PR #630, opened 2024-08-07
Add support for using AffineQuantizedTensor with `weights_only=True`

jerryzh168:
Summary:
`torch.load(file, weights_only=True)` is safer so ideally we can use that, by default it does not work with tensor subclasses, since now we have https://pytorch.org/docs/main/notes/serialization.html#torch.serialization.add_safe_globals we can add all tensor subclass classes and special types to globals so that these can work with `weights_only=True`

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/quantization/quant_api.py
+++ b/torchao/quantization/quant_api.py
@@ -48,6 +48,7 @@ from .GPTQ import (
 from .utils import _get_per_token_block_size
 import logging
 from .autoquant import autoquant, AutoQuantizableLinearWeight
+from torchao.utils import TORCH_VERSION_AFTER_2_5
 
 
 __all__ = [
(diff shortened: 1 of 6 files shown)


mikaylagawarecki (approved):
LGTM

torchao/quantization/quant_api.py
```diff
@@ -466,3 +481,7 @@ def int8_dynamic_activation_int8_semi_sparse_weight():
     """
     from torchao.dtypes import SemiSparseLayoutType
     return int8_dynamic_activation_int8_weight(layout_type=SemiSparseLayoutType())
+
+
+if TORCH_VERSION_AFTER_2_5:
```
mikaylagawarecki, line 486: A n00b qn: it wasn't super clear to me why these functions are used during unpickling

but I trust that you added them because they appeared in the error message :)

jerryzh168: yeah that's correct, these functions are serialized as well when we do dynamic quantization

msaroufim (approved):

PR #634, opened 2024-08-08
Refactor `_quantized_linear` for better extensibility

jerryzh168:
Summary:
Some popular ops like linear will get a lot of implementations based on the different characteristics of input and weight, e.g. int8 act + int8 weight, int8 act + int4 weight etc. For `AffineQuantizedTensor` rigth now all of these implementations are added to the main body of the implementation of linear dispatch, this makes the code hard to read and extend. 

We refactored the implementation for `_quantized_linear` op to take a list of (dispatch_condition, impl) and go through them one by one, this makes the body of _quantized_linear shorter and easier to maintain.

Alternatively we could also add more functionality to `implements`, e.g. add a secondary dispatch condition: `implements(func, dispatch_condition)`, but that is a much more complicated discussion that we can delay for later. a few questions we need to think about are:
* how do we allow people to override all implementations for a specific function? (used in autoquant)
* how do we make sure the dispatch condition people registered are called at the right order? e.g. if we have static quant, weight only quant, activation activation quant implementation/conditions, static quant (two inputs quantized) should come before the others, this might mean we also have to introduce the concept of secondary dispatch key here
* what happens to the dispatch table during inheritance? currently I think the dispatch table is just shared between parent and child, but if needed, we can make the dispatch table to be keyed on class as well and copy paste the dispatch table when a child class inherits it: ```cls._DISPATCH_TABLE[cls][func] = impl```

Test Plan:
regression tests

python tutorials/quantize_vit/run_vit_b_quant.py

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/dtypes/affine_quantized_tensor.py
+++ b/torchao/dtypes/affine_quantized_tensor.py
@@ -25,6 +25,7 @@ from torchao.dtypes.utils import (
     PlainLayoutType,
     is_device,
 )
+from torch.utils._python_dispatch import is_traceable_wrapper_subclass
 from dataclasses import dataclass
 from torchao.utils import TORCH_VERSION_AFTER_2_5
 
@@ -75,6 +76,10 @@ class AQTLayout(torch.Tensor):
 ##############################
 # Tensor Subclass Definition #
 ##############################
+
+_QLINEAR_DISPATCH_TABLE = {}
+def _register_quantized_linear_dispatch(dispatch_condition, impl):
+    _QLINEAR_DISPATCH_TABLE[dispatch_condition] = impl
 
 class AffineQuantizedTensor(torch.Tensor):
     """
@@ -157,7 +162,11 @@ class AffineQuantizedTensor(torch.Tensor):
 
     @staticmethod
     def _quantized_linear_op(input_tensor, weight_tensor, bias):
-        return _quantized_linear_op(input_tensor, weight_tensor, bias)
+        for dispatch_condition, impl in _QLINEAR_DISPATCH_TABLE.items():
+            if dispatch_condition(input_tensor, weight_tensor, bias):
+                return impl(input_tensor, weight_tensor, bias)
+
+        raise NotImplementedError("No specialized dispatch found for quantized linear op")
 
     def __tensor_flatten__(self):
         return ["layout_tensor"], [self.block_size, self.shape, self.quant_min, self.quant_max, self.zero_point_domain, self.dtype]
@@ -426,7 +435,7 @@ class PlainAQTLayout(AQTLayout):
         if func is aten.t.default:
             tensor = args[0]
             new = tensor.__class__(
-                tensor.int_data.view(tensor.shape[::-1]), tensor.scale, tensor.zero_point
+                tensor.int_data.view(tensor.shape[::-1]), tensor.scale, tensor.zero_point, tensor.layout_type
             )
             return return_and_correct_aliasing(func, args, kwargs, new)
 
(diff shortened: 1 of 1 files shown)


jerryzh168:
cc @kimishpatel we discussed about this a bit before, hopefully this makes extension outside of the repo easier

torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -157,7 +158,16 @@ def dequantize(self, output_dtype=None):
 
     @staticmethod
     def _quantized_linear_op(input_tensor, weight_tensor, bias):
-        return _quantized_linear_op(input_tensor, weight_tensor, bias)
+        cls = AffineQuantizedTensor
```
HDCharles, line 161: this feels really hacky, this should be a static method but its being hacked into a class method

torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -157,7 +158,16 @@ def dequantize(self, output_dtype=None):
 
     @staticmethod
     def _quantized_linear_op(input_tensor, weight_tensor, bias):
```
HDCharles, line 160: where does _quantized_linear_op get called now? The code flow feels much harder to follow now.

torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -666,238 +676,315 @@ def _aqt_is_uint4(aqt):
...
+def _register_mm_dispatch_helper(dispatch_condition, impl):
+    def _extract_mm_arg_and_apply(fn):
+        def f(func, types, args, kwargs):
+            input_tensor, weight_tensor, bias = (
+                args[0],
+                args[1],
+                None
+            )
+            return fn(input_tensor, weight_tensor.t())
+        return f
```
HDCharles, line 901: feels like these should be in utils if implements is

torchao/dtypes/affine_quantized_tensor.py
jerryzh168: I feel this should probably be moved to autoquant since it's only used there

HDCharles: we need a consistent way to extent aqt kernels and this is what we aligned on.

HDCharles: as is, this just breaks autoquant.

HDCharles: > I feel this should probably be moved to autoquant since it's only used there

It was used in aqt, its only used in autoquant in your PR because you removed it from aqt.

jerryzh168: I think moving this to autoquant is a small change? meaning subclass AffineQuantizedTensor and add this staticmethod, why does this break autoquant?

torchao/dtypes/affine_quantized_tensor.py
jerryzh168: yeah this is not called in AffineQuantizedTensor right now. I see, OK let me think about how we can make this work then

torchao/dtypes/affine_quantized_tensor.py
jerryzh168: does this has to be called in aqt? I didn't see why we have to do this from autoquant code

jerryzh168: yeah I understand, I think we can provide some other extension point for autoquant to accommodate this change, if people just keep adding things to `_quantized_linear_op` the op is just not going to be maintainable, also it will not possible for people to add new layout and dispatches to affine quantized tensor outside of torchao library

kimishpatel:
I need to look at the PR in more detail but why are we not proposing "extending" AffineQuantizedTensor class itself as a better solution. This allows for various packing and other things be also encoded by a class that wants different dispatch behavior. 

I there a reason why AffineQuantizedTensor class should take on this responsibility?

torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -157,7 +158,16 @@ def dequantize(self, output_dtype=None):
 
     @staticmethod
     def _quantized_linear_op(input_tensor, weight_tensor, bias):
-        return _quantized_linear_op(input_tensor, weight_tensor, bias)
+        cls = AffineQuantizedTensor
+        func = torch.nn.functional.linear
+        args = (input_tensor, weight_tensor, bias)
+        kwargs = {}
+        if hasattr(cls, "_DISPATCH_TABLE_WITH_CONDITION"):
```
kimishpatel, line 165: This feels like aqt expanding on its responsibility with some APIs whose need seems to have popped up now. I do feel AffineQuantizedTensor should not take on responsibility of dispatch at all, or if it does, it should just be dq-op-q. Rest should be by derived class.

I guess here you are asking users to register their favorite dispatch but I feel we are just implementing dispatcher now. I suspect subclass approach might be cleaner

jerryzh168: AQT have to do dispatch because cpu/cuda or other devices code path should be optimized by default I think

jerryzh168: but thinking about this dispatch table, it does feel very complicated. I'm planning to revert his part of the change, and just have this as a change to refactor `_quantized_op` function for now, we can discuss later about the support for lower bit kernels

jerryzh168:
> I need to look at the PR in more detail but why are we not proposing "extending" AffineQuantizedTensor class itself as a better solution. This allows for various packing and other things be also encoded by a class that wants different dispatch behavior.
> 
> I there a reason why AffineQuantizedTensor class should take on this responsibility?

so packing can be natively supported in the tensor as an attribute right now, it feels more complicated to do subclass than just adding an attribute? it might help to see the impact on all use cases though.

torchao/dtypes/affine_quantized_tensor.py
kimishpatel: > AQT have to do dispatch because cpu/cuda or other devices code path should be optimized by default I think

That is exactly what expanding responsibility is honestly. And this is why we end up with trying to figure out how to extend to accommodate different dispatch needs within AQT itself.

kimishpatel:
> this is just following what `torch.Tensor` is doing

Ok I need to understand this better perhaps

HDCharles:
looks ok to me, I think with these changes (especially the last version) we may want to add a unit test to check that the expected kernels are being used in autoquant, rather than what would normally be there from the dispatch. May be a todo for if we try a more complicated dispatch setup again.

HDCharles (approved):

jerryzh168:
> looks ok to me, I think with these changes (especially the last version) we may want to add a unit test to check that the expected kernels are being used in autoquant, rather than what would normally be there from the dispatch. May be a todo for if we try a more complicated dispatch setup again.

do we already have the check here: https://github.com/pytorch/ao/blob/acfb0da668121fc60a9e376213310abfa6bf92c5/test/integration/test_integration.py#L934

PR #635, opened 2024-08-08
Move Uintx out of prototype for future extension

jerryzh168:
Summary:
Thanks @vayuda for adding the initial version of Uintx tensor subclass we can now integrate this with `torch.uint1` to `torch.uint7` dtypes with some helpers to unblock the benefit of bitpacking (model size saving) to people first, and then we can gradually optimize the performance.

Also executorch is planning to integrate their low bit kernels with us, more native experience with these lower bit types will be required / useful there as well

Test Plan:
python test/dtypes/test_uintx.py

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/dtypes/uintx/Uintx.py
+++ b/torchao/dtypes/uintx/Uintx.py
@@ -1,17 +1,12 @@
-import functools
-import math
-from collections import defaultdict
-from typing import Any, Callable, Dict, Optional, Tuple, Union, List
+from typing import Tuple, List
 from dataclasses import dataclass
 import torch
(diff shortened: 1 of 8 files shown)


vayuda (approved):

torchao/dtypes/uintx/Uintx.py
vayuda, line 1: unused import _register_layout_cls from torchao.dtypes.utils
I used the function in AQT instead

msaroufim (approved):
LG, mind also updating the README here https://github.com/pytorch/ao#community-contributions

Issue #624, opened 2024-08-07
aqt peak memory performance worse than subclass apis


HDCharles:
see

https://github.com/pytorch/ao/pull/623

the new aqt apis have similar runtime perf but worse peak memory than the old apis for int8 dynamic and int4wo quant

probably this is due to needing to transpose the weight tensor for int8 dynamic, unsure for int4wo

Issue #554, opened 2024-07-29
Quantized Training


msaroufim:
Inspired by a recent back and forth with @gau-nernst 

Character.ai recently shared that they're working on quantized training https://research.character.ai/optimizing-inference/ where per @stephenroller they train models from scratch in int8 https://x.com/stephenroller/status/1816636257717436779

Historically we've invested more in QAT which @andrewor14 has led which is more of a technique to reduce perplexity when we do an eventual post training quantization.

Quantized training on the other hand actually quantizes the model at training time and so memory savings are observed both for training and inference

So when discussing quantized training there's a few aspects
1. Weights they can be in one: fp16, fp8, int8, int4 and below
3. Optimizer can be in one of: fp32, fp16, bf16, fp8, int8 and below

And if one were to ship this work, a bad combination can be validated at small scale (~600M parameter range) but a good idea needs to continuously be tested from (8b to 405b range) so each of these will need loss curves 

When choosing the starting point, we could either pretrain a model using quantized training or just finetune it and as long as the loss curves match the fp16 baselines then we are good. We'd also need to of course validate that memory savings are there and what the speedups/slowdowns are. 

And while we can merge a lot of the dtype conversion in AO and have some toy training loop in AO what I'm more optimistic about is having some end to end trainig recipe in https://github.com/pytorch/torchtitan @awgu  and an end to end finetuning recipe https://github.com/pytorch/torchtune @ebsmothers @joecummings

gau-nernst:
Just want to add, there is also activation/computation dtype and gradient dtype. In my exploration, I still use activation/computation in BF16 and gradient in BF16 to match weight-only quant inference. Activation/computation can be in lower precision dtype also, such as INT8 act - INT8 weight to match dynamic quantization inference, or FP8 act - FP8 weight to match current FP8 training recipe.

Lower precision gradient might not be possible? Will need to check existing works on this.

gau-nernst:
Some extra info for future reference

- Quantized training in JAX: [blog post](https://cloud.google.com/blog/products/compute/accurate-quantized-training-aqt-for-tpu-v5e) [github](https://github.com/google/aqt)
- INT8 training by Q-GaLore: [paper](https://arxiv.org/abs/2407.08296) [github](https://github.com/VITA-Group/Q-GaLore/blob/main/q_galore_torch/utils/quantization.py)

For evaluating the effectiveness of quantized training

[...]

PR #644, opened 2024-08-09
Add experimental INT8 quantized training

gau-nernst:
Address #554 (but don't close it)

This PR upstreams some of the exploration work done in https://github.com/gau-nernst/quantized-training

## Introduction (from the new README)

This folder contains experimental work on quantized training (QT). The main difference from quantization-aware training (QAT) is that in QT, we don't keep a high-precision copy of model weights. We take inspirations from:
- Q-GaLore: [[paper](https://arxiv.org/abs/2407.08296)] [[code](https://github.com/VITA-Group/Q-GaLore)]
- AQT: [[related paper](https://arxiv.org/abs/2105.03536)] [[code](https://github.com/google/aqt)]

Typically, low-precision weights cannot be trained directly due to quantization error: a small change in the quantized weight will be round down to zero. To tackle this problem, we use **stochastic rounding** for weight update. In simple terms, stochastic rounding will round up or down randomly, but with a higher chance if it is closer to that direction. For example, 0.8 will have 80% chance of rounding up and 20% of rounding down. It also follows that on average, stochastic rounding will estimate the floating point value exactly.

There are 2 main benefits for training in this way:
1. Reduce memory footprint. Also reduce communication bandwidth in distributed setting.
2. What you train is what you serve (WYTIWYS).

Currently we only support weight-only channel-wise INT8 symmetric quantization.

In this recipe, all linear weights are quantized to INT8 using channel-wise symmetric quantization `[-127, 127]`. In the forward and backward pass, the weights are upcast to activations' dtype (e.g. BF16). Therefore, their gradients are also in activations' dtype.

Usage

```python
from torchao.prototype.quantized_training import int8_weight_only_quantized_training
from torchao.prototype.low_bit_optim import AdamW

model = ...
quantize_(model, int8_weight_only_quantized_training())

optim = AdamW(model.parameters(), lr=3e-4)
```

It is recommended to use optimizers from `torchao.prototype.low_bit_optim` for quantized training, because they can automatically generate efficient fused optimizer kernel for `dequant->optimizer_step->quant` thanks to `torch.compile()`.

## Results

**Training loss**. Run the included benchmark. This launches an LLM pre-training using a Llama2-style model with 470M parameters on [TinyStories](https://huggingface.co/datasets/roneneldan/TinyStories). Add `--quantize int8_weight_only` for INT8 quantized training.

```
python python benchmarks/quantized_training/pretrain_llama2.py --seed 2024 --n_steps 10_000 --compile
```

image

**Memory benchmark**. 1B model; bs=4, seq_len=2048 -> 8192 toks/iter; activation checkpointing

Model | Peak memory (GB)
------|-----
BF16 eager | 11.06847
BF16 compile | 10.16915
INT8 QT eager | 10.11437
INT8 QT compile | 10.03365

In eager mode, the reduction in memory looks correct (1B model -> 1GB reduction from BF16->INT8). However, in compile mode, there is not a lot of memory reduction compared to BF16 compile. Maybe related to transposed weight? #624
 
## Extra results 

Some extra results from https://github.com/gau-nernst/quantized-training that was done before this PR.

**Llama2-style LLM pre-training on TinyStories**

470M params | 1B params
--------------|------------
image | image

The gap is much smaller for 1B params, indicating that larger model is easier to quantize (maybe dependent on the dataset too, perhaps 1B model is too big for TinyStories). Validation loss has the same gap as training loss.

Model | TinyStories validation loss
------|-------
470M BF16 | 0.9904
470M BF16 + INT8 PTQ | 0.9904
470M INT8 QT | 1.0389
470M INT8 QT + 8-bit Adam | 1.0927
1B BF16 | 0.9898
1B BF16 + INT8 PTQ | 0.9898
1B INT8 QT | 1.0000

**ViT fine-tuning**

Fine-tune [`timm/vit_giant_patch14_dinov2.lvd142m`](https://huggingface.co/timm/vit_giant_patch14_dinov2.lvd142m) (1B params) on [RESISC45](https://huggingface.co/datasets/timm/resisc45)

image

Model | RESISC45 val acc
-------|--------
BF16 model | 93.94%
BF16 model + INT8 PTQ | 93.92%
INT8 QT model | 92.40%

**LLM fine-tuning**

Fine-tune [SmolLM-1.7B](https://huggingface.co/HuggingFaceTB/SmolLM-1.7B) on [MetaMathQA](https://huggingface.co/datasets/meta-math/MetaMathQA)

image

Using LR=1e-5, only INT8 QT model can learn, while BF16 model cannot learn at all, probably due to LR being too small (BF16 only has ~3 decimal precision). Increasing LR to 1e-4, BF16 model can be trained now. Due to different LR, it's hard to compare BF16 LR=1e-4 and INT8 QT LR=1e-5 directly. Will update this if I have time to re-run INT8 QT with LR=1e-4.

Model | GSM8K acc
---|---
BF16 model (LR=1e-4) | 14.86
INT8 QT model (LR=1e-5) | 19.11

## Note on optimizer

Optimizer logic is still done in high precision (FP32 or BF16). To minimize errors, we should only re-quantize params once, at the end of the optimizer logic. However, existing PyTorch optimizers may modify the param in-place multiple times. e.g. AdamW

https://github.com/pytorch/pytorch/blob/32be3e942c3251dc50892334c6614a89327c122c/torch/optim/adamw.py#L384
 
Therefore, this PR also adds an alternative implementation of AdamW in `torchao.prototype.low_bit_optim.AdamW`, which only applies in-place update of param in the final step.

## Future ideas

- INT8 activation x INT8 weight. This can potentially leverage INT8 Tensor Cores, which is 2x faster than FP16/BF16 Tensor Cores.
- INT4 weight only (with group-wise quantization). This can be used with INT4 tinygemm deployment in mind (or other optimized INT4 kernels).
- FP8 activation x FP8 weight. The current FP8 training recipe can be seen as a form of QAT, which maintains a high-precision copy of model weights. We can eliminate the high-precision copy.


gau-nernst:
The failing test is related to #652 (custom optimizer + DTensor + torch.compile()). Pending fix.

benchmarks/benchmark_int8_qt.py
```diff
@@ -0,0 +1,148 @@
+# pre-train a mini Llama2 on TinyStories with INT8 quantized training
+# pip install transformers sentencepiece wandb
+#
+# BF16 baseline: python benchmarks/benchmark_int8_qt.py --seed 2024 --n_steps 10_000
+# INT8 QT:       python benchmarks/benchmark_int8_qt.py --seed 2024 --n_steps 10_000 --quantize int8_weight_only
+
+import os
+
+os.environ["PYTORCH_CUDA_ALLOC_CONF"] = "expandable_segments:True"
```
msaroufim, line 9: just curious were you running into ooms without this?

benchmarks/benchmark_int8_qt.py
```diff
@@ -0,0 +1,148 @@
...
+
+def get_loss(model: LlamaForCausalLM, batch: torch.Tensor):
+    return model(batch, labels=batch).loss
+
+
+def get_tinystories():
+    save_path = Path("tinystories.bin")
+
+    if not save_path.exists():
+        import sentencepiece as spm
```
msaroufim, line 33: perhaps create a seperate quantized training folder along with the dependencies for hf_hub and sentencepiece

benchmarks/benchmark_int8_qt.py
```diff
@@ -0,0 +1,148 @@
...
+
+    parser.add_argument("--project", default="int8_quantized_training")
+    parser.add_argument("--run_name")
+    parser.add_argument("--seed", type=int)
+    args = parser.parse_args()
+
+    if args.seed is not None:
+        torch.manual_seed(args.seed)
+
+    config = LlamaConfig(
```
msaroufim, line 93: did you also try the llama model we have in AO? CUrious what was more convenient here?

benchmarks/benchmark_int8_qt.py
```diff
@@ -0,0 +1,148 @@
...
+    if args.activation_checkpointing:
+        model.gradient_checkpointing_enable()
+    if args.quantize == "int8_weight_only":
+        quantize_(model, int8_weight_only_quantized_training())
+    elif args.quantize is not None:
+        raise ValueError(f"Unsupported quantize={args.quantize}")
+    print(f"No. of params: {sum(p.numel() for p in model.parameters()):,}")
+    print(f"No. of buffers: {sum(p.numel() for p in model.buffers()):,}")
+
+    # turn off these flags (set by quantize_()) to speed up compile time
```
msaroufim, line 111: huh interesting were the flags that bad? cc @HDCharles @jerryzh168 in case we need to change the defaults

benchmarks/benchmark_int8_qt.py
gau-nernst: I haven't. Will give jt a try

benchmarks/benchmark_int8_qt.py
gau-nernst: I think they are fine for inference. Maybe ok for training too, but I didn't want to wait 🤣

test/prototype/test_quantized_training.py
```diff
@@ -0,0 +1,225 @@
...
+            if compile_layer:
+                layer.compile()
+            fully_shard(layer)
+        fully_shard(fsdp_model)
+
+        base_optim = AdamW(base_model.parameters(), lr=1e-2)
+        fsdp_optim = AdamW(fsdp_model.parameters(), lr=1e-2)
+
+        # prevent segfault with torch.compile()
+        torch.set_float32_matmul_precision("highest")
```
msaroufim, line 194: interesting

torchao/prototype/low_bit_optim/adamw.py
```diff
@@ -150,8 +148,23 @@ def single_param_adamw(
...
+    def __init__(
+        self,
+        params,
+        lr=1e-3,
+        betas=(0.9, 0.999),
+        eps=1e-8,
+        weight_decay=1e-2,
+        amsgrad=False,
+    ) -> None:
+        """AdamW optimizer that supports quantized training (parameter is quantized)."""
```
msaroufim, line 166: Hmm did you explore other names? for this class?

torchao/prototype/quantized_training/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+This folder contains experimental work on quantized training (QT). The main difference from quantization-aware training (QAT) is that in QT, we don't keep a high-precision copy of model weights. We take inspirations from:
+- Q-GaLore: [[paper](https://arxiv.org/abs/2407.08296)] [[code](https://github.com/VITA-Group/Q-GaLore)]
+- AQT: [[related paper](https://arxiv.org/abs/2105.03536)] [[code](https://github.com/google/aqt)]
+
+Typically, low-precision weights cannot be trained directly due to quantization error: a small change in the quantized weight will be round down to zero. To tackle this problem, we use **stochastic rounding** for weight update. In simple terms, stochastic rounding will round up or down randomly, but with a higher chance if it is closer to that direction. For example, 0.8 will have 80% chance of rounding up and 20% of rounding down. It also follows that on average, stochastic rounding will estimate the floating point value exactly.
+
+There are 2 main benefits for training in this way:
+1. Reduce memory footprint. Also reduce communication bandwidth in distributed setting.
+2. What you train is what you serve (WYTIWYS).
```
msaroufim, line 11: curious where you got the terminology from?

msaroufim (approved):
Minor nits but LGTM, I see @bdhirsh is looking into the fsdp2 issue so let's wait and see if we should downgrade the pytorch pin before merge

torchao/prototype/quantized_training/int8.py
```diff
@@ -0,0 +1,247 @@
...
+
+from torchao.dtypes.utils import _dispatch__torch_dispatch__, _dispatch__torch_function__, _implements
+
+
+aten = torch.ops.aten
+c10d_functional = torch.ops.c10d_functional
+_c10d_functional = torch.ops._c10d_functional
+
+
+# the main difference of this tensor subclass from AffineQuantizedTensor:
```
msaroufim, line 15: perhaps @jerryzh168 has some feedback on similarities to AffineQuantizedTensor - granted i'd be fine if we punt on unification to another PR

torchao/prototype/quantized_training/README.md
gau-nernst: From Google https://github.com/google/aqt?tab=readme-ov-file#features. First used in their first AQT paper I think (https://arxiv.org/pdf/2105.03536, section 3.3, pointer 5). The Character.AI also alludes to this, though not using the exact words. Actually QAT can be considered "What you train is what you serve too", though Google might not think so 😄.

test/prototype/test_quantized_training.py
gau-nernst: Probably fixed with this https://github.com/pytorch/pytorch/pull/133173

benchmarks/benchmark_int8_qt.py
gau-nernst: Not for the experiments I ran here I think. I copied it over from my repo, and I think this flag is generally good.

torchao/prototype/quantized_training/int8.py
jerryzh168: yeah I feel we should just enable this in `AffineQuantizedTensor

torchao/prototype/quantized_training/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+Usage
+
+```python
+from torchao.prototype.quantized_training import int8_weight_only_quantized_training
+from torchao.prototype.low_bit_optim import AdamW
+from torchao.quantization.quant_api import quantize_
+
+model = ...
+quantize_(model, int8_weight_only_quantized_training())
```
jerryzh168, line 27: ah this is nice

test/prototype/test_quantized_training.py
```diff
@@ -0,0 +1,225 @@
...
+            loss_fp32.backward()
+            optim_fp32.step()
+            optim_fp32.zero_grad()
+
+            loss_int8.backward()
+            optim_int8.step()
+            optim_int8.zero_grad()
+
+
+class TestFSDP2(FSDPTest):
```
jerryzh168, line 146: it feels like we could probably add this as part of standard test suite, that we can use to sanity check if FSDP is supported for any dtype/tensor subclasses

benchmarks/benchmark_int8_qt.py
jerryzh168: you can turn it off setting `set_inductor_config=False`: https://github.com/pytorch/ao/blob/e7fc0ed0f2e130e44d250541aba71c7a7a985f15/torchao/quantization/quant_api.py#L268

torchao/prototype/quantized_training/int8.py
```diff
@@ -0,0 +1,247 @@
...
+# 1. F.linear is differentiable i.e. backward is defined.
+# 2. support stochastic rounding when casting from floating point.
+class Int8QTLinearWeight(Tensor):
```
andrewor14, line 18: Can you add a docstring to include some details on what kind of int8 quant this is doing (e.g. symmetric, -127 to 127)? Also you can move the comment above into the docstring

torchao/prototype/quantized_training/int8.py
```diff
@@ -0,0 +1,247 @@
...
+        scale = tensor.abs().amax(-1) / 127
+        tensor = tensor / scale.clip(1e-12).view(-1, 1)
+
+        if stochastic_rounding:
+            tensor = (tensor + torch.rand_like(tensor)).floor()
+        else:
+            tensor = tensor.round()
+
+        # NOTE: is clipping necessary?
+        tensor = tensor.clip(-128, 127).to(torch.int8)
```
andrewor14, line 68: actually is it -127 or -128? From your PR description it says -127

torchao/prototype/quantized_training/int8.py
```diff
@@ -0,0 +1,247 @@
...
+
+def int8_weight_only_quantized_training():
+    def apply_int8_linear_weight(linear: nn.Linear):
+        linear.weight = nn.Parameter(
+            Int8QTLinearWeight.from_float(linear.weight),
+            requires_grad=linear.weight.requires_grad,
+        )
+        return linear
+
+    return apply_int8_linear_weight
```
andrewor14, line 247: I think you need `_get_linear_subclass_inserter` here

torchao/prototype/quantized_training/int8.py
```diff
@@ -0,0 +1,247 @@
...
+    dtype = kwargs.get("dtype", args[0].dtype)
+    int_data = torch.zeros(size, device=device, dtype=torch.int8)
+    scale = torch.zeros(size[0], device=device, dtype=dtype)
+    return Int8QTLinearWeight(int_data, scale)
+
+
+# don't do anything. workaround for FSDP2. might give unexpected or wrong results.
+@Int8QTLinearWeight.implements([aten.view.default, aten.as_strided.default])
+def _(func, types, args, kwargs):
+    out = Int8QTLinearWeight(args[0].int_data, args[0].scale)
```
andrewor14, line 235: Can you explain why we don't need to do anything here? What happens if we call view on the inner tensors and return a copy?

torchao/prototype/quantized_training/int8.py
```diff
@@ -0,0 +1,247 @@
...
+
+
+aten = torch.ops.aten
+c10d_functional = torch.ops.c10d_functional
+_c10d_functional = torch.ops._c10d_functional
+
+
+# the main difference of this tensor subclass from AffineQuantizedTensor:
+# 1. F.linear is differentiable i.e. backward is defined.
+# 2. support stochastic rounding when casting from floating point.
```
andrewor14, line 17: I think `stochastic_rounding=False` when casting from floating point? (L73)

torchao/prototype/quantized_training/int8.py
```diff
@@ -0,0 +1,247 @@
...
+    def __tensor_flatten__(self):
+        return ["int_data", "scale"], []
+
+    @classmethod
+    def __tensor_unflatten__(cls, tensor_data_dict, tensor_attributes, outer_size=None, outer_stride=None):
+        return cls(tensor_data_dict["int_data"], tensor_data_dict["scale"], *tensor_attributes)
+
+    @staticmethod
+    @torch.no_grad()
+    def quantize(tensor: Tensor, stochastic_rounding: bool = False):
```
andrewor14, line 54: Would be good to add some comments here to explain why we need stochastic rounding and when. I saw you have some explanation in your README. Maybe good to add some of that here as well

torchao/prototype/quantized_training/int8.py
```diff
@@ -0,0 +1,247 @@
...
+        else:
+            tensor = tensor.round()
+
+        # NOTE: is clipping necessary?
+        tensor = tensor.clip(-128, 127).to(torch.int8)
+        return tensor, scale.to(original_dtype)
+
+    @classmethod
+    def from_float(cls, tensor: Tensor):
+        """Convert a float tensor into INT8 quantized weight. No stochastic rounding is performed.
```
andrewor14, line 73: What would happen if we just always apply stochastic rounding?

torchao/prototype/quantized_training/README.md
```diff
@@ -0,0 +1,42 @@
+# Quantized training
```
andrewor14, line 1: This is great! Thanks for adding such a detailed README

andrewor14 (approved):
Thanks @gau-nernst, looks great! One question I have is do the forward numerics match the inference path? E.g.
```
quantize_(model, int8_weight_only())  # vs
quantize_(model, int8_weight_only_quantized_training())
```
Do we want them to match?

Also, for Llama2, do you have the eval accuracies/perplexities compared to bf16 training?

torchao/prototype/quantized_training/int8.py
gau-nernst: `_get_linear_subclass_inserter()` will set `requires_grad=False`

https://github.com/pytorch/ao/blob/e7fc0ed0f2e130e44d250541aba71c7a7a985f15/torchao/quantization/quant_api.py#L261-L266

We can add `requires_grad` as extra argument to `_get_linear_subclass_inserter()`, but considering this is prototype/experimental, I would say leave it for future PR. Wdyt?

torchao/prototype/quantized_training/int8.py
gau-nernst: In terms of semantics, I think it's better to keep `from_float()` to perform normal rounding, as it is a dtype conversion. By right, stochastic rounding is only done in the optimizer step -> should be part of the optimizer logic. We use tensor subclass here so it's easier to implement.

In practice, once we do training, I think it shouldn't matter much if we stochastic round the first dtype conversion or not.

torchao/prototype/quantized_training/int8.py
gau-nernst: These ops are called by FSDP2 by these lines (there are a few mores, you can ctrl+F, but these are the first hits when FSDP2 runs)

https://github.com/pytorch/pytorch/blob/cd565bc45554f6c7d5659bff19a270bda58ce71d/torch/distributed/_composable/fsdp/_fsdp_param.py#L349
https://github.com/pytorch/pytorch/blob/cd565bc45554f6c7d5659bff19a270bda58ce71d/torch/distributed/_composable/fsdp/_fsdp_param.py#L466-L471

From what I understand, FSDP2 expects a "view" here i.e. not a copy. Since we do channel-wise quantization here, a view doesn't quite make sense since it changes the dimensions -> quantization scale should change too (i.e. group size is tied to tensor dims). Propagate the ops to inner tensor may not make much sense too. For example, `as_strided` can't be called on `.scale`, since scale has different shape from the outer tensor.

I looked into NF4 impl, and it seems like they also keep the inner tensors as is, only update the kwargs passing to `Tensor._make_wrapper_subclass()`, which change the outer tensor appearance.

https://github.com/pytorch/ao/blob/e7fc0ed0f2e130e44d250541aba71c7a7a985f15/torchao/dtypes/nf4tensor.py#L230-L270

So basically, it's only a hack to make it work with FSDP2. Normally we shouldn't call these ops at all! I don't know if passing the "expected kwargs" to `Tensor._make_wrapper_subclass()` is necessary. The FSDP2 test passes on my machine (with an earlier version of pytorch nightly), so I think it might be ok? This is why I also request review from @awgu to check what is the best practice / recommended way here.

supriyar:
> There are 2 main benefits for training in this way:

> Reduce memory footprint. Also reduce communication bandwidth in distributed setting.
> What you train is what you serve (WYTIWYS).

For 1, for the experiments run in this PR, what is the reduction in memory footprint vs bf16 training?

For 2, do you have any accuracy comparisons of the model (inference time) vs bf16

gau-nernst:
@andrewor14 @supriyar 

> do you have the eval accuracies/perplexities compared to bf16 training?
> do you have any accuracy comparisons of the model (inference time) vs bf16

I added eval results in the PR description. Reproduce it here for convenience. Also added PTQ eval.

Model | TinyStories validation loss
------|-------
470M BF16 | 0.9904
470M BF16 + INT8 PTQ | 0.9904
470M INT8 QT | 1.0389
470M INT8 QT + 8-bit Adam | 1.0927
1B BF16 | 0.9898
1B BF16 + INT8 PTQ | 0.9898
1B INT8 QT | 1.0000

For INT8 PTQ vs BF16, the loss only differs by the 5th decimal point. I double-checked to make sure I didn't make any mistakes. So at the current state, INT8 PTQ is still better, though perhaps INT8 QT needs slightly different hparams (e.g. larger LR).

> One question I have is do the forward numerics match the inference path

Currently they don't match for the following reasons
1. **Precision**: AQT keeps original dtype when doing quantization, while I upcast to FP32 before quantization, and downcast after that
2. **Calculate scale**: AQT uses `input.abs().amax() / 127.5`, while I use `input.abs().amax() / 127`
3. **Apply the scale**: AQT uses `input * (1 / scale)`, while I use `input / scale`

I checked that once I change my implementation to match AQT on the above 3 pointers, the inference results match exactly in eager mode.

Pointer 2 is especially important for quantized training. Empirical results from my previous experiments show that training will not converge if 127.5 is used. Likely because there will be clipping in the +ve side (i.e. +127.5 -> +127), which bias the results.

> for the experiments run in this PR, what is the reduction in memory footprint vs bf16 training?

Using 1B model; bs=4, seq_len=2048 -> 8192 toks/iter; activation checkpointing (I also put this table in the PR description)

Model | Peak memory (GB)
------|-----
BF16 eager | 11.06847
BF16 compile | 10.16915
INT8 QT eager | 10.11437
INT8 QT compile | 10.03365

In eager mode, the reduction in memory looks correct (1B model -> 1GB reduction from BF16->INT8). However, in compile mode, there is not a lot of memory reduction compared to BF16 compile. Maybe related to transposed weight? https://github.com/pytorch/ao/issues/624

andrewor14:
@gau-nernst Thanks for the detailed evaluation and explanation! I have a couple of follow-up questions but I'm OK with merging this as a prototype, but we should probably talk about the above caveats either in the README or in inline comments. Also curious if others have an opinion on this (cc @msaroufim @jerryzh168).

> I checked that once I change my implementation to match AQT on the above 3 pointers, the inference results match exactly in eager mode.

That's great. Can you include these 3 in the comments for your tensor subclass? This will help us when we refactor to use AQT in a future PR. 

> Pointer 2 is especially important for quantized training. Empirical results from my previous experiments show that training will not converge if 127.5 is used. Likely because there will be clipping in the +ve side (i.e. +127.5 -> +127), which bias the results.

That seems a bit fragile. Do you know why this is the case? Is there anyone else in the community that has reported something similar?

> In eager mode, the reduction in memory looks correct (1B model -> 1GB reduction from BF16->INT8). However, in compile mode, there is not a lot of memory reduction compared to BF16 compile. Maybe related to transposed weight? https://github.com/pytorch/ao/issues/624

Please add a TODO to investigate this. I think it's worth looking into because saving memory is an important reason for QT in the first place.

msaroufim:
also if you need to add any skip tests to get everything green let's do that, merge and just list out the remaining work that needs to be done in issues.

gau-nernst:
> That seems a bit fragile. Do you know why this is the case? Is there anyone else in the community that has reported something similar?

I suspect it's due to clipping. E.g. using 127.5, when you have 2 values [-1,+1] -> scale = 127.5, values are scaled to [-127.5,+127.5]. +127.5 can stochastic round to 127 or 128, but due to INT8 clipping, it will always be clipped to 127. In the end, `E[SR(x)] != x`. This will only happen to the largest positive value. Though it does seem surprising that this small difference can make such a big impact.

I'm not aware of works reporting this phenomenon. https://github.com/google/aqt also uses 127, though they are actually doing QAT. In terms of PTQ, not an expert but I think many libraries also uses 127 instead of 127.5? E.g. https://github.com/ggerganov/llama.cpp/blob/4b9afbbe9037f8a2d659097c0c7d9fce32c6494c/ggml/src/ggml-cuda/quantize.cu#L27. Though for PTQ, it's probably doesn't matter much whether 127 or 127.5 are used.

torchao/prototype/quantized_training/int8.py
andrewor14: Sounds good, can you add a comment in the code to explain this?

torchao/prototype/quantized_training/int8.py
andrewor14: Ok, add a TODO?

andrewor14: Ok, add a TODO comment?

torchao/prototype/quantized_training/int8.py
```diff
@@ -0,0 +1,271 @@
...
+        to FP32 before quantization, and downcast scale to original dtype.
+        2. Calculate scale: AQT uses `input.abs().amax() / 127.5`, while `input.abs().amax() / 127` is
+        done here.
+        3. Apply scale: AQT uses `input * (1 / scale)`, while this function performs `input / scale`.
+        """
+        original_dtype = tensor.dtype
+        tensor = tensor.float()
+
+        # absmax symmetric quantization
+        scale = tensor.abs().amax(-1) / 127
```
jerryzh168, line 77: so does this mean the quant_min/quant_max for int8 will be (-127, 127)? we could match this in post training quant by explicitly setting quant_min/quant_max I think

PR #645, opened 2024-08-09
float8 profiling: save torch_logs, and attach them to profiling trace

vkuzo:
Summary:

Adds two extra things to benchmarks/float8/profile_linear_float8.py:
1. for each profiling run, also saves the output of `TORCH_LOGS="output_code"`
2. for each profiling trace, creates a modified trace with the inductor code from (1) attached to triton kernel GPU events, which enables the user to click on the triton kernels in `chrome://tracing/` or perfetto, and see the triton code inline in the browser - no need to cross reference against logs anymore.

Screenshot 2024-08-09 at 10 25 27 AM

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:


y-sq (approved):
lgtm! Thanks for the changes, it will be very useful to view the triton kernels in the profiling trace.

PR #650, opened 2024-08-09
Add AffineQuantizedObserver

jerryzh168:
Summary:
In our static_quant flow tutorial we were still using observers from `torch.ao` which we plan to deprecate, this PR adds a more general observer for `AffineQuantizedTensor`, and has shown that we can replace the old observers (min max observer), there could be futhre work to improve perf, add new types of observation, e.g. tracking stats other than just min/max, moving average observer, histogram observer.

Test Plan:
python test/quantization/test_observer.py
python tutorials/calibration_flow/static_quant.py

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/quantization/observer.py
+++ b/torchao/quantization/observer.py
@@ -1,0 +1,166 @@
+import torch
+from .quant_primitives import (
+    _get_reduction_params,
+    choose_qparams_affine_with_min_max,
+    MappingType,
+    ZeroPointDomain,
+)
+
+from abc import ABCMeta, abstractmethod
+from dataclasses import dataclass
+from typing import Callable, List, Tuple, Optional, Any
+from functools import partial
+import logging
+logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True)
+class GranularityType:
+    pass
+
+@dataclass(frozen=True)
+class PerTensor(GranularityType):
+    pass
+
+@dataclass(frozen=True)
+class PerAxis(GranularityType):
+    axis: int
+
+# borrowed from torch.ao.quantization.observer
+class _PartialWrapper:
+    def __init__(self, p):
+        self.p = p
+
+    def __call__(self, *args, **keywords):
+        return self.p(*args, **keywords)
+
+    def __repr__(self):
+        return self.p.__repr__()
+
+    def with_args(self, *args, **kwargs):
+        return _with_args(self, *args, **kwargs)
+
+def _with_args(cls_or_self, *args, **kwargs):
+    r"""Wrapper that allows creation of class factories.
+
+    This can be useful when there is a need to create classes with the same
+    constructor arguments, but different instances.
+
+    Example::
+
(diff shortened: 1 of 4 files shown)


torchao/quantization/observer.py
```diff
@@ -0,0 +1,169 @@
...
+def get_block_size(input_shape: Tuple[int, ...], granularity_type: GranularityType) -> Tuple[int, ...]:
+    if isinstance(granularity_type, PerTensor):
+        return input_shape
+    elif isinstance(granularity_type, PerAxis):
+        block_size = list(input_shape)
+        block_size[granularity_type.axis] = 1
+        return tuple(block_size)
+    raise ValueError(f"Unsupported GranularityType: {granularity_type}")
+
+class AffineQuantizedObserver(torch.nn.Module):
```
vayuda, line 70: Is the precedent to add observers here or within the quantization file they are used for. I also see that you had to add a new function to `quant_primitives`

vayuda (approved):
Looks good for AWQ workflow

torchao/quantization/observer.py
```diff
@@ -0,0 +1,169 @@
...
+            self.scale_dtype,
+            self.zero_point_dtype,
+            self.preserve_zero,
+            self.zero_point_domain
+        )
+
+    return update_stats_min_max, calculate_qparams_min_max
+
+_update_stats_min_max, _calculate_qparams_min_max = get_min_max_funcs()
+AffineQuantizedMinMaxObserver = AffineQuantizedObserver.with_args(_update_stats_min_max, _calculate_qparams_min_max)
```
andrewor14, line 169: What's the reason behind returning these functions as Callables instead of overriding them in the base class? I feel it's cleaner to do the latter (also done in the torch.ao flows) so we don't have to pass functions around

andrewor14 (approved):

torchao/quantization/observer.py
jerryzh168: OK sounds good, I don't have a very strong preference on this point, the initial motivation is for people to reuse the same `AffineQuantizedObserver` everywhere, but sounds fine for them to inherit as well

cpuhrsch:
Out of curiosity (and ignorance), but why do we need to couple "Observer" to a particular dtype? Doesn't it just track statistics of a particular Tensor over time? 

Or put differently, is there an example of / plans for a Bfloat16Observer or FP8Observer?

jerryzh168:
@cpuhrsch please let me know if you want me to separate the affine quantized tensor arg list and minmax observer now. it will be something like this:

```
# has the min_val/max_val logic
class MinMaxObserver(...):
   ...
   def forward(...):
       ...

# has the list of args
class AffineQuantizedObserver(...):
   def __init__(...):

class AffineQuantizedMinMaxObserver(AffineQuantizedObserver, MinMaxObserver):
   def calculate_qparams(...):
```

jerryzh168:
will land for now, we can follow up to split the logic when there are more observers I think

PR #653, opened 2024-08-10
skip fsdp2 low bit optim test in nightlies

msaroufim:
Temporary workaround for https://github.com/pytorch/ao/issues/652

I didn't debug this but seems like it's some dynamic shape issue - cc @ezyang @gau-nernst


gau-nernst:
The FSDP test only run in nightly, so if we skip, it won't run at all. Will try if I can debug it today.
Also, are you looking at adding 2.4 to CI?

gau-nernst:
If this test succeeds in 2.4 (observing #654), I will approve this. This is also affecting my #644, probably because I also use optimizer from `low_bit_optim` for the quantized training FSDP test.

gau-nernst (approved):

PR #485, opened 2024-07-08
Make version check equal to or more than instead of more than

msaroufim:
EDIT: This PR is causing a bunch of IMA issues on PyTorch 2.3 which is unexpected

This PR solves 2 issues

1. Our current version guard checks were incorrect `TORCH_VERSION_AFTER_2_3(2.3)` would return false for 2.3 and that's because we intended to say (see below for more detail) and adding .dev was obscuring the actual intent
2. The second issue is the name we were using was confusing so renamed `TORCH_VERSION_AFTER_X_X` to `TORCH_VERSION_AT_LEAST_X_X,`

```python
import torch
torch.__version__ # prints 2.3.0
from torchao.utils import torch_version_at_least
torch_version_at_least("2.3.0.dev) # was returning false for 2.3.0 but True for 2.3.1
```

But after changes in this PR

```python
torch_version_at_least("2.3.0.dev") # True
torch_version_at_least("2.3.0") # True
```

Also tested with 3.5

```
Python 3.10.14 (main, May  6 2024, 19:42:50) [GCC 11.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from torchao.utils import torch_version_at_least
>>> from torchao.utils import torch_version_at_least
KeyboardInterrupt
>>> torch_version_at_least("2.5.0.dev")
True
>>> torch_version_at_least("2.5.0")
True
>>> import torch
>>> torch.__version__
'2.5.0.dev20240708+cu121'
>>> 
>>> from torchao.utils import TORCH_VERSION_AFTER_2_5
>>> TORCH_VERSION_AFTER_2_5
```

PR #654, opened 2024-08-10
Add PyTorch 2.4 tests in CI

msaroufim:
EDIT: will need to fix this first before merging this https://github.com/pytorch/ao/pull/679#issuecomment-2290444713

Before this PR we were testing nightlies which recently moved from PyTorch 2.4 and 2.5 meaning we don't have coverage for 2.4 now

I would have also liked to remove support for PyTorch 2.2 with the policy being "we support latest stable pytorch, the version before that and nightlies"

A few tests needed to be changed
* Autoquant with our nightlies now fail on pytorch 2.4 stable - cc @jerryzh168 @HDCharles 
* fp8 tests were moved to nightly only per @vkuzo's request
* Mixed_mm test @msaroufim 
* fsdp2 low bit optim tests on 2.4 which means they are completely skipped now @gau-nernst 

But will not make this change suddenly, cc some partners in case this is not OK @ebsmothers @joecummings @kimishpatel @Jack-Khuu @awgu @malfet


msaroufim:
@jerryzh168 there's 12 failing tests for AffineQuantized Tensor with torch 2.4 https://github.com/pytorch/ao/actions/runs/10335226306/job/28609531090?pr=654

This is concerning cause it means our latest stable release does not work with latest pytorch stable release

vkuzo:
is there a way to move the float8 tests to PT 2.5 in this PR?  It would be nice to just support the latest version for now to move fast.  It would be just switching https://github.com/pytorch/ao/blob/261d0a4feff1731163140d469ae69edb8a1da34b/test/float8/test_base.py#L21 and comparable lines in `test/float8/*` to target 2.5 instead.

msaroufim:
> is there a way to move the float8 tests to PT 2.5 in this PR? It would be nice to just support the latest version for now to move fast. It would be just switching

Yeah will land this first since right now our version guard logic is incorrect https://github.com/pytorch/ao/pull/485

jerryzh168:
> @jerryzh168 there's 12 failing tests for AffineQuantized Tensor with torch 2.4 [pytorch/ao/actions/runs/10335226306/job/28609531090?pr=654](https://github.com/pytorch/ao/actions/runs/10335226306/job/28609531090?pr=654)
> 
> This is concerning cause it means our latest nightly release does not work with latest pytorch stable release. I'm worried we have a similar with our official 0.4 release not working with PyTorch 2.4.
> 
> If so We might be forced to do a patch release. Can you please sync with @jcaip and figure this out

I can repro, opened a PR for export team to take a look here: https://github.com/pytorch/pytorch/issues/133262

jcaip:
cc @msaroufim @jerryzh168 

Testing locally with torch==2.4.0 and torchao==0.4.0 I don't see any failures for `test_integration.py` (The test that's failing in CI). Is there an easy way to test in CI to make sure? 

Screenshot 2024-08-12 at 4 48 36 PM

Screenshot 2024-08-12 at 4 55 33 PM


I'm ~70% sure it's [this](https://github.com/pytorch/ao/blob/accbdba821549dbbdfbf691190eb6e46478337b3/torchao/prototype/uintx/Uintx.py#L91) line that's causing the bug, since it's the only use of `get_plain` after branch cut.

msaroufim:
Ok that's good news the bug is then likely with the version guard code. Thanks!

jerryzh168:
> Testing locally with torch==2.4.0 and torchao==0.4.0 I don't see any failures for `test_integration.py` (The test that's failing in CI). Is there an easy way to test in CI to make sure?

oh I'm testing with torchao nightly

jerryzh168 (approved):

PR #668, opened 2024-08-13
add roofline estimation of float8 gemm + overhead

vkuzo:
Summary:

This PR adds a script to help estimate when it is expected to be faster to convert a `torch.nn.Linear` to float8, versus leaving it in bfloat16, for compute bound training use cases.

How we do this:
1. for bf16 and fp8 gemm time, use either benchmarks or roofline based on a percentage of peak compute bandwidth
2. for float8 overhead, enumerate the worst case of reads and writes performed by torch.compile, and convert to time by assuming we can achieve a percentae of peak memory bandwidth
3. compare (a) bf16_gemm_time and (b) fp8_gemm_time + fp8_overhead_time, and calculate the expected speedup.

Note that currently the gemm benchmarks require running a separate script, as documented in the argument descriptions.  We can make this nicer at a future time.

Test Plan:

```
python benchmarks/float8/float8_roofline.py \
    --outfile ~/local/tmp/20240813_del_del_dyn_bench_roofline.csv \
    --gemm_benchmarks_file ~/local/tmp/20240813_gemm_gpu_time_sweep_9_16.csv \
    --gemm_time_strategy benchmarks --model_torch_compile_limitations True \
    --scaling_type_input delayed --scaling_type_weight delayed \
    --scaling_type_grad_output dynamic | with-proxy gh gist create
- Creating gist...
✓ Created secret gist
https://gist.github.com/vkuzo/e3c4b274493140a7423c85d27863663a
```

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/benchmarks/float8/float8_roofline.py
+++ b/benchmarks/float8/float8_roofline.py
@@ -1,0 +1,391 @@
+"""
+This is a script to estimate the benefit from converting a `torch.nn.Linear`
+layer to float8, by estimating the difference in e2e GPU kernel time between:
+1. bf16 gemms in fwd and bwd, and 
+2. float8 gemms in fwd and bwd, and float8 overhead
+
+The gemm times are estimated either from direct measurements via benchmarks,
+or with a roofline estimation based on TOPS and peak compute bandwidth of an 
+NVIDIA H100.
+
+The float8 overhead times are estimated by counting memory reads and writes
+based on the specified float8 scaling, and estimating that we can achieve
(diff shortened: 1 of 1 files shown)


drisspg (approved):
Noob question does sympy really speed up this script?

vkuzo:
> Noob question does sympy really speed up this script?

Not sure about speedup.  The reason I used sympy is to easily see the contribution of arbitrary M, K, N to memory overhead.

PR #671, opened 2024-08-14
[Experimental] Float8 support in AQT

jainapurva:
Add float8 inference support to current Affine Quantized Tensor.

Test Plan : test/dtypes/test_affine_quantized.py


torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -246,6 +246,41 @@ def from_float_static(
             dtype=input_float.dtype,
         )
 
+    @classmethod
+    def from_float_float(
```
vkuzo, line 250: I think at this point `from_float`, `from_float_static`, `from_float_float` could use docblocks to explain how they differ from each other

vkuzo:
I think a good next step would be to add numerical tests, and ensure that this new object matches the numerical behavior of `Float8Tensor`.

torchao/dtypes/affine_quantized_tensor.py
jainapurva: I'm still trying to decide on how do we call them? I don't want to add another call like https://github.com/pytorch/ao/blob/174e630af2be8cd18bc47c5e530765a82e97f45b/torchao/dtypes/affine_quantized_tensor.py#L921. I'm thinking of making a dict for calls based on target_dtype

vkuzo: my 5c - writing docblocks for what is there now will help with deciding with naming, future refactors, etc.

jainapurva: Ok, will do that

jerryzh168: also we may not need this if we can just reuse the `from_float` code path, this is a prototype to see how much overlap there is as well I think

jainapurva: Yes, I don't want to update the from_float function before testing.

HDCharles:
can i ask whether we're sure we should include float8 tensors in AQT instead of another paradigm?

jainapurva:
> can i ask whether we're sure we should include float8 tensors in AQT instead of another paradigm?

AQT conceptually aligns a lot with fp8/fpx. Instead of writing a separate tensor subclass, it's more efficient to add float support to AQT. The concept of AQT is shared, the major difference is for dtype. This is an experimental PR, to test the feasibility. The design will be modified.

HDCharles:
(sorry didn't mean to close this, missclick)

torchao/quantization/quant_primitives.py
```diff
@@ -8,15 +8,24 @@
...
 )
 from torchao.utils import _register_custom_op
+from torchao.float8.float8_tensor import (
+    LinearMMConfig,
+    GemmInputRole,
+    Float8Tensor,
+)
 
+from torchao.float8.float8_scaling_utils import (
+    hp_tensor_to_float8_dynamic,
```
vkuzo, line 25: this is not a public API of `torchao.float8`, would be great to not use it here.  This function is an implementation detail and can change at any time, without any bc guarantees.

torchao/quantization/quant_primitives.py
```diff
@@ -692,6 +714,56 @@ def _choose_qparams_affine(
...
+def quantize_affine_float8(
+    input: torch.Tensor,
+    block_size: Tuple[int, ...],
+    output_dtype: torch.dtype,
+    quant_min: Optional[Union[int, float]] = None,
+    quant_max: Optional[Union[int, float]] = None,
+    zero_point_domain: ZeroPointDomain = ZeroPointDomain.INT,
+    reduce_amax: bool = False,
+) -> Float8Tensor:
+    return hp_tensor_to_float8_dynamic(
```
vkuzo, line 727: I thought adding float8 to AQT would mean the casts in AQT would be used, it's a bit confusing to use AQT but also use external utilies for casting.

torchao/quantization/quant_primitives.py
```diff
@@ -8,15 +8,24 @@
...
 
-
 from torchao.kernel.intmm import int_scaled_matmul
 from torchao.kernel.intmm import safe_int_mm
 from torchao.utils import (
     TORCH_VERSION_AT_LEAST_2_3,
     TORCH_VERSION_AT_LEAST_2_5,
 )
 from torchao.utils import _register_custom_op
+from torchao.float8.float8_tensor import (
```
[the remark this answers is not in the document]
jainapurva: Removed those. I was trying to experiment with it, but now it's not needed

torchao/quantization/quant_primitives.py
jainapurva: The final goal is for the AQT cast to be used. I was facing some errors while trying to use AQT's cast, was trying to resolve those.

torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -663,6 +700,99 @@ def get_plain(self) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
...
+        shape = int_data.shape
+        return torch.Tensor._make_wrapper_subclass(cls, shape, **kwargs)  # type: ignore[attr-defined]
+
+    def __init__(
+        self,
+        int_data: torch.Tensor,
+        scale: torch.Tensor,
+        layout_type: LayoutType,
+    ):
+        self.int_data = int_data
```
vkuzo, line 729: should this be renamed to `int_or_float_data`?  Ideally in a separate preceding PR from this one.

torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -269,6 +270,42 @@ def from_float_static(
             dtype=input_float.dtype,
         )
 
+    @classmethod
+    def from_float_float8(
```
vkuzo, line 274: @jerryzh168  can you help with a design on how to have `from_float`, `from_float_static`, etc extend to this use case?  Ideally we shouldn't special case a set of dtypes (float8) to have their own function.

test/dtypes/test_affine_quantized_float.py
```diff
@@ -0,0 +1,52 @@
...
+import unittest
+import tempfile
+from torchao.utils import (
+    TORCH_VERSION_AFTER_2_5,
+)
+
+
+class TestAffineQuantizedFloat(TestCase):
+    @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
+    def test_tensor_core_layout_transpose(self):
```
jerryzh168, line 18: we probably don't need to test this one I feel

test/dtypes/test_affine_quantized_float.py
```diff
@@ -0,0 +1,52 @@
...
+        for _ in range(10):
+            t = t.t()
+            aqt = aqt.t()
+            shape = t.shape
+            aqt_shape = aqt.shape
+            self.assertEqual(aqt_shape, shape)
+
+    @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
+    def test_weights_only(self):
+        for apply_quant in [float8_weight_only()]:
```
jerryzh168, line 38: this can be merged with test_affine_quantized.py file?

torchao/dtypes/affine_quantized_tensor.py
jainapurva: A combined function is better, will be refactoring it after testing float8

test/dtypes/test_affine_quantized_float.py
jainapurva: Yes, this will be merged

torchao/dtypes/affine_quantized_tensor.py
jerryzh168: yeah sure, I think we could have the two following final state:
* have separate `from_float_fpx` and `from_from_intx` since they have a bit different arg list
*  if we manage to generalize the arg list enough so it is reasonable merge the two then we can merge as well, I will discuss with Apurva and Driss about the args but at the first glance maybe preserve_zero is always going to be true and zero_point_domain may not apply here

torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -663,6 +700,99 @@ def get_plain(self) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
     def get_layout_type(self) -> LayoutType:
         return self.layout_type
 
+
+@register_layout_cls(FpxLayoutType)
+class FpxAQTLayout(AQTLayout):
```
jerryzh168, line 705: this is added in the fpx PR..do you want to wait until that is landed?

torchao/dtypes/affine_quantized_tensor.py
vkuzo: makes sense.

one thought, the `from_float` name will become more confusing if both the source and the target can also be various floating point bitwidths.  To clarify this in `torchao.float8`, I went with the `high_precision|hp` and `low_precision|lp` naming scheme

cpuhrsch: This is also why I like the idea of extending `to` and having our own factory functions that we can pass dtype enums.

jerryzh168: @vkuzo yeah makes sense, we can rename `from_float` to `from_high_precision` as well. as @cpuhrsch mentioned, this is not user facing API, we'll have `to` factory functions for various dtypes as the user facing API: https://github.com/pytorch/ao/blob/5c0e060a1a204d51ce6bc09901bc6f0574d2e38c/torchao/dtypes/affine_quantized_tensor.py#L990-L991

jerryzh168:
@jainapurva can you add this to https://github.com/pytorch/ao/blob/main/torchao/_models/llama/generate.py and https://github.com/pytorch/ao/blob/main/torchao/_models/llama/eval.py as well and get some e2e numbers for them

torchao/dtypes/affine_quantized_tensor.py
jainapurva: As discussed, I'm not using this layout anymore. I'll remove this code.

torchao/dtypes/affine_quantized_tensor.py
jerryzh168: @jainapurva as discussed from the meeting, let's merge this into `from_float` and add some guards on arguments

torchao/quantization/quant_api.py
```diff
@@ -488,6 +488,24 @@ def int8_dynamic_activation_int8_semi_sparse_weight():
...
+    def apply_float8wo_quant(weight):
+        # avoid circular dep
+        from torchao.dtypes import to_affine_quantized_float8
+
+        mapping_type = MappingType.SYMMETRIC
+        target_dtype = torch.float8_e4m3fn
+        eps = torch.finfo(torch.float32).eps
+        zero_point_dtype = torch.float32
+        block_size = (1, weight.shape[1])
+        return to_affine_quantized(input_float=weight, mapping_type=mapping_type, block_size=block_size, target_dtype=target_dtype,
```
drisspg, line 504: Should this be: to_affine_quantized_float8?

jerryzh168: we want to merge float8 into `to_affine_quantized` I think

torchao/quantization/quant_primitives.py
```diff
@@ -95,9 +101,12 @@ def _get_and_check_qmin_qmax(dtype, quant_min, quant_max):
     verify that they are within the range of possible quant_min/quant_max
     for dtype
     """
-    if dtype not in _DTYPE_TO_QVALUE_BOUNDS:
+    if dtype in FP8_TYPES:
+        quant_min_lower_bound, quant_max_upper_bound = torch.finfo(dtype).min, torch.finfo(dtype).max
```
jerryzh168, line 105: we don't even need these if we don't want fp8 to have quant_min/quant_max I think

torchao/quantization/quant_primitives.py
```diff
@@ -171,7 +180,7 @@ def quantize_affine(
       output_dtype (torch.dtype): requested dtype (e.g. torch.uint8) for output Tensor
       quant_min (Optional[int]): minimum quantized value for output Tensor, if not specified, it will be derived from dtype
       quant_max (Optional[int]): maximum quantized value for output Tensor, if not specified, it will be derived from dtype
-      zero_point_domain (ZeroPointDomain): the domain that zero_point is in, should be either integer or float
+      zero_point_domain (ZeroPointDomain): the domain that zero_point is in, should be eitehr integer or float
```
jerryzh168, line 183: please revert these changes, there is a PR that fixes these typos and seems it's reverted here

test/dtypes/test_affine_quantized.py
```diff
@@ -8,16 +8,21 @@
...
 )
-from torchao.utils import TORCH_VERSION_AT_LEAST_2_5
 
 import torch
 import unittest
 import tempfile
 
+from torchao.utils import (
+    TORCH_VERSION_AT_LEAST_2_5,
+    TORCH_VERSION_AFTER_2_5,
```
jerryzh168, line 23: nit: this can be removed, we are using `TORCH_VERSION_AT_LEAST_2_5` now

test/dtypes/test_affine_quantized.py
```diff
@@ -70,5 +75,21 @@ def test_to_device(self):
 
 
 
+class TestAffineQuantizedFloat(TestCase):
+    @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
+    def test_weights_only(self):
+        for apply_quant in [float8_weight_only()]:
```
jerryzh168, line 81: you can add `float8_weight_only()` to this list directly I think: https://github.com/pytorch/ao/blob/eb47c93ca2ae41f85a0d5d73441ad1bec7cb2405/test/dtypes/test_affine_quantized.py#L58

torchao/quantization/quant_api.py
jainapurva: Merged from_float_float8 into from_float

jainapurva:
> @jainapurva can you add this to https://github.com/pytorch/ao/blob/main/torchao/_models/llama/generate.py and https://github.com/pytorch/ao/blob/main/torchao/_models/llama/eval.py as well and get some e2e accuracy and perf numbers for them

Testing perf in this PR: https://github.com/pytorch/ao/pull/732

torchao/quantization/quant_primitives.py
jainapurva: We need it here
https://github.com/pytorch/ao/blob/efd480b43c14d0987e68834b501536d9f3330c68/torchao/quantization/quant_primitives.py#L270-L283

We don't want input for it. We can use default values.

torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -36,6 +37,22 @@
 
 aten = torch.ops.aten
 
+def validate_float8_params(
```
jerryzh168, line 40: please add type annotations for this as well

torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -36,6 +37,22 @@
...
+    assert input_float.is_floating_point(), "input_float must be a floating point tensor"
+    assert mapping_type in [MappingType.SYMMETRIC], "Only symmetric mapping is supported for float8"
+    assert target_dtype in FP8_TYPES, "target_dtype must be one of the follwoing: {}".format(FP8_TYPES)
+    assert quant_min is None, "quant_min must be None for float8"
+    assert quant_max is None, "quant_max must be None for float8"
+    assert eps is None, "eps must be None for float8"
+    assert scale_dtype is None, "scale_dtype must be None for float8"
+    assert zero_point_dtype is None, "zero_point_dtype must be None for float8"
+    assert preserve_zero is True, "preserve_zero must be True for float8"
+    assert zero_point_domain == ZeroPointDomain.INT, "zero_point_domain must be ZeroPointDomain.INT for float8"
```
jerryzh168, line 52: ideally these can be None as well, although it seems that integer quant requires these to be valid

torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -36,6 +37,22 @@
 
 aten = torch.ops.aten
 
+def validate_float8_params(
+    input_float, mapping_type, target_dtype, quant_min, quant_max, eps, scale_dtype, zero_point_dtype, preserve_zero, zero_point_domain, layout_type, use_hqq
+):
+    assert input_float.is_floating_point(), "input_float must be a floating point tensor"
```
jerryzh168, line 43: actually based on these it seems most of the args are irrelevant for float8, maybe just splitting the op makes more sense?

torchao/quantization/quant_api.py
```diff
@@ -488,6 +489,17 @@ def int8_dynamic_activation_int8_semi_sparse_weight():
     """
     return int8_dynamic_activation_int8_weight(layout_type=SemiSparseLayoutType())
 
+def float8_weight_only(target_dtype: torch.dtype = torch.float8_e4m3fn):
+    """
+    Applies float8 weight-only symmetric per-channel quantization to linear layers.
+    """
+    def apply_float8wo_quant(weight):
+        # avoid circular dep
```
jerryzh168, line 497: maybe you want to import `to_affine_quantized_floatx` here, I'm also refactoring this file to change the import to the file to avoid circular dep as well

torchao/dtypes/affine_quantized_tensor.py
jainapurva: Yes, created a new op from_float_to_floatx. It'll be calling the existing from_float but with some pre-defined param values. Also for future we'll need to rename and refactor these methods.

jerryzh168 (approved):

PR #672, opened 2024-08-14
Use `torch.uint1` to `torch.uint7` for Uintx tensor subclass

jerryzh168:
Summary:
Previously we are using bit_width for uintx quantization, but we can actually use `dtype` directly.

But there are still some workaround to convert from torch dtype to bit_width right now, if we want to remove all the hacks, we'd need to support Uintx tensor subclass properly and have `torch.uintx` dispatch to the tensor subclass. this is probably not the highest priority for now since good perf is more important.

Test Plan:
pytest test/dtypes/test_uintx.py

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/dtypes/test_affine_quantized.py
+++ b/test/dtypes/test_affine_quantized.py
@@ -9,13 +9,14 @@ from torchao.quantization.quant_api import (
     int8_dynamic_activation_int8_weight,
     int8_dynamic_activation_int8_semi_sparse_weight,
 )
+from torchao.dtypes import (
+    to_affine_quantized,
+)
+from torchao.utils import TORCH_VERSION_AT_LEAST_2_5
+
 import torch
 import unittest
 import tempfile
-from torchao.utils import (
-    TORCH_VERSION_AT_LEAST_2_5,
-)
-
 
 class TestAffineQuantized(TestCase):
     @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
--- a/test/dtypes/test_uintx.py
+++ b/test/dtypes/test_uintx.py
@@ -6,7 +6,10 @@ import torch
 
 from torchao.dtypes.uintx.Uintx import to_uintx
 from torchao.quantization.quant_api import quantize_, uintx_weight_only
-from torchao.utils import TORCH_VERSION_AT_LEAST_2_5
+from torchao.utils import (
+    TORCH_VERSION_AT_LEAST_2_3,
+    TORCH_VERSION_AT_LEAST_2_5,
+)
 
 from torchao.quantization.quant_primitives import (
     MappingType,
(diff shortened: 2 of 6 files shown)


torchao/dtypes/uintx/Uintx.py
```diff
@@ -11,10 +11,30 @@
...
+from torchao.utils import TORCH_VERSION_AFTER_2_3
 
 aten = torch.ops.aten
 
+# Note: Uintx does not work for torch 2.3 and below
+_DTYPE_TO_BIT_WIDTH = {}
+_BIT_WIDTH_TO_DTYPE = {}
+
+if TORCH_VERSION_AFTER_2_3:
+    _DTYPE_TO_BIT_WIDTH = {
```
vayuda, line 23: So this is the hacky fix until it is more natively supported

test/dtypes/test_affine_quantized.py
```diff
@@ -51,6 +55,44 @@ def test_weights_only(self):
...
+            l = torch.nn.Linear(128, 256, dtype=torch.bfloat16, device="cuda")
+            # make sure it runs
+            uintx_weight_only(dtype)(l)
+            l = torch.compile(l)
+            l(torch.randn(1, 128, dtype=torch.bfloat16, device="cuda"))
+
+
+    @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
+    @unittest.skipIf(not TORCH_VERSION_AFTER_2_3, "sub byte dtype requires torch 2.3+")
+    def test_uintx_model_size(self):
```
vayuda, line 72: This test seems like it belongs in `test_uintx.py` as a better assert for `test_uintx_weight_only_model_quant`.

torchao/dtypes/uintx/Uintx.py
jerryzh168: right, `torch.uintx` is a placeholder right now and nothing is supported on it, we'd need to add a proper tensor subclass and connect it with these dtypes in the future

test/dtypes/test_affine_quantized.py
jerryzh168: OK, I can move these

test/dtypes/test_uintx.py
```diff
@@ -6,7 +6,10 @@
 
 from torchao.dtypes.uintx.Uintx import to_uintx
 from torchao.quantization.quant_api import quantize_, uintx_weight_only
-from torchao.utils import TORCH_VERSION_AFTER_2_5
+from torchao.utils import (
+    TORCH_VERSION_AFTER_2_3,
```
vayuda, line 10: This import makes sense, but also is kind of useless since tests don't run unless Torch version is nightly. Maybe we can change tests to running if the version is after 2.3, and only calling `torch.compile()` if the torch version is nightly.

jerryzh168: sorry I don't quite follow, this is to guard against reference to `torch.uintx`, which is added in torch 2.3, so the tests will only run if torch is 2.3+

are you saying to add the version guard for tests as well?

also I didn't see a call for `torch.compile` in the test, are you saying we should add that

vayuda: Oh interesting, I didn't catch that we weren't using torch compile. That is intended usage so it should be tested.

 I was trying to say the guard for version 2.5 was for torch compile since the nested sub-classes issue was fixed only recently. Currently tests only run if version after 2.5 so the version after 2.3 guard is not really doing anything. 
 
 Now I see it guards the import which is used in a statement before the tests are executed so it is still needed

drisspg (approved):

PR #680, opened 2024-08-14
Add BSR subclass +torch.compile and clean up superblock

jcaip:
This PR adds in torch.compile support for block sparsity. 

In a custom op, we create the `sprase_bsr_tensor` from the explicit `crow_indices, col_indices, values`  tensors that are passed in to the custom op. 

I also created a tensor subclass which holds these same values. 

At dispatch, when we see a `torch.nn.functional.linear` call, we dispatch into our custom op  `torch.ops.blocksparse.linear`, using the tensors stored in the subclass. 

This will allow us to add a public API similar to `semi_sparse_weight()`, which I plan to do in a future PR. 

This PR also cleans up the superblock prototype implementation, as there was a lot of repeated code, and also adds in kernel tuning for BSR. 

For bfloat16 I see the following numbers, for a 1.23x gain:
```
New compile baseline: 63.431 ms
New compile + bsr: 53.514 ms 
New compile + bsr + tuning: 51.485 ms 
```


torchao/sparsity/prototype/superblock/utils.py
```diff
@@ -12,6 +12,25 @@
 import torch
 import torch.distributed as dist
 
+### IMAGENET UTILS
+@torch.inference_mode
+def benchmark_inference(warmup, iters, f, *args, **kwargs):
```
jerryzh168, line 17: why not use https://github.com/pytorch/ao/blob/5998389cb7304ed40f09251ed75e50632d00c142/torchao/utils.py#L51? we can add a warmup as well

jerryzh168 (approved):

torchao/sparsity/prototype/superblock/utils.py
jcaip: It was giving me different timings from my function. I'll try and debug the delta later, if I can find out whats causes it I'll switch over.

PR #679, opened 2024-08-14
retry version guard fix

msaroufim:

PR #684, opened 2024-08-15
Fix source version check

msaroufim:
@crcrpar can you try this instead?

Fixes https://github.com/pytorch/ao/pull/679#issuecomment-2290444713

This just basically takes advantage of the fact that every PyTorch version has the format `X.Y.Z` and then some stuff

```python

# Test cases
import torch
from torchao.utils import torch_version_at_least

torch.__version__ = "2.5.0a0+git9f17037"
print(f"{torch.__version__}: {torch_version_at_least('2.5.0')}")  # Should be True
print(f"{torch.__version__}: {torch_version_at_least('2.4.0')}")  # Should be True


torch.__version__ = "2.5.0.dev20240708+cu121"
print(f"{torch.__version__}: {torch_version_at_least('2.5.0')}")  # Should be True
print(f"{torch.__version__}: {torch_version_at_least('2.4.0')}")  # Should be True


torch.__version__ = "2.5.0"
print(f"{torch.__version__}: {torch_version_at_least('2.4.0')}")  # Should be True
print(f"{torch.__version__}: {torch_version_at_least('2.5.0')}")  # Should be True

torch.__version__ = "2.4.0"
print(f"{torch.__version__}: {torch_version_at_least('2.4.0')}")  # Should be True
print(f"{torch.__version__}: {torch_version_at_least('2.5.0')}")  # Should be False

```

### DIFF
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -1,0 +1,26 @@
+import unittest
+from unittest.mock import patch
+from torchao.utils import torch_version_at_least
+
+class TestTorchVersionAtLeast(unittest.TestCase):
+    def test_torch_version_at_least(self):
+        test_cases = [
+            ("2.5.0a0+git9f17037", "2.5.0", True),
+            ("2.5.0a0+git9f17037", "2.4.0", True),
+            ("2.5.0.dev20240708+cu121", "2.5.0", True),
+            ("2.5.0.dev20240708+cu121", "2.4.0", True),
+            ("2.5.0", "2.4.0", True),
+            ("2.5.0", "2.5.0", True),
+            ("2.4.0", "2.4.0", True),
+            ("2.4.0", "2.5.0", False),
+        ]
+
+        for torch_version, compare_version, expected_result in test_cases:
+            with patch('torch.__version__', torch_version):
+                result = torch_version_at_least(compare_version)
+
+                self.assertEqual(result, expected_result, f"Failed for torch.__version__={torch_version}, comparing with {compare_version}")
+                print(f"{torch_version}: {result}")
+
+if __name__ == '__main__':
+    unittest.main()
--- a/torchao/utils.py
+++ b/torchao/utils.py
@@ -7,6 +7,8 @@ import torch.nn.utils.parametrize as parametrize
 import itertools
 import time
 import warnings
+import re
+
 
 __all__ = [
     "benchmark_model",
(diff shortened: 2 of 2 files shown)


jerryzh168:
> @crcrpar can you try this instead?
> 
> Fixes [#679 (comment)](https://github.com/pytorch/ao/pull/679#issuecomment-2290444713)
> 
> This just basically takes advantage of the fact that every PyTorch version has the format `X.Y.Z` and then some stuff
> 
> ```python
> # Test cases
> import torch
> from torchao.utils import torch_version_at_least
> 
> torch.__version__ = "2.5.0a0+git9f17037"
> print(f"{torch.__version__}: {torch_version_at_least('2.5.0')}")  # Should be True
> print(f"{torch.__version__}: {torch_version_at_least('2.4.0')}")  # Should be True
> 
> 
> torch.__version__ = "2.5.0.dev20240708+cu121"
> print(f"{torch.__version__}: {torch_version_at_least('2.5.0')}")  # Should be True
> print(f"{torch.__version__}: {torch_version_at_least('2.4.0')}")  # Should be True
> 
> 
> torch.__version__ = "2.5.0"
> print(f"{torch.__version__}: {torch_version_at_least('2.4.0')}")  # Should be True
> print(f"{torch.__version__}: {torch_version_at_least('2.5.0')}")  # Should be True
> 
> torch.__version__ = "2.4.0"
> print(f"{torch.__version__}: {torch_version_at_least('2.4.0')}")  # Should be True
> print(f"{torch.__version__}: {torch_version_at_least('2.5.0')}")  # Should be False
> ```

maybe add these examples to tests?

jerryzh168 (approved):

test/test_utils.py
```diff
@@ -0,0 +1,26 @@
+import unittest
+from unittest.mock import patch
+from torchao.utils import torch_version_at_least
+
+class TestTorchVersionAtLeast(unittest.TestCase):
+    def test_torch_version_at_least(self):
+        test_cases = [
+            ("2.5.0a0+git9f17037", "2.5.0", True),
```
crcrpar, line 8: I'm confused, as I thought this should be False instead.
shouldn't 2.5.0a0 and 2.5.0.dev be <2.5.0 but >2.4?
I might misunderstand `torch_version_at_least` usage though

msaroufim: Strictly speaking this is an internal only utility that we leverage in case torch makes certain breaking changes or adds new features upstream. So I can't think of many cases where we'd do one thing with 2.5 nightlies vs 2.5 official releases.

crcrpar: understood.

felipemello1:
@msaroufim , when do you guys plan to make a new release? Is it monthly?

msaroufim:
We do monthly, next one is scheduled around Sep 5

PR #686, opened 2024-08-15
Scale to fp32 to make sure scaled_mm get float scalar scale input

qingquansong:
Scale to fp32 to make sure scaled_mm get float scalar scale input. This is helpful if our tensor is loaded with bf16 param type. Otherwise scale will be bf16 causing failure.

Need double check `preprocess_addmm` should extract the fp32 scale ideally


qingquansong:
Closed since it's related to pytorch lightning context manager casting buffer variable to bf16

PR #689, opened 2024-08-15
Change dispatch key for `_register_custom_op`

jerryzh168:
Summary:
`_register_custom_op` is a util for the ops to be preserved in torch.export for executorch, but after testing, Max found that the ops are decomposed during to_edge, and it's because of the dispatch key we are using, so we changed it to `CompositeExplicitAutograd` following https://github.com/pytorch/pytorch/blob/86aa327e4afba5b883341ed14d1c9a80e2441aaa/torch/ao/quantization/fx/_decomposed.py#L50

Test Plan:

confirmed `python tutorials/quantize_vit/run_vit_b_quant.py` also does not regress:
before:
elapsed_time:  9.0510380859375  milliseconds
after:
elapsed_time:  8.6919443359375  milliseconds

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/utils.py
+++ b/torchao/utils.py
@@ -190,7 +190,7 @@ def _register_custom_op(lib):
             op_name = fn.__name__[1:]
             schema = op_name + infer_schema(fn, mutates_args={})
             lib.define(schema)
-            lib.impl(op_name, fn, "CompositeImplicitAutograd")
+            lib.impl(op_name, fn, "CompositeExplicitAutograd")
 
             lib_namespace = lib.ns
             op = getattr(getattr(torch.ops, lib_namespace), op_name)


mcr229:
I think we just let CI run its course and see how it fairs.

zou3519 (commented):
You're not going to get autograd support for these operators in torch.compile if you mark them as CompositeExplicitAutograd. If you're OK with that, then we should block autograd support for eager-mode (that gets erroneously added) so that there are no eager vs compile differences

jerryzh168:
we still want this to be decomposed in torch.compile. Max is adding support in edge to preserve these ops instead, so this PR is not needed anymore

PR #495, opened 2024-07-10
[FEAT] Performance Profiler

jeromeku:
@msaroufim 

## Add Performance Profiler

Initial implementation of a performance profiler per [#426](https://github.com/pytorch/ao/issues/426).

## Overview

Primary contribution is a `TransformerPerformanceCounter` which records data movement and FLOPs across multiple contexts.  

Combined with a `DeviceSpec`, theoretical peak performance / utilization stats are accumulated for each individual context and in aggregate which can be used for `Speed of Light` / roofline analysis and other downstream performance profiling.

Motivation is to create a lightweight method for collecting useful performance stats using (mostly) `torch`-native features as a complement to `torch.profiler` and before diving into tools such as `nsight compute`.

## Details
Below is an example demonstrating the basic API.

```python
from torchao.profiler import CUDADeviceSpec, TransformerPerformanceCounter

# Device spec is a dataclass that contains device info such as name, peak bandwidth, peak FLOPs
# If these fields are not manually specified, they will be automatically populated using
# CUDA runtime functions exposed by `torch.cuda` and `triton.runtime.driver`  
device_spec = CUDADeviceSpec()

# The manager object tracks latency, data movement (bytes), and FLOPs across multiple contexts and
# maintains performance metrics for each context and in aggregate.
manager = TransformerPerformanceCounter(device_spec=device_spec)

# Prefill
with manager.count(label="prefill", num_tokens=x.numel()):
    out = model(encoded_prompt)
# Print recorded stats for "prefill" context
manager.print_summary(labels=["prefill"], show=True) 

# Decode
with manager.count(label="decode", num_tokens=1):
    out = model(out[-1])
# Print recorded stats for "decode" context
manager.print_summary(labels=["decode"], show=True) 

# Print accumulated stats across all contexts
manager.print_summary(show=True) # 
```
- `CUDADeviceSpec` is a lightweight dataclass that models device info
  - auto-populated with device-specific information using `torch.cuda` and `triton.runtime.driver` CUDA APIs where possible 
  - a hard-coded table is maintained in `torchao.profiler.device_spec._AVAILABLE_GPU_SPECS` to fill in peak FLOPs, though this should be possible to calculate directly using the `cudaDriver` API.
- `TransformerPerformanceCounter` uses `PerformanceCounterMode` under the hood to capture data movement and FLOPs
[...]

PR #690, opened 2024-08-16
[FEAT] Perf Profiler Update 

jeromeku:
@msaroufim 

Same as #495 except rebased on main.


jerryzh168:
@jeromeku really happy to see you updating this again, we were just discussing about this today. I think this is very important work and hope to merge this soon, I think one major thing before merging here is just to integrate the `PERF_COUNTER` etc. in torchao/_models/llama/perf_profile.py directly with https://github.com/pytorch/ao/blob/main/torchao/_models/llama/generate.py so we don't have duplications

cc @msaroufim @HDCharles

msaroufim:
Let's do code unification in a later PR

msaroufim (approved):
Thanks!!

PR #692, opened 2024-08-16
Relax QAT dtype assertion

andrewor14:
This was added originally for perf reasons specific to 8da4w, but the autograd.Function has since been adapted for more general use. A few users are hitting this assertion error.

More context: https://github.com/pytorch/torchtune/pull/1333


msaroufim (approved):
nice

PR #693, opened 2024-08-16
[do not land] profiler not working with torch.compile

jerryzh168:
Summary:
Current profiler only supports eager mdoe, putting up the PR so people can repro the issue Right now the best way to do perf analysis is still chrome trace I think

on compile mode:
Before turning on profiler
Average tokens/sec: 107.64
After:
Average tokens/sec: 5.39

Test Plan:
cd torchao/_models/llama
python generate.py --compile

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/_models/llama/generate.py
+++ b/torchao/_models/llama/generate.py
@@ -5,9 +5,10 @@
 # LICENSE file in the root directory of this source tree.
 import os
 import sys
+import textwrap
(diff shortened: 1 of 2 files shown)


jerryzh168:
cc @jeromeku wondering if you have any ideas on this

PR #696, opened 2024-08-17
Refactor quant_llm to work with affine quantized tensor

jerryzh168:
Summary:
We want to add quant_llm to affine quantized tensor as a general fp2-fp7 dtype, before that we need to refactor the current implementation to work with AffineQuantizedTensor first

Test Plan:
```
python test/prototype/test_fpx.py

# perf
# in torchao/_models/llama
python generate.py --compile --quantization fp6

# accuracy
python eval.py --quantization fp6 --compile

```

tok/s is matching the original now, fp6 is:
```
==========
Average tokens/sec: 115.62
Average Bandwidth: 573.39 GB/s
Peak Memory Usage: 6.97 GB
Model Size: 4.96 GB
```
perplexity with eval in _models/llama/ is:
```
wikitext: {'word_perplexity,none': 12.360869670148428, 'word_perplexity_stderr,none': 'N/A', 'byte_perplexity,none': 1.6003634603882877, 'byte_perplexity_stderr,none': 'N/A', 'bits_per_byte,none': 0.6783995944569804, 'bits_per_byte_stderr,none': 'N/A', 'alias': 'wikitext'}
```
matching previous result: 
FP6 E3M2 | 12.34 | 106.76

new numbers for benchmark_fp6.py
```
|   m |     k |     n |   fp6_latency (ms) |   fp16_latency (ms) |   speedup (d/s) |   correct |
|----:|------:|------:|-------------------:|--------------------:|----------------:|----------:|
|   1 |  8192 |  8192 |            66.7578 |             93.8563 |        1.40592  |         1 |
|   1 | 10240 |  8192 |            74.4154 |            111.698  |        1.501    |         1 |
|   1 | 57344 |  8192 |           236.078  |            577.7    |        2.44707  |         1 |
|   1 |  8192 | 28672 |           132.349  |            296.613  |        2.24114  |         1 |
|   2 |  8192 |  8192 |            63.3943 |             94.0825 |        1.48408  |         1 |
|   2 | 10240 |  8192 |            62.4372 |            114.443  |        1.83294  |         1 |
|   2 | 57344 |  8192 |           237.778  |            566.592  |        2.38286  |         1 |
|   2 |  8192 | 28672 |           133.844  |            304.042  |        2.27162  |         1 |
|   4 |  8192 |  8192 |            60.7482 |             94.2242 |        1.55106  |         1 |
|   4 | 10240 |  8192 |            61.2663 |            115.321  |        1.88229  |         1 |
|   4 | 57344 |  8192 |           238.502  |            567.895  |        2.38109  |         1 |
|   4 |  8192 | 28672 |           135.851  |            305.304  |        2.24735  |         1 |
|   8 |  8192 |  8192 |            54.8235 |             95.2624 |        1.73762  |         1 |
|   8 | 10240 |  8192 |            61.5713 |            116.808  |        1.89711  |         1 |
|   8 | 57344 |  8192 |           241.506  |            569.578  |        2.35845  |         1 |
|   8 |  8192 | 28672 |           139.488  |            307.304  |        2.20309  |         1 |
|  16 |  8192 |  8192 |            61.2497 |             96.2626 |        1.57164  |         1 |
|  16 | 10240 |  8192 |            60.8278 |            114.659  |        1.88497  |         1 |
|  16 | 57344 |  8192 |           264.154  |            572.281  |        2.16647  |         1 |
|  16 |  8192 | 28672 |           145.249  |            310.304  |        2.13636  |         1 |
|  32 |  8192 |  8192 |            60.7182 |             96.0235 |        1.58146  |         1 |
|  32 | 10240 |  8192 |            67.6144 |            114.384  |        1.69172  |         1 |
|  32 | 57344 |  8192 |           289.831  |            586.451  |        2.02342  |         1 |
|  32 |  8192 | 28672 |           175.66   |            326.459  |        1.85847  |         1 |
|  64 |  8192 |  8192 |            76.7881 |             97.5357 |        1.27019  |         1 |
|  64 | 10240 |  8192 |            91.239  |            116.467  |        1.2765   |         1 |
|  64 | 57344 |  8192 |           416.322  |            602.291  |        1.44669  |         1 |
|  64 |  8192 | 28672 |           257.579  |            335.085  |        1.3009   |         1 |
| 128 |  8192 |  8192 |           129.629  |            120.611  |        0.930432 |         1 |
| 128 | 10240 |  8192 |           156.548  |            136.184  |        0.86992  |         1 |
| 128 | 57344 |  8192 |           782.291  |            690.499  |        0.882663 |         1 |
| 128 |  8192 | 28672 |           478.238  |            425.829  |        0.890412 |         1 |
| 256 |  8192 |  8192 |           251.177  |            190.626  |        0.758932 |         1 |
| 256 | 10240 |  8192 |           300.058  |            225.371  |        0.751091 |         1 |
| 256 | 57344 |  8192 |          1433.14   |           1192.71   |        0.832238 |         1 |
| 256 |  8192 | 28672 |           876.324  |            648.411  |        0.739922 |         1 |
| 512 |  8192 |  8192 |           493.372  |            367.824  |        0.745531 |         1 |
| 512 | 10240 |  8192 |           592.824  |            448.459  |        0.756479 |         1 |
| 512 | 57344 |  8192 |          2808.56   |           2191.73   |        0.780376 |         1 |
| 512 |  8192 | 28672 |          1617.37   |           1163.74   |        0.719525 |         1 |
```
Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/dtypes/test_fpx.py
+++ b/test/dtypes/test_fpx.py
@@ -8,23 +8,27 @@ from torch.testing._internal.common_utils import (
     parametrize,
     run_tests,
 )
-from torchao.prototype.quant_llm import (
-    QuantLlmLinearWeight,
-    quant_llm_fpx_weight_only,
-    fp6_llm_weight_only,
+from torchao.dtypes.fpx import (
+    FpxTensorCoreAQTLayout,
+    FpxTensorCoreLayoutType,
     to_scaled_tc_fpx,
     from_scaled_tc_fpx,
 )
-from torchao.prototype.quant_llm.quant_llm import _pack_tc_fpx, _pack_tc_fp6
+from torchao.dtypes.fpx.fpx import _pack_tc_fpx, _pack_tc_fp6
 from torchao.prototype.custom_fp_utils import _f32_to_fpx_unpacked, _fpx_unpacked_to_f32
-from torchao.quantization.quant_api import quantize_
+from torchao.quantization import (
+    quantize_,
+    fpx_weight_only,
+)
+
+from torchao.utils import TORCH_VERSION_AT_LEAST_2_5
 
 
 _DEVICES = ["cpu"] + (["cuda"] if torch.cuda.is_available() else [])
 _FPx_DTYPES = [(3, 2), (2, 2)]
 
 
-class TestQuantLlmLinearWeight(TestCase):
+class TestFpxTensorCoreAQTLayout(TestCase):
     @parametrize("device", _DEVICES)
     def test_pack_tc_fp6_correctness(self, device):
         x = torch.randint(256, size=(256, 64), dtype=torch.uint8, device=device)
@@ -69,61 +73,40 @@ class TestQuantLlmLinearWeight(TestCase):
     @pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA not available")
     @parametrize("ebits,mbits", _FPx_DTYPES)
     def test_to_copy_device(self, ebits, mbits):
+        from torchao.quantization.quant_primitives import (
+            choose_qparams_affine_fpx,
+            quantize_affine_fpx,
+        )
+
         x = torch.randn(256, 64)
-        fpx = QuantLlmLinearWeight.from_float(x, ebits, mbits).cuda()
-        assert fpx.device.type == "cuda"
-        fpx = fpx.cpu()
-        assert fpx.device.type == "cpu"
+        scale = choose_qparams_affine_fpx(x, ebits, mbits)
+        x = quantize_affine_fpx(x, scale, ebits, mbits)
+        layout_type = FpxTensorCoreLayoutType(ebits, mbits)
+        fpx_layout_tensor = FpxTensorCoreAQTLayout.from_plain(x, scale, None, layout_type).cuda()
+        assert fpx_layout_tensor.device.type == "cuda"
+        fpx_layout_tensor = fpx_layout_tensor.cpu()
+        assert fpx_layout_tensor.device.type == "cpu"
 
     @pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA not available")
-    @parametrize("ebits,mbits", _FPx_DTYPES)
-    @parametrize("leading_dims", [(4,), (2, 4)])
-    @parametrize("bias", [False, True])
-    def test_quant_llm_linear_weight(self, ebits, mbits, bias, leading_dims):
-        OC, IC = 256, 64
-        device = "cuda"
-
-        fp16_weight = torch.randn(OC, IC, device=device, dtype=torch.half)
-        fp16_bias = torch.randn(OC, device=device, dtype=torch.half) if bias else None
-
-        fpx_weight = QuantLlmLinearWeight.from_float(fp16_weight, ebits, mbits)
-
-        x = torch.randn(*leading_dims, IC, device=device, dtype=torch.half)
-        out = torch.nn.functional.linear(x, fpx_weight, fp16_bias)
-        assert out.shape == leading_dims + (OC,)
-
-    @pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA not available")
+    @pytest.mark.skipif(not TORCH_VERSION_AT_LEAST_2_5, reason="quantization only works with torch.compile for 2.5+")
     @parametrize("ebits,mbits", _FPx_DTYPES)
     @parametrize("bias", [False, True])
-    def test_quant_llm_quantize(self, ebits, mbits, bias):
+    def test_fpx_weight_only(self, ebits, mbits, bias):
         N, OC, IC = 4, 256, 64
         device = "cuda"
 
-        linear = torch.nn.Linear(IC, OC, bias=bias, device=device)
+        linear = torch.nn.Linear(IC, OC, bias=bias, device=device, dtype=torch.half)
         fpx_linear = copy.deepcopy(linear)
-        quantize_(fpx_linear, quant_llm_fpx_weight_only(ebits, mbits))
+        quantize_(fpx_linear, fpx_weight_only(ebits, mbits))
 
         x = torch.randn(N, IC, device=device, dtype=torch.half)
         expected = fpx_linear(x)
         actual = torch.compile(fpx_linear, fullgraph=True)(x)
-        torch.testing.assert_close(actual, expected)
-
-    @pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA not available")
-    def test_fp6_llm_quantize(self):
-        N, OC, IC = 4, 256, 64
-        device = "cuda"
-
-        linear = torch.nn.Linear(IC, OC, device=device)
-        fpx_linear = copy.deepcopy(linear)
-        quantize_(fpx_linear, fp6_llm_weight_only())
-
-        x = torch.randn(N, IC, device=device, dtype=torch.half)
-        expected = fpx_linear(x)
-        actual = torch.compile(fpx_linear, fullgraph=True)(x)
+        # somehow compile now changes the result a bit
         torch.testing.assert_close(actual, expected)
 
 
-instantiate_parametrized_tests(TestQuantLlmLinearWeight)
+instantiate_parametrized_tests(TestFpxTensorCoreAQTLayout)
 
 
 if __name__ == "__main__":
--- a/torchao/dtypes/affine_quantized_tensor.py
+++ b/torchao/dtypes/affine_quantized_tensor.py
@@ -11,9 +11,9 @@ from torchao.quantization.quant_primitives import (
     MappingType,
     int_scaled_matmul,
     quantize_affine_hqq,
-)
-from torchao.quantization.utils import (
-    pack_tinygemm_scales_and_zeros,
+    choose_qparams_affine_fpx,
+    quantize_affine_fpx,
+    dequantize_affine_fpx,
 )
 from torch.utils._python_dispatch import return_and_correct_aliasing
 from torchao.dtypes.utils import (
@@ -33,6 +33,7 @@ from torchao.utils import (
     TorchAOBaseTensor,
     TORCH_VERSION_AT_LEAST_2_5,
 )
+from torchao.ops import quant_llm_linear
 
 aten = torch.ops.aten
 
@@ -45,6 +46,11 @@ class AQTLayout(TorchAOBaseTensor):
     Base class for the layout tensor for `AffineQuantizedTensor`
     """
     def get_plain(self) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+        """Get the plain (unpacked) Tensor for the layout Tensor
+
+        Returns int_data, scale and zero_point
+        Can be overwritten if other types of AQTLayout Tensor has different numbers of plain tensors
+        """
         pass
 
     def get_layout_type(self) -> LayoutType:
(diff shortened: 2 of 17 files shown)


msaroufim:
Since you're alraedy running this, what was the tok/s you were getting and was the output coherent?

jerryzh168:
> Since you're already running this, what was the tok/s you were getting and was the output coherent?

I'm only getting 20 tokens/sec right now, still trying to see what is the problem

gau-nernst:
Still digesting the change. Just some questions about high-level design:
- From what I understand, when we construct an AQT with FPx, AQT object will contain `FpxAQTLayout`, and then `FpxAQTLayout` will contain `FpxTensor`. It's a bit confusing. Do we need `FpxTensor`?
- I don't really like the fact that things are scattered in many files now. Make it hard to follow. Is it possible to make AQT expose a mechanism to register custom layout, so that everything related to a particular layout can stay in a single, standalone file? (maybe in the future, seems like a big design change)
- About the naming, should we keep some of the names somewhat related to the original work? i.e. Quant-LLM / FP6-LLM

jerryzh168:
> * From what I understand, when we construct an AQT with FPx, AQT object will contain `FpxAQTLayout`, and then `FpxAQTLayout` will contain `FpxTensor`. It's a bit confusing. Do we need `FpxTensor`?

yeah that's correct, FpxTensor is supposed to mimic the native dtype Tensor like uint4 tensor, if we add fpx dtypes (e.g. fp6e3m2 etc.) in core in the future, this tensor subclass will be used to back up these dtypes. it's similar to the UintxTensor

> * I don't really like the fact that things are scattered in many files now. Make it hard to follow. Is it possible to make AQT expose a mechanism to register custom layout, so that everything related to a particular layout can stay in a single, standalone file? (maybe in the future, seems like a big design change)

oh it's intermediate step, I'm planning to move quant_llm out of prototype in the next PR to reduce the size of the PR

yeah I think custom layout can already be registered separately, e.g. https://github.com/pytorch/ao/blob/2c8e3f368613f896fde14cc24b9268c9e125575e/torchao/dtypes/uintx/Uintx.py#L152 we 'll put fpx in a separate folder as well

> * About the naming, should we keep some of the names somewhat related to the original work? i.e. Quant-LLM / FP6-LLM
maybe this can live in kernel name? I'm trying to add this as a "standard" fpx dtype actually

jerryzh168:
> Since you're already running this, what was the tok/s you were getting and was the output coherent?

updated summary

jerryzh168:
please take a look again @msaroufim @gau-nernst @jcaip

torchao/dtypes/fpx/fpx.py
```diff
@@ -347,101 +354,81 @@ def from_scaled_tc_fpx(tensor: Tensor, ebits: int, mbits: int, scale=None) -> Te
...
     __torch_dispatch__ = classmethod(_dispatch__torch_dispatch__)
 
     @staticmethod
-    def __new__(cls, fpx_data: Tensor, scale: Tensor, ebits: int, mbits: int):
-        assert fpx_data.ndim == 2
-        assert fpx_data.dtype == torch.uint8
-        shape = (fpx_data.shape[0], fpx_data.shape[1] // (1 + ebits + mbits) * 8)
+    def __new__(cls, packed_fpx_data: Tensor, ebits: int, mbits: int):
+        assert packed_fpx_data.ndim == 2
+        assert packed_fpx_data.dtype == torch.uint8
```
drisspg, line 369: maybe add some docs on the expected layout of the packed data

jerryzh168: added

torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -259,6 +271,35 @@ def from_float_static(
             dtype=input_float.dtype,
         )
 
+    @classmethod
+    def from_float_fpx(
```
jerryzh168, line 275: will merge this into `from_float` in a separate PR, as discussed in the meeting

torchao/dtypes/fpx/fpx.py
```diff
@@ -347,101 +354,90 @@ def from_scaled_tc_fpx(tensor: Tensor, ebits: int, mbits: int, scale=None) -> Te
...
+    def get_plain(self):
+        """Returns the unpacked uint8 fpx data
+        """
+        return unpack_tc_fpx(self.packed_fpx_data, 1 + self.ebits + self.mbits)
 
-    return out.view(*act.shape[:-1], out_dim).to(act.dtype)
+    @classmethod
+    def from_uint8(cls, fpx_data: Tensor, ebits: int, mbits: int):
+        packed_fpx_data = pack_tc_fpx(fpx_data, 1 + ebits + mbits)
+        return cls(packed_fpx_data, ebits, mbits)
```
gau-nernst, line 426: My concern with this is that `pack_tc_fpx()` and `unpack_tc_fpx()` actually packs data in a layout optimized for Tensor Cores, as outlined in FP6-LLM paper (the idea should be similar to tinygemm TensorCoreLayout, though I don't think they are exactly identical). So I think this should be under a Layout class instead. This wouldn't be a generic FPx data (which would raise the question, what should a generic, placeholder FPx data layout should look like).

jerryzh168: I see.. OK in that case I can rename this, what is a good name? `FpxQuantLLMLayout`?

gau-nernst: That should be fine! `FpxLLMLayout` is also possible. No big preference on this.
Then we can mention the original paper+repo in the docstring.
Also, I think it's a good idea to add a short docstring about the layout of returned value from `.get_plain()` and expected layout of `.from_uint8(fpx_data)` i.e. unpacked FPx data with padding - `00SEEEMM` for FP6 E3M2. I think MX dtypes have some good docstring for this.

torchao/dtypes/fpx/fpx.py
```diff
@@ -347,101 +354,90 @@ def from_scaled_tc_fpx(tensor: Tensor, ebits: int, mbits: int, scale=None) -> Te
...
+class FpxTensor(Tensor):
+    """FpxTensor represents a Tensor with dtype fpx(ebits=a, mbits=b),
+    it has a internal tensor field of "packed_fpx_data", which is packed from the
+    uint8 unpacked data (the output of `quantize_affine_fpx` operator)
+
+    At a high level packing is done by grouping bits into 1 bit fragments (shards), 2 bit fragments and
+    4 bit fragments each fragments are packed separately and concatenated together.
+    For example for 6 bit dtype, we can extract the first 4 bits for all elements and pack them together
+    in a fragment, and extract the last 2 bits for all elements and pack them into fragment, in the end
+    we concatenate the fragments together.
```
gau-nernst, line 366: This is nice!

torchao/dtypes/fpx/fpx.py
jerryzh168: makes sense, updated

gau-nernst: Yea the problem is we don't know what FPx plain layout should look like for odd-bitwidth. For FP4, it's natural to pack 2 values in a byte. FP1 and FP2 does not make sense so I don't think they will exist 😄. For the rest, FP3, FP5, FP6, FP7, it's tricky to define their plain layout.

test/dtypes/test_fpx.py
```diff
@@ -69,61 +73,41 @@ def test_from_scaled_tc_fpx_compile(self, ebits, mbits, device):
...
+        scale = choose_qparams_affine_fpx(x, ebits, mbits)
+        x = quantize_affine_fpx(x, scale, ebits, mbits)
+        layout_type = FpxTensorCoreLayoutType(ebits, mbits)
+        fpx_layout_tensor = FpxTensorCoreAQTLayout.from_plain(x, scale, None, layout_type).cuda()
+        assert fpx_layout_tensor.device.type == "cuda"
+        fpx_layout_tensor = fpx_layout_tensor.cpu()
+        assert fpx_layout_tensor.device.type == "cpu"
 
     @pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA not available")
+    @pytest.mark.skipif(not TORCH_VERSION_AT_LEAST_2_5, reason="quantization only works with torch.compile for 2.5+")
```
gau-nernst, line 91: Just curious, what is the reason that the new change doesn't work with previous versions of PyTorch anymore? Don't think anyone is actively using this, so probably not too important to avoid BC.

test/dtypes/test_fpx.py
```diff
@@ -69,61 +73,41 @@ def test_from_scaled_tc_fpx_compile(self, ebits, mbits, device):
...
-        linear = torch.nn.Linear(IC, OC, device=device)
-        fpx_linear = copy.deepcopy(linear)
-        quantize_(fpx_linear, fp6_llm_weight_only())
-
-        x = torch.randn(N, IC, device=device, dtype=torch.half)
-        expected = fpx_linear(x)
-        actual = torch.compile(fpx_linear, fullgraph=True)(x)
-        torch.testing.assert_close(actual, expected)
+        # somehow compile now changes the result a bit
+        torch.testing.assert_close(actual, expected, rtol=5e-4, atol=5e-4)
```
gau-nernst, line 107: Might be because the dtype is changed to FP32. The FPx kernel is using FP16 for computation.

torchao/dtypes/fpx/README.md
```diff
@@ -5,15 +5,18 @@ This is a FP16 x FPx mixed matmul kernel optimized for io bound workloads per [F
...
 
 model = ...
 model.half()  # not necessary, but recommeneded to maintain accuracy
-quantize_(model, fp6_llm_weight_only())  # convert nn.Lineaer.weight to FP6 E3M2 in-place
+# fp6 with ebits = 3 and mbits = 2
+quantize_(model, fpx_weight_only(3, 2))
 
 # for generic FPx EyMz where x = 1 + y + z
-# quantize_(model, quant_llm_fpx_weight_only(2, 2))  # use FP5 E2M2 instead
+# quantize_(model, fpx_weight_only(2, 2))  # use FP5 E2M2 instead
```
gau-nernst, line 19: Probably can remove this

test/dtypes/test_fpx.py
```diff
@@ -8,23 +8,27 @@
...
+
+from torchao.utils import TORCH_VERSION_AT_LEAST_2_5
 
 
 _DEVICES = ["cpu"] + (["cuda"] if torch.cuda.is_available() else [])
 _FPx_DTYPES = [(3, 2), (2, 2)]
 
 
-class TestQuantLlmLinearWeight(TestCase):
+class TestFpxTensor(TestCase):
```
gau-nernst, line 31: This should be `TestFpxTensorCoreAQTLayout`?

torchao/dtypes/fpx/fpx.py
```diff
@@ -347,101 +355,74 @@ def from_scaled_tc_fpx(tensor: Tensor, ebits: int, mbits: int, scale=None) -> Te
 ]
 
 
-class QuantLlmLinearWeight(Tensor):
+class FpxTensor(Tensor):
```
gau-nernst, line 358: Is `FpxTensor` still used anywhere? If not, we can delete this class?

gau-nernst (approved):

gau-nernst:
Just rmb about this. We have to update the benchmark script too 😅 

https://github.com/pytorch/ao/blob/main/benchmarks/benchmark_fp6_llm.py

torchao/dtypes/fpx/fpx.py
jerryzh168: I thought grouping things with 1, 2, 4 bit shards feels like the standard approach? this is how uintx is packed as well?

gau-nernst: I wouldn't say it's "standard" (I might be wrong). @vayuda adopts that idea from FP6-LLM also 😅.

test/dtypes/test_fpx.py
jerryzh168: this is because after integrating this into AffineQuantizedTensor, we now have multiple levels of tensor subclass, and it does not work with torch.compile until we upgrade pytorch to after 2.5

test/dtypes/test_fpx.py
jerryzh168: OK, I can revert this one, I thought it was required

torchao/dtypes/fpx/fpx.py
jerryzh168: ah I see, OK, maybe I should change uintx to say it's optimized for something as well, instead of saying it's native..

Issue #642, opened 2024-08-08
NF4 quantization slower on 0.3 vs 0.1


ebsmothers:
Hi, we're observing a slowdown in our torchtune QLoRA recipe initialization after changing from version 0.1 to 0.3 (I haven't checked 0.4 yet but will do so shortly). This was first pointed out in https://github.com/pytorch/torchtune/issues/1246 and I believe the cause is coming from some changes in torchao. 

Repro: from a torchtune git install

```
# Just some commit hash from right before we upgraded to 0.3
git checkout 52e328337579e9b84ba7f2448b29a6de7c5d8db3
pip install torchao==0.1

# Save time.perf_counter() on init and then log the delta with perf_counter()
# here: https://github.com/pytorch/torchtune/blob/0a407712eda252573326074d33af0a66c2d2990e/recipes/lora_finetune_single_device.py#L539
tune run lora_finetune_single_device --config llama3/8B_qlora_single_device
>>> 15.1960636760341

# Do the same on 0.3
pip install torchao==0.3
# also need to comment some quant APIs out to fix import errors
tune run lora_finetune_single_device --config llama3/8B_qlora_single_device
>>> 95.78260190901347
```

msaroufim:
@drisspg @weifengpy  do you have any guesses as to what might be going on?

ebsmothers:
@msaroufim I haven't done that yet, but since it's on a fixed commit of torchtune my assumption is that it shouldn't matter (lmk if this is mistaken though)

weifengpy:
> @drisspg @weifengpy do you have any guesses as to what might be going on?

to avoid peaking GPU memory at NF4 quantization, Driss landed a fix that quantize in chunks  https://github.com/pytorch/ao/pull/196

if we checkout the PR and micro benchmark before/after the change, we should be able to confirm. 

If that's confirmed, it's feature 🍡 (not a bug) . NF4 quantization is one-time efforts. it only affects TTFB (checkpoint loading)

msaroufim:
interesting, the issue though is the quantization is taking on the order of 5 min relative to 10s that seems like a lot

weifengpy:
> interesting, the issue though is the quantization is taking on the order of 5 min relative to 10s that seems like a lot

good point. quantizing in chunks might not be the only reason

msaroufim:
Still investigating but I tried timing all the tests we have using `pytest --durations=0 test/dtypes/test_nf4.py`

[...]

PR #699, opened 2024-08-17
Add option to move param to `device` before quantization

### LINKED ISSUE #655: Per Layer Streaming Quantization
A tradeoff users have often complained, most recently @aredden about is either they
1. quantize on CPU and then push the model to GPU -> Slow quantization but VRAM efficient
2. Push to model to GPU and then quantize on GPU -> Fast quantization but needs lots of VRAM

Instead we could have a utility that sends one layer at a time to the gpu, quantizes it and then sends in a new layer synchronously. Granted this workflow seems to interact in a clunky way with `torch.compile` where we don't compile things layer wise and generally expect the model to be on the device where its compiled

gau-nernst:
Fixes #655 

Test

```python
import torch
from torchao._models.llama.model import Transformer, ModelArgs
from torchao.quantization.quant_api import quantize_, int8_weight_only
import time
import gc

def get_model():
    with torch.device("meta"):
        model = Transformer(ModelArgs())
    model.to_empty(device="cpu").bfloat16()
    return model

def clear_cuda():
    gc.collect()
    torch.cuda.empty_cache()
    torch.cuda.reset_peak_memory_stats()
    assert torch.cuda.max_memory_reserved() == 0

model = get_model()
time0 = time.perf_counter()
quantize_(model, int8_weight_only())
model.to("cuda")
torch.cuda.synchronize()
duration = time.perf_counter() - time0
print("Quantize on CPU:")
print(f"  - Time taken: {duration:.2f} s")
print(f"  - Peak memory: {torch.cuda.max_memory_reserved() / 1e9:.2f} GiB")
del model
clear_cuda()

model = get_model()
time0 = time.perf_counter()
model.to("cuda")
quantize_(model, int8_weight_only())
torch.cuda.synchronize()
duration = time.perf_counter() - time0
print("Quantize on CUDA:")
print(f"  - Time taken: {duration:.2f} s")
print(f"  - Peak memory: {torch.cuda.max_memory_reserved() / 1e9:.2f} GiB")
del model
clear_cuda()

model = get_model()
time0 = time.perf_counter()
quantize_(model, int8_weight_only(), device="cuda")
torch.cuda.synchronize()
duration = time.perf_counter() - time0
print("Move to CUDA and quantize each param individually:")
print(f"  - Time taken: {duration:.2f} s")
print(f"  - Peak memory: {torch.cuda.max_memory_reserved() / 1e9:.2f} GiB")
del model
clear_cuda()
```

```
Quantize on CPU:
  - Time taken: 10.48 s
  - Peak memory: 6.99 GiB
Quantize on CUDA:
  - Time taken: 1.96 s
  - Peak memory: 14.50 GiB
Move to CUDA and quantize each param individually:
  - Time taken: 1.94 s
  - Peak memory: 8.29 GiB
```

@jerryzh168 Do you have any suggestions what tests I should add for this?

This PR may help the slow NF4 quantization issue observed in torchtune too #642 (though I'm not sure what is the NF4 quantization API. Does it use `quantize_()`?)

### DIFF
--- a/test/quantization/test_quant_api.py
+++ b/test/quantization/test_quant_api.py
@@ -698,7 +698,6 @@ class TestQuantFlow(TestCase):
         torch.cuda.synchronize()
         time_baseline = time.perf_counter() - time0
         memory_baseline = torch.cuda.max_memory_allocated()
-        print(memory_baseline)
 
         del m
         reset_memory()
@@ -707,7 +706,6 @@ class TestQuantFlow(TestCase):
         quantize_(m, int8_weight_only(), device="cuda")
         time_streaming = time.perf_counter() - time0
         memory_streaming = torch.cuda.max_memory_allocated()
-        print(memory_streaming)
 
         for param in m.parameters():
             assert param.is_cuda
--- a/test/quantization/test_quant_api.py
+++ b/test/quantization/test_quant_api.py
@@ -54,6 +54,8 @@ from torchao._models.llama.model import Transformer, prepare_inputs_for_model
 from torchao.utils import unwrap_tensor_subclass
 import copy
 import tempfile
+import gc
+import time
 from torch.testing._internal.common_utils import TestCase
 
 
@@ -679,6 +681,38 @@ class TestQuantFlow(TestCase):
 
         res = m_copy(*example_inputs)
         self.assertEqual(res, ref)
+
+    @unittest.skipIf(not TORCH_VERSION_AT_LEAST_2_4, "Test only enabled for 2.4+")
+    @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
+    def test_quantized_model_streaming(self):
+        def reset_memory():
+            gc.collect()
+            torch.cuda.empty_cache()
+            torch.cuda.reset_peak_memory_stats()
+
+        reset_memory()
+        m = ToyLinearModel()
+        time0 = time.perf_counter()
+        m.to(device="cuda")
+        quantize_(m, int8_weight_only())
+        torch.cuda.synchronize()
+        time_baseline = time.perf_counter() - time0
+        memory_baseline = torch.cuda.max_memory_allocated()
+        print(memory_baseline)
+
+        del m
+        reset_memory()
+        m = ToyLinearModel()
+        time0 = time.perf_counter()
+        quantize_(m, int8_weight_only(), device="cuda")
+        time_streaming = time.perf_counter() - time0
+        memory_streaming = torch.cuda.max_memory_allocated()
+        print(memory_streaming)
+
+        for param in m.parameters():
+            assert param.is_cuda
+        self.assertLess(time_streaming, time_baseline * 1.1)
+        self.assertLess(memory_streaming, memory_baseline)
 
 
 if __name__ == "__main__":
--- a/test/quantization/test_quant_api.py
+++ b/test/quantization/test_quant_api.py
@@ -55,7 +55,6 @@ from torchao.utils import unwrap_tensor_subclass
 import copy
 import tempfile
 import gc
-import time
 from torch.testing._internal.common_utils import TestCase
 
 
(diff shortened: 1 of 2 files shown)


jerryzh168:
nf4 is not yet using `quantize_` API, it is module swapping linear with a LoraLinear: https://github.com/pytorch/ao/blob/b523f9f9e15b6fb80d10f585d9cf45e0c5e4d10e/test/dtypes/test_nf4.py#L544, maybe we could incorporate this in `quantize_` as well if there is a weight only quantization use case for nf4 as well. cc @drisspg 

the change makes sense, one thing I'm wondering is would it help further reduce cuda memory usage if you move the module back to cpu (original device) after quantization as well (and release the used cuda memory)? not sure if it would be slow though.

for tests, I think you'll probably need to construct a model with linear + non-linear modules and compare the max cuda memory for "Quantize on CUDA" and "Move to CUDA and quantize each param individually" and make sure the second one consumes less memory. you can add it in https://github.com/pytorch/ao/blob/main/test/quantization/test_quant_api.py I think

gau-nernst:
> nf4 is not yet using quantize_ API, it is module swapping linear with a LoraLinear

I see. So users of NF4 need to implement this logic i.e. torchtune will have to implement this instead of torchao.

> one thing I'm wondering is would it help further reduce cuda memory usage if you move the module back to cpu (original device) after quantization as well (and release the used cuda memory)? not sure if it would be slow though.

I can try this. But in the end, I think we should stick to 1 approach to avoid confusion to users.

jerryzh168:
> I see. So users of NF4 need to implement this logic i.e. torchtune will have to implement this instead of torchao.
> 
> >

yeah according to Driss, this LoraLinear code is framework specific so we don't want to offer an API for people

msaroufim:
> for tests, I think you'll probably need to construct a model with linear + non-linear modules and compare the max cuda memory for "Quantize on CUDA" and "Move to CUDA and quantize each param individually" and make sure the second one consumes less memory. you can add it in [main/test/quantization/test_quant_api.py](https://github.com/pytorch/ao/blob/main/test/quantization/test_quant_api.py?rgh-link-date=2024-08-17T03%3A49%3A01Z) I think

Agreed this is the right test

> the change makes sense, one thing I'm wondering is would it help further reduce cuda memory usage if you move the module back to cpu (original device) after quantization as well (and release the used cuda memory)? not sure if it would be slow though.

This scenario seems to optimize for people that have a GPU but want the model to run on the CPU, that seems rarer than the alternative of someone who wants to fit a model that in full precision does not fit on their GPU

Otherwise yeah once CI is green can merge this

gau-nernst:
> This scenario seems to optimize for people that have a GPU but want the model to run on the CPU, that seems rarer than the alternative of someone who wants to fit a model that in full precision does not fit on their GPU

The difference in peak memory between this PR and quantize on CPU (8.29 GiB vs 6.99 GiB) is probably due to the largest `nn.Linear` layer (LM head). I think what @jerryzh168 means is that, we can further eliminate this difference by moving quantized weight back to CPU one-by-one, so the peak VRAM after quantization will be the largest `nn.Linear` layer. When we move the whole CPU quantized model back to CUDA again, the peak memory should be the same as quantize on CPU (6.99 GiB), if memory fragmentation doesn't haunt us. But yea it's a bit more convoluted.

gau-nernst:
Using Llama2-7B

```
Baseline CPU: Quantize on CPU, then move the whole model to CUDA
  - Time taken: 10.48 s
  - Peak memory: 6.99 GiB

Baseline CUDA: Move the whole model to CUDA, then quantize on CUDA
  - Time taken: 1.96 s
  - Peak memory: 14.50 GiB

Approach 1: For each param, move to CUDA, quantize on CUDA
  - Time taken: 1.94 s
  - Peak memory: 8.29 GiB

Approach 2: For each param, move to CUDA, quantize on CUDA, move back to CPU. Then move the whole model to CUDA
  - Time taken: 5.58 s
  - Peak memory: 7.03 GiB
```

There are tradeoffs between approach 1 and 2. Lmk which one you prefer. I think we should only keep one approach to avoid confusion.

Another thing we can consider is to measure the size of the weight to decide what to do with it. E.g. if the weight exceeds xx size (which hopefully only the LM head is), we will always quantize on CPU -> avoid the memory spike. Again, this also adds seemingly unnecessary complexity.

msaroufim:
I like approach 1 better because it's a better tradeoff between the options we already have

msaroufim (approved):

jerryzh168 (approved):
Thanks for benchmarking different approaches @gau-nernst , yeah Approach 1 sounds good to me as well, it is simpler and easier to explain in API as well

PR #706, opened 2024-08-19
add ability to calculate amax in tiles

drisspg:
Stacked PRs:
 * #709
 * #707
 * __->__#706

--- --- ---

### add ability to calculate amax in tiles

ghstack-source-id: 83ccec3ec66f30b9d75146d0fc7b1137ea7574c4
Pull Request resolved: https://github.com/pytorch/ao/pull/682

### DIFF
--- a/torchao/float8/float8_scaling_utils.py
+++ b/torchao/float8/float8_scaling_utils.py
@@ -8,7 +8,7 @@
 Utilities for scaling high precision tensors to float8.
 """
 
-from typing import Optional
+from typing import Optional, Tuple, Literal
 
 import torch
 
@@ -36,6 +36,7 @@ def hp_tensor_to_float8_dynamic(
     linear_mm_config: LinearMMConfig,
     reduce_amax: bool = False,
     gemm_input_role: GemmInputRole = GemmInputRole.INPUT,
+    tile_size: Optional[Tuple[int, int]] = None,
 ) -> Float8Tensor:
     """
     Given a high precision tensor `hp_tensor`,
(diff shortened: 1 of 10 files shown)


torchao/float8/float8_utils.py
```diff
@@ -112,9 +110,12 @@ def tensor_to_amax(x: torch.Tensor, reduce_amax: bool = False) -> torch.Tensor:
 
 @torch.no_grad()
 def tensor_to_scale(
-    x: torch.Tensor, float8_dtype: torch.dtype, reduce_amax: bool = False
+    x: torch.Tensor,
+    float8_dtype: torch.dtype,
+    tile_size: Optional[Tuple[int, int]],
```
vkuzo, line 115: default value?

torchao/float8/float8_scaling_utils.py
```diff
@@ -116,11 +128,13 @@ def _maybe_initialize_amaxes_scales_for_float8_cast(
         # Note: we need to enable distributed reduction here in order
         # to match numerics between single GPU and multi GPU code for
         # activations and gradients
-        new_amax = tensor_to_amax(x, reduce_amax=reduce_amax)
+
+        # We hardcode TensorWise Scaling on delayed scaling for now
```
vkuzo, line 132: nit: throw an exception instead of silently changing the setting?

torchao/float8/float8_utils.py
drisspg: It would greatly reduce the the LOC of this PR but I have been wrestling with all our default values.

We we have like 3 abstraction layers where you can do casting with a mix of default and non default values. IMO I typically like have default values at the top most layer and then everything else requires args. This was my attempt to get us closer

torchao/float8/float8_scaling_utils.py
drisspg: Hmm so I actually I think for this PR:
`hp_tensor_to_float8_delayed` should actually assert that tile_size is None and then I can add the arg to this func with the same assert

torchao/float8/float8_utils.py
vkuzo: makes sense, maybe make tile_size keyword only to make the tensor wise case more readable?

torchao/float8/float8_scaling_utils.py
vkuzo: can also assert in `Float8LinearConfig.__post_init__`

PR #708, opened 2024-08-19
Add high-level operator interface

metascroy:
Summary:
Here is a draft of the high-level kernel operator.  This is something I quickly put together to get early feedback, and it will be iterated.  It closely follows what I had in the design doc, but I think it's easier to see how things fit together in the code.

The operator uses parallel_for and get_num_threads.  At build time, we need to link these to ET/PT implementations.  In this diff, they are single-threaded or OMP-backed implementations that I threw together.

The operator (inside channelwise_8bit_activation_groupwise_lowbit_weight.h) depends on a ukernel_config and tiling configs.  But in the wrapper, these can be hidden if desired (see examples/ for how).

Reviewed By: digantdesai

Differential Revision: D60321449


jerryzh168 (approved):
stamping

PR #711, opened 2024-08-20
Small fix for micro benchmark code

jerryzh168:
Summary:
There seems to be some problems running benchmark_aq.py:
```
torch._dynamo.exc.Unsupported: torch.* op returned non-Tensor int call_function <method 'size' of 'torch._C.TensorBase' objects>
```
when we run the benchmark with multiple shapes sometimes. But the problem will be gone if we reset the dynamo caches before each benchmark run

Test Plan:
python benchmarks/benchmark_aq.py

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/benchmarks/benchmark_aq.py
+++ b/benchmarks/benchmark_aq.py
@@ -17,6 +17,7 @@ from torchao.quantization.quant_api import (
     _replace_with_custom_fn_if_matches_filter,
 )
 import copy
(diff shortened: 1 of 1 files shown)


msaroufim (approved):

msaroufim:
Unfortunate change, can we also have some open issue maybe in core to flag

Issue #577, opened 2024-07-30
GPTQ implementation with tensor subclasses


HDCharles:
Problem: The current implementation of [GPTQ](https://arxiv.org/abs/2210.17323) (A technique used to improve quantization accuracy) relies on [model tracing](https://github.com/pytorch/ao/blob/main/torchao/quantization/GPTQ.py#L87) which cannot handle certain op sequences of various models, as an example, the [gpt-fast llama implementation’s kv_cache update](https://github.com/pytorch-labs/gpt-fast/blob/main/model.py#L80-L89) causes an issue. We worked around this for [our version](https://github.com/pytorch/ao/blob/main/torchao/_models/llama/model.py#L85-L98) where we use a flag to determine whether we use the fast op sequence or the one that works with tracing.

Goal: re-implement GPTQ as a technique which uses tensor subclasses to avoid this issue.

Background:
GPTQ requires us to, for each linear,
1) track all activations going to that linear
2) apply the GPTQ algorithm to update that linear’s weight
3) use the updated weight to generate outputs for that linear
4) repeat for the next linear

The main complication is that, to track n activations going to each linear, we’d normally run the model for n inputs. But if we update the weight of the 1st linear (and want to use the updated weight for the activations of the 2nd linear), we’d then have to run the model n more times. Meaning we’d have to run the model n*L (L=layers) times in total, which is extremely slow.

Instead, we want to run the model for each input, but ONLY up to the first linear, then pause, do the algorithm to update the weight, get the outputs for the updated weight and then, unpause and continue on until we hit the next linear….etc.

Previously we did this using[ tracing and a custom fx.Interpreter](https://github.com/pytorch/ao/blob/main/torchao/quantization/GPTQ.py#L57C25-L57C39) but as mentioned above, its a bit fragile. Instead we want to use a tensor subclass which will operate as a "multi-tensor", containing the tensors from several inputs. Each time it encounters a normal op, it will do the normal op to each constituent tensor, and generate a multi-tensor output so that it will propagate through the network. When it encounters a linear op, it will instead do the GPTQ weight update...etc outlined above.

Starting Point:
[...]

PR #716, opened 2024-08-20
[WIP] GPTQ MultiTensor refactor

danielpatrickhug:
This is a PR in response to issue #577 
The primary contributions of this PR are as follows:
1. Refactor the GPTQ module to use the MultiTensor class described [here](https://gist.github.com/HDCharles/a1b575bbf8875f994af8a01b225e1227) instead of the GenericGPTQRunner class which subclasses the fx.Interpreter.
2. Refactor the GPTQQuantizer class and the Int4WeightOnlyGPTQQuantizer class which previously depended on the GenericGPTQRunner class.


jerryzh168:
@danielpatrickhug thanks for working on this, could you follow the flow described in https://github.com/pytorch/ao/pull/721 for the refactor? we want to deprecate the "Quantizer" based APIs

torchao/quantization/GPTQ_MT.py
```diff
@@ -0,0 +1,446 @@
...
+            not skip_gptq
+            and cls.is_linear_layer(func)
+        )
+
+        if quantize_linear:
+            H = 0
+            total_batches = 0
+            
+        outputs = []
+        with torch._C.DisableTorchFunctionSubclass():
```
jerryzh168, line 128: this part can also be decoupled from the MultiTensor I feel

danielpatrickhug:
> @danielpatrickhug thanks for working on this, could you follow the flow described in #721 for the refactor? we want to deprecate the "Quantizer" based APIs
Sure no problem, I'll work on adapting it to this API format.

HDCharles:
> @danielpatrickhug thanks for working on this, could you follow the flow described in #721 for the refactor? we want to deprecate the "Quantizer" based APIs

i don't think we should combine a functional refactor with an API refactor. This should be a drop in replacement for what we had, otherwise we should make those API changes first and let OSS devs make their changes on top. I can do the API alignment after the functional refactor is complete.

PR #581, opened 2024-07-31
Add Auto-Round support

yiliu30:
Resolve https://github.com/pytorch/ao/issues/533

### Description

- [x] Integrated `Auto-Round` with `quantize`_ API using hooks + `MultiTensor`.
- [x] Exported the optimized `qweight` to `AffineQuantizedTensor` to leverage the `tinygemm` and `Uintx` kernels.
- [x] Evaluated the accuracy for `Llama2/3/3.1` on 5 popular `lm-eval` tasks  (more tests are on the way).
- [x] Added `Auto-Round` to the generation benchmarking for `Llama2/3`, (`Llama 3.1` not yet tested as it was landed a few days ago).
- [x] Small fix for the Llama model https://github.com/pytorch/ao/pull/769

### Usage

```python
from torchao.prototype.autoround.core import prepare_model_for_applying_auto_round_
from torchao.prototype.autoround.core import apply_auto_round

prepare_model_for_applying_auto_round_(
    model,
    is_target_module=is_target_module,
    bits=4,
    group_size=128,
    iters=200,
    device=device,
)

input_ids_lst = []
for data in dataloader:
    input_ids_lst.append(data["input_ids"].to(model_device))

multi_t_input_ids = MultiTensor(input_ids_lst)
out = model(multi_t_input_ids)

quantize_(model, apply_auto_round(), is_target_module)
```

For E2E examples, please refer [README.md](https://github.com/yiliu30/torchao-fork/tree/auto_round_support-3/torchao/prototype/autoround)

cc @thuang6 @ftian1 @wenhuach21

PR #721, opened 2024-08-21
Example for GPTQ-like calibration flow

jerryzh168:
Summary:
This PR adds example for GPTQ-like calibration flow, where we (1) optimize (quantize) one module at a time
(2) with each optimization step, we need to get a set of all calibration data (3) the output of each module is calculated based on the optimized (quantized) module, and then pass down to the next module

In this tutorial we mainly use two things:
(1) MultiTensor subclass https://gist.github.com/HDCharles/a1b575bbf8875f994af8a01b225e1227 (2) Module forward hooks

Potential use cases:
GPTQ, Auto-Round(https://github.com/pytorch/ao/pull/581), SpinQuant (according to Tijmen)

Test Plan:
python tutorials/calibration_flow/gptq_like.py

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/tutorials/calibration_flow/gptq_like.py
+++ b/tutorials/calibration_flow/gptq_like.py
@@ -1,18 +1,37 @@
 """
-This is a example flow for GPTQ like calibration flows, where we:
-(1) optimize (quantize) one module at a time
-(2) with each optimization step, we need to get a set of all calibration data
-(3) the output of each module is calculated based on the optimized (quantized) module, and then pass down to the next module
+Traditional calibration flow has the following flow (see static_quant.py for code examples):
+
+(1). insert input/output observers to the modules
+(2). run the model with calibration data so the observers in the model can record the statistics of the data flowing through them, observation does not change the output of a layer
+(3). convert the observed module to quantized module (or quantize the weights with the quantization parameters based on the observer statistics)
+
+By GPTQ like flow we mean a flow that does not fit into the above flow very well because
+(1) optimize (quantize) one layer (module) at a time and the the output of each layer is calculated based on the optimized (quantized) module, and then pass down to the next layer, this means layers are not independent
+(2) with each optimization step, we need to use all the input data for that layer instead of just some derived statistics like min_val/max_val
+
+To use the traditional flow, we'd need to
+(1). insert observers for the layer we want to optimize, that will record all the inputs
+(2). each time, run the entire model upto layer N, then optimize layer N, and then
+continue the process for layer N+1, this means we'll need to run O(N^2) layers in total.
+
+So we'd like to use a flow that only runs each layer a constant time so we get O(N) time complexity.
 
 In this tutorial we mainly use two things:
 (1) MultiTensor subclass https://gist.github.com/HDCharles/a1b575bbf8875f994af8a01b225e1227
-(2) Module forward hooks
+
+It stores a list of Tensors (calibration data). This is used to pass around all the calibration data to a layer, we can optimize the layer, and then output another MultiTensor object for future layers.
+
+(2) Module forward pre hooks (https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.register_forward_pre_hook)
+
+This is used for modifying the behavior of the forward function of `module`, it allows [modification](https://discuss.pytorch.org/t/use-forward-pre-hook-to-modify-nn-module-parameters/108498/2) of the module itself, and also allows modifying the input of the module.
+
+This can be used when we try to optimize (quantize) the layer, and then want the next layer to consume the output of the optimized layer directly.
 """
 import torch
 import torch.nn as nn
 from torch.utils._pytree import tree_flatten, tree_unflatten
 import gc
-from typing import Tuple, Dict
+from typing import Tuple, Dict, Any
 from torchao.quantization.utils import compute_error
 from torchao.dtypes import to_affine_quantized_static
 from torchao.quantization import quantize_
@@ -27,6 +46,8 @@ from torchao.quantization.quant_primitives import (
     MappingType,
     fake_quantize_affine,
 )
+
+torch.manual_seed(0)
 
 class MultiTensor(torch.Tensor):
     @staticmethod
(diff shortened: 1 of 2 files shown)


tutorials/calibration_flow/gptq_like.py
```diff
@@ -0,0 +1,273 @@
+"""
+This is a example flow for GPTQ like calibration flows, where we:
+(1) optimize (quantize) one module at a time
+(2) with each optimization step, we need to get a set of all calibration data
+(3) the output of each module is calculated based on the optimized (quantized) module, and then pass down to the next module
+
+In this tutorial we mainly use two things:
+(1) MultiTensor subclass https://gist.github.com/HDCharles/a1b575bbf8875f994af8a01b225e1227
```
msaroufim, line 8: n00b q but it's not super clear to me what is a multi tensor exactly and why it's relevant here?

tutorials/calibration_flow/gptq_like.py
```diff
@@ -0,0 +1,273 @@
...
+        self.two_linear1 = TwoLinear()
+        self.two_linear2 = TwoLinear(128, 256)
+
+    def forward(self, x, y):
+        x1 = self.two_linear1(x, y)
+        x2 = self.two_linear2(x1, x1)
+        return x2
+
+def _is_two_linear(mod, fqn):
+    return isinstance(mod, TwoLinear)
```
msaroufim, line 165: unnecessary indirection?

tutorials/calibration_flow/gptq_like.py
```diff
@@ -0,0 +1,273 @@
...
+        del observed_two_linear.input_scale1
+        del observed_two_linear.input_zp1
+        del observed_two_linear.input_scale2
+        del observed_two_linear.input_zp2
+        return observed_two_linear
+
+    return _apply_activation_static_quant
+
+
+example_inputs = (torch.randn(32, 64), torch.randn(32, 64),)
```
msaroufim, line 250: Up until this point this PR read as fairly abstract, I'm not sure who the intended audience is

Is this someone trying to implement a gptq like workflow in AO? That's an audience of 1? Also how are multitensors and forward hooks relevant for gptq exactly seems like there's a missing step in the explanation

tutorials/calibration_flow/gptq_like.py
```diff
@@ -0,0 +1,273 @@
...
+        kwargs: Dict[str, MultiTensor],
+        output: Tuple[MultiTensor],
+    ):
+        # remove the hook to avoid recursive calls
+        module._forward_hook_handle.remove()
+        # in this case args will be a tuple of 2 MultiTensors since
+        # the TwoLinear module takes 2 inputs, and each input will collect a list of normal Tensors
+
+        # we can use these two MultiTensors to calculate the quantization parameters for each input Tensor
+        act_obs1 = AffineQuantizedMinMaxObserver(MappingType.ASYMMETRIC, torch.uint8, granularity_type=PerTensor(), eps=torch.finfo(torch.float32).eps, scale_dtype=torch.float32, zero_point_dtype=torch.int32)
```
msaroufim, line 181: I'm still not sure I follow why the hello world example for static quantization should use 2 modules? Also the TwoLinear model seems kinda contrived and more of a test instead of a tutorial?

tutorials/calibration_flow/gptq_like.py
```diff
@@ -0,0 +1,273 @@
...
+            new_input2.append(fake_quantize_affine(inp, inp.shape, input_scale2, input_zp2, torch.uint8))
+
+        mt1 = MultiTensor(new_input1)
+        mt2 = MultiTensor(new_input2)
+
+        output_with_fake_quantized_inputs = module(mt1, mt2)
+        # we can return the modified output so it can be consumed by future modules
+        return output_with_fake_quantized_inputs
+
+    def _register_forward_hook(module: torch.nn.Module):
```
msaroufim, line 214: could you add a comment as to what the hook is trying to do?

tutorials/calibration_flow/gptq_like.py
```diff
@@ -0,0 +1,273 @@
...
+        act_scale, act_zero_point = observed_two_linear.input_scale1, observed_two_linear.input_zp1
+        input_quant_func1 = lambda x: to_affine_quantized_static(x, act_scale, act_zero_point, x.shape, target_dtype)
+        linear1.weight = torch.nn.Parameter(to_linear_activation_quantized(linear1.weight, input_quant_func1), requires_grad=False)
+
+        linear2 = observed_two_linear.linear2
+        # activation quantization
+        act_scale, act_zero_point = observed_two_linear.input_scale2, observed_two_linear.input_zp2
+        input_quant_func2 = lambda x: to_affine_quantized_static(x, act_scale, act_zero_point, x.shape, target_dtype)
+        linear2.weight = torch.nn.Parameter(to_linear_activation_quantized(linear2.weight, input_quant_func2), requires_grad=False)
+        del observed_two_linear.input_scale1
```
msaroufim, line 241: why do we need these deletions?

tutorials/calibration_flow/gptq_like.py
```diff
@@ -0,0 +1,273 @@
...
+    )
+
+# using a function to align with the API in quant_api
+def apply_activation_static_quant():
+    def _apply_activation_static_quant(observed_two_linear):
+        target_dtype = torch.uint8
+
+        linear1 = observed_two_linear.linear1
+        # activation quantization
+        act_scale, act_zero_point = observed_two_linear.input_scale1, observed_two_linear.input_zp1
```
msaroufim, line 232: I'm not sure I follow why this assignment is needed 
and I'm not sure i understand why `input_quant_func1` is needed? and why those couldn't be arguments to `to_linear_activation_quantized` I haven't carefully reviewed AQT so would appreciate some more context on what's the intent here

msaroufim (changes requested):

tutorials/calibration_flow/gptq_like.py
jerryzh168: MultiTensor is a Tensor that collects a list of Tensors (calibration data) and it can be used for optimizing a single module, and also the module can output the MultiTensor result for output. maybe I can add some motivation here

tutorials/calibration_flow/gptq_like.py
jerryzh168: this is used twice so probably better to keep it I think, also `_is_two_linear` has two args..

tutorials/calibration_flow/gptq_like.py
jerryzh168: this is any flow that's similar to GPTQ, that I described in the beginning, but I'll add more explanations/motivation.

current known use cases are GPTQ, AutoRound and SpinQuant as mentioned in the summary

tutorials/calibration_flow/gptq_like.py
jerryzh168: yeah it's a bit more complicated example, since I want to show the output of the module is modified in `        output_with_fake_quantized_inputs = module(mt1, mt2)`, I can change to one input and one module as well if you feel it's more important to keep it simple

tutorials/calibration_flow/gptq_like.py
jerryzh168: this is trying to demo modification to the module, we are deleting these args because these args are already used to quantize the weight and we are restoring the module to be nn.Linear here

tutorials/calibration_flow/gptq_like.py
jerryzh168: `input_quant_func1` is a general input quantization function that can be used to quantize activations, we can't hard code the args for `to_linear_activation_quantized` since the `input_quant_func` can be any function, it's not restricted to be using affine quantization here

we don't have docs for `to_linear_activation_quantized`, this is used to implement static and dynamic quantization for activations. I can add some docs for this

tutorials/calibration_flow/gptq_like.py
```diff
@@ -0,0 +1,263 @@
+"""
+Traditional calibration flow has the following flow (see static_quant.py for code examples):
```
jerryzh168, line 2: @msaroufim added a bit more context here, please let me know if this makes sense

msaroufim (approved):

Issue #698, opened 2024-08-17
[AQT-bug] `AffineQuantizedTensor` raises error with `.cuda()`


gau-nernst:
```python
import torch
from torchao.quantization.quant_api import quantize_, int8_weight_only

model = torch.nn.Linear(2048, 2048)
quantize_(model, int8_weight_only())
# model.to("cuda")  # this works
model.cuda()  # this doesn't
```

```
  File "/home/xx/code/ao/torchao/dtypes/affine_quantized_tensor.py", line 975, in _
    args[0].to(*args[1:], **kwargs)._apply_fn_to_data(torch.clone),
  File "/home/xx/code/ao/torchao/dtypes/affine_quantized_tensor.py", line 291, in to
    kwargs = self._get_to_kwargs(*args, **kwargs)
  File "/home/xx/code/ao/torchao/dtypes/affine_quantized_tensor.py", line 277, in _get_to_kwargs
    device, dtype, _, memory_format = torch._C._nn._parse_to(*args, **kwargs)
TypeError: to() received an invalid combination of arguments - got (device=torch.device, layout=torch.layout, dtype=torch.dtype, ), but expected one of:
 * (torch.device device = None, torch.dtype dtype = None, bool non_blocking = False, bool copy = False, *, torch.memory_format memory_format = None)
 * (torch.dtype dtype, bool non_blocking = False, bool copy = False, *, torch.memory_format memory_format = None)
 * (Tensor tensor, bool non_blocking = False, bool copy = False, *, torch.memory_format memory_format = None)
```

I faced similar problems with other subclasses that I implemented before. Basically need to remove the `layout` kwarg.

cc: @jerryzh168

jerryzh168:
thanks, yeah I think we want to add some basic tests to https://github.com/pytorch/ao/blob/main/test/dtypes/test_affine_quantized.py

Issue #710, opened 2024-08-20
Tensor subclass boilerplate can be consolidated


HDCharles:
A lot of code for tensor subclasses can likely be conslidated together into a base class that other classes can utilize

_get_to_kwargs:
https://github.com/pytorch/ao/blob/main/torchao/dtypes/affine_quantized_tensor.py#L64
https://github.com/pytorch/ao/blob/main/torchao/dtypes/affine_quantized_tensor.py#L276
(also needed for https://github.com/pytorch/ao/blob/main/torchao/quantization/autoquant.py#L40)

to:
https://github.com/pytorch/ao/blob/main/torchao/dtypes/affine_quantized_tensor.py#L594
https://github.com/pytorch/ao/blob/main/torchao/dtypes/affine_quantized_tensor.py#L290
https://github.com/pytorch/ao/blob/main/torchao/dtypes/affine_quantized_tensor.py#L423
(also needed for https://github.com/pytorch/ao/blob/main/torchao/quantization/autoquant.py#L40)

_apply_fn_to_data:
https://github.com/pytorch/ao/blob/main/torchao/dtypes/affine_quantized_tensor.py#L432

detach:
https://github.com/pytorch/ao/blob/main/torchao/dtypes/affine_quantized_tensor.py#L444

a default __repr__ would be nice (it caused bugs in the past)

if these were implemented using __tensor_flatten__ and __tensor_unflatten__ e.g. https://github.com/pytorch/ao/blob/main/torchao/dtypes/affine_quantized_tensor.py#L412-L421
you could write a general mixin/parent class that handles the general form of those methods

HDCharles:
@jerry39213gh thoughts?

jerryzh168:
yes, this makes sense I think, we can put this https://github.com/pytorch/ao/blob/main/torchao/dtypes/utils.py. we can also attach https://github.com/pytorch/ao/blob/ac8ce4ceb548c264a2f508f7b6a1f413d8d4454c/torchao/quantization/linear_activation_quantized_tensor.py#L98-L100 by default

there could be some issues of child class modifying the dispatch table for parent class, if this becomes an issue, we can add `class` to the dispatch table as well in the future.

PR #726, opened 2024-08-22
Fix affine quantized tensor to device calls

jerryzh168:
Summary:
Fixes: https://github.com/pytorch/ao/issues/698

Also added `TorchAOBaseTensor` addressing part of https://github.com/pytorch/ao/issues/710

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/utils.py
+++ b/torchao/utils.py
@@ -20,6 +20,7 @@ __all__ = [
     "_register_custom_op",
     "get_model_size_in_bytes",
     "unwrap_tensor_subclass",
+    "TorchAOBaseTensor",
     "TORCH_VERSION_AT_LEAST_2_2",
     "TORCH_VERSION_AT_LEAST_2_3",
     "TORCH_VERSION_AT_LEAST_2_4",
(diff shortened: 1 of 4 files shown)


torchao/utils.py
```diff
@@ -281,6 +282,30 @@ def unwrap_tensor_subclass(model, filter_fn=None):
...
+            if isinstance(arg, torch.layout):
+                args.remove(arg)
+        if "layout" in kwargs:
+            kwargs.pop("layout")
+        device, dtype, _, memory_format = torch._C._nn._parse_to(*args, **kwargs)
+        device = self.device if device is None else device
+        dtype = self.dtype if dtype is None else dtype
+        memory_format = (
+            memory_format if memory_format is not None else torch.preserve_format
+        )
```
gau-nernst, line 300: I have wondered about this for some time. Does it make sense to have `memory_format` applied to inner tensor? From what I understand, it's mainly for channels_last in convolution (https://pytorch.org/docs/stable/tensor_attributes.html#torch-memory-format). Maybe we can skip `memory_format` argument also?

gau-nernst: if the current tests pass, I think should be good for now!

gau-nernst (approved):

PR #713, opened 2024-08-20
cast `local_scale_tensor` to fp32 for precompute of fp8 dynamic scaling

crcrpar:
When a model is in bfloat16, precomputed scales seem to be in bfloat16.
This seems to cause the dtype mismatch, especially `scale_b` being bf16, after the first call of precompute:

```
# [rank0]:[rank0]:   File "/usr/local/lib/python3.10/dist-packages/torchao/float8/float8_linear.py", line 363, in forward
# [rank0]:[rank0]:     output = manual_float8_matmul.apply(input_fp8, weight_fp8.t())
# [rank0]:[rank0]:   File "/usr/local/lib/python3.10/dist-packages/torch/autograd/function.py", line 575, in apply
# [rank0]:[rank0]:     return super().apply(*args, **kwargs)  # type: ignore[misc]
# [rank0]:[rank0]:   File "/usr/local/lib/python3.10/dist-packages/torchao/float8/float8_linear.py", line 60, in forward
# [rank0]:[rank0]:     res_bits = torch.mm(input_fp8_reshaped, weight_fp8_t)
# [rank0]:[rank0]:   File "/usr/local/lib/python3.10/dist-packages/torchao/float8/float8_tensor.py", line 359, in __torch_dispatch__
# [rank0]:[rank0]:     return FLOAT8_OPS_TABLE[func](func, args, kwargs)
# [rank0]:[rank0]:   File "/usr/local/lib/python3.10/dist-packages/torchao/float8/float8_ops.py", line 181, in float8_mm
# [rank0]:[rank0]:     tensor_out = addmm_float8_unwrapped(
# [rank0]:[rank0]:   File "/usr/local/lib/python3.10/dist-packages/torchao/float8/float8_python_api.py", line 54, in addmm_float8_unwrapped
# [rank0]:[rank0]:     output = torch._scaled_mm(
# [rank0]:[rank0]: RuntimeError: Both scale_a and scale_b must be float (fp32) tensors.
```

PR #727, opened 2024-08-22
Specify output dtype to `torch.float32` in `_foreach_norm`

crcrpar:
follow-up of #713


weifengpy (approved):
this is even better for numerics. thanks for proposing

PR #732, opened 2024-08-22
Float8 inference examples

jainapurva:


drisspg (approved):
Looks good, can you also add the float8dq

PR #621, opened 2024-08-06
Add sparse marlin AQT layout

Diogo-V:
## Summary
Introduces a Layout for AQT class called [Sparse Marlin 2:4](https://github.com/IST-DASLab/Sparse-Marlin). 

## E2E Results
Screenshot 2024-09-06 at 8 57 06 AM

## Tests
test/sparsity/test_marlin.py

## Considerations for reviewers
1. To solve the issue with `torch.compile` I had to covert the section of [here](https://github.com/Diogo-V/ao/blob/f52925cb47c95eec95b1725a933fca99a8473d9d/torchao/sparsity/marlin/__init__.py#L239) and [here](https://github.com/Diogo-V/ao/blob/f52925cb47c95eec95b1725a933fca99a8473d9d/torchao/sparsity/marlin/__init__.py#L272) from numpy to torch. If I used torch.uint32, it would throw an error of `rshift_cpu not supported for uint32`. Using `torch.int64` seems to not cause any problems
2. I was able to completely remove the dependency from numpy, so we won't need to add it as a dependency

## Notes
* Closes #549 
* Since there is a deadline for the torchAO announcement tomorrow and, in the case that this work was to be part of it, I will be around today to work on any changes that this PR might need before it gets merged in.

Feel free to let me know if there is anything that needs to be refactored!

PR #733, opened 2024-08-22
Add sparse marlin 2:4 gemm op

Diogo-V:
## Description

This PR is a more concise version of #621 where only the gemm op and surrounding functions are implemented for a 2:4 [sparse marlin](https://github.com/IST-DASLab/Sparse-Marlin) kernel.

## What was done:
- [x] Tests to validate the gemm op as well as with `opcheck()` to check if `torch.compile` will work out of the box with it
- [x] Implemented functions to pack an int4 quantized tensor into a sparse marlin representation
- [x] Implemented functions to reverse the above process (to be later used when `dequantize()` is called)

## Notes:
* The cuda kernel was extracted from this [repo](https://github.com/neuralmagic/nm-vllm)

cc @jcaip

### DIFF
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -10,8 +10,9 @@ from torch.testing._internal.common_utils import (
     run_tests,
 )
 from torch.testing._internal.optests import opcheck
-from torchao.utils import is_fbcode, TORCH_VERSION_AT_LEAST_2_5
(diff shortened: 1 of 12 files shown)


jcaip (approved):
Thanks @Diogo-V!

msaroufim:
that was fast haha, 1 shot green ci ;)

Diogo-V:
Glad I could be of help!
Now, I have to maintain the streak ;)

PR #735, opened 2024-08-23
Fix docstring args names

kit1980:


msaroufim (approved):

jerryzh168:
@kit1980 we don't need to import to fbcode btw, we have a diff train that will handle the sync

jerryzh168 (approved):
thanks for the fix!

PR #736, opened 2024-08-23
Fix LR schedule handling for low-bit optimizers

### LINKED ISSUE #730: 4 bit Adam should support non constant lr
Our low bit optimizers were merged in HF https://github.com/huggingface/transformers/pull/31865 but

We have a known limitation that the 4 bit optimizer is not great when we don't have a constant learning rate

This is mentioned in the README https://github.com/pytorch/ao/tree/main/torchao/prototype/low_bit_optim

_Known issue: When learning rate is updated every step (e.g. using cosine learning rate scheduler), training speed is slower. This is because we have to convert learning rate to a CUDA tensor (which incurs expensive memory transfer cost), since torch.compile() will treat a Python float as a constant and trigger recompile whenever the value is changed._

However this is preventing @winglian from adopting this work

cc @gau-nernst @mlazos @janeyx99

gau-nernst:
Fixes #730

Tested with ViT-Base (~100M) params. Excellent speed up :smile:. The "before" is much slower than my previously reported numbers (3-4%) is probably because I use a much smaller model -> allocating CUDA LR tensor is much more evident.

image

Marking this as draft right now because I want to consolidate Adam and AdamW into a single file. It's becoming a bit troublesome to make the same edits in both files whenever I need to change anything :sweat_smile:.

cc @janeyx99 @msaroufim


msaroufim:
copy paste is mostly fine here to me (unless it really bothers you) even pytorch core has lots of copy paste between adam and adamw

gau-nernst:
@janeyx99 Since CPU scalar Tensor works with n-D CUDA Tensor, should `step` be on CPU also? I was following PyTorch core

https://github.com/pytorch/pytorch/blob/058302494cdf65fc59a1784fb9cea6c61e2be3f7/torch/optim/adamw.py#L99-L104

But incrementing `step` on CPU is definitely faster than doing it on CUDA?

@msaroufim It's just annoying when I make changes to the wrong file and need to be careful with the name - copy paste but still need to replace adam with adamw or vice versa :sweat_smile:

torchao/prototype/low_bit_optim/adam.py
```diff
@@ -20,7 +20,7 @@ def __init__(self, params, lr, betas, eps, weight_decay, amsgrad, *, block_size)
             raise ValueError("Invalid beta parameter at index 0: {}".format(betas[0]))
         if not 0.0 <= betas[1] < 1.0:
             raise ValueError("Invalid beta parameter at index 1: {}".format(betas[1]))
-        defaults = dict(lr=lr, betas=betas, eps=eps, weight_decay=weight_decay, amsgrad=amsgrad)
+        defaults = dict(lr=torch.tensor(lr), betas=betas, eps=eps, weight_decay=weight_decay, amsgrad=amsgrad)
```
janeyx99, line 23: is this the right place to set it? i would advocate for making this very obvious with users what is going on, and i would modify the AdamW constructor below.

janeyx99: unless you’re adamant that this optimizer will only support Tensor lrs. In pytorch/pytorch, we support python float lrs in eager because it is faster to compute python math than launch kernels, though that may be less relevant here.

janeyx99: The alternative I’d suggest for max visibility from user is changing line 165 below to be torch.tensor(1e-3)

janeyx99 (commented):
Direction is certainly what was discussed, so nice! Change looks mostly good to me. 

1. Regarding merging adam(w) together, there’s actually been an action item to merge the impls and make AdamW a subclass, so I understand exactly what you mean 🥲. That said, this change can def go in before you address that as this perf gap is more crucial to close.
2. Yes, step on CPU could be more performant (we normally have our steps be CPU tensors in core optims)
3. Logging/calling .item() esp on a CUDA tensor is expensive! just fyi

torchao/prototype/low_bit_optim/adam.py
gau-nernst: I think to keep it simple, we just enforce lr to be a scalar tensor here. If lr is not changed during training, whether it is a tensor or not would not matter. But if it is changed during training, we need lr to be a tensor anyway since torch.compile will recompile when python float lr changes value.

gau-nernst:
(I will update the benchmarks number in README. don't merge yet)

torchao/prototype/low_bit_optim/README.md
```diff
@@ -26,23 +26,23 @@ To use 4-bit Adam, replace the above with `Adam4bit`. Similarly for `AdamFp8`. Y
...
-ao 4-bit       |  7.72           |  7m 13s                  | 90.05
-lpmm 4-bit (*) |  7.73           | 11m 10s                  | 89.71
+Results for fine-tuning ViT-H (630M params) with BF16 AMP for 1 epoch, batch size 8, cosine LR scheduler, 4070Ti SUPER, fixed random seed:
+
+Adam impl       | max memory (GB) | imgs/s | accuracy
+----------------|-----------------|--------|----------
+PyTorch (fused) | 12.23           | 41.8   | 94.38
+bnb 8-bit       |  8.32           | 43.6   | 94.18
+ao 8-bit        |  8.33           | 42.6   | 94.25
+ao FP8 E4M3     |  9.27           | 44.1   | 94.40
```
gau-nernst, line 42: I re-ran the benchmark on main branch (without this PR) and the max memory for FP8 is the same. I'm suspecting something funny happening with torch.compile. The benchmark was done with `2.5.0.dev20240820`. I don't think it's a big issue, especially since FP8 optimizer is not popular yet (may change in the future though :eyes:). Re-running with 2.4 now. (probably won't re-run the rest with 2.4 since I'm lazy)

gau-nernst: FP8 optimizer

PyTorch version | max mem (GB) | imgs/s | acc
---|---|---|---
2.4.0 | 9.04 | 42.5 | 94.18
2.5.0.dev20240820 | 9.27 | 44.1 | 94.40

Definitely something funny with newer torch.compile :thinking:

torchao/prototype/low_bit_optim/adam.py
janeyx99: Yea, but I still think it’d be most visible/clear to users if the constructors of Adam and AdamW clearly set the default as Tensors and if this base would just error if lr was not a Tensor.

torchao/prototype/low_bit_optim/README.md
gau-nernst: On 2.3.1, Triton refused to run due to compute capability (though my GPU should support - 2.4.0 and nightly are fine)

```
Conversion from/to f8e4m3nv is only supported on compute capability >= 90Conversion from/to f8e4m3nv is only supported on compute capability >= 90

UNREACHABLE executed at ../../../lib/Conversion/TritonGPUToLLVM/ElementwiseOpToLLVM.cpp:823!
```

In the original benchmark numbers, FP8 optim was good (max memory is the same as 8-bit optim). But I don't rmb which PyTorch version I used back then 😅.

msaroufim (approved):

torchao/prototype/low_bit_optim/README.md
drisspg: This is just compiling the optimizer. There were some larger changes to the min cut partioner that affect how we split forward + bwd graphs ( determine what to recompute vs save).  These changes show up more in fp8 since you typically have long chains of ops to dequant or quant. But if there is no fwd/bwd I am not totally sure what might be happening

torchao/prototype/low_bit_optim/adam.py
janeyx99: Yes, I think it'd be good to have users pass in Tensor lrs, so they're most aware of what is going on. I think it is not great to switch it up under the hood and have the user be confused if there's ever an error regarding the Tensorness of the lr.

PR #740, opened 2024-08-23
Add Float8 Weight Only and FP8 weight + dynamic activation

drisspg:
# Summary

This PR makes some tweaks to the existing FP8 weight only work flow and adds float8_dynamic_activation_float8_weight to the quantize_ API. 

# New Apis made public:
```Python
    quantize_,
    float8_weight_only,
    float8_dynamic_activation_float8_weight,
)
```
```Python
def float8_dynamic_activation_float8_weight(
    target_dtype: torch.dtype = torch.float8_e4m3fn,
    activation_dtype: torch.dtype = torch.float8_e4m3fn,
    mm_config: ScaledMMConfig = ScaledMMConfig(use_fast_accum=True)
):
    """
    Applies float8 dynamic symmetric per-tensor quantization to both activations and weights of linear layers.
    Args:
        target_dtype (torch.dtype): The target data type for weight quantization. Default is torch.float8_e4m3fn.
        activation_dtype (torch.dtype): The target data type for activation quantization. Default is torch.float8_e4m3fn.
        mm_config (ScaledMMConfig): Configuration for the matrix multiplication. Default uses fast accumulation.
    """

```

``` Python
def float8_weight_only(target_dtype: torch.dtype = torch.float8_e4m3fn):
    """
    Applies float8 weight-only symmetric per-channel quantization to linear layers.
    
    Args:
        target_dtype (torch.dtype): The target data type for weight quantization. Default is torch.float8_e4m3fn.
    Note:
        The actual matmul will be computed in original precision of the weight tensor.
    """
```
### TODO
- Add Rowwise scaling option to top level api
- Next would be add float8_static_activation and hooking into calibration flow
- Adding these two APIs to autoquant
- Proper way of ensuring hardware support is met when quantizing for these types

### Changes
- Adds Float8Layout, the main reason for doing this is the ability to pass the mm_config down to matmul via the layout_type. This also adds the ability to do fake "transpose". This may still be needed but I have since learned that we register dispatch entires to nn.fucntional.linear. Thus the weight will not be ran through transpose. 
-  Wires up the TensorWise Scaling case to _scaled_mm
- Updates the Quant_primitives to handle the Symmetric no zero-point float to float dequant routine.
- Updates the AQTLayout's Typing to reflect the changes to dequant.
- I added various type hints and docs to function that I found helpful
- We were uncondtionally registering safe globals, gated that by behind the 2.5 flag
- Option for inverse scaling 

### Memory snapshot compare

With no_grad():

Screenshot 2024-08-28 at 11 46 16 PM

Normal conditions:
Screenshot 2024-08-28 at 11 47 04 PM

### FIX
@jerryzh168  If I decorate quantize affine:

```Python
@torch.no_grad()
def quantize_affine(
    input: torch.Tensor,
    block_size: Tuple[int, ...],
    scale: torch.Tensor,
    output_dtype: torch.dtype,
    quant_min: Optional[Union[int, float]] = None,
    quant_max: Optional[Union[int, float]] = None,
    zero_point_domain: ZeroPointDomain = ZeroPointDomain.INT,
) -> torch.Tensor:
    """
```
    
And 
```python
@torch.no_grad()
def choose_qparams_affine(
   input: torch.Tensor,
   mapping_type: MappingType,
   block_size: Tuple[int, ...],
   target_dtype: torch.dtype,
   quant_min: Optional[Union[int, float]] = None,
   quant_max: Optional[Union[int, float]] = None,
   eps: Optional[float] = None,
   scale_dtype: Optional[torch.dtype] = None,
   zero_point_dtype: Optional[torch.dtype] = None,
   preserve_zero: bool = True,
   zero_point_domain = ZeroPointDomain.INT,
) -> Tuple[torch.Tensor, torch.Tensor]:
```

I get the proper memory usage:
Screenshot 2024-08-28 at 11 57 28 PM

Is it okay If I land these as well or is there some other use case I am missing? I imagine all other input_tenosrs have been ints and havent propgated grads before

### Perf micro benchmarks
Runner Script:
https://gist.github.com/drisspg/3baf802f8c8631df2a549840f39e7e0d

Trace: [internal_link](https://interncache-all.fbcdn.net/manifold/perfetto-artifacts/tree/ui/index.html#!/?url=https://interncache-all.fbcdn.net/manifold/perfetto_internal_traces/tree/shared_trace/drisspg_88b2fa05-8715-4308-9a1f-b9ab23f7cb19_fp8_quant.json)

Max Autotune
Needs this fix: https://github.com/pytorch/pytorch/pull/134765
Trace: [internal_link](https://interncache-all.fbcdn.net/manifold/perfetto-artifacts/tree/ui/index.html#!/?url=https://interncache-all.fbcdn.net/manifold/perfetto_internal_traces/tree/shared_trace/drisspg_b5868ed6-806f-4118-b266-9228c2e9bb52_fp8_quant.json)

### DIFF
--- a/test/dtypes/test_affine_quantized.py
+++ b/test/dtypes/test_affine_quantized.py
@@ -10,11 +10,32 @@ from torchao.quantization.quant_api import (
     int8_dynamic_activation_int8_semi_sparse_weight,
     float8_weight_only,
 )
+from torch.testing._internal import common_utils
 from torchao.utils import TORCH_VERSION_AT_LEAST_2_5
 
 import torch
 import unittest
 import tempfile
+
+is_cuda_8_9 = torch.cuda.is_available() and torch.cuda.get_device_capability() >= (8, 9)
+
+
+def get_quantization_functions(do_sparse: bool, do_int4: bool):
+    base_functions = [
+        int8_weight_only(),
+        int8_dynamic_activation_int4_weight(),
+        int8_dynamic_activation_int8_weight(),
+    ]
+    if do_int4:
+        base_functions.append(int4_weight_only(group_size=32))
+
+    if do_sparse:
+        base_functions.append(int8_dynamic_activation_int8_semi_sparse_weight())
+
+    if is_cuda_8_9:
+        base_functions.append(float8_weight_only())
+
+    return base_functions
 
 
 class TestAffineQuantized(TestCase):
@@ -38,36 +59,36 @@ class TestAffineQuantized(TestCase):
             self.assertEqual(aqt_shape, shape)
 
     @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
-    def test_weights_only(self):
-        for apply_quant in [int4_weight_only(group_size=32), int8_weight_only(), int8_dynamic_activation_int4_weight(),
-                            int8_dynamic_activation_int8_weight(), int8_dynamic_activation_int8_semi_sparse_weight(), float8_weight_only()]:
-            l = torch.nn.Linear(128, 256, dtype=torch.bfloat16, device="cuda")
-            ql = apply_quant(l)
-            with tempfile.NamedTemporaryFile() as f:
-                torch.save(ql.state_dict(), f)
-                f.seek(0)
-                # `weights_only=True` is enabled for torch 2.5+
-                if TORCH_VERSION_AT_LEAST_2_5:
-                    _ = torch.load(f, weights_only=True)
-                else:
-                    _ = torch.load(f, weights_only=False)
+    @common_utils.parametrize("apply_quant", get_quantization_functions(True, True))
+    def test_weights_only(self, apply_quant):
+        l = torch.nn.Linear(128, 256, dtype=torch.bfloat16, device="cuda")
+        ql = apply_quant(l)
+        with tempfile.NamedTemporaryFile() as f:
+            torch.save(ql.state_dict(), f)
+            f.seek(0)
+            # `weights_only=True` is enabled for torch 2.5+
+            if TORCH_VERSION_AT_LEAST_2_5:
+                _ = torch.load(f, weights_only=True)
+            else:
+                _ = torch.load(f, weights_only=False)
 
     @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
-    def test_to_device(self):
-        from torchao.quantization import quantize_
-        for apply_quant in [int8_weight_only(), int8_dynamic_activation_int4_weight(), int8_dynamic_activation_int8_weight()]:
-            l = torch.nn.Linear(128, 256, dtype=torch.bfloat16)
-            ql = apply_quant(l)
-            ql.to("cuda")
+    @common_utils.parametrize("apply_quant", get_quantization_functions(False, False))
+    def test_to_device(self, apply_quant):
+        l = torch.nn.Linear(128, 256, dtype=torch.bfloat16)
+        ql = apply_quant(l)
+        ql.to("cuda")
 
-            l = torch.nn.Linear(128, 256, dtype=torch.bfloat16)
-            ql = apply_quant(l)
-            ql.to(device="cuda")
+        l = torch.nn.Linear(128, 256, dtype=torch.bfloat16)
+        ql = apply_quant(l)
+        ql.to(device="cuda")
 
-            l = torch.nn.Linear(128, 256, dtype=torch.bfloat16)
-            ql = apply_quant(l)
-            ql.cuda()
+        l = torch.nn.Linear(128, 256, dtype=torch.bfloat16)
+        ql = apply_quant(l)
+        ql.cuda()
 
+
+common_utils.instantiate_parametrized_tests(TestAffineQuantized)
 
 if __name__ == "__main__":
     run_tests()
--- a/test/dtypes/test_affine_quantized_float.py
+++ b/test/dtypes/test_affine_quantized_float.py
@@ -1,0 +1,141 @@
+from torchao.utils import (
+    TORCH_VERSION_AT_LEAST_2_5,
+    unwrap_tensor_subclass,
+)
+import pytest
+
+if not TORCH_VERSION_AT_LEAST_2_5:
+    pytest.skip("Unsupported PyTorch version", allow_module_level=True)
+
+from numpy import full
+from torch.testing._internal.common_utils import (
+    run_tests,
+)
+from torch._inductor.test_case import TestCase as InductorTestCase
+from torch.testing._internal import common_utils
+from torch.testing._internal.common_utils import (
+    TestCase,
+    run_tests,
+)
+from torch._dynamo.testing import CompileCounterWithBackend
+
+from torchao.quantization import (
+    quantize_,
+    float8_weight_only,
+    float8_dynamic_activation_float8_weight,
+)
+from torchao.float8.float8_utils import compute_error
+import torch
+import unittest
+import pytest
+import tempfile
+import copy
+import random
+
+from unittest.mock import patch
+
+
+random.seed(0)
+torch.manual_seed(0)
+
+is_H100 = torch.cuda.is_available() and torch.cuda.get_device_capability() >= (9, 0)
+is_cuda_8_9 = torch.cuda.is_available() and torch.cuda.get_device_capability() >= (8, 9)
+
+
+class ToyLinearModel(torch.nn.Module):
+    def __init__(self, in_features, out_features):
+        super().__init__()
+        self.linear1 = torch.nn.Linear(in_features, out_features, bias=False)
+        self.linear2 = torch.nn.Linear(out_features, in_features, bias=False)
+
+    def forward(self, x):
+        x = self.linear1(x)
+        x = self.linear2(x)
+        return x
+
+
+class TestAffineQuantizedFloat8Basic(TestCase):
+    @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
+    def test_tensor_core_layout_transpose(self):
+        l = torch.nn.Linear(128, 256, dtype=torch.bfloat16, device="cuda")
+        t = l.weight
+        shape = t.shape
+        apply_float8_weight_only_quant = float8_weight_only()
+        ql = apply_float8_weight_only_quant(l)
+        aqt = ql.weight
+        aqt_shape = aqt.shape
+        assert aqt_shape == shape
+
+        # transpose shape test
+        for _ in range(10):
+            t = t.t()
+            aqt = aqt.t()
+            shape = t.shape
+            aqt_shape = aqt.shape
+            assert aqt_shape == shape
+
+    @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
+    def test_weights_only_save_load(self):
+        with torch.no_grad():
+            for apply_quant in [float8_weight_only()]:
+                # TODO Fails when l requires grad
+                l = torch.nn.Linear(128, 256).eval().to(torch.bfloat16).to("cuda")
+                ql = apply_quant(l)
+                with tempfile.NamedTemporaryFile() as f:
+                    torch.save(ql.state_dict(), f)
+                    f.seek(0)
+                    # `weights_only=True` is enabled for torch 2.5+
+                    if TORCH_VERSION_AT_LEAST_2_5:
+                        _ = torch.load(f, weights_only=True)
+                    else:
+                        _ = torch.load(f, weights_only=False)
+
+
+class TestAffineQuantizedFloat8Compile(InductorTestCase):
+    @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
+    @unittest.skipIf(not is_cuda_8_9, "Need H100")
+    @common_utils.parametrize("dtype", [torch.bfloat16, torch.float32])
+    @common_utils.parametrize("mode", ["dynamic", "weight-only"])
+    @common_utils.parametrize("compile", [True, False])
+    # Inputs are (M,..), K, N
+    @common_utils.parametrize(
+        "sizes",
+        [
+            ((128,), 256, 128),
+            ((256,), 512, 256),
+            ((64,), 128, 64),
+            ((32, 128), 64, 256),
+            ((64, 256), 512, 128),
+        ],
+    )
+    def test_dynamic_fp8_linear(
+        self, dtype: torch.dtype, mode: str, compile: bool, sizes: tuple
+    ):
+        M, N, K = sizes
+        input_tensor = torch.randn(*M, K, dtype=dtype, device="cuda")
+
+        mode_map = {
+            "dynamic": float8_dynamic_activation_float8_weight,
+            "weight-only": float8_weight_only,
+        }
+
+        # Create a linear layer with bfloat16 dtype
+        model = ToyLinearModel(K, N).eval().to(dtype).to("cuda")
+
+        quantized_model = copy.deepcopy(model)
+        factory = mode_map[mode]()
+        quantize_(model, factory)
+
+        if compile:
+            quantized_model = torch.compile(quantized_model, fullgraph=True)
+
+        output_original = model(input_tensor)
+        output_quantized = quantized_model(input_tensor)
+
+        assert compute_error(output_original, output_quantized) > 20, "Error is too low"
+
+
+common_utils.instantiate_parametrized_tests(TestAffineQuantizedFloat8Compile)
+
+if __name__ == "__main__":
+    pytest.main([__file__])
--- a/test/dtypes/test_affine_quantized_float.py
+++ b/test/dtypes/test_affine_quantized_float.py
@@ -13,10 +13,6 @@ from torch.testing._internal.common_utils import (
 )
 from torch._inductor.test_case import TestCase as InductorTestCase
 from torch.testing._internal import common_utils
-from torch.testing._internal.common_utils import (
-    TestCase,
-    run_tests,
-)
 from torch._dynamo.testing import CompileCounterWithBackend
 
 from torchao.quantization import (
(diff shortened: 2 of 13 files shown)


jerryzh168:
@drisspg yeah the `no_grad` for quantize_affine/dequantize_affine should be OK I think. fake_quantize ops are reusing the same helper functions that's called by these ops but since the no_grad is not on the helper functions so it probably won't affect QAT, so as long as https://github.com/pytorch/ao/blob/cfabc13e72fd03934e62a2a03903bc1678235bed/test/quantization/test_qat.py#L436 passes it should be fine

test/dtypes/test_affine_quantized_float.py
```diff
@@ -0,0 +1,140 @@
...
+                # `weights_only=True` is enabled for torch 2.5+
+                if TORCH_VERSION_AT_LEAST_2_5:
+                    _ = torch.load(f, weights_only=True)
+                else:
+                    _ = torch.load(f, weights_only=False)
+
+
+class TestAffineQuantizedFloat8Compile(InductorTestCase):
+    @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
+    @unittest.skipIf(not is_cuda_8_9, "Need H100")
```
jerryzh168, line 95: we don't have H100 in CI I think?

test/dtypes/test_affine_quantized_float.py
```diff
@@ -0,0 +1,140 @@
...
+        factory = mode_map[mode]()
+        quantize_(model, factory)
+
+        if compile:
+            quantized_model = torch.compile(quantized_model, fullgraph=True)
+
+        output_original = model(input_tensor)
+        output_quantized = quantized_model(input_tensor)
+
+        assert compute_error(output_original, output_quantized) > 20, "Error is too low"
```
jerryzh168, line 134: nit: too high?

test/dtypes/test_affine_quantized_float.py
drisspg: Yeah need to fix this error message, since it we do have l40s which this is being ran on

torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -44,7 +47,13 @@ class AQTLayout(TorchAOBaseTensor):
     """
     Base class for the layout tensor for `AffineQuantizedTensor`
     """
-    def get_plain(self) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+    def get_plain(self) -> Tuple[torch.Tensor, torch.Tensor, Optional[torch.Tensor]]:
+        """ This method will be called when dequantizing the Tensor
+        It should return a tuple 3 things:
+            * The underlying data
```
jerryzh168, line 53: nit: we can add "unpacked" here as well I think

jerryzh168: also does optional zero point work? I thought None doesn't work in torch.compile

torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -270,9 +290,11 @@ def from_hp_to_floatx(
         cls,
         input_float: torch.Tensor,
         block_size: Tuple[int, ...],
-        target_dtype: torch.dtype = torch.float8_e4m3fn,
+        target_dtype: torch.dtype,
+        scale_dtype: Optional[torch.dtype] = None,
         layout_type: LayoutType = PlainLayoutType(),
```
jerryzh168, line 295: will fp8 work with `PlainLayoutType` or will it always be `Float8LayoutType`?

torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -376,6 +398,13 @@ def extra_repr(self):
         return f"inner_k_tiles={self.inner_k_tiles}"
 
 
+@dataclass(frozen=True)
+class Float8LayoutType(LayoutType):
+    mm_config: ScaledMMConfig
+
+    def pre_process(self, input: torch.Tensor) -> torch.Tensor:
+        return input
```
jerryzh168, line 406: this can be removed since it's the default implementation

torchao/dtypes/affine_quantized_tensor.py
drisspg: Yeah let me update the defaults

test/dtypes/test_affine_quantized_float.py
```diff
@@ -0,0 +1,140 @@
...
+        "sizes",
+        [
+            ((128,), 256, 128),
+            ((256,), 512, 256),
+            ((64,), 128, 64),
+            ((32, 128), 64, 256),
+            ((64, 256), 512, 128),
+        ],
+    )
+    def test_dynamic_fp8_linear(
```
jerryzh168, line 110: nit: name

torchao/dtypes/affine_quantized_tensor.py
drisspg: yeah compile seems to be working as expected here

torchao/dtypes/utils.py
```diff
@@ -143,3 +143,15 @@ def _get_layout_tensor_constructor(cls: Callable, layout_type_class: type(Layout
 
 def is_device(target_device_str: str, device: Union[str, torch.device]):
     return torch.device(device).type == target_device_str
+
+def get_out_shape(input_shape: Tuple[int], weight_shape: Tuple[int]) -> Tuple[int, int]:
```
jerryzh168, line 147: why is this related to weight_shape? I thought this is just something like `tmp = x_vals_int8.reshape(-1, x_vals_int8.shape[-1])`

torchao/quantization/quant_api.py
```diff
@@ -489,19 +492,77 @@ def int8_dynamic_activation_int8_semi_sparse_weight():
...
+            block_size=block_size,
+            target_dtype=target_dtype,
+            layout_type=Float8LayoutType(mm_config=None),
+        )
 
     return _get_linear_subclass_inserter(apply_float8wo_quant)
 
 
+def float8_dynamic_activation_float8_weight(
+    target_dtype: torch.dtype = torch.float8_e4m3fn,
```
jerryzh168, line 522: weight_dtype?

test/dtypes/test_affine_quantized_float.py
```diff
@@ -0,0 +1,140 @@
...
+        self.linear1 = torch.nn.Linear(in_features, out_features, bias=False)
+        self.linear2 = torch.nn.Linear(out_features, in_features, bias=False)
+
+    def forward(self, x):
+        x = self.linear1(x)
+        x = self.linear2(x)
+        return x
+
+
+class TestAffineQuantizedFloat8Basic(TestCase):
```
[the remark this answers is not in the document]
drisspg: I think this is a rebase bug, I see this test already in afifne quantized

torchao/quantization/quant_primitives.py
```diff
@@ -267,6 +267,11 @@ def _quantize_affine_no_dtype_cast(
     if zero_point is not None:
         zero_point = zero_point.view(shape_after_reduction)
 
+    if zero_point_domain is None:
```
jerryzh168, line 270: would it be better to check `zero_point` here? it seems that we want to say if there is no zero point, maybe zero_point_domain is None can be an assert inside the branch

torchao/quantization/quant_primitives.py
```diff
@@ -267,6 +267,11 @@ def _quantize_affine_no_dtype_cast(
...
+       quant = torch.clamp(input * scale.reciprocal(), quant_min, quant_max)
```
jerryzh168, line 271: how do we know if we should do `input * scale.reciprocal()` v.s. `input / scale` v.s. `input * (1.0 / scale)`...

torchao/quantization/quant_primitives.py
```diff
@@ -367,19 +372,29 @@ def _dequantize_affine_no_dtype_check(
...
     original_shape = input.shape
     input = input.view(shape_for_reduction)
     shape_after_reduction = shape_for_reduction
     for i in reduction_dims:
         shape_after_reduction[i] = 1
-    scale = scale.view(shape_after_reduction)
+    scale = scale.view(shape_after_reduction)   
+
+    # This case handles dequantization for float8
+    if zero_point_domain is None:
```
jerryzh168, line 391: similar comment for this one, I feel it probably makes more sense to use `zero_point` here

jerryzh168: although using `zero_point_domain` has one benefit, that is you can merge this implementation into the branch starting from L401, so I feel that makes sense as well, we just need to document it properly

torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -529,6 +558,110 @@ def from_plain(
         return cls(int_data_compressed, scale, zero_point, layout_type)
 
 
+@register_layout_cls(Float8LayoutType)
+class Float8AQTLayout(AQTLayout):
```
vkuzo, line 562: on mobile, but its unexpected to see this. I thought layout was representing how data is laid out in memory, in which case plain layout sounds right to me.

torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -529,6 +558,110 @@ def from_plain(
...
+    @classmethod
+    def from_plain(
+        cls,
+        data: torch.Tensor,
+        scale: torch.Tensor,
+        zero_point: Optional[torch.Tensor],
+        layout_type: LayoutType,
+    ):
+        """ Main entrypoint for constructing Float8Layout Tensor"""
+        assert _is_float8_type(data.dtype), f"Float8 Layout must be constructed from float8 dtype but got {data.dtype}"
```
jerryzh168, line 652: maybe add an assert for LayoutType as well if it only works with `Float8LayoutType`

torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -892,10 +1026,68 @@ def _linear_fp_act_int8_weight_impl(input_tensor, weight_tensor, bias):
...
 
+def _linear_fp_act_fp8_tensor_wise_weight_check(
+    input_tensor: torch.Tensor,
+    weight_tensor: AffineQuantizedTensor,
+    bias: Optional[torch.Tensor],
+) -> bool:
+    def check_aqt_tensorwise(aqt: AffineQuantizedTensor) -> bool:
+        return (
+            isinstance(aqt, AffineQuantizedTensor) and
+            isinstance(aqt.layout_tensor, Float8AQTLayout)
```
jerryzh168, line 1037: we typically check `aqt.layout_type` instead

test/dtypes/test_affine_quantized_float.py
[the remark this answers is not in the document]
drisspg: moved over

torchao/dtypes/affine_quantized_tensor.py
drisspg: actually I dont know what you mean by "unpacked"

torchao/quantization/quant_api.py
drisspg: agreed

torchao/quantization/quant_api.py
```diff
@@ -489,19 +492,77 @@ def int8_dynamic_activation_int8_semi_sparse_weight():
...
+            layout_type=Float8LayoutType(mm_config=None),
+        )
+
+        def input_quant_func(x: torch.Tensor):
+            activation = to_affine_quantized_floatx(
+                input_float=x,
+                block_size=x.shape,
+                target_dtype=activation_dtype,
+                scale_dtype=torch.float32,
+                layout_type=Float8LayoutType(mm_config=None),
```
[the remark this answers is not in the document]
drisspg: yup thanks, just realized that as well

torchao/quantization/quant_primitives.py
drisspg: Way back when Nikita told me that this is the fastest way to do it, I think we should update the others as well but probs in a separate PR since this is already pretty sprawling

torchao/dtypes/affine_quantized_tensor.py
drisspg: I left a comment about the need for this in the summary:
```
Adds Float8Layout, the main reason for doing this is the ability to pass the mm_config down to matmul via the layout_type. This also adds the ability to do fake "transpose". This may still be needed but I have since learned that we register dispatch entires to nn.fucntional.linear. Thus the weight will not be ran through transpose.
```
@jerryzh168 If this isnt true correct me though please

vkuzo: right, I'm saying it seems like this isnt using layout as intended? Do the abstractions need to change instead of tacking something like this on?

vkuzo: or, rename it so its not just about layout, etc

torchao/dtypes/affine_quantized_tensor.py
drisspg: I had one, but since from_plain is only ever called via the registration dispatch table I didnt think that was necessary, I can add though

torchao/dtypes/affine_quantized_tensor.py
drisspg: Is that okay for a follow up PR?

jerryzh168: layout is an abstraction to track all packing/format related things, this includes the behavior of transpose and how the format behaves in matmul I think

test/dtypes/test_affine_quantized.py
```diff
@@ -10,12 +10,32 @@
...
+        int8_weight_only(),
+        int8_dynamic_activation_int4_weight(),
+        int8_dynamic_activation_int8_weight(),
+    ]
+    if do_int4:
+        base_functions.append(int4_weight_only(group_size=32))
+    if do_sparse:
+        base_functions.append(int8_dynamic_activation_int8_semi_sparse_weight())
+
+    if is_cuda_8_9 and do_float8:  # You need to define this function
```
jerryzh168, line 34: what does the comment mean? also where is `do_float8`?

torchao/dtypes/affine_quantized_tensor.py
jerryzh168: oh it means data is not optimized for storage or computation, e.g. uint8 + quant_min=0, quant_max=15, will be the unpacked representation for uint4, when we pack it we can put 2 elements of uint4 data into one uint8 element.

torchao/quantization/quant_api.py
```diff
@@ -489,19 +492,80 @@ def int8_dynamic_activation_int8_semi_sparse_weight():
...
+    activation_dtype: torch.dtype = torch.float8_e4m3fn,
+    weight_dtype: torch.dtype = torch.float8_e4m3fn,
+    mm_config: Optional[ScaledMMConfig] = None
+):
+    """
+    Applies float8 dynamic symmetric per-tensor quantization to both activations and weights of linear layers.
+
+    Args:
+        target_dtype (torch.dtype): The target data type for weight quantization. Default is torch.float8_e4m3fn.
+        activation_dtype (torch.dtype): The target data type for activation quantization. Default is torch.float8_e4m3fn.
```
jerryzh168, line 532: please update the docs to align with the arg name as well

torchao/quantization/quant_api.py
```diff
@@ -489,19 +492,80 @@ def int8_dynamic_activation_int8_semi_sparse_weight():
...
+            layout_type=Float8LayoutType(mm_config=mm_config),
+        )
+
+        def input_quant_func(x: torch.Tensor):
+            activation = to_affine_quantized_floatx(
+                input_float=x,
+                block_size=x.shape,
+                target_dtype=activation_dtype,
+                scale_dtype=torch.float32,
+                layout_type=Float8LayoutType(mm_config=mm_config),
```
jerryzh168, line 557: do we want both to take the same mm_config?

torchao/quantization/quant_primitives.py
```diff
@@ -222,9 +222,9 @@ def _quantize_affine(
...
     zero_point: Optional[torch.Tensor],
     output_dtype: torch.dtype,
-    quant_min: Optional[int] = None,
-    quant_max: Optional[int] = None,
-    zero_point_domain: str = ZeroPointDomain.INT.name,
+    quant_min: Optional[Union[int, float, bool]] = None,
+    quant_max: Optional[Union[int, float, bool]] = None,
+    zero_point_domain: Optional[str] = ZeroPointDomain.INT.name,
 ) -> torch.Tensor:
     """op definition that has compatible signatures with custom op library
```
jerryzh168, line 229: maybe add a note for zero_point_domain here would be helpful I think

torchao/quantization/quant_api.py
```diff
@@ -489,19 +492,80 @@ def int8_dynamic_activation_int8_semi_sparse_weight():
     """
     return int8_dynamic_activation_int8_weight(layout_type=SemiSparseLayoutType())
 
+
 def float8_weight_only(target_dtype: torch.dtype = torch.float8_e4m3fn):
     """
     Applies float8 weight-only symmetric per-channel quantization to linear layers.
+    
+    Args:
+        target_dtype (torch.dtype): The target data type for weight quantization. Default is torch.float8_e4m3fn.
```
jerryzh168, line 501: nit: might make sense to use `weight_dtype` as well to align with `float8_dynamic_activation_float8_weight`

jerryzh168 (approved):
looks good overall, just had some nit comments inline

torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -217,6 +237,8 @@ def from_hp_to_intx(
             data = data.to(target_dtype)
         else:
             scale, zero_point = choose_qparams_affine(input_float, mapping_type, block_size, target_dtype, quant_min, quant_max, eps, scale_dtype, zero_point_dtype, preserve_zero, zero_point_domain)
+            if zero_point_domain is None:
+                zero_point = None
```
jerryzh168, line 241: this feels a bit confusing, should choose_qparams_affine return None for zero_point directly?

torchao/quantization/quant_primitives.py
```diff
@@ -159,15 +159,15 @@ def _get_reduction_params(block_size, input_size):
...
     input: torch.Tensor,
     block_size: Tuple[int, ...],
     scale: torch.Tensor,
     zero_point: Optional[torch.Tensor],
     output_dtype: torch.dtype,
-    quant_min: Optional[int] = None,
-    quant_max: Optional[int] = None,
+    quant_min: Optional[Union[int, float]] = None,
+    quant_max: Optional[Union[int, float]] = None,
     zero_point_domain: ZeroPointDomain = ZeroPointDomain.INT,
```
jerryzh168, line 171: I think you'll need to update this type to be Optional[ZeroPointDomain] as well right?

jainapurva: Currently infer_schema in torch doesn't support Optional[Union[int, float]]

jerryzh168: you mean the export test will fail?

test/dtypes/test_affine_quantized.py
drisspg: junk comment

torchao/quantization/quant_api.py
drisspg: ahh good call I will make this none, since we are storing config on weight

torchao/quantization/quant_api.py
drisspg: yeah will do

torchao/dtypes/affine_quantized_tensor.py
drisspg: yeah thats better, I got tunnel vision

drisspg: Actually can confirm I cant return an optional tensor from there

torchao/quantization/quant_primitives.py
drisspg: can you point me to the test?

torchao/float8/float8_python_api.py
```diff
@@ -30,14 +30,19 @@ def addmm_float8_unwrapped(
     output_scale: Optional[torch.Tensor] = None,
     bias: Optional[torch.Tensor] = None,
     use_fast_accum: bool = False,
+    inverse_scale: bool = True
```
vkuzo, line 33: does this mean `AffineQuantizedTensor` uses scale in the same way `Float8Tensor` uses 1 / scale, or something else?

torchao/quantization/quant_api.py
```diff
@@ -57,6 +59,7 @@
 from .utils import _get_per_token_block_size
 import logging
 from .autoquant import autoquant, AutoQuantizableLinearWeight
+from torchao.float8.float8_tensor import ScaledMMConfig
```
vkuzo, line 62: it seems like this PR makes `ScaledMMConfig` a public API.  Is this intended?  If yes, can we move this object to `float8/__init__.py`, maybe make a dataclass, ensure it's documented, etc?

drisspg: yeah Ill put up a PR

PR #746, opened 2024-08-25
[Low-bit optim] Add Llama2-7B finetune benchmarks

gau-nernst:
Update: change Llama3.1-8B-instruct to Llama2-7B

Fine-tune [Llama2-7B](https://huggingface.co/meta-llama/Llama-2-7b) on [Alpaca](https://huggingface.co/datasets/tatsu-lab/alpaca) dataset. Full BF16, 1 epoch, A100, fixed random seed. Benchmark is done with [torchtune](https://github.com/pytorch/torchtune). 

## Summary

AdamW impl       | Max memory (GB) | toks/s | `truthfulqa_mc2` acc | Compile time
-----------------|-----------------|--------|---------------------|-----
Not fine-tuned   | -               | -      | 38.95 | -
PyTorch (fused)  | 52 | ~4500 | 42.12 | ~4 min
bnb 8-bit        | 39 | ~4000 | 41.98 | ~4 min
ao 8-bit         | 39 | ~4000 | 42.41 | ~12 min
ao 4-bit         | 33 | ~3600 | 42.34 | ~4 min

NOTE:
- lpmm's 4-bit AdamW does not support BF16 weights -> not include in benchmark
- A100 does not support FP8 -> not include FP8 AdamW

Observations
- The reduction in peak memory looks correct: going from 16-bit to 8-bit -> 52 - 39 = 13GB reduction. Going from 8-bit to 4-bit -> 39 - 33 = 6GB reduction.
- Our 8-bit AdamW is only slightly slower than bnb, which is nice.
- **The compile time for 8-bit is HUGE** (12 min). Might need to find ways to mitigate this.
- Our 4-bit AdamW is quite slow, but compile fast. This is expected because we compile w/ dynamic shape for each param, while in 8-bit AdamW, we compile w/ static shape for all params. We do this way because 4-bit AdamW will have a memory bug when compiling for all params.

Command used (change `optimizer` and `checkpointer.output_dir` across runs)

> tune run full_finetune_single_device --config llama2/7B_full optimizer=torch.optim.AdamW optimizer.fused=True optimizer_in_bwd=False compile=True metric_logger=torchtune.utils.metric_logging.WandBLogger log_peak_memory_stats=True batch_size=16 log_every_n_steps=10 epochs=1 seed=2024 checkpointer.output_dir=experiments/adamw_baseline

## Fancy graphs!

**Compare across different n-bit optimizers**

image

**Compare 8-bit AdamW between ao and bnb**. The fact that the two graphs overlap show that our implementation is correct and competitive in speed (except compile time 😭)!

image


msaroufim (approved):
cc @mlazos who was looking at large compile times

gau-nernst:
@msaroufim Any blockers to merge this? The failing CPU test is unrelated, though I'm probably in charge of it since it's FP6-LLM 🌚. Seems like something change with CPU inductor.

Some thoughts on reducing compile time. There are 2 approaches to compile optimizer step in low-bit optim:
1. Compile optim step for single param i.e. `torch.compile(single_param_adam)`
2. Compile optim step for all params i.e. `torch.compile(param_groups_adam)`

Currently Adam8bit and AdamFp8 use approach (2) (with static shape) since it is faster (but compile much slower), while Adam4bit uses approach (1) (with dynamic shape) since there are excessive memory usage for "Adam4bit + approach (2)". Approach (1) requires dynamic shape to avoid hitting recompiles limit.

Now looking back, perhaps we can do approach (1) with static shape + temporarily remove recompile limit? I have seen FlexAttention doing this

https://github.com/pytorch/pytorch/blob/76710d4f95d1f920bdf56e4db4d6d71ef6c9aea2/torch/nn/attention/flex_attention.py#L989

It's probably safe to do so, since for a given model, the number of recompiles for `single_param_adam()` is fixed, though some models may have more recompiles than others (e.g. ViT vs LLM).

msaroufim:
I'm gonna add some of your comments here to the README since they're helpful

PR #748, opened 2024-08-26
Add INT8 mixed-precision training

gau-nernst:
Disclaimer: terminologies for INT8 training are generally not standardized yet. Thus, I use the terms with the following meaning:
- Quantized training: model weights are quantized. This is a strict requirement. Does not matter what is the compute precision. Examples of this: Q-GaLore, JetFire.
- INT8 mixed-precision training: model weights are in original precision, while some compute dtype might be INT8. I call it like this because it is similar to FP16/BF16 mixed-precision. One difference is that in FP16/BF16 mixed-precision, matmul will return FP16/BF16 outputs, while for INT8 mixed-precision, the returned dtype is not INT8. Examples include Google AQT and SwitchBack.

(WIP, will update PR description)

### DIFF
--- a/torchao/prototype/quantized_training/README.md
+++ b/torchao/prototype/quantized_training/README.md
@@ -2,7 +2,7 @@
 
 This folder contains experimental work on quantized training (QT). The main difference from quantization-aware training (QAT) is that in QT, we don't keep a high-precision copy of model weights. We take inspirations from:
 - Q-GaLore: [[paper](https://arxiv.org/abs/2407.08296)] [[code](https://github.com/VITA-Group/Q-GaLore)]
-- AQT: [[related paper](https://arxiv.org/abs/2105.03536)] [[code](https://github.com/google/aqt)]
+- JetFire: [[paper](https://arxiv.org/abs/2403.12422)] [[code](https://github.com/thu-ml/Jetfire-INT8Training)]
 
 Typically, low-precision weights cannot be trained directly due to quantization error: a small change in the quantized weight will be round down to zero. To tackle this problem, we use **stochastic rounding** for weight update. In simple terms, stochastic rounding will round up or down randomly, but with a higher chance if it is closer to that direction. For example, 0.8 will have 80% chance of rounding up and 20% of rounding down. It also follows that on average, stochastic rounding will estimate the floating point value exactly.
 
@@ -23,7 +23,7 @@ Usage
 ```python
 from torchao.prototype.quantized_training import int8_weight_only_quantized_training
 from torchao.prototype.low_bit_optim import _AdamW
-from torchao.quantization.quant_api import quantize_
+from torchao.quantization import quantize_
 
 model = ...
 quantize_(model, int8_weight_only_quantized_training())
@@ -46,8 +46,46 @@ BF16 compile    | 10.16915
 INT8 QT eager   | 10.11437
 INT8 QT compile | 10.03365
 
+## INT8 mixed-precision
+
+On NVIDIA GPUs, INT8 Tensor Cores can be up to 3x faster than their BF16/FP16 counterparts. In mixed-precision training, we can down-cast activations and weights dynamically to INT8 to leverage faster matmuls. However, since INT8 has very limited range [-128,127], we perform row-wise quantization, similar to how INT8 post-training quantization (PTQ) is done. Weight is still in original precision. This is inspired from prior works:
+
+- AQT: [[related paper](https://arxiv.org/abs/2105.03536)] [[code](https://github.com/google/aqt)]
+- SwitchBack: [[paper](https://arxiv.org/abs/2304.13013)]
+
+Usage
+
+```python
+from torchao.prototype.quantized_training import int8_mixed_precision_training, Int8MixedPrecisionConfig
+from torchao.quantization import quantize_
+
+model = ...
+config = Int8MixedPrecisionConfig(
+    forward=True,
+    backward_grad_input=True,
+    backward_grad_weight=True,
+)
+quantize_(model, int8_mixed_precision_training(config))
+
+# train model as usual
+```
+
+During training, there are 3 matmuls involved in each `nn.Linear` layer:
+- 1 in forward: `output = input @ weight.T`
+- 2 in backward:
+  - `grad_input = grad_output @ weight`
+  - `grad_weight = grad_output.T @ input`
+
+You can configure which matmul to be applied with INT8 mixed-precision using `Int8MixedPrecisionConfig` shown above. If convergence is an issue, we recommend leaving `backward_grad_weight` in original matmul precision, and also `backward_grad_input` if the issue still persists.
+
+Note:
+- When we only apply INT8 mixed-precision in the forward pass, this can be considered QAT.
+- When we only apply INT8 mixed-precision to `forward` and `backward_grad_input`, this is similar to SwitchBack. However, SwitchBack uses tensor-wise scaling for weight. For simplicity, we only support row-wise scaling.
+
+TODO: add some benchmarks
+
 ## Future ideas
 
-- INT8 activation x INT8 weight. This can potentially leverage INT8 Tensor Cores, which is 2x faster than FP16/BF16 Tensor Cores.
+- Tile-wise INT8 quantization to keep quantized weight for both forward and backward pass (similar to JetFire).
 - INT4 weight only (with group-wise quantization). This can be used with INT4 tinygemm deployment in mind (or other optimized INT4 kernels).
 - FP8 activation x FP8 weight. The current FP8 training recipe can be seen as a form of QAT, which maintains a high-precision copy of model weights. We can eliminate the high-precision copy.
--- a/torchao/prototype/quantized_training/int8.py
+++ b/torchao/prototype/quantized_training/int8.py
@@ -12,13 +12,41 @@ c10d_functional = torch.ops.c10d_functional
 _c10d_functional = torch.ops._c10d_functional
 
 
+@torch.no_grad()
+def quantize_int8_rowwise(tensor: Tensor, stochastic_rounding: bool = False):
+    """Normal rounding will always round down small changes in weight update. To tackle this problem,
+    stochastic rounding can be used, which has a low chance, but not zero, of rounding up. The
+    probability of rounding up is equal to x - ⌊x⌋, which indicates how close the value is to the next
+    integer value. Thus, stochastic rounding also approximates the floating point value exactly.
+
+    Currently this function differs from AQT's `int8_weight_only()` in the following way:
+    1. Precision: AQT keeps original dtype when doing quantization, while this function upcasts input
+    to FP32 before quantization. Output scale maintains the original input dtype.
+    2. Calculate scale: AQT uses `input.abs().amax() / 127.5`, while `input.abs().amax() / 127` is
+    done here.
+    3. Apply scale: AQT uses `input * (1 / scale)`, while this function performs `input / scale`.
+    """
+    # absmax symmetric quantization
+    scale = tensor.abs().amax(1) / 127  # same dtype as tensor
+    inv_scale = 1.0 / scale.float().clip(1e-12)
+    tensor = tensor.float() * inv_scale.view(-1, 1)  # slightly faster than divide directly
+
+    if stochastic_rounding:
+        tensor = (tensor + torch.rand_like(tensor)).floor()
+    else:
+        tensor = tensor.round()
+
+    tensor = tensor.clip(-128, 127).to(torch.int8)
+    return tensor, scale
+
+
 class Int8QTLinearWeight(Tensor):
     """INT8 symmetric quantization weight, with absmax scaling [-127, 127]. The main difference
     of this tensor subclass from AffineQuantizedTensor:
     1. `F.linear` is differentiable i.e. backward is defined.
     2. All in-place ops, such as `aten.copy_`, will perform stochastic rounding.
         `Int8QTLinearWeight.from_float()` does not perform stochastic rounding.
-    3. The numerics for quantization is slightly different. See `Int8QTLinearWeight.quantize()`
+    3. The numerics for quantization is slightly different. See `quantize_int8_rowwise()`
         for more details.
     """
 
@@ -55,42 +83,12 @@ class Int8QTLinearWeight(Tensor):
     def __tensor_unflatten__(cls, tensor_data_dict, tensor_attributes, outer_size=None, outer_stride=None):
         return cls(tensor_data_dict["int_data"], tensor_data_dict["scale"], *tensor_attributes)
 
-    @staticmethod
-    @torch.no_grad()
-    def quantize(tensor: Tensor, stochastic_rounding: bool = False):
-        """Normal rounding will always round down small changes in weight update. To tackle this problem,
-        stochastic rounding can be used, which has a low chance, but not zero, of rounding up. The
-        probability of rounding up is equal to x - ⌊x⌋, which indicates how close the value is to the next
-        integer value. Thus, stochastic rounding also approximates the floating point value exactly.
-
-        Currently this function differs from AQT's `int8_weight_only()` in the following way:
-        1. Precision: AQT keeps original dtype when doing quantization, while this function upcasts input
-        to FP32 before quantization, and downcast scale to original dtype.
-        2. Calculate scale: AQT uses `input.abs().amax() / 127.5`, while `input.abs().amax() / 127` is
-        done here.
-        3. Apply scale: AQT uses `input * (1 / scale)`, while this function performs `input / scale`.
-        """
-        original_dtype = tensor.dtype
-        tensor = tensor.float()
-
-        # absmax symmetric quantization
-        scale = tensor.abs().amax(-1) / 127
-        tensor = tensor / scale.clip(1e-12).view(-1, 1)
-
-        if stochastic_rounding:
-            tensor = (tensor + torch.rand_like(tensor)).floor()
-        else:
-            tensor = tensor.round()
-
-        tensor = tensor.clip(-128, 127).to(torch.int8)
-        return tensor, scale.to(original_dtype)
-
     @classmethod
     def from_float(cls, tensor: Tensor):
         """Convert a float tensor into INT8 quantized weight. No stochastic rounding is performed.
         This function is not differentiable.
         """
-        int_data, scale = cls.quantize(tensor.detach())
+        int_data, scale = quantize_int8_rowwise(tensor.detach())
         out = cls(int_data, scale)
         out.requires_grad_(tensor.requires_grad)
         return out
@@ -201,7 +199,7 @@ def _(func, types, args, kwargs):
         args[0].scale.copy_(args[1].scale, **kwargs)
 
     elif isinstance(args[0], Int8QTLinearWeight):
-        int_data, scale = Int8QTLinearWeight.quantize(args[1], stochastic_rounding=True)
+        int_data, scale = quantize_int8_rowwise(args[1], stochastic_rounding=True)
         args[0].int_data.copy_(int_data, **kwargs)
         args[0].scale.copy_(scale, **kwargs)
 
--- a/torchao/prototype/quantized_training/int8_mixed_precision.py
+++ b/torchao/prototype/quantized_training/int8_mixed_precision.py
@@ -1,0 +1,125 @@
+from typing import NamedTuple
+
+import torch
+import torch.nn.functional as F
+from torch import Tensor
+from torch.utils._triton import has_triton
+
+from .int8 import quantize_int8_rowwise
+
+if has_triton():
+    from .int8_mm import int8_mm_dequant
+
+else:
+
+    def int8_mm_dequant(A: Tensor, B: Tensor, A_scale_rowwise: Tensor, B_scale_colwise: Tensor) -> Tensor:
+        return (A * A_scale_rowwise.view(-1, 1)) @ (B * B_scale_colwise.view(1, -1))
+
+
+aten = torch.ops.aten
+
+
+class Int8MixedPrecisionConfig(NamedTuple):
+    forward: bool = False
+    backward_grad_input: bool = False
+    backward_grad_weight: bool = False
+
+
+class Int8MixedPrecisionLinearWeight(Tensor):
+    @staticmethod
+    @torch._dynamo.disable
+    def __new__(cls, data: Tensor, config: Int8MixedPrecisionConfig):
+        return Tensor._make_wrapper_subclass(
+            cls,
+            data.shape,
+            dtype=data.dtype,
+            device=data.device,
+        )
+
+    @torch._dynamo.disable
+    def __init__(self, data: Tensor, config: Int8MixedPrecisionConfig):
+        self._data = data
+        self.config = config
+
+    def __tensor_flatten__(self):
+        return ["_data"], [self.config]
+
+    @classmethod
+    def __tensor_unflatten__(cls, tensor_data_dict, tensor_attributes, outer_size=None, outer_stride=None):
+        return cls(tensor_data_dict["_data"], *tensor_attributes)
+
+    def __repr__(self):
+        return self._data.__repr__()
+
+    @classmethod
+    def __torch_function__(cls, func, types, args=(), kwargs=None):
+        kwargs = kwargs or dict()
+
+        if func is F.linear:
+            return _Int8MixedPrecisionLinear.apply(*args, **kwargs)
+
+        with torch._C.DisableTorchFunctionSubclass():
+            return func(*args, **kwargs)
+
+    @classmethod
+    def __torch_dispatch__(cls, func, types, args, kwargs):
+        if func in (aten.detach.default, aten.clone.default, aten._to_copy.default):
+            return cls(func(args[0]._data, *args[1:], **kwargs), args[0].config)
+
+        # TODO: some ops should return the original class i.e. in-place ops
+        args = [x._data if isinstance(x, cls) else x for x in args]
+        return func(*args, **kwargs)
+
+
+class _Int8MixedPrecisionLinear(torch.autograd.Function):
+    @staticmethod
+    def forward(ctx, input: Tensor, weight: Int8MixedPrecisionLinearWeight, bias: Tensor | None = None):
+        ctx.save_for_backward(input, weight)
+        ctx.bias = bias is not None
+
+        if weight.config.forward:
+            batch_dims = input.shape[:-1]
+            input = input.view(-1, weight.shape[1])
+            input_i8, input_scale = quantize_int8_rowwise(input)
+            weight_i8, weight_scale = quantize_int8_rowwise(weight)
+            out = int8_mm_dequant(input_i8, weight_i8.T, input_scale, weight_scale)
+            out = out.view(*batch_dims, weight.shape[0])
+        else:
+            out = input @ weight.T
+
+        out = out + bias if bias is not None else out
+        return out
+
+    @staticmethod
+    def backward(ctx, grad_output):
+        input, weight = ctx.saved_tensors
+        grad_input = grad_weight = grad_bias = None
+
+        weight: Int8MixedPrecisionLinearWeight
+        sr = weight.config.stochastic_rounding
+
+        batch_dims = grad_output.shape[:-1]
+        grad_output = grad_output.view(-1, weight.shape[0])
+        input = input.view(-1, weight.shape[1])
+
+        if ctx.needs_input_grad[0]:
+            if weight.config.backward_grad_input:
+                grad_output_i8, grad_output_scale = quantize_int8_rowwise(grad_output)
+                weight_i8_t, weight_scale = quantize_int8_rowwise(weight.T)
+                grad_input = int8_mm_dequant(grad_output_i8, weight_i8_t.T, grad_output_scale, weight_scale)
+            else:
+                grad_input = grad_output @ weight
+            grad_input = grad_input.view(*batch_dims, weight.shape[1])
+
+        if ctx.needs_input_grad[1]:
+            if weight.config.backward_grad_weight:
+                grad_output_i8_t, grad_output_scale = quantize_int8_rowwise(grad_output.T)
+                input_i8_t, input_scale = quantize_int8_rowwise(input.T)
+                grad_weight = int8_mm_dequant(grad_output_i8_t, input_i8_t.T, grad_output_scale, input_scale)
+            else:
+                grad_weight = grad_output.T @ input
+
+        if ctx.needs_input_grad[2] and ctx.bias:
+            grad_bias = grad_output.sum(0)
+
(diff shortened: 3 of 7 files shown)


vkuzo:
nice! I am on vacation for most of this week but will try to take a look.

gau-nernst:
(Not marking this ready for review yet because I'm still updating some benchmarks number. I'm seeing some discrepancy in peak memory between ao's built-in Llama and HF's Llama. Otherwise the main code should be ready for review.)

torchao/prototype/quantized_training/README.md
```diff
@@ -46,8 +46,46 @@ BF16 compile    | 10.16915
 INT8 QT eager   | 10.11437
 INT8 QT compile | 10.03365
 
+## INT8 mixed-precision
+
+On NVIDIA GPUs, INT8 Tensor Cores can be up to 3x faster than their BF16/FP16 counterparts. In mixed-precision training, we can down-cast activations and weights dynamically to INT8 to leverage faster matmuls. However, since INT8 has very limited range [-128,127], we perform row-wise quantization, similar to how INT8 post-training quantization (PTQ) is done. Weight is still in original precision. This is inspired by prior works:
```
vkuzo, line 51: wondering if you have any context on why we see a 3x peak speedup and not a 2x peak speedup?

torchao/prototype/quantized_training/README.md
```diff
@@ -46,8 +46,46 @@ BF16 compile    | 10.16915
...
+- SwitchBack: [[paper](https://arxiv.org/abs/2304.13013)]
+
+Usage
+
+```python
+from torchao.prototype.quantized_training import int8_mixed_precision_training, Int8MixedPrecisionConfig
+from torchao.quantization import quantize_
+
+model = ...
+config = Int8MixedPrecisionConfig(
```
vkuzo, line 63: if you want to align with float8, the naming of the fields would be `output`, `grad_input`, `grad_weight`, we felt that best aligned with the rest of PyTorch

torchao/prototype/quantized_training/int8_mixed_precision.py
```diff
@@ -0,0 +1,203 @@
...
+
+from .int8 import quantize_int8_rowwise
+
+if has_triton():
+    from .int8_mm import int8_mm_dequant
+
+else:
+
+    def int8_mm_dequant(A: Tensor, B: Tensor, A_scale_rowwise: Tensor, B_scale_colwise: Tensor) -> Tensor:
+        return (A * A_scale_rowwise.view(-1, 1)) @ (B * B_scale_colwise.view(1, -1))
```
vkuzo, line 18: nit: write out each statement line by line to make more readable?

```
A_scaled = A * A_scale_rowwise.view(-1, 1))
B_scaled = ...
return A_scaled @ B_scaled
```

torchao/prototype/quantized_training/README.md
gau-nernst: I did microbenchmark here https://github.com/gau-nernst/quantized-training?tab=readme-ov-file#matmul on my GPU (sm89). Using both CuBLAS (via `torch._int_mm()`) and triton matmul. Perhaps I can add that this speed up is measured on my specific GPU (4070Ti SUPER, sm89). I can probably spin up an A100 instance and do the microbenchmark there too.

torchao/prototype/quantized_training/int8_mixed_precision.py
```diff
@@ -0,0 +1,203 @@
...
+from torch import Tensor, nn
+from torch.utils._python_dispatch import return_and_correct_aliasing
+from torch.utils._triton import has_triton
+
+from torchao.dtypes.utils import _dispatch__torch_dispatch__, _dispatch__torch_function__, _implements
+
+from .int8 import quantize_int8_rowwise
+
+if has_triton():
+    from .int8_mm import int8_mm_dequant
```
vkuzo, line 13: have you tried using `torch._int_mm` with torch.compile in `max-autotune` mode to generate the matmul?

gau-nernst: I haven't. I can try.

A question. Can `torch_int_mm()` generate INT8 matmul fused with dequant? If yes, what is the proper way to write matmul that torch.compile will codegen it correctly? I see some versions of this in dynamo template but not very sure under which condition it will be codegen-ed.

Some problems I often face:
- `max-autotune` can take very long, especially for training.
- triton codegen can be finicky i.e. it may or may not do what we expect. Feels like sometimes it's clearer to manually write out the triton kernel (though we may miss out some other fusion opportunities)

torchao/prototype/quantized_training/README.md
gau-nernst: I can also add an INT8 microbenchmark script here, similar to `benchmarks/float8/bench_matmul.py`.

gau-nernst: Added the INT8 microbenchmark and its results in the PR description. Surprisingly, I only observe 3x speedup in consumer sm89 cards. Even L40s (also sm89) doesn't see such speed up :thinking:

I updated the README to "2x faster" so it's more conservative.

torchao/prototype/quantized_training/int8_mixed_precision.py
vkuzo: makes sense.  To scale to all preceding/subsequent ops, IMO compiler is the only scalable option.  If you are ok with just covering the matmul and not fusing preceding/subsequent ops, manual kernel works great.  I was more just wondering if you have have tried using the compiler.

gau-nernst:
I was trying this with torchtune codebase on A100 and the results are disastrous. Definitely something is wrong here. Even in eager mode, the speed is crawling. Will investigate more when I have time 😢. At the moment, seems like this only works well for consumer cards.

image

andrewor14:
> Edit 2: added A100 end2end benchmarks with benchmarks/quantized_training/pretrain_llama2.py in PR description. Seeing 25% speedup. The problem is indeed isolated to torchtune.

@gau-nernst Is there an issue on torchtune for this? If not please open one so they can take a look. Thanks.
Also, is the convergence benchmark in the PR description up-to-date? Results look great so far!

gau-nernst:
@andrewor14 The torchtune testing was only done locally. Will open an issue on torchtune once this PR is merged.

The convergence benchmark is up-to-date, but it was done in my experiment repo (https://github.com/gau-nernst/quantized-training) before this PR. The implementation should be the same. I don't want to add the full training code from my experiment repo to torchao/benchmarks since it might be too much? (250 LoC might not be too much 🤔 
 https://github.com/gau-nernst/quantized-training/blob/main/llm_pretrain.py. But still, I think the benchmark example in torchao should be minimal, while a proper convergence test should be done in torchtune/torchtitan).

Will try if I can run this with torchtitan. Hopefully won't get the same issue as torchtune...

torchao/prototype/quantized_training/int8.py
```diff
@@ -12,13 +12,41 @@
 _c10d_functional = torch.ops._c10d_functional
 
 
+@torch.no_grad()
+def quantize_int8_rowwise(tensor: Tensor, stochastic_rounding: bool = False):
+    """Normal rounding will always round down small changes in weight update. To tackle this problem,
+    stochastic rounding can be used, which has a low chance, but not zero, of rounding up. The
+    probability of rounding up is equal to x - ⌊x⌋, which indicates how close the value is to the next
+    integer value. Thus, stochastic rounding also approximates the floating point value exactly.
```
andrewor14, line 20: add a note saying stochastic rounding seems to be not needed for int8 mixed precision?

torchao/prototype/quantized_training/int8_mixed_precision.py
```diff
@@ -0,0 +1,247 @@
...
+        if ctx.needs_input_grad[2] and ctx.bias:
+            grad_bias = grad_output.sum(0)
+
+        return grad_input, grad_weight, grad_bias, grad_config
+
+
+def int8_mixed_precision_training(config: Int8MixedPrecisionConfig = _DEFAULT_CONFIG):
+    # TODO: right now `_get_linear_subclass_inserter()` will always set `requires_grad=False`
+    # when we have this out of prototype (or there are stable trainable tensor subclasses),
+    # update `_get_linear_subclass_inserter()` to allow `requires_grad=True`.
```
andrewor14, line 239: If we need to do this in multiple places then maybe it's better to just add the flag :). Feel free to do it in a future PR

andrewor14 (approved):
Thanks @gau-nernst, looks great! For this PR, please add a note on the current FSDP limitations and update the results in the PR descriptions. I think we should be good to merge once you do that. We can address the FSDP challenges in future PRs.

One thing I'm wondering is whether it makes sense to extend the int8 mixed-precision flow to support full int8 data flow similar to Jetfire in the future. How much of the existing code can we reuse?

torchao/prototype/quantized_training/int8_mixed_precision.py
```diff
@@ -0,0 +1,247 @@
...
+    def convert_linear(cls, module: nn.Module, config: Int8MixedPrecisionConfig = _DEFAULT_CONFIG):
+        if module.__class__ is nn.Linear:  # exact match, don't swap nn.Linear subclasses
+            module.__class__ = cls
+            module.config = config
+            return
+        for child in module.children():
+            cls.convert_linear(child, config)
+
+
+def _dynamic_int8_linear(input: Tensor, weight: Tensor) -> Tensor:
```
vkuzo, line 182: I think this should be called `mm` and not `linear`

torchao/prototype/quantized_training/int8_mixed_precision.py
```diff
@@ -0,0 +1,247 @@
...
+
+# FSDP op
+@implements(aten.split.Tensor)
+def _(func, types, args, kwargs):
+    data_list = func(args[0]._data, *args[1:], **kwargs)
+    return [Int8MixedPrecisionLinearWeight(x, args[0].config) for x in data_list]
+
+
+# alternative UX
+class Int8MixedPrecisionLinear(nn.Linear):
```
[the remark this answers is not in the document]
vkuzo: would be good to clearly document any converegence issues, regardless of whether subclasses or modules are used as the user facing UX

torchao/prototype/quantized_training/README.md
```diff
@@ -37,17 +37,63 @@ Only `torch.optim.Adam` and optimizers from `torchao.prototype.low_bit_optim` ar
...
+During training, there are 3 matmuls involved in each `nn.Linear` layer:
+- 1 in forward: `output = input @ weight.T`
+- 2 in backward:
+  - `grad_input = grad_output @ weight`
+  - `grad_weight = grad_output.T @ input`
+
+You can configure which matmul to be applied with INT8 mixed-precision using `Int8MixedPrecisionConfig` shown above. If convergence is an issue, we recommend leaving `grad_weight` in original matmul precision, and also `grad_input` if the issue still persists.
+
+Note:
+- When we only apply INT8 mixed-precision in the forward pass, this can be considered QAT.
```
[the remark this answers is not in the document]
gau-nernst: If we consider dtype casting an implementation detail, and only think QAT as doing quantization in forward pass during training (in terms of numerics), then it should be the same 😄.

torchao/prototype/quantized_training/int8_mixed_precision.py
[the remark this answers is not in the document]
gau-nernst: I think the FSDP issue should be considered a "bug", not convergence issue. For the tensor subclass implementation, when I implement `F.linear()` directly as `F.linear(act, weight._data)`, there are still bugs in FSDP i.e. loss curve diverge from baseline.

This Linear subclass does not help with FSDP bug though. I added it to try out, but it didn't work. Getting a totally different error when I call `_Int8MixedPrecisionLinear.apply()` inside `Int8MixedPrecisionLinear.forward()`

I will probably remove this Linear subclass. For implementing INT8 all-gather in the future, we need a tensor subclass weight anyway, so there is no value in having this Linear subclass if it doesn't solve anything.

torchao/prototype/quantized_training/int8_mixed_precision.py
gau-nernst: From my microbenchmarks, INT8 mm kernel is most performant with `A @ B.T` i.e. linear, thus I want to only call INT8 kernel with this layout. Need to enforce `A` and `B` in `A @ B.T` to be contiguous though, especially in backward pass. Will double check on this layout+perf stuff.

Also, by only doing `A @ B.T`, I only need to support row-wise scaling for both `A` and `B`.

gau-nernst:
@andrewor14 Thank you for the feedback.

@vkuzo For context, the current state of this PR works well for single-GPU, non-FSDP cases. Even torchtune works well now (will update the PR description with new results). I have been struggling to get this work correctly for FSDP in torchtitan. As discussed in my previous reply, I believe this is a bug in "tensor subclass + FSDP" interaction, since if I implement `F.linear` as `F.linear(act, weight._data)`, the bug is still there.

Looking at FP8 and NF4 implementations, I think I might have to implement it via `aten.mm` instead of `F.linear`. Appreciate if you can also share knowledge/experience with supporting tensor subclass weight in FSDP.

Since a branch cut has been made for 0.5 release, I think there is no rush to merge this PR. I will try to solve the FSDP bug over the weekend, then finalize the PR. Will mark the PR as draft now.

torchao/prototype/quantized_training/int8_mixed_precision.py
```diff
@@ -0,0 +1,234 @@
...
+    B_t_i8, B_scale_colwise = quantize_int8_rowwise(B.T)
+    return int8_mm_dequant(
+        A_i8.contiguous(),
+        B_t_i8.contiguous().T,
+        A_scale_rowwise,
+        B_scale_colwise,
+    )
+
+
+class _Int8MixedPrecisionLinear(torch.autograd.Function):
```
gau-nernst, line 176: @vkuzo I notice I might need to add `@torch._dynamo.allow_in_graph` here like what is done in FP8 for certain things to work. Do you have any idea why it is the case?

Also, it seems like calling a custom autograd.Function under `__torch_function__` (NOT `__torch_dispatch__`) results in silent bugs, even in eager mode. (I know that things under `__torch_dispatch__` doesn't have gradients/autograd is not working at dispatch level). Do you know if "calling a custom autograd.Function under `__torch_function__`" is supported in general?

vkuzo: > I notice I might need to add @torch._dynamo.allow_in_graph here like what is done in FP8 for certain things to work. Do you have any idea why it is the case?

that decorator allows the "create the subclass" node to live in one level of the torch.compile IR without being decomposed (the decomposition happens later).  At the time when float8 was built, the features we needed for float8 only worked with this decorator, and not when the constructor was decomposed right away.  I hear this might be different now, but I haven't checked myself (because the current way works fine).

vkuzo:
> Looking at FP8 and NF4 implementations, I think I might have to implement it via `aten.mm` instead of `F.linear`. Appreciate if you can also share knowledge/experience with supporting tensor subclass weight in FSDP.

If you do the same thing `Float8Tensor/NF4Tensor/MXTensor` are doing, things should work.  Let us know if you have more specific questions, happy to help.

torchao/prototype/quantized_training/int8_mixed_precision.py
```diff
@@ -0,0 +1,222 @@
...
+        *,
+        out: Optional[Tensor] = None,
+    ):
+        (data,) = all_gather_outputs
+        (config,) = metadata
+        if out is not None:
+            assert isinstance(out, Int8MixedPrecisionTrainingLinearWeight)
+            assert out.config == config
+            return
+        return Int8MixedPrecisionTrainingLinearWeight(data.to(param_dtype), config), all_gather_outputs
```
gau-nernst, line 131: @vkuzo The only thing not working with FSDP now (in both eager and compile mode) is `MixedPrecisionPolicy(param_dtype=torch.bfloat16)` with FP32 model. I think this is where it goes wrong. With this code, the loss diverges from baseline.

If I change to the below, loss becomes NaN.

```python
data = data.to(param_dtype)
return Int8MixedPrecisionTrainingLinearWeight(data, config), (data,)
```

If I don't do anything about `param_dtype`, I will get `AssertionError: FSDP reduce-scatter expects uniform gradient dtype but got {torch.bfloat16, torch.float32}`. 

Do you know what should be the correct way to do this? Also, what should `fsdp_post_all_gather()` return? I suppose this is a private API so there is no documentation.

I think one way to circumvent this is to override `dtype` of tensor subclass and keeps track of this metadata to propagate across ops (like NF4), and in matmul/linear, return the correct dtype. However, it seems a bit cumbersome for what seems to be simple.

vkuzo: Not directly answering your question, but here is how I built / encouraged folks to build things for float8 training:
1. focus on promising accuracy + performance first, take the simplest possible path through the PT stack to get good numbers
2. improve UX, as needed, after (1) is done

At a high level, that meant using module swaps and sidestepping a lot of missing functionality in torch.compile + tensor subclasses, to move faster on (1).  If you want, you are welcome to do the same here.  That would mean something like:
1. have a Int8TraningLinear, weight is a regular tensor
2. tensor subclasses are transient and only used to work around autograd limitations and compose with distributed
3. if you follow what float8 is doing, things will work with compile/distributed/autograd as we've fixed most/all of the issues

Note that this isn't aligned with torchao inference, and that alignment will have to happen at some future time, I'm just decoupling that from unblocking early research / getting promising performance + accuracy.

vkuzo: > Also, what should fsdp_post_all_gather() return?

This is an API designed to allow FSDP to communicate in lower precision.  Since you are currently always wrapping weights with subclasses, something that would sound right to me is:

* first, just make FSDP comms work in higher precision, this means you have to ser/deser your weight wrapper in these APIs
* then, in a separate PR, optinally enable doing the comms in int8, this means you quantize in the fsdp_pre_all_gather API and reconstruct the already quantized tensor in the fsdp_post_all_gather API

vkuzo: > I think one way to circumvent this is to override dtype of tensor subclass and keeps track of this metadata to propagate across ops (like NF4), and in matmul/linear, return the correct dtype. However, it seems a bit cumbersome for what seems to be simple.

hmm, not sure exactly what this means, but for float8, dtype is set to what we want autograd to see.  I understand this might be confusing to a subset of people, but I think being practical and just having things work e2e with a slight regression to what is intuitive has worked well in that case.  There isn't a single thing we can use as a "dtype" for arbitrary tensor subclasses anyways which composes with all the PyTorch systems.

gau-nernst: Yea I think "module swap + regular weight tensor + call custom autograd.Function in forward" is simpler to implement too. May need to add `@torch._dynamo.allow_in_graph` to the custom autograd.Function for compile to work. I tested this combo and it works for single GPU. Will try if it works for FSDP w/ mixed-precision param_dtype.

The main reason I started this out as tensor subclass is because of familiarity when I wrote other stuff in torchao. Also, for quantized weight before all-gather, tensor subclass for linear weight is needed (similar to `WeightWithDynamicFloat8CastTensor`) (though I agree that this shouldn't be a big reason at this stage). Anyway, single GPU performance is already pretty good 😄, only FSDP is left. Though I guess FSDP performance is more important for big corps 😄.

> This is an API designed to allow FSDP to communicate in lower precision

Yea I realize if the tensor subclass doesn't implement this method, the following ops will be called

```python
torch.ops.fsdp.all_gather_copy_in.default,
torch.ops.fsdp.split_with_sizes_copy.default,
torch.ops.c10d._allgather_base_.default,
```

It's not clear which op here should return the wrapped tensor (so that later `F.linear()` dispatch is correct). So maybe it's also possible to not use `fsdp_pre/post_all_gather()` if I don't need low-precision comms at the moment.

> first, just make FSDP comms work in higher precision, this means you have to ser/deser your weight wrapper in these APIs

Yes, this is my intention. But right now it doesn't work with `MixedPrecisionPolicy(param_dtype=torch.bfloat16)` as I said previously. There will be complaints about dtype (thus I thought about keeping tracking of a separate dtype metadata like NF4). If there is no MixedPrecisionPolicy, the current state of this PR works with FSDP, but very slow even with torch.compile.

vkuzo: > It's not clear which op here should return the wrapped tensor (so that later F.linear() dispatch is correct).

without taking a close look at the code, I think a good answer would be to handle all of these ops in your subclass properly.  Otherwise, I would expect the logic to get pretty brittle / hard to debug real fast.

gau-nernst: Just tested this combo "module swap + regular weight tensor + call custom autograd.Function in forward + `@torch._dynamo.allow_in_graph`" and it works with FSDP + MixedPrecisionPolicy now 😭. Seeing +15% speedup with 8x A100 for Llama3-8B pre-training. (for 1x A100, I see +40% speedup for Llama2-7B pre-training).

For `fsdp_post_all_gather()`, I guess I'm not supposed to change the data here, including dtype casting. Argument `param_dtype` should only be used to create "appearance" dtype for the tensor subclass. Maybe you can help me confirm it with FSDP people?

I will add back the module swap UX. I think I can leave the tensor subclass as is for future PR to implement INT8 comms (though right now I'm seeing a problem in backward all-gather 🌚 - I'm doing row-wise scaling for weight in forward, which is fine since weight is sharded along dim0, but in backward, I'm doing col-wise scaling (since we need to compute `grad_weight = grad_output.T @ input`) -> not compatible with dim0 sharding. Perhaps I need to support tensor-wise scaling for weight, which is the same as [SwitchBack](https://arxiv.org/abs/2304.13013) and FP8 recipes, though I suspect tensor-wise scaling for INT8 will have problems... OR tile-wise scaling (which requires more customized triton kernel...)

vkuzo: for some additional context, my high level plan for float8 was
1. eventually remove the module swap UX in favor of either module hooks or torch_function overrides (like you have in this PR), for better handling of unknown unknowns.  This will likely involve a good amount of e2e torch.compile work to ensure the new solution does not regress accuracy and performance
2. align with torchao inference, since (1) will not be doing exactly what torchao inference is doing since we will need either hooks or torch_function overrides, due to training-specific functionality.

Neither of these have a timeline yet, I think 2024Q4 would be a good time to start on (1), and (2) will depend on specifics of (1).  cc @jerryzh168 , @msaroufim  for context.

gau-nernst: @vkuzo I got FSDP w/ MixedPrecisionPolicy working for tensor subclass UX too! The trick is to do dtype casting in `fsdp_pre_all_gather()` instead of `fsdp_post_all_gather()`. Sadly `param_dtype` is not exposed to pre all-gather, so I think I need to add an extra field in `Int8MixedPrecisionTrainingConfig` as a workaround.

The performance is better than module swap API too! I'm seeing 20% speedup with tensor subclass compared to 15% speedup with module swap. (I think it might have to do with the use of `@torch._dynamo.allow_in_graph`, and since I put my quantization code inside the custom `autograd.Function`, it will not compile the quantization code?)

vkuzo: > I got FSDP w/ MixedPrecisionPolicy working for tensor subclass UX too!

that's great.  The torch_function+compile path was not ready in the past when float8 was built, great to hear that the story is better now.  If everything in fact works e2e nowadays with this approach, I could see float8 migrating to something like this in the future to unify with the rest of torchao better.

> The performance is better than module swap API too! I'm seeing 20% speedup with tensor subclass compared to 15% speedup with module swap.

that sounds like a bug, lmk if you have logs from a small toy model to demonstrate what is different, I'd be interested to check that out

gau-nernst: > that sounds like a bug, lmk if you have logs from a small toy model to demonstrate what is different, I'd be interested to check that out

What kind of logs do you need? Like profile trace? I don't keep this code anymore, so need some efforts to reproduce it. If you really want to check it out, I can produce it for you.

Also a side observation. Without `fsdp_pre/post_all_gather()`, I think we can't propagate tensor subclass across FSDP all-gather. I tried wrapping the outputs of the 3 FSDP ops mentioned earlier (`fsdp.all_gather_copy_in.default`, ...) but still could not propagate the tensor subclass. I think this is why NF4 also needs to use `fsdp_pre/post_all_gather()`, even though it doesn't exactly do dist comm quantization (just all-gather the pre-quantized weight, and wrap it with tensor subclass again).

torchao/prototype/quantized_training/int8_mixed_precision.py
```diff
@@ -0,0 +1,247 @@
...
+            cls.convert_linear(child, config)
+
+
+def _dynamic_int8_linear(input: Tensor, weight: Tensor) -> Tensor:
+    input_i8, input_scale = quantize_int8_rowwise(input)
+    weight_i8, weight_scale = quantize_int8_rowwise(weight)
+    return int8_mm_dequant(input_i8, weight_i8.T, input_scale, weight_scale)
+
+
+class _Int8MixedPrecisionLinear(torch.autograd.Function):
```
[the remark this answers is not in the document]
gau-nernst: For this pointer, I added doc to `_dynamic_int8_mm()` instead, and within `_Int8MixedPrecisionLinear`, it only calls the (quantize + int8 matmul + dequant) function -> inside this custom autograd.Function, we basically just swap `torch.mm` with `_dynamic_int8_mm()`

torchao/prototype/quantized_training/int8_mixed_precision.py
```diff
@@ -0,0 +1,247 @@
...
+
+import torch
+from torch import Tensor, nn
+from torch.utils._python_dispatch import return_and_correct_aliasing
+from torch.utils._triton import has_triton
+
+from torchao.utils import TorchAOBaseTensor
+
+from .int8 import quantize_int8_rowwise
+
```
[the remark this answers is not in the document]
gau-nernst: Added under `Int8MixedPrecisionTrainingLinearWeight`

gau-nernst:
Added torchtitan FSDP results to PR description. Looking very good 😄

@andrewor14 I have made all the changes I want to make. You can take a look through everything again to see if I haven't addressed any of your earlier feedback + any other changes you want.

M1 build wheels CI failing is unrelated.

torchao/prototype/quantized_training/int8_mixed_precision.py
vkuzo: > What kind of logs do you need? Like profile trace? I don't keep this code anymore, so need some efforts to reproduce it. If you really want to check it out, I can produce it for you.

feel free to ignore since looks like it will be a lot of work.  I can check it out from my machine after this lands, although it will be on different hardware.

gau-nernst: btw, just curious, do you have end2end speedup numbers for FP8? Specifically for Llama3-8B. I see there is one here https://github.com/pytorch/torchtitan/blob/main/docs/performance.md, but it's with tensor parallel and for Llama3-405B.

torchao/prototype/quantized_training/int8_mixed_precision.py
gau-nernst: I was playing around a bit with `torch._int_mm()` with max-autotune to codegen the fused matmul. Some observations:
- Seems like currently inductor can only fuse 1 epilogue multiplication. I think the relevant code is here? https://github.com/pytorch/pytorch/blob/90e12cf63d7ead464e9f99007b50b494e82298ca/torch/_inductor/kernel/mm.py#L757
- This might not be an issue, as the other multiplication can be fused into the subsequent op. However, in a small end2end benchmark, my explicit triton kernel is still much faster (50%).

Therefore, I think for now using the explicit triton kernel is still better.

What I'm curious is that is there a mechanism to register a custom op with torch compiler that provides 2 implementations (e.g. 1 hand-written fused Triton impl, and 1 pure PyTorch impl which can potentially be fused across op boundaries) so that the compiler can benchmark and choose the better one in a specific scenario.

andrewor14:
Looks good to me! I'm merging this and cherry-picking this into the 0.5 branch. Thanks for all your hard work figuring out the FSDP issue. If there are remaining issues let's address them in follow-up PRs.

andrewor14:
By the way @gau-nernst I suspect the reason why you weren't able to get good numerics with torchtune is because fine-tuning with C4 there is simply broken. I ran into this myself recently: https://github.com/pytorch/torchtune/issues/1526. Probably won't be able to test that particular workflow until they fix this problem.

torchao/prototype/quantized_training/int8_mixed_precision.py
vkuzo: this is the most recent polished numbers we have: https://dev-discuss.pytorch.org/t/enabling-float8-all-gather-in-fsdp2/2359 .  It's on LLaMa 3 70B, 1.5x speedup on 128 H100s with FSDP but no TP/SP.

gau-nernst: The setup there seems like doesn't use many special tricks yet (still just dynamic scaling) + FP8 all-gather (which probably is not too significant for an 8x A100 setup like I used). So there are still room to optimize INT8 training further :smile: (assuming INT8 and FP8 speedups over BF16 are similar) (or maybe using Llama3-70B will have larger speedup simply thanks to larger matmul -> more % time spent in matmul)

torchao/prototype/quantized_training/int8_mixed_precision.py
```diff
@@ -0,0 +1,231 @@
...
+            aten.split.Tensor,
+            aten.clone.default,
+        }:
+            # return new wrapped object
+            return pytree.tree_map_only(Tensor, lambda x: cls(x, config), out)
+        else:
+            # return new unwrapped object
+            return out
+
+    def fsdp_pre_all_gather(self, mesh):
```
awgu, line 117: By the way, I was thinking of changing the signature to:
```
def fsdp_pre_all_gather(
    self,
    mesh: DeviceMesh,
    module: nn.Module,  # owning module (e.g. `lin` if this is `lin.weight`)
    mp_policy: MixedPrecisionPolicy,  # policy of the owning FSDP module
):
```
as (1) having access to the `module` can allow using the module as a scratch space for any needed state and (2) having access to the `mp_policy` helps with situations like the one here.

My main concern with doing this is how to deal with the backward compatibility issue. Technically, FSDP2 is still in prototype release, so we are allowed to break BC. However, libraries like ao and torchtune may need to do if/else based on PyTorch version. Maybe it is better to do it sooner than later though.

awgu: Here we go: https://github.com/pytorch/pytorch/pull/136129
will need more time to test it -- fortunately the subclasses using the FSDP extension are all in ao now it seems 😆

Issue #744, opened 2024-08-24
Low bit optimizers quality


tsengalb99:
I saw that the quality numbers in https://github.com/pytorch/ao/tree/main/torchao/prototype/low_bit_optim changed recently. Was there a bug in the AO low bit optimizer implementation before?

msaroufim:
There was a bug with how we handled LR schedulers that make things go faster now https://github.com/pytorch/ao/pull/736

tsengalb99:
So nothing functionally changed and the optimizers are faster -> you were able to test on a larger model?

msaroufim:
Largest model we tried was on the order of 500M parameters, for 1B+ try it out and feel free share loss curves

tsengalb99:
Did you test pretraining or just finetuning only? Thanks

msaroufim:
This was a finetuning benchmark

Benchmark script for fine-tuning a [timm](https://github.com/huggingface/pytorch-image-models) model on [resisc45](https://huggingface.co/datasets/timm/resisc45) dataset is available at [benchmarks/benchmark_low_bit_adam.py](https://github.com/pytorch/ao/blob/main/benchmarks/benchmark_low_bit_adam.py)

We were hoping to expand to more fine-tuning benchmarks https://github.com/pytorch/torchtune here

And we're doing with fully quantized training here https://github.com/pytorch/ao/tree/main/torchao/prototype/quantized_training

tsengalb99:
I tried using the AO 8 bit optimizer in nanoGPT by swapping out AdamW with Adam8bit [here](https://github.com/karpathy/nanoGPT/blob/9755682b981a45507f6eb9b11eadef8cb83cebd5/model.py#L284) and keeping the embedding and non-decayed parameters (layernorms etc) in a regular torch AdamW instance. This strategy seems to work with the lpmm codebase, where I get close to FP32 performance, but with AO I get severely degraded performance. Is there anything special I need to do with the AO low bit optimizers? Also, I wasn't able to get the AO 4 bit optimizer to work out of the box. I had to disable the compile call [here](https://github.com/pytorch/ao/blob/af6803135d153689bf5da537815992d93ed72417/torchao/prototype/low_bit_optim/adam.py#L226) or compile would complain about the input to `lerp` being negative. I'm using torch 2.4.0 if that matters.

msaroufim:
[...]

PR #755, opened 2024-08-27
[low-bit optim] Fix Adam4bit support on PyTorch 2.3 and 2.4. Update AdamFp8 torch requirement

gau-nernst:
See https://github.com/pytorch/ao/issues/744#issuecomment-2312462039


msaroufim (approved):

gau-nernst:
Fixed issue with 4-bit Adam. Now 4-bit Adam works with PyTorch 2.3 likes in the past. Hopefully CI is grenn. The issue seems to be related to this https://github.com/pytorch/pytorch/issues/128649

I kinda feel conflicted about this change, since now the optimizer state is flattened, instead of having the same shape as param. Will try a better solution in the future. I think it has to do with dynamic compile also. 4-bit optim is giving us a lot of headaches 🤣.

msaroufim (approved):

PR #760, opened 2024-08-27
float8 training: add static scaling

vkuzo:
Summary:

This is useful for things such as:
* activation_with_bounded_range -> linear (can set static scale to activation range)
* bounding weight scales to known quantities if the modeling user can guarantee their magnitude throughout training

We don't have signal yet that this is useful for production things, but it would be good to land this to enable easy experimentation.

Test Plan:

Unit and integration tests pass:
```
./test/test_everything.sh
// note that there is a failure in `test_fsdp2.py` which is present on main
```

Use float8 profiling script to see GPU kernel time go down as we enable static scaling on a toy model:
https://gist.github.com/vkuzo/b2cf46f7cccb691125566873859ca39d

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/float8/test_base.py
+++ b/test/float8/test_base.py
@@ -134,6 +134,7 @@ class TestFloat8Tensor(unittest.TestCase):
         fp8_b.copy_(fp8_a)
         torch.testing.assert_close(fp8_a._data, fp8_b._data)
 
+    @pytest.mark.skip("broken")
     def test_weights_only_load(self):
         module = nn.Linear(16, 16)
         # Save model state dict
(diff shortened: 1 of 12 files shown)


test/float8/test_base.py
```diff
@@ -134,6 +134,7 @@ def test_copy_(self):
         fp8_b.copy_(fp8_a)
         torch.testing.assert_close(fp8_a._data, fp8_b._data)
 
+    @pytest.mark.skip("broken")
```
vkuzo, line 137: unrelated to this PR, but broken on main

test/float8/test_everything.sh
```diff
@@ -15,7 +15,7 @@ then
 ./test/float8/test_fsdp.sh
 ./test/float8/test_fsdp_compile.sh
 ./test/float8/test_dtensor.sh
-pytest test/float8/test_fsdp2/test_fsdp2.py
+python test/float8/test_fsdp2/test_fsdp2.py
```
vkuzo, line 18: unrelated to this PR, but calling this with `pytest` is broken on main

drisspg:
cc @sayakpaul  this PR will be very relevant for LoRA training

sayakpaul:
Thanks for pinging me here :) 

LMK how/if I can amend [my script](https://gist.github.com/sayakpaul/f0358dd4f4bcedf14211eba5704df25a) to benefit from this.

y-sq (approved):

PR #767, opened 2024-08-28
Revert "Refactor quant_llm to work with affine quantized tensor"

jerryzh168:
Reverts pytorch/ao#696

### DIFF
--- a/benchmarks/benchmark_fp6_llm.py
+++ b/benchmarks/benchmark_fp6_llm.py
@@ -1,16 +1,16 @@
 import torch
 import pandas as pd
 import torch.nn.functional as F
-# from torchao.prototype.quant_llm import QuantLlmLinearWeight
-from torchao.dtypes import to_affine_quantized_fpx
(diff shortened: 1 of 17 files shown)


jerryzh168:
looks like it breaks CI, it has an error that does not appear in the PR CI: https://github.com/pytorch/ao/actions/runs/10588415262/job/29340895949

gau-nernst (approved):

msaroufim (approved):

gau-nernst:
I think it might have to do with importing C++ extension. Depending on the import order, it might not get imported

https://github.com/pytorch/ao/blob/da8c5b969af2874f8af4f06508f3243ab747aef7/torchao/__init__.py#L25

Maybe that line should be inside `ops.py`? Or some kind of guards to prevent undefined ops because they are defined in C++?

jerryzh168:
yeah let me test putting this in ops.py first, that sounds like a reasonable fix, although historically we have been struggling with this import..

PR #772, opened 2024-08-28
[reland] Refactor quant_llm to work with affine quantized tensor (#696)

jerryzh168:
Summary:
We want to add quant_llm to affine quantized tensor as a general fp2-fp7 dtype, before that we need to refactor the current implementation to work with AffineQuantizedTensor first

Test Plan:
python test/prototype/test_quant_llm.py

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/benchmarks/benchmark_fp6.py
+++ b/benchmarks/benchmark_fp6.py
@@ -1,16 +1,16 @@
 import torch
 import pandas as pd
 import torch.nn.functional as F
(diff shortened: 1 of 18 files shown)


jerryzh168:
the problem is the previous error in CI does not really appear here, maybe we should do a forward fix. cc @msaroufim @gau-nernst

jerryzh168:
update: moved the import for quant_llm op, verified locally that "import torchao" works

PR #784, opened 2024-08-30
Move more utils to TorchAOBaseTensor

jerryzh168:
Summary:
This moves over _implements, _dispatch__torch_dispatch__, _dispatch__torch_function__, _register_layout_cls and _get_layout_tensor_constructor to `TorchAOBaseTensor` so when people inherit from this, they can get these utils directly

Test Plan:

rely on CI for other tests

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/prototype/test_low_bit_optim.py
+++ b/test/prototype/test_low_bit_optim.py
@@ -75,7 +75,7 @@ class TestQuantize(TestCase):
 
 
 class TestOptim(TestCase):
-    @pytest.mark.skipif(not TORCH_VERSION_AT_LEAST_2_3, reason="requires PyTorch >= 2.3")
+    @pytest.mark.skipif(not TORCH_VERSION_AT_LEAST_2_4, reason="requires PyTorch >= 2.3")
     @parametrize("optim_name", ["Adam8bit", "AdamW8bit", "Adam4bit", "AdamW4bit", "AdamFp8", "AdamWFp8"])
     @parametrize("dtype", [torch.float32, torch.bfloat16])
     @parametrize("device", _DEVICES)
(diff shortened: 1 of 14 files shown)


HDCharles:
can this handle the other ones from https://github.com/pytorch/ao/issues/710?

feels like there should be a way to set it up so that things like detach are handlable with a combination of __tensor_flatten__ and the constructor?

HDCharles (approved):
looks good otherwise

jerryzh168:
> can this handle the other ones from #710?
> 
> feels like there should be a way to set it up so that things like detach are handlable with a combination of **tensor_flatten** and the constructor?

yeah will do in a future PR

test/prototype/test_low_bit_optim.py
```diff
@@ -75,7 +75,7 @@ def test_quantize_4bit_with_qmap_compile(self, device):
 
 
 class TestOptim(TestCase):
-    @pytest.mark.skipif(not TORCH_VERSION_AT_LEAST_2_3, reason="requires PyTorch >= 2.3")
+    @pytest.mark.skipif(not TORCH_VERSION_AT_LEAST_2_4, reason="requires PyTorch >= 2.3")
```
jerryzh168, line 78: cc @gau-nernst multiple levels of inheritance breaks pytorch 2.3 version, so I had to use 2.4+ for this

gau-nernst: Personally I would prefer to keep supporting PyTorch 2.3 for low-bit optimizers. @msaroufim What do you think?

From the CI link you show, it seems like only 4-bit optim breaks, while 8-bit and Fp8 optim still work? Anyway, if we keep this change, also need to update the docs.

gau-nernst: After some investigation, the errors only happen to 4-bit optim, and because 4-bit optim uses dynamic-shape compile. Some strange interactions between inheritance and dynamic-shape :thinking:. Might have some fix soon..

PR #791, opened 2024-09-02
Skip fpx compile test

msaroufim:
Temporary workaround for https://github.com/pytorch/ao/issues/792

Skips the offending test, might be better than pinning the torch version like in https://github.com/pytorch/ao/pull/790 because we're close to a code freeze on Sep 5 and a pinned PyTorch version means we need to delay the release


msaroufim:
Planning on ninja landing so main can be green today

PR #805, opened 2024-09-04
Update the way scale is calculated for affine Symmetric

iseeyuan:
@TiRune identified this option. In the situation where the absolute value of quantized min and max can be different, like [-8, 7], we can calculate the scale factor with the pos and neg individually, and pick the larger one. It shows perplexity improvement in llama-like 4-bit weight quantized models. 

before:
'word_perplexity,none': 24.198390005931635
after:
'word_perplexity,none': 23.25360136363946

In this PR, one mapping type, `SYMMETRIC_MAX_POS_NEG` is added, to get the group symmetric quantization scales as mentioned above.

Please refer to the inline comments for the reasoning.


torchao/quantization/quant_primitives.py
```diff
@@ -730,8 +730,10 @@ def _choose_qparams_affine(
         max_val_pos = max_val
 
     if mapping_type == MappingType.SYMMETRIC.name:
-        max_val_pos = torch.max(-min_val_neg, max_val_pos)
-        scale = max_val_pos / (float(quant_max - quant_min) / 2)
+        smin = min_val_neg / float(quant_min)
```
jerryzh168, line 733: discussed offline that @iseeyuan will create a new mapping_type for this, and this is not always better than the existing way of computing scale. we can discuss the naming a bit later

TiRune: In what scenarios would it be worse? I would argue that this is always intended behavior, for any symmetric setting.

TiRune: There's another issue not mentioned yet, if you take the current scheme, export the weights that are qdqed, then load them back in again with the min-max quantizer, the results change everytime you do this. With this PR, the weights will stay the same even if you apply the min-max quantizer multiple times.

jerryzh168: @TiRune why you never want to clip values? it's always a trade off between clipping error and rounding error I think?

TiRune: For me, it's more about what's expected of a min-max quantizer. Sure, we always need to trade off clipping and rounding error. But that's what the MSE-based range setting, HQQ or those types of algorithms are for. The current choice for symmetric is kinda arbitrary to be 1/2 (q_min + q_max). 

To me, a min-max range setter is 1) it always includes both the min-and max in the range so there's no clipping error and 2) if it's applied twice the same result comes out i.e., f(f(x)) = f(x). We use this for e.g. to export fake quant weights and load them in another library like Executorch :D

TiRune: Another use-case of the min-max quantizer: Use it as the worst-case scenario for quantization to seed HQQ/MSE based range setting with.

E.g. you start with min-max, then search over shrinked versions of this range like in AWQ, you search over 0.99^N * the scale factor. In this algorithm version, you also expect the min-max quantizer to start off with 0 clipping error.

iseeyuan: Thanks @TiRune and @jerryzh168 ! To play safe incase there's usage of the old mapping, and quickly unblock the usage (without fixing all tests, especially those old tests that are planned to update), I added a new symmetric mapping type. 

Later if we notice that the the two symmetric mappings can merge to one, we can have a PR to merge them.

Please review the code and let me know if it make sense.

jerryzh168: @TiRune I see, yeah that makes sense I think. as @iseeyuan the current impl is used in all the existing impls so changing the behavior will be bc-breaking, it will be better to add a new mapping type.

test/quantization/test_quant_api.py
```diff
@@ -516,7 +516,7 @@ def test_quantized_tensor_subclass_8da4w(self):
         from torchao.quantization.quant_api import Int8DynActInt4WeightQuantizer
         from torchao.quantization.GPTQ import Int8DynActInt4WeightLinear
 
-        quantizer = Int8DynActInt4WeightQuantizer(groupsize=group_size)
+        quantizer = Int8DynActInt4WeightQuantizer(groupsize=group_size, mapping_type=MappingType.SYMMETRIC)
```
jerryzh168, line 519: nit: this can be the default?

torchao/quantization/quant_primitives.py
```diff
@@ -41,12 +41,18 @@ class MappingType(Enum):
     we'll use (-10.2, 10.2) as the range for floating point and map that to (-8, 7)
     e.g. scale = (10.2 - (-10.2)) / (7 - (-8))
 
+    SYMMETRIC_MAX_POS_NEG is a variant of symmetric mapping, where the scale is the max of smin
```
jerryzh168, line 44: how about `SYMMETRIC_NO_CLIPPING_ERR`? also cc @TiRune if there is suggestion for the name

jerryzh168:
error is relevant I think, maybe there is some changes to the QAT behavior of 8da4w

torchao/quantization/GPTQ.py
```diff
@@ -1022,13 +1024,15 @@ def __init__(
         precision: torch.dtype = torch.float32,
         scales_precision: torch.dtype = torch.float32,
         device: torch.device = torch.device("cpu"),
+        mapping_type: MappingType = MappingType.SYMMETRIC_MAX_POS_NEG
```
jerryzh168, line 1027: actually why is this set to the new type? I think it will be safer to keep the old one?

torchao/quantization/GPTQ.py
```diff
@@ -1022,13 +1024,15 @@ def __init__(
         precision: torch.dtype = torch.float32,
         scales_precision: torch.dtype = torch.float32,
         device: torch.device = torch.device("cpu"),
+        mapping_type: MappingType = MappingType.SYMMETRIC_NO_CLIPPING_ERR
```
jerryzh168, line 1027: it looks like this one is still not updated, is this intentional?

torchao/quantization/quant_primitives.py
```diff
@@ -41,12 +41,18 @@ class MappingType(Enum):
     we'll use (-10.2, 10.2) as the range for floating point and map that to (-8, 7)
     e.g. scale = (10.2 - (-10.2)) / (7 - (-8))
 
+    SYMMETRIC_NO_CLIPPING_ERR is a variant of symmetric mapping, where the scale is the max of smin
```
jerryzh168, line 44: cc @gau-nernst wondering if this will help int8 training as well: https://github.com/pytorch/ao/blob/1ce7da941b311ef6ab416d9c7a22be20e65b7495/torchao/prototype/quantized_training/int8.py#L65

gau-nernst: Possibly, will need to test. All INT8 training papers I see use 127 for both +ve and -ve though.

One potential concern I foresee is about quantization speed. During training we keep doing quantization, so having a fast one is crucial. Again, will need to try out / profile to be sure 😄.

torchao/quantization/GPTQ.py
iseeyuan: Oops, updated.

torchao/quantization/quant_primitives.py
jerryzh168: actually it seems even without this change 127 can be achieved with setting quant_min/quant_max to be (-127, 127)

yeah we could benchmark perf if we decide to use AQT for int8 training

test/quantization/test_qat.py
```diff
@@ -212,7 +213,7 @@ def test_qat_8da4w_quantizer(self):
         m = M()
         m2 = copy.deepcopy(m)
         qat_quantizer = Int8DynActInt4WeightQATQuantizer(groupsize=group_size)
-        ptq_quantizer = Int8DynActInt4WeightQuantizer(groupsize=group_size)
+        ptq_quantizer = Int8DynActInt4WeightQuantizer(groupsize=group_size, mapping_type=MappingType.SYMMETRIC)
```
jerryzh168, line 216: nit: if this is default now, we can remove the arg here I think

test/quantization/test_quant_primitives.py
```diff
@@ -164,7 +164,7 @@ def test_get_group_qparams_symmetric(self):
         scale_obs = scale_obs.reshape(weight.shape[0], -1)
 
         # assert that scales are identical
-        (scale_ao, _) = get_group_qparams_symmetric(weight, n_bit, groupsize, precision=torch.float16)
+        (scale_ao, _) = get_group_qparams_symmetric(weight, n_bit, groupsize, precision=torch.float16, mapping_type=MappingType.SYMMETRIC)
```
jerryzh168, line 167: same more these

test/quantization/test_quant_primitives.py
```diff
@@ -179,11 +179,27 @@ def test_choose_qparams_group_sym(self):
         precision = torch.float32
         scale, zero_point = choose_qparams_affine(input, mapping_type, block_size, dtype, eps=eps, scale_dtype=precision, zero_point_dtype=precision)
 
-        scale_ref, zp_ref = get_group_qparams_symmetric(input, n_bit=8, groupsize=2, precision=precision)
+        scale_ref, zp_ref = get_group_qparams_symmetric(input, n_bit=8, groupsize=2, precision=precision, mapping_type=mapping_type)
 
         self.assertTrue(torch.equal(scale, scale_ref))
         self.assertTrue(torch.equal(zero_point, zp_ref))
 
+    def test_choose_qparams_group_sym_pos_neg(self):
```
jerryzh168, line 187: nit: please update the name of the test as well

jerryzh168 (approved):
LGTM, thanks @iseeyuan and @TiRune

PR #806, opened 2024-09-04
Add aten.linear.default implementation to mx_ops

### LINKED ISSUE #796: NotImplementedError: aten.linear.default not implemented when using MXTensor
Hey I'm using the MX datatypes. It seems like the aten.linear.default function has not been implemented which causes the linear layers in the attenion layers not work with the MX datatypes.

Can you please implement this function in mx_ops.py? 
Thanks!

Ali-Flt:
Fixes #796

### DIFF
--- a/torchao/prototype/mx_formats/mx_ops.py
+++ b/torchao/prototype/mx_formats/mx_ops.py
@@ -156,3 +156,17 @@ def autocast_to_copy(aten_op, args, kwargs=None):
     )
     # print('after', res, res.dtype, res._orig_dtype)
     return res
+
+@implements([aten.linear.default])
+def mx_linear(aten_op, args, kwargs=None):
+    a = args[0]
+    b = args[1]
+    if len(args) > 2:
+        c = args[2]
+    else:
+        c = None
+    assert isinstance(a, MXTensor) and isinstance(b, MXTensor)
+    a_hp = a.to_dtype(a._orig_dtype)
+    b_hp = b.to_dtype(b._orig_dtype)
+    res = aten_op(a_hp, b_hp, c)
+    return res


vkuzo:
thanks! Could you also add a test to https://github.com/pytorch/ao/blob/main/test/prototype/mx_formats/test_mx_linear.py ?

Ali-Flt:
@vkuzo I tried different ways to trigger the use of aten.linear but couldn't

vkuzo:
I'd definitely recommend a test that fails before this PR and passes after this PR.  Would it work to wrap the code snippet you were using (which ended up calling into the linear override) into a test?

gau-nernst:
Just curious. Would it be better to implement `F.linear()` under `__torch_function__()` instead? Previously I also faced strange behavior on what aten ops will be dispatched by `F.linear()`, so implementing `F.linear()` directly would solve the problem.

vkuzo:
> Would it be better to implement F.linear() under __torch_function__() instead?

As of ~months ago, `__torch_dispatch__` was better supported with torch.compile, at least for the things we needed for float8.  I haven't checked if torch.compile + `__torch_function__` coverage is better now, would be good to check.

balaabhijit:
+1 Kudos for the fix! Can confirm I also ran into this same error without this PR

balaabhijit:
A simple test which will fail before this PR on:
```
torch==2.5.1+cu121
torchao==0.6.1
```

```python

import torch
import torch.nn as nn

from torchao.prototype.mx_formats.mx_linear import swap_linear_with_mx_linear

class MLP(nn.Module):
    def __init__(self, in_features: int = 128, out_features: int = 256):
        super().__init__()
        self.fc1 = nn.Linear(in_features=in_features, out_features=out_features)

    def forward(self, x):
        x = self.fc1(x)
        x = torch.nn.functional.gelu(x)
        return x


model = MLP()
# Does not hit the error with swap_linear_with_mx_inference_linear
swap_linear_with_mx_linear(
        model, elem_dtype=torch.float8_e4m3fn, block_size=32)

input_tensor = torch.randn(10, 128)
with torch.inference_mode():
    _ = model(input_tensor)


```

balaabhijit:
@vkuzo Thanks for the great insight on `__torch_dispatch__` vs `__torch_function__` this is super helpful

PR #807, opened 2024-09-04
[StaticQuant] add a linear observer class and test

drisspg:
Stacked PRs:
 * __->__#807

--- --- ---

[StaticQuant] add a linear observer class and test

This adds a new flow for calculating static activation scales. Rough example of flow:
```Python
import torch
import torch.nn as nn
from torchao.quantization.linear_observer_tensor import insert_observers_
from torchao.quantization.observer import (
    AffineQuantizedMinMaxObserver,
    PerTensor,
    MappingType
)

# Create observers
input_observer = AffineQuantizedMinMaxObserver(
    MappingType.SYMMETRIC,
    torch.float8_e4m3fn,
    granularity_type=PerTensor(),
    eps=torch.finfo(torch.float32).eps,
    scale_dtype=torch.float,
    zero_point_dtype=torch.int,
    zero_point_domain=None,
)

# Create a linear module
linear_module = nn.Linear(10, 20)

# Convert the linear module's weight to an observed tensor
insert_observers_(linear_module, input_observer, weight_observer=None)

# The linear_module can now be used as usual, with observers calculating statistics
output = linear_module(torch.randn(10, 10))

# Get the scale and zero point of the input observer
scale, zero_point = linear_module.weight.input_observer.calculate_qparams()
```

PR #808, opened 2024-09-04
[Float8] Make Inference and Training code independent

drisspg:
Stacked PRs:
 * #807
 * __->__#808

--- --- ---

# Summary

Remove Old FP8 inference flow and and switch over to Float8MMConfig
 in favor of the newly added support to AQT + quantize_ apis. It also completely separates and utilization on training code by creating a new `Float8MMConfig` and a addmm wrapper in the inference.py file


torchao/float8/float8_tensor.py
```diff
@@ -47,31 +45,41 @@
...
-# emulate: whether to emulate the matmuls in fp32
-# use_fast_accum: whether to use the fast-accumulation option for scaled_mm
-# fp8_output: whether to output the result of the scaled_mm in fp8
-# pad_inner_dim: whether to pad the inner dimension of a and b with 0s. This is needed for matmuls not aligned to 16.
-ScaledMMConfig = namedtuple(
-    "ScaledMMConfig",
-    ["emulate", "use_fast_accum", "fp8_output", "pad_inner_dim"],
-    defaults=[False, False, False, False],
-)
+class ScaledMMConfig(NamedTuple):
```
vkuzo, line 48: if this is the object to make public, can we move it to `config.py` to keep this file all-private?

alternatively, how about just creating this from a different public config, in case that works better with torchao inference code?

torchao/float8/float8_tensor.py
```diff
@@ -47,31 +45,41 @@
...
+    """
+    Configuration for the scaled_mm in the forward and backward pass.
+
+    Attributes:
+        emulate (bool): Whether to emulate the matmuls in fp32.
+        use_fast_accum (bool): Whether to use the fast-accumulation option for scaled_mm.
+        fp8_output (bool): Whether to output the result of the scaled_mm in fp8.
```
vkuzo, line 55: in a public API I'd expect this to be a dtype enumeration

torchao/float8/inference.py
```diff
@@ -7,246 +7,32 @@
...
-        linear_mm_config: Configuration settings for the matrix multiplication
-        reduce_amax: Whether to reduce the amax (absolute maximum) among the local distributed group.
-        static_quantization_scale: Optional tensor specifying the scale for activation. Default is None.
-
-    Returns:
-        Float8Tensor: The input tensor cast to Float8 (e4m3fn) format.
-
-    Note:
-        If the input tensor is already in Float8 format, it is returned as is without re-casting.
+class Float8MMConfig(NamedTuple):
```
vkuzo, line 16: can we make the names `ScaledMMConfig` and `Float8MMConfig` not be confusing?  Ideally it should be clear why there are two objects and in which context the user should use which object

drisspg: Sure, I was going to do ScaledMMConfigInference but that is even too verbose for me, any suggestions?

vkuzo: for training I went with a user facing `Float8GemmConfig` which is a dataclass and easy to explain (IMO), the way the data gets passed around is not user facing so details of `ScaledMMConfig` don't matter as much

tbh making `ScaledMMConfig` public and reusable between training + inference sounds right to me, if you really want this to be public without a dataclass wrapper

drisspg: It can not be a dataclass today because that wont work with compile, Lazos has a PR to make frozen dataclasses proxyable  but until then it has to be a named tuple. 

I love dataclasses but I dont understand why they are more understandable then a namedtuple, Is the problem that this name is similar to other names?

PR #815, opened 2024-09-05
[StaticQuant] Update how block_size is calculated with Observers

drisspg:
Stacked PRs:
 * #807
 * __->__#815

--- --- ---

[StaticQuant] Update how block_size is calculated with Observers

### DIFF
--- a/torchao/quantization/observer.py
+++ b/torchao/quantization/observer.py
@@ -16,14 +16,40 @@ logger = logging.getLogger(__name__)
 
 @dataclass(frozen=True)
 class GranularityType:
+    """
+    Base class for representing the granularity of quantization.
+
+    This class serves as a parent for specific granularity types used in 
+    quantization operations, such as per-tensor or per-axis quantization.
+    """
     pass
 
 @dataclass(frozen=True)
 class PerTensor(GranularityType):
+    """
+    Represents per-tensor granularity in quantization.
+
+    This granularity type calcualtes the quantization parameters
+    based off the entire tensor.
+    """
     pass
 
 @dataclass(frozen=True)
 class PerAxis(GranularityType):
+    """
+    Represents per-axis granularity in quantization.
+
+    This granularity type calcualtes different quantization parameters
+    along a specified axis of the tensor.
+
+    For example if the input tensor is shape [8, 16] and axis=0, then
+    the quantization parameters are calculated for each row of the tensor.
+    Giving a total of 8 quantization parameters.
+
+
+    Attributes:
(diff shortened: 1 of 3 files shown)


torchao/quantization/observer.py
```diff
@@ -16,14 +16,35 @@
...
+    """
+    Represents per-axis granularity in quantization.
+
+    This granularity type calcualtes different quantization parameters
+    along a specified axis of the tensor.
+
+    Attributes:
+        axis (int): The axis along which reduction is performed.
+    """
     axis: int
```
drisspg, line 48: @jerryzh168 
This kinda works backwards to what I would think

e.g with tensor:
9x2048

If I was calculating with axis wise and axis = 0, my first thought would be that this would produce min_values
1x2048 e.g. along the columns and vice versa  for axis=1 but it seems that this is flipped

jerryzh168: axis = 0 means one pair of qparams per element in axis = 0, in this case it would be 9 qparams.

this is consistent with the `ch_axis` arg we have before in: https://github.com/pytorch/pytorch/blob/f2a7228aed15f97cd4389b4240fa24e25b5eeb4f/torch/ao/quantization/observer.py#L690 and https://github.com/pytorch/pytorch/blob/7b280c31bac252c7741067a68b1675aefeb1da35/aten/src/ATen/native/native_functions.yaml#L7539

torchao/quantization/observer.py
```diff
@@ -130,8 +160,13 @@ def forward(self, input: torch.Tensor):
...
 
         input_detached = input.detach()
-        if self.block_size is None:
-            self.block_size = get_block_size(input_detached.shape, self.granularity_type)
+        assert self.granularity_type is not None, "granularity_type is None"
+        block_size = get_block_size(input_detached.shape, self.granularity_type)
+
+        # If we are doing PerTensor quantization then we do not need to reduce along a given
+        # Dimension. The input tensor can vary sizes and we dont want to cache a block_size
+        if self.block_size is None and not isinstance(self.granularity_type, PerTensor):
```
jerryzh168, line 168: I think it probably makes sense to remove `self.block_size` and always recalculate the block_size

jerryzh168 (approved):
looks great, thanks for adding docs and the additional test! had one comment inline, please address the comment before merging

torchao/quantization/observer.py
drisspg: you are right I thought it was needed in choose_qparams, but we are good

PR #819, opened 2024-09-05
[Float8Quant] Add rowwise scaling option to float8 dyanmic quant

drisspg:
Stacked PRs:
 * __->__#819

--- --- ---

[Float8Quant] Add rowwise scaling option to float8 dyanmic quant

# Summary
Adds an option to do PerRow scaling of the activation and weight, which can provide faster activation quantization and better numerics.

The api for calling this is:
```Python
import torch
import torch.nn as nn
from torchao.quantization.quant_api import quantize_, PerRow, float8_dynamic_activation_float8_weight

linear_module = nn.Linear(16, 32).to("cuda")

quantize_(linear_module, float8_dynamic_activation_float8_weight(granularity=PerRow())) 
```

### DIFF
--- a/test/dtypes/test_affine_quantized_float.py
+++ b/test/dtypes/test_affine_quantized_float.py
@@ -1,34 +1,30 @@
 from torchao.utils import (
     TORCH_VERSION_AT_LEAST_2_5,
-    unwrap_tensor_subclass,
 )
 import pytest
 
 if not TORCH_VERSION_AT_LEAST_2_5:
     pytest.skip("Unsupported PyTorch version", allow_module_level=True)
 
-from numpy import full
-from torch.testing._internal.common_utils import (
-    run_tests,
-)
 from torch._inductor.test_case import TestCase as InductorTestCase
 from torch.testing._internal import common_utils
-from torch._dynamo.testing import CompileCounterWithBackend
 
 from torchao.quantization import (
     quantize_,
     float8_weight_only,
     float8_dynamic_activation_float8_weight,
 )
+from torchao.quantization.observer import PerTensor, PerRow
 from torchao.float8.float8_utils import compute_error
 import torch
 import unittest
 import pytest
-import tempfile
 import copy
 import random
-
-from unittest.mock import patch
+from functools import partial
+from typing import Tuple
+from contextlib import nullcontext
+import io
 
 
 random.seed(0)
@@ -56,6 +52,9 @@ class TestAffineQuantizedFloat8Compile(InductorTestCase):
     @common_utils.parametrize("dtype", [torch.bfloat16, torch.float32])
     @common_utils.parametrize("mode", ["dynamic", "weight-only"])
     @common_utils.parametrize("compile", [True, False])
+    @common_utils.parametrize(
+        "granularity", [PerTensor(), PerRow()] if is_H100 else [PerTensor()]
+    )
     # Inputs are (M,..), K, N
     @common_utils.parametrize(
         "sizes",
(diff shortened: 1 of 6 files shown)


torchao/quantization/quant_api.py
```diff
@@ -550,6 +552,7 @@ def apply_float8wo_quant(weight):
 def float8_dynamic_activation_float8_weight(
     activation_dtype: torch.dtype = torch.float8_e4m3fn,
     weight_dtype: torch.dtype = torch.float8_e4m3fn,
+    granularity: Literal[PerTensor, RowWise] = PerTensor,
```
drisspg, line 555: @jerryzh168 @vkuzo Thoughts on what the public API should be with this? I stole the granularity from the observers module. Note I could also do: PerAxis(0) as the only supported option

torchao/quantization/observer.py
```diff
@@ -52,6 +52,17 @@ class PerAxis(GranularityType):
     """
     axis: int
 
+@dataclass(frozen=True)
+class RowWise(GranularityType):
```
vkuzo, line 56: n00b question, why is GranularityType a set of classes instead of an enum?

if it's for storing granularity settings later, maybe `GranularityType` isn't the right name?  That name screams enum to me.

torchao/quantization/quant_api.py
vkuzo: would be good to have granularity for each tensor (for example, rowwise for input1 and tensorwise for input2).

torchao/quantization/observer.py
drisspg: I think because some subclasses want to store more information, I.e.  PerAxis stores which axis and a PerBlock would store block_size in each dim

vkuzo: call it `GranularityConfig` then?

drisspg: @jerryzh168 how does this sound, I am not sure if this API his heavily used else where or is it okay to just change the name?

jerryzh168: GranularityConfig sounds reasonable I think, it's OK to change the name, it's only used in observer right now

jerryzh168: or just `Granularity`

torchao/quantization/quant_api.py
```diff
@@ -660,23 +698,46 @@ def float8_dynamic_activation_float8_weight(
     if mm_config is None:
         mm_config = Float8MMConfig(use_fast_accum=True)
 
-    #TODO we are hardcoding TensorWise scaling, will follow up PR for Tensorwise scaling
+    activation_granularity, weight_granularity = _validate_granularity(granularity)
+
+    def get_block_size(x: torch.Tensor, granularity: _fp8_granularities):
```
jerryzh168, line 703: why is this not added to the official one in https://github.com/pytorch/ao/blob/144445a7a8e988059555421caafa17e0c1678053/torchao/quantization/observer.py#L92

torchao/quantization/quant_api.py
```diff
@@ -641,17 +643,53 @@ def apply_float8wo_quant(weight):
     return _get_linear_subclass_inserter(apply_float8wo_quant)
 
 
+_fp8_granularities = Literal[PerTensor, PerRow]
+
+
+# Validate and process granularity input
+def _validate_granularity(
```
jerryzh168, line 650: nit: maybe name should be `normalize_granularity`

torchao/quantization/quant_api.py
```diff
@@ -65,6 +65,8 @@
 )
 from torchao.float8.inference import Float8MMConfig
 
+from torchao.quantization.observer import PerTensor, PerAxis, PerRow
```
jerryzh168, line 68: nit: PerAxis is not used

test/dtypes/test_affine_quantized_float.py
```diff
@@ -68,33 +66,81 @@ class TestAffineQuantizedFloat8Compile(InductorTestCase):
...
+            ), f"Quantization error is too high got a SQNR of {error}"
+
+    def test_invalid_granularity(self):
+        with pytest.raises(ValueError, match="Invalid granularity specification"):
+            float8_dynamic_activation_float8_weight(granularity="invalid")
+
+    def test_mismatched_granularity(self):
+        with pytest.raises(
+            ValueError,
+            match="Different granularities for activation and weight are not supported",
```
jerryzh168, line 120: so will this be supported in the future?

jerryzh168 (approved):
LGTM!

torchao/quantization/observer.py
drisspg: Im going to do this in a follow up PR

test/dtypes/test_affine_quantized_float.py
drisspg: Yup, going to add as a follow up, PerTensor + PerRowise should work if you broadcast the PerTensor to RowWise

torchao/quantization/quant_api.py
drisspg: Ohh great call!

PR #823, opened 2024-09-06
Update README

msaroufim:


supriyar (approved):
looks good, but I feel we need another readme revamp before the launch.

msaroufim:
For the main README? What would you revamp?

supriyar:
> For the main README? What would you revamp?

Few things come to mind -
* Introduction can be more about the present (can move the "from the team that bought you *fast" bit to the end)
* Highlight our key perf features upfront across different models and techniques. More perf tables would be nice.
* Highlight the main APIs like autoquant clearly
* No mention of sparsity

Feel like https://github.com/unslothai/unsloth readme format is a nice one to aim for

PR #831, opened 2024-09-06
[torchao] FFHT: just expect benchmark to be installed

swolchok:
* (to be filled)

The vendored benchmark setup didn't work right for me out of
the box, so I just changed the Makefile to use pkg-config to find
benchmark wherever it is installed.

Differential Revision: [D60194969](https://our.internmc.facebook.com/intern/diff/D60194969/)


swolchok:
moving to executorch for now until we figure out C++ cross-repo dep

PR #830, opened 2024-09-06
[torchao] Remove unneeded FFHT files

swolchok:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #833
* #832
* #831
* __->__ #830
* #829
* #648
* #647
* #646

We are just using FFHT to generate a kernel, so we don't need
the Python wrapper.
Differential Revision: [D60194971](https://our.internmc.facebook.com/intern/diff/D60194971/)

PR #829, opened 2024-09-06
[torchao] Clean commit of FFHT dependency in torchao/experimental/spinquant/FFHT

swolchok:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #833
* #832
* #831
* #830
* __->__ #829
* #648
* #647
* #646

This is https://github.com/FALCONN-LIB/FFHT .

Differential Revision: [D60194972](https://our.internmc.facebook.com/intern/diff/D60194972/)

PR #647, opened 2024-08-09
RFC: quantized fast hadamard transform

swolchok:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #833
* #832
* #831
* #830
* #829
* #648
* __->__ #647
* #646

Demonstrate that we can calculate a quantized fast hadamard transform with integer math only, except for adjusting the scale of the result. (Not sure if there is a reason to actually commit this -- do we have a use case for quantized FHT on CPU?)

Differential Revision: [D60866280](https://our.internmc.facebook.com/intern/diff/D60866280/)

**NOTE FOR REVIEWERS**: This PR has internal Meta-specific changes or comments, please review them on [Phabricator](https://our.internmc.facebook.com/intern/diff/D60866280/)!

PR #646, opened 2024-08-09
[torchao] Add fast_hadamard_transform and fast_hadamard_transform_28N kernels

swolchok:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #833
* #832
* #831
* #830
* #829
* #648
* #647
* __->__ #646

This adds a pair of kernels needed for SpinQuant (and QuaRot).

Differential Revision: [D60194968](https://our.internmc.facebook.com/intern/diff/D60194968/)

PR #832, opened 2024-09-06
[torchao] FFHT: ARM NEON port

swolchok:
* #833
* __->__ #832
* #831
* #830
* #829
* #648
* #647
* #646

Patch the code generator to be capable of generating NEON
code and leave it configured to do that since we already have the
checked-in generated AVX and SSE code (and no immediate plans to use
it). Generated code size was a potential issue so I also patched the
generator to 1) reuse generated code for previous smaller sizes where
applicable and 2) choose the smallest code that isn't more than 10%
slower than the very fastest code.

Differential Revision: [D60194970](https://our.internmc.facebook.com/intern/diff/D60194970/)

### DIFF
--- a/torchao/experimental/spinquant/FFHT/fht.h
+++ b/torchao/experimental/spinquant/FFHT/fht.h
@@ -8,9 +8,13 @@ extern "C" {
 #endif
 
 int fht_float(float *buf, int log_n);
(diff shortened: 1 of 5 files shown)


swolchok:
moving to executorch for now until we figure out C++ cross-repo dep

PR #835, opened 2024-09-06
[FSDP2] cast scale to float32 in precompute

weifengpy:
revert a recent PR that breaks unit tests https://github.com/pytorch/ao/pull/727 
we can revisit if we should apply fp32 upcasting consistently across float8 compute and precompute

It failed unit test at my devgpu but not sure why our CI did not catch it. maybe because of no H100 in CI? 

`pytest -s test/float8/test_fsdp2/test_fsdp2.py -k test_transformer_parity`

### DIFF
--- a/torchao/float8/fsdp_utils.py
+++ b/torchao/float8/fsdp_utils.py
@@ -59,7 +59,7 @@ def precompute_float8_dynamic_scale_for_fsdp(module: nn.Module) -> None:
         return
 
     # inf-norm is equivalent to max(abs(w))
-    max_weights = torch._foreach_norm(weights, ord=math.inf, dtype=torch.float32)  # Partial
+    max_weights = torch._foreach_norm(weights, ord=math.inf)  # Partial
     amax_tensor = torch.stack(max_weights)  # Partial
     # clamp is dispatched through DTensor
     # it will issue a single all-reduce
@@ -69,7 +69,7 @@ def precompute_float8_dynamic_scale_for_fsdp(module: nn.Module) -> None:
         scale_tensor = torch.clamp(scale_tensor, max=torch.finfo(torch.float16).max)
     local_scale_tensor = scale_tensor.to_local()
     for i, float8_linear in enumerate(float8_linears):
-        float8_linear.weight._local_tensor._precomputed_scale = local_scale_tensor[i]
+        float8_linear.weight._local_tensor._precomputed_scale = local_scale_tensor[i].to(torch.float32)
 
 
 # FSDP pads its local tensor on dim-0. The subclass should be preserved such


torchao/float8/fsdp_utils.py
```diff
@@ -69,7 +69,7 @@ def precompute_float8_dynamic_scale_for_fsdp(module: nn.Module) -> None:
         scale_tensor = torch.clamp(scale_tensor, max=torch.finfo(torch.float16).max)
     local_scale_tensor = scale_tensor.to_local()
     for i, float8_linear in enumerate(float8_linears):
-        float8_linear.weight._local_tensor._precomputed_scale = local_scale_tensor[i]
+        float8_linear.weight._local_tensor._precomputed_scale = local_scale_tensor[i].to(torch.float32)
```
weifengpy, line 72: precompute should align with Float8Linear https://github.com/pytorch/ao/blob/144445a7a8e988059555421caafa17e0c1678053/torchao/float8/float8_utils.py#L55

weifengpy:
cc @crcrpar that doing `to(torch.float32)` at the end have better numerics. sorry for misleading

torchao/float8/fsdp_utils.py
```diff
@@ -59,7 +59,7 @@ def precompute_float8_dynamic_scale_for_fsdp(module: nn.Module) -> None:
         return
 
     # inf-norm is equivalent to max(abs(w))
-    max_weights = torch._foreach_norm(weights, ord=math.inf, dtype=torch.float32)  # Partial
+    max_weights = torch._foreach_norm(weights, ord=math.inf)  # Partial
```
awgu, line 62: wondering why remove `dtype=torch.float32` here / why does it matter?

awgu:
It might be helpful if you can provide some explanation of which tensors' dtypes changed before/after revert.

torchao/float8/fsdp_utils.py
weifengpy: this line is mimicing float8 compute when we do abs(max) without upscaling
https://github.com/pytorch/ao/blob/144445a7a8e988059555421caafa17e0c1678053/torchao/float8/float8_utils.py#L101-L102C28

not sure if it improves groudtruth numerics, but at least this brings numeric on par with float8 compute https://github.com/pytorch/ao/blob/144445a7a8e988059555421caafa17e0c1678053/test/float8/test_fsdp2/test_fsdp2.py#L124-L132

awgu:
> revert a recent PR that regressed numerics:

I would mainly suggest to not phrase it as regressing numerics. It broke tests, so there was a test regression. However, this is just because there was a numeric mismatch now where the precompute code path does some computations in fp32 where the no-precompute code path does them in bf16.

The current change in the PR makes the two the same, but it makes the precompute path less accurate than before.

weifengpy:
> I would mainly suggest to not phrase it as regressing numerics. It broke tests, so there was a test regression

good point. I modifed the PR description to focus on unit test fixing. totally get your point that true numerics might be regressing

vkuzo (approved):

PR #842, opened 2024-09-08
Don't run mac builds per commit

### LINKED ISSUE #841: [CI] Mac M1 build wheels should not run for PR commits
I notice that in my PR, the Mac M1 build wheels task runs and fails, possibly because the PR came from my forked repo https://github.com/pytorch/ao/actions/runs/10750175620/job/29816022766?pr=748

@msaroufim mentions that the build wheel task should not run per commit.

cc @seemethere

msaroufim:
Fixes #841


seemethere (approved):
LGTM

PR #850, opened 2024-09-09
Update README.md

jerryzh168:
mark uintx to be prototype


jerryzh168:
probably should separate these into a different table

PR #851, opened 2024-09-09
Update build-wheels_m1.yml

joecummings:
Test


jcaip:
Closing as we've removed all m1 specific wheels last february cc @joecummings

PR #804, opened 2024-09-04
int4 fixes and improvements

HDCharles:
Summary:

1) added int4 to autoquant using hqq by default
2) fixes to hqq in normal int4 class so it can actually be used with normal UX
3) adding hqq to eval/generate
3) eval hqq to make sure its a reasonable default for autoquant 
4) running llama3 eval now that llama3 is working correctly (fixed in 3.1 PR)
5) testing hqq v GPTQ so we have a comparison in our benchmarks/eval 
6) GPTQ was broken -> fixes to utils and GPTQ to fix broken code

Test Plan:
benchmarks.sh (new autoquant-int4 benchmarks)

export CHECKPOINT_PATH=../../../checkpoints
export MODEL_REPO=meta-llama/Llama-2-7b-chat-hf
python eval.py --checkpoint_path $CHECKPOINT_PATH/$MODEL_REPO/model.pth python eval.py --checkpoint_path $CHECKPOINT_PATH/$MODEL_REPO/model.pth --quantization int8wo python eval.py --checkpoint_path $CHECKPOINT_PATH/$MODEL_REPO/model.pth --quantization int8dq --compile python eval.py --checkpoint_path $CHECKPOINT_PATH/$MODEL_REPO/model.pth --quantization int4wo-64-hqq python eval.py --checkpoint_path $CHECKPOINT_PATH/$MODEL_REPO/model.pth --quantization int4wo-64 python eval.py --checkpoint_path $CHECKPOINT_PATH/$MODEL_REPO/model.pth --quantization int4wo-64-gptq export MODEL_REPO=meta-llama/Meta-Llama-3-8B
python eval.py --checkpoint_path $CHECKPOINT_PATH/$MODEL_REPO/model.pth python eval.py --checkpoint_path $CHECKPOINT_PATH/$MODEL_REPO/model.pth --quantization int8wo python eval.py --checkpoint_path $CHECKPOINT_PATH/$MODEL_REPO/model.pth --quantization int8dq --compile python eval.py --checkpoint_path $CHECKPOINT_PATH/$MODEL_REPO/model.pth --quantization int4wo-64-hqq python eval.py --checkpoint_path $CHECKPOINT_PATH/$MODEL_REPO/model.pth --quantization int4wo-64 python eval.py --checkpoint_path $CHECKPOINT_PATH/$MODEL_REPO/model.pth --quantization int4wo-64-gptq

(see results in README.md)

Reviewers:

Subscribers:

Tasks:

Tags:

PR #866, opened 2024-09-10
Float8 autoquant weight only

jainapurva:
Added weight-only float8 quantization technique to autoquant. It runs the default kernel where dequantize is performed on float8 weight tensor followed by linear.

Updated the safe_int_mm to work on H100 for float8 dtypes.

**Test Results**

Llama 3.1 8b on H100

**Int8 Autoquant - weight only**

Float8 autoquant, with float16 and compile

- 	Time for inference 5: 0.99 sec total
- 	Average tokens/sec: 201.71
- 	Average Bandwidth: 1519.97 GB/s
- 	Peak Memory Usage: 12.04 GB
- 	Model Size: 7.54 GB

Autoquant for float16 model and compile - int8wo

- 	Time for inference 5: 0.98 sec total
- 	Average tokens/sec: 203.48
- 	Average Bandwidth: 1535.56 GB/s
- 	Peak Memory Usage: 11.10 GB
- 	Model Size: 7.55 GB


scripts/hf_eval.py
```diff
@@ -89,7 +89,7 @@ def all_linear(mod, name):
     with torch.no_grad():
         result = evaluate(
             HFLM(
-                pretrained=model.to(device),
+                pretrained=model,
```
jerryzh168, line 92: is this change needed

scripts/prepare.sh
jerryzh168, line 1: empty file?

scripts/hf_eval.py
jainapurva: This changeis removed from current PR, will be done in separate PR.

scripts/prepare.sh
jainapurva: Removed

HDCharles (changes requested):
some additional things you should add:


1) this option should probably detect the type of gpu you're using and only add float8 if its relevant
2) make sure if you run it on H100 for Llama that when the op in question is chosen, it actually speeds things up
3) add to CI like you mention, you can see the tests in the PR: 
https://github.com/pytorch/ao/pull/804/files

jainapurva:
> some additional things you should add:
> 
> 1. this option should probably detect the type of gpu you're using and only add float8 if its relevant
> 2. make sure if you run it on H100 for Llama that when the op in question is chosen, it actually speeds things up
> 3. add to CI like you mention, you can see the tests in the PR:
>    https://github.com/pytorch/ao/pull/804/files

As discussed offline. I've added the autoquant in a separate list and architecture check. New kernels need to be added in future to improve the time. The current time is equivalent with int8.

HDCharles (approved):
looks good

nit: I would give a full list of updates in your PR description if possible
i.e. you also add H100 support for safe_int_mm

nit: it would probably be good to have a unit test for safe_int_mm on H100 even if it has to be run manually

torchao/kernel/intmm.py
```diff
@@ -69,7 +69,10 @@ def safe_int_mm(input: torch.Tensor, mat2: torch.Tensor) -> torch.Tensor:
             input = (
                 input.contiguous()
             )  # (it seems the transpose makes cublas check the above j constraint on i)
-        return out_dtype(torch.ops.aten.mm.default, torch.int32, input, mat2)
+        try:
```
jerryzh168, line 72: maybe adding a comment to this would be helpful, how these two branches are handled?

jainapurva: The except is executed if it's a float8 dtype on H100, as there's no implementation for addmm_cuda for float8 dtypes

PR #867, opened 2024-09-10
README and benchmark improvements

HDCharles:
Summary:

quantization README:

1) added fp6 to benchmarks
2) rewrote autoquant section to give a higher level explanation before
   diving into the details
3) reordered affine quantization section to first show techniques then
   dive into details
4) added fp6 section
5) moved kv cache stuff to new section
6) added sparse-marlin section and removed sparse-marlin benchmark from
   top of README since we don't have a reasonable flow for users to use
   to apply it to their model without a pre-sparsified checkpoint.
7) added uintx section

Benchmarks Changes:

1) added instructions for adding things to benchmarks so everything
   stays consistent (in llama benchmark README)
2) organized/ran benchmarks for uintx and fp6 and sparse-marlin 3) added evaluations.sh to mirror benchmarks.sh
4) added sparse-marlin to eval.py
5) fixed some generate.py logging bugs
6) improved generate help quantization help text
7) fixed some eval.py bugs with uintx
8) added marlin to eval
9) fixed eval help text

sparsity readme:
1) added some details to sparsity

Test Plan:

benchmarks.sh
evaluations.sh

Reviewers:

Subscribers:

Tasks:

Tags:


torchao/sparsity/README.md
```diff
@@ -37,7 +37,6 @@ On Meta LLama3, we observe a 25% tok/s increase (180 -> 226) compared to our exi
 | Model       | Technique               | Tokens/Second | Memory Bandwidth (GB/s) | Peak Memory (GB) | Model Size (GB) |
 | ----------- | ----------------------- | ------------- | ----------------------- | ---------------- | --------------- |
 | Llama-3-8B  | Base (bfloat16)         |   95.64       | 1435.54                 | 16.43            | 15.01           |
-|             | int8dq                  |    8.61       |   64.75                 |  9.24            |  7.52           |
```
jerryzh168, line 40: do you want to remove https://github.com/pytorch/ao/blob/e283743b3cc4612bb641b88dca3670231724d396/torchao/_models/sam/benchmark.sh#L3-L4 as well?

jerryzh168 (approved):
LGTM, thanks for making all the benchmark/eval up to date!

torchao/_models/llama/benchmark_results.txt
```diff
@@ -1,28 +1,23 @@
+README BENCHMARKS
```
jerryzh168, line 1: btw, is this file manually organized?

torchao/_models/llama/benchmarks.sh
```diff
@@ -1,44 +1,28 @@
 export CHECKPOINT_PATH=../../../checkpoints # path to checkpoints folder
```
jerryzh168, line 1: I think one improvement for this script might be to use a loop to only specify `quantization` or whatever arg that changes, instead of spelling out all the args, it might be easier to see what benchmarks we are running

torchao/sparsity/README.md
jcaip: I think this is for LLaMa only, not for sam.

jcaip (approved):
lgtm

torchao/_models/llama/benchmarks.sh
HDCharles: that would be good if we were productionizing things but most of the time i'm commenting things out and only running 1 or 2 examples and having them all spelt out is really nice

PR #869, opened 2024-09-11
[Float8] Add static constructor that will be used in Observer workflow

drisspg:
Stacked PRs:
 * __->__#869

--- --- ---

[Float8] Add static constructor that will be used in Observer workflow

### Update
After much discussion with many folks, I think we can get the best of all the worlds
- Same serde contract
- torch.compile support
- weights only loading

#### Changes
- Introduce a new Static Subclass for activation casting: `LinearActivationQuantizedTensorStatic`
- Difference between the existing is that scale + optional zero point is stored as tensor attributes (critically not captured in the quant function)
- The input_quant func contract is `(Tensor, scale: Tensor, zero_point: `Optional[Tensor]`, **quant_kwargs)` where any additional kwargs cant contain tensors as values
- We pass the kwargs at construction in order to enable weights-only serialzation and not need a new "partial" function per instance, cc @mikaylagawarecki 
- We optionally return zero_point in the flatten if it is not None, cc @bdhirsh  ( this works idk what I was doing wrong before :) )
- 

### Previous notes
This PR explores ways to quantize_ a model with static scales. It implements 3 ways to do this:

1. using the exsitng LinearActivationTensor that where the input quant func captures a tensor
2. No activation subclass tensor, but rather registering the scale buffer on the model + adding in a forward_pre_hook
3. And A static ActivationSubclass with scale attribute updating the signature for the input_quant_func callable

The first  does not worth compile due to unlatten not knowing that tensors are in the callable
The second works with compile but changes how serde needs to be done
The third Maintains serde behavior and works with compile

### DIFF
--- a/test/dtypes/test_affine_quantized_float.py
+++ b/test/dtypes/test_affine_quantized_float.py
@@ -14,6 +14,10 @@ from torchao.quantization import (
     float8_weight_only,
     float8_dynamic_activation_float8_weight,
 )
+from torchao.quantization.quant_api import (
+    float8_static_activation_float8_weight,
+)
+from torchao.quantization.quant_primitives import choose_qparams_affine, MappingType
 from torchao.quantization.observer import PerTensor, PerRow
 from torchao.float8.float8_utils import compute_error
 import torch
@@ -50,7 +54,7 @@ class TestAffineQuantizedFloat8Compile(InductorTestCase):
     @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
     @unittest.skipIf(not is_cuda_8_9, "Requires GPU with compute capability >= 8.9")
     @common_utils.parametrize("dtype", [torch.bfloat16, torch.float32])
-    @common_utils.parametrize("mode", ["dynamic", "weight-only"])
+    @common_utils.parametrize("mode", ["dynamic", "weight-only", "static"])
     @common_utils.parametrize("compile", [True, False])
     @common_utils.parametrize(
         "granularity", [PerTensor(), PerRow()] if is_H100 else [PerTensor()]
@@ -60,37 +64,49 @@ class TestAffineQuantizedFloat8Compile(InductorTestCase):
         "sizes",
         [
             ((128,), 256, 128),
-            ((256,), 512, 256),
-            ((64,), 128, 64),
             ((32, 128), 64, 256),
-            ((64, 256), 512, 128),
         ],
     )
     def test_fp8_linear_variants(
         self, dtype: torch.dtype, mode: str, compile: bool, sizes: Tuple, granularity
     ):
-        raises = (
-            isinstance(granularity, PerRow)
-            and mode == "dynamic"
-            and dtype != torch.bfloat16
+        error_message = None
+        if isinstance(granularity, PerRow):
+            if mode == "dynamic" and dtype != torch.bfloat16:
+                error_message = "PerRow quantization only works for bfloat16 precision"
+            elif mode == "static":
+                error_message = (
+                    "Static quantization only supports PerTensor granularity"
+                )
+
+        error_context = (
+            pytest.raises(AssertionError, match=error_message)
+            if error_message
+            else nullcontext()
         )
-        context = (
-            nullcontext()
-            if not raises
-            else pytest.raises(
-                AssertionError,
-                match="PerRow quantization only works for bfloat16 precision",
-            )
-        )
-        with context:
+
+        with error_context:
             M, N, K = sizes
             input_tensor = torch.randn(*M, K, dtype=dtype, device="cuda")
-
+            # Get a "reasonable" scale for the input tensor even though
+            # we use the same scale for multiple activations
+            scale, _ = choose_qparams_affine(
+                input_tensor,
+                MappingType.SYMMETRIC,
+                input_tensor.shape,
+                torch.float8_e4m3fn,
+                scale_dtype=torch.float32,
+            )
             mode_map = {
                 "dynamic": partial(
                     float8_dynamic_activation_float8_weight, granularity=granularity
                 ),
                 "weight-only": float8_weight_only,
+                "static": partial(
+                    float8_static_activation_float8_weight,
+                    scale=scale,
+                    granularity=granularity,
+                ),
             }
 
             # Create a linear layer with bfloat16 dtype
@@ -98,7 +114,7 @@ class TestAffineQuantizedFloat8Compile(InductorTestCase):
 
             quantized_model = copy.deepcopy(model)
             factory = mode_map[mode]()
-            quantize_(model, factory)
+            quantize_(quantized_model, factory)
 
             if compile:
                 quantized_model = torch.compile(quantized_model, fullgraph=True)
@@ -145,14 +161,23 @@ class TestAffineQuantizedFloat8Compile(InductorTestCase):
 
     @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
     @unittest.skipIf(not is_cuda_8_9, "Requires GPU with compute capability >= 8.9")
-    @common_utils.parametrize("mode", ["dynamic", "weight-only"])
+    @common_utils.parametrize("mode", ["dynamic", "weight-only", "static"])
     def test_serialization(self, mode: str):
         # Create and quantize the model
         model = ToyLinearModel(16, 32).to(device="cuda")
-        if mode == "dynamic":
-            factory = float8_dynamic_activation_float8_weight()
-        else:
-            factory = float8_weight_only()
+
+        mode_map = {
+            "dynamic": partial(
+                float8_dynamic_activation_float8_weight, granularity=PerTensor()
+            ),
+            "weight-only": float8_weight_only,
+            "static": partial(
+                float8_static_activation_float8_weight,
+                scale=torch.tensor(1.0, dtype=torch.float32, device="cuda"),
+                granularity=PerTensor(),
+            ),
+        }
+        factory = mode_map[mode]()
         quantize_(model, factory)
 
         # Save the state dict to an in-memory buffer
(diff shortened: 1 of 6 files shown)


torchao/quantization/quant_api.py
```diff
@@ -755,6 +767,55 @@ def apply_float8_dynamic_activation_quant(weight: torch.Tensor):
     return _get_linear_subclass_inserter(apply_float8_dynamic_activation_quant)
 
 
+def float8_static_activation_float8_weight(
+    scale: torch.Tensor,
+    activation_dtype: torch.dtype = torch.float8_e4m3fn,
+    weight_dtype: torch.dtype = torch.float8_e4m3fn,
+    granularity: Optional[
```
vkuzo, line 774: docblock for this?

vkuzo (approved):

test/dtypes/test_affine_quantized_float.py
```diff
@@ -60,45 +62,57 @@ class TestAffineQuantizedFloat8Compile(InductorTestCase):
         "sizes",
         [
             ((128,), 256, 128),
-            ((256,), 512, 256),
```
drisspg, line 63: remove some extra configs that didnt werent needed for coverage

test/dtypes/test_affine_quantized_float.py
```diff
@@ -60,45 +62,57 @@ class TestAffineQuantizedFloat8Compile(InductorTestCase):
...
+                ),
             }
 
             # Create a linear layer with bfloat16 dtype
             model = ToyLinearModel(K, N).eval().to(dtype).to("cuda")
 
             quantized_model = copy.deepcopy(model)
             factory = mode_map[mode]()
-            quantize_(model, factory)
+            quantize_(quantized_model, factory)
```
drisspg, line 115: the quantize and non quantize were flipped causing us to compile the non quantized

drisspg:
There is a interesting issue here: 
```Shell
E               torch._dynamo.exc.TorchRuntimeError: Failed running call_function <built-in function linear>(*(FakeTensor(..., device='cuda:0', size=(32, 128, 256)), LinearActivationQuantizedTensor(AffineQuantizedTensor(data=FakeTensor(..., device='cuda:0', size=(64, 256))..., shape=torch.Size([64, 256]), block_size=torch.Size([64, 256]), device=cuda:0, dtype=torch.float32, requires_grad=False, layout_tensor=Float8AQTLayout(
E               float8_data=FakeTensor(..., device='cuda:0', size=(64, 256), dtype=torch.float8_e4m3fn),
E               scale=FakeTensor(..., device='cuda:0', size=()),
E               transposed=False, layout_type=Float8LayoutType(mm_config=Float8MMConfig(emulate=False, use_fast_accum=True, pad_inner_dim=False)))), functools.partial(<function _input_quant_func_dyanmic_fp8 at 0x7f5c17584160>, activation_granularity=PerTensor(), activation_dtype=torch.float8_e4m3fn, scale=tensor(0.0105, device='cuda:0'))), None), **{}):
E               Please convert all Tensors to FakeTensors first or instantiate FakeTensorMode with 'allow_non_fake_inputs'. Found in aten.view.default(tensor(0.0105, device='cuda:0', size=()), [1, 1, 1])
E               
E               from user code:
E                  File "/home/drisspg/meta/ao/test/dtypes/test_affine_quantized_float.py", line 46, in forward
E                   x = self.linear1(x)
E                 File "/home/drisspg/meta/pytorch/torch/nn/modules/linear.py", line 125, in forward
E                   return F.linear(input, self.weight, self.bias)
E               
```

I have a stash that patched this but broke other quant. The TLDR is that the input_quant_func captures a torch.Tensor. The 
[LinearActivationQuantizedTensor](https://github.com/pytorch/ao/blob/b4d076813beb5612434ed11027cdbd8895064458/torchao/quantization/linear_activation_quantized_tensor.py#L16) stores this callable and it is specified in as a tensor attributed not a one of the tensors via unflatten. And this inner tensor doesn't get fakified. TBH I am not really sure whats the best way to handle this. cc @bdhirsh @jerryzh168

vkuzo:
> The
[LinearActivationQuantizedTensor](https://github.com/pytorch/ao/blob/b4d076813beb5612434ed11027cdbd8895064458/torchao/quantization/linear_activation_quantized_tensor.py#L16) stores this callable

I mean...can we just not store the callable for the main paths?  It seems super hard to debug if the logic that changes numerics is stored in a callable instead of being clearly spelled out.  cc @jerryzh168

torchao/quantization/quant_api.py
```diff
@@ -755,6 +776,60 @@ def apply_float8_dynamic_activation_quant(weight: torch.Tensor):
...
+        Union[_fp8_granularities, Tuple[_fp8_granularities, _fp8_granularities]]
+    ] = None,
+    mm_config: Optional[Float8MMConfig] = None,
+):
+    """
+    Applies float8 static symmetric quantization to
+
+    Args:
+        scale (torch.Tensor): The scale tensor for activation quantization. TODO: does passing in 1 tensor scale make sense?
```
drisspg, line 792: This still exists,  is agood TODO. I want this flow to work with the insert_observers_ api

```Python
def is_weight_observer(mod: torch.Tensor, name: str) -> bool:
    return isinstance(mod, torch.nn.Linear) and isinstance(
        mod.weight, LinearActivationWeightObservedTensor
    )


def float8_static_activation_float8_weight_from_observers(
    lin_mod: torch.nn.Module,
):
    act_observer = lin_mod.weight.input_observer
    assert isinstance(
        act_observer, AffineQuantizedMinMaxObserver
    ), f"Float8 static quantization requires activation observer to be AffineQuantizedMinMaxObserver, got {type(act_observer)}"
    act_scale, act_zero_point = act_observer.calculate_qparams()
    quant_func = float8_static_activation_float8_weight(act_scale)
    # I dont like this next line
    lin_mod.weight = torch.nn.Parameter(lin_mod.weight.original_weight_tensor)
    return quant_func(lin_mod)

insert_observers_(m, acts_observer, weight_observer=None)
# calibrating / training
for _ in range(10):
    m(*example_inputs)

after_obs = m(*example_inputs)

# quantized linear represented as an nn.Linear with modified tensor subclass weights
# for both activation and weight quantization
quantize_(
    m,
    float8_static_activation_float8_weight_from_observers,
    filter_fn=is_weight_observer,
)
```

In which case it gets called in a "local" sense sense and 1 scale gets associated to 1 activation, but it is "possible" to use to quantize multiple linears with the same scale. I did find after sometesting though that this might be okay and is deff helpful for testing...

torchao/quantization/quant_api.py
```diff
@@ -682,20 +686,33 @@ def _input_quant_func_dyanmic_fp8(
     x: torch.Tensor,
     activation_granularity: _fp8_granularities,
     activation_dtype: torch.dtype,
+    scale: Optional[torch.Tensor] = None,
```
jerryzh168, line 689: nit: this function is now `dynamic` or `static` quant I think

torchao/quantization/quant_api.py
```diff
@@ -740,6 +757,10 @@ def apply_float8_dynamic_activation_quant(weight: torch.Tensor):
             scale_dtype=torch.float32,
             layout_type=Float8LayoutType(mm_config=mm_config),
         )
+        def input_quant_func(x: torch.Tensor):
```
jerryzh168, line 760: this is not used?

torchao/quantization/quant_api.py
jerryzh168: what is the problem with 1 scale gets associated to 1 activation?

torchao/quantization/observer.py
```diff
@@ -5,6 +5,10 @@
     MappingType,
     ZeroPointDomain,
 )
+from torchao.utils import (
+    TorchAOBaseTensor,
```
jerryzh168, line 9: nit: not used in this file?

torchao/quantization/linear_activation_static_quantized_tensor.py
```diff
@@ -0,0 +1,193 @@
...
+    return return_and_correct_aliasing(
+        func, args, kwargs, args[0]._apply_fn_to_data(torch.t)
+    )
+
+
+to_linear_activation_quantized_static = LinearActivationQuantizedTensorStatic.from_float
+
+if TORCH_VERSION_AT_LEAST_2_5:
+    # Allow a model with LinearActivationQuantizedTensor weights to be loaded with `weights_only=True`
+    torch.serialization.add_safe_globals([LinearActivationQuantizedTensorStatic])
```
jerryzh168, line 193: nit: maybe rename it to `LinearActivationStaticQuantizedTensor` to align with the file name

torchao/quantization/linear_activation_static_quantized_tensor.py
```diff
@@ -0,0 +1,193 @@
...
+            fn(self.scale),
+            fn(self.zero_point) if self.zero_point is not None else None,
+            self.quant_kwargs,
+        )
+
+    def to(self, *args, **kwargs):
+        kwargs = self._get_to_kwargs(*args, **kwargs)
+        device = kwargs.pop("device")
+        return self.__class__(
+            self.original_weight_tensor.to(**kwargs),
```
jerryzh168, line 132: should this be device as well?

torchao/quantization/linear_activation_static_quantized_tensor.py
```diff
@@ -0,0 +1,193 @@
...
+    @staticmethod
+    def _quantized_linear_op(
+        input_tensor: torch.Tensor, weight_tensor: torch.Tensor, bias: torch.Tensor
+    ):
+        input_quant_func_static = weight_tensor.input_quant_func_static
+        original_weight_tensor = weight_tensor.original_weight_tensor
+        scale = weight_tensor.scale
+        zero_point = weight_tensor.zero_point
+        quant_kwargs = weight_tensor.quant_kwargs
+        aqt = input_quant_func_static(
```
jerryzh168, line 103: nit: aqt should probably be renamed to something else, like quantized_tensor

torchao/quantization/quant_api.py
drisspg: This is what we want, but it is pretty easy to associate 1 scale with many activations:

`quantize_(model, float8_static_activation_float8_weight(scale=torch.tensor(1.0, device="cuda")))`
Which Is probably okay in some situations but idk

jerryzh168: you mean the same scale Tensor instance is used in multiple places? that seems fine?

PR #872, opened 2024-09-11
numeric difference with/without float8 all-gather when enabling torch.compile

weifengpy:
please see minified repro in https://github.com/pytorch/ao/pull/873

close this PR for now


weifengpy:
@bdhirsh @vkuzo wonder if you have some suggestions on how to debug torch.compile numerics? My current approach is enable TORCH_LOGS="output_code" and put breakpoints/printing in output code python files

PR #873, opened 2024-09-11
[DO NOT REV]debug numerics

weifengpy:
Summary:

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:


test/float8/test_fsdp2/fsdp2_common.py
```diff
@@ -26,7 +26,7 @@ def check_parity_no_mp(
     compile_transformer_block: bool = False,
 ):
     # TODO(before land): reorder args and make config not optional
-    for iter_idx in range(10):
+    for iter_idx in range(1000):
```
weifengpy, line 29: to reproduce on a single `Float8Linear`, I have to increase the iter_idx. The discrepency happens deterministically at iter_idx=167

PR #878, opened 2024-09-12
Unify scaled INT8 mm

### LINKED ISSUE #862: Unify scaled INT8 matmul
With the new addition of INT8 mixed-precision training, there are now 2 implementations of scaled INT8 matmul (INT8 matmul + dequant)

- https://github.com/pytorch/ao/blob/main/torchao/kernel/intmm_triton.py
- https://github.com/pytorch/ao/blob/main/torchao/prototype/quantized_training/int8_mm.py

I have identified the key differences

`intmm_triton.py` | `int8_mm.py`
---|---
Only fuse act scale | Fuse both act scale and weight scale
Scale step is `acc_i32 x scale` | Scale step is cast to fp32 `acc_i32.to(f32) x scale.to(f32)`
Different autotune configs | Different autotune configs

Ideally we should only keep 1. The tedious part is to validate there is no accuracy+speed regression, regardless of which final implementation we will adopt.

Here are the places that use `intmm_triton.py`

- https://github.com/pytorch/ao/blob/main/benchmarks/intmm.py 
- https://github.com/pytorch/ao/blob/main/torchao/quantization/utils.py (I think this file is legacy from module-swap API?)
- https://github.com/pytorch/ao/blob/main/torchao/dtypes/affine_quantized_tensor.py
- https://github.com/pytorch/ao/blob/main/test/kernel/test_autotuner.py

-> Basically ensure INT8 dynamic quantization for Llama and SAM benchmarks don't regress

Here are the places that use `int8_mm.py`

- https://github.com/pytorch/ao/blob/main/torchao/prototype/quantized_training/int8_mixed_precision.py
- https://github.com/pytorch/ao/blob/main/benchmarks/quantized_training/benchmark_int8mm.py (this is similar to the benchmark script for `intmm_triton.py` above

-> Ensure INT8 mixed-precision training doesn't regress

Another question. Is it ok to change `int_scaled_matmul()` signature to accept scales for both A and B instead of only for A?

gau-nernst:
Fixes #862

### DIFF
--- a/test/kernel/test_autotuner.py
+++ b/test/kernel/test_autotuner.py
@@ -10,7 +10,6 @@ import logging
 import os
 import unittest
 
-import pytest
 import torch
 from parameterized import parameterized
 
@@ -47,17 +46,10 @@ class TestQuantFlow(unittest.TestCase):
         assert out32_1.dtype == torch.int32
         out32_2 = intmm.int_matmul(x_int, w_int)
         assert out32_2.dtype == out32_1.dtype
-        torch.testing.assert_allclose(out32_1, out32_2)
+        torch.testing.assert_close(out32_1, out32_2)
 
-    @parameterized.expand(
-        [
-            ("cuda", torch.bfloat16),
-            ("cpu", torch.bfloat16),
-            ("cuda", torch.float16),
-            ("cpu", torch.float16),
-        ]
-    )
-    def test_int_scaled_mm(self, device, dtype):
+    @parameterized.expand([("cuda",), ("cpu",),])
+    def test_int_scaled_mm(self, device):
         if device == "cuda" and not torch.cuda.is_available():
             self.skipTest(f"{device} not available")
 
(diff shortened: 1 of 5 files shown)


torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -1160,31 +1160,14 @@ def _linear_int8_act_int8_weight_check(input_tensor, weight_tensor, bias):
...
     # 1. do the matrix form of dot(X_i, W_j)
-    #
-    #
     # 2. rescale the output
-    #
-    # in cases with large matrices, y_dot_int32 can grow sufficiently
-    # large that y_dot_int32 * a float16 scale is greater than the maximum
-    # value of a float 16, (which results in a value of inf even if multiplying
-    # by the other scale would bring it within the expected range)
-
```
gau-nernst, line 1173: overflow issue is now handled inside `int_scaled_matmul()`, which always perform multiplication in FP32.

test/kernel/test_autotuner.py
```diff
@@ -47,17 +46,10 @@ def test_int_mm(self, device, dtype):
...
 
-    @parameterized.expand(
-        [
-            ("cuda", torch.bfloat16),
-            ("cpu", torch.bfloat16),
-            ("cuda", torch.float16),
-            ("cpu", torch.float16),
-        ]
-    )
-    def test_int_scaled_mm(self, device, dtype):
```
gau-nernst, line 60: dtype is always overriden to BF16 later

gau-nernst:
The triton kernel in `torchao.kernel.intmm_triton` is highly coupled to `torchao.kernel.autotuner`, which makes it unnecessary complex to separate them out. `intmm_triton` also has inference in mind, with exhaustive autotuning and cached config, which is not suitable for training. Closing this for now.

Issue #391, opened 2024-06-18
[RFC] Tensor Subclass based Quantization API


jerryzh168:
Status: Draft
Updated: 06/17/2024

# Objective
In this doc we’ll talk about Tensor subclass based quantization API for modeling users and developers.

# Modeling User API
Modeling users refer to people who use quantization APIs to quantize their model for speed up, memory saving, power saving etc. Our main goal for modeling user API is for that to be easy to use without the need to fully understand technical details.

We expect users to use two types of APIs: (1). Manual quantization API with a direct API call (2). Automatic quantization API based on some objectives. It looks like the following:
## 1. Manual API call

```
from torchao import quantize

 # 1. quantize with name of some pre-packaged quantization technique
 # easiest way to apply quantization is to use string name for
 # some pre registered techniques
 m = nn.Sequential(nn.Linear(32, 1024), nn.Linear(1024, 32))
 # current options are: "int4_weight_only", "int8_weight_only", "int8_dynamic"
 m = quantize(m, "int4_weight_only")
 
# 2. customize some methods
# We also provide functions for these techniques so people can customize them
from torchao.quantization.quant_api import int4wo
m = nn.Sequential(nn.Linear(32, 1024), nn.Linear(1024, 32))
m = quantize(m, int4wo(groupsize=32))

# 3. write your own new apply_tensor_subclass
# You can also add your own apply_tensor_subclass by manually calling tensor subclass constructor
# on weight

# let's say user wants to use a new quantization technique/dtype that exposed factory function: to_affine_quantized

from torchao.dtypes import to_affine_quantized

# 3.1. weight only quantization with `to_affine_quantized`
def apply_to_affine_quantized(weight):
    # settings
    return to_affine_quantized(weight, ...)

model = quantize(model, apply_to_affine_quantized)

# 3.2. dynamic quant with `to_affine_quantized` for both input and weight

from torchao.quantization.quant_api import to_linear_act_quantized
def apply_dynamic_quant_my_dtype(weight):
    # weight settings
    ...

    def input_act_quant_func(x):
        # input settings
        return to_affine_quantized(x, ...)
    
    weight = to_affine_quantized(weight, ...)
    weight = to_linear_act_quantized(weight, input_act_quant_func)
    return weight

quantize(model, apply_dynamic_quant_my_dtype)

# 3.3. static quant using `my_dtype` for both input and weight
# A working example is work in progress, we'll update this later.

```

[...]

PR #880, opened 2024-09-12
W4A8 based on CUTLASS

alexsamardzic:
CUTLASS-based `s8s4_linear_cutlass()` operator is introduced, performing linear transformation over quantized 8-bit input and quantized 4-bit weight tensors, with corresponding floating point scale tensors attached.
    
A benchmark script, for comparing performance of MM based on this linear operator with MM over 16-bit floating point tensors is supplied in `benchmarks/benchmarks/benchmark_s8s4_cutlass.py`.
    
The Llama generator `script torchao/_models/llama/generate.py` is changed, to add `int8adq-int4w-symm` quantization as an option, that will in turn activate `s8s4_linear_cutlass()` operator.  With this type of quantization activated, i.e. if `generate.py` script run as follows:
```
python generate.py --compile --precision=torch.float16 -q int8adq-int4w-symm
```
the generator achieves around 133 tok/sec on A100, vs. around 93 tok/sec without quantization, i.e. when `generate.py` script run as follows:
```
python generate.py --compile --precision=torch.float16
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -93,6 +93,18 @@ def get_extensions():
             extra_compile_args["nvcc"].append("-g")
             extra_link_args.append("/DEBUG")
 
+    use_cutlass = False
+    if use_cuda and not IS_WINDOWS:
+        use_cutlass = True
+        this_dir = os.path.abspath(os.path.curdir)
+        cutlass_dir = os.path.join(this_dir, "third_party", "cutlass")
+        cutlass_include_dir = os.path.join(cutlass_dir, "include")
+    if use_cutlass:
+        extra_compile_args["nvcc"].extend([
+            "-DTORCHAO_USE_CUTLASS",
+            "-I" + cutlass_include_dir,
+        ])
+
     this_dir = os.path.dirname(os.path.curdir)
     extensions_dir = os.path.join(this_dir, "torchao", "csrc")
     sources = list(glob.glob(os.path.join(extensions_dir, "**/*.cpp"), recursive=True))
--- a/test/test_s8s4_linear_cutlass.py
+++ b/test/test_s8s4_linear_cutlass.py
@@ -1,0 +1,80 @@
+import itertools
+
+import torch
+
+import torchao
+from torchao.ops import s8s4_linear_cutlass
+from torchao.quantization.utils import group_quantize_tensor_symmetric
+from torchao.utils import compute_max_diff
+
+import pytest
+
+
+S8S4_LINEAR_CUTLASS_DTYPE = [torch.float16, torch.bfloat16]
+S8S4_LINEAR_CUTLASS_BATCH_SIZE = [1, 4, 8, 16, 32, 64]
+S8S4_LINEAR_CUTLASS_SIZE_MNK = [
+    (2, 512, 128),
+    (3, 2048, 2048),
+    (4, 3584, 640),
+    (13, 8704, 8576),
+    (26, 18944, 1664),
+    (67, 6656, 1408),
+]
+S8S4_LINEAR_CUTLASS_USE_BIAS = [False, True]
+S8S4_LINEAR_CUTLASS_TEST_PARAMS = list(
+    itertools.product(
+        S8S4_LINEAR_CUTLASS_DTYPE,
+        S8S4_LINEAR_CUTLASS_BATCH_SIZE,
+        S8S4_LINEAR_CUTLASS_SIZE_MNK,
+        S8S4_LINEAR_CUTLASS_USE_BIAS,
+    )
+)
+
+
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA not available")
+@pytest.mark.parametrize(
+    "dtype, batch_size, size_mnk, use_bias", S8S4_LINEAR_CUTLASS_TEST_PARAMS
+)
+def test_s8s4_linear_cutlass(dtype, batch_size, size_mnk, use_bias):
+    size_m, size_n, size_k = size_mnk
+
+    input = torch.randn((batch_size, size_m, size_k), dtype=dtype, device="cuda")
+    weight = torch.rand((size_n, size_k), dtype=dtype, device="cuda")
+    bias = torch.rand((size_n,), dtype=dtype, device="cuda") if use_bias else None
+
+    input_2d = input.view(-1, input.shape[-1])
+    input_2d_s8, input_2d_scales, input_2d_zeros = group_quantize_tensor_symmetric(
+        input_2d, 8, size_k, dtype
+    )
+    assert torch.all(input_2d_zeros == 0)
+    input_s8 = input_2d_s8.reshape(input.shape)
+    input_scales = input_2d_scales.reshape(input.shape[:-1])
+
+    weight_s8, weight_scales, weight_zeros = group_quantize_tensor_symmetric(
+        weight, 4, size_n, dtype
+    )
+    assert torch.all(weight_zeros == 0)
+    weight_s4 = ((weight_s8[:, 1::2] & 0xF) << 4) | (weight_s8[:, 0::2] & 0xF)
+
+    # If torch.nn.functional.linear(input, weight, bias) used as
+    # reference, the error would be too big.  The calculation below is
+    # approximately what s8s4_linear_cutlass kernel is doing (except
+    # that matrrix multiplication is over integers there)).
+    size_m_2d = input_2d.shape[0]
+    output_ref = (
+        (input_2d_s8.to(dtype) @ weight_s8.to(dtype).T)
+        * input_2d_scales.view(size_m_2d, 1)
+        * weight_scales.view(1, size_n)
+    )
+    if bias is not None:
+        output_ref += bias
+    output_ref = output_ref.reshape(input.shape[:-1] + (size_n,))
+
+    fn_inputs = (input_s8, input_scales, weight_s4, weight_scales, bias)
+    try:
+        output = s8s4_linear_cutlass(*fn_inputs)
+    except NotImplementedError as e:
+        pytest.xfail("s8s4_linear_cutlass() op not implemented")
+
+    max_diff = compute_max_diff(output, output_ref)
+    assert max_diff < 5e-3
--- a/torchao/csrc/cuda/s8s4_linear_cutlass/s8s4_linear_cutlass.cu
+++ b/torchao/csrc/cuda/s8s4_linear_cutlass/s8s4_linear_cutlass.cu
@@ -1,0 +1,536 @@
+#include <torch/extension.h>
+
+#include <ATen/ATen.h>
+#include <ATen/core/Tensor.h>
+#include <ATen/cuda/CUDAUtils.h>
+#include <c10/util/Exception.h>
+
+#if defined(TORCHAO_USE_CUTLASS) && !defined(_WIN32) &&                   \
+    defined(CUDA_VERSION) && (CUDA_VERSION >= 11080)
+#define BUILD_S8S4_LINEAR_CUTLASS
+#endif
+
+#if defined(BUILD_S8S4_LINEAR_CUTLASS)
+#include <cuda_runtime.h>
+#include <cutlass/cutlass.h>
+#include <cutlass/gemm/device/gemm_universal.h>
+#include <cutlass/epilogue/threadblock/fusion/visitors.hpp>
+#include <cutlass/gemm/kernel/default_gemm_universal_with_visitor.h>
+
+#define CUTLASS_STATUS_CHECK(status)                                      \
+  {                                                                       \
+    TORCH_CHECK(status == cutlass::Status::kSuccess,                      \
+                __func__, " : Got CUTLASS error: ",                       \
+                cutlassGetStatusString(status));                          \
+  }
+#endif
+
+namespace torchao {
+
+#if defined(BUILD_S8S4_LINEAR_CUTLASS)
+template<
+    typename ElementA,
+    typename ElementAScale,
+    typename ElementB,
+    typename ElementBScale,
+    typename ElementC,
+    typename ElementAccumulator,
+    typename ElementEpilogue,
+    typename ElementOutput,
+    typename ThreadblockShape,
+    typename WarpShape,
+    typename InstructionShape,
+    int NumStages,
+    bool use_tensor_c>
+void s8s4_linear_kernel_cutlass(
+    const at::Tensor& tensor_a, const at::Tensor& tensor_a_scale,
+    const at::Tensor& tensor_b, const at::Tensor& tensor_b_scale,
+    const at::Tensor& tensor_c, at::Tensor& tensor_d) {
+  using LayoutA = cutlass::layout::RowMajor;
+  using LayoutB = cutlass::layout::ColumnMajor;
+  using LayoutC = cutlass::layout::RowMajor;
+
+  const int m = tensor_a.size(0);
+  const int n = tensor_b.size(0);
+  const int k = tensor_a.size(1);
+
+  constexpr int AlignmentA = 128 / cutlass::sizeof_bits<ElementA>::value;
+  constexpr int AlignmentAScale =
+    128 / cutlass::sizeof_bits<ElementAScale>::value;
+  constexpr int AlignmentB = 128 / cutlass::sizeof_bits<ElementB>::value;
+  constexpr int AlignmentBScale =
+    128 / cutlass::sizeof_bits<ElementBScale>::value;
+  constexpr int AlignmentC = 128 / cutlass::sizeof_bits<ElementC>::value;
+  constexpr int AlignmentOutput =
+    128 / cutlass::sizeof_bits<ElementOutput>::value;
+
+  // Check for current CUTLASS limitations w.r.t. alignments.
+  TORCH_CHECK(k % AlignmentA == 0,
+              __func__, " : Number of columns of tensor A must be divisible ",
+              "by ", AlignmentA);
+  TORCH_CHECK(k % AlignmentB == 0,
+              __func__, " : Number of columns of tensor B must be divisible ",
+              "by ", AlignmentB);
+  TORCH_CHECK(n % AlignmentC == 0,
+              __func__, " : Number of columns of tensor C must be divisible ",
+              "by ", AlignmentC);
+
+  using SmArch = cutlass::arch::Sm80;
+  using ThreadblockSwizzle =
+      cutlass::gemm::threadblock::ThreadblockSwizzleStreamK;
+
+  constexpr auto NumEVTEpilogueStages = 1;
+
+  using TensorAScaleTileThreadMap =
+      cutlass::epilogue::threadblock::OutputTileThreadLayout<
+          ThreadblockShape,
+          WarpShape,
+          ElementAScale,
+          AlignmentAScale,
+          NumEVTEpilogueStages>;
+  using TensorBScaleTileThreadMap =
+      cutlass::epilogue::threadblock::OutputTileThreadLayout<
+          ThreadblockShape,
+          WarpShape,
+          ElementBScale,
+          AlignmentBScale,
+          NumEVTEpilogueStages>;
+  using TensorCTileThreadMap =
+      cutlass::epilogue::threadblock::OutputTileThreadLayout<
+          ThreadblockShape,
+          WarpShape,
+          ElementC,
+          AlignmentC,
+          NumEVTEpilogueStages>;
+  using OutputTileThreadMap =
+      cutlass::epilogue::threadblock::OutputTileThreadLayout<
+          ThreadblockShape,
+          WarpShape,
+          ElementOutput,
+          AlignmentOutput,
+          NumEVTEpilogueStages>;
+
+  using Accum = cutlass::epilogue::threadblock::VisitorAccFetch;
+
+  using TensorAScale =
+      cutlass::epilogue::threadblock::VisitorColBroadcast<
+          TensorAScaleTileThreadMap,
+          ElementAScale,
+          cute::Stride<cute::_1, cute::_0, int64_t>>;
+  using TensorAScaleArguments = typename TensorAScale::Arguments;
+
+  using TensorBScale =
+      cutlass::epilogue::threadblock::VisitorRowBroadcast<
+          TensorBScaleTileThreadMap,
+          ElementBScale,
+          cute::Stride<cute::_0, cute::_1, int64_t>>;
+  using TensorBScaleArguments = typename TensorBScale::Arguments;
+
+  using TensorCScalar =
+      cutlass::epilogue::threadblock::VisitorScalarBroadcast<ElementC>;
+  using TensorCTensor =
+      cutlass::epilogue::threadblock::VisitorRowBroadcast<
+          TensorCTileThreadMap,
+          ElementC,
+          cute::Stride<cute::_0, cute::_1, int32_t>>;
+  using TensorC =
+      std::conditional_t<use_tensor_c, TensorCTensor, TensorCScalar>;
+  using TensorCArguments = typename TensorC::Arguments;
+
+  using ApplyAScale = cutlass::epilogue::threadblock::VisitorCompute<
+      cutlass::multiplies, ElementEpilogue, ElementEpilogue,
+      cutlass::FloatRoundStyle::round_to_nearest
+  >;
+  using EVTApplyAScale = cutlass::epilogue::threadblock::Sm80EVT<
+      ApplyAScale,
+      Accum,
+      TensorAScale>;
+
+  using ApplyBScale = cutlass::epilogue::threadblock::VisitorCompute<
+      cutlass::multiplies, ElementEpilogue, ElementEpilogue,
+      cutlass::FloatRoundStyle::round_to_nearest
+  >;
+  using EVTApplyBScale = cutlass::epilogue::threadblock::Sm80EVT<
+      ApplyBScale,
+      EVTApplyAScale,
+      TensorBScale>;
+
+  using ApplySum = cutlass::epilogue::threadblock::VisitorCompute<
+      cutlass::plus, ElementEpilogue, ElementEpilogue,
+      cutlass::FloatRoundStyle::round_to_nearest
+  >;
+  using EVTApplySum = cutlass::epilogue::threadblock::Sm80EVT<
+      ApplySum,
+      EVTApplyBScale,
+      TensorC>;
+
+  using Output = cutlass::epilogue::threadblock::VisitorAuxStore<
+      OutputTileThreadMap, ElementOutput,
+      cutlass::FloatRoundStyle::round_to_nearest,
+      cute::Stride<int64_t, cute::_1, int64_t> // StrideMNL
+  >;
+
+  using EVTOutput = cutlass::epilogue::threadblock::Sm80EVT<
+      Output,
+      EVTApplySum>;
+
+  using EVTKernel =
+      typename cutlass::gemm::kernel::DefaultGemmWithVisitor<
+      ElementA, LayoutA, cutlass::ComplexTransform::kNone, AlignmentA,
+      ElementB, LayoutB, cutlass::ComplexTransform::kNone, AlignmentB,
+      ElementC, LayoutC, AlignmentC,
+      ElementAccumulator,
+      ElementEpilogue,
+      cutlass::arch::OpClassTensorOp,
+      SmArch,
+      ThreadblockShape,
+      WarpShape,
+      InstructionShape,
+      EVTOutput,
+      ThreadblockSwizzle,
+      NumStages,
+      cutlass::arch::OpMultiplyAddMixedInputUpcast,
+      NumEVTEpilogueStages
+  >::GemmKernel;
+
+  using Gemm = cutlass::gemm::device::GemmUniversalBase<EVTKernel>;
+
+  cutlass::gemm::GemmCoord problem_size(m, n, k);
+  constexpr auto SplitKFactor = 1;
+
+  TensorAScaleArguments tensor_a_scale_arguments{
+    (ElementAScale*)tensor_a_scale.data_ptr(),
+    ElementAScale(1),
+    {cute::_1{}, cute::_0{}, problem_size.m()}
+  };
+  TensorBScaleArguments tensor_b_scale_arguments{
+    (ElementBScale*)tensor_b_scale.data_ptr(),
+    ElementBScale(1),
+    {cute::_0{}, cute::_1{}, problem_size.n()}
+  };
+  TensorCArguments tensor_c_arguments{
+    [&]() -> TensorCArguments {
+      if constexpr (use_tensor_c) {
+        return {(ElementC*)tensor_c.data_ptr(),
+                ElementC(0),
+                {cute::_0{}, cute::_1{}, problem_size.n()}};
+      } else {
+        return {ElementC(0)};
+      }
+    }()
+  };
+  typename Output::Arguments output_arguments{
+    (ElementOutput*)tensor_d.data_ptr(),
+    {problem_size.n(), cute::_1{}, problem_size.mn().product()}
+  };
+  typename EVTOutput::Arguments callback_arguments{
+    {
+      {
+        {
+          {},                        // Accum
+          tensor_a_scale_arguments,  // TensorAScale
+          {}                         // ApplyAScale
+        },                           // EVTApplyAScale
+        tensor_b_scale_arguments,    // TensorBScale
+        {},                          // ApplyBScale
+      },                             // EVTApplyBScale
+      tensor_c_arguments,            // TensorC
+      {}                             // ApplySum
+    },                               // EVTApplySum
+    output_arguments                 // Output
+  };                                 // EVTOutput
+  constexpr auto AvailSms = -1;
+
+  typename Gemm::Arguments arguments(
+    cutlass::gemm::GemmUniversalMode::kGemm,
+    problem_size,
+    SplitKFactor,
+    callback_arguments,              // arguments of EVT callbacks
+    (ElementA*)tensor_a.data_ptr(),
+    (ElementB*)tensor_b.data_ptr(),
+    nullptr,                         // ptr C (unused)
+    nullptr,                         // ptr D (unused)
+    problem_size.mk().product(),     // batch stride A
+    problem_size.nk().product(),     // batch stride B
+    0,                               // batch stride C (unused)
+    0,                               // batch stride D (unused)
+    problem_size.k(),                // stride A
+    problem_size.k(),                // stride B
+    0,                               // stride C (unused)
+    0,                               // stride D (unused)
+    AvailSms);
+
+  Gemm gemm_op;
+
+  cutlass::Status status;
+
+  // Verify that GEMM operation with given arguments can be performed
+  // by CUTLASS.
+  status = gemm_op.can_implement(arguments);
+  CUTLASS_STATUS_CHECK(status);
+
+  // Allocate workspace for CUTLASS mixed datatypes GEMM kernel.
+  const auto workspace_size = Gemm::get_workspace_size(arguments);
+  auto workspace = tensor_a.new_empty({(int64_t)workspace_size},
+                                      at::TensorOptions().dtype(at::kByte));
+
+  // Initialize CUTLASS mixed datatypes GEMM object.
+  status = gemm_op.initialize(arguments, workspace.data_ptr(),
+                              at::cuda::getCurrentCUDAStream());
+  CUTLASS_STATUS_CHECK(status);
+
+  // Perform mixed datatypes GEMM operation.
+  status = gemm_op.run(at::cuda::getCurrentCUDAStream());
+  CUTLASS_STATUS_CHECK(status);
+}
+
+template<
+    typename ElementA,
+    typename ElementAScale,
+    typename ElementB,
+    typename ElementBScale,
+    typename ElementC,
+    typename ElementAccumulator,
+    typename ElementEpilogue,
+    typename ElementOutput,
+    bool use_tensor_c>
+void
+s8s4_linear_cutlass_dispatch_shapes(
+    const at::Tensor& tensor_a, const at::Tensor& tensor_a_scale,
+    const at::Tensor& tensor_b, const at::Tensor& tensor_b_scale,
+    const at::Tensor& tensor_c, at::Tensor& tensor_d) {
+  using InstructionShape = cutlass::gemm::GemmShape<16, 8, 32>;
+
+  // A minimal heuristic to improve performance for small number of
+  // inputs cases.
+  if (tensor_a.size(0) <= 16) {
+    using ThreadblockShape = cutlass::gemm::GemmShape<16, 128, 128>;
+    using WarpShape = cutlass::gemm::GemmShape<16, 32, 128>;
+    constexpr auto NumStages = 6;
+    s8s4_linear_kernel_cutlass<
+        ElementA, ElementAScale, ElementB, ElementBScale, ElementC,
+        ElementAccumulator, ElementEpilogue, ElementOutput,
+        ThreadblockShape, WarpShape, InstructionShape, NumStages, use_tensor_c>(
+            tensor_a, tensor_a_scale, tensor_b, tensor_b_scale, tensor_c,
+            tensor_d);
+  } else if (tensor_a.size(0) <= 32) {
+    using ThreadblockShape = cutlass::gemm::GemmShape<32, 128, 128>;
+    using WarpShape = cutlass::gemm::GemmShape<32, 32, 128>;
+    constexpr auto NumStages = 5;
+    s8s4_linear_kernel_cutlass<
+        ElementA, ElementAScale, ElementB, ElementBScale, ElementC,
+        ElementAccumulator, ElementEpilogue, ElementOutput,
+        ThreadblockShape, WarpShape, InstructionShape, NumStages, use_tensor_c>(
+            tensor_a, tensor_a_scale, tensor_b, tensor_b_scale, tensor_c,
+            tensor_d);
+  } else {
+    using ThreadblockShape = cutlass::gemm::GemmShape<64, 128, 128>;
+    using WarpShape = cutlass::gemm::GemmShape<64, 32, 128>;
+    constexpr auto NumStages = 4;
+    s8s4_linear_kernel_cutlass<
+        ElementA, ElementAScale, ElementB, ElementBScale, ElementC,
+        ElementAccumulator, ElementEpilogue, ElementOutput,
+        ThreadblockShape, WarpShape, InstructionShape, NumStages, use_tensor_c>(
+            tensor_a, tensor_a_scale, tensor_b, tensor_b_scale, tensor_c,
+            tensor_d);
+  }
+}
+
+#endif
+
+// Perform linear operation, using corresponding CUTLASS mixed
+// data-types GEMM kernel, to given arguments:
+//   result = (input * input_scale) @ (weight * weight_scale).T + bias
+// Notes: The "input_scale" tensor is expected to be a vector, of size
+// equal to number of rows of "input" tensor.  The "weight_scale"
+// tensor is expected to be a vector, of size equal to number of rows
+// of "weight" tensor. The "bias" tensor is expected to be a vector,
+// of size equal to number of rows of "weight" tensor.
+at::Tensor
+s8s4_linear_cutlass(const at::Tensor& input, const at::Tensor& input_scale,
+                    const at::Tensor& weight, const at::Tensor& weight_scale,
+                    const at::Tensor& bias) {
+#if defined(BUILD_S8S4_LINEAR_CUTLASS)
+  // For now, only CC 8.x devices are supported.
+  const auto dprops = at::cuda::getCurrentDeviceProperties();
+  const auto is_sm8x = dprops->major == 8;
+  TORCH_CHECK(is_sm8x,
+              __func__, " : Supported only on GPUs with compute capability "
+              "8.x");
+
+  // Validate datatypes of arguments.
+  TORCH_CHECK(input.dtype() == at::kChar,
+              __func__, " : The input datatype ", input.dtype(),
+              " not supported");
+  TORCH_CHECK(input_scale.dtype() == at::kHalf ||
+              input_scale.dtype() == at::kBFloat16,
+              __func__, " : The input scale datatype ", input_scale.dtype(),
+              " not supported");
+  TORCH_CHECK(weight.dtype() == at::kChar, " : The weight datatype ",
+              weight.dtype(), " not supported");
+  TORCH_CHECK(weight_scale.dtype() == input_scale.dtype(),
+              __func__, " : Expected weight scale datatype ",
+              input_scale.dtype(), ", got ", weight_scale.dtype());
+  if (bias.numel() > 0) {
+    TORCH_CHECK(bias.dtype() == input_scale.dtype(),
+                __func__, " : Expected bias datatype ", input_scale.dtype(),
+                ", got ", bias.dtype());
+  }
+
+  // Validate layouts of arguments.
+  TORCH_CHECK(input.dim() >= 2,
+              __func__, " : Expected input argument to be 2D or "
+              "higher-dimensional tensor, got ", input.dim(), " dims");
+  TORCH_CHECK(input.layout() == at::Layout::Strided,
+              __func__, " : Expected input argument to be strided, got layout ",
+              input.layout());
+  TORCH_CHECK(input_scale.dim() == input.dim() - 1,
+              __func__, " : Expected input scale argument to be ",
+              input.dim() - 1, "D tensor, got ", input_scale.dim(), " dims");
+  TORCH_CHECK(input_scale.layout() == at::Layout::Strided,
+              __func__, " : Expected input scale argument to be strided, got "
+              "layout ", input_scale.layout());
+  TORCH_CHECK(weight.dim() == 2,
+              __func__, " : Expected weight argument to be 2D tensor, got ",
+              weight.dim(), " dims");
+  TORCH_CHECK(weight.layout() == at::Layout::Strided,
+              __func__,
+              " : Expected weight argument to be strided, got layout ",
+              weight.layout());
+  TORCH_CHECK(weight_scale.dim() == 1 || weight_scale.dim() == 2,
+              __func__, " : Expected weight scale argument to be 1D or 2D ",
+              "tensor, got ", weight_scale.dim(), " dims");
+  TORCH_CHECK(weight_scale.layout() == at::Layout::Strided,
+              __func__, " : Expected weight scale argument to be strided, got "
+              "layout ", weight_scale.layout());
+  if (bias.numel() > 0) {
+    TORCH_CHECK(bias.dim() == 1,
+                __func__, " : Expected bias argument to be 1D tensor, got ",
+                bias.dim(), " dims");
+    TORCH_CHECK(bias.layout() == at::Layout::Strided,
+                __func__, " : Expected bias argument to be strided, got ",
+                "layout ", bias.layout());
+  }
+
+  // Squash the input tensor to 2D tensor.
+  const auto input_sizes = input.sizes().vec();
+  const auto input_2d = input.reshape({-1, input_sizes.back()});
+  const auto input_scale_sizes = input_scale.sizes().vec();
+  const auto input_scale_1d = input_scale.reshape({-1});
+  const auto weight_scale_1d = weight_scale.reshape({-1});
+
+  // Validate sizes of arguments.
+  TORCH_CHECK(input_2d.size(1) == 2 * weight.size(1),
+              __func__, " : Expected input argument to have ",
+              2 * weight.size(1), " columns, but got ", input_2d.size(1));
+  for (auto i = 0; i < input_scale_sizes.size(); ++i)
+    TORCH_CHECK(input_scale_sizes[i] == input_sizes[i],
+                __func__, " : Expected input scale argument size at position ",
+                i, " to be ", input_sizes[i], ", but got ",
+                input_scale_sizes[i]);
+  TORCH_CHECK(weight_scale_1d.numel() == weight.size(0),
+              __func__, " : Expected weight scale argument to have ",
+              weight.size(0), " elements, got ", weight_scale_1d.numel(),
+              " elements");
+  if (bias.numel() > 0) {
+    TORCH_CHECK(bias.numel() == weight.size(0),
+                __func__, " : Expected bias argument to have ", weight.size(0),
+                " elements, got ", bias.numel(), " elements");
+  }
+
+  // Validate strides of arguments.
+  const auto input_2d_strides = input_2d.strides();
+  TORCH_CHECK(input_2d_strides[0] >= 1 && input_2d_strides[1] == 1,
+              __func__, " : Expected input argument in row-major layout");
+  const auto input_scale_1d_strides = input_scale_1d.strides();
+  TORCH_CHECK(input_scale_1d_strides[0] == 1,
+              __func__, " : Expected input scale argument to be contiguous");
+  const auto weight_strides = weight.strides();
+  TORCH_CHECK(weight_strides[0] >= 1 && weight_strides[1] == 1,
+              __func__, " : Expected weight argument in row-major layout");
+  const auto weight_scale_1d_strides = weight_scale_1d.strides();
+  TORCH_CHECK(weight_scale_1d_strides[0] == 1,
+              __func__, " : Expected weight scale argument to be contiguous");
+  if (bias.numel() > 0) {
+    const auto bias_strides = bias.strides();
+    TORCH_CHECK(bias_strides[0] == 1,
+                __func__, " : Expected bias argument to be contiguous");
+  }
+
+  // Introduce alias names for arguments, according to the CUTLASS
+  // naming conventions.
+  const auto& tensor_a = input_2d;
+  const auto& tensor_a_scale = input_scale_1d;
+  const auto& tensor_b = weight;
+  const auto& tensor_b_scale = weight_scale_1d;
+  const auto& tensor_c = bias;
+
+  // Create output tensor.
+  at::Tensor tensor_d =
+      tensor_a_scale.new_empty({tensor_a.size(0), tensor_b.size(0)});
+
+  using ElementA = int8_t;
+  using ElementB = cutlass::int4b_t;
+  using ElementAccumulator = int32_t;
+  AT_DISPATCH_SWITCH(
+    input_scale.scalar_type(),
+    "s8s4_linear_cutlass",
+    AT_DISPATCH_CASE(
+      at::ScalarType::Half,
+      [&]() {
+        using ElementAScale = cutlass::half_t;
+        using ElementBScale = cutlass::half_t;
+        using ElementC = cutlass::half_t;
+        using ElementEpilogue = float;
+        using ElementOutput = cutlass::half_t;
+        if (bias.numel() > 0) {
+          s8s4_linear_cutlass_dispatch_shapes<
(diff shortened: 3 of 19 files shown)


cpuhrsch:
@alexsamardzic - Can we use the CUTLASS that ships with PyTorch? As in, should we change PyTorch to ship the headers used to build its CUTLASS kernels / does the PyTorch nightly already ship those? I see the test is using group size 128. I think it's ok if we don't necessarily support all group sizes or shapes right away.

We have some int4 support via the pattern matched in https://github.com/pytorch/pytorch/blob/main/torch/_inductor/fx_passes/post_grad.py#L345-L403 which dispatches to https://github.com/pytorch/pytorch/blob/dab7d646d55a2b6696d51dee4816a6743ec1ae5a/torch/_inductor/kernel/unpack_mixed_mm.py#L76 - would an extension for int4x2 X int8 of this be interesting here?

alexsamardzic:
> I'm on PTO today and tomorrow so will review asap, apologies for the delay

Thanks Mark - it's really just a draft, so not yet ready for review, but it would be useful to discuss points that I mentioned in my comment above.

alexsamardzic:
> @alexsamardzic - Can we use the CUTLASS that ships with PyTorch? As in, should we change PyTorch to ship the headers used to build its CUTLASS kernels / does the PyTorch nightly already ship those?

This CUTLASS version is also lagging behind.  My CUTLASS PR with mixed int4/int8 GEMM is merged after the latest (3.5.1) CUTLASS release, hopefully there will be a new release soon.  But in any case, this is a kind of problem that we'll have if we use more CUTLASS from torchao - for lots of time, the torchao build will have to be pointed to a bleeding edge CUTLASS checkout.

> I see the test is using group size 128. I think it's ok if we don't necessarily support all group sizes or shapes right away.

It uses group size 128 in order to force weight scale to be a vector, and not a matrix.  I tried to explain the issue in my comment above, if group quantization is obligatory here, then it's going to be rather complicated to make this work.

> We have some int4 support via the pattern matched in https://github.com/pytorch/pytorch/blob/main/torch/_inductor/fx_passes/post_grad.py#L345-L403 which dispatches to https://github.com/pytorch/pytorch/blob/dab7d646d55a2b6696d51dee4816a6743ec1ae5a/torch/_inductor/kernel/unpack_mixed_mm.py#L76 - would an extension for int4x2 X int8 of this be interesting here?

I'm just looking into the quantization code, to see is it possible to do it there - it's not hard to make this change, but CUTLASS in general doesn't support doing things before GEMM (while fusing operations after GEMM calculated is reasonably well supported), so it would be the best if the quantization code actually put the weight values in int4x2 format.

alexsamardzic:
Updated so that there is a new `int8_dynamic_activation_int4_weight_cutlass` quantization method available that, for now, would quantize both input and weight symmetrically, and won't use group quantization for weight (so weight scales are always a vector).  It should be now possible to try kernel on arbitrary models, if quantized by above quantization method.

torchao/quantization/quant_api.py
```diff
@@ -506,6 +508,41 @@ def int8_dynamic_activation_int4_weight(group_size=32, mapping_type=MappingType.
     return _get_linear_subclass_inserter(apply_int8_dynamic_activation_int4_weight_quant, group_size=group_size, mapping_type=mapping_type)
 
 
+def apply_int8_dynamic_activation_int4_weight_quant_cutlass(weight):
```
jerryzh168, line 511: can this be represented as a different `Layout` for int8 dynamic activation/int4 weight quantization? docs for Packing/Layout can be found in https://github.com/pytorch/ao/issues/391 "Layout and Packing" and simplified example in https://github.com/pytorch/ao/blob/main/tutorials/developer_api_guide/my_dtype_tensor_subclass.py

alexsamardzic: Thanks for the pointer!  Yes, this will need refinement on this and several other places, as I learn about doing things the "torchao way"; but my main goal initially is to connect the dots, so that some benchmarks could be run, and that we could verify that CUTLASS provides some value here.

alexsamardzic:
Made some minor updates, including added support for bfloat16.

<details>
 <summary>Micro-benchmarking script</summary>

```
import copy

import torch

from torchao.utils import (
    TORCH_VERSION_AT_LEAST_2_5,
    unwrap_tensor_subclass,
)
    quantize_,
    int8_dynamic_activation_int4_weight_cutlass,
)

# FIXME: change this!
_CUTLASS_DIR = ".../cutlass"


class ToyModel(torch.nn.Module):
    def __init__(self, nin, nout1, nout2):
        super().__init__()
        self.linear1 = torch.nn.Linear(nin, nout1)
        self.linear2 = torch.nn.Linear(nout1, nout2, bias=False)

    def forward(self, x):
        x = self.linear1(x)
        x = self.linear2(x)
        return x


methodq = int8_dynamic_activation_int4_weight_cutlass()
compile = False
dtype = torch.float16  # dtype = torch.bfloat16
device = "cuda"
bs, nin, nout1, nout2 = 256, 1024, 2048, 128

inputs = (torch.randn((1, bs, nin), dtype=dtype, device=device),)
model = ToyModel(nin, nout1, nout2).eval().to(dtype).to(device)
modelq = copy.deepcopy(model)

if compile:
    model = torch.compile(model, mode="max-autotune")

quantize_(modelq, methodq)
if not TORCH_VERSION_AT_LEAST_2_5:
    unwrap_tensor_subclass(modelq)

if compile:
    modelq = torch.compile(
        modelq,
        options={
            "max_autotune": True,
            "autotune_in_subproc": False,
            "max_autotune_gemm_backends": "Triton,CUTLASS",
            "cuda.cutlass_dir": _CUTLASS_DIR,
            "use_mixed_mm": True,
        },
    )


if __name__ == "__main__":
    from torchao.utils import benchmark_model

    nruns = 100
    torch._dynamo.reset()
    time = benchmark_model(model, nruns, inputs)
    timeq = benchmark_model(modelq, nruns, inputs)
    print(f"original model mean time  : {time:8.3f}")
    print(f"quantized model mean time : {timeq:8.3f}")
    print(f"speedup by quantization   : {time / timeq:8.3f}")
```
</details>

For particular shapes given in the script above, on A100 the micro-benchmark shows around 2x speedup over the case when float16 MM used, and around 1.8x speedup over the case when bfloat16 MM used.  (Note that this is for eager mode execution, as compilation to corresponding CUTLASS kernel is not yet supported by PyTorch.)

<details>
 <summary>Patch to run <tt>torchao/_models/llama/generate.py</tt></summary>

```
diff --git a/torchao/_models/llama/generate.py b/torchao/_models/llama/generate.py
index 5fb905d..e5b891b 100644
--- a/torchao/_models/llama/generate.py
+++ b/torchao/_models/llama/generate.py
@@ -206,6 +206,7 @@ def main(
             quantize_,
             int8_weight_only,
             int8_dynamic_activation_int8_weight,
+            int8_dynamic_activation_int4_weight_cutlass,
             int4_weight_only,
             fpx_weight_only,
             uintx_weight_only,
@@ -216,6 +217,8 @@ def main(
             quantize_(model, int8_weight_only())
         if "int8dq" in quantization:
             quantize_(model, int8_dynamic_activation_int8_weight())
+        if "w4a8-cutlass" in quantization:
+            quantize_(model, int8_dynamic_activation_int4_weight_cutlass())
         if "int4wo" in quantization:
             if "hqq" in quantization:
                 use_hqq=True
@@ -414,7 +417,7 @@ if __name__ == '__main__':
     parser.add_argument('--checkpoint_path', type=Path, default=Path("../../../checkpoints/meta-llama/Llama-2-7b-chat-hf/model.pth"), help='Model checkpoint path.')
     parser.add_argument('-q', '--quantization', type=str, 
         help=(
-            'Which quantization techniques to apply: int8dq, int8wo, fp6, int4wo-<groupsize>, int4wo-<groupsize>-hqq, autoquant, '
+            'Which quantization techniques to apply: int8dq, w4a8-cutlass, int8wo, fp6, int4wo-<groupsize>, int4wo-<groupsize>-hqq, autoquant, '
             +'autoquant-int4, uintx-<nbits>-<groupsize>, uintx-<nbits>-<groupsize>-hqq, sparse-marlin'
         )
     )
diff --git a/torchao/dtypes/affine_quantized_tensor.py b/torchao/dtypes/affine_quantized_tensor.py
index 1df3549..1252bb8 100644
--- a/torchao/dtypes/affine_quantized_tensor.py
+++ b/torchao/dtypes/affine_quantized_tensor.py
@@ -1158,6 +1158,7 @@ implements = AffineQuantizedTensor.implements
 # so that these can be shared by F.linear, aten.mm, aten.addmm dispatches
 
 def _linear_int8_act_int8_weight_check(input_tensor, weight_tensor, bias):
+    return False
     return (
         isinstance(input_tensor, AffineQuantizedTensor) and
         _aqt_is_int8_reduced_range(input_tensor) and
diff --git a/torchao/kernel/intmm.py b/torchao/kernel/intmm.py
index 3005cb1..451d0e6 100644
--- a/torchao/kernel/intmm.py
+++ b/torchao/kernel/intmm.py
@@ -54,6 +54,8 @@ if TORCH_VERSION_AT_LEAST_2_2:
             and k_is_nonzero_multiple_of_8
         )
 
+        bad_dimensions_for_cublas = False
+
         if device_cpu or bad_dimensions_for_cublas:
             # fallback path
             return torch.matmul(input.cpu().to(torch.int32), mat2.cpu().to(torch.int32)).to(
```
</details>

With the patch above, I was able to run Llama `generator.py` script.  The command to run is as follows:
```
python generate.py -q w4a8-cutlass
```
and the output is as follows (again, this is run on A100):
```
==========
Average tokens/sec: 10.21
Average Bandwidth: 33.78 GB/s
Peak Memory Usage: 14.22 GB
Model Size: 3.31 GB
```
while the reference output, for the case when no arguments supplied to `generate.py`, is as follows:
```
==========
Average tokens/sec: 32.87
Average Bandwidth: 434.31 GB/s
Peak Memory Usage: 13.62 GB
Model Size: 13.21 GB
```
So it's more than 3x slower, but the tokens/sec is not even that bad, considering that batch size is 1 here, and that the CUTLASS code has hard-coded for a block of threads to handle input tile size that is 128 for the same dimension.

So there is a room for improvement regarding the speed.  The text generated is garbage, however.  Even for the micro-benchmark above, output values visibly deviate from the values produced when native precision used (but at least they resemble each other).  I believe this is because of points I made in my [first comment](https://github.com/pytorch/ao/pull/880#issuecomment-2346325803) above, in particular because of not being possible to support group quantization here; so it's important to discuss these further.

alexsamardzic:
Made an update - turns out that actually CUTLASS needs a fix (posted below for now), and then `generate.py` script for Llama model would actually generate meaningful content.

<details>
<summary>CUTLASS fix</summary>

```
diff --git a/include/cutlass/epilogue/threadblock/default_epilogue_tensor_op.h b/include/cutlass/epilogue/threadblock/default_epilogue_tensor_op.h
index 1692cc30..5a1b164c 100644
--- a/include/cutlass/epilogue/threadblock/default_epilogue_tensor_op.h
+++ b/include/cutlass/epilogue/threadblock/default_epilogue_tensor_op.h
@@ -263,6 +263,44 @@ struct DefaultIteratorsTensorOp<
   static int const kFragmentsPerIteration = 2;
 };
 
+/// Partial specialization for bfloat16 <= int32_t x 8 epilogues avoids shared memory bank conflicts.
+template <
+  typename ThreadblockShape,
+  typename WarpShape,
+  typename InstructionShape,
+  typename ThreadMap
+>
+struct DefaultIteratorsTensorOp<
+  bfloat16_t, 
+  int32_t, 
+  8, 
+  ThreadblockShape, 
+  WarpShape, 
+  InstructionShape, 
+  ThreadMap> {
+  
+  using WarpTileIterator = cutlass::epilogue::warp::TileIteratorTensorOpMixed<
+    WarpShape,
+    InstructionShape,
+    int32_t,
+    32,
+    16,
+    8,
+    8
+  >;
+
+  using SharedLoadIterator = cutlass::epilogue::threadblock::SharedLoadIteratorMixed<
+    ThreadMap,
+    int32_t,
+    32,
+    16,
+    8,
+    8
+  >;
+
+  static int const kFragmentsPerIteration = 2;
+};
+
 /// Partial specialization for int8/int4b_t <= int32 x 16/8 epilogues avoids shared memory bank conflicts.
 /// Threadblock::kN = 256 still has bank conflicts.
 template <
```
</details>

On the other side, I tried with adapting tile sizes processed by block/warp of threads of corresponding CUTLASS kernel, in order to adapt to the fact that batch size is 1 here.  Here is an example of such change:
```--- a/torchao/csrc/cuda/s8s4_linear_cutlass/s8s4_linear_cutlass.cu
+++ b/torchao/csrc/cuda/s8s4_linear_cutlass/s8s4_linear_cutlass.cu
@@ -418,8 +418,8 @@ s8s4_linear_cutlass(const at::Tensor& input, const at::Tensor& input_scale,
   using ElementA = int8_t;
   using ElementB = cutlass::int4b_t;
   using ElementAccumulator = int32_t;
-  using ThreadblockShape = cutlass::gemm::GemmShape<128, 128, 64>;
-  using WarpShape = cutlass::gemm::GemmShape<64, 64, 64>;
+  using ThreadblockShape = cutlass::gemm::GemmShape<32, 128, 128>;
+  using WarpShape = cutlass::gemm::GemmShape<32, 32, 128>;
   using InstructionShape = cutlass::gemm::GemmShape<16, 8, 32>;
   AT_DISPATCH_SWITCH(
     input_scale.scalar_type(),
```
However, tokens/sec is not much improved this way.  Thus, the performance of this kernel for Llama model will require more work.

msaroufim (commented):
will make a second pass for the kernel code

torchao/quantization/quant_api.py
```diff
@@ -510,6 +512,41 @@ def int8_dynamic_activation_int4_weight(group_size=32, mapping_type=MappingType.
...
+    return weight
+
+def int8_dynamic_activation_int4_weight_cutlass():
+    """Applies int8 dynamic per token asymmetric activation quantization and int4 per group weight symmetric quantization to linear
+    This is used to produce a model for CUTLASS-based W4A8 kernel
+    """
+    return _get_linear_subclass_inserter(apply_int8_dynamic_activation_int4_weight_quant_cutlass)
+
+
 def int4_weight_only(group_size=128, layout_type=TensorCoreTiledLayoutType(inner_k_tiles=8), use_hqq=False):
```
cpuhrsch, line 550: Unrelated comment, what is this `use_hqq`? @jerryzh168 do you know?

torchao/csrc/cuda/s8s4_linear_cutlass/s8s4_linear_cutlass.cu
```diff
@@ -0,0 +1,487 @@
...
+    const at::Tensor& tensor_c, at::Tensor& tensor_d) {
+  using LayoutA = cutlass::layout::RowMajor;
+  using LayoutB = cutlass::layout::ColumnMajor;
+  using LayoutC = cutlass::layout::RowMajor;
+
+  const int m = tensor_a.size(0);
+  const int n = tensor_b.size(0);
+  const int k = tensor_a.size(1);
+
+  constexpr int AlignmentA = 128 / cutlass::sizeof_bits<ElementA>::value;
```
msaroufim, line 52: nit: mind adding a comment for why 128

Also how do you think about padding vs erroring

torchao/csrc/cuda/s8s4_linear_cutlass/s8s4_linear_cutlass.cu
```diff
@@ -0,0 +1,487 @@
...
+              __func__, " : Number of columns of tensor B must be divisible ",
+              "by ", AlignmentB);
+  TORCH_CHECK(n % AlignmentC == 0,
+              __func__, " : Number of columns of tensor C must be divisible ",
+              "by ", AlignmentC);
+
+  using SmArch = cutlass::arch::Sm80;
+  using ThreadblockSwizzle =
+      cutlass::gemm::threadblock::ThreadblockSwizzleStreamK;
+  constexpr auto NumStages = 4;
```
msaroufim, line 77: cutlass n00b but how do you pick these hyperparams?

setup.py
```diff
@@ -65,6 +65,12 @@ def get_extensions():
     extension = CUDAExtension if use_cuda else CppExtension
 
     if not IS_WINDOWS:
+        import cutlass_library
```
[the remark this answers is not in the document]
alexsamardzic: It's a recent addition to CUTLASS: a Python library that is able to generate C++ code for CUTLASS GEMM templates instantiation (which is nice to have, as these templates have dozen or more arguments, and it's oftentimes hard to get them right).  It's used in CUTLASS codegen for TorchInductor, like [here](https://github.com/pytorch/pytorch/tree/main/torch/_inductor/codegen/cuda).  However, recently CUTLASS itself also added a functionality to generate and compile C++ code for GEMM kernels, from a high-level specification in Python - this is part of `cutlass` Python package, see [here](https://github.com/NVIDIA/cutlass/blob/main/python/README.md).  Both `cutlass` and `cutlass_library` are available through [nvidia-cutlass](https://pypi.org/project/nvidia-cutlass/) pip package.  It's important to note that this package also contains all of the CUTLASS C++ header files, in order to make it possible to compile the C++ generated kernels.

setup.py
```diff
@@ -65,6 +65,12 @@ def get_extensions():
...
+        cutlass_library_dir = os.path.dirname(cutlass_library.__file__)
+        cutlass_include_dir = os.path.join(cutlass_library_dir, "source", "include")
+        # FIXME: remove this once CUTLASS package updated to include int4/int8 MM
+        cutlass_include_dir = "/data/quansight/scratch/cutlass/include"
```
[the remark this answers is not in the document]
alexsamardzic: I discussed this a bit in my first comment on this PR.  In order ao to compile after this PR eventually merged, CUTLASS C++ header files are to be made available.  There are at least two ways to do it:
1. To make CUTLASS repo a submodule of ao repo, just like PyTorch did it.
2. To make above mentioned `nvidia-cutlass` package a dependency of ao.

I'm leaning towards the later, and this is what above code, before "FIXME" is expecting.  However, in both of above cases, we'll certainly face an issue of having to depend on stuff that is not yet merged into CUTLASS, but we need it.  For example, at this very moment:

1. My CUTLASS [PR](https://github.com/NVIDIA/cutlass/pull/1413) with int4/int8 GEMM support for CUTLASS is merged, but CUTLASS team has not made a release in the meantime, so this functionality is only available in CUTLASS `main` branch, and also above mentioned `nvidia-cutlass` package doesn't contain it yet.
2. As mentioned in one of my comments above, while working in this PR, I found an omission in CUTLASS.  I created a CUTLASS [PR](https://github.com/NVIDIA/cutlass/pull/1826) with a fix, but this one is not yet merged, so neither CUTLASS `main` branch nor `nvidia-cutlass` package contain the fix at the moment, it's only available in my [branch](https://github.com/alexsamardzic/cutlass/tree/fix-evt-int32-accum-bfloat16-c).  So the only way to proceed with the development of my PR was to create a local copy of this branch - I created it in `/data/quansight/scratch/cutlass` directory on my machine; in order to try this PR, the local copy of this branch is to be created, and this last line in the snippet above is to be changed to the local directory.

From my experience with this stuff from PyTorch development based on CUTLASS, this is going to be permanent issue - if we decide to use CUTLASS in ao, the for the most of the time we'll need bleeding edge features.  So this is to be discussed further, IMO the best approach would be to build our own `nvidia-cutlass` package, from whatever CUTLASS branch we find the most appropriate.

torchao/quantization/quant_api.py
```diff
@@ -85,6 +85,7 @@
     "_get_subclass_inserter",
     "quantize_",
     "int8_dynamic_activation_int4_weight",
+    "int8_dynamic_activation_int4_weight_cutlass",
```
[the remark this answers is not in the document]
alexsamardzic: Now when I have the dots connected, in the sense that I can run a micro-benchmark, and also Lllama model, using this kernel, I'm working on a more detailed profiling, part of this is also comparing the performance of this kernel with `int8_dynamic_activation_int4_weight` kernel.  I'll report all my findings here when I'm done with the profiling.

test/test_s8s4_linear_cutlass.py
```diff
@@ -0,0 +1,51 @@
+# FIXME: move this test to the appropriate test file!!!
```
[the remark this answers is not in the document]
alexsamardzic: Yes.  Again, as mentioned in one of my comments above: At the moment, most of the "FIXME"-s in the PR are as I'm aware that I took shortcuts to make things work.  If/when we're happy with the main stuff, I'll revisit all of these, and redo them in the proper "ao-way".

torchao/csrc/cuda/s8s4_linear_cutlass/s8s4_linear_cutlass.cu
alexsamardzic: The 128 bits here is because of how tensor cores work (so it's not CUTLASS-specific), at least for SM 8.x.  It's related to the layout of tiles of matrix operands that single warp of thread is multiplying cooperatively.  The best explanation that I found so far is in GTC 2020 [talk](https://developer.nvidia.com/gtc/2020/video/s21745-vid), by CUTLASS team, around slide 15.

We can consider padding (maybe at the later stage?), I believe it would the best to incorporate padding together with the quantization.

torchao/csrc/cuda/s8s4_linear_cutlass/s8s4_linear_cutlass.cu
alexsamardzic: These, and others, are the CUTLASS GEMM C++ template arguments.  As mentioned above, there is dozen of these to set, but on the other side only small number of combinations of these arguments actually works.  The above mentioned `cutlass_library` package enumerates some of these working combinations.  The CUTLASS itself doesn't include any sort of heuristic for selection of these parameters, for example based on GEMM operand shapes.  So I had to hard-code some values, at least for now.  The values selected here are based on my previous experimentation with different combinations, and different operand shapes - in the sense that these values should provide acceptable performance for number of cases.  But certainly there are cases where these values are not good fit, Lllama inference, having batch size 1, is one such example.  So we may want to consider adding some heuristic here, but on the longer term we'd probably prefer to do support some auto-tuning, just like what is possible with Triton kernels.

torchao/quantization/quant_api.py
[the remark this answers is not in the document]
alexsamardzic: As a quick update here: Using the micro-benchmarking script above, it seems this PR is just 3-5% faster than `int8_dynamic_activation_int4_weight`.  However, on the Llama generator, it seems about 2x faster, when tokens/sec numbers compared.  (Remember that all the caveats from my first comment above still apply, so let's not jump into any conclusions for now.)

alexsamardzic:
(Pushed an update, where the branch is just rebased on the latest main.)

I did lots of profiling in the meantime, focusing primarily on running Llama generator (`torchao/_models/llama/generate.py`), using tokens/sec as performance measure, and comparing between this PR and W8A8DQ case (i.e. when model quantized using `int8_dynamic_activation_int8_weight`).  All of the results presented below were for A100 runs, the W8A8DQ run was as follows:
```
python generate.py -q int8dq
```
and the run for this PR was as follows (with the patch mentioned [above](https://github.com/pytorch/ao/pull/880#issuecomment-2356558669) applied beforehand):
```
python generate.py -q w4a8-cutlass
```

---
TLDR (note that each of these items could be verified by profiling W8A8DQ alone, without using this PR at all):

0. The CUTLASS MM kernel in case of this PR, and also the Triton kernel MM for the W8A8DQ are not, at least at this moment, the most critical for performance.  Instead, the other parts of the code, that are run each time along with the linear operator, are taking more execution time - see the remaining two items in the list.
1. The dispatch checks registered [here](https://github.com/pytorch/ao/blob/c187f876d2f15c02b350f5e1c50855b485132a55/torchao/dtypes/affine_quantized_tensor.py#L1724-L1736) are re-run over and over again.  These take considerable time, and also they make the performance depending on the position of registering the kernel, and corresponding check, in this list: if corresponding item moved between top and bottom of the list, the tokens/sec differ up to 10%. (@jerryzh168)
2. The dynamic quantization takes considerable time too, more than the MM kernel itself.  This could be improved, by working on fusing PyTorch operators used to perform quantization, or by implementing dedicated kernel(s) for dynamic quantization; also for Llama generator in particular by adjusting configs of these kernels to the fact that the number of inputs is 1.  (Still, IMO it's questionable is there any performance benefit in using dynamic quantization vs. weight quantization only.)
---

As an example for item 1 above, here are the performance results, as printed by `generate.py` script, in case when item registering given kernel and check moved to the first place in the list:
```
python generate -q int8dq
# ... lots of output here
==========
Average tokens/sec: 4.81
Average Bandwidth: 31.83 GB/s
Peak Memory Usage: 14.86 GB
Model Size: 6.62 GB

python generate -q w4a8-cutlass
# ... lots of output here
==========
Average tokens/sec: 10.31
Average Bandwidth: 34.11 GB/s
Peak Memory Usage: 14.22 GB
Model Size: 3.31 GB
```
and when moved to the last place in the list:
```
python generate -q int8dq
# ... lots of output here
==========
Average tokens/sec: 4.35
Average Bandwidth: 28.82 GB/s
Peak Memory Usage: 14.86 GB
Model Size: 6.62 GB

==========
# ... lots of output here
Average tokens/sec: 9.92
Average Bandwidth: 32.82 GB/s
Peak Memory Usage: 14.22 GB
Model Size: 3.31 GB
```

The generator runs are profiled using `pyinstrument`, and verified using `cProfile` and `nsys` profilers.  With the profiling run launched as follows:
```
python -m pyinstrument generate.py -q w4a8-cutlass
```
here is the relevant part of the `pyinstrument` output:

pyinstrument

So, for the attention segment of the model, one could see that everything related to running the linear operator takes about 34s in total.  Out of this time, 24s are spend in the dynamic quantization, while about 9.4s only are spent on the linear operator itself, and then out of these 9.4s, only 2.4s are spent on the CUTLASS MM kernel execution, while the rest of time get spent on checking to which kernel to dispatch (note that for this run, the check for applicability of the CUTLASS kernel is added last to the list) - these checks are not visible in this snippet, as `pyinstrument` by default suppresses calls that take shorter time, but attached below is full `pyinstrument` output to verify it.  The distribution of time spent is alike for the feed-forward part of the network - this could be also seen from the full output below.

Here is the `pyinstrument --show-all ...` output for the run above: [pyinstrument.txt](https://github.com/user-attachments/files/17269635/pyinstrument.txt).

As mentioned above, profiling results are verified using `cProfile` and `nsys`.  For example, for `nsys` run as follows:
```
nsys profile -w true -t cuda,nvtx,osrt,cudnn,cublas -s cpu -python-sampling=true $(which python) generate.py -q w4a8-cutlass
```
here is a screenshot of the timeline as shown by `nsys-ui`:

nsys-ui-1

Here, one could see that loading of model takes about 30s, then there is a short sequence of copying model to GPU and doing weights quantization, and then the rest of the timeline is the inference.  The CUTLASS MM kernel, designated as `Kernel2` here, takes less 30% of time of all of the CUDA kernels executed.  If timeline zoomed into a segment of time during the inference, one could see that CUDA kernels are not actually executed tightly (because the checks and dynamic quantization are actually a sequence of calls to PyTorch kernels that are not fused):

nsys-ui-2

cpuhrsch:
@alexsamardzic - Was the model torch.compile'd with mode 'max-autotune'? Also you can use [`torch.profiler`](https://github.com/pytorch-labs/segment-anything-fast/tree/e6aadeb86f3ae1f58c3f98e2a91e251716e0f2aa/experiments#kernel-traces) to generate kernel traces potentially a bit more quickly than with nsys (at least for rapid iteration). You can then open these with https://ui.perfetto.dev/

torchao/quantization/quant_api.py
jerryzh168: yeah this means use `hqq` algorithm to choose qparams and quantize the weight, since it is reusing the tinygemm kernel, we just added this as a separate option here

alexsamardzic:
> @alexsamardzic - Was the model torch.compile'd with mode 'max-autotune'?

Good question :smile:  Apparently, there is `--compile` option for the `generate.py` script to explicitly activate compilation.  So I re-did the analysis, this time using following command to run the generate script:
```
python generate.py --compile [-q parameter]
```
and here is a table with results for several different quantization techniques (again, all results are for A100):

| `-q parameter` | Average tokens/sec | Average Bandwidth in GB/s | Peak Memory Usage in GB | Model Size in GB |
| :--- | ---: | ---: | ---: | ---: |
| | 95.24 | 258.55 | 13.90 | 13.21 |
| `-q int8wo` | 155.31 | 1028.37 | 8.97 | 6.62 |
| `-q int4wo-32` | 186.70 | 774.98 | 5.31 | 4.15 |
| `-q int4wo-hqq` | 186.47 | 774.01 | 5.04 | 4.15 |
| `-q int8dq` | 49.64 | 328.72 | 9.44 | 6.62 |
| `-q w4a8-cutlass` | 86.14 | 285.08 | 4.75 | 3.31 |
| `-q w4a8-cutlass` (**tuned**) | 119.31 | 394.86 | 4.52 | 3.31 |

The "tuned" in the last row of the table above means that the size of tiles that each block of CUDA threads in the CUTLASS-based W4A8 kernel is processing is adjusted to the fact that single input is processed - more precisely, following patch is applied on top of this PR:

```
diff --git a/torchao/csrc/cuda/s8s4_linear_cutlass/s8s4_linear_cutlass.cu b/torchao/csrc/cuda/s8s4_linear_cutlass/s8s4_linear_cutlass.cu
index e60a69a..e85c9fc 100644
--- a/torchao/csrc/cuda/s8s4_linear_cutlass/s8s4_linear_cutlass.cu
+++ b/torchao/csrc/cuda/s8s4_linear_cutlass/s8s4_linear_cutlass.cu
@@ -418,8 +418,8 @@ s8s4_linear_cutlass(const at::Tensor& input, const at::Tensor& input_scale,
   using ElementA = int8_t;
   using ElementB = cutlass::int4b_t;
   using ElementAccumulator = int32_t;
-  using ThreadblockShape = cutlass::gemm::GemmShape<128, 128, 64>;
-  using WarpShape = cutlass::gemm::GemmShape<64, 64, 64>;
+  using ThreadblockShape = cutlass::gemm::GemmShape<32, 128, 128>;
+  using WarpShape = cutlass::gemm::GemmShape<32, 32, 128>;
   using InstructionShape = cutlass::gemm::GemmShape<16, 8, 32>;
   AT_DISPATCH_SWITCH(
     input_scale.scalar_type(),
```

The `pyinstrument` and `nsys` results are harder to interpret when compilation used.  It seems that what I mentioned as potential performance bottlenecks in my last comment above may be still affecting performance - as an illustration, here is a screenshot of the timeline as shown by `nsys-ui`, during the tokens generation:



So there is still large number of small kernels run between successive invocations of W4A8 kernel that take considerable time.  I'm looking further into this...

---

Note: PR is rebased on the latest main, and updated so that the only change needed to run the Llama generator is to point the code to the directory where the latest CUTLASS main is checked out.  In the meantime, my CUTLASS PR mentioned above is merged, so when version 3.6.0 of CUTLASS released, it will be enough to install `nvidia-cutlass` package to satisfy the dependency.

cpuhrsch:
@alexsamardzic So there's a --compile and --compile-prefill option. Maybe try applying both. One thing that should happen is the usage of CUDA graphs. Those should remove the spaces between these small kernels. You can also generate a profile using `--profile some_name_1` and then view it with https://ui.perfetto.dev/ . I recommend to `gzip` the resulting `some_name_1.json`. You can then also share that profile with us.

alexsamardzic:
> @alexsamardzic So there's a --compile and --compile-prefill option. Maybe try applying both. One thing that should happen is the usage of CUDA graphs. Those should remove the spaces between these small kernels. You can also generate a profile using `--profile some_name_1` and then view it with https://ui.perfetto.dev/ . I recommend to `gzip` the resulting `some_name_1.json`. You can then also share that profile with us.

OK - apparently there is an [issue](https://forums.developer.nvidia.com/t/nsys-profile-failed-when-using-pytorch-cudagraph/295523) between older versions of `nsys` and CUDA graphs.  After I upgraded `nsys`, the timeline shows that the W4A8 is taking most of the time:



This is matched by what Torch profiler shows (here is the full trace: 
[w4a8-cutlass.zip](https://github.com/user-attachments/files/17385076/w4a8-cutlass.zip)):



The `--compile_prefill` option helps a little bit, but not significantly.

@cpuhrsch Thanks for your help in running the model properly!

cpuhrsch:
Nice!

So we have `int8_dynamic_activation_int8_weight` and this adds `int8_dynamic_activation_int4_weight_cutlass`, can we call it `int8_dynamic_activation_int4_weight` instead or does that conflict with something else?

alexsamardzic:
> So we have `int8_dynamic_activation_int8_weight` and this adds `int8_dynamic_activation_int4_weight_cutlass`, can we call it `int8_dynamic_activation_int4_weight` instead or does that conflict with something else?

Yes, there is `int8_dynamic_activation_int4_weight` already there (but I'm unsure is it used at all).

cpuhrsch:
@alexsamardzic - Can you dig into what `int8_dynamic_activation_int4_weight` does? Maybe we can unify it. Then CUTLASS just becomes an implementation detail of `int8_dynamic_activation_int4_weight` (in cases where it's available).

alexsamardzic:
> @alexsamardzic - Can you dig into what `int8_dynamic_activation_int4_weight` does? Maybe we can unify it. Then CUTLASS just becomes an implementation detail of `int8_dynamic_activation_int4_weight` (in cases where it's available).

Yes: `int8_dynamic_activation_int4_weight` is just a quantization function, like all the others, but contrary to other quantization functions, for `AffineQuantizedTensor` there is no kernel check/impl registered for the case if tensor is quantized using this function.  The function is referenced only in some quant API tests, and not used anywhere else.  On the other side, there is a comment "for executorch" for one mention of this function, so I'm not sure is it used outside the torchao.

For `int8_dynamic_activation_int4_weight_cutlass`, I started from `int8_dynamic_activation_int4_weight` implementation, and changed couple things:

1. For now, `int8_dynamic_activation_int4_weight_cutlass` is restricted to symmetric quantization for both input and weight - as discussed in my first comment above, my CUTLASS-based W4A8 kernel still doesn't support zero-point handling.  On the other side, `int8_dynamic_activation_int4_weight` does asymmetric quantization of input, and user can select between symmetric and assymetric quantization of weights.
2. The result of `int8_dynamic_activation_int4_weight` is produced as one 4-bit quantized value per byte, while for `int8_dynamic_activation_int4_weight_cutlass`, I'm storing two 4-bit quantized values per byte, as this is how CUTLASS expects it.

Thus indeed these two could be unified into a single function, with some additional function arguments added.  Shall I do it this way for this PR?

cpuhrsch:
@alexsamardzic - so it seems that in order to use CUTLASS here you need

a) symmetric only
b) a special memory layout.

Now, it's useful to use CUTLASS on CUDA.

So if someone uses `int8_dynamic_activation_int4_weight` while on CUDA (and for certain shapes (and using symmetric quantization)) we could elect to create instances of AQT with a special CUTLASS layout so we can use the CUTLASS kernels.

For this approach we don't need new flags. We just need to carefully check the setting it's being used in. I'd prefer to try this first so we don't add more API surface.

alexsamardzic:
> For this approach we don't need new flags. We just need to carefully check the setting it's being used in. I'd prefer to try this first so we don't add more API surface.

Let me clarify what I'm proposing:  At the moment, with this PR we have two similar quantization methods:
```
def int8_dynamic_activation_int4_weight(
    group_size=32,
    mapping_type=MappingType.SYMMETRIC
)
```
and:
```
def int8_dynamic_activation_int4_weight_cutlass()
```
My idea was that I unify these two into:
```
def int8_dynamic_activation_int4_weight(
    group_size=32,
    mapping_type=MappingType.SYMMETRIC,
    input_mapping_type=MappingType.ASYMMETRIC, 
    pack_bits=False)
```
This way, it works unchanged for existing users of `int8_dynamic_activation_int4_weight`, becuase the defaults for the two new arguments of `int8_dynamic_activation_int4_weight` are set to match what existing `int8_dynamic_activation_int4_weight` implementation is doing.  For CUTLASS-based W4A8 kernel, one would have to quantize the model as follows (and this is for now only, namely if/when zero-point handling implemented into this PR, then `MappingType.ASYMMETRIC` will be another valid choice for `mapping_type` and `input_mapping_type`):
```
quantize_(
    model,
    int8_dynamic_activation_int4_weight(
        group_size=None,
        mapping_type=MappingType.SYMMETRIC,
        input_mapping_type=MappingType.SYMMETRIC,
        pack_bits=True)
)
```

The AQT works unchanged:  It gets created by some quantization method, possibly one of the above two.  When it comes to evaluation of a model, a kernel to use given AQT is searched for.  The W4A8 CUTLASS kernel is in the mix, it will get selected if AQT is on CUDA device, and if quantization method used was as in the last code snippet.

So users of existing API are not affected.  And the API change is of the same type as when a new callable, having some arguments and implementing a new quantization method, introduced - which I assume is allowed.

cpuhrsch:
Hm, interesting. The signature for `int8_dynamic_activation_int8_weight` is very different from `int8_dynamic_activation_int4_weight` at the moment.

In any case, I think what you suggest sounds good. I do think that `pack_bits` might be replaced by `layout` similar to how we do it for `int8_dynamic_activation_int8_weight`?

jerryzh168:
yeah I feel we should use layout and put `packe_bits` arg to layout and have special settings based on layout as well, similar to: https://github.com/pytorch/ao/blob/c87cc9b7286a46e9dfc076fa2417eb9b64ccc807/torchao/quantization/quant_api.py#L556

alexsamardzic:
> yeah I feel we should use layout and put `packe_bits` arg to layout and have special settings based on layout as well

OK. Looking into this:  So I'll implement a layout, let's name it `PackedInt4Layout`, which I guess will just "compress" each pair of 4-bit values into a single byte in its `pre_process` method.  Is it then necessary to create `PackedInt4TensorImpl` class, alike to other `TensorImpl` classes and if so, which methods should this class provide?  This part is somewhat unclear, as different `TensorImpl` classes are doing different things...

alexsamardzic:
Changed the quantization name from `w4a8-cutlass` to `int8adq-int4w-symm` in the `torchao/_models/llama/generate.py` script.

alexsamardzic:
Made the kernel built only if CUTLASS installation found at build time, also implemented a simple heuristic to improve the performance in case the number of inputs is small.

alexsamardzic:
A benchmark script added.

alexsamardzic:
This PR is ready for review now.

---

The PR introduces CUTLASS as torchao dependency.  The CUTLASS version should be 3.6.0 or higher, the `setup.py` script will check for its presence and not there, the CUTLASS-based `s8s4_linear_cutlass` operator will be compiled just so that it always returns an error.  The CUTLASS 3.6.0 is not yet released, hopefully it will be soon, but in the meantime in order for have mentioned operator built properly, one should clone the CUTLASS [repo](https://github.com/NVIDIA/cutlass), and then apply following patch to this PR:
```
diff --git a/setup.py b/setup.py
index b2e2c01..4cc5bf5 100644
--- a/setup.py
+++ b/setup.py
@@ -78,6 +78,10 @@ def get_extensions():
                 cutlass_library_dir = os.path.dirname(cutlass_library.__file__)
                 cutlass_include_dir = os.path.join(cutlass_library_dir, "source", "include")
 
+        # FIXME: remove this
+        use_cutlass = True
+        cutlass_include_dir = '/path-to-where-cutlass-repo-is-cloned/cutlass/include'
+
         extra_link_args = []
         extra_compile_args = {
             "cxx": [
```
Once CUTLASS 3.6.0 released, and corresponding version of `nvidia-cutlass` Python package available, the CUTLASS dependency will be fulfilled as usual: `pip install nvidia-cutlass`.

There are alternative approaches here: to include CUTLASS source within torchao, as PyTorch does, or to depend on CUTLASS from PyTorch (including moving `s8s4_linear_cutlass` operator implementation into PyTorch).  IMO, above is the cleanest approach, but we could discuss these too.

---

There is some additional work possible on top of this PR:

1. Scale factors and bias could be either of `torch.float16` or `torch.bfloat16` dtype.  It is possible to support `torch.float` here too, as well as various mixes of these dtypes.
2. Add support for zero-point - namely, as discussed in the first comment on this PR, the current implementation supports only symmetric quantization for both weight and activation tensors.
3. Further improve performance for smaller (`<256` or so) number of inputs.  That would require upstream changes in CUTLASS.
4. Consider adding support for group quantization.  That, too, would require upstream changes in CUTLASS.

IMO, this PR should be merged as is, and mentioned features could be eventually added later.

alexsamardzic:
Just pushed some minor fixes that should silent most of current CI issues.

The PR is good for a review, but overall there is no point in merging it before CUTLASS 3.6.0 released.  Then, we should update CI to have `nvidia-cutlass` package installed - only then the new functionality will be properly tested.

The PR contains new `benchmarks/benchmark_s8s4_cutlass.py` benchmarking script, but it produces a lengthy table as output.  Not sure is it good enough to just mention Lllama generator tok/sec instead?  Also, as mentioned above: if deemed important, it is possible to work later on CUTLASS to improve performance for this particular combination of operands, so the performance numbers may not be final.

cpuhrsch:
Yes Llama tok/sec would be useful. What change in performance does this PR cause? It's ok if a newer version of CUTLASS is required to run this. You could change the code to guard on that and/or manually install it given a build flag. Since we have our own OSS CI setup you could also add a one-off CI job to test on CUTLASS 3.6 to move along faster with this.

alexsamardzic:
> Yes Llama tok/sec would be useful. What change in performance does this PR cause? It's ok if a newer version of CUTLASS is required to run this. You could change the code to guard on that and/or manually install it given a build flag. Since we have our own OSS CI setup you could also add a one-off CI job to test on CUTLASS 3.6 to move along faster with this.

Pushed an update with commit message expanded.

The PR doesn't change performance of anything, the performance of new `s8s4_linear_cutlass` operator is discussed above.

The `setup.py` script already has a guard, checking for appropriate version of CUTLASS package - if this package not available, the `s8s4_linear_cutlass` operator will be compiled to basically report "not implemented" error.

Am looking into the CI setup in order to be able to properly test this PR in the CI until the CUTLASS 3.6.0 officially released...

alexsamardzic:
A Github workflow `.github/workflows/s8s4_linear_cutlass_test.yml` added, that should test the CUTLASS-based W4A8 kernel in the CI.

cpuhrsch:
I don't see the job in the list of jobs for this PR.

.github/workflows/s8s4_linear_cutlass_test.yml
```diff
@@ -0,0 +1,43 @@
+name: Run CUTLASS-based W4A8 kernel tests
+
+on:
+  push:
+    branches:
+      - w4a8-cutlass
```
cpuhrsch, line 6: I think you can enable this more generally to prevent regressions: https://github.com/pytorch/ao/blob/abff563ba515576fc48cd4ac0feb923dd65dc267/.github/workflows/pr-label-check.yml#L3

cpuhrsch:
@alexsamardzic - Do we need to add a new job for this or can we ensure that CUTLASS builds and gets tested as part of the existing tests?

torchao/dtypes/uintx/tensor_core_tiled_layout.py
```diff
@@ -33,6 +33,16 @@ def _aqt_is_tensor_core_tile_uint4(aqt):
     )
 
 
+def _aqt_is_tensor_core_tile_int4(aqt):
```
jerryzh168, line 36: need to rename this to remove `tensor_core_tile` in the name I think

torchao/quantization/quant_api.py
```diff
@@ -738,6 +746,25 @@ def _int8_symm_per_token_reduced_range_quant(x: torch.Tensor) -> torch.Tensor:
     )
 
 
+def _int8_symm_per_token_reduced_range_quant_cutlass(x: torch.Tensor) -> torch.Tensor:
```
jerryzh168, line 749: probably need to add this to https://github.com/pytorch/ao/blob/abff563ba515576fc48cd4ac0feb923dd65dc267/torchao/quantization/quant_api.py#L1170 for serialization support, can you add cutlass int4 layout to https://github.com/pytorch/ao/blob/abff563ba515576fc48cd4ac0feb923dd65dc267/test/dtypes/test_affine_quantized.py#L34 so it's tested for serialization as well

torchao/dtypes/__init__.py
```diff
@@ -17,6 +17,7 @@
 from .uintx import (
     BlockSparseLayout,
     Int4CPULayout,
+    Int4PackedLayout,
```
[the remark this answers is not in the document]
alexsamardzic: I can do it, but layout is not specific to CUTLASS - it's just 4-bit values that are packed by two in a 8-bit tensor.

torchao/dtypes/uintx/tensor_core_tiled_layout.py
alexsamardzic: Changed, and moved out of this file, as indeed it doesn't belong here.

torchao/quantization/quant_api.py
alexsamardzic: Made both changes.

alexsamardzic:
> @alexsamardzic - Do we need to add a new job for this or can we ensure that CUTLASS builds and gets tested as part of the existing tests?

It will be tested both through regression and nightly workflows; it's only that until CUTLASS 3.6 released, and until we decide how to make CUTLASS a dependency (more on this below),  CUTLASS headers has to be provided in a particular way.  I removed the new workflows, and changed regression workflow (`regression_test.yml`) temporarily so that it gets tested in CI.

alexsamardzic:
About CUTLASS:

[CUTLASS](https://github.com/NVIDIA/cutlass) is a heavily templated C++ header library by NVIDIA, primarily implementing GEMM (but also convolutions etc.) on tensor cores, for various combination of operands.  However, there is also some Python code in-there, that basically makes it possible to write given GEMM call in Python, and have corresponding C++ code implementing that call generated and compiled automatically.  Thus, CUTLASS gets packaged as a Python library `nvidia-cutlass`, that contains both its Python modules and C++ header files.

For writing CUTLASS-based GEMM kernels, in our case it's typically needed to first extend CUTLASS, to support a combination of operands that we're interested in.  As usual, the change is typically made first in a private fork of the CUTLASS repo, then it eventually gets merged into `main` branch of the CUTLASS repo, then after a new revision of CUTLASS is released it appears in `cutlass-nvidia` package too.  Note that the delay of including changes is increasing along the way.  For us, it's also relevant that PyTorch keeps CUTLASS repo as sub-module, and they have the sub-module pin moved from time to time, so the changes appear there with even more delay.

Now, the question is if we include CUTLASS-based kernels in torchao, how are we going to provide CUTLASS header files during the torchao build?  This PR already changes `setup.py` so that building CUTLASS-based kernels is optional, depending on are the CUTLASS header file found when torchao built; but obviously if we add CUTLASS-based kernels to torchao, we'd eventually want them to be built and used.  There are several options, each one with own trade-offs:

1. Like PyTorch, we can make CUTLASS a sub-module of torchao repo.  The advantage is that we could move the pin at our will, the disadvantage is that it complicates the structure of torchao repo.  Furthermore, the question is: do we have sub-module pointing to some kind of our fork of CUTLASS repo (in which case we have our eventual CUTLASS changes available as soon as we made it), or to CUTLASS official repo (in which case we are still facing the delay of merging our eventual CUTLASS changes into the CUTLASS `main`)?
2. We can add `nvidia-cutlass` package into the `dev-requirements.txt`.  This PR is written having this approach in mind.  IMO, this approach is the least hassle for torchao, but then we face an extended delay regarding using our eventual updates in CUTLASS upstream - we'll see them only when CUTLASS team makes a release.  CUTLASS releases typically happen each two months, but that's not the hard rule - right now, it's more than three months since the last release.  Another problem with this approach is that a developer waiting for a `nvidia-cutlass` package release with nededed updates, has in meantime to somehow point torchao `setup.py` to private copy of a CUTLASS tree, having the changes needed; same holds for CI testing (that's what I'm trying to do now with these temporary changes in the torchao CI workflow files), that makes the development cumbersome.  Maybe a solution is that we build and host our own version of `nvidia-cutlass` package, that we'd base on whatever fork of CUTLASS repo that suits us?
3. We can use CUTLASS provided by PyTorch.  For `@torch.compile` auto-tuning, I think that PyTorch package should contain CUTLASS headers, but it seems that this is not the case at the moment; I'll have to check this more carefully.  The problem with this approach would be that the delay of seeing CUTLASS updates that we need is longest - for example, at the moment PyTorch still has its CUTLASS pin set to version 3.4.1, that is rather old, and even when PyTorch team updates the pin, we'd have to wait for the next PyTorch release to have these changes available.

A variation of latest approach is to split this PR into two, and just move C++ part into PyTorch.  The disadvantage remains the same, but torchao would then have no worries about CUTLASS.  This may be worth consideration as there are already CUTLASS-based kernels in PyTorch that may be interesting for torchao (at least there are two of mine: 2:4 sparsity kernel and W8A16/W4A16 kernel - and I'm wondering, whatever we do, shall these sit together with this W4A8 kernel, be it in PyTorch or tochao code base?).  Moreover, as already mentioned, PyTorch has some CUTLASS-based support for `@torch.compile` auto-tuning and we'll probably want to extend on this: CUTLASS-based GEMM kernels are in my experience oftentimes slightly faster than GEMM kernels that `@torch.compile` generates in Triton (on the other side, CUTLASS-based GEMM kernels have disadvantages here too: they are very slow to compile when compared to Triton, and cannot be fused with other code compiled to Triton).  I don't think it's possible to do CUTLASS-based auto-tuning for `torch.comple` outside of PyTorch, so at least for this we'll have to employ CUTLASS from within PyTorch.  Finally, unless we employ the latest approach mentioned, we'll have to take care not to clash with whatever part of CUTLASS comes within PyTorch package.

So there is no clear cut, and we'll have to be careful what we decide here...  Any comments welcome.

@cpuhrsch @drisspg @jcaip

cpuhrsch:
So with torchao we can compile and ship our own CUDA binaries and since CUTLASS is header only why is option 1 so much more of a hassle than option 2? I don't think it's something that the users of our nightlies or stable releases will see.

If we want to prevent conflict with other versions of CUTLASS available, I think we should guard on the CUTLASS version within our own CUDA code, but that also seems reasonable to do. Otherwise there's a risk that we might compile different CUTLASS versions at once. Maybe there's also a risk we can overwrite symbols at link time. But the guards should help with that. 

I don't think we should depend on a fork within the submodule, but I do think it's ok to land our own more complex modifications / copies for various CUTLASS templates for a specific kernel first and then eventually deduplicate again.

@alexsamardzic can you give option 1 a go and see if it works in CI?

alexsamardzic:
> @alexsamardzic can you give option 1 a go and see if it works in CI?

It works - this is kind of how I run the tests at the moment: I've changed `.github/workflows/regression_test.yml` to clone the `main` branch of CUTLASS repo into the torchao top-level directory, and then I've changed `setup.py` of torchao to find CUTLASS include files there (ofc, these changes are just temporarily).  It could be seen [here](.github/workflows/regression_test.yml) that all tests from `test/test_s8s4_linear_cutlass.py` pass.

drisspg:
Sorry for responding late, but yeah I am pro Option 1 and anti cutlass fork. I think we should reach out to the cutlass folks to get a sense of when the 3.6 release to come out

alexsamardzic:
Pushed a variant that uses "option 1", let me check how it runs through CI.

jcaip:
Yeah +1 on including the CUTLASS headers in AO. I think it's advantageous to not be tied to PyTorch core for CUTLASS template updates. I know that Daniel is working on some custom kernels for 2:4 activation that need the latest CUTLASS version. 

>A variation of latest approach is to split this PR into two, and just move C++ part into PyTorch. The disadvantage remains the same, but torchao would then have no worries about CUTLASS. This may be worth consideration as there are already CUTLASS-based kernels in PyTorch that may be interesting for torchao (at least there are two of mine: 2:4 sparsity kernel and W8A16/W4A16 kernel - and I'm wondering, whatever we do, shall these sit together with this W4A8 kernel, be it in PyTorch or tochao code base?). 

Do you have an idea on how this would affect 2:4 support in core? I don't think we have a way to expose torchao ops / kernels in core but I may be wrong here.

alexsamardzic:
> Do you have an idea on how this would affect 2:4 support in core? I don't think we have a way to expose torchao ops / kernels in core but I may be wrong here.

I guess some period of deprecation could be used where users are pointed to the same functionality to torchao - just like it was discussed recently, for moving quantization related functionality.  On the other side, CUTLASS-based 2:4 sparsity operator is indeed somewhat coupled with cuSPARSELt-based one in the core; but maybe the whole `SparseSemiStructuredTensor` should be moved to torchao?

cpuhrsch (approved):

alexsamardzic:
Moved `third_party/cutlas` pin to freshly released `v3.6.0`.

cpuhrsch:
Merge at will @alexsamardzic - I assume the failed CI jobs are unrelated.

alexsamardzic:
> Merge at will @alexsamardzic - I assume the failed CI jobs are unrelated.

@jerryzh168 Do you maybe want to give another look to the changes on the Python side?

torchao/dtypes/uintx/int4_packed_layout.py
```diff
@@ -0,0 +1,160 @@
...
+    # TODO: use torch.int4
+    return (
+        aqt.tensor_impl.dtype == torch.int8
+        and aqt.quant_min == -8
+        and aqt.quant_max == 7
+    )
+
+
+@dataclass(frozen=True)
+class Int4PackedLayout(Layout):
```
jerryzh168, line 31: is it better to add cutlass in the name?

torchao/dtypes/uintx/int4_packed_layout.py
```diff
@@ -0,0 +1,160 @@
...
+
+    @classmethod
+    def from_plain(
+        cls,
+        int_data: torch.Tensor,
+        scale: torch.Tensor,
+        zero_point: torch.Tensor,
+        _layout: Layout,
+    ):
+        int_data_s4 = ((int_data[:, 1::2] & 0xF) << 4) | (int_data[:, 0::2] & 0xF)
```
jerryzh168, line 113: will be good to add an assert for `_layout`, like https://github.com/pytorch/ao/blob/fe5f11b2c58b452e01ba9ec7359629928b143619/torchao/dtypes/uintx/tensor_core_tiled_layout.py#L236

alexsamardzic:
Made changes suggested, and rebased on the latest main.

torchao/dtypes/uintx/cutlass_int4_packed_layout.py
```diff
@@ -0,0 +1,160 @@
...
+    )
+
+
+def _linear_int8_act_int4_weight_cutlass_impl(input_tensor, weight_tensor, bias):
+    from torchao.ops import s8s4_linear_cutlass
+
+    weight = weight_tensor.tensor_impl.int_data
+    weight_scale = weight_tensor.tensor_impl.scale
+    input = input_tensor.tensor_impl.int_data
+    input_scale = input_tensor.tensor_impl.scale
```
jerryzh168, line 156: looks like zero_point is not used, you can make zero_point to be None by setting zero_point_domain to be ZeroPointDomain.NONE in quant_api.py

torchao/quantization/quant_api.py
```diff
@@ -599,7 +600,12 @@ def apply_int8_dynamic_activation_int4_weight_quant(
     if act_mapping_type == MappingType.ASYMMETRIC:
         input_quant_func = _int8_asymm_per_token_quant
     elif act_mapping_type == MappingType.SYMMETRIC:
-        input_quant_func = _int8_symm_per_token_quant
+        if isinstance(layout, MarlinQQQLayout):
+            input_quant_func = _int8_symm_per_token_quant
+        elif isinstance(layout, CutlassInt4PackedLayout):
```
jerryzh168, line 605: you can set zero_point_domain to be ZeroPointDomain.NONE(https://github.com/pytorch/ao/blob/f7f20e9e8ba28a4910efb62e7b18243143d8d855/torchao/quantization/quant_primitives.py#L78) so that zero_point is None

torchao/dtypes/uintx/cutlass_int4_packed_layout.py
```diff
@@ -0,0 +1,160 @@
...
+        if func is aten.detach.default:
+            return return_and_correct_aliasing(
+                func, args, kwargs, args[0]._apply_fn_to_data(torch.detach)
+            )
+
+        raise NotImplementedError(
+            f"Int4PackedTensorImpl dispatch: attempting to run {func}, this is not supported"
+        )
+
+    def __tensor_flatten__(self):
```
jerryzh168, line 84: we can make zero_point optional and not return zero_point in the list, example: https://github.com/pytorch/ao/blob/f7f20e9e8ba28a4910efb62e7b18243143d8d855/torchao/experimental/_linear_8bit_act_xbit_weight_layout.py#L220, also need to handle this in `__tensor_unflatten__`: https://github.com/pytorch/ao/blob/f7f20e9e8ba28a4910efb62e7b18243143d8d855/torchao/experimental/_linear_8bit_act_xbit_weight_layout.py#L231

jerryzh168 (approved):
LGTM, one minor comment about zero_point

alexsamardzic:
> LGTM, one minor comment about zero_point

Again - made the change suggested, rebased on the latest main.  If CI goes green, will merge.

jerryzh168:
@alexsamardzic for W4A8, does cutlass also support int4 weight and fp8 activation as well?

alexsamardzic:
Yes, it does:
https://github.com/NVIDIA/cutlass/blob/main/examples/55_hopper_mixed_dtype_gemm/55_hopper_int4_fp8_gemm.cu.
But note that it won't be a simple update for this PR, as this one is for
Ampere, but Hopper kernels in CUTLASS are written differently.


On Fri, May 9, 2025, 7:02 AM Jerry Zhang ***@***.***> wrote:
>
> <
https://ci3.googleusercontent.com/meips/ADKq_NZ6ByBM1piM_eYnST5m6EGD-rdfGXsufbXAp4NmJh5L0rlKvYXv7Co984FwOTiDNE7P76uvx5bHyrd9AfzoJj7ckUR3Cff0YF6Yo-d6Xg=s0-d-e1-ft#https://avatars.githubusercontent.com/u/4958441?s=20&v=4>jerryzh168
left a comment (pytorch/ao#880)
>
> @alexsamardzic for W4A8, does cutlass also support int4 weight and fp8
activation as well?
>
> —
> Reply to this email directly, view it on GitHub, or unsubscribe.
> You are receiving this because you were mentioned.<
https://ci3.googleusercontent.com/meips/ADKq_Nao2yPzUQTO2Dy05H037GlrN-RQf7dUaGddk6pblrt_7wFeQVCD1XlDhKvJT6BbZHVb3TzGyXiZfUp8qPb8Hl4fll8fhF8p2iCROaBkPixagUQ00qsJuYGd5HXTZbLvCfYD4nml9i6JCDd3TA8JdUKCw64DELdRXpCq1dOR1KF3mZ8gQP_oxAL4aI_-YS9eMGlaOAjQRRJRxVTsEA_JwvvNTxwbIG-MFWv9kI5Yqnn6BqAjQMD2WCk=s0-d-e1-ft#https://github.com/notifications/beacon/A3U5CMNI4FUZL7XEVAMQTGD25QZF5A5CNFSM6AAAAABODFQLSSWGG33NNVSW45C7OR4XAZNMJFZXG5LFINXW23LFNZ2KUY3PNVWWK3TUL5UWJTVKYYNTO.gif>Message
ID: ***@***.***>

-- 
*Aleksandar Samardžić*
Scientific Software Developer

*Quansight* |  Your Data Experts

w: www.quansight.com <http://www.openteams.com/>  e:
***@***.*** ***@***.***>
<https://www.linkedin.com/company/quansight/>
<https://twitter.com/quansightai>

PR #881, opened 2024-09-12
Add overloads for NF4Tensor to support offloading

janeyx99:
Ran into issues when trying to offload activations when the activation was an NF4Tensor.

This PR adds `empty_like` and the device overload for `to`. There is some code dupe for the `to`, but I'm not sure if the codepath at 977 is ever hit. cc @drisspg @weifengpy

### DIFF
--- a/torchao/dtypes/nf4tensor.py
+++ b/torchao/dtypes/nf4tensor.py
@@ -156,6 +156,7 @@ def clone(func, *args, **kwargs):
 @implements(
     [
         aten.detach.default,
+        aten.empty_like.default,
     ]
 )
 def nf4_detach(aten_op, args, kwargs=None):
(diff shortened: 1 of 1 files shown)


drisspg (commented):
can you write a test pleaseeee

test/dtypes/test_nf4.py
```diff
@@ -270,6 +271,14 @@ def test_chunk_size_equivalence(self, dtype: torch.dtype, shape, chunk_size):
 
         torch.testing.assert_close(nf4_patched.quantized_data, nf4_base.quantized_data)
 
+    @parametrize("input_size", [(512 * 512,), (512, 512)])
+    def test_empty_like(self, input_size: Union[Tuple[int], int]):
+        nf4_tensor = to_nf4(torch.rand(input_size))
+        new_tensor = torch.empty_like(nf4_tensor, device=torch.device("cpu"))
```
jerryzh168, line 277: nit: to make sure `device` arg works, can you make sure the `nf4_tensor` is on gpu?

malfet, line 277: Not sure why you want to construct device here explicitly
```suggestion
        new_tensor = torch.empty_like(nf4_tensor, device="cpu")
```

malfet (approved):
If CI is green...

PR #882, opened 2024-09-12
Fix safe_int_mm for H100

jainapurva:


HDCharles (changes requested):
can you make sure the model runs on the specified device given these changes? it looksl ike they don't. I think you need to go back and if this is an issue with .to(device) after compile, can you change it so it does compile right at the end and does .to(device) right beforehand?

jerryzh168:
> can you make sure the model runs on the specified device given these changes? it looksl ike they don't. I think you need to go back and if this is an issue with .to(device) after compile, can you change it so it does compile right at the end and does .to(device) right beforehand?

@HDCharles should we refactor the quantization code in eval/benchmark/hf_eval to a util? lots of duplications right now I think

jainapurva:
Don't need this fix anymore

PR #884, opened 2024-09-13
Fix WOQ int8 failures

leslie-fang-intel:
Fix WOQ int8 failures as discussed in https://github.com/pytorch/ao/pull/843#issuecomment-2347682042

### DIFF
--- a/test/integration/test_integration.py
+++ b/test/integration/test_integration.py
@@ -826,6 +826,7 @@ class TestSubclass(unittest.TestCase):
     @torch._inductor.config.patch({"freezing": True})
     @unittest.skipIf(not TORCH_VERSION_AT_LEAST_2_4, "freeze requires torch 2.4 and after.")
     def test_int8_weight_only_quant_with_freeze(self, device, dtype):
+        torch._dynamo.reset()
         self._test_lin_weight_subclass_api_impl(
             _int8wo_api, device, 40, test_dtype=dtype
         )
--- a/test/integration/test_integration.py
+++ b/test/integration/test_integration.py
@@ -102,9 +102,7 @@ COMMON_DEVICE_DTYPE = list(itertools.product(COMMON_DEVICES, COMMON_DTYPES)).cop
 def _int8wo_api(mod):
     if TORCH_VERSION_AT_LEAST_2_4:
         quantize_(mod, int8_weight_only(), set_inductor_config=False)
-        if (
-            not TORCH_VERSION_AT_LEAST_2_5
-        ) or torch._inductor.config.freezing:
+        if not TORCH_VERSION_AT_LEAST_2_5:
             unwrap_tensor_subclass(mod)
     else:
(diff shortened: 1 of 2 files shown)


leslie-fang-intel:
cc @jerryzh168 @bdhirsh

test/integration/test_integration.py
```diff
@@ -102,7 +102,9 @@
 def _int8wo_api(mod):
     if TORCH_VERSION_AT_LEAST_2_4:
         quantize_(mod, int8_weight_only(), set_inductor_config=False)
-        if not TORCH_VERSION_AT_LEAST_2_5:
+        if (
+            not TORCH_VERSION_AT_LEAST_2_5
+        ) or torch._inductor.config.freezing:
```
jerryzh168, line 107: when is `freezing` set? how do user know that they need to call `unwrap_tensor_subclass` in that case?

leslie-fang-intel: Hi Jerry, `freezing` is setting in https://github.com/pytorch/ao/blob/8236a874479a9a9168e584c81dda8707f4c41006/test/integration/test_integration.py#L829 when running the UT. User can also set it as `torch._inductor.config.freezing=True` in script or using the environment variable as `TORCHINDUCTOR_FREEZING=1`

leslie-fang-intel: > how do user know that they need to call unwrap_tensor_subclass in that case?

That's a good idea. Did you mean we should write document somewhere?

jerryzh168: I see, thanks for the context! yeah I think it would be good to add this in the README: https://github.com/pytorch/ao/tree/main/torchao/quantization#workaround-with-unwrap_tensor_subclass-for-export-aoti-and-torchcompile-pytorch-24-and-before-only and link the issue as well

leslie-fang-intel: Thanks and adding the readme with issue link in this PR. please help to take a look again.

jerryzh168 (approved):
LGTM

leslie-fang-intel:
Sure, we should land the PR in PyTorch at first.

leslie-fang-intel:
Could you help to take a review https://github.com/pytorch/pytorch/pull/135928?

leslie-fang-intel:
Hi @jerryzh168, since https://github.com/pytorch/pytorch/pull/135928 has been merged into PyTorch, could you approve the CI run for this PR to check if it passes all tests?

leslie-fang-intel:
It seems `test_int8_weight_only_quant_subclass_api` with `float16` inputs still failed which is probably due to numerical issue. Will take a further look.

leslie-fang-intel:
Creating another PR: https://github.com/pytorch/pytorch/pull/136353 to fix the Float16 input failure.

leslie-fang-intel:
Hi @jerryzh168, the CI are all green now. I am going to land this PR. Thanks.

PR #886, opened 2024-09-13
Add Embedding Quantization to QAT module_swap flow

TiRune:
Summary: Adding the embedding quantizer in the same fashion as the other module swap setup.

Differential Revision: D62664322


torchao/quantization/GPTQ.py
```diff
@@ -965,6 +965,41 @@ def forward(self, input: torch.Tensor) -> torch.Tensor:
             self.precision,
         )
 
+	
+def _replace_embedding_4w(
```
jerryzh168, line 969: I'm wondering if this can be added at the user code side, since we are planning to deprecate the module swap API

TiRune: Please don't deprecate the module swap API - it's the easiest to work with and extend.
I'll likely have a headache if I wanted to make things work quickly and effectively with the tensor subclass stuff. 

If you guys have a few minutes, we can see how to add this to the tensor subclass stuff as well... but...

jerryzh168: OK, I think keeping multiple implementations of the same thing might be confusing, we can gather all the requirements and decide on the long term plan I think, I'm asking Andrew to take a stab first

PR #897, opened 2024-09-16
Add torchchat quantizer

metascroy:
Summary:
This diff adds a quantizer for the new torchao kernels that is similar to the Int8DynActInt4WeightQuantizer quantizer in torchchat (imported from from torchao.quantization.quant_api).  See the draft torchchat PR (https://github.com/pytorch/torchchat/pull/1070/) for how this can integrate with torchchat's quantization API.

I confirmed that models quantized with this are compatible with eager, compile, AOTI, and export to ExecuTorch in torchchat.  They do not run on ExecuTorch because we still have not written an ExecuTorch kernel wrapper.

jerryzh168 this does not use the new subclass API, and this is something I'd like to discuss further with you.  I'll set up a sync with you this week, but I wanted to have some API on the table to ground the discussion.

We do not currently have the required C++ methods implemented to support the new subclass API (e.g., we cannot unpack the packed weights from python; they are instead unpacked inline in the kernel).  From a torchchat user's perspective, I do not think this is important, but I'd like to discuss further.

Differential Revision: D62394341


torchao/experimental/quant_api.py
```diff
@@ -0,0 +1,277 @@
...
+                    + "Falling back to **slow** implementation _Int8DynActIntxWeightQuantizedLinearFallback."
+                )
+                qlinear = _Int8DynActIntxWeightQuantizedLinearFallback()
+                setattr(module, name, qlinear)
+                getattr(module, name).quantize_and_pack_weights(
+                    child.weight, nbit, group_size, has_weight_zeros
+                )
+
+
+class Int8DynActIntxWeightQuantizer:
```
cccclai, line 223: nit: the file name sounds like an interface but the content looks like all different kinds of implementation....

cccclai (approved):
Stamp because @digantdesai already approved internally

PR #899, opened 2024-09-17
Minor fix for logical operators precedence in _aqt_is_* checks.

alexsamardzic:


torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -1119,24 +1119,24 @@ def _aqt_is_int8(aqt):
...
+        (aqt.quant_min is None or aqt.quant_min == -128) and
+        (aqt.quant_max is None or aqt.quant_max == 127)
     )
 
 def _aqt_is_int8_reduced_range(aqt):
     return (
         aqt.layout_tensor.dtype == torch.int8 and
-        aqt.quant_min == -127 and
-        aqt.quant_max is None or aqt.quant_max == 127
+        (aqt.quant_min is not None and aqt.quant_min == -127) and
```
jerryzh168, line 1129: nit: probably don't need to check `not None` since `quant_min == -127` is stronger than that

torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -1119,24 +1119,24 @@ def _aqt_is_int8(aqt):
...
 
 def _aqt_is_uint4(aqt):
     """Check if an AffineQuantizedTensor is uint4 quantized Tensor"""
     # TODO: use torch.uint4
     return (
         aqt.layout_tensor.dtype == torch.int32 and
-        aqt.quant_min is None or aqt.quant_min == 0 and
-        aqt.quant_max is None or aqt.quant_max == 15
+        (aqt.quant_min is not None and aqt.quant_min == 0) and
+        (aqt.quant_max is not None and aqt.quant_max == 15)
```
jerryzh168, line 1139: same here, I think we can remove the `not None` checks

torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -1119,24 +1119,24 @@ def _aqt_is_int8(aqt):
...
 def _aqt_is_int8_reduced_range(aqt):
     return (
         aqt.layout_tensor.dtype == torch.int8 and
-        aqt.quant_min == -127 and
-        aqt.quant_max is None or aqt.quant_max == 127
+        (aqt.quant_min is not None and aqt.quant_min == -127) and
+        (aqt.quant_max is None or aqt.quant_max == 127)
     )
 
 def _aqt_is_uint4(aqt):
```
jerryzh168, line 1133: also probably would be better to change the name to something like `_aqt_is_tensor_core_tiled_uint4` (since this is the format expected by that layout, so users don't confuse this with a "general uint4 check"

jerryzh168 (approved):
looks good, thanks!

PR #902, opened 2024-09-18
Add example code for printing the operator and shapes in a model

jerryzh168:
Summary:
This will be useful for people to do understand the ops/shapes for a model that they are interested in optimizing, also helpful for microbenchmarks with target ops/shapes. This is used in https://github.com/pytorch/ao/issues/391 (search for `Microbenchmarks`)

Example output:
```
TORCH_FUNC=<built-in function linear> (M, K, N): 10 10 10
TORCH_FUNC=<method 'add' of 'torch._C.TensorBase' objects> args[0] shape: torch.Size([10, 10])

all linear shapes (M, K, N): [(10, 10, 10)]
```

Test Plan:
python tutorials/developer_api_guide/print_op_and_shapes.py

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/tutorials/developer_api_guide/print_op_and_shapes.py
+++ b/tutorials/developer_api_guide/print_op_and_shapes.py
@@ -1,0 +1,35 @@
+import torch
+
+linear_shapes = []
+from torch.overrides import TorchFunctionMode
+class TorchFunctionLoggingMode(TorchFunctionMode):
+    def __torch_function__(cls, func, types, args=(), kwargs=None):
+        if kwargs is None:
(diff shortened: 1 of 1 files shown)


msaroufim:
Can you show us what this prints on some model? Also why do this vs use autoquant?

vkuzo (approved):
lgtm!

jerryzh168:
> Can you show us what this prints on some model?

added in summary, this util is used in the torchao dev guide as a util for developers

PR #904, opened 2024-09-19
[float8] improve eager numerics for dynamic scales and gets on par with torch.compile

weifengpy:
Turning on/off float8 all-gather shows different numerics: https://github.com/pytorch/ao/pull/873

**Root cause**: `hp_tensor_to_float8_dynamic` have different numerics in eager vs compile
* compile does float32 upcasting at the begining of scale calculation, see `tmp0 = tl.load(in_ptr0 + (x0), None).to(tl.float32)` in triton kernels below
* eager does float32 upcasting at the end
* even if we do float32 upcasting at the begining for both compile and eager, triton has different numerics than eager. but float64 has bitwise on-par numerics

This PR does float64 upcasting to the begining of scale calculation

unit test: `TORCH_LOGS="output_code" pytest -s test/float8/test_base.py -k test_dynamic_scale_parity`

```
tmp0 = tl.load(in_ptr0 + (x0), None).to(tl.float32)
tmp1 = tl.load(in_ptr1 + (0))
tmp2 = tl.broadcast_to(tmp1, [XBLOCK])
tmp3 = tmp2.to(tl.float32)
tmp4 = tmp0 * tmp3
tmp5 = tmp4.to(tl.float32)
tmp6 = -448.0
tmp7 = triton_helpers.maximum(tmp5, tmp6)
tmp8 = 448.0
tmp9 = triton_helpers.minimum(tmp7, tmp8)
tmp10 = tmp9.to(tl.float32)
tmp11 = tmp10.to(tl.float8e4nv)
tl.store(out_ptr0 + (x0), tmp11, None)
```


test/float8/test_base.py
```diff
@@ -53,7 +56,7 @@
 is_cuda_8_9 = torch.cuda.is_available() and torch.cuda.get_device_capability() >= (8, 9)
 
 def bitwise_identical(a: Float8Tensor, b: Float8Tensor) -> bool:
-    assert torch.all(a._data == b._data).item(), "scales are not identical"
```
weifengpy, line 56: `._data` might be a typo in trunk since the error msg is about scales

test/float8/test_base.py
```diff
@@ -604,6 +607,35 @@ def test_small_amax_float16(self, float8_dtype):
...
+            float8_config,
+            gemm_input_role=GemmInputRole.WEIGHT,
+        )
+        float8_compile = torch.compile(hp_tensor_to_float8_dynamic)(
+            hp_tensor,
+            torch.float8_e4m3fn,
+            float8_config,
+            gemm_input_role=GemmInputRole.WEIGHT,
+        )
+        assert bitwise_identical(float8_eager, float8_compile)
```
weifengpy, line 638: without the PR, the numerics looks like following
eager `_scale=109.625` vs  compile `_scale=109.6444`

torchao/float8/float8_utils.py
```diff
@@ -42,6 +42,9 @@ def amax_to_scale(
         float8_dtype: The float8 dtype.
         orig_dtype: The original dtype of the tensor.
     """
+    # Preserve precision in amax-to-scale conversion
+    # and ensure on-par numerics with torch.compile
+    amax = amax.to(torch.float64)
```
weifengpy, line 47: I choose to upcast `amax` in `amax_to_scale` so we can still do bfloat16 all-reduce for amax. The change is also safer to `amax_buffer` in delayed scaling

test/float8/test_base.py
```diff
@@ -604,6 +607,39 @@ def test_small_amax_float16(self, float8_dtype):
...
+            float8_config,
+            gemm_input_role=GemmInputRole.WEIGHT,
+        )
+        float8_compile = torch.compile(hp_tensor_to_float8_dynamic)(
+            hp_tensor,
+            torch.float8_e4m3fn,
+            float8_config,
+            gemm_input_role=GemmInputRole.WEIGHT,
+        )
+        assert torch.equal(float8_eager._scale, float8_compile._scale)
```
weifengpy, line 642: without the PR, the numerics looks like following
eager _scale=109.625 vs compile _scale=109.6444

torch.compile has better numerics and match exactly with float64

torchao/float8/float8_utils.py
vkuzo: could you share why the upcasting happens?

awgu:
I am a bit surprised that we have to use fp64 in eager. Without the upcast to fp64, what is the dtype being used?

torchao/float8/float8_utils.py
weifengpy: I can look into inductor more on how it  achieved fp64

weifengpy:
> I am a bit surprised that we have to use fp64 in eager. Without the upcast to fp64, what is the dtype being used?

eager respect param dtype - if param is bf16, amax is bf16, scale compute is bf16. scale gets upcasted to fp32 at the very end

weifengpy:
converting to draft as upcasting to torch.float32 in eager should be enough. digging into inductor as there might be nuances around how bfloat16 is handled

torchao/float8/float8_tensor.py
```diff
@@ -163,7 +163,8 @@ def forward(
 
         DTensor Invariant: DTensor must always be the outer most tensor subclass
         """
-        tensor_scaled = tensor * scale
+        # scale is float32 thus upcasting tensor to match
+        tensor_scaled = tensor.to(torch.float32) * scale
```
weifengpy, line 167: without upcasting, the eager numeric is like `-157.00000000000000000000`, compile is like `-157.06507873535156250000`

weifengpy: torch.compile upcast tensor ahead, see `tmp0 = tl.load(in_ptr0 + (x0), None).to(tl.float32)` in following output code

```
@triton.jit
def triton_(in_ptr0, in_ptr1, out_ptr0, xnumel, XBLOCK : tl.constexpr):
    xnumel = 24576
    xoffset = tl.program_id(0) * XBLOCK
    xindex = xoffset + tl.arange(0, XBLOCK)[:]
    xmask = tl.full([XBLOCK], True, tl.int1)
    x0 = xindex
    tmp0 = tl.load(in_ptr0 + (x0), None).to(tl.float32)
```

torchao/float8/float8_utils.py
weifengpy: torch.compile actually upcasts to float32 with `tl.load(in_ptr0 + (x0), None).to(tl.float32)`. Upcasting to float64 further help because torch.compile and eager shows different numerics for 1.0 / float32 (but same numeric for float64)

The float32 numeric difference can be verified with 
```
import torch
def upcast_reciprocal(inp: torch.Tensor):
    return inp.reciprocal()
inp = torch.full([], 0.00817871093750000000, device="cuda", dtype=torch.float32)
eager_scale = upcast_reciprocal(inp)
compile_scale = torch.compile(upcast_reciprocal)(inp)
fp64_ground_truth = inp.to(torch.float64).reciprocal()
assert torch.equal(eager_scale, compile_scale), f"{eager_scale=} vs {compile_scale=}, {fp64_ground_truth=}"
```

test/float8/test_base.py
```diff
@@ -604,6 +607,40 @@ def test_small_amax_float16(self, float8_dtype):
...
+    )
+    @pytest.mark.parametrize(
+        "dtype",
+        [
+            torch.float32,
+            torch.bfloat16,
+            torch.float16,
+        ],
+    )
+    def test_dynamic_scale_parity(self, dtype: torch.dtype):
```
vkuzo, line 623: nit: move to `test_compile.py` since this is testing compile vs eager?

torchao/float8/float8_tensor.py
```diff
@@ -163,7 +163,8 @@ def forward(
 
         DTensor Invariant: DTensor must always be the outer most tensor subclass
         """
-        tensor_scaled = tensor * scale
+        # scale is float32 thus upcasting tensor to match
```
vkuzo, line 166: can we make this comment contain the context? something like

```
# Note: when the line below is compiled with `torch.compile`, `tensor` is automatically upcasted to `float32` to multiply with the scale
# In order to match numerics between eager and compile, we upcast manually here.
```

torchao/float8/float8_utils.py
```diff
@@ -42,6 +42,8 @@ def amax_to_scale(
         float8_dtype: The float8 dtype.
         orig_dtype: The original dtype of the tensor.
     """
+    # _scaled_mm requires float32 scale
```
vkuzo, line 45: nit: can we describe in more detail why we are upcasting here

torchao/float8/fsdp_utils.py
```diff
@@ -59,17 +59,17 @@ def precompute_float8_dynamic_scale_for_fsdp(module: nn.Module) -> None:
         return
 
     # inf-norm is equivalent to max(abs(w))
-    max_weights = torch._foreach_norm(weights, ord=math.inf)  # Partial
+    max_weights = torch._foreach_norm(weights, ord=math.inf, dtype=torch.float64)  # Partial
```
vkuzo, line 62: add comment to describe upcasting

vkuzo:
thanks for finding this, it would be great to have numerics match! At a high level, looks good.  Could we add two more things,
1. verify that there is no performance regression with torch.compile
2. ensure tests pass in CI

weifengpy:
> thanks for finding this, it would be great to have numerics match! At a high level, looks good. Could we add two more things,
> 
> 1. verify that there is no performance regression with torch.compile
> 2. ensure tests pass in CI

thanks for the feedback. will address them and publish for review

weifengpy:
verified perf is neutral and make sure all CI passes

for perf, I ran `benchmark_compiled_module(times=100, repeat=100)` from the output code of `pytest -s test/float8/test_compile.py -k test_dynamic_scale_numeric_parity`. I tested on shape = (1024, 1024), 
* without the change:  `0.000035 secs` P1612792187
* with the change: `0.000035 secs` P1612791058

looking into the output code, the difrerence is mainly `tmp3.to(tl.float64)` or `tmp3.to(tl.float32)`

weifengpy:
addressed feedback and is ready for review

torchao/float8/fsdp_utils.py
weifengpy: improved comment

torchao/float8/fsdp_utils.py
```diff
@@ -59,17 +59,20 @@ def precompute_float8_dynamic_scale_for_fsdp(module: nn.Module) -> None:
         return
 
     # inf-norm is equivalent to max(abs(w))
-    max_weights = torch._foreach_norm(weights, ord=math.inf)  # Partial
+    # keep consistent with float8_utils.amax_to_scale
+    # torch.compile and eager show different numerics for 1.0 / float32,
+    # upcast to float64 to ensure same numeric between compile and eager
+    max_weights = torch._foreach_norm(weights, ord=math.inf, dtype=torch.float64)  # Partial
```
awgu, line 65: `dtype=torch.float64` only changes the accumulation dtype? if there is no noticeable cost to this, I wonder if we should be doing this in more places 🤔

weifengpy: good question! Actually I just updated the code to do `_foreach_norm` in original precision and do float64 upcasting before calculating scales. That ensures consistent implementation between prcompute and `float8_utils.amax_to_scale`

back to your question, I checked `ForeachReduceOp.cu` and it's dispatching to lpnorm_cleanup<scalar_t, NormType::LInf, out_t>. Not sure what's inside `lpnorm_cleanup`. But inf-norm is just max(abs) so not sure if they accumulate numerics
https://github.com/pytorch/pytorch/blob/a28b40fa74470058ca57d77652b9601bece2f4d5/aten/src/ATen/native/cuda/ForeachReduceOp.cu#L534-L535C19

torchao/float8/fsdp_utils.py
```diff
@@ -59,17 +59,20 @@ def precompute_float8_dynamic_scale_for_fsdp(module: nn.Module) -> None:
...
+    # keep consistent with float8_utils.amax_to_scale
+    # torch.compile and eager show different numerics for 1.0 / float32,
+    # upcast to float64 to ensure same numeric between compile and eager
     max_weights = torch._foreach_norm(weights, ord=math.inf)  # Partial
     amax_tensor = torch.stack(max_weights)  # Partial
     # clamp is dispatched through DTensor
     # it will issue a single all-reduce
     amax_tensor = torch.clamp(amax_tensor, EPS)  # Replicate
-    scale_tensor = torch.finfo(torch.float8_e4m3fn).max / amax_tensor  # Replicate
+    scale_tensor = torch.finfo(torch.float8_e4m3fn).max / amax_tensor.to(torch.float64)  # Replicate
```
vkuzo, line 70: optional nit: upcast separately to make easier to read

```
amax_tensor = amax_tensor.to(torch.float64)
scale_tensor = ... / amax_tensor
```

weifengpy: updated as suggested

test/float8/test_compile.py
```diff
@@ -353,5 +359,42 @@ def test_sync_amax_func_cuda_graph_success():
...
+        torch.bfloat16,
+        torch.float16,
+    ],
+)
+def test_dynamic_scale_numeric_parity(dtype: torch.dtype):
+    scaling_type_weight = ScalingType.DYNAMIC
+    torch.manual_seed(42)
+    hp_tensor1 = torch.randn(16, 16, device="cuda", dtype=dtype)
+    hp_tensor2 = hp_tensor1.detach().clone()
+    float8_config = Float8LinearConfig(
```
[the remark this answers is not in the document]
weifengpy: good catch! I switched to `LinearMMConfig`. `Float8LinearConfig` was working because I did not call matmul that requires access to `self._linear_mm_config`.

vkuzo (approved):

PR #909, opened 2024-09-20
Rename Floating point to fp8

jainapurva:
The kernel checks for fp8 input and fp8 weights, hence need to rename from fp to fp8.


vkuzo (changes requested):
I'd recommend using `float8` everywhere to align with PyTorch convention of `float32`, `float16`, etc.  Fp8 may be more natural in how people talk about this, but IMO in this case consistency is better to remove all questions and have everything be named the same way across core and AO.

vkuzo (approved):
sorry, I meant to accept :) my feedback was for future work.

jerryzh168 (approved):

PR #914, opened 2024-09-21
Gptq mt refactor

danielpatrickhug:
This is a PR for the ongoing GPTQ MultiTensor refactor.
I added the e2e test and some atomic unit tests for the MultiTensor Class and the Int4WeightOnlyGPTQQuantizer class. I also refactored the non_in_place logic for readability and sped up the quantization. Finally, I added a new inputRecoder that utilized Muti Tensor and not MultiInput. @HDCharles


HDCharles (approved):
nit: the multitensor probably need a bit more though its not incredible urgent

I would check that

A) MultiTensor.pad, the tensor id's differ post padding
B) if you do an in place op with multiple inputs you get the expected (multiple) changes [this was the issue before, you'd only get the change from whichever ran list so testing this would be important]
C) if you do a not in place op, like functional.linear(MultiTensor(a1, a2), MultiTensorWeight(w1)) the final state of the weight is MultiTensorWeight(w1) rather than MultiTensorWeight(w1, w2) like it would be for an in place op.

test/quantization/test_quant_api.py
```diff
@@ -338,34 +338,35 @@ def test_8da4w_quantizer_eval(self):
...
+    # @unittest.skip("skipping until we get checkpoints for gpt-fast")
     def test_gptq_quantizer_int4_weight_only(self):
-        from torchao.quantization.GPTQ import Int4WeightOnlyGPTQQuantizer
-        from torchao._models._eval import InputRecorder, TransformerEvalWrapper
+        from torchao.quantization.GPTQ_MT import Int4WeightOnlyGPTQQuantizer
+        from torchao._models._eval import MultiTensorInputRecorder, TransformerEvalWrapper
         precision = torch.bfloat16
         device = "cuda"
-        checkpoint_path = Path("../checkpoints/meta-llama/Llama-2-7b-chat-hf/model.pth")
+        checkpoint_path = Path("checkpoints/meta-llama/Llama-2-7b-chat-hf/model.pth")
```
HDCharles, line 347: note by default tests are run from the tests dir and normally checkpoints are in the root of torchao after prepare.sh so this should probably be the previous path

msaroufim:
Haven't followed up here, is this good to merge?

test/quantization/test_quant_api.py
```diff
@@ -338,10 +338,10 @@ def test_8da4w_quantizer_eval(self):
             f"accuracy regressed from 8.23 to {result['results']['wikitext']['word_perplexity,none']}"
         )
 
-    @unittest.skip("skipping until we get checkpoints for gpt-fast")
+    # @unittest.skip("skipping until we get checkpoints for gpt-fast")
```
HDCharles, line 341: yeah i would add this back in, i removed it for my own manual testing but CI doesn't have lm_eval or the llama checkpoints so we run these manually.

HDCharles:
> Haven't followed up here, is this good to merge?

i've been talking to AIHD on discord mostly, i think we're trying to pass CI and then it'll be ready to merge

PR #918, opened 2024-09-23
[float8] fix typo in bitwise_identical unit test

weifengpy:
bitwise_identical is used in `pytest -s test/float8/test_base.py`. fix typos around checking scales


vkuzo (approved):
nice catch!

PR #923, opened 2024-09-23
clean up device checks in float8 unit test files

vkuzo:
Summary:

While working on rowwise scaling I noticed that some of the CUDA device capability checks we had in the test files did not make sense, cleaning this up.

Test Plan:

tests pass on my H100

CI, it should skip less tests now since CI only has CUDA capability 8, 9

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/float8/test_base.py
+++ b/test/float8/test_base.py
@@ -231,15 +231,6 @@ class TestFloat8Linear:
         linear_dtype: torch.dtype,
         linear_bias: bool,
     ):
-        if not emulate:
-            if not torch.cuda.is_available():
-                warnings.warn("CUDA not available")
-                pytest.skip()
-            elif torch.cuda.get_device_capability() < (9, 0):
-                warnings.warn(
-                    f"CUDA capability {torch.cuda.get_device_capability()} < (9.0)"
-                )
-                pytest.skip()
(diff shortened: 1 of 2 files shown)


test/float8/test_base.py
```diff
@@ -231,15 +231,6 @@ def test_linear(
         linear_dtype: torch.dtype,
         linear_bias: bool,
     ):
-        if not emulate:
-            if not torch.cuda.is_available():
-                warnings.warn("CUDA not available")
-                pytest.skip()
-            elif torch.cuda.get_device_capability() < (9, 0):
```
vkuzo, line 238: this line made this test skip in CI, not intended, removing.  Similar cases in the rest of this file.

test/float8/test_base.py
```diff
@@ -334,10 +315,6 @@ def test_autocast_outputs(
     @pytest.mark.parametrize("emulate", [True, False] if is_cuda_8_9 else [True])
     @unittest.skipIf(not torch.cuda.is_available(), "CUDA not available")
     def test_type_cast(self, linear_dtype: torch.dtype, emulate: bool):
-        emulate = (
```
vkuzo, line 337: the parametrization and skipping if cuda not available should already accomplish this

msaroufim (approved):

PR #928, opened 2024-09-24
Fix failing FP6 benchmark

tobiasvanderwerff:
Summary: 
I was playing around with FP6 and noticed that its benchmark script is no longer functional. It seems that it's still using an older API. This PR fixes the script in accordance with the FP6 correctness test in `ao/test/test_ops.py`:

https://github.com/pytorch/ao/blob/653efe98749124985155da494a372bea7fe4b383/test/test_ops.py#L62-L70

Here is the utility function `_create_floatx_inputs`:

https://github.com/pytorch/ao/blob/653efe98749124985155da494a372bea7fe4b383/test/test_ops.py#L36-L42

### DIFF
--- a/benchmarks/benchmark_fp6.py
+++ b/benchmarks/benchmark_fp6.py
@@ -1,7 +1,7 @@
 import torch
 import pandas as pd
 import torchao
-from torchao.dtypes.floatx import from_scaled_tc_floatx
+from torchao.dtypes.floatx import from_scaled_tc_floatx, to_scaled_tc_floatx
 from torchao.utils import benchmark_torch_function_in_microseconds
 from tqdm import tqdm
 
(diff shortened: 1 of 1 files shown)


jerryzh168:
thanks, this is using a higher level API, not old API. but it makes sense to keep consistent with test_ops I think

jerryzh168 (dismissed):

jerryzh168:
actually, maybe just adding a comment explaining this is enough, it's easier for end users to use the high level `quantize_` API and see the perf, since this is what they will be calling for the models in the end. see "Kernel Microbenchmarks" in https://github.com/pytorch/ao/issues/391 for how these benchmark script fit into the optimization process.

tobiasvanderwerff:
Thanks for the feedback. I don't think I made it clear enough, but currently the `fp6_benchmark.py` script is failing because of this line:

https://github.com/pytorch/ao/blob/858205a44fa6c10d84fa6084bfd8c526f1b85914/benchmarks/benchmark_fp6.py#L12

The `to_affine_quantized_floatx` function is not called with the right arguments (it expects 5 of them, but it passes only two), so in that sense the current benchmark script is outdated and worth updating. What do you think?

jerryzh168:
> Thanks for the feedback @jerryzh168. I don't think I made it clear enough, but currently the `fp6_benchmark.py` script is failing because of this line:
> 
> https://github.com/pytorch/ao/blob/858205a44fa6c10d84fa6084bfd8c526f1b85914/benchmarks/benchmark_fp6.py#L12
> 
> The `to_affine_quantized_floatx` function is not called with the right arguments (it expects 5 of them, but it passes only two), so in that sense the current benchmark script is outdated and worth updating. What do you think?

oh I see, yeah makes sense to update this, it should be `to_affine_quantized_fpx` I think, the renaming PR broken this..

tobiasvanderwerff:
That makes sense -- let me update the script with your change, @jerryzh168. Do you think it makes sense to open a new PR and close the current one?

tobiasvanderwerff:
I tested the change (it works) and created a new PR for it: #931. Closing this PR.

PR #931, opened 2024-09-24
Fix failing FP6 benchmark

tobiasvanderwerff:
The FP6 benchmark script `benchmarks/benchmark_fp6.py` is broken because of [this renaming PR](https://github.com/pytorch/ao/commit/8236a874479a9a9168e584c81dda8707f4c41006). The current PR fixes this by using the correct call to `to_affine_quantized_fpx` instead of `to_affine_quantized_floatx`.

https://github.com/pytorch/ao/blob/858205a44fa6c10d84fa6084bfd8c526f1b85914/benchmarks/benchmark_fp6.py#L12

Originally discussed [here](https://github.com/pytorch/ao/pull/928)

@jerryzh168

### DIFF
--- a/benchmarks/benchmark_fp6.py
+++ b/benchmarks/benchmark_fp6.py
@@ -1,7 +1,7 @@
 import torch
 import pandas as pd
 import torch.nn.functional as F
-from torchao.dtypes import to_affine_quantized_floatx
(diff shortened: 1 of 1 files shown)


jerryzh168 (approved):
thanks!

PR #933, opened 2024-09-24
[float8] all-reduce amax on dp mesh instead of global pg

weifengpy:
this PR makes sure FSDP2 + pipeline parallel work. For amax all-reduce, we should do allreduce on dp mesh, instead of global pg. otherwise it causes NCCL hanging

### DIFF
--- a/test/float8/test_fsdp2/test_fsdp2.py
+++ b/test/float8/test_fsdp2/test_fsdp2.py
@@ -312,7 +312,7 @@ class Test2DParallelMultiThread(FSDPTestMultiThread, TestFloat8Common):
 
         if self.rank in [0, 1]:
             # rank 0 and 1 are the 1st stage in the pipeline
-            # rank 2 and 4 are doing thing but waiting for the 1st stage
+            # rank 2 and 4 are doing nothing but waiting for the 1st stage
             float8_tensor = hp_tensor_to_float8_dynamic(
                 hp_tensor,
                 torch.float8_e4m3fn,
--- a/test/float8/test_fsdp2/test_fsdp2.py
+++ b/test/float8/test_fsdp2/test_fsdp2.py
@@ -17,10 +17,12 @@ import torch.distributed as dist
 import torch.nn as nn
 from torchao.float8.config import CastConfig, Float8LinearConfig, ScalingType
 from torchao.float8.float8_linear_utils import convert_to_float8_training
+from torchao.float8.float8_scaling_utils import hp_tensor_to_float8_dynamic
 from torchao.float8.fsdp_utils import WeightWithDynamicFloat8CastTensor
 from fsdp2_common import check_parity_bf16_mp, check_parity_no_mp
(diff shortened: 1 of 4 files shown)


weifengpy:
cc @tianyu-l @H-Huang

test/float8/test_fsdp2/test_fsdp2.py
```diff
@@ -293,6 +295,34 @@ def _get_curr_active_memory_mb(self) -> int:
...
+            hp_tensor = torch.randn(768, 32, device="cuda")
+            float8_tensor = hp_tensor_to_float8_dynamic(
+                hp_tensor,
+                torch.float8_e4m3fn,
+                Float8LinearConfig(
+                    cast_config_weight=CastConfig(scaling_type=ScalingType.DYNAMIC),
+                ),
+                gemm_input_role=GemmInputRole.WEIGHT,
+                reduce_amax=True,
+                device_mesh=dp_mesh
```
weifengpy, line 323: without the PR,  the unit test will throw NCCL timeout error because rank 2 and 4 did not participate in amax all-reduce

torchao/float8/float8_scaling_utils.py
```diff
@@ -36,6 +36,7 @@ def hp_tensor_to_float8_dynamic(
     linear_mm_config: LinearMMConfig,
     reduce_amax: bool = False,
     gemm_input_role: GemmInputRole = GemmInputRole.INPUT,
+    device_mesh = None,
```
weifengpy, line 39: DeviceMesh only exists in newer pytorch (2.4+). I tried to annotate device_mesh with string `"torch.distributed.DeviceMesh"`. but some CI run torch older pytorch 2.2 and complain about not being able to find DeviceMesh. I gave up on type hint. hopefully device_mesh itself is a strong indicator of DeviceMesh

awgu (approved):
SGTM!

will let @vkuzo make a pass too

torchao/float8/float8_scaling_utils.py
vkuzo: for now, we are intending to only test float8 code on the latest nightly to move fast and minimize maintenance load.  Do you know CI is giving you issues?  We have disabled CI on older PT versions per-test-file using version checks.

vkuzo (approved):

torchao/float8/float8_scaling_utils.py
weifengpy: I belive it's the following one. My guess - the function signature is checked at import time. For torch==2.2.2, if one of the test imported `float8_utils.py` or `float8_scaling_utils.py`, it will throw error

```
[test (CUDA 2.2.2, linux.g5.12xlarge.nvidia.gpu, torch==2.2.2 "numpy<2" , cuda, 12.1) / linux-job](https://github.com/pytorch/ao/actions/runs/11023026757/job/30613550360?pr=933#logs)
```

PR #934, opened 2024-09-24
Add new evaluation metrics

jainapurva:
Added new tests to llama/eval.sh for more extensive testing on larger metrics


jerryzh168:
if not sure if all eval metrics are relevant for llama, cc @HDCharles to take a look

HDCharles (changes requested):
see comment

jainapurva:
> the current solution is specifying all tests explicitly in benchmarks.sh and we made evals.sh to do the same though I haven't added all the tests there yet. If you want to make it easier to run those sets i'd maybe add them there?

I think we need more eval metrics to test the different techniques on llama. Though that could be done by updating the eval.sh instead. Also, we can add more benchmarks to the readme file in future.

drisspg:
@HDCharles just to confirm this is what you have in mind right?

HDCharles (approved):
this looks better

PR #938, opened 2024-09-24
Add composable QAT quantizer

andrewor14:
**Summary:** This is a utility for users who wish to apply multiple QAT quantizers to their models. In the near future, we expect to add an embedding QAT quantizer that composes with the existing linear QAT quantizers.

**Test Plan:**
python test/quantization/test_qat.py -k test_composable_qat_quantizer

### DIFF
--- a/torchao/quantization/prototype/qat/api.py
+++ b/torchao/quantization/prototype/qat/api.py
@@ -4,7 +4,7 @@
 # This source code is licensed under the license found in the
 # LICENSE file in the root directory of this source tree.
 
-from typing import Any, Optional
+from typing import Any, List, Optional
 
 import torch
 import torch.nn.functional as F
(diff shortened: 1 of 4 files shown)


torchao/quantization/prototype/qat/api.py
```diff
@@ -34,6 +34,41 @@
...
+    ) -> torch.nn.Module:
+        for quantizer in self.quantizers:
+            model = quantizer.prepare(model)
+        return model
+
+    def convert(
+        self, model: torch.nn.Module, *args: Any, **kwargs: Any
+    ) -> torch.nn.Module:
+        for quantizer in self.quantizers:
+            model = quantizer.convert(model)
```
jerryzh168, line 68: should the order be reserved

andrewor14: discussed offline: the quantizers are supposed to apply to different modules in the same model. Added some docs to clarify instead

jerryzh168 (approved):

PR #939, opened 2024-09-24
Supporting tensor parallelism for int8 weight only quant

jerryzh168:
Summary:
following https://github.com/pytorch/ao/blob/main/tutorials/developer_api_guide/tensor_parallel.py we can support tensor parallelism for int8 weight only quant, this is needed for torchchat

Test Plan:
python test/dtypes/test_affine_quantized_tensor_parallel.py

Reviewers:

Subscribers:

Tasks:

Tags:


kwen2501 (approved):
LGTM!
Wow, didn't thought that would be so straightforward!
Nice!

PR #941, opened 2024-09-24
fixing some issues with our support for 70/405B models

HDCharles:
Summary: download and convert scripts needed to be updated alongside model.py config files

Test Plan: python generate.py --checkpoint_path ../../../checkpoints/meta-llama/Meta-Llama-3.1-70B/model.pth

Reviewers:

Subscribers:

Tasks:

Tags:


jcaip (approved):
lgtm, will add nm checkpoint to the prepare script tonight

PR #223, opened 2024-05-07
Add FP16Act-FP6Weight Linear

gau-nernst:
Closes #208 

References:
- Paper: https://arxiv.org/abs/2401.14112
- Original implementation: https://github.com/usyd-fsalab/fp6_llm
- DeepSpeed implementation: https://github.com/microsoft/DeepSpeed/commit/ccfdb84e2a4a373ac657a99afd2d97e1d741b22b

TODO:
- [x] Port FP6 weight pre-packing
- [x] Port FP16Act-FP6Weight Linear kernel
- [x] Port FP6 -> FP16 dequantization
- [x] Port FP16 -> FP6 quantization
- [x] Add correctness check (wrt FP16-FP16 matmul) - Note: compare against dequant FP16 weight, not original weight
- [x] Add speed benchmarks
- [ ] Add split-k map from DeepSpeed: https://github.com/microsoft/DeepSpeed/blob/0b224edcf7d83713b95ad6b989694a8bdf01809e/deepspeed/inference/v2/kernels/core_ops/cuda_linear/cuda_linear.py (TODO in a future PR - autotune?)

`benchmarks/benchmark_fp6.py` results - 4070 Ti SUPER, PyTorch 2.3, CUDA 12.1

|    m |     k |     n |   fp6_latency (ms) |   fp16_latency (ms) |   speedup (d/s) |   correct |
|-----:|------:|------:|-------------------:|--------------------:|----------------:|----------:|
|    1 | 10240 |  8192 |          0.10459   |            0.272524 |        2.60565  |         1 |
|    1 |  8192 |  8192 |          0.0259702 |            0.218884 |        8.42828  |         1 |
|    1 | 57344 |  8192 |          0.590541  |            1.49265  |        2.5276   |         1 |
|    1 |  8192 | 28672 |          0.286375  |            0.773923 |        2.70248  |         1 |
|    2 | 10240 |  8192 |          0.105072  |            0.272984 |        2.59807  |         1 |
|    2 |  8192 |  8192 |          0.0283768 |            0.221068 |        7.79045  |         1 |
|    2 | 57344 |  8192 |          0.588764  |            1.50597  |        2.55785  |         1 |
|    2 |  8192 | 28672 |          0.289398  |            0.749859 |        2.5911   |         1 |
|    4 | 10240 |  8192 |          0.106225  |            0.273336 |        2.57318  |         1 |
|    4 |  8192 |  8192 |          0.0352987 |            0.221425 |        6.27289  |         1 |
[...]

Issue #208, opened 2024-04-28
FP6 dtype!


NicolasMejiaPetit:
### 🚀 The feature, motivation and pitch

https://arxiv.org/abs/2401.14112

I think you guys are really going to like this.
The deepspeed developers introduce FP6 datatype on cards without fp8 support, while maintaining full tensor core suppourt using a kernel they created called tc-fpX. Tests were done on a a100! And they achieved 1.69x-2.65x inference performance! And I assume this can be transferred over to training (with the exception of possibly the KV cache, and embedding module). This is really exiting, this will breathe new life into the rapidly aging a100 due to the introduction of the h100’s fp8.

It was merged into deepspeed in this commit: 
https://github.com/microsoft/DeepSpeed/commit/ccfdb84e2a4a373ac657a99afd2d97e1d741b22b

Getting this pushed into the Pytorch as a dtype, that would be a major win. These are the benefits FP6 provides:
![IMG_4696](https://github.com/pytorch/pytorch/assets/122953474/53beb625-fe5f-4790-9045-d873497b6705)

### Alternatives

These kernels shouldn’t be limited by only the a100, they theoretically could work on any card with uint8_t and fp16 support. Provided these kernels were only written for a100 so without modification it might only work on ampere cards.

### Additional context

The tc-FPx kernel essentially takes 4fp16 values, quantizes them to fp6 with some place holders. Then they get pushed into an array built of 3x Uint8_t. As shown here:
![IMG_4686](https://github.com/pytorch/pytorch/assets/122953474/91b77585-b6f3-4e38-9b97-f91749436a23)
![IMG_4688](https://github.com/pytorch/pytorch/assets/122953474/adcda5d5-4b40-4b49-ac33-a64e7e1183cb)
![IMG_4689](https://github.com/pytorch/pytorch/assets/122953474/c891eced-bcb9-4db3-813c-b34bb59ee9a0)

cc @jerryzh168 @jianyuh @raghuramank100 @jamesr66a @vkuzo @jgong5 @Xia-Weiwen @leslie-fang-intel

vkuzo:
This is great and the inference e2e integration like a good candidate for addition to https://github.com/pytorch/ao .  Let us know if you are interested in contributing!

[...]

PR #942, opened 2024-09-25
SM75 (Turing) support for FP6 kernel

tobiasvanderwerff:
The [FP6 CUDA kernel](https://github.com/pytorch/ao/tree/main/torchao/csrc/cuda/fp6_llm) is currently limited to >=SM80, i.e. at least Ampere generation NVIDIA GPUs. However, as mentioned in [this issue](https://github.com/pytorch/ao/issues/208), the FP6 kernel could theoretically work for any card with uint8_t and fp16 support.

This PR adds FP6 support for SM75 (Turing generation) GPUs. Given the modest support for SM75 GPUs (e.g. the NVIDIA T4) in `torchao`, I'm hoping this can be a step towards improving that.

Most important changes:
- Replace asynchronous copy operations with vectorized loads ([diff](https://github.com/pytorch/ao/compare/main...tobiasvanderwerff:ao:fp6-sm75#diff-1ff6c927363707476562ee05743bdcdbcc90aa7e8d420b045f231310fd663806R42))
- Replace `m16n8k16` Tensor core operation with two `m16n8k8` operations ([diff](https://github.com/pytorch/ao/compare/main...tobiasvanderwerff:ao:fp6-sm75#diff-d095cc529407429f6f9c6736383cdc658d0b4e870fe9d27f64a36e2b79c72177R91))
- Account for a difference in expected parameters for the `ldmatrix` operation ([diff](https://github.com/pytorch/ao/compare/main...tobiasvanderwerff:ao:fp6-sm75#diff-d095cc529407429f6f9c6736383cdc658d0b4e870fe9d27f64a36e2b79c72177R58))

Before this, FP6 failed to run on SM75 (specifically, it would error out with "RuntimeError: operator torchao::quant_llm_linear does not exist").

Below are the benchmark and correctness test results (using `ao/benchmarks/benchmark_fp6.py`). Compared to the [initial FP6 results](https://github.com/pytorch/ao/pull/223) done by @gau-nernst, we still see a speedup of at least 2x compared to FP16 in most cases.

Testing specs:
- NVIDIA T4 GPU (SM75 generation)
- Torch version: 2.6.0.dev20240917

|   m |     k |     n |   fp6_latency (ms) |   fp16_latency (ms) |   speedup (d/s) |   correct |
|----:|------:|------:|-------------------:|--------------------:|----------------:|----------:|
|   1 |  8192 |  8192 |           223.05   |             517.205 |         2.31878 |         1 |
|   1 |  8192 | 10240 |           267.519  |             644.85  |         2.41048 |         1 |
|   1 |  8192 | 57344 |          1409.66   |            3618.91  |         2.56722 |         1 |
|   1 | 28672 |  8192 |           760.036  |            1800.83  |         2.3694  |         1 |
|   2 |  8192 |  8192 |           226.652  |             550.078 |         2.42697 |         1 |
|   2 |  8192 | 10240 |           271.363  |             687.359 |         2.53299 |         1 |
|   2 |  8192 | 57344 |          1442.43   |            3795.44  |         2.63128 |         1 |
|   2 | 28672 |  8192 |           768.993  |            1945.42  |         2.52982 |         1 |
|   4 |  8192 |  8192 |           229.543  |             567.315 |         2.4715  |         1 |
|   4 |  8192 | 10240 |           276.362  |             690.747 |         2.49943 |         1 |
|   4 |  8192 | 57344 |          1457.8    |            3805.93  |         2.61074 |         1 |
|   4 | 28672 |  8192 |           776.344  |            1965.42  |         2.53163 |         1 |
|   8 |  8192 |  8192 |           237.398  |             573     |         2.41367 |         1 |
|   8 |  8192 | 10240 |           284.349  |             810.79  |         2.85139 |         1 |
|   8 |  8192 | 57344 |          1521.18   |            3823.49  |         2.51349 |         1 |
|   8 | 28672 |  8192 |           783.763  |            1988.13  |         2.53665 |         1 |
|  16 |  8192 |  8192 |           273.57   |             582.142 |         2.12795 |         1 |
|  16 |  8192 | 10240 |           334.392  |             815.086 |         2.43751 |         1 |
|  16 |  8192 | 57344 |          1908.2    |            3857.97  |         2.02178 |         1 |
|  16 | 28672 |  8192 |           896.452  |            2002.43  |         2.23373 |         1 |
|  32 |  8192 |  8192 |           449.671  |             659.112 |         1.46576 |         1 |
|  32 |  8192 | 10240 |           548.003  |             821.81  |         1.49965 |         1 |
|  32 |  8192 | 57344 |          3255.59   |            4543.11  |         1.39548 |         1 |
|  32 | 28672 |  8192 |          1531.63   |            2403.52  |         1.56925 |         1 |
|  64 |  8192 |  8192 |            58.3608 |             672.911 |        11.5302  |         0 |
|  64 |  8192 | 10240 |            60.6918 |             887.669 |        14.6258  |         0 |
|  64 |  8192 | 57344 |           433.124  |            4812.82  |        11.1119  |         0 |
|  64 | 28672 |  8192 |            58.4863 |            2505.37  |        42.8369  |         0 |
| 128 |  8192 |  8192 |            58.2865 |            1096.99  |        18.8206  |         0 |
| 128 |  8192 | 10240 |           121.688  |            1202.46  |         9.88147 |         0 |
| 128 |  8192 | 57344 |           767.995  |            6741.57  |         8.77814 |         0 |
| 128 | 28672 |  8192 |            58.5768 |            3579.83  |        61.1134  |         0 |
| 256 |  8192 |  8192 |           194.799  |            1731.37  |         8.888   |         0 |
| 256 |  8192 | 10240 |           212.755  |            2273.48  |        10.6859  |         0 |
| 256 |  8192 | 57344 |          1186.61   |           11964.1   |        10.0825  |         0 |
| 256 | 28672 |  8192 |           195.011  |            5868.9   |        30.0953  |         0 |
| 512 |  8192 |  8192 |           388.129  |            3043.3   |         7.84093 |         0 |
| 512 |  8192 | 10240 |           199.79   |            3544.64  |        17.7418  |         0 |
| 512 |  8192 | 57344 |            51.1464 |           19975.7   |       390.559   |         0 |
| 512 | 28672 |  8192 |           388.241  |           10728.5   |        27.6335  |         0 |

As you can see, the kernel produces correct results for `m < 64` but incorrect results for `m >= 64`. So in that sense this kernel is currently not functional for `m >= 64`. Note also that the latency is drastically reduced for `m >= 64`, indicating that the kernel is exiting early for some reason. I did some digging and documented my findings below.

### Why does `m >= 64` produce incorrect results?

After looking into this, I think the problem is allocation of too much shared memory. The A100 has quite a lot more shared memory (164 KB per SM) than the T4, so some assumptions made about shared memory may not hold for the T4.

First, I ran the following script to query the device properties of my T4 GPU:

```cpp
#include <cuda_runtime.h>
#include <iostream>

int main() {
    cudaDeviceProp prop;
    cudaGetDeviceProperties(&prop, 0);
    std::cout << "Compute Capability: " << prop.major << "." << prop.minor << std::endl;
    std::cout << "Shared memory per block: " << prop.sharedMemPerBlock << " bytes" << std::endl;
    std::cout << "Shared memory per multiprocessor: " << prop.sharedMemPerMultiprocessor << " bytes" << std::endl;
    return 0;
}

```

Which outputs the following:

```
Compute Capability: 7.5
Shared memory per block: 49152 bytes
Shared memory per multiprocessor: 65536 bytes
```

So the available shared memory is indeed highly reduced compared to the A100. At the threshold of `m>=64`, the kernel gets called with a `TilingConfig` where `WARP_COL_MMA_TENSORS` is set to 8, instead of 4:

https://github.com/pytorch/ao/blob/2dea31543635ad747e52047f5091fe0b490c3807/torchao/csrc/cuda/fp6_llm/fp6_linear.cu#L101-L103

This leads to an increased size of `TILE_N`:

https://github.com/pytorch/ao/blob/2dea31543635ad747e52047f5091fe0b490c3807/torchao/csrc/cuda/fp6_llm/configs.h#L56

Which in turn leads to an increased shared memory allocation:

https://github.com/pytorch/ao/blob/2dea31543635ad747e52047f5091fe0b490c3807/torchao/csrc/cuda/fp6_llm/configs.h#L62-L63

Printing `TilingConfig::SMEM_SIZE_C_TILE` during runtime shows that when `m` goes from 32 to 64, `TilingConfig::SMEM_SIZE_C_TILE` goes from 38912 to 69362. Since 69362 bytes is more shared memory than the T4 has per block and per SM, it seems likely that this is causing problems. (But somehow, this does not throw any kind of error.) Setting `WARP_COL_MMA_TENSORS` to a maximum of 4, i.e. setting `TilingConfig<4, 1, 4>` here:

https://github.com/pytorch/ao/blob/2dea31543635ad747e52047f5091fe0b490c3807/torchao/csrc/cuda/fp6_llm/fp6_linear.cu#L88-L94

and here:

https://github.com/pytorch/ao/blob/2dea31543635ad747e52047f5091fe0b490c3807/torchao/csrc/cuda/fp6_llm/fp6_linear.cu#L102-L108

leads to correct results for all benchmarks:

|   m |     k |     n |   fp6_latency (ms) |   fp16_latency (ms) |   speedup (d/s) |   correct |
|----:|------:|------:|-------------------:|--------------------:|----------------:|----------:|
|   1 |  8192 |  8192 |            223.036 |             516.966 |        2.31786  |         1 |
|   1 |  8192 | 10240 |            269.426 |             644.646 |        2.39266  |         1 |
|   1 |  8192 | 57344 |           1410.92  |            3648.77  |        2.58609  |         1 |
|   1 | 28672 |  8192 |            759.226 |            1804.14  |        2.37629  |         1 |
|   2 |  8192 |  8192 |            225.865 |             549.892 |        2.4346   |         1 |
|   2 |  8192 | 10240 |            271.288 |             687.137 |        2.53287  |         1 |
|   2 |  8192 | 57344 |           1441.76  |            3795.4   |        2.63247  |         1 |
|   2 | 28672 |  8192 |            765.302 |            1945.28  |        2.54185  |         1 |
|   4 |  8192 |  8192 |            230.63  |             567.303 |        2.45979  |         1 |
|   4 |  8192 | 10240 |            276.171 |             690.602 |        2.50063  |         1 |
|   4 |  8192 | 57344 |           1456.33  |            3805.46  |        2.61305  |         1 |
|   4 | 28672 |  8192 |            774.708 |            1965.41  |        2.53696  |         1 |
|   8 |  8192 |  8192 |            237.608 |             573.233 |        2.41252  |         1 |
|   8 |  8192 | 10240 |            284.422 |             811.645 |        2.85367  |         1 |
|   8 |  8192 | 57344 |           1521.34  |            3823.1   |        2.51298  |         1 |
|   8 | 28672 |  8192 |            786.987 |            1986.32  |        2.52395  |         1 |
|  16 |  8192 |  8192 |            272.092 |             581.971 |        2.13888  |         1 |
|  16 |  8192 | 10240 |            334.622 |             816.67  |        2.44058  |         1 |
|  16 |  8192 | 57344 |           1909.48  |            3857.44  |        2.02015  |         1 |
|  16 | 28672 |  8192 |            890.995 |            1978.75  |        2.22083  |         1 |
|  32 |  8192 |  8192 |            450.208 |             659.402 |        1.46466  |         1 |
|  32 |  8192 | 10240 |            547.148 |             825.829 |        1.50933  |         1 |
|  32 |  8192 | 57344 |           3261.63  |            4540.91  |        1.39222  |         1 |
|  32 | 28672 |  8192 |           1532.26  |            2396.69  |        1.56416  |         1 |
|  64 |  8192 |  8192 |            739.317 |             671.094 |        0.907721 |         1 |
|  64 |  8192 | 10240 |            886.388 |             839.778 |        0.947416 |         1 |
|  64 |  8192 | 57344 |           5192.64  |            4846.74  |        0.933387 |         1 |
|  64 | 28672 |  8192 |           2360.6   |            2637.51  |        1.1173   |         1 |
| 128 |  8192 |  8192 |           1243.43  |            1049.64  |        0.844148 |         1 |
| 128 |  8192 | 10240 |           1598.47  |            1226.33  |        0.767191 |         1 |
| 128 |  8192 | 57344 |           9111.21  |            6696.32  |        0.734954 |         1 |
| 128 | 28672 |  8192 |           4154.84  |            3649.96  |        0.878483 |         1 |
| 256 |  8192 |  8192 |           2493.73  |            1603.46  |        0.642994 |         1 |
| 256 |  8192 | 10240 |           3069.87  |            2045.3   |        0.666251 |         1 |
| 256 |  8192 | 57344 |          17182.4   |           11757.5   |        0.68428  |         1 |
| 256 | 28672 |  8192 |           8275.68  |            5849.96  |        0.706886 |         1 |
| 512 |  8192 |  8192 |           5030.14  |            3091.38  |        0.614572 |         1 |
| 512 |  8192 | 10240 |           5893.7   |            3557.85  |        0.603671 |         1 |
| 512 |  8192 | 57344 |          31538.6   |           19964.7   |        0.633024 |         1 |
| 512 | 28672 |  8192 |          16674     |           11045.6   |        0.662444 |         1 |

But as you can see, performance becomes quite slow now for `m>=64` (slower than FP16). I might look into the details to see if there's any optimizations that can be made for the T4 in this particular case.

I tried creating a guard that checks for sm_75 and `m>=64`, but this is causing problems. Specifically, I tried to check for `__CUDA_ARCH__` in `fp6_linear.cu` (as done [here](https://github.com/pytorch/ao/blob/2dea31543635ad747e52047f5091fe0b490c3807/torchao/csrc/cuda/fp6_llm/fp6_linear.cu#L17)), but this does not seem to work for some reason  (best reason I can think of is that `__CUDA_ARCH__` is not defined because it is not device code, but that doesn't explain why the guard does work at the beginning of the file). However, if I instead try to put the directive inside the device code, it seems that functions like `assert` don't have any effect (I'm not really sure why). 

So currently, the `m>=64` case is not handled well (it will lead to incorrect results) and I'd be happy to hear suggestions about how to deal with this case.

## Final thoughts
- I noticed the `splitK` paramater for the FP6 kernel is set based on a [dictionary](https://github.com/pytorch/ao/blob/2dea31543635ad747e52047f5091fe0b490c3807/torchao/dtypes/affine_quantized_tensor.py#L1347) that is optimized for the A100 GPU. It might not be optimal for the T4, but I don't have any insights into how much it matters.

Any feedback is very welcome!

### DIFF
--- a/benchmarks/benchmark_fp6.py
+++ b/benchmarks/benchmark_fp6.py
@@ -7,7 +7,7 @@ from torchao.utils import benchmark_torch_function_in_microseconds
 from tqdm import tqdm
 
 
-def benchmark(m: int, k: int, n: int):
+def benchmark(m: int, n: int, k: int):
     float_data = torch.randn(n, k, dtype=torch.half, device="cuda")
     fp6_weight = to_affine_quantized_fpx(float_data, FloatxTensorCoreLayoutType(3, 2))
     fp16_weight = fp6_weight.dequantize(torch.half)
--- a/benchmarks/benchmark_fp6.py
+++ b/benchmarks/benchmark_fp6.py
@@ -7,7 +7,7 @@ from torchao.utils import benchmark_torch_function_in_microseconds
 from tqdm import tqdm
 
 
-def benchmark(m: int, n: int, k: int):
+def benchmark(m: int, k: int, n: int):
     float_data = torch.randn(n, k, dtype=torch.half, device="cuda")
     fp6_weight = to_affine_quantized_fpx(float_data, FloatxTensorCoreLayoutType(3, 2))
     fp16_weight = fp6_weight.dequantize(torch.half)
@@ -43,7 +43,7 @@ if __name__ == "__main__":
 
     for m in tqdm([1 << i for i in range(10)]):
         for n, k in zip(n_vals, k_vals):
-            results.append(benchmark(m, n, k))
+            results.append(benchmark(m, k, n))
 
     df = pd.DataFrame(results)
     df.to_csv("fp6_llm_benchmark_results.csv", index=False)
--- a/torchao/csrc/cuda/fp6_llm/fp6_linear.cu
+++ b/torchao/csrc/cuda/fp6_llm/fp6_linear.cu
@@ -14,7 +14,7 @@
 // 
 // This file is adapted from https://github.com/usyd-fsalab/fp6_llm/blob/5df6737cca32f604e957e3f63f03ccc2e4d1df0d/fp6_llm/csrc/fp6_linear.cu
 
-#if !defined(__CUDA_ARCH__) || __CUDA_ARCH__ >= 800  // at least Ampere
+#if !defined(__CUDA_ARCH__) || __CUDA_ARCH__ >= 750  // at least Turing
 
 #include "kernel_matmul.cuh"
 #include "kernel_reduction.cuh"
--- a/torchao/csrc/cuda/fp6_llm/fp6_linear.cu
+++ b/torchao/csrc/cuda/fp6_llm/fp6_linear.cu
@@ -22,6 +22,22 @@
 #include <stdio.h>
 #include <assert.h>
 
+inline bool isSM75GPU() {
+    int device;
+    cudaError_t err = cudaGetDevice(&device);
+    if (err != cudaSuccess) {
+        return false;
+    }
+
+    cudaDeviceProp props;
+    err = cudaGetDeviceProperties(&props, device);
+    if (err != cudaSuccess) {
+        return false;
+    }
+
+    return (props.major == 7) && (props.minor == 5);
+}
+
 template<typename TilingConfig, typename OutputDataType, int EXPONENT, int MANTISSA>
 static void Kernel_Ex(cudaStream_t    stream,
                       const uint4     *Weight,
@@ -80,38 +96,51 @@ cudaError_t fpx_linear_kernel(cudaStream_t    stream,
     if(N_Global>64 && N_Global<=128)    N_PowerOf2 = 128;
     if(N_Global>128)                    N_PowerOf2 = ((N_Global-1)/128+1) * 128;
 
-    if (Split_K == 1) {
-        switch (N_PowerOf2) {
-            case 8:     Kernel_Ex<TilingConfig<4, 1, 1>, half, EXPONENT, MANTISSA>(stream, Weight, Scales, B, C, M_Global, N_Global, K_Global, Split_K);  break;
-            case 16:    Kernel_Ex<TilingConfig<4, 1, 2>, half, EXPONENT, MANTISSA>(stream, Weight, Scales, B, C, M_Global, N_Global, K_Global, Split_K);  break;
-            case 32:    Kernel_Ex<TilingConfig<4, 1, 4>, half, EXPONENT, MANTISSA>(stream, Weight, Scales, B, C, M_Global, N_Global, K_Global, Split_K);  break;
-            case 64:    Kernel_Ex<TilingConfig<4, 1, 8>, half, EXPONENT, MANTISSA>(stream, Weight, Scales, B, C, M_Global, N_Global, K_Global, Split_K);  break;
-            case 128:   Kernel_Ex<TilingConfig<4, 1, 8>, half, EXPONENT, MANTISSA>(stream, Weight, Scales, B, C, M_Global, N_Global, K_Global, Split_K);  break;
-            default:    if (N_PowerOf2 % 128 != 0) {
-                            printf("FP6LLM_API Error: Unsupported N dimension %d!\n", N_PowerOf2);
-                            return cudaErrorUnknown;
-                        }
-                        Kernel_Ex<TilingConfig<4, 1, 8>, half, EXPONENT, MANTISSA>(stream, Weight, Scales, B, C, M_Global, N_Global, K_Global, Split_K);  break;
+    if (isSM75GPU() && (N_PowerOf2 == 64 || N_PowerOf2 == 128 || N_PowerOf2 % 128 == 0)) {
+        // For SM75 and N >= 64, we use a different TilingConfig to deal with smaller shared memory.
+        if (Split_K == 1) {
+            Kernel_Ex<TilingConfig<4, 1, 4>, half, EXPONENT, MANTISSA>(stream, Weight, Scales, B, C, M_Global, N_Global, K_Global, Split_K);
+        } else {
+            Kernel_Ex<TilingConfig<4, 1, 4>, float, EXPONENT, MANTISSA>(stream, Weight, Scales, B, Reduction_Workspace, M_Global, N_Global, K_Global, Split_K);
+        }
+    } else {
+        if (Split_K == 1) {
+            switch (N_PowerOf2) {
+                case 8:     Kernel_Ex<TilingConfig<4, 1, 1>, half, EXPONENT, MANTISSA>(stream, Weight, Scales, B, C, M_Global, N_Global, K_Global, Split_K);  break;
+                case 16:    Kernel_Ex<TilingConfig<4, 1, 2>, half, EXPONENT, MANTISSA>(stream, Weight, Scales, B, C, M_Global, N_Global, K_Global, Split_K);  break;
+                case 32:    Kernel_Ex<TilingConfig<4, 1, 4>, half, EXPONENT, MANTISSA>(stream, Weight, Scales, B, C, M_Global, N_Global, K_Global, Split_K);  break;
+                case 64:    Kernel_Ex<TilingConfig<4, 1, 8>, half, EXPONENT, MANTISSA>(stream, Weight, Scales, B, C, M_Global, N_Global, K_Global, Split_K);  break;
+                case 128:   Kernel_Ex<TilingConfig<4, 1, 8>, half, EXPONENT, MANTISSA>(stream, Weight, Scales, B, C, M_Global, N_Global, K_Global, Split_K);  break;
+                default:    if (N_PowerOf2 % 128 != 0) {
+                                printf("FP6LLM_API Error: Unsupported N dimension %d!\n", N_PowerOf2);
+                                return cudaErrorUnknown;
+                            }
+                            Kernel_Ex<TilingConfig<4, 1, 8>, half, EXPONENT, MANTISSA>(stream, Weight, Scales, B, C, M_Global, N_Global, K_Global, Split_K);  break;
+            }
+        }
+        else {
+            switch (N_PowerOf2) {
+                case 8:     Kernel_Ex<TilingConfig<4, 1, 1>, float, EXPONENT, MANTISSA>(stream, Weight, Scales, B, Reduction_Workspace, M_Global, N_Global, K_Global, Split_K);  break;
+                case 16:    Kernel_Ex<TilingConfig<4, 1, 2>, float, EXPONENT, MANTISSA>(stream, Weight, Scales, B, Reduction_Workspace, M_Global, N_Global, K_Global, Split_K);  break;
+                case 32:    Kernel_Ex<TilingConfig<4, 1, 4>, float, EXPONENT, MANTISSA>(stream, Weight, Scales, B, Reduction_Workspace, M_Global, N_Global, K_Global, Split_K);  break;
+                case 64:    Kernel_Ex<TilingConfig<4, 1, 8>, float, EXPONENT, MANTISSA>(stream, Weight, Scales, B, Reduction_Workspace, M_Global, N_Global, K_Global, Split_K);  break;
+                case 128:   Kernel_Ex<TilingConfig<4, 1, 8>, float, EXPONENT, MANTISSA>(stream, Weight, Scales, B, Reduction_Workspace, M_Global, N_Global, K_Global, Split_K);  break;
+                default:    if (N_PowerOf2 % 128 != 0) {
+                                printf("FP6LLM_API Error: Unsupported N dimension %d!\n", N_PowerOf2);
+                                return cudaErrorUnknown;
+                            }
+                            Kernel_Ex<TilingConfig<4, 1, 8>, float, EXPONENT, MANTISSA>(stream, Weight, Scales, B, Reduction_Workspace, M_Global, N_Global, K_Global, Split_K);  break;
+            }
         }
     }
-    else {
-        switch (N_PowerOf2) {
-            case 8:     Kernel_Ex<TilingConfig<4, 1, 1>, float, EXPONENT, MANTISSA>(stream, Weight, Scales, B, Reduction_Workspace, M_Global, N_Global, K_Global, Split_K);  break;
-            case 16:    Kernel_Ex<TilingConfig<4, 1, 2>, float, EXPONENT, MANTISSA>(stream, Weight, Scales, B, Reduction_Workspace, M_Global, N_Global, K_Global, Split_K);  break;
-            case 32:    Kernel_Ex<TilingConfig<4, 1, 4>, float, EXPONENT, MANTISSA>(stream, Weight, Scales, B, Reduction_Workspace, M_Global, N_Global, K_Global, Split_K);  break;
-            case 64:    Kernel_Ex<TilingConfig<4, 1, 8>, float, EXPONENT, MANTISSA>(stream, Weight, Scales, B, Reduction_Workspace, M_Global, N_Global, K_Global, Split_K);  break;
-            case 128:   Kernel_Ex<TilingConfig<4, 1, 8>, float, EXPONENT, MANTISSA>(stream, Weight, Scales, B, Reduction_Workspace, M_Global, N_Global, K_Global, Split_K);  break;
-            default:    if (N_PowerOf2 % 128 != 0) {
-                            printf("FP6LLM_API Error: Unsupported N dimension %d!\n", N_PowerOf2);
-                            return cudaErrorUnknown;
-                        }
-                        Kernel_Ex<TilingConfig<4, 1, 8>, float, EXPONENT, MANTISSA>(stream, Weight, Scales, B, Reduction_Workspace, M_Global, N_Global, K_Global, Split_K);  break;
-        }
+
+    if (Split_K != 1) {
         // Reduction for SplitK
         dim3 GridDim((M_Global * N_Global) / REDUCTION_ELEMENT_PER_THREADBLOCK, 1, 1);
         dim3 BlockDim(WARP_SIZE, 1, 1);
         SplitK_Reduction<<<GridDim, BlockDim, 0, stream>>>(C, Reduction_Workspace, M_Global, N_Global, Split_K);
     }
+
     return cudaGetLastError();
 }
 
--- a/torchao/csrc/cuda/fp6_llm/fp6_linear.cu
+++ b/torchao/csrc/cuda/fp6_llm/fp6_linear.cu
@@ -13,6 +13,10 @@
 //    limitations under the License.
 // 
 // This file is adapted from https://github.com/usyd-fsalab/fp6_llm/blob/5df6737cca32f604e957e3f63f03ccc2e4d1df0d/fp6_llm/csrc/fp6_linear.cu
+//
+// MODIFICATION NOTE (2024-09-25): added SM75 support (https://github.com/pytorch/ao/pull/942):
+// - Modified the TilingConfig parameters for SM75 to deal with smaller shared memory
+//
 
 #if !defined(__CUDA_ARCH__) || __CUDA_ARCH__ >= 750  // at least Turing
 
--- a/torchao/csrc/cuda/fp6_llm/kernel_matmul.cuh
+++ b/torchao/csrc/cuda/fp6_llm/kernel_matmul.cuh
@@ -140,7 +140,9 @@ __global__ void QUANT_GEMM_Kernel(const uint4* Weight, const half* Scales,
     for(int j=0; j<REG_PER_THREAD_C_TENSOR_16_16; j++)
       c[i][j] = 0.0f;
   //
+  #if __CUDA_ARCH__ >= 800
   cp_async_wait_all();
+  #endif
   __syncthreads();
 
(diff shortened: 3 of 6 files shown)


gau-nernst:
@tobiasvanderwerff Thank you for your contribution! The debug when m>=64 is very insightful.

I was the one added this FP6-LLM kernel to torchao. I don't have enough CUDA knowledge to fully understand the changes, but it seems reasonable to me. Have you had the chance to discuss this with the original author about this change? https://github.com/usyd-fsalab/fp6_llm

I will let @msaroufim decide if we should support / improve the kernel for sm75 at all. Concerns include CI testing + deviation from upstream, as well as not worth the efforts due to old age.

General comments:
- Low performance for large M is to be expected. I don't think you need to try to optimize for it
- Regarding guard for M>=64, if the macro doesn't work, I think you can check with either `torch.cuda.get_device_capability()` or `cudaDeviceProp`
- Regarding the splitK, yea I don't really like it too. I wish we have an autotune for CUDA kernel, similar to triton 😄
- (Can do later, once everything is finalized) Can you add a summary of your changes at the top of each modified file?

Side question. Would you be able to modify this kernel to support BF16? Since all modern models use BF16, it would be useful. I figure it would be a matter of changing the dequant logic + MMA instructions, but again, I'm not too confident with my CUDA skills 😅.

tobiasvanderwerff:
@gau-nernst let me address your comments/questions one by one.

> Have you had the chance to discuss this with the original author about this change?

I did not have contact with the original authors. Looking at their repo and paper, I don't have the impression that they had GPUs before A100 generation in mind when they implemented FP6. My main motivation for writing this port is that I thought it was an interesting challenge to try and make this work on a slightly older GPU for a GPU poor person like myself who finds an A100 too expensive to rent for extended periods of time :)

> I will let @msaroufim decide if we should support / improve the kernel for sm75 at all. Concerns include CI testing + deviation from upstream, as well as not worth the efforts due to old age.

That's understandable. A little while back I asked around in the [GPU-mode discord](https://discord.com/channels/1189498204333543425/1285157077454553120) (formerly CUDA-mode), where @HDCharles mentioned there is potentially a demand for supporting more architectures than what is currently supported.

> Regarding guard for M>=64, if the macro doesn't work, I think you can check with either torch.cuda.get_device_capability() or cudaDeviceProp

That's a good idea -- I will try it out.

> (Can do later, once everything is finalized) Can you add a summary of your changes at the top of each modified file?

Err, you mean a comment in the C++/CUDA files? As in, add a comment at the top of each changed file saying how I added SM75 support to it? If so, I could do definitely do that.

> Side question. Would you be able to modify this kernel to support BF16? Since all modern models use BF16, it would be useful. I figure it would be a matter of changing the dequant logic + MMA instructions, but again, I'm not too confident with my CUDA skills 😅.

I'd love to try that, but unfortunately the T4 does not support BF16 (it requires SM80). So that would require an A100 to work with, which is currently not feasible for me.

gau-nernst:
@tobiasvanderwerff Thank you for your response

I'm just checking if you ask the original author about this change since he probably can give more feedback in terms of accuracy and perf wise. No worries!

Since bad perf for large M is to be expected (actually FP6-LLM kernel is also slower than FP16 on A100 for M>=128), I think we don't need to check for M>=64? Just use `__CUDA_ARCH__` to select tile size accordingly.

> Err, you mean a comment in the C++/CUDA files? As in, add a comment at the top of each changed file saying how I added SM75 support to it? If so, I could do definitely do that.

Yes, exactly.

> I'd love to try that, but unfortunately the T4 does not support BF16 (it requires SM80). So that would require an A100 to work with, which is currently not feasible for me.

No worries at all! In case you don't know, for sm80 stuff, Lightning AI gives out free 22hrs/month of L4. Some GPU clouds also provide rental service for consumer cards, like 3090 and 4090, which is much cheaper than A100. Just something to consider.

Some other pointers
- I think the build wheel CI does not compile for sm75. So interested users have to compile from source. Can you add a line in `torchao/dtypes/floatx/README.md` saying that sm75 support requires installing torchao from source?
https://github.com/pytorch/ao/blob/64719d5882ac3790e0018bed77d2d7b1e788ac5c/packaging/env_var_script_linux.sh#L16-L19
- If possible, can you run the llama2 benchmark for sanity check? https://github.com/pytorch/ao/blob/main/torchao/dtypes/floatx/README.md#end-to-end-benchmarks

msaroufim:
I like this PR but am a bit conflicted because we won't be running this in CI so won't get signal on whether SM75 is working and if fp6 in upstream is still actively maintained it'll make it more challenging to fix things upstream

gau-nernst:
@msaroufim Regarding CI testing, I think we can treat this as "not officially supported", like Windows support. So interested users have to compile from source + no guarantee that future versions will continue working. Since the change is quite small, I think it's still somewhat reasonable.

benchmarks/benchmark_fp6.py
```diff
@@ -7,7 +7,7 @@
 from tqdm import tqdm
 
 
-def benchmark(m: int, k: int, n: int):
+def benchmark(m: int, n: int, k: int):
```
HDCharles, line 10: is this change intentional? usually we talk about matmuls as mkn i.e. m x k activation and k x n weight (odd to reverse them and i'm unsure if benchmarks previously were assuming the other ordering)

HDCharles (approved):
i like hte PR overall, the slowdown for large m is not unexpected for kernels focusing on weight only quantization, we expect for bsz small, your io speed dominates (yielding a speedup) and then when bsz is very large you'll asymptotically approach fp16 x fp16 speed but for a intermediate range the quantization overhead is going to be maximally painful leading to a significant slowdown.

benchmarks/benchmark_fp6.py
gau-nernst: Looking at this again, I indeed made a mistake. The benchmark results are still correct, it's just the list of shapes are different.

I took the benchmark shapes from here: https://github.com/usyd-fsalab/fp6_llm/blob/ce76774bcfc26b325c1b558abcf1935026d9abbc/tests/python/run.sh - It's a bit confusing since the author use different variable names...

The code generating the list of shapes (under `__name__ == "__main__"`) are correct (follow the author), and it calls `benchmark(m, n, k)`. If you think we should benchmark a different sets of shapes, it should be good too!

In summary, this change corrects my previous mistake.

tobiasvanderwerff: Yes, it's intentional. The function signature was `def benchmark(m: int, k: int, n: int):`, but arguments were passed as `(m, n, k)`, so I thought that that was unnecessarily confusing and wanted to change the ordering in either the function call or the function signature. In the function itself, the shapes become m x k for the activation and n x k for the weight. 

I see one benchmark example (`benchmark_gpu_sparsity`, see below) where the ordering is m, k, n, so let me change the function signature back to that ordering for consistency.
https://github.com/pytorch/ao/blob/4b5b5ee599f47eb9245c370caa559641a1f0fbb2/benchmarks/benchmark_gpu_sparsity.py#L25

tobiasvanderwerff: Missed your comment before I posted my own @gau-nernst. Thanks for clarifying! I actually noticed that `m` gets passed as `n` to the actual kernel, which is slightly confusing. If you don't mind, I'll change this for consistency. I don't think it should affect the results, expect that `m` will be switched by `n` in the performance table.

gau-nernst: @tobiasvanderwerff You mean `k` and `n` right? Your current change looks correct. Yea it doesn't affect the results, it will only show different shapes instead.

tobiasvanderwerff: What I meant is slightly different. I'm referring to the fact that the original authors do some odd switching of the shapes in `fp6_linear.cu`. The arguments that get passed are `_in_feats` (activations of shape m x k) and `_weights` (shape n x k), but then they unpack the shapes as `M = _weights.size(0)`, `K = _in_feats.shape(0)`, `N = _in_feats.shape(1)` (see below). 

https://github.com/pytorch/ao/blob/4b5b5ee599f47eb9245c370caa559641a1f0fbb2/torchao/csrc/cuda/fp6_llm/fp6_linear.cu#L150-L158

So even though we pass the arguments correctly to the benchmark function as m, k, n, the names get switched inside the kernel. Anyway, this is mainly confusing when debugging the kernel, but it might actually be fine to just leave it as is, since it does not affect the shapes/results that get passed back from the kernel.

tobiasvanderwerff:
The kernel now handles the `m>=64` edge case to produce correct results for all shapes used in the benchmark (see https://github.com/pytorch/ao/pull/942/commits/650ba03d7df2bbb5c78e0620f71911ecc6c60927). 

I couldn't use the __CUDA_ARCH__ constant, so I'm instead checking the CUDA arch at runtime. It unfortunately makes the kernel launch code a bit uglier, since the template parameters of `TilingConfig` need to be known at compile time. I'm not the most experienced with C++ so there may be a better solution, but this is the best I could come up with to handle this.

torchao/csrc/cuda/fp6_llm/fp6_linear.cu
```diff
@@ -13,15 +13,35 @@
 //    limitations under the License.
 // 
 // This file is adapted from https://github.com/usyd-fsalab/fp6_llm/blob/5df6737cca32f604e957e3f63f03ccc2e4d1df0d/fp6_llm/csrc/fp6_linear.cu
+//
+// MODIFICATION NOTE (2024-09-25): added SM75 support (https://github.com/pytorch/ao/pull/942):
+// - Modified the TilingConfig parameters for SM75 to deal with smaller shared memory
```
gau-nernst, line 18: Great and concise summary!

gau-nernst, line 18: Great and concise summary!

torchao/csrc/cuda/fp6_llm/fp6_linear.cu
```diff
@@ -80,38 +100,51 @@ cudaError_t fpx_linear_kernel(cudaStream_t    stream,
...
-            case 16:    Kernel_Ex<TilingConfig<4, 1, 2>, half, EXPONENT, MANTISSA>(stream, Weight, Scales, B, C, M_Global, N_Global, K_Global, Split_K);  break;
-            case 32:    Kernel_Ex<TilingConfig<4, 1, 4>, half, EXPONENT, MANTISSA>(stream, Weight, Scales, B, C, M_Global, N_Global, K_Global, Split_K);  break;
-            case 64:    Kernel_Ex<TilingConfig<4, 1, 8>, half, EXPONENT, MANTISSA>(stream, Weight, Scales, B, C, M_Global, N_Global, K_Global, Split_K);  break;
-            case 128:   Kernel_Ex<TilingConfig<4, 1, 8>, half, EXPONENT, MANTISSA>(stream, Weight, Scales, B, C, M_Global, N_Global, K_Global, Split_K);  break;
-            default:    if (N_PowerOf2 % 128 != 0) {
-                            printf("FP6LLM_API Error: Unsupported N dimension %d!\n", N_PowerOf2);
-                            return cudaErrorUnknown;
-                        }
-                        Kernel_Ex<TilingConfig<4, 1, 8>, half, EXPONENT, MANTISSA>(stream, Weight, Scales, B, C, M_Global, N_Global, K_Global, Split_K);  break;
+    if (isSM75GPU() && (N_PowerOf2 == 64 || N_PowerOf2 == 128 || N_PowerOf2 % 128 == 0)) {
```
gau-nernst, line 103: I think a slightly cleaner way to do this is to make `isSM75GPU()` check the outermost if, so we can keep the original code in a block. something like

```C++
    if (isSM75GPU()) {
        // your new code
    } else {
        // original code
    }
```

Also, I think you can "stack" switch cases together to make things slightly less verbose

```C++
    case 64:
    case 128:
        // code for both 64 and 128
        break;
```

gau-nernst (approved):
Thanks for addressing my feedback! Everything looks very nice!

I think `__CUDA_ARCH__` is not available on host code, so we can't use it outside CUDA kernel.

torchao/csrc/cuda/fp6_llm/fp6_linear.cu
tobiasvanderwerff: Perhaps I misunderstand, but the original code is still in the second code block? I.e. I didn't change it. Currently, it looks like this:
```cpp
    if (isSM75GPU() && (N_PowerOf2 == 64 || N_PowerOf2 == 128 || N_PowerOf2 % 128 == 0)) {
        // edge cases where SM75 needs different parameters for N_PowerOf2 >= 64
    } else {
        // original code
    }
```
The only thing I changed was move the SplitK reduction outside of the block above, putting it at the end of the function so I didn't have to duplicate it:
```cpp
    if (Split_K != 1) {
        // Reduction for SplitK
        dim3 GridDim((M_Global * N_Global) / REDUCTION_ELEMENT_PER_THREADBLOCK, 1, 1);
        dim3 BlockDim(WARP_SIZE, 1, 1);
        SplitK_Reduction<<<GridDim, BlockDim, 0, stream>>>(C, Reduction_Workspace, M_Global, N_Global, Split_K);
    }
```

gau-nernst: Hmm I see, that is fine then. The diff makes it hard to see the actual changes, so I misunderstood.

tobiasvanderwerff:
# Benchmark results on Llama-2-7b-chat-hf

Testing specs:
- T4 GPU
- Torch version: 2.4.1

## float16

```bash
python generate.py --compile --precision float16
```

Average tokens/sec: 19.04
Average Bandwidth: 251.60 GB/s
Peak Memory Usage: 13.90 GB
Model Size: 13.21 GB



## fp6

```bash
python generate.py --compile --precision float16 --quantization fp6
```

Average tokens/sec: 40.52
Average Bandwidth: 200.93 GB/s
Peak Memory Usage: 6.61 GB
Model Size: 4.96 GB

tobiasvanderwerff:
# Fp6 eval results

```bash
python eval.py --compile --precision float16 --quantization fp6
```

wikitext:
- word_perplexity: 12.3653
- byte_perplexity: 1.6005
- bits_per_byte: 0.6785

msaroufim:
@gau-nernst feel free to merge this whenever you feel it's ready

PR #946, opened 2024-09-25
Float8 dynamic autoquant

jainapurva:
Added support to autoquant for float8 dynamically quantized linear weight, i.e float8 weight and activation. 
Added fallback path in safe_int_mm to support float8 with torch.compile  

**Benchmark:**
On llama3.1b - bfloat16 model 

For Float8 Dynamic Quant
- Time for inference: 14.55 sec total
- Average tokens/sec: 13.67
- Average Bandwidth: 103.05 GB/s
- Peak Memory Usage: 13.77 GB
- Model Size: 7.54 GB

For Float8 weight only
- Time for inference: 12.02 sec total
- Average tokens/sec: 16.65
- Average Bandwidth: 125.47 GB/s
- Peak Memory Usage: 11.97 GB
- Model Size: 7.54 GB

### DIFF
--- a/torchao/kernel/intmm.py
+++ b/torchao/kernel/intmm.py
@@ -53,6 +53,7 @@ if TORCH_VERSION_AT_LEAST_2_2:
             and j_is_nonzero_multiple_of_8
             and k_is_nonzero_multiple_of_8
         )
+        
         if device_cpu or bad_dimensions_for_cublas:
             # fallback path
             return torch.matmul(input.cpu().to(torch.int32), mat2.cpu().to(torch.int32)).to(
--- a/torchao/kernel/intmm.py
+++ b/torchao/kernel/intmm.py
@@ -37,8 +37,12 @@ if TORCH_VERSION_AT_LEAST_2_2:
         """
         # torch.compile path
         if dynamo_is_compiling() or "FakeTensor" in input.__repr__():
-            return out_dtype(torch.ops.aten.mm.default, torch.int32, input, mat2)
-
+            try:
+                return out_dtype(torch.ops.aten.mm.default, torch.int32, input, mat2)
+            except Exception:
+                # fallback path, would run on H100 for float8 dtypes 
+                # Exception on H100 float8 dtype : "addmm_cuda" not implemented for 'Float8_e4m3fn' 
+                return torch.matmul(input.to(torch.float32), mat2.to(torch.float32)).to(torch.int32)
         # error checking for cublas path
         assert (
             mat2.device == input.device
@@ -53,7 +57,6 @@ if TORCH_VERSION_AT_LEAST_2_2:
             and j_is_nonzero_multiple_of_8
             and k_is_nonzero_multiple_of_8
         )
-
         if device_cpu or bad_dimensions_for_cublas:
             # fallback path
             return torch.matmul(input.cpu().to(torch.int32), mat2.cpu().to(torch.int32)).to(
--- a/torchao/kernel/intmm.py
+++ b/torchao/kernel/intmm.py
@@ -37,12 +37,8 @@ if TORCH_VERSION_AT_LEAST_2_2:
         """
         # torch.compile path
         if dynamo_is_compiling() or "FakeTensor" in input.__repr__():
-            try:
-                return out_dtype(torch.ops.aten.mm.default, torch.int32, input, mat2)
-            except Exception:
-                # fallback path, would run on H100 for float8 dtypes 
-                # Exception on H100 float8 dtype : "addmm_cuda" not implemented for 'Float8_e4m3fn' 
-                return torch.matmul(input.to(torch.float32), mat2.to(torch.float32)).to(torch.int32)
+            return out_dtype(torch.ops.aten.mm.default, torch.int32, input, mat2)
+
         # error checking for cublas path
         assert (
             mat2.device == input.device
--- a/torchao/quantization/autoquant.py
+++ b/torchao/quantization/autoquant.py
@@ -27,6 +27,7 @@ __all__ = [
     "autoquant",
     "DEFAULT_AUTOQUANT_CLASS_LIST",
     "DEFAULT_INT4_AUTOQUANT_CLASS_LIST",
+    "OTHER_AUTOQUANT_CLASS_LIST",
 ]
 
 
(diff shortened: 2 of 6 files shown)


torchao/quantization/autoquant.py
```diff
@@ -492,6 +491,92 @@ def from_float(cls, weight):
...
+        Returns:
+            float: The benchmarked time for the autoquantization process.
+        """
+        if not _is_interpolate_mode(mode):
+            return super()._autoquant_test(act_mat, weight, bias, best_time, mode)
+
+        # SAM best is between .8 and 1, SDXL also performs best in this range
+        INTERPOLATION_CONSTANT = mode[1]
+        w_qtensor = cls.from_float(weight)
+        x_vals_float8, x_scales = quantize_activation_per_token_absmax(
```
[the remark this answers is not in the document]
jainapurva: Now updated it to return float8

torchao/quantization/autoquant.py
```diff
@@ -492,6 +491,92 @@ def from_float(cls, weight):
...
+            act_mat.reshape(-1, act_mat.shape[-1]), dtype=torch.float8_e4m3fn
+        )
+        quantized_matmul = (
+            lambda x_vals_float8, x_scales, w_vals_float8:
+                safe_int_mm(x_vals_float8, w_vals_float8) * x_scales
```
drisspg, line 560: the AQT path for float8 should call into; https://github.com/pytorch/ao/blob/72d2518398df3b49fd627c2368c2d5acce4b2540/torchao/float8/inference.py#L60
And not safe_mm

jainapurva: Done

torchao/kernel/intmm.py
```diff
@@ -37,8 +37,12 @@ def safe_int_mm(input: torch.Tensor, mat2: torch.Tensor) -> torch.Tensor:
         """
         # torch.compile path
         if dynamo_is_compiling() or "FakeTensor" in input.__repr__():
-            return out_dtype(torch.ops.aten.mm.default, torch.int32, input, mat2)
-
+            try:
```
drisspg, line 40: I dont understand this change, is it related to supporting fp8 with autoquant?

HDCharles: autoquant with dynamic int8 quant breaks on H100, this is addressing that

torchao/quantization/autoquant.py
```diff
@@ -492,6 +492,91 @@ def from_float(cls, weight):
...
+                    block_size=block_size,
+                    target_dtype=target_dtype,
+                    layout_type=layout_type,
+                    scale_dtype=torch.float32,
+        )
+        weight = super(AQFloat8DynamicallyQuantizedLinearWeight, cls).from_float(weight, input_quant_func)
+        return weight
+
+    @classmethod
+    def _autoquant_test(cls, act_mat, weight, bias, best_time, mode=["relu", None]):
```
HDCharles, line 537: did you test this in practice? we may need different constants for int8 and float8 dynamic, how does it perform in benchmarks and stuff. If you haven't really tested this on 2-3 models it may be better to just remove it and use the default method which will be very conservative under the interpolation mode and will still work reasonably under the relu mode.

torchao/quantization/utils.py
```diff
@@ -139,13 +139,12 @@ def _get_per_token_block_size(x: torch.Tensor) -> List[int]:
 # taken from
 # https://github.com/mit-han-lab/smoothquant/blob/2f87951dacfb9238d8d657f52ae83a82a3c9ba0c/smoothquant/fake_quant.py#L26
 # and slightly modified
-def quantize_activation_per_token_absmax(t):
+def quantize_activation_per_token_absmax(t, dtype=torch.int8):
```
HDCharles, line 142: does this actually work in practice, we're still using the same quant min/max, this seems inadvisable.

HDCharles (changes requested):
see comments, i would probably skip the quant test bit unless it gets tested on 2-3 models and the way the activation quantization is implemented seems like its going to cause issues because its only superficially been altered away from its normal int8 quantization.

torchao/quantization/autoquant.py
jainapurva: I've tested it on Llama, the numbers aren't great, but I can push it to next PR, with more benchmarks

torchao/kernel/intmm.py
drisspg: ohh okay so not directly related to this PR

torchao/quantization/autoquant.py
```diff
@@ -492,6 +492,47 @@ def from_float(cls, weight):
...
+        # input settings
+        def get_per_token_block_size(x):
+            block_size = list(x.shape)
+            for i in range(len(block_size)-1):
+                block_size[i] = 1
+            return block_size
+
+        input_target_dtype = torch.float8_e4m3fn
+        layout_type = Float8LayoutType()
+        input_quant_func = lambda x: to_affine_quantized_floatx(
```
drisspg, line 518: can we re-use this function: https://github.com/pytorch/ao/blob/60ffb86f28f69c4d485c88cb93ce2b5b6f9f39ea/torchao/quantization/quant_api.py#L685

torchao/quantization/autoquant.py
```diff
@@ -492,6 +492,47 @@ def from_float(cls, weight):
...
+    """
+    AutoQuantizable version of Float8DynamicallyQuantizedLinearWeight
+    """
+    @classmethod
+    def from_float(cls, weight):
+
+        # avoid circular dep
+        from torchao.dtypes import to_affine_quantized_floatx
+        # weight settings
+        def get_weight_block_size(x):
```
drisspg, line 505: @HDCharles we are baking in the scaling strategy, How does autoquant choose which strategy

HDCharles: each are separate classes, see how we handle groupsize for int4 quant

torchao/quantization/autoquant.py
jainapurva: Done

torchao/quantization/autoquant.py
jainapurva: I can look try implementing that in separate PR

torchao/quantization/autoquant.py
```diff
@@ -492,6 +494,46 @@ def from_float(cls, weight):
         block_size = (1, weight.shape[1])
         return super(AQFloat8WeightOnlyQuantizedLinearWeight, cls).from_hp_to_floatx(weight, block_size, target_dtype=cls.target_dtype, layout_type=Float8LayoutType())
 
+class AQFloat8DynamicallyQuantizedLinearWeight(AQMixin, LinearActivationQuantizedTensor):
```
drisspg, line 497: I think this looks good, lets name this PerRow scaling and lets only import PerRow above

drisspg (approved):

Issue #929, opened 2024-09-24
torchao.ops import error on aarch64 linux


ng-05:
Hello team,

I am compiling pytorch/ao package from source on aarch64 linux. 
The build and installtion seems to be working perfectly fine unitll I do and import : `import torchao.ops`

The issue is not with not any single op. Its with all the op registration in `ops.py`

To reproduce:
```

- git clone https://github.com/pytorch/ao.git --recursive
- cd ao
- python setup.py install 
- python -c 'import torchao.ops'

```

This import errors out with below error:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/sudouser/workspace/torch_ao_tests/ao/torchao/ops.py", line 42, in <module>
    def _(EXPONENT, MANTISSA, _in_feats, _weights, _scales, splitK = 1):
  File "/home/sudouser/workspace/torch_ao_tests/ao/torchao/ops.py", line 10, in decorator
    return torch.library.register_fake(f"{name}")(func)
  File "/home/sudouser/workspace/new_tools/tools/dev/pytorch/venv_latest/lib/python3.10/site-packages/torch/library.py", line 795, in register
    use_lib._register_fake(op_name, func, _stacklevel=stacklevel + 1)
  File "/home/sudouser/workspace/new_tools/tools/dev/pytorch/venv_latest/lib/python3.10/site-packages/torch/library.py", line 184, in _register_fake
    handle = entry.fake_impl.register(func_to_register, source)
  File "/home/sudouser/workspace/new_tools/tools/dev/pytorch/venv_latest/lib/python3.10/site-packages/torch/_library/fake_impl.py", line 31, in register
    if torch._C._dispatch_has_kernel_for_dispatch_key(self.qualname, "Meta"):
RuntimeError: operator torchao::quant_llm_linear does not exist
```

digantdesai:
cc @jerryzh168 - have you seen this before? Or do we have a ci for Aarch64 Linux?

msaroufim:
Generally the answer to do we support hardware architecture X is most likely gonna be no :P 

because AO isn't a pure python package we have to manually support CI for more HW architectures for now we support ARM Mac and Linux and that's about it

We've been very interested in supporting more HW architectures so hopefully we'll share some updates soon

jerryzh168:
> cc @jerryzh168 - have you seen this before? Or do we have a ci for Aarch64 Linux?

[...]

PR #949, opened 2024-09-26
Move op definition for custom kernel from C++ to Python

gau-nernst:
## Problem

Currently op definitions for our custom kernels are defined in C++. If, for some reasons, building torchao from source does not compile these C++ files, there will be undefined ops error when importing `torchao.ops`. #929

## Solution

Move these op definitions to Python to ensure that the ops are always defined, regardless of whether the C++ files are built.


msaroufim (approved):

elfisworking:
@gau-nernst i compile torchao 0.0.5 version by using this pr in HW device.  And i run the code following https://pytorch.org/torchtune/main/tutorials/qat_finetune.html. But the result is 
image
. can you help me ? thanks very much

gau-nernst:
@elfisworking It looks unrelated to this PR. What is the device you are running this on? Looks like NPU, which torchao probably does not support officially. And your screenshot only shows a warning, not an error, so things should still be fine? Anyway, you should open a separate issue if something is not working as intended.

elfisworking:
> @elfisworking It looks unrelated to this PR. What is the device you are running this on? Looks like NPU, which torchao probably does not support officially. And your screenshot only shows a warning, not an error, so things should still be fine? Anyway, you should open a separate issue if something is not working as intended.

thanks for  your reply. yes, i run torchao in npu that is a huawei aarch device. Following https://pytorch.org/torchtune/main/tutorials/qat_finetune.html,  when i run this code
>>> print(prepared_model.layers[0].attn)
MultiHeadAttention(
  (q_proj): Int8DynActInt4WeightQATLinear(in_features=4096, out_features=4096, bias=False)
  (k_proj): Int8DynActInt4WeightQATLinear(in_features=4096, out_features=1024, bias=False)
  (v_proj): Int8DynActInt4WeightQATLinear(in_features=4096, out_features=1024, bias=False)
  (output_proj): Int8DynActInt4WeightQATLinear(in_features=4096, out_features=4096, bias=False)
  (pos_embeddings): RotaryPositionalEmbeddings()
).

gau-nernst:
@elfisworking Yes, you should raise a separate issue.

PR #964, opened 2024-09-27
Run CI for gh branches

msaroufim:
@vkuzo lemme know if this works


vkuzo:
looks reasonable! seems tricky to test until it lands, but let's land it and see?

vkuzo (approved):

vkuzo:
thank you!

PR #969, opened 2024-09-28
add ruff lint rule to remove unused imports via ruff

jimexist:
- include stricter linter rules for the allowlisted files
- add more comments
- use python 11 for ruff run
- fix ruff to 0.6.8 to make it consistent

### DIFF
--- a/.github/workflows/ruff_linter.yml
+++ b/.github/workflows/ruff_linter.yml
@@ -25,13 +25,18 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install ruff
+        pip install ruff==0.6.8
     - name: Analyzing the code with ruff
       run: |
         ruff check .
-    - name: Check all Python files for syntax errors (E999) and undefined vars (F821)
+    - name: Check *all* Python files for F821, F823, and W191
       run: |
-        ruff check --isolated --select E999,F821
-    - name: Check well formatted code
+        # --isolated is used to skip the allowlist at all so this applies to all files
+        # please be careful when using this large changes means everyone needs to rebase
+        ruff check --isolated --select F821,F823,W191
+    - name: Check the allow-listed files for F,I
+      run: |
+        ruff check --select F,I
+    - name: Check the allow-listed files for well formatted code
       run: |
         ruff format --check
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -19,5 +19,5 @@ tabulate  # QOL for printing tables to stdout
 ninja
 
 # Linting
-ruff
+ruff==0.6.8
 pre-commit
--- a/test/dtypes/test_affine_quantized_float.py
+++ b/test/dtypes/test_affine_quantized_float.py
@@ -1,35 +1,36 @@
+import pytest
+
 from torchao.utils import (
     TORCH_VERSION_AT_LEAST_2_5,
 )
-import pytest
 
 if not TORCH_VERSION_AT_LEAST_2_5:
     pytest.skip("Unsupported PyTorch version", allow_module_level=True)
 
+import copy
+import io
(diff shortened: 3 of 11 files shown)


msaroufim:
So for linting feel free to add a new file or folder here and make all the changes you need https://github.com/pytorch/ao/blob/main/ruff.toml

It's tricky for us to merge something like this since eveyrone would need to rebase their code

jimexist:
> So for linting feel free to add a new file or folder here and make all the changes you need https://github.com/pytorch/ao/blob/main/ruff.toml
> 
> It's tricky for us to merge something like this since eveyrone would need to rebase their code

thank you for the reference. that makes sense.
i then only updated the GitHub workflow to include the new rule. this renders no difference on the list of included files in the toml config.

msaroufim:
Ok cool! Can you remove unused imports in files erroring in CI and we can merge this

jimexist:
> Ok cool! Can you remove unused imports in files erroring in CI and we can merge this

turns out the ruff linter is configured to:
1. format w.r.t the listed files
1. apply the limited ruleset for _all_ files

my change is to include a stricter linter rule but only applies to the allowlist, kind of between 1 and 2

.github/workflows/ruff_linter.yml
```diff
@@ -25,13 +25,18 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install ruff
+        pip install ruff==0.6.8
     - name: Analyzing the code with ruff
       run: |
         ruff check .
-    - name: Check all Python files for syntax errors (E999) and undefined vars (F821)
+    - name: Check *all* Python files for F821, F823, and W191
```
jimexist, line 32: e999 was deprecated and by default applied

.github/workflows/ruff_linter.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.9"]
+        python-version: ["3.11"]
```
msaroufim, line 18: so generally we want to make sure ao works on python versions 3.9+ so what's teh goal of this change

.github/workflows/ruff_linter.yml
```diff
@@ -25,13 +25,18 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install ruff
+        pip install ruff==0.6.8
```
msaroufim, line 28: how come?

.github/workflows/ruff_linter.yml
jimexist: 3.11 is newer but your context makes sense. I'll revert.

.github/workflows/ruff_linter.yml
jimexist: Rule set changes with versions so this is to avoid sudden breakage if they release a newer version

.github/workflows/ruff_linter.yml
jimexist, line 18:
```suggestion
        python-version: ["3.9"]
```

jimexist:
```suggestion
        python-version: ["3.9"]
```

.github/workflows/ruff_linter.yml
msaroufim: @drisspg any preference on this?

Cause btw now locally vs remotely there'll be ruff deviations https://github.com/pytorch/ao/blob/main/dev-requirements.txt#L22

drisspg: That's fine; I think we just need to also pin in dev-requirments.txt

msaroufim (approved):

PR #976, opened 2024-09-30
cleaned up and tested tp support

debajyotidatta:


torchao/_models/llama/generate.py
```diff
@@ -30,7 +31,7 @@ def device_sync(device):
 wd = Path(__file__).parent.parent.resolve()
 sys.path.append(str(wd))
 
-from torchao._models.llama.model import Transformer, prepare_inputs_for_model
+from model import Transformer, prepare_inputs_for_model
```
HDCharles, line 34: i think we want this to be the same as before, otherwise it can accidentally pull in gpt-fast model and it becomes annoying to test/debug

HDCharles (approved):
otherwise looks good

torchao/_models/llama/tp.py
```diff
@@ -0,0 +1,162 @@
...
+import torch.distributed as dist
+from torch import nn
+if os.uname().sysname != "Darwin":
+    from torch.distributed import _functional_collectives as funcol
+else:
+    # Distributed is not supported on MacOS
+    funcol = None
+
+from torchao._models.llama.model import Attention, FeedForward, Transformer
+from torchao.quantization.GPTQ import WeightOnlyInt4Linear
```
jerryzh168, line 21: we are not using`WeightOnlyInt4Linear` any more in torchao I think, is this just for GPTQ?

jerryzh168:
@debajyotidatta the way we are supporting tp by composing quantized tensor subclass and DTensor: https://github.com/pytorch/ao/pull/939, I believe the way its done in gpt-fast is outdated, would you be interested in integrating the new tp flow in generate.py instead?

debajyotidatta:
> @debajyotidatta the way we are supporting tp by composing quantized tensor subclass and DTensor: #939, I believe the way its done in gpt-fast is outdated, would you be interested in integrating the new tp flow in generate.py instead?

Sounds good! I will do that.

HDCharles (changes requested):
can you add try to do tensor parallel support along the lines of https://github.com/pytorch/ao/blob/main/tutorials/developer_api_guide/tensor_parallel.py

PR #977, opened 2024-09-30
[autoquant] Fix the autoquant multi_head_attention torch_function dispatch

IvanKobzarev:
* __->__ #977

Hotfix for the autoquant of vision transformer model.
What happens in the model:
```
multi_head_attention_forward(.. ao_subclass .. ) {
  torch.nn.functional.linear(.. ao_subclass ..)
}
```

ao_subclass 
```
  def __torch_function__(func, args) default path is:
        with torch._C.DisableTorchFunctionSubclass():
            return func(*args, **kwargs)
```       

         
  This results that mha goes into torch_function, but its implementation is run under DisableTorchFunctionSubclass which blocks downstream override of  torch.nn.functional.linear.

As a result it fails in eager mode.

The fix:
The hotfix is to handle multi_head_attention with unwrapping the quantized weight to the Tensor.
It breaks the logging from linear, so I hardcoded `error_on_unseen` to make False, to not fail by default.

The test:
```
import torch
import torchao
from torchao.utils import benchmark_model, profiler_runner
from torchvision import models
torch.set_float32_matmul_precision("high")
# Load Vision Transformer model
model = models.vit_b_16(weights=models.ViT_B_16_Weights.IMAGENET1K_V1)
# Set the model to evaluation mode
model.eval().cuda().to(torch.bfloat16)
# Input tensor (batch_size, channels, height, width)
inputs = (torch.randn(1, 3, 224, 224, dtype=torch.bfloat16, device='cuda'),)
## Quantization code - start
# int8 dynamic quantization act, int8 weight, see ao/torchao/quantization/README.md
# for APIs for earlier torch version and other quantization techniques
# temporary workaround to recover the perf with quantized model under torch.compile
torch.backends.mha.set_fastpath_enabled(False)
# for torch 2.4+

from torchao.quantization.quant_api import int8_dynamic_activation_int8_weight

torch.compile(model, backend="inductor", fullgraph=True, mode="max-autotune")(inputs[0])

model = torchao.autoquant(torch.compile(model, backend="eager", fullgraph=True))

## Quantization code - end
## compilation configs
torch._dynamo.config.automatic_dynamic_shapes = False
torch._inductor.config.force_fuse_int_mm_with_mul = True
torch._inductor.config.use_mixed_mm = True
## compilation configs end
# temporary workaround for the API to work with torch.compile
from torchao.utils import TORCH_VERSION_AT_LEAST_2_5
from torchao.utils import unwrap_tensor_subclass
# Must run with no_grad when optimizing for inference
with torch.no_grad():
    # warmup
    benchmark_model(model, 20, inputs)
    # benchmark
    print("elapsed_time: ", benchmark_model(model, 1000, inputs), " milliseconds")
    # Create a trace
    profiler_runner("quant.json.gz", benchmark_model, model, 5, inputs)

```

### DIFF
--- a/torchao/quantization/autoquant.py
+++ b/torchao/quantization/autoquant.py
@@ -19,6 +19,7 @@ from torchao.utils import TORCH_VERSION_AT_LEAST_2_3, TORCH_VERSION_AT_LEAST_2_5
 from torchao.quantization.utils import quantize_activation_per_token_absmax
 
 import torch.nn.functional as F
+import torch.utils._pytree as pytree
 
 __all__ = [
     "AutoQuantizableLinearWeight",
@@ -194,15 +195,16 @@ class AutoQuantizableLinearWeight(torch.Tensor):
             )
             cls.log_shape(mat1, w_autoquant, bias)
             return func(mat1, w_autoquant.weight, bias)
-        try:
-            with torch._C.DisableTorchFunctionSubclass():
-                return func(*args, **kwargs)
-        except:
-            print(f"ERR: subclass doesn't implement {func}")
+        elif func is torch.nn.functional.multi_head_attention_forward:
+            new_args = pytree.tree_map_only(cls, lambda x: x.weight, args)
+            return func(*new_args, **kwargs)
+
+        with torch._C.DisableTorchFunctionSubclass():
+            return func(*args, **kwargs)
 
     @classmethod
     def __torch_dispatch__(cls, func, types, args, kwargs):
-         if func is aten.detach.default:
+        if func is aten.detach.default:
             return return_and_correct_aliasing(func, args, kwargs, args[0]._apply_fn_to_data(torch.detach))
 
 @torch.no_grad()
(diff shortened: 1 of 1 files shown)


torchao/quantization/autoquant.py
```diff
@@ -194,15 +195,16 @@ def __torch_function__(cls, func, types, args=(), kwargs=None):
             )
             cls.log_shape(mat1, w_autoquant, bias)
             return func(mat1, w_autoquant.weight, bias)
-        try:
-            with torch._C.DisableTorchFunctionSubclass():
-                return func(*args, **kwargs)
-        except:
-            print(f"ERR: subclass doesn't implement {func}")
```
jerryzh168, line 201: problem is this just returns None I think, although it might be needed to implement `not error_on_unseen`, we need to think about how to do that if we change this

torchao/quantization/autoquant.py
```diff
@@ -553,7 +555,7 @@ def _change_autoquantizable_to_quantized(model, supress_autoquant_errors=True, *
         lambda mod, *args:
             hasattr(mod, "weight") and isinstance(mod.weight, AutoQuantizableLinearWeight)
     )
-    error_on_unseen=kwargs.pop("error_on_unseen", True)
+    error_on_unseen=kwargs.pop("error_on_unseen", False)
```
HDCharles, line 558: this cannot just be hard coded to false

HDCharles (changes requested):

HDCharles:
you need to create a test case for this issue

Issue #987, opened 2024-10-01
[RFC] Long Term QAT Flow


andrewor14:
Currently torchao QAT has two APIs, [tensor subclasses](https://github.com/pytorch/ao/blob/a4221df5e10ff8c33854f964fe6b4e00abfbe542/torchao/quantization/prototype/qat/api.py#L41) and [module swap](https://github.com/pytorch/ao/blob/a4221df5e10ff8c33854f964fe6b4e00abfbe542/torchao/quantization/prototype/qat/_module_swap_api.py#L39). The original plan was to deprecate and eventually remove the old module swap API in favor of the tensor subclass API. However, users are starting to rely on the module API for production uses due to gaps in the tensor subclass API. In this RFC, we discuss the few long term plans for these two APIs in torchao.

## API Today
We use a quantizer API today to abstract the implementation details from the user. Currently we support both tensor subclass and module swap APIs using different quantizers:

```
from torchao.quantization.prototype.qat import Int8DynActInt4WeightQATQuantizer 
from torchao.quantization.prototype.qat._module_swap_api import Int8DynActInt4WeightQATQuantizerModuleSwap

# tensor subclass version
qat_quantizer = Int8DynActInt4WeightQATQuantizer()
# module swap version
qat_quantizer = Int8DynActInt4WeightQATQuantizerModuleSwap()

# prepare: inserts fake quantizes but keeps everything in bf16
fake_quantized_model = qat_quantizer.prepare(model)

# train or fine-tune as before
train(fake_quantized_model)

# convert: actually quantizes the model to lower bit-widths
quantized_model = qat_quantizer.convert(fake_quantized_model)
```

## Module Swap vs Tensor Subclass

Although tensor subclasses are generally adopted in torchao, the main gap today are (1) the lack of general distributed support, and (2) steep learning curve. For these two reasons, some users prefer the module swap flow, and have begun implementing new features in this flow, such as embedding quantization and static QAT.

To summarize the pros and cons of both approaches:

  | Tensor subclass | Module swap
-- | -- | --
Consistency | ✓ The rest of torchao uses tensor subclasses, including PTQ, quantized training, float8 inference, and sparsity. These all use the same quantize_ API. | ✖ Diverges from PTQ flow (not necessarily a con)
Composability | ✓ Better composability with other tensor subclasses such as DTensor and NestedJaggedTensor | ✖ Pure module swap misses out on potential composability benefits with other tensor subclasses (no clear benefits for QAT today)
[...]

PR #989, opened 2024-10-01
Make module swap the main QAT flow again

andrewor14:
**Summary:** Following https://github.com/pytorch/ao/issues/987, this commit makes module swap the main QAT flow today. We remove all tensor subclass fake quantize injection logic since this is not needed in both the long term and the short term plans for QAT. In the short term, we will continue to use a full module swap flow, and only migrate to the long term flow once there is general distributed support for tensor subclasses and when tensor subclass composability provides meaningful benefits.

**Test Plan:**


andrewor14:
Closing this in favor of https://github.com/pytorch/ao/pull/1019 (exact same PR)

PR #990, opened 2024-10-01
Add example for awq like flow

jerryzh168:
Summary:
Added an example and util for awq like flow that applies extra equalization scale tensor to input activation

Test Plan:
python tutorials/calibration_flow/awq_like.py

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/quantization/linear_activation_scale.py
+++ b/torchao/quantization/linear_activation_scale.py
@@ -1,0 +1,160 @@
+from typing import Any, Callable, Dict, Optional
+
+import torch
+from torch.utils._python_dispatch import return_and_correct_aliasing
+
+from torchao.utils import (
+    TORCH_VERSION_AT_LEAST_2_5,
+    TorchAOBaseTensor,
+)
+
+__all__ = [
+    "WeightTensorWithLinearActivationScaleMetadata",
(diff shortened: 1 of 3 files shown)


torchao/quantization/linear_activation_scale.py
```diff
@@ -0,0 +1,165 @@
...
+        return cls(
+            tensor_data_dict["original_weight_tensor"],
+            tensor_data_dict["scale"],
+        )
+
+    @staticmethod
+    def _quantized_linear_op(
+        input_tensor: torch.Tensor, weight_tensor: torch.Tensor, bias: torch.Tensor
+    ):
+        original_weight_tensor = weight_tensor.original_weight_tensor
```
vayuda, line 77: In the example a quantized tensor is passed in so instead of being called `original_weight_tensor` shouldn't it be called `quantized_weight_tensor`

jerryzh168: the tensor subclass works with either quantized or not quantized tensor actually

`original_weight_tensor` means the original weight tensor that is not wrapped with this tensor subclass (in our case it's a quantized tensor)

vayuda (approved):

PR #995, opened 2024-10-02
Subclass API (#966)

metascroy:
Summary:

Adds new int8_dynamic_activation_intx_weight quantization with subclass API

Differential Revision: D62464487

### DIFF
--- a/torchao/experimental/docs/readme.md
+++ b/torchao/experimental/docs/readme.md
@@ -45,6 +45,31 @@ quantized_model = linear_quantizer.quantize(quantized_model)
 
 If you get stuck on the above steps, working examples for both linear and embedding are in torchao/experimental/tests/test_linear_8bit_act_xbit_weight_quantizer.py and torchao/experimental/tests/test_embedding_xbit_quantizer.py.  For example, running `python tests/test_linear_8bit_act_xbit_weight_quantizer.py` loads the ops, creates a toy model, quantizes the model, and runs it in eager, compile, AOTI, and exports the model.
 
+### Subclass API
+
+For linear, you can also use the new subclass API in torchao.
+
+```python
+import torch
+torch.ops.load_library("cmake-out/lib/libtorchao_ops_aten.dylib") # make sure this path is correct on your machine
+
+my_model = Model()
+
+from torchao.experimental.quant_api import int8_dynamic_activation_intx_weight
+from torchao.quantization.quant_api import quantize_
+quantize_(
+    my_model,
+    int8_dynamic_activation_intx_weight(
+        group_size=256,
+        nbit=4,
+        has_weight_zeros=False,
+    ),
+)
+```
+
+If you get stuck, consult
+`tests/test_linear_int8_dynamic_activation_intx_weight_subclass.py`.
+
 ## Available in torchchat
 
 TorchAO experimental kernels are [available in torchchat](https://github.com/pytorch/torchchat/blob/main/docs/quantization.md#experimental-torchao-lowbit-kernels), PyTorch's solution for running LLMs locally.  Torchchat integration uses similar steps to above.
--- a/torchao/experimental/quant_api.py
+++ b/torchao/experimental/quant_api.py
@@ -463,3 +463,56 @@ class IntxWeightEmbeddingQuantizer:
             },
         )
         return model
+
+
+from torchao.experimental._linear_8bit_act_xbit_weight_layout import Linear8BitActXBitWeightLayout
+from torchao.quantization.quant_api import (
+    _get_linear_subclass_inserter,
+    MappingType,
+    to_affine_quantized_intx,
+    ZeroPointDomain,
+)
+
+
+def int8_dynamic_activation_intx_weight(
+    group_size: int = 128,
+    nbit: int = 4,
+    has_weight_zeros: bool = False,
+    target: str = "native",
+):
+
+    def apply(weight):
+        assert weight.shape[-1] % group_size == 0
+        assert weight.device == torch.device("cpu"), "Only CPU is supported"
+        use_hqq = False
+        layout = Linear8BitActXBitWeightLayout(
+            nbit=nbit, group_size=group_size, target=target
+        )
+        mapping_type = MappingType.ASYMMETRIC
+        eps = torch.finfo(torch.float32).eps
+        block_size = (1, group_size)
+        target_dtype = torch.int32
+        quant_min = -(1 << (nbit - 1))
+        quant_max = (1 << (nbit - 1)) - 1
+        zero_point_dtype = torch.int8
+        preserve_zero = has_weight_zeros
+        zero_point_domain = ZeroPointDomain.INT if has_weight_zeros else ZeroPointDomain.NONE
+        # Note: this works differently than other quantizers because the dynamic
+        # activation quantization is fused with the kernel/op (and static activation quantization 
+        # is not supported).  
+        return to_affine_quantized_intx(
+            weight,
+            mapping_type,
+            block_size,
+            target_dtype,
+            quant_min,
+            quant_max,
+            eps,
+            zero_point_dtype=zero_point_dtype,
+            preserve_zero=preserve_zero,
+            zero_point_domain=zero_point_domain,
+            _layout=layout,
+            use_hqq=use_hqq,
+        )
+
+    return _get_linear_subclass_inserter(apply)
--- a/torchao/quantization/quant_primitives.py
+++ b/torchao/quantization/quant_primitives.py
@@ -60,9 +60,11 @@ class ZeroPointDomain(Enum):
 
     integer domain: quantized_val = (float_val / scale) (integer) + zero_point (integer)
     float domain: quantized_val = (float_val - (zero_point (float) - scale * mid_point)) / scale
+    none domain: quantized_val = (float_val / scale)
     """
     INT = auto()
     FLOAT = auto()
+    NONE = auto()
 
 class TorchAODType(Enum):
     """
@@ -344,6 +346,9 @@ def _quantize_affine_no_dtype_cast(
         quant = torch.clamp(
             torch.round(input * (1.0 / scale)) + zero_point, quant_min, quant_max
         )
+    elif zero_point_domain == ZeroPointDomain.NONE.name:
+        assert zero_point is None, "zero_point should be None when zero_point_domain is NONE"
+        quant = torch.clamp(torch.round(input * (1.0 / scale)), quant_min, quant_max)
     elif zero_point_domain is None:
         # This case handles quantization for float8 we expect no zero point and no zero point domain
         assert zero_point is None, "zero_point should be None when zero_point_domain is None"
@@ -476,6 +481,10 @@ def _dequantize_affine_no_dtype_check(
         if zero_point is not None:
             dequant = dequant - zero_point.to(torch.int32)
         dequant = dequant.to(output_dtype)
+        dequant = dequant * scale
+    elif zero_point_domain == ZeroPointDomain.NONE.name:
+        assert zero_point is None, "zero_point should be None when zero_point_domain is NONE"
+        dequant = input.to(output_dtype)
         dequant = dequant * scale
     elif zero_point_domain is None:
         # This case handles dequantization for float8 we expect no zero point and no zero point domain
@@ -813,15 +822,20 @@ def _choose_qparams_affine(
         assert mapping_type == MappingType.ASYMMETRIC.name
         scale = (max_val_pos - min_val_neg) / float(quant_max - quant_min)
         scale = torch.clamp(scale, min=eps)
-        if preserve_zero:
-            zero_point = quant_min - torch.round(min_val_neg / scale)
-            zero_point = torch.clamp(zero_point, quant_min, quant_max)
+        if zero_point_domain == ZeroPointDomain.NONE.name:
+            zero_point = None
         else:
-            assert zero_point_domain == ZeroPointDomain.FLOAT.name, "if not preserve_zero, zero_point must be in FLOAT domain"
-            mid_point = (quant_max + quant_min + 1) / 2
-            zero_point = min_val_neg + scale * mid_point
-
-    return scale.to(dtype=scale_dtype), zero_point.to(dtype=zero_point_dtype)
+            if preserve_zero:
+                zero_point = quant_min - torch.round(min_val_neg / scale)
+                zero_point = torch.clamp(zero_point, quant_min, quant_max)
+            else:
+                assert zero_point_domain == ZeroPointDomain.FLOAT.name, "if not preserve_zero, zero_point must be in FLOAT domain"
+                mid_point = (quant_max + quant_min + 1) / 2
+                zero_point = min_val_neg + scale * mid_point
+    
+    if zero_point is not None:
+        zero_point = zero_point.to(dtype=zero_point_dtype)
+    return scale.to(dtype=scale_dtype), zero_point
 
 
 # HQQ
--- a/torchao/experimental/_linear_8bit_act_xbit_weight_layout.py
+++ b/torchao/experimental/_linear_8bit_act_xbit_weight_layout.py
@@ -1,0 +1,375 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the license found in the
+# LICENSE file in the root directory of this source tree.
+
+from enum import auto, Enum
+
+import logging
+from typing import List, Optional, Tuple
+
+import torch
+from torch.utils._python_dispatch import return_and_correct_aliasing
+from torchao.dtypes.affine_quantized_tensor import (
+    AQTTensorImpl,
+    register_aqt_quantized_linear_dispatch,
+    register_layout,
+)
+from torchao.dtypes.utils import Layout
+from torchao.quantization.quant_primitives import (
+    MappingType,
+    ZeroPointDomain,
+)
+
+from torchao.quantization.quant_api import to_affine_quantized_intx
+
+
+logger = logging.getLogger(__name__)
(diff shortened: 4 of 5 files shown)


torchao/quantization/quant_primitives.py
```diff
@@ -300,7 +300,7 @@ def _quantize_affine_no_dtype_cast(
     elif zero_point_domain is None:
         # This case handles quantization for float8 we expect no zero point and no zero point domain
         assert zero_point is None, "zero_point should be None when zero_point_domain is None"
-        quant = torch.clamp(input * scale.reciprocal(), quant_min, quant_max)
+        quant = torch.clamp(torch.round(input * (1.0 / scale)), quant_min, quant_max)
```
metascroy, line 303: @jerryzh168 to confirm if this is OK.  It was needed to match behavior of other quantizer.

torchao/quantization/quant_primitives.py
```diff
@@ -766,13 +766,16 @@ def _choose_qparams_affine(
         assert mapping_type == MappingType.ASYMMETRIC.name
         scale = (max_val_pos - min_val_neg) / float(quant_max - quant_min)
         scale = torch.clamp(scale, min=eps)
-        if preserve_zero:
-            zero_point = quant_min - torch.round(min_val_neg / scale)
-            zero_point = torch.clamp(zero_point, quant_min, quant_max)
+        if zero_point_domain is None:
```
metascroy, line 769: @jerryzh168 confirm if this is OK.  It was needed to get scale-only quantization in affine_quantized_tensor

torchao/experimental/tests/test_linear_8bit_act_xbit_weight_subclass_quantizer.py
```diff
@@ -0,0 +1,153 @@
...
+        )
+
+        unwrapped_model = copy.deepcopy(model)
+        unwrap_tensor_subclass(model)
+
+        print("Exporting quantized model")
+        exported = torch.export.export(model, (activations,))
+        
+        print("Compiling quantized model")
+        compiled = torch.compile(unwrapped_model)
```
metascroy, line 135: @jerryzh168 do you see unification for compile and export coming soon?  The fact that one requires an unwrapped tensor subclass and the other requires a wrapped one makes using this API inconvenient in torchchat.

metascroy:
@kimishpatel @jerryzh168 moving review over to GH.  I hope I've addressed most of your concerns.

@jerryzh168, the fact that compile and export cannot handle the same model (export requires an unwrapped tensor subclass, compile requires a wrapped one, and eager can handle both) makes using this API inconvenient in torchchat.  Do you know if there is planned unification there?

torchao/experimental/_linear_8bit_act_xbit_weight_subclass_quantizer.py
```diff
@@ -0,0 +1,408 @@
...
+    assert input_tensor.dim() >= 3
+    lead_shape = input_tensor.shape[0:-2]
+    m, k = input_tensor.shape[-2], input_tensor.shape[-1]
+    n, k_ = weight_tensor.shape
+    assert k_ == k
+
+    input_tensor = input_tensor.reshape(-1, m, k)
+
+    res = [
+        _impl_2d(input_tensor[i, :, :], weight_tensor)
```
kimishpatel, line 382: Why are you doing it like this? You can just fuse first N dim. LIke line 379 should be
`input_tensor = input_tensor.reshape(-1, k)` no?

torchao/experimental/_linear_8bit_act_xbit_weight_subclass_quantizer.py
```diff
@@ -0,0 +1,408 @@
...
+        m, k = input_tensor.shape
+        n, k_ = weight_tensor.shape
+        assert k_ == k
+        group_size = weight_tensor.layout_tensor.layout_type.group_size
+        packed_weight = weight_tensor.layout_tensor.packed_weight
+
+        # TODO(T200095131): convert self.n, self.k, self.group_size to
+        # int when supported by AOTI
+        args = (
+            input_tensor,
```
kimishpatel, line 345: Does this expect input_tensor to be of fp32 type? Would it work with bf16? If not we should assert for those

metascroy: I can add an assert here too, although this is already asserted in the C++ layer

torchao/quantization/quant_primitives.py
jerryzh168: hmmm, it might be fine as long as all the tests passes I think

torchao/quantization/quant_primitives.py
jerryzh168: OK, should zero_point be `None` here?

torchao/experimental/_linear_8bit_act_xbit_weight_subclass_quantizer.py
```diff
@@ -0,0 +1,408 @@
...
+            ),
+            quant_min=None,  # TODO: why is this an arg for this function
+            quant_max=None,  # TODO: why is this an arg for this function
+            dtype=None,  # TODO: why is this an arg for this function
+            group_size=group_size,
+            output_dtype=torch.float32,
+        )
+
+        # Quantize activations
+        activation_scales, activation_zeros = choose_qparams_affine(
```
jerryzh168, line 269: dynamic quantization should be reusing affine quantized tensor, example: https://github.com/pytorch/ao/blob/900f9acccb8bc3c102d3a7095d1b1f0355057bec/torchao/quantization/quant_api.py#L586

why is this calling these functions here?

torchao/experimental/_linear_8bit_act_xbit_weight_subclass_quantizer.py
```diff
@@ -0,0 +1,408 @@
...
+        weight_qvals = weight_tensor.layout_tensor.packed_weight.to(torch.int32)
+        weight_scales = weight_tensor.layout_tensor.scale
+        weight_zeros = weight_tensor.layout_tensor.zero_point
+        group_size = weight_tensor.layout_tensor.layout_type.group_size
+        has_weight_zeros = weight_zeros is not None
+        m, k = input_tensor.shape
+        n, k_ = weight_tensor.shape
+        assert k_ == k
+
+        weights_dequantized = dequantize_per_channel_group(
```
jerryzh168, line 255: we tend to use quantize_affine/dequantize_affine I think, also this should probably be:

`weights_dequantized = weight_tensor.dequantize()`?

torchao/experimental/_linear_8bit_act_xbit_weight_subclass_quantizer.py
```diff
@@ -0,0 +1,408 @@
...
+
+
+def _linear_int8_dynamic_activation_intx_weight_check(
+    input_tensor, weight_tensor, bias
+):
+    layout_type = weight_tensor.layout_tensor.layout_type
+    return isinstance(layout_type, IntxWeightLayoutType) and bias is None
+
+
+def _linear_int8_dynamic_activation_intx_weight_fallback_impl(
```
jerryzh168, line 236: see fallback here: https://github.com/pytorch/ao/blob/900f9acccb8bc3c102d3a7095d1b1f0355057bec/torchao/dtypes/affine_quantized_tensor.py#L1760-L1764 I think you should be able to do the same here

torchao/experimental/_linear_8bit_act_xbit_weight_subclass_quantizer.py
```diff
@@ -0,0 +1,408 @@
...
+
+
+def _target_equal(target, other):
+    assert other in _VALID_TARGETS
+    assert target in _VALID_TARGETS
+    return target == other
+
+
+# This format is intended for use with int8 dynamic quantization
+class IntxWeightLayoutType(LayoutType):
```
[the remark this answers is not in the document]
metascroy: This layout targets the linear_8bit_act_xbit_weight kernels in torchao/experimental.  They can run on all PyTorch platforms (eager, AOTI, compile, and ExecuTorch), so adding ExecuTorch doesn't make sense.  Open to naming suggestions.

torchao/experimental/_linear_8bit_act_xbit_weight_subclass_quantizer.py
metascroy: That function doesn't look equivalent?  It looks like the quantization is symmetric.

torchao/quantization/quant_primitives.py
metascroy: I think the tests do not pass because it is slightly different quantization logic.  It looks more sensible to me to round before truncating, but I can also drop this change.

We can do a perplexity study when moving from the other quantizer to this one in torchchat.  But I have narrowed done this as being the only numerical difference between the two.

torchao/quantization/quant_primitives.py
metascroy: I could make it None, but that changes the return type of this method from Tuple[Tensor, Tensor] to Tuple[Tensor, Optional[Tensor]]

kimishpatel:
I plan to review some time tomorrow

torchao/experimental/_linear_8bit_act_xbit_weight_subclass_quantizer.py
metascroy: Used weight_tensor.dequantize() below

torchao/experimental/_linear_8bit_act_xbit_weight_subclass_quantizer.py
metascroy: Actually, I like keeping the assert in the C++ layer.  I think it makes sense for the op to assert its inputs rather than the thing calling the op to assert the inputs.

torchao/experimental/quant_api.py
```diff
@@ -317,3 +317,53 @@ def quantize(self, model: nn.Module) -> nn.Module:
...
+def int8_dyn_act_intx_weight(
+    group_size: int = 128,
+    nbit: int = 4,
+    has_weight_zeros: bool = False,
+    target: str = "native",
+):
+
+    def apply(weight):
+        assert weight.shape[-1] % group_size == 0
+        assert weight.device == torch.device("cpu"), "Only CPU is supported"
```
metascroy, line 340: Add CPU device assert here

torchao/quantization/quant_primitives.py
jerryzh168: I see, we don't want to break tests I think, but if this is better for torchchat we can create a new quant primitive op or add a new option here I feel

torchao/quantization/quant_primitives.py
jerryzh168: yeah I think making it None probably makes more sense here

torchao/experimental/_linear_8bit_act_xbit_weight_subclass_quantizer.py
jerryzh168: you can choose a different mapping type,
I mean we can use https://github.com/pytorch/ao/blob/900f9acccb8bc3c102d3a7095d1b1f0355057bec/torchao/quantization/quant_api.py#L613 to do activation quantization, maybe "Dynamic Activation Quantization + Weight Quantization" in https://github.com/pytorch/ao/issues/391 is clearer in explaining this, please let me know if it makes sense

torchao/experimental/_linear_8bit_act_xbit_weight_subclass_quantizer.py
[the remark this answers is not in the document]
jerryzh168: OK I feel it makes sense to just include the kernel name here then, like `Linear8BitActXBitWeightLayoutType` (also we are renaming LayoutType to Layout to make things clearer) you'll probably see this after rebase

torchao/experimental/_linear_8bit_act_xbit_weight_subclass_quantizer.py
metascroy: Sorry, I don't follow what you're suggesting.  I have weights_dequantized on line 260, and I need activations_dequantized so I can call torch.matmul(activations_dequantized, weights_dequantized.transpose(1, 0)) on line 296.

I'm not sure what I'm suppose to replace the code that generates activations_dequantized with.

jerryzh168 (changes requested):
some requested changes, please see comments

torchao/experimental/_linear_8bit_act_xbit_weight_subclass_quantizer.py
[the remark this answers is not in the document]
jerryzh168: @kimishpatel I don't quite follow why "it's not feasible to enumerate all possible ways in which weights can be packed" but I think we are not asking people to use AffineQuantizedTensor at this stage as I discussed in the post, so feel free to contribute in a way you feel makes more sense, we can always merge/refactor later if needed. although a high level API + brief description of implementation might be helpful for us to understand what you have in mind.

also for "Quant API shouldnt really be in the business of understanding packed layout information. This should really be left to subclasses of AQT" I think it might be better/easier to just copy paste AQT and create a new tensor subclass in that case instead of subclassing AQT, we are not clear whether we should have inheriting AQT as an extension point yet

torchao/quantization/quant_primitives.py
```diff
@@ -344,6 +346,8 @@ def _quantize_affine_no_dtype_cast(
         quant = torch.clamp(
             torch.round(input * (1.0 / scale)) + zero_point, quant_min, quant_max
         )
+    elif zero_point_domain == ZeroPointDomain.ZERO.name:
+        quant = torch.clamp(torch.round(input * (1.0 / scale)), quant_min, quant_max)
```
jerryzh168, line 350: nit: maybe add an assert here to check zero_point is None

torchao/experimental/docs/readme.md
jerryzh168, line 1: readme.md --> README.md?

torchao/experimental/_linear_8bit_act_xbit_weight_subclass_quantizer.py
jerryzh168, line 1: nit: I think we should remove the quantizer in the name, `_linear_8bit_act_xbit_weight_layout.py` might be more appropriate

torchao/experimental/_linear_8bit_act_xbit_weight_subclass_quantizer.py
```diff
@@ -0,0 +1,397 @@
...
+        group_size = weight_tensor.tensor_impl.get_layout().group_size
+        has_weight_zeros = weight_zeros is not None
+        m, k = input_tensor.shape
+        n, k_ = weight_tensor.shape
+        assert k_ == k
+
+        weights_dequantized = weight_tensor.dequantize()
+
+        # Quantize activations
+        activation_scales, activation_zeros = choose_qparams_affine(
```
jerryzh168, line 269: I think we can probably use `to_affine_quantized_intx` and `dequantize_affine` here to quantize activation?

torchao/experimental/_linear_8bit_act_xbit_weight_subclass_quantizer.py
```diff
@@ -0,0 +1,397 @@
...
+
+from enum import auto, Enum
+
+import logging
+from typing import List, Optional, Tuple
+
+import torch
+from torch.ao.quantization.fx._decomposed import (
+    dequantize_per_channel_group,
+    quantize_per_channel_group,
```
jerryzh168, line 15: these ops are a bit deprecated so we tend not to use these if possible

torchao/experimental/_linear_8bit_act_xbit_weight_subclass_quantizer.py
```diff
@@ -0,0 +1,397 @@
...
+    register_aqt_quantized_linear_dispatch,
+    register_layout,
+)
+from torchao.dtypes.utils import Layout
+from torchao.quantization.quant_primitives import (
+    choose_qparams_affine,
+    MappingType,
+    ZeroPointDomain,
+)
+from torchao.utils import TorchAOBaseTensor
```
jerryzh168, line 29: nit: looks like not used

jerryzh168 (approved):
LGTM, left a few more nit comments

torchao/quantization/quant_primitives.py
```diff
@@ -60,9 +60,11 @@ class ZeroPointDomain(Enum):
 
     integer domain: quantized_val = (float_val / scale) (integer) + zero_point (integer)
     float domain: quantized_val = (float_val - (zero_point (float) - scale * mid_point)) / scale
+    zero domain: quantized_val = (float_val / scale)
     """
     INT = auto()
     FLOAT = auto()
+    ZERO = auto()
```
jerryzh168, line 67: btw, I feel maybe NONE would be better?

torchao/experimental/quant_api.py
```diff
@@ -463,3 +463,56 @@ def quantize(self, model: nn.Module) -> nn.Module:
...
+from torchao.experimental._linear_8bit_act_xbit_weight_layout import Linear8BitActXBitWeightLayout
+from torchao.quantization.quant_api import (
+    _get_linear_subclass_inserter,
+    MappingType,
+    to_affine_quantized_intx,
+    ZeroPointDomain,
+)
+
+
+def int8_dyn_act_intx_weight(
```
jerryzh168, line 477: we are spelling out all the words in our API so far, so this should be `int8_dynamic_activation_intx_weight` I think

torchao/experimental/tests/test_linear_8bit_act_xbit_weight_subclass_quantizer.py
jerryzh168, line 1: can you rename the tests as well

jerryzh168 (approved):

Issue #1000, opened 2024-10-03
TestAffineQuantizedTensorParallel fails on H100


jainapurva:
The TestAffineQuantizedTensorParallel fails on H100 for bfloat16, float16 and float32 dtypes. Need to debug the reason and fix for it.

Error:
```
ERROR: test_tp_float32 (torchao.testing.utils.TorchAOTensorParallelTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/appy/.conda/envs/dev_ao/lib/python3.9/site-packages/torch/testing/_internal/common_distributed.py", line 541, in wrapper
    self._join_processes(fn)
  File "/home/appy/.conda/envs/dev_ao/lib/python3.9/site-packages/torch/testing/_internal/common_distributed.py", line 767, in _join_processes
    self._check_return_codes(elapsed_time)
  File "/home/appy/.conda/envs/dev_ao/lib/python3.9/site-packages/torch/testing/_internal/common_distributed.py", line 821, in _check_return_codes
    raise RuntimeError(
RuntimeError: Process 0 terminated or timed out after 300.0712020397186 seconds

----------------------------------------------------------------------
Ran 9 tests in 2701.009s

FAILED (errors=9)
/home/appy/.conda/envs/dev_ao/lib/python3.9/multiprocessing/resource_tracker.py:216: UserWarning: resource_tracker: There appear to be 36 leaked semaphore objects to clean up at shutdown
  warnings.warn('resource_tracker: There appear to be %d 
```

jainapurva:
The issue was caused due to pytorch version.

PR #1001, opened 2024-10-03
Temporary Fix: Skip TestAffineQuantizedTensorParallel on H100

jainapurva:
The current aqt test runs on bfloat16, float16 and float32, but the test doesn't run on H100 for these dtypes. As a temporray fix, skipping the test if H100

Created issue to track this: https://github.com/pytorch/ao/issues/1000

### DIFF
--- a/test/dtypes/test_affine_quantized_tensor_parallel.py
+++ b/test/dtypes/test_affine_quantized_tensor_parallel.py
@@ -8,10 +8,7 @@ class TestAffineQuantizedTensorParallel(TorchAOTensorParallelTestCase):
 
 is_H100 = torch.cuda.is_available() and torch.cuda.get_device_capability() >= (9, 0)
 
-if not is_H100:
-    copy_tests(TorchAOTensorParallelTestCase, TestAffineQuantizedTensorParallel, "aqt_tp")
-else:
-    print("Skipping TestAffineQuantizedTensorParallel because it doesn't run on H100")
+copy_tests(TorchAOTensorParallelTestCase, TestAffineQuantizedTensorParallel, "aqt_tp")
 
 if __name__ == "__main__":
     if not is_H100:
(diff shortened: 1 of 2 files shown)


jerryzh168 (approved):

test/dtypes/test_affine_quantized_tensor_parallel.py
```diff
@@ -1,12 +1,17 @@
...
+is_H100 = torch.cuda.is_available() and torch.cuda.get_device_capability() >= (9, 0)
 
 copy_tests(TorchAOTensorParallelTestCase, TestAffineQuantizedTensorParallel, "aqt_tp")
 
 if __name__ == "__main__":
-    run_tests()
+    if not is_H100:
+        run_tests()
+    else:
+        print("Skipping TestAffineQuantizedTensorParallel: not supported on H100")
```
msaroufim, line 17: I'd prefer we explicitly put a skip test on the skipped test, this will also mess up the pytest output

test/dtypes/test_affine_quantized_tensor_parallel.py
```diff
@@ -1,12 +1,17 @@
 from torchao.testing.utils import copy_tests, TorchAOTensorParallelTestCase
 from torch.testing._internal.common_utils import run_tests
 from torchao.quantization import int8_weight_only
+import torch
```
msaroufim, line 4: do you still see the error if you move the import torch statement above any ao imports?

jainapurva: Moving torch as the first import does resolve some issues, but there are still issues of compatibility with H100. I'm working on another PR to add support for H100 (specially float8). https://github.com/pytorch/ao/pull/1003

jainapurva:
- [ ]

PR #1003, opened 2024-10-03
Add Float8 support for AQT tensor parallel

jainapurva:
Add support to AQT Tensor Parallel for Float8 Weight Only

      - Added op support for aten.slice.Tensor
      - Added kernel in AQT dispatcher for float8-weight-only
      - Added AQT tensor parallel test case, can be run on H100

### DIFF
--- a/test/dtypes/test_affine_quantized_tensor_parallel.py
+++ b/test/dtypes/test_affine_quantized_tensor_parallel.py
@@ -1,12 +1,17 @@
+import torch
 from torchao.testing.utils import copy_tests, TorchAOTensorParallelTestCase
 from torch.testing._internal.common_utils import run_tests
-from torchao.quantization import int8_weight_only
+from torchao.quantization import int8_weight_only, float8_weight_only
 
-class TestAffineQuantizedTensorParallel(TorchAOTensorParallelTestCase):
-    pass
+class TestInt8woAffineQuantizedTensorParallel(TorchAOTensorParallelTestCase):
+    QUANT_METHOD_FN = staticmethod(int8_weight_only)
+copy_tests(TorchAOTensorParallelTestCase, TestInt8woAffineQuantizedTensorParallel, "int8wo_tp")
 
-
-copy_tests(TorchAOTensorParallelTestCase, TestAffineQuantizedTensorParallel, "aqt_tp")
+# Run only on H100
+if torch.cuda.is_available() and torch.cuda.get_device_capability() >= (9, 0):
+    class TestFloat8woAffineQuantizedTensorParallel(TorchAOTensorParallelTestCase):
+        QUANT_METHOD_FN = staticmethod(float8_weight_only)
+    copy_tests(TorchAOTensorParallelTestCase, TestFloat8woAffineQuantizedTensorParallel, "fp8wo_tp")
 
 if __name__ == "__main__":
     run_tests()
--- a/torchao/dtypes/affine_quantized_tensor.py
+++ b/torchao/dtypes/affine_quantized_tensor.py
@@ -1094,20 +1094,31 @@ class Float8AQTLayout(AQTLayout):
             return return_and_correct_aliasing(
                 func, args, kwargs, args[0]._apply_fn_to_data(torch.detach)
             )
-        if func is aten.clone.default:
+        elif func is aten.clone.default:
             return return_and_correct_aliasing(
                 func, args, kwargs, args[0]._apply_fn_to_data(torch.clone)
             )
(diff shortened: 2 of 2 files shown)


torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -1644,6 +1655,28 @@ def _linear_fp8_act_fp8_weight_impl(
         use_fast_accum=scaled_mm_config.use_fast_accum,
     ).reshape(out_shape)
 
+def _linear_fp_act_fp8_weight_check(
```
jerryzh168, line 1658: this is separate I guess, would be good to put this in a different PR?

jainapurva: This is needed to run float8 weight-only kernel, and support AQT tensor parallel.

test/dtypes/test_affine_quantized_tensor_parallel.py
```diff
@@ -1,12 +1,17 @@
 from torchao.testing.utils import copy_tests, TorchAOTensorParallelTestCase
 from torch.testing._internal.common_utils import run_tests
-from torchao.quantization import int8_weight_only
+from torchao.quantization import int8_weight_only, float8_weight_only
 
-class TestAffineQuantizedTensorParallel(TorchAOTensorParallelTestCase):
-    pass
+# class TestAffineQuantizedTensorParallel(TorchAOTensorParallelTestCase):
```
jerryzh168, line 5: please revert the changes for int8wo, we should make it explicit by spelling out the method here as well I think `QUANT_METHOD_FN = staticmethod(int8_weight_only)`

torchao/dtypes/affine_quantized_tensor.py
jerryzh168: I thought we already have fallback? https://github.com/pytorch/ao/blob/0ffbf856a2b61ed4dfb2c28376e8f4e801c9b962/torchao/dtypes/affine_quantized_tensor.py#L1723-L1730

jainapurva: https://github.com/pytorch/ao/blob/0ffbf856a2b61ed4dfb2c28376e8f4e801c9b962/torchao/dtypes/affine_quantized_tensor.py#L1691

It check through the dispatchers, to find a kernel, and if it doesn't find one, it raise a no dispatch kernel found error. Hence need a kernel in the dispatch table.

test/dtypes/test_affine_quantized_tensor_parallel.py
```diff
@@ -1,12 +1,15 @@
...
 
-class TestAffineQuantizedTensorParallel(TorchAOTensorParallelTestCase):
-    pass
+class TestInt8woAffineQuantizedTensorParallel(TorchAOTensorParallelTestCase):
+    QUANT_METHOD_FN = staticmethod(int8_weight_only)
+copy_tests(TorchAOTensorParallelTestCase, TestInt8woAffineQuantizedTensorParallel, "int8wo_tp")
 
-
-copy_tests(TorchAOTensorParallelTestCase, TestAffineQuantizedTensorParallel, "aqt_tp")
+# Uncomment the following to test float8wo_affine : works only on H100
```
jerryzh168, line 9: why not add a skip?

torchao/dtypes/affine_quantized_tensor.py
jerryzh168: not sure why this is required, but we can debug separately I think

jerryzh168 (approved):

test/dtypes/test_affine_quantized_tensor_parallel.py
```diff
@@ -1,12 +1,17 @@
...
-class TestAffineQuantizedTensorParallel(TorchAOTensorParallelTestCase):
-    pass
+class TestInt8woAffineQuantizedTensorParallel(TorchAOTensorParallelTestCase):
+    QUANT_METHOD_FN = staticmethod(int8_weight_only)
+copy_tests(TorchAOTensorParallelTestCase, TestInt8woAffineQuantizedTensorParallel, "int8wo_tp")
 
-
-copy_tests(TorchAOTensorParallelTestCase, TestAffineQuantizedTensorParallel, "aqt_tp")
+# Run only on H100
+if torch.cuda.is_available() and torch.cuda.get_device_capability() >= (9, 0):
```
jerryzh168, line 11: nit: might be better to use `unittest.skipIf` to annotate the class I think

PR #1006, opened 2024-10-03
training ir torchao migration

yushangdi:
Summary: Migrate capture_pre_autograd_graph to export_for_training.

Differential Revision: D63859678


test/dtypes/test_uint4.py
```diff
@@ -26,6 +25,12 @@
 )
 import copy
 
+has_export_for_training = False
+try:
```
jerryzh168, line 29: can this be done through a version guard? when does `export_for_training` start to work?

yushangdi: > can this be done through a version guard? when does `export_for_training` start to work?

for now it starts to exist in 2.5.0-rc1. I can make it a version guard. Thanks for the comment!

jerryzh168: we have some utils here btw: https://github.com/pytorch/ao/blob/0ffbf856a2b61ed4dfb2c28376e8f4e801c9b962/torchao/utils.py#L25-L29

but notice that it's does not exactly map to current meaning of version, e.g. 

TORCH_VERSION_AT_LEAST_2_5 == 2.5.0 or 2.5.0.dev
(instead of 2.5.0 + 2.6.0.dev)

jerryzh168 (approved):

yushangdi:
@jerryzh168 Can I get a stamp on the internall Diff as well? Thanks!

PR #1008, opened 2024-10-03
Pin torch nightly - Oct 1

jainapurva:
Due to memory leak and other additional compile issues with latest torch-nightly, pinning it to last stable version torch==2.6.0.dev20241001+cu121

### DIFF
--- a/test/prototype/test_quantized_training.py
+++ b/test/prototype/test_quantized_training.py
@@ -188,6 +188,7 @@ class TestQuantizedTraining(TestCase):
         assert snr(inputs_ref.grad, inputs_int8mp.grad) > 20
         assert snr(linear.weight.grad, linear_int8mp.weight.grad) > 20
 
+    @pytest.mark.skip('Flaky on CI')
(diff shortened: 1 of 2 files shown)


jerryzh168 (approved):
we can land after CI is green

test/prototype/test_quantized_training.py
```diff
@@ -188,6 +188,7 @@ def snr(ref, actual):
         assert snr(inputs_ref.grad, inputs_int8mp.grad) > 20
         assert snr(linear.weight.grad, linear_int8mp.weight.grad) > 20
 
+    @pytest.mark.skip('Flaky on CI')
```
msaroufim, line 191: should only skip if the version of pytorch is at least 2.6, the other tests seemed green

msaroufim (approved):
@jainapurva gonna merge this for now to unblock others from merging code but could you please open 2 separate issues
1. Assign to @gau-nernst the bitnet training skip
2. Another issue on the IMA error we're seeing

PR #1009, opened 2024-10-04
Re-run regression tests with CUDA_LAUNCH_BLOCKING

malfet:


msaroufim:
Huh funny this made the failure go away. Seems like we're dealing with a race condition?

PR #1016, opened 2024-10-04
[not for land] float8 training -> quantize_

vkuzo:
Summary:

This is just a smoke test of how this would look like.  Let's discuss?

Test Plan:

```
// only checked one test
pytest test/float8/test_numerics_integration.py -s -x
```

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/float8/test_numerics_integration.py
+++ b/test/float8/test_numerics_integration.py
@@ -25,6 +25,7 @@ from torchao.float8.float8_linear_utils import (
     linear_requires_sync,
     sync_float8_amax_and_scale_history,
 )
+from torchao.quantization.quant_api import quantize_, float8_for_training
 from torchao.float8.float8_utils import compute_error, IS_ROCM
 
 is_cuda_8_9 = torch.cuda.is_available() and torch.cuda.get_device_capability() >= (8, 9)
@@ -144,10 +145,12 @@ class TestFloat8NumericsIntegrationTest:
             cast_config_grad_output=cast_config_grad_output,
         )
 
-        convert_to_float8_training(
-            model_fp8,
-            config=config,
-        )
+        # convert_to_float8_training(
+        #     model_fp8,
+        #     config=config,
+        # )
+        quantize_(model_fp8, float8_for_training(config))
+        print(model_fp8)
 
         lr = 0.01
         optim_ref = torch.optim.SGD(model_ref.parameters(), lr=lr)
--- a/torchao/quantization/quant_api.py
+++ b/torchao/quantization/quant_api.py
@@ -73,6 +73,9 @@ from torchao.float8.inference import Float8MMConfig
 
 from torchao.quantization.observer import PerTensor, PerRow, get_block_size
 
+from torchao.float8 import Float8LinearConfig
+from torchao.float8.float8_linear import Float8Linear
+
 logger = logging.getLogger(__name__)
 
 __all__ = [
@@ -401,10 +404,25 @@ def _get_linear_subclass_inserter(constructor, *, allow_requires_grad=False, **k
     return insert_subclass
 
 def quantize_(
+    # NOT FOR LAND: this is the same across inference and torchao.float8
     model: torch.nn.Module,
+
+    # NOT FOR LAND: this seems to leak implementation details, in the future
+    # can this be named in a more generic way without mentioning the 
+    # implementation? Short term, seems ok to have float8 training use this
+    # evern though the implementation is different from 100% tensor subclass 
+    # based workflows.
     apply_tensor_subclass: Callable[[torch.nn.Module], torch.nn.Module],
+
+    # NOT FOR LAND: this is the same across inference and torchao.float8
     filter_fn: Optional[Callable[[torch.nn.Module, str], bool]] = None,
+
+    # NOT FOR LAND: this seems inference specific, in the future would we 
+    # consider renaming or some other way to make things more clear?
     set_inductor_config: bool = True,
+
+    # NOT FOR LAND: this seems inference specific, in the future would we 
+    # consider renaming or some other way to make things more clear?
     device: Optional[torch.types.Device] = None,
 ):
     """Convert the weight of linear modules in the model with `apply_tensor_subclass`, model is modified inplace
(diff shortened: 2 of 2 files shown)


torchao/quantization/quant_api.py
```diff
@@ -401,10 +404,25 @@ def insert_subclass(lin):
...
+    # evern though the implementation is different from 100% tensor subclass 
+    # based workflows.
     apply_tensor_subclass: Callable[[torch.nn.Module], torch.nn.Module],
+
+    # NOT FOR LAND: this is the same across inference and torchao.float8
     filter_fn: Optional[Callable[[torch.nn.Module, str], bool]] = None,
+
+    # NOT FOR LAND: this seems inference specific, in the future would we 
+    # consider renaming or some other way to make things more clear?
     set_inductor_config: bool = True,
```
msaroufim, line 422: yeah maybe these shouldn't be here

torchao/quantization/quant_api.py
```diff
@@ -401,10 +404,25 @@ def insert_subclass(lin):
...
+
+    # NOT FOR LAND: this seems inference specific, in the future would we 
+    # consider renaming or some other way to make things more clear?
     device: Optional[torch.types.Device] = None,
```
msaroufim, line 426: this one is more about quantizing layer by layer to avoid peak vram consumption in full precision. would training not also benefit from this?

Granted name is maybe not great

test/float8/test_numerics_integration.py
```diff
@@ -144,10 +145,12 @@ def test_encoder_fw_bw(
...
 
-        convert_to_float8_training(
-            model_fp8,
-            config=config,
-        )
+        # convert_to_float8_training(
+        #     model_fp8,
+        #     config=config,
+        # )
+        quantize_(model_fp8, float8_for_training(config))
```
msaroufim, line 152: this was shorter than i thought it would be :D

torchao/quantization/quant_api.py
```diff
@@ -829,6 +847,31 @@ def apply_float8_static_activation_quant(weight: torch.Tensor):
...
+# TO DISCUSS:
+# below currently just band aids float8 training in without actually changing 
+# how float8 training is configured. Is this fine? Is this a no-go? Somewhere in the
+# middle? How strongly do we care about config consistency among various 
+# projects? 
+# Vasiliy: I'm cool to conform to some other config standard, I would just say 
+# that a config (and not callables/partials) does seem more natural for 
+# training as the # of things to configure is significantly higher compared to
+# your average inference setup.
+def float8_for_training(
```
HDCharles, line 859: I think it would be cool if float8_for_training had usage options for both the config and non-config use cases. I don't know if thats feasible but i think having config is fine, it'd be a little better for random OSS users if there was a more english readable way to set it up.

or maybe its just down to some helpers? I think at the least we should have something like 

quantize_(mod, float8_for_training(config=get_default_config(dtype=fp16)))

torchao/quantization/quant_api.py
```diff
@@ -401,10 +404,25 @@ def insert_subclass(lin):
...
 def quantize_(
+    # NOT FOR LAND: this is the same across inference and torchao.float8
     model: torch.nn.Module,
+
+    # NOT FOR LAND: this seems to leak implementation details, in the future
+    # can this be named in a more generic way without mentioning the 
+    # implementation? Short term, seems ok to have float8 training use this
+    # evern though the implementation is different from 100% tensor subclass 
+    # based workflows.
     apply_tensor_subclass: Callable[[torch.nn.Module], torch.nn.Module],
```
HDCharles, line 415: yeah this should be named something different like constructor or transform function.

torchao/quantization/quant_api.py
HDCharles, line 422: we just used to get so many pings where it was like "its not faster" and unsurprisingly the individual didn't set the inductor config. I'm not sure the best way to handle this. Without it our apis also get 100% more complicated since its not a 1 line api anymore. We can move it into the constructors if necessary

torchao/quantization/quant_api.py
HDCharles: yeah, if someone has an fp32 model and want to do fp8 training on cuda, but the fp32 model doesn't fit on cuda, i'm confused how

quantize_(model, train_fp8, device="cuda")

isn't what we would want?

is it an issue because everything is supposed to be in the config?

test/float8/test_numerics_integration.py
vkuzo: note that setting the right config values may take longer

torchao/quantization/quant_api.py
vkuzo: I'm not saying it's bad, I'm just saying if training is going to be in here then this setting could be renamed `set_inductor_config_to_recommended_inference_settings`, etc

torchao/quantization/quant_api.py
vkuzo: usually "convert model to training" does not do anything to the weights, they stay in high precision throughout training.  Therefore, the issues of "quantize_ takes too long" or "quantize_ makes model not fit" do not appear in that context.

it's possible that someone might want to "quantize for training and modify the weights", just not something I see as a common thing

torchao/quantization/quant_api.py
HDCharles: yeah but its still either A) not on by default or B) users of non inference use cases have to disable it always.

the more i think about it, the more it seems that having different defaults for different use cases makes sense.

torchao/quantization/quant_api.py
HDCharles: yeah but its off by default

vkuzo: something like off by default and docblock says "for inference" sgtm

torchao/quantization/quant_api.py
```diff
@@ -401,10 +404,25 @@ def insert_subclass(lin):
...
+
+    # NOT FOR LAND: this is the same across inference and torchao.float8
     filter_fn: Optional[Callable[[torch.nn.Module, str], bool]] = None,
```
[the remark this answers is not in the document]
jerryzh168: but if it requires user to manually filter the modules, doesn't it make it more complicated?

alternatively I think the filtering can probably be done in the `apply_tensor_subclass` itself, so we can still keep the user API simple

torchao/quantization/quant_api.py
jerryzh168: maybe just move this to apply_tensor_subclass?

vkuzo: note: `set_inductor_config` is changing inductor configuration, that does not seem related to modifying modules of a model, so moving this to `apply_tensor_subclass` would be confusing to me

PR #1020, opened 2024-10-04
Add generic fake quantized linear for QAT

andrewor14:
* (to be filled)

Summary: This commit adds a generic fake quantized linear module
to replace the uses of the existing more specific QAT linears.
For example, `Int8DynActInt4WeightQATLinear` can be expressed
as follows:

```
from torchao.quantization.prototype.qat.api import FakeQuantizeConfig
from torchao.quantization.prototype.qat.linear import FakeQuantizedLinear

activation_config = FakeQuantizeConfig(
    bit_width=8,
    granularity="per_token",
    symmetric=False,
    dynamic=True,
)
weight_config = FakeQuantizeConfig(
    bit_width=4,
    group_size=8,
    symmetric=True,
    dynamic=True,
)
fq_linear = FakeQuantizedLinear(
    16, 32, False, activation_config, weight_config,
)
```

The main motivation is to provide a more flexible way to perform
QAT on models with linear layers. Previously, we would have to
create a new linear class every time we wish to experiment with
different fake quantization settings, e.g. different group size
or different bit width. Now we can express this easily using a
single linear module.

Test Plan:
python test/quantization/test_qat.py -k test_fake_quantize_config
python test/quantization/test_qat.py -k test_fake_quantized_linear_8da4w
python test/quantization/test_qat.py -k test_fake_quantized_linear_4w

### DIFF
--- a/torchao/quantization/prototype/qat/api.py
+++ b/torchao/quantization/prototype/qat/api.py
@@ -4,11 +4,75 @@
 # This source code is licensed under the license found in the
 # LICENSE file in the root directory of this source tree.
 
-from typing import Any, List
+from dataclasses import dataclass
+from enum import Enum
+from typing import Any, List, Optional
 
 import torch
 
 from torchao.quantization.unified import TwoStepQuantizer
+from torchao.quantization.quant_primitives import ZeroPointDomain
+
+
+class QuantizationGranularity(Enum):
+    PER_CHANNEL = "per_channel"
+    PER_TOKEN = "per_token"
+    PER_GROUP = "per_group"
+
+
+@dataclass
+class FakeQuantizeConfig:
+    """
+    Config for how to fake quantize weights or activations.
+
+    args:
+        bit_width: number of bits to simulate during fake quantization
+        granularity: granularity of scales and zero points, one of:
+            'per_token', 'per_channel', or 'per_group'
+        group_size: size of each group for 'per_group' granularity
+        symmetric: whether to use symmetric (default) or asymmetric quantization
+        scale_precision: scale dtype (default torch.fp32)
+        zero_point_precision: zero point dtype (default torch.int32)
+        zero_point_domain: whether zero point is in integer (default) or float domain
+        dynamic: whether to use dynamic (defualt) or static scale and zero points
+        range_learning: whether to learn scale and zero points during training (coming soon)
+    """
+    bit_width: int
+    granularity: Optional[QuantizationGranularity] = None
+    group_size: Optional[int] = None
+    symmetric: bool = True
+    scale_precision: torch.dtype = torch.float32
+    zero_point_precision: torch.dtype = torch.int32
+    zero_point_domain: ZeroPointDomain = ZeroPointDomain.INT
+    dynamic: bool = True
+    range_learning: bool = False
+
+    def __post_init__(self):
+        """
+        Verify that `group_size` and `granularity` are consistent.
+        """
+        if self.group_size is None and self.granularity is None:
+            raise ValueError("At least one of group_size or granularity must be set")
+        if self.granularity == QuantizationGranularity.PER_GROUP and self.group_size is None:
+            raise ValueError("Granularity is 'per_group' but no group_size was set")
+        if self.granularity != QuantizationGranularity.PER_GROUP and self.group_size is not None:
+            if self.granularity is None:
+                self.granularity = QuantizationGranularity.PER_GROUP
+            else:
+                raise ValueError(
+                    "Granularity is '%s' but group_size was set" % self.granularity.value
+                )
+        self._initialized = True
+
+    def __setattr__(self, name: str, value: Any):
+        """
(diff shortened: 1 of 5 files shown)


torchao/quantization/prototype/qat/fake_quantizer.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def forward(self, x: torch.Tensor):
+        """
+        Apply fake quantization to the tensor based on the bit-width,
+        granularity, symmetry, and other properties specified in the config.
+        """
+        if not self.enabled:
+            return x
+
+        if self.config.granularity == QuantizationGranularity.PER_TOKEN:
```
jerryzh168, line 50: we have some config here: https://github.com/pytorch/ao/blob/dec03131484b5717e9fb8b6c0c25f41ec30e9fef/torchao/quantization/observer.py#L20 might be good to reuse

torchao/quantization/prototype/qat/api.py
```diff
@@ -4,11 +4,75 @@
...
+class FakeQuantizeConfig:
+    """
+    Config for how to fake quantize weights or activations.
+
+    args:
+        bit_width: number of bits to simulate during fake quantization
+        granularity: granularity of scales and zero points, one of:
+            'per_token', 'per_channel', or 'per_group'
+        group_size: size of each group for 'per_group' granularity
+        symmetric: whether to use symmetric (default) or asymmetric quantization
```
jerryzh168, line 33: is it better to do `is_symmetric`, or maybe we should use `MappingType` I feel, we now have 3 types: https://github.com/pytorch/ao/blob/dec03131484b5717e9fb8b6c0c25f41ec30e9fef/torchao/quantization/quant_primitives.py#L36

torchao/quantization/prototype/qat/api.py
```diff
@@ -4,11 +4,75 @@
...
+        scale_precision: scale dtype (default torch.fp32)
+        zero_point_precision: zero point dtype (default torch.int32)
+        zero_point_domain: whether zero point is in integer (default) or float domain
+        dynamic: whether to use dynamic (defualt) or static scale and zero points
```
jerryzh168, line 37: might be clearer to do `is_dynamic`

torchao/quantization/prototype/qat/api.py
```diff
@@ -4,11 +4,75 @@
...
+        range_learning: whether to learn scale and zero points during training (coming soon)
+    """
+    bit_width: int
+    granularity: Optional[QuantizationGranularity] = None
+    group_size: Optional[int] = None
```
jerryzh168, line 42: can be reusing https://github.com/pytorch/ao/blob/dec03131484b5717e9fb8b6c0c25f41ec30e9fef/torchao/quantization/observer.py#L20 here and also we plan to rename it to `Granularity` for better clarity

torchao/quantization/quant_primitives.py
```diff
@@ -64,6 +64,18 @@ class ZeroPointDomain(Enum):
     INT = auto()
     FLOAT = auto()
 
+class TorchSignedInt(Enum):
```
jerryzh168, line 67: this can be more general since we may need some other things as well, like https://github.com/pytorch/ao/blob/52775076f833e7301fbac29eb2f4938461f1c94c/torchao/prototype/mx_formats/constants.py#L6-L8

so maybe just `TORCHAO_DTYPE` or something like that

andrewor14: I added a `TorchAODType` enum with fields like INT1 to INT7. What do you think?

torchao/quantization/prototype/qat/api.py
```diff
@@ -4,11 +4,224 @@
...
+
+        # Case 1: Granularity object
+        if isinstance(granularity, Granularity):
+            if not isinstance(granularity, (PerToken, PerAxis, PerGroup)):
+                raise ValueError("Granularity '%s' is not supported" % granularity)
+            if isinstance(granularity, PerAxis) and granularity.axis != 0:
+                raise ValueError("Only axis=0 is supported for PerAxis granularity")
+            return granularity
+
+        # Case 2: str granularity
```
jerryzh168, line 140: nit: will it be too complicated to have multiple ways to specify the same thing?

is this for BC or do people prefer to use str + some other args

jerryzh168 (approved):

torchao/quantization/quant_primitives.py
jerryzh168: yeah sounds good

jerryzh168:
for the error:
  E   AttributeError: module 'torch' has no attribute 'uint1'

these are available after 2.3, but we still have 2.2.2 tests I think

torchao/quantization/prototype/qat/api.py
andrewor14: Yeah people prefer to use str and ints. It's cumbersome to find where to import all our primitives I think

PR #1021, opened 2024-10-05
[low-bit optim] Fix load state dict when device is different

gau-nernst:
In `optim.load_state_dict(state_dict)`, if optim dtype != state_dict dtype, `aten._to_copy.default` is called. This PR simply implements this op and add appropriate tests.

**Update**: In PyTorch pre-2.4, calling `.to(device, dtype)` will not dispatch `aten._to_copy.default` when dtype is the same but device is different. Thus, I have to manually override `.to()` method instead. This is only done for PyTorch pre-2.4. FP8 is not affected since FP8 CUDA requires PyTorch 2.4 anyway. We can remove this hack once we drop 2.3 support.

### DIFF
--- a/test/prototype/test_low_bit_optim.py
+++ b/test/prototype/test_low_bit_optim.py
@@ -97,6 +97,29 @@ class TestOptim(TestCase):
         optim.step()
         optim.zero_grad()
 
+        # test serialization. also test the case CUDA optim loads CPU state dict
+        with tempfile.NamedTemporaryFile() as f:
+            torch.save(optim.state_dict(), f.name)
+            state_dict = torch.load(f.name, map_location="cpu")
+
+        model2 = copy.deepcopy(model)
+        optim2 = getattr(low_bit_optim, optim_name)(model2.parameters())
+        optim2.load_state_dict(state_dict)
+
+        for _ in range(2):
+            x = torch.randn(4, 32, device=device, dtype=dtype)
+
+            model(x).sum().backward()
+            optim.step()
+            optim.zero_grad()
+
+            model2(x).sum().backward()
+            optim2.step()
+            optim2.zero_grad()
+
+        for p1, p2 in zip(model.parameters(), model2.parameters()):
+            torch.testing.assert_close(p2, p1)
+
     @pytest.mark.skipif(bnb is None, reason="bitsandbytes is not availablle")
     @pytest.mark.skipif(not torch.cuda.is_available(), reason="bitsandbytes 8-bit Adam only works for CUDA")
     @pytest.mark.skipif(not TORCH_VERSION_AT_LEAST_2_3, reason="requires PyTorch >= 2.3")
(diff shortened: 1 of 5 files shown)


test/prototype/test_low_bit_optim.py
```diff
@@ -97,6 +97,29 @@ def test_optim_smoke(self, optim_name, dtype, device):
...
+            optim.zero_grad()
+
+            model2(x).sum().backward()
+            optim2.step()
+            optim2.zero_grad()
+
+        for p1, p2 in zip(model.parameters(), model2.parameters()):
+            torch.testing.assert_close(p2, p1)
+
     @pytest.mark.skipif(bnb is None, reason="bitsandbytes is not availablle")
```
bghira, line 123: small typo, `availablle -> available`
```suggestion
    @pytest.mark.skipif(bnb is None, reason="bitsandbytes is not available")
```

torchao/prototype/low_bit_optim/adam.py
```diff
@@ -109,6 +109,7 @@ def step(self, closure=None):
 
 # this will work with any optim state tensor subclass that implements aten.lerp.Scalar and aten.copy_.default
 # and param tensor subclass that implements aten.add_.Tensor, and aten.addcdiv_.default
+# NOTE: should we cast inputs to FP32 to ensure computations are always in FP32?
```
bghira, line 112: i override a few methods to cast the input to the weight dtype because the Flux model occasionally upcasts things to fp32 inside layernorm. you are saying fp32 is more correct than bf16?

torchao/prototype/low_bit_optim/subclass_4bit.py
```diff
@@ -60,8 +61,9 @@ def __tensor_unflatten__(cls, tensor_data_dict, tensor_attributes, outer_size=No
     def dequantize(self, output_dtype=None):
         codes = torch.stack([self.codes >> 4, self.codes & 0b1111], dim=-1)  # unpack
         float_data = dequant_with_qmap(codes, self.qmap, self.scale)
-        dtype = output_dtype or torch.get_default_dtype()
-        return float_data.view(self._shape).to(dtype)
+        if output_dtype is not None:
+            float_data = float_data.to(output_dtype)
+        return float_data.view(self._shape)
```
bghira, line 66: will we encounter non-contiguous tensor? if so, view cannot be used, reshape must be

msaroufim (approved):

torchao/prototype/low_bit_optim/adam.py
gau-nernst: For optimizer step, internal calculations should be done in FP32 to ensure accurate results.

torchao/prototype/low_bit_optim/subclass_4bit.py
gau-nernst: No, unlikely we will have non-contiguous tensor here, because this handles internal data of the tensor subclass, not outputs of another layer in a model.

PR #1024, opened 2024-10-07
Refactor `tiktoken` import bare except

ringohoffman:
We shouldn't suppress the import error if `tiktoken` is not installed. Also do the same thing for `sentencepiece`.

Instead of importing them at the top of the file, we can import them inside the only functions that use them; we can consider a different import structure if we end up needing to access them elsewhere.

This lazy importing should also decrease the load time of these modules.

cc: @msaroufim

### DIFF
--- a/torchao/_models/llama/tokenizer.py
+++ b/torchao/_models/llama/tokenizer.py
@@ -1,14 +1,8 @@
 # copied from https://github.com/pytorch-labs/gpt-fast/blob/main/tokenizer.py
+from __future__ import annotations
 
 import os
-import sentencepiece as spm
-try:
-    import tiktoken
-    from tiktoken.load import load_tiktoken_bpe
-except:
-    pass
 from pathlib import Path
-from typing import Dict
 
 class TokenizerInterface:
     def __init__(self, model_path):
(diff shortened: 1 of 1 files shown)


msaroufim (approved):

torchao/_models/llama/tokenizer.py
```diff
@@ -1,14 +1,8 @@
 # copied from https://github.com/pytorch-labs/gpt-fast/blob/main/tokenizer.py
+from __future__ import annotations
```
msaroufim, line 2: which python version are you using?

msaroufim (approved):

torchao/_models/llama/tokenizer.py
ringohoffman: I'm using 3.9. I saw that that was what you are using in CI.

https://github.com/pytorch/ao/blob/e2301e9dba91fa962d673fdc3b3f0002856a3ba7/.github/workflows/ruff_linter.yml#L18

I think you are right that for what I did, this wouldn't be required for 3.9. I will remove it.

```suggestion
```

PR #1030, opened 2024-10-08
SmoothQuant using tensor subclassing

Xia-Weiwen:
This PR implements SmoothQuant with tensor subclassing.
SmoothQuant is similar to AWQ with the following differences:
- Matmul is computed in int8 instead of floating point (at least at op level)
- The smoothing factor is calculated differently from the equalization scales of AWQ

It provides the following API for quantization:
- `insert_smooth_quant_observer_` inserts observers into the model in-place.
- `smooth_quant` applies SmoothQuant to each linear layer of the model. Use it by calling `torchao.quantization.quantize_`.

Two more APIs for quantization recipe tuning for advanced users:
- `save_smooth_quant_recipe` saves smoothing factors and quantization parameters of an observed model to a JSON file.
- `load_smooth_quant_recipe` loads these parameters from a file to a model with observers inserted.

More details can be found in the torchao/prototype/smoothquant/README.md.

Unit tests are added in test/prototype/test_smoothquant.py
An example is provided in torchao/prototype/smoothquant/example.py


torchao/prototype/smoothquant/api.py
```diff
@@ -0,0 +1,202 @@
...
+    Args:
+        smoothing_factor: The smoothing factor for the layer. Acquired from the layer's observer if None.
+        act_scales: The activation scales for the layer. Acquired from the layer's observer if None.
+        wei_scales: The weight scales for the layer. Acquired from the layer's observer if None.
+    """
+
+    def quantize_weight(observed_linear):
+        target_dtype = torch.int8
+        quant_min = _DTYPE_TO_QVALUE_BOUNDS[target_dtype][0]
+        quant_max = _DTYPE_TO_QVALUE_BOUNDS[target_dtype][1]
```
jerryzh168, line 172: these can be None btw if they are the default values for target_dtype

torchao/prototype/smoothquant/api.py
```diff
@@ -0,0 +1,202 @@
...
+        nonlocal smoothing_factor, act_scales, wei_scales
```
jerryzh168, line 173: why are these nonlocal?

torchao/prototype/smoothquant/api.py
```diff
@@ -0,0 +1,202 @@
...
+            dtype=observed_linear.weight.dtype
+        )
+        linear.weight = torch.nn.Parameter(constructor(observed_linear), requires_grad=False)
+        linear.bias = observed_linear.bias
+        return linear
+
+    return insert_subclass
+
+
+def save_smooth_quant_recipe(model: torch.nn.Module, save_path: str) -> Dict[str, torch.Tensor]:
```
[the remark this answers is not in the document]
Xia-Weiwen: We want to have an API to modify (tune) quantization parameters, i.e. the recipe here. Do you have any concern about adding this API?

torchao/prototype/smoothquant/api.py
Xia-Weiwen: Thanks. Now it's hard-coded to -127/127 because there is a check of quant_min/_max in AQT here: https://github.com/pytorch/ao/blob/e7b33bc91c831d10249c1222c8b4b667f18f28b7/torchao/dtypes/affine_quantized_tensor.py#L1420

torchao/prototype/smoothquant/api.py
Xia-Weiwen: Because I modified these arguments of the outer function. Now it's not needed anymore. Will remove. Thank.

Xia-Weiwen:
Hi @jerryzh168  I added a new tensor subclass `LinearActivationScaleQuantizedTensor` to support `x -> x/scale -> quantize x` for `torch.compile`.

If I use `LinearActivationQuantizedTensor`, the `x/sacle` is done outside the class (by `input_quant_func`) and there is a dynamo error about `scale` during `torch.compile`. I guess it's because the `scale` tensors are not on the graph in this case. Putting the `scale` in the weight tensor solves the problem. And `WeightTensorWithLinearActivationScaleMetadata` does not quantize activation.

Do you have any concern adding this new class? Thanks.

torchao/prototype/smoothquant/api.py
Xia-Weiwen: It's removed. Thanks.

torchao/prototype/smoothquant/api.py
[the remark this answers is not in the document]
jerryzh168: so the state_dict is supposed to be used by other APIs to tune quantization parameters? I think that's fine if you have this use case in mind, is the model with SmoothQuantObservedLinear not serializable by itself?

jerryzh168:
> Hi @jerryzh168 I added a new tensor subclass `LinearActivationScaleQuantizedTensor` to support `x -> x/scale -> quantize x` for `torch.compile`.
> 
> If I use `LinearActivationQuantizedTensor`, the `x/sacle` is done outside the class (by `input_quant_func`) and there is a dynamo error about `scale` during `torch.compile`. I guess it's because the `scale` tensors are not on the graph in this case. Putting the `scale` in the weight tensor solves the problem. And `WeightTensorWithLinearActivationScaleMetadata` does not quantize activation.
> 
> Do you have any concern adding this new class? Thanks.

I think in this case we should be composing `WeightTensorWithLinearActivationScaleMetadata` `LinearActivationQuantizedTensor` together, i.e.
```
weight = to_affine_quantized(float_weight, ...)
# this will quantize input
weight = to_linear_activation_quantized_tensor(weight)
# this will do x / scale
weight = to_weight_tensor_with_linear_activation_scale_metadata(weight)
```

in dispatch time, we first unwrap the outer most tensor subclass, which will be `WeightTensorWithLinearActivationScaleMetadata`, so we'll apply scale to activation, then `LinearActivationQuantizedTensor`, which will quantize activation, and then `AffineQuantizedTensor`

would this work?

torchao/prototype/smoothquant/api.py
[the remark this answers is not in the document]
Xia-Weiwen: `SmoothQuantObservedLinear` is serializable. However, a recipe is more flexible to tune parameters. Thanks.

Xia-Weiwen:
> I think in this case we should be composing `WeightTensorWithLinearActivationScaleMetadata` `LinearActivationQuantizedTensor` together, i.e.
> 
> ```
> weight = to_affine_quantized(float_weight, ...)
> # this will quantize input
> # use https://github.com/pytorch/ao/blob/c87cc9b7286a46e9dfc076fa2417eb9b64ccc807/torchao/quantization/weight_tensor_linear_activation_quantization.py#L13 for static quantization
> weight = to_linear_activation_quantized_tensor(weight)  # dynamic quant
> # this will do x / scale
> weight = to_weight_tensor_with_linear_activation_scale_metadata(weight)
> ```
> 
> in dispatch time, we first unwrap the outer most tensor subclass, which will be `WeightTensorWithLinearActivationScaleMetadata`, so we'll apply scale to activation, then `LinearActivationQuantizedTensor`, which will quantize activation, and then `AffineQuantizedTensor`
> 
> would this work?
> 
> the naming for different tensor subclasses is a bit confusing right now I think, we should cleanup a bit later

It works. Thanks

Xia-Weiwen:
Hi @jerryzh168 It's weird that if I add these lines https://github.com/pytorch/ao/blob/f595ed41b99685cc16fc480ca2218965bb812bed/torchao/kernel/intmm.py#L142C1-L146C1 to avoid overflow of float16, there will be a failure in test_spinquant.py, which does not use float16 at all. And there is another failure with CUDA nightly, whose log is cannot be loaded. And these failures cannot be reproduced in my local environment or on an AWS instance. 
So, I have to remove these line and also remove tests for float16. If users try to run in fp16, they will get overflow as well. Do you have any suggestions? Thanks.

jerryzh168:
> Hi @jerryzh168 It's weird that if I add these lines [`f595ed4`/torchao/kernel/intmm.py#L142C1-L146C1](https://github.com/pytorch/ao/blob/f595ed41b99685cc16fc480ca2218965bb812bed/torchao/kernel/intmm.py#L142C1-L146C1) to avoid overflow of float16, there will be a failure in `test_spinquant.py`, but the test does not use float16 at all.

what is the test failure? is it possible to do the dtype conversion before calling `int_scaled_matmul`

> And there is another failure with CUDA nightly, but its log cannot be loaded. And these failures cannot be reproduced in my local environment or on an AWS instance. So, I have to remove these line and also remove tests for float16. If users try to run in fp16, they will get overflow as well. Do you have any suggestions? Thanks.

I just saw the error, it is talking about some triton related things, I'll rerun first just to confirm

Xia-Weiwen:
> what is the test failure? is it possible to do the dtype conversion before calling `int_scaled_matmul`, e.g. before `y_dot_scaled = int_scaled_matmul(tmp, w_vals_int8_t, x_scales.reshape(-1, 1))` in affine_quantized_tensor.py?

The error is results not all close. One element exceed the tolerance by a small amount. As for dtype conversion, I didn't make such changes in affine_quantized_tensor.py.


> ```
> 
>   E1018 18:45:40.023529 436 site-packages/torch/_inductor/runtime/triton_heuristics.py:475]     assert lhs.shape[1].value >= 32, "small blocks not supported!"
>   E1018 18:45:40.023529 436 site-packages/torch/_inductor/runtime/triton_heuristics.py:475] AssertionError: small blocks not supported!
> ```

Thanks for the info. Maybe it's about shapes in the test case? I will modify and try again.

Xia-Weiwen:
@jerryzh168 I tried to move the dtype conversion to affine_quantized_tensor.py and it worked. Now all checks pass. Thanks.

torchao/prototype/smoothquant/readme.md
jerryzh168, line 1: nit: readme.md --> README.md

test/prototype/test_smoothquant.py
```diff
@@ -0,0 +1,176 @@
+from copy import deepcopy
+import os
+import pytest
+import torch
+import tempfile
+from torch.ao.quantization import MovingAverageMinMaxObserver
```
jerryzh168, line 6: can you add a followup (e.g. open an issue to implement moving average observer in torchao?)

test/prototype/test_smoothquant.py
```diff
@@ -0,0 +1,176 @@
...
+
+        def forward(self, x):
+            return self.fc(x)
+
+    m = Linear(bias).eval().to(idtype).to(device)
+    m_ref = deepcopy(m)
+    data = torch.randn(2, 32, dtype=idtype, device=device)
+
+    # calibrate
+    insert_smooth_quant_observer(m, alpha, quant_mode, n_calib_examples=1)
```
jerryzh168, line 74: nit: would it be useful to have a default value for alpha as well?

torchao/prototype/smoothquant/readme.md
```diff
@@ -0,0 +1,89 @@
...
+```
+To load a recipe, users should insert observers first. For example,
+```python
+insert_smooth_quant_observer(model)
+load_smooth_quant_recipe(model, "./smooth_quant_recipe.json")
+```
+
+## Benchmark
+Running the example with `torch.compile` on a NVIDIA A10G GPU.
+### meta-llama/Llama-2-7b-hf
```
jerryzh168, line 67: can you also report the perplexity for dynamic and static quant without using smoothquant?

torchao/prototype/smoothquant/readme.md
```diff
@@ -0,0 +1,89 @@
...
+```
+And load it by `--model-load-path`
+```bash
+python example.py -m MODLE_ID --device=<cuda or cpu> --quant-mode=<dynamic or static> --model-load-path ./quantized_model.pt
+```
+
+
+## Usage of API
+The following APIs are provided:
+- insert_smooth_quant_observer
```
jerryzh168, line 29:
```suggestion
- insert_smooth_quant_observer_
```

to make it clear it's inplace operation

torchao/prototype/smoothquant/core.py
```diff
@@ -0,0 +1,167 @@
...
+        self.n_calib_examples = n_calib_examples
+        self.inputs = []
+        self.device = self.weight.device
+        self.alpha = alpha
+        assert quant_mode in ["static", "dynamic"]
+        self.quant_mode = quant_mode
+        # act.shape = [mb, ic] (reshape if needed), wei.shape = [oc, ic]
+        # *_ic_obs are used to determine smoothing_factor
+        # *_mb/oc_obs are used to find qparams for quantization
+        self.act_ic_obs = PerChannelMinMaxObserver(
```
jerryzh168, line 67: nit: I think PerChannelMinMaxObserver can be replaced by our torchao AQT observer, see https://github.com/pytorch/ao/blob/f33cff7053530506d11085eb18e6b93f5e224faf/tutorials/calibration_flow/static_quant.py#L168 (need to set ch_axis correctly)

torchao/prototype/smoothquant/core.py
```diff
@@ -0,0 +1,167 @@
...
+from torchao.quantization.quant_primitives import (
+    MappingType,
+    ZeroPointDomain,
+)
+from torchao.quantization.observer import (
+    AffineQuantizedObserverBase, PerRow
+)
+
+
+class SmoothQuantObserver(AffineQuantizedObserverBase):
```
jerryzh168, line 15: it seems that smooth quant observer is not related to AQT, probably don't need to inherit from `AffineQuantizedObserverBase`?

torchao/prototype/smoothquant/core.py
Xia-Weiwen: Thanks for reviewing. Just curious. Why are we using observers in Torchao instead of those in PyTorch?

jerryzh168: PyTorch observers are not general enough since they only support per tensor, per channel etc. since now we also have per token, per group etc., observers in torchao are more general since we have the block_size arg that can represent all granularities. also pytorch observer has some args we want to get rid of like custom_qmin/qmax

torchao/prototype/smoothquant/readme.md
Xia-Weiwen: Thanks. Will rename.

test/prototype/test_smoothquant.py
Xia-Weiwen: Thanks. `insert_smooth_quant_observer` has a default value for alpha, which is 0.5. Here we test different alphas so it is given explicitly.

torchao/prototype/smoothquant/readme.md
Xia-Weiwen: Thanks. Will refactor.

torchao/prototype/smoothquant/core.py
```diff
@@ -0,0 +1,167 @@
...
+            eps=eps,
+        )
+        self.wei_ic_obs(self.weight)
+
+    @torch.no_grad()
+    def forward(self, input: torch.Tensor):
+        if self.quant_mode == "static":
+          # record inputs to find qparams for activation
+          if len(self.inputs) < self.n_calib_examples:
+              self.inputs.append(input.to("cpu").view(-1, input.size(-1)))
```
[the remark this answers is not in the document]
jgong5: Sounds more consistent to use `MultiTensor` here? Need to move `multi_tensor.py` to a common place too?

jgong5 (approved):
Overall LGTM.

torchao/prototype/smoothquant/core.py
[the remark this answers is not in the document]
Xia-Weiwen: Thanks. I am considering removing this part as we only need to keep track of the min/max values instead of entire tensors.

test/prototype/test_smoothquant.py
jerryzh168: oh OK sounds good

torchao/prototype/smoothquant/core.py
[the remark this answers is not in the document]
jerryzh168: yeah we can move that to utils if you need it, let me know @Xia-Weiwen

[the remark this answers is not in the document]
jerryzh168: actually I think it has some logic specific to autoround, maybe you can just copy paste for now

leslie-fang-intel (approved):

torchao/prototype/smoothquant/readme.md
Xia-Weiwen: Sure. I have collected more data and provided in the README

test/prototype/test_smoothquant.py
Xia-Weiwen: Now it is not used in the implementation although using a MovingAverageMinMaxObserver might produce better accuracy. Do we still need a followup?

torchao/prototype/smoothquant/core.py
Xia-Weiwen: Thanks. It's changed.

torchao/prototype/smoothquant/core.py
Xia-Weiwen: Thanks. I have refactored the observer part.

torchao/prototype/smoothquant/core.py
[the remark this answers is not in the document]
Xia-Weiwen: I have refactored this part and it's no longer needed to keep these input tensors.

torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -1369,7 +1368,10 @@ def _linear_int8_act_int8_weight_impl(input_tensor, weight_tensor, bias):
     w_vals_int8_t = weight_tensor.tensor_impl.int_data.contiguous().t()
     w_scales = weight_tensor.tensor_impl.scale
     tmp = x_vals_int8.reshape(-1, x_vals_int8.shape[-1])
-    y_dot_scaled = int_scaled_matmul(tmp, w_vals_int8_t, x_scales.reshape(-1, 1))
+    x_scales_dtype = x_scales.dtype
+    intermediate_dtype = torch.float if x_scales_dtype == torch.half else x_scales_dtype
```
jerryzh168, line 1372: can you add a comment on how do we decide the intermediate dtype here?

torchao/prototype/smoothquant/README.md
```diff
@@ -0,0 +1,98 @@
...
+| Static | 43.8051 | 11.2984 | 7.5791 | 19.5050 |
+
+Note*: Conventional quantization without SmoothQuant
+
+### meta-llama/Meta-Llama-3-8B
+Perplexity
+| Quant Method | alpha=0.25 | alpha=0.5 | alpha=0.75 | alpha=None* |
+|-|-|-|-|-|
+| Dynamic | 21.2475 | 8.8288 | 9.6514 | 8.3574 |
+| Static | 301.7118 | 18.0617 | 10.8343 | 278.9819 |
```
jerryzh168, line 81: so looks like it's more effective on static quant

jerryzh168 (approved):
looks good, thanks!

torchao/prototype/smoothquant/README.md
jerryzh168: did you also do a sanity check for perf to make sure this doesn't regress performance?

test/prototype/test_smoothquant.py
jerryzh168: is the change significant? if so maybe worth follow up

torchao/dtypes/affine_quantized_tensor.py
Xia-Weiwen: Sure. Added.

torchao/prototype/smoothquant/README.md
Xia-Weiwen: For performance, it's found from high to low
- Static quant
- Static quant + SmoothQuant
- Dynamic quant
- Dynamic quant + SmoothQuant

It's expected that SmoothQuant is slower because it inserts `div` on graph. Is it Ok?

Xia-Weiwen:
Hi @jerryzh168 I have updated this PR. Please take a look again. Thanks

test/prototype/test_smoothquant.py
Xia-Weiwen: Ok. I will open an issue then.

Xia-Weiwen:
BTW, I found torchao's observer behaves differently from pytorch's observer when running on cuda. torchao's observer has its `self.min_val` and `self.max_val` on the same device as input tensor but pytorch's observer always has them on cpu. Is that something that needs a fix? Thanks.

jerryzh168:
> BTW, I found torchao's observer behaves differently from pytorch's observer when running on cuda. torchao's observer has its `self.min_val` and `self.max_val` on the same device as input tensor but pytorch's observer always has them on cpu. Is that something that needs a fix? Thanks.

I see, I feel min_val/max_val being in the same device as input makes more sense? or are you saying we should add an option here?

torchao/prototype/smoothquant/README.md
jerryzh168: yeah that's fine as long as it's reasonable, it's just a sanity check

jerryzh168 (approved):

Xia-Weiwen:
> > BTW, I found torchao's observer behaves differently from pytorch's observer when running on cuda. torchao's observer has its `self.min_val` and `self.max_val` on the same device as input tensor but pytorch's observer always has them on cpu. Is that something that needs a fix? Thanks.
> 
> I see, I feel min_val/max_val being in the same device as input makes more sense? or are you saying we should add an option here?

Oh, I thought you might want them to have the same behavior. It's alright if that is not an issue.

PR #1034, opened 2024-10-08
gemlite integration in torchao

HDCharles:
gemlite integration in torchao

Summary:

This PR adds support for gemlite kernels in torchao using a subclass
integration with the gemlite_uintx_weight_only constructor. This works
for int4 grouped and ungrouped assymmetric oeight only quantization and
int8 symmetric ungrouped quantization for fp16 models. TP support
through DTensor is included in thsi PR

in the process of integrating gemlite into AQT i also made some fixes to
a few quant primitives that are being used which previously were not.

Test Plan:

test_integration.py -k "test_gemlite_layout"
test_affine_quantized_tensor_parallel.py -k "test_tp_gemlite"

see benchmarks.sh for gemlite benchmarks as well.

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/_models/llama/generate.py
+++ b/torchao/_models/llama/generate.py
@@ -171,7 +171,8 @@ def decode_n_tokens(
             )
             next_token, next_prob = next_token.clone(), next_prob.clone()
             input_pos += 1
-            new_tokens.append(next_token)
+            # in some instances not having this causes weird issues with the stored tokens when you run the next decode_one_token step
+            new_tokens.append(next_token.clone()) 
             callback(new_tokens[-1])
             new_probs.append(next_prob)
             cur_token = next_token
@@ -368,6 +369,7 @@ def main(
             int8_weight_only,
             quantize_,
             uintx_weight_only,
+            gemlite_uintx_weight_only,
         )
 
         from torchao.quantization.granularity import PerRow, PerTensor
@@ -377,6 +379,39 @@ def main(
             from torchao.prototype.spinquant import apply_spinquant
 
             apply_spinquant(model)
+        if "gemlite" in quantization:
+            import os, pwd
+            import gemlite
+            from gemlite.core import GemLiteLinearTriton, set_autotune
+            _quant_args = quantization.split("-")
+            bit_width = int(_quant_args[-2])
+            group_size = None if _quant_args[-1] == 'None' else int(_quant_args[-1])
+            try:
+                packing_bitwidth = int(_quant_args[-3])
+            except:
+                # if only 2 inputs found, use default value
+                packing_bitwidth = 32
+
+            quantize_(model, gemlite_uintx_weight_only(group_size, bit_width, packing_bitwidth))
+
+            # try to load gemlite kernel config
+            try:
+                GemLiteLinearTriton.load_config(f"/tmp/{pwd.getpwuid(os.getuid()).pw_gecos}_gemlite.json")
+                print(f"loaded gemlite kernel cache /tmp/{pwd.getpwuid(os.getuid()).pw_gecos}_gemlite.json")
+            except:
+                print(f"unable to load gemlite kernel cache /tmp/{pwd.getpwuid(os.getuid()).pw_gecos}_gemlite.json")
+
+            print("running gemlite warmup")
+            generate(
+                model,
+                encode_tokens(tokenizer, prompt, bos=True, device=device),
+                max_new_tokens,
+                batch_size,
+                interactive=False,
+                temperature=temperature,
+                top_k=top_k,
+            )
+            GemLiteLinearTriton.cache_config(f"/tmp/{pwd.getpwuid(os.getuid()).pw_gecos}_gemlite.json")
         if "int8wo" in quantization:
             quantize_(model, int8_weight_only())
         if "int8dq" in quantization:
@@ -959,7 +994,7 @@ if __name__ == "__main__":
 
     parser = argparse.ArgumentParser(description="Your CLI description.")
     parser.add_argument(
-        "--prefill_size", type=int, default=0, help="Whether to run in ttft mode"
+        "--prefill_size", type=int, default=None, help="Whether to run in ttft mode"
     )
     parser.add_argument(
         "--prompt", type=str, default="Hello, my name is", help="Input prompt."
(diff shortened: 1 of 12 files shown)


mobicham:
I see that you removed the pruning config, that will produce incorrect results for group-sizes lower than 128

HDCharles:
its there now, needed the custom op support stuff

torchao/_models/llama/generate.py
```diff
@@ -1053,6 +1088,7 @@ def callback(x):
     )
 
     args = parser.parse_args()
+    print(args)
```
jerryzh168, line 1091: nit: remove

torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -251,7 +261,8 @@ def from_hp_to_intx(
                 zero_point_domain,
             )
             # choose_qparams_affine is a custom op that does support returning optional Tensors. We thus set the zero_point to None if its domain is None
-            if zero_point_domain is None:
+            # TODO should probably consolidate ZeroPointDomain.NONE and None
```
jerryzh168, line 264: yeah we should not have both, also doing some refactor in https://github.com/pytorch/ao/pull/1402/files#diff-7c9b4c8c6d4ef9c47873263304a335d5cf56c3ac9f98ba10b994cd80dc9c2709L536

torchao/dtypes/affine_quantized_tensor_ops.py
```diff
@@ -20,6 +20,10 @@
     _linear_int8_act_int8_weight_block_sparse_check,
     _linear_int8_act_int8_weight_block_sparse_impl,
 )
+from torchao.dtypes.uintx.gemlite_layout import (
+    _linear_fp_act_int4_weight_gemlite_check,
```
jerryzh168, line 24: nit: fp16? seems like gemlite only works with fp16 right now, but can be a follow up PR

torchao/dtypes/uintx/gemlite_layout.py
```diff
@@ -0,0 +1,372 @@
...
+        _layout: Layout,
+    ):
+        from gemlite.core import DType, GemLiteLinearTriton, set_autotune
+
+        assert isinstance(
+            _layout, GemlitePackedLayout
+        ), f"GemliteAQTTensorImpl only works with GemliteLinearTriton but got {_layout}"
+        group_size, bit_width = _layout.group_size, _layout.bit_width
+
+        torch._dynamo.config.inline_inbuilt_nn_modules = False
```
jerryzh168, line 170: nit: do we need to restore the state afterwords

torchao/dtypes/uintx/gemlite_layout.py
```diff
@@ -0,0 +1,372 @@
...
+    def __torch_dispatch__(cls, func, types, args, kwargs):
+        kwargs = {} if kwargs is None else kwargs
+
+        # we don't handle transpose operations and just ignore them. In practice the only
+        # reason a transpsoe should occur is because the functional linear
+        # op can decompose into e.g. transpose + addmm so since we want
+        # to use the gemlite matmul kernel, which expects teh weight to be passed in as is,
+        # we ignore the transpose
+        if func is aten.detach.default or func is aten.t.default:
+            return return_and_correct_aliasing(
```
jerryzh168, line 271: would be good to record the transposed state probably: https://github.com/pytorch/ao/blob/200589beb900d4a39a880d4e31b942bed9d31f12/torchao/dtypes/uintx/tensor_core_tiled_layout.py#L306 in case it's used later

torchao/dtypes/uintx/gemlite_layout.py
```diff
@@ -0,0 +1,372 @@
...
+            weight_impl.zero_point,
+        ),
+    )
+
+
+def _linear_fp_act_int4_weight_gemlite_check(input_tensor, weight_tensor, bias):
+    return (
+        # input is native fp16 tensor
+        not is_traceable_wrapper_subclass(input_tensor)
+        # and input_tensor.dtype == torch.float16
```
jerryzh168, line 368: don't we need this?

jerryzh168 (approved):
looks good, I think we should merge now to unblock benchmarking and fix the follow up a bit later

torchao/quantization/quant_api.py
```diff
@@ -629,6 +630,33 @@ def int8_dynamic_activation_int4_weight(
...
+         size is more fine grained
+        `bit_width`: bit width of the quantized weight.
+        `packing_bitwidth`: bit width of the packed weight, should be 8 or 32. Can have performance impacts depending on hardware.
+        `contiguous`: if set, the weight will be packed as specified. Leaving it as None lets gemlite determine the best choice.
+    """
+
+    from torchao.dtypes.uintx.gemlite_layout import apply_gemlite_quant
+
+    use_hqq = True if bit_width == 4 else False
+    apply_fn = lambda weight: apply_gemlite_quant(
```
drisspg, line 654: Should we raise if gemlite is not installed?

jerryzh168: I think so, please feel free to review more, and we should have follow up fixes. 

just want to merge now to unblock benchmarking in sglang for now

torchao/dtypes/uintx/gemlite_layout.py
HDCharles: i think this is necessary for gemlite to function correctly,

torchao/dtypes/uintx/gemlite_layout.py
HDCharles: i think this may be better to be something handled by AQT itself rather than the tensor impl, in reality we don't want transpose causing packing and repacking of the weight, thus it should be tracked by the top level where the shape is actually being changed.

torchao/dtypes/uintx/gemlite_layout.py
HDCharles: its a little redundant since we block creation of the subclass unpon creation, normally when you mess this type of thing up it gives you a dtype mismatch error which is what will happen now, whereas adding this would likely bypass the error which i don't know if users would actually want.

could see it being either way though.

PR #1038, opened 2024-10-08
Move and rename GranularityType -> Granularity

andrewor14:
* #1020
* __->__ #1038

Summary: Move GranularityType to quant_primitives.py to be
consistent with other similar fields like MappingType and
ZeroPointDomain.

Test Plan: CI


torchao/quantization/quant_primitives.py
```diff
@@ -64,6 +65,74 @@ class ZeroPointDomain(Enum):
     INT = auto()
     FLOAT = auto()
 
+@dataclass(frozen=True)
+class Granularity:
```
jerryzh168, line 69: should this be put into its own file

jerryzh168 (approved):

torchao/quantization/quant_primitives.py
andrewor14: sure

PR #1043, opened 2024-10-09
Enable AWQ example on CPU

Xia-Weiwen:
Replace hardcoded `cuda` in the code with `args.device`.


vayuda (approved):
Good suggestion. Thought CPU was too slow so I didn't think about it.

Xia-Weiwen:
@vayuda Thanks for reviewing and merging. We also need AWQ on Intel CPUs.

PR #1072, opened 2024-10-14
Create header for packed weight ops

metascroy:
Summary:
This diff defines a packed_weight hader in torchao/experimental/ops/packed_weights_header.h.

The header is 16 bytes and has 4 fields:
* format: PackedWeightsFormat (enum)
* extra0: int
* extra1: int
* extra2: int

Whenever we have a new format type, we can add a field to the enum.  Currently I have a field for the format the universal kernels use, but MPS can have a different format, and KleidiAI also has its own format.

I modified the pack ops put this header and the start of the packed weights.  When the linear op runs, it reads the header to understand how the weights were packed.

Reviewed By: digantdesai

Differential Revision: D63498956


digantdesai:
Couple of things
* update the summary
* add a test to ensure we don't break BC accidently

digantdesai (approved):

Issue #998, opened 2024-10-03
Make Quant-LLM compatible with BF16


gau-nernst:
Quant-LLM code: https://github.com/pytorch/ao/tree/main/torchao/csrc/cuda/fp6_llm

Currently Quant-LLM kernel (backing FPx in torchao) only works with FP16. This creates a small divergence from other quantization methods, which all work with BF16. Since all recent models are trained and released with BF16, having BF16 support potentially improve accuracy for FPx models.

Might be over-simplifying, but I think it's just the matter of modifying dequant logic and MMA instructions (as well as update dtype in function signature appropriately)

https://github.com/pytorch/ao/blob/09b8b3c3156b99fa8356e2df30c1a65cbbf5cc91/torchao/csrc/cuda/fp6_llm/utils_parallel_dequant.cuh#L30-L45

https://github.com/pytorch/ao/blob/09b8b3c3156b99fa8356e2df30c1a65cbbf5cc91/torchao/csrc/cuda/fp6_llm/ptx_mma.cuh#L117-L125

cc @msaroufim @HDCharles 

I might try to do it myself, but I think it would be an interesting good first issue task too. @tobiasvanderwerff Would you be interested?

tobiasvanderwerff:
I'd love to work on this @gau-nernst :)

So if you're ok with me doing this I'll probably get started on it in the next few days.

DevyRuxpin:
Please assign, I may have a solution for this.

tobiasvanderwerff:
Hi @DevyRuxpin, I can understand that you want to work on this, but I have already made most of the necessary changes for this feature (see the diff [here](https://github.com/pytorch/ao/compare/main...tobiasvanderwerff:ao:fp6-bf16)). Since your solution is not complete yet (it removes FP16 support), it's probably best to avoid doing further duplicate work.

PR #1074, opened 2024-10-14
Issue #998 Fix

DevyRuxpin:
Possible solution to issue #998

### DIFF
--- a/torchao/csrc/cuda/fp6_llm/ptx_mma.cuh
+++ b/torchao/csrc/cuda/fp6_llm/ptx_mma.cuh
@@ -34,7 +34,7 @@
 #define PTX_MMA_CUH
 
 #include <cuda.h>
-#include <cuda_fp16.h>
+#include <cuda_bf16.h>  // Include BF16 header
 #include <cuda_runtime.h>
 
 #include <assert.h>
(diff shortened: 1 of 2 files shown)


msaroufim:
Might need to also add some tests so CI signal is helpful and just to be clear are we only supporting bf16 moving forward or both fp16 and bf16?

gau-nernst:
I think we shouldn't remove FP16 support, since it is the original dtype that Quant-LLM support. Not sure how we can structure it nicely (i.e. support both FP16 and BF16), but you can come up with a way first. Ideally it shouldn't have intrusive changes to the original code, so if there are upstream updates, it's easier for us to pull the new updates in. Maybe some kind of templates or just have a separate folder for BF16?

DevyRuxpin:
I will work on this further. Thank you for feedback, i will draw things back some on the changes and incorporate your suggestions. i will work on this tonight/tomorrow.

tobiasvanderwerff:
@gau-nernst @msaroufim I have already been working on a templated solution that supports both FP16 and BF16, and it is almost finished (see the diff [here](https://github.com/pytorch/ao/compare/main...tobiasvanderwerff:ao:fp6-bf16)). I think this would be a better solution than removing FP16 all together.

I'm just smoothing out some final bugs and it should be as good as ready. I didn't make a PR yet because I wanted to make sure that everything worked properly first. In hindsight, it might have been better to already create a PR to avoid the duplicate work done here (although I did say I was working on this feature in the original issue).

Issue #1063, opened 2024-10-12
CDN issue when building linux wheels


msaroufim:
Since Oct 7 our CI jobs on the main branch have been reliably red because everytime we download pytorch we run into this issue

This is a CDN issue that is out of control but escalated this interally

```
+ conda run -p /__w/_temp/conda_environment_11300887954 pip install torch --pre --index-url https://d3kup0pazkvub8.cloudfront.net/whl/nightly/cpu
ERROR conda.cli.main_run:execute(41): `conda run pip install torch --pre --index-url [https://d3kup0pazkvub8.cloudfront.net/whl/nightly/cpu`](https://d3kup0pazkvub8.cloudfront.net/whl/nightly/cpu%60) failed. (See above for error)
WARNING: The directory '/github/home/.cache/pip' or its parent directory is not owned or is not writable by the current user. The cache has been disabled. Check the permissions and owner of that directory. If executing pip with sudo, you should use sudo's -H flag.
ERROR: Wheel 'torch' located at /tmp/pip-unpack-nzix8h63/torch-2.6.0.dev20241011_cpu-cp310-cp310-linux_x86_64.whl is invalid.

Looking in indexes: https://d3kup0pazkvub8.cloudfront.net/whl/nightly/cpu
Collecting torch
  Downloading https://d3kup0pazkvub8.cloudfront.net/whl/nightly/cpu/torch-2.6.0.dev20241011%2Bcpu-cp310-cp310-linux_x86_64.whl (175.0 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━          134.2/175.0 MB 62.5 MB/s eta 0:00:01
```

atalman:
The PR that deployed Meta CDN is reverted: https://github.com/pytorch/test-infra/pull/5762

Issue #1062, opened 2024-10-11
Unpin PT nightly for CPU


msaroufim:
See https://github.com/pytorch/pytorch/issues/137826 for more context

PR #1075, opened 2024-10-14
Revert "Pin pt nightly CPU version"

HDCharles:
Reverts pytorch/ao#1061

seems this caused an issue

relevant to:
https://github.com/pytorch/ao/issues/1063
https://github.com/pytorch/ao/issues/1062

hopefully https://github.com/pytorch/ao/pull/1061 does not resurface and it was an env issue

### DIFF
--- a/.github/workflows/regression_test.yml
+++ b/.github/workflows/regression_test.yml
@@ -81,5 +81,6 @@ jobs:
         pip install ${{ matrix.torch-spec }}
         pip install -r dev-requirements.txt
         pip install .
+        export CONDA=$(dirname $(dirname $(which conda)))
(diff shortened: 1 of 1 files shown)


msaroufim (approved):
assuming nightly cpu job is green

EDIT: interesting yeah the error is reproducing https://github.com/pytorch/pytorch/issues/137826

cc @zailiwang

gau-nernst:
It seems like related to this https://stackoverflow.com/questions/49875588/importerror-lib64-libstdc-so-6-version-cxxabi-1-3-9-not-found

So something along the lines of a compiler (from CUDA) was used to compile the codegen CPU code but a different C++ stdlib (from system) was used to run the compiled code? Kinda strange that it didn't surface for CUDA nightly job.

msaroufim:
SUCCESS!

PR #1078, opened 2024-10-15
Add AQT tensor parallel for float8_dynamic_quant

jainapurva:
Added support for tensor parallel for float8_dynamic_activation_float8_weight 
- [x] - Supports PerTensor scaling
- [x] - Supports PerRow scaling

Added op implementations:
- aten.slice.Tensor
- aten.view.default

### DIFF
--- a/torchao/dtypes/affine_quantized_tensor.py
+++ b/torchao/dtypes/affine_quantized_tensor.py
@@ -1107,12 +1107,22 @@ class Float8AQTTensorImpl(AQTTensorImpl):
         elif func is aten.slice.Tensor:
             self, dim, start, end, step = fill_defaults(args, 5, [0, None, None, 1])
             if dim == 0:
-                return return_and_correct_aliasing(
-                    func, args, kwargs, args[0]._apply_fn_to_data(lambda x: aten.slice.Tensor(x, dim, start, end, step))
-                )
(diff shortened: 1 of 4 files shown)


torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -1107,12 +1110,19 @@ def __torch_dispatch__(cls, func, types, args, kwargs):
         elif func is aten.slice.Tensor:
             self, dim, start, end, step = fill_defaults(args, 5, [0, None, None, 1])
             if dim == 0:
+                #TODO: scale replecation should be dependent on block size
+                if self.scale.ndim == 1:
+                    return return_and_correct_aliasing(
+                        func, args, kwargs, args[0]._apply_fn_to_data(lambda x: aten.slice.Tensor(x, dim, start, end, step))
+                    )
+                else:
```
jerryzh168, line 1118: what is this case btw, is this ndim==0 or some other case?

jainapurva: Yes, it's ndim==0. I've updated the condition to check for ndim=0, instead of generic condition.

jerryzh168 (approved):
looks good, thanks!

PR #1079, opened 2024-10-15
Header bug fix

metascroy:
Summary: A last minute change created a compile error on the header.  This fixes the issue.

Differential Revision: D64370707

### DIFF
--- a/torchao/experimental/ops/linear_8bit_act_xbit_weight/packed_weights_header.h
+++ b/torchao/experimental/ops/linear_8bit_act_xbit_weight/packed_weights_header.h
@@ -17,19 +17,18 @@ torchao::ops::PackedWeightsHeader get_packed_weights_header_universal(
     int nr,
     int kr,
     int version = 1) {
-  TORCHAO_CHECK(
-      version >= 0 && version < 256, "version must be between 0 and 255");
-  TORCHAO_CHECK(
-      weight_nbit >= 1 && weight_nbit < 256,
-      "weight_nbit must be between 1 and 255");
   return torchao::ops::PackedWeightsHeader(
       torchao::ops::PackedWeightsFormat::linear_8bit_act_xbit_weight_universal,
-      {((static_cast<unsigned short>(version) << 8) |
-        static_cast<unsigned short>(weight_nbit)),
-       ((static_cast<unsigned short>(has_weight_zeros) << 8) |
-        static_cast<unsigned short>(has_bias)),
-       static_cast<unsigned short>(nr),
-       static_cast<unsigned short>(kr),
+      {version,
+       weight_nbit,
+       has_weight_zeros,
+       has_bias,
+       nr,
+       kr,
+       0,
+       0,
+       0,
+       0,
        0,
        0,
        0,
--- a/torchao/experimental/ops/packed_weights_header.h
+++ b/torchao/experimental/ops/packed_weights_header.h
@@ -7,17 +7,20 @@
 #pragma once
 #include <array>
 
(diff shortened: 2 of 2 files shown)


malfet:
I'm unsure what it fixes to be frank, perhaps better PR description would help clarify what it is trying to fix...

torchao/experimental/ops/linear_8bit_act_xbit_weight/packed_weights_header.h
```diff
@@ -17,19 +17,18 @@ torchao::ops::PackedWeightsHeader get_packed_weights_header_universal(
     int nr,
     int kr,
     int version = 1) {
-  TORCHAO_CHECK(
-      version >= 0 && version < 256, "version must be between 0 and 255");
```
malfet, line 21: What is wrong with this check?

torchao/experimental/ops/packed_weights_header.h
```diff
@@ -10,44 +10,47 @@
 #include <cassert>
 namespace torchao::ops {
 
-enum PackedWeightsFormat : unsigned short {
+enum PackedWeightsFormat : int {
```
malfet, line 13: Why this must be a signed type? `unsigned int` sounds like a better idea. Also `enum` inside the namespace is oxymoron, please use `enum class`
```suggestion
enum class PackedWeightsFormat : uint32_t {
```

torchao/experimental/ops/packed_weights_header.h
```diff
@@ -10,44 +10,47 @@
...
       PackedWeightsFormat format = PackedWeightsFormat::unknown,
-      params_type params = {0, 0, 0, 0, 0, 0, 0})
+      params_type params = {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0})
       : format{format}, params{params} {}
 
   inline static constexpr int size() {
-    static_assert(sizeof(format) + sizeof(params) == 16);
-    return 16;
+    static_assert(sizeof(magic) + sizeof(format) + sizeof(params) == 64);
+    return 64;
```
malfet, line 34: Why not use sizeof(XYZ)?
```suggestion
    return sizeof(PackedWeightsHeader);
```

torchao/experimental/ops/packed_weights_header.h
```diff
@@ -10,44 +10,47 @@
...
   }
 
   inline void write(void* packed_weights) const {
-    auto header = (unsigned short*)(packed_weights);
-    header[0] = (unsigned short)format;
+    auto header = (int*)(packed_weights);
```
malfet, line 38: In C++ better used C++ cast-style ops
```suggestion
    auto header = reinterpret_cast<int*>(packed_weights);
```

torchao/experimental/ops/linear_8bit_act_xbit_weight/packed_weights_header.h
metascroy: Before it existed to make sure version could be packed into a char (since both version/weight_nbit were packed into an unsigned short).  But the header now has more space, and version is now an int, so this bound check is no longer needed.

torchao/experimental/ops/packed_weights_header.h
metascroy: I thought this made it clearer that the header is 64 bytes at a glance.

digantdesai (approved):

digantdesai:
Address Nikita's comments before landing, stamping to unblock if you have already done so.

PR #1085, opened 2024-10-15
Add generic fake quantized embedding for QAT

andrewor14:
**Summary:** This is equivalent to https://github.com/pytorch/ao/pull/1020 but for nn.Embedding. This commit adds a generic fake quantized embedding module to replace the uses of the existing more specific QAT embeddings. For example, `Int4WeightOnlyQATEmbedding` can be expressed as follows:

```
from torchao.quantization.prototype.qat.api import FakeQuantizeConfig
from torchao.quantization.prototype.qat.embedding import FakeQuantizedEmbedding

weight_config = FakeQuantizeConfig(
    dtype=torch.int4,
    group_size=group_size,
    is_symmetric=True,
)
fq_embedding = FakeQuantizedEmbedding(16, 32, weight_config=weight_config)
```

**Test Plan:**
python test/quantization/test_qat.py -k test_qat_4w_embedding python test/quantization/test_qat.py -k test_fake_quantized_embedding_4w


jerryzh168 (approved):

torchao/quantization/prototype/qat/embedding.py
```diff
@@ -245,7 +306,7 @@ def __init__(
 
     def forward(self, x):
         from torchao._executorch_ops import _quantized_decomposed_dequantize_per_channel_group_wrapper
-        qmin, qmax = _get_qmin_qmax(self.bit_width)
+        qmin, qmax = _get_qmin_qmax(4)
```
jerryzh168, line 309: why is this hardcoded now?

torchao/quantization/prototype/qat/embedding.py
```diff
@@ -98,28 +160,31 @@ def _convert_helper(self, module: torch.nn.Module):
...
+                    scale_precision=scale_precision,
+                    zero_point_precision=zero_point_precision,
                     device=child.weight.device,
                 )
                 setattr(module, name, quantized_embedding)
 
                 # Load weights and qparams into quantized embedding
-                (qmin, qmax) = _get_qmin_qmax(self.bit_width)
-                (s, zp) = get_group_qparams_symmetric(child.weight, self.bit_width, child.group_size)
+                (qmin, qmax) = _get_qmin_qmax(4)
```
jerryzh168, line 184: I feel it might be better to keep `4` as an constant defined in the class so we can generalize later and we'll know more semantic information here for the arg

Issue #1013, opened 2024-10-04
ensure that there is only one "prototype" folder in torchao


vkuzo:
# context

Today `torchao` has various prototype folders:
1. https://github.com/pytorch/ao/tree/main/torchao/prototype
2. https://github.com/pytorch/ao/tree/main/torchao/quantization/prototype
3. https://github.com/pytorch/ao/tree/main/torchao/sparsity/prototype

# the task

This is confusing.  Let's just converge on one place, `torchao/prototype`.  This task is to move the code over without breaking things:
A. chat with @andrewor14  and confirm if it's ok to move `torchao/quantization/prototype` to `torchao/prototype/quantization`.  If yes, move it.
B. chat with @jcaip and confirm if it's ok to move `torchao/sparsity/prototype` to `torchao/prototype/sparsity`.  If yes, move it.

When moving the code, it would be good to verify that scripts/features/tests still work, CI is green, and update any hardcoded locations as needed.

Note: feel free to ignore `torchao/experimental` for now.

vkuzo:
cc @drisspg

PR #1088, opened 2024-10-16
Move quantization/prototype -> prototype/quantization

jainapurva:
Issue : https://github.com/pytorch/ao/issues/1013


jerryzh168:
thanks, this will be bc-breaking for QAT I think, will be good to import the diff and fix the torchtune stuff before landing I feel

vkuzo:
> thanks, this will be bc-breaking for QAT I think, will be good to import the diff and fix the torchtune stuff before landing I feel

If some of this code has internal callsites, should it move out of prototype?  Prototype shouldn't have any BC guarantees. cc @andrewor14

andrewor14:
Yeah sounds good, I will submit a separate PR to move QAT out of prototype and update all internal imports. @jainapurva for now can you leave the qat/ directory in its original place so you don't run into conflicts later? You can move the rest

andrewor14:
https://github.com/pytorch/ao/pull/1091

PR #1097, opened 2024-10-16
float8 training: add llama linear subgraph microbenchmarks

vkuzo:
Summary:

At a high level, this PR adds 4 microbenchmarks of `prev_op -> linear -> next_op` relevant to the LLaMa model family.  The goal is to benchmark and optimize float8 torch.compile overhead on these, with the assumption that speeding up these benchmarks will correspond to speeding up transformer workloads.

Specifically:
1. we add the definitions of relevant `prev_op -> linear -> next_op` subgraphs to `utils.py`
2. we make `profile_linear_float8.py` use the newly added subgraphs for individual profiling, configurable by subgraph name
3. we revamp `bench_linear_float8.py` to bring it up to speed with the latest best practices (too many changes to spell out), and also add the option to use these new subgraphs configurable by subgraph name

After this PR:
* there is a unified definition of relevant subgraphs
* `profile_linear_float8.py` is a convenience script to dive deeper into float8+compile on a single subgraph given an input shape
* `bench_linear_float8.py` is a convenience script to sweep performance of float8+compile on a single subgraph over a set of input shapes

Test Plan:

Ran both of the scripts and data looks right, and results for subgraphs not touched in this PR did not change.

Will follow up further (not in this PR) with any insights gained / next steps.

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/benchmarks/float8/bench_linear_float8.py
+++ b/benchmarks/float8/bench_linear_float8.py
@@ -6,44 +6,36 @@
 import argparse
 import copy
 from dataclasses import dataclass
-from itertools import product
 from pathlib import Path
 from typing import Callable, List, Optional, Tuple
 
 import pandas as pd
 
 import torch
+import torch.nn as nn
 import torch.utils.benchmark as benchmark
+from torchao.testing.float8.test_utils import get_test_float8_linear_config
+from torchao.float8 import convert_to_float8_training
 from torchao.float8.config import (
     CastConfig, 
     Float8LinearConfig, 
     ScalingType,
     ScalingGranularity,
+    Float8LinearRecipeName,
+    recipe_name_to_linear_config,
 )
 from torchao.float8.float8_linear import Float8Linear
 from torchao.float8.float8_linear_utils import (
     linear_requires_sync,
     sync_float8_amax_and_scale_history,
 )
-from torchao.float8.float8_tensor import ScaledMMConfig
-from utils import get_name_to_shapes_iter
+from utils import (
+    get_name_to_shapes_iter,
+    do_bench_using_profiling_wrapper_s,
+    subgraph_name_to_subgraph_and_inputs,
+)
 from tqdm import tqdm
 
-# estimating TOPs for matmuls in fp32, fp16, fp8
-# assuming A * B = C, with A being M * K, B being K * N, C being M * N
-
-# H100 SXM specs: bottom of https://www.nvidia.com/en-us/data-center/h100/
-h100_peak_flops_float32 = 67e12
-h100_peak_flops_fp16_tc = 1979e12
-h100_peak_tops_float8_tc = 3958e12
-
-dtype_to_peak_tops = {
-    torch.float32: h100_peak_flops_float32,
-    torch.float16: h100_peak_flops_fp16_tc,
-    torch.bfloat16: h100_peak_flops_fp16_tc,
-    torch.float8_e4m3fn: h100_peak_tops_float8_tc,
-    torch.float8_e5m2: h100_peak_tops_float8_tc,
-}
 
 # prevent splitting columns when printing a data frame
 pd.set_option("display.expand_frame_repr", False)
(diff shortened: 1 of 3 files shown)


benchmarks/float8/bench_linear_float8.py
```diff
@@ -6,44 +6,36 @@
...
-from torchao.float8.float8_tensor import ScaledMMConfig
-from utils import get_name_to_shapes_iter
+from utils import (
+    get_name_to_shapes_iter,
+    do_bench_using_profiling_wrapper_s,
+    subgraph_name_to_subgraph_and_inputs,
+)
 from tqdm import tqdm
 
-# estimating TOPs for matmuls in fp32, fp16, fp8
```
vkuzo, line 32: this shouldn't be here, this is more important for gemm-only benchmarks

benchmarks/float8/bench_linear_float8.py
```diff
@@ -102,7 +72,6 @@ def main(
     sweep_path: Optional[Path] = None,
     compile: bool = True,
     n_limit: Optional[int] = None,
-    fast_accum_filter: Optional[bool] = None,
```
vkuzo, line 105: we now use the values from `Float8LinearConfig` instead

benchmarks/float8/utils.py
```diff
@@ -350,3 +336,326 @@ def get_gpu_kernel_gemm_time_s(f, *args, **kwargs):
...
+# The goal here is to set up microbenchmarks on the fusion of float8 
+# scaling/casting that are relevant to transformer models, and break
+# them up in such a way that we can easily reason about fusions regarding
+# each linear layer without having to dig through large traces / long log files.
+# There is some duplication here (attn_norm and ffn_norm appear in multiple
+# places), and that's ok.
+#
+# Simplifying assumptions (may be relaxed in the future):
+# 1. don't model fusion with rotary embeddings and repeat-transpose
+# 2. don't model fusion with SDPA
```
vkuzo, line 465: cc @drisspg , curious on your thoughts on how future-proof this assumption is

benchmarks/float8/bench_linear_float8.py
```diff
@@ -112,137 +81,93 @@ def main(
...
-            linear_float8.forward_config = ScaledMMConfig(False, True, False)
-        else:
-            linear_float8.forward_config = ScaledMMConfig(False, False, False)
+        m_float8 = copy.deepcopy(m_ref)
+        convert_to_float8_training(m_float8, config=config)
 
-        input_tensor = torch.randn(M, K, device=device, dtype=dtype, requires_grad=True)
-        ref_forw_backward = lambda: linear_ref(input_tensor).sum().backward()
+        def ref_forw_backward(*input_tensors):
+            return m_ref(*input_tensors).sum().backward()
```
drisspg, line 132: The sum_backwards can have some interesting side effects because the back propping grad is broadcasted, maybe: torch.empty_like(out)

benchmarks/float8/bench_linear_float8.py
```diff
@@ -112,137 +81,93 @@ def main(
...
+            return m_float8(*input_tensors).sum().backward()
 
         if compile:
             ref_forw_backward = torch.compile(ref_forw_backward)
             float8_forw_backward = torch.compile(float8_forw_backward)
 
-        for _ in range(5):
-            ref_forw_backward()
-            float8_forw_backward()
+        n_warmup = 2
```
drisspg, line 143: nit: maybe move else if you think people will change warmup steps, or just write
# Warmups

benchmarks/float8/utils.py
drisspg: mmmmm I think this is okay, the haddamard transform from FA3 is something we would expect to be fused but I think we can just leave a TODO until we have low-p fused attention training support

benchmarks/float8/utils.py
```diff
@@ -350,3 +336,326 @@ def get_gpu_kernel_gemm_time_s(f, *args, **kwargs):
...
+    LN_LINEAR = "ln_linear"
+    NORM_FFN_NORM = "norm_ffn_norm"
+    NORM_FFN_NORM_SMALL = "norm_ffn_norm_small"
+    ATTN_WQKV_SUBGRAPH = "attn_wqkv_subgraph"
+    ATTN_WO_SUBGRAPH = "attn_wo_subgraph"
+    FFN_W13_SUBGRAPH = "ffn_w13_subgraph"
+    FFN_W2_SUBGRAPH = "ffn_w12_subgraph"
+
+
+def subgraph_name_to_subgraph_and_inputs(
```
drisspg, line 547: I think it would be nice if this was based off this had a few model sizes stamped out similar to 
https://github.com/pytorch-labs/ao_benchmarks/blob/8c5c95541198c1da2042d7028cab707e6541bb8a/experimental/smoothquant/debug/debug_llmi_blocks.py#L58-L80

drisspg (approved):

PR #1100, opened 2024-10-17
Add PyTorch 2.5 to regression test

### LINKED ISSUE #888: [CI] Add CI test for PyTorch 2.5.0rc
Just saw that PyTorch core made a branch cut for 2.5 release. Now nightly is 2.6. Probably a good idea to add CI to run on PyTorch 2.5rc? (don't even know if it's possible to install rc version of PyTorch). This would help to catch changes, especially new features, that for some reasons don't work with 2.5 but work with latest nightly.

Other miscs stuff:
- Add `TORCH_VERSION_AT_LEAST_2_6`
- Remove official support and CI for PyTorch 2.2?

gau-nernst:
Fixes #888


msaroufim:
Sorry for making a change to a draft PR but 2.2 is just way too old now so let's just get rid of it

gau-nernst:
No worries, I was about to ask you about 2.2 too. Just wanted to see if/which tests will fail with 2.5 (and sadly there are failing tests :cry:)

jerryzh168:
I think I've seen the error before, I think we'd just need to restrict all the tensor parallel tests to 2.6 and later

gau-nernst:
@jerryzh168 feel free to push the changes directly to this PR (or make a separate PR if you prefer that).

I already added `TORCH_VERSION_AT_LEAST_2_6` label in my previous BitNet PR.

gau-nernst:
@matthewdouglas Do you know why bnb is erroring out in this job? https://github.com/pytorch/ao/actions/runs/11396526050/job/31710520597?pr=1100

PyTorch 2.5 from pypi (built with CUDA 12.4) + bnb 0.42.0. I'm not sure why the CI install bnb 0.42 instead of 0.44 (perhaps due to another dependency?).

Seems like bnb 0.42 supports CUDA 12.1 but not CUDA 12.4? I guess we can install torch2.5 from https://download.pytorch.org/whl/cu121 in that case.

gau-nernst:
The torch2.5 CI job consistently get CUDA illegal memory access. I think the Galore kernel triggers it? (once illegal memory access is encountered, all subsequent CUDA ops will fail). Will try debug on my local machine.

I have skipped some distributed tests, so they only run on nightly (2.6).

There are still some failing AQT tests. @jerryzh168 perhaps you have some ideas?

torchao/prototype/galore/kernels/quant.py
```diff
@@ -22,17 +24,18 @@ def _dequant_kernel(
...
+    q_idx = tl.load(q_idx_ptr + offsets, mask=mask)
     tl.static_print(q_idx)
     # NOTE: Must upcast q_idx to int32 (q_idx is tl.uint8, which does not work for pointer indexing)
     q_vals = tl.load(qmap_ptr + q_idx.to(tl.int32))
-    absmax = tl.load(absmax_ptr + group_offsets)
+    absmax = tl.load(absmax_ptr + group_offsets)  # TODO: might need to add mask for this load also
 
     dq = q_vals * absmax
-    tl.store(dq_ptr + offsets, dq)
+    tl.store(dq_ptr + offsets, dq, mask=mask)
```
gau-nernst, line 38: @msaroufim This should fix the occasional illegal memory address (IMA) errors we observe.
- For some reasons torch2.5 in CI consistently produce this error (instead of randomly like other CI jobs)
- Running the galore tests alone will not reproduce this error due to CUDA caching allocator - the out-of-bounds addresses might still be valid address
  - To reproduce IMA error locally (without running the whole test suite), I allocate random tensors at the start of the test (e.g. `tensor = [torch.randn(torch.randint(100, 2000, ()).item(), torch.randint(100, 2000, ()).item(), device="cuda") for _ in range(100)]`)
- `tl.load(qmap_ptr + q_idx.to(tl.int32))` will never encounter out-of-bounds since `0 <= q_idx < 256` (due to uint8) and `qmap` always has size 256 (if it is used correctly)
- `tl.load(absmax_ptr + group_offsets)` MAY still face out-of-bounds error I think, but since `absmax` tensor is small, it's very likely that out-of-bound addresses are still valid. To be super safe, we would need to add mask for this one too. (sth like `mask=group_offsets < (M * N // GROUP_SIZE)` should be correct I think)

jerryzh168:
sorry just saw this, the AQT related errors seems to be related to inductor freezing, cc @leslie-fang-intel can you take a look?

leslie-fang-intel:
> sorry just saw this, the AQT related errors seems to be related to inductor freezing, cc @leslie-fang-intel can you take a look?

@gau-nernst did you mean the failure of `test/integration/test_integration.py::TestSaveLoadMeta::test_save_load_int8woqtensors_2_cpu`? I think the PR (https://github.com/pytorch/pytorch/pull/136353) to fix them landed in PyTorch main but not cherry-pick to 2.5 release branch. Can we skip them in PyTorch 2.5 regression test and keep it with main?

gau-nernst:
@leslie-fang-intel These are the failing tests (and their variants)

```
test/integration/test_integration.py::TestSubclass::test_int8_weight_only_quant_subclass_api_1_cpu
test/integration/test_integration.py::TestSubclass::test_int8_weight_only_quant_with_freeze_0_cpu
test/integration/test_integration.py::TestSaveLoadMeta::test_save_load_int8woqtensors_1_cpu
```

Not sure if they all have the same root cause. Can you add the appropriate skip test logic in my branch, since you would know best what to skip? (Not sure if you can push to my branch)

Also, do we need to add a doc somewhere about this?

leslie-fang-intel:
> @leslie-fang-intel These are the failing tests (and their variants)
> 
> ```
> test/integration/test_integration.py::TestSubclass::test_int8_weight_only_quant_subclass_api_1_cpu
> test/integration/test_integration.py::TestSubclass::test_int8_weight_only_quant_with_freeze_0_cpu
> test/integration/test_integration.py::TestSaveLoadMeta::test_save_load_int8woqtensors_1_cpu
> ```
> 
> Not sure if they all have the same root cause. Can you add the appropriate skip test logic in my branch, since you would know best what to skip? (Not sure if you can push to my branch)
> 
> Also, do we need to add a doc somewhere about this?

Seems I can't directly push to your branch. Stack a commit here https://github.com/leslie-fang-intel/ao/commit/1fdfa42b5bd76de5d99710bfd9354378c1d75684 and seems torch 2.5 testing are all green now as in: https://github.com/pytorch/ao/pull/1156.

There 2 issues regarding to these failures:

- Inductor freezing caused functional regression and which has been fixed in https://github.com/pytorch/pytorch/pull/136265 and documented in https://github.com/leslie-fang-intel/ao/blob/6fd1d6432c444c1aef24f70085f724c26b79bb88/torchao/quantization/README.md#workaround-with-unwrap_tensor_subclass-for-export-aoti-and-torchcompile-pytorch-24-and-before-only
- Inductor CPP backend regression fixed in https://github.com/pytorch/pytorch/pull/136353

leslie-fang-intel:
> Users have to use unwrap_tensor_subclass to get correct results? If this is the case, shouldn't the test do this instead of skipping it in 2.5?

Sure, update the skipping logic by using `unwrap_tensor_subclass` with `freezing` before torch2.6.

> It's a bug in PyTorch 2.5 that won't be fixed until there is a bug fix patch 2.5.1, if any? 

How to mark this PR: https://github.com/pytorch/pytorch/pull/136353 to cherry-pick in 2.5.1?

gau-nernst:
Replaced by #1156

PR #1111, opened 2024-10-17
Update smoke test

drisspg:
Stacked PRs:
 * __->__#1111

--- --- ---

### DIFF
--- a/.github/scripts/validate_binaries.sh
+++ b/.github/scripts/validate_binaries.sh
@@ -1,2 +1,6 @@
 pip install ${PYTORCH_PIP_PREFIX} torchao --index-url ${PYTORCH_PIP_DOWNLOAD_URL}
+# Intial smoke test, tries importing torchao
 python  ./test/smoke_tests/smoke_tests.py
+# Now we install dev-requirments and try to run the tests
+pip install -r dev-requirements.txt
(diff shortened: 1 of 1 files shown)


drisspg:
@atalman @msaroufim It looks like the validate binary might be a manual job, is it possible to test on the release branch w/ the cherry-pick:https://github.com/pytorch/ao/actions/runs/11393492434

.github/workflows/validate-binaries.yml
```diff
@@ -45,5 +45,7 @@ jobs:
       repository: "pytorch/ao"
       with_cuda: "enable"
       with_rocm: "disable"
-      smoke_test: "source ./.github/scripts/validate_binaries.sh"
+      smoke_test: |
+          pip install -r dev-requirements.txt
```
msaroufim, line 49: best to remove this line and only install torch, if the binaries crash because of a missing dependency that's bad

drisspg: My hope was that we can run the full test suite on the built binaries, and our test require dev-reqs to run

msaroufim: Ok maybe before installing requirements then you can run python -c "import torch"

msaroufim (approved):

PR #1113, opened 2024-10-18
support W4A8 Marlin kernel

HandH1998:
# Summary
We inroduce a mixed precision GEMM kernel for INT4-Weight and INT8-Activation. We implemented the W4A8 GEMM based on [Marlin GEMM](https://github.com/IST-DASLab/marlin). The kernel is designed to support our W4A8 quantization method [QQQ](https://github.com/HandH1998/QQQ). For more details on the kernel implementation, you can refer to our [paper](https://arxiv.org/pdf/2406.09904). The kernel demonstrates excellent performance and has been merged into the official vLLM project (see https://github.com/vllm-project/vllm/pull/5218). 

We hope the w4a8 GEMM can also provide a practical speedup for other W4A8 quantization methods in the community.
Additionally, since torchao is widely used in frameworks like [SGLang](https://github.com/sgl-project/sglang), we can extend support for W4A8 once the kernel is integrated into torchao.

# Performance
Here is the speedup over PyTorch FP16 GEMM (Calling CUTLASS) of all GEMMs under different numbers of input tokens. The weight matrix size is (N=8192, K=21760). You can reproduce the benchmark results using bench_w4a8.py in my [repo](https://github.com/HandH1998/marlin/tree/w4a8).
gemm_performance

### DIFF
--- a/torchao/quantization/marlin_qqq/README.md
+++ b/torchao/quantization/marlin_qqq/README.md
@@ -1,0 +1,6 @@
+# Marlin QQQ
+
+Marlin QQQ implementation adapted from the two below sources:
+
+* [QQQ](https://github.com/HandH1998/QQQ/tree/main)
+* [vllm](https://github.com/vllm-project/vllm/tree/main)
--- a/torchao/quantization/marlin_qqq/__init__.py
+++ b/torchao/quantization/marlin_qqq/__init__.py
@@ -1,0 +1,222 @@
+from typing import List, Tuple
+
+import numpy
+import torch
+
+from torchao.quantization.marlin_qqq.utils import (
+    marlin_permute_weights,
+    get_pack_factor,
+    qqq_quantize_weights,
+    const,
+)
+
+__all__ = [
+    "marlin_qqq_workspace",
+    "marlin_qqq_weights",
+    "get_qqq_scale_perms",
+    "get_qqq_weight_perm",
+    "marlin_qqq_permute_scales",
+    "marlin_qqq_quantize",
+]
+
+
+def marlin_qqq_workspace(
+    out_features: int,
+    min_thread_n: int = const.MIN_THREAD_N,
+    max_parallel: int = const.MAX_PARALLEL,
+) -> torch.Tensor:
+    """Creates a workspace for marlin qqq. The workspace is used to coordinate the locks
+    during the execution of the kernel.
+
+    Args:
+        out_features (int): The number of output features.
+        min_thread_n (int, optional): The minimum number of threads per block. Defaults to `MARLIN_QQQ_MIN_THREAD_N`.
+        max_parallel (int, optional): The maximum number of parallel threads. Defaults to `MARLIN_QQQ_MAX_PARALLEL`.
+    Returns:
+        torch.Tensor: The workspace tensor fully initialized with zeros.
+    """
+    assert (
+        out_features % min_thread_n == 0
+    ), f"out_features = {out_features}, min_thread_n = {min_thread_n}"
+    max_workspace_size = (out_features // min_thread_n) * max_parallel
+    return torch.zeros(max_workspace_size, dtype=torch.int, device="cuda")
+
+
+def marlin_qqq_weights(
+    q_w: torch.Tensor,
+    size_k: int,
+    size_n: int,
+    num_bits: int,
+    perm: torch.Tensor,
+    group_size: int,
+):
+    """Pack the quantized weights to the marlin format.
+
+    Args:
+        q_w (torch.Tensor): The quantized weight.
+        size_k (int): The number of input features.
+        size_n (int): The number of output features.
+        num_bits (int): The number of bits used for quantization.
+        perm (torch.Tensor): The weight permutation tensor.
+        group_size (int): The group size of quantization.
+    Returns:
+        torch.Tensor: The packed weight in marlin format.
+    """
+    # Permute
+    q_w = marlin_permute_weights(q_w, size_k, size_n, perm)
+
+    # Pack
+    pack_factor = get_pack_factor(num_bits)
+    orig_device = q_w.device
+
+    q_w = q_w.cpu().numpy().astype(numpy.uint32)
+
+    q_packed = numpy.zeros(
+        (q_w.shape[0], q_w.shape[1] // pack_factor), dtype=numpy.uint32
+    )
+    if group_size == size_k:
+        for i in range(pack_factor):
+            q_packed |= (q_w[:, i::pack_factor] & 0xF) << num_bits * i
+    else:
+        for i in range(pack_factor):
+            q_packed |= q_w[:, i::pack_factor] << num_bits * i
+
+    q_packed = torch.from_numpy(q_packed.astype(numpy.int32)).to(orig_device)
+
+    return q_packed
+
+
+def get_qqq_scale_perms() -> Tuple[List[int], List[int]]:
+    """Precompute permutations for the marlin scale shuffling.
+
+    Returns:
+        Tuple[List[int], List[int]]: Scale permutation list and
+        scale permutation list for a single group.
+    """
+
+    scale_perm: List[int] = []
+    for i in range(8):
+        scale_perm.extend([i + 8 * j for j in range(8)])
+    scale_perm_single: List[int] = []
+    for i in range(4):
+        scale_perm_single.extend([2 * i + j for j in [0, 1, 8, 9, 16, 17, 24, 25]])
+    return scale_perm, scale_perm_single
+
+
+# NOTE(HandH1998): QQQ employs different perms for per-group and per-channel weight quantization. # noqa: E501
+def get_qqq_weight_perm(num_bits: int, quant_type: str) -> torch.Tensor:
+    """Precompute permutations for the marlin weight shuffling.
+
+    Args:
+        num_bits (int): Number of bits to pack.
+        quant_type (str): The weight quantization type: per-group or per-channel.
+    Returns:
+        torch.Tensor: The weight permutation tensor.
+    """
+    perm_list: List[int] = []
+    for i in range(32):
+        perm1: List[int] = []
+        col = i // 4
+        for block in [0, 1]:
+            for row in [
+                4 * (i % 4),
+                4 * (i % 4) + 1,
+                4 * (i % 4) + 2,
+                4 * (i % 4) + 3,
+            ]:
+                perm1.append(16 * row + col + 8 * block)
+        for j in range(4):
+            perm_list.extend([p + 256 * j for p in perm1])
+
+    perm = numpy.array(perm_list)
+
+    assert quant_type in ["per-channel", "per-group"], "not supported quantization type"
+    if num_bits == 4:
+        if quant_type == "per-channel":
+            interleave = numpy.array([4, 0, 5, 1, 6, 2, 7, 3])
+        else:
+            interleave = numpy.array([0, 2, 4, 6, 1, 3, 5, 7])
+    else:
+        raise Exception("num_bits must be 4, got {}".format(num_bits))
+
+    perm = perm.reshape((-1, len(interleave)))[:, interleave].ravel()
+    perm = torch.from_numpy(perm)
+    return perm
+
+
+def marlin_qqq_permute_scales(
+    s_group: torch.Tensor,
+    s_channel: torch.Tensor,
+    size_k: int,
+    size_n: int,
+    group_size: int,
+) -> Tuple[torch.Tensor, torch.Tensor]:
+    """Converts per-group scale and per-channel scale to the format necessary for marlin.
+
+    Args:
+        s_group (torch.Tensor): The per-group quantization scale.
+        s_channel (torch.Tensor): The per-channel quantization scale.
+        size_k (int): The number of input features.
+        size_n (int): The number of output features.
+        group_size (int): The group size of quantization.
+    Returns:
+        Tuple[torch.Tensor, torch.Tensor]: The scale tensors in the marlin format.
+    """
+    scale_perm, scale_perm_single = get_qqq_scale_perms()
+    if group_size < size_k and group_size != -1:
+        s_group = s_group.reshape((-1, len(scale_perm)))[:, scale_perm]
+        s_channel = s_channel.reshape((-1, len(scale_perm_single)))[
+            :, scale_perm_single
+        ]
+        s_group = s_group.reshape((-1, size_n)).contiguous()
+    else:
+        s_channel = s_channel.reshape((-1, len(scale_perm_single)))[
+            :, scale_perm_single
+        ]
+    s_channel = s_channel.reshape((-1, size_n)).contiguous()
+
+    return s_group, s_channel
+
+
+def marlin_qqq_quantize(
+    w: torch.Tensor,
+    num_bits: int,
+    group_size: int,
+) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]:
+    """Quantize the weight and convert it to the marlin format.
+
+    Args:
+        w (torch.Tensor): The original weight.
+        num_bits (int): The number of bits for quantization.
+        group_size (int): The group size of quantization.
+    Returns:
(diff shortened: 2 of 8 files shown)


drisspg:
can we do some comparisons between this and https://github.com/pytorch/ao/pull/880?

jerryzh168:
thanks, looks pretty good, can you add the benchmark code in https://github.com/pytorch/ao/tree/main/benchmarks as well

torchao/quantization/marlin_qqq/README.md
```diff
@@ -0,0 +1,6 @@
+# Marlin QQQ
+
+Marlin QQQ implementation adapted from the two below sources:
```
msaroufim, line 3: Could you summarize the key changes made from the Marlin kernel here?

HandH1998: ok, I will do it these days.

HandH1998:
@drisspg @jerryzh168 I am working on the benckmark and I will give the comparisons with https://github.com/pytorch/ao/pull/880. The kernel is benchmarked on A100-80G GPU and it can work on SM > 8.0.

HandH1998:
@jerryzh168 @drisspg @msaroufim 
I have made the following modifications (code modifications refer to https://github.com/pytorch/ao/pull/621 and https://github.com/pytorch/ao/pull/880):
1. Added benchmark code for marlin_qqq_w4a8 GEMM in`benchmarks/benchmark_marlin_qqq.py`
2. Summarized the main differences between marlin_qqq_w4a8 GEMM and marlin_w4a16 GEMM in`torchao/quantization/marlin_qqq/README.md`
3. Supported marlin_qqq in `torchao/quantization/quant_api.py`
4. Added some unit tests in `test/quantization/test_marlin_qqq.py`

[w4a8-cutlass](https://github.com/pytorch/ao/pull/880) is a great work.  In comparison, we believe marlin_qqq_w4a8 can support weight per-group quantization in addition to weight per-channel quantization. However, marlin_qqq does have some limitations: it only supports symmetric quantization and the output dtype can only be `torch.float16`.

In addition, we also provide the performance of `torchao/_models/llama/generate.py` here. `-g128` means  weight per-group quantization and the group size is 128.

| `-q parameter` | Precison | Average tokens/sec | Average Bandwidth in GB/s | Peak Memory Usage in GB | Model Size in GB |
|:----------|:----------:|:----------:|:----------:|:----------:|:----------:|
| `--compile` | fp16 | 112.45 | 1486.00 | 13.93 | 13.21 |
| `-q marlin_qqq --compile` | w4a8 | 197.45 | 653.50 | 4.79 | 3.31 |
| `-q marlin_qqq --compile` | w4a8-g128 | 187.62 | 640.32 | 4.82 | 3.41 |

HandH1998:
@jerryzh168 @msaroufim @drisspg I have resolved the conficts. Look forward to see your new advice.

drisspg:
Hey @HandH1998  rekicking off the internal CI/CD here that failed

Overall this works looks very good, will do a more thorough review

drisspg:
It seems that you also applied formatting to a lot of files. This makes it pretty hard to review since all the changes get mixed together. 

I opened https://github.com/pytorch/ao/pull/1226 

Would you mind opening a separate PR for some of the other files you touched? And let's first add the formatting and then we can merge in your changes and make it easier to review.

HandH1998:
@drisspg I have removed the extra formatting in this PR, which should now simplify the review process.

alexsamardzic:
I've added a benchmarking script to #880, that makes it possible to compare the performance between the two W4A8 kernels.  As CUTLASS-based version doesn't support group quantization, at the moment it is only possible to make the comparison with `group_size=-1` in the Marlin-based version.  Marlin-based version performs clearly better for input sizes less than 256, while CUTLASS-based version is faster for input sizes of 256 and higher.  Consequently, Marlin-based version performs better on Llama generator too, with tokens/sec about 25% higher than CUTLASS-based version (note that in both cases I ran the generator as follows `python generate.py --compile --precision torch.float16 -q ...`).  Please note that the comparison is not completely apples-to-apples, as besides group quantization support there are other small differences between kernels, but still it seems this is pretty much current state of affairs regarding the performance.

Let me return the compliment by stating that this Marlin-based kernel is a great work too.  In particular, for me it clearly shows where the CUTLASS-based kernel should be improved.

torchao/quantization/quant_api.py
```diff
@@ -501,13 +522,18 @@ def apply_int8_dynamic_activation_int4_weight_quant(weight, group_size=32, mappi
...
+    else:
+        assert False, f"Unsupported input mapping type: {input_mapping_type}"
 
-    weight = to_affine_quantized_intx(weight, mapping_type, block_size, target_dtype, quant_min, quant_max, eps)
+    weight = to_affine_quantized_intx(weight, mapping_type, block_size, target_dtype, quant_min, quant_max, eps, _layout=layout)
     weight = to_linear_activation_quantized(weight, input_quant_func)
     return weight
 
-def int8_dynamic_activation_int4_weight(group_size=32, mapping_type=MappingType.SYMMETRIC):
+def int8_dynamic_activation_int4_weight(group_size=32, layout=PlainLayout(), mapping_type=MappingType.SYMMETRIC, input_mapping_type=MappingType.ASYMMETRIC):
```
drisspg, line 536: can you update the doc block, maybe with some description of how the supported layouts

torchao/quantization/marlin_qqq/README.md
```diff
@@ -0,0 +1,13 @@
+# Marlin QQQ
+
+Marlin QQQ kernel and Marlin kernel mainly have the following differences:
```
drisspg, line 3: Also can you add some description on the expected Sm support required, from the ifdefs looks like its sm80 +?

test/quantization/test_marlin_qqq.py
```diff
@@ -0,0 +1,129 @@
...
+
+class MarlinQQQ(TestCase):
+    def setUp(self):
+        super().setUp()
+        torch.manual_seed(0)
+
+        self.input = torch.randn((32, 16, 4096), dtype=torch.float16, device="cuda")
+        self.model = (
+            nn.Sequential(
+                nn.Linear(4096, 21504),
```
drisspg, line 31: probably a little bigger than needed for testing

torchao/quantization/marlin_qqq/__init__.py
```diff
@@ -0,0 +1,373 @@
...
+
+        max_q_val = 2**num_bits - 1
+        half_q_val = (max_q_val + 1) // 2
+
+        # Compute scale for each group
+        s_group = torch.max(torch.abs(w), 0, keepdim=True)[0]
+        s_group *= 2 / max_q_val  # 2 => symmetric
+
+        # Quantize
+        q_w = torch.round(w / s_group).int()
```
drisspg, line 90: cc @jerryzh168  I think we should make this block-size to -> max_abs a utility

torchao/quantization/marlin_qqq/__init__.py
```diff
@@ -0,0 +1,373 @@
...
+    perm = get_qqq_weight_perm(num_bits, quant_type)
+    q_w = marlin_permute_weights(q_w, size_k, size_n, perm)
+
+    # Pack
+    pack_factor = get_pack_factor(num_bits)
+    orig_device = q_w.device
+
+    q_w = q_w.cpu().numpy().astype(numpy.uint32)
+
+    q_packed = numpy.zeros(
```
drisspg, line 207: curious why we go to numpy for this?

torchao/quantization/marlin_qqq/utils.py
```diff
@@ -0,0 +1,186 @@
...
+                4 * (i % 4) + 2,
+                4 * (i % 4) + 3,
+            ]:
+                perm1.append(16 * row + col + 8 * block)
+        for j in range(4):
+            perm_list.extend([p + 256 * j for p in perm1])
+
+    perm = numpy.array(perm_list)
+
+    assert quant_type in ["per-channel", "per-group"], "not supported quantization type"
```
drisspg, line 127: quant_type seems like an enum would be more appropriate

torchao/quantization/quant_primitives.py
```diff
@@ -838,6 +840,122 @@ def _choose_qparams_affine(
...
 
 
+def choose_qparams_and_quantize_affine_qqq(
+    w: torch.Tensor,
+    mapping_type: str,
+    num_bits: int,
+    group_size: int,
+) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+    assert (
+        mapping_type == MappingType.SYMMETRIC
```
drisspg, line 850: seems like a dupe of the implementation of the stuff in utils can we directly call that func in here/ is this needed?

torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -329,37 +339,47 @@ def from_hp_to_intx(
...
-                block_size,
-                scale,
-                zero_point,
-                target_dtype,
-                quant_min,
-                quant_max,
-                zero_point_domain,
-            )
-            # Note: output will be uint8 tensor for sub byte tensors for now
+            if isinstance(_layout, MarlinQQQLayout):
```
drisspg, line 342: @jerryzh168 

All this if else branching makes me feel like something is wrong with the abstraction

drisspg:
Overall looking really good, would you mind reporting lib size increase from this PR? I plan to take another once over of the cuda code tomorrow and then once CI is good this should be good to go :)

torchao/quantization/marlin_qqq/utils.py
jerryzh168: we have https://github.com/pytorch/ao/blob/main/torchao/quantization/granularity.py btw

torchao/quantization/quant_api.py
```diff
@@ -487,9 +487,30 @@ def _int8_asymm_per_token_quant(x: torch.Tensor) -> torch.Tensor:
...
+        _get_per_token_block_size(x),
+        target_dtype,
+        eps=eps,
+        quant_min=quant_min,
+        quant_max=quant_max,
+        scale_dtype=torch.float32,
+    )
+
+
+def apply_int8_dynamic_activation_int4_weight_quant(weight, group_size=32, layout=PlainLayout(), mapping_type=MappingType.SYMMETRIC, input_mapping_type=MappingType.ASYMMETRIC):
```
jerryzh168, line 509: nit: probably clearer if we also rename the first `mapping_type` to `weight_mapping_type`, but this can be a follow up since we need to be careful about BC

torchao/quantization/marlin_qqq/__init__.py
jerryzh168: it's just https://github.com/pytorch/ao/blob/e41ca4ee41f5f1fe16c59e00cffb4dd33d25e56d/torchao/quantization/quant_primitives.py#L773-L777 I think it's possible to reuse already

torchao/quantization/marlin_qqq/__init__.py
HandH1998: Originally, `q_w` is `torch.uint32`, but since torch does not support `lshift_cuda` or `lshift_cpu`, we used `numpy+uint32` to implement the packing function. However, I have discovered that we can convert `q_w` to `torch.int64`, which supports `lshift_cuda`, and perform the packing on the GPU. This should be faster. I will update the code accordingly.

torchao/quantization/quant_api.py
HandH1998: Yes, I agree that this should be addressed later.

torchao/quantization/quant_primitives.py
HandH1998: have removed the stuff in utils and call the func here.

torchao/quantization/marlin_qqq/utils.py
HandH1998: I have modified the `quant_type` to `granularity` in  https://github.com/pytorch/ao/blob/main/torchao/quantization/granularity.py.

torchao/quantization/marlin_qqq/__init__.py
HandH1998: I have updated the code.

torchao/quantization/marlin_qqq/README.md
HandH1998: finished

test/quantization/test_marlin_qqq.py
HandH1998: finished

HandH1998:
@drisspg @jerryzh168 Thanks for your reviews. I have fixed the most issues according to your advice. The increased lib size is about 5M. Let's go forward :)

drisspg:
Lib increases from 3.5M -> 5.0M for provenance, I think this is acceptable

drisspg (approved):

drisspg:
@HandH1998  looks like there is 1 true failure on your PR would you mind fixing and then we can land?

torchao/dtypes/affine_quantized_tensor.py
jerryzh168: yeah I don't think this is intended actually, AffineQuantizedTensor is corresponding to a very specific quantize/dequantize algorithm, so if there is a different quantize/dequantize people will need to create another tensor subclass

torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -329,37 +339,47 @@ def from_hp_to_intx(
...
+                    quant_max,
+                    zero_point_domain,
+                )
+                # Note: output will be uint8 tensor for sub byte tensors for now
 
         data = _layout.post_process(data)
         tensor_impl_ctr = get_tensor_impl_constructor(type(_layout))
-        tensor_impl = tensor_impl_ctr(data, scale, zero_point, _layout)
+        if isinstance(_layout, MarlinQQQLayout):
+            tensor_impl = tensor_impl_ctr(data, s_group, s_channel, _layout)
```
jerryzh168, line 380: yeah it feels like it's better to create a new tensor subclass, we defined affine quantization being: https://github.com/pytorch/ao/blob/0e854ecbf75350d781fa60140ab1d3180a0404db/torchao/dtypes/affine_quantized_tensor.py#L136-L137 does it still hold for MarlinQQQ?

torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -653,6 +673,11 @@ def pre_process(self, input: torch.Tensor) -> torch.Tensor:
         return w_24.t()
 
 
+@dataclass(frozen=True)
+class MarlinQQQLayout(Layout):
```
jerryzh168, line 677: just a note: if you are creating a new tensor subclass, the layout/tensor_impl is not required if you feel it does not help in any way. unless you want to expose `MarlinQQQLayout` as a top level API for `int8_dynamic_activation_int4_weight`, as you do right now in the PR.

HandH1998: got it.

HandH1998:
> @HandH1998 looks like there is 1 true failure on your PR would you mind fixing and then we can land?

I will try to fix it soon.

torchao/dtypes/affine_quantized_tensor.py
HandH1998: @jerryzh168 make sense.

torchao/dtypes/affine_quantized_tensor.py
jerryzh168: yeah that sounds good to me

HandH1998: I have modified, please check it.

msaroufim (approved):
just need to fix lint

HandH1998:
> can you add a table similar to https://github.com/pytorch/ao/tree/main/torchao/quantization#sparse-marlin to README to show the performance? otherwise looks good to me

I have added it.

jerryzh168 (approved):
thanks!

jerryzh168:
please fix lint: https://github.com/pytorch/ao/actions/runs/11812082822/job/32933511633?pr=1113

zhyncs:
@HandH1998 It's so coooooool!

PR #1114, opened 2024-10-18
[AFQ] Optimize tensor_flatten for runtime

IvanKobzarev:
* __->__ #1114

__tensor_flatten__ is called at runtime when inputs are subclasses e.g. AffineQuantizedTensor

In case of using `float8_dynamic_activation_float8_weight` quantization, the activations will also be AQT.

There was external complain, that unwrap_tensor_subclasses sometimes exceeds in duration that compiled region execution.

Profiling shows that tensor_flatten attribute access will go through torch_function handling (AQT has it).
If to remove this torch_function dispatch for each getattr at runtime - tensor_flatten becomes (in my measurements x9 faster)

Screenshot 2024-10-18 at 14 31 33


jerryzh168 (approved):
do we need to do the same for TensorImpl tensor subclasses?

PR #1115, opened 2024-10-18
Remove old gptq

danielpatrickhug:
This PR is for the purpose of migrating several quantization classes in GPTQ.py to the new quantization.quantize_linear module before deleting the legacy GPTQ module following PR #914. This PR also handles changing the imports and README to correctly mirror the new structure as well as updating and removing tests in both test_qat and test_quant_api. @HDCharles @andrewor14 @jerryzh168 assisted and advised on how to execute this migration PR.


test/quantization/test_quant_api.py
```diff
@@ -225,7 +225,7 @@ def api(model):
     @unittest.skipIf(not TORCH_VERSION_AT_LEAST_2_3, "skipping when torch verion is 2.3 or lower")
     def test_8da4w_quantizer(self):
         from torchao.quantization.quant_api import Int8DynActInt4WeightQuantizer
-        from torchao.quantization.GPTQ import Int8DynActInt4WeightLinear
+        from torchao.quantization.quantize_linear import Int8DynActInt4WeightLinear
```
andrewor14, line 228: nit: can we call this `quantized_linear` instead?

torchao/quantization/GPTQ.py
```diff
@@ -1,1193 +0,0 @@
-# Copyright (c) Meta Platforms, Inc. and affiliates.
-# All rights reserved.
-#
-# This source code is licensed under the BSD-style license found in the
-# LICENSE file in the root directory of this source tree.
-
-
-# This source code is licensed under the license found in the
-# LICENSE file in the root directory of this source tree.
-
```
andrewor14, line 10: Can we keep this file and keep the legacy imports for now? It will help us migrate some known use cases. E.g. something like
```
from torchao.quantization.quantized_linear import (
    Int4WeightOnlyQuantizer,
    Int8DynActInt4WeightQuantizer,
    ...
)
```

test/quantization/test_qat.py
```diff
@@ -68,7 +68,7 @@
     TORCH_VERSION_AT_LEAST_2_5,
 )
 
-from torchao.quantization.GPTQ import (
+from torchao.quantization.quantize_linear import (
```
jerryzh168, line 71: this will be bc breaking since it's used by other repos like torchchat, so we'd probably want to keep backward compatibility by still keeping the old import path

torchao/quantization/quantize_linear.py
```diff
@@ -0,0 +1,469 @@
...
+import torch.nn.functional as F
+from torchao.quantization.unified import Quantizer
+from typing import Optional, Callable, Type, Dict, Any
+from torchao.quantization.utils import (
+    per_token_dynamic_quant,
+    groupwise_affine_quantize_tensor,
+    group_quantize_tensor_symmetric,
+)
+
+
```
andrewor14, line 14: maybe add a comment to say everything in this file is intended to be deprecated and eventually removed? Then we should also add the new alternative of running these using the quantize_ API, e.g.
```
    quantize_,
    int4_weight_only,
    int8_dynamic_activation_int4_weight,
)
quantize_(model, int4_weight_only())
quantize_(model, int8_dynamic_activation_int4_weight())
```

torchao/quantization/quant_api.py
```diff
@@ -99,14 +101,12 @@
...
-from .GPTQ import (
+from torchao.quantization.quantize_linear import (
     Int8DynActInt4WeightQuantizer,
-    Int8DynActInt4WeightGPTQQuantizer,
-
 )
 __all__ += [
     "Int8DynActInt4WeightQuantizer",
-    "Int8DynActInt4WeightGPTQQuantizer",
+
```
andrewor14, line 109: nit: remove this empty line?

torchao/quantization/__init__.py
```diff
@@ -37,6 +39,9 @@
     "compute_error",
     "Int4WeightOnlyGPTQQuantizer",
     "Int4WeightOnlyQuantizer",
+    "WeightOnlyInt4Linear",
+    "Int8DynActInt4WeightLinear",
+    "Int8DynActInt4WeightQuantizer",
```
andrewor14, line 44: Maybe we shouldn't expose these if they weren't exposed before, since these are meant to be legacy APIs?

test/quantization/test_quant_api.py
```diff
@@ -401,43 +336,6 @@ def test_gptq_quantizer_int4_weight_only(self):
             f"accuracy regressed from 7.76 to {result['results']['wikitext']['word_perplexity,none']}"
         )
 
-    @unittest.skip("skipping until we get checkpoints for gpt-fast")
-    def test_quantizer_int4_weight_only(self):
```
andrewor14, line 405: Does this PR just move the location of these classes, if so this test should still pass right, or am I missing something? @danielpatrickhug @HDCharles Is there a reason why we want to remove this test?

andrewor14:
Hi Daniel, thanks for cleaning this up. It looks great to me overall! I just had some comments about maintaining BC to give us time to migrate some known use cases.

test/quantization/test_quant_api.py
andrewor14: or maybe even `_quantized_linear` to discourage people from using it

test/quantization/test_quant_api.py
HDCharles: We don't want 2 int4 implementations

torchao/quantization/quantize_linear.py
danielpatrickhug: Hey, can you elaborate on this step a little bit more? I'm a bit confused.

andrewor14: Yeah sure, I mean we can express everything in this file using the new `quantize_` API:

```
# Before:
quantizer = Int4WeightOnlyQuantizer()  # or Int8DynActInt4WeightQuantizer
model = quantizer.quantize(model)

# After:
quantize_(model, int4_weight_only())  # or int8_dynamic_activation_int4_weight
```

We want to deprecate the Quantizer APIs eventually in favor of the quantize_ API, so we want to discourage people from using the quantizers in this file. I think adding a comment here explaining this context will help with that.

danielpatrickhug: Thank you, I added the comments to the top of the file noting the deprecation.

torchao/quantization/prototype/qat/linear.py
```diff
@@ -26,7 +24,11 @@
 from torchao.quantization.utils import get_group_qparams_symmetric
 from .api import FakeQuantizeConfig
 from .fake_quantizer import FakeQuantizer
-from .utils import (
+from torchao.quantization.utils import (
+    get_groupwise_affine_qparams,
+    groupwise_affine_quantize_tensor,
+)
+from torchao.quantization.prototype.qat.utils import (
```
andrewor14, line 31: sorry this is recently updated to `torchao.quantization.qat.utils`

andrewor14: Oh actually you'll face some conflicts with this file. Can you make the changes in `torchao/quantization/qat/linear.py` instead? Sorry for the merge conflicts

andrewor14 (approved):
Looks great. Thanks for the refactor!

andrewor14:
Oh actually I just realized. The latest changes don't actually remove the classes from GPTQ.py. We should still do that so we don't end up with duplicate copies of the same classes. What I meant was remove those classes from GPTQ.py and add the following imports to GPTQ.py to maintain legacy imports:

```
from torchao.quantization._quantized_linear import (
    Int4WeightOnlyQuantizer,
    Int8DynActInt4WeightLinear,
    Int8DynActInt4WeightQuantizer,
    WeightOnlyInt4Linear,
)
```

PR #1122, opened 2024-10-20
fixed `weights_only=True` load for `float8_dynamic_activation_float8_weight` in quant_api

### LINKED ISSUE #1118: Make float8_dynamic_activation_float8_weight work with `weights_only` load
# Summary

repro:
```Python
import torch
import torch.nn as nn
from torchao.quantization.quant_api import quantize_, PerRow, float8_dynamic_activation_float8_weight, float8_weight_only

linear_module = nn.Linear(16, 32).to(dtype=torch.bfloat16, device = "cuda")

quantize_(linear_module, float8_dynamic_activation_float8_weight(granularity=PerRow())) 

torch.save(linear_module.state_dict(), "linear_module.pt")

sd = torch.load("linear_module.pt", weights_only=True)
#
with torch.device("meta"):
    new_model = nn.Linear(16, 32)
    new_model.load_state_dict(sd, assign=True)
    
 ```
 
 This is because currently we store a partial on the class which is not safe serializable:
 
 https://github.com/pytorch/ao/blob/677af6f919baf2748f6117b62f320056c42fd0b0/torchao/quantization/quant_api.py#L768
 
 We should update this and other quant apis to instead save the kwargs needed for the original function and then register the underlying activation quant function. I already did this for static activation casting:
 
 https://github.com/pytorch/ao/blob/677af6f919baf2748f6117b62f320056c42fd0b0/torchao/quantization/weight_tensor_linear_activation_quantization.py#L45

### PR DESCRIPTION
**Pull request by p4arth**
Fixes #1118
Changes made:
1. Modified the `float8_dynamic_activation_float8_weight` function to have a flow like `float8_static_activation_float8_weight` inside the `quant_api.py` as suggested by @drisspg.
2. Added `quant_kwargs` in `linear_activation_quantized_tensor.py` so that they can be serialized using `torch.save` instead of unsafe serialization of partially initialized `input_quant_func`.

### DIFF
--- a/torchao/quantization/linear_activation_quantized_tensor.py
+++ b/torchao/quantization/linear_activation_quantized_tensor.py
@@ -26,6 +26,8 @@ class LinearActivationQuantizedTensor(TorchAOBaseTensor):
         on top of it
       `input_quant_func` (Callable[[torch.Tensor], torch.Tensor]): a function that takes a high precision floating point tensor and returns
         a quantized tensor, this is used to quantize input
+      `quant_kwargs` (Dict[str, Any]): Additional keyword arguments for the quantization function.
+        Restriction: Must not contain tensor values.
     """
     quant_kwargs: Dict[str, Any]
     
--- a/torchao/quantization/linear_activation_quantized_tensor.py
+++ b/torchao/quantization/linear_activation_quantized_tensor.py
@@ -1,5 +1,5 @@
 import torch
-from typing import Callable, Dict, Any
+from typing import Callable, Dict, Any, Optional
 from torch.utils._python_dispatch import return_and_correct_aliasing
 from torchao.utils import (
     TorchAOBaseTensor,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by jerryzh168 — DISMISSED** [review]
looks good I think, can you remove https://github.com/pytorch/ao/blob/f33cff7053530506d11085eb18e6b93f5e224faf/test/dtypes/test_affine_quantized_float.py#L199-L201 and run the test

**Review by jerryzh168 — APPROVED** [review]

File: torchao/quantization/linear_activation_quantized_tensor.py [hunk 1]
```diff
@@ -44,49 +47,60 @@ def __init__(
...
-        return ["original_weight_tensor"], [self.input_quant_func]
+        return ["original_weight_tensor"], [self.input_quant_func, self.quant_kwargs]
 
     @classmethod
     def __tensor_unflatten__(
         cls, tensor_data_dict, tensor_attributes, outer_size, outer_stride
     ):
         original_weight_tensor = tensor_data_dict["original_weight_tensor"]
-        input_quant_func, = tensor_attributes
+        input_quant_func, quant_kwargs = tensor_attributes
```
drisspg, line 67: These should not be tensor attributes

drisspg (changes requested):
Some comments

drisspg (approved):
Sorry, this looks great! Maybe update the doc similar to 
https://github.com/pytorch/ao/blob/85ec2097cb9120ccbb40dc39a92be4496bfcda77/torchao/quantization/weight_tensor_linear_activation_quantization.py#L29-L30

PR #1126, opened 2024-10-21
Float8 inference roofline benchmarks

jainapurva:


drisspg:
Looks good I think we also want to figure out a mechanism for capturing some of the preceding graph the attention to determine how well the dynamic casting is fused for dynamic techniques but otherwise this seeems very helpful!

PR #1141, opened 2024-10-22
is_linear fix for MHA

HDCharles:
Summary: filter fn may need access to parent types of module as in the case with mha

Test Plan: python /home/cdhernandez/local/ao/test/integration/test_integration.py -k "test_autoquant_mha"

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/integration/test_integration.py
+++ b/test/integration/test_integration.py
@@ -1296,6 +1296,32 @@ class TestAutoQuant(unittest.TestCase):
 
     @parameterized.expand(COMMON_DEVICE_DTYPE)
     @unittest.skipIf(not TORCH_VERSION_AT_LEAST_2_5, "autoquant requires 2.5+.")
+    def test_autoquant_mha(self, device, dtype):
+        class MHAModel(torch.nn.Module):
+            def __init__(self):
+                super().__init__()
+                self.mha = torch.nn.MultiheadAttention(4096, 32)
+                self.lin = torch.nn.Linear(4096, 4096)
+
+            def forward(self, x):
+                y = self.mha(x, x, x)[0]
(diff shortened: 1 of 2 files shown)


torchao/quantization/quant_api.py
```diff
@@ -234,6 +234,7 @@ def _is_linear(mod, *args):
         and not isinstance(mod.weight, AffineQuantizedTensor)
         and not isinstance(mod.weight, LinearActivationQuantizedTensor)
         and not isinstance(mod.weight, AffineFakeQuantizedTensor)
+        and not isinstance(mod, nn.modules.linear.NonDynamicallyQuantizableLinear)
```
jerryzh168, line 237: nit: maybe add a Note here, to say that if MHA is refactored to not use this module we'd need a different solution

jerryzh168 (approved):
looks good, thanks!

test/integration/test_integration.py
```diff
@@ -1278,6 +1278,32 @@ def test_autoquant_compile(self, device, dtype, m1, m2, k, n):
         sqnr = SQNR(out, out2)
         self.assertTrue(sqnr >= 30)
 
+    @parameterized.expand(COMMON_DEVICE_DTYPE)
+    @unittest.skipIf(not TORCH_VERSION_AT_LEAST_2_5, "autoquant requires 2.5+.")
```
jerryzh168, line 1282: according to CI, probably need to add skip if cuda is not available, or change this https://github.com/pytorch/ao/blob/629aee11a37d3b79f8352a247f39d35491da3464/test/integration/test_integration.py#L97 to not include cuda when it's not available

PR #1142, opened 2024-10-22
[ROCm] use dataclass for fnuz type setting

jeffdaily:
This allows the float8 type selection have reasonable defaults depending on the detected hardware.


drisspg (approved):
Looks good, I think you will need to format and will let CI see if anything kicks up

msaroufim (approved):
thanks! I went ahead and fixed the lint

PR #1145, opened 2024-10-23
Move files to prototype/sparsity

jainapurva:
Move files from sparsity/prototype -> prototype/sparsity
Maintain Backward-compatibility (will need a follow-up PR to clean up later)

Fixes: https://github.com/pytorch/ao/issues/1013


jerryzh168 (approved):
looks good, just need to make sure both OSS CI and internal CI passes before landing

Issue #1110, opened 2024-10-17
Named Symbol not found (torchchat #1298)


mikekgfb:
Quantized model gets a CUDA error "Named symbol not found".
see https://github.com/pytorch/torchchat/issues/1298

mikekgfb:
Cuda version:
NVIDIA-SMI 535.104.05             Driver Version: 535.104.05   CUDA Version: 12.2  

This ran on Google colab.  Detailed trace-back/repro: https://colab.research.google.com/drive/1PRneJBaS5TlJaIgc4Lwv2muiePp6T9Ss?usp=sharing

```
$nvidia-smi
Thu Oct 17 19:52:30 2024       
+---------------------------------------------------------------------------------------+
| NVIDIA-SMI 535.104.05             Driver Version: 535.104.05   CUDA Version: 12.2     |
|-----------------------------------------+----------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |         Memory-Usage | GPU-Util  Compute M. |
|                                         |                      |               MIG M. |
|=========================================+======================+======================|
|   0  Tesla T4                       Off | 00000000:00:04.0 Off |                    0 |
| N/A   39C    P8               9W /  70W |      0MiB / 15360MiB |      0%      Default |
|                                         |                      |                  N/A |
+-----------------------------------------+----------------------+----------------------+
                                                                                         
+---------------------------------------------------------------------------------------+
| Processes:                                                                            |
|  GPU   GI   CI        PID   Type   Process name                            GPU Memory |
|        ID   ID                                                             Usage      |
|=======================================================================================|
|  No running processes found                                                           |
+---------------------------------------------------------------------------------------+

$ python torchchat.py generate stories110M --quant torchchat/quant_config/cuda.json --prompt "It was a dark and stormy night, and"
2024-10-17 19:52:38.809413: E external/local_xla/xla/stream_executor/cuda/cuda_fft.cc:485] Unable to register cuFFT factory: Attempting to register factory for plugin cuFFT when one has already been registered
2024-10-17 19:52:39.031641: E external/local_xla/xla/stream_executor/cuda/cuda_dnn.cc:8454] Unable to register cuDNN factory: Attempting to register factory for plugin cuDNN when one has already been registered
[...]

PR #636, opened 2024-08-08
Add CUDA compute capability compile guard 

gau-nernst:
Fixes #632 

Tested on Google Colab

image

Issue #632, opened 2024-08-08
Make build from source easier on unsupported hardware


gau-nernst:
Compiling CUDA extensions on unsupported hardware (i.e. compute capability is too low) will error out. Reported by @danielpatrickhug on CUDA-MODE Discord.

<details>

<summary>Error logs</summary>

```
  ⚡ main ~/ao python setup.py develop
  running develop
  /home/zeus/miniconda3/envs/cloudspace/lib/python3.12/site-packages/setuptools/command/develop.py:42: EasyInstallDeprecationWarning: easy_install command is deprecated.
  !!
  
          ********************************************************************************
          Please avoid running ``setup.py`` and ``easy_install``.
          Instead, use pypa/build, pypa/installer or other
          standards-based tools.
  
          See https://github.com/pypa/setuptools/issues/917 for details.
          ********************************************************************************
  
  !!
    easy_install.initialize_options(self)
  /home/zeus/miniconda3/envs/cloudspace/lib/python3.12/site-packages/setuptools/_distutils/cmd.py:66: SetuptoolsDeprecationWarning: setup.py install is deprecated.
  !!
  
          ********************************************************************************
          Please avoid running ``setup.py`` directly.
          Instead, use pypa/build, pypa/installer or other
          standards-based tools.
  
          See https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html for details.
          ********************************************************************************
  
  !!
    self.initialize_options()
  running egg_info
  writing torchao.egg-info/PKG-INFO
  writing dependency_links to torchao.egg-info/dependency_links.txt
  writing requirements to torchao.egg-info/requires.txt
  writing top-level names to torchao.egg-info/top_level.txt
  reading manifest file 'torchao.egg-info/SOURCES.txt'
  adding license file 'LICENSE'
  writing manifest file 'torchao.egg-info/SOURCES.txt'
  running build_ext
  /home/zeus/miniconda3/envs/cloudspace/lib/python3.12/site-packages/torch/utils/cpp_extension.py:424: UserWarning: There are no g++ version bounds defined for CUDA version 12.1
    warnings.warn(f'There are no {compiler_name} version bounds defined for CUDA version {cuda_str_version}')
  building 'torchao._C' extension
  creating /teamspace/studios/this_studio/ao/build/temp.linux-x86_64-cpython-312
  creating /teamspace/studios/this_studio/ao/build/temp.linux-x86_64-cpython-312/torchao
[...]

PR #1147, opened 2024-10-23
BF16 support for Quant-LLM kernel

### LINKED ISSUE #998: Make Quant-LLM compatible with BF16

tobiasvanderwerff:
Closes #998 

This PR adds BF16 support for the existing FPx CUDA kernel ([here](https://github.com/pytorch/ao/tree/main/torchao/csrc/cuda/fp6_llm)), which was originally written for FP16 (see the [paper](https://arxiv.org/abs/2401.14112) for details). Since all recent models are trained and released with BF16, having BF16 support potentially improves accuracy for FPx models.

## Most important changes

1. Use C++ templates in the kernel to allow for either FP16 or BF16 input.
2. Change Python wrapper code for floatx/fpx to take into account BF16 inputs.
3. Add tests for bf16 ([test/test_ops.py](https://github.com/pytorch/ao/compare/main...tobiasvanderwerff:ao:fp6-bf16#diff-d183f2afc51d6a59bc70094e8f476d2468c45e415500f6eb60abad955e065156) and [test/dtypes/test_floatx.py](https://github.com/pytorch/ao/compare/main...tobiasvanderwerff:ao:fp6-bf16#diff-0e811fa3416cd87d9a25b4fb680890098c69aa33ca4db4d347d4a10cc41e0eb3)).
4. Change dequant logic for BF16 (see [FPx_FP16_Cast_4Way](https://github.com/pytorch/ao/compare/main...tobiasvanderwerff:ao:fp6-bf16#diff-d1a5769dc17cd4ffca57d2a4106d9780d8de281654bdcc91e88d64b49e09284dR35) function). This is simply a matter of changing the bit shifting to take into account the difference in exponents bits between FP16 and BF16 (5 vs. 8, respectively). Figure 6 in the original [paper](https://arxiv.org/abs/2401.14112) contains a helpful visualization of the bit storage pattern.
5. Change tensor core MMA instructions for BF16 (see [MMA_FP16_M16N8K16](https://github.com/pytorch/ao/compare/main...tobiasvanderwerff:ao:fp6-bf16#diff-d095cc529407429f6f9c6736383cdc658d0b4e870fe9d27f64a36e2b79c72177R97) function). Also straightforward, since it only involves changing the input types of the MMA instructions.
6. Modified exponent bias calculations (see [MultScale](https://github.com/pytorch/ao/compare/main...tobiasvanderwerff:ao:fp6-bf16#diff-d1a5769dc17cd4ffca57d2a4106d9780d8de281654bdcc91e88d64b49e09284dR66) function). This turned out to be a little more involved and required a custom solution. See the section below for more details.

## More details on modified exponent bias calculations

Adapting the FP6 kernel for BF16 introduces a complication regarding the exponent bias in the dequantization step compared to the original implementation for FP16. This required me to come up with a custom solution. For context: in section 4.2.2 and 5.3.1 of the FP6 paper, they mention that the exponent for FP16 (and also equivalent for BF16) is 

$E^{\text{f⁢p⁢16}} = E^{\text{f⁢p⁢x}} + \text{b⁢i⁢a⁢s}^{\text{f⁢p⁢16}} − \text{b⁢i⁢a⁢s}^{\text{f⁢p⁢x}}$

Adding the constant bias terms is computationally expensive to do during dequantization, so instead they set $E^{\text{f⁢p⁢16}} = E^{\text{f⁢p⁢x}}$ during dequantization, followed by multiplication later on with $2^{\text{b⁢i⁢a⁢s}^{\text{f⁢p⁢16}} − \text{b⁢i⁢a⁢s}^{\text{f⁢p⁢x}}}$ to get the equivalent result in a more efficient way. This works fine for FP16, since $\text{b⁢i⁢a⁢s}^{\text{fp⁢16}} = 15$ and so $2^{\text{b⁢i⁢a⁢s}^{\text{f⁢p⁢16}} − \text{b⁢i⁢a⁢s}^{\text{f⁢p⁢x}}} = 2^{15-3} = 2^{12}$ for FP6, which fits into a single 32-bit integer value.

Translating this to BF16 however, is a little more tricky. Since $\text{b⁢i⁢a⁢s}^{\text{bf⁢16}} = 127$ (see [Wikipedia](https://en.wikipedia.org/wiki/Bfloat16_floating-point_format)), this would amount to multiplication by $2^{\text{b⁢i⁢a⁢s}^{\text{bf⁢16}} − \text{b⁢i⁢a⁢s}^{\text{f⁢p⁢x}}} = 2^{127-3} = 2^{124}$, which is too large to fit into a 32-bit or even 64-bit number. To address this, I experimented with a few approaches, which I highlight below.

### Approach 1: Type punning (current choice)

EDIT: The solution below is now simplified by using the CUDA [ldexpf](https://docs.nvidia.com/cuda/cuda-math-api/cuda_math_api/group__CUDA__MATH__SINGLE.html#_CPPv46ldexpffi) function. Thanks to @gau-nernst for pointing me in the right direction!

$2^{124}$ is too large to fit into an integer, but it can fit into a 32-bit floating point number. Since the exponent for IEEE 754 floating point has 8 bits, it can represent a maximum exponent of $2^{127}$ (see [Wikipedia](https://en.wikipedia.org/wiki/Single-precision_floating-point_format)). So, we can represent $2^{124}$ by directly setting the exponent bits and constructing a float, like so:

```cpp
union {
    uint32_t u32;
    float f;
} tmp;
tmp.u32 = (124 + 127) << 23;  // 127=exponent bias, 23=mantissa
// tmp.f now contains a float with sign=0, mantissa=0, exponent=124, 
// which translates to the desired floating point value of 2^124
__nv_bfloat16 result = __hmul(*BF16_1,__float2bfloat16(tmp.f));
```

### Approach 2: Decompose the exponent bias into smaller values

A second approach is based on the insight that instead of multiplying by $2^{124}$, we can notice that $2^{124} = 2^{31} \cdot 2^{31} \cdot 2^{31} \cdot 2^{31}$. Since $2^{31}$ does fit into an ordinary 32-bit unsigned int, we can multiply 4 times by this value, like so:

```cpp
__nv_bfloat16 tmp1 = *BF16_1;
const __nv_bfloat16 BIAS = __float2bfloat16(1.0f * (uint32_t(1) << (124 / 4)));
#pragma unroll
for (int i = 0; i < 4; i++) {
    tmp1 = __hmul(tmp1, BIAS);
}
__nv_bfloat16 result = tmp1;
```

This approach works, but I think it is less preferrable to approach 1 since it introduces 3 additional multiplications.

## Benchmark

Tested on:
- A100 GPU
- Torch version: `2.6.0.dev20241022`
- CUDA version: 12.2

|   m |     k |     n |   fp6-fp16 latency (ms) |   fp16 latency (ms) |   speedup fp16 |   correct fp16 |   fp6-bf16 latency (ms) |   bf16 latency (ms) |   speedup bf16 |   correct bf16 |
|----:|------:|------:|------------------------:|--------------------:|---------------:|---------------:|------------------------:|--------------------:|---------------:|---------------:|
|   1 |  8192 |  8192 |                 47.8153 |             107.905 |       2.25671  |              1 |                 48.0822 |             108.038 |       2.24693  |              1 |
|   1 |  8192 | 10240 |                 56.5186 |             129.677 |       2.29441  |              1 |                 56.7016 |             129.925 |       2.29138  |              1 |
|   1 |  8192 | 57344 |                261.31   |             679.021 |       2.59853  |              1 |                261.341  |             681.108 |       2.60621  |              1 |
|   1 | 28672 |  8192 |                142.66   |             340.947 |       2.38992  |              1 |                142.768  |             341.311 |       2.39066  |              1 |
|   2 |  8192 |  8192 |                 48.1587 |             108.817 |       2.25954  |              1 |                 48.5589 |             108.727 |       2.23908  |              1 |
|   2 |  8192 | 10240 |                 56.8241 |             133.974 |       2.35769  |              1 |                 57.2074 |             134.379 |       2.34898  |              1 |
|   2 |  8192 | 57344 |                264.475  |             684.25  |       2.58721  |              1 |                264.451  |             683.928 |       2.58622  |              1 |
|   2 | 28672 |  8192 |                143.41   |             338.273 |       2.35878  |              1 |                143.463  |             338.418 |       2.35893  |              1 |
|   4 |  8192 |  8192 |                 49.0373 |             108.863 |       2.22     |              1 |                 49.3801 |             108.82  |       2.20372  |              1 |
|   4 |  8192 | 10240 |                 57.9573 |             134.769 |       2.32532  |              1 |                 58.1178 |             135.255 |       2.32726  |              1 |
|   4 |  8192 | 57344 |                269.995  |             688.162 |       2.5488   |              1 |                270.09   |             687.903 |       2.54694  |              1 |
|   4 | 28672 |  8192 |                144.317  |             339.184 |       2.35027  |              1 |                144.298  |             339.239 |       2.35096  |              1 |
|   8 |  8192 |  8192 |                 50.7631 |             109.538 |       2.15783  |              1 |                 50.9435 |             108.694 |       2.13361  |              1 |
|   8 |  8192 | 10240 |                 59.6324 |             136.478 |       2.28866  |              1 |                 59.8107 |             136.604 |       2.28394  |              1 |
|   8 |  8192 | 57344 |                281.025  |             694.71  |       2.47206  |              1 |                281.094  |             694.339 |       2.47013  |              1 |
|   8 | 28672 |  8192 |                146.231  |             340.169 |       2.32625  |              1 |                146.197  |             340.209 |       2.32707  |              1 |
|  16 |  8192 |  8192 |                 54.6856 |             111.104 |       2.0317   |              1 |                 54.8701 |             110.281 |       2.00986  |              1 |
|  16 |  8192 | 10240 |                 63.8146 |             139.737 |       2.18973  |              1 |                 63.9122 |             139.169 |       2.1775   |              1 |
|  16 |  8192 | 57344 |                310.406  |             705.661 |       2.27335  |              1 |                310.524  |             705.344 |       2.27147  |              1 |
|  16 | 28672 |  8192 |                153.435  |             342.533 |       2.23243  |              1 |                153.635  |             342.03  |       2.22625  |              1 |
|  32 |  8192 |  8192 |                 59.8124 |             109.968 |       1.83855  |              1 |                 60.0498 |             117.015 |       1.94863  |              1 |
|  32 |  8192 | 10240 |                 70.0894 |             137.461 |       1.96122  |              1 |                 70.0611 |             134.311 |       1.91706  |              1 |
|  32 |  8192 | 57344 |                361.694  |             713.476 |       1.97259  |              1 |                361.379  |             747.143 |       2.06748  |              1 |
|  32 | 28672 |  8192 |                158.193  |             352.667 |       2.22935  |              1 |                159.116  |             365.284 |       2.29571  |              1 |
|  64 |  8192 |  8192 |                 75.025  |             111.396 |       1.48478  |              1 |                 75.3286 |             116.787 |       1.55037  |              1 |
|  64 |  8192 | 10240 |                 86.4957 |             133.782 |       1.54669  |              1 |                 86.5149 |             137.81  |       1.59291  |              1 |
|  64 |  8192 | 57344 |                496.407  |             717.28  |       1.44494  |              1 |                495.186  |             848.441 |       1.71338  |              1 |
|  64 | 28672 |  8192 |                198.356  |             363.766 |       1.8339   |              1 |                198.211  |             378.897 |       1.91158  |              1 |
| 128 |  8192 |  8192 |                116.348  |             134.703 |       1.15776  |              1 |                115.112  |             136.111 |       1.18242  |              1 |
| 128 |  8192 | 10240 |                152.802  |             162.002 |       1.06021  |              1 |                151.848  |             144.769 |       0.953382 |              1 |
| 128 |  8192 | 57344 |                829.325  |             813.571 |       0.981003 |              1 |                821.493  |             860.491 |       1.04747  |              1 |
| 128 | 28672 |  8192 |                363.332  |             379.417 |       1.04427  |              1 |                358.096  |             377.712 |       1.05478  |              1 |
| 256 |  8192 |  8192 |                232.587  |             176.013 |       0.756765 |              1 |                232.328  |             178.489 |       0.768264 |              1 |
| 256 |  8192 | 10240 |                278.047  |             225.67  |       0.811624 |              1 |                275.367  |             215.422 |       0.78231  |              1 |
| 256 |  8192 | 57344 |               1517.54   |            1099.87  |       0.72477  |              1 |               1498.89   |            1092.88  |       0.729125 |              1 |
| 256 | 28672 |  8192 |                676.377  |             526.53  |       0.778457 |              1 |                667.652  |             536.409 |       0.803426 |              1 |
| 512 |  8192 |  8192 |                455.454  |             338.41  |       0.743016 |              1 |                451.99   |             328.303 |       0.726351 |              1 |
| 512 |  8192 | 10240 |                496.543  |             363.535 |       0.732133 |              1 |                492.433  |             410.334 |       0.833279 |              1 |
| 512 |  8192 | 57344 |               2528.04   |            2005.78  |       0.793411 |              1 |               2530.73   |            1963.43  |       0.775835 |              1 |
| 512 | 28672 |  8192 |               1338.7    |            1013.8   |       0.757304 |              1 |               1321.18   |             983.968 |       0.744764 |              1 |

## Final remarks

- I've added guards that should still make the kernel compatible with SM75, but I have not confirmed this because I was testing on an SM80 GPU.
- I've only tested this for FP6, but I think it should most likely also work for the other currently implemented FPx types.

CC @gau-nernst

### DIFF
--- a/benchmarks/benchmark_fp6.py
+++ b/benchmarks/benchmark_fp6.py
@@ -7,12 +7,13 @@ from torchao.utils import benchmark_torch_function_in_microseconds
 from tqdm import tqdm
 
 
-def benchmark(m: int, k: int, n: int):
-    float_data = torch.randn(n, k, dtype=torch.half, device="cuda")
+def benchmark(m: int, k: int, n: int, use_bf16=False):
+    dtype = torch.bfloat16 if use_bf16 else torch.half
+    float_data = torch.randn(n, k, dtype=dtype, device="cuda")
     fp6_weight = to_affine_quantized_fpx(float_data, FloatxTensorCoreLayout(3, 2))
-    fp16_weight = fp6_weight.dequantize(torch.half)
+    fp16_weight = fp6_weight.dequantize(dtype)
 
-    fp16_act = torch.randn(m, k, dtype=torch.half, device="cuda")
+    fp16_act = torch.randn(m, k, dtype=dtype, device="cuda")
     fp6_output = F.linear(fp16_act, fp6_weight)
     fp16_output = F.linear(fp16_act, fp16_weight)
 
@@ -28,7 +29,7 @@ def benchmark(m: int, k: int, n: int):
         "k": k,
         "n": n,
         "fp6_latency (ms)": fp6_time,
-        "fp16_latency (ms)": fp16_time,
+        f"{'bf16' if use_bf16 else 'fp16'}_latency (ms)": fp16_time,
         "speedup (d/s)": fp16_time / fp6_time,
         "correct": correct,
     }
@@ -39,11 +40,13 @@ if __name__ == "__main__":
     k_vals = (8192, 8192, 8192, 28672)
     n_vals = (8192, 10240, 57344, 8192)
 
+    use_bf16 = True
+
     results = []
 
     for m in tqdm([1 << i for i in range(10)]):
         for n, k in zip(n_vals, k_vals):
-            results.append(benchmark(m, k, n))
+            results.append(benchmark(m, k, n, use_bf16=use_bf16))
 
     df = pd.DataFrame(results)
     df.to_csv("fp6_llm_benchmark_results.csv", index=False)
--- a/benchmarks/benchmark_fp6.py
+++ b/benchmarks/benchmark_fp6.py
@@ -54,7 +54,7 @@ if __name__ == "__main__":
 
     results = []
 
-    for m in tqdm([1 << i for i in range(5)]):  # TODO: reset to 10
+    for m in tqdm([1 << i for i in range(10)]):
         for n, k in zip(n_vals, k_vals):
             results.append(benchmark(m, k, n))
 
--- a/benchmarks/benchmark_fp6.py
+++ b/benchmarks/benchmark_fp6.py
@@ -7,31 +7,43 @@ from torchao.utils import benchmark_torch_function_in_microseconds
 from tqdm import tqdm
 
 
-def benchmark(m: int, k: int, n: int, use_bf16=False):
-    dtype = torch.bfloat16 if use_bf16 else torch.half
-    float_data = torch.randn(n, k, dtype=dtype, device="cuda")
-    fp6_weight = to_affine_quantized_fpx(float_data, FloatxTensorCoreLayout(3, 2))
-    fp16_weight = fp6_weight.dequantize(dtype)
+def benchmark(m: int, k: int, n: int):
+    float_data_fp16 = torch.randn(n, k, dtype=torch.float16, device="cuda")
+    float_data_bf16 = torch.randn(n, k, dtype=torch.bfloat16, device="cuda")
+    fp6_weight_fp16 = to_affine_quantized_fpx(float_data_fp16, FloatxTensorCoreLayout(3, 2))
+    fp6_weight_bf16 = to_affine_quantized_fpx(float_data_bf16, FloatxTensorCoreLayout(3, 2))
+    fp16_weight = fp6_weight_fp16.dequantize(torch.float16)
+    bf16_weight = fp6_weight_bf16.dequantize(torch.bfloat16)
 
-    fp16_act = torch.randn(m, k, dtype=dtype, device="cuda")
-    fp6_output = F.linear(fp16_act, fp6_weight)
+    fp16_act = torch.randn(m, k, dtype=torch.float16, device="cuda")
+    bf16_act = fp16_act.to(torch.bfloat16)
+    fp6_output_fp16 = F.linear(fp16_act, fp6_weight_fp16)
+    fp6_output_bf16 = F.linear(bf16_act, fp6_weight_bf16)
     fp16_output = F.linear(fp16_act, fp16_weight)
+    bf16_output = F.linear(bf16_act, bf16_weight)
 
-    fp6_time = benchmark_torch_function_in_microseconds(F.linear, fp16_act, fp6_weight)
     fp16_time = benchmark_torch_function_in_microseconds(F.linear, fp16_act, fp16_weight)
+    bf16_time = benchmark_torch_function_in_microseconds(F.linear, bf16_act, bf16_weight)
+    fp6_time_fp16 = benchmark_torch_function_in_microseconds(F.linear, fp16_act, fp6_weight_fp16)
+    fp6_time_bf16 = benchmark_torch_function_in_microseconds(F.linear, bf16_act, fp6_weight_bf16)
 
     # follow https://github.com/usyd-fsalab/fp6_llm/blob/ce76774bcfc26b325c1b558abcf1935026d9abbc/tests/python/kernel_test.py
     # doesn't seem to be the right way to check for correctness
-    correct = (fp6_output - fp16_output).abs().mean() / fp16_output.abs().mean() < 1e-3
+    correct_fp16 = (fp6_output_fp16 - fp16_output).abs().mean() / fp16_output.abs().mean() < 1e-3
+    correct_bf16 = (fp6_output_bf16 - bf16_output).abs().mean() / bf16_output.abs().mean() < 1e-2
 
     return {
         "m": m,
         "k": k,
         "n": n,
-        "fp6_latency (ms)": fp6_time,
-        f"{'bf16' if use_bf16 else 'fp16'}_latency (ms)": fp16_time,
-        "speedup (d/s)": fp16_time / fp6_time,
-        "correct": correct,
+        "fp6-fp16 latency (ms)": fp6_time_fp16,
+        "fp16 latency (ms)": fp16_time,
+        "speedup fp16": fp16_time / fp6_time_fp16,
+        "correct fp16": correct_fp16,
+        "fp6-bf16 latency (ms)": fp6_time_bf16,
+        "bf16 latency (ms)": bf16_time,
+        "speedup bf16": bf16_time / fp6_time_bf16,
+        "correct bf16": correct_bf16,
     }
 
 
@@ -40,13 +52,11 @@ if __name__ == "__main__":
     k_vals = (8192, 8192, 8192, 28672)
     n_vals = (8192, 10240, 57344, 8192)
 
-    use_bf16 = True
-
     results = []
 
-    for m in tqdm([1 << i for i in range(10)]):
+    for m in tqdm([1 << i for i in range(5)]):  # TODO: reset to 10
         for n, k in zip(n_vals, k_vals):
-            results.append(benchmark(m, k, n, use_bf16=use_bf16))
+            results.append(benchmark(m, k, n))
 
     df = pd.DataFrame(results)
     df.to_csv("fp6_llm_benchmark_results.csv", index=False)
--- a/torchao/csrc/cuda/fp6_llm/fp6_linear.cu
+++ b/torchao/csrc/cuda/fp6_llm/fp6_linear.cu
@@ -31,21 +31,6 @@
 #include <torch/library.h>
 
 
-inline bool isSM75GPU() {
-    int device;
-    cudaError_t err = cudaGetDevice(&device);
-    if (err != cudaSuccess) return false;
-
-    int major, minor;
-    err = cudaDeviceGetAttribute(&major, cudaDevAttrComputeCapabilityMajor, device);
-    if (err != cudaSuccess) return false;
-
-    err = cudaDeviceGetAttribute(&minor, cudaDevAttrComputeCapabilityMinor, device);
-    if (err != cudaSuccess) return false;
-
-    return (major == 7) && (minor == 5);
-}
-
 template<typename TilingConfig, typename InputDataType, typename OutputDataType, int EXPONENT, int MANTISSA>
 static void Kernel_Ex(cudaStream_t    stream,
                       const uint4     *Weight,
@@ -106,7 +91,23 @@ cudaError_t fpx_linear_kernel(cudaStream_t    stream,
     if(N_Global>64 && N_Global<=128)    N_PowerOf2 = 128;
     if(N_Global>128)                    N_PowerOf2 = ((N_Global-1)/128+1) * 128;
 
-    if (isSM75GPU() && (N_PowerOf2 == 64 || N_PowerOf2 == 128 || N_PowerOf2 % 128 == 0)) {
+    // Check GPU Compute Capability
+    int device, major, minor;
+    cudaError_t err = cudaGetDevice(&device);
+    if (err != cudaSuccess) return err;
+    err = cudaDeviceGetAttribute(&major, cudaDevAttrComputeCapabilityMajor, device);
+    if (err != cudaSuccess) return err;
+    err = cudaDeviceGetAttribute(&minor, cudaDevAttrComputeCapabilityMinor, device);
+    if (err != cudaSuccess) return err;
+
+    if ((major < 7) || (major == 7 && minor < 5)) {
+        printf("FP6LLM_API Error: FP6LLM requires GPU with SM75 or higher!\n");
+        return cudaErrorUnknown;
+    }
+
+    const bool is_sm75_gpu = (major == 7) && (minor == 5);
+
+    if (is_sm75_gpu && (N_PowerOf2 == 64 || N_PowerOf2 == 128 || N_PowerOf2 % 128 == 0)) {
         // For SM75 and N >= 64, we use a different TilingConfig to deal with smaller shared memory.
         if (Split_K == 1) {
             Kernel_Ex<TilingConfig<4, 1, 4>, InputDataType, InputDataType, EXPONENT, MANTISSA>(stream, Weight, Scales, B, C, M_Global, N_Global, K_Global, Split_K);
@@ -156,19 +157,6 @@ cudaError_t fpx_linear_kernel(cudaStream_t    stream,
 
 
 // https://github.com/NVIDIA/apex/blob/master/csrc/type_shim.h
-#if __CUDA_ARCH__ == 750
-#define DISPATCH_HALF_AND_BF16(TYPE, NAME, ...)                                \
-  switch (TYPE) {                                                              \
-  case at::ScalarType::Half: {                                                 \
-    using torch_t = at::Half;                                                  \
-    using nv_t = half;                                                         \
-    __VA_ARGS__();                                                             \
-    break;                                                                     \
-  }                                                                            \
-  default:                                                                     \
-    AT_ERROR(#NAME, " not implemented for '", toString(TYPE), "'");            \
-  }
-#else
 #define DISPATCH_HALF_AND_BF16(TYPE, NAME, ...)                                \
   switch (TYPE) {                                                              \
   case at::ScalarType::Half: {                                                 \
@@ -186,7 +174,6 @@ cudaError_t fpx_linear_kernel(cudaStream_t    stream,
   default:                                                                     \
     AT_ERROR(#NAME, " not implemented for '", toString(TYPE), "'");            \
   }
-#endif
 
 namespace torchao {
 // MODIFICATION NOTE: dtype of _weights is changed to uint8
--- a/torchao/csrc/cuda/fp6_llm/fp6_linear.cu
+++ b/torchao/csrc/cuda/fp6_llm/fp6_linear.cu
@@ -41,16 +41,16 @@ inline bool isSM75GPU() {
     return (major == 7) && (minor == 5);
 }
 
-template<typename TilingConfig, typename OutputDataType, int EXPONENT, int MANTISSA>
-static void Kernel_Ex(cudaStream_t    stream,
-                      const uint4     *Weight,
-                      const half      *Scales,
-                      const half      *B,
-                      OutputDataType  *C,
-                      const size_t    M_Global,
-                      const size_t    N_Global,
-                      const size_t    K_Global, 
-                      int             Split_K) 
+template<typename TilingConfig, typename InputDataType, typename OutputDataType, int EXPONENT, int MANTISSA>
+static void Kernel_Ex(cudaStream_t        stream,
+                      const uint4         *Weight,
+                      const InputDataType *Scales,
+                      const InputDataType *B,
+                      OutputDataType      *C,
+                      const size_t        M_Global,
+                      const size_t        N_Global,
+                      const size_t        K_Global, 
+                      int                 Split_K) 
 {
     #ifdef DEBUG_MODE
         printf("\n");
@@ -59,7 +59,7 @@ static void Kernel_Ex(cudaStream_t    stream,
         printf("TILE_M: %d, TILE_K: %d, TILE_N: %d\n", TilingConfig::TILE_M, TilingConfig::TILE_K, TilingConfig::TILE_N);
     #endif
     static size_t SHMEM_SZ = max(TilingConfig::SMEM_SIZE_B_TILE+SMEM_SIZE_PER_TB_A_TILE, TilingConfig::SMEM_SIZE_C_TILE);
-    cudaFuncSetAttribute(QUANT_GEMM_Kernel<TilingConfig, OutputDataType, EXPONENT, MANTISSA>, cudaFuncAttributeMaxDynamicSharedMemorySize, SHMEM_SZ);
+    cudaFuncSetAttribute(QUANT_GEMM_Kernel<TilingConfig, InputDataType, OutputDataType, EXPONENT, MANTISSA>, cudaFuncAttributeMaxDynamicSharedMemorySize, SHMEM_SZ);
     size_t  dimN = (N_Global-1) / TilingConfig::TILE_N + 1;
     size_t  dimM = M_Global * Split_K / TilingConfig::TILE_M;
     dim3    GridDim(dimN, dimM, 1);
@@ -70,22 +70,23 @@ static void Kernel_Ex(cudaStream_t    stream,
                 GridDim.x, GridDim.y, GridDim.z, BlockDim.x, BlockDim.y, BlockDim.z, SHMEM_SZ);
         printf("\n");
     #endif
-    QUANT_GEMM_Kernel<TilingConfig, OutputDataType, EXPONENT, MANTISSA><<<GridDim, BlockDim, SHMEM_SZ, stream>>>
+    QUANT_GEMM_Kernel<TilingConfig, InputDataType, OutputDataType, EXPONENT, MANTISSA><<<GridDim, BlockDim, SHMEM_SZ, stream>>>
                     (Weight, Scales, B, C, M_Global, N_Global, K_Global, Split_K);
 }
 
-template<int EXPONENT, int MANTISSA>
-cudaError_t fpx_linear_kernel(cudaStream_t    stream,
-                              const uint4     *Weight,
-                              const half      *Scales,
-                              const half      *B,
-                              half            *C,
-                              const size_t    M_Global,
-                              const size_t    N_Global,
-                              const size_t    K_Global, 
-                              float           *Reduction_Workspace,  // Reduction_Workspace_Size = Split_K * M_Global * N_Global * sizeof(fp32)
-                              int             Split_K)
+template<typename InputDataType, int EXPONENT, int MANTISSA>
+cudaError_t fpx_linear_kernel(cudaStream_t        stream,
+                              const uint4         *Weight,
+                              const InputDataType *Scales,
+                              const InputDataType *B,
+                              InputDataType       *C,
+                              const size_t        M_Global,
+                              const size_t        N_Global,
+                              const size_t        K_Global, 
+                              float               *Reduction_Workspace,  // Reduction_Workspace_Size = Split_K * M_Global * N_Global * sizeof(fp32)
+                              int                 Split_K)
 {
+    static_assert(std::is_same<InputDataType, half>::value || std::is_same<InputDataType, __nv_bfloat16>::value, "Type must be float or __nv_bfloat16");
     assert(M_Global % 256 == 0);
     assert(K_Global % 64 == 0);
     assert(N_Global>0);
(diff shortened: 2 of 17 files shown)


torchao/csrc/cuda/fp6_llm/fp6_linear.cu
```diff
@@ -153,6 +154,26 @@ cudaError_t fpx_linear_kernel(cudaStream_t    stream,
...
+    using nv_t = half;                                                         \
+    __VA_ARGS__();                                                             \
+    break;                                                                     \
+  }                                                                            \
+  case at::ScalarType::BFloat16: {                                             \
+    using torch_t = at::BFloat16;                                              \
+    using nv_t = __nv_bfloat16;                                                \
+    __VA_ARGS__();                                                             \
+    break;                                                                     \
+  }                                                                            \
```
gau-nernst, line 172: If we want to keep sm75 support, I think we need to be careful whenever we use `__nv_bfloat16`. Might need to add more `#ifdef`.

torchao/csrc/cuda/fp6_llm/kernel_reduction.cuh
```diff
@@ -36,17 +36,22 @@
...
 {
-    half*  WARP_GPTR_C      = C                    + REDUCTION_ELEMENT_PER_THREADBLOCK * blockIdx.x;
-    float* WARP_GPTR_R      = Reduction_Workspace  + REDUCTION_ELEMENT_PER_THREADBLOCK * blockIdx.x;
-    half*  THREAD_GPTR_C    = WARP_GPTR_C          + threadIdx.x * HALF_PER_128BIT;    
-    float* THREAD_GPTR_R    = WARP_GPTR_R          + threadIdx.x * HALF_PER_128BIT;
+    static_assert(std::is_same<T, half>::value || std::is_same<T, __nv_bfloat16>::value, "Type must be float or __nv_bfloat16");
+    T*  WARP_GPTR_C      = C                    + REDUCTION_ELEMENT_PER_THREADBLOCK * blockIdx.x;
+    float* WARP_GPTR_R   = Reduction_Workspace  + REDUCTION_ELEMENT_PER_THREADBLOCK * blockIdx.x;
+    T*  THREAD_GPTR_C    = WARP_GPTR_C          + threadIdx.x * HALF_PER_128BIT;    
+    float* THREAD_GPTR_R = WARP_GPTR_R          + threadIdx.x * HALF_PER_128BIT;
```
gau-nernst, line 54: Minor: align the variable names

torchao/csrc/cuda/fp6_llm/ptx_mma.cuh
```diff
@@ -43,9 +46,9 @@
 // MODIFICATION NOTE: to support MSVC
 // - uint32_t __restrict__ Reg[][4] is changed to uint32_t (* __restrict__ Reg)[4]
 // - half __restrict__ (*read_SPTR) is changed to half (* __restrict__ read_SPTR)
-template <typename TilingConfig>
+template <typename T, typename TilingConfig>
 __device__ __forceinline__ void B_FromSharedToReg(uint32_t (* __restrict__ Reg)[4],
```
gau-nernst, line 50: Same as another comment. Do we need to care if T is half or bfloat16? (but this time the input argument is typed so maybe yes)

torchao/csrc/cuda/fp6_llm/utils_parallel_dequant.cuh
```diff
@@ -20,19 +20,23 @@
 
 #include <cuda.h>
 #include <cuda_fp16.h>
+// TODO: can cuda_bf16 be imported for SM75? How to guard against this? The guard below does not work outside of device code
```
gau-nernst, line 23: Where is bf16 used in host code?

torchao/csrc/cuda/fp6_llm/utils_parallel_dequant.cuh
```diff
@@ -58,10 +62,48 @@ __device__ __forceinline__ uint32_t MultScale(uint32_t PackedFP16Pair, half Scal
     return output;
 }
 
+template<int EXPONENT, int MANTISSA>
+__device__ __forceinline__ uint32_t MultScale(uint32_t PackedBF16Pair, __nv_bfloat16 Scale) {
```
gau-nernst, line 66: Will this help? https://docs.nvidia.com/cuda/cuda-math-api/cuda_math_api/group__CUDA__MATH__SINGLE.html#_CPPv46ldexpffi

gau-nernst (commented):
Awesome work! Leave some comments for discussion.

torchao/csrc/cuda/fp6_llm/utils_gmem.cuh
```diff
@@ -29,40 +29,41 @@
  * Copying A1/A2 from global memory to shared memory.
  * Usually 1024 or 2048 Bytes
  */
-template<int SMEM_SIZE_IN_BYTES_PER_WARP>
+template<typename T, int SMEM_SIZE_IN_BYTES_PER_WARP>
 __device__ __forceinline__ void CopyFromGlobalToShared_A(uint32_t* SPTR,
```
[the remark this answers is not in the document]
tobiasvanderwerff: Good point, if the goal is to minimize changes then there are a bunch of places where I can just use 'half' type for copy operations instead of 'T'.

torchao/csrc/cuda/fp6_llm/utils_parallel_dequant.cuh
tobiasvanderwerff: Thanks! This is definitely a less hacky solution :sweat_smile:

torchao/csrc/cuda/fp6_llm/utils_parallel_dequant.cuh
tobiasvanderwerff: I'll try to hack around a bit more to see if I can make it work.

torchao/csrc/cuda/fp6_llm/utils_parallel_dequant.cuh
gau-nernst: I was actually thinking about `ldexpf(*BF16_1, BIAS_OFFSET)` (`BF16_1` will be upcast to FP32). Then realize the original code does multiplication in FP16, which is a bit strange to me - not sure if it's faster (I have a feeling the GPU will use FP32 math to calculate it anyway). Anyway, the current change is fine!

torchao/csrc/cuda/fp6_llm/utils_parallel_dequant.cuh
tobiasvanderwerff: It should work now. I think the problem was that `#include <cuda_bf16.h>` actually needs to be there for host code as well, so I changed it to `#if !defined(__CUDA_ARCH__) || __CUDA_ARCH__ >= 800`

tobiasvanderwerff:
Here are some benchmark results on Llama-2-7b-chat-hf.

Tested on:
- A100 GPU
- Torch version: `2.6.0.dev20241023`
- CUDA version: 12.2

Wikitext perplexity (using `torchao/_models/llama/eval.py`):

```
fp6-float16:  12.3648
fp6-bfloat16: 12.3669
```

**FP16 perf.** 

```shell
python generate.py --compile --quantization fp6 --precision float16
```
Average tokens/sec: 111.55
Average Bandwidth: 553.13 GB/s
Peak Memory Usage: 6.69 GB
Model Size: 4.96 GB


**BF16 perf.**

```shell
python generate.py --compile --quantization fp6 --precision bfloat16
```

Average tokens/sec: 111.19
Average Bandwidth: 551.35 GB/s
Peak Memory Usage: 6.69 GB
Model Size: 4.96 GB

torchao/csrc/cuda/fp6_llm/utils_parallel_dequant.cuh
gau-nernst: Random thoughts. `float bias` can be made into a `constexpr` (in fact it can be a BF16 constexpr if you only do BF16 mul later). But I don't think there is a simple way apart from a big if-else clause (`ldexpf` cannot be used in a constexpr I think). Don't mind me, just some random thoughts on useless micro-optimizations :rofl:

benchmarks/benchmark_fp6.py
```diff
@@ -8,29 +8,42 @@
...
     fp16_time = benchmark_torch_function_in_microseconds(F.linear, fp16_act, fp16_weight)
+    bf16_time = benchmark_torch_function_in_microseconds(F.linear, bf16_act, bf16_weight)
+    fp6_time_fp16 = benchmark_torch_function_in_microseconds(F.linear, fp16_act, fp6_weight_fp16)
+    fp6_time_bf16 = benchmark_torch_function_in_microseconds(F.linear, bf16_act, fp6_weight_bf16)
 
     # follow https://github.com/usyd-fsalab/fp6_llm/blob/ce76774bcfc26b325c1b558abcf1935026d9abbc/tests/python/kernel_test.py
     # doesn't seem to be the right way to check for correctness
-    correct = (fp6_output - fp16_output).abs().mean() / fp16_output.abs().mean() < 1e-3
+    correct_fp16 = (fp6_output_fp16 - fp16_output).abs().mean() / fp16_output.abs().mean() < 1e-3
+    correct_bf16 = (fp6_output_bf16 - bf16_output).abs().mean() / bf16_output.abs().mean() < 1e-2
```
gau-nernst, line 33: Just curious. I saw that generally when BF16 is used, tolerance is quite higher than FP16. From your experience working on this, you do suspect any part of the code might result in this loss of precision? e.g. perhaps some parts are computed in BF16 instead of FP32. Or maybe it's just the way it is.

gau-nernst (approved):
Thank you for the great work! I will request Mark and Charles to take another scan at the PR.

Edit: almost forgot. @tobiasvanderwerff Can you update the docs also? Mainly this one I think https://github.com/pytorch/ao/tree/main/torchao/dtypes/floatx. You can mention that you extended the original kernel to work with BF16 :smile:. I think there might be other places where we mentioned FPx kernel only works with FP16.

torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -1636,15 +1636,15 @@ def _linear_f16_act_floatx_weight_check(input_tensor, weight_tensor, bias):
...
+def _linear_f16_bf16_act_floatx_weight_impl(input_tensor, weight_tensor, bias):
     from torchao.dtypes.floatx import _SPLIT_K_MAP
     from torchao.ops import quant_llm_linear
 
     act = input_tensor
     weight = weight_tensor
 
     out_dim, in_dim = weight.shape
-    act_reshaped = act.view(-1, in_dim).half()
+    act_reshaped = act.view(-1, in_dim)
```
msaroufim, line 1647: no need for a cast anymore?

torchao/csrc/cuda/fp6_llm/utils_core.cuh
```diff
@@ -98,13 +98,13 @@ __device__ __forceinline__ void core_mma_slice(float                     c[][REG
     #pragma unroll
     for (int i = 0; i < WARP_ROW_MMA_TENSORS; i++) {
         if(TilingConfig::WARP_COL_MMA_TENSORS==1) {
-            MMA_FP16_M16N8K16( c_uint_ptr[i], a_read[i], b_read[0] );
+            MMA_FP16_M16N8K16<USE_BF16>( c_uint_ptr[i], a_read[i], b_read[0] );
```
msaroufim, line 101: more of a nit: but USE_BF16 is now being piped through the entire codebase when that feels like it can be inferred from the type of an argument instead

msaroufim (approved):
+1 to adding docs, left a few random comments - this is close to getting landed

msaroufim (changes requested):
+1 to adding docs, left a few small comments - this is close to getting landed

torchao/csrc/cuda/fp6_llm/fp6_linear.cu
```diff
@@ -153,6 +156,39 @@ cudaError_t fpx_linear_kernel(cudaStream_t    stream,
 #include <ATen/cuda/CUDAContext.h>
 #include <torch/library.h>
 
+// https://github.com/NVIDIA/apex/blob/master/csrc/type_shim.h
+#if __CUDA_ARCH__ == 750
```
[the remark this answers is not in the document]
gau-nernst: We don't support <sm75 so equal is fine I think. Another look at this, maybe we can add this before BF16 case instead of having 2 `DISPATCH_HALF_AND_BF16`? something like

```C++
#if __CUDA_ARCH__ >= 800  // do we need to add ifdef ?
  case at::ScalarType::BFloat16: {                                             \
    using torch_t = at::BFloat16;                                              \
    using nv_t = __nv_bfloat16;                                                \
    __VA_ARGS__();                                                             \
    break;                                                                     \
  } 
#endif
```

torchao/dtypes/affine_quantized_tensor.py
tobiasvanderwerff: Is `input_tensor` allowed to be something other than float16 or bfloat16? I assumed it wasn't, so I thought that the input is always either float16 or bfloat16, making casting unnecessary. If we do cast here, how we would know whether to cast to float16 or bfloat16?

torchao/csrc/cuda/fp6_llm/fp6_linear.cu
```diff
@@ -70,22 +70,25 @@ static void Kernel_Ex(cudaStream_t    stream,
...
-                              half            *C,
+                              InputDataType   *C,
                               const size_t    M_Global,
                               const size_t    N_Global,
                               const size_t    K_Global, 
                               float           *Reduction_Workspace,  // Reduction_Workspace_Size = Split_K * M_Global * N_Global * sizeof(fp32)
                               int             Split_K)
 {
+    #if !defined(__CUDA_ARCH__) || __CUDA_ARCH__ >= 800
+    static_assert(std::is_same<InputDataType, half>::value || std::is_same<InputDataType, __nv_bfloat16>::value, "Type must be float or __nv_bfloat16");
```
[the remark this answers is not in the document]
tobiasvanderwerff: Yes!

torchao/csrc/cuda/fp6_llm/utils_core.cuh
tobiasvanderwerff: I get your point, but I'm not sure this can be avoided. The point is that for many of the functions (such as `core_mma_slice` shown above), the array input types don't indicate whether float16 or bfloat16 is used (e.g. using generic `uint32_t` types). If we derive the type from the input type, then the input type needs to be templated (which is what I started out with), but leads to a similar effect where you have to pipe the template type through the entire codebase. I've made an effort to only pass this `USE_BF16` parameter when it's strictly necessary.

torchao/csrc/cuda/fp6_llm/fp6_linear.cu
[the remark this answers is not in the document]
tobiasvanderwerff: Ok on second look I think this check for `__CUDA_ARCH__` is actually just flat out wrong here. It's executed as host code, which means `__CUDA_ARCH__` will not be defined and so `#if __CUDA_ARCH__ == 750` will always be false, even when a sm_75 GPU is used. I verified this by checking for `#if __CUDA_ARCH__ >= 800` to define the macro, and that fails on a sm_80 GPU. A better solution would be to check for `__CUDA_ARCH__` further downstream in device code.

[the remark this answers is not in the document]
tobiasvanderwerff: In fact, I think any check for `__CUDA_ARCH__` in `fp6_linear.cu` will always fail because  `__CUDA_ARCH__` is undefined since all of the functions in `fp6_linear.cu` are host functions. So that means that the following guard (which was already there from previous iterations of the kernel) is also redundant since `__CUDA_ARCH__` is never defined here:

https://github.com/pytorch/ao/blob/aa86005df445ba1221523a3edc31f869a7da4f1c/torchao/csrc/cuda/fp6_llm/fp6_linear.cu#L21

E.g. if I change this guard to something that should always fail if `__CUDA_ARCH__` is defined:
```cpp
#if !defined(__CUDA_ARCH__) || __CUDA_ARCH__ >= 99999
```
then it still evaluates to true on an A100 sm_80 GPU.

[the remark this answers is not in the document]
tobiasvanderwerff: I'm removing these guards for now, but it would be a good idea to also test this on sm75 to confirm that nothing unexpected happens. I might have time to test this on a Lightning Studio T4 in the next few days.

tobiasvanderwerff:
@msaroufim @gau-nernst I updated the docs at https://github.com/pytorch/ao/tree/main/torchao/dtypes/floatx. I didn't update the benchmark results with BF16 however, because I don't have access to the same machine that the existing benchmarks results used. I'm a bit short on time today so I may have missed some additional docs that should be updated -- I'll double check this later. I also intend to do a sm75 test because that would be good to check I think.

torchao/csrc/cuda/fp6_llm/fp6_linear.cu
```diff
@@ -18,8 +18,6 @@
 // - Modified the TilingConfig parameters for SM75 to deal with smaller shared memory
 //
 
-#if !defined(__CUDA_ARCH__) || __CUDA_ARCH__ >= 750  // at least Turing
```
gau-nernst, line 21: I think we still need to keep this. Although it seems like this file only contains host code, `fp6_linear.cu` will include `kernel_matmul.cuh` and `kernel_reduction.cuh` (which in turns include more header files), which has device code. I rmb adding this line in #636.

So what I think is happening on old devices (CUDA_ARCH < 7.5)
1. Host code is still compiled (in fact it is always compiled) e.g. `fpx_linear_kernel()`
2. Device code will not be compiled e.g. `CopyFromGlobalToShared_A()`
3. When trying to run the host function (`fpx_linear_kernel()`), you will get undefined symbol errors, which are the device code functions that were not compiled.

Without this guard, when you try to compile this file on CUDA_ARCH < 7.5, you will get undefined nv_bfloat16 symbol (or something along that line) I think. Didn't check my hypothesis, but that's what I'm suspecting is happening.

torchao/csrc/cuda/fp6_llm/fp6_linear.cu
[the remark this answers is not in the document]
gau-nernst: I was just suggesting, didn't know if nested macro will work or not.

It seems like `cuda_bf16.h` is always available on host code, even for sm<75. So for `DISPATCH_HALF_AND_BF16`, we might not need to care about CUDA_ARCH at all?

A mini test by me on google colab

image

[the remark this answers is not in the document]
tobiasvanderwerff: Thanks a lot for testing this @gau-nernst. The actual usage of the __nv_bfloat16 type is still guarded by `__CUDA_ARCH__`, so this means I can probably remove the guards for the `cuda_bf16.h` include as well as the `DISPATCH_HALF_AND_BF16` macro. I'll do a sm75 test to confirm.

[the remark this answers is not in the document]
tobiasvanderwerff: Correction: It does seem to be necessary to guard against sm<=75 on device code, e.g. as done in the [Pytorch codebase](https://github.com/pytorch/pytorch/blob/ceab24def4605019666fc6d9aba7df6724732784/aten/src/ATen/native/cuda/KernelUtils.cuh#L4). But on host code it should be okay, like you tested.

torchao/csrc/cuda/fp6_llm/utils_parallel_dequant.cuh
tobiasvanderwerff: I'm all for micro-optimizations :). The `bias` is now a `constexpr`:

```cpp
constexpr float power_of_two(int n) {
    return (n == 0) ? 1.0f : 2.0f * power_of_two(n - 1);
}
...
{
    constexpr float BIAS = power_of_two(BIAS_OFFSET);
}
```

I ran the benchmark and I can report that this makes absolutely no difference for performance :smile:

benchmarks/benchmark_fp6.py
tobiasvanderwerff: All I know is that BF16 has fewer bits for the fraction (mantissa) than FP16 (10 bits vs. 7 bits), so that leads to lower precision for BF16 compared to FP16. I can't think of any part of the FP6 kernel that would inherently lead to more loss of precision for BF16.

torchao/csrc/cuda/fp6_llm/fp6_linear.cu
tobiasvanderwerff: I just tested on SM75 and set the guard at the top of `fp6_linear.cu` to `#if !defined(__CUDA_ARCH__) || __CUDA_ARCH__ >= 800` specifically to test what would happen for an unsupported version (after verifying that the kernel does work correctly for SM75, see results below). The expected outcome would be that the code throws an error complaining about undefined symbols, but surprisingly, the code still compiles and runs. Earlier, this led me to believe that the guard does not do anything, but this doesn't seem to be what's happening. You're right that the guard does have an effect, but instead of failing, the kernel then returns incorrect results and has a constant runtime no matter the input shape (see results below). This seems to indicate that the kernel returns silently for unsupported SM versions instead of the expected outcome of erroring out.

Would you agree this is problematic? E.g. if someone compiles for SM70 and runs FP6 than the code will still work, but it will return garbage outputs. From this perspective, the guard actually seems to do more harm than good, since it gives the impression that the code runs correctly, even though it might not. So then leaving out the guard would at least catch some arch-specific errors (e.g. `__nv_bfloat16` not being defined).

A possible other solution would be to leave the guard in and check for sm>=75 at runtime (e.g. in the Python code that calls the FP6 kernel). Then the codebase can still compile for older architectures, while still making sure that the FP6 kernel cannot be executed. What do you think?

**Case 1**
```cpp
#if !defined(__CUDA_ARCH__) || __CUDA_ARCH__ >= 750  // tested on sm75
```
image


**Case 2**
```cpp
#if !defined(__CUDA_ARCH__) || __CUDA_ARCH__ >= 800  // tested on sm75
```
image

torchao/csrc/cuda/fp6_llm/fp6_linear.cu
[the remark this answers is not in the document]
tobiasvanderwerff: Correction 2: I tested on SM75 and the macro _does_ need to be guarded for SM75, otherwise it tries to compile device code with `nv_bfloat16` template type and error out since this is only available for SM>=80.

torchao/csrc/cuda/fp6_llm/utils_parallel_dequant.cuh
gau-nernst: Wow, that's an interesting and neat way to force constexpr 😄

torchao/csrc/cuda/fp6_llm/fp6_linear.cu
gau-nernst: That's.. interesting. I rmb we had an error before when trying to compile on old hardware #632, hence I added the CUDA_ARCH guard, and it compiles. Now thinking about it, I didn't test what would happen when I call the function on said old hardware, which would output garbage results, but no error, like you discovered.

I was following this https://github.com/pytorch/pytorch/blob/e98eac76b358fb4639b9e9ce6894014354d7b073/aten/src/ATen/native/cuda/int4mm.cu#L1 for the compile guard. And running tinygemm (this int4 mm kernel) on unsupported hardware will result in undefined symbol error at runtime. You can run this to reproduce on google colab https://github.com/pytorch/ao/issues/1110#issuecomment-2428090812.

I don't exactly like sm>=75 check at runtime (well, we already do that for sm75 launch params). I will try to play around with these compile guards to see their behavior. Now it seems like trying to support sm75 is getting very messy 🤣.

gau-nernst: @tobiasvanderwerff I created a thread over at GPU-MODE discord for easier discussion (writing it here in case you don't frequent discord much)

torchao/csrc/cuda/fp6_llm/fp6_linear.cu
```diff
@@ -141,17 +149,31 @@ cudaError_t fpx_linear_kernel(cudaStream_t    stream,
         // Reduction for SplitK
         dim3 GridDim((M_Global * N_Global) / REDUCTION_ELEMENT_PER_THREADBLOCK, 1, 1);
         dim3 BlockDim(WARP_SIZE, 1, 1);
-        SplitK_Reduction<<<GridDim, BlockDim, 0, stream>>>(C, Reduction_Workspace, M_Global, N_Global, Split_K);
+        SplitK_Reduction<InputDataType><<<GridDim, BlockDim, 0, stream>>>(C, Reduction_Workspace, M_Global, N_Global, Split_K);
     }
 
     return cudaGetLastError();
```
gau-nernst, line 155: Either remove this (so PyTorch will check CUDA error in subsequent kernel calls), or check this error (instead of just return). If we call the function but don't check the error, the error will be reset (subsequent `cudaGetLastError()` will return success).

I think we can remove this line and make `fpx_linear_kernel()` returns `void` instead. And if you want to check the error, add `cudaGetLastError()` and necessary checks at the end of `fp_eXmY_linear_forward_cuda()`. Thoughts?

tobiasvanderwerff: Good catch! I missed that subsequent kernel call. I've added an error check right after the call to `SplitK_Reduction`.

The return value from `fpx_linear_kernel` wasn't used anyway, so I think it's fine to make it a `void` return function. Instead, I'm not using calls to `TORCH_CHECK` to handle failure cases inside `fpx_linear_kernel`.

gau-nernst: I think we still need to call `CHECK_CUDA_KERNEL_LAUNCH()` when splitK=1?

tobiasvanderwerff: What do you mean? The `SplitK_Reduction` kernel is only called when `splitK != 1`, so only then a call to `CHECK_CUDA_KERNEL_LAUNCH()` is required?

gau-nernst (approved):
The new changes look good! Figuring out how to handle CUDA_ARCH took more efforts than expected, but now we all understand it better 😄.
I think there are some problems with CUDA CI right now. Will need to wait for that to be fixed.

msaroufim (approved):

fxmarty-amd:
cool work!

PR #1150, opened 2024-10-23
[reland][ROCm] use dataclass for fnuz type setting

jeffdaily:
Reland of #1142.  Revert the revert #1148.  Then check for torch.cuda.is_available() before trying to call torch.cuda.get_device_properties(0).

PR #1154, opened 2024-10-23
DO NOT LAND: Test for Jeff

msaroufim:
- **Revert "Revert "[ROCm] use dataclass for fnuz type setting" (#1148)"**
- **avoid RuntimeError if HIP GPUs not available**


msaroufim:
If this is green https://github.com/pytorch/ao/actions/runs/11487300374 then that means https://github.com/pytorch/ao/pull/1150 is safe to land

msaroufim:
All green

Screenshot 2024-10-23 at 1 25 43 PM

Issue #1014, opened 2024-10-04
[easy] delete `torchao/float8/float8_aten_api.py` and move the functionality to `float8_ops.py`


vkuzo:
The `torchao/float8/float8_aten_api.py` file is a bit outdated as it contains the project structure from way back before we had any float8 support in PyTorch core.  Now that float8 is well supported, we can move the `mm_float8_emulated` function to `torchao/float8/float8_ops.py` and delete the `float8_aten_api.py` file to simplify the code.

1. code to be refactored away: https://github.com/pytorch/ao/blob/main/torchao/float8/float8_aten_api.py
2. take `mm_float8_emulated`, and move it to https://github.com/pytorch/ao/blob/main/torchao/float8/float8_ops.py, properly updating the `float8_mm` and `float8_addmm` callsites.  Delete `float8_aten_api.py`
3. ensure local tests pass and CI is green

vkuzo:
cc @drisspg

PR #1155, opened 2024-10-23
Move float8_aten_api to float8_ops

jainapurva:
Reformat float8 for simplification
- move mm_float8_emulated function to inline at callsites in torchao/float8/float8_ops.py 
- deleted the float8_aten_api.py file

Test Plan: All existing tests passed

Issue: https://github.com/pytorch/ao/issues/1014


torchao/float8/float8_ops.py
```diff
@@ -481,3 +483,36 @@ def copy_fp8(aten_op, args, kwargs=None):
...
+
+
+def mm_float8_emulated(
+    m1,  # input 1 data
+    s1,  # input 1 scale
+    m2,  # input 2 data
+    s2,  # input 2 scale
+    dtype3,  # output dtype
+):
+    # naive implementation: dq -> op -> q
```
vkuzo, line 495: can we just inline this to the callsite, and delete the boilerplate code?

torchao/float8/float8_ops.py
```diff
@@ -481,3 +483,36 @@ def copy_fp8(aten_op, args, kwargs=None):
...
+    m1_fp32 = m1.float() / s1
+    m2_fp32 = m2.float() / s2
+    m3_fp32 = torch.mm(m1_fp32, m2_fp32)
+
+    return m3_fp32.to(dtype3)
+
+
+# ATen op placeholders
```
vkuzo, line 503: can delete from here to the bottom of the file after the top function is inlined

test/float8/test_base.py
```diff
@@ -8,9 +8,9 @@
 import itertools
 import random
 import re
-from typing import List, Tuple
 import unittest
 import warnings
+from typing import List, Tuple
```
vkuzo, line 13: nit: it's ok for this PR, but in general it's good to do formatting changes in separate PRs, since this is not related to the change you're trying to accomplish.  If you do end up doing this, good practice to comment and just say something like "this is unrelated, just linting, etc".

vkuzo (approved):
thank you! looks great if CI is green

PR #1156, opened 2024-10-24
Leslie/torch2.5

leslie-fang-intel:
test UT


gau-nernst:
Since your PR with the necessary changes is here, I think we can move the discussion from #1100 to this PR instead.

PR #1157, opened 2024-10-24
Replace torch.empty with torch.zeros

helunwencser:
Summary:
It turns out that it is unsafe to use `torch.empty` in oss environment because `torch.empty` creates tensor with uninitialized data. That means the buffer could be initialized with random values depends on what is left on that piece of memory. This causes code to generate inconsistent behavior.

This PR replaces `torch.empty` with `torch.zeros` to make sure that they are properly initialized and avoid inconsistent behaviors.

Differential Revision: D64875312


msaroufim (approved):
Assuming CI is green this seems OK to merge

PR #1160, opened 2024-10-24
Add module-swap UX for int8 mixed-precision training

gau-nernst:


gau-nernst:
strange CI errors. seem like some race conditions

PR #1164, opened 2024-10-24
Revert "Add support for groupwise quantization for int8 weight only quantization"

jerryzh168:
Reverts pytorch/ao#1121

turns out this is not needed


andrewor14:
@jerryzh168 By the way we might still need this for int8 embedding. If that's not the case we can reopen this.

PR #1173, opened 2024-10-25
[experimental] Kleidi - add operator level tests

digantdesai:
Don't still wire it up for the ops.

### DIFF
--- a/torchao/experimental/CMakeLists.txt
+++ b/torchao/experimental/CMakeLists.txt
@@ -5,6 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 cmake_minimum_required(VERSION 3.19)
+include(CMakeDependentOption)
 
 project(torchao)
 
@@ -21,6 +22,12 @@ if(NOT TORCHAO_INCLUDE_DIRS)
   set(TORCHAO_INCLUDE_DIRS ${CMAKE_CURRENT_SOURCE_DIR}/../..)
 endif()
 
+cmake_dependent_option(TORCHAO_BUILD_KLEIDIAI "Download, build, and link against Arm KleidiAI library" ON
+    "CMAKE_SYSTEM_PROCESSOR STREQUAL arm64" OFF)
+if(TORCHAO_BUILD_KLEIDIAI)
+    message(STATUS "Building with Arm KleidiAI library")
+    add_compile_definitions(TORCHAO_ENABLE_KLEIDI=1)
+endif()
 include(CMakePrintHelpers)
 
 add_compile_options("-Wall" "-Werror" "-Wno-deprecated")
--- a/torchao/experimental/CMakeLists.txt
+++ b/torchao/experimental/CMakeLists.txt
@@ -22,12 +22,7 @@ if(NOT TORCHAO_INCLUDE_DIRS)
   set(TORCHAO_INCLUDE_DIRS ${CMAKE_CURRENT_SOURCE_DIR}/../..)
 endif()
 
-cmake_dependent_option(TORCHAO_BUILD_KLEIDIAI "Download, build, and link against Arm KleidiAI library" ON
-    "CMAKE_SYSTEM_PROCESSOR STREQUAL arm64" OFF)
-if(TORCHAO_BUILD_KLEIDIAI)
-    message(STATUS "Building with Arm KleidiAI library")
-    add_compile_definitions(TORCHAO_ENABLE_KLEIDI=1)
-endif()
+option(TORCHAO_BUILD_KLEIDIAI "Download, build, and link against Arm KleidiAI library (arm64 only)" OFF)
 include(CMakePrintHelpers)
 
 add_compile_options("-Wall" "-Werror" "-Wno-deprecated")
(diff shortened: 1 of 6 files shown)


torchao/experimental/CMakeLists.txt
```diff
@@ -21,6 +22,12 @@ if(NOT TORCHAO_INCLUDE_DIRS)
   set(TORCHAO_INCLUDE_DIRS ${CMAKE_CURRENT_SOURCE_DIR}/../..)
 endif()
 
+cmake_dependent_option(TORCHAO_BUILD_KLEIDIAI "Download, build, and link against Arm KleidiAI library" ON
```
metascroy, line 25: Maybe leave this as cmake arg that gets passed in?

torchao/experimental/kernels/cpu/aarch64/kleidi/kai_matmul_clamp_f32_qai8dxp1x8_qsi4c32p4x8_1x4x32_neon_dotprod.h
```diff
@@ -94,7 +94,6 @@ void kernel(
     const void* activation_data,
     float clamp_min,
     float clamp_max) {
-  assert(output_m_stride == n);
```
metascroy, line 97: why remove?

torchao/experimental/ops/tests/test_linear_8bit_act_xbit_weight.cpp
```diff
@@ -36,14 +41,40 @@ UKernelConfig get_ukernel_config() {
       &ukernel::prepare_weight_data<weight_nbit, has_weight_zeros, has_bias>;
   config.kernel_fn =
       &ukernel::kernel<weight_nbit, has_weight_zeros, has_bias, has_clamp>;
+  } else {
+#if defined(TORCHAO_ENABLE_KLEIDI)
+    assert (weight_nbit == 4);
```
metascroy, line 46: This is fine in test, but asserts get removed in release build, so use TORCHAO_CHECK

torchao/experimental/ops/tests/test_linear_8bit_act_xbit_weight.cpp
```diff
@@ -36,14 +41,40 @@ UKernelConfig get_ukernel_config() {
...
+#else
+    assert (false);
+#endif // TORCHAO_ENABLE_KLEIDI
+  }
 
   return config;
 }
 
-template <int weight_nbit, bool has_weight_zeros, bool has_bias, bool has_clamp>
+template <int weight_nbit, bool has_weight_zeros, bool has_bias, bool has_clamp, bool has_kleidi=false>
```
metascroy, line 74: What do you think of changing void test_linear_8bit_act_xbit_weight(int m, int n, int k, int group_size) to void test_linear_8bit_act_xbit_weight(int m, int n, int k, int group_size, UkernelConfig config).

And then create 

template <int weight_nbit, bool has_weight_zeros, bool has_bias, bool has_clamp>
UKernelConfig get_ukernel_config_universal()

and 

UKernelConfig get_ukernel_config_kleidi()

and use these in the tests.

torchao/experimental/ops/tests/test_linear_8bit_act_xbit_weight.cpp
```diff
@@ -12,14 +12,19 @@
 #include <torchao/experimental/ops/memory.h>
```
metascroy, line 12: This file runs in buck in internal CI.  Check that kleidi is covered there too (unless kleidi is not pulled in internally).

torchao/experimental/CMakeLists.txt
digantdesai: sure.

torchao/experimental/kernels/cpu/aarch64/kleidi/kai_matmul_clamp_f32_qai8dxp1x8_qsi4c32p4x8_1x4x32_neon_dotprod.h
digantdesai: this was incorrect because the meaning of the argument `n` is not same a `N`. Kleidi should do `output_m_stride == N (not n)` issert internally as well.

torchao/experimental/ops/tests/test_linear_8bit_act_xbit_weight.cpp
digantdesai: It should be guarded with `TORCHAO_ENABLE_KLEIDI` but after kleidi lands internally we can enable it for aarch64 CI.

torchao/experimental/ops/tests/test_linear_8bit_act_xbit_weight.cpp
digantdesai: we should run tests in debug mode, but for the proper integration (which is not done in this PR) we should check and allow taking kleidi path only when `if constexpr (weight_4bit == 4)`.

torchao/experimental/ops/tests/test_linear_8bit_act_xbit_weight.cpp
digantdesai: yeah I like it, esp when we increase number of uKernel configs. Ok if we do it in the next PR?

metascroy: At least create a task for it in the lowbit kernel GSD so we don't forget

metascroy (approved):

torchao/experimental/ops/tests/test_linear_8bit_act_xbit_weight.cpp
digantdesai: [T206115417](https://www.internalfb.com/intern/tasks/?t=206115417)

PR #1176, opened 2024-10-26
Fixing eval.py to use GPTQ_MT for gptq

HDCharles:
Summary: small fix so we can assess gptq accuracy

Test Plan: python eval.py --checkpoint_path $CHECKPOINT_PATH/$MODEL_REPO/model.pth --quantization int4wo-64-gptq --calibration_limit 1 --limit 1

Reviewers:

Subscribers:

Tasks:

Tags:


jcaip (approved):
lgtm

PR #1192, opened 2024-10-29
intx weight only linear quantizer for mps

manuelcandales:
Differential Revision: D65079774

### DIFF
--- a/torchao/experimental/ops/mps/test/test_quantizer.py
+++ b/torchao/experimental/ops/mps/test/test_quantizer.py
@@ -1,0 +1,170 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the license found in the
+# LICENSE file in the root directory of this source tree.
+
+from typing import Optional
+import copy
+import itertools
+import os
+import sys
+
+import torch
+import torchao_mps_ops
+import unittest
+
+from parameterized import parameterized
+from torchao.experimental.quant_api import UIntxWeightOnlyLinearQuantizer
+from torchao.experimental.quant_api import _quantize
+
+
+class TestUIntxWeightOnlyLinearQuantizer(unittest.TestCase):
+    BITWIDTHS = range(1, 8)
+    GROUPSIZES = [32, 64, 128, 256]
+
+    # Currently, the quantization code in quant_api.py only supports K values
+    # multiple of group_size.
+    # TODO(mcandales): Generalize the code in quant_api.py and add tests to
+    # cover values of K not multiple of group_size.
+    def _model_setup(self):
+        group_size = 32
+        k0 = 96
+        k1 = 224
+        k2 = 160
+        n = 47
+        layers = [
+            torch.nn.Linear(k0, k1, bias=False),
+            torch.nn.Linear(k1, k2, bias=False),
+            torch.nn.Linear(k2, n, bias=False),
+        ]
+        model = torch.nn.Sequential(*layers)
+        return model, group_size, k0, n
+
+    def _quantize_model(self, model, precision, nbit, group_size):
+        quantizer = UIntxWeightOnlyLinearQuantizer(
+            device="mps",
+            precision=precision,
+            bitwidth=nbit,
+            groupsize=group_size,
+        )
+        quantized_model = copy.deepcopy(model)
+        quantized_model = quantizer.quantize(quantized_model)
+        return quantized_model
+
+    @parameterized.expand(BITWIDTHS)
+    def test_export(self, nbit):
+        model, group_size, k0, n = self._model_setup()
+        m = 3
+        activations = torch.randn(m, k0, dtype=torch.float32, device="mps")
+
+        quantized_model = self._quantize_model(model, torch.float32, nbit, group_size)
+        exported = torch.export.export(quantized_model, (activations,))
+
+        for node in exported.graph.nodes:
+            if node.op == "call_function":
+                self.assertTrue(
+                    str(node.target)
+                    == f"torchao._linear_fp_act_{nbit}bit_weight.default"
+                )
+
+    @parameterized.expand(BITWIDTHS)
+    def test_2d_output_device_and_shape(self, nbit):
+        model, group_size, k0, n = self._model_setup()
+        m = 3
+        activations = torch.randn(m, k0, dtype=torch.float32, device="mps")
+
+        quantized_model = self._quantize_model(model, torch.float32, nbit, group_size)
+        result = quantized_model(activations)
+        self.assertTrue(result.is_mps)
+        self.assertTrue(result.shape == (m, n))
+
+    @parameterized.expand(BITWIDTHS)
+    def test_3d_output_device_and_shape(self, nbit):
+        model, group_size, k0, n = self._model_setup()
+        leading_shape = (3, 5)
+        activations = torch.randn(*leading_shape, k0, dtype=torch.float32, device="mps")
+
+        quantized_model = self._quantize_model(model, torch.float32, nbit, group_size)
+        result = quantized_model(activations)
+        self.assertTrue(result.is_mps)
+        self.assertTrue(result.shape == (*leading_shape, n))
+
+    @parameterized.expand(itertools.product(BITWIDTHS, GROUPSIZES))
+    def test_valid_groupsizes(self, nbit, group_size):
+        k0 = 3 * group_size
+        k1 = 7 * group_size
+        n = 47
+        layers = [
+            torch.nn.Linear(k0, k1, bias=False),
+            torch.nn.Linear(k1, n, bias=False),
+        ]
+        model = torch.nn.Sequential(*layers)
+        m = 5
+        activations = torch.randn(m, k0, dtype=torch.float32, device="mps")
+
+        quantized_model = self._quantize_model(model, torch.float32, nbit, group_size)
+        result = quantized_model(activations)
+        self.assertTrue(result.is_mps)
+        self.assertTrue(result.shape == (m, n))
+
+    @parameterized.expand(BITWIDTHS)
+    def test_invalid_groupsizes(self, nbit):
+        group_size = 16
+        k0 = 3 * group_size
+        k1 = 7 * group_size
+        n = 47
+        layers = [
+            torch.nn.Linear(k0, k1, bias=False),
+            torch.nn.Linear(k1, n, bias=False),
+        ]
+        model = torch.nn.Sequential(*layers)
+
+        with self.assertRaises(ValueError):
+            self._quantize_model(model, torch.float32, nbit, group_size)
+
+    # TODO(mcandales): Consolidate with the reference impl in test_lowbit.py
+    def _reference_linear_lowbit_quant_weights(self, A, W, group_size, S, Z):
+        N = W.shape[0]
+        K = W.shape[1]
+        W = W.to(torch.float32)
+        scales = S.t().unsqueeze(2).repeat(1, 1, group_size).view(N, -1)[:, :K]
+        zeros = Z.t().unsqueeze(2).repeat(1, 1, group_size).view(N, -1)[:, :K]
+        W = scales * W + zeros
+        return torch.mm(A, W.t())
+
+    @parameterized.expand(BITWIDTHS)
+    def test_accuracy(self, nbit):
+        group_size = 32
+        m = 3
+        n = 7
+        k = 64
+        with torch.no_grad():
+            activations = torch.rand(m, k, dtype=torch.float32, device="mps")
+            model = torch.nn.Sequential(*[torch.nn.Linear(k, n, bias=False)])
+            quantized_model = self._quantize_model(
+                model, torch.float32, nbit, group_size
+            )
+            result = quantized_model(activations)
+
+            # Compute expected result
+            weight_cpu = model[0].weight.data
+            weight_qvals_cpu, weight_scales_cpu, weight_zeros_cpu = _quantize(
+                weight_cpu, group_size, nbit, True, torch.uint8
+            )
+            weight_scales_cpu = weight_scales_cpu.t()
+            weight_zeros_cpu = -weight_zeros_cpu.t() * weight_scales_cpu
+            expected = self._reference_linear_lowbit_quant_weights(
+                activations.cpu(),
+                weight_qvals_cpu,
+                group_size,
+                weight_scales_cpu,
+                weight_zeros_cpu,
+            )
+
+            # Compare results
+            torch.testing.assert_close(result.cpu(), expected, rtol=0.001, atol=0.001)
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/torchao/experimental/quant_api.py
+++ b/torchao/experimental/quant_api.py
@@ -25,10 +25,14 @@ handler.setFormatter(formatter)
 logger.addHandler(handler)
 
 
-def _quantize(vals: torch.Tensor, group_size: int, nbit: int, has_weight_zeros: bool):
+def _quantize(vals: torch.Tensor, group_size: int, nbit: int, has_weight_zeros: bool, signed=True):
     assert nbit >= 1 and nbit <= 8
-    qmin = -(1 << (nbit - 1))
-    qmax = (1 << (nbit - 1)) - 1
+    if signed:
+        qmin = -(1 << (nbit - 1))
+        qmax = (1 << (nbit - 1)) - 1
+    else:
+        qmin = 0
+        qmax = (1 << nbit) - 1
 
     n, k = vals.shape
     vals = vals.reshape(-1, group_size)
@@ -51,7 +55,7 @@ def _quantize(vals: torch.Tensor, group_size: int, nbit: int, has_weight_zeros: 
         zero_points=group_zeros,
         quant_min=qmin,
         quant_max=qmax,
-        dtype=torch.int8,
+        dtype=torch.int8 if signed else torch.uint8,
         group_size=group_size,
     )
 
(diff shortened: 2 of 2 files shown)


torchao/experimental/ops/mps/register.mm
```diff
@@ -82,6 +94,31 @@ Tensor linear_mps_kernel(
...
+
+  check_linear_mps_args<nbit>(A, B, group_size, SZ);
+
+  auto M = A.size(0);
+  auto N = B.size(0);
+
+  return at::empty({M, N}, A.options()).to("meta");
+}
+
 // LowBit Packing on CPU Backend
```
metascroy, line 122: Should we add headers to the packed weights?  Do we want mps/metal in the name of the packing op?

torchao/experimental/ops/mps/register.mm
```diff
@@ -65,6 +57,26 @@ Tensor linear_mps_kernel(
...
       ", 2]");
+}
+
+template <int nbit>
+Tensor linear_mps_kernel(
+    const Tensor& A,
+    const Tensor& B,
+    int64_t group_size,
+    const Tensor& SZ) {
+  TORCH_CHECK(
```
metascroy, line 68: Should we use TORCHAO_CHECK?

torchao/experimental/quant_api.py
```diff
@@ -463,3 +463,106 @@ def quantize(self, model: nn.Module) -> nn.Module:
...
+        weight_qvals = (weight_qvals + (1 << (nbit - 1))).to(torch.uint8)
+
+        self.weight_scales_and_zeros = torch.stack(
+            (weight_scales.t(), weight_zeros.t()), dim=2
+        )
+
+        self.packed_weights = self._pack_weights_op(weight_qvals.cpu()).to(device="mps")
+
+    def forward(self, x):
+        assert x.dim() == 2
```
metascroy, line 494: Should we make it work for arbitrary shape to match behavior of torch.linear?  See cpu impls for reshaping.

torchao/experimental/quant_api.py
```diff
@@ -463,3 +463,106 @@ def quantize(self, model: nn.Module) -> nn.Module:
...
+                    torch.ops.torchao, f"_linear_fp_act_{nbit}bit_weight"
+                ),
+            )
+            setattr(module, name, qlinear)
+            getattr(module, name).quantize_and_pack_weights(
+                child.weight, nbit, group_size
+            )
+
+
+class IntxWeightOnlyLinearQuantizer:
```
jerryzh168, line 524: can you follow https://github.com/pytorch/ao/pull/995 instead

manuelcandales: We will converge towards adopting tensor subclass based API. In the meantime, we will use module swap to unblock us on e2e validation

torchao/experimental/ops/mps/register.mm
manuelcandales: Headers to the packed weights? What do you mean by that?

torchao/experimental/quant_api.py
metascroy: @manuelcandales let's file a task for this though

torchao/experimental/ops/mps/register.mm
```diff
@@ -17,22 +17,14 @@
...
   auto K = A.size(1);
 
-  TORCH_CHECK(
-      A.is_mps(), __func__, "A is on ", A.device(), " but expected on mps");
-  TORCH_CHECK(
-      B.is_mps(), __func__, "B is on ", B.device(), " but expected on mps");
-  TORCH_CHECK(
-      SZ.is_mps(), __func__, "SZ is on ", SZ.device(), " but expected on mps");
-
   TORCH_CHECK(
```
kimishpatel, line 28: In a separate diff move to TORCHAO CHECK as scott suggested

torchao/experimental/ops/mps/register.mm
metascroy: We have a header to add to packed weights that identifies the packing format: https://fburl.com/gmws8i73.  When the kernel reads the packed weights, it can check the header to make sure the packing format is as expected.

It is added to both the CPU linear op (https://fburl.com/90ok50q9) and the CPU embedding op (https://fburl.com/ydhdq5ye).

I'm wondering if the packed weights for MPS should also have the header.  I guess if we decide to do that, it can come in a follow-up PR because this PR doesn't pack the weights.  cc @kimishpatel for his thoughts.

torchao/experimental/ops/mps/register.mm
manuelcandales: PR updated

torchao/experimental/quant_api.py
manuelcandales: done!

torchao/experimental/ops/mps/register.mm
manuelcandales: done!

torchao/experimental/quant_api.py
manuelcandales: done!

torchao/experimental/ops/mps/test/test_quantizer.py
```diff
@@ -0,0 +1,104 @@
...
+import sys
+
+import torch
+import torchao_mps_ops
+import unittest
+
+torchao_root: Optional[str] = os.getenv("TORCHAO_ROOT")
+assert torchao_root is not None, "TORCHAO_ROOT is not set"
+
+sys.path.insert(0, torchao_root)
```
kimishpatel, line 19: Mark a TODO here since this should not really be needed but likely needed because we are not integrated in torchao's setup.py

torchao/experimental/ops/mps/test/test_quantizer.py
```diff
@@ -0,0 +1,104 @@
...
+from torchao.experimental.quant_api import IntxWeightOnlyLinearQuantizer
+
+
+def parameterized(test_cases):
```
kimishpatel, line 23: isnt there something equivalent within torchao?

torchao/experimental/ops/mps/test/test_quantizer.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+class TestIntxWeightOnlyLinearQuantizer(unittest.TestCase):
+    cases = [(nbit,) for nbit in range(1, 8)]
+
+    def _model_setup(self):
+        k0 = 512
+        k1 = 256
+        k2 = 128
+        k3 = 1024
```
kimishpatel, line 40: Check for non power of 2 and odd/prime number sizes. If they are not supposed to work than have expect failure tests.

torchao/experimental/ops/mps/test/test_quantizer.py
```diff
@@ -0,0 +1,104 @@
...
+
+    @parameterized(cases)
+    def test_2d_output_device_and_shape(self, nbit):
+        model = self._model_setup()
+        group_size = 32
+        m = 3
+        activations = torch.randn(m, 512, dtype=torch.float32, device="mps")
+
+        quantized_model = self._quantize_model(model, torch.float32, nbit, group_size)
+        result = quantized_model(activations)
```
kimishpatel, line 86: Is this running actual metal kernels? Can we put this in CI? @metascroy are cpu stack being tested in CI?

torchao/experimental/ops/mps/test/test_quantizer.py
```diff
@@ -0,0 +1,104 @@
...
+        self.assertTrue(result.is_mps)
+        self.assertTrue(result.shape == (m, 1024))
```
kimishpatel, line 88: Can you test the results as well? Unless operators (not kernels) are being tested separately, it would be good to have tests for these.

kimishpatel (changes requested):
Left a few comments

torchao/experimental/ops/mps/test/test_quantizer.py
metascroy: Nothing is tested in the torchao GH CI stack.

There are kernel/op tests in internal CI, and end-to-end tests in torchchat CI.  We should probably add some CI jobs in torchao OSS, though.  @jerryzh168 are we OK to do this?

torchao/experimental/quant_api.py
```diff
@@ -463,3 +463,106 @@ def quantize(self, model: nn.Module) -> nn.Module:
...
+    ):
+        super().__init__()
+        self._pack_weights_op = pack_weight_op
+        self._linear_op = linear_op
+
+    def quantize_and_pack_weights(self, weights, nbit, group_size):
+        self.nbit = nbit
+        self.group_size = group_size
+
+        weight_qvals, weight_scales, weight_zeros = _quantize(
```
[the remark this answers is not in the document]
metascroy: _quantize calls quantize_per_channel_group

torchao/experimental/ops/mps/test/test_quantizer.py
metascroy: The script should work without this if you run `pip install .` from ao directory.

@manuelcandales but if you do add something like this, you should be able to use the source file's location to find torchao_root

torchao/experimental/ops/mps/register.mm
metascroy: @kimishpatel I don't follow this:

> AFAIU, for metal kernels to have different packed weights

You can have a header even if you only have one current format, no?

torchao/experimental/ops/mps/register.mm
```diff
@@ -144,4 +181,14 @@ Tensor pack_weights_cpu_kernel(const Tensor& W) {
...
 }
 
+TORCH_LIBRARY_IMPL(torchao, Meta, m) {
+  m.impl("_linear_fp_act_1bit_weight", &linear_mps_kernel_meta<1>);
+  m.impl("_linear_fp_act_2bit_weight", &linear_mps_kernel_meta<2>);
+  m.impl("_linear_fp_act_3bit_weight", &linear_mps_kernel_meta<3>);
+  m.impl("_linear_fp_act_4bit_weight", &linear_mps_kernel_meta<4>);
+  m.impl("_linear_fp_act_5bit_weight", &linear_mps_kernel_meta<5>);
+  m.impl("_linear_fp_act_6bit_weight", &linear_mps_kernel_meta<6>);
+  m.impl("_linear_fp_act_7bit_weight", &linear_mps_kernel_meta<7>);
```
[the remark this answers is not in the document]
manuelcandales: yes

torchao/experimental/ops/mps/register.mm
manuelcandales: Probably in a later PR, when I register ExecuTorch ops

torchao/experimental/ops/mps/register.mm
```diff
@@ -17,22 +17,14 @@
 
 // LowBit Quantized Linear on MPS Backend
 template <int nbit>
-Tensor linear_mps_kernel(
+void check_linear_mps_args(
```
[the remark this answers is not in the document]
manuelcandales: It doesn't need to return bool. TORCH_CHECK will throw if condition isn't satisfied.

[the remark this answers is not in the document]
manuelcandales: By the way, I refactored, and this is now in a separate diff

torchao/experimental/quant_api.py
```diff
@@ -463,3 +463,106 @@ def quantize(self, model: nn.Module) -> nn.Module:
...
+            weights, self.group_size, self.nbit, True
+        )
+        weight_qvals = (weight_qvals + (1 << (nbit - 1))).to(torch.uint8)
```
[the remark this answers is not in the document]
manuelcandales: Done and done! Separate PR #1202

torchao/experimental/quant_api.py
```diff
@@ -463,3 +463,106 @@ def quantize(self, model: nn.Module) -> nn.Module:
...
+
+        self.weight_scales_and_zeros = torch.stack(
```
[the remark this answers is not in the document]
manuelcandales: Done! #1202

torchao/experimental/ops/mps/test/test_quantizer.py
manuelcandales: Operators are being tested separately in test_lowbit.py.
But I will add tests here as well, to make sure the quantization computations are correct.

torchao/experimental/ops/mps/test/test_quantizer.py
kimishpatel: Yeah. Lets not rely on some env variables to be set.

malfet, line 19: This look like a debug code and should be removed...

torchao/experimental/ops/mps/test/test_quantizer.py
manuelcandales: Done!

torchao/experimental/ops/mps/test/test_quantizer.py
manuelcandales: I updated the values, to cover non-power of 2 values. But right now, the quantization code inside quant_api only supports values of K that are a multiple of group_size. I added a TODO, to address this in a future diff. I don't want to make many changes to the quant_api code in this diff, since that is also used by the lowbit cpu kernels

torchao/experimental/ops/mps/test/test_quantizer.py
manuelcandales: done!

manuelcandales: done!

torchao/experimental/ops/mps/test/test_quantizer.py
```diff
@@ -0,0 +1,104 @@
...
+from typing import Optional
+import copy
+import os
+import sys
+
+import torch
+import torchao_mps_ops
+import unittest
+
+torchao_root: Optional[str] = os.getenv("TORCHAO_ROOT")
```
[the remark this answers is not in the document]
manuelcandales: removed!

torchao/experimental/quant_api.py
```diff
@@ -516,3 +516,114 @@ def apply(weight):
...
+            (weight_scales.t(), weight_zeros.t()), dim=2
+        )
+
+        self.packed_weights = self._pack_weights_op(weight_qvals.cpu()).to(device="mps")
+
+    def forward(self, x):
```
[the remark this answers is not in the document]
manuelcandales: the op itself asserts this

torchao/experimental/ops/mps/test/test_quantizer.py
manuelcandales: well, there is something in torch.testing._internal, but then I ran into dependency issues with 'expecttest'.

torchao/experimental/ops/mps/test/test_quantizer.py
jerryzh168: yeah I think we are fine with adding new CI jobs for metal kernels

torchao/experimental/ops/mps/test/test_quantizer.py
jerryzh168: we use this in other places: https://github.com/pytorch/ao/blob/c546c5c4ea5acf46067876470a5fb52f14f7e7fb/test/integration/test_integration.py#L647, not sure why there is a dependency issue

torchao/experimental/quant_api.py
```diff
@@ -463,3 +463,106 @@ def quantize(self, model: nn.Module) -> nn.Module:
...
+        assert x.dim() == 2
+        return self._linear_op(
+            x, self.packed_weights, self.group_size, self.weight_scales_and_zeros
+        )
+
+
+def _replace_linear_with_quantized_linear_mps(module: nn.Module, kwargs={}):
```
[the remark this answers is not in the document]
manuelcandales: I didn't want to touch code used by the lowbit cpu kernels in this PR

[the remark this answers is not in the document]
kimishpatel: Can you at least add todo to the effect

torchao/experimental/quant_api.py
```diff
@@ -516,3 +522,113 @@ def apply(weight):
...
+
+    assert not isinstance(module, nn.Linear)
+    assert nbit >= 1 and nbit <= 7
+
+    for name, child in module.named_children():
+        if not isinstance(child, nn.Linear):
+            _replace_linear_with_quantized_linear_mps(child, kwargs)
+        else:
+            assert child.bias is None
+            qlinear = UIntxWeightOnlyQuantizedLinear(
```
kimishpatel, line 578: Do you have groupsize restriction? If so where is that asserted? I would have expected that groupsize will be constructor arg so that constructor can check and throw if the quantized linear supports it or not. I dont exactly like exceptions in this scenario but maybe thats a better choice because you cannot create an instance of quantized linear for invaild group size

torchao/experimental/quant_api.py
```diff
@@ -516,3 +522,113 @@ def apply(weight):
...
+                pack_weight_op=getattr(torch.ops.torchao, f"_pack_weight_{nbit}bit"),
+                linear_op=getattr(
+                    torch.ops.torchao, f"_linear_fp_act_{nbit}bit_weight"
+                ),
+            )
+            setattr(module, name, qlinear)
+            getattr(module, name).quantize_and_pack_weights(
```
kimishpatel, line 585:
```suggestion
           qlinear.quantize_and_pack_weights(
```

torchao/experimental/ops/mps/test/test_quantizer.py
```diff
@@ -0,0 +1,146 @@
...
+            bitwidth=nbit,
+            groupsize=group_size,
+        )
+        quantized_model = copy.deepcopy(model)
+        quantized_model = quantizer.quantize(quantized_model)
+        return quantized_model
+
+    @parameterized(cases)
+    def test_export(self, nbit):
+        model, group_size, k0, n = self._model_setup()
```
kimishpatel, line 66: Are all tests doing only group size of 32? If so we should test other group sizes as well including those that should result in exception.

kimishpatel (changes requested):
Left some comments

torchao/experimental/quant_api.py
```diff
@@ -25,10 +25,16 @@
...
+def _quantize(vals: torch.Tensor, group_size: int, nbit: int, has_weight_zeros: bool, dtype=torch.int8):
     assert nbit >= 1 and nbit <= 8
-    qmin = -(1 << (nbit - 1))
-    qmax = (1 << (nbit - 1)) - 1
+    if dtype == torch.int8:
+        qmin = -(1 << (nbit - 1))
+        qmax = (1 << (nbit - 1)) - 1
+    elif dtype == torch.uint8:
+        qmin = 0
+        qmax = (1 << nbit) - 1
```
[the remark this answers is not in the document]
manuelcandales: well, `quantize_per_channel_group` in line 54, takes a `dtype` argument, not a `signed` argument.
If I do as you say, i would have to pass to `quantize_per_channel_group` the expression `signed ? dtype.int8 : dtype.uint8`
Why do that? Isn't it cleaner to just do what I did?

torchao/experimental/quant_api.py
manuelcandales: It is a constructor arg to `UIntxWeightOnlyLinearQuantizer`. I am adding the assert in there

torchao/experimental/ops/mps/test/test_quantizer.py
manuelcandales: done!

torchao/experimental/ops/mps/test/test_quantizer.py
manuelcandales: done!

torchao/experimental/quant_api.py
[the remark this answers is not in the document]
kimishpatel: I just meant that you used signed for your _quantize function and leave quantize_per_channel_group as is.

Why? Because right now all calls to `_quantize` now has to pass int8 or uint8 (depending on what the default value is), even when doing < 8 bit quant. So that feels a bit unintuitive. Instead if you just do `_quantize(vals, groups_size, nbit, has_weight_zero, signed=False)` then you know that you want to quantize using nbit and use unsigned type.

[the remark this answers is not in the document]
manuelcandales: ok, sounds good, coming right up. But i can't leave `quantize_per_channel_group` as is. I do need to pass `signed ? dtype.int8 : dtype.uint8` to it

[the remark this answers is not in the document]
manuelcandales: done!

kimishpatel (approved):

kimishpatel (approved):

PR #1201, opened 2024-10-30
ROCm Support : Tile_Layout kernel

petrex:
This pull request introduces support for ROCm (Radeon Open Compute) in addition to CUDA for GPU acceleration. The changes primarily focus on enabling the build and execution of ROCm-specific code paths alongside existing CUDA paths. 

In this PR, I use `tensor_core_tiled_layout` as proof of concept, but it generalizes to other kernels (for example, fp6_llm or sparse_marlin) with minimum effort. Feedback are welcome

co-author : @lcskrishna

## Features:

# ROCm Support Integration:

* [`setup.py`](diffhunk://#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R49-R53): Added detection for ROCm and adjusted the logic for compiling GPU extensions based on the availability of CUDA or ROCm.
# Conditional Compilation for ROCm:

* [`torchao/csrc/cuda/tensor_core_tiled_layout/tensor_core_tiled_layout.cu`](diffhunk://#diff-29bb1a2fd9317c74c807a7f558f5de755af0def91b9a49c81c409f8e76f736ddL1-R1): Introduced conditional compilation directives to include ROCm-specific headers and adjust constants and operations for ROCm. 

These changes ensure that the codebase can compile and run efficiently on both CUDA and ROCm platforms, leveraging the best available GPU acceleration technology.

## Usage
With ROCm pytorch nightly docker , simply run `PYTORCH_ROCM_ARCH=gfx942 python setup.py install `

## Next

- [ ] AMD specific unit tests (tensor_core_tiled_layout)
- [ ] workload and platform specific optimization (tensor_core_tiled_layout)

PR #1206, opened 2024-10-31
ROCm Sparse Marlin Kernels

petrex:
Built on top pf https://github.com/pytorch/ao/pull/1201. This pull request introduces support for ROCm (Radeon Open Compute) for sparse marling kernel in addition to CUDA, enabling the code to run on AMD GPUs. 

The main changes involve conditional compilation to handle differences between CUDA and ROCm, as well as adding ROCm-specific intrinsics for MI300x.

co-author : @lcskrishna

---------

Key changes include:

### ROCm Support in `setup.py`:
* hip kernels generation
### Conditional Compilation in CUDA Source Files:

* Added conditional compilation directives to exclude certain code for ROCm and include ROCm-specific implementations. 

### ROCm-specific Implementations:

* Implemented ROCm-specific versions of functions and macros that are different from their CUDA counterparts, ensuring compatibility and performance on AMD GPUs. 

Next:

- [ ] validation and benchmark across workloads on MIxxx GPUs


setup.py
```diff
@@ -46,9 +46,11 @@ def read_version(file_path="version.txt"):
     CUDAExtension,
```
msaroufim, line 46: I think you might enjoy stack based PR development https://github.com/modularml/stack-pr

torchao/csrc/cuda/sparse_marlin/mem.h
```diff
@@ -19,6 +19,28 @@
 #include "base.h"
 
 namespace torchao {
+
```
msaroufim, line 22: @xw285cornell looking for some quick advice, do you recommend we support AMD by adding conditional compilation flags to our existing cuda kernels or be OK with some more copy paste?

msaroufim: Chatted offline and indeed ifdefs are the way to go

msaroufim:
Do you have performance numbers by any chance relative to fp16? wanna make sure the performance improvements are competitive with CUDA

petrex:
> Do you have performance numbers by any chance relative to fp16? wanna make sure the performance improvements are competitive with CUDA

still WIP, but would you share the benchmark you guys are using? will tray that on mi300x when the PR is ready.

msaroufim:
Ok holler at me again whenever you need a review. Really excited to see this land

drisspg:
For benchmarking it is a little ad hoc the best place for this today would be to verify on: https://github.com/pytorch/ao/blob/main/torchao/_models/llama/generate.py

msaroufim (approved):
seems good to me, I'll lean on @atalman and @jcaip for the final merge since the error you're seeing in CI does seem like an underlying infra issue. It's not a flake though, I tried rerunning it and it still fails

torchao/csrc/cuda/sparse_marlin/mem.h
```diff
@@ -19,6 +19,28 @@
...
+#ifdef USE_ROCM
```
jcaip, line 23: Should we gate on a specific ROCm version like we do for CUDA?

torchao/csrc/cuda/tensor_core_tiled_layout/tensor_core_tiled_layout.cu
```diff
@@ -30,38 +41,71 @@ inline __device__ bf16x2x4 convert_i4x8_to_bf16x2x4(uint32_t source) {
...
         : "=r"(h[ii])
         : "r"(i4s), "n"(MASK), "n"(I4s_TO_BF16s_MAGIC_NUM), "n"(immLut));
+#endif
   }
 
   // This is the BF16 {-136, -136} represented as an integer.
+#if defined(USE_ROCM)
+#if ROCM_VERSION >= 60200
+  auto BF16_BIAS = __bfloat162bfloat162(__hip_bfloat16(__hip_bfloat16_raw{0xC308}));
+  auto BF16_ONE = __bfloat162bfloat162(__hip_bfloat16(__hip_bfloat16_raw{0x3F80}));
```
jcaip, line 87: what does B16_ONE refer to here?

jcaip (commented):
Did you get a chance to try this and get benchmarking numbers? Curious to see how it compares. We should probably update the testing framework too for AMD

torchao/csrc/cuda/sparse_marlin/mem.h
petrex: Good point! What we need is a GPU arch check instead of rocm version check. I have added a GPU architecture check in the setup.py file. As a result, the kernel will now only be built for the MI300X architecture.

torchao/csrc/cuda/tensor_core_tiled_layout/tensor_core_tiled_layout.cu
petrex: thanks let me clean up a little bit. I'd like this PR focus on sparse_marlin. tensor_core_tile_layout.cu should go to https://github.com/pytorch/ao/pull/1201 instead.

petrex:
> Looks good! Did you get a chance to try this and get benchmarking numbers? Curious to see how it compares. We should probably update the testing framework too for AMD

thanks It is planned. I will update the benchmark PR.

torchao/csrc/cuda/tensor_core_tiled_layout/tensor_core_tiled_layout.cu
petrex: 0x3F80 un BF16 : Sign bit (0) + Exponent (111111) + Mantissa (00000000) = 1.0
Just renamed it in #1201 to reflect this.

jcaip (approved):
LGTM, should be good to merge once we fix the setup.py conflicts.

petrex:
> LGTM, should be good to merge once we fix the setup.py conflicts.

done.

PR #1207, opened 2024-10-31
Call narrow only for TensorCoreTiledLayout

jerryzh168:
Summary:
att, previously in https://github.com/pytorch/ao/pull/914 we added narrow op for all layout, the introduced narrow op breaks the pattern for int8 dynamic activation int4 weight quant for executorch, this PR guarded narrow op for tensor core tiled layout only

If similar things coming up in the future we can factor this into a proper API for Layout or TensorImpl

Test Plan:
python test/test_integration.py -k test_export

Reviewers:

Subscribers:

Tasks:

Tags:


HDCharles (approved):
looks good

mcr229 (approved):

PR #1211, opened 2024-11-01
[subclasses] Use __slots__ for micro optim of flatten/unflatten

IvanKobzarev:
* __->__ #1211


torchao/float8/fsdp_utils.py
```diff
@@ -258,6 +261,16 @@ def fsdp_post_all_gather(
 
 
 class WeightWithDelayedFloat8CastTensor(torch.Tensor):
+
+    __slots__ = [
```
weifengpy, line 265: curious is `__slots__` derived from `__tensor_flatten__` ?

IvanKobzarev: Yes, here I just collected all attributes used in __tensor_flatten__, __tensor_unflatten__

PR #1214, opened 2024-11-01
Expose FakeQuantizeConfigs in QAT quantizers

andrewor14:
**Summary:** This commit exposes the activation and weight `FakeQuantizeConfig`s in the existing QAT quantizers. These are helpful for implementing advanced functionality based on the quantization schemes represented by these quantizers, such as composing QAT + LoRA.

**Test Plan:**

```
>>> from torchao.quantization.qat.linear import Int8DynActInt4WeightQATQuantizer
>>> q = Int8DynActInt4WeightQATQuantizer()
>>> q.get_activation_fake_quantize_config()
FakeQuantizeConfig(dtype=torch.int8, granularity=PerToken(), mapping_type=<MappingType.ASYMMETRIC: 3>, scale_precision=torch.float32, zero_point_precision=torch.float32, zero_point_domain=<ZeroPointDomain.INT: 1>, is_dynamic=True, range_learning=False)
>>> q.get_weight_fake_quantize_config()
FakeQuantizeConfig(dtype=<TorchAODType.INT4: 4>, granularity=PerGroup(group_size=256), mapping_type=<MappingType.SYMMETRIC: 1>, scale_precision=torch.float32, zero_point_precision=torch.float32, zero_point_domain=<ZeroPointDomain.INT: 1>, is_dynamic=True, range_learning=False)
```

```
>>> from torchao.quantization.qat.linear import Int4WeightOnlyQATQuantizer
>>> q = Int4WeightOnlyQATQuantizer()
>>> q.get_activation_fake_quantize_config()
>>> q.get_weight_fake_quantize_config()
FakeQuantizeConfig(dtype=torch.uint4, granularity=PerGroup(group_size=256), mapping_type=<MappingType.ASYMMETRIC: 3>, scale_precision=torch.bfloat16, zero_point_precision=torch.bfloat16, zero_point_domain=<ZeroPointDomain.FLOAT: 2>, is_dynamic=True, range_learning=False)
```

### DIFF
--- a/torchao/quantization/qat/linear.py
+++ b/torchao/quantization/qat/linear.py
@@ -107,12 +107,23 @@ class FakeQuantizedLinear(torch.nn.Linear):
         return F.linear(x, w)
 
 
+class _LegacyQATQuantizer(TwoStepQuantizer):
+    """
+    Base class for sharing common methods across legacy QAT quantizers.
+    """
+    def get_activation_fake_quantize_config(self) -> Optional[FakeQuantizeConfig]:
+        return None
+
+    def get_weight_fake_quantize_config(self) -> Optional[FakeQuantizeConfig]:
+        return None
+
+
 # =========================================================
 # |   Linear int8 dynamic activations + int4 weight QAT   |
 # =========================================================
 
 
-class Int8DynActInt4WeightQATQuantizer(TwoStepQuantizer):
+class Int8DynActInt4WeightQATQuantizer(_LegacyQATQuantizer):
     """
     Quantizer for performing QAT on a model, where linear layers have int8
     dynamic per token fake quantized activations and int4 fake quantized
@@ -189,6 +200,12 @@ class Int8DynActInt4WeightQATQuantizer(TwoStepQuantizer):
             else:
                 self._convert_qat_linear_8da4w(child)
 
+    def get_activation_fake_quantize_config(self) -> Optional[FakeQuantizeConfig]:
+        return _get_8da4w_activation_config(self.scales_precision)
+
+    def get_weight_fake_quantize_config(self) -> Optional[FakeQuantizeConfig]:
+        return _get_8da4w_weight_config(self.groupsize, self.scales_precision)
+
 
 class Int8DynActInt4WeightQATLinear(FakeQuantizedLinear):
     """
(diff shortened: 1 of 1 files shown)


torchao/quantization/qat/linear.py
```diff
@@ -200,6 +250,7 @@ class Int8DynActInt4WeightQATLinear(FakeQuantizedLinear):
         precision: precision of weights
         scales_precision: precision of per group scales and zero points
     """
+    _QUANTIZER_CLS = Int8DynActInt4WeightQATQuantizer
```
jerryzh168, line 253: this is a bit weird, can the `activation_config` and `weight_config` be passed from caller in `__init__`?

jerryzh168: and the args like `groupsize`, `precision`, `scales_precision` should be derived from the config I feel, otherwise  it will be error prone since there will be mismatches between `Int8DynActInt4WeightQATLinear` and `Int8DynActInt4WeightQATQuantizer`

torchao/quantization/qat/linear.py
```diff
@@ -107,12 +107,23 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
         return F.linear(x, w)
 
 
+class _LegacyQATQuantizer(TwoStepQuantizer):
```
jerryzh168, line 110: nit: why is this called legacy, I thought we want to use the module swap API in the long term, or are we going to change this API to function call be still use module swap underneath?

torchao/quantization/qat/linear.py
andrewor14: I think that's also weird, because `Int8DynActInt4WeightQATLinear` should always have the same fake quantize configs. If users pass in other random configs then this linear will start doing unexpected things

torchao/quantization/qat/linear.py
andrewor14: yeah, in the long term we plan to use `quantize_` for module swap. These quantizers are too tailored to particular quantization schemes

torchao/quantization/qat/linear.py
andrewor14: Also this is just for the legacy code path. Eventually we want users to use `FakeQuantizedLinear` directly for everything, but for now we need to continue to expose this class for BC (many users are using it today)

jerryzh168: OK in that case, should we hard code the configs in `Int8DynActInt4WeightQATLinear`

andrewor14: Ok, but we'll have to duplicate the configs in that case (once in quantizer and once in linear class). Is that OK?

jerryzh168: hmmm, if you want to avoid mismatch, maybe you could refactor `_get_activation_fake_quantize_config` and `_get_weight_fake_quantize_config` to standalone helper functions so you don't need to set `_QUANTIZER_CLS` here, or considering you'll be deprecating `Int8DynActInt4WeightQATQuantizer`, it might be fine to duplicate as well

jerryzh168 (approved):

torchao/quantization/qat/linear.py
andrewor14: Ok, separated into standalone helper functions

PR #1216, opened 2024-11-02
[NF4] Add `quantize_()` API support for NF4

gau-nernst:
## Background

When I was working on INT8 training stuff a while back, I realized there was no issue to make NF4 training work with `quantize_()` API (i.e. swap plain tensor with tensor subclass). This is because autograd+compile still works correctly for `__torch_function__` (might not be the case in the past when NF4 was implemented) -> eliminate the need for custom `nn.Module` for downstream users as well as need for explicitly call to `linear_nf4()`.

## Usage

```python
from torchao import quantize_
from torchao.dtypes.nf4 import nf4_weight_only

model = ...
quantize_(model, nf4_weight_only())  # works for both training and inference. NF4 weight is non-trainable
```

This would also compose nicely with LoRA, making QLoRA implementation more seamless. e.g.

```python
from torch import nn

class LoRALinear(nn.Module):
    ...
    def forward(self, x):
        out = F.linear(x, self.weight)  # self.weight can be NF4Tensor
        out = out + self.lora_b(self.lora_a(x)) * (self.alpha / self.rank)
        return out

lora_layer = LoRALinear(...)
quantize_(
    lora_layer,
    nf4_weight_only(),
    filter_fn=lambda module, name: isinstance(module, LoRALinear),
)  # now this becomes QLoRA
```

### DIFF
--- a/test/dtypes/test_nf4.py
+++ b/test/dtypes/test_nf4.py
@@ -30,6 +30,7 @@ from torchao.dtypes.nf4tensor import (
     _INNER_TENSOR_NAMES_FOR_SHARDING,
     NF4Tensor,
     linear_nf4,
+    nf4_weight_only,
     to_nf4,
 )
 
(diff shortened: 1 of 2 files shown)


suryanshgupta9933:
This would be really neat.

msaroufim (approved):

PR #1219, opened 2024-11-04
Add a developer guide for exporting to executorch

jerryzh168:
Summary:
att, the requirement for exporting a quantized model to executorch is mainly that we want to preserve soem high level ops so they can be lowered to executorch ops, examples of ops that are already preserved are quantize_affine/dequantize_affine/choose_qparams_affine which can be matched in executorch for pattern matching, this PR adds an example for how to define and preserve a quantized embedding_byte op, the main util function we use is `torchao.utils._register_custom_op`

The expected output is that after export we see some high level ops (such as quantized embedding op) being preserved:

```
GraphModule(
  (0): Module(
    (parametrizations): Module(
      (weight): Module()
    )
  )
)



def forward(self, input):
    input, = fx_pytree.tree_flatten_spec(([input], {}), self._in_spec)
    _0_parametrizations_weight_original0 = getattr(self, "0").parametrizations.weight.original0
    _0_parametrizations_weight_original1 = getattr(self, "0").parametrizations.weight.original1
    input_1 = input
    embedding_byte = torch.ops.quant.embedding_byte.default(_0_parametrizations_weight_original0, [1, 128], _0_parametrizations_weight_original1, input_1);  _0_parametrizations_weight_original0 = _0_parametrizations_weight_original1 = input_1 = None
    return pytree.tree_unflatten((embedding_byte,), self._out_spec)
```

Test Plan:
python tutorials/developer_api_guide/export_to_executorch.py

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/tutorials/developer_api_guide/export_to_executorch.py
+++ b/tutorials/developer_api_guide/export_to_executorch.py
@@ -3,6 +3,7 @@ This tutorial shows how to preserve higher level operators in the model in order
 
 Specifically we define and preserved `torch.ops.quant.embedding_byte` op that works with quantized weights
 through `torch.export.export`, we can follow Executorch tutorials: https://pytorch.org/executorch/stable/tutorials/export-to-executorch-tutorial.html#lowering-to-edge-dialect to lower the model to executorch
+or rely on https://github.com/pytorch/executorch/tree/main/examples/models/llama and https://github.com/pytorch/torchchat libraries to export to target device.
 
 This can also support exporting the model to other platforms like ONNX as well.
 """
--- a/tutorials/developer_api_guide/export_to_executorch.py
+++ b/tutorials/developer_api_guide/export_to_executorch.py
@@ -1,0 +1,78 @@
+"""
+This tutorial shows how to preserve higher level operators in the model in order to be used in executorch
+"""
+import torch
+import torchao
+from my_dtype_tensor_subclass import (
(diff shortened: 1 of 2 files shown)


tutorials/developer_api_guide/export_to_executorch.py
```diff
@@ -0,0 +1,78 @@
...
+    m[0].weight = torch.nn.Parameter(to_my_dtype_extended(m[0].weight), requires_grad=False)
+    y_ref = m[0].weight.dequantize()[input]
+    y_q = m(input)
+    from torchao.quantization.utils import compute_error
+    sqnr = compute_error(y_ref, y_q)
+    assert sqnr > 45.0
+
+    # export
+    m_unwrapped = torchao.utils.unwrap_tensor_subclass(m)
+    m_exported = torch.export.export(m_unwrapped, (input,)).module()
```
andrewor14, line 70: are there steps after this that we should add or link to? e.g. `to_edge` or `to_backend`?

andrewor14 (approved):
Looks great! Can you also share the expected output of the tutorial in the PR description?

tutorials/developer_api_guide/export_to_executorch.py
jerryzh168: oh I can link a tutorial

PR #1225, opened 2024-11-05
support exlusion of params when using low bit optim

nighting0le01:
This PR allows, exclusion and inclusion of params layerwise when using low bit optimizers. this will allow for improving stability by running certain layers with 32 bit adam. https://huggingface.co/docs/bitsandbytes/main/en/optimizers

### DIFF
--- a/test/prototype/test_low_bit_optim.py
+++ b/test/prototype/test_low_bit_optim.py
@@ -19,6 +19,7 @@ from torchao.prototype.low_bit_optim.quant_utils import (
     quantize_4bit_with_qmap,
     _fp32_to_bf16_sr,
 )
+
 from torchao.utils import TORCH_VERSION_AT_LEAST_2_3, TORCH_VERSION_AT_LEAST_2_4, TORCH_VERSION_AT_LEAST_2_6
 
 try:
@@ -300,7 +301,72 @@ class TestOptim(TestCase):
             optim2.zero_grad()
 
             torch.testing.assert_close(loss1, loss2, msg=lambda msg: f"Iteration {idx}. {msg}")
+            
 
+    @pytest.mark.skipif(not TORCH_VERSION_AT_LEAST_2_3, reason="requires PyTorch >= 2.3")
+    @parametrize("optim_name", ["Adam8bit", "AdamW8bit", "Adam4bit", "AdamW4bit", "AdamFp8", "AdamWFp8"])
+    @parametrize("dtype", [torch.float32, torch.bfloat16])
+    @parametrize("device", _DEVICES)
+    def test_optim_exclude_low_bit_params(self, optim_name, dtype, device):
+        # Skip FP8 tests on devices that don't support it
+        if optim_name.endswith("Fp8") and device == "cuda":
+            if not TORCH_VERSION_AT_LEAST_2_4:
+                pytest.skip("FP8 CUDA requires PyTorch >= 2.4")
+            if torch.cuda.get_device_capability() < (8, 9):
+                pytest.skip("FP8 CUDA requires compute capability >= 8.9")
+                
+        model = nn.Sequential(
+            nn.Linear(32, 256),
+            nn.ReLU(),
+            nn.Linear(256, 32)
+        )
+        model.to(device=device, dtype=dtype)
+
+        params_to_exclude = [model[0].weight]
+
+        optim = getattr(low_bit_optim, optim_name)(
+            model.parameters(),
+            exclude_low_bit_optim_params=params_to_exclude
+        )
+
+        x = torch.randn(4, 32, device=device, dtype=dtype)
+        loss = model(x).sum()
+        loss.backward()
+        optim.step()
+        optim.zero_grad()
+        state = optim.state
+        excluded_param = params_to_exclude[0]
+        excluded_state = state[excluded_param]
+        exp_avg = excluded_state['exp_avg']
+        exp_avg_sq = excluded_state['exp_avg_sq']
+
+        if optim_name.endswith("8bit"):
+            quantized_state_types = (low_bit_optim.OptimState8bit,)
+        elif optim_name.endswith("4bit"):
+            quantized_state_types = (low_bit_optim.OptimState4bit,)
+        elif optim_name.endswith("Fp8"):
+            quantized_state_types = (low_bit_optim.OptimStateFp8,)
+        else:
+            quantized_state_types = ()
+
+        # Assert that the state tensors for the excluded parameter are not quantized
+        self.assertNotIsInstance(exp_avg, quantized_state_types)
+        self.assertNotIsInstance(exp_avg_sq, quantized_state_types)
+
+        for param in model.parameters():
+            if param is not excluded_param:
+                param_state = state[param]
+                exp_avg = param_state['exp_avg']
+                exp_avg_sq = param_state['exp_avg_sq']
+                self.assertIsInstance(exp_avg, quantized_state_types)
+                self.assertIsInstance(exp_avg_sq, quantized_state_types)
+
+        # Since the excluded parameter is not quantized, its data type should remain the same
+        self.assertEqual(excluded_param.dtype, dtype)
+
+        # Ensure that other parameters are still being updated correctly
+        for param in model.parameters():
+            self.assertIsNotNone(param.grad)
 
 class TestFSDP2(FSDPTest):
     @property
--- a/torchao/prototype/low_bit_optim/adam.py
+++ b/torchao/prototype/low_bit_optim/adam.py
@@ -13,7 +13,7 @@ from .quant_utils import _fp32_to_bf16_sr
 
 class _AdamBase(Optimizer):
     def __init__(
-        self, params, lr, betas, eps, weight_decay, amsgrad, *, block_size, bf16_stochastic_round, is_adamw
+        self, params, lr, betas, eps, weight_decay, amsgrad, *, block_size, bf16_stochastic_round, is_adamw,exclude_low_bit_optim_params=None
     ) -> None:
         if not 0.0 <= lr:
             raise ValueError("Invalid learning rate: {}".format(lr))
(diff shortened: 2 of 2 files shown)


test/prototype/test_low_bit_optim.py
```diff
@@ -300,7 +301,72 @@ def test_optim_bf16_stochastic_round_correctness(self):
...
+                exp_avg_sq = param_state['exp_avg_sq']
+                self.assertIsInstance(exp_avg, quantized_state_types)
+                self.assertIsInstance(exp_avg_sq, quantized_state_types)
+
+        # Since the excluded parameter is not quantized, its data type should remain the same
+        self.assertEqual(excluded_param.dtype, dtype)
+
+        # Ensure that other parameters are still being updated correctly
+        for param in model.parameters():
+            self.assertIsNotNone(param.grad)
```
gau-nernst, line 369: This is also redundant I think. The optimizer does not change `param.grad`. `param.grad` is computed only by the autograd engine (and I also think this will fail here, since you call `optim.zero_grad()` previously?)

torchao/prototype/low_bit_optim/adam.py
```diff
@@ -28,6 +28,10 @@ def __init__(
         self.block_size = block_size
         self.bf16_stochastic_round = bf16_stochastic_round
         self.is_adamw = is_adamw
+        # Store the IDs of parameters to exclude
+        self.exclude_low_bit_optim_params_ids = set(
+            id(p) for p in exclude_low_bit_optim_params
+        ) if exclude_low_bit_optim_params else set()
```
gau-nernst, line 34: I think you can hash tensor directly (it will use object id internally). PyTorch optimizer already hashes tensors when it uses params as keys in `self.state`.

torchao/prototype/low_bit_optim/adam.py
```diff
@@ -13,7 +13,7 @@
 
 class _AdamBase(Optimizer):
     def __init__(
-        self, params, lr, betas, eps, weight_decay, amsgrad, *, block_size, bf16_stochastic_round, is_adamw
+        self, params, lr, betas, eps, weight_decay, amsgrad, *, block_size, bf16_stochastic_round, is_adamw,exclude_low_bit_optim_params=None
```
gau-nernst, line 16: nit: formatting

gau-nernst (changes requested):
Thank you for the PR! Left some small comments.

torchao/prototype/low_bit_optim/adam.py
nighting0le01: thanks corrected

test/prototype/test_low_bit_optim.py
nighting0le01: thanks removed this

test/prototype/test_low_bit_optim.py
```diff
@@ -300,7 +301,72 @@ def test_optim_bf16_stochastic_round_correctness(self):
...
+        for param in model.parameters():
+            if param is not excluded_param:
+                param_state = state[param]
+                exp_avg = param_state['exp_avg']
+                exp_avg_sq = param_state['exp_avg_sq']
+                self.assertIsInstance(exp_avg, quantized_state_types)
+                self.assertIsInstance(exp_avg_sq, quantized_state_types)
+
+        # Since the excluded parameter is not quantized, its data type should remain the same
+        self.assertEqual(excluded_param.dtype, dtype)
```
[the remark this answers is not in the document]
nighting0le01: thanks removed this

test/prototype/test_low_bit_optim.py
```diff
@@ -300,7 +301,72 @@ def test_optim_bf16_stochastic_round_correctness(self):
...
+        self.assertNotIsInstance(exp_avg, quantized_state_types)
+        self.assertNotIsInstance(exp_avg_sq, quantized_state_types)
+
+        for param in model.parameters():
+            if param is not excluded_param:
+                param_state = state[param]
+                exp_avg = param_state['exp_avg']
+                exp_avg_sq = param_state['exp_avg_sq']
+                self.assertIsInstance(exp_avg, quantized_state_types)
+                self.assertIsInstance(exp_avg_sq, quantized_state_types)
```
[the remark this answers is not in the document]
nighting0le01: thanks incorporated this change

test/prototype/test_low_bit_optim.py
```diff
@@ -300,7 +301,72 @@ def test_optim_bf16_stochastic_round_correctness(self):
...
+        elif optim_name.endswith("4bit"):
+            quantized_state_types = (low_bit_optim.OptimState4bit,)
+        elif optim_name.endswith("Fp8"):
+            quantized_state_types = (low_bit_optim.OptimStateFp8,)
+        else:
+            quantized_state_types = ()
+
+        # Assert that the state tensors for the excluded parameter are not quantized
+        self.assertNotIsInstance(exp_avg, quantized_state_types)
+        self.assertNotIsInstance(exp_avg_sq, quantized_state_types)
```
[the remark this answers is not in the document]
nighting0le01: thanks incorporated this change

nighting0le01:
@gau-nernst  shall i also add configurable min_8bit_size like https://github.com/bitsandbytes-foundation/bitsandbytes/blob/9568735b21b9325e4789d6a5004517f2287f47c8/bitsandbytes/optim/optimizer.py#L603

over here :https://github.com/pytorch/ao/blob/71a442ae775e0ea5a541dcce637b128070d1243c/torchao/prototype/low_bit_optim/adam.py#L61

gau-nernst:
@nighting0le01 Adding something like `min_8bit_size` should be good. Though personally I don't know if having it is useful in any ways (does anyone use it / does adjust it help with stability?) If you still want to add it, maybe we can call it more generic, like `min_size_for_low_bit`, since we also have 4-bit and FP8.

Do you mind rebase/merge from main and make sure the tests pass?

nighting0le01:
hi @gau-nernst !

1. yes i have rebased and confirmed all test cases pass.
2. `min_size_for_low_bit`. why i propose this is to allow running gradient exploding or unstable layers in 32bit precision. similar motivation to https://huggingface.co/docs/bitsandbytes/main/en/optimizers#optimize-unstable-parameters
3. i can push it in another PR if you suggest

gau-nernst:
1. There are conflicts in your branch, hence I can't run the CI. Do you mind double-check? (from Github UI it shows `test_low_bit_optim.py` and `adam.py` have conflicts
2. From what I understand `min_size_for_low_bit` (or the original `min_8bit_size`) is to skip small params that don't contribute much memory savings if we use low-bit optim state for them (e.g. biases, norm params). How would it improve exploding gradients or instability? Usually instability appears in embedding layer or LM head I think (correct me if I'm wrong), which are large params but receive (somewhat) sparse gradients. In other words, how does increasing (or decreasing) the threshold help to improve stability? With this PR, the users can already select which specific params they want to keep optim state in original precision.

nighting0le01:
@gau-nernst  hi, sorry i was OOO for the last month, can you please run CI/CD now? verified test case passing locally

nighting0le01:
> @gau-nernst hi, sorry i was OOO for the last month, can you please run CI/CD now? verified test case passing locally

@gau-nernst  ran ruff check also now

test/prototype/test_low_bit_optim.py
```diff
@@ -32,6 +32,11 @@
     TORCH_VERSION_AT_LEAST_2_5,
     TORCH_VERSION_AT_LEAST_2_6,
 )
+from torchao.utils import (
+    TORCH_VERSION_AT_LEAST_2_3,
+    TORCH_VERSION_AT_LEAST_2_4,
+    TORCH_VERSION_AT_LEAST_2_6,
+)
```
gau-nernst, line 39: Don't reimport these

test/prototype/test_low_bit_optim.py
```diff
@@ -372,10 +377,56 @@ def test_optim_bf16_stochastic_round_correctness(self):
                 loss1, loss2, msg=lambda msg: f"Iteration {idx}. {msg}"
             )
 
+    @pytest.mark.skipif(not TORCH_VERSION_AT_LEAST_2_3, reason="requires PyTorch >= 2.3")
```
gau-nernst, line 380: In our CI, min PyTorch version is 2.3. We don't need to check >=2.3 anymore. You can remove this line

torchao/prototype/low_bit_optim/adam.py
gau-nernst: Can you try this? I don't think you need to use `id(p)` explicitly.

torchao/prototype/low_bit_optim/adam.py
```diff
@@ -59,7 +64,7 @@ def _new_buffer(self, p: Tensor, signed: bool):
         local_p = p.to_local() if isinstance(p, DTensor) else p
 
         # follow bitsandbytes, only quantize tensors >= 4096 values
-        if local_p.numel() >= 4096 and local_p.numel() % self.block_size == 0:
+        if p.numel() >= 4096 and p.numel() % self.block_size == 0 and id(p) not in self.exclude_low_bit_optim_params_ids:
```
gau-nernst, line 67: You should keep using `local_p` here for FSDP to work correctly (the check on divisibility should be done on local tensor, not the full tensor)

Furthermore, the check `id(p) not in self.exclude_low_bit_optim_params_ids` should be done before this. Just short-circuit it e.g. (`if p in self.exclude_low_bit_optim_params: return torch.zeros_like(p)`)

gau-nernst (changes requested):
Ruff lint is still failing. Can you double check?

Failing CUDA night seems to be unrelated.

Requested some changes because some of the code has been changed since you last opened this PR. Lmk if you have any questions.

jcaip:
cc @nighting0le01 are you still planning on working on this?

nighting0le01:
Sure let me land it this week

Issue #1190, opened 2024-10-29
[low-bit optim] Add COAT optimizer


gau-nernst:
Paper: https://arxiv.org/abs/2410.19313
Code: https://github.com/NVlabs/COAT (not available yet)

Seems like we already have most of the building blocks. The only new logic is "dynamic range expansion"

![](https://github.com/user-attachments/assets/d3c4da8d-82e7-486e-81d9-7b96fe842b66)

MirMustafaAli:
I would like to work on this @gau-nernst!!

gau-nernst:
@MirMustafaAli Go ahead and submit a PR 😄. Let me know if you face any problems.

MirMustafaAli:
@gau-nernst Which section should i look for to implement "dynamic range expansion"?. According to my understanding of repo it must be in float8 folder as it's aimed at hopper architecture utilizing type float8.  _Any pointers, PR's and reference methods which i can follow would be very much helpful for me_.

gau-nernst:
You can park it under `torchao/prototype/low_bit_optim`. The `float8/` folder is more for training stuff (fp8 matmul).

You can extend our current `OptimStateFp8`. See https://github.com/pytorch/ao/blob/000a49026459dd1dadf5ca34322d98e7b1680250/torchao/prototype/low_bit_optim/subclass_fp8.py. I think we only need to change the `quantize_fp8` function.

Another option is to create a separate optimizer. See https://github.com/pytorch/ao/blob/000a49026459dd1dadf5ca34322d98e7b1680250/torchao/prototype/low_bit_optim/adam.py. You can wrap all of the logic in a functional way (see `single_param_adam()` in the link above), and add boilerplate code for the optimizer (i.e. init optim states, call the functional optim step with torch.compile...)

MirMustafaAli:
Thanks!! Will work on your advice.

shreynath:
Hi! I'm Shreyaan, a Computer Engineering Student at Texas A&M getting started with open source contributions and I'd like to take a stab at this. From the thread it sounds like the cleanest path is extending OptimStateFp8 in torchao/prototype/low_bit_optim/subclass_fp8.py to implement the dynamic range expansion step from the COAT paper. Is this still available, and is the subclass_fp8.py extension approach still preferred over writing a standalone optimizer? The NVLABS code repo being unavailable means I'd be working from the paper directly — is that sufficient?

ATMmonitor667:
[...]

PR #1231, opened 2024-11-06
[low-bit optim] Add coat for float8 optimizer

MirMustafaAli:
This is a Work in Progress PR for #1190. 
 

As a draft PR, I have followed the first piece of advice by @gau-nernst of  _"extending OptimStateFp8"_.  Have **created a separate Dynamic Range Function** Instead of creating a different quantize_fp8 method  as it will be **applied before quantization** to **achieve larger representation range** of float8 datatypes and the **class will be storing value k**  to inverse the it after dequantization. 

**Requirements**:
TBA
**Additional Code/logic Added**:
TBA
**Logic/Code changes to existing codebase**:
TBA
**Outcome**:
TBA
**Scope of Usage**:
TBA
**Example**
TBA

**Changes**:
* _Dynamic Range Expansion Function_:  implementation of formula from the paper
* _Created OptimStateFp8WithDynamicRangeExpansion class by extending OptimStateFp8:_ by referencing the implementation of the OptimStatefp8. I have only overridden the dequantize method
* _Implemented aten.copy_.default and aten._to_copy.default for OptimStateFp8WithDynamicRangeExpansion_:

### Benchmarks

<b>Parameters</b>
| Parameter               | Value                                      |
|-------------------------|--------------------------------------------|
| Learning Rate (`lr`)    | 0.0001                                     |
| Automatic Mixed Precision (`amp`) | bf16                            |
| Seed                    | 42                                         |
| Model                   | timm/vit_base_patch16_224.augreg_in21k     |
| Optimizer (`optim`)     | AdamWFp8Ao_coat                            |
| Compile                 | False                                      |
| Profile                 | False                                      |
| Project                 | COAT-benchmarking                          |
| Number of Epochs        | 10                                         |
| Run Name                | AdamWFp8Ao_coat                            |
| Full BF16               | False                                      |
| Number of Workers       | 4                                          |
| Batch Size              | 1024                                       |
| Weight Decay            | 0                                          |
| Channels Last           | False                                      |
| Optimizer CPU Offload   | None                                       |
| Cosine LR Scheduler     | False                                      |
| Checkpoint Activations  | False                                      |

<b>Results</b>
<break>
W B Chart Nov 15 2024 (1)

W B Chart Nov 15 2024

W B Chart Nov 14 2024


gau-nernst:
I was thinking you can just add a flag to the current `OptimStateFp8`, something like `dynamic_range_expansion: bool`, instead of subclass-ing it.

MirMustafaAli:
> I was thinking you can just add a flag to the current `OptimStateFp8`, something like `dynamic_range_expansion: bool`, instead of subclass-ing it.

i have added the flag for optimstatefp8.  could you verify its right?

gau-nernst:
I think this requires a bit more work. You need to verify that you can create an optimizer with this (add test to https://github.com/pytorch/ao/blob/main/test/prototype/test_low_bit_optim.py) as well do some short training runs for sanity checks (using https://github.com/pytorch/ao/blob/main/benchmarks/benchmark_low_bit_adam.py).

I think for merging the PR, we should wait for the official code release to check numeric against them.

If you don't mind, we can discuss more details in GPU-MODE discord group https://discord.gg/gpumode. Just create a thread under torchao and tag me in (@gau.nernst)

MirMustafaAli:
sure!! i don't mind discussing in more details.

torchao/prototype/low_bit_optim/subclass_fp8.py
```diff
@@ -11,13 +11,23 @@
 DTYPE = torch.float8_e4m3fn
 
 
-def quantize_fp8(input: Tensor, block_size: int):
+def quantize_fp8(input: Tensor, block_size: int, apply_range_expansion: bool):
+
     shape = input.shape
     input = input.view(-1, block_size)
+    k =  aten.ones(input.shape[0], dtype=DTYPE, device=input.device)
```
gau-nernst, line 18: when `dynamic_range_expansion=False`, k should be None

torchao/prototype/low_bit_optim/subclass_fp8.py
```diff
@@ -11,13 +11,23 @@
...
+
+    if apply_range_expansion:
+        Rdtype = torch.finfo(DTYPE).max / torch.finfo(DTYPE).min # calculate the range of the dtype
+        Rx = input.abs().amax(-1).clip(1e-12) / input.abs().amin(-1).clip(1e-12) # range of input max and min
+        k = torch.log(Rdtype) / torch.log(Rx) # k calcuated
+        input = input.sign() * (input.abs() ** k.view(-1, 1))
```
gau-nernst, line 24: Add reference to the paper about this idea and calculation

torchao/prototype/low_bit_optim/subclass_fp8.py
```diff
@@ -46,6 +56,8 @@ def __init__(self, codes: Tensor, scale: Tensor):
         self.codes = codes
         self.scale = scale
         self.block_size = codes.numel() // scale.numel()
+        self.dynamic_range_expansion = dynamic_range_expansion
+        self.k = None
```
gau-nernst, line 60: `__init__()` for tensor subclass is supposed to only "wrap" the inner tensors. The arguments should be `codes`, `scale`, and `k`. Whether dynamic range expansion is used can be inferred from whether `k` is None

torchao/prototype/low_bit_optim/subclass_fp8.py
```diff
@@ -62,8 +74,13 @@ def dequantize(self, output_dtype=None):
         float_data = self.codes.float()
         float_data = float_data.view(-1, self.block_size) * self.scale.view(-1, 1)
 
+        if self.dynamic_range_expansion:
+            float_data = float_data.view(-1, self.block_size)
+            float_data =  float_data ** (1 / self.k.view(-1, 1))
```
gau-nernst, line 79: If we let `k=None` when `dynamic_range_expansion=False`, we can just check if `k is not None` here.

gau-nernst (changes requested):
Added some comments. You should add the `dynamic_range_expansion` flag to `.zeros()` method, because that is the entry point how `OptimStateFp8` is created (think factory function). `__init__()` method is only to wrap inner tensors, meant for internal usage. Users should not call `__init__()` directly.

torchao/prototype/low_bit_optim/subclass_fp8.py
```diff
@@ -11,13 +12,26 @@
 DTYPE = torch.float8_e4m3fn
 
 
-def quantize_fp8(input: Tensor, block_size: int):
+def quantize_fp8(input: Tensor, block_size: int, dynamic_range_expansion: bool):
+
     shape = input.shape
     input = input.view(-1, block_size)
+    k =  aten.ones(input.shape[0], dtype=DTYPE, device=input.device) if dynamic_range_expansion else None
```
gau-nernst, line 19: You can just set `k = None` here. It will be overwritten later if `dynamic_range_expansion=True`

torchao/prototype/low_bit_optim/subclass_fp8.py
```diff
@@ -62,15 +77,22 @@ def dequantize(self, output_dtype=None):
         float_data = self.codes.float()
         float_data = float_data.view(-1, self.block_size) * self.scale.view(-1, 1)
 
+        if self.k:
```
gau-nernst, line 80: I think this will error out when `k` is a tensor. `if self.k is not None` should be better.

torchao/prototype/low_bit_optim/subclass_fp8.py
```diff
@@ -88,12 +110,15 @@ def _(func, types, args, kwargs):
         assert dst.block_size == src.block_size
         dst.codes.copy_(src.codes)
         dst.scale.copy_(src.scale)
+        dst.k.copy_(src.k)
```
gau-nernst, line 113: Need to check if k is not None

gau-nernst (changes requested):

torchao/prototype/low_bit_optim/subclass_fp8.py
MirMustafaAli: understood

torchao/prototype/low_bit_optim/subclass_fp8.py
MirMustafaAli: okay will do it

torchao/prototype/low_bit_optim/subclass_fp8.py
MirMustafaAli: is it because this is ambiguous ?

torchao/prototype/low_bit_optim/subclass_fp8.py
MirMustafaAli: understood. since copy_ functionality if for tensor. it will lead to error if we implement it on. is my assumption correct? anything else i am missing?

torchao/prototype/low_bit_optim/adam.py
```diff
@@ -265,11 +266,28 @@ def __init__(
             bf16_stochastic_round=bf16_stochastic_round,
             is_adamw=False,
         )
+        self.dynamic_range_expansion = dynamic_range_expansion
 
     @staticmethod
-    def _subclass_zeros(p: Tensor, signed: bool, block_size: int):
-        return OptimStateFp8.zeros(p.shape, block_size, p.device)
+    def _subclass_zeros(p: Tensor, signed: bool, block_size: int, dynamic_range_expansion: bool):
+        return OptimStateFp8.zeros(p.shape, block_size, p.device, dynamic_range_expansion)
```
gau-nernst, line 273: You can make `_subclass_zeros()` a normal method instead of `staticmethod`, so that it can access `self.dynamic_range_expansion` directly -> don't need to pass in `dynamic_range_expansion` again. When I first wrote this, I didn't foresee such extension. If you want, you can remove `block_size` from all `_subclass_zeros()` signature too, since if we change it to a normal method, we can also access `self.block_size` directly.

test/prototype/test_low_bit_optim.py
```diff
@@ -150,6 +150,57 @@ def test_optim_smoke(self, optim_name, dtype, device):
 
         for p1, p2 in zip(model.parameters(), model2.parameters()):
             torch.testing.assert_close(p2, p1)
+    
+    @pytest.mark.skipif(
+        not TORCH_VERSION_AT_LEAST_2_4, reason="FP8 CUDA requires PyTorch >= 2.4"
+    )
+    @pytest.mark.skipif(
+        torch.cuda.get_device_capability() < (8, 9), reason="FP8 CUDA requires compute capability >= 8.9"
+    )
```
gau-nernst, line 159: Running this will fail on CPU CI

test/prototype/test_low_bit_optim.py
```diff
@@ -150,6 +150,57 @@ def test_optim_smoke(self, optim_name, dtype, device):
...
+    @parametrize(
+        "optim_name",
+        ["AdamFp8", "AdamWFp8"],
+    )
+    @parametrize("dtype", [torch.float32, torch.bfloat16])
+    @parametrize("device", _DEVICES)
+    @parametrize("optim_addon",["dynamic_range_expansion"])
+    def test_optim_addons(self, optim_name, dtype, device, optim_addon):
```
gau-nernst, line 167: You can make this `test_optim_fp8_coat_smoke()`. Don't need `optim_addon`, since when `dynamic_range_expansion=False`, we already tested it in the generic smoke test.

torchao/prototype/low_bit_optim/adam.py
```diff
@@ -51,8 +51,7 @@ def __setstate__(self, state):
             group.setdefault("amsgrad", False)
 
     # bring your own function to create zero-filled subclass
-    @staticmethod
-    def _subclass_zeros(p: Tensor, signed: bool, block_size: int):
+    def _subclass_zeros(self, p: Tensor, signed: bool):
```
gau-nernst, line 54: Nice work!

torchao/prototype/low_bit_optim/subclass_fp8.py
```diff
@@ -142,10 +172,11 @@ def _(func, types, args, kwargs):
...
         func(x.scale, *args[1:], **kwargs),
+        func(x.k, *args[1:], **kwargs) if x.k else None
     )
 
 
 if TORCH_VERSION_AT_LEAST_2_5:
     from torch.serialization import add_safe_globals
 
-    add_safe_globals([OptimStateFp8])
+    add_safe_globals([OptimStateFp8])
```
gau-nernst, line 182: I wonder why there is a diff here. Can you remove it? Perhaps missing new line?

gau-nernst (commented):
Thanks for the update. The PR is coming out nicely. There are some failing CI tests. Can you fix them, including the ruff linter?

Some extra items once that is finished:
- Update doc (link to the paper + usage)
- Run benchmark for sanity check https://github.com/pytorch/ao/blob/main/benchmarks/benchmark_low_bit_adam.py. I'm thinking comparing between BF16 baseline, FP8 optimizer, and FP8 COAT optimizer. Feel free to select a benchmark config suitable for you. And add the benchmark results in this PR description. Ideally, it should show that FP8 COAT is better than FP8 (though we might not observe it)

test/prototype/test_low_bit_optim.py
MirMustafaAli: Since float8 can only be run on hopper GPU's i added test. should i remove it?

torchao/prototype/low_bit_optim/subclass_fp8.py
MirMustafaAli: I have added and removed a newline, even tried with different spaces around it, but I’m still unsure why this is happening

MirMustafaAli: will try with spaces and see if i can remove it.

test/prototype/test_low_bit_optim.py
gau-nernst: The check should be inside the function, not as a decorator, like how it is done in `test_optim_smoke()`. Not exactly ideal, but we have to do it that way, since you can't call `torch.cuda.*` with CPU PyTorch.

torchao/prototype/low_bit_optim/subclass_fp8.py
gau-nernst: Do you by chance use a windows machine? maybe it's due to different newline character used by windows vs linux/macos. If that's the case, it's fine then.

torchao/prototype/low_bit_optim/adam.py
MirMustafaAli: Thank you

torchao/prototype/low_bit_optim/subclass_fp8.py
MirMustafaAli: i use macos

test/prototype/test_low_bit_optim.py
MirMustafaAli: understood i will work on it

torchao/prototype/low_bit_optim/subclass_fp8.py
```diff
@@ -47,7 +91,9 @@ def __init__(self, codes: Tensor, scale: Tensor):
         assert scale.ndim == 1
         self.codes = codes
         self.scale = scale
+        self.k = k
         self.block_size = codes.numel() // scale.numel()
+        self.sqrt_minmax_exp = sqrt_minmax_exp
 
     def __tensor_flatten__(self):
         return self.tensor_attrs, []
```
gau-nernst, line 99: When `k` and `sqrt_minmax_exp` is not None, you need to return them here also.

dev-requirements.txt
```diff
@@ -21,8 +21,7 @@ lm_eval
 diskcache
 pycocotools
 tqdm
-
-# Custom CUDA Extensions
+git+https://github.com/NVlabs/COAT.git#subdirectory=coat/optimizer/kernels # Custom CUDA Extensions
```
gau-nernst, line 24: Don't add this. CPU runner will fail to build CUDA extension. We will just test this locally.

torchao/prototype/low_bit_optim/subclass_fp8.py
MirMustafaAli: should i pass them instead of empty array?

gau-nernst: The first returned value (currently `self.tensor_attrs`) is a list of strings containing the names of tensor attributes. In this case, when there is no dynamic range extension, it's just `"codes", "scale"`. However, when there is dynamic range extension, you need to also add `"k", "sqrt_minmax_exp"`. IIRC, when they are None, you are not supposed to include them.

MirMustafaAli: Thank you for  explaining it. i have added them. looking forward to your approval

PR #1241, opened 2024-11-08
Fix Safe Load for NF4

drisspg:
Stacked PRs:
 * __->__#1241

--- --- ---

Fix Safe Load for NF4

### DIFF
--- a/torchao/utils.py
+++ b/torchao/utils.py
@@ -1,15 +1,14 @@
-import torch
-from typing import Tuple, Any, Callable
-from functools import reduce
 import functools
+import itertools
+import re
+import time
+from functools import reduce
 from importlib.metadata import version
 from math import gcd
-import torch.nn.utils.parametrize as parametrize
-import itertools
-import time
-import warnings
-import re
+from typing import Any, Callable, Tuple
 
+import torch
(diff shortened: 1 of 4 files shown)


torchao/utils.py
```diff
@@ -526,7 +585,24 @@ def fill_defaults(args, n, defaults_tail):
...
+    Example:
+    ```
+        class MyTensor(TorchAOBaseTensor):
+            pass
+
+        add_to_globals(MyTensor)
+    ```
+    """
+    if TORCH_VERSION_AT_LEAST_2_5:
+        torch.serialization.add_safe_globals([cls])
```
jerryzh168, line 608: or just inline?

drisspg: I think we are going to have to add this in many places I thought it would be nice to have a helper and point people to making new subclasses to this utility

jerryzh168: feels a bit overkill to save one line, but if you want to make a util it would be better to align with the signature of https://pytorch.org/docs/stable/notes/serialization.html#torch.serialization.add_safe_globals I think

jerryzh168 (approved):

PR #1255, opened 2024-11-08
Update torchao api reference and add contributor guide

jerryzh168:
Summary:
1. updated torchao api reference for quantization to include the APIs we want to expose, renamed torchao/quantization/linear_activation_weight_observer.py and removed the safe_int_mm and int_scaled_matmul from quant_primitives.py
2. added https://github.com/pytorch/ao/issues/391 to torchao docs

For torchao docs, I think we can include:
(1) API Reference (for modeling users)
(2) contributor guide (for researchers/developers)
(3) Tutorials (for modeling users)

for major feature announcement etc. users should refer to torchao repo README and release notes

Test Plan:
CI

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/docs/source/contributor_guide.rst
+++ b/docs/source/contributor_guide.rst
@@ -19,7 +19,7 @@ First we want to lay out the torchao stack::
 
   Quantization Algorithms/Flows: weight only/dynamic/static quantization, hqq, awq, gptq etc.
   ---------------------------------------------------------------------------------------------
-          Quantized Tensors (derived dtypes): AffineQuantizedTensor, CoodbookQuantizedTensor
+          Quantized Tensors (derived dtypes): AffineQuantizedTensor, CodebookQuantizedTensor
   ---------------------------------------------------------------------------------------------
     Quantization Primitive Ops/Efficient Kernels: matmul, quantize, dequantize
   ---------------------------------------------------------------------------------------------
@@ -208,6 +208,8 @@ Today we have some prototype low bit optimizers: `main/torchao/prototype/low_bit
 Quantized Training
 ******************
 Similar to low bit optimizers, we have quantized training prototype in `main/torchao/prototype/quantized_training <https://github.com/pytorch/ao/tree/main/torchao/prototype/quantized_training>`__, and we could extend AffineQuantizedTensor to support training as well, initial enablement is in progress, but there will be a lot of follow up work needed including making it work for different kernels etc.
+
+You can also checkout the tutorial for `Quantized Training <https://github.com/pytorch/ao/blob/main/tutorials/developer_api_guide/my_trainable_tensor_subclass.py>`__ that talks about how to make a dtype tensor subclass trainable.
 
 Case Study: How int4 weight only quantization works in torchao?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -600,75 +602,3 @@ Note: llama model (llama2/llama3) is our representative model for memory bound m
 Please checkout the ``--help`` option for each of the script to understand the supported options, e.g. you can use ``--profile=profile_path`` to get the chrome trace of the run to understand detailed `chrome trace <https://pytorch.org/tutorials/recipes/recipes/profiler_recipe.html#using-tracing-functionality>`__.
 
 Please let us know if there are any new important models that makes sense to be added to torchao model benchmark/eval folder.
-
-
-
-
-
-
-  
-  
-
-
-
-
-
-
-
-
-
-
-
-  
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-  
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-        
-
-
-
-
-
-
-
-
-
-
-
--- a/docs/source/contributor_guide.rst
+++ b/docs/source/contributor_guide.rst
@@ -1,0 +1,674 @@
+torchao Contributor Guide
+-------------------------
+
+.. toctree::
+ :maxdepth: 3
+
+Objective
+=========
+In this doc we’ll talk about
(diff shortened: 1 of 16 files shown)


docs/source/index.rst
```diff
@@ -1,9 +1,7 @@
 Welcome to the torchao Documentation
 =======================================
 
-**torchao** is an open-source library that provides the functionality
-to quantize and prune your models using native PyTorch. Our documentation is under development
-with more content coming soon.
+`**torchao** <https://github.com/pytorch/ao>`__ is a ibrary for custom data types & optimizations. Quantize and sparsify weights, gradients, optimizers & activations for inference and training using native PyTorch. Please checkout torchao `README <https://github.com/pytorch/ao#torchao-pytorch-architecture-optimization>`__ for an overall introduction to the library and recent highlight and updates. The documentation here will focus on 1. API Reference 2. Developer / Researcher Contribution Guide 3. Tutorials.
```
drisspg, line 4: nit, `library`

torchao/quantization/linear_activation_weight_observed_tensor.py
```diff
@@ -147,6 +148,8 @@ def _(func, types, args, kwargs):
         args[0].to(*args[1:], **kwargs)._apply_fn_to_data(torch.clone),
     )
 
+to_linear_activation_weight_observed = LinearActivationWeightObservedTensor.from_float
```
drisspg, line 151: one thing I dont like about this pattern is that it makes doc finding/IDE things worse

torchao/quantization/quant_api.py
```diff
@@ -363,7 +364,7 @@ def insert_observers_(
     def convert_to_linear_observer(linear_module: nn.Linear):
         # Wrap the weight with LinearActivationWeightObservedTensor and then with nn.Parameter
         linear_module.weight = nn.Parameter(
-            LinearActivationWeightObservedTensor.from_float(
```
drisspg, line 366: ditto hear, I dont know why this is worse than the other

torchao/quantization/linear_activation_weight_observed_tensor.py
jerryzh168: this is to be consistent with other things right now, i.e. we use factory functions to create tensors. for this tensor subclass specifically, it's more of a "util" tensor subclass, maybe we can have different rules, like for "dtype tensor subclass" we use `to_tensor_subclass` to create them, for "util" tensor subclasses we use `TensorSubclass.from_float` etc. but I'm not sure if it's worth it to distinguish between different types here

jerryzh168: I'm not very familiar with doc finding/IDE, any way to mitigate?

drisspg: From some googling I dont think there are many ways to mitigate. 
>this is to be consistent with other things right now, i.e. we use factory functions to create tensors. 
Like in my mind Object.method is the canonical way to define a factory function in Python. I never really understood why we did this function name aliasing? is it just to make it shorter that feels weird especially since its an internal dev facing api and not end model user facing. 

I dont need to die on this hill but if this is all for naming I am find with changing `from_float` to any other name you would prefer.

jerryzh168: it's for following things like https://pytorch.org/docs/stable/generated/torch.Tensor.to_sparse.html#torch.Tensor.to_sparse, https://pytorch.org/docs/stable/generated/torch.Tensor.to_sparse_bsc.html#torch.Tensor.to_sparse_bsc, https://pytorch.org/docs/stable/generated/torch.Tensor.to_sparse_coo.html#torch.Tensor.to_sparse_coo I think.

maybe this will help explain: https://docs.google.com/document/d/1vXVQEM8FKKF73vNUsSfG2QrkuD5z3Y3eYyj-qcrdJX4/edit?tab=t.0#bookmark=id.92qvtqebp9wz

I also don't care about this as much, I can change it back

drisspg (approved):

docs/source/contributor_guide.rst
```diff
@@ -0,0 +1,674 @@
...
+Other Quantization Flows
+########################
+
+For other quantization flow/algorithms that does not fit into any of the above, we also intend to provide examples for common patterns. For example, `GPTQ like quantization flow <https://github.com/pytorch/ao/blob/e283743b3cc4612bb641b88dca3670231724d396/tutorials/calibration_flow/gptq_like.py>`__ that is adopted by `Autoround <https://github.com/pytorch/ao/blob/e283743b3cc4612bb641b88dca3670231724d396/torchao/prototype/autoround/README.md>`__, it uses `MultiTensor <https://gist.github.com/HDCharles/a1b575bbf8875f994af8a01b225e1227>`__ and module hooks to optimize the module.
+
+If you are working on a new quantization algorithm/flow and not sure how to implement it in a PyTorch native way, please feel free to open an issue to describe how your algorithm works and we can help advise on the implementation details.
+
+Training
+########
+The above flow are mainly focused on inference, but low bit dtype Tensors can be used in training as well.
```
andrewor14, line 197: add a link to the subclass training tutorial? https://github.com/pytorch/ao/blob/main/tutorials/developer_api_guide/my_trainable_tensor_subclass.py

docs/source/contributor_guide.rst
```diff
@@ -0,0 +1,674 @@
...
+Note: the doc is heavily focused on inference right now, but we plan to expand to cover training techniques in the future as well.
+
+torchao Stack Overview
+======================
+
+First we want to lay out the torchao stack::
+
+  Quantization Algorithms/Flows: weight only/dynamic/static quantization, hqq, awq, gptq etc.
+  ---------------------------------------------------------------------------------------------
+          Quantized Tensors (derived dtypes): AffineQuantizedTensor, CoodbookQuantizedTensor
```
andrewor14, line 22: typo? CodebookQuantizedTensor

andrewor14 (approved):

PR #1269, opened 2024-11-12
[low-bit optim] Fix edge cases for FSDP2 integration

gau-nernst:
This PR fixes 2 issues that came up in torchtune when fine-tuning Llama3.2-vision 

**1. Sometimes there is a strange torch.compile() error with DTensor when there is `.grad` field.** https://github.com/pytorch/torchtune/issues/1978#issuecomment-2469538542

This is actually an old issue #652 (see https://github.com/pytorch/ao/issues/652#issuecomment-2285040894 for more details). However, it does not always happen: our CI test passes (for PyTorch 2.6, but PyTorch 2.5 has this issue), finetune Llama text (not multimodal) in torchtune has no issues, but finetuning Llama3.2-vision faces the error -> It's not clear why and how this happens. The error message seems to indicate that torch.compile() tries to do dynamic-shape, even though we are explicitly using `dynamic=False`

The solution is to call `.detach()` on param, which shares the same weight storage, but now it doesn't have `.grad` anymore. Thanks to this, low-bit optim + FSDP2 also work for PyTorch 2.5 CI now (previously it didn't).

I can't add a test for this, since I don't know how/when this happens.

**2. Wrong DTensor creation when there is uneven sharding** (i.e. 1st dim is not divisible by world size)

Usually we don't have uneven shards for LLMs, thus this error didn't surface. However, for ViT, it might be possible due to pos_embed: in some implementation pos_embed includes CLS token, hence the first dim is `num_visual_tokens + 1`.

The fix is simple: pass `shape` (and `stride`) to `DTensor.from_local()`. An appropriate test has also been added.


gau-nernst:
@vkuzo Do you have time to take a look at this PR? Thank you :pray:

msaroufim (approved):

vkuzo (approved):

PR #1275, opened 2024-11-12
[1/2] Wean off of PYBIND in favor of torch.ops.load_library

janeyx99:
Pave the path for python agnostic ao by removing depending on PYBIND (which is not python agnostic).

Concretely, what happened this PR?
- no more PYBIND, so no more init.cpp
- for all non-CUDA platforms, ao no longer has custom cpp extensions, and thus ao is a pure python lib
- so we skip auditwheel (which is used for when the platform is linux) for all non-cuda wheel builds. (This does also mean we now needlessly build the same python wheel for every other platform rocm, cpu, xpu...)
- no more PYBIND also means no more torchao._C, which we're replacing with a load_library call

This PR should have no failures. The next PR will be targeting the wheel process to only output one wheel for every python version.

* __->__ #1275


setup.py
```diff
@@ -110,6 +110,9 @@ def get_extensions():
     if use_cuda:
         sources += cuda_sources
 
+    if len(sources) == 0:
```
huydhn, line 113: `if not sources` sounds more pythonic?

janeyx99: ehhhh it looks less readable to me

janeyx99:
Closing in favor of #1276 as that one passes all CI. It looks like Build M1 Wheels does not work with ghstack

PR #1276, opened 2024-11-13
[1/2] Wean off of PYBIND in favor of torch.ops.load_library

janeyx99:
The no ghstack version of https://github.com/pytorch/ao/pull/1275 to see if the mac tests pass (they do).

Pave the path for python agnostic ao by removing depending on PYBIND (which is not python agnostic).

Concretely, what happened this PR?
- no more PYBIND, so no more init.cpp
- for all non-CUDA platforms, ao no longer has custom cpp extensions, and thus ao is a pure python lib
- so we skip auditwheel (which is used for when the platform is linux) for all non-cuda wheel builds. (This does also mean we now needlessly build the same python wheel for every other platform rocm, cpu, xpu...)
- no more PYBIND also means no more torchao._C, which we're replacing with a load_library call

This PR should have no failures. The next PR will be targeting the wheel process to only output one wheel for every python version. If you're curious, the next PR looks like https://github.com/pytorch/ao/pull/1277

### DIFF
--- a/packaging/post_build_script.sh
+++ b/packaging/post_build_script.sh
@@ -7,15 +7,14 @@
 
 set -eux
 
-WHEEL_NAME=$(ls dist/)
+# Prepare manywheel, only for CUDA.
+# The wheel is a pure python wheel for other platforms.
+if [[ "$CU_VERSION" == cu* ]]; then
+    WHEEL_NAME=$(ls dist/)
 
-pushd dist
-# Prepare manywheel
-manylinux_plat=manylinux2014_x86_64
-if [[ "$CU_VERSION" == "xpu" ]]; then
-    manylinux_plat=manylinux_2_28_x86_64
-fi
-auditwheel repair --plat "$manylinux_plat" -w . \
+    pushd dist
+    manylinux_plat=manylinux2014_x86_64
+    auditwheel repair --plat "$manylinux_plat" -w . \
     --exclude libtorch.so \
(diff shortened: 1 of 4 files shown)


msaroufim:
>  (This does also mean we now needlessly build the same python wheel for every other platform rocm, cpu, xpu...)

Ok guessing we'll delete those jobs in a future PR then?

packaging/post_build_script.sh
```diff
@@ -7,15 +7,14 @@
 
 set -eux
 
-WHEEL_NAME=$(ls dist/)
+# Prepare manywheel, only for CUDA.
+# The wheel is a pure python wheel for other platforms.
+if [[ "$CU_VERSION" == cu* ]]; then
```
msaroufim, line 12: is this also true for rocm?

msaroufim (approved):

packaging/post_build_script.sh
janeyx99: No, but it looks like the current .cu files are not built for rocm based on the `use_cuda` here https://github.com/pytorch/ao/blob/main/setup.py#L64

msaroufim: Ok! heads up to @petrex who has been working on adding custom hip kernel support

petrex: @janeyx99 This https://github.com/pytorch/ao/pull/1201 would build hip kernels.

PR #1283, opened 2024-11-14
Revert "pin nightlies to deal with std::badalloc"

jcaip:
Reverts pytorch/ao#1256


test/prototype/test_awq.py
```diff
@@ -70,7 +71,7 @@ def test_awq_loading(device, qdtype):
     
     model_save_path = "awq_model.pth"
     torch.save(m, model_save_path)
-    loaded_model = torch.load(model_save_path)
+    loaded_model = torch.load(model_save_path, assign=True)
```
jerryzh168, line 74: assign=True is an arg for `model.load_state_dict(...)` I think, see https://pytorch.org/ao/stable/serialization.html

malfet:
@jcaip do you know what caused it and what fixed it?

jcaip:
@malfet we switched over https://github.com/pytorch/ao/pull/1302 in this PR which fixed the jobs. Not sure about the root cause of the std::badalloc job but it was being triggerd on a jit test.

jerryzh168 (approved):

PR #1292, opened 2024-11-15
float8 delayed scaling: private API to fix user overriding buffers

vkuzo:
Summary:

Context: https://github.com/pytorch/torchtitan/issues/654

If the user has delayed scaling and FSDP float8 all-gather on, there is a subtle bug that can happen if the user calls
`model.to_empty(device="cuda")`:
1. to_empty recreates the buffers for tracking weight amax and scale
2. (1) leaves the buffers pointed to by Float8Linear.weight._amax_buffer, etc orphaned, because they don't participate in `to_empty`

I couldn't think of an easy and clean way to auto-fix this since we can't expect `torch.nn.Module` to know that our logic has multiple references to the same buffer, so exposing a private API for now until we can think of something better.

With the current fix, the user can then call
`_maybe_fixup_delayed_scaling_buffers` manually to relink the buffers to the correct new versions.

Test Plan: CI

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/float8/test_base.py
+++ b/test/float8/test_base.py
@@ -530,6 +530,29 @@ class TestFloat8Linear:
         with torch.inference_mode(mode=True):
             y = m(x)
 
+    @unittest.skipIf(not torch.cuda.is_available(), "CUDA not available")
+    def test_to_empty_delayed_scaling_with_float8_all_gather(self):
+        with torch.device("meta"):
+            m_ref = nn.Sequential(nn.Linear(32, 32))
+            config = Float8LinearConfig(
+                cast_config_input=CastConfig(scaling_type=ScalingType.DELAYED),
+                cast_config_weight=CastConfig(scaling_type=ScalingType.DELAYED),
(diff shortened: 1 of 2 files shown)


test/float8/test_base.py
```diff
@@ -530,6 +530,29 @@ def test_inference_mode(self):
...
+                enable_fsdp_float8_all_gather=True,
+            )
+            m_fp8 = convert_to_float8_training(m_ref, config=config)
+
+        assert m_fp8[0].fp8_amax_weight is m_fp8[0].weight._amax_buffer
+        assert m_fp8[0].fp8_amax_history_weight is m_fp8[0].weight._amax_history_buffer
+        assert m_fp8[0].fp8_scale_weight is m_fp8[0].weight._scale_buffer
+
+        m_fp8.to_empty(device="cuda")
+        m_fp8[0]._maybe_fixup_delayed_scaling_buffers()
```
weifengpy, line 550: we would need to call this inside torchtitan’s training loop?

vkuzo: yeah, which is definitely not ideal

Issue #1195, opened 2024-10-29
Add codebook (look up table based) quantization flow in torchao


jerryzh168:
Similar to affine quantization, we can implement codebook or look up table based quantization, which is another popular type of quantization, especially for lower bits like 4 bits or below (used in https://github.com/Vahe1994/AQLM, https://arxiv.org/abs/2402.04396 etc.). We can start with post training quantization and use k-means clustering to find the codebook / lookup table. You can check out https://github.com/pytorch/ao/issues/391 for the overall structure of torchao stack. Reference code for k-means can be found [here](https://github.com/apple/coremltools/blob/40f6705a6a4c5ef1a616f01b444f1d6e5a0c1f59/coremltools/optimize/torch/_utils/k_means.py#L1008).

After this we can also add more support for the advanced algorithms mentioned above.

API
```
quantize_(model, codebook_weight_only(dtype=torch.uint4))
```

Implementation details:
* [PR1] Ops
  * quantize_codebook(tensor, codebook)
  * dequantize_codebook(tensor, codebook)
* [PR2] Tensor Subclass
  * CodebookQuantizedTensor (similar to AffineQuantizedTensor)
    * clustering algorithm can be implemented in from_float function

Needs to flesh out the details of args etc. but can be done in the PR. I'd suggest to gradually add things and gather feedback.

Code Location: add a `codebook` folder under https://github.com/pytorch/ao/tree/main/torchao/prototype/quantization

malinjawi:
Hey @jerryzh168 I am new to torchao but this sounds like an issue I would want to investigate with my partner @harthi7. We will take a look and let you know how it goes. Cheers!

DerekLiu35:
Hi, I am also new to torchao and I would like to do this issue?

pawarmanasi07:
Hi! I'm interested in contributing to the implementation of the codebook quantization. Would it be helpful if I worked on [specific area - e.g., additional test cases, AQLM support, etc.]? Happy to coordinate with @DerekLiu35 to avoid duplicating effort.

jerryzh168:
I think the two immediate things are adding AQLM support and speedup. Adding AQLM in torchao will be a bit more convenient for users compared to using AQLM repo and then convert I think

pawarmanasi07:
Great! Let me know what I can start with.

DerekLiu35:
I'll focus on speeding up token generation, can coordinate more if @pawarmanasi07 also wants to work on that.

[...]

PR #1299, opened 2024-11-16
[WIP] Codebook quantization flow 

DerekLiu35:
This PR adds codebook quantization flow per https://github.com/pytorch/ao/issues/1195

#### **Usage**
```python
import torch
from torchao.prototype.quantization.codebook.codebook_quantized_tensor import CodebookQuantizedTensor

input_tensor = torch.randn(1024, 1024,  device='cuda')

block_size = (1, 1)
code_dtype = torch.uint4

quantized_tensor = CodebookQuantizedTensor.from_float(input_tensor, block_size, code_dtype)

dequantized_tensor = quantized_tensor.dequantize()
```

#### **ToDo**
- [ ] make `fit_kmeans` faster. Right now it takes >1 hour if you try to quantize a 1B model.


jerryzh168:
thanks for the contribution! yeah "> 1hour" seems a bit too slow, any ideas to speedup?

jerryzh168:
also after this is done, it would useful if you can add codebookquant to generate.py (https://github.com/pytorch/ao/blob/b714026478d87a98796498a9daf4edbb24845dca/torchao/_models/llama/generate.py#L209) and eval (https://github.com/pytorch/ao/blob/b714026478d87a98796498a9daf4edbb24845dca/torchao/_models/llama/eval.py#L71) to test the e2e model performance and accuracy

DerekLiu35:
> thanks for the contribution! yeah "> 1hour" seems a bit too slow, any ideas to speedup?

I think
   - For `block_size = (1, 1)`, It's similar to nf4tensor, so we can use absolute distance for scalars instead of euclidean distance
   - We could also decrease max_iter from 1000 to 200 for fit_kmeans but this would increase quantization error.

DerekLiu35:
changed max_iter from 1000 to 200 added codebook to eval and generate
I also added scales with groupsize 64 since when I ran eval codebook with torch.uint4 was getting high perplexity, the perplexity is still ~200 with scales

jerryzh168:
> changed max_iter from 1000 to 200 added codebook to eval and generate I also added scales with group size 64 since when I ran eval codebook with torch.uint4 it was getting high perplexity, the perplexity is still ~200 even with scales with group size 64

thanks, why the perplexity is so high? we get around 12/13 when using int4wo-64: https://github.com/pytorch/ao/tree/main/torchao/quantization#cuda-backend

DerekLiu35:
I found that it was because I was setting scales to the norm of each scale group instead of max of each scale group. If you set scales to be max of each scale group wikitext perplexity is ~11.6 for 
llama-3.2-3B. 
Though, AQLM initializes their scales as the norm of each scale group instead of max, but I guess they only do vector quantization and no scalar quantization.

jerryzh168:
Thanks if both performance and accuracy are reasonable, I think the main thing is to add a section for codebook quant: https://github.com/pytorch/ao/blob/main/torchao/quantization/README.md#uintx-quantization and show the perplexity and token/s result from eval.py and generate.py

torchao/prototype/quantization/codebook/codebook_ops.py
```diff
@@ -0,0 +1,317 @@
...
+
+    Returns:
+        torch.Tensor: The codebook tensor, shape (codebook_size, block_height, block_width).
+    """
+    if code_dtype == torch.int32:
+        codebook_size = 2**16
+    else:
+        codebook_size = _DTYPE_TO_QVALUE_BOUNDS[code_dtype][1] + 1
+
+    out_block_size, in_block_size = block_size
```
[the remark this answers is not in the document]
DerekLiu35: I made it 2d because that is how it was implemented in AQLM code. The code will be a little more complicated, but it should be possible to generalize to more than 2d.

DerekLiu35:
benchmarks were run on a single NVIDIA-A6000 GPU.

| Model       | Technique               | wikitext-perplexity | Tokens/Second | Memory Bandwidth (GB/s) | Peak Memory (GB) | Model Size (GB) |
| ----------- | ----------------------- | ------------------- | ------------- | ----------------------- | ---------------- | --------------- |
| Llama-3-8B  | Base (bfloat16)         |  7.590              |  32.36        |  485.71                 | 16.19            | 15.01           |
|             | codebook-4-64           |  9.533              |  1.73         |  8.62                   | 23.11            |  4.98           |
| Llama-3.1-8B| Base (bfloat16)         |  7.713              |  32.16        |  482.70                 | 16.35            | 15.01           |
|             | codebook-4-64           |  10.095             |  1.73         |  8.63                   | 23.11            |  4.98           |

Seems slow, might need custom kernels

torchao/prototype/quantization/codebook/codebook_quantized_tensor.py
```diff
@@ -0,0 +1,284 @@
...
+    def dequantize(self, output_dtype: Optional[torch.dtype] = None) -> torch.Tensor:
+        if output_dtype is None:
+            output_dtype = self.dtype
+
+        if isinstance(self.codes, UintxTensor):
+            codes = self.codes.get_plain()
+        else:
+            codes = self.codes
+        if codes.dtype == torch.uint8:
+            codes = codes.to(torch.int32)  # not sure how to index with uint8
```
jerryzh168, line 91: is this needed? this might be the reason why it's slow

DerekLiu35: I think so. I do indexing in `dequant = codebook[codes]` in `dequantize_codebook`. I got an error when I when I tried doing `codebook[codes]` when codes was uint8.

jerryzh168: what is the error? might be easy to support this in pytorch I feel

jerryzh168: only int and long are supported right now: https://github.com/pytorch/pytorch/blob/6e203ae6deaceb370e497bd50f2d02e894f5e9cc/aten/src/ATen/Dispatch.h#L801

DerekLiu35: UserWarning: indexing with dtype torch.uint8 is now deprecated, please use a dtype torch.bool instead.
IndexError: The shape of the mask [2] at index 0 does not match the shape of the indexed tensor [3, 3] at index 0.

jerryzh168: OK, let's add a TODO here to follow up and land this for now, is perplexity number expected here? it looks like it's slightly worse than int4wo: https://github.com/pytorch/ao/blob/main/torchao/quantization/README.md#cuda-backend

jerryzh168:
I think model level benchmark is fine for now, maybe add some unittest to test some basic functionality like getting codebook, quantize, dequantize before landing?

jerryzh168:
also for model level, is it possible to repro some accuracy result from AQLM: https://github.com/Vahe1994/AQLM/tree/main

torchao/prototype/quantization/codebook/codebook_quantized_tensor.py
DerekLiu35: I think perplexity is expected to be slightly worse than asymmetric int4wo (with scales and zero points) because I only add scales. But it should usually be better than symmetric int4wo (only scales, no zero points), though I initialize the centroids in kmeans randomly so it could be worse if centroids have bad initialization.

DerekLiu35:
I tested if I could convert an AQLM quantize model to my implementation as a sanity check and it seems like it works https://gist.github.com/DerekLiu35/c1cb9594c515e92c64762cbc8d087f7a.
Though, I get ~300 perplexity for llama-3-8b when I do similar setting as AQLM (block_size=(1, 8), code_dtype=torch.int32 (which makes codebook_size = 2**16), and scale_block_size=input.shape[1]). Maybe the extra tuning is necessary for lower perplexity?

jerryzh168:
Thanks! So the representation can be verified, but we are not sure how to repro the accuracy, and in original AQLM they'd need full model finetuning to get a good accuracy, seems like a good next step, if you are interested in improving this further

torchao/prototype/quantization/codebook/test_codebook_quant.py
```diff
@@ -0,0 +1,70 @@
...
+            self.input,
+            block_size=self.block_size,
+            code_dtype=self.code_dtype,
+            scale_block_size=self.scale_block_size,
+            chunk_size=self.chunk_size,
+        )
+
+        dequant = cqt.dequantize()
+
+        torch.testing.assert_close(dequant, self.input, atol=2, rtol=2)
```
jerryzh168, line 43: looks like a large drop, we could use a larger dtype for test I think, so we can get something closer

torchao/prototype/quantization/codebook/test_codebook_quant.py
```diff
@@ -0,0 +1,70 @@
...
+            scale_block_size=scale_block_size,
+            chunk_size=self.chunk_size,
+        )
+
+        dequant = cqt.dequantize()
+
+        torch.testing.assert_close(dequant, self.input, atol=0.1, rtol=0.1)
+
+        mse = torch.mean((dequant - self.input) ** 2).item()
+        self.assertLess(mse, 0.01)
```
jerryzh168, line 66: we typically use SQNR: https://github.com/pytorch/ao/blob/46b8796412eb350d1923091892850582d32737d0/torchao/quantization/utils.py#L50, numbers greater than 20 or 30 are reasonable

DerekLiu35:
> Thanks! So the representation can be verified, but we are not sure how to repro the accuracy, and in original AQLM they'd need full model finetuning to get a good accuracy, seems like a good next step, if you are interested in improving this further

Yeah, I'd definitely be interested in trying to implement AQLM to improve accuracy.

test/prototype/test_codebook_quant.py
```diff
@@ -0,0 +1,71 @@
...
+            self.input,
+            block_size=self.block_size,
+            code_dtype=self.code_dtype,
+            scale_block_size=self.scale_block_size,
+            chunk_size=self.chunk_size,
+        )
+
+        dequant = cqt.dequantize()
+
+        torch.testing.assert_close(dequant, self.input, atol=0.1, rtol=0.1)
```
jerryzh168, line 44: nit: it's fine to just rely on sqnr for error checking btw.

jerryzh168 (approved):
Thanks for the contribution and addressing all the comments!

PR #1302, opened 2024-11-18
migrate to linux_job_v2 and manylinux 2_28

HDCharles:
this is to handle the migration of dev-infra from manylinux_2_26 to 2_28: https://github.com/pytorch/pytorch/issues/123649

this required us update to linux_job_v2 for everything using nightly pytorch.
float8 tests: change to linux_job_v2 and remove binutils (they don't seem to be needed anymore)
nightly smoke tests: change to linux_job_v2
regression tests: only the nightlies need to use the new linux_job_v2, the rest needs to remain unchanged. We are currentlly pinning the pytorch version due to the bad_alloc tests but when that gets unpinned in https://github.com/pytorch/ao/pull/1283 the same changes for float8 test will need to be made to the nightly regression tests

these changes are working if CI is passing here

wheel builds: we update the manylinux version to 2_28, this was tested with https://github.com/pytorch/ao/actions/runs/11901019589 where all the cpu/cuda jobs pass, though we are still seeing failures with rocm jobs, that is a larger issue, unrelated to this PR, fixes for that are ongoing in https://github.com/pytorch/pytorch/issues/140631

### DIFF
--- a/.github/workflows/float8_test.yml
+++ b/.github/workflows/float8_test.yml
@@ -29,7 +29,7 @@ jobs:
             gpu-arch-type: "cuda"
             gpu-arch-version: "12.1"
 
-    uses: pytorch/test-infra/.github/workflows/linux_job.yml@main
+    uses: pytorch/test-infra/.github/workflows/linux_job_v2.yml@main
     with:
       timeout: 60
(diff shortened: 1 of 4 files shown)


atalman (approved):
lgtm

PR #1304, opened 2024-11-18
cmake torchao_ops_mps_linear_fp_act_xbit_weight

manuelcandales:
Summary: Move from setup.py to cmake for building custom torchao mps ops

Differential Revision: D66120124


torchao/experimental/ops/mps/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+
+include_directories(${TORCHAO_INCLUDE_DIRS})
+add_library(torchao_ops_mps_linear_fp_act_xbit_weight_aten SHARED aten/register.mm)
+
+target_include_directories(torchao_ops_mps_linear_fp_act_xbit_weight_aten PRIVATE "${TORCH_INCLUDE_DIRS}")
+target_link_libraries(torchao_ops_mps_linear_fp_act_xbit_weight_aten PRIVATE "${TORCH_LIBRARIES}")
+target_compile_definitions(torchao_ops_mps_linear_fp_act_xbit_weight_aten PRIVATE USE_ATEN=1)
+
+# Enable Metal support
+if (APPLE)
```
kimishpatel, line 33: Do we need to gate from specific version of OS? I think we should. both iOS and MacOS

torchao/experimental/ops/mps/build.sh
```diff
@@ -0,0 +1,20 @@
...
+echo "CMAKE_PREFIX_PATH: ${CMAKE_PREFIX_PATH}"
+export CMAKE_OUT=cmake-out
+
+cmake -DCMAKE_PREFIX_PATH=${CMAKE_PREFIX_PATH} \
+    -DCMAKE_INSTALL_PREFIX=${CMAKE_OUT} \
+    -S . \
+    -B ${CMAKE_OUT}
+cmake --build  ${CMAKE_OUT} -j 16 --target install --config Release
+
+rm ../../kernels/mps/src/metal_shader_lib.h
```
kimishpatel, line 20: wait what is this?

torchao/experimental/ops/mps/build.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash -eu
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the license found in the
+# LICENSE file in the root directory of this source tree.
+
+python ../../kernels/mps/codegen/gen_metal_shader_lib.py
```
kimishpatel, line 8: just add one line comment on what this does

kimishpatel: BTW in this script, why depend on yaml file for specification. Can you not just assume some pattern and include those metal files? Why add another step for someone to find their kernel needs extra step to be available to runtime

torchao/experimental/ops/mps/build.sh
```diff
@@ -0,0 +1,20 @@
...
+
+export CMAKE_PREFIX_PATH=$(python -c 'from distutils.sysconfig import get_python_lib; print(get_python_lib())')
```
kimishpatel, line 10: why cmake prefix path points to site-packages?

torchao/experimental/ops/mps/test/test_lowbit.py
```diff
@@ -4,10 +4,17 @@
...
 import torch
-import torchao_mps_ops
 import unittest
 
+path_libtorchao_ops_mps_aten = os.path.abspath(
+    os.path.join(
+        os.path.dirname(__file__), "../cmake-out/lib/libtorchao_ops_mps_linear_fp_act_xbit_weight_aten.dylib"
+    )
+)
+torch.ops.load_library(path_libtorchao_ops_mps_aten)
```
kimishpatel, line 16: - If and when we integrate experimental with torchao's pip installation, like pip install . or setup then at that type this lib should be installed in site-packages or some place like that instead of making it available via cmake-out in this way
- Put it under try catch and throw error message if not found

kimishpatel (changes requested):
mostly nits, but one question on file removal

kimishpatel (changes requested):
mostly nits, but one question on file removal

torchao/experimental/ops/mps/build.sh
manuelcandales: that's where the cmake stuff is. I am working on a conda environment.

torchao/experimental/ops/mps/build.sh
manuelcandales: after I am done building and installing, I don't need the generated metal_shader_lib.h anymore.

torchao/experimental/ops/mps/build.sh
manuelcandales: Because once I start optimizing the shaders, I will very quickly be in a situation where I will have multiple shaders per op, meaning, multiple versions. We want to benchmark them, and compare them. They should coexist. The yaml is the way to point at the best one, the "production one". We might end up in a place where one version is best for one platform/device, while another is best for another. If we get to that point, we can enrich the yaml to accommodate for those options. In the meantime, having this design allows me to very easily swap one version for another, and do comparisons.

torchao/experimental/ops/mps/build.sh
kimishpatel: ok then generated artifacts should be stored in build directory as temporary artifacts or somewhere else. Adding in the repo folder and removing it can leave the repo in dirty state if build fails. Also generally clean

torchao/experimental/ops/mps/build.sh
kimishpatel: ok sounds good

torchao/experimental/ops/mps/build.sh
kimishpatel: what do you mean thats where the cmake stuff is. prefix path is used for looking up packages https://cmake.org/cmake/help/latest/variable/CMAKE_PREFIX_PATH.html and other stuff

torchao/experimental/ops/mps/build.sh
```diff
@@ -0,0 +1,19 @@
...
+# All rights reserved.
+#
+# This source code is licensed under the license found in the
+# LICENSE file in the root directory of this source tree.
+
+cd "$(dirname "$BASH_SOURCE")"
+
+export CMAKE_PREFIX_PATH=$(python -c 'from distutils.sysconfig import get_python_lib; print(get_python_lib())')
+echo "CMAKE_PREFIX_PATH: ${CMAKE_PREFIX_PATH}"
+export CMAKE_OUT=$(python -c "import sys; print(sys.prefix)")/torchao_mps/cmake-out
```
kimishpatel, line 12: not-so-much-of-a-nit: why are you trying to install/cmake-out in python installation directory. Either pick the root directory of the repo or in the local directory and add cmake-out to .gitignore. https://fburl.com/p1n9rex7

torchao/experimental/ops/mps/test/test_lowbit.py
```diff
@@ -4,25 +4,31 @@
...
-                with self.subTest(case=case):
-                    func(self, *case)
+libname = "libtorchao_ops_mps_linear_fp_act_xbit_weight_aten.dylib"
+libpath = os.path.join(sys.prefix, "torchao_mps/cmake-out/lib/", libname)
 
-        return wrapper
+try:
+    torch.ops.load_library(libpath)
+except:
+    print(f"Failed to load library {libpath}")
```
kimishpatel, line 20:
```suggestion
    raise AttributeError(f"Failed to load library {libpath}")
```

kimishpatel: or ImportError

torchao/experimental/ops/mps/test/test_lowbit.py
```diff
@@ -4,25 +4,31 @@
...
+    raise
 
-    return decorator
+for nbit in range(1, 8):
+    op = getattr(torch.ops.torchao, f"_linear_fp_act_{nbit}bit_weight")
+    assert op is not None
```
kimishpatel, line 25: Please include appropriate error message as to which op was not found

torchao/experimental/ops/mps/test/test_quantizer.py
```diff
@@ -11,13 +11,27 @@
...
+
+try:
+    torch.ops.load_library(libpath)
+except:
+    print(f"Failed to load library {libpath}")
+    raise
+
+for nbit in range(1, 8):
+    op = getattr(torch.ops.torchao, f"_linear_fp_act_{nbit}bit_weight")
+    assert op is not None
```
kimishpatel, line 31: Same comment as before

torchao/experimental/ops/mps/test/test_quantizer.py
```diff
@@ -11,13 +11,27 @@
...
 from torchao.experimental.quant_api import UIntxWeightOnlyLinearQuantizer
 from torchao.experimental.quant_api import _quantize
 
+libname = "libtorchao_ops_mps_linear_fp_act_xbit_weight_aten.dylib"
+libpath = os.path.join(sys.prefix, "torchao_mps/cmake-out/lib/", libname)
+
+try:
+    torch.ops.load_library(libpath)
+except:
+    print(f"Failed to load library {libpath}")
```
kimishpatel, line 26: same as before

torchao/experimental/ops/mps/test/test_quantizer.py
```diff
@@ -11,13 +11,27 @@
...
+    raise
```
kimishpatel, line 27: This might need addressing when using this internally. I think the best way to do this is to do
```
try:
    for n bit.... loop to check for ops
except:
    try:
        load lib:
        for n bit...
    catch:
        throw error
```

kimishpatel (approved):
Stamping for now. Do consider addressing comments before landing

torchao/experimental/ops/mps/test/test_quantizer.py
manuelcandales: done!

torchao/experimental/ops/mps/build.sh
manuelcandales: done!

PR #1307, opened 2024-11-18
Added cpu support for llama generate.py/eval.py

jainapurva:
Updated the cuda dependencies for memory profiling, checked script on cpu.

Test plan: Ran script locally on Mac laptop


jerryzh168 (approved):
great, thanks! can you also paste result of testing on some quant options, like int8wo?

PR #1312, opened 2024-11-19
[NF4] `.to()` fixes

### LINKED ISSUE #1310: [NF4] Various bugs in how NF4 handles `.to()` to move to a different device
Reproduction

```python
import torch
from torch import nn
from torchao.dtypes.nf4tensor import to_nf4

x = torch.randn(1024, 1024)
x_nf4 = to_nf4(x)
print(x_nf4.cuda())  # this will dequantize NF4 -> unwanted
print(x_nf4.to(device="cuda"))  # this will raise error
print(x_nf4.to("cuda"))  # this will do the right thing

# .cpu() does not move .nf4 to CPU, because call_from_inner_tensors does not call the method on .nf4
x = torch.randn(1024, 1024).cuda()
x_nf4 = to_nf4(x).cpu()
print(x_nf4.quantized_data.device)  # cpu
print(x_nf4.nf4.device)  # cuda:0
print(x_nf4.to(torch.float32))  # error due to device mismatch

# not working with nn.Module
linear = nn.Linear(1024, 1024)
linear.weight = nn.Parameter(to_nf4(linear.weight.detach()), requires_grad=False)
linear.cuda()  # NF4 weight is not moved to CUDA
# linear.to("cuda")  # same problem

print(linear.weight.device)  # cuda:0
print(linear.weight.quantized_data.device)  # cpu
print(linear.weight.to(torch.float32).device)  # cpu
```

Summary:
1. `NF4Tensor.cuda()` will dequantize -> this is unwanted
2. `NF4Tensor.to(device="cuda")` will raise `IndexError`, since `args[1]` does not exist
3. `NF4Tensor.cpu()` does not move `.nf4` attribute -> cannot dequantize
4. Does not work with `nn.Module.to(device)`

- IMO, the semantics `NF4Tensor.to(torch.float32)` will dequantize is the culprit that causes these troubles + it is not consistent with AQT behavor. If `.to(dtype)` does not dequantize (only change appearance dtype), we only need to implement `aten._to_copy` instead of `Tensor.cpu`, `Tensor.to` and myriad of others. Though I understand this design is to make NF4 feels more like a true dtype.
- I think it makes more sense to designate `NF4Tensor.dequantize()` as the method to dequantize the tensor (also consistent with plain Tensor behavior, though plain `Tensor.dequantize()` will always return FP32), instead of the current situation (`NF4Tensor.dequantize()` is a static method for lookup table, while `NF4Tensor.get_original_weight()` does dequant)
- Changing this is BC, so we probably leave it as is.

gau-nernst:
Fixes #1310

### DIFF
--- a/torchao/dtypes/nf4tensor.py
+++ b/torchao/dtypes/nf4tensor.py
@@ -980,35 +980,44 @@ def implements_torch_function(torch_function):
 @implements_torch_function(torch.Tensor.to)
 def function_to_dtype(*args, **kwargs):
     tensor = args[0]
-    if isinstance(args[1], torch.dtype):
-        # Tensor.to(dtype, non_blocking, copy, memory_format)
-        return tensor.get_original_weight().to(*args[1:], **kwargs)
-    elif (
-        isinstance(args[1], torch.device)
-        or (
-            isinstance(args[1], str)
-            and (args[1] == "cpu" or args[1].startswith("cuda"))
+    device, dtype, non_blocking, convert_to_format = torch._C._nn._parse_to(
+        *args[1:], **kwargs
+    )
+
+    # dtype is specified -> dequantize
+    if dtype is not None:
+        return tensor.get_original_weight().to(
+            device, dtype, non_blocking, memory_format=convert_to_format
         )
-    ) and len(args) == 2:
-        # Tensor.to(device, non_blocking)
-        device = args[1]
-        updated_attrs = call_from_inner_tensors(tensor, "to", args[1:], kwargs)
-        updated_attrs["device"] = device
-        return NF4Tensor(*construct_nf4_args(tensor, updated_attrs))
-    else:
-        # Tensor.to(device, dtype, non_blocking, copy, memory_format)
-        # Tensor.to(other, non_blocking, copy)
-        raise NotImplementedError(
-            f"NF4Tensor.to({args[1:]}, {kwargs}) is not supported, passing to dispatch"
+
+    # dtype is not specified -> keep NF4
+    updated_attrs = dict(device=device)
+    tensor_attrs, _ = tensor.__tensor_flatten__()
+    for attr in tensor_attrs:
+        inner_tensor = getattr(tensor, attr)
+        updated_attrs[attr] = inner_tensor.to(
+            device, dtype, non_blocking, memory_format=convert_to_format
         )
+    return NF4Tensor(*construct_nf4_args(tensor, updated_attrs))
 
 
 @implements_torch_function(torch.Tensor.cpu)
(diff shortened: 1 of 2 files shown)


torchao/dtypes/nf4tensor.py
```diff
@@ -980,35 +980,44 @@ def decorator(func):
...
-    nf4tensor = args[0]
-    updated_attrs = call_from_inner_tensors(nf4tensor, "cpu", args[1:], kwargs)
-    updated_attrs["device"] = "cpu"
-    return NF4Tensor(*construct_nf4_args(nf4tensor, updated_attrs))
+    # Tensor.cpu(self, memory_format)
+    return args[0].to("cpu", *args[1:], **kwargs)
+
+
+@implements_torch_function(torch.Tensor.cuda)
+def function_cuda(*args, **kwargs):
```
drisspg, line 1011: does this func not work `call_from_inner_tensors`

drisspg (approved):

drisspg:
@ebsmothers Are there any ways to validate this against some tune flows to make sure it doesn't break anything?

torchao/dtypes/nf4tensor.py
gau-nernst: `call_from_inner_tensors()` does not call the method on `.scaler_mean` and `.nf4` attribute, hence I use `__tensor_flatten__` instead.

drisspg: I see. That's another anti-pattern since call_from_inner_tensors is typically applied for all the tensors and not just specific to sharding properties. But this makes sense. Maybe we should just like update it and have a flag that says ignore sharding or not.

drisspg: Otherwise, thank you. Yeah, I really didn't like the to calling dequant secretly, so I think this more aligns with what we've seen in the rest of the library.

gau-nernst: Yea I also penned some of my thoughts in #1310.

To clarify, this PR does not change the behavior "to calling dequant secretly (when dtype is specified)". Apart from fixes for `.scaler_mean` and `.nf4` attributes, this PR only changes `.cuda()` behavior to not dequantize (previously `.cuda()` will propagate to `aten._to_copy`, which dequantize), so it's more consistent with `.cpu()` as well as the general "not dequantize when dtype is not specified".

ebsmothers:
@drisspg sorry I missed your comment before so just looking now. Are there changes here that break BC? For what it's worth I patched the changes locally and ran `tune run --nnodes 1 --nproc_per_node 2 lora_finetune_distributed --config llama2/7B_qlora` and nothing broke.

Issue #1188, opened 2024-10-29
[FLOAT8] Add Hardware Compatibility Check for FP8 Quantization


drisspg:
#### Issue Summary
In our current implementation, we provide three APIs for model computation in FP8 format. However, for dynamic activation quant these FP8 computations are only supported on NVIDIA GPUs with SM89 and SM90 architectures. When models are quantized to FP8 on unsupported hardware, errors only occur during runtime, which can lead to confusion and wasted resources.

#### Proposed Solution
Check at the model quantization stage if the target hardware does not support FP8 computations and raise an error accordingly. This way, users are informed immediately if their hardware cannot handle FP8 quantization, rather than discovering it during runtime. Potentially point to weight-only quant which as more supported

Changes where to add errors:
```
    "float8_dynamic_activation_float8_weight",
    "float8_static_activation_float8_weight"
```

petrex:
good idea! I will add arch check for AMD GPUs as well.

PR #1314, opened 2024-11-19
Add hardware check to fp8 quant

jainapurva:
Add hardware check to ensure fp8 quantization only attempts runs on compatible hardware.

Test Plan: Ran float8_dynamic_quant on A100, MI300X, H100

Issue: https://github.com/pytorch/ao/issues/1188


torchao/quantization/quant_api.py
```diff
@@ -82,6 +82,7 @@
 from .utils import _get_per_token_block_size
 
 logger = logging.getLogger(__name__)
+is_cuda_8_9 = torch.cuda.is_available() and torch.cuda.get_device_capability() >= (8, 9)
```
jerryzh168, line 85: what about weight only quant, does it need CUDA 8.9 as well

jainapurva: For weight only, currently we just dequantize and perform the operation. The reason other techniques are cuda 8.9 dependent, is because scaled_mm is only supported on cuda >= 8.9

torchao/quantization/quant_api.py
```diff
@@ -939,6 +940,9 @@ def float8_dynamic_activation_float8_weight(
         mm_config (Float8MMConfig): Configuration for the matrix multiplication. Default uses fast accumulation.
 
     """
+    assert (
+        is_cuda_8_9
+    ), "Float8 dynamic activation quantization is only supported on CUDA 8.9 and above"
```
drisspg, line 945: This should also be supported on AMD. We should probably update this check.

cc @jeffdaily

drisspg (approved):

PR #584, opened 2024-08-01
Optimizer CPU offload for single GPU training

gau-nernst:
## Background

Currently there is no simple way to do optimizer CPU offload for single GPU training, although such feature exists for FSDP. DeepSpeed [ZeRO-Offload](https://www.deepspeed.ai/tutorials/zero-offload/) can work with single GPU, but it requires installing DeepSpeed, which can be complicated, and major changes to training loop (not convenient to switch between DeepSpeed and non-DeepSpeed).

Optimizer memory footprint is the largest in a training setup (2x model size for plain Adam), thus offloading optimizer to CPU would be greatly beneficial.

Below is a copy of optimizer CPU offload README

## Optimizer CPU Offload

This folder also implements optimizer CPU offload (i.e. ZeRO-Offload) for single GPU training. For multi-GPU training, you can use FSDP's built-in CPU offload.

```python
import torch
from torchao.prototype.low_bit_optim import CPUOffloadOptimizer

model = ...

# only offload optimizer state
optim = CPUOffloadOptimizer(model.parameters(), torch.optim.AdamW, fused=True)

# offload optimizer state AND gradients
optim = CPUOffloadOptimizer(model.parameters(), torch.optim.AdamW, offload_gradients=True, fused=True)
```

This will reduce GPU memory usage by optimizer state size, and additionally gradient size if `offload_gradients=True`. `CPUOffloadOptimizer` can wrap any base optimizer.

For saving and loading `CPUOffloadOptimizer`, it is important that you load model's weights BEFORE creating the optimizer, since we create a CPU copy of the parameters inside `CPUOffloadOptimizer.__init__()`. (TODO: we might want to have a method to synchronize CUDA and CPU params in either direction CPU->CUDA and CUDA->CPU, in case they are out of sync.)

```python
ckpt = torch.load("checkpoint.pth")

model = ...
model.load_state_dict(ckpt["model"])

optim = CPUOffloadOptimizer(model.parameters(), torch.optim.AdamW, fused=True)
optim.load_state_dict(ckpt["optim"])
```

NOTE:
- Since the optimizer step is done on CPU, it is highly recommended to use a fast CPU optimizer, such as `torch.optim.AdamW(fused=True)` (requires PyTorch 2.4). For other optimizers, you can try `torch.compile()` their optimizer step.
[...]

PR #1324, opened 2024-11-22
Enable CPU Offload for Intel GPU

dbyoung18:
# Background
Current CPU Offload in torchao only supports CUDA backend. We would like to add support for Intel GPU with the device option "xpu". 

# Details
- add "device" attribute to CPUOffloadOptimizer, default setting to "cuda"
- enhance and verify UT test_optim_cpu_offload_correctness & test_optim_cpu_offload_save_load pass on Intel GPU
- add "device" argument to benchmark_low_bit_adam.py. Users can use "--device xpu" to benchmark CPU Offload on Intel GPU. Currently it supports both full BF16 and BF16 AMP training w/ eager and compiled mode. Verified workloads on Intel GPU achieve memory saving and interleaving as expected as the description in reference PR:[ao#584](https://github.com/pytorch/ao/pull/584)

### DIFF
--- a/benchmarks/benchmark_low_bit_adam.py
+++ b/benchmarks/benchmark_low_bit_adam.py
@@ -4,7 +4,7 @@
 # - lpmm (4-bit optim): pip install yacs git+https://github.com/thu-ml/low-bit-optimizers.git
 # - DeepSpeed (ZeRO-Offload):
 #     sudo apt install libopenmpi-dev
-#     LD_LIBRARY_PATH=/usr/lib/x86_64-linux-gnu pip install mpi4p
+#     LD_LIBRARY_PATH=/usr/lib/x86_64-linux-gnu pip install mpi4py
 #     DS_BUILD_CPU_ADAM=1 pip install deepspeed --no-cache-dir
 #
 # To fine-tune a pre-trained ViT-Base on resisc45 dataset with BF16 AMP, using default AdamW optimizer from PyTorch core
@@ -98,6 +98,7 @@ def get_parser():
     parser.add_argument("--run_name", default="debug")
     parser.add_argument("--profile", action="store_true")
     parser.add_argument("--seed", type=int)
+    parser.add_argument("--device", type=str, choices=["cuda", "xpu"], default="cuda")
     return parser
 
 
@@ -128,9 +129,9 @@ def get_dloader(args, training: bool):
     )
 
 
-def get_amp_ctx(amp):
+def get_amp_ctx(amp, device):
     dtype = dict(bf16=torch.bfloat16, fp16=torch.float16, none=None)[amp]
-    return torch.autocast("cuda", dtype=dtype, enabled=amp != "none")
+    return torch.autocast(device, dtype=dtype, enabled=amp != "none")
 
 
 @torch.no_grad()
@@ -148,8 +149,8 @@ def evaluate_model(model, args):
         if args.channels_last:
             batch["image"] = batch["image"].to(memory_format=torch.channels_last)
 
-        with get_amp_ctx(args.amp):
-            all_preds.append(model(batch["image"].cuda()).argmax(1).cpu())
+        with get_amp_ctx(args.amp, args.device):
+            all_preds.append(model(batch["image"].to(args.device)).argmax(1).cpu())
 
     all_labels = torch.cat(all_labels, dim=0)
     all_preds = torch.cat(all_preds, dim=0)
@@ -192,7 +193,7 @@ if __name__ == "__main__":
         model.bfloat16()
     if args.channels_last:
         model.to(memory_format=torch.channels_last)
-    model.cuda()  # move model to CUDA after optionally convert it to BF16
+    model.to(args.device)  # move model to DEVICE after optionally convert it to BF16
     if args.compile:
         model.compile(fullgraph=True)
     print(f"Model parameters: {sum(p.numel() for p in model.parameters()):,}")
@@ -227,9 +228,9 @@ if __name__ == "__main__":
         optim_cls = OPTIM_MAP[args.optim]
 
         if args.optim_cpu_offload == "ao":
-            optim_cls = partial(low_bit_optim.CPUOffloadOptimizer, optimizer_class=optim_cls)
+            optim_cls = partial(low_bit_optim.CPUOffloadOptimizer, optimizer_class=optim_cls, device=args.device)
         elif args.optim_cpu_offload == "ao_offload_grads":
-            optim_cls = partial(low_bit_optim.CPUOffloadOptimizer, optimizer_class=optim_cls, offload_gradients=True)
+            optim_cls = partial(low_bit_optim.CPUOffloadOptimizer, optimizer_class=optim_cls, offload_gradients=True, device=args.device)
 
         optim = optim_cls(
             model.parameters(),
@@ -239,7 +240,7 @@ if __name__ == "__main__":
         )
 
     lr_schedule = CosineSchedule(args.lr, len(dloader) * args.n_epochs)
-    grad_scaler = torch.amp.GradScaler("cuda", enabled=args.amp == "fp16")
+    grad_scaler = torch.amp.GradScaler(args.device, enabled=args.amp == "fp16")
     log_interval = 10
     t0 = time.perf_counter()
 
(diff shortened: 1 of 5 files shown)


benchmarks/benchmark_low_bit_adam.py
```diff
@@ -4,7 +4,7 @@
 # - lpmm (4-bit optim): pip install yacs git+https://github.com/thu-ml/low-bit-optimizers.git
 # - DeepSpeed (ZeRO-Offload):
 #     sudo apt install libopenmpi-dev
-#     LD_LIBRARY_PATH=/usr/lib/x86_64-linux-gnu pip install mpi4p
+#     LD_LIBRARY_PATH=/usr/lib/x86_64-linux-gnu pip install mpi4py
```
gau-nernst, line 7: Thanks for fixing the typo!

benchmarks/benchmark_low_bit_adam.py
```diff
@@ -98,6 +98,7 @@ def get_parser():
     parser.add_argument("--run_name", default="debug")
     parser.add_argument("--profile", action="store_true")
     parser.add_argument("--seed", type=int)
+    parser.add_argument("--device", type=str, choices=["cuda", "xpu"], default="cuda")
```
gau-nernst, line 101: Can we auto-detect device instead? For a given PyTorch installation, only either one of cuda or xpu can be available  right? It's not possible to have both cuda and xpu PyTorch at the same time?

benchmarks/benchmark_low_bit_adam.py
```diff
@@ -248,15 +249,26 @@ def evaluate_model(model, args):
...
-        with torch.profiler.profile() if args.profile else nullcontext() as prof:
+        if args.profile:
+            activities = [torch.profiler.ProfilerActivity.CPU]
+            if args.device == "cuda":
+                activities.append(torch.profiler.ProfilerActivity.CUDA)
+            elif args.device == "xpu":
+                activities.append(torch.profiler.ProfilerActivity.XPU)
+
+            prof = torch.profiler.profile(activities=activities)
+        else:
```
gau-nernst, line 260: Can you check if on XPU PyTorch, PyTorch can auto-detect `ProfilerActivity.XPU`? https://github.com/pytorch/pytorch/blob/a8ab6b0938ed1788b0cad8982bcafaeaa5dea77d/torch/profiler/profiler.py#L61
Then we don't need to do that manually ourselves here.

torchao/prototype/low_bit_optim/cpu_offload.py
```diff
@@ -13,6 +13,7 @@ def __init__(
         optimizer_class: Type[Optimizer] = torch.optim.AdamW,
         *,
         offload_gradients: bool = False,
+        device: str = "cuda",
```
gau-nernst, line 16: Similarly to the above comment. Can we auto-detect device instead?

test/prototype/test_low_bit_optim.py
```diff
@@ -42,7 +42,12 @@
     lpmm = None
 
 
-_DEVICES = ["cpu"] + (["cuda"] if torch.cuda.is_available() else [])
+if torch.cuda.is_available():
+    _DEVICES = ["cpu", "cuda"]
+elif torch.xpu.is_available():
+    _DEVICES = ["cpu", "xpu"]
+else:
+    _DEVICES = ["cpu"]
```
gau-nernst, line 50: (just a note) We don't have XPU CI testing, so this doesn't do anything for now. But still good for local testing.

gau-nernst (approved):
Thanks for the feature addition! Hopefully once the device-agnostic API support arrives, we can eliminate the if-else checks 😆

torchao/prototype/low_bit_optim/cpu_offload.py
```diff
@@ -38,51 +40,57 @@ def __init__(
         if not isinstance(param_groups[0], dict):
             param_groups = [{"params": param_groups}]
 
-        self.param_cuda2cpu_map = dict()
+        self.param_d2h_map = dict()
         self.optim_dict = dict()
-        self.stream = torch.cuda.Stream()
+        self.device = device
+        self.stream = torch.Stream(device=self.device)
```
gau-nernst, line 46: 2.3 CI is failing because `torch.Stream` is only available on 2.5 I think. Perhaps we can do sth like `getattr(torch, device_type).Stream()`? (maybe do the same thing for other similar device API calls too?) Or just make this an if-else statement.

benchmarks/benchmark_low_bit_adam.py
dbyoung18: Checked it works w/ `torch.profiler.supported_activities()` to auto-detect `ProfilerActivity`. THX for ur review comment.

torchao/prototype/low_bit_optim/cpu_offload.py
dbyoung18: Add auto-detect available devices in torchao.utils. THX for ur review comment.

benchmarks/benchmark_low_bit_adam.py
gau-nernst: Since that works, I think we can just call `torch.profiler.profile()`. It will call `torch.profiler.supported_activities()` internally if `activities` is empty.

https://github.com/pytorch/pytorch/blob/a87925cc7e3a617542204e5f7f8e8ca766caaefe/torch/profiler/profiler.py#L669

torchao/prototype/low_bit_optim/cpu_offload.py
```diff
@@ -38,51 +38,53 @@ def __init__(
         if not isinstance(param_groups[0], dict):
             param_groups = [{"params": param_groups}]
 
-        self.param_cuda2cpu_map = dict()
+        self.param_d2h_map = dict()
         self.optim_dict = dict()
-        self.stream = torch.cuda.Stream()
+        self.device = get_available_devices()[-1]
+        assert ("cuda" in self.device or "xpu" in self.device), "CPU Offload currently only supports CUDA & XPU"
```
gau-nernst, line 44: Shouldn't this be `self.device in ("cuda", "xpu")` instead?

gau-nernst, line 44: Shouldn't this be `self.device in ("cuda", "xpu")` instead?

torchao/prototype/low_bit_optim/cpu_offload.py
dbyoung18: Replace related if-else statement w/ `getattr(torch, device_type).foo()`. And checked PyTorch v2.3 do have `torch.Stream` API and works well. Not sure the reason of 2.3 CI failure from the log. Since there are several updates about device setting, maybe we can trigger the CI again? THX for ur review comment.

torchao/prototype/low_bit_optim/cpu_offload.py
dbyoung18: Changed according to ur comment.

dbyoung18: Changed according to ur comment.

gau-nernst:
@dbyoung18 Can you run `ruff format` and push the formatted code? CUDA nightly is failing because of bitsandbytes calling `triton.ops` (I think later versions of triton doesn't have `triton.ops` anymore). It's not related but not sure if we can merge until that is fixed 😢. I think other PRs will be affected too.

Otherwise, everything else looks good already!

dbyoung18:
> @dbyoung18 Can you run `ruff format` and push the formatted code? CUDA nightly is failing because of bitsandbytes calling `triton.ops` (~I think later versions of triton doesn't have `triton.ops` anymore~ [bitsandbytes-foundation/bitsandbytes#1413](https://github.com/bitsandbytes-foundation/bitsandbytes/pull/1413)). It's not related but not sure if we can merge until that is fixed 😢. I think other PRs will be affected too.
> 
> Otherwise, everything else looks good already!

Done for ruff format. Hopes the bnb issue could be resolved soon. THX again for ur review and quick feedback:)

gau-nernst:
@dbyoung18 Can you merge from main? #1343 should fix the bnb issue.

Also, can you also update the doc here? https://github.com/pytorch/ao/tree/main/torchao/prototype/low_bit_optim#optimizer-cpu-offload

After that we are good to merge :smiley:

dbyoung18:
> @dbyoung18 Can you merge from main? #1343 should fix the bnb issue.
> 
> Also, can you also update the doc here? https://github.com/pytorch/ao/tree/main/torchao/prototype/low_bit_optim#optimizer-cpu-offload
> 
> After that we are good to merge 😃

Done for both. We have a plan to gradually support torch-ao & pytorch core on Intel GPU. For this PR it covers CPU Offload only and I will look into the remain part of low-bit optimizers for next step. Since meanwhile we are also on the way to upstream FlashAttention backend to pytorch core(target v2.6 or v2.7), would like to add benchmark data to the README when it's ready. So currently, I only modify the README to make the CPU-Offload part to cover XPU scope. THX for review and I am also looking forward to make further contributions soon.😃

gau-nernst:
Sounds good! The low-bit optimizers rely entirely on the tensor subclass + torch.compile() stack, so as long as there is a triton build that supports XPU backend, it should work out-of-the-box!

msaroufim (approved):

PR #1325, opened 2024-11-22
[float8] Re-enable slow-accum in the bwd of axis-wise scaling schemes

lw:
* #1326
* __->__ #1325

And circumvent the issue with the slow CUTLASS kernel by using the cuBLAS kernel + manual scaling.


torchao/float8/float8_python_api.py
```diff
@@ -37,19 +37,25 @@ def addmm_float8_unwrapped(
...
     b_inverse_scale = b_scale.reciprocal()
 
-    if output_dtype == torch.float32 and bias is not None:
+    post_inverse_scale = None
+    if (
+        a_scale.shape == (a_data.shape[0], 1)
+        and b_scale.shape == (1, b_data.shape[1])
+        and not use_fast_accum
+    ):
+        # The rowwise CUTLASS-based kernel is so slow without fast-accum that
```
vkuzo, line 46: just curious, do we have any OSS shareable evidence (perf/accuracy) on doing this versus rowwise with fast-accum off that we can add here?

lw: Screenshot 2024-12-03 at 19 09 19

I ran a quick benchmark on my H100 with a recent-ish version of PyTorch (nightly from Nov 12). I samples all MxNxK matmul shapes where each of M, N and K is a power of two between 512 and 16384. Here I'm plotting the slowdowns observed when activating slow-accum for the rowwise (CUTLASS-based) and tensorwise (cuBLAS-based) modes

In summary: in tensorwise we get a max slowdown of 50% (usually much less), with rowwise we typically are 2x as slow, with peaks of 4.5x as slow as fast-accum.

(I suspect that for very small shapes the benchmark was CPU-bound hence slow-accum looks as fast as fast-accum, but that's probably misleading)

lw:
Landing since Ruff is already broken on main

torchao/float8/float8_python_api.py
ngimel: fwiw in cuda 12.6.2+ perf of row-wise slow accum kernels is significantly better (slowdown is 50% or so, instead of 2-3x) but separate scaling might still come out ahead.

lw: @ngimel Do you know where those improvements come from? Is it just NVCC becoming better/smarter? Those rowwise kernels are built as part of PyTorch using CUTLASS so I expected CUTLASS upgrades would be more likely to improve perf...

ngimel: Now rerunning benchmarks I see bad perf even on the new version, must have messed something up the last time :-(.

PR #1278, opened 2024-11-13
Add Int4CPULayout and update int4 woq

yanbing-j:
https://github.com/pytorch/pytorch/pull/139611 is merged into PyTorch main branch.

PR #1330, opened 2024-11-22
Update nightly job to use 12.4 since 12.1 is deprecated

jerryzh168:
https://github.com/pytorch/ao/pull/1278#issuecomment-2492933680


jerryzh168:
using https://github.com/pytorch/ao/pull/1333 instead since I can't update this PR somehow

PR #1337, opened 2024-11-24
Update SAM AMG README with more descriptive install instructions

philipbutler:
@cpuhrsch 
Installing was not so obvious, so this should help others trying to test and contribute.

There's some extra commits from amending, I'm still kinda garbage at git


examples/sam2_amg_server/README.md
```diff
@@ -41,7 +41,26 @@ The 'ao' mode is a copy of the baseline with modifications to make the code more
...
+
+# Create and activate a virtual environment
+python -m venv venv && source venv/bin/activate
+
+# Install requirements for this example
+pip install -r examples/sam2_amg_server/requirements.txt
+
+# Install torch and ao nightlies
+#   (try to uninstall first to be safe, use CUDA 12.1 to be safe)
+pip uninstall torch torchao
```
cpuhrsch, line 54: I guess in a fresh environment this step would also be optional

examples/sam2_amg_server/README.md
```diff
@@ -41,7 +41,26 @@ The 'ao' mode is a copy of the baseline with modifications to make the code more
...
+
+pip install --pre torch torchvision torchaudio --index-url https://download.pytorch.org/whl/nightly/cu121
```
cpuhrsch, line 56: I think CUDA 12.4 should be better and available now. The main limitation here is the driver version (see https://docs.nvidia.com/cuda/archive/12.4.0/cuda-toolkit-release-notes/).

examples/sam2_amg_server/README.md
```diff
@@ -41,7 +41,26 @@ The 'ao' mode is a copy of the baseline with modifications to make the code more
...
+pip install --pre torchao --index-url https://download.pytorch.org/whl/nightly/cu121
+
+# Build from source
+python setup.py develop
```
cpuhrsch, line 60: This won't be necessary if the user installed torchao nightly above. This was really just needed to pull in the very latest change and/or to make changes locally and push them more easily. Also I think we had  uninstall torchao from above again so that this works.

examples/sam2_amg_server/README.md
```diff
@@ -41,7 +41,26 @@ The 'ao' mode is a copy of the baseline with modifications to make the code more
...
+
+# On your mark, get set...
+cd examples/sam2_amg_server/
 ```
 
 Download `sam2.1_hiera_large.pt` from https://github.com/facebookresearch/sam2?tab=readme-ov-file#download-checkpoints and put it into `~/checkpoints/sam2`
```
cpuhrsch, line 66: The download part I guess doesn't apply anymore now that the user can get this via the model type argument.

examples/sam2_amg_server/README.md
cpuhrsch: Yeah for now that seems necessary. I also found that the nightly package doesn't include the config files in https://github.com/pytorch/ao/tree/main/torchao/_models/sam2/configs so it's kind of the only choice.

cpuhrsch (approved):

cpuhrsch:
Looks good, let me know once you got a chance to give this another commit or two and then we can merge.

torchao/_models/sam2/configs/sam2/sam2_hiera_b+.yaml
```diff
@@ -2,18 +2,18 @@
 
 # Model
 model:
-  _target_: sam2.modeling.sam2_base.SAM2Base
+  _target_: torchao._models.sam2.modeling.sam2_base.SAM2Base
```
cpuhrsch, line 5: Ah, these are the sam2 configs. There's sam2 and then there's sam2.1. I had to switched to sam2.1, but maybe sam2 is still useful? I updated the sam2.1 configs in https://github.com/pytorch/ao/commit/b2e42ff6b4ed0134684932710a45241910de5fb8#diff-12caf984b30e7028df7fb867a984c5ccad3fde0a0783df2af26db9cef860fbe5 , but in any case it doesn't hurt to have sam2 support as well.

examples/sam2_amg_server/README.md
```diff
@@ -41,7 +41,26 @@ The 'ao' mode is a copy of the baseline with modifications to make the code more
...
+python -m venv venv && source venv/bin/activate
+
+# Install requirements for this example
+pip install -r examples/sam2_amg_server/requirements.txt
+
+# If you have an older version of torch in your current environment, uninstall it first
+pip uninstall torch
+
+# Install torch nightly
+pip install torch torchvision torchaudio
```
cpuhrsch, line 57: Hm, I think this would download the stable versions. You probably need to use the `--pre` variants again unless there's something else that causes nightlies to be selected.

examples/sam2_amg_server/README.md
jerryzh168, line 1: also how does `find sav_val -type f > sav_val_image_paths` in this file work? do we need to download the validation images some where?

cpuhrsch: Yes, let's not include hard links for the dataset download, because like you mentioned @philipbutler you have to agree to terms first. But it's ok to add instructions on where to download the dataset and how to do it. Just, please, no hard links to the actual images.

cpuhrsch:
@bhack - Yes, we're working a  more detailed description of this.

PR #1341, opened 2024-11-25
float8 training: move module attribute setting to sync function

vkuzo:
Summary:

This PR moves the setting of `is_amax_initialized` flag on `Float8Linear` to the `sync_float8_amax_and_scale_history` function.

There are two reasons for this:
1. the current logic does not work with torchtitan + delayed scaling + AC, failing with https://gist.github.com/vkuzo/70819a2cffb9346bf44ecd9079b8bf51 .
2. in general, stateful logic such as changing module attributes adds complexity. Even if we fix (1) in compile land, something else could break.

The `sync_float8_amax_and_scale_history` function is already called outside of the main model forward/backward, it's already required to be called at every iteration, it does not need to know about AC, and it seems like a great place to stash logic which isn't easily compileable such as this init code.

After this PR the `enable_amax_init` and `enable_pre_and_post_forward` config options are now no-ops. In a future PR we should add a deprecation warning, and eventually remove these.

Test Plan:

```
// this repo

// torchtitan
// requires https://github.com/pytorch/torchtitan/pull/698
with-proxy CONFIG_FILE="./train_configs/llama3_8b.toml" ./run_llama_train.sh --float8.enable_float8_linear --training.compile --float8.scaling_type_input delayed --float8.scaling_type_weight delayed --float8.scaling_type_grad_output delayed
```

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/float8/float8_linear_utils.py
+++ b/torchao/float8/float8_linear_utils.py
@@ -193,6 +193,9 @@ def sync_float8_amax_and_scale_history(model: torch.nn.Module, fp8_layers=None) 
             and we loop over all fp8_layers to sync and update amax scale histories.
             Users can use get_float8_layers to get all fp8 layers.
     """
+    # TODO(future): consider adding a flag to control setting the `is_amax_initialized`
+    # flag only on the first iteration.
+
     if fp8_layers is None:
         fp8_layers = get_float8_layers(model)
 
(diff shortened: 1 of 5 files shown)


torchao/float8/float8_linear_utils.py
```diff
@@ -309,10 +312,10 @@ def inner_func():
...
             child.fp8_scale_grad_output.copy_(new_grad_output_scales[idx])
 
-    # This allows for the compile to succede on the inner func and fail on the graph breaks
+    # This allows for the compile to succeed on the inner func and fail on the graph breaks
     # at the beginning and and of syncing
     inner_func()
 
     for child in fp8_layers:
-        # Set a flag to signal amaxes/scales are ready
-        child.amax_and_scale_synced = True
```
vkuzo, line 318: this was dead code which we forgot to delete in a previous PR, deleting here

weifengpy (approved):

Issue #658, opened 2024-08-12
Self compressing neural networks


msaroufim:
Self-Compressing Neural Networks is dynamic quantization-aware training that puts the size of the model in the loss

Paper: https://arxiv.org/pdf/2301.13142
Code: https://github.com/geohot/ai-notebooks/blob/master/mnist_self_compression.ipynb

aiepm:
I'm taking this one.

vedant-rupwal:
Hi everyone! I'm a student in the CodePath AI301 program and I'd love to work on this issue if it is still open and available.

I see there was a mention of it a few months ago, but since it isn't assigned, I'd like to take a crack at reproducing the dynamic QAT behavior described in the paper locally this week. Any pointers on where to start integrating this within the module: qat codebase would be greatly appreciated!

PR #1342, opened 2024-11-25
Self-Compression QAT and Linear

lliangthomas:
Implemented self-compression QAT and linear layer from this paper as a solution to #658 

If there are more features or more integration to be done with the current QAT schemes in TorchAO (self-compression seems orthogonal right now), let me know.

@msaroufim @HDCharles

### DIFF
--- a/torchao/quantization/qat/__init__.py
+++ b/torchao/quantization/qat/__init__.py
@@ -7,6 +7,7 @@ from .embedding import (
 from .linear import (
     Int4WeightOnlyQATQuantizer,
     Int8DynActInt4WeightQATQuantizer,
+    SelfCompressionQATQuantizer
 )
 
 __all__ = [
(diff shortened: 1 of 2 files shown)


torchao/quantization/qat/__init__.py
```diff
@@ -7,11 +7,13 @@
...
     Int8DynActInt4WeightQATQuantizer,
+    SelfCompressionQATQuantizer
 )
 
 __all__ = [
     "ComposableQATQuantizer",
     "Int4WeightOnlyQATQuantizer",
     "Int4WeightOnlyEmbeddingQATQuantizer",
     "Int8DynActInt4WeightQATQuantizer",
+    "SelfCompressionQATQuantizer"
```
andrewor14, line 18: This feels like a pretty experimental feature. I would exclude it from the top level QAT namespace

torchao/quantization/qat/linear.py
```diff
@@ -30,6 +30,82 @@
     _get_qmin_qmax,
 )
 
+class SelfCompressionQATQuantizer(torch.nn.Module):
```
andrewor14, line 33: This feels like a pretty experimental feature. I think this belongs better to a separate file or under torchao/prototype

PR #1344, opened 2024-11-26
Add support for quantize_() with Float8Linear module

jainapurva:
Added support for quantize_() API to work with models trained with float8, using Float8Linear.


drisspg (commented):
I think makes sense, can you add some tests

Some alternatives would be to maintain the Float8Linear structure and then swap the weight class dtype (feels weird).

I think this is good argument for subclasses since you can maintain structure and then assert all low_p subclasses have a `dequant` method and call that which will convert to fp32. I know Christian wants tensor.to(torch.float32) to do this but I think its tooo magical 

cc @vkuzo

torchao/quantization/quant_api.py
```diff
@@ -222,6 +224,9 @@ def _replace_with_custom_fn_if_matches_filter(
     Returns:
         None
     """
+    # If model is Float8Linear, convert it to Linear before moving forward
+    if isinstance(model, Float8Linear):
+        model = dequantize_float8_training(model)
```
vkuzo, line 229: can you just move your code snippet from the other file here:

```
if isinstance(model, Float8Linear):
    with torch.device("meta"):
        new_module = nn.Linear(model.in_features, model.out_features)
    new_module.weight = model.weight
    new_module.bias = model.bias        
    model = new_module
```

and not need any changes to `torchao/float8`?

jerryzh168: @vkuzo what do you think about having dequantizing a model as a separate API? it feels a bit weird to have this logic in `_replace_with_custom_fn_if_matches_filter` which is supposed to be a simple module replacement function I feel.

vkuzo: my 5c - 
1. we can make it work now without adding **any** public APIs, with minimal increase in complexity
2. if it's important to have a public API for "remove low precision training from a model", we can have that conversation in parallel

wdyt

jainapurva: As per the offline discussion with @jerryzh168, we should have a separate exposed API to convert Float8Linear to Linear. The flow would then be:

-> Float8 training
-> Dequantize Float8Linear to Linear
-> quantize_()

vkuzo (changes requested):
would be good to hear some motivation on why this needs a public API versus doing the same thing without a new API

torchao/quantization/quant_api.py
jerryzh168: the motivation for adding a new API is making the dequantizing step more explicit for user, instead of hide it in a module replacement function.

but agree this can happen in parallel. also it's probably not worth spending time to discuss as of now, and wait until there are more use cases might be better

torchao/quantization/quant_api.py
```diff
@@ -199,6 +201,22 @@ def change_linear_weights_to_int4_woqtensors(
 ########
 # TO BE DEPRECATED END
 ########
+def dequantize_float8_training(model: nn.Module) -> nn.Module:
```
jerryzh168, line 204: we want to inline this function in `_replace_with_custom_fn_if_matches_filter` for now?

jainapurva: Done

test/float8/test_base.py
```diff
@@ -751,7 +766,7 @@ def test_swap_root_linear_with_children_raises(self):
             config = Float8LinearConfig(emulate=emulate)
             with self.assertRaisesRegex(
                 AssertionError,
-                "Does not support a root nn.Linear with children",
+                "Does not support a root torch.nn.modules.linear with children",
```
jerryzh168, line 769: nit: might be better to revert this change

jerryzh168 (approved):
LGTM

test/float8/test_base.py
jainapurva: I reverted it.

PR #1351, opened 2024-11-26
[wip] float8 training: invert the meaning of scale

vkuzo:
Summary:

Needs some work before review.

Benefit 1: less kernels for delayed scaling, so easier to debug things.
Benefit 2: aligns with other tensor subclasses in torchao better.

Test Plan:

Some FSDP2 numerics tests are still failing, TODO debug

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/float8/float8_python_api.py
+++ b/torchao/float8/float8_python_api.py
@@ -34,8 +34,10 @@ def addmm_float8_unwrapped(
     as inputs. This is used to standardize the logic between subclassed and non subclassed
     versions of the linear module.
     """
-    a_inverse_scale = a_scale.reciprocal()
-    b_inverse_scale = b_scale.reciprocal()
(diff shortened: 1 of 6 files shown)


torchao/float8/float8_python_api.py
```diff
@@ -34,8 +34,10 @@ def addmm_float8_unwrapped(
     as inputs. This is used to standardize the logic between subclassed and non subclassed
     versions of the linear module.
     """
-    a_inverse_scale = a_scale.reciprocal()
-    b_inverse_scale = b_scale.reciprocal()
+    # a_inverse_scale = a_scale.reciprocal()
```
vkuzo, line 37: not having to do this is what reduces the # of kernels for delayed scaling

PR #1353, opened 2024-11-26
Benchamarking

metascroy:
Summary: Add benchmarks for experimental torchao kernels.

Differential Revision: D66512859

### DIFF
--- a/torchao/_models/llama/generate.py
+++ b/torchao/_models/llama/generate.py
@@ -217,7 +217,6 @@ def main(
             float8_weight_only,
             float8_dynamic_activation_float8_weight,
         )
-        from torchao.prototype.quantization.autoquant_v2 import autoquant_v2
         from torchao.utils import unwrap_tensor_subclass
 
         from torchao.quantization.granularity import PerTensor, PerRow
@@ -297,6 +296,29 @@ def main(
             dtype = _NBITS_TO_DTYPE[nbits]
             group_size = int(_quant_args[2])
             quantize_(model, uintx_weight_only(dtype, group_size, use_hqq=use_hqq))
+        elif "int8_dynamic_activation_intx_weight" in quantization:
+            from torchao.experimental.quant_api import int8_dynamic_activation_intx_weight
+            assert precision == torch.float32, "int8_dynamic_activation_intx_weight requires fp32 precision"
+
+            # Build kernels in temp location, and load them in torch
+            # This requires an ARM CPU
+            from torchao.experimental.temp_build import temp_build_and_load_torchao_ops
+            temp_build_and_load_torchao_ops(cmake_lists_path=os.path.dirname(os.path.realpath(__file__)) + "/../../experimental")
+
+            # Quantize model
+            _quant_args = quantization.split("-")
+            nbit = int(_quant_args[1])
+            assert nbit >= 1 and nbit <= 8, "nbits must be 1 to 8"
+            group_size = int(_quant_args[2])
+            has_weight_zeros = bool(_quant_args[3])
+            quantize_(
+                model,
+                int8_dynamic_activation_intx_weight(
+                    group_size=group_size,
+                    nbit=nbit,
+                    has_weight_zeros=has_weight_zeros,
+                ),
+            )
         elif "float8wo" in quantization:
             quantize_(model, float8_weight_only())
         elif "float8dq" in quantization:
@@ -309,6 +331,7 @@ def main(
                 granularity = PerTensor()
             quantize_(model, float8_dynamic_activation_float8_weight(granularity=granularity))
         elif "autoquant_v2" in quantization:
+            from torchao.prototype.quantization.autoquant_v2 import autoquant_v2
             from torchao._models._eval import InputRecorder
             from torchao._models.llama.model import prepare_inputs_for_model
 
--- a/torchao/experimental/temp_build.py
+++ b/torchao/experimental/temp_build.py
@@ -1,0 +1,43 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the license found in the
+# LICENSE file in the root directory of this source tree.
+
+import glob
+import subprocess
+import tempfile
+import torch
+
+def cmake_build_torchao_ops(cmake_lists_path, temp_build_dir):
+    from distutils.sysconfig import get_python_lib
+    print("Building torchao ops for ATen target")
+    cmake_prefix_path = get_python_lib()
+    subprocess.run(
+        [
+            "cmake",
+            "-DCMAKE_PREFIX_PATH=" + cmake_prefix_path,
+            "-DCMAKE_INSTALL_PREFIX=" + temp_build_dir.name,
+            "-S " + cmake_lists_path,
+            "-B " + temp_build_dir.name,
+        ]
+    )
+    subprocess.run(
+        [
+            "cmake",
+            "--build",
+            temp_build_dir.name,
+            "-j 16",
+            "--target install",
+            "--config Release",
+        ]
+    )
+
+def temp_build_and_load_torchao_ops(cmake_lists_path):
+    temp_build_dir = tempfile.TemporaryDirectory()
+    cmake_build_torchao_ops(cmake_lists_path, temp_build_dir)
+    libs = glob.glob(f"{temp_build_dir.name}/lib/libtorchao_ops_aten.*")
+    libs = list(filter(lambda l: (l.endswith("so") or l.endswith("dylib")), libs))
+    assert len(libs) == 1
+    torch.ops.load_library(libs[0])
+    print(f"TorchAO ops are loaded from {libs[0]}")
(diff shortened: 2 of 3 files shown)


torchao/_models/llama/generate.py
```diff
@@ -297,6 +296,60 @@ def main(
...
+        elif "int8_dynamic_activation_intx_weight" in quantization:
+            from torchao.experimental.quant_api import int8_dynamic_activation_intx_weight
+            assert precision == torch.float32, "int8_dynamic_activation_intx_weight requires fp32 precision"
+
+            # Build kernels in temp location, and load them in torch
+            import glob
+            import os
+            import subprocess
+            import tempfile
+            def cmake_build_torchao_ops(temp_build_dir):
```
jerryzh168, line 308: is this required? why not add to torchao default build?

jerryzh168 (approved):
LG thanks, had one question/comment

torchao/_models/llama/generate.py
jerryzh168: is this is required, I'd suggest to put these in a util

metascroy: Happy to add it to a default build for ARM CPU if you have a code pointer?

jerryzh168: that would be great, here is our build code for kernels I think: https://github.com/pytorch/ao/blob/5eb6339e0b6f413c74a3dfd5e7f53449474723fc/setup.py#L53

metascroy: It doesn't look like cmake is being used there, and it would require more refactoring to use it.  I factored out the install script into a utility, though.

jerryzh168: OK sounds good, we can start with this for now, maybe put this in setup.py in a separate PR. but I do feel we should do that so it's easy to use, @metascroy do you want to do that in a follow up PR?

jerryzh168: cc @msaroufim and @atalman if any of you can provide some guidance on how we can add cmake build script to torchao/setup.py that would be great

metascroy: Ok I'll land this for now.  On creating build scripts in torchao/setup.py, I'll leave it up to you all.  We already have build scripts for these kernels in ExecuTorch and torchchat, so they are not difficult to use on those surfaces.

jerryzh168: but if you add build scripts to torchao, will you still need to maintain these in torchchat and executorch?

metascroy: Yes, because both executorch/torchchat need to build executorch kernels, but if we added build scripts to torchao, they'd likely only build aten kernels (unless torchao takes a dependency on executorch).

jerryzh168: OK, so looks like it's the same kernel implementation, but different build script for torchao/executorch/torchchat in this case

torchao/_models/llama/generate.py
```diff
@@ -297,6 +296,29 @@ def main(
...
             group_size = int(_quant_args[2])
             quantize_(model, uintx_weight_only(dtype, group_size, use_hqq=use_hqq))
+        elif "int8_dynamic_activation_intx_weight" in quantization:
+            from torchao.experimental.quant_api import int8_dynamic_activation_intx_weight
+            assert precision == torch.float32, "int8_dynamic_activation_intx_weight requires fp32 precision"
+
+            # Build kernels in temp location, and load them in torch
+            # This requires an ARM CPU
+            from torchao.experimental.temp_build import temp_build_and_load_torchao_ops
+            temp_build_and_load_torchao_ops(cmake_lists_path=os.path.dirname(os.path.realpath(__file__)) + "/../../experimental")
```
kimishpatel, line 306: I dont follow why we are doing this way? Why can we not just try to load the ops and if not found raise exception with build steps needed? Ideally we can detect the platform and build this deps as part of some pre-req for running benchmarks under _model directory. When we are able to ship these kernels as part of pip package, we may not need this

torchao/_models/llama/generate.py
kimishpatel: Example in et for cmake https://github.com/pytorch/executorch/blob/main/setup.py#L496

torchao/_models/llama/generate.py
jerryzh168: > When we are able to ship these kernels as part of pip package, we may not need this

what is the blocker for this?

kimishpatel: we need to move out of the experimental. I think we need to follow up on this. I believe we have sufficient evidence now? @supriyar ?

metascroy: > I dont follow why we are doing this way? Why can we not just try to load the ops and if not found raise exception with build steps needed? Ideally we can detect the platform and build this deps as part of some pre-req for running benchmarks under _model directory. When we are able to ship these kernels as part of pip package, we may not need this

I guess by "just try to load the ops" you mean something like this: https://github.com/pytorch/executorch/blob/main/extension/llm/custom_ops/sdpa_with_kv_cache.py#L21-L34 (but with the except block replaced by install instructions)

We could do that, but in the current setup, won't the try block always fail when running this benchmarking script (unless we build/load the ops in setup.py)?  I'm also not sure what the instructions would say to make the script runnable without telling the user to modify the script by adding a torch.load_library line.  I guess we could ask them to define an environment variable with the library location?

kimishpatel: > but with the except block replaced by install instructions

> I'm also not sure what the instructions would say to make the script runnable without telling the user to modify the script by adding a torch.load_library line.

@metascroy Yes, but I dont follow the second part. Why user needs to modify the script. Can we not just `import torchao.experimental.lowbit_ops` which internally does try/except? But I do kind of get what you are doing here because any build scripts will ahve to figure out where to put the build artifact (.so) and then we need to load from there.

Ideally it should be installed as part of the setup instructions or pip package. So we can also follow something like https://github.com/pytorch/ao/blob/main/setup.py#L53 and add extra option to build with experimental lowbit quant features. So if user invokes the benchmarking script without building experimental kernels you can suggest please do `python setup.py --use_experimental_low_bit_kernels` or `pip install . --use_experimental_low_bit_kernels`. This feels a bit cleaner to me but curious to know your thoughts and also from @msaroufim

metascroy: > Can we not just import torchao.experimental.lowbit_ops which internally does try/except

This is not going to know where to load the ops with torch.load_library unless it is responsible for both building and loading.  There are a few options we could do:

1) If a user tries to load the ops, ask the user to build the ops and define the location in some environment variable like TORCHAO_OP_LIB_PATH.  This is external to the benchmarking script.  

2) If user tries to load the ops, ask the user to run the build script we provide in experimental (experimental/build_torchao_ops.sh).  We can look for ops where build script places them if TORCHAO_OP_LIB_PATH is not defined.  This is close to what is currently done in torchchat.

3) Build these ops as part of torchao's setup.py.  I'm not sure how well this works with cmake, though.

kimishpatel: 3 is my preference as I suggested earlier.

> I'm not sure how well this works with cmake, though.

Any specific concern or just general feeling?

metascroy: The concern is whether we can reuse the cmake setup stuff we already have (e.g., this function that sets up the parallel compute: https://github.com/pytorch/ao/blob/main/torchao/experimental/Utils.cmake#L7)?  If we bring in KleidiAI/CPUInfo via CMake, that will be more stuff to worry about.

I haven't used the torch CppExtension in setup.py, but it looks fairly simplistic compared to cmake.  Perhaps we could do something like what this blog does, if it does not already exist in PyTorch: https://martinopilia.com/posts/2018/09/15/building-python-extension.html

kimishpatel: Maybe I am missing something but Executorch's pybinding extension with xnnpack builds xnnpack which build with cpuinfo and pthreadpool and everything. Granted that is a whole lot to build but it does work in ET. So likely there is a nuance that you are worried about that I am not understanding

metascroy: I guess this goes back to one of my earlier comments to @jerryzh168: the current setup.py in torchao does not really support cmake, and it would require a good amount of refactoring to support it.  Currently setup.py in torchao is built on utilities in torch.utils.cpp_extension, which look somewhat simplistic and as far as I can tell, do not support cmake.

ET's setup.py defines a custom extension to support cmake and doing something similar to torchao looks like a sizable refactor to their setup?

kimishpatel: Ok I will have to rely on your answer for this since I havent looked at all the details of cpp_extension. I do remember it being simple but I dont know if there are ways to add deps as part of cpp_extension. I doubt but see if possible.

If not, I think it is worth proposing this for the sake of making ARM kernels available on mac builds. @drisspg any thoughts? discussion here is largely around more complex setup.py in order to allow us to build cpp package extensions that package cpu kernels and make it available as part of python package

metascroy: I edited setup.py to build the experimental kernels here: [D67777662](https://www.internalfb.com/diff/D67777662)

Need feedback from torchao folks on whether the changes are acceptable.

PR #1356, opened 2024-11-27
Update hardware check conditions

jainapurva:
Replace cuda architecture checks with the check function from torchao/utils:

- is_cuda_8_9 -> is_sm_at_least_89
- is_cuda_9_0 -> is_sm_at_least_90


jerryzh168 (approved):
thanks!

PR #1366, opened 2024-12-02
Lint fixes torchao files

jainapurva:
Lint fixing for

- torchao/utils.py
- torchao/ops.py
- torchao/_executorch_ops.py


jerryzh168 (approved):
thanks, we can ignore the bad alloc issue for now I think

PR #1372, opened 2024-12-03
Add INT8 SDPA path for CPU

Valentine233:
For the integration of INT8 SDPA in TorchAO, we design a feasible path by registering a customized pass of PyTorch and adding the pattern matcher and kernel in TorchAO.

**Steps:**
1. Register and implement the INT8 SDPA kernel, i.e. `torchao.ops.scaled_dot_product_int8`, for CPU.
2. Add the pattern matchers for INT8 SDPA., in order to replace the decomposed OPs with `torchao.ops.scaled_dot_product_int8`.
3. Register a customized pass of PyTorch by defining the above patterns as `torch._inductor.config.post_grad_custom_pre_pass`.

**Perf:**
The validation is launched for int8-bf16 on GNR machine. The script is similar as the [UT](https://github.com/pytorch/ao/pull/1372/files#diff-d96f56237c6ec522900d166495c242d9dd49139041553786b6f43f6d8f227aebR152).

| Model  | Mode  | E2E Speedup |
| ----------- | ------------ | ------------ |
| BertLarge | Throughput | 1.13 |
| BertLarge | Realtime | 1.03 |
| VIT | Throughput | 1.10 |
| VIT | Realtime | 1.03 |


Valentine233:
@drisspg @jerryzh168 @jgong5 @leslie-fang-intel Please help review for the POC, thanks!

drisspg:
WIll do some more review on this later today but I think we might want to make a sub folder in quantization for 'inductor_patterns" since this is a pretty particular workflow and we should add a good readme explaining how it can be used and its limitations

vkuzo:
> Register a customized pass of PyTorch by defining the above patterns as torch._inductor.config.joint_custom_pre_pass.

Do I understand correctly that this using torch.compile to change the numerics of the model by hooking up a quantization pass to inductor?  If yes, can this live in prototype for now? I'd have concerns about using torch.compile passes to change numerics being the official API, some of the challenges here include breaking the assumption that a compiler does not meaningfully change numerics.

jerryzh168:
> > Register a customized pass of PyTorch by defining the above patterns as torch._inductor.config.joint_custom_pre_pass.
> 
> Do I understand correctly that this using torch.compile to change the numerics of the model by hooking up a quantization pass to inductor? If yes, can this live in prototype for now? I'd have concerns about using torch.compile passes to change numerics being the official API, some of the challenges here include breaking the assumption that a compiler does not meaningfully change numerics.

I believe numerics changes happens in pt2e quant api this should only do fusion

torchao/quantization/sfdp_int8_fx_pass.py
```diff
@@ -0,0 +1,686 @@
...
+            "extra_check": extra_check,
+            "scalar_workaround": workaround,
+        }
+
+
+@functools.lru_cache(None)
+def _sfdp_init_int8():
+    for key, register_replacement_kwargs in _gen_sfdp_patterns_int8():
+        register_replacement(**register_replacement_kwargs)
+    config.joint_custom_pre_pass = patterns.apply
```
jerryzh168, line 686: is this the official API to add new fusion passes to inductor? what if we have multiple fusion passes that we need to add? i.e. we probably want to move all intel quant passes to torchao in the future as well

torchao/quantization/sfdp_int8_fx_pass.py
```diff
@@ -0,0 +1,686 @@
...
+    v_scale,
+    a_zp,
+    a_scale,
+    o_zp,
+    o_scale,
+    dropout,
+):
+    # int8-mix-fp32 QUANTIZED SDPA with mask
+    q = query.permute([0, 2, 1, 3])
+    q = torch.ops.quantized_decomposed.dequantize_per_tensor.default(
```
jerryzh168, line 46: we are moving away from these ops, it would be better if you can use our newer quantize_affine/dequantize_affine ops directly from https://github.com/pytorch/ao/blob/main/torchao/quantization/quant_primitives.py

torchao/quantization/sfdp_int8_fx_pass.py
Valentine233: According to https://github.com/pytorch/pytorch/blob/main/torch/_inductor/config.py#L165, I suppose only one fusion pass could be assigned, also the same case for other customized passes. It is better to expand this to a list of passes. Maybe need more comments from @Chillee @eellison.

vkuzo:
> > > Register a customized pass of PyTorch by defining the above patterns as torch._inductor.config.joint_custom_pre_pass.
> > 
> > 
> > Do I understand correctly that this using torch.compile to change the numerics of the model by hooking up a quantization pass to inductor? If yes, can this live in prototype for now? I'd have concerns about using torch.compile passes to change numerics being the official API, some of the challenges here include breaking the assumption that a compiler does not meaningfully change numerics.
> 
> I believe numerics changes happens in pt2e quant api this should only do fusion

I see that `rtol` and `atol` are pretty significant when comparing the baseline to the workflow introduced in this PR, is that intended?  Are you saying that with the proper setup we can do the comparison between using this inductor pass vs not using it to pass with rtol/atol near zero?

Valentine233:
> > > > Register a customized pass of PyTorch by defining the above patterns as torch._inductor.config.joint_custom_pre_pass.
> > > 
> > > 
> > > Do I understand correctly that this using torch.compile to change the numerics of the model by hooking up a quantization pass to inductor? If yes, can this live in prototype for now? I'd have concerns about using torch.compile passes to change numerics being the official API, some of the challenges here include breaking the assumption that a compiler does not meaningfully change numerics.
> > 
> > 
> > I believe numerics changes happens in pt2e quant api this should only do fusion
> 
> I see that `rtol` and `atol` are pretty significant when comparing the baseline to the workflow introduced in this PR, is that intended? Are you saying that with the proper setup we can do the comparison between using this inductor pass vs not using it to pass with rtol/atol near zero?

There exists some numeric issues for the kernel currently, and I would update it finally. Ideally, the rtol is expected to be 1.0 for quantized dtype.

test/test_ops.py
```diff
@@ -38,502 +39,623 @@
...
+
+        self.assertEqual(actual, math_ref, atol=3.0, rtol=5e-6)
+
+    # def _create_floatx_inputs(self, ebits: int, mbits: int, BS: int, OC: int, IC: int, device, dtype):
+    #     # Randomly initialize each byte
+    #     nbits = 1 + ebits + mbits
+    #     floatx_weight = torch.randint(256, (OC, IC // 8 * nbits), dtype=torch.uint8)
+    #     scale = torch.rand(OC).to(dtype) + 0.5
+    #     fp16_act = torch.rand(BS, IC).to(dtype) + 0.5
+    #     return floatx_weight.to(device), scale.to(device), fp16_act.to(device)
```
drisspg, line 169: fix before land

torchao/ops.py
```diff
@@ -71,6 +72,56 @@ def _(
...
+    q_scale: float = 1.0,
+    k_zp: int = 0,
+    k_scale: float = 1.0,
+    v_zp: int = 0,
+    v_scale: float = 1.0,
+    a_zp: int = 0,
+    a_scale: float = 1.0,
+    o_zp: int = 0,
+    o_scale: float = 1.0,
+) -> Tensor:
```
drisspg, line 93: Lets add a good doc block here of all the args and their semantis

torchao/csrc/cpu/sdpa.cpp
```diff
@@ -0,0 +1,2168 @@
...
+  TORCH_CHECK(!attn_mask.defined() ||
+          (attn_mask.dim() == 2 || attn_mask.dim() == 4),
+    "_scaled_dot_product_int8_cpu: Attention mask dim in {2, 4}");
+
+  // fallback math path
+  // at::Tensor output = sdpa_int8_math_impl(query, key, value,
+  //   dropout_p, is_causal, attn_mask, scale,
+  //   q_zp, q_scale,
+  //   k_zp, k_scale,
+  //   v_zp, v_scale,
```
drisspg, line 2143: decent amount of commented code here

vkuzo:
@jerryzh168  or @drisspg , could you help make sure the PR summary includes the high level description of the flow and explains why the inductor pass is used to swap to the final kernel, this isn't expected, it's somewhat makes sense to me but it would be good to just explain what the plan is in a way that's easily discoverable

test/quantization/test_sfdp_int8_fx_pass.py
```diff
@@ -0,0 +1,199 @@
...
+        context_layer = torch.matmul(attention, v)
+        context_layer = context_layer.permute(0, 2, 1, 3).contiguous()
+        context_layer = context_layer.view(
+            context_layer.size()[:-2] + (self.all_head_size,)
+        )
+        return self.dense(context_layer)
+
+def _generate_qdq_quantized_model(mod, inputs, quantizer):
+    with torch.no_grad():
+        export_model = capture_pre_autograd_graph(mod, inputs)
```
jerryzh168, line 78: `capture_pre_autograd_graph` is deprecated, please use `export_for_training`: https://pytorch.org/tutorials/prototype/pt2e_quant_ptq.html

jerryzh168:
@vkuzo are you referring to `atol=1.0` (https://github.com/pytorch/ao/pull/1372/files#diff-fe0aa67bd65dd5da118abe44f45104170ba9871cb14bd446a448d056599d462aR189)? this does look a bit large. I think there is some expected difference when we fuse `dq - op - q` patterns to real quantized ops, it's probably expected to have slight changes to numerics, so it's a bit different compared to floating point fusions

torchao/ops.py
```diff
@@ -71,6 +72,56 @@ def _(
     return _in_feats.new_empty((BS, OC))
 
 
+def scaled_dot_product_int8(
```
jerryzh168, line 75: please add `for_cpu` to the op name, since we will likely add some gpu ops as well

test/test_ops.py
Valentine233: Fixed.

test/quantization/test_sfdp_int8_fx_pass.py
Valentine233: Thanks, fixed.

msaroufim:
@Valentine233 do you have any perf benchmarks? Naively I would have imagined that SDPA isn't as crucial to implement on CPU because the caches are are larger and hardware managed

Valentine233:
> do you have any perf benchmarks? Naively I would have imagined that SDPA isn't as crucial to implement on CPU because the caches are are larger and hardware managed but perhaps my mathematical intuition of what CPU SDPA is lacking so anything you could share would be super helpful

We also encounter cache problems on CPU for long sequence lengths, so the fused SDPA does have perf benefits. I would share the results in the descriptions, as soon as they are prepared.

torchao/csrc/cpu/int8_sdpa.cpp
```diff
@@ -0,0 +1,2382 @@
...
+    int64_t in_rows,
+    int64_t in_cols,
+    int64_t ldi,
+    int64_t ldo) {
+  for (int64_t r=0; r<in_rows; r++) {
+    for (int64_t c=0; c<in_cols; c++) {
+      *(dst + c * ldo + r) = *(src + r * ldi + c);
+    }
+  }
+}
```
sanchitintel, line 856: nit: Does compiler auto-vectorization work well in this case? I'm guessing you've already determined that this code isn't significant enough to optimize. Just curious.

Thanks

torchao/csrc/cpu/int8_sdpa.cpp
```diff
@@ -0,0 +1,2382 @@
...
+
+template <typename scalar_t>
+inline void do_copy(
```
sanchitintel, line 859: nit: Not used anywhere?

sanchitintel: Resolved, thanks!

torchao/csrc/cpu/int8_sdpa.cpp
Valentine233: Thanks! The transposition code is not significantly time-consuming, so this is currently a scalar version. The vectorized code would be added in the future if necessary.

sanchitintel: OK, thanks!

torchao/ops.py
Valentine233: Thanks, the doc block is added.

torchao/csrc/cpu/sdpa.cpp
Valentine233: Thanks, the unused codes are removed.

torchao/ops.py
Valentine233: This is supposed to be used by all the backends, and each backend could register its own implementation. For example, for CPU path https://github.com/pytorch/ao/pull/1372/files#diff-eaf2387d03cf16395487f5f4162420a8e84bb89e9f1221a01474d2f87ff449ddR2080-R2082.

Or do you think the API for CPU and GPU could be different?

Valentine233:
@jerryzh168 @drisspg Please help review the PR, thanks!
BTW, do you know how to get the permission to trigger CI and add labels for TorchAO repo?

Valentine233:
@jerryzh168 @drisspg Hi, could you review this PR when you are free? And is it possible to give me the permission for triggering CI and adding label? Thanks so much!

setup.py
```diff
@@ -53,6 +53,10 @@ def read_version(file_path="version.txt"):
     and platform.system() == "Darwin"
 )
 
+use_cpp_avx512 = os.getenv("USE_AVX512", "1") == "1" and platform.system() == "Linux"
```
sanchitintel, line 56: Can we use PyTorch APIs such as `torch._C._cpu._is_avx512_supported`? Thanks!

Valentine233: Thanks for the suggestion. It works!

Valentine233: Update: The API is new, so we would encounter error for torch with lower versions.

Valentine233:
@jerryzh168 @drisspg Hi, the PR has been rebased and all CI tests pass now. Please help review, thanks!
Also see the perf results in description, cc @msaroufim.

leslie-fang-intel (commented):
The added CI seems failed, @Valentine233 please take a look.

setup.py
```diff
@@ -305,6 +320,12 @@ def get_extensions():
 
     # Collect C++ source files
     sources = list(glob.glob(os.path.join(extensions_dir, "**/*.cpp"), recursive=True))
+    if IS_WINDOWS:
+        # Remove csrc/cpu/*.cpp on Windows due to the link issue: unresolved external symbol PyInit__C
+        cpp_sources = list(
```
drisspg, line 325: nit; "excluded_sources"

torchao/ops.py
drisspg: I think its fine as is, we are planning to add a variant for FAv3 like fp8 attention which would be slightly different I imagine, but it will be prototype

cc @jbschlosser

setup.py
Valentine233: Thanks, modified!

Valentine233:
> The added CI seems failed, @Valentine233 please take a look.

Thanks, fixed!

test/prototype/inductor/test_int8_sdpa_fusion.py
```diff
@@ -0,0 +1,215 @@
...
+                else None,
+            )
+            enable_autocast = dtype == torch.bfloat16
+            with (
+                torch.no_grad(),
+                torch.amp.autocast(
+                    self.device, enabled=enable_autocast, dtype=torch.bfloat16
+                ),
+            ):
+                _int8_sdpa_init()
```
leslie-fang-intel, line 184: I think it should be invoked when user imports `torchao`, why we need to register these passes explicitly?

torchao/quantization/sfdp_int8_fx_pass.py
leslie-fang-intel: Can we report an issue to track it? If multiple libraries register the `joint_custom_pre_pass`, it will fail to work implicitly.

torchao/prototype/inductor/fx_passes/int8_sdpa_fusion.py
```diff
@@ -0,0 +1,581 @@
...
+    )
+    _register_int8_sdpa_pattern(
+        _get_int8_sdpa_final_pattern(
+            has_mask=True, is_batch_size_1=True, is_reduced_type=True, has_convert=False
+        )
+    )
+
+
+def _register_int8_sdpa_lowerings():
+    _register_int8_sdpa_fp32_lowering()
```
leslie-fang-intel, line 568: instead of register it one by one, can we put them into arguments of a register function?

torchao/prototype/inductor/fx_passes/int8_sdpa_fusion.py
```diff
@@ -0,0 +1,581 @@
...
+            aten.expand.default,
+            int8_sdpa_softmax_pattern,
+            Arg(),
+        ),
+        Arg(),
+    )
+
+
+def _get_int8_sdpa_final_pattern(
+    has_mask: bool, is_batch_size_1: bool, is_reduced_type: bool, has_convert: bool
```
leslie-fang-intel, line 389: Please write the comments for these arguments

torchao/prototype/inductor/fx_passes/int8_sdpa_fusion.py
```diff
@@ -0,0 +1,581 @@
...
+            CallFunction(
+                aten.clone.default,
+                int8_sdpa_k_basic_pattern,
+                memory_format=Arg(),
+            ),
+            Arg(),
+        )
+
+
+def _get_int8_sdpa_v_pattern(is_batch_size_1: bool, has_convert: bool):
```
leslie-fang-intel, line 191: Can we merge the `q/k/v` pattern into one function?

torchao/csrc/cpu/int8_sdpa.cpp
```diff
@@ -0,0 +1,1918 @@
...
+        sdpa_int8_fused_kernel(output, query, key, value,
+            dropout_p, is_causal, attn_mask, scale,
+            q_zp, q_scale,
+            k_zp, k_scale,
+            v_zp, v_scale,
+            a_zp, a_scale,
+            o_zp, o_scale);
+        return output.transpose(1, 2);
+    } else {
+        return sdpa_int8_math_kernel(query, key, value,
```
leslie-fang-intel, line 1891: Feels we can merge these 2 `sdpa_int8_math_kernel`.

torchao/csrc/cpu/int8_sdpa.cpp
```diff
@@ -0,0 +1,1918 @@
...
+  }
+  // Heuristic to decide whether to use one parallel loop or not
+  uint32_t l2_cache_size = at::cpu::L2_cache_size();
+  int64_t num_thread = at::get_num_threads();
+  int64_t attn_size = q_split_size * kv_seq_len * sizeof(int32_t) * num_thread;
+  bool use_one_parallel_loop = (batchSize * num_head > num_thread) &&
+      (attn_size > 1.5 * l2_cache_size);
+  if (use_one_parallel_loop) {
+    if (!attn_mask.has_value()) {
+      if (q_split_size == 256) {
```
leslie-fang-intel, line 1671: Please merge these 2 branches for use_one_parallel_loop

test/prototype/inductor/test_int8_sdpa_fusion.py
Valentine233: This could not be registered unless we do the explicit invoking, just like https://github.com/pytorch/pytorch/blob/main/torch/_inductor/fx_passes/post_grad.py#L117.

torchao/prototype/inductor/fx_passes/int8_sdpa_fusion.py
Valentine233: Thanks, modified!

torchao/prototype/inductor/fx_passes/int8_sdpa_fusion.py
Valentine233: Thanks, added!

torchao/prototype/inductor/fx_passes/int8_sdpa_fusion.py
Valentine233: Thanks, modified!

torchao/csrc/cpu/int8_sdpa.cpp
Valentine233: Thanks, modified!

torchao/csrc/cpu/int8_sdpa.cpp
Valentine233: Thanks, modified!

leslie-fang-intel (approved):

Valentine233:
@drisspg @jerryzh168 Hi, do you have any other comments for this PR?

test/prototype/inductor/test_int8_sdpa_fusion.py
```diff
@@ -0,0 +1,215 @@
...
+        not TORCH_VERSION_AT_LEAST_2_7, reason="int8 sdpa requires torch 2.7 or later"
+    )
+    @pytest.mark.skipif(IS_WINDOWS, reason="int8 sdpa does not support windows yet")
+    @config.patch({"freezing": True})
+    def _test_sdpa_int8_rewriter(self):
+        import torch.ao.quantization.quantizer.x86_inductor_quantizer as xiq
+        from torch.ao.quantization.quantize_pt2e import convert_pt2e, prepare_pt2e
+        from torch.ao.quantization.quantizer.x86_inductor_quantizer import (
+            X86InductorQuantizer,
+        )
```
jerryzh168, line 154: please import from `torchao.quantization.pt2e.quantize_pt2e` and `torchao.quantization.pt2e.quantizer` now, we just duplicated these into torchao

test/prototype/inductor/test_int8_sdpa_fusion.py
```diff
@@ -0,0 +1,215 @@
...
+                )
+                export_model = export_for_training(
+                    mod,
+                    inputs,
+                    strict=True,
+                ).module()
+                prepare_model = prepare_pt2e(export_model, quantizer)
+                prepare_model(*inputs)
+                convert_model = convert_pt2e(prepare_model)
+                torch.ao.quantization.move_exported_model_to_eval(convert_model)
```
jerryzh168, line 198: same for this one, `torchao.quantization.pt2e.move_exported_model_to_eval`

torchao/prototype/inductor/fx_passes/int8_sdpa_fusion.py
```diff
@@ -0,0 +1,370 @@
...
+    )
+    if input_name == "key":
+        # do transpose
+        int8_sdpa_qkv_pattern_before_dequant = CallFunction(
+            aten.permute.default,
+            int8_sdpa_qkv_pattern_before_dequant,
+            Arg(),
+        )
+    int8_sdpa_qkv_basic_pattern = CallFunction(
+        torch.ops.quantized_decomposed.dequantize_per_tensor.default,
```
jerryzh168, line 117: just FYI, we will want to deprecate this op and replace it with `torch.ops.torchao.dequantize_affine` in the future

test/prototype/inductor/test_int8_sdpa_fusion.py
```diff
@@ -0,0 +1,215 @@
...
+
+    def transpose_for_scores(self, x: torch.Tensor) -> torch.Tensor:
+        new_x_shape = x.size()[:-1] + (
+            self.num_attention_heads,
+            self.attention_head_size,
+        )
+        x = x.view(new_x_shape)
+        return x.permute([0, 2, 1, 3])
+
+    def forward(self, x, mask):
```
jerryzh168, line 49: do we also want to support quantizing F.sdpa?

jerryzh168 (approved):

test/prototype/inductor/test_int8_sdpa_fusion.py
Valentine233: Do you mean to fuse `dequant+sdpa+quant` into `quantized_sdpa`? I suppose we have no case like this now. Can add more fusion patterns in the future if needed.

jerryzh168:
@leslie-fang-intel @Xia-Weiwen do you want to move other inductor quant passes here as well

test/prototype/inductor/test_int8_sdpa_fusion.py
jerryzh168: oh I mean, F.sdpa -> breakdown to smaller ops -> quantize

Xia-Weiwen:
> @leslie-fang-intel @Xia-Weiwen do you want to move other inductor quant passes here as well

I think so since the `q` and `dq` ops are deprecated in PyTorch core https://github.com/pytorch/ao/pull/1372#discussion_r2049907829. We may need to think more about it, such as the pace to do it and how to keep backward compatibility. We may also need to add more registration points for custom passes in Inductor to accommodate all passes. Leslie may comment more if needed. Thanks.

test/prototype/inductor/test_int8_sdpa_fusion.py
Valentine233: Thanks and updated!

test/prototype/inductor/test_int8_sdpa_fusion.py
Valentine233: Thanks and updated!

torchao/prototype/inductor/fx_passes/int8_sdpa_fusion.py
Valentine233: Well noted, thanks!

torchao/ops.py
```diff
@@ -49,6 +49,9 @@
     "mx_fp4_bf16(Tensor a, Tensor b, Tensor a_scale, Tensor b_scale) -> Tensor",
     tags=[torch._C.Tag.needs_fixed_stride_order],
 )
+lib.define(
```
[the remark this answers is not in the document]
Valentine233: Thanks and modified!

leslie-fang-intel:
> > @leslie-fang-intel @Xia-Weiwen do you want to move other inductor quant passes here as well
> 
> I think so since the `q` and `dq` ops are deprecated in PyTorch core [#1372 (comment)](https://github.com/pytorch/ao/pull/1372#discussion_r2049907829). We may need to think more about it, such as the pace to do it and how to keep backward compatibility. We may also need to add more registration points for custom passes in Inductor to accommodate all passes. Leslie may comment more if needed. Thanks.

Yean, since the new `q` and `dq` ops like `torch.ops.torchao.dequantize_affine` will be registered in TorchAO, I feel we need to register patterns with these new ops in TorchAO.

test/prototype/inductor/test_int8_sdpa_fusion.py
Valentine233: Now we firstly do the exporting, and then pt2e. One problem is that: `F.sdpa` is still one op after exporting, so it needs to somehow breakdown before applying pt2e.

jerryzh168: right, we'll need some rewrite util for that, and user should call that explicitly if they want to do that

jerryzh168:
will need to revert this one due to internal build errors, please land again

torchao/quantization/sfdp_int8_fx_pass.py
Valentine233: Report an issue: https://github.com/pytorch/pytorch/issues/151876

kimishpatel:
I also dont see clear description of what int8 sdpa really entails numerically. I implemented something quantized sdpa where q, k and v are per token quantized but output of sdpa is still fp32. I dont know if that is the intent here as well. But if we are landing this in AO then we should make sure such implementations are aligned. cc; @jerryzh168 @drisspg

leslie-fang-intel:
> I also dont see clear description of what int8 sdpa really entails numerically. I implemented something quantized sdpa where q, k and v are per token quantized but output of sdpa is still fp32. I dont know if that is the intent here as well. But if we are landing this in AO then we should make sure such implementations are aligned.

I think the output data type should be the same as input which should be `u8`. Hi @kimishpatel, could you share more details about your test case? so cc @Valentine233 can also help to take a look.

kimishpatel:
> > I also dont see clear description of what int8 sdpa really entails numerically. I implemented something quantized sdpa where q, k and v are per token quantized but output of sdpa is still fp32. I dont know if that is the intent here as well. But if we are landing this in AO then we should make sure such implementations are aligned.
> 
> I think the output data type should be the same as input which should be `u8`. Hi @kimishpatel, could you share more details about your test case? so cc @Valentine233 can also help to take a look.

I think we have treated quantized sdpa as dynamic quantized linear such that q, k and v are int8 but output is fp32. In larger part to avoid doing softmax in low precision which is where of the numerics issues of sdpa lie

kimishpatel:
In any case if you can update the description or add comment as to what the op's numerics are it would be great, e.g. q @ k is int8 @ int8 -> int8? what about softmax etc.

Valentine233:
> In any case if you can update the description or add comment as to what the op's numerics are it would be great, e.g. q @ k is int8 @ int8 -> int8? what about softmax etc.

Here is the workflow for quantized SDPA:
image

kimishpatel:
> > In any case if you can update the description or add comment as to what the op's numerics are it would be great, e.g. q @ k is int8 @ int8 -> int8? what about softmax etc.
> 
> Here is the workflow for quantized SDPA: image

Have you validated impact of this on any real models?

Valentine233:
> Have you validated impact of this on any real models?

Sure, for Bertlarge and VIT. You can see the results in the description part.

kimishpatel:
> > Have you validated impact of this on any real models?
> 
> Sure, for Bertlarge and VIT. You can see the results in the description part.

actually i was asking for accuracy measurements and more so for decoder only model as well.

However I see that reported speed up is in the range of 3-15%. is that right?

PR #1377, opened 2024-12-04
[float8] Re-enable slow-accum in the bwd of axis-wise scaling schemes

lw:
And circumvent the issue with the slow CUTLASS kernel by using the cuBLAS kernel + manual scaling.


lw:
Re-submission of https://github.com/pytorch/ao/pull/1325

vkuzo (approved):

PR #1382, opened 2024-12-05
Enable float8 attention support (q/k/v)

jerryzh168:
Summary:
This PR integrates flashattention 3 kernel: https://github.com/Dao-AILab/flash-attention/blob/1feb711f46563960fc10a8e659c93c300619504b/flash_attn/flash_attn_interface.py#L1102 to float8 affine quantized tensor.

To use the kernel, right now we need to manually add quantize call for q/k/v before sdpa op, but we can explore other APIs in the future

@sijiac is working on adding new variations of attention implementation in the future (per row, per column, per block scaling etc.).

Test Plan:
python test/dtypes/test_affine_quantized_float.py -k test_float8_attention

### SAM2
tested on sam2 and seems to be a bit slower than before, this is reasonable because sam2 is using 16 and 32 head dimension, but fa3 requires 64 being the minimum size, we need to do some padding to make this work (pad 32 to 64) which is expected to increase runtime significantly.

### llama2
llama2 without fallback: doesn't work because `attn_mask` is not supported.

### llama2 numerics only
(just for testing, code is not checked in) tested on llama2 (with fallback to test numerics):

since `attn_mask` is not supported in flashattention 3 kernel, it's using the fallback path: https://github.com/pytorch/ao/pull/1382/files#diff-3019e8f38b0919dbaba5aa1329a697e89fc98749e35a7bdc274c71a0d3738ec2R285
```
no quantize attn:

wikitext: {'alias': 'wikitext', 'word_perplexity,none': 12.2451228989592, 'word_perplexity_stderr,none': 'N/A', 'byte_perplexity,none': 1.5975503222785694, 'byte_perplexity_stderr,none': 'N/A', 'bits_per_byte,none': 0.675861376279183, 'bits_per_byte_stderr,none': 'N/A'}

quantize attn:

wikitext: {'alias': 'wikitext', 'word_perplexity,none': 12.294668124182962, 'word_perplexity_stderr,none': 'N/A', 'byte_perplexity,none': 1.5987571166545238, 'byte_perplexity_stderr,none': 'N/A', 'bits_per_byte,none': 0.6769507810861697, 'bits_per_byte_stderr,none': 'N/A'}
```
Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/dtypes/test_affine_quantized_float.py
+++ b/test/dtypes/test_affine_quantized_float.py
@@ -9,7 +9,6 @@ if not TORCH_VERSION_AT_LEAST_2_5:
 
 import copy
 import io
-import math
 import random
 import unittest
 from contextlib import nullcontext
@@ -18,7 +17,6 @@ from typing import Tuple
 
 import pytest
 import torch
-from einops import rearrange, repeat
 from torch._inductor.test_case import TestCase as InductorTestCase
 from torch.testing._internal import common_utils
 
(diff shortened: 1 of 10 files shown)


torchao/dtypes/floatx/float8_layout.py
```diff
@@ -311,3 +311,80 @@ def _linear_fp_act_fp8_weight_impl(
...
+    # for libc10.so
+    import torch
+    from hopper.flash_attn_interface import flash_attn_func
+
+    q_tensor_impl = q.tensor_impl
+    assert not q_tensor_impl.transposed
+    q_float8_data = q_tensor_impl.float8_data
+    # change from scalar to tensor of size [1]
+    q_scale = q_tensor_impl.scale
+    q_scale = torch.tensor([q_scale], device=q_scale.device)
```
drisspg, line 360: are the scales on host?

torchao/dtypes/affine_quantized_tensor_ops.py
```diff
@@ -277,6 +240,64 @@ def _(func, types, args, kwargs):
...
+        args[1].quant_max,
+        args[1].zero_point_domain,
+        output_dtype=sliced_scale.dtype,
+    )
+
+
+@implements(torch.nn.functional.scaled_dot_product_attention)
+def _(func, types, args, kwargs):
+    q, k, v = args[:3]
+    if not _sdpa_float8_check(q, k, v, kwargs):
```
drisspg, line 285: what if we just raise for now? since we only have support for fp8. I thikn doing the fallback is unintuitive here for most users. When we add the int8 kernel we can undo

torchao/dtypes/affine_quantized_tensor_ops.py
```diff
@@ -277,6 +240,64 @@ def _(func, types, args, kwargs):
         return func(input_tensor, weight_tensor)
 
 
+@implements(torch.nn.functional.embedding)
```
drisspg, line 243: Is this related?

test/dtypes/test_affine_quantized_float.py
```diff
@@ -288,7 +290,182 @@ def test_fp8_weight_dimension_warning(self):
             )
 
 
+# copied from https://github.com/Dao-AILab/flash-attention/blob/1feb711f46563960fc10a8e659c93c300619504b/tests/test_util.py#L185
+
+
+def construct_local_mask(
```
drisspg, line 296: I have variants of these that dont require einops
https://github.com/pytorch/pytorch/blob/f9af86de0138646540980148a781b7709f03c998/test/test_transformers.py#L2381

test/dtypes/test_affine_quantized_float.py
```diff
@@ -288,7 +290,182 @@ def test_fp8_weight_dimension_warning(self):
...
+        k = torch.randn([64, 8, 8, 64], dtype=torch.bfloat16, device="cuda")
+        v = torch.randn([64, 8, 8, 64], dtype=torch.bfloat16, device="cuda")
+
+        m = MyModel().eval()
+        # it differs a lot from the non-quantized implementation
+        # sqnr = -2.5
+        # ref = m(q, k, v)
+
+        # but matches the custom attention implementation in flash attention repo
+        ref = attention_ref(q, k, v)[0]
```
drisspg, line 462: can't we just compare to scaled_dot_product_attention

drisspg (commented):
Overall looks good, left a few comments. I think we can add the int8 kernel when it gets added as well for CPU

test/dtypes/test_affine_quantized_float.py
drisspg: I guess you say in your PR summmary that there is a big deviation? that doesn't seem good to me and we probalby need to pin point what is going on

torchao/dtypes/floatx/float8_layout.py
```diff
@@ -311,3 +311,80 @@ def _linear_fp_act_fp8_weight_impl(
...
+    v_scale = v_tensor_impl.scale
+    v_scale = torch.tensor([v_scale], device=v_scale.device)
+
+    dropout_p = kwargs.get("dropout_p", None)
+    assert (
+        dropout_p is None or dropout_p == 0.0
+    ), "dropout_p should be set to 0.0 during inference"
+    causal = kwargs.get("causal", False)
+
+    out = flash_attn_func(
```
drisspg, line 380: Flash_attn and SDPA have different semantics

FLash exepcts B, S, H, D
and SDPA expectes B, H, S, D requiring a  transposition

torchao/dtypes/floatx/float8_layout.py
jerryzh168: you mean `q_scale` before we call `torch.tensor`? they should be using the same device as original weight I think, so should be on cuda before

torchao/dtypes/affine_quantized_tensor_ops.py
jerryzh168: sure, just to confirm, you mean when we add int8 kernel, we want to add fallback back? or just add a new dispatch to int8 kernel but still raise for the other cases

test/dtypes/test_affine_quantized_float.py
jerryzh168: yes, it does not match `scaled_dot_product_attention`, yeah we need to figure this out in the end, but since @sijiac is developing new kernels, we can probably wait a bit

torchao/dtypes/floatx/float8_layout.py
jerryzh168: ah, thanks, maybe that's why they do not match..I can try tomorrow

torchao/dtypes/affine_quantized_tensor_ops.py
drisspg: I'm saying lets have this dispatch be: if float8 check passes calla flash float8, otherwise raise a nice error. We can probably have a config that will allow users to opt into the dequantize and call higher precision sdpa

torchao/quantization/README.md
```diff
@@ -360,6 +360,36 @@ We have kernels that do 8-bit dynamic quantization of activations and uintx grou
 
 You try can out these apis with the `quantize_` api as above alongside the constructor `int8_dynamic_activation_intx_weight`.  An example can be found in `torchao/_models/llama/generate.py`.
 
+### Float8 `scaled_dot_product_attention` Support
```
vkuzo, line 363: should this be in prototype?  The API doesn't match anything else we have in torchao, and in the PR summary you also mention that this isn't a final API.

torchao/quantization/README.md
```diff
@@ -360,6 +360,36 @@ We have kernels that do 8-bit dynamic quantization of activations and uintx grou
...
+from torchao.quantization.quant_api import _float8_symmetric_per_tensor_quant
+import torch.nn.functional as F
+
+class MyModel(torch.nn.Module):
+    def forward(self, q, k, v, float8_quantize=False):
+        if float8_quantize:
+            q = _float8_symmetric_per_tensor_quant(q)
+            k = _float8_symmetric_per_tensor_quant(k)
+            v = _float8_symmetric_per_tensor_quant(v)
+        return F.scaled_dot_product_attention(q, k, v)
```
vkuzo, line 380: If we already require a modeling code change, what's the benefit of overriding F.sdpa instead of just calling a quantized attention op directly?  I think with the API shown in this PR just calling a quantized op is way simpler.

torchao/quantization/README.md
jerryzh168: yeah this is prototype, this section is the prototype section, I can add a description on top

torchao/dtypes/affine_quantized_tensor_ops.py
```diff
@@ -277,6 +240,64 @@ def _(func, types, args, kwargs):
...
+        args[1].zero_point_domain,
+        output_dtype=sliced_scale.dtype,
+    )
+
+
+@implements(torch.nn.functional.scaled_dot_product_attention)
+def _(func, types, args, kwargs):
+    q, k, v = args[:3]
+    if not _sdpa_float8_check(q, k, v, kwargs):
+        # dequantize and call original op
```
vkuzo, line 286: I would expect this path to raise an exception instead of dequantizing and calling the high precision op.  If I'm asking an API to quantize an op, the two cases I expect is either (1) successfully quantize an op or (2) throw an exception with an error message.  I really don't expect or want (3) silently ignore my API call.

torchao/quantization/README.md
vkuzo: it's not clear that this is prototype, the section title is "Other Available Quantization Techniques".  It would be great to clarify the expections on APIs in this section!

torchao/dtypes/affine_quantized_tensor_ops.py
jerryzh168: yeah makes sense, updated

torchao/dtypes/floatx/float8_layout.py
jerryzh168: thanks @drisspg this helped close the gap, I can compare against bfloat16 SDPA directly now

test/dtypes/test_affine_quantized_float.py
jerryzh168: removed this since we are matching original floating point output now

test/dtypes/test_affine_quantized_float.py
jerryzh168: fixed now, somehow accidently had the wrong q/k/v order before

torchao/_models/llama/model.py
```diff
@@ -311,11 +313,25 @@ def forward(self, x: Tensor, freqs_cis: Tensor, mask: Optional[Tensor], input_po
...
         k = k.repeat_interleave(self.n_head // self.n_local_heads, dim=1)
         v = v.repeat_interleave(self.n_head // self.n_local_heads, dim=1)
+
+        # quantize q/k/v with per tensor float8 quantization
+        padded = False
+        if _QUANTIZE_ATTN:
+            from torchao.quantization.quant_api import _float8_symmetric_per_tensor_quant
+            original_dtype = v.dtype
+            if q.shape[-1] in [64, 128, 256]:
+                q = _float8_symmetric_per_tensor_quant(q)
```
drisspg, line 323: We also likely need/want to apply the hadamard transform. I don't remember off hand if this is include in fav3 api

torchao/_models/sam2/modeling/sam/transformer.py
bhack, line 1: Can we update the current SAM2 readme with all the ao optimizations we have introduced?

torchao/_models/llama/model.py
bhack: https://pytorch.org/blog/hadacore/

PR #1385, opened 2024-12-06
Move MarlinQQQTensor out of AQT

jainapurva:
Reorganization of AQT: Move MarlinQQQTensor to marlin_qqq_tensor file, with layouts and kernels.

### DIFF
--- a/torchao/dtypes/README.md
+++ b/torchao/dtypes/README.md
@@ -1,0 +1,19 @@
+# README
+
+## File Structure of the `dtypes` Folder
+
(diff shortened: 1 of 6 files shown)


jerryzh168 (approved):
adding a README.md in https://github.com/pytorch/ao/tree/main/torchao/dtypes that explains the file structure related to aqt would be helpful I think

jerryzh168:
also It's fine to not mark this as bc-breaking since MarlinQQQ is prototype feature

Issue #1264, opened 2024-11-12
ZeroPointDomain as an arguments


airMeng:
## Context
Current ZeroPointDomain is [bound to the layout](https://github.com/pytorch/ao/blob/2ba1a61fe1244560325b5051b5d3c10044553be0/torchao/quantization/quant_api.py#L607)
```python
        ...
        zero_point_domain = ZeroPointDomain.FLOAT

        # Sparse Marlin only supports symmetric quantization.
        # NOTE: If we start having lots of layouts that require different configurations,
        # we should consider moving this logic somewhere else.
        if isinstance(layout, MarlinSparseLayout):
            ...
            zero_point_domain = ZeroPointDomain.INT
```
Ideally, we should allow the data types of zero points to be specified as arguments. There are two main benefits:

1. Memory Efficiency: Integer zero points can significantly reduce memory footprint. For example, with a group_size=32, using int4 zero points instead of bf16 can save 0.375 bits per element.
2. Community Compatibility: Integer zero points have a well-established ecosystem of recipes and kernels. Making this option available in TorchAO would allow us to leverage these resources more effectively.

## Proposals
Add an optional argument to let users specify the data types of zero points:
```diff
diff --git a/torchao/quantization/quant_api.py b/torchao/quantization/quant_api.py
index 476cc229..08716dbc 100644
--- a/torchao/quantization/quant_api.py
+++ b/torchao/quantization/quant_api.py
@@ -568,7 +568,7 @@ def int8_dynamic_activation_int4_weight(


 def int4_weight_only(
-    group_size=128, layout=TensorCoreTiledLayout(inner_k_tiles=8), use_hqq=False
+    group_size=128, layout=TensorCoreTiledLayout(inner_k_tiles=8), use_hqq=False, default_fp_zp=True
 ):
     """
     Applies uint4 weight-only asymmetric per-group quantization to linear layers, using
@@ -604,7 +604,7 @@ def int4_weight_only(
         eps = 1e-6
         preserve_zero = False
         zero_point_dtype = torch.bfloat16
-        zero_point_domain = ZeroPointDomain.FLOAT
+        zero_point_domain = default_fp_zp ? ZeroPointDomain.FLOAT : ZeroPointDomain.INT

         # Sparse Marlin only supports symmetric quantization.
         # NOTE: If we start having lots of layouts that require different configurations,
```

An example usage
```python
from torchao.quantization.quant_api import (
    quantize_,
    int8_dynamic_activation_int8_weight,
[...]

PR #1401, opened 2024-12-11
Expose zero_point_domain as arguments

airMeng:
As discussed in https://github.com/pytorch/ao/issues/1264, expose zero_point_domain to the users.

```
from torchao.quantization.quant_primitives import ZeroPointDomain
    quantize_,
    int4_weight_only,
)
quantize_(m, int4_weight_only(zero_point_domain=...))
```

However, to run the model E2E with integer zero points, we need to separate scales and zero points tensors. The first PR to enable this is https://github.com/pytorch/pytorch/pull/137566

@jgong5 @mingfeima @leslie-fang-intel @liangan1

### DIFF
--- a/test/dtypes/test_affine_quantized.py
+++ b/test/dtypes/test_affine_quantized.py
@@ -16,7 +16,7 @@ from torchao.quantization import (
     int8_dynamic_activation_int8_weight,
     int8_weight_only,
 )
-from torchao.quantization.quant_primitives import MappingType
+from torchao.quantization.quant_primitives import MappingType, ZeroPointDomain
 from torchao.utils import (
     TORCH_VERSION_AT_LEAST_2_5,
     TORCH_VERSION_AT_LEAST_2_6,
@@ -24,7 +24,7 @@ from torchao.utils import (
 )
 
 
-def get_quantization_functions(do_sparse: bool, do_int4: bool, device: str = "cuda"):
+def get_quantization_functions(do_sparse: bool, do_int4: bool, device: str = "cuda", int4_zp_int: bool = False):
     base_functions = [
         int8_weight_only(),
         int8_dynamic_activation_int4_weight(),
@@ -36,8 +36,14 @@ def get_quantization_functions(do_sparse: bool, do_int4: bool, device: str = "cu
             base_functions.append(
                 int4_weight_only(group_size=32, layout=Int4CPULayout())
             )
+            if int4_zp_int:
+                base_functions.append(
+                    int4_weight_only(group_size=32, layout=Int4CPULayout(), zero_point_domain=ZeroPointDomain.INT)
+                )
         else:
             base_functions.append(int4_weight_only(group_size=32))
+            if int4_zp_int:
+                base_functions.append(int4_weight_only(group_size=32, zero_point_domain=ZeroPointDomain.INT))
 
     if do_sparse:
         base_functions.append(
@@ -71,7 +77,7 @@ class TestAffineQuantized(TestCase):
             self.assertEqual(aqt_shape, shape)
 
     @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
-    @common_utils.parametrize("apply_quant", get_quantization_functions(True, True))
+    @common_utils.parametrize("apply_quant", get_quantization_functions(True, True, "cuda", True))
     def test_weights_only(self, apply_quant):
         linear = torch.nn.Linear(128, 256, dtype=torch.bfloat16, device="cuda")
         ql = apply_quant(linear)
--- a/test/dtypes/test_affine_quantized.py
+++ b/test/dtypes/test_affine_quantized.py
@@ -24,7 +24,9 @@ from torchao.utils import (
 )
 
 
-def get_quantization_functions(do_sparse: bool, do_int4: bool, device: str = "cuda", int4_zp_int: bool = False):
+def get_quantization_functions(
+    do_sparse: bool, do_int4: bool, device: str = "cuda", int4_zp_int: bool = False
+):
     base_functions = [
         int8_weight_only(),
         int8_dynamic_activation_int4_weight(),
@@ -38,7 +40,11 @@ def get_quantization_functions(do_sparse: bool, do_int4: bool, device: str = "cu
             )
             if int4_zp_int:
                 base_functions.append(
-                    int4_weight_only(group_size=32, layout=Int4CPULayout(), zero_point_domain=ZeroPointDomain.INT)
+                    int4_weight_only(
+                        group_size=32,
+                        layout=Int4CPULayout(),
+                        zero_point_domain=ZeroPointDomain.INT,
+                    )
                 )
         else:
             base_functions.append(int4_weight_only(group_size=32))
@@ -75,7 +81,9 @@ class TestAffineQuantized(TestCase):
             self.assertEqual(aqt_shape, shape)
 
     @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
-    @common_utils.parametrize("apply_quant", get_quantization_functions(True, True, "cuda", True))
+    @common_utils.parametrize(
+        "apply_quant", get_quantization_functions(True, True, "cuda", True)
+    )
     def test_weights_only(self, apply_quant):
         linear = torch.nn.Linear(128, 256, dtype=torch.bfloat16, device="cuda")
         ql = apply_quant(linear)
--- a/test/dtypes/test_affine_quantized.py
+++ b/test/dtypes/test_affine_quantized.py
@@ -42,8 +42,6 @@ def get_quantization_functions(do_sparse: bool, do_int4: bool, device: str = "cu
                 )
         else:
             base_functions.append(int4_weight_only(group_size=32))
-            if int4_zp_int:
-                base_functions.append(int4_weight_only(group_size=32, zero_point_domain=ZeroPointDomain.INT))
 
     if do_sparse:
         base_functions.append(
--- a/test/quantization/test_quant_primitives.py
+++ b/test/quantization/test_quant_primitives.py
@@ -108,7 +108,6 @@ def _groupwise_affine_quantize_tensor_from_qparams(
     zeros = zeros.reshape(-1, 1)
     max_int = 2**n_bit - 1
     min_int = 0
-    # import pdb; pdb.set_trace()
     if zero_point_domain == ZeroPointDomain.FLOAT:
         min_val = zeros - scales * (2 ** (n_bit - 1))
         w_int4x8 = (
--- a/test/quantization/test_quant_primitives.py
+++ b/test/quantization/test_quant_primitives.py
@@ -57,7 +57,13 @@ def check_idempotent(self, fn, *args, **kwargs):
 
 
 # Legacy tinygemm ops
-def _get_groupwise_affine_qparams(w, n_bit=4, groupsize=128, dtype=torch.bfloat16, zero_point_domain=ZeroPointDomain.FLOAT):
+def _get_groupwise_affine_qparams(
+    w,
+    n_bit=4,
+    groupsize=128,
+    dtype=torch.bfloat16,
+    zero_point_domain=ZeroPointDomain.FLOAT,
+):
     if groupsize > w.shape[-1]:
         groupsize = w.shape[-1]
     assert groupsize > 1
@@ -73,7 +79,7 @@ def _get_groupwise_affine_qparams(w, n_bit=4, groupsize=128, dtype=torch.bfloat1
     quant_min = 0
     quant_max = max_int
     scales = (max_val - min_val).clamp(min=1e-6) / max_int
-    if (zero_point_domain == ZeroPointDomain.FLOAT):
+    if zero_point_domain == ZeroPointDomain.FLOAT:
         zeros = min_val + scales * (2 ** (n_bit - 1))
         zeros = zeros.to(dtype=dtype).reshape(w.shape[0], -1)
     else:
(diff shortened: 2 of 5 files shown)


test/dtypes/test_affine_quantized.py
```diff
@@ -36,8 +36,14 @@ def get_quantization_functions(do_sparse: bool, do_int4: bool, device: str = "cu
             base_functions.append(
                 int4_weight_only(group_size=32, layout=Int4CPULayout())
             )
+            if int4_zp_int:
+                base_functions.append(
+                    int4_weight_only(group_size=32, layout=Int4CPULayout(), zero_point_domain=ZeroPointDomain.INT)
```
jerryzh168, line 41: can you add a new section in https://github.com/pytorch/ao/blob/main/torchao/quantization/README.md#other-available-quantization-techniques to report the e2e model level numbers for int4wo cpu?

torchao/quantization/quant_api.py
```diff
@@ -630,7 +630,8 @@ def int8_dynamic_activation_int4_weight(
 
 
 def int4_weight_only(
-    group_size=128, layout=TensorCoreTiledLayout(inner_k_tiles=8), use_hqq=False
+    group_size=128, layout=TensorCoreTiledLayout(inner_k_tiles=8), use_hqq=False,
+    zero_point_domain=ZeroPointDomain.FLOAT
```
jerryzh168, line 634: do we need to expose this? if each layout is tied to certain settings, maybe we can do this for now: https://github.com/pytorch/ao/blob/039cef4ad546716aa04cd54c461feb173f7fe403/torchao/quantization/quant_api.py#L675

jerryzh168: we are also brainstorming to how to best structure these settings as well, we might be reducing some of the options here in the future

airMeng: I think layout is for weight, while zero_point_domain is for zero points. So it would be better to decouple it.
But I agree we can structure these things, better to align with the most common recipes while maintaining the capabilities to let users extend themselves.

test/dtypes/test_affine_quantized.py
airMeng: The PR focus on the exposure of the API. The first backend to really run the models still depends on https://github.com/pytorch/pytorch/pull/137566, in which we register ```_weight_int4pack_mm_with_scale_and_zeros``` and make zeros not the same dtype of scales available in PyTorch.  cc @liangan1 here for awareness

However, if you prefer the 2 changes (zero_point_domain as arguments, separate scales and zeros tensors for int4_mm) in 1 PR, I am fine too.

torchao/quantization/quant_api.py
jerryzh168: yeah I understand that they are separate. it's more of we are using layout as a tool to select kernels right now I think, e.g. the sparse layout example that I linked, user don't need to worry about making all the settings correct, just need to specify the sparse layout and the rest is set automatically. It is a bit confusing, but we can think about how to make it clearer in the future, one alternative I can think of is just expose a top level api for each kernel.

test/dtypes/test_affine_quantized.py
jerryzh168: oh I see, so this is for xpu, I'd suggest to add a new layout for xpu and configure everything through that layout for now, instead of exposing these options to user, something like this:
```
def int4_weight_only(...., layout):
        if isinstance(layout, Int4XPULayout):
            # set zero_point_domain and other settings here
            ...
```

does this work for your use case?

airMeng: I'd put it not only for XPU but for all backends. CPU and CUDA side might also want integer zero points, as what other popular recipes do. For XPU we are also working on floating zero points,

Do you prefer I binding the integer zp to XPU in this PR temporarily?

jerryzh168: oh OK, if you want to support both, then it makes sense to expose I think, just make sure to assert in places when it's not supported

torchao/quantization/quant_api.py
```diff
@@ -665,9 +666,8 @@ def apply_int4_weight_only_quant(weight):
         quant_min = 0
         quant_max = 15
         eps = 1e-6
-        preserve_zero = False
+        preserve_zero = zero_point_domain == ZeroPointDomain.INT
         zero_point_dtype = torch.bfloat16
-        zero_point_domain = ZeroPointDomain.FLOAT
 
         # Sparse Marlin only supports symmetric quantization.
         # NOTE: If we start having lots of layouts that require different configurations,
```
jerryzh168, line 673: might be better to add some assert for each of the layout for now, otherwise people can get invalid configurations, e.g. right now we have:
```
TensorCoreTiledLayout: zero point domain has to be FLOAT, and preserve_zero has to be False
MarlinSparseLayout: zero point has to be int, preserve zero has to be True
IntCPULayout: same as TensorCoreTiledLayout?
...
```

jerryzh168 (approved):
looks good overall, please add some assert in `int4_weight_only` to make it clearer what is supported.

airMeng:
@jerryzh168 assert added, could you give a review?

torchao/quantization/quant_api.py
```diff
@@ -665,9 +666,12 @@ def apply_int4_weight_only_quant(weight):
         quant_min = 0
         quant_max = 15
         eps = 1e-6
-        preserve_zero = False
+        preserve_zero = zero_point_domain == ZeroPointDomain.INT
         zero_point_dtype = torch.bfloat16
-        zero_point_domain = ZeroPointDomain.FLOAT
+
+        if isinstance(layout, TensorCoreTiledLayout):
```
jerryzh168, line 672: I think zero_point_domain should probably default to None, which means we'll set it based on layout, but when it is not None, we'd need to check it for TensorCoreTiledLayout, MarlinSparseLayout and IntCPULayout

jerryzh168 (changes requested):
requesting changes to the default value of zero_point_domain

torchao/quantization/quant_api.py
airMeng: fixed in https://github.com/pytorch/ao/pull/1401/commits/1dbf9aa7bf4b1c0d4aba686e598d4264ed03ac64

torchao/quantization/quant_api.py
```diff
@@ -665,9 +666,15 @@ def apply_int4_weight_only_quant(weight):
         quant_min = 0
         quant_max = 15
         eps = 1e-6
-        preserve_zero = False
+        preserve_zero = zero_point_domain == ZeroPointDomain.INT
```
jerryzh168, line 669: this is probably not longer valid, since we use zero_point_domain == None by default so user don't need to worry about it. to make this clearer I'd suggest the following:

```
assert layout in layout_to_default_zero_point.keys(), f"Only support layout: {layout_to_default_zero_point.keys()}"
if zero_point_domain is None:
    zero_point_domain = layout_to_default_zero_point[layout]
    preserve_zero = layout_to_default_preserve_zero[layout]
else:
    assert zero_point_domain  in layout_to_supported_zero_point_domain[layout], "...."
```

airMeng: Could you check https://github.com/pytorch/ao/pull/1401/commits/17b8bf1ac39ea741c431afbb088109b2123781cd?

torchao/quantization/quant_api.py
```diff
@@ -665,17 +679,21 @@ def apply_int4_weight_only_quant(weight):
...
         zero_point_dtype = torch.bfloat16
-        zero_point_domain = ZeroPointDomain.FLOAT
+
+        assert layout in layout_to_zero_point_domain.keys(), f"Only support layout: {layout_to_zero_point_domain.keys()}"
+        if zero_point_domain is None:
+            # the first value is the default one
+            zero_point_domain = layout_to_zero_point_domain[layout][0]
+            preserve_zero = layout_to_preserve_zeros[layout]
+        else:
+            assert zero_point_domain  in layout_to_zero_point_domain[layout], f"Layout only support {layout_to_zero_point_domain[layout]}"
```
jerryzh168, line 690: nit: extra space

jerryzh168 (approved):
looks good, thanks!

torchao/quantization/quant_api.py
```diff
@@ -110,6 +111,18 @@
     "Int8DynActInt4WeightGPTQQuantizer",
 ]
 
+# update according to the support matrix
+layout_to_zero_point_domain = {
```
jerryzh168, line 115: nit: capitalize, like https://github.com/pytorch/ao/blob/ebc43034e665bcda759cf9ef9c2c207057c5eeb1/torchao/quantization/quant_primitives.py#L100

test/quantization/test_quant_primitives.py
```diff
@@ -97,17 +106,29 @@ def _groupwise_affine_quantize_tensor_from_qparams(
...
     min_int = 0
-    w_int4x8 = (
-        to_quant.sub(min_val)
-        .div(scales)
-        .round()
-        .clamp_(min_int, max_int)
-        .to(torch.int32)
-        .reshape_as(w)
-    )
+    # import pdb; pdb.set_trace()
```
jerryzh168, line 111: nit: remove

airMeng:
@facebook-github-bot label "topic: improvement"

torchao/quantization/quant_api.py
```diff
@@ -665,17 +679,22 @@ def apply_int4_weight_only_quant(weight):
         quant_min = 0
         quant_max = 15
         eps = 1e-6
-        preserve_zero = False
         zero_point_dtype = torch.bfloat16
-        zero_point_domain = ZeroPointDomain.FLOAT
+
+        nonlocal zero_point_domain
+        assert any(isinstance(layout, support_layout) for support_layout in LAYOUT_TO_ZERO_POINT_DOMAIN.keys()), f"Only support layout: {LAYOUT_TO_ZERO_POINT_DOMAIN.keys()}"
```
jerryzh168, line 685: `type(layout)` should be enough?

airMeng: fix

torchao/quantization/quant_api.py
```diff
@@ -630,7 +643,8 @@ def int8_dynamic_activation_int4_weight(
 
 
 def int4_weight_only(
-    group_size=128, layout=TensorCoreTiledLayout(inner_k_tiles=8), use_hqq=False
+    group_size=128, layout=TensorCoreTiledLayout(inner_k_tiles=8), use_hqq=False,
+    zero_point_domain=None
 ):
     """
     Applies uint4 weight-only asymmetric per-group quantization to linear layers, using
```
jerryzh168, line 650: nit: please update docs for zero_point_domain before landing

airMeng: update in https://github.com/pytorch/ao/pull/1401/commits/51a4505800e43fd2093098180c401bbd7a38a62d

torchao/quantization/README.md
```diff
@@ -202,6 +202,17 @@ We also have a unified quantized tensor subclass that implements how to get a qu
 #### Layouts
 We extended the `layout` concept to represent different packing formats for a tensor. `AffineQuantizedTensor` supports `plain` and `tensor_core_tiled` layout. `plain` layout is used for `int8_weight_only` and `int8_dynamic_activation_int8_weight` and also as a default layout. `tensor_core_tiled` layout is used for `int4_weight_only` quantization and is packing the weights in a format that is compatible with tinygemm [int4mm](https://github.com/pytorch/pytorch/blob/39357ba06f48cda7d293a4995aa5eba2a46598b5/aten/src/ATen/native/native_functions.yaml#L4138) kernels.
 
+### Zero Point Domains
+```ZeroPointDomain``` is used to control the data types of zero points. ```None``` represents symmetric quantization, while ```ZeroPointDomain.FLOAT``` and ```ZeroPointDomain.INT``` indicate asymmetric quantization. For detailed implementation of different zero point data types, refer to [the reference implementation]((../../test/quantization/test_quant_primitives.py)).
```
jerryzh168, line 206: we have MappingType to mean symmetric or asymmetric quant, zero_point_domain is referring to zero_point, None means zero_point is None, FLOAT means zero_point is in the floating point domain and INT means integer domain, please see: https://github.com/pytorch/ao/blob/46b8796412eb350d1923091892850582d32737d0/torchao/quantization/quant_primitives.py#L68

torchao/quantization/quant_api.py
```diff
@@ -650,6 +664,7 @@ def int4_weight_only(
          size is more fine grained, choices are [256, 128, 64, 32]
         `layout`: layout type for quantized tensor, default is `TensorCoreTiledLayout(inner_k_tiles=8)`
         `use_hqq`: whether to use hqq or default quantization mode, default is False
+        `zero_point_domain`: data type of zeros points, choices are [None(default), ZeroPointDomain.FLOAT, ZeroPointDomain.INT, ZeroPointDomain.NONE]
```
jerryzh168, line 667: nit: probably adding that `None` means we'll set zero_point_domain based on layout

torchao/quantization/README.md
airMeng: https://github.com/pytorch/ao/pull/1401/commits/e4237e006d11624633555ac40283ee6dd8349f34

torchao/quantization/quant_api.py
airMeng: https://github.com/pytorch/ao/pull/1401/commits/e4237e006d11624633555ac40283ee6dd8349f34

airMeng:
@jerryzh168 anything more before merged?

jerryzh168:
No we can merge now

PR #1402, opened 2024-12-11
[WIP] Use `int_scaled_matmul` with `int8_dynamic_activation_int8_weight(act_mapping_type=MappingType.ASYMMETRIC)`

sanchitintel:
## Feature

Use `int_scaled_matmul` with asymmetrically int8 quantized activation & symmetrically int8 quantized weight by applying compensation for zero point of activation.

### Motivation
Currently, optimizing GEMMs by using asymmetrically int8 quantized activation & symmetrically quantized weight with the `int8_dynamic_activation_int8_weight` API poses a problem - torchao currently does not use `torch._int_mm` for this case, so in case of frozen weights (inference, and Inductor freezing config enabled), the frozen int8 weights are folded into FP32 weights (as `aten.mm`'s second argument) during Inductor's constant-folding passes (during freezing).

With `sym act, sym wgt` case, `torch._int_mm` is being used, and that makes it easier to leverage frozen int8 weights with Inductor pattern-matching & use Inductor max-autotune mode.

This PR does something similar for `asym act, sym wgt` case by using `torch._int_mm` with int8 activation & weights, and applying compensation corresponding to the activation's zero points.

This change makes it possible to leverage this GEMM's pattern with Inductor pattern-matching & using Inductor max-autotune to fuse the whole GEMM (I'd add its support in a PyTorch PR).

### TODO

 - [ ] Report if there's any eager mode slowdown with this approach
 - [ ] Create PyTorch PR & report inference speedup

### DIFF
--- a/torchao/dtypes/uintx/plain_layout.py
+++ b/torchao/dtypes/uintx/plain_layout.py
@@ -220,7 +220,7 @@ def _linear_fp_act_int8_weight_impl(input_tensor, weight_tensor, bias):
     return y
 
 
-def _linear_int8_act_int8_weight_check(input_tensor, weight_tensor, bias):
+def _linear_sym_int8_act_sym_int8_weight_check(input_tensor, weight_tensor, bias):
     return (
         isinstance(input_tensor, AffineQuantizedTensor)
         and _aqt_is_int8_reduced_range(input_tensor)
@@ -231,7 +231,7 @@ def _linear_int8_act_int8_weight_check(input_tensor, weight_tensor, bias):
     )
 
 
-def _linear_int8_act_int8_weight_impl(input_tensor, weight_tensor, bias):
+def _linear_sym_int8_act_sym_int8_weight_impl(input_tensor, weight_tensor, bias):
     #
     # 1. do the matrix form of dot(X_i, W_j)
     #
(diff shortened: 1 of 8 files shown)


torchao/dtypes/uintx/plain_layout.py
```diff
@@ -266,3 +266,58 @@ def _linear_int8_act_int8_weight_impl(input_tensor, weight_tensor, bias):
...
     return y
+
+
+def _linear_asym_int8_act_sym_int8_weight_check(input_tensor, weight_tensor, bias):
+    return (
+        isinstance(input_tensor, AffineQuantizedTensor)
+        and _aqt_is_int8(input_tensor)
+        # ZeroPointDomain.NONE did not work for weight in int8_dynamic_activation_int8_weight
+        # Uncommenting the next line works in eager mode, but Dynamo runs into a problem with it.
+        #and torch.equal(weight_tensor.tensor_impl.zero_point, torch.zeros_like(weight_tensor.tensor_impl.zero_point))
```
sanchitintel, line 277: This line is a blocker. Uncommenting it results in a Dynamo related error probably because the data of the weight tensor can't be accessed during Dynamo tracing. `ZeroPointDomain.NONE` did not work for weight in `int8_dynamic_activation_int8_weight`.

Keeping this line commented is incorrect, since this approach is only meant for symmetric activation of weights. So, need to find another way to ascertain here that the weights are symmetrically quantized.

jerryzh168: is this because of the specific way you are comparing the tensors? 

also what is the error for `ZeroPointDomain.NONE`?

sanchitintel: Sorry my late response, @jerryzh168!
Thanks for taking a look!


> is this because of the specific way you are comparing the tensors?

I'm using `torch.equal` to compare tensors. This error is reported in logs generated with `TORCH_LOGS="+dynamo"` -

```
  File "torch/utils/_stats.py", line 21, in wrapper
    return fn(*args, **kwargs)
  File "torch/_subclasses/fake_tensor.py", line 1276, in __torch_dispatch__
    return self.dispatch(func, types, args, kwargs)
  File "torch/_subclasses/fake_tensor.py", line 1817, in dispatch
    return self._cached_dispatch_impl(func, types, args, kwargs)
  File "torch/_subclasses/fake_tensor.py", line 1387, in _cached_dispatch_impl
    output = self._dispatch_impl(func, types, args, kwargs)
  File "torch/_subclasses/fake_tensor.py", line 2355, in _dispatch_impl
    op_impl_out = op_impl(self, func, *args, **kwargs)
  File "torch/_subclasses/fake_impls.py", line 546, in data_dep
    raise DataDependentOutputException(func)
torch._subclasses.fake_tensor.DataDependentOutputException: aten.equal.default
```

This runtime error is reported when the process exits -

```
torch._dynamo.exc.InternalTorchDynamoError: RuntimeError: Cannot access data pointer of Tensor (e.g. FakeTensor, FunctionalTensor). If you're using torch.compile/export/fx, it is likely that we are erroneously tracing into a custom kernel. To fix this, please wrap the custom kernel into an opaque custom op. Please see the following for details: https://pytorch.org/tutorials/advanced/custom_ops_landing_page.html
```

sanchitintel: > also what is the error for `ZeroPointDomain.NONE`?

Runtime error at https://github.com/pytorch/ao/blob/039cef4ad546716aa04cd54c461feb173f7fe403/torchao/quantization/quant_primitives.py#L895

jerryzh168: oh, this check seems wrong: https://github.com/pytorch/ao/blob/039cef4ad546716aa04cd54c461feb173f7fe403/torchao/quantization/quant_primitives.py#L891 it should be zero_point_domain != ZeroPointDomain.NONE

sanchitintel: Thanks again for your help, @jerryzh168!

It looks like conditions pertaining to both `None` and `ZeroPointDomain.NONE` are required here, as `None` is being exclusively used for FP8.

I changed the implementation to return `None` for `zero_point` if `zero_point_domain` is `ZeroPointDomain.NONE`.

It seems most code-coverage is tested through CUDA UTs, so if CUDA UTs would fail, I'll revise the implementation to return a zero point even when `zero_point_domain` would be `ZeroPointDomain.NONE`.

torchao/quantization/quant_primitives.py
```diff
@@ -888,14 +888,19 @@ def _choose_qparams_affine(
                 "preserve_zero == False is not supported for symmetric quantization"
             )
         if (
-            zero_point_domain is not None
+            zero_point_domain != ZeroPointDomain.NONE.name
+            and zero_point_domain != None
```
jerryzh168, line 892: I feel we can probably remove support for None since it's the same as ZeroPointDomain.NONE.name

torchao/quantization/quant_primitives.py
```diff
@@ -888,14 +888,19 @@ def _choose_qparams_affine(
...
             and zero_point_domain != ZeroPointDomain.INT.name
         ):
             raise ValueError(
-                "zero_point_domain != ZeroPointDomain.INT is not supported for symmetric quantization"
+                "Except a None value for zero_point_domain, Only ZeroPointDomain.NONE and ZeroPointDomain.INT"
+                " are supported for symmetric quantization."
             )
+        if zero_point_domain == ZeroPointDomain.NONE.name:
```
jerryzh168, line 899: thanks for the fix, looks like this is not tested before. can you add a test for the new code path?

also this op is becoming too complicated..we want to split

torchao/quantization/quant_primitives.py
sanchitintel: Thanks again for reviewing!

Some other places in the codebase are also using both `ZeroPointDomain.NONE.name` and None separately:

https://github.com/pytorch/ao/blob/31234dbc3210a0833cd922856107dc004d088ce1/torchao/quantization/quant_primitives.py#L543

I unified these two cases as one in the latest commit, but I'm not sure if changes in `__tensor_unflatten__` & `__tensor_flatten__` methods of some classes may be required at some other places in the codebase to ensure that they can deal with a `None` zero-point when TorchDynamo would be used . I'll run CUDA-only UTs at my end tomorrow morning to verify.

torchao/quantization/quant_primitives.py
sanchitintel: > can you add a test for the new code path?

This case is being tested in a UT I added in `test/quantization/test_quant_primitives.py`

> also this op is becoming too complicated..we want to split

Please advise if you're referring to splitting `_choose_qparams_affine`.
If so, I could split it up into smaller methods. Thanks!

jerryzh168: yeah I meant splitting choose_qparams_affine/quantize_affine/dequantize to different variations and reduce the complexity of the most common path, this includes removing preserve_zero, zero_point_domain args and just have different variations of choose_qparams_affine/quantize_affine/dequantize. this should be done separately though since it will be a large change

sanchitintel:
Closing in favor of #1556, which fixes the ZeroPointDomain.NONE implementation. Thanks!

PR #1404, opened 2024-12-12
Enable ciflow/benchmark

huydhn:
This label can then be added to PR to trigger benchmark jobs.  The label can only be used after this lands though, so it's not possible to test it out in this PR.

Alternatively, people can run the benchmark manually using https://github.com/pytorch/ao/actions/workflows/dashboard_perf_test.yml, which is useful if there are parameters to customize.


jerryzh168:
thanks, how do we verify?

jerryzh168 (approved):

PR #1415, opened 2024-12-13
Add new QAT API through quantize_

andrewor14:
**Summary:** This commit adds a new QAT API that can be used with the existing `quantize_`. This is an alternative to the old QAT *Quantizer APIs, which are much less flexible. The new API can be used as follows:

```
from torchao import quantize_
from torchao.quantization.qat import (
    FakeQuantizeConfig,
    intx_quantization_aware_training,
)
my_model = ...
activation_config = FakeQuantizeConfig(torch.int8, "per_token", is_symmetric=False)
weight_config = FakeQuantizeConfig(torch.int4, group_size=32)
quantize_(
    my_model,
    intx_quantization_aware_training(activation_config, weight_config),
)
```

**Test Plan:**
python test/quantization/test_qat.py -k test_quantize_api


torchao/quantization/qat/embedding.py
```diff
@@ -85,6 +82,27 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+            mod.embedding_dim,
+            mod.padding_idx,
+            mod.max_norm,
+            mod.norm_type,
+            mod.scale_grad_by_freq,
+            mod.sparse,
+            weight_config=weight_config,
+            device=mod.weight.device,
+        )
+        if mod.weight.device != torch.device("meta"):
```
jerryzh168, line 102: qq: why is this check needed?

torchao/quantization/qat/api.py
```diff
@@ -239,6 +239,52 @@ def __setattr__(self, name: str, value: Any):
...
+        if isinstance(mod, torch.nn.Linear):
+            return FakeQuantizedLinear.from_linear(
+                mod,
+                activation_config,
+                weight_config,
+            )
+        elif isinstance(mod, torch.nn.Embedding):
+            if activation_config is not None:
+                raise ValueError("Embedding does not support QAT for activations")
+            return FakeQuantizedEmbedding.from_embedding(mod, weight_config)
```
jerryzh168, line 279: the function takes both `activation_config` and `weight_config`, but in this case we only use `weight_config`

should we be separating these two branches to 2 APIs?

torchao/quantization/qat/embedding.py
andrewor14: added a comment, basically this would error in the distributed setting since the model is instantiated on the meta device for non-zero rank

torchao/quantization/qat/api.py
andrewor14: I feel it's better to only have one API, otherwise we'll need a separate API for each type of layer we support. Also it's more consistent with PTQ where we only have one `int8_weight_only` and let the `filter_fn` decide which layers to apply the transformation on (instead of having separate `int8_weight_only_linear` and `int8_weight_only_embedding`)

jerryzh168: oh the separation I'm referring to is to separate int8_act_int4_weight and int8_weight_only here, not linear v.s. embedding

andrewor14: oh I see, you mean separate this into the following?
```
intx_activation_intx_weight_quantization_aware_training
intx_weight_only_quantization_aware_training
```
I can do that, but the naming seems a bit long... any suggestions?

andrewor14: alternatively:
```
intx_quantization_aware_training (for both act + weight)
intx_weight_only_quantization_aware_training
```

andrewor14: Ok, separated into these two for now. Please let me know if you have better suggestions for the naming:
```
intx_quantization_aware_training (for both act + weight)
intx_weight_only_quantization_aware_training
```

jerryzh168: maybe intx_quantization_aware_training can be `intx_dynamic_quantization_aware_training`?

jerryzh168: or just expand to match ptq and then add a `qat` as suffix, that might be clearer?

jerryzh168 (approved):

andrewor14:
Thanks, merging this!

PR #1419, opened 2024-12-16
Move Int4CPULayout to int4_cpu_layout.py

yanbing-j:
This PR is to move Int4CPULayout to int4_cpu_layout.py, which is required in https://github.com/pytorch/ao/pull/1278#discussion_r1884352308.


jerryzh168 (approved):
thanks!

PR #1422, opened 2024-12-16
metal lowbit kernels: optimized 2-bit, 3-bit and 4-bit shaders

manuelcandales:
Adapts optimized 4-bit shader from PyTorch (MLX inspired) and adds similarly optimized 2-bit shader.
Adds optimized 3-bit shader
Restricts N to be multiple of 4 and adjusts tests accordingly.

Performance (tokens/sec via torchchat):

Llama 3.2 1B (llama3.2-1b-base): 
1bit: 28.0688
2bit: 31.2422
3bit: 30.1294
4bit: 30.7905
5bit: 28.1504
6bit: 28.4321
7bit: 27.3991

Llama 3.1 8B (llama3.1-base):
1bit: 7.4459
2bit: 15.6508
3bit: 15.3086
4bit: 16.1268
5bit: 6.7308
6bit: 6.4887
7bit: 6.4537

Notice that the performance of all n-bit kernels is similar for the 1B-parameter model. This optimization is felt when running the 8B-parameter model. Notice the jump from 6-7 tok/sec to 15-16 tok/sec when comparing the non-optimized kernels to the optimized 2-bit, 3-bit and 4-bit.


torchao/experimental/kernels/mps/metal/int3mm_opt.metal
```diff
@@ -0,0 +1,146 @@
...
+    uint k_block_index = k / group_size;
+    uint scales_group_offset = (k_block_index * N + n);
+
+    vecT scales =
+        (reinterpret_cast<constant vecT *>(scales_ptr + scales_group_offset))[0];
+    vecT zeros =
+        (reinterpret_cast<constant vecT *>(zeros_ptr + scales_group_offset))[0];
+    float4 zeros_float = float4(zeros);
+
+    float4 a_valA = float4(A_ptr[k / 4]);
```
kimishpatel, line 70: Also just use float4 a_val[2];?

torchao/experimental/kernels/mps/metal/int4mm_opt.metal
```diff
@@ -0,0 +1,180 @@
...
+    ushort b_val3 = (reinterpret_cast<constant ushort *>(
+        B_ptr + (k + 3 * K) / k_pack_factor))[0];
+    b_mat[0] = scales[0] * float4(float(b_val0 & 0x000f), float(b_val0 & 0x00f0),
+                               float(b_val0 & 0x0f00), float(b_val0 & 0xf000));
+    b_mat[1] = scales[1] * float4(float(b_val1 & 0x000f), float(b_val1 & 0x00f0),
+                               float(b_val1 & 0x0f00), float(b_val1 & 0xf000));
+    b_mat[2] = scales[2] * float4(float(b_val2 & 0x000f), float(b_val2 & 0x00f0),
+                               float(b_val2 & 0x0f00), float(b_val2 & 0xf000));
+    b_mat[3] = scales[3] * float4(float(b_val3 & 0x000f), float(b_val3 & 0x00f0),
+                               float(b_val3 & 0x0f00), float(b_val3 & 0xf000));
```
kimishpatel, line 141: we should just do similar unpack thing you had before but should validate whether that has any perf implication

torchao/experimental/kernels/mps/metal/int3mm_opt.metal
```diff
@@ -0,0 +1,146 @@
+#include <metal_simdgroup>
+#include <metal_stdlib>
+using namespace metal;
+
+inline void unpack_3bit(const uchar3 b, thread float* w) {
+  w[0] = float(((b[0] & 1) << 2) | (b[1] & 3));
```
kimishpatel, line 6: I dont know which of these shifts are avoidable and can be accounted for via act_div_scale stuff done in 4-bit kernel

torchao/experimental/kernels/mps/src/dispatch.h
```diff
@@ -20,4 +20,18 @@ inline void dispatch_mm(
       threadsPerThreadgroup:MTLSizeMake(std::min(maxThreadsPerGroup, M), 1, 1)];
 }
 
+inline void dispatch_mm_N4x32(
```
kimishpatel, line 23: Lets adopt similar naming as other libs like xnnpack. Mr1xNr4_per_TG (threadgroup)

torchao/experimental/kernels/mps/src/lowbit.h
```diff
@@ -125,6 +125,7 @@ void linear_lowbit_quant_weights_mps(
     int32_t N,
     const std::string_view type_str) {
   assert(K % 8 == 0);
```
kimishpatel, line 127: I dont know if this is a sufficient condition for optimized operators. Because when k is not multiple of what thread group handles per iteration, I dont know what we do. Or we probably do something wrong? Or maybe not because each thread in the threadgroup handles at most 8 k values per iteration.

torchao/experimental/kernels/mps/metal/int4mm_opt.metal
```diff
@@ -0,0 +1,180 @@
...
+  result += simd_shuffle_down(result, 4);
+  result += simd_shuffle_down(result, 8);
+  result += simd_shuffle_down(result, 16);
+  if (tid_in_simdgroup % threads_per_channel == 0) {
+    reinterpret_cast<device vecT *>(output_data + m * N)[n / 4] = vecT(result);
+  }
+}
+
+#define INSTANTIATE_INT4MM(DTYPE, GSIZE)                                       \
+  template [[host_name("int4pack_mm_" #GSIZE "_" #DTYPE)]] kernel void         \
```
kimishpatel, line 157: optimized and non optimized have the same host name?

torchao/experimental/kernels/mps/test/test_lowbit.mm
```diff
@@ -200,22 +200,23 @@ void run_test(int32_t m, int32_t k, int32_t n, int32_t group_size) {
...
-  run_test<T, nbit>(3, 64, 5, 64);
-  run_test<T, nbit>(7, 64, 23, 64);
-  run_test<T, nbit>(17, 120, 23, 128);
-  run_test<T, nbit>(17, 128, 23, 128);
-  run_test<T, nbit>(41, 144, 23, 128);
-  run_test<T, nbit>(41, 128, 23, 128);
-  run_test<T, nbit>(81, 8, 1, 256);
-  run_test<T, nbit>(19, 256, 17, 256);
-  run_test<T, nbit>(1, 1000, 81, 256);
+  run_test<T, nbit>(1, 8, 4, 32);
```
kimishpatel, line 203: Not sure if this path is actually testing optimized kernels at all

kimishpatel (changes requested):
When adding optimized version of something we should have some sort of benchmarking numbers. Ideally I would like that to come from standalone benchmark but fo rnow you can report what you got from torchchat

torchao/experimental/kernels/mps/metal.yaml
```diff
@@ -1,14 +1,17 @@
+- func: Vec4Type
+  file: common.metal
+
 - func: int1mm
-  file: divbit.metal
+  file: int1mm.metal
 
 - func: int2mm
-  file: divbit.metal
+  file: int2mm_opt.metal
```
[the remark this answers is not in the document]
manuelcandales: From the very beginning, the intent behind the structure of my code, has been to enable multiple versions of the shaders to coexist in the repo, to facilitate comparisons. Only one is "plugged in" at a given moment. That's the purpose of the `metal.yaml` and the `LowBitConfig` in `lowbit.h`.
So, yes, I am keeping the original int3mm.metal.
I think this should be fine for as long as we are inside an experimental folder, no?

torchao/experimental/kernels/mps/metal/int1mm.metal
```diff
@@ -0,0 +1,97 @@
+#include <metal_stdlib>
+using namespace metal;
+
+/**
+ * 1-Bit Quantized Linear.
+ *
+ * @param[A] M x K unquantized input tensor of floating point dtype (Float, Half, BFloat16)
```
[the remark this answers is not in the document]
manuelcandales: Ok

torchao/experimental/kernels/mps/metal/int1mm.metal
```diff
@@ -0,0 +1,97 @@
...
+        rc += a_val4 * (scale * float(w_val4) + zero);
+        rc += a_val5 * (scale * float(w_val5) + zero);
+        rc += a_val6 * (scale * float(w_val6) + zero);
+        rc += a_val7 * (scale * float(w_val7) + zero);
+      }
+    }
+    outputData[m * N + n] = T(rc);
+}
+
+#define INSTANTIATE_INT1MM(DTYPE, GSIZE)                                 \
```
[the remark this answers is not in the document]
manuelcandales: Sounds good.

torchao/experimental/kernels/mps/metal/int2mm_opt.metal
```diff
@@ -0,0 +1,138 @@
...
+  constant uchar *B_ptr = B + ((n * K) / k_pack_factor);
+
+  thread float4 result = float4(0.0);
+  // We multipy group of 4 channels with these scales.
+  // Because corresponding values from weight matrix are effectively left
+  // shifted. This is to avoid doing right shift on those values which ends up
+  // affecting performance. This is the trick applied in MLX kernels.
+  float4 act_div_scales = {1.f, 1 / 4.f, 1 / 16.f, 1 / 64.f};
+
+  for (; k < K; k += k_jump) {
```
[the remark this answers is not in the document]
manuelcandales: Why? See my answer to your comment below about K % 8 not being sufficient condition.

torchao/experimental/kernels/mps/metal/int3mm_opt.metal
manuelcandales: Ok.

torchao/experimental/kernels/mps/metal/int4mm_opt.metal
manuelcandales: Exactly, future PR :)

torchao/experimental/kernels/mps/metal/int3mm_opt.metal
manuelcandales: In future PRs I will experiment with different packing schemes, to see which one is optimal. This is the next thing I will do with 3-bit. But I don't want to do that in this PR.
Now, this packing isn't friendly with the act_div_scale stuff. This is because this packing spreads each weight value to multiple bytes. It isn't as clean as 4-bit where every value was contained in a single byte.
I am not saying that we won't use the act_div_scale trick in a future PR. But if you accept that we are not modifying the packing in this PR, then you should be ok with us not trying to complicate this PR with the act_div_scale trick either.

torchao/experimental/kernels/mps/src/dispatch.h
manuelcandales: Sure, sounds good.

torchao/experimental/kernels/mps/metal/int4mm_opt.metal
manuelcandales: See my answer to your first comment.
Yes, int4mm.metal is committed to the repo, but it isn't plugged in, int4mm_opt.metal is the one plugged in.

torchao/experimental/kernels/mps/metal/int2mm_opt.metal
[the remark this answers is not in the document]
kimishpatel: yeah sorry. it should work. I realized that in my latter comment

torchao/experimental/kernels/mps/metal/int3mm_opt.metal
kimishpatel: I am ok with not complicating this PR. I agree that how values are packed has impact on what you can do.

torchao/experimental/kernels/mps/src/lowbit.h
kimishpatel: yeah makes sense

torchao/experimental/kernels/mps/test/test_lowbit.mm
kimishpatel: Ok so then my concern is that at any point in time you are only testing the kernel that the codegen is using via yaml struct. This is bit problematic to me. I dont know if I should hold this PR off to fix that but we definitely should fix it

manuelcandales: If you prefer, I can delete the shaders not used, until we have a mechanism to test of all of the shaders in the metal directory. Does that sound fair?

kimishpatel (approved):
i trust the ones that are resolved as will be fixed, will get fixed. rest looks ok

torchao/experimental/kernels/mps/metal/int1mm.metal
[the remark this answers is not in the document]
manuelcandales: Done!

PR #1432, opened 2024-12-18
Gemlite fixes

HDCharles:
Summary:

shapes need to be divisible by 128 or they will not work with gemlite need fp32 accumulation for groupsize None on int4

Test Plan:

python test_integration.py -k "test_gemlite" (new test for non divisible shape)a

python generate.py --checkpoint_path $CHECKPOINT_PATH/$MODEL_REPO/model.pth --compile --precision float16 --quantization gemlite-8-4-None  --write_result benchmark_results.txt python generate.py --checkpoint_path
$CHECKPOINT_PATH/$MODEL_REPO/model.pth --compile --precision float16 --quantization gemlite-32-4-None  --write_result benchmark_results.txta

(previously these gave nonsense responses)

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/_models/llama/generate.py
+++ b/torchao/_models/llama/generate.py
@@ -3,6 +3,7 @@
 
 # This source code is licensed under the license found in the
 # LICENSE file in the root directory of this source tree.
+import json
 import os
 import platform
 import sys
@@ -18,11 +19,6 @@ import torch._inductor.config
 import torchao
 from torchao.quantization.quant_primitives import MappingType
 from torchao.utils import get_model_size_in_bytes, TORCH_VERSION_AT_LEAST_2_5
-from torchao._models.utils import (
-    get_arch_name,
-    write_json_result_ossci,
-    write_json_result_local,
-)
 
 torch.sparse.SparseSemiStructuredTensor._FORCE_CUTLASS = False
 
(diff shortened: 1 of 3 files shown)


torchao/_models/llama/generate.py
```diff
@@ -41,6 +37,14 @@ def elapsed_time(self, other_event):
         return abs(other_event.event_time - self.event_time) * 1000
 
 
+def get_arch_name() -> str:
```
jerryzh168, line 40: why these changes? is this some rebase issue

torchao/dtypes/uintx/gemlite_layout.py
```diff
@@ -173,6 +182,10 @@ def from_plain(
             exhaustive=False,
             use_cuda_graph=False,
         )
+        if _layout.group_size == None and _layout.bit_width == 4:
+                from gemlite.core import GEMLITE_ACC_DTYPE
+                from gemlite.dtypes import DType
+                GEMLITE_ACC_DTYPE[DType.FP16] = DType.FP32
```
mobicham, line 188: This will only work when all the layers use the same group_size, which is ok for now. 
The other option will be using this https://github.com/mobiusml/gemlite/blob/master/gemlite/core.py#L87 but for now let's keep it like this

jerryzh168:
landed in https://github.com/pytorch/ao/pull/1435, please feel free to submit any follow up fixes

torchao/dtypes/uintx/gemlite_layout.py
HDCharles: I tested this manually, it works in all cases even when there are different group sizes.

mobicham: I mean when different layers use different settings within the same model, but let's not worry about that !

PR #1434, opened 2024-12-18
[float8nocompile] Add unit tests validating model outputs, weight gradients, input gradients match expected values

danielvegamyhre:
**Summary**

The purpose of this change is to add a unit test validating weights and gradients are equal for the production float8 conversion path and the new float8nocompile path.

Note: this is different than the example in `examples/example.py` which is just a simple runnable e2e example documenting the UX.

**Test plan**:

Test is passing.

```
[danvm@devgpu006.vll6 ~/local/ao/torchao/prototype/float8nocompile/test (test)]$ pytest test.py
========================================================= test session starts =========================================================
platform linux -- Python 3.13.1, pytest-7.4.0, pluggy-1.5.0
rootdir: /data/users/danvm/ao
plugins: hypothesis-6.122.3
collected 1 item                                                                                                                      

test.py .                                                                                                                       [100%]

========================================================== warnings summary ===========================================================
test.py:15
  /data/users/danvm/ao/torchao/prototype/float8nocompile/test/test.py:15: PytestCollectionWarning: cannot collect test class 'TestModel' because it has a __init__ constructor (from: torchao/prototype/float8nocompile/test/test.py)
    class TestModel(nn.Module):

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
==================================================== 1 passed, 1 warning in 2.52s =====================================================
[danvm@devgpu006.vll6 ~/local/ao/torchao/prototype/float8nocompile/test (test)]$ 
```

### DIFF
--- a/torchao/prototype/float8nocompile/test/test.py
+++ b/torchao/prototype/float8nocompile/test/test.py
@@ -3,11 +3,9 @@ import torch
 import torch.nn as nn
 
 from torchao.float8.float8_linear_utils import convert_to_float8_training
-
 from torchao.prototype.float8nocompile.float8nocompile_linear_utils import (
     convert_to_float8_nocompile_training,
 )
-
 from torchao.utils import TORCH_VERSION_AT_LEAST_2_5
 
 if not TORCH_VERSION_AT_LEAST_2_5:
@@ -46,14 +44,19 @@ def test_model_weights_and_gradients(model1, model2):
     model2 = model2.to(torch.bfloat16).to(device)
 
     # compare production float8 linear conversion with no-compile version
-    convert_to_float8_nocompile_training(model1)
     convert_to_float8_training(model2)
+    convert_to_float8_nocompile_training(model1)
 
-    input_data = torch.randn(16, 32, dtype=torch.bfloat16).to(device)
+    input_tensor = torch.randn(
+        16, 32, requires_grad=True, dtype=torch.bfloat16, device=device
+    )
+    input_copy1 = input_tensor.clone().detach().requires_grad_(True)
+    input_copy2 = input_tensor.clone().detach().requires_grad_(True)
+
     loss_fn = nn.MSELoss()
 
-    output1 = model1(input_data)
-    output2 = model2(input_data)
+    output1 = model1(input_copy1)
+    output2 = model2(input_copy2)
 
     loss1 = loss_fn(output1, torch.zeros_like(output1))
     loss2 = loss_fn(output2, torch.zeros_like(output2))
@@ -61,7 +64,8 @@ def test_model_weights_and_gradients(model1, model2):
     loss1.backward()
     loss2.backward()
 
-    # compare the weights and gradients of both models
+    # compare the outputs, weight gradients, and input gradients
+    assert torch.allclose(output1, output2, atol=0, rtol=0)
+    assert torch.allclose(input_copy1.grad, input_copy2.grad, atol=0, rtol=0)
     for param1, param2 in zip(model1.parameters(), model2.parameters()):
-        assert torch.allclose(param1.data, param2.data, atol=0, rtol=0)
         assert torch.allclose(param1.grad, param2.grad, atol=0, rtol=0)
--- a/torchao/prototype/float8nocompile/test/test.py
+++ b/torchao/prototype/float8nocompile/test/test.py
@@ -1,0 +1,74 @@
+import pytest
+import torch
+import torch.nn as nn
+
+from torchao.float8.float8_linear_utils import convert_to_float8_training
+from torchao.prototype.float8nocompile.float8nocompile_linear_utils import (
+    convert_to_float8_nocompile_training,
+)
+from torchao.utils import TORCH_VERSION_AT_LEAST_2_5
+
+if not TORCH_VERSION_AT_LEAST_2_5:
+    raise AssertionError("torchao.float8 requires PyTorch version 2.5 or greater")
+
+
+class TestModel(nn.Module):
+    def __init__(self):
+        super().__init__()
+        self.layers = nn.Sequential(
+            nn.Linear(32, 64, bias=False),
+            nn.Linear(64, 32, bias=False),
+        )
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        return self.layers(x)
+
+
+@pytest.fixture
+def model1():
+    torch.manual_seed(0)
+    return TestModel()
+
+
+@pytest.fixture
+def model2():
+    torch.manual_seed(0)
+    return TestModel()
+
+
+def test_model_weights_and_gradients(model1, model2):
+    assert torch.cuda.is_available()
+    device = torch.device("cuda")
+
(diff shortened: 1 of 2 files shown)


danielvegamyhre:
cc @vkuzo for review

torchao/prototype/float8nocompile/test/test.py
```diff
@@ -0,0 +1,74 @@
...
+
+        loss1.backward()
+        loss2.backward()
+
+        optim1.step()
+        optim2.step()
+
+        # compare the weights and gradients of both models
+        for param1, param2 in zip(model1.parameters(), model2.parameters()):
+            assert torch.allclose(param1.data, param2.data, atol=1e-3, rtol=1e-3)
```
vkuzo, line 73: I think it should match with atol=0 and rtol=0, does that work?

You can also construct a simpler test to test the same thing with atol=0 and rtol=0:
1. start with sample input
2. do one forward and one backward (no optimizer update)
3. compare output, weight gradient, input gradient

If there is a divergence, the setup above will be easier to debug compared to a full training loop

danielvegamyhre: Yeah I used 1e-3 because when I tested with 1e-5 there was divergence. I will look into it further.

danielvegamyhre: As I test I used the official `convert_to_float8_training` implementation rather than my own, and there is still divergence, specifically on the weight gradients

danielvegamyhre: Actually I'm wondering if a slight amount of divergence in model outputs (and thus gradients) is expected when converting from bf16 to fp8, since we are using a lower precision data type there will be some amount of rounding error during the conversion, right?

danielvegamyhre: Ah, was your comment comment was specifically talking about requiring an exact match for `param.data`, not `param.gradient` as well? If so then this makes sense and I've updated the code accordingly.

I also removed the optimizers since they aren't needed to verify the weights and gradients.

vkuzo: if your first model is converted to float8 with the production path with default settings and your second model is converted to float8 with your path, then you should hopefully be able to match bitwise between them.  There isn't a strong need to compare to a bfloat16 baseline here.

danielvegamyhre: I see, I misunderstood what we wanted to compare here. That makes sense, I've updated the code accordingly to compare production float8 linear conversion with no-compile version and verify the param data and grads are an exact match.

torchao/prototype/float8nocompile/test/test.py
```diff
@@ -0,0 +1,67 @@
...
+    device = torch.device("cuda")
+
+    model1 = model1.to(torch.bfloat16).to(device)
+    model2 = model2.to(torch.bfloat16).to(device)
+
+    # compare production float8 linear conversion with no-compile version
+    convert_to_float8_nocompile_training(model1)
+    convert_to_float8_training(model2)
+
+    input_data = torch.randn(16, 32, dtype=torch.bfloat16).to(device)
```
vkuzo, line 52: I'd recommend the following:

```
input_data_1 = torch.randn(...).requires_grad_()
input_data_2 = copy.deepcopy(input_data_1)
# now you can test gradients on the above for equality
```

torchao/prototype/float8nocompile/test/test.py
```diff
@@ -0,0 +1,67 @@
...
+    output1 = model1(input_data)
+    output2 = model2(input_data)
+
+    loss1 = loss_fn(output1, torch.zeros_like(output1))
+    loss2 = loss_fn(output2, torch.zeros_like(output2))
+
+    loss1.backward()
+    loss2.backward()
+
+    # compare the weights and gradients of both models
```
vkuzo, line 64: the weights should match since they matched at model creation and there is no optimizer step, so no need to test that

here are three things to test:
* output1 vs output2
* weight1.grad vs weight2.grad
* input1.grad vs input2.grad

danielvegamyhre: Ok done. I had to play around with how I cloned the input tensor to ensure the copies had both `requires_grad==True` and `is_leaf==True` so the gradients would not be `None`. The tests are passing now.

danielvegamyhre: For my own learning, is the reason we care about input tensor gradients because certain techniques like saliency maps rely on using the input gradients? I didn't consider this at first and was only considering the case of updating model weights, so I am just curious.

torchao/prototype/float8nocompile/test/test.py
```diff
@@ -0,0 +1,73 @@
...
+def test_model_weights_and_gradients(model1, model2):
+    assert torch.cuda.is_available()
+    device = torch.device("cuda")
+
+    model1 = model1.to(torch.bfloat16).to(device)
+    model2 = model2.to(torch.bfloat16).to(device)
+
+    # compare production float8 linear conversion with no-compile version
+    convert_to_float8_nocompile_training(model1)
+    convert_to_float8_training(model2)
```
vkuzo, line 52: readability nit: order the baseline first and the experiment second, IMO that's easier to read for the average code reader.  Right now it's the other way around.

torchao/prototype/float8nocompile/test/test.py
```diff
@@ -0,0 +1,73 @@
...
+
+    input_tensor = torch.randn(16, 32, requires_grad=True, dtype=torch.bfloat16)
```
vkuzo, line 54: you can create on cuda directly with `torch.randn(..., device=device)`

torchao/prototype/float8nocompile/test/test.py
```diff
@@ -0,0 +1,73 @@
...
+
+    loss1 = loss_fn(output1, torch.zeros_like(output1))
+    loss2 = loss_fn(output2, torch.zeros_like(output2))
+
+    loss1.backward()
+    loss2.backward()
+
+    # compare the outputs, weight gradients, and input gradients
+    for param1, param2 in zip(model1.parameters(), model2.parameters()):
+        assert torch.allclose(output1, output2, atol=0, rtol=0)
```
vkuzo, line 71: nit: move outside the loop over params, otherwise you're comparing the same thing N times

torchao/prototype/float8nocompile/test/test.py
```diff
@@ -0,0 +1,73 @@
...
+        assert torch.allclose(param1.grad, param2.grad, atol=0, rtol=0)
+        assert torch.allclose(input_copy1.grad, input_copy2.grad, atol=0, rtol=0)
```
vkuzo, line 73: nit: move outside the loop over params, otherwise you're comparing the same thing N times

vkuzo (approved):

torchao/prototype/float8nocompile/test/test.py
danielvegamyhre: ah that was silly of me... thanks

PR #1445, opened 2024-12-19
[float8nocompile]: Triton kernels for conversion to float8 dtypes for forward pass of Float8LinearNoCompile

danielvegamyhre:
**Summary**

These changes include:

- Triton kernels for high-precision to fp8 conversion for the forward pass
- Wrapper function which orchestrates the conversion kernels and returns the result as Float8Tensor
- Unit test to compare production conversion path with this new path

**Test plan:**
- Unit tests for these kernels are passing

**Next steps**
- Wrap using torch.autograd.Function subclass w/ backward() passing through gradients unchanged, to make conversion to Float8Tensor differentiable 
- Use `torch.mm` (which wraps the cuBLAS float8 matmul routine) w/  Float8Tensors to perform the matmul needed for the linear layer

### DIFF
--- a/torchao/prototype/float8nocompile/kernels/fp8_dynamic_tensorwise.py
+++ b/torchao/prototype/float8nocompile/kernels/fp8_dynamic_tensorwise.py
@@ -61,7 +61,7 @@ def _fp8_scale(
 ):
     # calculate global amax across all blocks
     global_amax = tl.zeros([1], dtype=tl.float64)
-    num_blocks = (num_elements + BLOCK_SIZE - 1) // BLOCK_SIZE
+    num_blocks = tl.cdiv(num_elements, BLOCK_SIZE)
     for i in range(num_blocks):
         block_max = tl.load(block_amaxes_ptr + i)
         global_amax = tl.maximum(global_amax, block_max)
@@ -110,7 +110,9 @@ def triton_hp_tensor_to_float8_dynamic(
     gemm_input_role: GemmInputRole = GemmInputRole.INPUT,
 ) -> Float8Tensor:
 
-    BLOCK_SIZE = 8
+    assert hp_tensor.is_contiguous(), "tensor must be contiguous"
+
+    BLOCK_SIZE = 8  # TODO(danielvegamyhre): tune this for perf
 
     num_elements = hp_tensor.numel()
     orig_shape = hp_tensor.shape
--- a/torchao/prototype/float8nocompile/kernels/fp8_dynamic_tensorwise.py
+++ b/torchao/prototype/float8nocompile/kernels/fp8_dynamic_tensorwise.py
@@ -15,6 +15,7 @@ import triton.language as tl
 
 from torchao.float8.float8_tensor import Float8Tensor, GemmInputRole, LinearMMConfig
 
+EPS = 1e-12
 
 FP8_DTYPE_MAP = {
     torch.int8: tl.int8,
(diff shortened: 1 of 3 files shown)


torchao/prototype/float8nocompile/kernels/fp8_dynamic_tensorwise.py
```diff
@@ -0,0 +1,113 @@
...
+):
+    offs = tl.arange(0, BLOCK_SIZE)
+
+    # get amax
+    amax = tl.zeros([1], dtype=tl.float64)
+    for i in range(0, n_elements, BLOCK_SIZE):
+        block_offs = (i * BLOCK_SIZE) + offs
+        block_mask = block_offs < n_elements
+        vals = tl.load(input_ptr + block_offs, mask=block_mask).to(input_dtype)
+        amax = tl.maximum(amax, tl.max(tl.abs(vals)))
```
vkuzo, line 54: I think you might want atomic max here to prevent race conditions

torchao/prototype/float8nocompile/kernels/fp8_dynamic_tensorwise.py
```diff
@@ -0,0 +1,113 @@
...
+    n_elements,
+    input_dtype: tl.constexpr,
+    tensor_out_dtype: tl.constexpr,
+    BLOCK_SIZE: tl.constexpr,
+    EPS: tl.constexpr,
+):
+    offs = tl.arange(0, BLOCK_SIZE)
+
+    # get amax
+    amax = tl.zeros([1], dtype=tl.float64)
```
vkuzo, line 49: if I'm reading this code right, the `amax` value will be calculated per a single cell in the grid - I would have expected the same amax to be used for all cells in the grid, there should be a synchronization step somewhere

torchao/prototype/float8nocompile/kernels/fp8_dynamic_tensorwise.py
```diff
@@ -0,0 +1,113 @@
...
+    for i in range(0, n_elements, BLOCK_SIZE):
+        block_offs = (i * BLOCK_SIZE) + offs
+        block_mask = block_offs < n_elements
+        vals = tl.load(input_ptr + block_offs, mask=block_mask).to(input_dtype)
```
vkuzo, line 53: I would have expected a single `tl.load` per grid cell, if I'm understanding this code correctly it's both launching a grid and also iterating through every element of the tensor from each cell in the grid.  I could be missing something

danielvegamyhre: Oh I think I see the problem, I didn't update the grid dimensions when I changed the kernel implementation strategy. In this simple strategy there should only be 1 thread block / cell, which finds the global amax via loading one block at a time into SRAM in a loop as seen here. 

My original strategy was to divide the tensor into some number of blocks, compute a local amax for each block, write those to a shared buffer of some sort, then have a second step which computes the global amax from those local amaxes. However, I decided I wanted to try this simple approach first to just get the code functionally correct to start.

Unfortunately a SEV caused me to lose access to my devgpu but I will verify the fix tomorrow then make it more parallelized.

danielvegamyhre: @vkuzo I went back to the drawing board and rewrote this using the original strategy I had in mind (described in my previous comment) and it works now. It's a bit more complicated and requires breaking up the conversion process into separate kernels, but it works and uses a higher degree of parallelism so should be reasonably performant. 

However, I am wondering if it would be more efficient for each block in the final `_to_fp8()` kernel to locally compute the global scale from the previously computed `block_amaxes`, rather than have the separate kernel `_scale()` do it before hand and share that scale value everywhere. My reasoning is that all instances of the `_to_fp8()` kernel will be waiting on the `_scale()` kernel to finish before they can launch, and the memory access latency overhead of the `_scale()` kernel moving data back and forth from HBM <-> SRAM will be slower than just computing the scale in `_to_fp8()` directly.

If time allows I'd like to benchmark/profile both approaches.

danielvegamyhre:
Adding @drisspg for review as well, curious to get your thoughts on this initial implementation of triton kernels for fp8 conversion

torchao/prototype/float8nocompile/kernels/fp8_dynamic_tensorwise.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+def triton_hp_tensor_to_float8_dynamic(
+    hp_tensor: torch.Tensor,
+    fp8_dtype: torch.dtype,
+    linear_mm_config: LinearMMConfig,
+    gemm_input_role: GemmInputRole = GemmInputRole.INPUT,
+) -> Float8Tensor:
+
+    BLOCK_SIZE = 8
```
drisspg, line 113: Not sure if this PR is targetting Perf but this seems way to small

torchao/prototype/float8nocompile/kernels/fp8_dynamic_tensorwise.py
```diff
@@ -0,0 +1,175 @@
...
+    block_amaxes_ptr,
+    scale_out_ptr,
+    num_elements,
+    fp8_dtype_max,
+    BLOCK_SIZE: tl.constexpr,
+    EPS: tl.constexpr,
+):
+    # calculate global amax across all blocks
+    global_amax = tl.zeros([1], dtype=tl.float64)
+    num_blocks = (num_elements + BLOCK_SIZE - 1) // BLOCK_SIZE
```
drisspg, line 64: nit: tl.ceil_div

drisspg (commented):
One big thing here is that all your kernels assume contiguous inputs, either assert that or update to add for striding.

torchao/prototype/float8nocompile/kernels/fp8_dynamic_tensorwise.py
danielvegamyhre: Added a TODO comment here to tune this for perf. This is just a temporary placeholder value while I get the core functionality working e2e.

torchao/prototype/float8nocompile/kernels/fp8_dynamic_tensorwise.py
danielvegamyhre: Done

danielvegamyhre:
> One big thing here is that all your kernels assume contiguous inputs, either assert that or update to add for striding.

Thanks, added an assertion that input tensor is contiguous, as well as a unit test to validate this behavior.

drisspg (approved):

drisspg:
Side note here another fun way to do it: https://github.com/drisspg/transformer_nuggets/blob/a4c66bbeebaa479ad8b6ed82d7efbafa41b17260/transformer_nuggets/fp8/scaled_quant.py#L139

PR #1447, opened 2024-12-19
[Feat]: Add support for kleidiai quantization schemes

ng-05:
Description:
Allow int8_dynamic_activation_intx_weight to work with aten  _dyn_quant_matmul_4bit op
    
Needs : pytorch/pytorch#134124 or Pytorch > 2.6.0

### DIFF
--- a/torchao/dtypes/affine_quantized_tensor.py
+++ b/torchao/dtypes/affine_quantized_tensor.py
@@ -204,7 +204,6 @@ class AffineQuantizedTensor(TorchAOBaseTensor):
         zero_point_domain: ZeroPointDomain = ZeroPointDomain.INT,
         _layout: Layout = PlainLayout(),
         use_hqq: bool = False,
-        bias: Optional[torch.Tensor] = None
     ):
         """Convert a high precision tensor to an integer affine quantized tensor."""
         original_shape = input_float.shape
@@ -277,11 +276,7 @@ class AffineQuantizedTensor(TorchAOBaseTensor):
 
         data = _layout.post_process(data)
         tensor_impl_ctr = get_tensor_impl_constructor(type(_layout))
-        args = [data, scale, zero_point, _layout]
-        # Only PackedLinearInt8DynamicActivationIntxWeightLayout() with "aten" target supports bias
-        if bias is not None:
-            args.append(bias)
-        tensor_impl = tensor_impl_ctr(*args)
+        tensor_impl = tensor_impl_ctr(data, scale, zero_point, _layout)
         return cls(
             tensor_impl,
             block_size,
--- a/torchao/dtypes/affine_quantized_tensor.py
+++ b/torchao/dtypes/affine_quantized_tensor.py
@@ -204,6 +204,7 @@ class AffineQuantizedTensor(TorchAOBaseTensor):
         zero_point_domain: ZeroPointDomain = ZeroPointDomain.INT,
         _layout: Layout = PlainLayout(),
         use_hqq: bool = False,
+        bias: Optional[torch.Tensor] = None
     ):
         """Convert a high precision tensor to an integer affine quantized tensor."""
         original_shape = input_float.shape
@@ -276,7 +277,11 @@ class AffineQuantizedTensor(TorchAOBaseTensor):
 
         data = _layout.post_process(data)
         tensor_impl_ctr = get_tensor_impl_constructor(type(_layout))
-        tensor_impl = tensor_impl_ctr(data, scale, zero_point, _layout)
+        args = [data, scale, zero_point, _layout]
+        # Only PackedLinearInt8DynamicActivationIntxWeightLayout() with "aten" target supports bias
+        if bias is not None:
+            args.append(bias)
+        tensor_impl = tensor_impl_ctr(*args)
         return cls(
             tensor_impl,
             block_size,
--- a/torchao/experimental/packed_linear_int8_dynamic_activation_intx_weight_layout.py
+++ b/torchao/experimental/packed_linear_int8_dynamic_activation_intx_weight_layout.py
@@ -5,6 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 import logging
+from enum import Enum, auto
 from typing import Optional, Tuple
 
 import torch
@@ -20,6 +21,9 @@ from torchao.dtypes.utils import AQTTensorImpl, Layout
 from torchao.quantization.quant_primitives import (
     ZeroPointDomain,
 )
+from torchao.utils import (
+    TORCH_VERSION_AT_LEAST_2_6,
+)
 
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.WARNING)
@@ -31,17 +35,33 @@ formatter = logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(messag
 handler.setFormatter(formatter)
 logger.addHandler(handler)
 
+class Target(Enum):
+    """Enum that indicates the backend target"""
+
+    NATIVE = auto()
+    ATEN = auto()
+
+def target_from_str(target: str) -> Target:
+    if target.lower() == "native":
+        return Target.NATIVE
+    elif target.lower() == "aten":
+        return Target.ATEN
+    else:
+        raise ValueError(f"Invalid target: {target}")
 
 class PackedLinearInt8DynamicActivationIntxWeightLayout(Layout):
     bit_width: Optional[int]
     group_size: Optional[int]
     has_weight_zeros: Optional[bool]
+    # The target platform for the layout, 'native' or 'aten'
+    target: Optional[Target]
 
     def __init__(
         self,
         bit_width: Optional[int] = None,
         group_size: Optional[int] = None,
         has_weight_zeros: Optional[bool] = None,
+        target: Optional[str] = "native",
     ):
         if bit_width is not None:
             assert bit_width >= 1 and bit_width <= 8, "bit_width must be 1 to 8"
@@ -51,6 +71,7 @@ class PackedLinearInt8DynamicActivationIntxWeightLayout(Layout):
         self.bit_width = bit_width
         self.group_size = group_size
         self.has_weight_zeros = has_weight_zeros
+        self.target = target_from_str(target)
 
         if not self.has_params_set():
             assert (
@@ -60,13 +81,14 @@ class PackedLinearInt8DynamicActivationIntxWeightLayout(Layout):
             ), "bit_width, group_size, and has_weight_zeros must be None if has_params_set is False"
 
     def extra_repr(self):
-        return f"group_size={self.group_size}, bit_width={self.bit_width}, has_weight_zeros={self.has_weight_zeros}"
+        return f"group_size={self.group_size}, bit_width={self.bit_width}, has_weight_zeros={self.has_weight_zeros}, target={self.target}"
 
     def has_params_set(self) -> bool:
         return (
             (self.bit_width is not None)
             and (self.group_size is not None)
             and (self.has_weight_zeros is not None)
+            and (self.target is not None)
         )
 
 
@@ -125,9 +147,11 @@ class PackedLinearInt8DynamicActivationIntxWeightAQTTensorImpl(AQTTensorImpl):
         scale: torch.Tensor,
         zero_point: Optional[torch.Tensor],
         layout: Layout,
+        bias: Optional[torch.Tensor] = None,
     ):
         assert isinstance(layout, PackedLinearInt8DynamicActivationIntxWeightLayout)
         assert layout.has_params_set(), "PackedLinearInt8DynamicActivationIntxWeightLayout params must be set before calling from_plain"
+        assert layout.target in {Target.NATIVE, Target.ATEN}, f"Unexpected target: {layout.target}"
 
         # TODO(T200095131): remove group_size_tensor, n_tensor, k_tensor
         # when AOTI supports int
@@ -135,6 +159,13 @@ class PackedLinearInt8DynamicActivationIntxWeightAQTTensorImpl(AQTTensorImpl):
         group_size_tensor = torch.empty(0, layout.group_size, dtype=torch.int8)
         n_tensor = torch.empty(0, n, dtype=torch.int8)
         k_tensor = torch.empty(0, k, dtype=torch.int8)
+
+        if layout.target == Target.ATEN:
+            assert TORCH_VERSION_AT_LEAST_2_6, f"aten target is requires torch version > 2.6.0"
+            int_data = int_data.add(8)
+            int_data = (int_data[::,1::2]  << 4 | int_data[::,::2] ).to(torch.uint8)
+            packed_weight = torch.ops.aten._dyn_quant_pack_4bit_weight(int_data, scale, bias, layout.group_size, k, n)
+            return cls(packed_weight, layout, group_size_tensor, n_tensor, k_tensor)
 
         if layout.has_weight_zeros:
             args = [
@@ -211,16 +242,13 @@ class PackedLinearInt8DynamicActivationIntxWeightAQTTensorImpl(AQTTensorImpl):
 def _linear_check(input_tensor, weight_tensor, bias):
     layout = weight_tensor.tensor_impl.get_layout()
     return isinstance(layout, PackedLinearInt8DynamicActivationIntxWeightLayout) and (
-        bias is None
+        bias is None or layout.target == Target.ATEN # Aten target allows bias
     )
 
 
 def _linear_impl(input_tensor, weight_tensor, bias):
-    assert (
-        bias is None
-    ), "bias in linear is not supported for PackedLinearInt8DynamicActivationIntxWeightAQTTensorImpl"
 
-    def _impl_2d(input_tensor, weight_tensor):
+    def _impl_2d_native(input_tensor, weight_tensor):
         assert input_tensor.dim() == 2
         assert weight_tensor.dim() == 2
 
@@ -255,6 +283,31 @@ def _linear_impl(input_tensor, weight_tensor, bias):
             torch.ops.torchao, f"_linear_8bit_act_{bit_width}bit{wzp_suffix}_weight"
         )(*args)
 
+    def _impl_2d_aten(input_tensor, weight_tensor):
+        assert input_tensor.dim() == 2
+        assert weight_tensor.dim() == 2
+
+        m, k = input_tensor.shape
+        n, k_ = weight_tensor.shape
+        assert k_ == k
+        group_size = weight_tensor.tensor_impl.get_layout().group_size
+        packed_weight = weight_tensor.tensor_impl.packed_weight
+        return torch.ops.aten._dyn_quant_matmul_4bit(
+            input_tensor, packed_weight, group_size, k_, n)
+
+    target = weight_tensor.tensor_impl.get_layout().target
+
+    if target == Target.ATEN:
+        assert (
+            TORCH_VERSION_AT_LEAST_2_6 == 1
+        ), "Target.ATEN requires torch >= 2.6.0"
+        _impl_2d = _impl_2d_aten
+    elif target == Target.NATIVE:
+        _impl_2d = _impl_2d_native
+        assert (
+            bias is None
+        ), "bias in linear is not supported for PackedLinearInt8DynamicActivationIntxWeightAQTTensorImpl with target 'native' "
+
     if input_tensor.dim() == 2:
         return _impl_2d(input_tensor, weight_tensor)
 
@@ -267,7 +320,6 @@ def _linear_impl(input_tensor, weight_tensor, bias):
     res = _impl_2d(input_tensor.reshape(-1, k), weight_tensor)
     res = res.reshape(*lead_shape, m, n)
     return res
-
 
 register_aqt_quantized_linear_dispatch(
     _linear_check,
--- a/torchao/experimental/packed_linear_int8_dynamic_activation_intx_weight_layout.py
+++ b/torchao/experimental/packed_linear_int8_dynamic_activation_intx_weight_layout.py
@@ -12,6 +12,8 @@ import torch
 from torch.utils._python_dispatch import return_and_correct_aliasing
 
 from torchao.dtypes.affine_quantized_tensor import (
+    AffineQuantizedTensor,
+    get_tensor_impl_constructor,
     register_layout,
 )
 from torchao.dtypes.affine_quantized_tensor_ops import (
@@ -20,7 +22,11 @@ from torchao.dtypes.affine_quantized_tensor_ops import (
 from torchao.dtypes.utils import AQTTensorImpl, Layout
 from torchao.quantization.quant_primitives import (
     ZeroPointDomain,
+    MappingType,
+    choose_qparams_affine,
+    quantize_affine,
 )
+
 from torchao.utils import (
     TORCH_VERSION_AT_LEAST_2_6,
 )
@@ -325,3 +331,78 @@ register_aqt_quantized_linear_dispatch(
     _linear_check,
     _linear_impl,
 )
+
+
+class PackedLinearInt8DynamicActivationIntxWeightAtenTensor(AffineQuantizedTensor):
+    """
+    PackedLinearInt8DynamicActivationIntxWeightAtenTensor quantized tensor subclass which inherits AffineQuantizedTensor class.
+    """
+
+    @classmethod
+    def from_hp_to_intx(
+        cls,
+        input_float: torch.Tensor,
+        mapping_type: MappingType,
+        block_size: Tuple[int, ...],
+        target_dtype: torch.dtype,
+        quant_min: Optional[int] = None,
+        quant_max: Optional[int] = None,
+        eps: Optional[float] = None,
+        scale_dtype: Optional[torch.dtype] = None,
+        zero_point_dtype: Optional[torch.dtype] = None,
+        preserve_zero: bool = True,
+        zero_point_domain: Optional[ZeroPointDomain] = ZeroPointDomain.INT,
+        _layout: Layout = PackedLinearInt8DynamicActivationIntxWeightLayout(),
+        use_hqq: bool = False,
+        bias: Optional[torch.Tensor] = None
+    ):
+        assert use_hqq == False, f"PackedLinearInt8DynamicActivationIntxWeightTensor can not support HQQ optimization"
+        assert isinstance(
+            _layout, PackedLinearInt8DynamicActivationIntxWeightLayout), f"PackedLinearInt8DynamicActivationIntxWeightTensor can only support PackedLinearInt8DynamicActivationIntxWeightLayout(). Provided {_layout}"
+        assert _layout.target == Target.ATEN, f"PackedLinearInt8DynamicActivationIntxWeightTensor requires target 'aten'."
+        original_shape = input_float.shape
+        input_float = _layout.pre_process(input_float)
+
+        scale, zero_point = choose_qparams_affine(
+            input_float,
+            mapping_type,
+            block_size,
+            target_dtype,
+            quant_min,
+            quant_max,
+            eps,
+            scale_dtype,
+            zero_point_dtype,
+            preserve_zero,
+            zero_point_domain,
+        )
+        # choose_qparams_affine is a custom op that does support returning optional Tensors. We thus set the zero_point to None if its domain is None
+        # TODO should probably consolidate ZeroPointDomain.NONE and None
+        if zero_point_domain is None or zero_point_domain == ZeroPointDomain.NONE:
+            zero_point = None
+        data = quantize_affine(
+            input_float,
+            block_size,
+            scale,
+            zero_point,
+            target_dtype,
+            quant_min,
+            quant_max,
+            zero_point_domain,
+        )
+        # Note: output will be uint8 tensor for sub byte tensors for now
+
+        data = _layout.post_process(data)
+        tensor_impl_ctr = get_tensor_impl_constructor(type(_layout))
+        tensor_impl = tensor_impl_ctr(data, scale, zero_point, _layout, bias)
+        return cls(
+            tensor_impl,
+            block_size,
+            original_shape,
+            quant_min,
+            quant_max,
+            zero_point_domain,
+            dtype=input_float.dtype,
+        )
+
+to_packedlinearint8dynamicactivationintxweight_quantized_intx = PackedLinearInt8DynamicActivationIntxWeightAtenTensor.from_hp_to_intx
--- a/torchao/experimental/quant_api.py
+++ b/torchao/experimental/quant_api.py
@@ -494,6 +494,7 @@ class IntxWeightEmbeddingQuantizer:
 
 from torchao.experimental.packed_linear_int8_dynamic_activation_intx_weight_layout import (
     PackedLinearInt8DynamicActivationIntxWeightLayout,
+    to_packedlinearint8dynamicactivationintxweight_quantized_intx,
 )
 from torchao.quantization.linear_activation_quantized_tensor import (
     to_linear_activation_quantized,
@@ -576,6 +577,7 @@ def int8_dynamic_activation_intx_weight(
         )
     bit_width = dtype_to_bit_width[weight_dtype]
     layout_arg = layout
+    propagate_bias = isinstance(layout_arg, PackedLinearInt8DynamicActivationIntxWeightLayout) and layout_arg.target=="aten"
 
     def apply(weight, bias: Optional[torch.Tensor] = None):
         if isinstance(granularity, PerGroup):
@@ -591,6 +593,10 @@ def int8_dynamic_activation_intx_weight(
 
         layout = layout_arg
         scale_dtype = None
+        tensor_quantizer = to_affine_quantized_intx
+        quant_min = -(1 << (bit_width - 1))
+        quant_max = (1 << (bit_width - 1)) - 1
+
         if isinstance(layout, PackedLinearInt8DynamicActivationIntxWeightLayout):
             assert (
                 weight.device == torch.device("cpu")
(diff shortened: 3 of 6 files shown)


torchao/quantization/GPTQ.py
jerryzh168, line 1: we plan to deprecate the APIs in this file actually, you should be extending APIs from this file instead: https://github.com/pytorch/ao/blob/main/torchao/quantization/quant_api.py

ng-05: Thanks, Noted!

ng-05:
Hello @jerryzh168 ,
We want to support two diff type of int4 schemes. 
1. symmetric_groupwise -> groupsize [ 32, 64, 128 etc ]
2. symmetric_channelwise -> groupsize is equal to channelsize of the matmul weights


How should we take this input from user regarding quantization schemes. Groupsize parameter can not server the purpose as channelsize will change for diff matmuls in a model?

Currently I am using "scheme" parameter to differentiate between the two.
[aarch64_cpu_channelwise.json](https://github.com/ARM-software/Tool-Solutions/blob/main/ML-Frameworks/pytorch-aarch64/examples/gen_ai_utils/quant_configs/aarch64_cpu_channelwise.json)
[aarch64_cpu_groupwise.json](https://github.com/ARM-software/Tool-Solutions/blob/main/ML-Frameworks/pytorch-aarch64/examples/gen_ai_utils/quant_configs/aarch64_cpu_groupwise.json)

jerryzh168:
> How should we take this input from user regarding quantization schemes. Groupsize parameter can not server the purpose as channelsize will change for diff matmuls in a model?

yeah, you can use https://github.com/pytorch/ao/blob/main/torchao/quantization/granularity.py (PerGroup and PerAxis(axis=0)), examples: https://github.com/pytorch/ao/blob/070345d9676f3dd6cdd325e987764fc3c71ccaf5/torchao/quantization/quant_api.py#L1069, https://github.com/pytorch/ao/blob/070345d9676f3dd6cdd325e987764fc3c71ccaf5/tutorials/calibration_flow/static_quant.py#L168

ng-05:
> > How should we take this input from user regarding quantization schemes. Groupsize parameter can not server the purpose as channelsize will change for diff matmuls in a model?
> 
> yeah, you can use https://github.com/pytorch/ao/blob/main/torchao/quantization/granularity.py: PerGroup and PerAxis(axis=0) (assuming channel dimension is 0), examples:
> 
> https://github.com/pytorch/ao/blob/070345d9676f3dd6cdd325e987764fc3c71ccaf5/torchao/quantization/quant_api.py#L1069
> 
> ,
> https://github.com/pytorch/ao/blob/070345d9676f3dd6cdd325e987764fc3c71ccaf5/tutorials/calibration_flow/static_quant.py#L168

Thanks for the inputs @jerryzh168.

I have initial change ready which extends [int4_weight_only](https://github.com/pytorch/ao/blob/main/torchao/quantization/quant_api.py#L687) quantizer. 

The 4 bit KleidiAI kernels quantizes the weight in torchao and input to 8 bit within the kernel itself instead of quantizing the input in the torchao the way [int8_dynamic_activation_int4_weight](https://github.com/pytorch/ao/blob/main/torchao/quantization/quant_api.py#L631) does. 
For this reason I am extending the int4_weight_only api. I am slightly confused if the intention of this api is to convey NO input quantisation to user?

Currently neither int4_weight_only nor int8_dynamic_activation_int4_weight fully aligns with the way kelidiai 4 bit kernels are working. 

I feel int4_weight_only is closest to what we want to do, what are your thoughts on this?

jerryzh168:
> I feel int4_weight_only is closest to what we want to do, what are your thoughts on this?

yeah int4_weight_only means no input quantization, I think it aligns better with `int8_dynamic_activation_int4_weight`, you can use a different layout and customize the logic for input quantization.

we also have https://github.com/pytorch/ao/blob/47383779c0975fd7ec213bc7b07c34ca5a1e5abb/torchao/experimental/quant_api.py#L485 that is the same as your use case. there is some ongoing refactors/updates there as well right now

ng-05:
Hello @jerryzh168 , I am planning to migrate ```int8_dynamic_activation_intx_weight``` api  to ```int8_dynamic_activation_intx_weight_v2```. 
For now I have kept the API separate for review and testing. 

Can you please review this change, specially the change the in ```_get_linear_subclass_inserter``` which allow bias propagation. The bias needed by ```torch.ops.aten._dyn_quant_pack_4bit_weight```.

torchao/experimental/_linear_8bit_act_xbit_weight_layout.py
```diff
@@ -56,22 +62,27 @@ class Linear8BitActXBitWeightLayout(Layout):
     nbit: int
     group_size: int
 
-    # The target platform for the layout, either 'native' or 'fallback'.
+    # The target platform for the layout, 'native', 'fallback' or 'aten'
     target: Target
 
+    # Allow bias access via layout
+    bias: Optional[torch.Tensor] = None
```
jerryzh168, line 69: layout is more of a "type" actually, why is bias Tensor passed here?

torchao/quantization/quant_api.py
```diff
@@ -450,15 +450,18 @@ def _linear_extra_repr(self):
...
 
-def _get_linear_subclass_inserter(constructor, *, allow_requires_grad=False, **kwargs):
+def _get_linear_subclass_inserter(constructor, *, allow_requires_grad=False, propagate_bias=False, **kwargs):
     """Helper function to apply the constructor that quantizes the weight Tensor (with additional kwargs)
     to the weight of linear module
     """
 
     def insert_subclass(lin):
         requires_grad = allow_requires_grad and lin.weight.requires_grad
+        args = [lin.weight]
```
jerryzh168, line 460: nit: I feel putting optional args in kwargs might be better

jerryzh168 (approved):
looks good to me overall, can you add some tests?

jerryzh168:
> I am also not sure if int8_dynamic_activation_intx_weight* quantizer can be accessed by torchchat currently? Do you have an example how torchchat can pass args like granularity, mapping_type from torchchat cli to torchao ?

I don't think we need to expose these fine grained args to torchchat cli, we just need these high level args like: https://github.com/pytorch/torchchat/blob/main/torchchat/quant_config/mobile.json

we are also working on migrating torchchat to use torchao quant api btw

torchao/experimental/_linear_8bit_act_xbit_weight_layout.py
```diff
@@ -100,6 +110,12 @@ def _pack_weights_native(
             torch.empty(0, group_size, dtype=torch.int8),
         ]
 
+    if TORCH_VERSION_AT_LEAST_2_6 and layout.target == Target.ATEN:
```
kimishpatel, line 113: If torch version is not 2.6 but layout.target == aten, then what happens? Should you just assert that it is not supported?

torchao/experimental/_linear_8bit_act_xbit_weight_layout.py
```diff
@@ -353,6 +377,51 @@ def _impl_2d(input_tensor, weight_tensor):
...
+        )
+    assert (
+        weight_tensor.tensor_impl.get_layout().nbit == 4
+    ), f"Only 4 bit is supported"
+    assert (
+        TORCH_VERSION_AT_LEAST_2_6 == 1
+    ), "Target.ATEN requires torch >= 2.6.0"
+    # aten supports bias for kleidiAI but not for default fallback op
+    if not torch.backends.kleidiai.is_available():
+        print("TODO bias == None")
```
kimishpatel, line 396: assert bias == None, <msg>

torchao/experimental/_linear_8bit_act_xbit_weight_layout.py
```diff
@@ -353,6 +377,51 @@ def _impl_2d(input_tensor, weight_tensor):
...
+
+        m, k = input_tensor.shape
+        n, k_ = weight_tensor.shape
+        assert k_ == k
+        group_size = weight_tensor.tensor_impl.get_layout().group_size
+        packed_weight = weight_tensor.tensor_impl.packed_weight
+        return torch.ops.aten._dyn_quant_matmul_4bit(
+            input_tensor, packed_weight, group_size, k_, n)
+
+    if input_tensor.dim() == 2:
```
kimishpatel, line 411: Do you have this requirement?

torchao/experimental/quant_api.py
```diff
@@ -482,6 +492,104 @@ def quantize(self, model: nn.Module) -> nn.Module:
         return model
 
 
+_intx_granularity = Union[PerGroup, PerRow]
+
+
+def int8_dynamic_activation_intx_weight_v2(
```
metascroy, line 498: I'm currently refactoring int8_dynamic_activation_intx_weight quantizer to use layout instead of target for the packing format: https://github.com/pytorch/ao/pull/1553.  I think this should provide more flexibility longterm.

torchao/experimental/_linear_8bit_act_xbit_weight_layout.py
```diff
@@ -153,7 +169,7 @@ def get_layout(self) -> Layout:
     def get_plain(
         self,
     ) -> Tuple[torch.Tensor, Optional[torch.Tensor], Optional[torch.Tensor]]:
-        if self.get_layout().target == Target.FALLBACK:
+        if self.get_layout().target == Target.FALLBACK or self.get_layout().target == Target.ATEN:
             return self.packed_weight, self.scale, self.zero_point
```
metascroy, line 173: IIUC when using Target.ATEN, self.packed_weight is not the int_data, so I'm not sure get_plain is correct here?

torchao/experimental/quant_api.py
kimishpatel, line 498: is there a significant diffrence between this and int8_dynamic_activation_intx_weight?

kimishpatel (changes requested):
I understand that this quant API now connects kernels we landed in aten with quant API. If the kernels you guys landed in aten are actually new ops, unlike int4pack_mm and friends, then why did we land them there in the first place. In order to reach those kernels you need ao dep anyway? (@digantdesai I know you tagged me on that PR but i never really deep dived into that so maybe you have context here)

Besides taht i have a couple of questions.
- In the current form it is only making aten op you guys added available via tensor subclass api, so what happens to say torch.compile (maybe this works?) or AOTI usecase?
- I would also like to see if we can leverage this op in executorch, for which integration into AO would have been a better choice compared to this being aten op
- If kleidi's op performs better than whats in this repo (and note that @digantdesai has actually integrated some of the kleidi ops that I guess you guys are aware of), then can we just use that op directly or have a path to kleidi's impl for the cpu ops that exist under experimental/ops?

kimishpatel:
> > Hello @jerryzh168 , I am planning to migrate `int8_dynamic_activation_intx_weight` api to `int8_dynamic_activation_intx_weight_v2`. For now I have kept the API separate for review and testing.
> > Can you please review this change, specially the change the in `_get_linear_subclass_inserter` which allow bias propagation. The bias needed by `torch.ops.aten._dyn_quant_pack_4bit_weight`.
> > I am also not sure if int8_dynamic_activation_intx_weight* quantizer can be accessed by torchchat currently? Do you have an example how torchchat can pass args like granularity, mapping_type from torchchat cli to torchao ?
> 
> torchchat does not currently use int8_dynamic_activation_intx_weight, but instead a submodule swap API here: [`main`/torchao/experimental/quant_api.py#L438](https://github.com/pytorch/ao/blob/main/torchao/experimental/quant_api.py#L438)
> 
> We will be switching torchchat to use int8_dynamic_activation_intx_weight instead, but I first need to land some changes for perf/clarity: #1553

Any specific reason why use subclass API instead of module swap?

torchao/experimental/_linear_8bit_act_xbit_weight_layout.py
ng-05: I want to access the bias to be packed with my weights and scales. I can not find any other existing way to pass bias to ```from_plain()``` api via https://github.com/pytorch/ao/blob/ad61822f4101625092dd8d2d59d4be6438341791/torchao/dtypes/affine_quantized_tensor.py#L281

How do you think I should access bias in the packing function here. https://github.com/pytorch/ao/blob/ad61822f4101625092dd8d2d59d4be6438341791/torchao/experimental/_linear_8bit_act_xbit_weight_layout.py#L77

torchao/quantization/quant_api.py
ng-05: If you put bias in optional kwargs, all the functions similar to "apply()" will start complaining and need changes

torchao/experimental/_linear_8bit_act_xbit_weight_layout.py
ng-05: Good Catch. Should be assert

torchao/experimental/_linear_8bit_act_xbit_weight_layout.py
ng-05: Good Catch, thanks. I Missed this

torchao/experimental/_linear_8bit_act_xbit_weight_layout.py
ng-05: Yes we only support 2d Input tensor in ```torch.ops.aten._dyn_quant_matmul_4bit```

ng-05:
> I understand that this quant API now connects kernels we landed in aten with quant API. If the kernels you guys landed in aten are actually new ops, unlike int4pack_mm and friends, then why did we land them there in the first place. In order to reach those kernels you need ao dep anyway? (@digantdesai I know you tagged me on that PR but i never really deep dived into that so maybe you have context here)
> 
> Besides taht i have a couple of questions.
> 
> * In the current form it is only making aten op you guys added available via tensor subclass api, so what happens to say torch.compile (maybe this works?) or AOTI usecase?
> * I would also like to see if we can leverage this op in executorch, for which integration into AO would have been a better choice compared to this being aten op
> * If kleidi's op performs better than whats in this repo (and note that @digantdesai has actually integrated some of the kleidi ops that I guess you guys are aware of), then can we just use that op directly or have a path to kleidi's impl for the cpu ops that exist under experimental/ops?

I am unaware of executorch status and what performance you get with klediai kernels over there. I tested this change with torch.compile() and it seems to be working fine.

metascroy:
> > > Hello @jerryzh168 , I am planning to migrate `int8_dynamic_activation_intx_weight` api to `int8_dynamic_activation_intx_weight_v2`. For now I have kept the API separate for review and testing.
> > > Can you please review this change, specially the change the in `_get_linear_subclass_inserter` which allow bias propagation. The bias needed by `torch.ops.aten._dyn_quant_pack_4bit_weight`.
> > > I am also not sure if int8_dynamic_activation_intx_weight* quantizer can be accessed by torchchat currently? Do you have an example how torchchat can pass args like granularity, mapping_type from torchchat cli to torchao ?
> > 
> > 
> > torchchat does not currently use int8_dynamic_activation_intx_weight, but instead a submodule swap API here: [`main`/torchao/experimental/quant_api.py#L438](https://github.com/pytorch/ao/blob/main/torchao/experimental/quant_api.py#L438)
> > We will be switching torchchat to use int8_dynamic_activation_intx_weight instead, but I first need to land some changes for perf/clarity: #1553
> 
> Any specific reason why use subclass API instead of module swap?

My understanding from @jerryzh168 is that long-term, torchao plans to support pt2e and subclass/quantize_ based quantization long-term.  I believe torchchat is working on (and has already partially completed) moving module-swap based quantization over to use quantize_ (cc @Jack-Khuu to keep me honest there).

torchao/experimental/quant_api.py
metascroy: In the new flow, fallback is achieved by passing layout=PlainLayout() to int8_dynamic_activation_intx_weight.  Native is achieved by passing PackedLinearInt8DynamicActivationIntxWeightLayout() as layout.

You would need to add a new layout for your op, e.g., PackedLinearInt8DynamicActivationIntxWeightKleidiAILayout or something (for what you call target=ATEN in this PR).

If you integrate the KleidiAI kernels with torchao as @digantdesai mentioned, you should be able to reuse PackedLinearInt8DynamicActivationIntxWeightLayout and KleidiAI can be dispatched to inside the torchao ops when appropriate.  If absolutely needed, we can add a target to the constructor for PackedLinearInt8DynamicActivationIntxWeightLayout. 

> I do not want to build and load torchao ops lib if I can directly leverage aten op

As an FYI, these ops are now built and loaded when torchao is pip installed locally.  Albeit they are still not present in the nightlies.

kimishpatel:
> > I understand that this quant API now connects kernels we landed in aten with quant API. If the kernels you guys landed in aten are actually new ops, unlike int4pack_mm and friends, then why did we land them there in the first place. In order to reach those kernels you need ao dep anyway? (@digantdesai I know you tagged me on that PR but i never really deep dived into that so maybe you have context here)
> > Besides taht i have a couple of questions.
> > 
> > * In the current form it is only making aten op you guys added available via tensor subclass api, so what happens to say torch.compile (maybe this works?) or AOTI usecase?
> > * I would also like to see if we can leverage this op in executorch, for which integration into AO would have been a better choice compared to this being aten op
> > * If kleidi's op performs better than whats in this repo (and note that @digantdesai has actually integrated some of the kleidi ops that I guess you guys are aware of), then can we just use that op directly or have a path to kleidi's impl for the cpu ops that exist under experimental/ops?
> 
> I am unaware of executorch status and what performance you get with klediai kernels over there. I tested this change with torch.compile() and it seems to be working fine.

are you planning to move dynamic quantized ops in aten to torchao?

jerryzh168:
> 
> Any specific reason why use subclass API instead of module swap?

you mean quantize_ API right? it is the officially supported API for inference path of torchao

jerryzh168:
agree with @kimishpatel that we want these ops in ao instead of aten, but I talked to @digantdesai last time he explains here: https://github.com/pytorch/pytorch/issues/143289#issuecomment-2555819375

torchao/experimental/_linear_8bit_act_xbit_weight_layout.py
jerryzh168: If it does not fit into `AQT`, I think it's fine to create a new tensor subclass

jerryzh168: layout is following the design of https://pytorch.org/docs/stable/tensor_attributes.html#torch.layout

digantdesai:
> are you planning to move dynamic quantized ops in aten to torchao?

@kimishpatel - long term yes, but in the short term we might support `aten` backend here. Two main blockers, 
(1) as I said earlier on the ATen op PR https://github.com/pytorch/pytorch/issues/143289#issuecomment-2555819375 about Kleidi in two places, and 
(2) solidify out the torchAO apis (bias packing, layout clean up from Scott R., module swap to dispatch), move out of experimental, and integrate Kleidi ops with low-bit kernels.

I think if we unblock leveraging ATen op for now from here for eager/compile, we can solve (1) and (2) without blocking actual LLM use cases, this was a time sensitive demo request from Arm side for PT2.6.

torchao/experimental/_linear_8bit_act_xbit_weight_layout.py
ng-05: What is the idea behind get_plain() api? Does it return the unpacked original tensors of the Linear layer? If yes, why is it returning the packed weights in original implementation?
@metascroy what do you mean by ```int_data```  in your first comment?

torchao/experimental/quant_api.py
ng-05: I will add ```has_bias``` and ```target``` to ```PackedLinearInt8DynamicActivationIntxWeightLayout()```

kimishpatel:
> > are you planning to move dynamic quantized ops in aten to torchao?
> 
> @kimishpatel - long term yes, but in the short term we might support `aten` backend here. Two main blockers, (1) as I said earlier on the ATen op PR [pytorch/pytorch#143289 (comment)](https://github.com/pytorch/pytorch/issues/143289#issuecomment-2555819375) about Kleidi in two places, and (2) solidify out the torchAO apis (bias packing, layout clean up from Scott R., module swap to dispatch), move out of experimental, and integrate Kleidi ops with low-bit kernels.
> 
> I think if we unblock leveraging ATen op for now from here for eager/compile, we can solve (1) and (2) without blocking actual LLM use cases, this was a time sensitive demo request from Arm side for PT2.6.

Thanks for the context Digant. Lets make sure we make progress on the long term front

ng-05:
I have done the changes as per the review comments and latest refactoring ```(PackedLinearInt8DynamicActivationIntxWeightLayout)```
Can I please get a review @metascroy @digantdesai

torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -206,6 +206,7 @@ def from_hp_to_intx(
         zero_point_domain: Optional[ZeroPointDomain] = ZeroPointDomain.INT,
         _layout: Layout = PlainLayout(),
         use_hqq: bool = False,
+        bias: Optional[torch.Tensor] = None
```
jerryzh168, line 209: probably better to inherit from AQT and customize this for now I feel, see https://github.com/pytorch/ao/blob/cf453360dd3e09394657172f8e8d8da23cfbf043/torchao/dtypes/uintx/marlin_qqq_tensor.py#L31 for an example

torchao/quantization/quant_api.py
jerryzh168: why? I thought we still only put it when propagate_bias is True

kimishpatel:
I have expressed my high level concerns which has been discussed. So I am gonna leave rest of the review to @metascroy and @digantdesai

torchao/dtypes/affine_quantized_tensor.py
ng-05: Thanks for your input @jerryzh168 . I have made the changes as per your suggestion.

torchao/quantization/quant_api.py
ng-05: sorry I thought you wanted lin.weight to be passed as part of kwargs. lin.bias can be passed as part of kwargs. This change is done

torchao/experimental/packed_linear_int8_dynamic_activation_intx_weight_layout.py
```diff
@@ -31,17 +41,33 @@
...
+        return Target.NATIVE
+    elif target.lower() == "aten":
+        return Target.ATEN
+    else:
+        raise ValueError(f"Invalid target: {target}")
 
 class PackedLinearInt8DynamicActivationIntxWeightLayout(Layout):
     bit_width: Optional[int]
     group_size: Optional[int]
     has_weight_zeros: Optional[bool]
```
metascroy, line 61: The packed weights from Kleidi have bias packed with them, right?  If so, let's add has_bias: Optional[bool] here to layout.

torchao/experimental/quant_api.py
```diff
@@ -531,13 +538,28 @@ def int8_dynamic_activation_intx_weight(
...
-    except AttributeError:
-        raise Exception(
-            "TorchAO experimental kernels are not loaded.  To install the kernels, run `USE_CPP=1 pip install .` from ao on a machine with an ARM CPU."
-            + "  Alternatively, use layout=PlainLayout() with int8_dynamic_activation_intx_weight, but note that doing so will result in much slower performance."
-        )
+
+    if target == "aten":
+        if not isinstance(layout, PackedLinearInt8DynamicActivationIntxWeightLayout) or \
+                weight_dtype != torch.int4 or \
+                has_weight_zeros != True or \
```
metascroy, line 545: It looks like the KleidiAI op does not take the zero points during packing (scale-only quantization)?  So shouldn't has_weight_zeros be false?

torchao/experimental/quant_api.py
```diff
@@ -506,6 +512,7 @@ def int8_dynamic_activation_intx_weight(
     weight_dtype: torch.dtype = torch.int4,
     granularity: Union[PerRow, PerGroup] = PerGroup(128),
     has_weight_zeros: bool = False,
+    target: str = "native",
```
metascroy, line 515: I think it would be better to pass this in layout's constructor because it isn't related to the quantization intent, but packing format/kernel selection e.g., layout= PackedLinearInt8DynamicActivationIntxWeightLayout(target="native")

torchao/experimental/quant_api.py
```diff
@@ -531,13 +538,28 @@ def int8_dynamic_activation_intx_weight(
...
+                has_weight_zeros != True or \
+                weight_mapping_type != MappingType.SYMMETRIC:
+            raise NotImplementedError(
+                f"target 'aten' requires:\n"
+                f"- layout to be PackedLinearInt8DynamicActivationIntxWeightLayout,\n"
+                f"- has_weight_zeros to be True,\n"
+                f"- weight_dtype to be torch.int4,\n"
+                f"- weight_mapping_type to be MappingType.SYMMETRIC"
+            )
+    elif not isinstance(layout, PlainLayout):
```
metascroy, line 554: Guard this try/except on `if isisntance(layout, PackedLinearInt8DynamicActivationIntxWeightLayout)` instead.  In case other layout is added in future, guarding on not PlainLayout is too broad.

torchao/experimental/quant_api.py
```diff
@@ -584,25 +612,30 @@ def apply(weight):
...
-            zero_point_domain=ZeroPointDomain.INT
-            if has_weight_zeros
-            else ZeroPointDomain.NONE,
-            _layout=layout,
-        )
+            if target == "aten":
+                assert TORCH_VERSION_AT_LEAST_2_6, f"aten target is requires torch version > 2.6.0"
+                if torch.backends.kleidiai.is_available():
+                    if isinstance(granularity, PerGroup):
+                        scale_dtype = torch.bfloat16 # KleidiAI kernel requires bfloat16 scale_dtype
```
metascroy, line 621: Only bfloat16 on PerGroup, but not on PerRow?

torchao/experimental/quant_api.py
```diff
@@ -531,13 +538,28 @@ def int8_dynamic_activation_intx_weight(
...
     """
-    try:
-        torch.ops.torchao._pack_8bit_act_4bit_weight
-    except AttributeError:
-        raise Exception(
-            "TorchAO experimental kernels are not loaded.  To install the kernels, run `USE_CPP=1 pip install .` from ao on a machine with an ARM CPU."
-            + "  Alternatively, use layout=PlainLayout() with int8_dynamic_activation_intx_weight, but note that doing so will result in much slower performance."
-        )
+
+    if target == "aten":
```
metascroy, line 542: Can this be something like:

```
if isinstance(layout, PackedLinearInt8DynamicActivationIntxWeightLayout):
    if taget == "aten":
        # Do KleidiAI checks
    
    if target == "native":
        # Do try/except import logic
```

metascroy:
Overall I think it's close.  Can you be sure to run `torchao/experimental/tests/test_packed_linear_int8_dynamic_activation_intx_weight_layout.py` to make sure the tests pass.  It is not currently enabled in CI.

You should also add some test cases for your new target to that file to check accuracy/exportability.

torchao/experimental/packed_linear_int8_dynamic_activation_intx_weight_layout.py
```diff
@@ -268,8 +327,82 @@ def _impl_2d(input_tensor, weight_tensor):
...
     return res
 
-
 register_aqt_quantized_linear_dispatch(
     _linear_check,
     _linear_impl,
 )
+
+
+class PackedLinearInt8DynamicActivationIntxWeightAtenTensor(AffineQuantizedTensor):
```
metascroy, line 336: Why do we need a subclass of AffineQuantizedTensor?  What is wrong with using the existing to_affine_quantized_intx method?

ng-05: This is suggested by one of the Pytorch reviewer: https://github.com/pytorch/ao/pull/1447#discussion_r1919517652
from_plain method needs specialization to accomodate bias. AffineQuantizedTensor provides a generic flow where specialization was possible but Jerry suggested to seprate the flow.

ng-05:
> Overall I think it's close. Can you be sure to run `torchao/experimental/tests/test_packed_linear_int8_dynamic_activation_intx_weight_layout.py` to make sure the tests pass. It is not currently enabled in CI.
> 
> You should also add some test cases for your new target to that file to check accuracy/exportability.
> 
> I'll be out for the week of January 20, but @digantdesai should be able to help with review during that time.

I can not run this test as the op build is broken for aarch64 linux.

torchao/experimental/quant_api.py
ng-05: Yes , thats the implementation detail for groupwise kernel. Channelwise( PerRow ) kernel can accept float32 scales

torchao/experimental/packed_linear_int8_dynamic_activation_intx_weight_layout.py
ng-05: We can detect bias propagation  from target=="aten" or from has_bias. I dont think we need both if we are making target part of PackedLinearInt8DynamicActivationIntxWeightLayout().
Where do you want to use ```has_bias``` and what is the usecase?

torchao/experimental/quant_api.py
ng-05: Refactored this

torchao/experimental/quant_api.py
ng-05: I have refactored the checks logic. It is now better encapsulated

ng-05:
@digantdesai @jerryzh168 can I please get your latest review on this. I have addressed all the review comments from @metascroy

torchao/experimental/packed_linear_int8_dynamic_activation_intx_weight_layout.py
digantdesai: Curious, (1) why adding bias as optional `to_affine_quantized_intx` isn't an option? or alternatively, (2) for native kernels we will add bias support sooner rather than later, so should we use this for both native and aten?

digantdesai:
> I can not run this test as the op build is broken for aarch64 linux.

Can you remind me of this issue? how can we resolve this?

digantdesai (commented):
I think with extended AQT it is close. Left some comments.

ng-05:
> > I can not run this test as the op build is broken for aarch64 linux.
> 
> Can you remind me of this issue? how can we resolve this?

We need to check for ```aarch64``` in cmake processor along with arm64. Even after that I am seeing few build issues. I did not have a deeper look at them.

torchao/experimental/packed_linear_int8_dynamic_activation_intx_weight_layout.py
```diff
@@ -125,9 +153,11 @@ def from_plain(
         scale: torch.Tensor,
         zero_point: Optional[torch.Tensor],
         layout: Layout,
+        bias: Optional[torch.Tensor] = None,
     ):
         assert isinstance(layout, PackedLinearInt8DynamicActivationIntxWeightLayout)
         assert layout.has_params_set(), "PackedLinearInt8DynamicActivationIntxWeightLayout params must be set before calling from_plain"
+        assert layout.target in {Target.NATIVE, Target.ATEN}, f"Unexpected target: {layout.target}"
```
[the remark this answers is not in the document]
ng-05: if bias is none and target layout.target != ATEN then execution will never reach this point. It will be executed via the "native" target which has bias assert.  Please check here: https://github.com/pytorch/ao/pull/1447/files#diff-3e4ffa192fe9921999dd6a798fc3fa620377896ef9ba65245b1e5ab8c0d7d344R593

torchao/experimental/packed_linear_int8_dynamic_activation_intx_weight_layout.py
```diff
@@ -31,17 +41,33 @@
 handler.setFormatter(formatter)
 logger.addHandler(handler)
 
+class Target(Enum):
+    """Enum that indicates the backend target"""
+
+    NATIVE = auto()
+    ATEN = auto()
```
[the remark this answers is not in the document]
ng-05: I decided to stick to aten so that more ops from aten can be added in future if needed. This enables this layout to work with torchao ops and aten ops.
```dyn_quant_matmul_4bit``` is not supposed to be KleidiAI specific only

ng-05:
@digantdesai 
for this comment :
Curious, (1) why adding bias as optional to_affine_quantized_intx isn't an option? or alternatively, (2) for native kernels we will add bias support sooner rather than later, so should we use this for both native and aten?

Please check 
https://github.com/pytorch/ao/pull/1447#discussion_r1922297356 and https://github.com/pytorch/ao/pull/1447#discussion_r1919517652

This change was suggested by @jerryzh168

digantdesai:
> @digantdesai for this comment : Curious, (1) why adding bias as optional to_affine_quantized_intx isn't an option? or alternatively, (2) for native kernels we will add bias support sooner rather than later, so should we use this for both native and aten?
> 
> Please check [#1447 (comment)](https://github.com/pytorch/ao/pull/1447#discussion_r1922297356) and [#1447 (comment)](https://github.com/pytorch/ao/pull/1447#discussion_r1919517652)
> 
> This change was suggested by @jerryzh168

What about (2)?

digantdesai:
I am ok with merging this for now, and since we will have to figure out details between native and aten for intx, we can resolve this later.

Stamping to unblock you.

torchao/experimental/quant_api.py
metascroy: I think different people have used the term symmetric differently.  Some people use it to mean the zero point is 0, but other people use it to mean the range is symmetric (e.g., -7 to +7).

It looks like you are using an asymmetric quant range of [-8, +7] with zeropoint being 0.  The quantizer should do this if you specify has_weight_zero=False (which does scale-only quantization and is tied to zero_point_domain being NONE) and weight_mappinMappingType.ASYMMETRIC (indicating the asymmetric range).

I introduced has_weight_zeros earlier because @kimishpatel did not like symmetric being synonymous with the zero point being 0.

metascroy:
> > Overall I think it's close. Can you be sure to run `torchao/experimental/tests/test_packed_linear_int8_dynamic_activation_intx_weight_layout.py` to make sure the tests pass. It is not currently enabled in CI.
> > You should also add some test cases for your new target to that file to check accuracy/exportability.
> > I'll be out for the week of January 20, but @digantdesai should be able to help with review during that time.
> 
> I can not run this test as the op build is broken for aarch64 linux.

I added CI to run torchao/experimental/tests/test_packed_linear_int8_dynamic_activation_intx_weight_layout.py so you don't have to do it locally (.github/workflows/torchao_experimental_test.yml).  Just make sure the "Run TorchAO Experimental Tests" job finishes successfully before merging.

I don't see any tests for your new quantization code added.  For now, can you add a new test file (e.g., torchao/experimental/tests/test_packed_linear_int8_dynamic_activation_intx_weight_layout_target_aten.py) that tests your quantizer/kernel code.  Later I can merge the existing test_packed_linear_int8_dynamic_activation_intx_weight_layout.py and the tests for aten target in the new file you create into one test. 

You can your test to CI by adding a line here: https://github.com/pytorch/ao/blob/main/.github/workflows/torchao_experimental_test.yml#L42

digantdesai:
Good point about tests, @ng-05 can we please add them? Also in the CI now?

ng-05:
> Good point about tests, @ng-05 can we please add them? Also in the CI now?

Done.

ng-05:
> > > Overall I think it's close. Can you be sure to run `torchao/experimental/tests/test_packed_linear_int8_dynamic_activation_intx_weight_layout.py` to make sure the tests pass. It is not currently enabled in CI.
> > > You should also add some test cases for your new target to that file to check accuracy/exportability.
> > > I'll be out for the week of January 20, but @digantdesai should be able to help with review during that time.
> > 
> > 
> > I can not run this test as the op build is broken for aarch64 linux.
> 
> I added CI to run torchao/experimental/tests/test_packed_linear_int8_dynamic_activation_intx_weight_layout.py so you don't have to do it locally (.github/workflows/torchao_experimental_test.yml). Just make sure the "Run TorchAO Experimental Tests" job finishes successfully before merging.
> 
> I don't see any tests for your new quantization code added. For now, can you add a new test file (e.g., torchao/experimental/tests/test_packed_linear_int8_dynamic_activation_intx_weight_layout_target_aten.py) that tests your quantizer/kernel code. Later I can merge the existing test_packed_linear_int8_dynamic_activation_intx_weight_layout.py and the tests for aten target in the new file you create into one test.
> 
> You can your test to CI by adding a line here: https://github.com/pytorch/ao/blob/main/.github/workflows/torchao_experimental_test.yml#L42

Thank you very much. I have done the requested changes around tests. Please review

torchao/experimental/packed_linear_int8_dynamic_activation_intx_weight_layout.py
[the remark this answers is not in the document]
digantdesai: But the aten implementation and the packing routine it selects is Kleidi specific.

PR #1449, opened 2024-12-19
Add gemlite kernel option to autoquant

jerryzh168:
Summary:
Refactored gemlite quant api a bit and added gemlite as an option to autoquant: `GEMLITE_INT4_AUTOQUANT_CLASS_LIST` and also `ALL_AUTOQUANT_CLASS_LIST`

Test Plan:
python generate.py --checkpoint_path $CHECKPOINT_PATH/$MODEL_REPO/model.pth --compile --compile_prefill --precision float16 --quantization autoquant-gemlite-int4

(gemlite only works for float16)
Reviewers:

Subscribers:

Tasks:

Tags:


torchao/dtypes/uintx/gemlite_layout.py
```diff
@@ -46,15 +46,14 @@ def get_gemlite_quant_kwargs(bit_width, group_size):
     return kwargs
 
 
-def apply_gemlite_quant(
+def get_gemlite_aqt_kwargs(
     weight,
     group_size=64,
     bit_width=4,
     packing_bitwidth=8,
     contiguous=None,
```
mobicham, line 54: this one is using `packing_bitwidth=8` by default, with the other functions are using 32. Ideally this one should be using 32 too

mobicham (commented):
Thank you Jerry! Overall looks ok to me, I haven't tested it though and I am not familar with autoquant api, so would make sense if Charles could look at it as well.

torchao/dtypes/uintx/gemlite_layout.py
jerryzh168: will change, looks like it should be OK

drisspg (approved):

PR #1455, opened 2024-12-20
[float8nocompile] Add alternate Triton kernels for FP8 conversion which use atomic_max-based algo instead of reduction-based algo

danielvegamyhre:
**Summary**

- Add new Triton kernels to use a different approach for calculating the global amax
- Have a single shared global amax tensor of size 1
- All blocks compute local amax then perform a thread-safe max operation `tl.atomic_max` with the shared global max tensor to find the true global max once the kernel is complete
- The advantage of this strategy is we don't have to allocate a shared buffer of size `num_elements // BLOCK_SIZE` outside of the kernels, which prevents us from autotuning the block size since we need to know it ahead of time to allocate the buffer.

Also made the algorithm used configurable and added unit tests for both.

**Test plan**
- `pytest test/test.py` - Passing
- `python3 benchmark/benchmark.py` yields large performance improvement as seen below. It's now markedly better than production path eager execution but not quite as good as compiled yet.

Performance of reduction based kernel with un-tuned block size of 8:
```
  input_size  high_precision_dtype      eager_time    compiled_time    float8nocompile
-------------  ----------------------  ------------  ---------------  -----------------
 65500         torch.float32                599.299          298.101    94446
 65500         torch.bfloat16               649.674          394.535    94386.3
     1.05e+06  torch.float32                640.5            332.171   104449
     1.05e+06  torch.bfloat16               685.822          421.365   104372
     1.68e+07  torch.float32               1963.09          1214.32    280825
     1.68e+07  torch.bfloat16              1828.16          1051.67    261710
     2.68e+08  torch.float32              24129.8          16287.2          3.39791e+06
     2.68e+08  torch.bfloat16             21603.2          12389.9          3.39515e+06
```

Performance of atomic max based kernel with auto-tuned block size:
```
  input_size  high_precision_dtype      eager_time    compiled_time    float8nocompile
------------  ----------------------  ------------  ---------------  -----------------
65500         torch.float32                599.055          298.543            372.018
65500         torch.bfloat16               649.523          394.457            413.137
    1.05e+06  torch.float32                640.497          332.419            413.503
    1.05e+06  torch.bfloat16               685.584          421.296            453.472
    1.68e+07  torch.float32               1963.72          1215.19            1415.5
    1.68e+07  torch.bfloat16              1829.28          1051.86            1298.55
    2.68e+08  torch.float32              24126.2          16294.3            19124.8
    2.68e+08  torch.bfloat16             21592.5          12390.6            16485.7
```

### DIFF
--- a/torchao/prototype/float8nocompile/kernels/fp8_dynamic_tensorwise.py
+++ b/torchao/prototype/float8nocompile/kernels/fp8_dynamic_tensorwise.py
@@ -9,7 +9,6 @@ Triton kernels for scaling high precision tensors to float8.
 """
 
 import torch
-
 import triton
 import triton.language as tl
 
@@ -30,11 +29,18 @@ FP8_DTYPE_MAP = {
     torch.float64: tl.float64,
 }
 
+kernel_configs = [
+    triton.Config({"BLOCK_SIZE": 128}, num_warps=1),
+    triton.Config({"BLOCK_SIZE": 256}, num_warps=2),
+    triton.Config({"BLOCK_SIZE": 512}, num_warps=4),
+]
 
+
+@triton.autotune(configs=kernel_configs, key=["input_size"])
 @triton.jit
 def _block_amax(
     input_ptr,
-    block_amaxes_ptr,
+    amax_ptr,
     num_elements,
     input_dtype: tl.constexpr,
     BLOCK_SIZE: tl.constexpr,
@@ -47,37 +53,15 @@ def _block_amax(
     block_mask = block_offs < num_elements
     vals = tl.load(input_ptr + block_offs, mask=block_mask).to(input_dtype)
     block_amax = tl.max(tl.abs(vals), axis=0)
-    tl.store(block_amaxes_ptr + block_id, block_amax)
+    tl.atomic_max(amax_ptr, block_amax)
 
 
-@triton.jit
-def _fp8_scale(
-    block_amaxes_ptr,
-    scale_out_ptr,
-    num_elements,
-    fp8_dtype_max,
-    BLOCK_SIZE: tl.constexpr,
-    EPS: tl.constexpr,
-):
-    # calculate global amax across all blocks
-    global_amax = tl.zeros([1], dtype=tl.float64)
-    num_blocks = tl.cdiv(num_elements, BLOCK_SIZE)
-    for i in range(num_blocks):
-        block_max = tl.load(block_amaxes_ptr + i)
-        global_amax = tl.maximum(global_amax, block_max)
-
-    # compute scale, must be fp32
-    scale = (fp8_dtype_max / tl.clamp(global_amax, min=EPS, max=float("inf"))).to(
-        tl.float32
-    )
-    scale_off = tl.arange(0, 1)
-    tl.store(scale_out_ptr + scale_off, scale)
-
-
+@triton.autotune(configs=kernel_configs, key=["input_size"])
 @triton.jit
 def _to_fp8(
     input_ptr,
-    scale_ptr,
+    scale_out_ptr,
+    amax_ptr,
     out_ptr,
     num_elements,
     fp8_dtype_min,
@@ -87,11 +71,18 @@ def _to_fp8(
     BLOCK_SIZE: tl.constexpr,
     EPS: tl.constexpr,
 ):
-    # load previously computed scale
-    scale = tl.load(scale_ptr)
+    # compute scale, must be fp32
+    global_amax = tl.load(amax_ptr)
+    scale = (fp8_dtype_max / tl.clamp(global_amax, min=EPS, max=float("inf"))).to(
+        tl.float32
+    )
+    # only one program needs to store the scale
+    block_id = tl.program_id(axis=0)
+    if block_id == 0:
+        scale_offs = tl.arange(0, 1)
+        tl.store(scale_out_ptr + scale_offs, scale)
 
     # load block of input tensor
-    block_id = tl.program_id(axis=0)
     block_start = block_id * BLOCK_SIZE
     block_offs = block_start + tl.arange(0, BLOCK_SIZE)
     mask = block_offs < num_elements
@@ -109,10 +100,7 @@ def triton_hp_tensor_to_float8_dynamic(
     linear_mm_config: LinearMMConfig,
     gemm_input_role: GemmInputRole = GemmInputRole.INPUT,
 ) -> Float8Tensor:
-
     assert hp_tensor.is_contiguous(), "tensor must be contiguous"
-
-    BLOCK_SIZE = 8  # TODO(danielvegamyhre): tune this for perf
 
     num_elements = hp_tensor.numel()
     orig_shape = hp_tensor.shape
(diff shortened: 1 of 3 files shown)


drisspg (commented):
Looks good. I'm not sure I would completely throw away the old though since I think tuning the split reduction could get us a lot of performance and is deterministic which is nice

danielvegamyhre:
> Looks good. I'm not sure I would completely throw away the old though since I think tuning the split reduction could get us a lot of performance and is deterministic which is nice

Hmm ok I will make the strategy used configurable.

danielvegamyhre:
> Looks good. I'm not sure I would completely throw away the old though since I think tuning the split reduction could get us a lot of performance and is deterministic which is nice

@drisspg I updated the PR to keep both implementation and make the kernel algorithm used configurable, and updated unit tests to exercise both paths.

One thing I noticed though is that the atomic max strategy was failing non-deterministically, I had to add a call ` torch.cuda.synchronize()` after the global amax kernel to fix this.

torchao/prototype/float8nocompile/kernels/fp8_dynamic_tensorwise.py
```diff
@@ -31,8 +31,86 @@
...
+    BLOCK_SIZE: tl.constexpr,
+    EPS: tl.constexpr,
+):
+    # compute local amax for each block
+    block_id = tl.program_id(axis=0)
+    block_start = block_id * BLOCK_SIZE
+    block_offs = block_start + tl.arange(0, BLOCK_SIZE)
+    block_mask = block_offs < num_elements
+    vals = tl.load(input_ptr + block_offs, mask=block_mask).to(input_dtype)
+    block_amax = tl.max(tl.abs(vals), axis=0)
```
vkuzo, line 68: why is there an `axis=0` here?

torchao/prototype/float8nocompile/kernels/fp8_dynamic_tensorwise.py
```diff
@@ -31,8 +31,86 @@
...
+):
+    # compute scale, must be fp32
+    global_amax = tl.load(amax_ptr)
+    scale = (fp8_dtype_max / tl.clamp(global_amax, min=EPS, max=float("inf"))).to(
+        tl.float32
+    )
+
+    # only one program needs to store the scale
+    block_id = tl.program_id(axis=0)
+    if block_id == 0:
```
vkuzo, line 95: not sure why this is needed, maybe delete since doing per-program custom logic increases complexity?

if you want to expose the scale to be able to test it numerically, IMO having your atomics kernel output the scale is simpler

torchao/prototype/float8nocompile/kernels/fp8_dynamic_tensorwise.py
```diff
@@ -126,47 +202,80 @@ def triton_hp_tensor_to_float8_dynamic(
...
-        fp8_output,
-        num_elements,
-        fp8_dtype_min,
-        fp8_dtype_max,
-        input_dtype=tl_input_dtype,
-        output_dtype=tl_output_dtype,
-        BLOCK_SIZE=BLOCK_SIZE,
-        EPS=EPS,
-    )
+        torch.cuda.synchronize()
```
vkuzo, line 223: why is this needed?

if helpful, I previously experimented with atomic max here (https://github.com/vkuzo/pytorch_scripts/tree/main/reduction_hack) and didn't need synchronization for correct numerics, although I didn't test the logic all the way up to the float8 conversion

vkuzo:
> Performance of atomic max based kernel with auto-tuned block size:

A few comments about measurement:
1. feel free to only report bfloat16 to keep things simple
2. for `input_size`, it would be good to report the 3d tensor shape instead of the 1d number of elements, to make it easy for the reader to map back to real tensor shapes in common models

torchao/prototype/float8nocompile/kernels/fp8_dynamic_tensorwise.py
danielvegamyhre: The unit test for conversion was failing non-deterministically, and forcing synchronization here was the only solution that reliably fixed it. This isn't ideal though so am hoping to find a better way.

danielvegamyhre:
> > Performance of atomic max based kernel with auto-tuned block size:
> 
> A few comments about measurement:
> 
> 1. feel free to only report bfloat16 to keep things simple
> 2. for `input_size`, it would be good to report the 3d tensor shape instead of the 1d number of elements, to make it easy for the reader to map back to real tensor shapes in common models

Makes sense, done! New example output:

```
input_shape    high_precision_dtype      eager_time    compiled_time    float8nocompile
-------------  ----------------------  ------------  ---------------  -----------------
(16, 4096)     torch.bfloat16               649.843          395.745            412.926
(256, 4096)    torch.bfloat16               685.372          421.237            449.762
(4096, 4096)   torch.bfloat16              1828.54          1053.8             1244.41
(65536, 4096)  torch.bfloat16             21576            12438.9            15619.6
```

torchao/prototype/float8nocompile/kernels/fp8_dynamic_tensorwise.py
danielvegamyhre: It isn't necessary, i've removed it.

torchao/prototype/float8nocompile/kernels/fp8_dynamic_tensorwise.py
danielvegamyhre: >not sure why this is needed, maybe delete since doing per-program custom logic increases complexity?

Since all N thread blocks / program ids will calculate the same scale value for the tensor, we only need 1 of them to store the scale output. I figured having all N programs write the same value to the same memory location would be redundant. 

>if you want to expose the scale to be able to test it numerically, IMO having your atomics kernel output the scale is simpler

I need the scale to use in the Float8Tensor constructor, but yeah I agree it would be simpler to have the atomic amax kernel do the scale calculation right there rather than storing the amax then using it to calculate the scale here. I will refactor.

vkuzo: > I need the scale to use in the Float8Tensor constructor

that's a good reason, just put that in the comment

> I figured having all N programs write the same value to the same memory location would be redundant.

I think having a triton program have if statements which branch by program ID is a pattern to avoided (even if it works) because it's suprising / isn't really how people commonly write triton kernels.  IMO we can just refactor your atomic kernel code to also write out the scale to get to the same place with simpler code.

danielvegamyhre: > > I need the scale to use in the Float8Tensor constructor
> 
> that's a good reason, just put that in the comment

Done

> > I figured having all N programs write the same value to the same memory location would be redundant.
> 
> I think having a triton program have if statements which branch by program ID is a pattern to avoided (even if it works) because it's suprising / isn't really how people commonly write triton kernels. IMO we can just refactor your atomic kernel code to also write out the scale to get to the same place with simpler code.

Ok good to know, thanks - refactored to remove this pattern and simplify the code, let me know what you think.

vkuzo (approved):

Issue #556, opened 2024-07-30
low precision training upcoming feature tracker


vkuzo:
This is a running list of planned features for low precision training.  As features are completed we plan to delete them from this list, to keep things simple.

# MX

## pytorch/pytorch

* [in progress] improving torch.compile performance for to_mx cast across dim1: https://github.com/pytorch/pytorch/issues/149982

## pytorch/torchao

* [in progress] performance tracker: https://github.com/pytorch/ao/issues/1768

# float8

## performance

* [in progress] optimize torch.compile performance for float8 tensorwise scaling/casting kernels
  * [fixed behind a flag, off by default] https://github.com/pytorch/pytorch/issues/130015
  * [planned] https://github.com/pytorch/pytorch/issues/128063
* [in progress] ensure that float8 rowwise scaling is performant with TP and async TP https://github.com/pytorch/pytorch/issues/149990

## distributed

* [planned] verify integration with PP

## new features

* [planned] MoE support: https://github.com/pytorch/ao/issues/2147
* [planned] float8 SDPA (priority TBD)

## ecosystem

* [in progress] enable TP in torchtune's float8 integration: https://github.com/pytorch/torchtune/issues/2682

## other

* [planned] expose float8 training via the quantize_ API
* [planned] migrate `torchao.float8` code to `torchao.quantization` for better unification with the rest of torchao, in a BC-preserving way: https://github.com/pytorch/ao/issues/894

airMeng:
Has SDPA supported FP8 already like flash-attention https://github.com/Dao-AILab/flash-attention/blob/main/hopper/benchmark_flash_attention_fp8.py does?

vkuzo:
> Has SDPA supported FP8 already like flash-attention https://github.com/Dao-AILab/flash-attention/blob/main/hopper/benchmark_flash_attention_fp8.py does?

we would be interesting in accepting a contribution for this!  To me, the priority is not clear as we have not seen convincing evidence that SDPA + float8 can achieve good accuracy numerically.

avizon-aws:
@vkuzo , is there a plan for MXFP8 all-gather? If so, by when is the feature expected to be enabled?

vkuzo:
> [@vkuzo](https://github.com/vkuzo) , is there a plan for MXFP8 all-gather? If so, by when is the feature expected to be enabled?

we have considered this for float8 and mx, but it has not made the priority cut yet.  Could you share some context about your use case, and whether you need this for FSDP or TP?

PR #1469, opened 2025-01-02
Add a register_replacement to fix float8 delayed scaling kernel fusion issues in torchao/float8

y-sq:
**Summary**:
The original pr in inductor, https://github.com/pytorch/pytorch/pull/143464

As suggested in the comments, we moved the patterns to torchao repo.
Note that it needs the pytorch repo to contain this pr https://github.com/pytorch/pytorch/pull/139321 to work.

**Test Plans**:
Run float8 training script. Amax and cast are fused in delayed scaling; dynamic scaling is not affected. Also tested with preceding ops.

The delayed scaling kernel also looks reasonable to me, [https://fburl.com/phabricator/iqmlollk](https://l.facebook.com/l.php?u=https%3A%2F%2Ffburl.com%2Fphabricator%2Fiqmlollk&h=AT1D_5oMCX-aALM9tYctFwC6Xh5Ef-YSPYyrw65XrE09Xd76JLfL9uo2WnyA3uwcNcx9_WppKaXXBkzLnc-UUG_mO4j0mtOaruGcjCbHyuVhGvYEZkOF980gvkM7kLVeD24o0QLCVpqh0c32MAch47lMTKA)

A simple performance test: [D67517255](https://www.internalfb.com/diff/D67517255)
(It needs some temporary changes to force recompute weight and activation in backward.)

- I used the "recompute weight + activation in backward" case to test, as it will likely become the default choice. 
- "recompute weight" is enabled by "force_recompute_fp8_weight_in_bwd=enable_activation_checkpointing". It might be later changed to other options. So I made the changes in a local testing scripts for now.

```
TORCH_LOGS="fusion" TORCHINDUCTOR_LOOP_ORDERING_AFTER_FUSION=1  buck run @mode/opt scripts/shuqiyang/test_inductor:test_float8 --  ~/local/tmp/20241120_test --dtype_filter float8 --scaling_type_input delayed --scaling_type_weight delayed --scaling_type_grad_output delayed --enable_activation_checkpointing True  2>&1 | tee ~/test_compile_b.txt
```

Linear - 4096x4096x4096
- Before: `0.7000332674418591` ms
- After: `0.6707110309278329` ms
- Delta: `4.4%`

With layer_norm:
- Before: `0.753488252631584`
- After: `0.7374823118279574`
- Delta: `2.2%`

Witi sigmoid:
- Before: `0.7153208260869579`
- After: `0.6845765714285723`
- Delta: `4.5%`

Differential Revision: D67758184

### DIFF
--- a/torchao/float8/README.md
+++ b/torchao/float8/README.md
@@ -82,6 +82,9 @@ from torchao.utils import TORCH_VERSION_AT_LEAST_2_5
 if not TORCH_VERSION_AT_LEAST_2_5:
     raise AssertionError("torchao.float8 requires PyTorch version 2.5 or greater")
 
+# Recommended: enable additional torchinductor passes to improve the performance of delayed scaling
+torchao.float8._prototype_register_float8_delayed_scaling_inductor_passes()
+
 # create model and sample input
 m = nn.Sequential(
     nn.Linear(2048, 4096),
@@ -172,7 +175,7 @@ For small shapes, a combination of (2) and (3) leads to speedup < 1.  For medium
 
 ## Scaling type vs speedup
 
-Delayed scaling is theoretically faster than dynamic scaling because of reduced read/write traffic requirements.  Today, torch.compile has a couple of limitations (see the performance section of https://github.com/pytorch/ao/issues/556) which prevent us from reaching the optimal behavior for delayed scaling, so the observed performance of delayed scaling is close to that of dynamic scaling. As the torch.compile limitations are fixed, we expect delayed scaling to eventually become more performant compared to dynamic scaling.
+Delayed scaling is theoretically faster than dynamic scaling because of reduced read/write traffic requirements. Today, torch.compile has a couple of limitations (see the performance section of https://github.com/pytorch/ao/issues/556) which prevent us from reaching the optimal behavior for delayed scaling without workarounds.  We have a prototype workaround (API subject to change) with the `torchao.float8._prototype_register_float8_delayed_scaling_inductor_passes()` API to improve delayed scaling performance.
 
 ## torch.compile behavior vs speedup
 
--- a/torchao/float8/inductor_utils.py
+++ b/torchao/float8/inductor_utils.py
@@ -1,0 +1,126 @@
+import functools
+import inspect
+import traceback
+from collections import deque
+
+import torch
+
+
+def amax_with_scaling_pattern(tensor_x_inp, scale_x, fp8_dtype, fp8_max):
+    tensor_x = tensor_x_inp.to(torch.float32) * scale_x
+    tensor_x = tensor_x.clamp(min=-1 * fp8_max, max=fp8_max)
+    tensor_x = tensor_x.to(fp8_dtype)
+    amax = torch.max(torch.abs(tensor_x_inp))
+    return (tensor_x, amax)
+
+
+def amax_with_scaling_tiled_replacement(tensor_x_inp, scale_x, fp8_dtype, fp8_max):
+    tensor_x = tensor_x_inp.to(torch.float32) * scale_x
+    tensor_x = tensor_x.clamp(min=-1 * fp8_max, max=fp8_max)
+    tensor_x = tensor_x.to(fp8_dtype)
+    amax_1 = torch.max(torch.abs(tensor_x_inp), dim=-1).values
+    amax = torch.max(amax_1)
+    return (tensor_x, amax)
+
+
+# The amax_with_scaling_pattern will also match dynamic scaling cases, we want to avoid that.
+# `scale_x` of delayed scaling comes from the previous iteration, instead of from `tensor_x_inp`.
+# We check that `scale_x` is not a dependency of `tensor_x_inp`
+def fp8_delayed_scaling_extra_check(match):
+    scale_x_inputs = deque([match.kwargs["scale_x"]])
+    max_num_node_to_check = 20  # Don't traverse too many nodes
+    current_num_node = 0
+    while len(scale_x_inputs) > 0 and current_num_node < max_num_node_to_check:
+        current_node = scale_x_inputs.popleft()
+        for n in current_node.all_input_nodes:
+            if n == match.kwargs["tensor_x_inp"]:
+                return False
+            scale_x_inputs.append(n)
+            current_num_node += 1
+    return True
+
+
+def partialize_and_update_signature(func, **kwargs):
+    """
+    Equivalent to functools.partial but also updates the signature on returned function
+    """
+    original_sig = inspect.signature(func)
+    parameters = original_sig.parameters
+
+    new_parameters = {
+        key: value for key, value in parameters.items() if key not in kwargs
+    }
+    new_sig = inspect.Signature(parameters=list(new_parameters.values()))
+
+    partial_func = functools.partial(func, **kwargs)
+
+    def wrapper(*args, **kwargs):
+        return partial_func(*args, **kwargs)
+
+    wrapper.__signature__ = new_sig  # type: ignore[attr-defined]
+    wrapper.__name__ = func.__name__
+
+    return wrapper
+
+
+def register_fp8_delayed_scaling_patterns_inner():
+    from torch._inductor.fx_passes.post_grad import (
+        pass_patterns as post_grad_patterns_all,
+    )
+    from torch._inductor.pattern_matcher import fwd_only, register_replacement
+
+    post_grad_patterns = post_grad_patterns_all[1]  # medium priority
+
+    if torch.cuda.is_available():
+        for fp8_dtype in [
+            torch.float8_e4m3fn,
+            torch.float8_e5m2,
+            torch.float8_e4m3fnuz,
+            torch.float8_e5m2fnuz,
+        ]:
+            # torch.float16 has the same pattern as torch.bfloat16, because they both needs `tensor_x_inp.to(torch.float32)`
+            for dtype in [torch.float32, torch.bfloat16]:
+                device = "cuda"
+                register_replacement(
+                    partialize_and_update_signature(
+                        amax_with_scaling_pattern,
+                        fp8_dtype=fp8_dtype,
+                        fp8_max=torch.finfo(fp8_dtype).max,
+                    ),
+                    partialize_and_update_signature(
+                        amax_with_scaling_tiled_replacement,
+                        fp8_dtype=fp8_dtype,
+                        fp8_max=torch.finfo(fp8_dtype).max,
+                    ),
+                    [
+                        torch.tensor((16, 16), device=device, dtype=dtype),
+                        torch.tensor(2.0, device=device, dtype=torch.float32),
+                    ],
+                    fwd_only,
+                    post_grad_patterns,
+                    extra_check=fp8_delayed_scaling_extra_check,
+                )
+
+
+"""
+This a short-term workaround of the delayed scaling performance issue.
+It explicitly replaces `max(x)` with `max(max(x, dim=-1))`, enabling the fusion of amax scaling factor calculation and fp8 casting.
+
+Usage:
+    To use this solution, add the following line at the beginning of your user code:
+    torchao.float8._prototype_register_float8_delayed_scaling_inductor_passes()
+"""
+
+
+def _prototype_register_float8_delayed_scaling_inductor_passes() -> None:
+    # To make the fp8 delayed scaling pattern work, we need a fix pr from inductor, https://github.com/pytorch/pytorch/pull/139321
+    # Will throw the error if the pattern registration did not work, up to user to decide what to do with it
+    try:
+        register_fp8_delayed_scaling_patterns_inner()
+    except AssertionError as e:
+        if "assert pattern_repr not in _seen_patterns" in traceback.format_exc():
+            print(
+                f"Caught duplicated patterns in register_fp8_delayed_scaling_patterns: {traceback.format_exc()}",
+                "\nPlease update your pytorch dependency to the latest main branch to fix it.\n",
+            )
+        raise e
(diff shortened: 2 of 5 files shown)


y-sq:
The failed tests in `Run Regression Tests / test-nightly (CUDA Nightly, linux.g5.12xlarge.nvidia.gpu, --pre torch` are from non-float8 folders, seems unrelated to this pr?

Screenshot 2025-01-02 at 3 45 24 PM

torchao/float8/stateful_float8_linear.py
```diff
@@ -36,6 +36,9 @@
     WeightWithDynamicFloat8CastTensor,
     WeightWithStaticFloat8CastTensor,
 )
+from torchao.float8.inductor_utils import register_fp8_delayed_scaling_patterns
+
+register_fp8_delayed_scaling_patterns()
```
vkuzo, line 41: I would prefer that we have an explicit API to enable this instead of unexpectedly registering a call on module import.  How about something like below?

```
def _prototype_register_float8_delayed_scaling_inductor_passes():
    """
    Note: this is a prototype API and is subject to change.
    TODO writeme
    """
    ...
```

Then, we can mention this in the delayed scaling section of `README.md`

torchao/float8/inductor_utils.py
```diff
@@ -0,0 +1,128 @@
...
+    wrapper.has_run = False
+    wrapper.result = None
+    return wrapper
+
+
+@run_once
+def register_fp8_delayed_scaling_patterns() -> bool:
+    # To make the fp8 delayed scaling pattern work, we need a fix pr from inductor, https://github.com/pytorch/pytorch/pull/139321
+    # Added the try-catch block to ignore the failed pattern if the current torch verions doesn't include the fix pr.
+    try:
```
vkuzo, line 124: can this check for something specific instead of a blanket try-catch?

torchao/float8/inductor_utils.py
```diff
@@ -0,0 +1,128 @@
...
+                        torch.tensor((16, 16), device=device, dtype=dtype),
+                        torch.tensor(2.0, device=device, dtype=torch.float32),
+                    ],
+                    fwd_only,
+                    post_grad_patterns,
+                    extra_check=fp8_delayed_scaling_extra_check,
+                )
+
+
+def run_once(f):
```
vkuzo, line 106: can delete this if you change the API to a function user has to call

torchao/float8/inductor_utils.py
```diff
@@ -0,0 +1,128 @@
+import functools
+import inspect
+from collections import deque
+
+import torch
+
+E4M3_MAX_POS = torch.finfo(torch.float8_e4m3fn).max
+E5M2_MAX_POS = torch.finfo(torch.float8_e5m2).max
```
vkuzo, line 8: can we also handle `torch.float8_e4m3fnuz` and `torch.float8_e5m2fnuz`

torchao/float8/inductor_utils.py
```diff
@@ -0,0 +1,128 @@
...
+        else:
+            return wrapper.result
+
+    wrapper.has_run = False
+    wrapper.result = None
+    return wrapper
+
+
+@run_once
+def register_fp8_delayed_scaling_patterns() -> bool:
```
vkuzo, line 121: can we add a docblock describing that this replaces `max(x)` with `max(max(x, dim=-1))` for cases relevant to delayed scaling

torchao/float8/inductor_utils.py
y-sq: I modified to explicitly log out the error and ask the user to update pytorch version if the duplicate pattern error is caught. Calling register_fp8_delayed_scaling_patterns without the inductor fix pr will now get:
```
Caught duplicated patterns in register_fp8_delayed_scaling_patterns: Traceback (most recent call last):
  File "/data/users/shuqiyang/fbsource/buck-out/v2/gen/fbcode/d22fb8d26889267e/scripts/shuqiyang/test_inductor/__test_float8__/test_float8#link-tree/torchao/float8/inductor_utils.py", line 100, in register_fp8_delayed_scaling_patterns
    register_fp8_delayed_scaling_patterns_inner()
  File "/data/users/shuqiyang/fbsource/buck-out/v2/gen/fbcode/d22fb8d26889267e/scripts/shuqiyang/test_inductor/__test_float8__/test_float8#link-tree/torchao/float8/inductor_utils.py", line 79, in register_fp8_delayed_scaling_patterns_inner
    register_replacement(
  File "/data/users/shuqiyang/fbsource/buck-out/v2/gen/fbcode/d22fb8d26889267e/scripts/shuqiyang/test_inductor/__test_float8__/test_float8#link-tree/torch/_inductor/pattern_matcher.py", line 1398, in register_replacement
    assert pattern_repr not in _seen_patterns
AssertionError
 
Please update your pytorch dependency to the latest main branch to fix it.
```

y-sq:
@vkuzo thanks for your comments, I update the pr to use an explicit API; add readme and docblock; add the two additional fp8 dtypes; add explicit check and hint for update pytorch version for the inductor fix pr.

y-sq:
With the pattern replacement:
```
TORCHINDUCTOR_LOOP_ORDERING_AFTER_FUSION=1 TORCH_LOGS="fusion"  python benchmarks/float8/profile_linear_float8.py ~/tmp/log_delayed_scaling_3 --dtype_filter float8 --scaling_type_input delayed --scaling_type_weight delayed --scaling_type_grad_output delayed --enable_activation_checkpointing True --add_inductor_metadata_to_trace False --enable_float8_delayed_scaling_inductor_passes True
```
```
Summary of time (ms) by kernel category

 experiment     1_float8
category               
0_gemm            0.353
1_f8_overhead     0.161
2_other           0.028
All               0.542
```

Baseline:
```
--enable_float8_delayed_scaling_inductor_passes True
```
```
Summary of time (ms) by kernel category

 experiment     1_float8
category               
0_gemm            0.349
1_f8_overhead     0.198
2_other           0.022
All               0.570
```

We can get 5% update by `--enable_float8_delayed_scaling_inductor_passes True`.

torchao/float8/README.md
```diff
@@ -174,6 +174,12 @@ For small shapes, a combination of (2) and (3) leads to speedup < 1.  For medium
 
 Delayed scaling is theoretically faster than dynamic scaling because of reduced read/write traffic requirements.  Today, torch.compile has a couple of limitations (see the performance section of https://github.com/pytorch/ao/issues/556) which prevent us from reaching the optimal behavior for delayed scaling, so the observed performance of delayed scaling is close to that of dynamic scaling. As the torch.compile limitations are fixed, we expect delayed scaling to eventually become more performant compared to dynamic scaling.
 
+As a short-term workaround of the delayed scaling performance issue, we implemented `_prototype_register_float8_delayed_scaling_inductor_passes`, which explicitly replaces `max(x)` with `max(max(x, dim=-1))`, enabling the fusion of amax scaling factor calculation and fp8 casting.
+To use this solution, simply add the following line at the beginning of your user code:
+```
+torchao.float8._prototype_register_float8_delayed_scaling_inductor_passes()
```
vkuzo, line 180: can you add this directly to the code sample for delayed scaling, and add a comment there that this API is subject to change?

torchao/float8/README.md
```diff
@@ -174,6 +174,12 @@ For small shapes, a combination of (2) and (3) leads to speedup < 1.  For medium
 
 Delayed scaling is theoretically faster than dynamic scaling because of reduced read/write traffic requirements.  Today, torch.compile has a couple of limitations (see the performance section of https://github.com/pytorch/ao/issues/556) which prevent us from reaching the optimal behavior for delayed scaling, so the observed performance of delayed scaling is close to that of dynamic scaling. As the torch.compile limitations are fixed, we expect delayed scaling to eventually become more performant compared to dynamic scaling.
```
vkuzo, line 175: let's change this section to something like

Delayed scaling is theoretically faster than dynamic scaling because of reduced read/write traffic requirements. Today, torch.compile has a couple of limitations (see the performance section of https://github.com/pytorch/ao/issues/556) which prevent us from reaching the optimal behavior for delayed scaling without workarounds.  We have a prototype workaround (API subject to change) with the `torchao.float8._prototype_register_float8_delayed_scaling_inductor_passes()` API to improve delayed scaling performance.

torchao/float8/README.md
y-sq: Yes, I'll update the pr to move this part to the delayed scaling code samples.

torchao/float8/inductor_utils.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def _prototype_register_float8_delayed_scaling_inductor_passes() -> bool:
+    # To make the fp8 delayed scaling pattern work, we need a fix pr from inductor, https://github.com/pytorch/pytorch/pull/139321
+    try:
+        register_fp8_delayed_scaling_patterns_inner()
+        return True
+    except Exception as e:
+        if (
+            isinstance(e, AssertionError)
```
vkuzo, line 122: if this is the error you expect from a too-old PyTorch version, just catch `AssertionError` directly?

torchao/float8/inductor_utils.py
```diff
@@ -0,0 +1,133 @@
...
+            and "assert pattern_repr not in _seen_patterns" in traceback.format_exc()
+        ):
+            print(
+                f"Caught duplicated patterns in register_fp8_delayed_scaling_patterns: {traceback.format_exc()}",
+                "\nPlease update your pytorch dependency to the latest main branch to fix it.\n",
+            )
+        else:
+            print(
```
vkuzo, line 130: I think it's better to not catch this and let the exception bubble up

y-sq: makes sense, I'll change this part, and leave a comment here to explain that if the user sees an assertion error, they can try to update the pytorch version to fix?

torchao/float8/inductor_utils.py
eellison: You can also check if _seen_patterns is defined, and if it is clear, remove the last entry after every registration ?

torchao/float8/inductor_utils.py
y-sq: @vkuzo thanks for the comments. I updated the readme and exception try-catch parts.

torchao/float8/inductor_utils.py
y-sq: Hi @eellison , thanks for the comment. Do we expect the pattern matcher pr to merge soon? If so, I think we can rely on that and just use the latest pytorch branch; otherwise, I can "remove the last entry after every registration" to workaround the duplicate pattern issue.

eellison: it landed - hopefully it doesnt get reverted. there was an internal import of `_seen_patterns` that failed when I removed it. leaving that in for now so should stick this time.

torchao/float8/inductor_utils.py
```diff
@@ -0,0 +1,126 @@
...
+    try:
+        register_fp8_delayed_scaling_patterns_inner()
+        return True
+    except AssertionError:
+        if "assert pattern_repr not in _seen_patterns" in traceback.format_exc():
+            print(
+                f"Caught duplicated patterns in register_fp8_delayed_scaling_patterns: {traceback.format_exc()}",
+                "\nPlease update your pytorch dependency to the latest main branch to fix it.\n",
+            )
+        return False
```
vkuzo, line 126: can we re-throw the exception instead of swallowing it, and make the function void instead of returning a boolean status?  I think a clean signature for this function would be:

```
# throws an exception if the registration did not work, up to user to decide what to do with it
def _prototype_register_float8_delayed_scaling_inductor_passes() -> None: ...
```

torchao/float8/inductor_utils.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+def _prototype_register_float8_delayed_scaling_inductor_passes() -> bool:
+    # To make the fp8 delayed scaling pattern work, we need a fix pr from inductor, https://github.com/pytorch/pytorch/pull/139321
+    try:
+        register_fp8_delayed_scaling_patterns_inner()
+        return True
+    except AssertionError:
+        if "assert pattern_repr not in _seen_patterns" in traceback.format_exc():
+            print(
```
vkuzo, line 122: this is good, but you can still re-throw the exception after printing this nice hint on how to solve it

torchao/float8/README.md
```diff
@@ -82,6 +82,10 @@ from torchao.utils import TORCH_VERSION_AT_LEAST_2_5
 if not TORCH_VERSION_AT_LEAST_2_5:
     raise AssertionError("torchao.float8 requires PyTorch version 2.5 or greater")
 
+# [Optional] Enable a short-term workaround of the delayed scaling performance issue.
```
vkuzo, line 85: how about:

```
# Recommended: enable additional torchinductor passes to improve the performance of delayed scaling
```

torchao/float8/README.md
```diff
@@ -82,6 +82,10 @@ from torchao.utils import TORCH_VERSION_AT_LEAST_2_5
...
+# Explicitly replaces `max(x)` with `max(max(x, dim=-1))`, enabling the fusion of amax scaling factor calculation and fp8 casting.
```
vkuzo, line 86: nit: just remove this line? It will probably confuse end users.

vkuzo:
looks good! Final request: can we ensure the new functionality is covered with a test, just as a smoke check that this does not affect numerics?  We'll need to ensure that the passes can be easily unregistered when the test is done to prevent affecting other tests.

eellison:
should we land ?

y-sq:
> should we land ?

yes, I'll add a test tomorrow and will land the pr after getting a review approval

eellison (approved):

y-sq:
Updates: Added the test case.

PR #1471, opened 2025-01-02
add fp8 conversion kernel for transpose in row major format

danielvegamyhre:
* #1488
* #1475
* #1474
* #1473
* #1472
* __->__ #1471

### DIFF
--- a/torchao/prototype/float8nocompile/kernels/fp8_dynamic_tensorwise.py
+++ b/torchao/prototype/float8nocompile/kernels/fp8_dynamic_tensorwise.py
@@ -93,6 +93,67 @@ def _to_fp8_row_major(
     key=["num_elements"],
 )
 @triton.jit
+def _to_fp8_row_major_t(
+    input_ptr,
+    out_ptr,
+    scale_ptr,
+    num_elements: int,
+    fp8_dtype_min: float,
+    fp8_dtype_max: float,
+    input_num_rows: int,
+    input_num_cols: int,
+    output_num_rows: int,
+    output_num_cols: int,
+    input_stride_row: int,
+    input_stride_col: int,
+    output_stride_row: int,
+    output_stride_col: int,
+    input_dtype: tl.constexpr,
+    output_dtype: tl.constexpr,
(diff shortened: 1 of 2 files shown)


drisspg (commented):
Can we not just use the same kernel and pass in the strides of the output tensor we will be filling?

danielvegamyhre:
> Can we not just use the same kernel and pass in the strides of the output tensor we will be filling?

I don't think so, but I could be wrong. 

In the non-transposed row major output, the input and output dims and strides are the same, so the input column offsets correspond to output column strides, and input row offsets correspond to output row strides:

```python
    out_offs = (
        block_col_offs[None, :] * out_stride_col              
        + block_row_offs[:, None] * out_stride_row
    )
```

However, with the transpose written in row major format, the input/output dims and strides are not the same, and the columns need to be written as rows, and vice versa. So we need to use output row strides with the column offsets, and output column strides with the row offsets.

```python
    out_offs = (
        block_col_offs[:, None] * output_stride_row       # input column offsets use output row stride
        + block_row_offs[None, :] * output_stride_col     # input row offsets use output column stride
    )
```

What do you think? Am I missing something?

drisspg:
Ohhh I meant can you just add a tl.contexpr flag gating whether we do the transpose or not?

danielvegamyhre:
> Ohhh I meant can you just add a tl.contexpr flag gating whether we do the transpose or not?

Oh I see, yeah during the development process when I was actively figuring out what logical differences would be needed between transpose vs non-transpose versions, it was easier for me to keep things cleanly separated. I'm happy to unify them now though, if you think reducing the API surface would be preferable.

drisspg:
Up to you, I think it might just be easier to maintain and it functionally equivalent

danielvegamyhre:
> Up to you, I think it might just be easier to maintain and it functionally equivalent

Makes sense. Would you mind if I do the refactor as a follow-up PR in order to avoid the merge conflicts with all the PRs higher in the stack?

vkuzo (approved):

PR #1481, opened 2025-01-02
[cleanup][4/x] unify weight casting

vkuzo:
Summary:

Not ready for review yet, performance regression because tensorwise
abs+max and weight casting is happening twice between fwd and bwd.
Limitation of something in PT2 stack?

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo:
* (to be filled)

vkuzo:
`config.force_recompute_fp8_weight_in_bwd` doesn't work with FSDP2 with float8 all-gather properly yet, looking

drisspg (approved):

PR #1486, opened 2025-01-03
Additional fixes for autoquant serialization

jerryzh168:
Summary:
Can't overwrite `.layout` attribute of a Tensor since `tensor.layout` should be `torch.layout`

discovered when loading the autoquantized model https://huggingface.co/jerryzh168/llama3-8b-autoquant

Test Plan:
tested locally with huggingface transformers
Reviewers:

Subscribers:

Tasks:

Tags:


drisspg (approved):
L

PR #1512, opened 2025-01-07
add handling for batch dim in float8nocompile

danielvegamyhre:
This PR includes changes for torchtitan + float8nocompile compatibility.

**Summary**
- Add handling for input tensors with 3+ dimensions, which is important for practical use with input tenors containing a batch dimension.
- Allow the Float8LinearConfig to be passed in as a parameter in convert_to_float8_nocompile_training() rather than hardcoded as the default, for maintain parity in torchtitan between float8 and float8nocompile code paths, ensuring they both use the same config and thus tests can be compared 1:1 (see [here](https://github.com/pytorch/torchtitan/blob/90567fc9827ffdf17bdd0349cd5276c662d0769a/torchtitan/float8.py#L98)) 
- Adds additional unit tests for comparing the prod vs prototype float8 conversion autograd functions


danielvegamyhre:
* (to be filled)

torchao/prototype/float8nocompile/float8nocompile_linear_test.py
```diff
@@ -0,0 +1,73 @@
...
+    # prototype forward. expects non-transposed weight
+    out_prototype = matmul_with_args_in_hp.apply(y_input_bf16, y_weight_bf16, config, linear_mm_config, KernelAlgorithm.ATOMIC_MAX)
+
+    # compare
+    assert torch.allclose(out_prod, out_prototype, atol=1e-3, rtol=1e-3)
+
+    out_prod.sum().backward()
+    out_prototype.sum().backward()
+
+    assert torch.allclose(x_input_bf16.grad, y_input_bf16.grad, atol=1e-3, rtol=1e-3)
```
vkuzo, line 72: can we compare to production float8 with bitwise equality instead?  It would be a more useful comparison with a higher chance to catch issues.

drisspg: I would be surprised if the amax version is bitwise identical since one is doing atomics and one is doing split reduction. Even the split reduction case has a chance to be flakey incase any of the split reduction params change ( which is internal to inductor)

vkuzo: yeah, I guess the amax + scaling should be bitwise equivalent but the gemm may not be.  I would still compare between the two float8 versions with a narrow tolerance instead of comparing between bf16 and float8 with a wide tolerance.

danielvegamyhre:
> > Allow the Float8LinearConfig to be passed in as a parameter in convert_to_float8_nocompile_training() rather than hardcoded as the default.
> 
> maybe explain why this is needed?

Updated description with reasoning for this

torchao/prototype/float8nocompile/float8nocompile_linear_test.py
danielvegamyhre: Changed these checks to verify bitwise equality and they are passing.

vkuzo:
> Allow the Float8LinearConfig to be passed in as a parameter in convert_to_float8_nocompile_training() rather than hardcoded as the default, for maintain parity in torchtitan between float8 and float8nocompile code paths, ensuring they both use the same config and thus tests can be compared 1:1 (see [here](https://github.com/pytorch/torchtitan/blob/90567fc9827ffdf17bdd0349cd5276c662d0769a/torchtitan/float8.py#L98))

well, if we add the config, we'll have to support every option? :)  Given that this is a prototype, doesn't matter that much.  The proper way to do it would be to either assert on unsupported config options or support every option.

torchao/prototype/float8nocompile/float8nocompile_linear.py
```diff
@@ -80,15 +80,19 @@ def forward(self, input: torch.Tensor) -> torch.Tensor:
         return output
 
     @classmethod
-    def from_float(cls, mod, kernel_algo: KernelAlgorithm = KernelAlgorithm.ATOMIC_MAX):
+    def from_float(
+        cls,
+        mod,
+        config: Float8LinearConfig,
```
vkuzo, line 86: add comment specifying that only default config is supported and any non-defaults are silently ignored

torchao/prototype/float8nocompile/float8nocompile_linear_test.py
```diff
@@ -0,0 +1,90 @@
...
+)
+def test_matmul_with_args_in_hp(input_shape: tuple[int, int]):
+    assert torch.cuda.is_available()
+    device = "cuda"
+
+    # high precision inputs
+    input_bf16 = torch.randn(
+        input_shape, dtype=torch.bfloat16, device=device, requires_grad=True
+    )
+    x_input_bf16 = input_bf16.clone().detach().to(device).requires_grad_(True)
```
vkuzo, line 28: nit: the `x` and `y` var names are a bit confusing, as `y = torch.mm(x, w)` is widely used in a similar context

vkuzo (approved):

torchao/prototype/float8nocompile/float8nocompile_linear.py
danielvegamyhre: done, updated docstring as well

torchao/prototype/float8nocompile/float8nocompile_linear_test.py
danielvegamyhre: makes sense, updated vars prefixed with `x_` to prefix with `prod_`, and updated vars prefixed with `y_` to prefixed with `prototype_`

PR #1516, opened 2025-01-07
fix bug in tl.store mask for kernel _to_fp8_row_major_t_and_non_t

danielvegamyhre:
Discovered this bug while prototyping torchtitan integration with float8nocompile in https://github.com/pytorch/torchtitan/pull/778, the symptom was the grads went to NaN during training.

The issue was a mismatch between how output offsets for writing the transposed tensor are calculated, versus the offsets used in creating the output mask. Output offsets for writing were correct but the offsets used in the output mask were backwards accidentally.


danielvegamyhre:
* (to be filled)

vkuzo:
looks like there is a gap in test coverage, is it doable to add a test to this PR which fails before the fix and passes after the fix?

danielvegamyhre:
> looks like there is a gap in test coverage, is it doable to add a test to this PR which fails before the fix and passes after the fix?

ah thanks for the reminder, just updated the e2e training test with a new test case with large inputs, which I've confirmed fails without this change and passes with this change.

vkuzo (approved):

PR #1517, opened 2025-01-07
[float8nocompile] support option to not precompute fp8 tensor for backward

danielvegamyhre:
**Summary**
Below are the differences in what we compute during fp8 conversion in the forward/backward pass, depending on if `no_precompute_for_backward` is enabled or not.

AC enabled:
- Compute `input_fp8_row_major` in the forward pass and compute `input_fp8_col_major` in the backward pass (i.e., don't precompute anything for backward)

AC not enabled:
- Compute `input_fp8_row_major` and `input_fp8_col_major` in the forward pass, saving the `input_fp8_col_major` for backward.

**Performance**
Benchmarks with torchtitan training Llama3 8b on 8 H100s with 4096 seq len, using full AC and selective per layer AC are below. Comparing with production float8 + compiling linear layers, this prototype roughly matched tokens/second while peak memory usage is less than or equal.

Screenshot 2025-01-14 at 4 11 49 PM


danielvegamyhre:
* (to be filled)

vkuzo:
nice! two questions:
1. is it possible to detect if the user is using AC instead of hardcoding it?  cc @soulitzer if this API already exists or if it's doable to add it in the future
2. do the kernels used in this PR match what torch.compile is doing, holding AC setting constant?

danielvegamyhre:
> 1. is it possible to detect if the user is using AC instead of hardcoding it?  cc @soulitzer if this API already exists or if it's doable to add it in the future

I talked with Jeffrey about this last week and my understanding is right now there is no API to detect if AC is enabled or not, though he said it could be a reasonable feature request.

>do the kernels used in this PR match what torch.compile is doing, holding AC setting constant?

I'm not sure how torch.compile handles AC enablement, i'll double check

danielvegamyhre:
Performance benchmarking with torchtitan shows this type of AC actually degrades performance of the Float8NoCompileLinear layer. 

I will include the explanation of this and gap analysis in a doc summarizing the project and results.

@vkuzo are we good to merge this, or should we close given the above?

torchao/prototype/float8nocompile/float8nocompile_linear.py
```diff
@@ -36,47 +38,57 @@ def __init__(self, *args, **kwargs):
...
-        # TODO(danielvegamyhre): support for FSDP once dependencies are implemented
-        output = matmul_with_args_in_hp.apply(
-            input,
-            self.weight,
-            self.config,
-            self.linear_mm_config,
-            self.kernel_algo,
-        )
+        if self.use_activation_checkpointing:
+            output = checkpoint(
```
soulitzer, line 74: The issue here is that you don't really want to be wrapping a single operator with activation checkpoint.

Detecting whether you're currently in AC and doing the right thing based on that is probably still what you want.

danielvegamyhre: I've updated the implementation to:

1) Not use torch.util.checkpoint.checkpoint
2) Rename the feature `no_precompute_for_backward` to avoid confusion with torch native AC

I've also updated the PR description to include performance benchmarking of this setting compared to others.

danielvegamyhre:
@vkuzo I updated the PR description with final benchmarks utilizing this `no_precompute_for_backward_feature` showing for Full AC and Selective Per Layer AC, the tokens/sec is equivalent and peak memory usage is lower than prod float8+compiling linear layers: https://github.com/pytorch/ao/pull/1517 

I think since this is an improvement over the baseline of prod float8+compile linear layers, this should be good to merge. However, I am still unclear on why both approaches have equal or greater memory usage than bfloat16 eager mode.

torchao/prototype/float8nocompile/float8nocompile_linear.py
```diff
@@ -111,7 +113,40 @@ def from_float(
...
+    def forward(
+        ctx,
+        input_hp: torch.Tensor,
+        weight_hp: torch.Tensor,
+        config: Float8LinearConfig,
+        linear_mm_config: LinearMMConfig,
+        kernel_algo: KernelAlgorithm,
+        no_precompute_for_backward: bool,
+    ):
+        if no_precompute_for_backward:
```
vkuzo, line 125: nit: just have two top-level functions instead?  People usually expect `torch.autograd.Function` to have a single forward and backward.

torchao/prototype/float8nocompile/float8nocompile_linear_utils.py
```diff
@@ -45,10 +46,30 @@ def convert_to_float8_nocompile_training(
...
+        no_precompute_for_backward=no_precompute_for_backward,
     )
     return swap_linear_layers(
         module,
         from_float,
         module_filter_fn=module_filter_fn,
     )
+
+
+def no_precompute_for_backward_every_nth_layer(model: nn.Module, n: int):
```
vkuzo, line 61: I don't think I would land this here, I'd instead just ask the user to implement this on their own.  This is more of a training loop / AC design kind of thing.

vkuzo (approved):

torchao/prototype/float8nocompile/float8nocompile_linear_utils.py
danielvegamyhre: I see, yeah I orginally had this in torchtitan training code but felt like this might be a better home for it. I can remove it from this PR.

torchao/prototype/float8nocompile/float8nocompile_linear.py
danielvegamyhre: Done

PR #1518, opened 2025-01-07
Lint torchao folder

jainapurva:
Lint all folders in torchao, with an exception for experimental. Added experimental to exclude list.


ruff.toml
```diff
@@ -3,20 +3,13 @@
 # To add a new path: Simply add it to the 'include' list.
 # Example: To lint all files in every subfolder of 'test', add "test/**/*"
 include = [
-    "torchao/float8/**/*.py",
```
drisspg, line 6: remove the include list

jainapurva: Did that in another PR #1521  . A couple more folders were to be linted before removing include

jainapurva: Had to divide the changes in multiple PRs to avoid extremely long PR

PR #1531, opened 2025-01-09
Skip calling unwrap_tensor_subclass for torch 2.7+

jerryzh168:
Summary:
att

Test Plan:
python test/integration/test_integration.py -k TestExport

AOTI test also works, but there is some AOT compile issues, reproducible by python test/integration/test_integration.py -k TestAOTI.test_aoti_06, error: https://gist.github.com/jerryzh168/824840130f30ae11bfc1506a7500b931

Reviewers:

Subscribers:

Tasks:

Tags:


drisspg (approved):
Awesome!

PR #1539, opened 2025-01-10
Add support for eager mode performance

jerryzh168:
Summary:

Added "compile" filed to "extra_info" that allows us to record eager mode performance as well

context is eager, eager + compile, eager + compile + autoquant can all have performance improvements/changes over time, so we want to track:

(1) eager perf on some previous date (configurable by user) (2) current eager perf
(3) current compile perf
(4) current autoqunat + compile perf

Test Plan:
tested locally:
https://gist.github.com/jerryzh168/2a15322b0c8f40f35e52956837c67fec

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/examples/sam2_amg_server/server.py
+++ b/examples/sam2_amg_server/server.py
@@ -508,7 +508,7 @@ def main(
 
     # since autoquant is replicating what furious mode is doing, don't use these two together
     if autoquant_type is not None:
-        assert not furious, "use autoquant can't be used together with furious"
+        assert not furious, "autoquant can't be used together with furious"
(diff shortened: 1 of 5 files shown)


examples/sam2_amg_server/server.py
```diff
@@ -569,9 +569,22 @@ def main(
 
         if output_json_path:
             headers = ["name", "dtype", "device", "arch", "metric", "actual", "target"]
+            headers = [
```
huydhn, line 572: This look like a typo as `headers` is redefined

huydhn:
Let me start the dashboard speedup part now this is landed

PR #1543, opened 2025-01-10
SAM2: Rerun batch size 1 experiments on latest nightly

cpuhrsch:
Updated results and track version of torch and torchvision


examples/sam2_amg_server/compile_export_utils.py
```diff
@@ -332,9 +332,6 @@ def load_exported_model(
 def set_fast(
     mask_generator, task_type, loaded_exported_model=False, allow_recompiles=True
 ):
-    if task_type == "":
```
jerryzh168, line 335: this is for autoquant benchmark, or should we start with exported model for autoquant as well?

PR #1515, opened 2025-01-07
Add CUTLASS-based W4A4

gau-nernst:
Closes #1406

Thanks to #880, we now have a CUTLASS (3.6.0) copy in torchao. Adding W4A4 is pretty straight-forward, similar to how W4A8 is done. ~~This is largely copied from my other repo, so I didn't exactly follow @alexsamardzic's style. Requesting a first round of review.~~
- Update: Thanks to @alexsamardzic's refactoring efforts, now W4A8 and W4A4 share most of the code, with template instantiation for each GEMM type.

Note: this is more for doing experiments with W4A4 easier. Personally I don't think it's too useful at the moment, since W4A4 accuracy is probably quite bad.

TODO:
- [x] Hook up to AQT
- [x] Benchmark script and get benchmark results for A100 and 4090
- [ ] (Maybe) Do some tuning + heuristics based on problem size

<details>
<summary><b>A100 SXM4</b> (from vast.ai)</summary>

|   m |     k |     n |   fp16_latency (ms) |   W4A8 latency (ms) |   W4A8 speedup (d/s) |   W4A4 latency (ms) |   W4A4 speedup (d/s) |
|----:|------:|------:|--------------------:|--------------------:|---------------------:|--------------------:|---------------------:|
|   1 |  8192 |  8192 |             110.592 |              51.2   |              2.16    |              38.912 |              2.84211 |
|   1 |  8192 | 10240 |             123.904 |              56.32  |              2.2     |              45.056 |              2.75    |
|   1 |  8192 | 57344 |             569.344 |             206.848 |              2.75248 |             156.672 |              3.63399 |
|   1 | 28672 |  8192 |             299.008 |             107.52  |              2.78095 |              92.16  |              3.24444 |
|   2 |  8192 |  8192 |             108.544 |              50.176 |              2.16327 |              38.912 |              2.78947 |
|   2 |  8192 | 10240 |             129.024 |              57.344 |              2.25    |              45.056 |              2.86364 |
|   2 |  8192 | 57344 |             581.632 |             206.848 |              2.81188 |             157.696 |              3.68831 |
|   2 | 28672 |  8192 |             293.888 |             107.52  |              2.73333 |              92.16  |              3.18889 |
|   4 |  8192 |  8192 |             109.568 |              50.176 |              2.18367 |              38.912 |              2.81579 |
|   4 |  8192 | 10240 |             130.048 |              57.344 |              2.26786 |              45.056 |              2.88636 |
[...]

PR #1545, opened 2025-01-10
Refactor s8s4_linear_cutlass()

alexsamardzic:


alexsamardzic:
Refactored the code so that it is easier to extend it with combinations of quantized operands data types other than W4A8, like one proposed in #1515.  Disadvantage: compilation time increased.

drisspg (approved):

PR #1548, opened 2025-01-10
[super ugly not working code] use shim.h instead of Tensor

janeyx99:
I KNOW THIS DOES NOT BUILD

this is a poc change to see what making custom ops use aoti shim.h would look like

### DIFF
--- a/torchao/csrc/cuda/tensor_core_tiled_layout/tensor_core_tiled_layout.cu
+++ b/torchao/csrc/cuda/tensor_core_tiled_layout/tensor_core_tiled_layout.cu
@@ -1,10 +1,12 @@
 #if !defined(__CUDA_ARCH__) || __CUDA_ARCH__ >= 800  // at least Ampere
 
-#include <ATen/ATen.h>
-#include <ATen/core/Tensor.h>
-#include <ATen/cuda/CUDAContext.h>
+// #include <ATen/ATen.h>
+// #include <ATen/core/Tensor.h>
 #include <ATen/DeviceGuard.h>
+#include <ATen/cuda/CUDAContext.h>
 #include <c10/cuda/CUDAGuard.h>
+#include <torch/csrc/inductor/aoti_torch/c/shim.h>
+#include <torch/csrc/inductor/aoti_torch/utils.h>
 #include <torch/extension.h>
 
 template <typename U, typename V>
@@ -15,18 +17,18 @@ constexpr __host__ __device__ auto divUp(U a, V b) -> decltype(a + b) {
 }
 constexpr int32_t kWarpSize = 32;
 
-//Simple data structure to represent 4 pairs of bfloat16s, used for vectorized dequantization
-//https://github.com/pytorch/pytorch/blob/b6689e0fb83a1578959ab0d9c6d2d9e11f7df21a/aten/src/ATen/native/cuda/int4mm.cu#L178-L180
-struct __align__(16) bf16x2x4 {
-  __nv_bfloat162 vals[4];
-};
+// Simple data structure to represent 4 pairs of bfloat16s, used for vectorized
+// dequantization
+// https://github.com/pytorch/pytorch/blob/b6689e0fb83a1578959ab0d9c6d2d9e11f7df21a/aten/src/ATen/native/cuda/int4mm.cu#L178-L180
+struct __align__(16) bf16x2x4 { __nv_bfloat162 vals[4]; };
 
-//Copied from https://github.com/pytorch/pytorch/blob/b6689e0fb83a1578959ab0d9c6d2d9e11f7df21a/aten/src/ATen/native/cuda/int4mm.cu#L195C1-L241C1
+// Copied from
+// https://github.com/pytorch/pytorch/blob/b6689e0fb83a1578959ab0d9c6d2d9e11f7df21a/aten/src/ATen/native/cuda/int4mm.cu#L195C1-L241C1
 inline __device__ bf16x2x4 convert_i4x8_to_bf16x2x4(uint32_t source) {
   bf16x2x4 result;
   constexpr int kElements = 8;
 
-  uint32_t* h = reinterpret_cast<uint32_t*>(&result);
+  uint32_t *h = reinterpret_cast<uint32_t *>(&result);
   uint32_t const source_i4s = source;
 
   // First, we extract the i4s and construct an intermediate fp16 number.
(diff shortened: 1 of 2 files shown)


torchao/csrc/cuda/tensor_core_tiled_layout/tensor_core_tiled_layout.cu
```diff
@@ -15,18 +19,18 @@ constexpr __host__ __device__ auto divUp(U a, V b) -> decltype(a + b) {
 }
 constexpr int32_t kWarpSize = 32;
 
-//Simple data structure to represent 4 pairs of bfloat16s, used for vectorized dequantization
```
janeyx99, line 18: UGH IGNORE the random VSCode formats--Idk how autoformat is turned on in this repo for me + will figure it out later, skip this line til 170!!!

torchao/csrc/cuda/tensor_core_tiled_layout/tensor_core_tiled_layout.cu
```diff
@@ -146,121 +157,210 @@ __global__ void _dequantize_int4_kernel(
...
+      int2 *v_i32x2 = reinterpret_cast<int2 *>(v);
 
-    #pragma unroll
+#pragma unroll
       for (int i = 0; i < 4; ++i) {
         reinterpret_cast<int2 *>(&pOut[ks[i]])[0] = v_i32x2[i];
       }
     }
   }
 }
```
janeyx99, line 169: start looking below!

torchao/csrc/cuda/tensor_core_tiled_layout/libtorch.cpp
```diff
@@ -0,0 +1,245 @@
...
+        : library_(kind, ns, k, file, line) {}
+private:
+    torch::Library library_; // Actual Library object
+};
+
+StableLibrary::StableLibrary(StableLibrary::Kind kind, std::string ns, std::optional<c10::DispatchKey> k, const char* file, uint32_t line)
+    : lib_(&TorchLibraryOpaque(Library::Kind::IMPL, ns, k, file, line)) {}
+
+StableLibrary& StableLibrary::impl(std::string name, void (*fn)(void **, int64_t, int64_t)) {
+  auto boxed_function = [fn](const c10::OperatorHandle &op, torch::jit::Stack *stack) {
```
kimishpatel, line 196: are these, `torch::jit:*` and IValue, considered stable?

torchao/csrc/cuda/tensor_core_tiled_layout/libtorch.cpp
```diff
@@ -0,0 +1,245 @@
...
+//          torch::CppFunction::makeFromBoxedFunction<
+//              boxed_unpack_tensor_core_tiled_layout>());
+//   // m.impl("torchao::dequantize_tensor_core_tiled_layout",
+//   // &_dequantize_tensor_core_tiled_layout);
+//   m.impl("torchao::dequantize_tensor_core_tiled_layout",
+//          torch::CppFunction::makeFromBoxedFunction<
+//              boxed_dequantize_tensor_core_tiled_layout>());
+// }
+
+class StableLibrary::TorchLibraryOpaque {
```
kimishpatel, line 184: I am guessing that, from design perspective, this is the interface layer with registration API that is shipped with different versions of libtorch? But the user code, for custom ops, never relies on at::Tensor?

torchao/csrc/cuda/tensor_core_tiled_layout/tensor_core_tiled_layout.cu
```diff
@@ -282,35 +399,75 @@ at::Tensor _unpack_tensor_core_tiled_layout(
...
+        <<<grid, kWarpSize, 0, stream>>>(packed_w_pta32, out_pta32);
   }
 
   return out;
 }
 
-TORCH_LIBRARY_IMPL(torchao, CUDA, m) {
-  m.impl("torchao::unpack_tensor_core_tiled_layout", &_unpack_tensor_core_tiled_layout);
-  m.impl("torchao::dequantize_tensor_core_tiled_layout", &_dequantize_tensor_core_tiled_layout);
+void voidyvoid_boxed_ATH_unpack_tensor_core_tiled_layout(void **stack,
```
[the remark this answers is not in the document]
kimishpatel: on serious note though, will user have to write this function with `void** stack`? It would actually be good if user just write with stable API. LIke AtenTensorHandle

torchao/csrc/cuda/tensor_core_tiled_layout/libtorch.cpp
janeyx99: no! I haven't gotten a chance to document what I'm trying to do, but everything in libtorch.cpp should eventually make its way back to libtorch (not stable!!)

everything in libtorch.h is supposed to be stable.

torchao/csrc/cuda/tensor_core_tiled_layout/libtorch.cpp
janeyx99: Yea, this is to allow for registering libtorch-agnostic custom ops which cannot use at::Tensor or IValue, etc. in their schema

torchao/csrc/cuda/tensor_core_tiled_layout/tensor_core_tiled_layout.cu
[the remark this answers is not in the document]
janeyx99: Yea, to be clear, there are two portions we expect the user to provide for now:
1. their custom op, which should directly use AtenTensorHandle (we can make that easier by wrapping AtenTensorHandle with a headeronly C++ API layer)
2. how their custom op should be registered within our dispatcher stack (this is the point of the voidyvoid function). There may be a way to automatically generate this for the user given the schema of their custom op from (1) but that is a next step. Currently, we are going to expect users to provide this.

torchao/csrc/cuda/tensor_core_tiled_layout/tensor_core_tiled_layout.cu
```diff
@@ -146,121 +155,167 @@ __global__ void _dequantize_int4_kernel(
...
-  TORCH_CHECK(
-      group_size == 32 || group_size == 64 || group_size == 128 ||
-      group_size == 256);
+  TORCH_CHECK(group_size == 32 || group_size == 64 || group_size == 128 ||
+              group_size == 256);
   TORCH_CHECK(numQGroups == K / group_size);
-  TORCH_CHECK(scales_and_zeros.dim() == 3);
-  TORCH_CHECK(scales_and_zeros.size(1) == N);
-  TORCH_CHECK(scales_and_zeros.size(2) == 2);
+
```
msaroufim, line 225: i might be a bit out of the loop but have you considered helpers like the below so the code before and after looks similar

```cpp
inline int64_t get_dim(aoti_tensor_t tensor) {
    int64_t dim;
    aoti_torch_get_dim(tensor, &dim);
    return dim;
}

inline int32_t get_dtype(aoti_tensor_t tensor) {
    int32_t dtype;
    aoti_torch_get_dtype(tensor, &dtype);
    return dtype;
}

inline int64_t get_size(aoti_tensor_t tensor, int64_t dim) {
    int64_t size;
    aoti_torch_get_size(tensor, dim, &size);
    return size;
}
```

janeyx99: The plan is to eventually have a top level helper C++ API make this easier, but yes the shim.c style API is verbose.

PR #1551, opened 2025-01-11
Remove temp build files from torchao

metascroy:
Summary: Removes temp build artifacts from experimental.  Now the kernels are built and loaded with `USE_CPP=1 pip install .` from ao.

Reviewed By: jerryzh168

Differential Revision: D67807207

### DIFF
--- a/torchao/_models/llama/generate.py
+++ b/torchao/_models/llama/generate.py
@@ -548,14 +548,15 @@ def main(
                 precision == torch.float32
             ), "int8_dynamic_activation_intx_weight requires fp32 precision"
 
-            # Build kernels in temp location, and load them in torch
-            # This requires an ARM CPU
-            from torchao.experimental.temp_build import temp_build_and_load_torchao_ops
-
-            temp_build_and_load_torchao_ops(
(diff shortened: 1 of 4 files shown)


torchao/_models/llama/generate.py
```diff
@@ -548,14 +548,15 @@ def ffn_or_attn_only(mod, fqn):
...
-                + "/../../experimental"
-            )
+            try:
+                torch.ops.torchao._pack_8bit_act_4bit_weight
+            except:
+                print(
+                    "Unable to load experimental torchao kernels.  Performance will be slow."
+                )
+                print(
+                    "To install the kernels, run `USE_CPP=1 pip install .` from ao on a machine with an ARM CPU"
```
jerryzh168, line 558: is this included in the default build as well?

jerryzh168 (approved):

torchao/_models/llama/generate.py
metascroy: I quickly glanced and it does not look like the default nightly build for M1 mac defines USE_CPP=1, although adding it should be pretty simple.

PR #1049, opened 2024-10-10
Rename AQT#2 LayoutType -> Layout

jainapurva:
### PR changes:
Current Layout/Layout type confusing and not corresponding to pytorch layout

- Current LayoutType class and aqt.layout_type corresponds to tensor.layout in pytorch
- Rename
      - “LayoutType” to “Layout”
      - “aqt.layout_type” to “aqt._layout” for more alignment with pytorch native Tensor
      - _layout -> layout for public API

This change also breaks backwards compatibility for all users that explicitly defined the layout_type parameter:

### BC-breaking note:
Prev PR: #1046 -> Layout renamed to TensorImpl

```py
# Before
class AQTLayout(TorchAOBaseTensor):
...

# After
class AQTTensorImpl(TorchAOBaseTensor):
...
```

- LayoutType's renamed to Layout
- For quant_api's () the layout_type param has been renamed to layout.
- Affected APIs:
      - int4_weight_only
      - int8_dynamic_activation_int8_weight
      - int8_dynamic_activation_int8_semi_sparse_weight

```py
# Before change
def int4_weight_only(group_size=128, layout_type=TensorCoreTiledLayoutType(inner_k_tiles=8), use_hqq=False):
...

quantize_(int4_weight_only(group_size=128, layout_type=TensorCoreTiledLayoutType(inner_k_tiles=8), use_hqq=False))

# After change
def int4_weight_only(group_size=128, layout=TensorCoreTiledLayout(inner_k_tiles=8), use_hqq=False):
...

quantize_(int4_weight_only(group_size=128, layout=TensorCoreTiledLayout(inner_k_tiles=8), use_hqq=False))
```

PR #1556, opened 2025-01-13
 Consolidate `ZeroPointDomain.NONE` & `None` zero point domains

sanchitintel:
## Summary

- Both `ZeroPointDomain.NONE` & `None` zero point domains were being used. The latter was being used for float8. This PR consolidates both & retains `ZeroPointDomain.NONE`
- Using `ZeroPointDomain.NONE` [would now produce `None` for zero-point](https://github.com/pytorch/ao/blob/cd7abef6a6c37940be88a8b6e24fa66bba344f71/torchao/quantization/quant_primitives.py#L927-L930)
- `int8_dynamic_activation_int8_weight` now uses `ZeroPointDomain.NONE` as weight zero point domain (as weight is quantized symmetrically to int8).

Some of the older changes in this PR (such as supporting torch.compile with optional zero_point) were rendered redundant by more recent changes in the main branch, so I removed them & modified the description accordingly. Thanks!

### BC-breaking note:

A `None` value for `zero_point_domain` would no longer be accepted. Please use `ZeroPointDomain.NONE` instead.
A `None` zero-point would be generated if  `ZeroPointDomain.NONE` would be used.

#### _Before this PR_:

`None` value of `zero_point_domain` was being used for quantization to float8 dtypes to indicate the absence of a zero-point.
`ZeroPointDomain.NONE` was being used for quantization to non-float8 dtypes to indicate the absence of a zero-point.
`choose_qparams_affine` returned a non-None zero-point for symmetric quantization.

#### _After this PR lands_:
`ZeroPointDomain.NONE` value for `zero_point_domain` should be used to denote the absence of a zero-point.
zero point in such cases would be `None`.

###### For the following user-facing APIs, a `None` `zero_point_domain` can no longer be used with the following APIs (ZeroPointDomain.NONE should be used instead of `None`) -
`torchao.quantization.quant_api.int4_weight_only`

##### `zero_point_domain` is no longer optional for the following developer-facing APIs -

`torchao.quantization.quant_primitives.choose_qparams_affine`
`torchao.quantization.quant_primitives.choose_qparams_affine_with_min_max`
`torchao.quantization.quant_primitives.fake_quantize_affine_cachemask`
`torchao.quantization.quant_primitives.quantize_affine`
`torchao.quantization.observer.AffineQuantizedObserverBase`
`torchao.dtypes.uintx.marlin_qqq_tensor.MarlinQQQTensor.from_hp_to_intx`
`torchao.dtypes.AffineQuantizedTensor.from_hp_to_intx`
`torchao.dtypes.AffineQuantizedTensor.from_hp_to_intx_static`

###### For the following developer-facing APIs, a `None` `zero_point_domain` can no longer be used with the following APIs (ZeroPointDomain.NONE should be used instead of `None`) -

`torchao.quantization.quant_primitives.fake_quantize_affine`
`torchao.quantization.quant_primitives.dequantize_affine`
`torchao.quantization.qat.FakeQuantizeConfig`

##### For the following internal APIs (whose names begin with an underscore character), `zero_point_domain` can no longer be `None` (`ZeroPointDomain.NONE` should be used instead of `None`): 
`torchao.quantization.qat._ToAffineFakeQuantized`


torchao/quantization/quant_primitives.py
```diff
@@ -389,13 +389,10 @@ def _quantize_affine_no_dtype_cast(
...
             zero_point is None
         ), "zero_point should be None when zero_point_domain is NONE"
-        quant = torch.clamp(torch.round(input * (1.0 / scale)), quant_min, quant_max)
-    elif zero_point_domain is None:
-        # This case handles quantization for float8 we expect no zero point and no zero point domain
-        assert (
-            zero_point is None
-        ), "zero_point should be None when zero_point_domain is None"
-        quant = torch.clamp(input * scale.reciprocal(), quant_min, quant_max)
+        if _is_float8_type(input.dtype):
```
sanchitintel, line 392: This check is incorrect because it'd work in the dequantize case, and not the quantize case (this one). So I'll fix it.

Is there some way to ascertain if we're trying to quantize to float8 here? Sorry, this path requires an Nvidia GPU of a higher Compute Capability than the one I have locally. Thanks!

cc @drisspg

drisspg: It seems to me like the main behavioral change we want here when there is no zero point domain is to know whether we should round or not round the scale values before clamping. Is that correct?

sanchitintel: Yes, that's right. Thanks!

sanchitintel: @drisspg, but this behavior differs for float8 and non-float8 quantization, so I'm wondering how to differentiate both cases, since the output dtype is not exposed to this function (viz. `_quantize_affine_no_dtype_cast`). Thanks!

Would it be okay to match `quant_min` & `quant_max` with that of FP8 types? I feel like practically, doing so would not lead to any side-effects. Thanks!

jerryzh168: we really should be separating out these ops so it's easier to maintain I think, especially for fp8, we'd want separate quant primitive ops, cc @jainapurva 

for now I think whatever passes test would be OK. the target dtype is only available out side of this function: https://github.com/pytorch/ao/blob/de5c6e12f3f8cf3ce3d682880873034baf3cb4ad/torchao/quantization/quant_primitives.py#L338, so there is no way to distinguish it from here, I think you'll need to pass around a flag from the callsite of this function instead

jerryzh168 (approved):
LGTM

sanchitintel:
Thanks for reviewing, @jerryzh168! I'll fix the lint error once other CI checks would complete.

sanchitintel:
Looks like the failing UT behaves differently on AMD machines used by both of those CI instances due to different CPU ISAs than my local machine, and Inductor codegen for some specific case is failing on those machines due to difference in vectorization support.

I'll get access to a VM corresponding to the underlying AWS instances, and will try to have that issue fixed in PyTorch. Thanks!

jerryzh168:
for https://github.com/pytorch/ao/actions/runs/12760810579/job/35567540271?pr=1556 you can install pre-commit

`pip install pre-commit`

and run it:

`pre-commit run`

then it will run the formatting before every `git commit`

test/integration/test_integration.py
```diff
@@ -959,25 +976,49 @@ def _test_lin_weight_subclass_api_impl(
...
+            f"API failed, no compile dtype={test_dtype}, (m, k, n)={test_shape}",
         )
 
         mod_qc = torch.compile(mod, mode="max-autotune")
         test_comp = mod_qc(x)
         self.assertGreater(
             SQNR(ref_f, test_comp),
             min_sqnr,
-            f"{api.__name__} failed when compiled with dtype={test_dtype}, (m, k, n)={test_shape}",
+            f"API failed when compiled with dtype={test_dtype}, (m, k, n)={test_shape}",
```
drisspg, line 991: Why this change?

test/integration/test_integration.py
```diff
@@ -959,25 +976,49 @@ def _test_lin_weight_subclass_api_impl(
...
+                COMMON_DTYPES,
+                ACT_MAPPING_TYPES,
+                WEIGHT_ZERO_POINT_DOMAINS,
+            )
+        )
+    )
+    def test_int8_dynamic_quant_subclass_api(
+        self, device, dtype, act_mapping, weight_zero_point_domain
+    ):
+        from functools import partial
```
drisspg, line 1007: Nit global import

torchao/quantization/quant_primitives.py
```diff
@@ -371,6 +372,7 @@ def _quantize_affine_no_dtype_cast(
     zero_point: Optional[torch.Tensor],
     quant_min: Union[int, float],
     quant_max: Union[int, float],
+    quant_dtype: Optional[torch.dtype],
     zero_point_domain: Optional[str] = ZeroPointDomain.INT.name,
```
drisspg, line 376: This shouldn't be optional any more right?

sanchitintel: Oops, I missed it. Thank you!

test/integration/test_integration.py
sanchitintel: Because of `api` using `functools.partial`, it wouldn't have `__name__` attribute anymore.

drisspg (approved):

drisspg (approved):

test/quantization/test_observer.py
```diff
@@ -199,7 +200,6 @@ def test_linear_observer_tensor(self, observe_weight: bool):
             input_scale.item(),
             max_val / max_fp8,
         )
-        self.assertIsNotNone(input_zero_point)
```
jerryzh168, line 202: is there a change of behavior when you change zero_point_domain for None to ZeroPointDomain.NONE?

test/quantization/test_quant_primitives.py
```diff
@@ -838,6 +838,32 @@ def test_fake_quantize_affine_cachemask(self):
         torch.testing.assert_close(dequantized, fake_quantized)
         torch.testing.assert_close(expected_mask, mask)
 
+    # ZeroPointDomain.NONE should work
+    def test_none_zero_point_domain(self):
```
jerryzh168, line 842: we could also have a test for zero_point_domain being None and throw an IllegalArgumentError exception I think

sanchitintel: Thanks again for reviewing! I added code for raising `ValueError` if `zero_point_domain` would be `None`, but only for `choose_qparams_affine` and `choose_qparams_affine_with_min_max`.

For other public-facing APIs that have `zero_point_domain` as a parameter, I added asserts instead.

Please advise if this is fine, or if other places in the code should also throw exceptions instead of asserting.

Thanks!

torchao/quantization/quant_api.py
```diff
@@ -857,6 +857,7 @@ def int8_dynamic_activation_int8_weight(
     layout=PlainLayout(),
     act_mapping_type=MappingType.SYMMETRIC,
     weight_only_decode=False,
+    weight_zp_domain=ZeroPointDomain.NONE,
```
sanchitintel, line 860: Hi @jerryzh168, I added a new argument `weight_zp_domain` to `int8_dynamic_activation_int8_weight` & made it `ZeroPointDomain.NONE` by default, since weight is being symmetrically quantized here.

Instead of this change, I could also have instead called `to_affine_quantized_intx` with `zero_point_domain=ZeroPointDomain.NONE`, but I didn't do so because in the main branch, the default value of `ZeroPointDomain.INT` is being used here.

Thanks!

test/quantization/test_observer.py
jerryzh168: I see, so what is the meaning of `zero_point_domain` == None before?

torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -302,10 +303,8 @@ def from_hp_to_intx_static(
         zero_point_domain: Optional[ZeroPointDomain] = ZeroPointDomain.INT,
         _layout: Layout = PlainLayout(),
     ):
+        assert zero_point_domain is not None, "zero_point_domain must not be None"
```
jerryzh168, line 306: zero_point_domain should not be Optional in L303 I think

jerryzh168: also maybe raise ValueError here, that might be more user friendly than assertion

torchao/quantization/quant_api.py
jerryzh168: actually here, I think we can just hardcode the zero_point_domain in when we call `to_affine_aquantized_intx` in this function, if it's always true for int8_act_int8_weight, unless you have a use case that is not doing this?

test/quantization/test_observer.py
sanchitintel: Some APIs were creating a `None` `zero_point` when `zero_point_domain` `ZeroPointDomain.NONE` or `None` was used, while `choose_qparams_affine` was not.

torchao/dtypes/affine_quantized_tensor.py
sanchitintel: Thank you! Made those modifications

test/quantization/test_observer.py
sanchitintel: Hi @jerryzh168, is it possible that some torchao users' code may be expecting a non-None `zero_point` with zero_point_domain `ZeroPointDomain.NONE`/None, making this change BC-breaking for them? Thanks!

drisspg: I think that most usages of this function are internal to torchao so that its okay to BC break, you can add the label just to be sure

sanchitintel: Thanks for your advice, @drisspg!
Could you please help add such a label as GitHub isn't displaying an option to me for adding it? Thanks!

jerryzh168 (approved):
LGTM, please rebase and land

test/quantization/test_observer.py
jerryzh168: I added a bc-breaking label, please also write a bc-breaking note similar to https://github.com/pytorch/ao/pull/1049

sanchitintel: Thanks again, @jerryzh168! I added a note & rebased the PR.

PR #1566, opened 2025-01-15
[float8nocompile] add triton kernel which does fp8 conversion to col major and transpose in col major at once

danielvegamyhre:
This will be useful when no AC is used. We need fp8 weight transpose in col major (forward pass), and fp8 weight in col major (forward pass). Right now we always compute them separately using 2 kernels. With no AC, it makes sense to have a single kernel read once and write twice, rather than dispatch 2 separate kernels.

Performance benchmarking in #1567 shows tok/sec improvement at the cost of some memory, which is a fine trade-off when not using AC.


danielvegamyhre:
* (to be filled)

vkuzo (approved):

PR #1577, opened 2025-01-17
INT4 XPU enabling

airMeng:
~~The PR is a draft currently.~~

The PR will add 2 kinds of INT4 support on XPU: floating zero points and integer zero points, following the discussion in https://github.com/pytorch/ao/issues/1264.

Integer zero points which has been natively supported via OneDNN https://github.com/pytorch/pytorch/pull/137566

Floating zero points, the default behaviour in this repo, supported by https://github.com/intel/torch-xpu-ops/pull/1130, more implementations on the way.


test/dtypes/test_affine_quantized.py
```diff
@@ -46,6 +47,18 @@ def get_quantization_functions(
                         zero_point_domain=ZeroPointDomain.INT,
                     )
                 )
+        elif device == "xpu" and  TORCH_VERSION_AT_LEAST_2_6:
```
jerryzh168, line 50: 2_7 or 2_6?

test/integration/test_integration.py
```diff
@@ -1079,6 +1084,8 @@ def test_int4_weight_only_quant_subclass_api_grouped(self, device, dtype):
         layout_list = []
         if device == "cpu" and TORCH_VERSION_AT_LEAST_2_6:
             layout_list.append(Int4CPULayout())
+        elif device == "xpu" and TORCH_VERSION_AT_LEAST_2_6:
```
jerryzh168, line 1087: here as well, 2_6 or 2_7?

torchao/dtypes/uintx/int4_xpu_layout.py
```diff
@@ -0,0 +1,283 @@
...
+                    f"Int4XPUAQTTensorImpl dispatch: attempting to run {func}, with dim={dim}, that is not supported"
+                )
+
+        raise NotImplementedError(
+            f"Int4XPUAQTTensorImpl dispatch: attempting to run {func}, this is not supported"
+        )
+
+    __torch_function__ = torch._C._disabled_torch_function_impl
+
+    def get_plain(self) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
```
jerryzh168, line 210: btw for this one, we have some unpacking op for tensor core tiled layout that we should really be using: https://github.com/pytorch/ao/blob/cf453360dd3e09394657172f8e8d8da23cfbf043/torchao/csrc/cuda/tensor_core_tiled_layout/tensor_core_tiled_layout.cu#L311-L312

test/dtypes/test_affine_quantized.py
airMeng: depend on the https://github.com/pytorch/pytorch/pull/137566. It's really just a draft, so not yet ready for review. I will ping you when ready :)

jerryzh168:
btw why the op is added in https://github.com/pytorch/pytorch/pull/137566 instead of in torchao? any plans to move it to torchao?

torchao/dtypes/uintx/int4_xpu_layout.py
airMeng: sure. I will give a check.

airMeng:
> btw why the op is added in [pytorch/pytorch#137566](https://github.com/pytorch/pytorch/pull/137566) instead of in torchao? any plans to move it to torchao?

@mingfeima @EikanWang can you comment?

mingfeima:
> > btw why the op is added in [pytorch/pytorch#137566](https://github.com/pytorch/pytorch/pull/137566) instead of in torchao? any plans to move it to torchao?
> 
> @mingfeima @EikanWang can you comment?

The situation is different for XPU (the intel GPUs) from CPU and CUDA here. Not sure that whether providing sycl or oneDNN xpu ops in ao is a feasible solution.

test/dtypes/test_affine_quantized.py
airMeng: now it should be 2_7

test/integration/test_integration.py
sunjiweiswift: 2.7

sunjiweiswift:
@jerryzh168 pls review again·

test/dtypes/test_affine_quantized.py
```diff
@@ -70,9 +75,10 @@ def get_quantization_functions(
                 base_functions.append(int4_dynamic_activation_int4_weight())
 
     if do_sparse:
-        base_functions.append(
-            int8_dynamic_activation_int8_weight(layout=SemiSparseLayout())
-        )
+        if device != "xpu":
```
jerryzh168, line 78: nit: we can merge the two ifs here

test/dtypes/test_affine_quantized.py
```diff
@@ -101,32 +110,31 @@ def test_tensor_core_layout_transpose(self):
...
+        for apply_quant in apply_quant_list:
+            linear = torch.nn.Linear(128, 256, dtype=torch.bfloat16, device=device)
+            if isinstance(apply_quant, AOBaseConfig):
+                quantize_(linear, apply_quant)
+                ql = linear
             else:
-                _ = torch.load(f, weights_only=False)
-
-    @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
+                # TODO(#1690): delete this once config migration is done
```
jerryzh168, line 123: cc @vkuzo we can delete these now?

torchao/dtypes/uintx/int4_xpu_layout.py
```diff
@@ -0,0 +1,433 @@
...
+        self.transposed = False
+        self._layout = _layout
+        self.scale = scale
+        self.zero = zero
+
+    def __tensor_flatten__(self):
+        if self.scale_and_zero is not None:
+            return ["packed_weight", "scale_and_zero"], [self.transposed, self._layout]
+        else:
+            return ["packed_weight", "scale", "zero"], [self.transposed, self._layout]
```
jerryzh168, line 214: why do we have two formats here? maybe should split into multiple layouts?

torchao/quantization/quant_api.py
```diff
@@ -798,6 +802,13 @@ class Int4WeightOnlyConfig(AOBaseConfig):
     zero_point_domain: Optional[ZeroPointDomain] = ZeroPointDomain.NONE
     set_inductor_config: bool = True
 
+    mapping_type = MappingType.ASYMMETRIC
+    block_size = (1, group_size)
```
jerryzh168, line 806: this seems to be a duplication of `group_size` arg?

torchao/quantization/quant_api.py
```diff
@@ -798,6 +802,13 @@ class Int4WeightOnlyConfig(AOBaseConfig):
...
+    target_dtype = torch.int32
+    quant_min = 0
+    quant_max = 15
+    eps = 1e-6
+    zero_point_dtype = torch.bfloat16
```
jerryzh168, line 811: also why all these ares are added to the layout itself?

torchao/quantization/utils.py
```diff
@@ -454,10 +472,11 @@ def groupwise_affine_dequantize_tensor_from_qparams(
     )
 
 
-def groupwise_affine_quantize_tensor(w, n_bit=4, groupsize=128, dtype=torch.bfloat16):
-    scales, zeros = get_groupwise_affine_qparams(w, n_bit, groupsize, dtype)
+def groupwise_affine_quantize_tensor(w, n_bit=4, groupsize=128, dtype=torch.bfloat16, int_zp=False):
+    scales, zeros = get_groupwise_affine_qparams(w, n_bit, groupsize, dtype, int_zp=int_zp)
```
jerryzh168, line 476: nit, for clarity: int_zp --> zero_point_domain_is_int

torchao/quantization/GPTQ.py
```diff
@@ -546,6 +546,14 @@ def linear_forward_int4(
             groupsize,
             scales_and_zeros.to(scales_precision),
         ).to(dtype=x.dtype)
+    elif is_device(x.device.type, "xpu") and TORCH_VERSION_AT_LEAST_2_7:
+        c = torch.ops.aten._weight_int4pack_mm_with_scales_and_zeros(
```
jerryzh168, line 550: also why do we have this function? can the slicing (scales_and_zeros[0] and scales_and_zeros[1]) be done in the `_weight_int4pack_mm` itself?

torchao/dtypes/uintx/int4_xpu_layout.py
airMeng: integer zp and floating zp
I don't split into 2 layout because from user side it will be confusing

current:
```
quantize_(model, int4_weight_only(group_size=32, layout=Int4XPULayout(), zero_point_domain=ZeroPointDomain.INT))
quantize_(model, int4_weight_only(group_size=32, layout=Int4XPULayout(), zero_point_domain=ZeroPointDomain.Float))
```

but if different layouts

```python
quantize_(model, int4_weight_only(group_size=32, layout=Int4XPULayoutIntZP(), zero_point_domain=ZeroPointDomain.INT))
quantize_(model, int4_weight_only(group_size=32, layout=Int4XPULayoutFloatZP(), zero_point_domain=ZeroPointDomain.Float))
```

I think the current implementation is more straightforward for users.

jerryzh168: but layout defines how we store the packed weights actually, using a single layout for multiple things is breaking this abstraction I feel

is the concern around specifying zero_point_domain multiple times? we could remove that and just infer the zero_point_domain from layout I think (the latter API)

airMeng: Since only XPU supports integer zp, can I move it in the next PR?
``` layout defines how we store the packed weights actually``` it should include the layout of scales and zeros, right?

Comment (jerryzh168) [reply torchao/dtypes/uintx/int4_xpu_layout.py:214]: > Since only XPU supports integer zp, can I move it in the next PR?

what is this referring to?

> layout defines how we store the packed weights actually it should include the layout of scales and zeros, right?

yeah that's correct, ideally I think we should not use layout to control whether we have packed weight / scale_and_zero / scale, zero, the duplication should actually happen in the tensor level (we create different tensor subclass tensors), not layout. feel free to go that route if want.

File: test/dtypes/test_affine_quantized.py (hunk 4 above)
Comment (airMeng) [reply test/dtypes/test_affine_quantized.py]: merged

File: torchao/dtypes/uintx/int4_xpu_layout.py (hunk 6 above)
Comment (airMeng) [reply torchao/dtypes/uintx/int4_xpu_layout.py:214]: Can I separate into different layouts, and bind the zero point domain into each layout in the next PR?

File: torchao/quantization/utils.py (hunk 9 above)
Comment (airMeng) [reply torchao/quantization/utils.py]: done

File: torchao/quantization/GPTQ.py (hunk 10 above)
Comment (airMeng) [reply torchao/quantization/GPTQ.py]: removed GPTQ support in this PR to simply the logic. I will open another PR after I seperate the layouts

File: torchao/prototype/hqq/hqq_tinygemm_linear.py [hunk 11]
```diff
@@ -166,6 +167,10 @@ def process_hqq_quants(self, W_q, meta):
             self.weight_int4pack = torch.ops.aten._convert_weight_to_int4pack_for_cpu(
                 W_q_torch, self.inner_k_tiles
             )
+        if is_device(W_q.device.type, "Xpu") and TORCH_VERSION_AT_LEAST_2_7:
```
jerryzh168, line 170: nit: "xpu"?

torchao/quantization/quant_api.py
```diff
@@ -798,6 +802,13 @@ class Int4WeightOnlyConfig(AOBaseConfig):
     zero_point_domain: Optional[ZeroPointDomain] = ZeroPointDomain.NONE
     set_inductor_config: bool = True
 
+    # mapping_type = MappingType.ASYMMETRIC
```
jerryzh168, line 805: nit: remove

torchao/quantization/quant_primitives.py
```diff
@@ -407,6 +407,8 @@ def _quantize_affine_no_dtype_cast(
     shape_after_reduction = shape_for_reduction
     for i in reduction_dims:
         shape_after_reduction[i] = 1
+    if shape_after_reduction[0] == 12288:
```
jerryzh168, line 410: remve?

torchao/quantization/quant_primitives.py
```diff
@@ -954,6 +956,7 @@ def _choose_qparams_affine(
             if preserve_zero:
                 zero_point = quant_min - torch.round(min_val_neg / scale)
                 zero_point = torch.clamp(zero_point, quant_min, quant_max)
+                zero_point_dtype = torch.int32
```
jerryzh168, line 959: why is this set here?

torchao/quantization/utils.py
```diff
@@ -315,7 +316,7 @@ def dequantize_per_channel(int_repr, scales, zero_points, out_dtype=torch.float3
     return dequantized
 
 
-def get_groupwise_affine_qparams(w, n_bit=4, groupsize=128, dtype=torch.bfloat16):
+def get_groupwise_affine_qparams(w, n_bit=4, groupsize=128, dtype=torch.bfloat16, zero_point_domain_is_int=False):
```
jerryzh168, line 319: nit: I'm wondering if we should just expose zero_point_domain as an arg directly here

torchao/quantization/quant_api.py
```diff
@@ -850,6 +860,7 @@ def _int4_weight_only_transform(
             zero_point_domain in LAYOUT_TO_ZERO_POINT_DOMAIN[type(layout)]
         ), f"Layout only support {LAYOUT_TO_ZERO_POINT_DOMAIN[layout]}"
 
+    preserve_zero = LAYOUT_TO_PRESERVE_ZEROS[type(layout)] if zero_point_domain!=ZeroPointDomain.INT else True
```
jerryzh168, line 863: does `zero_point_dtype` need to change after this is set

jerryzh168 (approved):
looks fine for now, but at some point we will probably add new tensor subclass tensors for these

torchao/quantization/quant_primitives.py
airMeng: In fact ````preserve_zero``` and INT zero point domain couples here, I think it is duplicated someway
The reason for setting this parameter as an int is that many places calling this function use the default floating parameter.

torchao/quantization/quant_api.py
airMeng: same as the above. In fact preserve_zero and INT zero point domain couples.

torchao/quantization/utils.py
airMeng: which way do you prefer? how about
```
def get_groupwise_affine_qparams(w, n_bit=4, groupsize=128, data_dtype=torch.bfloat16, scale_dtype=torch.bfloat16, zero_dtype=torch.bfloat16):
```

torchao/quantization/quant_api.py
jerryzh168: there are three things, preserve_zero = {True, False}, zero_point_domain = {FLOAT, INT, NONE} and zero_point_dtype = {float, int, ...}

it's true that not all combinations are valid, but I don't think they are coupled, see https://github.com/pytorch/ao/blob/64bcf4c25755a783685ba7383000b3bf722523c1/torchao/quantization/quant_primitives.py#L755-L770, zero_point_domain might be something we can remove though. cc @jainapurva

torchao/quantization/utils.py
jerryzh168: sure, this sounds good. what about preserve_zero and zero_point_domain? I don't think these can be fully inferred?

torchao/quantization/quant_primitives.py
jerryzh168: preserve_zero talks about whether zero (in original floating point domain) is exactly representable or not, it's not coupled with zero point domain I think, even zero is exactly representation, we can still choose zero_point_domain to be in float

airMeng: yes, from the math side they are not related. but the code here implies this, see the condition dispatch from Line954 to 966. We need a refactor here.

torchao/quantization/quant_api.py
airMeng: how about https://github.com/pytorch/ao/pull/1577/commits/946f530512eb91d4947a8488804148071d9051bb, expose it as an independent argument?

torchao/quantization/quant_primitives.py
```diff
@@ -951,13 +951,14 @@ def _choose_qparams_affine(
         if zero_point_domain == ZeroPointDomain.NONE.name:
             zero_point = None
         else:
-            if preserve_zero:
+            if zero_point_domain == ZeroPointDomain.INT.name:
```
jerryzh168, line 954: why changing this condition?

torchao/quantization/quant_primitives.py
jerryzh168: I think it would be better to do assert instead of changing condition if there is coupling?

torchao/quantization/quant_primitives.py
airMeng: with https://github.com/pytorch/ao/pull/1577/commits/946f530512eb91d4947a8488804148071d9051bb, ```preserve_zero``` and ```zero_point_domain``` has be decoupled. ```preserve_zero``` has already taken effects https://github.com/pytorch/ao/blob/946f530512eb91d4947a8488804148071d9051bb/torchao/quantization/quant_primitives.py#L903-L908

and the conditional checks here should all be based on the value of zero_point_domain.

jerryzh168: I see, OK, can you restructure the if statement a bit, to have:

```
if zero_point_domain == ZeroPointDomain.NONE.name:
   ...
elif zero_point_domain == ZeroPointDomain.INT.name:
   ...
elif zero_point_domain == ZeroPointDomain.FLOAT.name:
   ...
else:
    raise
```

torchao/quantization/quant_api.py
jerryzh168: yeah this makes sense

torchao/quantization/quant_primitives.py
airMeng: done in https://github.com/pytorch/ao/pull/1577/commits/269d7a34c76181525fe9236934a4c3384023690b

torchao/quantization/utils.py
jerryzh168: is this going to be updated? I feel it's a bit weird to introduce a boolean flag for zero_point_domain when we can just pass zero_point_domain itself around

torchao/quantization/utils.py
```diff
@@ -342,15 +343,14 @@ def get_groupwise_affine_qparams(w, n_bit=4, groupsize=128, dtype=torch.bfloat16
         eps,
         scale_dtype=scale_dtype,
         zero_point_dtype=zero_point_dtype,
-        preserve_zero=False,
-        zero_point_domain=ZeroPointDomain.FLOAT,
+        preserve_zero=zero_point_domain_is_int,
+        zero_point_domain=ZeroPointDomain.FLOAT if not zero_point_domain_is_int else ZeroPointDomain.INT
```
jerryzh168, line 347: this could just be zero_point_domain if you pass it around

airMeng: Could you check whether https://github.com/pytorch/ao/pull/1577/commits/287d069e6f88bcf832de0849b29d079e5fe52fe9 is okay?

torchao/quantization/utils.py
airMeng: https://github.com/pytorch/ao/pull/1577/commits/287d069e6f88bcf832de0849b29d079e5fe52fe9 is it okay?

torchao/quantization/utils.py
jerryzh168: yeah looks good, thanks!

airMeng:
@jerryzh168 could you trigger the workflow? Is it possible to grant the write access to this repo then I can trigger CI myself?

jerryzh168:
@airMeng just invited you as a contributor

airMeng:
@jerryzh168 if no more comment, I will squash and merge it.

torchao/quantization/quant_primitives.py
```diff
@@ -955,19 +955,19 @@ def _choose_qparams_affine(
         scale = torch.clamp(scale, min=eps)
         if zero_point_domain == ZeroPointDomain.NONE.name:
             zero_point = None
+        elif zero_point_domain == ZeroPointDomain.INT.name:
+            zero_point = quant_min - torch.round(min_val_neg / scale)
+            zero_point = torch.clamp(zero_point, quant_min, quant_max)
+            zero_point_dtype = torch.int32
```
jerryzh168, line 961: I still feel we probably don't want to set the dtype here?

test/quantization/test_quant_primitives.py
```diff
@@ -135,7 +136,9 @@ def _groupwise_affine_quantize_tensor_from_qparams(
         )
 
     if TORCH_VERSION_AT_LEAST_2_5:
-        if not (is_device(w.device.type, "cpu") and TORCH_VERSION_AT_LEAST_2_6):
+        if (not (is_device(w.device.type, "cpu") and TORCH_VERSION_AT_LEAST_2_6)) and (
```
jerryzh168, line 139: these are getting complicated and hard to maintain, maybe you can help with making these easier like encapsulate these conversion functions in a helper function somehow

torchao/quantization/quant_primitives.py
jerryzh168: we can merge after this one is changed to assert I think

airMeng: sure

test/quantization/test_quant_primitives.py
airMeng: could you check whether https://github.com/pytorch/ao/pull/1577/commits/613cad7b1f42805efa878561cf36b3e976761037 works?

torchao/quantization/quant_primitives.py
airMeng: could you check whether https://github.com/pytorch/ao/pull/1577/commits/613cad7b1f42805efa878561cf36b3e976761037 works?

torchao/utils.py
```diff
@@ -676,6 +676,18 @@ def is_sm_at_least_100():
     )
 
 
+def check_cpu_version(device, version="2.6.0"):
```
jerryzh168, line 679: nit: I feel a more descriptive name might be better, this is a bit vague

test/quantization/test_quant_primitives.py
```diff
@@ -744,8 +748,8 @@ def test_groupwise_affine_dequantize_tensor_from_qparams(self):
                 zeros = torch.randint(0, 15, (10, 2), dtype=torch.int32)
             if TORCH_VERSION_AT_LEAST_2_5:
                 input_tmp = input
-                if not (
-                    is_device(input.device.type, "cpu") and TORCH_VERSION_AT_LEAST_2_6
+                if (not (check_cpu_version(input.device))) and (
```
jerryzh168, line 751: my comment was actually meant to say that we can encapsulate these format changes to some function `input_tmp = (input[::, ::2] << 4 | input[::, 1::2]).to(torch.uint8)` so we don't need to have these across the codebase since it's pretty error prone and we are not sure when to use them

torchao/utils.py
jerryzh168: checking cuda is fine, but this check is specific for cpu and xpu right, and related to the change of the int4mm operator

jerryzh168: > checking cpu and torch version 2.8

when do we need this?

airMeng: https://github.com/pytorch/ao/blob/31f119e002c1bef8cd8fa91e5523ca5033a9ecdf/torchao/dtypes/uintx/int4_cpu_layout.py#L114-L122

sorry, should be 2_6 and 2_5

jerryzh168: this is getting too complicated I feel, can we just drop the support for some pytorch versions?

jerryzh168: it's fine if you want to keep the version number, but still it would be good to change the function name to make it clearer I think, also this doesn't have to block the PR, please feel free to merge and fix later

airMeng: > this is getting too complicated I feel, can we just drop the support for some pytorch versions?

there are still regression tests in current CI, as early as torch 2.3. But I agree if we should only support the latest version since AO is an experimental(?) innovation project

jerryzh168: yeah we're only committed to support the most recent 2 version of pytorch, but here I meant drop support for CPU layout for older pytorch versions and just keep one

PR #1578, opened 2025-01-17
[CPU] Fix registration of int4wo linear implementation on CPU

Xia-Weiwen:
**Summary**
Int4wo on CPU does not run into expected mm op (`torch.ops.aten._weight_int4pack_mm_for_cpu`). It seems to be a regression after some refactoring of related code. This PR fixes it by registering a linear impl for the `Int4CPULayout`, which calls `torch.ops.aten._weight_int4pack_mm_for_cpu` for computation. The new impl is enabled for torch>=2.6. The new impl does not require dtype to be bfloat16. It supports fp32, fp16, bf16 for both weight and activation.

**Test plan**
```
python test/quantization/test_quant_api.py -k test_int4wo_cpu
```

### DIFF
--- a/test/quantization/test_quant_api.py
+++ b/test/quantization/test_quant_api.py
@@ -773,9 +773,7 @@ class TestQuantFlow(TestCase):
         if x_dim == 3:
             example_inputs = (example_inputs[0].unsqueeze(0),)
 
-        with torch.no_grad(), torch.autocast(
-            "cpu", enabled=(dtype != torch.float), dtype=dtype
-        ):
+        with torch.no_grad():
             quantize_(m, int4_weight_only(group_size=32, layout=Int4CPULayout()))
             # ensure the expected op is in the code
             _, code = torch._inductor.utils.run_and_get_code(
@@ -783,6 +781,7 @@ class TestQuantFlow(TestCase):
                 *example_inputs,
             )
             assert "_weight_int4pack_mm_for_cpu" in code[0]
+            assert "aten.mm.default" not in code[0]
 
 
 class TestMultiTensorFlow(TestCase):
--- a/test/quantization/test_quant_api.py
+++ b/test/quantization/test_quant_api.py
@@ -762,24 +762,31 @@ class TestQuantFlow(TestCase):
         self.assertLess(memory_streaming, memory_baseline)
 
     @unittest.skipIf(not TORCH_VERSION_AT_LEAST_2_6, "Test only enabled for 2.6+")
-    def test_int4wo_cpu(self):
+    @common_utils.parametrize(
+        "dtype", [torch.float, torch.bfloat16, torch.half]
+    )
+    @common_utils.parametrize(
+        "x_dim", [2, 3]
+    )
+    def test_int4wo_cpu(self, dtype, x_dim):
         from torchao.dtypes import Int4CPULayout
 
         device = "cpu"
-        for dtype in (torch.float, torch.bfloat16, torch.half):
-            m = ToyLinearModel().eval().to(dtype).to(device)
-            example_inputs = m.example_inputs(dtype=dtype, device=device)
+        m = ToyLinearModel().eval().to(dtype).to(device)
+        example_inputs = m.example_inputs(dtype=dtype, device=device)
+        if x_dim == 3:
+            example_inputs = (example_inputs[0].unsqueeze(0),)
 
-            with torch.no_grad(), torch.autocast(
-                "cpu", enabled=(dtype != torch.float), dtype=dtype
-            ):
-                quantize_(m, int4_weight_only(group_size=32, layout=Int4CPULayout()))
-                # ensure the expected op is in the code
-                _, code = torch._inductor.utils.run_and_get_code(
-                    torch.compile(m, fullgraph=True, dynamic=True),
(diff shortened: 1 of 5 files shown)


test/quantization/test_quant_api.py
```diff
@@ -761,6 +761,29 @@ def reset_memory():
...
+    def test_int4wo_cpu(self, dtype, x_dim):
+        from torchao.dtypes import Int4CPULayout
+
+        device = "cpu"
+        m = ToyLinearModel().eval().to(dtype).to(device)
+        example_inputs = m.example_inputs(dtype=dtype, device=device)
+        if x_dim == 3:
+            example_inputs = (example_inputs[0].unsqueeze(0),)
+
+        with torch.no_grad(), torch.autocast(
```
leslie-fang-intel, line 776: Since the model and inputs have already been converted to target data type, why we still need to enable autocast here?

torchao/dtypes/uintx/tensor_core_tiled_layout.py
```diff
@@ -76,14 +75,9 @@ def _linear_bf16_act_uint4_weight_impl(input_tensor, weight_tensor, bias):
...
     groupsize = weight_tensor.block_size[1]
-    if is_device(input_tensor.device.type, "cpu") and TORCH_VERSION_AT_LEAST_2_6:
-        y = torch.ops.aten._weight_int4pack_mm_for_cpu(
-            act_mat.contiguous(), packed_weight, groupsize, scale_and_zero
-        )
-    else:
-        y = torch.ops.aten._weight_int4pack_mm(
-            act_mat.contiguous(), packed_weight, groupsize, scale_and_zero
-        )
+    y = torch.ops.aten._weight_int4pack_mm(
```
leslie-fang-intel, line 78: why we change `_weight_int4pack_mm_for_cpu` to `_weight_int4pack_mm`? I remember `_weight_int4pack_mm` didn't register for CPU.

sanchitintel: @Xia-Weiwen moved the CPU implementation to `torchao/dtypes/uintx/int4_cpu_layout.py`, where support for more activation dtypes has also been added.

This code now pertains to CUDA.

sanchitintel:
> Int4wo on CPU does not run into expected mm op (`torch.ops.aten._weight_int4pack_mm_for_cpu`)

The FX IR pattern you shared with me after running a toy model did have a call to `torch.ops.aten._weight_int4pack_mm_for_cpu`, but the whole pattern corresponding to it was very weird (even had a aten.mm call at the end), so something is indeed broken.
If we had run an LLM with torchchat with int4 WoQ & simply searched for `torch.ops.aten._weight_int4pack_mm_for_cpu` in the FX graph of the model, we may probably not have discovered this issue.

Is it possible to add a UT that uses `torch.compile`, and also somehow checks if the pattern is as expected? 

Thanks!

test/quantization/test_quant_api.py
Xia-Weiwen: Thanks. `autocast` is removed. The `_weight_int4pack_mm_for_cpu` is not registered for autocast.

torchao/dtypes/uintx/tensor_core_tiled_layout.py
Xia-Weiwen: The explanation from @sanchitintel is correct. This part of code would never be run on CPU because the layout is for CUDA only.

Xia-Weiwen:
> > Int4wo on CPU does not run into expected mm op (`torch.ops.aten._weight_int4pack_mm_for_cpu`)
> 
> The [FX IR pattern](https://gist.github.com/sanchitintel/8f6ab270daf9e008338866fa49184f78) you shared with me after running a toy model did have a call to `torch.ops.aten._weight_int4pack_mm_for_cpu`, but the whole pattern corresponding to it was very weird (even had a aten.mm call at the end), so something is indeed broken. If we had run an LLM with torchchat with int4 WoQ & simply searched for `torch.ops.aten._weight_int4pack_mm_for_cpu` in the FX graph of the model, we may probably not have discovered this issue.
> 
> Is it possible to add a UT for a small model that uses `torch.compile`, and also somehow checks if the pattern is as expected? Perhaps, the corresponding UT in `test/integration/test_integration.py` could be modified.
> 
> Thanks!

We found that the weird graph is generated because the reference path is used for computation, i.e., dequantize weight -> aten.linear. With this PR, the graph is what we expect now. And I have also updated the UT to ensure `aten.mm` won't appear in the graph.

Xia-Weiwen:
cc. @yanbing-j @mingfeima @Valentine233 I tried to add you as viewers but your names didn't appear in the list.

mingfeima (approved):

mingfeima:
> cc. @yanbing-j @mingfeima @Valentine233 I tried to add you as viewers but your names didn't appear in the list.

we are not on the list of ao :)

leslie-fang-intel (approved):

Xia-Weiwen:
Hi @jerryzh168 @gau-nernst @HDCharles @alexsamardzic Could you please review this PR? Thanks.

drisspg (approved):

PR #1585, opened 2025-01-17
[rfc] enable direct configuration in quantize_

vkuzo:
# summary

This PR is a POC for passing per-workflow arguments to `quantize_` directly, without wrapping them in a `Callable`.  

High level motivation: passing direct configuraton is intuintive and widely used in similar contexts across various projects.  Passing configuration wrapped in a callable is IMO not intuitive, hard to understand and debug, and we have evidence that it pushes a significant portion of users from building on top of torchao.

# user facing API proposed changes

## signature of quantize_

```python
#
# before
#
def quantize(
    model: torch.nn.Module,
    ...,
): ...

#
# after - intermediate state (for the time that we need to keep the old syntax around)
# we need to clarify whether the intermediate state is needed for 1 to 2 releases, or if we should
# just go directly from before to after
#
def quantize(
    model: torch.nn.Module,
    apply_tensor_subclass: Union[Callable[[torch.nn.Module], torch.nn.Module], AOBaseWorkflowConfig],
    ...,
): ...

#
# after - long term state
#
def quantize(
    model: torch.nn.Module,
    config: AOBaseWorkflowConfig,
    ...,
): ...
```

## usage example

An example for `int4_weight_only`

```python
#
# before
#
quantize_(m, int4_weight_only(group_size=32))

#
# after, with new user facing names
#
quantize_(m, Int4WeightOnlyWorkflowConfig(group_size=32))

#
# AND, after, with BC names
#
quantize_(m, int4_weight_only(group_size=32))
```

# developer facing proposed changes

See the PR details for examples, but they can be summarized as:

```python
#
# old
#

# quantize_ calls the instance of calling this function on each module of the model
def int4_weight_only(group_size: int, ...) -> Callable:

    def new_callable(weight: torch.Tensor):
        # configuration is captured here via local variables
        ...
        
    # return type is a Callable
    return _get_linear_subclass_inserter(new_callable)

#
# new
#

# config base class
class AOBaseWorkflowConfig(abc.ABC):
    pass

# user facing configuration of a workflow
@dataclass
class Int4WeightOnlyWorkflowConfig(AOBaseWorkflowConfig):
    group_size: int = 128
    ...

# not user facing transform of a module according to a worfklow's configuration
@register_quantize_module_handler(Int4WeightOnlyWorkflowConfig)
def _int4_weight_only_transform(
    module: torch.nn.Module, 
    config: UserFacingWorkflowConfig,
) -> torch.nn.Module:
    # map to AQT, not user facing
    ...
```

## current status

At this point this is a POC, and I've shown how it can work on three user facing workflows:
* PTQ's `int4_weight_only`
* QAT's `intx_quantization_aware_training` and `from_intx_quantization_aware_training`

# next steps

discuss more broadly

Test Plan:

```
pytest test/quantization/test_quant_api.py -s -x -k test_int4_weight_only_numerics
pytest test/quantization/test_qat.py -s -x -k test_quantize_api_standalone
pytest test/quantization/test_qat.py -s -x -k test_quantize_api_convert_path
```

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/quantization/qat/api.py
+++ b/torchao/quantization/qat/api.py
@@ -9,11 +9,15 @@ from typing import Any, Callable, List, Optional, Union
 
 import torch
 
+from torchao.core.config import AOBaseWorkflowConfig
 from torchao.quantization.granularity import (
     Granularity,
     PerAxis,
     PerGroup,
     PerToken,
+)
+from torchao.quantization._transform_module import (
+    register_quantize_module_handler,
 )
 from torchao.quantization.quant_primitives import (
     _SUB_BYTE_INT_BOUNDS,
@@ -239,12 +243,26 @@ class FakeQuantizeConfig:
             super().__setattr__(name, value)
 
 
-def intx_quantization_aware_training(
-    activation_config: Optional[FakeQuantizeConfig] = None,
-    weight_config: Optional[FakeQuantizeConfig] = None,
-) -> Callable:
+@dataclass
+class IntXQuantizationAwareTrainingWorkflowConfig(AOBaseWorkflowConfig):
+    activation_config: Optional[FakeQuantizeConfig] = None
+    weight_config: Optional[FakeQuantizeConfig] = None
+
+
+# for BC
+intx_quantization_aware_training = IntXQuantizationAwareTrainingWorkflowConfig
+
+
+@register_quantize_module_handler(IntXQuantizationAwareTrainingWorkflowConfig)
+def _intx_quantization_aware_training_transform(
+    module: torch.nn.Module,
+    config: IntXQuantizationAwareTrainingWorkflowConfig,
+) -> torch.nn.Module:
     """
-    Return a function that applies fake quantization to a `torch.nn.Module`.
+    THIS IS NOT A PUBLIC API - any usage of this outside of torchao
+    can break at any time.
+
+    Apply fake quantization to a `torch.nn.Module`.
     to be used with :func:`~torchao.quantization.quant_api.quantize_`.
 
     Example usage::
@@ -267,37 +285,32 @@ def intx_quantization_aware_training(
     `torch.nn.Embedding` with an activation config, then we will raise
     ValueError as these are not supported.
     """
+    from .embedding import FakeQuantizedEmbedding
+    from .linear import FakeQuantizedLinear
 
-    def _insert_fake_quantize(mod: torch.nn.Module):
-        """
-        Swap the given module with its corresponding fake quantized version.
-        """
-        from .embedding import FakeQuantizedEmbedding
-        from .linear import FakeQuantizedLinear
+    mod = module
+    activation_config = config.activation_config
+    weight_config = config.weight_config
 
-        if isinstance(mod, torch.nn.Linear):
-            return FakeQuantizedLinear.from_linear(
-                mod,
-                activation_config,
-                weight_config,
-            )
-        elif isinstance(mod, torch.nn.Embedding):
-            if activation_config is not None:
-                raise ValueError(
-                    "Activation fake quantization is not supported for embedding"
-                )
-            return FakeQuantizedEmbedding.from_embedding(mod, weight_config)
-        else:
+    if isinstance(mod, torch.nn.Linear):
+        return FakeQuantizedLinear.from_linear(
+            mod,
+            activation_config,
+            weight_config,
+        )
+    elif isinstance(mod, torch.nn.Embedding):
+        if activation_config is not None:
             raise ValueError(
-                "Module of type '%s' does not have QAT support" % type(mod)
+                "Activation fake quantization is not supported for embedding"
             )
+        return FakeQuantizedEmbedding.from_embedding(mod, weight_config)
+    else:
+        raise ValueError("Module of type '%s' does not have QAT support" % type(mod))
 
-    return _insert_fake_quantize
 
-
-def from_intx_quantization_aware_training() -> Callable:
+class FromIntXQuantizationAwareTrainingWorkflowConfig(AOBaseWorkflowConfig):
     """
-    Return a function that converts a model with fake quantized modules,
+    Object that knows how to convert a model with fake quantized modules,
     such as :func:`~torchao.quantization.qat.linear.FakeQuantizedLinear`
     and :func:`~torchao.quantization.qat.linear.FakeQuantizedEmbedding`,
     back to model with the original, corresponding modules without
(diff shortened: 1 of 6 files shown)


drisspg:
Is there someone on the quanty/core side of things that can chime in on this API cc @Andrew Or

Tbh this feels almost too similar to the original since this isn't a Pure dataclass and has the _transform func.

I imagine people want:
```Python
quantize_(model, THE_CONFIG_TYPE(weight_scheme="f38", a_scheme"a12", ...)"
```

FWIW : https://github.com/pytorch/ao/blob/1240b19fd719d54af64c2d4d8b5cc33aba345dce/scripts/hf_eval.py#L68-L147

I think this code is a good example of the downsides of the current approach. This has high cognitive load IMO and is hard to tell what exactly is being chosen when. I think a good sign that the new API is better if it also could clean up this type of code

vkuzo:
> Is there someone on the quanty/core side of things that can chime in on this API cc @andrew Or
> 
> Tbh this feels almost too similar to the original since this isn't a Pure dataclass and has the _transform func.
> 
> I imagine people want:
> 
> ```python
> quantize_(model, THE_CONFIG_TYPE(weight_scheme="f38", a_scheme"a12", ...)"
> ```
> 
> FWIW :
> 
> https://github.com/pytorch/ao/blob/1240b19fd719d54af64c2d4d8b5cc33aba345dce/scripts/hf_eval.py#L68-L147
> 
> I think this code is a good example of the downsides of the current approach. This has high cognitive load IMO and is hard to tell what exactly is being chosen when. I think a good sign that the new API is better if it also could clean up this type of code

So, I'm actually highly in favor of moving the `transform` function out of user facing land to keep the config clean.  However, we don't have consensus on this on the team, with @HDCharles  liking the colocation.  I'm a big fan of easy-to-think-about incremental changes moving in the right direction, so currently I'm just punting this discussion till later, and "enforcing" that by keeping the `transform` function private with an underscore and "scary" docblocks.

If in the future we moved transforms out, I'd be even happier.

andrewor14:
> So, I'm actually highly in favor of moving the transform function out of user facing land to keep the config clean. However, we don't have consensus on this on the team, with @HDCharles liking the colocation.

I think another way of address this is to just register a handler for all configs we want to handle in `quantize_`. E.g.

```
def quantize_(model: nn.Module, config: BaseConfig, ...existing args):
    if type(config) in _QUANTIZE_CONFIG_HANDLER:
        _QUANTIZE_CONFIG_HANDLER[type(config)](model, config)
    else:
        raise ValueError("Unknown config type")

@register_quantize_handler(AffineQuantizationConfig)
def _do_affine_quantization(model: nn.Module, config: AffineQuantizationConfig)

@register_quantize_handler(QATConfig)
def _do_qat(model: nn.Module, config: QATConfig)
```

This will keep `quantize_` small, and we won't need a separate `transform_fn` argument so the signature will be simpler. Configs will also be simpler and just be configs; they don't need to know anything about how they will be used.

torchao/quantization/quant_api.py
```diff
@@ -684,14 +708,10 @@ def gemlite_uintx_weight_only(
...
 
 
-def int4_weight_only(
-    group_size=128,
-    layout=TensorCoreTiledLayout(inner_k_tiles=8),
-    use_hqq=False,
-    zero_point_domain=None,
-):
+@dataclass
+class int4_weight_only(ao_base_workflow_config):
```
andrewor14, line 712: What do you think about having `int4_weight_only` be a function that returns a config instead:
```
def int4_weight_only(group_size: int):
    return AffineQuantizationConfig(
        weight_dtype=torch.int4,
        weight_group_size=group_size,
        ...
        activation_dtype=None,
    )
```

Then we only need one config class for AQT in general instead of one per quant setting. Also the user can still call `quantize_(model, int4_weight_only(group_size=32))` exactly the same way as before.

andrewor14: Update: decided to do 1 class per quant setting offline

vkuzo:
summary of offline discussion:
1. will change to a registration API for transforms, as suggested by @andrewor14 
2. will describe BC plan in more detail
3. will get some user feedback

torchao/quantization/quant_api.py
```diff
@@ -468,7 +480,10 @@ def insert_subclass(lin):
 
 def quantize_(
     model: torch.nn.Module,
-    apply_tensor_subclass: Callable[[torch.nn.Module], torch.nn.Module],
+    # apply_tensor_subclass: Callable[[torch.nn.Module], torch.nn.Module],
+    apply_tensor_subclass: Union[
+        Callable[[torch.nn.Module], torch.nn.Module], AOBaseWorkflowConfig
```
andrewor14, line 485: I think there are two BC breaking steps:
1. Change the arg name to "config"
2. Change the arg type to `AOBaseWorkflowConfig` only

Based on my quick search I feel we can just do (1) now. I haven't found any use cases yet that call this with a keyword argument, so this seems relatively safe. For (2) maybe we should deprecate with warning first and remove after a couple releases? I feel this use case (users passing in arbitrary functions) is harder to grep for.

Curious what others think as well @drisspg @HDCharles

torchao/quantization/quant_api.py
```diff
@@ -684,14 +720,10 @@ def gemlite_uintx_weight_only(
...
 
 
-def int4_weight_only(
-    group_size=128,
-    layout=TensorCoreTiledLayout(inner_k_tiles=8),
-    use_hqq=False,
-    zero_point_domain=None,
-):
+@dataclass
+class Int4WeightOnlyWorkflowConfig(AOBaseWorkflowConfig):
```
andrewor14, line 724: Looks great. I feel eventually we do want an intermediate `AffineQuantizationConfig` that lets users set the different dtypes, symmetric, dynamic etc, but I think we can add that later

torchao/quantization/quant_api.py
```diff
@@ -711,59 +743,84 @@ def int4_weight_only(
...
+# for BC
+# TODO maybe change other callsites
+int4_weight_only = Int4WeightOnlyWorkflowConfig
+
+
+@register_quantize_module_handler(Int4WeightOnlyWorkflowConfig)
+def _int4_weight_only_transform(
+    module: torch.nn.Module, config: Int4WeightOnlyWorkflowConfig
+) -> torch.nn.Module:
+    # TODO(future PR): perhaps move this logic to a different file, to keep the API
```
andrewor14, line 761: agree with this. We should move all configs and associated transforms to a separate file

andrewor14 (commented):
New proposal looks great to me!

torchao/quantization/qat/api.py
```diff
@@ -239,12 +243,26 @@ def __setattr__(self, name: str, value: Any):
             super().__setattr__(name, value)
 
 
-def intx_quantization_aware_training(
-    activation_config: Optional[FakeQuantizeConfig] = None,
-    weight_config: Optional[FakeQuantizeConfig] = None,
-) -> Callable:
+@dataclass
+class IntXQuantizationAwareTrainingWorkflowConfig(AOBaseWorkflowConfig):
```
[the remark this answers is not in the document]
vkuzo: I think keeping the base class as `AOBaseWorkflowConfig` and removing `Workflow` from child classes could work here

vkuzo:
moving development to https://github.com/pytorch/ao/pull/1595 to move to stacked PRs

PR #1592, opened 2025-01-21
Sparsity getting started docs

jainapurva:
Sparsity section in getting started


docs/source/sparsity.rst
```diff
@@ -1,4 +1,920 @@
...
+
+.. code-block:: py
+
+   from torchao.sparsity.sparse_api import sparsify_
+   from torchao.prototype.sparsity.superblock.blocksparse import block_sparse_weight
+
+   model = model.cuda()
+   sparsify_(model, block_sparse_weight())
+
+Goal
```
jcaip, line 182: I think everything before this can stay in the README, everything below makes sense for docs.

jainapurva: Ok. I think we can still add the supported APIs in api ref, even the prototype ones. #1590

jcaip (approved):

docs/source/index.rst
```diff
@@ -81,25 +82,32 @@ Welcome to the torchao Documentation
...
+   getting-started
 
 .. toctree::
    :glob:
    :maxdepth: 1
-   :caption: Contributor Guide
+   :caption: Developer Notes
 
    contributor_guide
+   sparsity.rst
```
andrewor14, line 95: I think you don't need .rst here

Issue #1501, opened 2025-01-06
Bug: MXLinear backward pass implementation


balancap:
In the current implementation of the `MXLinear` layer:
```python
def forward(self, x):
       x_mx = MXTensor.to_mx(x, self.elem_dtype, self.block_size)
       w_mx = MXTensor.to_mx(self.weight, self.elem_dtype, self.block_size)
       y = F.linear(x_mx, w_mx, self.bias)
       y = NoopFwToMXBw.apply(y, self.elem_dtype, self.block_size)
       return y
```
there is only a single MX quantization step of the output gradient (in `NoopFwToMXBw`).

However, following the MX microscaling paper, there should be 4 quantizations happening: two for output gradient (on 2 different axes), one for the activation and one for the weights (different from the forward ones). 
![microscaling-fwd-bwd](https://github.com/user-attachments/assets/60421c7d-8709-41c1-aa84-3955567d4d1b)

**Why does it matter:** even though not officially confirmed by hardware vendors, it is clear that MX matmuls can only be fully optimized if the quantization axis correspond to the reduction axis for both operands. Hence, running MX backward pass on next gen hardware will require the 4 quantization steps presented above. Changing of axis for the MX quantization result in a different quantization error, meaning that the current implementation is potentially not giving a full picture of what will be MX training on real hardware.

**Potential fix**: I believe we need a full implementation of forward+backward pass of `blockwise_quantize_linear` function, manually handling the backward pass quantization steps.

vkuzo:
Hi @balancap , https://github.com/pytorch/ao/issues/932 is also related.  This is a KP and I agree with your issue summary.  I also agree that we'll have to have a `torch.autograd.Function` which specifies how the quantization is done for all the 3 gemms in fwd/bwd.  

As soon as the first hardware vendor releases the official specs, we on the torchao core team are going to sprint on updating this code to match the final hardware support.  We would welcome contributions if anyone is looking to do this update earlier!

vkuzo:
I put up https://github.com/pytorch/ao/pull/1593 to address this in emulation, and we'll follow up again once we know more about the hardware.

balancap:
Thanks @vkuzo for the fix. Any chance we could have a new release (minor?) of TorchAO integrating it?

balancap:
[...]

Issue #932, opened 2024-09-24
[Question] Difference in MXLinear vs MXInferenceLinear grouping direction


Abhijit-2592:
In [MXLinear](https://github.com/pytorch/ao/blob/26e790df61e23b2ba340c36b84eb9940fec100bb/torchao/prototype/mx_formats/mx_linear.py#L63) layer it seems like we are directly passing the `self.weight (out_features, in_features)` to the `MXTensor.to_mx()` But in [MXInferenceLinear](https://github.com/pytorch/ao/blob/26e790df61e23b2ba340c36b84eb9940fec100bb/torchao/prototype/mx_formats/mx_linear.py#L86) we are passing `self.weight.t() (in_features, out_features)`. Doesn't this affect the grouping within the [MXTensor](https://github.com/pytorch/ao/blob/26e790df61e23b2ba340c36b84eb9940fec100bb/torchao/prototype/mx_formats/mx_tensor.py#L80) AFAIU, we reshape the tensor to `view(-1, group_size)` and take `torch.amax(tensor, 1)` to calculate the MXTensor. Wouldn't the max values change like below?

<img width="611" alt="Screenshot 2024-09-24 at 1 00 25 PM" src="https://github.com/user-attachments/assets/61614dd5-80e5-4814-ba78-1518771e78a2">

vkuzo:
Hi @Abhijit-2592 ,

In `MXInferenceLinear`, the relevant code snippet is:

```
        new_mod.weight_mx = MXTensor.to_mx(
            mod.weight.t().contiguous(), elem_dtype, block_size=block_size
        ).t()
```

Source: https://github.com/pytorch/ao/blob/26e790df61e23b2ba340c36b84eb9940fec100bb/torchao/prototype/mx_formats/mx_linear.py#L86

There are two calls to `t()` in this snippet (one on line 2, one on line 3), so it should work correctly, this code changes the memory layout without changing the tensor sizes.  Please reopen if that does not explain your question.

balaabhijit:
Hi @vkuzo thanks for your prompt response. It doesn't seem like they are exactly equivalent. In the above code you have provided, you `t()` before passing to `MXTensor.to_mx()` and then transpose it back before saving it as `new_mod.weight_mx` this ensures the shapes are right but I don't think the elements exactly match. Could you please clarify? I have also provided the snipper for your reference

```python

import torch
print(torch.__version__)
_ = torch.manual_seed(1234)

import torchao
print(torchao.__version__)
[...]

PR #1593, opened 2025-01-21
mx cleanup [2/x]: refactor mx gemm

vkuzo:
Summary:

Refactors the MX gemm emulation code to properly emulate the memory layout
constraints we expect from the future mx-enabled hardware, where we
expect:
* the first argument to the mx gemm to be required row-major memory
  format
* the second argument to the mx gemm to be required col-major memory
  format

Fixes https://github.com/pytorch/ao/issues/1501 and https://github.com/pytorch/ao/issues/932

Note that two morally unrelated issues were uncovered with this
refactor:
1. when autocast is on, compile is no longer matching eager numerics.
   Since the "before this PR" state isn't really representative of the
   world, I'm treating this as a newly uncovered issue, and we can fix
   it in a future PR.
2. our transpose logic for fp4 packed into two elements per byte doesn't
   work for tensors of shape (M, 1), because we currently rely on the
   `is_contiguous()` function to see if our tensor was transposed. We
   could work around, but punting that until a time that becomes
   important. I expect most tensors in real world usage with MX to not
   hit this case.

Test Plan:

```
```

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo:
* (to be filled)

drisspg (approved):

PR #1595, opened 2025-01-22
[bc-breaking] enable direct configuration in quantize_

vkuzo:
# summary

This PR enables passing per-workflow arguments to `quantize_` directly, without wrapping them in a `Callable`.  

Motivation: passing direct configuraton is intuintive and widely used in similar contexts across various projects.  Passing configuration wrapped in a callable is IMO not intuitive, hard to understand and debug, and we have evidence that it pushes a portion of users from building on top of torchao.
 
We will keep the old callable syntax supported by `quantize_` for one release cycle, and delete it afterwards.  We will keep the old names as aliases for new names going forward (example: `int4_weight_only` as an alias of `Int4WeightOnlyConfig`) to keep existing callsites working without changes.

# user facing API changes

## signature of quantize_

```python
#
# before
#
def quantize(
    model: torch.nn.Module,
    ...,
): ...

#
# after - intermediate state, support both old and new for one release
#
def quantize(
    model: torch.nn.Module,
    config: Union[AOBaseConfig, Callable[[torch.nn.Module], torch.nn.Module]],
    ...,
): ...

#
# after - long term state
#
def quantize(
    model: torch.nn.Module,
    config: AOBaseConfig,
    ...,
): ...
```

## usage example

An example for `int4_weight_only`

```python
#
# before
#
quantize_(m, int4_weight_only(group_size=32))

#
# after, with new user facing names
#
quantize_(m, Int4WeightOnlyConfig(group_size=32))

#
# AND, after, with BC names
#
quantize_(m, int4_weight_only(group_size=32))
```

# developer facing changes

See the PR details for examples, but they can be summarized as:

```python
#
# old
#

# quantize_ calls the instance of calling this function on each module of the model
def int4_weight_only(group_size: int, ...) -> Callable:

    def new_callable(weight: torch.Tensor):
        # configuration is captured here via local variables
        ...
        
    # return type is a Callable
    return _get_linear_subclass_inserter(new_callable)

#
# new
#

# config base class
class AOBaseConfig(abc.ABC):
    pass

# user facing configuration of a workflow
@dataclass
class Int4WeightOnlyConfig(AOBaseConfig):
    group_size: int = 128
    ...

# not user facing transform of a module according to a worfklow's configuration
@register_quantize_module_handler(Int4WeightOnlyConfig)
def _int4_weight_only_transform(
    module: torch.nn.Module, 
    config: Int4WeightOnlyConfig,
) -> torch.nn.Module:
    # map to AQT, not user facing
    ...
```

# current status

The current PR migrates three user facing workflows:
* PTQ's `int4_weight_only`
* QAT's `intx_quantization_aware_training` and `from_intx_quantization_aware_training`

I've chosen to migrate one PTQ and two QAT workflows to prove generality of the new flow, but avoid a high LOC in this PR to make it easier to review.  We will migrate the rest of the workflows in future PRs, detailed below:
* int8_dynamic_activation_int4_weight
* int8_dynamic_activation_int8_weight
* int8_dynamic_activation_int8_semi_sparse_weight
* int8_weight_only
* float8_weight_only
* float8_dynamic_activation_float8_weight
* float8_static_activation_float8_weight
* uintx_weight_only
* fpx_weight_only
* gemlite_uintx_weight_only
* callsites from the prototype folder

After a release cycle, we will delete the old callable syntax.

Test Plan:

```
pytest test/quantization/test_quant_api.py -s -x -k test_int4_weight_only_numerics
pytest test/quantization/test_qat.py -s -x -k test_quantize_api_standalone
pytest test/quantization/test_qat.py -s -x -k test_quantize_api_convert_path
```

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo:
* __->__ #1595

torchao/core/config.py
```diff
@@ -0,0 +1,13 @@
+import abc
+
+
+# directory location for this might need more polish
+class AOBaseWorkflowConfig(abc.ABC):
```
drisspg, line 5: Super Nit: maybe just AOBaseConfig

torchao/quantization/_transform_module.py
```diff
@@ -0,0 +1,17 @@
+from typing import Callable, Dict
+
+import torch
+
+from torchao.core.config import AOBaseWorkflowConfig
+
+_QUANTIZE_CONFIG_HANDLER: Dict[
```
drisspg, line 7: alternative we could use `singledispatch` do this: _QUANTIZE_CONFIG_HANDLER: Dict[
    AOBaseWorkflowConfig,
    Callable[[torch.nn.Module, AOBaseWorkflowConfig], torch.nn.Module],
] = {}


def register_quantize_module_handler(config_type):
    def decorator(func):
        _QUANTIZE_CONFIG_HANDLER[config_type] = func

    return decorator
    
    
    which would make quantize_ generic and then we register instantations baed off of config type

torchao/quantization/_transform_module.py
```diff
@@ -0,0 +1,17 @@
...
+    AOBaseWorkflowConfig,
+    Callable[[torch.nn.Module, AOBaseWorkflowConfig], torch.nn.Module],
+] = {}
+
+
+def register_quantize_module_handler(config_type):
+    def decorator(func):
```
drisspg, line 14: if we want to keep it this way maybe lets add a functools.wraps

torchao/core/config.py
vkuzo: sgtm

torchao/quantization/_transform_module.py
vkuzo: if the first argument of `quantize_` was the configuration, I would see the benefit in using `singledispatch`.  The first argument is currently the `model` though.

drisspg (approved):

test/dtypes/test_affine_quantized.py
```diff
@@ -180,8 +187,13 @@ def apply_uint6_weight_only_quant(linear):
...
     @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
     def test_print_quantized_module(self, apply_quant):
+        print(apply_quant)
         linear = torch.nn.Linear(128, 256, dtype=torch.bfloat16, device="cuda")
-        ql = apply_quant(linear)
+        if isinstance(apply_quant, AOBaseConfig):
+            quantize_(linear, apply_quant)
+            ql = linear
+        else:
+            ql = apply_quant(linear)
```
andrewor14, line 196: once we migrate all functions to configs we won't need this check anymore right? Should we add a TODO to remove it?

torchao/core/config.py
```diff
@@ -0,0 +1,10 @@
+import abc
```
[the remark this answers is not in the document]
andrewor14: slightly stronger preference is I feel "core" shouldn't appear in the import, so users should be able to do this:
```
from torchao.config import AOBaseConfig
```
but we can do that by adding this to `__init__.py`

torchao/quantization/qat/api.py
```diff
@@ -315,22 +328,31 @@ def from_intx_quantization_aware_training() -> Callable:
         )
```
andrewor14, line 328: need to update the docstring here in the previous line

torchao/quantization/qat/api.py
```diff
@@ -269,37 +287,32 @@ def intx_quantization_aware_training(
     `torch.nn.Embedding` with an activation config, then we will raise
```
andrewor14, line 287: I can't comment up there but need to update the docstring in L282

torchao/core/config.py
```diff
@@ -0,0 +1,10 @@
...
+
+
+class AOBaseConfig(abc.ABC):
+    """
+    If a workflow config inherits from this then `quantize_` knows
+    how to a apply it to a model.
+    """
```
andrewor14, line 8: should we add a paragraph here or under quantize_ about how this is related to `register_quantize_module_handler`, so users who wish to add their own configs know how to do it?

torchao/quantization/quant_api.py
```diff
@@ -507,39 +520,36 @@ def quantize_(
...
-    )
+    if isinstance(config, AOBaseConfig):
+        handler = _QUANTIZE_CONFIG_HANDLER[type(config)]
+        # for each linear in the model, apply the transform if filtering passes
+        _replace_with_custom_fn_if_matches_filter(
+            model,
+            handler,
+            _is_linear if filter_fn is None else filter_fn,
+            device=device,
+            extra_args=(config,),
```
andrewor14, line 535: alternatively we can pass in a lambda, then we don't need to add `extra_args` or pass in `config`:
```
replace_fn = lambda mod: handler(mod, config)
```
seems simpler

vkuzo: I'm really not a fan of passing callables around, it's easy when the callable is simple but easy for future people to tack ugly stuff on and increase complexity.  Non-callable args make it harder to make the code ugly in the future.

andrewor14: oh sorry, I meant pass in `replace_fn` instead of `handler`, like:
```
replace_fn = lambda mod: handler(mod, config)
_replace_with_custom_fn_if_matches_filter(
            model,
            replace_fn,
            _is_linear if filter_fn is None else filter_fn,
            device=device,
)
```
either way you're passing a callable

vkuzo: hmm, still not a fan of `replace_fn = lambda mod: handler(mod, config)`.  This changes `replace_fn` from a stateless callable to a stateful callable, where the state is hard to inspect.  It's less LOC but harder to debug IMO.

test/quantization/test_qat.py
```diff
@@ -1185,7 +1185,7 @@ def test_qat_prototype_bc(self):
     @unittest.skipIf(
         not TORCH_VERSION_AT_LEAST_2_4, "skipping when torch version is 2.4 or lower"
     )
-    def test_quantize_api(self):
+    def test_quantize_api_standalone(self):
```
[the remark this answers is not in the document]
vkuzo: it's convenient from being able to filter for only this test from the commandline.  I can remove it if you'd like.

test/dtypes/test_affine_quantized.py
```diff
@@ -195,7 +208,11 @@ def test_flatten_unflatten(self, device, dtype):
         apply_quant_list = get_quantization_functions(False, True, device)
         for apply_quant in apply_quant_list:
             linear = torch.nn.Linear(128, 256, dtype=dtype, device=device)
-            ql = apply_quant(linear)
+            if isinstance(apply_quant, AOBaseConfig):
+                quantize_(linear, apply_quant)
+            else:
+                # TODO(#1690): delete this once config migration is done
+                ql = apply_quant(linear)
```
msaroufim, line 215: have a few partners where we need to forward fix BC issues including HuggingFace transformers, Optimimum, SGLang and Diffusers

vkuzo: @msaroufim  do you have a link?

I don't expect any BC breakages of people using the `quantize_` API as specified in the docs.  The BC breaking change would be if people are applying their transform on linear layers directly, without using `quantize_`.

vkuzo: HF callsite: https://github.com/huggingface/transformers/blob/1feebb5b4150882deabddd190a541f336f3be817/src/transformers/quantizers/quantizer_torchao.py#L199

SGLANG callsite: https://github.com/sgl-project/sglang/blob/2f47d710ae9cb1bdbbe0fe2392a0634827d257b3/python/sglang/srt/layers/torchao_utils.py#L39

Diffusers callsite: https://github.com/huggingface/diffusers/blob/7fb481f840b5d73982cafd1affe89f21a5c0b20b/src/diffusers/quantizers/torchao/torchao_quantizer.py#L234

we should definitely test these, but they look like they will be unaffected to me

PR #1599, opened 2025-01-22
Add Float8QuantizedTensor (AQT subclass) and replace to_affine_quantized_floatx with to_affine_quantized_float8 in quantization APIs

danielvegamyhre:
**Context**

Currently, AQT has the method [from_hp_to_floatx](https://github.com/pytorch/ao/blob/5d1444bdef6df15eb89c4c5716ede1c5f8677798/torchao/dtypes/affine_quantized_tensor.py#L351) for float8 quantization, and [from_hp_to_fpx](https://github.com/pytorch/ao/blob/5d1444bdef6df15eb89c4c5716ede1c5f8677798/torchao/dtypes/affine_quantized_tensor.py#L399) for low precision floating point data types like fp6 (technically can support fp1-fp7).

`from_hp_to_floatx` re-uses [from_hp_to_intx](https://github.com/pytorch/ao/blob/5d1444bdef6df15eb89c4c5716ede1c5f8677798/torchao/dtypes/affine_quantized_tensor.py#L352), which in turn uses these [generic quantization primitives](https://github.com/pytorch/ao/blob/5d1444bdef6df15eb89c4c5716ede1c5f8677798/torchao/dtypes/affine_quantized_tensor.py#L250-L277). 

Overall, in the current state the float8 path is a bit confusing for developers, due to both the naming ("floatx") and the use of generic functions which include a bunch of params which are unrelated to float8 quantization.

**Summary of changes**

The goal of this PR stack is to refactor this to have a clean separation of concerns, and simpler internal API surfaces for code using in float8 quantization for inference.

Specifically:
- Separate quantization primitives for float8 
- Integrate those new quant primitives into AQT 
- Integrate new AQT methods into float8 quantization APIs <------------------- (this PR)

**Note**: I will add float8 static quantization in a separate set of PRs.


danielvegamyhre:
* (to be filled)

jerryzh168:
thanks, we also want to split out a Float8 (and floatx) specific AQT implementations as well, I talked to @jainapurva before

danielvegamyhre:
> thanks, we also want to split out a Float8 (and floatx) specific AQT implementations as well, I talked to @jainapurva before

Yep that makes sense, when I talked to her earlier she said she is planning to create these AQT subclasses, so I decided to do this part of the refactor.

torchao/dtypes/__init__.py
```diff
@@ -38,6 +34,7 @@
     "to_affine_quantized_fpx",
     "to_affine_quantized_floatx",
```
jainapurva, line 35: Please remove floatx, float8 should replace floatx.

jainapurva:
> > thanks, we also want to split out a Float8 (and floatx) specific AQT implementations as well, I talked to @jainapurva before
> 
> 
> 
> Yep that makes sense, when I talked to her earlier she said she is planning to create these AQT subclasses, so I decided to do this part of the refactor.

Yes, we want all the instances replaced. Autoquant is using it for Float8. Hence would be better to rename it float8

danielvegamyhre:
> > > thanks, we also want to split out a Float8 (and floatx) specific AQT implementations as well, I talked to @jainapurva before
> > 
> > 
> > Yep that makes sense, when I talked to her earlier she said she is planning to create these AQT subclasses, so I decided to do this part of the refactor.
> 
> Yes, we want all the instances replaced. Autoquant is using it for Float8. Hence would be better to rename it float8

Done!

torchao/dtypes/float8/float8_layout.py
```diff
@@ -209,19 +215,64 @@ def __repr__(self):
         )
 
 
+class Float8QuantizedTensor(AffineQuantizedTensor):
```
vkuzo, line 218: I'm not a fan of this, this introduces one more abstraction (`Float8QuantizedTensor`), while keeping the complexity of `AffineQuantizedTensor`.  I think either staying with AQT or just writing a float8 tensor without using AQT would seem more attractive.

danielvegamyhre: Interesting perspective - cc @jainapurva @jerryzh168 thoughts on this?

danielvegamyhre:
Discussed offline, closing until internal discussions are finalized.

PR #1601, opened 2025-01-22
Jcaip/llm bsr

jcaip:
This PR promotes Supermask and block sparsity from prototype -> `torchao.sparsity`

It adds a new public API for `SupermaskLinear`, which users can use to add Supermask to their models with

```
sparsify_(model, lambda x: SupermaskLinear.to_dense(x, sparsity_level=0.9)
```

I have also modified all the existing supermask sam testing code to use this new API. 

It also ports over the triton addmm kernels from core, to let us modify them as needed. I've added padding support into the triton kernel, which was a 4 tok/s improvement (214 -> 218). 

* Adds padding to BSR

On a H100 benchmarking with the following commands
```
export MODEL_REPO=meta-llama/Meta-Llama-3.1-8B

python generate.py --checkpoint_path $CHECKPOINT_PATH/$MODEL_REPO/model.pth --compile --compile_prefill --write_result benchmark_results.txt --prefill_size 8192 --profile baseline_prefill
python generate.py --checkpoint_path $CHECKPOINT_PATH/$MODEL_REPO/model.pth --compile --compile_prefill --write_result benchmark_results.txt --prefill_size 8192 --sparsity bsr --profile bsr_prefill
python generate.py --checkpoint_path $CHECKPOINT_PATH/$MODEL_REPO/model.pth --compile --compile_prefill --write_result benchmark_results.txt --profile baseline
python generate.py --checkpoint_path $CHECKPOINT_PATH/$MODEL_REPO/model.pth --compile --compile_prefill --write_result benchmark_results.txt --sparsity bsr --profile bsr
```

yields a 134 -> 218 tok/s improvemnt on LLM decoding.


torchao/prototype/sparsity/superblock/bsr_triton_ops.py
```diff
@@ -0,0 +1,2541 @@
...
+        for _ in range(row_nnz):
+            values_block = tl.load(values_block_ptrs)
+
+            # find which row of dense needs to get loaded
+            # for multiplication with values_block.
+            dense_row_idx = tl.load(col_index_nnz_ptr)
+            offsets = tl.arange(0, 16)[None, :]
+            dense_block = tl.load(
+                dense_block_ptrs + dense_tiled_row_stride * dense_row_idx,
+                mask=offsets < BLOCKSIZE_COL,
```
jcaip, line 2480: cc @cpuhrsch masking added in here for the padding

torchao/prototype/sparsity/superblock/bsr_triton_ops.py
```diff
@@ -0,0 +1,2541 @@
...
+        nnz_offset = tl.load(crow_indices_offset_ptr)
+        nnz_offset_next = tl.load(crow_indices_offset_ptr + crow_indices_stride)
+
+        # Compute nnz for the row with number row_block_pid.
+        row_nnz = nnz_offset_next - nnz_offset
+
+        row_block_arange = tl.arange(0, BLOCKSIZE_ROW)
+        inner_block_arange = tl.arange(0, BLOCKSIZE_INNER)
+
+        if BLOCKSIZE_COL < 16 or BLOCKSIZE_COL % 16 != 0:
```
jcaip, line 2426: This is the padding logic (need to do this properly instead of hardcoding 16)

torchao/prototype/sparsity/superblock/blocksparse.py
```diff
@@ -206,6 +241,69 @@ def block_sparse_detach(func, types, args, kwargs):
...
+                                 bsr.col_indices(),
+                                 new_values)
+
+    if isinstance(bsr, torch.Tensor) and isinstance(t, BlockSparseTensor):
+        return my_mul(t, bsr)
+    return my_mul(bsr, t)
+
+
+@implements(aten.sum.dim_IntList)
+def block_sparse_sum(func, types, args, kwargs):
```
jcaip, line 287: cc @cpuhrsch This computes the sum properly for the fast path reduction, but doesn't work with compile because of L300 `temp_sum = bsr.values()[start:stop]` which errors out on data dependent flow.

I think we can add a new kernel to the `bsr_dense_addmm` implementation to handle the fast path there instead, and rewrite this using triton

cpuhrsch: Right, so here it's useful to view crow_indices and values as a NestedTensor and then use sum from there :) This is possible because values + crow_indices is like values + offsets.

PR #1607, opened 2025-01-23
[WIP] speed up CodebookQuantizedTensor inference

DerekLiu35:
This PR tries to speedup inference for https://github.com/pytorch/ao/issues/1195

Currently I've ported code1x16 matrix multiplication kernels from AQLM to torchao 

#### **Usage**
[Demo notebook with preliminary tests](https://gist.github.com/DerekLiu35/54965431689b0578c89411b2a4ab1b61)

- Preliminary tests show a ~2x speedup for linear layer compared to the fallback implementation.
- Matches AQLM's performance for `(1, 8)` block sizes 
- quantization error of kernel seems much higher than fallback implementation, need to investigate more

### **ToDo**

- [ ] Improve performance for block sizes of `(1, 1)`.
- [ ] Add triton fall back for other block sizes
- [ ] Add tests.
- [ ] Update benchmarks

Would appreciate feedback on approach and preliminary results
@jerryzh168 @pawarmanasi07


jcaip:
cc @DerekLiu35 are you still planning to work on this?

PR #1615, opened 2025-01-24
Move tensor_flatten/unflatten from AQT -> TensorAOBaseClass

jainapurva:


torchao/utils.py
```diff
@@ -573,6 +573,35 @@ def _get_to_kwargs(self, *args, **kwargs):
         }
         return kwargs
 
+    def __tensor_flatten__(self):
```
jerryzh168, line 576: this is AQT specific, so shouldn't be here I think.

what can be here is a general `__tensor_flatten__` and `__tensor_unflatten__` that can take a list of tensor attr names and other attribute names and fill in the remaining code

PR #1627, opened 2025-01-27
Prepare for -DPy_LIMITED_API flag in pytorch #145764

janeyx99:
PyTorch core is going to start enforcing the Py_LIMITED_API flag definition for best practices + sanity whenever the py_limited_api is passed in. This PR prepares AO to smoothly handle that change when the nightly is updated.

PT PR:  https://github.com/pytorch/pytorch/pull/145764

Test plan:
This should be a noop for both build and runtime.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,6 +14,7 @@ from setuptools import Extension, find_packages, setup
 
 current_date = datetime.now().strftime("%Y%m%d")
 
+PY3_9_HEXCODE = "0x03090000"
 
 def get_git_commit_id():
     try:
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -212,24 +214,26 @@ def get_extensions():
...
-        ]
+        ],
     }
 
     if not IS_WINDOWS:
-        extra_compile_args["cxx"] = [
-            "-O3" if not debug_mode else "-O0",
-            "-fdiagnostics-color=always",
-        ]
+        extra_compile_args["cxx"].extend(
```
drisspg, line 225: good fix

setup.py
```diff
@@ -14,6 +14,8 @@
 
 current_date = datetime.now().strftime("%Y%m%d")
 
+PY3_9_HEXCODE = "0x03090000"
```
drisspg, line 17: this strange AF to me as a flag

drisspg (approved):

setup.py
janeyx99: lol yea welcome to the undocumented world of python

PR #1635, opened 2025-01-29
Add boilerplate code

jainapurva:
* #1636
* __->__ #1635


torchao/utils.py
```diff
@@ -573,6 +585,58 @@ def _get_to_kwargs(self, *args, **kwargs):
...
+        if not hasattr(self, "_layout"):
+            return None
+        return self._layout
+
+
+implements = TorchAOBaseTensor.implements
+aten = torch.ops.aten
+
+
+@implements(aten.detach.default)
```
jerryzh168, line 613: one thing here is what would happen when a tensor subclasses TorchAOBaseTensor and tried to overwrite these functions, to allow child tensor classes to overwrite these functions I think we'd have to copy the table https://github.com/pytorch/ao/blob/7b0d2ce50baaa2a137eb9d438a076544c43096a3/torchao/utils.py#L402 when we detect a new child class is created, e.g.

```
def _implements(cls, aten_ops_or_torch_fns):
    # making sure we are querying the attribute from current class, not parent
    # please check if this works
    if "_ATEN_OP_OR_TORCH_FN_TABLE" not in dir(cls):
         # copy the table from parent

def _dispatch__torch_function__(cls, func, types, args=(), kwargs=None):
    # making sure we are querying the attribute from current class, not parent
    # please check if this works
    if "_ATEN_OP_OR_TORCH_FN_TABLE" not in dir(cls):
         # copy the table from parent if it exists
```

torchao/utils.py
```diff
@@ -512,9 +513,15 @@ def _get_tensor_impl_constructor(
...
 
+"""
+TensorAOBase subclass is a util tensor subclass that provides commonly used functions, and should be inherited to define a new tensor subclass
+"""
+
+
 class TorchAOBaseTensor(torch.Tensor):
     """A util tensor subclass that provides commonly used functions
-       new tensor subclass can inherit it to get all the utility functions
+       new tensor subclass can inherit it to get all the utility functions, and
```
vkuzo, line 523: what are the pros and cons of using inheritance here versus just having utility functions and tensors using what they need from utility functions, without inheritance?  It might be hard to come up with a `TorchAOBaseTensor` which is generic enough to truly handle all the important use cases in torchao.

jerryzh168: makes sense, we could have both I think

jainapurva: As per my understanding:
Option 1: We can move out all the util functions, and make an independent tensor-subclass (inheriting from torch.Tensor), then developer will have to define/inherit util functions.
Option 2: We can add the minimum needed util functions in TorchAOBaseTensor, so that it gives a base start to developer and build on top of it.

jerryzh168: depends on how much can be reused by other tensor subclasses I think, if the TorchAOBaseTensor is very small then it may not make sense to have it any more and it will not be useful for existing inference tensor subclasses. I feel it might be better to have both the current TorchAOBaseTensor and some utils that can be reused by other tensor subclasses as a starting point and then adapt based on use cases

torchao/utils.py
```diff
@@ -539,39 +567,45 @@ def _(func, types, args, kwargs):
...
+    ):
+        raise NotImplementedError("Subclasses must implement __tensor_unflatten__")
+
+    def __repr__(self):
+        raise NotImplementedError("Subclasses must implement __repr__")
+
+    def _apply_fn_to_data(self, fn):
+        raise NotImplementedError("Subclasses must implement _apply_fn_to_data")
 
-    def _get_to_kwargs(self, *args, **kwargs):
```
jerryzh168, line 556: this should be preserved I think, since it's called in child classes

jerryzh168 (approved):

PR #1636, opened 2025-01-29
Create separate float8 tensor subclass

jainapurva:
Float8Tensor, a new subclass for float8, independent of AQT. It's a BC breaking change, hence will need to proceed with it, by deprecating the current API, and issuing a user warning about the deprecation. The new API will be available to users for use. After a substancial amount of time, once we know user has replaced the deprecated API with the updated one, we can delete it's code

* __->__ #1636
* #1635

### DIFF
--- a/torchao/dtypes/affine_quantized_tensor.py
+++ b/torchao/dtypes/affine_quantized_tensor.py
@@ -338,64 +338,6 @@ class AffineQuantizedTensor(TorchAOBaseTensor):
         )
 
     @classmethod
-    def from_hp_to_floatx(
-        cls,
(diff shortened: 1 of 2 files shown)


vkuzo (changes requested):
Since this is BC-breaking, it would be great to see a more thorough writeup in the PR summary of why we want this and how this fits into the overall strategy of torchao.

jainapurva:
> Since this is BC-breaking, it would be great to see a more thorough writeup in the PR summary of why we want this and how this fits into the overall strategy of torchao.

Yes, this PR is still a work in progress. Will add a thorough write-up

PR #1625, opened 2025-01-27
[not for land] hook up MX to CUDA 12.8 cuBLAS MX gemm

vkuzo:
Summary:

Requires https://github.com/pytorch/pytorch/pull/145562/files

None of this is for land - just testing for now as we work on a long
term support plan.

Status:

## performance

* [done] torch._scaled_mm performance as expected: seeing about 2x speedup from bf16 to mxfp8, and 4x from bf16 to nvfp4
* [not in this PR] fast scaling kernels

## numerics - mxfp8
* [done] hello world
* [done] check numerics with A and B set to eye-one and scales set to all-ones
* [done] numerics match fp32 baseline and mx emulation with a single tile (M == K == N == 128), with torch.randn inputs
* [done] numerics match fp32 baseline for multiple tiles, of various shapes
* [done] (not in this PR) train llama with real MX gemm and get decent numerics

## numerics - nvfp4
* [done] hello world
* [not done] repeat all steps of mxfp8 numerics testing

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

PR #1637, opened 2025-01-29
Add mx_fp8_bf16 kernel

drisspg:
Stacked PRs:
 * #1661
 * __->__#1637

--- --- ---

Add mx_fp8_bf16 kernel

Will flesh out more but this moves over the kernel from here: https://github.com/drisspg/driss_torch/blob/2813322f0b0f9a0f0fc8d382090ad0aaecf3468a/src/mx_fp8_bf16.cu#L162

This does fp8xfp8 w/ E8m0 scales and group_size hard coded to 32. The format for the scales is the same as that for cublasLT. I have created a pytorch function that converts the [n_rows, n_cols//32] scales into the expected format:
https://github.com/drisspg/transformer_nuggets/blob/382cb0f19a5f615827174289b8ef552419d51fea/transformer_nuggets/mx/to_blocked.py#L11
This was surprisingly hard fought and would not have been possible w/ @albanD 😊

This allows this PR: https://github.com/pytorch/ao/pull/1625 to not have any dependencies on PT core updates while we add the required dtypes and bindings to cublas: https://github.com/pytorch/pytorch/pull/145562

### Follow up

Config needs more tuning

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -215,10 +215,7 @@ def get_extensions():
     extra_link_args = []
     extra_compile_args = {
         "cxx": [f"-DPy_LIMITED_API={PY3_9_HEXCODE}"],
-        "nvcc": [
-            "-O3" if not debug_mode else "-O0",
-            "-t=0",
-        ],
+        "nvcc": ["-O3" if not debug_mode else "-O0", "-t=0", "-std=c++17"],
     }
 
     if not IS_WINDOWS:
@@ -257,12 +254,16 @@ def get_extensions():
         use_cutlass = True
         cutlass_dir = os.path.join(third_party_path, "cutlass")
         cutlass_include_dir = os.path.join(cutlass_dir, "include")
+        cutlass_tools_include_dir = os.path.join(
+            cutlass_dir, "tools", "util", "include"
+        )
         cutlass_extensions_include_dir = os.path.join(cwd, extensions_cuda_dir)
     if use_cutlass:
         extra_compile_args["nvcc"].extend(
             [
                 "-DTORCHAO_USE_CUTLASS",
                 "-I" + cutlass_include_dir,
+                "-I" + cutlass_tools_include_dir,
                 "-I" + cutlass_extensions_include_dir,
             ]
         )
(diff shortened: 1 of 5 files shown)


vkuzo (approved):
nice! if CI is green - looks good! I think this should have at least one numerical test though.  Can be a follow-up PR if needed.

setup.py
```diff
@@ -218,6 +218,7 @@ def get_extensions():
         "nvcc": [
             "-O3" if not debug_mode else "-O0",
             "-t=0",
+             "-std=c++20"
```
albanD, line 221: :fire:

albanD (approved):
Very cool!

setup.py
drisspg: I don't think we can land this ahahh I like it but I'm sure somewhere there we don't support a new enough compiler

albanD: In ao, you most likely can if you don't care about windows.
If you do care about windows, "most likely" -> "most likely can't" 
:D

torchao/ops.py
```diff
@@ -615,3 +618,29 @@ def _(
         dtype=input_scale.dtype,
         device=input.device,
     )
+
+
+def mx_fp8_bf16(A: Tensor, B: Tensor, A_scale: Tensor, B_scale: Tensor):
```
vkuzo, line 623: should we make it clear this is prototype and subject to change at any time?

torchao/ops.py
```diff
@@ -592,3 +593,31 @@ def _(
     bias: Tensor,
 ) -> Tensor:
     return input_scale.new_empty(*input.shape[:-1], weight.shape[0])
+
+
+def mx_fp8_bf16(A: Tensor, B: Tensor, A_scale: Tensor, B_scale: Tensor):
+    """Defines a matmul between two fp8 tensors w/ MX scales in E8MO and returns a bf16 tensor.
```
vkuzo, line 599: nit: should we clarify this only supports `float8_e4m3fn` for now?  

If we wanted to add `float8_e5m2` or the `nuz` variants, would it also go in this function?

PR #1642, opened 2025-01-29
Q dq layout

metascroy:


torchao/experimental/quant_api.py
```diff
@@ -598,15 +596,21 @@ def apply(weight):
...
             _layout=layout,
         )
 
         # Note that PackedLinearInt8DynamicActivationIntxWeightLayout has dynamic activation quantization fused
         # with the kernel and it should not be applied separately
-        if not isinstance(layout, PackedLinearInt8DynamicActivationIntxWeightLayout):
+        if not any(
+            isinstance(layout, layout_class)
+            for layout_class in [
+                QDQLayout,
```
metascroy, line 610: @jerryzh168 the QDQLayout here fuses the activation quantization ops with the weight ops rather than use to_affine_quantized_intx.

The reason is because using to_affine_quantized_intx means "torch.ops.quant.choose_qparams_affine.default" op will show up in export instead of "torch.ops.quantized_decomposed.choose_qparams_per_token_asymmetric.default" (wanted by ExecuTorch).  By fusing, I can control the ops used in register_aqt_quantized_linear_dispatch.

Is there a better way of doing this dispatch?  Something like register_aqt_quantized_linear_dispatch, but for the activations.

jerryzh168: I think `torch.ops.quant.choose_qparams_affine.default` covers `torch.ops.quantized_decomposed.choose_qparams_per_token_asymmetric.default` already? I think we should move executorch to use `torch.ops.quant.choose_qparams_affine.default`, @mcr229 and @digantdesai has worked on this before

metascroy: I was trying to make the exported graph equivalent to what is Int8DynActInt4WeightQuantizer produces.  If XNNPACK partitioner can handle torch.ops.quant.choose_qparams_affine.default, then that also works.  cc @digantdesai @mcr229 to confirm.

torchao/experimental/q_dq_layout.py
```diff
@@ -0,0 +1,78 @@
...
+    return isinstance(layout, QDQLayout)
+
+
+def _linear_impl(input_tensor, weight_tensor, bias):
+    x = per_token_dynamic_quant(input_tensor)
+
+    w_int_data = weight_tensor.tensor_impl.int_data
+    w_scale = weight_tensor.tensor_impl.scale
+    w_zero_point = weight_tensor.tensor_impl.zero_point
+    assert len(weight_tensor.block_size) == 2
```
kimishpatel, line 57: Can you add some meaningful message?

torchao/experimental/quant_api.py
kimishpatel: > I think `torch.ops.quant.choose_qparams_affine.default` covers `torch.ops.quantized_decomposed.choose_qparams_per_token_asymmetric.default` already? I think we should move executorch delegation to use `torch.ops.quant.choose_qparams_affine.default`, @mcr229 and @digantdesai has worked on this before

@jerryzh168 moving executorch delegates and flow to use ops from torch.ops.quant we need to make PT2E flow do the same thing. I think we have circled on this in the past but not clear if we converged whether quantized_decomposed ops and PT2E workflow should move to AO or torch.ops.quant should mvoe to pytorch core. Can you say what the plan is?

torchao/experimental/tests/test_q_dq_layout.py
```diff
@@ -0,0 +1,173 @@
...
+                )
+
+                quantized_model_reference = copy.deepcopy(model)
+                quantize_(
+                    quantized_model_reference,
+                    int8_dynamic_activation_intx_weight(
+                        weight_dtype=weight_dtype,
+                        granularity=granularity,
+                        has_weight_zeros=has_weight_zeros,
+                        layout=PlainLayout(),
```
kimishpatel, line 66: I cant figure out what this one dispatches to

kimishpatel: code pointer?

torchao/experimental/tests/test_q_dq_layout.py
```diff
@@ -0,0 +1,173 @@
...
+        has_weight_zeros = False
+        layers = [
+            torch.nn.Linear(512, 256, bias=False),
+        ]
+        model = torch.nn.Sequential(*layers)
+        activations = torch.randn(1, 512, dtype=torch.float32)
+
+        to_export_with_old_api = copy.deepcopy(model)
+
+        print("Quantizing model")
```
kimishpatel, line 101: remove

torchao/experimental/tests/test_q_dq_layout.py
```diff
@@ -0,0 +1,173 @@
...
+                has_weight_zeros=has_weight_zeros,
+                layout=QDQLayout(),
+            ),
+        )
+        eager_results = model(activations)
+
+        unwrap_tensor_subclass(model)
+
+        print("Exporting quantized model")
+        exported = torch.export.export(model, (activations,), strict=True)
```
kimishpatel, line 116: export_for_training?

torchao/experimental/tests/test_q_dq_layout.py
```diff
@@ -0,0 +1,173 @@
...
+        # TODO: there are slight differences in the results because exported_results uses
+        # asymmetric with zero_point_domain NONE (has_weight_zeros=False)
+        # and results_from_old_api uses symmetric (but with an asymmetric range)
+        # I think the new API might make more sense, but need more thought
+        # results_from_old_api = exported_from_old_api.module()(activations)
+        # self.assertTrue(torch.allclose(exported_results, results_from_old_api))
+
+
+if __name__ == "__main__":
+    unittest.main()
```
kimishpatel, line 173: Why this? You can just run this with pytest, right?

torchao/experimental/tests/test_q_dq_layout.py
metascroy: This took me a while to figure out, but I think it dispatches to the fallback path here: https://github.com/pytorch/ao/blob/main/torchao/dtypes/affine_quantized_tensor_ops.py#L171-L201

torchao/experimental/tests/test_q_dq_layout.py
metascroy: I could, but pytest doesn't show me messages like "Testing weight_dtype={weight_dtype}, has_weight_zeros={has_weight_zeros}", which was helpful for debugging.

I can remove the print statements and use pytest, though.

torchao/experimental/quant_api.py
mcr229: yea choose_qparams_affine is a supported node for the xnnpack partitioner:
https://github.com/pytorch/executorch/blob/073397357118feef0fca91326ed612ce5c60d53b/backends/xnnpack/partition/config/quant_affine_configs.py#L57

jerryzh168: @kimishpatel we want to move pt2e workflow to ao, just sent you the plan in internal chat

torchao/experimental/tests/test_q_dq_layout.py
jerryzh168: we also have some feedback that we should not have a fallback path like this, maybe we can create a dispatch for it as well to make it clearer and remove the fallback path

torchao/experimental/quant_api.py
kimishpatel: Sounds good.

torchao/experimental/tests/test_q_dq_layout.py
kimishpatel: > we also have some feedback that we should not have a fallback path like this, maybe we can create a dispatch for it as well to make it clearer and remove the fallback path

Yeah that would be cleaner.

kimishpatel (approved):
Looks good to me. I presume you will address the pending comments and nothing critical is block, approving

torchao/experimental/quant_api.py
metascroy: Sounds good!  I'll change QDQ to use ops in torch.ops.quant like choose_qparams_affine and dequantize_affine.  This will make it equivalent to PlainLayout today.

PR #516, opened 2024-07-17
[Experimental] Support rounding before calculating the largest power of 2

NicoleMayer:
There are 3 methods for calculating MX block scaling:
1. FLOOR: this method is recommended by the OCP MX Spec 1.0 and uses `X = 2^floor(log2(max_abs(v))-max_exp)` 
2. CEIL: this method avoids overflow issues, but small values may shift to 0 due to a large scaling factor. It uses `X = 2^ceil(log2(max_abs(v))-max_exp)`
3. EVEN: this method is a trade-off between Option 1 and Option 2. It uses `X = 2^(floor(log2(rounding(max_abs(v)))-max_exp))`

We support all three methods for calculating MX block scaling using the `ScaleCalculationMode` config. By default, we use the `EVEN` method for better accuracy.

PR #1643, opened 2025-01-30
mx: add ceil and RNE rounding modes to the cast from fp32 to e8m0

vkuzo:
Summary:

Why we want this: the MX spec describes floor rounding, which isn't always the most accurate.  It is good to have more options to experiment with.

This is a copy-pasta of https://github.com/pytorch/ao/pull/516 with the
modifications being keeping FLOOR the default more, and removing e3m0, credit
to NicoleMayer for the original code. I don't have a way to check bitwise
equivalency with the most recent cuBLAS version yet, but will come back
and add tests when I do.

Test Plan:

```
```

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo:
* (to be filled)

drisspg (approved):

PR #1645, opened 2025-01-30
Change TORCH_LIBRARY to TORCH_LIBRARY_FRAGMENT

metascroy:
This is to prevent conflicts between CPU/MPS because torch requires all ops be defined in same block.

Test plan: CI + 

```
cd ao/torchao/experimental/ops/mps
bash build.sh
python test/test_lowbit.py
```

### DIFF
--- a/torchao/experimental/ops/mps/linear_fp_act_xbit_weight_aten.mm
+++ b/torchao/experimental/ops/mps/linear_fp_act_xbit_weight_aten.mm
@@ -163,7 +163,7 @@ Tensor pack_weights_cpu_kernel(const Tensor& W) {
   return B;
 }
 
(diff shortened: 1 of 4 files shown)


torchao/experimental/ops/mps/linear_fp_act_xbit_weight_aten.mm
metascroy, line 1: @manuelcandales how do I test the change?  Is there a script I should run?

manuelcandales:
```
cd ao/torchao/experimental/ops/mps
bash build.sh
python test/test_lowbit.py
```

manuelcandales (approved):

PR #1647, opened 2025-01-31
[Fix]: Fallback to KleidiAI channelwise kernel groupsize isnt suitable

nikhil-arm:
Description:
1. Some models can have certain odd shapes which can not be used with blocked quantization. Fallback to channelwise quantization for those shapes.
2. Fix Formatting issue with experimental tests

### DIFF
--- a/torchao/experimental/quant_api.py
+++ b/torchao/experimental/quant_api.py
@@ -587,8 +587,6 @@ def int8_dynamic_activation_intx_weight(
                 f"granularity must be PerGroup or PerRow, got {granularity}"
             )
 
-        assert weight.shape[-1] % group_size == 0
-
         layout = layout_arg
         scale_dtype = None
         tensor_quantizer = to_affine_quantized_intx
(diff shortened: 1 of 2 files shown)


ng-05:
@digantdesai @metascroy can you please help with reviewing this

torchao/experimental/quant_api.py
```diff
@@ -628,15 +621,28 @@ def apply(weight, bias: Optional[torch.Tensor] = None):
...
                     TORCH_VERSION_AT_LEAST_2_6
                 ), "aten target is requires torch version > 2.6.0"
+                # Fallback to Channelwise scheme if group_size is too big
+                if weight.shape[-1] < group_size:
+                    logger.warning(f"Changing group_size to {
+                                   weight.shape[-1]}. Weight shape {weight.shape} can not support group_size {group_size}.")
+                    group_size = weight.shape[-1]
                 if torch.backends.kleidiai.is_available():
-                    if isinstance(granularity, PerGroup):
+                    if weight.shape[-1] != group_size and group_size % 32 == 0:
```
digantdesai, line 630: `weight.shape[-1] != group_size` I don't understand this constraint.

IIRC the constraint is `k % group_size == 0 and group_size % 32 == 0`

torchao/experimental/quant_api.py
```diff
@@ -628,15 +621,28 @@ def apply(weight, bias: Optional[torch.Tensor] = None):
                 assert (
                     TORCH_VERSION_AT_LEAST_2_6
                 ), "aten target is requires torch version > 2.6.0"
+                # Fallback to Channelwise scheme if group_size is too big
+                if weight.shape[-1] < group_size:
```
vkuzo, line 625: I'd recommend adding a top-level configuration option which clearly tells the user "group size can be changed for certain weight shapes" to handle this case, and throwing an exception if that config setting isn't on.

nikhil-arm: Because groupwise is more accurate and we might want to selectively fallback for certain unsupported layers. But I am closing this, user can always filter nn module and use multiple calls to quantizer with correct configuration.

PR #1650, opened 2025-02-01
more stringent test for CPUOffloadOptimizer

ngc92:
Problems with CPUOffloadOptimizer's  synchronization (cf. #1649)  are not detected by the current test case, as it is very benign, giving ample opportunity for transfers to complete even without explicit sync.

These changes try to make the situation a bit more challenging, decreasing the arithmetic density of the model and ensuring that the critical path is as short as possible.

This PR is deliberately not based on top of #1649, to show that this new test actually catches the synchronization problem. After that PR is merged, tests should pass again.


gau-nernst:
@ngc92 You can add `self.stream.synchronize()` at the end of optim step in this PR, then I will convert #1649 to only address the LR scheduler issue.

Thank you again for identifying the issue and coming up with an appropriate test!

gau-nernst (approved):
Failing ruff check is not caused by this change.

PR #1660, opened 2025-02-04
Tensor parallel support for fpx dtype

jainapurva:


torchao/dtypes/floatx/floatx_tensor_core_layout.py
```diff
@@ -598,6 +598,14 @@ def __torch_dispatch__(cls, func, types, args, kwargs):
                     lambda x: x.to(device=kwargs.pop("device", None))
                 ),
             )
+        elif func is aten.slice.Tensor:
+            return return_and_correct_aliasing(
+                func, args, kwargs, args[0]._apply_fn_to_data(aten.slice.Tensor)
```
jerryzh168, line 603: really? is this correct?

Issue #1666, opened 2025-02-05
[MX] Support mixed MXFP4/FP6/FP8 linear layer


balancap:
Blackwell hardware natively supports any combination of MXFP4/FP6/FP8 in matmuls. See PTX and Cutlass documentation:
* https://docs.nvidia.com/cuda/parallel-thread-execution/#tcgen05-kind-shapes 
* https://github.com/NVIDIA/cutlass/blob/main/tools/library/src/reference/block_scaled_gemm_mixed8bitsa.cu

According the [MX paper](https://arxiv.org/abs/2310.10537), and macore generally the large quantization literature, there is advantages to use different bitwidth for weights, activations and gradients. It would be very useful in `mx_mm` and `MXLinear` to support this more general setting.

balancap:
#1667 is adding mixed element dtypes support to `mx_mm`. A similar general interface could be added to `MXLinear`
```python
class MXLinear(torch.nn.Linear):
    @classmethod
    @torch.no_grad()
    def from_float(cls, mod, in_elem_dtype, w_elem_dtype, grad_elem_dtype, block_size):
        ...
```

jainapurva:
@vkuzo @drisspg

balancap:
Closing the issue as #1667 has been merged

PR #1667, opened 2025-02-05
Support mixed MX element dtype in `mx_mm` function and `MXLinear`.

balancap:
Following the MXFP and quantization literature, it is useful to support different element dtypes for activations, weights and gradients. This PR is simply adding a more general interface to `mx_mm`. A similar choice could be done with `MXLinear`

General issue: #1666

### DIFF
--- a/torchao/prototype/mx_formats/mx_linear.py
+++ b/torchao/prototype/mx_formats/mx_linear.py
@@ -23,25 +23,31 @@ class mx_mm(torch.autograd.Function):
     # 1.       input @ weight_t    = output     (forward pass)
     # 2. grad_output @ weight      = grad_input (backward pass)
     # 3.     input_t @ grad_output = grad_weight (backward pass)
+    # 
+    # input, weight and grad_output have each their own MX element dtype.
 
     @staticmethod
     def forward(
         ctx,
         input_hp: torch.Tensor,
         weight_hp: torch.Tensor,
-        elem_dtype: Any,
+        in_elem_dtype: Any,
+        w_elem_dtype: Any,
+        grad_elem_dtype: Any,
         block_size: int,
     ):
         ctx.save_for_backward(input_hp, weight_hp)
-        ctx.elem_dtype = elem_dtype
+        ctx.in_elem_dtype = in_elem_dtype
+        ctx.w_elem_dtype = w_elem_dtype
+        ctx.grad_elem_dtype = grad_elem_dtype
         ctx.block_size = block_size
 
         # input @ weight_t = output
         input_orig_shape = input_hp.shape
         input_hp_r = input_hp.reshape(-1, input_orig_shape[-1])
 
-        input_mx_r_dim0 = MXTensor.to_mx(input_hp_r, elem_dtype, block_size)
-        weight_mx_dim0 = MXTensor.to_mx(weight_hp, elem_dtype, block_size)
+        input_mx_r_dim0 = MXTensor.to_mx(input_hp_r, in_elem_dtype, block_size)
+        weight_mx_dim0 = MXTensor.to_mx(weight_hp, w_elem_dtype, block_size)
         output = torch.mm(input_mx_r_dim0, weight_mx_dim0.t())
         output = output.reshape(*input_orig_shape[:-1], output.shape[-1])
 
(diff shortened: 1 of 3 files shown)


torchao/prototype/mx_formats/mx_linear.py
```diff
@@ -23,25 +23,31 @@ class mx_mm(torch.autograd.Function):
     # 1.       input @ weight_t    = output     (forward pass)
     # 2. grad_output @ weight      = grad_input (backward pass)
     # 3.     input_t @ grad_output = grad_weight (backward pass)
+    # 
+    # input, weight and grad_output have each their own MX element dtype.
```
vkuzo, line 27: nit: "can have"?

vkuzo:
this makes sense, it would be great to cover with a test

the easiest place to test it would be here (https://github.com/pytorch/ao/blob/8afd10ed4b22b3cabd80184062c4ad58001bc68a/test/prototype/mx_formats/test_mx_linear.py#L47), and that requires adding this to `MXLinear`.  Would you be interested in doing that in this PR?

by the way, https://github.com/pytorch/pytorch/issues/146414 outlines bringing MX dtypes to PyTorch core, and we plan to evolve `torchao/prototype/mx_formats/` accordingly

torchao/prototype/mx_formats/mx_linear.py
balancap: Done

balancap:
I added the support of this feature in `MXLinear` too. In order to avoid breaking the interface (and keeping things simple in the single dtype case), you can now pass either a single element dtype or a tuple of 3.

I expanded the coverage in the test you mentioned (plus a small test on the factory side to check the 2 cases above are working properly).

Thanks for the link on PyTorch MX plan :+1: I would assume that the MX "simulated" mode is going to stay in TorchAO for some time, as it is very useful for testing + getting ready for MX hardware until it is widely available.

torchao/prototype/mx_formats/mx_linear.py
```diff
@@ -87,13 +97,20 @@ class MXLinear(torch.nn.Linear):
...
+    Input, weight and grad_output can have each their own MX element dtype
+    by passing a tuple of `elem_dtype` to the factory method `from_float`.
     """
 
     @classmethod
     @torch.no_grad()
     def from_float(cls, mod, elem_dtype, block_size):
         mod.__class__ = MXLinear
-        mod.elem_dtype = elem_dtype
+        # Single element dtype passed for input, weight and gradient.
```
vkuzo, line 109: nit: can we do

```
def from_float(
    ...,
    elem_dtype=torch.float8_e4m3fn,
    elem_dtype_weight_override=None,
    elem_dtype_grad_output_override=None,
    ...
): ...
```

we plan to create a proper config object for this in the future, but for now would be good to keep things simple and avoid mixing types in the API (such as dtype vs tuple)

vkuzo:
> I would assume that the MX "simulated" mode is going to stay in TorchAO for some time, as it is very useful for testing + getting ready for MX hardware until it is widely available.

yep! great to hear this is useful.

torchao/prototype/mx_formats/mx_linear.py
balancap: Should I then enforce named argument in `MXLinear.from_float` and `swap_linear_with_mx_linear` for `block_size` and `filter_fn`? And have a default `block_size=32`?

vkuzo: sounds reasonable!

balancap: Just pushed a fix commit with:
```python
def from_float(
        cls,
        mod,
        elem_dtype,
        elem_dtype_weight_override=None,
        elem_dtype_grad_output_override=None,
        *,
        block_size=32,
    ):
```
and similarly for `swap_linear_with_mx_linear`

vkuzo (approved):
looks good, thank you!  Please feel free to merge if CI is green.

note: we will likely change the UX of this workflow in the near future (add a top-level config, etc) as we add Blackwell support, we'll make sure to keep these options in the new UX!

balancap:
@vkuzo Unfortunately, the H100 Float8 test runner seems to have had an issue starting.

vkuzo:
failure is transient, I merged it.  Thank you!

PR #1671, opened 2025-02-05
Add CUTLASS-based row-wise scaled sparse FP8 kernel

alexsamardzic:


alexsamardzic:
The kernel is ready and passes smoke test.

Remaining tasks:
- [ ] Write a converter to SM90 sparse semi-structured format
- [ ] Validate the kernel on proper test inputs
- [ ] Check compile warnings about the performance loss, write the benchmark
- [ ] Write Python-side code: sparsify/quantize method, Llama generator extension, etc.
- [ ] Provide that kernel is built with SM90a flags when torchao detects H100 card as SM90
- [ ] Further unify CUDA code with `rowwise_scaled_linear_cutlass` code
- [ ] Implement a meaningful config selection

@cpuhrsch @drisspg

test/test_rowwise_scaled_linear_sparse_cutlass.py
```diff
@@ -0,0 +1,178 @@
...
+        XQ_WQ_DTYPES,
+        BATCH_SIZE,
+        SIZE_MNK,
+        USE_BIAS,
+        BIAS_DTYPE,
+    )
+)
+
+
+def rand_sparse_semi_structured(r, c, dtype, device, choice=None):
```
jcaip, line 65: FYI: I think we have a variant of this in here: https://github.com/pytorch/ao/blob/cc6244c864416926877fc469f6d46db900a90f61/torchao/sparsity/utils.py#L26

alexsamardzic: Fixed.

alexsamardzic:
Testing this PR revealed that the sparse compressor in CUTLASS is not treating -0.0 values as zeros.  The upstream fix is proposed [here](https://github.com/NVIDIA/cutlass/pull/2122).

alexsamardzic:
This PR is ready for review.  It contains:

1. An implementation of two new CUTLASS-based operators:
   - Converter to sparse format for FP8 data and SM9x arch, in `torchao/csrc/cuda/to_sparse_semi_structured_cutlass_sm9x`.
   - Row-wise scaled linear operator implementation for sparse FP8 weight and FP8 activation in `torchao/csrc/cuda/rowwise_scaled_linear_sparse_cutlass`.  For parallel compilation, each operator template instantiation is in a separate `.cu` file.
2. The test for later operator in `test/test_rowwise_scaled_linear_sparse_cutlass.py` (not all tests will pass at the moment because of #1766), and the micro-benchmark in `benchmarks/benchmark_rowwise_scaled_linear_sparse_cutlass.py`.
3. The corresponding layout and `TensorImpl` class implementations in `torchao/dtypes/floatx/cutlass_semi_sparse_layout.py`.  Because of a CUTLASS issue with handling minus zero values when compressing dense to sparse tensor, `from_plain()` method here contains a temporary workaround (the fix for this CUTLASS issue is in the works: NVIDIA/cutlass#2110).
4. The remaining glue code on the Python side in `torchao/ops.py`, `torchao/dtypes/affine_quantized_tensor.py` and `torchao/quantization/quant_api.py`, including definition of new config `Float8DynamicActivationFloat8SemiSparseWeightConfig` for the `quantize_()` method.
5. An update to `torchao/_models/llama/generate.py` script, to make it possible to test the new quantization and linear operator within the context of Llama - run with `python generate.py --compile --sparsity semi -q float8dq`.
6. Some minor updates for CUTLASS-based integer W4A4/W4A8 stuff.

I'll address the performance tuning (through CUTLASS run-time config selection), as well as further unification with row-wise scaled linear non-sparse kernels, that are both mentioned as remaining tasks above, in a separate PR.

@drisspg The `setup.py` changes are about activating `gencode` flags for SM90a when the build is for SM90 - it's clumsy, but it works, so hopefully we could use this approach until eventually switching to CMake builds for the extensions.  I'm adding you as a reviewer because of this; also, please add reviewer(s), whoever may be the most appropriate, for the Python side of the code.

@jcaip If you think there is a need, we may discuss eventually exposing mentioned new operators through `SparseSemiStructuredTensor`.

@gau-nernst With this PR, it's possible to try CUTLASS-based W4A4 operator from the Llama generator - run with `python generate.py --compile --sparsity semi -q int4dq-4` (be sure to fetch the model beforehand - instructions are [here](https://github.com/pytorch/ao/tree/main/torchao/_models/llama)).  The output is not meaningful, maybe it's because the quantization is too tight, but we may want to investigate it further.

torchao/csrc/cuda/rowwise_scaled_linear_sparse_cutlass/rowwise_scaled_linear_sparse_cutlass.cuh
```diff
@@ -0,0 +1,523 @@
...
+  using ElementAccumulator = float;
+  using ElementCompute = float;
+
+  using ProblemShape = cute::Shape<int, int, int, int>;
+
+  // If KernelTmaWarpSpecializedPingpong used for kernel schedule, the
+  // performance is really bad; on the other side, using
+  // KernelTmaWarpSpecializedPingpongFP8FastAccum doesn't seem to
+  // affect the precision much - thus, sticking with it.
+  using KernelSchedule =
```
jcaip, line 81: How much of a performance degredation is slow accumulation? Is it about the same 2x we see for the non-[sparse](https://github.com/pytorch/ao/pull/1325/files#r1868193786) version? 

We have some recipes where this makes a difference for the final accuracy.

torchao/dtypes/uintx/cutlass_int4_packed_layout.py
```diff
@@ -110,8 +108,7 @@ def from_plain(
         _layout: Layout,
     ):
         assert zero_point is None or torch.all(zero_point == 0)
-
-        int_data_s4 = ((int_data[:, 1::2] & 0xF) << 4) | (int_data[:, 0::2] & 0xF)
+        int_data_s4 = ((int_data[..., 1::2] & 0xF) << 4) | (int_data[..., 0::2] & 0xF)
```
jcaip, line 111: just curious here, what's the difference between `...` and `:`?

jcaip (commented):
cc @alexsamardzic Took a first pass, but looks good so far

Did you have any numbers for generate.py? I can try grabbing some if not.

torchao/_models/llama/generate.py
```diff
@@ -564,16 +578,24 @@ def ffn_or_attn_only(mod, fqn):
...
         elif "float8dq" in quantization:
-            granularity = str(quantization.split("-")[-1])
-            if granularity == "tensor":
-                granularity = PerTensor()
-            elif granularity == "row":
-                granularity = PerRow()
+            if sparsity and "semi" in sparsity:
+                quantize_(
+                    model,
+                    Float8DynamicActivationFloat8SemiSparseWeightConfig(),
```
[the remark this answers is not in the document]
jcaip: Actually, this might not be possible with the CUTLASS implementation since we need a separate `AQTTensorImpl` since the values and metadata are separated.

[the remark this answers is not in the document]
alexsamardzic: Yeah, I was confused what to do here too...  I followed what I think sparse Marlin is doing: basically, quantization followed by sparsification is still treated as quantization, with some rearrangement of quantized values.  This is also alike to how, for example, quantization to int4 is treated, first like quantization to int8 with -8/7 as min/max quantized values, and then packing of two quantized values into a single byte.

Let me know if you think there is a better way to handle this, to have `sparsify_()` somehow covered.

torchao/csrc/cuda/rowwise_scaled_linear_sparse_cutlass/rowwise_scaled_linear_sparse_cutlass.cuh
alexsamardzic: For Llama generate.py, it's indeed close to 2x; for the benchmark (`benchmarks/benchmark_rowwise_scaled_linear_sparse_cutlass.py`) of the operator only, it's even worse - up to 3x for some shapes.

I was thinking about making `use_fast_accum` an optional argument for the operator.  However, that would mean eventually having this as a template argument, with two options, that would double number of templates instantiated during the compilation, that would slow down the compilation twice.  So I think it's better to keep it as is for now, it could be added if a need arise.  (For the same reason, the number of options for scales/bias/output data types in `.cu` file is kept to minimum - this is stated explicitly in a comment in the `.cuh` file.).

torchao/dtypes/uintx/cutlass_int4_packed_layout.py
alexsamardzic: Need to look into 3D and higher-dimensional tensors to spot the difference: for example, for 3D tensor `x`, it holds that `x[..., -1]` is `x[:, :, -1]`, while `x[:, -1]` is `x[:, -1, :]`.

torchao/_models/llama/generate.py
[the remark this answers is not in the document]
jcaip: This should be fine for now - sparsify_ calls quantize under the hood, should it should also be able to use `Float8DynamicActivationFloat8SemiSparseWeightConfig`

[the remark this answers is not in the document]
alexsamardzic: Yep - for the generate script, it works this way.  Furthermore, I believe the safer approach is to first quantize and then sparsify, as the other way around for CUTLASS it's not guaranteed that meta-data tensor (holding the sparsity structure info) will stay valid between original and quantized tensors.

jcaip (approved):
lgtm, feel free to merge

PR #1681, opened 2025-02-06
add a deprecation warning for float8 delayed and static scaling

vkuzo:
Summary:

As titled, the complexity tax for these features is high and there no
known real use cases, as the community is overwhelmingly using dynamic
scaling. So, IMO we should deprecate this.

Test Plan: CI

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo:
* __->__ #1681

drisspg (approved):

PR #1687, opened 2025-02-10
Support MXFP6 packing and fused unpack-dequantise kernel

alex-titterton:
Good afternoon! Following recent developments and increased support for MXFP formats, it would be useful to support efficient packing for `MXFP6` to benefit from the decrease in memory consumption and bandwidth requirements vs `(MX)FP8`.

`MXFP6` has shown to perform similarly well compared with `MXFP8` in LLM inference tasks, and with sufficient QAT even as well as `float32`, e.g. in the [MXFP paper](https://arxiv.org/abs/2310.10537).

This PR packs the bits representing the `FP6` values in a `4+2` fashion as is done in the [FP6 LLM paper](https://arxiv.org/abs/2401.14112), and supports both `E2M3` and `E3M2` variants. Packing is done via standalone Triton kernel, with unpacking and dequantisation performed via a fused kernel for better performance.

Tests have been added in `test_custom_cast.py` and `test_mx_tensor.py` to cover accuracy in quantise-pack-unpack-dequantise with various FP6 values (min/max norm, min/max subnorm, `-0.0` etc for both `E2M3` and `E3M2` variants) as well as checking packed tensor dimensions.

Note: due to the `4+2` packing scheme this requires the packing dimension to be a multiple of `4` since the packed dimension will be 3/4 of this. However the typical MX block size is `32` (—> `24` when packed), and HW implementations tend to require dims to be multiples of `16` or `32`, so this should not be a problem. The relevant test case dimensions have been changed from `6` to `8` and the MX block sizes from `2` to `4` where applicable in order to accommodate this requirement.

### DIFF
--- a/torchao/prototype/mx_formats/custom_cast.py
+++ b/torchao/prototype/mx_formats/custom_cast.py
@@ -24,6 +24,8 @@ from torchao.prototype.mx_formats.constants import (
     E8M0_EXPONENT_BIAS,
     E8M0_EXPONENT_NAN_VAL,
     F4_E2M1_EXP_BIAS,
+    F6_E2M3_EXP_BIAS,
+    F6_E3M2_EXP_BIAS,
     F32_EXP_BIAS,
 )
 
@@ -43,6 +45,12 @@ EBITS_F6_E3M2, MBITS_F6_E3M2 = 3, 2
 
 SIGN_MASK_F4 = 0x8  # 1000
 MANTISSA_MASK_F4 = 0x1  # 0001
+
+SIGN_MASK_F6_E2M3 = 0x20  # 100000
+MANTISSA_MASK_F6_E2M3 = 0x7  # 000111
+
+SIGN_MASK_F6_E3M2 = 0x20  # 100000
+MANTISSA_MASK_F6_E3M2 = 0x3  # 000011
 
 ZERO_BITS_F32 = 0x0
 ZERO_POINT_FIVE_BITS_F32 = 0x3F000000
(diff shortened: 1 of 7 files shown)


vkuzo:
@alex-titterton , so sorry I missed this in my review stack.  Overall this looks great, thank you!  Would you mind rebasing (as we've been iterating on MX rapidly recently), and then we can review / get this in?

torchao/prototype/mx_formats/custom_cast.py
```diff
@@ -313,6 +321,300 @@ def triton_f4_to_scaled_bf16_kernel(
...
+    def _fp6_packed_to_bf16(
+        packed_4bits_a,
+        packed_4bits_b,
+        packed_2bits,
+        sign_mask_f6,
+        mbits_f6,
+        f6_exp_bias,
+        mbits_f32,
+        f32_exp_bias,
+    ):
```
vkuzo, line 334: didn't read this line by line but I trust the test cases :)

vkuzo:
> unpacking and dequantisation performed via a fused kernel for better performance.

by the way, wondering if there are any performance numbers you can share on this kernel?

alex-titterton:
> > unpacking and dequantisation performed via a fused kernel for better performance.
> 
> by the way, wondering if there are any performance numbers you can share on this kernel?

Thanks @vkuzo -- basic timings for unpacking with/without the fused kernel for tensor shapes ranging from `[1024, 32]` to `[1024, 8192]` suggest around a 2x improvement with the fused kernel.

I'll get this refactored 👍

vkuzo:
> a 2x improvement with the fused kernel.

nice!

alex-titterton:
Hi @vkuzo -- I see merging is blocked due to a merge commit which I accidentally created when trying to resolve the conflicts from the recent updates to `main`. I've tried to get out of this by rebasing, but the block still appears to be there. 

Would it be best for me to open a fresh PR with a clean commit?

vkuzo:
> Would it be best for me to open a fresh PR with a clean commit?

sure, that works!  There have been a lot of changes lately so it may be easier to just reapply instead of trying to rebase.

alex-titterton:
> > Would it be best for me to open a fresh PR with a clean commit?
> 
> sure, that works! There have been a lot of changes lately so it may be easier to just reapply instead of trying to rebase.

Thanks @vkuzo -- I've opened a new PR #1810 which should make things a bit cleaner.

Cheers,
Alex

PR #1706, opened 2025-02-12
Outline Benchmark Quant APIs

jainapurva:
This is the first PR, in the benchmarking effort. It provides the outline to building microbenchmarking framework for quant api's in torchao.

The different inputs like quantization techniques, matrix sizes, compile, sparsity etc, will be given as input to the python script. The options are re-defined in the scipt for quantization techniques, and a developer can add new quant technique. The script will generate a csv with performance numbers, and that'll be used to plot charts and as an input to dashboard 

Included in current PR:
   - Existing quant techniques
   - Inference on Toy linear model (single linear layer)
   - Compile option

### Run benchmarking:
``` 
python benchmarks/microbenchmarks/config_parser.py --config benchmarks/microbenchmarks/configs/benchmark_config.yml 
```

## Sample configs/benchmark_config.yml:

```YAML
# Sample configuration for inference kernel benchmarks

# For multiple quantizations and shapes
quantizations:
  - "baseline"
  - "int8wo"
  - "int4wo-128"
  - "int4wo-128-hqq"

model_params:
  matrix_shapes:
    - name: "custom"
      shapes: [
        [1024, 1024, 1024],  # [m, k, n]
        [2048, 4096, 1024],
        [4096, 4096, 1024]
      ]
    # - name: "llama" # Pre-defined shapes like llama, square, other-model shapes, etc.
  precision: "torch.bfloat16"
  compile: false
  device: "cuda"
  model_type: "linear"
```


benchmarks/microbenchmarks/bench_inference_kernels.py
```diff
@@ -0,0 +1,144 @@
...
+    float8_weight_only,
+    int4_weight_only,
+    int8_weight_only,
+    quantize_,
+)
+
+
+def parse_quantization_arg(quantization_input: List[str]):
+    # Define regex patterns for quantization techniques
+    patterns = {
```
jainapurva, line 22: Will be adding more patterns here, for all quant apis

PR #1708, opened 2025-02-13
SAM2: Use torch.export for VOS

cpuhrsch:


cpuhrsch:
Unrelated reformatting required: `Would reformat: torchao/float8/config.py`

PR #1711, opened 2025-02-13
update torchao READMEs with new configuration APIs

vkuzo:
Summary:

This updates the README files with the names of the new workflow
configuration APIs.

Test Plan: CI

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo:
* (to be filled)

drisspg (approved):

README.md
```diff
@@ -29,16 +29,16 @@ For inference, we have the option of
...
+quantize_(m, Int4WeightOnlyConfig())
 ```
 
-For gpt-fast `int4_weight_only()` is the best option at bs=1 as it **2x the tok/s and reduces the VRAM requirements by about 65%** over a torch.compiled baseline.
+For gpt-fast `Int4WeightOnlyConfig()` is the best option at bs=1 as it **2x the tok/s and reduces the VRAM requirements by about 65%** over a torch.compiled baseline.
 
-If you don't have enough VRAM to quantize your entire model on GPU and you find CPU quantization to be too slow then you can use the device argument like so `quantize_(model, int8_weight_only(), device="cuda")` which will send and quantize each layer individually to your GPU.
+If you don't have enough VRAM to quantize your entire model on GPU and you find CPU quantization to be too slow then you can use the device argument like so `quantize_(model, Int8WeightOnlyConfig(), device="cuda")` which will send and quantize each layer individually to your GPU.
 
 If you see slowdowns with any of these techniques or you're unsure which option to use, consider using [autoquant](./torchao/quantization/README.md#autoquantization) which will automatically profile layers and pick the best way to quantize each layer.
```
andrewor14, line 43: I think this file is missing a few, e.g. I still see `intx_quantization_aware_training`, `int8_weight_only_quantized_training`, and `fpx_weight_only`

vkuzo: thank you! I just updated `intx_quantization_aware_training` and `fpx_weight_only`.  

`int8_weight_only_quantized_training` is a prototype API and is not migrated yet - we can do this after the 0.9.0 release.

PR #1719, opened 2025-02-14
Make TorchAO cpp/Python extension

drisspg:
Summary: Adds the ability to build our Native code

Differential Revision: D69634772

### DIFF
--- a/torchao/csrc/cuda/rowwise_scaled_linear_cutlass/rowwise_scaled_linear_cutlass_s8s4.cu
+++ b/torchao/csrc/cuda/rowwise_scaled_linear_cutlass/rowwise_scaled_linear_cutlass_s8s4.cu
@@ -1,5 +1,4 @@
 #include <torch/library.h>
-
 #include "rowwise_scaled_linear_cutlass.cuh"
 
 namespace torchao {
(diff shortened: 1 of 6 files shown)


torchao/csrc/cuda/rowwise_scaled_linear_cutlass/rowwise_scaled_linear_cutlass_s8s4.cu
```diff
@@ -13,11 +12,16 @@ rowwise_scaled_linear_cutlass_s8s4(
               __func__, " : The input datatypes combination ", xq.dtype(),
               " for xq and ", wq.dtype(), " for wq is not supported");
 
-  // Dispatch to appropriate kernel template.
+#if defined(BUILD_ROWWISE_SCALED_LINEAR_CUTLASS)
```
alexsamardzic, line 15: Was there a specific reason for this change, and alike in the "s4s4" file?  The .cuh header file included in both of them contains checking of `BUILD_ROWWISE_SCALED_LINEAR_CUTLASS`, so that it doesn't have to be duplicated here.  Also, in my upcoming PR, I've added `#include <cutlass/cutlass.h>` if that was the problem for referring to `cutlass::int4b_t` datatype.

drisspg: The problem was the headers weren't included wihen this flag isn't defiend so using the namespace didnt work

So once you add feel free to undo this

alexsamardzic: Got it, thanks for fixing it!

PR #1731, opened 2025-02-18
float8 training: make the "config from recipe" API polished

vkuzo:
Summary:

This PR makes the API that takes a recipe name (enum or string) and
returns a `Float8LinearConfig` instance more polished and ready for
usage in README.md docs and by partner callsites such as torchtitan and
torchtune.

Test Plan:

```
```

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo:
* (to be filled)

torchao/float8/config.py
```diff
@@ -146,6 +146,13 @@ class Float8GemmConfig:
     use_fast_accum: bool = False
 
 
+# Pre-made recipes for common configurations
+class Float8LinearRecipeName(enum.Enum):
+    TENSORWISE = "tensorwise"
+    ROWWISE = "rowwise"
+    ROWWISE_WITH_GW_HP = "rowwise_with_gw_hp"
```
HDCharles, line 153: nit: wondering if gw_hp should be clarified somewhere as far as what it means?
it might also make sense to allow a fully written out version

HDCharles (approved):

Issue #1690, opened 2025-02-10
migration of `quantize_` workflow configuration from callables to configs


vkuzo:
# summary

This issue tracks the migration of `quantize_` per-workflow configuration from Callables to configs..  

We are migrating the way `quantize_` workflows are configured from callables (tensor subclass inserters) to direct configuration (config objects).  Motivation: align with the rest of the ecosystem, enable inspection of configs after instantiation, remove a common source of confusion.

**What is changing:**

Specifically, here is how the signature of `quantize_`'s second argument will change:

```python
#
# torchao v0.8.0 and before
#
def quantize(
    model: torch.nn.Module,
    apply_tensor_subclass: Callable[[torch.nn.Module], torch.nn.Module],
    ...,
): ...

#
# torchao v0.9.0
#
def quantize(
    model: torch.nn.Module,
    config: Union[AOBaseConfig, Callable[[torch.nn.Module], torch.nn.Module]],
    ...,
): ...

#
# torchao v0.10.0 or later (exact version TBD)
#
def quantize(
    model: torch.nn.Module,
    config: AOBaseConfig,
    ...,
): ...
```

1. the name of the second argument to `quantize_` changed from `apply_tensor_subclass` to `config`.  Since the vast majority of callsites today are passing in configuration with a positional argument, this change should not affect most people.
2. the type of the second argument to `quantize_` will change from `Callable[[torch.nn.Module], torch.nn.Module]` to `config: AOBaseConfig`, following a deprecation process detailed below.
3. for individual workflows, the user facing API name changed from snake case (`int8_weight_only`) to camel case (`Int8WeightOnlyConfig`).  All argument names for each config are kept as-is.  We will keep the old snake case names (`int8_weight_only`) around and alias them to the new names (`int8_weight_only = Int8WeightOnlyConfig`), to avoid breaking callsites.  We plan to keep the old names forever.  Here are all the workflow config name changes:

| old name (will keep working) | new name (recommended) |
| --- | --- |
| `int4_weight_only` | `Int4WeightOnlyConfig` |
| `float8_dynamic_activation_float8_weight` | `Float8DynamicQuantizationFloat8WeightConfig` |
| `float8_static_activation_float8_weight` | `Float8StaticActivationFloat8WeightConfig` |
| `float8_weight_only` | `Float8WeightOnlyConfig` |
| `fpx_weight_only` | `FPXWeightOnlyConfig` |
[...]

PR #1734, opened 2025-02-19
promote blocksparse from prototype, make it faster

jcaip:
This PR promotes block sparsity from prototype in torchao. 

Chiefly, it ports over the triton addmm blocksparse kernels from core, and makes several performance improvements to them. 

All of the numbers reported below are for an H100, with blocksize=64 and sparsity_level=0.9. The default dense baseline is 134 tok/s

1) Adds padding support to the triton kernel for dense matrices with dimension < 16, like those we run into during decoding. (214 -> 218 tok/s)
2) Changes the default [num_stages](https://github.com/triton-lang/triton/discussions/512) parameter from 1 to 4. This has a large effect on performance, and it seemed like the default kernel autotuning either does not modify or deems this parameter to be unimportant for some reason. (218 -> 263 tok/s). 
3) Adds an env_var, BSR_AUTOTUNE, that users can use if they want to do kernel autotuning on top of the default parameters. (263 -> 266 tok/s) This seems to matter more for bs=n compute bound workloads, where I see a reduction from 0.3855 to 0.3745s on bs=8192 prefill (roughly 3%)

So in total we are seeing a **1.985x** speedup 🚀

I've also updated the documentation to not reference prototype - planning on updating the diagram in a subsequent PR. 

### Testing
I added a new test case for the padding inputs and moved the test file out of prototype. 
```
python test/sparsity/test_sparse_api.py
```

### Benchmarking
```
export MODEL_REPO=meta-llama/Meta-Llama-3.1-8B

python generate.py --checkpoint_path $CHECKPOINT_PATH/$MODEL_REPO/model.pth --compile --compile_prefill --write_result benchmark_results.txt --prefill_size 8192 --profile baseline_prefill
python generate.py --checkpoint_path $CHECKPOINT_PATH/$MODEL_REPO/model.pth --compile --compile_prefill --write_result benchmark_results.txt --prefill_size 8192 --sparsity bsr --profile bsr_prefill
python generate.py --checkpoint_path $CHECKPOINT_PATH/$MODEL_REPO/model.pth --compile --compile_prefill --write_result benchmark_results.txt --profile baseline
python generate.py --checkpoint_path $CHECKPOINT_PATH/$MODEL_REPO/model.pth --compile --compile_prefill --write_result benchmark_results.txt --sparsity bsr --profile bsr
```


vkuzo:
might be good to consider getting the changes from https://github.com/pytorch/ao/issues/1690 in here since you are making a major API change, it will save you a migration in the future.

cpuhrsch (approved):

PR #1742, opened 2025-02-20
Allow for scales to be in new e8m0 dtype

drisspg:
Stacked PRs:
 * #1762
 * __->__#1742

--- --- ---

Allow for scales to be in new e8m0 dtype

### DIFF
--- a/torchao/ops.py
+++ b/torchao/ops.py
@@ -1,3 +1,5 @@
+import functools
+
 import torch
 from torch import Tensor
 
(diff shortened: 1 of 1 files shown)


torchao/ops.py
```diff
@@ -606,6 +608,26 @@ def _(
...
 
 
+@functools.lru_cache()
+def _get_dtypes():
+    if hasattr(torch, "float8_e8m0fnu"):
+        return (torch.uint8, torch.float8_e8m0fnu)
+    return (torch.uint8,)
+
+
+def _check_scale_dtypes(A_scale, B_scale):
```
vkuzo, line 618: would be good to delete this later to reduce complexity and just support e8m0

drisspg: totally agree

PR #1746, opened 2025-02-20
Validate binaries on all OS's

atalman:


.github/workflows/validate-binaries.yml
```diff
@@ -40,7 +44,6 @@ jobs:
     with:
       package_type: "wheel"
       version: ${{ inputs.version }}
-      os: "linux"
       channel: ${{ inputs.channel }}
       repository: "pytorch/ao"
       with_cuda: "enable"
```
HDCharles, line 49: does this need to be changed or does the validate-binaries file handle enabling/disabling cuda?

PR #1751, opened 2025-02-21
Fix potential out-of-bound access in int8_mm.py

mark14wu:
This PR fixes a bug in `_scaled_int8_mm_kernel` allowing unmasked `tl.load()` calls when `K` was even, but not necessarily a multiple of `BLOCK_K`, causing out-of-bounds reads. The fix changes the check to `K % BLOCK_K == 0`, ensuring unmasked loads occur only when the last block is fully valid.

### DIFF
--- a/torchao/prototype/quantized_training/int8_mm.py
+++ b/torchao/prototype/quantized_training/int8_mm.py
@@ -79,7 +79,7 @@ def _scaled_int8_mm_kernel(
     BLOCK_N: tl.constexpr,
     BLOCK_K: tl.constexpr,
     GROUP_M: tl.constexpr = 8,
-    EVEN_K: tl.constexpr = False,
+    EVEN_K: tl.constexpr = True,
     COL_SCALE_SCALAR: tl.constexpr = False,
 ):
     # based on triton.ops.matmul
--- a/torchao/prototype/quantized_training/int8_mm.py
+++ b/torchao/prototype/quantized_training/int8_mm.py
@@ -99,7 +99,7 @@ def _scaled_int8_mm_kernel(
 
     acc = tl.zeros((BLOCK_M, BLOCK_N), dtype=tl.int32)
     for k in range(K, 0, -BLOCK_K):
-        if EVEN_K:
+        if K % BLOCK_K == 0:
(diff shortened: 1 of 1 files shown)


mark14wu:
cc @Jokeren

Jokeren:
@mark14wu A clearer fix can be done using `@triton.heuristics`

gau-nernst:
Thank you for the interest! Didn't know there are people looking at this code 😄.

I copied this code from PyTorch Inductor though. Does that mean Inductor is susceptible to this bug too? (or inductor allocates padded tensors so it's not an issue?)

Like @Jokeren said, `@triton.heuristic` sounds like a good idea. We would want to keep a `tl.constexpr` there so that we don't do `if` for every iteration.

Another question. It seems like we might do out-of-bound reads for M and N dims too?

Jokeren:
> I copied this code from PyTorch Inductor though. Does that mean Inductor is susceptible to this bug too?

Yes indeed, but our tool, a triton sanitizer, cannot work for pytorch inductor at this moment. Hao may want to explore this direction when he starts intern at meta this summer. 

> It seems like we might do out-of-bound reads for M and N dims too?

M and N dims are safe because we have protections with `rm%M` and `rn%N`.

@mark14wu please go ahead refactor the code with triton.heuristic

drisspg:
There are ways to turn inductor code into triton only repos

mark14wu:
Now refactored with `triton.heuristics`.

torchao/prototype/quantized_training/int8_mm.py
```diff
@@ -54,6 +54,11 @@
 
 
 @triton.autotune(configs=configs, key=["M", "N", "K", "stride_ak", "stride_bk"])
+@triton.heuristics(
+    {
+        "EVEN_K": lambda args: args["K"] % args["BLOCK_M"] == 0
```
gau-nernst, line 59: Shouldn't it be `BLOCK_K`?

mark14wu: Sorry, it's a typo. Fixed now.

PR #1759, opened 2025-02-21
Micro-benchmark inference

jainapurva:
This is the first PR, in the benchmarking effort. It provides the outline to setup inference microbenchmarking for quant api's in torchao.

The different inputs like quantization techniques, matrix sizes, compile, sparsity etc, will be given as input to the python script. The options are re-defined in the scipt for quantization techniques, and a developer can add new quant technique. The script will generate a csv with performance numbers, and that'll be used to plot charts and as an input to dashboard . The script performs the following tasks:

- Take input as .yml
- Performs benchmarking for quantize_ APIs eval time for configurations
- Record all the config params and their respective time in csv file.
- Test cases

_Future PRs will include more config options and process the generated results._

### benchmark_config.yml
```
# Sample configuration for inference kernel benchmarks
quantization_config_recipe_names:
  - "baseline"  # String format is same as llama/generate.py
  - "int8wo"
  - "int4wo-128"
output_dir: "benchmarks/microbenchmarks/test/results"  # Directory for results and plots
model_params:
  matrix_shapes:
    - name: "custom"
      shapes: [
        [1024, 1024, 1024],  # [m, k, n]
        [2048, 4096, 1024],
        [4096, 4096, 1024]
      ]
  high_precision_dtype: "torch.bfloat16"
  use_torch_compile: true
  torch_compile_mode: "max-autotune"
  device: "cuda"  # Change this to "cuda", "mps", "xpu", or "cpu" as needed
  model_type: "linear"
```

### Run command:
```
python benchmarks/microbenchmarks/benchmark_runner.py --config benchmarks/microbenchmarks/benchmark_config.yml
```
#### Output will be stored in 
```
benchmarks/microbenchmarks/results/results.csv
```


benchmarks/microbenchmarks/test/benchmark_config.yml
```diff
@@ -0,0 +1,19 @@
...
+output_dir: "benchmarks/microbenchmarks/test/results"  # Directory for results and plots
+model_params:
+  matrix_shapes:
+    - name: "custom"
+      shapes: [
+        [1024, 1024, 1024],  # [m, k, n]
+        [2048, 4096, 1024],
+        [4096, 4096, 1024]
+      ]
+  precision: "torch.bfloat16"
```
vkuzo, line 16: nit: there are many different precisions in this context.  Maybe this can be named more specifically, such as "high_precision_dtype"?

benchmarks/microbenchmarks/test/benchmark_config.yml
```diff
@@ -0,0 +1,19 @@
...
+  compile: "max-autotune"
```
vkuzo, line 17: `compile_mode`?

benchmarks/microbenchmarks/test/benchmark_config.yml
```diff
@@ -0,0 +1,19 @@
+# Sample configuration for inference kernel benchmarks
+quantizations:
```
vkuzo, line 2: `quantization_config_recipe_names`?

benchmarks/microbenchmarks/test/benchmark_config.yml
```diff
@@ -0,0 +1,19 @@
...
+  - "baseline"
+  - "int8wo"
+  - "int4wo-128"
+  - "int4wo-128-hqq"
+output_dir: "benchmarks/microbenchmarks/test/results"  # Directory for results and plots
+model_params:
+  matrix_shapes:
+    - name: "custom"
+      shapes: [
```
vkuzo, line 11: how would this work for layers where there is more than one linear, and the weights have different shapes?

jainapurva: The code currently supports only a single linear layer model type, in future PRs, it'll be able to support different models, including more than one linear layer, with different weight shapes. 

The input in shapes would be a list of lists, where each [m,k,n] signifies one linear layer.
Example:
```
quantizations:
  - "baseline"
  - "int8wo"
  - "int4wo-128"
  - "int4wo-128-hqq"
output_dir: "benchmarks/microbenchmarks/test/results"  # Directory for results and plots
model_params:
  matrix_shapes:
    - name: "custom"
      shapes: [
        [[1024, 1024, 1024],  # Each entry signifies the shapes to benchmark on for the model. 
        [2048, 4096, 1024]], # Each [m,k,n] is for a single linear layer
        [[4096, 4096, 1024],
        [1024, 1024, 1024]]
      ]
  precision: "torch.bfloat16"
  compile: "max-autotune"
  device: "cuda"  # Change this to "cuda", "mps", "xpu", or "cpu" as needed
  model_type: "double-linear"
```
Each entry is a list of list for different linear layers
[[1024, 1024, 1024], 
 [2048, 4096, 1024]],

benchmarks/microbenchmarks/benchmark_inference.py
```diff
@@ -0,0 +1,67 @@
...
+        config.m,
+        config.k,
+        config.n,
+        dtype=config.precision,
+        device=config.device,
+    )
+
+    # Use quantize_ to apply each quantization function to the model
+    m_copy = deepcopy(base_model).eval().to(config.device)
+    m_copy = quantize_model(m_copy, config.quantization)
```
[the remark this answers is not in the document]
jainapurva: The logic has multiple if/else, and will become more cumbersome with params like sparsity, autoquant techniques. Also, if we abstract it, we might be able to use it for e2e models also. Reducing the number of places a developer might need to make changes, and register their kernels/apis

benchmarks/microbenchmarks/benchmark_inference.py
```diff
@@ -0,0 +1,67 @@
+"""
+Inference benchmark runner
+
+This script runs inference benchmarks and generates a micro-benchmarking report for it.
+- run() function is the main entry point for running inference benchmarks.
+"""
+
+from copy import deepcopy
```
drisspg, line 8: can you add a readme describing the local flow for users

jainapurva: Sure, will also add the flow diagram to it, and example.

benchmarks/microbenchmarks/benchmark_inference.py
```diff
@@ -0,0 +1,67 @@
...
+    # Use quantize_ to apply each quantization function to the model
+    m_copy = deepcopy(base_model).eval().to(config.device)
+    m_copy = quantize_model(m_copy, config.quantization)
+
+    if config.compile:
+        print("Compiling model....")
+        m_copy = torch.compile(m_copy, mode=config.compile_mode, fullgraph=True)
+
+    # Run benchmarks
+    result = {**config.__dict__}
```
drisspg, line 48: why not just call to_dict?

benchmarks/microbenchmarks/test/test_benchmark_inference.py
```diff
@@ -0,0 +1,36 @@
...
+            "compile": False,
+            "device": "cpu",  # Use CPU for testing
+            "model_type": "linear",
+        }
+        self.config = BenchmarkConfig(
+            quantization="baseline",
+            params=self.params,
+            shape_name="test",
+            shape=[16, 32, 8],  # Small shape for testing
+            output_dir="benchmarks/microbenchmarks/test/test_output/",
```
drisspg, line 20: For testing maybe use a tmp dir

benchmarks/microbenchmarks/utils.py
```diff
@@ -0,0 +1,380 @@
...
+def ffn_or_attn_only(mod, fqn):
+    return isinstance(mod, torch.nn.Linear) and (
+        "feed_forward" in fqn or "attention" in fqn
+    )
+
+
+def quantize_model(
+    model: torch.nn.Module,
+    quantization: str,
+    **kwargs,
```
vkuzo, line 131: I wouldn't include this unless it's used, it isn't used in this PR.

I also would prefer either everything as string or everything as arguments, mixing both is confusing.

benchmarks/microbenchmarks/benchmark_inference.py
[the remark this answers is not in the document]
vkuzo: I think if you name it something like

```
quantization_string_to_quantized_model(m_copy, config.quantization_string)
```

it would be a lot easier to understand

benchmarks/microbenchmarks/utils.py
```diff
@@ -0,0 +1,380 @@
...
+            model, Float8DynamicActivationFloat8WeightConfig(granularity=granularity)
+        )
+    else:
+        if not TORCH_VERSION_AT_LEAST_2_5:
+            unwrap_tensor_subclass(model)
+    return model
+
+
+# Function to benchmark model evaluation - e2e eval run
+def benchmark_model_inference_in_microseconds(model, input_data):
```
HDCharles, line 290: this entire functions is effectively reinventing the wheel, you should use one of the benchmark functions e.g. 

https://github.com/pytorch/ao/blob/main/torchao/quantization/autoquant.py#L343 or 
https://pytorch.org/docs/stable/benchmark_utils.html

benchmarks/microbenchmarks/utils.py
```diff
@@ -0,0 +1,380 @@
...
+    # Returns model run time in seconds
+    if torch.cuda.is_available():
+        torch.cuda.synchronize()
+
+    # warm up
+    for _ in range(2):
```
HDCharles, line 296: for example the warmup without torch.no_grad vs run with torch.no_grad is going to create diverging behaviors.

also, in practice if the first run of the model is where compile actually occurs, you're not really warming things up.

benchmarks/microbenchmarks/test/benchmark_config.yml
```diff
@@ -0,0 +1,20 @@
+# Sample configuration for inference kernel benchmarks
+quantization_config_recipe_names:
+  - "baseline"
+  - "int8wo"
+  - "int4wo-128"
+  - "int4wo-128-hqq"
```
HDCharles, line 6: accuracy techniques like hqq shouldn't be in benchmarks. Its the same kernel as the other, just a different quant technique.

benchmarks/microbenchmarks/utils.py
```diff
@@ -0,0 +1,380 @@
...
+    return isinstance(mod, torch.nn.Linear) and not ffn_only(mod, fqn)
+
+
+def ffn_or_attn_only(mod, fqn):
+    return isinstance(mod, torch.nn.Linear) and (
+        "feed_forward" in fqn or "attention" in fqn
+    )
+
+
+def quantize_model(
```
vkuzo, line 128: after looking in more detail at what this is doing, how about the following instead:

```
# convert from string to AOBaseConfig
def string_to_config(quant_string: str) -> AOBaseConfig

# that's it, once you have a config, you can pass it around and call quantize_|sparsify_directly on it from your callsite
```

benchmarks/microbenchmarks/utils.py
```diff
@@ -0,0 +1,380 @@
...
+    sparsity = kwargs.get("sparsity", None)
+    precision = kwargs.get("precision", None)
+
+    # Quantization techniques
+    if "baseline" in quantization:
+        return model
+    if "int8wo" in quantization:
+        quantize_(model, Int8WeightOnlyConfig())
+    if "int8dq" in quantization:
+        if sparsity and "semi" in sparsity:
```
vkuzo, line 154: if this is doing "apply sparsity and then apply quantization", IMO it's cleaner to just have two config strings and call your API twice using the same function, instead of trying to branch here

benchmarks/microbenchmarks/utils.py
```diff
@@ -0,0 +1,380 @@
...
+
+    if torch.cuda.is_available():
+        torch.cuda.empty_cache()
+        torch.cuda.reset_peak_memory_stats()
+
+    if hasattr(torch, "_dynamo"):
+        torch._dynamo.reset()
+
+
+def generate_results_csv(
```
vkuzo, line 355: nit: can we also pretty-print to stdout?  most of the time I just want to see the numbers in stdout and I don't need to export them elsewhere

benchmarks/microbenchmarks/utils.py
jainapurva: Having everything as string makes sense. Once we finalize https://github.com/pytorch/ao/pull/1774, I can use the same for benchmarking.

benchmarks/microbenchmarks/utils.py
jainapurva: Done

benchmarks/microbenchmarks/utils.py
jainapurva: Yes, I feel Timer serves the usecase here, and added a first run for warmup

benchmarks/microbenchmarks/utils.py
jainapurva: Makes sense. quantize_ has a config, hence supporting that now, but sparsify_ doesn't support AOBaseConfig yet. Will add the support for sparsity once config is added https://github.com/pytorch/ao/pull/1856

benchmarks/microbenchmarks/test/benchmark_config.yml
```diff
@@ -0,0 +1,19 @@
...
+model_params:
+  matrix_shapes:
+    - name: "custom"
+      shapes: [
+        [1024, 1024, 1024],  # [m, k, n]
+        [2048, 4096, 1024],
+        [4096, 4096, 1024]
+      ]
+  high_precision_dtype: "torch.bfloat16"
+  compile: true
```
HDCharles, line 16: nit: I tend to prefer to consolidate multiple variables into 1, like compile vs compile_mode is kind of redundant when you could just have 

compile: "max-autotune" or compile: "false"

HDCharles (approved):
lgtm

benchmarks/microbenchmarks/README.md
```diff
@@ -0,0 +1,83 @@
...
+- **utils.py**: Contains utility functions and configuration classes
+- **test\/**: Test files and sample configurations
+
+## Usage
+
+1. Create a configuration YAML file (see example below)
+2. Run the benchmark using:
+
+```bash
+python -m benchmarks.microbenchmarks.benchmark_runner --config path/to/config.yml
```
vkuzo, line 26: is there a way to use without a yaml? IMO there should be

benchmarks/microbenchmarks/benchmark_inference.py
```diff
@@ -0,0 +1,74 @@
...
+            m_copy = torch.compile(m_copy, mode=config.compile, fullgraph=True)
+
+    # Run benchmarks
+    result = {**config.to_dict()}
+
+    # Benchmark time to run an inference call for quantized model
+    model_time = benchmark_model_inference_in_microseconds(
+        model=m_copy, input_data=input_data
+    )
+    result["benchmark_model_inference_in_microseconds"] = model_time
```
vkuzo, line 61: it would be nice to also measure peak memory usage

vkuzo, line 61: nit: "benchmark_model_inference_in_microseconds" sounds like a function name, I think a name such as "model_inference_time_ms" is more descriptive of this value

benchmarks/microbenchmarks/test/benchmark_config.yml
```diff
@@ -0,0 +1,18 @@
...
+model_params:
+  matrix_shapes:
+    - name: "custom"
+      shapes: [
+        [1024, 1024, 1024],  # [m, k, n]
+        [2048, 4096, 1024],
+        [4096, 4096, 1024]
+      ]
+  high_precision_dtype: "torch.bfloat16"
+  compile: "max-autotune"
```
vkuzo, line 16: this is confusing, it is not clear what "compile" means here and this variable name does not match argument names from the `torch.compile` API.  IMO something like `use_torch_compile` and `torch_compile_mode='max_autotune'` is clearer.

benchmarks/microbenchmarks/README.md
jainapurva: yml gives us options to add many use cases simultaneously, with multiple parameters. In generate we couldn't compare multiple quantizations, also writing a long terminal cmd used to be cumbersome.

benchmarks/microbenchmarks/benchmark_inference.py
jainapurva: Yes, memory usage and other performance metrics will be added in follow-up diffs

benchmarks/microbenchmarks/utils.py
```diff
@@ -0,0 +1,381 @@
...
+
+    Args:
+        quantization (str): quantization method to be used
+        **kwargs: additional arguments to be passed to the quantization method
+
+    Returns:
+        AOBaseConfig: Quantization configuration object
+    """
+    high_precision_dtype = kwargs.get("high_precision_dtype", torch.bfloat16)
+    if "int4wo" in quantization and not HAS_TRITON:
```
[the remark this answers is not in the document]
jainapurva: This format is being followed in llama/generate.py, hence I kept the same string format until a standard string format is defined in #1774.

benchmarks/microbenchmarks/README.md
vkuzo: when using a benchmark for performance work ("speed up kernel x"), it's way more convenient to use command line arguments compared to changing yaml files

benchmarks/microbenchmarks/test/benchmark_config.yml
vkuzo: two arguments sounds right to me, IMO the number of "new arguments to remember" should be minimized, and anything that matches the same args from other functions should match the arg names from those functions exactly.  It's more LOC, but it's easier to understand with less ambiguity.

benchmarks/microbenchmarks/utils.py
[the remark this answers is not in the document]
vkuzo: makes sense, add a comment with this context?

benchmarks/microbenchmarks/README.md
jainapurva: As discussed offline, command line arguments can be added, but I would prefer doing that in following PRs. I'll add a few TODO's in PR, to give an idea on how and where it would be added.

benchmarks/microbenchmarks/utils.py
```diff
@@ -0,0 +1,381 @@
...
+        self.model_type = params.get("model_type", "linear")
+        self.output_dir = output_dir
+        self.name = f"benchmark_{self.quantization}_{self.model_type}_m{self.m}_k{self.k}_n{self.n}{'_compile' if self.compile else ''}"
+
+    @staticmethod
+    def _parse_precision(precision_str: str) -> torch.dtype:
+        """Convert string precision to torch dtype"""
+        return getattr(torch, precision_str.split(".")[-1])
+
+    def to_dict(self) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
jainapurva: We're not storing the results in config. Currently I'm just getting the config details in the results dictionary, and then adding the results to it, but it's not modifying the BenchmarkConfig object. I do think it would be more convenient for processing if I add a BenchmarksResults class, with config as one attribute, and the other attributes for model_inference_time_in_ms, peak_memory_usage, etc.

PR #1760, opened 2025-02-22
Move import into function definition

jcaip:
To fix builds broken on pytest import


jcaip:
This will be fixed in https://github.com/pytorch/ao/pull/1761 instead

PR #1784, opened 2025-02-26
Move torchao/_models to benchmarks/_models

jainapurva:
Move torchao/_models to benchmarks/_models. The benchmarks folder will be main folder for all models and benchmarkings in AO


torchao/prototype/quantization/mixed_precision/scripts/BO_acc_throughput.py
```diff
@@ -18,15 +18,17 @@
...
-    _load_model,
+from torchao._models.generate import (
     decode_one_token,
-    device_sync,
-    encode_tokens,
     prefill,
 )
-from torchao._models.llama.model import Transformer, prepare_inputs_for_model
-from torchao._models.llama.tokenizer import get_tokenizer
+from torchao._models.llm.model import Transformer, prepare_inputs_for_model
```
jerryzh168, line 25: nit: why `llm` instead of `llama`

test/quantization/test_gptq_mt.py
```diff
@@ -5,8 +5,8 @@
 import torch.nn.functional as F
 from torch.testing._internal.common_utils import run_tests
 
-from torchao._models.llama.model import Transformer, prepare_inputs_for_model
-from torchao._models.llama.tokenizer import get_tokenizer
+from torchao._models.llm.model import Transformer, prepare_inputs_for_model
```
jerryzh168, line 8: why these can't be moved to under benchmark as well?

torchao/prototype/quantization/mixed_precision/scripts/BO_acc_throughput.py
jainapurva: The config has mistral and other models also, hence thought llm would be a better name than llama.

https://github.com/pytorch/ao/blob/677aca7968a45355363652e51398265fdaf87d49/torchao/_models/llm/model.py#L74-L158

jerryzh168: maybe the config can live somewhere else, `llm` might be a bit too general I feel

jainapurva: Ok, makes sense

drisspg (approved):

PR #1785, opened 2025-02-26
metal lowbit kernels: pip install

manuelcandales:
Integrate lowbit metal kernels into setup.py installation

```
USE_CPP=1 TORCHAO_BUILD_EXPERIMENTAL_MPS=1 pip install .
```

Running the above command under the `ao` directory on a Mac will also build the metal kernels.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -75,6 +75,20 @@ from torch.utils.cpp_extension import (
     CUDAExtension,
 )
 
+build_torchao_experimental_mps = (
+    os.getenv("TORCHAO_BUILD_EXPERIMENTAL_MPS") == "1"
+    and build_torchao_experimental
+    and torch.mps.is_available()
+)
+
+if os.getenv("TORCHAO_BUILD_EXPERIMENTAL_MPS") == "1":
+    if use_cpp != "1":
+        print("Building experimental MPS ops requires USE_CPP=1")
+    if not platform.machine().startswith("arm64") or platform.system() != "Darwin":
+        print("Experimental MPS ops require Apple Silicon.")
+    if not torch.mps.is_available():
+        print("MPS not available. Skipping compilation of experimental MPS ops.")
+
 # Constant known variables used throughout this file
 cwd = os.path.abspath(os.path.curdir)
 third_party_path = os.path.join(cwd, "third_party")
@@ -174,6 +188,8 @@ class TorchAOBuildExt(BuildExtension):
         if not os.path.exists(self.build_temp):
             os.makedirs(self.build_temp)
 
+        build_mps_ops = "ON" if build_torchao_experimental_mps else "OFF"
+
         subprocess.check_call(
             [
                 "cmake",
@@ -181,8 +197,10 @@ class TorchAOBuildExt(BuildExtension):
                 "-DCMAKE_BUILD_TYPE=" + build_type,
                 # Disable now because 1) KleidiAI increases build time, and 2) KleidiAI has accuracy issues due to BF16
                 "-DTORCHAO_BUILD_KLEIDIAI=OFF",
+                "-DTORCHAO_BUILD_MPS_OPS=" + build_mps_ops,
                 "-DTorch_DIR=" + torch_dir,
                 "-DCMAKE_LIBRARY_OUTPUT_DIRECTORY=" + extdir,
+                "-DCMAKE_INSTALL_PREFIX=cmake-out",
             ],
             cwd=self.build_temp,
         )
--- a/torchao/experimental/ops/mps/test/test_lowbit.py
+++ b/torchao/experimental/ops/mps/test/test_lowbit.py
@@ -10,10 +10,7 @@ import unittest
 import torch
 from parameterized import parameterized
 
-libname = "libtorchao_ops_mps_aten.dylib"
-libpath = os.path.abspath(
-    os.path.join(os.path.dirname(__file__), "../cmake-out/lib/", libname)
-)
+import torchao  # noqa: F401
 
 try:
     for nbit in range(1, 8):
@@ -21,6 +18,10 @@ try:
         getattr(torch.ops.torchao, f"_pack_weight_{nbit}bit")
 except AttributeError:
     try:
+        libname = "libtorchao_ops_mps_aten.dylib"
+        libpath = os.path.abspath(
+            os.path.join(os.path.dirname(__file__), "../cmake-out/lib/", libname)
+        )
         torch.ops.load_library(libpath)
     except:
         raise RuntimeError(f"Failed to load library {libpath}")
--- a/.gitignore
+++ b/.gitignore
@@ -375,3 +375,4 @@ checkpoints/
 
 # Experimental
 torchao/experimental/cmake-out
+torchao/experimental/deps
--- a/torchao/experimental/CMakeLists.txt
+++ b/torchao/experimental/CMakeLists.txt
@@ -16,6 +16,7 @@ if (NOT CMAKE_BUILD_TYPE)
 endif()
 
 option(TORCHAO_BUILD_EXECUTORCH_OPS "Building torchao ops for ExecuTorch." OFF)
+option(TORCHAO_BUILD_MPS_OPS "Building torchao MPS ops" OFF)
 
 
 if(NOT TORCHAO_INCLUDE_DIRS)
@@ -51,6 +52,12 @@ if(CMAKE_SYSTEM_PROCESSOR STREQUAL "arm64")
         torchao_ops_linear_8bit_act_xbit_weight_aten
         torchao_ops_embedding_xbit_aten
     )
+    if (TORCHAO_BUILD_MPS_OPS)
+        message(STATUS "Building with MPS support")
+        add_subdirectory(ops/mps)
+        target_link_libraries(torchao_ops_aten PRIVATE torchao_ops_mps_aten)
+    endif()
+
     install(
         TARGETS torchao_ops_aten
         EXPORT _targets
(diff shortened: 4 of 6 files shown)


setup.py
```diff
@@ -174,15 +174,19 @@ def build_cmake(self, ext):
         if not os.path.exists(self.build_temp):
             os.makedirs(self.build_temp)
 
+        build_mps_ops = "ON" if torch.mps.is_available() else "OFF"
```
metascroy, line 177: Should we have an environment variable, e.g., TORCHAO_BUILD_MPS that can be toggled?

We plan to start moving the CPU kernels out of experimental, and we would like to build those without building MPS until they too are moved out of experimental.

torchao/__init__.py
```diff
@@ -35,6 +35,10 @@
     # running the script `torchao/experimental/build_torchao_ops.sh <aten|executorch>`
     # For more information, see https://github.com/pytorch/ao/blob/main/torchao/experimental/docs/readme.md
     experimental_lib = list(Path(__file__).parent.glob("libtorchao_ops_aten.*"))
+    if len(experimental_lib) == 0 and Path.cwd().name == "ao":
+        experimental_lib = list(
+            (Path(__file__).parent.parent / "build").rglob("libtorchao_ops_aten.*")
```
metascroy, line 40: Where is build?  Is this the folder where MPS kernels are placed?

.github/workflows/torchao_experimental_test.yml
```diff
@@ -36,6 +36,7 @@ jobs:
           pip install --extra-index-url "https://download.pytorch.org/whl/nightly/cpu" torch=="2.6.0.dev20250104"
           pip install numpy
           pip install pytest
+          pip install pyyaml
           USE_CPP=1 pip install .
       - name: Run python tests
```
metascroy, line 41: Should you add the MPS tests to CI?

setup.py
```diff
@@ -174,15 +174,19 @@ def build_cmake(self, ext):
...
+
         subprocess.check_call(
             [
                 "cmake",
                 ext.sourcedir,
                 "-DCMAKE_BUILD_TYPE=" + build_type,
                 # Disable now because 1) KleidiAI increases build time, and 2) KleidiAI has accuracy issues due to BF16
                 "-DTORCHAO_BUILD_KLEIDIAI=OFF",
+                "-DTORCHAO_BUILD_MPS_OPS=" + build_mps_ops,
```
manuelcandales, line 186: I did add a flag. So, you aren't forced to build the MPS ops always

setup.py
manuelcandales: I added a flag, see my other comment

torchao/__init__.py
manuelcandales: This is independent of MPS kernels. When you run `USE_CPP=1 pip install .`, that `build` folder is created, and it contains the .dylib.
I added this, to cover then case when importing torchao from the ao directory, in which case the local module is imported, instead of the one installed in the environment

.github/workflows/torchao_experimental_test.yml
manuelcandales: MPS test added to CI. I renamed the existing job to `test-cpu-ops`. Now we have two jobs under the TorchAO experimental workflow: `test-cpu-ops` and `test-mps-ops`

torchao/__init__.py
metascroy: I think this is a bit brittle, and we should only load the library from the installed package.  If you want to load the library from a local build folder, you can manually call torch.load_library.

Generally, I don't think we should cater to case where users are mixing the local repo with the installed package, other than maybe give a warning.  Note, you can pip install this without cloning the repo locally if you just want to use it:

```
USE_CPP=1 pip install git+https://github.com/pytorch/ao.git@main
```

What do you think?

setup.py
metascroy: I saw this flag, but it's tied to "torch.mps.is_available()" rather than also being tied to a user controlled environment variable.

Maybe it's OK

setup.py
```diff
@@ -174,15 +174,19 @@ def build_cmake(self, ext):
...
                 "-DTorch_DIR=" + torch_dir,
                 "-DCMAKE_LIBRARY_OUTPUT_DIRECTORY=" + extdir,
+                "-DCMAKE_INSTALL_PREFIX=cmake-out",
```
metascroy, line 189: Why is this necessary now?  Is it required for MPS?

.github/workflows/torchao_experimental_test.yml
```diff
@@ -56,3 +57,39 @@ jobs:
           sh build_and_run_tests.sh
           rm -rf /tmp/cmake-out
           popd
+
+  test-mps-ops:
```
kimishpatel, line 61: Enabling CI can be a separate step from enabling mps via pip install

setup.py
kimishpatel: also note that not all cpu kernels move out of experimental. Meaning building cpu kernels in experimental may continue until we decide some newly added ops/features are ready to move out of experimental

setup.py
kimishpatel: I would also rename these build flags as `TORCHAO_BUILD_EXPERIMENTAL_MPS`

torchao/__init__.py
kimishpatel: I agree that this is brittle. If you are locally doing `pip install` then you want the libs to be installed in `site-packages` regardless of where you are building the artifacts and thats where you want to load from

kimishpatel (changes requested):
Manuel, why do you wanna enable lowbit kernels via pip install quite yet. Is this in prep for the talk? Besides 4-bit where we still dont have performant gemm (only gemv), I would probably keep this out of pip install

manuelcandales:
The main value added by this PR, is the ability to easily install the MPS kernels locally with the same simple command used to install the CPU kernels: `USE_CPP=1 pip install .`

> I would probably keep this out of pip install
This is still a local pip install. It isn't distributed with pip. If you do `pip install torchao`, you won't get these kernels yet.

.github/workflows/torchao_experimental_test.yml
manuelcandales: @metascroy suggested I added it in an earlier comment, and I did

setup.py
manuelcandales: Yes, it is needed by the CMakeLists.txt used for the MPS ops

torchao/experimental/ops/mps/test/test_lowbit.py
```diff
@@ -10,17 +10,18 @@
...
-)
+import torchao  # noqa: F401
 
 try:
     for nbit in range(1, 8):
         getattr(torch.ops.torchao, f"_linear_fp_act_{nbit}bit_weight")
         getattr(torch.ops.torchao, f"_pack_weight_{nbit}bit")
 except AttributeError:
     try:
+        libname = "libtorchao_ops_mps_aten.dylib"
```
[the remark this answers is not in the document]
manuelcandales: The test doesn't **need** to manually load the library. I didn't add this line, I just moved it under the try/catch block
In line 13 we `import torchao`, which would load the library if torchao was installed using `USE_CPP=1 pip install .`
I could have deleted the code to load the dylib, but I decided not to, because if the user is doing dev work and building the library using the existing bash script, it is nice to still be able to run the test.

setup.py
manuelcandales: > I saw this flag, but it's tied to "torch.mps.is_available()" rather than also being tied to a user controlled environment variable.

The CPU ops aren't tied to a user controlled environment variable either. You always build them when running on Apple Silicon. Same for cuda. The cuda ops are always built if cuda and the CUDA toolkit are available. It isn't controlled by an environment variable. Why would MPS be any different?

We can add an environment variable to control whether we build experimental ops or not. That makes sense. But that should be a separate PR, because it would affect all of the experimental ops, including the CPU ones, not just MPS. But I don't think we should have an environment variable for MPS vs CPU.

torchao/__init__.py
manuelcandales: Ok, I agree with this. I think it sucks that we can't import them from the root `ao` directory. But I agree with this being brittle. Will remove.

setup.py
manuelcandales: @kimishpatel More reason to have a flag for all experimental ops, rather than an flag specific to MPS ops.

setup.py
metascroy: For CPU, I do plan to add some more fine grained control with environment variables, especially because KleidiAI increases the build time by a non-trivial amount.

Something like:

```
class CppBuildOptions:
    def __init__(self, default = "0"):        
        self.build_type = "Debug" if use_debug_mode() else "Release"
        self.build_cpu = platform.machine().startswith("arm64") and (os.getenv("TORCHAO_BUILD_CPU", self.default) == "1)
        self.build_kleidiai = platform.machine().startswith("arm64") and (os.getenv("TORCHAO_BUILD_KLEIDIAI", self.default) == "1")
        self.build_experimental_mps = (platform.system() == "Darwin") and torch.mps.is_available() and (os.getenv("TORCHAO_BUILD_EXPERIMENTAL_MPS", self.default) == "1")
```

We also plan to start moving CPU ops out of experimental soon, so that's a reason why they aren't exactly the same as MPS in terms of "always build" vs. not.

I think adding a user supplied TORCHAO_BUILD_EXPERIMENTAL_MPS might be a good medium for @kimishpatel's concern on why these are being built in setup.py at all.  They are still pip installable with:

```
USE_CPP=1 TORCHAO_BUILD_EXPERIMENTAL_MPS=1 pip install .
```
with all the benefits that brings with auto-loading the kernels at __init__.py, but their build is clearly controlled.

Note that torchchat can still enable these by default with that above install command if it chooses.

.github/workflows/torchao_experimental_test.yml
metascroy: The CI does test whether the setup.py and __init__.py are building/loading the kernels correctly, so I do think it makes sense.

setup.py
manuelcandales: Implemented!

setup.py
```diff
@@ -53,6 +53,11 @@ def read_version(file_path="version.txt"):
     and platform.system() == "Darwin"
 )
 
+build_torchao_experimental_mps = (
+    os.getenv("TORCHAO_BUILD_EXPERIMENTAL_MPS") == "1"
```
metascroy, line 57: shouldn't you also add the torch.mps.is_available thing here too?

metascroy (approved):

PR #1791, opened 2025-02-27
Add support for copy_ for plain layout and tensor core tiled layout

jerryzh168:
Summary:
att, only support copy_ from AQT to another AQT with same metadata (shapes etc.)

Tested int4wo, int8wo, int8dq

This is to support the `param.data.copy_` usage in vllm/sglang

Test Plan:
python test/dtypes/test_affine_quantized.py -k test_copy_

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/dtypes/test_affine_quantized.py
+++ b/test/dtypes/test_affine_quantized.py
@@ -214,7 +214,6 @@ class TestAffineQuantized(TestCase):
         "apply_quant", get_quantization_functions(False, True, "cuda", False)
     )
     def test_copy_(self, apply_quant):
-        print("apply_quant:", apply_quant)
         linear = torch.nn.Linear(128, 256, dtype=torch.bfloat16, device="cuda")
         ql = apply_quant(linear)
         linear2 = torch.nn.Linear(128, 256, dtype=torch.bfloat16, device="cuda")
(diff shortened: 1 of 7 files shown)


test/dtypes/test_affine_quantized.py
```diff
@@ -175,6 +175,23 @@ def test_print_quantized_module(self, apply_quant):
         ql = apply_quant(linear)
         assert "AffineQuantizedTensor" in str(ql)
 
+    @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
```
drisspg, line 178: Can you add some tests to ensure not same_metadata case errors

drisspg (approved):
Looks good just the one note on the test

jithunnair-amd:
@jerryzh168 This broke ROCm CI: https://hud.pytorch.org/pytorch/ao/commit/79e3366e273dcc50e0300384a1d0d6b1cc8d5e1f
I'm not sure if we plan to follow the same approach as we do in the PyTorch project of reverting and relanding PRs that break ROCm CI. But if we do, please revert this PR and add the `ciflow/rocm` label to kick off ROCm CI on this PR so you can ensure it's green before relanding.

cc @msaroufim @supriyar

PR #1793, opened 2025-02-28
[float8] add perf benchmarks for float8 training with rowwise + tensorwise scaling

danielvegamyhre:
**Summary**
- Add float8 training performance benchmarks for rowwise + tensorwise scaling.
- Add repro steps for these benchmarks.

### DIFF
--- a/torchao/float8/README.md
+++ b/torchao/float8/README.md
@@ -214,17 +214,20 @@ and tensorwise scaling. The training benchmarks were all run using:
 - Steps 100
 - `torch.compile`
 - FSDP2
+- pytorch version: `2.7.0a0+gitb98af95`
+- torchao version: `0.10.0+git890e0ac8`
+- torchtitan version: `0.0.2`
 
-| Model         | Scaling                           | Activation checkpointing | Peak Memory (GB)  | Median tokens/second | Speedup over baseline
-| ------------- | --------------------------------- | ------------------------ | ------------------| -------------------- | ---------------------
-| Llama3-8b     |  none (bf16)                      | per op SAC               | 47.65             |  6150                | -
-| Llama3-8b     |  tensorwise with optimal settings | per op SAC               | 47.77             |  7689.5              | 25.03%
-| Llama3-8b     |  rowwise                          | per op SAC               | 47.79             |  6768                | 10.05%
+
+| Model         | Scaling                            | Activation checkpointing | Peak Memory (GB)  | Median tokens/second | Speedup over baseline
+| ------------- | ---------------------------------- | ------------------------ | ------------------| -------------------- | ---------------------
+| Llama3-8b     |  none (bfloat16)                   | per op SAC               | 47.65             |  6150                | -
+| Llama3-8b     |  tensorwise with float8 all-gather | per op SAC               | 47.77             |  7689.5              | 25.03%
+| Llama3-8b     |  rowwise with bfloat16 all-gather  | per op SAC               | 47.79             |  6768                | 10.05%
 
 **Important notes**:
-- Speedups increase as M,K,N (GEMM dimensions) increase. Speedups as high as 1.5x have been measured with larger shapes ((example)[https://pytorch.org/blog/training-using-float8-fsdp2/]).
+- E2E speedups increase as M,K,N (GEMM dimensions) increase. Speedups as high as 1.5x have been measured with larger shapes ((example)[https://pytorch.org/blog/training-using-float8-fsdp2/]).
 - Rowwise scaling is better at handling outliers than tensorwise scaling, so these recipes are different points on the accuracy vs performance curve.
-- Tensorwise scaling benchmarks were ran with optimal settings, namely: `enable_fsdp_float8_all_gather`, `precompute_float8_dynamic_scale_for_fsdp`, `force_recompute_fp8_weight_in_bwd`.
 
 **Reproducing training benchmarks**
 To reproduce these benchmarks, you can follow these steps:
--- a/torchao/float8/README.md
+++ b/torchao/float8/README.md
@@ -237,7 +237,7 @@ including [downloading a tokenizer](https://github.com/pytorch/torchtitan?tab=re
 2. Install torchao following these [steps](https://github.com/pytorch/ao/tree/main?tab=readme-ov-file#installation).
 3. From the `torchao/float8/benchmarking/` directory, you can run the following commands to reproduce the benchmarks above:
    - bf16 + compile: `TORCHTITAN_ROOT=<path> ./float8_training_benchmark.sh`
-   - float8 tensorwise: `TORCHTITAN_ROOT=<path> FLOAT8_RECIPE_WITH_BEST_SETTINGS="tensorwise" ./float8_training_benchmark.sh`
-   - float8 rowwise: `TORCHTITAN_ROOT=<path> FLOAT8_RECIPE_WITH_BEST_SETTINGS="rowwise" ./float8_training_benchmark.sh`
+   - float8 tensorwise with float8 all-gather + compile: `TORCHTITAN_ROOT=<path> FLOAT8_RECIPE_WITH_BEST_SETTINGS="tensorwise" ./float8_training_benchmark.sh`
+   - float8 rowwise with bf16 all-gather + compile: `TORCHTITAN_ROOT=<path> FLOAT8_RECIPE_WITH_BEST_SETTINGS="rowwise" ./float8_training_benchmark.sh`
 
 See the float8 training benchmarking [guide](.torchao/float8/benchmarking/README.md) for more details.
--- a/torchao/float8/README.md
+++ b/torchao/float8/README.md
@@ -215,15 +215,15 @@ and tensorwise scaling. The training benchmarks were all run using:
 - `torch.compile`
 - FSDP2
 
-| Model         | Scaling      | Activation checkpointing | Median tokens/second      | Peak Memory (GB) |
-| ------------- | ------------ | ------------------------ | ------------------------- | ---------------- |
-| Llama3-8b     |  none (bf16) | per op SAC               | 6019                      | 47.65            |
-| Llama3-8b     |  tensorwise  | per op SAC               | 7190                      | 47.77            |
-| Llama3-8b     |  rowwise     | per op SAC               | 6649                      | 47.79            |
+| Model         | Scaling      | Activation checkpointing | Peak Memory (GB)  | Median tokens/second | Speedup over basline
+| ------------- | ------------ | ------------------------ | ------------------| -------------------- | ---------------------
+| Llama3-8b     |  none (bf16) | per op SAC               | 47.65             |  6019                | -
+| Llama3-8b     |  tensorwise  | per op SAC               | 47.77             |  7190                | 19.45%
+| Llama3-8b     |  rowwise     | per op SAC               | 47.79             |  6649                | 10.47%
 
-In these benchmarks tensorwise scaling achieved ~8% higher tokens/second over rowwise scaling, and ~19.5% higher than the bf16 baseline.
-However, it is important to note that rowwise scaling has been shown to yield improvments in training loss/accuracy due to reduced quantization error, particularly
-when training large models for many steps.
+**Important notes**:
+- Speedups increase as M,K,N (GEMM dimensions) increase. Speedups as high as 1.5x have been measured with larger shapes ((example)[https://pytorch.org/blog/training-using-float8-fsdp2/]).
+- Rowwise scaling is better at handling outliers than tensorwise scaling, so these recipes are different points on the accuracy vs performance curve.
 
 **Reproducing training benchmarks**
 To reproduce these benchmarks, you can follow these steps:
--- a/torchao/float8/README.md
+++ b/torchao/float8/README.md
@@ -215,7 +215,7 @@ and tensorwise scaling. The training benchmarks were all run using:
 - `torch.compile`
 - FSDP2
 
-| Model         | Scaling      | Activation checkpointing | Peak Memory (GB)  | Median tokens/second | Speedup over basline
+| Model         | Scaling      | Activation checkpointing | Peak Memory (GB)  | Median tokens/second | Speedup over baseline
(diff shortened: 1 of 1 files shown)


torchao/float8/README.md
```diff
@@ -269,3 +269,37 @@ python test/float8/test_fsdp2/test_fsdp2.py
...
+
+- Single-node training on 8xH100 GPUs
+- Batch size 2
+- Sequence length 8192
+- Steps 100
+- `torch.compile`
+- FSDP2
+
+| Model         | Scaling     | Activation checkpointing | Average tokens/second     | Peak Memory (GB) |
+| ------------- | ----------- | ------------------------ | ------------------------- | ---------------- |
```
vkuzo, line 286: we should include the baseline (bf16 + compile) here so it's clear what the speedup is from baseline

torchao/float8/README.md
```diff
@@ -269,3 +269,37 @@ python test/float8/test_fsdp2/test_fsdp2.py
...
+| Llama3-8b     |  tensorwise | per op SAC               | 3217.4                    | 75.47            |
```
vkuzo, line 287: The TPS number looks a bit low, I was expecting something in the 7k range.  What machine are you using?  I'm currently on `devgpu006.vll6` and I've noticed in the past couple of weeks torchtitan performance has been about 0.5x of what it was ~weeks ago, so it's either something local to my machine or a regression somewhere in the stack.

torchao/float8/README.md
```diff
@@ -269,3 +269,37 @@ python test/float8/test_fsdp2/test_fsdp2.py
...
+To reproduce these benchmarks, you can follow these steps:
+
+1. On a machine with 8 H100 GPUs, clone torchtitan and follow local installation [steps](https://github.com/pytorch/torchtitan?tab=readme-ov-file#installation),
+including [downloading a tokenizer](https://github.com/pytorch/torchtitan?tab=readme-ov-file#downloading-a-tokenizer).
+2. Install torchao following these [steps](https://github.com/pytorch/ao/tree/main?tab=readme-ov-file#installation).
+2. From the torchtitan root directory, you can run the following command to reproduce the benchmarks. Note
+    - Run float8 training with tensorwise scaling: `NGPU=8 CONFIG_FILE="./train_configs/llama3_8b.toml" ./run_llama_train.sh --training.steps=100 --training.batch_size=2 --training.compile --model.converters="float8"`
+    - Run float8 training with rowwise scaling: `NGPU=8 CONFIG_FILE="./train_configs/llama3_8b.toml" ./run_llama_train.sh --training.steps=100 --training.batch_size=2 --training.compile --model.converters="float8" --float8.recipe_name="rowwise"`
+
+Benchmark results were calculated by averaging the tokens/second over the first 100 training steps, excluding step 1 which includes
```
vkuzo, line 304: nit: how about median instead of average?

torchao/float8/README.md
```diff
@@ -269,3 +269,37 @@ python test/float8/test_fsdp2/test_fsdp2.py
...
+that rowwise scaling has been shown to yield improvments in training loss/accuracy due to reduced quantization error, particularly
+when training large models for many steps.
+
+**Reproducing training benchmarks**
+To reproduce these benchmarks, you can follow these steps:
+
+1. On a machine with 8 H100 GPUs, clone torchtitan and follow local installation [steps](https://github.com/pytorch/torchtitan?tab=readme-ov-file#installation),
+including [downloading a tokenizer](https://github.com/pytorch/torchtitan?tab=readme-ov-file#downloading-a-tokenizer).
+2. Install torchao following these [steps](https://github.com/pytorch/ao/tree/main?tab=readme-ov-file#installation).
+2. From the torchtitan root directory, you can run the following command to reproduce the benchmarks. Note
```
vkuzo, line 300: this is great!  IMO we should mention that the syntax of how to reproduce these from torchtitan may change (from torchtitan, if they change their arguments), and include the torch, torchao, torchtitan versions used to generate these.

torchao/float8/README.md
danielvegamyhre: > What machine are you using? I'm currently on `devgpu006.vll6`

I'm on this machine as well, I'll try using a new one.

> hmm, or this may be related to batch_size 2 I see on L279. Was this changed from torchtitan defaults?

default batch size is 1, I actually increased it to 2 to get the TPS as high as we could within the memory capacity constraint of an H100. I also was thinking batch size of 1 is probably not super common for production training runs so was trying to make it more representative of real workloads.

torchao/float8/README.md
danielvegamyhre: Done

torchao/float8/README.md
danielvegamyhre: I created a benchmarking script to automate this and added a short guide for it: https://github.com/pytorch/ao/pull/1802

updated this PR to use it

torchao/float8/README.md
danielvegamyhre: > I'm currently on devgpu006.vll6 and I've noticed in the past couple of weeks torchtitan performance has been about 0.5x of what it was ~weeks ago

Good catch, I switched machines and perf was much better. Updated PR accordingly

torchao/float8/README.md
```diff
@@ -269,3 +269,38 @@ python test/float8/test_fsdp2/test_fsdp2.py
...
+- `torch.compile`
+- FSDP2
+
+| Model         | Scaling      | Activation checkpointing | Median tokens/second      | Peak Memory (GB) |
+| ------------- | ------------ | ------------------------ | ------------------------- | ---------------- |
+| Llama3-8b     |  none (bf16) | per op SAC               | 6019                      | 47.65            |
+| Llama3-8b     |  tensorwise  | per op SAC               | 7190                      | 47.77            |
+| Llama3-8b     |  rowwise     | per op SAC               | 6649                      | 47.79            |
+
+In these benchmarks tensorwise scaling achieved ~8% higher tokens/second over rowwise scaling, and ~19.5% higher than the bf16 baseline.
```
vkuzo, line 291: nit: how about
* add column with "speedup over baseline" instead of only explaining it in a sentence
* saying something like "rowwise scaling is better at handling outliers than tensorwise scaling, so these recipes are different points on the accuracy vs performance curve"
* saying that speedups increase as M,K,N increase, and pointing to blogs such as https://pytorch.org/blog/training-using-float8-fsdp2/ where e2e speedups as high as 1.5x are quoted.  This is just to clarify that the 1.2x shown here is not the max speedup - it's just the speedup given the benchmark setup.

danielvegamyhre:
Looks like test failure is related to https://github.com/pytorch/ao/issues/1799

torchao/float8/README.md
danielvegamyhre: done, let me know what you think

vkuzo:
Just to confirm, for tensorwise scaling, I see that https://github.com/pytorch/ao/blob/main/benchmarks/float8/training/float8_training_benchmark.sh is using recipe-lookup-by name.

Unfortunately, for tensorwise scaling, this is correct but not optimal.  We also need to enable the following flags to enable float8 all-gather with FSDP2, and those flags are currently not supported when using titan's recipe-string-to-recipe lookup:

```
# source: https://github.com/pytorch/torchtitan/blob/main/docs/float8.md
CONFIG_FILE="./train_configs/llama3_8b.toml" ./run_train.sh --model.converters="float8" --float8.enable_fsdp_float8_all_gather --float8.precompute_float8_dynamic_scale_for_fsdp --float8.force_recompute_fp8_weight_in_bwd --training.compile
```

We have https://github.com/pytorch/torchtitan/issues/901 to track making the torchtitan side of this better.  Any chance we  can update the tensorwise benchmark to include these flags, and also call out in the table that the tensorwise recipe has float8 all-gather for FSDP?

torchao/float8/README.md
```diff
@@ -202,3 +202,38 @@ python test/float8/test_fsdp2/test_fsdp2.py
...
+and tensorwise scaling. The training benchmarks were all run using:
+
+- Single-node training on 8xH100 GPUs
+- Batch size 1
+- Sequence length 8192
+- Steps 100
+- `torch.compile`
+- FSDP2
+
+| Model         | Scaling      | Activation checkpointing | Peak Memory (GB)  | Median tokens/second | Speedup over basline
```
vkuzo, line 218: nit: `baseline` (typo)

danielvegamyhre:
> Just to confirm, for tensorwise scaling, I see that https://github.com/pytorch/ao/blob/main/benchmarks/float8/training/float8_training_benchmark.sh is using recipe-lookup-by name.
> 
> Unfortunately, for tensorwise scaling, this is correct but not optimal. We also need to enable the following flags to enable float8 all-gather with FSDP2, and those flags are currently not supported when using titan's recipe-string-to-recipe lookup:
> 
> ```
> # source: https://github.com/pytorch/torchtitan/blob/main/docs/float8.md
> CONFIG_FILE="./train_configs/llama3_8b.toml" ./run_train.sh --model.converters="float8" --float8.enable_fsdp_float8_all_gather --float8.precompute_float8_dynamic_scale_for_fsdp --float8.force_recompute_fp8_weight_in_bwd --training.compile
> ```
> 
> We have [pytorch/torchtitan#901](https://github.com/pytorch/torchtitan/issues/901) to track making the torchtitan side of this better. Any chance we can update the tensorwise benchmark to include these flags, and also call out in the table that the tensorwise recipe has float8 all-gather for FSDP?

Sure, but I thought enabling float8 all gather would make it less of a 1:1 comparison with rowwise? Or is the goal here just to showcase the peak achievable speedup using all optimal configs for each scaling strategy?

vkuzo:
> Or is the goal here just to showcase the peak achievable speedup using all optimal configs for each scaling strategy?

Yes, IMO that's what this should do.  We should call out any features which are not implemented (or impossible to implement), but at the end we want the best speedup with each recipe, with each appropriate knobs turned on.

danielvegamyhre:
> > Or is the goal here just to showcase the peak achievable speedup using all optimal configs for each scaling strategy?
> 
> Yes, IMO that's what this should do. We should call out any features which are not implemented (or impossible to implement), but at the end we want the best speedup with each recipe, with each appropriate knobs turned on.

@vkuzo I managed to find a machine without perf regression and reran all benchmarks, using the optimal configs for tensorwise scaling this time, so this is ready for another look when you have a sec

torchao/float8/README.md
```diff
@@ -202,3 +202,39 @@ python test/float8/test_fsdp2/test_fsdp2.py
...
+- Batch size 1
+- Sequence length 8192
+- Steps 100
+- `torch.compile`
+- FSDP2
+
+| Model         | Scaling                           | Activation checkpointing | Peak Memory (GB)  | Median tokens/second | Speedup over baseline
+| ------------- | --------------------------------- | ------------------------ | ------------------| -------------------- | ---------------------
+| Llama3-8b     |  none (bf16)                      | per op SAC               | 47.65             |  6150                | -
+| Llama3-8b     |  tensorwise with optimal settings | per op SAC               | 47.77             |  7689.5              | 25.03%
```
vkuzo, line 221: how about something like `tensorwise with float8 all-gather` and `rowwise with float8 all-gather`? "optimal settings" should be true for all the rows in this table, it's just that the actual settings change.

torchao/float8/README.md
```diff
@@ -202,3 +202,39 @@ python test/float8/test_fsdp2/test_fsdp2.py
...
+- Single-node training on 8xH100 GPUs
+- Batch size 1
+- Sequence length 8192
+- Steps 100
+- `torch.compile`
+- FSDP2
+
+| Model         | Scaling                           | Activation checkpointing | Peak Memory (GB)  | Median tokens/second | Speedup over baseline
+| ------------- | --------------------------------- | ------------------------ | ------------------| -------------------- | ---------------------
+| Llama3-8b     |  none (bf16)                      | per op SAC               | 47.65             |  6150                | -
```
vkuzo, line 220: nit: `bfloat16`, to match how we spell dtypes in the rest of PyTorch?

torchao/float8/README.md
```diff
@@ -202,3 +202,39 @@ python test/float8/test_fsdp2/test_fsdp2.py
 # make sure to turn on torch.compile to get the best performance
 ./benchmarks/float8/bench_linear_float8.py -o ../tmp/test.txt --compile
 ```
+
+### Training benchmarks
+
+[Torchtitan](https://github.com/pytorch/torchtitan) was used to benchmark float8 training performance, for both rowwise
+and tensorwise scaling. The training benchmarks were all run using:
+
+- Single-node training on 8xH100 GPUs
```
vkuzo, line 211: can we add PyTorch version, torchtitan version, torchao version?  Ideally the script could display them.

torchao/float8/README.md
```diff
@@ -202,3 +202,39 @@ python test/float8/test_fsdp2/test_fsdp2.py
...
+- FSDP2
+
+| Model         | Scaling                           | Activation checkpointing | Peak Memory (GB)  | Median tokens/second | Speedup over baseline
+| ------------- | --------------------------------- | ------------------------ | ------------------| -------------------- | ---------------------
+| Llama3-8b     |  none (bf16)                      | per op SAC               | 47.65             |  6150                | -
+| Llama3-8b     |  tensorwise with optimal settings | per op SAC               | 47.77             |  7689.5              | 25.03%
+| Llama3-8b     |  rowwise                          | per op SAC               | 47.79             |  6768                | 10.05%
+
+**Important notes**:
+- Speedups increase as M,K,N (GEMM dimensions) increase. Speedups as high as 1.5x have been measured with larger shapes ((example)[https://pytorch.org/blog/training-using-float8-fsdp2/]).
```
vkuzo, line 225: nit: "E2e speedups as high as 1.5x..."

vkuzo (approved):
looks great, thank you!

torchao/float8/README.md
danielvegamyhre: I thought about this but the thing is, tensorwise has 3 settings enabled, so listing all 3 would cause the column to become huge and make the table formatting clunky.

vkuzo: I think it's important to specify that float8 all-gather is used for tensorwise, and not important to list out the three specific settings used to enable that feature.

PR #1801, opened 2025-02-28
bugfix clean_release_notes.py

HDCharles:
1) Deprecation(s) name needs to be consistent or else it wont find that dict entry

2) need to handle escape char of regex


andrewor14:
Good catch, thanks for fixing!

PR #1808, opened 2025-03-01
Silence loud error on torchao cpu builds

msaroufim:
before this fix, just importing torchao in a cpu only environment would print `import error: triton not found`

Came up live in this talk by @manuelcandales https://www.youtube.com/watch?v=0vv9K0RvOfo&ab_channel=GPUMODE

To repro on your M1 laptop

```
USE_CPP=0 pip install .
python
import torchao
```


msaroufim:
Errors seem like flakes

drisspg (approved):
`

PR #1810, opened 2025-03-03
Support MXFP6 packing and fused unpack-dequantise kernel (conflicts resolved)

alex-titterton:
Updated version of #1687 to resolve some merge conflicts.

Good afternoon! Following recent developments and increased support for MXFP formats, it would be useful to support efficient packing for MXFP6 to benefit from the decrease in memory consumption and bandwidth requirements vs (MX)FP8.

MXFP6 has shown to perform similarly well compared with MXFP8 in LLM inference tasks, and with sufficient QAT even as well as float32, e.g. in the [MXFP paper](https://arxiv.org/abs/2310.10537).

This PR packs the bits representing the FP6 values in a 4+2 fashion as is done in the [FP6 LLM paper](https://arxiv.org/abs/2401.14112), and supports both E2M3 and E3M2 variants. Packing is done via standalone Triton kernel, with unpacking and dequantisation performed via a fused kernel for better performance.

Tests have been added in test_custom_cast.py and test_mx_tensor.py to cover accuracy in quantise-pack-unpack-dequantise with various FP6 values (min/max norm, min/max subnorm, -0.0 etc for both E2M3 and E3M2 variants) as well as checking packed tensor dimensions.

Note: due to the 4+2 packing scheme this requires the packing dimension to be a multiple of 4 since the packed dimension will be 3/4 of this. However the typical MX block size is 32 (—> 24 when packed), and HW implementations tend to require dims to be multiples of 16 or 32, so this should not be a problem. The relevant test case dimensions have been changed from 6 to 8 and the MX block sizes from 2 to 4 where applicable in order to accommodate this requirement.

Note: I've added a bool flag into `config.py` to enable/disable FP6 packing, but it doesn't belong to any class as such. I wasn't sure where best to put it following the restructuring of the config file, and so for now it's just accessed from other functions/classes as `config.pack_fp6`.

### DIFF
--- a/torchao/prototype/mx_formats/config.py
+++ b/torchao/prototype/mx_formats/config.py
@@ -15,6 +15,7 @@ from torchao.prototype.mx_formats.constants import (
     SUPPORTED_ELEM_DTYPES,
 )
 
+pack_fp6 = True
 
 class MXGemmKernelChoice(Enum):
     # always available - MX operands are dequantized and a high precision
--- a/torchao/prototype/mx_formats/config.py
+++ b/torchao/prototype/mx_formats/config.py
@@ -61,8 +61,9 @@ class MXLinearConfig:
     # If True, uses a custom triton kernel for fp4 dequantize
     use_fp4_custom_triton_dequant_kernel: bool = False
 
-    # If True, packs 4xFP6 into 3xuint8 containers, using custom triton kernels (fused unpack/dequantize)
-    pack_fp6 = False
+    # If True, packs 4xFP6 into 3xuint8 containers for inference, using custom triton
+    # kernels (fused unpack/dequantize). Training not currently supported.
+    pack_fp6 = True
 
     def __post_init__(self):
         # validate elem_dtype and its overrides
--- a/torchao/prototype/mx_formats/config.py
+++ b/torchao/prototype/mx_formats/config.py
@@ -63,7 +63,7 @@ class MXLinearConfig:
 
     # If True, packs 4xFP6 into 3xuint8 containers for inference, using custom triton
     # kernels (fused unpack/dequantize). Training not currently supported.
-    pack_fp6 = True if hasattr(torch.library, 'custom_op') else False
+    pack_fp6 = True if hasattr(torch.library, "custom_op") else False
 
     def __post_init__(self):
         # validate elem_dtype and its overrides
--- a/torchao/prototype/mx_formats/config.py
+++ b/torchao/prototype/mx_formats/config.py
@@ -63,7 +63,7 @@ class MXLinearConfig:
 
     # If True, packs 4xFP6 into 3xuint8 containers for inference, using custom triton
     # kernels (fused unpack/dequantize). Training not currently supported.
-    pack_fp6 = True
+    pack_fp6 = True if hasattr(torch.library, 'custom_op') else False
 
     def __post_init__(self):
         # validate elem_dtype and its overrides
--- a/torchao/prototype/mx_formats/config.py
+++ b/torchao/prototype/mx_formats/config.py
@@ -15,7 +15,6 @@ from torchao.prototype.mx_formats.constants import (
     SUPPORTED_ELEM_DTYPES,
 )
 
-pack_fp6 = True
 
 class MXGemmKernelChoice(Enum):
     # always available - MX operands are dequantized and a high precision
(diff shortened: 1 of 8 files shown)


torchao/prototype/mx_formats/config.py
```diff
@@ -15,6 +15,7 @@
     SUPPORTED_ELEM_DTYPES,
 )
 
+pack_fp6 = True
```
vkuzo, line 18: can we put this inside of `MXLinearConfig` and default to `False`?

alex-titterton: Makes sense -- I've updated this to use the `MXLinearConfig` class and defaulted to `False`.

vkuzo:
Thank you! I took a closer look at the PR after the rebase, and just have a couple of follow-up questions:
1. Is this inference only, or do you expect usage during training?  If inference only it may be simpler to just omit this from training code.
2. would it be possible to add a reference plain PyTorch kernel for the packing, similar to how our packed fp4x2 currently works?  That can be in a separate PR is needed, but at least we should have a TODO comment tracking it.

alex-titterton:
> 2. would it be possible to add a reference plain PyTorch kernel for the packing, similar to how our packed fp4x2 currently works?  That can be in a separate PR is needed, but at least we should have a TODO comment tracking it.

I've added a reference implementation just now 👍

vkuzo:
> I'd say the use case for inference is of course clearer, but for training I could imagine it being a benefit perhaps more in the case of fine-tuning on hardware with limited memory capacity/bandwidth (i.e. when the dimensions of the model have been nailed down such that the dimensions shouldn't pose a problem for packing). On that basis I think it's perhaps worth having as an option where the default is not to pack, but it can be used for either training or inference?

Makes sense.  How about adding it to inference on-by-default for fp6, and we revisit training when the use case is more clear?

alex-titterton:
> > I'd say the use case for inference is of course clearer, but for training I could imagine it being a benefit perhaps more in the case of fine-tuning on hardware with limited memory capacity/bandwidth (i.e. when the dimensions of the model have been nailed down such that the dimensions shouldn't pose a problem for packing). On that basis I think it's perhaps worth having as an option where the default is not to pack, but it can be used for either training or inference?
> 
> Makes sense. How about adding it to inference on-by-default for fp6, and we revisit training when the use case is more clear?

Sorry, just to clarify: do you mean to leave the training support parts in but off-by-default, or remove entirely?

vkuzo:
hmm, thinking out loud here:
* packing weights to fp6 seems clearly useful for inference: we save model size on disk, and potentially throughput if the workload is memory bound
* packing weights/activations/grads to fp6 value to training is not yet clear

that context would point me to say "on by default in inference" and "not supported at all in training to minimize complexity, unless someone articulates the actual benefit" - thoughts?

alex-titterton:
> hmm, thinking out loud here:
> 
> * packing weights to fp6 seems clearly useful for inference: we save model size on disk, and potentially throughput if the workload is memory bound
> * packing weights/activations/grads to fp6 value to training is not yet clear
> 
> that context would point me to say "on by default in inference" and "not supported at all in training to minimize complexity, unless someone articulates the actual benefit" - thoughts?

That sounds reasonable. The same `MXLinearConfig` class is used for both training and inference, but I've tidied it up such that there's no `pack_fp6` in any of the training wrappers etc.

vkuzo:
thank you, looks great!  I have just kicked off a CI run

> The same MXLinearConfig class is used for both training and inference,

yep, we will likely improve this soon as we iterate on the MX flows.  Having it in the same place is temporary.

vkuzo:
would you mind running the linter and amending the PR?

```
cd torchao
ruff format
ruff check . --fix
```

vkuzo:
Also, looks like CI is red, seems like the new functionality and tests are not being properly skipped when triton is not available.  I think copying how the fp4 triton packing kernel handles this would be one way to resolve.

alex-titterton:
> Also, looks like CI is red, seems like the new functionality and tests are not being properly skipped when triton is not available. I think copying how the fp4 triton packing kernel handles this would be one way to resolve.

Ah, I see it - one of the dummy function definitions still had `MX_BLOCK_SIZE: tl.constexpr` in the case where triton is unavailable. Fixed it now, sorry about that!

alex-titterton:
Looks like the tests are failing for `torch==2.3` since the `torch.library.custom_op` module (used in `custom_cast.py` to allow for `torch.compile` with fp6 packing) only exists in `torch>=2.4`. A lot of tests are skipped when `torch<2.4`, but this is done after the imports are run and so the error will still occur.

One option is if I create a dummy no-op for `pack_uint6` in the case where `torch<2.4`, and in that case force the `pack_fp6` flag to `False`? It's a little fiddly, but it would allow `torch==2.3` to still be supported just without fp6 packing.

alex-titterton:
I've added a proposed solution whereby `@torch.library.custom_op` decorated function declarations are only called if `torch >= 2.4`. If using `torch < 2.4` FP6 packing is disabled by default, and trying to use it will throw a more informative error message. @vkuzo hopefully this looks ok to you!

vkuzo:
thanks! I have kicked off another round of CI.

PR #1812, opened 2025-03-03
fix USE_CPP=0 during local build

vkuzo:
Summary:

Looks like this env var no longer matches its original intent (Python-only build)
as various other c++ extensions were added, fixing.

Test Plan:

```
with-proxy USE_CPP=0 pip install -e .
```

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,7 +47,7 @@ use_cpp = os.getenv("USE_CPP")
 
 import platform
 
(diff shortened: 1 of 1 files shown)


vkuzo:
* __->__ #1812

vkuzo:
CI is failing, needs some work

PR #1822, opened 2025-03-04
Remove libOMP linking for experimental kernels

Jack-Khuu:
As mentioned in https://github.com/pytorch/torchchat/pull/1493, we no longer need to explicitly link to OMP libs


Jack-Khuu:
@metascroy @manuelcandales Is there CI that auto checks for this? 

How should we test this?

metascroy:
It looks like it fails with linking error if you don't make it explicit.

It might have to do with the version of pytorch being used in the test.  Do you know which version has OMP bundled?

Jack-Khuu:
I think you're right, it was updated circa https://github.com/pytorch/pytorch/pull/145870

I'll bump the pinned PT

jcaip (approved):
lgtm, feel free to merge once CI passes

Jack-Khuu:
Remade PR: https://github.com/pytorch/ao/pull/1828

This one got flooded on a bad rebase

PR #1826, opened 2025-03-04
Optionally enable KleidiAI + clean up setup.py flags

metascroy:
This PR,

* Guards universal kernels behind TORCHAO_BUILD_CPU_AARCH64
* Cleans up CMakeExtension to include cmake flags
* Introduces BuildOptions that control what CPU/MPS kernels are built.  Universal kernels are built if TORCHAO_BUILD_CPU_AARCH64 is specified,  KleidiAI kernels are built if TORCHAO_BUILD_KLEIDIAI is specified, and MPS kernels are built if TORCHAO_BUILD_EXPERIMENTAL_MPS is specified.  On Arm-based Macs, TORCHAO_BUILD_CPU_AARCH64 is enabled by default when USE_CPP=1.


setup.py
```diff
@@ -75,19 +76,53 @@ def use_debug_mode():
...
+        if self.build_kleidi_ai:
+            assert (
+                self.build_cpu_aarch64
+            ), "TORCHAO_BUILD_KLEIDIAI requires TORCHAO_BUILD_CPU_AARCH64 be set"
+
+        # TORCHAO_BUILD_EXPERIMENTAL_MPS is disabled by default.
+        self.build_experimental_mps = self._os_bool_var(
+            "TORCHAO_BUILD_EXPERIMENTAL_MPS", default=False
+        )
+        if self.build_experimental_mps:
```
manuelcandales, line 110: also need to assert _is_arm64 here. the experimental mps ops are intended for apple silicon

metascroy: Added arm64

PR #1829, opened 2025-03-04
float8 matmul bench: make peak tops lookup dynamic

vkuzo:
Summary:

Instead of hardcoding H100 numbers, makes the script choose between H100
and B200 based on available GPU.

Test Plan:

```
python benchmarks/float8/bench_matmul.py
```

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo:
* (to be filled)

drisspg (approved):

PR #1832, opened 2025-03-04
Update float8 training benchmark script tensorwise configs

danielvegamyhre:
### Summary of changes
- Update float8 training benchmarking script to use the following additional configurations, in order to capture more optimal perf numbers:
    - `--float8.enable_fsdp_float8_all_gather`
    - `--float8.precompute_float8_dynamic_scale_for_fsdp` 
    - `--float8.force_recompute_fp8_weight_in_bwd`
- Move float8 training benchmark readme into the `/benchmarks/float8/training` folder (it was incorrectly placed in the parent directory)

### Test plan
- `TORCHTITAN_ROOT=${HOME}/local/torchtitan FLOAT8_RECIPE="tensorwise" ./float8_training_benchmark.sh` uses correct float8 args


benchmarks/float8/training/float8_training_benchmark.sh
```diff
@@ -22,7 +22,11 @@ fi
 
 # validate recipe name
 if [ -n "${FLOAT8_RECIPE}" ]; then
-  FLOAT8_ARGS="--model.converters="float8" --float8.recipe_name=${FLOAT8_RECIPE}"
+  if [ "${FLOAT8_RECIPE}" == "tensorwise" ]; then
```
vkuzo, line 25: makes sense, but maybe we shouldn't call it recipe here, since it's not really the "recipe" value from our recipe name to lookup function?  how about `TORCHTITAN_BEST_SETTINGS_RECIPE` or something like that?

It's very annoying :(

vkuzo (approved):

benchmarks/float8/training/float8_training_benchmark.sh
danielvegamyhre: Hmm, we have "tensorwise" as a valid recipe name in the torchao config lookup function [here](https://github.com/pytorch/ao/blob/883dc6513bafc60195d9c12b9160afe2ee2e1dae/torchao/float8/config.py#L304). Are your referring to this [part](https://github.com/pytorch/torchtitan/blob/9a2f5784c3812125df283f423ec6e141359f1fc5/torchtitan/components/float8.py#L63) of torchtitan which basically uses `None` recipe as tensorwise instead of the string?

danielvegamyhre: Makes sense, changed it to `FLOAT8_RECIPE_WITH_BEST_SETTINGS` to communicate to the user we are using additional settings which are optimal for the chosen recipe.

danielvegamyhre:
CI failures on macos are unrelated to this change, merging

PR #1836, opened 2025-03-05
Migrate to config for Int8DynamicActivationIntxWeightConfig

metascroy:
This PR:

* Migrates to Int8DynamicActivationIntxWeightConfig
* Merges PackedLinearInt8DynamicActivationIntxWeightLayout  to use the same quantizer, and merges the tests


torchao/dtypes/affine_quantized_tensor.py
metascroy, line 1: @drisspg @jerryzh168 are we ok adding tensor_impl_ctr_kwargs to from_hp_to_intx.

It can be used to propagate a bias when constructing the weight tensor subclass via from_plain.

torchao/experimental/tests/test_int8_dynamic_activation_intx_weight.py
```diff
@@ -54,37 +45,111 @@ def test_accuracy(self):
...
+                weight_dtype=weight_dtype,
+                granularity=granularity,
+                has_weight_zeros=has_weight_zeros,
+                layout=reference_layout,
+            ),
+        )
+
+        with torch.no_grad():
+            result = quantized_model(activations)
+            expected_result = quantized_model_reference(activations)
```
Jack-Khuu, line 139: nit: We can factor out the creation of expected_results since it's just PlainLayout in both cases (different models)

torchao/experimental/quant_api.py
```diff
@@ -530,39 +526,46 @@ def int8_dynamic_activation_intx_weight(
...
+                when PackedLinearInt8DynamicActivationIntxWeightLayout is unavailable.
     """
 
-    def is_torchao_op_skippable(layout):
-        return isinstance(layout, PlainLayout) or (
-            isinstance(layout, PackedLinearInt8DynamicActivationIntxWeightLayout)
-            and layout.target == Target.ATEN
-        )
+    weight_dtype: torch.dtype = torch.int4
+    granularity: Union[PerRow, PerGroup] = PerRow()
```
Jack-Khuu, line 538: Why not 
`    granularity: Union[PerRow, PerGroup] = PerGroup(128),`

like `int8_dynamic_activation_intx_weight`?

torchao/experimental/quant_api.py
```diff
@@ -530,39 +526,46 @@ def int8_dynamic_activation_intx_weight(
...
-            torch.ops.torchao._pack_8bit_act_4bit_weight
-        except AttributeError:
-            raise Exception(
-                "TorchAO experimental kernels are not loaded.  To install the kernels, run `USE_CPP=1 pip install .` from ao on a machine with an ARM CPU."
-                + " You can also set target to 'aten' if you are using ARM CPU."
-                + "  Alternatively, use layout=PlainLayout() with int8_dynamic_activation_intx_weight, but note that doing so will result in much slower performance."
-            )
+
+@register_quantize_module_handler(Int8DynamicActivationIntxWeightConfig)
+def _int8_dynamic_activation_intx_weigh_transform(
```
Jack-Khuu, line 550:
```suggestion
def _int8_dynamic_activation_intx_weight_transform(
```

torchao/experimental/quant_api.py
```diff
@@ -571,107 +574,111 @@ def is_torchao_op_skippable(layout):
...
+        group_size = granularity.group_size
+    elif isinstance(granularity, PerRow):
+        group_size = weight.shape[-1]
+    else:
+        raise ValueError(f"granularity must be PerGroup or PerRow, got {granularity}")
+
+    tensor_impl_ctr_kwargs = None
+    if isinstance(layout, PackedLinearInt8DynamicActivationIntxWeightLayout):
+        # We need to create a new layout object for each module because when
+        # granulairty is PerRow, the layout objects cannot share the group_size
```
Jack-Khuu, line 588:
```suggestion
        # granularity is PerRow, the layout objects cannot share the group_size
```

torchao/experimental/packed_linear_int8_dynamic_activation_intx_weight_layout.py
```diff
@@ -312,113 +314,35 @@ def _impl_2d_aten(input_tensor, weight_tensor):
 
     target = weight_tensor.tensor_impl.get_layout().target
 
+    if weight_tensor.tensor_impl.get_layout().has_bias:
+        assert (
+            bias is None
+        ), "bias should be None because it is already packed with the weights (has_bias=True)"
```
Jack-Khuu, line 320: nit: `if: assert`; also fine with leaving it as-is for legibility

```suggestion
      assert (
          not weight_tensor.tensor_impl.get_layout().has_bias or bias is None
      ), "bias should be None because it is already packed with the weights (has_bias=True)"
```

Jack-Khuu (approved):
Mostly nits

Not terribly familiar with this code, but passes the gut test

torchao/experimental/quant_api.py
metascroy: PerRow is safer default because it doesn't depend on input data size.  I expect users should always specify this parameter

torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -276,7 +277,11 @@ def from_hp_to_intx(
 
         data = _layout.post_process(data)
         tensor_impl_ctr = get_tensor_impl_constructor(type(_layout))
-        tensor_impl = tensor_impl_ctr(data, scale, zero_point, _layout)
+        if tensor_impl_ctr_kwargs is None:
+            tensor_impl_ctr_kwargs = {}
+        tensor_impl = tensor_impl_ctr(
+            data, scale, zero_point, _layout, **tensor_impl_ctr_kwargs
+        )
```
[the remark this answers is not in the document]
metascroy: I'd like to hear from @drisspg  or someone from torchao on this change.

Not so much on the style preference, but more so on whether they're OK adding tensor_impl_ctr_kwargs to the to_affine_quantized_intx signature.

torchao/experimental/quant_api.py
```diff
@@ -571,107 +574,111 @@ def is_torchao_op_skippable(layout):
...
-                        "- weight_mapping_type to be MappingType.SYMMETRIC or MappingType.SYMMETRIC_NO_CLIPPING_ERR"
-                    )
-                assert (
-                    TORCH_VERSION_AT_LEAST_2_6
-                ), "aten target is requires torch version > 2.6.0"
-                if torch.backends.kleidiai.is_available():
-                    if isinstance(granularity, PerGroup):
-                        scale_dtype = (
-                            torch.bfloat16
-                        )  # KleidiAI kernel requires bfloat16 scale_dtype
```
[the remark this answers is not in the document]
nikhil-arm: Aten KleidiAI kernel requires scale_dtype as  torch.bfloat16 otherwise it would fallback to ref implementation. Also the input to aten kernel needs to be float32

torchao/dtypes/affine_quantized_tensor.py
vkuzo: not super familiar with this code, but as long as this doesn't change the BC surface sgtm

if this is controversial, can we separate this from the config migration? I'd love to see that piece land asap.

vkuzo (approved):
didn't read the code in detail, but it would be great to migrate this to config soon so we can disable the old path

please feel free to wait for a proper review if needed

torchao/experimental/packed_linear_int8_dynamic_activation_intx_weight_layout.py
```diff
@@ -44,61 +39,66 @@
...
+        return Target.AUTO
     elif target.lower() == "aten":
         return Target.ATEN
     else:
         raise ValueError(f"Invalid target: {target}")
 
 
 class PackedLinearInt8DynamicActivationIntxWeightLayout(Layout):
-    bit_width: Optional[int]
-    group_size: Optional[int]
```
drisspg, line 62: why remove thiese?

torchao/dtypes/affine_quantized_tensor.py
metascroy: The new tensor_impl_ctr_kwargs has a default value of None, so it shouldn't change how any existing call sites work.  The CI also passes.

vkuzo: sgtm, let's land if no other concerns?

torchao/experimental/packed_linear_int8_dynamic_activation_intx_weight_layout.py
metascroy: Added them back

torchao/dtypes/affine_quantized_tensor.py
metascroy: Ok I can land after CI passes.

I don't have concerns on the changes in torchao/experimental/*.  I mostly wanted feedback from someone in torchao on this change from torchao/dtypes/affine_quantized_tensor.py

vkuzo: I mean,

```
tensor_impl_ctr_kwargs: Optional[dict] = None,
```

is pretty hard to follow.  IMO it would be better to refactor the code to just pass bias directly instead of adding a layer of indirection.

However, this is nitty, not a part of BC, and I want to see the config part land, so how about we chat about ^ in parallel and if it needs fixing someone can do that in a future PR?

metascroy: The reason for using tensor_impl_ctr_kwargs  instead of "bias" is I thought it would be more extensible in future. 
 Currently if you use to_affine_quantized_intx to do quantization, there is no way to forward other args to your tensor subclass's "from_layout(data, scale, zero_point, _layout)" method.  Here we want to forward bias, but in future someone might want to forward something else.
 
What is currently done is the code in torchao/experimental has its own copy of to_affine_quantized_intx that supports bias.  The downside of this is I fear the two might drift apart going forward.

With all that said, I can refactor this PR to only contain the config change and put up the tensor_impl_ctr_kwargs change in another PR.

vkuzo: I don't get it, why can't we just add `bias` as an argument and pass it?

torchao/experimental/quant_api.py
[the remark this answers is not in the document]
nikhil-arm: bfloat16 only for groupwise. float32 for channelwise

PR #1838, opened 2025-03-05
enable compile with mxfp8 and mxfp4 cutlass gemm

vkuzo:
Summary:

Enables torch.compile e2e on MXLinear with the `mxfp8_cutlass` and `mxfp4_cutlass` recipes.  To do this, we need to add a tag on the op definitions to make inductor honor the col-major layout.

Note that something else is broken with these ops and bias=True - leaving that for a future PR.

Test Plan:

```
python benchmarks/float8/profile_lowp_training.py ~/local/tmp/20250305_test --mx_recipe_name mxfp8_cutlass

# uncomment the hardware check, then can also run the unit test
pytest test/prototype/mx_formats -s -k test_linear_compile
```

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo:
* (to be filled)

torchao/ops.py
```diff
@@ -27,8 +27,16 @@
 lib.define(
     "rowwise_scaled_linear_cutlass_s8s4(Tensor input, Tensor input_scale, Tensor weight, Tensor weight_scale, Tensor bias) -> Tensor"
 )
-lib.define("mx_fp8_bf16(Tensor a, Tensor b, Tensor a_scale, Tensor b_scale) -> Tensor")
-lib.define("mx_fp4_bf16(Tensor a, Tensor b, Tensor a_scale, Tensor b_scale) -> Tensor")
+# Note: we need to add the `torch._C.Tag.needs_fixed_stride_order` tag in order for inductor
+# to honor the layout constraints for `b` in the two ops below.
+lib.define(
```
drisspg, line 32: Whoops I meant to add this before

Good catch

drisspg (approved):

Issue #1821, opened 2025-03-03
Bias quantization for prequantized checkpoints


jackzhxng:
Following up from a chat with @jainapurva 

For private internal model enablement purposes, we would like to request support for bias quantization in prequantized checkpoint loading. At the moment we are doing manual source transformation after loading the prequantized checkpoint here https://github.com/pytorch/executorch/blob/main/examples/models/llama/source_transformation/pre_quantization.py#L40 into your deprecated `Int8DynActInt4WeightLinear`, which doesn't support bias quantization.

PR #1845, opened 2025-03-05
Add bias support for Int8DynActInt4WeightLinear

andrewor14:
**Summary:** Previously, when we see a linear with bias, we simply do not swap it to `Int8DynActInt4WeightLinear` and leave it as is. Now we do swap it, but bias is not quantized and passed to F.linear in full precision.

Fixes https://github.com/pytorch/ao/issues/1821

**Test Plan:**
python test/quantization/test_quant_api.py -k test_8da4w_quantizer_linear_bias

### DIFF
--- a/test/quantization/test_qat.py
+++ b/test/quantization/test_qat.py
@@ -1043,22 +1043,10 @@ class TestQAT(unittest.TestCase):
     )
     def test_replace_linear_8da4w(self):
         module = torch.nn.ModuleList(
-            [torch.nn.Linear(in_features=256, out_features=50, bias=True)]
-        )
-        _replace_linear_8da4w(
-            module,
-            256,
-            False,
-            torch.float32,
-            torch.float32,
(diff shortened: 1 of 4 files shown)


andrewor14:
> @andrewor14 By bias quantization, does it mean quantized bias, or just support for bias in quantization?

Right now this doesn't support quantized bias, just loading a quantized model where the linear has a non-existent bias. I think we can add actual bias quantization later. In general quantizing the bias doesn't seem to be very common, however, as the benefits are not super significant

jainapurva (approved):

jackzhxng (commented):
So with this PR, we will be able to represent unquantized bias with Int8DynActInt4WeightLinear?

andrewor14:
> So with this PR, we will be able to represent unquantized bias with Int8DynActInt4WeightLinear?

Yep

PR #1847, opened 2025-03-06
[wip] sparse marlin rocm compilation

jcaip:
cc @petrex I tried to get this working to run some benchmarks and was running into some issues. 

Does this require a specific ROCm version? I'm using 6.3.3 and I'm running into some stuff like
```
error: use of undeclared identifier '__builtin_amdgcn_global_load_lds'
```

Aside from that, I ran into a lot of issues with namespace issues with __half2, which I think I fixed by using `::__half2`, and one other bug with the hip version of `cudaSetAttribute` needing a function pointer.


jcaip:
Thanks @petrex, happy to share more debug info / hop on a call and walk you through the issues if that's helpful.

PR #1848, opened 2025-03-07
Add stack for qkv projection

drisspg:


torchao/dtypes/floatx/float8_layout.py
```diff
@@ -154,6 +154,50 @@ def __torch_dispatch__(cls, func, types, args, kwargs):
...
+
+            # Check if all tensors are Float8AQTTensorImpl
+            if all(isinstance(t, Float8AQTTensorImpl) for t in tensors):
+                # Stack the float8_data tensors
+                stacked_float8_data = func([t.float8_data for t in tensors], dim=dim)
+
+                # Handle scales based on their dimensionality
+                if all(t.scale.ndim == 0 for t in tensors):
+                    # Use the maximum scale value from all tensors for numerical stability
+                    stacked_scale = torch.max(torch.cat([t.scale for t in tensors]))
```
vkuzo, line 170: would this change numerics if max_scale != original scale?  I would have expected something like a rescale of the data if the scale is changed.

PR #1853, opened 2025-03-07
migrate prototype/awq to configs

vkuzo:
Summary:

As titled

Test Plan:

```
// note: this fails on weights only load, but the failure happens
// after my changes, and already exists on main branch
pytest test/prototype/test_awq.py -s -x
```

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo:
* (to be filled)

danielvegamyhre (approved):
Change itself looks good to me but this doesn't sound good:

```
// note: this fails on weights only load, but the failure happens
// after my changes, and already exists on main branch
pytest test/prototype/test_awq.py -s -x
```

If the bug exists on main and is unrelated to these changes, how has the test been passing CI?  Sounds like we should create a bug to track this and maybe the feature owner can take a look

drisspg (approved):
Is the model change purposeful

vkuzo:
> Is the model change purposeful

ah, that snuck in.  No, will remove.

vkuzo:
> Change itself looks good to me but this doesn't sound good:
> 
> ```
> // note: this fails on weights only load, but the failure happens
> // after my changes, and already exists on main branch
> pytest test/prototype/test_awq.py -s -x
> ```
> 
> If the bug exists on main and is unrelated to these changes, how has the test been passing CI? Sounds like we should create a bug to track this and maybe the feature owner can take a look

in general I agree, but given that this is the prototype folder the responsibility to do what you suggested would be on the owner of awq - I'm not sure who that is :)  In general for prototype folder there aren't guarantees that we'll spend our time to migrate things, if people want their thing to continue to keep working they should work on moving the feature out of prototype.

PR #1851, opened 2025-03-07
config migration: smoothquant

vkuzo:
Summary:

Migrates smoothquant to direct configs

Test Plan:

```
pytest test/prototype/test_smoothquant.py -s -x
```

Reviewers:

Subscribers:

Tasks:

Tags:

PR #1854, opened 2025-03-07
migrates prototype/mixed_precision to configs

vkuzo:
Summary:

Note: had to remove int4/int8 functionality to simplify the refactor.
Whoever uses this script next and needs that functionality can add this
back.

Test Plan:

```
pytest test/prototype/test_mixed_precision.py -s -x
```

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo:
* (to be filled)

torchao/prototype/quantization/mixed_precision/scripts/naive_intNwo.py
```diff
@@ -18,6 +23,25 @@ def intN_weight_only(group_size=32, n=8, symmetric=False):
         quantize_(model, intN_weight_only(n=your_bit_choice, group_size=group_size), optional_filter_func_for_desired_layers_to_quantize)
     """
 
+    group_size: int = 32
+    n: int = 8
+    symmetric: bool = False
+
+
+# for bc
+intN_weight_only = IntNWeightOnlyConfig
```
drisspg, line 32: Imo these prototype folders we should just break BC

drisspg (approved):
See comment

torchao/prototype/quantization/mixed_precision/scripts/naive_intNwo.py
vkuzo: I agree, but it's actually easier to keep BC here (to not change callsites) - if someone wants to change the callsites of all these prototype features, that would sgtm in a separate PR.

drisspg: I guess, but also changing callsites within a repro is - I find - not that hard and ultimately helps create a more unified api as we move everything to configs + registrations

vkuzo: ok will do

vkuzo: @drisspg , I take it back.  See how the scope of https://github.com/pytorch/ao/pull/1851 expanded from removing the bc name.  Some of these other prototypes have even more callsites throughout the codebase, some of those callsites aren't easily testable, etc - I just don't think it's worth the effort to tie removing old names to this PR stack since we have to keep BC anyways for some of the APIs.

PR #1857, opened 2025-03-07
migrate nf4 to configs

vkuzo:
Summary:

1. migrate nf4 to configs
2. create a separate, temporary, workflow file for nf4 to get around
   circular import errors. A future person should properly make nf4 a
   workflow in torchao's dir structure, but that is out of scope for
   this PR.

Note that this technically breaks "BC", but nf4 isn't an official API so there isn't really BC.  This does not break the torchtune integration.

Test Plan:

```
pytest test/dtypes/test_nf4.py
```

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo:
* (to be filled)

andrewor14 (approved):

PR #1867, opened 2025-03-11
Add tensor_impl_kwargs to to_affine_quantized_intx

metascroy:
Tensor subclasses in torchao/experimental pack unquantized bias with the quantized weights during quantize_ for performance reasons.  This requires they pass bias to the subclass impl constructor, but there is no way to to do with to_affine_quantized_intx.

This PR adds tensor_impl_kwargs to to_affine_quantized_intx and uses it in torchao/experimental.


torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -204,6 +204,7 @@ def from_hp_to_intx(
         zero_point_domain: ZeroPointDomain = ZeroPointDomain.INT,
         _layout: Layout = PlainLayout(),
         use_hqq: bool = False,
+        tensor_impl_ctr_kwargs: Optional[dict] = None,
```
vkuzo, line 207: IMO it would be better to either spell out the argument you want explicitly, or create a new function for what you need.

vkuzo (changes requested):
the intent sgtm, requesting changes to implement it directly instead of creating a new layer of indirection

torchao/dtypes/affine_quantized_tensor.py
metascroy: I think @jerryzh168 rejected a previous change that spelled out bias explicitly.  He instead asked that we create a new subclass of _AffineQuantizedTensor.

We can stick with this plan, but it does duplicate almost all of the code in from_hp_to_intx.

jerryzh168: I think it's fine to duplicate code to keep the main path simple, we also want to split the code for hqq as well

PR #1869, opened 2025-03-11
triton kernel to cast to mx across dim0 and dim1

vkuzo:
Summary:

This is a start of the short term workaround for https://github.com/pytorch/pytorch/issues/148682 .  tl;dr; is that today torch.compile is not yet generating good fusions for the cast to mx across dim0 and dim1, so we can add custom kernels to do this until the torch.compile support is better.  We take a simple strategy to minimize overall reads and writes: read the input in 32x32 tiles, and write out the row-major and col-major casted values and scales directly from each tile.

The kernel added in this PR is ok-ish: it's ~2x faster than torch.compile across a range of shapes with MKN from 1024 to 16384, but it only hits around (redacted) 8 TBPS peak mem bandwidth.  Impact on e2e benchmarks of fwd+bwd of MXLinear:

```
# K = 1024: speedup over bf16 0.19 -> 0.28
# K = 16384: speedup over bf16 0.91 -> 1.07
```

Some future improvements we can make:
1. add an outer_tile and set it to 128, keeping inner tile at 32 (see below, will likely get us to redacted of peak bandwidth).  128 because outer_tile 256+ seems to run out of shared memory on B200.
2. write out swizzled scales (not currently in this benchmark, but will help with e2e)
3. see if we can make the col-major output and row-major and col-major writes coalesced (didn't do any research on how yet)

For now I want to check this in as just a kernel + benchmark without an integration to `MXLinear`.  If we make this kernel significantly faster, we can integrate. 

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo:
* (to be filled)

torchao/prototype/mx_formats/custom_cast.py
```diff
@@ -1080,3 +1082,240 @@ def _(uint8_data):
...
+
+        # quant to float8
+        # TODO(this PR): clamp?
+        col_normalized = col_normalized.to(tl.float8e4nv)
+
+        # Store the row-normalized result in row-major format
+        tl.store(output_row_major_ptr + row_major_offsets, row_normalized, mask=mask)
+
+        # Store the column-normalized result in column-major format
+        tl.store(output_col_major_ptr + col_major_offsets, col_normalized, mask=mask)
```
vkuzo, line 1202: if I comment this line out, I get to 60% of peak mem bandwidth (nearly matching an "add one" kernel). uncoalesced writes maybe...

torchao/prototype/mx_formats/custom_cast.py
```diff
@@ -1080,3 +1082,498 @@ def _(uint8_data):
...
+        col_scale = torch.empty(
+            n_cols, n_rows // inner_block_size, dtype=torch.uint8, device=x.device
+        )
+
+        # Calculate grid dimensions based on tile size
+        grid_rows = triton.cdiv(n_rows, row_tile_size)
+        grid_cols = triton.cdiv(n_cols, col_tile_size)
+
+        # inner_block_size = 32
+        rename_me_tile_size = row_tile_size * col_tile_size // inner_block_size
```
eellison, line 1514: lol i wasnt sure what to call this either..

torchao/prototype/mx_formats/custom_cast.py
```diff
@@ -1080,3 +1082,498 @@ def _(uint8_data):
...
+            n_cols=n_cols,
+            COL_TILE_SIZE=col_tile_size,
+            ROW_TILE_SIZE=row_tile_size,
+            RENAME_ME_TILE_SIZE=rename_me_tile_size,
+            INNER_BLOCK_SIZE=inner_block_size,
+        )
+
+        return (
+            output_col_major.t(),
+            col_scale.reshape(-1, 1).view(torch.float8_e8m0fnu),
```
eellison, line 1531: Because the triton is just accessing `data_ptr` you could initialize these as the correct layout if you wanted.

torchao/prototype/mx_formats/custom_cast.py
```diff
@@ -1080,3 +1082,498 @@ def _(uint8_data):
...
+        # add offset for inner blocks
+        factor = ROW_TILE_SIZE // INNER_BLOCK_SIZE
+
+        # needs better name
+        jump_vals_per_col = (n_rows - ROW_TILE_SIZE) // INNER_BLOCK_SIZE
+        # tl.device_print("jump_vals_per_col", jump_vals_per_col)
+
+        # [0, 1, 2, 3, 4, 5, 6, 7] -> [0, 1, 4, 5, 8, 9, 12, 13]
+        col_scale_indices = col_scale_indices + (
+            tl.floor(col_scale_indices / factor) * jump_vals_per_col
```
eellison, line 1460: any reason to not just use integer division here instead of `/` and floor ?

torchao/prototype/mx_formats/custom_cast.py
```diff
@@ -1080,3 +1082,498 @@ def _(uint8_data):
...
+        ).to(tl.int32)
+        # tl.static_print(col_scale_indices)
+        # tl.device_print("indices", col_scale_indices)
+
+        # TODO(future): mask
+        tl.store(col_scale_start_ptr + col_scale_indices, col_scale_e8m0)
```
eellison, line 1466: Should we add some runtime asserts about block divisibility ?

torchao/prototype/mx_formats/custom_cast.py
```diff
@@ -1080,3 +1082,498 @@ def _(uint8_data):
...
+        # bf16_exp_bias = 127
+
+        # Find the maximum absolute value for each row
+        max_abs = tl.max(x, axis=axis)
+        # return max_abs, max_abs.to(tl.uint8)
+
+        # TODO 1: rewrite as bit shifts, see https://github.com/pytorch/ao/pull/1908/files
+        # before: 1.7 TB/s
+        # after: ?
+        scale_e8m0_unbiased = tl.floor(tl.log2(max_abs + epsilon)) - target_max_pow2
```
eellison, line 1109: I would suspect theres a better way of getting the scale avoiding doing a float log. i guess this is what the TODO is for.

eellison (approved):
looks good ! did not review indexing math too closely but hoepfully will be replaced with torch.compile soon.

vkuzo:
closing this in favor of https://github.com/pytorch/ao/pull/1932, I split the dim1 logic so we can have this PR as an easy historical reference

PR #1873, opened 2025-03-11
[float8] Bug fix: do not override requires_grad=False when enable_float8_all_gather=True

### LINKED ISSUE #1871: `Float8Linear.from_float()` overrides `requires_grad=False` when `enable_fsdp_float8_all_gather=True`
Hi, I think the issue is here: https://github.com/pytorch/ao/blob/d3b481b55c9717eefcb9132a01661036a7ea39d6/torchao/float8/float8_linear.py#L414-L420

I'm guessing we could fix it by changing that to:

```python
            new_mod.weight = torch.nn.Parameter(
                WeightWithDynamicFloat8CastTensor(
                    new_mod.weight,
                    new_mod.linear_mm_config,
                    new_mod.config.cast_config_weight.target_dtype,
                ),
                requires_grad=new_mod.weight.requires_grad
            )
```

danielvegamyhre:
Fixes #1871

I confirmed the update unit test fails without the fix, and passes with the fix.

cc @epwalsh who reported the bug

### DIFF
--- a/test/float8/test_fsdp2/test_fsdp2.py
+++ b/test/float8/test_fsdp2/test_fsdp2.py
@@ -67,8 +67,9 @@ class TestFloat8Common:
         return module
 
     def init_transformer(
-        self, weight_tying: bool, 
-        dtype: Optional[torch.dtype] = None, 
+        self,
+        weight_tying: bool,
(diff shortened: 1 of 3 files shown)


vkuzo (approved):
thank you!

PR #1875, opened 2025-03-12
Config Serilization No Deps

drisspg:
# Summary
Similar to https://github.com/pytorch/ao/pull/1806 but without using Pydantic

This PR introduces functionality for serializing and deserializing `AOBaseConfig` objects to and from dictionaries 

The new `config_to_dict` and `config_from_dict` functions allow these configurations to be:

- Saved to disk
- Shared between applications
- Versioned for compatibility

Another main difference from that PR is that AOBaseConfig remains unchanged besides the overridable class var which is set to 1 by default

This makes it so that configs can opt-in, which hopefully is as cheap is adding your config to the test file, and ensuring that all types that are needed for reconstruction can be found in torchao.quantization import module

### Serialization Format

When an `AOBaseConfig` object is serialized, it creates a dictionary with:

```python
{
    "_type": "ConfigClassName",  # Just the class name, not full module path
    "_version": 1,               # Version from the class's VERSION attribute
    "_data": {                   # Actual configuration parameters
        "param1": value1,
        "param2": value2,
        # Nested objects also get serialized with their types
    }
}
```

### Special Cases

The `ConfigJSONEncoder` handles several types of objects:

1. **AOBaseConfig subclasses**: Primary config objects
2. **NamedTuple types**: Common for grouped parameters
3. **Dataclasses**: Needed for our granularity objects
4. **torch.dtype objects**: the 1 native torch object that so far shows up in our configs
5. **Layout objects**: Layout objects 
6. **Enum values**: For options like `MappingType.SYMMETRIC`

### Version Management

Each config class defines a `VERSION` class variable (defaulting to 1). During deserialization, we enforce:

```python
if stored_version != current_version:
    raise VersionMismatchError(type_path, stored_version, current_version)
```

This prevents using outdated configurations with newer code that might expect different parameters.

### Restricted Reconstruction

I only allow importing types from torchao.quantization module path or from torch

```python
try:
    module = importlib.import_module("torchao.quantization")
    cls = getattr(module, type_path)
except (ImportError, AttributeError) as e:
    raise ValueError(f"Failed to find class {type_path} in torchao.quantization: {e}")
```

### Serializing a Configuration

```python
from torchao.core.config import config_to_dict
import json

# Create a configuration
config = Int4WeightOnlyConfig(group_size=64)

# Convert to dictionary
config_dict = config_to_dict(config)

# Save to JSON file
with open("config.json", "w") as f:
    json.dump(config_dict, f)
```

This produces
```Json
{"_type": "Int4WeightOnlyConfig", "_version": 1, "_data": {"group_size": 64, "layout": {"_type": "TensorCoreTiledLayout", "_version": 1, "_data": {"inner_k_tiles": 8}}, "use_hqq": false, "zero_point_domain": {"_type": "ZeroPointDomain", "_data": "NONE"}}}
```

### Deserializing a Configuration

```python
from torchao.core.config import config_from_dict
import json

# Load from JSON file
with open("config.json", "r") as f:
    config_dict = json.load(f)

# Rebuild configuration object
config = config_from_dict(config_dict)

print(config)
```

Prints
```Shell
Int4WeightOnlyConfig(group_size=64, layout=TensorCoreTiledLayout(inner_k_tiles=8), use_hqq=False, zero_point_domain=<ZeroPointDomain.NONE: 3>)
```

## YAML Support

The serialized format uses basic types (strings, numbers, booleans, lists, dictionaries) that are compatible with YAML. Users can:

1. Serialize to dict using `config_to_dict`
2. Use PyYAML to convert the dict to YAML
3. For deserialization, parse YAML to dict, then use `config_from_dict`

Example:
```python
import yaml
from torchao.core.config import config_to_dict, config_from_dict

# Serialize to YAML
config_dict = config_to_dict(my_config)
yaml_str = yaml.dump(config_dict)

# Deserialize from YAML
loaded_dict = yaml.safe_load(yaml_str)
loaded_config = config_from_dict(loaded_dict)
```
The yaml string 

```YAML
_data:
  group_size: 64
  layout:
    _data:
      inner_k_tiles: 8
    _type: TensorCoreTiledLayout
    _version: 1
  use_hqq: false
  zero_point_domain:
    _data: NONE
    _type: ZeroPointDomain
_type: Int4WeightOnlyConfig
_version: 1
```

## Future Improvements

- Migration logic for handling version changes
- We could add more machinery for making custom types in AOBaseconfigs easier to support

As well we should ensure that all quant-api's configs are added to the test files

### DIFF
--- a/test/quantization/test_config_serialization.py
+++ b/test/quantization/test_config_serialization.py
@@ -1,0 +1,172 @@
+import json
+import os
+import tempfile
+from dataclasses import dataclass
+from unittest import mock
+
+import pytest
+import torch
+
+from torchao.core.config import (
+    AOBaseConfig,
+    VersionMismatchError,
+    config_from_dict,
+    config_to_dict,
+)
+from torchao.quantization.quant_api import (
+    Float8DynamicActivationFloat8WeightConfig,
+    Float8WeightOnlyConfig,
+    FPXWeightOnlyConfig,
+    GemliteUIntXWeightOnlyConfig,
+    Int4DynamicActivationInt4WeightConfig,
+    Int4WeightOnlyConfig,
+    Int8DynamicActivationInt4WeightConfig,
+    Int8DynamicActivationInt8WeightConfig,
+    Int8WeightOnlyConfig,
+    PerRow,
+    UIntXWeightOnlyConfig,
+)
+from torchao.sparsity.sparse_api import BlockSparseWeightConfig, SemiSparseWeightConfig
+
+# Define test configurations as fixtures
+configs = [
+    Float8DynamicActivationFloat8WeightConfig(),
+    Float8DynamicActivationFloat8WeightConfig(granularity=PerRow()),
+    Float8WeightOnlyConfig(
+        weight_dtype=torch.float8_e4m3fn,
+    ),
+    UIntXWeightOnlyConfig(dtype=torch.uint1),
+    Int4DynamicActivationInt4WeightConfig(),
+    Int4WeightOnlyConfig(
+        group_size=32,
+    ),
+    Int8DynamicActivationInt4WeightConfig(
+        group_size=64,
+    ),
+    Int8DynamicActivationInt8WeightConfig(),
+    # Int8DynamicActivationInt8WeightConfig(layout=SemiSparseLayout()),
+    Int8WeightOnlyConfig(
+        group_size=128,
+    ),
+    UIntXWeightOnlyConfig(
+        dtype=torch.uint3,
+        group_size=32,
+        use_hqq=True,
+    ),
+    GemliteUIntXWeightOnlyConfig(
+        group_size=128,  # Optional, has default of 64
+        bit_width=8,  # Optional, has default of 4
+        packing_bitwidth=8,  # Optional, has default of 32
+        contiguous=True,  # Optional, has default of None
+    ),
+    FPXWeightOnlyConfig(ebits=4, mbits=8),
+    # Sparsity configs
+    SemiSparseWeightConfig(),
+    BlockSparseWeightConfig(blocksize=128),
+]
+
+
+# Create ids for better test naming
+def get_config_ids(configs):
+    if not isinstance(configs, list):
+        configs = [configs]
+    return [config.__class__.__name__ for config in configs]
+
+
+@pytest.mark.parametrize("config", configs, ids=get_config_ids)
+def test_reconstructable_dict_file_round_trip(config):
+    """Test saving and loading reconstructable dicts to/from JSON files."""
+    # Get a reconstructable dict
+    reconstructable = config_to_dict(config)
+
+    # Create a temporary file to save the JSON
+    with tempfile.NamedTemporaryFile(
+        mode="w+", suffix=".json", delete=False
+    ) as temp_file:
+        # Write the reconstructable dict as JSON
+        json.dump(reconstructable, temp_file)
+        temp_file_path = temp_file.name
+
+    try:
+        # Read back the JSON file
+        with open(temp_file_path, "r") as file:
+            loaded_dict = json.load(file)
+
+        # Reconstruct from the loaded dict
+        reconstructed = config_from_dict(loaded_dict)
+
+        # Check it's the right class
+        assert isinstance(reconstructed, config.__class__)
+
+        # Verify attributes match
+        for attr_name in config.__dict__:
+            if not attr_name.startswith("_"):  # Skip private attributes
+                original_value = getattr(config, attr_name)
+                reconstructed_value = getattr(reconstructed, attr_name)
+
+                # Special handling for torch dtypes
+                if (
+                    hasattr(original_value, "__module__")
+                    and original_value.__module__ == "torch"
+                ):
+                    assert (
+                        str(original_value) == str(reconstructed_value)
+                    ), f"Attribute {attr_name} mismatch after file round trip for {config.__class__.__name__}"
+                else:
+                    assert (
+                        original_value == reconstructed_value
+                    ), f"Attribute {attr_name} mismatch after file round trip for {config.__class__.__name__}"
+
+    finally:
+        # Clean up the temporary file
+        if os.path.exists(temp_file_path):
+            os.unlink(temp_file_path)
+
+
+# Define a dummy config in a non-allowed module
+@dataclass
+class DummyNonAllowedConfig(AOBaseConfig):
+    VERSION = 2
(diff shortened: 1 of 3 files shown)


torchao/quantization/__init__.py
```diff
@@ -46,7 +46,9 @@
     AffineQuantizedObserverBase,
 )
 from .quant_api import (
+    CutlassInt4PackedLayout,
     Float8DynamicActivationFloat8WeightConfig,
+    Float8MMConfig,
```
vkuzo, line 51: why is this needed?

test/quantization/test_config_serialization.py
```diff
@@ -0,0 +1,116 @@
...
+
+# Create ids for better test naming
+def get_config_ids(configs):
+    if not isinstance(configs, list):
+        configs = [configs]
+    return [config.__class__.__name__ for config in configs]
+
+
+@pytest.mark.parametrize("config", configs, ids=get_config_ids)
+def test_reconstructable_dict_file_round_trip(config):
```
vkuzo, line 66: looks great!

torchao/core/config.py
```diff
@@ -26,4 +33,214 @@ def _transform(
...
+            return result
+        except TypeError:
+            pass
+
+        # Default case - return as is
+        # (This will be processed by standard JSON encoder later)
+        return value
+
+
+def config_to_dict(config: AOBaseConfig) -> Dict[str, Any]:
```
vkuzo, line 152: question: why not put this and `config_from_dict` as methods on `AOBaseConfig`?

vkuzo:
> ensuring that all types that are needed for reconstruction can be found in torchao.quantization import module. "I only allow importing types from torchao.quantization module path or from torch".

Can this be relaxed to allow prototype and sparsity configs to be serializable?

Thoughts on how handling of static scales would work with this?

Overall looks great!

torchao/quantization/__init__.py
drisspg: For provenance made it so that we only allow reconstruction imports in module `torchao.quantization` which is our BC API

drisspg:
1. Added a global allow lists that can be used for allowing more config locations

2. For handling static scales thats interesting. It is not possible to save a tensor to json ( well you can but not comically so)
So you would update the type specific handlers to add special logic for tensors. Probably saying "scale = STATIC" or idk up to us. And then at reconstruction time you need to figure out where we put that data, which realistically should live on the subclasses stored in the state dict. SO you need to essentially init the config in an invalid state (scale = PLACEHOLDER) and then after the fact inject the right scale, a little handwavy but I think its possible

torchao/quantization/__init__.py
drisspg: And really this should have been here

torchao/core/config.py
andrewor14: I think it's better to just expose `AOBaseConfig` rather than 2 separate functions. We need to expose `AOBaseConfig` anyway for users who wish to write custom configs, since we are removing the ability to pass in Callables into quantize_ in this release

torchao/core/config.py
```diff
@@ -26,4 +33,226 @@ def _transform(
...
+            try:
+                return cls(obj_data)
+            except:
+                return obj_data
+
+    # Process nested structures for dictionary obj_data
+    processed_data = {}
+    for key, value in obj_data.items():
+        if isinstance(value, dict) and "_type" in value and "_data" in value:
+            # Recursively handle nested configs
```
andrewor14, line 241: Do we have any existing configs that have nested configs or fields with data structures? I feel our configs so far have been pretty simple?

torchao/core/config.py
```diff
@@ -26,4 +33,226 @@ def _transform(
...
+            cls = getattr(module, type_path)
+            break  # Found the class, exit the loop
+        except (ImportError, AttributeError):
+            continue  # Try the next module
+
+    # If we couldn't find the class in any allowed module, raise an error
+    if cls is None:
+        allowed_modules_str = ", ".join(ALLOWED_AO_MODULES)
+        raise ValueError(
+            f"Failed to find class {type_path} in any of the allowed modules: {allowed_modules_str}"
```
andrewor14, line 217: What's the motivation for this? Why not just make all configs, including custom user configs, serializable?

torchao/core/config.py
drisspg: Sorry one other point is that for config_from_dict you dont know the base config youare trying to builld from and having this as the ABCs' class method feels kinda weird to me. These should mostly be implementation details for other libraries and really called by users

andrewor14 (approved):
Looks great! Definitely prefer this to the Pydantic version and separating serde from the string API. Just a couple questions:

1. How do we want versioning to work exactly? We definitely need to bump the version when we remove or modify an existing field, but do we necessarily need to bump the version when adding a new field with a default value? In this case, should we still fail on version mismatch or just let it load the default value?

2. Do we need to handle nested configs / data structures? I feel inner configs (e.g. `Float8MMConfig`) don't need to be `TorchAOBaseConfig`, so technically we just need to handle inner dataclasses/named tuples?

torchao/core/config.py
drisspg: Vasiliy mentioned there might be one for float8 training but I could find it for config wrapping configs

We do have configs that wrap objects that need to be handled specially, the float inference configs have this where MMconfig is internal to it and needs special care

vkuzo: > We do have configs that wrap objects that need to be handled specially, the float inference configs have this where MMconfig is internal to it and needs special care

I don't think MMConfig should be in the BC surface / serializable / etc, that seems like a code smell.

drisspg: I disagree it is a convenient  way for users to modulate mm behavior on a per instance level and unpacking this struct into three separate booleans doen't feel beneficial. 

That being said there are alot of examples of these nested objects, `Layouts`, `granularities`, etc.. for which this behavior is necessary

torchao/quantization/__init__.py
```diff
@@ -56,6 +58,8 @@
     Int8DynamicActivationInt4WeightConfig,
     Int8DynamicActivationInt8WeightConfig,
     Int8WeightOnlyConfig,
+    PlainLayout,
```
vkuzo, line 61: is this exposed in public API / on HF / etc?

I don't think this abstraction is clean and I don't want to expose it, this is not really a "layout" rather than a "layout with some other arguments mixed in for convenience".  We should talk about what to do with this before exposing it.

PR #1879, opened 2025-03-13
Add bias support to torchao kernels

metascroy:
Summary: This diff adds bias support for torchao kernels / quantizer.

Differential Revision: D71093679


Jack-Khuu (commented):
Passes the gut tests

Minor nit to use ternary operator, but up to you whether you want to grep replace it

Jack-Khuu (approved):

PR #1886, opened 2025-03-13
Add module swap quantization API from Quanty

andrewor14:
**Summary:** This commit adds a module-swap-based PTQ API from Quanty, including:

- Quantized linear and embedding modules
- `IntQuantizer` to specify how to quantize weights and activations
- `CodeBookQuantizer` as an alternative to IntQuantizer
- Implementation of K-means to be used for codebook quantization
- Range setting and data getter utility

These new APIs will complement our existing `quantize_` API, which is primarily used for tensor-subclass-based quantization today (though it can also support module swaps). All APIs introduced in this commit are under prototype and highly subject to change. In particular, we plan to delete `quantize_module_swap` and `QuantizationRecipe`, and instead integrate this flow with the `quantize_` API by creating a new `AOBaseConfig`.

All code is migrated from Quanty and written by @TiRune.

**Test Plan:**
python test/quantization/module_swap/test_*

### DIFF
--- a/test/prototype/module_swap_quantization/test_kmeans_codebook.py
+++ b/test/prototype/module_swap_quantization/test_kmeans_codebook.py
@@ -1,0 +1,56 @@
+import copy
+import unittest
+from typing import Union
+
+import torch
+import torch.nn as nn
+
+from torchao.prototype.quantization.module_swap import (
+    CodeBookQuantizer,
+    QuantizedLinear,
+)
+from torchao.prototype.quantization.module_swap.algorithms import kmeans_codebook
+
+
+class SimpleTestNetwork(nn.Module):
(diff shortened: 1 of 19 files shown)


test/quantization/module_swap/test_kmeans_codebook.py
```diff
@@ -0,0 +1,56 @@
+import copy
```
vkuzo, line 1: IMO these tests should be in `test/prototype/quantization/module_swap`

vkuzo (approved):
a short README.md of `prototype/module_swap_quantization` would be nice as well

andrewor14:
Thanks, added the README. Merging this!

PR #1890, opened 2025-03-14
[float8] Make check for rowwise scales more explicit/readable

danielvegamyhre:
## Description

Make check for rowwise scales more explicit/readable in `addmm_float8_unwrapped`.


vkuzo (approved):

vkuzo:
fix ruff before landing

jerryzh168:
this should be `not user facing` I think, since it does not affect end users

PR #1907, opened 2025-03-17
mx cast performance bench

vkuzo:
Summary:

On B200, our MX cast is no longer memory bound.
* https://www.internalfb.com/phabricator/paste/view/P1758361186 has a list of things we can do to improve performance
* this PR creates a simple benchmark so we can track how each change impacts performance

Test Plan:

Current results, before any improvements: https://www.internalfb.com/phabricator/paste/view/P1758363159

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo:
* (to be filled)

drisspg (approved):

PR #1911, opened 2025-03-17
mx cast: remove clamping of output tensor for torch.compile path

vkuzo:
Summary:

In eager mode, cast to float8 is unsaturated
In triton, it is saturated
For this kernel, we are compute bound so doing the saturated cast in
triton without the clamp provides a speedup

Filed placeholder issue (in torchao) to add a way to cast to float8 with saturation in eager mode: https://github.com/pytorch/ao/issues/1912.  We may need some core changes for this.

Speedup after vs before: 1.52x

* bench before: https://www.internalfb.com/phabricator/paste/view/P1758491240
* bench after: https://www.internalfb.com/phabricator/paste/view/P1758531336

Test Plan:

```
```

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo:
* (to be filled)

drisspg (approved):

PR #1924, opened 2025-03-19
Update float8 tests to run nightly

jainapurva:
Update the Float8 tests to run nightly or as an action on CI, instead of running on every PR.

### DIFF
--- a/.github/workflows/float8_test.yml
+++ b/.github/workflows/float8_test.yml
@@ -50,7 +50,3 @@ jobs:
         pip install .
         pytest test/float8 --verbose -s
         pytest test/dtypes/test_affine_quantized_float.py --verbose -s
-        # Float8nocompile tests
-        cd torchao/prototype/float8nocompile
-        pytest kernels/ --verbose -s
-        pytest test/train_test.py --verbose -s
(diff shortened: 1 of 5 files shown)


jainapurva:
* (to be filled)

.github/workflows/float8_test.yml
```diff
@@ -53,3 +49,8 @@ jobs:
         pip install -r dev-requirements.txt
         pip install .
         pytest test/float8 --verbose -s
+        pytest test/dtypes/test_affine_quantized_float.py --verbose -s
+        # Float8nocompile tests
+        cd torchao/prototype/float8nocompile
```
danielvegamyhre, line 54: Thanks for thinking about float8nocompile for this, but IMO we shouldn't mix in prototype tests with production float8 tests in the nightly CI. The reason is I don't think prototype bugs should cause production test suites to fail or block releases, etc.

IMO float8nocompile tests should only run on changes to the code in float8nocompile/ directory (as is the current configuration [here](https://github.com/pytorch/ao/blob/ab3792e3d91e04f85992a659c1664a6a1a6d733c/.github/workflows/float8nocompile_test.yaml#L15)).

jainapurva: Ok, makes sense. I've removed those from the workflow.

jainapurva:
Not needed anymore

PR #1940, opened 2025-03-24
Add config selection for row-wise scaled FP8 sparse CUTLASS-based kernel

alexsamardzic:


alexsamardzic:
Improves Llama generator performance for about 10%, also visibly improves performance for most shapes in the [benchmark](https://github.com/pytorch/ao/blob/main/benchmarks/benchmark_rowwise_scaled_linear_sparse_cutlass.py).

drisspg (approved):
Any perf numbers would be great to see 😊

Issue #1936, opened 2025-03-21
[Torchao Experimental] Create meta tensors directly in meta kernels


metascroy:
In meta kernels in torchao/experimental/ops, we do things like:

```
return torch::empty({num_out, k}).to("meta");
```

We should create meta tensors directly if possible.

swolchok:
suspect it's`torch::empty({num_out, k}, torch::kMeta)`

swolchok:
(and you may not strictly need to prefix `kMeta` with `torch::`

metascroy:
> (and you may not strictly need to prefix `kMeta` with `torch::`)

I'll try this and put up a PR next week.

metascroy:
Fixed in [Add dynamic shape support for lowbit kernels #1942](https://github.com/pytorch/ao/pull/1942)

PR #1942, opened 2025-03-24
Add dynamic shape support for lowbit kernels

metascroy:
To support dynamic shapes, the linear/embedding meta kernels must be written in python.

The packing meta kernels are still in C++ because 1) packing shapes are not dynamic, and 2) the packed size is not readily available in python.  For these, I move to create meta tensors directly per https://github.com/pytorch/ao/issues/1936.

### DIFF
--- a/torchao/experimental/tests/test_int8_dynamic_activation_intx_weight.py
+++ b/torchao/experimental/tests/test_int8_dynamic_activation_intx_weight.py
@@ -216,6 +216,53 @@ class TestInt8DynamicActivationIntxWeight(unittest.TestCase):
             aoti_results = fn(activations)
             self.assertTrue(torch.allclose(eager_results, aoti_results))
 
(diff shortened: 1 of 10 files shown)


torchao/experimental/tests/test_int8_dynamic_activation_intx_weight.py
```diff
@@ -216,6 +216,53 @@ def test_export_compile_aoti_PackedLinearInt8DynamicActivationIntxWeightLayout(
...
+    def test_export_dynamic_shape_PackedLinearInt8DynamicActivationIntxWeightLayout(
+        self,
+    ):
+        granularity = PerRow()
+        m = 3
+        k0 = 512
+        k1 = 256
+        k2 = 128
+        k3 = 1024
+        weight_dtype = torch.int4
```
jackzhxng, line 228: Maybe would be good to loop through and test all of the possible granularities from 2-7

jackzhxng (approved):
Synced offline, feel free to merge

PR #1963, opened 2025-03-26
Update linear.h

metascroy:
Summary:
This diff updates the interface for universal kernels to mirrored the update we made in KleidiAI a couple weeks ago.  The changes do look large, but most are moving files around.

Changes:

* Within cpu/aarch64/linear, there is a new folder channelwise_8bit_activation_groupwise_lowbit_weight that contains:
    * include.h (new version of linear.h, with shared activation packing and weight packing functions), the 3 kernel files (moved, but no changes made), pack_activations (moved but no changes), and pack_weights (moved but no changes).
    * The new shared embedding ops (this diff stack) need a new include/namespace update to reflect these changes.

Note that linear.h is preserved to not break existing code at op level, but eventually it will be deleted and the op code will use the new interface.

Reviewed By: digantdesai

Differential Revision: D71357293


metascroy:
@digantdesai can I get a stamp on PR.  You already approved diff.

manuelcandales (approved):

PR #1964, opened 2025-03-26
Match QAT prepare and convert numerics exactly

andrewor14:
**Summary:** Previously, `Int8DynActInt4QATQuantizer` had slightly diverging numerics between the prepare and convert steps. This is because the prepare step uses quantization primitives shared with AQT (specifically `quantize_affine` and `dequantize_affine`), while the convert step relies on old ops from the `torch.ops.quantized_decomposed` namespace. The diverging numerics is negligible for small models, but the quantization errors begin to compound for larger models with many linear layers.

More specifically, there are three different places where the divergence occurs during activation quantization:

1. **Choose qparams.** The prepare step casts the qparams to `torch.float32`, whereas the convert step casts the scales to `torch.float64` and zero points to `torch.int64`.

2. **Quantize.** The prepare step performs round before adding zero points and uses torch functions, while the convert step adds before rounding and uses torch tensor methods.
```
# Prepare
x = torch.clamp(
    torch.round(x * (1.0 / scale)) + zero_point, qmin, qmax,
)

# Convert
x = ( 
    x.mul(1.0 / scale)
    .add(zero_point)
    .round()
    .clamp(qmin, qmax)
    .to(quantize_dtype)
)
```

3. **Dequantize.** The prepare step casts to `torch.int32` before adding the zero points, and casts back to the original dtype before multiplying the scale. The convert step only casts at the very end.
```
# Prepare
x = x.to(torch.int32) - zero_point.to(torch.int32)
x = x.to(orig_dtype)
x = x * scale

# Convert
x = x - zero_point
x = x * scale 
x = x.to(orig_dtype)
```

This commit makes the convert path use the same torchao quantization primitives as the prepare path, thereby resolving the 3 above differences. Now, the prepare and convert steps match exactly in terms of numerics over many trials.

**Test Plan:**
```
python test/quantization/test_qat.py -k test_fake_quantize_per_token_vs_convert
```

### DIFF
--- a/torchao/_executorch_ops.py
+++ b/torchao/_executorch_ops.py
@@ -5,6 +5,8 @@
 # LICENSE file in the root directory of this source tree.
 import torch
 
+# TODO: delete these ops
+
 
 def _quantized_decomposed_quantize_per_channel_group_wrapper(*args, **kwargs):
     """
--- a/torchao/quantization/utils.py
+++ b/torchao/quantization/utils.py
@@ -539,36 +539,48 @@ def group_quantize_tensor_symmetric(
     return w_int8, scales, zeros
 
 
-def per_token_dynamic_quant(input: torch.Tensor) -> torch.Tensor:
-    orig_dtype = input.dtype
-    # TODO: we may need to make the choose_qparams op configurable
-    from torchao._executorch_ops import (
-        _quantized_decomposed_choose_qparams_per_token_asymmetric_wrapper,
-    )
-
-    (
-        scales,
-        zero_points,
-    ) = _quantized_decomposed_choose_qparams_per_token_asymmetric_wrapper(
-        input, torch.int8
-    )
-
-    # TODO: get these from torch.int8
+def per_token_dynamic_quant(
+    input: torch.Tensor,
+    scale_dtype: torch.dtype = torch.float32,
+    zero_point_dtype: torch.dtype = torch.float32,
+) -> torch.Tensor:
+    mapping_type = MappingType.ASYMMETRIC
+    block_size = _get_per_token_block_size(input)
     quant_min = -128
     quant_max = 127
-    from torchao._executorch_ops import _quantized_decomposed_quantize_per_token_wrapper
+    quant_dtype = torch.int8
+    output_dtype = input.dtype
 
-    input = _quantized_decomposed_quantize_per_token_wrapper(
-        input, scales, zero_points, quant_min, quant_max, torch.int8
+    scales, zero_points = choose_qparams_affine(
+        input,
+        mapping_type,
+        block_size,
+        quant_dtype,
+        quant_min,
+        quant_max,
+        scale_dtype=scale_dtype,
+        zero_point_dtype=zero_point_dtype,
     )
-    from torchao._executorch_ops import (
-        _quantized_decomposed_dequantize_per_token_wrapper,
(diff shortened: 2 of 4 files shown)


torchao/_executorch_ops.py
```diff
@@ -3,7 +3,207 @@
...
+# ============================
+
+
+quantized_decomposed_lib.define(
+    "quantize_per_token_aqt(Tensor input, Tensor scales, Tensor zero_points, "
+    "int quant_min, int quant_max, ScalarType dtype) -> Tensor"
+)
+
+
+@impl(quantized_decomposed_lib, "quantize_per_token_aqt", "CompositeExplicitAutograd")
```
jerryzh168, line 26: why adding new ops instead of reusing the quantize_affine and dequantize_affine ops?

andrewor14: It's because executorch expects these custom ops when lowering the quantized model. Seems like making `quantize_affine` a custom op is a bit overkill

jerryzh168: quantize_affine/dequantize_affine are already preserved as custom ops and supported by xnnpack lowering though

jerryzh168: see https://github.com/pytorch/ao/blob/d32afefdb6b8420cc31c9ed307497f4d305dd49c/test/integration/test_integration.py#L2039-L2041, does this work?

andrewor14: oh I see, yeah let me give that a try then. Does that mean executorch and XNNPACK need to understand all the arg types, like `MappingType` and `ZeroPointDomain`?

jerryzh168: these args are converted to strings, yeah they'll need to understand the strings

andrewor14: Ok updated, take another look?

torchao/quantization/utils.py
```diff
@@ -539,36 +539,50 @@ def group_quantize_tensor_symmetric(
...
     quant_min = -128
     quant_max = 127
-    from torchao._executorch_ops import _quantized_decomposed_quantize_per_token_wrapper
+    quant_dtype = torch.int8
+    output_dtype = input.dtype
 
-    input = _quantized_decomposed_quantize_per_token_wrapper(
-        input, scales, zero_points, quant_min, quant_max, torch.int8
+    scales, zero_points = (
+        _quantized_decomposed_choose_qparams_per_token_asymmetric_wrapper(
```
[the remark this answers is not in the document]
andrewor14: It's because QAT uses this, which is different from `choose_qparams_affine`: https://github.com/pytorch/ao/blob/09c276039e35d914be31484d39dc25ee3049e188/torchao/quantization/qat/fake_quantizer.py#L73

I actually tried making it call `choose_qparams_affine` but the math is slightly different, so the test ended up failing instead

jerryzh168 (approved):

jerryzh168:
Thanks for the updates, will be good to setup a deprecation plan for quantized_decomposed.choose_qparams op then

torchao/quantization/utils.py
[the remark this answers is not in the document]
andrewor14: Actually just got it to work with `choose_qparams_affine` with `MappingType.ASYMMETRIC`

PR #1968, opened 2025-03-26
Move Int8DynamicActivationIntxWeightConfig out of experimental

metascroy:
This PR moves Int8DynamicActivationIntxWeightConfig and its quantizer into torchao.quantization.quant_api (out of experimental).  Int8DynamicActivationIntxWeightConfig is refactored to closely mirror Int8DynamicActivationInt4WeightConfig when weight_dtype=torch.int4, layout=QDQLayout().

Quantization in Int8DynamicActivationIntxWeightConfig is done with QDQLayout, and then packing is done separately with make_packed_linear_int8_dynamic_activation_intx_weight_tensor.  This is to separate the quantization algorithm from the storage.  Both the packed and QDQLayout quantize with the same algorithm, and this is made explicit.

Example API usage:
```
quantize_(
      model,
      Int8DynamicActivationIntxWeightConfig(
          weight_dtype=torch.int4,
          weight_granularity=PerGroup(32),
          weight_mapping_type=MappingType.ASYMMETRIC,
          weight_zero_point_domain=ZeroPointDomain.NONE,
          layout=QDQLayout(),
      ),
  )
```


torchao/quantization/quant_api.py
```diff
@@ -664,6 +672,164 @@ def _int8_dynamic_activation_int4_weight_transform(
     return module
 
 
+@dataclass
+class Int8DynamicActivationIntxWeightConfig(AOBaseConfig):
+    """
```
metascroy, line 677: @andrewor14  can you have a look at this comment if there are any issues with it working well with QAT workflow with FakeQuantizeConfig.

torchao/quantization/quant_api.py
```diff
@@ -664,6 +672,164 @@ def _int8_dynamic_activation_int4_weight_transform(
...
+        act_mapping_type: The type of mapping to use for the activation quantization.
+            Must be one of MappingType.ASYMMETRIC or MappingType.SYMMETRIC.
+        layout: The layout to use for the packed weight tensor:
+            - PackedLinearInt8DynamicActivationIntxWeightLayout: this layout is optimized for CPU performance.
+            - QDQLayout: this layout represents the quantization with Q/DQ quant primitives, and is intended for
+                export applications like ExecuTorch.
+    """
+
+    weight_dtype: torch.dtype = torch.int8
+    weight_granularity: Union[PerRow, PerGroup] = PerGroup(32)
```
andrewor14, line 701: I feel we should just make the type here `Granularity` and throw an error for unsupported types, so we don't tie ourselves to specific granularity in the signature itself

torchao/quantization/quant_api.py
```diff
@@ -664,6 +672,164 @@ def _int8_dynamic_activation_int4_weight_transform(
     return module
 
 
+@dataclass
+class Int8DynamicActivationIntxWeightConfig(AOBaseConfig):
```
andrewor14, line 676: Does this match the numerics of `Int8DynamicActivationInt4WeightConfig` exactly if we choose weight_dtype = `torch.int4`? Is that a goal of this?

torchao/quantization/quant_api.py
```diff
@@ -664,6 +672,164 @@ def _int8_dynamic_activation_int4_weight_transform(
...
+            - PackedLinearInt8DynamicActivationIntxWeightLayout: this layout is optimized for CPU performance.
+            - QDQLayout: this layout represents the quantization with Q/DQ quant primitives, and is intended for
+                export applications like ExecuTorch.
+    """
+
+    weight_dtype: torch.dtype = torch.int8
+    weight_granularity: Union[PerRow, PerGroup] = PerGroup(32)
+    weight_zero_point_domain: ZeroPointDomain = ZeroPointDomain.NONE
+    weight_mapping_type: MappingType = MappingType.SYMMETRIC
+    weight_scale_dtype: Optional[torch.dtype] = None
```
andrewor14, line 704: I notice there's no `weight_zero_point_dtype` here. Is this assuming weight will always be symmetric? FWIW in the corresponding QAT `FakeQuantizeConfig` we do have `zero_point_precision` as well as `scale_precision`

torchao/quantization/quant_api.py
```diff
@@ -664,6 +672,164 @@ def _int8_dynamic_activation_int4_weight_transform(
...
+    act_mapping_type: MappingType = MappingType.ASYMMETRIC
+    layout: Layout = PackedLinearInt8DynamicActivationIntxWeightLayout(
```
andrewor14, line 706: What happens if we use this layout for cuda or other non-CPU backends? Are the numerics the same / is it still optimized? Does this also work with PlainLayout? Just wondering if this is the right default

torchao/quantization/quant_api.py
andrewor14: Whether or not this is the goal maybe we should document this somewhere, either here or in `Int8DynamicActivationInt4WeightConfig`'s docstring, because users may be confused about this

torchao/quantization/quant_api.py
```diff
@@ -664,6 +672,164 @@ def _int8_dynamic_activation_int4_weight_transform(
...
+        target_dtype=torch.int8,
+        quant_min=quant_min,
+        quant_max=quant_max,
+        eps=torch.finfo(torch.float32).eps,
+        scale_dtype=weight_scale_dtype,
+        zero_point_dtype=torch.int8 if has_weight_zeros else None,
+        preserve_zero=has_weight_zeros
+        or (weight_mapping_type == MappingType.SYMMETRIC),
+        zero_point_domain=weight_zero_point_domain,
+        _layout=QDQLayout(),
```
andrewor14, line 790: Do we just ignore `config.layout` here? Or does that refer to activation layout only?

torchao/quantization/quant_api.py
metascroy: Asymmetric is supported.  weight_zero_point_dtype is set to torch.int8 if weight_zero_point_domain=ZeroPointDomain.INT, else it is None if weight_zero_point_domain=ZeroPointDomain.NONE.

torchao/quantization/quant_api.py
metascroy: It works with QDQLayout, which subclasses PlainLayout(), but explicitly defines the linear impl.

Today this is done via a fallback path with PlainLayout() that @jerryzh168 mentioned might be removed.

We could make QDQLayout the default.  PackedLinearInt8DynamicActivationIntxWeightLayout only works on CPU.

torchao/quantization/quant_api.py
metascroy: See comment on line 775.  The QDQLayout is used for quantization algorithm.

The packing for layout PackedLinearInt8DynamicActivationIntxWeightLayout is handled on block at 804.

torchao/quantization/quant_api.py
metascroy: It does match numerics exactly when weight_dtype = torch.int4

metascroy:
@andrewor14 @jerryzh168 any more concerns on this PR?

torchao/dtypes/uintx/packed_linear_int8_dynamic_activation_intx_weight_layout.py
```diff
@@ -311,139 +334,112 @@ def _impl_2d_aten(input_tensor, weight_tensor):
...
-    MappingType,
-    choose_qparams_affine,
-    choose_qparams_and_quantize_affine_hqq,
-    quantize_affine,
 )
+from torchao.dtypes.utils import AQTTensorImpl, Layout
 
 
-class _AffineQuantizedTensor(AffineQuantizedTensor):
+def make_packed_linear_int8_dynamic_activation_intx_weight_tensor(
```
jerryzh168, line 343: would it make sense to have a separate tensor subclass for this layout and it can inherit from AQT?

torchao/experimental/tests/test_quant_passes.py
```diff
@@ -9,48 +9,68 @@
...
+        layer_to_weight_granularity = {}
         for weight_dtype in [getattr(torch, f"int{i}") for i in range(1, 9)]:
-            for has_weight_zeros in [True, False]:
-                for has_bias in [True, False]:
-                    idx = len(layers)
-                    layer_to_weight_dtype[idx] = weight_dtype
-                    layer_to_has_weight_zeros[idx] = has_weight_zeros
-                    layers.append(torch.nn.Linear(64, 64, bias=has_bias))
-        activations = torch.randn(2, 1, 64, dtype=torch.float32)
+            for weight_mapping_type in [MappingType.ASYMMETRIC, MappingType.SYMMETRIC]:
```
jerryzh168, line 33: you can use https://github.com/pytorch/ao/blob/9516764a97147231c72377bc1067c5e997de02f5/torchao/prototype/float8nocompile/benchmark/benchmark.py#L74 to reduce indentation

torchao/experimental/quant_api.py
```diff
@@ -648,6 +648,10 @@ def __post_init__(self):
         )
 
 
+# For BC
+int8_dynamic_activation_intx_weight = Int8DynamicActivationIntxWeightConfig
```
jerryzh168, line 652: do we need BC for experimental APIs

jerryzh168 (approved):
LGTM

torchao/dtypes/uintx/packed_linear_int8_dynamic_activation_intx_weight_layout.py
jerryzh168: > Perhaps that would be OK, but I'm wondering how that intersects with the quantize_ config (which has layout).

you can continue to use layout as an abstraction if you feel that is useful

> Note that this layout does not override anything in AQT, so it doesn't need to be a separate class.

OK then seems to be OK to keep using AQT

> I just wanted a way for users to construct this AQT from plain data, which is kind of useful if people want to use them outside of the quantize_ API. There is no easy, packaged up way for users to do that with AQT right now.

it seems that using default AQT constructor is fine for now? or do you feel we should add another constructor function?

torchao/experimental/quant_api.py
```diff
@@ -948,11 +808,13 @@ def quantize(
         # Quantize unembeddings
         quantize_(
             model,
-            Int8DynamicActivationIntxWeightConfig(
+            Int8DynamicActivationIntxWeightConfig_NonExperimental(
```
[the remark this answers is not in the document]
metascroy: This is the Int8DynamicActivationIntxWeightConfig in torchao.quantization.quant_api (the one we just added in this PR).

This is to avoid name conflict with torchao.experimental.quant_api.Int8DynamicActivationIntxWeightConfig, which I'm keeping around with exception that tells people to use new API.  This is just in case someone in torchchat or ET updates the ao pin before I get a chance, and they're wondering why things are broken.

Eventually I'll move it after I do pin bumps in those projects.

torchao/experimental/quant_api.py
metascroy: This is just in case someone in torchchat or ET updates the ao pin before I get a chance, and they're wondering why things are broken.

Eventually I'll move it after I do pin bumps in those projects.

torchao/dtypes/uintx/packed_linear_int8_dynamic_activation_intx_weight_layout.py
metascroy: > or do you feel we should add another constructor function

let's wait and see if others want that functionality.  The current way to construct an AQT from plain would be something like:

```
layout = PackedLinearInt8DynamicActivationIntxWeightLayout(target=target)
tensor_impl = PackedLinearInt8DynamicActivationIntxWeightAQTTensorImpl.from_plain(
        int_data, scale, zero_point, layout, bias
)
aqt = AffineQuantizedTensor(
    tensor_impl,
    block_size=(1, group_size),
    shape=int_data.shape,
    quant_min=qmin,
    quant_max=qmax,
    zero_point_domain=ZeroPointDomain.INT
    if has_weight_zeros
    else ZeroPointDomain.NONE,
)
```

which is isn't the most intuitive IMO.

jerryzh168: this actually looks reasonable to me, similar to torch.Tensor where people can construct a TensorImpl and pass it to Tensor constructor

jerryzh168:
also labeling this as a new feature, might be helpful to write down how to use the API in summary so we can copy paste for release notes

andrewor14:
Looks great! @metascroy can you add a unit test in a separate PR showing this new config matches Int8DynamicActivationInt4WeightConfig numerics exactly if `weight_dtype = torch.int4`?

PR #1969, opened 2025-03-26
Initial prototype of differentiable _scaled_grouped_mm function

danielvegamyhre:
## Summary

The `_grouped_scaled_mm` function in torchao will do:
- Dynamic, float8 rowwise quantization on inputs
- Use these float8 inputs with the grouped scaled mm kernel in pytorch core and return the result
- Do this in a differentiable way

Note this prototype only handles A=2D, B=3D. 

## Test plan
- Added unit tests verifying the correctness of the forward pass (outputs) and backward pass (gradients)
- Verified torch.compile works with no graph breaks

## Example usage

```python
from torchao.prototype.scaled_grouped_mm import _scaled_grouped_mm

...

out = _scaled_grouped_mm(
    x,             # 2D high precision input tensor
    params,        # 3D high precision weights
    offs=offs,     # 1D int32 group offsets
    out_dtype=out_dtype,
)
```

### DIFF
--- a/torchao/float8/float8_tensor.py
+++ b/torchao/float8/float8_tensor.py
@@ -101,19 +101,19 @@ def choose_scaled_mm_config(
     b_linear_mm_config: LinearMMConfig,
 ):
     if a_role is GemmInputRole.INPUT and b_role is GemmInputRole.WEIGHT:
-        assert (
-            a_linear_mm_config.output == b_linear_mm_config.output
-        ), f"linear_mm_config.output mismatch: {a_linear_mm_config.output} vs {b_linear_mm_config.output}"
+        assert a_linear_mm_config.output == b_linear_mm_config.output, (
+            f"linear_mm_config.output mismatch: {a_linear_mm_config.output} vs {b_linear_mm_config.output}"
+        )
         return a_linear_mm_config.output
     elif a_role is GemmInputRole.GRAD_OUTPUT and b_role is GemmInputRole.WEIGHT:
-        assert (
-            a_linear_mm_config.grad_input == b_linear_mm_config.grad_input
-        ), f"linear_mm_config.grad_input mismatch: {a_linear_mm_config.grad_input} vs {b_linear_mm_config.grad_input}"
+        assert a_linear_mm_config.grad_input == b_linear_mm_config.grad_input, (
+            f"linear_mm_config.grad_input mismatch: {a_linear_mm_config.grad_input} vs {b_linear_mm_config.grad_input}"
+        )
         return a_linear_mm_config.grad_input
     elif a_role is GemmInputRole.GRAD_OUTPUT and b_role is GemmInputRole.INPUT:
-        assert (
-            a_linear_mm_config.grad_weight == b_linear_mm_config.grad_weight
-        ), f"linear_mm_config.grad_weight mismatch: {a_linear_mm_config.grad_weight} vs {b_linear_mm_config.grad_weight}"
+        assert a_linear_mm_config.grad_weight == b_linear_mm_config.grad_weight, (
+            f"linear_mm_config.grad_weight mismatch: {a_linear_mm_config.grad_weight} vs {b_linear_mm_config.grad_weight}"
+        )
         return a_linear_mm_config.grad_weight
     else:
         raise AssertionError(f"unexpected a_role {a_role} and b_role {b_role}")
@@ -154,9 +154,9 @@ class _ToFloat8ConstrFunc(torch.autograd.Function):
         bits_fp8 = to_fp8_saturated(tensor_scaled, float8_dtype)
 
         if isinstance(bits_fp8, DTensor):
-            assert isinstance(
-                scale, DTensor
-            ), "Expected Float8 scale to be a DTensor if bits_fp8 is a DTensor"
+            assert isinstance(scale, DTensor), (
+                "Expected Float8 scale to be a DTensor if bits_fp8 is a DTensor"
+            )
             bits_mesh = bits_fp8.device_mesh
             bits_placements = bits_fp8.placements
             local_bits = bits_fp8.to_local()
--- a/torchao/float8/float8_tensor.py
+++ b/torchao/float8/float8_tensor.py
@@ -317,7 +317,6 @@ class Float8Tensor(torch.Tensor):
             None,
             0,
             -1,
-            -2,
         ), f"unsupported axiswise_dim {axiswise_dim}"
         self._axiswise_dim = axiswise_dim
 
--- a/torchao/float8/float8_tensor.py
+++ b/torchao/float8/float8_tensor.py
@@ -101,19 +101,19 @@ def choose_scaled_mm_config(
     b_linear_mm_config: LinearMMConfig,
 ):
     if a_role is GemmInputRole.INPUT and b_role is GemmInputRole.WEIGHT:
-        assert a_linear_mm_config.output == b_linear_mm_config.output, (
-            f"linear_mm_config.output mismatch: {a_linear_mm_config.output} vs {b_linear_mm_config.output}"
-        )
+        assert (
+            a_linear_mm_config.output == b_linear_mm_config.output
+        ), f"linear_mm_config.output mismatch: {a_linear_mm_config.output} vs {b_linear_mm_config.output}"
         return a_linear_mm_config.output
     elif a_role is GemmInputRole.GRAD_OUTPUT and b_role is GemmInputRole.WEIGHT:
-        assert a_linear_mm_config.grad_input == b_linear_mm_config.grad_input, (
-            f"linear_mm_config.grad_input mismatch: {a_linear_mm_config.grad_input} vs {b_linear_mm_config.grad_input}"
-        )
+        assert (
+            a_linear_mm_config.grad_input == b_linear_mm_config.grad_input
+        ), f"linear_mm_config.grad_input mismatch: {a_linear_mm_config.grad_input} vs {b_linear_mm_config.grad_input}"
         return a_linear_mm_config.grad_input
     elif a_role is GemmInputRole.GRAD_OUTPUT and b_role is GemmInputRole.INPUT:
-        assert a_linear_mm_config.grad_weight == b_linear_mm_config.grad_weight, (
-            f"linear_mm_config.grad_weight mismatch: {a_linear_mm_config.grad_weight} vs {b_linear_mm_config.grad_weight}"
-        )
+        assert (
+            a_linear_mm_config.grad_weight == b_linear_mm_config.grad_weight
+        ), f"linear_mm_config.grad_weight mismatch: {a_linear_mm_config.grad_weight} vs {b_linear_mm_config.grad_weight}"
         return a_linear_mm_config.grad_weight
     else:
         raise AssertionError(f"unexpected a_role {a_role} and b_role {b_role}")
@@ -154,9 +154,9 @@ class _ToFloat8ConstrFunc(torch.autograd.Function):
         bits_fp8 = to_fp8_saturated(tensor_scaled, float8_dtype)
 
         if isinstance(bits_fp8, DTensor):
-            assert isinstance(scale, DTensor), (
-                "Expected Float8 scale to be a DTensor if bits_fp8 is a DTensor"
-            )
+            assert isinstance(
+                scale, DTensor
+            ), "Expected Float8 scale to be a DTensor if bits_fp8 is a DTensor"
             bits_mesh = bits_fp8.device_mesh
             bits_placements = bits_fp8.placements
             local_bits = bits_fp8.to_local()
@@ -313,11 +313,7 @@ class Float8Tensor(torch.Tensor):
             linear_mm_config if linear_mm_config is not None else LinearMMConfig()
         )
         self._gemm_input_role = gemm_input_role
-        assert axiswise_dim in (
-            None,
-            0,
-            -1,
-        ), f"unsupported axiswise_dim {axiswise_dim}"
+        assert axiswise_dim in (None, 0, -1), f"unsupported axiswise_dim {axiswise_dim}"
         self._axiswise_dim = axiswise_dim
 
         return self
--- a/torchao/float8/float8_tensor.py
+++ b/torchao/float8/float8_tensor.py
@@ -316,8 +316,8 @@ class Float8Tensor(torch.Tensor):
         assert axiswise_dim in (
             None,
             0,
-            1,
             -1,
+            -2,
         ), f"unsupported axiswise_dim {axiswise_dim}"
         self._axiswise_dim = axiswise_dim
 
--- a/torchao/float8/float8_tensor.py
+++ b/torchao/float8/float8_tensor.py
@@ -313,7 +313,7 @@ class Float8Tensor(torch.Tensor):
             linear_mm_config if linear_mm_config is not None else LinearMMConfig()
         )
         self._gemm_input_role = gemm_input_role
-        assert axiswise_dim in (None, 0, -1), f"unsupported axiswise_dim {axiswise_dim}"
+        assert axiswise_dim in (None, 0, 1, -1), f"unsupported axiswise_dim {axiswise_dim}"
         self._axiswise_dim = axiswise_dim
 
         return self
--- a/torchao/float8/float8_tensor.py
+++ b/torchao/float8/float8_tensor.py
@@ -313,7 +313,12 @@ class Float8Tensor(torch.Tensor):
             linear_mm_config if linear_mm_config is not None else LinearMMConfig()
         )
         self._gemm_input_role = gemm_input_role
-        assert axiswise_dim in (None, 0, 1, -1), f"unsupported axiswise_dim {axiswise_dim}"
+        assert axiswise_dim in (
+            None,
+            0,
+            1,
+            -1,
+        ), f"unsupported axiswise_dim {axiswise_dim}"
         self._axiswise_dim = axiswise_dim
 
         return self
--- a/torchao/prototype/grouped_mm/__init__.py
+++ b/torchao/prototype/grouped_mm/__init__.py
@@ -1,10 +1,13 @@
-
 from typing import Optional
 
 import torch
+
 from torchao import float8
-from torchao.float8.float8_scaling_utils import hp_tensor_to_float8_dynamic, get_maybe_axiswise_dim
 from torchao.float8.config import Float8LinearConfig, Float8LinearRecipeName
+from torchao.float8.float8_scaling_utils import (
+    get_maybe_axiswise_dim,
+    hp_tensor_to_float8_dynamic,
+)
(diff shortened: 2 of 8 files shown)


torchao/float8/float8_tensor.py
```diff
@@ -313,7 +313,7 @@ def __new__(
             linear_mm_config if linear_mm_config is not None else LinearMMConfig()
         )
         self._gemm_input_role = gemm_input_role
-        assert axiswise_dim in (None, 0, -1), f"unsupported axiswise_dim {axiswise_dim}"
+        assert axiswise_dim in (None, 0, 1, -1), f"unsupported axiswise_dim {axiswise_dim}"
```
danielvegamyhre, line 316: Note for reviewer: I had to do this to handle the 3D B tensor, which the user will pass in transposed (e.g. `_scaled_grouped_mm(A, B.transpose(-2, -1), ...)`). The rowwise scales for a 3D transposed tensor need to be along dim 1, which was not supported. 

Example:

B.shape = (B,N,K)
B.transpose(-2,-1).shape = (B,K,N)
B_scales need to be (B,1,N) and then be squeezed into (B,N) to prepare for torch._scaled_grouped_mm input

torchao/float8/float8_tensor.py
```diff
@@ -313,7 +313,12 @@ def __new__(
             linear_mm_config if linear_mm_config is not None else LinearMMConfig()
         )
         self._gemm_input_role = gemm_input_role
-        assert axiswise_dim in (None, 0, -1), f"unsupported axiswise_dim {axiswise_dim}"
+        assert axiswise_dim in (
+            None,
+            0,
+            -1,
+            -2,
```
danielvegamyhre, line 320: Need to allow -2 so we can do rowwise scaling of the 2D subtensors embedded in 3D tensors.

e.g.

- A = (M,K) and B = (B,K,N)
- A_scale should be (M,)
- B_scale should be (B,N) <-- because we computed rowwise scales for the (K,N) subtensors

danielvegamyhre:
cc @tianyu-l @ngimel please feel free to share any thoughts or feedback on this

torchao/prototype/grouped_mm/__init__.py
```diff
@@ -0,0 +1,370 @@
...
+    float8_recipe: Float8LinearRecipeName,
+    offs: torch.Tensor,
+    out_dtype: Optional[torch.dtype] = None,
+) -> torch.Tensor:
+    """
+    This function performs dynamic float8 quantization on the input tensors A and B using the given recipe,
+    then performs a scaled grouped GEMM and returns the results.
+
+    Args:
+        A (torch.Tensor): The first input tensor, which must be a 2D tensor of shape (M * num_groups, K).
```
vkuzo, line 25: clarify dtype of A, B, offs?

torchao/prototype/grouped_mm/__init__.py
```diff
@@ -0,0 +1,370 @@
...
+
+from torchao.float8.config import (
+    Float8LinearConfig,
+    Float8LinearRecipeName,
+    ScalingGranularity,
+)
+from torchao.float8.float8_utils import tensor_to_scale, to_fp8_saturated
+
+
+def _grouped_scaled_mm(
```
vkuzo, line 13: nit: can we move this out of `__init__.py` into its own file?

torchao/prototype/grouped_mm/__init__.py
```diff
@@ -0,0 +1,370 @@
+from typing import Optional, Tuple
```
vkuzo, line 1: seems like folder name should be `grouped_scaled_mm` instead of `grouped_mm`?  Or do we plan to add a `grouped_mm` here too?

torchao/prototype/grouped_mm/__init__.py
```diff
@@ -0,0 +1,370 @@
...
+    Float8LinearRecipeName,
+    ScalingGranularity,
+)
+from torchao.float8.float8_utils import tensor_to_scale, to_fp8_saturated
+
+
+def _grouped_scaled_mm(
+    A: torch.Tensor,
+    B: torch.Tensor,
+    float8_recipe: Float8LinearRecipeName,
```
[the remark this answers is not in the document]
danielvegamyhre: Makes sense, done.

torchao/prototype/grouped_mm/__init__.py
danielvegamyhre: Done

torchao/prototype/grouped_mm/__init__.py
danielvegamyhre: Updated folder name

torchao/prototype/grouped_mm/__init__.py
danielvegamyhre: Done

torchao/prototype/scaled_grouped_mm/scaled_grouped_mm.py
```diff
@@ -0,0 +1,417 @@
...
+    """
+    Args:
+        offs: (int32 torch.Tensor): list of group offsets
+
+    Returns:
+        list of integers representing the group sizes
+    """
+    group_sizes = []
+    start_idx = 0
+    for end_idx in offs.tolist():
```
ngimel, line 401: this is causing device-host synchronization, and we should avoid it, relying on upstream ops to create suitable inputs, or in worst case, on _scaled_grouped_mm implementation itself to throw an assert

torchao/prototype/scaled_grouped_mm/scaled_grouped_mm.py
```diff
@@ -0,0 +1,417 @@
...
+
+        # Fetch float8 config from specified recipe name.
+        float8_config = Float8LinearConfig.from_recipe_name(
+            Float8LinearRecipeName.ROWWISE
+        )
+
+        # Store what we need for backward.
+        ctx.save_for_backward(A, B)
+        ctx.float8_config = float8_config
+        ctx.offs = offs
```
ngimel, line 95: `offs` is also a tensor so better you you use `save_for_backward`

danielvegamyhre: Done

torchao/prototype/scaled_grouped_mm/scaled_grouped_mm.py
```diff
@@ -0,0 +1,415 @@
...
+        )
+        grad_output_scaled = grad_output.to(torch.float32) * grad_output_scales
+        grad_output_fp8_row_major = to_fp8_saturated(
+            grad_output_scaled, float8_config.cast_config_grad_output.target_dtype
+        )
+
+        # Convert B to non-transposed, float8, column-major for right operand of grouped GEMM
+        # needed for grad_A: grad_output @ B.
+        # Since B was transposed before entry to forward, we need to transpose it back here for this.
+        B_non_transposed_col_major = B.contiguous().transpose(-2, -1)
```
ngimel, line 166: It might be better to scale and transpose B in forward, and store only quantized version (to minimize memory)

torchao/prototype/scaled_grouped_mm/scaled_grouped_mm.py
```diff
@@ -0,0 +1,415 @@
...
+    A_fp8_col_major = torch.empty_like(A_col_major, dtype=target_dtype)
+    A_scales = torch.empty(
+        A_fp8_col_major.size(1) * num_groups,
+        dtype=torch.float32,
+        device=A_fp8_col_major.device,
+    )
+
+    start_idx = 0
+    next_scale_idx = 0
+    for end_idx in offs.tolist():
```
ngimel, line 279: here also would be better to have a triton kernel computing scales that could read offs on the device, to avoid syncs

torchao/prototype/scaled_grouped_mm/scaled_grouped_mm.py
danielvegamyhre: Good catch, removed this assertion for now and will rely on kernel side assert, to avoid device-host sync

torchao/prototype/scaled_grouped_mm/scaled_grouped_mm.py
```diff
@@ -0,0 +1,417 @@
...
+        # A shape: (M, K)
+        # A_scales shape: (M,1)
+        A_scales = tensor_to_scale(
+            A,
+            float8_config.cast_config_input.target_dtype,
+            scaling_granularity=float8_config.cast_config_input.scaling_granularity,
+            axiswise_dim=-1,
+            round_scales_to_power_of_2=float8_config.round_scales_to_power_of_2,
+        )
+        A_scaled = A.to(torch.float32) * A_scales
```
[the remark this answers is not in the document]
danielvegamyhre: That's how we do the conversion, multiply the high precision tensor by the scales, then do the clamping and dtype conversion in `to_fp8_saturated` on the line below. You can see the analogous production code [here](https://github.com/pytorch/ao/blob/5e4d50ca1f1ea8564b372e5a28536834366bf09b/torchao/float8/float8_tensor.py#L153-L154).

Or are you asking why we multiply instead of divide, it has to do with how we calculate the scale [here](https://github.com/pytorch/ao/blob/5e4d50ca1f1ea8564b372e5a28536834366bf09b/torchao/float8/float8_utils.py#L45-L48). The scale is basically the maximum value representable in the float8 dtype, divided by the maximum absolute value in the target part of the tensor (i.e., scale = 448 / amax). 

So if amax is larger than 448, multiplying will shrink the tensor values to fit in the dynamic range of the float8 dtype. If amax is smaller than 448, it will grow the values to fill the dynamic range of the float8 dtype.

torchao/prototype/scaled_grouped_mm/scaled_grouped_mm.py
danielvegamyhre: Yeah this implementation is not performant, my first goal was to get accurate numerics. As a follow up I can write a triton kernel to avoid this device-host sync and for loop.

torchao/prototype/scaled_grouped_mm/scaled_grouped_mm.py
danielvegamyhre: good idea, done. hopefully torch.compile can do some fusion here and read B once and write both outputs simultaneously (float transposed column major, float8 non-transposed column major).

torchao/prototype/scaled_grouped_mm/scaled_grouped_mm.py
```diff
@@ -0,0 +1,403 @@
...
+        ), f"shape {A.shape} and {B.shape} are not compatible for _scaled_grouped_mm"
+
+        # Due to hardware requirements, the right operand in a scaled grouped GEMM must be column-major.
+        if not _is_column_major(B):
+            B_col_major = B.transpose(-2, -1).contiguous().transpose(-2, -1)
+        else:
+            B_col_major = B
+
+        # Fetch float8 config from specified recipe name.
+        float8_config = Float8LinearConfig.from_recipe_name(
```
vkuzo, line 83: it's suprising to see a config created here, why not just inline the logic you need without worrying about configs?  IMO dealing with configs would be for when this API is about to be productionized.

danielvegamyhre: I agree, I used it here for this prototype because in the test code i need to use `matmul_with_hp_or_float8_args` (which requires a Float8 config) to compute the reference forward/backward - and trying to get the outputs/grads to match was already pretty tricky, so to start I wanted to just reference the same `Float8LinearConfig.from_recipe_name(Float8LinearRecipeName.ROWWISE)` everywhere, to minimize room for accidental differences in how a particular tensor is quantized in test code vs implementation, etc.

I am happy to refactor to inline everything in the implementation, but for the test code we still need the float8 config.

torchao/prototype/scaled_grouped_mm/scaled_grouped_mm.py
```diff
@@ -0,0 +1,371 @@
...
+    """
+    This function performs dynamic float8 quantization with row-wise scaling
+    on the input tensors A and B using the given recipe, then performs a
+    scaled grouped GEMM and returns the results.
+
+    Args:
+        A (bf16/float32 torch.Tensor): The first high-precision input tensor, which must be a 2D tensor of shape (M * num_groups, K).
+        B (bf16/float32 torch.Tensor): The second high-precision input tensor which must be 3D, which must be shape (B, K, N).
+        offs (int32 torch.Tensor): The offsets to use to mark the starting index of each group in the input tensor of shape.
+        float8_recipe (Float8LinearRecipeName): The recipe to use for dynamic float8 quantization.
```
vkuzo, line 24: remove?

torchao/prototype/scaled_grouped_mm/scaled_grouped_mm.py
```diff
@@ -0,0 +1,371 @@
...
+) -> torch.Tensor:
+    """
+    This function performs dynamic float8 quantization with row-wise scaling
+    on the input tensors A and B using the given recipe, then performs a
+    scaled grouped GEMM and returns the results.
+
+    Args:
+        A (bf16/float32 torch.Tensor): The first high-precision input tensor, which must be a 2D tensor of shape (M * num_groups, K).
+        B (bf16/float32 torch.Tensor): The second high-precision input tensor which must be 3D, which must be shape (B, K, N).
+        offs (int32 torch.Tensor): The offsets to use to mark the starting index of each group in the input tensor of shape.
```
vkuzo, line 23: can we clarify if this is for A, B or both? A, right?

torchao/prototype/scaled_grouped_mm/scaled_grouped_mm.py
```diff
@@ -0,0 +1,371 @@
...
+        float8_recipe (Float8LinearRecipeName): The recipe to use for dynamic float8 quantization.
+        out_dtype (Optional[torch.dtype]): The dtype of the output tensor. Currently only torch.bfloat16 is supported.
+        use_fast_accum (bool): Whether to use fast accumulation or not. Default is False.
```
vkuzo, line 26: remove?

torchao/prototype/scaled_grouped_mm/scaled_grouped_mm.py
```diff
@@ -0,0 +1,371 @@
...
+        ), "B must be float32 or bfloat16"
+        assert offs.dtype == torch.int32, "offs must be int32"
+
+        # Assert A and B dims are compatible for a scaled grouped GEMM.
+        assert A.size(-1) == B.size(
+            -2
+        ), f"shape {A.shape} and {B.shape} are not compatible for _scaled_grouped_mm"
+
+        # Due to hardware requirements, the right operand in a scaled grouped GEMM must be column-major.
+        if not _is_column_major(B):
```
vkuzo, line 73: nit: I'd prefer letting the caller do this instead, and this function can just assert that the layout is what is needed for the kernel

torchao/prototype/scaled_grouped_mm/scaled_grouped_mm.py
```diff
@@ -0,0 +1,371 @@
...
+            round_scales_to_power_of_2=True,
+        )
+        B_scaled = B.to(torch.float32) * B_scales
+        B_fp8_col_major = to_fp8_saturated(B_scaled, torch.float8_e4m3fn)
+
+        # Precompute B_non_transposed_col_major for backward, to save memory by storing the
+        # low precision B tensor instead of the high precision B tensor.
+        # In the backward this is needed for grad_A: grad_output @ B.
+        # Since B was transposed before entry to forward, we need to transpose it back here for this.
+        B_non_transposed_col_major = B.contiguous().transpose(-2, -1)
```
vkuzo, line 109: this naming is confusing, if the original variable is `B`, then this looks like `B_transposed`.  IMO it would be cleanest to do something like:
* input is B
* B transposed is B_t

or 
* input is B_t
* B_t transposed is B

torchao/prototype/scaled_grouped_mm/scaled_grouped_mm.py
```diff
@@ -0,0 +1,371 @@
...
+    """
+    This function checks if the input tensor is column-major.
+
+    Args:
+        x (torch.Tensor): The input tensor to be checked.
+
+    Returns:
+        A boolean indicating whether the input tensor is column-major.
+    """
+    return x.stride(-2) == 1 and x.stride(-1) > 1
```
vkuzo, line 371: does this work for 4d/5d/etc tensors? if not, maybe assert that rank is 3?

torchao/prototype/scaled_grouped_mm/scaled_grouped_mm.py
danielvegamyhre: Yeah I agree, I considered this as well. The problem is I'm trying to make the naming consistent with torch._scaled_grouped_mm, which calls the tensors A and B, but it checks that  B [must be transposed](https://github.com/pytorch/pytorch/blob/d25acac357ff8663a7787e57e6bc5e69987a8f9a/aten/src/ATen/native/cuda/Blas.cpp#L1497) - although really what it's enforcing is column major format, so it's a bit confusing. 

For now I changed the naming to option 2 above, which will make the python code here clearer, with the trade-off being it will no longer be consistent with the kernel naming. I think that's fine though, I doubt too many people will be diving into the kernel code.

torchao/prototype/scaled_grouped_mm/scaled_grouped_mm.py
```diff
@@ -0,0 +1,371 @@
...
+        A: torch.Tensor,
+        B: torch.Tensor,
+        offs: torch.Tensor,
+        out_dtype: Optional[torch.dtype] = None,
+    ) -> torch.Tensor:
+        # torchao _scaled_grouped_mm only supports A=2D, B=3D.
+        assert A.ndim == 2, "A must be 2D"
+        assert B.ndim == 3, "B must be 3D"
+
+        assert (
```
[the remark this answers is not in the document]
danielvegamyhre: Some are specific to this implementation (only support 2d A and 3D B for now, the primary use case), but some are redundant, yes. I found the device-side assertions to be opaque sometimes, I often had to read the kernel code to figure out why it wasn't working, to see the exact condition that was failing if the error was a bit ambiguous. So my goal here was to make the requirements more transparent and easier to debug. 

(I think since scaled grouped mm is a kernel executing on the GPU, if a check fails, then on the CPU side the error message we can't see the actual line of code with the conditoin that failed, it just points to the entrypoint `torch._scaled_grouped_mm`)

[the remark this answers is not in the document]
ngimel: ndim assertions are not device side, so they have correct stack trace. Device-side assertions are put in when it's impossible to check the same thing on the host without introducing host-device sync, so they happen for a reason and unfortunately can't be improved. To get correct stack trace run with CUDA_LAUNCH_BLOCKING=1

PR #1979, opened 2025-03-28
add PyTorch MG style group gemm reference utils

lessw2020:
This PR adds a single reference file for computing forward and backward (dx, dw) results for MG style group gemms.
It uses PyTorch directly via manual looping and PyTorch Autograd to compute reference values to be used in verifying other (Triton, torch.compile) group Gemm results. 

The idea here is for both fp8 and bf16 group gemms to use the same reference generation to ensure consistent unit testing.


danielvegamyhre:
LGTM, will leave approval for @vkuzo

PR #1982, opened 2025-03-31
[sparsity] Add PartialLinear module for structured sparsity

lakshminarasimmanv:
This PR adds the PartialLinear module to `torchao.prototype.sparsity` as suggested by @jcaip in pytorch/pytorch#149853.

## Implementation Details
- PartialLinear is a linear layer that implements structured sparsity by connecting each output neuron to only the top-k input features by weight magnitude
- Supports dynamic connectivity updates during training with configurable update frequency
- Provides both masked-dense computation and fully sparse computation paths

The implementation is placed in a PartialLinear folder as suggested. I'm open to any feedback or modifications needed to better integrate with the torchao ecosystem and sparsity framework.

This addresses the TODO in pytorch/pytorch's `linear.py` module that mentions "PartialLinear - maybe in sparse?".

Related issue: pytorch/pytorch#135091

cc @jcaip

### DIFF
--- a/torchao/prototype/sparsity/PartialLinear/partial_linear.py
+++ b/torchao/prototype/sparsity/PartialLinear/partial_linear.py
@@ -1,0 +1,181 @@
+import math
+import torch
+from torch import Tensor
+from torch.nn import functional as F, init
+from torch.nn.parameter import Parameter
+from torch.nn import Module
+
+class PartialLinear(Module):
+    r"""Applies a linear transformation where each output feature connects to only the top-k
+    input features by weight magnitude: :math:`y = x * (M \odot W^T) + b`,
+    where $\odot$ is the element-wise product and $M$ is a binary mask.
+    This module implements a form of structured sparsity that can reduce computation
+    and memory usage during inference.
+    Args:
+        in_features: size of each input sample
+        out_features: size of each output sample
+        top_k: number of weights to retain per output feature (default: in_features // 2)
+        bias: If set to ``False``, the layer will not learn an additive bias.
+            Default: ``True``
+        update_mask_every: update the mask every N forward passes during training (default: 50)
+    Shape:
+        - Input: :math:`(*, H_\text{in})` where :math:`*` means any number of
+          dimensions including none and :math:`H_\text{in} = \text{in\_features}`.
+        - Output: :math:`(*, H_\text{out})` where all but the last dimension
+          are the same shape as the input and :math:`H_\text{out} = \text{out\_features}`.
+    Attributes:
+        weight: the learnable weights of the module of shape
+            :math:`(\text{out\_features}, \text{in\_features})`. The values are
+            initialized from :math:`\mathcal{U}(-\sqrt{k}, \sqrt{k})`, where
+            :math:`k = \frac{1}{\text{in\_features}}`
+        mask: binary mask of shape :math:`(\text{out\_features}, \text{in\_features})`
+            indicating which weights are retained (1) or pruned (0)
+        bias:   the learnable bias of the module of shape :math:`(\text{out\_features})`.
+                If :attr:`bias` is ``True``, the values are initialized from
+                :math:`\mathcal{U}(-\sqrt{k}, \sqrt{k})` where
+                :math:`k = \frac{1}{\text{in\_features}}`
+        is_sparse_forward: when True, uses a completely sparse computation for forward pass
+    Examples::
+        >>> m = nn.PartialLinear(20, 30, top_k=5)  # Each output connected to top 5 inputs
+        >>> input = torch.randn(128, 20)
+        >>> output = m(input)
+        >>> print(output.size())
+        torch.Size([128, 30])
+    """
+    __constants__ = ["in_features", "out_features", "top_k", "update_mask_every", "is_sparse_forward"]
+    in_features: int
+    out_features: int
+    top_k: int
+    update_mask_every: int
+    is_sparse_forward: bool
+    weight: Tensor
+
+    def __init__(
+        self,
+        in_features: int,
+        out_features: int,
+        top_k: int = None,
+        bias: bool = True,
+        update_mask_every: int = 50,
+        is_sparse_forward: bool = False,
+        device=None,
+        dtype=None,
+    ) -> None:
+        factory_kwargs = {"device": device, "dtype": dtype}
+        super().__init__()
+
+        self.in_features = in_features
+        self.out_features = out_features
+
+        # Default to half of the input features if not specified
+        if top_k is None:
+            top_k = max(1, in_features // 2)
+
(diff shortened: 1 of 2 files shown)


test/prototype/test_partial_linear.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class TestPartialLinear(unittest.TestCase):
+    def test_basic_functionality(self):
+        in_features = 20
+        out_features = 10
+        top_k = 5
+        m = PartialLinear(in_features, out_features, top_k=top_k)
+        
+        self.assertEqual(m.in_features, in_features)
```
jcaip, line 20: let's put this in the sparsity subfolder in tests

test/prototype/test_partial_linear.py
```diff
@@ -0,0 +1,44 @@
...
+# LICENSE file in the root directory of this source tree.
+import unittest
+
+import torch
+
+from torchao.prototype.sparsity.PartialLinear.partial_linear import PartialLinear
+
+
+class TestPartialLinear(unittest.TestCase):
+    def test_basic_functionality(self):
```
jcaip, line 14: nit: better to do something like test_partial_linear

test/prototype/test_partial_linear.py
```diff
@@ -0,0 +1,44 @@
...
+        
+        x = torch.randn(32, in_features)
+        output = m(x)
+        self.assertEqual(output.shape, (32, out_features))
+        
+        x_single = torch.randn(in_features)
+        output_single = m(x_single)
+        self.assertEqual(output_single.shape, (out_features,))
+    
+    def test_invalid_input(self):
```
jcaip, line 35: `test_invalid_k` is a better test name

torchao/prototype/sparsity/PartialLinear/partial_linear.py
```diff
@@ -0,0 +1,183 @@
...
+        self.is_sparse_forward = is_sparse_forward
+        self._forward_counter = 0
+
+        # Create a full weight matrix
+        self.weight = Parameter(
+            torch.empty((out_features, in_features), **factory_kwargs)
+        )
+
+        # Create a binary mask for the weights
+        mask = torch.ones((out_features, in_features), **factory_kwargs)
```
jcaip, line 90: can we make mask `bool` or do we need differentiable masking for training?

torchao/prototype/sparsity/PartialLinear/partial_linear.py
```diff
@@ -0,0 +1,183 @@
...
+        dtype=None,
+    ) -> None:
+        factory_kwargs = {"device": device, "dtype": dtype}
+        super().__init__()
+
+        self.in_features = in_features
+        self.out_features = out_features
+
+        # Default to half of the input features if not specified
+        if top_k is None:
```
jcaip, line 73: I think better to just throw a value error or make this an arg instead of a kwarg

torchao/prototype/sparsity/PartialLinear/partial_linear.py
```diff
@@ -0,0 +1,183 @@
...
+            _, top_k_indices = weight_mag.topk(self.top_k, dim=1)
+
+            # Set mask to 1 for top-k weights for each output
+            for i in range(self.out_features):
+                new_mask[i, top_k_indices[i]] = 1.0
+
+            # Update the mask
+            self.mask.copy_(new_mask)
+
+    def _sparse_forward(self, input):
```
jcaip, line 133: Why do we need this / what is this doing? I don't understand when you would use this `is_sparse_forward` flag.

torchao/prototype/sparsity/PartialLinear/partial_linear.py
```diff
@@ -0,0 +1,183 @@
...
+        if self.is_sparse_forward:
+            return self._sparse_forward(input)
+
+        # Apply the mask to the weights
+        masked_weight = self.weight * self.mask
+
+        # Use the masked weights for the linear transformation
+        return F.linear(input, masked_weight, self.bias)
+
+    def extra_repr(self) -> str:
```
jcaip, line 179: Is this function used anywere?

jcaip (changes requested):
thanks for the PR :) 

Long term, I think our "training" functionality for sparsity in torchao needs some TLC. We currently have superblock and accelerated 2:4 sparse training, but these each have their own entrypoints, we should standardize under `sparsify_`. We can do something similar for this as well. 

Also it would be great to get memory benchmarks on a simple nn.Linear layer.

lakshminarasimmanv:
Thank you @jcaip for the feedback and guidance!

I'll take a look at the requested changes and address them.

Regarding standardizing under `sparsify_`: I agree this would be valuable for long-term integration. Would you like me to attempt integrating with the existing sparsity framework in this PR, or would you prefer to keep this as a standalone implementation for now and handle the standardization in a follow-up effort?

I'll also work on adding memory benchmarks comparing to nn.Linear.

I appreciate your guidance and direction on how best to align this with the torchao ecosystem.

Looking forward to your further thoughts!

jcaip:
cc @lakshminarasimmanv just curious here - are you working on this for fun or do you have a use-case in mind that needs this? If you have a specific use case I would be really interested in learning more. 

> Regarding standardizing under sparsify_: I agree this would be valuable for long-term integration. Would you like me to attempt integrating with the existing sparsity framework in this PR, or would you prefer to keep this as a standalone implementation for now and handle the standardization in a follow-up effort?

Let's keep this self contained for now, adding to sparsify_ can be done is a subsequent PR. If you're interested in doing that I'm happy to review and accept :)

Feel free to ping me when you're ready for another review or have any questions!

lakshminarasimmanv:
Thanks for asking about my motivation, @jcaip!

I've been working with structured sparsity approaches to optimize transformer models, where I noticed that attention heads often focus strongly on a subset of input features. This seemed like a natural use case for a partial connectivity approach that could reduce computation while preserving model quality.

When exploring the PyTorch codebase to understand implementation patterns, I came across the "TODO: PartialLinear" comment and the related issue, which aligned perfectly with what I was exploring.

Also BTW, I've updated the code to address your requested changes. I've also been running benchmarks comparing `Linear` and `PartialLinear` across different model sizes and will share those results in the next day or two.

I'm definitely interested in contributing actively to PyTorch and would be happy to maintain this implementation and help with the sparsity framework going forward.

Looking forward to your thoughts on the changes!

jcaip (commented):
Thanks @lakshminarasimmanv, a couple more comments

> I've been working with structured sparsity approaches to optimize transformer models, where I noticed that attention heads often focus strongly on a subset of input features. This seemed like a natural use case for a partial connectivity approach that could reduce computation while preserving model quality.

Sounds cool! do you have any data / experiments you're running? I would createa README.md and mention that there, to help others understand how and why to use your code.

PR #1988, opened 2025-04-01
Register codebook quant ops

jerryzh168:
Summary:
Register the codebook quant / dequant ops as custom ops so they can be recongnized after export

Test Plan:
python test/prototype/test_codebook_quant.py -k test_export

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/prototype/quantization/codebook/codebook_ops.py
+++ b/torchao/prototype/quantization/codebook/codebook_ops.py
@@ -182,7 +182,6 @@ def choose_qparams_codebook(
     Returns:
         torch.Tensor: The codebook tensor, shape (codebook_size, *block_size).
     """
-    breakpoint()
     if code_dtype == torch.int32:
         codebook_size = 2**16
     else:
--- a/torchao/prototype/quantization/codebook/codebook_ops.py
+++ b/torchao/prototype/quantization/codebook/codebook_ops.py
@@ -11,8 +11,13 @@ from torchao.quantization.quant_primitives import (
     _DTYPE_TO_QVALUE_BOUNDS,
     _SUB_BYTE_UINT_BOUNDS,
 )
+from torchao.utils import _register_custom_op
+
+quant_lib = torch.library.Library("quant", "FRAGMENT")
+register_custom_op = _register_custom_op(quant_lib)
 
 
+@register_custom_op
 def quantize_codebook(
     input: torch.Tensor,
     codebook: torch.Tensor,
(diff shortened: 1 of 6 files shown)


torchao/prototype/quantization/codebook/codebook_ops.py
```diff
@@ -90,20 +95,24 @@ def quantize_codebook(
...
     """
     Reconstructs the original tensor from codes and the codebook.
 
     Args:
-        codes (torch.Tensor): Indices of codebook entries for each block,
-                                          shape (d1//b1, d2//b2, ..., dN//bN).
+        codes (torch.Tensor): torch.int32 dtype, indices of codebook entries for each block,
+                              shape (d1//b1, d2//b2, ..., dN//bN).
         codebook (torch.Tensor): Codebook tensor used for quantization,
                                  shape (k, b1, b2, ..., bN) where b_i are block sizes.
```
metascroy, line 113: nit: say what k is

torchao/prototype/quantization/codebook/codebook_ops.py
```diff
@@ -90,20 +95,24 @@ def quantize_codebook(
     return codes.to(code_dtype)
 
 
+@register_custom_op
 def dequantize_codebook(
```
metascroy, line 99: IIUC, this does not look like it supports granularity, which we will want.

From what I can, k is the idx range, e.g., for 4-bit quantization, k = 16.  Each idx=i is mapped to a tensor codebook[i].  So we have 1 codebook for the tensor that maps indices to tensor.  

This seems a bit complicated to me.  For CoreML, the default is each idx maps to a scalar (but they also support mapping to a vector).  I'm not sure if anyone will need tensor-valued look up values.

But we do want granularity in the sense that we can have one codebook per channel, grouped channel, tensor, etc.

jerryzh168: I think the `scale_block_size` in `choose_qparams_codebook` or the shape of `scales` in the dequant op is supposed to allow us to control the  granularity, the `block_size`arg seems to have a different meaning than the `block_size` in other ops, so we should probably rename it, may guess is the `block_size` of tensor values that share the same kmeans cluster value.

jerryzh168: oh wait, the granularity of codebook is separate, let me take a look again

metascroy (approved):
Stamping to unblock as discussed

torchao/prototype/quantization/codebook/codebook_ops.py
jerryzh168: will update docs after I update the code to support block_size

PR #2001, opened 2025-04-02
Add gguf q4_k quantization

jerryzh168:
Summary:
Didn't implement the algorithm to choose_qparams from gguf, since it's complicated, e.g. https://github.com/ggml-org/llama.cpp/blob/f423981ac806bf031d83784bcb47d2721bc70f97/ggml/src/ggml-quants.c#L744 and https://github.com/ggml-org/llama.cpp/blob/f423981ac806bf031d83784bcb47d2721bc70f97/ggml/src/ggml-quants.c#L827C14-L827C28

but implemented a simple choose_qparams that can fit the gguf format: Q4_K: w = q * block_scale(6-bit) + block_min(6-bit)

The goal to port gguf format here in torchao is trying to compose that with our existing accuracy preserving techniques like GPTQ, AutoRound, QAT to see if we can help improve the accuracy.

also produced https://huggingface.co/jerryzh168/phi4-mini-torchao-gguf-q4_k with this change and verified with lm-eval that it has good accuracy

Test Plan:
python test/prototype/test_gguf_quant.py

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/prototype/test_gguf_quant.py
+++ b/test/prototype/test_gguf_quant.py
@@ -15,7 +15,7 @@ class TestGGUFQuantization(unittest.TestCase):
     def setUp(self):
         torch.manual_seed(123)
         self.input = torch.randn(2, 256, dtype=torch.float32)
-        self.n_super_blocks = 8
+        self.n_blocks_per_superblock = 8
         self.block_size = (1, 32)
         self.dtype = torch.uint4
 
@@ -34,7 +34,7 @@ class TestGGUFQuantization(unittest.TestCase):
     def test_gguf_quantized_tensor_from_float(self):
         gqt = GGUFQuantizedTensor.from_float(
             self.input,
-            self.n_super_blocks,
+            self.n_blocks_per_superblock,
             self.dtype,
         )
 
--- a/test/prototype/test_gguf_quant.py
+++ b/test/prototype/test_gguf_quant.py
@@ -5,10 +5,10 @@ import torch
 from torchao.prototype.quantization.gguf import (
     GGUFQuantizedTensor,
     GGUFWeightOnlyConfig,
-    choose_qparams_gguf,
 )
-from torchao.quantization import quantize_
+from torchao.quantization.quant_primitives import choose_qparams_gguf
 from torchao.quantization.utils import compute_error
+from torchao.quantization import quantize_
 
 
 class TestGGUFQuantization(unittest.TestCase):
--- a/test/prototype/test_gguf_quant.py
+++ b/test/prototype/test_gguf_quant.py
@@ -1,3 +1,9 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+
 import unittest
 
 import torch
@@ -6,9 +12,9 @@ from torchao.prototype.quantization.gguf import (
     GGUFQuantizedTensor,
     GGUFWeightOnlyConfig,
 )
+from torchao.quantization import quantize_
 from torchao.quantization.quant_primitives import choose_qparams_gguf
 from torchao.quantization.utils import compute_error
-from torchao.quantization import quantize_
 
 
 class TestGGUFQuantization(unittest.TestCase):
--- a/test/prototype/test_gguf_quant.py
+++ b/test/prototype/test_gguf_quant.py
@@ -1,0 +1,53 @@
+import unittest
+
+import torch
+
+from torchao.prototype.quantization.gguf import (
+    GGUFQuantizedTensor,
+    GGUFWeightOnlyConfig,
+    choose_qparams_gguf,
+)
+from torchao.quantization import quantize_
+from torchao.quantization.utils import compute_error
+
+
+class TestGGUFQuantization(unittest.TestCase):
+    def setUp(self):
+        torch.manual_seed(123)
+        self.input = torch.randn(2, 256, dtype=torch.float32)
+        self.n_super_blocks = 8
+        self.block_size = (1, 32)
+        self.dtype = torch.uint4
+
+    def test_choose_qparams_gguf(self):
+        (
+            super_block_scale_scale,
+            super_block_min_scale,
+            quantized_block_scale,
+            quantized_block_min,
+        ) = choose_qparams_gguf(self.input, self.block_size, self.dtype)
+
+        assert super_block_scale_scale.shape, (2, 8)
+        assert super_block_min_scale.shape, (2, 8)
+        assert quantized_block_scale.shape, (2, 32)
+
+    def test_gguf_quantized_tensor_from_float(self):
+        gqt = GGUFQuantizedTensor.from_float(
+            self.input,
+            self.n_super_blocks,
+            self.dtype,
+        )
+
+        dequant = gqt.dequantize()
+
+        sqnr = compute_error(dequant, self.input)
+        self.assertGreater(sqnr, 30)
+
+    def test_quantize_api(self):
+        m = torch.nn.Sequential(torch.nn.Linear(256, 64))
+        quantize_(m, GGUFWeightOnlyConfig())
+        assert type(m[0].weight) == GGUFQuantizedTensor
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/torchao/prototype/quantization/gguf/gguf_quantized_tensor.py
+++ b/torchao/prototype/quantization/gguf/gguf_quantized_tensor.py
@@ -1,0 +1,237 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from dataclasses import dataclass
+from typing import Optional
+
+import torch
+
+from torchao.core.config import AOBaseConfig
+from torchao.quantization.quant_primitives import (
+    choose_qparams_gguf,
+    dequantize_gguf,
+    quantize_gguf,
+)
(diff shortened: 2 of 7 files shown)


kimishpatel:
havent looked at the code, but does it also implement super block scale?

jerryzh168:
> havent looked at the code, but does it also implement super block scale?

yeah this is exactly what this PR is implementing :)

torchao/quantization/quant_primitives.py
```diff
@@ -1039,6 +1044,172 @@ def reshape_w(w):
...
+
+    block_scale_absmax = torch.amax(
+        torch.abs(block_scale_clone), dim=reduction_dims, keepdim=False
+    )
+    block_min_absmax = torch.amax(
+        torch.abs(block_min_clone), dim=reduction_dims, keepdim=False
+    )
+
+    # 2. get super_block_scale_scale and super_block_min_scale
+    quant_max = 2**6 - 1
```
kimishpatel, line 1101: is this for 6 bit?

kimishpatel: oh nevermind. this is for 6bit scale value. rename this appropriately else it is confusing

torchao/prototype/quantization/gguf/gguf_quantized_tensor.py
```diff
@@ -0,0 +1,237 @@
...
+def _gguf_weight_only_transform(
+    module: torch.nn.Module,
+    config: GGUFWeightOnlyConfig,
+):
+    """
+    Applies gguf weight-only quantization to linear layers.
+
+    Args:
+        dtype: torch.uint1 to torch.uint8, torch.int32 supported.
+        n_super_blocks: the number of super blocks in a 256 element block for gguf, e.g. when it is 8
```
metascroy, line 224: It seems this is blocks_per_superblock, not n_super_blocks?

torchao/prototype/quantization/gguf/gguf_quantized_tensor.py
```diff
@@ -0,0 +1,237 @@
...
+            quantized_block_scale,
+            quantized_block_min,
+            int_data,
+            input_float.shape,
+            dtype=torch.float16,
+        )
+
+
+@dataclass
+class GGUFWeightOnlyConfig(AOBaseConfig):
```
metascroy, line 209: Have you checked that this is generic enough to capture all of their superblock affine schemes?

torchao/prototype/quantization/gguf/gguf_quantized_tensor.py
```diff
@@ -0,0 +1,237 @@
...
+class GGUFQuantizedTensor(TorchAOBaseTensor):
+    """
+    A Tensor subclass that when applied to a weight used in a linear op/module,
+    changes that linear op to a weight-only int4 quantized linear op with groupwise
+    affine quantization on the weight.
+    """
+
+    @staticmethod
+    def __new__(
+        cls,
```
metascroy, line 41: Does GGUF have a utility that can construct their packed tensors from these values?

torchao/prototype/quantization/gguf/gguf_quantized_tensor.py
compilade: If it included the number of bits per sub-block scales and mins (both are usually the same), then it would be easier to adapt to more types.

Some types use 4-bit sub-block scales and mins (`Q2_K`), others 6-bit (`Q3_K`, `Q4_K`, `Q5_K`) and others 8-bit (`Q6_K`).

torchao/prototype/quantization/gguf/gguf_quantized_tensor.py
compilade: Not yet for k-quants, sorry. It is planned though, but it will take some time.

But it would not be a drop-in replacement here anyway since the `gguf` Python package uses Numpy for its calculations, not PyTorch. (at least for now)

test/prototype/test_gguf_quant.py
```diff
@@ -0,0 +1,53 @@
+import unittest
+
+import torch
+
+from torchao.prototype.quantization.gguf import (
```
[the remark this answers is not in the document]
jerryzh168: haven't explored how to export yet, will do in next PR

torchao/prototype/quantization/gguf/gguf_quantized_tensor.py
jerryzh168: yeah it will be easy to extend to other variations, I just start with Q4_K for now

torchao/prototype/quantization/gguf/gguf_quantized_tensor.py
jerryzh168: thanks for the context @compilade, I think maybe we can try to see if we can use the export script from autoround: https://github.com/intel/auto-round/blob/eb793488638105980f18ed7a40301d29d5c75ca0/auto_round/export/export_to_gguf/convert.py#L1159-L1169

compilade: > I think maybe we can try to see if we can use the export script from autoround

This will likely be very, very slow (more than 20x the C version) since they re-implement `make_qkx2_quants` in Numpy: <https://github.com/intel/auto-round/blob/eb793488638105980f18ed7a40301d29d5c75ca0/auto_round/export/export_to_gguf/quant.py#L164>.

Technically, you may not necessarily need the search for the scales and mins to pack to `Q4_K`, assuming you already have them and/or they are trainable parameters.

I think the way you did it with separate tensors for the scales and mins does seem appropriate to avoid some of the complexity of the packing format (since from what I understand, this is intended to be some in-memory quantization used in QAT? Do correct me if I'm wrong). (the scales and mins in `Q4_K` [are notoriously packed together in 12 bytes]( <https://github.com/ggml-org/llama.cpp/blob/7a84777f42a9b3ba47db5d20b7662f8ddf92f652/gguf-py/gguf/quants.py#L484-L496>), but that's not relevant if this is intended as an in-memory quantization with different kernels than the ones in `ggml`)

You only need to worry about the packing format if you are exporting to GGUF. (Dequantization is implemented in the upstream `gguf` Python package for most types (including k-quants and i-quants) already, it's only quantization which is currently limited to `Q4_0`, `Q4_1`, `Q5_0`, `Q5_1`, and `Q8_0` because k-quants in Python were too slow to pack (although this could change after <https://github.com/ggml-org/llama.cpp/pull/12557>, which might simplify the search for scales (and mins, if generalized)))

It would be possible, though, to add an API to the upstream `gguf` Python package which would skip the search for the scales and mins but still allow quantizing to `Q4_K` and similar, but I'm not sure how it would/should be used.

jerryzh168: > This will likely be very, very slow (more than 20x the C version) since they re-implement make_qkx2_quants in Numpy: [intel/auto-round@eb79348/auto_round/export/export_to_gguf/quant.py#L164](https://github.com/intel/auto-round/blob/eb793488638105980f18ed7a40301d29d5c75ca0/auto_round/export/export_to_gguf/quant.py#L164).
> Technically, you may not necessarily need the search for the scales and mins to pack to Q4_K, assuming you already have them and/or they are trainable parameters.

yes that's correct I think, we may have to adapt that code a bit for torchao quants to work, but we'll be providing scale/min from torchao (current plan is to use GPTQ, AutoRound or QAT), so won't need to run that path, will need to make sure this path is taken instead: https://github.com/intel/auto-round/blob/eb793488638105980f18ed7a40301d29d5c75ca0/auto_round/export/export_to_gguf/quant.py#L326

> if this is intended as an in-memory quantization with different kernels than the ones in ggml)

we do want to target ggml kernels in the end still, but overall goal here is try to leverage existing torchao post training accuracy preserving techniques like GPTQ, AutoRound etc. and quantization aware training techniques to see if we can help improve the accuracy of various gguf quantization schemes through composing with these existing techniques (and relying on user data).

Regarding search algorithms in gguf, yeah I feel they are quite complicated (make_qx_quant, make_qp_quant etc. I also haven't looked at imatrix as stuff), and it might be error prone for me to port them here, also I didn't see them anywhere else, @compilade can you share how these are derived at the high level, and how we can understand them better.

metascroy (approved):

jerryzh168:
thanks, I'll merge now since CI is happy, will add more docs next time

torchao/prototype/quantization/gguf/gguf_quantized_tensor.py
compilade: @jerryzh168

Late reply, but maybe still useful:

> Regarding search algorithms in gguf, yeah I feel they are quite complicated

Yes, they are. I would not recommend re-implementing them here (unless you really want to restructure the search algorithms to make them more PyTorch-friendly to make them fast enough (which will need to be done at some point for the `gguf` Python library in the `llama.cpp` repo)).
Packing, on the other hand, is more approachable.

> can you share how these are derived at the high level, and how we can understand them better. i.e. the high level motivation/direction/constraints that you are working with

These search algorithms, from my understanding, are weighted rounding algorithms, which search for the best integers which, when scaled (and/or offset, depending on the type), minimize the weighted squared error.

Scales make the norm of the quantized search space irrelevant because it can always be adjusted. The grid of integers can be traversed linearly to cumulatively find the best approximation (conceptually by trying all pre-rounding scales which would result in a distinct rounding).
Offsets project the search space on the hyperplane perpendicular to `[1, 1, 1, ...]` (aka, the mean is zero-ed), which makes it a bit more complicated to traverse.

> are all of them trying to minimize the rounding error? what about clamping error? 

The search algorithms for the scales and mins are minimizing the weighted error. By extension, this also includes rounding error and clamping error.

> why don't you use data to improve quantization accuracy?

This is the purpose of `imatrix`, which basically allows to make the search algorithms activation-aware through the weights of the error (with per-channel importance or similar).

PR #2015, opened 2025-04-04
Fix slice and padding for TensorCoreTiledLayout

jerryzh168:
Summary:
Previously some of the code paths are not exercised, so the bug was not discovered

but there are some bug related to slice operation and padding, basically scale and zero_point are not padded before, this results in errors when it is required.

Test Plan:
python test/dtypes/test_affine_quantized.py -k test_slice

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/dtypes/uintx/gemlite_layout.py
+++ b/torchao/dtypes/uintx/gemlite_layout.py
@@ -297,16 +297,9 @@ class GemliteAQTTensorImpl(TensorCoreTiledAQTTensorImpl):
         if func is aten.slice.Tensor:
             self, dim, start, end, step = fill_defaults(args, 5, [0, None, None, 1])
             if dim == 0:
-                int_data, scale, zero_point = self.get_plain()
-                int_data = aten.slice.Tensor(int_data, dim, start, end, step)
-                int_data = self._layout.post_process(int_data)
-                sliced = self.from_plain(int_data, scale, zero_point, self._layout)
-                return return_and_correct_aliasing(func, args, kwargs, sliced)
-            elif dim == 1:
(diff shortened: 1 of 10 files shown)


torchao/dtypes/uintx/gemlite_layout.py
```diff
@@ -297,14 +297,31 @@ def __torch_dispatch__(cls, func, types, args, kwargs):
         if func is aten.slice.Tensor:
```
drisspg, line 297: can we update this class to use the @dispatch pattern

torchao/dtypes/uintx/int4_cpu_layout.py
```diff
@@ -192,31 +192,26 @@ def __torch_dispatch__(cls, func, types, args, kwargs):
 
         if func is aten.slice.Tensor:
             self, dim, start, end, step = fill_defaults(args, 5, [0, None, None, 1])
```
drisspg, line 194: same for this

torchao/dtypes/uintx/tensor_core_tiled_layout.py
```diff
@@ -335,31 +350,25 @@ def __torch_dispatch__(cls, func, types, args, kwargs):
 
         if func is aten.slice.Tensor:
```
drisspg, line 351: also same for this

torchao/dtypes/uintx/gemlite_layout.py
jerryzh168: sure, can I do this in a separate PR?

drisspg (approved):

PR #2017, opened 2025-04-04
Update torchao.prototype.parq and add 4-bit Llama 3.2 1B benchmark

lisjin:
We would like to merge recent changes from our open sourced library at https://github.com/facebookresearch/parq.

We have also benchmarked 4-bit Llama 3.2 1B fine-tuned for 25K steps on fineweb-edu using torchtune. We used PARQ's [`MaxUnifQuantizer`](https://github.com/facebookresearch/parq/blob/b41f00e7e59fb6dc939ce371d2b6f40f71582895/parq/quant/uniform.py#L114) and [`ProxHardQuant`](https://github.com/facebookresearch/parq/blob/b41f00e7e59fb6dc939ce371d2b6f40f71582895/parq/optim/proxmap.py#L28) proximal mapping, which is equivalent to STE. Below are the relevant training config changes to the [llama3_2/1B_full.yaml](https://github.com/pytorch/torchtune/blob/84aea68d90ff5f8e65af33fd004e7b84424bc712/recipes/configs/llama3_2/1B_full.yaml) recipe.
```
batch_size: 8
epochs: 1
optimizer:
  _component_: torch.optim.AdamW
  lr: 4e-5
  weight_decay: 0.0
  betas: [0.9, 0.95]
  fused: True

lr_scheduler:
  _component_: torchtune.training.lr_schedulers.get_cosine_schedule_with_warmup
  num_warmup_steps: 2500
 ```

As shown in the table below, the resulting 4-bit model achieves well under 10% accuracy on most commonsense reasoning benchmarks relative to the pre-trained model.
    Tasks    |16-bit|4-bit |% diff
-------------|-----:|-----:|-----:
arc_challenge|0.3805|0.3575|-6.4  
arc_easy     |0.6309|0.6077|-3.7  
hellaswag    |0.6081|0.5423|-10.8 
piqa         |0.7410|0.7122|-3.9  
winogrande   |0.6022|0.5549|-7.8

### DIFF
--- a/torchao/prototype/parq/__init__.py
+++ b/torchao/prototype/parq/__init__.py
@@ -1,3 +1,9 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+
 from .optim import (  # noqa: F401
     ProxBinaryRelax,
     ProxHardQuant,
--- a/torchao/prototype/parq/optim/__init__.py
+++ b/torchao/prototype/parq/optim/__init__.py
@@ -1,3 +1,9 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+
 from .binarelax import ProxBinaryRelax  # noqa: F401
 from .parq import ProxPARQ  # noqa: F401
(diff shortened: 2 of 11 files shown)


andrewor14:
Hi @lisjin, thanks for the update! The results look great. As discussed offline, we usually only add submodules if other parts of torchao are using the submodules, which is not the case here. Do you mind making the latest changes to PARQ in the prototype version here in torchao instead?

lisjin:
> Do you mind making the latest changes to PARQ in the prototype version here in torchao instead?

Of course! I just removed the submodule and updated the prototype version instead.

andrewor14 (approved):

PR #2037, opened 2025-04-10
Fix QDQLayout

metascroy:
When testing QDQLayout on evaluation benchamarks.  Depending on whether int_data's dtype was torch.int8 vs. torch.int32, I would see terrible or good metrics.  But in both cases, the dequantized values were identical.

I traced the issue down to __new__ in the tensor impl class:
```
kwargs["dtype"] = int_data.dtype
```

If I set this to torch.int32, the problem would resolve.  The problem would also resolve if I give QDQLayout its own tensor impl, rather than inherit from PlainLayout's tensor impl (even keeping the dtype as int_data.dtype when int_data was torch.int8).  This PR does the latter for the fix.

### DIFF
--- a/torchao/dtypes/uintx/q_dq_layout.py
+++ b/torchao/dtypes/uintx/q_dq_layout.py
@@ -24,19 +24,199 @@ handler.setFormatter(formatter)
 logger.addHandler(handler)
 
 
-from torchao.dtypes.utils import PlainLayout
+from dataclasses import dataclass
+from typing import Optional, Tuple
+
+from torch.utils._python_dispatch import (
+    return_and_correct_aliasing,
+)
+
+from torchao.dtypes.utils import AQTTensorImpl, Layout
+from torchao.utils import fill_defaults
+
+aten = torch.ops.aten
 
 
-class QDQLayout(PlainLayout):
+@dataclass(frozen=True)
+class QDQLayout(Layout):
     pass
 
 
-from torchao.dtypes.uintx.plain_layout import PlainAQTTensorImpl
+def _same_metadata(self: "QDQTensorImpl", src: "QDQTensorImpl") -> bool:
+    return (
+        isinstance(self, QDQTensorImpl)
+        and isinstance(src, QDQTensorImpl)
+        and self.shape == src.shape
+        and self.int_data.shape == src.int_data.shape
+        and self.scale.shape == src.scale.shape
+        and (self.zero_point is None and src.zero_point is None)
+        or (
+            self.zero_point is not None
+            and src.zero_point is not None
+            and self.zero_point.shape == src.zero_point.shape
+        )
+        and type(self._layout) == type(src._layout)
+    )
 
 
 @register_layout(QDQLayout)
-class _Impl(PlainAQTTensorImpl):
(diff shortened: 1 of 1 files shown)


torchao/dtypes/uintx/q_dq_layout.py
```diff
@@ -24,19 +24,199 @@
...
+            )
+
+        elif func is aten.t.default:
+            tensor = args[0]
+            new = tensor.__class__(
+                tensor.int_data.t(), tensor.scale, tensor.zero_point, tensor._layout
+            )
+            return return_and_correct_aliasing(func, args, kwargs, new)
+
+        elif func is aten.slice.Tensor:
```
jerryzh168, line 172: the previous impl might have some issues, I now changed to always slice both int_data, scale and zp: https://github.com/pytorch/ao/blob/a3b857fcc6dbf89c2e88c887689f271e2c11859d/torchao/dtypes/uintx/tensor_core_tiled_layout.py#L353

torchao/dtypes/uintx/q_dq_layout.py
```diff
@@ -24,19 +24,199 @@
...
+        scale: torch.Tensor,
+        zero_point: Optional[torch.Tensor],
+        _layout: Layout,
+    ):
+        kwargs = {}
+        kwargs["device"] = int_data.device
+        kwargs["layout"] = (
+            kwargs.get("layout") if kwargs.get("layout", False) else int_data.layout
+        )
+        kwargs["dtype"] = int_data.dtype
```
jerryzh168, line 86: isn't this the same: https://github.com/pytorch/ao/blob/a3b857fcc6dbf89c2e88c887689f271e2c11859d/torchao/dtypes/uintx/plain_layout.py#L67

metascroy: Yes, that's what is it in PlainLayout, too.  This file just copies PlainLayout under the new name QDQLayout.  Previously, it subclassed from PlainLayout, but this lead to issues.  I could overcome the issue in two ways:

1. Change to "kwargs["dtype"] = torch.int32" in PlainLayout, and keep the subclassing
2. Keep "kwargs["dtype"] = int_data.dtype", but do not subclass PlainLayout.

torchao/dtypes/uintx/q_dq_layout.py
metascroy: This is what it is in PlainLayout right now.  Is that no longer correct?

jerryzh168: yeah, I think we should update that, I only updated tensor core tiled layout last time

torchao/dtypes/uintx/q_dq_layout.py
jerryzh168: I see, sounds good to copy. just curious, where does the issue come from? if the QDQLayout and PlainLayout have the same implementation?

metascroy: I wasn't able to pinpoint it

jerryzh168 (approved):

PR #2042, opened 2025-04-11
Using autoround's implementation for gguf q4_k

jerryzh168:
Summary:
Since autoround has export code for GGUF, we can reuse that by also porting over their (naive) quant algorithms for gguf

Test Plan:
TODO

Reviewers:

Subscribers:

Tasks:

Tags:


ambroser53:
Is there any plans to finish and release this feature? Has it been shown to work as is?

jerryzh168:
we have deprioritized this work due to unclear needs, but open to community contributions, I remember there were some issues with using lm-eval with gguf models before: https://github.com/EleutherAI/lm-evaluation-harness/issues/2887 so no way to confirm the implementation is correct or not until that is fixed

PR #2043, opened 2025-04-11
Enabling MOE Quantization using linear decomposition

HDCharles:
Enabling MOE Quantization using linear decomposition

Summary: This PR is a first step at optimizing moe inference using
torchAO. The goal for this step is to enable existing quantization
kernels and workflows to work for moe quantization by decomposing the
group gemm into a sequence of unbalanced linear ops that can use the
existing quantized kernels. To enable this we had to add support for
quantizing these 3D tensors as well as slicing and indexing. 2 methods
of achieving this were implemented. for int8wo, int8dq, int4wo, fp8wo,
fp8dq, the underlying quantized tensor subclass was adapted to both
support 3D tensors, indexing and slicing, as well as an updated
transformation function that can handle the
ConditionalFeedForwardAOQuantizable modules if the filter funciton in
quantize_ is used to target the aforementioned module. For some complex kernels
which use packed data that couldn't be made to easily work in 3D, we
also added FakeExtraDimTensor which can transform any
quantized tensor subclass into supporting the necessary slice and index
operations for moe quantization. This option is enabled by using
MoeQuantConfig.

This can be applied to huggingface llama4 for instance as shown int he
llama4_quant.py example. Since the hf moe module is implemented in a way
that's not condusive to quantization, it first requires a module swap to
the MOEFeedForwardAOQuantizable.

```python
# API
from torchao.quantization.prototype.moe_quant.utils import cond_ffn_filter, MoEQuantConfig
from torchao.quantization.quant_api import quantize_, Int8WeightOnlyConfig

 quantize_(model, MoEQuantConfig(Int8WeightOnlyConfig()), filter_fn=cond_ffn_filter)
 model=torch.compile(model, mode="reduce-overhead", fullgraph=is_single_token_inference)
 # there's unavoidable graph break in multi token inference during token shuffling
```

note the MoEQuantConfig has an attribute use_fake_extra_dim_tensor that defaults to AS_FALLBACK but can be set to TRUE or FALSE as well to either automatically enable or disable that option.

Benchmarks:

|             | batchsize 1 |             | batchsize 8 |              |             |
|-------------|-------------|-------------|-------------|--------------|-------------|
| Technique   | tok/s       | memory (GB) | tok/s       | tok/s* batch | memory (GB) |
| None        |       78.35 |       93.76 |        18.2 |       145.64 |       94.12 |
| int8wo      |        6.14 |       49.13 |        5.01 |        40.09 |       49.23 |
| int8wo-base |        98.4 |       48.87 |        4.94 |        39.56 |        49.2 |
| int4wo      |       14.25 |       30.21 |       11.84 |        94.75 |       30.19 |
| int4wo-base |       79.38 |       36.15 |       10.29 |        82.29 |       36.12 |
| fp8wo       |         3.2 |       50.31 |        2.88 |        23.08 |       50.29 |
| fp8wo-base  |       59.41 |       52.07 |        2.98 |        23.81 |       52.05 |
| fp8dq       |        9.78 |       50.92 |        4.08 |        32.61 |       50.89 |
| fp8dq-base  |       45.92 |       53.97 |        3.78 |        30.23 |       53.94 |

Note: its unclear why the memory usage increases when going from batchsize 1 to batchsize 8, my supposition is that because batchsize 1 can be compiled with fullgraph, the kernels are likely to use heavier fusion which can result in larger kernels.

Memory Profile of Quantization of MoE Modules:
image

note: quantization can be done on cpu as in llama4_quant.py so that if only the int4 model fits on cuda you can still run the model. this memory trace was taken from the mixtral moe memory profile by first loading the whole model to cuda.

Test Plan:
python test/quantization/test_moe_quant.py

python
test/torchao/experimental/tests/test_int8_dynamic_activation_intx_weight.py
-k "test_moe_quant_intx"

sh torchao/_models/mixtral-moe/run.sh

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/_models/mixtral-moe/generate.py
+++ b/torchao/_models/mixtral-moe/generate.py
@@ -238,8 +238,8 @@ def main(
 
     from torchao.quantization.prototype.moe_quant.utils import (
         MoEQuantConfig,
+        UseFakeExtraDimTensor,
         cond_ffn_filter,
-        UseFakeExtraDimTensor
     )
     from torchao.quantization.quant_api import (
         Float8DynamicActivationFloat8WeightConfig,
@@ -260,28 +260,42 @@ def main(
             config = MoEQuantConfig(Int8WeightOnlyConfig())
 
         elif "int8wo" in moe_quant:
-            config = MoEQuantConfig(Int8WeightOnlyConfig(), use_fake_extra_dim_tensor=UseFakeExtraDimTensor.TRUE)
+            config = MoEQuantConfig(
+                Int8WeightOnlyConfig(),
+                use_fake_extra_dim_tensor=UseFakeExtraDimTensor.TRUE,
+            )
 
         elif "int8dq-base" in moe_quant:
             config = MoEQuantConfig(Int8DynamicActivationInt8WeightConfig())
 
         elif "int8dq" in moe_quant:
-            config = MoEQuantConfig(Int8DynamicActivationInt8WeightConfig(), use_fake_extra_dim_tensor=UseFakeExtraDimTensor.TRUE)
+            config = MoEQuantConfig(
+                Int8DynamicActivationInt8WeightConfig(),
+                use_fake_extra_dim_tensor=UseFakeExtraDimTensor.TRUE,
+            )
 
         elif "int4wo-base" in moe_quant:
             config = MoEQuantConfig(Int4WeightOnlyConfig())
 
         elif "int4wo" in moe_quant:
-            config = MoEQuantConfig(Int4WeightOnlyConfig(), use_fake_extra_dim_tensor=UseFakeExtraDimTensor.TRUE)
+            config = MoEQuantConfig(
+                Int4WeightOnlyConfig(),
+                use_fake_extra_dim_tensor=UseFakeExtraDimTensor.TRUE,
+            )
 
         elif "fp8wo-base" in moe_quant:
             config = MoEQuantConfig(Float8WeightOnlyConfig())
 
         elif "fp8wo" in moe_quant:
-            config = MoEQuantConfig(Float8WeightOnlyConfig(), use_fake_extra_dim_tensor=UseFakeExtraDimTensor.TRUE)
+            config = MoEQuantConfig(
+                Float8WeightOnlyConfig(),
+                use_fake_extra_dim_tensor=UseFakeExtraDimTensor.TRUE,
+            )
 
         elif "fp8dq-base" in moe_quant:
-            config = MoEQuantConfig(Float8DynamicActivationFloat8WeightConfig(granularity=PerRow()))
+            config = MoEQuantConfig(
+                Float8DynamicActivationFloat8WeightConfig(granularity=PerRow())
+            )
 
         elif "fp8dq" in moe_quant:
             config = MoEQuantConfig(
--- a/torchao/_models/mixtral-moe/generate.py
+++ b/torchao/_models/mixtral-moe/generate.py
@@ -208,7 +208,6 @@ def main(
     assert checkpoint_path.is_file(), checkpoint_path
     tokenizer_path = checkpoint_path.parent / "tokenizer.model"
     assert tokenizer_path.is_file(), str(tokenizer_path)
-    print(f"Memory used: {torch.cuda.max_memory_reserved() / 1e9:.02f} GB")
     print(f"Using device={device}")
     precision = torch.bfloat16
     is_chat = "chat" in str(checkpoint_path)
@@ -220,10 +219,10 @@ def main(
 
     print("Loading model ...")
     t0 = time.time()
-    model = _load_model(checkpoint_path, device, precision)
+    model = _load_model(checkpoint_path, "cpu", precision)
 
-    device_sync(device=device)  # MKG
     print(f"Time to load model: {time.time() - t0:.02f} seconds")
+    t0 = time.time()
 
     tokenizer = SentencePieceProcessor(model_file=str(tokenizer_path))
     encoded = encode_tokens(tokenizer, prompt, bos=True, device=device)
@@ -299,7 +298,12 @@ def main(
 
         if config is not None:
             quantize_(model, config, filter_fn=cond_ffn_filter)
-            torch.cuda.reset_peak_memory_stats()
+            print(f"Time to apply quantization to model: {time.time() - t0:.02f} seconds")
+    
+    model.to(device=device)
+    device_sync(device=device)
+
+    print(f"C: {torch.cuda.max_memory_reserved() / 1e9:.02f} GB")
 
     if compile:
         # moe quant + compile causes repeated warnings
@@ -382,7 +386,7 @@ def main(
 
         if not interactive:
             pass
-            print(tokenizer.decode(y[0].tolist()))
+            # print(tokenizer.decode(y[0].tolist()))
         else:
             print()
         tokens_generated = y.size(-1) - prompt_length
--- a/torchao/_models/mixtral-moe/generate.py
+++ b/torchao/_models/mixtral-moe/generate.py
@@ -386,16 +386,16 @@ def main(
 
         if not interactive:
             pass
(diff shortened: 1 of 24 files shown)


torchao/_models/mixtral-moe/generate.py
```diff
@@ -0,0 +1,393 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+
+# This source code is licensed under the license found in the
+# LICENSE file in the root directory of this source tree.
+import itertools
```
drisspg, line 6: Do we need all this? 

I understand it is good to have an MOE model but is it important to have our own definition here?

torchao/dtypes/affine_quantized_tensor_ops.py
```diff
@@ -490,10 +492,50 @@ def _(func, types, args, kwargs):
         self.quant_max,
         self.zero_point_domain,
         dtype=self.dtype,
-        strides=self.stride(),
+        strides=self.stride() if len(block_size)==2 else None,
```
drisspg, line 495: why this?

HDCharles: the stride can change for 3 dimensional tensors when doing slicing and indexing ops.

torchao/_models/mixtral-moe/generate.py
HDCharles: we've discussed adding a moe model for a while and this was the main one for testing and benchmarking and demonstrating the technique.

torchao/dtypes/floatx/float8_layout.py
```diff
@@ -159,6 +159,13 @@ def __torch_dispatch__(cls, func, types, args, kwargs):
             raise ValueError(
                 f"Not supported args for copy_ due to metadata mistach: {args[0], args[1]}"
             )
+        elif func is aten.select.int or func is aten.index.Tensor:
```
drisspg, line 162: Nit: `elif func in [aten.select.int, aten.index.Tensor]:`

torchao/dtypes/uintx/tensor_core_tiled_layout.py
```diff
@@ -93,11 +93,13 @@ def _linear_bf16_act_uint4_weight_impl(input_tensor, weight_tensor, bias):
...
 
     # groupwise int4 quantization
-    groupsize = weight_tensor.block_size[1]
-    y = torch.ops.aten._weight_int4pack_mm(
-        act_mat.contiguous(), packed_weight, groupsize, scale_and_zero
-    )
-
+    groupsize = weight_tensor.block_size[-1]
+    if act_mat.numel() == 0:
+        y = act_mat
```
drisspg, line 98: maybe a note, on "handiling for empty inputs"

torchao/dtypes/uintx/tensor_core_tiled_layout.py
```diff
@@ -416,11 +452,16 @@ def block_size(self):
 
         scale, zero = unpack_tinygemm_scales_and_zeros(self.scale_and_zero)
         cur_shape = self.shape
-        assert len(cur_shape) == 4
+        if len(cur_shape) == 5:
```
drisspg, line 455: cc @jerryzh168 since you are also making some changes to how slicing is handled

torchao/experimental/tests/test_int8_dynamic_activation_intx_weight.py
```diff
@@ -629,6 +629,53 @@ def test_identical_to_Int8DynActInt4WeightQATQuantizer(
...
+        )
+        from torchao.quantization.utils import compute_error
+
+        with torch.device("cpu"):
+            model = MOEFeedForwardAOQuantizable(512, 256, 8, 2).to(torch.bfloat16)
+            x = torch.randn(8, 512, dtype=torch.bfloat16)
+
+        out = model(x).clone()
+
+        # base_config = Int8DynamicActivationIntxWeightConfig()
```
drisspg, line 654: nit remove

torchao/quantization/transform_module.py
```diff
@@ -47,5 +47,6 @@ def _transform(
     @functools.wraps(config_type)
     def decorator(func):
         _QUANTIZE_CONFIG_HANDLER[config_type] = func
+        return func  # needed to make the functions usable externally
```
drisspg, line 50: ?

torchao/quantization/quant_api.py
```diff
@@ -1592,10 +1655,39 @@ def _float8_dynamic_activation_float8_weight_transform(
...
+            quantized_weight = _float8_dynamic_activation_float8_weight_quantize_tensor(
+                weight, config
+            )
+            new_weight = torch.nn.Parameter(quantized_weight, requires_grad=False)
+            setattr(module, weight_attr, new_weight)
+        return module
+    else:
+        assert hasattr(module, "weight"), (
+            "applying float8 dynamic activation quant requires module to have weight attribute"
+            + "but {module} does not have one"
```
drisspg, line 1683: f"..."

torchao/quantization/transform_module.py
HDCharles: if you wanted to use any of the transform functions in quant_api.py that use the decorator, it would say that the function is actually just `None` since nothing was being returned after the function was passed into the decorator. They were only stored in the transform function dict so they were only usable through quantize_. This solves that problem. Since for MoE we use these transform functions, this was necessary and doesn't affect anything.

torchao/quantization/prototype/moe_quant/utils.py
```diff
@@ -0,0 +1,261 @@
...
+                return func(*args, **kwargs)
+        except Exception as e:
+            print(f"ERR: subclass {cls} doesn't implement {func}, got error: {e}")
+
+    @classmethod
+    def __torch_dispatch__(cls, func, types, args, kwargs):
+        kwargs = {} if kwargs is None else kwargs
+
+        if func == aten.slice.Tensor:
+            self, dim, start, end, step = fill_defaults(args, 5, [0, None, None, 1])
```
drisspg, line 175: we should probably error better when dim is not 0 right?

torchao/quantization/quant_api.py
```diff
@@ -1126,9 +1120,32 @@ def _int4_weight_only_transform(
...
+
+
+@register_quantize_module_handler(Int4WeightOnlyConfig)
+def _int4_weight_only_transform(
+    module: torch.nn.Module, config: Int4WeightOnlyConfig
+) -> torch.nn.Module:
+    if config.set_inductor_config:
+        torchao.quantization.utils.recommended_inductor_config_setter()
+
+    if "ConditionalFeedForwardAOQuantizable" in str(type(module)):
```
drisspg, line 1133: my main question here is that it feels bad we need to re-write every quantize handler with this if elf branch especially for a prototype feature

It seems like these module handlers should be able to specify handling their specific nn module individually

HDCharles: I agree, I initially wanted to just use a custom transform function since we're using a custom filter function which targetting a different type of module but there's no way to do that with quantize_. I spoke to vasiliy who said that the config is supposed to define the quantize function and this was expected.

So if quantize_(model, config, new_filter_fn, *new_transform_fn*) is out, the alternative is to overwrite the existing transform function like in the PR to detect the module and behave differently, there's not really a good prototype-y way to do it and honestly the changes to quantization code so that they can handle 3D tensors is the bigger issue imo.

We could also rewrite the registration handler to do the registration based on module type, but a significant benfit of using tensor subclasses is that any module with a nn.funcitonal.linear op and a module.weight attribute can be quantized without issue. We would lose that if we hardcode the transform function to only apply to certain modules.

The best alternative in my mind would be targetting the tensors directly and changing behavior for 2D/3D tensor rather than module type. However this is a significant change to things since everything is module focused and there's no good way to detect which tensors are being used by which ops without a graph or knowledge of the module structure.

PR #2044, opened 2025-04-11
[reland][ROCm] preshuffled weight mm

jeffdaily:
Adds SwizzleTensor subclass that wraps a Tensor and reorders the contents to be suitable for HIPBLASLT_ORDER_COL16_4R8. SwizzleTensor intercepts torch.mm and replaces with custom calls to hipblaslt.


jeffdaily:
@mxz297 @jerryzh168 please re-review, kick of CI, thanks.

mxz297:
@jeffdaily "test-mps-ops" still seems to be failing to compile with 

      /Users/ec2-user/runner/_work/ao/ao/torchao/csrc/rocm/swizzle/swizzle.cpp:1:10: fatal error: 'hip/hip_runtime.h' file not found
      #include <hip/hip_runtime.h>
               ^~~~~~~~~~~~~~~~~~~

I wonder if we should just guard the whole source file under #if USE_ROCM

jeffdaily:
> @jeffdaily "test-mps-ops" still seems to be failing to compile with
> 
> ```
>   /Users/ec2-user/runner/_work/ao/ao/torchao/csrc/rocm/swizzle/swizzle.cpp:1:10: fatal error: 'hip/hip_runtime.h' file not found
>   #include <hip/hip_runtime.h>
>            ^~~~~~~~~~~~~~~~~~~
> ```
> 
> I wonder if we should just guard the whole source file under #if USE_ROCM

Done.

jerryzh168:
is this not fixed?

```
__w/ao/ao/pytorch/ao/torchao/csrc/rocm/swizzle/swizzle.cpp:4:10: fatal error: hip/hip_runtime.h: No such file or directory
    4 | #include <hip/hip_runtime.h>
      |          ^~~~~~~~~~~~~~~~~~~
compilation terminated.
ninja: build stopped: subcommand failed.
```

jerryzh168:
https://github.com/pytorch/ao/actions/runs/14604068152/job/40968945843?pr=2103

mxz297:
@jeffdaily  @jerryzh168  This is very strange....  So, previously, this error
```
__w/ao/ao/pytorch/ao/torchao/csrc/rocm/swizzle/swizzle.cpp:4:10: fatal error: hip/hip_runtime.h: No such file or directory
    4 | #include <hip/hip_runtime.h>
      |          ^~~~~~~~~~~~~~~~~~~
compilation terminated.
ninja: build stopped: subcommand failed.
```

only shows up on non-AMD platform, so we added a commit that will guard the whole source file under #if USE_ROCM. And we indeed no longer saw this failure anymore. And i saw clean merge signals before doing the merge.

So, i am a little bit surprised by:
(1) why the test failure did not show up before the merge?
(2) The failure now is on AMD platforms. Jeff have tested this on OSS platform and I have tested this inside meta. I feels like more likely the test platform does not have proper rocm setup?

@jeffdaily  Are you able to repro these rocm failures somehow?

jerryzh168:
the error appears in internal diff as well: https://www.internalfb.com/diff/D73052566 I think we should revert for now?

mxz297:
@jerryzh168  relied in the internal diff, but it seems like some other failure, which feels like caused by some other diff, though

petrex:
I am seeing the same error in wheel build. 
Maybe we are missing proper tool-chain/ rocm headers in the `pytorch/manylinux2_28-builder:rocm6.2.4` ? or just some env var? @amdfaa

HDCharles:
looking at our code, we have:

#if defined(USE_ROCM)
#include <hip/hip_bf16.h>
#include <hip/hip_fp16.h>
#include <hip/hip_runtime.h>
#endif

in tensor_core_tiled_layout.cu

Is the hip include here not gated correctly?

jeffdaily:
That looks gated correctly.  The CI build is missing -I/opt/rocm for some reason.  The header files are there, but flag is missing.

torchao/__init__.py
```diff
@@ -43,13 +43,14 @@
...
 
-from . import dtypes, optim, testing
+from . import dtypes, optim, swizzle, testing
 
 __all__ = [
     "dtypes",
     "autoquant",
     "optim",
     "quantize_",
+    "swizzle",
```
jerryzh168, line 53: why is this added to top level? should this be in prototype for now?

torchao/swizzle/__init__.py
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
jerryzh168, line 1: we don't want to create a new folder under torchao for this tensor/op I think..

jeffdaily: Where do you recommend for it to go?

jerryzh168: is this prototype? we can add to torchao/prototype for now

PR #1916, opened 2025-03-18
Duplicate pt2e quant code from pytorch/pytorch to pytorch/ao

jerryzh168:
Summary:
First step of https://dev-discuss.pytorch.org/t/torch-ao-quantization-migration-plan/2810

core logic of pt2e are duplicated, and also ported test here

Next:
* move meta internal callsites to depend on torchao
* Docs will be migrated separately

Test Plan:
pytest test/quantization/pt2e_flow

Reviewers:

Subscribers:

Tasks:

Tags:

PR #2048, opened 2025-04-12
Copy and adapt pt2e quantization code to torchao

jerryzh168:
Summary:
First step of dev-discuss.pytorch.org/t/torch-ao-quantization-migration-plan/2810

core logic of pt2e are duplicated, and also ported test here

previous unfinished migration: https://github.com/pytorch/ao/pull/1916

Next:

move meta internal callsites to depend on torchao
Docs will be migrated separately

Test Plan:
pytest test/quantization/pt2e_flow

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/integration/test_integration.py
+++ b/test/integration/test_integration.py
@@ -2038,8 +2038,10 @@ class TestExport(unittest.TestCase):
         self.assertTrue(torch.equal(after_export, ref))
         if api is _int8da_int4w_api:
             targets = [n.target for n in model.graph.nodes]
-            self.assertTrue(torch.ops.quant.choose_qparams_affine.default in targets)
-            self.assertTrue(torch.ops.quant.quantize_affine.default in targets)
+            self.assertTrue(
+                torch.ops.torchao_quant.choose_qparams_affine.default in targets
+            )
+            self.assertTrue(torch.ops.torchao_quant.quantize_affine.default in targets)
             self.assertFalse(torch.ops.aten.narrow.default in targets)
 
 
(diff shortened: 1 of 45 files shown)


drisspg (approved):

andrewor14 (approved):
Hi Jerry, can we add a limited set of tests to show that PT2E in pytorch core vs torchao are numerically equivalent, so users will have more confidence they can migrate without any changes in numerics? We can do this in a follow up PR since this one is already 25k lines

test/integration/test_integration.py
```diff
@@ -2038,8 +2038,8 @@ def forward(self, x):
         self.assertTrue(torch.equal(after_export, ref))
         if api is _int8da_int4w_api:
             targets = [n.target for n in model.graph.nodes]
-            self.assertTrue(torch.ops.quant.choose_qparams_affine.default in targets)
-            self.assertTrue(torch.ops.quant.quantize_affine.default in targets)
+            self.assertTrue(torch.ops.torchao.choose_qparams_affine.default in targets)
+            self.assertTrue(torch.ops.torchao.quantize_affine.default in targets)
```
andrewor14, line 2042: Are these numerically equivalent to the `torch.ops.quant` counterparts? Should we add tests to show that?

andrewor14:
Another question is how do we ensure both copies of PT2E remain the same? E.g. if we fix a bug in PyTorch core PT2E, is there an easy way to remind us to fix the same bug in torchao's PT2E? (and vice versa?)

test/integration/test_integration.py
jerryzh168: yeah the ones in pytorch are copy pasted from torchao so should be the same

jerryzh168:
> Hi Jerry, can we add a limited set of tests to show that PT2E in pytorch core vs torchao are numerically equivalent, so users will have more confidence they can migrate without any changes in numerics? We can do this in a follow up PR since this one is already 25k lines

sure, can do this a bit later

leslie-fang-intel (approved):

PR #2058, opened 2025-04-15
Remove zero_point_domain from quant configs

metascroy:
ZeroPointDomain will be removed from the quant primitives, so this PR removes it from the configs and instead relies on MappingType to convey similar information.

### DIFF
--- a/torchao/dtypes/uintx/packed_linear_int8_dynamic_activation_intx_weight_layout.py
+++ b/torchao/dtypes/uintx/packed_linear_int8_dynamic_activation_intx_weight_layout.py
@@ -156,48 +156,109 @@ class PackedLinearInt8DynamicActivationIntxWeightAQTTensorImpl(AQTTensorImpl):
         zero_point: Optional[torch.Tensor],
         layout: Layout,
         bias: Optional[torch.Tensor] = None,
+        *,
+        validate_inputs: bool = True,
     ):
         assert isinstance(layout, PackedLinearInt8DynamicActivationIntxWeightLayout)
-        assert layout.has_params_set(), "PackedLinearInt8DynamicActivationIntxWeightLayout params must be set before calling from_plain"
         assert layout.target in [
             t for t, _ in _TARGET_AND_STR
         ], f"Unexpected target: {layout.target}"
+        assert layout.has_params_set(), "PackedLinearInt8DynamicActivationIntxWeightLayout params must be set before calling from_plain"
 
-        int_types = [torch.int8, torch.int16, torch.int32, torch.int64]
-
-        n, k = int_data.shape
-        assert int_data.dtype in int_types, f"int_data.dtype must be {int_types}"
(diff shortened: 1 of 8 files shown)


jerryzh168:
please provide some context in the PR summary on the motivation on the change and the summary of the changes in the PR

torchao/experimental/tests/test_int8_dynamic_activation_intx_weight.py
```diff
@@ -370,7 +355,7 @@ def test_export_QDQLayout(self):
             "torch.ops.torchao.choose_qparams_affine.default(input_1, 'ASYMMETRIC', [1, 512], torch.int8, None, None, None, torch.float64, torch.int64)",
```
jerryzh168, line 355: nit: I think the test should probably just be checking the ops unless strictly necessary, otherwise the test might be a bit fragile, there is no guarantee on the names of intermediate values in fx graph

torchao/dtypes/uintx/packed_linear_int8_dynamic_activation_intx_weight_layout.py
```diff
@@ -358,79 +418,35 @@ def make_packed_linear_int8_dynamic_activation_intx_weight_tensor(
...
-            assert zero_point.min().item() >= qmin
-            assert zero_point.max().item() <= qmax
+    has_weight_zeros = True
+    if zero_point is None:
+        has_weight_zeros = False
+    else:
+        zero_point_min = zero_point.min().item()
+        zero_point_max = zero_point.max().item()
+        if zero_point_min == 0 and zero_point_max == 0:
+            has_weight_zeros = False
```
jerryzh168, line 437: nit: might be better to have a util to check for zero_point being all zero and use it everywhere to keep consistent

torchao/dtypes/uintx/packed_linear_int8_dynamic_activation_intx_weight_layout.py
```diff
@@ -156,48 +156,109 @@ def from_plain(
         zero_point: Optional[torch.Tensor],
         layout: Layout,
         bias: Optional[torch.Tensor] = None,
+        *,
+        validate_inputs: bool = True,
```
jerryzh168, line 160: why is this needed? I think ideally should be guaranteed by the choose qparams

jerryzh168 (approved):
looks good, just left some nit comments

torchao/dtypes/uintx/packed_linear_int8_dynamic_activation_intx_weight_layout.py
metascroy: You do not need to go through the quant config to construct a packed tensor.  You can construct one directly with from_plain.

When going from quant_config, it sets validate_inputs = False in from_plain because it's already enforced by construction.  But if someone calls from_plain outside of this, I wanted the inputs to be validated.

PR #2060, opened 2025-04-15
Match QAT prepare and convert numerics exactly for bf16 and fp16

andrewor14:
**Summary:** The previous PR #1964 got this to match for fp32, but there were three additional sources of numerical discrepancies with bf16:

1. QAT asymmetric per token choose qparams diverged from `choose_qparams_affine`, which had simpler logic
2. QAT per token fake quantize cast the input to fp32 before fake quantizing them
3. QAT symmetric per group choose qparams used a hardcoded eps value that did not match `choose_qparams_affine`

These are both resolved in this commit: (1) QAT now uses `choose_qparams_affine` instead of the custom function for asymmetric per token, which is now deleted, (2) QAT no longer casts the input to fp32, and (3) QAT now uses an eps value that corresponds to the input dtype. The result is exact match in numerics between the prepare and convert steps for both fp32, bf16, and fp16.

**Test Plan:**

python test/quantization/test_qat.py -k test_fake_quantize_per_token_vs_convert

### DIFF
--- a/test/quantization/test_qat.py
+++ b/test/quantization/test_qat.py
@@ -12,6 +12,7 @@ import unittest
 
 import torch
 import torch.nn.functional as F
+from parameterized import parameterized
 from torch.ao.quantization.fx._decomposed import quantized_decomposed_lib  # noqa: F401
 
 from torchao import quantize_
@@ -40,7 +41,6 @@ from torchao.quantization.qat.linear import (
     Int8DynActInt4WeightQATLinear,
 )
 from torchao.quantization.qat.utils import (
-    _choose_qparams_per_token_asymmetric,
     _fake_quantize_per_channel_group,
     _fake_quantize_per_token,
     _GenericFakeQuantize,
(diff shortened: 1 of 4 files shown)


test/quantization/test_qat.py
```diff
@@ -1449,7 +1468,16 @@ def test_fake_quantize_per_token_vs_convert(self):
...
         not TORCH_VERSION_AT_LEAST_2_4, "skipping when torch version is 2.4 or lower"
     )
-    def test_qat_8da4w_prepare_vs_convert(self):
+    def test_fake_quantize_per_token_vs_convert_fp32(self):
+        self._test_fake_quantize_per_token_vs_convert(torch.float32)
+
+    @unittest.skipIf(
+        not TORCH_VERSION_AT_LEAST_2_4, "skipping when torch version is 2.4 or lower"
+    )
+    def test_fake_quantize_per_token_vs_convert_bf16(self):
```
jerryzh168, line 1477: nit: do we have float16 as well? also can probably use parametrization for these

jerryzh168 (approved):

jerryzh168:
seems like some relevant tests are failing in CI? https://github.com/pytorch/ao/actions/runs/14568001515/job/40860205831

petrex:
I am seeing `TestQAT.test_qat_8da4w_prepare_vs_convert_*` failing in CI. Shall we skip the tests before fixes landed?

PR #2062, opened 2025-04-16
Fix unwrap_tensor_subclass

metascroy:
unwrap_tensor_subclass fails if called multiple times with error AttributeError: 'UnwrapTensorSubclass' object has no attribute 'rebuild_stack'.

This PR fixes the error by not registering a parametrization on a tensor that has already been unwrapped.

### DIFF
--- a/torchao/utils.py
+++ b/torchao/utils.py
@@ -314,6 +314,8 @@ def unwrap_tensor_subclass(model, filter_fn=None):
             and type(child.weight) is not torch.nn.Parameter
             and isinstance(child.weight, torch.Tensor)
             and issubclass(type(child.weight), torch.Tensor)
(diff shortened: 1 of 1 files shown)


jerryzh168:
do we still need to use `unwrap_tensor_subclass` now?  I thought it was not needed after some pytorch update

jerryzh168 (approved):

jerryzh168:
also please don't forget to add a label for new PRs: http://github.com/pytorch/ao/actions/runs/14482040173/job/40620799537?pr=2062

metascroy:
> do we still need to use `unwrap_tensor_subclass` now? I thought it was not needed after some pytorch update

It does not look like it's working in ET right now, so I'm using unwrap_tensor_subclass for now

PR #2063, opened 2025-04-16
Fix compile issues

metascroy:
When trying to bump the torchao pin in ET, I see compile issues related to the new SDPA ops: https://github.com/pytorch/executorch/pull/10142

I think it relates to not including inline after TORCHAO_ALWAYS_INLINE in header files.  In this PR, I searched for all instances of TORCHAO_ALWAYS_INLINE in torchao and added inline after if it is missing.

### DIFF
--- a/torchao/experimental/kernels/cpu/aarch64/benchmarks/benchmark_bitpacking.cpp
+++ b/torchao/experimental/kernels/cpu/aarch64/benchmarks/benchmark_bitpacking.cpp
@@ -27,7 +27,7 @@ template <
     typename pack_8_values_fn_type,
     typename vec_pack_64_values_fn_type,
     typename vec_pack_128_values_fn_type>
-TORCHAO_ALWAYS_INLINE void pack_uint_odd_bit_values(
+TORCHAO_ALWAYS_INLINE inline void pack_uint_odd_bit_values(
     pack_8_values_fn_type pack_8_values_func,
     vec_pack_64_values_fn_type vec_pack_64_values_func,
     vec_pack_128_values_fn_type vec_pack_128_values_func,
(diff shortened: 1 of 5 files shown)


torchao/experimental/kernels/cpu/aarch64/benchmarks/benchmark_bitpacking.cpp
```diff
@@ -27,7 +27,7 @@ template <
     typename pack_8_values_fn_type,
     typename vec_pack_64_values_fn_type,
     typename vec_pack_128_values_fn_type>
-TORCHAO_ALWAYS_INLINE void pack_uint_odd_bit_values(
+TORCHAO_ALWAYS_INLINE inline void pack_uint_odd_bit_values(
```
kimishpatel, line 30: why do you need to include inline after TORCHAO_ALWAYS_INLINE. Should it not expand into inline?

metascroy: Previously I noticed it was required for internal builds.  One of the compile errors I see in the ET pin bump PR is:

```
  /pytorch/executorch/extension/llm/custom_ops/../../../third-party/ao/torchao/experimental/kernels/cpu/aarch64/matmul/matmul_utils.h:74:28: error: inlining failed in call to ‘always_inline’ ‘void torchao::kernels::cpu::aarch64::quantized_matmul::utils::transpose_4x4(const float32_t*, size_t, float32x4_t (&)[4])’: function body can be overwritten at link time
```

metascroy: @kimishpatel I confirmed this does fix the issue by sshing into the ci machine in ExecuTorch

PR #2064, opened 2025-04-16
[scaled grouped mm] add triton kernels for float8 rowwise quantization with per-group/jagged scales

danielvegamyhre:
Next PR in stack: https://github.com/pytorch/ao/pull/2077

## Motivation
- In the first iteration of the differentiable scaled grouped mm prototype, to perform float8 quantization with per-group/"jagged" scales, we naively iterate through the offsets on the CPU in for loop, launching separate scaling kernels for each subtensor (see [here](https://github.com/pytorch/ao/blob/a61e3024e220753d592cff9ef2d77c104fde1609/torchao/prototype/scaled_grouped_mm/scaled_grouped_mm.py#L253)). This approach requires host-device sync, to loop over the values via `offs.tolist()` on the CPU, which negatively impacts performance.
- Solution: do all of this in a single triton kernel to avoid host-device sync and improve performance.

## Summary

This PR adds 2 Triton kernels:
- `_triton_fp8_row_major_jagged_rowwise_scales`: performs dynamic float8 quantization with row-wise scales for each group (as determined by the offsets), and writes the output in row-major memory layout. This kernel is to be used for quantizing left operand of the scaled grouped mm in the backward pass of the differentiable scaled grouped mm, to compute `grad_weight` (see [here](https://github.com/pytorch/ao/blob/a61e3024e220753d592cff9ef2d77c104fde1609/torchao/prototype/scaled_grouped_mm/scaled_grouped_mm.py#L184-L191)).
- `_triton_fp8_col_major_jagged_colwise_scales`: performs dynamic float8 quantization with row-wise scales *computed over logical columns* for each group (as determined by the offsets), and writes the output in col-major memory layout. This kernel is to be used for quantization of the right operand of the scaled grouped mm in the backward pass of the differentiable scaled grouped mm, to compute `grad_weight` (see [here](https://github.com/pytorch/ao/blob/a61e3024e220753d592cff9ef2d77c104fde1609/torchao/prototype/scaled_grouped_mm/scaled_grouped_mm.py#L192-L197)).

## Test plan
- Added unit tests confirming triton kernels produce same outputs as corresponding pure pytorch implementations.

## Performance
- Adding a benchmark script which compares **compiled torch native code** versus **triton kernels** for the float8 quantization w/ per-group scales. 
- To be clear, this does NOT benchmark the e2e differentiable scaled grouped mm, just the quantization of these 2D tensors with per-group scales.
- The benchmark measures the wall clock time in microseconds to perform [both conversions](https://github.com/pytorch/ao/blob/7fa9c69dc0999023add31d000d4750e0ac2cd799/torchao/prototype/scaled_grouped_mm/scaled_grouped_mm.py#L182-L197) needed for the `grad_weight = grad_output_t @ input` in the backward pass.
- Initial perf numbers are below and show huge speedups, by avoiding the slow host-device sync necessary in the torch native code:
```
input_shape      n_groups  high_precision_dtype      torch_time_us    triton_time_us
-------------  ----------  ----------------------  ---------------  ----------------
(256, 4096)             4  torch.bfloat16                  1109.9             82.133
(256, 4096)             8  torch.bfloat16                  2369.09            84.036
(256, 4096)            16  torch.bfloat16                  4469.31            82.534
(4096, 4096)            4  torch.bfloat16                  1289.95            88.023
(4096, 4096)            8  torch.bfloat16                  2358.04            88.563
(4096, 4096)           16  torch.bfloat16                  4430.14            84.527
(65536, 4096)           4  torch.bfloat16                  7810.48            83.115
(65536, 4096)           8  torch.bfloat16                  8658.81            85.899
(65536, 4096)          16  torch.bfloat16                 10339.6             85.017
```

## Next steps
- Measure e2e differentiable scaled grouped mm perf with vs without these changes
- Integrate the triton kernels into the autograd func
- Do any additional perf improvements as necessary
- Add JaggedFloat8Tensor subclass w/ override for torch._grouped_mm (bf16 op)
- Add user-facing prototype API for converting MoE layers to use float8 training

### DIFF
--- a/torchao/prototype/scaled_grouped_mm/kernels/jagged_float8_scales.py
+++ b/torchao/prototype/scaled_grouped_mm/kernels/jagged_float8_scales.py
@@ -182,7 +182,9 @@ def _triton_fp8_row_major_jagged_rowwise_scales(
             block_row_offs[:, None] * stride_input_row
             + block_col_offs[None, :] * stride_input_col
         )
-        block_mask = (block_row_offs[:, None] < M) & (block_col_offs[None, :] < group_col_end_idx)
+        block_mask = (block_row_offs[:, None] < M) & (
+            block_col_offs[None, :] < group_col_end_idx
+        )
         data = tl.load(input_ptr + block_offs, mask=block_mask, other=0.0).to(
             input_dtype
         )
--- a/torchao/prototype/scaled_grouped_mm/kernels/jagged_float8_scales.py
+++ b/torchao/prototype/scaled_grouped_mm/kernels/jagged_float8_scales.py
@@ -1,0 +1,348 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+Triton kernels for scaling high precision tensors to float8.
+"""
+import itertools
+from typing import Tuple
+
+import torch
+import triton
+import triton.language as tl
+
+from torchao.prototype.scaled_grouped_mm.utils import _is_column_major
+
+EPS = 1e-12
+
+FP8_DTYPE_MAP = {
+    torch.int8: tl.int8,
+    torch.int16: tl.int16,
+    torch.int32: tl.int32,
+    torch.int64: tl.int64,
+    torch.float8_e4m3fn: tl.float8e4nv,
+    torch.float8_e5m2: tl.float8e5,
+    torch.float16: tl.float16,
+    torch.bfloat16: tl.bfloat16,
+    torch.float32: tl.float32,
+    torch.float64: tl.float64,
+}
+
+block_sizes = [128, 256]
+kernel_configs_2D = [
+    triton.Config({"BLOCK_SIZE_ROWS": block_size_rows, "BLOCK_SIZE_COLS": block_size_cols})
+    for block_size_rows in block_sizes
+    for block_size_cols in block_sizes
+]
+
+
+def triton_fp8_row_major_jagged_rowwise_scales(
+    hp_tensor: torch.Tensor,
+    offsets: torch.Tensor,
+    output_dtype: torch.dtype = torch.float8_e4m3fn,
+    round_scales_to_power_of_2: bool = False,
+) -> Tuple[torch.Tensor, torch.Tensor]:
+    """
+    Converts a high precision tensor to a float8 tensor is row-major memory layout,
+    using 'jagged' rowwise scales (i.e., separate scales for each group/subtensor as
+    determined by the offsets).
+
+    Args:
+        - hp_tensor: 2D high precision tensor to be converted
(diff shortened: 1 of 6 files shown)


danielvegamyhre:
cc @drisspg @ngimel I'd appreciate your thoughts/reviews on this when you have time

danielvegamyhre:
also cc @vkuzo for visibility upon return from recharge

torchao/prototype/scaled_grouped_mm/kernels/jagged_float8_scales.py
```diff
@@ -0,0 +1,356 @@
...
+
+import torch
+import triton
+import triton.language as tl
+
+from torchao.prototype.scaled_grouped_mm.utils import _is_column_major
+
+EPS = 1e-12
+
+FP8_DTYPE_MAP = {
```
drisspg, line 23: Nit: I think there is a built in 1

torchao/prototype/scaled_grouped_mm/kernels/jagged_float8_scales.py
```diff
@@ -0,0 +1,356 @@
...
+
+    # determine start and end column idx for this group
+    group_col_start_idx = tl.load(
+        offsets_ptr + offset_idx - 1, mask=offset_idx > 0, other=0
+    )
+    group_col_end_idx = tl.load(offsets_ptr + offset_idx)
+    block_row_offs = block_row_id * BLOCK_SIZE_ROWS + tl.arange(0, BLOCK_SIZE_ROWS)
+
+    # compute rowwise amaxes for this group
+    amax_buffer = tl.zeros((BLOCK_SIZE_ROWS,), dtype=tl.float64)
```
drisspg, line 150: have we kept this in fp64 elsewhere?

drisspg (approved):

torchao/prototype/scaled_grouped_mm/kernels/jagged_float8_scales.py
danielvegamyhre: hmm good point, in the float8 training code the amax is in original precision (bf16/fp32) and we convert the amax to fp64 when calculating the scale: https://github.com/pytorch/ao/blob/0045d885521aac773b89740195ca8c7d78bbc16e/torchao/float8/float8_utils.py#L45

Updated the kernel to do the same here for consistency and verified tests are still passing.

torchao/prototype/scaled_grouped_mm/kernels/jagged_float8_scales.py
danielvegamyhre: I found [this](https://github.com/triton-lang/triton/blob/981e987eed9053b952f81153bc0779c99d8c642e/python/triton/runtime/jit.py#L448), is this what you're referring to? Since it's just dtype string/name -> dtype string/name map there would be some extra steps to use it, not sure if it's worth it

drisspg: yeah no worries, leave as is

PR #2070, opened 2025-04-17
Set use_hqq to True by default for int4 weight only quant

jerryzh168:


HDCharles:
the issue is that things like GPTQ depend on the expected behavior of quantization methods. We can hard code those to set use_hqq = False but I think the assumption is that in general, something like hqq is not the default.

HDCharles (approved):
at minimum the techniques that rely on default behavior need to have this disabled.

jerryzh168:
makes sense, looks like this doesn't affect GPTQ: https://github.com/pytorch/ao/blob/a96eeb1c7d7ba24cf0ccfc105141729acfed22bf/torchao/quantization/GPTQ_MT.py#L705

jerryzh168:
oh nvm, it's also used for Marlin Sparse layout, it feels like we should have a separate layout for hqq

jerryzh168:
OK let's do this a bit later, after we split the HQQ stuff out of AQT

Issue #2082, opened 2025-04-19
Flatten pt2e/pt2e folder


jerryzh168:
Now we copy pasted the pt2e quant to torchao, we want to follow up with
some cleanup, one task is to remove the pt2e/pt2e folder and flatten
the files into the pt2e folder

PR #2083, opened 2025-04-19
[1/n] Move a few utils under pt2e/pt2e/ to pt2e/

jerryzh168:
Summary:
https://github.com/pytorch/ao/issues/2082

Test Plan:
pytest test/quantization/pt2e

Reviewers:

Subscribers:

Tasks:

Tags:


torchao/quantization/pt2e/__init__.py
```diff
@@ -13,15 +22,6 @@
...
-    _allow_exported_model_train_eval as allow_exported_model_train_eval,
-)
-from torchao.quantization.pt2e.pt2e.export_utils import (
-    _move_exported_model_to_eval as move_exported_model_to_eval,
-)
-from torchao.quantization.pt2e.pt2e.export_utils import (
-    _move_exported_model_to_train as move_exported_model_to_train,
-)
 
 from .fake_quantize import (
```
metascroy, line 26: Should we add default_dynamic_fake_quant to top level?

jerryzh168: yes, will do in a separate PR, my plan is to first flatten pt2e/pt2e and then make sure we have everything in top level

metascroy (approved):

torchao/quantization/pt2e/__init__.py
jerryzh168: let me do it in next PR

PR #2085, opened 2025-04-21
Fix numeric mismatches

metascroy:
For embedding, these changes ensure:

* IntXQuantizationAwareTrainingConfig
* IntxWeightOnlyConfig
* Int4WeightOnlyEmbeddingQATQuantizer (prepare / convert)

match for various model/scale dtypes (tested in torchao/experimental/tests/test_embedding_xbit_quantizer.py).

For linear, these changes ensures:
* Int8DynamicActivationIntxWeightConfig
* Int8DynamicActivationInt4WeightConfig
* IntXQuantizationAwareTrainingConfig
* Int8DynActInt4WeightQATQuantizer (prepare / convert)

match for various model/scale dtypes (tested in torchao/experimental/tests/test_int8_dynamic_activation_intx_weight.py).

The exception is the PTQ configs are very different than QAT configs for asymmetric weights (not default setting).  Making these match will require discussion.

dynamic activation quantization has several similar, but slightly different functions used throughout torchao (they generally agree when dtype is torch.float32), but they should probably be unified:
* torchao.utils.per_token_dynamic_quant (used in Int8DynActInt4WeightQATQuantizer's convert)
* torchao.quantization.qat.utils._choose_qparams_per_token_asymmetric (used in QAT configs like IntXQuantizationAwareTrainingConfig)
* torchao.quantization.quant_api._int8_asymm_per_token_quant (used in PTQ configs like Int8DynamicActivationInt4WeightConfig and Int8DynamicActivationIntxWeightConfig)

### DIFF
--- a/torchao/quantization/GPTQ.py
+++ b/torchao/quantization/GPTQ.py
@@ -933,7 +933,10 @@ def linear_forward_8da4w(
     groupsize,
     precision,
 ):
-    x = per_token_dynamic_quant(x, scale_dtype=precision, zero_point_dtype=precision)
+    # to match torchao.quantization.quant_api._int8_asymm_per_token_quant
+    x = per_token_dynamic_quant(
+        x, scale_dtype=torch.float64, zero_point_dtype=torch.int64
+    )
     # TODO: verify and remove following reshape code
     # origin_x_size = x.size()
     # x = x.reshape(-1, origin_x_size[-1])
--- a/torchao/quantization/GPTQ.py
+++ b/torchao/quantization/GPTQ.py
@@ -935,8 +935,11 @@ def linear_forward_8da4w(
 ):
     # to match torchao.quantization.quant_api._int8_asymm_per_token_quant
     x = per_token_dynamic_quant(
-        x, scale_dtype=torch.float64, zero_point_dtype=torch.int64
+        x, scale_dtype=torch.float32, zero_point_dtype=torch.float32
     )
+    # x = per_token_dynamic_quant(
+    #     x, scale_dtype=torch.float64, zero_point_dtype=torch.int64
+    # )
     # TODO: verify and remove following reshape code
     # origin_x_size = x.size()
     # x = x.reshape(-1, origin_x_size[-1])
--- a/torchao/quantization/GPTQ.py
+++ b/torchao/quantization/GPTQ.py
@@ -18,6 +18,7 @@ import torch.nn.functional as F
 from torch.utils._pytree import tree_flatten, tree_unflatten
 
 from torchao.dtypes.utils import is_device
+from torchao.quantization.quant_api import _int8_asymm_per_token_quant
 from torchao.utils import (
     TORCH_VERSION_AT_LEAST_2_3,
     TORCH_VERSION_AT_LEAST_2_6,
@@ -931,13 +932,11 @@ def linear_forward_8da4w(
     zeros,
     out_features,
     groupsize,
-    precision,
+    output_precision,
 ):
     # uses fp32 to match torchao.quantization.quant_api._int8_asymm_per_token_quant
     # and activation_scale_dtype in QAT configs
-    x = per_token_dynamic_quant(
-        x, scale_dtype=torch.float32, zero_point_dtype=torch.float32
-    )
+    x = _int8_asymm_per_token_quant(x).dequantize()
     # TODO: verify and remove following reshape code
     # origin_x_size = x.size()
     # x = x.reshape(-1, origin_x_size[-1])
@@ -957,7 +956,7 @@ def linear_forward_8da4w(
         torch.int8,
         quant_min,
         quant_max,
-        output_dtype=precision,
+        output_dtype=output_precision,
     )
 
     # x = x.to(torch.float16)
--- a/torchao/quantization/GPTQ.py
+++ b/torchao/quantization/GPTQ.py
@@ -933,13 +933,11 @@ def linear_forward_8da4w(
     groupsize,
     precision,
 ):
-    # to match torchao.quantization.quant_api._int8_asymm_per_token_quant
+    # uses fp32 to match torchao.quantization.quant_api._int8_asymm_per_token_quant
+    # and activation_scale_dtype in QAT configs
     x = per_token_dynamic_quant(
         x, scale_dtype=torch.float32, zero_point_dtype=torch.float32
     )
-    # x = per_token_dynamic_quant(
-    #     x, scale_dtype=torch.float64, zero_point_dtype=torch.int64
-    # )
     # TODO: verify and remove following reshape code
     # origin_x_size = x.size()
     # x = x.reshape(-1, origin_x_size[-1])
--- a/torchao/quantization/GPTQ.py
+++ b/torchao/quantization/GPTQ.py
@@ -18,7 +18,6 @@ import torch.nn.functional as F
 from torch.utils._pytree import tree_flatten, tree_unflatten
 
 from torchao.dtypes.utils import is_device
-from torchao.quantization.quant_api import _int8_asymm_per_token_quant
 from torchao.utils import (
     TORCH_VERSION_AT_LEAST_2_3,
     TORCH_VERSION_AT_LEAST_2_6,
@@ -936,7 +935,12 @@ def linear_forward_8da4w(
 ):
     # uses fp32 to match torchao.quantization.quant_api._int8_asymm_per_token_quant
     # and activation_scale_dtype in QAT configs
-    x = _int8_asymm_per_token_quant(x).dequantize()
+    # TODO: in future add ability to specify activation_scale_dtype to PTQ configs
+    # and enable similar change here
+    x = per_token_dynamic_quant(
+        x, scale_dtype=torch.float32, zero_point_dtype=torch.float32
+    )
+
     # TODO: verify and remove following reshape code
     # origin_x_size = x.size()
     # x = x.reshape(-1, origin_x_size[-1])
--- a/torchao/quantization/qat/embedding.py
+++ b/torchao/quantization/qat/embedding.py
@@ -240,6 +240,7 @@ class Int4WeightOnlyEmbeddingQATQuantizer(TwoStepQuantizer):
                     group_size,
                     precision=scale_precision,
                 )
+                zp = zp.to(zero_point_precision)
                 q_weight = _quantized_decomposed_quantize_per_channel_group_wrapper(
                     child.weight,
                     s,
--- a/torchao/quantization/qat/embedding.py
+++ b/torchao/quantization/qat/embedding.py
@@ -235,7 +235,10 @@ class Int4WeightOnlyEmbeddingQATQuantizer(TwoStepQuantizer):
                 # Load weights and qparams into quantized embedding
                 (qmin, qmax) = _get_qmin_qmax(self.bit_width)
                 (s, zp) = get_group_qparams_symmetric(
-                    child.weight, self.bit_width, group_size, precision=scale_precision,
+                    child.weight,
+                    self.bit_width,
+                    group_size,
+                    precision=scale_precision,
                 )
                 q_weight = _quantized_decomposed_quantize_per_channel_group_wrapper(
                     child.weight,
--- a/torchao/quantization/qat/embedding.py
+++ b/torchao/quantization/qat/embedding.py
@@ -177,6 +177,7 @@ class Int4WeightOnlyEmbeddingQATQuantizer(TwoStepQuantizer):
                 scale_precision=self.scale_precision,
                 zero_point_precision=self.zero_point_precision,
                 device=child.weight.device,
+                dtype=child.weight.dtype,
             )
             # In distributed training, the model may be instantiated
             # on the meta device, in which case there is no need to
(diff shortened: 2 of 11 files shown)


andrewor14:
Hi @metascroy, these should already be fixed in https://github.com/pytorch/ao/pull/2060. Can you verify?

metascroy:
> Hi @metascroy, these should already be fixed in #2060. Can you verify?

I think the bugs in the embedding quantizer (casting output to int instead of float) are still there, but let me verify the other one.

metascroy:
> > Hi @metascroy, these should already be fixed in #2060. Can you verify?
> 
> I think the bugs in the embedding quantizer (casting output to int instead of float) are still there, but let me verify the other one.

I tried rebasing on your fixes, and they don't resolve the embedding issue and there is still an issue with linear.  The problem is inconsistency in scale_dtype for activations.  In most places it is fp32, but in some places it is set to scale_dtype from weights.

torchao/quantization/quant_api.py
```diff
@@ -568,14 +568,16 @@ def _int8_asymm_per_token_quant(x: torch.Tensor) -> torch.Tensor:
     """This is defined here instead of local function to support serialization"""
     mapping_type = MappingType.ASYMMETRIC
     target_dtype = torch.int8
+    scale_dtype = torch.float32
```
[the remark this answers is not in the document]
metascroy: This is not used in PT2E code.

On ET side, XNNPACK does not look at the scale dtype when detecting a lowering pattern, although need to make sure it still suceeds on ET pin bump: https://github.com/pytorch/executorch/blob/main/backends/xnnpack/utils/quant_utils.py#L195

I can make the zero_point_dtype = torch.int8.  The danger there is if people don't use torchao's dequant API, they can overflow by subtracting two int8s.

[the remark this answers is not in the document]
kimishpatel: > This is not used in PT2E code.
> 
> On ET side, XNNPACK does not look at the scale dtype when detecting a lowering pattern, although need to make sure it still suceeds on ET pin bump: https://github.com/pytorch/executorch/blob/main/backends/xnnpack/utils/quant_utils.py#L195
> 
> I can make the zero_point_dtype = torch.int8. The danger there is if people don't use torchao's dequant API, they can overflow by subtracting two int8s.

Will the flow every produce et model with ops which belong here https://github.com/pytorch/executorch/blob/main/kernels/quantized/quantized.yaml? If not thats probably fine.

Separately, I imagine we will have to add support for these ops for quantized kv cache as well. Right now it is using the older ops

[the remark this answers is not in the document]
metascroy: No, those ops are all the old quant primitives.

This will show up in ET as a new quant primitive in the torchao namespace.  Currently only XNNPACK can recognize it.

torchao/quantization/qat/embedding.py
```diff
@@ -367,20 +374,24 @@ def __init__(
...
+            [1, self.group_size],
             self.scale,
             self.zero_point,
+            torch.int8,
             qmin,
             qmax,
-            torch.int8,
-            self.group_size,
-            x.dtype,
+            output_dtype=self.output_dtype,
```
andrewor14, line 394: should output dtype be the same as `x.dtype`? Then we don't need a separate arg?

torchao/quantization/qat/embedding.py
```diff
@@ -227,13 +228,17 @@ def _convert_helper(self, module: torch.nn.Module):
...
                 setattr(module, name, quantized_embedding)
 
                 # Load weights and qparams into quantized embedding
                 (qmin, qmax) = _get_qmin_qmax(self.bit_width)
                 (s, zp) = get_group_qparams_symmetric(
-                    child.weight, self.bit_width, group_size
+                    child.weight,
+                    self.bit_width,
+                    group_size,
+                    precision=scale_precision,
```
andrewor14, line 241: we probably need to set the zero point precision separately to be more correct? I think the function doesn't let you set them separately, so maybe we need to do an extra cast after calling it here, like `zp.to(zero_point_precision)`?

torchao/quantization/GPTQ.py
```diff
@@ -933,7 +933,11 @@ def linear_forward_8da4w(
     groupsize,
     precision,
```
andrewor14, line 934: should we rename this to `output_precision` then?

torchao/quantization/qat/linear.py
```diff
@@ -270,7 +273,7 @@ def __init__(
         precision: torch.dtype = torch.float32,
         scales_precision: torch.dtype = torch.float32,
     ) -> None:
-        activation_config = _get_8da4w_activation_config(scales_precision)
+        activation_config = _get_8da4w_activation_config(torch.float32)
```
andrewor14, line 276: Is this change needed? Looks like `scales_precision` already defaults to `torch.float32`?

torchao/quantization/qat/linear.py
```diff
@@ -219,7 +219,10 @@ def _convert_qat_linear_8da4w(self, module: torch.nn.Module):
                 n_bit = 4
                 (qmin, qmax) = _get_qmin_qmax(n_bit)
                 (s, zp) = get_group_qparams_symmetric(
-                    child.weight, n_bit, config.group_size
+                    child.weight,
+                    n_bit,
+                    config.group_size,
+                    precision=config.scale_precision,
```
andrewor14, line 225: same here, probably need to set `zp = zp.to(config.zero_point_precision)` to be more correct

torchao/quantization/GPTQ.py
```diff
@@ -933,7 +933,11 @@ def linear_forward_8da4w(
...
 ):
-    x = per_token_dynamic_quant(x, scale_dtype=precision, zero_point_dtype=precision)
+    # uses fp32 to match torchao.quantization.quant_api._int8_asymm_per_token_quant
+    # and activation_scale_dtype in QAT configs
+    x = per_token_dynamic_quant(
+        x, scale_dtype=torch.float32, zero_point_dtype=torch.float32
```
andrewor14, line 939: Looks like in `_int8_asymm_per_token_quant` you set the zero point to torch.int8, so this is different? A bit confused here

metascroy: That difference doesn't actually matter because the zero point domain is int.  So the zero points will be int8 numbers (which can be represented in fp32).

But maybe it makes sense to just use quant_api._int8_asymm_per_token_quant directly here, rather than use per_token_dynamic_quant?

torchao/quantization/qat/embedding.py
metascroy: For numeric purposes, it doesn't matter because the zero point is an int8 number, which you can represent with floating point number.

But I'll add the extra cast.

torchao/quantization/qat/embedding.py
metascroy: No, this is an embedding op, so x.dtype is an int (the index).  But the weights should not be cast to integers.

This was one of the existing numerical issues in the embedding convert step

torchao/quantization/qat/linear.py
metascroy: In the PTQ routines, scale_precision on the dynamic activation quantization is always FP64 (see _int8_asymm_per_token_quant, https://fburl.com/uf90caon; in this diff I change it to FP32, though.), even if the scale_dtype on the weight or the input_dtype are set to something else.

torchao/quantization/GPTQ.py
kimishpatel: yeah I would prefer us to be consistent in scale and zero point dtype. In fact if we could make scale fp32 and zero point int32 that would be fine. zero point should actually be same as quant dtype

torchao/quantization/qat/embedding.py
kimishpatel: yeah but it would be better to be consistent to have out zero point dtype separately specified. for example if you were to quantize bias to in32 than you would want zero point in int32 although usually it is just zero. 

Besides, if this is for affine quantization than maybe have function name reflect that

torchao/quantization/qat/linear.py
andrewor14: I see, makes sense. We should definitely update the docstring to mention this then, something like

"Note: We hardcode activation scales to use `torch.fp32`, but allow users to specify the weight scales (defaults to `torch.fp32`). To get exact numerical match with `Int8DynamicActivationInt4WeightConfig`, users must use the same dtype for both the weights and the scales. Here `scales_precision` refers specifically to the weight scales only, not the activation scales."

torchao/quantization/GPTQ.py
andrewor14: Yeah agree that we should just use `torch.int8` here for zero points to be consistent, even if it doesn't change the numerics in practice. We probably shouldn't use `_int8_asymm_per_token_quant` here because that would introduce subclasses into a module swap only flow.

torchao/quantization/quant_api.py
```diff
@@ -706,7 +708,6 @@ def _int8_dynamic_activation_int4_weight_transform(
     # weight settings
     block_size = (1, group_size)
     target_dtype = torch.int8
-    eps = torch.finfo(torch.float32).eps
```
andrewor14, line 709: Good catch. I had to do the same in my PR, since the the quant primitives were using `torch.finfo(torch.bfloat16).eps` instead

andrewor14:
Thanks @metascroy, the changes make sense to me. I think what we can do to make it clearer is add additional documentation (maybe to the docstring of Int8DynActInt4WeightQATLinear) to clarify what is needed to make the numerics match end-to-end. We can also add to the QAT README but I can do that separately if you'd like.

torchao/quantization/GPTQ.py
metascroy: I meant to use _int8_asymm_per_token_quant followed be dequantize, e.g., _int8_asymm_per_token_quant(x).dequantize()

torchao/quantization/qat/linear.py
metascroy: > > In the PTQ routines, scale_precision on the dynamic activation quantization is always FP64 (see _int8_asymm_per_token_quant, https://fburl.com/uf90caon; in this diff I change it to FP32, though.), even if the scale_dtype on the weight or the input_dtype are set to something else.
> 
> why is that?

I don't know if there is a great reason for that, it's just how LinearActivationQuantizedTensor works (https://github.com/pytorch/ao/blob/main/torchao/quantization/linear_activation_quantized_tensor.py#L24).

Unlike the weights, torchao does not serialize parameters related to activation quantization right now.  Instead it serializes a function that does the activation quantization (which is opaque).  To support serialization, the function is added to torch.serialization.add_safe_globals.

I think to support specifying information about the activation quantization, we would have to serialize this information in the LinearActivationQuantizedTensor (https://github.com/pytorch/ao/blob/main/torchao/quantization/linear_activation_quantized_tensor.py#L24), or register many variants of _int8_asymm_per_token_quant with different dtypes to torch.serialization.add_safe_globals.  The former probably makes more sense.

cc @jerryzh168

andrewor14:
Hi @metascroy, another thing is I think we can unskip the failing test `test_qat_8da4w_prepare_vs_convert` after this PR. I tried this out here: https://github.com/pytorch/ao/pull/2131. Do you want to incorporate the changes there into this PR?

torchao/quantization/qat/linear.py
jerryzh168: is the question around we hard coded the args with linear activation quantized tensor? we can use https://github.com/pytorch/ao/blob/cdced21f9b87bb69930c9f0fae79fbe3c2ef8bcb/torchao/quantization/quant_api.py#L1712 to support passing around args I think

torchao/quantization/GPTQ.py
andrewor14: > _int8_asymm_per_token_quant(x).dequantize()

This would still introduce AQT into this path right? I think I would prefer to keep these separate. IMO just calling the same primitives with unit tests seems enough

metascroy:
> Hi @metascroy, another thing is I think we can unskip the failing test `test_qat_8da4w_prepare_vs_convert` after this PR. I tried this out here: #2131. Do you want to incorporate the changes there into this PR?

Added your changes to this PR

andrewor14 (approved):

andrewor14:
Looks great, thanks for fixing!

metascroy:
@andrewor14 @jerryzh168 I cannot import this to Phabricator ([D73691281](https://www.internalfb.com/diff/D73691281)).  I've rebased on main multiple times, and always run into the same issue.

jerryzh168:
most likely it's because some inconsistencies between OSS and phabricator code, you can try again after we land the recent diff train diffs I think

metascroy:
> most likely it's because some inconsistencies between OSS and phabricator code, you can try again after we land the recent diff train diffs I think

I guess I'll land in OSS and go on top of the diff train then

kimishpatel:
> I think what we can do to make it clearer is add additional documentation (maybe to the docstring of Int8DynActInt4WeightQATLinear) to clarify what is needed to make the numerics match end-to-end.

If this is referring to matching numerics e2e with QAT numerics post conver, then my mental model is that user should not have to do anything, no?

PR #2091, opened 2025-04-21
PARQ quantizer support for torchao's weight-only configs

lisjin:
This is the first step in supporting `torchao.quantize_` for PARQ trained models. I target only `Int4WeightOnlyConfig` and `IntxWeightOnlyConfig` for now since PARQ does not have activation quantization.

Instead of converting the state (e.g., scale, zero point) from PARQ's existing quantizers to torchao format, I decided to create a new quantizer `UnifTorchaoQuantizer`. This quantizer calls torchao's quantization primitives `choose_qparams_affine`, `quantize_affine`, `dequantize_affine` to ensure parity between the two QAT methods.

@metascroy It would be great if you could check the correctness of how the quantizer in `TestUnifTorchaoQuantizer.test_intx_weight_only` is initialized. I'm not sure if I missed any subtleties with int8.


test/prototype/test_parq.py
```diff
@@ -97,9 +117,96 @@ def test_ternarybit_lsbq_parq_prox(self):
...
+            target_dtype=torch.int32,
+            quant_min=0,
+            quant_max=2**b - 1,
+            eps=1e-6,
+            preserve_zero=False,
+        )
+        self.compare_quantized_models(model, m_ref, quantizer, b)
+
+    @unittest.skipIf(not TORCH_VERSION_AT_LEAST_2_6, "Test only enabled for 2.6+")
+    @unittest.skipIf(_DEVICE == "cpu", "Need GPU available")
```
metascroy, line 176: This should work on CPU in default QDQLayout.

Only certain layouts require CPU.

test/prototype/test_parq.py
```diff
@@ -97,9 +117,96 @@ def test_ternarybit_lsbq_parq_prox(self):
...
+        model = M(m=self.out_dim, n=self.out_dim).to(_DEVICE)
+        model.reset_parameters()
+        m_ref = copy.deepcopy(model)
+
+        config = IntxWeightOnlyConfig(granularity=PerGroup(self.group_size))
+        quantize_(m_ref, config)
+
+        # based off torchao.quantization.quant_api._intx_weight_only_transform
+        b = 8
+        q_dtype = torch.int8
```
metascroy, line 187: Why only int8?  Can we do torch.int1, torch.int2, ..., torch.int8?

metascroy:
@lisjin can you give a little code snippet of our QAT prepare/convert would work for this API?

I'm having trouble following.  Here are some example code snippets from other APIs: https://fb.workplace.com/groups/pytorch.edge2.team/permalink/1186139489308568/

test/prototype/test_parq.py
lisjin: Definitely! I wasn't aware these data types exist but will include them.

test/prototype/test_parq.py
```diff
@@ -53,52 +78,184 @@ def forward(self, x):
...
+        m_ref = copy.deepcopy(model)
+        quantize_(m_ref, Int4WeightOnlyConfig(group_size))
+
+        b = 4
+        quantizer = self.int4_torchao_quantizer()
+        self.compare_quantized_models(model, m_ref, quantizer, b, group_size)
+
+    @unittest.skipIf(not TORCH_VERSION_AT_LEAST_2_6, "Test only enabled for 2.6+")
+    @common_utils.parametrize("b", [2, 3, 4, 8])
+    @common_utils.parametrize("group_size", [32, 512])
```
lisjin, line 191: Added parametrized test cases that loop over (int2, int3, int4, int8) data types.

test/prototype/test_parq.py
```diff
@@ -53,52 +78,184 @@ def forward(self, x):
...
+        for n, module in model.named_modules():
+            if not _is_linear(module):
+                continue
+
+            p_orig = getattr(orig_model, n).weight  # PARQ weight
+            p = module.weight.dequantize()  # PARQ weight after quantize_
+            p_ref = getattr(m_ref, n).weight.dequantize()  # natively quantize_
+
+            self.assertTrue(p_orig.equal(p_ref))
+            self.assertTrue(p.equal(p_ref))
```
lisjin, line 254: @metascroy Lines 241–243 shows how to convert a PARQ quantized model using `quantize_`. Since PARQ stores full-precision "latent" copies of each quantized parameter, I just load those values into the model before calling `quantize_`. Line 254 shows that we match the natively quantized model `m_ref`.

torchao/prototype/parq/optim/quantopt.py
```diff
@@ -106,6 +108,28 @@ def quantize_(
         quants.copy_(Q)
         return q
 
+    @torch.no_grad()
+    def torchao_quantize_(self, model):
```
lisjin, line 112: @metascroy Here is a new optimizer-level method within PARQ that (1) loads the latent full-precision weights and (2) recursively quantizes all modules using `torchao.quantize_`. Just to double-check, will it also work on modules besides `torch.nn.Linear`?

andrewor14:
Hi @lisjin, do you mind adding a code snippet on the main README on what the end-to-end flow would look like? My understanding is you can just replace the `LSBQuantizer` there with your new `UnifTorchaoQuantizer`. Then what happens after training? Do we call `quantize_(model, Int4WeightOnlyConfig)` as before? Would be good to clarify

torchao/prototype/parq/quant/uniform_torchao.py
```diff
@@ -0,0 +1,123 @@
...
+    def __init__(
+        self,
+        symmetric: bool,
+        target_dtype: Optional[torch.dtype] = None,
+        quant_min: Optional[Union[int, float]] = None,
+        quant_max: Optional[Union[int, float]] = None,
+        eps: Optional[float] = None,
+        preserve_zero: bool = True,
+        zero_point_domain: ZeroPointDomain = ZeroPointDomain.NONE,
+        config: Optional[AOBaseConfig] = None,
```
andrewor14, line 40: Hmm, a bit weird that this isn't used inside the class at all. Can we just let the user pass this into `torchao_quantize_` instead?

lisjin:
@andrewor14 Thanks for the feedback—I removed config from UnifTorchaoQuantizer. In the README, I've also added a side-by-side comparison of PARQ vs. torchao prepare and convert steps. After PARQ training, we call `optimizer.torchao_quantize_(model, config)`. Let me know if there's anything missing.

andrewor14:
Looks great, thanks @lisjin! The README is very clear.

One thing I want to discuss is whether we can just use a new `PARQConfig` instead so the PARQ flow looks more like the existing torchao QAT flow. This is current convert flow in the PR now:
```
config = IntXWeightOnlyConfig(weight_dtype=torch.int4, granularity=PerGroup(32))
optimizer.torchao_quantize_(model, config)
```

What do you think about something like this instead?
```
inner_config = IntXWeightOnlyConfig(weight_dtype=torch.int4, granularity=PerGroup(32))
parq_config = PARQConfig(optimizer, inner_config)
quantize_(model, parq_config)
```
Also curious if @metascroy has any thoughts on this

torchao/prototype/parq/optim/quantopt.py
```diff
@@ -106,6 +109,27 @@ def quantize_(
...
+        }
+
+        def inner_quantize_(model):
+            for module in model.children():
+                for param in module.parameters(recurse=False):
+                    if param.data_ptr() in param_set:
+                        quantize_(module, config)
+                        break
+
+                inner_quantize_(module)
```
lisjin, line 129: @andrewor14 I think the `quantize_(model, PARQConfig(optimizer, inner_config))` interface looks cleaner, but I see a potential issue with applying this `inner_quantize_` function. PARQ uses `param_set` to filter which modules to quantize. However, `quantize_` has its own recursive function that quantizes parameters based off `filter_fn` mapping `nn.Module` -> `bool`. I'm not sure how to reconcile the two.

andrewor14: I see, yeah doesn't seem trivial to reconcile these two. Maybe this is OK then

torchao/prototype/parq/optim/quantopt.py
```diff
@@ -106,6 +109,27 @@ def quantize_(
...
+            p.data_ptr()
+            for group in self.regularized_param_groups()
+            for p in group["params"]
+        }
+
+        def inner_quantize_(model):
+            for module in model.children():
+                for param in module.parameters(recurse=False):
+                    if param.data_ptr() in param_set:
+                        quantize_(module, config)
```
andrewor14, line 126: By the way this call itself is also recursive, so I feel this will end up quantizing the same submodules multiple times? Maybe we should express this as part of a filter_fn, something like:

```
def _filter_fn(module: torch.nn.Module) -> bool:
    for param in module.parameters(recurse=False):
        if param.data_ptr() in param_set:
            return True
    return False

quantize_(model, config, _filter_fn)
```

andrewor14 (approved):
Looks good to me other than the recursion comment. @metascroy any other thoughts?

torchao/prototype/parq/optim/quantopt.py
andrewor14: Another idea is:

```
filter_fn = optimizer.get_filter_fn()
quantize_(model, config, filter_fn)
```
where `optimizer.get_filter_fn` returns a function that checks against the `param_set`. Then you don't need `PARQConfig`

andrewor14: Another idea is:

```
filter_fn = optimizer.get_filter_fn()
quantize_(model, config, filter_fn)
```
where `optimizer.get_filter_fn` returns a function that checks against the `param_set`. Then you don't need `PARQConfig`. What do you think?

torchao/prototype/parq/optim/quantopt.py
lisjin: Good point! I'll add this change in

metascroy:
Looks good to me!  Thanks @lisjin!

Can we add an end-to-end test_intx_weight_only_e2e for intx (with various x-values), similar to test_int4_weight_only_e2e?

torchao/prototype/parq/optim/quantopt.py
metascroy: Usually to do non-linear, we need a third filter_fn argument, e.g., for embedding:

```
quantize_(model, IntxWeightOnlyConfig(torch.int4), lambda m, fqn: isinstance(m, torch.nn.Embedding))
```

By default the filter function in quantize_ is torch.nn.Linear, which is why you don't usually see it specified.

torchao/prototype/parq/quant/uniform_torchao.py
```diff
@@ -0,0 +1,135 @@
...
+        target_dtype: Optional[torch.dtype] = None,
+        quant_min: Optional[Union[int, float]] = None,
+        quant_max: Optional[Union[int, float]] = None,
+        eps: Optional[float] = None,
+        preserve_zero: bool = True,
+        zero_point_domain: ZeroPointDomain = ZeroPointDomain.INT,
+    ) -> None:
+        super().__init__(center=False)
+
+        self.mapping_type = (
```
metascroy, line 42: Nit: pass mapping type directly?

torchao/prototype/parq/quant/uniform_torchao.py
```diff
@@ -0,0 +1,135 @@
...
+        self.zero_point_domain = zero_point_domain
+
+    def _init_quant_min_max(self, b: int) -> None:
+        if self.quant_min is None or self.quant_max is None:
+            assert b in _BIT_WIDTH_TO_DTYPE, f"Unsupported bitwidth {b}"
+            self.quant_min, self.quant_max = _DTYPE_TO_QVALUE_BOUNDS[
+                _BIT_WIDTH_TO_DTYPE[b]
+            ]
+        if self.target_dtype is None:
+            self.target_dtype = torch.int8
```
metascroy, line 59: Just make torch.int8 the default in __init__?

test/prototype/test_parq.py
```diff
@@ -53,52 +93,179 @@ def forward(self, x):
...
+            Int4UnifTorchaoQuantizer(),
+            ProxHardQuant(),
+            quant_per_channel=True,
+        )
+        self.compare_parq_convert(model, m_ref, optimizer, config)
+
+    @unittest.skipIf(not TORCH_VERSION_AT_LEAST_2_6, "Test only enabled for 2.6+")
+    @unittest.skipIf(_DEVICE == "cpu", "Need GPU available")
+    @common_utils.parametrize("b", [2, 3, 4, 8])
+    def test_intx_weight_only_e2e(self, b: int = 2, group_size: int = 32):
```
lisjin, line 247: @metascroy Thanks for looking it over! I've added this end-to-end test, along with `mapping_type=MappingType.SYMMETRIC` and `target_dtype=torch.int8` defaults for `UnifTorchaoQuantizer`

test/prototype/test_parq.py
```diff
@@ -53,52 +93,179 @@ def forward(self, x):
...
+        # do not update model weights, just quantize
+        optimizer.zero_grad()
+        optimizer.step()
+
+        orig_model = copy.deepcopy(model)  # save copy of PARQ quantized model
+
+        # equivalent to torchao's convert step
+        model.eval()
+        optimizer.restore_latent_params()
+        quantize_(model, config, filter_fn=optimizer._get_filter_fn(model))
```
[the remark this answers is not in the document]
andrewor14: Thanks! One final nit is if we expect users to call this, we should remove the underscore before `get_filter_fn` since it's now public facing API

torchao/prototype/parq/README.md
```diff
@@ -7,23 +7,104 @@ This library applies QAT without modifying model-level code. It instead interfac
...
+</tr>
+<tr>
+<td>convert</td>
+<td valign="top">
+
+```python
+config = IntXWeightOnlyConfig(
+    weight_dtype=torch.int4, granularity=PerGroup(32)
+)
+optimizer.torchao_quantize_(model, config)
```
andrewor14, line 73: also need to update this

PR #2103, opened 2025-04-22
Update ruff version to 0.11.6

jerryzh168:
Summary:
Use more recent versions so people can have less problems when they just install the latest ruff

Test Plan:
CI

Reviewers:

Subscribers:

Tasks:

Tags:


drisspg (approved):
Maybe give a heads up to some folks since this will cause merge conflicts

jerryzh168:
many pre-existing errors

PR #2119, opened 2025-04-23
Add AOPerModuleConfig

jerryzh168:
Summary

This allows:

* per module config by name, e.g. linear: config1, embedding: config2 skipping config by module name
* specify default config for all supported modules by default (modules that passes the filter function)

```
config1 = Int4WeightOnlyConfig(group_size=32)
config2 = Int8WeightOnlyConfig()
config = AOPerModuleConfig({"_default": config1, "linear2": config2})

config1 = Int4WeightOnlyConfig(group_size=32)
config2 = Int8WeightOnlyConfig()
config = AOPerModuleConfig({"linear1": config1, "linear2": config2})
```
Test Plan:
python test/quantization/test_quant_api.py -k test_ao_per_module_config

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/quantization/test_quant_api.py
+++ b/test/quantization/test_quant_api.py
@@ -973,6 +973,7 @@ class TestQuantFlow(TestCase):
         assert isinstance(model.linear2.weight, AffineQuantizedTensor)
         assert isinstance(model.linear2.weight._layout, PlainLayout)
 
+    @unittest.skipIf(not TORCH_VERSION_AT_LEAST_2_6, "Need torch 2.6+")
     def test_ao_per_module_config_embedding_linear(self):
         weight_dtype = torch.int8
         granularity = PerGroup(8)
--- a/test/quantization/test_quant_api.py
+++ b/test/quantization/test_quant_api.py
@@ -19,6 +19,7 @@ from torch.ao.quantization.quantizer.xnnpack_quantizer import (
     get_symmetric_quantization_config,
 )
 from torch.testing._internal import common_utils
+from torch.testing._internal.common_quantization import TestHelperModules
 from torch.testing._internal.common_utils import TestCase
 
 from torchao import quantize_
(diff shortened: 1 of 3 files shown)


torchao/quantization/quant_api.py
```diff
@@ -307,6 +307,52 @@ def _replace_with_custom_fn_if_matches_filter(
...
+    """
+    A variant of _replace_with_custom_fn_if_matches_filter where replacement_fn takes module name as well
+        ...
+        replacement_fn (Callable[[torch.nn.Module, str], torch.nn.Module]): The function to replace matching modules.
+        ...
+
+    Returns:
+        None
+    """
+    if isinstance(model, Float8Linear):
```
drisspg, line 327: maybe just extract this part and reuse

torchao/quantization/quant_api.py
```diff
@@ -1900,6 +1957,30 @@ def _fpx_weight_only_transform(
     return module
 
 
+@dataclass
+class AOPerModuleConfig(AOBaseConfig):
```
drisspg, line 1961: maybe a quick lil doc block here saying its FQN, etc

torchao/quantization/quant_api.py
```diff
@@ -1900,6 +1957,30 @@ def _fpx_weight_only_transform(
...
+    module_name_to_config: Dict[str, Optional[AOBaseConfig]] = field(
+        default_factory=dict
+    )
+
+
+def _ao_per_module_config_handler(
+    module: torch.nn.Module, module_name: str, config: AOPerModuleConfig
```
drisspg, line 1968: module_fqn ?

torchao/quantization/quant_api.py
```diff
@@ -1900,6 +1957,30 @@ def _fpx_weight_only_transform(
...
 
 
+@dataclass
+class AOPerModuleConfig(AOBaseConfig):
+    module_name_to_config: Dict[str, Optional[AOBaseConfig]] = field(
+        default_factory=dict
+    )
+
+
+def _ao_per_module_config_handler(
```
drisspg, line 1967: acutally what does this do?

drisspg (commented):
I like this approach more

torchao/quantization/quant_api.py
jerryzh168: can't really reuse, in L341 we need to call a different function as well

torchao/quantization/quant_api.py
jerryzh168: so for each module (+ module name), just checks the dictionary mapping from module name / fqn to specific configs, if there is a match, we apply the config in dictionary to the module.

drisspg: Ohh this makes total sense thanks

drisspg (approved):

test/quantization/test_quant_api.py
```diff
@@ -933,6 +938,34 @@ def test_workflow_e2e_numerics(self, config):
         sqnr = compute_error(y_ref, y_q)
         assert sqnr >= 16.5, f"SQNR {sqnr} is too low"
 
+    @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
```
drisspg, line 941: can you add a test for Linear configs + embedding configs

PR #2122, opened 2025-04-24
Fix cuda compile error with bf16

metascroy:
Summary: T222166791

Differential Revision: D73562284

### DIFF
--- a/torchao/csrc/cuda/fp6_llm/utils_parallel_dequant.cuh
+++ b/torchao/csrc/cuda/fp6_llm/utils_parallel_dequant.cuh
@@ -25,9 +25,7 @@
 
 #include <cuda.h>
 #include <cuda_fp16.h>
-#if !defined(__CUDA_ARCH__) || __CUDA_ARCH__ >= 800
 #include <cuda_bf16.h>
-#endif
 #include <cuda_runtime.h>
 
 /*
@@ -70,9 +68,9 @@ constexpr float power_of_two(int n) {
     return (n == 0) ? 1.0f : 2.0f * power_of_two(n - 1);
 }
 
-#if !defined(__CUDA_ARCH__) || __CUDA_ARCH__ >= 800
 template<int EXPONENT, int MANTISSA>
 __device__ __forceinline__ uint32_t MultScale(uint32_t PackedBF16Pair, __nv_bfloat16 Scale) {
+#if __CUDA_ARCH__ >= 800
     constexpr int BIAS_OFFSET = (int(1) << (8-1)) - (int(1) << (EXPONENT-1));
     constexpr float BIAS = power_of_two(BIAS_OFFSET);
     __nv_bfloat16* BF16_1 = reinterpret_cast<__nv_bfloat16*>(&PackedBF16Pair);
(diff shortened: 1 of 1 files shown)


torchao/csrc/cuda/fp6_llm/utils_parallel_dequant.cuh
```diff
@@ -70,9 +70,9 @@ constexpr float power_of_two(int n) {
     return (n == 0) ? 1.0f : 2.0f * power_of_two(n - 1);
 }
 
+#if !defined(__CUDA_ARCH__) || __CUDA_ARCH__ >= 800
```
jerryzh168, line 73: why does this look reversed to me? should this be if defined and >= 800?

metascroy: I thought that was weird as well.  But I see it defined in this way in multiple files in torchao (they could all be wrong).

But what I'm doing in this PR is matching the if/else macro on the import (lines 27-29):

```
#include <cuda_fp16.h>
#if !defined(__CUDA_ARCH__) || __CUDA_ARCH__ >= 800
#include <cuda_bf16.h>
#endif
```

with this function.

__nv_bfloat16 is defined in  cuda_bf16.h, but that is currently guarded by #if !defined(__CUDA_ARCH__) || __CUDA_ARCH__ >= 800.

jerryzh168: I see, this does look weird

jerryzh168: cc @gau-nernst can you take a look, is this intentional?

gau-nernst: What was the original error? (i.e. why is this PR needed?)

`__CUDA_ARCH__` is only defined for device (CUDA) code. `#if !defined(__CUDA_ARCH__) || __CUDA_ARCH__ >= 800` means it is true if it is **host code** OR CUDA code with sm>=80. There are some rules about how we should use it https://docs.nvidia.com/cuda/cuda-c-programming-guide/#cuda-arch, otherwise it will lead to undefined behavior.

For the example you mentioned above, we need to also include `cuda_bf16.h` header in host code, otherwise host code won't have access to BF16 typedef. Similarly, `__global__` functions (CUDA kernels) must be visible in **both** host code and device code (and have the same signature). Hence, in many cases, it's easier to have all the functions defined, and leave the implementation empty if `__CUDA_ARCH__ < xxx`.

metascroy: So rather than this change, I should remove the guards around cuda_bf16.h in lines 28-30, i.e., change:

```
#if !defined(__CUDA_ARCH__) || __CUDA_ARCH__ >= 800
#include <cuda_bf16.h>
#endif
```

to:

```
#include <cuda_bf16.h>
```

Is that right?

metascroy: I went ahead and removed the guards on the header (#if !defined(__CUDA_ARCH__) || __CUDA_ARCH__ >= 800)

metascroy: Any more concerns here @jerryzh168 @gau-nernst?

jerryzh168 (approved):

PR #2128, opened 2025-04-25
[CPU] Enable DA8W4 on CPU

Xia-Weiwen:
**Summary**
This PR enables DA8W4 on CPU.
- It adds a new layout `Int8DynamicActInt4WeightCPULayout` and its implementation
- It adds two custom ops:
  - `da8w4_linear_prepack_cpu` for weight packing
  - `da8w4_linear_cpu` for A8W4 GEMM.
- It adds C++ kernels for the two new custom ops

The feature supports symmetric and asymmetric quantization of activation.

The ops and kernels won't be available unless
- torchao is built from source with `USE_CPP_KERNELS=1` on Linux with an X86 CPU with AVX512.
- torchao is run on Linux with an X86 CPU with AVX512.
- PyTorch version >= 2.7

To get the best performance, one needs a CPU with AMX support.

**Implementation details**
- The weight-packing kernel is implemented with AVX512 intrinsics if available. Otherwise, a reference path is used.
- The GEMM kernel uses `at::cpublas` brgemm utilities from Pytorch core if available.
- In the GEMM kernel, if M is large (>4)
  - if brgemm is available, brgemm is used.
  - otherwise, fallback to reference implementation
- In the GEMM kernel, if M is small (<=4):
  - if AVX512_VNNI is available, the kernel uses AVX512_VNNI intrinsics.
  - otherwise, go to the same path for large M.
- All utilities functions used in the kernel are implemented with AVX512 if available. Otherwise fall back to reference implementation.

**Usage**
```python
quantize_(
    model,
    int8_dynamic_activation_int4_weight(
        group_size=32,  # or 64, 128
        layout=Int8DynamicActInt4WeightCPULayout(),
        act_mapping_type=MappingType.SYMMETRIC,  # or MappingType.ASYMMETRIC
    ),
)
```

**Test plan**
```
pytest test/quantization/test_quant_api.py -k test_8da4w_cpu
```


test/quantization/test_quant_api.py
```diff
@@ -875,6 +875,31 @@ def test_int4wo_cpu(self, dtype, x_dim, use_hqq):
...
+                int8_dynamic_activation_int4_weight(
+                    group_size=32, layout=Int4CPULayout()
+                ),
+            )
+            # ensure the expected op is in the code
+            _, code = torch._inductor.utils.run_and_get_code(
+                torch.compile(m, fullgraph=True, dynamic=True),
+                *example_inputs,
+            )
+            assert "_weight_int4pack_mm_for_cpu" in code[0]
```
jerryzh168, line 900: I remember this op is weight only quant?

Xia-Weiwen: Yes. This op is used here because we don't have an op to compute da8w4 on CPU yet. So it will fallback to explicit dequantization and call of this op.

leslie-fang-intel (approved):

test/quantization/test_quant_api.py
Xia-Weiwen: I have re-implemented the path. Now we use a new layout for da8w4. Thanks.

Xia-Weiwen:
@leslie-fang-intel This PR is updated to use a new layout. Please review again. Thanks.

torchao/dtypes/uintx/int4_cpu_layout.py
```diff
@@ -352,3 +352,134 @@ def _linear_fp_act_uint4_weight_cpu_impl(input_tensor, weight_tensor, bias):
...
+        weight_shape = self.packed_weight.shape
+        N = weight_shape[0]
+        K = weight_shape[1] * 2
+        groups = self.scales.numel() // N
+        group_size = K // groups
+        return (1, group_size)
+
+    def get_plain(self) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+        plain_weight = torch.stack(
+            ((self.packed_weight << 4) >> 4, self.packed_weight >> 4), dim=-1
```
leslie-fang-intel, line 483: Not sure if my understanding is right, but here it seems we didn't completely restore the original weight as the inverse of `from_plain`?

torchao/dtypes/uintx/int4_cpu_layout.py
```diff
@@ -352,3 +352,134 @@ def _linear_fp_act_uint4_weight_cpu_impl(input_tensor, weight_tensor, bias):
...
+        if func is aten.t.default:
+            """we don't need to repack the weight and just rely on external
+            shape being changed and record the status of transpose/no-transpose
+            """
+            transposed = DA8W4CPUAQTTensorImpl(
+                args[0].packed_weight,
+                args[0].scales,
+                args[0].qzeros,
+                not args[0].transposed,
+                args[0]._layout,
```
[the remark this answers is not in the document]
Xia-Weiwen: Thanks for the comment. There is a comment explaining this just above. And this is a copy from the `Int4XPULayout`. We have it here because `Int4XPULayout` has it. And I am afraid there is not a UT for it. And it seems unused when running models. Do you want me to remove it?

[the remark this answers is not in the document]
leslie-fang-intel: yean, how about we remove it at now and add it back with the UT when met the use case.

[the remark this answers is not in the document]
Xia-Weiwen: I have removed it. Thanks.

Xia-Weiwen:
Hi @jerryzh168 Could you please review this PR? Thanks.

Xia-Weiwen:
Hi @jerryzh168 Could you please review this PR? Thanks.

Xia-Weiwen:
Hi @jerryzh168 Could you please review this PR? Thanks.

Xia-Weiwen:
Hi @leslie-fang-intel Please review this PR again. I have also added the kernel code in this PR. It showed reasonable performance in internal benchmarks. Thanks.

torchao/csrc/cpu/da8w4_linear.cpp
```diff
@@ -0,0 +1,746 @@
...
+
+static bool use_cpublas_checked = false;
+static bool use_cpublas = false;
+
+bool da8w4_can_pack_weight() {
+#if defined(CPU_CAPABILITY_AVX512)
+  if (use_cpublas_checked) {
+    return use_cpublas;
+  }
+  use_cpublas = at::native::cpublas::could_pack(at::kByte);
```
leslie-fang-intel: It requires `AMX` to make this check as `True` but in the `setup.py` which only requires `vnni`. I think we should align these.

Xia-Weiwen: Thanks for the comments. We don't need the AMX flag here because the code in Torchao does not depends on AMX. `brgemm` is in torch core, not here. We need AVX512_VNNI because we use AVX512_VNNI intrinsics explicitly here in Torchao.
Another example is the INT8 SDPA implementation, where only AVX512 flag is used because it only has AVX512 code in Torchao and it also uses brgemm from torch core.

Xia-Weiwen:
> Please also describe how we choose different implementations based on the CPU Info.

I have added more details in the description. Thanks.

torchao/csrc/cpu/da8w4_linear.cpp
leslie-fang-intel: May be we can change the name of this function to `da8w4_use_cpublas`? If it means packing the weight to the format for cpublas is needed.

torchao/csrc/cpu/da8w4_linear.cpp
```diff
@@ -0,0 +1,746 @@
...
+    for (int n = 0; n < N / 2; ++n) {
+      int32_t b = (int32_t)B[k * ldb + n];
+      dqB[k * N + n * 2] = (b & 0xf) - qzeros[n];
+      dqB[k * N + n * 2 + 1] = (b >> 4) - qzeros[n];
+    }
+  }
+}
+#endif
+
+#if defined(CPU_CAPABILITY_AVX512_VNNI)
```
leslie-fang-intel, line 292: Will it be a issue if user build the package on platform with VNNI support but run it on legacy platform.

torchao/csrc/cpu/da8w4_linear.cpp
Xia-Weiwen: Thanks for the comment. I have changed it to `cpublas_can_pack`.

torchao/csrc/cpu/da8w4_linear.cpp
Xia-Weiwen: Thanks for the comment. I have added a runtime check.

leslie-fang-intel (approved):

Xia-Weiwen:
Hi @jerryzh168 Could you please review this PR? Thanks. It's changed a lot since your last review.

Xia-Weiwen:
Hi @jerryzh168 Could you please review this PR? Thanks.

torchao/quantization/quant_api.py
```diff
@@ -728,9 +761,17 @@ def _int8_dynamic_activation_int4_weight_transform(
...
 
+    if isinstance(layout, Int8DynamicActInt4WeightCPULayout):
+        # Int8DynamicActInt4WeightCPULayout requires bias to be in float32
+        if module.bias is not None:
+            module.bias = torch.nn.Parameter(module.bias.float(), requires_grad=False)
+
     # input settings
     if act_mapping_type == MappingType.ASYMMETRIC:
-        input_quant_func = _int8_asymm_per_token_quant
+        if isinstance(layout, Int8DynamicActInt4WeightCPULayout):
```
jerryzh168, line 767: nit: is it possible to reuse `Int4CPULayout`? the config already asays int8 dynamic activation so it's a bit redundant to include it again here

torchao/dtypes/uintx/int4_cpu_layout.py
```diff
@@ -350,3 +340,290 @@ def _linear_fp_act_uint4_weight_cpu_impl(input_tensor, weight_tensor, bias):
     if bias is not None:
         y += bias
     return y.to(orig_dtype)
+
+
+@dataclass(frozen=True)
+class Int8DynamicActInt4WeightCPULayout(Layout):
```
jerryzh168: it looks like you can just reuse Int4CPULayout

torchao/dtypes/uintx/int4_cpu_layout.py
```diff
@@ -350,3 +340,290 @@ def _linear_fp_act_uint4_weight_cpu_impl(input_tensor, weight_tensor, bias):
...
+    """Layout class for da8w4 CPU layout for affine quantized tensor"""
+
+    pass
+
+
+@register_layout(Int8DynamicActInt4WeightCPULayout)
+class DA8W4CPUAQTTensorImpl(Int4CPUAQTTensorImpl):
```
jerryzh168: oh I see, OK if you need a separate Impl then makes sense to have a separate layout

torchao/dtypes/uintx/int4_cpu_layout.py
jerryzh168: can you move the layout and impl to a separate file?

torchao/dtypes/uintx/int4_cpu_layout.py
```diff
@@ -350,3 +340,290 @@ def _linear_fp_act_uint4_weight_cpu_impl(input_tensor, weight_tensor, bias):
...
+        int_data: torch.Tensor,
+        scale: torch.Tensor,
+        zero_point: torch.Tensor,
+        _layout: Layout,
+    ):
+        assert isinstance(_layout, Int8DynamicActInt4WeightCPULayout)
+        assert int_data.dtype == torch.int8, "DA8W4 CPU: expects uint8 weight"
+        assert int_data.shape[1] % 2 == 0, "DA8W4 CPU: expects even number of columns"
+        # int8 -> uint8
+        int_data = (int_data + 8).to(torch.uint8)
```
jerryzh168: why not just have uint8 data directly from the quant primitive ops?

torchao/dtypes/uintx/int4_cpu_layout.py
```diff
@@ -350,3 +340,290 @@ def _linear_fp_act_uint4_weight_cpu_impl(input_tensor, weight_tensor, bias):
...
+        if scale.dim() == 1:
+            scale.unsqueeze_(-1)
+        scale = scale.to(torch.float)
+        if zero_point.dim() == 1:
+            zero_point.unsqueeze_(-1)
+        zero_point = zero_point.to(torch.int8) + 8
```
jerryzh168: can you configure dtypes of int_data, scale, zero_point and shapes in the call to `to_affine_quantized_intx`?

torchao/dtypes/uintx/int4_cpu_layout.py
Xia-Weiwen: Yes. We need a different impl from W16W4 because the ISA (AMX and VNNI) requires different memory formats of weight for computation in BF16 or INT8. Thanks.

torchao/quantization/quant_api.py
Xia-Weiwen: Since it needs a separate impl, a new layout is needed. Thanks.

torchao/dtypes/uintx/int4_cpu_layout.py
Xia-Weiwen: Sure. Done.

torchao/dtypes/uintx/int4_cpu_layout.py
Xia-Weiwen: Thanks for the suggestion. I have improved this part.

test/quantization/test_quant_api.py
```diff
@@ -695,6 +697,72 @@ def test_int4wo_cpu(self, dtype, x_dim, use_hqq):
...
+            )
+            y, code = torch._inductor.utils.run_and_get_code(
+                torch.compile(m, fullgraph=True, dynamic=True),
+                *example_inputs,
+            )
+            # ensure the expected op is in the code
+            assert "torch.ops.torchao.da8w4_linear_cpu.default" in code[0]
+            quantize_(
+                m2,
+                int8_dynamic_activation_int4_weight(
```
jerryzh168, line 742: nit: can you use the new API: `Int8DynamicActivationInt4WeightConfig` instead of `int8_dynamic_activation_int4_weight`?

torchao/quantization/quant_api.py
```diff
@@ -728,9 +761,17 @@ def _int8_dynamic_activation_int4_weight_transform(
     quant_min = -8
     quant_max = 7
 
+    if isinstance(layout, Int8DynamicActInt4WeightCPULayout):
```
jerryzh168: can this happen in kernel? we have dtype conversions like this: https://github.com/pytorch/ao/blob/28989031568710ec6739eb901708eb392782cfd1/torchao/dtypes/uintx/plain_layout.py#L260

jerryzh168 (approved):

torchao/quantization/quant_api.py
Xia-Weiwen: Thanks for the comment. I have moved this to `_linear_int8_act_int4_weight_cpu_impl`.

test/quantization/test_quant_api.py
Xia-Weiwen: Thanks. Done.

PR #2131, opened 2025-04-25
Unskip test_qat_8da4w_prepare_vs_convert

andrewor14:
Following @metascroy's investigation in #2085, we can unskip this test, which was caused by activation scales having different precisions between prepare and convert.

**Test Plan:**


petrex:
thanks @andrewor14 let me test in on rocm as well.

andrewor14:
Seems like it's still failing, need to investigate further

andrewor14:
Ok I think I fixed it for fp16 as well. @petrex try again?

andrewor14:
Unskipped as part of #2085

PR #2140, opened 2025-04-28
[PT2E][X86] Migrate fusion passes in Inductor to torchao

Xia-Weiwen:
### Summary
In this PR, we migrate the fusion passes of quantized ops for X86Inductor backend from PyTorch Inductor source code to Torchao. This is the first step to migrate quantization-related fusion passes in PyTorch core to Torchao.
**With this PR landed, we can add fusion passes for new ops in Torchao instead of in PyTorch core. So, we want this PR merged early.**
We plan to do the migration in the following steps:
1. Copy fusion passes from PyTorch core to Torchao (this PR)
2. Depreacte and remove fusion passes in PyTorch core (TODO)
3. Switch to new quantize/dequantize ops in Torchao (TODO)

(Step 2 and 3 have no dependency on each other and can be reordered.)

Fusion passes need to be registered to Inductor before calling torch.compile. And it would be less user-friendly if we ask users to register them in their code. There are two options for automatic registration:
1. Wrap registration in the lowering API `lower_pt2e_quantized_to_x86`
2. Register passes when importing `torchao.quantization.pt2e.quantizer.x86_inductor_quantizer`

The problem with option 1 is that we should allow registering other fusion passes out of tree. So, we decide to go with option 2.

### Test plan
We copied related UTs from https://github.com/pytorch/pytorch/blob/main/test/inductor/test_mkldnn_pattern_matcher.py
The test cases are run only with torch nightly since some torch features are only available in nightly, such as `onednn.qconv_pointwise`.
Use the following cmd to run tests
```
pytest test/quantization/pt2e/test_x86inductor_fusion.py
```

### Explanation of implementation

- In this PR, we mostly copy the code from torch Inductor https://github.com/pytorch/pytorch/blob/main/torch/_inductor/fx_passes/quantization.py, using internal functions, methods and utilities in Inductor by importing them directly. We think it's the simplest way to register the fusion passes.
- For now, the fusion passes in torch Inductor will co-exist with the passes registered in torchao. There won't be an issue because duplicate passes won't be applied twice. It's because the patterns no longer exist in graph after fusion, and one pattern won't be matched twice.
- In the future, we will switch to the new quantize/dequantize ops in torchao when they are ready. At that time, the patterns registered in torchao will be different from those in Inductor. After that, the passes in torch Inductor will be deprecated and eventually removed.


Xia-Weiwen:
Hi @jerryzh168 @jansel Could you please review this PR? We would like to hear your comments especially on (1) if it sounds ok to you that we copy Inductor code here in Torchao with Inductor's internal utilities, (2) if it is ok that we keep duplicate passes for now. Thanks!

jansel:
I think out of tree passes are fine.

Do we need a better registration system so the changes can be local to a specific torch.compile() call rather than mutating globals?

cc @eellison

jerryzh168 (approved):
should the passes live in inductor or under

jerryzh168:
I also feel hiding compile API in `lower_pt2e_quantized_to_x86` is not a good idea and compile stack should allow registering fusion passes out of tree

torchao/quantization/pt2e/lowering.py
```diff
@@ -12,21 +14,48 @@
...
+            import torch._inductor.config
+
+            from torchao.prototype.inductor.fx_passes.quantization import (
+                _register_quantization_weight_pack_pass,
+                quant_lift_up,
+            )
+
+            torch._inductor.config.pre_grad_custom_pass = quant_lift_up
+            _register_quantization_weight_pack_pass()
+            FUSION_PATH_REGISTERED = True
```
jerryzh168, line 52: can this part happen during import of `x86_inductor_quantizer`?

Xia-Weiwen:
> I think out of tree passes are fine.
> 
> Do we need a better registration system so the changes can be local to a specific torch.compile() call rather than mutating globals?
> 
> cc @eellison

Thanks for your comments. We will just keep the current implementation then.
As for a new registration system, maybe we can have something similar to `pre_grad_custom_pass`?

jansel:
Yeah, I think this might be cleaner with something like `pre_grad_custom_pass` instead of global registration.

leslie-fang-intel (commented):
Hi @Xia-Weiwen, will you add the registration system in PyTorch firstly then refine this PR?

Xia-Weiwen:
> Hi @Xia-Weiwen, will you add the registration system in PyTorch firstly then refine this PR?

No. I plan to keep the current implementation. When new registration system is added in Inductor by Meta Inductor team, I will switch to that in another PR.

torchao/quantization/pt2e/lowering.py
```diff
@@ -12,21 +14,48 @@
...
+            global torch
+            import torch._inductor.config
+
+            from torchao.prototype.inductor.fx_passes.quantization import (
+                _register_quantization_weight_pack_pass,
+                quant_lift_up,
+            )
+
+            torch._inductor.config.pre_grad_custom_pass = quant_lift_up
+            _register_quantization_weight_pack_pass()
```
leslie-fang-intel, line 51: I'm a bit concerned about this. Not sure how we should handle it, but it seems like it need to be take care

- The patterns from `quantization.py in TorchAO` will be registered here once
- And inside `torch.compile` when freezing turns on, the same patterns from `pytorch/torch/_inductor/fx_passes/quantization.py in Torch Core` will be registered again.

torchao/quantization/pt2e/lowering.py
```diff
@@ -12,21 +14,48 @@
...
+        if not FUSION_PATH_REGISTERED:
+            global torch
+            import torch._inductor.config
+
+            from torchao.prototype.inductor.fx_passes.quantization import (
+                _register_quantization_weight_pack_pass,
+                quant_lift_up,
+            )
+
+            torch._inductor.config.pre_grad_custom_pass = quant_lift_up
```
leslie-fang-intel, line 50: Be careful to check if there is any other `pre_grad_custom_pass` registered before, check this https://github.com/pytorch/pytorch/issues/151876 cc @Valentine233 who are working on it.

torchao/prototype/inductor/fx_passes/quantization.py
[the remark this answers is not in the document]
Xia-Weiwen: Thanks. I have moved it as you suggested.

torchao/quantization/pt2e/lowering.py
Xia-Weiwen: Thanks. I have changed per your suggestion.

torchao/quantization/pt2e/lowering.py
Xia-Weiwen: Thanks for the comments. As we discussed offline and I have explained in the summary above, duplicate passes will be applied only once because once applied, the pattern is gone.

torchao/quantization/pt2e/lowering.py
Xia-Weiwen: Thanks for the comment. It is potentially unsafe. I will modify this part after the `pre_grad_custom_pass` is refactored.

Xia-Weiwen:
> I also feel hiding compile API in `lower_pt2e_quantized_to_x86` is not a good idea and compile stack should allow registering fusion passes out of tree

Thanks for the comments. I have moved the registration out of the lowering function. Please review again. Thanks.

Xia-Weiwen:
Hi @jerryzh168 Could you please review this PR again? Thanks

Xia-Weiwen:
Hi @jerryzh168 Do you have any concerns about this PR? Thanks.

torchao/quantization/pt2e/quantizer/x86_inductor_quantizer.py
```diff
@@ -1625,3 +1625,17 @@ def _same_shape(n1: Node, n2: Node):
...
+import torch._inductor.config
+
+from torchao.quantization.pt2e.inductor_passes.x86 import (
+    _register_quantization_weight_pack_pass,
+    quant_lift_up,
+)
+from torchao.utils import TORCH_VERSION_AT_LEAST_2_8
+
+if TORCH_VERSION_AT_LEAST_2_8:
+    torch._inductor.config.pre_grad_custom_pass = quant_lift_up
```
jerryzh168, line 1640: just confirming, is this the final API already or do we still expect more changes here?

jerryzh168 (approved):

torchao/quantization/pt2e/quantizer/x86_inductor_quantizer.py
Xia-Weiwen: I think this is subject to change. Discussions are still underway. We plan to modify this part after the API is changed. Thanks.

PR #2154, opened 2025-05-01
Implement dtensor.shard_dim_alltoall, aten.contiguous, aten.chunk

nathan-az:
These three ops threw NotImplemented errors when attempting to use TP with FP8 in torchtune. Feel free to make corrections or take this into another PR. 

Note that this was done rather blindly, but relieved the errors and appears to get near identical results to BF16 TP training, but may not be correct. Specifically for the `shard_dim_alltoall` - I'm not sure whether this needs type checking and only works on tensorwise-sharded tensors, like the other collectives (e.g. `allgather_fp8`).

image

Example run above shows quite similar losses. 

Certainly needs a review from someone more familiar with the underlying operations. CC @andrewor14 @vkuzo 

Related to, but does not resolve https://github.com/pytorch/torchtune/issues/2629 as it still does not play nicely with `compile` due to mismatching guessed metadata. I have not been able to resolve this.


danielvegamyhre:
>These three ops threw NotImplemented errors when attempting to use TP with FP8 in torchtune. 

@nathan-az to clarify, are you using tensorwise or rowwise scales for TP + FP8 in torchtune?

nathan-az:
tensorwise @danielvegamyhre

andrewor14:
@danielvegamyhre do you know how TP + FP8 + compile works in torchtitan? We have not been able to get this working in torchtune for some reason, wondering if the setup is somehow different. In torchtune we do [compile -> convert_to_float8_training -> apply TP](https://github.com/pytorch/torchtune/blob/1823abd22558a1e74228bf08e21dccbbca33f661/recipes/full_finetune_distributed.py#L575), and we skip the output layer for compile and float8. The TP plans we tried live [here](https://github.com/pytorch/torchtune/blob/1823abd22558a1e74228bf08e21dccbbca33f661/torchtune/models/llama3/_parallelism.py#L27). Seems like it's failing with `aten.chunk NotImplementedError` even with regular non-float8 TP plans. Is it set up differently in torchtitan somehow?

danielvegamyhre:
> @danielvegamyhre do you know how TP + FP8 + compile works in torchtitan? We have not been able to get this working in torchtune for some reason, wondering if the setup is somehow different. In torchtune we do [compile -> convert_to_float8_training -> apply TP](https://github.com/pytorch/torchtune/blob/1823abd22558a1e74228bf08e21dccbbca33f661/recipes/full_finetune_distributed.py#L575), and we skip the output layer for compile and float8. The TP plans we tried live [here](https://github.com/pytorch/torchtune/blob/1823abd22558a1e74228bf08e21dccbbca33f661/torchtune/models/llama3/_parallelism.py#L27). Seems like it's failing with `aten.chunk NotImplementedError` even with regular non-float8 TP plans. Is it set up differently in torchtitan somehow?

Yes, I've actually been looking into it more lately to implement a PoC fp8 rowwise all-gather for TP, which is needed to support fusing all-gather-matmuls in async TP (https://github.com/pytorch/pytorch/issues/149990). 

In torchtitan certain layers are parallelized with standard bf16 parallelism primitives ([code](https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/llama3/parallelize_llama.py#L135-L150)).

If - and only if - float8 w/ tensorwise scaling is used, we parallellize w/ the special float8 TP parallelism primitives to do all-gathers in fp8 w/ tensorwise scales ([code](https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/llama3/parallelize_llama.py#L154-L197). 

In torchtune for fp8 training/fine-tuning it looks like you're unconditionally using the float8 TP parallelism primitives for everything ([link](https://github.com/pytorch/torchtune/blob/1823abd22558a1e74228bf08e21dccbbca33f661/torchtune/models/llama3/_parallelism.py#L65)), which is different than torchtitan. As a starting point I'd suggest trying to mirror the torchtitan TP parallelism plan, which would work.

nathan-az:
2 notes here:

1. Is there any opposition to supporting the operations implemented in this PR (or are they implemented incorrectly / in a way that only works for tensorwise operations that needs changing)? I ask because this appears to unblock the non-compile training. I'm happy to abandon if this PR is overall problematic, or make changes if there are limitations/gates we should put in place such as checking scaling type.
2. @danielvegamyhre it looks to me like `torchtune`'s parallelisation is actually identical to the `torchtitan` tensorwise-case you outlined, in which the standard TP primitives are used for token embeddings and output, with the fp8 classes being used only within the layers ([reference](https://github.com/pytorch/torchtune/blob/1823abd22558a1e74228bf08e21dccbbca33f661/torchtune/models/llama3/_parallelism.py#L36)).

andrewor14:
> In torchtune for fp8 training/fine-tuning it looks like you're unconditionally using the float8 TP parallelism primitives for everything ([link](https://github.com/pytorch/torchtune/blob/1823abd22558a1e74228bf08e21dccbbca33f661/torchtune/models/llama3/_parallelism.py#L65)), which is different than torchtitan. As a starting point I'd suggest trying to mirror the torchtitan TP parallelism plan, which would work.

Thanks. Also just to clarify, torchtune FP8 + TP also fails with the regular TP plan ([this one](https://github.com/pytorch/torchtune/blob/1823abd22558a1e74228bf08e21dccbbca33f661/torchtune/models/llama3/_parallelism.py#L63), with regular ColwiseParallel etc.). Any ideas why?

danielvegamyhre:
>@danielvegamyhre it looks to me like torchtune's parallelisation is actually identical to the torchtitan tensorwise-case you outlined, in which the standard TP primitives are used for token embeddings and output, with the fp8 classes being used only within the layers ([reference](https://github.com/pytorch/torchtune/blob/1823abd22558a1e74228bf08e21dccbbca33f661/torchtune/models/llama3/_parallelism.py#L36)).

Ah, I misread that - thanks. In any case, I'm confused why a `shard_dim_all_to_all` is required for float8 TP in torchtune but not with torchtitan? Looking at the TP parallelization plan in torchtitan and torchtune, they both seem to use a TP + SequenceParallel strategy very similar to the one described here: https://arxiv.org/pdf/2205.05198, and in this strategy there should only be all-gathers and reduce-scatters (in theory)... @tianyu-l do you have any thoughts on this?

nathan-az:
**Config**:

<details>

```yaml
batch_size: 1
checkpointer:
  _component_: torchtune.training.FullModelHFCheckpointer
  checkpoint_files:
    filename_format: model-{}-of-{}.safetensors
    max_filename: '00004'
  model_type: LLAMA3
  output_dir: ${output_dir}
  recipe_checkpoint: null
  checkpoint_dir: inputs/model
clip_grad_norm: 100.0
compile: false
custom_sharded_layers: []

data_parallel_shard_dim: 1
data_parallel_replicate_dim: 1
tensor_parallel_dim: 4

enable_fp8_training: true
fp8_recipe_name: tensorwise
tensor_parallel_plan:
  _component_: torchtune.models.llama3._parallelism._fp8_llama_tp_plan

dataset:
  _component_: torchtune.datasets.alpaca_dataset
  packed: True  # True increases speed
  train_on_input: true
device: cuda
dtype: bf16
enable_activation_checkpointing: true
enable_activation_offloading: false
epochs: 1
fsdp_cpu_offload: false
gradient_accumulation_steps: 4
log_every_n_steps: 1
log_level: INFO
log_peak_memory_stats: true
loss:
  _component_: torchtune.modules.loss.LinearCrossEntropyLoss
  mask_pre_projection: False
max_steps_per_epoch: 10

metric_logger:
  _component_: torchtune.training.metric_logging.MLFlowLogger
  experiment_name: fp8_tp_debug_tensorwise
  run_name: base_tp_no_opt

optimizer:
  _component_: torchao.optim.AdamW8bit
  lr: 4.0e-05
optimizer_in_bwd: false
output_dir: outputs
resume_from_checkpoint: false
seed: 100
shuffle: true
tokenizer:
  max_seq_len: 1024
  path: inputs/model/original/tokenizer.model
  _component_: torchtune.models.llama3.llama3_tokenizer
model:
  _component_: torchtune.models.llama3_1.llama3_1_8b
```

</details>

Above is the config I'm testing with, branched from `ao` main and `torchtune` main. Small modifications I made to `torchtune` - 

* un-gate FP8 + TP
* Fix TP plan to take (but not change) `model` as an arg
* Force `max-autotune` when compiling flex attention to work with my hardware (L40s).
* Distributing `labels` in the `linear_loss` case to fix mixed DTensor/Tensor in TP operations

I then ran to see which errors I got, implementing one at a time.

1. NotImplementedError: attempting to run _dtensor.shard_dim_alltoall.default
2. NotImplementedError: attempting to run aten.contiguous.default
3. NotImplementedError: attempting to run aten.chunk.default

I resolved one at a time, and this was the order they came (although due to the async nature of CUDA I'm not sure how meaningful this order is). Let me know if I can provide any other information.

joecummings:
@andrewor14  @danielvegamyhre Any update on this?

nathan-az:
Closing this - https://github.com/pytorch/torchtune/pull/2761 stopped these errors from raising.

Looks like there are still issues with FP8 + TP + compile (testing with llama 3.3). Would be great to know if anyone is looking into this.

Issue #1621, opened 2025-01-26
Unittests Migration Progress


osbm:
Source: #1321 
I thought opening small pull requests would be easier for me to manage and easier for you to review. Gradual but controlled process to simplify current testing library situation. I love this package. Thank you all for your efforts. If i can be a little of help with this i would be very happy.
- dora
- [ ] test_dora_fusion.py
- [ ] test_dora_layer.py
- dtypes
- [ ] test_affine_quantized_float.py
- [x] test_affine_quantized.py
- [x] test_affine_quantized_tensor_parallel.py
- [ ] test_bitnet.py
- [ ] test_bitpacking.py
- [x] test_floatx.py
- [ ] test_nf4.py
- [ ] test_uint2.py
- [x] test_uint4.py
- [ ] test_uintx.py
- float8
- [ ] test_base.py
- [ ] test_compile.py
- [ ] test_dtensor.py
- [ ] test_dtensor.sh
- [ ] test_everything.sh

- [ ] test_fsdp2_tp.py
- [ ] test_fsdp_compile.py
- [ ] test_fsdp_compile.sh
- [ ] test_fsdp.py
- [ ] test_fsdp.sh
- [ ] test_numerics_integration.py
- float8/test_fsdp2
- [ ] test_fsdp2.py
- galore
- [ ] memory_analysis_utils.py
- [ ] model_configs.py
- [ ] profile_memory_usage.py
- [ ] profiling_utils.py
- [ ] README.md
- hqq
- [x] test_hqq_affine.py
- [ ] test_triton_mm.py
- [ ] test_triton_qkv_fused.py
- integration
- [x] test_integration.py
- kernel
- [ ] galore_test_utils.py
- [x] test_autotuner.py
- [ ] test_fused_kernels.py
- [ ] test_galore_downproj.py
- prototype
- [ ] test_autoround.py
- [ ] test_awq.py
- [ ] test_bitpacking_gen.py
- [x] test_codebook_quant.py
- [ ] test_device_spec.py
- [ ] test_low_bit_optim.py
- [x] test_mixed_precision.py
- [x] test_parametrization.py
- [ ] test_performance_counter.py
- [ ] test_quantized_training.py
- [x] test_scheduler.py
- [ ] test_smoothquant.py
- [x] test_sparse_api.py
- [x] test_sparsifier.py
- [x] test_sparsity_utils.py
- [ ] test_spinquant.py
- [x] test_splitk.py
- [x] test_structured_sparsifier.py
[...]

PR #2158, opened 2025-05-01
Removes pinned version for pytest

syed-ahmed:
Unless 7.4.0 is actually required, we should remove this so that pip is able to resolve dependencies in other environments:
```
ERROR: Cannot install pytest==7.4.0 because these package versions have conflicting dependencies.                                                                                                    
                                                                                                                                                                                                     
The conflict is caused by:                                                                                                                                                                           
    The user requested pytest==7.4.0                                                                                                                                                                 
    The user requested (constraint) pytest==8.1.1                                                                                                                                                    

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip to attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
```

### DIFF
--- a/.github/workflows/regression_test.yml
+++ b/.github/workflows/regression_test.yml
@@ -64,32 +64,38 @@ jobs:
             torch-spec: 'torch==2.5.1 --index-url https://download.pytorch.org/whl/cu121'
             gpu-arch-type: "cuda"
             gpu-arch-version: "12.6"
+            dev-requirements-overrides: "s/^pytest$/pytest==7.4.0/"
           - name: CUDA 2.6
             runs-on: linux.g5.12xlarge.nvidia.gpu
             torch-spec: 'torch==2.6.0'
             gpu-arch-type: "cuda"
             gpu-arch-version: "12.6"
+            dev-requirements-overrides: ""
           - name: CUDA 2.7
             runs-on: linux.g5.12xlarge.nvidia.gpu
             torch-spec: 'torch==2.7.0'
             gpu-arch-type: "cuda"
             gpu-arch-version: "12.6"
+            dev-requirements-overrides: ""
 
           - name: CPU 2.5.1
             runs-on: linux.4xlarge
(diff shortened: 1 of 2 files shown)


jerryzh168:
seems to be OK to me, cc @msaroufim @drisspg @atalman ?

msaroufim:
IIRC we wanted to get rid of pytest entirely, might be time to accelerate the transition?

jerryzh168:
@msaroufim I see, then we'll need to finish this one: https://github.com/pytorch/ao/issues/1621

drisspg:
We use pretty minimal set of the PyTest features so I think it is this is fine for now, unless the failures in this PR are in fact caused by unpinning

jerryzh168:
CI errors might be related I think, @syed-ahmed can you confirm

syed-ahmed:
@jerryzh168 Yeah, looks like the errors are real. It's happening because pytest does have a new requirement: https://github.com/pytorch/pytorch/issues/127517. Looks like we workaround it: https://github.com/pytorch/pytorch/pull/136158 and that's why the CI passes for PyTorch 2.6 and 2.7. 

I've added a `dev-requirements-overrides: "s/^pytest$/pytest==7.4.0/"` in the github workflow of the 2.5 tests, and we can remove it after the next PyTorch release. Is that ok?

msaroufim (approved):

msaroufim:
@metascroy @manuelcandales seems like the MPS jobs are flaking

msaroufim:
Actually I messed up merging this, I  didn't realize sentencepiece has some test utilities at build time

msaroufim:
Actually I did not mess up seems like it's a flake and occurs even if i downgrade pytest again https://github.com/pytorch/ao/pull/2166

Main is healthy

Issue #2157, opened 2025-05-01
nightly build for mac stops on 0422


guangy10:
For example running `pip install torchao==0.11.0.dev20250430 --extra-index-url "https://download.pytorch.org/whl/nightly/cpu"` will not work on mac. The latest build available on mac was from 0422.

guangy10:
cc: @jerryzh168

atalman:
Something changes on 0419

guangy10:
@atalman Is this fixed? Running `pip install torchao==0.11.0.dev20250506 --extra-index-url "https://download.pytorch.org/whl/nightly/cpu"` on mac will still end up with the same issue:

```
Looking in indexes: https://pypi.org/simple, https://download.pytorch.org/whl/nightly/cpu
[...]

PR #2161, opened 2025-05-01
[testing]Triaging ROCm wheel build

petrex:
Triage ROCm CI build error that's not reproducible locally

- disable sparse marlin kernel build for ROCm wheel


petrex:
@amdfaa @jeffdaily

petrex:
cross compilation should and expect to work. need to look else where.

petrex:
related to #2157 #2170

PR #2162, opened 2025-05-02
Update QAT README.md

### LINKED ISSUE #2155: QAT docs
In the [QAT README](https://github.com/pytorch/ao/tree/main/torchao/quantization/qat#quantize_-api-recommended), there is a suggestion:

---
To fake quantize embedding in addition to linear, you can additionally call the following with a filter function during the prepare step:
```
quantize_(
    m,
    IntXQuantizationAwareTrainingConfig(weight_config=weight_config),
    filter_fn=lambda m, _: isinstance(m, torch.nn.Embedding),
)
```
---

Would the correct approach not be as follows?
```
from torchao.quantization.quant_api import _is_linear
quantize_(
    m,
    IntXQuantizationAwareTrainingConfig(weight_config=weight_config),
    filter_fn=lambda m, _: isinstance(m, torch.nn.Embedding) or _is_linear(m),
)
```

SalmanMohammadi:
Closes #2155.

### DIFF
--- a/torchao/quantization/qat/README.md
+++ b/torchao/quantization/qat/README.md
@@ -115,10 +115,11 @@ To fake quantize embedding in addition to linear, you can additionally call
 the following with a filter function during the prepare step:
 
 ```
+from torchao.quantization.quant_api import _is_linear
 quantize_(
     m,
     IntXQuantizationAwareTrainingConfig(weight_config=weight_config),
-    filter_fn=lambda m, _: isinstance(m, torch.nn.Embedding),
+    filter_fn=lambda m, _: isinstance(m, torch.nn.Embedding) or _is_linear(m),
 )
 ```
 


jerryzh168 (approved):

jerryzh168:
cc @andrewor14

torchao/quantization/qat/README.md
```diff
@@ -115,10 +115,11 @@ To fake quantize embedding in addition to linear, you can additionally call
 the following with a filter function during the prepare step:
 
 ```
+from torchao.quantization.quant_api import _is_linear
 quantize_(
     m,
     IntXQuantizationAwareTrainingConfig(weight_config=weight_config),
-    filter_fn=lambda m, _: isinstance(m, torch.nn.Embedding),
+    filter_fn=lambda m, _: isinstance(m, torch.nn.Embedding) or _is_linear(m),
```
andrewor14, line 122: Actually this is only if you want to use the same configuration for embedding and linear. I kept them as two separate calls because in the above example linear additionally has activation quantization.

SalmanMohammadi: Yeah you're right and this example won't work if you try and apply a config with activation quantization to both linear and embedding layers at the same time. 
You can stack calls to `quantize` right? Would the right way to go about this two `quantize` calls, one which iflters for linear, then another which filters for embeddings?

andrewor14: Yeah if you have slightly different quantization configurations for embedding and linear the right way would be two separate `quantize_` calls. This is by design because we don't want to complicate `quantize_` to accept a dictionary of configs

PR #2174, opened 2025-05-05
Fixing aliasing behavior for slice in AQT TensorCoreTiledLayout

jerryzh168:
Summary:
slice op is supposed to preserve aliasing (output of slice should alias the input), but this is not true for TensorCoreTiledLayout (used by int4wo), and some others like gemlite

Reason is that we do unpacking, pading and prepacking right now, which creates new tensors.

We fixes it in this PR by doing slicing on the packed inner Tensor directly, specifically packed_weight and scale_and_zero in TensorCoreTiledLayout.

Test Plan:
python test/dtypes/test_affine_quantized.py -k test_slice_and_copy_int4wo

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/dtypes/test_affine_quantized.py
+++ b/test/dtypes/test_affine_quantized.py
@@ -387,6 +387,31 @@ class TestAffineQuantizedBasic(TestCase):
         # make sure it runs
         torch.matmul(x, w.t())
 
+    @common_utils.parametrize("device", ["cuda"])
+    @common_utils.parametrize("dtype", [torch.bfloat16])
+    @skip_if_no_cuda()
+    @skip_if_rocm("ROCm enablement in progress")
+    def test_slice_and_copy_int4wo(self, device, dtype):
+        l = torch.nn.Linear(1024, 1024).to("cuda").to(torch.bfloat16)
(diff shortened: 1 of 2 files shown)


test/dtypes/test_affine_quantized.py
```diff
@@ -387,6 +387,31 @@ def test_matmul(self, device, dtype):
...
+        # dummy_l has random input (shouldn't be 0)
+        dummy_l = torch.nn.Linear(1024, 1024).to("cuda").to(torch.bfloat16)
+        quantize_(dummy_l, Int4WeightOnlyConfig())
+        quantized = dummy_l.weight
+        quantized = quantized.narrow(0, 0, 512)
+
+        param_data.copy_(quantized)
+
+        # making sure param.data is updated
+        assert param.data.dequantize()[0][0] != 0
```
drisspg, line 412: Can we also just assert that the storage_ptrs are the same

jerryzh168: sg, added

drisspg (approved):

Issue #2094, opened 2025-04-22
[Quant][PT2E] AffineQuantized observers failed Resnet18


Xia-Weiwen:
**Error message**
```
  File "/home/weiwen/torchao/torchao/quantization/pt2e/pt2e/_affine_quantization.py", line 739, in forward
    self.min_val.shape == min_val.shape
AssertionError: Can't update existing min_val - shape mismatch, self.min_val:torch.Size([]) != min_val:torch.Size([1, 1])
```

**How to reproduce**
```python
import torch
import torchao
import torchvision
import copy

from torchao.quantization.pt2e.quantize_pt2e import (
    convert_pt2e,
    prepare_pt2e,
)
import torchao.quantization.pt2e.quantizer.x86_inductor_quantizer as xiq
from torchao.quantization.pt2e.quantizer.x86_inductor_quantizer import X86InductorQuantizer
from torchao.quantization.pt2e import move_exported_model_to_eval
from torch.ao.quantization.pt2e.lowering import lower_pt2e_quantized_to_x86


def get_quant_config():
    from torchao.quantization.pt2e.pt2e._affine_quantization import (
        AffineQuantizedMinMaxObserver,
        AffineQuantizedMovingAverageMinMaxObserver,
    )
    from torchao.quantization.pt2e.observer import MappingType, PerAxis, PerTensor
    from torchao.quantization.pt2e.quantizer.quantizer import QuantizationSpec
    from torchao.quantization.pt2e.quantizer.xnnpack_quantizer_utils import QuantizationConfig
    act_quantization_spec = QuantizationSpec(
        dtype=torch.uint8,
        quant_min=0,
        quant_max=255,
        qscheme=torch.per_tensor_affine,
        is_dynamic=False,
        observer_or_fake_quant_ctr=AffineQuantizedMovingAverageMinMaxObserver.with_args(
            target_dtype=torch.uint8,
            mapping_type=MappingType.ASYMMETRIC,
            granularity=PerTensor(),
            averaging_constant=1,
        ),
    )
    weight_quantization_spec = QuantizationSpec(
        dtype=torch.int8,
        quant_min=-128,
        quant_max=127,
        qscheme=torch.per_channel_symmetric,
        ch_axis=0,  # 0 corresponding to oc
        is_dynamic=False,
        observer_or_fake_quant_ctr=AffineQuantizedMinMaxObserver.with_args(
            target_dtype=torch.int8,
            mapping_type=MappingType.SYMMETRIC,
[...]

Issue #2112, opened 2025-04-23
[PT2E] observers does not handle inputs with different shapes correctly


Xia-Weiwen:
**Summary**
The observers in Torchao, such as `AffineQuantizedMinMaxObserver`, support different quantization granularity by keeping block sizes. For example, if granularity is `PerTensor` and input shape is `(16, 3, 224, 224)`, the observer keeps block sizes = `(16, 3, 224, 224)`.
However, the block sizes would be wrong if inputs with different shapes are passed in. For example, if another input with shape = `(16, 3, 56, 56)` comes, the block sizes are updated as `(16, 3, 56, 56)`. It is wrong for inputs with shape = `(16, 3, 224, 224)`.

**How to reproduce**
The code to reproduce is similar as the reproducer here: https://github.com/pytorch/ao/issues/2094
One needs to bypass the https://github.com/pytorch/ao/issues/2094 issue by changing source code manually and also run the `converted_model` after `convert_pt2e`:
```python
    converted_model = convert_pt2e(prepared_model)
    move_exported_model_to_eval(converted_model)
    print("[info] converted_model =\n", converted_model)
    converted_model(*example_inputs)  # add this line
```

jerryzh168:
yeah we haven't test per tensor that much I think, we'd need to fix per tensor use case

jerryzh168:
I have thought about this scenario a bit before, I think the fix for this one would be using `granularity` instead of `block_size` in the observer, please feel free to take this if you have time @Xia-Weiwen

Xia-Weiwen:
Do you mean using granularity in all cases or only for per-tensor?

Xia-Weiwen:
I think using block size = -1 for per-tensor also makes sense. What do you think?

jerryzh168:
> Do you mean using granularity in all cases or only for per-tensor?

just for per tensor for now, we can assert that only one of `granularity` and `block_size` can be valid, and use the one that's available

jerryzh168:
> I think using block size = -1 for per-tensor also makes sense. What do you think?

yeah this could make sense, but then the question is can we express all dynamic shapes with block_size?

jerryzh168:
> I think so. For a certain dimension:
> 
> * If group size > 1 is specified, we just use that group size as block size
> * If group size == 1, such as per-channel and per token, we use block size = 1
> * If it's not group-wise quantization, such as per-tensor and the other dimension of per-channel, we use block size = -1
> 
[...]

PR #2177, opened 2025-05-06
[PT2E] Fix per-tensor observer issue with varing shape & rank

Xia-Weiwen:
Fixes https://github.com/pytorch/ao/issues/2094 and https://github.com/pytorch/ao/issues/2112
We may find inputs with varying shapes and ranks, e.g. when running Resnet18. The current implementation is based on block_size, which is not enough for such cases. The fix is simple: use block_size = -1 for each dimension for per-tensor quantization and update block_size for each input when inserting q/dq in `convert`.


torchao/quantization/pt2e/observer.py
```diff
@@ -1891,6 +1891,10 @@ def convert(self, model: torch.fx.GraphModule, observer_node: Node):
             assert self.original_dtype is not None, (
                 "Expecting original_dtype to be populated"
             )
+            # Since input shape & rank may change (e.g. Resnet18), here we need to update block_size for each input
+            self.block_size = get_block_size(
```
jerryzh168, line 1895: when does this happen? can you give an example?

Xia-Weiwen: To reproduce the issue, you may run the code here: https://github.com/pytorch/ao/issues/2094#issue-3010187171
You will have to using `-1` for `block_size` without updating of `self.block_size` here.

jerryzh168: are you saying the rank / number of dimension changes for input as well? can we use a single `-1` to represent this case?

Xia-Weiwen: > are you saying the rank / number of dimension changes for input as well?

Yes

> can we use a single -1 to represent this case?

I think it's doable. But there are checks to guard `len(self.block_size) == len(input.shape)`. We need to handle the special case for per-tensor quant at these locations. Is it ok?

Xia-Weiwen:
@jerryzh168 Could you please review this PR? Thanks.

Xia-Weiwen:
Hi @jerryzh168  @drisspg Could you please review this PR? I am not sure if the current implementation is what you expected. Thanks.

torchao/quantization/pt2e/_affine_quantization.py
```diff
@@ -113,7 +113,8 @@ def _get_reduction_params(block_size, input_size):
           shape_for_reduction: (3, 3, 5, 2, 10)
           reduction_dim: [0, 1, 3, 4]
     """
-    assert len(block_size) == len(input_size)
```
jerryzh168, line 116: is this still used? we should be using the code in quant_primitives.py I think

torchao/quantization/pt2e/_affine_quantization.py
```diff
@@ -113,7 +113,8 @@ def _get_reduction_params(block_size, input_size):
...
+    assert block_size == [-1] or len(block_size) == len(input_size)
+    block_size = [-1] * len(input_size) if block_size == [-1] else block_size
```
jerryzh168, line 117: is this correct? I think here we need to fill in real dimensions

torchao/quantization/pt2e/_affine_quantization.py
Xia-Weiwen: Yes. It's still used when running the prepared model (model after `prepare_pt2e`). Is it a bug? Do I need to fix it, too?

torchao/quantization/pt2e/_affine_quantization.py
Xia-Weiwen: Seems both are working. I will change it to real dimensions. Thanks.

torchao/quantization/pt2e/_affine_quantization.py
Xia-Weiwen: I am using the observers defined here: torchao/quantization/pt2e/_affine_quantization.py

jerryzh168: I think we should be using the ones in torchao/quantization/observer.py eventually

only occurrence seems to be https://github.com/pytorch/ao/blob/554cb60c750e6ef31bbcafec74bb76a4578902da/test/quantization/pt2e/test_quantize_pt2e.py#L2531 and we want to update it I think

so if you are adding new things I'd recommend use the one from torchao.quantization

Xia-Weiwen: Hi @jerryzh168 For the issue I mentioned, do you plan to work on it or do you want us to fix it? Thanks.

Xia-Weiwen: Hi @jerryzh168 Do you have any suggestions on this? Thanks.

jerryzh168: I thought we already use the one from torchao: https://github.com/pytorch/ao/blob/212d912dcbfb43b13cc4bee533d3d1844d76b7c1/torchao/quantization/pt2e/prepare.py#L35 but if you saw we are using torch.ao please go ahead and change them

Xia-Weiwen: @jerryzh168 Thanks for the reply. I did not mean `torch.ao`. I meant there are two versions of such utilities in torchao, `torchao.quantization.pt2e` and `torchao.quantization`. For example,
https://github.com/pytorch/ao/blob/5153bd3ce9fc4e873a00d7a24000114ce93a2899/torchao/quantization/pt2e/observer.py#L84
and
https://github.com/pytorch/ao/blob/5153bd3ce9fc4e873a00d7a24000114ce93a2899/torchao/quantization/observer.py#L32
The PT2E flow in torchao uses those in `torchao.quantization.pt2e` while you said you wanted to switch to `torchao/quantization/observer.py`.
So, I was asking whether you would switch to `torchao/quantization/observer.py` in PT2E flow first. Do you have any suggestions on that? Thanks.

jerryzh168: oh I see, yeah for now use `torchao/quantization/observer.py` would be better I think, we haven't finalized the folder structure for this one yet

Xia-Weiwen: @jerryzh168 Am I supposed to wait until you finalize the folder structure? Thanks.

Xia-Weiwen:
This PR is stale and looks like it's not an urgent task to switch to Torchao's observer for PT2E. So, let me close it and open a new one if needed in the future.

PR #2179, opened 2025-05-06
Eval hf models using lm_eval

jainapurva:


torchao/_models/llama/eval_hf.py
```diff
@@ -0,0 +1,182 @@
...
+#         return_tensors="pt",
+#     ).to("cuda")
+#     generated_ids = quantized_model.generate(**inputs, max_new_tokens=128)
+#     output_text = tokenizer.batch_decode(
+#         generated_ids, skip_special_tokens=True, clean_up_tokenization_spaces=False
+#     )
+#     print("Response:", output_text[0][len(prompt):])
+
+
+def model_throughput(
```
jerryzh168, line 88: nit: maybe you can add a memory measurement as well

jainapurva: Done

torchao/_models/README.md
```diff
@@ -1,4 +1,43 @@
-## SAM2
+# LLAMA
+
+## Eval on Llama 3.1 8B and Llama 3.2 3B
+
+We use lm-eval tasks for evaluating TorchAO Quantization APIs on HuggingFace models. The results are in the table below:
```
HDCharles, line 5: would specify what task this is for.

benchmarks/microbenchmarks/utils.py
```diff
@@ -291,6 +292,15 @@ def string_to_config(
         else:
             granularity = PerTensor()
         return Float8DynamicActivationFloat8WeightConfig(granularity=granularity)
+    if "gemlitewo" in quantization:
```
HDCharles, line 295: gemlite can be 4 bit or 8 bit, should probably specify that this is for 4 bit

https://github.com/pytorch/ao/blob/main/torchao/quantization/quant_api.py#L968

HDCharles (approved):
looks good aside from those small comments

PR #2180, opened 2025-05-06
Set eps in end-to-end QAT flow

andrewor14:
**Summary:** This commit does two things:

(1) Allow users to set eps in `FakeQuantizeConfig`
(2) For other parts of the QAT flow, set eps to `torch.finfo(torch.float32).eps` for input linear activations to match the existing hardcoded input activation scale dtype (which is fp32)

The motivation is to enable users who wish to lower their models to XNNPACK. This would require them to use the following combination of dtypes during training for end-to-end numerical match:

- input activations: bf16
- input activation scales: fp32
- input activation eps: `torch.finfo(torch.float32).eps`
- weight: bf16
- weight scales: bf16
- weight eps: `torch.finfo(torch.bfloat16).eps`

However, today there is no way to specify the above in any of the QAT flows. For the recommended `FakeQuantizeConfig` flow, we always use `torch.finfo(x.dtype).eps`, where x is bf16 in this case, and there is no way for users to configure this. This is resolved by (1).

For the legacy `Int8DynActInt4QATQuantizer` flow, we hardcode input activation scales to always use fp32 in
https://github.com/pytorch/ao/pull/2085, but did not set the corresponding eps. Today, this also uses `torch.finfo(x.dtype).eps` by default, where x is bf16, and so we use the wrong eps value. This is resolved by (2).

**Test Plan:**
python test/quantization/test_qat.py -k test_fake_quantize_config_eps
python test/quantization/test_qat.py -k test_qat_8da4w_eps

### DIFF
--- a/torchao/quantization/GPTQ.py
+++ b/torchao/quantization/GPTQ.py
@@ -938,7 +938,10 @@ def linear_forward_8da4w(
     # TODO: in future add ability to specify activation_scale_dtype to PTQ configs
     # and enable similar change here
     x = per_token_dynamic_quant(
-        x, scale_dtype=torch.float32, zero_point_dtype=torch.float32
+        x,
+        scale_dtype=torch.float32,
+        zero_point_dtype=torch.float32,
+        eps=torch.finfo(torch.float32).eps,
     )
 
     # TODO: verify and remove following reshape code
--- a/torchao/quantization/qat/fake_quantizer.py
+++ b/torchao/quantization/qat/fake_quantizer.py
@@ -81,6 +81,7 @@ class FakeQuantizer(torch.nn.Module):
                 target_dtype=self.config.dtype,
                 quant_min=qmin,
                 quant_max=qmax,
+                eps=self.config.eps,
                 scale_dtype=self.config.scale_precision,
                 zero_point_dtype=self.config.zero_point_precision,
             )
@@ -117,6 +118,7 @@ class FakeQuantizer(torch.nn.Module):
                     bit_width,
                     group_size,
                     scale_precision,
+                    eps=self.config.eps,
                 )
             else:
                 (self.scale, self.zero_point) = get_groupwise_affine_qparams(
@@ -124,6 +126,7 @@ class FakeQuantizer(torch.nn.Module):
                     bit_width,
                     group_size,
                     scale_precision,
+                    eps=self.config.eps,
                 )
             self.zero_point = self.zero_point.to(zero_point_precision)
 
--- a/torchao/quantization/utils.py
+++ b/torchao/quantization/utils.py
@@ -324,6 +324,7 @@ def get_groupwise_affine_qparams(
     dtype=torch.bfloat16,
     zero_point_domain=ZeroPointDomain.FLOAT,
     preserve_zero=False,
+    eps=None,
 ):
     if groupsize > w.shape[-1]:
         groupsize = w.shape[-1]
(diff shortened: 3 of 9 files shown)


torchao/quantization/utils.py
```diff
@@ -539,7 +540,8 @@ def get_group_qparams_symmetric(
     assert n_bit <= 8, f"unsupported n_bit: {n_bit}"
 
     block_size = (1, groupsize)
-    eps = torch.finfo(w.dtype).eps
+    if eps is None:
+        eps = torch.finfo(w.dtype).eps
```
metascroy, line 544: Isn't this already the behavarior inside choose_qparams_affine?

torchao/quantization/qat/fake_quantizer.py
```diff
@@ -117,8 +118,10 @@ def _per_channel_or_group_forward(self, x: torch.Tensor):
                     bit_width,
                     group_size,
                     scale_precision,
+                    eps=self.config.eps,
                 )
             else:
+                # TODO: validate eps against 1e-6 here
```
metascroy, line 124: Why not pipe eps through to get_groupwise_affine_qparams?

torchao/quantization/utils.py
andrewor14: Yeah, I could just change FakeQuantizer to use that instead, but seems like a bigger change, prefer to leave that in a future PR to keep this one isolated

torchao/quantization/qat/fake_quantizer.py
andrewor14: Sure, I didn't do it because `get_groupwise_affine_qparams` hardcodes it to 1e-6 today, but I can add the validation in this PR instead

metascroy: But you can make the default value 1e-6 if not provided to match behavior today?

torchao/quantization/GPTQ.py
```diff
@@ -938,7 +938,10 @@ def linear_forward_8da4w(
     # TODO: in future add ability to specify activation_scale_dtype to PTQ configs
     # and enable similar change here
     x = per_token_dynamic_quant(
-        x, scale_dtype=torch.float32, zero_point_dtype=torch.float32
+        x,
+        scale_dtype=torch.float32,
+        zero_point_dtype=torch.float32,
+        eps=torch.finfo(torch.float32).eps,
```
metascroy, line 944: We also need to update here: https://github.com/pytorch/ao/blob/137b0795acb3282ce622948b1537e20914186eea/torchao/quantization/quant_api.py#L625

metascroy (approved):

metascroy:
@andrewor14 maybe we can relax the failing test_export_compile_aoti_PackedLinearInt8DynamicActivationIntxWeightLayout test to just look for the op name?

andrewor14:
Hey @metascroy, looks like there's a new test failure and it seems real? I think it's complaining that qdq ops are not fused. I wonder if it's because there's a new eps value that broke the pattern matching somehow. Can you take a look?

```
python torchao/experimental/tests/test_quant_passes.py -k test_replace_q_dq_patterns_with_quantized_linear_ops_pass
```

metascroy:
> Hey @metascroy, looks like there's a new test failure and it seems real? I think it's complaining that qdq ops are not fused. I wonder if it's because there's a new eps value that broke the pattern matching somehow. Can you take a look?
> 
> ```
> python torchao/experimental/tests/test_quant_passes.py -k test_replace_q_dq_patterns_with_quantized_linear_ops_pass
> ```

Traveling today, but I'll have a look later in afternoon.

metascroy:
> > Hey @metascroy, looks like there's a new test failure and it seems real? I think it's complaining that qdq ops are not fused. I wonder if it's because there's a new eps value that broke the pattern matching somehow. Can you take a look?
> > ```
> > python torchao/experimental/tests/test_quant_passes.py -k test_replace_q_dq_patterns_with_quantized_linear_ops_pass
> > ```
> 
> Traveling today, but I'll have a look later in afternoon.

Pushed a fix to your PR

andrewor14:
Ok, remaining test failure seems unrelated and happens on other PR. I think this is good to land.

PR #2190, opened 2025-05-09
[float] document e2e training -> inference flow

danielvegamyhre:
## Summary

Document the E2E training => inference flow with examples.

### DIFF
--- a/torchao/float8/README.md
+++ b/torchao/float8/README.md
@@ -238,12 +238,10 @@ The first step in the E2E is to train your model and save a checkpoint. The seco
 ```python
 import torch
 from torch import nn
+import torch.nn.functional as F
 
 from torchao.float8.float8_linear_utils import convert_to_float8_training
 from torchao.float8.float8_linear import Float8Linear
-
-import torch
-import torch.nn as nn
 from torchao.float8 import convert_to_float8_training
 from torchao.utils import TORCH_VERSION_AT_LEAST_2_5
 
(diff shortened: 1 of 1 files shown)


danielvegamyhre:
cc @andrewor14

andrewor14 (approved):
Tested this out locally too, works for me. Thanks!

andrewor14:
(Might need to add a distributed checkpoint section but we can do that in a separate PR)

torchao/float8/README.md
```diff
@@ -230,3 +230,95 @@ including [downloading a tokenizer](https://github.com/pytorch/torchtitan?tab=re
...
+    output = m(x)
+    # use fake labels for demonstration purposes
+    fake_labels = torch.ones_like(output)
+    loss = F.mse_loss(output, fake_labels)
+    loss.backward()
+    optimizer.step()
+
+# save the model
+torch.save({
+    'model': m,
```
danielvegamyhre, line 289: In practice the model would be in some modeling file, and the training code and inference code would both import it separately, in order to avoid the need to deserialize the python model definition w/ `torch.load(...., weights_only=False)`, which has some security risks. 

However, I was aiming to have these be copy/paste-able runnable standalone examples, which seemed to require this bad practice. Thoughts @andrewor14 @vkuzo?

vkuzo: I think a good way to do it as as follows:
1. create a reproducible model definition
2. create a new instance of (1), train it, save weights to checkpoint
3. create a new instance of (1), load weights from checkpoint, finetune it or do inference

there is no saving of model definition with `torch.save` needed in the flow as above

danielvegamyhre: That's what I originally tried actually, but it doesn't work because the weights in the serialized/checkpointed model from step (2) are registered under different names (prefixed with `_orig_mod`) than the freshly initialized model in step (3). 

I solved this by saving the converted model definition directly in torch.save and loading the model state dict into that, but it's not ideal imo. I'm curious how torchtitan/torchtune do this as well

danielvegamyhre: Ok going to merge this for now, we can discuss alternatives async if you want.

PR #2170, opened 2025-05-05
Fix linux cpu builds. Resolves nightly build for mac stops on 0422

atalman:
Fixes https://github.com/pytorch/ao/issues/2157

This reverts 2 PRS: 
https://github.com/pytorch/ao/pull/2044
https://github.com/pytorch/ao/pull/2093
Which introduced the regression.

This works: ``torchao-0.11.0.dev20250505+cpu-py3-none-any.whl`` generates correct file.
Please see details on https://github.com/pytorch/ao/issues/2157

PR #2163, opened 2025-05-02
Uses torch.version.cuda to compile CUDA extensions

syed-ahmed:
Fixes: https://github.com/pytorch/ao/issues/2152

Tested this by installing a cpu only version of PyTorch and then running `pip install -e . -vvv`. "PyTorch GPU support is not available. Skipping compilation of CUDA extensions" is in the logs, confirming that the previous behavior is preserved.
```
Using pip 25.1 from /usr/local/lib/python3.12/dist-packages/pip (python 3.12)                 
Non-user install because site-packages writeable                                              
Created temporary directory: /tmp/pip-build-tracker-a0n8qmb2                                  
Initialized build tracking at /tmp/pip-build-tracker-a0n8qmb2                                 
Created build tracker: /tmp/pip-build-tracker-a0n8qmb2                                        
Entered build tracker: /tmp/pip-build-tracker-a0n8qmb2                                        
Created temporary directory: /tmp/pip-install-tn_td1mv                                        
Created temporary directory: /tmp/pip-ephem-wheel-cache-iflb4k6r                              
Obtaining file:///workspace/ao                                                                
  Added file:///workspace/ao to build tracker '/tmp/pip-build-tracker-a0n8qmb2'               
  Running setup.py (path:/workspace/ao/setup.py) egg_info for package from file:///workspace/a
o                                                                                             
  Created temporary directory: /tmp/pip-pip-egg-info-j4p5drqz                                 
  Running command python setup.py egg_info                                                    
  INFO:root:running egg_info                                                                  
  INFO:root:creating /tmp/pip-pip-egg-info-j4p5drqz/torchao.egg-info                          
  INFO:root:writing /tmp/pip-pip-egg-info-j4p5drqz/torchao.egg-info/PKG-INFO                  
  INFO:root:writing dependency_links to /tmp/pip-pip-egg-info-j4p5drqz/torchao.egg-info/depend
ency_links.txt                                                                                
  INFO:root:writing requirements to /tmp/pip-pip-egg-info-j4p5drqz/torchao.egg-info/requires.t
xt                                                                                            
  INFO:root:writing top-level names to /tmp/pip-pip-egg-info-j4p5drqz/torchao.egg-info/top_lev
el.txt                                                                                        
  INFO:root:writing manifest file '/tmp/pip-pip-egg-info-j4p5drqz/torchao.egg-info/SOURCES.txt
'                                                                                             
[...]

PR #2193, opened 2025-05-09
Uses torch.version.cuda to compile CUDA extensions

syed-ahmed:
@atalman the changes in https://github.com/pytorch/ao/pull/2170 removed my changes from https://github.com/pytorch/ao/pull/2163. Note that torch.cuda.is_available() fails in a system where CUDA toolkit is available for compilation but there are no GPUs. As a result, extension won't get compiled in that system.


atalman (approved):
lgtm

PR #2197, opened 2025-05-12
Forward fix lint

msaroufim:
We don't have clear instructions on linting

https://github.com/pytorch/ao/actions/runs/14962782164/job/42027402557


gau-nernst (approved):
In my last PR, for some reasons, CI didn't run when I pushed new changes after you approved, hence I didn't catch the lint issue. Sorry about that.

I think a better way is to use pre-commit? And the lint CI can run pre-commit as well -> single source of lint

PR #2210, opened 2025-05-14
primitive scale fix

YIWENX14:
Differential Revision: D74446877

### DIFF
--- a/torchao/quantization/quant_primitives.py
+++ b/torchao/quantization/quant_primitives.py
@@ -1270,6 +1270,8 @@ def choose_qparams_affine_with_min_max(
     if eps is None:
         eps = torch.finfo(min_val.dtype).eps
 
+    scale_device = min_val.device
+
     if preserve_zero:
         min_val_neg = torch.min(min_val, torch.zeros_like(min_val))
         max_val_pos = torch.max(max_val, torch.zeros_like(max_val))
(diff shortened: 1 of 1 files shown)


torchao/quantization/quant_primitives.py
```diff
@@ -948,7 +952,9 @@ def _choose_qparams_affine(
         scale = torch.clamp(scale, min=eps)
     else:
         assert mapping_type == MappingType.ASYMMETRIC.name
-        scale = (max_val_pos - min_val_neg) / float(quant_max - quant_min)
+        scale = (max_val_pos - min_val_neg) / torch.tensor(
```
drisspg, line 955: what if you did:
`(max_val_pos - min_val_neg) / (quant_max - quant_min).to(torch.float32)`

YIWENX14: Casting to `float32` doesn't help the discrepancy on CPU vs GPU.

drisspg: Ahh sorry I see that quant_max and quant_mins are ints, why is this better than the existing? 
I am confused as to what this is trying to do?  

```Py
import torch
from transformer_nuggets.utils.tracing import LoggingMode
a = torch.randn(2, 3, device="cuda")
with LoggingMode():
    out = a / float(32 - 12)
    print(out)


with LoggingMode():
    out = a / torch.tensor(float(32 - 12), dtype=a.dtype, device=a.device)
    print(out)
    
 ```
 Produces:
 
 ```Py
 $1: f32[2, 3] = aten.div.Tensor($0, 20.0)
tensor([[-0.0313, -0.0028, -0.0344],
        [ 0.0299,  0.0099, -0.0426]], device='cuda:0')
$0: f32[] = aten.lift_fresh.default($0)
$2: f32[2, 3] = aten.div.Tensor($1, $0)
tensor([[-0.0313, -0.0028, -0.0344],
        [ 0.0299,  0.0099, -0.0426]], device='cuda:0')
```

jerryzh168 (approved):
OK, I think it's fine to merge as long as it doesn't break existing tests, there are some flexibility in these quant primitive ops since we didn't really define these very precisely

torchao/quantization/quant_primitives.py
kimishpatel: Driss, the issue is "float" as in python float is actually float64

drisspg: So this isn't about fixing a device but about fixing the dtype?

PR #2213, opened 2025-05-15
Add selective weight loading decode kernel for activation sparsity

jcaip:
Summary:

This PR adds in a kernel to accelerate $$xW^T$$ when x is sparse and we are memory bound.

The idea here is that we can avoid loading the columns of $W$ that correspond to the zero elements of $x$.

This lets us accelerate activation sparsity for bs=1 decode use cases.

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/kernel/sparse_gemv/sparse_gemv.py
+++ b/torchao/kernel/sparse_gemv/sparse_gemv.py
@@ -1,0 +1,139 @@
+# adapted from deja vu
+
+from typing import Optional
+
+import torch
+import triton
+import triton.language as tl
+def init_to_zero(*names):
+    def init_func(nargs):
+        for name in names:
+            nargs[name].zero_()
+    return init_func
+
+# NOTE: will need to warm up kernels each time, triton autotune caching isn't a thing right now
+
+configs=[
+    triton.Config({"BLOCK_M": 64, "BLOCK_N": 128}, num_warps=2, pre_hook=init_to_zero("Y")), 
+
+    triton.Config({"BLOCK_M": 64, "BLOCK_N": 64}, num_warps=4, pre_hook=init_to_zero("Y")),
+    triton.Config({"BLOCK_M": 8, "BLOCK_N": 128}, num_warps=2, pre_hook=init_to_zero("Y")),
+    triton.Config({"BLOCK_M": 16, "BLOCK_N": 256}, num_warps=4, pre_hook=init_to_zero("Y")),
+    triton.Config({"BLOCK_M": 16, "BLOCK_N": 256}, num_warps=4, pre_hook=init_to_zero("Y")),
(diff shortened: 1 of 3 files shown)


torchao/dtypes/floatx/cutlass_semi_sparse_layout.py
```diff
@@ -15,15 +15,51 @@
...
+from torchao.dtypes.utils import AQTTensorImpl, Layout, get_out_shape
 from torchao.ops import (
     rowwise_scaled_linear_sparse_cutlass_f8f8,
-    to_sparse_semi_structured_cutlass_sm9x_f8,
 )
 
 aten = torch.ops.aten
 
 
+def _pad_dense_input(dense_input: torch.Tensor) -> torch.Tensor:
```
jerryzh168, line 26: I think ideally we don't do padding at this level, since this interferes with aliasing semantics, and make some op implementation impossible, like slice. is it possible to move this to the kernel itself?

torchao/dtypes/floatx/cutlass_semi_sparse_layout.py
```diff
@@ -66,11 +97,12 @@ def __new__(
         )
         kwargs["dtype"] = sparse.dtype
         kwargs["requires_grad"] = False
-        shape = (sparse.shape[0], 2 * sparse.shape[-1])
+        # shape = (sparse.shape[0], 2 * sparse.shape[-1])
```
jerryzh168, line 100: remove?

torchao/dtypes/floatx/cutlass_semi_sparse_layout.py
```diff
@@ -80,6 +112,7 @@ def __init__(
         self.meta = meta
         self.scale = scale
         self._layout = _layout
+        self._shape = shape
```
jerryzh168, line 115: why do we need a separate shape, instead of deriving from the `sparse` tensor?

jerryzh168: nit: also sparse naming might be a bit vague, renaming to `sparse_data` or something might be better

torchao/dtypes/floatx/cutlass_semi_sparse_layout.py
```diff
@@ -210,3 +257,46 @@ def _linear_fp8_act_fp8_weight_sparse_cutlass_impl(input_tensor, weight_tensor,
     )
 
     return out
+
+
+def _linear_fp8_act_sparse_fp8_weight_cutlass_check(input_tensor, weight_tensor, bias):
+    from torchao.dtypes.floatx import Float8Layout
+
+    res = (
+        isinstance(input_tensor, AffineQuantizedTensor)
```
jerryzh168, line 266: just FYI, I feel we could simplify this list of checks if don't need to support a variety of kernels that all use the same subclassed tensor, it only needs to be specific enough so we can dispatch to the correct kernel. maybe this can be simplified if we split the AQT in the future.

PR #2207, opened 2025-05-13
[reland2][ROCm] preshuffled weight mm

jeffdaily:

PR #2219, opened 2025-05-19
Manually specify flags if no arch set

drisspg:
Stacked PRs:
 * __->__#2219

--- --- ---

Manually specify flags if no arch set


drisspg:
I am trying to unblock: https://github.com/pytorch/ao/pull/2207#issuecomment-2891533439
I don't know what the spanning versions we should add by default though are

mxz297:
@drisspg  this pr seems to also fail but with some different failures... Are trunk tests all healthy?

PR #2221, opened 2025-05-19
Make torchao pt2e prepare/convert functions compatible with quantizers in torch.ao

metascroy:
Currently we cannot use torchao's prepare/convert functions with quantizers based on torch.ao.quantization.quantizer.quantizer.Quantizer.  This makes migration difficult to do in pieces.

In this PR, we add logic to prepare/convert functions to recognize torch.ao.quantization.quantizer.quantizer.Quantizer and call the previous prepare/convert APIs.

### DIFF
--- a/torchao/quantization/pt2e/quantize_pt2e.py
+++ b/torchao/quantization/pt2e/quantize_pt2e.py
@@ -95,6 +95,15 @@ def prepare_pt2e(
         # run calibration
         # calibrate(m, sample_inference_data)
     """
+    # We will temporarily make prepare_pt2e backward compatible with quantizers that configs, observers,
+    # and fake quantizers from torch.ao instead of torchao
+    if isinstance(quantizer, torch.ao.quantization.quantizer.quantizer.Quantizer):
+        from torch.ao.quantization.quantize_pt2e import (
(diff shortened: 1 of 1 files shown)


torchao/quantization/pt2e/quantize_pt2e.py
```diff
@@ -217,6 +235,32 @@ def _quant_node_constraint(n: Node) -> bool:
...
+        if isinstance(
+            m, torch.ao.quantization.fake_quantize.FakeQuantize
+        ) or isinstance(m, torch.ao.quantization.observer.ObserverBase):
+            is_torch_ao_prepared = True
+        if isinstance(m, torchao_FakeQuantize) or isinstance(m, torchao_ObserverBase):
+            is_torchao_prepared = True
+    assert is_torch_ao_prepared or is_torchao_prepared, (
+        "Must be prepared using torch.ao or torchao, but did not see any FakeQuantize or ObserverBase modules in model: "
+        + str([f"{n}: {type(m)}" for n, m in model.named_modules()])
+    )
```
jerryzh168, line 256: there might be models that does not have observer/fake_quant inserted, I think we could remove this assert

jerryzh168 (approved):

PR #2225, opened 2025-05-19
use correct fp8 quantization dtype for AMD GPU

yuguo68:
Reviewed By: dakechen1993

Differential Revision: D75021458

### DIFF
--- a/test/float8/test_base.py
+++ b/test/float8/test_base.py
@@ -56,6 +56,7 @@ from torchao.float8.float8_utils import (
     tensor_to_scale,
 )
 from torchao.testing.float8.test_utils import get_test_float8_linear_config
+from torchao.utils import is_MI300, is_ROCM
 
 random.seed(0)
 torch.manual_seed(0)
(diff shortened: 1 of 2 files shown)


yuguo68:
@jerryzh168 , @drisspg could you help review? thanks! seems I cannot add reviewers on the right panel.

drisspg:
So previously the expectation was just that users would manually pass int he FP8 dtypes but I think this is fine for now, can you add a test a small test for AMD machines?

yuguo68:
> So previously the expectation was just that users would manually pass int he FP8 dtypes but I think this is fine for now, can you add a test a small test for AMD machines?

@drisspg thanks for the review! from a user perspective I would much prefer the choice between torch.float8_e4m3fnuz/torch.float8_e4m3fn is handled internally. added a small test for AMD MI300X.

yuguo68:
@drisspg could you take a second look? thanks!

drisspg (approved):

PR #2235, opened 2025-05-21
GPTQ updates

HDCharles:
Summary:

1) reorganized GPTQ
 a) got rid of old GPTQ and renamed GPTQ_MT to GPTQ
 b) moved new GPTQ to prototype
 c) moved quantized linear modules in GPTQ.py to linear_quant_modules.py
2) removed dependence on lm_eval for input_recorder
 a) created new input recorder that doesn't depend on lm_eval
 b) made lm_eval input recorder depend on new generic input_recorder
 c) made TransformerEvalWrapper the base class and made LMEvalInputRecorder inherit from it (rather than vice versa like before)
 d) updated apis generally to work with new input recorder (inputs have to be passed in as though they were being passed into the model, so input_recorder(*args) rather than input_recorder(args)
3) reorganized GPTQ tests
 a) moved tests from test_quant_api.py to test_gptq.py
 b) added new test that can be run in CI that doesn't depend on
 c) removed all the 8aw4 tests that we never got working (is this fine?)
 lm_eval/llama weights
 c) got rid of test_gptq_mt.py (consolidated into test_gptq where relevant.
4) added new documentation for lm_eval
 a) new readme and eval benchmarks for GPTQ
 b) comments in GPTQ.py
5) GPTQ improvements
 a) tested compilation of hessian calculation and parts of faster quant,
 generally they were slower or buggy. Possible to get a speedup but it was inconsistent so removed it.
 b) reimplemented faster quant while trying to compile it(improved speed by 2-5% and code is clearer while trying to compile pasrts)
 c) moved helper functions out of the class. They're largely generic and
 this is less cluttered. May need to revisit how generic they are if new GPTQQuantizers are made.
 d) some improvements to the duplication checking and copying to be
 faster when possible (previously MultiTensor.unpad had checks but at the point that its called, we already did equality checks so we don't need them in unpad.
 e) fixed some bugs due to this not being in CI and things changing for
 int4wo tensor subclass.
 6) BC
  a) got rid of Int8DynActInt4WeightGPTQQuantizer since it was unused. Can re-add if desired.
  b) for other imports, maintained BC, previous impoirts from quantization/GPTQ.py now go to quantization/GPTQ/__init__.py
  c) InputRecorder -> LMEvalInputRecorder but left BC import in as an option.

Test Plan:

1) `python test_gptq.py`

note1: the skipped test test_gptq_quantizer_int4_weight_only also ran.
note2: we now have a CI ready test in test_gptq using the generic input recorder.

2) I verified that all activation match between old GPTQ (non MT) and current
   GPTQ, this can be seen by passing the test_gptq_quantizer_int4_weight_only as mentioned above but was also verified by comparing debug outputs and printing activation values for first 3 multi tensors.

3) eval benchmarks:

```shell

export MODEL_REPO=meta-llama/Llama-2-7b-chat-hf
python eval.py --checkpoint_path $CHECKPOINT_PATH/$MODEL_REPO/model.pth
python eval.py --checkpoint_path $CHECKPOINT_PATH/$MODEL_REPO/model.pth
--quantization int4wo-64
python eval.py --checkpoint_path $CHECKPOINT_PATH/$MODEL_REPO/model.pth
--quantization int4wo-gptq-64 --calibration_limit 10

export MODEL_REPO=meta-llama/Meta-Llama-3-8B
python eval.py --checkpoint_path $CHECKPOINT_PATH/$MODEL_REPO/model.pth
python eval.py --checkpoint_path $CHECKPOINT_PATH/$MODEL_REPO/model.pth
--quantization int4wo-64
python eval.py --checkpoint_path $CHECKPOINT_PATH/$MODEL_REPO/model.pth
--quantization int4wo-gptq-64 --calibration_limit 10
```

see README.md for results but they show GPTQ is working

Reviewers:

Subscribers:

Tasks:

Tags:


torchao/quantization/qat/linear.py
```diff
@@ -10,7 +10,7 @@
 import torch.nn.functional as F
 
 from torchao.dtypes.utils import is_device
-from torchao.quantization.GPTQ import (
+from torchao.quantization.linear_quant_modules import (
```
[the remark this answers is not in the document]
HDCharles: I can add that

torchao/prototype/GPTQ/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Results
+
+We tested the GPTQ implementation using the llama model in torchao/_models/llama with 10 calibration samples from lm_eval.
+
+| Technique:     | Llama-2-7b-chat-hf | Meta-Llama-3-8B |   |   |
+|----------------|--------------------|-----------------|---|---|
+| bf16           |             12.245 |           7.441 |   |   |
+| int4wo-64      |             12.876 |           8.316 |   |   |
+| gptq-int4wo-64 |             12.523 |           8.026 |   |   |
```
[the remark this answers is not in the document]
HDCharles: gah

torchao/_models/_eval.py
```diff
@@ -92,171 +115,39 @@ def tok_encode(self, string: str, **kwargs):
...
-        # Output `something` with the correct shape to keep eval going
-        return torch.randn(
-            (1, T, self.vocab_size), dtype=torch.bfloat16, device=self._device
-        )
-
     def _model_generate(self, context, max_length, eos_token_id):
         raise Exception("unimplemented")
 
 
-class InputRecorder(eval_wrapper):
```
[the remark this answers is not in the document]
HDCharles: No its LMEvalInputRecorder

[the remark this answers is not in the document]
HDCharles: i can make it BC

test/quantization/test_gptq.py
```diff
@@ -0,0 +1,202 @@
...
+from torchao.utils import (
+    TORCH_VERSION_AT_LEAST_2_4,
+)
+
+torch.manual_seed(0)
+
+
+class TestGPTQ(TestCase):
+    @unittest.skip("skipping until we get checkpoints for gpt-fast")
+    def test_gptq_quantizer_int4_weight_only(self):
```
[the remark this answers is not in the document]
HDCharles: we can upgrade it but this test and the exact value of the result allow us to lock in the numerics. I made a bunch of changes and i'd rather do more test enhancements in another PR, this test was my north star since I knew the old GPTQ implementation was working.

torchao/quantization/README.md
```diff
@@ -396,53 +396,26 @@ The `quantize_` and `autoquant` apis now automatically use our recommended induc
...
-    groupsize,
-)
-model.setup_caches(max_batch_size=1, max_seq_length=calibration_seq_length)
-model = quantizer.quantize(model, inputs).cuda()
+from torchao.prototype.GPTQ import MultiTensorInputRecorder, Int4WeightOnlyGPTQQuantizer
 
+model = get_model()
+
+# first gather inputs
+input_recorder=MultiTensorInputRecorder()
```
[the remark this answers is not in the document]
HDCharles: ruff failing me

torchao/_models/_eval.py
```diff
@@ -272,91 +163,47 @@ def record_inputs(
...
-    def get_inputs(self):
-        return self.inputs
+    def get_inputs(self):  # for BC
+        return self.get_recorded_inputs()
+
+    def get_recorded_inputs(self):
+        return self.base_input_recorder.get_recorded_inputs()
+
+    def get_recorded_args_and_kwargs(self):
+        return self.base_input_recorder.get_recorded_args_and_kwargs()
```
jerryzh168, line 173: nit: get_recorded_args(self) might be more consistent with get_recorded_args_and_kwargs?

does recorded inputs support recording multiple args?

torchao/optim/adam.py
```diff
@@ -180,7 +180,7 @@ def single_param_adam(
     bias_correction2 = 1 - beta2**step
 
     # keep high precision copy for param update
-    exp_avg_f32 = exp_avg.float().lerp(grad_f32, 1 - beta1)
+    exp_avg_f32 = exp_avg.float().lerp(grad_f32, 1 - beta1).float()
```
jerryzh168, line 183: is this relevant to the PR?

torchao/quantization/GPTQ/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+```python
+from torchao.quantization import MultiTensorInputRecorder, Int4WeightOnlyGPTQQuantizer
+
+model = get_model() # user provided function
+
+# first gather inputs
+input_recorder = MultiTensorInputRecorder()
+for i in range(calibration_limit):
+    args = get_next_input() # user provided function
```
jerryzh168, line 20: optional nit: might be useful to have some real data example (e.g. https://github.com/pytorch/ao/blob/c4250a4bfa495a260ea009e23a9b8202bb93b52d/torchao/prototype/awq/example.py#L21), but I'm also planning to take a look at this soon

jerryzh168 (approved):
Thanks, I think the updated API looks good

torchao/quantization/GPTQ/README.md
```diff
@@ -0,0 +1,74 @@
...
+    input_recorder(*args) # compare to model(*args)
+     # note: can do input_recorder(*args, **kwargs) if needed
+
+# then perform GPTQ
+quantizer = Int4WeightOnlyGPTQQuantizer() # quantization parameters like group_size can be set here
```
jerryzh168, line 25: also maybe adding some example on how people can adapt this to their own quantization might be useful as well I think

torchao/quantization/qat/linear.py
```diff
@@ -102,7 +102,7 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
             w = self.weight_fake_quantizer(self.weight)
         else:
             w = self.weight
-        return F.linear(x, w, self.bias)
+        return F.linear(x.to(w.dtype), w, self.bias)
```
jerryzh168, line 105: is this change needed?

torchao/quantization/__init__.py
```diff
@@ -191,9 +194,7 @@
     "PerRow",
     "PerToken",
     "LinearActivationQuantizedTensor",
-    "Int4WeightOnlyGPTQQuantizer",
     "Int4WeightOnlyQuantizer",
-    "Int8DynActInt4WeightGPTQQuantizer",
```
jerryzh168, line 196: this is still used in ET (code): https://github.com/pytorch/executorch/blob/66dfc46686ebcb317efdf13580e869e7e4e2f0cc/examples/models/llama/source_transformation/quantize.py#L184 so probably better to remove the this in ET before landing the PR

torchao/_models/_eval.py
HDCharles: i wanted to maintain the old API which is used in a few places just to get data.

torchao/quantization/qat/linear.py
HDCharles: i am completely flummoxed by this error i'm seeing without it. Somehow these changes broke sparsity, qat and low bit optimizers without actually touching any of their code. I'm at a loss as to how and trying to figure out whats going on. If i can figure out how to fix it without that then i'll do that but thus far that actually seemed to have fixed it.

torchao/optim/adam.py
HDCharles: gone

torchao/quantization/__init__.py
HDCharles: see https://github.com/pytorch/executorch/pull/11274

PR #2238, opened 2025-05-22
Try fixing CI by pinning pytest

atalman:
Looks like pytest is failing with Segfault even when I execute simple pytest --version
```
exec: line 14:   238 Segmentation fault      (core dumped) python -u -m pytest --version
```
Hence Pinning the pytest to previous version fixes the Segfault

Test PR: https://github.com/pytorch/ao/pull/2237
Failure: https://github.com/pytorch/ao/actions/runs/15176248032/job/42676994119?pr=2237#step:15:1427

We would probably need to followup on this to understand why this have happended, since latest pytest update was March 2, 2025: https://pypi.org/project/pytest/8.3.5/

### DIFF
--- a/.github/workflows/regression_test.yml
+++ b/.github/workflows/regression_test.yml
@@ -64,7 +64,7 @@ jobs:
             torch-spec: 'torch==2.5.1 --index-url https://download.pytorch.org/whl/cu121'
             gpu-arch-type: "cuda"
             gpu-arch-version: "12.6"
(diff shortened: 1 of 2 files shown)


jerryzh168 (approved):

drisspg (approved):

drisspg:
Is this saying that the regression test requries 7.4 but main is on 8.3? 

Also great job finding this!

PR #2240, opened 2025-05-22
Enable fp16+int4 mixed precission path for int4 xpu path with int zero point

liangan1:
**Backgroup**
For XPU device, when user select the int zero point, the _torch.ops.aten._weight_int4pack_mm_with_scales_and_zeros kernel_ operator will be used to do A16W4 computation. Both Afp16W4 and ABF16int4 are supported in this op on XPU device, while only the BF16 activation is supported in the torchAO now, In this PR we want to unlock the FP16 activation support.

### DIFF
--- a/torchao/dtypes/affine_quantized_tensor_ops.py
+++ b/torchao/dtypes/affine_quantized_tensor_ops.py
@@ -46,8 +46,8 @@ from torchao.dtypes.uintx.int4_cpu_layout import (
 from torchao.dtypes.uintx.int4_xpu_layout import (
     _linear_bf16_act_uint4_weight_float_zero_check,
     _linear_bf16_act_uint4_weight_float_zero_impl,
-    _linear_bf16_act_uint4_weight_int8_zero_check,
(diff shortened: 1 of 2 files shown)


liangan1:
@jerryzh168 can you help to review?

liangan1:
@EikanWang

liangan1:
@@EikanWang

jerryzh168 (approved):
can you add a test? maybe add one under https://github.com/pytorch/ao/blob/4d5f65711f7c53985d09e3a8c6aa8d8549f7d5a4/test/dtypes/test_affine_quantized.py#L299 for now, current tests are not very well structured

liangan1:
> can you add a test? maybe add one under
> 
> https://github.com/pytorch/ao/blob/4d5f65711f7c53985d09e3a8c6aa8d8549f7d5a4/test/dtypes/test_affine_quantized.py#L299
> 
> for now, current tests are not very well structured

Yes. I checked the test in test_affine_quantized.py. xpu is not enabled for a lot of UTs(no only for dtype), I will open a new pr to enable these UTs.

PR #2243, opened 2025-05-22
Rename AOPerModuleConfig to ModuleFqnToConfig

jerryzh168:
Summary:
to be more explicit on what this config means

not marking bc-breaking since we haven't officially advertised this API

Test Plan:
CI

Reviewers:

Subscribers:

Tasks:

Tags:


drisspg (approved):

metascroy:
@jerryzh168 I think this is BC breaking for the hummingbird models we published that use this API

PR #2253, opened 2025-05-23
Fix Per Row scaling for inference

drisspg:
Stacked PRs:
 * #2258
 * #2256
 * __->__#2253

--- --- ---

## Summary

In somewhere in the myriad of refactors we broke per-row scaling. This had no effect becuase dequant just so happend to work and a global scale, so we test now

### DIFF
--- a/test/dtypes/test_affine_quantized_float.py
+++ b/test/dtypes/test_affine_quantized_float.py
@@ -25,6 +25,7 @@ import torch
 from torch._inductor.test_case import TestCase as InductorTestCase
 from torch.testing._internal import common_utils
 
+from torchao.dtypes.floatx.float8_layout import Float8AQTTensorImpl
 from torchao.float8.float8_utils import compute_error
 from torchao.quantization import (
     Float8DynamicActivationFloat8WeightConfig,
@@ -42,6 +43,9 @@ from torchao.quantization.quant_api import (
 from torchao.quantization.quant_primitives import (
     MappingType,
     choose_qparams_affine,
+    choose_qparams_affine_float8,
+    dequantize_affine_float8,
+    quantize_affine_float8,
 )
 from torchao.utils import (
     is_sm_at_least_89,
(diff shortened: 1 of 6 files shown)


torchao/dtypes/floatx/float8_layout.py
```diff
@@ -333,13 +333,12 @@ def _linear_fp8_act_fp8_weight_impl(
     input_scale = input_tensor.tensor_impl.scale
     # Handle case where input tensor is more than 2D
     inpt_data = inpt_data.reshape(-1, inpt_data.shape[-1])
-
     # Handle rowwise case
     if _is_rowwise_scaled(weight_tensor):
         assert _is_rowwise_scaled(input_tensor), (
             "Input tensor must be rowwise block size"
         )
-        w_scale = w_scale.unsqueeze(-1).T
```
danielvegamyhre, line 342: Why was this necessary? Is this intentional?

torchao/quantization/quant_primitives.py
```diff
@@ -1970,20 +1970,38 @@ def choose_qparams_affine_float8(
...
+            block_size, tensor.shape
+        )
+        tensor_reshaped = tensor.view(shape_for_reduction)
+        max_abs = tensor_reshaped.abs().amax(dim=reduction_dims, keepdim=True)
+
+        scale = max_abs / quant_max
+        # Reshape scale back to match the expected output shape
+        # The scale tensor should have the same shape as the input divided by block_size
+        output_shape = [
+            input_size // block_size[i] if block_size[i] > 1 else input_size
```
danielvegamyhre, line 2000: I don't think we need the if/else here do we, since if block_size[i] is 1 and we divide by 1, it's still just input_size

torchao/quantization/quant_api.py
```diff
@@ -1434,18 +1434,21 @@ def _float8_weight_only_transform(
 _fp8_granularities = Union[PerTensor, PerRow]
 
 
-# Validate and process granularity input
 def _normalize_granularity(
     granularity: Optional[
-        Union[_fp8_granularities, Tuple[_fp8_granularities, _fp8_granularities]]
+        Union[
+            _fp8_granularities,
```
danielvegamyhre, line 1440: general comment/question not specific to this PR: is it conventional to have python type aliases in snake case like this? seems odd to me, I expected camel case

torchao/quantization/quant_primitives.py
```diff
@@ -2027,5 +2045,24 @@ def dequantize_affine_float8(
     # upcasted to `float32` to divide by the scale, since scale is a fp32 for float8 quantization.
     # In order to match numerics between eager and compile, we upcast manually here.
     fp8_tensor = tensor.to(torch.float32)
-    hp_tensor = fp8_tensor * scale
+    # For block-wise quantization, we need to broadcast the scale to match tensor dimensions
+    if scale.shape != tensor.shape:
+        # Calculate the block size from the shape difference
+        block_size = tuple(
+            tensor.shape[i] // scale.shape[i]
+            if scale.shape[i] != tensor.shape[i]
```
danielvegamyhre, line 2053: do we need this if/else? seems like if `scale.shape[i] == tensor.shape[i]`, then `tensor.shape[i] // scale.shape[i]` will already equal 1.

torchao/dtypes/floatx/float8_layout.py
drisspg: TBH I am not entirely sure what happend to all this code but rowwise scales was  stored in shape (N_rows) shape

torchao/quantization/quant_api.py
drisspg: yeah good call changed

test/dtypes/test_affine_quantized_float.py
```diff
@@ -297,21 +297,53 @@ def test_fp8_weight_dimension_warning(self):
...
     )
-    def test_mm_float8dq(self):
+    @common_utils.parametrize(
+        "in_features,out_features", [(512, 1024), (256, 768), (1024, 512)]
+    )
+    @common_utils.parametrize(
+        "leading_shape", [(1,), (8,), (16,), (2, 8,), (2, 2, 16,)]
+    )  # fmt: skip
+    @common_utils.parametrize("bias", [True, False])
+    def test_mm_float8dq(self, in_features, out_features, leading_shape, bias: bool):
```
danielvegamyhre, line 307: since this only tests PerRow maybe the test name should reflect that

torchao/float8/inference.py
```diff
@@ -106,3 +115,66 @@ def _is_rowwise_scaled(x) -> bool:
...
+            list[FP8Granularity],
+        ]
+    ],
+) -> Tuple[FP8Granularity, FP8Granularity]:
+    processed_granularity = None
+    if granularity is None:
+        processed_granularity = (PerTensor(), PerTensor())
+    elif isinstance(granularity, (PerTensor, PerRow)):
+        processed_granularity = (granularity, granularity)
+    elif isinstance(granularity, (tuple, list)) and len(granularity) == 2:
```
danielvegamyhre, line 137: Why do we support passing in different granularities for activations if weights, since they both must be the same?

drisspg: There was a todo, they need not be, and this was forward looking to things like deepseek and more granular varaints

torchao/quantization/quant_primitives.py
```diff
@@ -1970,20 +1970,37 @@ def choose_qparams_affine_float8(
...
+    block_size: Optional[Tuple[int, ...]] = None,
 ) -> torch.Tensor:
     """
     Calculates float8 scaling factor for the given high precision tensor, using tensorwise granularity.
 
     Args:
         tensor (torch.Tensor): Input tensor to be quantized.
         float8_dtype (torch.dtype): Data type of the quantized tensor (e.g., torch.float8_e4m3fn, torch.float8_e5m2).
+        scale_dtype (torch.dtype): Data type of the scaling factor (e.g., torch.float32).
+        block_size (Optional[Tuple[int, ...]]): Block size for block-wise quantization. If None, tensorwise quantization is used.
```
danielvegamyhre, line 1982: was support for blockwise quantization intentionally included in this PR? this looks like the quant primitive support for it but presumably other things are still needed like scaling granularity (which only is PerTensor or PerRow right now), etc.

drisspg: This primitive, can/should be generic to block-size - which is what I did arguablly I hsould update the doc string a little, but its up to the caller to pass in the right block_size which in this case is only 2 difffernt forms today per-tensor and per-row

danielvegamyhre (approved):

drisspg:
```
Processed prompts: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 5/5 [00:00<00:00, 15.06it/s, est. speed input: 108.45 toks/s, output: 195.81 toks/s]
Prompt: 'Why is Pytorch 2.0 the best machine learning compiler?', Generated text: " Ok,\nI am five months after changing? This one's"
Prompt: 'Hello, my name is', Generated text: ' Samuel Rodriguez.\r\n\r\nIf you want to know some regulars of'
Prompt: 'The president of the United States is', Generated text: ' revered has taken old wine beefs, morgs.\n'
Prompt: 'The capital of France is', Generated text: ' ruling Tauresearchednow.com offers a regularsolutions'
Prompt: 'The future of AI is', Generated text: ' moving... I thought about implementing logic, physics are one transform'
[rank0]:[W524 10:43:22.504024005 ProcessGroupNCCL.cpp:1538] Warning: WARNING: destroy_process_group() was not called before progr
```
Per Tensor is broken, it has to do with how we are slicing

torchao/quantization/quant_primitives.py
```diff
@@ -1970,21 +1970,92 @@ def choose_qparams_affine_float8(
...
     """
+    quant_max = torch.finfo(float8_dtype).max
     # only tensorwise scaling is supported for now:
-    quant_min, quant_max = torch.finfo(float8_dtype).min, torch.finfo(float8_dtype).max
-    min_val_neg = torch.min(tensor)
-    max_val_pos = torch.max(tensor)
-    max_val_pos = torch.max(-min_val_neg, max_val_pos)
-    scale = max_val_pos / (float(quant_max - quant_min) / 2)
-    return scale.to(dtype=scale_dtype)
+    if block_size is None:
```
jerryzh168, line 1986: is this per tensor? should we standardize on this? some other place are probably using -1 to represent this

PR #2257, opened 2025-05-23
Patch the _is_conv_node function

cccclai:
Summary: Add the conv padding ops in torch/ao only, will add a separate PR for the ones in pytorch

Differential Revision: D75323215


cccclai:
It seems like I can't add reviews. Also can't add labels.

Issue #2239, opened 2025-05-22
Tensor Subclass + VLLM Compile


drisspg:
# VLLM Torch.compile Issue Tracker

## Summary

This document tracks the existing issue with the way VLLM uses `torch.compile` and tensor subclasses.

**TLDR**: VLLM doesn't setup `aotdispatch` correctly, causing subclass flattening to not take place.

@zou3519 has theoretically fixed this issue with: https://github.com/vllm-project/vllm/pull/17057, which enables standalone compile.

### MXFP4 Model Testing

**Command**: 
```bash
python vllm/sample_output.py --model_name "data/mxfp4-Qwen2-7B-Instruct" --compile True
```

**Issue**: The swizzle kernel needs to be enabled because without it, the dynamic control flow in `to_blocked` will error.

**Error encountered**:
```
torch._inductor.exc.InductorError: RuntimeError: Failed to import /tmp/torchinductor_drisspg/zw/czwhsx3gfghqtfk5wbhit2p2xpgr3yxil54pbsi2mtnirmrnrths.py
IndentationError: unexpected indent (czwhsx3gfghqtfk5wbhit2p2xpgr3yxil54pbsi2mtnirmrnrths.py, line 173)

Set TORCHDYNAMO_VERBOSE=1 for the internal stack trace (please do this especially if you're reporting a bug to PyTorch). For even more developer context, set TORCH_LOGS="+dynamo"
```

**Note**: I suspect that it doesn't like inlining the  user defined Triton kernel correctly.
cc @oulgen  if you have any ideas here

### FP8 Model Testing

**Command**:
```bash
python vllm/sample_output.py --model_name "data/fp8-Qwen2-7B-Instruct" --compile True
```

**Results**: ✅ **Working**

**First run compilation time**:
```
INFO 05-21 22:29:11 [monitor.py:33] torch.compile takes 136.26 s in total
```

**Sample outputs**:
- **Prompt**: 'Why is Pytorch 2.0 the best machine learning compiler?'  
  **Generated**: ' PyTorch 2.0, currently not released officially, is anticipated to be a significant upgrade in several areas including performance, features, and usability...'

- **Prompt**: 'Hello, my name is'  
  **Generated**: ' Mandy Lowry, and I am a certified financial planner and a long-time friend of the Girls' Town...'

- **Prompt**: 'The president of the United States is'  
  **Generated**: " the leader of the government of the United States. The president is also the commander-in-chief of the United States Armed Forces..."

[...]

PR #2258, opened 2025-05-24
integration-vllm-test

drisspg:
Stacked PRs:
 * __->__#2258

--- --- ---

integration-vllm-test

### DIFF
--- a/test/integration/test_vllm.py
+++ b/test/integration/test_vllm.py
@@ -1,0 +1,252 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+
+import importlib.util
+import os
+import random
+import shutil
+from pathlib import Path
+from typing import List
+
+import numpy as np
+import pytest
+import torch
+
+from torchao.utils import TORCH_VERSION_AT_LEAST_2_7
+
+if not TORCH_VERSION_AT_LEAST_2_7:
+    pytest.skip("Requires PyTorch 2.7 or higher", allow_module_level=True)
+
+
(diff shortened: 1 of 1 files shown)


test/integration/test_vllm.py
```diff
@@ -0,0 +1,245 @@
...
+import pytest
+import torch
+
+from torchao.utils import TORCH_VERSION_AT_LEAST_2_7
+
+if not TORCH_VERSION_AT_LEAST_2_7:
+    pytest.skip("Requires PyTorch 2.7 or higher", allow_module_level=True)
+
+
+VLLM_AVAILABLE = importlib.util.find_spec("vllm") is not None
```
jerryzh168, line 23: does this work for you?

test/integration/test_vllm.py
```diff
@@ -0,0 +1,245 @@
...
+    @classmethod
+    def setup_class(cls):
+        """Set up test environment."""
+        # Set seeds for reproducibility
+        cls.set_seed(42)
+
+        # Set vLLM environment variables
+        os.environ["VLLM_USE_V1"] = "1"
+        os.environ["VLLM_ENABLE_V1_MULTIPROCESSING"] = "0"
+        os.environ["VLLM_TEST_STANDALONE_COMPILE"] = "1"
```
jerryzh168, line 106: are you missing `VLLM_DISABLE_COMPILE_CACHE`?

can you write down for each flag about whether they are always required v.s. temporary etc.

jerryzh168 (approved):

test/integration/test_vllm.py
drisspg: I made note here: https://github.com/pytorch/ao/issues/2239

You can see that `VLLM_TEST_STANDALONE_COMPILE` which wil lultimately be the default doesn't makes it so that we can compile subclasses

PR #2269, opened 2025-05-29
[float8 training] remove duplicate override for view

danielvegamyhre:
I noticed there's 2 overrides for `aten.view.default`:

1. standard desugaring op at the top.
2. special `float8_view` override lower down. 

The 2nd one overwrites the 1st entry in the FLOAT8_OPS_TABLE, so we can remove it.

### DIFF
--- a/torchao/float8/float8_ops.py
+++ b/torchao/float8/float8_ops.py
@@ -100,7 +100,6 @@ def implements(aten_ops):
 
 @implements(
     [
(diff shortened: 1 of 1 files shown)


danielvegamyhre:
> how about also changing the definition of `implements` to throw an exception if a value is overridden, to prevent future similar bugs?

Good idea, done

PR #2280, opened 2025-05-30
Fix QAT range learning, ensure scales get gradients

andrewor14:
**Summary:** The previous `_GenericFakeQuantized` nulled all gradients except the ones for the input. This is problematic for range learning because scales and zero points are now `nn.Parameters` and actually require gradients. This commit fixes this by reducing the scope of the `autograd.Function` to `torch.round` only, so QAT can just call the fake quantization primitives directly.

Note: Part of the dequantize math currently casts the inputs and the zero points to int32. However, autograd doesn't work with integer math and this part of the code path is now visible to autograd. To make this work, this commit also removes this dtype cast.

Note: This change means we no longer do cachemask and so our numerics no longer matches those of pytorch/pytorch's
fake quantization ops.

**Test Plan:**
Updated the following test to check for scales and weights being updated:
python test/quantization/test_qat.py -k test_qat_range_learning

### DIFF
--- a/test/quantization/test_qat.py
+++ b/test/quantization/test_qat.py
@@ -49,7 +49,6 @@ from torchao.quantization.qat.linear import (
 from torchao.quantization.qat.utils import (
     _fake_quantize_per_channel_group,
     _fake_quantize_per_token,
-    _GenericFakeQuantize,
     _get_qmin_qmax,
 )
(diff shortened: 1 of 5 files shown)


jerryzh168:
looks like there are some CI failures

PR #2214, opened 2025-05-15
Fixes MX formats build for blackwell

syed-ahmed:
When `TORCH_CUDA_ARCH_LIST` doesn't have `9.0a`, `setup.py` first compiles the files in `sources` and then compiles the files in `cutlass_90a_sources`, overwriting the `_C.abi3.so` that was produced before:
```
    [1/7] /usr/local/cuda/bin/nvcc --generate-dependencies-with-compile --dependency-output /opt/pytorch/ao/build/temp.linux-aarch64-cpython-312/torchao/csrc/cuda/tensor_core_tiled_layout/tensor_core_tiled_layout.o.d -I/usr/local/lib/python3.12/dist-packages/torch/include -I/usr/local/lib/python3.12/dist-packages/torch/include/torch/csrc/api/include -I/usr/local/cuda/include -I/usr/include/python3.12 -c -c /opt/pytorch/ao/torchao/csrc/cuda/tensor_core_tiled_layout/tensor_core_tiled_layout.cu -o /opt/pytorch/ao/build/temp.linux-aarch64-cpython-312/torchao/csrc/cuda/tensor_core_tiled_layout/tensor_core_tiled_layout.o -D__CUDA_NO_HALF_OPERATORS__ -D__CUDA_NO_HALF_CONVERSIONS__ -D__CUDA_NO_BFLOAT16_CONVERSIONS__ -D__CUDA_NO_HALF2_OPERATORS__ --expt-relaxed-constexpr --compiler-options ''"'"'-fPIC'"'"'' -DNDEBUG -O3 -t=0 -std=c++17 -DTORCHAO_USE_CUTLASS -I/opt/pytorch/ao/third_party/cutlass/include -I/opt/pytorch/ao/third_party/cutlass/tools/util/include -I/opt/pytorch/ao/torchao/csrc/cuda -DCUTE_USE_PACKED_TUPLE=1 -DCUTE_SM90_EXTENDED_MMA_SHAPES_ENABLED -DCUTLASS_ENABLE_TENSOR_CORE_MMA=1 -DCUTLASS_DEBUG_TRACE_LEVEL=0 --ftemplate-backtrace-limit=0 -DTORCH_API_INCLUDE_EXTENSION_H -DPy_LIMITED_API=0x03090000 -DTORCH_EXTENSION_NAME=_C -gencode=arch=compute_100,code=sm_100 -gencode=arch=compute_100a,code=sm_100a -gencode=arch=compute_90,code=sm_90
[...]

PR #2284, opened 2025-05-31
Fix Bug in MX Builds

drisspg:
Stacked PRs:
 * __->__#2284

--- --- ---

Fix Bug in MX Builds

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -223,6 +223,55 @@ def check_submodules():
         )
 
 
+def get_cuda_version_from_nvcc():
+    """Get CUDA version from nvcc if available."""
+    try:
+        result = subprocess.check_output(
+            ["nvcc", "--version"], stderr=subprocess.STDOUT
+        )
+        output = result.decode("utf-8")
+        # Look for version line like "release 12.6"
+        for line in output.split("\n"):
+            if "release" in line.lower():
+                parts = line.split()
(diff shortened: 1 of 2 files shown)


gau-nernst (approved):

gau-nernst:
Actually as long as we have a sufficiently new CUDA toolkit (12.8 I think?), we can compile for sm100a? Regardless of actual compute capability available at compile time.

The downside is that compile time is longer for non-sm100a users building from source.

drisspg:
So there was 2 problems:
1. the accidental downgrade of Cutlass fixed in the PR
2. The second is that cuda extensions doesn't let you compile w/ perfile flags, In the past for cutlass utilzing the 'a' variants if you try to compile with anything but that sm+a you would get a compile time error, e.g. https://github.com/pytorch/ao/pull/2214

I think the problem right now is that I unconditionally remove  remove the sources 

e.g. 
if I set 90a, and 100a I get

INFO:root:copying build/lib.linux-x86_64-cpython-312/torchao/_C.abi3.so -> torchao
INFO:root:copying build/lib.linux-x86_64-cpython-312/torchao/_C_cutlass_90a.abi3.so -> torchao
INFO:root:copying build/lib.linux-x86_64-cpython-312/torchao/_C_cutlass_100a.abi3.so -> torchao

which is right

PR #2285, opened 2025-05-31
Build mxfp4 kernel for sm120a

gau-nernst:
## Update (2025/06/14)

Added benchmark script, courtesy to @drisspg. Perhaps you can run it for B200 as well

At 575W (default)

575w_mx_8192_8192_20250614_110313

At 400W - this is just for me since I'm running this card at 400W

400w_mx_8192_8192_20250614_110032

For reference (5090 is last column)

image

At stock power limit, the card can comfortably reach speed of light. The interesting bit is in MX-FP8 -> it looks like the kernel from `torch._scaled_mm()` (I guess it's from CuDNN/CuBLAS?) is using FP16 accumulate 🤔

## Update (2025/06/11)

I narrowed down the issue to template - if the kernel is inside a templated function, even if I don't use any template arguments, I will get the runtime error below (`cudaFuncSetAttribute() returned error: invalid resource handle`). It might be an issue with cutlass or my environment (nvcc version, compiler...).

Hence, the solution is to create a separate source file for sm120a, without any templated functions. When we support nvfp4 in the future, we can either manually duplicate the code again, use macro, or have a python script to codegen the cutlass kernel creation.

Other details of this PR:
- Add `sm120a` extension
- Modify torch library loading logic: For cutlass kernels with architecture-specific targets (smXY**a**), it only loads those matching the current GPU's compute capability. The limitation of this code is that it won't work correctly for multi-GPU setup with different compute capabilities.

Other alternatives that I have considered for the torch library loading logic:
- Basically we need a runtime check to select sm100a or sm120a kernel
- Due to `setuptools.Extension`'s limitation, sm100a and sm120a kernels must stay in separate shared library files. This eliminates the option of doing runtime check in C++.
- Hence, the runtime check must be in Python. I think of 2 options
  - 1. Name 2 different ops (e.g. `mx_fp4_bf16_sm100a` and `mx_fp4_bf16_sm120a`), and dispatch the correct op in Python
  - 2. Use the same op name, but only load the torch library corresponding to current GPU at startup -> **I go with this approach**

## Original (2025/05/31)

Just making some quick changes here to see if I can build mxfp4 kernel on 5090 (sm120). Eventually this will be put under `torchao._C_cutlass_120a`?

Setting `-DCUTLASS_DEBUG_TRACE_LEVEL=1` so I can see debug trace.

To build (using `torch==2.8.0.dev20250530+cu128`)

```
TORCH_CUDA_ARCH_LIST=12.0a uv pip install -e . -v --no-build-isolation
```

Running `pytest test/prototype/mx_formats/test_mx_mm.py -v`

```
/home/thien/code/ao/third_party/cutlass/include/cutlass/gemm/device/gemm_universal_adapter.h:244    workspace_bytes: 0
/home/thien/code/ao/third_party/cutlass/include/cutlass/gemm/device/gemm_universal_adapter.h:312  GemmUniversal::initialize() - workspace 0, stream: null
/home/thien/code/ao/third_party/cutlass/include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_cooperative.hpp:201  to_underlying_arguments():
/home/thien/code/ao/third_party/cutlass/include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_cooperative.hpp:214    WARNING: Arguments do not include a valid SM count.
  For optimal performance, populate the arguments KernelHardwareInfo struct with the SM count.
/home/thien/code/ao/third_party/cutlass/include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_cooperative.hpp:218  to_underlying_arguments(): Setting persistent grid SM count to 170
/home/thien/code/ao/third_party/cutlass/include/cutlass/gemm/kernel/sm90_gemm_tma_warpspecialized_cooperative.hpp:224    WARNING: Arguments do not include a valid max cluster count.
  For optimal performance, populate the arguments KernelHardwareInfo struct with the max_active_clusters.
/home/thien/code/ao/third_party/cutlass/include/cutlass/gemm/device/gemm_universal_adapter.h:336    Setting smem size to 101376
/home/thien/code/ao/third_party/cutlass/include/cutlass/gemm/device/gemm_universal_adapter.h:343    cudaFuncSetAttribute() returned error: invalid resource handle
```

`cudaFuncSetAttribute() returned error: invalid resource handle` means that the function is invalid? https://github.com/NVIDIA/cutlass/blob/ad7b2f5e84fcfa124cb02b91d5bd26d238c0459e/include/cutlass/gemm/device/gemm_universal_adapter.h#L338, which is quite strange...

For reference, I can build and run the example from Cutlass here https://github.com/NVIDIA/cutlass/blob/v3.9.2/examples/79_blackwell_geforce_gemm/79a_blackwell_geforce_nvfp4_bf16_gemm.cu. The changes in this PR has been taken from this example. When building with `CUTLASS_DEBUG_TRACE_LEVEL=1`, there are also warnings in `sm90_gemm_tma_warpspecialized_cooperative.hpp`, so that is probably not the issue.

@drisspg

cc @alexsamardzic in case you faced this error with Cutlass before


drisspg:
The first thing that comes to mind is that example is doing NVfp4 where all our recipes are doing MXfp4, e.g. https://github.com/pytorch/ao/pull/2285/files#diff-e155558499c3b1fbab1b5d3b60f032bf1e636908a8ef50a1de33bff518107019R240-R241 needs to change as well. For inference we have MXFP8 and MXFP4 support I am planning to add an NVFP4 scaling recipe next, that being said I would imagine that MXFP4 is supported on 5090..


cc @syed-ahmed

syed-ahmed:
Per cutlass docs, I believe MXFP4 is supported in 5090: https://github.com/NVIDIA/cutlass/blob/9d165a3b8ef446a7ff3db198413f82bcb83f46fe/media/docs/cpp/blackwell_functionality.md#blackwell-sm120-gemms

However note the section that talks about the differences with sm100. So it's possible we need more changes to the kernel in torch ao. Also what CUDA version are you using? I'd assume you'd need a fairly recent CUDA version. I'll try to guide more next week.

gau-nernst:
@syed-ahmed I'm using CUDA 12.9

The strange thing is that the cutlass example works, but the one in torchao doesn't. I carefully compared the two, and I don't spot any difference in the template arguments.

syed-ahmed:
How about the test? Are the inputs similar to the cutlass example?

setup.py
```diff
@@ -557,12 +562,29 @@ def get_extensions():
...
                 "mx_kernels",
-                "mx_fp_cutlass_kernels.cu",
+                "mx_fp_cutlass_kernels_sm100a.cu",
             ),
         ]
         # Remove from main sources to prevent compilation with other architectures
         sources = [
-            s for s in sources if os.path.basename(s) != "mx_fp_cutlass_kernels.cu"
+            s
+            for s in sources
```
drisspg: nit I wouldn't hate a little helper func defined above to do all this stripping

torchao/__init__.py
```diff
@@ -25,8 +25,21 @@
 
     so_files = list(Path(__file__).parent.glob("_C*.so"))
     if len(so_files) > 0:
+        compute_capability = (
+            torch.cuda.get_device_capability() if torch.cuda.is_available() else None
+        )
+
         for file in so_files:
+            # only load architecture-specific target if the current GPU matches that target
```
drisspg, line 33: 👍

torchao/csrc/cuda/mx_kernels/mx_fp_cutlass_kernels_sm120a.cu
```diff
@@ -0,0 +1,195 @@
...
+
+#if defined(BUILD_MX_KERNELS_CUTLASS)
+namespace {
+
+using namespace cute;
+
+void run_gemm(at::Tensor& a, at::Tensor& b, at::Tensor& a_scale,
+             at::Tensor& b_scale, at::Tensor& out, int M, int K, int N) {
+
+  using MmaTileShape        = Shape<_128,_128,_128>;
```
drisspg, line 44: haven't looked in depth too much at what args are avail curious  how much tuning is needed here

gau-nernst: Here are the list of possible tile sizes https://github.com/NVIDIA/cutlass/blob/main/media/docs/cpp/blackwell_functionality.md#tile-size

benchmarks/mx_formats/mm_bench.py
```diff
@@ -0,0 +1,200 @@
...
+        result = run_experiment(config)
+        results.append(Experiment(config=config, result=result))
+    print_results(results, save_path)
+
+    if graph and save_path is not None:
+        df = pd.read_csv(save_path)
+        plot_tflops_comparison(df, save_path)
+
+
+if __name__ == "__main__":
```
vkuzo: can we unify this code with https://github.com/pytorch/ao/blob/main/benchmarks/float8/bench_matmul.py instead?  I know the path says `float8` but it would be good to have it all in one place.

gau-nernst: I have added mxfp4 to that script. You can review it. Thank you

vkuzo: thank you!

benchmarks/float8/bench_matmul.py
```diff
@@ -107,35 +114,55 @@ def run(
 
         del A
 
-        # raw float8 matmul (upper bound for what we can achive in eager mode)
-        # TODO(future): add e5m2
-        d1, d2, d3 = torch.float8_e4m3fn, torch.float8_e4m3fn, dtype
-        A = torch.zeros(M, K, device=device, dtype=d1)
-        B = torch.zeros(K, N, device=device, dtype=d2).t().contiguous().t()
+        if use_fp4:
+            A = torch.zeros(M, K // 2, device=device, dtype=torch.int8).view(
```
drisspg: just now looking at this but the performance on Nvidia w/ zeros vs non zero filled data is very very different

I think in a follow up PR we should make the zero filled an option vs randn distributed

benchmarks/mx_formats/mm_bench.py
gau-nernst: Any changes you want to make? If not, I will merge 🙏

benchmarks/float8/bench_matmul.py
vkuzo: can even be in this PR, I'd vote for randn as default

gau-nernst: 1. Should there be an option to choose between randn and zeros? Or just use randn for everything
2. For FP4, should I do randn in FP32/BF16 then convert to FP4, or just do `randint(0, 255, dtype=uint8).view(float4_e2m1fn_x2)`?

drisspg: ^ I think the option for zeros can come later

2. I would do the first randn in Higher precision then cast

gau-nernst: Do you mind taking a final look then merge? The failing tests seem unrelated.

jerryzh168:
sorry this breaks internal CI, we'' revert first to unblock diff train

```
clang++: warning: argument unused during compilation: '-pie' [-Wunused-command-line-argument]
ld.lld: error: duplicate symbol: torchao::mx_fp4_bf16(at::Tensor, at::Tensor, at::Tensor, at::Tensor)
>>> defined at __stripped__/torchao/csrc/cuda/mx_kernels/mx_fp_cutlass_kernels_sm100a.cu.pic.stripped.o:(torchao::mx_fp4_bf16(at::Tensor, at::Tensor, at::Tensor, at::Tensor)) in archive buck-out/v2/gen/fbcode/38bba46f0633e4ed/pytorch/ao/___C__/lib_C.stripped.pic.a
>>> defined at __stripped__/torchao/csrc/cuda/mx_kernels/mx_fp_cutlass_kernels_sm120a.cu.pic.stripped.o:(.text.unlikely._ZN7torchao11mx_fp4_bf16EN2at6TensorES1_S1_S1_+0x0) in archive buck-out/v2/gen/fbcode/38bba46f0633e4ed/pytorch/ao/___C__/lib_C.stripped.pic.a
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
```

gau-nernst:
Yea I guess we need to do something like this, if we don't use cmake to build each source file with different flags

https://github.com/pytorch/pytorch/blob/40a785103cf94a1dbc3e0e43d1ed6c41fb60bedb/aten/src/ATen/native/cuda/RowwiseScaledMM.cu#L4

PR #2286, opened 2025-05-31
[optim] Fix bug when default dtype is BF16

gau-nernst:
Happened in #2235

When `torch.set_default_dtype(torch.bfloat16)` is set, `qmap = torch.tensor(...)` becomes BF16. Hence, dequantizing the low-bit tensor subclass results in BF16, leading to dtype mismatch in Adam logic.


gau-nernst:
@msaroufim Can I get a stamp on this? Thank you!

msaroufim (approved):

PR #2289, opened 2025-06-02
Add Float8ActInt4WeightQATQuantizer

andrewor14:
**Summary:** This commit adds a QAT quantizer that performs float8 dynamic activation + int4 symmetric per channel weight fake quantization. Note that there is no corresponding config for float8 QAT yet. This will be added in a future PR.

**Test Plan:**
python test/quantization/test_qat.py -k test_float8_fake_quantize
python test/quantization/test_qat.py -k test_qat_fp8a4w_quantizer

### DIFF
--- a/test/quantization/test_qat.py
+++ b/test/quantization/test_qat.py
@@ -17,6 +17,9 @@ from parameterized import parameterized
 from torch.ao.quantization.fx._decomposed import quantized_decomposed_lib  # noqa: F401
 
 from torchao import quantize_
+from torchao.float8.config import ScalingGranularity
+from torchao.float8.float8_scaling_utils import hp_tensor_to_float8_dynamic
+from torchao.float8.float8_tensor import LinearMMConfig
 from torchao.quantization.granularity import (
     PerAxis,
     PerGroup,
@@ -40,15 +43,18 @@ from torchao.quantization.qat.embedding import (
 )
 from torchao.quantization.qat.fake_quantizer import (
     FakeQuantizer,
+    _Float8RowwiseActivationFakeQuantizer,
 )
 from torchao.quantization.qat.linear import (
     FakeQuantizedLinear,
+    Float8ActInt4WeightQATQuantizer,
     Int4WeightOnlyQATLinear,
     Int8DynActInt4WeightQATLinear,
 )
 from torchao.quantization.qat.utils import (
     _fake_quantize_per_channel_group,
     _fake_quantize_per_token,
+    _Float8RowwiseFakeQuantize,
     _get_qmin_qmax,
 )
 from torchao.quantization.quant_api import (
@@ -68,6 +74,7 @@ from torchao.quantization.unified import (
 )
 from torchao.quantization.utils import (
     _get_per_token_block_size,
+    compute_error,
     get_group_qparams_symmetric,
     get_groupwise_affine_qparams,
     groupwise_affine_quantize_tensor,
@@ -1474,7 +1481,6 @@ class TestQAT(unittest.TestCase):
         numerics that match exactly over N trials.
         """
         from torchao.quantization.qat import Int8DynActInt4WeightQATQuantizer
-        from torchao.quantization.utils import compute_error
 
         num_trials = 1000
         group_size = 16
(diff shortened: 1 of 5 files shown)


torchao/quantization/qat/fake_quantizer.py
```diff
@@ -186,3 +189,32 @@ def __repr__(self) -> str:
...
+            scale = tensor_to_scale(
+                x,
+                self.FLOAT8_DTYPE,
+                scaling_granularity=self.scaling_granularity,
+                axiswise_dim=get_maybe_axiswise_dim(-1, self.scaling_granularity),
+            )
+            x_fq = x / scale
+            x_fq = x_fq.to(torch.float8_e4m3fn).to(original_dtype)
+            x_fq = x_fq * scale
+            return x_fq.to(original_dtype)
```
andrewor14: Hi @vkuzo, can you check the math here for fp8 QAT?

torchao/quantization/qat/fake_quantizer.py
```diff
@@ -186,3 +189,32 @@ def __repr__(self) -> str:
...
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        if self.enabled:
+            original_dtype = x.dtype
+            scale = tensor_to_scale(
+                x,
+                self.FLOAT8_DTYPE,
+                scaling_granularity=self.scaling_granularity,
+                axiswise_dim=get_maybe_axiswise_dim(-1, self.scaling_granularity),
+            )
+            x_fq = x / scale
```
vkuzo: this should be a multiply instead of a divide

source: https://github.com/pytorch/ao/blob/f0f1f6ce470d2854eb8fafc5dda3b4ffd80f1fe5/torchao/float8/float8_tensor.py#L153

you also should clamp before casting, source: https://github.com/pytorch/ao/blob/f0f1f6ce470d2854eb8fafc5dda3b4ffd80f1fe5/torchao/float8/float8_tensor.py#L154

torchao/quantization/qat/fake_quantizer.py
```diff
@@ -186,3 +189,32 @@ def __repr__(self) -> str:
...
+    """
+
+    FLOAT8_DTYPE = torch.float8_e4m3fn
+
+    def __init__(self, scaling_granularity: ScalingGranularity):
+        super().__init__()
+        self.enabled = True
+        self.scaling_granularity = scaling_granularity
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
```
vkuzo, line 201: is there a backward that should be defined somewhere with STE?

torchao/quantization/qat/fake_quantizer.py
vkuzo: I'd recommend adding a unit test on this logic, you can check for SQNR >= ~20

andrewor14: added

test/quantization/test_qat.py
```diff
@@ -17,6 +17,9 @@
 from torch.ao.quantization.fx._decomposed import quantized_decomposed_lib  # noqa: F401
 
 from torchao import quantize_
+from torchao.float8.config import ScalingGranularity
```
drisspg, line 20: I kinda hate that we have ScalingGranuliarty and Ganularity of the other FP8 inference APIs

drisspg (approved):

test/quantization/test_qat.py
vkuzo: I think this is worth fixing before landing.  @andrewor14 , how about just using rowwise scaling (since I assume that the one you want) and removing the option to confugure it? That will at least keep this problem away from the BC surface of QAT in a way that we can more easily fix later.

andrewor14: Yeah sure

andrewor14:
Removed ScalingGranularity from the public API `Float8ActInt4WeightQATQuantizer`, please have another look @vkuzo

torchao/quantization/qat/utils.py
```diff
@@ -16,6 +19,37 @@
...
+        x: torch.Tensor,
+        float8_dtype: torch.dtype,
+        scaling_granularity: ScalingGranularity,
+        axiswise_dim: Optional[int],
+    ):
+        original_dtype = x.dtype
+        scale = tensor_to_scale(
+            x,
+            float8_dtype,
+            scaling_granularity=scaling_granularity,
```
vkuzo: IMO better to just put the granularity you want here to also remove it from the middle layer

andrewor14:
Ok, I'm merging this. The latest commit doesn't use any float8 training classes or functions. The implementation is also hidden so we can always change this if needed. Please let me know if there are any follow-up issues or concerns @vkuzo @drisspg

PR #2292, opened 2025-06-03
Removing DocBlock to unblock MXFP4 w/ Unwrap Tensor

drisspg:
Stacked PRs:
 * __->__#2292

--- --- ---

Removing DocBlock to unblock MXFP4 w/ Unwrap Tensor

See: https://github.com/pytorch/pytorch/issues/155006 for more details


drisspg:
Richard has some details, essentially pre-compile sets some configs in inductor that are not as tested

zou3519 (approved):

PR #1865, opened 2025-03-10
remove the `set_inductor_config` argument of `quantize_`.

vkuzo:
Summary:

Finishing the migration in https://github.com/pytorch/ao/issues/1715.  Specifically:
* the `set_inductor_config` argument of `quantize_` is removed
* the following workflows get a `set_inductor_config` argument, defaulting to True: AWQUIntXConfig, IntNWeightOnlyConfig, SmoothQuantConfig, Int8DynamicActivationInt4WeightConfig, Int4DynamicActivationInt4WeightConfig, GemliteUIntXWeightOnlyConfig, Int4WeightOnlyConfig, Int8WeightOnlyConfig, Int8DynamicActivationInt8WeightConfig, Float8WeightOnlyConfig, Float8DynamicActivationFloat8WeightConfig, Float8DynamicActivationFloat8WeightConfig, Float8StaticActivationFloat8WeightConfig, Float8StaticActivationFloat8WeightConfig, UIntXWeightOnlyConfig, FPXWeightOnlyConfig.  I skipped quantized training and non-AQT workflows from this, since I expect them to not need this setting.

Before

```python
def quantize_(
    model: torch.nn.Module,
    config: AOBaseConfig,
    filter_fn: Optional[Callable[[torch.nn.Module, str], bool]] = None,
    set_inductor_config: Optional[bool] = None,
    device: Optional[torch.types.Device] = None,
):  ...
```

After

```python
def quantize_(
    model: torch.nn.Module,
    config: AOBaseConfig,
    filter_fn: Optional[Callable[[torch.nn.Module, str], bool]] = None,
    device: Optional[torch.types.Device] = None,
):  ...

# see top of PR description for list of affected configs
class Int8DynamicActivationInt4WeightConfig(AOBaseConfig):
    ...
    set_inductor_config: bool = True
```

Test Plan:

```
pytest test/quantization/test_quant_api.py -s -x -k test_workflow_e2e_numerics
```

Reviewers:

Subscribers:

Tasks:

Tags:

PR #2294, opened 2025-06-03
make bfs_graph_trace as internal function

Gasoonjia:
Summary: makes bfs_trace_with_node_process as a internal graph utils.

Differential Revision: D75888360

### DIFF
--- a/test/quantization/pt2e/test_numeric_debugger.py
+++ b/test/quantization/pt2e/test_numeric_debugger.py
@@ -22,7 +22,7 @@ from torchao.quantization.pt2e import (
     generate_numeric_debug_handle,
     prepare_for_propagation_comparison,
 )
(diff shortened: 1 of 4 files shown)


larryliu0820:
Looks ok to me? @jerryzh168

larryliu0820 (approved):

Gasoonjia:
what's the best release note for this pr? it introduces BC issue for users and should have a release note label.

janeyx99 (commented):
Looks good to me, please do write a BC breaking note for the PR body for best visibility. The reason for it becoming private would be that it was never intended to be public and that the usage of this function by end users is not expected to be high.

jerryzh168 (approved):
see an example for bc-breaking note here: https://github.com/pytorch/ao/pull/1865

but I don't think there are users using this right now so probably OK to skip as well

Issue #2228, opened 2025-05-20
[Quant] Can quant not be decomposed on inductor?


shiyang-weng:
torch.ops.torchao.dequantize_affine decomposed to convert_element_type and mul.
Inductor will do constant_fold before pattern matching
On constant_fold, inductor replace fp8 weight and some previous operations with fp32 weight
Is this as expected?

Now register_decomposition on [register_decomposition](https://github.com/pytorch/ao/blob/96aec6a3e713687c1728a20a08d5c54db0344377/torchao/utils.py#L226)

This sample test can reproduce the issue

```python

import os

os.environ["OMP_NUM_THREADS"] = "1"
os.environ["TORCHINDUCTOR_FREEZING"] = "1"
os.environ["TORCH_COMPILE_DEBUG"] = "0"
os.environ["TORCHDYNAMO_PRINT_GUARD_FAILS"] = "0"

from typing import Callable, List, Optional, Union
import torch
from torch import nn
import torchao
#import torchao.quantization.pt2e.quantizer.x86_inductor_quantizer as xiq

def dequantize_per_tensor(
        input: torch.Tensor,
        scale: torch.Tensor,
        output_dtype: torch.dtype
) -> torch.Tensor:
    res = torch.ops.torchao.dequantize_affine(
        input=input,
        block_size=input.shape,
        scale=scale,
        zero_point=torch.tensor(0),
        input_dtype=torch.float8_e4m3fn,
    )
    if output_dtype != torch.float:
        res = res.to(output_dtype)
    return res

def quantize_per_tensor(
        input: torch.Tensor,
        scale: torch.Tensor,
) -> torch.Tensor:
    return torch.ops.torchao.quantize_affine(
        input=input,
        block_size=input.shape,
        scale=scale,
        zero_point=torch.tensor(0),
        output_dtype=torch.float8_e4m3fn,
    )

class Perceptron(torch.nn.Module):
    def __init__(
        self,
        in_size: int,
        out_size: int,
        bias: bool = True,
        activation: Union[
            torch.nn.Module,
            Callable[[torch.Tensor], torch.Tensor],
        ] = torch.relu,
        device: Optional[torch.device] = None,
        dtype: torch.dtype = torch.float32,
    ) -> None:
        super().__init__()
        self._out_size = out_size
        self._in_size = in_size
        self._linear: nn.Linear = nn.Linear(
            self._in_size,
            self._out_size,
            bias=bias,
            device=device,
            dtype=dtype,
        )
[...]

PR #2299, opened 2025-06-04
skip quant/dequant decomposed

shiyang-weng:
Fix https://github.com/pytorch/ao/issues/2228

> What we want to do now is to enable FP8 quantization in PyTorch. And similar as INT8 quantization, we need to insert quantize and dequantize ops into the graph.
> However we met problems with these q/dq ops both in the PyTorch core and Torchao.
> 
> PyTorch core:
> 
> The quantize_per_tensor op does not support FP8. We want to fix it via https://github.com/pytorch/pytorch/pull/153601. And as you commented, the op is deprecated.
> Torchao:
> 
> In the fusion pass in Inductor, we want to match the pattern fp8_weight -> torchao.dequantize_affine_float8 -> fp32_op and fuse it as fp8_weight -> weight_pack -> fp8_op. We have done so for INT8 PT2E quantization. However, the pattern matching pass is applied after a constant folding pass in Inductor:
> https://github.com/pytorch/pytorch/blob/100ec0b34aeff2b948dae33937857d0c86cf1646/torch/_inductor/fx_passes/freezing_patterns.py#L69C1-L74C1
> After constant_fold(gm), the pattern will be folded as fp32_weight -> fp32_op. Then the original pattern cannot be found any more and the FP8 semantics is lost since the pattern is entirely in fp32 now.
> For INT8, the int8_weight -> quantized_decomposed.dequantize_per_channel -> fp32_op pattern won't be folded because we mark quantized_decomposed.dequantize_per_channel impure so that it won't be folded: https://github.com/pytorch/pytorch/blob/100ec0b34aeff2b948dae33937857d0c86cf1646/torch/_inductor/constant_folding.py#L139C1-L149C1 . But for the torchao.dequantize_affine_float8, we cannot do this because
> It is an op from Torchao, which is unknown to the constant folder
> It is decomposed to smaller ops, so we cannot put it in the list as a single op.

1. Aligned dispatch_key with [pt](https://github.com/pytorch/pytorch/blob/6fb62931593fc10252e4994cd1a0595ebf8e990f/torch/ao/quantization/fx/_decomposed.py#L49). quantize_affine and dequantize_affine will not be decomposed
2. Register meta func for q/dq
3. Register dq for uintx. Previously, dequant would be decomposed into small ops without the need for registration.


torchao/quantization/quant_primitives.py
```diff
@@ -377,6 +378,23 @@ def _quantize_affine(
...
+    output_dtype: torch.dtype,
+    quant_min: Optional[Union[int, float]] = None,
+    quant_max: Optional[Union[int, float]] = None,
+    zero_point_domain: ZeroPointDomain = ZeroPointDomain.INT,
+) -> torch.Tensor:
+    return torch.empty(input.shape, dtype=output_dtype, device=input.device)
+
+
+_quantize_affine, _quantize_affine_meta = register_custom_op_with_meta(
+    _quantize_affine, _quantize_affine_meta)
```
Xia-Weiwen: Can we separate the registrations for meta and non-meta device and also use them as decorators instead calling it? Then it would be better aligned with the rest of the file.

LevelDownRefine: > Can we separate the registrations for meta and non-meta device and also use them as decorators instead calling it? Then it would be better aligned with the rest of the file.

Done

Xia-Weiwen (approved):

torchao/utils.py
```diff
@@ -221,11 +221,12 @@ def decorator(fn):
...
             schema = op_name + infer_schema(fn, mutates_args={})
             lib.define(schema)
-            lib.impl(op_name, fn, "CompositeImplicitAutograd")
+            lib.impl(op_name, fn, dispatch_key)
 
             lib_namespace = lib.ns
             op = getattr(getattr(torch.ops, lib_namespace), op_name)
-            register_decomposition([op])(fn)
+            if dispatch_key == "CompositeImplicitAutograd":
+                register_decomposition([op])(fn)
```
leslie-fang-intel, line 229: Feeling adding a flag in `_register_custom_op` to indicate whether this op should be decomposed will be clearer, cc @jerryzh168 what's your suggestions?

torchao/dtypes/uintx/uintx_layout.py
```diff
@@ -208,6 +209,16 @@ def _(func, types, args, kwargs):
     )
 
 
+@implements(torch.ops.torchao.dequantize_affine)
+def _(func, types, args, kwargs):
+    return return_and_correct_aliasing(
+        func.default,  # work around
```
LevelDownRefine, line 215: Better way may be
``` python
@implements(torch.ops.torchao.dequantize_affine.default)
    return return_and_correct_aliasing(
        func, 
        args,
        kwargs,
        _dequantize_affine_impl(*args),
    )
```

But there will be following issue. NotImplementedError: UintxTensor dispatch: attempting to run unimplemented operator/function: func=<OpOverload(op='aten.to', overload='dtype')>, types=(<class 'torchao.dtypes.uintx.uintx_layout.UintxTensor'>,)

LevelDownRefine:
Remove this pr still has this accuracy error.
So this error may not be caused by this patch
```shell
$ conda list |grep torch
pytorch-triton            3.3.1+gitc8757738          pypi_0    pypi
torch                     2.8.0.dev20250608+cu126          pypi_0    pypi
torchao                   0.12.0+git3aa9361           dev_0    <develop>
torchaudio                2.8.0.dev20250609+cu126          pypi_0    pypi
torchvision               0.23.0.dev20250609+cu126          pypi_0    pypi
$ git log -1
commit 3aa93619466739c9d9845e1db3bfb2ff0f464857 (HEAD, origin/main, origin/HEAD, main)
$ rm -rf /tmp/torchinductor_pt-gpu/ && /data/wengshiy/conda/envs/ao/bin/pytest -s test/integration/test_integration.py::TestSubclass
       if q.is_Number and p.is_Number:
>           assert p >= 0, p
E           torch._inductor.exc.InductorError: AssertionError: -420769046757011/1000000000000000
E           
../conda/envs/ao/lib/python3.10/site-packages/torch/utils/_sympy/functions.py:488: InductorError
```

leslie-fang-intel (approved):

LevelDownRefine:
Could you help reivew this pr? @jerryzh168 
Looks CI errors are unrelated.

And I need to fill out CLA, do you know who I can fill out on the "point of contact"?

test/quantization/test_quant_primitives.py
```diff
@@ -829,6 +833,84 @@ def test_fake_quantize_affine_cachemask(self):
...
+        with torch.no_grad():
+            # reference implementation using generic primitives
+            expected_scale = torch.tensor(2.0)
+            expected_quantized = quantize_affine(
+                input,
+                input.shape,
+                expected_scale,
+                output_dtype=float8_dtype,
+                quant_min=torch.finfo(float8_dtype).min,
+                quant_max=torch.finfo(float8_dtype).max,
```
jerryzh168, line 877: I remember now we have [quantize_affine_float8](https://github.com/pytorch/ao/blob/0afa4c1bd28c82921e360ddbd1b27c9d6da5b947/torchao/quantization/quant_primitives.py#L2078)?

torchao/dtypes/uintx/uintx_layout.py
jerryzh168: are you using uintx layout for float8? why is this changed?

torchao/utils.py
jerryzh168: yeah I feel adding a flag might be easier here

jerryzh168:
for CLA I think you can just follow the process in the CI:

```
In order for us to review and merge your suggested changes, please sign at https://code.facebook.com/cla. If you are contributing on behalf of someone else (eg your employer), the individual CLA may not be sufficient and your employer may need to sign the corporate CLA.

Once the CLA is signed, our tooling will perform checks and validations. Afterwards, the pull request will be tagged with CLA signed. The tagging process may take up to 1 hour after signing. Please give it that time before contacting us about it.

If you have received this in error or have any questions, please contact us at [cla@meta.com](mailto:cla@meta.com?subject=CLA%20for%20pytorch%2Fao%20%232299). Thanks!
```

torchao/dtypes/uintx/uintx_layout.py
LevelDownRefine: > are you using uintx layout for float8? why is this changed?

dequantize_affine/quantize_affine also used in uintx. 
Without this patch, dequant would be decomposed, so it's ok for uintx, but now it is not decomposed. It is necessary to be found on [_dispatch__torch_function__](https://github.com/pytorch/ao/blob/0eedbf3f839a1c2ca40f8e7d2cfb795a2073d510/torchao/utils.py#L427).
So I use _implements to add it.

test/quantization/test_quant_primitives.py
LevelDownRefine: Without this patch, dequantize_affine/quantize_affine will be decomposed both on int8 and fp8.
If you think it's better to handle fp8 operators separately, I will create another pr to register fp8 operators separately

jerryzh168: we should use quantize_affine_float8 / dequantize_affine_float8 for float8 I think

torchao/dtypes/uintx/uintx_layout.py
jerryzh168: > But there will be following issue. NotImplementedError: UintxTensor dispatch: attempting to run unimplemented operator/function: func=<OpOverload(op='aten.to', overload='dtype')>, types=(<class 'torchao.dtypes.uintx.uintx_layout.UintxTensor'>,)

for this, you could implement `to` in the UintxTensor as well I think

torchao/utils.py
```diff
@@ -179,7 +179,7 @@ def find_multiple(n: int, *args: int) -> int:
     return n + k - (n % k)
 
 
-def _register_custom_op(lib):
+def _register_custom_op(lib, dispatch_key="CompositeImplicitAutograd"):
```
jerryzh168, line 182: are you planning to change dispatch key to a flag of whether to decompose or not

LevelDownRefine:
Fixed on https://github.com/pytorch/ao/pull/2379

PR #2305, opened 2025-06-04
[sparse] marlin fixes

jcaip:
Summary:

This PR updates sparse-marlin to not use CPU tensors and updates it to be compatible with Int4WeightOnl.

Test Plan:
```
pytest test/sparsity/test_marlin.py
```

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/dtypes/uintx/marlin_sparse_layout.py
+++ b/torchao/dtypes/uintx/marlin_sparse_layout.py
@@ -130,7 +130,7 @@ class MarlinSparseAQTTensorImpl(AQTTensorImpl):
         cls,
         int_data: torch.Tensor,
         scale: torch.Tensor,
-        zero_point: torch.Tensor,
+        zero: torch.Tensor,
         meta: torch.Tensor,
         _layout: Layout,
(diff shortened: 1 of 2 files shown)


torchao/dtypes/uintx/marlin_sparse_layout.py
```diff
@@ -151,16 +151,17 @@ def __init__(
...
         _layout: Layout,
         original_shape: torch.Size,
         group_size: int,
         num_bits: int,
     ):
         self.int_data = int_data
+        self.scale_and_zero = None
         self.scale = scale
-        self.zero_point = zero_point
+        self.zero = zero
```
jerryzh168, line 164: why do we have both scale_and_zero and zero?

jcaip: It's to maintain compatibility with https://github.com/pytorch/ao/blob/main/torchao/dtypes/uintx/int4_xpu_layout.py, otherwise the checks fail because the `TensorImpl` doesn't have this attribute.

PR #2308, opened 2025-06-05
Add slicing support for fbgemm fp8 and int4

jerryzh168:
Summary:
att, this is needed in vllm
Note that irregular shapes will require padding, which is not implemented right now, we can add that if it's required by the model

Test Plan:
```
python test/dtypes/test_fbgemm_fp8.py -k test_slice 
python test/dtypes/test_fbgemm_int4.py -k test_slice
```

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/dtypes/test_fbgemm_fp8.py
+++ b/test/dtypes/test_fbgemm_fp8.py
@@ -25,23 +25,86 @@ from torchao.utils import (
 
 
 @unittest.skipIf(not TORCH_VERSION_AT_LEAST_2_8, "Need pytorch 2.8+")
+@unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
+@unittest.skipIf(not is_sm_at_least_90(), "Nedd sm90+")
 class TestFbgemmFp8Tensor(TestCase):
-    @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
-    @unittest.skipIf(not is_sm_at_least_90(), "Nedd sm90+")
+    def setUp(self):
+        self.config = FbgemmConfig(
+            input_dtype=e4m3_dtype,
+            weight_dtype=e4m3_dtype,
+            output_dtype=torch.bfloat16,
+        )
+
     def test_linear(self):
         dtype = torch.bfloat16
         device = "cuda"
         input = torch.randn(1, 128, dtype=dtype, device=device)
         linear = torch.nn.Linear(128, 256, dtype=dtype, device=device)
         original = linear(input)
-        config = FbgemmConfig(
-            input_dtype=e4m3_dtype,
-            weight_dtype=e4m3_dtype,
-            output_dtype=torch.bfloat16,
-        )
-        quantize_(linear, config)
+        quantize_(linear, self.config)
         quantized = linear(input)
         self.assertTrue(compute_error(original, quantized) > 20)
+
+    def test_slice(self):
+        dtype = torch.bfloat16
+        device = "cuda"
+        dummy = torch.nn.Linear(256, 256, bias=False, dtype=dtype, device=device)
+        dummy1 = torch.nn.Linear(256, 64, bias=False, dtype=dtype, device=device)
+        dummy1.weight = torch.nn.Parameter(
+            dummy.weight.narrow(0, 0, 64), requires_grad=False
+        )
+        dummy2 = torch.nn.Linear(128, 256, dtype=dtype, device=device)
+        dummy2.weight = torch.nn.Parameter(
+            dummy.weight.narrow(1, 0, 128), requires_grad=False
+        )
+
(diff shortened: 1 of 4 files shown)


test/dtypes/test_fbgemm_fp8.py
```diff
@@ -34,15 +41,43 @@ def test_linear(self):
...
+        weight1 = dummy.weight.narrow(0, 0, 64)
+        weight2 = dummy.weight.narrow(1, 0, 128)
+
+        # check for sliced weight, before and after float8 quantization
+        # does not differ too much
+        input = torch.randn(2, 256, dtype=dtype, device=device)
+        res_ref = dummy1(input)
+        dummy.weight = torch.nn.Parameter(weight1, requires_grad=False)
+        res = dummy(input)
+        assert compute_error(res, res_ref) > 25
```
drisspg, line 77: can this be made more concrete, e.g. we can test the exact elements match

test/dtypes/test_fbgemm_int4.py
```diff
@@ -33,16 +41,43 @@ def test_linear(self):
...
+        dummy1.weight = torch.nn.Parameter(
+            dummy.weight.narrow(0, 0, 64), requires_grad=False
+        )
+        dummy2 = torch.nn.Linear(128, 256, dtype=dtype, device=device)
+        dummy2.weight = torch.nn.Parameter(
+            dummy.weight.narrow(1, 0, 128), requires_grad=False
+        )
+
+        quantize_(dummy, self.config)
+        weight1 = dummy.weight.narrow(0, 0, 64)
```
drisspg, line 62: ditto here

torchao/dtypes/fbgemm_fp8_tensor.py
```diff
@@ -146,6 +155,43 @@ def _(func, types, args, kwargs):
     )
 
 
+@implements(aten.slice.Tensor)
+def _(func, types, args, kwargs):
+    self, dim, start, end, step = fill_defaults(args, 5, [0, None, None, 1])
```
drisspg, line 198: I think this stuff is error prone, do you think that : https://github.com/pytorch/ao/blob/488ecd4cf871035b091981a557a9204a5d50bf5b/torchao/dtypes/floatx/float8_layout.py#L302

is helpful here?

test/dtypes/test_fbgemm_int4.py
jerryzh168: we can have exact match if dequantize op is implemented I think, right now it's just a sanity check, can we follow up this later?

torchao/dtypes/fbgemm_fp8_tensor.py
jerryzh168: the scale here is not 2d actually, but might be useful if we resize that to 2D/3D

drisspg: I dont think this relies on 2d vs 3d no?

test/dtypes/test_fbgemm_int4.py
drisspg: Can you ensure that the weight1 's data + its scale is exactly the right slice of dummy.weight

jerryzh168: yeah I can add an assert for that to check the internal data

torchao/dtypes/fbgemm_fp8_tensor.py
jerryzh168: does this assume scale has the same shape as data: https://github.com/pytorch/ao/blob/488ecd4cf871035b091981a557a9204a5d50bf5b/torchao/dtypes/floatx/float8_layout.py#L321

drisspg: hmm yeah it does, we should probably assert

test/dtypes/test_fbgemm_fp8.py
jerryzh168: updated above, please take a look again

torchao/dtypes/fbgemm_int4_tensor.py
```diff
@@ -151,13 +153,96 @@ def _(func, types, args, kwargs):
...
+        for tensor_name in self_tensors:
+            getattr(self, tensor_name).copy_(getattr(src, tensor_name))
+        return
+    raise ValueError(
+        f"Not supported args for copy_ due to metadata mismatch: {args[0], args[1]}"
+    )
+
+
+@implements(aten.slice.Tensor)
+def _(func, types, args, kwargs):
```
drisspg, line 190: Maybe a doc block here on what is supported and the semantics this code is not that easy to grok

drisspg (approved):

PR #2313, opened 2025-06-05
[BE/docs] Add float8 training api ref to docsite

danielvegamyhre:
Test plan:
- Ran local preview of docsite and confirmed API ref renders properly. I will move ScalingGranularity to be available at the module level in a subsequent PR, so that it will appear in the API ref too (it's needed for user-facing API to convert model to use float8 rowwise training)

### DIFF
--- a/docs/source/api_ref_float8.rst
+++ b/docs/source/api_ref_float8.rst
@@ -1,0 +1,30 @@
+.. _api_quantization:
+
+====================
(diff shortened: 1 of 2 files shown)


jerryzh168 (approved):

danielvegamyhre:
fyi @vkuzo for awareness

PR #2321, opened 2025-06-05
[BE] Make internal torchao.float8 functions private

danielvegamyhre:


danielvegamyhre:
> anything used by other people?

Yes, I did an internal search for each function and added aliases to the old/public name for those that had users.

PR #2323, opened 2025-06-06
Add round_scales_to_power_of_2 option for float quantization

drisspg:
Stacked PRs:
 * #2398
 * __->__#2323

--- --- ---

Add round_scales_to_power_of_2 option for float quantization

This adds support for rounding scaling factors down to the nearest power of 2
for float quantization, following the pattern established in Float8LinearConfig.

Key changes:
- Add round_scales_to_power_of_2 parameter to all float quantization configs
- Update choose_qparams_affine_floatx and to_scaled_tc_floatx functions to apply power of 2 rounding
- Thread the parameter through all relevant function calls in quant_api.py

Lets users who train with this setting run in inference

### DIFF
--- a/test/dtypes/test_affine_quantized_float.py
+++ b/test/dtypes/test_affine_quantized_float.py
@@ -28,7 +28,10 @@ from torch.testing._internal import common_utils
 from torchao.dtypes.floatx.float8_layout import Float8AQTTensorImpl, preprocess_scale
 from torchao.float8.float8_utils import compute_error
 from torchao.quantization import (
+    Float8DynamicActivationFloat8SemiSparseWeightConfig,
     Float8DynamicActivationFloat8WeightConfig,
+    Float8StaticActivationFloat8WeightConfig,
+    Float8WeightOnlyConfig,
     float8_dynamic_activation_float8_weight,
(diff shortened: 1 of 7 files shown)


test/dtypes/test_affine_quantized_float.py
```diff
@@ -630,6 +633,100 @@ def test_float8_tensor_slicing_functional_correctness(self, granularity):
...
+        if min_sm == 90 and not is_sm_at_least_90():
+            self.skipTest("Requires GPU with compute capability >= 9.0")
+
+        device = "cuda"
+        dtype = torch.bfloat16
+        model = torch.nn.Linear(64, 32, bias=False).to(device).to(dtype)
+
+        config = config_factory()
+        if isinstance(
+            config, Float8DynamicActivationFloat8SemiSparseWeightConfig
```
danielvegamyhre, line 757: Shouldn't the `min_sm` check above already handle this? Or is this because the cutlass kernel is only built on sm90 / 90a?

Test case for Float8SemiSparse is failing on h100 due to CUDA backend not supporting operator (kernel not built?) so just wondering

PR #2325, opened 2025-06-06
moe quant with dedicated kernels [wip]

HDCharles:
Summary:

extending the torchao moe support to have more performant kernels. This PR supports both scaled_grouped_mm and fbgemm's grouped_gemm_fp8_rowwise though it seems like grouped_gemm_fp8_rowwise is a bit buggy (need to make a clear repro)

todo: run benchmarks, debug fbgemm kernel, unit tests

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/prototype/moe_quant/kernels.py
+++ b/torchao/prototype/moe_quant/kernels.py
@@ -1,0 +1,155 @@
+import torch
+import torch.nn.functional as F
+from torchao.quantization.utils import _torchtitan_available, _fbgemm_available
+
+grouped_gemm_fp8_rowwise = None
+if _fbgemm_available:
+    try:
+        from fbgemm_gpu.experimental.gemm.triton_gemm.grouped_gemm import grouped_gemm_fp8_rowwise
+    except:
+        pass    
+
+__all__ = ["fp8_dq_moe_op",
+           "manual_pad",
+           "torchtitan_pad",
+        ]
+
+
+def fp8_dq_moe_op(input, w1, w2, w3, expert_indices, scores, fast_accum=True, use_fbgemm_kernel=False):
+    # parameters
+    orig_in_shape = input.shape
+    input.reshape(-1, orig_in_shape[-1])
+    num_tokens, dim = input.shape
+    num_experts, expert_dim, _ = w1.shape
+    scores = scores.view(-1, scores.shape[-1])
+    top_k = scores.shape[-1]
+    total_activations = num_tokens*top_k
+    
+    # preprocess indices
+    expert_indices = expert_indices.view(-1)
+    activation_shuffle = expert_indices.argsort(stable=True)
+    token_shuffle = activation_shuffle.div(top_k).floor().to(torch.int64)
+    num_tokens_per_expert = torch.histc(expert_indices, bins=num_experts, min=0, max=num_experts)
+    
+    # padding
+    alignment = 16
+    if _torchtitan_available:
+        num_ranks = 1
+        padded_indices, m_offsets = torchtitan_pad(num_tokens_per_expert, alignment, num_ranks)
+    else:
+        padded_indices, m_offsets = manual_pad(num_tokens_per_expert, alignment)
+
+    pad_len = padded_indices.shape[0]
+    valid_values = padded_indices >= 0
+    
+    # get data for weights
+    w1_fp8 = w1.original_weight_tensor.tensor_impl.float8_data
+    w1_scale = w1.original_weight_tensor.tensor_impl.scale.squeeze()
+    w1_qfunc = w1.input_quant_func
+    w1_quant_kwargs = w1.quant_kwargs
+
+    w3_fp8 = w3.original_weight_tensor.tensor_impl.float8_data
+    w3_scale = w3.original_weight_tensor.tensor_impl.scale.squeeze()
+
+    w2_fp8 = w2.original_weight_tensor.tensor_impl.float8_data
+    w2_scale = w2.original_weight_tensor.tensor_impl.scale.squeeze()
+    w2_qfunc = w2.input_quant_func
+    w2_quant_kwargs = w2.quant_kwargs
+
+
+    # quantize then shuffle input
+    q_input = w1_qfunc(input, **w1_quant_kwargs)
+    q_input_data = q_input.tensor_impl.float8_data
+    q_input_scale = q_input.tensor_impl.scale.squeeze()
+    input_fp8 = torch.zeros((pad_len, q_input_data.shape[-1]), dtype=q_input_data.dtype, device=q_input_data.device)
+    input_scale = torch.zeros(pad_len, dtype=q_input_scale.dtype, device=q_input_scale.device)
+    input_fp8[valid_values] = q_input_data[token_shuffle]
+    input_scale[valid_values] = q_input_scale[token_shuffle] if q_input_scale.numel()>1 else q_input_scale
+    
+    if use_fbgemm_kernel:
+        assert grouped_gemm_fp8_rowwise is not None, "fbgemm kernel requires fbgemm-gpu-genai to be installed: https://github.com/pytorch/FBGEMM/blob/main/fbgemm_gpu/experimental/gen_ai/README.md"
+        y1 = grouped_gemm_fp8_rowwise(input_fp8, w1_fp8.reshape(-1, w1_fp8.shape[-1]), m_offsets, input_scale, w1_scale.reshape(-1), use_fast_accum=True)
+        y3 = grouped_gemm_fp8_rowwise(input_fp8, w3_fp8.reshape(-1, w3_fp8.shape[-1]), m_offsets, input_scale, w3_scale.reshape(-1), use_fast_accum=True)
+        
+        y = F.silu(y1)*y3
+        
+        y_q = w2_qfunc(y, **w2_quant_kwargs)
+
+        y_fp8 = y_q.tensor_impl.float8_data
+        y_scale = y_q.tensor_impl.scale.squeeze()
+        out = grouped_gemm_fp8_rowwise(y_fp8, w2_fp8.view(-1, w1_fp8.shape[-1]), m_offsets, y_scale, w2_scale.view(-1), use_fast_accum=fast_accum)
+        # unpad and combine output with weights
+        out = out[valid_values]
+        sorted_scores = scores.reshape(-1,1)[activation_shuffle]
+        out = out*sorted_scores
+
+        # sum weighted outputs
+        final_out = torch.zeros_like(input)
+        final_out = final_out.scatter_add(
+            dim=0,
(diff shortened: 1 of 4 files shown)


torchao/prototype/moe_quant/quantizable_moe_modules.py
```diff
@@ -105,11 +108,20 @@ def forward(
...
                 .reshape(x.shape)
             )
             return final_out
+        
+        # fp8 dq moe
+        elif (
+            isinstance(self.w1, torchao.quantization.linear_activation_quantized_tensor.LinearActivationQuantizedTensor) and 
+            isinstance(self.w1.original_weight_tensor._layout, torchao.dtypes.floatx.float8_layout.Float8Layout)
+        ):
+            final_out = fp8_dq_moe_op(x, self.w1, self.w2, self.w3, expert_indices, scores)
```
jerryzh168: is it possible to call this op without modifying the source model? 

is there a gropup_mm for bfloat16 that we can overwrite and dispatch to scaled_grouped_mmm?

alexsamardzic: Yes, there is `_grouped_mm` in PyTorch core that does that.

alexsamardzic:
PR to hopefully remove need for padding groups is here: pytorch/pytorch#155466.

torchao/prototype/moe_quant/kernels.py
```diff
@@ -0,0 +1,155 @@
...
+    expert_indices = expert_indices.view(-1)
+    activation_shuffle = expert_indices.argsort(stable=True)
+    token_shuffle = activation_shuffle.div(top_k).floor().to(torch.int64)
+    num_tokens_per_expert = torch.histc(expert_indices, bins=num_experts, min=0, max=num_experts)
+    
+    # padding
+    alignment = 16
+    if _torchtitan_available:
+        num_ranks = 1
+        padded_indices, m_offsets = torchtitan_pad(num_tokens_per_expert, alignment, num_ranks)
```
ngimel: heads up, soon we won't need padding once #155466 lands

torchao/prototype/moe_quant/kernels.py
```diff
@@ -0,0 +1,155 @@
...
+    # quantize then shuffle input
+    q_input = w1_qfunc(input, **w1_quant_kwargs)
+    q_input_data = q_input.tensor_impl.float8_data
+    q_input_scale = q_input.tensor_impl.scale.squeeze()
+    input_fp8 = torch.zeros((pad_len, q_input_data.shape[-1]), dtype=q_input_data.dtype, device=q_input_data.device)
+    input_scale = torch.zeros(pad_len, dtype=q_input_scale.dtype, device=q_input_scale.device)
+    input_fp8[valid_values] = q_input_data[token_shuffle]
+    input_scale[valid_values] = q_input_scale[token_shuffle] if q_input_scale.numel()>1 else q_input_scale
+    
+    if use_fbgemm_kernel:
```
ngimel, line 55: we have fbgemm-like kernels available via autotuning in torch.compile, thanks to #155138, do you think we still need separate fbgemm path?

torchao/prototype/moe_quant/quantizable_moe_modules.py
HDCharles: that's a better integration point but i'm not sure i'll be able to complete that before i have to head out on leave.

also i'd probably make that a separate PR instead of combining everything into one since that would be a significant change to the base moe integration.

alexsamardzic:
PR pytorch/pytorch#155466, that makes it possible to avoid padding, is merged.  Here is a quick patch to remove padding (note that it also disables FBGEMM altogether, so `_scaled_grouped_mm` implementation from PyTorch core is used): 

[096_fuse_moeb-diff.txt](https://github.com/user-attachments/files/20804025/096_fuse_moeb-diff.txt)

I'm not completely sure that scale tensor shape adjustment I made here are correct, but in any case this patch, together with latest PyTorch used, will make all the tests in `test_moe_quant.py` pass.

ngimel:
@alexsamardzic we still would need padding for backward where K could possibly become 0?

alexsamardzic:
> @alexsamardzic we still would need padding for backward where K could possibly become 0?

This PR is not concerned about backward, but I would say @danielvegamyhre is touching on it: #2405.  In any case, you have a point, here is a diff for `_grouped_mm` tests in PyTorch to demonstrate the issue:

<details>
<summary>diff</summary>

```diff
diff --git a/test/test_matmul_cuda.py b/test/test_matmul_cuda.py
index 4e64c807425..96667a79440 100644
--- a/test/test_matmul_cuda.py
+++ b/test/test_matmul_cuda.py
@@ -354,15 +354,15 @@ class TestMatmulCuda(TestCase):
     @unittest.skipIf(TEST_WITH_ROCM, "ROCm doesn't support CUTLASS")
     @xfailIfSM100OrLater
     @unittest.skipIf(not SM90OrLater, "Grouped gemm supported on SM90")
-    @parametrize("strided", [False, True])
-    @parametrize("a_row_major", [False, True])
-    @parametrize("b_row_major", [False, True])
-    @parametrize("use_torch_compile", [False, True])
+    @parametrize("strided", [True])
+    @parametrize("a_row_major", [True])
+    @parametrize("b_row_major", [True])
+    @parametrize("use_torch_compile", [True, False])
     def test_grouped_gemm_2d_3d(self, strided, a_row_major, b_row_major, use_torch_compile):
         device = "cuda"
         dtype = torch.bfloat16
         s_int = int(strided)
-        m, n, k, n_groups = 16, 32, 64, 4
+        m, n, k, n_groups = 3, 32, 64, 5
         if a_row_major:
             a = torch.randn(m * n_groups, k * (1 + s_int), device=device, dtype=dtype)[:, :k]
         else:
@@ -388,6 +388,7 @@ class TestMatmulCuda(TestCase):
             a.grad = None
             b.grad = None
             offs = torch.arange(m, n_groups * m + 1, m, device="cuda", dtype=torch.int32)
+            offs = torch.tensor([0, 1, 6, 6, 15], device="cuda", dtype=torch.int32)
             if check_zero_size:
                 offs[0] = offs[1]
````
</details>

If offsets changed say to `[1, 3, 5, 6, 15]`, so that there are no zero sizes, it would work fine.   I'm going to see is it possible to further refine [these checks](https://github.com/pytorch/pytorch/blob/4da98351b9e231bbbe83b5c590cf7d2ea382333a/aten/src/ATen/native/cuda/GroupMMCommon.cuh#L50-L54), in order to make it work for this case.

Edit: see [here](https://github.com/pytorch/pytorch/pull/155466#issuecomment-2991933436).

**Comment by alexsamardzic** [comment]
Here is slightly changed diff:  [096_fuse_moeb-diff.txt](https://github.com/user-attachments/files/20907002/096_fuse_moeb-diff.txt).  To be applied after PR rebased on latest main.

Some end-to-end performance numbers for Mixtral model, for current version of the PR:

First, this patch is to be applied to enforce auto-tuning for all cases:
```diff
diff --git a/torchao/_models/mixtral-moe/generate.py b/torchao/_models/mixtral-moe/generate.py
index 11a53043..10da20f7 100644
--- a/torchao/_models/mixtral-moe/generate.py
+++ b/torchao/_models/mixtral-moe/generate.py
@@ -337,10 +337,10 @@ def main(
 
         if batch_size == 1 and (isinstance(moe_quant, str) and "base" in moe_quant):
             decode_one_token = torch.compile(
-                decode_one_token, mode="reduce-overhead", fullgraph=True
+                decode_one_token, mode="max-autotune"
             )
         else:
-            decode_one_token = torch.compile(decode_one_token, mode="reduce-overhead")
+            decode_one_token = torch.compile(decode_one_token, mode="max-autotune")
 
         if args.compile_prefill:
             prefill = torch.compile(prefill, fullgraph=True, dynamic=True)
```

For each run below, `--checkpoint_path=.../checkpoints/mistralai/Mixtral-8x7B-Instruct-v0.1/model.pth` should be added to command line, skipped for brevity.

First, a variant that goes through [this branch](https://github.com/pytorch/ao/blob/186708f41679f44cc239776b05c54bc7e1da4d70/torchao/_models/mixtral-moe/model.py#L361-L387) of the model forward function:

```sh
$ python generate.py --compile --moe_quant fp8dq

Average tokens/sec: 10.23
Memory used: 51.21 GB
model size: 48.37
```

Then a variant that goes through [this branch](https://github.com/pytorch/ao/blob/186708f41679f44cc239776b05c54bc7e1da4d70/torchao/_models/mixtral-moe/model.py#L328-L353):

```sh
$ python generate.py --compile --moe_quant fp8dq-base

Average tokens/sec: 56.34
Memory used: 59.14 GB
model size: 48.37
```

Finally, for a variant that will utilize auto-tuned `_scaled_grouped_mm` i.e. go through [this branch](https://github.com/pytorch/ao/blob/186708f41679f44cc239776b05c54bc7e1da4d70/torchao/_models/mixtral-moe/model.py#L354-L360) of the model, one could skip the first branch by adding `if False and...` [here](https://github.com/pytorch/ao/blob/186708f41679f44cc239776b05c54bc7e1da4d70/torchao/_models/mixtral-moe/model.py#L328), and then:

```sh
$ python generate.py --compile --moe_quant fp8dq-base

Average tokens/sec: 101.24
Memory used: 59.14 GB
model size: 48.37
```

If auto-tuning for `_scaled_grouped_mm` disabled (e.g. by simply removing the meta registration), the tokens/sec is practically the same, and that is kind of expected as the batches are of very small size.

Again, this all could be further improved, leaving it at that for now.

ngimel:
@alexsamardzic so pytorch now with removed paddiing restrictions is strictly better than fbgemm?

alexsamardzic:
> @alexsamardzic so pytorch now with removed paddiing restrictions is strictly better than fbgemm?

The FBGEMM kernel is not included in the results above.  To have it activated, on top of changes mentioned for the last run (that was about using `_scaled_grouped_mm`), the `False` [here](https://github.com/pytorch/ao/blob/186708f41679f44cc239776b05c54bc7e1da4d70/torchao/_models/mixtral-moe/model.py#L39)  should be changed to `True`.  After these changes:

```sh
$ python generate.py --compile --moe_quant fp8dq-base

Average tokens/sec: 23.51
Memory used: 59.14 GB
model size: 48.37
```
Plus, the output is garbage.  However, regarding the performance, note [this](https://github.com/pytorch/ao/blob/186708f41679f44cc239776b05c54bc7e1da4d70/torchao/prototype/moe_quant/kernels.py#L60), i.e. the compilation is at the moment disabled around calls to FBGEMM kernel; and if enabled it would error out.  I'm not sure @HDCharles would be interested in working further on that branch, but IMO both FBGEMM and PyTorch Triton kernels should have similar performance, so FBGEMM kernel usage may be safely skipped.

torchao/prototype/moe_quant/kernels.py
vkuzo: does that require max-autotune?

PR #2337, opened 2025-06-08
Add support for bmm and `to` for fbgemm Tensor

jerryzh168:
Summary:
att, this PR adds support for running quantized bmm, the quantized bmm kernel for int4 and fp8 (with dynamic activation quantization) requires transpose of weights in order to run, so added transpose_input to the convert function to transpose the weights first

Test Plan:
python test/dtypes/test_fbgemm_fp8.py -k test_bmm
python test/dtypes/test_fbgemm_int4.py -k test_bmm

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/dtypes/fbgemm_fp8_tensor.py
+++ b/torchao/dtypes/fbgemm_fp8_tensor.py
@@ -18,6 +18,7 @@ from torchao.utils import (
 
 __all__ = [
     "to_fbgemm_fp8",
+    "FbgemmFp8Tensor",
 ]
 
 aten = torch.ops.aten
@@ -74,11 +75,22 @@ class FbgemmFp8Tensor(TorchAOBaseTensor):
     def _quantization_type(self):
         return f"shape={self.shape}, activation_scale_ub={self.activation_scale_ub}, device={self.device}"
 
+    def to(self, *args, **kwargs):
+        kwargs = self._get_to_kwargs(*args, **kwargs)
+        device = kwargs.pop("device")
+        return self.__class__(
+            self.float8_data.to(device),
+            self.scale.to(device),
+            self.activation_scale_ub.to(device),
+            self.dtype,
+        )
+
     @classmethod
     def from_float(
         cls,
         w: torch.Tensor,
         activation_scale_ub: Optional[float] = None,
+        transpose_input: bool = False,
     ):
         if activation_scale_ub is None:
             activation_scale_ub = 1200.0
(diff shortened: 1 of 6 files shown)


torchao/dtypes/fbgemm_fp8_tensor.py
```diff
@@ -88,6 +89,12 @@ def from_float(
             dtype=torch.float,
             device=w.device,
         )
+        if transpose_input:
+            if w.ndim == 3:
```
drisspg, line 104: nit: maybe just w.transpose(-1, -2)

torchao/dtypes/fbgemm_fp8_tensor.py
```diff
@@ -141,6 +148,39 @@ def _(func, types, args, kwargs):
...
+    if not input_tensor.is_floating_point():
+        raise NotImplementedError(
+            f"{func} is not implemented for non floating point input"
+        )
+
+    orig_act_size = input_tensor.size()
+
+    # not used
+    num_tokens = torch.empty([input_tensor.size(0)], device=input_tensor.device)
+    xq, x_scale = torch.ops.fbgemm.quantize_fp8_per_row(
```
drisspg, line 166: This ot use num_tokens feels weird, maybe make an issue  on fbgemm? or update the op to not need

torchao/dtypes/fbgemm_int4_tensor.py
```diff
@@ -82,13 +82,20 @@ def from_float(
...
         block_size: List[int],
+        transpose_input: bool = False,
     ):
         assert len(block_size) == w.ndim, (
             f"Expecting the length of block_size to be equal to the dimension of the weight, got {block_size=} and {w.ndim=}"
         )
         if int4_row_quantize_zp is None:
             raise ImportError("Requires fbgemm-gpu-genai >= 1.2.0")
 
+        if transpose_input:
```
drisspg, line 105: ditto here

torchao/dtypes/fbgemm_int4_tensor.py
```diff
@@ -146,6 +153,30 @@ def _(func, types, args, kwargs):
     return res
 
 
+@implements(torch.bmm)
+def _(func, types, args, kwargs):
+    input_tensor, weight_tensor = (
+        args[0],
+        args[1],
+    )
+    if not input_tensor.is_floating_point():
```
drisspg: nit: Is this guard needed?  Like is this a common situation to run into

jerryzh168: this may not be needed, just copy pasted from other impls

torchao/dtypes/fbgemm_fp8_tensor.py
jerryzh168: yeah I checked with @jiawenliu64 and this arg is indeed only used in internal use cases, he was recommending to use the triton op, although I found the triton op is a bit slower, maybe it requires some tuning. I'll double check

drisspg (approved):

PR #2342, opened 2025-06-09
Implemented a new test case for LUT quantization

szyszyzys:


torchao/experimental/kernels/cpu/aarch64/tests/test_utils_lut.h
```diff
@@ -0,0 +1,321 @@
...
+  GroundTruthStrategy strategy;
+  int group_size = -1;
+  int nbit = 4;
+  int random_seed = 123;
+};
+
+
+
+std::pair<std::vector<int8_t>, std::vector<int8_t>>
+generate_simple_u_to_s_lut_and_indices(
```
metascroy: What does "simple_u_to_s" mean?

szyszyzys: The term "simple_u_to_s" refers to converting an unsigned index to a signed index. The example provided is a simplified scenario where we only need to map values to each slot. I will work on making the function name more descriptive and readable.

torchao/experimental/kernels/cpu/aarch64/tests/test_utils_lut.h
```diff
@@ -0,0 +1,321 @@
...
+
+
+  /**
+  * @brief Generates a Look-Up Table (LUT) from pre-computed quantization parameters.
+  *
+  * This function creates a dequantization LUT that maps every possible value of the
+  * input data type (T_in) to its corresponding floating-point representation for
+  * each quantization group.
+  *
+  * @tparam T_in The data type of the quantized values (e.g., int8_t). This
```
metascroy, line 63: Where is the T_in defined in the code?

torchao/experimental/kernels/cpu/aarch64/tests/test_utils_lut.h
```diff
@@ -0,0 +1,321 @@
...
+  *              determines the size of each individual LUT (e.g., 256 entries for int8_t).
+  * @tparam T_zp The data type of the zero-points (e.g., int8_t).
+  * @param scales A vector of scale values, one for each quantization group.
+  * @param zeros A vector of zero-point values, one for each quantization group.
+  * @param has_zeros A flag indicating if the zero-points should be used.
+  * @return A flattened std::vector<float> containing all the group LUTs concatenated.
+  */
+  std::vector<int8_t> generate_requant_lut_from_params(
```
metascroy, line 71: Why are we returning int8_t type, and not float32_t for LUT?

torchao/experimental/kernels/cpu/aarch64/tests/test_utils_lut.h
```diff
@@ -0,0 +1,321 @@
...
+    int nbit,
+    const std::vector<int8_t>& zeros,
+    bool has_zeros) {
+  // Determine the range of the low-bit quantized values.
+  const int q_min = -(1 << (nbit - 1));
+  const int q_max = (1 << (nbit - 1)) - 1;
+  const size_t lut_size_per_group = static_cast<size_t>(q_max) - q_min + 1;
+  const int lut_index_offset = q_min;
+
+  const int num_groups = zeros.size();
```
metascroy, line 81: What is this when has_zeros=false?

torchao/experimental/kernels/cpu/aarch64/tests/test_utils_lut.h
```diff
@@ -0,0 +1,321 @@
...
+  std::vector<int8_t> luts(num_groups * lut_size_per_group);
+
+  for (int group_idx = 0; group_idx < num_groups; ++group_idx) {
+    const int8_t zero_point = has_zeros ? zeros[group_idx] : 0;
+
+    for (int q_val = q_min; q_val <= q_max; ++q_val) {
+      size_t lut_idx = group_idx * lut_size_per_group + (q_val - lut_index_offset);
+      // The LUT stores the result of (quantized_value - zero_point)
+      luts[lut_idx] = static_cast<int8_t>(q_val - zero_point);
```
metascroy, line 90: Why is the lut output dtype int8_t?

torchao/experimental/kernels/cpu/aarch64/tests/test_utils_lut.h
szyszyzys: I updated the implementation but forgot to update the function description. I will provide a new description.

torchao/experimental/kernels/cpu/aarch64/tests/test_utils_lut.h
```diff
@@ -0,0 +1,321 @@
...
+
+      for (int i = 0; i < num_param_sets; ++i) {
+          scales[i] = scale_dis(gen);
+          zeros[i] = static_cast<T_zp>(zero_dis(gen));
+      }
+      return {scales, zeros};
+  }
+
+  // Helper to perform quantization
+  static std::vector<T_in> quantize_input(
```
metascroy, line 292: What is this for?

Can't we just generate the random LUT for testing?

torchao/experimental/kernels/cpu/aarch64/tests/test_utils_lut.h
szyszyzys: In the context of test_channelwise_8bit_activation_groupwise_lowbit_weight_lut, the code requires an int8_t type to function correctly. I initially designed the implementation to serve as a second quantization step, hence the int8_t output. However, I agree that it should be more flexible to accommodate different data types for various scenarios. I attempted to address this in the code but overlooked this function. I will make the necessary adjustments.

torchao/experimental/kernels/cpu/aarch64/tests/test_utils_lut.h
szyszyzys: It could lead to errors. I should not rely on the size of zeros when has_zeros is false.

torchao/experimental/kernels/cpu/aarch64/tests/test_utils_lut.h
szyszyzys: Yes, we can generate a random LUT for testing purposes. My initial design aimed to extend this function to accommodate more scenarios, such as secondary quantization. This would allow it to work in conjunction with other components. My goal was to ensure that it could be integrated and tested as part of a larger system.

torchao/experimental/kernels/cpu/aarch64/tests/test_utils_lut.h
metascroy: Yeah, let's come up with a better name, e.g., maybe "generate_random_int8_lut_and_indices" or something like that

torchao/experimental/kernels/cpu/aarch64/tests/test_utils_lut.h
metascroy: I'm not sure I follow how you're planning on using this.  For LUT quantization, we won't be quantizing in this way.

We also shouldn't re-invent the quantization logic because we have this elsewhere in other tests, e.g., here: https://github.com/pytorch/ao/blob/main/torchao/experimental/kernels/cpu/aarch64/tests/test_utils.h#L110.

If you want to refactor that code to use non-aarch64 code, that'd be great, but I'm not sure it's related to the LUT project.

torchao/experimental/kernels/cpu/aarch64/tests/test_utils_lut.h
```diff
@@ -0,0 +1,321 @@
...
+              size_t lut_idx = i * lut_size_per_set + (q_val_int - lut_idx_offset);
+              float dequant_val = scales[i] * (static_cast<float>(q_val_int) - static_cast<float>(zeros[i]));
+              lut[lut_idx] = static_cast<T_lut>(dequant_val); // Cast to final LUT type
+          }
+        }
+      return lut;
+  }
+
+  // Helper to compute the reference output
+  static std::vector<float> compute_expected_output(
```
metascroy, line 335: Is this doing the matrix product of activations * dequantized_weights?

It looks like it's just dequantizing the weights?

szyszyzys: I may have overdesigned this part. My initial intention was to accommodate a more complex setup, such as integrating with other quantization methods or using different LUT designs. That's why I used the GroundTruthStrategy to set the dequantization strategies. It seems unnecessary in this context.

torchao/experimental/kernels/cpu/aarch64/tests/test_utils_lut.h
szyszyzys: Got it. Thanks!

PR #2344, opened 2025-06-09
[BE] Rename qparams for tinygemm

jainapurva:
Rename 
- quantize_affine_float_zero_point -> quantize_affine_tinygemm
- dequantize_affine_float_zero_point -> dequantize_affine_tinygemm

### DIFF
--- a/torchao/dtypes/affine_quantized_tensor.py
+++ b/torchao/dtypes/affine_quantized_tensor.py
@@ -26,14 +26,14 @@ from torchao.quantization.quant_primitives import (
     choose_qparams_and_quantize_affine_hqq,
     dequantize_affine,
     dequantize_affine_float8,
-    dequantize_affine_float_zero_point,
     dequantize_affine_floatx,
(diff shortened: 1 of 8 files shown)


torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -26,14 +26,14 @@
     choose_qparams_and_quantize_affine_hqq,
     dequantize_affine,
     dequantize_affine_float8,
-    dequantize_affine_float_zero_point,
```
jerryzh168, line 29: can you do a search to see if these are used outside of torchao?

jerryzh168 (approved):

torchao/dtypes/affine_quantized_tensor.py
jerryzh168: oh I did a search, looks like not

jainapurva: Yup, found the same

PR #2345, opened 2025-06-10
Add inplace quantizer examples

cccclai:
Summary:
Add a quantizer example for in place ops

Rollback Plan:

Differential Revision: D76312488

### DIFF
--- a/torchao/quantization/pt2e/constant_fold.py
+++ b/torchao/quantization/pt2e/constant_fold.py
@@ -93,6 +93,24 @@ class ConstantFolder(torch.fx.Interpreter):
         self.deferred_value = object()
         self.skip_folding_node_fn = skip_folding_node_fn
 
+        # Identify mutable buffers by finding copy_ operations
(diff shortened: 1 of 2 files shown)


torchao/quantization/pt2e/constant_fold.py
```diff
@@ -92,6 +92,24 @@ def __init__(
         self.lifted_constant_names = lifted_constant_names
         self.deferred_value = object()
         self.skip_folding_node_fn = skip_folding_node_fn
+        
+        # Identify mutable buffers by finding copy_ operations
+        self.mutable_buffers = self._find_mutable_buffers()
+
+    def _find_mutable_buffers(self) -> set[torch.fx.Node]:
```
jerryzh168, line 99: why is this change needed? is there a test that exercises this code path?

cccclai: Yeah, the added test will fail if we have this code path. The `quantize_per_tensor_default` will be folded together with the mutable buffer, which is not what we want

PR #2346, opened 2025-06-10
Add Tutorial on E2E integration into VLLM and minimal Subclass

drisspg:
Add Tutorial on E2E integration into VLLM and minimal Subclass


drisspg:
@andrewor14  do tutorials/md files automagically get put into our docs?

andrewor14:
> @andrewor14 do tutorials/md files automagically get put into our docs?

No you need to put them under docs/

andrewor14:
Actually I think this tutorial belongs better to the docs page :). The tutorials/ folder is not super visible. I don't imagine people will purposefully click into it and stumble upon this highly important tutorial, so maybe move it to the docs page instead? Unfortunately this means you will need to convert the markdown here to rst format

tutorials/torchao_vllm_integration.md
```diff
@@ -0,0 +1,450 @@
...
+To make a new TorchAO quantization method work with VLLM, you need to implement minimal tensor subclass operations that support **tensor parallelism**. VLLM uses `narrow()` and copy_ to move data from host cpu loaded in a state dict to the device, these  require these specific aten operations:
+
+```python
+@implements([aten.detach.default, aten.alias.default])
+def _(func, types, args, kwargs):
+    return return_and_correct_aliasing(
+        func, args, kwargs, args[0]._apply_fn_to_data(func)
+    )
+
+@implements([aten._to_copy.default])
```
jerryzh168: I remember this impl is incorrect, can you follow https://github.com/pytorch/ao/blob/16e2d0ae32c2ed231872616aee941d21dbda0df8/torchao/dtypes/fbgemm_int4_tensor.py#L208 for a working example?

tutorials/torchao_vllm_integration.md
```diff
@@ -0,0 +1,450 @@
...
+def _(func, types, args, kwargs):
+    return return_and_correct_aliasing(
+        func, args, kwargs, args[0]._apply_fn_to_data(torch.clone)
+    )
+
+@implements([aten.slice.Tensor])
+def _(func, types, args, kwargs):
```
jerryzh168: should we provide a real impl here as well

tutorials/torchao_vllm_integration.md
```diff
@@ -0,0 +1,450 @@
...
+@dataclass
+class MyNewQuantConfig(AOBaseConfig):
+    """Configuration for your new quantization method"""
+    bits: int = 4
+    group_size: int = 128
+    VERSION: ClassVar[int] = 1
+
+class MyQuantizedTensor(TorchAOBaseTensor):
+    """Your tensor subclass inheriting from TorchAOBaseTensor"""
+
```
jerryzh168: we might be able to improve this a bit, see https://github.com/pytorch/ao/blob/16e2d0ae32c2ed231872616aee941d21dbda0df8/torchao/dtypes/fbgemm_int4_tensor.py#L36-L37 we can probably put tensor_flatten and unflatten in TorchAOBaseTensor now

drisspg: Yeah was thinking about this as well

tutorials/torchao_vllm_integration.md
drisspg: Good catch, the one I mentioned is for autocast

drisspg:
@svekars  I remember you mentioned that we can use Markdown isntead of RST to write docs, is that true?

tutorials/torchao_vllm_integration.md
```diff
@@ -0,0 +1,465 @@
+# TorchAO Integration with VLLM: Architecture and Usage Guide
+
+This tutorial provides a comprehensive overview of how TorchAO integrates with VLLM, and what needs to be implemented to have a new technique work E2E.
+
+## Table of Contents
+* [Configuration System](#configuration-system)
+* [Usage Examples](#usage-examples)
+* [Supported Quantization Methods](#supported-quantization-methods)
+* [Advanced Features](#advanced-features)
+* [Performance Considerations](#performance-considerations)
```
andrewor14: I think these 3 are dead links right now (no corresponding sections)

tutorials/torchao_vllm_integration.md
```diff
@@ -0,0 +1,465 @@
...
+
+### 1. Quantizing Models with HuggingFace Integration
+
+```python
+from transformers import TorchAoConfig, AutoModelForCausalLM
+from torchao.quantization import Int4WeightOnlyConfig
+
+# Create quantization configuration
+quantization_config = TorchAoConfig(
+    quant_type=Int4WeightOnlyConfig(group_size=128, use_hqq=True)
```
andrewor14: Do we have a list of configs that are currently supported through `TorchAoConfig`?

tutorials/torchao_vllm_integration.md
```diff
@@ -0,0 +1,465 @@
...
+
+@implements([aten.slice.Tensor])
+def _(func, types, args, kwargs):
+    "Example impl for weird quant scheme which has 1 scale per data element"
+    self, dim, start, end, step = fill_defaults(args, 5, [0, None, None, 1])
+    assert step == 1, f"Only step=1 is supported, got step={step}"
+    if end >= self.shape[dim]:
+        end = self.shape[dim]
+    shape = list(self.shape)
+    shape[dim] = end - start
```
andrewor14: There's a lot of code in this tutorial. I feel the level of detail here is a bit distracting from the overall flow. What do you think about having a separate .py file with an actual runnable example, and then referencing that in the tutorial with some omitted code?

tutorials/torchao_vllm_integration.md
```diff
@@ -0,0 +1,465 @@
...
+
+# Quantize model
+quantization_config = TorchAoConfig(Int4WeightOnlyConfig(group_size=128, use_hqq=True))
+model = AutoModelForCausalLM.from_pretrained(
+    "meta-llama/Llama-2-7b-hf",
+    quantization_config=quantization_config
+)
+
+# REQUIRED: Must use safe_serialization=False for tensor subclasses
+model.push_to_hub("your-username/llama-2-7b-int4", safe_serialization=False)
```
andrewor14: I feel this code block is duplicated from above? Should we just mention `safe_seiralization` must be False up there?

tutorials/torchao_vllm_integration.md
```diff
@@ -0,0 +1,465 @@
...
+        k: deserialize_if_function(v) for k, v in metadata.items()
+    }
+
+    # Reconstruct tensor subclass
+    return TensorSubclass.__tensor_unflatten__(
+        plain_tensors, deserialized_metadata, None, None
+    )
+```
+
+However this is not quite ready yet.
```
andrewor14: I feel 90% users who will read this tutorial won't care much about this part. Maybe move to the bottom?

tutorials/torchao_vllm_integration.md
```diff
@@ -0,0 +1,465 @@
...
+
+1. **For Production**: Use `safe_serialization=False` when pushing quantized models
+2. **For Sharing**: Consider providing both quantized and unquantized versions
+3. **For Loading**: VLLM automatically handles tensor subclass reconstruction during model loading
+
+The TorchAO team is actively working on solutions to enable proper safetensors support for tensor subclasses.
+
+## Integration Architecture Diagrams
+
+### 1. High-Level Model Flow: Transformers → VLLM + TorchAO
```
andrewor14: Add some short text here to talk about the high level? How are these diagrams related to each other?

tutorials/torchao_vllm_integration.md
drisspg: thanks

docs/source/torchao_vllm_integration.md
```diff
@@ -0,0 +1,398 @@
+(torchao_vllm_integration)=
+# TorchAO Integration with VLLM: Architecture and Usage Guide
```
andrewor14: Maybe drop TorchAO from the title? It's a bit long in the side bar now compared to the other pages
Screenshot 2025-06-11 at 7 03 37 PM

docs/source/torchao_vllm_integration.md
```diff
@@ -0,0 +1,398 @@
...
+        self.quantized_data = quantized_data
+        self.scale = scale
+
+    def __tensor_flatten__(self) -> Tuple[List[str], List]:
+        """Serialize tensor subclass into plain tensors and metadata"""
+        return self.tensor_data_attrs, [getattr(self, attr) for attr in self.tensor_attributes]
+
+    @classmethod
+    def __tensor_unflatten__(cls, tensor_data_dict: Dict[str, torch.Tensor], tensor_attributes: List,
+                           outer_size: Optional[torch.Size], outer_stride: Optional[Tuple]) -> "MyQuantizedTensor":
```
andrewor14: formatting is weird here, maybe just do 1 per line?
Screenshot 2025-06-11 at 7 05 07 PM

docs/source/torchao_vllm_integration.md
```diff
@@ -0,0 +1,398 @@
...
+
+(create-tensor-subclass)=
+#### 1. Create Your Tensor Subclass
+
+```python
+from torchao.core.config import AOBaseConfig
+from torchao.utils import TorchAOBaseTensor
+
+@dataclass
+class MyNewQuantConfig(AOBaseConfig):
```
andrewor14, line 167: Will this config work automatically with `TorchAoConfig` or is there a hardcoded list somewhere for configs that HF knows about?

docs/source/torchao_vllm_integration.md
```diff
@@ -0,0 +1,398 @@
...
+```
+
+(compilation-benefits)=
+#### Compilation Benefits
+The overhead of tensor subclasses disappears with {func}`torch.compile`, this is on by default in VLLM.
+
+(trade-off-tensor-subclasses)=
+### Trade Off of Tensor Subclasses
+1. **Compilation**: is essential for removing subclass overhead. Without it unless your model is extremely gpu bound the overhead of dispatch on the CPU can severely impact performance.
+2. The checkpoint defines the behavior of the model. You might be saying "don't all checkpoints do this". This is true, however people typically solely think of a torch.Tensor as its data. When in actuality its a true class where it brings the Dispatcher and all the kernels ATen has registered to it. When you define your tensor subclass, you are building a separate little world. One w/ a different representation of data, but also one where you need to explicitly define what ops you support and have implementations for all the hardware you want to support. This can feel a little like spooky action at a distance at first. But it can be very powerful. Case in point is being able to support TP with only 3 definitions.
```
andrewor14, line 294: Link to this related section here? https://docs.pytorch.org/ao/stable/subclass_basic.html#what-are-tensor-subclasses. "For more details, please refer to ..."

andrewor14 (approved):
Looks great!

docs/source/torchao_vllm_integration.md
drisspg: No hard coded list

PR #2365, opened 2025-06-12
Add _apply_fn_to_data method to TorchAOBaseTensor base class

### LINKED ISSUE #2349: Add _apply_fn_to_data in AOBaseClass
# Summary

This pattern is very common and can be implemented generically.

The only times this will change is when we need to spoof our actual size, which is uncommon NJT is the only one I can think of
```Py
def _apply_fn_to_data(self, fn: Callable):
    """Applies a fn to all tensor components stored on this class"""
    tensor_names, ctx = self.__tensor_flatten__()

    # Apply the function to each tensor component
    new_tensors = {}
    for name in tensor_names:
        new_tensors[name] = fn(getattr(self, name))

    return self.__class__.__tensor_unflatten__(
        new_tensors,
        ctx,
        None,  # outer_size parameter
        None,  # outer_stride parameter
    )
```

krikera:
- Implements generic pattern for applying functions to tensor components
   - Uses __tensor_flatten__ and __tensor_unflatten__ pattern
   - Fixes #2349

### DIFF
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -49,6 +49,14 @@ class TestTorchAOBaseTensor(unittest.TestCase):
         with self.assertRaisesRegex(NotImplementedError, "arg_types"):
             l.weight = torch.nn.Parameter(MyTensor(l.weight))
 
+    def test_apply_fn_to_data(self):
+        self.assertTrue(hasattr(TorchAOBaseTensor, "_apply_fn_to_data"))
+        self.assertTrue(callable(getattr(TorchAOBaseTensor, "_apply_fn_to_data")))
+
+        method = getattr(TorchAOBaseTensor, "_apply_fn_to_data")
+        self.assertFalse(isinstance(method, classmethod))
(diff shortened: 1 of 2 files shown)


jerryzh168:
thanks, if you are interested, can you add these as well https://github.com/pytorch/ao/blob/0afa4c1bd28c82921e360ddbd1b27c9d6da5b947/torchao/dtypes/fbgemm_fp8_tensor.py#L48-L60

test/test_utils.py
```diff
@@ -49,6 +49,14 @@ def __init__(self, data):
         with self.assertRaisesRegex(NotImplementedError, "arg_types"):
             l.weight = torch.nn.Parameter(MyTensor(l.weight))
 
+    def test_apply_fn_to_data(self):
```
jerryzh168, line 52: might be good to add a small tensor and use apply_fn_to_data to implement some common methods like clone, and make sure this works functionally

PR #2371, opened 2025-06-13
[ci] fix pt2e x86 unit tests

jcaip:
Summary:

It looks like this
[change](https://github.com/pytorch/pytorch/commit/61e13782ddddf9e957c984ef11d7bb7643b871e7#diff-40b3d6cc2026cae8f139b15c4b0b05fd2e69c2715be0ce15648de79b0e15a4eb) in core was made and the corresponding test change was not duplicated here.

To unbreak CI, I copied over the test changes from the core PR, but I this just skips the test, a proper fix is still needed.

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:


jcaip:
cc @Xia-Weiwen hey just wanted to follow up on https://github.com/pytorch/ao/pull/2140 and see if you're planning to deprecate in core soon?

jerryzh168 (approved):

PR #2372, opened 2025-06-13
Gemlite generate.py fix

mobicham:
Fixes `generate.py` benchmarking with gemlite.

Also, the current code gives OOM on smaller gpus. By putting the weights on cpu first, we can avoid this issue.


torchao/_models/llama/generate.py
```diff
@@ -244,7 +244,7 @@ def encode_tokens(tokenizer, string, bos=True, device=default_device):
 
 def _load_model(checkpoint_path, device, precision):
     checkpoint = torch.load(
-        str(checkpoint_path), mmap=True, weights_only=True, map_location=device
+        str(checkpoint_path), mmap=True, weights_only=True, map_location="cpu"
```
jerryzh168, line 247: actually is the map_location change correct?

mobicham: Yes, without this, you can't load a Llama3-8B fp16 on a 24GB gpu, you just get OOM.
This is the part I mentioned:
> Also, the current code gives OOM on smaller gpus. By putting the weights on cpu first, we can avoid this issue.

jerryzh168: oh OK, but do we need to change it somewhere or it has to run in cpu

I think if user request to run on some device but it fails, we should probably not silently change the device, but instead ask user to use a different device instead

mobicham: No it still runs on gpu, there's a `model.to(device)`: https://github.com/mobicham/ao/blob/862e0cbd90f8cc5f992ae66e779022912fa4d93a/torchao/_models/llama/generate.py#L245-L256

The issue is that, loading the weights via `map_location='cuda'` + doing `model.to(device)` for some reason uses more vram leading to oom. Maybe there's a cleaner way of doing it.

jerryzh168: oh OK, makes sense, this should be OK, some quantized format does not support loading in CPU but for full/half precision model it should be fine

PR #2379, opened 2025-06-16
[float8] Prevent quantize_affine_float8/dequantize_affine_float8 decomposed on inductor

LevelDownRefine:
Fix https://github.com/pytorch/ao/issues/2228

> What we want to do now is to enable FP8 quantization in PyTorch. And similar as INT8 quantization, we need to insert quantize and dequantize ops into the graph.
> However we met problems with these q/dq ops both in the PyTorch core and Torchao.
> 
> PyTorch core:
> 
> The quantize_per_tensor op does not support FP8. We want to fix it via https://github.com/pytorch/pytorch/pull/153601. And as you commented, the op is deprecated.
> Torchao:
> 
> In the fusion pass in Inductor, we want to match the pattern fp8_weight -> torchao.dequantize_affine_float8 -> fp32_op and fuse it as fp8_weight -> weight_pack -> fp8_op. We have done so for INT8 PT2E quantization. However, the pattern matching pass is applied after a constant folding pass in Inductor:
> https://github.com/pytorch/pytorch/blob/100ec0b34aeff2b948dae33937857d0c86cf1646/torch/_inductor/fx_passes/freezing_patterns.py#L69C1-L74C1
> After constant_fold(gm), the pattern will be folded as fp32_weight -> fp32_op. Then the original pattern cannot be found any more and the FP8 semantics is lost since the pattern is entirely in fp32 now.
> For INT8, the int8_weight -> quantized_decomposed.dequantize_per_channel -> fp32_op pattern won't be folded because we mark quantized_decomposed.dequantize_per_channel impure so that it won't be folded: https://github.com/pytorch/pytorch/blob/100ec0b34aeff2b948dae33937857d0c86cf1646/torch/_inductor/constant_folding.py#L139C1-L149C1 . But for the torchao.dequantize_affine_float8, we cannot do this because
> It is an op from Torchao, which is unknown to the constant folder
> It is decomposed to smaller ops, so we cannot put it in the list as a single op.

Need to prevent q/dq decomposed both on fp8 and int8.
Based on discussion on https://github.com/pytorch/ao/pull/2299
Create this PR to prevent fp8 q/dq decomposed firstly.
On this PR, 
1. Register quantize_affine_float8 and dequantize_affine_float8 to torch.ops.torchao
2. Aligned dispatch_key with [pt](https://github.com/pytorch/pytorch/blob/6fb62931593fc10252e4994cd1a0595ebf8e990f/torch/ao/quantization/fx/_decomposed.py#L49). quantize_affine_float8 and dequantize_affine_float8 will not be decomposed.
3. Register meta func for quantize_affine_float8 and dequantize_affine_float8


LevelDownRefine:
The error of ci is irrelevant.
Will rebase

torchao/utils.py
```diff
@@ -179,7 +179,7 @@ def find_multiple(n: int, *args: int) -> int:
     return n + k - (n % k)
 
 
-def _register_custom_op(lib):
+def _register_custom_op(lib, decomposed=True):
```
jerryzh168: nit: can be more explicit, e.g. `inductor_decomposed`

jerryzh168 (approved):

torchao/utils.py
LevelDownRefine: Done

torchao/quantization/quant_primitives.py
```diff
@@ -2270,6 +2270,7 @@ def _expand_scale_to_tensor_shape(
     return expanded_scale
 
 
+@_register_custom_op(quant_lib, False)
```
jerryzh168, line 2273: we can probably make `quantize_affine_float8` and `choose_qparams_affine_float8` public as well since it's used in inductor lowering. cc @jainapurva @drisspg

LevelDownRefine:
The two ci errors are irrelevant. Not find the errors on pt2.7.1
And I change the ut path. Could you help re-run the ci. @jerryzh168

LevelDownRefine:
> The two ci errors are irrelevant. Not find the errors on pt2.7.1 And I change the ut path. Could you help re-run the ci. @jerryzh168

The two ci errors are irrelevant. Could we land this pr? @jerryzh168

PR #2383, opened 2025-06-16
remove torchao dependency from torchao build script

vkuzo:
Summary:

We should not use torchao to build torchao. Before this PR, I got stuck in a loop where I had a broken install (for some other reason) and could not uninstall torchao with this setup script, because the script required a working torchao to run.

Commenting out the offending code in this PR, leaving for a future person to actually fix it.

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -85,8 +85,6 @@ build_macos_arm_auto = use_cpp == "1" and is_arm64 and is_macos
 #       └── TORCHAO_PARALLEL_BACKEND → Backend selection (aten_openmp, executorch, etc.)
 
 
-from torchao.utils import TORCH_VERSION_AT_LEAST_2_7
-
 version_prefix = read_version()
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -388,16 +386,18 @@ def get_extensions():
...
-            if torch._C._cpu._is_avx512_supported():
-                extra_compile_args["cxx"].extend(
-                    [
-                        "-DCPU_CAPABILITY_AVX512",
-                        "-march=native",
-                        "-mfma",
-                        "-fopenmp",
-                    ]
-                )
+        # TODO(future PR): make this work without using `TORCH_VERSION_AT_LEAST_2_7`,
```
drisspg, line 389: cc @metascroy

drisspg (approved):

PR #2384, opened 2025-06-16
Add pt2e tutorials to torchao doc page

jerryzh168:
Summary:
att, after we migrate pt2e quant code from pytorch to torchao, now we also want to migrate the docs as well

Test Plan:
check generated docs

Reviewers:

Subscribers:

Tasks:

Tags:


docs/source/index.rst
```diff
@@ -43,3 +44,9 @@ for an overall introduction to the library and recent highlight and updates.
    static_quantization
    pretraining
    torchao_vllm_integration
+   tutorials_source/pt2e_quant_ptq
+   tutorials_source/pt2e_quant_qat
+   tutorials_source/pt2e_quant_x86_inductor
+   tutorials_source/pt2e_quant_xpu_inductor
+   tutorials_source/pt2e_quantizer
+   tutorials_source/openvino_quantizer
```
andrewor14, line 58: Is it possible to group these somehow? Maybe a new "PT2E Tutorials" section or something? Right now it looks kind of cluttered and makes other torchao tutorials look small

Screenshot 2025-06-16 at 4 50 05 PM

docs/source/tutorials_source/openvino_quantizer.rst
```diff
@@ -0,0 +1,250 @@
...
+We will start by performing the necessary imports, capturing the FX Graph from the eager module.
+
+.. code-block:: python
+
+    import copy
+    import openvino.torch
+    import torch
+    import torchvision.models as models
+    from torch.ao.quantization.quantize_pt2e import convert_pt2e
+    from torch.ao.quantization.quantize_pt2e import prepare_pt2e
```
andrewor14: need to update these?

docs/source/tutorials_source/pt2e_quant_x86_inductor.rst
```diff
@@ -0,0 +1,314 @@
...
+We will start by performing the necessary imports, capturing the FX Graph from the eager module.
+
+::
+
+    import torch
+    import torchvision.models as models
+    import copy
+    from torch.ao.quantization.quantize_pt2e import prepare_pt2e, convert_pt2e
+    import torch.ao.quantization.quantizer.x86_inductor_quantizer as xiq
+    from torch.ao.quantization.quantizer.x86_inductor_quantizer import X86InductorQuantizer
```
andrewor14: same here, maybe just search "torch.ao.quantization"

docs/source/tutorials_source/pt2e_quant_x86_inductor.rst
```diff
@@ -0,0 +1,314 @@
...
+  quantizer.set_global(xiq.get_default_x86_inductor_quantization_config(is_qat=True))
+  prepared_model = prepare_qat_pt2e(exported_model, quantizer)
+
+  # train omitted
+
+  converted_model = convert_pt2e(prepared_model)
+  # we have a model with aten ops doing integer computations when possible
+
+  # move the quantized model to eval mode, equivalent to `m.eval()`
+  torch.ao.quantization.move_exported_model_to_eval(converted_model)
```
andrewor14: did we also migrate this helper function over? I feel this is needed for QAT

docs/source/index.rst
andrewor14: Also what do you think about renaming these tutorials to make the titles look more consistent, e.g.

```
PyTorch 2 Export: Post-Training Quantization (PTQ) (prototype)
PyTorch 2 Export: Quantization-Aware Training (QAT) (prototype)
PyTorch 2 Export: Quantization with X86 Backend through Inductor
PyTorch 2 Export: Quantization with Intel GPU Backend through Inductor
PyTorch 2 Export: How to Write Your Own Quantizer
PyTorch 2 Export: Quantization for OpenVINO torch.compile Backend
```

andrewor14: If these are all under a "PyTorch 2 Export Tutorials" section I feel we can even just drop the prefix in front of each one of these (but keep them on the actual tutorial page if that's possible, not sure how though)

docs/source/tutorials_source/pt2e_quant_x86_inductor.rst
jerryzh168: yeah we did

docs/source/index.rst
jerryzh168: sure, updated

jerryzh168: I have put these in a separate category

jerryzh168:
cc @Xia-Weiwen @leslie-fang-intel we are moving these tutorials to torchao docs, please take a look
cc @ZhiweiYan-96 @EikanWang @liangan1 @riverliuintel @CuiYifeng on xpu tutorials
cc @daniil-lyakhov @anzr299 @alexsu52 @ynimmaga @AlexKoff88 on openvino tutorials

I just moved the docs, haven't really checking whether these new APIs works or not since it may require updates in the backend quantizers. please help check the new APIs and also help update backend related code to make sure the tutorials works as well

daniil-lyakhov:
Hello, @jerryzh168! 
Unfortunately, the OpenVINOQuantizer tutorial does not work with the torchao, but migration is not straightforward for us:
NNCF and OpenVINOQuantizer have to migrate to torchao. But NNCF supports other quantizers, like XNNPACKQuantizer and X86InductorQuantizer. We are using them to run precommit tests, but with the update, we have to add torchao and executorch (master branch) as a dependencies. (because 0.6.0 XNNPACKQuantizer still uses torch.ao)

So the question is - when there will be the next executorch release? Will it depend on the latest torch==2.7.1?
To enable tests as they are, we have to build the repo from the source, so, before doing that, I would like your opinion on that. Is it better to wait until the executorch release or better start to use master branch immediately?

jerryzh168:
> But NNCF supports other quantizers, like XNNPACKQuantizer and X86InductorQuantizer.

@daniil-lyakhov XNNPACKQuantizer is moved to executorch and X86InductorQuantizer is also moved to torchao can you change dependency to these?

ET next release is end of July: https://github.com/pytorch/executorch/issues/11075, torchao next release will happen soon.

> Will it depend on the latest torch==2.7.1?

I heard from @metascroy that it will depend on 2.8

I'm not sure how the dependency situation in nncf, but typically the way we are doing it is have some version check for the required libraries like https://github.com/pytorch/ao/blob/63a91d75f5705fb6ffff82112a7dd4e2363f8ad4/torchao/utils.py#L363-L369 and implement the new changes in a branch while keep both new and old paths working, and provide support for old path for a few releases and then remove them.

for example currently torchao only supports pytorch version >= 2.5.1

it probably doesn't have to block the doc, I can put up the doc first and feel free to update the new torchao doc and remove the old one when you are ready

PR #2387, opened 2025-06-16
Add support for resharding for fbgemm configs

jerryzh168:
Summary:
added transpose and cat op support, and also some custom transpose/reshape/unflatten support for resharding.

In the future we should probably provide examples for using distributed checkpoint for resharding

Test Plan:
python test/dtypes/test_fbgemm_int4.py -k test_transpose python test/dtypes/test_fbgemm_int4.py -k test_cat python test/dtypes/test_fbgemm_fp8.py -k test_transpose python test/dtypes/test_fbgemm_fp8.py -k test_cat

Reviewers:

Subscribers:

Tasks:

Tags:


drisspg:
Why are these ops needed? is it for DCP?

test/dtypes/test_fbgemm_fp8.py
```diff
@@ -146,6 +146,35 @@ def test_to_device(self):
...
+        # weight: (256, 128)
+        linear2 = torch.nn.Linear(128, 256, dtype=torch.bfloat16)
+
+        quantize_(linear1, self.config)
+        quantize_(linear2, self.config)
+
+        cat_weight1 = torch.cat([linear1.weight, linear2.weight], dim=0)
+        cat_weight2 = torch.cat([linear1.weight, linear2.weight], dim=1)
+        self.assertTrue(cat_weight1.shape, (512, 128))
+        self.assertTrue(cat_weight2.shape, (256, 256))
```
drisspg: can you also assert equality of bits

torchao/dtypes/fbgemm_fp8_tensor.py
```diff
@@ -27,22 +27,30 @@
 class FbgemmFp8Tensor(TorchAOBaseTensor):
     """
     TODO: needs padding for cutlass kernels
```
drisspg, line 45: what about the other attributes

torchao/dtypes/fbgemm_fp8_tensor.py
```diff
@@ -27,22 +27,30 @@
...
+    Args:
+        data_to_scale_dim: the dim mapping from float8_data to scale, e.g.
+          float8_data: (batch_size, output_channel, input_channel)
+          scale: (batch_size, output_channel) (since it's per row quantization)
```
drisspg: can we keep the scale shape the same rank as float8_data

torchao/dtypes/fbgemm_fp8_tensor.py
```diff
@@ -27,22 +27,30 @@
...
+          data_to_scale_dim: {0: 0, 1: 1}
```
drisspg: This explanation isn't very helpful / I dont know what this is doing

torchao/dtypes/fbgemm_fp8_tensor.py
```diff
@@ -82,6 +90,53 @@ def to(self, *args, **kwargs):
             self.float8_data.to(device),
             self.scale.to(device),
             self.activation_scale_ub.to(device),
+            self.data_to_scale_dim,
+            self.dtype,
+        )
+
+    def _transpose_and_reshape(self):
+        """This is added for resharding support, since the resharding logic for the model we are
```
drisspg, line 141: Do these these next two functions  need to be methods or can they be implementations of the actual ops

torchao/dtypes/fbgemm_fp8_tensor.py
```diff
@@ -82,6 +90,53 @@ def to(self, *args, **kwargs):
...
+        working with only support 2D
+        """
+        assert len(self.shape) == 3, (
+            f"Only expected to be used when the Tensor is 3D, got {len(self.shape)}"
+        )
+        dim0, dim1, dim2 = self.shape
```
drisspg, line 153: I dont understand this

torchao/dtypes/fbgemm_int4_tensor.py
```diff
@@ -144,16 +227,20 @@ def _(func, types, args, kwargs):
         args[1],
         args[2] if len(args) > 2 else None,
     )
-    orig_act_size = input_tensor.size()
+    orig_input_size = input_tensor.size()
```
drisspg, line 243: intentional?

jerryzh168: yeah, trying to be more consistent with naming

test/dtypes/test_fbgemm_fp8.py
jerryzh168: sure

torchao/dtypes/fbgemm_fp8_tensor.py
jerryzh168: these should be methods, it's specific for the hack we are doing

torchao/dtypes/fbgemm_fp8_tensor.py
jerryzh168: this is a bit confusing, removed

torchao/dtypes/fbgemm_fp8_tensor.py
jerryzh168: done

torchao/dtypes/fbgemm_fp8_tensor.py
jerryzh168: added docs for fp8 and int4

torchao/dtypes/fbgemm_fp8_tensor.py
jerryzh168: this is specific to the hack, we'll transpose the weight first and then quantize, so (dim0, dim2, dim1) is the original shape

we are restoring the shape to original shape to resharding here

torchao/dtypes/fbgemm_fp8_tensor.py
```diff
@@ -26,7 +26,16 @@
 
 class FbgemmFp8Tensor(TorchAOBaseTensor):
     """
+    Float8 Rowwise Quantized (weight) Tensor, with float8 rowwise dynamic quantization for activation.
     TODO: needs padding for cutlass kernels
+
+    Tensor Attributes:
+        float8_data: float8 raw data, dtype torchao.float8.config.e4m3_dtype
```
vkuzo: to clarify, does this mean there is a dependency on `torchao.float8.config.e4m3_dtype`? If so, I think the dependency should be refactored away to a common utility, it's not expected for that config to affect anything other than the `torchao.float8` workflow.

torchao/dtypes/fbgemm_int4_tensor.py
```diff
@@ -32,20 +32,44 @@
 
 
 class FbgemmInt4Tensor(TorchAOBaseTensor):
+    """
+    Groupwise int4 weight only quantization
+
+    Tensor Attributes:
+        packed_weight: packed int4 weight, either 2D (N, K/2) or 3D (B, N, K/2), last dimension is packed
```
vkuzo, line 39: when is a weight a 3D tensor, and why is the batch dimension in here? could you share a specific example

torchao/dtypes/fbgemm_fp8_tensor.py
jerryzh168: yeah, OK sure

torchao/dtypes/fbgemm_int4_tensor.py
jerryzh168: this is used in MoE weights I think, using the weights for bmm is a concrete example in https://github.com/pytorch/ao/blob/e29b9bd0015f9c0b62c3982b32a485028c4da548/test/dtypes/test_fbgemm_fp8.py#L116

torchao/core/config.py
```diff
@@ -284,3 +296,28 @@ def config_from_dict(data: Dict[str, Any]) -> AOBaseConfig:
...
+    """
+
+    # The preferred e4m3 type.
+    e4m3_dtype = torch.float8_e4m3fn
+
+    # The preferred e5m2 type.
+    e5m2_dtype = torch.float8_e5m2
+
+    def __post_init__(self):
+        if torch.version.hip and torch.cuda.is_available() and is_MI300():
```
vkuzo: I don't like this, I think magic methods should be avoided. It isn't super clear what would happen here, for example, if someone saves a checkpoint on an nvidia device and loads it on a rocm device which needs these other flavors.

I would recommend making the user change the float8 dtypes to the fnuz variants explicitly.

vkuzo (dismissed):

torchao/core/config.py
jerryzh168: would it complicate the user API? then they'd need to be aware whether they are quantizing for CUDA or ROCm?

jcaip: >  It isn't super clear what would happen here, for example, if someone saves a checkpoint on an nvidia device and loads it on a rocm device which needs these other flavors.

Just FYI this errors currently for torchao fp8, as the weight tensors will be loaded in the ocp variant (from the checkpoint), while the activation tensors will be the fnuz variant. 

> would it complicate the user API? then they'd need to be aware whether they are quantizing for CUDA or ROCm?

I think yes, but this is really a MI300x only problem - they only have support for fnuz fp8, MI350x will have support for OCP fp8. So maybe the solution here is to just throw a warning / not implemented error when we detecting using ocp fp8 on MI300x.

torchao/dtypes/fbgemm_fp8_tensor.py
drisspg: This function needs better description of what its trying to do

torchao/dtypes/fbgemm_fp8_tensor.py
```diff
@@ -82,6 +94,61 @@ def to(self, *args, **kwargs):
...
+
+    def _unflatten(self, num_experts):
+        """This is added for resharding support, since the resharding logic for the model we are
+        working with only support 2D
+        """
+        float8_data = self.float8_data
+        scale = self.scale
+        float8_data = float8_data.unflatten(0, (num_experts, -1)).squeeze(dim=0)
+        scale = scale.unflatten(0, (num_experts, -1)).squeeze(dim=0)
+        if self.rowwise_dim == 0:
```
drisspg, line 188: what is going on here

torchao/dtypes/fbgemm_fp8_tensor.py
```diff
@@ -264,11 +331,78 @@ def _(func, types, args, kwargs):
...
+@implements(aten.cat.default)
+def _(func, types, args, kwargs):
+    """Concatenate multiple float8 quantized tensors
+    (scale and float8_data has the same rank)
+    If the concatenation dimension is not the same as rowwise_dim, then we can just concatenate the
+    float8_data and scale directly
+    If the concatention dimension is the same as rowwise_dim, theoretically we should either
+      (1) check that scales from all tensors are equal and use the first scale
+      (2) dequantize and requantize
+    but for now we just use the first scale directly, which might have slight implication on accuaracy
```
drisspg, line 472: why can't we do 2?

torchao/dtypes/fbgemm_fp8_tensor.py
jerryzh168: this is just trying to figure out the rowwise quant dimension after unflatten op based on the original rowwise_dim

torchao/dtypes/fbgemm_fp8_tensor.py
jerryzh168: no dequantize kernels for fbgemm right now I think

torchao/core/config.py
jerryzh168: OK, I have reverted this change and have been using the dtype from `torch` namespace directly

torchao/dtypes/fbgemm_fp8_tensor.py
jerryzh168: current implementation in llama-stack is also incorrect I think, but it still works somehow, the example output still makes sense and is similar to original output

torchao/dtypes/fbgemm_fp8_tensor.py
jerryzh168: added some more descriptions, I don't have much details beyond the implementation itself actually, please let me know if it helps

jerryzh168:
landed in different forms in https://github.com/pytorch/ao/tree/main/torchao/quantization/quantize_/workflows

PR #2394, opened 2025-06-17
Add part 2 of end-to-end tutorial: fine-tuning

andrewor14:
This commit adds the QAT tutorial and a general structure for the fine-tuning tutorial, which all also include QLoRA and float8 quantized fine-tuning. It also connects the 3 tutorial parts (pre-training, fine-tuning, and serving) into one cohesive end-to-end flow with some visuals and text.

Preview it yourself here: https://docs-preview.pytorch.org/pytorch/ao/2394/finetuning.html

Screenshot 2025-06-17 at 5 35 18 PM


jerryzh168:
thanks, the same image should probably be added to other tutorials as well

docs/source/finetuning.rst
```diff
@@ -0,0 +1,291 @@
...
+
+.. image:: ../../torchao/quantization/qat/images/qat_diagram.png
+
+There are multiple options for using TorchAO's QAT for fine-tuning:
+1. Directly use our QAT APIs with your own training loop
+2. Use our integration with `TorchTune <https://github.com/pytorch/torchtune>`__
+3. Use our integratino with `Axolotl <https://github.com/axolotl-ai-cloud/axolotl>`__
+
+
+Option 1: TorchAO QAT API
```
jerryzh168: when do we want people use this?

andrewor14:
> thanks, the same image should probably be added to other tutorials as well

Yeah I added the corresponding images in part 1 and part 3

docs/source/finetuning.rst
andrewor14: I think if people have their own fine-tuning frameworks they want to use, e.g. the Llama-3.2 1B/3B quantized release did not use torchtune or axolotl

jerryzh168: nit: I think we can order these from most popular to least popular

jerryzh168 (approved):

docs/source/finetuning.rst
andrewor14: Ok, I will put the integrations first since we want OSS users to use those first

Issue #2388, opened 2025-06-16
Implement an AWQ algorithm with dynamic activation quantization for ExecuTorch


metascroy:
TorchAO has [implemented AWQ](https://github.com/pytorch/ao/tree/main/torchao/prototype/awq), and we'd like to extend the implementation to cover dynamic activation quantization to support lowering to ExecuTorch.

1. We should modify the existing AWQ algorithm to support QDQLayout [here](https://github.com/pytorch/ao/blob/main/torchao/prototype/awq/core.py#L118C22-L118C33) to support ExecuTorch

2.  The scales should have 8-bit dynamic activation quantization applied before computing the AWQ scaling.

With the above changes, we should be able to quantize a model with AWQ and lower it to ExecuTorch following instructions similar to [here](https://huggingface.co/pytorch/Qwen3-4B-8da4w).

metascroy:
cc @kimishpatel as FYI

kimishpatel:
@metascroy this one https://github.com/pytorch/ao/blob/main/torchao/prototype/awq/core.py#L118 is really in the observer. You need to replace it like here https://github.com/pytorch/ao/blob/main/torchao/prototype/awq/api.py#L150 or here https://github.com/pytorch/ao/blob/main/torchao/prototype/awq/example.py#L239, right?

metascroy:
> [@metascroy](https://github.com/metascroy) this one https://github.com/pytorch/ao/blob/main/torchao/prototype/awq/core.py#L118 is really in the observer. You need to replace it like here https://github.com/pytorch/ao/blob/main/torchao/prototype/awq/api.py#L150 or here https://github.com/pytorch/ao/blob/main/torchao/prototype/awq/example.py#L239, right?

Yes, I think those need to be modified as well.

It's actually a bit strange to me that the observer is not using the same quantization as the "convert" method.  Seems like they should be aligned?

jerryzh168:
actually as I thought through this more, it may not always work for dynamic quantization, the reasoning is the follows:

The way it works is it preserves the salient weights, in weight only quant, the only accuracy loss comes from weight quantization, so preserving the salient weights can help reduce the only source of error without affecting other things. In dynamic quant, activation quantization is also important, so if we scale the activation it could have some effect on the accuracy as well, I think this could explain why we can't always get better results with applying awq to dynamic quant.

kimishpatel:
> actually as I thought through this more, it may not always work for dynamic quantization, the reasoning is the follows:
[...]

PR #2399, opened 2025-06-18
[WIP] Add AWQ quantization with QDQLayout support for ExecuTorch

kimishpatel:
Still WIP

Good chunk of this was written by Claude

This commit implements AWQ (Activation-aware Weight Quantization) with QDQLayout support and 8-bit dynamic activation quantization for ExecuTorch compatibility, addressing GitHub issue #2388.

Key features:
- AWQObserverQDQ: Enhanced observer with 8-bit dynamic activation quantization
- QDQLayout integration for ExecuTorch compatibility
- AWQQDQConfig: Configuration for the new quantization approach
- Complete API integration with quantize_() function
- Comprehensive test suite with 8 test cases
- Usage example demonstrating the full workflow

The implementation extends the existing AWQ algorithm to support:
1. QDQLayout (Quantize-Dequantize Layout) for ExecuTorch export
2. 8-bit dynamic quantization of activation scales for better compression
3. Scale search algorithm maintaining AWQ's core optimization approach
4. Seamless integration with existing torchao quantization infrastructure

Usage:
```python
from torchao.prototype.awq import (
    insert_awq_observer_qdq_, AWQQDQConfig, _is_awq_observed_linear_qdq
)

insert_awq_observer_qdq_(model, ...)
quantize_(model, AWQQDQConfig(...), filter_fn=_is_awq_observed_linear_qdq)
```

### DIFF
--- a/torchao/prototype/awq/executorch_awq.py
+++ b/torchao/prototype/awq/executorch_awq.py
@@ -1,0 +1,441 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+AWQ implementation with QDQLayout support and dynamic activation quantization for ExecuTorch.
+
+This module extends the existing AWQ implementation to support:
+1. QDQLayout (Quantize-Dequantize Layout) for ExecuTorch compatibility
+2. 8-bit dynamic activation quantization for scales
+3. Improved quantization workflow for ExecuTorch deployment
+"""
+
+import types
+from dataclasses import dataclass
+from typing import Optional, Callable, Dict, Any
+
+import torch
+import torch.nn.functional as F
+
+import torchao
+from torchao.core.config import AOBaseConfig
+from torchao.dtypes import to_affine_quantized_intx
+from torchao.dtypes.uintx.q_dq_layout import QDQLayout
+from torchao.quantization.quant_primitives import _DTYPE_TO_BIT_WIDTH
+from torchao.quantization import to_weight_tensor_with_linear_activation_scale_metadata
+from torchao.quantization.granularity import PerGroup
+from torchao.quantization.observer import AffineQuantizedObserverBase
+from torchao.quantization.quant_api import (
+    _linear_extra_repr,
+    _replace_with_custom_fn_if_matches_filter,
+)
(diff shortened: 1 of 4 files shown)


kimishpatel:
@metascroy dont need to quite review yet, just wanted to put it up

jerryzh168:
ah thanks @kimishpatel, I also started thinking about this one, was planning to have a general AWQConfig that has a base_config of Int8DynamicActivationInt4Weight or other base configs, but it's OK we start with this first as well

torchao/prototype/awq/executorch_awq.py
```diff
@@ -0,0 +1,441 @@
...
+        for i in range(self.scale_search_space_size):
+            ratio = i * 1.0 / self.scale_search_space_size
+            scales = self.average.pow(ratio).to(self.weight.dtype)
+            scales = scales / (scales.max() * scales.min()).sqrt()
+
+            # Create quantized weight with QDQLayout
+            layout = QDQLayout()
+            tensor_dtype = torch.int8
+
+            quantized_weight = to_affine_quantized_intx(
```
jerryzh168, line 178: one thing that's slightly different is now we have to add dynamic quant here as well in observer so that scale is calculated in the context of the input will be quantized later.

kimishpatel:
> Int8DynamicActivationInt4Weight

Happy to make changes as seem apprpriate. This was mainly an exercise to leverage claude code more than an attempt to implement awq for dynamic act quant

torchao/prototype/awq/executorch_awq.py
kimishpatel: yeah tahts exactly, right? I had thought about this but forgot to instruct claude

jerryzh168: oh I see, I think we can refactor awq to work with general ao configs and it might be easier for Claude to add the new functionality: https://github.com/pytorch/ao/pull/2400

kimishpatel: my plan was to ask claude to do refactor to consolidate code in a subsequent commit. I can take a look at your PR too. Or if you want to comandeer this one, I am fine as well.

BTW I thought about what you said about dynamic act quant. Thing that worries me is that if we derive scale that is post dynamic act quantization then we cannot apply that scale before activation quantization. I mean we could but it wont be the same thing. So I am wondering whether we should do that or not. Might be just worth doing experiment also

PR #2403, opened 2025-06-18
Align scale dtype with weight in GPTQ

xiaowangintel:
Summary:
For general usage, align the data type of scale with weight Instead of the default use of bfloat16.

### DIFF
--- a/torchao/quantization/GPTQ/GPTQ.py
+++ b/torchao/quantization/GPTQ/GPTQ.py
@@ -295,7 +295,7 @@ class MultiTensor(torch.Tensor):
                     SQNR(DQ, DQ_from_qtensor),
                 )
 
-                qparams2 = cls.get_qparams_func(W)
+                qparams2 = cls.get_qparams_func(W, W.dtype)
                 Q2 = cls.quantize_func(W, qparams2)
                 DQ2 = cls.dequantize_func(Q2, qparams2).to(W.dtype)
                 old_q_out = (
@@ -444,7 +444,9 @@ class MultiTensor(torch.Tensor):
                 group_end = min(group_start + group_size, columns)
                 if group_start % group_size == 0:
                     # needed for when group_size == columns so only calculate qparams once
-                    cur_qparams = cls.get_qparams_func(W[:, group_start:group_end])
+                    cur_qparams = cls.get_qparams_func(
+                        W[:, group_start:group_end], orig_dtype
+                    )
                     all_qparams.append(cur_qparams)
 
                 for index in range(group_start, group_end):  # within each group
@@ -679,10 +681,11 @@ class Int4WeightOnlyGPTQQuantizer(GPTQQuantizer):
         else:
             self.zero_point_domain = ZeroPointDomain.FLOAT
 
-        self.get_qparams_func = lambda w: get_groupwise_affine_qparams(
+        self.get_qparams_func = lambda w, precision: get_groupwise_affine_qparams(
             w,
             n_bit,
             group_size,
+            dtype=precision,
             zero_point_domain=self.zero_point_domain,
         )
         self.quantize_func = (


xiaowangintel:
@liangan1 @jerryzh168 please help to review this pr.

liangan1 (approved):
LGTM

torchao/quantization/GPTQ/GPTQ.py
```diff
@@ -295,7 +295,7 @@ def __torch_function__(
                     SQNR(DQ, DQ_from_qtensor),
                 )
 
-                qparams2 = cls.get_qparams_func(W)
+                qparams2 = cls.get_qparams_func(W, W.dtype)
```
jerryzh168, line 298: thoughts on changing https://github.com/pytorch/ao/blob/994a4ba6c869854fcaa6ca7e118fcbd75e6c28cc/torchao/quantization/utils.py#L335C5-L335C10 to use `w.dtype` by default?

xiaowangintel: No. For gptq algorithm, the original implement for weight scales calculation does not specify scale dtype, and use default dtype=bfloat16 of get_groupwise_affine_qparams function. And it's not suitable for other model-precision. We want to align the scale data type with linear weight primitive type. Therefore, the modification passes a dtype parameter to  get_groupwise_affine_qparams.

jerryzh168: you mean it's required for gptq algorithm to use hardcoded bfloat16 by default?

> And it's not suitable for other model-precision.

what does this mean?

xiaowangintel: We want to enable float16 precision on gptq on intel GPU. But, we encountered an issue of data type misalignment, got bfloat16 scales of weight quantization following original implementation.

xiaowangintel: This pr aims to align scales dtype of weight quantization with original weight dtype.

jerryzh168: would this work for you:

```
def get_groupwise_affine_qparams(
    w,
    n_bit=4,
    groupsize=128,
    dtype=None,
    zero_point_domain=ZeroPointDomain.FLOAT,
    preserve_zero=False,
    eps=None,
):
   if dtype is None:
       dtype = w.dtype
   ...
```

torchao/quantization/GPTQ/GPTQ.py
```diff
@@ -444,7 +444,9 @@ def faster_quant(cls, H, W, device):
                 group_end = min(group_start + group_size, columns)
                 if group_start % group_size == 0:
                     # needed for when group_size == columns so only calculate qparams once
-                    cur_qparams = cls.get_qparams_func(W[:, group_start:group_end])
+                    cur_qparams = cls.get_qparams_func(
+                        W[:, group_start:group_end], orig_dtype
```
jerryzh168, line 448: will `orig_dtype` here be different from `W.dtype`

Edit:
seems like not, `orig_dtype = W.dtype`

xiaowangintel: Yes, they are different. Reading https://github.com/pytorch/ao/blob/7f7470337897230fb19dbc9c2d326d780c080a25/torchao/quantization/GPTQ/GPTQ.py#L407, model weight are converted to float in gptq quantization calibration.

torchao/quantization/GPTQ/GPTQ.py
xiaowangintel: No, the reasons given below.

torchao/quantization/GPTQ/GPTQ.py
jerryzh168: OK I see

jerryzh168 (approved):

PR #2404, opened 2025-06-18
Fix ruff broken on main

andrewor14:


jerryzh168:
oh I thought the ruff error on that PR was flaky

PR #2405, opened 2025-06-18
[float8 moe training] make using triton kernels for per-group scaling configurable

danielvegamyhre:
## Summary
- Make using triton kernels for per-group scaling configurable so we can measure speedup from avoiding d2h sync when benchmarking e2e training, and validate the expected speedup. We can remove this before graduating out of prototype.
- Improve benchmarking script, make compile configurable via an argument (I discovered a [bug](https://github.com/pytorch/pytorch/issues/156325) in torch.compile tune_scaled_grouped_mm so want the benchmark script to be usable as a repro for debugging)

### DIFF
--- a/torchao/prototype/moe_training/benchmarks/benchmark_scaled_grouped_mm.py
+++ b/torchao/prototype/moe_training/benchmarks/benchmark_scaled_grouped_mm.py
@@ -110,19 +110,12 @@ def run_experiment(
         torch.cuda.synchronize()
 
     # benchmark torch
-    if args.compile:
-        compiled_fwd_bwd = torch.compile(forward_backward)
-        warmup(compiled_fwd_bwd, A, B_t, offs, use_triton=False)
-        start_time_ns = time.perf_counter_ns()
-        compiled_fwd_bwd(A, B_t, offs, use_triton=False)
-        torch_time_ns = time.perf_counter_ns() - start_time_ns
-        torch_time_us = torch_time_ns / 1e3
-    else:
-        warmup(forward_backward, A, B_t, offs, use_triton=False)
-        start_time_ns = time.perf_counter_ns()
-        forward_backward(A, B_t, offs, use_triton=False)
-        torch_time_ns = time.perf_counter_ns() - start_time_ns
-        torch_time_us = torch_time_ns / 1e3
+    torch_func = torch.compile(forward_backward) if args.compile else forward_backward
+    warmup(torch_func, A, B_t, offs, use_triton=False)
+    start_time_ns = time.perf_counter_ns()
+    torch_func(A, B_t, offs, use_triton=False)
+    torch_time_ns = time.perf_counter_ns() - start_time_ns
+    torch_time_us = torch_time_ns / 1e3
 
     # benchmark triton
     warmup(forward_backward, A, B_t, offs, use_triton=True)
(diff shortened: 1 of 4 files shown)


danielvegamyhre:
@drisspg @vkuzo for review

torchao/prototype/moe_training/benchmarks/benchmark_scaled_grouped_mm.py
```diff
@@ -92,26 +98,53 @@ def warmup(func, *args, **kwargs):
...
+            use_triton_for_per_group_scales=use_triton,
+        )
         out.sum().backward()
-
-    # bench triton
-    warmup(forward_backward, A, B_t, offs)
+        torch.cuda.synchronize()
+
+    # benchmark torch
+    if args.compile:
```
drisspg: nit: func = compiled_fwd_bwd if compile else forward_backward
to make it easier to read

torchao/prototype/moe_training/conversion_utils.py
```diff
@@ -28,7 +28,7 @@ class MoETrainingConfig(AOBaseConfig):
     For all other ops, ScaledGroupedMMTensor behaves like a regular torch.Tensor.
     """
 
-    pass
+    use_triton_for_per_group_scales: bool = True
```
drisspg, line 32: maybe a lil note: used for testing but will remove .. etc

torchao/prototype/moe_training/scaled_grouped_mm.py
```diff
@@ -195,25 +216,48 @@ def backward(ctx, grad_output: torch.Tensor):
 
         # grad_B is a special case. both operands of the grouped gemm will be 2D with offsets determing the "groups."
         # Compute scales for grad_output_t and A, which are both 2D tensors with offsets which define the "jagged" groups.
-        grad_output_t_fp8_row_major, grad_output_t_scales = (
-            triton_fp8_row_major_jagged_rowwise_scales(
-                grad_output_t_row_major,
+        if use_triton_for_per_group_scales:
+            grad_output_t_fp8_row_major, grad_output_t_scales = (
+                triton_fp8_row_major_jagged_rowwise_scales(
```
drisspg: nit yu could also do the function assignment trick here to make it easier to read

torchao/prototype/moe_training/tensor.py
```diff
@@ -12,9 +12,20 @@ class ScaledGroupedMMTensor(torch.Tensor):
 
     grouped_mm_func_name = "_grouped_mm"
     offs_arg_name = "offs"
+    use_triton_for_per_group_scales = True
 
-    def __init__(self, data: torch.Tensor):
+    def __new__(cls, data: torch.Tensor, use_triton_for_per_group_scales: bool = True):
+        cls.use_triton_for_per_group_scales = use_triton_for_per_group_scales
```
drisspg: this looks weird, I think you should just let it fall through to __init__ and then set it on the instance

danielvegamyhre: I agree, I did that at first but then in the __torch_function__ i don't have access to the instance because it's a classmethod. I'd like to make it less weird though, am open to ideas. 

Technically since this is just temporary for benchmarking comparison, I could just condition on an env var and avoid plumbing this through everywhere entirely, but that seemed weird as well. In retrospect I actually think it would be cleaner though. What do you think?

torchao/prototype/moe_training/benchmarks/benchmark_scaled_grouped_mm.py
danielvegamyhre: Done

torchao/prototype/moe_training/scaled_grouped_mm.py
danielvegamyhre: good call, done

torchao/prototype/moe_training/tensor.py
drisspg: One inputs to torch function has to be one of these subclasses, right?  So can't you just grab it from whatever instance is the subclass?

danielvegamyhre: True, that should work - updated.

drisspg (approved):

PR #2407, opened 2025-06-18
Groupwise low bit LUT based model quantization.

szyszyzys:


metascroy (changes requested):
The unit tests are failing here, and the PR needs to be broken up into smaller pieces.

PR #2408, opened 2025-06-18
WIP NVfp4

drisspg:
WIP NVfp4


drisspg:
@vkuzo Curious if you agree we should roll this into the existing mx tensor?

torchao/prototype/mx_formats/mx_subclass.py
```diff
@@ -151,7 +156,116 @@ def _mx_inference_linear_transform(
...
+        block_size=config.block_size,
+    )
+
+    input_quant_func = _input_activation_quant_func_nvfp4
+    input_quant_kwargs = {
+        "block_size": config.block_size,
+        "scale": None,
+    }
+
+    quantized_weight = to_linear_activation_quantized(
```
vkuzo: can we just write the logic here instead of using `to_linear_activation_quantized`?  I remember same feedback on the mxfp4 inference tensor.

torchao/prototype/mx_formats/mx_subclass.py
```diff
@@ -151,7 +156,116 @@ def _mx_inference_linear_transform(
...
+    weight_dtype: torch.dtype = field(default_factory=_get_nvfp4_dtype)
+
+    # NVFP4 uses E4M3 scales
+    scale_dtype: torch.dtype = torch.float8_e4m3fn
+
+    # CUBLAS is preferred for NVFP4 with VEC16_UE4M3 support
+    gemm_kernel_choice: MXGemmKernelChoice = MXGemmKernelChoice.CUBLAS
+
+    # Set some magic perf settings
+    set_inductor_config: bool = False
```
vkuzo: nit: move outside of this API? It's convenient to have it here but also breaks separation of concerns.

torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -0,0 +1,489 @@
...
+        # Unpack the float
+        data_unpacked = unpack_uint4(self._data.view(torch.uint8))
+        data_f32 = f4_unpacked_to_f32(data_unpacked)
+
+        # Reshape and scale
+        orig_shape = self.shape
+        M, K = orig_shape[0], orig_shape[1]
+        data_f32 = data_f32.view(M, K // self._block_size, self._block_size)
+        scale_e4m3_reshaped = self._scale_e4m3.view(M, K // self._block_size, 1)
+        data_scaled = data_f32 * scale_e4m3_reshaped.to(torch.float32)
```
vkuzo, line 210: is there a second rescaling somewhere?

torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -0,0 +1,489 @@
...
+    assert a._data.is_contiguous()
+    assert b._data.t().is_contiguous()
+    assert a._block_size == 16, f"NVFP4 requires block_size=16, got {a._block_size}"
+    assert b._block_size == 16, f"NVFP4 requires block_size=16, got {b._block_size}"
+
+    # Get tensor dimensions
+    M, K = a.shape[0], a.shape[1]
+    N = b.shape[1]
+
+    # Swizzle Dizzle
```
vkuzo, line 425: lol

torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -0,0 +1,489 @@
...
+    )
+    assert data_hp.numel() % block_size == 0, "unsupported"
+    assert data_hp.is_contiguous(), "unsupported"
+    assert block_size == 16, "NVFP4 requires block_size=16"
+
+    orig_shape = data_hp.shape
+    data_hp = data_hp.reshape(orig_shape[0], -1, block_size)
+
+    max_abs = torch.amax(torch.abs(data_hp), dim=-1)
+    block_scales = torch.clamp(max_abs / F4_E2M1_MAX, min=E4M3_EPS, max=F8E4M3_MAX)
```
vkuzo, line 469: maybe we can add an epsilon to guard against all-zeros?

torchao/prototype/mx_formats/nvfp4_tensor.py
drisspg: ~~my two-level scaling is ont working correctly yet~~
Fixed

torchao/prototype/mx_formats/mx_subclass.py
drisspg: So you can't just move the logic out here, the entirety of the forward behavior has to be "wrapped" by the subclass. Currently there are two ways to do that, without changing nn.modules. 
1. Like above; this is subclass composition 
2. The other is to copy the same behavior into the implementations of the ops, 
e.g.

NVFP4's dispatch would need to copy: https://github.com/pytorch/ao/blob/7192edffe9df060edcce1733a9b333c44d70d19c/torchao/quantization/linear_activation_quantized_tensor.py#L135-L186

Not the end of the world. But for some subclasses that serve dual purpose (dyanmic + weight only, + static, + training) it can be alot of switch statements in the ops as opposed to having the base subclass + some sugar

torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -0,0 +1,538 @@
...
+
+    max_abs = torch.amax(torch.abs(data_hp), dim=-1)
+    # These scales are currently in fp32, we are going to `quantize` them to e4m3
+    block_scale = max_abs / F4_E2M1_MAX
+
+    out_scales = None
+    if per_tensor_scale is None:
+        # We are doing single level scaling
+        block_scale_fp8 = torch.clamp(block_scale, min=E4M3_EPS, max=F8E4M3_MAX).to(
+            torch.float8_e4m3fn
```
drisspg, line 589: The down up down pattern of casts is probs overkill

torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -0,0 +1,538 @@
...
+        b_scale_blocked.view(torch.float8_e4m3fn),
+        bias=bias,
+        out_dtype=a._orig_dtype,
+        # scale_result=scale_result,
+    )
+
+    # THIS IS A WORKAROUND:
+    # RuntimeError: CUDA error: CUBLAS_STATUS_INVALID_VALUE when calling
+    if scale_result is not None:
+        result = result * scale_result
```
gau-nernst: I think if there is `bias`, `scale_result` should be applied before `bias`?

drisspg: You are totally right

drisspg:
Weight only fails w/ compile and bisected to:

Likely from the work around to get triton to not error on e2m1
```Py

Disabling lowerings fixed the issue.
Starting bisect by getting upper bound.
Upper bound of 38 found for inductor.
Bisecting inductor - lowerings (Range: [0, 38], Midpoint: 19)
Bisecting inductor - lowerings (Range: [20, 38], Midpoint: 29)
Bisecting inductor - lowerings (Range: [30, 38], Midpoint: 34)
Bisecting inductor - lowerings (Range: [35, 38], Midpoint: 36)
Bisecting inductor - lowerings (Range: [35, 36], Midpoint: 35)
Binary search completed for inductor - lowerings. The bisect number is 36. Debug info: convert_element_type_5
Bisection status deleted.
   Bisection result: BisectionResult(backend='inductor', subsystem='lowerings', bisect_number=36, debug_info='convert_element_type_5')

6. Testing inductor config workarounds for WEIGHT_ONLY:
   {'inductor.coordinate_descent_tuning': False}           ERROR
   {'inductor.force_fuse_int_mm_with_mul': False}          ERROR
   {'inductor.post_grad_passes': False}                    ERROR
   {'inductor.pattern_matcher': False}                     ERROR
   {'inductor.epilogue_fusion': False}                     ERROR
   {'inductor.max_autotune': False}                        ERROR
   {'triton.autotune_pointwise': False}                    ✗ 3.1dB
   {'inductor.benchmark_kernel': False}                    ERROR
   {'inductor.aggressive_fusion': False}                   ERROR

7. Testing other compile backends:
   Backend 'eager': SQNR = 20.00 dB ✓
   Backend 'aot_eager': SQNR = 20.00 dB ✓
skipping cudagraphs due to skipping cudagraphs due to cpu device (_tensor_constant0). Found from : 
   File "/home/drisspg/.conda/envs/nightly/lib/python3.12/site-packages/torch/_dynamo/external_utils.py", line 70, in inner
    return fn(*args, **kwargs)

   Backend 'cudagraphs': SQNR = 20.00 dB ✓
   ```

**Comment by drisspg** [comment]
@vkuzo updated to use the mm_config

File: torchao/prototype/mx_formats/mx_subclass.py [hunk 8]
```diff
@@ -151,7 +150,90 @@ def _mx_inference_linear_transform(
...
+        )
+        assert self.weight_dtype == torch.float4_e2m1fn_x2, (
+            f"NVFP4 requires weight_dtype=float4_e2m1fn_x2, got {self.weight_dtype}"
+        )
+        assert self.scale_dtype == torch.float8_e4m3fn, (
+            f"NVFP4 requires scale_dtype=float8_e4m3fn, got {self.scale_dtype}"
+        )
+        assert self.block_size == 16, (
+            f"NVFP4 requires block_size=16, got {self.block_size}"
+        )
```
gau-nernst: just curious. what's the point of exposing all of these when only a specific value is accepted.

torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -0,0 +1,619 @@
...
+    assert b._block_size == 16, f"NVFP4 requires block_size=16, got {b._block_size}"
+
+    M, K = a.shape[0], a.shape[1]
+    N = b.shape[1]
+
+    # Swizzle Dizzle
+    a_scale = a._scale_e4m3.view(M, K // a._block_size)
+    b_scale = b._scale_e4m3.view(N, K // b._block_size)
+    a_scale_blocked = to_blocked(a_scale)
+    b_scale_blocked = to_blocked(b_scale)
```
gau-nernst, line 429: I have wondered about this for MX-dtype as well. It makes sense for MX-dtype to have scale swizzling here since we may want the layout to be vendor-neutral. But NVFP4 is specific to NVIDIA, so why not put this under `to_nvfp4()`?

torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -0,0 +1,619 @@
...
+        out_dtype=a._orig_dtype,
+        # scale_result=scale_result,  # Not supported yet
+    )
+
+    if scale_result is not None:
+        result = result * scale_result
+
+    # Add bias after scaling if needed
+    if should_add_bias_separately:
+        result = result + bias
```
gau-nernst: Not sure how important it is, but if `scale_result` is FP32 (NVIDIA uses FP32 for the per-tensor scale), `result` will become FP32 due to upcasting rule.

gau-nernst (approved):
Just some comments

torchao/prototype/mx_formats/mx_subclass.py
drisspg: This is a good point, my original intent was to not make another subclass, and just merge in w/ mxfp, cc @vkuzo I imagine that we wan't this separated? I started to work on a observer for this since without  it this just a worse mxfp4.

torchao/prototype/mx_formats/nvfp4_tensor.py
drisspg: ahh nice catch, thanks

torchao/prototype/mx_formats/nvfp4_tensor.py
drisspg: This is a good question, for mx I recently confirmed that AMD does not require swizziling, this is a good point about NVFP4, I am actually going update this code path to 	`cache` the swizzled layout for the weight

torchao/prototype/mx_formats/mx_subclass.py
vkuzo: I'm open, I guess the only thing this has in common with mxfp4 is the fp4 part.  Either separate or together could work.  Since you already have it separate, seems fine?

torchao/prototype/mx_formats/mx_subclass.py
```diff
@@ -151,7 +150,93 @@ def _mx_inference_linear_transform(
...
 
+def _get_nvfp4_dtype():
+    """Factory function for NVFP4 dtype defaults."""
+    if not TORCH_VERSION_AT_LEAST_2_8:
+        raise RuntimeError("NVFP4InferenceConfig requires PyTorch 2.8 or later")
+    return torch.float4_e2m1fn_x2
+
+
+@dataclass
+class NVFP4InferenceConfig(AOBaseConfig):
```
vkuzo, line 155: is there precedent for `Inference` vs `Training` in the config?  Would we use the same config for, say, QAT?

if you want to align this later when bringing out of prototype, sgtm

torchao/prototype/mx_formats/mx_subclass.py
```diff
@@ -151,7 +150,93 @@ def _mx_inference_linear_transform(
...
+
+    # NVFP4 uses FP4 data
+    activation_dtype: torch.dtype = field(default_factory=_get_nvfp4_dtype)
+    weight_dtype: torch.dtype = field(default_factory=_get_nvfp4_dtype)
+
+    # NVFP4 uses E4M3 scales
+    scale_dtype: torch.dtype = torch.float8_e4m3fn
+
+    # CUBLAS is preferred for NVFP4 with VEC16_UE4M3 support
+    gemm_kernel_choice: MXGemmKernelChoice = MXGemmKernelChoice.CUBLAS
```
vkuzo: I added this originally for debugging, not sure if we want this to be exposed publically for production. thoughts?  I kind of want the best kernel to be selected for me as the default, and the ability to override for debugging.

torchao/prototype/mx_formats/mx_subclass.py
```diff
@@ -151,7 +150,93 @@ def _mx_inference_linear_transform(
...
+            f"NVFP4 requires block_size=16, got {self.block_size}"
+        )
+
+
+@register_quantize_module_handler(NVFP4InferenceConfig)
+def _nvfp4_inference_linear_transform(
+    module: torch.nn.Linear, config: NVFP4InferenceConfig
+):
+    """Quantization handler for NVFP4InferenceConfig"""
+    assert is_sm_at_least_100(), "NVFP4 is only supported on sm100+ machines"
```
vkuzo: can we add an emulation mode, and move this check to the non-emulated path? it's really nice to have for having CI, and for being able to debug most things without needing a new GPU card

vkuzo: doesn't have to be in this PR, for the future

torchao/prototype/mx_formats/nvfp4_tensor.py
vkuzo: I'm still confused.  If there is per-tensor scaling, we scale by the per-tensor scale to convert to nvfp4.  Do we need to undo this scaling when converting back to high precision?

vkuzo: I see, it looks like it lives in `self.get_scales()`.  Maybe we can rename the function to make it obvious that both scales are there, like `self.get_blockwise_and_maybe_tensorwise_scales()`?

torchao/prototype/mx_formats/nvfp4_tensor.py
vkuzo: the swizzling is kernel specific, IMO makes sense to have it here close to the kernel call in absence of other motivating factors

torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -0,0 +1,619 @@
...
+
+    return result
+
+
+@implements([torch.nn.functional.linear, aten.linear.default])
+def nvfp4_linear(func, types, args, kwargs):
```
vkuzo, line 466: curious on why we need both `linear` and `mm` flavors instead of picking one? I thought `linear` was for torch function and mm variants for torch dispatch?

torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -0,0 +1,619 @@
...
+        args[1],
+        args[2] if len(args) > 2 else None,
+    )
+
+    if not isinstance(weight_tensor, NVFP4Tensor):
+        raise NotImplementedError("NVFP4Tensor: weight must be NVFP4Tensor")
+
+    config = weight_tensor.mm_config
+
+    if config == NVFP4MMConfig.WEIGHT_ONLY:
```
vkuzo, line 478: thanks for doing this, this is very clear and the reader doesn't need to jump around the codebase to understand what's happening

torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -0,0 +1,619 @@
...
+
+    Args:
+        amax: Per-tensor amax tensor
+
+    Returns:
+        torch.Tensor: Per-tensor scale tensor
+    """
+    return (
+        torch.clamp(amax / F8E4M3_MAX, min=E4M3_EPS, max=F8E4M3_MAX)
+        .to(torch.float8_e4m3fn)
```
vkuzo: why is this cast to float8 necessary?

torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -0,0 +1,619 @@
...
+        scaled_block_scales_fp32 = scaled_block_scales_fp8.to(torch.float32)
+        # We "temporarily" dequant the scaled_block_scales_fp32 to get the per_tensor_scale
+        # To apply to data
+        total_scale = per_tensor_scale * scaled_block_scales_fp32
+        data_scaled = data_hp / total_scale.unsqueeze(-1)
+        out_scales = scaled_block_scales_fp8
+
+    data_scaled = torch.clamp(data_scaled, -F4_E2M1_MAX, F4_E2M1_MAX)
+    data_scaled = data_scaled.view(orig_shape)
+    data_lp = f32_to_f4_unpacked(data_scaled.float())
```
vkuzo, line 613: I remember this function was being pretty slow when we wrote the emulation code, curious how performance is now?  There are likely things we can do to make it faster if this is in the hot path.

torchao/prototype/mx_formats/mx_subclass.py
drisspg: so the `weight only` mode is fully emulated, I will unskip for those

torchao/prototype/mx_formats/nvfp4_tensor.py
drisspg: We only need the mm and addmm in most circumstances, however if you run under inference mdoe that is pre-dispatch you end up see these linear ops

torchao/prototype/mx_formats/mx_subclass.py
vkuzo: I mean an emulation mode for all recipes, where the gemm would happen in high precision, for testing and debugging

torchao/prototype/mx_formats/mx_subclass.py
drisspg: Yeah this is a lil funky, from a quick look it seems like QAT has its own naming conventions `[IntXQuantizationAwareTrainingConfig](https://github.com/pytorch/ao/blob/cedadc741954f47a9e9efac2aa584701f125bc73/torchao/quantization/qat/api.py#L242)`
I agree though that we should reconcile before moving out of prototype, will add a todo

torchao/prototype/mx_formats/mx_subclass.py
drisspg: Kept this when I was thinking we were going to merge in to mx. I am going to remove since we expect that cublas to be the most performant in 99% of cases with the exception of maybe decode gemmv shapes. 

Especially since we dont actually have a cutlass binding yet for this flow

torchao/prototype/mx_formats/nvfp4_tensor.py
drisspg: Ahhh good catch, this one shouldn't be needed since we only ever use it in fp32

torchao/prototype/mx_formats/nvfp4_tensor.py
drisspg: Yeah @gaunerst has a nice triton kernel for doing the cast + pack that I am going to try in a follow up PR

PR #2410, opened 2025-06-18
[float8] add _auto_filter_for_recipe to float8

danielvegamyhre:
Part of https://github.com/pytorch/torchtitan/issues/1207

## Problem
- float8 rowwise + vanilla TP in torchtitan had flat perf with respect to bfloat. 
- RCA In https://github.com/pytorch/torchtitan/issues/1207 found attention.wk and attention.wv layers were so small that float8 rowwise conversion resulted in significant slowdown (approx 40%) for those linears, thus the perf benefits from fp8 rowwise conversion on larger linears were nullified.
- This is because the default `filter_fqns` for float8 model conversion are fine for the fp8 tensorwise recipe, but bad for the float8 rowwise recipe. 

### Solution 

This has been a footgun for various users as well (including Poolside), so I created an "auto filter" (https://github.com/pytorch/ao/pull/2410) which automatically filters Linears for a given float8 recipe, by checking for the following criteria:

1. dims not divisible by 16 (hardware requirement for float8)
2. dim sizes below thresholds that will result in worse perf **for that given recipe**, using simple heuristics based on the linked recipe perf tables above.
3. fqn matches one of the user defined `filter_fqns`

I integrated a [PoC into torchtitan](https://github.com/pytorch/torchtitan/pull/1319) and the auto filter improved fp8 rowwise perf both local Llama3 8b run and Llama3 70b MAST run, compared to the default filter_fn we have now.

It prevents users from hitting this common footgun, while also preserving the flexibility to define their model-specific fqns.

## Results 
See https://github.com/pytorch/torchtitan/issues/1207 for Llama3 70b results, TL;DR is filtering wk and wv improves TPS  ~10% for vanilla TP and ~15% for async TP.

### DIFF
--- a/torchao/float8/float8_linear_utils.py
+++ b/torchao/float8/float8_linear_utils.py
@@ -5,7 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 import logging
 from functools import partial
-from typing import Callable, List, Optional
+from typing import Callable, List, Optional, Union
 
 import torch.nn as nn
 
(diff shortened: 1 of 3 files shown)


danielvegamyhre:
cc @vkuzo for review

torchao/float8/float8_linear_utils.py
```diff
@@ -113,3 +114,78 @@ def convert_to_float8_training(
         from_float,
         module_filter_fn=module_filter_fn,
     )
+
+
+def auto_filter_for_recipe(
```
vkuzo: nit: prefix with underscore

torchao/float8/float8_linear_utils.py
```diff
@@ -113,3 +114,78 @@ def convert_to_float8_training(
...
+        return False
+
+    # All dims must be divisible by 16 due to float8 hardware requirements.
+    K, N = mod.weight.shape
+    dims_multiples_of_16 = K % 16 == 0 and N % 16 == 0
+    if not dims_multiples_of_16:
+        return False
+
+    # Dims below these thresholds will result in worse performance
+    # (see https://github.com/pytorch/ao/tree/main/torchao/float8#rowwise-scaling)
```
vkuzo, line 163: note: this table is hardware specific.  I haven't calculated what the relevant dimensions are for consumer cards.  It's going to be hard to make this bulletproof, but and experimental API without BC guarantees is fine.

torchao/float8/float8_linear_utils.py
```diff
@@ -113,3 +114,78 @@ def convert_to_float8_training(
...
+    if is_filtered_fqn:
+        return False
+
+    # All dims must be divisible by 16 due to float8 hardware requirements.
+    K, N = mod.weight.shape
+    dims_multiples_of_16 = K % 16 == 0 and N % 16 == 0
+    if not dims_multiples_of_16:
+        return False
+
+    # Dims below these thresholds will result in worse performance
```
vkuzo: s/will/may/g, it also depends on M

torchao/float8/float8_linear_utils.py
```diff
@@ -113,3 +114,78 @@ def convert_to_float8_training(
...
```
vkuzo: s/will/may/g

torchao/float8/float8_linear_utils.py
vkuzo: nit: expose in float8/__init__.py to avoid having callers import from this file

vkuzo (approved):
looks good, can we make sure the name has an underscore and also add a test before landing?

torchao/float8/float8_linear_utils.py
danielvegamyhre: Added a note about this in the docstring and in these comments.

PR #2420, opened 2025-06-21
enable to_mxfp8 cast for DTensor

vkuzo:
Summary:

1. add a test for casting a DTensor to mxfp8
2. make the test pass:
  a. remove addition of epsilon, it's not supported in DTensor world but
  we also don't need it anymore since we are no longer using `log2`
  anywhere.
  b. replace `<<` with `torch.bitwise_left_shift` and `>>` with
  `torch.bitwise_right_shift`.  The short versions are silently broken
  for DTensor inputs, but the verbose versions work.
3. set up the wiring for testing mxfp8 with TP on a toy model. Note that
   making this work is split for the next PR, as this PR got too large.

Test Plan:

```bash
./test/prototype/mx_formats/test_dtensor.sh
./test/float8/test_dtensor.sh
pytest test/prototype/mx_formats/
```

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/prototype/mx_formats/test_dtensor.py
+++ b/test/prototype/mx_formats/test_dtensor.py
@@ -1,0 +1,98 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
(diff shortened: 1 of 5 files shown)


vkuzo:
* #2434
* __->__ #2420

PR #2421, opened 2025-06-22
Add support for Int4GroupwisePreshuffleTensor for fbgemm

jerryzh168:
Stacked PRs:
 * #2474
 * #2463
 * #2479
 * #2437
 * #2422
 * __->__#2421

--- --- ---

### Add support for Int4GroupwisePreshuffleTensor for fbgemm

Summary:
Note: slice is not working yet, others are working

Test Plan:
python test/dtypes/test_int4_groupwise_preshuffle.py

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/quantization/quantize_/__init__.py
+++ b/torchao/quantization/quantize_/__init__.py
@@ -1,0 +1,9 @@
+from .int4_groupwise_preshuffle_tensor import (
+    Int4GroupwisePreshuffleTensor,
+)
+
+Int4GroupwisePreshuffleTensor.__module__ = "torchao.quantization"
+
+__all__ = [
+    "Int4GroupwisePreshuffleTensor",
+]
--- a/torchao/quantization/quantize_/int4_groupwise_preshuffle_tensor.py
+++ b/torchao/quantization/quantize_/int4_groupwise_preshuffle_tensor.py
@@ -1,0 +1,353 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+
+
+import importlib.util
+from typing import List
+
+import torch
+from torch.utils._python_dispatch import return_and_correct_aliasing
+
+from torchao.utils import (
+    TORCH_VERSION_AT_LEAST_2_5,
+    TorchAOBaseTensor,
+    fill_defaults,
+)
+
+__all__ = [
+    "Int4GroupwisePreshuffleTensor",
+]
+
+aten = torch.ops.aten
+
+
+if importlib.util.find_spec("fbgemm_gpu") is None:
+    quantize_int4_preshuffle = None
+else:
+    from fbgemm_gpu.experimental.gen_ai.quantize import quantize_int4_preshuffle
+
+
+class Int4GroupwisePreshuffleTensor(TorchAOBaseTensor):
+    """
+    Groupwise int4 weight only quantization
+
+    Tensor Attributes:
+        packed_weight: packed int4 weight, either 2D (N, K/2) or 3D (B, N, K/2), last dimension is packed
+        group_scale: (K/group_size, N) for 2D Tensor, (B, N, K/group_size) for 3D Tensor
+               dtype is the same as the original Tensor dtype
+        group_zero: (K/group_size, N) for 2D Tensor, (B, N, K/group_size) for 3D Tensor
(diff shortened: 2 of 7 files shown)


torchao/dtypes/int4_groupwise_preshuffle_tensor.py
```diff
@@ -0,0 +1,424 @@
...
+
+if importlib.util.find_spec("fbgemm_gpu") is None:
+    quantize_int4_preshuffle = None
+else:
+    from fbgemm_gpu.experimental.gen_ai.quantize import quantize_int4_preshuffle
+
+
+class Int4GroupwisePreshuffleTensor(TorchAOBaseTensor):
+    """
+    Groupwise int4 weight only quantization
```
vkuzo: can we add a description of preshuffle to this docblock?

torchao/dtypes/int4_groupwise_preshuffle_tensor.py
```diff
@@ -0,0 +1,424 @@
...
+    "Int4GroupwisePreshuffleTensor",
+]
+
+aten = torch.ops.aten
+
+
+if importlib.util.find_spec("fbgemm_gpu") is None:
+    quantize_int4_preshuffle = None
+else:
+    from fbgemm_gpu.experimental.gen_ai.quantize import quantize_int4_preshuffle
```
vkuzo: is this a prototype API? If yes, should the torchao version also be prototype? what does "experimental" mean in the folder structure here?

torchao/dtypes/int4_groupwise_preshuffle_tensor.py
```diff
@@ -0,0 +1,424 @@
...
+        self.group_scale,
+        self.group_zero,
+        self.group_size,
+        shape_multiplier,
+        tensor_shape,
+    )
+    return return_and_correct_aliasing(func, args, kwargs, new)
+
+
+to_int4_groupwise_preshuffle = Int4GroupwisePreshuffleTensor.from_float
```
vkuzo: nit: why not just call the original function, to reduce the amount of indirection?

jerryzh168: OK, makes sense, this was also from previous design that we want to create some factory methods but it's not too important now I think. I can remove it

torchao/dtypes/int4_groupwise_preshuffle_tensor.py
jerryzh168: it is stable and production ready, and used in production. it's just bad naming according to @jwfromm, and they have a plan to get rid of it

torchao/quantization/quantize_/int4_groupwise_preshuffle_tensor.py
```diff
@@ -0,0 +1,343 @@
...
+
+    Non-Tensor Attributes:
+        group_size: the group size for groupwise quantization
+        shape_multiplier: is the multipler from packed_weight to the real weight, since
+        we pack the weight for int4, for example, when we pack the last dimension for
+        a 2D tensor, the shape_multiplier will be [1, 2]
+        shape: shape of the original Tensor
+
+    Note:
+       preshuffle means the weight is rearranged for more efficient use of loading instructions
```
vkuzo: it would be good to share the specifics of the preshuffle transformation, either here or via a link a user can follow

torchao/quantization/quantize_/__init__.py
```diff
@@ -0,0 +1,9 @@
+from .int4_groupwise_preshuffle_tensor import (
+    Int4GroupwisePreshuffleTensor,
+)
+
+Int4GroupwisePreshuffleTensor.__module__ = "torchao.quantization"
```
vkuzo, line 5: can we confirm (by actually testing it) that we can change the directory location later without breaking BC?

jerryzh168: I have a test that verifies the loaded weight have module path `torchao.quantization.Int4GroupwisePreshuffleTensor`, this (`type(tensor)`)is used in the load code path: https://github.com/pytorch/pytorch/blob/d4b8857e51a089b7e0e722689398c5c3ada274c9/torch/_tensor.py#L262 which gives us good confidence that it would work as long as we do this

but I can do a e2e test a bit later by uploading the file in huggingface hub and change the path locally to verify as well

jerryzh168: added in https://github.com/pytorch/ao/pull/2437

PR #2425, opened 2025-06-23
[float8 moe training] Add TP support

danielvegamyhre:
Note: this should be merged AFTER this bug fix: https://github.com/pytorch/ao/pull/2451 I will rebase and retest all of this once that's merged.

## Summary
- Add TP support for routed experts and shared expert.
    - Make target dim of scale squeze() ops explicit to handle both 2D and 3D "A" tensors (routed experts case has 2D "A", shared expert has 3D "A").
    - Make `offs` optional to handle shared_expert case where num_experts=1 (scaled grouped GEMM only processing 1 expert)
- Add debug logging

## Test plan
- Added integration test using torchtitan llama4 TP implementation. Test cases for (1) routed experts, and  (2) routed experts + shared expert.
- Manual testing with torchtitan llama4 debug model with TP=2, targeting routed experts AND shared experts works ([logs](https://www.internalfb.com/phabricator/paste/view/P1849252048)). 
- Manual testing with torch titan llama4 debug model with FSDP=2 + TP=2 confirms this 2D parallelism is working for routed experts ([logs](https://www.internalfb.com/phabricator/paste/view/P1848969337))

## Limitations
- 2D parallel witih FDSP+TP for **shared experts** is not yet supported yet (see [comment](https://github.com/pytorch/ao/pull/2425#issuecomment-2997952768)) below). Need to debug this, which I will do in a subsequent PR.

### DIFF
--- a/test/prototype/moe_training/test_tp.py
+++ b/test/prototype/moe_training/test_tp.py
@@ -7,7 +7,7 @@
 #
 # To run these unit tests, use the following command:
 #
-# torchrun --nproc_per_node=${NUM_GPUS} -m pytest test_fsdp.py
+# torchrun --nproc_per_node=${NUM_GPUS} -m pytest test_tp.py
 #
 #######################################################################
 
@@ -18,7 +18,7 @@ import pytest
 import torch
 from torch import distributed as dist
 from torch import nn
-from torch.distributed._composable.fsdp import fully_shard
+from torch.distributed.device_mesh import DeviceMesh, init_device_mesh
 from torch.nn import functional as F
 
 # this feature requires CUDA and SM89+
@@ -34,6 +34,7 @@ from torchao.quantization.quant_api import quantize_
 
 # this test requires torchtitan
 try:
+    from torchtitan.experiments.llama4.infra.parallelize import apply_moe_tp
     from torchtitan.experiments.llama4.model.args import TransformerModelArgs
     from torchtitan.experiments.llama4.model.moe import MoE
 except ImportError:
@@ -43,25 +44,32 @@ except ImportError:
     pytest.skip(allow_module_level=True)
 
 
-def test_moe_float8_training_fsdp():
+@pytest.mark.parametrize(
+    "target_fqns",
+    [
+        ["experts"],
+        ["experts,shared_expert"],
+    ],
+)
+def test_moe_float8_training_tp_sp(target_fqns: list[str]):
     assert torch.cuda.is_available()
 
     # setup distributed for fsdp
-    setup_distributed()
+    mesh = setup_distributed()
 
     # define model args
-    target_fqns = ["experts"]
     model_args = TransformerModelArgs(
         moe_enabled=True,
         num_experts=8,
         dim=256,
+        vocab_size=1024,
     )
     init_std = 0.02
     device = torch.device("cuda")
 
     # reference bf16 MoE
     ref_model = MoE(model_args).to(torch.bfloat16).cuda()
-    torch.manual_seed(42)
+    torch.manual_seed(1)
     ref_model.init_weights(init_std, device)
 
     # target MoE for testing conversion
@@ -88,9 +96,9 @@ def test_moe_float8_training_fsdp():
         target_fqns=target_fqns,
     )
 
-    # FSDP2
-    fully_shard(model)
-    fully_shard(ref_model)
+    # apply TP
+    apply_moe_tp(model, mesh)
+    apply_moe_tp(ref_model, mesh)
 
     # inputs
     batch, seq, dim = 8, 2048, 256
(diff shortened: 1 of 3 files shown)


danielvegamyhre:
Error with FSDP=2, TP=2 targeting both routed experts AND shared expert. The issue is specific to the shared expert using 2D parallelism. Will debug and resolve in separate PR.

The logs are a bit confusing, I first see an error in meta registration that "B" tensor is fp32 instead of bf16. This is odd, since I'm not using `torch.compile` and I thought meta registrations were only used for compile.
```
File "/home/danvm/.conda/envs/torchtitan/lib/python3.13/site-packages/torch/_meta_registrations.py", line 7527, in _meta_grouped_mm_common
...
RuntimeError: Expected inputs of BF16 type but got mat_a.dtype=torch.bfloat16 and mat_b.dtype=torch.float32.
```

Then a few lines later, I see my log lines during the forward pass, just before the grouped mm, confirming the "B" tensor (W1) is bf16, not fp32:

```
[rank0]:X dtype: torch.bfloat16
[rank0]:W1 dtype: torch.bfloat16
[rank0]:W1 type: <class 'torch.distributed.tensor.DTensor'>
[rank0]:W1.to_local() type: <class 'torchao.prototype.moe_training.tensor.ScaledGroupedMMTensor'>
[rank0]:W1.to_local() dtype: torch.bfloat16
```

(as an aside, it's strange these log lines appear AFTER the error has already occured (?). I assume it must be due to how log writes are buffered)

Then at the end of the logs, I see a different error related to strides/sizes not matching a storage of size 0, but i'm guessing this a downstream affect of the first error:

```
RuntimeError: setStorage: sizes [1, 256, 256], strides [65536, 256, 1], storage offset 0, and itemsize 2 requiring a storage size of 131072 are out of bounds for storage of size 0
```

Full logs: https://www.internalfb.com/phabricator/paste/view/P1850071143

danielvegamyhre:
cc @drisspg @vkuzo for review

fyi @tianyu-l @lessw2020 @ngimel for awareness as well

test/prototype/moe_training/test_tp.py
```diff
@@ -0,0 +1,219 @@
...
+
+import pytest
+import torch
+from torch import distributed as dist
+from torch import nn
+from torch.distributed.device_mesh import DeviceMesh, init_device_mesh
+from torch.nn import functional as F
+
+# this feature requires CUDA and SM89+
+if not torch.cuda.is_available() or torch.cuda.get_device_capability() < (8, 9):
```
drisspg, line 25: nit we have some helpers for this in ao/utils

test/prototype/moe_training/test_tp.py
```diff
@@ -0,0 +1,219 @@
...
+    )
+
+from torchao.float8.float8_utils import compute_error
+from torchao.prototype.moe_training.conversion_utils import MoETrainingConfig
+from torchao.prototype.moe_training.tensor import ScaledGroupedMMTensor
+from torchao.quantization.quant_api import quantize_
+
+# this test requires torchtitan
+try:
+    from torchtitan.experiments.llama4.infra.parallelize import apply_moe_tp
```
drisspg, line 37: we should add this test to test_float8 -> https://github.com/pytorch/ao/blob/994a4ba6c869854fcaa6ca7e118fcbd75e6c28cc/.github/workflows/float8_test.yml#L46

test/prototype/moe_training/test_tp.py
```diff
@@ -0,0 +1,219 @@
...
+    for param1, param2 in zip(model.parameters(), ref_model.parameters()):
+        param_grad_sqnr = compute_error(param1.grad, param2.grad)
+        assert param_grad_sqnr.item() >= 25.0, (
+            f"SQNR must be >= 25.0, got {param_grad_sqnr.item()}."
+        )
+
+    dist.destroy_process_group()
+
+
+def _validate_model_conversion(
```
drisspg, line 143: did I review another PR that had teh same util? if so maybe put into torchao.testing so we can reuse

test/prototype/moe_training/test_tp.py
```diff
@@ -0,0 +1,219 @@
...
+    world_size = int(os.environ["WORLD_SIZE"])
+    dist.init_process_group("nccl", rank=rank, world_size=world_size)
+    device_mesh = init_device_mesh("cuda", (world_size,))
+    # seed must be the same in all processes
+    torch.manual_seed(1)
+    torch.cuda.set_device(rank)
+    return device_mesh
+
+
+def apply_moe_tp(
```
drisspg, line 184: this is always specific to module structure e.g. the fqn's right?

torchao/prototype/moe_training/conversion_utils.py
```diff
@@ -8,6 +14,8 @@
     register_quantize_module_handler,
 )
 
+logger: logging.Logger = logging.getLogger(__name__)
```
drisspg: side note, we should setup better logging in torchao
alas: https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library

just getting the root logger going w/ null handler

drisspg (approved):

danielvegamyhre:
Closing in favor of https://github.com/pytorch/ao/pull/2473 but @drisspg  I addressed your comments in the new PR. The new PR is not ready for review yet, still some work to do. Will tag you when it's ready

PR #2431, opened 2025-06-24
test rowwise fp32

y-sq:
Summary:
Running rowwise scaling on fp32 tensors got the error, P1794222725
```
RuntimeError: Only bf16 high precision output types are supported for row-wise scaling.
```

This pr adds an option to explicitly use bfloat16 as the output of rowwise_scaled, and cast it back to the original precision.

It can be enabled by setting
```
config = dataclasses.replace(config, convert_dtypes_for_rowwise_scaled_mm=True)
```

Differential Revision: D73552660

### DIFF
--- a/torchao/float8/float8_ops.py
+++ b/torchao/float8/float8_ops.py
@@ -54,6 +54,12 @@ def addmm_float8_unwrapped(
         a_inverse_scale = a_inverse_scale.new_ones(())
         b_inverse_scale = a_inverse_scale.new_ones(())
 
+    # work around torch._scaled_mm not having float32 output type
+    # TODO(pytorch/pytorch#156771): remove this once torch._scaled_mm supports float32 output
+    orig_dtype = output_dtype
+    if orig_dtype in (torch.float16, torch.float32) and is_rowwise_scaling:
+        output_dtype = torch.bfloat16
+
     post_bias = None
     if output_dtype == torch.float32:
         # Bias is not supported by _scaled_mm when output is fp32
@@ -75,6 +81,9 @@ def addmm_float8_unwrapped(
         output *= post_inverse_scale
     if post_bias is not None:
         output += post_bias
+
+    if orig_dtype in (torch.float16, torch.float32) and is_rowwise_scaling:
+        output = output.to(orig_dtype)
 
     return output
 
--- a/test/float8/test_base.py
+++ b/test/float8/test_base.py
@@ -34,9 +34,7 @@ from torchao.float8.config import (
     e5m2_dtype,
 )
 from torchao.float8.float8_linear import Float8Linear
-from torchao.float8.float8_linear_utils import (
-    convert_to_float8_training,
-)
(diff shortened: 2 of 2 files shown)


torchao/float8/config.py
```diff
@@ -204,6 +204,11 @@ class Float8LinearConfig:
     # same value in the forward pass as the backward passes.
     round_scales_to_power_of_2: bool = False
 
+    # This is a workaround for using rowwise_scaled_mm for non-bf16 tensors.
+    # Currently, rowwise_scaled_mm only supports bf16 outputs.
+    # We workaround this by using bf16 as rowwise_scaled_mm output, and cast back to the original precision.
+    convert_dtypes_for_rowwise_scaled_mm: bool = False
```
vkuzo, line 210: 1. can we prefix the name with underscore to signify an experimental flag
2. can we add file an issue to make `torch._scaled_mm` support float32 output, and add a TODO here in a comment to delete this flag once the issue is resolved

torchao/float8/float8_ops.py
```diff
@@ -54,6 +55,11 @@ def addmm_float8_unwrapped(
         a_inverse_scale = a_inverse_scale.new_ones(())
         b_inverse_scale = a_inverse_scale.new_ones(())
 
+    orig_dtype = output_dtype
+
+    if convert_dtypes_for_rowwise_scaled_mm and is_rowwise_scaling:
+        output_dtype = torch.bfloat16
+
```
vkuzo, line 62: instead of adding a flag, TBH I think we can just enable this on-by-default, like this:

file issue in PyTorch core to add float32 output to scaled_mm

```python
output_dtype_to_use = output_dtype
if is_rowwise_scaling:
    # work around torch._scaled_mm not having float32 output type
    # TODO(issue number): remove this once torch._scaled_mm supports float32 output
    output_dtype_to_use = torch.bfloat16
output = torch._scaled_mm(..., output_dtype_to_use, ...)
...
if is_rowwise_scaling and output_dtype == torch.float32:
    # work around torch._scaled_mm not having float32 output type
    # TODO(issue number): remove this once torch._scaled_mm supports float32 output
    output = output.to(orig_dtype)
```

y-sq: makes sense, I'll change to enable by default and file an issue.

y-sq:
Updated to enable the workaround by default. Included fp16 and fp32 dtypes in the existing test cases. The additional changes are formatting things generated by linter. 
(May need some time for this pr to be updated with the diff.)

The pytorch issue: https://github.com/pytorch/pytorch/issues/156771

vkuzo:
@y-sq , maybe export again?

y-sq:
@vkuzo sorry there were some un-synced files between github and fbcode so the previous export all failed. The pr should be updated now.

vkuzo (approved):
lgtm, let's make sure CI is good before landing!

PR #2432, opened 2025-06-24
solve the test issue

Gasoonjia:
Differential Revision: D77191199
torchao numeric debug infra have to rely on pytorch 2.8 nightly and beyond to trace the node transformation between exported graph and its lifted module; update the test decorator to skip the tests on lower version pytorch


jerryzh168:
nit: it would be better to put some quick summary to explain the context I think

Gasoonjia:
> nit: it would be better to put some quick summary to explain the context I think

Thanks! Added

PR #2437, opened 2025-06-24
Add support for float8 activation for Int4PreshuffledTensor

jerryzh168:
Stacked PRs:
 * #2474
 * #2463
 * #2479
 * __->__#2437

--- --- ---

### Add support for float8 activation for Int4PreshuffledTensor

Summary:
Added basic op support like linear and bmm, we have both float8 and bf16 in the same Tensor
because it's the same dtype, only difference is whether the activation is quantized or not. Although
there is some differneces in implementation:

bf16 activaton:
* group_scale
* group_zero

fp8 activation
* group_scale
* row_scale

Test Plan:
python test/quantization/quantize_/workflows/int4/test_int4_preshuffled_tensor.py

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/quantization/quant_api.py
+++ b/torchao/quantization/quant_api.py
@@ -67,8 +67,8 @@ from torchao.quantization.linear_activation_weight_observed_tensor import (
     LinearActivationWeightObservedTensor,
 )
 from torchao.quantization.observer import AffineQuantizedObserverBase, get_block_size
-from torchao.quantization.quantize_ import (
-    Int4GroupwisePreshuffleTensor,
+from torchao.quantization.quantize_.workflows import (
+    Int4PreshuffledTensor,
 )
 from torchao.quantization.transform_module import (
     _QUANTIZE_CONFIG_HANDLER,
@@ -2056,6 +2056,7 @@ class FbgemmConfig(AOBaseConfig):
        weight_dtype (torch.dtype): weight dtype of the kernel
        output_dtype (torch.dtype): output dtype of the kernel
        group_size (int): The group size for weight
+       preshuffle (bool): whether preshuffle the weights or not
     """
 
     input_dtype: torch.dtype
@@ -2073,7 +2074,7 @@ def _(module: torch.nn.Module, config: FbgemmConfig) -> torch.nn.Module:
 
     _SUPPORTED_DTYPES = {
         (torch.bfloat16, torch.int4, torch.bfloat16),
-        (e4m3_dtype, e4m3_dtype, torch.bfloat16),
+        (torch.float8_e4m3fn, torch.float8_e4m3fn, torch.bfloat16),
     }
 
     if (
@@ -2082,8 +2083,10 @@ def _(module: torch.nn.Module, config: FbgemmConfig) -> torch.nn.Module:
         and (config.output_dtype == torch.bfloat16)
     ):
         if config.preshuffle:
-            weight = Int4GroupwisePreshuffleTensor.from_float(
-                module.weight, config.block_size
+            weight = Int4PreshuffledTensor.from_float(
+                module.weight,
+                config.block_size,
+                activation_dtype=torch.bfloat16,
             )
         else:
             weight = to_fbgemm_int4(
@@ -2093,6 +2096,20 @@ def _(module: torch.nn.Module, config: FbgemmConfig) -> torch.nn.Module:
         module.weight = torch.nn.Parameter(weight, requires_grad=False)
         module.extra_repr = types.MethodType(_linear_extra_repr, module)
         return module
+    if (
+        (config.input_dtype == e4m3_dtype)
+        and (config.weight_dtype == torch.int4)
+        and (config.output_dtype == torch.bfloat16)
+    ):
+        if config.preshuffle:
+            weight = Int4PreshuffledTensor.from_float(
+                module.weight,
+                config.block_size,
+                activation_dtype=torch.float8_e4m3fn,
+            )
+            module.weight = torch.nn.Parameter(weight, requires_grad=False)
+            module.extra_repr = types.MethodType(_linear_extra_repr, module)
+            return module
     elif (
         (config.input_dtype == e4m3_dtype)
         and (config.weight_dtype == e4m3_dtype)
--- a/test/quantization/quantize_/workflows/int4/test_int4_preshuffled_tensor.py
+++ b/test/quantization/quantize_/workflows/int4/test_int4_preshuffled_tensor.py
@@ -4,14 +4,18 @@
 # This source code is licensed under the BSD 3-Clause license found in the
 # LICENSE file in the root directory of this source tree.
 
+import tempfile
 import unittest
 
 import torch
 from torch.testing._internal.common_utils import (
     TestCase,
+    instantiate_parametrized_tests,
+    parametrize,
     run_tests,
 )
 
+from torchao.float8.config import e4m3_dtype
 from torchao.quantization import (
     FbgemmConfig,
     quantize_,
@@ -23,6 +27,45 @@ from torchao.utils import (
     is_sm_at_least_90,
 )
 
+if TORCH_VERSION_AT_LEAST_2_8:
+    BF16_ACT_CONFIG = FbgemmConfig(
+        input_dtype=torch.bfloat16,
+        weight_dtype=torch.int4,
+        output_dtype=torch.bfloat16,
+        block_size=[1, 128],
+        preshuffle=True,
+    )
+
+    BF16_ACT_BMM_CONFIG = FbgemmConfig(
+        input_dtype=torch.bfloat16,
+        weight_dtype=torch.int4,
+        output_dtype=torch.bfloat16,
+        block_size=[1, 1, 128],
+        preshuffle=True,
+    )
+
+    FP8_ACT_CONFIG = FbgemmConfig(
+        input_dtype=e4m3_dtype,
+        weight_dtype=torch.int4,
+        output_dtype=torch.bfloat16,
+        block_size=[1, 128],
+        preshuffle=True,
+    )
+
+    FP8_ACT_BMM_CONFIG = FbgemmConfig(
+        input_dtype=e4m3_dtype,
+        weight_dtype=torch.int4,
+        output_dtype=torch.bfloat16,
+        block_size=[1, 1, 128],
+        preshuffle=True,
+    )
+
+else:
+    BF16_ACT_CONFIG = None
+    BF16_ACT_BMM_CONFIG = None
+    FP8_ACT_CONFIG = None
+    FP8_ACT_BMM_CONFIG = None
+
 
 @unittest.skipIf(not TORCH_VERSION_AT_LEAST_2_8, "Need pytorch 2.8+")
 @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
@@ -30,35 +73,25 @@ from torchao.utils import (
 @unittest.skipIf(
     not _is_fbgemm_genai_gpu_available(), "Requires fbgemm-gpu-genai >= 1.2.0"
 )
-class TestInt4GroupwisePreshuffleTensor(TestCase):
+class TestInt4PreshuffledTensor(TestCase):
     def setUp(self):
-        self.config = FbgemmConfig(
-            input_dtype=torch.bfloat16,
-            weight_dtype=torch.int4,
-            output_dtype=torch.bfloat16,
-            block_size=[1, 128],
-            preshuffle=True,
-        )
-        self.bmm_config = FbgemmConfig(
-            input_dtype=torch.bfloat16,
-            weight_dtype=torch.int4,
-            output_dtype=torch.bfloat16,
-            block_size=[1, 1, 128],
-            preshuffle=True,
-        )
         self.GPU_DEVICES = ["cuda"] if torch.cuda.is_available() else []
 
-    def test_linear(self):
+    @parametrize("config", [BF16_ACT_CONFIG, FP8_ACT_CONFIG])
+    def test_linear(self, config):
         dtype = torch.bfloat16
         device = "cuda"
         input = torch.randn(1, 128, dtype=dtype, device=device)
         linear = torch.nn.Linear(128, 256, dtype=dtype, device=device)
         original = linear(input)
-        quantize_(linear, self.config)
+        quantize_(linear, config)
         quantized = linear(input)
         self.assertTrue(compute_error(original, quantized) > 20)
 
-    def test_bmm(self):
+    # Note: this order will error out: `Got bad cuda status: an illegal memory access was encountered at line: 449`
+    # @parametrize("bmm_config", [BF16_ACT_BMM_CONFIG, FP8_ACT_BMM_CONFIG])
+    @parametrize("bmm_config", [FP8_ACT_BMM_CONFIG, BF16_ACT_BMM_CONFIG])
+    def test_bmm(self, bmm_config):
         class M(torch.nn.Module):
             def __init__(self, weight):
                 super().__init__()
@@ -74,31 +107,45 @@ class TestInt4GroupwisePreshuffleTensor(TestCase):
         m = M(weight).eval()
         original = m(input)
         m.weight = torch.nn.Parameter(m.weight.transpose(1, 2).contiguous())
-        quantize_(m, self.bmm_config, filter_fn=lambda x, fqn: True)
+        quantize_(m, bmm_config, filter_fn=lambda x, fqn: True)
         quantized = m(input)
         self.assertTrue(compute_error(original, quantized) > 18)
 
-    def test_to_device(self):
+    @parametrize("config", [BF16_ACT_CONFIG, FP8_ACT_CONFIG])
+    def test_to_device(self, config):
         for device in self.GPU_DEVICES:
             linear = torch.nn.Linear(128, 256, dtype=torch.bfloat16)
-            quantize_(linear, self.config)
+            quantize_(linear, config)
             linear.to(device)
 
             linear = torch.nn.Linear(128, 256, dtype=torch.bfloat16)
-            quantize_(linear, self.config)
+            quantize_(linear, config)
             linear.to(device=device)
 
             linear = torch.nn.Linear(128, 256, dtype=torch.bfloat16)
-            quantize_(linear, self.config)
+            quantize_(linear, config)
             linear.to(device)
 
-    def test_module_path(self):
+    @parametrize("config", [BF16_ACT_CONFIG, FP8_ACT_CONFIG])
+    def test_module_path(self, config):
         linear = torch.nn.Linear(128, 256, dtype=torch.bfloat16)
-        quantize_(linear, self.config)
+        quantize_(linear, config)
         self.assertEqual(
             str(type(linear.weight)),
-            "<class 'torchao.quantization.Int4GroupwisePreshuffleTensor'>",
+            "<class 'torchao.quantization.Int4PreshuffledTensor'>",
         )
+
+        with tempfile.NamedTemporaryFile() as f:
+            torch.save(linear.state_dict(), f)
+            f.seek(0)
+            state_dict = torch.load(f)
+            self.assertEqual(
+                str(type(state_dict["weight"])),
+                "<class 'torchao.quantization.Int4PreshuffledTensor'>",
+            )
+
+
+instantiate_parametrized_tests(TestInt4PreshuffledTensor)
 
 
 if __name__ == "__main__":
--- a/torchao/quantization/__init__.py
+++ b/torchao/quantization/__init__.py
@@ -87,8 +87,8 @@ from .quant_primitives import (
     dequantize_affine,
     quantize_affine,
 )
-from .quantize_ import (
-    Int4GroupwisePreshuffleTensor,
+from .quantize_.workflows import (
+    Int4PreshuffledTensor,
 )
 from .smoothquant import (
     SmoothFakeDynamicallyQuantizedLinear,
@@ -153,7 +153,7 @@ __all__ = [
     "ModuleFqnToConfig",
     "FbgemmConfig",
     # tensor subclasses
-    "Int4GroupwisePreshuffleTensor",
+    "Int4PreshuffledTensor",
     # smooth quant - subject to change
     "get_scale",
     "SmoothFakeDynQuantMixin",
--- a/torchao/quantization/quantize_/workflows/__init__.py
+++ b/torchao/quantization/quantize_/workflows/__init__.py
@@ -1,0 +1,7 @@
+from .int4.int4_preshuffled_tensor import (
+    Int4PreshuffledTensor,
+)
+
+__all__ = [
+    "Int4PreshuffledTensor",
+]
--- a/torchao/quantization/quantize_/workflows/int4/int4_preshuffled_tensor.py
+++ b/torchao/quantization/quantize_/workflows/int4/int4_preshuffled_tensor.py
@@ -6,7 +6,7 @@
 
 
 import importlib.util
-from typing import List
+from typing import List, Optional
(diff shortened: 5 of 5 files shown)


test/dtypes/test_float8_activation_int4_groupwise_preshuffle.py
```diff
@@ -0,0 +1,166 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
vkuzo: should this test be in the `test/quantization/quantize_` folder?

test/dtypes/test_float8_activation_int4_groupwise_preshuffle.py
```diff
@@ -0,0 +1,166 @@
...
+    is_sm_at_least_90,
+)
+
+
+@unittest.skipIf(not TORCH_VERSION_AT_LEAST_2_8, "Need pytorch 2.8+")
+@unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
+@unittest.skipIf(not is_sm_at_least_90(), "Nedd sm90+")
+class TestInt4GroupwisePreshuffleTensor(TestCase):
+    def setUp(self):
+        self.config = FbgemmConfig(
```
vkuzo: or, maybe organize tests by the config?  it would make sense for tests for everything in `FbgemmConfig` to be together.

test/dtypes/test_float8_activation_int4_groupwise_preshuffle.py
```diff
@@ -0,0 +1,166 @@
...
+        dtype = torch.bfloat16
+        device = "cuda"
+        input = torch.randn(1, 128, dtype=dtype, device=device)
+        linear = torch.nn.Linear(128, 256, dtype=dtype, device=device)
+        original = linear(input)
+        quantize_(linear, self.config)
+        quantized = linear(input)
+        self.assertTrue(compute_error(original, quantized) > 20)
+
+    # @unittest.skip("WIP: this doesn't work yet")
```
vkuzo: remove

test/dtypes/test_float8_activation_int4_groupwise_preshuffle.py
```diff
@@ -0,0 +1,166 @@
...
+    def test_slice(self):
```
vkuzo: is there an integration test we can write to cover this? If this is needed for TP, maybe just have an integration test for TP which is easy to run for all of these configs?

test/dtypes/test_float8_activation_int4_groupwise_preshuffle.py
```diff
@@ -0,0 +1,166 @@
...
+        quantize_(dummy_l, self.config)
+        quantized = dummy_l.weight
+        quantized = quantized.narrow(0, 0, 512)
+
+        param_data.copy_(quantized)
+
+        # making sure param.data is updated
+        assert param.data.packed_weight[0][0] != orig_value
+
+    def test_bmm(self):
```
vkuzo: unit tests are nice, feels like it would be good to also have an integration test to cover all of these ops in one go

torchao/quantization/quant_api.py
```diff
@@ -2040,6 +2040,8 @@ class FbgemmConfig(AOBaseConfig):
        weight_dtype (torch.dtype): weight dtype of the kernel
        output_dtype (torch.dtype): output dtype of the kernel
        group_size (int): The group size for weight
+       preshuffle (bool): whether preshuffle the weights or not
+       activation_dtype_for_int4 (str): the dtype for activation for int4 weight, either bf16 or fp8
```
vkuzo: it's confusing to have both `input_dtype` and `activation_dtype_for_int4`, what are your thoughts

torchao/dtypes/float8_activation_int4_groupwise_preshuffle_tensor.py
```diff
@@ -0,0 +1,447 @@
...
+    quantize_int4_preshuffle = None
+    quantize_fp8_row = None
+else:
+    from fbgemm_gpu.experimental.gen_ai.quantize import (
+        quantize_fp8_row,
+        quantize_int4_preshuffle,
+    )
+
+
+class Float8ActivationInt4GroupwisePreshuffleTensor(TorchAOBaseTensor):
```
vkuzo: why not extend the existing Int4GroupwisePreshuffleTensor

jerryzh168: sorry this should be removed

test/dtypes/test_float8_activation_int4_groupwise_preshuffle.py
jerryzh168: yeah, the test file to review is test_int4_groupwise_preshuffle.py, I forgot to remove this file

test/dtypes/test_float8_activation_int4_groupwise_preshuffle.py
jerryzh168: e2e tests can be the ones we run in vllm/sglang integration on real models I think?

test/dtypes/test_float8_activation_int4_groupwise_preshuffle.py
jerryzh168: this is required to run the model in vllm, can add to https://github.com/pytorch/ao/blob/main/test/integration/test_vllm.py when the API is more mature

test/dtypes/test_float8_activation_int4_groupwise_preshuffle.py
jerryzh168: was planning to remove FbgemmConfig as well soon

torchao/quantization/quant_api.py
jerryzh168: yeah it's a bit confusing, although this is temporary, I'm deprecating this later in the stack: https://github.com/pytorch/ao/pull/2474

I'll remove this one and just use `input_dtype`

test/integration/test_serialization_bc.py
jerryzh168, line 1: @vkuzo added model serialization bc test here

test/quantization/quantize_/int4/test_int4_groupwise_preshuffle_tensor.py
```diff
@@ -23,6 +27,45 @@
...
+    FP8_ACT_CONFIG = FbgemmConfig(
+        input_dtype=e4m3_dtype,
+        weight_dtype=torch.int4,
+        output_dtype=torch.bfloat16,
+        block_size=[1, 128],
+        preshuffle=True,
+    )
+
+    FP8_ACT_BMM_CONFIG = FbgemmConfig(
+        input_dtype=e4m3_dtype,
```
drisspg: Is this consistent w/ our other APIs? I thought its call `activation_dytpe`: https://github.com/pytorch/ao/blob/e9644da9208bdae48ff6ff485be30de6ca0e779a/torchao/quantization/quant_api.py#L1593

test/quantization/quantize_/int4/test_int4_groupwise_preshuffle_tensor.py
```diff
@@ -32,33 +75,23 @@
...
         linear = torch.nn.Linear(128, 256, dtype=dtype, device=device)
         original = linear(input)
-        quantize_(linear, self.config)
+        quantize_(linear, config)
         quantized = linear(input)
         self.assertTrue(compute_error(original, quantized) > 20)
 
-    def test_bmm(self):
+    # Note: this order will error out: `Got bad cuda status: an illegal memory access was encountered at line: 449`
+    # @parametrize("bmm_config", [BF16_ACT_BMM_CONFIG, FP8_ACT_BMM_CONFIG])
```
drisspg: that a pretty weird..

torchao/quantization/quant_api.py
```diff
@@ -2067,7 +2068,9 @@ def _(module: torch.nn.Module, config: FbgemmConfig) -> torch.nn.Module:
     ):
         if config.preshuffle:
             weight = Int4GroupwisePreshuffleTensor.from_float(
-                module.weight, config.block_size
+                module.weight,
+                config.block_size,
+                activation_dtype="bf16",
```
drisspg: called activation_dtype here

test/quantization/quantize_/int4/test_int4_groupwise_preshuffle_tensor.py
jerryzh168: probably not, we are removing this config, see the last PR in the stack

torchao/quantization/quantize_/int4/int4_groupwise_preshuffle_tensor.py
```diff
@@ -36,10 +40,16 @@ class Int4GroupwisePreshuffleTensor(TorchAOBaseTensor):
 
     Tensor Attributes:
         packed_weight: packed int4 weight, either 2D (N, K/2) or 3D (B, N, K/2), last dimension is packed
-        group_scale: (K/group_size, N) for 2D Tensor, (B, N, K/group_size) for 3D Tensor
-               dtype is the same as the original Tensor dtype
-        group_zero: (K/group_size, N) for 2D Tensor, (B, N, K/group_size) for 3D Tensor
-               dtype is the same as the original Tensor dtype
+        for bf16 activation:
```
drisspg: are there expectations on strides for these two different versions?

torchao/quantization/quantize_/int4/int4_groupwise_preshuffle_tensor.py
```diff
@@ -61,13 +70,26 @@ class Int4GroupwisePreshuffleTensor(TorchAOBaseTensor):
...
       perform a single 64 bit load to move all needed weights for the thread into register memory.
+
+    Note for float8 activation int4 weight kernel:
+      float8 activation int4 weight kernel doesn't work with zero_point, since it use table lookup approach which
+      requires symmetric quantization
     """
 
-    tensor_data_attrs = ["packed_weight", "group_scale", "group_zero"]
+    tensor_data_attrs = ["packed_weight", "group_scale"]
+    optional_tensor_data_attr1 = "group_zero"
```
drisspg: is there a reason why these can't always be None?

test/quantization/quantize_/int4/test_int4_groupwise_preshuffle_tensor.py
jerryzh168: yeah Josh says this is because some tensors are on CPU

torchao/quantization/quantize_/int4/int4_groupwise_preshuffle_tensor.py
jerryzh168: I think all of these should be contiguous

torchao/quantization/quantize_/int4/int4_groupwise_preshuffle_tensor.py
```diff
@@ -77,36 +99,53 @@ def __new__(
...
         self.group_zero = group_zero
+        self.row_scale = row_scale
         self.shape_multiplier = shape_multiplier
         self.group_size = group_size
 
     def __tensor_flatten__(self):
-        return self.tensor_data_attrs, [
-            getattr(self, attr) for attr in self.tensor_attributes
-        ]
+        if getattr(self, self.optional_tensor_data_attr1) is None:
```
drisspg: this smells weird

torchao/quantization/quantize_/int4/int4_groupwise_preshuffle_tensor.py
jerryzh168: not sure, the non-preshuffle one does return a zero_point, but I can double check with Josh

torchao/quantization/quantize_/int4/int4_groupwise_preshuffle_tensor.py
```diff
@@ -77,36 +99,53 @@ def __new__(
...
+        tensors.append(tensor_data_dict.get(cls.optional_tensor_data_attr2, None))
         return cls(
-            *[tensor_data_dict[name] for name in cls.tensor_data_attrs],
+            *tensors,
             *tensor_attributes,
         )
 
     def _apply_fn_to_data(self, fn):
+        tensors = [fn(getattr(self, name)) for name in self.tensor_data_attrs]
+        t1 = getattr(self, self.optional_tensor_data_attr1)
```
drisspg: seems like they should also go in a list?

torchao/quantization/quantize_/int4/int4_groupwise_preshuffle_tensor.py
```diff
@@ -137,10 +177,17 @@ def from_float(
         cls,
         w: torch.Tensor,
         block_size: List[int],
+        activation_dtype: str = "bf16",
```
drisspg: why a string?

torchao/quantization/quantize_/int4/int4_groupwise_preshuffle_tensor.py
jerryzh168: it can be a list of exactly two elements, that if one exists the other will not exist. it's a bit weird anyways, I'm open to just hardcode these as well, which one do you feel is better?

torchao/quantization/quantize_/int4/int4_groupwise_preshuffle_tensor.py
jerryzh168: this is a string in the fbgemm op API

jerryzh168:
> Can you add serialization tests

you mean serialization of the models? it's moved to https://github.com/pytorch/ao/pull/2463/files#diff-9f6b6c4b39656e797cfda97536a4cf8a82004c64da518ad524637b471b716739, I don't exactly remember the reason 

for config serialization we did a config refactor in the last PR, I can add after we are aligned on what config should look like

torchao/quantization/quantize_/int4/int4_groupwise_preshuffle_tensor.py
vkuzo: I think we should be consistent with other torchao APIs here instead of trying to match fbgemm

torchao/quantization/quantize_/int4/__init__.py
```diff
@@ -0,0 +1,7 @@
+from .int4_groupwise_preshuffle_tensor import (
```
vkuzo: can we have some kind of `workflows` folder? `torchao/quantization/quantize_/workflows/int4/...`

jerryzh168: sure, but just trying to understand, what are the other folders under quantize_ that are not workflows?

torchao/quantization/quantize_/int4/int4_groupwise_preshuffle_tensor.py
jerryzh168: from Josh: it definitely helps a bit and doesnt hurt performance noticeably

I don't think we want to sacrifice accuracy for the sake of uniformity of code here?

torchao/quantization/quantize_/int4/int4_groupwise_preshuffle_tensor.py
jerryzh168: updated to hardcoded names

torchao/quantization/quant_api.py
```diff
@@ -2040,6 +2040,7 @@ class FbgemmConfig(AOBaseConfig):
        weight_dtype (torch.dtype): weight dtype of the kernel
        output_dtype (torch.dtype): output dtype of the kernel
        group_size (int): The group size for weight
```
andrewor14, line 2058: seems like this field doesn't exist anymore, and `block_size` and `activation_scale_ub` are not documented

torchao/quantization/quant_api.py
```diff
@@ -2040,6 +2040,7 @@ class FbgemmConfig(AOBaseConfig):
...
+       preshuffle (bool): whether preshuffle the weights or not
```
andrewor14, line 2059: add default value = False here?

torchao/quantization/quant_api.py
```diff
@@ -2077,6 +2080,20 @@ def _(module: torch.nn.Module, config: FbgemmConfig) -> torch.nn.Module:
...
+        and (config.output_dtype == torch.bfloat16)
+    ):
+        if config.preshuffle:
+            weight = Int4GroupwisePreshuffleTensor.from_float(
+                module.weight,
+                config.block_size,
+                activation_dtype=torch.float8_e4m3fn,
+            )
+            module.weight = torch.nn.Parameter(weight, requires_grad=False)
+            module.extra_repr = types.MethodType(_linear_extra_repr, module)
```
andrewor14, line 2111: nit: these two lines are duplicated 3 times, can refactor?

torchao/quantization/quantize_/int4/int4_groupwise_preshuffle_tensor.py
```diff
@@ -36,10 +40,16 @@ class Int4GroupwisePreshuffleTensor(TorchAOBaseTensor):
...
-               dtype is the same as the original Tensor dtype
-        group_zero: (K/group_size, N) for 2D Tensor, (B, N, K/group_size) for 3D Tensor
-               dtype is the same as the original Tensor dtype
+        for bf16 activation:
+            group_scale: (K/group_size, N) for 2D Tensor, (B, N, K/group_size) for 3D Tensor
+                   dtype is the same as the original Tensor dtype
+            group_zero: (K/group_size, N) for 2D Tensor, (B, N, K/group_size) for 3D Tensor
+                   dtype is the same as the original Tensor dtype
+        for float8 activation:
+            group_scale: (K/group_size/8, 8, N) for 2D Tensor, (B, K/group_size/8, 8, N) for 3D Tensor
```
andrewor14: why the extra /8? Also what's the second dimension?

torchao/quantization/quantize_/int4/int4_groupwise_preshuffle_tensor.py
```diff
@@ -77,36 +97,52 @@ def __new__(
...
-        packed_weight,
-        group_scale,
-        group_zero,
-        group_size,
-        shape_multiplier,
-        shape,
+        packed_weight: torch.Tensor,
+        group_scale: torch.Tensor,
+        group_zero: Optional[torch.Tensor],
+        row_scale: Optional[torch.Tensor],
```
andrewor14: Seems like we expect exactly one of these two fields to be defined. Can we error if neither or both are defined?

torchao/quantization/quantize_/int4/int4_groupwise_preshuffle_tensor.py
```diff
@@ -77,36 +97,52 @@ def __new__(
...
         self.shape_multiplier = shape_multiplier
         self.group_size = group_size
 
     def __tensor_flatten__(self):
-        return self.tensor_data_attrs, [
-            getattr(self, attr) for attr in self.tensor_attributes
-        ]
+        if getattr(self, "group_zero") is None:
+            assert getattr(self, "row_scale") is not None
+            return self.tensor_data_attrs + ["row_scale"], [
```
andrewor14: I think to simplify this we can just set `tensor_data_attrs` properly in the constructor, depending on which one of `row_scale` or `group_zero` is defined

torchao/quantization/quantize_/int4/int4_groupwise_preshuffle_tensor.py
```diff
@@ -77,36 +97,52 @@ def __new__(
...
+            *tensors,
             *tensor_attributes,
         )
 
     def _apply_fn_to_data(self, fn):
+        tensors = [fn(getattr(self, name)) for name in self.tensor_data_attrs]
+        t1 = getattr(self, "group_zero")
+        tensors.append(fn(t1) if t1 is not None else None)
+        t2 = getattr(self, "row_scale")
+        tensors.append(fn(t2) if t2 is not None else None)
```
andrewor14: same here, you can delete these 4 lines if you set `tensor_data_attrs` in the constructor

torchao/quantization/quantize_/int4/int4_groupwise_preshuffle_tensor.py
```diff
@@ -137,29 +174,52 @@ def from_float(
...
         assert len(block_size) == w.ndim, (
             f"Expecting the length of block_size to be equal to the dimension of the weight, got {block_size=} and {w.ndim=}"
         )
+
+        _SUPPORTED_ACT_DTYPES = [torch.bfloat16, torch.float8_e4m3fn]
+        assert activation_dtype in _SUPPORTED_ACT_DTYPES, (
+            f"activation dtype {activation_dtype} is not supported, supported ones are: {_SUPPORTED_ACT_DTYPES}"
+        )
+
+        _DTYPE_TO_STR = {
```
andrewor14: nit: you can just make this `_SUPPORTED_ACT_DTYPES_TO_STR` so you don't need another list above

torchao/quantization/quantize_/int4/int4_groupwise_preshuffle_tensor.py
```diff
@@ -189,21 +250,42 @@ def _(func, types, args, kwargs):
...
-        B, M, _ = input_tensor.shape
-        _, N, _ = wq.shape
-        res = torch.empty((B, M, N), device=input_tensor.device, dtype=torch.bfloat16)
-        for i in range(B):
-            res[i] = torch.ops.fbgemm.bf16i4bf16_shuffled(
-                input_tensor[i], wq[i], group_scale[i], group_zero[i]
+    # bf16 activation
+    if weight_tensor.group_zero is not None:
+        group_zero = weight_tensor.group_zero.contiguous()
+        if input_tensor.ndim == 3 and wq.ndim == 3:
```
andrewor14: is this running gemm by batch? Maybe add a comment to explain?

torchao/quantization/quant_api.py
jerryzh168: we are going to remove this config soon in https://github.com/pytorch/ao/pull/2474

torchao/quantization/quantize_/int4/int4_groupwise_preshuffle_tensor.py
jerryzh168: this seems to be the same as tinygemm, it's probably the tiling size, not exactly clear about the details though

torchao/quantization/quantize_/int4/int4_groupwise_preshuffle_tensor.py
jerryzh168: OK, sounds good

torchao/quantization/quantize_/int4/int4_groupwise_preshuffle_tensor.py
jerryzh168: actually after second thought, it can't be deleted, we still need an explicit None in cases when the attribute is None

andrewor14: hmm yeah you're right... alternatively you could do
```
tensor_data_attrs = ["packed_weight", "group_scale", "group_zero", "row_scale"]
```
then here you can do
```
tensors = [getattr(self, name) for name in self.tensor_data_attrs]
tensors = [fn(t) for t in tensors if t is not None else None]
```
but it's not much simpler than what you have

jerryzh168: yeah there is also the other functions that use this

torchao/quantization/quantize_/int4/int4_groupwise_preshuffle_tensor.py
jerryzh168: yeah this was copied over from some internal fbgemm code, now I thought about this again, we probably should remove this, since linear will only have 2d weight

PR #2438, opened 2025-06-24
Store NVFP4 block scales in swwizzled layout on tensor

drisspg:
Stacked PRs:
 * #2439
 * __->__#2438

--- --- ---

Store NVFP4 block scales in swizzled layout on tensor

For llama3 70b no TP sizes w/ 1024 tokens: 15% E2E speedup

## In eager Before: https://fburl.com/7w3j6b1q
nvfp4 
Screenshot 2025-06-24 at 4 02 42 PM
Runtime: 2436.98 μs per iteration

## In eager After: https://fburl.com/s7ggvm94
nvfp4 Runtime: 2356.77 μs per iteration
Screenshot 2025-06-24 at 4 03 27 PM

# In compile

## Before:  https://fburl.com/1gvfjjlu
nvfp4 Runtime: 576.14 μs per iteration
Screenshot 2025-06-24 at 4 11 36 PM

## After: https://fburl.com/usp1xelj
nvfp4 Runtime: 486.69 μs per iteration
Screenshot 2025-06-24 at 4 11 55 PM

```
Throughput: 47.12 requests/s, 19998.00 total tokens/s, 9635.87 output tokens/s
Total num prompt tokens:  225190
Total num output tokens:  209407
```

### DIFF
--- a/test/prototype/mx_formats/test_mx_tensor.py
+++ b/test/prototype/mx_formats/test_mx_tensor.py
@@ -657,3 +657,301 @@ def test_nvfp4_reconstruction(dtype, shape, use_per_tensor_scale):
     assert x.t().dtype == x_reconstructed_t.dtype, (
         f"Transpose dtype mismatch: {x.t().dtype} vs {x_reconstructed_t.dtype}"
     )
+
+
+@pytest.mark.parametrize(
+    "shape",
+    [
+        (128, 4),
+        (256, 8),
+        (100, 3),
+        (4, 4),
+        (50, 10),
+        (384, 12),
+    ],
+)
+@pytest.mark.parametrize(
+    "use_triton_kernel", [False, True] if torch.cuda.is_available() else [False]
+)
+@pytest.mark.skipif(
+    not TORCH_VERSION_AT_LEAST_2_8, reason="torch.compile requires PyTorch 2.8+"
+)
+def test_to_blocked_from_blocked_roundtrip(shape, use_triton_kernel: bool):
+    from torchao.prototype.mx_formats.utils import from_blocked, to_blocked
+
+    rows, cols = shape
+    device = "cuda" if torch.cuda.is_available() else "cpu"
+
+    original = torch.randint(0, 255, (rows, cols), device=device, dtype=torch.uint8)
+
+    blocked = to_blocked(original, use_triton_kernel=use_triton_kernel)
+    reconstructed = from_blocked(blocked, rows, cols)
+
+    torch.testing.assert_close(
+        original,
+        reconstructed,
+        atol=0.0,
+        rtol=0.0,
+        msg=f"Roundtrip failed for shape {shape} with use_triton_kernel={use_triton_kernel}",
+    )
+
+
+@pytest.mark.parametrize("is_swizzled_scales", [False, True])
+@pytest.mark.parametrize(
+    "shape",
+    [
+        (32, 64),
+        (16, 32),
+        (64, 128),
+        (384, 128),
+    ],
+)
+@pytest.mark.skipif(
+    not TORCH_VERSION_AT_LEAST_2_8, reason="torch.compile requires PyTorch 2.8+"
+)
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA not available")
+def test_nvfp4_swizzled_scales_construction(is_swizzled_scales, shape):
+    """
+    Test that NVFP4Tensor can be constructed with swizzled scales and
+    that the _is_swizzled_scales flag is set correctly.
+    """
+    from torchao.prototype.mx_formats.nvfp4_tensor import NVFP4Tensor
+
+    M, K = shape
+    data = torch.randn(M, K, device="cuda", dtype=torch.bfloat16)
+
+    tensor = NVFP4Tensor.to_nvfp4(data, is_swizzled_scales=is_swizzled_scales)
+    assert tensor._is_swizzled_scales == is_swizzled_scales
+    reconstructed = tensor.to_dtype(torch.bfloat16)
+    assert reconstructed.shape == data.shape
+
+
+@pytest.mark.parametrize(
+    "slice_dim,slice_spec",
+    [
+        # Row slicing - must align with 128-row boundaries
+        pytest.param(0, slice(0, 128), id="slice_rows[0:128]"),
+        pytest.param(0, slice(128, 256), id="slice_rows[128:256]"),
+        # Column slicing - must align with 64-column boundaries (4 scale columns * 16 block_size)
+        pytest.param(1, slice(0, 64), id="slice_cols[0:64]"),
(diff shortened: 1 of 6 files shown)


drisspg:
@vkuzo I plan to refactor the test from test_mx_tensor into a different file in the same folder after these two land to avoid rebase conflicts

drisspg:
Yup this is for inference

```Py

import torch
from typing import Literal

from transformer_nuggets.misc.mlp import FeedForward
from transformer_nuggets.utils.benchmark import (
    profiler,
    benchmark_do_bench_in_microseconds,
    benchmark_cuda_function_in_microseconds,
)

from torchao.prototype.mx_formats.mx_subclass import (
    MXFPInferenceConfig,
    NVFP4InferenceConfig,
)
from torchao.prototype.mx_formats import MXGemmKernelChoice
import os
from rich import print

from jsonargparse import CLI

torch._logging.set_logs(fusion=True)
torch._logging.set_logs(graph=True)
os.environ["TORCHINDUCTOR_FORCE_DISABLE_CACHES"] = "1"


def main(
    num_tokens: int = 1024,
    flavor: str = "8B",
    mode: str = "mxfp8",
    bf16: bool = False,
    setting: Literal["max-autotune", "inductor"] = "max-autotune",
    backend: Literal["inductor", "eager"] = "inductor",
    use_triton: bool = False,
):
    device = "cuda"
    dtype = torch.bfloat16
    assert flavor in ("8B", "70B", "405B")
    assert mode in ("mxfp8", "mxfp4", "nvfp4")
    if bf16:
        model = FeedForward.llama3_mlp(flavor).to(device=device, dtype=dtype)
        inpt = model.get_input(num_tokens, device, dtype)
        model = torch.compile(
            model, fullgraph=True, dynamic=False, mode=setting, backend=backend
        )

        for _ in range(5):
            model(inpt)
        with profiler("data/bf16", with_stack=True) as p:
            for _ in range(5):
                model(inpt)
                p.step()
        runtime_us = benchmark_cuda_function_in_microseconds(lambda: model(inpt))
        print(
            f"[bold green]Bf16 Runtime:[/bold green] {runtime_us:.2f} μs per iteration"
        )

    model = FeedForward.llama3_mlp(flavor).to(device=device, dtype=dtype)
    inpt = model.get_input(num_tokens, device, dtype)
    match mode:
        case "mxfp8":
            elem_dtype = torch.float8_e4m3fn
            choice = MXGemmKernelChoice.CUBLAS
            config = MXFPInferenceConfig(
                activation_dtype=elem_dtype,
                weight_dtype=elem_dtype,
                block_size=32,
                gemm_kernel_choice=choice,
            )
        case "mxfp4":
            elem_dtype = torch.float4_e2m1fn_x2
            choice = MXGemmKernelChoice.CUTLASS
            config = MXFPInferenceConfig(
                activation_dtype=elem_dtype,
                weight_dtype=elem_dtype,
                block_size=32,
                gemm_kernel_choice=choice,
            )
        case "nvfp4":
            config = NVFP4InferenceConfig(use_triton_kernel=use_triton)
        case _:
            raise ValueError(f"Unknown mode {mode}")

    quantize_(model, config)
    model = torch.compile(
        model, fullgraph=True, dynamic=False, mode="max-autotune", backend=backend
    )

    for _ in range(5):
        model(inpt)

    # Use profiler for detailed profiling
    with profiler(f"data/mx_fp_8_{"subclass"}_{mode}", with_stack=True) as p:
        for _ in range(5):
            model(inpt)
            p.step()

    # Use benchmark function for pretty printed runtime
    runtime_us = benchmark_do_bench_in_microseconds(lambda: model(inpt))
    print(f"[bold green]{mode} Runtime:[/bold green] {runtime_us:.2f} μs per iteration")


if __name__ == "__main__":
    CLI(main)

```

> training wouldn't benefit from this, at least without fused quantize+scale_swizzle
Yup this is a proxy for inference where weights are static,  like you said for training since the scales change at every grad step, I dont think this would be much help

test/prototype/mx_formats/test_mx_tensor.py
```diff
@@ -657,3 +657,303 @@ def assert_sqnr_gt_threshold(orig, new, threshold):
...
+
+    original = torch.randint(0, 255, (rows, cols), device=device, dtype=torch.uint8)
+
+    blocked = to_blocked(original, use_triton_kernel=use_triton_kernel)
+    reconstructed = from_blocked(blocked, rows, cols)
+
+    torch.testing.assert_close(
+        original,
+        reconstructed,
+        atol=1e-6,
```
vkuzo: should these tolerances be 0.0?

test/prototype/mx_formats/test_mx_tensor.py
```diff
@@ -657,3 +657,303 @@ def assert_sqnr_gt_threshold(orig, new, threshold):
...
+    M, K = 32, 64
+    data = torch.randn(M, K, device="cuda", dtype=torch.bfloat16)
+
+    # Create tensors with both storage methods
+    regular_tensor = NVFP4Tensor.to_nvfp4(data, store_swizzled=False)
+    swizzled_tensor = NVFP4Tensor.to_nvfp4(data, store_swizzled=True)
+
+    # Get scales from both tensors
+    regular_scales = regular_tensor.get_hp_scales()
+    swizzled_scales = swizzled_tensor.get_hp_scales()
```
vkuzo, line 951: this is confusing, I would have expected something like this:

```python
get_hp_scales: always returns unswizzled scales (fast if stored unswizzled)
get_swizzled_scales: always returns swizzled scales (fast if stored swizzled)
```

it's definitely a gotcha to have `get_hp_scales` return different things based on internal state

torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -94,6 +99,7 @@ def __new__(
         )
 
         self._scale_e4m3 = blockwise_scales
+        self._swizzled_scales = store_swizzled
```
vkuzo: nit: something like `_is_swizzled_scales` to make it clear this is a boolean, `swizzled_scales` sounds like a tensor

torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -77,12 +80,14 @@ def __new__(
         block_size,
         orig_dtype,
         mm_config=NVFP4MMConfig.DYNAMIC,
+        store_swizzled=False,
```
vkuzo: nit: match this arg name to attribute name? It's always a bit confusing when they are differently named, and in this case seems like there is no benefit to different names

test/prototype/mx_formats/test_mx_tensor.py
vkuzo: it looks like `get_hp_scales` does always return unswizzled scales, so should the check below be an exact match?

PR #2443, opened 2025-06-26
Add exportable coreml codebook quantization op

jerryzh168:
Summary:
Added CoreML codebook quant (Palettization): https://apple.github.io/coremltools/docs-guides/source/opt-palettization-overview.html#palettization-overview
* supports group_size `per_grouped_channel`
* doesn't support vector quantization yet, but will be easy to turn on if needed
* ops added: choose_qparams_and_quantize_codebook, dequantize_codebook
* also enabled support for export, these two ops will be preserved after exporta
* Added CodebookWeightOnlyConfig(dtype, group_size) that can be used with quantize_ to quantize the Tensor

Test Plan:
python test/prototype/test_coreml_codebook.py

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/prototype/test_codebook_coreml.py
+++ b/test/prototype/test_codebook_coreml.py
@@ -1,0 +1,91 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+import unittest
+
+import torch
+
+from torchao.prototype.quantization.codebook_coreml import (
+    CodebookQuantizedTensor,
+    CodebookWeightOnlyConfig,
+    choose_qparams_and_quantize_codebook_coreml,
+)
+from torchao.quantization import quantize_
+from torchao.quantization.utils import compute_error
+from torchao.testing.utils import skip_if_no_cuda
+from torchao.utils import TORCH_VERSION_AT_LEAST_2_6, is_package_at_least
+
+
+@unittest.skipIf(
+    not is_package_at_least("coremltools", "8.3.0"), "Requires coremltools >= 8.3.0"
+)
+class TestCodebookQuantization(unittest.TestCase):
+    def setUp(self):
+        torch.manual_seed(123)
+        self.input = torch.randn(100, 256, dtype=torch.float32)
+        self.code_dtype = torch.uint8
+        self.block_size = [-1, 4]
+        self.nbits = 8
+
+    def test_choose_qparams_codebook(self):
+        codebook, wq = choose_qparams_and_quantize_codebook_coreml(
+            self.input,
+            self.code_dtype,
+            self.block_size,
+        )
+        group_size = self.block_size[-1]
+        self.assertEqual(codebook.shape, (256 // group_size, 2**self.nbits, 1))
+        self.assertEqual(wq.shape, (100, 256))
+
+        self.assertFalse(torch.isnan(codebook).any())
+        self.assertFalse(torch.isnan(wq).any())
+
+    def test_codebook_quantized_tensor_from_float(self):
+        cqt = CodebookQuantizedTensor.from_float(
+            self.input,
+            self.code_dtype,
+            self.block_size,
+        )
+
+        dequant = cqt.dequantize()
+        sqnr = compute_error(dequant, self.input)
+        self.assertGreater(sqnr, 30)
+
+    def test_codebook_quantized_tensor_from_float2(self):
+        block_size = [-1, 16]
+        code_dtype = torch.uint4
+
+        cqt = CodebookQuantizedTensor.from_float(
+            self.input,
+            code_dtype,
+            block_size,
+        )
+
+        dequant = cqt.dequantize()
+
+        sqnr = compute_error(dequant, self.input)
+        self.assertGreater(sqnr, 18)
+
+    def test_quantize_api(self):
+        m = torch.nn.Sequential(torch.nn.Linear(64, 64))
+        quantize_(
+            m,
+            CodebookWeightOnlyConfig(dtype=self.code_dtype, block_size=self.block_size),
+        )
+        assert type(m[0].weight) == CodebookQuantizedTensor
+
+    @skip_if_no_cuda()
+    @unittest.skipIf(not TORCH_VERSION_AT_LEAST_2_6, "requires 2.6+.")
+    def test_export(self):
+        m = torch.nn.Sequential(torch.nn.Linear(128, 64)).to(torch.float32)
+        quantize_(m, CodebookWeightOnlyConfig(self.code_dtype, self.block_size))
+        example_inputs = (torch.randn(1, 128, dtype=torch.float32),)
+        m = torch.export.export(m, example_inputs).module()
+        targets = [n.target for n in m.graph.nodes]
+        self.assertTrue(torch.ops.quant.dequantize_codebook.default in targets)
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/torchao/prototype/quantization/codebook_coreml/__init__.py
+++ b/torchao/prototype/quantization/codebook_coreml/__init__.py
@@ -1,0 +1,13 @@
+from .api import CodebookWeightOnlyConfig
+from .codebook_ops import (
+    choose_qparams_and_quantize_codebook_coreml,
+    dequantize_codebook,
+)
+from .codebook_quantized_tensor import CodebookQuantizedTensor
+
+__all__ = [
+    "CodebookQuantizedTensor",
+    "CodebookWeightOnlyConfig",
+    "choose_qparams_and_quantize_codebook_coreml",
+    "dequantize_codebook",
+]
--- a/torchao/prototype/quantization/codebook_coreml/api.py
+++ b/torchao/prototype/quantization/codebook_coreml/api.py
@@ -1,0 +1,54 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+
+from dataclasses import dataclass
+from typing import List
+
+import torch
+
+from torchao.core.config import AOBaseConfig
+from torchao.prototype.quantization.codebook_coreml.codebook_quantized_tensor import (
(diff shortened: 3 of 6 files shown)


test/prototype/test_coreml_codebook.py
```diff
@@ -0,0 +1,87 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+import unittest
+
+import torch
+
+from torchao.prototype.quantization.coreml_codebook import (
```
metascroy: Should we call it coreml_codebook?  Why not something more generic?

torchao/prototype/quantization/coreml_codebook/api.py
```diff
@@ -0,0 +1,52 @@
...
+    Args:
+        dtype: torch.uint1 to torch.uint8, torch.int32 supported.
+    Returns:
+        Callable for quantization transformation.
+    """
+    dtype = config.dtype
+    group_size = config.group_size
+    weight = module.weight
+
+    if weight.numel() > 2**27:
```
metascroy: Where is this coming from?  Shouldn't it be the number of weight values in a group (where k-means is applied), rather than total number of weight values?

Also, I think we should error out rather than silently not quantize?

torchao/prototype/quantization/coreml_codebook/codebook_ops.py
```diff
@@ -0,0 +1,149 @@
...
+from torchao.utils import _register_custom_op
+
+quant_lib = torch.library.Library("quant", "FRAGMENT")
+register_custom_op = _register_custom_op(quant_lib)
+
+
+# wrapper around coreml util: https://github.com/apple/coremltools/blob/1c0e5cb1c1e3ab759af107b54f2be18b7c03f8aa/coremltools/models/neural_network/quantization_utils.py#L363
+@torch.no_grad
+@register_custom_op
+def choose_qparams_and_quantize_codebook(
```
metascroy: If this just wraps coremltools, maybe that should be reflected in the name, e.g., choose_qparams_and_quantize_with_coreml_codebook?

Let me add @YifanShenSZ as a reviewer to make sure we're using the latest/best quantization methods, and not depending on things they plan to deprecate.  I also saw k-mean code inside https://github.com/apple/coremltools/tree/main/coremltools/optimize/torch/palettization

torchao/prototype/quantization/coreml_codebook/codebook_quantized_tensor.py
```diff
@@ -0,0 +1,199 @@
...
+    def __repr__(self):
+        return (
+            f"{self.__class__.__name__}(codes={self.codes}, codebook={self.codebook}, group_size={self.group_size} "
+            f"shape={self.shape}, device={self.device}, dtype={self.dtype}, requires_grad={self.requires_grad})"
+        )
+
+    def _quantization_type(self):
+        return f"shape={self.shape}, group_size={self.group_size}, codebook_size={self.codebook.size(0)}, device={self.device}, code_dtype={self.codes.dtype}"
+
+    def dequantize(self, output_dtype: Optional[torch.dtype] = None) -> torch.Tensor:
```
metascroy, line 78: Should output_dtype be a property defined in init?

torchao/prototype/quantization/coreml_codebook/codebook_quantized_tensor.py
```diff
@@ -0,0 +1,199 @@
...
+        if output_dtype is None:
+            output_dtype = self.dtype
+
+        if isinstance(self.codes, UintxTensor):
+            codes = self.codes.get_plain()
+        else:
+            codes = self.codes
+        if codes.dtype != torch.int32:
+            # TODO: Investigate and support not casting to torch.int32 for indexing to improve performance
+            codes = codes.to(torch.int32)
```
metascroy: Why do we need to cast to int32?

torchao/prototype/quantization/coreml_codebook/codebook_quantized_tensor.py
```diff
@@ -0,0 +1,199 @@
...
+            code_dtype (torch.dtype): The dtype of the codes.
+            chunk_size (int): The chunk size to use during quantization (to control memory usage).
+        """
+
+        codebook, codes = choose_qparams_and_quantize_codebook(
+            input_tensor.to(torch.float32), code_dtype, group_size
+        )
+
+        if code_dtype in _DTYPE_TO_BIT_WIDTH:
+            codes = UintxTensor.from_uint8(codes, dtype=code_dtype)
```
metascroy: Does this mean codes are packed uintx tensors?  What about after export?

torchao/prototype/quantization/coreml_codebook/codebook_quantized_tensor.py
```diff
@@ -0,0 +1,199 @@
...
+         corresponds to a block in the original tensor. Shape is `(codebook_size, out_block_size, in_block_size)`.
+      block_size (Tuple[int, ...]): Granularity of quantization, specifying the dimensions of tensor
+         blocks that share the same quantization parameters.
+      shape (torch.Size): Shape of the original high-precision tensor.
+      dtype (torch.dtype): dtype of the original high-precision tensor.
+    """
+
+    @staticmethod
+    def __new__(
+        cls,
```
metascroy: Should we have a layout?

test/prototype/test_coreml_codebook.py
jerryzh168: need to think a bit about this, codebook is already taken by existing aqlm codebook quant: https://github.com/pytorch/ao/tree/main/torchao/prototype/quantization/codebook

we can just make this official as well, if it's going to be used by coreml I think it's going to stay?

torchao/prototype/quantization/coreml_codebook/api.py
jerryzh168: this is copied from previous version, I can remove as well

torchao/prototype/quantization/coreml_codebook/codebook_ops.py
jerryzh168: sounds good, I can add coreml to name

torchao/prototype/quantization/coreml_codebook/codebook_quantized_tensor.py
jerryzh168: indexing is only supported for int32 I think, see https://github.com/pytorch/ao/pull/1299#discussion_r1874097098

torchao/prototype/quantization/coreml_codebook/codebook_quantized_tensor.py
jerryzh168: yeah, this is also copied over from previous version, we can remove as well, would you prefer using qmin/qmax here

torchao/prototype/quantization/coreml_codebook/codebook_quantized_tensor.py
jerryzh168: that's not always needed, what are the use cases do you have?

metascroy: To distinguish between Uintx and QDQLayout?  Or just use QDQLayout

torchao/prototype/quantization/coreml_codebook/codebook_quantized_tensor.py
metascroy: I don't think we need qmin/qmax because they are indices.  I think it's OK to have a dtype/bitwidth that specifies that range, e.g., torch.uint3, but I was worried about using UintxTensor because it isn't convenient to unpack after export (vs. just using an uint8 tensor).

test/prototype/test_coreml_codebook.py
metascroy: So the choose_qparams_and_quantize can be coreml specific I guess because you're just wrapping coremltools.

But can't the dequantize ops and tensor be shared?

jerryzh168: maybe, but aqlm doesn't have a use case yet, so I'd save from trying to align the shapes/format/args etc. for now, aqlm path is also using scales, I think it's a bit different at least for now

torchao/prototype/quantization/coreml_codebook/codebook_quantized_tensor.py
jerryzh168: yeah makes sense, so uint8 is fine for storing uint1-uint7 in this case right? I can remove this part

torchao/prototype/quantization/coreml_codebook/codebook_quantized_tensor.py
jerryzh168: let's just use QDQLayout for now (as a default and no need to specify). until there is another use case

torchao/prototype/quantization/coreml_codebook/codebook_ops.py
jerryzh168: also will use `codebook_coreml` instead, following existing conventions that backend comes in the end

torchao/prototype/quantization/coreml_codebook/codebook_quantized_tensor.py
metascroy: Yes, uint8 is fine.  But there should be some arg that indicates the range/dtype

torchao/prototype/quantization/coreml_codebook/codebook_quantized_tensor.py
```diff
@@ -0,0 +1,199 @@
...
+        if isinstance(self.codes, UintxTensor):
+            codes = self.codes.get_plain()
+        else:
+            codes = self.codes
+        if codes.dtype != torch.int32:
+            # TODO: Investigate and support not casting to torch.int32 for indexing to improve performance
+            codes = codes.to(torch.int32)
+        return dequantize_codebook(
+            codes,
+            self.codebook,
```
metascroy: Should we have a granularity?

metascroy (approved):

torchao/prototype/quantization/coreml_codebook/codebook_quantized_tensor.py
jerryzh168: we'll use block_size for now, it can be extended to other granularities

PR #2444, opened 2025-06-26
[CPU] Fix ref path of DA8W4 cpp kernel

Xia-Weiwen:
Add a check of `CPU_CAPABILITY_AVX512` macro in the weight prepacking kernel to avoid errors when environments for building and deploying are different.


leslie-fang-intel (approved):
LGTM

PR #2446, opened 2025-06-26
fix scale shape in mx_format conversion

103yiran:


vkuzo:
could you share what is broken, and ideally add a test?

103yiran:
> could you share what is broken, and ideally add a test?

@vkuzo I found that the dimensions of `scale_e8m0_biased` and `data_lp` are not consistent, and the lowest dimension of `scale_e8m0_biased` is 1.

vkuzo:
hi @103yiran , thanks for the report!  I created https://github.com/pytorch/ao/pull/2506 to fix this with a test case, will close this PR.

PR #2447, opened 2025-06-26
float8 readme: remove duplication

vkuzo:
We had two duplicate example training loops in float8 readme, removing and making the same example work for all recipes

### DIFF
--- a/torchao/float8/README.md
+++ b/torchao/float8/README.md
@@ -12,16 +12,12 @@ and composable with key systems such as autograd, ```torch.compile``` and distri
 
 # Single GPU User API
 
-## float8 linear with dynamic tensorwise scaling
-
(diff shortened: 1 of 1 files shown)


danielvegamyhre (approved):

danielvegamyhre:
Did we copy/paste + run the example and confirm it doesn't produce nans, loss goes down, etc? With learning rate of 0.1 on these toy models with a `y.sum().backward()` loss function, I've seen NaN losses. My theory is that since we are basically training the model to predict all 0s to minimize the sum, with the high learning rate of 0.1 it takes a large step into nan territory.

vkuzo:
> Did we copy/paste + run the example and confirm it doesn't produce nans, loss goes down, etc?

Would be good for a future PR! It's just a toy example for now.

PR #2449, opened 2025-06-26
[NF4] Support nf4 tensor shard and gather

mori360:
Add  nf4_all_gather_into_tensor and scatter_nf4tensor to enable dispatch of `scatter` and `all_gather_into_tensor`
Add unit test to show that nf4 tensor keeps the same after distribute and gather.

### DIFF
--- a/test/dtypes/test_nf4.py
+++ b/test/dtypes/test_nf4.py
@@ -14,28 +14,28 @@ from typing import Tuple, Union
 import pytest
 import torch
 import torch.nn.functional as F
+
+import torchao
+from packaging import version
 from torch import nn
 from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import (
-    CheckpointWrapper,
     apply_activation_checkpointing,
+    CheckpointWrapper,
 )
 from torch.distributed.fsdp.wrap import ModuleWrapPolicy
 from torch.testing._internal.common_distributed import skip_if_lt_x_gpu
 from torch.testing._internal.common_fsdp import FSDPTest
 from torch.testing._internal.common_utils import (
-    TestCase,
     instantiate_parametrized_tests,
     parametrize,
     run_tests,
+    TestCase,
 )
-
-import torchao
-from packaging import version
 from torchao.dtypes._nf4tensor_api import nf4_weight_only
 from torchao.dtypes.nf4tensor import (
     _INNER_TENSOR_NAMES_FOR_SHARDING,
-    NF4Tensor,
     linear_nf4,
+    NF4Tensor,
     to_nf4,
 )
 from torchao.testing.utils import skip_if_rocm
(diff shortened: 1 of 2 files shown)


msaroufim:
Presumably we should observe a comms time reduction? would be nice to see some profiles

mori360:
> Presumably we should observe a comms time reduction? would be nice to see some profiles

Are there any baseline I could compare with?

torchao/dtypes/nf4tensor.py
```diff
@@ -273,19 +310,35 @@ def nf4_slice(aten_op, args, kwargs=None):
...
+        if nf4tensor.numel() != nf4tensor.size()[0] * nf4tensor.size()[1]:
+            raise NotImplementedError("NF4Tensor size does not match numel.")
+        for attr in _INNER_TENSOR_NAMES_FOR_SHARDING:
+            attr_size = [getattr(nf4tensor, attr).size()]
+            updated_attrs[attr] = aten_op(
+                getattr(nf4tensor, attr), *attr_size, **kwargs
+            )
+            updated_attrs.update(
+                {
+                    "stride": (nf4tensor.size()[1], 1),
```
weifengpy: why hardcode the stride?

torchao/dtypes/nf4tensor.py
```diff
@@ -273,19 +310,35 @@ def nf4_slice(aten_op, args, kwargs=None):
...
+            updated_attrs = apply_to_inner_tensors(nf4tensor, aten_op, args[1:], kwargs)
+            updated_attrs.update(
+                {
+                    "size": [nf4tensor.numel()],
+                    "stride": (1,),
+                }
+            )
+    else:
+        updated_attrs = {}
+        if nf4tensor.numel() != nf4tensor.size()[0] * nf4tensor.size()[1]:
```
weifengpy: is it possible to have out of bound access with nf4tensor.size()[1]? in the else branch, could `len(size) == 0`?

torchao/dtypes/nf4tensor.py
```diff
@@ -22,7 +22,44 @@
...
+def scatter_nf4tensor(func, *args, **kwargs):
+    output_tensor = args[0][0][0]
+    input_tensors = args[0][1]
+    new_attr, update_work = [], []
+    for attr in _INNER_TENSOR_NAMES_FOR_SHARDING:
+        input_attrs = []
+        if input_tensors:
+            for input_tensor in input_tensors[0]:
+                if hasattr(input_tensor, attr):
+                    input_attrs.append(getattr(input_tensor, attr))
```
weifengpy, line 57: what happens when tensor are not evenly divisible? or is there is possibility for uneven sharding?

torchao/dtypes/nf4tensor.py
```diff
@@ -22,7 +22,44 @@
 c10d_functional = torch.ops.c10d_functional
 
 
-NF4_OPS_TABLE: Dict[Any, Any] = {}
+def nf4_all_gather_into_tensor(func, *args, **kwargs):
+    nf4tensor = args[0][0]
```
weifengpy, line 28: can we assert len(args) and len(args[0]) before accessing them?

test/dtypes/test_nf4.py
```diff
@@ -741,6 +741,46 @@ def _test_qlora_fsdp2(
...
 
 
+class TestComm(FSDPTest):
+    @property
+    def world_size(self) -> int:
+        return 2
+
+    @pytest.mark.skipif(
+        version.parse(torch.__version__).base_version < "2.4.0",
+        reason="torch >= 2.4 required",
```
weifengpy: which api is needed in 2.4? DTensor?

test/dtypes/test_nf4.py
```diff
@@ -435,7 +435,7 @@ def test_tensor_view_valid(self, input_size: Union[Tuple[int], int]):
             inner_tensor = getattr(viewed_tensor, attr)
             self.assertEqual(inner_tensor.size(0), inner_tensor.numel())
 
-    @parametrize("input_size", [(512 * 512,), (512, 512)])
+    @parametrize("input_size", [(512 * 512,)])
```
weifengpy, line 448: why removing (512, 512) ?

weifengpy:
> Presumably we should observe a comms time reduction? would be nice to see some profiles

perf should be on-par. This BE refactoring upstreams NF4 specific logic from torchtune to DTensor. It creates an example for people to follow to handle tensor subclass + DTensor state dict

test/dtypes/test_nf4.py
mori360: tensor=torch.randn(512,512) and tensor.view(512,512) is now valid after changes at nf4_view, move it to `test_tensor_2d_view_valid`

torchao/dtypes/nf4tensor.py
mori360: fixed

torchao/dtypes/nf4tensor.py
mori360: if sharded unevenly(did not go through the split dispatch), we will compare the input and output sizes here

PR #2452, opened 2025-06-27
Graduate debug handle in torchao

Gasoonjia:
Summary:
This diff maket torchao's debugging infra fully leverage node["from_node"] info and get rid of debug handle.

debug handle, we will miss you 🫡

Reviewed By: jerryzh168

Differential Revision: D76628702

### DIFF
--- a/torchao/quantization/pt2e/_numeric_debugger.py
+++ b/torchao/quantization/pt2e/_numeric_debugger.py
@@ -30,6 +30,21 @@ FROM_NODE_KEY = "from_node"
 log = logging.getLogger(__name__)
 
 
+@dataclass(frozen=True)
+class NodeSourceDebugInfo:
+    """
+    Contains node source information for locating the node in the original graph.
+    This replaces the numeric debug handle approach with direct node source info.
+    """
+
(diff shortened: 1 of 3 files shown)


torchao/quantization/pt2e/_numeric_debugger.py
```diff
@@ -30,6 +30,18 @@
...
 
+@dataclass(frozen=True)
+class NodeSourceDebugInfo:
+    """
+    Contains node source information for debugging purposes.
+    This replaces the numeric debug handle approach with direct node source info.
+    """
+
+    name: str
+    graph_id: int
```
jerryzh168, line 44: could you add some comments here about what is graph_id?

jerryzh168 (approved):

torchao/quantization/pt2e/_numeric_debugger.py
Gasoonjia: added

jerryzh168: thanks, I see, is graph_id enough for operator level debugging?

PR #2455, opened 2025-06-27
[moe training] Cast to mixed precision policy param dtype in fsdp_pre_all_gather hook

danielvegamyhre:
## Stack
- https://github.com/pytorch/ao/pull/2455 <- this PR

## Summary

- After examining this [example](https://github.com/pytorch/pytorch/blob/4500a4aa50141ed30e093ef8491b30d1d1287348/test/distributed/_composable/fsdp/test_fully_shard_extensions.py#L117) and discussing with @weifengpy, we think that when implementing fsdp pre/post all gather hooks for a tensor subclass, the developer must handle casting params to the MP policy param dtype in the pre all-gather hook themselves.
    - I thought the downcasting to save network bandwidth via bf16 all gather instead of fp32 would be done automatically by fsdp, but it appears when implementing these hooks we have to handle it ourselves. 
    - Tracking issue to determine contract between MP policy and tensor subclass extension: https://github.com/pytorch/pytorch/issues/157395
- Update fsdp_post_all_gather_hook to more correctly handle case where `out != None` (see code comments for details)
- Remove `dtype` param from ScaledGroupedMM tensor, as it is no longer needed when doing the casting in pre all gather rather than post all gather.

## Test plan
- `./test/prototype/moe_training/test_fsdp.sh`
- Manual test with torchtitan llama4 debug model: `NGPU=2 CONFIG_FILE="./torchtitan/experiments/llama4/train_configs/debug_model.toml" ./run_train.sh --training.steps=10 --model.converters="float8" --float8.recipe_name="rowwise"  --float8.moe_fqns_prototype="experts" `

### DIFF
--- a/torchao/prototype/moe_training/tensor.py
+++ b/torchao/prototype/moe_training/tensor.py
@@ -46,16 +46,15 @@ class ScaledGroupedMMTensor(torch.Tensor):
     def __new__(
         cls,
         tensor: torch.Tensor,
-        dtype: torch.dtype,
     ):
-        logger.info(f"ScaledGroupedMMTensor __new__: tensor.dtype={tensor.dtype}, dtype: {dtype}, shape: {tensor.shape}")
+        # logger.info(f"ScaledGroupedMMTensor __new__: tensor.dtype={tensor.dtype}, dtype: {dtype}, shape: {tensor.shape}")
         return torch.Tensor._make_wrapper_subclass(
             cls,
             tensor.size(),
             strides=tensor.stride(),
             storage_offset=tensor.storage_offset(),
             memory_format=suggest_memory_format(tensor),
-            dtype=dtype,
+            dtype=tensor.dtype,
             layout=tensor.layout,
             device=tensor.device,
             pin_memory=tensor.is_pinned(),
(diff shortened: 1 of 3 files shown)


torchao/prototype/moe_training/tensor.py
```diff
@@ -156,6 +162,13 @@ def fsdp_post_all_gather(
         out: Optional[torch.Tensor] = None,
     ):
         (data,) = all_gather_outputs
+        logger.debug(
+            f"ScaledGroupedMMTensor fsdp_post_all_gather: data.dtype={data.dtype}, param_dtype: {param_dtype}"
+        )
+
+        if out is not None:
+            return
```
weifengpy, line 172: how did you decide when to return, and when to `out.copy_(data)`

weifengpy: is out.storage the same to data.storage? if yes, it's correct to return direclty

danielvegamyhre: I considered doing `out.copy_(data)` like in this [example](https://github.com/pytorch/pytorch/blob/20e40492b046b9287726d3ec656117e4dc38f0e2/test/distributed/_composable/fsdp/test_fully_shard_extensions.py#L130), but found it resulted in the following error:

`RuntimeError: a leaf Variable that requires grad is being used in an in-place operation.`

I also found that FSDP2 still seems to work when returning directly without doing this, so I omitted it. 

Can you share more details on how `out` is intended to be used in this fsdp_post_all_gather hook (as opposed to returning the outputs directly)?

danielvegamyhre: Discussed offline and updated, see code comments

danielvegamyhre:
cc @drisspg @vkuzo for review

torchao/prototype/moe_training/tensor.py
```diff
@@ -156,6 +153,25 @@ def fsdp_post_all_gather(
...
+                    == out._data.untyped_storage().data_ptr()
+                )
+            else:
+                assert out._data.dtype == param_dtype, (
+                    f"{out._data.dtype} {param_dtype}"
+                )
+                out._data.copy_(data)
+            return
+
+        # For training step 0, out=None, so we need to return a new ScaledGroupedMMTensor.
```
vkuzo, line 174: do we have a test for this?

vkuzo (approved):

torchao/prototype/moe_training/tensor.py
danielvegamyhre: We have a test for float MoE + FSDP training. We don't have a test verifying which code branch is followed in this fsdp_post_all_gather hook at training step 0 vs 1, but I think the FSDP test alone is sufficient. Let me know if you have other thoughts.

danielvegamyhre:
MPS test failures are unrelated to this change

PR #2458, opened 2025-06-29
Fix MX + vllm

drisspg:
Stacked PRs:
 * __->__#2458

--- --- ---

Fix MX + vllm

# Summary

We calcuate scales by viewing 2d (B,H) inputs as (B,h//bsisze, bsize) and then we reduce the last dim to 1. This flatten was causing the same metatdata check to fail


danielvegamyhre (approved):
LGTM, in the PR description or a comment could you mention the issue this fixes and why/how? For readers of the code to be able to easily understand the need for this shape change. 

Also do we have a vllm integration test that would catch this issue?

vkuzo:
add a test?

PR #2461, opened 2025-06-30
Register choose_qparams_affine_float8 as custom op

angelayi:


test/integration/test_integration.py
```diff
@@ -2076,6 +2078,26 @@ def forward(self, x):
             self.assertTrue(torch.ops.torchao.choose_qparams_affine.default in targets)
             self.assertTrue(torch.ops.torchao.quantize_affine.default in targets)
             self.assertFalse(torch.ops.aten.narrow.default in targets)
+    
+    def test_export_float8(self):
```
jerryzh168, line 2086: need this skip according to CI: https://github.com/pytorch/ao/blob/c2a6568a04075acc371a338206216bb65536fb27/test/dtypes/test_affine_quantized_float.py#L74-L76

PR #2470, opened 2025-07-01
increase torch dynamo cache size limit to support all tests

Gasoonjia:
Summary: Some tests was suppressed due to cache size limitation. This diff increase the cache size to unblock the issue.

Reviewed By: jerryzh168

Differential Revision: D76794879


jerryzh168:
plese fix ruff

PR #2471, opened 2025-07-01
Fix docstrings for quantization API docs

jainapurva:


torchao/quantization/linear_activation_quantized_tensor.py
```diff
@@ -288,7 +288,7 @@ def _(func, types, args, kwargs):
     )
 
 
-to_linear_activation_quantized = LinearActivationQuantizedTensor.from_float
+to_linear_activation_quantized = LinearActivationQuantizedTensor.from_float  # Converts a float tensor to LinearActivationQuantizedTensor for dynamic activation quantization
```
jerryzh168, line 291: FYI: we are going to remove these to functions I think to reduce the indirections

jerryzh168 (approved):

torchao/quantization/linear_activation_quantized_tensor.py
jainapurva: Sounds good. In this PR, I wish to keep the docstring intact, and remove it all together with the function.

PR #2473, opened 2025-07-02
[moe training] Add TP support for routed experts

danielvegamyhre:
## Stack
- https://github.com/pytorch/ao/pull/2473 <- this PR

## Summary
- Adds TP integration test and testing utils file for shared functions used in single GPU / FSDP / TP training tests
- Update _scaled_grouped_mm to support 3D A tensor, which is needed for the shared expert ([code](https://github.com/pytorch/torchtitan/blob/c08c9d4962ea843dd786b850d1716861955b9a9f/torchtitan/experiments/llama4/model/moe.py#L316)).
- Make `offs` optional, to handle shared_expert case where num_experts=1 (no group offsets needed since there's only 1 token group).

## Test plan
- `./test/prototype/moe_training/test_tp.sh`

### DIFF
--- a/test/prototype/moe_training/test_tp.py
+++ b/test/prototype/moe_training/test_tp.py
@@ -1,0 +1,245 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+######################################################################
+#
+# To run these unit tests, use the following command:
+#
+# torchrun --nproc_per_node=${NUM_GPUS} -m pytest test_tp.py
+#
+#######################################################################
+
+import copy
+import os
+
+import pytest
+import torch
+from torch import distributed as dist
+from torch import nn
+from torch.distributed._tensor import DTensor
+from torch.distributed.device_mesh import DeviceMesh, init_device_mesh
+from torch.distributed.tensor import Partial, Replicate, Shard
+from torch.nn import functional as F
+
+try:
+    from torch.distributed.tensor.parallel import (
+        PrepareModuleInputOutput,
+        parallelize_module,
+    )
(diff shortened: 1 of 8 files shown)


test/prototype/moe_training/test_tp.py
```diff
@@ -0,0 +1,235 @@
...
+from torch.distributed.device_mesh import DeviceMesh, init_device_mesh
+from torch.distributed.tensor import Partial, Replicate, Shard
+from torch.distributed.tensor.parallel import (
+    PrepareModuleInputOutput,
+    parallelize_module,
+)
+from torch.nn import functional as F
+
+# this feature requires CUDA and SM89+
+if not torch.cuda.is_available() or torch.cuda.get_device_capability() < (8, 9):
```
vkuzo, line 41: is the bf16 group gemm working on A100? If yes, I would vote for adding an emulation mode and running this test in emulation mode.  We do this for float8 and MX training.

test/prototype/moe_training/testing_utils.py
```diff
@@ -0,0 +1,33 @@
+from torch import nn
+
+from torchao.prototype.moe_training.tensor import ScaledGroupedMMTensor
+
+
+def _validate_model_conversion(
```
vkuzo, line 6: nit: why do we need recursion to check this?

danielvegamyhre: It's just a generic way of checking all target FQNs were converted properly, and verifying all non-target FQNs were correctly not converted. It can easily be applied when we extend tests to other MoE models as well beyond just the torchtitan llama4 one I started with.

test/prototype/moe_training/test_tp.py
danielvegamyhre: I will look into this and add emulation if bf16 grouped gemm builds on a100

PR #2463, opened 2025-06-30
Add Float8RowwiseTensor

jerryzh168:
Add Float8RowwiseTensor

Summary:
Splits out the float8 rowwise quantized path (both act and weight) of AQT to Float8RowwiseTensor

Next: could potentially incorporate the per tensor activation path there as well
Next: we can split the per tensor weight path to another Tensor as well, so we can deprecate AQT path for float8

Test Plan:
python test/dtypes/test_affine_quantized_float.py
python test/quantization/quantize_/test_float8_rowwise_tensor.py

Reviewers:

Subscribers:

Tasks:

Tags:

PR #2474, opened 2025-07-02
Add all fbgemm kernel Tensors into Int4WeightOnlyConfig and Float8DynamicActivationInt4WeightConfig

jerryzh168:
Stacked PRs:
 * __->__#2474
 * #2463
 * #2479
 * #2437

--- --- ---

### Add all fbgemm kernel Tensors into Int4WeightOnlyConfig and Float8DynamicActivationInt4WeightConfig

Summary:
att, we will deprecate FbgemmConfig since it's a single kernel.
we'd like to categorize things to derived dtype + packed format

Test Plan:
python test/quantization/quantize_/test_int4_groupwise_preshuffle.py

Reviewers:

Subscribers:

Tasks:

Tags:


torchao/quantization/quant_api.py
```diff
@@ -1214,6 +1239,39 @@ def _int4_weight_only_transform(
     return module
 
 
+@dataclass
+class Float8ActivationInt4WeightConfig(AOBaseConfig):
+    group_size: int = 128
+    use_preshuffle: bool = False
+    kernel: str = "fbgemm"
```
andrewor14: Use `GemmKernelChoice` here?

torchao/quantization/quant_api.py
```diff
@@ -1214,6 +1239,39 @@ def _int4_weight_only_transform(
...
+    assert use_preshuffle, (
+        f"only use_preshuffle == True is supported right now, got: {use_preshuffle}"
+    )
+    assert kernel == "fbgemm", f"only fbgemm kernel is supported, got: {kernel}"
+    weight = module.weight
+    block_size = tuple([1 for _ in range(weight.ndim - 1)] + [group_size])
+    new_weight = Int4GroupwisePreshuffleTensor.from_float(
+        module.weight,
+        block_size,
+        activation_dtype="fp8",
```
andrewor14: should we use a `torch.dtype` here instead of a string

torchao/quantization/__init__.py
```diff
@@ -44,6 +44,7 @@
 from .quant_api import (
     CutlassInt4PackedLayout,
     FbgemmConfig,
```
andrewor14, line 46: What's the plan for `FbgemmConfig`? Looks like it was added only ~1.5 months ago but it's technically public API. Do we know if anyone's using it already? I don't think it's released yet so wonder if it's OK to just remove it?

torchao/quantization/quant_api.py
```diff
@@ -1134,15 +1136,38 @@ def _int4_weight_only_quantize_tensor(weight, config):
...
+            )
+            return new_weight
+        else:
+            new_weight = Int4GroupwiseTensor.from_float(
+                weight,
+                block_size,
+            )
+            return new_weight
+
     mapping_type = MappingType.ASYMMETRIC
```
andrewor14, line 1181: should we assert `GemmKernelChoice == ATEN` here and `use_preshuffle == False`? Maybe also add a comment to say this is the tinygemm kernel route?

torchao/quantization/quant_api.py
andrewor14: also add a docstring?

torchao/quantization/quant_api.py
```diff
@@ -1214,6 +1239,39 @@ def _int4_weight_only_transform(
...
 
+@dataclass
+class Float8ActivationInt4WeightConfig(AOBaseConfig):
+    group_size: int = 128
+    use_preshuffle: bool = False
+    kernel: str = "fbgemm"
+
+
+@register_quantize_module_handler(Float8ActivationInt4WeightConfig)
+def _(module: torch.nn.Module, config: Int4WeightOnlyConfig) -> torch.nn.Module:
```
andrewor14: wrong config type here, also should we call this `_float8_activation...transform` to be consistent with other configs?

torchao/quantization/__init__.py
jerryzh168: we'll remove it, it is used in some internal script but we'll update these as well

torchao/quantization/quant_api.py
jerryzh168: updated the structure after some discussions, please take a look again

torchao/quantization/quant_api.py
jerryzh168: this is updated, please take a look again

torchao/quantization/quant_api.py
jerryzh168: done

torchao/quantization/quantize_/workflows/packing_format.py
```diff
@@ -0,0 +1,13 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+
+from enum import Enum
+
+
+class PackingFormat(str, Enum):
```
andrewor14: add some docstrings?

torchao/quantization/quant_api.py
```diff
@@ -1123,6 +1125,9 @@ class Int4WeightOnlyConfig(AOBaseConfig):
     zero_point_domain: Optional[ZeroPointDomain] = ZeroPointDomain.NONE
     set_inductor_config: bool = True
     preserve_zero: Optional[bool] = None
+    # since not all tensors are migrated to the new structure yet,
+    # we use `_legacy' to represent the previous layout
+    packing_format: PackingFormat = "_legacy"
```
andrewor14: Can "legacy" mean different things for different configs? I wonder if we should make this optional instead, where None represents "legacy"?

torchao/quantization/quantize_/workflows/packing_format.py
andrewor14: also are there concerns about python version compatibility? I remember we stayed away from str enums before because of this

torchao/quantization/quantize_/workflows/int4/int4_tensor.py
```diff
@@ -31,22 +30,22 @@
     pack_int4 = None
 
 
-class FbgemmInt4Tensor(TorchAOBaseTensor):
-    tensor_data_attrs = ["packed_weight", "scale", "zero_point"]
-    tensor_attributes = ["group_size", "shape"]
+class Int4Tensor(TorchAOBaseTensor):
```
andrewor14, line 33: Add some docstrings here? In particular I think it may be a bit confusing that `Int4WeightOnlyConfig` does not always give you this tensor (sometimes gives you AQT instead). Should we clarify that we're targeting the fbgemm kernels here and not the tinygemm ones? How is this different from Int4PreshuffledTensor?

torchao/quantization/quant_api.py
```diff
@@ -1220,6 +1243,38 @@ def _int4_weight_only_transform(
     return module
 
 
+@dataclass
+class Float8ActivationInt4WeightConfig(AOBaseConfig):
```
andrewor14, line 1254: docstrings?

andrewor14: also can you add the new APIs introduced in this PR to `ao/docs/source/api_ref_quantization.rst`, since these are now public APIs?

torchao/quantization/quant_api.py
```diff
@@ -1140,15 +1145,33 @@ def _int4_weight_only_quantize_tensor(weight, config):
...
+        )
+        return new_weight
+    elif packing_format == "plain":
+        new_weight = Int4Tensor.from_float(
+            weight,
+            block_size,
+        )
+        return new_weight
+
+    assert packing_format == "_legacy", f"Unsupported packing_format: {packing_format}"
```
andrewor14: nit: use the enums (e.g. `PackingFormat.LEGACY`) here and other places instead of raw strings to be more robust?

torchao/quantization/quantize_/workflows/packing_format.py
```diff
@@ -0,0 +1,13 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
andrewor14: actually is this the right place for this file? It's not really a "workflow", more like a "util class" or "common"

andrewor14 (approved):
Looks great, left some comments mostly about documentation

torchao/quantization/quantize_/workflows/packing_format.py
jerryzh168: oh this one is moved to another folder, let me delete from here, please check the previous PR for docstrings

yeah this is fine, strEnum is introduced in 3.11, we can use that after 3.10 EOL, I can add a comment

torchao/quantization/quantize_/workflows/packing_format.py
jerryzh168: yeah this is moved to common, see previous PR

torchao/quantization/quantize_/workflows/int4/int4_tensor.py
jerryzh168: sure, will add some docs

> Should we clarify that we're targeting the fbgemm kernels here and not the tinygemm ones?

fbgemm is faster when batch_size is larger https://github.com/pytorch/ao/pull/2273

> How is this different from Int4PreshuffledTensor?

just packing format is different, this one is "plain" the other one is "preshuffled"

torchao/quantization/quant_api.py
jerryzh168: yeah legacy just means no packing format, (it's implemented with AQT), I plan to remove the support for legacy at some point and don't want to complicate the typing here

PR #2475, opened 2025-07-02
[moe training] Add 2D parallel (FSDP2 + TP) tests for routed experts

danielvegamyhre:
## Stack
- https://github.com/pytorch/ao/pull/2481 
- https://github.com/pytorch/ao/pull/2475 <- this PR

## Summary
- New integration test uses NGPU=4, with FDSP=2 and TP=2. Applies FSDP2 first, then applies TP.
- Note all tests currently only apply float8 to routed experts, not shared experts.

## Test plan 
- `./test/prototype/moe_training/test_fsdp_tp.sh`

### DIFF
--- a/test/prototype/moe_training/test_fsdp_tp.sh
+++ b/test/prototype/moe_training/test_fsdp_tp.sh
@@ -1,0 +1,1 @@
+torchrun --nproc_per_node=4 --local-ranks-filter=0 -m pytest test/prototype/moe_training/test_fsdp_tp.py -s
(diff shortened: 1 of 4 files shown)


test/prototype/moe_training/test_fsdp_tp.sh
```diff
@@ -0,0 +1 @@
+torchrun --nproc_per_node=4 --local-ranks-filter=0 -m pytest test/prototype/moe_training/test_fsdp_tp.py -s
```
vkuzo, line 1: at this point I'd also add `test_everything.sh` which runs all of the shell scripts, same as we do for float8 training

vkuzo (approved):

danielvegamyhre:
MPS test failure is unrelated to this change

PR #2476, opened 2025-07-02
[CPU] Add concat-linear fusion pass for da8w4

Xia-Weiwen:
**Summary**
This PR adds a concat-linear fusion pass for da8w4 on CPU. The pass fuses the following pattern
```
    da8w4_linear_cpu(x, ..., w1, ...) -- y1
  /
x --da8w4_linear_cpu(x, ..., w2, ...) -- y2
  \...
    da8w4_linear_cpu(x, ..., wN, ...) -- yN
```
to
```
x -- da8w4_linear_cpu(x, ..., w_concat, ...) -- y_concat -- split -- (y1, y2, yN)
```
The fusion pass is registered as a custom post_grad pass in Inductor. The pass takes effect only when `torch._inductor.config.cpp.enable_concat_linear` is true.

Benchmarks show that total CPU time of linear is reduced by >5% with concat linear when running Llama3.1-8B with 32 cores on a 6th gen of Intel(R) Xeon(R).

**Test plan**
```
pytest test/quantization/test_quant_api.py -k test_8da4w_concat_linear_cpu
```

### DIFF
--- a/test/quantization/test_quant_api.py
+++ b/test/quantization/test_quant_api.py
@@ -763,6 +763,65 @@ class TestQuantFlow(TestCase):
             assert torch.allclose(dqw1, dqw1_ref)
             assert torch.allclose(dqw2, dqw2_ref)
 
+    @unittest.skipIf(
+        "CPU" not in torch._C._dispatch_dump("torchao::da8w4_linear_cpu"),
+        reason="cpp kernels not built",
+    )
+    @unittest.skipIf(not TORCH_VERSION_AT_LEAST_2_8, "Test only enabled for 2.8+")
+    @common_utils.parametrize("x_dim", [2, 3])
+    @common_utils.parametrize("bias", [True, False])
+    def test_8da4w_concat_linear_cpu(self, x_dim, bias):
+        N, K = 64, 128
+
+        class Mod(torch.nn.Module):
+            def __init__(self, bias):
+                super().__init__()
+                self.linear1 = torch.nn.Linear(K, N, bias=bias)
+                self.linear2 = torch.nn.Linear(K, N, bias=bias)
+                self.linear3 = torch.nn.Linear(K, N, bias=bias)
+
+            def forward(self, x):
+                a = self.linear1(x)
+                b = self.linear2(x)
+                c = self.linear3(x)
+                return a + b + c
+
+        dtype = torch.bfloat16
+        device = "cpu"
+        m = Mod(bias).eval().to(dtype).to(device)
+        x_shape = [2] * x_dim
+        x_shape[-1] = K
+        x = torch.rand(x_shape, dtype=dtype, device=device)
+        with torch.no_grad():
+            quantize_(
+                m,
+                Int8DynamicActivationInt4WeightConfig(
+                    group_size=32,
+                    layout=Int8DynamicActInt4WeightCPULayout(),
+                    act_mapping_type=MappingType.SYMMETRIC,
+                ),
+            )
+            # Need to turn on freezing to get the pattern
+            # set enable_concat_linear to true to enable the fusion
+            with torch._inductor.config.patch(
+                {"freezing": True, "cpp.enable_concat_linear": True}
+            ):
+                y, code = torch._inductor.utils.run_and_get_code(
+                    torch.compile(m, fullgraph=True, dynamic=True),
+                    x,
+                )
+            # ensure the expected op occurs only once in the code after fusion
+            # The trailing "(" is to avoid matching the op in the comment
+            assert code[0].count("torch.ops.torchao.da8w4_linear_cpu.default(") == 1
+            with torch._inductor.config.patch(
+                {"freezing": True, "cpp.enable_concat_linear": False}
+            ):
+                y_ref, code = torch._inductor.utils.run_and_get_code(
+                    torch.compile(m, fullgraph=True, dynamic=True),
+                    x,
+                )
+            assert torch.allclose(y, y_ref)
+
     # TODO(#1690): move to new config names
     @unittest.skipIf(not TORCH_VERSION_AT_LEAST_2_4, "Test only enabled for 2.4+")
     @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
--- a/torchao/dtypes/uintx/dyn_int8_act_int4_wei_cpu_layout.py
+++ b/torchao/dtypes/uintx/dyn_int8_act_int4_wei_cpu_layout.py
@@ -3,6 +3,7 @@
 #
 # This source code is licensed under the BSD 3-Clause license found in the
 # LICENSE file in the root directory of this source tree.
+import operator
 from dataclasses import dataclass
 from typing import Tuple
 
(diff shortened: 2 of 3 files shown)


Xia-Weiwen:
Hi @CaoE Could you please also review this PR? I cannot add you as a reviewer.

torchao/dtypes/uintx/dyn_int8_act_int4_wei_cpu_layout.py
```diff
@@ -310,3 +311,240 @@ def _linear_int8_act_int4_weight_cpu_impl(input_tensor, weight_tensor, bias):
     y = y.reshape(*orig_act_size[:-1], orig_out_features)
 
     return y.to(orig_dtype)
+
+
+# Inductor FX passes for concat linear
+def _is_valid_concat_linear_da8w4_fusion(computation_nodes):
```
leslie-fang-intel: Feels we should move these code to somewhere else since it's a graph optimization pass. Perhaps we should put all the graph optimization passes together.

torchao/dtypes/uintx/dyn_int8_act_int4_wei_cpu_layout.py
```diff
@@ -310,3 +311,240 @@ def _linear_int8_act_int4_weight_cpu_impl(input_tensor, weight_tensor, bias):
...
+            return get_hash_for_files((__file__,))
+
+    da8w4_concat_linear_pass = DA8W4ConcatLinearCpuPass()
+    device = "cpu"
+    init_backend_registration()
+    device_scheduling = get_scheduling_for_device(device)
+    device_python_wrapper = get_wrapper_codegen_for_device(device, False)
+    device_cpp_wrapper = get_wrapper_codegen_for_device(device, True)
+    device_custom_pass = da8w4_concat_linear_pass
+    register_backend_for_device(
```
leslie-fang-intel: Using this API here feels a bit confusing, since we're not trying to register a new backend—the CPP backend should already be registered. We only intend to register a custom pass for it.

torchao/dtypes/uintx/dyn_int8_act_int4_wei_cpu_layout.py
Xia-Weiwen: Yes. I agree. However, the existing FX passes are put under the PT2E directory, which is not suitable for the pass in this PR. We may need more inputs from Meta. I will ask Jerry for review later.

torchao/dtypes/uintx/dyn_int8_act_int4_wei_cpu_layout.py
Xia-Weiwen: I feel it confusing too but it is the only way to register a custom post_grad pass in Inductor. We may need to discuss with Meta about how to refine the registration of custom passes in Inductor later.

Xia-Weiwen: > I feel it confusing too but it is the only way to register a custom post_grad pass in Inductor. We may need to discuss with Meta about how to refine the registration of custom passes in Inductor later.

Correction: it is the only way to register a custom post_grad **gm** pass in Inductor

torchao/dtypes/uintx/dyn_int8_act_int4_wei_cpu_layout.py
Xia-Weiwen: I have moved it to a separate location. Thanks.

torchao/dtypes/uintx/dyn_int8_act_int4_wei_cpu_layout.py
leslie-fang-intel: Could we extend it in PyTorch?

torchao/Inductor_passes/cpu.py
```diff
@@ -0,0 +1,251 @@
...
+                return get_hash_for_files((__file__,))
+
+        da8w4_concat_linear_pass = DA8W4ConcatLinearCpuPass()
+        device = "cpu"
+        init_backend_registration()
+        device_scheduling = get_scheduling_for_device(device)
+        device_python_wrapper = get_wrapper_codegen_for_device(device, False)
+        device_cpp_wrapper = get_wrapper_codegen_for_device(device, True)
+        device_custom_pass = da8w4_concat_linear_pass
+        register_backend_for_device(
```
leslie-fang-intel: Could we extend this implementation instead of using the `register_backend_for_device` in PyTorch?

Xia-Weiwen: Yes. We need to discuss with Meta first. For now, I think we better use the existing path to register a custom pass. Thanks.

Xia-Weiwen:
Hi @jerryzh168 Could you please review this PR? This PR adds new Inductor passes and we would like to hear your suggestions on where to put the code. Thanks.

torchao/Inductor_passes/cpu.py
Xia-Weiwen: I found that we can get gm by `graph.owning_module` so we can use `post_grad_custom_post_pass` to apply the pass. Then we don't need to use the `register_backend_for_device` API. Thanks.

torchao/prototype/inductor/fx_passes/da8w4_concat_linear_fusion_cpu.py
```diff
@@ -0,0 +1,216 @@
...
+                                graph.erase_node(user)
+
+
+# Define and register a custom pass for concat linear
+# We always register the pass when calling this function
+# but it only takes effect when config.cpp.enable_concat_linear is set to True
+def register_da8w4_concat_linear_cpu_pass():
+    from torch._inductor import config as inductor_config
+
+    inductor_config.post_grad_custom_post_pass = _concat_linear_dq8w4_cpu
```
leslie-fang-intel, line 216: > I found that we can get gm by graph.owning_module so we can use post_grad_custom_post_pass to apply the pass. Then we don't need to use the register_backend_for_device API. Thanks.

But this one may silently cause conflict right?

Xia-Weiwen: Yes, exactly. We need to extend the passes to lists in PyTorch.

test/quantization/test_quant_api.py
```diff
@@ -765,6 +765,65 @@ def test_8da4w_cpu(self, dtype, x_dim, bias, bs, sym_quant_a):
...
             assert torch.allclose(dqw2, dqw2_ref)
 
+    @unittest.skipIf(
+        "CPU" not in torch._C._dispatch_dump("torchao::da8w4_linear_cpu"),
+        reason="cpp kernels not built",
+    )
+    @unittest.skipIf(not TORCH_VERSION_AT_LEAST_2_8, "Test only enabled for 2.8+")
+    @common_utils.parametrize("x_dim", [2, 3])
+    @common_utils.parametrize("bias", [True, False])
+    def test_8da4w_concat_linear_cpu(self, x_dim, bias):
```
jerryzh168: nit: you probably want to create a new file for each type of quant to reduce interference with other quant methods

torchao/prototype/inductor/fx_passes/da8w4_concat_linear_fusion_cpu.py
jerryzh168: cc @jansel @eellison

jerryzh168:
LGTM, @Xia-Weiwen I had a message in slack, wondering if you want to migrate the cpu related stuff to experimental folder to be more consistent with the rest of the CPU kernels: https://github.com/pytorch/ao/tree/main/torchao/experimental

test/quantization/test_quant_api.py
Xia-Weiwen: @jerryzh168 Thanks for the suggestion. Do you mean creating a new file of unit tests for each type of quant?

Xia-Weiwen:
> LGTM, @Xia-Weiwen I had a message in slack, wondering if you want to migrate the cpu related stuff to experimental folder to be more consistent with the rest of the CPU kernels: https://github.com/pytorch/ao/tree/main/torchao/experimental

Thanks for reviewing and sorry for the late reply.

test/quantization/test_quant_api.py
jerryzh168: yeah, similar to https://github.com/pytorch/ao/blob/main/test/quantization/quantize_/test_int4_groupwise_preshuffle.py but you can put it under `test/quantization/` folder for now

for some context, we are also thinking of splitting the AQT tensor to reduce the code that people need to interact with when they add a new type of quantization

Xia-Weiwen: ok. I have put them in a separate file. Please take a look again. Thanks.

jerryzh168 (approved):

torchao/prototype/inductor/fx_passes/da8w4_concat_linear_fusion_cpu.py
leslie-fang-intel: > Yes, exactly. We need to extend the passes to lists in PyTorch.

Still feels extend current design to avoid this conflict will be a better solution. Let's add some notes for the potential conflict at least.

Xia-Weiwen: Hi @jansel @eellison I think we need to extend the custom passes in Inductor (to make them lists). I will probably submit a PR for it. The custom passes added by https://github.com/pytorch/pytorch/pull/154841 does not meet our needs. Do you have comments? Thanks.

eellison: List registration makes sense to me.

eellison: List registration makes sense to me.

Xia-Weiwen: Thanks

PR #2481, opened 2025-07-02
[WIP] [moe training] Add tests for 3D parallel (FDSP + TP + EP) for routed experts

danielvegamyhre:
WIP because it requires https://github.com/pytorch/pytorch/pull/157216 to land first

## Stack
- https://github.com/pytorch/ao/pull/2481 <- this PR

## Summary
- Test applies 3D parallelism to routed experts, where FSDP=2, TP=2, EP=2

## Test plan
- `./test/prototype/moe_training/test_fsdp_tp_ep.sh`

### DIFF
--- a/test/prototype/moe_training/test_fsdp_tp_ep.py
+++ b/test/prototype/moe_training/test_fsdp_tp_ep.py
@@ -7,7 +7,7 @@
 #
 # To run these unit tests, use the following command:
 #
-# torchrun --nproc_per_node=${NUM_GPUS} -m pytest test_fsdp_tp.py
+# torchrun --nproc_per_node=${NUM_GPUS} -m pytest test_fsdp_tp_ep.py
 #
 #######################################################################
 
@@ -74,7 +74,7 @@ except ImportError:
         # ["experts,shared_expert"],
     ],
 )
-def test_moe_float8_training_fsdp_tp(target_fqns: list[str]):
+def test_moe_float8_training_fsdp_tp_ep(target_fqns: list[str]):
     assert torch.cuda.is_available()
 
     # setup distributed for tp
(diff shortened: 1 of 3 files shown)


test/prototype/moe_training/test_fsdp_tp_ep.py
```diff
@@ -0,0 +1,263 @@
...
+    quantize_(model, config=config, filter_fn=moe_module_filter_fn)
+
+    # validate that only the experts were converted
+    _validate_model_conversion(
+        model,
+        target_fqns=target_fqns,
+    )
+
+    # apply TP and EP
+    apply_moe_ep_tp(
```
danielvegamyhre, line 123: cc @tianyu-l I would appreciate your feedback on the parallelism configuration for this test.

Note:
1. This is all using your PR for local testing: https://github.com/pytorch/torchtitan/pull/1324
2. The target model is just importing the Llama4 MoE layer from torchtitan directly, not the full transformer.
3. `apply_moe_ep_tp` is a modified version of the torchtitan function, which just removes the for-loop over transformer blocks (since there is no transformer), but is otherwise identical.
4. I'm attempting to test something like a single-host dp2ep configuration. 

Is this correct/useful or do you recommend any adjustments?

tianyu-l: TP wrapping seems good.
FSDP + EP wrapping looks good for `MoE.experts`
FSDP wrapping for `MoE.router.gate` and `MoE.shared_expert` is not the same as dp2ep EP.

In dp2ep EP, we do the following, assuming using your "dp2 ep2 tp2" setup
apply FSDP on `MoE.experts` with dp2
apply EP on `MoE.experts` with ep2 (or EP+TP together with ep2 tp2)
apply FSDP on `MoE` with flattened "dp2 x ep2"

If you don't do this, what happens is you could have sharded `MoE.router.gate` and `MoE.shared_expert` by degree 4, but you only shard them with degree 2. This is may be fine for your test, just not identical to what I did.

You can take a look at how I did `apply_fsdp` and set up device mesh in `torchtitan/distributed/parallel_dims.py`

PR #2484, opened 2025-07-03
[mps] Add offsets to enable aoti

angelayi:
* #2501
* #2485
* __->__ #2484

In AOTI we allocate all weights in one blob and index into the blob based on the offsets. This caused an accuracy error when running the lowbit kernels since it assumes all the offsets are 0.


malfet (approved):

torchao/experimental/kernels/mps/src/lowbit.h
```diff
@@ -74,10 +74,15 @@ using DispatchFn =
 
 inline void linear_lowbit_quant_weights_mps_impl(
     id<MTLBuffer> a_buf,
+    size_t a_offset,
```
malfet: Wouldn't it be better to have like a tuple of buf + offset?
I.e. in function with 10+ arguments its easy to make a mistake about arg order

manuelcandales: +1 to this

angelayi: updated to using a pair!

manuelcandales (approved):

PR #2486, opened 2025-07-03
Add packing activation and packing weight.

szyszyzys:
Summary: Add packing for activation and weights for group wise low bit LUT based quantization kernel.

Reviewed By: metascroy

Differential Revision: D77312714


metascroy (approved):

metascroy:
LGTM!  You can ignore the failing mps test, I think it is broken on main, see https://github.com/pytorch/ao/issues/2487

PR #2499, opened 2025-07-07
Tutorial for benchmarking

jainapurva:
This pull request introduces comprehensive documentation updates for the TorchAO benchmarking framework

* **`docs/source/benchmarking_overview.md`**: Added a detailed tutorial on using the TorchAO benchmarking framework. It includes steps for adding APIs, model architectures, and CI dashboard integration, along with troubleshooting tips and best practices.

* **`docs/source/benchmarking_user_faq.md`**: Introduced a new FAQ section to address common benchmarking use cases, with a placeholder for future content.


docs/source/microbenchmarking.rst
```diff
@@ -0,0 +1,330 @@
...
+  # ... rest of existing code ...
+
+Now we can use this recipe throughout the benchmarking framework.
+
+.. note::
+
+  If the ``AOBaseConfig`` uses input parameters, like bit-width, group-size etc, you can pass them appended to the string config in input
+  For example, for ``GemliteUIntXWeightOnlyConfig`` we can pass it-width and group-size as ``gemlitewo-<bit_width>-<group_size>``
+
+2. Add a Model to Benchmarking Recipes
```
jerryzh168: for this one can people add huggingface models easily?

docs/source/microbenchmarking.rst
```diff
@@ -0,0 +1,330 @@
+Microbenchmarking Tutorial
+==========================
+
+This tutorial will guide you through using the TorchAO microbenchmarking framework. The tutorial contains different use cases for benchmarking your API and integrating with the dashboard.
+
+1. :ref:`Add an API to benchmarking recipes`
```
jerryzh168: nit: I think the order should be from most used to least used, also the doc feels more like a developer facing "API reference" which is centers around how to develop the benchmarking tool further (add new things etc.), instead of "using" the benchmarking tool.

Wondering if this could be structured in a way that is more user centered, that is structure this in terms of use cases / user flows:

* what is the most common use case, and what should be the flow / interaction with the tool?

and introduce what people need to do in different use cases.

one use case I have, is that I added a new tensor, e.g. for fbgemm, I'd like to know how fbgemm quant compare against existing quant on existing microbenchmarks or some models

e.g. how can the benchmarking tool simplify the changes needed to generate these tables: https://github.com/pytorch/ao/pull/2273 and https://github.com/pytorch/ao/pull/2276

Another could be for kernel developers who is writing cutlass or triton kernels, what are the pieces they need to interact with.

And we can then optimize / simplify each flow after we have that.

But maybe I'm talking about a different doc, this doc can be useful as a "API reference" as well

drisspg (commented):
One minor overall nit is that I think we should write docs in markdown as often as possible since it is much more of a lingua franca

docs/source/microbenchmarking.rst
jainapurva: Thanks @jerryzh168, this is a very helpful suggestion. There should be two docs, one for API reference, another should be for different use-cases, something like an FAQ or help doc

docs/source/microbenchmarking.rst
jainapurva: By adding model, if we mean adding architecture, or custom shapes for architectures, then they can. This is a microbenchmarking recipe, hence to add a model, we'll need to add it's full architecture (like llama), for generating lower level benchmarking numbers. For micro-benchmarking, it doesn't support specifying a hf-model name and importing it. That functionality can be included in future developments.

docs/source/benchmarking_user_faq.md
```diff
@@ -0,0 +1,142 @@
...
+## Run the performance benchmarking on your PR
+
+### 1. Add label to your PR
+To trigger the benchmarking CI workflow on your pull request, you need to add a specific label to your PR. Follow these steps:
+
+1. Go to your pull request on GitHub.
+2. On the right sidebar, find the "Labels" section.
+3. Click on the "Labels" dropdown and select "ciflow/benchmark" from the list of available labels.
+
+Adding this label will automatically trigger the benchmarking CI workflow for your pull request.
```
jerryzh168: can you add where can we see the results as well

jerryzh168: also will it run after each new commit is added or if the commit is updated?

docs/source/benchmarking_user_faq.md
```diff
@@ -0,0 +1,142 @@
...
+### 2. Manually trigger benchmarking workflow on your github branch
+To manually trigger the benchmarking workflow for your branch, follow these steps:
+
+1. Navigate to the "Actions" tab in your GitHub repository.
+2. Select the benchmarking workflow from the list of available workflows. For microbenchmarks, it's `Microbenchmarks-Perf-Nightly`.
+3. Click on the "Run workflow" button.
+4. In the dropdown menu, select the branch.
+5. Click the "Run workflow" button to start the benchmarking process.
+
+This will execute the benchmarking workflow on the specified branch, allowing you to evaluate the performance of your changes.
```
jerryzh168: what happens when people:

(1). push a new commit
(2). update a existing commit

?

jerryzh168: also when do we use (1) and when do we use (2)? seems like they are doing the same thing, if so maybe just keep one is enough

docs/source/benchmarking_user_faq.md
```diff
@@ -0,0 +1,142 @@
...
+```
+
+### Interpreting Results
+
+The benchmark results include:
+
+- **Speedup**: Performance improvement compared to baseline (bfloat16)
+- **Memory Usage**: Peak memory consumption during inference
+- **Latency**: Time taken for inference operations
+- **Profiling Data**: Detailed performance traces (when enabled)
```
jerryzh168: is there some duplicationg between these and L80-L84

docs/source/benchmarking_user_faq.md
```diff
@@ -0,0 +1,142 @@
...
+
+Results are saved in CSV format with columns for:
```
jerryzh168: nit: I think you can just show a small example output here

docs/source/benchmarking_user_faq.md
```diff
@@ -0,0 +1,142 @@
+# Benchmarking Use-Case FAQs
+
+This guide is intended to provide instructions for the most fequent benchmarking use-case. If you have any use-case that is not answered here, please create an issue here: [TorchAO Issues](https://github.com/pytorch/ao/issues)
+
+## Table of Contents
```
jerryzh168: this is closer to use case but still not really end use cases yet I feel, I think it might be helpful to describe scenarios like:
(1) integrating new quantization techniques (e.g. a new float8 technique)
people might be interested to understand overall performance, accuracy, memory footprint impact of the new technique and how it compared to the existing ones
In terms of commits, only last commit is important here I think

(2) kernel optimizations
people might be interested only on performance, and maybe trace is also important
they may put up a PR and will go through multiple commits, or just update a single commit multiple times, and they want to understand the performance differences between different commits

(3) performance regression tracking
what is the entry point for this one? will people receive an email when some threshold is passed?

(4) end users
is there a dashboard we can show end users for them to understand if they are using torchao, what speedup and accuracy drop they can expect on certain technique, device, for certain model architecture, etc.

docs/source/benchmarking_user_faq.md
jainapurva: If we add label, it'll run for every commit we add to the PR

docs/source/benchmarking_user_faq.md
jainapurva: This will run only for the latest change on the branch, it won't trigger automatically on every commit, for that we'll need to add label to the PR

jainapurva:
@jerryzh168 As discussed offline, I'll address your comments in the follow-up PR for the end-user tutorial  https://github.com/pytorch/ao/pull/2512

docs/source/benchmarking_user_faq.md
jainapurva: Addressed in PR: https://github.com/pytorch/ao/pull/2512

docs/source/benchmarking_user_faq.md
jainapurva: Will be addressed in PR https://github.com/pytorch/ao/pull/2512

docs/source/benchmarking_overview.md
```diff
@@ -0,0 +1,215 @@
+# Benchmarking Overview
```
jerryzh168: also `API Guide` might be more accurate

PR #2505, opened 2025-07-08
[CPU] Add support for dynamic float8 act float8 weight on CPU

Xia-Weiwen:
**Summary**
This PR adds support for dynamic float8 act float8 weight quantization on CPU.
It adds
- A new layout: `Float8DynamicActFloat8WeightCPULayout`
- Two new ops: `float8_linear_prepack_cpu` and `float8_linear_cpu`
- CPP kernels for the two new ops

The kernel computes FP8 GEMM with BF16 dtype.

**Test plan**
```
pytest test/quantization/test_dynamic_float8_linear_cpu.py
```

### DIFF
--- a/torchao/csrc/cpu/float8_linear.cpp
+++ b/torchao/csrc/cpu/float8_linear.cpp
@@ -1,0 +1,522 @@
+#include <torch/all.h>
+#include <ATen/cpu/vec/vec.h>
+#include <ATen/native/CPUBlas.h>
+#include <c10/util/Unroll.h>
+
+namespace torchao {
+
+namespace {
+
+#define BLOCK_N 32
+
+static bool cpublas_checked = false;
+static bool cpublas_can_pack = false;
+
+bool cpublas_could_pack() {
+  // the could_pack check requires AMX support implicitly
+  if (cpublas_checked) {
+    return cpublas_can_pack;
+  }
+  cpublas_can_pack = at::native::cpublas::could_pack(at::kBFloat16);
+  cpublas_checked = true;
+  return cpublas_can_pack;
+}
+
+/*
+return: packed_weight, packed_scales
+*/
+std::tuple<at::Tensor, at::Tensor>
+float8_linear_prepack_impl(
+    const at::Tensor& weight,
+    const at::Tensor& scales) {
+  // weight shape = [N, K]
+  // scales shape = [N, G]
+  TORCH_CHECK(weight.dim() == 2,
+              "Float8 linear CPU: Weight should be a 2D tensor for packing");
+  TORCH_CHECK(weight.size(1) % 2 == 0,
+              "Float8 linear CPU: Weight should have even number of columns for packing");
+
+  auto new_scales = scales;
+  if (new_scales.dim() == 1) {
+    new_scales.unsqueeze_(1);
+  }
+  new_scales = new_scales.to(at::kFloat);
+  int N = weight.size(0);
+  int K = weight.size(1);
+  int G = scales.size(1);
+  int group_size = K / G;
+  int block_k = group_size > 128 ? 128 : group_size;
+  while (K % block_k != 0) {
+    block_k /= 2;
+  }
+  TORCH_CHECK(block_k > 0 && block_k <= group_size,
+              "Float8 linear CPU: Invalid block_k size, should be in (0, group_size]");
+  constexpr int block_n = BLOCK_N;
+  int Nc = N / block_n;
+  int Kc = K / block_k;
+
+  // Reorder weight to [N/block_n, K/block_k, block_k, block_n]
+  // Reorder scales to [N/block_n, G, block_n]
+  auto weight_view = weight.view({Nc, block_n, Kc, block_k});
+  at::Tensor weight_reordered = weight_view.permute({0, 2, 3, 1}).contiguous();
+  at::Tensor blocked_weight;
+  at::Tensor blocked_scales = new_scales.view({Nc, block_n, G}).permute({0, 2, 1}).contiguous();
+
+#if defined(CPU_CAPABILITY_AVX512)
+  if (cpublas_could_pack()) {
+    constexpr int vnni_size = 2; // for float16
+    blocked_weight = at::empty({Nc, Kc, block_k, block_n}, weight.options());
+    auto weight_ptr = reinterpret_cast<uint8_t*>(weight_reordered.data_ptr());
+    auto blocked_weight_ptr = reinterpret_cast<uint8_t*>(blocked_weight.data_ptr());
+    int64_t num_blocks = Nc * Kc;
+    at::parallel_for(0, num_blocks, 1, [&](int64_t begin, int64_t end) {
+      for (const auto i : c10::irange(begin, end)) {
+        auto in_ptr = weight_ptr + i * block_k * block_n;
+        auto out_ptr = blocked_weight_ptr + i * block_k * block_n;
+
+        // Reorder weight block to VNNI
+        // plain shape = [block_k, block_n]
+        // packed shape = [block_k / VNNI_SIZE, block_n, VNNI_SIZE] viewed as [block_k, block_n]
+        constexpr int n_group_size = 8;
+        constexpr int n_group = block_n / n_group_size; // 4
+        for (int nb = 0; nb < n_group; ++nb) {
+          for (int k = 0; k < block_k; k += vnni_size) {
+            for (int ni = 0; ni < n_group_size; ++ni) {
+              for (int ki = 0; ki < vnni_size; ++ki) {
+                int src_idx = nb * n_group_size + ni + (k + ki) * block_n;
+                int dst_idx = (nb * n_group_size + ni) * vnni_size + k * block_n + ki;
+                *(out_ptr + dst_idx) = *(in_ptr + src_idx);
+              }
+            }
+          }
+        }
+      }
+    });
+  } else
+#endif
+  {
+    blocked_weight = weight_reordered;
+  }
+
+  return std::make_tuple(std::move(blocked_weight), std::move(blocked_scales));
+}
+
+#if defined(CPU_CAPABILITY_AVX512)
+alignas(64) static uint16_t e4m3_to_16bit[256];
+
+template <typename T>
+static void initialize_e4m3_to_16bit_tables() {
+  // run only once
+  static bool initialized_16bit = false;
+  if (!initialized_16bit) {
+    for (uint8_t u8 = 0; u8 < 256; ++u8) {
+      auto value = static_cast<T>(c10::bit_cast<at::Float8_e4m3fn>(u8));
+      uint16_t value_bits = c10::bit_cast<uint16_t>(value);
+      e4m3_to_16bit[u8] = value_bits;
+      if (u8 == 255) {
+        break;
+      }
+    }
+    initialized_16bit = true;
+  }
+}
+
+template <typename T>
+static void cvt_e4m3_16bit_intrinsic_lut(
+    const at::Float8_e4m3fn* __restrict__ in,
+    T* out,
+    int64_t len) {
+  for (size_t i = 0; i < len; i += 64) {
+    __m512i fp8_vec = _mm512_loadu_si512((__m512i*)&in[i]);
+    __m128i group0 = _mm512_castsi512_si128(fp8_vec);
+    __m128i group1 = _mm512_extracti32x4_epi32(fp8_vec, 1);
+    __m128i group2 = _mm512_extracti32x4_epi32(fp8_vec, 2);
+    __m128i group3 = _mm512_extracti32x4_epi32(fp8_vec, 3);
+
+    __m512i indices0 = _mm512_cvtepu8_epi32(group0);
+    __m512i indices1 = _mm512_cvtepu8_epi32(group1);
+    __m512i indices2 = _mm512_cvtepu8_epi32(group2);
+    __m512i indices3 = _mm512_cvtepu8_epi32(group3);
+
+    // Gather BF16 conversion results from the lookup table.
+    __m512i bf16_i32_vec0 = _mm512_i32gather_epi32(indices0, e4m3_to_16bit, 2);
+    __m512i bf16_i32_vec1 = _mm512_i32gather_epi32(indices1, e4m3_to_16bit, 2);
+    __m512i bf16_i32_vec2 = _mm512_i32gather_epi32(indices2, e4m3_to_16bit, 2);
+    __m512i bf16_i32_vec3 = _mm512_i32gather_epi32(indices3, e4m3_to_16bit, 2);
+
+    // Helper lambda: Convert 16 32-bit ints (in a __m512i) to 16 16-bit ints.
+    auto convert_32_to_16 = [](__m512i vec) -> __m256i {
+      return _mm512_cvtepi32_epi16(vec);
+    };
+
+    __m256i bf16_i16_vec0 = convert_32_to_16(bf16_i32_vec0);
+    __m256i bf16_i16_vec1 = convert_32_to_16(bf16_i32_vec1);
+    __m256i bf16_i16_vec2 = convert_32_to_16(bf16_i32_vec2);
+    __m256i bf16_i16_vec3 = convert_32_to_16(bf16_i32_vec3);
+
+    _mm256_storeu_si256((__m256i*)(out + i + 0), bf16_i16_vec0);
+    _mm256_storeu_si256((__m256i*)(out + i + 16), bf16_i16_vec1);
+    _mm256_storeu_si256((__m256i*)(out + i + 32), bf16_i16_vec2);
+    _mm256_storeu_si256((__m256i*)(out + i + 48), bf16_i16_vec3);
+  }
+}
+
+static void _convert_B_to_bf16(
+    const at::Float8_e4m3fn* __restrict__ B,
+    at::BFloat16* dqB,
+    int64_t len) {
+  initialize_e4m3_to_16bit_tables<at::BFloat16>();
+  int tail = len % 64;
+  cvt_e4m3_16bit_intrinsic_lut<at::BFloat16>(B, dqB, len - tail);
+  for (int i = len - tail; i < len; ++i) {
+    dqB[i] = (at::BFloat16)B[i];
+  }
+}
+
+static void _convert_A_to_bf16(
+    const at::Float8_e4m3fn* __restrict__ A,
+    at::BFloat16* dqA,
+    int64_t M,
+    int64_t K,
+    int64_t lda) {
+  initialize_e4m3_to_16bit_tables<at::BFloat16>();
+  for (int m = 0; m < M; ++m) {
+    int tail = K % 64;
+    int body = K - tail;
+    cvt_e4m3_16bit_intrinsic_lut<at::BFloat16>(A + m * lda, dqA + m * K, body);
+    for (int k = body; k < K; ++k) {
+      dqA[m * K + k] = (at::BFloat16)A[m * lda + k];
+    }
+  }
+}
+
+template <bool accum, int64_t N>
+static void _dequant_and_store(
+    float* __restrict__ output,
+    const float* __restrict__ input,
+    const float* __restrict__ scale_a,
+    const float* __restrict__ scale_b,
+    int M,
+    int ldi,
+    int ldo,
+    int ldsa = 1) {
+  for (int m = 0; m < M; ++m) {
+    float a_scale = *(scale_a + m * ldsa);
+    __m512 va_scale = _mm512_set1_ps(a_scale);
+    int n = 0;
+#pragma GCC unroll 2
+    for (; n < N; n += 16) {
+      __m512 vc_f = _mm512_loadu_ps(input + m * ldi + n);
+      __m512 vc_f_mul = _mm512_mul_ps(vc_f, va_scale);
+      __m512 vb_s = _mm512_loadu_ps(scale_b + n);
+      vc_f_mul = _mm512_mul_ps(vc_f_mul, vb_s);
+      if constexpr (accum) {
+        __m512 vo = _mm512_loadu_ps(output + m * ldo + n);
+        _mm512_storeu_ps(output + m * ldo + n, _mm512_add_ps(vo, vc_f_mul));
+      } else {
+        _mm512_storeu_ps(output + m * ldo + n, vc_f_mul);
+      }
+    }
+    for (; n < N; ++n) {
+      float dq_val = input[m * ldi + n] * a_scale * scale_b[n];
+      if constexpr (accum) {
+        output[m * ldo + n] += dq_val;
+      } else {
+        output[m * ldo + n] = dq_val;
+      }
+    }
+  }
+}
+
+#else
+static void _convert_B_to_bf16(
+    const at::Float8_e4m3fn* B,
+    at::BFloat16* dqB,
+    int64_t len) {
+  for (int i = 0; i < len; ++i) {
+    dqB[i] = (at::BFloat16)B[i];
+  }
+}
+
+static void _convert_A_to_bf16(
+    const at::Float8_e4m3fn* __restrict__ A,
+    at::BFloat16* dqA,
+    int64_t M,
+    int64_t K,
+    int64_t lda) {
+  for (int m = 0; m < M; ++m) {
+    for (int k = 0; k < K; ++k) {
+      dqA[m * K + k] = (at::BFloat16)A[m * lda + k];
+    }
+  }
+}
+#endif
+
+template <bool cpublas_can_pack, int64_t N>
+void _dequant_gemm_accum(
+    float* C,
+    const at::Float8_e4m3fn* A,
+    const float* scales_a,
+    const at::Float8_e4m3fn* B,
+    const float* scales_b,
+    int64_t M,
+    int64_t K,
+    int64_t lda,
+    int64_t ldc) {
+  // Compute GEMM fp8 * fp8 -> fp32
+  // Then apply scales and store results
+  at::BFloat16 dqB[K * N];
+  _convert_B_to_bf16(B, dqB, K * N);
+  at::BFloat16 dqA[M * K];
+  _convert_A_to_bf16(A, dqA, M, K, lda);
+#if defined(CPU_CAPABILITY_AVX512)
+  if constexpr (cpublas_can_pack) {
+    float C_f32[M * N];
+    at::native::cpublas::brgemm(
+        M,
+        N,
+        K,
+        K /*lda*/,
+        N /*ldb*/,
+        N /*ldc*/,
+        false /* add_C */,
+        dqA,
+        dqB,
+        C_f32,
+        true /* is_vnni */);
+    _mm_prefetch(B + N * K, _MM_HINT_T0);
+    _mm_prefetch(A + K, _MM_HINT_T0);
+    _dequant_and_store<true, N>(
+        C,
+        C_f32,
+        scales_a,
+        scales_b,
+        M,
+        N /*ldi*/,
+        ldc,
+        1 /*ldsa*/);
+  } else
+#endif
+  {
+    for (int64_t i = 0; i < M; ++i) {
+      for (int64_t j = 0; j < N; ++j) {
+        float sum = 0;
+        for (int64_t k = 0; k < K; ++k) {
+          sum += ((float)dqA[i * K + k] * dqB[k * N + j]);
+        }
+        C[i * ldc + j] += sum * scales_a[i] * scales_b[j];
+      }
+    }
+  }
+}
+
+template<int64_t N>
+inline void copy_bias(const float* bias_ptr, float* y_buf, int64_t m) {
+  if (bias_ptr) {
+    for (int i = 0; i < m; ++i) {
+      int j = 0;
+#if defined(CPU_CAPABILITY_AVX512)
+#pragma GCC unroll 2
+      for (; j < N; j += 16) {
+        __m512 bias_vec = _mm512_loadu_ps(bias_ptr + j);
+        _mm512_storeu_ps(y_buf + i * N + j, bias_vec);
+      }
+#endif
+      for (; j < N; ++j) {
+        y_buf[i * N + j] = bias_ptr[j];
+      }
+    }
+  } else { // initialize to zero
+    for (int i = 0; i < m; ++i) {
+      int j = 0;
+#if defined(CPU_CAPABILITY_AVX512)
+#pragma GCC unroll 2
+      for (; j < N; j += 16) {
+        __m512 zero_vec = _mm512_setzero_ps();
+        _mm512_storeu_ps(y_buf + i * N + j, zero_vec);
+      }
+#endif
+      for (; j < N; ++j) {
+        y_buf[i * N + j] = 0;
+      }
+    }
+  }
+}
+
+template<typename out_dtype, int64_t N>
+inline void store_out(const float* y_buf, out_dtype* c_ptr, int64_t m, /* int64_t n, */ int64_t lda) {
+  for (int i = 0; i < m; ++i) {
+    int j = 0;
+    if constexpr (std::is_same<out_dtype, float>::value) {
+#if defined(CPU_CAPABILITY_AVX512)
+#pragma GCC unroll 2
+      for (; j < N; j += 16) {
+        __m512 y_vec = _mm512_loadu_ps(y_buf + i * N + j);
+        _mm512_storeu_ps(c_ptr + i * lda + j, y_vec);
+      }
+#endif
+      for (; j < N; ++j) {
+        c_ptr[i * lda + j] = y_buf[i * N + j];
+      }
+    } else if constexpr (std::is_same<out_dtype, at::BFloat16>::value) {
+#if defined(CPU_CAPABILITY_AVX512)
+#pragma GCC unroll 2
+      for (; j < N; j += 16) {
+        __m512 y_vec = _mm512_loadu_ps(y_buf + i * N + j);
+        __m256i y_bf16_vec = at::vec::cvtfp32_bf16(y_vec);
+        _mm256_storeu_si256(reinterpret_cast<__m256i*>(c_ptr + i * lda + j), y_bf16_vec);
+      }
+#endif
+      for (; j < N; ++j) {
+        c_ptr[i * lda + j] = at::BFloat16(y_buf[i * N + j]);
+      }
+    } else if constexpr (std::is_same<out_dtype, at::Half>::value) {
+#if defined(CPU_CAPABILITY_AVX512)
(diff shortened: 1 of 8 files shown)


Xia-Weiwen:
Hi @chunyuan-w @mingfeima Could you please review this PR? Thanks.

chunyuan-w:
Should we move the conversion vec code to this file? https://github.com/pytorch/pytorch/blob/cd995bfb2aac8891465809be3ce29543bd524287/aten/src/ATen/cpu/vec/vec512/vec512_float8.h

Similar to this PR: https://github.com/pytorch/pytorch/pull/152417

Xia-Weiwen:
> Should we move the conversion vec code to this file? https://github.com/pytorch/pytorch/blob/cd995bfb2aac8891465809be3ce29543bd524287/aten/src/ATen/cpu/vec/vec512/vec512_float8.h
> 
> Similar to this PR: [pytorch/pytorch#152417](https://github.com/pytorch/pytorch/pull/152417)

Thanks for the comment. If we move it to PyTorch, a problem might be that we need to check if the function is available at compile time. We may do it step by step, and for now it might be better that we keep it here.

torchao/csrc/cpu/float8_linear.cpp
```diff
@@ -0,0 +1,522 @@
...
+      return 128;
+    }
+  }();
+  int64_t Mc = (M + block_m - 1) / block_m;
+  bool parallel_on_M = M > 128;
+  int64_t num_blocks = parallel_on_M ? Mc * Nc : Nc;
+
+  // scales shape = [Nc, G, block_n]
+  int64_t num_groups = weight_scales.size(1);
+  int64_t group_size = K / num_groups;
```
chunyuan-w: Do we support the case where `K % num_groups != 0`?

Xia-Weiwen: We don't support it. It is guarded by the quantization utility in Torchao, such as https://github.com/pytorch/ao/blob/aee079503e6d882e798fd58a42780f5b98ae2126/torchao/quantization/quant_primitives.py#L293
I have also added a `TORCH_CHECK` here. Thanks.

jerryzh168:
1. I think we want to add this with the new design
2. I feel this should probably be added to prototype first, and if we get wider adoptions we can move to official API

jerryzh168:
please move to the new design, and also put this prototype/quantization/ folder?

torchao/quantization/quantize_/workflows/float8/float8_opaque_tensor.py
```diff
@@ -0,0 +1,227 @@
...
+
+class Float8OpaqueTensor(TorchAOBaseTensor):
+    """
+    Float8 dynamic activation float8 weight on CPU. The weight tensor is reordered to a blocked layout
+    for better memory locality from [N, K] to [N/block_n, K/block_k, block_k, block_n], where block_n = 32
+    and block_k depends on group-size for quantization (=32/64/128). And the innermost block with shape
+    [block_k, block_n] may be further reordered to VNNI layout depending on supported CPU ISA.
+
+    Tensor Attributes:
+        qdata: Reordered float8 weight on CPU with shape = [N/block_n, K/block_k, block_k, block_n].
```
jerryzh168, line 41: is this really opaque? oh OK, it's because "And the innermost block with shape
    [block_k, block_n] may be further reordered to VNNI layout depending on supported CPU ISA."?

Xia-Weiwen: Yes, that's right.

test/quantization/quantize_/workflows/float8/test_float8_opaque_tensor.py
```diff
@@ -0,0 +1,259 @@
...
+    )
+    @unittest.skipIf(not torch_version_at_least("2.6.0"), "Test only enabled for 2.6+")
+    @common_utils.parametrize("dtype", [torch.float, torch.bfloat16, torch.half])
+    @common_utils.parametrize("x_dim", [2, 3])
+    @common_utils.parametrize("bias", [True, False])
+    def test_dynamic_float8_linear_ref_cpu(self, dtype, x_dim, bias):
+        device = "cpu"
+        # the shape is not supported by cpp kernel, so the ref path will be used.
+        m = ToyLinearModel(120, 120, bias=bias).eval().to(dtype).to(device)
+        bs = 4
```
mingfeima: use `common_utils.parametrize`

test/quantization/quantize_/workflows/float8/test_float8_opaque_tensor.py
```diff
@@ -0,0 +1,259 @@
...
+        "CPU" not in torch._C._dispatch_dump("torchao::float8_linear_cpu"),
+        reason="cpp kernels not built",
+    )
+    @unittest.skipIf(not torch_version_at_least("2.6.0"), "Test only enabled for 2.6+")
+    @common_utils.parametrize("dtype", [torch.float, torch.bfloat16, torch.half])
+    @common_utils.parametrize("x_dim", [2, 3])
+    @common_utils.parametrize("bias", [True, False])
+    @common_utils.parametrize("bs", [1, 160])
+    def test_dynamic_float8_linear_cpu(self, dtype, x_dim, bias, bs):
+        device = "cpu"
```
mingfeima, line 65: does `torch.ao` has test cases for cuda right now?

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
```diff
@@ -0,0 +1,618 @@
...
+  at::Tensor blocked_weight;
+  at::Tensor blocked_scales = is_per_tensor ? new_scales.view({1}) : new_scales.view({Nc, block_n, G}).permute({0, 2, 1}).contiguous();
+
+#if defined(CPU_CAPABILITY_AVX512)
+  if (cpublas_could_pack()) {
+#ifdef CPUBLAS_BRGEMM_F8F8F32
+    constexpr int vnni_size = 4; // for fp8
+#else
+    constexpr int vnni_size = 2; // for float16
+#endif
```
mingfeima: try to abstract common logic to util or common header with template speficfication

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
```diff
@@ -0,0 +1,618 @@
...
+      if (u8 == 255) {
+        break;
+      }
+    }
+    initialized_16bit = true;
+  }
+}
+
+template <typename T>
+static void cvt_e4m3_16bit_intrinsic_lut(
```
mingfeima: this LUT implenetation is roughly 2x slower than the approach we selected on sglang. You can contact Yanbing for details.

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
```diff
@@ -0,0 +1,618 @@
...
+    // Gather BF16 conversion results from the lookup table.
+    __m512i bf16_i32_vec0 = _mm512_i32gather_epi32(indices0, e4m3_to_16bit, 2);
+    __m512i bf16_i32_vec1 = _mm512_i32gather_epi32(indices1, e4m3_to_16bit, 2);
+    __m512i bf16_i32_vec2 = _mm512_i32gather_epi32(indices2, e4m3_to_16bit, 2);
+    __m512i bf16_i32_vec3 = _mm512_i32gather_epi32(indices3, e4m3_to_16bit, 2);
+
+    // Helper lambda: Convert 16 32-bit ints (in a __m512i) to 16 16-bit ints.
+    auto convert_32_to_16 = [](__m512i vec) -> __m256i {
+      return _mm512_cvtepi32_epi16(vec);
+    };
```
mingfeima: i seldom use lambda in intrisics, this may lead to register spill if it is called in a unrolled function.

use a MACRO instead.

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
```diff
@@ -0,0 +1,618 @@
...
+
+    __m256i bf16_i16_vec0 = convert_32_to_16(bf16_i32_vec0);
+    __m256i bf16_i16_vec1 = convert_32_to_16(bf16_i32_vec1);
+    __m256i bf16_i16_vec2 = convert_32_to_16(bf16_i32_vec2);
+    __m256i bf16_i16_vec3 = convert_32_to_16(bf16_i32_vec3);
+
+    _mm256_storeu_si256((__m256i*)(out + i + 0), bf16_i16_vec0);
+    _mm256_storeu_si256((__m256i*)(out + i + 16), bf16_i16_vec1);
+    _mm256_storeu_si256((__m256i*)(out + i + 32), bf16_i16_vec2);
+    _mm256_storeu_si256((__m256i*)(out + i + 48), bf16_i16_vec3);
```
mingfeima: even with LUT, try to write back with __m512i, use a marco:
```
#define M256_INSERT_M512(lo, hi) \
    _mm512_inserti64x4(_mm512_castsi256_si512(lo), hi, 1)

```

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
```diff
@@ -0,0 +1,618 @@
...
+  initialize_e4m3_to_16bit_tables<at::BFloat16>();
+  for (int m = 0; m < M; ++m) {
+    int tail = K % 64;
+    int body = K - tail;
+    cvt_e4m3_16bit_intrinsic_lut<at::BFloat16>(A + m * lda, dqA + m * K, body);
+    for (int k = body; k < K; ++k) {
+      dqA[m * K + k] = (at::BFloat16)A[m * lda + k];
+    }
+  }
+}
```
mingfeima: can we combine these two function into one, seems duplicated to me.

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
```diff
@@ -0,0 +1,618 @@
...
+  float a_scale, b_scale;
+  __m512 va_scale;
+  __m512 vb_scale;
+  for (int m = 0; m < M; ++m) {
+    if constexpr (act_quant_mode == PER_GROUP) {
+      a_scale = *(scale_a + m * ldsa);
+      va_scale = _mm512_set1_ps(a_scale);
+    }
+    int n = 0;
+#pragma GCC unroll 2
```
mingfeima: i usually don't reply on compiler unroll.

use c10::UNROLL instead. for storing back to C, which is MxN, unroll in both two dimensions.

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
```diff
@@ -0,0 +1,618 @@
...
+        N /*ldb*/,
+        N /*ldc*/,
+        false /* add_C */,
+        dqA,
+        dqB,
+        C_f32,
+        true /* is_vnni */);
+#endif
+    _mm_prefetch(B + N * K, _MM_HINT_T0);
+    _mm_prefetch(A + K, _MM_HINT_T0);
```
mingfeima: is this trying to prefetch the next block? won't be too affective.

we do prefetch on K dimension with distance 64 or 128, this should be done by brgemm, we won't need this prefetch.

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
```diff
@@ -0,0 +1,618 @@
...
+inline void store_out(
+    const float* y_buf,
+    out_dtype* c_ptr,
+    int64_t M,
+    int64_t lda,
+    const float* scales_a,
+    const float* scales_b,
+    const float* bias) {
+  float a_scale = 1.0, b_scale = 1.0;
+#if defined(CPU_CAPABILITY_AVX512)
```
mingfeima: mingle #if defined(CPU_CAPABILITY_AVX512) inside function at such level would be hard to read.

i prefer more human readable way:

```
#if defined(CPU_CAPABILITY_AVX512)
template<typename out_dtype, int64_t N, int act_quant_mode, int wei_quant_mode>
inline void store_out(...)
#else
template<typename out_dtype, int64_t N, int act_quant_mode, int wei_quant_mode>
inline void store_out(...)
#endif
```

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
```diff
@@ -0,0 +1,618 @@
...
+      return 32;
+    } else if (M < 96) {
+      return 64;
+    } else {
+      return 128;
+    }
+  }();
+  int64_t Mc = (M + block_m - 1) / block_m;
+  bool parallel_on_M = M > 128;
+  int64_t num_blocks = parallel_on_M ? Mc * Nc : Nc;
```
mingfeima: this is deciding BLOCK_M size and parallel strategy, should be common logic, move it to header.

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
```diff
@@ -0,0 +1,618 @@
...
+
+  const at::Float8_e4m3fn* a_ptr = input_view.data_ptr<at::Float8_e4m3fn>();
+  const float* a_scales_ptr = input_scales.data_ptr<float>();
+  const at::Float8_e4m3fn* b_ptr = weight.data_ptr<at::Float8_e4m3fn>();
+  const float* b_scales_ptr = weight_scales.data_ptr<float>();
+  out_dtype* c_ptr = output.data_ptr<out_dtype>();
+  const float* bias_ptr = bias.has_value() ? bias.value().data_ptr<float>() : nullptr;
+
+  at::parallel_for(0, num_blocks, 1, [&](int64_t begin, int64_t end) {
+    auto y_buf = new (std::align_val_t(8)) float[block_m * block_n];
```
mingfeima: block_m and block_n are not constexpr, yes compiler allows you to do it now, but i preferred to use a dispatch:

```
#define AT_DISPATCH_GEMM_BLOCK_SIZES(...)
```

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
```diff
@@ -0,0 +1,618 @@
...
+    for (const auto i : c10::irange(begin, end)) {
+      int64_t mc = parallel_on_M ? i / Nc : 0;
+      int64_t nc = parallel_on_M ? i % Nc : i;
+      int64_t mc_end = parallel_on_M ? mc + 1 : Mc;
```
mingfeima: use `data_index_init` and `data_index_step` to get rid of integer div which is slow.

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
```diff
@@ -0,0 +1,618 @@
...
+
+      for (int mci = mc; mci < mc_end; ++mci) {
+        int64_t m_size = mci * block_m + block_m > M ? M - mci * block_m : block_m;
+        memset(y_buf, 0, sizeof(float) * m_size * block_n);
```
mingfeima: `memset` is avx2. use at::vec::map do this job.

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
```diff
@@ -0,0 +1,618 @@
...
+            y_buf /*C*/,
+            a_ptr + mci * block_m * K + kci * block_k /*A*/,
+            scales_a /*scales_a*/,
+            b_ptr + (nc * Kc + kci) * block_n * block_k /*B*/,
+            scales_b /*scales_b*/,
+            m_size /*M*/,
+            block_k /*K*/,
+            K /*lda*/,
+            block_n /*ldc*/,
+            ldsa /*ldsa*/);
```
mingfeima:
```suggestion
_micro_gemm<cpublas_can_pack, block_n, act_quant_mode, wei_quant_mode>(
    /* C */        y_buf,
    /* A */        a_ptr + mci * block_m * K + kci * block_k,
    /* scales_a */ scales_a,
    /* B */        b_ptr + (nc * Kc + kci) * block_n * block_k,
    /* scales_b */ scales_b,
    /* M */        m_size,
    /* K */        block_k,
    /* lda */      K,
    /* ldc */      block_n,
    /* ldsa */     ldsa);
```

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
mingfeima: allocate temp buffer with `at::empty()` which has been optimized with cpu allocator, outside the parallel loop and divide into threads by yourself

`new` should not appear inside omp loop, this is slow function.

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
```diff
@@ -0,0 +1,618 @@
...
+  int wei_quant_mode = weight_scales.numel() == 1 ? PER_TENSOR :
+                       weight_scales.numel() == N ? PER_ROW :
+                       PER_GROUP;
+  // Case to fall back
+  if (weight.dim() == 2) {
+    TORCH_CHECK(act_quant_mode != PER_GROUP && wei_quant_mode != PER_GROUP,
+      "FP8 linear: Per-group quantization is not supported in the fallback path");
+    auto y_fp32 = at::linear(input.to(at::kFloat).mul(input_scales), weight.to(at::kFloat).mul(weight_scales), bias);
+    return y_fp32.to(output_dtype);
+  }
```
mingfeima: we don't care this fallback path performance, right?

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
```diff
@@ -0,0 +1,618 @@
...
+                       input_scales.numel() == input.numel() / input.size(-1) ? PER_ROW :
+                       PER_GROUP;
+  int wei_quant_mode = weight_scales.numel() == 1 ? PER_TENSOR :
+                       weight_scales.numel() == N ? PER_ROW :
+                       PER_GROUP;
+  // Case to fall back
+  if (weight.dim() == 2) {
+    TORCH_CHECK(act_quant_mode != PER_GROUP && wei_quant_mode != PER_GROUP,
+      "FP8 linear: Per-group quantization is not supported in the fallback path");
+    auto y_fp32 = at::linear(input.to(at::kFloat).mul(input_scales), weight.to(at::kFloat).mul(weight_scales), bias);
```
mingfeima:
```suggestion
    auto y_fp32 = at::linear(input.to(at::kFloat).mul_(input_scales), weight.to(at::kFloat).mul_(weight_scales), bias);
```

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
```diff
@@ -0,0 +1,618 @@
...
+      std::tuple<
+          /*output_dtype*/ at::ScalarType,
+          /*cpublas_can_pack*/ bool,
+          /*act_quant_mode*/ int,
+          /*wei_quant_mode*/ int>,
+      std::tuple<
+          enumerate_dispatcher<at::ScalarType, at::ScalarType::Float, at::ScalarType::BFloat16, at::ScalarType::Half>,
+          enumerate_dispatcher<bool, false, true>,
+          enumerate_dispatcher<int, PER_TENSOR, PER_ROW, PER_GROUP>,
+          enumerate_dispatcher<int, PER_TENSOR, PER_ROW, PER_GROUP>>>::
```
mingfeima: honestly this really is not human readable. And do we have scenarios where  A and B have different quant_mode? Or a_quant_mode and b_quant_mode exists in certain combinations?

prefer old fashioned macro dispatch:
```
AT_DISPATCH_FLOATINF_TYPES(
  AT_DISTACH_BOOL_VALUES(
    AT_DISPATCH_QUANT_MODE(
       ...
    )
  )
)
```

test/quantization/quantize_/workflows/float8/test_float8_opaque_tensor.py
Xia-Weiwen: These test cases are for CPU only. CUDA has its own implementation and test cases.

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
Xia-Weiwen: I think this is not very common. It's for fp8 kernel only.

mingfeima: something like:
```C++
template <typename T> struct vnni_traits;
template <> struct vnni_traits<at::Half> { static constexpr int size = 2; };
template <> struct vnni_traits<at::BFloat16> { static constexpr int size = 2; };
template <> struct vnni_traits<at::float8_e4m3> { static constexpr int size = 4; };
template <> struct vnni_traits<at::float8_e5m2> { static constexpr int size = 4; };
template <> struct vnni_traits<int8_t> { static constexpr int size = 4; };

template <typename T> constexpr int get_vnni_size() { return vnni_traits<T>::size; }
```

used by
```C++
constexpr int vnni_size = get_vnni_size<T>();
```

test/quantization/quantize_/workflows/float8/test_float8_opaque_tensor.py
Xia-Weiwen: Thanks. Updated.

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
Xia-Weiwen: Thanks. Added.

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
Xia-Weiwen: Thanks. Updated.

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
Xia-Weiwen: Thanks. This code is removed and now we use the new fp8->bf16 code.

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
Xia-Weiwen: Thanks. This code is removed and now we use the new fp8->bf16 code.

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
Xia-Weiwen: Thanks. Updated.

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
Xia-Weiwen: Thanks. Updated.

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
Xia-Weiwen: Thanks. Updated.

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
Xia-Weiwen: Thanks. Updated.

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
Xia-Weiwen: Thanks. Moved to utils.h.

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
Xia-Weiwen: Thanks. Updated. Now we are using at::empty.

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
Xia-Weiwen: Thanks. Updated.

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
Xia-Weiwen: Thanks. I studied `at:;vec::map` and found that it read an input buffer and writes an output buffer, which seems inefficient for setting buffers to zero. I added a util function with AVX512 to zero a buffer.

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
Xia-Weiwen: Thanks. Updated.

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
Xia-Weiwen: Thanks. Now we use `at::empty`.

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
Xia-Weiwen: That's right.

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
Xia-Weiwen: Thanks. Updated.

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
Xia-Weiwen: Thanks. I have added a few macros for dispatch in utils.h.

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
```diff
@@ -0,0 +1,584 @@
...
+  at::Tensor blocked_weight;
+  at::Tensor blocked_scales = is_per_tensor ? new_scales.view({1}) : new_scales.view({Nc, block_n, G}).permute({0, 2, 1}).contiguous();
+
+#if defined(CPU_CAPABILITY_AVX512)
+  if (cpublas_could_pack()) {
+#ifdef CPUBLAS_BRGEMM_F8F8F32
+    constexpr int vnni_size = get_vnni_size<at::Float8_e4m3fn>(); // for fp8
+#else
+    constexpr int vnni_size = get_vnni_size<at::BFloat16>(); // for bfloat16
+#endif
```
mingfeima, line 89: looks better now~

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
```diff
@@ -0,0 +1,584 @@
...
+    int64_t ldsa) {
+  // If FP8 brgemm is not available, convert A/B to bf16 for computation
+  // Compute GEMM fp8 * fp8 -> fp32 (or bf16 * bf16 -> fp32)
+  // If per_group quant, apply scales. Otherwise, don't apply scales here
+  // Finally accumulate and store results
+#ifndef CPUBLAS_BRGEMM_F8F8F32
+  at::BFloat16 dqB[K * N];
+  cvt_f8e4m3_to_bf16(B, dqB, K, N, N);
+  at::BFloat16 dqA[M * K];
+  cvt_f8e4m3_to_bf16(A, dqA, M, K, lda);
```
mingfeima: try not to use dynamic memory allocation for arrays.

use `at::empty()` to allocate memory.

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
```diff
@@ -0,0 +1,584 @@
...
+#endif
+#if defined(CPU_CAPABILITY_AVX512)
+  if constexpr (cpublas_can_pack) {
+    float C_f32[M * N];
```
mingfeima: same as before, for example, you can do
```
at::Tensor buffer = at::empty({K * N + M * K + M * N * 2}, at::kBFloat16);
at::BFloat16* qdB = buffer.data_ptr<at::BFloat16>();
at::BFloat16* qdA = qdB + K * N;
float* C = reinterpret_cast<float>(qdA + M * K);
```

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
Xia-Weiwen: Thanks. Now these buffers are allocated outside the microkernel.

torchao/csrc/cpu/aten_kernels/float8_linear.cpp
Xia-Weiwen: Thanks. Now these buffers are allocated outside the microkernel.

Xia-Weiwen:
Hi @jerryzh168 Could you please review this PR again? Thanks.

Xia-Weiwen:
Hi @jerryzh168 Could you please review this PR again? Thanks.

torchao/quantization/observer.py
```diff
@@ -78,8 +80,13 @@ def get_block_size(
         block_size = list(input_shape)
         block_size[granularity.axis] = 1
         return tuple(block_size)
-    elif isinstance(granularity, PerRow):
+    elif isinstance(granularity, (PerRow, PerToken)):
```
jerryzh168, line 83: maybe create a separate PR to merge these two: https://github.com/pytorch/ao/blob/18dbe875a0ce279739dda06fda656e76845acaac/torchao/quantization/pt2e/observer.py#L1783?

and move it to `torchao/quantization/utils` ?

jerryzh168: seems like the pt2e one is not used that much, so should be relatively easy to remove

torchao/quantization/qat/fake_quantize_config.py
```diff
@@ -423,6 +423,11 @@ def _infer_fake_quantize_configs(
         (act_granularity, weight_granularity) = _normalize_granularity(
             base_config.granularity
         )
+        assert act_granularity == weight_granularity and isinstance(
```
jerryzh168, line 425: split fake quant changes to a separate file?

torchao/quantization/quant_api.py
```diff
@@ -1739,6 +1762,7 @@ class Float8DynamicActivationFloat8WeightConfig(AOBaseConfig):
     kernel_preference: KernelPreference = KernelPreference.AUTO
     set_inductor_config: bool = True
     version: int = 2
+    packing_format: Float8PackingFormat = Float8PackingFormat.PLAIN
```
jerryzh168: nit: `packing_format` --> `float8_packing_format`

torchao/quantization/observer.py
Xia-Weiwen: Sure. I will have another PR to do it.

torchao/quantization/qat/fake_quantize_config.py
Xia-Weiwen: Hi @jerryzh168 This file is about fake quant. Do you mean a separate file or a separate PR?

Xia-Weiwen: And this PR needs this change for fake quant because this PR moves the checks from inside `_normalize_granularity` out. And the same for similar changes elsewhere.

torchao/quantization/quant_api.py
Xia-Weiwen: Thanks. Updated.

test/quantization/quantize_/workflows/float8/test_float8_opaque_tensor.py
```diff
@@ -0,0 +1,259 @@
...
+            * 0.1,
+        )
+
+    def forward(self, x):
+        x = self.linear1(x)
+        x = self.linear2(x)
+        return x
+
+
+class TestDynamicFloat8Linear(TestCase):
```
jerryzh168, line 54: nit: `TestFloat8OpaqueTensor`?

test/quantization/quantize_/workflows/float8/test_float8_opaque_tensor.py
```diff
@@ -0,0 +1,259 @@
...
+            (PerTensor(), PerTensor()),
+            (PerTensor(), PerRow()),
+            (PerTensor(), PerGroup(64)),
+        ],
+    )
+    @common_utils.parametrize("dtype", [torch.float, torch.bfloat16, torch.half])
+    @common_utils.parametrize("x_dim", [2, 3])
+    @common_utils.parametrize("bias", [True, False])
+    @common_utils.parametrize("bs", [1, 128])
+    def test_dynamic_float8_linear_per_tensor_cpu(
```
jerryzh168, line 104: this is per tensor activation? might be good to clarify

test/quantization/quantize_/workflows/float8/test_float8_opaque_tensor.py
```diff
@@ -0,0 +1,259 @@
...
+        m = ToyLinearModel(256, 256, bias=bias).eval().to(dtype).to(device)
+        example_inputs = m.example_inputs(batch_size=bs, dtype=dtype, device=device)
+        if x_dim == 3:
+            example_inputs = (example_inputs[0].unsqueeze(0),)
+        y = m(*example_inputs)
+
+        with torch.no_grad():
+            quantize_(
+                m,
+                get_config([PerRow(), PerGroup(group_size)]),
```
jerryzh168, line 185: why are these tests not combined into the same one? seems all of them are very similar

torchao/float8/inference.py
```diff
@@ -205,22 +206,19 @@ def _normalize_granularity(
         ]
     ],
 ) -> Tuple[FP8Granularity, FP8Granularity]:
+    supported_granularities = (PerTensor, PerRow, PerGroup)
```
jerryzh168, line 209: is this only supported for CPU? I think we also rely on this in the cuda path, and it will be surprising if it says supported here but error out somewhere else

torchao/quantization/qat/fake_quantize_config.py
jerryzh168: sorry, separate PR

jerryzh168 (changes requested):
looks mostly fine, I'd suggest to split the PR to
* `get_block_size` changes
* add a normalize_granularity for cpu in quantization/quantize_/workflows/float8/? since cuda doesn't support PerBlock yet
* PR to add float8 linear op? - also probably need some tests for the op itself
* then the PR for Float8Tensor

jerryzh168 (changes requested):
looks mostly fine, I'd suggest to split the PR to
* `get_block_size` changes
* add a normalize_granularity for cpu in quantization/quantize_/workflows/float8/? since cuda doesn't support PerBlock yet
* PR to add float8 linear op? - also probably need some tests for the op itself
* then the PR for Float8Tensor

Xia-Weiwen:
> looks mostly fine, I'd suggest to split the PR to
> 
> * `get_block_size` changes
> * add a normalize_granularity for cpu in quantization/quantize_/workflows/float8/? since cuda doesn't support PerBlock yet
> * PR to add float8 linear op? - also probably need some tests for the op itself
> * then the PR for Float8Tensor

Thanks. I will split it into multiple PRs.

Xia-Weiwen:
We split this PR into the following smaller ones:
- https://github.com/pytorch/ao/pull/3039 (relanded by https://github.com/pytorch/ao/pull/3059)
- https://github.com/pytorch/ao/pull/3052
- https://github.com/pytorch/ao/pull/3075

Closing this one.

PR #2506, opened 2025-07-08
enforce that `MXTensor` scale dimensions are consistent with data

vkuzo:
Summary:

Ensures that if the data dims are (M, K) then scale dims are (M, K //
block_size).  Previously the scale dims were (M, K // block_size, 1).

No logic change in surrounding code, but this is definitely more
correct.

Test Plan:

```
pytest test/prototype/mx_formats
./test/prototype/mx_formats/test_mx_dtensor.sh
```

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo:
* __->__ #2506

PR #2509, opened 2025-07-08
[MoE training] torch.compile support for ScaledGroupedMMTensor

danielvegamyhre:
## Summary
- Return 2nd arg (metadata) from flatten to make tensor subclass traceable for compile
- Wrap triton kernels in custom ops for compile 
- Add e2e training tests with compile

## Test plan


danielvegamyhre:
cc @vkuzo @drisspg for review

Issue #2487, opened 2025-07-03
MPS experimental test is broken in torchao


metascroy:
The MPS experimental test appears to be broken, e.g., https://github.com/pytorch/ao/actions/runs/16056086755/job/45310718172?pr=2486

@manuelcandales can you have a look?

PR #2531, opened 2025-07-11
Update torchao_experimental_test.yml

metascroy:
Disable MPS tests until https://github.com/pytorch/ao/issues/2487 is addressed.  cc @manuelcandales


drisspg (approved):
Please wait till all ci is green before land

PR #2539, opened 2025-07-14
Avoid configuring the root logger on import

carmocca:
`logging.debug` (and `.info`, ...) will `basicConfig()` automatically if there are no handlers
https://github.com/python/cpython/blob/3d8c38f6db0fea7845aafb92fe6bc795b536a367/Lib/logging/__init__.py#L2222-L2223

So the `debug` message in torchao's `__init__.py` adds an extra `StreamHandler` that can result in duplicate logs for downstream applications that import `torchao` and configure their logger at runtime.

torchao should instead use its own logger (as implemented).

### DIFF
--- a/torchao/__init__.py
+++ b/torchao/__init__.py
@@ -20,6 +20,8 @@ try:
 except PackageNotFoundError:
     __version__ = "unknown"  # In case this logic breaks don't break the build
 
+logger = logging.getLogger(__name__)
+
 try:
     from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


carmocca:
cc @andrewor14 (random ping hoping to get a review, sorry and thanks!)

andrewor14:
Thank you @carmocca, we should probably do this for the rest of the repo too. Merging this.

PR #2541, opened 2025-07-14
[BE] [float8] Run test_everything.sh in float8 test CI using linux.aws.h100.4

### LINKED ISSUE #2477: Support running multi-device tests in CI
For float8 training, the test_everything.sh script requires multiple GPUs for FSDP/TP tests, so we currently don't run in CI as it's not configured for multi-device jobs. We should figure out how to run these multi-device tests in CI. This would also be useful for some of our new MoE training parallelism tests.

danielvegamyhre:
Fixes #2477


danielvegamyhre:
fyi @vkuzo

.github/workflows/float8_test.yml
```diff
@@ -53,6 +53,11 @@ jobs:
         uv pip install -r dev-requirements.txt
         uv pip install vllm
         pip install .
-        pytest test/float8 --verbose -s
-        pytest test/integration --verbose -s
-        pytest test/dtypes/test_affine_quantized_float.py --verbose -s
```
drisspg, line 58: This removes tests

drisspg (changes requested):
this is the main way we are testing our integrations, add back

Also seems that there still exsits tests that can run non not 4 gpus that are worth doing

danielvegamyhre:
>Also seems that there still exsits tests that can run non not 4 gpus that are worth doing

Can you clarify what you mean here? We could use `linux.aws.h100.8` for tests that require GPUs (e.g., 3d parallel). I have an in progress MoE test that requires 8 GPUs (#2481) but for the float8 distributed tests they require 4 max i believe

.github/workflows/float8_test.yml
danielvegamyhre: thanks, added those back.

drisspg (approved):
My comment was just that we now are requesting 4 gpus for all of these tests where previously we only needed 1, I am not sure if this ends up being harder to schedule / add smore delays

danielvegamyhre:
> My comment was just that we now are requesting 4 gpus for all of these tests where previously we only needed 1, I am not sure if this ends up being harder to schedule / add smore delays

Hmm I see, perhaps we can see how it goes over the next week or so and if it's slowing us down then we can revisit

vkuzo (changes requested):
how about:
1.  split `test_everything.sh` into one piece for single GPU and another one for multi GPU
2. keep the single GPU one in the current target
3. make a new target for multi GPU

vkuzo:
@danielvegamyhre IMO worth reverting or forward fixing asap, because this PR seems to run the single GPU float8 tests twice, once from the original code and once again from `test_everything.sh`

danielvegamyhre:
Fix forward: https://github.com/pytorch/ao/pull/2561

PR #2542, opened 2025-07-14
Add QLoRA and FP8 to finetuning tutorial (part 2)

andrewor14:
This is part 2 of the end-to-end tutorial. Previously we already had QAT. This commit also adds QLoRA and FP8. To preview, visit https://docs-preview.pytorch.org/pytorch/ao/2542/finetuning.html

### DIFF
--- a/docs/source/finetuning.rst
+++ b/docs/source/finetuning.rst
@@ -284,10 +284,123 @@ schemes, but these are not customizable unlike the above example.
 Quantized Low-Rank Adaptation (QLoRA)
 #####################################
 
-(Coming soon!)
+Low-Rank Adaptation (LoRA) refers to freezing the original model,
+and instead training a set of new "adapter" parameters that are a
+small fraction of the original parameters, thereby significantly
+reducing the memory footprint during training. QLoRA is an extension
+of LoRA that additionally quantizes the frozen original model
+parameters to 4-bits, thereby further reducing the memory footprint.
+
+TorchAO offers an implementation of the NF4 data type proposed in
+the original `QLoRA paper <https://arxiv.org/pdf/2305.14314>`__.
+This implementation expresses NF4 as a tensor subclass through the
+`NF4Tensor <https://docs.pytorch.org/ao/stable/generated/torchao.dtypes.NF4Tensor.html>`__,
+which composes cleanly with other PyTorch features like `torch.compile`
+and FSDP2. Users can convert a high precision tensor to NF4 simply
+by calling `torchao.dtypes.to_nf4 <https://docs.pytorch.org/ao/stable/generated/torchao.dtypes.to_nf4.html>`__.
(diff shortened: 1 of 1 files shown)


docs/source/finetuning.rst
```diff
@@ -284,10 +284,94 @@ schemes, but these are not customizable unlike the above example.
...
+          device: Optional[torch.device] = None,
+          dtype: Optional[torch.dtype] = None,
+          **quantization_kwargs,
+      ):
+          super().__init__(in_dim, out_dim, bias=bias, device=device, dtype=dtype)
+          # No need to train these in QLoRA
+          self.weight.requires_grad_(False)
+          if self.bias is not None:
+              self.bias.requires_grad_(False)
+          nf4_weight = to_nf4(self.weight, **quantization_kwargs)
```
jerryzh168, line 320: can we extend this to support any quantization? I just added a AOBaseTensorConfig that might be able to help: https://github.com/pytorch/ao/pull/2463

docs/source/finetuning.rst
```diff
@@ -284,10 +284,94 @@ schemes, but these are not customizable unlike the above example.
...
+for training. For a full example of how TorchTune incorporates the `NF4Tensor`
+in its QLoRA fine-tuning recipe, see their implementation of
+`LoRALinear <https://github.com/pytorch/torchtune/blob/a6290a5b40758f13bca61c386bc8756a49ef417e/torchtune/modules/peft/lora.py#L19>`__.
+You can also try it out by running the following command,
+or refer to their `QLoRA tutorial <https://docs.pytorch.org/torchtune/stable/tutorials/qlora_finetune.html>`__
+for more details.
+
+.. code::
+
+  tune run lora_finetune_single_device --config llama3_2/3B_qlora_single_device.yaml
```
jerryzh168, line 342: what about integration with the transformer peft library?

also isn't tune deprecated

jerryzh168 (approved):
looks great, thanks!

docs/source/finetuning.rst
andrewor14: Ok, I'll add a note to say it's possible to extend to other quantization, but want to keep this example NF4 since that's what's used in the QLoRA paper

docs/source/finetuning.rst
andrewor14: wow I didn't realize we had an integration with peft, don't think this was documented in any of our docs? Will add a note here

For torchtune I don't think there's a mature replacement yet from pytorch so I feel it's OK. Also that's where most of our fine-tuning integrations live today

jerryzh168: yeah: https://huggingface.co/docs/peft/en/developer_guides/quantization#torchao-pytorch-architecture-optimization, I'll test this path a bit soon as well

PR #2548, opened 2025-07-15
Increase timeout for regression tests

wdvr:


danielvegamyhre:
@wdvr hmm operation was cancelled after 2.5 hrs (ROCM) and 1hr (H100)

PR #2550, opened 2025-07-15
add custom op wrapping mxfp8 dim1 cast cuda kernel

danielvegamyhre:
Stacked PRs:
 * #2554
 * #2553
 * #2552
 * #2551
 * __->__#2550

--- --- ---

### add custom op wrapping mxfp8 dim1 cast cuda kernel

### DIFF
--- a/torchao/prototype/mx_formats/kernels.py
+++ b/torchao/prototype/mx_formats/kernels.py
@@ -16,7 +16,11 @@ from torchao.prototype.custom_fp_utils import (
     _f32_to_floatx_unpacked,
     _floatx_unpacked_to_f32,
 )
-from torchao.utils import TORCH_VERSION_AT_LEAST_2_4, TORCH_VERSION_AT_LEAST_2_7
+from torchao.utils import (
+    TORCH_VERSION_AT_LEAST_2_4,
+    TORCH_VERSION_AT_LEAST_2_7,
+    is_sm_at_least_100,
+)
 
 # TODO(future): if needed, make the below work on previous PyTorch versions,
 # just need to hunt down the previous location of `libdevice`. An assert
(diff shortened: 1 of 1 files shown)


torchao/prototype/mx_formats/kernels.py
```diff
@@ -1730,3 +1734,82 @@ def triton_quantize_nvfp4(
...
+            )
+        else:
+            output_colwise = x.new_empty(0, dtype=torch.float8_e4m3fn)
+            scales_colwise = x.new_empty(0, dtype=torch.float8_e8m0fnu)
+
+        return output_rowwise, output_colwise, scales_rowwise, scales_colwise
+
+else:
+
+    def mxfp8_quantize_cuda(
```
drisspg, line 1817: Nit I hate this pattern it would be cool to create a decorator that
@requires_triton

or something that will raise this error if not avail

danielvegamyhre: agree this is not ideal. i like the decorator idea, will try it

danielvegamyhre: The thing is I also don't want to do the import if not sm100 because the kernel won't exist... I can write a decorator for this but it won't be very generic

danielvegamyhre: Hmm it seems like with a decorator approach there are problems with torch custom op internals trying to infer the schema. 
```
../.conda/envs/ao/lib/python3.13/site-packages/torch/_library/infer_schema.py:69: in error_fn
    raise ValueError(f"infer_schema(func): {what} Got func with signature {sig})")
E   ValueError: infer_schema(func): Parameter func must have a type annotation. Got func with signature (func))
````

I solved this by reversing the order of the custom op decorator and new decorator so the signature is preserved:

```python
@requires_sm100
@torch.library.custom_op("torchao::mxfp8_quantize_cuda", mutates_args=())
def mxfp8_quantize_cuda(
...
```

However, I then get an error when trying to register a fake for the custom op:

```
torchao/prototype/mx_formats/kernels.py:1770: in <module>
    @mxfp8_quantize_cuda.register_fake
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E   AttributeError: 'function' object has no attribute 'register_fake'
```

I inspected the object returned by the decorator and managed to figure out how to get it to work, but it is ugly and think it will probably have side effects.... :

```python
# ... in the decorator ... 
   return decorator.__closure__[0].cell_contents
```

Here's the full working decorator: https://www.internalfb.com/phabricator/paste/view/P1870964511

IMO I am not confident enough in this implementation to include it atm (I think it probably won't work correctly on non-sm100) and prefer the predictability of the statement (for now).

drisspg: cc @zou3519 Do you think that multiple decorators should be supported / supportable?

vkuzo: would be good to keep the code inside this file consistent, I'd vote for a separate PR fixing all the kernels with a better fallback syntax

torchao/prototype/mx_formats/kernels.py
```diff
@@ -1730,3 +1734,82 @@ def triton_quantize_nvfp4(
...
+# MXFP8 CUDA kernel is only built on SM100+
+if is_sm_at_least_100():
+    from torchao.prototype import mxfp8_cuda
+
+    @torch.library.custom_op("torchao::mxfp8_quantize_cuda", mutates_args=())
+    def mxfp8_quantize_cuda(
+        x: torch.Tensor,
+        rowwise: bool = False,
+        colwise: bool = True,
+        scaling_mode: str = "floor",
```
vkuzo, line 1748: would be good to use the scaling mode enum in as many places as possible, and convert to string right before the call into the c++ wrapper

vkuzo (approved):

torchao/prototype/mx_formats/kernels.py
zou3519: What's the goal? To error out when the operator gets executed and we are < sm100?

What's wrong with:
```
@torch.library.custom_op("torchao::mxfp8_quantize_cuda", mutates_args=())
def mxfp8_quantize_cuda(...):
    if not_is_sm100(...):
        raise RuntimeError(...)
```

The multiple decorators can work though. My rec is:
```
@torch.library.custom_op("torchao::mxfp8_quantize_cuda", mutates_args=())
@requires_sm100
def mxfp8_quantize_cuda(
```
but because requires_sm100 wraps a function, it should also propagate the type annotations to the wrapped function. That'll let torch.library.custom_op read from the type annotations.

torchao/prototype/mx_formats/kernels.py
danielvegamyhre: Makes sense, done

danielvegamyhre: Nevermind, I remembered why I had to use a string in the first place, it seems custom ops don't place nicely with enums:

```
torchao/prototype/mx_formats/kernels.py:1746: in <module>
    @torch.library.custom_op("torchao::mxfp8_quantize_cuda", mutates_args=())
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../.conda/envs/ao/lib/python3.13/site-packages/torch/_library/custom_ops.py:148: in inner
    schema_str = torch.library.infer_schema(fn, mutates_args=mutates_args)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../.conda/envs/ao/lib/python3.13/site-packages/torch/_library/infer_schema.py:127: in infer_schema
    if annotation_type.__origin__ is tuple:
       ^^^^^^^^^^^^^^^^^^^^^^^^^^
E   AttributeError: type object 'ScaleCalculationMode' has no attribute '__origin__'
```

zou3519: yeah, you can't pass an enum to a custom op.

PR #2556, opened 2025-07-15
[float8] bump float8 ci test on h100 to cuda 12.6

danielvegamyhre:
Test uses torch nighly w/ cuda 12.6 but `torch-spec` specifies 12.4 still, need to bump it.


danielvegamyhre:
cc @jerryzh168

Issue #2483, opened 2025-07-03
torchao installation should be clearly called out in the README.md


vkuzo:
This used to be here but looks like it got removed.  Here is what I think we should have:
* a dedicated "Installation" header.  Right now installation is under "Quick Start" which kind of makes sense but not 100% clear.
* bring back documentation about `USE_CPP=0` flag

PR #2559, opened 2025-07-16
Update README with PEFT integration + installation

andrewor14:
Highlight our missing PEFT LoRA integration and add a section for installation (fixes https://github.com/pytorch/ao/issues/2483).

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@
 [![](https://img.shields.io/badge/torchao-documentation-blue?color=DE3412)](https://docs.pytorch.org/ao/stable/index.html)
 [![license](https://img.shields.io/badge/license-BSD_3--Clause-lightgrey.svg)](./LICENSE)
 
-[Latest News](#-latest-news) | [Overview](#-overview) | [Quick Start](#-quick-start)  | [Integrations](#-integrations) | [Inference](#-inference) | [Training](#-training) | [Videos](#-videos) | [Citation](#-citation)
+[Latest News](#-latest-news) | [Overview](#-overview) | [Quick Start](#-quick-start)  | [Installation](#-installation) | [Integrations](#-integrations) | [Inference](#-inference) | [Training](#-training) | [Videos](#-videos) | [Citation](#-citation)
 
 </div>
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -106,14 +89,40 @@ speedup: 6.9x
...
+</details>
+
+
 ## 🔗 Integrations
 
 TorchAO is integrated into some of the leading open-source libraries including:
 
 * HuggingFace transformers with a [builtin inference backend](https://huggingface.co/docs/transformers/main/quantization/torchao) and [low bit optimizers](https://github.com/huggingface/transformers/pull/31865)
 * HuggingFace diffusers best practices with `torch.compile` and TorchAO in a standalone repo [diffusers-torchao](https://github.com/huggingface/diffusers/blob/main/docs/source/en/quantization/torchao.md)
+* HuggingFace PEFT for QLoRA using TorchAO as their [quantization backend](https://huggingface.co/docs/peft/en/developer_guides/quantization#torchao-pytorch-architecture-optimization)
```
salmanmohammadi: @andrewor14 AFAIK HF has yet to support QLoRA with ao. However, I would love to see this in HF, as we experience difficulties in composing the bnb implementation with e.g. FSDP2/TP.

andrewor14: I think this is loading a quantized model (only int8 supported now) and then running LoRA on it?

```
quantization_config = TorchAoConfig(quant_type="int8_weight_only")
base_model = AutoModelForCausalLM.from_pretrained(model_id, quantization_config=quantization_config)
peft_config = LoraConfig(...)
model = get_peft_model(base_model, peft_config)
```

That's what I mean by QLoRA here, though it's not "QLoRA based on NF4" in the traditional sense like in the paper

andrewor14: > we experience difficulties in composing the bnb implementation with e.g. FSDP2/TP

Yeah we had some discussions on this with the bnb team too. I think the way to enable this would be to implement a new version that uses our NF4 tensor subclass (or alternate 4-bit dtype tensor subclass like MXFP4 and NVFP4). That should work out of the box with FSDP2

andrewor14: (Updated QLoRA to LoRA here to avoid confusion)

PR #2561, opened 2025-07-16
[BE] [CI] Set up 1xL4, 1xH100, 4xH100 CI workflows 

danielvegamyhre:
## Summary
Multi-GPU tests are not currently run in CI. This PR makes the following changes to support this:

- Instead of splitting CI tests by feature (namely, float8), we split by hardware:
    - 1xH100 tests (single gpu integration, float8 tests)
    - 4xH100 tests (multi-gpu float8 distributed tests)
    - 1xL4 tests (singlegp8 integration, float8 tests)

As a next step I'll add our MoE training tests to the 4xH100 workflow

### DIFF
--- a/.github/workflows/float8_multi_gpu_test.yml
+++ b/.github/workflows/float8_multi_gpu_test.yml
@@ -1,4 +1,4 @@
-name: Run Float8 Tests
+name: Run Float8 Multi-GPU Tests
 
 on:
   push:
@@ -11,7 +11,7 @@ on:
       - 'gh/**'
 
 concurrency:
-  group: float8_test-${{ github.workflow }}-${{ github.ref == 'refs/heads/main' && github.run_number || github.ref }}
+  group: float8_multi_gpu_test-${{ github.workflow }}-${{ github.ref == 'refs/heads/main' && github.run_number || github.ref }}
   cancel-in-progress: true
 
 env:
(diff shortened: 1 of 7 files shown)


.github/workflows/float8_multi_gpu_test.yml
```diff
@@ -0,0 +1,51 @@
...
+env:
+  HF_TOKEN: ${{ secrets.HF_TOKEN }}
+
+jobs:
+  test:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: H100
```
danielvegamyhre: Note for reviewers: multi-gpu tests will only run on 4xH100, so no coverage of SM89 in this set up. I'm not aware of any multi-gpu hosts for sm89 atm, but can look into it

danielvegamyhre:
cc @vkuzo @drisspg for review

test/float8/test_everything_single_gpu.sh
```diff
@@ -0,0 +1,18 @@
...
+#!/bin/bash
+
+# terminate script on first error
+set -e
+
+pytest test/float8/test_base.py --verbose -s
+pytest test/float8/test_compile.py --verbose -s
+pytest test/float8/test_numerics_integration.py --verbose -s
+pytest test/float8/test_auto_filter.py --verbose -s
+pytest test/integration --verbose -s
```
drisspg: nit: These are not specific to float8, but I wanted to amke sure they were run on h100 node. It feels a little strange to have in thsi file, can we just keep all the tests closer to the workflow so when people wnat to expand h100 tests they can update.

Also the job name should be changed in retrospect to h100 tests or whatever

danielvegamyhre: Hmm, I see. Yeah I guess we can split workflows based on hardware type, that would also make scheduling/allocation faster if all tests that need 1xH100 share that job, all tests that need 4xH100 share that job, etc.

vkuzo: splitting tests by product and CI jobs by hardware type makes sense to me

danielvegamyhre: Split the CI workflows into:
- 1xH100
- 4xH100
- 1xL4 (SM89)

(We could also name them SM89/SM90)

This is ready for another look.

drisspg (approved):

PR #2554, opened 2025-07-15
integrate mxfp8 dim1 cast kernel choice enum into MXLinear

danielvegamyhre:
Stacked PRs:
 * __->__#2554
 * #2553
 * #2552
 * #2551

--- --- ---

## Summary
- Add `MXFP8Dim1CastKernelChoice` enum and replace all uses of boolean flag `use_fp8_dim1_cast_triton_kernel` with it. (Default to Triton for now)
- Update tests accordingly and verify they are passing.

## Test plan
- `pytest test/prototype/mx_formats/test_mx_linear.py -k eager_vs_hp`
- `pytest test/prototype/mx_formats/test_mx_linear.py -k compile` 

## Next steps
- Integrate into torchtitan for e2e fsdp training tests once this stack lands. Torchtitan PR: https://github.com/pytorch/torchtitan/pull/1401
- Dtensor tests still having issues both with Triton and CUDA: `./test/prototype/mx_formats/test_mx_dtensor.sh`
   - Triton error (known issue): `RuntimeError: Attempting to use FunctionalTensor on its own. Instead, please use it with a corresponding FunctionalTensorMode()`
   - Cuda error in dtensor op dispatch: `assert res.ndim == 0, "output tensor should be scalar!"`

PR #2564, opened 2025-07-16
integration of new mxfp8 casting cuda kernel

danielvegamyhre:
integration of new mxfp8 casting cuda kernel


danielvegamyhre:
@vkuzo @drisspg I squashed the remaining changes in the original stack into the same PR, so that the tests would be in the same PR as the changes.

drisspg:
Didn't read any code yet:
```
Update triton_scale_swizzle kernel to accept both row major and col major inputs (since cuda kernel writes scale in col major, to avoid uncoalesced global accesses)
```
I think updating to just accept generically strided inputs and writing out row major (required for mm kernels) is good, is that what you did?

danielvegamyhre:
> Didn't read any code yet:
> 
> ```
> Update triton_scale_swizzle kernel to accept both row major and col major inputs (since cuda kernel writes scale in col major, to avoid uncoalesced global accesses)
> ```
> 
> I think updating to just accept generically strided inputs and writing out row major (required for mm kernels) is good, is that what you did?

Yep, that's correct

test/prototype/mx_formats/test_mx_dtensor.py
```diff
@@ -25,6 +25,7 @@
 from tqdm import tqdm
 
 from torchao.prototype.mx_formats import MXLinearConfig
+from torchao.prototype.mx_formats.config import MXFP8CastKernelChoice
```
vkuzo: sorry, `MXFP8Dim1CastKernelChoice`? since for dim0 we are always using torch.compile

torchao/prototype/mx_formats/config.py
```diff
@@ -33,6 +33,12 @@ class MXGemmKernelChoice(Enum):
     CUBLAS = "cublas"
 
 
+class MXFP8CastKernelChoice(Enum):
```
vkuzo: add some comments on what the options are?

torchao/prototype/mx_formats/mx_linear.py
```diff
@@ -15,21 +15,41 @@
...
 ):
-    a_data, a_scale = triton_to_mxfp8_dim1(a, block_size)
+    if cast_kernel_choice == MXFP8CastKernelChoice.TRITON:
+        a_data, a_scale = triton_to_mxfp8_dim1(a, block_size)
+    elif cast_kernel_choice == MXFP8CastKernelChoice.CUDA:
+        _, a_data, _, a_scale = mxfp8_quantize_cuda(
+            a,
+            rowwise=False,
+            colwise=True,
+            scaling_mode="floor",
```
vkuzo, line 48: TODO for later to allow choice of scaling modes

vkuzo (approved):

test/prototype/mx_formats/test_mx_dtensor.py
danielvegamyhre: It was originally `MXFP8Dim1CastKernelChoice` but [here](https://github.com/pytorch/ao/pull/2554#discussion_r2210140775) we discussed naming it `MXFP8CastKernelChoice` for potentially adding support for dim0 and dim0+dim1 casts as well. I don't have strong opinions either way, I went ahead and changed it back to `MXFP8Dim1CastKernelChoice`

torchao/prototype/mx_formats/mx_linear.py
danielvegamyhre: Added todo on the custom op itself, with an explanation why we currently are using a string param

PR #2568, opened 2025-07-17
[BE] Convert quantization internal methods private

jainapurva:
As a part of clean code, we're converting internal implementations private and removing old deprecated APIs.

- **torchao/quantization/autoquant.py**
  - `check_cache` -> `_check_cache`
  - `update_cache` -> `_update_cache`
  - `AUTOQUANT_CACHE` -> `_AUTOQUANT_CACHE`

- **torchao/quantization/utils.py**
  - `guard_dtype_size` -> `_guard_dtype_size`
  - `quant_int8_per_token_matmul` -> `_quant_int8_per_token_matmul`
  - `quantize_activation_per_token_absmax` -> `_quantize_activation_per_token_absmax`
  - `quant_int8_dynamic_per_token_linear` -> `_quant_int8_dynamic_per_token_linear`

### DIFF
--- a/torchao/quantization/autoquant.py
+++ b/torchao/quantization/autoquant.py
@@ -27,8 +27,8 @@ from torchao.quantization.quant_primitives import (
     ZeroPointDomain,
 )
 from torchao.quantization.utils import (
+    _quantize_activation_per_token_absmax,
     compute_error,
-    quantize_activation_per_token_absmax,
 )
 from torchao.utils import (
     TORCH_VERSION_AT_LEAST_2_3,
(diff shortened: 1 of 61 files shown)


vkuzo (approved):

torchao/quantization/autoquant.py
```diff
@@ -66,11 +66,11 @@
 AUTOQUANT_CACHE = {}
```
andrewor14: can you also make this one private?

torchao/quantization/utils.py
```diff
@@ -33,10 +32,8 @@
 
 __all__ = [
     "compute_error",
-    "_apply_logging_hook",
     "quantize_activation_per_token_absmax",
     "quant_int8_dynamic_per_token_linear",
```
andrewor14: which other functions in this list can also be privatized? Looks like these 2 at least aren't used outside of torchao (and I don't think we intend them to be). Seems like there are a few more

andrewor14 (approved):
Thanks, seems like there may be more functions in these two files that can be privatized

PR #2576, opened 2025-07-21
Add StretchedUnifTorchaoQuantizer

lisjin:
This PR adds a new stretched uniform quantizer for PARQ, which empirically performs well for 2- and 3-bit QAT. Main differences:
- Fractional `quant_min=-2**(b - 1) + 0.5` and `quant_max=2**(b - 1) - 0.5` values
- Per-block `min_val`, `max_val` are computed by taking a multiple of the mean over absolute values (instead of absmax)

As in #2091, I also compare the resulting PARQ quantized weights with those quantized with torchao's module swap + `quantize_` API. To support this, I created a new tensor subclass `StretchedAffineQuantizedTensor` and config `StretchedIntxWeightOnlyConfig` to handle floating point `quant_min`, `quant_max`, and `zero_point` values.

### DIFF
--- a/test/prototype/test_parq.py
+++ b/test/prototype/test_parq.py
@@ -21,10 +21,12 @@ from torchao.prototype.parq.optim import (
 from torchao.prototype.parq.quant import (
     Int4UnifTorchaoQuantizer,
     LSBQuantizer,
+    StretchedUnifTorchaoQuantizer,
     TernaryUnifQuantizer,
     UnifQuantizer,
     UnifTorchaoQuantizer,
 )
+from torchao.prototype.parq.quant.quant_api import StretchedIntxWeightOnlyConfig
 from torchao.prototype.parq.quant.uniform_torchao import _BIT_WIDTH_TO_DTYPE
 from torchao.quantization.granularity import PerGroup
 from torchao.quantization.qat import (
@@ -35,11 +37,11 @@ from torchao.quantization.qat import (
 from torchao.quantization.quant_api import (
     Int8DynamicActivationIntxWeightConfig,
     IntxWeightOnlyConfig,
-    MappingType,
     _is_linear,
     int4_weight_only,
     quantize_,
 )
+from torchao.quantization.quant_primitives import MappingType
 from torchao.utils import (
     TORCH_VERSION_AT_LEAST_2_4,
     TORCH_VERSION_AT_LEAST_2_6,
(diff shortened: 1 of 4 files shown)


torchao/prototype/parq/quant/quant_api.py
```diff
@@ -0,0 +1,175 @@
...
+        target_dtype: torch.dtype,
+        b: int,
+        quant_min: Optional[float] = None,
+        quant_max: Optional[float] = None,
+        eps: Optional[float] = None,
+        scale_dtype: Optional[torch.dtype] = None,
+        zero_point_dtype: Optional[torch.dtype] = None,
+        preserve_zero: bool = False,
+        zero_point_domain: ZeroPointDomain = ZeroPointDomain.FLOAT,
+        _layout: Layout = PlainLayout(),  # noqa: B008
```
metascroy: Should this be QDQLayout or PlainLayout?

torchao/prototype/parq/quant/quant_api.py
```diff
@@ -0,0 +1,175 @@
...
+        scale_method: str = "mean",
+    ):
+        original_shape = input_float.shape
+        input_float = _layout.pre_process(input_float)
+
+        dim = None
```
metascroy: Is it possible to use this function here: https://github.com/pytorch/ao/blob/main/torchao/quantization/quant_primitives.py#L271

torchao/prototype/parq/quant/quant_api.py
```diff
@@ -0,0 +1,175 @@
...
+
+to_stretched_affine_quantized_intx = StretchedAffineQuantizedTensor.from_hp_to_intx
+
+
+@dataclass
+class StretchedIntxWeightOnlyConfig(IntxWeightOnlyConfig):
+    b: Optional[int] = None
+    quant_min: Optional[float] = None
+    quant_max: Optional[float] = None
+    mapping_type: MappingType = MappingType.ASYMMETRIC
```
metascroy: It doesn't look like symmetric supports float point domain: https://github.com/pytorch/ao/blob/main/torchao/quantization/quant_primitives.py#L1468-L1470

Does it make sense to have ZeroPointDomain.INT or MappingType.SYMMETRIC for StretchedIntxWeightOnlyConfig?

If not, maybe we remove those from the config?

torchao/prototype/parq/quant/quant_api.py
```diff
@@ -0,0 +1,175 @@
...
+        data = quantize_affine(
+            input_float,
+            block_size,
+            scale,
+            zero_point,
+            target_dtype,
+            quant_min=quant_min,
+            quant_max=quant_max,
+        )
+        data, scale, zero_point = _layout.post_process(
```
metascroy, line 140: IIUC, the zero_point isn't some fixed fp value, but can vary slightly based on the ranges.

So what the CPU kernel can support is a fp32 scale times a int8-LUT value.

So if we want the grid {-3.5, -1.5, 1.5, 3.5}, we instead use the LUT = {-7, -3, 3, 7} with s_table = 0.5.

In addition to having this LUT for the whole tensor, we can have an FP32 scale s at a per-group granularity.  Dequantization is then:

w_dequantized = s * s_table * LUT[idx]

It's not clear to be that the affine scheme here is representable in that way.  IIUC, you have:

w_dequantized = s * (qval - z)

 So qval - z could define the LUT, but it looks like we'd have a different LUT per group_size values because z changes every group_size values?  Is that right?

lisjin: Thanks for the comments @metascroy!

Just to clarify from our chat earlier, `zero_point=-0.5` is the same across all groups. (I flipped the sign since it's standard to add zero_point during quantization.)

```python
w_quantized = torch.round(x / s + zero_point)
w_dequantized = s * (w_quantized - zero_point)
```
For the 2-bit case, we set `s` so that `x / s` is restricted to range [-1.5, 1.5]. Since `zero_point=-0.5`, `w_quantized` lies in the grid {-2, -1, 0, 1}.

It seems like we don't need to use LUT format since this is well-supported by an affine scheme. Maybe it would be worth supporting for latency comparisons though (and to avoid float `zero_point`).

test/prototype/test_parq.py
```diff
@@ -273,7 +275,84 @@ def test_intx_weight_only_e2e(self, b: int = 2, group_size: int = 32):
             ProxHardQuant(),
             quant_per_channel=True,
         )
-        self.compare_parq_convert(model, m_ref, optimizer, config)
+        compare_parq_convert(model, m_ref, optimizer, config)
+
+
+class TestStretchedUnifTorchaoQuantizer(common_utils.TestCase):
```
lisjin, line 281: New test case that ensures equivalence between PARQ's original `UnifQuantizer` implementation and the new `StretchedUnifTorchaoQuantizer`

torchao/prototype/parq/quant/quant_api.py
```diff
@@ -0,0 +1,221 @@
...
+
+    q_abs = input_float.abs()
+    max_val = torch.minimum(
+        b * q_abs.mean(dim=reduction_dims, keepdim=True),
+        torch.amax(q_abs, dim=reduction_dims, keepdim=True),
+    ).clamp_(min=eps)
+
+    scale = max_val / quant_max
+    scale = scale.to(dtype=scale_dtype, device=input_float.device)
+    zero_point = torch.full_like(scale, -0.5, dtype=zero_point_dtype)
```
lisjin, line 59: Here's the logic for initializing the scale based on multiples of per-group absolute value means. I also manually set the zero point to be the same across groups.

torchao/prototype/parq/quant/quant_api.py
metascroy: It is well supported by an affine scheme where zero_point is a float, but we do not have CPU kernel support for this.

But if zero_point is always 0.5, then w_quantized - zero_point is just some value in [1.5, 1.5], and this could define an LUT, so I think we can hook into the kernel in that way.

We just need the LUT to be integer, so we can define the LUT as [-3, -1, 1, 3] and then divide the scales in half.

PR #2590, opened 2025-07-24
Only assert when CUDA is available.

Stonepia:
This commit performs the capability checks only when CUDA is available, so that we have better support for third-party devices like Intel GPU.

### DIFF
--- a/torchao/float8/inference.py
+++ b/torchao/float8/inference.py
@@ -167,7 +167,7 @@ def _check_hardware_support(
             raise ValueError(
                 f"Invalid granularity type: {_granularity}, only PerTensor or PerRow are supported."
             )
-
-        assert is_sm_at_least_89() or is_MI300(), (
-            "Float8 dynamic quantization requires CUDA compute capability ≥8.9 or MI300+."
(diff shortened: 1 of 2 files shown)


torchao/float8/inference.py
```diff
@@ -167,7 +167,7 @@ def _check_hardware_support(
             raise ValueError(
                 f"Invalid granularity type: {_granularity}, only PerTensor or PerRow are supported."
             )
-
-        assert is_sm_at_least_89() or is_MI300(), (
-            "Float8 dynamic quantization requires CUDA compute capability ≥8.9 or MI300+."
-        )
```
liangan1, line 173: This change will impact other device, such as cpu or any unknow device. Suggest to add common utlis function to judge the fp8 capability and apply it to all fp8 related changes. This function should only ensure the CUDA compute capability ≥8.9 or MI300+ and XPU device is available now.

liangan1:
Can you show the accuracy result for dq-fp8 for both CUDA and XPU based on the LLama-3.1-8B?

PR #2592, opened 2025-07-24
Add Float8BlockwiseLinear with Triton kernels for quantization and GEMMs

danielvegamyhre:
Stacked PRs:
 * __->__#2592

--- --- ---

## Summary
- Add "Float8BlockwiseLinear" and make it differentiable with autograd func to support training
- Add Triton kernels for (1) activation quant, (2) weight quant, and (3) GEMM based on these [DeepGemm inference](https://github.com/deepseek-ai/DeepSeek-V3/blob/main/inference/kernel.py) kernel  and rename GEMM to explicitly include expected scaling granularity of operands in the function names (`fp8_gemm` => `blockwise_fp8_gemm_1x128_128x128`).
- Add new Triton kernel need for backward: `blockwise_fp8_gemm_1x128_1x128_kernel` for dW calculation where both left and right operands have activation scaling granularity (1 x block_size). This is a modified version of the kernel above, so it accepts 1x128 scaling for both operands.
- GEMM kernels do accumulation in fp32 and cast output to bfloat16.
- Modify all quantization kernels to use EPS to guard against NaNs upon division by 0.
- Add tests verifying numerics by enforcing reasonable SQNR
- Added benchmarking script for comparing Triton kernels vs FBGEMM vs DeepGEMM quantization kernels.

## Why use Triton kernels instead of DeepGEMM cutlass kernels?
The GEMM APIs in @vkuzo's PoC [here](https://github.com/pytorch/ao/pull/2386/files) no longer exist in DeepGemm. I tried using the new GEMM APIs (`fp8_gemm_nt` etc), and:
- On B200, with both Vasiliy's PR and my PR, got device-side asserts on this [line](https://github.com/deepseek-ai/DeepGEMM/blob/187656694f7f69e3e7975617a68bc3387680a7e1/csrc/jit/kernel_runtime.hpp#L49), that were not immediately clear how to resolve.
- On H100, I only tried Vasiliy's PR, but got undefined symbols error from CUDA, despite using CUDA toolkit 12.8+ as stated in the readme.

Since our only goal is a functional skeleton and not performance, rather than spend more time on this, I just used the existing Triton kernels we had and made a modified GEMM (1 line change) to support blockwise_fp8_gemm_1x128_1x128_kernel.

If we want to replace these Triton GEMMs with the Cutlass ones later to see if perf is better (it probably is), we can do that.

## Note on numerics
Interestingly, the reference DeepGemm triton quantization kernels do NOT use EPS/clamping to prevent division by 0. This resulted in my unit tests passing (where inputs were from a normal distributed), but NaNs occuring in TorchTitan training runs, where actual activation values sometimes had amax of 0.

I updated the kernels to use the same EPS guards as torchao.float8, and this fixed the Nans. 

## Test plan
- `pytest test/prototype/blockwise_fp8/test_blockwise_kernels.py`
- `pytest test/prototype/blockwise_fp8/test_blockwise_linear.py`

## Torchtitan PoC integration results
- [Logs](https://www.internalfb.com/phabricator/paste/view/P1882038839) showing converted model, nice decreasing loss, but very low throughput (~5.1k vs ~9.8k for bfloat16).
- I think perf may be bad because of (1) all the eager .t().contiguous() type of transformations done, and (2) the DeepGemm triton GEMM requires the B tensor be in row-major, then does strided reads to get it in col-major in SMEM to do the fp8 wgmma? Same thing as with fp8 flex attention I think..


danielvegamyhre:
cc @vkuzo @drisspg for review

test/prototype/test_fp8_blockwise_kernels.py
```diff
@@ -0,0 +1,160 @@
...
+    A = torch.randn(M, K).cuda()
+    B = torch.randn(N, K).cuda()
+    C = A @ B.T
+    A_q, A_s = fp8_blockwise_act_quant(A, dtype=dtype)
+    B_q, B_s = fp8_blockwise_weight_quant(B, dtype=dtype)
+    C_q = blockwise_fp8_gemm(A_q, A_s, B_q, B_s)
+    error = torch.norm(C - C_q) / torch.norm(C)
+    print(f"Relative Error: {error.item():.6f}")
+
+    assert error < 0.1, "Quantize gemm error is too high"
```
drisspg: Can you use sqnr everywhere match w/ existing numerics testing

torchao/prototype/blockwise_fp8/kernels.py
```diff
@@ -277,3 +277,231 @@ def fp8_blockwise_weight_dequant(
     )
     fp8_blockwise_weight_dequant_kernel[grid](x, s, y, M, N, BLOCK_SIZE=block_size)
     return y
+
+
+# original implementation from fbgemm_gpu:
+# https://github.com/pytorch/FBGEMM/blob/b19401e913fcdff536dc097fa3013a0a9d66256e/fbgemm_gpu/experimental/gemm/triton_gemm/fp8_gemm.py#L3091
+def triton_quantize_fp8_block(
```
drisspg, line 444: since we have an optional runtime dependency on fbgemm can we just call their kernel directly?

drisspg:
```
(32768, 32768)  (1,128)          5732.42      7960.21        7830.56
(32768, 32768)  (128,128)       13692.4        669.664       7831.14
```

this number is kinda weird to me, do you have memory bandwidth calcs? I dont immediately get why there is a 10x delta in group wise vs blockwise

torchao/prototype/blockwise_fp8/kernels.py
danielvegamyhre: Yes that is the desired end state. For now I have tried and have had repeated problems getting it to work so far (`fbgemm-gpu-genai`), e.g. undefined symbols. Tried on both H100 and B200 and got different undefined symbol errors

test/prototype/test_fp8_blockwise_kernels.py
danielvegamyhre: Updated to use SQNR

danielvegamyhre:
>this number is kinda weird to me, do you have memory bandwidth calcs? I dont immediately get why there is a 10x delta in group wise vs blockwise

Yeah I agree it's odd, will try adding some mem bw calcs, was thinking about checking with Josh / fbgemm team as well if perhapst here is a different kernel they use for activation quant.

danielvegamyhre:
@drisspg @vkuzo this is ready for another look, numerics look good and torchtitan loss curve looks good. perf is bad for llama3 8b right now.

(accidentally squashed my stack into a single pr with stack-pr, sorry for the large PR, I can break it up if necessary)

drisspg:
Can you try scaled_mm for this, if you install cuda 12.9 nightly this should work

danielvegamyhre:
> Can you try scaled_mm for this, if you install cuda 12.9 nightly this should work

Sure, I think I'm actually going to close this PR and open a new stack, because for learning I wrote some kernels for blockwise quant ops and GEMMs specific for training and they work better than repurposing these inference kernels. 

I will check out _scaled_mm support for blockwise GEMM as well.


Update: tried it out and it seems like torch._scaled_mm does not yet support groupwise scaling of A + blockwise scaling of B, which we need for dsv3. 

scaled_mm currently requires both A and B use the same scaling scheme for both A and B:
-  1x128 for A and 128x1 for B, or
- 128x128 for A and 128x128 for B

It does not support:
- 1x128 for A and 128x128 for B

I will check with Luca if this is intentional or if I'm missing something.

```python
 51  ->         breakpoint()
 52             out = torch._scaled_mm(
 53                 x_fp8,
 54                 weight_t_fp8,
 55                 scale_a=1.0 / x_scale,
 56                 scale_b=1.0 / weight_t_scale,

# input with (1x128 scales)
(Pdb) x_fp8.shape
torch.Size([256, 4096])
(Pdb) x_scale.shape
torch.Size([256, 32])
# weight.T with (128x128) scales
(Pdb) weight_t_fp8.shape
torch.Size([4096, 128256])
(Pdb) weight_t_scale.shape
torch.Size([32, 1002])

...

RuntimeError: Invalid scaling configuration.
- For TensorWise scaling, a and b should be float8, scales should be float and singletons.
- For RowWise scaling, a and b should be float8, scales should be float, scale_a should be (256, 1) and scale_b should be (1, 128256), and both should be contiguous.
- For BlockWise 1x128 scaling, a and b should be float8, scales should be float, scale_a should be (256, 32) and scale_b should be (32, 128256), and both should be outer-dim-major.
- For BlockWise 128x128 scaling, a and b should be float8, scales should be float, scale_a should be (2, 32) and scale_b should be (32, 1002), and both should be near-inner-dim-major (with 16-byte aligned strides).
- For Blockwise 1x32 scaling, a and b should be float8, scales should be float8_e8m0fnu, scale_a should have 32768 elements and scale_b should have 16416768 elements, and both should be contiguous.
- For Blockwise 1x16 scaling, a and b should be float4 (packed 2x), scales should be float8_e4m3fn, scale_a should have 131072 elements and scale_b should have 65667072 elements, and both should be contiguous.
Got a.dtype()=Float8_e4m3fn, scale_a.dtype()=Float, scale_a.size()=[256, 32], scale_a.stride()=[32, 1], b.dtype()=Float8_e4m3fn, scale_b.dtype()=Float, scale_b.size()=[32, 1002] and scale_b.stride()=[1002, 1]

```

PR #2600, opened 2025-07-25
wip MoE refactor

HDCharles:
Summary:

now that the pytorch grouped_mm kernels don't require padding, refactoring the moe implementation to use that rather than what was there before.

DONE
-implement moe with grouped_mm [x]
-add handling for generic module swap to AOQuantizable (MoEMapping) [x] -refactor MoEQuantConfig to swap generic modules [x]

TODO
-add dispatch from grouped_mm to linear decomposition of quantized kernel
-compare linear decomposition vs new linear decomposition vs grouped_mm for eager, compile, autotuned compile linear decomposition
-compare linear decomposition vs new linear decomposition for quantized kernels
-add scaled_group_gemm and fbgemm kernel (probably in a new PR)

ISSUE:
the autotuned grouped_mm kernels don't give the correct output, but then work in eager and compile with reduce-overhead. why?

see new_run.log output, first 2 runs are fine, line 144 is nonsense

Test Plan:

sh run.sh

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/_models/mixtral-moe/generate.py
+++ b/torchao/_models/mixtral-moe/generate.py
@@ -14,6 +14,9 @@ import torch._dynamo.config
 import torch._inductor.config
 
 from torchao.utils import get_model_size_in_bytes
+from torchao.prototype.moe_quant import MoEFeedForwardAOQuantizable
+from torchao.quantization.quant_api import _replace_with_custom_fn_if_matches_filter
+from model import MoEFeedForward
 
 torch.manual_seed(0)
 
@@ -187,7 +190,6 @@ def _load_model(checkpoint_path, device, precision):
 
 B_INST, E_INST = "[INST]", "[/INST]"
 
-
 def main(
     prompt: str = "Hello, my name is",
     interactive: bool = False,
@@ -199,6 +201,7 @@ def main(
     checkpoint_path: Path = Path("checkpoints/mistralai/Mixtral-8x7B-v0.1/model.pth"),
     compile: bool = True,
     compile_prefill: bool = False,
+    compile_mode: str = "reduce-overhead",
     moe_quant: Optional[str] = None,
     profile: Optional[Path] = None,
     memory_profile: Optional[Path] = None,
@@ -236,10 +239,12 @@ def main(
         ]
     )
 
-    from torchao.prototype.moe_quant.utils import (
+    from torchao.prototype.moe_quant import (
         MoEQuantConfig,
+        MoEMapping,
         UseFakeExtraDimTensor,
-        cond_ffn_filter,
+        MoEFeedForwardAOQuantizable,
+
     )
     from torchao.quantization.quant_api import (
         Float8DynamicActivationFloat8WeightConfig,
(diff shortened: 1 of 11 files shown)


alexsamardzic:
This one replaces #2325, right?

I'm struggling to run the `run.sh` script (i.e. `generate.py` script), keep getting "CUDA out of memory" errors on H100...  Are you using PyTorch built from source and if not, which PyTorch package version do you have installed?

Would you mind finding `mm_grouped.py` file in your PyTorch installation, then changing `can_use_triton_kernel()` function there to just return `False`, and then re-trying?  This will force eager (non-Triton) version of grouped MM kernel to be used even for max-autotune; namely, I suspect that the garbage output may not be from grouped MM Triton kernel itself, but maybe from max-autotuning the whole layer, and that would test it.

As a side note, it seems that `MoEFeedForwardAOQuantizable` should be imported for [this](https://github.com/pytorch/ao/blob/f51485a28fa0be44d144014b89f9f65c0966172a/torchao/prototype/moe_quant/utils.py#L77) and [this](https://github.com/pytorch/ao/blob/f51485a28fa0be44d144014b89f9f65c0966172a/torchao/prototype/moe_quant/utils.py#L83).

HDCharles:
> This one replaces #2325, right?
> 
> I'm struggling to run the `run.sh` script (i.e. `generate.py` script), keep getting "CUDA out of memory" errors on H100... Are you using PyTorch built from source and if not, which PyTorch package version do you have installed?
> 
> Would you mind finding `mm_grouped.py` file in your PyTorch installation, then changing `can_use_triton_kernel()` function there to just return `False`, and then re-trying? This will force eager (non-Triton) version of grouped MM kernel to be used even for max-autotune; namely, I suspect that the garbage output may not be from grouped MM Triton kernel itself, but maybe from max-autotuning the whole layer, and that would test it.
> 
> As a side note, it seems that `MoEFeedForwardAOQuantizable` should be imported for [this](https://github.com/pytorch/ao/blob/f51485a28fa0be44d144014b89f9f65c0966172a/torchao/prototype/moe_quant/utils.py#L77) and [this](https://github.com/pytorch/ao/blob/f51485a28fa0be44d144014b89f9f65c0966172a/torchao/prototype/moe_quant/utils.py#L83).

can you run it with batch_size 1? 

i'll try the fix

yeah i haven't done the quantization dispatch stuff yet.

torchao/prototype/moe_quant/utils.py
```diff
@@ -220,94 +415,16 @@ def __torch_dispatch__(cls, func, types, args, kwargs):
...
-    FALSE = auto()
-    AS_FALLBACK = auto()
-
-
-@dataclass
-class MoEQuantConfig(AOBaseConfig):
-    """Configuration for applying quantization to MoE
-    Args:
-        `base_config`: normal AO Config
+class DummyModule(torch.nn.Module):
```
vkuzo: I think a better solution is to make torchao APIs work on parameters.  The current workaround is fine for prototype, but we'd want more proper support for non-prototype.

torchao/_models/mixtral-moe/model.py
```diff
@@ -310,7 +310,7 @@ def apply_rotary_emb(x: Tensor, freqs_cis: Tensor) -> Tensor:
 # T'(e) tokens for expert e
 
 
-class MOEFeedForwardAOQuantizable(nn.Module):
+class MoEFeedForwardAOQuantizable(nn.Module):
```
vkuzo: It seems unlikely that people are going to swap their MoE module to AO's version.  Can we just target `torch._grouped_mm` calls directly without requiring a module swap?

alexsamardzic: What would it mean to "target" it specifically?  If given model compiled, the compiled version of this operator will be used anyway, not sure what else torchao could do about it...

alexsamardzic:
> can you run it with batch_size 1?

Nope, with both batch_size 1 and 8, it runs out of memory.

torchao/_models/mixtral-moe/model.py
```diff
@@ -347,6 +347,14 @@ def __init__(self, config):
             torch.empty(config.num_experts, config.intermediate_size, config.dim)
         )  # E, I, D
         self.num_experts = config.num_experts
+        self.perf_is_optimized = False
+
+    def optimize_perf(self):
```
vkuzo: nit: name something like `fuse_w1_w3`, `optimize_perf` is ambiguous

torchao/_models/mixtral-moe/generate.py
```diff
@@ -255,71 +267,62 @@ def main(
...
-            config = MoEQuantConfig(Int8WeightOnlyConfig())
+            config.base_config = Int8WeightOnlyConfig()
 
         elif "int8wo" in moe_quant:
-            config = MoEQuantConfig(
-                Int8WeightOnlyConfig(),
-                use_fake_extra_dim_tensor=UseFakeExtraDimTensor.TRUE,
-            )
+            config.base_config = Int8WeightOnlyConfig()
+            config.use_fake_extra_dim_tensor=UseFakeExtraDimTensor.TRUE
```
vkuzo: what does this do?

torchao/_models/mixtral-moe/generate.py
```diff
@@ -255,71 +267,62 @@ def main(
...
-            config = MoEQuantConfig(
-                Int4WeightOnlyConfig(),
-                use_fake_extra_dim_tensor=UseFakeExtraDimTensor.TRUE,
-            )
+            config.base_config = Int4WeightOnlyConfig()
+            config.use_fake_extra_dim_tensor=UseFakeExtraDimTensor.TRUE
 
         elif "fp8wo-base" in moe_quant:
-            config = MoEQuantConfig(Float8WeightOnlyConfig())
+            config.base_config = Float8WeightOnlyConfig()
```
vkuzo, line 296: I'd personally focus on float8 dynamic quant for now, it's more important to get good perf + accuracy on a small set of techniques than to achieve broad coverage of techniques.

torchao/_models/mixtral-moe/generate.py
```diff
@@ -474,6 +477,12 @@ def callback(x):
         action="store_true",
         help="Whether to compile the prefill (improves prefill perf, but higher compile times)",
     )
+    parser.add_argument(
+        "--compile_mode",
+        type=str,
+        default="reduce-overhead",
+        help="which torch.compile mode to use: reduce-overhead or max-autotune, does nothing if --compile is not set.",
```
vkuzo, line 494: do we expect a meaningful % of users to use max-autotune? if not, I'd say p1 and do it later.

torchao/_models/mixtral-moe/model.py
```diff
@@ -225,41 +225,39 @@ def forward(
...
-#         x = x.view(-1, self.dim)
-#         # T = num_tokens, E = num_experts, D = hidden dim, A = activated experts
-#         # x: [T, D]
-#         scores = self.gate(x) # [T, E]
-#         expert_weights = F.softmax(scores, dim=-1)
-#         expert_weights, expert_indices = torch.topk(expert_weights, self.num_activated_experts, dim=-1) # [T, A], [T, A]
-#         expert_weights /= expert_weights.sum(dim=-1, keepdim=True) # [T, A]
-#         expert_outs = self.cond_ffn(x, expert_indices)
-#         return torch.einsum('tai,ta -> ti', expert_outs, expert_weights)
+class ConditionalFeedForward(nn.Module):
```
vkuzo, line 249: why not just make this use `torch._grouped_mm` and remove the AOQuantizeable version?

torchao/_models/mixtral-moe/model.py
```diff
@@ -225,41 +225,39 @@ def forward(
         y = self.wo(y)
         return y
 
+class MoEFeedForward(nn.Module):
```
vkuzo, line 227: is this the same code as in the original mixtral model definition? if not, do we have a test testing for numerical equivalency?

torchao/prototype/moe_quant/utils.py
HDCharles: We can do that but nothing right now uses grouped_mm, if there's a use case for it, it'll be pretty trivial to add.

torchao/_models/mixtral-moe/model.py
HDCharles: We can do that but nothing right now uses grouped_mm, if there's a use case for it, it'll be pretty trivial to add.

torchao/_models/mixtral-moe/model.py
HDCharles: outdated

torchao/_models/mixtral-moe/generate.py
HDCharles: yeah, I have done that, I thought we'd be able to do dispatch grouped_mm -> linear -> quantized linear for hte rest of the tensor subclasses but it doesn't work so i've gone back to what it was initially and this PR mostly focuses on fp8dq and bf16

torchao/_models/mixtral-moe/generate.py
HDCharles: Initially, to support things generally i created a tensor subclass called FakeExtraDimTensor that takes traditional 2D tensor subclasses and handles the slicing and indexing ops needed for MoE. This enables stuff like ARM quantization where i wasn't able to procure a working dev env to actually natively support the 3D functionality. In practice its slower than the native 3D support but more general which may just not be worth it.

torchao/_models/mixtral-moe/generate.py
HDCharles: @alexsamardzic added it so i wanted to test and see the speedup, at least for these shapes it doesn't make a huge difference so its not hugely important.

torchao/_models/mixtral-moe/model.py
HDCharles: I was hoping to do MoE quant without direct model modification, previously i manually modified the model but with the new swapping stuff its no longer needed.

torchao/_models/mixtral-moe/model.py
HDCharles: yes its the original code. We don't have tests for numerical equivalence because i don't think that implementation is actually a source of truth. It can't actually do batchsize 8 due to the crazy dim join going on.

torchao/dtypes/floatx/float8_layout.py
```diff
@@ -243,13 +253,41 @@ def _(func, types, args, kwargs):
...
+        Float8AQTTensorImpl(
+            args[0].float8_data,
+            args[0].scale,
+            not args[0].transposed,
+            args[0]._layout,
+        ),
+    )
+
+
+@implements(grouped_mm)
```
jerryzh168, line 274: do you want to make changes to Float8Tensor directly? since we are moving away from AQT

HDCharles: i'll do that in a new PR

danielvegamyhre:
> now that the pytorch grouped_mm kernels don't require padding

To clarify, we no longer need to pad empty token groups for experts assigned 0 tokens, so the token group size isn't 0. However, padding to 16 byte alignment for slowest moving dim (stride 1) is still required, at least for fp8 scaled grouped mm. Does this align with your understanding as well, or am I missing something?

torchao/_models/mixtral-moe/generate.py
```diff
@@ -255,71 +265,72 @@ def main(
 
     if moe_quant:
         torch._dynamo.config.capture_dynamic_output_shape_ops = True
-        config = None
+        config = MoEQuantConfig(
```
danielvegamyhre, line 268: minor point but the training config is called `MoETrainingConfig` so it would be more clear to call this `MoEInferenceConfig`, to remove any potential ambiguity present in the name `MoEQuantConfig`

alexsamardzic:
> To clarify, we no longer need to pad empty token groups for experts assigned 0 tokens, so the token group size isn't 0. However, padding to 16 byte alignment for slowest moving dim (stride 1) is still required, at least for fp8 scaled grouped mm. Does this align with your understanding as well, or am I missing something?

The limitations come from TMA, [this test](https://github.com/pytorch/pytorch/blob/da1f608ca33f3062535d0a4866d95db19e72fcbd/test/test_matmul_cuda.py#L502) show what's possible (the most relevant part is "2d/3d", in particular row-major/row-major), and what you stated is correct.  Note that for this particular case, besides 0 tokens, token group could consist of 1, 2, ... or any number of tokens, under condition that slowest moving dim is aligned to 16 bytes.  Both eager and compiled version of grouped MM kernels have checks in place to detect wrong alignments.

PR #2601, opened 2025-07-25
Add daily lib integration test

jerryzh168:
Summary:
* We are separating integration tests since it tends to be more noisy than the other tests
* run it daily instead of on every PR to reduce the cost of running the tests

Test Plan:
CI

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/.github/workflows/lib_integration_test.yml
+++ b/.github/workflows/lib_integration_test.yml
@@ -1,0 +1,62 @@
+name: Run Library Integration Tests
+
+on:
+  schedule:
+    # 6 am PST every day
+    - cron: "0 14 * * *"
+  workflow_dispatch:
+
+concurrency:
+  group: library_integration_test-${{ github.workflow }}-${{ github.ref == 'refs/heads/main' && github.run_number || github.ref }}
+  cancel-in-progress: true
+
+env:
+  HF_TOKEN: ${{ secrets.HF_TOKEN }}
+
+jobs:
+  test:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: SM-89
+            runs-on: linux.g6.4xlarge.experimental.nvidia.gpu
+            torch-lib-spec: '--pre fbgemm-gpu-genai --index-url https://download.pytorch.org/whl/nightly/cu126'
+            gpu-arch-type: "cuda"
+            gpu-arch-version: "12.6"
+          - name: H100
+            runs-on: linux.aws.h100
+            torch-lib-spec: '--pre torchvision torchaudio fbgemm-gpu-genai --index-url https://download.pytorch.org/whl/nightly/cu126'
+            gpu-arch-type: "cuda"
+            gpu-arch-version: "12.4"
+    permissions:
+      id-token: write
+      contents: read
+    uses: pytorch/test-infra/.github/workflows/linux_job_v2.yml@main
+    with:
+      timeout: 60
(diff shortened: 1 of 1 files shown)


.github/workflows/lib_integration_test.yml
```diff
@@ -0,0 +1,59 @@
...
+        conda activate venv
+        export PATH=/opt/rh/devtoolset-10/root/usr/bin/:$PATH
+        python -m pip install --upgrade pip
+        pip install uv
+        uv pip install ${{ matrix.torch-lib-spec }}
+        uv pip install -r dev-requirements.txt
+        pip install .
+        # to not interfere with pytorch version
+        git clone https://github.com/vllm-project/vllm.git
+        cd vllm
```
vkuzo, line 56: is this installing vllm into the torchao directory? if yes, maybe we can change it to install it side by side instead? 

```
root
  /torchao
  /vllm
```

.github/workflows/lib_integration_test.yml
```diff
@@ -0,0 +1,59 @@
...
+
+jobs:
+  test:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: SM-89
+            runs-on: linux.g6.4xlarge.experimental.nvidia.gpu
+            torch-lib-spec: '--pre fbgemm-gpu-genai --index-url https://download.pytorch.org/whl/nightly/cu126'
```
vkuzo, line 24: seems like this is testing nightly pytorch + latest torchao + latest vllm?

how about switching it to stable pytorch + latest torchao + stable vllm, so the only thing moving often is torchao? Otherwise I feel it will be noisy.

vkuzo (approved):

.github/workflows/lib_integration_test.yml
jerryzh168: oh OK, makes sense

.github/workflows/lib_integration_test.yml
jerryzh168: let me switch a bit later or add the stable version test later, since we are actively adding new things now, we need fixes from fbgemm-gpu-genai

jerryzh168: opened https://github.com/pytorch/ao/issues/2607 to track

.github/workflows/lib_integration_test.yml
```diff
@@ -0,0 +1,62 @@
...
+        export TORCHAO_DIR=$(pwd)
+        cd ..
+        # install vllm from source so that we can use the latest
+        # pytorch and fbgemm-gpu-genai
+        git clone https://github.com/vllm-project/vllm.git
+        cd vllm
+        git submodule update --init --recursive
+        python use_existing_torch.py
+        pip install -r requirements/build.txt
+        pip install --no-build-isolation -e .
```
huydhn, line 60: Per my comment in https://github.com/pytorch/ao/pull/2463#discussion_r2231679747, please don't land this PR with `pip install --no-build-isolation -e .` vLLM from source as it kills our H100 cluster.  I have created an issue on our end to have better isolation for this case https://github.com/pytorch-labs/pytorch-gha-infra/issues/766.

huydhn: Alternatively, we could consider using vLLM Docker image to avoid building it altogether https://github.com/pytorch/ao/pull/2610

PR #2603, opened 2025-07-25
Misc fixes to prepare for adding Float8Tensor

jerryzh168:
Summary:
* Moved some float8 related util function to torchao.float8.inference
* renamed _choose_qparams_affine_float8 to _choose_scale_float8
* added hp_value_lb and hp_value_ub to _choose_scale_float8
* added `__all__` to torchao/core/config.py

Test Plan:
pytest test/dtypes/test_affine_quantized_float.py -k test_choose_scale_float8_bounds

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/float8/inference.py
+++ b/torchao/float8/inference.py
@@ -7,7 +7,7 @@
 Defines an nn module designed to be used during inference
 """
 
-from typing import NamedTuple, Optional, Tuple, Union
+from typing import List, NamedTuple, Optional, Tuple, Union
 
 import torch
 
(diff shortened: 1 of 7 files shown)


torchao/float8/inference.py
```diff
@@ -7,7 +7,7 @@
 Defines an nn module designed to be used during inference
 """
 
-from typing import NamedTuple, Optional, Tuple, Union
+from typing import List, NamedTuple, Optional, Tuple, Union
```
vkuzo, line 10: I don't think `torchao.float.inference` is the best directory for this, because `torchao.float8` does not cleanly fit in the directory structure.  Is there a better place for these utils?

torchao/quantization/quant_primitives.py
```diff
@@ -2194,19 +2196,25 @@ def _choose_qparams_affine_float8(
...
+        hp_value_lb (Optional[float]): the lower bound for high precision floating point value for calculating scale
+        hp_value_ub (Optional[float]): the upper bound for high precision floating point value for calculating scale
     """
     quant_max = torch.finfo(float8_dtype).max
     # only tensorwise scaling is supported for now:
     if len(block_size) == 0:
         max_abs = tensor.abs().max()
+        if hp_value_lb is not None or hp_value_ub is not None:
+            max_abs = torch.clamp(max_abs, min=hp_value_lb, max=hp_value_ub)
+        print("max abs:", max_abs)
```
vkuzo: nit: remove

torchao/float8/inference.py
jerryzh168: makes sense, let me try to move this a bit later? I think the most correct place is "torchao/quantization/quantize_/workflows/float8/" which is not created yet

vkuzo: sgtm

PR #2604, opened 2025-07-25
NVFP4 -> Use more of e4m3 range for block_scales

drisspg:
NVFP4 -> Use more of e4m3 range for block_scales


torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -723,18 +723,21 @@ def nvfp4_addmm(func, types, args, kwargs):
...
     Returns:
-        torch.Tensor: Per-tensor scale tensor
+        torch.Tensor: Per-tensor scale for two-level NVFP4 scaling
     """
-    return torch.clamp(amax / F8E4M3_MAX, min=E4M3_EPS, max=F8E4M3_MAX).to(
-        torch.float32
-    )
+    return torch.clamp(
+        amax / (F8E4M3_MAX * F4_E2M1_MAX), min=E4M3_EPS, max=F8E4M3_MAX
+    ).to(torch.float32)
```
gau-nernst: small q: is there a guarantee on the dtype of `amax`? Would it be better to cast `amax` to FP32 before doing division? (so that division is accurate). Then don't need to do FP32 cast after `torch.clamp()`

gau-nernst (approved):

torchao/prototype/mx_formats/nvfp4_tensor.py
drisspg: No guarantee, I think that is cleaner, let me re-arrange

PR #2610, opened 2025-07-26
Add vLLM x TorchAO integration workflow

huydhn:
This adds a workflow to run vLLM x TorchAO tests with the latest vLLM main.  The setup is vLLM main x PyTorch stable x FBGEMM stable x TorchAO PR/main commits.

If there are not enough H100 capacity and you observe queueing, please let me know.  We might need to tweak the workflow to run on H100 only when pushing to main.

### Testing

https://github.com/pytorch/ao/actions/runs/16535494171/job/46769051375

### DIFF
--- a/.github/workflows/vllm_ao_integration_test.yml
+++ b/.github/workflows/vllm_ao_integration_test.yml
@@ -29,7 +29,6 @@ jobs:
       id-token: write
       contents: read
     runs-on: ${{ matrix.runs-on }}
-    if: github.event_name == 'push' || github.event_name == 'workflow_dispatch' || (github.event_name == 'pull_request' && matrix.runs-on != 'linux.aws.h100')
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4


huydhn:
@jerryzh168 ~~I run the same logic to install fbgemm-gen-ai from nightly and build TorchAO, but running pytest currently ends up with this error loading fbgemm https://github.com/pytorch/ao/actions/runs/16535051236/job/46767869902?pr=2610#step:7:400.  Do you know what I did wrong here?~~

I need to install `fbgemm-gen-ai` from stable to avoid rebuilding it.  I think we could stay with that, but let me know if it's ok.  In this setup we have, vLLM main x PyTorch stable x FBGEMM stable x TorchAO PR/main commits

jerryzh168:
current torchao main doesn't work with fbgemm gpu genai stable I think, maybe just use stable for everything and we'll make sure next stable torchao release can work with fbgemm gpu genai stable

huydhn:
> current torchao main doesn't work with fbgemm gpu genai stable I think, maybe just use stable for everything and we'll make sure next stable torchao release can work with fbgemm gpu genai stable

Interesting, the tests pass https://github.com/pytorch/ao/actions/runs/16537610493?pr=2610, it means that torchao main is working with vLLM main.  Is that what we are looking for here?  I guess if we need to test torchao vs fbgemm gpu genai, it makes more sense to have another workflow to test only these twos?

jerryzh168:
> > current torchao main doesn't work with fbgemm gpu genai stable I think, maybe just use stable for everything and we'll make sure next stable torchao release can work with fbgemm gpu genai stable
> 
> Interesting, the tests pass [pytorch/ao/actions/runs/16537610493?pr=2610](https://github.com/pytorch/ao/actions/runs/16537610493?pr=2610), it means that torchao main is working with vLLM main. Is that what we are looking for here? I guess if we need to test torchao vs fbgemm gpu genai, it makes more sense to have another workflow to test only these twos?

OK it could be because we haven't landed everything we want in main yet, and some of our new changes would require fbgemm main

we do have tests to test torchao + fbgemm gpu genai (nightly) in our CI

.github/workflows/vllm_ao_integration_test.yml
```diff
@@ -0,0 +1,91 @@
+name: Run Integration Test against vLLM main
+
+on:
+  push:
+    branches:
+      - main
```
jerryzh168, line 6: nit: nightly might be enough I think

.github/scripts/run_vllm_tests.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+set -eux
+
+# vLLM docker image is using CUDA 12.8 and python 3.12
+pip install --pre fbgemm-gpu-genai --index-url https://download.pytorch.org/whl/cu128
```
jerryzh168, line 6: fbgemm gpu genai nightly would require pytorch nightly I think, and with that we probably need vllm to depend no torch nightly, is that doable with docker? or docker is build with stable pytorch?

PR #2613, opened 2025-07-28
Remove double baseline calculations for CI microbenchmarks

jainapurva:
This pull request introduces significant updates to the benchmarking framework, focusing on measuring latency for eager modes, and adding caching mechanisms for baseline performance. 

* Introduced `_BASELINE_CACHE` to store eager and compile baseline inference times, reducing redundant computations; resulted in reducing the CI runtime substantially.
* Calculate performance results for compile and eager mode
* Removed the `use_torch_compile` parameter as compile and eager performance are being calculated by default


benchmarks/microbenchmarks/benchmark_inference.py
```diff
@@ -34,15 +35,70 @@
...
+# Baseline caching
+#
+# ``_BASELINE_CACHE`` maps a unique key to a tuple
+# ``(eager_baseline_time, compile_baseline_time)``.  See ``_make_cache_key`` for the key
+# construction.  Users should not access this cache directly; it is
+# internal to this module.  The cache intentionally holds the
+# uncompiled base model so that quantized versions can be derived
+# without mutating the cached copy.
+
+_BASELINE_CACHE: Dict[Tuple, Tuple[float, float]] = {}
```
jerryzh168, line 50: nit: add comment for what key is and maybe give an example

benchmarks/microbenchmarks/benchmark_inference.py
```diff
@@ -51,28 +107,46 @@ def run(config: BenchmarkConfig) -> BenchmarkResult:
...
+
+            result.eager_baseline_inference_time_in_ms = eager_baseline_time
+            result.compile_baseline_inference_time_in_ms = compile_baseline_time
+        else:
+            # Retrieve cached values
+            cached_eager_time, cached_compile_time = _BASELINE_CACHE[cache_key]
+            result.eager_baseline_inference_time_in_ms = cached_eager_time
+            result.compile_baseline_inference_time_in_ms = cached_compile_time
+
+        # At this point, ``base_model`` is an uncompiled model ready for quantization,
```
jerryzh168, line 145: `base_model` could be compiled in L124 right?

benchmarks/microbenchmarks/benchmark_inference.py
```diff
@@ -101,24 +175,39 @@ def run(config: BenchmarkConfig) -> BenchmarkResult:
...
         )
 
-        # Calculate speedup w.r.t. baseline
-        result.speedup = round(
-            result.baseline_inference_time_in_ms / result.model_inference_time_in_ms, 2
+        # Compute eager speedup relative to baseline
+        result.eager_speedup_on_baseline = round(
+            result.eager_baseline_inference_time_in_ms
+            / result.eager_model_inference_time_in_ms,
+            2,
```
jerryzh168: nit: pass by keyword arg to show what this is

benchmarks/microbenchmarks/benchmark_inference.py
```diff
@@ -101,24 +175,39 @@ def run(config: BenchmarkConfig) -> BenchmarkResult:
...
-                m_copy, mode=config.torch_compile_mode, fullgraph=True
-            )
-
         # Store result in model for memory profiling
         m_copy._benchmark_result = result
 
-        # Benchmark time to run an inference call for quantized model
+        # Measure inference time for quantized model
+        print("Benchmarking eager quantized model.....")
+        result.eager_model_inference_time_in_ms = model_inference_time_in_ms(
```
jerryzh168: nit: add quantized somewhere in the name?

benchmarks/microbenchmarks/benchmark_inference.py
```diff
@@ -101,24 +175,39 @@ def run(config: BenchmarkConfig) -> BenchmarkResult:
...
+            model=m_copy, input_data=input_data
+        )
+
+        # Measure inference time for compiled quantized model
         print("Benchmarking quantized model.....")
-        result.model_inference_time_in_ms = model_inference_time_in_ms(
+        m_copy = torch.compile(m_copy, mode=config.torch_compile_mode, fullgraph=True)
+        result.compile_model_inference_time_in_ms = model_inference_time_in_ms(
```
jerryzh168: same for this one

benchmarks/microbenchmarks/benchmark_inference.py
```diff
@@ -101,24 +175,39 @@ def run(config: BenchmarkConfig) -> BenchmarkResult:
...
+            / result.eager_model_inference_time_in_ms,
+            2,
+        )
+        # Compute compile speedup relative to baseline
+        result.compile_speedup_on_baseline = round(
+            result.compile_baseline_inference_time_in_ms
+            / result.compile_model_inference_time_in_ms,
+            2,
+        )
+        # Compute compile speedup for quantized model relative to eager quantized model
```
jerryzh168, line 209: do we need to do this comparison? I think it might be more useful to just compare eager quantized v.s. eager baseline and compile quantized v.s. compile baseline, since these shows the speedup in different serving environments

vkuzo:
this seems ok, I think it would be even better if the code was refactored to only measure the baseline once and compare each experiment against it.  This way, complexity is lower and there is no need for a cache.

high level:

```python
baseline_metrics = calc_baseline_metrics(...)
for experiment_config in experiment_configs:
    experiment_metrics = calc_experiment_metrics(...)
    speedup_vs_baseline = calc_speedup(experiment_metrics, baseline_metrics)
```

non-blocking comment, up to you

benchmarks/microbenchmarks/benchmark_inference.py
jainapurva: Yes, sorry it was an older comment

benchmarks/microbenchmarks/benchmark_inference.py
jainapurva: In addition to the basic quantized vs non-quantized in both eager and compile mode, I feel it is necessary to see the relative performance difference between compile and eager mode for a quantized model, as we need to track if the performance difference is increasing/decreasing with updates to compile or with newer kernels. Currently compile gives good performance, but in future, if we plan to improve the eager performance, then this metric will help track the performance difference.

jerryzh168: OK, comments need to be updated I think, I remember you were comparing different things before

torchao/testing/model_architectures.py
```diff
@@ -156,6 +156,7 @@ def create_model_and_input_data(
         high_precision_dtype (torch.dtype): data type of the model
         m, k, n (int): dimensions of the model and input data
     """
+    torch.manual_seed(42)
```
jerryzh168: should this be here?

benchmarks/microbenchmarks/benchmark_inference.py
jainapurva: I was calculating the same values, just renamed the variables. Changed compile_model_inferene_time_in_ms -> quantized_model_compiled_inference_time_in_ms and eager_model_inference_time_in_ms -> quantized_model_eager_inference_time_in_ms

benchmarks/microbenchmarks/benchmark_inference.py
```diff
@@ -34,15 +35,70 @@
     create_model_and_input_data,
 )
 
+# -----------------------------------------------------------------------------
+# Baseline caching
+#
+# ``_BASELINE_CACHE`` maps a unique key constructed using _make_cache_key(config) -> (model_type, m, k, n, high_precision_dtype, device, torch_compile_mode) to a tuple
+# ``(eager_baseline_time, compile_baseline_time)``.  See ``_make_cache_key`` for the key
+# construction.  Users should not access this cache directly; it is
+# internal to this module.
```
jerryzh168, line 44: can you also add a bit detail on when the cache is useful as well

torchao/testing/model_architectures.py
jainapurva: Added it here to ensure the model and data created using this method is consistent. This is the method being used in benchmarking to created models and input_data

jerryzh168: it's probably better to do this in the test file itself I think, changing this here is a bit confusing

jerryzh168 (approved):

PR #2618, opened 2025-07-28
Add Float8BlockwiseLinear for training

danielvegamyhre:
Add Float8BlockwiseLinear for training


vkuzo:
the integration looks good, but IMO we should use pytorch native code for quantization of weights/activations, and only add triton kernels when we're confident they match the pytorch native code with bitwise accuracy.  Without this, it's hard to trust and debug the accuracy of this prototype.

danielvegamyhre:
> the integration looks good, but IMO we should use pytorch native code for quantization of weights/activations, and only add triton kernels when we're confident they match the pytorch native code with bitwise accuracy. Without this, it's hard to trust and debug the accuracy of this prototype.

Makes sense, I updated prior PR to assert bitwise equivalence, let me know what you think.

test/prototype/blockwise_fp8_training/test_blockwise_linear.py
```diff
@@ -0,0 +1,72 @@
...
+import pytest
+import torch
+
+triton = pytest.importorskip("triton", reason="Triton required to run this test")
+
+from torchao.float8.float8_utils import compute_error
+from torchao.prototype.blockwise_fp8_training.linear import Float8BlockwiseLinear
+
+
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA not available")
```
vkuzo, line 25: sm90?

test/prototype/blockwise_fp8_training/test_blockwise_linear.py
```diff
@@ -0,0 +1,72 @@
...
+def test_blockwise_quant_linear_fwd_bwd(
+    in_features,
+    out_features,
+    batch_size,
+    block_size,
+):
+    if in_features % block_size != 0 or out_features % block_size != 0:
+        pytest.skip(f"Dimensions must be divisible by block_size={block_size}")
+
+    torch.random.manual_seed(0)
```
vkuzo: usually people do this one next to the imports, and then use `copy.deepcopy` to create copies of models.  It's unusually to set the seed twice to get the same effect, even if it works.

danielvegamyhre: Makes sense, updated.

PR #2622, opened 2025-07-29
Integrate PARQ with lowbit Arm CPU kernels

metascroy:
This PR integrates PARQ with lowbit CPU kernels by translating PARQ to an LUT tensor.

### DIFF
--- a/test/prototype/test_dynamic_activation_lut.py
+++ b/test/prototype/test_dynamic_activation_lut.py
@@ -3,6 +3,7 @@
 #
 # This source code is licensed under the BSD 3-Clause license found in the
 # LICENSE file in the root directory of this source tree.
+
 import platform
 import sys
 from copy import deepcopy
--- a/.github/workflows/torchao_experimental_test.yml
+++ b/.github/workflows/torchao_experimental_test.yml
@@ -53,6 +53,7 @@ jobs:
           pytest torchao/experimental/tests/test_int8_dynamic_activation_intx_weight.py
           python torchao/experimental/tests/test_embedding_xbit_quantizer.py
           python torchao/experimental/tests/test_quant_passes.py
+          pytest -s test/prototype/test_dynamic_activation_lut.py
       - name: Run kernels/cpu/aarch64/tests
         if: runner.os == 'macOS'
         run: |
@@ -106,7 +107,7 @@ jobs:
   #         conda run -n test-mps-ops-env pip install torch --index-url "https://download.pytorch.org/whl/nightly/cpu"
   #     - name: Print torch version
   #       run: |
-          
+
   #         conda run -n test-mps-ops-env python -c "import torch; print(torch.__version__)"
   #     - name: Install requirements
   #       run: |


test/prototype/test_dynamic_activation_lut.py
```diff
@@ -0,0 +1,161 @@
...
+    quantize_(parq_model, config)
+
+    # Apply dynamic activation to parq model.  This will serve as the LUT reference
+    parq_model_with_dyn_quant = deepcopy(parq_model)
+    quantize_(
+        parq_model_with_dyn_quant,
+        Int8DynamicActivationConfig(),
+        # We have to explicitly provide filter_fn because the default linear filter
+        # excludes modules with AffinQUnatizedTensor weights
+        filter_fn=lambda m, fqn: isinstance(m, torch.nn.Linear),
```
lisjin, line 104: Should I add a patch somewhere so that StretchedAffineQuantizedTensor is included in the default filter_fn?

test/prototype/test_dynamic_activation_lut.py
```diff
@@ -0,0 +1,161 @@
...
+        )
+    )
+    quantize_(lut_model, conversion_config, filter_fn=conversion_config.get_filter_fn())
+
+    # Run both models and compare
+    parq_out = parq_model(activations)
+    parq_with_dyn_quant_out = parq_model_with_dyn_quant(activations)
+    lut_out = lut_model(activations)
+
+    assert torch.allclose(parq_out, parq_with_dyn_quant_out, atol=1e-1, rtol=1e-1)
```
lisjin, line 121: Just curious, is using such a high atol and rtol expected when comparing to the model with 8-bit activations? Maybe this means we should always use quantized activations during QAT in the future.

torchao/prototype/quantization/dynamic_activation_lut/api.py
```diff
@@ -0,0 +1,83 @@
...
+    # Construct LUT as 2 * ([q_min, q_max] - 0.5)
+    assert torch.all(zero_point == -0.5)
+    lut = torch.arange(q_min, q_max + 1)
+    lut = 2 * lut + 1
+
+    # Construct idx values
+    qval_idx = int_data - q_min
+
+    # Construct scale
+    scale = scale.reshape(-1).to(torch.float32)
```
lisjin, line 71: Should this be using weight.dtype instead of torch.float32? This is what we use in [choose_qparams_stretched_affine](https://github.com/pytorch/ao/blob/3515cb66b6b0a8f538d302877a81eb52632eee20/torchao/prototype/parq/quant/quant_api.py#L39), which follows [_choose_qparams_affine](https://github.com/pytorch/ao/blob/main/torchao/quantization/quant_primitives.py#L1533).

lisjin (approved):
This looks great to me—thanks @metascroy!

test/prototype/test_dynamic_activation_lut.py
metascroy: I think most entries are accurate within 2 decimal places, but for CI it's important to give a buffer because we don't want it to be flakey.

But we will know more on dynamic quant accuracy evaluations from the work you're doing now

torchao/prototype/quantization/dynamic_activation_lut/api.py
metascroy: The kernel only supports fp32 scales, so we cast to fp32, no matter the original dtype.

test/prototype/test_dynamic_activation_lut.py
metascroy: Let's hold off on that now.

I'd like to get thoughts from people in ao first

PR #2628, opened 2025-07-29
[bc-breaking] Generalize FakeQuantizeConfig beyond intx

andrewor14:
* (to be filled)

**Summary:** The existing `FakeQuantizeConfig` performs only
intx quantization, but we plan to extend QAT to other dtypes
such as fp8 and nvfp4 in the near future. This is the necessary
refactor before that. Specifically:

```
# New abstract class
FakeQuantizeConfigBase
# Rename
FakeQuantizeConfig -> IntxFakeQuantizeConfig
```

In the future, we will have other types of `FakeQuantizeConfigBase`
for float dtypes that users can pass in instead of the existing
Intx one.

**BC-breaking notes:** For BC, we keep around the old names to
reference the new ones. However, this commit is still BC-breaking
in the sense that a few APIs now accept the abstract
`FakeQuantizeConfigBase` instead. For the most part, this abstract
class will be hidden from the user.

Before:
```
activation_config = FakeQuantizeConfig(torch.int8, "per_token", is_symmetric=False)
weight_config = FakeQuantizeConfig(torch.int4, group_size=32)
```

After:
```
activation_config = IntxFakeQuantizeConfig(torch.int8, "per_token", is_symmetric=False)
weight_config = IntxFakeQuantizeConfig(torch.int4, group_size=32)
```

**Test Plan:**

### DIFF
--- a/torchao/quantization/qat/fake_quantize_config.py
+++ b/torchao/quantization/qat/fake_quantize_config.py
@@ -1,0 +1,262 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+
+# This source code is licensed under the license found in the
+# LICENSE file in the root directory of this source tree.
+
+import abc
+from dataclasses import dataclass
+from typing import Any, Optional, Union
+
+import torch
+
+from torchao.quantization.granularity import (
+    Granularity,
+    PerAxis,
+    PerGroup,
+    PerToken,
+)
+from torchao.quantization.quant_primitives import (
+    _SUB_BYTE_INT_BOUNDS,
+    _SUB_BYTE_UINT_BOUNDS,
+    MappingType,
+    TorchAODType,
+    ZeroPointDomain,
+)
+
+
+@dataclass
+class FakeQuantizeConfigBase(abc.ABC):
+    """
+    Base class for representing fake quantization config.
+    """
+
+    pass
+
+
+@dataclass
+class IntxFakeQuantizeConfig(FakeQuantizeConfigBase):
+    """
+    Config for how to fake quantize weights or activations.
+
+    Args:
+        dtype: dtype to simulate during fake quantization, e.g. torch.int8.
+            For PyTorch versions older than 2.6, you may use `TorchAODType` to represent
+            torch.int1 to torch.int7 instead, e.g. TorchAODType.INT4.
+        granularity: granularity of scales and zero points, e.g. PerGroup(32).
+            We also support the following strings:
+               1) 'per_token': equivalent to PerToken()
+               2) 'per_channel': equivalent to PerAxis(0)
+               3) 'per_group': equivalent to PerGroup(group_size), must be combined
+                   with separate `group_size` kwarg, Alternatively, just set the
+                   `group_size` kwarg and leave this field empty.
+        mapping_type: whether to use symmetric (default) or asymmetric quantization
+            Alternatively, set `is_symmetric` (bool) and leave this field empty.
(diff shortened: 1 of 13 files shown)


drisspg:
Just to confirm, we are changing the name of FakeQuantizeConfig to intxFakeQuantizeConfig, But we are also keeping around the fake quantize config object as a rename of int x fake quantize config ? And then in two releases we will remove it?

andrewor14:
> Just to confirm, we are changing the name of FakeQuantizeConfig to intxFakeQuantizeConfig, But we are also keeping around the fake quantize config object as a rename of int x fake quantize config ? And then in two releases we will remove it?

Yes keeping around `FakeQuantizeConfig` is just for BC. I think many users are using it today (it's part of our recommended API). Will deprecate in a future PR

torchao/quantization/qat/fake_quantize_config.py
```diff
@@ -0,0 +1,262 @@
...
+from torchao.quantization.quant_primitives import (
+    _SUB_BYTE_INT_BOUNDS,
+    _SUB_BYTE_UINT_BOUNDS,
+    MappingType,
+    TorchAODType,
+    ZeroPointDomain,
+)
+
+
+@dataclass
```
drisspg: I dont think you need this dataclass decorator here

torchao/quantization/qat/fake_quantize_config.py
```diff
@@ -0,0 +1,262 @@
...
+    scale_precision: torch.dtype
+    zero_point_precision: torch.dtype
+    zero_point_domain: ZeroPointDomain
+    is_dynamic: bool = True
+    range_learning: bool = False
+    eps: Optional[float] = None
+
+    def __init__(
+        self,
+        dtype: Union[torch.dtype, TorchAODType],
```
drisspg, line 97: I wonder if you can type this as `Literal[...]` so that it only allows for int inputs

torchao/quantization/qat/fake_quantize_config.py
```diff
@@ -0,0 +1,262 @@
...
+        self.mapping_type = self._get_mapping_type(mapping_type, is_symmetric)
+        self.scale_precision = scale_precision
+        self.zero_point_precision = zero_point_precision
+        self.zero_point_domain = zero_point_domain
+        self.is_dynamic = is_dynamic
+        self.range_learning = range_learning
+        self.eps = eps
+
+        # Validate dtype
+        all_dtypes = [torch.int8, torch.uint8]
```
drisspg, line 123: similar to this

drisspg (approved):

torchao/quantization/qat/fake_quantize_config.py
andrewor14: we actually have a lot of dtypes we allow, like all of `int2-8` and `uint2-8`, will be too verbose I think

drisspg: just define a `Allowed Types` above and use it

andrewor14: Hmm I just tried it but didn't really like it. I think I prefer a simpler signature like just `torch.dtype` (we can drop `TorchAODType` soon, only needed for pytorch 2.5 and before) and do the validation in init. Also this is just moving existing code

andrewor14: Hmm I just tried it but didn't really like it. I think I prefer a simpler signature like just `torch.dtype` (we can drop `TorchAODType` soon, only needed for pytorch 2.5 and before) and do the validation in init

PR #2629, opened 2025-07-29
New multi-step QAT API

andrewor14:
* #2666
* #2641
* __->__ #2629

**Summary:** This commit adds a new multi-step QAT API with the
main goal of simplifying the existing UX. The new API uses the
same `QATConfig` for both the prepare and convert steps, and
automatically infers the fake quantization configs based on
a PTQ base config provided by the user:

```Py
    quantize_,
    Int8DynamicActivationInt4WeightConfig
)
from torchao.quantization.qat import QATConfig

\# prepare
base_config = Int8DynamicActivationInt4WeightConfig(group_size=32)
qat_config = QATConfig(base_config, step="prepare")
quantize_(m, qat_config)

\# train (not shown)

\# convert
quantize_(m, QATConfig(base_config, step="convert"))
```

The main improvements include:
- A single config for both prepare and convert steps
- A single quantize_ for convert (instead of 2)
- No chance for incompatible prepare vs convert configs
- Much less boilerplate code for most common use case
- Simpler config names

For less common use cases such as experimentation, users can
still specify arbitrary fake quantization configs for
activations and/or weights as before. This is still important
since there may not always be a corresponding PTQ base config.
For example:

```Py
from torchao.quantization.qat import IntxFakeQuantizeConfig, QATConfig

activation_config = IntxFakeQuantizeConfig(torch.int8, "per_token", is_symmetric=False)
weight_config = IntxFakeQuantizeConfig(torch.int4, group_size=32)
qat_config = QATConfig(
    activation_config=activation_config,
    weight_config=weight_config,
    step="prepare",
)
quantize_(model, qat_config)

\# train and convert same as above (not shown)
```

**BC-breaking notes:** This change by itself is technically not
BC-breaking since we keep around the old path, but will become
so when we deprecate and remove the old path in the future.

Before:
```Py
\# prepare
activation_config = IntxFakeQuantizeConfig(torch.int8, "per_token", is_symmetric=False)
weight_config = IntxFakeQuantizeConfig(torch.int4, group_size=32)
qat_config = IntXQuantizationAwareTrainingConfig(activation_config, weight_config),
quantize_(model, qat_config)

\# train (not shown)

\# convert
quantize_(model, FromIntXQuantizationAwareTrainingConfig())
quantize_(model, Int8DynamicActivationInt4WeightConfig(group_size=32))
```

After: (see above)

**Test Plan:**
```
```


torchao/quantization/qat/api.py
```diff
@@ -11,19 +11,213 @@
...
+        ValueError: If `base_config` is None but `step` is "convert"
+        ValueError: If the config is applied on a module that is not a
+            `torch.nn.Linear` or `torch.nn.Embedding`, or it is applied on
+            `torch.nn.Embedding` with an activation config
+    """
+
+    base_config: Optional[AOBaseConfig]
+    activation_config: Optional[FakeQuantizeConfigBase]
+    weight_config: Optional[FakeQuantizeConfigBase]
+    step: str
```
vkuzo: nit: make a string enum, and accept the user argument as both a string (no change) and an enum value

vkuzo:
> This change by itself is technically not
BC-breaking since we keep around the old path, but will become
so when we deprecate and remove the old path in the future.

Let's start a github issue with timeline + owner of deprecation, and throw a warning on the old path pointing them to that issue?

andrewor14:
> Let's start a github issue with timeline + owner of deprecation, and throw a warning on the old path pointing them to that issue?

Sure, filed one here: https://github.com/pytorch/ao/issues/2630

torchao/quantization/qat/api.py
```diff
@@ -5,25 +5,230 @@
...
+from .embedding import FakeQuantizedEmbedding
 from .fake_quantize_config import (
     FakeQuantizeConfig,  # noqa: F401, for BC
     FakeQuantizeConfigBase,
+    _infer_fake_quantize_configs,
 )
+from .linear import FakeQuantizedLinear
+
+
+class QATConfigStep(str, Enum):
```
drisspg: Ohhh I wish we can get to 3.11 and use StrEnum

jerryzh168: nit: just `QATStep`?

jerryzh168: also do we need a load/prepare_for_loading step as well? is there a use case to load a QAT model

andrewor14: > also do we need a load/prepare_for_loading step as well? is there a use case to load a QAT model

don't think so, after convert it should look like a regular PTQ'ed model. Also we can always add it later if we need to

torchao/quantization/qat/api.py
```diff
@@ -5,25 +5,230 @@
...
+        self.activation_config = activation_config
+        self.weight_config = weight_config
+        self.step = step
+        self.__post_init__()
+
+    def __post_init__(self):
+        self.step = self.step.lower()
+        all_step_values = [s.value for s in QATConfigStep]
+        if self.step not in all_step_values:
+            raise ValueError("`step` must be one of %s" % all_step_values)
```
jerryzh168: nit: this can be removed now? will it be cause by the enum itself?

andrewor14: seems like it's still needed: https://github.com/pytorch/ao/blob/551703147e050e8770c411019a7230de87c20e72/test/quantization/test_qat.py#L1288

jerryzh168 (approved):
Thanks, LGTM

torchao/quantization/qat/api.py
```diff
@@ -5,25 +5,230 @@
...
+        self.activation_config = activation_config
+        self.weight_config = weight_config
+        self.step = step
+        self.__post_init__()
+
+    def __post_init__(self):
+        self.step = self.step.lower()
+        all_step_values = [s.value for s in QATStep]
+        if self.step not in all_step_values:
+            raise ValueError("`step` must be one of %s" % all_step_values)
```
jerryzh168: nit: I think fstrings is more modern? `f"... {all_step_values}"`

PR #2635, opened 2025-07-30
Clarifying the meaning of VERSION in AOBaseConfig

jerryzh168:
Summary:
the VERSION means default version for child configs, it should never change. Also added test to enforce this

Test Plan:
python test/core/test_config.py
Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/core/test_config.py
+++ b/test/core/test_config.py
@@ -187,5 +187,19 @@ def test_version_mismatch():
             config_from_dict(reconstructable)
 
 
+def test_default_version():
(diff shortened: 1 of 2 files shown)


test/core/test_config.py
```diff
@@ -187,5 +187,16 @@ def test_version_mismatch():
             config_from_dict(reconstructable)
 
 
+def test_default_version():
+    """Making sure the default version for a new config inheritign from AOBaseConfig is always 1"""
```
vkuzo: nit: add the reason why it should always be 1 here

PR #2641, opened 2025-07-30
Deprecate old QAT APIs

andrewor14:
* (to be filled)

**Summary:** Deprecates QAT APIs that should no longer be used.
Print helpful deprecation warning to help users migrate.

**Test Plan:**
```
python test/quantization/test_qat.py -k test_qat_api_deprecation
```

### DIFF
--- a/torchao/quantization/qat/api.py
+++ b/torchao/quantization/qat/api.py
@@ -24,6 +24,7 @@ from .fake_quantize_config import (
     _infer_fake_quantize_configs,
 )
 from .linear import FakeQuantizedLinear
+from .utils import _log_deprecation_warning
 
 
 class QATConfigStep(str, Enum):
@@ -224,11 +225,11 @@ def _qat_config_transform(
         return _QUANTIZE_CONFIG_HANDLER[type(base_config)](module, base_config)
 
 
-# TODO: deprecate
 @dataclass
 class IntXQuantizationAwareTrainingConfig(AOBaseConfig):
     """
-    (Will be deprecated soon)
+    (Deprecated) Please use :class:`~torchao.quantization.qat.QATConfig` instead.
+
     Config for applying fake quantization to a `torch.nn.Module`.
     to be used with :func:`~torchao.quantization.quant_api.quantize_`.
 
@@ -256,9 +257,13 @@ class IntXQuantizationAwareTrainingConfig(AOBaseConfig):
     activation_config: Optional[FakeQuantizeConfigBase] = None
     weight_config: Optional[FakeQuantizeConfigBase] = None
 
+    def __post_init__(self):
+        _log_deprecation_warning(self)
+
 
 # for BC
-intx_quantization_aware_training = IntXQuantizationAwareTrainingConfig
+class intx_quantization_aware_training(IntXQuantizationAwareTrainingConfig):
+    pass
 
 
 @register_quantize_module_handler(IntXQuantizationAwareTrainingConfig)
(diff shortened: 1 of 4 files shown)


torchao/quantization/qat/utils.py
```diff
@@ -104,3 +106,42 @@ def _get_qmin_qmax(n_bit: int, symmetric: bool = True):
...
+    Log a helpful deprecation message pointing users to the new QAT API,
+    only once per deprecated class.
+    """
+    global _LOGGED_DEPRECATED_CLASSES
+    if old_api_object.__class__ in _LOGGED_DEPRECATED_CLASSES:
+        return
+    _LOGGED_DEPRECATED_CLASSES.add(old_api_object.__class__)
+    logger = logging.getLogger(old_api_object.__module__)
+    logger.warning(
+        """'%s' is deprecated and will be removed in a future release. Please use the following API instead:
```
jerryzh168, line 117: is there a concrete time for deprecation?

jerryzh168 (approved):

torchao/quantization/qat/api.py
```diff
@@ -256,9 +257,13 @@ class IntXQuantizationAwareTrainingConfig(AOBaseConfig):
     activation_config: Optional[FakeQuantizeConfigBase] = None
     weight_config: Optional[FakeQuantizeConfigBase] = None
 
+    def __post_init__(self):
```
jerryzh168, line 260: I heard typing_extensions.deprecated is more IDE friendly: https://github.com/pytorch/pytorch/pull/153892#discussion_r2097943320

torchao/quantization/qat/utils.py
andrewor14: yeah it's in the issue, deprecated this version (0.13.0) and removed the next (0.14.0)

torchao/quantization/qat/api.py
andrewor14: hmm I don't think torchao has a dependency on `typing_extensions`?

andrewor14: hmm I don't think torchao has a dependency on `typing_extensions`? Not sure if we want to add a new one

andrewor14: hmm I don't think torchao has a dependency on `typing_extensions`? Do we want to add a new one for this?

jerryzh168: I see, OK probably not worth to add deps since it looks like right now there is no deps for torchao

torchao/quantization/qat/utils.py
```diff
@@ -104,3 +106,42 @@ def _get_qmin_qmax(n_bit: int, symmetric: bool = True):
         qmin = 0
         qmax = 2**n_bit - 1
     return (qmin, qmax)
+
+
+# log deprecation warning only once per class
+_LOGGED_DEPRECATED_CLASSES = set[type]()
```
jerryzh168: actually I'm also looking at how to do warning once, does `warnings.warn` work for you?

drisspg: the global set approach is the only way to do it

jerryzh168: it seems that warnings.warn by default only prints once though? https://docs.python.org/3/library/warnings.html#the-warnings-filter

andrewor14: Looks like I was able to get `warnings.warn` to log only once, updated

PR #2646, opened 2025-07-31
skip rocm for moe training tests

danielvegamyhre:
ROCm is not supported for MoE training right now, and a couple new tests were added that don't include the pytest skip decorator.

This will resolve these failures: https://github.com/pytorch/ao/actions/runs/16626977101/job/47046001244


jainapurva (approved):

danielvegamyhre:
test failures are unrelated, merging

petrex:
Thanks. To the best of my knowledge, that kernel relies on CUTLASS. AMD/ROCm is currently working on comparable support.

Issue #2649, opened 2025-07-31
Deprecation for Float8DynamicActivationFloat8WeightConfig and Float8WeightOnlyConfig and the models


jerryzh168:
This issue is tracking the deprecation of the (1) configs (2) model checkpoints quantized with these configs.

What is deprecated:
1. We added VERSION 2 config in https://github.com/pytorch/ao/pull/2463, and switched the default version to 2 in TBD, the VERSION 1 config is now deprecated, please use VERSION 2 config to quantize the model
2. the quantized checkpoints quantized with VERSION 1 config previously is deprecated as well, and we plan to remove the ability to load these checkpoints in March 2026 (9 months from now)

PR #2650, opened 2025-08-01
Bump version for float8 dynamic quant and weight only quant configs

jerryzh168:
Stacked PRs:
 * #2710
 * #2687
 * #2474
 * #2651
 * __->__#2650

--- --- ---

### Bump version for float8 dynamic quant and weight only quant configs

Summary:
This PR changes the default VERSION for Float8DynamicActivationFloat8WeightConfig and Float8WeightOnlyConfig from 1 to 2
and makes the VERSION 1 config and VERSION 1 quantized models deprecated, more details in: https://github.com/pytorch/ao/issues/2649

Also extended current config serialization to work with multiple config versions

Deprecation Note:
```
from transformers import AutoModelForCausalLM, AutoTokenizer

model_name = "torchao-testing/opt-125m-float8dq-row-v1-0.13-dev"
quantized_model = AutoModelForCausalLM.from_pretrained(
    model_name,
    torch_dtype="bfloat16",
    device_map="cuda",
)

/data/users/jerryzh/ao/torchao/core/config.py:249: UserWarning: Stored version is not the same as current default version of the config: stored_version=1, current_version=2, please check the deprecation warning
  warnings.warn(
/data/users/jerryzh/ao/torchao/dtypes/floatx/float8_layout.py:113: UserWarning: Models quantized with VERSION 1 of Float8DynamicActivationFloat8WeightConfig is deprecated and will no longer be supported in a future release, please upgrade torchao and quantize again, or download a newer torchao checkpoint, see https://github.com/pytorch/ao/issues/2649 for more details
  warnings.warn(

```

Suggestion: upgrade torchao to 0.13 and later and generate the checkpoint again:
```
```
Or download the checkpoint again (please let us know if the checkpoint is not updated)

Test Plan:
tested with serializing a model with VERSION 1 config and load it, and checks warnings are properly printed
```
python test/integration/test_loading_deprecated_checkpoint.py
```

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/core/test_config.py
+++ b/test/core/test_config.py
@@ -7,6 +7,7 @@
 import json
 import os
 import tempfile
+import warnings
 from dataclasses import dataclass
 from unittest import mock
 
@@ -15,7 +16,6 @@ import torch
 
 from torchao.core.config import (
     AOBaseConfig,
-    VersionMismatchError,
     config_from_dict,
     config_to_dict,
 )
@@ -151,7 +151,9 @@ def test_reconstructable_dict_file_round_trip(config):
 # Define a dummy config in a non-allowed module
 @dataclass
 class DummyNonAllowedConfig(AOBaseConfig):
-    VERSION = 2
+    # NOTE: must be `version: int` (with type annotations) to
+    # overload the version variable from AOBaseConfig
+    version: int = 2
     value: int = 42
 
 
@@ -172,11 +174,11 @@ def test_disallowed_modules():
         reconstructed = config_from_dict(reconstructable)
         assert isinstance(reconstructed, DummyNonAllowedConfig)
         assert reconstructed.value == 42
-        assert reconstructed.VERSION == 2
+        assert reconstructed.version == 2
 
 
 def test_version_mismatch():
-    """Test that version mismatch raises an error during reconstruction."""
+    """Test that version mismatch prints a warning during reconstruction."""
     # Create a config
     dummy_config = DummyNonAllowedConfig()
     reconstructable = config_to_dict(dummy_config)
@@ -186,17 +188,19 @@ def test_version_mismatch():
 
     # Patch to allow the module but should still fail due to version mismatch
     with mock.patch("torchao.core.config.ALLOWED_AO_MODULES", {__name__}):
-        with pytest.raises(
-            VersionMismatchError,
-            match="Version mismatch for DummyNonAllowedConfig: stored version 1 != current version 2",
-        ):
+        with warnings.catch_warnings(record=True) as caught_warnings:
             config_from_dict(reconstructable)
+            assert any(
+                "Stored version is not the same as current default version of the config"
+                in str(w.message)
+                for w in caught_warnings
+            ), "Didn't get expected warning message for version mismatch"
 
 
 def test_default_version():
     """Making sure the default version for a new config inheriting from AOBaseConfig is always 1
-    because it's the default VERSION that all children has when they haven't explicitly
-    defined a VERSION class variable
+    because it's the default version that all children has when they haven't explicitly
+    defined a version class variable
     """
 
     @dataclass
@@ -204,7 +208,7 @@ def test_default_version():
         pass
 
     config = DummyConfig()
-    assert config.VERSION == 1, "Default version must be 1"
+    assert config.version == 1, "Default version must be 1"
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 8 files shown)


test/core/test_config.py
```diff
@@ -186,11 +186,13 @@ def test_version_mismatch():
...
-            VersionMismatchError,
-            match="Version mismatch for DummyNonAllowedConfig: stored version 1 != current version 2",
-        ):
+        with warnings.catch_warnings(record=True) as caught_warnings:
             config_from_dict(reconstructable)
+            assert any(
+                "Stored version is not the same as current default version of the config"
+                in str(w.message)
+                for w in caught_warnings
+            ), "Din't get expected warning message for version mismatch"
```
drisspg: nit: didn't

drisspg:
This pr is a little confusing, we can do this a number of ways , still continue to 
* support serialization of version 1, this looks like maintaining a load path so that configs w/ version one cant transformed into version 2
* the argument may be that this is alot of card and hard to do correct ( I am not sure of this) In which case it looks like we are just warning and not erroring? Why is that

jerryzh168:
> This pr is a little confusing, we can do this a number of ways , still continue to
> 
> * support serialization of version 1, this looks like maintaining a load path so that configs w/ version one cant transformed into version 2
> * the argument may be that this is alot of card and hard to do correct ( I am not sure of this) In which case it looks like we are just warning and not erroring? Why is that

So in terms of config:
* version 1 can be accessed through explicitly specifying Float8DynamicActivationFloat8WeightConfig(..., VERSION=1) (but we will break this in 0.14)
* version 2 is the default after the PR

in terms of checkpoints
* version 1 checkpoint can still continue to be loaded, with a warning, we will break this path in 0.14 (in 2 releases)


in terms of converting checkpoints, that doesn't sound necessary, the reasoning is the exposure of current float8 checkpoint is still limited and it's OK to break and just not keep two code paths

> In which case it looks like we are just warning and not erroring? Why is that

I didn't quite get the question, but please let me know if it's still unclear

drisspg:
Ci failures look real

torchao/quantization/quant_api.py
```diff
@@ -1506,6 +1506,9 @@ class Float8WeightOnlyConfig(AOBaseConfig):
 
 def _float8_weight_only_quant_tensor(weight, config):
     if config.VERSION == 1:
+        warnings.warn(
```
vkuzo, line 1509: I missed it in previous PRs, but why `VERSION` instead of `version`?  If there is no good reason, can we change it before it's too late?

jerryzh168: it's from here: https://github.com/pytorch/ao/blob/3b4bc9869d933927b2547d8231feab69789a80d4/torchao/core/config.py#L61

should we change both or ignore `VERSION` in TorchAOBaseConfig?

jerryzh168: it is a bit confusing now since now it's a instance variable instead of a class variable

jerryzh168: OK updated it to instance variable and renamed to `version`

torchao/core/config.py
```diff
@@ -114,13 +106,13 @@ def default(self, o):
...
             for f in dataclasses.fields(o):
-                if f.name != "VERSION":
+                if f.name != "version":
                     data_dict[f.name] = self.encode_value(getattr(o, f.name))
 
             return {
                 # Only store the class name for dataclasses too
                 "_type": o.__class__.__name__,
-                "_version": getattr(o.__class__, "VERSION", 1),
+                "_version": getattr(o, "version", 1),
```
drisspg: should the 3rd arg here be the _DEFAULT_VERSION?

torchao/core/config.py
```diff
@@ -253,10 +244,11 @@ def config_from_dict(data: Dict[str, Any]) -> AOBaseConfig:
             f"Failed to find class {type_path} in any of the allowed modules: {allowed_modules_str}"
         )
 
-    # Check version - require exact match
```
drisspg, line 256: I feel like in general versions will error right? like shouldnt we have this path but just let instances figure out if they should error or not at deserialization

torchao/core/config.py
jerryzh168: oh, makes sense

torchao/core/config.py
jerryzh168: they can still error, see quant_api.py, they can error when `config.version` has a deprecated version number

but erroring out here would mean older version are not supported at all

drisspg (approved):

PR #2652, opened 2025-08-01
fix bc breakage flex path

msaroufim:
Update because of https://github.com/pytorch/pytorch/pull/159437

### DIFF
--- a/torchao/prototype/inductor/int8_sdpa_lowering.py
+++ b/torchao/prototype/inductor/int8_sdpa_lowering.py
@@ -3,10 +3,14 @@ from typing import Optional
 import sympy
 import torch
 from torch._inductor.ir import ChoiceCaller, FixedLayout, TensorBox, get_fill_order
+
 try:
     from torch._inductor.kernel.flex_attention import construct_strides, maybe_realize
 except ModuleNotFoundError:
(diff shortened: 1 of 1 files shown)


jerryzh168 (approved):

torchao/prototype/inductor/int8_sdpa_lowering.py
```diff
@@ -3,7 +3,14 @@
 import sympy
 import torch
 from torch._inductor.ir import ChoiceCaller, FixedLayout, TensorBox, get_fill_order
-from torch._inductor.kernel.flex_attention import construct_strides, maybe_realize
+
+try:
+    from torch._inductor.kernel.flex_attention import construct_strides, maybe_realize
```
drisspg: we should just copy these helpers over

msaroufim: The main con is now depending on 2 other internal functions `realize_inputs` and `tree_map` but perhaps those are fairly stable now

drisspg: yeah but my functions are private too lol -> I will break (if I need to with no regards for ao)

PR #2644, opened 2025-07-31
Convert model inference test from pytest to unittest

namgyu-youn:
**Summary:**
Converts TorchAO model (`Transformer`) inference test from Pytest into Unittest. Annotations are added for better readability.
- Related to: #1621 

cc @osbm

PR #2400, opened 2025-06-18
Make AWQ more general

jerryzh168:
Summary:
* Added AWQConfig that takes a base config and made corresponding changes
in other parts of the flow

Test Plan:
Tested on Phi4-mini and Qwen3-8B

Qwen3-8B

|Task | calibration_limit | no-awq | awq |
|-----|------------------| ------|------|
|leaderboard_math_hard (v3) | 2 | 0.3543 | 0.4371 |
|gpqa_main_zeroshot | 50 | 0.32 | 0.36 |
|mmlu | 5 | 0.7372 | 0.7463 |
|bbh | 1 | 0.7385 | 0.7556|

Phi4-mini

| Task | calibration_limit | no-awq | awq |
|------|------------------|--------|-----|
| mmlu_pro | 2 | 0.4057 | 0.4757 |
| gsm8k | 5 | 0.72 | 0.76 |

Reviewers:

Subscribers:

Tasks:

Tags:

PR #2659, opened 2025-08-01
Convert SmoothQuant test to unittest

namgyu-youn:
Fix part of #1621 

Migrate test file (`test/prototype/test_smoothquant.py`) framework from pytest to unittest.

### DIFF
--- a/test/prototype/test_smoothquant.py
+++ b/test/prototype/test_smoothquant.py
@@ -8,6 +8,7 @@ import unittest
 from copy import deepcopy
 
 import torch
+from torch.testing._internal import common_utils
 
 from torchao.prototype.smoothquant import (
     SmoothQuantConfig,
@@ -21,7 +22,6 @@ from torchao.quantization.utils import (
     dequantize_per_channel,
     dynamically_quantize_per_channel,
 )
-from torchao.testing import common_utils
 from torchao.utils import (
     TORCH_VERSION_AT_LEAST_2_5,
 )
@@ -73,10 +73,6 @@ class TestSmoothQuant(unittest.TestCase):
     @common_utils.parametrize("input_dtype", [torch.float, torch.bfloat16, torch.half])
     def test_smoothquant_accuracy(self, bias, alpha, quant_mode, device, input_dtype):
         """Test the margin error of SmoothQuant across bias, alpha, dtype, etc."""
-        self._run_compute_accuracy_test(bias, alpha, quant_mode, device, input_dtype)
-
-    def _run_compute_accuracy_test(self, bias, alpha, quant_mode, device, input_dtype):
-        """Single compute accuracy test"""
 
         class SimpleLinear(torch.nn.Module):
             def __init__(self, bias: bool):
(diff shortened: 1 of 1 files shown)


jerryzh168 (approved):
Thanks!

jerryzh168:
@namgyu-youn if you are interested in contributing more, making smoothquant API similar to AWQ (https://github.com/pytorch/ao/pull/2400) is a good task

we are planning to graduate these two features to official support

namgyu-youn:
> @namgyu-youn if you are interested in contributing more, making smoothquant API similar to AWQ (#2400) is a good task
> 
> we are planning to graduate these two features to official support

@jerryzh168 Thanks for your suggestion. I will look into it before resolving more #1621 . Also, could you check #2644? It fixes part of #1621 (same as this PR).

test/prototype/test_smoothquant.py
```diff
@@ -62,33 +63,32 @@ def setUpClass(cls):
             # This test case will trigger recompilation many times, so set a large cache_size_limit here
             torch._dynamo.config.cache_size_limit = 128
 
-        # Define test parameter ranges
-        cls.bias_options = [True, False]
-        cls.alpha_options = [None, 0.5, 0.75]  # None means conventional quantization
-        cls.quant_mode_options = ["static", "dynamic"]
-        cls.devices = ["cpu"] + (["cuda"] if torch.cuda.is_available() else [])
-        cls.input_dtypes = (torch.float, torch.bfloat16, torch.half)
+    def _setup_quantized_model(self, model, alpha, quant_mode, calibration_data):
```
jerryzh168: I feel it's probably fine to inline these as well, there are not many lines of code here

test/prototype/test_smoothquant.py
```diff
@@ -53,143 +51,240 @@ def forward(self, x):
...
+    @common_utils.parametrize("quant_mode", ["static", "dynamic"])
+    @common_utils.parametrize(
+        "device", ["cpu"] + (["cuda"] if torch.cuda.is_available() else [])
+    )
+    @common_utils.parametrize("input_dtype", [torch.float, torch.bfloat16, torch.half])
+    def test_smoothquant_accuracy(self, bias, alpha, quant_mode, device, input_dtype):
+        """Test the margin error of SmoothQuant across bias, alpha, dtype, etc."""
+        self._run_compute_accuracy_test(bias, alpha, quant_mode, device, input_dtype)
+
+    def _run_compute_accuracy_test(self, bias, alpha, quant_mode, device, input_dtype):
```
jerryzh168: nit: can also remove this function, seems no need to define a function here

test/prototype/test_smoothquant.py
```diff
@@ -53,143 +51,240 @@ def forward(self, x):
...
+            torch.testing.assert_close(
+                q_out,
+                reference_out.to(input_dtype),
+                atol=tolerance,
+                msg=f"Quantized output differs from reference for "
+                f"bias={bias}, alpha={alpha}, quant_mode={quant_mode}, "
+                f"device={device}, dtype={input_dtype}",
+            )
+
+    def _compute_reference_out(self, m_ref, data, alpha, quant_mode, bias, input_dtype):
```
jerryzh168: if it's used only once, I'd suggest not to define new functions

test/prototype/test_smoothquant.py
```diff
@@ -53,143 +51,240 @@ def forward(self, x):
...
-
-    assert out is not None
-    assert save_load_out is not None
-    assert torch.allclose(out, save_load_out)
+    @common_utils.parametrize("input_dtype", [torch.float, torch.bfloat16, torch.half])
+    def test_save_load_recipe(self, alpha, quant_mode, device, input_dtype):
+        """Test save/load recipe functionality."""
+        self._run_save_load_recipe_test(alpha, quant_mode, device, input_dtype)
+
+    def _run_save_load_recipe_test(self, alpha, quant_mode, device, input_dtype):
```
jerryzh168: also this one

test/prototype/test_smoothquant.py
namgyu-youn: Thanks, those functions are undefined for more brevity in https://github.com/pytorch/ao/pull/2659/commits/3ffdd10a280dc53614390af2d6d1d3edc5de401f .

test/prototype/test_smoothquant.py
```diff
@@ -53,143 +51,227 @@ def forward(self, x):
...
-    is_observed_linear = lambda m, fqn: isinstance(m, SmoothQuantObservedLinear)
-    quantize_(m, SmoothQuantConfig(), is_observed_linear)
-    with torch.inference_mode():
+class TestSmoothQuant(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        """Set up class-level configuration for tests."""
+        # Skip tests on ROCm (AMD GPU) due to compatibility issues
+        if torch.version.hip is not None:
+            raise unittest.SkipTest("Skipping the tests in ROCm")
```
jerryzh168: nit: does `unittest.skip(...)` work?

namgyu-youn: @jerryzh168 https://docs.pytorch.org/docs/stable/notes/hip.html shows how `torch.version.hip` works. It is `None` in CUDA, `str (ROCm version)` in ROCm, and `None` in CPU-only. That is, this test is skipped only when ROCm available.

jerryzh168: oh I mean using `unittest.skip` instead of raise the SkipTest error

namgyu-youn: I see, but there is no conditioner in the `unittest.skip` . How about using `unittest.skipif`? See https://docs.python.org/3/library/unittest.html#unittest.skipIf for the reference, and I like this change because the code can be more brevity

jerryzh168: yeah sounds good, thanks

PR #2660, opened 2025-08-01
Replace `torch.norm` with `torch.linalg.vector_norm` for PyTorch future update

namgyu-youn:
Based on https://docs.pytorch.org/docs/stable/generated/torch.norm.html, `torch.norm` will be deprecated. Therefore, this update replaces `torch.norm()` with `torch.linalg.vector_norm()` for the foreseeable future.
- See also : https://github.com/search?q=repo%3Apytorch%2Fao%20.norm(&type=code

### DIFF
--- a/test/prototype/test_blockwise_triton.py
+++ b/test/prototype/test_blockwise_triton.py
@@ -41,7 +41,7 @@ def test_blockwise_quant_dequant(_, N, K, dtype):
     x = torch.randn(N, K).cuda()
     qx, s = fp8_blockwise_weight_quant(x, dtype=dtype)
     x_reconstructed = fp8_blockwise_weight_dequant(qx, s)
-    error = torch.norm(x - x_reconstructed) / torch.norm(x)
+    error = torch.linalg.vector_norm(x - x_reconstructed) / torch.linalg.vector_norm(x)
     print(f"Relative Error: {error.item():.6f}")
 
     assert error < 0.1, "Quant-Dequant error is too high"
@@ -66,7 +66,7 @@ def test_blockwise_fp8_gemm(M, N, K, dtype):
     A_q, A_s = fp8_blockwise_act_quant(A, dtype=dtype)
     B_q, B_s = fp8_blockwise_weight_quant(B, dtype=dtype)
     C_q = blockwise_fp8_gemm(A_q, A_s, B_q, B_s)
-    error = torch.norm(C - C_q) / torch.norm(C)
+    error = torch.linalg.vector_norm(C - C_q) / torch.linalg.vector_norm(C)
     print(f"Relative Error: {error.item():.6f}")
 
     assert error < 0.1, "Quantize gemm error is too high"
--- a/test/prototype/test_quantized_training.py
+++ b/test/prototype/test_quantized_training.py
@@ -213,7 +213,9 @@ class TestQuantizedTraining(TestCase):
 
         def snr(ref, actual):
             error = actual - ref
-            return 20 * torch.log10(ref.norm() / error.norm())
+            return 20 * torch.log10(
+                torch.linalg.vector_norm(ref) / torch.linalg.vector_norm(error)
+            )
 
         assert snr(outputs_ref, outputs_int8mp) > 20
         assert snr(inputs_ref.grad, inputs_int8mp.grad) > 20
--- a/torchao/float8/float8_utils.py
+++ b/torchao/float8/float8_utils.py
@@ -144,8 +144,8 @@ def compute_error(x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
         x: The original tensor.
         y: The tensor to compare to the original tensor.
     """
-    Ps = torch.norm(x)
-    Pn = torch.norm(x - y)
+    Ps = torch.linalg.vector_norm(x)
+    Pn = torch.linalg.vector_norm(x - y)
     return 20 * torch.log10(Ps / Pn)
 
 
--- a/torchao/prototype/parq/quant/lsbq.py
+++ b/torchao/prototype/parq/quant/lsbq.py
@@ -70,7 +70,7 @@ def compute_v_per_channel(p: Tensor, dim: Optional[int] = None, ternary: bool = 
     r = r.sub(v * binary_sign(r))
 
     # compute least squares error, then select the `v` minimizes it
-    costs = r.norm(dim=dim)
+    costs = torch.linalg.vector_norm(r, dim=dim)
     indices = costs.argmin(dim=dim, keepdim=True)
     v_best = v_cands.gather(1, indices)
     return v_best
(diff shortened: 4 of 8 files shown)


jerryzh168:
thanks, according to the linked doc, it looks like there is three functions:

Use [torch.linalg.vector_norm()](https://docs.pytorch.org/docs/stable/generated/torch.linalg.vector_norm.html#torch.linalg.vector_norm) when computing vector norms and [torch.linalg.matrix_norm()](https://docs.pytorch.org/docs/stable/generated/torch.linalg.matrix_norm.html#torch.linalg.matrix_norm) when computing matrix norms. For a function with a similar behavior as this one see [torch.linalg.norm()](https://docs.pytorch.org/docs/stable/generated/torch.linalg.norm.html#torch.linalg.norm).

and the closest one is `torch.linalg.norm()` that probably works for both vector and matrix?

namgyu-youn:
> thanks, according to the linked doc, it looks like there is three functions:
> 
> Use [torch.linalg.vector_norm()](https://docs.pytorch.org/docs/stable/generated/torch.linalg.vector_norm.html#torch.linalg.vector_norm) when computing vector norms and [torch.linalg.matrix_norm()](https://docs.pytorch.org/docs/stable/generated/torch.linalg.matrix_norm.html#torch.linalg.matrix_norm) when computing matrix norms. For a function with a similar behavior as this one see [torch.linalg.norm()](https://docs.pytorch.org/docs/stable/generated/torch.linalg.norm.html#torch.linalg.norm).
> 
> and the closest one is `torch.linalg.norm()` that probably works for both vector and matrix?

@jerryzh168 In my last experience, `torch.linalg.vector_norm()` has better visibility because developers can directly check that it is a vector norm. How about using `torch.linalg.vector_norm()`?

jerryzh168:
> > thanks, according to the linked doc, it looks like there is three functions:
> > Use [torch.linalg.vector_norm()](https://docs.pytorch.org/docs/stable/generated/torch.linalg.vector_norm.html#torch.linalg.vector_norm) when computing vector norms and [torch.linalg.matrix_norm()](https://docs.pytorch.org/docs/stable/generated/torch.linalg.matrix_norm.html#torch.linalg.matrix_norm) when computing matrix norms. For a function with a similar behavior as this one see [torch.linalg.norm()](https://docs.pytorch.org/docs/stable/generated/torch.linalg.norm.html#torch.linalg.norm).
> > and the closest one is `torch.linalg.norm()` that probably works for both vector and matrix?
> 
> @jerryzh168 In my last experience, `torch.linalg.vector_norm()` has better visibility because developers can directly check that it is a vector norm. How about using `torch.linalg.vector_norm()`?

does it work for matrix? or just vector?

namgyu-youn:
> > @jerryzh168 In my last experience, `torch.linalg.vector_norm()` has better visibility because developers can directly check that it is a vector norm. How about using `torch.linalg.vector_norm()`?
> 
> does it work for matrix? or just vector?

@jerryzh168 [torch.linalg.vector_norm](https://docs.pytorch.org/docs/stable/generated/torch.linalg.vector_norm.html#torch.linalg.vector_norm) only works for the vector norm as we discussed. We can use [torch.linalg.norm](https://docs.pytorch.org/docs/stable/generated/torch.linalg.norm.html#torch.linalg.norm) for both (matrix & vector) uses, but in my las experience, `torch.linalg.vector_norm` helped me clarify at https://github.com/NVIDIA/TensorRT-Model-Optimizer/pull/206 and https://github.com/facebookresearch/optimizers/pull/182.

`linalg.norm` might interrupt developers' understanding of it. So I prefer using `linalg.vector_norm` and `linalg.matrix_norm` more.  But I am ready to use `linalg.norm`; let me know which aligns more.

jerryzh168:
> > > @jerryzh168 In my last experience, `torch.linalg.vector_norm()` has better visibility because developers can directly check that it is a vector norm. How about using `torch.linalg.vector_norm()`?
> > 
> > 
> > does it work for matrix? or just vector?
> 
> @jerryzh168 [torch.linalg.vector_norm](https://docs.pytorch.org/docs/stable/generated/torch.linalg.vector_norm.html#torch.linalg.vector_norm) only works for the vector norm as we discussed. We can use [torch.linalg.norm](https://docs.pytorch.org/docs/stable/generated/torch.linalg.norm.html#torch.linalg.norm) for both (matrix & vector) uses, but in my las experience, `torch.linalg.vector_norm` helped me clarify at [NVIDIA/TensorRT-Model-Optimizer#206](https://github.com/NVIDIA/TensorRT-Model-Optimizer/pull/206) and [facebookresearch/optimizers#182](https://github.com/facebookresearch/optimizers/pull/182).
> 
> `linalg.norm` might interrupt developers' understanding of it. So I prefer using `linalg.vector_norm` and `linalg.matrix_norm` more. But I am ready to use `linalg.norm`; let me know which aligns more.

I don't quite get it, if it only works for vector norm, what happens for matrix inputs? or we expect all callsites to only have vector inputs?

jerryzh168:
oh sorry, after reading the doc more closely, I think I'm a mistaken about what "vector" describes, it is referring to output instead of input

jerryzh168 (approved):

namgyu-youn:
@jerryzh168  it seems that [CI failure](https://hud.pytorch.org/pr/pytorch/ao/2660#48628837745) (`AttributeError: module 'torch' has no attribute 'int1'`) occurred due to a PyTorch version error. https://github.com/ant-research/MagicQuill/issues/97 suggested upgrading PyTorch, showing CI failure is not related, but not certain about the true reason.

Therefore, we might face troubleshooting about this; **feel free to close this PR if you think issue is too chained**. This PR addresses deprecated functions in the foreseeable future, but I am fine to close it if it breaks CI.

jerryzh168:
@namgyu-youn I think you can use https://github.com/pytorch/ao/blob/fbe3df96f0e96c36c05d0ce9214cae6b00c8d2ad/test/quantization/quantize_/workflows/int4/test_int4_opaque_tensor.py#L36 to skip the tests that need torch.int1 (added in pytorch 2.6 I think)

namgyu-youn:
> @namgyu-youn I think you can use
> 
> https://github.com/pytorch/ao/blob/fbe3df96f0e96c36c05d0ce9214cae6b00c8d2ad/test/quantization/quantize_/workflows/int4/test_int4_opaque_tensor.py#L36
> 
> to skip the tests that need torch.int1 (added in pytorch 2.6 I think)

Sorry I didn't make it clear. What I wanted to ask was, "should I add SkipTest to pass (Int1 support) CI?". Wondered why this CI broken for this PR, but I haven't found true reason. (maybe its related to other PRs)

jerryzh168:
@namgyu-youn maybe rebase? generally we should already skip the int1 tests for lower torch versions

namgyu-youn:
@jerryzh168 Rebase was right; the reason was an old CI run. Since 2.5.0 check was dropped at https://github.com/pytorch/ao/pull/2720, we don't have to investigate into it; CI fail seems unrelated, I think.

jerryzh168 (approved):

PR #2662, opened 2025-08-01
Update regression_test_rocm.yml

amdfaa:
Migrate mi300s to gfx942


petrex (commented):
lgtm.

petrex (approved):

petrex:
Quick question : Do we have more capacity than before, or is this coming from a larger allocation in the new pool?

huydhn (approved):

huydhn:
Some tests are failing, but I assume they are pre-existing failures.  Let's land this anyway as there is no mi300 runner anymore.

PR #2663, opened 2025-08-01
[MoE training] Assert expert weights are column-major; preserve subclass with transpose

danielvegamyhre:
Once https://github.com/pytorch/torchtitan/pull/1517 lands we need to:
- remove the temporary workaround of dynamic memory layout transformation from row-major to col-major before every grouped gemm, and just assert weights are col-major.
- preserve subclass through transpose ops, so the `weight.transpose(-2, -1)` op doesn't lose the subclass and thus no override to fp8 grouped gemm is done


danielvegamyhre:
cc @vkuzo @drisspg this is ready for review, the associated titan PR has landed https://github.com/pytorch/torchtitan/pull/1517

drisspg (approved):

danielvegamyhre:
confirmed test failures are unrelated, and we are looking into these red CI tests to get them fixed

PR #2666, opened 2025-08-01
Add NVFP4 QAT

andrewor14:
* (to be filled)

**Summary:** This commit adds a QAT flow for NVFP4, following the
numerics in `NVFP4Tensor` closely but without the dtyping casting,
swizzling, and the packing/unpacking. Users can call this flow as follows:

```
from torchao.quantization.qat import NVFP4FakeQuantizeConfig, QATConfig

qat_config = QATConfig(
    activation_config=NVFP4FakeQuantizeConfig(),
    weight_config=NVFP4FakeQuantizeConfig(),
    step="prepare",
)
quantize_(model, qat_config)
```

**Test Plan:**
```
```

### DIFF
--- a/test/quantization/test_qat.py
+++ b/test/quantization/test_qat.py
@@ -12,6 +12,7 @@ import unittest
 import warnings
 from typing import List
 
+import pytest
 import torch
 import torch.nn.functional as F
 from parameterized import parameterized
@@ -44,6 +45,7 @@ from torchao.quantization.qat.embedding import (
 )
 from torchao.quantization.qat.fake_quantize_config import (
     IntxFakeQuantizeConfig,
+    NVFP4FakeQuantizeConfig,
 )
 from torchao.quantization.qat.fake_quantizer import (
     FakeQuantizer,
@@ -112,8 +114,8 @@ class M(torch.nn.Module):
         self.sub = Sub()
         self.linear2 = torch.nn.Linear(256, 512, bias=False).to(torch.float)
 
-    def example_inputs(self):
-        return (torch.randn(1, 512).to(torch.float),)
+    def example_inputs(self, device: torch.device = None):
+        return (torch.randn((1, 512), device=device).to(torch.float),)
 
     def _get_all_weight_qparams(self) -> List[torch.Tensor]:
         return [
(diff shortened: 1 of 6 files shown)


test/quantization/test_qat.py
```diff
@@ -1884,6 +1886,32 @@ def test_qat_api_deprecation(self):
...
+        quantize_(m, qat_config)
+
+        # Compare prepared values
+        torch.manual_seed(self.SEED)
+        x = m.example_inputs("cuda")
+        out = m(*x)
+        baseline_out = baseline_model(*x)
+        sqnr = compute_error(out, baseline_out).item()
+        # Use same SQNR threshold as `test_nvfp4_reconstruction`
+        # TODO: why is this 0.0 when `use_per_tensor_scale=True`?
```
drisspg: That is a seems to be a bug it use_per_tensor_scale should be higher, probably supposed to be 10.0

torchao/quantization/qat/fake_quantize_config.py
```diff
@@ -36,6 +36,22 @@ class FakeQuantizeConfigBase(abc.ABC):
...
+    according to https://developer.nvidia.com/blog/introducing-nvfp4-for-efficient-and-accurate-low-precision-inference/.
+
+    Fake quantization numerics follow `NVFP4Tensor` closely: https://github.com/pytorch/ao/blob/main/torchao/prototype/mx_formats/nvfp4_tensor.py.
+
+    Args:
+        use_per_tensor_scale (bool): Whether to use two-level per-tensor fp32 scaling
+            after the initial fp8 (e4m3) block-wise scaling.
+    """
+
+    use_per_tensor_scale: bool = False
```
drisspg: We should default to true

drisspg:
Any numeric studies on how / if this improves quant error? Even if its pretty trivial setup

test/quantization/test_qat.py
```diff
@@ -1952,6 +1953,45 @@ def test_infer_fp8_int4_config(self):
...
+
+        self._test_quantize_api_against_ptq(
+            NVFP4InferenceConfig(),
+            target_prepare_sqnr=8,
+            target_convert_sqnr=float("inf"),
+        )
+
+    @unittest.skipIf(not _CUDA_IS_AVAILABLE, "skipping when cuda is not available")
+    @parametrize("use_per_tensor_scale", [True, False])
+    def test_qat_nvfp4(self, use_per_tensor_scale: bool):
```
drisspg, line 1972: can we add a test where we compare the outputs of the converted qat model to the ptq only model with the same config and assert they are identicallly equal

torchao/quantization/qat/fake_quantizer.py
```diff
@@ -91,14 +95,60 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        if x.dim() == 3:
+            x = x.view(-1, x.shape[-1])
+        if self.config.use_per_tensor_scale:
+            tensor_amax = torch.max(torch.abs(x))
+            per_tensor_scale = per_tensor_amax_to_scale(tensor_amax)
+        else:
+            per_tensor_scale = None
+
+        # quantize
+        scale, q = _nvfp4_quantize(
```
drisspg: no grad?

torchao/quantization/qat/fake_quantizer.py
```diff
@@ -91,14 +95,60 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+            x,
+            block_size=block_size,
+            per_tensor_scale=per_tensor_scale,
+            skip_dtype_cast_and_packing=True,
+        )
+        if self.config.use_per_tensor_scale:
+            scale = scale * per_tensor_scale
```
drisspg: Just to confirm this is applied both the input activation and the weight right?

andrewor14: yep

torchao/quantization/qat/fake_quantizer.py
andrewor14: added STE instead (see _Float8Round)

torchao/quantization/qat/fake_quantizer.py
```diff
@@ -91,14 +95,60 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        from torchao.prototype.mx_formats.nvfp4_tensor import (
+            _nvfp4_quantize,
+            per_tensor_amax_to_scale,
+        )
+
+        block_size = 16
+        original_shape = x.shape
+        if x.dim() == 3:
+            x = x.view(-1, x.shape[-1])
```
jerryzh168: why does this happen here? can this happen in quant primtive ops?

andrewor14: I found that this was necessary for activations during training (for the batch size). Not sure if it was necessary for inference? @drisspg

test/quantization/test_qat.py
andrewor14: Actually that test already exists (`test_quantize_api_nvfp4` does this, notice `converted_sqnr` is inf)

PR #2676, opened 2025-08-04
Add quantized tensor for groupwise lut based quantization

szyszyzys:
Reviewed By: metascroy

Differential Revision: D79119915

### DIFF
--- a/torchao/prototype/quantization/codebook_groupwise/codebook_quantized_tensor.py
+++ b/torchao/prototype/quantization/codebook_groupwise/codebook_quantized_tensor.py
@@ -1,0 +1,220 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the license found in the
+# LICENSE file in the root directory of this source tree.
+
+from typing import Optional
+
+import torch
+import torch.nn.functional as F
+from torch.utils._python_dispatch import return_and_correct_aliasing
+
+from torchao.utils import TorchAOBaseTensor
+
+# --- C++ Op Accessor Functions ---
+
+
+def get_pack_op(weight_nbit: int):
+    """Gets the C++ packing function from the 'torchao' namespace."""
+    op_name = f"_pack_groupwise_{weight_nbit}bit_weight_with_lut"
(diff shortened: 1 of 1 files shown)


jerryzh168:
does this have to be a new tensor? the codebook_coreml CodebookQuantizedTensor has a block_size that can represent groupwise as well?

szyszyzys:
> does this have to be a new tensor? the codebook_coreml CodebookQuantizedTensor has a block_size that can represent groupwise as well?

Yes, we can indeed reuse the CodebookQuantizedTensor by incorporating the new function call into the C++ file. The primary reason I created a new class was due to the differing definitions of block size between [codebook_coreml](https://github.com/pytorch/ao/blob/ca5f7887c24aaeab21bc4b3519ea9802f754d710/torchao/prototype/quantization/codebook_coreml/codebook_ops.py#L23C1-L24C1) and [codebook](https://github.com/pytorch/ao/blob/ca5f7887c24aaeab21bc4b3519ea9802f754d710/torchao/prototype/quantization/codebook/codebook_ops.py#L153C5-L153C28), which I found to be somewhat confusing. 
This function utilizes elements from both as the backend for codebook generation. To avoid this confusion, I opted to create a new tensor that doesn't use the term "block size," as using "lut_group_size" and "scale_group_size" can be more intuitive and make it easier to understand the differences.
However, your suggestion makes sense, and I agree that we can have a single CodebookQuantizedTensor shared across all the implementations.

jerryzh168:
we could remove the [codebook](https://github.com/pytorch/ao/blob/ca5f7887c24aaeab21bc4b3519ea9802f754d710/torchao/prototype/quantization/codebook/codebook_ops.py#L153C5-L153C28) I think, it's added as a prototype before but no use case yet. codebook_coreml is the one we want to use, and also the block_size that's been defined there is the one we want to use.

jerryzh168:
also `scale_group_size` can probably be generalized to `scale_block_size` as well to be more future proof

szyszyzys:
> also `scale_group_size` can probably be generalized to `scale_block_size` as well to be more future proof

Got it. I'll proceed with these changes.

PR #2680, opened 2025-08-04
Update op_groupwise_lowbit_weight_lut-impl.h

metascroy:
The following headers should not be used in the impl file because it creates an ExecuTorch compile issue:

```
#include <ATen/Functions.h>
#include <torch/library.h>
```

This removes them.


szyszyzys:
Yes we should remove these 2 lines.

szyszyzys (approved):
Looks good for me. We should remove these 2 line.

PR #2686, opened 2025-08-05
[CPU][float8] Add scaled_embedding_bag kernel

shiyang-weng:
Implemented FP8 QEmbeddingBag on CPU, currently supporting:
- `include_last_offset=True`
- `mode="sum"`

Next steps 
1. expand supported modes.
2. Use fp8 instructions instead


test/quantization/test_quant_api.py
```diff
@@ -858,6 +859,57 @@ def test_int4wo_cuda_serialization(self):
             # load state_dict in cuda
             model.load_state_dict(sd, assign=True)
 
+    @unittest.skipIf(
+        "CPU" not in torch._C._dispatch_dump("torchao::qembeddingbag"),
+        reason="cpp kernels not built",
+    )
+    def test_embeddingbag_cpu(self):
```
jerryzh168: the test should be added here I think: https://github.com/pytorch/ao/blob/main/test/test_ops.py

torchao/csrc/cpu/qembeddingbag.cpp
```diff
@@ -0,0 +1,211 @@
...
+  }
+  return _mm512_loadu_ps(buf);
+}
+
+void _mm512_store_ps_cvt_e4m3(at::Float8_e4m3fn* result, const __m512 x, float * buf) {
+  _mm512_storeu_ps(buf, x);
+  for (int i = 0; i < 16; i++) {
+    result[i] = static_cast<at::Float8_e4m3fn>(buf[i]);
+  }
+}
```
Xia-Weiwen: Maybe use `at::vec`?

torchao/csrc/cpu/qembeddingbag.cpp
```diff
@@ -0,0 +1,170 @@
...
+  int64_t emb_dim = qweight.size(1);
+
+  auto index_type = indices.scalar_type();
+  auto qtype = qweight.scalar_type();
+  float w_scale = w_scales.data_ptr<float>()[0];
+
+  TORCH_CHECK(indices.is_contiguous());
+  TORCH_CHECK(offsets.is_contiguous() && offsets.scalar_type() == index_type);
+  TORCH_CHECK(qweight.is_contiguous());
+  TORCH_CHECK(qweight.dim() == 2);
```
Xia-Weiwen: Also check weight dtype?

torchao/ops.py
```diff
@@ -1106,3 +1109,19 @@ def _(
...
+    qweight: Tensor,
+    indices: Tensor,
+    offsets: Tensor,
+    w_scales: Tensor,
+    o_scale: float,
+    mode: int,
+    include_last_offset: bool,
+) -> Tensor:
+    assert include_last_offset == True
+    # if include_last_offset:
```
Xia-Weiwen: Remove?

torchao/csrc/cpu/qembeddingbag.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+at::Tensor qembeddingbag_impl(const at::Tensor &qweight,
+                              const at::Tensor &indices,
+                              const at::Tensor &offsets,
+                              const at::Tensor &w_scales, double o_scale,
+                              const int64_t mode, bool include_last_offset) {
+  // Only support include_last_offset == True and mode ==
+  // at::native::EmbeddingBagMode::SUM
+  TORCH_CHECK(include_last_offset);
+  TORCH_CHECK(mode == at::native::EmbeddingBagMode::SUM);
```
Xia-Weiwen: Better to add error messages and TODO here

LevelDownRefine:
@jerryzh168 Could you help review this pr

torchao/ops.py
```diff
@@ -70,6 +70,9 @@
 lib.define(
     "da8w4_linear_cpu(Tensor input, Tensor input_scales, Tensor input_qzeros, Tensor weight, Tensor weight_scales, Tensor weight_qzeros, Tensor compensation, Tensor? bias, ScalarType output_dtype) -> Tensor"
 )
+lib.define(
+    "qembeddingbag(Tensor qweight, Tensor indices, Tensor offsets, Tensor weight_scale, float o_scale, int mode, bool include_last_offset) -> Tensor"
```
jerryzh168: is this the same as https://github.com/pytorch/pytorch/blob/371eacb2ae4ecdabc52ea4634ed21558df2f3bab/aten/src/ATen/native/native_functions.yaml#L2368C1-L2369C1? with the only difference of qweight being float8?

Xia-Weiwen: @jerryzh168 Thanks for reviewing. Yes, I think so, except that the implementation in this PR has limited functionality so far.

LevelDownRefine: This operator is used for inference. So I did not add any parameters related to the gradient, including scale_grad_by_freq, sparse, per_sample_weights, padding_idx.

jerryzh168: I think we should add this to pytorch directly if that's the case, float8 is a native dtype in pytorch, so I think it makes most of the sense to just add the functionality there, we can error out in the op if some arg combination is not supported or invalid for float8

LevelDownRefine: Intel's platform has fp8 instructions. When we are ready, we hope to update this kernel based on fp8 instructions. As far as I know, the latest GCC is required. Is it difficult to support in PyTorch?

jerryzh168: I'm not sure, can you open an issue for this in pytorch/pytorch?

LevelDownRefine: And for PT API, PT use qtensor and no need to use scale, and we use scale here. Could we land on ao?

jerryzh168: oh I meant can we add to normal embedding op in PT: https://github.com/pytorch/pytorch/blob/f5bf5147ad18994c9a6e0f565d7831362bf5a18a/aten/src/ATen/native/native_functions.yaml#L2360, not quantized one

Xia-Weiwen: So, since this PR adds a quantized version of this op, do you think it better to be added in Torchao rather than in torch core? Thanks.

jerryzh168: yeah my question is can this be implemented with extending the embedding_bag op in pytorch and do the scaling in torchao? or will performance be a concern here

LevelDownRefine: This is a memory bound operator. Repeated reading and writing will lead to significant performance degradation.  For example, if we originally need to read and write once(this situation will also occur many times for DLRM), we will need to read and write twice after do the scaling separately, and the performance will be reduced by half.

jerryzh168: OK, sounds good, maybe rename this to `_scaled_embedding_bag` to follow these ops: https://github.com/pytorch/pytorch/blob/31a41daff49f2cde941d8b9e35cb2eaeeb606c0d/aten/src/ATen/native/native_functions.yaml#L7135

using `_` to indicating it's prototype op since you may want to update the arg list expand hardware coverage etc. later

test/test_ops.py
```diff
@@ -780,5 +780,69 @@ def test_swizzle_mm():
...
+        test_out = torch.ops.torchao._scaled_embedding_bag(
+            fp8_weight,
+            indices,
+            offsets,
+            weight_scale,
+            1.0,
+            mode_enum,
+            include_last_offset,
+        ).to(dtype)
+        torch.testing.assert_close(refe_out, test_out, atol=0, rtol=0)
```
jerryzh168: is this too strict?

LevelDownRefine: changed to 1e-5

torchao/ops.py
LevelDownRefine: Done

jerryzh168:
> @pytorchbot merge

we just manually merge with the button in torchao

jerryzh168:
also is this op built by default? I think ideally it can be optional so it does not impact the normal build. we have seen some errors when some other kernels from prototype feature that breaks the torchao build

LevelDownRefine:
> also is this op built by default? I think ideally it can be optional so it does not impact the normal build. we have seen some errors when some other kernels from prototype feature that breaks the torchao build

Like other kernel on cpu/*.cpp, it is not built by default and built only with USE_CPU_KERNELS=1.

PR #2687, opened 2025-08-05
Align Int4Tensor implementation details with the design of Float8Tensor

jerryzh168:
Stacked PRs:
 * #2738
 * #2710
 * __->__#2687

--- --- ---

### Align Int4Tensor implementation details with the design of Float8Tensor

Summary:
Int4Tensor is the non-preshuffled version of int4 quantized Tensor, data is [N, K/2], scale/zero_point has shape: [K/group_size, N]

Multiple fixes for Int4Tensor to align with the design of Float8Tensor (only calling fbgemm ops)
* defined `tensor_data_names` and `tensor_attribute_names` so we can remove some of the implementations from TorchAOBaseTensor
* Migrated op implementation and tests from https://github.com/pytorch/ao/pull/2387

Note: This is just refactoring Int4Tensor, no BC related changes in this PR

Int4Tensor path is exposed in version 2 of `Int4WeightOnlyConfig` (default version is still 1, which is using the old AQT path

Test Plan:
python test/quantization/quantize_/workflows/int4/test_int4_tensor.py

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/quantization/quantize_/workflows/int4/test_int4_tensor.py
+++ b/test/quantization/quantize_/workflows/int4/test_int4_tensor.py
@@ -8,25 +8,21 @@ import unittest
 
 import torch
 from torch.testing._internal.common_utils import (
-    TestCase,
+    instantiate_parametrized_tests,
+    parametrize,
     run_tests,
 )
 
-from torchao.quantization import (
-    Int4WeightOnlyConfig,
-    quantize_,
-)
+from torchao.quantization import Int4WeightOnlyConfig, quantize_
 from torchao.quantization.utils import compute_error
-from torchao.utils import (
-    TORCH_VERSION_AT_LEAST_2_8,
-    is_sm_at_least_90,
-)
+from torchao.testing.utils import TorchAOIntegrationTestCase
+from torchao.utils import TORCH_VERSION_AT_LEAST_2_8, is_sm_at_least_90
 
 
 @unittest.skipIf(not TORCH_VERSION_AT_LEAST_2_8, "Need pytorch 2.8+")
 @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
 @unittest.skipIf(not is_sm_at_least_90(), "Nedd sm90+")
-class TestInt4Tensor(TestCase):
+class TestInt4Tensor(TorchAOIntegrationTestCase):
     def setUp(self):
         self.config = Int4WeightOnlyConfig(
             group_size=128,
@@ -61,50 +57,46 @@ class TestInt4Tensor(TestCase):
         quantize_(dummy, self.config)
         weight1 = dummy.weight.narrow(0, 0, 64)
         weight2 = dummy.weight.narrow(1, 0, 128)
-        self.assertEqual(weight1._data, dummy.weight._data.narrow(0, 0, 64))
+        self.assertEqual(weight1.qdata, dummy.weight.qdata.narrow(0, 0, 64))
         self.assertEqual(weight1.scale, dummy.weight.scale.narrow(1, 0, 64))
-        self.assertEqual(weight2._data, dummy.weight._data.narrow(1, 0, 64))
+        self.assertEqual(weight1.zero_point, dummy.weight.zero_point.narrow(1, 0, 64))
+        self.assertEqual(weight2.qdata, dummy.weight.qdata.narrow(1, 0, 64))
         self.assertEqual(weight2.scale, dummy.weight.scale.narrow(0, 0, 1))
+        self.assertEqual(weight2.zero_point, dummy.weight.zero_point.narrow(0, 0, 1))
 
         # check for sliced weight, before and after float8 quantization
         # does not differ too much
         input = torch.randn(2, 256, dtype=dtype, device=device)
         res_ref = dummy1(input)
-        dummy.weight = torch.nn.Parameter(weight1, requires_grad=False)
+        dummy.weight = torch.nn.Parameter(weight1.contiguous(), requires_grad=False)
         res = dummy(input)
         assert compute_error(res, res_ref) > 20
 
         input = torch.randn(2, 128, dtype=dtype, device=device)
         res_ref = dummy2(input)
-        dummy.weight = torch.nn.Parameter(weight2, requires_grad=False)
+        dummy.weight = torch.nn.Parameter(weight2.contiguous(), requires_grad=False)
         res = dummy(input)
         assert compute_error(res, res_ref) > 15
 
-    def test_slice_and_copy_(self):
+    def test_slice_preserves_aliasing(self):
+        config = self.config
         l = torch.nn.Linear(1024, 1024).to("cuda").to(torch.bfloat16)
         l.weight = torch.nn.Parameter(
             torch.zeros(1024, 1024, dtype=torch.bfloat16, device="cuda")
         )
-        quantize_(l, self.config)
+        quantize_(l, config)
         param = l.weight
         param_data = param.data
         param_data = param_data.narrow(0, 0, 512)
-        assert param.data._data.data_ptr() == param_data._data.data_ptr()
+        # Making sure the aliasing is preserved in sliced quantized Tensor
+        assert param.data.qdata.data_ptr() == param_data.qdata.data_ptr()
         assert param.data.scale.data_ptr() == param_data.scale.data_ptr()
         assert param.data.zero_point.data_ptr() == param_data.zero_point.data_ptr()
-        orig_value = param.data._data[0][0].item()
 
-        # dummy_l has random input (shouldn't be 0)
-        dummy_l = torch.nn.Linear(1024, 1024).to("cuda").to(torch.bfloat16)
-        quantize_(dummy_l, self.config)
-        quantized = dummy_l.weight
-        quantized = quantized.narrow(0, 0, 512)
+    def test_slice_and_copy_similar_to_vllm(self):
+        self._test_slice_and_copy_similar_to_vllm(self.config)
 
-        param_data.copy_(quantized)
-
-        # making sure param.data is updated
-        assert param.data._data[0][0] != orig_value
-
+    @unittest.skipIf(not is_sm_at_least_90(), "Nedd sm90+")
     def test_bmm(self):
         class M(torch.nn.Module):
             def __init__(self, weight):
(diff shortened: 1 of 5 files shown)


test/quantization/quantize_/workflows/int4/test_int4_tensor.py
```diff
@@ -7,20 +7,79 @@
...
+from torchao.quantization import Int4WeightOnlyConfig, quantize_
 from torchao.quantization.utils import compute_error
-from torchao.utils import (
-    TORCH_VERSION_AT_LEAST_2_8,
-    is_sm_at_least_90,
-)
+from torchao.utils import TORCH_VERSION_AT_LEAST_2_8, is_sm_at_least_90
+
+
+class Experts(nn.Module):
```
vkuzo: can we reuse from `Float8Tensor` test?

test/quantization/quantize_/workflows/int4/test_int4_tensor.py
```diff
@@ -7,20 +7,79 @@
...
+    instantiate_parametrized_tests,
+    parametrize,
     run_tests,
 )
 
-from torchao.quantization import (
-    Int4WeightOnlyConfig,
-    quantize_,
-)
+from torchao.prototype.moe_quant.utils import MoEQuantConfig
```
vkuzo: it's unexpected that this is here, because this is a prototype workflow

test/quantization/quantize_/workflows/int4/test_int4_tensor.py
```diff
@@ -80,31 +139,62 @@ def test_slice(self):
...
         param_data = param_data.narrow(0, 0, 512)
-        assert param.data._data.data_ptr() == param_data._data.data_ptr()
+        # Making sure the aliasing is preserved in sliced quantized Tensor
+        assert param.data.qdata.data_ptr() == param_data.qdata.data_ptr()
         assert param.data.scale.data_ptr() == param_data.scale.data_ptr()
-        assert param.data.zero_point.data_ptr() == param_data.zero_point.data_ptr()
-        orig_value = param.data._data[0][0].item()
 
-        # dummy_l has random input (shouldn't be 0)
+    def test_slice_and_copy_similar_to_vllm(self):
```
vkuzo, line 96: all of these tests seem very similar to `Float8Tensor`, is there an opportunity to unify?

torchao/quantization/quantize_/workflows/int4/int4_tensor.py
```diff
@@ -35,75 +31,38 @@ class Int4Tensor(TorchAOBaseTensor):
...
-            self._data.to(device),
-            self.scale.to(device),
-            self.zero_point.to(device),
-            self.block_size,
-            self.shape,
-        )
-
     @classmethod
-    def from_float(
+    def to_int4(
```
vkuzo: why rename this? `from_float` seems cleaner, maybe `from_high_precision` might be even more clear.  Same feedback for `Float8Tensor`.

torchao/quantization/quantize_/workflows/int4/int4_tensor.py
```diff
@@ -158,9 +117,10 @@ def _(func, types, args, kwargs):
     orig_act_size = input_tensor.size()
     orig_out_features = weight_tensor.shape[-2]
 
+    input_tensor = input_tensor.reshape(-1, input_tensor.shape[-1])
     res = torch.ops.fbgemm.bf16i4bf16_rowwise(
         input_tensor,
-        weight_tensor._data.contiguous(),
+        weight_tensor.qdata.contiguous(),
```
vkuzo: is it expected that the tensors are not contiguous? if not, can we assert for this instead of calling `contiguous`?

torchao/quantization/quantize_/workflows/int4/int4_tensor.py
```diff
@@ -294,13 +214,257 @@ def _(func, types, args, kwargs):
...
+
+@implements(aten.squeeze.dim)
+def _(func, types, args, kwargs):
+    self, dim = args
+
+    # Squeeze qdata
+    qdata = self.qdata.squeeze(dim=dim)
+
+    # For scale and zero_point, we need to squeeze based on the tensor layout
+    # Int4Tensor has scale and zero_point with shape (K/group_size, N) for 2D or (B, N, K/group_size) for 3D
```
vkuzo, line 454: this is confusing, would you mind sharing some context on why this is true? why are the 2d and 3d cases different for the order of K/group_size and N stored on disk?

jerryzh168: sorry some of these comments are not updated, but I found that for both 2D and 3D the order is the same when I tested the code, will do a pass for the comments for this one

test/quantization/quantize_/workflows/int4/test_int4_tensor.py
jerryzh168: sure

test/quantization/quantize_/workflows/int4/test_int4_tensor.py
jerryzh168: OK, I can copy this to testing utils

test/quantization/quantize_/workflows/int4/test_int4_tensor.py
jerryzh168: we could, this is mostly copy pasted from there, maybe wait until there are 3 of them and we can have some utils?

torchao/quantization/quantize_/workflows/int4/int4_tensor.py
jerryzh168: this is to align with Float8Tensor, sure, can rename this to `from_high_precision` for both

torchao/quantization/quantize_/workflows/int4/int4_tensor.py
jerryzh168: the non-contiguous comes from the reshape ops like transpose, view I think, but the kernel will need these to be contiguous, I can try changing these to assert and do the contiguous operation in user side to see if it works

torchao/quantization/quantize_/workflows/int4/int4_tensor.py
jerryzh168: how about `from_float_hp`? this is a bot more precise than `high_precision` I think

test/quantization/quantize_/workflows/int4/test_int4_tensor.py
```diff
@@ -5,22 +5,63 @@
...
+        base_config_handler = _QUANTIZE_CONFIG_HANDLER[type(config.base_config)]
+        dummy_mod = DummyModule(weight)
+        quant_mod = base_config_handler(dummy_mod, config.base_config)
+        return quant_mod.weight
+
+    return _moe_quant_tensor_base(weight, config)
+
+
+@register_quantize_module_handler(TestOnlyMoEQuantConfig)
+def moe_quant_fn(module, config: TestOnlyMoEQuantConfig):
```
vkuzo: this is really confusing, could you share the result of `print(model)` after this function has been applied?

if it's going to print model with parameters wrapped in `Int4Tensor`, can we just wrap the parameters directly without all of these layers of abstraction?

test/quantization/quantize_/workflows/int4/test_int4_tensor.py
vkuzo: I think it makes sense to unify now specifically because this is testing the same thing, `slice_and_copy_similar_to_vllm`, so we know the use case is the same

test/quantization/quantize_/workflows/int4/test_int4_tensor.py
```diff
@@ -126,20 +202,215 @@ def forward(self, x):
...
+            ],
+            dim=1,
+        )
+        ref_scale = linear1.weight.scale
+        ref_zero_point = linear1.weight.zero_point
+        self.assertEqual(cat_qweight2.qdata, ref_data)
+        self.assertEqual(cat_qweight2.scale, ref_scale)
+        self.assertEqual(cat_qweight2.zero_point, ref_zero_point)
+
+    def test_moe_weight_reshape_ops(self):
```
vkuzo, line 213: if this is the same use case with Float8Tensor, IMO unify

torchao/quantization/quantize_/workflows/int4/int4_tensor.py
```diff
@@ -35,75 +31,38 @@ class Int4Tensor(TorchAOBaseTensor):
...
-            self._data.to(device),
-            self.scale.to(device),
-            self.zero_point.to(device),
-            self.block_size,
-            self.shape,
-        )
-
     @classmethod
-    def from_float(
+    def from_float_hp(
```
vkuzo: you  can do `from_hp` if you want as well.  `from_float_hp` sounds unusual.

torchao/quantization/quantize_/workflows/int4/int4_tensor.py
vkuzo: I would have expected the weights to be stored in a format aligned with what the kernel needs, without any need for just-in-time layout transforms.  Does this match how the current code works?

torchao/testing/model_architectures.py
```diff
@@ -177,3 +178,63 @@ def create_model_and_input_data(
     else:
         raise ValueError(f"Unknown model type: {model_type}")
     return model, input_data
+
+
+class Experts(nn.Module):
```
vkuzo: maybe call it something like `FeedForwardWithExperts`? `Experts` is ambiguous

test/quantization/quantize_/workflows/int4/test_int4_tensor.py
vkuzo: if this is working around the fact that `quantize_` needs to work on modules, IMO we should change quantize_ to handle this instead of working around? seems important for MoEs.

test/quantization/quantize_/workflows/int4/test_int4_tensor.py
jerryzh168: OK

torchao/quantization/quantize_/workflows/int4/int4_tensor.py
jerryzh168: normally it is, but the weights also goes through some transformations like the ones we listed in `test_moe_weight_reshape_ops` which makes weight / scale etc. non-contiguous I think, but I can try to do call contiguous in user code, that might be cleaner I think

test/quantization/quantize_/workflows/int4/test_int4_tensor.py
jerryzh168: yeah the parameters are wrapped in Int4Tensor, this is just applying quantization to each of the moe weights: w1, w2 and w3

I can inline these for now. can follow up with how to have an API for weights + configs separately

test/quantization/quantize_/workflows/int4/test_int4_tensor.py
jerryzh168: yeah it's the same, sure

torchao/testing/model_architectures.py
jerryzh168: this is adapted from https://github.com/meta-llama/llama-models/blob/a9c89c471f793423afd4cc3ca8671d6e56fe64cb/models/llama4/moe.py#L22, how about renaming to `LLama4Experts` to make it more specific

test/quantization/quantize_/workflows/int4/test_int4_tensor.py
jerryzh168: most of the tests in these two files are the same actually, should we just have a common TestCase and both of them inherit from it?

torchao/quantization/quantize_/workflows/int4/int4_tensor.py
jerryzh168: turns out the contiguous is not implemented properly, just fixed that and we can remove contiguous calls in linear/bmm now

test/quantization/quantize_/workflows/int4/test_int4_tensor.py
jerryzh168: probably not worth changing API right now since MoE quant is also moving, let me know if current code looks good

torchao/quantization/quantize_/workflows/int4/int4_tensor.py
```diff
@@ -35,75 +31,38 @@ class Int4Tensor(TorchAOBaseTensor):
     int4 quantization with plain (default) packing format (for all granularities)
 
     Tensor Attributes:
-        _data: packed int4 weight, either 2D (N, K/2) or 3D (B, N, K/2), last dimension is packed
-        scale: (K/group_size, N) for 2D Tensor, (B, N, K/group_size) for 3D Tensor, where B is batch size,
+        qdata: packed int4 weight, either 2D (N, K/2) or 3D (B, N, K/2), last dimension is packed
```
drisspg, line 34: is this the new naming convention ?

torchao/quantization/quantize_/workflows/int4/int4_tensor.py
```diff
@@ -35,75 +31,38 @@ class Int4Tensor(TorchAOBaseTensor):
...
 
     Non-Tensor Attributes:
         block_size: the block size for quantization, representing the granularity, for example groupwise quantization will have block_size (1, group_size)
         shape: the shape of the original Tensor
     """
 
-    tensor_data_attrs = ["_data", "scale", "zero_point"]
-    tensor_attributes = ["block_size", "shape"]
+    tensor_data_names = ["qdata", "scale", "zero_point"]
+    tensor_attribute_names = ["block_size", "shape"]
```
drisspg, line 46: tensor_attribute_names  is a lil weird to me
are these attributes that are `tensors` and thus should go int he right unflatten location?

torchao/quantization/quantize_/workflows/int4/int4_tensor.py
```diff
@@ -294,13 +224,262 @@ def _(func, types, args, kwargs):
...
+            assert torch.equal(tensor_0.zero_point, tensors[i].zero_point)
+        cat_scale = scales[0]
+        cat_zero_point = zero_points[0]
+
+    # Calculate new shape based on the concatenated qdata shape
+    new_shape = list(cat_qdata.shape)
+    new_shape[-1] *= 2
+    new_shape = list(new_shape)
+
+    new = tensor_0.__class__(
```
drisspg: can we just use the Int4Tensor constructor or does this change need to be grabbed at runtime

test/quantization/quantize_/workflows/int4/test_int4_tensor.py
```diff
@@ -8,25 +8,21 @@
...
-)
+from torchao.testing.utils import TorchAOIntegrationTestCase
+from torchao.utils import TORCH_VERSION_AT_LEAST_2_8, is_sm_at_least_90
 
 
 @unittest.skipIf(not TORCH_VERSION_AT_LEAST_2_8, "Need pytorch 2.8+")
 @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
 @unittest.skipIf(not is_sm_at_least_90(), "Nedd sm90+")
-class TestInt4Tensor(TestCase):
+class TestInt4Tensor(TorchAOIntegrationTestCase):
```
vkuzo, line 25: are we keeping both old and new? If we are keeping the old version working, I would expect this test case to not have any changes, as it would test the old version.

torchao/quantization/quantize_/workflows/int4/int4_tensor.py
jerryzh168: yeah we will use `qdata` for raw quantized data now for all tensors, `data` doesn't work because it's already used by normal torch.Tensor, and also I prefer not to use `_data` and make all attribute start with `_`

torchao/quantization/quantize_/workflows/int4/int4_tensor.py
jerryzh168: yeah.. this somewhat follows what unflatten and flatten functions names these things, the `tensor` means the tensor subclass instance, meaning the attributes of the tensor subclass instance, instead of "tensor attributes"

I could remove `tensor_` as well to make it less confusing? probably better to do in a separate PR

torchao/quantization/quantize_/workflows/int4/int4_tensor.py
jerryzh168: I can update to calling Int4Tensor directly as well, I guess this is useful in case the Tensor is inherited though, but not a concern right now

test/quantization/quantize_/workflows/int4/test_int4_tensor.py
jerryzh168: old version meaning the version using AQT? we are keeping AQT, but this test does not test the AQT path, it only tests the new Int4Tensor, and we are updating Int4Tensor in this PR

vkuzo: can we update the PR summary with context on this? Migrations are always confusing and clearly laying out what is changing with BC, what is breaking BC, and what is not changing will help get a good review.

jerryzh168: OK added, no BC related changes in this PR

vkuzo (approved):

PR #2689, opened 2025-08-05
[CPU][FP8] Support FP8 SDPA for CPU backend

Valentine233:
Description:
1. Reuse the schema of `qscaled_dot_product`, and extend the FP8 dtype.
2. Support the fused attention and fallback math kernels for FP8 SDPA.
3. Support the pattern match for FP8 SDPA.

### DIFF
--- a/test/prototype/inductor/test_qsdpa_fusion.py
+++ b/test/prototype/inductor/test_qsdpa_fusion.py
@@ -151,14 +151,14 @@ class TestSDPAPatternRewriterTemplate(TestCase):
     @skipIfRocm
     @unittest.skipIf(
         not torch_version_at_least("2.7.0"),
-        reason="int8 sdpa requires torch 2.7 or later",
+        reason="qsdpa requires torch 2.7 or later",
     )
     @unittest.skipIf(
         "CPU" not in torch._C._dispatch_dump("torchao::qscaled_dot_product"),
         reason="cpp kernels not built",
     )
     @config.patch({"freezing": True})
-    def _test_sdpa_int8_rewriter(self):
+    def _test_qsdpa_rewriter(self):
         import torchao.quantization.pt2e.quantizer.x86_inductor_quantizer as xiq
         from torchao.quantization.pt2e.quantize_pt2e import convert_pt2e, prepare_pt2e
         from torchao.quantization.pt2e.quantizer.x86_inductor_quantizer import (
@@ -213,9 +213,7 @@ if HAS_CPU:
 
     class SDPAPatternRewriterCpuTests(TestSDPAPatternRewriterTemplate):
         device = "cpu"
-        test_sdpa_int8_rewriter_cpu = (
-            TestSDPAPatternRewriterTemplate._test_sdpa_int8_rewriter
-        )
+        test_qsdpa_rewriter_cpu = TestSDPAPatternRewriterTemplate._test_qsdpa_rewriter
 
 
 if __name__ == "__main__":
--- a/torchao/csrc/cpu/aten_kernels/quantized_sdpa.cpp
+++ b/torchao/csrc/cpu/aten_kernels/quantized_sdpa.cpp
@@ -2509,7 +2509,6 @@ at::Tensor _qscaled_dot_product_cpu(
 #ifdef CPU_CAPABILITY_AVX512
       if (at::native::cpublas::could_pack(dtype)) {
           at::Tensor output = at::empty_like(query, query.options()).transpose(1, 2);
-          std::cout << "int8_sdpa_fused_kernel" << std::endl;
           int8_sdpa_fused_kernel(output, query, key, value,
               dropout_p, is_causal, attn_mask, scale,
               q_scale, q_zp,
@@ -2520,7 +2519,6 @@ at::Tensor _qscaled_dot_product_cpu(
           return output.transpose(1, 2);
       } else {
 #endif // CPU_CAPABILITY_AVX512
-          std::cout << "int8_sdpa_math_kernel" << std::endl;
           return int8_sdpa_math_kernel(query, key, value,
               dropout_p, is_causal, attn_mask, scale,
               q_scale, q_zp,
@@ -2536,7 +2534,6 @@ at::Tensor _qscaled_dot_product_cpu(
 // CPUBLAS_BRGEMM_F8F8F32 is defined if FP8 BRGEMM is supported in PyTorch CPUBlas.
       if (at::native::cpublas::could_pack(dtype)) {
           at::Tensor output = at::empty_like(query, query.options()).transpose(1, 2);
-          std::cout << "fp8_sdpa_fused_kernel" << std::endl;
           fp8_sdpa_fused_kernel(output, query, key, value,
               dropout_p, is_causal, attn_mask, scale,
               q_scale, k_scale,
@@ -2545,7 +2542,6 @@ at::Tensor _qscaled_dot_product_cpu(
           return output.transpose(1, 2);
       } else {
 #endif // CPU_CAPABILITY_AVX512 && CPUBLAS_BRGEMM_F8F8F32
-          std::cout << "fp8_sdpa_math_kernel" << std::endl;
           return fp8_sdpa_math_kernel(query, key, value,
               dropout_p, is_causal, attn_mask, scale,
               q_scale, k_scale,


Valentine233:
@CaoE @jianan-gu Please help review, thanks~

torchao/csrc/cpu/quantized_sdpa.cpp
```diff
@@ -1622,10 +1775,371 @@ sdpa_int8_fused_kernel_impl(
   at::native::cpublas::brgemm_release();
 }
 
+// FP8 - kernel with f8f8f8 GEMM
+template <typename scalar_t, typename mask_t,
+          int64_t q_split_size, int64_t kv_split_size>
+inline typename std::enable_if_t<std::is_same_v<scalar_t, at::Float8_e4m3fn>, void>
+fp8_sdpa_fused_kernel_impl(
```
CaoE: Is it possible to merge int8 and fp8 implementations ?

test/prototype/inductor/test_int8_sdpa_fusion.py
```diff
@@ -157,7 +157,7 @@ def _check_common(
     )
     @config.patch({"freezing": True})
     def _test_sdpa_int8_rewriter(self):
-        from torch.export import export_for_training
+        from torch.export import export
```
CaoE: If this test covers fp8, we'd better rename it.

CaoE: Also the file name.

torchao/csrc/cpu/quantized_sdpa.cpp
```diff
@@ -725,13 +769,122 @@ inline void copy_value_with_pad(
...
+  if (i < size) {
+    auto tmp0 = at::vec::Vectorized<float>::loadu(a + i, size - i);
+    auto tmp1 = tmp0 - vec_max;
+    auto tmp2 = tmp1.exp_u20();
+    vec_tmp_sum = at::vec::Vectorized<float>::set(vec_tmp_sum, vec_tmp_sum + tmp2, size - i);
+    auto tmp3 = tmp2 * vec_scale;
+    auto tmp4 = at::vec::clamp(tmp3, vec_min_val, vec_max_val);
+    _store(out + i, tmp4, size - i);
+  }
+  val = vec_tmp_sum.reduce_add();
```
CaoE: Does this function need NaN guard ?

torchao/csrc/cpu/quantized_sdpa.cpp
```diff
@@ -1873,30 +2500,63 @@ at::Tensor _qscaled_dot_product_cpu(
...
+              dropout_p, is_causal, attn_mask, scale,
+              q_scale, q_zp,
+              k_scale, k_zp,
+              v_scale, v_zp,
+              a_scale, a_zp,
+              o_scale, o_zp);
+          return output.transpose(1, 2);
+      } else {
+#endif // CPU_CAPABILITY_AVX512
+          std::cout << "int8_sdpa_math_kernel" << std::endl;
```
CaoE: Seems to be an omission. Remove this.

torchao/csrc/cpu/quantized_sdpa.cpp
```diff
@@ -1873,30 +2500,63 @@ at::Tensor _qscaled_dot_product_cpu(
...
-            o_scale, o_zp);
-        return output.transpose(1, 2);
-    } else {
-  #endif // CPU_CAPABILITY_AVX512
-        return sdpa_int8_math_kernel(query, key, value,
+  if (dtype == at::ScalarType::Byte) {
+#ifdef CPU_CAPABILITY_AVX512
+      if (at::native::cpublas::could_pack(dtype)) {
+          at::Tensor output = at::empty_like(query, query.options()).transpose(1, 2);
+          std::cout << "int8_sdpa_fused_kernel" << std::endl;
```
CaoE: Seems to be an omission.

torchao/csrc/cpu/quantized_sdpa.cpp
```diff
@@ -1873,30 +2500,63 @@ at::Tensor _qscaled_dot_product_cpu(
...
+          std::cout << "fp8_sdpa_fused_kernel" << std::endl;
+          fp8_sdpa_fused_kernel(output, query, key, value,
+              dropout_p, is_causal, attn_mask, scale,
+              q_scale, k_scale,
+              v_scale, a_scale,
+              o_scale);
+          return output.transpose(1, 2);
+      } else {
+#endif // CPU_CAPABILITY_AVX512 && CPUBLAS_BRGEMM_F8F8F32
+          std::cout << "fp8_sdpa_math_kernel" << std::endl;
```
CaoE: Same as above.

torchao/csrc/cpu/quantized_sdpa.cpp
```diff
@@ -1873,30 +2500,63 @@ at::Tensor _qscaled_dot_product_cpu(
...
-            v_scale, v_zp,
-            a_scale, a_zp,
-            o_scale, o_zp);
-        return output.transpose(1, 2);
-    } else {
-  #endif // CPU_CAPABILITY_AVX512
-        return sdpa_int8_math_kernel(query, key, value,
+  if (dtype == at::ScalarType::Byte) {
+#ifdef CPU_CAPABILITY_AVX512
+      if (at::native::cpublas::could_pack(dtype)) {
```
CaoE: Do we always need pack on supported platforms ? Is there any cases where do packing is slower than plain format ?

torchao/csrc/cpu/quantized_sdpa.cpp
```diff
@@ -1873,30 +2500,63 @@ at::Tensor _qscaled_dot_product_cpu(
...
-  #endif // CPU_CAPABILITY_AVX512
+#ifdef CPU_CAPABILITY_AVX512
+      }
+#endif // CPU_CAPABILITY_AVX512
+  } else if (dtype == at::ScalarType::Float8_e4m3fn) {
+#if defined(CPUBLAS_BRGEMM_F8F8F32) && defined(CPU_CAPABILITY_AVX512)
+// CPUBLAS_BRGEMM_F8F8F32 is defined if FP8 BRGEMM is supported in PyTorch CPUBlas.
+      if (at::native::cpublas::could_pack(dtype)) {
+          at::Tensor output = at::empty_like(query, query.options()).transpose(1, 2);
+          std::cout << "fp8_sdpa_fused_kernel" << std::endl;
```
CaoE: Same as above.

torchao/csrc/cpu/quantized_sdpa.cpp
```diff
@@ -1834,6 +2424,43 @@ at::Tensor sdpa_int8_math_kernel(
   return output;
 }
 
+at::Tensor fp8_sdpa_math_kernel(
```
CaoE: Is there any tests for the ref implementation ?

torchao/csrc/cpu/quantized_sdpa.cpp
Valentine233: We can not merge these two implementations, as fp8 uses flash attention while int8 does not.

torchao/csrc/cpu/quantized_sdpa.cpp
Valentine233: We would not get an extreme large value here, because this is part of the safe-softmax where the max value has been subtracted before.

torchao/csrc/cpu/quantized_sdpa.cpp
Valentine233: Thanks for your suggestion!
For the cases we care about, for example landing zone models, we have confirmed that packing is better.
For the general cases, we can tune for need_pack in the future.

test/prototype/inductor/test_int8_sdpa_fusion.py
Valentine233: Thanks and modified.

torchao/csrc/cpu/quantized_sdpa.cpp
Valentine233: Thanks and modified.

torchao/csrc/cpu/quantized_sdpa.cpp
Valentine233: Thanks and modified.

torchao/csrc/cpu/quantized_sdpa.cpp
Valentine233: Thanks and modified.

torchao/csrc/cpu/quantized_sdpa.cpp
Valentine233: Thanks and modified.

torchao/csrc/cpu/quantized_sdpa.cpp
Valentine233: The ref path can be tested when `CPUBLAS_BRGEMM_F8F8F32` is not set and this MACRO is defined in PyTorch. So, the fused attention or math ref one, which kernel to go, depends on the version of PyTorch. I have locally validated the both cases.

Valentine233:
@drisspg @jerryzh168 Could you kindly help review? Thanks~

jerryzh168 (approved):
stamping since this is a pretty independent change

PR #2694, opened 2025-08-05
Allow no quantization during QATConfig convert

andrewor14:
**Summary:** This commit adds back the functionality to swap `FakeQuantized*` modules back to the corresponding `torch.nn.*` without performing post-training quantization:
```
QATConfig(base_config=None, step="convert")
```

This has the exact same functionality as this deprecated config:
```
FromIntXQuantizationAwareTrainingConfig()
```

This functionality is added back since it may be useful to users who wish to save QAT trained checkpoints from models containing only `torch.nn.*` modules (not `FakeQuanitzed*` modules), e.g. when training and inference need to happen on different machines:
```
quantize_(model, QATConfig(base_ptq_config, step="prepare"))
train(model)
quantize_(model, QATConfig(step="convert"))
torch.save(model.state_dict(), "my_checkpoint.pt")

# On a different machine
model.load_state_dict(torch.load("my_checkpoint.pt"))
quantize_(model, base_ptq_config)
```

**Test Plan:**
```
python test/quantization/test_qat.py  -k qat_config_init
python test/quantization/test_qat.py  -k qat_api_convert_no_quantization
```

### DIFF
--- a/torchao/quantization/qat/api.py
+++ b/torchao/quantization/qat/api.py
@@ -115,8 +115,10 @@ class QATConfig(AOBaseConfig):
         ValueError: If `base_config` and `activation_config` are both specified
         ValueError: If `base_config` and `weight_config` are both specified
         ValueError: If neither `base_config` nor `weight_config` is specified
+             and `step` is "prepare"
+        ValueError: If either `activation_config` or `weight_config` is specified
+             and `step` is "convert"
         ValueError: If `step` is not one of "prepare" or "convert"
-        ValueError: If `base_config` is None but `step` is "convert"
         ValueError: If the config is applied on a module that is not a
             `torch.nn.Linear` or `torch.nn.Embedding`, or it is applied on
             `torch.nn.Embedding` with an activation config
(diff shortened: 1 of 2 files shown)


jerryzh168:
I see, this is the step="load" in AWQConfig, would it be easier to just add that

andrewor14:
I don't think we need a different step? This is just separating the convert step into two phases: (1) swap back to nn.Linear, and (2) quantize the model, and these can happen on different machines

torchao/quantization/qat/api.py
```diff
@@ -148,18 +150,26 @@ def __post_init__(self):
...
                 "Cannot specify both `base_config` and `activation_config`"
             )
         if self.base_config is not None and self.weight_config is not None:
             raise ValueError("Cannot specify both `base_config` and `weight_config`")
-        if self.base_config is None and self.step == "convert":
-            raise ValueError("`base_config` must be specified in the convert step")
+        if (
+            self.step == QATStep.PREPARE
+            and self.base_config is None
+            and self.weight_config is None
```
lisjin, line 162: @andrewor14 Would it be possible to allow base_config=None and weight_config=None? This would support PARQ's activation-only quantization use case in #2743.

andrewor14: yeah let's do that separately in your PR

PR #2695, opened 2025-08-05
[pt2e] Avoid getting model device once per node

andrewor14:
**Summary:** Previously, we call `assert_and_get_unqiue_device` once per node in both prepare and convert. This is expensive and unnecessary since the model device is the same across all nodes, so we should just call this once in the beginning and reuse the same model device across all the nodes.

torchao version of https://github.com/pytorch/pytorch/pull/159901

Note: The prepare path is not completely done yet, since we are blocked on the pytorch PR on being merged. It's different from convert since it still calls utility functions from `torch.ao.quantization.fx`.

**Test Plan:**
```
python test/quantization/pt2e/test_quantize_pt2e.py
```

### DIFF
--- a/torchao/quantization/pt2e/observer.py
+++ b/torchao/quantization/pt2e/observer.py
@@ -1908,10 +1908,18 @@ class AffineQuantizedObserverBase(ABC, torch.nn.Module):
             else:
                 scale, zero_point = self.calculate_qparams()
                 scale_node = create_getattr_from_value(
-                    model, model.graph, "_scale", scale
+                    model,
+                    model.graph,
+                    "_scale",
+                    scale,
+                    scale.device if isinstance(scale, torch.Tensor) else None,
                 )
                 zero_point_node = create_getattr_from_value(
-                    model, model.graph, "_zero_point", zero_point
+                    model,
+                    model.graph,
+                    "_zero_point",
+                    zero_point,
+                    zero_point.device if isinstance(zero_point, torch.Tensor) else None,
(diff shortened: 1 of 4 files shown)


jerryzh168 (approved):

torchao/quantization/pt2e/observer.py
```diff
@@ -1885,7 +1885,9 @@ def convert(self, model: torch.fx.GraphModule, observer_node: Node):
             )
 
         from torchao.quantization.pt2e.utils import create_getattr_from_value
+        from torchao.utils import _assert_and_get_unique_device
```
jerryzh168: btw, this is still going to be per node, but here it's probably free to get the device from scale tensor? can we do that

andrewor14: oh you're right, let me try

torchao/quantization/pt2e/observer.py
```diff
@@ -1908,10 +1908,18 @@ def convert(self, model: torch.fx.GraphModule, observer_node: Node):
             else:
                 scale, zero_point = self.calculate_qparams()
                 scale_node = create_getattr_from_value(
-                    model, model.graph, "_scale", scale
+                    model,
+                    model.graph,
+                    "_scale",
+                    scale,
+                    scale.device,
```
jerryzh168: seems like scale could be non-Tensor as well, so here we'd need a device from `model` instead, same for `zero_point`

PR #2671, opened 2025-08-04
[moe training] add benchmark script for moe layer

danielvegamyhre:
Stacked PRs:
 * #2698
 * #2697
 * #2696
 * __->__#2671

--- --- ---

[moe training] add benchmark script for moe layer

## Summary
- Add benchmarking script (with profiling) for a standalone MoE layer. This isolation makes perf analysis + optimization easier.

PR #2696, opened 2025-08-05
[moe training] add fp8 rowwise kernels for expert weights

danielvegamyhre:
Stacked PRs:
 * #2698
 * #2697
 * __->__#2696
 * #2671

--- --- ---

[moe training] add fp8 rowwise kernels for expert weights

## Summary
- torch.compile is too slow for quantizing expert weights along dim1 for backward (see https://github.com/pytorch/pytorch/issues/159769)
- I wrote a triton kernel to do this with better perf. We can remove it once torch.compile is better. 
- This PR is focused on numerical accuracy. Perf benchmarking is in next PR in stack. 

## Test plan
- `pytest test/prototype/moe_training/test_kernels.py`

### DIFF
--- a/torchao/prototype/moe_training/kernels/float8_rowwise.py
+++ b/torchao/prototype/moe_training/kernels/float8_rowwise.py
@@ -1,0 +1,251 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+
+
+from typing import Tuple
+
+import torch
+import triton
+import triton.language as tl
+
+EPS = 1e-12
+
+FP8_DTYPE_MAP = {
+    torch.int8: tl.int8,
+    torch.int16: tl.int16,
+    torch.int32: tl.int32,
+    torch.int64: tl.int64,
+    torch.float8_e4m3fn: tl.float8e4nv,
+    torch.float8_e5m2: tl.float8e5,
+    torch.float16: tl.float16,
+    torch.bfloat16: tl.bfloat16,
+    torch.float32: tl.float32,
+    torch.float64: tl.float64,
+}
+
+block_sizes = [16]
+num_warps = [4]
+num_stages = [2]
+kernel_configs_2D = [
+    triton.Config(
+        {"BLOCK_SIZE_N": block_size, "BLOCK_SIZE_K": block_size * 2},
+        num_warps=warps,
+        num_stages=stages,
+    )
+    for block_size in block_sizes
+    for warps in num_warps
+    for stages in num_stages
+]
+
+from torch.library import triton_op, wrap_triton
+
+
+@triton_op("torchao::triton_fp8_rowwise_transpose_rhs", mutates_args={})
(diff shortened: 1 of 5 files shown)


danielvegamyhre:
@zou3519 here is a different problem occurring with `wrap_triton` in this PR. I added a new triton kernel and wrapped, but when I try to use `torch.compile` now I get an error about using a FunctionalTensor...

### Repro
1. Check out this PR
2. Run unit test using compile: `pytest test/prototype/moe_training/test_training.py  -k test_moe_float8_training[True-target_fqns0]` 

### Error
```
>       raise RuntimeError(
            "Attempting to use FunctionalTensor on its own. Instead, please use it with a corresponding FunctionalTensorMode()"
        )
E       torch._dynamo.exc.BackendCompilerFailed: backend='inductor' raised:
E       RuntimeError: Attempting to use FunctionalTensor on its own. Instead, please use it with a corresponding FunctionalTensorMode()
E       
E       While executing %_grouped_mm : [num_users=1] = call_function[target=torch._grouped_mm](args = (%bfloat16, %transpose), kwargs = {offs: %offsets})
E       Original traceback:
E         File "/home/danvm/torchtitan/torchtitan/experiments/llama4/model/moe.py", line 109, in _run_experts_grouped_mm
E           torch._grouped_mm(
E       
E       Use tlparse to see full graph. (https://github.com/pytorch/tlparse?tab=readme-ov-file#tlparse-parse-structured-pt2-logs)
E       
E       Set TORCHDYNAMO_VERBOSE=1 for the internal stack trace (please do this especially if you're reporting a bug to PyTorch). For even more developer context, set TORCH_LOGS="+dynamo"

```

torchao/prototype/moe_training/kernels/float8_rowwise.py
```diff
@@ -0,0 +1,251 @@
...
+
+    # compute scales with local amax, using axis=0 because for each expert,
+    # we are reading the non-transposed input, and want to compute the scales
+    # along axis=1 for the transposed input.
+    amaxes = tl.max(tl.abs(input_data), axis=1).to(tl.float64)  # (K,)
+    scales = (fp8_dtype_max / tl.clamp(amaxes, min=EPS, max=float("inf"))).to(
+        tl.float32
+    )
+    if round_scales_to_power_of_2:
+        scales = tl.exp2(tl.floor(tl.log2(scales)))
```
vkuzo, line 173: this seems expensive, can we just extract the bits?

torchao/prototype/moe_training/utils.py
```diff
@@ -145,6 +145,41 @@ def _to_2d_jagged_float8_tensor_rowwise(
...
+        input_hp_t,
+        target_dtype,
+        scaling_granularity=ScalingGranularity.AXISWISE,
+        axiswise_dim=-2,
+        round_scales_to_power_of_2=round_scales_to_power_of_2,
+    )  # (E, 1, K)
+
+    # Apply scales to tensor and convert to float8.
+    tensor_scaled = input_hp_t.to(torch.float32) * scales
+    float8_tensor = to_fp8_saturated(tensor_scaled, target_dtype)
```
vkuzo, line 176: this is confusing because it sounds like `Float8TrainingTensor`, maybe name it `float8_data`?

PR #2699, opened 2025-08-05
[not4land] H100 CI test

jerryzh168:
Summary:

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:


jerryzh168:
confirmed that changing these conditions will change the results for kernel generation, but adding `torch.compiler.reset` resolves the issue

PR #2702, opened 2025-08-06
Add __init__.py for group wise lut quantization package

szyszyzys:
Reviewed By: metascroy

Differential Revision: D79119958

### DIFF
--- a/torchao/prototype/quantization/codebook_groupwise/__init__.py
+++ b/torchao/prototype/quantization/codebook_groupwise/__init__.py
@@ -1,0 +1,4 @@
+from .api import GroupwiseLutWeightConfig
+from .codebook_quantized_tensor import GroupwiseLutQuantizedTensor
+
+__all__ = ["GroupwiseLutQuantizedTensor", "GroupwiseLutWeightConfig"]


jerryzh168:
still planning to merge this with coreml_codebook_quant tensor?

metascroy:
@szyszyzys your code needs to go through the ruff linter

szyszyzys:
> @szyszyzys your code needs to go through the ruff linter

I initially used `ruff format` to check the formatting while ignoring import formatting. Now, I'm using `ruff check --select I --fix` to address the issue.

szyszyzys:
> still planning to merge this with coreml_codebook_quant tensor?

Yes, I plan to merge it in a future commit. I'm currently considering how to seamlessly integrate the separate LUT and scale group sizes (or LUT and scale block sizes) into the coreml_codebook_quant tensor. I've now ensured that all the code uses a consistent definition of block size, which will make it easier to adapt the coreml_codebook_quant.

Changing the `coreml_codebook_quant` tensor to include `LUT block size` and `scale block size` should make the code functional. Although CoreML currently doesn't support generating scales, it would be straightforward to use the block size as the LUT block size. However, testing whether the scales improve performance could be challenging. 
Additionally, using a single block size might not be very extensible, as there could be more advanced ways to handle grouping in the future. Do you have any ideas on how to approach this?

PR #2705, opened 2025-08-06
fix swizzle kernel w/ fullgraph

drisspg:
Stacked PRs:
 * __->__#2705

--- --- ---

fix swizzle kernel w/ fullgraph


eellison (approved):
Should we flip switch on using to_blocked ?

PR #2706, opened 2025-08-06
Track API usage

andrewor14:


torchao/float8/float8_linear_utils.py
```diff
@@ -101,6 +102,7 @@ def convert_to_float8_training(
     Returns:
      nn.Module: The modified module with swapped linear layers.
     """
+    torch._C._log_api_usage_once("torchao.quantization.convert_to_float8_training")
```
andrewor14: @vkuzo I put this under the `torchao.quantization` namespace based on some past discussions on code reorg that hasn't happened yet. Do you prefer this namespace compared to `torchao.float8`, which is the current location?

andrewor14: also any other float8 APIs you'd like to track?

jerryzh168 (approved):
quantize_ and pt2e looks good, these will only track internal usage right?

andrewor14:
> quantize_ and pt2e looks good, these will only track internal usage right?

yeah I think so

torchao/float8/float8_linear_utils.py
danielvegamyhre: is the reason for preemptively logging here ahead of the eventual migration so that querying usage metrics is easier (i.e., no need to aggregate results of 2 queries)?

andrewor14: yeah and also we can start tracking earlier, not sure when the migration will happen

vkuzo: IMO better to match existing code structure, and if a refactor happens we'll deal with the logging change.  I say this because there is no timeline or owner for actually migrating `torchao.float8` at the moment.

PR #2710, opened 2025-08-07
Support `optional_tensor_names` in TorchAOBaseTensor

jerryzh168:
Support `optional_tensor_names` in TorchAOBaseTensor

Summary:
Allows subclasses inheriting from TorchAOBaseTensor to have optional tensor attributes, updated
all common util functions to support `optional_tensor_names` list, including `__tensor_flatten__`, `__tensor_unflatten__`, ops like aten._to_copy, contiguous, alias etc.

Test Plan:
python test/test_utils.py

Reviewers:

Subscribers:

Tasks:

Tags:


drisspg (commented):
can you update the doc block for AoBaseTensor I think having class types so that users know what types it contains e.g. you can have optional_tensor_data_names: Optional[List[Tensor]]

so w/ a description of why you want to populate this field, what functions it can help you implement, etc...

jerryzh168:
> can you update the doc block for AoBaseTensor I think having class types so that users know what types it contains e.g. you can have optional_tensor_data_names: Optional[List[Tensor]]
> 
> so w/ a description of why you want to populate this field, what functions it can help you implement, etc...

sg, updated the doc, we can't add these fields by default since it's optional to define these fields

PR #2711, opened 2025-08-07
Let quantizers add custom meta data to GraphModule

per:
Make the 'custom' meta data field in the GraphModule be propageted with any data that is added by the quantizer.

### DIFF
--- a/torchao/quantization/pt2e/quantize_pt2e.py
+++ b/torchao/quantization/pt2e/quantize_pt2e.py
@@ -116,6 +116,8 @@ def prepare_pt2e(
     model = quantizer.transform_for_annotation(model)
     quantizer.annotate(model)
     quantizer.validate(model)
+    # Store the 'custom' meta data if any added by quantizer
+    annotated_meta = model.meta.get("custom", {})
     model = prepare(
         model,
         node_name_to_scope,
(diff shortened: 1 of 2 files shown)


per:
@msaroufim, @cpuhrsch: You are marked as code owners, so reaching out to you. Who do I ping to get the workflow running and reviewers added?

torchao/quantization/pt2e/quantize_pt2e.py
```diff
@@ -116,13 +116,17 @@ def calibrate(model, data_loader):
     model = quantizer.transform_for_annotation(model)
     quantizer.annotate(model)
     quantizer.validate(model)
+    # Store the 'custom' meta data if any added by quantizer
+    annotated_meta = model.meta.get("custom", {})
```
digantdesai, line 120: what's the intended use for this?

per: I want to pass on a list of the annotated nodes to the partitioner and then on to the exported programs graph_module in order to validate the quantization folding decisions in the backend when we start to mix INT and FP. The ET part of it (passing on the 'custom' meta to the edge program) is also needed, but otherwise the 'custom' field is handled in the re-exports to survive all the way to that point.

per: @digantdesai I realised that the nodes 'custom' meta is propagated correctly already, so closing this PR.

Issue #2630, opened 2025-07-30
Deprecate and remove old QAT API


andrewor14:
Following https://github.com/pytorch/ao/pull/2629, we plan to change the main QAT API (more details in the PR description). Today we keep around the following classes for BC without any warning:

```
# Replaced by QATConfig
IntXQuantizationAwareTrainingConfig
FromIntXQuantizationAwareTrainingConfig
intx_quantization_aware_training_config
from_intx_quantization_aware_training_config

# Replaced by IntxFakeQuantizeConfig
FakeQuantizeConfig

# Replaced by IntxFakeQuantizer
FakeQuantizer
```

The plan is to deprecate these configs before v0.13.0, migrate all known users, and remove these configs before v0.14.0. All references to the old API in the docs and README should already be made in https://github.com/pytorch/ao/pull/2629.

andrewor14:
Just testing

@claude can you see this message?

andrewor14:
@claude is this issue done?

PR #2714, opened 2025-08-07
Generalize FakeQuantizer beyond intx

andrewor14:
* __->__ #2714

**Summary:** Similar to https://github.com/pytorch/ao/pull/2628,
but for `FakeQuantizer`. It is cleaner to isolate the logic of
each quantizer in separate classes, e.g. intx vs nvfp4 vs fp8.
Naming change:

```
FakeQuantizer -> IntxFakeQuantizer
```

**BC-breaking notes:** This is technically not BC-breaking yet
since we are just deprecating the old APIs while keeping them
around. It will be when we do remove the old APIs in the future
according to https://github.com/pytorch/ao/issues/2630.

Before:
```
config = IntxFakeQuantizeConfig(torch.int8, "per_channel")
FakeQuantizer(config)
```

After:
```
config = IntxFakeQuantizeConfig(torch.int8, "per_channel")
IntxFakeQuantizer(config) # or
FakeQuantizerBase.from_config(config)
```

**Test Plan:**
```
```

### DIFF
--- a/docs/source/api_ref_qat.rst
+++ b/docs/source/api_ref_qat.rst
@@ -28,7 +28,8 @@ Custom QAT APIs
     IntxFakeQuantizeConfig
     FakeQuantizedLinear
     FakeQuantizedEmbedding
-    FakeQuantizer
+    FakeQuantizerBase
+    IntxFakeQuantizer
     linear.enable_linear_fake_quant
     linear.disable_linear_fake_quant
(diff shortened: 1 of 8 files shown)


jerryzh168:
if not BC breaking, probably don't need to a bc-breaking note, seems like more of a deprecation note

PR #2716, opened 2025-08-07
Don't call erase if node is already erased in batch norm fusion.

abeakkas:
Summary:
The erase call results in a warning log if the node is already erased by the dead code elimination earlier.

Checking if the node is already erased so we don't unnecessarily pollute the log with warnings.

Differential Revision: D79846881


abeakkas:
Not sure how to add a topic label. I don't think I have the permission.

jerryzh168 (approved):

PR #2717, opened 2025-08-08
Remove dep on protype MoEQuantConfig

jerryzh168:
Summary:
att

Test Plan:
python test/quantization/quantize_/workflows/float8/test_float8_tensor.py

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/testing/model_architectures.py
+++ b/torchao/testing/model_architectures.py
@@ -8,6 +8,7 @@ import re
 
 import torch
 import torch.nn as nn
+import torch.nn.functional as F
 
 
 # TODO: Refactor torchao and tests to use these models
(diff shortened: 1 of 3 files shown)


HDCharles (approved):
looks good

torchao/testing/model_architectures.py
```diff
@@ -177,3 +178,63 @@ def create_model_and_input_data(
     else:
         raise ValueError(f"Unknown model type: {model_type}")
     return model, input_data
+
+
+class Llama4Experts(nn.Module):
```
HDCharles: would be good for this to actual match Llama4

https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama4/modeling_llama4.py#L45-L74

since we've adopted those attribute and parameter names for the AOQuantizable modules

jerryzh168: this is from https://github.com/meta-llama/llama-models/blob/a9c89c471f793423afd4cc3ca8671d6e56fe64cb/models/llama4/moe.py#L22, we can ask POC there to see if it makes sense to refactor that or not separately

PR #2719, opened 2025-08-08
Deprecate old TORCH_VERSION variables

andrewor14:
* __->__ #2719

**Summary:** This commit deprecates the following variables:

```
TORCH_VERSION_AT_LEAST_2_5
TORCH_VERSION_AT_LEAST_2_4
TORCH_VERSION_AT_LEAST_2_3
TORCH_VERSION_AT_LEAST_2_2
TORCH_VERSION_AFTER_2_5
TORCH_VERSION_AFTER_2_4
TORCH_VERSION_AFTER_2_3
TORCH_VERSION_AFTER_2_2
```

As of this commit, the latest released version of PyTorch is 2.8,
which means we can drop support for 2.5 and before since we only
support 3 of the latest releases.

The next commit will remove usages of all of these variables
from within torchao.

**Test Plan:**
```
python test/test_utils.py -k torch_version_deprecation
```

### DIFF
--- a/torchao/utils.py
+++ b/torchao/utils.py
@@ -8,6 +8,7 @@ import importlib
 import itertools
 import re
 import time
+import warnings
 from functools import reduce
 from importlib.metadata import version
 from math import gcd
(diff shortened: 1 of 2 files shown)


torchao/utils.py
```diff
@@ -377,13 +378,52 @@ def torch_version_at_least(min_version):
     return is_fbcode() or compare_versions(torch.__version__, min_version) >= 0
 
 
+def _deprecated_torch_version_at_least(version_str: str) -> str:
+    version_str_var_name = "_".join(version_str.split(".")[:2])
+    deprecation_msg = f"TORCH_VERSION_AT_LEAST_{version_str_var_name} is deprecated and will be removed in torchao 0.14.0"
+    return _BoolDeprecationWrapper(
+        torch_version_at_least(version_str),
```
jerryzh168, line 389: why is current torch version related to this? this is talking about deprecation of these variables right? if so, then these variables should be deprecated now, regardless of what the system torch version is?

andrewor14: yeah I can deprecate them all. I think these variables exist so we don't keep calling this function, but yeah we definitely don't need to expose them

andrewor14:
> e.g. for TORCH_VERSION_AT_LEAST("2.7.0"), commonly we might expect this to be 2.7.0 stable and 2.8.0.dev and above, and 2.7.0.dev is not greater than 2.7.0 (because after 2.6.0 release, the version will be 2.7.0.dev until 2.7.0 is released)
but currently, 2.7.0 and 2.7.0.dev will return True

You mean 2.7.0.dev should not be True because of the following ordering?
```
2.6.0 (stable) < 2.7.0.dev < 2.7.0 (stable)
```
But currently the code thinks:
```
# current state, but this is wrong
2.6.0 (stable) < 2.7.0 (stable) <= 2.7.0.dev
```
Is that right?

andrewor14:
Filed an issue here: https://github.com/pytorch/ao/issues/2722. I think we should fix that separately because this PR doesn't change any semantics in its current state.

torchao/utils.py
andrewor14: Ok just deprecated them all

jerryzh168:
> > e.g. for TORCH_VERSION_AT_LEAST("2.7.0"), commonly we might expect this to be 2.7.0 stable and 2.8.0.dev and above, and 2.7.0.dev is not greater than 2.7.0 (because after 2.6.0 release, the version will be 2.7.0.dev until 2.7.0 is released)
> > but currently, 2.7.0 and 2.7.0.dev will return True
> 
> You mean 2.7.0.dev should not be True because of the following ordering?
> 
> ```
> 2.6.0 (stable) < 2.7.0.dev < 2.7.0 (stable)
> ```
> 
> But currently the code thinks:
> 
> ```
> # current state, but this is wrong
> 2.6.0 (stable) < 2.7.0 (stable) <= 2.7.0.dev
> ```
> 
> Is that right?

yeah exactly

torchao/utils.py
jerryzh168: OK thanks, also the deprecation doesn't need to depend on the current pytorch version I think

we could also add a cache for `torch_version_at_least` as well I think

andrewor14: > deprecation doesn't need to depend on the current pytorch version

Do you mean just mention `TORCH_VERSION_AT_LEAST_* is deprecated` in the warning instead of including the version number?

jerryzh168: so the current deprecation message seems to be printed only if the current pytorch version is above some `version_str`, what I'm proposing is just remove this check and always print the deprecation

andrewor14: the warning should always be printed I think. The current pytorch version is not used to decide whether or not we print it (it's only passed to `torch_version_at_least`)

jerryzh168: oh OK, I misunderstood, so the bool value is to pass around the return value of the `torch_version_at_least` function while also print the message

jerryzh168 (approved):

torchao/utils.py
andrewor14: yeah exactly

PR #2720, opened 2025-08-08
Drop support for PyTorch 2.5 and before

andrewor14:
* (to be filled)

**Summary:** We gate on the PyTorch version throughout the repo.
Recently PyTorch 2.8 was released, so the oldest PyTorch version
we need to support is 2.6. After this commit, we assume the user
is running PyTorch 2.6+, and remove all references to the following
variables, which are deprecated.

```
TORCH_VERSION_AT_LEAST_2_6
TORCH_VERSION_AT_LEAST_2_5
TORCH_VERSION_AT_LEAST_2_4
TORCH_VERSION_AT_LEAST_2_3
TORCH_VERSION_AT_LEAST_2_2
TORCH_VERSION_AFTER_2_5
TORCH_VERSION_AFTER_2_4
TORCH_VERSION_AFTER_2_3
TORCH_VERSION_AFTER_2_2
```

**Test Plan:** CI

### DIFF
--- a/test/quantization/pt2e/test_duplicate_dq.py
+++ b/test/quantization/pt2e/test_duplicate_dq.py
@@ -11,6 +11,7 @@ import unittest
 from typing import Any
 
 import torch
+from torch.export import export_for_training
 from torch.testing._internal.common_quantization import QuantizationTestCase
 from torch.testing._internal.common_utils import IS_WINDOWS, run_tests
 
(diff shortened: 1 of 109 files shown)


test/quantization/pt2e/test_duplicate_dq.py
```diff
@@ -11,6 +11,7 @@
 from typing import Any
 
 import torch
+from torch.export import export_for_training
```
jerryzh168, line 14: separate PR: we want to use `torch.export.export` now if the min pt version support is 2.6 I think

jerryzh168 (approved):

test/quantization/pt2e/test_duplicate_dq.py
andrewor14: Just quickly did this here: https://github.com/pytorch/ao/pull/2724

PR #2721, opened 2025-08-08
Remove old `change_linear_weights_to_*` APIs

andrewor14:
* #2724
* __->__ #2721
* #2720

**Summary:** This commit removes these super old quantization
APIs that aren't even accessible by the user:

```
change_linear_weights_to_int8_dqtensors
change_linear_weights_to_int8_woqtensors
change_linear_weights_to_int4_woqtensors
```

**BC-breaking notes:**

Before:
```
change_linear_weights_to_int8_dqtensors(model)
change_linear_weights_to_int8_woqtensors(model)
change_linear_weights_to_int4_woqtensors(model)
```

After:
```
quantize_(model, Int8WeightOnlyConfig())
quantize_(model, Int4WeightOnlyConfig())
```

**Test Plan:** CI


torchao/quantization/quant_api.py
```diff
@@ -120,9 +120,6 @@
     ZeroPointDomain,
 )
 from .subclass import (
```
jerryzh168, line 122: this file can be deleted as well?

jerryzh168: also probably https://github.com/pytorch/ao/blob/6cfa47705f60ea614695b52b4b120ac5fd84d1cb/torchao/quantization/dynamic_quant.py and https://github.com/pytorch/ao/blob/6cfa47705f60ea614695b52b4b120ac5fd84d1cb/torchao/quantization/weight_only.py

andrewor14: Looks like classes from this file is still being used in autoquant. I don't think we can delete or even deprecate yet...

andrewor14: Ok, created an issue here instead: https://github.com/pytorch/ao/issues/2745. Let's do them separately because there are autoquant/benchmark dependencies

jerryzh168:
BC breaking note? we can direct them to new APIs I think

Issue #2612, opened 2025-07-27
Missing benchmark for `sparse24_sm90_sparsify` overhead


namgyu-youn:
Hi, I have a question about your team's workshop, [Accelerating Transformer Inference and Training with 2:4 Activation Sparsity](https://arxiv.org/abs/2503.16672). It seems that the benchmark for overhead when converting Squared-ReLU into SparseGEMM is missing.

For the background, Squared-ReLU has high (>85%) sparsity and accuracy loss compared to SWiGLU. Since Squared-ReLU has a high sparsity, it opens a ticket for **compression** to Sparse-GEMM with 2:4 sparsity. Compression logic is in Figure 3 and the following:

https://github.com/pytorch/ao/blob/ebfe1736c4442970835b6eda833c0bc5a1ce2dda/torchao/prototype/sparsity/activation/srelu_linear.py#L57-L64

But I haven't found the overhead for compression (`sparse24_sm90_sparsify`). Does the new benchmark for `sparse24_sm90_sparsify` make sense? Maybe it would be integrated into `benchmarks/benchmark_e2e_fp8_sparse_linear.py`.

jainapurva:
@jcaip Can you please look into this

namgyu-youn:
@jainapurva Sorry for the distraction, but could you let me know who is engaging in the `sparsity` module? Not certain if there is any progress or roadmap because the last commit log was 3 months ago. Even more, there is no comment about sparsity issues (in #2612 and #2517).

Personally, I am looking at other issues but am more interested in `sparsity` module. Could you lead me for deeper contribution?

jainapurva:
@namgyu-youn As of today, we're not actively building sparsity, but we'll support the existing sparsity module. If you're interested in contributing in sparsity, @jcaip is the main POC for it. He's currently on vacation, and I'll be happy to assist in his absence.

namgyu-youn:
> [@namgyu-youn](https://github.com/namgyu-youn) As of today, we're not actively building sparsity, but we'll support the existing sparsity module. If you're interested in contributing in sparsity, [@jcaip](https://github.com/jcaip) is the main POC for it. He's currently on vacation, and I'll be happy to assist in his absence.

[...]

PR #2728, opened 2025-08-11
Make SmoothQuant more General

namgyu-youn:
Summary:
- Added SmoothQuantConfig as a base config and made corresponding changes in other parts of the flow
- Related Issue/PR: https://github.com/pytorch/ao/pull/2400 

Test Plan:
- Qwen 3-8B (static/dynamic) with `example.py` and unittest
- Additional tests for more tasks required

ETC
- Fix typo in README.md for SmoothQuant

### DIFF
--- a/test/prototype/test_smoothquant.py
+++ b/test/prototype/test_smoothquant.py
@@ -3,7 +3,6 @@
 #
 # This source code is licensed under the BSD 3-Clause license found in the
 # LICENSE file in the root directory of this source tree.
-import tempfile
 import unittest
 from copy import deepcopy
 
@@ -13,9 +12,6 @@ from torch.testing._internal import common_utils
 from torchao.prototype.smoothquant import (
     SmoothQuantConfig,
     SmoothQuantObservedLinear,
-    insert_smooth_quant_observer_,
-    load_smooth_quant_recipe,
-    save_smooth_quant_recipe,
 )
 from torchao.quantization import quantize_
 from torchao.quantization.utils import (
@@ -86,14 +82,15 @@ class TestSmoothQuant(unittest.TestCase):
         test_data = torch.randn(2, 32, dtype=input_dtype, device=device)
 
         # Step 1: Setup quantized model with observer insertion and calibration
-        insert_smooth_quant_observer_(m, alpha, quant_mode)
+        config = SmoothQuantConfig(step="prepare", alpha=alpha, quant_mode=quant_mode)
+        quantize_(m, config)
 
         # Perform calibration with test data
         m(test_data)
 
         # Apply quantization configuration
-        is_observed_linear = lambda m, fqn: isinstance(m, SmoothQuantObservedLinear)
-        quantize_(m, SmoothQuantConfig(), is_observed_linear)
+        config.step = "convert"
+        quantize_(m, config)
 
         # Apply compilation if supported
         if TORCH_VERSION_AT_LEAST_2_5:
@@ -174,6 +171,43 @@ class TestSmoothQuant(unittest.TestCase):
                 f"device={device}, dtype={input_dtype}",
             )
 
+    def test_observer_insertion(self):
+        """Test that PREPARE step correctly inserts SmoothQuantObservedLinear."""
+
+        class SimpleLinear(torch.nn.Module):
+            def __init__(self, bias: bool):
+                super().__init__()
+                self.fc = torch.nn.Linear(32, 32, bias)
+
+            def forward(self, x):
+                return self.fc(x)
+
+        m = SimpleLinear(True).eval()
+
+        # Before quantization - should be regular Linear
+        self.assertIsInstance(m.fc, torch.nn.Linear)
+        self.assertNotIsInstance(m.fc, SmoothQuantObservedLinear)
+
+        # PREPARE step - should insert observers
+        config = SmoothQuantConfig(step="prepare", alpha=0.5, quant_mode="dynamic")
+        quantize_(m, config)
+
+        # After PREPARE - should be SmoothQuantObservedLinear
+        self.assertIsInstance(m.fc, SmoothQuantObservedLinear)
+        self.assertTrue(hasattr(m.fc, "obs"))
+
+        # Test calibration
+        test_data = torch.randn(2, 32)
+        m(test_data)
+
+        # CONVERT step - should produce regular Linear with quantized weights
+        config.step = "convert"
+        quantize_(m, config)
+
+        # After CONVERT - should be regular Linear again (but quantized)
+        self.assertIsInstance(m.fc, torch.nn.Linear)
+        self.assertNotIsInstance(m.fc, SmoothQuantObservedLinear)
+
     @unittest.skip("This test is broken on recent PyTorch, TODO(#1639): fix it")
     @common_utils.parametrize("alpha", [None, 0.5, 0.75])
     @common_utils.parametrize("quant_mode", ["static", "dynamic"])
@@ -181,19 +215,19 @@ class TestSmoothQuant(unittest.TestCase):
         "device", ["cpu"] + (["cuda"] if torch.cuda.is_available() else [])
     )
     @common_utils.parametrize("input_dtype", [torch.float, torch.bfloat16, torch.half])
-    def test_save_load_recipe(self, alpha, quant_mode, device, input_dtype):
-        """Test save/load recipe functionality."""
+    def test_two_step_quantization(self, alpha, quant_mode, device, input_dtype):
+        """Test two-step quantization process (PREPARE -> CONVERT)."""
         dataset_size = 20
         layer_dims = (512, 256, 128)  # Input, hidden, output dimensions
         n_calib_examples = 10
         sequence_length = 5
 
         # Create two identical models for comparison
-        m = ToyLinearModel(*layer_dims).eval().to(input_dtype).to(device)
-        m_save_load = deepcopy(m)
+        m1 = ToyLinearModel(*layer_dims).eval().to(input_dtype).to(device)
+        m2 = deepcopy(m1)
 
         # Generate calibration dataset
-        dataset = m.example_inputs(
+        dataset = m1.example_inputs(
             dataset_size,
             sequence_length=sequence_length,
             dtype=input_dtype,
@@ -201,71 +235,43 @@ class TestSmoothQuant(unittest.TestCase):
         )
         calibration_data = dataset[:n_calib_examples]
 
-        # Step 1: Setup first quantized model with observer insertion and calibration
-        insert_smooth_quant_observer_(m, alpha, quant_mode)
+        # Step 1: PREPARE - Insert observers
+        config = SmoothQuantConfig(step="prepare", alpha=alpha, quant_mode=quant_mode)
+        quantize_(m2, config)
 
-        # Perform calibration with calibration data
+        # Step 2: Calibration
         for data in calibration_data:
-            m(data)
+            m2(data)
 
         # Apply quantization configuration
         is_observed_linear = lambda m, fqn: isinstance(m, SmoothQuantObservedLinear)
-        quantize_(m, SmoothQuantConfig(), is_observed_linear)
+        quantize_(m2, SmoothQuantConfig(), is_observed_linear)
 
         # Apply compilation if supported
         if TORCH_VERSION_AT_LEAST_2_5:
-            m = torch.compile(m, fullgraph=True)
+            m2 = torch.compile(m2, fullgraph=True)
 
-        # Step 2: Setup save/load model with recipe functionality
-        insert_smooth_quant_observer_(m_save_load, alpha, quant_mode)
-        for example in calibration_data:
-            m_save_load(example.to(device))
+        # Step 4: Validate outputs on full dataset
+        with torch.inference_mode():
+            m2_outputs = []
 
-        # Step 3: Test save/load recipe functionality
-        with tempfile.NamedTemporaryFile() as temp_file:
-            save_path = temp_file.name
-            save_smooth_quant_recipe(m_save_load, save_path)
-            load_smooth_quant_recipe(m_save_load, save_path)
+            for data in dataset:
+                # Remove batch dimension for model input
+                input_tensor = data.squeeze(0)
+                m2_output = m2(input_tensor)
+                m2_outputs.append(m2_output)
 
-            # Step 4: Complete quantization for save/load model
-            is_observed_linear = lambda m, fqn: isinstance(m, SmoothQuantObservedLinear)
-            quantize_(m_save_load, SmoothQuantConfig(), is_observed_linear)
+            # Concatenate all outputs
+            m2_result = torch.cat(m2_outputs)
 
-            if TORCH_VERSION_AT_LEAST_2_5:
-                m_save_load = torch.compile(m_save_load, fullgraph=True)
+            self.assertIsNotNone(m2_result, "Quantized model output should not be None")
 
-            # Step 5: Validate outputs on full dataset
-            with torch.inference_mode():
-                original_outputs = []
-                save_load_outputs = []
-
-                for data in dataset:
-                    # Remove batch dimension for model input
-                    input_tensor = data.squeeze(0)
-
-                    original_output = m(input_tensor)
-                    save_load_output = m_save_load(input_tensor)
-
-                    original_outputs.append(original_output)
-                    save_load_outputs.append(save_load_output)
-
-                # Concatenate all outputs for comparison
-                original_result = torch.cat(original_outputs)
-                save_load_out = torch.cat(save_load_outputs)
-
-                self.assertIsNotNone(
-                    original_result, "Original model output should not be None"
-                )
-                self.assertIsNotNone(
-                    save_load_out, "Save/load model output should not be None"
-                )
-
-                torch.testing.assert_close(
-                    original_result,
-                    save_load_out,
-                    msg=f"Save/load recipe should produce identical results for "
-                    f"alpha={alpha}, quant_mode={quant_mode}, device={device}, dtype={input_dtype}",
-                )
+            # Check that model produces reasonable outputs
+            self.assertFalse(
+                torch.isnan(m2_result).any(),
+                f"Quantized model should not produce NaN values for "
+                f"alpha={alpha}, quant_mode={quant_mode}, device={device}, dtype={input_dtype}",
+            )
 
 
 common_utils.instantiate_parametrized_tests(TestSmoothQuant)
--- a/torchao/prototype/smoothquant/README.md
+++ b/torchao/prototype/smoothquant/README.md
@@ -1,4 +1,4 @@
-# SmothQuant quantization
+# SmoothQuant quantization
 This is a native PyTorch implementation of the algorithm described in [this paper](https://arxiv.org/abs/2211.10438).
 
 In this implementation, weights are smoothed (equalized) and quantized to int8 during quantization. Activations are smoothed and quantized to int8 at runtime. Quantization is done either dynamically or statically. If activations are dynamically quantized, qparams (i.e., scales) are found at runtime while qparams are found during quantization for static quantization. For dynamic quantization, activations are quantized per token. And for static quantization, activations are quantized per tensor. Generally, dynamic quantization produces better accuracy while static quantization has better latency. In both cases, weights and activations are symmetrically quantized.
@@ -6,21 +6,21 @@ In this implementation, weights are smoothed (equalized) and quantized to int8 d
 ## Quick start
 Run the example code with
 ```bash
-python example.py -m MODLE_ID --device=<cuda or cpu> --quant-mode=<dynamic or static>
+python example.py -m MODEL_ID --device=<cuda or cpu> --quant-mode=<dynamic or static>
 # An example
 python example.py -m meta-llama/Llama-2-7b-hf --device=cuda --quant-mode=dynamic
 ```
 To use the `torch.compile` for speedup, add `--compile`. You may want to export `TORCHINDUCTOR_FREEZING=1` for even better performance.
 ```bash
-TORCHINDUCTOR_FREEZING=1 python example.py -m MODLE_ID --device=<cuda or cpu> --quant-mode=<dynamic or static> --compile
+TORCHINDUCTOR_FREEZING=1 python example.py -m MODEL_ID --device=<cuda or cpu> --quant-mode=<dynamic or static> --compile
 ```
 To save a quantized model for reuse, specify `--model-save-path`
 ```bash
-python example.py -m MODLE_ID --device=<cuda or cpu> --quant-mode=<dynamic or static> --model-save-path ./quantized_model.pt
+python example.py -m MODEL_ID --device=<cuda or cpu> --quant-mode=<dynamic or static> --model-save-path ./quantized_model.pt
 ```
 And load it by `--model-load-path`
 ```bash
-python example.py -m MODLE_ID --device=<cuda or cpu> --quant-mode=<dynamic or static> --model-load-path ./quantized_model.pt
+python example.py -m MODEL_ID --device=<cuda or cpu> --quant-mode=<dynamic or static> --model-load-path ./quantized_model.pt
 ```
 
 
--- a/torchao/prototype/smoothquant/api.py
+++ b/torchao/prototype/smoothquant/api.py
@@ -5,18 +5,13 @@
 # LICENSE file in the root directory of this source tree.
 import types
 from dataclasses import dataclass
-from typing import Dict, Optional
+from typing import Optional
 
 import torch
 
 import torchao
 from torchao.core.config import AOBaseConfig
 from torchao.dtypes import to_affine_quantized_intx, to_affine_quantized_intx_static
-from torchao.prototype.smoothquant.core import (
-    SmoothQuantObservedLinear,
-    SmoothQuantObserver,
-)
-from torchao.quantization import quantize_
 from torchao.quantization.linear_activation_quantized_tensor import (
     to_linear_activation_quantized,
 )
@@ -25,109 +20,52 @@ from torchao.quantization.linear_activation_scale import (
 )
 from torchao.quantization.quant_api import (
     _linear_extra_repr,
-    _replace_with_custom_fn_if_matches_filter,
 )
 from torchao.quantization.quant_primitives import MappingType
 from torchao.quantization.transform_module import (
     register_quantize_module_handler,
 )
 from torchao.quantization.utils import _get_per_token_block_size
-from torchao.quantization.weight_tensor_linear_activation_quantization import (
-    to_weight_tensor_with_linear_activation_quantization_metadata,
+
+from .core import (
+    SmoothQuantObservedLinear,
+    SmoothQuantObserver,
+    SmoothQuantStep,
 )
 
 
-def insert_smooth_quant_observer_(
-    model: torch.nn.Module, alpha: Optional[float] = 0.5, quant_mode: str = "dynamic"
-):
+@dataclass
+class SmoothQuantConfig(AOBaseConfig):
     """
-    Inserts SmoothQuantObserver into Linear layers of a given model.
+    Configuration for SmoothQuant quantization when passed into quantize_()
 
     Args:
-        model: The model to be modified (in place). Ensure model is on the desired device for calibration
+        step (SmoothQuantStep): The step for SmoothQuant process
+            PREPARE: insert SmoothQuant Observers to linear layers
+            CONVERT: convert the observed linear modules to quantized modules
         alpha: The alpha value to determine smoothing factor. Factor = 1 if alpha is None, which means
-               falling back to conventional quantization.
+            Fall back to conventional quantization if None
         quant_mode: dynamic or static quantization of activation
+        smoothing_factor: The smoothing factor for the layer. Acquired from the layer's observer if None.
+        act_scales: The activation scales for the layer. Acquired from the layer's observer if None.
+        wei_scales: The weight scales for the layer. Acquired from the layer's observer if None.
+        set_inductor_config: if True, adjusts `torchinductor` settings to recommended values.
     """
-    _is_linear = lambda m, fqn: isinstance(m, torch.nn.Linear)
 
-    quant_min, quant_max = -127, 127
-    eps = torch.finfo(torch.float32).eps
+    step: SmoothQuantStep
+    alpha: Optional[float] = 0.5
+    quant_mode = "dynamic"
+    smoothing_factor: Optional[torch.Tensor] = None
+    act_scales: Optional[torch.Tensor] = None
+    wei_scales: Optional[torch.Tensor] = None
+    set_inductor_config = True
 
-    def replace_with_observer(layer):
-        # creates observer and replaces linear layers with observed linear layers
-        observer = SmoothQuantObserver(
-            layer.weight,
-            alpha,
-            quant_mode,
-            quant_min=quant_min,
-            quant_max=quant_max,
-            eps=eps,
-        )
-        return SmoothQuantObservedLinear.from_float(layer, observer)
-
-    _replace_with_custom_fn_if_matches_filter(model, replace_with_observer, _is_linear)
-
-
-def save_smooth_quant_recipe(
-    model: torch.nn.Module, save_path: str
-) -> Dict[str, torch.Tensor]:
-    """
-    Save smoothing_factors, act_scales, and wei_scales for each SmoothQuantObservedLinear layer in the model.
-    """
-    result = {}
-
-    def recurse(module: torch.nn.Module, name: str = ""):
-        for child_name, child in module.named_children():
-            full_name = f"{name}.{child_name}" if name else child_name
-
-            # Apply the analysis function to this layer
-            if isinstance(child, SmoothQuantObservedLinear):
-                smoothing_factor, act_scales, wei_scales = child.obs.calculate_qparams()
-                result[full_name + ".smoothing_factor"] = smoothing_factor
-                result[full_name + ".act_scales"] = act_scales
-                result[full_name + ".wei_scales"] = wei_scales
-
-            # Recurse into child modules
-            recurse(child, full_name)
-
-    recurse(model)
-
-    torch.save(result, save_path)
-
-
-def load_smooth_quant_recipe(
-    model: torch.nn.Module, recipe_path: str, device=None
-) -> torch.nn.Module:
-    recipe = torch.load(recipe_path, weights_only=True)
-
-    def recurse(module: torch.nn.Module, name: str = ""):
-        if isinstance(module, SmoothQuantObservedLinear):
-            smoothing_factor = recipe.get(name + ".smoothing_factor", None)
-            act_scales = recipe.get(name + ".act_scales", None)
-            wei_scales = recipe.get(name + ".wei_scales", None)
-            if device is not None:
-                module.to(device=device)
-            # act_scales is None for dynamic quantization
-            if any(x is None for x in (smoothing_factor, wei_scales)):
-                return module
-            is_observed_linear = lambda m, fqn: isinstance(m, SmoothQuantObservedLinear)
-            wrapper = torch.nn.Sequential(module)
-            quantize_(
-                wrapper,
-                SmoothQuantConfig(smoothing_factor, act_scales, wei_scales),
-                is_observed_linear,
-            )
-            return wrapper[0]
-
-        mod_new = module
-
-        for child_name, child in module.named_children():
-            full_name = f"{name}.{child_name}" if name else child_name
-            setattr(mod_new, child_name, recurse(child, full_name))
-        return mod_new
-
-    recurse(model)
+    def __post_init__(self):
+        self.step = self.step.lower() if isinstance(self.step, str) else self.step.value
+        all_step_values = [s.value for s in SmoothQuantStep]
+        if self.step not in all_step_values:
+            raise ValueError(f"{self.step} is not one of {all_step_values}")
+        assert self.quant_mode in ["static", "dynamic"]
 
 
 class _ActQuantizer:
@@ -145,46 +83,66 @@ class _ActQuantizer:
         )
 
     def static_quantize(self, input, scale, zero_point):
+        # Use tensor-wise quantization for static mode
+        # This matches the expected behavior for SmoothQuant static quantization
         return to_affine_quantized_intx_static(
             input,
             scale,
             zero_point,
-            list(input.shape),
+            (1,) + (1,) * (input.ndim - 1),
             self.target_dtype,
             self.quant_min,
         )
-
-
-@dataclass
-class SmoothQuantConfig(AOBaseConfig):
-    """
-    Configuration for quantizing linear layers when passed into quantize_()
-
-    Args:
-        smoothing_factor: The smoothing factor for the layer. Acquired from the layer's observer if None.
-        act_scales: The activation scales for the layer. Acquired from the layer's observer if None.
-        wei_scales: The weight scales for the layer. Acquired from the layer's observer if None.
-        set_inductor_config: if True, adjusts `torchinductor` settings to recommended values.
-    """
-
-    smoothing_factor: Optional[torch.Tensor] = None
-    act_scales: Optional[torch.Tensor] = None
-    wei_scales: Optional[torch.Tensor] = None
-    set_inductor_config: bool = True
 
 
 @register_quantize_module_handler(SmoothQuantConfig)
 def _smooth_quant_transform(
     module: torch.nn.Module,
     config: SmoothQuantConfig,
-):
-    smoothing_factor = config.smoothing_factor
-    act_scales = config.act_scales
-    wei_scales = config.wei_scales
+) -> torch.nn.Module:
+    step = config.step
+    observed_linear = None
+
+    if step == SmoothQuantStep.PREPARE:
+        observer = SmoothQuantObserver(
+            module.weight,
+            config.alpha,
+            config.quant_mode,
+            quant_min=-127,
+            quant_max=127,
+            eps=torch.finfo(torch.float32).eps,
+        )
+        return SmoothQuantObservedLinear.from_float(module, observer)
+
+    elif step == SmoothQuantStep.CONVERT:
+        if not isinstance(module, SmoothQuantObservedLinear):
+            print(
+                f"convert: module is not SmoothQuantObservedLinear, skipping: {type(module)}"
+            )
+            return module
+        observed_linear = module
+    else:
+        raise ValueError(f"Unexpected step: {step}")
+
+    # Convert observed linear to quantized linear
     if config.set_inductor_config:
         torchao.quantization.utils.recommended_inductor_config_setter()
-    observed_linear = module
 
+    # Get quantization parameters
+    if all(x is not None for x in (config.smoothing_factor, config.wei_scales)):
+        smoothing_factor, act_scales, wei_scales = (
+            config.smoothing_factor,
+            config.act_scales,
+            config.wei_scales,
+        )
+        weight = observed_linear.weight * smoothing_factor
+    else:
+        smoothing_factor, act_scales, wei_scales = (
+            observed_linear.obs.calculate_qparams()
+        )
+        weight = observed_linear.obs.weight * smoothing_factor
+
+    # Create new linear layer
     linear = torch.nn.Linear(
         observed_linear.in_features,
         observed_linear.out_features,
@@ -194,38 +152,26 @@ def _smooth_quant_transform(
     )
     linear.bias = observed_linear.bias
 
+    # Quantize weights
     target_dtype = torch.int8
-    # act_scales is None for dynamic quantization thus not checked
-    if any(x is None for x in (smoothing_factor, wei_scales)):
-        factor, x_scale, w_scales = observed_linear.obs.calculate_qparams()
-        weight = observed_linear.obs.weight * factor
-    else:
-        factor, x_scale, w_scales = smoothing_factor, act_scales, wei_scales
-        weight = observed_linear.weight * factor
     weight = weight.to(observed_linear.weight.dtype)
     block_size = (1, weight.size(1))
-    wei_zero_points = torch.zeros_like(w_scales, dtype=torch.int64)
+    wei_zero_points = torch.zeros_like(wei_scales, dtype=torch.int64)
+
     qw = to_affine_quantized_intx_static(
-        weight,
-        w_scales,
-        wei_zero_points,
-        block_size,
-        target_dtype,
+        weight, wei_scales, wei_zero_points, block_size, target_dtype
     )
 
-    if x_scale is None:
-        # dynamic quant
-        qw = to_linear_activation_quantized(
-            qw, _ActQuantizer(target_dtype).dynamic_quantize
-        )
-    else:
-        # static quant
-        x_zero_point = torch.zeros_like(x_scale, dtype=torch.int64)
-        qw = to_weight_tensor_with_linear_activation_quantization_metadata(
-            qw, _ActQuantizer(target_dtype).static_quantize, x_scale, x_zero_point
-        )
+    # Apply activation quantization
+    qw = to_linear_activation_quantized(
+        qw, _ActQuantizer(target_dtype).dynamic_quantize
+    )
 
-    qw = to_weight_tensor_with_linear_activation_scale_metadata(qw, factor.to(qw.dtype))
+    # Add smoothing factor metadata
+    qw = to_weight_tensor_with_linear_activation_scale_metadata(
+        qw, smoothing_factor.to(qw.dtype)
+    )
     linear.weight = torch.nn.Parameter(qw, requires_grad=False)
-    linear.extra_repr = types.MethodType(_linear_extra_repr, module)
+    linear.extra_repr = types.MethodType(_linear_extra_repr, linear)
+
     return linear
--- a/torchao/prototype/smoothquant/core.py
+++ b/torchao/prototype/smoothquant/core.py
@@ -3,15 +3,19 @@
 #
 # This source code is licensed under the BSD 3-Clause license found in the
 # LICENSE file in the root directory of this source tree.
+from enum import Enum
 from typing import Optional
 
 import torch
 import torch.nn.functional as F
 
 from torchao.quantization.observer import AffineQuantizedMinMaxObserver, PerAxis
-from torchao.quantization.quant_primitives import (
-    MappingType,
-)
+from torchao.quantization.quant_primitives import MappingType
+
+
+class SmoothQuantStep(str, Enum):
+    PREPARE = "prepare"
+    CONVERT = "convert"
 
 
 class SmoothQuantObserver(torch.nn.Module):
@@ -39,36 +43,28 @@ class SmoothQuantObserver(torch.nn.Module):
         super().__init__()
         assert weight.ndim == 2
         self.weight = weight
-        self.inputs = []
         self.device = self.weight.device
         self.alpha = alpha
-        assert quant_mode in ["static", "dynamic"]
         self.quant_mode = quant_mode
         self.quant_min = quant_min
         self.quant_max = quant_max
-        self.eps = eps
+        self.eps = eps or torch.finfo(torch.float32).eps
         # act.shape = [mb, ic] (reshape if needed), wei.shape = [oc, ic]
         # *_ic_obs are used to determine smoothing_factor
         # wei_oc_obs is used to find qparams for quantization
         self.act_ic_obs = AffineQuantizedMinMaxObserver(
-            MappingType.SYMMETRIC,
-            torch.int8,
-            PerAxis(-1),
-            eps=eps,
+            MappingType.SYMMETRIC, torch.int8, PerAxis(-1), eps=self.eps
         )
         self.wei_ic_obs = AffineQuantizedMinMaxObserver(
-            MappingType.SYMMETRIC,
-            torch.int8,
-            PerAxis(-1),
-            eps=eps,
+            MappingType.SYMMETRIC, torch.int8, PerAxis(-1), eps=self.eps
         )
         self.wei_oc_obs = AffineQuantizedMinMaxObserver(
             MappingType.SYMMETRIC,
             torch.int8,
             PerAxis(0),
-            quant_min=quant_min,
-            quant_max=quant_max,
-            eps=eps,
+            quant_min=self.quant_min,
+            quant_max=self.quant_max,
+            eps=self.eps,
         )
         self.wei_ic_obs(self.weight)
 
@@ -78,7 +74,7 @@ class SmoothQuantObserver(torch.nn.Module):
         return input
 
     def calculate_qparams(self):
-        # 1 Get min/max per IC from observers
+        # Step 1: Get min/max per input channel (IC) from observers
         wei_min_per_ic = self.wei_ic_obs.min_val
         wei_max_per_ic = self.wei_ic_obs.max_val
         act_min_per_ic = self.act_ic_obs.min_val
@@ -89,43 +85,44 @@ class SmoothQuantObserver(torch.nn.Module):
         w_abs_max_per_ic = (
             torch.max(torch.abs(wei_min_per_ic), torch.abs(wei_max_per_ic)) + self.eps
         )
-        # 2 calculate the smoothing factor
+
+        # Step 2: Calculate smoothing factor
         if self.alpha is None:
             # fall back to conventional quantization if alpha is None
-            smoothing_factor = torch.ones_like(
-                x_abs_max_per_ic,
-                dtype=x_abs_max_per_ic.dtype,
-                device=x_abs_max_per_ic.device,
-            )
+            smoothing_factor = torch.ones_like(x_abs_max_per_ic)
         else:
             smoothing_factor = torch.pow(x_abs_max_per_ic, self.alpha) / torch.pow(
                 w_abs_max_per_ic.to(x_abs_max_per_ic.device), 1 - self.alpha
             )
-        # 3 apply smoothing factor to activations and find scales for static quantization
+
+        # Step 3: Calculate activation scales for static quantization
         act_scales = None
         if self.quant_mode == "static":
-            act_min_per_ic_new = act_min_per_ic / smoothing_factor.reshape(
+            act_min_new = act_min_per_ic / smoothing_factor.reshape(
                 act_min_per_ic.shape
             )
-            act_max_per_ic_new = act_max_per_ic / smoothing_factor.reshape(
+            act_max_new = act_max_per_ic / smoothing_factor.reshape(
                 act_max_per_ic.shape
             )
-            min_val_per_tensor = torch.min(act_min_per_ic_new)
-            max_val_per_tensor = torch.max(act_max_per_ic_new)
-            min_val_neg = torch.min(
-                min_val_per_tensor, torch.zeros_like(min_val_per_tensor)
-            )
-            max_val_pos = torch.max(
-                max_val_per_tensor, torch.zeros_like(max_val_per_tensor)
-            )
-            max_val_pos = torch.max(-min_val_neg, max_val_pos)
-            act_scale = max_val_pos / (float(self.quant_max - self.quant_min) / 2)
-            act_scales = act_scale.to(self.device)
-        # 4 update weight and find scales
+
+            # Calculate global scale (scalar)
+            global_min = torch.min(act_min_new)
+            global_max = torch.max(act_max_new)
+            abs_max = torch.max(torch.abs(global_min), torch.abs(global_max))
+            act_scale = abs_max / (float(self.quant_max - self.quant_min) / 2)
+
+            # Create scalar tensor for tensor-wise quantization
+            act_scales = act_scale.reshape(()).to(self.device)  # Ensure scalar shape
+
+        # Step 4: Update weight and find scales
         self.wei_oc_obs(self.weight * smoothing_factor.to(self.device))
         wei_scales, _ = self.wei_oc_obs.calculate_qparams()
-        # 5 return results
-        return smoothing_factor.to(self.device), act_scales, wei_scales.to(self.device)
+
+        return (
+            smoothing_factor.to(self.device),
+            act_scales,
+            wei_scales.to(self.device),
+        )
 
 
 class SmoothQuantObservedLinear(torch.nn.Linear):
@@ -133,27 +130,25 @@ class SmoothQuantObservedLinear(torch.nn.Linear):
         self,
         in_features: int,
         out_features: int,
-        bias: bool,
         obs: SmoothQuantObserver,
+        bias: bool = True,
         device=None,
         dtype=None,
     ):
         super().__init__(in_features, out_features, bias, device, dtype)
-        assert isinstance(obs, SmoothQuantObserver)
         self.obs = obs
 
     def forward(self, input: torch.Tensor):
         input = self.obs(input)
-        output = F.linear(input, self.weight, self.bias)
-        return output
+        return F.linear(input, self.weight, self.bias)
 
     @classmethod
     def from_float(cls, float_linear: torch.nn.Linear, obs: SmoothQuantObserver):
         observed_linear = cls(
             float_linear.in_features,
             float_linear.out_features,
-            float_linear.bias is not None,
             obs,
+            float_linear.bias is not None,
             device=float_linear.weight.device,
             dtype=float_linear.weight.dtype,
         )
--- a/torchao/prototype/smoothquant/example.py
+++ b/torchao/prototype/smoothquant/example.py
@@ -15,8 +15,6 @@ from transformers import AutoModelForCausalLM, AutoTokenizer
 
 from torchao.prototype.smoothquant import (
     SmoothQuantConfig,
-    SmoothQuantObservedLinear,
-    insert_smooth_quant_observer_,
 )
 from torchao.quantization import quantize_
 
@@ -137,8 +135,11 @@ def wikitext2_ppl(
         print(f"Time to load model: {time.time() - t0:.02f} seconds")
         print("running calibration")
         t0 = time.time()
-        # insert observers to find average magnitude and calculate scales
-        insert_smooth_quant_observer_(model, alpha, quant_mode)
+        # Step 1: Insert observers to find average magnitude and calculate scales
+        config = SmoothQuantConfig(step="prepare", alpha=alpha, quant_mode=quant_mode)
+        quantize_(model, config)
+
+        # Step 2: Calibration
         calibration_data = get_calib_dataset(
             tokenizer=tokenizer, n_samples=calibration_size, block_size=sequence_length
         )
@@ -147,10 +148,11 @@ def wikitext2_ppl(
             batch.to("cpu")
         print(f"time for calibration: {time.time() - t0:.02f} seconds")
 
-        is_observed_linear = lambda m, fqn: isinstance(m, SmoothQuantObservedLinear)
+        # Step 3: Convert to quantized model
         print(f"running SmoothQuant with {quant_mode} quantization")
         t0 = time.time()
-        quantize_(model, SmoothQuantConfig(), is_observed_linear)
+        config.step = "convert"
+        quantize_(model, config)
         print(f"time for quantization: {time.time() - t0:.02f} seconds")
         if model_save_path is not None:
             print(f"Saving quantized model to {model_save_path}")
@@ -239,7 +241,7 @@ if __name__ == "__main__":
         args.quant_mode,
         args.calibration_samples,
         args.device,
-        args.precision,
+        precision_dtype,
         args.seq_len,
         args.compile,
         args.model_load_path,
--- a/torchao/prototype/smoothquant/__init__.py
+++ b/torchao/prototype/smoothquant/__init__.py
@@ -1,15 +1,13 @@
-from .api import (
-    SmoothQuantConfig,
-    insert_smooth_quant_observer_,
-    load_smooth_quant_recipe,
-    save_smooth_quant_recipe,
+from .api import SmoothQuantConfig
+from .core import (
+    SmoothQuantObservedLinear,
+    SmoothQuantObserver,
+    SmoothQuantStep,
 )
-from .core import SmoothQuantObservedLinear
 
 __all__ = [
-    "insert_smooth_quant_observer_",
-    "load_smooth_quant_recipe",
-    "save_smooth_quant_recipe",
     "SmoothQuantConfig",
+    "SmoothQuantStep",
+    "SmoothQuantObserver",
     "SmoothQuantObservedLinear",
 ]


namgyu-youn:
@jerryzh168 Could you please look into this PR? It was inspired by https://github.com/pytorch/ao/pull/2659#issuecomment-3146079199 for more generalized SmoothQuant API.

jerryzh168:
Thanks @namgyu-youn this is a step towards that but not fully general yet, it seems to be a quick change to add it though, commented inline.

also it seems smoothquant is not very popular at the moment: https://huggingface.co/models?search=smoothquant, so I'd like to wait a bit before we invest more effort to it, let me know if you are interested to contribute more to torchao, we have many more higher priority issues that you can help with I think

jerryzh168 (approved):

test/prototype/test_smoothquant.py
```diff
@@ -82,14 +55,15 @@ def forward(self, x):
         test_data = torch.randn(2, 32, dtype=input_dtype, device=device)
 
         # Step 1: Setup quantized model with observer insertion and calibration
-        insert_smooth_quant_observer_(m, alpha, quant_mode)
+        config = SmoothQuantConfig(step="prepare", alpha=alpha, quant_mode=quant_mode)
```
jerryzh168: it you want to really make it general, the `quant_mode` has to be changed to base_config, and we'll do a general quantization like this https://github.com/pytorch/ao/blob/2eae09b2258aaa2e0b575f9efa22e9ea00da5acb/torchao/prototype/awq/api.py#L104-L106

torchao/prototype/smoothquant/api.py
```diff
@@ -194,38 +152,26 @@ def _smooth_quant_transform(
...
-        weight = observed_linear.obs.weight * factor
-    else:
-        factor, x_scale, w_scales = smoothing_factor, act_scales, wei_scales
-        weight = observed_linear.weight * factor
     weight = weight.to(observed_linear.weight.dtype)
     block_size = (1, weight.size(1))
-    wei_zero_points = torch.zeros_like(w_scales, dtype=torch.int64)
+    wei_zero_points = torch.zeros_like(wei_scales, dtype=torch.int64)
+
     qw = to_affine_quantized_intx_static(
```
jerryzh168: here this should be updated to https://github.com/pytorch/ao/blob/2eae09b2258aaa2e0b575f9efa22e9ea00da5acb/torchao/prototype/awq/api.py#L104-L106 to make it truly general

namgyu-youn:
> Thanks @namgyu-youn this is a step towards that but not fully general yet, it seems to be a quick change to add it though, commented inline.
> 
> also it seems smoothquant is not very popular at the moment: https://huggingface.co/models?search=smoothquant, so I'd like to wait a bit before we invest more effort to it, let me know if you are interested to contribute more to torchao, we have many more higher priority issues that you can help with I think

Thanks for the kind info, and I truly love your team's work after reviewing [TorchAO: CodeML @ ICML 2025](https://arxiv.org/abs/2507.16099).

The recently updated [contribution guide](https://docs.pytorch.org/ao/stable/contributor_guide.html) could be a great choice for the next contribution, but personally I prefer the sparsity (pruning) module more. Unfortunately, I heard the main POC (@jcaip) is on vacation, making it hard for me to progress. The following are my recent activities related to the sparsity module:
1. Since Wanda was already introduced, I recently introduced Wanda++ at https://github.com/pytorch/ao/pull/2537.
2. Computation overhead was missing in your team's workshop (not certain because of my lack of knowledge), and opened issue at https://github.com/pytorch/ao/issues/2612
3. Also interested in Activation compression https://github.com/pytorch/ao/issues/1920, but I have to learn more about it.

If there is no huge progress for the sparsity module, quantization (new APIs or primitive ops) might be a next step. Let me know if there is a good-second-issue about it.

p.s. Could you please check https://github.com/pytorch/ao/pull/2644 ? It hasn't merged yet after being approved (no CI broken). Also, https://github.com/pytorch/ao/pull/2660 has been waiting for review (I am fine to close this because it is low-priority).

torchao/prototype/smoothquant/api.py
namgyu-youn: Oh, I missed it. The Quantization API should be a choice, thanks!

test/prototype/test_smoothquant.py
namgyu-youn: Thanks, I missed it. Directly using the Quantization API is a choice.

namgyu-youn:
Test result (`test_smoothquant.py`):
```
$ python test/prototype/test_smoothquant.py
..............................................
----------------------------------------------------------------------
Ran 46 tests in 15.208s

OK
```

namgyu-youn:
@jerryzh168 Hi, I am happy to show you more generalized SmoothQuant API by using Quantization API (`torchao/quantization/quant_api.py`) at https://github.com/pytorch/ao/pull/2728/commits/ba89d03d0d49ec16cace1bdf25475eaa23a57e2c. Could you review this PR?

test/prototype/test_smoothquant.py
```diff
@@ -13,252 +12,227 @@
...
+        self.assertNotIsInstance(m.linear1, SmoothQuantObservedLinear)
 
-    @unittest.skip("This test is broken on recent PyTorch, TODO(#1639): fix it")
     @common_utils.parametrize("alpha", [None, 0.5, 0.75])
-    @common_utils.parametrize("quant_mode", ["static", "dynamic"])
     @common_utils.parametrize(
-        "device", ["cpu"] + (["cuda"] if torch.cuda.is_available() else [])
+        "base_config",
+        [
+            int8_dynamic_activation_int8_weight(),
```
jerryzh168: nit: this API is deprecated, use `Int8DynamicActivationInt8WeightConfig` instead

torchao/prototype/smoothquant/README.md
```diff
@@ -1,26 +1,26 @@
...
+# SmoothQuant quantization
 This is a native PyTorch implementation of the algorithm described in [this paper](https://arxiv.org/abs/2211.10438).
 
 In this implementation, weights are smoothed (equalized) and quantized to int8 during quantization. Activations are smoothed and quantized to int8 at runtime. Quantization is done either dynamically or statically. If activations are dynamically quantized, qparams (i.e., scales) are found at runtime while qparams are found during quantization for static quantization. For dynamic quantization, activations are quantized per token. And for static quantization, activations are quantized per tensor. Generally, dynamic quantization produces better accuracy while static quantization has better latency. In both cases, weights and activations are symmetrically quantized.
 
 ## Quick start
 Run the example code with
 ```bash
-python example.py -m MODLE_ID --device=<cuda or cpu> --quant-mode=<dynamic or static>
+python example.py -m MODEL_ID --device=<cuda or cpu> --quant-mode=<dynamic or static>
```
jerryzh168: do you need to update the API here as well? since we now take a config instead of just static/dynamic quant mode flag

torchao/prototype/smoothquant/example.py
```diff
@@ -137,8 +137,15 @@ def wikitext2_ppl(
         print(f"Time to load model: {time.time() - t0:.02f} seconds")
         print("running calibration")
         t0 = time.time()
-        # insert observers to find average magnitude and calculate scales
-        insert_smooth_quant_observer_(model, alpha, quant_mode)
+        # Step 1: Insert observers to find average magnitude and calculate scales
+        config = SmoothQuantConfig(
+            base_config=int8_dynamic_activation_int8_weight(),
```
jerryzh168: can generalize the example API to take quant type configs now, see https://github.com/pytorch/ao/blob/751d7f64892d1bc407f07e033a7978df52534f62/torchao/prototype/awq/example.py#L307

torchao/prototype/smoothquant/api.py
```diff
@@ -5,186 +5,112 @@
...
+        observer = SmoothQuantObserver(
+            weight=module.weight,
+            alpha=config.alpha,
+            quant_min=-127,
+            quant_max=127,
+            eps=torch.finfo(torch.float32).eps,
+        )
+        return SmoothQuantObservedLinear.from_float(module, observer)
+
+    elif step == SmoothQuantStep.CONVERT:
```
jerryzh168, line 88: nit: we might have to add a similar PREPARE_FOR_LOADING as well so that the quantized model can be loaded in transformers, similar to awq: https://github.com/pytorch/ao/blob/5c0d6a3fa6b86f7711165fb1c4a4bdf6bd771944/torchao/prototype/awq/api.py#L77, callsite: https://github.com/pytorch/ao/blob/5c0d6a3fa6b86f7711165fb1c4a4bdf6bd771944/torchao/prototype/awq/example.py#L264

torchao/prototype/smoothquant/README.md
namgyu-youn: Yes now we can update here because there is no need for `--quant-mode`, I will fix it!

torchao/prototype/smoothquant/example.py
namgyu-youn: Thanks, but how about using `Int8DynamicActivationInt8WeightConfig` as a default in here and devide PR? It might require checking which APIs are compatiable with `SmoothQuantConfig`, and building unittest.

~btw, we can uniform commonly used utils functions in AWQ and SmoothQuant: `get_calib_dataset`, `wiki2_eval`, and `quantize_and_eval`.~

jerryzh168: yeah sure

torchao/prototype/smoothquant/core.py
```diff
@@ -31,44 +35,34 @@ def __init__(
...
             MappingType.SYMMETRIC,
             torch.int8,
             PerAxis(0),
-            quant_min=quant_min,
-            quant_max=quant_max,
-            eps=eps,
+            quant_min=self.quant_min,
+            quant_max=self.quant_max,
+            eps=self.eps,
         )
```
jerryzh168: are these still needed?

torchao/prototype/smoothquant/example.py
```diff
@@ -4,185 +4,168 @@
...
+
+    # Step 3: Convert to quantized model
+    print("running SmoothQuant convert")
+    t0 = time.time()
+    quant_config.step = SmoothQuantStep.CONVERT
+    quantize_(model, quant_config)
+    print(f"time for convert: {time.time() - t0:.02f} seconds")
+
+    # Set up config for loading
+    quant_config.step = SmoothQuantStep.PREPARE_FOR_LOADING
```
jerryzh168, line 125: does this work? you can check if it works by the following:

```
export MODEL=YOUR_SAVED_SMOOTHQUANT_MODEL
lm_eval --model hf --model_args pretrained=$MODEL --tasks $TASK --device cuda:0 --batch_size auto --limit 50

# vllm
export MODEL=YOUR_SAVED_SMOOTHQUANT_MODEL
python benchmarks/benchmark_latency.py --input-len 256 --output-len 256 --model $MODEL
```

namgyu-youn: Hoped so because it works similarly to AWQ, but just tested it with the following code for assurance and got the log message:

```python
import tempfile
import torch
from transformers import AutoModelForCausalLM, AutoTokenizer
from torchao.prototype.smoothquant import SmoothQuantConfig
from torchao.prototype.smoothquant.core import SmoothQuantStep
from torchao.prototype.smoothquant.example import quantize_and_eval
from torchao.quantization.quant_api import Int8DynamicActivationInt8WeightConfig

MODEL_NAME = "microsoft/DialoGPT-small"

# Step 1: Create quantized model
with tempfile.NamedTemporaryFile(suffix='.pt', delete=False) as f:
    model_path = f.name

quantize_and_eval(MODEL_NAME, 0.5, ['PPL'], 256, 5, 'cuda', torch.float32, False, model_path, None)

# Step 2: Test PREPARE_FOR_LOADING
tokenizer = AutoTokenizer.from_pretrained(MODEL_NAME)
tokenizer.pad_token = tokenizer.eos_token

model = AutoModelForCausalLM.from_pretrained(MODEL_NAME, torch_dtype=torch.float32).cuda()
quantize_(model, SmoothQuantConfig(
    base_config=Int8DynamicActivationInt8WeightConfig(),
    step=SmoothQuantStep.PREPARE_FOR_LOADING,
    alpha=0.5,
))

# Test inference
test_input = tokenizer('Hello world', return_tensors='pt').to('cuda')
with torch.no_grad():
    output = model(**test_input)
    generated = model.generate(**test_input, max_length=20, do_sample=False)

print(f"✓ Inference: {output.logits.shape}")
print(f"✓ Generation: {tokenizer.decode(generated[0], skip_special_tokens=True)}")
```

```bash
Loading model on cuda...
Time to load model: 1.86 seconds
running SmoothQuant prepare and calibrate
Repo card metadata block was not found. Setting CardData to empty.
Token indices sequence length is longer than the specified maximum sequence length for this model (1443 > 1024). Running this sequence through the model will result in indexing errors
time for prepare and calibration: 5.20 seconds
running SmoothQuant convert
time for convert: 0.04 seconds
Saving model to /tmp/tmpqeme5s1r.pt
`loss_type=None` was set in the config but it is unrecognised.Using the default loss: `ForCausalLMLoss`.
Setting `pad_token_id` to `eos_token_id`:50256 for open-end generation.
✓ Inference: torch.Size([1, 4, 50257])
✓ Generation: TorchAO TorchAO
```

For sure, we should benchmark them with your suggestion, but I want to carefully suggest dividing its PR.

torchao/prototype/smoothquant/README.md
```diff
@@ -1,65 +1,61 @@
...
+for data in calibration_dataset:
     model(data)
-save_smooth_quant_recipe(model, "./smooth_quant_recipe.json")
-```
-To load a recipe, users should insert observers first. For example,
-```python
-insert_smooth_quant_observer_(model)
-load_smooth_quant_recipe(model, "./smooth_quant_recipe.json")
+
+# Step 3: Convert
```
jerryzh168, line 49: I think ideally we can add a tutorial doc for how to save transformer models for vllm/lm-eval as well: https://github.com/pytorch/ao/blob/083361bc3f7addc505a0f994a923f4ae9f54388e/torchao/prototype/awq/example.py#L223-L265

basically: prepare, convert, and then manually set the config step to "prepare_for_loading"
and then upload the model.

after this the model should be able to be used with vllm and lm-eval.

torchao/prototype/smoothquant/api.py
```diff
@@ -5,186 +5,131 @@
...
+                f"convert: module is not SmoothQuantObservedLinear, skipping: {type(module)}"
+            )
+            return module
+        observed_linear = module
+    else:
+        raise ValueError(f"Unexpected step: {step}")
+
+    # Convert observed linear to quantized linear
     if config.set_inductor_config:
         torchao.quantization.utils.recommended_inductor_config_setter()
```
jerryzh168: this should be removed I think

torchao/prototype/smoothquant/api.py
```diff
@@ -5,186 +5,131 @@
...
-    observed_linear = module
 
+    # Get quantization parameters
+    if all(x is not None for x in (config.smoothing_factor, config.wei_scales)):
```
jerryzh168: when is this branch taken? how do people generate these parameters?

torchao/prototype/smoothquant/core.py
```diff
@@ -3,23 +3,27 @@
...
 
 class SmoothQuantObserver(torch.nn.Module):
     def __init__(
         self,
         weight: torch.Tensor,
         alpha: Optional[float] = 0.5,
-        quant_mode: str = "static",  # or dynamic
         quant_min: Optional[int] = None,
         quant_max: Optional[int] = None,
         eps: Optional[float] = None,
```
jerryzh168: can we remove these args as well, I think these are not needed

torchao/prototype/smoothquant/core.py
```diff
@@ -31,44 +35,34 @@ def __init__(
...
-            eps=eps,
+            MappingType.SYMMETRIC, torch.int8, PerAxis(-1), eps=self.eps
         )
         self.wei_ic_obs = AffineQuantizedMinMaxObserver(
-            MappingType.SYMMETRIC,
-            torch.int8,
-            PerAxis(-1),
-            eps=eps,
+            MappingType.SYMMETRIC, torch.int8, PerAxis(-1), eps=self.eps
         )
```
jerryzh168: are these hardcoded to int8? what if we want to apply them to other types of quantization?

torchao/prototype/smoothquant/core.py
```diff
@@ -78,7 +72,7 @@ def forward(self, input: torch.Tensor):
         return input
 
     def calculate_qparams(self):
-        # 1 Get min/max per IC from observers
+        # Step 1: Get min/max per input channel (IC) from observers
         wei_min_per_ic = self.wei_ic_obs.min_val
         wei_max_per_ic = self.wei_ic_obs.max_val
         act_min_per_ic = self.act_ic_obs.min_val
```
jerryzh168: so we only need min_val/max_val right? might be easier to just do this ourselves instead of relying on AffineQuantizedMinMaxObserver? we can copy over some of the main logic to record min_val/max_val based on granularity of scale (PerAxis) as well

torchao/prototype/smoothquant/example.py
```diff
@@ -4,185 +4,168 @@
...
-):
-    model.eval()
-    tokenizer.pad_token = tokenizer.eos_token
-    tokenizer.padding_side = "right"
-    tokenizer.add_eos_token = False
-
-    print("Loading dataset")
-    t0 = time.time()
+# TODO: Uniform this with torchao/prototype/awq/example.py and expand more tasks
+def benchmark(model, tokenizer, max_seq_length=512, tasks=["PPL"], device="cuda"):
```
jerryzh168, line 24: I was hoping to remove these in the future and just rely on vllm for performance evaluation and lm-eval for model quality evaluation

torchao/prototype/smoothquant/example.py
jerryzh168: OK sounds good to divide the PR

torchao/prototype/smoothquant/example.py
namgyu-youn: Filed an issue at https://github.com/pytorch/ao/issues/2815 for testing more quantization API and benchmarks. And these task sound good to me; let me engage it after this PR.

torchao/prototype/smoothquant/core.py
namgyu-youn: Oh sorry I misunderstood it. What we only need here are `min_val/max_val` same as AWQ. The workflow should be updated based on it.

torchao/prototype/smoothquant/api.py
namgyu-youn: Yes we only use the smooth factor in here. `wei_scales` and `act_scales` should be totally removed.

torchao/prototype/smoothquant/core.py
namgyu-youn: I missed updating it after testing `Int8DynamicActivationInt8WeightConfig`. It should be fixed.

torchao/prototype/smoothquant/core.py
namgyu-youn: yes they can be removed with `AffineQuantizedMinMaxObserver`, thanks.

torchao/prototype/smoothquant/api.py
```diff
@@ -5,186 +5,119 @@
...
+    # Get quantization parameters
+    smoothing_factor = (
+        config.smoothing_factor
+        if config.smoothing_factor is not None
+        else observed_linear.obs.calculate_qparams()
+    )
+    weight = observed_linear.weight * smoothing_factor
+
+    # Create new linear layer
     linear = torch.nn.Linear(
```
jerryzh168: nit: one trick we can have when creating these linear weights is to create them in meta device: https://github.com/vllm-project/vllm/blob/c86af22f31838ee654c856279ac5110ae3fdb2cc/vllm/model_executor/layers/quantization/torchao.py#L159 to save memory I think, similar for awq

torchao/prototype/smoothquant/core.py
```diff
@@ -3,157 +3,98 @@
...
+    CONVERT = "convert"
+    PREPARE_FOR_LOADING = "prepare_for_loading"
 
 
 class SmoothQuantObserver(torch.nn.Module):
     def __init__(
         self,
         weight: torch.Tensor,
+        bias: Optional[torch.Tensor],
+        base_config: AOBaseConfig,
```
jerryzh168: actually seems like `base_config` is not used here?

torchao/prototype/smoothquant/api.py
jerryzh168: please remove this before landing

torchao/prototype/smoothquant/api.py
```diff
@@ -5,186 +5,119 @@
...
         smoothing_factor: The smoothing factor for the layer. Acquired from the layer's observer if None.
-        act_scales: The activation scales for the layer. Acquired from the layer's observer if None.
-        wei_scales: The weight scales for the layer. Acquired from the layer's observer if None.
         set_inductor_config: if True, adjusts `torchinductor` settings to recommended values.
     """
 
+    base_config: AOBaseConfig
+    step: SmoothQuantStep
+    alpha: Optional[float] = 0.5
     smoothing_factor: Optional[torch.Tensor] = None
```
jerryzh168: should this be removed as well? seems like not used?

torchao/prototype/smoothquant/api.py
```diff
@@ -5,186 +5,119 @@
...
-    act_scales: Optional[torch.Tensor] = None
-    wei_scales: Optional[torch.Tensor] = None
     set_inductor_config: bool = True
```
jerryzh168: also this flag, I don't think we need this

torchao/prototype/smoothquant/api.py
```diff
@@ -5,186 +5,119 @@
...
+
+    # Convert observed linear to quantized linear
     if config.set_inductor_config:
         torchao.quantization.utils.recommended_inductor_config_setter()
-    observed_linear = module
 
+    # Get quantization parameters
+    smoothing_factor = (
+        config.smoothing_factor
+        if config.smoothing_factor is not None
```
jerryzh168: probably just remove this arg, I don't see when we'll use it

torchao/prototype/smoothquant/core.py
```diff
@@ -3,157 +3,98 @@
...
 
     def forward(self, input: torch.Tensor):
         input = self.obs(input)
-        output = F.linear(input, self.weight, self.bias)
-        return output
+        return F.linear(input, self.weight, self.bias)
 
     @classmethod
     def from_float(cls, float_linear: torch.nn.Linear, obs: SmoothQuantObserver):
         observed_linear = cls(
```
jerryzh168: also here, can use https://github.com/vllm-project/vllm/blob/c86af22f31838ee654c856279ac5110ae3fdb2cc/vllm/model_executor/layers/quantization/torchao.py#L159 to save memoery

torchao/prototype/smoothquant/README.md
```diff
@@ -68,29 +64,15 @@ Running the example with `torch.compile` on a NVIDIA A10G GPU.
 Perplexity
 | Quant Method | alpha=0.25 | alpha=0.5 | alpha=0.75 | alpha=None* |
 |-|-|-|-|-|
-| Dynamic | 8.1872 | 7.4257 | 7.2518 | 7.5509 |
-| Static | 43.8051 | 11.2984 | 7.5791 | 19.5050 |
+| SmoothQuant | - | - | - | - |
```
jerryzh168: I think we'd need to run some initial testing to make sure the refactor works? otherwise we might be landing non-working code

torchao/prototype/smoothquant/api.py
namgyu-youn: Removing it looks good because it can be computed without initialization, thanks.

torchao/prototype/smoothquant/api.py
namgyu-youn: Thanks for the reminder. I will remove it.

torchao/prototype/smoothquant/core.py
namgyu-youn: Since the observer only computes smoothing factor, `base_config` can be removed, thanks.

torchao/prototype/smoothquant/README.md
namgyu-youn: This is the omitted section and will be updated after the refactor is finished. I will only add int8 dynamic in this PR, but am definitely interested in expanding to vLLM benchmarks with more quantization APIs.

torchao/prototype/smoothquant/core.py
namgyu-youn: Thanks for teaching me meta devices again!

torchao/prototype/smoothquant/api.py
namgyu-youn: Thanks for teaching me the Meta device. LGTM for memory saving when transformations are done before loading the actual data.

torchao/prototype/smoothquant/README.md
jerryzh168: do you plan to update this before this PR is merged or later? I feel we should update this before the PR can be merged to make sure this is working

performance benchmark can come later, but we should have accuracy test to make sure smoothquant implementation is correct I think

namgyu-youn: Sorry I didn't make it clear; **it will be updated in this PR** after refactoring is finished. Actually, I have been checking its performance using `TinyLlama/TinyLlama-1.1B-Chat-v1.0` for each commit. The architecture is quite different, so I didn't mention them, although the accuracy (perplexity) was affordable. Following is one of the experiment results:

image

Similar to AWQ, it will be updated using the `Llama-2-7b-chat-hf` model. Please feel free to direct commit about it because my setup is quite different from your team's setup (1xA100 80GB SXM4 instance).

jerryzh168: OK please request review when you are ready

torchao/prototype/smoothquant/README.md
```diff
@@ -1,98 +1,70 @@
...
-### meta-llama/Meta-Llama-3-8B
-Perplexity
-| Quant Method | alpha=0.25 | alpha=0.5 | alpha=0.75 | alpha=None* |
-|-|-|-|-|-|
-| Dynamic | 21.2475 | 8.8288 | 9.6514 | 8.3574 |
-| Static | 301.7118 | 18.0617 | 10.8343 | 278.9819 |
+## Benchmarks
 
-Note*: Conventional quantization without SmoothQuant
+Evaluation perplexity numbers were calculated using the script in `smoothquant/example.py`. For Llama-2-7b-chat-hf, performance benchmarks were calculated using the `torchao/_models/llama/generate.py` script and run on a 1xA100 48GB PCIe interface.
```
namgyu-youn: @jerryzh168 Since my setup (1xA100 48GB PCIe) is not the same as your team (1xA100 80GB SXM4 instance), the result can be quite different. Please feel free to update it if needed.

jerryzh168: that's fine I feel, we are mainly interested in the perplexity changes with and without AWQ I think, could you show that? you can use https://github.com/pytorch/ao/blob/main/torchao/_models/llama/eval.py to get the perplexity I think, or try to add a non-awq version in example.py

namgyu-youn:
Benchmark Result:

image

image

image

image

torchao/prototype/smoothquant/README.md
namgyu-youn: I have been reviewing https://github.com/pytorch/ao/blob/main/torchao/_models/llama/eval.py, but applying it now to SmoothQuant makes me overwhelmed. Therefore, I built a comparison and result without `lm_eval`. Although the experiment setting is quite different, the result shows there is meaningful speed gain. Could you look into it? Results are attached in https://github.com/pytorch/ao/pull/2728#issuecomment-3224073953.

namgyu-youn: Again, I am engaging in benchmarks  (AWQ vs. SmoothQuant) within vLLM and lm_eval, but have to learn more. The future plan is filed at https://github.com/pytorch/ao/issues/2815, so we can separate PRs; please do not hesitate to assign it to me.

jerryzh168:
perplexity is lower the better I think, it feels like there might be some issues with the implementation of somoothquant?

jerryzh168:
you can also try to follow the awq functionality test: https://github.com/pytorch/ao/blob/23f8a22a8d7d6bfbc2a4ec29d6ca88f1eda5a369/test/prototype/test_awq.py#L64 to make sure smoothquant is implemented correctly

namgyu-youn:
> perplexity is lower the better I think, it feels like there might be some issues with the implementation of somoothquant?

Table 6 shows there is a slight perplexity increase with speedup (latency). Could you elaborate more about the issue?

Also, in this benchmark, I used SeqLen (`max_seq_length`) 512 and calibration limit 128, unlike AWQ because:
- SeqLen: 128, 256, 512, and 1024 were used in the paper (Table 8 and 10)
- Calibration: It isn't specified in the paper, but in my last experience, many researches selected 128 > for stable learning.


> you can also try to follow the awq functionality test:
> 
> https://github.com/pytorch/ao/blob/23f8a22a8d7d6bfbc2a4ec29d6ca88f1eda5a369/test/prototype/test_awq.py#L64
> 
> to make sure smoothquant is implemented correctly

We have a similar function, `test_smoothquant_accuracy` at https://github.com/pytorch/ao/pull/2728/files#diff-ece749169545f50ad6e0a8d09b55d9c68752f6ec9e0accf30c61b8250bdf350cR75. But it doesn't compute error range like AWQ because, in my opinion, the test configuration is not enough for the real error range.

jerryzh168:
> > perplexity is lower the better I think, it feels like there might be some issues with the implementation of somoothquant?
> 
> Table 6 shows there is a slight perplexity increase with speedup (latency). Could you elaborate more about the issue?
> 
> Also, in this benchmark, I used SeqLen (`max_seq_length`) 512 and calibration limit 128, unlike AWQ because:
> 
> * SeqLen: 128, 256, 512, and 1024 were used in the paper (Table 8 and 10)
> * Calibration: It isn't specified in the paper, but in my last experience, many researches selected 128 > for stable learning.
> 
> > you can also try to follow the awq functionality test:
> > https://github.com/pytorch/ao/blob/23f8a22a8d7d6bfbc2a4ec29d6ca88f1eda5a369/test/prototype/test_awq.py#L64
> > 
> > to make sure smoothquant is implemented correctly
> 
> We have a similar function, `test_smoothquant_accuracy` at [#2728 (files)](https://github.com/pytorch/ao/pull/2728/files#diff-ece749169545f50ad6e0a8d09b55d9c68752f6ec9e0accf30c61b8250bdf350cR75). But it doesn't compute error range like AWQ because, in my opinion, the test configuration is not enough for the real error range.

perplexity should be decrease when we apply smoothquant I think, since it's the lower the better

In terms of performance, I think we don't expect large changes there, since it's just applying some additional scaling factor to activation
> We have a similar function, test_smoothquant_accuracy at [#2728 (files)](https://github.com/pytorch/ao/pull/2728/files#diff-ece749169545f50ad6e0a8d09b55d9c68752f6ec9e0accf30c61b8250bdf350cR75). But it doesn't compute error range like AWQ because, in my opinion, the test configuration is not enough for the real error range.

yeah I think we should make sure the unittest is really useful to test whether smoothquant is effective, if it's not useful right now, it would be better to improve it a bit so it's useful. otherwise we don't have visibility to whether we have implemented smoothquant correctly or not

namgyu-youn:
> perplexity should be decrease when we apply smoothquant I think, since it's the lower the better
 
Could you please link reference about this? I know how perplexity (ppl) works, but based on the [paper](https://arxiv.org/pdf/2211.10438), Table 7 (the following) shows there is slight ppl increase with SmoothQuant.

image


> 
> > We have a similar function, test_smoothquant_accuracy at [#2728 (files)](https://github.com/pytorch/ao/pull/2728/files#diff-ece749169545f50ad6e0a8d09b55d9c68752f6ec9e0accf30c61b8250bdf350cR75). But it doesn't compute error range like AWQ because, in my opinion, the test configuration is not enough for the real error range.
> 
> yeah I think we should make sure the unittest is really useful to test whether smoothquant is effective, if it's not useful right now, it would be better to improve it a bit so it's useful. otherwise we don't have visibility to whether we have implemented smoothquant correctly or not

Expanding unittest to compute real error range can be a future plan, but I think experiment results in this PR can substitute it. We can design it using https://github.com/pytorch/ao/blob/main/.github/workflows/1xH100_tests.yml, but this is future plan, not in this PR because experiment results can substitute it.

jerryzh168:
oh sorry, I meant smoothquant should decrease perplexity compared to original dynamic quant, not compare to original high precision model.

so it should be perplexity (bfloat16, float16, float32) < perplexity (smoothquant + int8 dynamic quant) < perplexity (int8 dynamic quant)

wondering if you can show `perplexity (smoothquant + int8 dynamic quant) < perplexity (int8 dynamic quant)` in the testing. this will give us confidence that the implementation is correct

namgyu-youn:
> so it should be perplexity (bfloat16, float16, float32) < perplexity (smoothquant + int8 dynamic quant) < perplexity (int8 dynamic quant)
> 
> wondering if you can show `perplexity (smoothquant + int8 dynamic quant) < perplexity (int8 dynamic quant)` in the testing. this will give us confidence that the implementation is correct

I see, evaluating all the cases seems much better. I will update `example.py` for this, but it can be a little bit heavy for first-time users; maybe we can go with tutorial for this after this PR.

namgyu-youn:
Benchmark result for A4W8-dynamic without SmoothQuant:
screenshot-2025-08-28_11-59-57
screenshot-2025-08-28_12-04-04

jerryzh168 (approved):
Thanks!

torchao/prototype/smoothquant/README.md
```diff
@@ -1,98 +1,86 @@
...
-| Static | 301.7118 | 18.0617 | 10.8343 | 278.9819 |
+### Benchmark Results
 
-Note*: Conventional quantization without SmoothQuant
+| Precision | Quantization | Perplexity | Tokens/sec | PPL Change | Speed Change |
+|-----------|--------------|------------|------------|------------|--------------|
+| float16   | -                  | 6.93       | 625        | -          | -           |
+| bfloat16  | -                  | 6.93       | 667        | -          | -           |
+| bfloat16* | -                  | 6.93       | 27    🐌   | -          | -           |
+| float16   | A4W8-dynamic       | 7.35       | 1,016      | +6.07%     | +39.51%   |
```
jerryzh168: W4A8?

namgyu-youn:
@jerryzh168 hi, I am happy to share perplexity (https://github.com/pytorch/ao/pull/2728#issuecomment-3231759033; origin < SmoothQuant < A4W8) scored **same** as we discussed. 

Could you look into it? It seems that SmoothQuant API truly works! :tada:

torchao/prototype/smoothquant/README.md
namgyu-youn: oh yes I missed so updated docs in https://github.com/pytorch/ao/pull/2728/commits/4c2d55c0999e87165c7af092e3a73bc6eee0de0b.

torchao/prototype/smoothquant/README.md
namgyu-youn: Yeah tutorial is needed because `example.py` in both are quiet heavy. But as I reviewed, `TransformerEvalWrapper` might not work correctly so I will look into "how can we correctly wrap quantized (SmoothQuant/AWQ) models".

Is it okay to go with this first? I think we can go with tutorial after verifying it.

torchao/prototype/smoothquant/README.md
```diff
@@ -1,98 +1,81 @@
...
-Use `--alpha` to specify the alpha parameter. Add `--disable-smooth-quant` to run quantization without SmoothQuant.
+| Precision | Quantization | Perplexity | Tokens/sec | PPL Change | Speed Change |
+|-----------|--------------|------------|------------|------------|--------------|
+| bfloat16  |  -             | 6.93       | 667        |  -         |  -          |
+| bfloat16* |  -             | 6.93       | 27    🐌   |  -         |  -          |
+| bfloat16  | W4A8-dynamic   | 7.35       | 1,967      | +6.07%     | +33.89%     |
+| bfloat16  | W4A8-dynamic** | 7.03       | **1,972**  | **+1.39%** | +33.82%     |
+| float16   |  -             | 6.93       | 625        |  -         |  -          |
+| float16   | W4A8-dynamic   | 7.29       | 523        | +5.21%     | -19.42%     |
+| float16   | W4A8-dynamic** | 6.94       | 516        | **+0.21%** | -21.23%     |
```
namgyu-youn: Although bfloat16 works well, float16 increase latency. Also, `torch.compile` seems not compatiable with SmoothQuant. How about removing `args.precision` and `torch.compile` overally for more brevity? 

But not certain about it because we can investigate to them for better performance (just maybe..).

torchao/prototype/smoothquant/README.md
```diff
@@ -1,98 +1,81 @@
...
+- **Speed Improvement**: Most configurations show 35-40% speed improvement with both W4A8 and SmoothQuant-W4A8
+- **Quality Trade-off**: Slight perplexity increase (~1-1.4%) in most cases
+- **Compilation Impact**: Using `--compile` flag significantly degrades performance (768% slower)
+- **Best Configuration**: `bfloat16` without `--compile` provides optimal balance
 
-**Environment**
-- AWS g5.12xlarge instance
-- torch==2.6.0.dev20241017+cu124
-- python==3.12.6
+> Note: Unlike AWQ, this benchmark isn't computed using the script in `vllm/benchmarks` or `lm_eval`. vLLM benchmark will be introduced in foreseeable future. See https://github.com/pytorch/ao/issues/2815 for more information.
```
namgyu-youn, line 82: Please assign this issue to me. I am looking a way to wrap quantized (AWQ & SmoothQuant) models, one of the milestone for this issue.

test/prototype/test_smoothquant.py
```diff
@@ -13,252 +12,252 @@
...
+
     @classmethod
     def setUpClass(cls):
         """Set up class-level configuration for tests."""
         # This test case will trigger recompilation many times, so set a large cache_size_limit here
         torch._dynamo.config.cache_size_limit = 128
 
-    @unittest.skip("This test is broken on recent PyTorch, TODO(#1639): fix it")
-    @common_utils.parametrize("bias", [True, False])
+    # TODO: Update after #2729 merged
```
namgyu-youn: p.s., @jerryzh168 sorry for interrupting your schedule, but could you look at this PR? CI seems happy, and waiting for your review.

jerryzh168: reviewed, please see comments

torchao/prototype/smoothquant/README.md
jerryzh168: sounds good

torchao/prototype/smoothquant/api.py
jerryzh168: @namgyu-youn L170-173 can be removed I think

torchao/prototype/smoothquant/api.py
jerryzh168: this is still there as well

torchao/prototype/smoothquant/api.py
jerryzh168: seems like not updated?

torchao/prototype/smoothquant/README.md
```diff
@@ -1,65 +1,61 @@
...
+# SmoothQuant quantization
 This is a native PyTorch implementation of the algorithm described in [this paper](https://arxiv.org/abs/2211.10438).
 
 In this implementation, weights are smoothed (equalized) and quantized to int8 during quantization. Activations are smoothed and quantized to int8 at runtime. Quantization is done either dynamically or statically. If activations are dynamically quantized, qparams (i.e., scales) are found at runtime while qparams are found during quantization for static quantization. For dynamic quantization, activations are quantized per token. And for static quantization, activations are quantized per tensor. Generally, dynamic quantization produces better accuracy while static quantization has better latency. In both cases, weights and activations are symmetrically quantized.
 
 ## Quick start
 Run the example code with
 ```bash
-python example.py -m MODLE_ID --device=<cuda or cpu> --quant-mode=<dynamic or static>
+python example.py --repo MODEL_ID --quant smoothquant --device <cuda or cpu>
```
jerryzh168: nit: repo --> `model_id` might be clearer

jerryzh168 (changes requested):
actually, looks like many of the previous comments are not addressed yet, please re-request review after addressing these comments, thanks!

torchao/prototype/smoothquant/README.md
namgyu-youn: Yeah I used `--repo` same as AWQ, but `--model` is a much well used name, thanks.

namgyu-youn: Yeah I used `--repo` same as AWQ, but `--model` is a much well used name, thanks.

torchao/prototype/smoothquant/example.py
```diff
@@ -4,185 +4,276 @@
...
+        .eval()
+        .to(device)
+    )
+    print(f"Time to load model: {time.time() - t0:.02f} seconds")
+
+    # Step 1: Prepare - insert observers
+    print("running SmoothQuant prepare and calibrate")
+    t0 = time.time()
+    quant_config = SmoothQuantConfig(
+        base_config=Int8DynamicActivationInt8WeightConfig(),
```
jerryzh168, line 101: so this is int8 dynamic activation + int8 weight? instead of int8 + int4? did you use a different one when you run the test?

namgyu-youn: Oh I recorded the wrong name.. The experiment result is W8A8, not W4A8. README should be fixed for this.

torchao/prototype/smoothquant/api.py
namgyu-youn: This was removed from https://github.com/pytorch/ao/pull/2728/commits/fc8ae4d83040fa8ccfe6967c415b1be0ac8cdfef#diff-67509204789182caac4e5912b46078e13e62ae5fe7834a576b0bf6abb524eefaL55; please check Ln55.

torchao/prototype/smoothquant/api.py
namgyu-youn: ~It was fixed in https://github.com/pytorch/ao/pull/2728/commits/46d5d317de3253c2957cb76a1ca0e462d63d79c0; please check Ln 119~

It was removed in https://github.com/pytorch/ao/pull/2728/commits/fc8ae4d83040fa8ccfe6967c415b1be0ac8cdfef, please check Ln48

torchao/prototype/smoothquant/api.py
namgyu-youn: `act_scales`, `wei_scales` are deprecated parameter; please check https://github.com/pytorch/ao/pull/2728#discussion_r2308766624 (or https://github.com/pytorch/ao/commit/fc8ae4d83040fa8ccfe6967c415b1be0ac8cdfef - Ln48)

torchao/prototype/smoothquant/README.md
jerryzh168: sure, performance is not the main thing here, we are mainly interested in accuracy

torchao/prototype/smoothquant/api.py
```diff
@@ -5,227 +5,122 @@
...
 
-    qw = to_weight_tensor_with_linear_activation_scale_metadata(qw, factor.to(qw.dtype))
+    # Quantize weights
+    base_config_handler = _QUANTIZE_CONFIG_HANDLER[type(base_config)]
+    dummy_mod = DummyModule(weight)
+    quant_mod = base_config_handler(dummy_mod, base_config)
+    qw = quant_mod.weight
+
+    # Add smoothing factor metadata
+    qw = to_weight_tensor_with_linear_activation_scale_metadata(
```
jerryzh168, line 120: btw, a followup here is to align with awq: https://github.com/pytorch/ao/pull/2753

we can start by making it work for https://github.com/pytorch/ao/blob/main/torchao/quantization/quantize_/workflows/intx/intx_unpacked_to_int8_tensor.py

torchao/prototype/smoothquant/core.py
```diff
@@ -3,160 +3,95 @@
...
-        )
+        with torch.device("meta"):
+            observed_linear = cls(
+                float_linear.in_features,
+                float_linear.out_features,
+                obs,
+                device=float_linear.weight.device,
+                dtype=float_linear.weight.dtype,
+            )
         observed_linear.weight = float_linear.weight
```
jerryzh168, line 98: bias should be assigned as well?

torchao/prototype/smoothquant/example.py
```diff
@@ -4,185 +4,263 @@
...
+        step=SmoothQuantStep.PREPARE,
+        alpha=alpha,
+    )
+    quantize_(model, quant_config)
 
-if __name__ == "__main__":
+    # Step 2: Calibration
+    calibration_data = get_calib_dataset(
```
jerryzh168, line 108: so TransformerEval doesn't work here? ideally we use that in the end since we are using lm-eval to evaluation, and it would be nice if users can just specify a list of tasks and we can do the calibration on them

torchao/prototype/smoothquant/example.py
```diff
@@ -4,185 +4,263 @@
...
+    model = (
+        AutoModelForCausalLM.from_pretrained(model_id, torch_dtype=torch.bfloat16)
+        .eval()
+        .to(device)
+    )
+    base_results = benchmark(
+        model, tokenizer, max_seq_length, tasks=tasks, device=device
+    )
+
+    # Case 2: W4A8-dynamic without SmoothQuant
```
jerryzh168: nit: comments and print strings need to be updated from W4A8 to A8W8

although it might be more useful that we should a A8W4 example (in the future)

torchao/prototype/smoothquant/core.py
```diff
@@ -3,160 +3,95 @@
...
-            float_linear.out_features,
-            float_linear.bias is not None,
-            obs,
-            device=float_linear.weight.device,
-            dtype=float_linear.weight.dtype,
-        )
+        with torch.device("meta"):
+            observed_linear = cls(
+                float_linear.in_features,
+                float_linear.out_features,
```
jerryzh168, line 92: why bias is not mentioned?

torchao/prototype/smoothquant/api.py
```diff
@@ -5,227 +5,122 @@
...
-        qw = to_weight_tensor_with_linear_activation_quantization_metadata(
-            qw, _ActQuantizer(target_dtype).static_quantize, x_scale, x_zero_point
+        raise ValueError(f"Unexpected step: {step}")
+
+    # Compute smoothed weight parameters
+    smoothing_factor = observed_linear.obs.calculate_qparams()
+    weight = observed_linear.weight * smoothing_factor
+
+    # Create new linear layer
+    with torch.device("meta"):
```
jerryzh168, line 103: please make sure bias is assigned for this one as well

torchao/prototype/smoothquant/api.py
```diff
@@ -5,227 +5,122 @@
...
-    observed_linear = module
-
-    linear = torch.nn.Linear(
-        observed_linear.in_features,
-        observed_linear.out_features,
-        observed_linear.bias is not None,
-        device=observed_linear.weight.device,
-        dtype=observed_linear.weight.dtype,
-    )
-    linear.bias = observed_linear.bias
```
jerryzh168, line 195: you can see it's assigned in previously

torchao/prototype/smoothquant/example.py
namgyu-youn: I haven't tried `TransformerEval` here because in my last review, AWQ seems not supporting `TransformerEval` yet.

And what we are looking for is surely **supporting vLLM or lm-eval ecosystem** for users within AWQ, SmoothQuant, etc. Finding correct wraper might be sub-issue for https://github.com/pytorch/ao/issues/2815, but I think we can file it at the PR and link to it directly.

torchao/prototype/smoothquant/example.py
namgyu-youn: Thanks. Personally, W4A8 has higher priority than vLLM benchmarks. But for the promotion, both seems attractive.

torchao/prototype/smoothquant/api.py
namgyu-youn: yes bias should be added, thanks.

torchao/prototype/smoothquant/core.py
namgyu-youn: yes bias should be here for preventing over-fitting.

torchao/prototype/smoothquant/api.py
namgyu-youn: Interesting, I haven't thought about computation cost, but wondered why this function's name is too long. Direct wrapping seems much better, so I will look into it after this PR, thanks.

torchao/prototype/smoothquant/core.py
namgyu-youn: Yes bias should be here same as old version.

torchao/prototype/smoothquant/example.py
jerryzh168: OK sounds good

torchao/prototype/smoothquant/README.md
```diff
@@ -1,98 +1,81 @@
...
-# static quant
-TORCHINDUCTOR_FREEZING=1 python example.py -m <model_id> --device=cuda --quant-mode=static --compile
-```
-Use `--alpha` to specify the alpha parameter. Add `--disable-smooth-quant` to run quantization without SmoothQuant.
+| Precision dtype | Quantization | Perplexity | Tokens/sec | PPL Change | Speed Change |
+|-----------|--------------|------------|------------|------------|--------------|
+| bfloat16  |  -             | 6.93       | 667        |  -         |  -          |
+| bfloat16* |  -             | 6.93       | 27    🐌   |  -         |  -          |
+| bfloat16  | W8A8-dynamic   | 7.35       | 1,967      | +6.07%     | +33.89%     |
+| bfloat16  | W8A8-dynamic** | 7.03       | **1,972**  | **+1.39%** | +33.82%     |
```
jerryzh168: so this is smoothquant? might be better to call out?

torchao/prototype/smoothquant/api.py
```diff
@@ -5,227 +5,123 @@
...
-        set_inductor_config: if True, adjusts `torchinductor` settings to recommended values.
+        base_config: Base quantization configuration that SmoothQuant is applied on top of
+        step (SmoothQuantStep): The step for SmoothQuant process
+            PREPARE: insert SmoothQuant Observers to linear layers
+            CONVERT: convert the observed linear modules to quantized modules
+            PREPARE_FOR_LOADING: convert the floating point model to a dummy smoothquant quantized model, so we can
+            load the quantized weights through copy_ later
+        alpha: The alpha value to determine smoothing factor. Factor = 1 if alpha is None, which means
+            Fall back to conventional quantization if None
+        smoothing_factor: Main concept of SmoothQuant, which is a factor to smooth the quantization.
```
jerryzh168: is this removed?

jerryzh168 (approved):
looks good thanks, please see some final inline comments

torchao/prototype/smoothquant/example.py
namgyu-youn: btw, I just found I thought wrapping quiet complex. Because we can import pre-trained model from HuggingFace (HF) and save to HF keeping its architecture, we can directly use HF format in vLLM (yes vLLM supports HF input format).

Therefore, wrapping is really easy, just use `model.save_pretrained()` instead of `torch.save()`. This should not be sub-issue, and quiet simple. I locally test using it, and it works well. Hope to show you full version after this PR!

torchao/prototype/smoothquant/README.md
namgyu-youn: asterisk (*) means `torch.compile`, and double-asterisk (**) means SmoothQuant+W8A8. Maybe we can update them for better readability.

torchao/prototype/smoothquant/api.py
namgyu-youn: They all exist. Are you asking for `alpha` or `smoothing_factor` annotations removal? Let me know if I am missing something.

jerryzh168: the smoothing_factor argument is removed from the config I think, so we should remove the comment

namgyu-youn: I see, thanks.

test/prototype/test_smoothquant.py
```diff
@@ -13,252 +12,252 @@
...
+        config = SmoothQuantConfig(
+            base_config=base_config, step=SmoothQuantStep.PREPARE, alpha=alpha
+        )
+        quantize_(m2, config)
 
-        # Perform calibration with calibration data
+        # Step 2: Calibration
         for data in calibration_data:
-            m(data)
+            m2(data.squeeze(0).to(input_dtype))
```
jerryzh168: why is this `squeeze(0)` needed?

test/prototype/test_smoothquant.py
```diff
@@ -13,252 +12,252 @@
...
-
-        class SimpleLinear(torch.nn.Module):
-            def __init__(self, bias: bool):
-                super().__init__()
-                self.fc = torch.nn.Linear(32, 32, bias)
-                self.fc.weight.data = torch.randn_like(self.fc.weight.data)
-
-            def forward(self, x):
-                return self.fc(x)
+    def test_smoothquant_accuracy(self, alpha, base_config, device, input_dtype):
```
jerryzh168, line 77: where is the error check in this test?

test/prototype/test_smoothquant.py
```diff
@@ -13,252 +12,252 @@
...
+
+        # CONVERT step - should produce regular Linear with quantized weights
+        config.step = SmoothQuantStep.CONVERT
+        quantize_(m, config)
+
+        # After CONVERT - should be regular Linear again (but quantized)
+        self.assertIsInstance(m.linear1, torch.nn.Linear)
+        self.assertNotIsInstance(m.linear1, SmoothQuantObservedLinear)
+
+    def test_prepare_for_loading(self):
```
jerryzh168: to really test this, I think we need to do the following:

1. one path with `PREPARE_FOR_LOADING` as you have done here, let's say `model1`
2. another path with `PREPARE` + `CONVERT`, save the model (torch.save(model2.state_dict())
3. load the state_dict from file, and `model1.load_state_dict(...)`
4. compare the result of `model1(*example_inputs)` (with loaded weights) and `model2(*example_inputs)`, they should match exactly

jerryzh168: Additionally you can also add one for vllm use pattern, like this: https://github.com/pytorch/ao/blob/568c1932a16ae9f30d48da214a88dc0013e98ed8/test/prototype/test_awq.py#L155, this will make sure that vllm works

torchao/prototype/smoothquant/example.py
jerryzh168: `TransformerEvalWrapper` wraps over lm-eval actually, not sure how vllm is related here actually

test/prototype/test_smoothquant.py
namgyu-youn: This unittest checks: 1. Is `q_out` exist (can be computed)?, 2. Are dimension of `q_out` and `ref_out`same?

We can try expanding it to error range, but not certain as I mentioned in https://github.com/pytorch/ao/pull/2728#issuecomment-3225351336.

test/prototype/test_smoothquant.py
namgyu-youn: It's because `example_inputs()` is updated to use sequence output, same as AWQ:

https://github.com/pytorch/ao/blob/568c1932a16ae9f30d48da214a88dc0013e98ed8/test/prototype/test_awq.py#L28-L36

Also, since we agreed to keep `ToyLinearModel` in `test_awq.py` (in https://github.com/pytorch/ao/pull/2729#discussion_r2311746752), reusing it for SmoothQuant seems to be a choice; done at https://github.com/pytorch/ao/pull/2728/commits/ba58e89f7f0a0ee23731aa94e53bd5503e1f6640.

torchao/prototype/smoothquant/example.py
namgyu-youn: Let me rewrite about this; Maybe there is no need to use `TransformerEvalWrapper` here. For more explanation, workflows can be summarized as:
1. Quantization: Setup -> Calibration -> Scaling -> Quantize -> **Save** to PyTorch checkpoint (`*.pt`) format
2. Benchmark: **Load** quantized PyTorch/HF model -> Benchmark in vLLM -> Visualize

At now, AWQ and SmoothQuant are using `torch.save()` (save to `*.pt`).
https://github.com/pytorch/ao/blob/568c1932a16ae9f30d48da214a88dc0013e98ed8/torchao/prototype/awq/example.py#L269

But for the vLLM benchmark, **we have to use the HF format**. Luckily, saving to HF format is really easy because we are loading a pretrained model from HF to apply AWQ/SmoothQuant. That is, we are already using saved model architecture (configs) from HF. So, without debugging `TransformerEvalWrapper`, we can just replace `torch.save()` with HF commands; please check https://github.com/namgyu-youn/ao/commit/2c30bb78a73b03eac020c022c12a2e65c05b97a7 for more info.

test/prototype/test_smoothquant.py
namgyu-youn: Sounds good, but not certain because:
1. We are using toy (linear) model here, not real use-case (transformer-based) with the following

https://github.com/pytorch/ao/blob/568c1932a16ae9f30d48da214a88dc0013e98ed8/torchao/_models/llama/model.py#L243

3. Even after we update toy model to transformer-based, we might face high resource requirement.
4. What you are looking for (1-4) is pipeline test, not small unit test (too large size).

So my suggestion is, how about keeping (let's check only if each functions are produce-able) them and engage in auto-release like https://github.com/pytorch/ao/pull/2906 ? This seems right place for pipeline test I guess.

test/prototype/test_smoothquant.py
jerryzh168: I think we should include an error testing here, otherwise this is not accuracy/functionality testing

the error testing can be:

1. quantize model with base_config - model1
2. quantize model with SmoothQuantConfig(base_config) - model2
3. compute the sqnr of model1(*example_Inputs) with original_model(*example_inputs), and sqnr of model2(*example_inputs) with original_model(*example_inputs), to make sure second sqnr is higher

test/prototype/test_smoothquant.py
jerryzh168: why ToyLinear model doesn't work? we have awq test here btw: https://github.com/pytorch/ao/blob/568c1932a16ae9f30d48da214a88dc0013e98ed8/test/prototype/test_awq.py#L107

test/prototype/test_smoothquant.py
jerryzh168: I see, it seems like every time `data` is used, it always need a `data.squeeze(0)`? would it make more sense to remove the leading dimension `1` in https://github.com/pytorch/ao/blob/568c1932a16ae9f30d48da214a88dc0013e98ed8/test/prototype/test_awq.py#L32 instead?

test/prototype/test_smoothquant.py
namgyu-youn: oh I meant although linear model works, in my (very naive) guess, it would be not enough for real use-case, because most of the architecture are transformer-based.

test/prototype/test_smoothquant.py
namgyu-youn: Sounds good to me. I will move to `assert` for comparing base quantized with SmoothQuant, thanks.

test/prototype/test_smoothquant.py
namgyu-youn: Seems much better, how about deviding PRs? Because this PR address SmoothQuant API, we can open that small refactoring PR I think.

test/prototype/test_smoothquant.py
jerryzh168: yeah that's fine but we should at least have some test for functionality with ToyLinear I feel

torchao/prototype/smoothquant/example.py
jerryzh168: using `get_calib_dataset` to get calibration dataset is a bit limited isn't it? that's why we want to use `TransformerEvalWrapper` which allows people to use any available tasks in lm-eval

but this is not blocking, feel free have a separate PR to introduce `TransformerEvalWrapper`

test/prototype/test_smoothquant.py
```diff
@@ -13,252 +12,220 @@
...
+            # Skip int4 weight tests for now due to reference implementation mismatch
+            # Int8DynamicActivationInt4WeightConfig(),
+        ],
     )
-    @common_utils.parametrize("input_dtype", [torch.float, torch.bfloat16, torch.half])
-    def test_save_load_recipe(self, alpha, quant_mode, device, input_dtype):
-        """Test save/load recipe functionality."""
+    @common_utils.parametrize("device", ["cpu", "cuda"])
+    @common_utils.parametrize("input_dtype", torch.bfloat16)
+    def test_two_step_quantization(self, alpha, base_config, device, input_dtype):
```
jerryzh168: I think this can be removed if we have `test_smoothquant_accuracy` implemented as suggested, or merge these two

torchao/prototype/smoothquant/example.py
namgyu-youn: Oh sorry, I misunderstood `TransformerEvalWrapper` as a kind of converter at first look. Using independent calibration extractor (`get_calib_dataset`) is definitely not the goal; thanks for correcting it.

test/prototype/test_smoothquant.py
namgyu-youn: Yes we can remove this unit test if we focus on the accuracy comparison (i.e., perplexity: W8A8-dynamic < SmoothQaunt) I feel; thanks for teaching me.

test/prototype/test_smoothquant.py
namgyu-youn: Jerry, I tried some experiments for `assert SmoothQuant < base`, but it seems the error comparison is not producible.

The main reason is that **the linear model is not enough to collect outliers**. In other words, we are losing activation (outlier) distribution, which is the key of AWQ and SmoothQuant. Therefore, in my opinion, using transformer-based (`Transformer`) or something is the right place we can reproduce it. 

https://github.com/pytorch/ao/blob/568c1932a16ae9f30d48da214a88dc0013e98ed8/torchao/_models/llama/model.py#L243

How about adding a TODO for this issue and keeping the unit test? I totally agree with error testing as you mentioned, but it can't be applied using `ToyLinearModel` I think (same issue at https://github.com/pytorch/ao/pull/2729#discussion_r2311780828).

jerryzh168: maybe this: https://github.com/pytorch/ao/blob/85557135c93d3429320a4a360c0ee9cb49f84a00/test/integration/test_integration.py#L208 can you copy this test first, and I think we should follow up with migrating more tests there I feel

namgyu-youn: Sure, please check https://github.com/pytorch/ao/commit/8cd24006df6e3ff08c1a7bfb51ea232a65744d4a

jerryzh168: I feel we should have at least one test to make sure smoothquant is producing reasonable result, you can adapt https://github.com/pytorch/ao/blob/85557135c93d3429320a4a360c0ee9cb49f84a00/test/integration/test_integration.py#L208 for the current implementation so we have some sanity check at least I think, otherwise the accuracy seems to not be tested at all

namgyu-youn: Yes, we are only checking "if error is producable (not None)," not "if error of SmoothQuant is lower than error of original quan". The pasted test is done at https://github.com/pytorch/ao/pull/2728/commits/50dedb89e655211b3bca2705cc4ca5bb1c15b3a4 , and it should be updated I feel. Feel free to ping me if there is any issue for this after this PR.

jerryzh168: > Yes, we are only checking "if error is producable (not None)," not "if error of SmoothQuant is lower than error of original quan".

can you add some sanity check for accuracy like https://github.com/pytorch/ao/blob/85557135c93d3429320a4a360c0ee9cb49f84a00/test/integration/test_integration.py#L208?

> The pasted test is done at 50dedb8 , and it should be updated I feel. Feel free to ping me if there is any issue for this after this PR.

not sure what this means, I didn't see accuracy check in the update

jerryzh168 (changes requested):
please add a test for sanity checking the accuracy / functionality of smoothquant implementation, see comments inline

jerryzh168:
the error seems to be real: https://github.com/pytorch/ao/actions/runs/17455759907/job/49707357714?pr=2728 you can't import from test files, can you define the ToyLinearModel in test_smoothquant itself?

test/integration/test_integration.py
jerryzh168, line 1: can you revert changes to this file? the smoothquant here is referring to a different smoothquant implementation

jerryzh168 (changes requested):
please revert changes to test_integration.py

also can you run the tests locally first?

namgyu-youn:
> also can you run the tests locally first?

Unfortunately, I am unavailable to L40s. Here is the locally tested result in A100 80GB PCIe MIG instance. 

<details>
<summary>Result of integration test after revert</summary>

```bash
$ pytest test/integration --verbose -s
===================================================================================================== warnings summary =====================================================================================================
.venv/lib/python3.10/site-packages/triton/runtime/autotuner.py:97
.venv/lib/python3.10/site-packages/triton/runtime/autotuner.py:97
  /home/elicer/ao/.venv/lib/python3.10/site-packages/triton/runtime/autotuner.py:97: DeprecationWarning: warmup, rep, and use_cuda_graph parameters are deprecated. See https://github.com/triton-lang/triton/pull/4496 for details.
    warnings.warn(("warmup, rep, and use_cuda_graph parameters are deprecated. See "

torchao/utils.py:408
  /home/elicer/ao/torchao/utils.py:408: UserWarning: TORCH_VERSION_AT_LEAST_2_8 is deprecated and will be removed in torchao 0.14.0
    warnings.warn(self.msg)

test/integration/test_integration.py::TestSubclass::test_int8_weight_only_quant_subclass_3_cuda
  /home/elicer/ao/.venv/lib/python3.10/site-packages/torch/_inductor/compile_fx.py:282: UserWarning: TensorFloat32 tensor cores for float32 matrix multiplication available but not enabled. Consider setting `torch.set_float32_matmul_precision('high')` for better performance.
    warnings.warn(

test/integration/test_integration.py: 49 warnings
  /home/elicer/ao/torchao/utils.py:408: UserWarning: TORCH_VERSION_AT_LEAST_2_7 is deprecated and will be removed in torchao 0.14.0
    warnings.warn(self.msg)

test/integration/test_integration.py::SmoothquantIntegrationTest::test_on_dummy_distilbert
  /home/elicer/ao/test/integration/test_integration.py:1429: DeprecationWarning: torch.ao.quantization is deprecated and will be removed in 2.10. 
  For migrations of users: 
  1. Eager mode quantization (torch.ao.quantization.quantize, torch.ao.quantization.quantize_dynamic), please migrate to use torchao eager mode quantize_ API instead 
  2. FX graph mode quantization (torch.ao.quantization.quantize_fx.prepare_fx,torch.ao.quantization.quantize_fx.convert_fx, please migrate to use torchao pt2e quantization API instead (prepare_pt2e, convert_pt2e) 
  3. pt2e quantization has been migrated to torchao (https://github.com/pytorch/ao/tree/main/torchao/quantization/pt2e) 
  see https://github.com/pytorch/ao/issues/2259 for more details
    model_copy2 = torch.ao.quantization.quantize_dynamic(

test/integration/test_integration.py::SmoothquantIntegrationTest::test_on_dummy_distilbert
  /home/elicer/ao/.venv/lib/python3.10/site-packages/torch/ao/quantization/quantize.py:566: DeprecationWarning: torch.ao.quantization is deprecated and will be removed in 2.10. 
  For migrations of users: 
  1. Eager mode quantization (torch.ao.quantization.quantize, torch.ao.quantization.quantize_dynamic), please migrate to use torchao eager mode quantize_ API instead 
  2. FX graph mode quantization (torch.ao.quantization.quantize_fx.prepare_fx,torch.ao.quantization.quantize_fx.convert_fx, please migrate to use torchao pt2e quantization API instead (prepare_pt2e, convert_pt2e) 
  3. pt2e quantization has been migrated to torchao (https://github.com/pytorch/ao/tree/main/torchao/quantization/pt2e) 
  see https://github.com/pytorch/ao/issues/2259 for more details
    convert(model, mapping, inplace=True)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
================================================================================ 152 passed, 207 skipped, 55 warnings in 124.51s (0:02:04) =================================================================================
```
</details>


<details>
<summary>Result of SmoothQuant test</summary>

```bash
$ pytest test/prototype/test_smoothquant.py --verbose -s
=================================================================================================== test session starts ====================================================================================================
platform linux -- Python 3.10.15, pytest-8.4.2, pluggy-1.6.0 -- /home/elicer/ao/.venv/bin/python3
cachedir: .pytest_cache
hypothesis profile 'default'
rootdir: /home/elicer/ao
plugins: hypothesis-6.138.14
collecting ... TMA benchmarks will be running without grid constant TMA descriptor.
collected 6 items                                                                                                                                                                                                          

test/prototype/test_smoothquant.py::TestSmoothQuant::test_observer_insertion_base_config0 PASSED
test/prototype/test_smoothquant.py::TestSmoothQuant::test_prepare_for_loading_base_config0 PASSED
test/prototype/test_smoothquant.py::TestSmoothQuant::test_smoothquant_accuracy_alpha_0_5_base_config0_device_cpu_bfloat16 convert: module is not SmoothQuantObservedLinear, skipping: <class 'torch.nn.modules.linear.Linear
'>                                                                                                                                                                                                                          PASSED
test/prototype/test_smoothquant.py::TestSmoothQuant::test_smoothquant_accuracy_alpha_0_5_base_config0_device_cuda_bfloat16 convert: module is not SmoothQuantObservedLinear, skipping: <class 'torch.nn.modules.linear.Linea
r'>                                                                                                                                                                                                                         PASSED
test/prototype/test_smoothquant.py::TestSmoothQuant::test_smoothquant_accuracy_alpha_0_75_base_config0_device_cpu_bfloat16 convert: module is not SmoothQuantObservedLinear, skipping: <class 'torch.nn.modules.linear.Linea
r'>                                                                                                                                                                                                                         PASSED
test/prototype/test_smoothquant.py::TestSmoothQuant::test_smoothquant_accuracy_alpha_0_75_base_config0_device_cuda_bfloat16 convert: module is not SmoothQuantObservedLinear, skipping: <class 'torch.nn.modules.linear.Line
ar'>                                                                                                                                                                                                                        PASSED

===================================================================================================== warnings summary =====================================================================================================
.venv/lib/python3.10/site-packages/triton/runtime/autotuner.py:97
.venv/lib/python3.10/site-packages/triton/runtime/autotuner.py:97
  /home/elicer/ao/.venv/lib/python3.10/site-packages/triton/runtime/autotuner.py:97: DeprecationWarning: warmup, rep, and use_cuda_graph parameters are deprecated. See https://github.com/triton-lang/triton/pull/4496 for 
details.                                                                                                                                                                                                                        warnings.warn(("warmup, rep, and use_cuda_graph parameters are deprecated. See "

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
============================================================================================== 6 passed, 2 warnings in 4.98s ===============================================================================================
```
</details>

jerryzh168:
please skip the failed test when there is no cuda like this: https://github.com/pytorch/ao/blob/d35c2ce93efb1db059b3250a593f5b64332a2617/test/quantization/quantize_/workflows/int4/test_int4_tensor.py#L28

namgyu-youn:
> please skip the failed test when there is no cuda like this:
> 
> https://github.com/pytorch/ao/blob/d35c2ce93efb1db059b3250a593f5b64332a2617/test/quantization/quantize_/workflows/int4/test_int4_tensor.py#L28

Done, could you look into it?

Xia-Weiwen:
This PR is titled "Make SmoothQuant more General" however it removes static quant support and prevents UT running in CPU-only environments, which actually makes SmoothQuant more specific.

namgyu-youn:
> This PR is titled "Make SmoothQuant more General" however it removes static quant support and prevents UT running in CPU-only environments, which actually makes SmoothQuant more specific.

@Xia-Weiwen We decided to split the PR to support more quantization APIs, as discussed in https://github.com/pytorch/ao/pull/2728#discussion_r2284819151. In fact, what "general" refers to here is the new SmoothQuant API structure (`config -> quantize (convert)`). Please check the following unit test and reproduce it for more quantization APIs:

https://github.com/pytorch/ao/blob/0d3217d07ec112bf92fb9d7383938d884b411502/test/prototype/test_smoothquant.py#L127-L164

PR #2734, opened 2025-08-11
[moe training] work around wrap_triton bug by using normal custom ops instead for fp8 rowwise kernels 

danielvegamyhre:
Stacked PRs:
 * #2769
 * #2767
 * #2765
 * #2762
 * #2756
 * #2749
 * __->__#2734
 * #2733

--- --- ---

## Summary
- I discovered this bug (https://github.com/pytorch/pytorch/issues/160333) with wrap_triton + tensor subclasses, and as a fix we can simply use traditional custom ops, rather than wrap_triton.
- As of 8/12 a short-term fix has landed in PT core (https://github.com/pytorch/pytorch/pull/160341), but this would require users to use a very recent pytorch nightly, whereas custom ops are well supported in 2.7+ release, and the work to use custom ops is already done in this PR, so I don't see a compelling reason to migrate back. Open to feedback on this though.

### DIFF
--- a/benchmarks/prototype/moe_training/benchmark_kernels.py
+++ b/benchmarks/prototype/moe_training/benchmark_kernels.py
@@ -15,8 +15,8 @@ from tqdm import tqdm
 from triton.testing import do_bench
 
 from torchao.prototype.moe_training.kernels.jagged_float8_scales import (
-    triton_fp8_col_major_jagged_colwise_scales,
-    triton_fp8_row_major_jagged_rowwise_scales,
+    triton_fp8_per_group_colwise_scales,
+    triton_fp8_per_group_rowwise_scales,
 )
 from torchao.prototype.moe_training.utils import (
     torch_to_float8_per_group_colwise,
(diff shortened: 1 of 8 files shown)


vkuzo:
prototype feature requiring PT nightly sounds good to me tbh.  Are there specific users you are looking to unblock who can't use nightly? if not, I'd say then keep this nightly-only until we move out of prototype.

vkuzo (approved):
accepting to unblock

danielvegamyhre:
> prototype feature requiring PT nightly sounds good to me tbh. Are there specific users you are looking to unblock who can't use nightly? if not, I'd say then keep this nightly-only until we move out of prototype.

Makes sense, I will make a note to revisit this. We could migrate back to wrap_triton, it just seems like all downsides (more limited pytorch version compatibility) without any clear upsides (functionality is the same as normal custom ops). I guess one upside is we could remove the "register_fake"? I also just have more confidence in normal custom ops maturity and stability as a feature, versus wrap_triton.

vkuzo:
> this would require users to use a very recent pytorch nightly, whereas custom ops are well supported in 2.7+ release

I was just saying prototype features should require nightly, it's usually not worth it to also support older versions as that adds complexity.  No opinion on triton_op vs custom_op.

Issue #2271, opened 2025-05-29
QAT range learning tracker


andrewor14:
This issue trackers follow-up feature requests regarding QAT range learning after the initial PR: https://github.com/pytorch/ao/pull/2033

- [ ] Do not learn zero points for symmetric quantization
- [ ] Add option to learn actual ranges instead of scales and zero points directly
- [ ] For weight-only quantization, allow users to initialize the new parameters in `__init__` instead
- [ ] Allow users to save learned scales and zero points and reload them in a new training job

PR #2739, opened 2025-08-11
don't learn zero points for symmetric quantization

liangel-02:
summary: 

part of https://github.com/pytorch/ao/issues/2271
rather than learning both scales and zero points, zero points are always zero for symmetric quantization. both are learnable parameters still for asymmetric quantization.

### DIFF
--- a/test/quantization/test_qat.py
+++ b/test/quantization/test_qat.py
@@ -21,7 +21,12 @@ from torchao import quantize_
 from torchao.float8.config import ScalingGranularity
 from torchao.float8.float8_scaling_utils import hp_tensor_to_float8_dynamic
 from torchao.float8.float8_training_tensor import LinearMMConfig
-from torchao.quantization.granularity import PerAxis, PerGroup, PerRow, PerToken
+from torchao.quantization.granularity import (
+    PerAxis,
+    PerGroup,
+    PerRow,
+    PerToken,
+)
 from torchao.quantization.linear_quant_modules import (
     _replace_linear_8da4w,
     _replace_linear_int4,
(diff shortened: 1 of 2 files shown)


test/quantization/test_qat.py
```diff
@@ -1685,6 +1685,7 @@ def test_qat_range_learning(self):
             range_learning=True,
             scale_precision=torch.float32,
             zero_point_precision=torch.float32,
+            is_symmetric=True,
```
andrewor14: can you test both True and False here by adding parameterization to this test?

Here's an example: https://github.com/pytorch/ao/blob/853f87d1bd56d8d6c980a518cf9fa08fbb20135c/test/quantization/test_qat.py#L1516

test/quantization/test_qat.py
```diff
@@ -1704,10 +1705,20 @@ def test_qat_range_learning(self):
...
+
+        qparams = m._get_all_weight_qparams()
+        for i in range(0, len(qparams), 2):
+            scale = qparams[i]
+            zero_point = qparams[i + 1]
+            self.assertIsInstance(scale, torch.nn.Parameter)
+            self.assertTrue(scale.requires_grad)
+            self.assertTrue(scale in params)
+            if config.is_symmetric:
+                self.assertFalse(isinstance(zero_point, torch.nn.Parameter))
```
andrewor14, line 1730: also assert zero point is always zero?

andrewor14 (approved):
Looks great, thank you!

test/quantization/test_qat.py
```diff
@@ -1704,10 +1705,20 @@ def test_qat_range_learning(self):
         # All scales and zero points should be in `m.parameters()`
         initialize_fake_quantizers(m, example_inputs)
         params = set(m.parameters())
-        for t in m._get_all_weight_qparams():
-            self.assertIsInstance(t, torch.nn.Parameter)
-            self.assertTrue(t.requires_grad)
-            self.assertTrue(t in params)
+
+        qparams = m._get_all_weight_qparams()
+        for i in range(0, len(qparams), 2):
```
andrewor14: actually this is a bit error prone. How about we separate this into `_get_all_weight_scales` and `_get_all_weight_zero_points`, and just point `_get_all_weight_qparams` to the concatenation of these two lists? Then here we can test scales and zero points separately

Issue #2713, opened 2025-08-07
[fp8 blockwise training] try using torch._scaled_mm instead of Triton kernels for fp8 gemms


danielvegamyhre:
We have an initial prototype of DeepSeekV3 style fp8 blockwise training done [here](https://github.com/pytorch/ao/blob/main/torchao/prototype/blockwise_fp8_training/linear.py). Numerics are accurate but performance has not been optimized yet.

Initial tests with a local torchtitan integration on my H100 devgpu show the blockwise GEMM kernels are slower than expected. NCU analysis shows uncoalesced global accesses causing major slowdowns, but rather than optimize these kernels, it's probably a better idea to use `torch._scaled_mm` instead, which recently added support for DSV3 style fp8 GEMMs using a CUTLASS kernel which is likely much more performant than the Triton kernels. This will also be more consistent with our other float8 tensorwise and rowwise training recipes, which use torch._scaled_mm.

We should do the following:

1. Add benchmarking script(s) that compares runtime of:
    - Performance of [blockwise_fp8_gemm_1x128_128x128](https://github.com/pytorch/ao/blob/143c3a60451727f9fba56289b6fa74cfdb04b440/torchao/prototype/blockwise_fp8_training/kernels.py#L106) vs torch._scaled_mm
    - Performance of [blockwise_fp8_gemm_1x128_128x1](https://github.com/pytorch/ao/blob/143c3a60451727f9fba56289b6fa74cfdb04b440/torchao/prototype/blockwise_fp8_training/kernels.py#L214C5-L214C35) vs torch._scaled_mm
    - (see [here](https://github.com/pytorch/ao/blob/143c3a60451727f9fba56289b6fa74cfdb04b440/torchao/prototype/blockwise_fp8_training/linear.py#L26) for context on how/where these gemms are used for context)
    - Here is an [example](https://github.com/pytorch/ao/blob/main/benchmarks/float8/bench_grouped_mm.py) benchmark script for something similar that can be used as a starting point.
2. If microbenchmarks show torch._scaled_mm is faster, update the blockwise fp8 linear to use this gemm. 

Note torch._scaled_mm has some slightly different stride/mem layout requirements for the inputs. You will see this in the error message that it throws if you try to directly swap it out with the triton gemms.

danielvegamyhre:
This is done, see blockwise training prototype

PR #2740, opened 2025-08-12
Add benchmarking and testing for blockwise fp8 GEMM using Triton and …

roman-y-wu:
issue: #2713 
- Introduced a new benchmarking script to compare performance between Triton kernels and torch._scaled_mm for blockwise fp8 GEMM operations.
- Added tests for correctness and performance of scaled_mm implementations, including various matrix sizes commonly used in LLMs.
- Implemented scaled_mm kernels to support blockwise fp8 GEMM operations, preserving scaling precision.
- Enhanced the Float8BlockwiseLinear class to support both Triton and scaled_mm backends.
- Included error handling and edge case tests for the new implementations.

### DIFF
--- a/benchmarks/prototype/blockwise_fp8_training/benchmark_triton_vs_scaled_mm.py
+++ b/benchmarks/prototype/blockwise_fp8_training/benchmark_triton_vs_scaled_mm.py
@@ -5,12 +5,12 @@
 # LICENSE file in the root directory of this source tree.
 
 """
-Microbenchmark script to compare Triton kernels vs torch._scaled_mm
+Microbenchmark script to compare Triton kernels vs torch._scaled_mm native blockwise scaling
 for blockwise fp8 GEMM operations.
 
 This provides a proper 1:1 comparison between the Triton blockwise implementation
-and the torch._scaled_mm block-by-block approach, both preserving blockwise 
-scaling precision without approximations.
+and the torch._scaled_mm native blockwise scaling with CUDA 12.9+, as recommended 
+by danielvegamyhre to avoid uncoalesced memory access issues in Triton kernels.
 """
 
 import torch
@@ -51,9 +51,10 @@ def blockwise_fp8_scaled_mm_1x128_128x128_reference(
     b_scale: torch.Tensor,
 ) -> torch.Tensor:
     """
-    Reference implementation using the improved torch._scaled_mm blockwise approach.
-    This provides a proper 1:1 comparison with the Triton kernel by using the
-    same block-by-block processing to preserve blockwise scaling precision.
+    Reference implementation using native torch._scaled_mm with blockwise scaling.
+    This uses the CUDA 12.9+ native blockwise scaling support to provide optimal
+    performance through direct CUTLASS kernel usage, avoiding the uncoalesced
+    memory access issues present in Triton kernels.
     """
     from torchao.prototype.blockwise_fp8_training.scaled_mm_kernels import (
         blockwise_fp8_gemm_scaled_mm_1x128_128x128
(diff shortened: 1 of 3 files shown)


benchmarks/prototype/blockwise_fp8_training/benchmark_triton_vs_scaled_mm.py
```diff
@@ -0,0 +1,248 @@
...
+    """
+    Reference implementation using torch._scaled_mm for blockwise fp8 GEMM.
+    This is a simplified version - the actual implementation needs to handle
+    blockwise scaling properly.
+    """
+    a_mm, a_scale_recip, b_mm, b_scale_recip = prepare_blockwise_scaled_mm_tensors(
+        a_fp8, a_scale, b_fp8, b_scale
+    )
+    
+    # For now, use tensorwise scaling as a baseline comparison
```
danielvegamyhre: I'm confused, why are we comparing against tensorwise here? Tensorwise will have substantially different perf than blockwise, we need to do a 1:1 comparison here

torchao/prototype/blockwise_fp8_training/scaled_mm_kernels.py
```diff
@@ -0,0 +1,253 @@
...
+    a_s: torch.Tensor,  # (M, K // block_size) reciprocals of scales
+    b: torch.Tensor,  # (K, N) in fp8, column-major
+    b_s: torch.Tensor,  # (K // block_size, N // block_size) reciprocals of scales
+    block_size: int = 128,
+) -> torch.Tensor:
+    """
+    Blockwise fp8 GEMM using torch._scaled_mm instead of Triton kernel.
+
+    This is a simplified implementation that approximates blockwise scaling
+    using row-wise and column-wise scaling supported by torch._scaled_mm.
```
danielvegamyhre: torch._scaled_mm supports DSV3 style groupwise and blockwise scaling now, using CUDA 12.9. We need to use that, we can't approximate

danielvegamyhre (commented):
Thanks for taking a crack at this! I left a couple comments, there are some serious issues with this approach I think, we can't use other scaling types and other GEMMs as an approximation, we need to perform these benchmarks against torch._scaled_mm + CUDA 12.9 that dispatches to the actual cutlass kernel in PT core for groupwise+blockwise scaling.

torchao/prototype/blockwise_fp8_training/scaled_mm_kernels.py
roman-y-wu: Thank you for the comments, I made the edit that the problem should be fixed.

danielvegamyhre: @romanwu10 awesome, I made a note to take a look at this as soon as I can. thanks!

danielvegamyhre: Can you share performance benchmark results comparing the triton gemm and torch._scaled_mm? Include in the PR description if you can?

torchao/prototype/blockwise_fp8_training/scaled_mm_kernels.py
```diff
@@ -0,0 +1,217 @@
...
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+Implementation of blockwise fp8 GEMM operations using torch._scaled_mm
+as an alternative to custom Triton kernels.
+
+This implementation uses block-by-block processing with torch._scaled_mm to maintain
+blockwise scaling precision, providing accurate results comparable to the Triton kernels.
+While torch._scaled_mm doesn't natively support arbitrary blockwise scaling, the
```
danielvegamyhre: `torch._scaled_mm` does natively support blockwise scaling, see [here](https://github.com/pytorch/pytorch/blob/371eacb2ae4ecdabc52ea4634ed21558df2f3bab/aten/src/ATen/cuda/CUDABlas.h#L144-L145). You just need to upgrade to CUDA 12.9 to use it :)

That is the goal of this issue, to support torch._scaled_mm with native blockwise scaled GEMMs (not emulated ones like this).

danielvegamyhre: This unit test should help you figure out how to use the torch._scaled_mm native blockwise scaled gemm: https://github.com/pytorch/pytorch/blob/371eacb2ae4ecdabc52ea4634ed21558df2f3bab/test/test_matmul_cuda.py#L1373

roman-y-wu: Updated. Thank you so much!

PR #2742, opened 2025-08-12
Introduce IntxOpaqueTensor to replace PackedInt8DynamicActivationIntxWeightLayout in AQT

metascroy:
This adds IntxOpaqueTensor to replace the AQT tensor with PackedInt8DynamicActivationIntxWeightLayout since AQT will be removed.

The test plan are the new unit tests.


torchao/quantization/quant_api.py
```diff
@@ -546,6 +549,11 @@ def quantize_(
         )
 
 
+def _int8_asymm_per_token_quant_hp(x: torch.Tensor) -> torch.Tensor:
```
jerryzh168: nit: fake_quant sounds more appropriate I think

torchao/quantization/quant_api.py
```diff
@@ -739,6 +747,11 @@ class Int8DynamicActivationIntxWeightConfig(AOBaseConfig):
     weight_scale_dtype: Optional[torch.dtype] = None
     act_mapping_type: MappingType = MappingType.ASYMMETRIC
     layout: Layout = QDQLayout()
+    packing_format: PackingFormat = PackingFormat.UNPACKED_TO_INT8
+
+    # Used with PackingFormat.TILED
+    compute_target: Optional[ComputeTarget] = None
```
jerryzh168: can this be merged with KernelPreference?

torchao/quantization/quantize_/common/packing_format.py
```diff
@@ -40,3 +40,8 @@ class PackingFormat(str, Enum):
     Unpacked means the subbyte quantized data is stored as int8
     """
     UNPACKED_TO_INT8 = "unpacked_to_int8"
+
+    """
+    Tiled means the tiling used by CPU kernels
+    """
+    TILED = "tiled"
```
jerryzh168: TILE_PACKED? or update the tensor name to IntxTiledTensor?

torchao/quantization/quantize_/workflows/intx/intx_tile_packed_tensor.py
```diff
@@ -0,0 +1,345 @@
...
+        bit_width: the bit width for quantization (can be 1 - 8)
+        block_size: the block size for quantization, representing the granularity, for example groupwise quantization will have block_size (1, group_size)
+        shape: the shape of the original Tensor
+        dtype: dtype for activations/outputs
+        packed_weights_has_zeros: whether zeros are present in packed_weights
+        packed_weights_has_bias: whether bias is present in packed_weights
+        compute_target: the compute target for the packed data.  Compute targets may pack the data differently.  See ComputeTarget enum for details.
+    """
+
+    tensor_data_names = ["packed_weights"]
```
jerryzh168: nit: qdata

torchao/quantization/quant_api.py
```diff
@@ -786,30 +799,71 @@ def __post_init__(self):
...
+            assert compute_target is not None, (
+                "Must specify a compute target for PackingFormat.TILED"
+            )
+            new_weight = IntxTilePackedTensor.from_intx_unpacked_tensor(
+                new_weight, bias=new_bias, compute_target=compute_target
+            )
+            new_bias = None  # bias is packed with weights
+        else:
+            assert packing_format == PackingFormat.UNPACKED_TO_INT8
+            new_weight = to_linear_activation_quantized(
```
jerryzh168: btw, we are planning to move away from `to_linear_activation_quantized` as well, to reduce the abstractions, we are implementing dynamic activation in the tensor itself, see https://github.com/pytorch/ao/blob/72b35bf5b0c2b07a8f668b0078d1f919b43ca598/torchao/quantization/quantize_/workflows/float8/float8_tensor.py#L244-L247 for example. also api: https://docs.pytorch.org/ao/main/quantization_overview.html#dynamic-activation-and-weight-quantization

this can be a separate PR though

torchao/quantization/quant_api.py
```diff
@@ -1970,6 +2036,31 @@ def _uintx_weight_only_transform(
     return module
 
 
+def _adjust_scale_dtype_in_intx_unpacked_tensor(
+    intx_unpacked_tensor: IntxUnpackedTensor,
+    hp_tensor: torch.Tensor,
+    scale_dtype: torch.dtype,
+) -> None:
+    """
+    Adjusts the scale_dtype on IntxUnpackedTensor.
```
jerryzh168: this seems complicated, any way to get rid of this?

torchao/experimental/tests/test_int8_dynamic_activation_intx_weight.py
jerryzh168, line 1: nit: update the test file name, also why is this not moved to the test/quantization/quantize_/... folder?

torchao/quantization/quantize_/workflows/intx/intx_tile_packed_tensor.py
metascroy: But it contains scales/bias too, so is qdata the best name?

In the unpacked tensor, qdata refers to just the quantized values

torchao/experimental/tests/test_int8_dynamic_activation_intx_weight.py
metascroy: I can move the test and update the name

torchao/quantization/quant_api.py
metascroy: I really don't like this, but it is to keep numeric backward compatibility in test cases when the scale_dtype is set to something different that input type (which I think will not be a common case).

I could remove this and deprecate scale_dtype from the config in a follow-up diff if it doesn't break any tests.  But I'd rather that diff be small and isolated so I can run it against tests in fbcode.

torchao/quantization/quant_api.py
metascroy: Ok, let me change that then

torchao/quantization/quant_api.py
metascroy: It's similar, but not exactly the same as kernel preference.

jerryzh168: oh OK, I remembered that we have discussed about this

torchao/quantization/quantize_/workflows/intx/intx_tile_packed_tensor.py
jerryzh168: does it always contain scale/bias? if so we can do `qdata_and_params` maybe

torchao/quantization/quant_api.py
jerryzh168: sounds good, yeah will be good to remove, it's fine to do in a future PR

torchao/quantization/quantize_/workflows/intx/intx_tile_packed_tensor.py
metascroy: I think packed_weights is better because it is produced by a pack_weight function.

torchao/quantization/quantize_/workflows/intx/intx_unpacked_tensor.py
```diff
@@ -25,13 +27,22 @@
...
     "IntxUnpackedTensor",
+    "ActivationQuantization",
 ]
 
 aten = torch.ops.aten
 
 _FLOAT_TYPES: List[torch.dtype] = [torch.float16, torch.bfloat16, torch.float32]
 
 
+class ActivationQuantization(enum.Enum):
```
jerryzh168: why are we doing this? the recommended way is https://github.com/pytorch/ao/blob/af2cf1e3e619cc95e095c57bb517eaad2da5be97/torchao/quantization/quantize_/workflows/float8/float8_tensor.py#L243-L247 or just inline the quantization code in linear

torchao/quantization/quantize_/workflows/intx/intx_tile_packed_tensor.py
```diff
@@ -0,0 +1,350 @@
...
+__all__ = [
+    "IntxTilePackedTensor",
+]
+
+aten = torch.ops.aten
+
+
+class ComputeTarget(enum.Enum):
+    """
+    This packs the tensor for PyTorch CPU kernels in ATen.
```
jerryzh168: might be good to add some description on how this differs from KernelPreference

torchao/quantization/quantize_/workflows/intx/intx_unpacked_tensor.py
jerryzh168, line 1: also why is this change here? should this be a separate PR?

torchao/quantization/quantize_/workflows/intx/intx_unpacked_tensor.py
metascroy: The quantization code is inlined in linear in lines 233-249.

The enum was to support applying activation quant (ActivationQuantization. DYNAMIC_INT8_ASYMMETRIC_PER_TOKEN) vs. weight only (None).  I made it an enum so it could be extended.

I can use _choose_quant_func_and_quantize_tensor instead, but out of curiosity, why is that is the design to have_choose_quant_func_and_quantize_tensor in a common directory?  It seems its implementation would just have many if/else if based on various tensor subclasses?

torchao/quantization/quantize_/workflows/intx/intx_unpacked_tensor.py
metascroy: This PR is about updating int8_dynamic_activation_intx_weight to version 2.  Most of the work to do that is creating v2 of the tile_packed tensor.  But it also required updating the unpacked tensor to support dynamic activation quant.

I can move the  unpacked to a separate PR, though.

torchao/quantization/quantize_/workflows/intx/intx_unpacked_tensor.py
jerryzh168: the reason for `_choose_quant_func_and_quantize_tensor` having if/else was for readability and removing indirections, alternative is to have a config I think

but this is not always required, you can also call the activation quantization in the linear op directly as well, without using this function, especially when you only have a single possible type of activation quantization

metascroy:
@jerryzh168 I've updated this PR to use the new opaque tensor

torchao/quantization/quant_api.py
```diff
@@ -835,6 +840,16 @@ def _int8_dynamic_activation_intx_weight_quantize_tensor(weight, bias, config):
 
         new_bias = bias
 
+        # Create packed tensor
+        if packing_format == PackingFormat.OPAQUE:
+            assert compute_target is not None, (
+                "Must specify a compute target for PackingFormat.TILE_PACKED"
```
jerryzh168: nit: update comment

metascroy:
> lg, is this the last change before we can bump the version?

should be.  Just need to do the packing format refactor you mentioned

PR #2743, opened 2025-08-12
Allow per-group quantizers in QuantOptimizer, fix state_dict

lisjin:
Changes to make torchao's version of PARQ compatible with current QAT pipeline:
1. Bug fix: When using FSDP 2, the QuantOptimizer state cannot be loaded correctly from a checkpoint because it has an extra "qat_state" key. Fix this by removing qat_state entirely and have the user set optimizer.num_steps manually after loading checkpoint.
2. New feature: Allow param_groups to be quantized using different bit-widths and/or quantization methods. If the user passes a "quant_cls" and optional "quant_kwargs" keys in a param_group, QuantOptimizer will override the default self.quantizer object with one initialized from these values.
3. New feature: Make PARQ compatible with new QAT API by relaxing the constraints in QATConfig. Before, it required that base_config or weight_config must be provided during the prepare step. Now, it accepts when only activation_config is provided. Now we can migrate from deprecated`*IntXQuantizationAwareTrainingConfig` classes.


torchao/quantization/qat/api.py
```diff
@@ -214,15 +212,21 @@ def _qat_config_transform(
...
-        else:
-            # Unrelated module, ignore
-            return module
-        return _QUANTIZE_CONFIG_HANDLER[type(base_config)](module, base_config)
+
+        return (
+            module
+            if base_config is None or is_fakequant
+            else _QUANTIZE_CONFIG_HANDLER[type(base_config)](module, base_config)
+        )
```
lisjin: @andrewor14 I relaxed the constraints in the convert step so that `base_config=None` is valid. Is this okay?

torchao/quantization/qat/api.py
```diff
@@ -148,18 +148,16 @@ def __post_init__(self):
...
+                "At least one of `base_config`, `activation_config` or `weight_config` must be specified"
             )
         if self.base_config is not None and self.activation_config is not None:
             raise ValueError(
                 "Cannot specify both `base_config` and `activation_config`"
             )
         if self.base_config is not None and self.weight_config is not None:
             raise ValueError("Cannot specify both `base_config` and `weight_config`")
-        if self.base_config is None and self.step == "convert":
-            raise ValueError("`base_config` must be specified in the convert step")
```
andrewor14: ah I'm removing this here: https://github.com/pytorch/ao/pull/2694

torchao/quantization/qat/api.py
andrewor14: Actually I already did it here: https://github.com/pytorch/ao/pull/2694. You should be able to revert these changes after I land it

lisjin: Thanks! I left a comment on that PR since I run into an error about empty base_config and weight_config after rebasing.

andrewor14: I see, please feel free to remove or relax that error in this PR. That PR doesn't account for activation only QAT yet

torchao/quantization/quant_api.py
```diff
@@ -1312,6 +1313,31 @@ def _float8_cutlass_quant_sparse(
     )
 
 
+@dataclass
+class Int8DynActOnlyConfig(AOBaseConfig):
```
andrewor14: Any chance we can keep this under prototype in PARQ? I can't think of any use cases outside of PARQ at the moment

andrewor14: Also nit on the name, maybe `Int8DynamicActivationOnlyConfig` is better (not a big fan of "DynAct" even though we still have a lot of legacy code using that naming)

torchao/quantization/__init__.py
```diff
@@ -144,6 +145,7 @@
     "Int8DynamicActivationIntxWeightConfig",
     "Int4WeightOnlyConfig",
     "Float8DynamicActivationInt4WeightConfig",
+    "Int8DynActOnlyConfig",
```
andrewor14: I wouldn't expose this config under this namespace

torchao/quantization/quant_api.py
lisjin: Sure, I've just fixed both issues

andrewor14:
Please also update the PR description to reflect the latest design.

PR #2746, opened 2025-08-12
Rename `Float8ActivationInt4WeightConfig` to `Float8DynamicActivationInt4WeightConfig`

jerryzh168:
Rename `Float8ActivationInt4WeightConfig` to `Float8DynamicActivationInt4WeightConfig`

Summary:
This was a mistake, we need to align the name with other dynamic quant configs

Test Plan:
CI

Reviewers:

Subscribers:

Tasks:

Tags:


jerryzh168:
technically breaking BC for previous PR, but there is no stable release that has this API and we don't expect to have users now for this one, so marking this as not user facing

andrewor14 (approved):

PR #2748, opened 2025-08-12
Fix setup develop

metascroy:
This enables you to run the following on macOS:

```
python setup.py develop
```

Python extensions always look for so files with specific names.

### DIFF
--- a/torchao/experimental/op_lib.py
+++ b/torchao/experimental/op_lib.py
@@ -22,14 +22,16 @@ potential_paths = [
 
 
 def find_and_load_libtorchao_ops(potential_paths):
+    # import torchao._experimental_aten_ops
+
     for lib_path in potential_paths:
-        libs = list(lib_path.glob("libtorchao_ops_aten.*"))
+        libs = list(lib_path.glob("_experimental_aten_ops.*"))
 
         if not libs:
             continue
 
         assert len(libs) == 1, (
-            f"Expected to find one libtorchao_ops_aten.* library at {lib_path}, but found {len(libs)}"
+            f"Expected to find one _experimental_aten_ops.* library at {lib_path}, but found {len(libs)}"
         )
 
         target_lib = libs[0]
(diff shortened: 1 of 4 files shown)


torchao/experimental/tests/test_embedding_xbit_quantizer.py
```diff
@@ -183,10 +183,9 @@ def test_shared_embedding(self):
         self.assertTrue(torch.allclose(result, exported_result))
 
         # Check the shared_embedding and linear ops use the same lifted weight
-        weight = "b_getattr_l__fn_____0___unembedding_packed_weights"
         expected_lines = [
-            f"torch.ops.torchao._shared_embedding_4bit.default({weight}, 4096, 131, 4096, reshape)",
-            f"torch.ops.torchao._linear_8bit_act_4bit_weight.default(linear, {weight}, 4096, 131, 4096)",
+            "torch.ops.torchao._shared_embedding_4bit.default",
```
metascroy, line 187: Removing weight because its name changes over time and it makes CI flakey

swolchok (commented):
> Python extensions always look for so files with specific names.

I was skeptical of this, but sure enough .so is the more popular extension in my site-packages directory. No complaints from my end but would like to give @drisspg a chance to review.

torchao/experimental/op_lib.py
```diff
@@ -22,14 +22,16 @@
 
 
 def find_and_load_libtorchao_ops(potential_paths):
+    # import torchao._experimental_aten_ops
```
drisspg: nit: maybe a doc string?

drisspg (approved):
I think this is fine, is there any doc on libs are searched and this name schema you mention?

metascroy:
> I think this is fine, is there any doc on libs are searched and this name schema you mention?

Added  docstring to op load function, but this is not user-facing.  It is called inside torchao's __init__.py automatically.

PR #2750, opened 2025-08-12
Revert "Remove calls to contiguous in the implementation of Float8Tensor"

jerryzh168:
Reverts pytorch/ao#2747


jerryzh168:
will submit a forward fix, it's simple fix to test

PR #2753, opened 2025-08-12
Update AWQ implementation to not use extra wrapper tensor subclass

jerryzh168:
Summary:
We want to remove the extra wrapper tensor subclass `to_weight_tensor_with_linear_activation_scale_metadata`, althopugh it is composable with different tensor subclasses, it complicates the flow and reduces locality of code (compute logic spread between the LinearActivationScale wrapper and the real tensor subclasses)

Instead we want to implement activation scaling in the tensor itself, we added a `SupportsActivationPreScaling` protocol that can be implemented by the tensor subclass, that needs to have `act_pre_scale` defined

* we'll use `tensor.act_pre_scale` to set the pre scale tensor
* in linear we can get the activation_scale and do `input_tensor = input_tensor * weight_tensor.act_pre_scale` before calling gemm kernels

This change does not break BC.

Test Plan:
python test/prototype/test_awq.py
python test/quantization/quantize_/workflows/int4/test_int4_tensor.py

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/prototype/test_awq.py
+++ b/test/prototype/test_awq.py
@@ -14,7 +14,7 @@ from torch.testing._internal.common_utils import (
 )
 
 from torchao.prototype.awq import AWQConfig, AWQStep
-from torchao.quantization import FbgemmConfig, Int4WeightOnlyConfig, quantize_
+from torchao.quantization import Int4WeightOnlyConfig, quantize_
 from torchao.utils import _is_fbgemm_genai_gpu_available
 
 
@@ -73,13 +73,7 @@ class TestAWQ(TestCase):
         m = ToyLinearModel(l1, l2, l3).eval().to(original_dtype).to(device)
 
         # baseline quantization
-        base_config = FbgemmConfig(
-            input_dtype=torch.bfloat16,
-            weight_dtype=torch.int4,
-            output_dtype=torch.bfloat16,
-            block_size=[1, group_size],
-            preshuffle=False,
-        )
+        base_config = Int4WeightOnlyConfig(group_size=group_size, version=2)
         m_baseline = copy.deepcopy(m)
         quantize_(m_baseline, base_config)
 
@@ -129,13 +123,7 @@ class TestAWQ(TestCase):
         calibration_data = dataset[:n_calibration_examples]
 
         # calibrate
-        base_config = FbgemmConfig(
-            input_dtype=torch.bfloat16,
-            weight_dtype=torch.int4,
-            output_dtype=torch.bfloat16,
-            block_size=[1, group_size],
-            preshuffle=False,
-        )
+        base_config = Int4WeightOnlyConfig(group_size=group_size, version=2)
         quant_config = AWQConfig(base_config, step=AWQStep.PREPARE)
         quantize_(m, quant_config)
 
@@ -189,13 +177,7 @@ class TestAWQ(TestCase):
         calibration_data = dataset[:n_calibration_examples]
 
         # calibrate
-        base_config = FbgemmConfig(
-            input_dtype=torch.bfloat16,
-            weight_dtype=torch.int4,
-            output_dtype=torch.bfloat16,
-            block_size=[1, group_size],
-            preshuffle=False,
-        )
+        base_config = Int4WeightOnlyConfig(group_size=group_size, version=2)
         quant_config = AWQConfig(base_config, step=AWQStep.PREPARE)
         quantize_(m, quant_config)
 
--- a/torchao/prototype/awq/api.py
+++ b/torchao/prototype/awq/api.py
@@ -10,10 +10,10 @@ from dataclasses import dataclass
 import torch
 
 from torchao.core.config import AOBaseConfig
-from torchao.quantization import to_weight_tensor_with_linear_activation_scale_metadata
 from torchao.quantization.quant_api import (
     _linear_extra_repr,
 )
+from torchao.quantization.quantize_.common import SupportsActivationPreScaling
 from torchao.quantization.transform_module import (
     _QUANTIZE_CONFIG_HANDLER,
     register_quantize_module_handler,
@@ -105,7 +105,12 @@ def _awq_transform(
     dummy_mod = DummyModule(observed_linear.weight * equalization_scale)
     quant_mod = base_config_handler(dummy_mod, config.base_config)
     qw = quant_mod.weight
-    qw = to_weight_tensor_with_linear_activation_scale_metadata(qw, equalization_scale)
+    assert isinstance(qw, SupportsActivationPreScaling), (
+        "weight must support activation scaling through implementing `SupportsActivationPreScaling`"
+    )
+    # since we want to do `act` * `act_pre_scale` during runtime for speed, we'll save the
+    # reciprocal of the `equalization_scale`
+    qw.act_pre_scale = 1.0 / equalization_scale
 
     linear = torch.nn.Linear(
         observed_linear.in_features,
--- a/torchao/quantization/quantize_/common/protocol.py
+++ b/torchao/quantization/quantize_/common/protocol.py
@@ -1,0 +1,22 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Protocols for some functionalities in tensor subclasses"""
+
+from typing import Optional, Protocol, runtime_checkable
+
+import torch
+
+
+@runtime_checkable
+class SupportsActivationPreScaling(Protocol):
+    """Protocol for activation scale that should be multiplied with activation before quantization,
+    or before we use activation in matrix multiplications, used for algorithms like AWQ
+
+    A class that have `act_pre_scale: Optional[torch.Tensor]` attribute implements the Protocol
+    """
+
+    act_pre_scale: Optional[torch.Tensor]
--- a/torchao/quantization/quantize_/workflows/int4/int4_tensor.py
+++ b/torchao/quantization/quantize_/workflows/int4/int4_tensor.py
@@ -5,7 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 
-from typing import List
+from typing import List, Optional
 
 import torch
 from torch.utils._python_dispatch import return_and_correct_aliasing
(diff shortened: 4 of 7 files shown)


torchao/quantization/quantize_/workflows/int4/int4_tensor.py
```diff
@@ -43,23 +44,25 @@ class Int4Tensor(TorchAOBaseTensor):
...
+    def __init__(self, qdata, scale, zero_point, activation_scale, block_size, shape):
         self.qdata = qdata
         self.scale = scale
         self.zero_point = zero_point
+        self.activation_scale = activation_scale
         self.block_size = block_size
 
     def _quantization_type(self):
-        return f"shape={self.shape}, block_size={self.block_size}, device={self.device}"
+        return f"shape={self.shape}, has_activation_scale={self.activation_scale is not None}, block_size={self.block_size}, device={self.device}"
```
vkuzo: nit: most tensors will have this `None`, so I would only add it to the string representation if it's not `None` to keep things short for the common case

torchao/quantization/quantize_/workflows/int4/int4_tensor.py
```diff
@@ -113,12 +117,17 @@ def _(func, types, args, kwargs):
...
+    assert isinstance(weight_tensor, Int4Tensor)
+
     assert weight_tensor.qdata.is_contiguous(), "Expected qdata to be contiguous"
     assert weight_tensor.scale.is_contiguous(), "Expected scale to be contiguous"
     assert weight_tensor.zero_point.is_contiguous(), (
         "Expected zero_point to be contiguous"
     )
 
+    if weight_tensor.activation_scale is not None:
+        input_tensor = input_tensor / weight_tensor.activation_scale
```
vkuzo: I think multiplying would be faster, can we change the scale definition to do that?

torchao/quantization/quantize_/workflows/int4/int4_tensor.py
```diff
@@ -43,23 +44,25 @@ class Int4Tensor(TorchAOBaseTensor):
     """
 
     tensor_data_names = ["qdata", "scale", "zero_point"]
+    optional_tensor_data_names = ["activation_scale"]
     tensor_attribute_names = ["block_size", "shape"]
 
-    def __new__(cls, qdata, scale, zero_point, block_size, shape):
+    def __new__(cls, qdata, scale, zero_point, activation_scale, block_size, shape):
```
vkuzo: I feel like an optional argument like this should be last, after non-optional ones like block_size and shape

torchao/quantization/quantize_/workflows/int4/int4_tensor.py
vkuzo: also, how about `input_scale`, to standardize with https://docs.pytorch.org/docs/stable/generated/torch.nn.functional.linear.html

vkuzo: actually, maybe `act_scale` or `act_prescale` to standardize with `Float8Tensor.act_quant_kwargs`

if the activation is quantized, do we need to combine the scales to just do the multiplication once? this may also affect the design

test/prototype/test_awq.py
```diff
@@ -136,13 +130,7 @@ def test_awq_loading(self):
         calibration_data = dataset[:n_calibration_examples]
 
         # calibrate
-        base_config = FbgemmConfig(
```
andrewor14, line 132: should we delete this public config before people start using it? (in a separate PR)

torchao/quantization/quantize_/common/mixin.py
```diff
@@ -0,0 +1,20 @@
...
+    """Mixin for activation scale, using for algorithms like AWQ
+    If a class inherits from this mixin, we'll have `set_activation_scale`
+    defined on it and the class must have "activation_scale" as an optional tensor data attribute
+    """
+
+    def set_activation_scale(self, scale):
+        assert hasattr(self, "activation_scale"), (
+            f"tensor {type(self)} does not have attribute `activation_scale`"
+        )
+        self.activation_scale = scale
```
andrewor14: should we also add this as an instance variable, like in L15
```
activation_scale: Optional[torch.Tensor]
```

torchao/prototype/awq/api.py
```diff
@@ -105,7 +105,10 @@ def _awq_transform(
     dummy_mod = DummyModule(observed_linear.weight * equalization_scale)
     quant_mod = base_config_handler(dummy_mod, config.base_config)
     qw = quant_mod.weight
-    qw = to_weight_tensor_with_linear_activation_scale_metadata(qw, equalization_scale)
```
andrewor14, line 108: seems like we should deprecate `WeightTensorWithLinearActivationScaleMetadata` in favor of the mixin? Can we file an issue for that?

torchao/quantization/quantize_/workflows/int4/int4_tensor.py
jerryzh168: makes sense although currently this order is mandated by the order of tensor and non-tensor attribute right now: https://github.com/pytorch/ao/blob/a1a9632b9e3aaaefe1299fbb067268dbb7501ab1/torchao/utils.py#L727-L731

currently all tensor attributes comes before non-tensor attributes, is it better to have:
required_tensor_attrs, required_attrs, optional_tensor_attrs, optional_attrs for every tensor?

vkuzo: got it

is this kind of change bc breaking? if yes, is there an opportunity to make all of these keyword-only arguments to remove from bc?

torchao/quantization/quantize_/workflows/int4/int4_tensor.py
jerryzh168: sure, I can use `act_scale`

> if the activation is quantized, do we need to combine the scales to just do the multiplication once? this may also affect the design

it doesn't work if activation is already quantized, we expect this to be applied before static or dynamic quantization

test/prototype/test_awq.py
jerryzh168: yeah, will do

torchao/quantization/quantize_/common/mixin.py
jerryzh168: we want to use TorchAOBaseTensor so have to define this in `optional_tensor_data_names`

torchao/prototype/awq/api.py
jerryzh168: yes, we'll do this along with the migration or when a new tensor is added I think, no need to migrate the old implementations now

torchao/quantization/quantize_/workflows/int4/int4_tensor.py
jerryzh168: if we have a stable release of these, then yes, although right now we don't have these in any of the official stable release yet

some of the data args are required for the construction to make sense I think, do you mean making all attributes keyword-only arg?

drisspg:
Looks like we are switching to more of a protocol style, personally I think that 

``` Python
from typing import Protocol, runtime_checkable

@runtime_checkable
class SupportsActivationScaling(Protocol):
    """A protocol for tensor-like objects that support activation scaling."""
    
    activation_scale: torch.Tensor | None
  
    def set_activation_scale(self, scale: torch.Tensor) -> None:
    	...


# int4
class Int4Tensor(TorchAOBaseTensor):
    # ... existing implementation ...
    
    # Just add the required attribute and method
    activation_scale: torch.Tensor | None = None

    def set_activation_scale(self, scale: torch.Tensor) -> None:
        self.activation_scale = scale
...
        

  def _awq_transform(...):
    # ...
    assert isinstance(qw, SupportsActivationScaling), (
        "weight must support activation scaling by conforming to the SupportsActivationScaling protocol"
    )

   ```
   
   Is nicer since we don't have to do multiple inheritance, I just really dont like multiple inheritance

**Comment by jerryzh168** [comment]
> Looks like we are switching to more of a protocol style, personally I think that
> 
> ```python
> from typing import Protocol, runtime_checkable
> 
> @runtime_checkable
> class SupportsActivationScaling(Protocol):
>     """A protocol for tensor-like objects that support activation scaling."""
>     
>     activation_scale: torch.Tensor | None
>   
>     def set_activation_scale(self, scale: torch.Tensor) -> None:
>     	...
> 
> 
> # int4
> class Int4Tensor(TorchAOBaseTensor):
>     # ... existing implementation ...
>     
>     # Just add the required attribute and method
>     activation_scale: torch.Tensor | None = None
> 
>     def set_activation_scale(self, scale: torch.Tensor) -> None:
>         self.activation_scale = scale
> ...
>         
> 
>   def _awq_transform(...):
>     # ...
>     assert isinstance(qw, SupportsActivationScaling), (
>         "weight must support activation scaling by conforming to the SupportsActivationScaling protocol"
>     )
> ```
> 
> Is nicer since we don't have to do multiple inheritance, I just really dont like multiple inheritance

OK, will try thanks

File: torchao/quantization/quantize_/workflows/int4/int4_tensor.py [hunk 7]
```diff
@@ -36,30 +36,54 @@ class Int4Tensor(TorchAOBaseTensor):
                dtype is the same as the original Tensor dtype
         zero_point: (K/group_size, N) for 2D Tensor, (B, K/group_size, N) for 3D Tensor, where B is batch size,
                dtype is the same as the original Tensor dtype
+        act_scale (Optional[Tensor]): Optional per row scale for activation Tensor, if present,
```
vkuzo: nit: remove "per row", it should work with any scale where the shapes are broadcastable to the original tensor

torchao/quantization/quantize_/workflows/int4/int4_tensor.py
```diff
@@ -36,30 +36,54 @@ class Int4Tensor(TorchAOBaseTensor):
...
+               we'll multiply activation Tensor with act_scale before applying dynamic
```
vkuzo: how does this compose with dynamic quant?  Is there a faster path where we only scale once, instead of scaling by `act_scale`, then finding max, then scaling again by the inner scale?

torchao/quantization/quantize_/workflows/int4/int4_tensor.py
vkuzo: I think the variable name should make it clear that this scale is not related to the scale which happens inside of activation quantization.

jerryzh168: OK sg

torchao/quantization/quantize_/workflows/int4/int4_tensor.py
jerryzh168: we will first do act_scale before dynamic quantization, and then do the dynamic quantization. we don't have the faster path right now, not sure if this is needed or whether this optimization will actually yield speedup yet

torchao/quantization/quantize_/common/protocol.py
```diff
@@ -0,0 +1,22 @@
...
+
+"""Protocols for some functionalities in tensor subclasses"""
+
+from typing import Optional, Protocol, runtime_checkable
+
+import torch
+
+
+@runtime_checkable
+class SupportsActivationScaling(Protocol):
```
vkuzo: nit: `Prescaling`, to match `act_pre_scale`

torchao/quantization/quantize_/workflows/int4/int4_tensor.py
```diff
@@ -113,12 +140,17 @@ def _(func, types, args, kwargs):
...
+    assert isinstance(weight_tensor, Int4Tensor)
+
     assert weight_tensor.qdata.is_contiguous(), "Expected qdata to be contiguous"
     assert weight_tensor.scale.is_contiguous(), "Expected scale to be contiguous"
     assert weight_tensor.zero_point.is_contiguous(), (
         "Expected zero_point to be contiguous"
     )
 
+    if weight_tensor.act_pre_scale is not None:
+        input_tensor = input_tensor * weight_tensor.act_pre_scale
```
vkuzo, line 152: lg, I think we should also test this from the int4 test case file, since this is not specific to AWQ.

vkuzo (approved):

PR #2770, opened 2025-08-14
Fix missing QuantOptimizer methods

lisjin:
This is a patch to yesterday's #2743.
- Returned the overrides to state_dict and load_state_dict in QuantOptimizer, which were mistakenly removed
- Replaced the "quant_cls" and "quant_kwargs" keys in param_group to "quantizer" so that the user can instantiate the class outside of the optimizer. Otherwise, we can't use JSON to serialize nonstandard types like torch.dtype.

### DIFF
--- a/torchao/prototype/parq/quant/uniform_torchao.py
+++ b/torchao/prototype/parq/quant/uniform_torchao.py
@@ -142,7 +142,7 @@ class UnifTorchaoQuantizer(Quantizer):
 
 
 class StretchedUnifTorchaoQuantizer(UnifTorchaoQuantizer):
-    def __init__(self, b: int, int_shift: float = 0.5) -> None:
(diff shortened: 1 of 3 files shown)


torchao/prototype/parq/quant/uniform_torchao.py
```diff
@@ -142,7 +142,7 @@ def quantize(
 
 
 class StretchedUnifTorchaoQuantizer(UnifTorchaoQuantizer):
-    def __init__(self, b: int, int_shift: float = 0.5) -> None:
+    def __init__(self, b: int, int_shift: float = 0.5, **kwargs) -> None:
```
jerryzh168, line 145: what is the kwargs for?

lisjin: It lets us customize the parent class defaults, like mapping_type or target_dtype. Thanks for looking this over, by the way

PR #2773, opened 2025-08-15
mx_formats: make emulated tests pass on H100, and add to CI

vkuzo:
Summary:

As we work toward moving `mx_formats` out of prototype, we will need CI
coverage.  We are unlikely to get NVIDIA B200 machines in CI any time
soon, so we'll have to rely on emulation for now.

The current PR runs our existing mx tests on an H100 and fixes all
encountered issues.  These were broken because we don't have these tests
in our H100 CI yet.

Finally, adds the MX tests to CI:
1. most tests are added to the 1xH100 workflow
2. the dtensor tests are added to the 4xH100 workflow

Test Plan:

```bash
./test/prototype/mx_formats/test_mx_dtensor.sh
```

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo:
* __->__ #2773

PR #2777, opened 2025-08-15
Update mx_formats README.md

vkuzo:
add e2e torchtitan benchmarks on LLaMa 3 8B for mxfp8 training

### DIFF
--- a/torchao/prototype/mx_formats/README.md
+++ b/torchao/prototype/mx_formats/README.md
@@ -7,14 +7,34 @@ in native PyTorch.  We are currently in prototype and are actively working on op
 
 | workflow | emulation | performance | accuracy |
 | --- | --- | --- | --- |
-| training with mxfp8 | ✅ | 🚧 [active development](https://github.com/pytorch/ao/issues/1768) | ✅ |
-| inference (weight-only) with mxfp8, mxfp6, mxfp4 | ✅ | 🔲 | 🔲 |
-
-We plan to add the following features in the near future:
-* other inference workflows such as dynamic quantization
-* a unified training to inference workflow
+| training with mxfp8 | ✅ | ✅ | ✅ |
+| inference with mxfp8, mxfp6, mxfp4 | ✅ | 🔲 | 🔲 |
 
 ℹ️ <em>See the [feature tracker](https://github.com/pytorch/ao/issues/556) and the [performance tracker](https://github.com/pytorch/ao/issues/1768) for upcoming features.</em>
+
+## Training e2e benchmarks on NVIDIA B200
+
+- Single-node training on 8xB100 GPUs, batch size 1, sequence length 8192, steps 100, `torch.compile`, FSDP2, per-op SAC
+- pytorch version: `2.9.0.dev20250815+cu128`, torchao version: `0.13.0+gite4e681be6`, torchtitan commit: `6fc499f6f5b32151a799188be2208cfb09faed30`
+
+| Model         | Scaling                            | Peak Memory (GB)  | Median tokens/second | Speedup over baseline
+| ------------- | ---------------------------------- | ------------------| -------------------- | ---------------------
+| Llama3-8b     |  none (bfloat16)                   | 33.71             |  8307.5              | -
+| Llama3-8b     |  float8 tensorwise (f8 all-gather) | 33.38             |  10417.0             | 25.4%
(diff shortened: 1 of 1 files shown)


torchao/prototype/mx_formats/README.md
```diff
@@ -7,15 +7,35 @@ in native PyTorch.  We are currently in prototype and are actively working on op
...
-* other inference workflows such as dynamic quantization
-* a unified training to inference workflow
+| training with mxfp8 | ✅ | ✅ | ✅ |
+| inference with mxfp8, mxfp6, mxfp4 | ✅ | 🔲 | 🔲 |
 
 ℹ️ <em>See the [feature tracker](https://github.com/pytorch/ao/issues/556) and the [performance tracker](https://github.com/pytorch/ao/issues/1768) for upcoming features.</em>
 
+## Training e2e benchmarks on NVIDIA B200
+
+- Single-node training on 8xB100 GPUs, batch size 1, sequence length 8192, steps 100, `torch.compile`, FSDP2, per-op SAC
```
drisspg: Can you specify that this is a power throttled variant

vkuzo: should also say B200 instead of B100, will fix

torchao/prototype/mx_formats/README.md
```diff
@@ -7,15 +7,35 @@ in native PyTorch.  We are currently in prototype and are actively working on op
...
+- pytorch version: `2.9.0.dev20250815+cu128`, torchao version: `0.13.0+gite4e681be6`, torchtitan commit: `6fc499f6f5b32151a799188be2208cfb09faed30`
+
+| Model         | Scaling                            | Peak Memory (GB)  | Median tokens/second | Speedup over baseline
+| ------------- | ---------------------------------- | ------------------| -------------------- | ---------------------
+| Llama3-8b     |  none (bfloat16)                   | 33.71             |  8307.5              | -
+| Llama3-8b     |  float8 tensorwise (f8 all-gather) | 33.38             |  10417.0             | 25.4%
+| Llama3-8b     |  mxfp8_cublas                      | 33.88             |  9969.0              | 20.0%
+| Llama3-8b     |  mxfp8_cublas_rceil                | 33.88             |  9642.0              | 16.1%
```
danielvegamyhre, line 25: this is odd, rceil uses the hardware acclerated fp32 -> e8m0  casting instruction, it should be faster than floor, and when i did benchmarking it was faster than floor. any idea what could be going on here?

torchao/prototype/mx_formats/README.md
```diff
@@ -7,15 +7,35 @@ in native PyTorch.  We are currently in prototype and are actively working on op
...
 
+## Training e2e benchmarks on NVIDIA B200
+
+- Single-node training on 8xB100 GPUs, batch size 1, sequence length 8192, steps 100, `torch.compile`, FSDP2, per-op SAC
+- pytorch version: `2.9.0.dev20250815+cu128`, torchao version: `0.13.0+gite4e681be6`, torchtitan commit: `6fc499f6f5b32151a799188be2208cfb09faed30`
+
+| Model         | Scaling                            | Peak Memory (GB)  | Median tokens/second | Speedup over baseline
+| ------------- | ---------------------------------- | ------------------| -------------------- | ---------------------
+| Llama3-8b     |  none (bfloat16)                   | 33.71             |  8307.5              | -
+| Llama3-8b     |  float8 tensorwise (f8 all-gather) | 33.38             |  10417.0             | 25.4%
```
danielvegamyhre, line 23: ah, when i benchmarked fp8 tensorwise vs mxfp8 and found they were roughly the same throughput, i didn't use fp8 all-gather, i wonder if that explains this difference

torchao/prototype/mx_formats/README.md
danielvegamyhre: Are we not using torch.compile here? When I ran benchmarks during mxfp8 dim1 cast cuda kernel development i was getting ~13.5k TPS on FSDP2 only 8xb200.. can double check the exact config

torchao/prototype/mx_formats/README.md
vkuzo: I think it only uses the accelerated instruction in the dim1 kernel.  Once we make it also use that instruction in dim0, it should beat floor.

PR #2785, opened 2025-08-17
integrate torch._scaled_mm into Float8BlockwiseLinear and add bench script

danielvegamyhre:
Stacked PRs:
 * #2829
 * #2827
 * #2826
 * __->__#2785

--- --- ---

integrate torch._scaled_mm into Float8BlockwiseLinear and add bench script

## Summary
- Add `use_triton` flag to Float8BlockwiseLinear` that defaults to False. When True, use triton for gemms. When False, use torch._scaled_mm for gemms.
- Compute scales in column major format, for compatibility with torch._scaled_mm.
- Add some autotuner configs for IO bounds dynamic quantization kernels.
- Rename triton kernels to be prefixed with "triton_" for clarity
- Write out reciprocal scales directly from triton quant kernels, rather than writing regular scales then doing 1.0/scales to pass in reciprocals to GEMM for rescaling outputs. This is necessary because in some cases the division op (1.0/scales) changes the memory layout from col major to row major, which triggers assertions on mem layout

## Benchmarks

Benchmarking an eager linear forward + backward pass with llama4 shapes, using torch._scaled_mm improves perf by ~2x but still slower than bf16. Next step is to look at the trace to see which quantization kernels are slowing things down.

```
    M     N     K  out_dtype         bf16_mm_linear_us    fp8_triton_linear_us    fp8_scaled_mm_linear_us
-----  ----  ----  --------------  -------------------  ----------------------  -------------------------
16640  5120  8192  torch.bfloat16              5349.79                 12643.5                    7746.98
16640  8192  5120  torch.bfloat16              6100.62                 12912.6                    7783.31
```


drisspg:
Do we actually need these triton kernels? It seems like scaled_mm is pretty universally, is there a valid case for the small shapes you mentioned previously  

I guess the other valid argument is that some people aren't using new enough triton to have this fp8 mm support

danielvegamyhre:
> Do we actually need these triton kernels? It seems like scaled_mm is pretty universally, is there a valid case for the small shapes you mentioned previously
> 
> I guess the other valid argument is that some people aren't using new enough triton to have this fp8 mm support

No, scaled_mm is definitely better, I just wanted to keep the triton gemms so as a side-project for learning I can try optimizing them to match scaled_mm using additional features like TMA, etc.

torchao/prototype/moe_training/scaled_grouped_mm.py
```diff
@@ -48,7 +48,7 @@ def _scaled_grouped_mm(
     """
     # TODO: Remove logging once prototype is more mature. This is currently very useful for development and debugging.
     if scaling_type == MoEScalingType.FP8_ROWWISE:
-        # print("Using fp8 rowwise scaled_grouped_mm")
+        print("Using fp8 rowwise scaled_grouped_mm")
```
drisspg: !

danielvegamyhre: Yeah this shouldn't be commented out, this was mistakenly left in a previous PR. So fixing it here.

drisspg: we shouldnt be printing though.. maybe logging but forsure not printing

danielvegamyhre: Agreed, I had this as logging originally but changed to print when I noticed the logs didn't show up in torchtitan training runs when I was trying to get ScaledGroupedMMTensor to work e2e, and wanted to focus on the task at hand rather than debug whatever torchtitan is doing with the logging module.

I switched back to logging statements in this PR: https://github.com/pytorch/ao/pull/2835

danielvegamyhre: @drisspg I landed https://github.com/pytorch/ao/pull/2835 and rebased on top of main, so the print statements are logging now.

danielvegamyhre:
autoquant test failure unrelated, landing

Issue #2722, opened 2025-08-08
`torch_version_at_least` semantics are incorrect


andrewor14:
We have the following test cases right now:

https://github.com/pytorch/ao/blob/6cfa47705f60ea614695b52b4b120ac5fd84d1cb/test/test_utils.py#L16-L21

But this is wrong. The right order should be:
```
2.4.0 (stable) < 2.5.0.dev / 2.5.0.git < 2.5.0 (stable)
```
So actually `2.5.0.dev` and `2.5.0.git` are not "at least" 2.5.0.

PR #2786, opened 2025-08-18
fix incorrect torch version test

namgyu-youn:
**Summary:**
PyTorch pre-release/dev versions have `a0/dev` in their name. Therefore, the right order is the following:
```
2.4.0 (stable) < 2.5.0.dev / 2.5.0.git < 2.5.0 (stable)
```
For correct order, this PR applies true order within stable and pre-release/dev.

- fix: https://github.com/pytorch/ao/issues/2722

**Test plan:** CI


torchao/utils.py
```diff
@@ -368,7 +368,14 @@ def is_fbcode():
...
 def torch_version_at_least(min_version):
-    return is_fbcode() or compare_versions(torch.__version__, min_version) >= 0
+    from packaging.version import parse as parse_version
+
+    if is_fbcode():
+        return True
+
+    # Parser for local identifiers
+    current_version = re.sub(r"\+.*$", "", torch.__version__)
+    return parse_version(current_version) >= parse_version(min_version)
```
jerryzh168: I have some impression that we don't want to do this, but @msaroufim would have more context here

msaroufim: seems fine to merge altho we probably want to delete the compare_versions function

namgyu-youn: `compare_versions & parse_version` are unavailable to used here because `parse_version` only extracts `\d+\.\d+\.\d+` (e.g., 2.5.0→[2, 5, 0]). Therefore, we can inject more parsers (e.g., a0, dev) into `parse_version`, but I am not certain because `check_cpu_version & check_xpu_version` are chained with them.

andrewor14: seems like we want to change `compare_version` and `parse_version` to fix this instead, since it's used in multiple places?

andrewor14:
Hi @namgyu-youn, thanks for fixing this. I think we should actually fix `compare_versions` and `parse_version` instead, since they are used in a few different places. I'm actually wondering if we can express the comparisons like this:

```
# parse_version
2.5.0+cu126 -> [2, 5, 0]
2.5.0 -> [2, 5, 0]
2.5.0a0+git9f17037 -> [2, 5, -1]
2.5.0.dev20240708+cu121 -> [2, 5, -1]
2.4.0 -> [2, 4, 0]
```

Basically always return a list of 3 numbers, and -1 represents a pre-release. Then we can just compare as follows:

```
# don't need compare_versions anymore
parse_version(x) >= parse_version(y)

# These are True
[2, 5, -1] < [2, 5, 0]
[2, 5, -1] > [2, 4, 99]
```

To me this is simpler than the existing `compare_versions` logic, which I think is a bit unintuitive to understand. What do you think?

namgyu-youn:
> Hi @namgyu-youn, thanks for fixing this. I think we should actually fix `compare_versions` and `parse_version` instead, since they are used in a few different places. I'm actually wondering if we can express the comparisons like this:
> 
> ```
> # parse_version
> 2.5.0+cu126 -> [2, 5, 0]
> 2.5.0 -> [2, 5, 0]
> 2.5.0a0+git9f17037 -> [2, 5, -1]
> 2.5.0.dev20240708+cu121 -> [2, 5, -1]
> 2.4.0 -> [2, 4, 0]
> ```
> 
> Basically always return a list of 3 numbers, and -1 represents a pre-release. Then we can just compare as follows:
> 
> ```
> # don't need compare_versions anymore
> parse_version(x) >= parse_version(y)
> 
> # These are True
> [2, 5, -1] < [2, 5, 0]
> [2, 5, -1] > [2, 4, 99]
> ```
> 
> To me this is simpler than the existing `compare_versions` logic, which I think is a bit unintuitive to understand. What do you think?

Sounds good to me. Customized `parse_version` was just a careful suggestion, and I also agree with `-1` for better logic. Let me fix it with co-authoring; thanks again!

test/test_utils.py
```diff
@@ -16,14 +16,14 @@
 class TestTorchVersion(unittest.TestCase):
     def test_torch_version_at_least(self):
         test_cases = [
-            ("2.5.0a0+git9f17037", "2.5.0", True),
```
andrewor14, line 19: we should keep the full version string in the test case

torchao/utils.py
```diff
@@ -348,27 +348,35 @@ def _is_float8_type(dtype: torch.dtype) -> bool:
...
-    match = re.match(r"(\d+\.\d+\.\d+)", version_string)
+    """
+    Parse version string representing pre-release with -1
+
+    Examples: "2.5.0" -> [2, 5, 0], "2.5.0.dev" -> [2, 5, -1]
+    """
+    version = re.sub(r"\+.*$", "", version_string)
+
+    # Check for pre-release indicators (including all common patterns)
+    is_prerelease = bool(re.search(r"(a|b|dev)", version))
```
andrewor14: Seems like this is simpler:
```
is_prerelease = bool(re.search(r"(git|dev)", version_string))
match = re.match(r"(\d+)\.(\d+)\.(\d+)", version_string)
```
any other patterns we need to search for? @jerryzh168

torchao/utils.py
```diff
@@ -348,27 +348,35 @@ def _is_float8_type(dtype: torch.dtype) -> bool:
...
+    match = re.match(r"(\d+)\.(\d+)\.(\d+)", version)
     if match:
-        version = match.group(1)
-        return [int(x) for x in version.split(".")]
+        major, minor, patch = map(int, match.groups())
+        if is_prerelease:
+            patch = -1
```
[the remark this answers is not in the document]
andrewor14: Please add this case to the test as well

torchao/utils.py
jerryzh168: git and dev should be enough I think, we have git locally and dev in nightly

torchao/utils.py
jerryzh168: btw we also have https://github.com/pytorch/ao/blob/43b4106d282c617efd6137f78b5b10cd349fce4c/torchao/utils.py#L1002, should we just reuse that? not sure if this works for pre-releases as well, could you check?

I vaguely remember at some point that we don't want to use parse version, but don't remember why though

torchao/utils.py
```diff
@@ -348,27 +348,33 @@ def _is_float8_type(dtype: torch.dtype) -> bool:
...
 def parse_version(version_string):
-    # Extract just the X.Y.Z part from the version string
-    match = re.match(r"(\d+\.\d+\.\d+)", version_string)
+    """
+    Parse version string representing pre-release with -1
+
+    Examples: "2.5.0.dev20240708+cu121" -> [2, 5, -1], "2.5.0" -> [2, 5, 0]
+    """
+    # Check for pre-release indicators
+    is_prerelease = bool(re.search(r"(git|dev)", version_string))
```
jerryzh168, line 357: please also verify the version string for pytorch, if you build it from source

torchao/utils.py
namgyu-youn: > btw we also have
> 
> https://github.com/pytorch/ao/blob/43b4106d282c617efd6137f78b5b10cd349fce4c/torchao/utils.py#L1002
> 
> , should we just reuse that? not sure if this works for pre-releases as well, could you check?
> I vaguely remember at some point that we don't want to use parse version, but don't remember why though

### Installation
```bash
# 2.8.0 stable
pip install torch --index-url https://download.pytorch.org/whl/cpu
# 2.9.0 pre-release (nighty)
pip install torch --index-url https://download.pytorch.org/whl/nightly/cpu
```

### Test code
```
test_cases = ["2.7.0a0+git9f17037", "2.7.0", "2.8.0a0+git9f17037", "2.8.0", "2.9.0a0+git9f17037", "2.9.0"]

print(torch.__version__)
print(torch.version)

for test in test_cases:
	print(torch_version_at_least(test), is_package_at_least(package_name="torch", min_version=test))
```

### Result
```bash
# stable
2.8.0+cpu
<module 'torch.version' from '/home/youn17/Documents/git/ao/.venv/lib/python3.12/site-packages/torch/version.py'>
True False
True True
False False
False False
False True
False True

# pre-release (nighty)
2.9.0.dev20250821+cpu
<module 'torch.version' from '/home/youn17/Documents/git/ao/.venv/lib/python3.12/site-packages/torch/version.py'>
True True
True True
True False
False True
False True
False True
```

It seems that `is_package_at_least()` doesn't work for both stable/pre-release cases because`version()` returns the module object;  `__version__()` returns the **actual version string**. In my naive guess, `torch_version_at_least` and `is_package_at_least` can be consolidated because they share "return true if installed <package/torch> is higher than required" (with PyTorch pre-release detection).

torchao/utils.py
namgyu-youn: > git and dev should be enough I think, we have git locally and dev in nightly

Aren't the above comments mentioning `git` and `dev` enough for a pre-release pattern? Let me know if I am missing something.

andrewor14: Yeah should be `git` if you build from source, this is mine: `2.9.0a0+git1f19003`

andrewor14:
Thanks @namgyu-youn, this looks good to me! The test failures are unrelated.

torchao/utils.py
jerryzh168: yeah I was referring to torchao as well but then realized this is for torch

PR #2788, opened 2025-08-18
nvfp4 tensor: switch to TorchAOBaseTensor

vkuzo:
Summary:

Switches NVFP4Tensor to inherit from TorchAOBaseTensor, and removes the
custom tensor flatten/unflatten to use the base class version.

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/quantization/quantize_/workflows/int4/test_int4_marlin_sparse_tensor.py
+++ b/test/quantization/quantize_/workflows/int4/test_int4_marlin_sparse_tensor.py
@@ -21,6 +21,7 @@ from torchao.quantization import (
 )
 from torchao.quantization.utils import compute_error
 from torchao.sparsity.sparse_api import apply_fake_sparsity
+from torchao.testing.utils import skip_if_rocm
 from torchao.utils import (
     TORCH_VERSION_AT_LEAST_2_8,
 )
@@ -38,6 +39,7 @@ class TestInt4MarlinSparseTensor(TestCase):
     def setUp(self):
         self.GPU_DEVICES = ["cuda"] if torch.cuda.is_available() else []
 
+    @skip_if_rocm("ROCm enablement in progress")
     @parametrize("config", [BF16_ACT_CONFIG])
     @parametrize(
         "sizes",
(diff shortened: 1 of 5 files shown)


vkuzo:
* #2790
* __->__ #2788

torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -59,20 +60,22 @@ class NVFP4Tensor(torch.Tensor):
...
 
-    qdata: torch.Tensor
-    _scale_e4m3: torch.Tensor
-    _per_tensor_scale: Optional[torch.Tensor]
-    _block_size: int
-    _orig_dtype: torch.dtype
-    _is_swizzled_scales: bool
-    mm_config: NVFP4MMConfig
-    use_triton_kernel: bool
+    tensor_data_names = ["qdata", "_scale_e4m3"]
```
jerryzh168, line 70: nit: remove underscores?

vkuzo: not related to current PR, we can do this in a future PR

torchao/utils.py
```diff
@@ -810,6 +810,8 @@ def __tensor_flatten__(self):
                     if maybe_tensor is not None:
                         tensor_data_names.append(tensor_data_name)
 
+            # TODO(future PR): also return names of tensor attributes for easier
```
jerryzh168, line 813: not sure if we could change it, that's what we've been using in the past

torchao/prototype/mx_formats/nvfp4_tensor.py
drisspg: from a typehint / ide experience this kinda sucks

is it not possible to have the actual attributes typehinted on the class w/ AOBaseTensor?

jerryzh168: we can try to do that as an improvement for future maybe, not sure how to do that yet, maybe we can do some inspection

torchao/utils.py
vkuzo: I think it's worth changing, right now it's a list of attribute values without the names, which is way harder to debug than a key:value dictionary.

PR #2789, opened 2025-08-18
fixing torchao rocm ci test

liangel-02:


test/quantization/quantize_/workflows/int4/test_int4_marlin_sparse_tensor.py
```diff
@@ -38,6 +39,7 @@ class TestInt4MarlinSparseTensor(TestCase):
     def setUp(self):
         self.GPU_DEVICES = ["cuda"] if torch.cuda.is_available() else []
 
+    @skip_if_rocm("ROCm enablement in progress")
```
andrewor14, line 42: Will it also fail in `test_to_device`? Maybe need to add it there too? (or move the skip to the class itself)

PR #2790, opened 2025-08-18
nvfp4 tensor: refactor weight-only vs dynamic quant

vkuzo:
Summary:

Refactors `NVFP4Tensor` to use `act_quant_kwargs`, to follow the design
of recently added `Float8Tensor`.

Note that chose not to use `_choose_quant_func_and_quantize_tensor` as
we do not support any activation types other than nvfp4. This can be
relaxed in the future if needed.

This is still not the final API, might need to make more tweaks before
we bring out of prototype.

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo:
* __->__ #2790

torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -141,9 +155,11 @@ def to_nvfp4(
             block_size: Block size for quantization (must be 16)
             per_tensor_scale: Optional pre-computed absolute maximum for calibration.
                 If provided, uses per-tensor scaling. If None, uses block-wise scaling only.
-            mm_config: Matrix multiplication configuration
+            per_tensor_scale: Optional pre-computed absolute maximum for calibration for activation
```
jerryzh168: is this supposed to be `act_per_tensor_scale`? also is it expected that docstring is the same as the previous item

PR #2792, opened 2025-08-18
Add load and run tests for checkpoints that we want to have BC

jerryzh168:
Stacked PRs:
 * #2793
 * __->__#2792

--- --- ---

### Add load and run tests for checkpoints that we want to have BC

Summary:
Added load and run tests to make sure previously saved checkpoints can continue to load and run.

includes FP8, INT4 and INT4 + preshuffled checkpoints since these might reach larger audience

Test Plan:
python test/integration/test_load_and_run_checkpoint.py

Reviewers:

Subscribers:

Tasks:

Tags:


test/integration/test_load_and_run_checkpoint.py
```diff
@@ -0,0 +1,123 @@
...
+_DEPRECATED_MODEL_INFO = [
+    (
+        "torchao-testing/opt-125m-float8dq-row-v1-0.13-dev",
+        1,
+        "Float8DynamicActivationFloat8WeightConfig",
+    ),
+]
+
+_MODEL_NAMES = [
+    "torchao-testing/single-linear-FP8-v2-0.13-dev",
```
vkuzo: overall looks great, can we have a more exact naming scheme? use the actual config name instead of "FP8", etc

vkuzo: where is the code to create these models? can we check it in so it's easily extendable?

test/integration/test_load_and_run_checkpoint.py
```diff
@@ -0,0 +1,123 @@
...
+            )
+
+        tokenizer = AutoTokenizer.from_pretrained(model_name)
+        prompt = ("Hello, my name is",)
+        inputs = tokenizer(
+            prompt,
+            return_tensors="pt",
+        ).to("cuda")
+        generated_ids = quantized_model.generate(**inputs, max_new_tokens=128)
+        # make sure it runs
```
vkuzo: instead of testing that it runs, can we save the logits, set temperature to 0 and then test for bitwise equivalency?

test/integration/test_load_and_run_checkpoint.py
```diff
@@ -0,0 +1,123 @@
...
+            model_name, filename="model_inputs.pt"
+        )
+        with open(downloaded_example_inputs, "rb") as f:
+            example_inputs = torch.load(f)
+        downloaded_output = hf_hub_download(model_name, filename="model_output.pt")
+        with open(downloaded_output, "rb") as f:
+            ref_output = torch.load(f)
+
+        output = model(*example_inputs)
+        self.assertTrue(torch.allclose(output, ref_output))
```
vkuzo: can this have bitwise equivalency?

test/integration/test_load_and_run_checkpoint.py
jerryzh168: OK, code is in the model card of these models, I can add a sample in the test as well

test/integration/test_load_and_run_checkpoint.py
```diff
@@ -0,0 +1,160 @@
...
+@unittest.skipIf(not is_sm_at_least_89(), "Nedd sm89+")
+@unittest.skipIf(
+    is_fbcode(),
+    "Skipping the test in fbcode for now, not sure how to download from transformers",
+)
+class TestLoadAndRunCheckpoint(TestCase):
+    def _test_single_linear_helper(self, model_name):
+        from huggingface_hub import hf_hub_download
+
+        downloaded_model = hf_hub_download(model_name, filename="model.bin")
```
vkuzo: to clarify, this is a state dict? maybe filename can reflect this?  Also, should the extension be `pt` to match PyTorch convention?

test/integration/test_load_and_run_checkpoint.py
```diff
@@ -0,0 +1,160 @@
...
+    def test_single_linear(self, model_name):
+        """Test that we can load and run the quantized linear checkpoint with saved sample input
+        and match the saved output, to make sure there is no BC breaking changes
+        when we make changes to tensor subclass implementations
+        """
+        if (
+            "Float8DynamicActivationFloat8WeightConfig" in model_name
+            and not is_sm_at_least_90()
+        ):
+            return unittest.skip("FP8 checkpoint is produced in SM90+")
```
vkuzo: should we ensure we have CI with sm90+ running this test?

test/integration/test_load_and_run_checkpoint.py
```diff
@@ -0,0 +1,160 @@
...
+)
+from transformers import AutoModelForCausalLM, AutoTokenizer, TorchAoConfig
+
+from torchao.utils import is_fbcode, is_sm_at_least_89, is_sm_at_least_90
+
+# please check model card for how to generate these models
+
+_DEPRECATED_SINGLE_LINEAR_MODEL_NAMES = [
+    # model card: https://huggingface.co/torchao-testing/single-linear-Float8DynamicActivationFloat8WeightConfig-v1-0.13-dev
+    "torchao-testing/single-linear-Float8DynamicActivationFloat8WeightConfig-v1-0.13-dev",
```
vkuzo: nit: specify what here is torchao version and what is config version? `torchao-testing/model:single_linear-config:Float8DynamicActivationFloat8WeightConfig-config_version:1-torchao:0.13-dev`?

test/integration/test_load_and_run_checkpoint.py
jerryzh168: yeah this is a state_dict, this is trying to match the convention of huggingface models: https://huggingface.co/pytorch/Qwen3-8B-INT4/blob/main/pytorch_model-00001-of-00002.bin

I can match pt convention as well, since it's also what people are used to for state dict

test/integration/test_load_and_run_checkpoint.py
jerryzh168: yeah it's tested here: https://github.com/pytorch/ao/blob/083361bc3f7addc505a0f994a923f4ae9f54388e/.github/workflows/1xH100_tests.yml#L50

added this skip because it fails in this test: https://github.com/pytorch/ao/blob/083361bc3f7addc505a0f994a923f4ae9f54388e/.github/workflows/1xL4_tests.yml#L50 (SM89)

test/integration/test_load_and_run_checkpoint.py
jerryzh168: this seems a bit long, would changing `0.13-dev` to `0.13.dev` be clearer that this is a torchao version?

also just found that `:` is not supported in hf model naming, only `-`, `.` and `_` are supported.

can also move around the 0.13.dev to make it clearer as well, please let me know

vkuzo: > this seems a bit long, would changing 0.13-dev to 0.13.dev be clearer that this is a torchao version?

I think we should go for clarity here, if it's long that's ok.  Can we pick one of `-`, `.`, `_` as a delimiter, and make the name be formatted with `key1.val1,key2.val2,...`, with whatever delimiters between pairs and key:val that we can use

vkuzo (approved):
would be good to make the naming exact

test/integration/test_load_and_run_checkpoint.py
jerryzh168: does this have to be in the name, how about adding these in model card descriptions? there are some restrictions to the naming, including max length of 96:
```
Only regular characters and '-', '_', '.' are accepted. '--' and '..' are forbidden. '-' and '.' cannot start or end the name. The name cannot end with ".git" or ".ipynb". Max length is 96.
```

tried Float8 and it's 111: `model_single-linear-config_Float8DynamicActivationFloat8WeightConfig-config-version_v2-torchao-version_0.13.dev`

jerryzh168:
updated model card to include exact descriptions due to naming length restrictions

PR #2795, opened 2025-08-19
Fix batch norm folding in `prepare_pt2e` for multiple conv->BN chains sharing the same conv weights

subhankarpal:
### Summary
For a model with multiple conv->BN chains that have the same conv weights, `prepare_pt2e` applies batch norm folding repeatedly for each chain, causing incorrect conv weights in the output prepared model. This fix ensures batch norm folding is performed only once per unique conv weight set, preventing multiple folds and preserving correct model behavior.

### Testplan
```
python test/quantization/pt2e/test_quantize_pt2e.py -k test_chunked_bn_fusion
```

### DIFF
--- a/torchao/quantization/pt2e/utils.py
+++ b/torchao/quantization/pt2e/utils.py
@@ -671,6 +671,7 @@ def fold_bn_weights_into_conv_node(
     conv_bias_node: Optional[Node],
     bn_node: Node,
     m: GraphModule,
+    fake_fuse: bool=False # removes the BN nodes but doesn't change the conv weights
 ) -> None:
     # conv args: input, weight, bias, stride, padding, dilation, ...
     conv_w = _get_tensor_constant_from_node(conv_weight_node, m)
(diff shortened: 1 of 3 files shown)


torchao/testing/model_architectures.py
```diff
@@ -22,6 +22,26 @@ def forward(self, x):
         return x
 
 
+class ToyCNNModel(nn.Module):
```
jerryzh168: nit: better to be more specific I think, like `ConvWithSharedWeightInExportedModel`

torchao/quantization/pt2e/utils.py
```diff
@@ -781,9 +795,15 @@ def _fuse_conv_bn_(m: GraphModule) -> None:
...
+            conv_node,
+            conv_weight_node,
+            conv_bias_node,
+            bn_node,
+            m,
+            (conv_weight_node in fused_convs_weight_nodes)
         )
-
+        fused_convs_weight_nodes.add(conv_weight_node)
+    del fused_convs_weight_nodes
```
jerryzh168: nit: probably no need to delete this explicitly

subhankarpal:
> see some nit comments inline

Fixed, thanks.

jerryzh168:
@subhankarpal no need for manual import, our oncall will import this in diff train

amdfaa:
This PR appears to be causing failures for both rocm and cuda: https://hud.pytorch.org/failure?name=Run%20Regression%20Tests%20on%20ROCm%20%2F%20test-nightly%20(ROCM%20Nightly%2C%20linux.rocm.gpu.gfx942.2%2C%20--pre%20torch%20--index-url%20https%3A%2F%2Fdownload.pyt...%20%2F%20linux-job&jobName=undefined&failureCaptures=test%2Fintegration%2Ftest_integration.py%3A%3ATestAutoQuant%3A%3Atest_autoquant_hp_float

PR #2806, opened 2025-08-19
mxtensor: refactor activation quant to use direct logic

vkuzo:
Summary:

Refactors MXTensor's inference quantization from using linear activation
wrapper tensor to quantizing the activation directly in the `mm`
function dispatch, for simplicity and to align with design of our new
tensors.

Test Plan:

```bash
```

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/prototype/mx_formats/mx_ops.py
+++ b/torchao/prototype/mx_formats/mx_ops.py
@@ -91,8 +91,8 @@ def _addmm_mx_dispatch(
     if gemm_choice in (MXGemmKernelChoice.CUBLAS, MXGemmKernelChoice.CUTLASS):
         # real MX gemm backed by torchao's CUTLASS kernels
         M, K, N = a.shape[0], a.shape[1], b.shape[1]
-        assert a._data.is_contiguous()
-        assert b._data.t().is_contiguous()
+        assert a.qdata.is_contiguous()
+        assert b.qdata.t().is_contiguous()
         assert a._block_size == 32, f"Invalid block size {a._block_size}"
         assert b._block_size == 32, f"Invalid block size {b._block_size}"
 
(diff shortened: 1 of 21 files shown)


vkuzo:
* __->__ #2806

torchao/prototype/mx_formats/mx_ops.py
```diff
@@ -80,12 +80,26 @@ def _get_gemm_choice(
...
 ) -> torch.Tensor:
     """
     Core implementation shared between mx_mm and mx_addmm.
     The only difference is whether bias is None or not.
     """
+
+    if not isinstance(a, MXTensor):
+        assert b.act_quant_kwargs is not None, "weight-only quant not yet supported"
+        k = b.act_quant_kwargs
+        a = MXTensor.to_mx(
```
jerryzh168, line 93: nit: rename to `from_hp` for all tensor subclasses for consistency (feel free to do in separate PR)

torchao/prototype/mx_formats/mx_tensor.py
```diff
@@ -68,6 +72,16 @@
 EBITS_F8_E5M2, MBITS_F8_E5M2 = 5, 2
 
 
+@dataclass
+class QuantizeTensorToMXKwargs(QuantizeTensorKwargs):
```
jerryzh168, line 76: is this used / tested now? I think we'd need to add a handling for it in https://github.com/pytorch/ao/blob/083361bc3f7addc505a0f994a923f4ae9f54388e/torchao/quantization/quantize_/common/quantize_tensor_kwargs.py#L33 in order to use it

torchao/prototype/mx_formats/mx_ops.py
jerryzh168: oh it's defined here, makes sense as well since it's only used once, should we remove _choose_quant_func_and_quantize_tensor

jerryzh168 (approved):

torchao/prototype/mx_formats/mx_tensor.py
vkuzo: I think `_choose_quant_func_and_quantize_tensor` makes sense if there is support for different quantized activation formats, if the kernels only support one quantized activation format then simpler to just quantize directly, like I do in this PR.  We can always switch to that function later as this is not a part of BC surface.

jerryzh168: makes sense

PR #2808, opened 2025-08-19
turn float8 inference kernel check test back on

vkuzo:
Summary:

This test was likely flaky because torch.compile with mode
`reduce-overhead` turns on CUDA graphs, CUDA graphs launches a
couple of extra kernels to copy data to the graph and copy it back, and
the # of extra kernels launched varied between local env and CI.

For now, removing `reduce-overhead` mode so we can have a solid test for
default settings. Also, extending the test to the v2 config to ensure we
know if expected # of kernels in new production path breaks.

Test Plan:

```bash
pytest test/dtypes/test_affine_quantized_float.py -k test_expected_kernels_on_gpu -s
```

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/dtypes/test_affine_quantized_float.py
+++ b/test/dtypes/test_affine_quantized_float.py
@@ -36,6 +36,7 @@ from torchao.quantization.quant_primitives import (
     _quantize_affine_float8,
     choose_qparams_affine,
 )
+from torchao.quantization.quantize_.common import KernelPreference
 from torchao.utils import (
     is_sm_at_least_89,
     is_sm_at_least_90,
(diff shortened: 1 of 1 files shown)


vkuzo:
* __->__ #2808

test/dtypes/test_affine_quantized_float.py
```diff
@@ -756,14 +750,23 @@ def test_expected_kernels_on_gpu(self, granularity, torch_compile_mode):
...
+        if float8_config_version == 1:
+            config = Float8DynamicActivationFloat8WeightConfig(
+                granularity=granularity, version=1
+            )
+        else:
+            assert float8_config_version == 2
+            config = Float8DynamicActivationFloat8WeightConfig(
+                granularity=granularity,
+                version=2,
+                kernel_preference=KernelPreference.TORCH,
```
jerryzh168, line 762: what about other kernel preferences? I think the default people use is `KerenelPreference.AUTO`, although the implementation might change over time for this option

jerryzh168 (approved):

test/dtypes/test_affine_quantized_float.py
vkuzo: we can add them in future PRs, for now it would be good to resume coverage in CI

vkuzo:
failures look unrelated, landing

PR #2818, opened 2025-08-20
Fix autoquant tests failed due to changes to benchmark_gpu

jerryzh168:
Summary:
benchmark_gpu is updated from triton to inductor by default: https://github.com/pytorch/pytorch/blob/8047cde0f3a27f3afa218792b8464d5e0c9d942f/torch/_inductor/runtime/benchmarking.py#L316, and it only accepts "min" or "all" return mode, causing autoquant to fail in nightly

Test Plan:
CI

Reviewers:

Subscribers:

Tasks:

Tags:


jerryzh168:
can repro in the most recent pytorch version actually, let me check

PR #2834, opened 2025-08-21
Relax parq start/end check

yasmin496:
Summary: Relax parq start/end check to unblock dbg runs after D80536544

Reviewed By: Ninja91

Differential Revision: D80709175


lisjin:
It looks like this change was removed internally—closing it for now

PR #2841, opened 2025-08-21
[mxfp8 moe] replace per group scaling with conventional scaling

danielvegamyhre:
Stacked PRs:
 * #2844
 * __->__#2841
 * #2835

--- --- ---

[mxfp8 moe] replace per group scaling with conventional scaling

## Summary

Now that token group group aligment size is configurable in torchtitan (https://github.com/pytorch/torchtitan/pull/1503), in MXFP8 training we [set token group alignment size to 32](https://github.com/pytorch/torchtitan/blob/fd230800d2385e782f4443e24f1e45d83a6e6b0d/torchtitan/components/quantization/mx.py#L65-L66). This is to ensure when we quantize along that dim, we have (1) no scaling groups crossing logically distinct tensor boundaries, and (2) no “partial” scaling group at the the end (e..g, 1x17 chunk because the tensor ends).

- There are a couple places I had to call `.contiguous()` to meet the constraints of `to_mx()`. We should look into ways to avoid this, if possible.

By padding to avoid the case where scaling groups cross token group boundaries, we can now migrate to the plain `to_mx()` API.

## Additional context
Note the previous pytorch looped implementation was just intended to be used as a basic reference to check future per-group scaling kernels correctness for the simple case (token groups divisible by block size), but after thinking about it more, I realized any potential per-group scaling kernels we write would need to do padding inside them anyways to handle the case where scaling groups cross token group boundaries, thus we should just do the padding ahead of time and avoid the need for custom kernels all together.

### DIFF
--- a/torchao/prototype/moe_training/scaled_grouped_mm.py
+++ b/torchao/prototype/moe_training/scaled_grouped_mm.py
@@ -5,7 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 import logging
-from typing import Optional, Tuple
+from typing import Optional
 
 import torch
 
@@ -19,8 +19,6 @@ from torchao.prototype.moe_training.kernels import (
 )
 from torchao.prototype.moe_training.utils import (
     _is_column_major,
-    _to_mxfp8_per_group_colwise,
-    _to_mxfp8_per_group_rowwise,
 )
 from torchao.prototype.mx_formats.mx_tensor import to_mx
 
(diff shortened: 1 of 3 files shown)


torchao/prototype/moe_training/scaled_grouped_mm.py
```diff
@@ -295,7 +293,22 @@ def forward(
         # Cast B_t per-expert to mxfp8 across dim1.
         # B_t_mx shape: (E, K, N)
         # B_t_scale shape: (E, K//block_size, N)
-        B_t_scale, B_t_mx = _to_mxfp8_3d_expert_weights_dim1(B_t, block_size=block_size)
+
+        # To cast B_t per-expert to mxfp8 across dim1, we transpose the experts, cast along dim -1, then untranspose.
+        # B_mx shape: (E, N, K)
+        # B_scale shape: (E, N, K//block_size)
+        B_scales_dim2, B_mx_dim2 = to_mx(
```
vkuzo, line 300: nit: I think the right variable name would be `B_scales_...` and `B_data_...`, MX is the combination of data + scale and it's misleading to refer to just the data as "MX".

I'd say either use MXTensor, or use "scale" + "data"

vkuzo (approved):
looks right, I think var naming should be adjusted. Can be a separate PR.

torchao/prototype/moe_training/scaled_grouped_mm.py
danielvegamyhre: Makes sense, will update the var naming pattern in a follow up

Issue #2673, opened 2025-08-04
[moe training] fsdp2 bug for llama4 shared experts where num_experts=1


danielvegamyhre:
# Summary
- When applying low precision MoE training (fp8 rowwise in this case) there is a bug with the ScaledGroupedMMTensor subclass I believe. Specifically, we see fsdp2 complain about the all gather inputs shape in the error: 

```
    File "/home/danvm/.conda/envs/torchtitan/lib/python3.13/site-packages/torch/distributed/fsdp/_fully_shard/_fsdp_param_group.py", line 314, in unshard
      self._all_gather_result = foreach_all_gather(
                                ~~~~~~~~~~~~~~~~~~^
          self.fsdp_params,
          ^^^^^^^^^^^^^^^^^
      ...<4 lines>...
          self._all_gather_comm,
          ^^^^^^^^^^^^^^^^^^^^^^
      )
      ^
    File "/home/danvm/.conda/envs/torchtitan/lib/python3.13/site-packages/torch/utils/_contextlib.py", line 120, in decorate_context
      return func(*args, **kwargs)
    File "/home/danvm/.conda/envs/torchtitan/lib/python3.13/site-packages/torch/distributed/fsdp/_fully_shard/_fsdp_collectives.py", line 248, in foreach_all_gather
      param_all_gather_inputs = _get_param_all_gather_inputs(fsdp_params)
    File "/home/danvm/.conda/envs/torchtitan/lib/python3.13/site-packages/torch/utils/_contextlib.py", line 120, in decorate_context
      return func(*args, **kwargs)
    File "/home/danvm/.conda/envs/torchtitan/lib/python3.13/site-packages/torch/distributed/fsdp/_fully_shard/_fsdp_collectives.py", line 327, in _get_param_all_gather_inputs
      param_all_gather_inputs[i] = fsdp_param.all_gather_inputs
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/home/danvm/.conda/envs/torchtitan/lib/python3.13/site-packages/torch/distributed/fsdp/_fully_shard/_fsdp_param.py", line 727, in all_gather_inputs
      raise AssertionError(
      ...<4 lines>...
      )
[...]

PR #2842, opened 2025-08-21
test fsdp2 Moe

weifengpy:


weifengpy:
trying to repro https://github.com/pytorch/ao/issues/2673 but this passed. need to sync more with @danielvegamyhre

PR #2845, opened 2025-08-22
Add Int4PlainInt32Tensor

liangan1:
This PR is used to enable the Int4PlainInt32Tensor. The pacing format name is "plain_int32"
**Testcase:**
`bash
test/quantization/quantize_/workflows/int4/test_int4_plain_int32_tensor.py
`


torchao/quantization/quantize_/common/packing_format.py
```diff
@@ -40,3 +40,6 @@ class PackingFormat(str, Enum):
     Unpacked means the subbyte quantized data is stored as int8
     """
     UNPACKED_TO_INT8 = "unpacked_to_int8"
+
+    "int4_xpu_int_zp is referring to the format used by int4 weight-only quantization on XPU with int zero point, which is a groupwise quantization format."
+    INT4_XPU_INT_ZP = "int4_xpu_int_zp"
```
jerryzh168: please don't include int4 and xpu in the name, can you name this in terms of of how the quantized data is packed?

torchao/quantization/quantize_/workflows/int4/int4_xpu_tensor.py
```diff
@@ -0,0 +1,182 @@
...
+aten = torch.ops.aten
+
+
+class Int4XPUTensorIntZP(TorchAOBaseTensor):
+    """
+    int4 weight-only quantization on XPU with oneDNN as backend (groupwise quantization only)
+
+    Tensor Attributes:
+        qdata: packed int4 weigh, always viewed as a 2D (N, K/2) tensor, last dimension is packed
+               preshuffling is specific to CPU kernels, see Note below.
```
jerryzh168: is this correct? can you update this to describe how xpu tensor is packed? examples can be found in https://github.com/pytorch/ao/blob/c93bc7d2725e0f5af1c266d90f6f0838c17c4e92/torchao/quantization/quantize_/common/packing_format.py#L18

I guess the detailed doc can exist in packing_format.py, here maybe a high level summary of how the packing format looks like is good enough

torchao/quantization/quantize_/common/packing_format.py
liangan1: The int4 weight xpu is a plain format tensor according  to this doc, it just pack 2 int4 weight elements in a byte and then store the 4*int4 as int32. So I change it to the plain.

torchao/quantization/quantize_/workflows/int4/int4_xpu_tensor.py
liangan1: Updated.

torchao/quantization/quantize_/common/packing_format.py
jerryzh168: I see, we have plain that stores 2*int4 as int8, can you reuse it or would need a new one? https://github.com/pytorch/ao/blob/main/torchao/quantization/quantize_/workflows/int4/int4_tensor.py

jerryzh168: @liangan1 can you use `PLAIN_INT32` for packing_format, and rename things accordingly (tensor subclass, files etc.)

liangan1: Thanks @jerryzh168. I have added PLAIN_INT32 to be used by the xpu int4. Per my understanding,  the packing format should be a dispatch policy to select the right tensor subclassing and a tensor subclass should cover a specific quantization recipe. So I suppose I should keep the current tensor name for int4 xpu. 
In this PR, we just want to enable the int xpu with int zp domain. The current oneDNN backend can not support the float zp as CUDA/CPU backend and the feature is WIP.  I plain to reuse this packing format in the future and dispatch the tensor with the zero point domain information.

torchao/quantization/quantize_/workflows/__init__.py
```diff
@@ -14,6 +14,9 @@
 from .int4.int4_tensor import (
     Int4Tensor,
 )
+from .int4.int4_xpu_tensor import (
+    Int4XPUTensorIntZP,
```
jerryzh168: nit: we should rename the tensor names etc. as well

torchao/quantization/quantize_/workflows/int4/int4_xpu_tensor.py
```diff
@@ -0,0 +1,181 @@
...
+            zero_point,
+            target_dtype,
+            quant_min,
+            quant_max,
+        )
+        assert int_data.dtype == torch.int32, (
+            "torch.ops.aten._convert_weight_to_int4pack expects `int32` dtype"
+        )
+        packed_weight = (int_data[::, 1::2] << 4 | int_data[::, ::2]).to(torch.uint8)
+        packed_weight = torch.ops.aten._convert_weight_to_int4pack(
```
jerryzh168: actually is this int8 or int32?

if it's int8, seems like we should probably just merge into Int4Tensor, it has the same packing format [N, K/2] (int8 dtype)

liangan1: It is int32 with [N, K/8], pls refer to https://github.com/intel/torch-xpu-ops/blob/ab6e7b0b518a8ed921f85f8d48dc3c2551dbdcfa/src/ATen/native/xpu/WeightInt4Pack.cpp#L7

torchao/quantization/quantize_/workflows/__init__.py
liangan1: Per my understanding, the packing format should show the weight packing format, but I also want add the zp domian information in this tensor, so keep the tensor name as Int4XPUTensorIntZP.

jerryzh168: the packing format should contain a description of how the quantized weights are laid out in memory, it should not mention `XPU`, and also not `INTZP` since these are not relevant to how the quantized weights laid out.

can you change it to something like `Int4PlainInt32` since `PLAIN_INT32` is the packing_format?

torchao/quantization/quantize_/common/packing_format.py
jerryzh168: you can reuse the packing format and the tensor for float32 zero_point as well in the future I think, but today we structure tensor subclass by: dtype + packing_format, so `Int4PlainInt32` might be better

torchao/quantization/quantize_/workflows/int4/int4_xpu_tensor.py
jerryzh168: OK sounds good

torchao/quantization/quantize_/workflows/__init__.py
liangan1: Int4PlainInt32 is good for me. But I still have some concern about the relation between packing_format and subclassing tensor? it must be 1 x 1?

jerryzh168: I think so, it's 1x1, we are also going to do a follow up refactor to have Int4PackingFormat instead having a global PackingFormat, since there is not much reuse

theoretically, 1 packing format can correspond to multiple tensors with different dtypes, but we found that there is not much reuse of packing_format, that's why we want to move it to local to each dtype (int4, float8, intx)

what's the concern for this?

torchao/quantization/quantize_/common/packing_format.py
liangan1: Done. change it to Int4PlainInt32. pls help to review again.

torchao/quantization/quantize_/workflows/__init__.py
liangan1: Thanks. Your explanation is clear enough. No more concern now.

torchao/quantization/quantize_/workflows/int4/int4_plain_int32_tensor.py
```diff
@@ -0,0 +1,181 @@
...
+
+        original_shape = w.shape
+        mapping_type = MappingType.ASYMMETRIC
+        target_dtype = torch.int32
+        quant_min = 0
+        quant_max = 15
+        eps = 1e-6
+        scale_dtype = None
+        zero_point_dtype = torch.int32
+        scale, zero_point = _choose_qparams_affine(
```
jerryzh168: nit: we can use the official op? https://github.com/pytorch/ao/blob/f02354d82463e21e760b1ea5b03ee448f3ee6fe3/torchao/quantization/quant_primitives.py#L1194 same for quantize_affine

test/quantization/quantize_/workflows/int4/test_int4_plain_int32_tensor.py
```diff
@@ -0,0 +1,89 @@
...
+from torchao.utils import (
+    torch_version_at_least,
+)
+
+
+def get_config(group_size):
+    return Int4WeightOnlyConfig(
+        group_size=group_size,
+        packing_format="plain_int32",
+        zero_point_domain=ZeroPointDomain.INT,
```
jerryzh168: nit: we don't need this anymore I think, also we want to remove ZeroPointDomain in the future

torchao/quantization/__init__.py
```diff
@@ -162,9 +163,10 @@
     "FbgemmConfig",
     # tensor subclasses
     "Int4Tensor",
+    "Int4PlainInt32",
     "Int4PreshuffledTensor",
     "Int4MarlinSparseTensor",
-    "IntxOpaqueTensor",
+    "IntxOpaqueTensor",
```
jerryzh168: nit: extra spaces?

torchao/quantization/quantize_/workflows/int4/int4_plain_int32_tensor.py
```diff
@@ -0,0 +1,181 @@
...
+
+aten = torch.ops.aten
+
+
+class Int4PlainInt32(TorchAOBaseTensor):
+    """
+    int4 weight-only quantization on XPU with oneDNN as backend (groupwise quantization only)
+
+    Tensor Attributes:
+        qdata: packed int4 weigh, always viewed as a 2D (N, K/2) tensor
```
jerryzh168: nit: would be good to callout the dtype here, since it's important in this case

torchao/quantization/quantize_/workflows/int4/int4_plain_int32_tensor.py
```diff
@@ -0,0 +1,181 @@
...
+        scale: (K/group_size, N), dtype is the same as the original Tensor dtype
+        zero_point: (K/group_size, N)
+
+    Non-Tensor Attributes:
+        block_size: the block size for quantization, representing the granularity, for groupwise quantization, will have block_size (1, group_size).
+                    we only support group_size = 32/64/128.
```
jerryzh168: add an assert in `from_hp`?

torchao/quantization/quantize_/workflows/int4/int4_plain_int32_tensor.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+class Int4PlainInt32(TorchAOBaseTensor):
+    """
+    int4 weight-only quantization on XPU with oneDNN as backend (groupwise quantization only)
+
+    Tensor Attributes:
+        qdata: packed int4 weigh, always viewed as a 2D (N, K/2) tensor
+        scale: (K/group_size, N), dtype is the same as the original Tensor dtype
+        zero_point: (K/group_size, N)
```
jerryzh168: also dtypes for these for more clarity

torchao/quantization/quantize_/workflows/int4/int4_plain_int32_tensor.py
liangan1: Done.

test/quantization/quantize_/workflows/int4/test_int4_plain_int32_tensor.py
liangan1: Removed. But I have a question that how to selelct the int zp domain for user if there is no this param?

torchao/quantization/__init__.py
liangan1: fixed.

torchao/quantization/quantize_/workflows/int4/int4_plain_int32_tensor.py
liangan1: Updated.

torchao/quantization/quantize_/workflows/int4/int4_plain_int32_tensor.py
liangan1: Corrected. There is no block size limitation for XPU backend.

torchao/quantization/quantize_/workflows/int4/int4_plain_int32_tensor.py
liangan1: Add the dtype information for zero_point.

torchao/quantization/quantize_/workflows/int4/int4_plain_int32_tensor.py
```diff
@@ -0,0 +1,180 @@
...
+aten = torch.ops.aten
+
+
+class Int4PlainInt32Tensor(TorchAOBaseTensor):
+    """
+    int4 weight-only quantization on XPU with oneDNN as backend (groupwise quantization only)
+
+    Tensor Attributes:
+        qdata: (N, K/8), packed int4 weight, the data type is int32 here with 4*(int4*2)
+        scale: (K/group_size, N), dtype is the same as the original Tensor dtype
```
jerryzh168, line 34: btw, might also be good to specify what is the supported and optimized original tensor dtype as well, I feel we have been a bit vague on this for all these tensors

test/quantization/quantize_/workflows/int4/test_int4_plain_int32_tensor.py
jerryzh168: we'll know how to quantize based on the type of tensor, so user just need to choose the packing_format

jerryzh168:
please rebase, and also fix the CI error as well, need to skip the test when there is no xpu I think

maybe update the Summary to make sure the naming are correct as well

torchao/quantization/quantize_/workflows/int4/int4_plain_int32_tensor.py
liangan1: Added more comments about the original data type. Both  half and bfloat16 are supported.

liangan1:
cc @xiaowangintel

liangan1:
> please rebase, and also fix the CI error as well, need to skip the test when there is no xpu I think
> 
> maybe update the Summary to make sure the naming are correct as well

Done. @jerryzh168  pls help review again.

torchao/quantization/quantize_/workflows/int4/int4_plain_int32_tensor.py
```diff
@@ -0,0 +1,180 @@
...
+    def _quantization_type(self):
+        return f"shape={self.shape}, block_size={self.block_size}, device={self.device}"
+
+    @classmethod
+    def from_hp(
+        cls,
+        w: torch.Tensor,
+        block_size: List[int],
+    ):
+        assert w.ndim == 2 and w.device.type == "xpu", (
```
jerryzh168, line 75: if `the original data type can be half and bfloat16` can you add an assert here as well?

test/quantization/quantize_/workflows/int4/test_int4_plain_int32_tensor.py
```diff
@@ -0,0 +1,86 @@
...
+    return Int4WeightOnlyConfig(
+        group_size=group_size,
+        packing_format="plain_int32",
+        version=2,
+    )
+
+
+@unittest.skipIf(not torch_version_at_least("2.8.0"), "Need pytorch 2.8+")
+@unittest.skipIf(not torch.xpu.is_available(), "XPU not available")
+class Int4PlainInt32Tensor(TestCase):
```
jerryzh168, line 38: we probably need more tests like serailization etc. but can add these later

jerryzh168 (approved):

torchao/quantization/quantize_/workflows/int4/int4_plain_int32_tensor.py
liangan1: Done.

test/quantization/quantize_/workflows/int4/test_int4_plain_int32_tensor.py
liangan1: Ok. we are working on the XPU CI enabling in other PRs. Pls refer to https://github.com/pytorch/ao/issues/2917

PR #2744, opened 2025-08-12
[WIP] simple MoE inference API

vkuzo:
Summary:

Not ready for review yet

Test Plan:

```python
pytest test/prototype/moe_inference/test_base.py -s -x
```

Reviewers:

Subscribers:

Tasks:

Tags:

PR #2846, opened 2025-08-22
Fix Llama4 example

yiliu30:


torchao/prototype/moe_quant/llama4_quant.py
```diff
@@ -58,7 +58,10 @@ def convert_fn(module):
 
 
 model_id = "meta-llama/Llama-4-Scout-17B-16E-Instruct"
-model = Llama4ForCausalLM.from_pretrained(model_id, torch_dtype=torch.bfloat16)
+
+dtype = torch.bfloat16
+torch.set_default_dtype(dtype)
```
yiliu30, line 63: To make sure the input and model weights are the same dtype.

torchao/prototype/moe_quant/quantizable_moe_modules.py
```diff
@@ -30,7 +30,7 @@ def __init__(
     def forward(self, x: Tensor) -> Tensor:
         batch_size = x.shape[0]
         x = x.view(-1, self.hidden_dim)  # x: [T, D]
-        scores = self.router(x)  # [T, E]
+        scores = self.router(x)[0]  # [T, E]
```
yiliu30, line 33: The original router returns `router_scores` and `router_logits`.
https://github.com/huggingface/transformers/blob/19ffe0219dae122203f9726669f88ef1c6ea3bb4/src/transformers/models/llama4/modeling_llama4.py#L143

yiliu30:
Hi @liangel-02 @andrewor14   Looks like the failed CI checks are not related to this PR. Could you help retrigger them?

torchao/prototype/moe_quant/quantizable_moe_modules.py
jerryzh168: router here is a nn.Linear, not `Llama4Router` actually, see L21?

yiliu30: This `rounter` was replaced with the original rounter.
https://github.com/pytorch/ao/blob/9d01b43ecbaa14433762fd6787409808c022a8cd/torchao/prototype/moe_quant/llama4_quant.py#L48-L53

jerryzh168: so this module does not run by itself? seems quite confusing

yiliu30: The `MOEFeedForwardAOQuantizable` was used, but it seems that its `rounter` shouldn’t be quantized in order to preserve accuracy. Could you confirm that? @HDCharles

jerryzh168:
@yiliu30 I'd recommend wait a bit until https://github.com/pytorch/ao/pull/2744 is landed for moe support

yiliu30:
> @yiliu30 I'd recommend wait a bit until #2744 is landed for moe support

Sure, thanks for the information.

PR #2848, opened 2025-08-22
[mxfp8 moe] add support for fbgemm 2d-3d mx8mx8bf16 grouped gemm

danielvegamyhre:
Stacked PRs:
 * #2886
 * #2882
 * #2879
 * __->__#2848

--- --- ---

## Summary 
- fbgemm recently added a 2d-3d mxfp8 grouped gemm in https://github.com/pytorch/FBGEMM/pull/4710
- This PR integrates the new gemm into the MoE training code base for the following 2d-3d grouped gemms:
    - `output = input @ weight^T`
    - `grad_input = grad_output @ weight` 
- Add new mxfp8 utils `to_blocked_per_group_2d` (for input scales) and `to_blocked_per_group_3d` (for weight scales). These are pytorch reference implementations that are not performant. We can implement equivalent triton kernels for them later.
- Notes on fbgemm API and pytorch grouped mm API:
    - x must be shape (Mg, K) and row major / contiguous
    - x scales must have been preprocessed to have **per-group blocked layout** and be contiguous
    - weights must be shape (E, N, K) and row major / contiguous
    - weight scales must have been pre-processed to have **per-group blocked layout** and be contiguous
    - group sizes is a vector containing the size of each token group in the `x` tensor
    - starting_row_after_padding corresponds to `x_scales` tensor and must be size `len(group_sizes) + 1` where the first starting row is always 0, and each value corresponds to the starting row of group[i] in the `x_scales` tensor AFTER padding
- Refactor `_emulated_mxfp8_scaled_grouped_mm_2d_3d` to have same function signature and input constraints as the fbgemm API
  
## Test plan
- `pytest test/prototype/moe_training/test_scaled_grouped_mm.py -k mx`
- `pytest test/prototype/moe_training/test_training.py -k mx`

### DIFF
--- a/test/prototype/moe_training/test_scaled_grouped_mm.py
+++ b/test/prototype/moe_training/test_scaled_grouped_mm.py
@@ -230,25 +230,26 @@ def compute_reference_forward(
 @pytest.mark.parametrize("num_experts", (1, 8, 16))
 def test_emulate_mxfp8_grouped_gemm_2d_3d(M, K, N, num_experts):
     x = torch.randn(M, K, dtype=torch.bfloat16, device="cuda")
-    w_t = torch.randn(num_experts, K, N, dtype=torch.bfloat16, device="cuda")
+    w = torch.randn(num_experts, N, K, dtype=torch.bfloat16, device="cuda")
     offs = generate_jagged_offs(num_experts, M)
-    x_ref, w_t_ref, offs_ref = x.clone(), w_t.clone(), offs.clone()
+    x_ref, w_ref, offs_ref = x.clone(), w.clone(), offs.clone()
 
     # Quantize inputs to mxpf8 for emulated mxfp8 scaled grouped mm
     block_size = 32
-    x_scale, x_mx = to_mx(x, elem_dtype=torch.float8_e4m3fn, block_size=block_size)
+    x_scale, x_fp8 = to_mx(x, elem_dtype=torch.float8_e4m3fn, block_size=block_size)
 
     # To cast B_t per-expert to mxfp8 across dim1, we transpose the experts, cast along dim -1, then untranspose.
-    w_scale, w_mx = to_mx(
-        w_t.transpose(-2, -1).contiguous(),
+    w_scale, w_fp8 = to_mx(
+        w,
         elem_dtype=torch.float8_e4m3fn,
         block_size=block_size,
     )
-    w_t_scale, w_t_mx = w_scale.transpose(-2, -1), w_mx.transpose(-2, -1)
 
-    ref_out = torch._grouped_mm(x_ref, w_t_ref, offs=offs_ref, out_dtype=torch.bfloat16)
+    ref_out = torch._grouped_mm(
+        x_ref, w_ref.transpose(-2, -1), offs=offs_ref, out_dtype=torch.bfloat16
+    )
     out = _emulated_mxfp8_scaled_grouped_mm_2d_3d(
-        x_mx, x_scale, w_t_mx, w_t_scale, offs=offs, out_dtype=torch.bfloat16
+        x_fp8, x_scale, w_fp8, w_scale, offs=offs, out_dtype=torch.bfloat16
     )
 
     sqnr = compute_error(ref_out, out)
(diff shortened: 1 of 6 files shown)


danielvegamyhre:
@drisspg @vkuzo this is working for all tested cases now and is ready for review/land

torchao/prototype/moe_training/scaled_grouped_mm.py
```diff
@@ -402,12 +400,30 @@ def backward(ctx, grad_out: torch.Tensor):
 def _emulated_mxfp8_scaled_grouped_mm_2d_3d(
     A_mx: torch.Tensor,
     A_scale: torch.Tensor,
-    B_t_mx: torch.Tensor,
-    B_t_scale: torch.Tensor,
+    B_mx: torch.Tensor,
```
vkuzo, line 403: nit: `_mx` should be for a combination of raw data and scale, if `B_mx` is just the data then better to call it something else

torchao/prototype/mx_formats/utils.py
```diff
@@ -99,3 +99,76 @@ def _to_blocked_single(scales: Tensor) -> Tensor:
...
+        x_scales: Tensor with per group scales in blocked format concatenated into one tensor.
+        group_offs: Tensor of shape (num_groups,) which contains the end index of each group along the Mg dimension.
+        Mg: total size of all groups summed together
+        K: K dim size
+
+    Returns:
+        blocked_scales: Tensor
+        start_row_after_padding: Tensor of shape (num_groups,) which contains the start row after padding for each group.
+    """
+    from fbgemm_gpu.experimental.gemm.triton_gemm.fp4_quantize import _to_blocked
```
vkuzo, line 120: is this the same function as the one we have in torchao? https://github.com/pytorch/ao/blob/6f035e851f93e841f7307de3932b00ecf929461f/torchao/prototype/mx_formats/utils.py#L18

vkuzo (approved):
stamping since this is prototype

torchao/prototype/mx_formats/utils.py
danielvegamyhre: Will test if they're the same and replace if we can - I was having trouble getting the kernel working without CUDA errors so was trying to minimize differences between fbgemm unit test code and this torchao code path.

drisspg: should be the exact same

test/prototype/moe_training/test_scaled_grouped_mm.py
```diff
@@ -230,25 +230,26 @@ def compute_reference_forward(
 @pytest.mark.parametrize("num_experts", (1, 8, 16))
 def test_emulate_mxfp8_grouped_gemm_2d_3d(M, K, N, num_experts):
     x = torch.randn(M, K, dtype=torch.bfloat16, device="cuda")
-    w_t = torch.randn(num_experts, K, N, dtype=torch.bfloat16, device="cuda")
+    w = torch.randn(num_experts, N, K, dtype=torch.bfloat16, device="cuda")
     offs = generate_jagged_offs(num_experts, M)
```
drisspg, line 234: does this create random split lengths?

torchao/prototype/moe_training/kernels/mxfp8.py
```diff
@@ -0,0 +1,135 @@
...
+    B_fp8: torch.Tensor,
+    A_scales: torch.Tensor,
+    A_scales_blocked: torch.Tensor,
+    B_scales: torch.Tensor,
+    B_scales_blocked: torch.Tensor,
+    offs: torch.Tensor,
+    group_sizes: torch.Tensor,
+    starting_row_after_padding: torch.Tensor,
+):
+    logger.info(f"offs: {offs}, dtype: {offs.dtype}")
```
drisspg, line 111: nit: since these are format strings we always access these values even if logger is not enabled

test/prototype/moe_training/test_scaled_grouped_mm.py
danielvegamyhre: Random group offsets that are multiples of `multiple_of` param (default 16) in ascending sorted order. I should probably change the call to `multiple_of=32` here but I guess 16 works for these test cases.

torchao/prototype/moe_training/kernels/mxfp8.py
danielvegamyhre: ah, I always forget that. best practice to use `logger.info("offs: ", offs) rather than format strings to avoid this, is that right?

drisspg: https://realpython.com/python-f-strings/#lazy-evaluation-in-logging

danielvegamyhre: Will update

Issue #2833, opened 2025-08-21
[mxfp8 moe training] Add mxfp8 to FSDP and TP tests


danielvegamyhre:
Currently we have basic mxfp8 forward + backward supported in tochao scaled_grouped_mm with dynamic mxfp8 quantization: https://github.com/pytorch/ao/blob/main/test/prototype/moe_training/test_fsdp.py#L86

We need to add MXFP8 to these tests.

PR #2849, opened 2025-08-22
[mxfp8 moe training] Add mxfp8 to FSDP tests

liangel-02:
addressing #2833 

updating test to include mxfp8: `torchrun --nproc_per_node=${NUM_GPUS} -m pytest test_fsdp.py `, all tests pass

### DIFF
--- a/test/prototype/moe_training/test_fsdp.py
+++ b/test/prototype/moe_training/test_fsdp.py
@@ -51,8 +51,6 @@ except ImportError:
     (MoEScalingType.MXFP8, 28.0)
 ])
 def test_moe_float8_training_fsdp(recipe: MoEScalingType, min_out_sqnr: float):
-# def test_moe_float8_training_fsdp():
-    print(f"recipe: {recipe}")
     assert torch.cuda.is_available()
 
     # setup distributed for fsdp
@@ -88,7 +86,6 @@ def test_moe_float8_training_fsdp(recipe: MoEScalingType, min_out_sqnr: float):
             if target_fqn in cur_fqn:
                 return True
         return False
-    print("quantizing...")
     # quantize test model
     config = MoETrainingConfig(recipe)
     # config = MoETrainingConfig()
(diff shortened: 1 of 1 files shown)


danielvegamyhre:
@liangel-02 we need to use this torchtitan API:

```python
set_token_group_alignment_size_m(16) # fp8
```

or

```python
set_token_group_alignment_size_m(32) # mxfp8
```


This is because TMA (some background [here](https://pytorch.org/blog/hopper-tma-unit/)) requires the slowest moving dim (stride of 1) to be 128 bit (16 byte) aligned. 

In the backward pass, when grad_weight = grad_output_t @ input, the “M” dimension (flattened token groups) become this “stride 1” dim. Therefore, each token group must be 16 byte aligned for this grouped gemm. 

- For bf16, this means groups divisible by 8 elements (16 bytes / 2 bytes per elem).
- For fp8 groups divisible by 16 elements  (16 bytes / 1 byte per elem).  
- For mxfp8, I enforce they must be divisible by `block_size` (32) not because of TMA, but to ensure when we quantize along that dim, we have (1) no scaling groups crossing logically distinct tensor boundaries, and (2) no “partial” scaling group at the the end (e..g, 1x17 chunk because the tensor ends).

This might all be a bit confusing without background knowledge on GPU architecture and MoE models, but don't worry! Plenty of time to learn.

For now, just implement this setting in some clean/interpretable way in the test code and it should (hopefully) work.

test/prototype/moe_training/test_fsdp.py
```diff
@@ -83,7 +95,8 @@ def moe_module_filter_fn(mod: nn.Module, cur_fqn: str) -> bool:
         return False
 
     # quantize test model
-    config = MoETrainingConfig()
+    config = MoETrainingConfig(recipe)
+    # config = MoETrainingConfig()
```
danielvegamyhre: remove commented code before landing

danielvegamyhre (approved):
lgtm! 1 minor comment before landing, thanks

PR #2856, opened 2025-08-22
[moe training] add test case for shared expert in distributed tests

danielvegamyhre:
Stacked PRs:
 * __->__#2856

--- --- ---

[moe training] add test case for shared expert in distributed tests

## Summary
- After https://github.com/pytorch/torchtitan/pull/1561 landed in torchtitan, the hang when using quantized training techniques on MoEs is now resolved. We can add test cases for this now.

## Test plan
- `cd test/prototype/moe_training && ./test_everything.sh`


danielvegamyhre:
doing this as part of https://github.com/pytorch/ao/pull/2905 instead

PR #2860, opened 2025-08-22
Port metadata from the linear node onto the reference custom op for int4

kimishpatel:
* __->__ #2860

Summary:
Allow for numerical debugger in ExecuTorch use the from_node info for
correlation

Test Plan:
CI

Reviewers:

Subscribers:

Tasks:

Tags:


torchao/quantization/pt2e/reference_representation_rewrite.py
```diff
@@ -1153,12 +1184,15 @@ def reference_representation_rewrite(model: GraphModule) -> GraphModule:
...
             replacement.recompile()  # type: ignore[attr-defined]
-            replace_pattern_with_filters(
+            matches = replace_pattern_with_filters(
                 model,
                 pattern,
                 replacement,
                 match_filters=rewrite_info.filter_fn,
                 ignore_literals=rewrite_info.ignore_literals,
             )  # type: ignore[arg-type]
+            if rewrite_info.port_metadata_fn:
```
jerryzh168, line 1194: why is this not added to https://github.com/pytorch/ao/blob/main/torchao/quantization/pt2e/quantizer/port_metadata_pass.py?

kimishpatel: reference rewrite is an option pass. Or at least thats how I was using it in my debug. In reference rewrite the rules for porting metadata is not as generic, so you have to let each pattern dictate what they want to do

jerryzh168 (approved):
OK, I'm also wondering if we should move reference representation rewrite and port metadata etc. outside of torchao, since it seems that it's not used by most of torchao users, would it make more sense for this to be in ET?

kimishpatel:
> OK, I'm also wondering if we should move reference representation rewrite and port metadata etc. outside of torchao, since it seems that it's not used by most of torchao users, would it make more sense for this to be in ET?

Yeah it could. Not opposed to it. I do think it is useful for debugging especially when doing lowbit quant. Even when doing say fp4 perhaps? But if this doesnt seem of interest in forseeable future than I am find moving it out

jerryzh168:
we could add some logging and decide in a few months, my guess is we don't have people using these right now, at least the reference rewrite

PR #2861, opened 2025-08-22
Update IntxUnpackedTensor to support dynamic activation

metascroy:
This updates IntxUnpackedTensor to support dynamic activation, and allows v2 inside the Int8DynamicActivatinoIntxWeightConfig.

### DIFF
--- a/torchao/quantization/quantize_/workflows/intx/intx_unpacked_tensor.py
+++ b/torchao/quantization/quantize_/workflows/intx/intx_unpacked_tensor.py
@@ -19,7 +19,6 @@ from torchao.quantization.quant_primitives import (
 )
 from torchao.quantization.utils import _get_per_token_block_size
 from torchao.utils import (
-    TORCH_VERSION_AT_LEAST_2_5,
     TorchAOBaseTensor,
     fill_defaults,
 )
@@ -300,6 +299,5 @@ def _(func, types, args, kwargs):
 
 IntxUnpackedTensor.__module__ = "torchao.quantization"
 
-if TORCH_VERSION_AT_LEAST_2_5:
-    # Allow a model with IntxUnpackedTensor weights to be loaded with `weights_only=True`
-    torch.serialization.add_safe_globals([IntxUnpackedTensor])
+# Allow a model with IntxUnpackedTensor weights to be loaded with `weights_only=True`
+torch.serialization.add_safe_globals([IntxUnpackedTensor])
--- a/torchao/quantization/quantize_/workflows/intx/intx_unpacked_tensor.py
+++ b/torchao/quantization/quantize_/workflows/intx/intx_unpacked_tensor.py
@@ -17,7 +17,9 @@ from torchao.quantization.quant_primitives import (
     dequantize_affine,
     quantize_affine,
 )
+from torchao.quantization.utils import _get_per_token_block_size
 from torchao.utils import (
+    TORCH_VERSION_AT_LEAST_2_5,
     TorchAOBaseTensor,
     fill_defaults,
 )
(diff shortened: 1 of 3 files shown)


torchao/quantization/quantize_/workflows/intx/intx_unpacked_tensor.py
```diff
@@ -199,10 +216,18 @@ def _(func, types, args, kwargs):
...
+    assert isinstance(weight_tensor, IntxUnpackedTensor)
+
+    # Apply dynamic activation quant
+    if weight_tensor.apply_activation_quantization:
+        input_tensor = IntxUnpackedTensor.from_hp(
+            hp_tensor=input_tensor,
+            block_size=_get_per_token_block_size(input_tensor),
+            target_dtype=torch.int8,
+            mapping_type=MappingType.ASYMMETRIC,
+        ).dequantize()
```
jerryzh168: is this the only possible setting people will want?

torchao/quantization/quantize_/workflows/intx/intx_unpacked_tensor.py
```diff
@@ -199,10 +216,18 @@ def _(func, types, args, kwargs):
...
         args[2] if len(args) > 2 else None,
     )
-    if isinstance(input_tensor, IntxUnpackedTensor):
-        input_tensor = input_tensor.dequantize()
-    if isinstance(weight_tensor, IntxUnpackedTensor):
-        weight_tensor = weight_tensor.dequantize()
+    assert isinstance(weight_tensor, IntxUnpackedTensor)
+
+    # Apply dynamic activation quant
+    if weight_tensor.apply_activation_quantization:
```
jerryzh168: also probably rename to `apply_int8_act_asym_per_token_quant` to be more specific probably

torchao/quantization/quantize_/workflows/intx/intx_unpacked_tensor.py
jerryzh168, line 1: nit: also this file and the tensor subclass should be renamed to intx_unpacked_to_int8_tensor to match the packing format name I think

metascroy: Renaming will make reviewing harder.  Do you want it renamed in this PR or a follow-up PR?

torchao/quantization/quantize_/workflows/intx/intx_unpacked_tensor.py
metascroy: This is what Arm/XNNPACK/ExecuTorch need.

Not sure if someone will come in the future an want something else.

torchao/quantization/quantize_/workflows/intx/intx_unpacked_tensor.py
metascroy: Actually github seems to show when files are renamed.  I'll go ahead an update this PR with the renaming

torchao/quantization/quantize_/workflows/intx/__init__.py
```diff
@@ -1,5 +1,5 @@
-from .intx_unpacked_tensor import IntxUnpackedTensor
+from .intx_unpacked_to_int8_tensor import IntxUnpackedToInt8Tensor
```
jerryzh168: we don't need this file I think, we can import from `.intx.intx_unpacked_to_int8_tensor` directly

torchao/quantization/quantize_/workflows/intx/intx_unpacked_to_int8_tensor.py
```diff
@@ -54,12 +55,27 @@ class IntxUnpackedTensor(TorchAOBaseTensor):
...
 
     tensor_data_names = ["qdata", "scale", "zero_point"]
-    tensor_attribute_names = ["target_dtype", "block_size", "dtype"]
-
-    def __new__(cls, qdata, scale, zero_point, target_dtype, block_size, dtype):
+    tensor_attribute_names = [
+        "target_dtype",
+        "block_size",
+        "dtype",
+        "apply_int8_act_asym_per_token_quant",
```
jerryzh168, line 66: nit: make this arg optional and default to False?

metascroy: It is optional and defaulted to false in the class method from_hp.

__new__ and __init__ aren't really user facing, so I think it's better that everything be explicitly defined?

jerryzh168: I see, OK

jerryzh168 (approved):

PR #2866, opened 2025-08-24
Split implements and implements_torch_function

Krishn1412:
1) Added two registers _ATEN_OP_TABLE and _TORCH_FN_TABLE instead of one.

2) Split the decorator into two.

### DIFF
--- a/torchao/utils.py
+++ b/torchao/utils.py
@@ -392,39 +392,45 @@ and dispatching to these implementations.
 """
 
 
-def _implements(cls, aten_ops_or_torch_fns):
-    """Use this decorator to implement a function for an aten ops in __torch_dispatch__
-    (if user passed in a list of ops)
-    or torch function in __torch_function__ (if user passed in a single object)
-
-    class MyTensor(torch.Tensor):
-        ...
-        implements = classmethod(_implements)
-
-    implements = MyTensor.implements
-
-    @implements(torch.nn.functional.linear):
-    def _(func, types, args, kwargs):
-        ...
-
-    """
-    if not hasattr(cls, "_ATEN_OP_OR_TORCH_FN_TABLE"):
-        cls._ATEN_OP_OR_TORCH_FN_TABLE = {}
-
-    if cls not in cls._ATEN_OP_OR_TORCH_FN_TABLE:
-        cls._ATEN_OP_OR_TORCH_FN_TABLE[cls] = {}
-
-    if not isinstance(aten_ops_or_torch_fns, (list, tuple)):
-        aten_ops_or_torch_fns = [aten_ops_or_torch_fns]
+def _implements(cls, aten_ops):
+    """Decorator to implement aten ops for __torch_dispatch__."""
+    if not hasattr(cls, "_ATEN_OP_TABLE"):
+        cls._ATEN_OP_TABLE = {}
+    if cls not in cls._ATEN_OP_TABLE:
+        cls._ATEN_OP_TABLE[cls] = {}
+    if not isinstance(aten_ops, (list, tuple)):
+        aten_ops = [aten_ops]
 
     def decorator(func):
-        for op in aten_ops_or_torch_fns:
+        for op in aten_ops:
 
-            @functools.wraps(op)
-            def wrapper(f, types, args, kwargs):
-                return func(f, types, args, kwargs)
+            @functools.wraps(func)
+            def wrapper(f, types, args, kwargs, _func=func):
+                return _func(f, types, args, kwargs)
 
-            cls._ATEN_OP_OR_TORCH_FN_TABLE[cls][op] = wrapper
+            cls._ATEN_OP_TABLE[cls][op] = wrapper
+        return func
+
+    return decorator
+
+
+def _implements_torch_function(cls, torch_fns):
+    """Decorator to implement __torch_function__."""
+    if not hasattr(cls, "_TORCH_FN_TABLE"):
+        cls._TORCH_FN_TABLE = {}
+    if cls not in cls._TORCH_FN_TABLE:
+        cls._TORCH_FN_TABLE[cls] = {}
+    if not isinstance(torch_fns, (list, tuple)):
+        torch_fns = [torch_fns]
+
+    def decorator(func):
+        for fn in torch_fns:
+
+            @functools.wraps(func)
+            def wrapper(f, types, args, kwargs, _func=func):
+                return _func(f, types, args, kwargs)
+
+            cls._TORCH_FN_TABLE[cls][fn] = wrapper
         return func
 
     return decorator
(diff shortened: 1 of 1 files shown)


jerryzh168 (commented):
thanks! there are some more things to do I think.

(1). we'll need to modify all callsites that's using `implements` on some aten function to use `implements_torch_function` instead
(2). need to make sure these two things can compose,
e.g. https://github.com/pytorch/ao/blob/bc2c83e1a408c5682cd8debc7515eabbddd28fe2/torchao/quantization/quantize_/workflows/float8/float8_tensor.py#L232 is using both, so:

```
@implements(aten.linear)
@implements_torch_function(F.linear)
def _(...):
     ...
```

should work,  please add a test in https://github.com/pytorch/ao/blob/f03a737582b6a247fa86301678b4e9ebdd8fca57/test/test_utils.py#L89 to make sure

torchao/dtypes/affine_quantized_tensor_ops.py
```diff
@@ -262,9 +262,11 @@ def _register_aqt_quantized_linear_dispatches():
 _register_aqt_quantized_linear_dispatches()
 
 implements = AffineQuantizedTensor.implements
+implements_torch_function = AffineQuantizedTensor.implements_torch_function
 
 
-@implements([torch.nn.functional.linear, aten.linear.default])
+@implements([aten.linear.default])
```
jerryzh168, line 268: nit: btw, we don't need a list if it's a single op, i.e. we can do `@implements(aten.linear.default)`

test/test_utils.py
```diff
@@ -120,6 +121,48 @@ def __init__(self, qdata, attr, device=None):
...
+
+        # check they both point to the same function wrapper
+        aten_wrapper = TorchAOBaseTensor._ATEN_OP_TABLE[TorchAOBaseTensor][
+            torch.ops.aten.linear.default
+        ]
+        torchfn_wrapper = TorchAOBaseTensor._TORCH_FN_TABLE[TorchAOBaseTensor][F.linear]
+
+        # check if they wrap the same underlying function
+        self.assertEqual(aten_wrapper.__wrapped__, fake_linear)
+        self.assertEqual(torchfn_wrapper.__wrapped__, fake_linear)
```
jerryzh168: these are testing implementation details, I think it might be better to change these to test the user code instead, for example:
```
class MyTensor(TorchAOBaseTensor):
    ...
    # something like this https://github.com/pytorch/ao/blob/87769675a3e12209c4c30cd4e8563de7099d9d21/test/test_utils.py#L239-L252
    
counter = 0

def fake_linear(f, types, args, kwargs):
       x, w, b = args
       counter += 1
       return torch.matmul(x, w.T) + (b if b is not None else 0)

linear = torch.nn.Linear(...)
# swap the weight to MyTensor: https://github.com/pytorch/ao/blob/87769675a3e12209c4c30cd4e8563de7099d9d21/test/test_utils.py#L254-L256
linear.weight = MyTensor(...)

# trigger F.linear
linear(*example_inputs)

# check the value of counter to make sure the function called       

# trigger aten.linear
with torch.inference_mode():
    linear(*example_inputs)

# check the value of counter to make sure the function called       
```

jerryzh168:
seems like there is a conflict with main, please rebase

test/test_utils.py
```diff
@@ -340,6 +341,54 @@ def __init__(
...
+                counter["calls"] += 1
+                return torch.matmul(x, w_plain.T) + (b_plain if b_plain is not None else 0)
+
+            
+            l = torch.nn.Linear(2, 3)
+            orig_w = l.weight.detach().clone()
+            orig_b = l.bias.detach().clone() if l.bias is not None else None
+
+            p = torch.nn.Parameter(orig_w)
+            p.data = MyTensor(orig_w, "attr", None)
```
jerryzh168: nit: `p.data` is not a recommended API I think, just initialize the parameter with MyTensor would be better:

```
l.weight = torch.nn.Parameter(MyTensor(orig_w, "attr", None))
```

jerryzh168:
please fix ruff errors as well

jerryzh168:
please rebase will make sure to merge this soon to prevent new conflicts

jerryzh168:
looks like there are some issues with the rebase

Krishn1412:
Hey @jerryzh168, looks like I merged a lot of other commits, could you help me to resolve this mess please? TIA.

Krishn1412:
@jerryzh168 could you please merge it now. TIA.

torchao/prototype/quantization/dynamic_activation_lut/int8_dynamic_activation_lut_tensor.py
jerryzh168, line 1: can you remove this change?

Krishn1412:
@jerryzh168, its done. Sorry about the back and forth, it will be much more efficient going forward.

Krishn1412:
Hello @jerryzh168, can you please take a look at the tests when you have time. Seems to be some problem with contiguous call but I can't figure out how it is related to my change.

torchao/utils.py
```diff
@@ -867,12 +875,13 @@ def __init_subclass__(cls, **kwargs):
         # inherit the torch function and dispatch implementations from direct parent classes
         # e.g. for `class C(B, A)`, C.__bases__ == (B, A)
         for parent in cls.__bases__:
-            if parent in cls._ATEN_OP_OR_TORCH_FN_TABLE:
-                cls._ATEN_OP_OR_TORCH_FN_TABLE[cls].update(
-                    cls._ATEN_OP_OR_TORCH_FN_TABLE[parent]
-                )
+            if hasattr(cls, "_ATEN_OP_TABLE") and parent in cls._ATEN_OP_TABLE:
```
jerryzh168, line 877: nit: we don't need `hasattr(cls, "_ATEN_OP_TABLE")` check since it should be assigned in L860-L867 right?

jerryzh168: actually I feel the code here might be a bit problematic, shouldn't this be checking and calling `parent._ATEN_OP_TABLE` instead?

jerryzh168: this can be fixed in a separate though

jerryzh168:
for test failure, I think the issue might be this function: https://github.com/pytorch/ao/blob/b3b545fda4b860d5c033b30b071c9cfbf4e392e8/torchao/utils.py#L503 should be `implements_function`

Krishn1412:
@jerryzh168, can run approve the tests.

Krishn1412:
Hey @jerryzh168, thanks for merging this. Are there any other open tickets you could recommend? TIA.

torchao/utils.py
jerryzh168: @Krishn1412 one quick follow up is this item I think, can you help add a test to test subclassing works now and make sure this is working properly?

Krishn1412: Sure thing @jerryzh168

PR #2868, opened 2025-08-25
Move CPU kernels out of experimental

metascroy:
This PR:

* Moves torchao/experimental/kernels -> torchao/csrc/cpu/torch_free_kernels.  This directory houses CPU kernels that do not depend on torch and are written with raw pointers.

* Moves torchao/experimental/ops -> torchao/csrc/cpu/shared_kernels.  This directory houses CPU kernels that are shared between ATen and ExecuTorch, and can be built for either runtime based on compile-time flags.

In addition, it:

* Fixes include paths and CMake files to make the above changes buildable.
* Adjusts setup.py

The PR is on the larger side, but most changes are not meaningful.  For review, I suggest focusing on the changes to setup.py.

Test plan: CI tests


torchao/csrc/cpu/shared_kernels/internal/parallel-aten-impl.h
```diff
@@ -19,10 +19,6 @@ void torchao::parallel_1d(const int64_t begin, const int64_t end, const F& f) {
   });
 }
 
-inline void torchao::set_num_threads(int num_threads) {
```
metascroy, line 22: Removing these because code in csrc should have a path to being torch-agnostic, and there are not plans to support set_num_threads.  Note this function is not actually used by the kernels, so we can just delete it.

setup.py
```diff
@@ -472,10 +485,22 @@ def get_extensions():
...
+
+    # Exclude C++ CPU sources that are built by CMake
+    cpu_cmake_sources = glob.glob(
+        os.path.join(extensions_dir, "cpu", "torch_free_kernels", "**", "*.cpp"),
+        recursive=True,
+    )
+    cpu_cmake_sources += glob.glob(
+        os.path.join(extensions_dir, "cpu", "shared_kernels", "**", "*.cpp"),
+        recursive=True,
+    )
```
kimishpatel, line 498: which are teh corresponding lines that are deleted? I expected some other parts of this file that were building old cpp files to be deleted

torchao/csrc/cpu/shared_kernels/benchmarks/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
kimishpatel, line 1: why is this file seems created and not moved?

kimishpatel (approved):
skimmed through. left a couple of comments. You might have to change some BUCK /TARGETS files

setup.py
metascroy: The CMake builder now points to csrc instead of experimental (https://fburl.com/6lyvku2d).  So it isn't deleted, but just looks for the CMakeLists in a different file.

The logic here is to remove the cpu_cmake_sources from sources on line 498 because they are built with cmake later on.  Otherwise it will try to compile them without cmake.  And this doesn't compile because the appropriate flags are not defined.

torchao/csrc/cpu/shared_kernels/benchmarks/CMakeLists.txt
metascroy: It's moved from here: https://github.com/pytorch/ao/blob/main/torchao/experimental/ops/benchmarks/CMakeLists.txt

But enough changed from the original that it isn't showing as moved

metascroy:
> skimmed through. left a couple of comments. You might have to change some BUCK /TARGETS files

It doesn't look like the BUCK / TARGETS files are mirrored in OSS.

I also don't think csrc/cpu is mirrored in xplat (experimental was the only folder on torchao with mirroring set up), so I'm not sure if a lot of the buck tests will run anymore.

kimishpatel:
> > skimmed through. left a couple of comments. You might have to change some BUCK /TARGETS files
> 
> It doesn't look like the BUCK / TARGETS files are mirrored in OSS.
> 
> I also don't think csrc/cpu is mirrored in xplat (experimental was the only folder on torchao with mirroring set up), so I'm not sure if a lot of the buck tests will run anymore.

I think it will be lot easier to start this migration starting internally

setup.py
kimishpatel: Got it

jerryzh168:
what about the kernels from intel: 
```
da8w4_linear.cpp
int8_sdpa.cpp
scaled_embedding_bag.cpp
```

looks like they are in top level right now? should this be moved to a proper place?

PR #2874, opened 2025-08-25
TorchAOBaseTensor `__tensor_flatten__` and `__tensor_unflatten__` use…

jerryzh168:
… tensor attribute dict

Summary:
Previously we return tensor attribute list, but changing to tensor attribute dict is more debuggable

Test Plan:
python test/integration/test_load_and_run_checkpoint.py

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo (approved):
thank you!

PR #2876, opened 2025-08-25
fix ci import error 

liangel-02:


test/prototype/moe_training/test_fsdp.py
```diff
@@ -45,6 +45,12 @@
         "torchtitan not installed, skipping MoE tests.", allow_module_level=True
     )
 
+if torch.version.hip is not None:
```
danielvegamyhre: Can you add this to test_tp.py and test_fsdp_tp.py as well?

petrex (approved):
I see module level skip here. any plan to enable the test suite in the future?

danielvegamyhre:
> I see module level skip here. any plan to enable the test suite in the future?

We would love to have AMD support in the future, but have no concrete plan/timeline for it. We are currently focused on B200 and H100 related workflows for PTC in October. If you'd like to contribute AMD support for MoE training code it would be welcome!

PR #2879, opened 2025-08-26
[mxfp8 moe training] refactor all var names with suffix _mx to _data for clarity

danielvegamyhre:
Stacked PRs:
 * #2897
 * #2894
 * #2886
 * #2882
 * __->__#2879

--- --- ---

[mxfp8 moe training] refactor all var names with suffix _mx to _fp8 for clarity

### DIFF
--- a/torchao/prototype/moe_training/scaled_grouped_mm.py
+++ b/torchao/prototype/moe_training/scaled_grouped_mm.py
@@ -122,7 +122,7 @@ class _Float8GroupedMM(torch.autograd.Function):
             round_scales_to_power_of_2=True,
         )
         A_scaled = A.to(torch.float32) * A_scales
-        A_fp8_row_major = to_fp8_saturated(A_scaled, torch.float8_e4m3fn)
+        A_data_row_major = to_fp8_saturated(A_scaled, torch.float8_e4m3fn)
 
         # Convert B to float8, column-major for right operand of grouped GEMM.
         # B_t shape: (E, K, N)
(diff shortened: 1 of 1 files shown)


danielvegamyhre:
> sg, but if you're using `A_scales` I'd say the corresponding name is something like `A_data`, not `A_fp8`.

Yeah, hmm, scales and data are both different fp8 formats so I can see how this still could be unclear. The thing is, we already have `A` input tensor, so `A_data` vs `A` both sound like the same thing, and it gets confusing. I considered changing "A" to "A_hp" but A_hp vs A_data is still confusing. I figured "A" vs "A_fp8" was the clearest distinction, but A_hp vs A_data could also work I guess.

torchao/prototype/moe_training/scaled_grouped_mm.py
```diff
@@ -291,13 +291,13 @@ def forward(
         ctx.out_dtype = out_dtype
         ctx.emulated = emulated
 
-        # A_mx shape: (M, K)
+        # A_fp8 shape: (M, K)
```
vkuzo: I'm not a fan of `A_f8`, this isn't going to scale to other dtypes.  `A_data` is IMO better and more representative of the MX format being composed of raw data and scale.

vkuzo (approved):
stamping because prototype

torchao/prototype/moe_training/scaled_grouped_mm.py
danielvegamyhre: Alright I went ahead and changed _fp8 suffix to _data everywhere

Issue #2338, opened 2025-06-09
Distributing ao tensor subclasses in .safetensors checkpoints


mikaylagawarecki:
### Context 
The current status quo for distributing ao weights on huggingface is as checkpoints produced by `torch.save` ([examples here](https://huggingface.co/collections/pytorch/torchao-quantized-phi-4-mini-instruct-681566f123acc6fed345cb1a)). The reasoning is that by default, safetensors only supports saving state dictionaries of plain tensors, whereas ao weights tend to be tensor subclasses.

@drisspg and I had a conversation last week about what would be necessary for ao weights be distributed as `.safetensors` files rather than `.pt` files on huggingface. **My understanding from him is this *might* be necessary for checkpoints to be marked as official on huggingface**

 Since ao tensor subclasses are wrapper tensor subclasses that have plain tensors + metadata, it should theoretically be possible to decompose subclasses into tensors and metadata --> save tensors + metadata to a safetensors file (with special handling for non-json serializable metadata) with `safetensors.torch.save_file` --> have a helper on top of `safetensors.torch.load_file` that reconstructs the subclass.

I wrote a simple prototype of what this would looks like and am looking 
1. To stimulate discussion on what saving ao subclasses in safetensors format might look like/whether this is needed
2. For feedback on whether a solution like below covers the important cases/is a viable solution.

### Simple example
We took the more straightforward case of a [`LinearActivationQuantizedTensor`](https://github.com/pytorch/ao/blob/12398428a0573374c3971d9cb66d47185a11feff/torchao/quantization/linear_activation_quantized_tensor.py#L24-L119) with dynamic quantization to fbgemmfp8tensor, which I understand from Driss might be the main case to target.

My understanding here is that a `LinearActivationQuantizedTensor` would take
- `original_weight_tensor`: a plain tensor
- `input_quant_func`: in this case `to_fbgemm_fp8`, which maps to `FbgemmFp8Tensor.from_float`
- `quant_kwargs`: in this case, `activation_scale_ub`

Given this, I generated a **rudimentary** script that would

1. Extract tensor and non-tensor attributes from `LinearActivationQuantizedTensor`
2. Pass all information necessary to reconstruct the subclass to the [`metadata` argument](https://huggingface.co/docs/safetensors/en/api/torch#safetensors.torch.save_file) to `safetensors.torch.save_file` 
[...]

PR #2881, opened 2025-08-26
safetensors support 

liangel-02:
### Context

Currently we serialize and distribute torchao quantized model with [pytorch native APIs](https://docs.pytorch.org/ao/main/serialization.html), specifically `torch.save(model.state_dict())`and `model.load_state_dict(state_dict, assign=True)`

### Summary

This PR builds out the functionality `save_tensor_subclass_dict` and `load_tensor_subclass_dict` using `save_file` and `load_file` from safetensors library for FP8 based off of the script [here](https://github.com/pytorch/ao/issues/2338). 

### Test Plan 

Outputs match after saving/loading model.state_dict()


torchao/prototype/quantization/safetensor.py
```diff
@@ -0,0 +1,206 @@
...
+    combined_metadata["tensor_names"] = json.dumps(list(tensor_dict.keys()))
+
+    if additional_metadata:
+        combined_metadata.update(additional_metadata)
+
+    save_file(combined_tensors_dict, file_path, metadata=combined_metadata)
+    print(f"Saved {len(tensor_dict)} tensor subclasses to {file_path} with metadata")
+
+
+if __name__ == "__main__":
```
jerryzh168: this should be in a separate test file in https://github.com/pytorch/ao/tree/main/test/prototype

eg. test/prototype/safetensors/test_safetensors.py

torchao/prototype/quantization/safetensor.py
jerryzh168, line 1: nit: `safetensors` to match the library name

torchao/prototype/quantization/safetensor.py
```diff
@@ -0,0 +1,206 @@
...
+
+        if tensor_type == "Float8Tensor":
+            dtype = ALLOWED_QUANT_DTYPES.get(tensor_metadata.get("dtype"))
+            granularity = ALLOWED_GRANUALARITY.get(tensor_metadata.get("granularity"))
+            kernel_preference = json.loads(tensor_metadata.get("kernel_preference"))
+            hp_value_lb = tensor_metadata.get("hp_value_lb")
+            hp_value_ub = tensor_metadata.get("hp_value_ub")
+            mm_config = Float8MMConfig(*json.loads(tensor_metadata.get("mm_config")))
+
+            from torchao.quantization.quantize_.workflows import Float8Tensor
```
jerryzh168: nit: we should import from `torchao.quantization` so that we don't expose the internal implementation details

torchao/prototype/quantization/safetensors_support.py
```diff
@@ -0,0 +1,184 @@
...
+    combined_metadata = {}
+    combined_tensors_dict = {}
+
+    for tensor_name, tensor in tensor_dict.items():
+        if tensor.__class__.__name__ == "Tensor":
+            tensors_dict = {"data": tensor}
+            metadata = {"tensor_type": "Tensor"}
+            metadata["device"] = (
+                "cuda:0" if tensor.device == torch.device("cuda:0") else "cpu"
+            )
```
jerryzh168: can this just be `str(tensor.device)`

torchao/prototype/quantization/safetensors_support.py
```diff
@@ -0,0 +1,184 @@
...
+        tensor_dict: Dictionary of tensor subclasses to save, with keys as tensor names
+        file_path: Path where to save the tensors
+        additional_metadata: Optional additional metadata to include
+    """
+
+    combined_metadata = {}
+    combined_tensors_dict = {}
+
+    for tensor_name, tensor in tensor_dict.items():
+        if tensor.__class__.__name__ == "Tensor":
```
jerryzh168: maybe: `type(tensor) is torch.Tensor` to avoid the more private APIs

torchao/prototype/quantization/safetensors_support.py
```diff
@@ -0,0 +1,184 @@
...
+            elif value:
+                metadata[item] = json.dumps(value)
+
+    return metadata, tensors_dict
+
+
+def save_tensor_subclass_dict(
+    tensor_dict: Dict[str, Dict[str, torch.Tensor]],
+    file_path: str,
+    additional_metadata: Optional[Dict[str, str]] = None,
```
jerryzh168: is this used? we can remove if not I think

torchao/prototype/quantization/safetensors_support.py
```diff
@@ -0,0 +1,184 @@
...
+
+    Returns:
+        Tuple of (metadata, tensors_dict) where:
+        - metadata: Dictionary with metadata needed to reconstruct the tensor
+        - tensors_dict: Dictionary with tensors to save
+    """
+    metadata = {}
+    tensors_dict = {}
+
+    if tensor.__class__.__name__ == "Float8Tensor":
```
jerryzh168: same here, we can do:
```
from torchao.quantization import Float8Tensor
if isinstance(tensor, Float8Tensor):
    ...
```
I think

torchao/prototype/quantization/safetensors_support.py
```diff
@@ -0,0 +1,184 @@
...
+        metadata["tensor_type"] = "Float8Tensor"
```
jerryzh168: I feel we can do a full path here, might be more robust (torchao.quantization.Float8Tensor), also can we get this from tensor itself instead of hardcode? (not sure what is `tensor.__class__` but something like that might give the full path)

torchao/prototype/quantization/safetensors_support.py
```diff
@@ -0,0 +1,184 @@
...
+        for item, value in tensor.__dict__.items():
+            if isinstance(value, granularity.PerRow):
+                metadata["granularity"] = "PerRow"
+            elif isinstance(value, QuantizeTensorToFloat8Kwargs):
+                metadata["dtype"] = "torch.float8_e4m3fn"
```
jerryzh168: can we just use `item` for these?

jerryzh168: for `QuantizeTensorToFloat8Kwargs`, can we define a serialization for that, or just make this object serializable

torchao/prototype/quantization/safetensors_support.py
```diff
@@ -0,0 +1,184 @@
...
+            elif item == "qdata":
+                tensors_dict["qdata"] = value
+                metadata["qdata_device"] = (
+                    "cuda:0" if value.device == torch.device("cuda:0") else "cpu"
+                )
```
jerryzh168: is this needed? what happens if we just save the tensor directly

torchao/prototype/quantization/safetensors_support.py
```diff
@@ -0,0 +1,163 @@
...
+            result[tensor_name] = Float8Tensor(
+                qdata=tensor_tensors["qdata"],
+                scale=tensor_tensors["scale"],
+                block_size=json.loads(tensor_metadata.get("block_size")),
+                mm_config=mm_config,
+                hp_value_lb=act_quant_kwargs.hp_value_lb,
+                hp_value_ub=act_quant_kwargs.hp_value_ub,
+                act_quant_kwargs=act_quant_kwargs,
+                kernel_preference=act_quant_kwargs.kernel_preference,
+                dtype=act_quant_kwargs.float8_dtype,
```
jerryzh168: these value are not from `act_quant_kwargs`, these should be serialized on its own

torchao/prototype/quantization/safetensors_support.py
```diff
@@ -0,0 +1,163 @@
...
+                # Remove the prefix
+                tensor_tensors[key[len(tensor_name) + 1 :]] = value
+
+        tensor_type = tensor_metadata.get("tensor_type")
+
+        if tensor_type == str(Float8Tensor):
+            # not the same as mm_config in act_quant_kwargs
+            mm_config = json.loads(tensor_metadata.get("mm_config"))
+            if mm_config:
+                mm_config = Float8MMConfig(*mm_config)
```
jerryzh168: why do we need to initialize mm_config instead of just reading from whatever is serialized?

torchao/prototype/quantization/safetensors_support.py
```diff
@@ -0,0 +1,163 @@
...
+    combined_metadata = {}
+    combined_tensors_dict = {}
+
+    for tensor_name, tensor in tensor_dict.items():
+        if isinstance(tensor, Float8Tensor):
+            metadata, tensors_dict = create_metadata_for_tensor_subclass(tensor)
+        elif isinstance(tensor, torch.Tensor):
+            tensors_dict = {"data": tensor}
+            metadata = {"tensor_type": str(torch.Tensor)}
+            metadata["device"] = str(tensor.device)
```
jerryzh168: do we need to serialize device?

torchao/prototype/quantization/safetensors_support.py
```diff
@@ -0,0 +1,163 @@
...
+    tensor_dict: Dict[str, Dict[str, torch.Tensor]],
+    file_path: str,
+):
+    """
+    Save a dictionary of tensor subclasses with appropriate metadata.
+
+    Args:
+        tensor_dict: Dictionary of tensor subclasses to save, with keys as tensor names
+        file_path: Path where to save the tensors
+        additional_metadata: Optional additional metadata to include
```
jerryzh168: nit: remove?

torchao/prototype/quantization/safetensors_support.py
```diff
@@ -0,0 +1,163 @@
...
+        if tensor_type == str(Float8Tensor):
+            # not the same as mm_config in act_quant_kwargs
+            mm_config = json.loads(tensor_metadata.get("mm_config"))
+            if mm_config:
+                mm_config = Float8MMConfig(*mm_config)
+
+            act_quant_kwargs_dict = json.loads(tensor_metadata.get("act_quant_kwargs"))
+            act_quant_kwargs = config_from_dict(act_quant_kwargs_dict)
+
+            result[tensor_name] = Float8Tensor(
```
jerryzh168: wondering for these, can we just have:
1. a tensor data dict mapping from {tensor_data_attr_name: tensor}
2. a tensor attribute dict mapping from {tensor_attr_name: non-Tensor attributes}

and use 
```
Float8Tensor(
  **tensor_data_dict,
  **tensor_attribute_dict,
)
```
to create the Float8Tensor?

torchao/prototype/quantization/QuantizeTensorToFloat8KwargsJSON.py
```diff
@@ -0,0 +1,56 @@
...
+        return json_dict
+
+
+def config_from_dict(data: Dict[str, Any]) -> QuantizeTensorToFloat8Kwargs:
+    """
+    Create QuantizeTensorToFloat8Kwargs instance from a dictionary.
+    """
+    saved_mm_config = json.loads(data.get("mm_config"))
+    if saved_mm_config:
+        saved_mm_config = Float8MMConfig(*saved_mm_config)
```
jerryzh168: why do we need to do this here?

torchao/prototype/quantization/QuantizeTensorToFloat8KwargsJSON.py
```diff
@@ -0,0 +1,56 @@
...
+
+    saved_granularity = ALLOWED_GRANUALARITY.get(data.get("granularity"))
+    if not saved_granularity:
+        saved_granularity = PerRow()
```
jerryzh168: also this, I think we should not have these

torchao/prototype/quantization/safetensors_support.py
```diff
@@ -0,0 +1,163 @@
...
+        Tuple of (metadata, tensors_dict) where:
+        - metadata: Dictionary with metadata needed to reconstruct the tensor
+        - tensors_dict: Dictionary with tensors to save
+    """
+    metadata = {}
+    tensors_dict = {}
+
+    if isinstance(tensor, Float8Tensor):
+        metadata["tensor_type"] = str(tensor.__class__)
+        for item, value in tensor.__dict__.items():
```
jerryzh168: also not sure if we want to save everything here, can you check what is in `__dict__` and if it contains more than we need

torchao/prototype/quantization/safetensors_JSONEncoders.py
```diff
@@ -0,0 +1,175 @@
...
+    "torchao.quantization.quantize_.workflows",
+}
+
+
+class Float8TensorAttributeJSONEncoder(json.JSONEncoder):
+    def default(self, o):
+        if isinstance(o, Float8Tensor):
+            tensor_attr_dict = {}
+
+            for tensor_attribute_name in o.optional_tensor_attribute_names:
```
jerryzh168: to generalize, both `tensor_attribute_names` and `optional_tensor_attribute_names` should be handled

torchao/prototype/quantization/safetensors_JSONEncoders.py
```diff
@@ -0,0 +1,175 @@
...
+from torchao.quantization.granularity import PerRow
+from torchao.quantization.quantize_.workflows.float8.float8_tensor import (
+    QuantizeTensorToFloat8Kwargs,
+)
+
+ALLOWED_QUANT_DTYPES = {
+    "torch.float8_e4m3fn": torch.float8_e4m3fn,
+    # add to me
+}
+ALLOWED_GRANUALARITY = {"PerRow()": PerRow()}
```
jerryzh168: nit: these are no longer used

torchao/prototype/quantization/safetensors_JSONEncoders.py
```diff
@@ -0,0 +1,175 @@
...
+    "torchao.dtypes",
+    "torchao.prototype.awq",
+    "torchao.quantization.quantize_.common",
+    "torchao.quantization.quantize_.workflows",
+}
+
+
+class Float8TensorAttributeJSONEncoder(json.JSONEncoder):
+    def default(self, o):
+        if isinstance(o, Float8Tensor):
```
jerryzh168: we can add an assert for now

torchao/prototype/quantization/safetensors_JSONEncoders.py
```diff
@@ -0,0 +1,175 @@
...
+            # if it can't handle the type
+            result = self.default(value)
+            return result
+        except TypeError:
+            pass
+
+        return value
+
+
+def config_from_dict(data: Dict[str, Any]) -> Float8Tensor:
```
jerryzh168: this should not be `config_from_dict` I think, seems like `object_from_dict` is more appropriate

torchao/prototype/quantization/safetensors_JSONEncoders.py
liangel-02: since it's recursive it won't always be Float8Tensor (assertion will fail)

torchao/prototype/quantization/safetensors_support.py
```diff
@@ -0,0 +1,106 @@
...
+    )
+    return result
+
+
+def save_tensor_subclass_dict(
+    tensor_dict: Dict[str, Dict[str, torch.Tensor]],
+    file_path: str,
+):
+    """
+    Save a dictionary of tensor subclasses with appropriate metadata.
```
jerryzh168: might be helpful to write down the end result, what is saved in file.

also for `load_tensor_subclass_dict`, it would be good to write down what is loaded as well I think

torchao/prototype/quantization/safetensors_support.py
```diff
@@ -0,0 +1,106 @@
...
+            raise ValueError(f"Unsupported tensor type: {tensor_type}")
+
+    print(
+        f"Loaded {len(tensor_names)} tensor subclasses from {file_path} with metadata"
+    )
+    return result
+
+
+def save_tensor_subclass_dict(
+    tensor_dict: Dict[str, Dict[str, torch.Tensor]],
```
jerryzh168: nit: we can name this `state_dict`, to align with what people are used to

torchao/prototype/quantization/safetensors_serialization.py
```diff
@@ -0,0 +1,160 @@
...
+
+        if isinstance(o, torch.dtype):
+            return {"_type": "torch.dtype", "_data": str(o).split(".")[-1]}
+
+        if isinstance(o, enum.Enum):
+            # Store the full path for enums to ensure uniqueness
+            return {"_type": f"{o.__class__.__name__}", "_data": o.name}
+
+        if isinstance(o, list):
+            return [self.encode_value(item) for item in o]
```
jerryzh168: nit: we can handle dict as well

torchao/prototype/quantization/safetensors_serialization.py
```diff
@@ -0,0 +1,160 @@
...
+    "torchao.quantization",
+    "torchao.dtypes",
+    "torchao.quantization.quantize_.common",
+    "torchao.quantization.quantize_.workflows",
+}
+
+
+class Float8TensorAttributeJSONEncoder(json.JSONEncoder):
+    def default(self, o):
+        if isinstance(o, Float8Tensor):
```
jerryzh168: to generalize this in future PRs, please take a look at docs for TorchAOBaseTensor: https://github.com/pytorch/ao/blob/3bf21d0982f8beee2063999eaaa804084ca8110e/torchao/utils.py#L771, that should help with how to implement the generalization I think.

test/prototype/safetensors/test_safetensors.py
```diff
@@ -0,0 +1,47 @@
...
+    def test_safetensors(self):
+        config = Float8DynamicActivationFloat8WeightConfig(granularity=PerRow())
+        model = torch.nn.Sequential(
+            torch.nn.Linear(32, 256, dtype=torch.bfloat16, device="cuda")
+        )
+        quantize_(model, config)
+        example_inputs = (torch.randn(2, 32, dtype=torch.bfloat16, device="cuda"),)
+        ref_output = model(*example_inputs)
+
+        save_tensor_state_dict(model.state_dict(), "fp8_weights.safetensors")
```
drisspg: we should use tmp file if this is saved to disk

torchao/prototype/quantization/safetensors_support.py
```diff
@@ -0,0 +1,138 @@
...
+
+        if tensor_type == Float8Tensor.__name__:
+            tensor_metadata["_data"].update(tensor_tensors)
+            result[tensor_name] = object_from_dict(tensor_metadata)
+        elif tensor_type == torch.Tensor.__name__:
+            result[tensor_name] = tensor_tensors["_data"]
+        else:
+            raise ValueError(f"Unsupported tensor type: {tensor_type}")
+
+    print(
```
drisspg: nit: probably shoulnd't print, but can log

torchao/prototype/quantization/safetensors_support.py
```diff
@@ -0,0 +1,138 @@
...
+            for key, value in tensors_dict.items()
+        }
+
+        combined_metadata[tensor_name] = metadata
+        combined_tensors_dict.update(prefixed_tensors_dict)
+
+    combined_metadata["tensor_names"] = json.dumps(list(tensor_dict.keys()))
+
+    save_file(combined_tensors_dict, file_path, metadata=combined_metadata)
+    print(f"Saved {len(tensor_dict)} tensor subclasses to {file_path} with metadata")
```
drisspg: ditto

torchao/prototype/quantization/safetensors_serialization.py
jerryzh168, line 1: nit: I think we should create a `safetensors` or `safetensors_support` folder and put all safetensor related things there

test/prototype/safetensors/test_safetensors.py
jerryzh168, line 1: also this path should match the code file path, e.g. `test/prototype/quantization/safetensors/test_safetensors_support.py`

torchao/prototype/quantization/safetensors_support.py
```diff
@@ -0,0 +1,143 @@
...
+    combined_tensors_dict = {}
+
+    for tensor_name, tensor in tensor_dict.items():
+        if isinstance(tensor, Float8Tensor):
+            tensors_dict = {}
+            for tensor_data_name in tensor.tensor_data_names:
+                tensors_dict[tensor_data_name] = getattr(tensor, tensor_data_name)
+
+            metadata = json.dumps(tensor, cls=Float8TensorAttributeJSONEncoder)
+        elif isinstance(tensor, torch.Tensor):
```
mikaylagawarecki: Did you mean to check the type here
```suggestion
        elif type(tensor) is torch.Tensor:
```

Otherwise, I think every other ao tensor subclass other than `Float8Tensor` will be going through this path rather than erroring as unsupported. 

Please also add a test to verify unsupported types error as expected

torchao/prototype/quantization/safetensors_serialization.py
```diff
@@ -0,0 +1,163 @@
...
+    def encode_value(self, value):
+        """Helper method to recursively encode a value"""
+        # Try to use default for custom type
+        try:
+            # This will handle all our special cases and raise TypeError
+            # if it can't handle the type
+            result = self.default(value)
+            return result
+        except TypeError:
+            pass
```
mikaylagawarecki: n00b q: why is it ok to `pass` on the `TypeError` here?

Separately, it seems like we might potentially recurse infinitely between `encode_value`/`default` if there are circular references, so we might eventually need some sort of memo if that happens, but I doubt you will run into this so it's likely ok for this PR

torchao/prototype/quantization/safetensors_serialization.py
```diff
@@ -0,0 +1,163 @@
...
+
+    # Try to find the class in any of the allowed modules
+    cls = None
+    for module_path in ALLOWED_AO_MODULES:
+        try:
+            module = importlib.import_module(module_path)
+            cls = getattr(module, type_path)
+            break  # Found the class, exit the loop
+        except (ImportError, AttributeError):
+            continue  # Try the next module
```
mikaylagawarecki: Sorry if I missed this in one of the earlier comments, why do we do it like this rather than saving the `__module__` of classes? Is the purpose to preserve backward compatibility even if classes are moved around?

torchao/prototype/quantization/safetensors_serialization.py
```diff
@@ -0,0 +1,163 @@
...
+
+ALLOWED_AO_MODULES = {
+    "torchao.quantization",
+    "torchao.dtypes",
+    "torchao.quantization.quantize_.common",
+    "torchao.quantization.quantize_.workflows",
+}
+
+
+class Float8TensorAttributeJSONEncoder(json.JSONEncoder):
```
mikaylagawarecki: nice!

torchao/prototype/quantization/safetensors_serialization.py
```diff
@@ -0,0 +1,163 @@
...
+            return {
+                "_type": o.__class__.__name__,
+                "_data": data_dict,
+            }
+
+        if isinstance(o, torch.dtype):
+            return {"_type": "torch.dtype", "_data": str(o).split(".")[-1]}
+
+        if isinstance(o, enum.Enum):
+            # Store the full path for enums to ensure uniqueness
```
mikaylagawarecki: What does this comment mean?

torchao/prototype/quantization/safetensors_serialization.py
mikaylagawarecki: While you only really need a limited subset of classes, given the `ALLOWED_AO_MODULES`, the "untrusted user input" (contents of the checkpoint) can technically cause the code to create anything in the whole module thats allowlisted (which is a lot of things e.g. https://github.com/pytorch/ao/blob/main/torchao/quantization/__init__.py)

I'm not a security expert, but iiuc the purpose of moving away from `torch.save` appears to be to give more legitimacy to the safety of the checkpoint. I do think there is a tricky/fine line here where we still need to do some limited amount of "code execution instructed by contents of the checkpoint" on top of safetensors here (i.e. creating the tensor subclasses/enums). So I think we should also be careful to constrain the "surface of attack" here

While the `import/getattr` loop is convenient I think it might be best to avoid this.


Using an explicit allowlist (to constrain the surface of attack) might be preferable here, wdyt?

```
ALLOWED_CLASSES = {
    "Float8Tensor": torchao.quantization.Float8Tensor,
    "SomeConfig": torchao.dtypes.SomeConfig,
    # Only explicitly allowed classes
}
```

cc @jerryzh168 as I'm not sure whether in the end state we just need a limited subset of classes or everything in the module

torchao/prototype/quantization/safetensors_serialization.py
liangel-02: to address your first comment, for primitive types where they don't match any custom encoding we've defined, we call `super().default(o)`. this always raises a `TypeError` so catching and passing on it lets us return the value as is

torchao/prototype/quantization/safetensors_serialization.py
liangel-02: we want to store the full path to ensure uniqueness since sometimes different enum classes might have members w/ the same name

mikaylagawarecki: I see a nit: confusion I had while reading this was that full path sounds like it refers to `{o.__class__}.{o.__name__}` but we are storing `{o.__class__.__name__}`

torchao/prototype/quantization/safetensors_serialization.py
jerryzh168: yeah I think it makes sense to allowlist only specific classes that we need, we don't need everything in the module.

torchao/prototype/quantization/safetensors_serialization.py
jerryzh168: it will be handled by standard JSONEncoder according to https://github.com/pytorch/ao/blob/f0cca9982e09be1f3cd928c1bf8c0423d554e89a/torchao/core/config.py#L169-L170 but I'm not exactly sure if this true, maybe we can add a test for this to confirm?

torchao/prototype/quantization/safetensors_serialization.py
jerryzh168, line 1: also should be `prototype/safetensors_support/` instead of under the `prototype/quantization` I think, since safetensors support is for all torchao checkpoints

torchao/prototype/quantization/safetensors_serialization.py
liangel-02: i think our current test covers this logic

torchao/prototype/safetensors/safetensors_serialization.py
```diff
@@ -0,0 +1,160 @@
...
+    def default(self, o):
+        if isinstance(o, Float8Tensor):
+            tensor_attr_dict = {}
+            all_tensor_attributes = (
+                o.optional_tensor_attribute_names + o.tensor_attribute_names
+            )
+
+            for tensor_attribute_name in all_tensor_attributes:
+                attribute = getattr(o, tensor_attribute_name)
+                serialized_attribute = self.encode_value(attribute)
```
jerryzh168: nit: might be better to say `encoded_attribute` I think, serialization is an overloaded term

torchao/prototype/safetensors/safetensors_serialization.py
```diff
@@ -0,0 +1,160 @@
...
+import torchao
+from torchao.quantization import Float8Tensor
+from torchao.quantization.quantize_.common import KernelPreference
+from torchao.quantization.quantize_.workflows import QuantizeTensorToFloat8Kwargs
+
+ALLOWED_CLASSES = {
+    "Float8Tensor": torchao.quantization.Float8Tensor,
+    "Float8MMConfig": torchao.float8.inference.Float8MMConfig,
+    "QuantizeTensorToFloat8Kwargs": QuantizeTensorToFloat8Kwargs,
+    "PerRow": torchao.quantization.granularity.PerRow,
```
jerryzh168: I think we will have more, like "PerTensor" at least

also we should probably use `torchao.quantization.PerRow` to not expose implementation details

torchao/prototype/safetensors/safetensors_serialization.py
```diff
@@ -0,0 +1,160 @@
...
+import enum
+import json
+from typing import Any, Dict
+
+import torch
+
+import torchao
+from torchao.quantization import Float8Tensor
+from torchao.quantization.quantize_.common import KernelPreference
+from torchao.quantization.quantize_.workflows import QuantizeTensorToFloat8Kwargs
```
jerryzh168, line 11: looks like some imports are not used?

liangel-02: these are used to create the ALLOWED_CLASSES dict

mikaylagawarecki (approved):
Thanks for restricting to ALLOWED_CLASSES

torchao/prototype/safetensors/safetensors_serialization.py
```diff
@@ -0,0 +1,160 @@
...
+
+ALLOWED_CLASSES = {
+    "Float8Tensor": torchao.quantization.Float8Tensor,
```
jerryzh168: oh sorry, I meant this can be `Float8Tensor`

torchao/prototype/safetensors/safetensors_serialization.py
```diff
@@ -0,0 +1,160 @@
...
+    "Float8MMConfig": torchao.float8.inference.Float8MMConfig,
+    "QuantizeTensorToFloat8Kwargs": QuantizeTensorToFloat8Kwargs,
+    "PerRow": torchao.quantization.PerRow,
```
jerryzh168, line 17: also would need a "PerTensor", that is also a valid option for `Float8DynamicActivationFloat8WeightConfig`

PR #2883, opened 2025-08-26
Fix Float8Tensor quantize op kernrel preference dispatch

jerryzh168:
Stacked PRs:
 * #2884
 * __->__#2883

--- --- ---

### Fix Float8Tensor quantize op kernrel preference dispatch

Summary:
Previously if user specifies kernel_preference == "fbgemm", we'll use torch ops like `_choose_scale_float8` and `_quantize_affine_float8` to quantize the high precision Tensor
into a float8 Tensor

this PR makes sure we use fbgemm kernels when kernel_preference is "fbgemm", meaning: `torch.ops.triton.quantize_fp8_row` for per row, and `torch.ops.fbgemm.quantize_fp8_per_tensor`
for per tensor (while `torch.ops.fbgemm.quantize_fp8_per_tensor` has some issues right now and we'll enable later when it's fixed)

This doesn't have impact on BC, meaning old serialized model can still be loaded and run, only thing is fixing the kernel choice for fbgemm kernel preference
means users who requested FBGEMM kernelpreference now actually run fbgemm quantize op instead of torch op

Test Plan:
python test/quantization/quantize_/workflows/float8/test_float8_tensor.py -k test_expected_gpu_kernel_fbgemm

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/quantization/quantize_/workflows/float8/test_float8_tensor.py
+++ b/test/quantization/quantize_/workflows/float8/test_float8_tensor.py
@@ -10,6 +10,8 @@ from contextlib import nullcontext
 from typing import Tuple
 
 import torch
+from torch._inductor.utils import run_and_get_code
+from torch.testing import FileCheck
 from torch.testing._internal import common_utils
 from torch.testing._internal.common_utils import (
     run_tests,
@@ -85,6 +87,14 @@ class TestFloat8Tensor(TorchAOIntegrationTestCase):
         kernel_preference: KernelPreference,
         sizes: Tuple,
     ):
+        if (
+            isinstance(granularity, PerTensor)
+            and kernel_preference == KernelPreference.FBGEMM
+        ):
+            return unittest.skip(
+                "per tensor with fbgemm kernel preferece does not work yet"
+            )
+
         error_message = None
         if isinstance(granularity, PerRow):
             if mode == "dynamic" and dtype != torch.bfloat16:
@@ -237,7 +247,11 @@ class TestFloat8Tensor(TorchAOIntegrationTestCase):
         other_kernel_preferences = [
             KernelPreference.AUTO,
         ]
-        if _is_fbgemm_genai_gpu_available() and is_sm_at_least_90():
+        if (
+            _is_fbgemm_genai_gpu_available()
+            and is_sm_at_least_90()
+            and not isinstance(granularity, PerTensor)
+        ):
             other_kernel_preferences.append(KernelPreference.FBGEMM)
 
         quantized_outputs = {}
@@ -398,6 +412,32 @@ class TestFloat8Tensor(TorchAOIntegrationTestCase):
         granularity = PerRow()
         config = Float8DynamicActivationFloat8WeightConfig(granularity=granularity)
         self._test_moe_weight_reshape_ops(config)
+
+    # TODO: we have some other tests living in https://github.com/pytorch/ao/blob/4ecc89edd7b5cfc12e6f80854c85d04c472a0eb0/test/dtypes/test_affine_quantized_float.py#L743
+    # that should be moved here after v1 config is deprecated:
+    # https://github.com/pytorch/ao/issues/2649
+    @unittest.skipIf(not is_sm_at_least_90(), "Nedd sm90+")
+    def test_expected_gpu_kernel_fbgemm(self):
+        """Making sure KernelPreference.FBGEMM calls correct quantize and gemm kernels"""
+        torch.compiler.reset()
+
+        M, K, N = 128, 256, 512
+        m = torch.nn.Sequential(
+            torch.nn.Linear(K, N, device="cuda", dtype=torch.bfloat16)
+        )
+        config = Float8DynamicActivationFloat8WeightConfig(
+            granularity=PerRow(),
+            kernel_preference=KernelPreference.FBGEMM,
+        )
+        quantize_(m, config)
+        m = torch.compile(m)
+        x = torch.randn(M, K, device="cuda", dtype=torch.bfloat16)
+        out, code = run_and_get_code(m, x)
+
+        # check at least one occurrence of the quantize op and rowwise gemm op
+        FileCheck().check_count(
+            "torch.ops.triton.quantize_fp8_row.default", 1
+        ).check_count("torch.ops.fbgemm.f8f8bf16_rowwise.default", 1).run(code[0])
 
 
 common_utils.instantiate_parametrized_tests(TestFloat8Tensor)
--- a/torchao/quantization/quantize_/common/kernel_preference.py
+++ b/torchao/quantization/quantize_/common/kernel_preference.py
@@ -26,7 +26,7 @@ class KernelPreference(str, Enum):
     """
     TORCH = "torch"
 
-    """Use fbgemm quantize and quantized mm kernels, requires fbgemm_gpu_genai library
+    """Use quantize and quantized mm kernels from fbgemm_gpu_genai library, requires fbgemm_gpu_genai library
     """
     FBGEMM = "fbgemm"
 
--- a/torchao/quantization/quantize_/workflows/float8/float8_tensor.py
+++ b/torchao/quantization/quantize_/workflows/float8/float8_tensor.py
@@ -22,7 +22,7 @@ from torchao.float8.inference import (
     preprocess_data,
     preprocess_scale,
 )
-from torchao.quantization.granularity import PerRow
+from torchao.quantization.granularity import PerRow, PerTensor
 from torchao.quantization.observer import get_block_size
 from torchao.quantization.quant_primitives import (
     _choose_scale_float8,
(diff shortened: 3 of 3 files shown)


test/dtypes/test_affine_quantized_float.py
```diff
@@ -789,7 +789,7 @@ def test_expected_kernels_on_gpu(self, granularity, float8_config_version):
             # three triton kernels for quantizing the activation:
             # kernel 1: x_max_tmp = max(x, ...)
             # kernel 2: x_max = max(x_max_tmp)
-            # kernel 3: x_float8 = to_float8(x, x_max)
+            # kernel 3: x_float8 = Float8Tensor.from_hp(x, x_max)
```
vkuzo: this is pseudocode, not the actual function name

test/quantization/quantize_/workflows/float8/test_float8_tensor.py
```diff
@@ -255,6 +255,8 @@ def test_kernel_preference_numerical_equivalence(self, granularity, sizes):
         kp_and_res = list(quantized_outputs.items())
         for i in range(len(kp_and_res)):
             kp, res = kp_and_res[i]
+            print("kp:", kp)
```
drisspg: !

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
```diff
@@ -163,7 +163,7 @@ def dequantize(self, output_dtype: Optional[torch.dtype] = None) -> torch.Tensor
         return _dequantize_affine_float8(qdata, scale, output_dtype)
 
     @classmethod
-    def to_float8(
+    def from_hp(
```
drisspg: Whats the context behind the name change?

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
```diff
@@ -177,18 +177,29 @@ def to_float8(
...
-            )
+            and is_sm_at_least_90()
+            and isinstance(granularity, PerRow)
+            and float8_dtype == torch.float8_e4m3fn
+            and hp_value_lb is None
         ):
-            assert float8_dtype == torch.float8_e4m3fn, (
-                f"Only torch.float8_e4m3fn is supported, got: {float8_dtype}"
+            # optimized path for auto and per row quantization
+            kernel_choice = "triton"
```
vkuzo: let's make it clear what is "triton" and how it is related to `KernelPreference`

vkuzo (changes requested):

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
jerryzh168: This is from some previous comments from Vasiliy, but `Float8Tensor.from_hp` seems to make more sense compare to `Float8Tensor.to_float8`

vkuzo: seems like a separate change, would be good to split into a different PR

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
jerryzh168: this is in L180, but I can move it here

vkuzo: IMO we shouldn't have anything in here which is not in `KernelPreference`, it's too confusing.  Not confusing options would be either extending `KernelPreference`, or creating a new thing if you need more granularity.

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
jerryzh168: OK splitted to https://github.com/pytorch/ao/pull/2893

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
```diff
@@ -177,18 +177,30 @@ def to_float8(
...
+            and isinstance(granularity, PerRow)
+            and float8_dtype == torch.float8_e4m3fn
+            and hp_value_lb is None
         ):
-            assert float8_dtype == torch.float8_e4m3fn, (
-                f"Only torch.float8_e4m3fn is supported, got: {float8_dtype}"
+            # for per row quantization and kernel_preference auto setting
+            # we'll use triton quantize kernel for best performance
+            kernel_choice = "triton"
+        elif kernel_preference == KernelPreference.FBGEMM and hp_value_lb is None:
```
vkuzo: `elif kernel_preference == KernelPreference.FBGEMM and hp_value_lb is None:` should be changed to `elif kernel_preference == KernelPreference.FBGEMM`.  If the user asked for fbgemm, we should give them fbgemm.

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
jerryzh168: OK I added a `KernelPreference.TRITON` that use triton op for quantize and torch op for gemm, since there are no gemm ops for triton

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
jerryzh168: sounds good

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
```diff
@@ -202,7 +217,39 @@ def to_float8(
...
+                maybe_hp_value_ub_tensor = None
+            # not used
+            num_tokens = torch.empty([hp_tensor.size(0)], device=hp_tensor.device)
+            if isinstance(granularity, PerRow):
+                data, scale = torch.ops.fbgemm.quantize_fp8_per_row(
+                    hp_tensor, num_tokens, scale_ub=maybe_hp_value_ub_tensor
+                )
+            else:
+                assert isinstance(granularity, PerTensor), (
+                    f"Expected per tensor, got {granularity}"
```
drisspg, line 223: shoudln't we just error here?

drisspg: I dont love the silent fallbacks when users manually specify things

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
```diff
@@ -256,6 +303,9 @@ def _(func, types, args, kwargs):
                 kernel_choice = "fbgemm"
         elif weight_tensor.kernel_preference == KernelPreference.FBGEMM:
             kernel_choice = "fbgemm"
+        elif weight_tensor.kernel_preference == KernelPreference.TRITON:
+            # no triton gemm op is available, so we'll fallback to torch
+            kernel_choice = "torch"
```
drisspg: also weird, your `kernel choice` is doing double duty and now a recipe. that recipe is also not very clear from your initial doc

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
jerryzh168: OK, I can use error for this option for now

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
jerryzh168: do you mean kernel choice is used for both quantize and gemm?

what is the initial doc you are referring to?

drisspg: Yeah exactly, The doc is just the code block and reading on the kernel choice doc string

jerryzh168: > kernel choice is used for both quantize and gemm

yeah that's a decision we made before, according to Josh there is no need to have kernel level choice for now, just to keep things simple.

we did mention this in the KernelPreference doc I think?

torchao/quantization/quantize_/common/kernel_preference.py
```diff
@@ -30,5 +30,9 @@ class KernelPreference(str, Enum):
     """
     FBGEMM = "fbgemm"
 
+    """Use triton quantize and quantized mm kernels (if available), requires fbgemm_gpu_genai library, if no triton kernel for the quantize op or mm kernel is available, we'll fallback to torch ops
+    """
+    TRITON = "triton"
```
vkuzo: this name isn't coherent with the rest of the enum.  We already have an `FBGEMM` option which does not say anything about cutlass vs triton and therefore already includes these kernels.  I think you have two options:
1. have the fbgemm option pick the best kernel (cutlass vs triton) for the user.  I prefer this one.
2. make it clear that "FBGEMM" does not mean "FBGEMM", but really means "FBGEMM_CUTLASS", and also add "FBGEMM_TRITON".  I don't really like this option.

vkuzo (changes requested):

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
vkuzo: IMO it should be:
* fbgemm - use all fbgemm kernels, error out if something is not supported
* torch - use all torch kernels, error out if something is not supported
* auto - torchao decides what to do

we should not use torch kernels in the fbgemm setting, as that is not honoring what the user asked for

torchao/quantization/quantize_/common/kernel_preference.py
jerryzh168: OK thanks, yeah 1 seems easiest for now, will update to that. unless there is request to distinguish these in the future

test/quantization/quantize_/workflows/float8/test_float8_tensor.py
```diff
@@ -85,6 +85,14 @@ def test_fp8_linear_variants(
         kernel_preference: KernelPreference,
         sizes: Tuple,
     ):
+        if (
+            isinstance(granularity, PerTensor)
+            and kernel_preference == KernelPreference.FBGEMM
+        ):
```
drisspg, line 93: Nit: lets Xfail this

vkuzo:
> Previously we didn't handle kernel_preference == "fbgemm" properly for the quantize op,
this PR makes sure we dispatch to fbgemm kernels when kernel_preference is fbgemm

Can you explain specifically what did not work, and why it works after this PR?  It would also be good to have a test which fails before this PR and passes after this PR.

test/quantization/quantize_/workflows/float8/test_float8_tensor.py
jerryzh168: we are using unittest, seems like we can't do `return unittest.expectedFailure("...")`?

```
.../ao/test/quantization/quantize_/workflows/float8/test_float8_tensor.py", line 92, in test_fp8_linear_variants
    return unittest.expectedFailure(
  File ".../python3.10/unittest/case.py", line 148, in expectedFailure
    test_item.__unittest_expecting_failure__ = True
AttributeError: 'str' object has no attribute '__unittest_expecting_failure__'
```

but let me know if there is an example to do expectedFailure conditionally instead of skipping entire test

jerryzh168:
@vkuzo sure, updated the PR summary and added a test for this one and next PR as well

vkuzo:
> fixing the kernel choice for fbgemm kernel preference, which is supposed to be a developer facing API

note that I don't think this is true, all values of `KernelPreference` are user facing

test/quantization/quantize_/workflows/float8/test_float8_tensor.py
```diff
@@ -399,6 +413,29 @@ def test_moe_weight_reshape_ops(self):
         config = Float8DynamicActivationFloat8WeightConfig(granularity=granularity)
         self._test_moe_weight_reshape_ops(config)
 
+    def test_expected_gpu_kernel_fbgemm(self):
```
vkuzo, line 420: I think this test should be together with the other tests we have which check the same thing for other settings of this config, currently in `test_affine_quantized_float.py`.  Can we add a TODO to unify?

jerryzh168: yeah I think we can put everything here after we deprecate the AQT path in 9 months

jerryzh168:
> > fixing the kernel choice for fbgemm kernel preference, which is supposed to be a developer facing API
> 
> note that I don't think this is true, all values of `KernelPreference` are user facing

makes sense, it is user facing

PR #2884, opened 2025-08-26
Float8Tensor per row quantization pass bias to fbgemm kernel

jerryzh168:
Float8Tensor per row quantization pass bias to fbgemm kernel

Summary:
Previously bias is not passed to fbgemm kernel for float8 per row quant,
this PR adds it.

Difference is we should have a faster float8 per row quantized kernel, without changing
numerics or other things.

Test Plan:
```
python test/dtypes/test_affine_quantized_float.py -k test_expected_kernels_on_gpu
python test/quantization/quantize_/workflows/float8/test_float8_tensor.py
```

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/quantization/quantize_/workflows/float8/float8_tensor.py
+++ b/torchao/quantization/quantize_/workflows/float8/float8_tensor.py
@@ -300,6 +300,8 @@ def _(func, types, args, kwargs):
                 "Expected fbgemm_gpu_genai package to be installed"
             )
             assert is_sm_at_least_90(), "Expected SM90+ for fbgemm_gpu_genai"
+            mm_config = weight_tensor.mm_config
+            assert mm_config is not None
 
             out_shape = get_out_shape(input_tensor.shape, weight_tensor.shape)
             xq = input_tensor.qdata.reshape(-1, input_tensor.qdata.shape[-1])
@@ -315,6 +317,8 @@ def _(func, types, args, kwargs):
                     wq,
                     x_scale,
                     w_scale,
+                    bias=bias,
+                    use_fast_accum=mm_config.use_fast_accum,
                 ).reshape(out_shape)
             else:
                 assert _is_tensorwise_scaled(weight_tensor)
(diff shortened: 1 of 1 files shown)


vkuzo:
looks reasonable, can we add a test which covers the new path?

jerryzh168:
@drisspg yeah test is this one:

```python test/quantization/quantize_/workflows/float8/test_float8_tensor.py -k test_expected_gpu_kernel_fbgemm```

but it's probably better to have a more exhaustive one for all options (auto and torch) as well

it is a performance "bug" I think, the effect is that we are not using the most efficient path if there is a bias

File: test/quantization/quantize_/workflows/float8/test_float8_tensor.py [hunk 1]
```diff
@@ -435,9 +437,13 @@ def test_expected_gpu_kernel_fbgemm(self):
         out, code = run_and_get_code(m, x)
 
         # check at least one occurrence of the quantize op and rowwise gemm op
+        # check that there is no `triton_poi_fused_add_0` since the bias add should
+        # happen in the `f8f8bf16_rowwise.default` op instead of separately
         FileCheck().check_count(
             "torch.ops.triton.quantize_fp8_row.default", 1
-        ).check_count("torch.ops.fbgemm.f8f8bf16_rowwise.default", 1).run(code[0])
+        ).check_count("torch.ops.fbgemm.f8f8bf16_rowwise.default", 1).check_not(
+            "triton_poi_fused_add_0"
```
jerryzh168: cc @drisspg we explicitly test that `triton_poi_fused_add_0` is not generated, to make sure there is no additional `res + bias` there in this code path

vkuzo: it's not safe to depend on inductor generated kernel names such as `triton_poi_fused_add_0`, I think it would be better to ensure that there are no additional kernels called after `torch.ops.fbgemm.f8f8bf16_rowwise.default` is called.

jerryzh168: OK, updated

drisspg (approved):

PR #2886, opened 2025-08-27
[mxfp8 moe training] add per group blocked scale kernels

danielvegamyhre:
Stacked PRs:
 * #2897
 * #2894
 * __->__#2886

--- --- ---

[mxfp8 moe training] add per group blocked scale kernels for 2d input activations

## Summary
- We currently use a pytorch loop based impl for converting mxfp8 scales to block swizzled format on a per-group basis
- This of course is not optimal for perf and results in d2h sync 
- This PR implements a Triton kernel which does this conversion, without doing a d2h sync then looping on the host
- Note: to simplify the kernel, we pre-compute the start row of each group in the block padded output scales tensor (see `compute_per_group_blocked_scale_offsets`), but this is just a couple standard torch ops and shouldn't cause a d2h sync. There is probably still room for optimization here by doing this in the kernel somehow, but we'll take things one step at a time.

## Test plan
- `pytest test/prototype/moe_training/test_kernels.py  -k blocked`

## Performance
- Low memory bandwidth utilization but 14x faster than existing torch implementation
```
input_shape      torch_time_us    triton_time_us    torch_mem_bw_gbps    triton_mem_bw_gbps  triton_speedup
-------------  ---------------  ----------------  -------------------  --------------------  ----------------
(16640, 160)           866.848            60.416                6.261                89.831  14.35x
```


torchao/prototype/moe_training/kernels/mxfp8_blocked_scales.py
```diff
@@ -0,0 +1,253 @@
...
+    dest_indices = r_mod_32 * 16 + r_div_32 * 4 + col_offs
+
+    # Flatten
+    dest_indices_flat = tl.reshape(dest_indices, (BLOCK_ROWS * BLOCK_COLS))
+
+    # For this group and col block, we iterate through row blocks, reading (BLOCK_ROWS, BLOCK_COLS) from the input scales.
+    # We track how many row blocks we have iterated through.
+    block_row_id = 0
+    current_start_row = input_group_start_row
+    while current_start_row < input_group_end_row:
```
danielvegamyhre, line 216: Note for reviewer: I think we can probably do this without a loop, and just parallelize across row blocks as well (like in the original impl for dense models). Need to think about it some more.

drisspg: lets add as a follow up / todo

drisspg (approved):
can you make sure this is cudagraphable

torchao/prototype/moe_training/kernels/mxfp8_gemms.py
```diff
@@ -108,6 +110,9 @@ def _log_inputs(
     group_sizes: torch.Tensor,
     starting_row_after_padding: torch.Tensor,
 ):
+    if not DEBUG:
```
drisspg, line 116: I think you can just grab the logging state e.g something logging.Debug

danielvegamyhre: Yeah I tried this (make log level configurable and using logger.debug) and it wasn't working as expected so I just did this hack to refocus on the main work. Will add TODO

PR #2888, opened 2025-08-27
Support QAT int4 v1 path for BC

andrewor14:
**Summary:** `Int4WeightOnlyConfig` supports version 1 (targeting tinygemm) and version 2 (targeting fbgemm). However, the latter requires a new dependency (fbgemm_gpu_genai >= 1.2.0), which is problematic for torchao integrations with other frameworks. For now, we should continue to support the v1 path for BC.

**Test Plan:**
```
python test/quantization/test_qat.py -k test_infer_int4_weight_only_config
```


vkuzo:
I think this PR should also test the int4 workflow with version 1, can we add that?

andrewor14:
> I think this PR should also test the int4 workflow with version 1, can we add that?

added a test

test/quantization/test_qat.py
```diff
@@ -1933,6 +1934,22 @@ def test_quantize_api_fp8_int4(self):
...
+    @parametrize("version", [1, 2])
+    def test_quantize_api_int4(self, version: int):
+        """
+        Test the following:
+            quantize_(model, QATConfig(Int4WeightOnlyConfig(), step="prepare"))
+            quantize_(model, QATConfig(Int4WeightOnlyConfig(), step="convert"))
+        """
+        self._test_quantize_api_against_ptq(
+            Int4WeightOnlyConfig(version=version),
+            target_prepare_sqnr=12,
```
jerryzh168, line 1949: does this mean the prepare numerics does not match convert numerics very well? will this be an issue

jerryzh168 (approved):

PR #2897, opened 2025-08-28
[mxfp8 moe training] use dim1 cast cuda kernel in bwd

danielvegamyhre:
Stacked PRs:
 * __->__#2897

--- --- ---

[mxfp8 moe training] use dim1 cast cuda kernel in bwd

This way we can eliminate the .contiguous() call, as well as use the dim1 cast CUDA kernel which is the fastest option (vs inductor codgen or triton).

### DIFF
--- a/torchao/prototype/moe_training/scaled_grouped_mm.py
+++ b/torchao/prototype/moe_training/scaled_grouped_mm.py
@@ -20,7 +20,13 @@ from torchao.prototype.moe_training.kernels import (
 from torchao.prototype.moe_training.utils import (
     _is_column_major,
 )
+from torchao.prototype.mx_formats.config import (
+    MXFP8Dim1CastKernelChoice,
+    MXGemmKernelChoice,
+    ScaleCalculationMode,
+)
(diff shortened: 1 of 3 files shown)


torchao/prototype/moe_training/scaled_grouped_mm.py
```diff
@@ -20,6 +20,12 @@
 from torchao.prototype.moe_training.utils import (
     _is_column_major,
 )
+from torchao.prototype.mx_formats.config import (
+    MXFP8Dim1CastKernelChoice,
+    MXGemmKernelChoice,
+    ScaleCalculationMode,
+)
+from torchao.prototype.mx_formats.mx_linear import _to_mxfp8_dim1_kernel_wrapper
```
vkuzo: nit: move to utils file if you want to reuse it

vkuzo (approved):

torchao/prototype/moe_training/scaled_grouped_mm.py
danielvegamyhre: done

PR #2623, opened 2025-07-29
Disable register_da8w4_concat_linear_cpu_pass

henrylhtsang:
Summary: at

partial revert for now.

Differential Revision: D79130500

PR #2907, opened 2025-08-29
Fix FX Graph Cache issue in register_da8w4_concat_linear_cpu_pass

cyxlily:
Fix the bug that the FX Graph Cache was being bypassed when using the register_da8w4_concat_linear_cpu_pass, preventing cache hits on subsequent model runs.
Implement DA8W4ConcatLinearCPUPass that inherits from CustomGraphPass. Ensure it can be serialized and saved as fxgraph properly. Add the unit test. When saving fxgraph, the fxgraph_cache_bypass shuold remain at 0, confirming that the custom pass is no longer being rejected by the cache system.


jerryzh168:
it's disabled in https://github.com/pytorch/ao/pull/2623 because of internal failures, so we'll need to import this to make sure no internal failures I think

Xia-Weiwen:
> it's disabled in #2623 because of internal failures, so we'll need to import this to make sure no internal failures I think

Thanks

Xia-Weiwen:
Hi @jerryzh168 May I know the results of internal checks? Thanks.

test/quantization/test_da8w4_cpu.py
```diff
@@ -163,6 +163,11 @@ def forward(self, x):
             # ensure the expected op occurs only once in the code after fusion
             # The trailing "(" is to avoid matching the op in the comment
             assert code[0].count("torch.ops.torchao.da8w4_linear_cpu.default(") == 1
+
+            # ensure the custom DA8W4ConcatLinearCPUPass is properly cached as fxgraph
```
jerryzh168: how does the added test make sure `DA8W4ConcatLinearCPUPass` is properly cached?

jerryzh168:
I checked with @henrylhtsang and he says that this should fine

test/quantization/test_da8w4_cpu.py
Xia-Weiwen: It checks the counter. Do you think the comment here should be improved? Thanks.

jerryzh168: yeah, probably some explanation would be helpful, on how the cache works and why the counters would show that `DA8W4ConcatLinearCPUPass` is cached, I don't quite get it right now

cyxlily: How about change the comment to "# ensure the custom DA8W4ConcatLinearCPUPass is not bypassed when saving as fxgraph"?

Xia-Weiwen:
Hi @jerryzh168 How is the internal check going? BTW, the PR has been updated since your last import. Thanks.

test/quantization/test_da8w4_cpu.py
```diff
@@ -163,6 +163,11 @@ def forward(self, x):
             # ensure the expected op occurs only once in the code after fusion
             # The trailing "(" is to avoid matching the op in the comment
             assert code[0].count("torch.ops.torchao.da8w4_linear_cpu.default(") == 1
+
+            # ensure the custom DA8W4ConcatLinearCPUPass is not bypassed when saving as fxgraph
```
jerryzh168: can you update the comments to make it clearer what is being tested? why does this test the cache bypass?

test/quantization/test_da8w4_cpu.py
jerryzh168: still not clear what this means, can you add a more detailed explanation on how the caching works by default, what is the desired behavior for caching and how does the test makes sure the desired behavior occurs?

just adding the content of summary to the comment might be enough, and also make sure the variable naming is clearer

test/quantization/test_da8w4_cpu.py
```diff
@@ -172,6 +177,9 @@ def forward(self, x):
                 )
             assert torch.allclose(y, y_ref)
 
+            disable_fxgraph_cache_bypass = counters["inductor"]["fxgraph_cache_bypass"]
```
jerryzh168: I guess the variable naming is a bit confusing probably, I understand this is making sure the with the new pass, fx graph cache is not bypassing anything now

test/quantization/test_da8w4_cpu.py
cyxlily: Thanks. It's updated.

test/quantization/test_da8w4_cpu.py
cyxlily: Thanks. It's updated.

cyxlily:
@jerryzh168 Could you help to merge?

jerryzh168:
sorry I don't know how to merge this because I can't make the meta internal-only changes check pass, can you create a new PR?

jerryzh168:
merged now thanks @cyxlily @Xia-Weiwen

jerryzh168:
@cyxlily @Xia-Weiwen is it possible to do inductor changes only on your side? this PR has caused problems twice already, I don't think it's worth it to try to land it again

Xia-Weiwen:
Hi @jerryzh168 May I know what the new issue is?

Xia-Weiwen:
Hi @jerryzh168 I feel it is ok for us to enable the fusion pass in our script. Please feel free to revert it if needed. However, we are just curious about the issue you met and wondering if that means inductor fusion passes are not so friendly to extend or the CI should be improved somehow.

jerryzh168:
> Hi @jerryzh168 May I know what the new issue is?

similar issues as before I think, cache hit becomes 0 after this PR

jerryzh168:
> Hi @jerryzh168 I feel it is ok for us to enable the fusion pass in our script. Please feel free to revert it if needed. However, we are just curious about the issue you met and wondering if that means inductor fusion passes are not so friendly to extend or the CI should be improved somehow.

yeah will let you know if there is any followup afterwards

PR #2912, opened 2025-08-29
Add hqq support for Int4TilePackedTo4dTensor

jerryzh168:
Summary:
Enabled `use_hqq` option for Int4TilePackedTo4dTensor, instead of calling quant primitive ops for tinygemm to quantize the high precision tensor, the `use_hqq=True` path will quantize with `_choose_qparams_and_quantize_affine_hqq` that help improve accuracy for int4 weight only quantization, but still reuse the tinygemm kernel for speedup

Test Plan:
python test/quantization/quantize_/workflows/int4/test_int4_tile_packed_to_4d_tensor.py

Accuracy test (sanity check) to make sure hqq improves accuracy:
```
sh release.sh --model_id Qwen/Qwen3-8B --quants INT4 --push_to_hub

no hqq checkpoint: https://huggingface.co/jerryzh168/Qwen3-8B-INT4-non-hqq
hqq checkpoint: https://huggingface.co/jerryzh168/Qwen3-8B-INT4

export MODEL=jerryzh168/Qwen3-8B-INT4-non-hqq
export TASK=mmlu
lm_eval --model hf --model_args pretrained=$MODEL --tasks $TASK --device cuda:0 --batch_size auto

|      Groups      |Version|Filter|n-shot|Metric|   |Value |   |Stderr|
|------------------|------:|------|------|------|---|-----:|---|-----:|
|mmlu              |      2|none  |      |acc   |↑  |0.7019|±  |0.0036|
| - humanities     |      2|none  |      |acc   |↑  |0.6036|±  |0.0066|
| - other          |      2|none  |      |acc   |↑  |0.7403|±  |0.0076|
| - social sciences|      2|none  |      |acc   |↑  |0.8083|±  |0.0070|
| - stem           |      2|none  |      |acc   |↑  |0.7069|±  |0.0078|

export MODEL=jerryzh168/Qwen3-8B-INT4
lm_eval --model hf --model_args pretrained=$MODEL --tasks $TASK --device cuda:0 --batch_size auto

|      Groups      |Version|Filter|n-shot|Metric|   |Value |   |Stderr|
|------------------|------:|------|------|------|---|-----:|---|-----:|
|mmlu              |      2|none  |      |acc   |↑  |0.7040|±  |0.0036|
| - humanities     |      2|none  |      |acc   |↑  |0.5962|±  |0.0065|
| - other          |      2|none  |      |acc   |↑  |0.7470|±  |0.0075|
| - social sciences|      2|none  |      |acc   |↑  |0.8177|±  |0.0069|
| - stem           |      2|none  |      |acc   |↑  |0.7114|±  |0.0078|

hqq improves the accuracy for mmlu slightly.
```

Reviewers:

Subscribers:

Tasks:

Tags:


torchao/quantization/quantize_/workflows/int4/int4_tile_packed_to_4d_tensor.py
```diff
@@ -76,6 +84,7 @@ def from_hp(
         cls,
         hp_tensor: torch.Tensor,
         block_size: List[int],
+        use_hqq: bool = False,
```
mergennachin: have you thought of making it non binary but enum based? 

default
hqq

so that we can expand if necessary

jerryzh168: we haven't seen new updates to this one in that past year, I feel it's unlikely we will get more algorithms here.

vkuzo: the boolean feels hacky, IMO either a new method `from_hp_with_hqq` or an enum would be cleaner

jerryzh168: OK, I can add a str enum for this then, since both of you preferred that over the current state

torchao/quantization/quantize_/workflows/int4/int4_choose_qparams_algorithm.py
```diff
@@ -0,0 +1,20 @@
...
+
+from enum import Enum
+
+
+# can switch to StrEnum (https://docs.python.org/3/library/enum.html#enum.StrEnum)
+# after python 3.10 is end of life (https://devguide.python.org/versions/)
+class Int4ChooseQParamsAlgorithm(str, Enum):
+    """Variant of quantization algorithm to calculate scale and zero_point
+
+    * tinygemm: the choose qparams algorithm native for tinygemm kernel
```
vkuzo: can we describe what this actually does

jerryzh168: Sure, updated with some pseudo code describing the core logic

torchao/quantization/quant_api.py
```diff
@@ -1118,6 +1125,12 @@ def _int4_weight_only_quantize_tensor(weight, config):
 
     if config.version == 2:
         block_size = list(block_size)
+
```
mergennachin, line 1128: can you assert that "legacy" use_hqq and int4_choose_qparams_algorithm logic should match

mergennachin: Namely, if use_hqq=True then assert int4_choose_qparams_algorithm==HQQ

(We should still allow use_hqq=False and int4_choose_qparams_algorithm=FANCY_ALGORITHM for future use-casess)

jerryzh168: I'm planning to deprecate `use_hqq` flag for v2 config at some point actually, otherwise it might be confusing

jerryzh168: current state of `Int4WeightOnlyConfig` is a bit weird since both v1 and v2 config options are present, but we can deprecate v1 in 2 releases I think (in 0.15)

mergennachin: Yeah exactly, until we deprecate having that check would be good.

jerryzh168: OK sounds good, this would mean we have to leave `use_hqq` visible to people use v2 config as well, but sounds OK, we can always update the model card a bit later

jerryzh168: actually this will break BC for v1 configs as well

torchao/quantization/quant_api.py
```diff
@@ -1069,12 +1070,22 @@ class Int4WeightOnlyConfig(AOBaseConfig):
...
+         currently support TINYGEMM ("tinygemm") and HQQ ("hqq")
         `layout`: layout type for quantized tensor, default is `TensorCoreTiledLayout(inner_k_tiles=8)`
         `use_hqq`: whether to use hqq or default quantization mode, default is False
         `zero_point_domain`: data type of zeros points, choices are [ZeroPointDomain.FLOAT, ZeroPointDomain.INT, ZeroPointDomain.NONE]
         `set_inductor_config`: if True, adjusts `torchinductor` settings to recommended values.
         `preserve_zero`: whether to preserve zero, default is None. Will be set to True if zero_point_domain is ZeroPointDomain.INT
-        `packing_format`: the packing format for int4 tensor, available from version 2 and above
+        `version`: version of the config to use, only subset of above args are valid for version 1, and subset of above args are valid for version 2, default is 1, see note for more details
+
+    Note:
```
jerryzh168, line 1075: cc @mergennachin I added some more docs here, please let me know if this helps, only a subset of args will be used for each of the version right now

mergennachin: Documentation is good, but why not also do assertion? 

If ignored field are present, don't you wanna throw an exception to the developer?

mergennachin: Okay, talked offline with @jerryzh168 

The current approach seems fine. The increased complexity seems not worth it since we have to change the type to Optional.

jerryzh168:
I'm merging this now to unblock future PRs, please feel free to add more comment if there is anything else that should be fixed

PR #2913, opened 2025-08-30
SpinQuant rotate bias

rohansjoshi:
Summary: Added bias rotation. This is needed to apply SpinQuant R2 to models which have bias such as Qwen models.

Differential Revision: D81352249


andrewor14 (approved):
Thanks! Can you also add a quick unit test for these models? (can be separate PR)

jerryzh168:
I have the impression that spinquant implementation in torchao doesn't work before because it didn't improve accuracy, wondering if this would help fix the issue

previous accuracy test results can be found in https://github.com/pytorch/ao/pull/983 I think

PR #2924, opened 2025-09-02
Set seed in numeric tests to make them more reliable

metascroy:
This PR sets a random seed in the numeric tests to make them deterministic and more reliable.


test/prototype/test_dynamic_activation_lut.py
```diff
@@ -83,6 +83,7 @@ def run_before_and_after_tests():
 @pytest.mark.parametrize("lead_dim", [(5,), (2, 3)])
 @pytest.mark.skipif(not is_arm64_mac, reason="requires arm64 mac")
 def test_parq_conversion(dtype, granularity, bit_width, lead_dim):
+    torch.manual_seed(0)
```
jerryzh168, line 86: seems OK, but wondering if we migrate to unittest, if this is set automatically in some base class like TestCase.

metascroy: I'm not sure if it is or not...

PR #2928, opened 2025-09-03
[WIP]Add common int4 quantization support in torchao

orangeH25:


torchao/dtypes/affine_quantized_tensor.py
```diff
@@ -150,6 +154,18 @@ def dequantize(self, output_dtype: Optional[torch.dtype] = None) -> torch.Tensor
         elif isinstance(self._layout, Float8Layout):
             data, scale, _ = self.tensor_impl.get_plain()
             return _dequantize_affine_float8(data, scale, output_dtype)
+        elif isinstance(self._layout, Int4StorageQuantizationLayout):
```
jerryzh168, line 157: we changed our stack and Layout is no longer used, please check https://docs.pytorch.org/ao/main/quantization_overview.html and https://docs.pytorch.org/ao/main/contributor_guide.html

orangeH25: Thanks for the update! I'll check the new stack and follow the guidance

PR #2937, opened 2025-09-04
Improve QAT fp8-int4 numerics

andrewor14:
**Summary:** This commit improved the prepare vs convert SQNR of fp8-int4 QAT from 12 to 22. This is achieved by mimicking the numerics of the target FBGEMM fp8-int4 kernel more closely. In particular, FBGEMM first quantizes the weights to fp8, and then uses max abs values to compute the scale, which is significantly different from what torchao's quant primitives do.

**Test Plan:**
```
python test/quantization/test_qat.py -k test_fbgemm_fp8_primitives
python test/quantization/test_qat.py -k test_fbgemm_int4_primitives
python test/quantization/test_qat.py -k test_quantize_api_fp8_int4
```


torchao/quantization/qat/fake_quantize_config.py
```diff
@@ -77,6 +77,24 @@ def __post_init__(self):
             )
 
 
+@dataclass
+class Int4WeightFBGEMMFakeQuantizeConfig(FakeQuantizeConfigBase):
```
jerryzh168: nit: maybe not mention FBGEMM, but `Preshuffled` instead to allign with the tensor name

vkuzo:
can you explain what exactly is different in default path vs this path?

andrewor14:
> can you explain what exactly is different in default path vs this path?

Sure, there are a few main differences. The new path mimics the fbgemm numerics by:
1. Quantizing the weights first to fp8 and then to int4
2. Using max abs value to compute the scale (the old path uses max - min)
3. Using eps=1e-12 for fp8 quant and 1e-6 for int4 quant (old path uses `torch.finfo(dtype).eps`)

torchao/quantization/qat/fake_quantize_config.py
andrewor14: ok, should we use the following naming scheme?
```
Int4PreshuffledTensor -> Int4WeightPreshuffledFakeQuantizeConfig
Int4Tensor -> Int4WeightFakeQuantizeConfig (future PR)
```

jerryzh168: yeah, if the fake quant path uses the quant ops specific to Int4PreshuffledTensor (in from_hp), then I think we can just call it `Int4WeightPreshuffledFakeQuantizeConfig`

test/quantization/test_qat.py
```diff
@@ -1950,6 +1951,19 @@ def test_quantize_api_int4(self, version: int):
...
+    @unittest.skipIf(not _CUDA_IS_AVAILABLE, "skipping when cuda is not available")
+    def test_quantize_api_int8_int4(self):
+        """
+        Test the following:
+            quantize_(model, QATConfig(Int8DynamicActivationInt4WeightConfig(), step="prepare"))
+            quantize_(model, QATConfig(Int8DynamicActivationInt4WeightConfig(), step="convert"))
+        """
+        self._test_quantize_api_against_ptq(
+            Int8DynamicActivationInt4WeightConfig(group_size=32),
+            target_prepare_sqnr=30,
```
jerryzh168, line 1963: trying to understand why this is not inf, is this because fake quant does not do dtype conversion?

test/quantization/test_qat.py
```diff
@@ -2033,6 +2046,126 @@ def test_qat_nvfp4(self, use_per_tensor_scale: bool):
...
+        x1 = torch.randn([128, 256], dtype=torch.bfloat16).cuda()
+        x2 = copy.deepcopy(x1)
+        x3 = copy.deepcopy(x1)
+
+        # (1) Just call `quantize_int4_preshuffle`
+        (q1, (scale1, _)) = quantize_int4_preshuffle(x1, group_size, dtype="fp8")
+
+        # (2) Call `quantize_int4_preshuffle` but skip packing and shuffling
+        (q2, _) = quantize_fp8_row(x2)
+        (q2, scale2) = int4_row_quantize(q2, group_size)
```
jerryzh168, line 2121: is this what happens in `quantize_int4_preshuffle`? why do we first quantize to fp8 then to int4 instead of just quantizing to int4?

torchao/quantization/quant_primitives.py
```diff
@@ -2288,9 +2301,7 @@ def _quantize_affine_float8(
     tensor_scaled = tensor_fp32 / scale_expanded
     max_value = torch.finfo(float8_dtype).max
     tensor_clamped = tensor_scaled.clamp(min=-max_value, max=max_value)
-    if cast_to_float8_dtype:
-        tensor_clamped = tensor_clamped.to(float8_dtype)
-    return tensor_clamped
+    return _RoundToFloat8.apply(tensor_clamped, float8_dtype)
```
jerryzh168, line 2304: so fake quant also do casting? I thought we don't do casting during fake quant

test/quantization/test_qat.py
andrewor14: yeah it happens here, not sure the reason: https://github.com/pytorch/FBGEMM/blob/3ca2859adc0ae24b1214ccacedff24ea5fce9be5/fbgemm_gpu/experimental/gen_ai/gen_ai/quantize.py#L138

test/quantization/test_qat.py
andrewor14: not sure, but this is enough to recover significant accuracy degradation in most cases. I did verify the qparams dtypes are also matching but haven't investigated further, may continue this later separately

torchao/quantization/quant_primitives.py
andrewor14: I think fp8 numerics is close enough with bf16/fp32 that we can do this during QAT. This helps mimic the convert numerics a bit closer

andrewor14: fp8-int4 prepare vs convert sqnr drops from 22.375 to 19.25 without this

test/quantization/test_qat.py
```diff
@@ -2033,6 +2046,126 @@ def test_qat_nvfp4(self, use_per_tensor_scale: bool):
...
+
+        # (3) Reference implementation for QAT without the dequantize
```
jerryzh168, line 2123: for this, should we just initialize `Int4WeightPreshuffledFakeQuantizer` can call it here?

jerryzh168 (approved):
LGTM!

test/quantization/test_qat.py
jerryzh168: looks like this test is not related to the changes, since it's `Int8DynamicActivationInt4WeightConfig`

torchao/quantization/qat/fake_quantizer.py
```diff
@@ -93,13 +98,73 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        super().__init__()
+        self.config = config
+        torch._C._log_api_usage_once(
+            "torchao.quantization.qat.Int4WeightPreshuffledFakeQuantizer"
+        )
+
+    def forward(self, w: torch.Tensor) -> torch.Tensor:
+        return self._forward(w)[0]
+
+    def _forward(self, w: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
```
jerryzh168: didn't see this function being used elsewhere except for in L121, why do we need this?

andrewor14: oh sorry, we don't need this. This was leftover when I tried to write a test using this class

test/quantization/test_qat.py
andrewor14: I tried doing that but it's tricky because we don't want the dequantize steps here (but they're needed in the fake quantizer)

PR #2938, opened 2025-09-04
[CPU] Support int8 scaled embedding bag

shiyang-weng:


test/test_ops.py
```diff
@@ -784,14 +786,12 @@ def test_swizzle_mm():
...
 )
 @pytest.mark.parametrize(
-    "multi_hot, batch_size, vector_size, index_type",
+    "multi_hot, batch_size, vector_size, index_type, qtype",
     EMBEDINGBAG_TEST_PARAMS,
     ids=str,
 )
-def test_scaled_embedding_bag_cpu(multi_hot, batch_size, vector_size, index_type):
-    qtype = torch.float8_e4m3fn
+def test_scaled_embedding_bag_cpu(multi_hot, batch_size, vector_size, index_type, qtype):
```
Xia-Weiwen: It would be better to separate test cases for int8 and fp8 and use a helper function for the common part.

Xia-Weiwen (approved):

test/test_ops.py
LevelDownRefine: Done

metascroy:
As FYI, this PR conflicts with https://github.com/pytorch/ao/pull/2940, which is being landed from Phabricator right now

torchao/csrc/cpu/scaled_embedding_bag.cpp
```diff
@@ -11,19 +11,55 @@ namespace torchao {
 namespace {
 
 #if defined(CPU_CAPABILITY_AVX512)
```
metascroy: Move this file to torchao/csrc/cpu/aten_kernels

LevelDownRefine: Done

LevelDownRefine:
@jerryzh168 Could you help review this pr?

jerryzh168 (approved):
are these ops built by default?

LevelDownRefine:
> are these ops built by default?

Are not built by default

PR #2940, opened 2025-09-04
Move top-level CPU kernels to csrc/cpu/aten_kernels

metascroy:
This PR moves top-level CPU kernels in csrc/cpu to csrc/cpu/aten_kernels.  We now have 3 kernel directories under csrc/cpu:

* csrc/cpu/torch_free_kernels
* csrc/cpu/shared_kernels
* csrc/cpu/aten_kernels

We also add a readme that describes what each of these folders contains.


metascroy:
> does build still work? any changes needed to build script?

I'll let CI catch build issues

PR #2941, opened 2025-09-04
Fix xnnpack export

metascroy:
The previous version of IntxUnpackedToInt8Tensor had a reshape operator after export.  Although this does not change any numerics, it does interfere with XNNPACK's ability to recognize and lower the dynamic quantization pattern.

This PR removes the reshape operator, and adjusts the unit tests to look for 0 reshapes.

### DIFF
--- a/test/quantization/quantize_/workflows/intx/test_intx_unpacked_to_int8_tensor.py
+++ b/test/quantization/quantize_/workflows/intx/test_intx_unpacked_to_int8_tensor.py
@@ -23,7 +23,7 @@ from torchao.quantization import (
     MappingType,
     quantize_,
 )
-from torchao.quantization.granularity import PerGroup
+from torchao.quantization.granularity import PerAxis, PerGroup
 from torchao.quantization.qat import IntxFakeQuantizeConfig, QATConfig
 from torchao.quantization.quantize_.common import PackingFormat
 from torchao.quantization.utils import compute_error
@@ -156,7 +156,7 @@ class TestIntxUnpackedToInt8Tensor(TestCase):
             model,
             Int8DynamicActivationIntxWeightConfig(
                 weight_dtype=torch.int4,
-                weight_granularity=PerGroup(64),
+                weight_granularity=PerAxis(0),
                 weight_mapping_type=MappingType.SYMMETRIC,
                 packing_format=PackingFormat.UNPACKED_TO_INT8,
                 version=2,
--- a/test/quantization/quantize_/workflows/intx/test_intx_unpacked_to_int8_tensor.py
+++ b/test/quantization/quantize_/workflows/intx/test_intx_unpacked_to_int8_tensor.py
@@ -27,7 +27,7 @@ from torchao.quantization.granularity import PerGroup
 from torchao.quantization.qat import IntxFakeQuantizeConfig, QATConfig
 from torchao.quantization.quantize_.common import PackingFormat
 from torchao.quantization.utils import compute_error
-from torchao.utils import torch_version_at_least
+from torchao.utils import torch_version_at_least, unwrap_tensor_subclass
 
 
 @unittest.skipIf(not torch_version_at_least("2.7.0"), "Need pytorch 2.7+")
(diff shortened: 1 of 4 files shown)


torchao/quantization/quantize_/workflows/intx/intx_unpacked_to_int8_tensor.py
```diff
@@ -221,12 +263,7 @@ def _(func, types, args, kwargs):
 
     # Apply dynamic activation quant
     if weight_tensor.apply_int8_act_asym_per_token_quant:
```
metascroy: @jerryzh168 do you think we should make apply_int8_act_asym_per_token_quant an enum instead of a boolean.  It will make it easier to extend if we want in the future.

jerryzh168: yeah sure, if you anticipate more cases then it makes sense for it to be a enum I think

torchao/quantization/quantize_/workflows/intx/intx_unpacked_to_int8_tensor.py
```diff
@@ -207,6 +213,42 @@ def dequantize(self):
         )
 
 
+def _fake_apply_int8_act_asym_per_token_quant(hp_tensor):
```
jerryzh168: actually maybe a closer name for this is quantize_and_dequantize

to not confuse with fake_quant, which does not quantize to the real int values in quantize_affine

test/quantization/quantize_/workflows/intx/test_intx_unpacked_to_int8_tensor.py
```diff
@@ -169,17 +169,52 @@ def test_export_int8_dyn_act_intx_weight_config(self):
...
+                weight_dtype=torch.int4,
+                weight_granularity=PerGroup(64),
+                weight_mapping_type=MappingType.SYMMETRIC,
+                packing_format=PackingFormat.UNPACKED_TO_INT8,
+                version=2,
+            ),
+        )
+        eager_results = model(activations)
+
+        unwrap_tensor_subclass(model)
```
jerryzh168, line 203: is this still needed? we want to deprecate this util actually

metascroy: It's not needed, but there is a difference in the exported graph when it's present, so I want to test both with and without it for 0 reshapes.

torchao/quantization/quantize_/workflows/intx/intx_unpacked_to_int8_tensor.py
metascroy: I'll add just in case for future-proofing.

jerryzh168: also `apply_int8_act_asym_per_token_quant` should be `apply_int8_act_asym_per_token_quant_dequant` I think

test/quantization/quantize_/workflows/intx/test_intx_unpacked_to_int8_tensor.py
jerryzh168: is this still used in ET? ideally we can remove them I think, not sure what is the pytorch version compatibility guarantee in ET

torchao/quantization/quantize_/workflows/intx/intx_unpacked_to_int8_tensor.py
```diff
@@ -221,12 +263,7 @@ def _(func, types, args, kwargs):
 
     # Apply dynamic activation quant
     if weight_tensor.apply_int8_act_asym_per_token_quant:
-        input_tensor = IntxUnpackedToInt8Tensor.from_hp(
-            hp_tensor=input_tensor,
-            block_size=_get_per_token_block_size(input_tensor),
-            target_dtype=torch.int8,
-            mapping_type=MappingType.ASYMMETRIC,
-        ).dequantize()
+        input_tensor = _fake_apply_int8_act_asym_per_token_quant(input_tensor)
```
jerryzh168: why is this changed to a function call btw?

metascroy: choose_qparams returns a rank 0 tensor instead of a rank 1 when there is only 1 number, and this fails the assert in the constructor of IntxUnpackedToInt8Tensor (scale.shape == n_blocks).

torchao/quantization/quantize_/workflows/intx/intx_unpacked_to_int8_tensor.py
```diff
@@ -32,6 +32,14 @@
 _FLOAT_TYPES: List[torch.dtype] = [torch.float16, torch.bfloat16, torch.float32]
 
 
+class ActivationQuantization(enum.Enum):
```
jerryzh168: you can use `ActivationQuantization(str, Enum)` to avoid checking for str btw

example https://github.com/pytorch/ao/blob/b34c10379a3ae1dfa2e7c89bcf8cc02906b9d381/torchao/quantization/quantize_/common/kernel_preference.py#L12-L14

torchao/quantization/quantize_/workflows/intx/intx_unpacked_to_int8_tensor.py
jerryzh168: why do you need this assert? if you pass around block_size to quantize/dequantize op, this will be handled correctly

torchao/quantization/quantize_/workflows/intx/intx_opaque_tensor.py
```diff
@@ -15,6 +15,7 @@
 from torchao.quantization.quant_primitives import _DTYPE_TO_BIT_WIDTH
 from torchao.quantization.quantize_.workflows.intx.intx_unpacked_to_int8_tensor import (
     IntxUnpackedToInt8Tensor,
+    ActivationQuantization,
```
jerryzh168: naming is a bit too general I feel, how about:

```
IntxUnpackedToInt8ActivationQuantization
```

we can move this to `intx` folder and rename to `IntxActivationQuantization` if this enum can be reused by other intx tensors

test/quantization/quantize_/workflows/intx/test_intx_unpacked_to_int8_tensor.py
metascroy: It is still used.  We can try removing from ET and see if any tests break.

torchao/quantization/quantize_/workflows/intx/intx_unpacked_to_int8_tensor.py
metascroy: So if a user constructs the tensor using from_hp, it's not a concern.

But the user can also construct the tensor directly by supplying qdata, scale, zero_point and block_size separately, and in that case I think it makes sense to assert

PR #2944, opened 2025-09-05
[safetensor enablement] add fn to check if metadata is torchao

liangel-02:
**Summary**
adding util function to check if metadata is generated by our `flatten_tensor_state_dict` logic and accompanying tests. 

**Test**
```python test/prototype/safetensors/test_safetensors_support.py```

### CODE REVIEW

File: torchao/prototype/safetensors/safetensors_utils.py [hunk 1]
```diff
@@ -159,3 +161,20 @@ def object_from_dict(data: Dict[str, Any]):
         return cls(**processed_data)
     except Exception as e:
         raise ValueError(f"Failed to create instance of {cls.__name__}: {e}")
+
+
+def is_metadata_torchao(metadata: Dict[str, Any]):
+    if not metadata or "tensor_names" not in metadata:
+        return False
+
+    all_tensors = json.loads(metadata["tensor_names"])
```
jerryzh168: nit: `all_tensor_names`

test/prototype/safetensors/test_safetensors_support.py
```diff
@@ -41,6 +42,7 @@ def test_safetensors(self):
         model = torch.nn.Sequential(
             torch.nn.Linear(32, 256, dtype=torch.bfloat16, device="cuda")
         )
+        assert is_metadata_torchao(model.state_dict()) == False
```
jerryzh168: nit: `assert not`

jerryzh168: why do we pass state_dict to this function?

test/prototype/safetensors/test_safetensors_support.py
jerryzh168, line 1: can we have a separate test for the util, maybe manually construct 1-2 metadata_dict that doesn't follow the format, and one that follows the format to make sure the function is implemented correct

torchao/prototype/safetensors/safetensors_utils.py
```diff
@@ -159,3 +161,20 @@ def object_from_dict(data: Dict[str, Any]):
         return cls(**processed_data)
     except Exception as e:
         raise ValueError(f"Failed to create instance of {cls.__name__}: {e}")
+
+
+def is_metadata_torchao(metadata: Dict[str, Any]):
```
jerryzh168, line 172: maybe use `metadata_dict` to be precise, for both function name and arg name

also can you add a docstring for this function?

torchao/prototype/safetensors/safetensors_utils.py
jerryzh168, line 1: also for this file, I think we can define a `__all__` in the beginning of the file to include all public facing functions/objects, like: https://github.com/pytorch/ao/blob/e7b310b8441ad62ed28608b77685a29242f245f5/torchao/utils.py#L21

test/prototype/safetensors/test_safetensors_support.py
jerryzh168: probably should be in a separate test file as well I think, to align with where it is implemented

torchao/prototype/safetensors/safetensors_utils.py
```diff
@@ -159,3 +167,20 @@ def object_from_dict(data: Dict[str, Any]):
...
+def is_metadata_dict_torchao(metadata: Dict[str, Any]):
+    if not metadata or "tensor_names" not in metadata:
+        return False
+
+    all_tensor_names = json.loads(metadata["tensor_names"])
+
+    for tensor_name in all_tensor_names:
+        if not isinstance(metadata[tensor_name], str):
+            return False
+        tensor_dict = json.loads(metadata[tensor_name])
```
jerryzh168: this might fail right?

torchao/prototype/safetensors/safetensors_utils.py
```diff
@@ -159,3 +167,20 @@ def object_from_dict(data: Dict[str, Any]):
         return cls(**processed_data)
     except Exception as e:
         raise ValueError(f"Failed to create instance of {cls.__name__}: {e}")
+
+
+def is_metadata_dict_torchao(metadata: Dict[str, Any]):
+    if not metadata or "tensor_names" not in metadata:
+        return False
+
+    all_tensor_names = json.loads(metadata["tensor_names"])
```
jerryzh168: also this might fail?

test/prototype/safetensors/test_safetensors_utils.py
```diff
@@ -0,0 +1,43 @@
...
+from torchao.prototype.safetensors.safetensors_utils import is_metadata_dict_torchao
+from torchao.quantization.granularity import PerRow
+from torchao.quantization.quant_api import Float8DynamicActivationFloat8WeightConfig
+
+
+class TestSafeTensorsUtils(TestCase):
+    @parametrize(
+        "metadata_dict",
+        [
+            {"format": "pt"},
```
jerryzh168: can you include all False cases?

```
if not metadata or "tensor_names" not in metadata:
if not isinstance(metadata[tensor_name], str):
if tensor_type not in ALLOWED_TENSORS:

and the two json.loads that might fail
```

torchao/prototype/safetensors/safetensors_utils.py
```diff
@@ -159,3 +167,20 @@ def object_from_dict(data: Dict[str, Any]):
...
+
+    for tensor_name in all_tensor_names:
+        if not isinstance(metadata[tensor_name], str):
+            return False
+        tensor_dict = json.loads(metadata[tensor_name])
+        tensor_type = tensor_dict.get("_type")
```
jerryzh168, line 192: do we need to have a default value for this get?

torchao/prototype/safetensors/safetensors_utils.py
```diff
@@ -159,3 +167,20 @@ def object_from_dict(data: Dict[str, Any]):
...
+
+
+def is_metadata_dict_torchao(metadata: Dict[str, Any]):
+    if not metadata or "tensor_names" not in metadata:
+        return False
+
+    all_tensor_names = json.loads(metadata["tensor_names"])
+
+    for tensor_name in all_tensor_names:
+        if not isinstance(metadata[tensor_name], str):
```
jerryzh168: also there could be no `tensor_name` in metadata

torchao/prototype/safetensors/safetensors_utils.py
```diff
@@ -159,3 +167,20 @@ def object_from_dict(data: Dict[str, Any]):
         return cls(**processed_data)
     except Exception as e:
         raise ValueError(f"Failed to create instance of {cls.__name__}: {e}")
+
+
+def is_metadata_dict_torchao(metadata: Dict[str, Any]):
```
jerryzh168: nit: metadata --> metada_dict

test/prototype/safetensors/test_safetensors_utils.py
```diff
@@ -0,0 +1,59 @@
...
+from torchao.quantization.granularity import PerRow
+from torchao.quantization.quant_api import Float8DynamicActivationFloat8WeightConfig
+
+
+@unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
+class TestSafeTensorsUtils(TestCase):
+    @parametrize(
+        "metadata_dict",
+        [
+            {},
```
jerryzh168: nit: please also add some comments for each of these to indicate what we are testing as well

torchao/prototype/safetensors/safetensors_utils.py
```diff
@@ -159,3 +167,28 @@ def object_from_dict(data: Dict[str, Any]):
...
     except Exception as e:
         raise ValueError(f"Failed to create instance of {cls.__name__}: {e}")
+
+
+def is_metadata_dict_torchao(metadata_dict: Dict[str, Any]):
+    if not metadata_dict or "tensor_names" not in metadata_dict:
+        return False
+    try:
+        all_tensor_names = json.loads(metadata_dict["tensor_names"])
+    except Exception:
```
jerryzh168: we should catch only the error from failing `json.loads` I think

jerryzh168: according to https://docs.python.org/3/library/json.html#json.load, there is 
```
[JSONDecodeError](https://docs.python.org/3/library/json.html#json.JSONDecodeError) – When the data being deserialized is not a valid JSON document.

[UnicodeDecodeError](https://docs.python.org/3/library/exceptions.html#UnicodeDecodeError) – When the data being deserialized does not contain UTF-8, UTF-16 or UTF-32 encoded data.
```

Not sure if there are other errors though

torchao/prototype/safetensors/safetensors_utils.py
```diff
@@ -159,3 +167,28 @@ def object_from_dict(data: Dict[str, Any]):
...
+        return False
+
+    for tensor_name in all_tensor_names:
```
jerryzh168, line 183: can `all_tensor_names` be empty? if it's empty is it going to be a valid torchao checkpoint?

torchao/prototype/safetensors/safetensors_utils.py
jerryzh168: oh looks like it doesn't error out: https://docs.python.org/3/library/stdtypes.html#dict.get, then I think adding a comment here to say that if "_type" doesn't exist, the following check is still valid is good enough

liangel-02: .get() will return None if key doesn't exist, i'll add a comment

torchao/prototype/safetensors/safetensors_utils.py
liangel-02: adding TypeError as well

test/prototype/safetensors/test_safetensors_utils.py
```diff
@@ -0,0 +1,67 @@
...
+@unittest.skipIf(not is_sm_at_least_89(), "Need sm89+")
+class TestSafeTensorsUtils(TestCase):
+    @parametrize(
+        "metadata_dict",
+        [
+            {},  # empty dict
+            {"format": "pt"},  # no tensor_names
+            {
+                "0.weight": {"_type": "Int4Tensor"},  # tensor data not str
+                "tensor_names": '["0.weight", "0.bias"]',
```
jerryzh168: `all_tensor_names` is empty or not a list is not tested?

test/prototype/safetensors/test_safetensors_utils.py
```diff
@@ -0,0 +1,67 @@
...
+            },
+            {
+                "tensor_names": '["0.weight", "0.bias"]',  # missing corresponding tensors
+            },
+            {
+                "0.weight": '{"_type": "Int4Tensor"}',  # unsupported tensor subclass
```
jerryzh168: this is not tested I think
```
not isinstance(
            metadata_dict[tensor_name], str
        ):
```

test/prototype/safetensors/test_safetensors_utils.py
```diff
@@ -0,0 +1,67 @@
...
+            },
+            {
+                "0.weight": "{}",  # missing _type key
+                "tensor_names": '["0.weight", "0.bias"]',
+            },
+            {"tensor_names": PerRow()},  # json.loads() fails for tensor_names
+            {
+                "0.weight": PerRow(),  # json.loads() fails for tensor
+                "tensor_names": '["0.weight", "0.bias"]',
+            },
```
jerryzh168: ideally these can be ordered according to code I think, so that we know where the error might occur as we execute `is_metadata_dict_torchao` and make sure we test things exhaustively

test/prototype/safetensors/test_safetensors_utils.py
jerryzh168: also for this test, might be better to have a Tensor class name that we are not likely to use, like `Int4Tensor_NOT_SUPPORTED`, so that we don't need to update this in the future

liangel-02:
```
{
    "0.weight": {"_type": "Int4Tensor"},  # tensor data not str
    "tensor_names": '["0.weight", "0.bias"]',
},
```

  i think this covers, will reorder for clarity

test/prototype/safetensors/test_safetensors_utils.py
```diff
@@ -0,0 +1,73 @@
...
+            },  # json.loads() fails for metadata_dict["tensor_names"]
+            {"tensor_names": []},  # not tensor_names
+            {"tensor_names": "Int4Tensor_NOT_SUPPORTED"},  # tensor_names not a list
+            {
+                "tensor_names": '["0.weight", "0.bias"]',  # tensor_name not in metadata_dict
+            },
+            {
+                "0.weight": {
+                    "_type": "Int4Tensor_NOT_SUPPORTED"
+                },  # tensor data not str
```
jerryzh168: nit: might be better to just use a number here to make it less confusing

test/prototype/safetensors/test_safetensors_utils.py
```diff
@@ -0,0 +1,73 @@
...
+    @parametrize(
+        "metadata_dict",
+        [
+            {},  # not metadata_dict
+            {"format": "pt"},  # "tensor_names" not in metadata_dict
+            {
+                "tensor_names": PerRow()
+            },  # json.loads() fails for metadata_dict["tensor_names"]
+            {"tensor_names": []},  # not tensor_names
+            {"tensor_names": "Int4Tensor_NOT_SUPPORTED"},  # tensor_names not a list
```
jerryzh168: nit: same here, might be better to just use a number instead of `"Int4Tensor_NOT_SUPPORTED"` to make it more obvious

test/prototype/safetensors/test_safetensors_utils.py
```diff
@@ -0,0 +1,73 @@
...
+                "tensor_names": '["0.weight", "0.bias"]',  # tensor_name not in metadata_dict
+            },
+            {
+                "0.weight": {
+                    "_type": "Int4Tensor_NOT_SUPPORTED"
+                },  # tensor data not str
+                "tensor_names": '["0.weight", "0.bias"]',
+            },
+            {
+                "0.weight": PerRow(),  # json.loads() fails for metadata_dict[tensor_name]
```
jerryzh168: does this one also fail the previous test? since PerRow is not a string?

test/prototype/safetensors/test_safetensors_utils.py
```diff
@@ -0,0 +1,71 @@
...
+            {"tensor_names": "0"},  # tensor_names not a list
+            {
+                "tensor_names": '["0.weight", "0.bias"]',  # tensor_name not in metadata_dict
+            },
+            {
+                "0.weight": {"_type": "0"},  # tensor data not str
+                "tensor_names": '["0.weight", "0.bias"]',
+            },
+            {
+                "0.weight": "PerRow()",  # json.loads() fails for metadata_dict[tensor_name]
```
jerryzh168: does this actually fail?

liangel-02: yes, i also double checked

torchao/prototype/safetensors/safetensors_support.py
```diff
@@ -15,19 +15,20 @@
 
 def unflatten_tensor_state_dict(
     tensors_data_dict: Dict[str, Any],
-    metadata_dict: Dict[str, Any],
+    metadata: Dict[str, Any],
```
jerryzh168, line 18: fine for now, but in the future it would be better to have separate PR for these renaming changes

output.png
jerryzh168, line 1: revert?

PR #2945, opened 2025-09-05
Update model script for INT8-INT4

metascroy:
Updates INT8-INT4 to use version 2 configs, and fixes issue with tied embeddings.


.github/scripts/torchao_model_releases/quantize_and_upload.py
```diff
@@ -520,7 +522,7 @@ def _untie_weights_and_save_locally(model_id):
 _mobile_inference_recipe = """
 # Running in a mobile app
 (TODO: pte file name generation)
```
jerryzh168: can remove this line if the naming is already corrected

PR #2947, opened 2025-09-05
First attempt at QuantOptimizer.torchao_convert

lisjin:


torchao/prototype/parq/optim/quantopt.py
```diff
@@ -106,32 +124,87 @@ def quantize_(
...
             return False
 
-        return _filter_fn
+        for param_set in self._param_sets():
+            yield partial(_filter_fn, param_set=param_set)
+
+    def torchao_convert(
+        self,
+        model: nn.Module,
+        packing_format: PackingFormat = PackingFormat.UNPACKED_TO_INT8,
```
metascroy: We probably don't need this as an arg.  PackingFormat.UNPACKED_TO_INT8 is a canonical packing format, and we can convert to other ones after convert.

torchao/prototype/parq/quant/config_torchao.py
```diff
@@ -0,0 +1,141 @@
...
+        scale = scale.reshape(*n_blocks)
+        zero_point = zero_point.reshape(*n_blocks)
+
+        weight = IntxUnpackedToInt8Tensor(
+            qdata=qdata,
+            scale=scale,
+            zero_point=zero_point,
+            target_dtype=getattr(torch, f"int{config.b}"),
+            block_size=block_size,
+            dtype=weight.dtype,
```
lisjin, line 105: @metascroy Here's the new logic for converting the param quantized with StretchedUnifTorchaoQuantizer to IntxUnpackedToInt8Tensor format that you suggested last week

torchao/prototype/parq/optim/quantopt.py
```diff
@@ -91,6 +93,23 @@ def __repr__(self) -> str:
...
+        for group in self.regularized_param_groups():
+            group["num_steps"] = value
+            return
+
+    @num_steps.deleter
+    def num_steps(self) -> None:
+        for group in self.regularized_param_groups():
+            group.pop("num_steps", None)
+            return
+
```
lisjin, line 117: @Ninja91 FYI this is the upstream change for D81526700

Ninja91: Noice!

torchao/prototype/parq/quant/config_torchao.py
```diff
@@ -0,0 +1,143 @@
...
+    StretchedUnifTorchaoQuantizer,
+)
+
+
+@dataclass
+class StretchedIntxWeightOnlyConfig(IntxWeightOnlyConfig):
+    b: Optional[int] = None
+    quant_min: Optional[int] = None
+    quant_max: Optional[int] = None
+    activation_quantization: Optional[str] = "int8_asym_per_token"
```
metascroy, line 53: Why does a weight-only config have activation quantization?

torchao/prototype/parq/quant/config_torchao.py
```diff
@@ -0,0 +1,143 @@
...
+)
+from .uniform_torchao import (
+    _BIT_WIDTH_TO_DTYPE,
+    Int4UnifTorchaoQuantizer,
+    StretchedUnifTorchaoQuantizer,
+)
+
+
+@dataclass
+class StretchedIntxWeightOnlyConfig(IntxWeightOnlyConfig):
```
metascroy: Does StretchedIntxWeightOnlyConfig need to subclass IntxWeightOnlyConfig?

Can it be its own config?  It can still use the IntxUnpackedToInt8Tensor.

torchao/prototype/parq/quant/config_torchao.py
```diff
@@ -0,0 +1,143 @@
...
+        config = Int4WeightOnlyConfig(group_size=block_size, **kwargs)
+    elif isinstance(quantizer, StretchedUnifTorchaoQuantizer):
+        config = StretchedIntxWeightOnlyConfig(
+            b=b,
+            quant_min=quantizer.quant_min,
+            quant_max=quantizer.quant_max,
+            granularity=granularity,
+            version=version,
+        )
+    elif is_embed:
```
metascroy: Maybe this is more accurately "is_weight_only" than "is_embed"?

metascroy:
Thanks!  Overall looks great!  Left some comments.

torchao/prototype/parq/quant/config_torchao.py
lisjin: Good catch, I changed the config name to include dynamic activations. Passing activation_quantization=None will give you weight-only quantization. Let me know if this is confusing and we should have a separate StretchedIntxWeightOnlyConfig instead.

torchao/prototype/parq/quant/config_torchao.py
```diff
@@ -0,0 +1,200 @@
...
+    return config
+
+
+def is_hf_model(model: nn.Module) -> bool:
+    return TRANSFORMERS_AVAIL and isinstance(
+        getattr(model, "config", None), PretrainedConfig
+    )
+
+
+def save_hf_quantization_config(
```
andrewor14: right now this sounds like saving it to disk, maybe a better name is `attach_hf_quantization_config`?

torchao/prototype/parq/quant/config_torchao.py
```diff
@@ -0,0 +1,200 @@
...
+    model: nn.Module,
+    filter_fns: list[Callable[nn.Module, bool]],
+    configs: list[AOBaseConfig],
+) -> None:
+    """Save torchao quantization config to Hugging Face model."""
```
andrewor14: can you elaborate in the docstring what `filter_fns` does? Seems like you attach the config only if it returns true

torchao/prototype/parq/quant/config_torchao.py
```diff
@@ -0,0 +1,200 @@
...
+            weight_granularity=granularity,
+            weight_mapping_type=quantizer.mapping_type,
+            act_mapping_type=MappingType.ASYMMETRIC,
+            intx_packing_format=IntxPackingFormat.UNPACKED_TO_INT8,
+            version=version,
+        )
+    return config
+
+
+def is_hf_model(model: nn.Module) -> bool:
```
andrewor14: are these meant to be called by the user? If not can you add an underscore before it?

torchao/prototype/parq/optim/quantopt.py
```diff
@@ -106,32 +126,72 @@ def quantize_(
...
             for p in module.parameters(recurse=False):
                 if p.data_ptr() in param_set:
                     return True
             return False
 
-        return _filter_fn
+        for param_set in self._param_sets():
+            yield partial(_filter_fn, param_set=param_set)
+
+    def torchao_convert(self, model: nn.Module) -> None:
```
andrewor14: can you add some docs on when this is supposed to be called? Maybe add an example to the README?

torchao/prototype/parq/optim/quantopt.py
```diff
@@ -106,32 +126,72 @@ def quantize_(
...
+                quantizer,
+                weight_only,
+                device,
+                group["quant_bits"],
+                group.get("quant_block_size"),
+            )
+
+            # for HF model, delay quantize_ to after save_hf_quantization_config
+            # since it modifies p.data_ptr() and breaks filter_fn
+            if not save_hf_config:
```
metascroy: Nit: why not always delay?  After the loop you can have:
```
if save_hf_config:
    save_hf_quantization_config(model, filter_fns, configs)

for config, filter_fn in zip(configs, filter_fns):
    quantize_(model, config, filter_fn=filter_fn)
```

I agree with @andrewor14 that attach_hf_quantization_config is a better name

test/prototype/test_parq.py
```diff
@@ -115,22 +126,32 @@ def compare_parq_convert(
...
         if not _is_linear(module):
             continue
 
         p_orig = getattr(orig_model, n).weight  # PARQ weight
-        p = module.weight.dequantize()  # PARQ weight after quantize_
         p_ref = getattr(m_ref, n).weight.dequantize()  # native quantize_
-
         torch.testing.assert_close(p_orig, p_ref, atol=0, rtol=0)
+
+        p = module.weight.dequantize()  # PARQ weight after quantize_
```
metascroy, line 148: Can we also test that the forward function model(*example_inputs) gives expected result?

torchao/prototype/parq/quant/config_torchao.py
metascroy: I do think it is a bit confusing because it is either weight-only or dynamic based on activation_quantization.

Either use two configs, or call it "StretchedIntxWeightConfig" (so no reference to dynamic quant or weight-only in name)

torchao/prototype/parq/quant/config_torchao.py
```diff
@@ -0,0 +1,200 @@
...
+except ImportError:
+    TRANSFORMERS_AVAIL = False
+
+
+@dataclass
+class Int8DynamicActivationStretchedIntxWeightConfig(AOBaseConfig):
+    granularity: Granularity = PerAxis(0)
+    scale_dtype: Optional[torch.dtype] = None
+    layout: Layout = QDQLayout()
+    version: int = 1
```
metascroy: Layouts are being deprecated, so maybe make version = 2 the default?

metascroy:
Overall, I think it looks great!  Approving PR, but address the comments above before landing.

metascroy (approved):

lisjin:
Thanks for the comments @andrewor14 @metascroy :) I think I've addressed them all now, but let me know if I missed anything!

torchao/prototype/parq/quant/config_torchao.py
```diff
@@ -0,0 +1,211 @@
...
+            f"axis must be 0 with PerAxis, but got {granularity.axis}"
+        )
+        group_size = weight.shape[-1]
+    else:
+        raise ValueError(f"granularity must be PerGroup or PerAxis, got {granularity}")
+
+    block_size = (1, group_size)
+    target_dtype = torch.int8
+    q_args = (weight, mapping_type, block_size, target_dtype, config.b)
+    if config.version == 2:
```
jerryzh168, line 80: for this it's probably fine to break BC soon since it's a prototype feature?

lisjin: Got it, thanks. I kept the version=1 convention for this config to preserve the old functionality (StretchedAffineQuantizedTensor). The new version=2 will convert tensors to IntxUnpackedToInt8Tensor, which is consistent with other config classes like IntxWeightOnlyConfig

@metascroy Should we remove StretchedAffineQuantizedTensor entirely?

torchao/prototype/parq/optim/quantopt.py
lisjin: Good call, I've updated the prepare and convert examples in torchao/prototype/parq/README.md

torchao/prototype/parq/quant/config_torchao.py
metascroy: I have no objections to removing it.  I assumed you wanted it for Int4CPULayout, but if you don't need that, I think it's fine to remove.

On ExecuTorch side, IntxUnpackedToInt8 tensor will be sufficient

lisjin: I think you're right about CPU support. I'll leave it for now but consider removing it in the future

jerryzh168: Int4CPULayout is also moved to v2 already: https://github.com/pytorch/ao/pull/2845 you can set `int4_packing_format` to `plain_int32` to get it

PR #2950, opened 2025-09-06
Skip expanding scales for rowwise fp8 quantize

andrewor14:
**Summary:** https://github.com/pytorch/ao/pull/2253 added a step in `quantize_affine_float8` to expand the scales for blockwise quantization. The purpose of this step is to make the scales always broadcastable with the input tensor. However, this is unnecessary for rowwise quantization, which already has broadcastable shapes, e.g.

```
scale = [32, 1]
input = [32, 16]
```

Today, we will `repeat_interleave` the above scales to pad the scale tensor until it reaches `[32, 16]`, which adds non-trivial memory and latency overhead. This commit adds a fast path to skip this expanding step if we detect rowwise quantization.

**Test Plan:**
```
python test/quantization/test_quant_primitives.py -k test_maybe_expand_scale_to_tensor_shape
```

Also compared fine-tuning Qwen3-1.7B with fp8-fp8 QAT using batch size 32 on a single H100 GPU:
- Before: 25.34 GB peak memory, 3047.25 tok/s
- After: 22.53 GB peak memory, 3358.49 tok/s
- This PR uses 11.1% less memory and is 10.2% faster

### DIFF
--- a/test/quantization/test_quant_primitives.py
+++ b/test/quantization/test_quant_primitives.py
@@ -16,6 +16,7 @@ from torchao.quantization.quant_primitives import (
     _choose_qparams_affine_tinygemm,
     _fake_quantize_affine,
     _fake_quantize_affine_cachemask,
+    _maybe_expand_scale_to_tensor_shape,
     choose_qparams_affine,
     dequantize_affine,
     quantize_affine,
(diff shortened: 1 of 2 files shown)


test/quantization/test_quant_primitives.py
```diff
@@ -771,6 +772,20 @@ def test_fake_quantize_affine_cachemask(self):
         torch.testing.assert_close(dequantized, fake_quantized)
         torch.testing.assert_close(expected_mask, mask)
 
+    def test_maybe_expand_scale_to_tensor_shape(self):
+        # rowwise quantization: if all dimensions match except for the last one,
```
drisspg, line 776: we don't need the it to be only the last one right, its basically if it is expandable?

torchao/quantization/quant_primitives.py
```diff
@@ -2242,6 +2243,10 @@ def _expand_scale_to_tensor_shape(
         # Scalar scale - can broadcast naturally
         return scale
 
+    # For rowwise quantization, just return the scale as is
+    if scale.shape[:-1] == target_shape[:-1] and scale.shape[-1] == 1:
```
drisspg: you could probably do something fun, like 
```py
def is_trivial_expandable(scale, target_shape):
    return all(a == b or a == 1 for a, b in zip(scale.shape, target_shape))
```

PR #2951, opened 2025-09-06
Experimental folder deprecation part 2/x

metascroy:
In a previous PR, we migrated all C++ CPU kernel code out of experimental into csrc.

This PR starts to migrate the python code:

* Github workflow torchao_experimental_test.yml is renamed to regression_test_aarch64.yml 
* Tests in torchao/experimental/tests are migrated to test/quantization.  Some broken tests were deleted.
* torchao.experimental.op_lib (meta kernel registration) is migrated to torchao/csrc_meta_ops.py

### DIFF
--- a/torchao/__init__.py
+++ b/torchao/__init__.py
@@ -53,7 +53,6 @@ if not skip_loading_so_files:
 
         # The following registers meta kernels for some CPU kernels
         from torchao.csrc_meta_ops import *  # noqa: F403
-        # from torchao. import *  # noqa: F403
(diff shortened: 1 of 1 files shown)


kimishpatel (approved):
I presume all the deleted files code has already been moved?

metascroy:
> I presume all the deleted files code has already been moved?

Yes, except where the test is broken and no longer makes sense (e.g., there is a load ops test that no longer makes sense for how the ops are loaded now).

PR #2953, opened 2025-09-07
Switch IntxWeightOnlyConfig/Int8DynamicActivationIntxWeightConfig to v2

metascroy:
This PR switches IntxWeightOnlyConfig/Int8DynamicActivationIntxWeightConfig to v2.

### DIFF
--- a/.github/scripts/torchao_model_releases/quantize_and_upload.py
+++ b/.github/scripts/torchao_model_releases/quantize_and_upload.py
@@ -230,12 +230,10 @@ from torchao.quantization.granularity import PerGroup, PerAxis
 embedding_config = IntxWeightOnlyConfig(
     weight_dtype=torch.int8,
     granularity=PerAxis(0),
-    version=2,
 )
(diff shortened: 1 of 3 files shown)


torchao/experimental/tests/test_int8_dynamic_activation_intx_weight_config_v1.py
```diff
@@ -473,6 +471,7 @@ def test_identical_to_Int8DynamicActivationInt4WeightConfig(
                 weight_mapping_type=mapping_type,
                 weight_scale_dtype=None,
                 act_mapping_type=act_mapping_type,
+                version=1,
```
jerryzh168, line 474: might be better to split these changes to a separate PR I think, like: https://github.com/pytorch/ao/pull/2958

metascroy: I'll split this PR into 3:

1. Migrate existing v1 test to use version=1
2. Replace compute target by packing format
3. Switch to v2 by default

PR #2956, opened 2025-09-08
[mxfp8 moe training] per group scale conversion to blocked format with groups along K dim (for 2d2d grouped gemm)

danielvegamyhre:
## Summary
- We have a triton kernel that does per group mxfp8 scale conversion to blocked format, for 2d-3d grouped gemms, where the groups are along the `total_M` dimension in `(total_M, K) @ (E, K, N)`
- We now need triton kernels that does per group conversion for 2d-2d grouped gemms, where the groups are along the scaled dim: `(M, total_K) @ (total_K, N)`. 

#### LHS operand for 2d-3d MXFP8 grouped gemm
This is the existing kernel, the layout is much simpler for the 2d-3d case where the groups are along M:

Screenshot 2025-09-09 at 9 11 18 AM

#### LHS operand for 2d-2d MXFP8 grouped gemm
When groups are along the scaled dim being contracted, the memory layout is more complicated, as we have to represent separate standalone "row of blocks major" layouts in subtensors that are part of a larger parent tensor. 

Screenshot 2025-09-09 at 8 45 48 AM

## Other
- I also removed the `Mg` param from the `torch_to_blocked_per_group_2d` function (used for 2d-3d grouped gemms), since it is not used.

## Test plan
- `pytest test/prototype/moe_training/test_kernels.py -k test_mxfp8_per_group_blocked_scales_2d2d -s `

### DIFF
--- a/test/prototype/moe_training/test_kernels.py
+++ b/test/prototype/moe_training/test_kernels.py
@@ -22,13 +22,13 @@ from torchao.prototype.moe_training.kernels.jagged_float8_scales import (
     triton_fp8_per_group_rowwise_scales,
 )
 from torchao.prototype.moe_training.kernels.mxfp8_blocked_scales import (
-    compute_per_group_blocked_scale_offsets,
-    compute_per_group_blocked_scale_offsets_2d2d_lhs,
-    torch_to_blocked_per_group_2d,
-    torch_to_blocked_per_group_2d2d_lhs,
+    compute_blocked_scale_offsets_for_K_groups,
+    compute_blocked_scale_offsets_for_M_groups,
+    torch_to_blocked_2d_K_groups,
+    torch_to_blocked_2d_M_groups,
     torch_to_blocked_per_group_3d,
-    triton_mx_block_rearrange_per_group_2d,
-    triton_mx_block_rearrange_per_group_2d2d_lhs,
+    triton_mx_block_rearrange_2d_K_groups,
+    triton_mx_block_rearrange_2d_M_groups,
     triton_mx_block_rearrange_per_group_3d,
 )
 from torchao.prototype.moe_training.utils import (
@@ -229,15 +229,15 @@ def test_mxfp8_per_group_blocked_scales_2d(
     )
 
     # torch reference
-    ref_out_scales, _ = torch_to_blocked_per_group_2d(
+    ref_out_scales, _ = torch_to_blocked_2d_M_groups(
         e8m0_scales, input_group_offsets, k, block_size=block_size
     )
 
     # triton kernel
-    _, output_group_offsets = compute_per_group_blocked_scale_offsets(
+    _, output_group_offsets = compute_blocked_scale_offsets_for_M_groups(
         input_group_offsets
     )
-    triton_out_scales = triton_mx_block_rearrange_per_group_2d(
+    triton_out_scales = triton_mx_block_rearrange_2d_M_groups(
         e8m0_scales,
         input_group_offsets,
         output_group_offsets,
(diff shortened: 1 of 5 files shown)


test/prototype/moe_training/test_kernels.py
```diff
@@ -266,3 +269,49 @@ def test_mxfp8_per_group_blocked_scales_3d(
...
+    ref_out_scales, ref_start_cols_after_padding = torch_to_blocked_per_group_2d2d_lhs(
+        e8m0_scales,
+        input_group_offsets,
+    )
+
+    # triton kernel
+    _, output_group_offsets = compute_per_group_blocked_scale_offsets_2d2d_lhs(
+        input_group_offsets
+    )
+    assert torch.allclose(output_group_offsets, ref_start_cols_after_padding), (
```
drisspg: torch.equal?

torchao/prototype/moe_training/kernels/mxfp8_blocked_scales.py
```diff
@@ -58,6 +60,57 @@ def torch_to_blocked_per_group_2d(
     return blocked_scales, start_row_after_padding
 
 
+def torch_to_blocked_per_group_2d2d_lhs(
```
drisspg: nit: maybe soemthig like 2d_kmajor since this is only operates on 1 tensor

torchao/prototype/moe_training/kernels/mxfp8_blocked_scales.py
```diff
@@ -357,3 +441,152 @@ def triton_scale_swizzle_per_group_3d(
...
+    BLOCK_ROWS: tl.constexpr,
+    BLOCK_COLS: tl.constexpr,
+    DEBUG: tl.constexpr = False,
+):
+    group_pid = tl.program_id(0)
+    block_row_pid = tl.program_id(1)
+
+    # Input scales row range for this group
+    input_group_start_col = tl.load(
+        orig_offsets + group_pid - 1, mask=group_pid > 0, other=0
```
drisspg, line 533: nit, we dont actually need a mask here right?

torchao/prototype/moe_training/kernels/mxfp8_blocked_scales.py
```diff
@@ -357,3 +441,152 @@ def triton_scale_swizzle_per_group_3d(
...
+    )
+    input_group_end_col = tl.load(
+        orig_offsets + group_pid, mask=group_pid < num_groups, other=0
```
drisspg: nit we dont need a mask for this load right?

torchao/prototype/moe_training/kernels/mxfp8_blocked_scales.py
```diff
@@ -357,3 +441,152 @@ def triton_scale_swizzle_per_group_3d(
...
+    )
+    # Output scales start row we will begin writing to
+    output_group_start_col = tl.load(
```
drisspg: ditto

torchao/prototype/moe_training/kernels/mxfp8_blocked_scales.py
```diff
@@ -357,3 +441,152 @@ def triton_scale_swizzle_per_group_3d(
...
+        output_scales_group_offsets + group_pid, mask=group_pid < num_groups, other=0
+    )
+
+    # Calculate destination indices for each row and col in block swizzled layout.
```
drisspg: can you make this a helper jit function and reuse elsewwhere, its fine if you stack that commit on this

torchao/prototype/moe_training/kernels/mxfp8_blocked_scales.py
danielvegamyhre: Hmm for 2d-3d and 2d-2d, the 2d tensors are all K major aren't they? I was actually thinking we should just remove the `_lhs` suffix from the name, since I think the kernel will actually work for both LHS and RHS operands in the 2d2d grouped gemm, given shapes `(M, total_K)` and `(N, total_K)`

How about: `torch_to_blocked_per_group_for_2d2d` (added "for")

danielvegamyhre: Renamed this func and others for clarity that the difference is about grouping along M dim vs K dim

torchao/prototype/moe_training/kernels/mxfp8_blocked_scales.py
danielvegamyhre: Updated

torchao/prototype/moe_training/kernels/mxfp8_blocked_scales.py
danielvegamyhre: good idea, done

danielvegamyhre:
@drisspg i finished addressing your comments, ready for another look

PR #2957, opened 2025-09-08
Fix ROCM QAT test failure

andrewor14:
https://github.com/pytorch/ao/actions/runs/17551856077/job/49846025788
```
AssertionError: Object comparison failed: torch.float8_e4m3fnuz != torch.float8_e4m3fn
```


jerryzh168:
@vkuzo I remember you mentioned that we don't want to use `from torchao.float8.config import e4m3_dtype`, is the alternative just to set these dtype manually at callsite base on whether rocm or not rocm?

PR #2959, opened 2025-09-08
IntxWeightOnlyConfig/Int8DynamicIntxWeightConfig v2 migration: use version=1 in tests

metascroy:
Use version=1 in old tests

### DIFF
--- a/torchao/quantization/quantize_/workflows/intx/intx_opaque_tensor.py
+++ b/torchao/quantization/quantize_/workflows/intx/intx_opaque_tensor.py
@@ -29,6 +29,16 @@ __all__ = [
 aten = torch.ops.aten
 
 
+def is_kernel_library_loaded():
+    loaded = False
+    try:
+        _check_torchao_ops_loaded()
+        loaded = True
+    except Exception:
+        pass
+    return loaded
+
+
 class ComputeTarget(enum.Enum):
     """
     This packs the tensor for PyTorch CPU kernels in ATen.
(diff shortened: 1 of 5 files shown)


torchao/quantization/quantize_/workflows/intx/intx_opaque_tensor.py
```diff
@@ -29,6 +29,16 @@
 aten = torch.ops.aten
 
 
+def is_kernel_library_loaded():
```
jerryzh168: why is this defined here? should this be in some util path?

jerryzh168: also is the naming `is_kernel_library_loaded` too broad?

metascroy: It's defined here because intx_opaque_tensor is the user-API for accessing these kernels.

On the name being too broad: I think when considered with its full include path (which include intx_opaque_tensor), it is not too broad.

jerryzh168: `intx_opaque_tensor` is an implementation detail I think, I don't think we want to import from this file, can you import from `quantize_.workflows` instead?

metascroy: Sure, but I don't expect people to ever import is_kernel_library_loaded.

It's only used for our CI tests to know when to skip.  I could also add an underscore?

jerryzh168: I see, yeah in that case adding an underscore would be better I think

metascroy: Added underscore

PR #2960, opened 2025-09-08
Remove compute target from intx_opaque_tensor

metascroy:
This PR removes compute target from intx_opaque_tensor

### DIFF
--- a/torchao/quantization/quantize_/workflows/intx/intx_opaque_tensor.py
+++ b/torchao/quantization/quantize_/workflows/intx/intx_opaque_tensor.py
@@ -6,7 +6,7 @@
 
 
 import logging
-from typing import Optional, Union
+from typing import Optional
 
 import torch
 
@@ -30,6 +30,7 @@ __all__ = [
 
 aten = torch.ops.aten
 
+
 def _is_kernel_library_loaded():
     loaded = False
     try:
@@ -38,6 +39,7 @@ def _is_kernel_library_loaded():
     except Exception:
         pass
     return loaded
+
 
 class IntxOpaqueTensor(TorchAOBaseTensor):
     """
@@ -116,9 +118,7 @@ class IntxOpaqueTensor(TorchAOBaseTensor):
         tensor: IntxUnpackedToInt8Tensor,
         *,
         bias: Optional[torch.Tensor] = None,
-        intx_packing_format: Union[
-            IntxPackingFormat, str
-        ] = IntxPackingFormat.OPAQUE_TORCHAO_AUTO,
+        intx_packing_format: IntxPackingFormat = IntxPackingFormat.OPAQUE_TORCHAO_AUTO,
     ):
         """
         Constructs a IntxOpaqueTensor from an IntxUnpackedToInt8Tensor.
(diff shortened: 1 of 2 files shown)


torchao/quantization/quantize_/workflows/intx/intx_opaque_tensor.py
```diff
@@ -71,7 +45,7 @@ class IntxOpaqueTensor(TorchAOBaseTensor):
         dtype: dtype for activations/outputs
         packed_weights_has_zeros: whether zeros are present in packed_weights
         packed_weights_has_bias: whether bias is present in packed_weights
-        compute_target: the compute target for the packed data.  Compute targets may pack the data differently.  See ComputeTarget enum for details.
+        packing_format: the packing format for the packed data.  See IntxPackingFormat enum for details.
```
jerryzh168: nit: link to `IntxPackingFormat`, example: https://github.com/pytorch/ao/blob/ac5ab7e630bc54c867775f1fe420a44df9e37409/torchao/quantization/qat/api.py#L73

torchao/quantization/quantize_/workflows/intx/intx_opaque_tensor.py
```diff
@@ -120,10 +94,10 @@ def __init__(
         self.block_size = block_size
         self.packed_weights_has_zeros = packed_weights_has_zeros
         self.packed_weights_has_bias = packed_weights_has_bias
-        self.compute_target = compute_target
+        self.packing_format = packing_format
```
jerryzh168: should we use `intx_packing_format` instead?

metascroy: I prefer packing_format, but if you have strong preference, I can change it.  It should be clear from the type.

jerryzh168: yeah I feel it will be clearer to not confuse packing_format from different dtypes

torchao/quantization/quantize_/workflows/intx/intx_opaque_tensor.py
```diff
@@ -347,4 +330,4 @@ def _(func, types, args, kwargs):
 
 IntxOpaqueTensor.__module__ = "torchao.quantization"
 
-torch.serialization.add_safe_globals([IntxOpaqueTensor, ComputeTarget])
+torch.serialization.add_safe_globals([IntxOpaqueTensor, IntxPackingFormat])
```
jerryzh168: feels like IntxPackingFormat should be added in the `intx_packing_format.py` instead

torchao/quantization/quantize_/workflows/intx/intx_opaque_tensor.py
```diff
@@ -134,15 +108,24 @@ def from_intx_unpacked_to_int8_tensor(
         tensor: IntxUnpackedToInt8Tensor,
         *,
         bias: Optional[torch.Tensor] = None,
-        compute_target: Union[ComputeTarget, str] = ComputeTarget.TORCHAO_AUTO,
+        intx_packing_format: Union[
```
jerryzh168: nit: this can just be `IntxPackingFormat` since it's already inherits from both str and Enum

Issue #2896, opened 2025-08-28
[CPU][FP8][Inductor] How to support fp8 quant for inductor on CPU


shiyang-weng:
What we want to do is to enable FP8 quantization in PyTorch. Similar to INT8 quantization, this requires inserting quantize and dequantize operations into the computational graph. In order to reuse pattern matching logic of int8, we need register FP8 quant and dequant.

To address this, we attempted to register quant in [#2379](https://github.com/pytorch/ao/pull/2379), but the PR was reverted in [#2672](https://github.com/pytorch/ao/pull/2672) because it caused performance regression on H100 GPUs.

It will take a lot of effort to find the root cause of GPU regression.
Maybe we can register quant specifically for CPU, but this requires defining and registering a separate function for CPU.
@jerryzh168 @vkuzo Do you have some suggestions about it?
cc @Xia-Weiwen 

I create following test to show the issue.
```python
import os

os.environ["OMP_NUM_THREADS"] = "1"
os.environ["TORCHINDUCTOR_FREEZING"] = "1"
os.environ["TORCH_COMPILE_DEBUG"] = "1"
os.environ["TORCHDYNAMO_PRINT_GUARD_FAILS"] = "1"

import torch
import torchao

dtype = torch.float
qtype = torch.float8_e4m3fn

def dequantize_per_tensor(
    tensor: torch.Tensor,
    scale: float,
    output_dtype: torch.dtype
) -> torch.Tensor:
    res = torchao.quantization.quant_primitives._dequantize_affine_float8(
        tensor=tensor,
        scale=torch.tensor([scale]),
        output_dtype=torch.float
    )
    return res

def quantize_per_tensor(
    tensor: torch.Tensor,
    scale: float,
) -> torch.Tensor:
    return torchao.quantization.quant_primitives._quantize_affine_float8(
        tensor=tensor,
        scale=torch.tensor([scale]),
        float8_dtype=torch.float8_e4m3fn,
    )


class FP8QDQLinear(torch.nn.Module):
    def __init__(self, in_features, out_features):
        super().__init__()
        self.weight = torch.randn((out_features, in_features),).to(qtype)
        self.weight_scale = 1.0
        self.scale = 1.0
        self.bias = None

    def forward(self, input):
        weight = dequantize_per_tensor(
            self.weight.data,
            self.weight_scale,
            dtype,
        )
        q_input = quantize_per_tensor(
            input,
            self.scale,
        )

        dq_input = dequantize_per_tensor(
[...]

PR #2961, opened 2025-09-09
[Float8] add non-decomposed version of quantize/dequantize ops for fp8

shiyang-weng:
Fix https://github.com/pytorch/ao/issues/2896

What we want to do is to enable FP8 quantization in PyTorch. Similar to INT8 quantization, this requires inserting quantize and dequantize operations into the computational graph. In order to reuse pattern matching logic of int8, we need register FP8 quant and dequant.

To address this, we attempted to register quant in https://github.com/pytorch/ao/pull/2379, but the PR was reverted in https://github.com/pytorch/ao/pull/2672 because it caused performance regression on H100 GPUs. And there is no need to register q/dq on CUDA.

Based on the above reasons, I register quant specifically for CPU.


LevelDownRefine:
@vkuzo Could you help review this PR?

jerryzh168:
> @vkuzo Could you help review this PR?

is this urgent? Vasiliy is not available recently and will be back next week

LevelDownRefine:
> is this urgent? Vasiliy is not available recently and will be back next week

Thanks for letting me know. Not urgent. We can wait for him back next week

LevelDownRefine:
@vkuzo Could you help review this PR?

Xia-Weiwen:
Hi @vkuzo Thanks for your suggestion. May I know what kind of flag you were talking about? A global flag, or an argument passed to quantization APIs?

Xia-Weiwen:
> Hi @vkuzo Thanks for your suggestion. May I know what kind of flag you were talking about? A global flag, or an argument passed to quantization APIs?

Hi @vkuzo @jerryzh168 Could you share a little more about the design? Thanks.

jerryzh168:
I think the decision of decompose or not decompose should be static? if we want consistent behavior for the same op across cuda and cpu, it might be better to have separate ops I feel

Xia-Weiwen:
Hi @jerryzh168 Would you think it better to have a non-decomposed and a decomposed version of the op than a CPU and a CUDA version? We did a similar thing here: https://github.com/pytorch/pytorch/blob/df4ebddbe0fa2306fb8acd09b20265046d968c10/torch/ao/quantization/fx/_decomposed.py#L1206
also @vkuzo

jerryzh168:
yeah just a different op seems to be the only alternative here

LevelDownRefine:
Created quantize_affine_float8_non_decomposed and dequantize_affine_float8_non_decomposed separately for non-decomposed

jerryzh168:
LGTM, cc @vkuzo can you take a look again

PR #2969, opened 2025-09-09
forward fix shape guards handling for executorch

avikchaudhuri:
Summary:
D80713603 introduces a call module to a generated `_guards_fn` by export that ET pass infra cannot handle right now.

It is possible to omit this generation by using `ep.module(check_guards=False)` but there are too many places to update, so meanwhile we can explicitly remove the generated instruction as in here.

Differential Revision: D82030581

### DIFF
--- a/torchao/quantization/pt2e/lowering.py
+++ b/torchao/quantization/pt2e/lowering.py
@@ -57,7 +57,7 @@ def lower_pt2e_quantized_to_x86(
     lowered_model = (
         torch.export.export(model, example_inputs, strict=True)
         .run_decompositions(_post_autograd_decomp_table())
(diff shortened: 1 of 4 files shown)


Burton2000:
This seems to still be a problem now that executorch 1.0.0 has updated to use pytorch 2.9.0 and torchao 0.14.0

I tried this fix locally and seems to solve the issue, any reason it wasn't merged or is there a different solution?

azakhtyamov:
still a problem for torchao=0.14.1 and executorch=1.0.0

Burton2000:
@digantdesai if I were to raise this pull request again would it be possible to get merged?

PR #2974, opened 2025-09-10
Revert "[CPU] Support int8 scaled embedding bag"

metascroy:
Reverts pytorch/ao#2938

This is failing internal tests with compile error.  Please remove the unused variable qtype and reland, see https://github.com/pytorch/ao/pull/2972.

### DIFF
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -779,10 +779,19 @@ EMBEDINGBAG_TEST_PARAMS = list(
 )
 
 
-def _test_scaled_embedding_bag_cpu_helper(
-    multi_hot, batch_size, vector_size, index_type, qtype
-):
+@pytest.mark.skipif(
+    "CPU" not in torch._C._dispatch_dump("torchao::_scaled_embedding_bag"),
+    reason="cpp kernels not built",
+)
+@pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


jerryzh168 (approved):

Xia-Weiwen:
Hi @jerryzh168 @metascroy Sorry for the issue you met with the code. We are going to resubmit a PR. However, I have a few questions about this reversion:
- Why did the unused variable produce a compile time error instead of a warning? How did you compile it? Maybe we can use the same way to compile to find such unused variables in the future.
- Why not landing this PR https://github.com/pytorch/ao/pull/2972 instead of reverting the entire https://github.com/pytorch/ao/pull/2938?

Thanks.

jerryzh168:
for "Why not landing this PR https://github.com/pytorch/ao/pull/2972 instead of reverting the entire https://github.com/pytorch/ao/pull/2938?"

it's because many internal tests are failing so would be hard to verify if the fix is enough or not one by one

jerryzh168:
error is:
```
8: error: unused variable 'qtype' [-Werror,-Wunused-variable]
  177 |   auto qtype = qweight.scalar_type();
      |        ^~~~~
1 error generated.
```

Xia-Weiwen:
Got it. Thanks.

Issue #2932, opened 2025-09-03
importing torchao.prototype.mxfp8_cuda extension does not work from torchao wheels built by CI


vkuzo:
Repro:

1. install a recent torchao wheel using `pip`, 0.13.0 RC or 0.14.x nightly: `pip install torchao==0.13.0 --extra-index-url https://download.pytorch.org/whl/test/cu128`
2. try to import NVFP4: `from torchao.prototype.mx_formats import NVFP4InferenceConfig`

You'll see an error:

```bash
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/root/miniconda3/envs/py3.11/lib/python3.11/site-packages/torchao/prototype/mx_formats/__init__.py", line 8, in <module>
    from torchao.prototype.mx_formats.inference_workflow import (
  File "/root/miniconda3/envs/py3.11/lib/python3.11/site-packages/torchao/prototype/mx_formats/inference_workflow.py", line 20, in <module>
    from torchao.prototype.mx_formats.mx_tensor import MXTensor, QuantizeTensorToMXKwargs
  File "/root/miniconda3/envs/py3.11/lib/python3.11/site-packages/torchao/prototype/mx_formats/mx_tensor.py", line 47, in <module>
    from torchao.prototype.mx_formats.kernels import (
  File "/root/miniconda3/envs/py3.11/lib/python3.11/site-packages/torchao/prototype/mx_formats/kernels.py", line 1481, in <module>
    from torchao.prototype import mxfp8_cuda
ImportError: cannot import name 'mxfp8_cuda' from 'torchao.prototype' 
```

Root cause: the `mxfp8_cuda` extension contains our custom c++ kernels for casting to mxfp8 across dim1.  The extension is:
1. enabled to build if CUDA 12.8 is present (here: https://github.com/pytorch/ao/blob/85557135c93d3429320a4a360c0ee9cb49f84a00/setup.py#L641)
2. actually builds a kernel only for CUDA capability 10.0 (here: https://github.com/pytorch/ao/blob/85557135c93d3429320a4a360c0ee9cb49f84a00/torchao/csrc/cuda/mx_kernels/mxfp8_quantize.cuh#L28)

In today's CI jobs, we are building with CUDA 12.8 but we don't have NVIDIA B200s available, so we pass (1) and fail (2).

[...]

PR #2979, opened 2025-09-11
Add nvcc flags to explicitly build mxfp8 dim1 cast kernel for sm100a 

danielvegamyhre:
Fixes https://github.com/pytorch/ao/issues/2932

I think we need to explicitly add the nvcc flags to build for sm100a in the extension itself. Right now, we check if the `build_for_sm100a` flag is true, which is set to true for cuda 12.8+, but we don't actually modify the nvcc args passed in to build the extension.

Seems like building from source is accidentally working? Looking into this..

## Test plan
- CI Job building for CUDA 12.8 DOES have `building 'torchao.prototype.mxfp8_cuda' extension` logs but does NOT have this [warning](https://github.com/pytorch/ao/blob/186aeb01664687d14108ada420c475cc783e1643/torchao/csrc/cuda/mx_kernels/mxfp8_quantize.cuh#L30C6-L30C25) (`"MXFP8 quantization requires SM90+ (Hopper) or SM100+ (Blackwell) architecture. Kernel will be disabled for this architecture."`) in it: https://github.com/pytorch/ao/actions/runs/17631942858/job/50100852741 
    - (previously, without this fix, in the CI build logs indicating the extension was being built, but also see the warning that cuda arch not supported so kernel will not be built)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -677,7 +677,11 @@ def get_extensions():
                     ],
                     extra_compile_args={
                         "cxx": ["-std=c++17", "-O3"],
-                        "nvcc": nvcc_args + ["-gencode=arch=compute_100a,code=sm_100a"],
+                        "nvcc": nvcc_args
+                        + [
+                            "-gencode=arch=compute_100,code=sm_100",
+                            "-gencode=arch=compute_120,code=compute_120",
+                        ],
                     },
                 ),
             )


drisspg:
We have separate 100a modules if you look lower in the file

danielvegamyhre:
> We have separate 100a modules if you look lower in the file

Are you referring to this? https://github.com/pytorch/ao/blob/cc351513188d4936cf4d8cc305738d07e51b348a/setup.py#L706

setup.py
```diff
@@ -677,7 +677,7 @@ def get_extensions():
                     ],
                     extra_compile_args={
                         "cxx": ["-std=c++17", "-O3"],
-                        "nvcc": nvcc_args,
+                        "nvcc": nvcc_args + ["-gencode=arch=compute_100a,code=sm_100a"],
```
drisspg: update to 100, and add 120

danielvegamyhre:
Confirmed build is successful and `pytest test/prototype/mx_formats/test_mx_linear.py -k test_linear_eager_vs_hp` still passes

PR #2984, opened 2025-09-11
Updates LUT tensor and new convert API

metascroy:
For better integration with the parq optimizer, we add a convert method for LUT tensors.  We can now do:

With https://github.com/pytorch/ao/pull/2947, we can now do:
```
optimizer.torchao_convert(model)
```

With the changes in this PR, the converted model can then be optimized for CPU with:
```
from torchao.prototype.int8_lut_tensor import convert_model
convert_model(model)
```

convert_model converts IntxUnpackedToInt8Tensor's to optimized Int8LutTensor's for better performance on Arm CPUs.

### DIFF
--- a/test/prototype/test_int8_lut_tensor.py
+++ b/test/prototype/test_int8_lut_tensor.py
@@ -9,7 +9,6 @@ from copy import deepcopy
 import pytest
 import torch
 
-from torchao.prototype.conversion.api import _convert_to_optimized_model_for_aarch64
 from torchao.prototype.parq.quant import (
     StretchedIntxWeightConfig,
     StretchedUnifTorchaoQuantizer,
@@ -17,6 +16,7 @@ from torchao.prototype.parq.quant import (
 from torchao.prototype.quantization.int8_lut_tensor.int8_lut_tensor import (
     _is_kernel_library_loaded,
 )
+from torchao.prototype.tensor_conversion.api import _convert_model_for_aarch64
 from torchao.quantization import quantize_
 from torchao.quantization.granularity import PerAxis, PerGroup
 from torchao.quantization.utils import compute_error
@@ -75,7 +75,7 @@ def test_parq_conversion(dtype, granularity, bit_width, lead_dim):
 
     # Convert PARQ model to lowbit LUT model
     lut_model = deepcopy(parq_model)
-    _convert_to_optimized_model_for_aarch64(lut_model, tensor_type="int8_lut_tensor")
+    _convert_model_for_aarch64(lut_model, tensor_type="int8_lut_tensor")
 
     # Run both models and compare
     parq_out = parq_model(activations)
@@ -111,7 +111,7 @@ def test_export(dtype, granularity, bit_width, lead_dim):
     activations = parq_model.example_inputs(lead_dim=lead_dim)
     quantize_(parq_model, config)
 
-    _convert_to_optimized_model_for_aarch64(parq_model)
+    _convert_model_for_aarch64(parq_model)
 
     ep = torch.export.export(parq_model, (activations,))
 
--- a/torchao/prototype/tensor_conversion/api.py
+++ b/torchao/prototype/tensor_conversion/api.py
@@ -8,7 +8,7 @@ import torch
 import torch.nn as nn
 
 
-def _convert_module_to_int8_lut_tensor(module):
+def _convert_linear_weight_to_int8_lut_tensor(module):
     from torchao.prototype.quantization.int8_lut_tensor import Int8LutTensor
 
     assert isinstance(module, nn.Linear)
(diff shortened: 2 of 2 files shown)


torchao/prototype/quantization/int8_lut_tensor/api.py
```diff
@@ -0,0 +1,37 @@
...
+import torch
+import torch.nn as nn
+
+from torchao.prototype.quantization.int8_lut_tensor.int8_lut_tensor import (
+    Int8LutTensor,
+)
+from torchao.quantization.quantize_.workflows import IntxUnpackedToInt8Tensor
+
+
+def convert_model(model):
```
jerryzh168: is this the `_convert_to_optimized_tensor_based_on_hardware` function that we plan to use for vllm? can you move this to torchao/prototype/quantization/api.py?

torchao/prototype/quantization/int8_lut_tensor/api.py
```diff
@@ -0,0 +1,37 @@
...
+                    f"Skipping converting {name} to Int8LutTensor because its weight is not an IntxUnpackedToInt8Tensor"
+                )
+                continue
+            try:
+                new_weight = Int8LutTensor.from_intx_unpacked_to_int8_tensor(
+                    weight, bias=module.bias
+                )
+                module.weight = torch.nn.Parameter(new_weight, requires_grad=False)
+                module.bias = None
+            except Exception as e:
```
jerryzh168: is it possible to explicitly list all expected exception here?

torchao/prototype/quantization/int8_lut_tensor/api.py
metascroy: This is different, but I do plan to call this function as part of that API.

I can move this function to that location anyway, though.

Should I use torchao/prototype/quantization/api.py or would torchao/prototype/conversion/api.py be better?

jerryzh168: `torchao/prototype/conversion/api.py` sounds good I think, we can keep each feature in separate folders

torchao/prototype/conversion/api.py
```diff
@@ -0,0 +1,48 @@
...
+    assert isinstance(module, nn.Linear)
+    weight = module.weight
+    new_weight = Int8LutTensor.from_intx_unpacked_to_int8_tensor(
+        weight, bias=module.bias
+    )
+    module.weight = torch.nn.Parameter(new_weight, requires_grad=False)
+    module.bias = None
+
+
+def _convert_to_optimized_model_for_aarch64(
```
jerryzh168: so this is not supposed to be included in the main `_convert_to_optimized_model_based_on_hardware`, but called separately by user?

torchao/prototype/conversion/api.py
```diff
@@ -0,0 +1,48 @@
...
+    model,
+    *,
+    tensor_type="int8_lut_tensor",
```
jerryzh168: why do we have this arg? so we just use this arg to do the conversion instead of automatically decide what's best?

if so maybe should rename the function name to not include `optimized` and `aarch64`

torchao/prototype/conversion/api.py
metascroy: It can be called by _convert_to_optimized_model_based_on_hardware in one of the if/else blocks.

But ET might call it directly since ET doesn't care about the GPU optimizations.

torchao/prototype/conversion/api.py
metascroy: In a follow-up PR, I'm going to set this to "auto", which just picks the best.  But for unit testing, it's quite helpful to force a specific conversion.

torchao/prototype/conversion/__init__.py
jerryzh168, line 1: @metascroy also maybe `tensor_conversion` to be more specific?

torchao/prototype/conversion/api.py
jerryzh168: OK, might be better to remove optimized and aarch64 then since the target is user specified

torchao/prototype/conversion/api.py
```diff
@@ -0,0 +1,48 @@
...
+
+        weight = module.weight
+        if not isinstance(weight, IntxUnpackedToInt8Tensor):
+            print(
+                f"Skipping converting {name} to IntxOpaqueTensor because its weight is not an IntxUnpackedToInt8Tensor"
+            )
+            continue
+
+        if tensor_type == "int8_lut_tensor":
+            _convert_module_to_int8_lut_tensor(module)
```
jerryzh168: also `_convert_module_to_int8_lut_tensor` is a weird name, probably you meant `_convert_module_weight_to_int8_lut_tensor`, also since it's only linear, you can do `_convert_linear_weight_to_int8_lut_tensor`

torchao/prototype/conversion/api.py
metascroy: But all the supported targets are optimized aarch64 targets

I can remove optimized, though, if you want to reserve that for auto

jerryzh168: OK sounds good

metascroy:
Any more concerns here @jerryzh168?

PR #2990, opened 2025-09-12
[mxfp8 moe training] add compile support

danielvegamyhre:
Stacked PRs:
 * #3004
 * #3002
 * #2999
 * #2998
 * __->__#2990

--- --- ---

### [mxfp8 moe training] add compile support

- wrap triton kernels in custom ops
- fix d2h sync caused by doing `padded_rows = output_scale_group_offsets[-1]`. By looking at the trace I found this was doing a .item() under the hood and causing a d2h sync, so now instead I'm just using the upper bound of the padding needed via `padded_rows = rows + num_groups * 128` and avoiding the d2h sync.

### Test plan
- `pytest test/prototype/moe_training/test_training.py `

### Microbenchmarks with compile

```
A_shape        B_shape           recipe                  bf16_e2e_us    scaled_e2e_us  scaled_e2e_speedup      bf16_fwd_us    scaled_fwd_us  scaled_fwd_speedup
-------------  ----------------  --------------------  -------------  ---------------  --------------------  -------------  ---------------  --------------------
(16640, 5120)  (1, 8192, 5120)   MoEScalingType.MXFP8        4268.5           3402.75  1.254x                      1513.76          1675.81  0.903x
(16640, 5120)  (4, 8192, 5120)   MoEScalingType.MXFP8        3968.88          4282.53  0.927x                      1126.21          2222.37  0.507x
(16640, 5120)  (16, 8192, 5120)  MoEScalingType.MXFP8        4900.77          8091.55  0.606x                      1262.66          9047.7   0.14x
(16640, 5120)  (64, 8192, 5120)  MoEScalingType.MXFP8        8432.61         21453.3   0.393x                      1788.94         14476.4   0.124x
```

### Perf analysis for M=16640, G=4, N=8192, K=5120
- [bf16 trace](https://github.com/user-attachments/files/22308898/bf16_profile.json)
- [mxfp8 trace](https://github.com/user-attachments/files/22308899/scaled_profile.json)

Looking at the trace to see why it's slower, here are my initial takeways:

1. **Mxfp8 grouped GEMMs are 1.57x faster than bf16 grouped gemms, but should be 2.1x faster**: (mxpf8 avg = ~700us, bf16 avg = ~1.1ms). This contradicts the grouped GEMM microbenchmarking we did, which indicates for shape M=16640, G=4, K=5120, N=8192 we should be getting a ~2.1x speedup, but we only are getting a ~1.57x speedup. So we need to determine what is going on here. cc @cthi @slayton58 who may be able to help with this

2. **Blocked scale swizzling kernels look okay**: 2 of the 3 of the handwritten triton kernels for converting scales to blocked swizzled format are very fast. One may need more optimization (TBD):
    - `triton_scale_swizzle_M_groups` = 18us avg
    - `triton_scale_swizzle_K_groups` = 14us avg
    - `triton_scale_swizzle_per_group_3d` = ~250us avg (not surprising it's longer since 3d tensor is much more data than 2d activations, will get some mem bw benchmarks on these kernels though)

3. **Mxfp8 dim1 cast CUDA kernel looks good**: ~100us average, used on 2d RHS operands. This kernel achieves ~5300gbps mem bw, around 66% of the peak 8TB/s bandwidth, so it can potentially still be improved.

5. **Torch inductor codegen kernels are pretty slow** (worst offender is ~1.2ms, longer than the grouped GEMM itself). This is probably largely due to [stray *.contiguous() call I was forced to do by to_mx API limitations](https://github.com/pytorch/ao/blob/cc65dc5d4908902a0de0ecc86cba733e95ddce2c/torchao/prototype/moe_training/scaled_grouped_mm.py#L363) but I'm guessing it's also that inductor codegen is slow, as it has been historically for various cases in fp8 rowwise, fp8 blockwise, and mxfp8. So we should try to get the [mxfp8 dim1 cuda kernel](https://github.com/pytorch/ao/blob/85557135c93d3429320a4a360c0ee9cb49f84a00/torchao/csrc/cuda/mx_kernels/mxfp8_quantize.cuh) working for 3d tensors so we can get the double win of avoiding the .contiguous() calls and using the faster kernel / achieves higher mem bw utilization than torch.compile / triton. I can do this next, or perhaps @slayton58 may be interested in this. (Perhaps we can just reshape 3d->2d, use the existing kernel, then reshape back to 3d? Need to think about this.)

### DIFF
--- a/benchmarks/prototype/moe_training/benchmark_scaled_grouped_mm_dq.py
+++ b/benchmarks/prototype/moe_training/benchmark_scaled_grouped_mm_dq.py
@@ -28,6 +28,9 @@ device = torch.device("cuda")
 # Needed since changing args to function causes recompiles
 torch._dynamo.config.cache_size_limit = 1000
 
+# Dynamic shapes hurt performance
+torch._dynamo.config.automatic_dynamic_shapes = False
+
 
 @dataclass(frozen=True)
 class ExperimentConfig:
--- a/test/prototype/moe_training/test_kernels.py
+++ b/test/prototype/moe_training/test_kernels.py
@@ -213,7 +213,7 @@ def test_fp8_rowwise_3d_transpose_rhs_reduction(round_scales_to_power_of_2: bool
 @pytest.mark.parametrize(
     "m,k,n_groups", [(256, 256, 4), (16640, 5120, 16), (16640, 8192, 16)]
 )
-def test_mxfp8_per_group_blocked_scales_2d(
+def test_triton_mx_block_rearrange_2d_M_groups(
     m: int,
     k: int,
     n_groups: int,
@@ -271,11 +271,14 @@ def test_mxfp8_per_group_blocked_scales_3d(
     )
 
 
+@pytest.mark.skip(
+    "Temporarily disable and use e2e training numerical tests instead. See: https://github.com/pytorch/ao/pull/2990#discussion_r2354167396"
+)
 @skip_if_rocm("ROCm enablement in progress")
 @pytest.mark.parametrize("m", [256, 512, 1024, 5120])
 @pytest.mark.parametrize("total_k", [512, 1024, 2048, 4096, 8192, 16384])
 @pytest.mark.parametrize("n_groups", [1, 4, 8, 16])
-def test_mxfp8_per_group_blocked_scales_2d2d(
+def test_triton_mx_block_rearrange_2d_K_groups(
     m: int,
     total_k: int,
     n_groups: int,
--- a/torchao/prototype/moe_training/kernels/mxfp8_blocked_scales.py
+++ b/torchao/prototype/moe_training/kernels/mxfp8_blocked_scales.py
@@ -4,6 +4,7 @@ import torch
 import triton
 import triton.language as tl
 from torch import Tensor
+from torch.library import triton_op, wrap_triton
 
 from torchao.prototype.mx_formats.utils import to_blocked
 from torchao.utils import ceil_div
@@ -29,7 +30,14 @@ def torch_to_blocked_2d_M_groups(
 
     assert x_scales.ndim == 2, "x_scales must be 2D"
     assert block_size == 32, "Only block_size=32 is supported for now"
-    blocked_scales_list = []
+    total_M, _ = x_scales.shape
+    num_groups = group_offs.shape[0]
+
+    # Each group will require a variable amount of padding, so to avoid d2h sync causing by iterating over each group,
+    # the Triton kernenl will use an upper bound of adding 128 padding rows to each group.
+    # (This torch impl is used as a reference for correctness, so we must match the triton kernel's impl).
+    total_M_padded = total_M + num_groups * 128
+    blocked_scales = x_scales.new_zeros(total_M_padded, K // block_size)
     start_row_after_padding_list = [0]
     group_start_idx = 0
     for i, group_end_idx in enumerate(group_offs.tolist()):
@@ -42,7 +50,6 @@ def torch_to_blocked_2d_M_groups(
         # Convert group scales to blocked format
         group_scales = x_scales[group_start_idx:group_end_idx]
         group_scales_blocked = to_blocked(group_scales)
-        blocked_scales_list.append(group_scales_blocked)
 
         # Calculate the start row after padding
         scaling_groups_per_row = K // block_size
@@ -50,11 +57,17 @@ def torch_to_blocked_2d_M_groups(
         new_start_row = prev_start_row_after_padding + rows_for_group
         start_row_after_padding_list.append(new_start_row)
 
+        # Write output to subtensor
+        group_rows_padded = ceil_div(group_size, 128) * 128
+        blocked_scales[
+            prev_start_row_after_padding : prev_start_row_after_padding
+            + group_rows_padded,
+            :,
+        ] = group_scales_blocked.reshape(-1, K // block_size)
+
         # Update next group start index
         group_start_idx = group_end_idx
 
-    blocked_scales = torch.cat(blocked_scales_list, dim=0).contiguous()
-    blocked_scales = blocked_scales.reshape(-1, K // 32)
     start_row_after_padding = torch.tensor(
         start_row_after_padding_list, device=x_scales.device, dtype=torch.int64
     )
@@ -79,34 +92,44 @@ def torch_to_blocked_2d_K_groups(
     """
     assert x_scales.ndim == 2, "x_scales must be 2D"
     assert block_size == 32, "Only block_size=32 is supported for now"
-    blocked_scales_list = []
+    M, total_K = x_scales.shape
+    padded_M = ceil_div(M, 128) * 128
+    num_groups = group_offs.shape[0]
+
+    # Each group will require a variable amount of padding, so to avoid d2h sync causing by iterating over each group,
+    # Triton kernel will use an upper bound of adding 4 padding cols to each group.
+    # (This torch impl is used as a reference for correctness, so we must match the triton kernel's impl).
+    total_K_padded = total_K + num_groups * 4
+    blocked_scales = x_scales.new_zeros(padded_M, total_K_padded)
+
     start_col_after_padding_list = [0]
     group_start_idx = 0
     for i, group_end_idx in enumerate(group_offs.tolist()):
         group_size = group_end_idx - group_start_idx
-        prev_start_row_after_padding = start_col_after_padding_list[i]
+        prev_start_col_after_padding = start_col_after_padding_list[i]
         if group_size == 0:
-            start_col_after_padding_list.append(prev_start_row_after_padding)
+            start_col_after_padding_list.append(prev_start_col_after_padding)
             continue
 
         # Convert group scales to blocked format
         group_scales = x_scales[:, group_start_idx:group_end_idx]
         group_scales_blocked = to_blocked(group_scales)
         cols_after_padding = ceil_div(group_size, 4) * 4
-        blocked_scales_list.append(group_scales_blocked)
+
+        # Write output to subtensor
+        blocked_scales[
+            :,
+            prev_start_col_after_padding : prev_start_col_after_padding
+            + cols_after_padding,
+        ] = group_scales_blocked.reshape(-1, cols_after_padding)
 
         # Calculate the start row after padding
-        new_start_col = prev_start_row_after_padding + cols_after_padding
+        new_start_col = prev_start_col_after_padding + cols_after_padding
         start_col_after_padding_list.append(new_start_col)
 
         # Update next group start index
         group_start_idx = group_end_idx
 
-    # blocked_scales = torch.cat(blocked_scales_list, dim=1)
-    M = x_scales.shape[0]
-    padded_M = ceil_div(M, 128) * 128
-    blocked_scales = torch.cat(blocked_scales_list)
-    blocked_scales = blocked_scales.reshape(padded_M, -1)
     start_cols_after_padding = torch.tensor(
         start_col_after_padding_list, device=x_scales.device, dtype=torch.int64
     )
@@ -192,6 +215,7 @@ def compute_blocked_scale_offsets_for_K_groups(
     return group_sizes, starting_col_after_padding
 
 
+@triton_op("torchao::triton_mx_block_rearrange_2d_M_groups", mutates_args={})
 def triton_mx_block_rearrange_2d_M_groups(
     scales_tensor: torch.Tensor,
     input_group_end_offsets: torch.Tensor,
@@ -216,14 +240,16 @@ def triton_mx_block_rearrange_2d_M_groups(
         "Expected element size to be 1 byte (8 bits)"
     )
     rows, cols = scales_tensor.shape
-    num_groups = input_group_end_offsets.numel()
+    num_groups = input_group_end_offsets.shape[0]
 
-    # Final offset is the total number of rows in the tensor
-    padded_rows = output_group_start_offsets[-1]
+    # Final offset is the total number of rows in the tensor.
+    # Padding needing per group is variable/data dependent, so we just pad each group by
+    # the upper bound of 128 rows to avoid a d2h sync caused by iterating over each group.
+    padded_rows = rows + num_groups * 128
 
     num_col_blocks = ceil_div(cols, 4)
     padded_cols = num_col_blocks * 4
-    output = scales_tensor.new_empty((padded_rows, padded_cols))
+    output = scales_tensor.new_zeros((padded_rows, padded_cols))
 
     # Output block stride for the rearranged format
     BLOCK_ROWS, BLOCK_COLS = 128, 4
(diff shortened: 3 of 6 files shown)


danielvegamyhre:
Possible wrap_triton bug:

- Unit tests using compile pass: `pytest test/prototype/moe_training/test_training.py -s`
- Benchmarks using compile hit this error, *inside the custom op* (??):

Command: `python benchmarks/prototype/moe_training/benchmark_scaled_grouped_mm_dq.py --compile`

Update: disabling dynamic shapes resolved the error

Error:
```

  File "/home/danvm/ao/torchao/prototype/moe_training/kernels/mxfp8_blocked_scales.py", line 228, in triton_mx_block_rearrange_2d_M_groups
    output = scales_tensor.new_empty((padded_rows, padded_cols))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
torch._dynamo.exc.TorchRuntimeError: Dynamo failed to run FX node with fake tensors: call_function torchao.triton_mx_block_rearrange_2d_M_groups.default(*(FakeTensor(..., device='cuda:0', size=(16640, 160), dtype=torch.float8_e8m0fnu), FakeTensor(..., device='cuda:0', size=(s21,), dtype=torch.int32), FakeTensor(..., device='cuda:0', size=(s21 + 1,), dtype=torch.int64)), **{}): got RuntimeError('Cannot call numel() on tensor with symbolic sizes/strides\nException raised from throw_cannot_call_with_symbolic at /home/danvm/pytorch/c10/core/TensorImpl.cpp:291 (most recent call first):\nframe #0: c10::Error::Error(c10::SourceLocation, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) + 0x88 (0x7f35c53946c8 in /home/danvm/pytorch/torch/lib/libc10.so)\nframe #1: c10::TensorImpl::throw_cannot_call_with_symbolic(char const*) const + 0x78 (0x7f35c5325476 in /home/danvm/pytorch/torch/lib/libc10.so)\nframe #2: <unknown function> + 0x7938f (0x7f35c537038f in /home/danvm/pytorch/torch/lib/libc10.so)\nframe #3: <unknown function> + 0x49f856 (0x7f35c589f856 in /home/danvm/pytorch/torch/lib/libtorch_python.so)\nframe #4: <unknown function> + 0x484fde (0x7f35c5884fde in /home/danvm/pytorch/torch/lib/libtorch_python.so)\nframe #5: python() [0x545869]\n<omitting python frames>\nframe #11: python() [0x42c4c2]\nframe #13: python() [0x625fb6]\nframe #23: python() [0x457d8d]\nframe #24: python() [0x5d6bb6]\nframe #26: torch::handle_torch_function_no_python_arg_parser(c10::ArrayRef<_object*>, _object*, _object*, char const*, _object*, char const*, torch::TorchFunctionName) + 0x405 (0x7f35c603d6d5 in /home/danvm/pytorch/torch/lib/libtorch_python.so)\nframe #27: <unknown function> + 0x6a1865 (0x7f35c5aa1865 in /home/danvm/pytorch/torch/lib/libtorch_python.so)\nframe #28: <unknown function> + 0x17bd9d3 (0x7f35b93bd9d3 in /home/danvm/pytorch/torch/lib/libtorch_cpu.so)\nframe #29: <unknown function> + 0x6a7d90 (0x7f35c5aa7d90 in /home/danvm/pytorch/torch/lib/libtorch_python.so)\nframe #30: <unknown function> + 0x6a4c8d (0x7f35c5aa4c8d in /home/danvm/pytorch/torch/lib/libtorch_python.so)\nframe #31: <unknown function> + 0xc56c70 (0x7f35c6056c70 in /home/danvm/pytorch/torch/lib/libtorch_python.so)\nframe #32: <unknown function> + 0xc5708d (0x7f35c605708d in /home/danvm/pytorch/torch/lib/libtorch_python.so)\nframe #33: <unknown function> + 0x3b7dc6 (0x7f35c57b7dc6 in /home/danvm/pytorch/torch/lib/libtorch_python.so)\nframe #34: python() [0x5429c4]\nframe #36: python() [0x56b000]\nframe #38: python() [0x457d8d]\nframe #41: <unknown function> + 0xc66254 (0x7f35c6066254 in /home/danvm/pytorch/torch/lib/libtorch_python.so)\nframe #42: <unknown function> + 0xc58d6b (0x7f35c6058d6b in /home/danvm/pytorch/torch/lib/libtorch_python.so)\nframe #43: <unknown function> + 0xc66a96 (0x7f35c6066a96 in /home/danvm/pytorch/torch/lib/libtorch_python.so)\nframe #44: <unknown function> + 0x6a7d90 (0x7f35c5aa7d90 in /home/danvm/pytorch/torch/lib/libtorch_python.so)\nframe #45: <unknown function> + 0x6a4c8d (0x7f35c5aa4c8d in /home/danvm/pytorch/torch/lib/libtorch_python.so)\nframe #46: <unknown function> + 0x17bd7cb (0x7f35b93bd7cb in /home/danvm/pytorch/torch/lib/libtorch_cpu.so)\nframe #47: <unknown function> + 0x6a7d90 (0x7f35c5aa7d90 in /home/danvm/pytorch/torch/lib/libtorch_python.so)\nframe #48: <unknown function> + 0x6a4c8d (0x7f35c5aa4c8d in /home/danvm/pytorch/torch/lib/libtorch_python.so)\nframe #49: <unknown function> + 0x596736b (0x7f35bd56736b in /home/danvm/pytorch/torch/lib/libtorch_cpu.so)\nframe #50: <unknown function> + 0x9e685f (0x7f35c5de685f in /home/danvm/pytorch/torch/lib/libtorch_python.so)\nframe #51: <unknown function> + 0x9e6b5c (0x7f35c5de6b5c in /home/danvm/pytorch/torch/lib/libtorch_python.so)\nframe #52: <unknown function> + 0x8de7e9 (0x7f35c5cde7e9 in /home/danvm/pytorch/torch/lib/libtorch_python.so)\nframe #53: <unknown function> + 0x3b7dc6 (0x7f35c57b7dc6 in /home/danvm/pytorch/torch/lib/libtorch_python.so)\nframe #54: python() [0x5429c4]\nframe #57: python() [0x42c4c2]\nframe #59: python() [0x625fb6]\n')

from user code:
   File "/home/danvm/ao/benchmarks/utils.py", line 12, in fwd_bwd
    out = fn(*args, **kwargs)
  File "/home/danvm/ao/torchao/prototype/moe_training/scaled_grouped_mm.py", line 71, in _scaled_grouped_mm
    return _MXFP8GroupedMM.apply(
  File "/home/danvm/ao/torchao/prototype/moe_training/scaled_grouped_mm.py", line 323, in forward
    A_scales_blocked = triton_mx_block_rearrange_2d_M_groups(
  File "/home/danvm/pytorch/torch/_library/custom_ops.py", line 676, in __call__
    return self._opoverload(*args, **kwargs)

Set TORCHDYNAMO_VERBOSE=1 for the internal stack trace (please do this especially if you're reporting a bug to PyTorch). For even more developer context, set TORCH_LOGS="+dynamo"
```

cc @zou3519 any idea what could be going on here? this is with torch built from source on 9/10

danielvegamyhre:
Disable dynamic shapes worked as a workaround

danielvegamyhre:
Only forward is getting compiled for some reason, not backward: https://manifold.edge.x2p.facebook.net/v0/read/tree/logs/.tmp7B4CM7/index.html?bucketName=tlparse_reports&apiKey=tlparse_reports-key&withPayload=1&timeoutMsec=10000

This is with torch built from source with cuda 12.9 at 53b8bdb97774114ca02948fed47f2fd49996c564 (sept 12th)
Screenshot 2025-09-12 at 1 21 38 PM

zou3519:
Do you have a function in C++ that calls Tensor.sizes()? The main fix for that error message is to change Tensor.sizes() in C++ to Tensor.sym_sizes()

torchao/prototype/moe_training/tensor.py
```diff
@@ -125,28 +125,31 @@ def unwrap(t):
...
         if func not in _ops_to_preserve_subclass:
             return out
 
-        # wrap outputs back into ScaledGroupedMMTensor for ops that do preserve subclass
-        return pytree.tree_map_only(
-            torch.Tensor,
-            lambda x: ScaledGroupedMMTensor(x, scaling_type),
-            out,
-        )
+        # Only wrap tensor outputs, prevent double wrapping
```
bdhirsh: what would cause double wrapping to have happened previously? I would expect that as long as `args_unwrapped` above only consists of plain tensors, then `out` here should be only plain tensors as well

torchao/prototype/moe_training/tensor.py
```diff
@@ -27,7 +27,7 @@
     torch.ops.aten.copy_.default,
     torch.ops.aten.view.default,
     torch.ops.aten.as_strided.default,
-    torch.ops.aten._to_copy.default,
+    torch.ops.aten._to_copy.default,  # for *.to(dtype)
```
bdhirsh, line 30: it seems a bit strange to me that `_to_copy` should **always** preserve the subclass. For example - if I do `aten._to_copy(mxfp8_tensor, dtype=torch.bfloat16)`, I would expect the output to **not** be a subclass, right?

Put another way - I'd imagine that you need to carefully implement `_to_copy` to only preserve the subclass if we are not doing a cross-dtype copy to a higher precision

torchao/prototype/moe_training/tensor.py
danielvegamyhre: Tried to repro the original issue to answer this, but now using the latest torch nightly build it doesn't repro anymore....  good news I guess

torchao/prototype/moe_training/tensor.py
danielvegamyhre: >For example - if I do aten._to_copy(mxfp8_tensor, dtype=torch.bfloat16), I would expect the output to not be a subclass, right?

It would seem that way, but we actually do want to preserve the subclass in this case, because of how torchtitan integrates with grouped_mm [here](https://github.com/pytorch/torchtitan/blob/main/torchtitan/models/moe.py#L117)

Basically, kernels that torch._grouped_mm and torch._scaled_grouped_mm dispatch to only accept bf16 as the input/output dtype, yet torchtitan wants to support doing training in full fp32 precision, so they do this cast here.

benchmarks/prototype/moe_training/benchmark_scaled_grouped_mm_dq.py
```diff
@@ -28,6 +28,9 @@
 # Needed since changing args to function causes recompiles
 torch._dynamo.config.cache_size_limit = 1000
 
+# Workaround for https://github.com/pytorch/ao/pull/2990#issuecomment-3285762681
+torch._dynamo.config.automatic_dynamic_shapes = False
```
vkuzo, line 32: is this needed for fast e2e? if yes, add to README.md?

danielvegamyhre: Oh, we can remove this workaround now - seems like using the latest nightly, the issue is resolved. Updated.

danielvegamyhre: So actually yes, while the bug/crash no longer occurs with dynamic shapes, perf is worse with dynamic shapes: https://www.internalfb.com/phabricator/paste/view/P1949814669

So I think we should recommend disabling. Will make a note of this to put in the readme

drisspg: No we dont need to recommend in the readme. Benchmarks artificially hit automatic dynamic shapes because we are sweeping certain dims. In practice the dims that are being swept won't be dynamic or if they are then its up to the users to decide how they want to handle this -> dynamic=false w/ more recompiles or the default behavior

test/prototype/moe_training/test_kernels.py
```diff
@@ -272,10 +272,10 @@ def test_mxfp8_per_group_blocked_scales_3d(
 
 
 @skip_if_rocm("ROCm enablement in progress")
-@pytest.mark.parametrize("m", [256, 512, 1024, 5120])
-@pytest.mark.parametrize("total_k", [512, 1024, 2048, 4096, 8192, 16384])
-@pytest.mark.parametrize("n_groups", [1, 4, 8, 16])
-def test_mxfp8_per_group_blocked_scales_2d2d(
+@pytest.mark.parametrize("m", [256])
```
drisspg: Intentional?

torchao/prototype/moe_training/kernels/mxfp8_blocked_scales.py
```diff
@@ -216,14 +240,14 @@ def triton_mx_block_rearrange_2d_M_groups(
         "Expected element size to be 1 byte (8 bits)"
     )
     rows, cols = scales_tensor.shape
-    num_groups = input_group_end_offsets.numel()
+    num_groups = input_group_end_offsets.shape[0]
 
     # Final offset is the total number of rows in the tensor
-    padded_rows = output_group_start_offsets[-1]
+    padded_rows = rows + num_groups * 128
```
drisspg, line 248: can you write a `note: padding to max_possible: ...`
and then link to that note throughout the codebase this is a pretty common pattern

torchao/prototype/moe_training/kernels/mxfp8_blocked_scales.py
```diff
@@ -479,13 +505,13 @@ def triton_mx_block_rearrange_2d_K_groups(
...
     # Calculate blocks needed
-    num_groups = input_group_end_offsets.numel()
+    num_groups = input_group_end_offsets.shape[0]
     num_row_blocks = ceil_div(rows, 128)
     padded_rows = num_row_blocks * 128
 
     # output_group_start_offsets always starts with 0 and ends with the total number of cols
-    padded_cols = output_group_start_offsets[-1]
-    output = scales_tensor.new_empty((padded_rows, padded_cols))
+    padded_cols = cols + num_groups * 4
```
drisspg, line 516: ditto

torchao/prototype/moe_training/kernels/mxfp8_blocked_scales.py
```diff
@@ -79,34 +92,44 @@ def torch_to_blocked_2d_K_groups(
     """
     assert x_scales.ndim == 2, "x_scales must be 2D"
     assert block_size == 32, "Only block_size=32 is supported for now"
-    blocked_scales_list = []
+    M, total_K = x_scales.shape
```
danielvegamyhre, line 95: Note for reviewers: the changes to the torch reference implementation `torch_to_blocked_2d_K_groups` (to account for upper-bound based padding, i.e. `total_padded_K = cols + num_groups * 4`) is not working properly yet. It does not match the triton kernel in unit tests, but I am sure the triton kernel is correct because the e2e training tests validate correct gradient numerics. As further evidence, replacing the triton kernel with this torch impl in the training code, and the gradient numerics are garbage. 

Need to figure out the correct way to represent "row of blocks"-major PER GROUP in torch native code, before landing this.

test/prototype/moe_training/test_kernels.py
danielvegamyhre: Nope, was for debugging this issue: https://github.com/pytorch/ao/pull/2990/files#r2354167396 which I still need to figure out. Anyway just reverted the test cases.

PR #2997, opened 2025-09-13
Support Int4OpaqueTensor for AWQ

cyxlily:
Add act_pre_scale into Int4OpaqueTensor for AWQ.

### DIFF
--- a/test/prototype/test_awq.py
+++ b/test/prototype/test_awq.py
@@ -8,6 +8,7 @@ import tempfile
 import unittest
 
 import torch
+from parameterized import parameterized
 from torch.testing._internal.common_utils import (
     TestCase,
     run_tests,
@@ -41,12 +42,14 @@ class ToyLinearModel(torch.nn.Module):
         x = self.linear3(x)
         return x
 
+devices = ["cpu"]
+if (
+    torch.cuda.is_available()
+    and _is_fbgemm_genai_gpu_available()
+    and TORCH_VERSION_AT_LEAST_2_6
+):
+    devices.append("cuda")
 
-@unittest.skipIf(not torch.cuda.is_available(), reason="CUDA not available")
-@unittest.skipIf(
-    not _is_fbgemm_genai_gpu_available(),
-    reason="need to install fbgemm_gpu_genai package",
-)
 class TestAWQ(TestCase):
     def test_awq_config(self):
         base_config = Int4WeightOnlyConfig()
@@ -61,8 +64,8 @@ class TestAWQ(TestCase):
         with self.assertRaisesRegex(ValueError, "is not one of"):
             AWQConfig(base_config, step="not_supported")
 
-    def test_awq_functionality(self):
-        device = "cuda"
+    @parameterized.expand([(device,) for device in devices])
+    def test_awq_functionality(self, device):
         dataset_size = 100
         l1, l2, l3 = 512, 256, 128
         original_dtype = torch.bfloat16  # tinygemm kernel only uses bfloat16 inputs
@@ -73,7 +76,15 @@ class TestAWQ(TestCase):
         m = ToyLinearModel(l1, l2, l3).eval().to(original_dtype).to(device)
 
         # baseline quantization
-        base_config = Int4WeightOnlyConfig(group_size=group_size)
+        if device == "cuda":
+            base_config = Int4WeightOnlyConfig(group_size=group_size, version=2)
+        elif device == "cpu":
+            base_config = Int4WeightOnlyConfig(
+                group_size=group_size, packing_format="opaque", version=2
+            )
+            torch.manual_seed(1234)
+        else:
+            assert False, "Unsupported device: {}".format(device)
         m_baseline = copy.deepcopy(m)
         quantize_(m_baseline, base_config)
 
@@ -102,10 +113,11 @@ class TestAWQ(TestCase):
 
         loss_awq = (ref_out - awq_out).pow(2).mean().item()
         loss_base = (ref_out - baseline_out).pow(2).mean().item()
+
         assert loss_awq < loss_base
 
-    def test_awq_loading(self):
-        device = "cuda"
+    @parameterized.expand([(device,) for device in devices])
+    def test_awq_loading(self, device):
         dataset_size = 100
         l1, l2, l3 = 512, 256, 128
         original_dtype = torch.bfloat16  # tinygemm kernel only uses bfloat16 inputs
@@ -123,7 +135,14 @@ class TestAWQ(TestCase):
         calibration_data = dataset[:n_calibration_examples]
 
         # calibrate
-        base_config = Int4WeightOnlyConfig(group_size=group_size)
+        if device == "cuda":
+            base_config = Int4WeightOnlyConfig(group_size=group_size, version=2)
+        elif device == "cpu":
+            base_config = Int4WeightOnlyConfig(
+                group_size=group_size, packing_format="opaque", version=2
+            )
+        else:
+            assert False, "Unsupported device: {}".format(device)
         quant_config = AWQConfig(base_config, step=AWQStep.PREPARE)
         quantize_(m, quant_config)
 
@@ -152,14 +171,14 @@ class TestAWQ(TestCase):
         assert awq_save_load_out is not None
         assert torch.allclose(awq_out, awq_save_load_out, atol=1e-2)
 
-    def test_awq_loading_vllm(self):
+    @parameterized.expand([(device,) for device in devices])
+    def test_awq_loading_vllm(self, device):
         """Simulate weight loading in vllm:
         * prepare model weight to the same format (awq weight)
         * use weight.copy_(state_dict["weight"]) to copy over the quantized weights from checkpoint
 
         There is also a slicing op that is ommitted here, overall e2e is tested in tests in vllm repo
         """
-        device = "cuda"
         dataset_size = 100
         l1, l2, l3 = 512, 256, 128
         original_dtype = torch.bfloat16  # tinygemm kernel only uses bfloat16 inputs
@@ -177,7 +196,14 @@ class TestAWQ(TestCase):
         calibration_data = dataset[:n_calibration_examples]
 
         # calibrate
-        base_config = Int4WeightOnlyConfig(group_size=group_size)
+        if device == "cuda":
+            base_config = Int4WeightOnlyConfig(group_size=group_size, version=2)
+        elif device == "cpu":
+            base_config = Int4WeightOnlyConfig(
+                group_size=group_size, packing_format="opaque", version=2
+            )
+        else:
+            assert False, "Unsupported device: {}".format(device)
         quant_config = AWQConfig(base_config, step=AWQStep.PREPARE)
         quantize_(m, quant_config)
 
--- a/test/quantization/quantize_/workflows/int4/test_int4_opaque_tensor.py
+++ b/test/quantization/quantize_/workflows/int4/test_int4_opaque_tensor.py
@@ -20,6 +20,7 @@ from torchao.quantization import (
     quantize_,
 )
 from torchao.quantization.utils import compute_error
+from torchao.quantization.quantize_.common import SupportsActivationPreScaling
 from torchao.utils import (
     torch_version_at_least,
 )
@@ -28,7 +29,8 @@ from torchao.utils import (
 def get_config(group_size):
     return Int4WeightOnlyConfig(
         group_size=group_size,
-        int4_packing_format="opaque",
+        packing_format="opaque",
+        version=2,
     )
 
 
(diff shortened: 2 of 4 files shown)


test/prototype/test_awq.py
```diff
@@ -73,7 +76,15 @@ def test_awq_functionality(self):
         m = ToyLinearModel(l1, l2, l3).eval().to(original_dtype).to(device)
 
         # baseline quantization
-        base_config = Int4WeightOnlyConfig(group_size=group_size)
+        if device == "cuda":
+            base_config = Int4WeightOnlyConfig(group_size=group_size, version=2)
```
Xia-Weiwen: The original code for CUDA does not specify `version=2` but the new code does it. Do we need this change?

test/prototype/test_awq.py
```diff
@@ -123,7 +135,14 @@ def test_awq_loading(self):
         calibration_data = dataset[:n_calibration_examples]
 
         # calibrate
-        base_config = Int4WeightOnlyConfig(group_size=group_size)
+        if device == "cuda":
+            base_config = Int4WeightOnlyConfig(group_size=group_size, version=2)
```
Xia-Weiwen: ditto

test/prototype/test_awq.py
```diff
@@ -177,7 +196,14 @@ def test_awq_loading_vllm(self):
...
```
Xia-Weiwen: ditto

test/quantization/quantize_/workflows/int4/test_int4_opaque_tensor.py
```diff
@@ -75,7 +77,23 @@ def test_module_path(self, dtype):
...
+        assert isinstance(qw, SupportsActivationPreScaling), (
+            "Expected int4 tensor supports activation prescaling"
+        )
+        assert qw.act_pre_scale is None, "Default `act_pre_scale` is None"
+        _ACT_PRE_SCALE = 2
+        qw.act_pre_scale = _ACT_PRE_SCALE
+        quantized = linear(input)
 
+        # making sure activation pre scaling is successfully applied to the activation
+        self.assertTrue(compute_error(original * _ACT_PRE_SCALE, quantized) > 20)
```
Xia-Weiwen: Why does this test "activation pre scaling is successfully applied to the activation" as you explained in the comment? What if `compute_error(original * _ACT_PRE_SCALE, quantized)` becomes < 20 after we apply the scale?

test/quantization/quantize_/workflows/int4/test_int4_opaque_tensor.py
```diff
@@ -75,7 +77,23 @@ def test_module_path(self, dtype):
...
+    def test_activation_prescaling(self):
+        dtype = torch.bfloat16
+        input = torch.randn(1, 128, dtype=dtype)
+        linear = torch.nn.Linear(128, 256, bias=False, dtype=dtype)
+        original = linear(input)
+        quantize_(linear, get_config(group_size=128))
+        qw = linear.weight
+        assert isinstance(qw, SupportsActivationPreScaling), (
+            "Expected int4 tensor supports activation prescaling"
+        )
```
Xia-Weiwen, line 89: Could you explain why `qw` is an instance of `SupportsActivationPreScaling`? `qw` should be an instance of `Int4OpaqueTensor`, right?

torchao/prototype/awq/example.py
```diff
@@ -254,14 +295,21 @@ def quantize_and_eval(
         quantize_(model, quant_config)
         print(f"time for convert: {time.time() - t0:.02f} seconds")
         quant_config = AWQConfig(base_config, step="prepare_for_loading")
-        model.config.quantization_config = TorchAoConfig(quant_config)
+        #model.config.quantization_config = TorchAoConfig(quant_config)
```
Xia-Weiwen: Is this change needed?

Xia-Weiwen (commented):
LGTM but I have a few questions.

Xia-Weiwen:

torchao/prototype/awq/example.py
```diff
@@ -223,10 +259,15 @@ def quantize_and_eval(
...
 
-        base_config = Int4WeightOnlyConfig(group_size=group_size)
+        if device == "cuda":
+            base_config = Int4WeightOnlyConfig(group_size=group_size, version=2)
+        elif device == "cpu":
+            base_config = Int4WeightOnlyConfig(
+                group_size=group_size, packing_format="opaque", version=2
+            )
+        else:
+            assert False, "Unsupported device: {}".format(device)
```
mingfeima: i am not very familar with the concept here, could you explain why cpu needs `opaque` packing_format?

torchao/prototype/awq/example.py
```diff
@@ -254,14 +295,21 @@ def quantize_and_eval(
...
         group_size = int(quant.split("-")[1])
         print(f"running {quant} quantization with group size {group_size}")
         # TODO: enable after migration: https://github.com/pytorch/ao/issues/2752
         # use_hqq = "hqq" in quant
-        base_config = Int4WeightOnlyConfig(group_size=group_size, version=2)
+        if device == "cuda":
+            base_config = Int4WeightOnlyConfig(group_size=group_size, version=2)
+        elif device == "cpu":
+            base_config = Int4WeightOnlyConfig(
+                group_size=group_size, packing_format="opaque", version=2
```
jerryzh168: version=2 can be removed now, it's the default now

torchao/prototype/awq/example.py
jerryzh168: it's because packing_format describes a fix format of how the quantized weight data are laid out in memory, but int4 cpu has a format that is based on specific hardwares/tensor shapes etc.: https://github.com/pytorch/ao/blob/045c9597a858ac6fb8b590c9adeb102877cfe10b/torchao/quantization/quantize_/workflows/int4/int4_opaque_tensor.py#L45-L47

test/quantization/quantize_/workflows/int4/test_int4_opaque_tensor.py
cyxlily: > Could you explain why `qw` is an instance of `SupportsActivationPreScaling`? `qw` should be an instance of `Int4OpaqueTensor`, right?

qw is both an instance of Int4OpaqueTensor and an instance of SupportsActivationPreScaling. Because SupportsActivationPreScaling is a Protocol (a formal interface). This protocol defines a single, specific requirement: any class that has an act_pre_scale: Optional[torch.Tensor] attribute is automatically considered to implement the protocol. Therefore, the Int4OpaqueTensor class is written to include this required act_pre_scale attribute. Because it has this attribute, it satisfies the SupportsActivationPreScaling protocol. The @runtime_checkable decorator allows Python's isinstance() function to verify this compliance at runtime.
https://github.com/pytorch/ao/blob/main/torchao/quantization/quantize_/common/protocol.py#L14-L22

Xia-Weiwen: Thanks for the explanation

test/quantization/quantize_/workflows/int4/test_int4_opaque_tensor.py
cyxlily: > Why does this test "activation pre scaling is successfully applied to the activation" as you explained in the comment? What if `compute_error(original * _ACT_PRE_SCALE, quantized)` becomes < 20 after we apply the scale?

This test is not checking if the quantization is accurate, but rather that the activation pre-scaling mechanism is functioning correctly. The > 20 condition is a sanity check to ensure the scaled quantization (input * 2 before INT4 quantization) produces a result that is significantly different from the original (BF16 result * 2), proving the scale factor was applied. 
If compute_error(original * _ACT_PRE_SCALE, quantized) became < 20, it would mean that the quantized result with pre-scaling is almost identical to the exactly scaled FP16 result, which implies that the pre-scaling or quantization isn't being applied correctly.

Xia-Weiwen: Thanks. Could you update the comments to make it clearer? Needless to add too many details though.

test/prototype/test_awq.py
cyxlily: You are right, we don't need version=2 for CUDA, I update and remove this change.

test/prototype/test_awq.py
cyxlily: Removed

test/prototype/test_awq.py
cyxlily: Removed

test/quantization/quantize_/workflows/int4/test_int4_opaque_tensor.py
cyxlily: Sure, I update comments and tests.

torchao/prototype/awq/example.py
cyxlily: No, I update and remove this change.

test/prototype/test_awq.py
```diff
@@ -5,17 +5,17 @@
...
+from parameterized import parameterized
 from torch.testing._internal.common_utils import (
     TestCase,
     run_tests,
 )
 
 from torchao.prototype.awq import AWQConfig, AWQStep
 from torchao.quantization import Int4WeightOnlyConfig, quantize_
-from torchao.utils import _is_fbgemm_genai_gpu_available
+from torchao.utils import TORCH_VERSION_AT_LEAST_2_6, _is_fbgemm_genai_gpu_available
```
Xia-Weiwen: Please use the new `torch_version_at_least("2.6.0") instead.

torchao/prototype/awq/example.py
```diff
@@ -6,17 +6,23 @@
...
     AWQConfig,
 )
-from torchao.quantization import (
-    quantize_,
-)
+from torchao.quantization import Int4WeightOnlyConfig, quantize_
+
+inductor_config.cpp_wrapper = True
+inductor_config.max_autotune = True
+inductor_config.max_autotune_gemm_backends = "CPP,ATEN"
```
Xia-Weiwen: This script will also be used for CUDA. So, I think `triton` is needed here. Or let's simply remove this line to use the default ones.

torchao/prototype/awq/example.py
cyxlily: Thanks, I update and remove version=2

test/quantization/quantize_/workflows/int4/test_int4_opaque_tensor.py
```diff
@@ -76,6 +78,43 @@ def test_module_path(self, dtype):
...
+        _ACT_PRE_SCALE = 2
+        manual_scaled_quantized = linear1(input * _ACT_PRE_SCALE)
+        qw2 = linear2.weight
+        qw2.act_pre_scale = _ACT_PRE_SCALE
+        auto_scaled_quantized = linear2(input)
+
+        # Making sure activation pre scaling is successfully applied to the activation.
+        # manual_scaled_quantized (input * 2 → quantize with act_pre_scale=None) should equal
+        # auto_scaled_quantized (original input → quantize with act_pre_scale=2),
+        # Proving that the act_pre_scale factor correctly applies input scaling
```
Xia-Weiwen:
```suggestion
        # Make sure activation pre scaling is successfully applied to the activation.
```
Let's make it more concise.

test/quantization/quantize_/workflows/int4/test_int4_opaque_tensor.py
```diff
@@ -76,6 +78,43 @@ def test_module_path(self, dtype):
...
+        self.assertEqual(manual_scaled_quantized, auto_scaled_quantized)
+
+        # Making sure quantization with pre-scaling is successfully applied to the activation.
+        # The error > 20 indicats that quantized computation with activation pre-scaling
+        # produces significantly different results from simply scaling the original
+        # floating-point output, confirming that pre-scaling is applied during
+        # quantization rather than post-processing.
```
Xia-Weiwen:
```suggestion
        # If pre-scaling is auto-applied, the quantization error should be low, i.e., compute_error (SQNR) is high
```
This may be simpler.

torchao/prototype/awq/example.py
Xia-Weiwen: Please remove the `version=2` here since it's the default.

Xia-Weiwen (approved):
Thanks.

test/quantization/quantize_/workflows/int4/test_int4_opaque_tensor.py
```diff
@@ -28,7 +29,7 @@
 def get_config(group_size):
     return Int4WeightOnlyConfig(
         group_size=group_size,
-        int4_packing_format="opaque",
+        packing_format="opaque",
```
jerryzh168: this should be `int4_packing_format` I think

test/quantization/quantize_/workflows/int4/test_int4_opaque_tensor.py
```diff
@@ -76,6 +77,37 @@ def test_module_path(self, dtype):
...
+        qw2.act_pre_scale = _ACT_PRE_SCALE
+        auto_scaled_quantized = linear2(input)
+
+        # Making sure activation pre scaling is successfully applied to the activation.
+        self.assertEqual(manual_scaled_quantized, auto_scaled_quantized)
+
+        # If pre-scaling is auto-applied, the quantization error should be low,
+        # i.e., compute_error (SQNR) is high
+        self.assertTrue(
+            compute_error(original * _ACT_PRE_SCALE, auto_scaled_quantized) > 20
```
jerryzh168: nit: `original` --> `original_output`

torchao/prototype/awq/example.py
cyxlily: Removed

test/quantization/quantize_/workflows/int4/test_int4_opaque_tensor.py
cyxlily: Thanks, updated

test/quantization/quantize_/workflows/int4/test_int4_opaque_tensor.py
cyxlily: Thanks, updated

test/quantization/quantize_/workflows/int4/test_int4_opaque_tensor.py
cyxlily: Thanks, updated

test/quantization/quantize_/workflows/int4/test_int4_opaque_tensor.py
cyxlily: Thanks, updated

torchao/prototype/awq/example.py
cyxlily: OK, I removed version=2 in CPU

PR #2998, opened 2025-09-13
[mxfp8 moe training] use dim1 cast cuda kernel for 3d weights by reshaping to 2d

danielvegamyhre:
Stacked PRs:
 * #3004
 * #3002
 * #2999
 * __->__#2998

--- --- ---

[mxfp8 moe training] use dim1 cast cuda kernel for 3d weights by reshaping to 2d

- Doing `to_mx(B_t.contiguous())` is unspeakably slow (see perf analysis in previous PR in stack)
- As a workaround, we can use the faster dim1 cast cuda kernel by reshaping the 3d weights to 2d, casting, then reshaping back 2d. I wasn't able to find a way to reshape the 2d, column major quantized tensor -> 3d column major tensor, so I was forced to use .t().contiguous().t() pattern, which is not ideal for perf, yet still faster than doing `to_mx(B_t.contiguous())`.

## Next steps
- Based on benchmarks and traces, quantizing 3d expert weights scales poorly as number of experts increases, both with the `to_mx` method and the 2d dim1 cast CUDA kernel method. This is likely due to the .contiguous() call required for both methods.
- We should update the dim1 cast CUDA kernel to handle 3 inputs, writing directly to col major format, so we can avoid this expensive transformation to column major.
    - We could also update the CUTLASS grouped gemm to handle NT/TN/TT/NN layouts but I think passing in args with different memory layouts could affect kernel perf, need to think about this more.

## Test plan
- `pytest test/prototype/moe_training/test_training.py `

## Benchmarks
Before:
```
A_shape        B_shape           recipe                  bf16_e2e_us    scaled_e2e_us  scaled_e2e_speedup      bf16_fwd_us    scaled_fwd_us  scaled_fwd_speedup
-------------  ----------------  --------------------  -------------  ---------------  --------------------  -------------  ---------------  --------------------
(16640, 5120)  (1, 8192, 5120)   MoEScalingType.MXFP8        4268.5           3402.75  1.254x                      1513.76          1675.81  0.903x
(16640, 5120)  (4, 8192, 5120)   MoEScalingType.MXFP8        3968.88          4282.53  0.927x                      1126.21          2222.37  0.507x
(16640, 5120)  (16, 8192, 5120)  MoEScalingType.MXFP8        4900.77          8091.55  0.606x                      1262.66          9047.7   0.14x
(16640, 5120)  (64, 8192, 5120)  MoEScalingType.MXFP8        8432.61         21453.3   0.393x                      1788.94         14476.4   0.124x
```

After:
```
A_shape        B_shape           recipe                  bf16_e2e_us    scaled_e2e_us  scaled_e2e_speedup      bf16_fwd_us    scaled_fwd_us  scaled_fwd_speedup
-------------  ----------------  --------------------  -------------  ---------------  --------------------  -------------  ---------------  --------------------
(16640, 5120)  (1, 8192, 5120)   MoEScalingType.MXFP8        4920.32          3057.79  1.609x                      1299.92          1091.58  1.191x
(16640, 5120)  (4, 8192, 5120)   MoEScalingType.MXFP8        3886.21          3402.82  1.142x                      1087.39           931.36  1.168x
(16640, 5120)  (16, 8192, 5120)  MoEScalingType.MXFP8        5769.02          5384.91  1.071x                      1411.1           1222.7   1.154x
(16640, 5120)  (64, 8192, 5120)  MoEScalingType.MXFP8        8455.23         12846.2   0.658x                      1796.1           2968.21  0.605x
```

### DIFF
--- a/benchmarks/utils.py
+++ b/benchmarks/utils.py
@@ -25,8 +25,13 @@ def bench_fwd_bwd_microseconds(
 
 def bench_fwd_microseconds(fn, *args, use_compile=False, fullgraph=True, **kwargs):
     fn_compiled = torch.compile(fn, fullgraph=fullgraph) if use_compile else fn
+
+    def inference_fn(*args, **kwargs):
+        with torch.no_grad():
+            return fn_compiled(*args, **kwargs)
+
     return benchmark_cuda_function_in_microseconds(
-        fn_compiled,
+        inference_fn,
         *args,
         **kwargs,
     )
--- a/torchao/prototype/moe_training/scaled_grouped_mm.py
+++ b/torchao/prototype/moe_training/scaled_grouped_mm.py
@@ -58,7 +58,6 @@ def _scaled_grouped_mm(
     """
     # TODO: Remove logging once prototype is more mature. This is currently very useful for development and debugging.
     if scaling_type == MoEScalingType.FP8_ROWWISE:
-        logger.debug("Using fp8 rowwise for _scaled_grouped_mm")
         return _Float8GroupedMM.apply(
             A,
             B_t,
@@ -66,7 +65,6 @@ def _scaled_grouped_mm(
             out_dtype,
         )
     elif scaling_type == MoEScalingType.MXFP8:
-        logger.debug("Using mxfp8 for _scaled_grouped_mm")
         block_size = 32  # TODO: should we make this configurable? plumb it through in a config somehow?
         return _MXFP8GroupedMM.apply(
             A,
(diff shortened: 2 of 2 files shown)


benchmarks/utils.py
```diff
@@ -25,8 +25,13 @@ def fwd_bwd(*args, **kwargs):
 
 def bench_fwd_microseconds(fn, *args, use_compile=False, fullgraph=True, **kwargs):
     fn_compiled = torch.compile(fn, fullgraph=fullgraph) if use_compile else fn
+
+    def inference_fn(*args, **kwargs):
```
vkuzo, line 29: why do we need this? forward in training is not run with `torch.inference_mode()`

danielvegamyhre: I know, I just wanted to measure just the quantizations ops + grouped gemm needed to produce the forward output, without pre-computing and saving things for backward, so that I can more easily see if the quant ops for fwd or bwd are what is slow without generating a trace to look at. IIRC if I didn't use inference mode, the forward graph still precomputed stuff.

drisspg: just use `no_grad`

drisspg:
Crawling up the stack but do we need this if we are adding a kernel higher up?

danielvegamyhre:
> Crawling up the stack but do we need this if we are adding a kernel higher up?

Yes, for now, because it is still currently the fastest method when E < 8. New CUDA Kernel for 3d tensors is only faster for E>8 (see relevant PR). I want to change this though and just use one kernel everywhere, would appreciate your thoughts on the kernel design. I checked NCU and it flags some minor (~15%) potential speedups from resolving bank conflicts and uncoalesced global accesses but am not confident that is really the issue, since those are present for the 2d kernel as well.

PR #2999, opened 2025-09-13
[moe training] add benchmarks for dsv3 236b, 671b shapes; reorganize benchmarks dir

danielvegamyhre:
Stacked PRs:
 * #3004
 * #3002
 * __->__#2999
 * #2998

--- --- ---

[moe training] organize bench dir by quant type; rename bench scripts to match kernel names

### Benchmarks for Llama4 16e, DSV3 236b, DSV3 671b with various experts per device (i.e., different EP degrees)

Llama4 16e shapes with 1-8 experts per device
```
M,N,K,G                  recipe                  bf16_fwd_bwd_us    scaled_fwd_bwd_us  scaled_fwd_bwd_speedup      bf16_fwd_us    scaled_fwd_us  scaled_fwd_speedup
-----------------------  --------------------  -----------------  -------------------  ------------------------  -------------  ---------------  --------------------
(16384, 8192, 5120, 1)   MoEScalingType.MXFP8           4403.49              2992.1    1.472x                         1236.1            823.344  1.501x
(16384, 8192, 5120, 2)   MoEScalingType.MXFP8           4225.02              3198      1.321x                         1180.58           795.648  1.484x
(16384, 8192, 5120, 4)   MoEScalingType.MXFP8           4100.06              3540.16   1.158x                         1150.82           918.576  1.253x
(16384, 8192, 5120, 8)   MoEScalingType.MXFP8           4246.43              3752.99   1.131x                         1192.67          1029.31   1.159x
(128000, 8192, 5120, 1)  MoEScalingType.MXFP8          42062.4              24263.6    1.734x                        10494             5960.54   1.761x
(128000, 8192, 5120, 2)  MoEScalingType.MXFP8          47400.4              26359.8    1.798x                        22849.5           7594.02   3.009x
(128000, 8192, 5120, 4)  MoEScalingType.MXFP8          36716.6              25263.1    1.453x                        10177             6054.02   1.681x
(128000, 8192, 5120, 8)  MoEScalingType.MXFP8          35439.1              27204.2    1.303x                        10147.6           5802.88   1.749x
```

DSV3 671b shapes with 1-8 experts per device
```
(16384, 2048, 7168, 1)   MoEScalingType.MXFP8           1628.3               1533.95   1.062x                          455.536          419.872  1.085x
(16384, 2048, 7168, 2)   MoEScalingType.MXFP8           1456.1               1496.16   0.973x                          457.536          415.856  1.1x
[...]

PR #2513, opened 2025-07-09
Add CUDA kernel for MXFP8 dim1 casting

danielvegamyhre:
# Add CUDA kernel for MXFP8 dim1 casting

Co-authored-by: Less Wright <lessw@etrillium.com>

## Summary
- Add a CUDA kernel to do mxfp8 dim1 casting, which benchmarks show is ~1.4x faster than the existing Triton kernel, and benchmarking Llama3 8b training with torchtitan show this translating to a 1.5% - 2.5% e2e training speedup. This prototype was developed/explored in a personal repo, and now this PR begins the migration into torchao.
- We used this [TE kernel](https://github.com/NVIDIA/TransformerEngine/blob/4c7095ca94da5c44db9a5ba606a483c9751ae076/transformer_engine/common/util/cast_kernels.cuh#L61) as a starting point (big thanks to @lessw2020 for setting up this C++ extension with the parts we needed to iterate on! https://github.com/danielvegamyhre/private-torchao/pull/1)
- Subsequent PRs will migrate the integration, torch.compile support etc. to make reviewing easier. This PR is only adds (1) the CUDA kernel, (2) the C++ extension to make it usable in Python, (3) numerical test, and (4) kernel benchmarking. 

## Numerics changes
I made the following changes to get matching numerics for the columnwise/dim1 scaling path:
- Implemented the floor scaling mode recipe, to match numerics https://github.com/danielvegamyhre/private-torchao/pull/9
- Make scaling mode (floor, rceil) configurable https://github.com/danielvegamyhre/private-torchao/pull/16
- Add support for both fp32 and bf16 input tensor types. https://github.com/danielvegamyhre/private-torchao/pull/10
- Updates scale shapes to match torchao https://github.com/danielvegamyhre/private-torchao/pull/8

## Performance improvements
I also made the following changes to improve perf for the columnwise/dim1 scaling path:
- Fix uncoalesced global accesses when writing colwise scales https://github.com/danielvegamyhre/private-torchao/pull/11
- Use reciprocal multiplication instead of division in loop along dim1 scaling group https://github.com/danielvegamyhre/private-torchao/pull/12
- Write quantized output to col-major directly via TMA smem -> gmem changes, to avoid extra memory layout  transformation on output to get in required memory layout for the GEMM https://github.com/danielvegamyhre/private-torchao/pull/18
[...]

PR #3002, opened 2025-09-14
[mxfp8 moe training] add CUDA kernel to quantize 3d tensor colwise

danielvegamyhre:
Stacked PRs:
 * #3034
 * #3033
 * #3004
 * __->__#3002

--- --- ---

[mxfp8 moe training] add CUDA kernel to quantize 3d tensor colwise

## Summary
- This PR adds a new CUDA kernel specifically for quantizing 3d expert weights shape (E,N,K) along the N dimension and writing directly to column major format. 
     - Design: I create separate input/output TMA descriptors for each expert, and process each 2d expert in parallel using the same method that the 2d dim1 quantization kernel uses. The 2d kernel achieves 85% peak memory bandwidth utilization, so hopefully we can achieve similar perf for 3d.
- The existing methods for quantizing 3d expert weights both scale very poorly. I have verified this via benchmarking and traces (see previous PR), and hypothesize that it is due to required `.contiguous()` calls:
    - Using `to_mx` only scales along the last dim and requires contiguos inputs. So this requires transposing contiguous tensor (E,N,K)  -> (E,K,N) then calling .contiguous() to scale along the N dim (needed for backwards)
    - Using the existing CUDA kernel for casting along dim1 is possible, by treating the 3d input tensor as a 2d tensor of shape (E*N, K). However, this produces a 2d  output tensor in column major format, and there is no way to reshape and restride the tensor to be 3d again AND preserve the column major format, such that numerics are preserved. Thus, we have to transform the output to column major afterwards, requiring a .contiguous() call.

## Test plan
- Added tests that verify numerical accuracy

## Kernel microbenchmarks

Perf is decent for large E and abysmal for small E. Need to investigate this.

Update (9/15): NCU shows 3d kernel operating on (2,8192,5120) tensor is actually compute bound (??)

```
input_shape         to_mx_us    cuda_2d_us    cuda_3d_us    to_mx_gbps    cuda_2d_gbps    cuda_3d_gbps
----------------  ----------  ------------  ------------  ------------  --------------  --------------
(1, 8192, 5120)      117.92         69.776       242.112      1078.19          1822.11         525.128
(2, 8192, 5120)      431.264       105.536       249.728       589.615         2409.41        1018.23
(4, 8192, 5120)      848.992       195.584       297.376       599.015         2600.21        1710.16
(8, 8192, 5120)     1682.59        379.904       412.992       604.495         2677.3         2462.8
(16, 8192, 5120)    3350.53        775.984       615.536       607.139         2621.49        3304.82
(64, 8192, 5120)   13352          3150.66       1959.71        609.418         2582.62        4152.11
```

### DIFF
--- a/torchao/csrc/cuda/mx_kernels/mxfp8_quantize.cuh
+++ b/torchao/csrc/cuda/mx_kernels/mxfp8_quantize.cuh
@@ -22,6 +22,7 @@
 #include <cuda/barrier>
 #include <cuda/ptx>
 
+
 #define MIN_CUDA_SM 1000 // SM90 = 900, SM100 = 1000
 
 // Check if we're compiling for supported architecture
@@ -697,7 +698,7 @@ __global__ void __launch_bounds__(MXFP8_THREADS_PER_CHUNK)
           for (int y = 0; y < MXFP8_SHMEM_DIM_Y; y++) {
             for (int x = 0; x < MXFP8_SHMEM_DIM_X; x++) {
               printf("in_sh[%d][%d][%d] = %f\n", b, y, x,
-                     (float)in_sh[b][y][x]);
+                     DataTypeTraits<IType>::to_float(in_sh[b][y][x]));
             }
           }
         }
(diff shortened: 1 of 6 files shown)


danielvegamyhre:
@slayton58 @ngimel i would be curious to get your thoughts on ways to improve this kernel for quantizing 3d expert weights (E,N,K) along the N dim, where weights are contiguous. It uses nearly identical logic to the 2d dim1 cast kernel (which achieves ~85% mem bw utilization), yet the perf is much worse (~8% to 62% peak mem bw, depending on input size - see benchmarks in PR description).

I think the culprit might be how i'm allocating all the TMA descriptors and passing them in, the overhead might be too much for small E? NCU has not flagged anything particularly helpful so far. Strangely, for E=2 it shows the kernel is compute bound with 78% compute throughput % and 38% memory bandwidth %.

Additional context: torch.compile and handwritten triton kernels were both slow for mxfp8 quant for RHS operands where we scale colwise (32x1 granularity) e.g., (triton hit 56% peak mem bw). So I added a CUDA kernel here which I derived from a TE kernel which achieves ~85% peak mem bw (https://github.com/pytorch/ao/pull/2513). Basically we stripped out internal TE types, added support for different scale calculation modes (floor, rceil) to align with torchao numerics, then resolved some perf issues resulting from those changes to get reasonable perf.

Now, I'm finding quantizing 3d expert weights along dim1 is scaling extremely poorly as number of experts increases  (see this PR's description for details, and see https://github.com/pytorch/ao/pull/2999 for benchmarks). So I added a similar CUDA kernel to our mxfp8_cuda extension specifically for quantizing 3d expert weights colwise and writing directly to col major format we need it in. 

The first approach I tried was just updating the 2d kernel to handle 3d tensors by treating it as a 2d tensor of shape (E*N, K) but the coordinate mapping / pointer arithemetic became a complicated mess that wasn't working. So I made a new kernel, that is similar to the 2d kernel but passes in separate input/output TMA descriptors for each expert, then the kernel operates on each 2d expert with logical separation, in parallel.

ngimel:
Discussed offline, likely culprit is cudaMallocManaged calls on the hot path https://github.com/pytorch/ao/pull/3002/files#diff-7ddc6623d9efea4ee4f4bdb3cdd7ef16ec3d3bc8bc974be85311125e464efb3dR1326, we should be able to create just a single descriptor and do the loads using the correct offsets.

danielvegamyhre:
> Discussed offline, likely culprit is cudaMallocManaged calls on the hot path https://github.com/pytorch/ao/pull/3002/files#diff-7ddc6623d9efea4ee4f4bdb3cdd7ef16ec3d3bc8bc974be85311125e464efb3dR1326, we should be able to create just a single descriptor and do the loads using the correct offsets.

I've been trying this today, progress so far is:
- Scales for all experts are correct 
- Quantized data is correct for the expert=0 subtensor, but is all 0s for all other experts

This indicates the new input TMA descriptor and async loads are working properly, but the output TMA descriptor and/or async stores are not.

I think this is because the input tensor is in simple row major format, which can easily be represented in a TMA descriptor with shape `(E*N, K)` with stride K. However, the output data needs to be in "column major PER expert" format, so strides (N*K, 1, N). The 2d output TMA descriptor + `ptx::cp_async_bulk_tensor_2d_global_to_shared` do not seem capable of representing this layout so far (or could be a skill issue on my part, haha)

I will push a WIP PR on top of this stack to show the difference between this multiple-tma-descriptor approach, which is at least functionally correct, versus the single tma descriptor approach.

slayton58:
@danielvegamyhre I think there's a couple of options, can we do something like: 
```
# out : [E, N, K], stride: [N*K, 1, N]#
out.transpose(2,1) # Now [E, K, N], stride: [N*K, N, 1], representable by row-major TMA descriptor (2d or otherwise)
# compute
out.transpose(1, 2) # Back to original form
```
or, we can try using 3d TMAs directly - `ptx::cp_async_bulk_tensor_3d_global_to_shared` is the relevant read invocation, and a 3d descriptor would have to be created for this.
or (finally) we can ignore TMA writes, and use regular global stores (so STG) - we shouldn't need the async, pipeline-able part of the TMA instruction.

danielvegamyhre:
@slayton58 row major -> transpose to per expert col major is a great idea! Trying it now. 

>or, we can try using 3d TMAs directly

Yeah I've considered this, it is probably the "proper" way to do it but would require a larger refactor, hopefully the transpose method works.

drisspg:
can we just merge the top of stack w/ this pR

torchao/csrc/cuda/mx_kernels/mxfp8_quantize.cuh
```diff
@@ -900,10 +901,244 @@ __global__ void __launch_bounds__(MXFP8_THREADS_PER_CHUNK)
   // #endif
 }
 
+// 3D MXFP8 quantization kernel using 2D TMA
+template <typename IType, typename OType, size_t SCALE_DIM_Y,
+          size_t SCALE_DIM_X, ScaleCalculationMode ScalingMode>
+__global__ void __launch_bounds__(MXFP8_THREADS_PER_CHUNK)
+    mxfp8_quantize_kernel_3d(
+        const CUtensorMap* tensor_maps_input,
```
drisspg, line 909: Do we need to link against the cuda drivers in order to build this?

I think we fixed this at the top of the stack?

drisspg (approved):

torchao/csrc/cuda/mx_kernels/mxfp8_quantize.cuh
danielvegamyhre: do you mean in the create tensor map functions? the other examples I saw do reference the cuda driver to get the cuTensorMapEncodeTiled function pointer, so I did the same for 3d tensor map creation.

PR #3009, opened 2025-09-16
Optimize Load Time 

petrex:
TLDR:  This pull request introduces lazy import mechanics to the `torchao` package, significantly reducing its cold import time.

The changes implement module-level `__getattr__` and `__dir__` to expose submodules and top-level attributes only when accessed, rather than importing all heavy dependencies up front. Comprehensive tests and a benchmarking script are also added to validate and measure the improvements.

**Core: Lazy import implementation in `torchao`:**
- Refactored `torchao/__init__.py` to use module-level `__getattr__` and `__dir__`, enabling lazy loading of submodules and re-exported attributes such as `autoquant` and `quantize_`, which now load only when accessed. This avoids the eager import of heavy dependencies, reducing import-time overhead.
- Updated `__all__` to reflect the new lazy import structure, ensuring correct symbol exposure for users and tools.

**Testing:**
- Added `test/core/test_lazy_import.py` to verify that top-level import of `torchao` is lightweight and that submodules and attributes are loaded only upon access, ensuring correctness of the new lazy import mechanism.

**Benchmarking:**
- Introduced `scripts/benchmark_import_torchao.py`, a script to measure and compare cold import times for both eager and lazy import modes, providing quantitative evidence of the performance improvement.

**Internal improvements:**
- Removed unnecessary eager imports and clarified comments in `torchao/__init__.py`, especially around experimental and heavy submodules.
- Added necessary imports and setup for lazy import logic in `torchao/__init__.py`.


petrex:
duplicate #3024

PR #3011, opened 2025-09-16
Prevent union-finding cycles for shared qspecs

martinlsm:
Certain graphs caused infinite recursion when unioning nodes to groups based on shared quantization specs while implicit sharing was enabled. The problem occurred when `NodeOrEdge` A with its own `QuantizationSpec` received an edge (in `shared_with_map`) to `EdgeOrNode` B which in turn had a `SharedQuantizationSpec` pointing back to A. While in this state, looking up the quantization spec in `_unwrap_shared_qspec` resulted in an endless recursion loop and the program crashed.

Remedy this problem by, prior to unioning two trees, checking if the root of the parent has a `SharedQuantizationSpec` pointing to the root of the child; if that is the case, reverse the edge by letting parent point to the child. This will ensure that no cycle like the one described above is formed.

Add a test case which this commit fixes; namely, a graph where one input edge is shared between two ops. This graph would cause trigger cyclic reference bug that was seen prior to this patch.

### DIFF
--- a/torchao/quantization/pt2e/prepare.py
+++ b/torchao/quantization/pt2e/prepare.py
@@ -154,17 +154,17 @@ def _union(
     root_child = _find_root_edge_or_node(child, shared_with_map)
 
     parent_qspec = edge_or_node_to_qspec[root_parent]
-    if not (
+    if (
         isinstance(parent_qspec, SharedQuantizationSpec)
         and parent_qspec.edge_or_node == root_child
     ):
-        # union the two trees by pointing the root of child to root of parent
-        shared_with_map[root_child] = root_parent
-    else:
         # Parent already references child with a shared qspec. We would create
         # a cycle if we formed an edge from the child to the parent. Therefore,
         # we reverse the edge in this particular case.
         shared_with_map[root_parent] = root_child
+    else:
+        # union the two trees by pointing the root of child to root of parent
+        shared_with_map[root_child] = root_parent
 
 
 def _update_shared_with(
--- a/torchao/quantization/pt2e/prepare.py
+++ b/torchao/quantization/pt2e/prepare.py
@@ -153,15 +153,18 @@ def _union_if_no_cycle(
     root_parent = _find_root_edge_or_node(parent, shared_with_map)
     root_child = _find_root_edge_or_node(child, shared_with_map)
 
-    # Union the two trees by pointing the root of child to root of parent
-    # unless the parent qspec points to the root of child; that would create an
-    # undesired cycle and the nodes are in practice already unioned.
     parent_qspec = edge_or_node_to_qspec[root_parent]
     if not (
         isinstance(parent_qspec, SharedQuantizationSpec)
         and parent_qspec.edge_or_node == root_child
     ):
+        # union the two trees by pointing the root of child to root of parent
         shared_with_map[root_child] = root_parent
+    else:
+        # Parent already references child with a shared qspec (would create a
+        # cycle). The sets are effectively already unified via the shared
+        # qspec, so we leave shared_with_map unchanged.
+        pass
 
 
 def _update_shared_with(
--- a/torchao/quantization/pt2e/prepare.py
+++ b/torchao/quantization/pt2e/prepare.py
@@ -143,7 +143,7 @@ def _find_root_edge_or_node(
     return root
 
 
-def _union_if_no_cycle(
+def _union(
     parent: EdgeOrNode,
     child: EdgeOrNode,
     shared_with_map: dict[EdgeOrNode, EdgeOrNode],
@@ -161,10 +161,10 @@ def _union_if_no_cycle(
         # union the two trees by pointing the root of child to root of parent
         shared_with_map[root_child] = root_parent
     else:
-        # Parent already references child with a shared qspec (would create a
-        # cycle). The sets are effectively already unified via the shared
-        # qspec, so we leave shared_with_map unchanged.
-        pass
+        # Parent already references child with a shared qspec. We would create
+        # a cycle if we formed an edge from the child to the parent. Therefore,
+        # we reverse the edge in this particular case.
+        shared_with_map[root_parent] = root_child
 
 
 def _update_shared_with(
@@ -181,7 +181,7 @@ def _update_shared_with(
         parent = qspec.edge_or_node
         # we point from edge_or_node to the node that it is sharing_with, e.g.
         # qspec for a = SharedQuantizationSpec(b) means `a` points to `b`
-        _union_if_no_cycle(parent, child, shared_with_map, edge_or_node_to_qspec)
+        _union(parent, child, shared_with_map, edge_or_node_to_qspec)
 
 
 def _unwrap_shared_qspec(
@@ -263,9 +263,7 @@ def _union_input_edge_with(
         # since dtype is the same (we may want to extend this to be a more strict check
         # in the future)
         # so we point from `input_edge` to `arg` (output of the argument)
-        _union_if_no_cycle(
-            edge_or_node, input_edge, shared_with_map, edge_or_node_to_qspec
-        )
+        _union(edge_or_node, input_edge, shared_with_map, edge_or_node_to_qspec)
 
 
 def _get_edge_or_node_to_group_id(
--- a/torchao/quantization/pt2e/prepare.py
+++ b/torchao/quantization/pt2e/prepare.py
@@ -1,5 +1,6 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates.
 # All rights reserved.
+# Copyright 2025 Arm Limited and/or its affiliates.
 #
 # This source code is licensed under the BSD 3-Clause license found in the
 # LICENSE file in the root directory of this source tree.
@@ -142,22 +143,32 @@ def _find_root_edge_or_node(
     return root
 
 
-def _union(
+def _union_if_no_cycle(
     parent: EdgeOrNode,
     child: EdgeOrNode,
     shared_with_map: dict[EdgeOrNode, EdgeOrNode],
+    edge_or_node_to_qspec: dict[EdgeOrNode, QuantizationSpecBase],
 ) -> None:
     """Merge the subtree for `child` with `parent`, the order is important here"""
     root_parent = _find_root_edge_or_node(parent, shared_with_map)
     root_child = _find_root_edge_or_node(child, shared_with_map)
-    # union the two trees by pointing the root of child to root of parent
-    shared_with_map[root_child] = root_parent
+
+    # Union the two trees by pointing the root of child to root of parent
+    # unless the parent qspec points to the root of child; that would create an
+    # undesired cycle and the nodes are in practice already unioned.
+    parent_qspec = edge_or_node_to_qspec[root_parent]
+    if not (
+        isinstance(parent_qspec, SharedQuantizationSpec)
+        and parent_qspec.edge_or_node == root_child
+    ):
+        shared_with_map[root_child] = root_parent
 
 
 def _update_shared_with(
     child: EdgeOrNode,
     qspec: QuantizationSpecBase,
     shared_with_map: dict[EdgeOrNode, EdgeOrNode],
+    edge_or_node_to_qspec: dict[EdgeOrNode, QuantizationSpecBase],
 ):
     """Update the `shared_with_map` based on the qspec, this applies the `SharedQuantizationSpec`
     configuration and established the relationship between `edge_or_node` with the edge/node that it
@@ -167,7 +178,7 @@ def _update_shared_with(
         parent = qspec.edge_or_node
         # we point from edge_or_node to the node that it is sharing_with, e.g.
         # qspec for a = SharedQuantizationSpec(b) means `a` points to `b`
-        _union(parent, child, shared_with_map)
+        _union_if_no_cycle(parent, child, shared_with_map, edge_or_node_to_qspec)
 
 
 def _unwrap_shared_qspec(
@@ -249,7 +260,9 @@ def _union_input_edge_with(
         # since dtype is the same (we may want to extend this to be a more strict check
         # in the future)
         # so we point from `input_edge` to `arg` (output of the argument)
-        _union(edge_or_node, input_edge, shared_with_map)
+        _union_if_no_cycle(
+            edge_or_node, input_edge, shared_with_map, edge_or_node_to_qspec
+        )
 
 
 def _get_edge_or_node_to_group_id(
@@ -311,7 +324,9 @@ def _get_edge_or_node_to_group_id(
     for edge_or_node, qspec in edge_or_node_to_qspec.items():
         if isinstance(edge_or_node, torch.fx.Node):
             output_node = edge_or_node
-            _update_shared_with(output_node, qspec, shared_with_map)
+            _update_shared_with(
+                output_node, qspec, shared_with_map, edge_or_node_to_qspec
+            )
         else:
             input_edge = edge_or_node
             input_edge_root_qspec = _unwrap_shared_qspec(
@@ -332,9 +347,6 @@ def _get_edge_or_node_to_group_id(
                 # because we will point the root of (node1, node2) (in this case node1) to the root of (node1, node3)
                 # Step 3. and when we process (node1, node3), it can try to point to node1 as well, then we'll
                 # have a circular dependency
-                # the following order works around this issue, but this does not allow arbitrary configuration
-                # of sharing so it might break in a different case in the future, when it breaks
-                # quantizer writer can check the notes here to debug the issue
 
                 # sharing with other users of the producer node
                 # (arg, user)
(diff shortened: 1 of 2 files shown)


martinlsm:
This PR is the proposed solution to the bug I initially reported here in the ExecuTorch repo: pytorch/executorch#13842

jerryzh168:
Thanks, can you add a test in https://github.com/pytorch/ao/blob/main/test/quantization/pt2e/test_quantize_pt2e.py for us to understand the issue better?

also for the test, add a comment that follows this format: https://github.com/pytorch/ao/blob/4dffb40280ea7b0e1732c580d08df58d0134c543/torchao/quantization/pt2e/prepare.py#L278 will be better for people to follow as well

torchao/quantization/pt2e/prepare.py
```diff
@@ -142,22 +143,32 @@ def _find_root_edge_or_node(
...
+
+    # Union the two trees by pointing the root of child to root of parent
+    # unless the parent qspec points to the root of child; that would create an
+    # undesired cycle and the nodes are in practice already unioned.
+    parent_qspec = edge_or_node_to_qspec[root_parent]
+    if not (
+        isinstance(parent_qspec, SharedQuantizationSpec)
+        and parent_qspec.edge_or_node == root_child
+    ):
+        shared_with_map[root_child] = root_parent
```
jerryzh168: could you add a comment to explain what happens (the state of things) in the else branch as well?

kimishpatel:
> The problem occurred when NodeOrEdge A with its own QuantizationSpec received an edge (in shared_with_map) to EdgeOrNode B

I dont follow why there A -> B edge if B's quant spec points to A

torchao/quantization/pt2e/prepare.py
martinlsm: I've attempted to clarify it. Have a look at it now.

martinlsm:
> Thanks, can you add a test in https://github.com/pytorch/ao/blob/main/test/quantization/pt2e/test_quantize_pt2e.py for us to understand the issue better?
> 
> also for the test, add a comment that follows this format:
> 
> https://github.com/pytorch/ao/blob/4dffb40280ea7b0e1732c580d08df58d0134c543/torchao/quantization/pt2e/prepare.py#L278
> 
> will be better for people to follow as well

I've added a test now. Have a look and see what you think :)

martinlsm:
I managed to find a simpler model to reproduce the problem. Let me get back again with an update.

martinlsm:
> > The problem occurred when NodeOrEdge A with its own QuantizationSpec received an edge (in shared_with_map) to EdgeOrNode B
> 
> I dont follow why there A -> B edge if B's quant spec points to A

It's quite complicated, but I will attempt to explain. We have the following model:
```
class M(torch.nn.Module):
    def forward(self, x):
        a = x.clone()
        b = torch.eq(a, x)
        return b
```
And the following sitatuation of qspecs:
- _clone_
  - Input `(x,clone)` has its own `QuantizationSpec`
  - Output `clone` shares qspec with input: `SharedQuantizationSpec((x,clone))`
- _eq_
  - First input `(clone,eq)` has its own `QuantizationSpec`
  - Second input `(x,eq)` shares spec with the first input: `SharedQuantizationSpec((clone,eq))`
  - Output `eq` is not quantized (bool output)

Then algorithm `_get_edge_or_node_to_group_id` produces extra edges in `shared_with_map` when implicit sharing is enabled. These edges point downstream in the graph.

This is what happens when the problem occurs:
1. The algorithm processes the first edge `(x,clone)`: implicit sharing will produce edge `shared_with_map[(x,clone)] = (x,eq)` because eq is another user of x with an identical qspec.
2. The algorithm processes the second edge `(clone,eq)`: implicit sharing will cause `(clone,eq)` to be shared with `clone` (sharing with producer node), but `clone` has a `SharedQuantizationSpec((x,clone))`, and in step 1, we made `(x,clone)` share with (point to) `(x,eq)`. So `(x,eq)` is the node we end up at and we produce the edge `shared_with_map[(clone,eq)] = (x,eq)`
3. Now things has gone bad because when the algorithm processes the last edge `(x,eq)` and tries to find its qspec, it gets stuck in a loop because it looks up `(clone,eq)` via its shared qspec, but ends back again at `(x,eq)` because of the edge we produced in step 2. This is the endless loop.

Maybe there's a smarter way to solve this problem but I could only come up with the solution to skip producing the edge at step 2. If we skip forming an edge in step 2, the nodes/edges will still be unioned in the end because when the last edge  `(x,eq)` is processed, we will union it with `(clone,eq)` by forming the edge in the correct direction (`shared_with_map[(x,eq)] = (clone,eq)`, and not the other way around as we did in step 2.).

jerryzh168:
> 3\. Now things has gone bad because when the algorithm processes the last edge `(x,eq)` and tries to find its qspec, it gets stuck in a loop because it looks up `(clone,eq)` via its shared qspec, but ends back again at `(x,eq)` because of the edge we produced in step 2. This is the endless loop.

thanks for the detailed example! this is very helpful, could you add this explanation to the test as well?

looking at the example, seems like this is still a relatively simple case, what if it takes a few more steps for (clone, eq) to point back to (x, eq)? 

It seems to me, at the high level, we need some ordering of the output nodes and input edges, so that instead of have the pointer always point from whatever edges we are processing to another node or edge, as we are doing in the implicit sharing code, we always follow the ordering instead. for example in this case, we have:
```
name                        ordering
x                                0
(x, clone)                 1
clone                        2
(clone, eq)               3
(x, eq)                      4
eq                             5
```

the problem right now is (clone, eq) points to (x, eq), and then later (x, eq) points to (clone ,eq) again, forming a loop, due to lack of ordering. let's say, we follow an order that's listed above (when sharing, only higher ordering entities will point to lower ordering entities, not the reverse), even if we need to union (clone, eq) (3) and (x, eq) (4) multiple times, we always union the higher with the lower, and we'll always have (x, eq) points to (clone, eq) for both step 2 and 3 since the ordering of (x, eq) is higher than (clone, eq).

We could add a new ordering map probably to assign the ordering for each entities (output node and input edge) in the graph I think.

kimishpatel:
> the problem right now is (clone, eq) points to (x, eq), and then later (x, eq) points to (clone ,eq) again, forming a loop, due to lack of ordering.

I dont think thats the issue. Loop is really of (x, eq) -> (clone, eq) -> clone node -> (x, clone) -> (x, eq) loop.

kimishpatel:
> * The algorithm processes the first edge `(x,clone)`: implicit sharing will produce edge `shared_with_map[(x,clone)] = (x,eq)` because eq is another user of x with an identical qspec.

Given (x, clone) has its own quant spec why do we end up with shared_with_map[(x, clone)] = (x, eq)? If anything I would have expected, after processing (x, clone),  shared_with_map to be empty OR with implicit sharing shared_with_map[(x, eq)] = (x, clone). Basically which ever edge is annotated to have its own qspec should never be a key in the `shared_with_map`

kimishpatel:
> > * The algorithm processes the first edge `(x,clone)`: implicit sharing will produce edge `shared_with_map[(x,clone)] = (x,eq)` because eq is another user of x with an identical qspec.
> 
> Given (x, clone) has its own quant spec why do we end up with shared_with_map[(x, clone)] = (x, eq)? If anything I would have expected, after processing (x, clone), shared_with_map to be empty OR with implicit sharing shared_with_map[(x, eq)] = (x, clone). Basically which ever edge is annotated to have its own qspec should never be a key in the `shared_with_map`

Now in this `First input (clone,eq) has its own QuantizationSpec` becomes problematic because implicit sharing would force it to map to clone node. So maybe some ordering of which edge is the one owning qspec should be done. And maybe if you do explicit cycle finding then you would denote (x, eq) as the leader

jerryzh168:
> > * The algorithm processes the first edge `(x,clone)`: implicit sharing will produce edge `shared_with_map[(x,clone)] = (x,eq)` because eq is another user of x with an identical qspec.
> 
> Given (x, clone) has its own quant spec why do we end up with shared_with_map[(x, clone)] = (x, eq)? If anything I would have expected, after processing (x, clone), shared_with_map to be empty OR with implicit sharing shared_with_map[(x, eq)] = (x, clone). Basically which ever edge is annotated to have its own qspec should never be a key in the `shared_with_map`

even something annotated with its own spec, it could share with other nodes / edges when they are the "same", e.g.
op1 -> op2
     \ --> op3

output node of op1, and input edge (op1, op2), input edge (op1, op3) are the same thing since we just need to insert one observer, when implicit sharing is enabled.

jerryzh168:
> > the problem right now is (clone, eq) points to (x, eq), and then later (x, eq) points to (clone ,eq) again, forming a loop, due to lack of ordering.
> 
> I dont think thats the issue. Loop is really of (x, eq) -> (clone, eq) -> clone node -> (x, clone) -> (x, eq) loop.

I think (clone, eq) points to (x, eq) IIUC, but cc @martinlsm to confirm, it might be helpful to write down `edge_or_node_to_qspec` and the exact content of `shared_with_map` after each step

jerryzh168:
my understanding is the following:

Screenshot 2025-09-18 at 17 01 21

kimishpatel:
btw can you not detect cycles explicitly? I think the issue is that we need to have one leader node that takes the responsibility of owning the qspec. In situations like it like "i dont have qspec, go ask this other guy" and then we have guy x, asking guy y, askin guy z, asking guy x. So it seems to me that for implicit sharing we have to denote one node in the cycle as leader.

martinlsm:
> > the problem right now is (clone, eq) points to (x, eq), and then later (x, eq) points to (clone ,eq) again, forming a loop, due to lack of ordering.
> 
> I dont think thats the issue. Loop is really of (x, eq) -> (clone, eq) -> clone node -> (x, clone) -> (x, eq) loop.

I agree with @kimishpatel . Even if we decide some order by sorting, there could be shared qspecs pointing in the reverse order and thus we could risk forming a loop if I understand things correctly.

martinlsm:
> my understanding is the following:
> 
> Screenshot 2025-09-18 at 17 01 21

This is close to correct, except that in step 2., when processing (clone,eq), it is not shared with (x,eq) as _another user_. Another user in this context would be _another user of_ clone, which there are none, i.e., eq is the only user of clone and that is what we are currently processing.

(clone,eq) instead ends up pointing to (x,eq) by sharing with previous output clone. clone is in the current state pointing to (x,eq) because:
- clone has a shared qspec to (x,clone)
- (x,clone) points to (x,eq) in `shared_with_map` because we formed that edge in the previous step when we processed (x,clone).

So when (clone,eq) wants to share with clone, it arrives at (x,eq).

martinlsm:
> btw can you not detect cycles explicitly? I think the issue is that we need to have one leader node that takes the responsibility of owning the qspec. In situations like it like "i dont have qspec, go ask this other guy" and then we have guy x, asking guy y, askin guy z, asking guy x. So it seems to me that for implicit sharing we have to denote one node in the cycle as leader.

I just submitted a new version now that is kind of what you suggest here I guess. Instead of skipping forming the union, I instead reverse the edge in the `_union`. This feels a bit more robust since we still ensure that the nodes are unioned and it's less risk that we'll miss some edge case I guess...

> looking at the example, seems like this is still a relatively simple case, what if it takes a few more steps for (clone, eq) to point back to (x, eq)?

Then regarding this good argument, I don't believe it is possible to be more steps than _one_ because `_union` calls `_find_root_edge_or_node` on both the parent's and the child's subtree, and that function performs path compression so any indirections will be eliminated. Please correct me if I'm wrong here though because this stuff is complicated!

martinlsm:
> thanks for the detailed example! this is very helpful, could you add this explanation to the test as well?

Oh, right I forgot about this one. Will sort that out on Monday. But in the meantime, let me know what you think about the new solution if you got the time.

jerryzh168:
> > > the problem right now is (clone, eq) points to (x, eq), and then later (x, eq) points to (clone ,eq) again, forming a loop, due to lack of ordering.
> > 
> > 
> > I dont think thats the issue. Loop is really of (x, eq) -> (clone, eq) -> clone node -> (x, clone) -> (x, eq) loop.
> 
> I agree with @kimishpatel . Even if we decide some order by sorting, there could be shared qspecs pointing in the reverse order and thus we could risk forming a loop if I understand things correctly.

I think it similar, if we have ordering, then even if user specifies a sharing with a reverse direction, we'll have the node/edge with a lower ordering_id to hold the final qspec, so we will have to modify the `edge_or_node_to_qspec` dictionary as well

personally I feel the ordering map will help make it easier since no need to worry about cycle detection anymore, but both works I feel

torchao/quantization/pt2e/prepare.py
```diff
@@ -146,18 +147,31 @@ def _union(
...
+    if not (
+        isinstance(parent_qspec, SharedQuantizationSpec)
+        and parent_qspec.edge_or_node == root_child
+    ):
+        # union the two trees by pointing the root of child to root of parent
+        shared_with_map[root_child] = root_parent
+    else:
+        # Parent already references child with a shared qspec. We would create
+        # a cycle if we formed an edge from the child to the parent. Therefore,
+        # we reverse the edge in this particular case.
```
jerryzh168, line 163: will this be more confusing than just assign an ordering before hand?

jerryzh168: also I haven't thought through, but wondering if it's possible that root_child can go around and end up pointing to root_parent again

kimishpatel:
> > > > the problem right now is (clone, eq) points to (x, eq), and then later (x, eq) points to (clone ,eq) again, forming a loop, due to lack of ordering.
> > > 
> > > 
> > > I dont think thats the issue. Loop is really of (x, eq) -> (clone, eq) -> clone node -> (x, clone) -> (x, eq) loop.
> > 
> > 
> > I agree with @kimishpatel . Even if we decide some order by sorting, there could be shared qspecs pointing in the reverse order and thus we could risk forming a loop if I understand things correctly.
> 
> I think it similar, if we have ordering, then even if user specifies a sharing with a reverse direction, we'll have the node/edge with a lower ordering_id to hold the final qspec, so we will have to modify the `edge_or_node_to_qspec` dictionary as well
> 
> personally I feel the ordering map will help make it easier since no need to worry about cycle detection anymore, but both works I feel

I think cycle detection is probably more easy to reason about. it is clear in its intent. Unless there is a significant concern on other aspect, I would suggest we do cycle detection

torchao/quantization/pt2e/prepare.py
jerryzh168: cc @kimishpatel is this the cycle detection you have in mind?

seems OK to me, if this is the only thing that's needed

kimishpatel: I am not sure. Isnt the cycle already formed (parent_qspec.edge_or_node == root_child) before we come here. it feels we are detecting that and correcting it. I might be wrong though. I

martinlsm: @kimishpatel Prior to forming the problematic union, we are in the state that is shown in figure below about to assign `shared_with_map[(clone,eq)] = (x,eq)`. So there's not already a union we are correct. We are just reversing the green edge by assigning `shared_with_map[(x,eq)] = (clone,eq)` to make the edge point in the same direction as the blue one (`edge_or_node_to_qspec`).

prepare_state (2)

martinlsm:
> > thanks for the detailed example! this is very helpful, could you add this explanation to the test as well?
> 
> Oh, right I forgot about this one. Will sort that out on Monday. But in the meantime, let me know what you think about the new solution if you got the time.

I have done an update regarding this now.

torchao/quantization/pt2e/prepare.py
```diff
@@ -146,18 +147,31 @@ def _union(
...
+    edge_or_node_to_qspec: dict[EdgeOrNode, QuantizationSpecBase],
 ) -> None:
     """Merge the subtree for `child` with `parent`, the order is important here"""
     root_parent = _find_root_edge_or_node(parent, shared_with_map)
     root_child = _find_root_edge_or_node(child, shared_with_map)
-    # union the two trees by pointing the root of child to root of parent
-    shared_with_map[root_child] = root_parent
+
+    parent_qspec = edge_or_node_to_qspec[root_parent]
+    if not (
```
jerryzh168: nit: reversing the condition and swap if / else branches might be clearer

jerryzh168 (approved):
LGTM, seems like it's the only change that's needed to prevent a loop?

kimishpatel (approved):
ok looks good

torchao/quantization/pt2e/prepare.py
martinlsm: I agree with you! The branches are swapped now.

martinlsm:
> LGTM, seems like it's the only change that's needed to prevent a loop?

I think it's the only thing needed yes! Proving it would be hard but I have not yet found a case where it does not work. But if this problem is ever to be seen again, it would just have to revisited I guess.

So we are good to go and can merge this? Thanks guys for the useful feedback and discussion in this PR!

PR #3015, opened 2025-09-16
Fix torchao_convert, remove StretchedAffineQuantizedTensor

lisjin:
Follow-up to `QuantOptimizer.torchao_convert` added in #2947:
1. Use the "_default" key for the dict passed to ModuleFqnToConfig instead of listing all linear weight names.
2. Fix unwanted quantization of tied weights when using "_default". Some HF models have a `_tied_weights_keys` variable (e.g., in [Qwen3Model](https://github.com/huggingface/transformers/blob/91393fe4cc3266a05bc0d129e34ff5f761bb46e2/src/transformers/models/qwen3/modeling_qwen3.py#L430)).
    i. Suppose we tie `lm_head.weight` with `model.embed_tokens.weight`. Even though they're tied, their names both show up in `model.named_parameters()`. This is a problem since HF's quantization_config will try to apply the "_default" config to the tied LM head, which may differ from the embedding's. @metascroy This is a major issue for our mixed quantization of embeddings and LM head.
3. @jerryzh168 Remove StretchedAffineQuantizedTensor to align with deprecated version=1 in torchao configs.

### DIFF
--- a/test/prototype/test_parq.py
+++ b/test/prototype/test_parq.py
@@ -54,9 +54,16 @@ def split_param_groups(model) -> tuple[list, list, list]:
     params_quant, params_embed, params_no_quant = [], [], []
 
     def get_param_groups(model):
+        seen_data_ptrs = set()  # avoid duplicates in case of tied weights
         for module in model.children():
             is_linear = _is_linear(module)
             for n, p in module.named_parameters():
+                if n == "weight":
+                    data_ptr = p.data_ptr()
+                    if data_ptr in seen_data_ptrs:
+                        continue
+                    seen_data_ptrs.add(data_ptr)
+
                 if is_linear and n == "weight":
                     params_quant.append(p)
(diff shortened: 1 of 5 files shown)


metascroy:
> This is a problem since HF's quantization_config will try to apply the "_default" config to the tied LM head, which may differ from the embedding's

Can you say more here with an example code snippet?

lisjin:
> > This is a problem since HF's quantization_config will try to apply the "_default" config to the tied LM head, which may differ from the embedding's
> 
> Can you say more here with an example code snippet?

@metascroy Sure, [here](https://github.com/huggingface/transformers/blob/cd74917ffc3e8f84e4a886052c5ab32b7ac623cc/src/transformers/quantizers/quantizer_torchao.py#L274–L277) is where HF applies quantize_ given a "_default" key. Their filter_fn doesn't doesn't check for tied parameters, so it would apply the default config to LM head tied to the embeddings.

PR #3018, opened 2025-09-17
fix: avoid removing from tuple in _get_to_kwargs

orangeH25:
This PR fixes a bug in `_get_to_kwargs` where `args` is a tuple but the code attempted to call `args.remove(arg)`, leading to:
```bash
    def _get_to_kwargs(self, *args, **kwargs):
        # `torch._C._nn._parse_to` can't handle `layout` argument
        for arg in args:
            if isinstance(arg, torch.layout):
>               args.remove(arg)
                ^^^^^^^^^^^
E               AttributeError: 'tuple' object has no attribute 'remove'

torchao/utils.py:722: AttributeError
```

Temporary test case in test/test_utils.py::TestTorchAOBaseTensor(unittest.TestCase):
```python
    def test__get_to_kwargs_with_layout(self):
        MyClass = TorchAOBaseTensor()
        import torch
        MyClass._get_to_kwargs(torch.strided,device="cuda")
```

Temporary test command:
```bash
pytest --count=1 test/test_utils.py::TestTorchAOBaseTensor::test__get_to_kwargs_with_layout
```

### DIFF
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -340,6 +340,9 @@ class TestTorchAOBaseTensor(unittest.TestCase):
         )
         self._test_default_impls_helper(lp_tensor, lp_tensor_for_copy)
 
+    def test__get_to_kwargs_with_layout(self):
+        MyClass = TorchAOBaseTensor()
+        MyClass._get_to_kwargs(torch.strided,device="cuda")
 
 if __name__ == "__main__":
     unittest.main()
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -144,6 +144,10 @@ class TestTorchAOBaseTensor(unittest.TestCase):
         self.assertTrue(torch.equal(lp_tensor.qdata, reconstructed.qdata))
         self.assertEqual(lp_tensor.attr, reconstructed.attr)
 
+        # test _get_to_kwargs
+        _ = lp_tensor._get_to_kwargs(torch.strided, device="cuda")
+        _ = lp_tensor._get_to_kwargs(layout=torch.strided, device="cuda")
+
         # `to` / `_to_copy`
         original_device = lp_tensor.device
         lp_tensor = lp_tensor.to("cuda")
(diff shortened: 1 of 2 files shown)


orangeH25:
Hey @jerryzh168 @huydhn  , Please help to review this one,thank you.

jerryzh168:
@orangeH25 thanks! can you add the test to `test/test_utils.py`?

orangeH25:
> @orangeH25 thanks! can you add the test to `test/test_utils.py`?

Got it, I'll add the test

orangeH25:
Hey @jerryzh168 , the test case has been added. Please review it again. Thanks!

test/test_utils.py
```diff
@@ -340,6 +340,9 @@ def __init__(
         )
         self._test_default_impls_helper(lp_tensor, lp_tensor_for_copy)
 
+    def test__get_to_kwargs_with_layout(self):
+        MyClass = TorchAOBaseTensor()
```
jerryzh168: this is not the intended use for TorchAOBaseTensor, maybe you can add a test item in https://github.com/pytorch/ao/blob/18dbe875a0ce279739dda06fda656e76845acaac/test/test_utils.py#L108 since `_get_to_kwargs` is also a default impl

jerryzh168: we can probably expand the op test for `to`: https://github.com/pytorch/ao/blob/18dbe875a0ce279739dda06fda656e76845acaac/test/test_utils.py#L148-L152

and also add an explicit test for `_get_to_kwargs` as well

orangeH25: Thanks, I'll modify the test

jerryzh168 (approved):

test/test_utils.py
```diff
@@ -144,6 +144,10 @@ def _test_default_impls_helper(self, lp_tensor, lp_tensor_for_copy):
         self.assertTrue(torch.equal(lp_tensor.qdata, reconstructed.qdata))
         self.assertEqual(lp_tensor.attr, reconstructed.attr)
 
+        # test _get_to_kwargs
+        _ = lp_tensor._get_to_kwargs(torch.strided, device="cuda")
+        _ = lp_tensor._get_to_kwargs(layout=torch.strided, device="cuda")
```
jerryzh168, line 149: btw, testing the higher level call to `to` op directly might be better here, `lp_tensor.to(...)`

jerryzh168:
I'll merge for now and follow up to change the call to `to`

PR #3021, opened 2025-09-17
Add Sparsify overhead benchmark

namgyu-youn:
Summary:
This PR adds sparsify overhead benchmark, omitted in ICLR workshop paper: https://arxiv.org/abs/2503.16672
- fix: https://github.com/pytorch/ao/issues/2612

In the paper, there are two parts for the benchmark: 1) Sparsify operation overhead, 2) Sparse-GEMM kernel performance. Part 1) was omitted from the original benchmark, so this PR adds the missing sparsify-only benchmark comparing `torchao.sparse24_sm90_sparsify` against `torch._cslt_compress` (cuSPASRELt) baseline.

Test plan: CI

### DIFF
--- a/benchmarks/benchmark_e2e_fp8_sparse_linear.py
+++ b/benchmarks/benchmark_e2e_fp8_sparse_linear.py
@@ -41,18 +41,16 @@ def benchmark(num_tokens, hidden_size=8192, intermediate_size=8192):
     fp16_time = benchmark_microseconds(ffn_ref, input_tensor)
 
     # Sparsify-only benchmarks
-    X_scale = torch.empty([num_tokens, 1], device="cuda", dtype=torch.float32)
-    ao_cusparse_time = benchmark_microseconds(
-        lambda: torch.ops.torchao.sparse24_sm90_sparsify(
+    ao_fast_sparsification_time = benchmark_microseconds(
+        torch.ops.torchao.sparse24_sm90_sparsify(
             input_tensor,
             "cutlass",
             "srelu",
             "largest",
             dtype=torch.float8_e4m3fn,
-            scale=X_scale,
         )
     )
-    cusparse_time = benchmark_microseconds(lambda: torch._cslt_compress(input_tensor))
+    cusparse_time = benchmark_microseconds(torch._cslt_compress, input_tensor)
 
     # bf16
     ffn_clone = (
@@ -131,10 +129,10 @@ def benchmark(num_tokens, hidden_size=8192, intermediate_size=8192):
         "fp8_c_time (us)": fp8_c_time,
         "fp8_c_sparse_time (us)": fp8_c_sparse_time,
         "fp8_c_activation_sparse_time (us)": fp8_c_activation_sparse_time,
-        "ao_cusparse_time (us)": ao_cusparse_time,
-        "cusparse_compress_time (us)": cusparse_time,
+        "ao_fast_sparsification_time (us)": ao_fast_sparsification_time,
+        "cusparse*_compress_time (us)": cusparse_time,
         "speedup": fp8_c_time / fp8_c_activation_sparse_time,
-        "sparsify_speedup": cusparse_time / ao_cusparse_time,
+        "sparsify_speedup": cusparse_time / ao_fast_sparsification_time,
     }
 
 
--- a/benchmarks/benchmark_e2e_fp8_sparse_linear.py
+++ b/benchmarks/benchmark_e2e_fp8_sparse_linear.py
@@ -45,12 +45,12 @@ def benchmark(num_tokens, hidden_size=8192, intermediate_size=8192):
         torch.ops.torchao.sparse24_sm90_sparsify(
             input_tensor,
             "cutlass",
-            "srelu",
+            "identity",
             "largest",
             dtype=torch.float8_e4m3fn,
         )
     )
-    cusparse_time = benchmark_microseconds(torch._cslt_compress, input_tensor)
+    cusparselt_time = benchmark_microseconds(torch._cslt_compress, input_tensor)
 
     # bf16
     ffn_clone = (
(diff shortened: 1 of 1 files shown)


namgyu-youn:
@jcaip Please review this PR, thanks.

jcaip:
@namgyu-youn Can you share the results of your benchmark script?

benchmarks/benchmark_e2e_fp8_sparse_linear.py
```diff
@@ -40,6 +40,20 @@ def benchmark(num_tokens, hidden_size=8192, intermediate_size=8192):
     input_tensor = torch.randn(num_tokens, hidden_size).to(torch.bfloat16).cuda()
     fp16_time = benchmark_microseconds(ffn_ref, input_tensor)
 
+    # Sparsify-only benchmarks
+    X_scale = torch.empty([num_tokens, 1], device="cuda", dtype=torch.float32)
+    ao_cusparse_time = benchmark_microseconds(
+        lambda: torch.ops.torchao.sparse24_sm90_sparsify(
+            input_tensor,
+            "cutlass",
+            "srelu",
```
jcaip: this should be "identity" here instead

benchmarks/benchmark_e2e_fp8_sparse_linear.py
```diff
@@ -40,6 +40,20 @@ def benchmark(num_tokens, hidden_size=8192, intermediate_size=8192):
...
+            "largest",
+            dtype=torch.float8_e4m3fn,
+            scale=X_scale,
+        )
+    )
+    cusparse_time = benchmark_microseconds(lambda: torch._cslt_compress(input_tensor))
```
jcaip: do you need this lambda? Can you just pass in like we do in L41 above:

```
cusparse_time = benchmark_microseconds(torch._cslt_compress, input_tensor)
```

benchmarks/benchmark_e2e_fp8_sparse_linear.py
```diff
@@ -40,6 +40,20 @@ def benchmark(num_tokens, hidden_size=8192, intermediate_size=8192):
     input_tensor = torch.randn(num_tokens, hidden_size).to(torch.bfloat16).cuda()
     fp16_time = benchmark_microseconds(ffn_ref, input_tensor)
 
+    # Sparsify-only benchmarks
+    X_scale = torch.empty([num_tokens, 1], device="cuda", dtype=torch.float32)
+    ao_cusparse_time = benchmark_microseconds(
+        lambda: torch.ops.torchao.sparse24_sm90_sparsify(
```
jcaip: same nit as below

benchmarks/benchmark_e2e_fp8_sparse_linear.py
```diff
@@ -40,6 +40,20 @@ def benchmark(num_tokens, hidden_size=8192, intermediate_size=8192):
...
+            input_tensor,
+            "cutlass",
+            "srelu",
+            "largest",
+            dtype=torch.float8_e4m3fn,
+            scale=X_scale,
```
jcaip: I think you can pass in `None` to scale for the fairest comparison.

namgyu-youn:
> @namgyu-youn Can you share the results of your benchmark script? 

@jcaip unfortunately not available to H100 HBM, please feel free to edit for benchmarks result

benchmarks/benchmark_e2e_fp8_sparse_linear.py
```diff
@@ -117,7 +131,10 @@ def benchmark(num_tokens, hidden_size=8192, intermediate_size=8192):
         "fp8_c_time (us)": fp8_c_time,
         "fp8_c_sparse_time (us)": fp8_c_sparse_time,
         "fp8_c_activation_sparse_time (us)": fp8_c_activation_sparse_time,
+        "ao_cusparse_time (us)": ao_cusparse_time,
```
jcaip: nit: I think something like `ao_fast_sparsification_time` is a better var name.

benchmarks/benchmark_e2e_fp8_sparse_linear.py
```diff
@@ -117,7 +131,10 @@ def benchmark(num_tokens, hidden_size=8192, intermediate_size=8192):
...
+        "cusparse_compress_time (us)": cusparse_time,
```
jcaip: cusparselt* instead of cusparse so we don't get confused :)

benchmarks/benchmark_e2e_fp8_sparse_linear.py
jcaip: please update this :)

jcaip (changes requested):

benchmarks/benchmark_e2e_fp8_sparse_linear.py
```diff
@@ -117,7 +129,10 @@ def benchmark(num_tokens, hidden_size=8192, intermediate_size=8192):
         "fp8_c_time (us)": fp8_c_time,
         "fp8_c_sparse_time (us)": fp8_c_sparse_time,
         "fp8_c_activation_sparse_time (us)": fp8_c_activation_sparse_time,
+        "ao_fast_sparsification_time (us)": ao_fast_sparsification_time,
+        "cusparse*_compress_time (us)": cusparse_time,
```
jcaip: cusparse_time isnt a good name for this because there is a seperate cusparse library, aside from cusparselt. please use cusparselt here instead

Also looks like theres a typo in the string?

benchmarks/benchmark_e2e_fp8_sparse_linear.py
namgyu-youn: Thanks for the reminder; I missed it.

benchmarks/benchmark_e2e_fp8_sparse_linear.py
namgyu-youn: Thanks, I didn't know it due to my lack of background.

jcaip (approved):

PR #3022, opened 2025-09-17
Pass QAT learned qparams in convert

andrewor14:
**Summary:** Add support to pass scales and zero points learned during QAT range learning to the PTQ base config. Currently only the following configs support this feature:

```
IntxWeightOnlyConfig
Int8DynamicActivationInt4WeightConfig
Int8DynamicActivationIntxWeightConfig
```

During the convert phase, QAT will detect if range learning was used during training, and pass the learned scales and zero points as custom qparams to the quantized tensor subclass, so PTQ will produce more consistent numerics.

Fixes part of https://github.com/pytorch/ao/issues/2271.

**Test Plan:**
```
python test/quantization/test_qat.py -k test_range_learning_convert_pass_qparams
```

### DIFF
--- a/test/quantization/test_qat.py
+++ b/test/quantization/test_qat.py
@@ -10,7 +10,7 @@
 import copy
 import unittest
 import warnings
-from typing import List
+from typing import List, Type
 
 import torch
 import torch.nn.functional as F
@@ -2303,6 +2303,57 @@ class TestQAT(TestCase):
             atol=0,
             rtol=0,
         )
+
+    @parametrize(
+        "base_config_cls",
+        [
+            IntxWeightOnlyConfig,
+            Int8DynamicActivationInt4WeightConfig,
+            Int8DynamicActivationIntxWeightConfig,
+        ],
+    )
+    def test_range_learning_convert_pass_qparams(
+        self, base_config_cls: Type[AOBaseConfig]
+    ):
+        """
+        Verify that range learning QAT can pass qparams from the prepared
+        model to the convert model.
+        """
+        group_size = 32
+        config = IntxFakeQuantizeConfig(
+            torch.int4,
+            group_size=group_size,
+            is_symmetric=True,
+            is_dynamic=False,
+            range_learning=True,
+        )
+        m = M()
+        example_inputs = m.example_inputs()
+        quantize_(m, QATConfig(weight_config=config, step="prepare"))
+        initialize_fake_quantizers(m, example_inputs)
+
+        # convert and verify scales are what we expect
+        scale1 = m.linear1.weight_fake_quantizer.scale
+        scale2 = m.linear2.weight_fake_quantizer.scale
+        sub_scale = m.sub.linear.weight_fake_quantizer.scale
+        if base_config_cls == Int8DynamicActivationInt4WeightConfig:
+            base_config = base_config_cls()
+            quantize_(m, QATConfig(base_config, step="convert"))
+            torch.testing.assert_close(
+                m.linear1.weight.original_weight_tensor.tensor_impl.scale, scale1
+            )
+            torch.testing.assert_close(
+                m.linear2.weight.original_weight_tensor.tensor_impl.scale, scale2
+            )
+            torch.testing.assert_close(
+                m.sub.linear.weight.original_weight_tensor.tensor_impl.scale, sub_scale
+            )
+        else:
+            base_config = base_config_cls(torch.int4, PerGroup(group_size))
+            quantize_(m, QATConfig(base_config, step="convert"))
+            torch.testing.assert_close(m.linear1.weight.scale, scale1)
+            torch.testing.assert_close(m.linear2.weight.scale, scale2)
+            torch.testing.assert_close(m.sub.linear.weight.scale, sub_scale)
 
 
 instantiate_parametrized_tests(TestQAT)
--- a/torchao/quantization/qat/api.py
+++ b/torchao/quantization/qat/api.py
@@ -21,6 +21,7 @@ from .embedding import FakeQuantizedEmbedding
 from .fake_quantize_config import (
     FakeQuantizeConfig,  # noqa: F401, for BC
     FakeQuantizeConfigBase,
+    IntxFakeQuantizeConfig,
     _infer_fake_quantize_configs,
 )
 from .linear import FakeQuantizedLinear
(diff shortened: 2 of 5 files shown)


torchao/quantization/qat/api.py
```diff
@@ -227,15 +228,26 @@ def _qat_config_transform(
...
         assert config.weight_config is None, "unexpected `weight_config`"
+        kwargs = {}
         if isinstance(module, FakeQuantizedLinear):
+            # Optionally pass custom scales and zero points to base config handler
+            weight_config = module.weight_fake_quantizer.config
+            if (
+                isinstance(weight_config, IntxFakeQuantizeConfig)
+                and weight_config.range_learning
+            ):
+                kwargs["custom_scales"] = weight_config.scale
```
metascroy: In intx_unpacked_to_int8_tensor, you called this custom_scale not custom_scales

torchao/quantization/quantize_/workflows/intx/intx_unpacked_to_int8_tensor.py
```diff
@@ -177,20 +177,25 @@ def from_hp(
...
-        scale, zero_point = choose_qparams_affine(
-            hp_tensor,
-            mapping_type,
-            block_size,
-            target_dtype=torch.int8,
-            quant_min=qmin,
-            quant_max=qmax,
-            zero_point_dtype=torch.int8,
-        )
+        if custom_scale is None or custom_zero_point is None:
```
metascroy: assert custom_scale and custom_zero_point are both None or both not None

torchao/quantization/quant_api.py
```diff
@@ -2207,6 +2215,7 @@ def _intx_weight_only_quantize_tensor(weight, config):
                 block_size,
                 weight_dtype,
                 mapping_type=mapping_type,
+                **kwargs,
             )
             if scale_dtype is not None and scale_dtype != weight.dtype:
                 _adjust_scale_dtype_in_intx_unpacked_tensor(
```
metascroy, line 2249: Presumably you don't want this to run if you passed custom scales?

torchao/quantization/qat/api.py
andrewor14: ah nice catch

torchao/quantization/quant_api.py
andrewor14: I looked into this a bit and I think it's actually fine to run this? Basically we use the custom scale first and then cast it to the configured scale dtype (otherwise `custom_scale` will have to leak into this file, which I hope to avoid)

metascroy: I think you're right

metascroy (approved):

test/quantization/test_qat.py
```diff
@@ -2304,6 +2304,42 @@ def shuffle_and_pack(t: torch.Tensor, scale: torch.Tensor) -> torch.Tensor:
...
+        """
+        group_size = 32
+        config = IntxFakeQuantizeConfig(
+            torch.int4,
+            group_size=group_size,
+            is_symmetric=True,
+            is_dynamic=False,
+            range_learning=True,
+            scale_precision=torch.float32,
+            zero_point_precision=torch.float32,
```
metascroy: We should not use float32 zeros we we want to lower to XNNPACK.

We should convert the int32 zeros to int8 when passing in custom_zero_point

torchao/quantization/quant_api.py
```diff
@@ -809,7 +809,9 @@ def __post_init__(self):
                     )
 
 
-def _int8_dynamic_activation_intx_weight_quantize_tensor(weight, bias, config):
+def _int8_dynamic_activation_intx_weight_quantize_tensor(
+    weight, bias, config, **kwargs
```
jerryzh168: spell out the kwargs might be better?

torchao/quantization/qat/api.py
```diff
@@ -227,15 +228,26 @@ def _qat_config_transform(
...
         assert config.weight_config is None, "unexpected `weight_config`"
+        kwargs = {}
         if isinstance(module, FakeQuantizedLinear):
+            # Optionally pass custom scales and zero points to base config handler
+            weight_config = module.weight_fake_quantizer.config
+            if (
+                isinstance(weight_config, IntxFakeQuantizeConfig)
+                and weight_config.range_learning
+            ):
+                kwargs["custom_scale"] = module.weight_fake_quantizer.scale
```
jerryzh168: also wondering if this should be part of config?

torchao/quantization/quant_api.py
andrewor14: like `tensor_subclass_kwargs`?

torchao/quantization/qat/api.py
andrewor14: these configs are actually per-model, but scales and zero points are per layer (linear)

jerryzh168: I see

torchao/quantization/quant_api.py
jerryzh168: no, I mean use `custom_scale` and `custom_zero_point`, since kwargs is more complicated than this

andrewor14: done

torchao/quantization/quant_api.py
```diff
@@ -847,12 +862,16 @@ def _int8_dynamic_activation_intx_weight_quantize_tensor(weight, bias, config):
             intx_packing_format == IntxPackingFormat.UNPACKED_TO_INT8
             or intx_packing_format in opaque_formats
         ), f"Unsupported packing format: {intx_packing_format}"
+        if custom_zero_point.dtype == torch.int32:
+            custom_zero_point = custom_zero_point.to(torch.int8)
```
jerryzh168: should we do assert instead?

torchao/quantization/quant_api.py
```diff
@@ -2202,11 +2235,15 @@ def _intx_weight_only_quantize_tensor(weight, config):
 
     if config.version == 2:
         if config.intx_packing_format == IntxPackingFormat.UNPACKED_TO_INT8:
+            if custom_zero_point.dtype == torch.int32:
```

torchao/quantization/quant_api.py
andrewor14: this is to make the flow work by default without user having to specify the zero point dtype (QAT's default is int32), so adding an assert will fail that

PR #3029, opened 2025-09-18
Add main tensor conversion API for packed tensors

jerryzh168:
Summary:
Added `convert_to_packed_tensor_based_on_current_hardware` to convert a tensor from the unpacked / plain version to a packed version

This is to enable vllm for packed weights, vllm will do a slice for the quantized weight, but slicing is not always supported for all torchao tensor subclasses. So we want to first ship an plain / unpacked checkpoint and then convert to the packed version using this API

Test Plan:
pytest test/prototype/test_tensor_conversion.py

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/prototype/tensor_conversion/api.py
+++ b/torchao/prototype/tensor_conversion/api.py
@@ -7,7 +7,14 @@
 import torch
 import torch.nn as nn
 
-from torchao.quantization.quantize_.workflows import IntxUnpackedToInt8Tensor
+# TODO: move the function to torchao.utils
+from torchao.dtypes.utils import is_device
+from torchao.quantization import (
+    Int4PreshuffledTensor,
+    Int4Tensor,
+    IntxUnpackedToInt8Tensor,
+)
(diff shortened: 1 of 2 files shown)


torchao/prototype/tensor_conversion/api.py
```diff
@@ -156,3 +163,19 @@ def _convert_model_for_aarch64(
             raise ValueError(f"Unexpected tensor_type={tensor_type}")
 
     return model
+
+
+def _convert_to_packed_tensor_based_on_current_hardware(tensor: TorchAOBaseTensor):
```
andrewor14: Is the intention for 3rd party libraries like vLLM to call this? If so maybe we should remove the underscore. It should be OK to break BC later since this is under prototype

andrewor14 (approved):

torchao/prototype/tensor_conversion/api.py
jerryzh168: yeah this is for vllm, sure

PR #3030, opened 2025-09-18
Avoid normalization layers in HF's quantization_config

lisjin:
This is a follow-up to #3015. I found that setting `model.config.quantization_config` with a "_default" key will quantize not just linear layer weights, but all weights, including normalization layers.

The culprit is still [this](https://github.com/huggingface/transformers/blob/5f6e278a5177d8b85945a2cdb6b776dacee34914/src/transformers/quantizers/quantizer_torchao.py#L277) line from TorchAoHfQuantizer.create_quantized_param:
```python
quantize_(module, c, filter_fn=lambda x, fqn: True)
```

In order to avoid quantizing non-linear weights, I had to manually add all module names for normalization layers to `modules_to_not_convert`, which is an argument to `TorchAoConfig`.

@andrewor14 Do you know if this is the intended behavior? I don't see why they aren't using the default `filter_fn` for `quantize_`.

### DIFF
--- a/test/prototype/test_parq.py
+++ b/test/prototype/test_parq.py
@@ -27,9 +27,13 @@ from torchao.prototype.parq.quant import (
     UnifQuantizer,
     UnifTorchaoQuantizer,
 )
-from torchao.prototype.parq.quant.config_torchao import TRANSFORMERS_AVAIL, _is_hf_model
+from torchao.prototype.parq.quant.config_torchao import (
+    TRANSFORMERS_AVAIL,
+    _attach_hf_quantization_config,
+    _is_hf_model,
+)
 from torchao.prototype.parq.quant.uniform_torchao import _BIT_WIDTH_TO_DTYPE
-from torchao.quantization.granularity import PerGroup
+from torchao.quantization.granularity import PerAxis, PerGroup
 from torchao.quantization.qat import IntxFakeQuantizeConfig, QATConfig
 from torchao.quantization.quant_api import (
     Int4WeightOnlyConfig,
@@ -48,6 +52,10 @@ from torchao.utils import (
 )
 
 _DEVICE = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+
+if TRANSFORMERS_AVAIL:
+    from transformers import PretrainedConfig, TorchAoConfig
+    from transformers.quantizers.quantizer_torchao import TorchAoHfQuantizer
 
 
 def split_param_groups(model) -> tuple[list, list, list]:
@@ -206,8 +214,11 @@ class M(nn.Module):
 
         if embedding and tied_weights:
             assert self.embedding.weight.shape == self.linear2.weight.shape
-            self.linear2.weight = self.embedding.weight
+            self.tie_weights()
             self._tied_weights_keys.append("linear2.weight")
+
+    def tie_weights(self):
+        self.linear2.weight = self.embedding.weight
 
     def reset_parameters(self):
         for module in (self.linear1, self.linear2):
@@ -222,6 +233,9 @@ class M(nn.Module):
             k = self.embedding.num_embeddings
             inputs = torch.randint(1, k, (1, self.linear1.in_features), device=device)
         return inputs
+
+    def get_input_embeddings(self) -> nn.Module:
+        return self.embedding
 
     def forward(self, x):
         x = self.embedding(x)
@@ -506,8 +520,6 @@ class TestInt8DynamicActivationTorchaoQuantizer(common_utils.TestCase):
 
         attach_hf_config = False
         if TRANSFORMERS_AVAIL:
-            from transformers import PretrainedConfig
-
             model.config = PretrainedConfig()  # pretend this is a HF model
             attach_hf_config = _is_hf_model(model)
             self.assertTrue(attach_hf_config)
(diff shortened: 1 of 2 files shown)


andrewor14:
I think that's the intended behavior but just wanted to cc @jerryzh168 to confirm.

There are two ways to skip quantizing certain layers today:
1. [`ModuleFqnToConfig`](https://huggingface.co/docs/transformers/main/quantization/torchao#1-skip-quantization-for-certain-layers), more flexible but also a bit more verbose
2. [`modules_to_not_convert`](https://huggingface.co/docs/transformers/en/main_classes/quantization#transformers.TorchAoConfig), just specify a list of module names to not convert

However, neither of these accept regex today so you'll have to specify all the modules you want to skip manually, which may be a bit brittle

torchao/prototype/parq/quant/config_torchao.py
```diff
@@ -193,8 +193,9 @@ def _attach_hf_quantization_config(
         if not hasattr(module, "weight"):
             continue
 
+        # Do not quantize pointers to tied weights or normalization layers
         data_ptr = module.weight.data_ptr()
-        if data_ptr in seen_data_ptrs:  # do not re-quantize tied weight
+        if data_ptr in seen_data_ptrs or name.endswith("norm"):
```
andrewor14: maybe it's better to filter out the class instead? Is there a common class we can use here for that?

lisjin: I was thinking about checking for something like `isinstance(module, (nn.LayerNorm, nn.RMSNorm))`, but it looks like HF defines a model-specific [`Qwen3RMSNorm`](https://github.com/huggingface/transformers/blob/91393fe4cc3266a05bc0d129e34ff5f761bb46e2/src/transformers/models/qwen3/modeling_qwen3.py#L50) that only inherits from `nn.Module`

jerryzh168: "norm" in name might be more general I think

jerryzh168: also using `data_ptr` seems a bit fragile and hacky, any other way to detect sharing? (like checking id of weight?)

lisjin: @jerryzh168 Could you share more about checking id of the weight? I've used `data_ptr` consistently in the PARQ code and [this](https://dev-discuss.pytorch.org/t/weight-sharing-on-cuda/701/2) suggests it's a valid way to check that the underlying storage is the same (correct me if I'm wrong)

jerryzh168: OK, I was thinking of `id(weight)` but that's probably not going to work since it's python object id

although, since you are working with transformer models, is it possible to rely on some other information/API to detect shared weights?

e.g. https://huggingface.co/pytorch/Qwen3-4B-INT8-INT4#untie-embedding-weights might be helpful? specifically `find_tied_parameters`

lisjin: Sure, I added a check for membership in `model._tied_weights_keys` instead of `seen_data_ptrs`. I also added a test case for what happens in `PreTrainedModel.from_pretrained` when given a quantization config. That method iterates over `model.state_dict().items()` instead of `model.named_parameters()`, which includes tied weights.

jerryzh168: thanks, btw `named_parameters(remove_duplicate=False)` will probably give you the tied weights

test/prototype/test_parq.py
```diff
@@ -530,6 +542,55 @@ def test_int8_dynamic_activation_intx_e2e(
...
+        configs = [embed_config]
+        filter_fns = [lambda m: isinstance(m, nn.Embedding)]
+        _attach_hf_quantization_config(model, filter_fns, configs, module_to_config)
+
+        quantization_config = getattr(model.config, "quantization_config", None)
+        self.assertTrue(isinstance(quantization_config, TorchAoConfig))
+        self.assertTrue(quantization_config.modules_to_not_convert == ["linear2"])
+
+        # Simulate transformers.PreTrainedModel.from_pretrained
+        hf_quantizer = TorchAoHfQuantizer(
```
jerryzh168: do we need to call this directly? can this be achieved with normal `PreTrainedModel.from_pretrained(..., quantization_config=TorchAoConfig(quantization_config, modules_to_not_convert=...))`

I'm not sure how stable all these APIs for `TorchAoHfQuantizer` are

torchao/prototype/parq/quant/config_torchao.py
lisjin: I see. I think model._tied_weights_keys will give us the names of the pointers. I'm not sure what method remove_duplicates is using to distinguish pointer params from the reference

test/prototype/test_parq.py
lisjin: Ok, I've replaced it with PreTrainedModel.from_pretrained

test/prototype/test_parq.py
```diff
@@ -50,6 +55,84 @@
...
+        ):
+            super().__init__(**kwargs)
+            self.m = m
+            self.n = n
+            self.k = k
+            self.bias = bias
+            self.embedding = embedding
+            self.tied_weights = tied_weights
+
+    class PreTrainedM(M, PreTrainedModel):
```
jerryzh168, line 116: oh I meant just using some specific model defined in transformers, and use the public APIs, just making sure, would the tests work for existing models in transformers?

jerryzh168 (approved):
lg, see comments inline, just want to make sure that the APIs we are using are applicable to all huggingface models

torchao/prototype/parq/quant/config_torchao.py
jerryzh168: remove_duplicate=False is not removing the duplicated params, so you can probably find the tied params manually in some way I think, but using tied_weights_keys sounds reasonable as well since it seems to be defined in most of transformer model definitions

test/prototype/test_parq.py
lisjin: Yes, existing transformers models also inherit from `PreTrainedModel`. `AutoModelForCausalLM.from_pretrained(..., quantization_config=quantization_config)` can be tested in the same way

metascroy:
> I think that's the intended behavior but just wanted to cc @jerryzh168 to confirm.

This is not at all what I would expect for how default\_ to behave since the "default" in quantize\_ is to only apply to linear.

@jerryzh168 this behavior means we're quantizing more than we expect in our hummingbird models?

jerryzh168:
> This is not at all what I would expect for how default_ to behave since the "default" in quantize_ is to only apply to linear.

yeah this is the intended behavior since now user can do the filtering manually when they generate ModuleFqnToConfig

it will still fail if the module does not have `.weight` though, due to the implementation details of quantize_

for the models we quantize, I think it's still just linears that are quantized, otherwise these are likely to fail: https://github.com/pytorch/ao/blob/3d48174441edb7940c5aae17a4c1125d9b91ab60/torchao/quantization/quant_api.py#L965-L977

although it does make sense to have better error messages when it's not what we expect

PR #3034, opened 2025-09-19
[mxfp8 moe training] update 3d quant colwise scaling kernel to use single input/output TMA descriptors

danielvegamyhre:
Stacked PRs:
 * __->__#3034

--- --- ---

[mxfp8 moe training] update 3d quant colwise scaling kernel to use single input/output TMA descriptors

## Summary
- CUDA Kernel for 3d quantization across cols added in #3004 has worse perf than other methods for small num_experts, and is only better for large num_experts. 
- We hypothesize this is because cudaMallocManaged of separate TMA descriptors per expert, which is a slow/blocking function. The overhead is constant, and thus more noticeable for small inputs.
- In this PR, I redesign the kernel to use single input and output TMA descriptor for the whole 3d tensor.
    - For the input, it is in simple row major format, so I can read from specific experts by adjusting the TMA row offset during the async TMA load.
    - For the output, it is a more complex "column major per expert" format, so I use a 3d TMA descriptor with the specific shape and strides needed. I transpose the row major data in SMEM before doing the async TMA store to GMEM to get it in col major per expert format.

## Test plan
- Add unit tests for Llama4 and DeepSeekV3 shapes
- `sanitize pytest test/prototype/moe_training/test_kernels.py `

## Performance
devgpu having issues, perf for everything isworse than usual, but the relative difference between the cuda_3d vs other methods is still useful for now:

```
input_shape         to_mx_us    cuda_2d_us    cuda_3d_us    to_mx_gbps    cuda_2d_gbps    cuda_3d_gbps
----------------  ----------  ------------  ------------  ------------  --------------  --------------
(1, 8192, 5120)      816.608       595.936       466.736       155.693         213.345         272.402
(2, 8192, 5120)     1174.82        919.776       508.48        216.442         276.458         500.078
(4, 8192, 5120)     1631.71       1159.58        631.712       311.672         438.571         805.049
(8, 8192, 5120)     2621.44       1602.82        839.424       388             634.582        1211.69
(16, 8192, 5120)    4612.42       2772.93       1291.07        441.035         733.606        1575.62
(64, 8192, 5120)   15916.1        8551.81       4112.35        511.241         951.489        1978.66
```

### DIFF
--- a/torchao/csrc/cuda/mx_kernels/mxfp8_quantize.cuh
+++ b/torchao/csrc/cuda/mx_kernels/mxfp8_quantize.cuh
@@ -22,7 +22,6 @@
 #include <cuda/barrier>
 #include <cuda/ptx>
 
-
 #define MIN_CUDA_SM 1000 // SM90 = 900, SM100 = 1000
 
 // Check if we're compiling for supported architecture
@@ -37,6 +36,17 @@
 #else
 #define HAS_NATIVE_FP8_CONVERSION 0
 #endif
+
+// Macro to check CUDA error.
+#define CUDA_CHECK(call)                                                      \
+do {                                                                          \
+    cudaError_t err = call;                                                   \
+    if (err != cudaSuccess) {                                                 \
+        fprintf(stderr, "CUDA Error in %s at line %d: %s\n",                  \
+                __FILE__, __LINE__, cudaGetErrorString(err));                 \
+        throw std::runtime_error(cudaGetErrorString(err));                    \
+    }                                                                         \
+} while (0)
 
 enum class DType {
   kByte,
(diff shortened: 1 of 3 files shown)


ngimel (approved):

torchao/csrc/cuda/mx_kernels/mxfp8_quantize.cuh
```diff
@@ -941,15 +984,18 @@ __global__ void __launch_bounds__(MXFP8_THREADS_PER_CHUNK)
...
+  const int expert_logical_base_row = expert_idx * N;
 
   // The destination shared memory buffer of a bulk tensor operation should be
   // 128 e8m0_t aligned
   __shared__ alignas(128)
       IType in_sh[MXFP8_BUFFERS_NUM][MXFP8_SHMEM_DIM_Y][MXFP8_SHMEM_DIM_X];
+
+  // SMEM buffer for expert must be 3d since we use cp async bulk tensor 3d ptx instruction.
+  // We parallelize across experts, so leading "E" dim will always be 1 for single expert.
+  const size_t smem_e = 1;
```
drisspg: nit: constexpr

ngimel:
What are cuda_2d numbers in the benchmark? Running in a loop for each expert?

danielvegamyhre:
> What are cuda_2d numbers in the benchmark? Running in a loop for each expert?

The benchmarks are from [this script](https://github.com/pytorch/ao/blob/main/benchmarks/prototype/moe_training/mxfp8/bench_quantize_3d.py).

The "cuda_2d" benchmarks are referencing this [function](https://github.com/pytorch/ao/blob/ae12e429d4acb29db00d3264e2fa99187d686906/torchao/prototype/moe_training/scaled_grouped_mm.py#L451) which uses the 2d cuda colwise quantization on the 3d tensor, by reshaping it in pytorch from (E*N, K), quantizing, then reshaping the output and scales appropriately back to 3d, to match the expectations of torch._scaled_grouped_mm mxfp8 2d-3d grouped gemm.

The key issue with this method is that the quantized (E*N, K) is in column major format, and I couldn't find a way to reshape/view back to (E,N,K) with per expert column major format by simplying mutating the tensor metadata - I had to do a physical memory layout transformation [here](https://github.com/pytorch/ao/blob/ae12e429d4acb29db00d3264e2fa99187d686906/torchao/prototype/moe_training/scaled_grouped_mm.py#L482), which is not ideal. 

So I called that method `cuda_2d` in the script since it's using the 2d quantization kernel (name could be better, just trying to keep it concise).

torchao/csrc/cuda/mx_kernels/mxfp8_quantize.cuh
```diff
@@ -345,6 +344,56 @@ inline CUtensorMapDataType get_dtype_for_tma(DType dtype) {
...
+                                         DType dtype,
+                                         const size_t E,
+                                         const size_t N,
+                                         const size_t K,
+                                         uint32_t shmem_e,
+                                         uint32_t shmem_n,
+                                         uint32_t shmem_k,
+                                         const size_t type_num_bits) {
+  // Get function pointer to cuTensorMapEncodeTiled
+  static void *driver_ptr = nullptr;
```
ngimel: since both 2d and 3d map are using this, I think you should use a function that would return it (to not initialize it twice)

danielvegamyhre: Hmm good point. Updated to do something like a singleton pattern, holding driver ptr as a global/static var that starts as null then we only initialize it once whenever the first kernel is called. Let me know if that is what you had in mind / will work. (It builds and tests pass)

torchao/csrc/cuda/mx_kernels/mxfp8_quantize.cuh
```diff
@@ -345,6 +344,56 @@ inline CUtensorMapDataType get_dtype_for_tma(DType dtype) {
   }
 }
 
+static void *driver_ptr = nullptr;
```
ngimel: I'd prefer something like 
```
void * get_driver_ptr() {
static void * driver_ptr = nullptr;
  if (!driver_ptr) {
    cudaDriverEntryPointQueryResult result;
    cudaGetDriverEntryPoint("cuTensorMapEncodeTiled", &driver_ptr,
                            cudaEnableDefault, &result);
  }
  return driver_ptr;
}
```
that gets called from both create_3D_tensor_map and create_2D_tensor_map, but this would work too. If you are going with this you should put it in an anonymous namespace to not pollute other files that may include this, `driver_ptr` is pretty generic.

danielvegamyhre: Makes sense - updated.

torchao/csrc/cuda/mx_kernels/mxfp8_quantize.cuh
```diff
@@ -345,6 +344,61 @@ inline CUtensorMapDataType get_dtype_for_tma(DType dtype) {
   }
 }
 
+void* get_driver_ptr() {
+  // Only initialize driver_ptr once during the lifetime of the program.
+  static void *driver_ptr = nullptr;
+  if (!driver_ptr) {
+    cudaDriverEntryPointQueryResult result;
+    cudaGetDriverEntryPoint("cuTensorMapEncodeTiled", &driver_ptr,
```
ngimel: also, you should error check this call

danielvegamyhre: thanks, done

PR #3012, opened 2025-09-16
Update SmoothQuant to use SupportsActivationPreScaling protocol instead of external wrapper

namgyu-youn:
### Summary
Update SmoothQuant to use SupportsActivationPreScaling protocol instead of external wrapper

Similar to AWQ (https://github.com/pytorch/ao/pull/2753), SmoothQuant
now uses direct protocol implementation instead of
`to_weight_tensor_with_linear_activation_scale_metadata` wrapper.

- Add `act_pre_scale` attribute to `LinearActivationQuantizedTensor`
- Apply pre-scaling in all dispatch methods
- Remove external wrapper dependency

### Test Plan
```bash
python torchao/prototype/smoothquant/example.py
python test/prototype/test_smoothquant.py
python test/integration/test_integration.py
```

Issue #2752, opened 2025-08-12
Migrating from AffineQuantizedTensor + Layouts to new structure of tensor subclasses


jerryzh168:
Note: AffineQuantizedTensor + Layouts are scheduled to be deprecated in 0.17.0 and removed in 0.18.0!

Context:
Previously we use AffineQuantizedTensor for many of our use cases including int4, float8, intx, floatx. It introduces some complicated abstractions like Layout, people have been saying it's a bit hard to understand, and there are many indirections in the code.

As an effort simplify the code base and make it easier to contribute to, we have been adding new features with a different structure in mind. Now we want to structure Tensors by "dtype" and "packing_format", e.g. we'll have Int4PreshuffledTensor, Int8Tensor, Float8Tensor, instead of having AffineQuantizedTensor and multiple layouts.

Please check out our updated docs for the new tensor subclass organization structure and guide for design:
* quantization overview: https://docs-preview.pytorch.org/pytorch/ao/2723/quantization_overview.html
* contributor guide: https://docs-preview.pytorch.org/pytorch/ao/2723/contributor_guide.html
* Examples of tensor subclasses following new design: https://github.com/pytorch/ao/tree/main/torchao/quantization/quantize_/workflows

## migration status

All layouts has been deleted, we will start deleting AffineQuantizedTensor now, @andrewor14 currently working on this.

| inference config name | current status | plan | POC | notes |
| ----------------------------------------- | ---- |  --- | --- | --- |
| `Int8WeightOnlyConfig` | v1 exists | create v2, then deprecate v1 | ? | https://github.com/pytorch/ao/pull/3391 added v2, https://github.com/pytorch/ao/pull/3407 cleaned up, can deprecate in 0.17.0, done: https://github.com/pytorch/ao/pull/4151 |
| `Int8DynamicActivationInt8WeightConfig` | v1 exists | create v2, then deprecate v1, delete v1: https://github.com/pytorch/ao/pull/4019 | ? | https://github.com/pytorch/ao/pull/3391 added v2, https://github.com/pytorch/ao/pull/3407 cleaned up, can deprecate in 0.17.0, done: https://github.com/pytorch/ao/pull/4151 |
| `Int8DynamicActivationInt4WeightConfig` | v1 exists | move to prototype | ? | https://github.com/pytorch/ao/pull/3491 moved to prototype in 0.16.0, can deprecate in 0.17.0, done: https://github.com/pytorch/ao/pull/3884 |
[...]

Issue #2919, opened 2025-09-02
torchao release compatibility table


vkuzo:
This issue describes the compatibility matrix between torchao releases and its dependencies.  If you are seeing an error when importing `torchao` that looks like this,

```bash
(pytorch_nightly) [vasiliy@devgpu007.eag6 ~/local]$ python -c "import torchao"
Fatal Python error: Aborted
```

then most likely you can resolve this error by ensuring that the `torch` version in your environment is compatible with the `torch` version used to build your `torchao` version.

# torch

| torchao version | torch version | torch version, torchao's Python API only |
|----------------|----------------|------------------------------------|
| 0.15.0dev (nightly)  | 2.10.0dev (nightly)      | 2.10.0, 2.9.0, 2.8.0     |
| 0.14.1                 | 2.9.0                      | 2.9.0, 2.8.0, 2.7.1                                               |
| 0.13.0                 | 2.8.0                     | 2.8.0, 2.7.1, 2.6.0                            |
| 0.12.0                 | 2.7.1, 2.6.0, 2.5.0  | n/a                                                  |

# fbgemm_gpu

`torchao` has an optional runtime dependency on `fbgemm_gpu`. Please see https://docs.pytorch.org/FBGEMM/general/Releases.html for the compatibility matrix for `fbgemm_gpu`.  Note that while `torchao`'s Python API supports multiple `torch` versions, each `fbgemm_gpu` version only supports a single `torch` version.  Therefore, if you are using `torchao` together with `fbgemm_gpu`, you should use the `torch` version corresponding to your `fbgemm_gpu` version.

jerryzh168:
any ideas when we can fix the compatibility of torchao nightly with torch? currently it's blocking tests in vllm: https://github.com/vllm-project/vllm/pull/21982/files#diff-2fe466060a88bb6a57175df8ca7175849db82a2cf2ba082295d481ab57e58868R512

yushangdi:
I'm also having the same issue here. Blocking testing executorch with torch nightly.

I have torch on latest master (with some python-only local changes) `'2.9.0a0+git3564a8a'`, and torchao `torchao-0.14.0.dev20250909+cu126`.

yushangdi:
> I'm also having the same issue here. Blocking testing executorch with torch nightly.
> 
> I have torch on latest master (with some python-only local changes) `'2.9.0a0+git3564a8a'`, and torchao `torchao-0.14.0.dev20250909+cu126`.

[...]

PR #3038, opened 2025-09-21
Add Int8Tensor for clearer interface

namgyu-youn:
**Summary**:
Introduce new tensor subclass API for int8 quantization with clearer interface.

The main change can be summarized to the following:
- Old: Complex affine transform (`AffineQuantizedTensor`) with separate layout handling
- New: Direct int8 tensor with scaling factor and zero point

Related Issue/PR: https://github.com/pytorch/ao/pull/3012#issuecomment-3305023078 #2752

**Test plan**:
test/quantization/quantize_/workflows/int8/test_int8_tensor.py

### DIFF
--- a/test/quantization/quantize_/workflows/int8/test_int8_tensor.py
+++ b/test/quantization/quantize_/workflows/int8/test_int8_tensor.py
@@ -58,20 +58,28 @@ class ToyTwoLinearModel(torch.nn.Module):
 class TestInt8Tensor(TorchAOIntegrationTestCase):
     def setUp(self):
         super().setUp()
+
+        self.test_shape = (4, 3)
+        self.dtype = torch.bfloat16
+        self.batch_size = 32
+        self.int8_min = -128
+        self.int8_max = 127
+
         torch.manual_seed(42)
-        self.weight_fp = torch.randn(4, 3, dtype=torch.bfloat16)
-        self.input_fp = torch.randn(4, 3, dtype=torch.bfloat16)
-        self.bias = torch.randn(4, dtype=torch.bfloat16)
-        self.block_size = [4, 3]
+        self.weight_fp = torch.randn(*self.test_shape, dtype=self.dtype)
+        self.input_fp = torch.randn(*self.test_shape, dtype=self.dtype)
+        self.bias = torch.randn(self.test_shape[0], dtype=self.dtype)
+        self.block_size = list(self.test_shape)
 
     def test_creation_and_attributes(self):
         """Test tensor creation, dtypes, and ranges"""
         tensor = Int8Tensor.from_hp(self.weight_fp, self.block_size)
 
-        self.assertEqual(tensor.shape, (4, 3))
+        self.assertEqual(tensor.shape, self.test_shape)
         self.assertEqual(tensor.qdata.dtype, torch.int8)
         self.assertTrue(
-            torch.all(tensor.qdata >= -128) and torch.all(tensor.qdata <= 127)
+            torch.all(tensor.qdata >= self.int8_min)
+            and torch.all(tensor.qdata <= self.int8_max)
         )
 
     @common_utils.parametrize("dtype", [torch.bfloat16, torch.float16])
@@ -88,12 +96,13 @@ class TestInt8Tensor(TorchAOIntegrationTestCase):
             Int8WeightOnlyConfig(version=2),
         ],
     )
-    def test_int8_linear_variants(
+    def test_int8_linear_quantization_accuracy(
         self,
         dtype: torch.dtype,
         sizes: tuple,
         config,
     ):
+        """Test quantization preserves reasonable accuracy"""
         M, N, K = sizes
         input_tensor = torch.randn(*M, K, dtype=dtype, device="cuda")
 
@@ -108,14 +117,16 @@ class TestInt8Tensor(TorchAOIntegrationTestCase):
         output_quantized = m_q(input_tensor)
 
         error = compute_error(output_original, output_quantized)
-        assert error > 20, f"Quantization error is too high got a SQNR of {error}"
+        assert error > 20, (
+            f"Quantization quality is too low, SQNR: {error}dB (expected > {20}dB)"
+        )
 
     @common_utils.parametrize("dtype", [torch.bfloat16, torch.float16])
-    def test_static_dynamic_quantization(self, dtype):
-        """Test static and dynamic quantization"""
+    def test_quantization_shapes(self, dtype):
+        """Test static and dynamic quantization output shapes"""
         K, N = 128, 64
         weight = torch.randn(N, K, dtype=dtype, device="cuda")
-        input_tensor = torch.randn(32, K, dtype=dtype, device="cuda")
+        input_tensor = torch.randn(self.batch_size, K, dtype=dtype, device="cuda")
 
         # Dynamic quantization (runtime scale computation)
         dynamic_tensor = Int8Tensor.from_hp(weight, block_size=[N, K])
@@ -126,8 +137,8 @@ class TestInt8Tensor(TorchAOIntegrationTestCase):
             mapping_type=MappingType.SYMMETRIC,
             block_size=(input_tensor.shape[0], K),
             target_dtype=torch.int8,
-            quant_min=-128,
-            quant_max=127,
+            quant_min=self.int8_min,
+            quant_max=self.int8_max,
             scale_dtype=dtype,
             zero_point_dtype=torch.int8,
         )
@@ -145,25 +156,29 @@ class TestInt8Tensor(TorchAOIntegrationTestCase):
         dynamic_output = torch.nn.functional.linear(input_tensor, dynamic_tensor)
         static_output = torch.nn.functional.linear(input_tensor, static_tensor)
 
-        self.assertEqual(dynamic_output.shape, (32, N))
-        self.assertEqual(static_output.shape, (32, N))
+        expected_shape = (self.batch_size, N)
+        self.assertEqual(dynamic_output.shape, expected_shape)
+        self.assertEqual(static_output.shape, expected_shape)
         self.assertEqual(dynamic_output.dtype, dtype)
         self.assertEqual(static_output.dtype, dtype)
 
     @unittest.skip("granularity parameter not supported in current API")
     @common_utils.parametrize("granularity", [PerTensor(), PerRow()])
     def test_slice_preserves_aliasing(self, granularity):
+        slice_size = 512
+        tensor_size = 1024
+
         config = Int8DynamicActivationInt8WeightConfig(
             granularity=granularity, version=2
         )
-        l = torch.nn.Linear(1024, 1024).to("cuda").to(torch.bfloat16)
+        l = torch.nn.Linear(tensor_size, tensor_size).to("cuda").to(torch.bfloat16)
         l.weight = torch.nn.Parameter(
-            torch.zeros(1024, 1024, dtype=torch.bfloat16, device="cuda")
+            torch.zeros(tensor_size, tensor_size, dtype=torch.bfloat16, device="cuda")
         )
         quantize_(l, config)
         param = l.weight
         param_data = param.data
-        param_data = param_data.narrow(0, 0, 512)
+        param_data = param_data.narrow(0, 0, slice_size)
         # Making sure the aliasing is preserved in sliced quantized Tensor
         assert param.data.qdata.data_ptr() == param_data.qdata.data_ptr()
         assert param.data.scale.data_ptr() == param_data.scale.data_ptr()
@@ -179,20 +194,27 @@ class TestInt8Tensor(TorchAOIntegrationTestCase):
     @common_utils.parametrize("dtype", [torch.bfloat16, torch.float16])
     def test_slice(self, config, device, dtype):
         """Test tensor slicing"""
-        dummy = torch.nn.Linear(256, 256, bias=False, dtype=dtype, device=device)
+        tensor_size = 256
+        slice_sizes = (64, 128)
+
+        dummy = torch.nn.Linear(
+            tensor_size, tensor_size, bias=False, dtype=dtype, device=device
+        )
         quantize_(dummy, config)
 
-        weight1 = dummy.weight.clone().narrow(0, 0, 64)
-        weight2 = dummy.weight.clone().narrow(1, 0, 128)
+        weight1 = dummy.weight.clone().narrow(0, 0, slice_sizes[0])
+        weight2 = dummy.weight.clone().narrow(1, 0, slice_sizes[1])
 
-        self.assertEqual(weight1.qdata, dummy.weight.qdata.narrow(0, 0, 64))
-        self.assertEqual(weight2.qdata, dummy.weight.qdata.narrow(1, 0, 128))
+        self.assertEqual(weight1.qdata, dummy.weight.qdata.narrow(0, 0, slice_sizes[0]))
+        self.assertEqual(weight2.qdata, dummy.weight.qdata.narrow(1, 0, slice_sizes[1]))
 
         # Int8DynamicActivationInt8WeightConfig uses per-row (PerRow)
         # Int8WeightOnlyConfig uses per-tensor (PerTensor)
         if isinstance(config, Int8DynamicActivationInt8WeightConfig):
             # PerRow: dim 0 slicing affects scale, dim 1 doesn't
-            self.assertEqual(weight1.scale, dummy.weight.scale.narrow(0, 0, 64))
+            self.assertEqual(
+                weight1.scale, dummy.weight.scale.narrow(0, 0, slice_sizes[0])
+            )
             self.assertEqual(weight2.scale, dummy.weight.scale)
         else:
             # PerTensor: scale unchanged by slicing
@@ -211,20 +233,33 @@ class TestInt8Tensor(TorchAOIntegrationTestCase):
             x_int8.dequantize()[0], x_int8_0.dequantize(), atol=0, rtol=0
         )
 
-    def test_error_handling_and_dequant(self):
-        """Test input validation and dequantization accuracy"""
-        with self.assertRaises((AssertionError, ValueError, RuntimeError)):
-            Int8Tensor.from_hp(torch.randn(5), [1])
+    def test_invalid_input_handling(self):
+        """Test input validation with specific error types"""
+        invalid_tensor = torch.randn(5)
+        incompatible_block_size = [1]
 
-        with self.assertRaises((AssertionError, ValueError, RuntimeError)):
+        with self.assertRaises(
+            ValueError, msg="Should reject incompatible tensor dimensions"
+        ):
+            Int8Tensor.from_hp(invalid_tensor, incompatible_block_size)
+
+        with self.assertRaises(
+            ValueError, msg="Should reject mismatched block size dimensions"
+        ):
             Int8Tensor.from_hp(self.weight_fp, [1])
 
+    def test_dequantization_accuracy(self):
+        """Test dequantization accuracy separately"""
         test_data = torch.tensor([[1.0, -1.0]], dtype=torch.bfloat16)
         tensor = Int8Tensor.from_hp(test_data, [1, 2])
 
         dequantized = tensor.dequantize()
         self.assertEqual(dequantized.shape, test_data.shape)
-        self.assertLess(torch.abs(dequantized - test_data).max().item(), 0.1)
+        self.assertLess(
+            torch.abs(dequantized - test_data).max().item(),
+            0.1,
+            msg=f"Dequantization error exceeds tolerance of {0.1}",
+        )
 
 
 if __name__ == "__main__":
--- a/torchao/quantization/quantize_/workflows/int8/int8_tensor.py
+++ b/torchao/quantization/quantize_/workflows/int8/int8_tensor.py
@@ -133,7 +133,7 @@ class Int8Tensor(TorchAOBaseTensor):
 
         if tuple(block_size) == w.shape:
             # per-tensor
-            scale = scale.expand(w.shape)
+            pass
         elif len(scale.shape) == 1:
             # per-row, 1D -> 2D
             scale = scale.unsqueeze(-1)
@@ -190,22 +190,38 @@ def _(func, types, args, kwargs):
         w_vals_t = weight_tensor.qdata.contiguous().t()
         w_scales = weight_tensor.scale
 
-        tmp = x_vals.reshape(-1, x_vals.shape[-1])
-        x_scales_dtype = x_scales.dtype
+        tmp_shape = (-1, x_vals.shape[-1])
+        tmp = x_vals.view(tmp_shape)
 
         # Cast fp16 scale to float
         intermediate_dtype = (
-            torch.float if x_scales_dtype == torch.half else x_scales_dtype
+            torch.float if x_scales.dtype == torch.half else x_scales.dtype
         )
         # Note: CUDA doesn't support int32/int64 matmul, so we convert to float
         # Error message is NotImplementedError: "addmm_cuda" not implemented for 'Int'
         # This may introduce minor numerical differences compared to int arithmetic
         y_dot = torch.mm(tmp.to(intermediate_dtype), w_vals_t.to(intermediate_dtype))
-        y_dot_scaled = y_dot * x_scales.reshape(-1, 1).to(intermediate_dtype)
 
-        result = (y_dot_scaled * w_scales).reshape(
-            *x_vals.shape[:-1], y_dot_scaled.shape[-1]
-        )
+        # Apply activation scale
+        is_per_tensor_act = x_scales.numel() == 1
+        if is_per_tensor_act:
+            y_dot.mul_(x_scales.to(intermediate_dtype))
+        else:
+            # For block-wise activation scale, reshape to match y_dot
+            x_scales_reshaped = x_scales.view(y_dot.shape[0], -1)
+            y_dot.mul_(x_scales_reshaped.to(intermediate_dtype))
+
+        # Apply weight scale
+        is_per_tensor_weight = w_scales.numel() == 1
+        if is_per_tensor_weight:
+            result = y_dot.mul_(w_scales.to(intermediate_dtype))
+        else:
+            # Per-row weight scale - transpose and broadcast
+            w_scales_broadcast = w_scales.t().expand_as(y_dot)
+            result = y_dot.mul_(w_scales_broadcast.to(intermediate_dtype))
+
+        # Reshape back to original shape
+        result = result.view(*x_vals.shape[:-1], result.shape[-1])
         result = result.to(activation_tensor.dtype)
     else:
         # FP × INT8 (weight-only)


torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,106 @@
...
+
+from torchao.utils import TorchAOBaseTensor
+
+__all__ = ["Int8PlainInt8Tensor"]
+
+aten = torch.ops.aten
+
+
+# TODO: Implement block-wise quantization using block_size
+class Int8PlainInt8Tensor(TorchAOBaseTensor):
```
jerryzh168: nit: we can just use `Int8Tensor` if it's plain, since that's the default

jerryzh168:
can you add a version 2 and expose this tensor through https://github.com/pytorch/ao/blob/852518572b16c1fe2d3d0139dd7c86604a73aaea/torchao/quantization/quant_api.py#L1497? similar to https://github.com/pytorch/ao/blob/852518572b16c1fe2d3d0139dd7c86604a73aaea/torchao/quantization/quant_api.py#L1752

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,106 @@
...
+@implements([torch.nn.functional.linear, aten.linear.default])
+def _(func, types, args, kwargs):
+    """quantization: float -> int8"""
+    input_tensor, weight_tensor, bias = (
+        args[0],
+        args[1],
+        args[2] if len(args) > 2 else None,
+    )
+
+    if isinstance(input_tensor, Int8PlainInt8Tensor):
```
jerryzh168: we also need to quantize input_tensor in this function now, please check https://github.com/pytorch/ao/blob/9d88c164e7953acb1c12c7cf09f583976326ee69/torchao/quantization/quantize_/workflows/float8/float8_tensor.py#L263-L266

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,193 @@
...
+
+    if isinstance(input_tensor, Int8Tensor):
+        # INT8 × INT8 (dynamic)
+        x_int32 = input_tensor.qdata.to(torch.int32)
+        w_int32 = weight_tensor.qdata.to(torch.int32).t()
+
+        result = torch.mm(x_int32.view(-1, x_int32.size(-1)), w_int32)
+        scale = input_tensor.scale.view(-1, 1) * weight_tensor.scale.unsqueeze(0)
+        result = result.to(scale.dtype) * scale
+        result = result.view(*input_tensor.shape[:-1], -1)
```
jerryzh168: this is not the same as https://github.com/pytorch/ao/blob/122b30753b91d9f409a241fcb71b4d3089eb6522/torchao/dtypes/uintx/plain_layout.py#L269-L315?

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,193 @@
...
+    else:
+        # FP × INT8 (static)
```
jerryzh168: also this is the code for weight only quant I think: https://github.com/pytorch/ao/blob/122b30753b91d9f409a241fcb71b4d3089eb6522/torchao/dtypes/uintx/plain_layout.py#L250

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
jerryzh168: can you add a test to check the kernel that's used similar to https://github.com/pytorch/ao/blob/8e2ca35ea603349e71c2467e10fd371e34bf52bc/test/quantization/quantize_/workflows/float8/test_float8_tensor.py#L420 as well?

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,193 @@
...
+@dataclass
+class QuantizeTensorToInt8Kwargs(QuantizeTensorKwargs):
+    """Tensor kwargs for creating int8 tensor (either activation or weight)
+
+    Args:
+        kernel_preference (KernelPreference): kernel preference for ops like matmul, grouped matmul etc.
+        block_size (Optional[list[int]]): block size for quantization granularity
+    """
+
+    kernel_preference: KernelPreference = KernelPreference.AUTO
```
jerryzh168: seems like no multiple kernel preferences right now right? if so, we can remove this for now

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -0,0 +1,88 @@
...
+from torchao.quantization.quantize_.workflows.int8.int8_tensor import (
+    Int8Tensor,
+    QuantizeTensorToInt8Kwargs,
+)
+from torchao.quantization.utils import compute_error
+from torchao.testing.utils import TorchAOIntegrationTestCase
+
+
+@unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
+class TestInt8Tensor(TorchAOIntegrationTestCase):
```
jerryzh168, line 58: for test, maybe try to follow https://github.com/pytorch/ao/blob/main/test/quantization/quantize_/workflows/int4/test_int4_marlin_sparse_tensor.py for now and also add some tests for slicing? 

https://github.com/pytorch/ao/blob/8e2ca35ea603349e71c2467e10fd371e34bf52bc/test/quantization/quantize_/workflows/float8/test_float8_tensor.py#L158-L216 and https://github.com/pytorch/ao/blob/8e2ca35ea603349e71c2467e10fd371e34bf52bc/test/quantization/quantize_/workflows/float8/test_float8_tensor.py#L278

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
namgyu-youn: We can remove this flag, but how about adding TODO for real kernel preference? Keeping current structure might be helpful for it.

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
namgyu-youn: Done at https://github.com/pytorch/ao/pull/3038/commits/93835507aaf7003e353000ff28275074b4ffb4b2 , thanks for pointing it out.

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
namgyu-youn: > can you add a test to check the kernel that's used similar to
> 
> https://github.com/pytorch/ao/blob/8e2ca35ea603349e71c2467e10fd371e34bf52bc/test/quantization/quantize_/workflows/float8/test_float8_tensor.py#L420
> 
> as well?

Yes linked workflow should be better to prevent overhead, I will fix it.

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
namgyu-youn: Yes linked unit test is helpful for slicing (`PerTensor`, `PerRow`) test, but I didn't implemented `granularity` in this PR yet for smaller PR size. Can I address it after this PR?

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
jerryzh168: we don't have different kernel options for this one I think

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,237 @@
...
+        w: torch.Tensor,
+        block_size: list[int],
+        act_quant_kwargs: Optional[QuantizeTensorToInt8Kwargs] = None,
+        kernel_preference: KernelPreference = KernelPreference.AUTO,
+    ):
+        if w.dim() != 2 or len(block_size) != 2:
+            raise ValueError("Expected 2D tensor and block_size length 2")
+
+        # Rounding function from high precision dtype
+        scale = w.abs().max(dim=-1, keepdim=True)[0] / 127.0
```
jerryzh168: looks like block_size is not used? why is that?

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
jerryzh168: I don't think the slicing tests are specific to a granularity, you should be able to adapt it for the currently supported granularity I think

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
jerryzh168: you can checkout https://github.com/pytorch/ao/blob/8c5c33e0410e87d817c0dd154f7699734112ce0b/torchao/dtypes/uintx/plain_layout.py#L232 for expected granularity

also this should be using these quant primitive ops: https://github.com/pytorch/ao/blob/8c5c33e0410e87d817c0dd154f7699734112ce0b/torchao/quantization/quantize_/workflows/int4/int4_marlin_sparse_tensor.py#L79-L97, arguments can be found by tracing through the code path for int8 in https://github.com/pytorch/ao/blob/8c5c33e0410e87d817c0dd154f7699734112ce0b/torchao/quantization/quant_api.py#L1566 and https://github.com/pytorch/ao/blob/8c5c33e0410e87d817c0dd154f7699734112ce0b/torchao/dtypes/affine_quantized_tensor.py#L325

this might require a bit too much context, let me know if you would like us to take over

namgyu-youn: Thanks, surely want to take over! Drafted this PR for those updates, but will look into it today (6 hours later)

btw, version 2 is updated at https://github.com/pytorch/ao/commit/c53dad03d0f6f0779c31876ab1232464b93ea566 (version 1 is default)

jerryzh168 (commented):
please rebase, and let me know when this is ready for review again @namgyu-youn

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -0,0 +1,218 @@
...
+            self.block_size,
+            act_quant_kwargs=QuantizeTensorToInt8Kwargs(),
+        )
+
+        reference = torch.nn.functional.linear(self.input_fp, self.weight_fp, self.bias)
+        result_dynamic = torch.nn.functional.linear(
+            self.input_fp, weight_q8_dynamic, self.bias
+        )
+
+        self.assertEqual(result_dynamic.shape, reference.shape)
```
jerryzh168: nit: probably add a test for `compute_error` comparing floating point weight and int8+int8 weight as well

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -0,0 +1,218 @@
...
+        output_original = m(input_tensor)
+        output_quantized = m_q(input_tensor)
+
+        error = compute_error(output_original, output_quantized)
+        assert compute_error(output_original, output_quantized) > 20, (
+            f"Quantization error is too high got a SQNR of {error}"
+        )
+
+    def test_linear_operations(self):
+        """Test fp+int8 and int8+int8 linear ops"""
```
jerryzh168: this is not int8+int8 I think? this is weight only quant

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -0,0 +1,218 @@
...
+        reference = torch.nn.functional.linear(self.input_fp, self.weight_fp, self.bias)
+        result_fp = torch.nn.functional.linear(self.input_fp, weight_q8, self.bias)
+        result_q8 = torch.nn.functional.linear(input_q8, weight_q8, self.bias)
+
+        self.assertEqual(result_fp.shape, reference.shape)
+        self.assertEqual(result_q8.shape, reference.shape)
+        self.assertTrue(compute_error(result_fp, reference) > 10)
+        self.assertTrue(compute_error(result_q8, reference) > 10)
+
+    def test_dynamic_quantization(self):
```
jerryzh168: I think you can remove these 2 tests actually, since they are already tested in `test_int8_linear_variants`

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -0,0 +1,218 @@
...
+    def test_slice(self, config, device, dtype):
+        """Test tensor slicing"""
+        dummy = torch.nn.Linear(256, 256, bias=False, dtype=dtype, device=device)
+        quantize_(dummy, config)
+
+        weight1 = dummy.weight.clone().narrow(0, 0, 64)
+        weight2 = dummy.weight.clone().narrow(1, 0, 128)
+
+        self.assertEqual(weight1.qdata, dummy.weight.qdata.narrow(0, 0, 64))
+        self.assertEqual(weight2.qdata, dummy.weight.qdata.narrow(1, 0, 128))
```
jerryzh168: nit: add assert for scale as well?

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -0,0 +1,218 @@
...
+
+    def test_transpose(self):
```
jerryzh168: is this used anywhere? for most of the tensors we actually don't support transpose so far, we tend to add this only when need

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -0,0 +1,218 @@
...
+
+        self.assertEqual(transposed.shape, (3, 4))
+        self.assertEqual(transposed.block_size, [3, 4])
+
+    def test_select(self):
+        """Test select operation"""
+        weight_q8 = Int8Tensor.from_hp(self.weight_fp, self.block_size)
+        selected = weight_q8.select(0, 0)
+
+        self.assertEqual(selected.shape, (3,))
```
jerryzh168: test the data as well?

jerryzh168: you can follow this: https://github.com/pytorch/ao/blob/5346f0e169c58aeba04076ace2f99e6b4f158102/test/quantization/quantize_/workflows/float8/test_float8_tensor.py#L449

torchao/quantization/quant_api.py
```diff
@@ -1376,22 +1379,30 @@ def __post_init__(self):
...
+            weight,
+            mapping_type,
+            block_size,
+            target_dtype,
+            eps=eps,
+            zero_point_dtype=zero_point_dtype,
+        )
+    else:
+        assert config.version == 2, f"Unexpected version: {config.version}"
+        block_size = [weight.shape[0], weight.shape[1]]
```
jerryzh168, line 1390: this should be the same as L1393 I think, you can extract L1390-L1393 out of the first if branch and use that I think

torchao/quantization/quant_api.py
```diff
@@ -1572,19 +1585,31 @@ def get_weight_block_size(x):
...
+            _layout=layout,
+            zero_point_domain=weight_zero_point_domain,
+        )
+        quantized_weight = to_linear_activation_quantized(
+            quantized_weight, input_quant_func
+        )
+    else:
+        quantized_weight = Int8Tensor.from_hp(
+            weight,
+            block_size=get_weight_block_size(weight),
```
jerryzh168: nit: can calculate `block_size` outside of the if/else

torchao/quantization/quantize_/common/quantize_tensor_kwargs.py
```diff
@@ -52,5 +54,10 @@ def _choose_quant_func_and_quantize_tensor(
             quant_kwargs.hp_value_ub,
             quant_kwargs.kernel_preference,
         )
+    elif isinstance(quant_kwargs, QuantizeTensorToInt8Kwargs):
+        return Int8Tensor.from_hp(
+            tensor,
+            quant_kwargs.block_size or [1, tensor.shape[-1]],
```
jerryzh168: nit: why not make block_size mandatory?

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,287 @@
...
+
+@dataclass
+class QuantizeTensorToInt8Kwargs(QuantizeTensorKwargs):
+    """Tensor kwargs for creating int8 tensor (either activation or weight)
+
+    Args:
+        block_size (Optional[list[int]]): block size for quantization granularity
+    """
+
+    block_size: Optional[list[int]] = None
```
jerryzh168: why is this optional?

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,287 @@
...
+    """
+
+    tensor_data_names = ["qdata", "scale"]
+    tensor_attribute_names = ["block_size", "_shape"]
+    optional_tensor_attribute_names = [
+        "act_quant_kwargs",
+        "dtype",
+    ]
+
+    def __new__(
```
jerryzh168, line 64: nit: please annotate the args with types to be clearer

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,287 @@
...
+        dtype=None,
+    ):
+        kwargs = {
+            "device": qdata.device,
+            "dtype": dtype or scale.dtype,
+            "requires_grad": False,
+        }
+        return torch.Tensor._make_wrapper_subclass(cls, shape, **kwargs)
+
+    def __init__(
```
jerryzh168, line 79: same here

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,287 @@
...
+        block_size,
+        shape,
+        act_quant_kwargs=None,
+        dtype=None,
+    ):
+        super().__init__()
+        self.qdata = qdata
+        self.scale = scale
+        self.block_size = block_size
+        self._shape = shape
```
jerryzh168: nit: we don't need to set shape here, since it will be set in `torch.Tensor._make_wrapper_subclass`

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,287 @@
...
+    def dequantize(self, output_dtype: Optional[torch.dtype] = None) -> torch.Tensor:
+        """Dequantize int8 tensor to floating point"""
+        dtype = output_dtype or self.dtype or self.scale.dtype
+
+        qdata_fp = self.qdata.to(dtype)
+        scale = self.scale.to(dtype)
+
+        # Reshape 1D scale to [N, 1] for broadcasting with [N, K] qdata
+        if scale.ndim == 1:
+            scale = scale.unsqueeze(1)
```
jerryzh168: is this needed?

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,287 @@
...
+            sliced_scale,
+            tensor.block_size,
+            sliced_shape,
+            tensor.act_quant_kwargs,
+            tensor.dtype,
+        ),
+    )
+
+
+@implements(aten.transpose.int)
```
jerryzh168: we don't need this yet I think, we can remove for now and add later when needed

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,287 @@
...
+        args[2],
+        args[3],
+        args[4] if len(args) > 4 else 1,
+    )
+
+    sliced_scale = tensor.scale
+    if dim == 0 and tensor.scale.ndim >= 1:
+        sliced_scale = aten.slice.Tensor(tensor.scale, 0, start, end, step)
+
+    sliced_shape = list(
```
jerryzh168: why not get the shape from sliced tensor directly?

jerryzh168: can you check https://github.com/pytorch/ao/blob/c96f2dd0c980ad100ea9e7c574dead8f19451955/torchao/quantization/quantize_/workflows/float8/float8_tensor.py#L419? I'm not sure if the current implementation is enough to cover all cases actually

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
namgyu-youn: Isn't we already comparing 1) bflot16 vs. int8-quant, 2) float16 vs. int8-quant by dtype parameterization?

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
namgyu-youn: Could you tell me why there is no need to support transposition for quantized tensors? I thought it was just a type of tensor. If we remove this, how can users transpose it like `Tensor.transpose()` ?

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
namgyu-youn: It was wrong type hint because api can't work without granularity, mandatory (not-optional) should be right.

torchao/quantization/quant_api.py
namgyu-youn: Isn't dividing logics much safer and easier to deprecate old API in the future? Other APIs like `_float8_weight_only_quant_tensor` also have been used with this convention without a common branch.

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,319 @@
...
+class QuantizeTensorToInt8Kwargs(QuantizeTensorKwargs):
+    """Tensor kwargs for creating int8 tensor (either activation or weight)
+
+    Args:
+        block_size (list[int]): block size for quantization granularity
+        static_scale (Optional[torch.Tensor]): pre-computed scale for static quantization
+    """
+
+    block_size: list[int]
+    static_scale: Optional[torch.Tensor] = None
```
jerryzh168, line 41: I haven't seen static quant use cases before, I'd recommend to remove for now

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,319 @@
...
+implements = Int8Tensor.implements
+
+
+@implements([aten.dequantize.self])
+def _(func, types, args, kwargs):
+    """dequantization: int8 -> float"""
+    return args[0].dequantize()
+
+
+@implements([torch.nn.functional.linear, aten.linear.default])
```
jerryzh168: nit: implements is refactored now: https://github.com/pytorch/ao/pull/2866/files

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
jerryzh168: just haven't seen people using it yet, I think we should implement as little as possible to keep maintainence burden low

torchao/quantization/quant_api.py
jerryzh168: it's fine to duplicate I think, but the current code for block_size doesn't support 3d though

torchao/quantization/quant_api.py
```diff
@@ -1576,18 +1589,30 @@ def get_weight_block_size(x):
...
+            eps=eps,
+            zero_point_dtype=zero_point_dtype,
+            _layout=layout,
+            zero_point_domain=weight_zero_point_domain,
+        )
+        quantized_weight = to_linear_activation_quantized(
+            quantized_weight, input_quant_func
+        )
+    else:
+        quantized_weight = Int8Tensor.from_hp(
```
jerryzh168, line 1602: this doesn't include activation quantization?

jerryzh168, line 1602: also we should add an assert for version here as well, same as the other place

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
namgyu-youn: I barely remember there are only a few use cases in other issues, but it seems https://github.com/pytorch/ao/pull/3089 is a kind of use case. Could you check it again?

jerryzh168: oh I see, OK then we can add this now

torchao/quantization/quant_api.py
namgyu-youn: Okay then I will keep this branch and update the assert for 3D check.

torchao/utils.py
```diff
@@ -433,74 +433,62 @@ def __eq__(self, other):
...
-    if not hasattr(cls, "_ATEN_OP_OR_TORCH_FN_TABLE"):
-        cls._ATEN_OP_OR_TORCH_FN_TABLE = {}
+    return decorator
 
-    if cls not in cls._ATEN_OP_OR_TORCH_FN_TABLE:
-        cls._ATEN_OP_OR_TORCH_FN_TABLE[cls] = {}
 
-    if not isinstance(aten_ops_or_torch_fns, (list, tuple)):
-        aten_ops_or_torch_fns = [aten_ops_or_torch_fns]
+def _implements_torch_function(cls, torch_fns):
```
jerryzh168: why these changes? is there some issue with rebase?

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,304 @@
...
+
+    assert isinstance(weight_tensor, Int8Tensor), (
+        f"Expected weight to be Int8Tensor, got {type(weight_tensor)}"
+    )
+
+    if weight_tensor.act_quant_kwargs is not None:
+        if not isinstance(activation_tensor, Int8Tensor):
+            if weight_tensor.act_quant_kwargs.static_scale is not None:
+                # INT8 × INT8 (static): symmetric quantization only
+                static_scale = weight_tensor.act_quant_kwargs.static_scale
```
jerryzh168: OK if this is needed I think it should be included in `_choose_quant_func_and_quantize_tensor` as well?

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,304 @@
...
+            scale = scale.view(*scale.shape, *[1] * (qdata_fp.ndim - scale.ndim))
+
+        return qdata_fp * scale
+
+
+implements = Int8Tensor.implements
+implements_torch_function = Int8Tensor.implements_torch_function
+
+
+@implements([aten.dequantize.self])
```
jerryzh168: is this needed? if not we should remove for now

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,304 @@
...
+    def dequantize(self, output_dtype: Optional[torch.dtype] = None) -> torch.Tensor:
+        """Dequantize int8 tensor to floating point"""
+        dtype = output_dtype or self.dtype or self.scale.dtype
+
+        qdata_fp = self.qdata.to(dtype)
+        scale = self.scale.to(dtype)
+
+        # Reshape scale to broadcast
+        if scale.numel() > 1 and scale.shape != qdata_fp.shape:
+            scale = scale.view(*scale.shape, *[1] * (qdata_fp.ndim - scale.ndim))
```
jerryzh168: is this needed?

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,304 @@
...
+    optional_tensor_attribute_names = [
+        "act_quant_kwargs",
+        "dtype",
+    ]
+
+    def __new__(
+        cls: type,
+        qdata: torch.Tensor,
+        scale: torch.Tensor,
+        block_size: list[int],
```
jerryzh168, line 68: nit: I remember `list` has a higher python version requirements, so probably better to change this to `List` from `typing` I think

torchao/quantization/quantize_/common/quantize_tensor_kwargs.py
jerryzh168: this one is still not resolved yet

torchao/quantization/quant_api.py
```diff
@@ -2443,6 +2406,21 @@ def _module_fqn_to_config_handler(
     return module
 
 
+def _unwrap_float8_linear(module: Float8Linear) -> nn.Linear:
```
jerryzh168, line 2480: some rebase issue?

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
jerryzh168: yeah this test should be removed

torchao/quantization/quant_api.py
namgyu-youn: Oh actually we already doing 3D-check at `from_hp()`, by using

```python
if w.dim() != 2 or len(block_size) != 2:
    raise ValueError("Expected 2D tensor and block_size length 2")
```

torchao/utils.py
namgyu-youn: There was no merge conflict, so I overwrote this file, but I regret this commit; let me know if https://github.com/pytorch/ao/pull/3038/commits/0a45f902e57659c3dfae0d6bf6895954e49aac92 should be reverted.

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
namgyu-youn: It is needed for block-level granularity. For example,

1. Row-wise: If scale shape is (64, 1) and w_q (quantized weight shape) is (256, 512), we can naturally broadcast them
2. Channel-wise: If scale shape is (512,) and w_q is (256, 512), we can naturally broadcast them
3. Block-size granularity: If scale shape is (32, 64) and w_q is (256, 512), we have to rescale to broadcast them.

But we can also reuse `_maybe_expand_scale_to_tensor_shape`, similar to:

https://github.com/pytorch/ao/blob/4b79f9ec5d6951ebdbcbdc70da529861618b8d8b/torchao/quantization/quantize_/workflows/float8/float8_tensor.py#L149-L154

and 

https://github.com/pytorch/ao/blob/f3fc5e75c4225917480210df69eb1a09edc63292/torchao/quantization/quant_primitives.py#L2407-L2421

namgyu-youn:
Updated log:
- Made `block_size` mandatory (resolves https://github.com/pytorch/ao/pull/3038#discussion_r2437348445, https://github.com/pytorch/ao/pull/3038#discussion_r2407653212)
- Fix static quantization flows (resolves: https://github.com/pytorch/ao/pull/3038#discussion_r2437343855)
- Update `block_size` 3D-check logic (related to: https://github.com/pytorch/ao/pull/3038#discussion_r2407645952)
- Replace type hint `List()` with `list()`: https://github.com/pytorch/ao/pull/3038#discussion_r2437347794

~To reviewers: 
Unfortunately, I can't build and run local tests, caused by https://github.com/pytorch/ao/issues/2919, after trying downgrade and gradual installation. Please feel free to direct commit if `test_int8_tensor.py` fails.~

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
namgyu-youn: Thanks, it is only for `List`, not for `Dict`, `Tuple`, etc.?

torchao/utils.py
```diff
@@ -12,7 +12,7 @@
 from functools import reduce
 from importlib.metadata import version
 from math import gcd
-from typing import Any, Callable, Optional, Type
+from typing import Any, Callable, Optional
```
jerryzh168: please fix rebase to not have these changes, or open a new PR if you don't know how to fix rebase

torchao/utils.py
jerryzh168: yeah I don't think it's relevant, so the changes should be reverted, but this looks like some issue related to rebase

torchao/quantization/quant_api.py
namgyu-youn: Sorry, I assumed https://github.com/pytorch/ao/pull/3038/commits/0a45f902e57659c3dfae0d6bf6895954e49aac92 was a wrong way, which was the start of the rebase issue. The solution looks like dropping relevant commits using rebase.

But rebasing after all those commits is overwhelming to me. So, I really don't want to open a duplicate PR, but may I reopen the PR and link to this PR? I just want to remove unrelevant codes change log.

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -0,0 +1,221 @@
...
+        self.assertEqual(tensor.qdata.dtype, torch.int8)
+        self.assertTrue(
+            torch.all(tensor.qdata >= -128) and torch.all(tensor.qdata <= 127)
+        )
+
+    @common_utils.parametrize("dtype", [torch.bfloat16, torch.float16])
+    @common_utils.parametrize(
+        "sizes",
+        [
+            ((128,), 256, 128),
```
jerryzh168, line 89: does 3D inputs work? e.g. ((32, 128,), 256, 128),

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -0,0 +1,221 @@
...
+        quantize_(m_q, config)
+
+        output_original = m(input_tensor)
+        output_quantized = m_q(input_tensor)
+
+        error = compute_error(output_original, output_quantized)
+        assert error > 20, f"Quantization error is too high got a SQNR of {error}"
+
+    @common_utils.parametrize("dtype", [torch.bfloat16, torch.float16])
+    def test_static_quantization(self, dtype):
```
jerryzh168: can you add a test the static quant config and test that as well `Int8StaticActivationInt8WeightConfig`

or maybe you can remove this for now and coordinate with @Xia-Weiwen (https://github.com/pytorch/ao/pull/3089/files#diff-bf4d50867e3d649de2d89146592bf47d2f258c4c19126c8acf0e120ee904b726) to add the static quant support separately?

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -0,0 +1,221 @@
...
+        param_data = param.data
+        param_data = param_data.narrow(0, 0, 512)
+        # Making sure the aliasing is preserved in sliced quantized Tensor
+        assert param.data.qdata.data_ptr() == param_data.qdata.data_ptr()
+        assert param.data.scale.data_ptr() == param_data.scale.data_ptr()
+
+    @common_utils.parametrize(
+        "config",
+        [
+            Int8DynamicActivationInt8WeightConfig(version=2),
```
jerryzh168, line 189: will need to test the static quant as well, if that is added

torchao/quantization/quant_api.py
jerryzh168, line 1: changes in this PR should be reverted as well

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,281 @@
...
+            sliced_qdata,
+            sliced_scale,
+            block_size,
+            tensor.act_quant_kwargs,
+            tensor.dtype,
+        ),
+    )
+
+
+@implements(aten.select.int)
```
jerryzh168, line 278: is this tested? if not, please remove for now

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,281 @@
...
+    sliced_qdata = func(tensor.qdata, dim, start, end, step)
+
+    if tensor.scale.numel() == 1:
+        # Per-tensor quantization - scale doesn't change
+        sliced_scale = tensor.scale
+    elif dim < tensor.scale.ndim and tensor.scale.shape[dim] > 1:
+        # Block-wise quantization - need to slice the scale appropriately
+        sliced_scale = func(tensor.scale, dim, start, end, step)
+    else:
+        sliced_scale = tensor.scale
```
jerryzh168: can you match the implementation with Float8Tensor? https://github.com/pytorch/ao/blob/53b5efdac921a38fd15e8d3ac8191c3927140287/torchao/quantization/quantize_/workflows/float8/float8_tensor.py#L449-L456

torchao/quantization/quant_api.py
jerryzh168: @namgyu-youn please feel free to close and reopen a new one if it's hard to fix rebase issue, seems like it's still not fully fixed

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
jerryzh168: probably also for Dict and Tuple, I have only tried list before

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
Xia-Weiwen: Both are ok for me. Thanks.

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
namgyu-youn: No, 3D input raise `ValueError` triggered by `from_hp()`: https://github.com/pytorch/ao/pull/3038#discussion_r2438473717

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
namgyu-youn: Interested in both way, but smaller PR should be better I feel, and I will remove static flag.

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
namgyu-youn: Is there a use case with old (< 3.9) python version? I remember `list`, `dict`, and `tuple` is natively supported to 3.10 https://docs.astral.sh/ruff/rules/non-pep585-annotation/.

Because PEP585 (type hint) is chained with pre-commit issue, prefer to focus on new versions (no need for `from typing import List, Dict, Tuple`). How about using `list, dict, tuple` focusing on new python versions? There might be a new issue by PEP 585 if we go with `Dict, List, Tuple` I feel.

torchao/quantization/quant_api.py
namgyu-youn: Thanks, just want to remove unrelevant change logs as you mentioned.

namgyu-youn:
Updated logs:
In https://github.com/pytorch/ao/pull/3038/commits/062f3ccae219169a2501588e7195d119fbbb73ff
- Add static/dynamic unit test: https://github.com/pytorch/ao/pull/3038#discussion_r2457609875
- Remove relevant dequantization function: https://github.com/pytorch/ao/pull/3038#discussion_r2437344941
- Revert `List -> list`: https://github.com/pytorch/ao/pull/3038#discussion_r2458616350
- Update api to support per-row and per-tensor granularity: related to https://github.com/pytorch/ao/pull/3038#discussion_r2380385907
- Update tensor slicing operation similar to `Float8Tensor`: https://github.com/pytorch/ao/pull/3038#discussion_r2457626291

In https://github.com/pytorch/ao/pull/3038/commits/680cec98af3eb59e3d6d2a73f5b9d0de94efd3b5
- Update `setUp` for common used args
- Split `test_error_handling_and_dequant` unit test into `test_invalid_input_handling & test_dequantization_accuracy`

namgyu-youn:
@jerryzh168 sorry for the multiple-PRs again; reopened PR (https://github.com/pytorch/ao/pull/3241) is copy-pasted after the last commit in this PR and resolves rebase errors. Please check the above comment (change log; https://github.com/pytorch/ao/pull/3038#issuecomment-3442839212) first, and please take a look at https://github.com/pytorch/ao/pull/3241, thanks.

PR #3042, opened 2025-09-22
generalize torch compatibility check

liangel-02:
**Summary**

per #2919, torchao .so files built using PyTorch 2.8.0 are not ABI compatible with PyTorch 2.9+. this PR modifies the check added in #2908 for future versions of torchao. we keep the checks for current stable versions for clarity and add a check for most recent versions (torch 2.9.0.dev and torchao 0.14.0dev). this check should be updated with every release to correctly reflect compatibility.

**Testing**

bash script `./scripts/test_torch_version_torchao_version_compatibility.sh`. this script was ran locally to test and should be updated and run before every release/version bump.


torchao/utils.py
```diff
@@ -378,6 +378,14 @@ def torch_version_at_least(min_version):
     return parse_version(torch.__version__) >= parse_version(min_version)
 
 
+def torch_version_greater_than(min_version):
```
andrewor14: I think we can just reuse at_least? Are we concerned about the ABI breaking change between 2.8.1 vs 2.8.0?

.github/scripts/validate_binaries.sh
```diff
@@ -4,3 +4,52 @@ python  ./test/smoke_tests/smoke_tests.py
 # Now we install dev-requirments and try to run the tests
 pip install -r dev-requirements.txt
 pytest test --verbose -s
+
+# Testing compatibility
+# We know that torchao .so files built using PyTorch 2.8.0 are not ABI compatible with PyTorch 2.9+.
```
andrewor14: Link to the issue?

.github/scripts/validate_binaries.sh
```diff
@@ -4,3 +4,52 @@ python  ./test/smoke_tests/smoke_tests.py
...
+        exit 1
+    fi
+}
+
+# Uninstall torch
+pip uninstall torch
+# Install compatible version of torch
+pip install torch==2.8.0
+# Build torchao
+pip install --pre torchao --index-url https://download.pytorch.org/whl/nightly/cu129
```
andrewor14: Maybe I'm missing something but this isn't actually building torchao right, just pulling the pre-built binary? Like in the future when the latest stable torch version is 2.10.0 or something, we still want to test building torchao against 2.8.0 right?

torchao/__init__.py
```diff
@@ -28,14 +30,11 @@
 # otherwise, assume it's prebuilt by torchao's build scripts and we can make
 #   assumptions about the PyTorch version used to build it.
 if (not "+git" in __version__) and not ("unknown" in __version__):
-    # torchao v0.13.0 is built with PyTorch 2.8.0. We know that torchao .so
-    # files built using PyTorch 2.8.0 are not ABI compatible with PyTorch 2.9+.
+    # We know that torchao .so files built using PyTorch 2.8.0 are not ABI compatible with PyTorch 2.9+.
     # The following code skips importing the .so files if PyTorch 2.9+ is
     # detected, to avoid crashing the Python process with "Aborted (core
     # dumped)".
-    # TODO(#2901, and before next torchao release): make this generic for
```
andrewor14, line 36: I think we should keep the link to the issue but remove the TODO

torchao/utils.py
andrewor14: discussed offline, this is to guard against nightlies like 2.9.0.dev (which is not at_least 2.9.0)

jerryzh168: is `at_least` (>=) better than `greater_than` (>)?

liangel-02: `at_least` will return False for torch nightlies (e.g. 2.9.0.dev...) because the `parse_version` function will evaluate patch to -1 for pre-release

liangel-02: https://github.com/pytorch/ao/pull/3042#discussion_r2370018264

will make this function priv

torchao/__init__.py
```diff
@@ -5,6 +5,8 @@
 
 import torch
 
+from torchao.utils import torch_version_greater_than
```
vkuzo: I'd advice against importing anything general here, the check for "can we use c++ kernels" should not depend on any internal library code.  Just copy-paste what you need, or if you do import then you need to make sure the file being imported has no chance of doing anything funky.

torchao/__init__.py
```diff
@@ -28,14 +30,11 @@
...
-    # torchao v0.13.0 is built with PyTorch 2.8.0. We know that torchao .so
-    # files built using PyTorch 2.8.0 are not ABI compatible with PyTorch 2.9+.
+    # We know that torchao .so files built using PyTorch 2.8.0 are not ABI compatible with PyTorch 2.9+.
     # The following code skips importing the .so files if PyTorch 2.9+ is
     # detected, to avoid crashing the Python process with "Aborted (core
     # dumped)".
-    # TODO(#2901, and before next torchao release): make this generic for
-    # future torchao and torch versions
-    if __version__.startswith("0.13.0") and str(torch.__version__) >= "2.9":
+    if torch_version_greater_than("2.8.0"):
```
vkuzo: this doesn't look right, you want something like:

```
if torch == 2.8.0 and torchao == 0.13.0:
    good
elif torch == 2.9.0 and torchao == 0.14.0:
    good
elif ...
    good
else:
    bad
```

vkuzo: https://github.com/pytorch/ao/issues/2919

torchao/__init__.py
```diff
@@ -20,24 +21,50 @@
 except PackageNotFoundError:
     __version__ = "unknown"  # In case this logic breaks don't break the build
 
+
 logger = logging.getLogger(__name__)
 
+
+def parse_version(version_string):
```
andrewor14: this should be private

torchao/__init__.py
```diff
@@ -20,24 +21,50 @@
...
+    """
+    Parses the major and minor of a torch version
+
+    Examples:
+        - 2.9.0dev... (pre-release) becomes [2, 9]
+        - 2.8.0 becomes [2, 8]
+        - 2.8.1 becomes [2, 8]
+    """
+    # Check for pre-release indicators
```
andrewor14, line 35: we can remove this comment?

.github/scripts/validate_binaries.sh
```diff
@@ -1,6 +1,67 @@
...
-python  ./test/smoke_tests/smoke_tests.py
-# Now we install dev-requirments and try to run the tests
-pip install -r dev-requirements.txt
-pytest test --verbose -s
+# pip install ${PYTORCH_PIP_PREFIX} torchao --index-url ${PYTORCH_PIP_DOWNLOAD_URL}
+# # Intial smoke test, tries importing torchao
+# python  ./test/smoke_tests/smoke_tests.py
+# # Now we install dev-requirments and try to run the tests
+# pip install -r dev-requirements.txt
+# pytest test --verbose -s
```
andrewor14: uncomment these?

.github/scripts/validate_binaries.sh
```diff
@@ -1,6 +1,67 @@
...
+check_torchao_import "false" ""
+
+# Uninstall torch
+pip uninstall torch
+# Uninstall torchao
+pip uninstall torchao
+# Install non-ABI stable torch version with older version of torchao
+pip install --pre torch --index-url https://download.pytorch.org/whl/nightly/cu129
+# Build torchao
+pip install torchao==0.13.0
```
andrewor14: I think this is fine, but the comment is wrong. This doesn't actually build torchao, but rather installs a version of torchao (0.13.0) that was built against 2.8.0

.github/scripts/validate_binaries.sh
```diff
@@ -1,6 +1,67 @@
...
+}
+
+# Uninstall torch
+pip uninstall torch
+# Install compatible version of torch
+pip install torch==2.8.0
+# Build torchao
+pip install torchao==0.13.0
+# Test 1: Should import successfully without warning
+check_torchao_import "false" ""
```
andrewor14: probably don't need to test 2.8.0 vs 0.13.0 compatibility for every future release from now on

.github/scripts/validate_binaries.sh
```diff
@@ -1,6 +1,67 @@
...
+
+# Uninstall torch
+pip uninstall torch
+# Uninstall torchao
+pip uninstall torchao
+# Install compatible version of torch (nightly 2.9.0dev)
+pip install --pre torch --index-url https://download.pytorch.org/whl/nightly/cu129
+# Build torchao (nightly 0.14.0dev...)
+pip install --pre torchao --index-url https://download.pytorch.org/whl/nightly/cu129
```
andrewor14: Will these two always be true? E.g. let's say we release 0.14.0 and bump the version, now it's 0.15.0dev vs 2.9.0dev, then this will fail right? Does this mean we need to keep updating `__init__.py` to add a new check every time we bump the torchao version?

andrewor14 (approved):
Looks good to me, thanks! Just left a few minor questions

torchao/__init__.py
```diff
@@ -20,24 +21,49 @@
...
 
+
+def _parse_version(version_string):
+    """
+    Parses the major and minor of a torch version
+
+    Examples:
+        - 2.9.0dev... (pre-release) becomes [2, 9]
+        - 2.8.0 becomes [2, 8]
+        - 2.8.1 becomes [2, 8]
```
vkuzo: 2.8.1 has a different ABI from 2.8.0, so this function should return three things (2, 8, 0) instead of two (2, 8)

.github/scripts/validate_binaries.sh
```diff
@@ -4,3 +4,64 @@ python  ./test/smoke_tests/smoke_tests.py
 # Now we install dev-requirments and try to run the tests
 pip install -r dev-requirements.txt
 pytest test --verbose -s
+
+# Testing compatibility
```
vkuzo: can this be a separate script?

torchao/__init__.py
```diff
@@ -20,24 +21,47 @@
...
+    # The following code skips importing the .so files if incompatible torch version is detected,
+    # to avoid crashing the Python process with "Aborted (core
     # dumped)".
-    # TODO(#2901, and before next torchao release): make this generic for
-    # future torchao and torch versions
-    if __version__.startswith("0.13.0") and str(torch.__version__) >= "2.9":
+    # .dev for nightlies since 2.9.0 has not been released
+    if _parse_version(torch.__version__) == _parse_version(
+        "2.9.0.dev"
+    ) and _parse_version(__version__) == _parse_version("0.14.0"):
```
jerryzh168: why is this only torch 2.9.0 nightly and torchao == 0.14.0 instead of something like torch >= 2.8 and torchao >= 0.13.0?

.github/scripts/test_compatibility.sh
```diff
@@ -0,0 +1,51 @@
...
+}
+
+# Uninstall torch
+pip uninstall torch
+# Uninstall torchao
+pip uninstall torchao
+# Install compatible version of torch (nightly 2.9.0dev)
+pip install --pre torch --index-url https://download.pytorch.org/whl/nightly/cu129
+# Installs nightly torchao (0.14.0dev...)
+pip install --pre torchao --index-url https://download.pytorch.org/whl/nightly/cu129
```
jerryzh168: should we test more combinations? e.g.

torchao = (0.13, 0.14.dev) 
pytorch = (2.8, 2.9.dev, 2.9 (future))

torchao/__init__.py
liangel-02: we want to do stricter checks like == instead of >=
if we did torch >= x and torchao >= y we can have false positives if torchao version is y+2 but torch is x+1 when y+2 is built against torch x+2 (so not guaranteed compatible with torch x+1)

we also only need to check the latest torchao version against the latest torch version because this code will never run for a previous torchao version (it is redundant to consider the case when torchao == 0.13.0 because if that were the case, this code would not exist)

.github/scripts/test_compatibility.sh
liangel-02: adding more test, but don't need to test previous versions i think (see https://github.com/pytorch/ao/pull/3042#discussion_r2376589479, https://github.com/pytorch/ao/pull/3042#discussion_r2383236934)

torchao/__init__.py
jerryzh168: > if we did torch >= x and torchao >= y we can have false positives if torchao version is y+2 but torch is x+1 when y+2 is built against torch x+2 (so not guaranteed compatible with torch x+1)

I think changing to `>=` for both would skip all future inclusions of this? e.g. torchao >= 0.14.0.dev and torch >= 2.9.0.dev (this will be satisfied for all current and future torch and torchao versions)

explanation for 0.13.0 and below makes sense

jerryzh168: also to clarify, current version for torchao should be `0.14.0.dev` I think, not sure if you intended to check 0.14.0 or not

torchao/__init__.py
```diff
@@ -20,24 +22,54 @@
...
+        if is_prerelease:
+            patch = -1
+        return [major, minor, patch]
+    else:
+        raise ValueError(f"Invalid version string format: {version_string}")
+
+
+# users can set env var TORCH_INCOMPATIBLE=1 to skip loading .so files
+# this way, if they are using an incompatbile torch version, they can still use the API by setting the env var
+skip_loading_so_files = bool(os.getenv("TORCH_INCOMPATIBLE", False))
```
vkuzo: nit: name the env var something like `TORCHAO_SKIP_LOADING_SO_FILES` to be very specific about what it is doing.

torchao/__init__.py
```diff
@@ -20,24 +22,54 @@
...
+    # The following code skips importing the .so files if incompatible torch version is detected,
+    # to avoid crashing the Python process with "Aborted (core
     # dumped)".
-    # TODO(#2901, and before next torchao release): make this generic for
-    # future torchao and torch versions
-    if __version__.startswith("0.13.0") and str(torch.__version__) >= "2.9":
+    # .dev for nightlies since 2.9.0 has not been released
+    if _parse_version(torch.__version__) == _parse_version(
+        "2.9.0.dev"
+    ) and _parse_version(__version__) == _parse_version("0.14.0"):
```
vkuzo: let's rewrite this to be more obvious:

```python
skip_loading_so_files = False

if bool(os.getenv("TORCH_INCOMPATIBLE", False)):
    # user overrise
    skip_loading_so_files = True
elif (version check here):
    # incompatible lib versions
    skip_loading_so_files = True
...
if skip_loading_so_files:
    ... emit warning ...
else:
    ... load so files ...
```

vkuzo (changes requested):
comments inline

torchao/__init__.py
vkuzo: I think you should spell out all the versions here:

```python
if torch_version == X and torchao_version == Y:
  ...
elif torch_version == X+1 and torchao_version == Y+1:
  ...
else:
  ...
```

this removes all confusion about what is supported or not.  It's ok that `if torch_version == X and torchao_version == Y` will become dead code, IMO more important to make it crystal clear what's going on.  IMO having the entire support matrix here is the best way to do that.

If you really don't want dead code, I'd still have a comment very clearly explaining why only the latest thing is here and what the behavior was before.

scripts/test_torch_version_torchao_version_compatibility.sh
```diff
@@ -0,0 +1,68 @@
+# Testing compatibility
```
vkuzo, line 1: nit: rename to `test_torch_version_torchao_version_compatibility.sh` to be specific

.github/scripts/test_compatibility.sh
```diff
@@ -0,0 +1,68 @@
...
+        exit 1
+    fi
+}
+
+# Uninstall torch
+pip uninstall torch
+# Uninstall torchao
+pip uninstall torchao
+# Install compatible version of torch (nightly 2.9.0dev)
+pip install --pre torch --index-url https://download.pytorch.org/whl/nightly/cu129
```
vkuzo: can we install a specific version here to be clear about what we are testing

scripts/test_torch_version_torchao_version_compatibility.sh
```diff
@@ -0,0 +1,95 @@
...
```
vkuzo, line 1: nit: remove double ".sh" from file name

.github/workflows/validate-binaries.yml
```diff
@@ -45,5 +45,5 @@ jobs:
       repository: "pytorch/ao"
       with_cuda: "enable"
       with_rocm: "disable"
-      smoke_test: "source ./.github/scripts/validate_binaries.sh"
+      smoke_test: "source ./.github/scripts/validate_binaries.sh && source ./.github/scripts/test_compatibility.sh"
```
vkuzo: point to new name?

torchao/__init__.py
```diff
@@ -20,28 +22,63 @@
...
+    # The following code skips importing the .so files if incompatible torch version is detected,
+    # to avoid crashing the Python process with "Aborted (core
     # dumped)".
-    # TODO(#2901, and before next torchao release): make this generic for
-    # future torchao and torch versions
-    if __version__.startswith("0.13.0") and str(torch.__version__) >= "2.9":
-        logger.warning(
-            f"Skipping import of cpp extensions due to incompatible torch version {torch.__version__} for torchao version {__version__}"
-        )
+    if _parse_version(torch.__version__) == _parse_version("2.8.0") and _parse_version(
```
vkuzo: nit:

```python
torch_cur_version = _parse_version(torch.__version__)
torchao_cur_version = _parse_version(__version__)
v2_8_0 = _parse_version('2.8.0')
v0_13_0 = _parse_version('0.13.0')

if torch_cur_version == v2_8_0 and torchao_cur_version == v0_13_0:
    ...
elif ...:
    ...
```

.github/scripts/test_torch_version_torchao_version_compatibility.sh
```diff
@@ -0,0 +1,95 @@
...
+    fi
+}
+
+## torch == 2.8.0, torchao == 0.13.0
+# Uninstall torch
+pip uninstall torch
+# Uninstall torchao
+pip uninstall torchao
+# Install current compatible version of torch (2.8.0)
+pip install torch==2.8.0
```
vkuzo: can we extract the versions out in variables and put them at the top of the script? That would make it easy for the person doing the torchao release to change them when the version increments

vkuzo:
```bash
# previous versions
PREV_TORCH_VERSION = 2.8.0
PREV_TORCHAO_VERSION = 0.13.0

# commands to install current torch version
pip install --pre torch --index-url https://download.pytorch.org/whl/nightly/cu129
```

IMO don't test torchao nightly because that should be the same as building it from source for the purposes of this script

scripts/test_torch_version_torchao_version_compatibility.sh
```diff
@@ -0,0 +1,95 @@
...
+## torch == 2.9.0.dev..., torchao from source, env var = True
+# Uninstall torch
+pip uninstall torch
+# Uninstall torchao
+pip uninstall torchao
+# Install compatible version of torch (nightly 2.9.0dev)
+pip install torch==2.9.0.dev20250905+cu129 --index-url https://download.pytorch.org/whl/nightly/cu129
+# Build torchao from source
+python setup.py develop
+# Should import with warning because optional env var is set to true
```
vkuzo, line 68: I think this test should
1. install torch nightly
2. build torchao from source
3. uninstall torch
4. install torch stable
5. import with env var

this is testing the import which will crash if we don't use the env var

.github/workflows/validate-binaries.yml
```diff
@@ -45,5 +45,5 @@ jobs:
       repository: "pytorch/ao"
       with_cuda: "enable"
       with_rocm: "disable"
-      smoke_test: "source ./.github/scripts/validate_binaries.sh"
+      smoke_test: "source ./.github/scripts/validate_binaries.sh && source ./.github/scripts/test_torch_version_torchao_version_compatibility.sh"
```
vkuzo: thoughts on having the release owner run this manually and removing it from here?  It seems like having this in CI could run into edge cases when a new torch version is released but a new torchao version is not released on the same day?

torchao/__init__.py
```diff
@@ -20,28 +22,67 @@
...
-    # torchao v0.13.0 is built with PyTorch 2.8.0. We know that torchao .so
-    # files built using PyTorch 2.8.0 are not ABI compatible with PyTorch 2.9+.
-    # The following code skips importing the .so files if PyTorch 2.9+ is
-    # detected, to avoid crashing the Python process with "Aborted (core
+if bool(os.getenv("TORCHAO_SKIP_LOADING_SO_FILES", False)):
+    # user override
+    # users can set env var TORCH_INCOMPATIBLE=1 to skip loading .so files
+    # this way, if they are using an incompatbile torch version, they can still use the API by setting the env var
+    skip_loading_so_files = True
+    # if torchao version has "+git", assume it's locally built and we don't know
```
vkuzo: align comment indent to L57?

vkuzo (approved):
thank you for improving this!

scripts/test_torch_version_torchao_version_compatibility.sh
```diff
@@ -0,0 +1,82 @@
+# Testing compatibility
```
vkuzo, line 1: since this script is not run from CI anymore, can we move it to a different location, for example `ao/scripts/`?

PR #3048, opened 2025-09-23
[mxfp8 moe training] mxfp8_on_device_all_to_all_v kernel

danielvegamyhre:
## Summary
- Trace analysis of Llama4 mxfp8 training on b200 (FSDP=4, EP=4, seq_len=8192, local_batch_size=8) shows exposed a2a comms constitute a huge % of total runtime (see trace). 
- We can improve this by doing dynamic quant on inputs -> a2a on mxfp8 data and scales -> dequantize outputs, which should reduce bytes sent over the network by ~45%. There will be some overhead from dynamic quant/dequant, but this is on the order of microseconds, and the a2a here is 200+ms, so should be a good trade-off and solid net perf benefit.
- My main concern here is how I'm allocating symmetric memory buffers and necessary rendezvous overhead - the idea is for this to only happen once on the first forward pass then be re-used, but remains to be seen if this works as I expect in practice. 
Screenshot 2025-09-23 at 12 58 12 PM

## Changes
Using the [bf16 all_to_all_v](https://github.com/pytorch/torchtitan/blob/3e1b843ecd91a80dbf56e16f2eae7b637209ebcb/torchtitan/experiments/deepseek_v3/symm_mem_recipes/triton_on_device_all_to_all_v.py) kernel from torchtitan as a starting point, this PR implements:
- `_mxfp8_on_device_all_to_all_v` kernel using Triton + Symmetric Memory for inter-device comms
- Autograd func wrapper to make it differentiable
- Unit tests verifying correctness of both forward and backward pass, by comparing against standard fp32 `all_to_all_single_autograd` and validating SQNR is high (>28).
- I also refactored to separate quantization kernels and comms kernels into separate files/modules.

## Benchmarks
```
input_shape        num_splits    bf16_us    mxfp8_us
---------------  ------------  ---------  ----------
(8, 8192, 5120)             8     4032.3     1256.14
```

## Next steps
- Refactor to use NVSHMEM apis directly for inter-node a2a
- torchtitan integration by extending this [configurable EP a2a pattern](https://github.com/pytorch/torchtitan/pull/1569/files#diff-15c119f6ad6a2475659b31edbb7a4a7cc8b58c96a2ad508ef546e8f6aeb31ca7R442-R472) 
- Extend to all_to_all_v_2d to fuse in the local token shuffle so we can do it in low precision, moving fewer bytes around -> making it faster, then dequantize only after the token shuffle.
- Look into loop unrolling, which is [used in the bf16 implementation](https://github.com/pytorch/torchtitan/blob/3e1b843ecd91a80dbf56e16f2eae7b637209ebcb/torchtitan/experiments/deepseek_v3/symm_mem_recipes/triton_on_device_all_to_all_v.py#L100)

### DIFF
--- a/torchao/prototype/moe_training/kernels/mxfp8/comms.py
+++ b/torchao/prototype/moe_training/kernels/mxfp8/comms.py
@@ -57,7 +57,8 @@ class MXFP8OnDeviceAllToAllV(torch.autograd.Function):
 
         # Quantize input
         block_size = 32
-        input_scales, input_data = to_mx(
+        to_mx_c = torch.compile(to_mx)
+        input_scales, input_data = to_mx_c(
             input,
             elem_dtype=torch.float8_e4m3fn,
             block_size=block_size,
@@ -237,8 +238,7 @@ def _mxfp8_on_device_all_to_all_v(
     output_scales: torch.Tensor,
     output_splits: torch.Tensor,
     group: dist.ProcessGroup = dist.group.WORLD,
-    BLOCKS_PER_REMOTE_RANK: int = 1,
-    UNROLL_FACTOR: int = 8,
+    BLOCKS_PER_REMOTE_RANK: int = 8,
     BLOCK_SIZE: int = 16384,
 ):
     assert input.dim() == 2, f"{input.shape}"
@@ -274,7 +274,6 @@ def _mxfp8_on_device_all_to_all_v(
         rank=input_hdl.rank,
         world_size=input_hdl.world_size,
         BLOCKS_PER_REMOTE_RANK=BLOCKS_PER_REMOTE_RANK,
-        UNROLL_FACTOR=UNROLL_FACTOR,
         BLOCK_SIZE=BLOCK_SIZE,
         num_warps=1,
     )
@@ -296,7 +295,6 @@ def _mxfp8_all_to_all_v_kernel(
     rank: tl.constexpr,
     world_size: tl.constexpr,
     BLOCKS_PER_REMOTE_RANK: tl.constexpr,
-    UNROLL_FACTOR: tl.constexpr,
     BLOCK_SIZE: tl.constexpr,
 ):
     blockwise_barrier(signal_pad_ptrs, None, rank, world_size, sem="relaxed")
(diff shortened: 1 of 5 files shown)


torchao/prototype/moe_training/kernels/mxfp8/comms.py
```diff
@@ -0,0 +1,429 @@
...
+    rank: tl.constexpr,
+    world_size: tl.constexpr,
+    BLOCKS_PER_REMOTE_RANK: tl.constexpr,
+    UNROLL_FACTOR: tl.constexpr,
+    BLOCK_SIZE: tl.constexpr,
+):
+    blockwise_barrier(signal_pad_ptrs, None, rank, world_size, sem="relaxed")
+    sync_threads()
+
+    remote_rank = tl.program_id(0) // BLOCKS_PER_REMOTE_RANK
```
kwen2501, line 303: note that this works only in a case where number of blocks >= number of ranks. 
Otherwise, we'd need to use the few blocks we have to loop over the remote_rank space.

torchao/prototype/moe_training/kernels/mxfp8/comms.py
```diff
@@ -0,0 +1,429 @@
...
+    input_hdl = symm_mem.rendezvous(input, group=group)
+    input_splits_hdl = symm_mem.rendezvous(input_splits, group=group)
+    input_scales_hdl = symm_mem.rendezvous(input_scales, group=group)
+
+    input_ptrs = input_hdl.buffer_ptrs_dev
+    input_splits_ptrs = input_splits_hdl.buffer_ptrs_dev
+    input_scales_ptrs = input_scales_hdl.buffer_ptrs_dev
+    signal_pad_ptrs = input_hdl.signal_pad_ptrs_dev
+    dim = output.shape[1]
+    dim_scaling_groups = input_scales.shape[-1]
```
kwen2501, line 261: Curious, how big is `dim_scaling_groups` in general?

kwen2501 (approved):
LGTM! Thanks for quantizing the kernel!

torchao/prototype/moe_training/kernels/mxfp8/comms.py
danielvegamyhre: For an input shape (M, K), it is the number of columns in the scales tensor, which will be shape `(M, K//block_size)`, where block_size usually is 32.

I may rename the variable to be something more clear. (At minimum I will clarify in the doc string)

PR #3049, opened 2025-09-23
Support NVFP4 dynamic per tensor scale

andrewor14:
* #3050
* __->__ #3049

**Summary:** This commit adds an option for the existing
`NVFP4InferenceConfig` to dynamically compute an appropriate
fp32 per tensor scale to support the two level scaling
according to the NVFP4 specification:
https://developer.nvidia.com/blog/introducing-nvfp4-for-efficient-and-accurate-low-precision-inference/.

While two level scaling is supported in `NVFP4Tensor`, today
there is no config API for users to call this. The existing
`NVFP4InferenceConfig` only supports single level scaling
because including an explicit `per_tensor_scale` field would
make serialization tricky.

In the future, we should add an end-to-end calibration flow
so users can compute an appropriate per tensor scale for the
activations first, and then pass this to `NVFP4Tensor` as a
static scale, similar to the proposal in https://github.com/pytorch/ao/issues/2572.

**Test Plan:**
```
pytest test/quantization/test_qat.py -k test_quantize_api_nvfp4
```

Also did a quick benchmark before and after:
```
import copy
import time
import torch
from torch._inductor.utils import do_bench_using_profiling
from torchao.prototype.mx_formats import NVFP4InferenceConfig

m_mx1 = torch.nn.Linear(1024, 1024, bias=True, dtype=torch.bfloat16, device="cuda")
m_mx2 = copy.deepcopy(m_mx1)
config1 = NVFP4InferenceConfig(use_dynamic_per_tensor_scale=False)
config2 = NVFP4InferenceConfig(use_dynamic_per_tensor_scale=True)
quantize_(m_mx1, config=config1)
quantize_(m_mx2, config=config2)
m_mx1 = torch.compile(m_mx1, fullgraph=True, backend="aot_eager")
m_mx2 = torch.compile(m_mx2, fullgraph=True, backend="aot_eager")

def run_no_per_tensor_scale():
   for _ in range(1000):
      m_mx1(torch.randn(512, 1024, device="cuda", dtype=torch.bfloat16))

def run_with_per_tensor_scale():
   for _ in range(1000):
      m_mx2(torch.randn(512, 1024, device="cuda", dtype=torch.bfloat16))

no_per_tensor_scale_elapsed_ms = do_bench_using_profiling(run_no_per_tensor_scale)
with_per_tensor_scale_elapsed_ms = do_bench_using_profiling(run_with_per_tensor_scale)

print("No per_tensor_scale = ", no_per_tensor_scale_elapsed_ms, "ms")
print("With per_tensor_scale = ", with_per_tensor_scale_elapsed_ms, "ms")
```

On a single B200:
```
No per_tensor_scale =  105.45409099999787 ms
With per_tensor_scale =  137.24245699997056 ms
```


drisspg:
Instead of using time can you use something like
https://github.com/drisspg/transformer_nuggets/blob/419095ebee68e6c7c8f511df73fbbbbc0601d96c/transformer_nuggets/utils/benchmark.py#L72-L75

to get cuda time

andrewor14:
> Instead of using time can you use something like
https://github.com/drisspg/transformer_nuggets/blob/419095ebee68e6c7c8f511df73fbbbbc0601d96c/transformer_nuggets/utils/benchmark.py#L72-L75

updated

vkuzo:
128, 64, 256 is also a really small shape so not sure perf measurements are representative

andrewor14:
> 128, 64, 256 is also a really small shape so not sure perf measurements are representative

Ok, updated to 512, 1024, 1024, it seems pretty similar still

PR #3050, opened 2025-09-23
Improve QAT nvfp4 numerics

andrewor14:
* __->__ #3050

**Summary:** Similar to https://github.com/pytorch/ao/pull/2986,
this commit improves the prepare vs convert SQNR of NVFP4 QAT
from 12 to inf. This is achieved by refactoring NVFP4 QAT to
mimic the PTQ numerics exactly, using a new linear class to
incorporate both the quantization and mm logic.

**Unit tests:**
```
python test/quantization/test_qat.py -k test_quantize_api_nvfp4
```

**End-to-end tests:**
Fine-tuning Llama3.2-3B with and without this PR in axolotl:
- fine-tune for 1 epoch on yahma/alpaca-cleaned
- batch size 512, learning rate 2e-5, no gradient accumulation

Wikitext:

- With this PR, QAT nvfp4 quantized model achieved 15% lower
  perplexity than the quantized baseline
- Without this PR, QAT nvfp4 quantized model was about the
  same as the quantized baseline

```
==> Llama3.2-3B_baseline_bs512/eval_float.log <==
|        |       |none  |     0|word_perplexity|↓  |9.418|±  |   N/A|

==> Llama3.2-3B_baseline_bs512/eval_quantized.log <==
|        |       |none  |     0|word_perplexity|↓  |10.3681|±  |   N/A|

# QAT with this PR (quantized)
==> Llama3.2-3B_qat_bs512/eval_quantized.log <==
|        |       |none  |     0|word_perplexity|↓  |10.2281|±  |   N/A|
```


test/quantization/test_qat.py
```diff
@@ -2086,9 +2085,14 @@ def test_quantize_api_nvfp4(self, use_per_tensor_scale: bool):
...
 
+        if use_per_tensor_scale:
+            target_prepare_sqnr = 36
+        else:
+            target_prepare_sqnr = float("inf")
+
         self._test_quantize_api_against_ptq(
             NVFP4InferenceConfig(use_dynamic_per_tensor_scale=use_per_tensor_scale),
-            target_prepare_sqnr=12,
+            target_prepare_sqnr=target_prepare_sqnr,
```
drisspg: can you explain this I would expect that per-tensor scaling would have less error and thus a higher sqnr? but also I dont know what prepare sqnr means?

torchao/prototype/custom_fp_utils.py
```diff
@@ -120,31 +123,41 @@ def _f32_to_floatx_unpacked(x: Tensor, ebits: int, mbits: int) -> Tensor:
...
-    normal_x = normal_x.to(torch.uint8)
+    if not fake_quantize:
+        normal_x = normal_x.to(torch.uint8)
 
     #
     # combine the branches
     #
-    x = torch.full_like(x, max_int, dtype=torch.uint8)
+    if fake_quantize:
+        x = torch.full_like(x, max_int, dtype=torch.int32)
```
drisspg: why this?

torchao/prototype/custom_fp_utils.py
```diff
@@ -120,31 +123,41 @@ def _f32_to_floatx_unpacked(x: Tensor, ebits: int, mbits: int) -> Tensor:
...
+    else:
+        x = torch.full_like(x, max_int, dtype=torch.uint8)
     x = torch.where(denormal_mask, denormal_x, x)
     x = torch.where(normal_mask, normal_x, x)
 
     # add sign back
     sign_lp = sign >> (MBITS_F32 + EBITS_F32 - mbits - ebits)
-    sign_lp = sign_lp.to(torch.uint8)
+    if not fake_quantize:
+        sign_lp = sign_lp.to(torch.uint8)
```
drisspg: would this be cleaner if all of these used the STE?

torchao/prototype/custom_fp_utils.py
```diff
@@ -24,7 +24,9 @@ def _n_ones(n: int) -> int:
 F32_EXP_BIAS = _n_ones(EBITS_F32 - 1)
 
 
-def _f32_to_floatx_unpacked(x: Tensor, ebits: int, mbits: int) -> Tensor:
+def _f32_to_floatx_unpacked(
+    x: Tensor, ebits: int, mbits: int, fake_quantize: bool = False
```
vkuzo: do all tests pass if you just make your behavior the default?  It's not ideal to have product-logic-specific branches like this.

torchao/prototype/custom_fp_utils.py
```diff
@@ -120,31 +123,41 @@ def _f32_to_floatx_unpacked(x: Tensor, ebits: int, mbits: int) -> Tensor:
...
+        sign_lp = sign_lp.to(torch.uint8)
     # Right shift of a negative signed integer can fill the least significant
     # bits with either 1s or 0s, depending on the implementation. Since PyTorch
     # doesn't have an uint32 dtype, we mask out these bits to get just the
     # f4 sign bit
     sign_lp = sign_lp & sign_mask
     x = x | sign_lp
 
-    return x.to(torch.uint8)
+    if fake_quantize:
```
vkuzo: I don't understand this, if the target dtype is 8 bits or smaller, why do you want to return more bits here?

vkuzo (changes requested):

torchao/prototype/custom_fp_utils.py
andrewor14: I removed the branches by adding a compute dtype instead, please take another look. I want to avoid changing numerics for existing use cases

torchao/prototype/custom_fp_utils.py
andrewor14: for QAT we want to avoid actually casting the value down to 8 bits, so we use higher precision to simulate this

torchao/prototype/custom_fp_utils.py
andrewor14: wrapped all of this in an STE instead

test/quantization/test_qat.py
andrewor14: This actually refers to the SQNR between (1) QAT fake quantized model and (2) PTQ actual quantized model, so a lower SQNR here just means these two don't match exactly.

I'm not sure why per tensor is not inf here. I suspect there's more logic in this case so there's larger chance for divergence. I did also try calling `NVFP4Tensor` directly for QAT just for a sanity check, but I wasn't able to get inf even with that.

torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -834,7 +833,7 @@ def _nvfp4_quantize(
     data_scaled = torch.clamp(data_scaled, -F4_E2M1_MAX, F4_E2M1_MAX)
     data_scaled = data_scaled.view(orig_shape)
     if skip_dtype_cast_and_packing:
-        return out_scales.to(torch.float32), data_scaled.to(orig_dtype)
+        return _Float8Round.apply(out_scales), data_scaled
```
vkuzo: hmm, don't you want to wrap the entire `_nvfp4_quantize` call with STE instead?

vkuzo:
> Simulate f4_unpacked_to_f32 and f32_to_f4_unpacked,
but in torch.int32 instead of torch.uint8

remove this from PR summary since looks like this is removed from the code?

torchao/prototype/qat/nvfp4.py
```diff
@@ -12,6 +14,41 @@
...
+    """
+
+    @staticmethod
+    def forward(
+        ctx, x: torch.Tensor, per_tensor_scale: Optional[torch.Tensor]
+    ) -> torch.Tensor:
+        block_size = 16
+
+        # quantize
+        scale, q = _nvfp4_quantize(
```
vkuzo: umm, how about returning `NVFP4Tensor` here (wrapped with STE), and then just dequantizing it?  then you don't need `skip_packing` flag at all, you don't need to care about `f4_unpacked_to_f32`, etc?

vkuzo: specifically you can wrap `NVFP4Tensor.to_nvfp4` with STE, and then you don't need to care about the details

vkuzo: I guess you have STE here, so you just need NVFP4Tensor

test/quantization/test_qat.py
andrewor14: Update: they're both inf now. The per-tensor case wasn't inf before because we weren't simulating `_addmm_nvfp4_dispatch` in QAT, but instead just called `F.linear` on `NVFP4Tensor.to_nvfp4().to_dtype()`, which doesn't give the same numerics

torchao/quantization/qat/fake_quantize_config.py
```diff
@@ -444,12 +444,16 @@ def _infer_fake_quantize_configs(
     elif isinstance(base_config, NVFP4InferenceConfig):
         if NVFP4MMConfig.DYNAMIC:
```
jerryzh168, line 445: nit: if this is a boolean, might be good to say NVFP4MMConfig.is_dynamic I think, although probably not relevant to this PR

torchao/prototype/qat/nvfp4.py
```diff
@@ -23,47 +22,159 @@ class NVFP4FakeQuantizeConfig(FakeQuantizeConfigBase):
...
+        use_triton_kernel (bool): Whether to use triton kernels during fake quantization
     """
 
     use_per_tensor_scale: bool = True
+    use_swizzled_scales: bool = False
+    use_triton_kernel: bool = False
 
 
-class NVFP4FakeQuantizer(FakeQuantizerBase):
+class _NVFP4FakeQuantizedLinearForward(torch.autograd.Function):
```
vkuzo: seems like this isn't `FakeQuantizedLinear`, it's actually doing quantized forward and only fake quantizing in the backward?  The name should correctly describe what this is doing.

torchao/quantization/qat/linear.py
```diff
@@ -130,13 +130,27 @@ def to_linear(self) -> torch.nn.Linear:
             new_linear.bias = self.bias
         return new_linear
 
-    @classmethod
+    @staticmethod
```
jerryzh168: why this change?

I guess it's because we may return `NVFP4FakeQuantizedLinear` now? would it be better to do this separately in this case?

torchao/quantization/qat/fake_quantize_config.py
andrewor14: ok, can fix separately since this is a PTQ config

torchao/quantization/qat/linear.py
andrewor14: it's for getting the correct linear class here: https://github.com/pytorch/ao/blob/main/torchao/quantization/qat/api.py#L211

jerryzh168: maybe it's better to the check there? and keep this `from_linear` intact?

since we typically expect `some_class.from_linear` to return an instance of `some_class`

PR #3053, opened 2025-09-24
support select.int for Float8Tensor

vkuzo:
Summary:

This is useful for stitching together 2D weights to a 3D weight, specifically this happens in vLLM for HF models where expert weights are 2D.

Test Plan:

```bash
pytest test/quantization/quantize_/workflows/float8/test_float8_tensor.py -x -s -k index
```

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/quantization/quantize_/workflows/float8/float8_tensor.py
+++ b/torchao/quantization/quantize_/workflows/float8/float8_tensor.py
@@ -617,6 +617,28 @@ def _(func, types, args, kwargs):
     return return_and_correct_aliasing(func, args, kwargs, new)
 
 
+@implements(aten.select.int)
+def _(func, types, args, kwargs):
+    old_float8_tensor, dim, index = args
+    assert dim == 0, f"Float8Tensor aten.select.int with {dim=} is not yet supported"
+    assert len(old_float8_tensor.qdata.shape) == len(old_float8_tensor.scale.shape), (
+        "unsupported"
+    )
+    assert len(old_float8_tensor.qdata.shape) == len(old_float8_tensor.block_size), (
+        "unsupported"
+    )
+    new_float8_tensor = old_float8_tensor.__class__(
+        old_float8_tensor.qdata[index],
+        old_float8_tensor.scale[index],
+        old_float8_tensor.block_size[1:],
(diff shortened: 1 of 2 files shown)


torchao/quantization/quantize_/workflows/float8/float8_tensor.py
```diff
@@ -617,6 +617,22 @@ def _(func, types, args, kwargs):
...
 
 
+@implements(aten.select.int)
+def _(func, types, args, kwargs):
+    old_float8_tensor, dim, index = args
+    assert dim == 0, f"Float8Tensor aten.select.int with {dim=} is not yet supported"
+    new_float8_tensor = old_float8_tensor.__class__(
+        old_float8_tensor.qdata[index],
+        old_float8_tensor.scale[index],
+        old_float8_tensor.block_size[1:],
```
jerryzh168, line 633: feels like we have to do an assert somewhere to make sure the rank of all three are the same to be safe

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
```diff
@@ -617,6 +617,22 @@ def _(func, types, args, kwargs):
     return return_and_correct_aliasing(func, args, kwargs, new)
 
 
+@implements(aten.select.int)
+def _(func, types, args, kwargs):
+    old_float8_tensor, dim, index = args
+    assert dim == 0, f"Float8Tensor aten.select.int with {dim=} is not yet supported"
+    new_float8_tensor = old_float8_tensor.__class__(
```
jerryzh168, line 630: nit: is it better to use `Float8Tensor` directly here? or `old_float8_tensor.__class__`? we have been using `Float8Tensor` in other functions so far

jerryzh168 (approved):

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
vkuzo: I used the syntax in the function right above which uses `__class__`, so I guess there is already a mix?  If we want consistency, maybe we can do a codemod in a future PR.

PR #3055, opened 2025-09-24
[Intel GPU] Enable optim SR test

arlesniak:
Enable optim_bf16_stochastic_round_correctness test for XPU

### DIFF
--- a/test/test_low_bit_optim.py
+++ b/test/test_low_bit_optim.py
@@ -419,8 +419,8 @@ class TestOptim(TestCase):
         for p1, p2 in zip(model1.parameters(), model2.parameters()):
             torch.testing.assert_close(p2, p1)
 
-    def test_optim_bf16_stochastic_round_correctness(self):
-        device = "cuda" if torch.cuda.is_available() else "cpu"
+    @parametrize("device", _DEVICES)
+    def test_optim_bf16_stochastic_round_correctness(self, device):
         torch.manual_seed(2024)
         model1 = nn.Sequential(nn.Linear(32, 1024), nn.ReLU(), nn.Linear(1024, 128))
         model1.to(device)


test/test_low_bit_optim.py
```diff
@@ -419,8 +419,8 @@ def test_optim_cpu_offload_save_load(self):
         for p1, p2 in zip(model1.parameters(), model2.parameters()):
             torch.testing.assert_close(p2, p1)
 
+    @parametrize("device", _DEVICES)
     def test_optim_bf16_stochastic_round_correctness(self):
```
jerryzh168: device arg is not added to the function

liangan1: @arlesniak  __DEVICES_ is the global variable in this file, I don't think your change is needed.

arlesniak: The "test_optim_bf16_stochastic_round_correctness" should be parametrized with _DEVICES global variable, which contains other devices (XPU if detected). Without the PR changes the choice was cuda, cpu only.

liangan1: Are you sure? According to this code,https://github.com/pytorch/ao/blob/main/torchao/utils.py#L143 when the xpu is available in your ENV, the _DEVICES will contain the XPU.

arlesniak: Let me be more precise. You're right, the _DEVICES list will contain XPU if detected. Without the PR changes the test will be triggered once with the hardcoded device choice - cuda or cpu: https://github.com/pytorch/ao/blob/main/test/test_low_bit_optim.py#L423, regardless of _DEVICES content. 
After the PR changes, the test will be triggered separately for every device from the _DEVICES including XPU (similiar to some other parametrized tests i.e. test_quantize_8bit_with_qmap_correctness etc).

EikanWang (approved):

liangan1 (approved):

PR #3057, opened 2025-09-24
Fix torchAO shape check on fp8 tensors

RandySheriff:
Differential Revision: D83184235


jcaip (approved):
lgtm, thanks for the fix!

PR #3061, opened 2025-09-25
[XPU] Support _weight_int8pack_mm path for all backends and INT8 quantization on Intel XPU

xiaowangintel:
# Motivation
There is the XMX intrinsic on the Intel GPUs, e.g., Intel Arc B580, to speedup the int8 computation. This PR is target to ensure the int8 quantization works on the XPU and user can use same quantization config with CUDA. In addition, we support _weight_int8pack_mm path for all backends to replace mm + mul on woq-int8 mode, and expect to improve performance.

# Implementation
The _weight_int8pack_mm kernel for XPU backend is supported in https://github.com/pytorch/pytorch/pull/160938. Similar kernels are supported on CPU and CUDA.

# Usage
Users can use same quantization config, such as Int8StaticActivationInt8WeightConfig, Int8DynamicActivationInt8WeightConfig and Int8WeightOnlyConfig as CUDA to run the int8 models on XPU. This device is dispatched by the PyTorch op.
```

#For static int8 quantization
            granularity=granularity,
        )

#For  woq-int8 quantization
from torchao.quantization import quantize_, Int8WeightOnlyConfig
quantize_(model, Int8WeightOnlyConfig(version=2, granularity=PerRow()))

```

### DIFF
--- a/test/quantization/quantize_/workflows/int8/test_int8_tensor.py
+++ b/test/quantization/quantize_/workflows/int8/test_int8_tensor.py
@@ -26,7 +26,7 @@ from torchao.quantization.quantize_.common import (
 from torchao.quantization.utils import compute_error, get_block_size
 from torchao.testing.model_architectures import ToyTwoLinearModel
 from torchao.testing.utils import TorchAOIntegrationTestCase
-from torchao.utils import torch_version_at_least
+from torchao.utils import get_current_accelerator_device, torch_version_at_least
 
 INT8_TEST_CONFIGS = [
     Int8WeightOnlyConfig(version=2, granularity=PerTensor()),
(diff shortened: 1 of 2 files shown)


torchao/dtypes/uintx/plain_layout.py
```diff
@@ -253,16 +253,25 @@ def _linear_fp_act_int8_weight_impl(input_tensor, weight_tensor, bias):
     # return torch.ops.aten._weight_int8pack_mm(input_tensor.contiguous(), w_vals_int8_t, weight_tensor.tensor_impl.scale)
 
     # per channel int8 weight only quantizated mm
-    w_vals_int8_t = weight_tensor.tensor_impl.int_data.t()
+    w_vals_int8 = weight_tensor.tensor_impl.int_data
```
jerryzh168: this is a code path for int8 cuda as well I think, changing it has a risk of perf regressions

also this is the older stack, I'd suggest to migrate first, WIP here for int8 + plain layout: https://github.com/pytorch/ao/pull/3038

liangan1:
@xiaowangintel  Pls resolve the UT fails

liangan1 (approved):

liangan1:
@xiaowangintel  you may need also to add the usage for the woq path.

torchao/dtypes/uintx/plain_layout.py
liangan1: @jerryzh168 this pr is based on the Int8Tensor introduced by #3038, can you help to review it again.

xiaowangintel:
This is the older stack. The new software stack already supports Intel GPUs.

PR #3070, opened 2025-09-26
mxtensor: support clone

vkuzo:
Summary:

This is needed for saving HF models with mxfp4 weights to disk.

Test Plan:

```bash
pytest test/prototype/mx_formats/test_mx_tensor.py -s -k test_clone
```

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo:
* #3081
* #3080
* #3079
* #3078
* #3072
* __->__ #3070

torchao/prototype/mx_formats/mx_ops.py
```diff
@@ -352,3 +352,16 @@ def autocast_to_copy(func, types, args, kwargs):
...
     return tensor
+
+
+@implements([aten.clone.default])
+def mx_clone(func, types, args, kwargs):
+    self = args[0]
+    memory_format = kwargs.get("memory_format", None)
+
+    if memory_format is not None:
+        clone_fn = lambda x: x.clone(memory_format=memory_format)
```
danielvegamyhre, line 363: Why not just avoid the if statement with `clone_fn = lambda x: x.clone(**kwargs)`? Do we need to explicitly exclude other kwargs and only include memory format if specified, for some reason?

vkuzo: this is copy-pasta from https://github.com/pytorch/ao/blob/7690612a018fea6c0ed4efeaa4809c0798d12786/torchao/prototype/mx_formats/nvfp4_tensor.py#L346, IMO let's separate cleanup of this into a separate PR

PR #3059, opened 2025-09-25
[Reland] Unify get_block_size

Xia-Weiwen:
**Summary**
It's a reopen of https://github.com/pytorch/ao/pull/3039 but retains backward compatibility. And UTs are not changed to ensure backward compatibility.
This PR merges the two definitions of `get_block_size` in ao/torchao/quantization/pt2e/observer.py and torchao/quantization/observer.py and put it in torchao/quantization/utils.py
It also removes duplicate definitions of granularities in the pt2e path.

**Test plan**
This util function is used in many places so it should have been covered by all kinds of UT.

PR #3052, opened 2025-09-24
[CPU] Add ops for float8 linear

Xia-Weiwen:
**Summary**
This is part of a previous PR https://github.com/pytorch/ao/pull/2505 since the original is too big.
This PR adds two ops for float8 linear on CPU, one for weight packing and the other for computation.
- float8_linear_prepack_cpu
- float8_linear_cpu

They will be used for float8 tensor subclass in the future.

**Test plan**
```
pytest -sv test/test_ops.py -k test_float8_linear_cpu
```

PR #3075, opened 2025-09-26
[CPU] add Float8OpaqueTensor for dynamic float8 act float8 weight

Xia-Weiwen:
**Summary**
We split the original big PR https://github.com/pytorch/ao/pull/2505 into the following smaller ones:
- https://github.com/pytorch/ao/pull/3039 (relanded by https://github.com/pytorch/ao/pull/3059)
- https://github.com/pytorch/ao/pull/3052
- And this PR https://github.com/pytorch/ao/pull/3075, which as the Float8OpaqueTensor for dynamic float8 act float8 weight quantization on CPU

**Test plan**
```
pytest -sv test/quantization/quantize_/workflows/float8/test_float8_opaque_tensor.py
```

### DIFF
--- a/test/quantization/quantize_/workflows/float8/test_float8_opaque_tensor.py
+++ b/test/quantization/quantize_/workflows/float8/test_float8_opaque_tensor.py
@@ -70,7 +70,7 @@ class TestFloat8OpaqueTensor(TestCase):
                 return
         device = "cpu"
         m = ToyTwoLinearModel(256, 256, 256, dtype, device, bias).eval()
-        example_inputs = m.example_inputs(batch_size=bs, dtype=dtype)
+        example_inputs = m.example_inputs(batch_size=bs)
         if x_dim == 3:
             example_inputs = (example_inputs[0].unsqueeze(0),)
         y = m(*example_inputs)
@@ -104,7 +104,7 @@ class TestFloat8OpaqueTensor(TestCase):
         """
         device = "cpu"
         m = ToyTwoLinearModel(120, 120, 120, dtype, device, bias).eval()
-        example_inputs = m.example_inputs(batch_size=bs, dtype=dtype)
+        example_inputs = m.example_inputs(batch_size=bs)
         if x_dim == 3:
             example_inputs = (example_inputs[0].unsqueeze(0),)
         y = m(*example_inputs)
--- a/torchao/quantization/quant_api.py
+++ b/torchao/quantization/quant_api.py
@@ -1718,22 +1718,6 @@ def _input_activation_quant_func_fp8(
     return activation
 
 
-def _input_activation_quant_cpu_fp8(
-    x: torch.Tensor,
-    activation_granularity: FP8Granularity,
-    activation_dtype: torch.dtype,
-):
-    """Dynamic quantize activation to fp8 for CPU."""
-    block_size = get_block_size(x.shape, activation_granularity)
-    return to_affine_quantized_floatx(
-        input_float=x,
-        block_size=block_size,
-        target_dtype=activation_dtype,
-        scale_dtype=torch.float32,
-        _layout=PlainLayout(),
-    )
-
-
 def _fp8_mm_compat(weight: torch.Tensor) -> bool:
     """
     Check if a weight tensor meets float8 quantization requirements.
@@ -1843,37 +1827,36 @@ def _float8_dynamic_activation_float8_weight_quantize_tensor(weight, config):
     # Ensure works on device
     activation_granularity, weight_granularity = granularity
 
-    # Note: right now we assume it's weights of conv2d and conv3d purely based
-    # on the dimension of weight, currently there is no conflict with linear 2d
-    # and moe weights 3d
-    # if we need to support conv1d, which also has 3d weight, we may have to
-    # pass around the module as well to distinguish between conv1d and 3d moe weight
-    if weight.dim() in [4, 5]:
-        # weights for conv2d or 3d
-        assert isinstance(activation_granularity, PerTensor) and isinstance(
-            weight_granularity, PerTensor
-        ), "4D/5D tensor only supports per tensor activation and weight quantization"
+    if float8_packing_format == Float8PackingFormat.PLAIN:
+        # Note: right now we assume it's weights of conv2d and conv3d purely based
+        # on the dimension of weight, currently there is no conflict with linear 2d
+        # and moe weights 3d
+        # if we need to support conv1d, which also has 3d weight, we may have to
+        # pass around the module as well to distinguish between conv1d and 3d moe weight
+        if weight.dim() in [4, 5]:
+            # weights for conv2d or 3d
+            assert isinstance(activation_granularity, PerTensor) and isinstance(
+                weight_granularity, PerTensor
+            ), (
+                "4D/5D tensor only supports per tensor activation and weight quantization"
+            )
 
-        # conv3d weight dim: (C_out, C_in, K1, K2, K3)
-        # conv2d weight dim: (C_out, C_in, K1, K2)
-        # skip quantization when either C_out or C_in
-        # is not a multiple of 16
-        if weight.shape[0] % 16 != 0 or weight.shape[1] % 16 != 0:
+            # conv3d weight dim: (C_out, C_in, K1, K2, K3)
+            # conv2d weight dim: (C_out, C_in, K1, K2)
+            # skip quantization when either C_out or C_in
+            # is not a multiple of 16
+            if weight.shape[0] % 16 != 0 or weight.shape[1] % 16 != 0:
+                return weight
+
+        elif not _fp8_mm_compat(weight):
+            # TODO(future PR): this should really throw an exception instead of silently
+            # not doing what the user asked
             return weight
 
-    elif float8_packing_format == Float8PackingFormat.PLAIN and not _fp8_mm_compat(
-        weight
-    ):
-        # TODO(future PR): this should really throw an exception instead of silently
-        # not doing what the user asked
-        return weight
-
-    if float8_packing_format == Float8PackingFormat.PLAIN and isinstance(
-        weight_granularity, PerRow
-    ):
-        assert weight.dtype == torch.bfloat16, (
-            "PerRow quantization only works for bfloat16 precision input weight"
-        )
+        if isinstance(weight_granularity, PerRow):
+            assert weight.dtype == torch.bfloat16, (
+                "PerRow quantization only works for bfloat16 precision input weight"
+            )
 
     if config.version == 1:
         warnings.warn(
--- a/torchao/quantization/quantize_/workflows/float8/float8_opaque_tensor.py
+++ b/torchao/quantization/quantize_/workflows/float8/float8_opaque_tensor.py
@@ -18,7 +18,9 @@ from torchao.quantization.quant_primitives import (
     _choose_scale_float8,
     _quantize_affine_float8,
 )
-from torchao.quantization.utils import get_block_size
+from torchao.quantization.quantize_.common import (
+    _choose_quant_func_and_quantize_tensor,
+)
 from torchao.utils import (
     TorchAOBaseTensor,
 )
@@ -237,13 +239,11 @@ def _(func, types, args, kwargs):
                     "input and weight should have the same group size but got"
                     f" {weight_tensor.block_size[1]} and {group_size}"
                 )
-        act_block_size = get_block_size(act_mat.shape, granularity)
-        act_scale = _choose_scale_float8(
-            act_mat,
-            float8_dtype=torch.float8_e4m3fn,
-            block_size=act_block_size,
+        input_tensor = _choose_quant_func_and_quantize_tensor(
+            act_mat, weight_tensor.act_quant_kwargs
         )
-        act_mat = _quantize_affine_float8(act_mat, act_scale, torch.float8_e4m3fn)
+        act_mat = input_tensor.qdata
+        act_scale = input_tensor.scale
     else:
         raise NotImplementedError(
             "Activation quantization args not provided for Float8OpaqueTensor"
--- a/torchao/testing/model_architectures.py
+++ b/torchao/testing/model_architectures.py
@@ -64,12 +64,12 @@ class ToyTwoLinearModel(torch.nn.Module):
         )
 
     # Note: Tiny-GEMM kernel only uses BF16 inputs
-    def example_inputs(self, batch_size=1, dtype=None):
+    def example_inputs(self, batch_size=1):
         return (
             torch.randn(
                 batch_size,
                 self.linear1.in_features,
-                dtype=dtype or self.dtype,
+                dtype=self.dtype,
                 device=self.device,
             ),
         )


Xia-Weiwen:

Xia-Weiwen:
Hi @mingfeima @jerryzh168 @andrewor14 Could you please review this PR? Thanks.

test/quantization/quantize_/workflows/float8/test_float8_opaque_tensor.py
```diff
@@ -0,0 +1,164 @@
...
+    )
+    @unittest.skipIf(not torch_version_at_least("2.6.0"), "Test only enabled for 2.6+")
+    @common_utils.parametrize("dtype", [torch.float, torch.bfloat16, torch.half])
+    @common_utils.parametrize("x_dim", [2, 3])
+    @common_utils.parametrize("bias", [True, False])
+    @common_utils.parametrize("bs", [1, 160])
+    @common_utils.parametrize(
+        "x_granularity",
+        [PerTensor(), PerRow(), PerGroup(32), PerGroup(64), PerGroup(128)],
+    )
```
mingfeima, line 58: does torch.ao support per block quantization, e.g. deepseek style?

torchao/quantization/quantize_/workflows/float8/float8_opaque_tensor.py
```diff
@@ -0,0 +1,227 @@
...
+__all__ = [
+    "Float8OpaqueTensor",
+]
+
+aten = torch.ops.aten
+
+
+class Float8OpaqueTensor(TorchAOBaseTensor):
+    """
+    Float8 dynamic activation float8 weight on CPU. The weight tensor is reordered to a blocked layout
```
mingfeima, line 40: should it be `Float8 dynamic quantized float8 weight on CPU`

torchao/quantization/quantize_/workflows/float8/float8_opaque_tensor.py
```diff
@@ -0,0 +1,227 @@
...
+    for better memory locality from [N, K] to [N/block_n, K/block_k, block_k, block_n], where block_n = 32
+    and block_k depends on group-size for quantization (=32/64/128). And the innermost block with shape
+    [block_k, block_n] may be further reordered to VNNI layout depending on supported CPU ISA.
+
+    Tensor Attributes:
+        qdata: Reordered float8 weight on CPU with shape = [N/block_n, K/block_k, block_k, block_n].
```
mingfeima, line 46: are we computing with float32 here, as the weight is not packed in vnni2 format.

test/quantization/quantize_/workflows/float8/test_float8_opaque_tensor.py
Xia-Weiwen: Thanks for the comment. The supported granularity varies among different quantization methods in Torchao. For float8 da8w8 on CPU, it does not support the block-wise quantization used in DeepSeek.

torchao/quantization/quantize_/workflows/float8/float8_opaque_tensor.py
Xia-Weiwen: The expression here is interpreted as "Float8 dynamic activation" "float8 weight" on CPU, which means activation is dynamic quantized to float8 and weight is static quantized to float8. This is aligned with other parts in Torchao.

torchao/quantization/quantize_/workflows/float8/float8_opaque_tensor.py
Xia-Weiwen: We are computing with bf16 or fp8, depending on ISA. The exposed shape does not have the VNNI dimension but the memory layout is VNNI-2 or VNNI-4 if ISA is supported.

Xia-Weiwen:
Hi @mingfeima @jerryzh168 @andrewor14  Though this PR depends on https://github.com/pytorch/ao/pull/3100, could you please review this PR? Thanks.

torchao/quantization/quantize_/workflows/float8/float8_opaque_tensor.py
```diff
@@ -0,0 +1,227 @@
...
+            )
+            assert K % block_size[1] == 0, (
+                f"Expecting in_features {K} to be multiple of group_size {block_size[1]}, but got {K}"
+            )
+        scale = _choose_scale_float8(
+            hp_tensor,
+            float8_dtype=torch.float8_e4m3fn,
+            block_size=block_size,
+        )
+        data = _quantize_affine_float8(hp_tensor, scale, torch.float8_e4m3fn)
```
jerryzh168, line 181: do you need to use https://github.com/pytorch/ao/blob/c96f2dd0c980ad100ea9e7c574dead8f19451955/torchao/quantization/quant_primitives.py#L2425?

torchao/float8/inference.py
```diff
@@ -230,6 +231,47 @@ def _normalize_granularity(
     return processed_granularity
 
 
+def _normalize_granularity_opaque_tensor(
```
jerryzh168: why this can't reuse the other `normalize_granularity_tensor`?

torchao/float8/types.py
```diff
@@ -12,8 +12,9 @@
 from typing import TYPE_CHECKING, Union
 
 if TYPE_CHECKING:
-    from torchao.quantization.granularity import PerRow, PerTensor
+    from torchao.quantization.granularity import PerGroup, PerRow, PerTensor
 
 
 # Define FP8Granularity type alias to break circular import dependencies
 FP8Granularity = Union["PerTensor", "PerRow"]
+FP8GranularityCPU = Union["PerTensor", "PerRow", "PerGroup"]
```
jerryzh168: I feel we can reuse and extend `FP8Granularity` and assert only part of the options are supported for GPU right now

torchao/quantization/quant_api.py
```diff
@@ -1722,6 +1726,26 @@ def _input_activation_quant_func_fp8(
...
+    x: torch.Tensor,
+    activation_granularity: FP8Granularity,
+    activation_dtype: torch.dtype,
+):
+    """Dynamic quantize activation to fp8 for CPU."""
+    if not isinstance(activation_granularity, PerGroup):
+        block_size = get_block_size(x.shape, activation_granularity)
+    else:
+        group_size = activation_granularity.group_size
+        block_size = (*([1] * (len(x.shape) - 1)), group_size)
```
jerryzh168: why is this not included in `get_block_size`?

torchao/quantization/quant_api.py
```diff
@@ -1807,17 +1837,20 @@ def _float8_dynamic_activation_float8_weight_quantize_tensor(weight, config):
     activation_value_lb = config.activation_value_lb
     activation_value_ub = config.activation_value_ub
     kernel_preference = config.kernel_preference
+    float8_packing_format = config.float8_packing_format
 
     # Ensure works on device
-    _check_hardware_support(granularity)
+    is_cpu = weight.device.type == "cpu"
+    if not is_cpu:
+        _check_hardware_support(granularity)
```
jerryzh168: can you move this to version 1? and then version 2 can do this check in the tensor itself probably

torchao/quantization/quant_api.py
```diff
@@ -1807,17 +1837,20 @@ def _float8_dynamic_activation_float8_weight_quantize_tensor(weight, config):
...
+    if not is_cpu and not _fp8_mm_compat(weight):
         # TODO(future PR): this should really throw an exception instead of silently
         # not doing what the user asked
         return weight
 
-    if isinstance(weight_granularity, PerRow):
+    if not is_cpu and isinstance(weight_granularity, PerRow):
         assert weight.dtype == torch.bfloat16, (
             "PerRow quantization only works for bfloat16 precision input weight"
         )
```
jerryzh168: also these checks, I feel we can move these to version 1 branch for now and deprecate later, we can add the checks to tensors for version 2

torchao/quantization/quantize_/workflows/float8/float8_opaque_tensor.py
Xia-Weiwen: Thanks. Since we are not using Inductor for fusion like PT2E, it should be OK here.

torchao/float8/inference.py
Xia-Weiwen: Thanks. Updated

torchao/float8/types.py
Xia-Weiwen: Thanks. Updated.

torchao/quantization/quant_api.py
Xia-Weiwen: Updated. Thanks.

torchao/quantization/quant_api.py
Xia-Weiwen: Sure. Thanks.

torchao/quantization/quant_api.py
Xia-Weiwen: Sure. thanks.

Xia-Weiwen: Moving this to `version=1` branch causes CI failures. I will keep them as is. Maybe it can be improved later. Thanks.

Xia-Weiwen: The data type check is kept here and the `_fp8_mm_compat` check is moved to `version=1`. Thanks.

Xia-Weiwen:
@jerryzh168 Could you please review this PR again? Thanks.

torchao/float8/inference.py
```diff
@@ -204,28 +205,41 @@ def _normalize_granularity(
             list[FP8Granularity],
         ]
     ],
+    supported_granularities: tuple[FP8Granularity] = (PerTensor, PerRow),
+    support_different_granularities: bool = False,
```
jerryzh168: this is weird, I think we should have normalize_granularity to only do normalize, not also validation

Xia-Weiwen: I feel the same actually. Where should we put the validation? Thanks.

jerryzh168: this seems to `_normalize_and_validate_granularities`, can you define separate functions for both float8 tensor and float8 opque tensor in the tensor file itself? i.e. float8_tensor.py and float8_opque_tensor.py

probably will be clearer if you do this in a separate PR, that is move the original `_normalize` function to float8_tensor.py and change all the callsites first, and then in this PR you just need to add a new one for float8_opque_tensor.py

Xia-Weiwen: Sounds good. Will do. Thanks

Xia-Weiwen: How about version=1? Call `_validate_granularity` explicitly? In that case, `_validate_granularity` cannot be bound to a specific tensor type I guess. And `_normalize_granularity` (with checks) is called elsewhere too:
- https://github.com/pytorch/ao/blob/838dcebb13be7cd7b7ae5d1d8c4e674c89a5d7b8/torchao/quantization/quant_api.py#L1945
- https://github.com/pytorch/ao/blob/838dcebb13be7cd7b7ae5d1d8c4e674c89a5d7b8/torchao/quantization/qat/fake_quantize_config.py#L422

How shall we do validation at these locations?

Xia-Weiwen: Hi @jerryzh168 Do you have any suggestions for this? Thanks.

jerryzh168: I think just create `_normalize_and_validate_granularities` for both float8_tensor.py and float8_opaque_tensor.py in their own files without sharing anything seems to be the cleanest

jerryzh168: oh actually looks like https://github.com/pytorch/ao/blob/838dcebb13be7cd7b7ae5d1d8c4e674c89a5d7b8/torchao/quantization/qat/fake_quantize_config.py#L422 needs a pure normalize function

I think you can do the following:

torchao/quantization/quantize_/workflows/float8/utils.py
_normalize_granularity (make a single element to be a tuple)

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
_validate_granluarity (per row, per tensor)

torchao/quantization/quantize_/workflows/float8/float8_opaque_tensor.py
_validate_granluarity (per row, per tensor, per block)

Xia-Weiwen:
Hi @jerryzh168 It's been awhile since last update of this PR because of reversions of prior PRs. I have rebased this PR. And for the `_normalize_granularities` thing we discussed above (https://github.com/pytorch/ao/pull/3075#discussion_r2433348592), I have added a method `_normalize_and_check_granularities` to `Float8OpaqueTensor`. This brings minimal changes and keeps the code clean. Could you please review this PR again? Thanks.

test/quantization/quantize_/workflows/float8/test_float8_opaque_tensor.py
```diff
@@ -0,0 +1,164 @@
...
+            torch.save(linear.state_dict(), f)
+            f.seek(0)
+            state_dict = torch.load(f)
+            self.assertEqual(
+                str(type(state_dict["weight"])),
+                "<class 'torchao.quantization.Float8OpaqueTensor'>",
+            )
+
+
+common_utils.instantiate_parametrized_tests(TestFloat8OpaqueTensor)
```
jerryzh168: nit: seems that we can add this  to `TestFloat8OpaqueTensor` class as a decorator

test/quantization/quantize_/workflows/float8/test_float8_opaque_tensor.py
```diff
@@ -0,0 +1,164 @@
...
+    @unittest.skipIf(
+        "CPU" not in torch._C._dispatch_dump("torchao::float8_linear_cpu"),
+        reason="cpp kernels not built",
+    )
+    @unittest.skipIf(not torch_version_at_least("2.6.0"), "Test only enabled for 2.6+")
+    @common_utils.parametrize("dtype", [torch.float, torch.bfloat16, torch.half])
+    @common_utils.parametrize("x_dim", [2, 3])
+    @common_utils.parametrize("bias", [True, False])
+    @common_utils.parametrize("bs", [4, 128])
+    def test_dynamic_float8_linear_ref(self, dtype, x_dim, bias, bs):
```
jerryzh168: what does ref mean? what's the difference between this test and previous one?

test/quantization/quantize_/workflows/float8/test_float8_opaque_tensor.py
```diff
@@ -0,0 +1,164 @@
...
+            if w_granularity.group_size != x_granularity.group_size:
+                return
+        device = "cpu"
+        m = ToyLinearModel(256, 256, bias=bias).eval().to(dtype).to(device)
+        example_inputs = m.example_inputs(batch_size=bs, dtype=dtype, device=device)
+        if x_dim == 3:
+            example_inputs = (example_inputs[0].unsqueeze(0),)
+        y = m(*example_inputs)
+
+        with torch.no_grad():
```
jerryzh168: can do https://github.com/pytorch/ao/commit/8e3b3da5444e69434a5a0997767d6a38a8b40bf8 in setUp

test/quantization/quantize_/workflows/float8/test_float8_opaque_tensor.py
```diff
@@ -0,0 +1,164 @@
...
+
+def get_config(granularity):
+    return Float8DynamicActivationFloat8WeightConfig(
+        activation_dtype=torch.float8_e4m3fn,
+        granularity=granularity,
+        float8_packing_format="opaque",
+    )
+
+
+class ToyLinearModel(torch.nn.Module):
```
jerryzh168: we also just landed a util for this https://github.com/pytorch/ao/commit/b4ec4cbab799b5320fc1ca094e0398b3ad967dbf

test/quantization/quantize_/workflows/float8/test_float8_opaque_tensor.py
```diff
@@ -0,0 +1,164 @@
...
+import torch
+from torch.testing._internal import common_utils
+from torch.testing._internal.common_utils import (
+    TestCase,
+    run_tests,
+)
+
+from torchao import quantize_
+from torchao.quantization import PerGroup, PerRow, PerTensor
+from torchao.quantization.quant_api import (
```
jerryzh168: nit: we can just import from `torchao.quantization`, I feel in the end we might be able to make quant_api.py as implementation detail and don't expose to users

torchao/quantization/quant_api.py
```diff
@@ -1834,12 +1862,12 @@ def _float8_dynamic_activation_float8_weight_quantize_tensor(weight, config):
         if weight.shape[0] % 16 != 0 or weight.shape[1] % 16 != 0:
             return weight
 
-    elif not _fp8_mm_compat(weight):
+    elif not is_cpu and not _fp8_mm_compat(weight):
```
jerryzh168: maybe check packing_format instead of device? since we are trying to make this not device specific

torchao/quantization/quant_api.py
```diff
@@ -1834,12 +1862,12 @@ def _float8_dynamic_activation_float8_weight_quantize_tensor(weight, config):
...
         # TODO(future PR): this should really throw an exception instead of silently
         # not doing what the user asked
         return weight
 
-    if isinstance(weight_granularity, PerRow):
+    if not is_cpu and isinstance(weight_granularity, PerRow):
```

torchao/quantization/quant_api.py
```diff
@@ -1849,6 +1877,7 @@ def _float8_dynamic_activation_float8_weight_quantize_tensor(weight, config):
             "Config Deprecation: version 1 of Float8DynamicActivationFloat8WeightConfig is deprecated and will no longer be supported in a future release, please use version 2, see https://github.com/pytorch/ao/issues/2649 for more details"
         )
 
+        _check_hardware_support(granularity)
```
jerryzh168, line 1866: this function name seems too general, but we can improve this later

torchao/quantization/quant_api.py
```diff
@@ -1898,12 +1939,6 @@ def _float8_dynamic_activation_float8_weight_transform(
     *,
     parameter_name: str = "weight",
 ):
-    assert is_sm_at_least_89() or is_MI300(), (
```
jerryzh168, line 1901: why is this removed? is it because it's already checked in `_check_hardware_support`?

torchao/quantization/quant_api.py
```diff
@@ -1898,12 +1939,6 @@ def _float8_dynamic_activation_float8_weight_transform(
...
-        "Float8 dynamic activation quantization is only supported on CUDA>=8.9 and MI300+"
-    )
-    if config.set_inductor_config:
```
jerryzh168, line 1904: also why is this removed?

torchao/quantization/quantize_/workflows/float8/float8_opaque_tensor.py
```diff
@@ -0,0 +1,273 @@
...
+from typing import List, Optional, Tuple, Union
+
+import torch
+
+from torchao.quantization.granularity import (
+    PerGroup,
+    PerRow,
+    PerTensor,
+)
+from torchao.quantization.observer import get_block_size
```
jerryzh168: why importing from observer? I thought it's moved to torchao.quantization.utils?

torchao/quantization/quantize_/workflows/float8/float8_opaque_tensor.py
```diff
@@ -0,0 +1,273 @@
...
+    )
+    assert isinstance(weight_tensor, Float8OpaqueTensor), (
+        f"Expected weight_tensor to be Float8OpaqueTensor, got: {type(weight_tensor)}"
+    )
+    assert weight_tensor.ndim in [2, 4]
+    assert input_tensor.size(-1) == weight_tensor.size(-1), (
+        f"Shapes of input and weight do not match, input:{input_tensor.shape}, weight: {weight_tensor.shape}"
+    )
+
+    act_mat = input_tensor.contiguous()
```
jerryzh168, line 219: isn't this going to be slow?

torchao/quantization/quantize_/workflows/float8/float8_opaque_tensor.py
```diff
@@ -0,0 +1,273 @@
...
+                    "input and weight should have the same group size but got"
+                    f" {weight_tensor.block_size[1]} and {group_size}"
+                )
+        act_block_size = get_block_size(act_mat.shape, granularity)
+        act_scale = _choose_scale_float8(
+            act_mat,
+            float8_dtype=torch.float8_e4m3fn,
+            block_size=act_block_size,
+        )
+        act_mat = _quantize_affine_float8(act_mat, act_scale, torch.float8_e4m3fn)
```
jerryzh168: why is this not using https://github.com/pytorch/ao/blob/b4ec4cbab799b5320fc1ca094e0398b3ad967dbf/torchao/quantization/quantize_/workflows/float8/float8_tensor.py#L311

torchao/quantization/quantize_/workflows/float8/float8_opaque_tensor.py
```diff
@@ -0,0 +1,273 @@
...
+        )
+
+    # float8 quantized linear operation
+    y = torch.ops.torchao.float8_linear_cpu.default(
+        act_mat,
+        act_scale,
+        packed_weight,
+        scale,
+        bias.float() if bias is not None else bias,  # requires bias to be float
+        torch.float,  # out_dtype
```
jerryzh168: shouldn't this align with the original activation dtype `orig_dtype`? oh or are you trying to do this for better precision?

torchao/quantization/quantize_/workflows/float8/float8_opaque_tensor.py
```diff
@@ -0,0 +1,273 @@
...
+                    f" {block_size[1]} and {group_size}"
+                )
+        if act_per_group_quant or wei_per_group_quant:
+            assert N % 32 == 0, (
+                f"Expecting out_features {N} to be multiple of 32, but got {N}"
+            )
+            assert K % block_size[1] == 0, (
+                f"Expecting in_features {K} to be multiple of group_size {block_size[1]}, but got {K}"
+            )
+        scale = _choose_scale_float8(
```
jerryzh168, line 176: I recently found https://github.com/pytorch/ao/issues/3324, does it affect your per tensor use case?

test/quantization/quantize_/workflows/float8/test_float8_opaque_tensor.py
Xia-Weiwen: Updated. Thanks.

test/quantization/quantize_/workflows/float8/test_float8_opaque_tensor.py
Xia-Weiwen: It tests the fallback path in the kernel. I have updated the name and added a comment to explain this. Thanks.

test/quantization/quantize_/workflows/float8/test_float8_opaque_tensor.py
Xia-Weiwen: Updated. Thanks.

test/quantization/quantize_/workflows/float8/test_float8_opaque_tensor.py
Xia-Weiwen: Updated. Thanks.

test/quantization/quantize_/workflows/float8/test_float8_opaque_tensor.py
Xia-Weiwen: Updated. Thanks.

torchao/quantization/quant_api.py
Xia-Weiwen: Updated. Thanks.

torchao/quantization/quant_api.py
Xia-Weiwen: Updated. Thanks.

torchao/quantization/quant_api.py
Xia-Weiwen: Yeah. This function is not added by this PR. I just moved it here for `config.version == 1`.

torchao/quantization/quant_api.py
Xia-Weiwen: It's probably due to rebase conflicts. I have added this back. Thanks.

torchao/quantization/quant_api.py
Xia-Weiwen: It's probably due to rebase conflicts. I have added this back. Thanks.

torchao/quantization/quantize_/workflows/float8/float8_opaque_tensor.py
Xia-Weiwen: Updated. Thanks.

torchao/quantization/quantize_/workflows/float8/float8_opaque_tensor.py
Xia-Weiwen: On CPU, we require input tensors to be contiguous. In fact, we almost always get contiguous inputs. So, the reordering won't actually happen. Here it just ensures the assumption.

torchao/quantization/quantize_/workflows/float8/float8_opaque_tensor.py
Xia-Weiwen: Thanks for the pointer. However, `_choose_quant_func_and_quantize_tensor` does the following:
```python
    if isinstance(quant_kwargs, QuantizeTensorToFloat8Kwargs):
        return Float8Tensor.from_hp(...)
```
Unfortunately, `Float8OpaqueTensor` also uses `QuantizeTensorToFloat8Kwargs` so it cannot distinguish them.
Besides, in the implementation of `Float8Tensor`, activation is quantized by `Float8Tensor.from_hp` to a `Float8Tensor` and then unwrapped to get the quantized tensor data for computation. And this part of logic is not exposed to users. So, I feel that it's unnecessary to use `Float8OpaqueTensor.from_hp` to quantize the activation then unwrap it. It looks good to quantize it with `_quantize_affine_float8`.
What do you think? If you want `Float8OpaqueTensor` to be aligned with `Float8Tensor`, we may need to define a counterpart of `QuantizeTensorToFloat8Kwargs` for `Float8OpaqueTensor` so that we can distinguish them. Thanks.

torchao/quantization/quantize_/workflows/float8/float8_opaque_tensor.py
Xia-Weiwen: Updated. Thanks.

torchao/quantization/quantize_/workflows/float8/float8_opaque_tensor.py
Xia-Weiwen: Thanks for the reminder. I didn't meet any issue with this.

Xia-Weiwen:
Hi @jerryzh168 I have updated this PR per your comments. Could you please review again? Thanks.

torchao/quantization/quant_api.py
```diff
@@ -1716,6 +1718,22 @@ def _input_activation_quant_func_fp8(
     return activation
 
 
+def _input_activation_quant_cpu_fp8(
```
jerryzh168: is this function used?

torchao/quantization/quant_api.py
```diff
@@ -1834,12 +1861,16 @@ def _float8_dynamic_activation_float8_weight_quantize_tensor(weight, config):
         if weight.shape[0] % 16 != 0 or weight.shape[1] % 16 != 0:
             return weight
 
-    elif not _fp8_mm_compat(weight):
+    elif float8_packing_format == Float8PackingFormat.PLAIN and not _fp8_mm_compat(
```
jerryzh168: to make these less complicated, can you lift the `float8_packing_format == Float8PackingFormat.PLAIN` before line 1851

torchao/quantization/quantize_/workflows/float8/float8_opaque_tensor.py
jerryzh168: if it's always contiguous I feel it might be better to do an assert here

torchao/quantization/quantize_/workflows/float8/float8_opaque_tensor.py
jerryzh168: we should add one of `QuantizeTensorToFloat8Kwargs` for each tensor I think, so should create `QuantizeTensorToOpaqueFloat8Kwargs`

jerryzh168: yeah it's optional, if you feel it's better to not use it, it's OK as well.

main thing is using this will reduce duplication and make it easier to adapt in the future

torchao/testing/model_architectures.py
```diff
@@ -64,12 +64,12 @@ def __init__(
         )
 
     # Note: Tiny-GEMM kernel only uses BF16 inputs
-    def example_inputs(self, batch_size=1):
+    def example_inputs(self, batch_size=1, dtype=None):
```
jerryzh168: when do you use the dtype that's not the same as the original weight dtype?

test/quantization/quantize_/workflows/float8/test_float8_opaque_tensor.py
```diff
@@ -0,0 +1,150 @@
...
+        self, dtype, x_dim, bias, bs, x_granularity, w_granularity
+    ):
+        if isinstance(x_granularity, PerGroup):
+            if not isinstance(w_granularity, PerGroup):
+                return
+            if w_granularity.group_size != x_granularity.group_size:
+                return
+        device = "cpu"
+        m = ToyTwoLinearModel(256, 256, 256, dtype, device, bias).eval()
+        example_inputs = m.example_inputs(batch_size=bs, dtype=dtype)
```
jerryzh168: dtype here seems to be the same as the one taken by model? so you don't need to specify it right?

torchao/quantization/quantize_/workflows/float8/float8_opaque_tensor.py
Xia-Weiwen: I mean in most cases it's contiguous so we don't need to worry about performance. But we cannot guarantee that. It is acceptable for us if the input tensor is not contiguous and it's slow to make it contiguous.

torchao/testing/model_architectures.py
Xia-Weiwen: I have removed this. Thanks.

torchao/quantization/quantize_/workflows/float8/float8_opaque_tensor.py
Xia-Weiwen: Thanks for the suggestion. Since input tensors should be plain, we can reuse `QuantizeTensorToFloat8Kwargs` and `_choose_quant_func_and_quantize_tensor` here. I have updated this part.

torchao/quantization/quant_api.py
Xia-Weiwen: Thanks. Removed.

torchao/quantization/quant_api.py
Xia-Weiwen: Updated. Thanks.

test/quantization/quantize_/workflows/float8/test_float8_opaque_tensor.py
Xia-Weiwen: Updated. Thanks.

Xia-Weiwen:
Hi @jerryzh168 Please review again. Thanks.

torchao/quantization/quantize_/workflows/float8/float8_opaque_tensor.py
jerryzh168: OK sounds good

jerryzh168 (approved):

PR #3080, opened 2025-09-26
mxtensor: switch to AOBaseTensor dispatch

vkuzo:
Summary:

Deletes the custom op dispatch logic in `MXTensor` and switches to the
general one in `AOBaseTensor`.

To enable this, we move the mx ops to the same file as `MXTensor`. This
is to avoid the need for callsites to separately import `mx_ops.py`.

Note that there are a couple of custom mx ops which could instead use
the general implementation, leaving that for future PRs to keep this one
small.

Test Plan:

```
pytest test/prototype/mx_formats -s -x
```

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo:
* #3081
* __->__ #3080

torchao/prototype/mx_formats/mx_tensor.py
```diff
@@ -624,3 +615,294 @@ def to_mx(
 
     # Do not force the MXTensor type on the returned tensor
     __torch_function__ = torch._C._disabled_torch_function_impl
+
+
+implements = MXTensor.implements
+
+
+@implements([aten.detach.default, aten.alias.default])
```
jerryzh168, line 623: these are defined in TorchAOBaseTensor as well I think

vkuzo: yes, we can move in future PRs if needed, I wanted to minimize changes in this PR

Issue #2948, opened 2025-09-05
Deprecation for Int4WeightOnlyConfig (version 1) and the models


jerryzh168:
This issue is tracking the deprecation of the (1) configs (2) model checkpoints quantized with these configs.

What is deprecated:
* We added version 2 Int4WeightOnlyConfig in various PRs in https://github.com/pytorch/ao/issues/2752 and switched the default version to 2 in https://github.com/pytorch/ao/pull/2949, the version 1 config is now deprecated, please use version 2 config to quantize the model
* the quantized checkpoints quantized with version 1 config previously is deprecated as well, and we plan to remove the support to load these checkpoints after pytorch 2.11 release (around 9 months from now)

Timeline:
0.14.0: annouce deprecation for version 1 config
after all tensors are migrated: remove support for version 1 config
after pytorch 2.11 release: remove support for version 1 checkpoints

jerryzh168:
It's deprecated and deleted now

PR #3083, opened 2025-09-26
Add quantize_ nn.Parameter support

jcaip:
This PR adds in support for quantizing `nn.Parameter` to `quantize_`. 

### bc-breaking changes

The top level `quantize_` API has the following bc-breaking changes:

1) Passing in both`filter_fn` and `ModuleFqnToConfig` is no longer supported and will now throw a value error if both are specified. Previously, we would quantize all modules that were both matched by `filter_fn` and specified in `ModuleFqnToConfig`. Users should now manually specify `filter_fn=None` when using `ModuleFqnToConfig`/`FqnToConfig`. 
2) The semantics of `filter_fn=None` have changed. Previously passing in `None` would default to `_is_linear` when running `quantize_`. Now when `filter_fn=None` is specified we ignore `filter_fn` completely and only rely on `FqnToConfig` to quantize the model. Note that this is equivalent to passing in `filter_fn=lambda mod, fqn: True` in the previous API. 
3) The default `filter_fn` has changed from `None` to `_is_linear` and `_default` in `ModuleFqnToConfig` now only applies to linear layers. Previously `_default` would apply to all modules that passed `filter_fn`. We plan to deprecate `_default` in the future, please see #3229 for more details. 

Before:
```python
model = torch.nn.Sequential(
    torch.nn.Linear(128, 128), 
    torch.nn.Linear(128, 128), 
    torch.nn.Conv2d(128, 128, 3, 1, 1), 
).cuda().to(torch.bfloat16)

config = ModuleFqnToConfig({
    "0": Float8DynamicActivationFloat8WeightConfig(), 
})

# these are equivalent
quantize_(model, config, filter_fn=_is_linear)
quantize_(model, config, filter_fn=None)
quantize_(model, config)
```
```
> Sequential(
  (0): Linear(in_features=128, out_features=128, weight=Float8Tensor(self.act_quant_kwargs=QuantizeTensorToFloat8Kwargs(float8_dtype=torch.float8_e4m3fn, granularity=PerTensor(), mm_config=None, hp_value_lb=None, hp_value_ub=None, kernel_preference=<KernelPreference.AUTO: 'auto'>), self.block_size=[128, 128], self.mm_config=Float8MMConfig(emulate=False, use_fast_accum=True, pad_inner_dim=False), self.scale.shape=torch.Size([1, 1]), self.kernel_preference=<KernelPreference.AUTO: 'auto'>))
  (1): Linear(in_features=128, out_features=128, bias=True)
  (2): Conv2d(128, 128, kernel_size=(3, 3), stride=(1, 1), padding=(1, 1))
)
```
After:
```python
# user must specify None
quantize_(model, config, filter_fn=None)
```
```
> Sequential(
  (0): Linear(in_features=128, out_features=128, weight=Float8Tensor(self.act_quant_kwargs=QuantizeTensorToFloat8Kwargs(float8_dtype=torch.float8_e4m3fn, granularity=PerTensor(), mm_config=None, hp_value_lb=None, hp_value_ub=None, kernel_preference=<KernelPreference.AUTO: 'auto'>), self.block_size=[128, 128], self.mm_config=Float8MMConfig(emulate=False, use_fast_accum=True, pad_inner_dim=False), self.scale.shape=torch.Size([1, 1]), self.kernel_preference=<KernelPreference.AUTO: 'auto'>))
  (1): Linear(in_features=128, out_features=128, bias=True)
  (2): Conv2d(128, 128, kernel_size=(3, 3), stride=(1, 1), padding=(1, 1))
)
```
After:
```python
# these now error
quantize_(model, config, filter_fn=_is_linear)
quantize_(model, config)
```
```
> ValueError: Custom filter_fn and FqnToConfig were both specified. Only filter_fn=None is supported when FqnToConfig is specified.
```

#### Example for _default changes:
Before:
```python
class MyModule(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.register_parameter("weight", torch.nn.Parameter(torch.randn(128, 128)))

model = torch.nn.Sequential(
    torch.nn.Linear(128, 128), 
    torch.nn.Linear(128, 128), 
    MyModule(),
).cuda().to(torch.bfloat16)

config = ModuleFqnToConfig({
    "_default": Float8DynamicActivationFloat8WeightConfig(), 
})

quantize_(model, config, filter_fn=lambda mod, fqn: isinstance(mod, torch.nn.Linear) or isinstance(mod, MyModule))
```
```
> Sequential(
  (0): Linear(in_features=128, out_features=128, weight=Float8Tensor(self.act_quant_kwargs=QuantizeTensorToFloat8Kwargs(float8_dtype=torch.float8_e4m3fn, granularity=PerTensor(), mm_config=None, hp_value_lb=None, hp_value_ub=None, kernel_preference=<KernelPreference.AUTO: 'auto'>), self.block_size=[128, 128], self.mm_config=Float8MMConfig(emulate=False, use_fast_accum=True, pad_inner_dim=False), self.scale.shape=torch.Size([1, 1]), self.kernel_preference=<KernelPreference.AUTO: 'auto'>))
  (1): Linear(in_features=128, out_features=128, weight=Float8Tensor(self.act_quant_kwargs=QuantizeTensorToFloat8Kwargs(float8_dtype=torch.float8_e4m3fn, granularity=PerTensor(), mm_config=None, hp_value_lb=None, hp_value_ub=None, kernel_preference=<KernelPreference.AUTO: 'auto'>), self.block_size=[128, 128], self.mm_config=Float8MMConfig(emulate=False, use_fast_accum=True, pad_inner_dim=False), self.scale.shape=torch.Size([1, 1]), self.kernel_preference=<KernelPreference.AUTO: 'auto'>))
  (2): MyModule(in_features=128, out_features=128, weight=Float8Tensor(self.act_quant_kwargs=QuantizeTensorToFloat8Kwargs(float8_dtype=torch.float8_e4m3fn, granularity=PerTensor(), mm_config=None, hp_value_lb=None, hp_value_ub=None, kernel_preference=<KernelPreference.AUTO: 'auto'>), self.block_size=[128, 128], self.mm_config=Float8MMConfig(emulate=False, use_fast_accum=True, pad_inner_dim=False), self.scale.shape=torch.Size([1, 1]), self.kernel_preference=<KernelPreference.AUTO: 'auto'>))
)
```

After:
```python
# only linear is applied for default
quantize_(model, config, filter_fn=None)
```
```
> Sequential(
  (0): Linear(in_features=128, out_features=128, weight=Float8Tensor(self.act_quant_kwargs=QuantizeTensorToFloat8Kwargs(float8_dtype=torch.float8_e4m3fn, granularity=PerTensor(), mm_config=None, hp_value_lb=None, hp_value_ub=None, kernel_preference=<KernelPreference.AUTO: 'auto'>), self.block_size=[128, 128], self.mm_config=Float8MMConfig(emulate=False, use_fast_accum=True, pad_inner_dim=False), self.scale.shape=torch.Size([1, 1]), self.kernel_preference=<KernelPreference.AUTO: 'auto'>))
  (1): Linear(in_features=128, out_features=128, weight=Float8Tensor(self.act_quant_kwargs=QuantizeTensorToFloat8Kwargs(float8_dtype=torch.float8_e4m3fn, granularity=PerTensor(), mm_config=None, hp_value_lb=None, hp_value_ub=None, kernel_preference=<KernelPreference.AUTO: 'auto'>), self.block_size=[128, 128], self.mm_config=Float8MMConfig(emulate=False, use_fast_accum=True, pad_inner_dim=False), self.scale.shape=torch.Size([1, 1]), self.kernel_preference=<KernelPreference.AUTO: 'auto'>))
  (2): MyModule()
)
```

### Summary

`ModuleFqnToConfig` has been renamed to `FqnToConfig`, which now accepts both module fqn and parameter fqns. `ModuleFqnToConfig` has been aliased to maintain BC.  The keys to `FqnToConfig` can be one of the following (in order of precedence):

1) exact parameter FQN
```python
quant_config = FqnToConfig({
    "linear1.weight": Float8DynamicActivationFloat8WeightConfig(
        granularity=PerRow(),
    ),
})
```
2) exact module FQN 
```python
quant_config = FqnToConfig({
    "linear1": Float8DynamicActivationFloat8WeightConfig(
        granularity=PerRow(),
    ),
})
```
3) regex that matches parameter FQN (prepended by `re:`)
```python
quant_config = FqnToConfig({
    "re:linear*.weight": Float8DynamicActivationFloat8WeightConfig(
        granularity=PerRow(),
    ),
})
```
4) regex that matches module FQN (prepended by `re:`)
```python
quant_config = FqnToConfig({
    "re:linear*": Float8DynamicActivationFloat8WeightConfig(
        granularity=PerRow(),
    ),
})
```
5) _default, only applies to `nn.Linear` layers
```python
quant_config = FqnToConfig({
    "_default": Float8DynamicActivationFloat8WeightConfig(
        granularity=PerRow(),
    ),
})
```

To enable support for parameter fqn for a paticular config, we need to add the `parameter_name` kwarg into the config signature, and update `CUSTOM_PARAM_QUANTIZATION_SUPPOTED_CONFIGS`. See the changes [here](https://github.com/pytorch/ao/pull/3083/files#diff-bf4d50867e3d649de2d89146592bf47d2f258c4c19126c8acf0e120ee904b726R1874) for more details. 

`Float8DynamicActivationFloat8WeightConfig` has been enabled by this PR, but other configs will throw an `NotImplementedError`. 

### Test Plan

1) unit tests for new config:
```
pytest test/quantization/test_quant_api.py::TestFqnToConfig
```

2) regression test for ModuleFqnToConfig
```
pytest test/quantization/test_quant_api.py -k test_module_fqn_to_config
```
3) Make sure that we can load old HF checkpoints to maintain BC, run [this](https://huggingface.co/torchao-testing/opt-125m-ModuleFqnToConfig-v1-regex-0.14.0.dev#test-loading)

4) Make sure that this doesn't break BC with transformers
```
pytest tests/quantization/torchao_integration/test_torchao.py -k test_module_fqn_to_config
```

5) make sure that this doesn't break BC in VLLM:
```
pytest tests/quantization/test_torchao.py
```
___

## How do our configs translate for MoEs?

Currently, we define a bunch of configs that are for dense nn.Linear modules, how do these configs translate in the case of MoE inference? 

### Some background on MoE inference
There are two ways that forwards is implemented for MoE

- For loop of `nn.Linear` - In this case, we break down the 3d weight x activation matmul into a for loop of 2d weight x activation matmuls. This can be seen [here](https://github.com/huggingface/transformers/blob/6cade29278c4aee3f174f8950f97a3873bdb212f/src/transformers/models/gpt_oss/modeling_gpt_oss.py#L117). 

**In this case, I argue that the semantics of the configs do not change at all from the normal `nn.Linear` case, as we are just doing a bunch of normal 2d linear matmuls.** 

- bmm/grouped mm on the 3d weights / activations directly. 

**For this case, we'd need to add additional op support (bmm) for forwards. Depending on whether the subclass is an AQT subclass or non AQT subclass this will be added differently.**

I plan to only support parameter quantization for non-AQT subclasses, my reasoning being that those are the most popular / important configs anyway (Float8Dynamic, Int4WeightOnly). 

Below is a breakdown of what Configs map to AQT / non-AQT subclasses:
| not using AQT | AffineQuantizedTensor |
|-----------|---------------|
| Float8DynamicActivationFloat8WeightConfig | FPXWeightOnlyConfig |
| Float8DynamicActivationInt4WeightConfig | Float8WeightOnlyConfig |
| Float8StaticActivationFloat8WeightConfig | Float8DynamicActivationFloat8SemiSparseWeightConfig |
| Int4WeightOnlyConfig (v2) | GemliteUIntXWeightOnlyConfig |
|  | Int4DynamicActivationInt4WeightConfig |
|  | Int8DynamicActivationInt4WeightConfig |
|  | Int8DynamicActivationInt8WeightConfig |
|  | Int8WeightOnlyConfig |
|  | IntxWeightOnlyConfig |
|  | UIntXWeightOnlyConfig |

For these the majority of the semantics remain the same, the only semantics that really changes is `PerRow` granularity. and there's a very natural extension of `PerRow` to the 3d case (apply on the last dimension). 

I took a look at the keys of the non-AQT configs below and what they would mean for MoEs. 

#### Float8DynamicActivationFloat8WeightConfig
```
[('activation_dtype', <class 'torch.dtype'>),
 ('weight_dtype', <class 'torch.dtype'>),
 ('granularity',
  typing.Union[ForwardRef('PerTensor'), ForwardRef('PerRow'), typing.List[typing.Union[ForwardRef('PerTensor'), ForwardRef('PerRow')]], NoneType]),
 ('mm_config', typing.Optional[torchao.float8.inference.Float8MMConfig]),
 ('activation_value_lb', typing.Optional[float]),
 ('activation_value_ub', typing.Optional[float]),
 ('kernel_preference', <enum 'KernelPreference'>),
 ('set_inductor_config', <class 'bool'>),
 ('version', <class 'int'>)]
```

`activation_dtype`, `weight_dtype`, `activation_value_lb`, `activation_value_ub` all do not change meaning semantically. 
`granularity=PerTensor()` does not change semantic meaning - we still use a single tensor to scale the entire weight tensor. 
`granularity=PerRow()` does change meaning - we now calculate a scale for each row for the last dimension [-1] i.e for a weight of (E, N, K) we would expect PerRow to create scales of block size (1, 1, K). 
`mm_config` `kernel_preference` and `set_inductor_config` stay the same as well. 

#### Float8StaticActivationFloat8WeightConfig
```
[('scale', <class 'torch.Tensor'>),
 ('activation_dtype', <class 'torch.dtype'>),
 ('weight_dtype', <class 'torch.dtype'>),
 ('granularity',
  typing.Union[ForwardRef('PerTensor'), ForwardRef('PerRow'), typing.Tuple[typing.Union[ForwardRef('PerTensor'), ForwardRef('PerRow')], typing.Union[ForwardRef('PerTensor'), ForwardRef('PerRow')]], NoneType]),
 ('mm_config', typing.Optional[torchao.float8.inference.Float8MMConfig]),
 ('set_inductor_config', <class 'bool'>)]
 ```
`scale` should be passed in as a 3d tensor instead of a 2d tensor in the case of `PerRow` granularity

#### Float8DynamicActivationInt4WeightConfig
```
[('int4_packing_format', <enum 'Int4PackingFormat'>)]
```

int4_packing_format - Only "preshuffled" is supported and Int4PreshuffledTensor [supports](https://github.com/pytorch/ao/blob/895573980e085b02a2c6abbc82239bae7f1318d6/torchao/quantization/quantize_/workflows/int4/int4_preshuffled_tensor.py#L154) 3d weights. 

#### Int4WeightOnlyConfig
```
[('group_size', <class 'int'>),
 ('layout',
  typing.Optional[torchao.dtypes.uintx.tensor_core_tiled_layout.TensorCoreTiledLayout]),
 ('use_hqq', <class 'bool'>),
 ('zero_point_domain',
  typing.Optional[torchao.quantization.quant_primitives.ZeroPointDomain]),
 ('set_inductor_config', <class 'bool'>),
 ('preserve_zero', typing.Optional[bool]),
 ('int4_packing_format', <enum 'Int4PackingFormat'>),
 ('int4_choose_qparams_algorithm', <enum 'Int4ChooseQParamsAlgorithm'>),
 ('version', <class 'int'>)]
 ```

`group_size`, `int4_packing_format`, `int4_choose_qparams_algorithm`, `set_inductor_config` are the only things that are set for v2 config, 

I don't think these semantics of these change, although there are some packing formats that do not support 3d weights.  It looks like (`Int4PackingFormat.PLAIN_INT32`,  `Int4PackingFormat.MARLIN_SPARSE`).

### DIFF
--- a/test/quantization/test_quant_api.py
+++ b/test/quantization/test_quant_api.py
@@ -714,7 +714,7 @@ class TestQuantFlow(TestCase):
         config1 = Int4WeightOnlyConfig(
             group_size=32, int4_packing_format="tile_packed_to_4d"
         )
-        config2 = Float8DynamicActivationFloat8WeightConfig()
+        config2 = Float8WeightOnlyConfig()
         config = ModuleFqnToConfig(
             {
                 "re:linear.*": config2,


torchao/quantization/quant_api.py
```diff
@@ -2399,6 +2371,36 @@ def _module_fqn_to_config_handler(
     return module
 
 
+@dataclass
```
jcaip, line 2377: I think in the future we can consolidate some of the logic of `ParamFqnToConfig` and `ModuleFqnToConfig`

jerryzh168:
current AOBaseConfig is more for linear weights, can it be extended to param config cleanly?

vkuzo:
> Add in ParamFqnToConfig config
> This new config is very similar to ModuleFqnToConfig except it takes in nn.Parameter FQNs and also supports regexs.

Would it work to stick with `ModuleFqnToConfig` and update its meaning, to avoid introducing a new object with a lot of similarities with the old object?  Pseudocode of what it could do:

```python
def handle_module(model, fqn, config):
    if has_parameter(model, fqn):
        ... new behavior for parameters, apply parameter swap config ...
    elif has_parameter(model, fqn + '.weight'):
        ... old behavior, apply parameter swap config ...
    elif has_module(model, fqn):
        ... old behavior, apply module swap ...
```

jcaip:
> Would it work to stick with ModuleFqnToConfig and update its meaning, to avoid introducing a new object with a lot of similarities with the old object? 

Yeah, we can do this. Do you think we should keep the `ModuleFqnToConfig` name? It's a little confusing I feel to pass in parameter fqn but it's also being used by huggingface and vllm so I think it would be better to keep it as is.

jcaip:
> current AOBaseConfig is more for linear weights, can it be extended to param config cleanly?

Yes I believe so, especially in the case of the Config object itself. We attach everything to the weight parameter for nn.Linear, so this allows us to specify the parameter name instead of assuming it's "weight". 

The only thing that does not map cleanly IMO is the `module_registration`:

```
        # non user facing code
        @register_quantize_module_handler(WorkflowFooConfig)
        def _transform(
            mod: torch.nn.Module,
            config: WorkflowFooConfig,
        ) -> torch.nn.Module:
            # the transform is implemented here, usually a tensor sublass
            # weight swap or a module swap
```

I think we should define the transform for parameters as the base case (aka `@register_quantize_handler`) , and use that for the module flow (assuming the parameter is module.weight), since it's the more general case.

vkuzo:
> Do you think we should keep the ModuleFqnToConfig name? It's a little confusing I feel to pass in parameter fqn but it's also being used by huggingface and vllm so I think it would be better to keep it as is.

IMO we should change the current name and keep the old name for BC:

```python
ParamOrModuleFqnToConfig = ...

# for bc
ModuleFqnToConfig = ParamOrModuleFqnToConfig
```

vkuzo:
> I think we should define the transform for parameters as the base case

To me it seems that the transform has to be for modules, because it is inplace.  User can target a parameter if they want to, but the transform function always runs on a module that owns the parameter.

torchao/quantization/quant_api.py
```diff
@@ -2352,6 +2371,36 @@ def _module_fqn_to_config_handler(
...
+
+
+def _param_fqn_to_config_handler(
+    mod_containing_param: torch.nn.Module, fqn: str, config: ParamFqnToConfig
+):
+    for name, param in list(mod_containing_param.named_parameters()):
+        # skip if not direct child
+        if "." not in name:
+            for pattern in config.param_fqn_to_config:
+                if re.match(pattern, f"{fqn}.{name}"):
```
jerryzh168: so it applies to all params, regardless of what it is? e.g. bias? should we be more specific in what people are configuring?

andrewor14:
Does this mean we need to refactor all supported configs to use this structure?

```
@register_quantized_param_handler(config)
def _float8_dynamic_activation_float8_weight_quantize_tensor(...):
    # returns quantized tensor

def _float8_dynamic_activation_float8_weight_transform(...):
    module.weight = _float8_dynamic_activation_float8_weight_quantize_tensor(...)
    return module
```

torchao/quantization/quant_api.py
```diff
@@ -524,6 +526,23 @@ def quantize_(
     torch._C._log_api_usage_once("torchao.quantization.quantize_")
 
     filter_fn = _is_linear if filter_fn is None else filter_fn
+    if isinstance(model, nn.Parameter):
+        handler = _QUANTIZE_CONFIG_PARAM_HANDLER[type(config)]
```
andrewor14: nicer error message if the config is not registered?

torchao/quantization/quant_api.py
```diff
@@ -524,6 +526,23 @@ def quantize_(
...
+    if isinstance(config, ParamFqnToConfig):
+        _replace_with_custom_fn_if_matches_filter_with_name(
+            model,
+            _param_fqn_to_config_handler,
+            lambda mod, fqn: any(
+                re.match(pattern, f"{fqn}.{name}")
+                for pattern in config.param_fqn_to_config
+                for name, _ in mod.named_parameters()
+                if "." not in name
+            ),
```
andrewor14: does this ignore the `filter_fn` arg then? Do we need special handling if the `filter_fn` clashes with the fqns here?

andrewor14: Also I find this lambda a bit difficult to read with the double for loops. Can we rewrite it as a function for clarity?

torchao/quantization/quant_api.py
```diff
@@ -2399,6 +2371,36 @@ def _module_fqn_to_config_handler(
     return module
 
 
+@dataclass
+class ParamFqnToConfig(AOBaseConfig):
+    """Per param configurations for torchao quantize_ API
+
+    Args:
+        `param_fqn_to_config`: Dict[str, Optional[AOBaseConfig]]: a dictionary from
+         the fully qualified name of the parameter to the AOBaseConfig that we want to apply to that parameter.
```
andrewor14: so is the contract that all `AOBaseConfigs` can be applied to both modules and parameters? If users pass in a config that doesn't have parameter support yet, is there a way to raise a nice error message?

vkuzo:
> One thing I'm wondering is if we can provide a more general API where users don't need to specify the FQNs

I'd note that for the use case of HF model -> quantize_ using the HF integration -> vLLM, we want FQNs because the other libraries think in terms of FQNs, and FQNs are serializeable.  IMO we should either have both FQN to config and filter_fn fully supported, or deprecate filter_fn.

torchao/quantization/quant_api.py
jcaip: I think we should consider the regex syntax separately, I can remove from this PR. 

One thing I would like would be for `quantize_` log the modules/params it's swapping so it's easy to see what the difference is.

jcaip:
>Does this mean we need to refactor all supported configs to use this structure?

I was thinking we can just decorate the `from_hp` classmethod instead of having to refactor everything.

torchao/quantization/quant_api.py
jcaip: > so is the contract that all AOBaseConfigs can be applied to both modules and parameters?

Ideally yes, although I think we would just enable this for non-AQT subclasses from the start. For AQT subclasses / configs we can add functionality after we port the subclass out. 

> If users pass in a config that doesn't have parameter support yet, is there a way to raise a nice error message?

Yeah, I can add this.

torchao/quantization/quant_api.py
```diff
@@ -2381,6 +2398,133 @@ def _fpx_weight_only_transform(
     return module
 
 
+@dataclass
+class ModuleOrParamFqnToConfig(AOBaseConfig):
```
vkuzo: how about just adding the logic to `ModuleFqnToConfig`, as I suggested in one of my previous comments

jcaip: We'll have some merge conflicts with #3084, do you want to land yours first @jerryzh168 and then I can rebase?

jerryzh168: yeah sure, I'll update the PR now

torchao/quantization/quant_api.py
```diff
@@ -2382,27 +2389,140 @@ def _fpx_weight_only_transform(
...
+        isinstance(param, TorchAOBaseTensor)
+        for _, param in top_level_named_parameters_list
+    ):
+        return mod_containing_param
+
+    for name, param in top_level_named_parameters_list:
+        for pattern, param_config in config.module_or_param_fqn_to_config.items():
+            full_param_fqn = f"{fqn}.{name}"
+            if (pattern == full_param_fqn) or (
+                pattern[:3] == "re:" and re.search(pattern[3:], f"{fqn}.{name}")
```
jerryzh168: should we use `re.fullmatch`? since that's the behavior we want right?

jcaip: Yeah I'm onboard, will update this PR.

jerryzh168: I also updated https://github.com/pytorch/ao/pull/3084/files we can align the implementation to check regex as well

torchao/quantization/quant_api.py
vkuzo: ping on my original comment

jcaip: > > Do you think we should keep the ModuleFqnToConfig name? It's a little confusing I feel to pass in parameter fqn but it's also being used by huggingface and vllm so I think it would be better to keep it as is.
> 
> IMO we should change the current name and keep the old name for BC:
> 
> ```python
> ParamOrModuleFqnToConfig = ...
> 
> # for bc
> ModuleFqnToConfig = ParamOrModuleFqnToConfig
> ```

I thought we agreed on renaming `ModueFqnToConfig` but unifying on a single object? Is there another comment you're referring to?

torchao/quantization/quant_api.py
```diff
@@ -2385,42 +2380,157 @@ def _fpx_weight_only_transform(
 
 
 @dataclass
-class ModuleFqnToConfig(AOBaseConfig):
-    """Per module configurations for torchao quantize_ API
+class ModuleOrParamFqnToConfig(AOBaseConfig):
+    """Configuration class for applying different quantization configs to modules or parameters based on their fully qualified names (FQNs).
+
+    This extends the functionality of ModuleFqnToConfig to support parameter-level quantization configurations
```
jerryzh168: nit: comment seems stale

torchao/quantization/quant_api.py
vkuzo: you are right, I misread the code, sorry

vkuzo:
> > Add in ParamFqnToConfig config
> > This new config is very similar to ModuleFqnToConfig except it takes in nn.Parameter FQNs and also supports regexs.
> 
> Would it work to stick with `ModuleFqnToConfig` and update its meaning, to avoid introducing a new object with a lot of similarities with the old object? Pseudocode of what it could do:
> 
> ```python
> def handle_module(model, fqn, config):
>     if has_parameter(model, fqn):
>         ... new behavior for parameters, apply parameter swap config ...
>     elif has_parameter(model, fqn + '.weight'):
>         ... old behavior, apply parameter swap config ...
>     elif has_module(model, fqn):
>         ... old behavior, apply module swap ...
> ```

@jcaip  would this be simpler than having two transform registration systems?

torchao/quantization/quant_api.py
vkuzo: > This PR adds in support for quantizing nn.Parameter to quantize_by adding a new config, ModuleOrParamFqnToConfig. This new config is very similar to ModuleFqnToConfig except it also accepts nn.Parameter FQNs.

I missed it because the PR summary still said we are adding a new object, can we update the summary to reflect the current state of the PR

jcaip:
> > Would it work to stick with `ModuleFqnToConfig` and update its meaning, to avoid introducing a new object with a lot of similarities with the old object? Pseudocode of what it could do:
> > ```python
> > def handle_module(model, fqn, config):
> >     if has_parameter(model, fqn):
> >         ... new behavior for parameters, apply parameter swap config ...
> >     elif has_parameter(model, fqn + '.weight'):
> >         ... old behavior, apply parameter swap config ...
> >     elif has_module(model, fqn):
> >         ... old behavior, apply module swap ...
> > ```
> 
> @jcaip would this be simpler than having two transform registration systems?

cc @vkuzo 

Hmm, I think the pseudocode mentioned here vs the logic in the PR and having two transform registration systems are a bit orthogonal. It's possible to have one registration system with the logic in the PR as well. I'm assuming your main concern is with having two registration systems? Let me know if that's not the case.

IMO it's about the same complexity to have one registration system vs two. My main preference for having two registration systems is that it reduces the amount of work we have to do to enable other Configs for parameter quantization - we just need to add the decorator to our `from_hp` or `from_float` class function. In the case of having a shared registration system, we'd need to modify each existing `transform` [function](https://github.com/pytorch/ao/blob/c96f2dd0c980ad100ea9e7c574dead8f19451955/torchao/quantization/quant_api.py#L651) manually to add non-weight param support.

vkuzo:
> I'm assuming your main concern is with having two registration systems?

yes, and even further IMO we should have a single "modify module inplace" paradigm instead of having one paradigm for modules and one for parameters

> My main preference for having two registration systems is that it reduces the amount of work we have to do to enable other Configs for parameter quantization

IMO we should go for the solution where the resulting code is the simplest, if that involves manual work that seems OK to me, and we can parallelize the conversions if you don't want to do them alone.  Reducing the work to convert but ending up with two systems seems like trading dev time now for increased system complexity later.

jcaip:
OK I'll update the PR to use a single registration system. 

> yes, and even further IMO we should have a single "modify module inplace" paradigm instead of having one paradigm for modules and one for parameters

One thing I want to point out is that it's difficult to supports stuff like our vLLM [integration](https://github.com/vllm-project/vllm/blob/f11e3c516be3d88733ea4b0c79f47e33cc319197/vllm/model_executor/layers/quantization/torchao.py#L192), where we pass in a [parameter](https://github.com/vllm-project/vllm/blob/f11e3c516be3d88733ea4b0c79f47e33cc319197/vllm/model_executor/layers/quantization/torchao.py#L192) that's not tied to any module, with a single "modify module inplace" paradigm.

vkuzo:
> One thing I want to point out is that it's difficult to supports stuff like our vLLM [integration](https://github.com/vllm-project/vllm/blob/f11e3c516be3d88733ea4b0c79f47e33cc319197/vllm/model_executor/layers/quantization/torchao.py#L192), where we pass in a [parameter](https://github.com/vllm-project/vllm/blob/f11e3c516be3d88733ea4b0c79f47e33cc319197/vllm/model_executor/layers/quantization/torchao.py#L192) that's not tied to any module, with a single "modify module inplace" paradigm.

I think "everything is parameters" is also a valid solution, I just don't think we should have both - let's pick one?

torchao/quantization/quant_api.py
```diff
@@ -2385,42 +2382,159 @@ def _fpx_weight_only_transform(
...
+    Users can either explicitly specify a specific FQN or pass in a regex pattern to match FQNs.
+
+    `quantize_` will first try to replace all modules matching the keys of ModuleOrParamFqnToConfig.module_or_param_fqn_to_config,
+    It will then will try to replace any parameters that match the keys, ignoring modules that have already been transformed by the previous flow (modules that contain AOBaseTensor parameters):
 
     Args:
-        `module_fqn_to_config`: typing.OrderedDict[str, Optional[AOBaseConfig]]: an
-         ordered dictionary from
-             (1). fully qualified name (fqn) of module or
+        module_fqn_to_config (OrderedDict[str, Optional[AOBaseConfig]]): An ordered dictionary mapping
```
jerryzh168: nit: use `typing.OrderedDict` since it's different from `collections.OrderedDict`

torchao/quantization/quant_api.py
```diff
@@ -2385,42 +2382,159 @@ def _fpx_weight_only_transform(
...
+    Note:
+        - Only processes top-level parameters (those directly accessible as module attributes)
+        - Skips parameters that are already TorchAOBaseTensor instances to avoid double quantization
+        - Uses the first matching pattern for each parameter
+        - Sets quantized parameters as non-differentiable (requires_grad=False)
+
+    Raises:
+        NotImplementedError: If a configuration type doesn't have a registered parameter handler.
+    """
+    top_level_named_parameters_list = [
```
jerryzh168: nit: is this the same as `list(dict(mod_containing_param.named_parameters()).items())`

torchao/quantization/quant_api.py
```diff
@@ -2385,42 +2382,159 @@ def _fpx_weight_only_transform(
...
+    if any(
+        isinstance(param, TorchAOBaseTensor)
+        for _, param in top_level_named_parameters_list
+    ):
+        return mod_containing_param
+
+    for name, param in top_level_named_parameters_list:
+        for pattern, param_config in config.module_or_param_fqn_to_config.items():
+            full_param_fqn = f"{fqn}.{name}"
+            if (pattern == full_param_fqn) or (
```
jerryzh168: btw, if we want exact match (`==`) to take precedence, I think it has to be a separate check,

```
if pattern == full_param_fqn:
   ...
elif pattern.startswith("re:") and ...:
   ...
```

A test of 
```
model: with linear1 module
config: {"re:linear.*": config1, "linear1": config2}
```
and linear1 should be quantized with config2 instead of config1 should catch it

test/quantization/test_quant_api.py
```diff
@@ -901,5 +904,157 @@ def test_config_deprecation(self):
...
+            .to(torch.bfloat16)
+            .cuda()
+        )
+        quant_config = ModuleOrParamFqnToConfig(
+            {
+                # only this config should be applied, as module fqn takes precedence
+                "0": Float8DynamicActivationFloat8WeightConfig(
+                    granularity=PerRow(),
+                ),
+                "re:.*weight": Float8DynamicActivationFloat8WeightConfig(
```
jerryzh168: we should test the reverse order I think, to make sure `0` takes precedence

test/quantization/test_quant_api.py
```diff
@@ -901,5 +904,157 @@ def test_config_deprecation(self):
...
+                "shared_expert.gate_proj": Float8DynamicActivationFloat8WeightConfig(
+                    granularity=PerTensor(),
+                ),
+            }
+        )
+
+        quantize_(
+            model,
+            quant_config,
+        )
```
jerryzh168, line 896: checks?

torchao/quantization/quant_api.py
jerryzh168: also to correct the naming, we can add a  `module_or_param_fqn_to_config` field and use that for version 2, and go through the normal version update path like other configs as well I think

jerryzh168 (commented):
@jcaip can you add `ModuleOrParamFqnToConfig` to torchao docs as well? I would like to link to it in transformer docs

torchao/quantization/quant_api.py
vkuzo: how about `FqnToConfig`, and assert inline that the thing pointed to by FQN is a module or a parameter, throw an exception on other attributes?  IMO simpler name that will cover the known use cases.

torchao/quantization/quant_api.py
vkuzo: how about just `fqn_to_config`

torchao/quantization/quant_api.py
jcaip: That makes sense to me.

torchao/quantization/quant_api.py
jerryzh168: yeah sounds good

torchao/quantization/quant_api.py
```diff
@@ -528,16 +515,22 @@ def quantize_(
...
-    if isinstance(config, ModuleFqnToConfig):
+    if isinstance(config, ModuleOrParamFqnToConfig):
         _replace_with_custom_fn_if_matches_filter_with_name(
             model,
             _module_fqn_to_config_handler,
             filter_fn,
             device=device,
             extra_args=(config,),
         )
+        _replace_with_custom_fn_if_matches_filter_with_name(
```
jerryzh168: also can we just do a single replacement? that might be simpler

the older functionality is a special case of the new one with param_name="weight" so seems like we can use the same code path for everything?

jcaip: Yeah, before I was doing this in two passes because we were doing `re.search`, so something like `re:linear` would match the fqn of the module, the weight, and the bias. But now that we're using `re.fullmatch` I think we can unify this.

torchao/quantization/quant_api.py
```diff
@@ -2385,42 +2382,159 @@ def _fpx_weight_only_transform(
...
+    * "_default", fallback for **all modules** if no match for all previous keys
+        (Note, when using `_default`, the config is applied to all modules, to apply
+        it to only a subset of modules, e.g. with some types, it's better to filter
+        the modules that we don't want to quantize before hand and configure them to
+        None, e.g. `{"re:.+norm.+": None, "_default": linear_config}`)
+
+    Note:
+        - The order of patterns in the OrderedDict may matter as only the first matching pattern is applied
+        - Parameters that are already TorchAOBaseTensor instances are skipped to avoid double quantization
+        - "_default" is ignored for parameter replacement.
```
jerryzh168, line 2408: OK if we merge the replacement code, I guess this option has to be valid for params as well, or we need to rename this to something else? or just remove this option?

jcaip: This is really default linear, since we assume `filter_fn` is `_is_linear` if not specified, I think it makes the most sense to remove, it seems like users could just use the regex support to apply to all linear explicitly by their FQN.

jerryzh168: yeah makes sense, can you keep BC in this PR, and we can change all the published checkpoints and deprecate this separately

docs/source/torchao_vllm_integration.md
```diff
@@ -54,22 +54,21 @@ assert isinstance(config, AOBaseConfig)
 All quantization configurations inherit from {class}`torchao.core.config.AOBaseConfig`, which provides serialization and validation capabilities.
 ```
 
-(module-level-configuration)=
-### 3. Module-Level Configuration
+(fqn-configuration)=
+### 3. FQN Configuration
 
-For granular control, use `ModuleFqnToConfig`:
```
andrewor14, line 60: Looks like we also document this in serving.md, can you update that doc as well?

test/quantization/test_quant_api.py
```diff
@@ -901,5 +905,242 @@ def test_config_deprecation(self):
...
+
+        assert isinstance(model.experts.gate_up_proj, Float8Tensor)
+        assert isinstance(model.shared_expert.gate_proj.weight, Float8Tensor)
+        assert model.shared_expert.gate_proj.weight.scale.numel() == 1
+
+    def test_quantize_modle_exact_match_preference(self):
```
andrewor14: nit: typo modle

torchao/quantization/quant_api.py
```diff
@@ -528,16 +514,19 @@ def quantize_(
     torch._C._log_api_usage_once("torchao.quantization.quantize_")
 
     filter_fn = _is_linear if filter_fn is None else filter_fn
```
andrewor14: is this default `filter_fn` going to have unexpected consequences if people are using `FqnToConfig`? E.g. let's say someone literally just wants to quantize a very specific parameter:
```
quantize_(model, FqnToConfig({"layers.0.some.parameter": Int4WeightOnlyConfig()}))
```
If I'm reading the code correctly, right now we do the replacement if either (1) we match the filter_fn, or (2) we match the fqn. Would the above unexpectedly quantize all the other linear layers in the model?

torchao/quantization/quant_api.py
```diff
@@ -2385,62 +2380,183 @@ def _fpx_weight_only_transform(
...
 
     Args:
-        `module_fqn_to_config`: typing.OrderedDict[str, Optional[AOBaseConfig]]: an
-         ordered dictionary from
-             (1). fully qualified name (fqn) of module or
+        module_fqn_to_config (OrderedDict[str, Optional[AOBaseConfig]]): An ordered dictionary mapping
+            regex patterns (as strings) to quantization configurations.
+
+            The patterns can be one of the follows:
+             (1). fully qualified name (fqn) of module or paramter or
```
andrewor14: typo: paramter

torchao/quantization/quant_api.py
```diff
@@ -2385,62 +2380,183 @@ def _fpx_weight_only_transform(
...
+    Users can either explicitly specify a specific FQN or pass in a regex pattern to match FQNs.
+
+    `quantize_` will first try to replace all modules matching the keys of FqnToConfig.fqn_to_config,
+    It will then will try to replace any parameters that match the keys, ignoring modules that have already been transformed by the previous flow (modules that contain AOBaseTensor parameters):
 
     Args:
-        `module_fqn_to_config`: typing.OrderedDict[str, Optional[AOBaseConfig]]: an
-         ordered dictionary from
-             (1). fully qualified name (fqn) of module or
+        module_fqn_to_config (OrderedDict[str, Optional[AOBaseConfig]]): An ordered dictionary mapping
```
andrewor14: the docstring still references the old arg name I think

torchao/quantization/quant_api.py
```diff
@@ -2385,62 +2380,183 @@ def _fpx_weight_only_transform(
...
         default_factory=OrderedDict
     )
+    version: int = 1
 
     def __post_init__(self):
-        torch._C._log_api_usage_once("torchao.quantization.ModuleFqnToConfig")
+        torch._C._log_api_usage_once("torchao.quantization.FqnToConfig")
+        if len(self.module_fqn_to_config) > 0 and len(self.fqn_to_config) > 0:
+            warnings.warn(
+                "Both module_fqn_to_config and fqn_to_config are specified, only fqn_to_config will be used"
```
andrewor14: I feel this is going to be a silent error for some users, should we just ban this case for simplicity? It's not for BC

torchao/quantization/quant_api.py
```diff
@@ -2385,62 +2380,183 @@ def _fpx_weight_only_transform(
...
+            )
+        if len(self.module_fqn_to_config) > 0 and len(self.fqn_to_config) == 0:
```
andrewor14, line 2424: nit: if you throw an error above then this can become:
```
if len(self.module_fqn_to_config) > 0:
    assert len(self.fqn_to_config) == 0
    self.fqn_to_config = self.module_fqn_to_config
```
and you don't need the rest of the cases (probably don't need to update `self.module_fqn_to_config` to match `self.fqn_to_config`?)

torchao/quantization/quant_api.py
```diff
@@ -2385,62 +2380,183 @@ def _fpx_weight_only_transform(
...
+            if default is not None:
+                c = default
+    return c
 
-    if c is not None:
-        handler = _QUANTIZE_CONFIG_HANDLER[type(c)]
-        return handler(module, c)
 
-    return module
+def select_module_if_filter_fn_or_contains_params_matching_pattern(
```
andrewor14: private?

torchao/quantization/quant_api.py
jcaip: Yeah, we should just ValueError here.

torchao/quantization/quant_api.py
jcaip: In this case, replacement won't do anything as the other linear layers aren't specified in the config. I can add a test for this though.

torchao/quantization/quant_api.py
```diff
@@ -2385,62 +2380,186 @@ def _fpx_weight_only_transform(
...
+    return module
+
+
+def _get_config_for_fqn(fqn: str, config: FqnToConfig):
+    """Helper function to get the config for a given fqn from an FqnToConfig object.
+
+        Args:
+            fqn (str): The fully qualified name to match against the config patterns.
+            config (FqnToConfig): The FqnToConfig object containing mapping of FQNs or regex patterns to quantization configs.
+    torchao/quantization/quant_api.py
```
andrewor14: remove?

torchao/quantization/quant_api.py
andrewor14: Yeah I think it would be good to verify this, from the code it seems we do the replacement if we match either the `filter_fn` or the `config` (not and). Would also be good to clearly document the semantics of `filter_fn` in the docstring in this case

vkuzo: yeah, I think the semantic should be
* if both fqn_to_config and filter_fn specified, both have to match for config to be applied (AND, not OR)
* else, use whichever one is applied

it seems like we should consider breaking BC here and change the default `filter_fn` to `is_linear`, so that if user passes in `filter_fn == None` then only `fqn_to_config` is applied?

torchao/quantization/quant_api.py
```diff
@@ -2385,64 +2380,195 @@ def _fpx_weight_only_transform(
...
+                continue
+            elif re.fullmatch(maybe_module_or_param_fqn_pattern[3:], fqn):
+                # we'll apply the config for first fully matched pattern
+                c = config.fqn_to_config[maybe_module_or_param_fqn_pattern]
+                found = True
+                break
+    return found, c
+
+
+def _select_module_if_filter_fn_or_contains_params_matching_pattern(
```
vkuzo: IMO should be AND, not OR

torchao/quantization/quant_api.py
```diff
@@ -528,16 +514,19 @@ def quantize_(
...
 
     filter_fn = _is_linear if filter_fn is None else filter_fn
-    if isinstance(config, ModuleFqnToConfig):
+    if isinstance(config, FqnToConfig):
         _replace_with_custom_fn_if_matches_filter_with_name(
             model,
-            _module_fqn_to_config_handler,
-            filter_fn,
+            _fqn_to_config_handler,
+            partial(
```
vkuzo: seems like we are passing one callable and one callable wrapping a callable into a fuction, seems a bit hard to follow.  Have we considered just writing this directly instead?

jcaip: I can write this as a lambda, if that's a bit clearer to you?
```python
lambda mod, fqn: filter_fn(mod, fqn) and select_with_module(mod, fqn, config=config)
```

torchao/quantization/quant_api.py
jcaip: In my mind, if someone specifies a fqn in the config, it's pretty clear that they want to quantize it. So I think AND is kind of a footgun here, especially if the default `filter_fn` is `is_linear`. i.e. First time user wants to quantize a parameter, adds an entry to FqnToConfig, and the new param doesn't get quantized because the default filter_fn is `is_linear`. I guess we can just throw a warning in this instance though. 

cc @jerryzh168 what do you think? I'll defer to whatever's most popular with the team.

vkuzo: No good options, but I think AND is the most correct long term semantic here, paired with changing the default `filter_fn`.

We could also punt and just say that specifying fqn_to_config and non-default filter_fn is not supported, and resolve the part where they are both specified later.

jerryzh168: I think we can remove `filter_fn`, it seems to be confusing and unclear when it's strictly needed. I don't see why the filtering can't happen beforehand when it's strictly necessary, or in the quant functions itself, we skip modules that we don't want to support and put up a warning.

for the current PR, AND does make more sense. 
1. we can have all MoE uses cases specifiy filter_fn=None for now 
2. refactor all use cases to only use fqn, and doing filter_fn beforehand if needed
3. remove `filter_fn` in a separate PR after all refactor is done

jcaip: Sounds good to me, ill update the pr

vkuzo: agreed on removing `filter_fn` longer term

I think it is used pretty widely though, so maybe not in this PR and we do it separately with a proper deprecation?  We can punt in this PR by just throwing an exception if `fqn_to_config` is provided along with a non-default `filter_fn`.

andrewor14: `filter_fn` has a lot of internal uses, and it's how many users apply quantization/QAT to linear and embedding separately today. We should do a careful deprecation of this and make sure existing use cases have a good alternative

vkuzo: @andrewor14 , any thoughts on "We can punt in this PR by just throwing an exception if fqn_to_config is provided along with a non-default filter_fn."?

andrewor14: > We can punt in this PR by just throwing an exception if fqn_to_config is provided along with a non-default filter_fn

Yeah sounds good to me

torchao/quantization/quant_api.py
```diff
@@ -96,42 +96,28 @@
...
-    Int4WeightOnlyGPTQQuantizer,
-)
-from .granularity import (
-    Granularity,
-    PerAxis,
-    PerGroup,
-    PerRow,
-    PerTensor,
-)
+from .GPTQ import Int4WeightOnlyGPTQQuantizer
```
vkuzo: are the style changes intended? If yes, can we separate into a different PR?

jerryzh168:
can you add a BC breaking Notes section since it's breaking BC?

test/quantization/test_quant_api.py
```diff
@@ -841,5 +840,273 @@ def test_config_deprecation(self):
...
+    def test_filter_fn_and_fqn_to_config_error(self):
+        """Test that specifying non-default filter_fn and FqnToConfig raises ValueError."""
+
+        # Create a simple model
+        model = torch.nn.Sequential(torch.nn.Linear(10, 5).cuda().bfloat16())
+
+        # Create config with unsupported parameter handler
+        quant_config = FqnToConfig(
+            {
+                "0.weight": Float8DynamicActivationFloat8WeightConfig(
```
jerryzh168, line 1066: should we check the precendence of `"0"` v.s. `"0.param"` as well? I think `"0.param"` should have precendence over `"0"`

torchao/quantization/quant_api.py
```diff
@@ -2385,62 +2396,200 @@ def _fpx_weight_only_transform(
...
+             (1). fully qualified name (fqn) of module or parameter
              (2). regex of fully qualified name (in python `re` module regex format), should
                   start with prefix "re:" or
              (3). "_default"
-         to the config that we want to apply to the module or None
+         to the config that we want to apply to the module/param or None
 
          Config key ordered by precedence:
            * fully qualified module name, e.g. `language.layers.0.q_proj`
+           * fully qualified parameter name, e.g. `language.layers.0.q_proj.weight`
```
jerryzh168, line 2391: should this have higher precedence over the first one?

torchao/quantization/quant_api.py
```diff
@@ -2385,62 +2396,200 @@ def _fpx_weight_only_transform(
...
            * regex for module names, must start with `re:`, e.g. `re:language\.layers\..+\.q_proj`,
-             whiever regex fully matches the module fqn first will be applied
+             whichever regex fully matches the module fqn first will be applied
              (order of keys for dictionary are kept consistent since we are using OrderedDict)
-           * "_default", fallback for **all modules** if no match for all previous keys
+           * regex for parameter names, must start with `re:`, e.g. `re:language\.layers\..+\.q_proj.weight`.
```
jerryzh168: maybe this one should also have higher precedence over the regex for modules, although not sure if it's easy to implement

torchao/quantization/quant_api.py
```diff
@@ -2385,62 +2396,200 @@ def _fpx_weight_only_transform(
...
+        # This code handles BC compatibility with `ModuleFqnToConfig`. It ensures that `self.module_fqn_to_config` and `self.fqn_to_config` share the same object.
+        self.module_fqn_to_config = self.fqn_to_config
 
-def _module_fqn_to_config_handler(
-    module: torch.nn.Module, module_fqn: str, config: ModuleFqnToConfig
-):
+        # TODO we plan to deprecate `_default later, so raise a warning if we find it passed in`
+        if "_default" in self.fqn_to_config:
+            warnings.warn(
+                "Config Deprecation: _default is deprecated and will no longer be supported in a future release."
```
jerryzh168: nit: we can create an issue to track, similar to https://github.com/pytorch/ao/issues/2948

torchao/quantization/quant_api.py
jcaip: Yeah we can just say param fqn > module_fqn in general. I think it should be easy to change this ontop of the recent changes.

jerryzh168:
also `bc-breaking changes` is still not clear to me, can you list one note for each API call?

Before:
```
API 1
```

After:
```
API1 meaning change or updated way to have the same behavior of API1
```

?

jcaip:
> also `bc-breaking changes` is still not clear to me, can you list one note for each API call?
> 
> Before:
> 
> ```
> API 1
> ```
> 
> After:
> 
> ```
> API1 meaning change or updated way to have the same behavior of API1
> ```
> 
> ?

Yeah will do, sorry I'm still resolving some some bugs with the vllm integration so haven't gotten around to it.

torchao/quantization/quant_api.py
```diff
@@ -1662,22 +1674,31 @@ def _float8_weight_only_quant_tensor(weight, config):
...
+    *,
+    parameter_name: str = "weight",
 ) -> torch.nn.Module:
     if config.set_inductor_config:
         torchao.quantization.utils.recommended_inductor_config_setter()
 
-    assert hasattr(module, "weight"), (
-        "applying int8 weight only quant requires module to have weight attribute"
+    assert hasattr(module, parameter_name), (
+        "applying float8 weight only quant requires module to have {parametr_name} attribute"
```
vkuzo: nit: fix typo, `parameter_name` instead of `parametr_name`

torchao/quantization/quant_api.py
```diff
@@ -2385,64 +2412,211 @@ def _fpx_weight_only_transform(
...
+                "Config Deprecation: _default is deprecated and will no longer be supported in a future release. Please see https://github.com/pytorch/ao/issues/3229 for more details."
+            )
+
+
+# maintain BC
+ModuleFqnToConfig = FqnToConfig
+
+# for now, we need to keep track of what configs support custom param quantization.
+# Once we've updated all the transform functions to take in a custom_param kwarg, we can delete this object and the subsequent check
+CUSTOM_PARAM_QUANTIZATION_SUPPOTED_CONFIGS = {
```
vkuzo: nit: `CUSTOM_PARAM_QUANTIZATION_SUPPOTED_CONFIGS` to `CUSTOM_PARAM_QUANTIZATION_SUPPORTED_CONFIGS`

also, can we add a TODO to migrate the rest of the callsites, and make sure it has an owner and a timeline

test/quantization/test_quant_api.py
```diff
@@ -609,7 +613,7 @@ def test_module_fqn_to_config_default(self):
         config = ModuleFqnToConfig({"_default": config1, "linear2": config2})
         model = ToyLinearModel().cuda().to(dtype=torch.bfloat16)
         example_inputs = model.example_inputs(device="cuda", dtype=torch.bfloat16)
-        quantize_(model, config)
+        quantize_(model, config, filter_fn=None)
```
vkuzo, line 616: if the user is using `AOBaseConfig` which is not an instance of `FQNToConfig`, then there is no BC breaking change, right?

jcaip: Yeah, filter_fn=None will default to _is_linear when it's not an instance of `FQNToConfig`

PR #3084, opened 2025-09-26
Add module fqn regex support for ModuleFqnToConfig

jerryzh168:
Summary:
To simplify the config file for torchao quantized models we want to allow people to configure the ModuleFqnToConfig through regex, e.g. `linear*`, `language.layers.*.gate_proj`

Test Plan:
python test/quantization/test_quant_api.py -k test_module_fqn_to_config_module_name_regex

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/quantization/test_quant_api.py
+++ b/test/quantization/test_quant_api.py
@@ -34,6 +34,7 @@ from torchao.dtypes import (
     TensorCoreTiledLayout,
 )
 from torchao.quantization import (
+    Int4Tensor,
     LinearActivationQuantizedTensor,
     PerGroup,
 )
@@ -67,6 +68,7 @@ from torchao.quantization.subclass import (
 from torchao.quantization.utils import compute_error
 from torchao.testing.utils import skip_if_rocm
 from torchao.utils import (
+    _is_fbgemm_genai_gpu_available,
     is_sm_at_least_89,
     is_sm_at_least_90,
     torch_version_at_least,
@@ -690,6 +692,20 @@ class TestQuantFlow(TestCase):
         assert isinstance(model.linear1.weight._layout, TensorCoreTiledLayout)
         assert isinstance(model.linear2.weight, AffineQuantizedTensor)
         assert isinstance(model.linear2.weight._layout, PlainLayout)
+
+    @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
+    @unittest.skipIf(
+        not _is_fbgemm_genai_gpu_available(), "Requires fbgemm-gpu-genai >= 1.2.0"
+    )
+    def test_module_fqn_to_config_module_name_regex(self):
+        config1 = Int4WeightOnlyConfig(group_size=32)
+        config = ModuleFqnToConfig({"linear*": config1})
+        model = ToyLinearModel().cuda().to(dtype=torch.bfloat16)
+        example_inputs = model.example_inputs(device="cuda", dtype=torch.bfloat16)
+        quantize_(model, config)
+        model(*example_inputs)
+        assert isinstance(model.linear1.weight, Int4Tensor)
+        assert isinstance(model.linear2.weight, Int4Tensor)
 
     def test_module_fqn_to_config_embedding_linear(self):
         weight_dtype = torch.int8
--- a/torchao/quantization/quant_api.py
+++ b/torchao/quantization/quant_api.py
@@ -16,6 +16,7 @@ and mixed GEMM kernels
 """
 
 import logging
+import re
 import types
 import warnings
 from dataclasses import dataclass, field
@@ -2389,8 +2390,14 @@ def _module_fqn_to_config_handler(
         # Maybe: we can add module type specific config in the future, in needed
         c = config.module_fqn_to_config[module_fqn]
     else:
-        # fallback to use default if no module specific config is provided
-        c = config.module_fqn_to_config.get("_default", None)
+        for maybe_module_fqn_pattern in config.module_fqn_to_config:
+            if re.search(maybe_module_fqn_pattern, module_fqn):
+                # we'll apply the first matched pattern
+                c = config.module_fqn_to_config[maybe_module_fqn_pattern]
+                break
+        else:
+            # fallback to use default if no module specific config is provided
+            c = config.module_fqn_to_config.get("_default", None)
 
     if c is not None:
         handler = _QUANTIZE_CONFIG_HANDLER[type(c)]


vkuzo:
does it work for quantizing a HuggingFace model and then running it in vLLM?

jerryzh168:
> does it work for quantizing a HuggingFace model and then running it in vLLM?

not yet, we need follow up changes in these repos to handle this, i.e. https://github.com/huggingface/transformers/blob/071eb5334f5a9ac2c7a13515219be8a272388ec6/src/transformers/quantizers/quantizer_torchao.py#L302 and https://github.com/vllm-project/vllm/blob/8bf8f4582208ac7af230512ff5f3ac1dc36d5222/vllm/model_executor/layers/quantization/torchao.py#L126

vkuzo:
I'd love to see the following in this PR:
1. a user facing description of what the new proposed syntax is.  If this changes the FQN to always be a regex, it would be good to also have either proof that this does not change existing behavior, or a list of what specific existing behavior is changed and validation that changing it and breaking BC is acceptable.
2. draft PRs in any repos affected by this change (huggingface transformers, vLLM) demonstrating that the e2e behavior will work as intended

IMO it's hard to know if this PR is landable without having the two things above

jerryzh168:
@vkuzo just updated the PR to include more docs and details on precedence of configs and please check out the summary for the e2e tests in transformers and vllm. Please take a look again

test/quantization/test_quant_api.py
```diff
@@ -691,6 +694,38 @@ def test_module_fqn_to_config_module_name(self):
         assert isinstance(model.linear2.weight, AffineQuantizedTensor)
         assert isinstance(model.linear2.weight._layout, PlainLayout)
 
+    @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
+    @unittest.skipIf(
```
vkuzo: IMO we should keep basic functionality tests independent from runtime dependencies such as `fbgemm-gpu-genai`.  Can we refactor these test cases to only use PyTorch native functionality?

test/quantization/test_quant_api.py
```diff
@@ -691,6 +694,38 @@ def test_module_fqn_to_config_module_name(self):
...
+        not _is_fbgemm_genai_gpu_available(), "Requires fbgemm-gpu-genai >= 1.2.0"
+    )
+    def test_module_fqn_to_config_precedence(self):
+        """Testing that full path config takes precedence over
+        regex config in ModuleFqnToConfig
+        """
+        config1 = Int4WeightOnlyConfig(group_size=32)
+        config2 = Float8DynamicActivationFloat8WeightConfig()
+        config = ModuleFqnToConfig({"linear1": config1, "linear*": config2})
```
vkuzo: can we also test `linear*` being specified first and `linear1` second

torchao/quantization/quant_api.py
```diff
@@ -2367,10 +2368,19 @@ class ModuleFqnToConfig(AOBaseConfig):
 
     Args:
         `module_fqn_to_config`: Dict[str, Optional[AOBaseConfig]]: a dictionary from
-         the fully qualified name of module to the AOBaseConfig that we want to apply to the module.
-         Also has a special key: "_default", if "_default" is present in the dictionary,
-         the config for "_default" will be applied to all the remaining modules that does not have
-         per module configuration specified.
+             (1). fully qualified name (fqn) of module or
+             (2). regex of fully qualified name or
```
vkuzo: clarify that this is using Python's `re` module regex format

torchao/quantization/quant_api.py
```diff
@@ -2367,10 +2368,19 @@ class ModuleFqnToConfig(AOBaseConfig):
...
+             (3). "_default"
+         to the AOBaseConfig that we want to apply to the module
+
+         Config key ordered by precedence:
+           * fully qualified module name, e.g. `language.layers.0.q_proj`
+           * regex for module names, e.g. `language.layers.*.q_proj`
```
vkuzo: clarify that regexes will be matched in the order that they appear in the dictionary

test/quantization/test_quant_api.py
jerryzh168: sure, that should work as well

torchao/quantization/quant_api.py
jerryzh168: will do, also I think we need to change to OrderedDict to keep the order consistent

jerryzh168:
Updates:
1. addressed comments from @vkuzo 
2. updated type of module_fqn_to_config from Dict to OrderedDict, checked that it still works with https://huggingface.co/torchao-testing/opt-125m-ModuleFqnToConfig-v1-regex-0.14.0.dev#test-loading giving the same outputs

torchao/quantization/quant_api.py
```diff
@@ -2389,8 +2404,14 @@ def _module_fqn_to_config_handler(
         # Maybe: we can add module type specific config in the future, in needed
         c = config.module_fqn_to_config[module_fqn]
     else:
-        # fallback to use default if no module specific config is provided
-        c = config.module_fqn_to_config.get("_default", None)
+        for maybe_module_fqn_pattern in config.module_fqn_to_config:
+            if re.search(maybe_module_fqn_pattern, module_fqn):
```
jcaip: Thoughts on changing this logic from any match in the string, to must match the entire string?

I.e. for some regex `r`, we actually do `re.search(^r$, module_fqn`). Reason I'm proposing this is because for gpt-oss we have `gate_up_proj` and `gate_up_proj_bias`, and only one should be quantized.

jerryzh168: should this be done in the regex itself? like `...gate_up_proj$`

jcaip: yeah I'm just a little worried people will be dumb like me and accidentally quantize both, but im fine with how it is now too.

jcaip (approved):
lgtm, ill do the same logic in #3084

torchao/quantization/quant_api.py
jerryzh168: I could add a note in the doc I guess, although I feel we should not change the meaning of the regex in the code, but leave this to user instead

vkuzo: > Reason I'm proposing this is because for gpt-oss we have gate_up_proj and gate_up_proj_bias, and only one should be quantized.

that's a good point.  If the user specifies `gate_up_proj`, seems like with the current logic it would apply an exact match to `gate_up_proj`, and a regex match to `gate_up_proj_bias`?  That's definitely a gotcha, IMO we should ensure every key is either a regex or an exact match to remove the ambiguity.

jerryzh168: sorry I think I meant to use `fullmatch` function, not `search` (https://docs.python.org/3/library/re.html#search-vs-match)

but discussed offline that we want to go with explicitly calling out regex configs. I plan to use a prefix of `re:`, same as llm-compressor

torchao/quantization/quant_api.py
```diff
@@ -2409,8 +2405,16 @@ def _module_fqn_to_config_handler(
         # Maybe: we can add module type specific config in the future, in needed
         c = config.module_fqn_to_config[module_fqn]
```
vkuzo, line 2406: nit: assert that the pattern does not start with `re:`, for clarity

vkuzo (approved):

vkuzo:
thank you!

torchao/quantization/quant_api.py
jcaip: I can update this in my PR.

PR #3085, opened 2025-09-27
[BE] Remove Float8Linear from quant_api.py

### LINKED ISSUE #3069: Conversion of dense vs MoE order should not matter
Currently, for torchao MoE training APIs, you must convert MoE layers first, then dense layers second. This is because if you convert the nn.Linears into Float8Linear/MXLinear first, there is logic that will *unconvert* them if you apply any other conversion to the model afterwards: https://github.com/pytorch/ao/blob/8c5c33e0410e87d817c0dd154f7699734112ce0b/torchao/quantization/quant_api.py#L202-L207

We should improve this so conversion order does not matter.

danielvegamyhre:
Fixes #3069 

Discussed issue with @vkuzo offline, references to Float8Linear in quant_api.py conversion code are technical debt that can be removed.

### DIFF
--- a/torchao/quantization/quant_api.py
+++ b/torchao/quantization/quant_api.py
@@ -1661,15 +1661,10 @@ def _float8_weight_only_transform(
         "applying int8 weight only quant requires module to have weight attribute"
         + " but {module} does not have one"
     )
-    # If model we're quantizing for inference was trained with torchao float8 training
-    # and checkpointed with the Float8Linears, we need to convert them back to
-    # regular nn.Linears so we can apply inference quantization techniques to them.
+
     if isinstance(module, Float8Linear):
-        with torch.device("meta"):
-            new_module = nn.Linear(module.in_features, module.out_features)
-        new_module.weight = module.weight
-        new_module.bias = module.bias
-        module = new_module
+        module = _unwrap_float8_linear(module)
+
     new_weight = _float8_weight_only_quant_tensor(module.weight, config)
 
     module.weight = torch.nn.Parameter(new_weight, requires_grad=False)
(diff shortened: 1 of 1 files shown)


vkuzo (approved):
lg if CI is green

jerryzh168:
I don't think it's as simple as just removing it? @jainapurva might have more context on why this is needed IIRC

jainapurva:
> I don't think it's as simple as just removing it? @jainapurva might have more context on why this is needed IIRC

This code was added to enable quantization using quantize_ api of torchao's fp8 trained model, which is a Float8Linear Tensor. To remove this, we need to test the following flow: TorchAO's Fp8 pre-trained model (model will have Float8Linear layer), then quantize the model using quantize_ api.

danielvegamyhre:
> > I don't think it's as simple as just removing it? @jainapurva might have more context on why this is needed IIRC
> 
> This code was added to enable quantization using quantize_ api of torchao's fp8 trained model, which is a Float8Linear Tensor. To remove this, we need to test the following flow: TorchAO's Fp8 pre-trained model (model will have Float8Linear layer), then quantize the model using quantize_ api.

The code being removed *dequantizes* the model (swaps Float8Linear -> nn.Linear) using the quantize_ api. Why do we need the quantize_ api to do this?

jainapurva:
The quantize_ api would only worked for linear layers, hence first we used
to convert Float8Linear into Linear layers, and then those linear layers
would be quantized

vkuzo:
It does not make sense to me to put workflow-specific logic in a general utility such as `_replace_with_custom_fn_if_matches_filter`.  IMO we should land this PR to remove the technical debt and re-enable the "quantize a float8 trained model for inferen) logic in a follow-up PR, if needed.

danielvegamyhre:
> It does not make sense to me to put workflow-specific logic in a general utility such as `_replace_with_custom_fn_if_matches_filter`. IMO we should land this PR to remove the technical debt and re-enable the "quantize a float8 trained model for inferen) logic in a follow-up PR, if needed.

I agree with this.... I will look into the test failures and fix before landing

danielvegamyhre:
@vkuzo @jerryzh168 I moved the Float8Linear -> nn.Linear conversion logic from `quantize_(...)` to the fp8/fpx related inference quantization config handlers. The test that was failing is passing locally now. Let me know what you think.

PR #2982, opened 2025-09-11
Update deprecated parameters in Hugging Face library

namgyu-youn:
Summary:
In `from_pretrained()` method in `huggingface/transformers`, `torch_dtype` is deprecated and `dtype` replaces it. To prevent deprecation warnings, this PR replaces `torch_dtype` with `dtype`.

Test plan: CI

Reference: https://github.com/huggingface/transformers/pull/39782

PR #3089, opened 2025-09-28
Static quant support for SmoothQuant

Xia-Weiwen:
**Summary**
This PR adds static quant support for SmoothQuant by adding a new `Int8StaticActivationInt8WeightConfig` configuration. Static quantization will generally have better latency & throughput as it saves the overhead of runtime qparam selection.
In the implementation:
- Activation is only per-tensor quantized to support dynamic shape of activation.
- The `SmoothQuantObserver` returns act scale along with the smoothing factor.
- The act scale of each layer is set to `Int8StaticActivationInt8WeightConfig` for transformation of each linear layer.
  - Note: The `Int8StaticActivationInt8WeightConfig` is not suitable for general static quantization (although it works), users should use PT2E in that case. It's because the act scale for the config are global instead of per-linear-layer, which is the same as `Float8StaticActivationFloat8WeightConfig`

**Test plan**
This PR also updates the test cases for SmoothQuant:
- Support CPU-only environment
- Bug fix: the linear module in the UT is not transformed to SmoothQuantLinear since the linear module itself is the parent module.
- Add outliers to example inputs to simulate the case which SmoothQuant is intended to handle.

```
```

### DIFF
--- a/test/integration/test_integration.py
+++ b/test/integration/test_integration.py
@@ -42,12 +42,14 @@ from torchao.quantization.quant_api import (
     Int4WeightOnlyConfig,
     Int8DynamicActivationInt4WeightConfig,
     Int8DynamicActivationInt8WeightConfig,
+    Int8StaticActivationInt8WeightConfig,
     Int8WeightOnlyConfig,
     _replace_with_custom_fn_if_matches_filter,
     quantize_,
 )
 from torchao.quantization.quant_primitives import (
     MappingType,
+    choose_qparams_affine,
     dequantize_affine,
 )
 from torchao.quantization.smoothquant import (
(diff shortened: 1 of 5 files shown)


jerryzh168:
@Xia-Weiwen I think it's better to wait until the Int8Tensor migration is done

Xia-Weiwen:
> @Xia-Weiwen I think it's better to wait until the Int8Tensor migration is done

Thanks for the info

torchao/prototype/smoothquant/example.py
```diff
@@ -82,23 +85,30 @@ def quantize_and_eval(
...
+    static_quant_act: bool,
+    compile: bool,
 ):
     print(f"Loading model on {device}...")
     torch.manual_seed(34)
     t0 = time.time()
     tokenizer = AutoTokenizer.from_pretrained(model_id)
     model = (
-        AutoModelForCausalLM.from_pretrained(model_id, dtype=torch.bfloat16)
+        AutoModelForCausalLM.from_pretrained(model_id, torch_dtype=torch.bfloat16)
```
namgyu-youn, line 96: nit: `torch_dtype` is deprecated; please check https://github.com/pytorch/ao/pull/2982 for more info

torchao/prototype/smoothquant/example.py
```diff
@@ -155,10 +170,12 @@ def compare_models(
     torch.manual_seed(34)
     tokenizer = AutoTokenizer.from_pretrained(model_id)
     model = (
-        AutoModelForCausalLM.from_pretrained(model_id, dtype=torch.bfloat16)
+        AutoModelForCausalLM.from_pretrained(model_id, torch_dtype=torch.bfloat16)
```
namgyu-youn, line 173: same here

torchao/prototype/smoothquant/example.py
```diff
@@ -167,11 +184,13 @@ def compare_models(
     print("Benchmarking W8A8-dynamic without SmoothQuant...")
     torch.manual_seed(34)
     w8a8_model = (
-        AutoModelForCausalLM.from_pretrained(model_id, dtype=torch.bfloat16)
+        AutoModelForCausalLM.from_pretrained(model_id, torch_dtype=torch.bfloat16)
```
namgyu-youn, line 187: and here :)

torchao/prototype/smoothquant/example.py
```diff
@@ -147,6 +160,8 @@ def compare_models(
     device: str,
     model_save_path: str,
     model_save_hf_hub_path: str,
+    static_quant_act: bool,
+    compile: bool,
```
namgyu-youn, line 164: Could you share result for `torch.compile` with static-quant? Not sure for the reason, but it decreased Token/Sec within dynamic-quant, and discussed to remove at https://github.com/pytorch/ao/pull/2728#discussion_r2307463913 .

Xia-Weiwen:
This PR is out of date. We need to use the new int8 tensor API.

PR #3098, opened 2025-09-29
Patch same metadata

larryliu0820:


torchao/utils.py
```diff
@@ -525,6 +525,8 @@ def _(func, types, args, kwargs):
         )
 
     def _same_metadata(self: TorchAOBaseTensor, src: TorchAOBaseTensor) -> bool:
+        if not (hasattr(self, "tensor_data_names") and hasattr(src, "tensor_data_names")):
```
jerryzh168, line 528: we do have a guard for this already: https://github.com/pytorch/ao/blob/5cbbd73feba9573f4614dbbf5e59338b5feda454/torchao/utils.py#L861 so the problem should be somewhere else I think

PR #3101, opened 2025-09-30
Improve SmoothQuant test cases

Xia-Weiwen:
**Summary**
The test cases for SmoothQuant have the following issues:
- All cases are skipped in CPU-only environments
- The test case `test_smoothquant_accuracy` uses a single linear as the model, but only submodules can be converted to `SmoothQuantObservedLinear`. So, the single linear is never converted and the case does not actually test the results of SmoothQuant linear. When running with `pytest -sv`, the following warning messages are printed:
  - `convert: module is not SmoothQuantObservedLinear, skipping: <class 'torch.nn.modules.linear.Linear'>`

This PR fixes the issues.

**Test plan**

### DIFF
--- a/test/prototype/test_smoothquant.py
+++ b/test/prototype/test_smoothquant.py
@@ -15,8 +15,14 @@ from torchao.prototype.smoothquant import (
 )
 from torchao.prototype.smoothquant.core import SmoothQuantStep
 from torchao.quantization import quantize_
+from torchao.quantization.linear_activation_scale import (
+    WeightTensorWithLinearActivationScaleMetadata,
+)
 from torchao.quantization.quant_api import (
     Int8DynamicActivationInt8WeightConfig,
+)
+from torchao.quantization.utils import (
+    compute_error as SQNR,
 )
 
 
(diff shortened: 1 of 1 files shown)


Xia-Weiwen:
CC @namgyu-youn I cannot add you as a reviewer.

jerryzh168 (approved):
LGTM, thanks for the fix

namgyu-youn (approved):
Injecting outliers looks good to me, thanks for the fix.

Xia-Weiwen:
> Injecting outliers looks good to me, thanks for the fix.

Thanks. Actually, the cases may fail if we don't. That means alpha=0.5 is worse than 0 with some random inputs.

test/prototype/test_smoothquant.py
```diff
@@ -34,16 +40,19 @@ def example_inputs(
...
-        ]
+        # For SmoothQuant tests, we intentionally insert some outliers to input features
+        x = torch.randn(
+            batch_size,
+            sequence_length,
+            self.linear1.in_features,
+            dtype=dtype,
+            device=device,
+        )
+        n_outliers = max(1, int(x.size(-1) * 0.1))
```
namgyu-youn, line 51: 0.1 ratio of outliers are too high I think. Probably < 0.05 ?

test/prototype/test_smoothquant.py
```diff
@@ -34,16 +40,19 @@ def example_inputs(
...
+        # Randomly select outlier features
+        outlier_indices = torch.randperm(x.size(-1))[:n_outliers]
+        x[:, :, outlier_indices] *= 10.0
```
namgyu-youn, line 54: 10 weights for outliers might not be enough for outliers. Could you try > 20 to reproduce outliers?

namgyu-youn (changes requested):
I think we can reproduce outliers by adjusting ratio and its size, please check comments

namgyu-youn:
> Thanks. Actually, the cases may fail if we don't. That means alpha=0.5 is worse than 0 with some random inputs.

Yeah outlier is a key to produce SmoothQuant. Thanks for this change

PR #3103, opened 2025-09-30
[mxfp8 moe training] mxfp8 a2a with d2h sync

danielvegamyhre:
## Summary
- Add differentiable mxfp8 a2a implemented using all_to_all_single functional collective, which requires the caller do a d2h sync to get input_splits/output_splits on the host, as required by the all_to_all_single API.
- This is in contrast to "on device / sync-free"  impl using Triton + Symmetric memory (#3088) which theoretically should achieve better performance in a model like **experimental** DSV3 in torchtitan, which natively supports overallocated symmetric memory buffers for data exchange of `inputs` (fwd) and `grad_output` (bwd), and passes these padded buffers through the rest of the downstream grouped_mm, scatter_add etc ops. This approach is experimental because if there is sufficient expert load skew at some point during the training run and the overallocated sym mem buffs are not large enough, the run will crash.
- Non-experimental titan dsv3/llama4 do NOT use this method, opting instead to do a d2h sync in order to allocate the exact memory needed for incoming tokens. In such an approach, the mxfp8 on device kernel is actually worse, because it must do a d2h sync to do an expensive, synchronizing aten::item (slice op) to get the actual received tokens from the overallocated buffer (both in fwd and bwd). 
- Therefore, we implement this simpler mxfp8 a2a sync kernel, which has the exact same approach as the non-experimental models in titan described above, only we quant/dequant the inputs outputs.

## Test plan
- Added test case: `pytest test/prototype/moe_training/mxfp8/test_mxfp8_a2a.py  -k MXFP8AllToAllVSyncTest` 

## Benchmarks

Baseline: 4916us
Mxfp8 sync a2a: 4207us (~1.17x speedup)

mxfp8 perf is basically the same as bf16 baseline. Looking at traces to find out why, we can see the all_to_alls themselves (fp8 data and e8m0 scales) are 1.51x faster than the bf16 a2a at 4916us vs (3167+81)=3248us, but the quant/dequant ops are consuming the entire 

bf16 trace:
Screenshot 2025-09-30 at 10 57 12 AM

mxfp8 trace:
Screenshot 2025-09-30 at 10 58 46 AM

### DIFF
--- a/test/prototype/moe_training/mxfp8/test_mxfp8_a2a.py
+++ b/test/prototype/moe_training/mxfp8/test_mxfp8_a2a.py
@@ -7,6 +7,7 @@ if not torch.cuda.is_available() or torch.cuda.get_device_capability() != (10, 0
 import torch.distributed as dist
 import torch.distributed._symmetric_memory as symm_mem
 from torch.distributed._functional_collectives import (
+    all_to_all_single,
     all_to_all_single_autograd,
 )
 from torch.nn import functional as F
@@ -23,13 +24,14 @@ from torchao.float8.float8_utils import (
 )
 from torchao.prototype.moe_training.kernels.mxfp8.comms import (
     mxfp8_on_device_all_to_all_v,
+    to_mxfp8_a2a_dequant,
 )
 
 from ..testing_utils import generate_split_sizes
 
 
 @instantiate_parametrized_tests
-class MXFP8AllToAllVTest(MultiProcessTestCase):
+class MXFP8OnDeviceAllToAllVTest(MultiProcessTestCase):
     def setUp(self) -> None:
         super().setUp()
         self._spawn_processes()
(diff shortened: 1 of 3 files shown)


torchao/prototype/moe_training/kernels/mxfp8/comms.py
```diff
@@ -455,3 +454,116 @@ def _exchange_row_offsets(
     output_offset_for_remote_rank = tl.sum(output_split_sizes)
 
     return input_offset_for_remote_rank, output_offset_for_remote_rank, num_rows_to_read
+
+
+class MXFP8AllToAllVSync(torch.autograd.Function):
```
vkuzo: nit: name it something like `ToMXFP8ThenAll2AllThenToBf16` to make it clear what dtype transitions are happening

test/prototype/moe_training/mxfp8/test_mxfp8_a2a.py
```diff
@@ -143,5 +145,126 @@ def test_a2a_fwd_bwd(self):
...
+        try:
+            torch.manual_seed(42 + self.rank)
+            self._init_device()
+
+            group_name = dist.group.WORLD.group_name
+            symm_mem.enable_symm_mem_for_group(group_name)
+
+            tokens_per_ep_rank = 8192
+            dim = 2048
+            input_tensor = torch.randn(
```
vkuzo, line 187: optional: if you quantize this to mxfp8 and then dequantize, you can then test for bitwise equality at the end

vkuzo (approved):

torchao/prototype/moe_training/kernels/mxfp8/comms.py
danielvegamyhre: Makes sense, updated to use more explicit naming. The high precision dtype can be bf16 or fp32 so I used `ToMXFP8AllToAllVDequant`

PR #3104, opened 2025-09-30
Replace param_group quantizer instance with QuantOptimizer attribute

lisjin:
Fix compatibility with open source QAT training/finetuning scripts.

Details: FSDP2 `state_dict` and `load_state_dict` methods will raise errors if param_group contains a quantizer object. (This is necessary for quantizing embeddings differently from other params.) I replaced the `param_group["quantizer"]` key with a `group_quantizer_map` mapping between `regularized_param_groups` indices and quantizers in `QuantOptimizer`.

### DIFF
--- a/test/prototype/test_parq.py
+++ b/test/prototype/test_parq.py
@@ -21,7 +21,6 @@ from torchao.prototype.parq.optim import (
 from torchao.prototype.parq.quant import (
     Int4UnifTorchaoQuantizer,
     LSBQuantizer,
-    Quantizer,
     StretchedIntxWeightConfig,
     StretchedUnifTorchaoQuantizer,
     TernaryUnifQuantizer,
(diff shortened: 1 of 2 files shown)


test/prototype/test_parq.py
```diff
@@ -162,29 +161,29 @@ def build_param_groups(
...
-            }
-        )
+        param_groups.append({"params": params_embed, "quant_bits": 4})
     return param_groups
 
 
+def get_optim_kwargs(base_optimizer, embedding=True, quant_cls=UnifTorchaoQuantizer):
+    optim_kwargs = {}
+    if embedding:
+        group_idx = len(base_optimizer.param_groups) - 2
```
metascroy: Why this?  It seems brittle?

lisjin: I agree—I just changed it to be more generalizable

metascroy (approved):

PR #3108, opened 2025-10-01
make scale shape 2d and match qdata shape in NVFP4Tensor

vkuzo:
Summary:

Makes `NVFP4Tensor._scale_e4m3` shape be consistent with `NVFP4Tensor.shape`.  Specifically:

1. if the scales are not swizzled, contiguous data shape `(M, K)` corresponds to scale shape `(M, K // 16)`
2. if the scales are swizzled, contiguous data shape `(M, K)` corresponds to scale shape `(ceil_div(M, 128) * 32, ceil_div(K, 64) * 16)`

If we transpose axes 0 and 1, both qdata and scale now get transposed.

I want this because we need to reason about scales when combining 2d MoE weights into 3d weights, and having the scale shape match the data will make that reasoning easier (in a future PR).

Test Plan:

tests pass on a B200:

```bash
```

also, a dense model quantized with torchao nvfp4 runs in vLLM before and after this PR

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/prototype/mx_formats/test_nvfp4_tensor.py
+++ b/test/prototype/mx_formats/test_nvfp4_tensor.py
@@ -21,6 +21,7 @@ from torchao.prototype.mx_formats.nvfp4_tensor import (
     per_tensor_amax_to_scale,
     unpack_uint4,
 )
+from torchao.prototype.mx_formats.utils import ceil_div
 from torchao.quantization.utils import compute_error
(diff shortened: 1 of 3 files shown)


vkuzo:
* #3109
* __->__ #3108

vkuzo:
need to fix swizzle slicing tests

PR #3110, opened 2025-10-01
Add scale-only version of the HQQ algorithm for IntxWeightOnlyConfig/Int8DynamicActivationIntxWeightConfig

metascroy:
This PR introduces a scale-only version of the HQQ algorithm for  IntxWeightOnlyConfig/Int8DynamicActivationIntxWeightConfig, which we find improves model quality on mmlu from 0.50 to 0.55 when quantizing Gemma3-4B.

### DIFF
--- a/torchao/quantization/qat/api.py
+++ b/torchao/quantization/qat/api.py
@@ -4,6 +4,7 @@
 # This source code is licensed under the license found in the
 # LICENSE file in the root directory of this source tree.
 
+import copy
 from dataclasses import dataclass
 from enum import Enum
 from typing import Any, List, Optional, Tuple
@@ -232,6 +233,7 @@ def _qat_config_transform(
         # Optionally pass custom scales and zero points to base config handler
         # This is only for range learning and only applies to weights
         kwargs = {}
+        has_custom_scale_and_zero_point = False
         weight_config = module.weight_fake_quantizer.config
         if (
             isinstance(weight_config, IntxFakeQuantizeConfig)
@@ -239,6 +241,7 @@ def _qat_config_transform(
         ):
             kwargs["custom_scale"] = module.weight_fake_quantizer.scale
             kwargs["custom_zero_point"] = module.weight_fake_quantizer.zero_point
+            has_custom_scale_and_zero_point = True
 
         # Swap FakeQuantizedLinear -> nn.Linear
         # Swap FakeQuantizedEmbedding -> nn.Embedding
@@ -253,6 +256,12 @@ def _qat_config_transform(
                 f"Encountered unexpected module {module}, should never happen"
             )
         if base_config is not None:
+            # If passing custom scales and zero points, we need to disable the choose_qparam_algorithm on the config
+            if has_custom_scale_and_zero_point and hasattr(
+                base_config, "intx_choose_qparams_algorithm"
+            ):
+                base_config = copy.deepcopy(base_config)
+                base_config.intx_choose_qparams_algorithm = None
             return _QUANTIZE_CONFIG_HANDLER[type(base_config)](
                 module, base_config, **kwargs
             )
--- a/torchao/quantization/quant_api.py
+++ b/torchao/quantization/quant_api.py
@@ -78,6 +78,7 @@ from torchao.quantization.quantize_.workflows import (
     Int4PreshuffledTensor,
     Int4Tensor,
     Int4TilePackedTo4dTensor,
+    IntxChooseQParamsAlgorithm,
     IntxOpaqueTensor,
     IntxPackingFormat,
     IntxUnpackedToInt8Tensor,
@@ -748,6 +749,7 @@ class Int8DynamicActivationIntxWeightConfig(AOBaseConfig):
         `intx_packing_format`: The format to use for the packed weight tensor (version 2 only).
             - unpacked_to_int8: this format is the default and is intended for export applications like ExecuTorch.
             - opaque_torchao_auto: this format is optimized for CPU performance.
+        `intx_choose_qparams_algorithm`: The algorithm to use for choosing the quantization parameters.
         `version`: version of the config to use, only subset of above args are valid based on version, see note for more details.
 
         Note:
@@ -766,6 +768,9 @@ class Int8DynamicActivationIntxWeightConfig(AOBaseConfig):
     act_mapping_type: MappingType = MappingType.ASYMMETRIC
     layout: Layout = QDQLayout()
     intx_packing_format: IntxPackingFormat = IntxPackingFormat.UNPACKED_TO_INT8
+    intx_choose_qparams_algorithm: IntxChooseQParamsAlgorithm = (
+        IntxChooseQParamsAlgorithm.AFFINE
+    )
 
     version: int = 2
 
@@ -830,6 +835,7 @@ def _int8_dynamic_activation_intx_weight_quantize_tensor(
     act_mapping_type = config.act_mapping_type
     layout = config.layout
     intx_packing_format = config.intx_packing_format
+    intx_choose_qparams_algorithm = config.intx_choose_qparams_algorithm
 
     assert weight.dim() == 2, (
         f"Int8DynamicActivationIntxWeightConfig only works for 2-d Tensor, got: {weight.dim()}"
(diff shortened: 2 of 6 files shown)


jerryzh168:
Thanks! Add some unit tests for this path as well?

torchao/quantization/quant_api.py
```diff
@@ -2202,6 +2216,7 @@ def __post_init__(self):
         assert self.mapping_type in [
             MappingType.ASYMMETRIC,
             MappingType.SYMMETRIC,
+            MappingType.SYMMETRIC_NO_CLIPPING_ERR,
```
mergennachin, line 2219: need to update the assert error message

metascroy:
> Thanks! Add some unit tests for this path as well?

Added new E2E test

torchao/quantization/quantize_/workflows/intx/intx_choose_qparams_algorithm.py
```diff
@@ -0,0 +1,23 @@
...
+
+    """
+    Uses `torchao.quantization.quant_primitives.choose_qparams_affine`
+    """
+    AFFINE = "affine"
+
+    """
+    Uses `torchao.quantization.quant_primitives._choose_qparams_and_quantize_scale_only_hqq`
+    """
+    HQQ = "hqq"
```
mergennachin: There are two variants of HQQ today, which makes it confusing:

_choose_qparams_and_quantize_affine_hqq -- which is used in version 1 of the config when use_hqq=True

_choose_qparams_and_quantize_scale_only_hqq - this one

torchao/quantization/quant_primitives.py
```diff
@@ -2125,6 +2126,99 @@ def _choose_qparams_and_quantize_affine_hqq(
...
+    for _ in range(max(1, iters)):
+        # Quantize using current scale
+        # Qg.shape = [n, n_groups, group_size]
+        Qg = _r(Wg / scale.unsqueeze(-1)).clamp(qmin, qmax)
+
+        # Solve least-square problem min_{s} ||Wg - s * Qg||^2 and project
+        # solution onto positive space, or take previous value
+        num = (Wg * Qg).sum(dim=2, dtype=torch.float32)  # [n, n_groups]
+        den = (Qg * Qg).sum(dim=2, dtype=torch.float32)  # [n, n_groups]
+        scale = torch.where(den > 0, num / den, prev_scale)
```
mergennachin, line 2200: can `den` be really really small? dividing num by den may result really big scales

torchao/quantization/qat/api.py
```diff
@@ -270,6 +256,12 @@ def _qat_config_transform(
                 f"Encountered unexpected module {module}, should never happen"
             )
         if base_config is not None:
+            # If passing custom scales and zero points, we need to disable the choose_qparam_algorithm on the config
+            if has_custom_scale_and_zero_point and hasattr(
+                base_config, "intx_choose_qparams_algorithm"
+            ):
+                base_config = copy.deepcopy(base_config)
+                base_config.intx_choose_qparams_algorithm = None
```
mergennachin, line 266: Can you print a debug log that it is disabled?

torchao/quantization/qat/api.py
```diff
@@ -270,6 +256,12 @@ def _qat_config_transform(
                 f"Encountered unexpected module {module}, should never happen"
             )
         if base_config is not None:
+            # If passing custom scales and zero points, we need to disable the choose_qparam_algorithm on the config
+            if has_custom_scale_and_zero_point and hasattr(
+                base_config, "intx_choose_qparams_algorithm"
+            ):
```
mergennachin, line 263: do you wanna specifically check for Int8DynamicActivationIntxWeightConfig and IntxWeightOnlyConfig types only?

torchao/quantization/quantize_/workflows/intx/intx_choose_qparams_algorithm.py
scottroy: These configs are used in ExecuTorch, and never supported HQQ in any previous version.

use_hqq was used in version 1 of the int4 server config.

If you think it would be clearer, I could call the algorithm enum HQQ_SCALE_ONLY or HQQ_NO_ZERO_POINT

torchao/quantization/quant_primitives.py
scottroy: den is a sum of square integers, so it will be at least 1 if positive.

torchao/quantization/qat/api.py
scottroy: I originally did this, but it causes a circular import error because the file that defines the configs also imports from this file.

torchao/quantization/quantize_/workflows/intx/intx_choose_qparams_algorithm.py
mergennachin: > If you think it would be clearer, I could call the algorithm enum HQQ_SCALE_ONLY or HQQ_NO_ZERO_POINT

Yeah, let's do HQQ_SCALE_ONLY then

mergennachin:
This is great btw

Why would one ever not do HQQ_scale_only on XNNPACK on ET if it improves the accuracy compared to naive? It doesn't take a long time to quantize, right? Roughly a few minutes?

Hope we can make it to the 0.14 branch cut, which is Oct 6.

test/quantization/quantize_/workflows/intx/test_intx_unpacked_to_int8_tensor.py
```diff
@@ -61,6 +61,22 @@ def test_linear(self):
         error = compute_error(original, quantized)
         self.assertTrue(error > 20)
 
+    def test_hqq(self):
+        dtype = torch.bfloat16
+        device = "cpu"
+        config = IntxWeightOnlyConfig(
```
mergennachin, line 67: Also another test for Int8DynamicActivationIntxWeightConfig too

torchao/quantization/quantize_/workflows/intx/intx_choose_qparams_algorithm.py
```diff
@@ -0,0 +1,23 @@
...
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+
+from enum import Enum
+
+
+# can switch to StrEnum (https://docs.python.org/3/library/enum.html#enum.StrEnum)
+# after python 3.10 is end of life (https://devguide.python.org/versions/)
+class IntxChooseQParamsAlgorithm(str, Enum):
```
mergennachin, line 12: Also, do we need to introduce yet another class? 

Can we just extend existing Int4ChooseQParamsAlgorithm and add affine and hqq_scale_only? 

And then rename/promote Int4ChooseQParamsAlgorithm to IntxChooseQParamsAlgorithm in a follow-up PR?

metascroy:
> This is great btw
> 
> Why would one ever not do HQQ_scale_only on XNNPACK on ET if it improves the accuracy compared to naive? It doesn't take a long time to quantize, right? Roughly a few minutes?
> 
> Hope we can make it to the 0.14 branch cut, which is Oct 6.

Yes, I want to measure on some more models and benchmarks and if results are neutral to positive, I'll update the default in etLLM.

torchao/quantization/quantize_/workflows/intx/intx_choose_qparams_algorithm.py
metascroy: In torchao's refactor (removing AffineQuantizedTensor), the direction is for subclasses to not share higher-level abstractions, but instead define their own enums.  This is how packing format works as well (intx_packing_format for the intx subclass, and int4_packing_format for the int4 subclass).

I'll let @jerryzh168 comment here as well

mergennachin: Will defer to @jerryzh168 then

jerryzh168: yeah we want local abstractions instead of global abstractions unless it's required.

torchao/quantization/qat/api.py
jerryzh168, line 1: btw, are the changes in this file tested as well?

metascroy: I think this exisiting unit test covers them: https://github.com/pytorch/ao/blob/main/test/quantization/test_qat.py#L2321

torchao/quantization/qat/api.py
jerryzh168: oh so the new algorithm is not supported, should you error out instead when user specifies the hqq algorithm instead of silent ignore?

scottroy: This is not silent ignore.  This is used in QAT range learning, where the scale/zero_point are passed when constructing the subclass.  But I changed the subclass to require you specify choose algorithm as None if you pass in scale/zero.

Issue #2745, opened 2025-08-12
Deprecate and remove subclass.py, dynamic_quant.py, and weight_only.py


andrewor14:
These 3 files contain old quantization flows that are not really supported, but they are not even deprecated as of this issue:

- [ ] Deprecate [subclass.py](https://github.com/pytorch/ao/blob/6cfa47705f60ea614695b52b4b120ac5fd84d1cb/torchao/quantization/subclass.py)
- [ ] Deprecate [dynamic_quant.py](https://github.com/pytorch/ao/blob/6cfa47705f60ea614695b52b4b120ac5fd84d1cb/torchao/quantization/dynamic_quant.py)
- [ ] Deprecate [weight_only.py](https://github.com/pytorch/ao/blob/6cfa47705f60ea614695b52b4b120ac5fd84d1cb/torchao/quantization/weight_only.py)
- [ ] Remove subclass.py
- [ ] Remove dynamic_quant.py
- [ ] Remove weight_only.py

pengu2004:
Hey, I'm new here and would love to do this, is that okay? @andrewor14

andrewor14:
@pengu2004 Yeah for sure, please go ahead

namgyu-youn:
@andrewor14 hi, opened PR for resolving this issue. Please check https://github.com/pytorch/ao/pull/3115

Issue #1639, opened 2025-01-29
make smoothquant more PT2 friendly


vkuzo:
torchao's smoothquant recently broke after a change to PyTorch core: https://github.com/pytorch/pytorch/issues/145733 .  We should make the updates suggested by @anijain2305 in that issue to our code.  I actually think we should go a bit farther and go with something like

```
#
# before
#
class _ActQuantizer:
    def __init__(self, target_dtype, quant_min=-127):
        self.target_dtype = target_dtype
        self.quant_min = quant_min

    def dynamic_quantize(self, input):
        return to_affine_quantized_intx(
            input,
            MappingType.SYMMETRIC,
            _get_per_token_block_size(input),
            self.target_dtype,
            self.quant_min,
        )

    def static_quantize(self, input, scale, zero_point):
        return to_affine_quantized_intx_static(
            input,
            scale,
            zero_point,
            list(input.shape),
            self.target_dtype,
            self.quant_min,
        )

#
# after
#
@dataclass
class _ActQuantConfig:
    target_dtype: torch.dtype
    quant_min: int = -127

# then, logic elsewhere chooses whether to call static or dynamic quant based on the contents of an instance of `_ActQuantConfig`
``` 

My feedback here is similar in spirit to https://github.com/pytorch/ao/pull/1595 - IMO it's simpler and safer to pass around dumb config objects and use them to choose which function to call, instead of encoding the "which function to call" information in the config as a callable object.

vkuzo:
we should also unpin the nightlies (undo most of https://github.com/pytorch/ao/pull/1608) when this is fixed

namgyu-youn:
@vkuzo Looks like this issue was resolved by https://github.com/pytorch/ao/pull/2728 ? The linked PR introduced new SmoothQuant workflows with configuration and quantization API call.

vkuzo:
nice, you can try seeing if the TODO in https://github.com/pytorch/ao/blob/afd6096931c793dc63e9417e2e78f671edc35398/test/prototype/test_smoothquant.py#L83 works with recent code

namgyu-youn:
> nice, you can try seeing if the TODO in
> 
> [ao/test/prototype/test_smoothquant.py](https://github.com/pytorch/ao/blob/afd6096931c793dc63e9417e2e78f671edc35398/test/prototype/test_smoothquant.py#L83)
> 
[...]

PR #3115, opened 2025-10-02
Drop old quantization flows

namgyu-youn:
**Summary**: 
This PR drops the following old quantization flows:
```
torchao/quantization/subclass.py
torchao/quantization/weight_only.py
torchao/quantization/dynamic_quant.py
```

Applies corresponding change in other parts for removing old quantization flows:
```bash
torchao/quantization/quant_api.py
torchao/quantization/utils.py
torchao/quantization/autoquant.py

# Remove old test cases
test/integration/test_integration.py

# SmoothQuant with old workflow is dropped
# Use torchao/prototype/smoothquant instead with new api
torchao/quantization/smoothquant.py
```

- fix: #2745 

**Test plan**: test/integration/test_integration.py


benchmarks/benchmark_aq.py
namgyu-youn, line 1: How about inlining `_int8wo_api`, `_int8da_int8w_api`, `_int4wo_api` ? They are used only once across codebase.

benchmarks/benchmark_aq.py
```diff
@@ -186,20 +174,16 @@ def _bench_quantized_tensor_subclass_perf(api, ref_api, M, N, K, kwargs=None):
 
     for M, N, K in all_shapes:
         _bench_quantized_tensor_subclass_perf(
```
namgyu-youn, line 110: Temporarily updated to use new APIs 2 times to fix CI, but maybe we can update `_bench_quantized_tensor_subclass_perf` to compare only original vs. new quantization flows?

benchmarks/benchmark_aq.py
andrewor14: yeah I think that's fine if they're only used in benchmarks

torchao/quantization/smoothquant.py
```diff
@@ -1,266 +0,0 @@
-# Copyright (c) Meta Platforms, Inc. and affiliates.
```
andrewor14, line 1: Hey @namgyu-youn I don't think we want to remove smoothquant? This wasn't part of the issue: https://github.com/pytorch/ao/issues/2745. Can you add these back?

test/integration/test_integration.py
```diff
@@ -210,262 +198,6 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
-class SmoothquantUnitTest(unittest.TestCase):
```
andrewor14, line 213: Same here, please add the smoothquant tests back

torchao/prototype/quantization/autoquant_v2.py
```diff
@@ -40,9 +40,7 @@
     MappingType,
     ZeroPointDomain,
 )
-from torchao.quantization.subclass import (  # noqa
-    Int8DynamicallyQuantizedLinearWeight,
-    Int8WeightOnlyQuantizedLinearWeight,
+from torchao.quantization.subclass import (
     QuantizedLinearWeightBase,
```
andrewor14: didn't we remove this class in this PR? Seems we need to delete this import completely?

torchao/quantization/utils.py
```diff
@@ -272,37 +250,6 @@ def _quant_int8_per_token_matmul(
     return y
 
 
-def dynamically_quantize_per_channel(x, quant_min, quant_max, target_dtype):
```
andrewor14, line 275: Hey @namgyu-youn I don't think this is part of the scope, can you add this function back? The issue https://github.com/pytorch/ao/issues/2745 is only referring to everything in `subclass.py`, `dynamic_quant.py`, and `weight_only.py`

andrewor14 (commented):
Hi @namgyu-youn, thanks for working on this. I think it looks good overall, but seems like we removed some things outside the scope of https://github.com/pytorch/ao/issues/2745, like smoothquant. Can you please add these back?

benchmarks/benchmark_aq.py
andrewor14: also cc @jainapurva, can you take a look at the benchmark changes?

test/integration/test_integration.py
namgyu-youn: In SmoothQuant API side, they (`torchao/quantization/smoothquant.py` & `torchao/quantization/prototype/smoothquant/`) are just a two different implementations, so we can revert them. 

But in tests side, because we dropped

```
from torchao.quantization.subclass import (
    Int4WeightOnlyQuantizedLinearWeight,
    Int8DynamicallyQuantizedLinearWeight,
    Int8WeightOnlyQuantizedLinearWeight,
)
```
, 
tests (`SmoothquantUnitTest`) can't be maintained. So my suggestion is, how about dropping the old SmoothQuant API also in this PR? The new API also resolved https://github.com/pytorch/ao/issues/1639 and has better structure I think

torchao/quantization/utils.py
namgyu-youn: Reverted; it was misunderstanding while checking api structure.

torchao/prototype/quantization/autoquant_v2.py
namgyu-youn: Done, thanks for pointing it out

test/integration/test_integration.py
andrewor14: oh I see, yeah if the test uses the old APIs then maybe we can just remove the test

andrewor14: hmm actually I misunderstood, I think we should decide what to do with smoothquant separately. Seems like it is orthogonal to the 3 files that we're deleting, and I don't think it's using any of the other deleted APIs

andrewor14:
Hey @namgyu-youn I see smoothquant is still being deleted in this PR. I think we should treat that as a separate issue since we have two smoothquant implementations and it's not clear how we want to reconcile these two yet. As far as I understand smoothquant is actually not related to https://github.com/pytorch/ao/issues/2745. Can we scope our changes in this PR to that issue?

namgyu-youn:
> Hey @namgyu-youn I see smoothquant is still being deleted in this PR. I think we should treat that as a separate issue since we have two smoothquant implementations and it's not clear how we want to reconcile these two yet. As far as I understand smoothquant is actually not related to #2745. Can we scope our changes in this PR to that issue?

@andrewor14 Understood, addressing it separately looks better I feel, so reverted SmoothQuant api at https://github.com/pytorch/ao/pull/3115/commits/6ef8503c4fbbd2bbf75113d60d18369ae6c0be88 .

test/integration/test_integration.py
andrewor14, line 213: @namgyu-youn Thanks for adding back smoothquant. I think now we can add back this test too?

namgyu-youn: Please check https://github.com/pytorch/ao/pull/3115#discussion_r2411215582 again. We should drop this test or reproduce with new quantization APIs.

andrewor14: Can you explain which part of the smoothquant test touches the removed API? I took a quick look and didn't find any references to the removed APIs

namgyu-youn: It's because we drop functions in `torchao/quantization/subclass.py`: `Int4WeightOnlyQuantizedLinearWeight`, `Int8DynamicallyQuantizedLinearWeight`, `Int8WeightOnlyQuantizedLinearWeight`

and `SmoothquantUnitTest` depend on those functions

andrewor14: I searched the uses of these 3 classes:
```
Int4WeightOnlyQuantizedLinearWeight
Int8DynamicallyQuantizedLinearWeight
Int8WeightOnlyQuantizedLinearWeight
```
None of them seem to show up in `SmoothquantUnitTest`. Seems like they're only used in `TestSubclass`. E.g. this smoothquant test doesn't use it, so we shouldn't delete it:
https://github.com/pytorch/ao/blob/main/test/integration/test_integration.py#L216-L232

Can you verify again?

namgyu-youn: Oh sorry, you are right, the smoothquant test is unrelated. Reverted at https://github.com/pytorch/ao/pull/3115/commits/4651e63e7357ee8c7902a4b0710d6ceaf98c8007; sorry for that again.

torchao/quantization/__init__.py
```diff
@@ -100,21 +100,11 @@
...
     IntxUnpackedToInt8Tensor,
 )
-from .smoothquant import (
-    SmoothFakeDynamicallyQuantizedLinear,
-    SmoothFakeDynQuantMixin,
-    get_scale,
-    set_smooth_fq_attribute,
-    smooth_fq_linear_to_inference,
-    swap_linear_with_smooth_fq_linear,
-)
```
andrewor14, line 110: Hi @namgyu-youn, looks like there's more we need to add back here

torchao/quantization/__init__.py
```diff
@@ -172,13 +162,6 @@
     "Int4TilePackedTo4dTensor",
     "Float8Tensor",
     "Int4OpaqueTensor",
-    # smooth quant - subject to change
-    "get_scale",
-    "SmoothFakeDynQuantMixin",
-    "SmoothFakeDynamicallyQuantizedLinear",
-    "swap_linear_with_smooth_fq_linear",
-    "smooth_fq_linear_to_inference",
-    "set_smooth_fq_attribute",
```
andrewor14, line 181: and here

andrewor14:
Looks great, just 2 more places where we need add back smoothquant. Once you do that I think we can go ahead and merge this

test/integration/test_integration.py
```diff
@@ -1324,30 +1200,6 @@ def test_save_load_int4woqtensors(self, device, dtype):
         self._test_handle_save_load_meta_impl(_int4wo_api, device, 20, test_dtype=dtype)
 
 
-class TorchCompileUnitTest(unittest.TestCase):
```
andrewor14, line 1327: Hey @namgyu-youn sorry I found another one. Can you do another search for your diff to see if there's more smoothquant changes to revert? Is this the last one?

PR #3117, opened 2025-10-02
enable select for NVFP4Tensor

vkuzo:
Summary:

This is useful for vLLM 2d -> 3d MoE weight surgery

Test Plan:

unit tests:
```
```

Also, after this PR we can run a Qwen 1.5 MoE model quantized with nvfp4
in vLLM, with https://github.com/vllm-project/vllm/pull/25480

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo:
* __->__ #3117

torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -634,6 +634,25 @@ def nvfp4_view_op(func, types, args, kwargs):
...
 
+@implements([aten.select.int])
+def nvfp4_select(func, types, args, kwargs):
+    old, dim, index = args
+    assert dim == 0, f"NVFP4Tensor aten.select.int with {dim=} is not yet supported"
+    assert len(old.qdata.shape) == len(old._scale_e4m3.shape), "unsupported"
+    new = old.__class__(
+        old.qdata[index],
+        old._scale_e4m3[index],
+        old._block_size,
```
jcaip, line 645: should a dim get knocked off block size after you select?

vkuzo: currently `block_size` is an integer for this tensor, 16 for NVFP4.  If we change it to a multidimensional block, we'd have to update this code.

PR #3128, opened 2025-10-07
[mxfp8 moe training] add triton kernel for mxfp8 quantization along dim0

danielvegamyhre:
[mxfp8 moe training] add triton kernel for mxfp8 quantization along dim0


torchao/prototype/mx_formats/kernels.py
```diff
@@ -1038,6 +1040,175 @@ def to_mxfp8_dim1_kernel(
...
 
+    @triton.autotune(
+        configs=_get_mxfp8_dim1_kernel_autotune_configs(),
+        key=["n_rows", "n_cols", "INNER_BLOCK_SIZE"],
+    )
+    @triton.jit
+    def to_mxfp8_dim0_kernel(
+        x_ptr,  # pointer to input tensor
+        output_ptr,  # pointer to output tensor (row-normalized)
+        row_scale_ptr,  # pointer to store row-wise maximum absolute values
```
vkuzo, line 1049: should the comment say scale or max_abs? also row-wise should really say "across dim0"?

torchao/prototype/mx_formats/kernels.py
```diff
@@ -1038,6 +1040,175 @@ def to_mxfp8_dim1_kernel(
...
+        jump_vals_per_row = (n_cols - COL_TILE_SIZE) // INNER_BLOCK_SIZE
+
+        # example transformation (specifics depend on tile sizes):
+        # [0, 1, 2, 3, 4, 5, 6, 7] -> [0, 1, 4, 5, 8, 9, 12, 13]
+        row_scale_indices = row_scale_indices + (
+            (row_scale_indices // BLOCKS_PER_COL_TILE) * jump_vals_per_row
+        )
+
+        # Store the scales
+        tl.store(row_scale_start_ptr + row_scale_indices, row_scale_e8m0)
```
vkuzo: should this either have a mask or a TODO to add masking?

vkuzo (approved):

torchao/prototype/mx_formats/kernels.py
danielvegamyhre: I added masking in https://github.com/pytorch/ao/pull/3131 where I refactored/simplified the kernel, but I'll go ahead and move it here

torchao/prototype/mx_formats/kernels.py
danielvegamyhre: the dim1 kernel said "column-wise" here, so i just changed it to say "row-wise" here for consistency. dim0/dim1 is also fine. The comment shouldn't "maximum absolute values" for sure though, I will remove that from both kernels.

torchao/prototype/mx_formats/kernels.py
```diff
@@ -829,6 +829,8 @@ def _(uint8_data):
     import triton.language as tl
     from torch.library import triton_op, wrap_triton
 
+    print("importing triton ops")
```
drisspg: remove please

torchao/prototype/mx_formats/kernels.py
```diff
@@ -1038,6 +1040,175 @@ def to_mxfp8_dim1_kernel(
...
+        jump_vals_per_row = (n_cols - COL_TILE_SIZE) // INNER_BLOCK_SIZE
+
+        # example transformation (specifics depend on tile sizes):
+        # [0, 1, 2, 3, 4, 5, 6, 7] -> [0, 1, 4, 5, 8, 9, 12, 13]
+        row_scale_indices = row_scale_indices + (
+            (row_scale_indices // BLOCKS_PER_COL_TILE) * jump_vals_per_row
+        )
+
+        # Store the scales
+        tl.store(row_scale_start_ptr + row_scale_indices, row_scale_e8m0)
```
drisspg: is this storing in swizzled layout?

danielvegamyhre: No, we have separate kernels for that. Low hanging fruit could be to write directly to blocked swizzled format here, like Cursor mentioned they're doing in this blog https://cursor.com/blog/kernels

torchao/prototype/mx_formats/kernels.py
danielvegamyhre: Woops, was debugging an import error, thanks

torchao/prototype/mx_formats/kernels.py
drisspg: Its a non zero perf win for activation casting in inference. We can leave as a follow up but should be pretty easy to just have aconstexpr path do this

PR #3130, opened 2025-10-07
Update version to 0.15.0

jainapurva:
Updated the version from `0.14.0` to `0.15.0`.

### DIFF
--- a/.github/workflows/regression_test.yml
+++ b/.github/workflows/regression_test.yml
@@ -53,7 +53,7 @@ jobs:
         pip install .
         export CONDA=$(dirname $(dirname $(which conda)))
         export LD_LIBRARY_PATH=$CONDA/lib/:$LD_LIBRARY_PATH
-        pytest test/quantization/pt2e/test_x86inductor_fusion.py --verbose -s
+        pytest test --verbose -s
   test:
     strategy:
       fail-fast: false


jerryzh168 (approved):

andrewor14:
@liangel-02 anything we need to bump for the version compatibility?

liangel-02:
> @liangel-02 anything we need to bump for the version compatibility?

1. In https://github.com/pytorch/ao/blob/main/torchao/__init__.py#L76, update the condition to `2.9.0dev` and `0.14.0` (instead of `0.14.0dev`). Then, add a new condition for `0.15.0dev` and corresponding torch version - we want to add another "load .so files" condition when the user builds with the compatible version of torch for the updated torchao version.

2.  Add a new test for the new version in https://github.com/pytorch/ao/blob/main/scripts/test_torch_version_torchao_version_compatibility.sh . Manually run the bash script to verify these changes: `.scripts/test_torch_version_torchao_version_compatibility.sh`

3. We can also update the compatibility table in #2919 to reflect the latest updates 

@vkuzo please comment if there's anything missing

torchao/__init__.py
```diff
@@ -65,14 +65,19 @@ def _parse_version(version_string):
     v2_8_0 = _parse_version("2.8.0")
     v0_13_0 = _parse_version("0.13.0")
     v2_9_0_dev = _parse_version("2.9.0.dev")
```
jainapurva, line 67: As 2.9 has not been released yet, we might still need to call it dev, but the branch cut has been done in core, and the dev version has been updated to 2.10.0.dev. @liangel-02 How do we handle this?

PR #3135, opened 2025-10-09
add learnable_fake_quantize in pt2e

navsud:
Summary: Learnable Fake Quantize is a popular technique used especially for low-precision QAT. This was available in fx quantization at https://github.com/pytorch/pytorch/blob/main/torch/ao/quantization/_learnable_fake_quantize.py#L10?, but not in pt2e quantization flow. This change adds learnable fake quantize in pt2e, with some changes, especially related to how channel wise quantization works.

Differential Revision: D83542550

### DIFF
--- a/torchao/quantization/pt2e/learnable_fake_quantize.py
+++ b/torchao/quantization/pt2e/learnable_fake_quantize.py
@@ -1,0 +1,233 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+
+from typing import Optional
+
+import torch
+from torch.nn.parameter import Parameter
+
+from torchao.quantization.pt2e.fake_quantize import FakeQuantizeBase
+
+__all__ = ["LearnableFakeQuantize"]
+
+
+class LearnableFakeQuantize(FakeQuantizeBase):
+    r"""Generalized extension of the FakeQuantize module.
+
+    This is an extension of the FakeQuantize module, which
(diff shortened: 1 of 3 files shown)


torchao/quantization/pt2e/learnable_fake_quantize.py
jerryzh168, line 1: can you correct the name to `fake_quantizer`?

navsud: @jerryzh168 
I'm trying to be consistent with `FakeQuantize()` class at: https://github.com/pytorch/ao/blob/main/torchao/quantization/pt2e/fake_quantize.py. 
I'm ok to change the file name: `learnable_fake_quantize.py` to `learnable_fake_quantizer.py` and the class name from `LearnableFakeQuantize()` to `LearnableFakeQuantizer()` if that is what you meant.

andrewor14: Yeah this is for pt2e flow, so I feel the name `FakeQuantize` is more consistent with the other classes?

jerryzh168: `FakeQuantize` is the wrong name I think since it's a verb not noun, it's not consistent with `Observer` which is a noun. We can correct, and deprecate `FakeQuantize` itself as we add new things.

But it's OK to do this separately as well

PR #3138, opened 2025-10-09
[CI] Use single-GPU runners for ROCm MI3xx CI

jithunnair-amd:
If the job doesn't **need** >1 GPUs, we can reduce the runner GPU count to 1

cc @jagadish-amd


jithunnair-amd:
Merging this PR since it [successfully scheduled](https://github.com/pytorch/ao/actions/runs/18572814398/job/52950687299) on the 1-GPU runners; the test failures are unrelated to this PR.

PR #3151, opened 2025-10-10
Fix broken test_affine_quantized_tensor_parallel test after DeviceMesh

danielvegamyhre:
This stack made DeviceMesh public attributes read only as they normally shouldn't be modified after init: https://github.com/pytorch/pytorch/pull/164954

So we need to set the backing private attribute `mesh._device` instead of `mesh.device` in this test now to fix it.


andrewor14:
Just gonna merge this now to unbreak CI, thanks Daniel!

PR #3153, opened 2025-10-10
Testing subclassing

Krishn1412:
Added two tests to check if the subclassing logic defined in TorchAOBaseTensor __init_subclass_ _ works correctly. Refer the [comment](https://github.com/pytorch/ao/pull/2866#discussion_r2411724582).

The first test fails for:

```
self.assertIsNot(Parent._ATEN_OP_TABLE, Child._ATEN_OP_TABLE)
self.assertIsNot(Parent._TORCH_FN_TABLE, Child._TORCH_FN_TABLE)
```

telling us that Parent._ATEN_OP_TABLE and Child._ATEN_OP_TABLE are the same object, meaning the class attribute dictionary was shared, not re-initialized per subclass.

### DIFF
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -416,6 +416,49 @@ class TestTorchAOBaseTensor(unittest.TestCase):
         Parent._ATEN_OP_TABLE[Parent]["new_op"] = "added_later"
         self.assertNotIn("new_op", Child._ATEN_OP_TABLE[Child])
 
+    def test_subclassing_with_real_op(self):
+        counter = {"calls": 0}
+
+        class Parent(TorchAOBaseTensor):
+            tensor_data_names = ["qdata"]
+            tensor_attribute_names = ["attr"]
+
+            def __new__(cls, qdata, attr):
+                r = torch.Tensor._make_wrapper_subclass(cls, qdata.shape)
+                r.qdata = qdata
+                r.attr = attr
+                return r
+
+            def __init__(self, qdata, attr):
+                pass
+
+        # Real op implementation
+        @Parent.implements([torch.ops.aten.cat.default])
+        def _cat_op(func, types, args, kwargs):
+            counter["calls"] += 1
+            return func(*args, **kwargs)
+
+        class Child(Parent):
+            tensor_data_names = ["qdata"]
+            tensor_attribute_names = ["attr"]
+
+        # Table checks
+        self.assertIn(torch.ops.aten.cat.default, Parent._ATEN_OP_TABLE[Parent])
+        self.assertIn(torch.ops.aten.cat.default, Child._ATEN_OP_TABLE[Child])
+
+        # Ensure child table is distinct
+        self.assertIsNot(Parent._ATEN_OP_TABLE, Child._ATEN_OP_TABLE)
+
+        # calling the op through the child tensor
+        t1 = torch.randn(2, 3)
+        t2 = torch.randn(2, 3)
+        child_tensor1 = Child(t1, "a")
+        child_tensor2 = Child(t2, "b")
+
+        torch.ops.aten.cat.default([child_tensor1, child_tensor2], 0)
+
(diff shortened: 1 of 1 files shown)


Krishn1412:
Is that the expected behavior @jerryzh168 ? If it is, then the [if statement](https://github.com/pytorch/ao/pull/2866#discussion_r2383567801) will work. But it does seem problematic. Shouldn't each subclass rebuilt its own copy of the dict?

test/test_utils.py
```diff
@@ -392,6 +392,49 @@ def fake_linear(func, types, args, kwargs):
             counter["calls"], 2, "Expected fake_linear to be called via aten.t.default"
         )
 
+    def test_subclassing(self):
+        class Parent(TorchAOBaseTensor):
```
jerryzh168, line 396: the test should be doing this:
```
1. define Parent class
2. Parent class implements some method

# can do a real implementation or just some dummy one
@Parent.implements(aten.cat)
def _(..):
   ...
   
   
3. some child class inherits from parent: Child(Parent)

4. make sure the same op still works
by calling the op (e.g. aten.cat) and make sure it works (make sure it's called)

can do some additional table checks
```

test/test_utils.py
```diff
@@ -392,6 +392,49 @@ def fake_linear(func, types, args, kwargs):
...
+            tensor_attribute_names = ["attr"]
+
+        Parent._ATEN_OP_TABLE[Parent]["op_parent"] = "parent_impl"
+        Parent._TORCH_FN_TABLE[Parent]["fn_parent"] = "parent_fn_impl"
+
+        class Child(Parent):
+            tensor_data_names = ["qdata"]
+            tensor_attribute_names = ["attr"]
+
+        # ensure child has copied parent ops
```
jerryzh168, line 407: yeah I think idea is the same, but just changing it to real op implementation with `implements` and `implements_torch_function` would be better

test/test_utils.py
```diff
@@ -392,6 +392,49 @@ def fake_linear(func, types, args, kwargs):
...
+
+        # ensure the top-level dicts are distinct (not inherited)
+        self.assertIsNot(Parent._ATEN_OP_TABLE, Child._ATEN_OP_TABLE)
+        self.assertIsNot(Parent._TORCH_FN_TABLE, Child._TORCH_FN_TABLE)
+
+        # change the parent's op after subclass creation — should not leak
+        Parent._ATEN_OP_TABLE[Parent]["new_op"] = "added_later"
+        self.assertNotIn("new_op", Child._ATEN_OP_TABLE[Child])
+
+    def test_multiple_inheritance(self):
```
jerryzh168, line 462: yeah testing multiple inheritance is helpful as well, thanks

jerryzh168:
> Is that the expected behavior @jerryzh168 ? If it is, then the [if statement](https://github.com/pytorch/ao/pull/2866#discussion_r2383567801) will work. But it does seem problematic. Shouldn't each subclass rebuilt its own copy of the dict?

yeah, each sublcass should have their own dict, I think the current impl is problematic

Krishn1412:
@jerryzh168 ,
the new test with real op fails for the same line
`self.assertIsNot(Parent._ATEN_OP_TABLE, Child._ATEN_OP_TABLE).`

Also when it is called using the child tensor, it works.

For the inheritance test case,
it fails at
`self.assertEqual(C._ATEN_OP_TABLE[C]["shared"], "from_b")`
AssertionError: 'from_a' != 'from_b'.

Should we start working on changing the implementation?

jerryzh168:
sorry I missed the comment

> the new test with real op fails for the same line
> self.assertIsNot(Parent._ATEN_OP_TABLE, Child._ATEN_OP_TABLE).

yeah we need to make sure parent and child classes are not sharing the table, we should fix this

> For the inheritance test case,

yeah seems reasonable to me as well, please feel free to start changing this

PR #3158, opened 2025-10-12
Update Float8Tensor for GRPO training in unsloth

andrewor14:
**Summary:** Support a few extra ops called during GRPO loop in unsloth/vllm for Float8Tensor.

**Test Plan:**
```
python test/quantization/quantize_/workflows/float8/test_float8_tensor.py -k test_fp8_matmul_lora_variants
python test/quantization/quantize_/workflows/float8/test_float8_tensor.py -k test_to_dtype_layout
python test/quantization/quantize_/workflows/float8/test_float8_tensor.py -k test_has_compatible_shallow_copy_type
python test/quantization/quantize_/workflows/float8/test_float8_tensor.py -k test_transpose
```

### DIFF
--- a/test/quantization/quantize_/workflows/float8/test_float8_tensor.py
+++ b/test/quantization/quantize_/workflows/float8/test_float8_tensor.py
@@ -18,6 +18,7 @@ from torch.testing._internal.common_utils import run_tests
 from torchao.quantization import (
     Float8DynamicActivationFloat8WeightConfig,
     Float8WeightOnlyConfig,
+    Granularity,
     PerBlock,
     PerRow,
     PerTensor,
@@ -42,6 +43,8 @@ torch._dynamo.config.cache_size_limit = 128
 class ToyLinearModel(torch.nn.Module):
     def __init__(self, in_features, out_features, bias):
         super().__init__()
+        self.in_features = in_features
+        self.out_features = out_features
         self.linear1 = torch.nn.Linear(in_features, out_features, bias=bias)
         self.linear2 = torch.nn.Linear(out_features, in_features, bias=bias)
 
@@ -49,6 +52,21 @@ class ToyLinearModel(torch.nn.Module):
         x = self.linear1(x)
         x = self.linear2(x)
         return x
+
+    def check_weight_scaling(self, granularity: Granularity):
+        qs1 = self.linear1.weight.scale
+        qs2 = self.linear2.weight.scale
+        N, K = (self.out_features, self.in_features)
+        if granularity == PerTensor():
+            assert qs1.shape == (1, 1)
+            assert qs2.shape == (1, 1)
+        elif granularity == PerRow():
+            assert qs1.shape == (N, 1)
+            assert qs2.shape == (K, 1)
+        else:
+            assert granularity == (PerBlock([1, 128]), PerBlock([128, 128]))
+            assert qs1.shape == (N // 128, K // 128)
+            assert qs2.shape == (K // 128, N // 128)
 
 
 class ToyConvModel(torch.nn.Module):
@@ -71,6 +89,47 @@ class ToyConvModel(torch.nn.Module):
 
     def forward(self, x):
         return self.conv(x)
+
+
+class ToyLoRAModel(torch.nn.Module):
+    def __init__(
+        self,
+        in_features: int,
+        out_features: int,
+        lora_rank: int,
+        device: torch.device,
+    ):
+        super().__init__()
+        self.in_features = in_features
+        self.out_features = out_features
+        self.linear = torch.nn.Linear(
+            in_features,
+            out_features,
+            bias=False,
+            device=device,
+        )
+        self.lora_A = torch.nn.Parameter(
+            torch.randn(in_features, lora_rank, device=device),
+        )
+        self.lora_B = torch.nn.Parameter(
+            torch.randn(lora_rank, out_features, device=device),
+        )
+
+    def forward(self, x):
+        matmul_out = torch.matmul(x, self.linear.weight.t())
+        lora_out = x @ self.lora_A @ self.lora_B
+        return matmul_out + lora_out
+
+    def check_weight_scaling(self, granularity: Granularity):
+        qs = self.linear.weight.scale
+        N, K = (self.out_features, self.in_features)
+        if granularity == PerTensor():
+            assert qs.shape == (1, 1)
+        elif granularity == PerRow():
+            assert qs.shape == (N, 1)
+        else:
+            assert granularity == (PerBlock((1, 128)), PerBlock((128, 128)))
+            assert qs.shape == (N // 128, K // 128)
 
 
 # TODO: move tests in test_affine_quantized_float.py here after we migrated all implementations
@@ -112,10 +171,75 @@ class TestFloat8Tensor(TorchAOIntegrationTestCase):
         dtype: torch.dtype,
         mode: str,
         compile: bool,
-        granularity,
+        granularity: Granularity,
         kernel_preference: KernelPreference,
         sizes: Tuple,
         bias: bool,
+    ):
+        _, N, K = sizes
+        self._test_fp8_matmul_model(
+            dtype,
+            mode,
+            compile,
+            granularity,
+            kernel_preference,
+            sizes,
+            bias,
+            ToyLinearModel(K, N, bias),
+        )
+
+    @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
+    @unittest.skipIf(
+        not is_sm_at_least_89(), "Requires GPU with compute capability >= 8.9"
+    )
+    @common_utils.parametrize("dtype", [torch.bfloat16, torch.float32])
+    @common_utils.parametrize("mode", ["dynamic", "weight-only"])
+    @common_utils.parametrize("compile", [True, False])
+    @common_utils.parametrize("granularity", [PerTensor(), PerRow()])
+    @common_utils.parametrize(
+        "kernel_preference",
+        [KernelPreference.AUTO, KernelPreference.TORCH, KernelPreference.FBGEMM],
+    )
+    # Inputs are (M,..), K, N
+    @common_utils.parametrize(
+        "sizes",
+        [
+            ((128,), 256, 128),
+            ((32, 128), 64, 256),
+        ],
+    )
+    def test_fp8_matmul_lora_variants(
+        self,
+        dtype: torch.dtype,
+        mode: str,
+        compile: bool,
+        granularity: Granularity,
+        kernel_preference: KernelPreference,
+        sizes: Tuple,
+    ):
+        _, N, K = sizes
+        model = ToyLoRAModel(K, N, lora_rank=8, device=torch.device("cpu"))
+        self._test_fp8_matmul_model(
+            dtype,
+            mode,
+            compile,
+            granularity,
+            kernel_preference,
+            sizes,
+            bias=False,
+            model=model.to("cuda"),
+        )
+
+    def _test_fp8_matmul_model(
+        self,
+        dtype: torch.dtype,
+        mode: str,
+        compile: bool,
+        granularity: Granularity,
+        kernel_preference: KernelPreference,
+        sizes: Tuple,
+        bias: bool,
+        model: torch.nn.Module,
     ):
         if isinstance(granularity, PerTensor):
             if kernel_preference is KernelPreference.FBGEMM:
@@ -172,9 +296,7 @@ class TestFloat8Tensor(TorchAOIntegrationTestCase):
         with error_context:
             M, N, K = sizes
             input_tensor = torch.randn(*M, K, dtype=dtype, device="cuda")
-
-            # Create a linear layer with bfloat16 dtype
-            model = ToyLinearModel(K, N, bias).eval().to(dtype).to("cuda")
+            model = model.eval().to(dtype).to("cuda")
 
             quantized_model = copy.deepcopy(model)
 
@@ -190,18 +312,7 @@ class TestFloat8Tensor(TorchAOIntegrationTestCase):
             quantize_(quantized_model, config)
 
             # ensure weight scaling is what we expect
-            qs1 = quantized_model.linear1.weight.scale
-            qs2 = quantized_model.linear2.weight.scale
-            if granularity == PerTensor():
-                assert qs1.shape == (1, 1)
-                assert qs2.shape == (1, 1)
-            elif granularity == PerRow():
-                assert qs1.shape == (N, 1)
-                assert qs2.shape == (K, 1)
-            else:
-                assert granularity == (PerBlock([1, 128]), PerBlock([128, 128]))
-                assert qs1.shape == (N // 128, K // 128)
-                assert qs2.shape == (K // 128, N // 128)
+            quantized_model.check_weight_scaling(granularity)
 
             if compile:
                 quantized_model = torch.compile(quantized_model, fullgraph=True)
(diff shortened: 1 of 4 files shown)


test/quantization/quantize_/workflows/float8/test_float8_tensor.py
```diff
@@ -148,6 +150,64 @@ def test_fp8_linear_variants(
...
+            kernel_preference=kernel_preference,
+        )
+        weight_tensor_fp8 = Float8Tensor.from_hp(
+            weight_tensor,
+            granularity=granularity,
+            mm_config=mm_config,
+            kernel_preference=kernel_preference,
+        )
+        output_tensor = torch.matmul(input_tensor, weight_tensor.t())
+        output_tensor_fp8 = torch.matmul(input_tensor_fp8, weight_tensor_fp8.t())
```
jerryzh168: this is not used through the quantize_ API?

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
```diff
@@ -248,14 +248,56 @@ def from_hp(
...
+    input_tensor, weight_tensor = args[0], args[1]
+    return _float8_mm_impl(input_tensor, weight_tensor)
+
+
+@implements(aten.addmm_.default)
+def _(func, types, args, kwargs):
+    output_tensor, input_tensor, weight_tensor = (
+        args[0],
+        args[1],
+        args[2] if len(args) > 2 else None,
```
jerryzh168: why is weight tensor optional?

jerryzh168: should we add assert for the last 2 kwargs as well https://github.com/pytorch/pytorch/blob/82ff07c7884d478ddd5d638bebbb938e55c9bebf/aten/src/ATen/native/native_functions.yaml#L7214

jerryzh168: also I thought one of mat1 and mat2 should be bias_tensor?

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
```diff
@@ -665,6 +708,68 @@ def _(func, types, args, kwargs):
...
+    if "dtype" in kwargs:
+        assert kwargs["dtype"] == self.dtype
+    if "layout" in kwargs:
+        assert kwargs["layout"] == self.layout
+    # if device is the same, treat this like a no-op
+    device = kwargs.get("device")
+    if device == self.device:
+        return self
+    # otherwise, move all inner tensors to the new device
+    new_tensor = self.__class__(
```
jerryzh168: nit: we can use `Float8Tensor` instead of `self.__class__` to be consistent

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
```diff
@@ -665,6 +708,68 @@ def _(func, types, args, kwargs):
     return return_and_correct_aliasing(func, args, kwargs, new)
 
 
+@implements(torch.ops.aten.to.dtype_layout)
```
jerryzh168: can this be implemented in TorchAOBaseTensor? seems nothing special: https://github.com/pytorch/ao/blob/3577306c8b32517afe8eb6eb7e84335601180598/torchao/utils.py#L583

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
```diff
@@ -665,6 +708,68 @@ def _(func, types, args, kwargs):
...
+        self.dtype,
+    )
+    return return_and_correct_aliasing(func, args, kwargs, new_tensor)
+
+
+# This is called during _apply() to see if we can shallow
+# copy the content of one tensor into another. For now,
+# we only allow shallow copy if both tensors are `Float8Tensor`
+# and have the same shape.
+@implements_torch_function(torch._has_compatible_shallow_copy_type)
```
jerryzh168: if nohing special, we could also implement this in TorchAOBaseTensor I think

test/quantization/quantize_/workflows/float8/test_float8_tensor.py
jerryzh168: if this can be accessed through quantize_ API then we can merge the test with test_linear_variants I think

andrewor14: I don't think this can be accessed through the quantize_ API unfortunately, nn.Linear will dispatch to F.linear first

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
```diff
@@ -248,14 +248,56 @@ def from_hp(
...
+    weight_tensor: torch.Tensor,
+) -> torch.Tensor:
+    assert isinstance(weight_tensor, Float8Tensor), (
+        f"Don't expect to reach here with an override other than weight currently, {type(input_tensor)} {type(weight_tensor)}"
+    )
+    is_transposed = weight_tensor.qdata.stride(-2) < weight_tensor.qdata.stride(-1)
+    # For matmul(x, w.t()), just call the linear implementation
+    # For matmul(x, w), just dequantize for now, we can optimize later
+    if is_transposed:
+        return _float8_linear_impl(input_tensor, weight_tensor.t())
```
vkuzo: going from `mm` to `linear` is going backwards in terms of levels of abstraction, IMO this should be changed to not call `linear`-named utils.

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
```diff
@@ -248,14 +248,56 @@ def from_hp(
...
+    else:
+        return torch.matmul(input_tensor, weight_tensor.dequantize())
```
vkuzo: mm to matmul is also going to a higher level thing, better to call `torch.mm` here

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
```diff
@@ -248,14 +248,56 @@ def from_hp(
...
     input_tensor, weight_tensor, bias = (
         args[0],
         args[1],
         args[2] if len(args) > 2 else None,
     )
+    return _float8_linear_impl(input_tensor, weight_tensor, bias)
+
+
+@implements(aten.mm.default)
+@implements_torch_function(torch.matmul)
```
vkuzo, line 268: `mm` and `matmul` are not the same op, `matmul` supports broadcasting, IMO I'd not mix them together

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
andrewor14: yeah the first tensor is the bias, also added the asserts

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
vkuzo: `_float8_mm_impl` seems confusing, IMO this should be refactor to cleanly override individual torch or aten ops and ensure that the logic of when to do weight-only vs dynamic quant is consistent everywhere

vkuzo (changes requested):
plz clean up `_float8_mm_impl`

test/quantization/quantize_/workflows/float8/test_float8_tensor.py
vkuzo: could you share a code snippet of how the callsite is creating these Float8Tensor objects

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
```diff
@@ -248,14 +248,151 @@ def from_hp(
...
+        )
+        kernel_choice = "torch"
+    return kernel_choice
+
+
+def _call_fbgemm_f8f8bf16_matmul(
+    input_tensor: Float8Tensor,
+    weight_tensor: Float8Tensor,
+    bias: Optional[torch.Tensor] = None,
+    weight_is_already_transposed: bool = False,
```
vkuzo: instead of this flag, just transpose at the callsite to match the meaning of `matmul`

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
vkuzo, line 305: the current assumption here is that `input_tensor` is in high precision unless quantized by this logic.  If you are introducing a path where this isn't true, let's change this to

```
if act_quant_kwargs is not None:
    assert not isinstance(input_tensor, AOBaseTensor), "unsupported"
    ...
```

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
```diff
@@ -361,6 +427,37 @@ def _(func, types, args, kwargs):
         )
 
 
+def _float8_matmul_impl(
```
vkuzo: how about

1. define _float8_mm_impl as the float8 version of `torch.mm`, as the lowest level shared code of maybe quantizing the input and then choosing a gemm
2. all other functions (matmul, linear, etc) call _float8_mm_impl

it's a bit confusing to have two different paths for linear and matmul

test/quantization/quantize_/workflows/float8/test_float8_tensor.py
vkuzo: I think a test for actually doing whatever the real callsite is doing would be good here, i.e. define a toy model, then define how the float8 matmul is injected, then run that.

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
andrewor14: The reason behind this flag is to prevent unnecessary double transpose when we call linear, since the fbgemm op expects the weight to be in the linear format (already transposed). So if we don't have this flag:
```
1. linear calls _float8_mm_impl(input, weight.t())
2. _float8_mm_impl calls weight.t() before calling torch.ops.fbgemm.f8f8bf16
```
If we just transpose the weight for linear, we may end up slowing linear down, is that OK?

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
andrewor14: This was done to avoid double transpose in the linear path (which doesn't happen today, see [this comment](https://github.com/pytorch/ao/pull/3158#discussion_r2481727760)). I agree that ideally everything should go through `_float8_mm_impl`, but doing so may add overhead for the linear path, should I go ahead and merge the implementations anyway?

andrewor14: Ok, just did some benchmark on the double transpose thing, seems like it didn't introduce much overhead. I refactored the code to the way you suggested, please have another look

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
andrewor14: added the assert

test/quantization/quantize_/workflows/float8/test_float8_tensor.py
andrewor14: > could you share a code snippet of how the callsite is creating these Float8Tensor objects

Sure, here's the callsite for matmuls on the Float8Tensor: https://github.com/unslothai/unsloth/blob/2267b5c5532957141a33bfa5bb9f0b220a4b3efe/unsloth/kernels/fast_lora.py#L68. Then they do some patching to replace the mlp module to call this autograd function instead of the original linear layers: https://github.com/unslothai/unsloth/blob/2267b5c5532957141a33bfa5bb9f0b220a4b3efe/unsloth/models/llama.py#L2860. The `Float8Tensors` were previously already created and saved in the checkpoint.

> I think a test for actually doing whatever the real callsite is doing would be good here, i.e. define a toy model, then define how the float8 matmul is injected, then run that.

Sounds good, I can add that

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
```diff
@@ -254,21 +254,59 @@ def from_hp(
...
+        args[1],
+        args[2],
+    )
+    assert kwargs.get("alpha", 1) == 1, "only alpha=1 is supported"
+    assert kwargs.get("beta", 1) == 1, "only beta=1 is supported"
+    out = _float8_mm_impl(input_tensor, weight_tensor)
+    return bias_tensor.add_(out)
+
+
+def _float8_mm_impl(
```
vkuzo: nit: can we call it `addmm` since bias is added inside?

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
```diff
@@ -300,6 +338,7 @@ def _(func, types, args, kwargs):
             mm_config = weight_tensor.mm_config
             assert mm_config is not None
             assert not _is_128_128_scaled(weight_tensor), "unimplemented"
+            weight_tensor = weight_tensor.t()
```
vkuzo: nit: just inline this?

```
res = torch.ops.fbgemm.f8f8bf16_rowwise(
  ...
  wq.t(),
  ...,
  w_scale.t(),
  ...
)
```

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
```diff
@@ -334,28 +373,25 @@ def _(func, types, args, kwargs):
             assert kernel_choice == "torch"
             scaled_mm_config = weight_tensor.mm_config
             assert scaled_mm_config is not None
-            out_shape = get_out_shape(input_tensor.shape, weight_tensor.shape)
+            out_shape = (*input_tensor.shape[:-1], weight_tensor.shape[1])
```
vkuzo: technically `addmm` and `mm` don't support broadcasting of the gemm operands, would be more correct to enforce `x` and `w` to be 2d here and move the 3d shape support to `matmul` and `linear`.  If you want to do a TODO to punt, sounds ok to me.

test/quantization/quantize_/workflows/float8/test_float8_tensor.py
```diff
@@ -801,6 +901,38 @@ def test_slice_3d_operation(self, granularity, slice_dim, tensor_shape):
 
         self.assertEqual(sliced_dequantized, sliced_original)
 
+    def test_to_dtype_layout(self):
```
vkuzo, line 921: why do we need this?

test/quantization/quantize_/workflows/float8/test_float8_tensor.py
```diff
@@ -801,6 +901,38 @@ def test_slice_3d_operation(self, granularity, slice_dim, tensor_shape):
...
+        x = torch.randn(128, 512, device="cuda", dtype=torch.bfloat16)
+        x_fp8 = Float8Tensor.from_hp(x)
+        y_fp8 = torch.ops.aten.to.dtype_layout(
+            x_fp8, dtype=x_fp8.dtype, layout=x_fp8.layout, device="cpu"
+        )
+        self.assertEqual(y_fp8.dtype, x_fp8.dtype)
+        self.assertEqual(y_fp8.layout, x_fp8.layout)
+        self.assertEqual(y_fp8.device, torch.device("cpu"))
+
+    def test_has_compatible_shallow_copy_type(self):
```
vkuzo, line 931: why do we need this?

test/quantization/quantize_/workflows/float8/test_float8_tensor.py
```diff
@@ -112,10 +161,74 @@ def test_fp8_linear_variants(
...
+        self,
+        dtype: torch.dtype,
+        mode: str,
+        compile: bool,
+        granularity: Granularity,
+        kernel_preference: KernelPreference,
+        sizes: Tuple,
+    ):
+        _, N, K = sizes
+        self._test_fp8_matmul_model(
```
vkuzo, line 222: is there any need to test the backward?

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
andrewor14: Hmm I tried doing this but looks like `_is_rowwise_scaled` expects a rowwise `Float8Tensor`, and `weight_tensor` is colwise scaled at this point (it was previously transposed in linear). Also `wq` and `w_scale` are used in the tensorwise case too

test/quantization/quantize_/workflows/float8/test_float8_tensor.py
andrewor14: These two are called when the transformer trainer moves the model to a different device before training, e.g. https://github.com/pytorch/pytorch/blob/main/torch/nn/modules/module.py#L938

test/quantization/quantize_/workflows/float8/test_float8_tensor.py
andrewor14: I think it's OK since the backward just dequantizes and does a basic bf16 matmul? I can still add it if you'd like

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
vkuzo: how about

```
if _is_rowwise_scaled(weight_tensor.t()): ...
```

IMO worth changing since we are inside `addmm`, and `addmm` expects the a to be (M, K) and b to be (K, N), changing that is a bit confusing

test/quantization/quantize_/workflows/float8/test_float8_tensor.py
vkuzo: makes sense, add to `test_fp8_matmul_lora_variants`?  IMO the goal of that test is to test the actual use case e2e, so if the actual use case includes the dtype movement then the test should too.

PR #3160, opened 2025-10-13
Small updates to main torchao README.md

vkuzo:
Minor simplifications of main README.md:
1. move 2025H1 news to the section where user has to expand to see
2. simplify key value prop sentence to "TorchAO is an easy to use quantization library for native PyTorch"
3. add some missing `bash` and `python` modifiers to code blocks
4. rename "float8" to "quantized training" to match the rest of the document
5. move "sparse training" and "memory efficient optimizers" to the section where user has to expand to see


README.md
```diff
@@ -47,8 +47,7 @@
 
 ## 🌅 Overview
 
-TorchAO is a PyTorch-native model optimization framework leveraging quantization and sparsity to provide an end-to-end, training-to-serving workflow
-for AI models. TorchAO works out-of-the-box with `torch.compile()` and `FSDP2` across most HuggingFace PyTorch models. Key features include:
+TorchAO is an easy to use quantization library for native PyTorch. TorchAO works out-of-the-box with `torch.compile()` and `FSDP2` across most HuggingFace PyTorch models. Key features include:
```
andrewor14, line 50: This looks fine to me, wondering if @supriyar have any thoughts on the wording since it's different from the paper now

supriyar: it's fine, since the new statement does not mean other one isn't true anymore. I think @vkuzo is trying to simplify the message users see in first glance of the repo.

PR #3162, opened 2025-10-13
Simplify quantization README.md

vkuzo:
Move features which are not top of mind in 2025 to a section near the bottom where user can click to expand.  For now, keep all the docs as they are, we can make further changes in future PRs.

Specifically, we move the following sections to the bottom:
1. autoquant
2. AQT in-depth deep dive
3. k-v cache quantization

Note that the PR looks long, but the other sections that appear in the PR are actually no-change, this is just git choosing to display those lines as "added" even though they were not touched because of the way 1-3 above are moved to the bottom.


torchao/quantization/README.md
supriyar, line 1: can we move H100 results to the top in benchmarks and maybe remove llama2 numbers for A100 since we have llama3?

vkuzo: there is a different PR to remove the LLaMa 2 numbers.  Agree, we can bump H100 in a separate PR.

PR #3165, opened 2025-10-13
Add parq utility to create an optimizer

metascroy:
Adds utility to create an optimizer like this:

```
quant_configs_and_filter_fns = [
    (QuantConfig(bitwidth=2, group_size=128), use_2bit),
    (QuantConfig(bitwidth=3, group_size=256), use_3bit),
    (QuantConfig(bitwidth=4, group_size=256), use_4bit),
    (QuantConfig(bitwidth=8), use_8bit),
 ]

optimizer = create_optimizer(model, quant_configs_and_filter_fns, base_optimizer_cls, base_optimizer_kwargs)
```

The filter functions are (module, fqn) -> bool, e.g.,

```
def use_2bit(m, fqn):
        if not fqn.endswith(".weight"):
            return False

        # Layers 0-3
        for i in range(4):
            if fqn.startswith(f"model.layers.{i}."):
                for key in ["v_proj", "gate_proj", "up_proj"]:
                    if key in fqn:
                        return True

        # Layers 4-31
        for i in range(4, 32):
            if fqn.startswith(f"model.layers.{i}."):
                for key in ["q_proj", "v_proj", "gate_proj", "up_proj", "down_proj"]:
                    if key in fqn:
                        return True

        return False
```

### DIFF
--- a/torchao/prototype/parq/api.py
+++ b/torchao/prototype/parq/api.py
@@ -1,5 +1,7 @@
 from dataclasses import dataclass
-from typing import Optional
+from typing import Any, Callable, Dict, List, Optional, Tuple, Type
+
+import torch
 
 from torchao.prototype.parq.optim import QuantOptimizer
 from torchao.prototype.parq.quant import (
@@ -29,7 +31,12 @@ class QuantConfig:
             object.__setattr__(self, "quantizer", q)
 
 
-def create_param_groups_and_group_quantizer_map(model, quant_configs_and_filter_fns):
+def create_param_groups_and_group_quantizer_map(
+    model: torch.nn.Module,
+    quant_configs_and_filter_fns: List[
+        Tuple[QuantConfig, Callable[[torch.nn.Module, str], bool]]
+    ],
+):
     param_groups = []
     group_quantizer_map = {}
     for idx, (config, _) in enumerate(quant_configs_and_filter_fns):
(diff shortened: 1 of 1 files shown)


torchao/prototype/parq/api.py
```diff
@@ -0,0 +1,136 @@
...
+            "quant_bits": config.bitwidth,
+        }
+        if config.group_size is not None:
+            param_group["quant_block_size"] = config.group_size
+        param_group["_quantizer"] = config.quantizer
+        param_groups.append(param_group)
+
+    # Non-quantized group at end so that index in param_groups
+    # is the index in the subset of quantized param groups, which is
+    # used in defining group_quantizer_map
```
lisjin, line 55: Standardizing the indices this way is a great idea

torchao/prototype/parq/api.py
```diff
@@ -0,0 +1,136 @@
...
+
+    return param_groups, group_quantizer_map
+
+
+from torchao.prototype.parq import ProxHardQuant
+
+
+def create_optimizer(
+    model,
+    quant_configs_and_filter_fns,
```
lisjin: Should we add a type hint here?

torchao/prototype/parq/api.py
```diff
@@ -0,0 +1,136 @@
...
+                    matching_config = config
+                    print(f"{config.bitwidth},{config.group_size}")
+                else:
+                    raise ValueError(
+                        f"Found multiple matching configs for {param_name}. Previous match={matching_config}, new match={config}."
+                    )
+
+        # If no match, add to no-quant group at last idx
+        if matching_config is None:
+            print("NONE")
```
lisjin, line 94: Is this for debugging?

metascroy: After each parameter, it prints out config (bitwidth,groupsize or NONE)

PR #3182, opened 2025-10-15
[WIP] Move float8 cutlass sparse layout to Float8SemiSparseTensor

bbeckca:
Moving float8 cutlass sparse layout into its own class:
https://github.com/pytorch/ao/blob/main/torchao/dtypes/floatx/cutlass_semi_sparse_layout.py

cc @jerryzh168 @jcaip @danielvegamyhre

### DIFF
--- a/test/quantization/quantize_/workflows/float8/test_float8_semi_sparse.py
+++ b/test/quantization/quantize_/workflows/float8/test_float8_semi_sparse.py
@@ -5,9 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 import unittest
-from torchao.quantization.quantize_.workflows.float8.float8_semi_sparse_tensor import Float8SemiSparseTensor
-from torchao.quantization.quantize_.workflows.float8.float8_tensor import Float8Tensor
-from torchao.float8.inference import Float8MMConfig
+
 import torch
 from torch.testing._internal.common_utils import (
     TestCase,
@@ -15,6 +13,12 @@ from torch.testing._internal.common_utils import (
     parametrize,
     run_tests,
 )
+
+from torchao.float8.inference import Float8MMConfig
+from torchao.quantization.quantize_.workflows.float8.float8_semi_sparse_tensor import (
+    Float8SemiSparseTensor,
+)
+from torchao.quantization.quantize_.workflows.float8.float8_tensor import Float8Tensor
 from torchao.sparsity.sparse_api import apply_fake_sparsity
 from torchao.testing.utils import skip_if_rocm
 from torchao.utils import is_sm_at_least_90
@@ -44,19 +48,24 @@ class TestFloat8SemiSparseTensor(TestCase):
         linear = torch.nn.Linear(K, N, dtype=dtype, device=device)
 
         apply_fake_sparsity(linear)
-        
+
         mm_config = Float8MMConfig(use_fast_accum=True)
-        input_fp8 = Float8Tensor.from_hp(input, float8_dtype=torch.float8_e4m3fn, mm_config=mm_config)
-        
-        weight_fp8 = Float8Tensor.from_hp(linear.weight.data, float8_dtype=torch.float8_e4m3fn, mm_config=mm_config)
-        dense_output = torch.nn.functional.linear(input_fp8, weight_fp8, linear.bias)
-        
-        weight_sparse_fp8 = Float8SemiSparseTensor.from_hp(linear.weight.data, [1, K])
-        sparse_output = torch.nn.functional.linear(input_fp8, weight_sparse_fp8, linear.bias)
-        
-        torch.testing.assert_close(
-            dense_output, sparse_output, atol=3e-1, rtol=3e-1
+        input_fp8 = Float8Tensor.from_hp(
+            input, float8_dtype=torch.float8_e4m3fn, mm_config=mm_config
         )
+
+        weight_fp8 = Float8Tensor.from_hp(
+            linear.weight.data, float8_dtype=torch.float8_e4m3fn, mm_config=mm_config
+        )
+        dense_output = torch.nn.functional.linear(input_fp8, weight_fp8, linear.bias)
+
+        weight_sparse_fp8 = Float8SemiSparseTensor.from_hp(linear.weight.data, [1, K])
+        sparse_output = torch.nn.functional.linear(
+            input_fp8, weight_sparse_fp8, linear.bias
+        )
+
+        torch.testing.assert_close(dense_output, sparse_output, atol=3e-1, rtol=3e-1)
+
 
 instantiate_parametrized_tests(TestFloat8SemiSparseTensor)
 
--- a/torchao/quantization/quantize_/workflows/float8/float8_semi_sparse_tensor.py
+++ b/torchao/quantization/quantize_/workflows/float8/float8_semi_sparse_tensor.py
@@ -86,7 +86,7 @@ implements_torch_function = Float8SemiSparseTensor.implements_torch_function
 @implements(aten.t.default)
 def _(func, types, args, kwargs):
     from torch.utils._python_dispatch import return_and_correct_aliasing
-    
+
     self = args[0]
     new = Float8SemiSparseTensor(
         sparse=self.sparse,
@@ -98,8 +98,10 @@ def _(func, types, args, kwargs):
 
 def _linear_fp8_semi_sparse(input_tensor, weight_tensor, bias):
     from torchao.ops import rowwise_scaled_linear_sparse_cutlass_f8f8
-    from torchao.quantization.quantize_.workflows.float8.float8_tensor import Float8Tensor
-    
+    from torchao.quantization.quantize_.workflows.float8.float8_tensor import (
+        Float8Tensor,
+    )
+
     if isinstance(input_tensor, Float8Tensor):
         input = input_tensor.qdata
         input_scale = input_tensor.scale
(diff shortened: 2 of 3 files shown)


torchao/quantization/quantize_/common/packing_format.py
```diff
@@ -32,3 +32,4 @@ class PackingFormat(str, Enum):
     needed for the rest of the system to understand the specific format that's adopted.
     """
     OPAQUE = "opaque"
+    # todo: add semi-sparse
```
bbeckca: @jerryzh168 It seems we may want to add a packing format for sparse. Wondering if there's a preference between adding it here or in a separate file ([similar to int4](https://github.com/pytorch/ao/blob/30082cb3d0e6f86b1bae5e75ef4284e3a4b5ece6/torchao/quantization/quantize_/workflows/int4/int4_packing_format.py#L4)) for float8?

jcaip: Do we need packing format if we have a separate config? It looks like packing format is mostly to support different Int4WeightOnlyConfig kernel options (tinygemm, sparse marlin, etc).

test/quantization/quantize_/workflows/float8/test_float8_semi_sparse.py
```diff
@@ -0,0 +1,108 @@
...
+from torchao.quantization import (
+    Float8WeightOnlyConfig,
+    quantize_,
+)
+from torchao.quantization.utils import compute_error
+from torchao.sparsity.sparse_api import apply_fake_sparsity
+from torchao.testing.utils import skip_if_rocm
+from torchao.utils import torch_version_at_least
+
+BF16_ACT_CONFIG = Float8WeightOnlyConfig(
```
jcaip: I don't think this config makes sense, it's not something we support. From what I understand this is a bf16 a + fp8 sparse weight? We only have kernel support for fp8xfp8 +2:4 sparse matmul, no support for mixed  input dtypes currently.

torchao/quantization/quantize_/workflows/float8/float8_semi_sparse_tensor.py
```diff
@@ -0,0 +1,114 @@
...
+            meta,
+            scale,
+        )
+
+
+implements = Float8SemiSparseTensor.implements
+implements_torch_function = Float8SemiSparseTensor.implements_torch_function
+
+
+@implements(aten.linear.default)
```
jcaip: We'll also need to make sure `mm` and `addmm` are supported ops as well. The arg order is different from linear but it should be the same logic.

bbeckca: Sounds good, I'm onboard with that. Mind if I add those ops in a follow-up diff after this lands?

torchao/quantization/quantize_/common/packing_format.py
bbeckca: Good point, I noticed that we seem to replace the dense weight with quantized semi-sparse [in the transform](https://github.com/pytorch/ao/blob/ed4cd34323ce58c4e82d1561a72abee3356eb1ad/torchao/quantization/quant_api.py#L1881) Would it make more sense to integrate Float8SemiSparseTensor here rather than gating with packing-format as I proposed previously? cc @jerryzh168

test/quantization/quantize_/workflows/float8/test_float8_semi_sparse.py
bbeckca: You're right, it seems I should be mirroring test_fp8_cutlass_sparse ([from test_sparse_api.py](https://github.com/pytorch/ao/blob/ed4cd34323ce58c4e82d1561a72abee3356eb1ad/test/sparsity/test_sparse_api.py#L132)) instead
with the difference being using the new flag/config which exposes the tensor subclass being added?

torchao/quantization/quantize_/workflows/float8/float8_semi_sparse_tensor.py
jcaip: Yup that's fine with me :)

test/quantization/quantize_/workflows/float8/test_float8_semi_sparse.py
jcaip: I think `Float8DynamicActivationFloat8SemiSparseWeightConfig` should eventually resolve to your subclass.
But I would like to hold off on that until 
1) We're out of the QRT period (cc @RandySheriff just FYI on refactor plans)
2) We have the same functionality (addmm/mm support)

torchao/quantization/quantize_/common/packing_format.py
jcaip: Yup I think that the transform should call your subclass.

test/quantization/quantize_/workflows/float8/test_float8_semi_sparse.py
bbeckca: Sounds good. In that case, should we sequence the changes as follows?
1. Land `Float8SemiSparseTensor` with linear support
2. Add mm / addmm ops for feature parity
3. Integrate into `Float8DynamicActivationFloat8SemiSparseWeightConfig` after QRT

bbeckca: @jcaip @jerryzh168 Mind confirming that you're onboard with this direction?

jcaip: Yeah, that sounds good to me.

test/quantization/quantize_/workflows/float8/test_float8_semi_sparse.py
```diff
@@ -0,0 +1,74 @@
...
+    @skip_if_rocm("ROCm enablement in progress")
+    @parametrize(
+        "sizes",
+        [
+            ((128,), 256, 128),
+            ((32, 128), 512, 128),
+            ((2, 32, 128), 256, 128),
+        ],
+    )
+    def test_sparse_vs_dense_fp8(self, sizes):
```
bbeckca, line 42: @jcaip Updated testing to follow style of [test_sparse_apu.py](https://fburl.com/18n157bf. For now I'm omitting any config related changes until QRT, however this diff does include all ops (linear, addmm, mm) so that integration can be done as follow up.

Could I get feedback on the construction of the implementations and test before adding similar for addmm, mm and adding polish?

test/quantization/quantize_/workflows/float8/test_float8_semi_sparse.py
```diff
@@ -0,0 +1,74 @@
...
+        input_fp8 = Float8Tensor.from_hp(
+            input, float8_dtype=torch.float8_e4m3fn, mm_config=mm_config
+        )
+
+        weight_fp8 = Float8Tensor.from_hp(
+            linear.weight.data, float8_dtype=torch.float8_e4m3fn, mm_config=mm_config
+        )
+        dense_output = torch.nn.functional.linear(input_fp8, weight_fp8, linear.bias)
+
+        weight_sparse_fp8 = Float8SemiSparseTensor.from_hp(linear.weight.data, [1, K])
```
jcaip, line 62: nit: .detach() instead of .data?

torchao/quantization/quantize_/workflows/float8/float8_semi_sparse_tensor.py
```diff
@@ -0,0 +1,153 @@
...
+    _quantize_affine_float8,
+)
+from torchao.utils import TorchAOBaseTensor
+
+__all__ = ["Float8SemiSparseTensor"]
+aten = torch.ops.aten
+
+
+class Float8SemiSparseTensor(TorchAOBaseTensor):
+    tensor_data_names = ["sparse", "meta", "scale"]
```
jcaip, line 22: nit: we should use [ `compressed_values`, `metadata` ]  instead of `sparse` and `meta` here.

torchao/quantization/quantize_/workflows/float8/float8_semi_sparse_tensor.py
```diff
@@ -0,0 +1,153 @@
...
+            meta,
+            scale,
+        )
+
+
+implements = Float8SemiSparseTensor.implements
+implements_torch_function = Float8SemiSparseTensor.implements_torch_function
+
+
+@implements(aten.t.default)
```
jcaip, line 86: Why do you have to implement transpose? Transpose on a sparse matrix is kind of tricky, we should probably throw an `ValueError` if it's called.

torchao/quantization/quantize_/workflows/float8/float8_semi_sparse_tensor.py
```diff
@@ -0,0 +1,153 @@
...
+    return return_and_correct_aliasing(func, args, kwargs, new)
+
+
+def _linear_fp8_semi_sparse(input_tensor, weight_tensor, bias):
+    from torchao.ops import rowwise_scaled_linear_sparse_cutlass_f8f8
+    from torchao.quantization.quantize_.workflows.float8.float8_tensor import (
+        Float8Tensor,
+    )
+
+    if isinstance(input_tensor, Float8Tensor):
```
jcaip, line 105: nit: what does this conditional do?

torchao/quantization/quantize_/workflows/float8/float8_semi_sparse_tensor.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+@implements([aten.mm.default, aten.addmm.default])
+def _(func, types, args, kwargs):
+    if func == aten.addmm.default:
+        bias, input_tensor, weight_tensor = args
+    else:  # aten.mm.default
+        input_tensor, weight_tensor = args
+        bias = None
+
```
jcaip, line 135: I think you may need to do some transpose trickery here to support mm and addmm, 

my understanding is that `linear(x, w, bias)` will return `xW^t + bias` so for mm / addmm you need to pass in a transposed weight

jcaip (commented):
cc @bbeckca I think the implementation and test look good, left a couple of comments.

bbeckca:
Abandoning in favor of #3258.

PR #3183, opened 2025-10-15
[not4land] Some fixes for MXFP8

jerryzh168:


test/prototype/mx_formats/test_inference_workflow.py
```diff
@@ -108,18 +116,22 @@ def test_inference_workflow_mx(elem_dtype, bias: bool, compile: bool, emulate: b
...
     quantize_(m_mx, config=config)
+    print("m_mx:", m_mx)
+        
     if compile:
         m_mx = torch.compile(m_mx, fullgraph=True)
 
-    x = torch.randn(128, 32, device="cuda", dtype=torch.bfloat16)
-    y_ref = m(x)
-    y_mx = m_mx(x)
+    with torch.inference_mode():
```
vkuzo, line 124: is this what makes the test fail, without changes to the product code?

torchao/prototype/mx_formats/mx_tensor.py
```diff
@@ -752,6 +761,14 @@ def mx_addmm(func, types, args, kwargs):
     b = args[2]
     return _addmm_mx_dispatch(a, b, func, bias=bias)
 
+@implements([aten.linear.default])
+def mx_linear(func, types, args, kwargs):
```
vkuzo, line 765: I think the reshape should live here, the `addmm` op is supposed to have 2d inputs

torchao/prototype/mx_formats/mx_tensor.py
```diff
@@ -752,6 +761,14 @@ def mx_addmm(func, types, args, kwargs):
...
+    assert isinstance(args[0], torch.Tensor) and isinstance(args[1], MXTensor)
+    a = args[0]
+    b = args[1]
+    bias = args[2] if len(args) > 2 else None
+    return _addmm_mx_dispatch(a, b.t(), func, bias=bias)
```
vkuzo, line 770: IMO either don't pass `func` into `addmm_mx_dispatch`, or we should rename it to something like `addmm_or_linear_mx_dispatch`

PR #3189, opened 2025-10-16
Fix TORCHAO_SKIP_LOADING_SO_FILES behavior

andrewor14:
**Summary:** Today, if users want to force skip loading the .so files, they can pass in `TORCHAO_SKIP_LOADING_SO_FILES=1`. However, if they pass in `TORCHAO_SKIP_LOADING_SO_FILES=0` or `TORCHAO_SKIP_LOADING_SO_FILES=false`, we will still skip loading these files. This commit fixes this behavior by:

1. Renaming this env var to `TORCHAO_FORCE_SKIP_LOADING_SO_FILES`
2. Only accepting value of "1" for this env var

**Test Plan:**

```
$ TORCHAO_FORCE_SKIP_LOADING_SO_FILES=1 python -c "import torchao"
Skipping import of cpp extensions due to TORCHAO_FORCE_SKIP_LOADING_SO_FILES=1

# No effect
$ TORCHAO_FORCE_SKIP_LOADING_SO_FILES=0 python -c "import torchao"
$ TORCHAO_FORCE_SKIP_LOADING_SO_FILES=False python -c "import torchao"
$ TORCHAO_FORCE_SKIP_LOADING_SO_FILES=false python -c "import torchao"
```

### DIFF
--- a/torchao/__init__.py
+++ b/torchao/__init__.py
@@ -45,9 +45,12 @@ def _parse_version(version_string):
 
 
 skip_loading_so_files = False
-if bool(os.getenv("TORCHAO_SKIP_LOADING_SO_FILES", False)):
+force_skip_loading_so_files = (
+    os.getenv("TORCHAO_FORCE_SKIP_LOADING_SO_FILES", "0") == "1"
+)
+if force_skip_loading_so_files:
     # user override
-    # users can set env var TORCH_INCOMPATIBLE=1 to skip loading .so files
+    # users can set env var TORCHAO_FORCE_SKIP_LOADING_SO_FILES=1 to skip loading .so files
     # this way, if they are using an incompatbile torch version, they can still use the API by setting the env var
     skip_loading_so_files = True
 # if torchao version has "+git", assume it's locally built and we don't know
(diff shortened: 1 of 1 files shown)


torchao/__init__.py
```diff
@@ -45,11 +45,16 @@ def _parse_version(version_string):
 
 
 skip_loading_so_files = False
-if bool(os.getenv("TORCHAO_SKIP_LOADING_SO_FILES", False)):
+user_skip_loading_so_files = os.getenv("TORCHAO_SKIP_LOADING_SO_FILES")
```
vkuzo: I think we should only enable it if it equals to `1`, to simplify this code

andrewor14: so if users set it to anything else other than "1", do you mean we should set this flag to False or ignore the setting?

vkuzo: the check should be `TORCHAO_SKIP_LOADING_SO_FILES == "1"`, not `bool(TORCHAO_SKIP_LOADING_SO_FILES)`

torchao/__init__.py
```diff
@@ -83,10 +86,15 @@ def _parse_version(version_string):
...
-        Please see GitHub issue #2919 for more info"
-    )
+    if force_skip_loading_so_files:
+        logger.warning(
+            "Skipping import of cpp extensions due to TORCHAO_FORCE_SKIP_LOADING_SO_FILES=1"
+        )
+    else:
+        logger.warning(
+            f"Skipping import of cpp extensions due to incompatible torch version {torch.__version__} for torchao version {__version__} \
+            Please see GitHub issue #2919 for more info"
```
vkuzo: can we change this to a URL to the issue people can copy-paste?

PR #3191, opened 2025-10-16
[Intel GPU] enable use of dinov2 models for offload benchmark_low_bit_adam

arlesniak:
Fix for the assertion:
`AssertionError: Input height (224) doesn't match model (518).`
when running benchmark_low_bit_adam with i.e. timm/vit_giant_patch14_dinov2.lvd142m (results for the offload benchmarks)

### DIFF
--- a/benchmarks/benchmark_low_bit_adam.py
+++ b/benchmarks/benchmark_low_bit_adam.py
@@ -18,6 +18,16 @@
 #   --amp bf16 \
 #   --optim AdamW
 #
+# To fine-tune a pre-trained ViT-Giant DINOv2 with full BF16, using AdamW optimizer with CPU offload
+# (https://github.com/pytorch/ao/blob/main/torchao/optim/README.md#optimizer-cpu-offload)
+# python benchmark_low_bit_adam.py \
+#   --model "timm/vit_giant_patch14_dinov2.lvd142m" \
+#   --full_bf16 \
+#   --optim AdamW \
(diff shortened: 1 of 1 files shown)


liangan1:
@arlesniak this is  the benchmark part, you also need to enable the realted UTs

arlesniak:
> @arlesniak this is the benchmark part, you also need to enable the realted UTs

The PR changes are solely for the purpose of the benchmark when dinov2 models are used with it. The root cause of the error was the hardcoded input size, fixed with more flexible solution.These changes don't affect the TorchAo functionality and IMHO are not to be unit tested.

EikanWang:
@liangan1 , @arlesniak , is the issue a XPU-specific issue? According to the error message, it should be a general issue, rigth?

arlesniak:
@EikanWang  It's a general issue, the fix is needed for the benchmark to be run with models on IntelGPU and others. (https://github.com/pytorch/ao/blob/main/torchao/optim/README.md)

benchmarks/benchmark_low_bit_adam.py
```diff
@@ -120,11 +120,18 @@ def get_parser():
 
 def get_dloader(args, training: bool):
     transforms = [v2.ToImage()]
+    input_size = (
```
jerryzh168, line 134: can you add a example benchmark command for xpu in the start of the file as a comment?

jerryzh168 (approved):

PR #3196, opened 2025-10-17
add option to save profiling traces in inference roofline script

vkuzo:
Summary:

convenient to analyze differences between roofline and observed

tl;dr; of findings:

mxfp8
1. need to pre-swizzle weights
2. torch.compile gives us two kernels, will repurpose the manual
   training kernel for this, will need to add pre-swizzling. Longer
   term, can see if fbgemm_gpu one is faster.

mxfp4
1. need to pre-swizzle weights
2. need a faster gemm (can use fbgemm_gpu)
3. need a fused activation quant kernel (can use fbgemm_gpu)

nvfp4
1. need to speed up existing triton activation quant kernel, currently
   it doesn't autotune anything so probably some easy wins here. Longer
   term can also benchmark vs fbgemm_gpu

Test Plan:

```bash
CUDA_VISIBLE_DEVICES=5 python benchmarks/float8/float8_inference_roofline.py ~/local/tmp/20251016_inference_nvfp4.csv --recipe_name nvfp4 --save_profile_traces True
```

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo:
* #3200
* #3198
* __->__ #3196

danielvegamyhre (approved):

PR #3198, opened 2025-10-17
mxtensor: make scale shape match qdata

vkuzo:
Summary:

Given `qdata` shape M, K, makes `scale` shape be M, K // block_size

This will be important to keep the logic sane as we add pre-swizzling
in a future PR.

Test Plan:

```
```

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo:
* #3202
* #3200
* __->__ #3198

PR #3200, opened 2025-10-17
mxtensor: add pre-swizzle support

vkuzo:
Summary:

Adds the ability to pre-swizzle scales for `MXTensor`,
and turns it on for the inference workflow.

For activations, this is no-change for now but if we write a fused
kernel we'll hook into the pre-swizzled path.

For weights, this is a performance win already as now we swizzle ahead of
time.

Rough magnitude of the weight pre-swizzling win:
on M, K, N == 4096, 4096, 4096, the inference fwd speedup on mxfp8
increases from 1.24x to 1.30x

Name of `_is_swizzled_scales` is not final, but IMO we should finalize it in a future PR together with NVFP4Tensor. For now I'm staying consistent with NVFP4Tensor.

Test Plan:

```bash
// correctness
CUDA_VISIBLE_DEVICES=5 pytest test/prototype/mx_formats/ -s

// performance
CUDA_VISIBLE_DEVICES=5 python benchmarks/float8/float8_inference_roofline.py ~/local/tmp/20251017_test.csv --recipe_name mxfp8_cublas --shape_gen_name pow2_extended
// before: https://www.internalfb.com/phabricator/paste/view/P1996942931
// after: https://www.internalfb.com/phabricator/paste/view/P1996941798
```

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo:
* #3202
* __->__ #3200

torchao/prototype/mx_formats/mx_tensor.py
```diff
@@ -526,31 +544,6 @@ def __new__(
             torch.float8_e5m2,
             torch.uint8,
         ), "unsupported"
-        if elem_dtype in (
```
vkuzo, line 529: this code doesn't really have a strong purpose, removing instead of making it handle swizzling

torchao/prototype/mx_formats/utils.py
```diff
@@ -190,5 +211,210 @@ def _to_mxfp8_dim1_kernel_wrapper(
             gemm_kernel_choice,
             False,
             None,
+            is_swizzled_scales,
         )
     return mx_tensor
+
+
+def _swizzle_aware_slice(
```
vkuzo, line 219: extracted this out, the only things that change are the various shape calculations (fp8 vs fp4 data, 32 vs 16 block size)

torchao/prototype/mx_formats/utils.py
```diff
@@ -190,5 +211,210 @@ def _to_mxfp8_dim1_kernel_wrapper(
...
+    x: torch.Tensor, dim, start, end, step
+) -> Tuple[torch.Tensor, torch.Tensor]:
+    """
+    Input: NVFP4Tensor or MXTensor
+    Output: sliced qdata and scale, does the right thing for unswizzled and swizzled scales
+    """
+
+    M, K = x.shape[0], x.shape[1]
```
drisspg, line 227: Nit I should probably have used a (m/n) generic term like rows, columns

PR #3209, opened 2025-10-18
require --no-build-isolation in torchao builds

vkuzo:
Summary:

In recent versions of `setuptools`, the behavior of using PEP 517 (enforce build isolation, https://peps.python.org/pep-0517/) flipped from False to True.  This broke torchao builds as our `setup.py` file depends on not having build isolation, as it imports build extension class definitions from `torch` (https://github.com/pytorch/ao/blob/94dee9c485f9bedad652b6f24ab90dc0dec6cafc/setup.py#L106).

A long term fix would be modernize the torchao build system to honor PEP 517.  However, to unbreak CI and buy us some time, in the short term we can just require build isolation for building torchao. This PR implements the short term workaround.

An AI-generated writeup I used to debug this, with more details: https://gist.github.com/vkuzo/5e6f282fda816c2cd4bce8c7fc1f30e2

Note that I had to add a `pyproject.toml` file to make `pip install .` run without warnings.

Test Plan:

```bash
// both of the commands below work
with-proxy USE_CPP=0 pip install -e . --no-build-isolation
with-proxy USE_CPP=1 pip install -e . --no-build-isolation --verbose

// will check CI
```

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/packaging/pre_build_script.sh
+++ b/packaging/pre_build_script.sh
@@ -17,4 +17,4 @@ else
 fi
 pip install $PYTORCH_DEP
 
-pip install setuptools wheel twine auditwheel
+pip install setuptools wheel twine auditwheel numpy
--- a/.github/scripts/ci_test_xpu.sh
+++ b/.github/scripts/ci_test_xpu.sh
@@ -8,7 +8,7 @@ export CXX=/usr/bin/g++
 export SCCACHE_DISABLE=1
 
 python3 -m pip install torch torchvision torchaudio pytorch-triton-xpu --index-url https://download.pytorch.org/whl/nightly/xpu --force-reinstall --no-cache-dir 
-cd torchao && python3 setup.py install && cd ..
+cd torchao && pip install . --no-build-isolation && cd ..
 
 python3 -c "import torch; import torchao; print(f'Torch version: {torch.__version__}')"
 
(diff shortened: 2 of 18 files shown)


vkuzo:
note: `test-cpu-ops` and`test-nightly` are failing in main branch for other reasons, so looks like CI is as green as it can be

vkuzo:
@atalman , I'm guessing if we go forward with this we'll also need to modify https://github.com/pytorch/test-infra/blob/a287d59209eb9221fa2428c3930efdc8643ad801/.github/workflows/build_wheels_linux.yml#L245 to use --no-build-isolation, gated to torchao only?

atalman:
HI @vkuzo test-infra will execute this build: https://github.com/pytorch/test-infra/blob/a287d59209eb9221fa2428c3930efdc8643ad801/.github/workflows/build_wheels_linux.yml#L264

```
 ${CONDA_RUN} python setup.py clean
 ${CONDA_RUN} python setup.py bdist_wheel
```

This is still supported

packaging/pre_build_script.sh
```diff
@@ -17,4 +17,4 @@ else
 fi
 pip install $PYTORCH_DEP
 
-pip install setuptools wheel twine auditwheel
+pip install setuptools wheel twine auditwheel numpy
```
vkuzo, line 20: `numpy` was added to DTensor here: https://github.com/pytorch/pytorch/pull/165451, so now we need to add it to the environment for successful wheel builds

vkuzo:
CI failures are unrelated, merging

PR #3215, opened 2025-10-20
Add per tensor fp8 quantization support for conv3d

jerryzh168:
Summary:
att, we added support of quantization conv3d weights, with Float8DynamicActivationFloat8WeightConfig

API:
```
config = Float8DynamicActivationFloat8WeightConfig(
   granularity=PerTensor(),
)

_is_conv3d = lambda m, fqn: isinstance(m, torch.nn.Conv3d)

quantize_(quantized_model, config, filter_fn=_is_conv3d)
```

Test Plan:
pytest test/quantization/quantize_/workflows/float8/test_float8_tensor.py -k test_fp8_conv_variants

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/quantization/quant_api.py
+++ b/torchao/quantization/quant_api.py
@@ -1813,7 +1813,12 @@ def _float8_dynamic_activation_float8_weight_quantize_tensor(weight, config):
     _check_hardware_support(granularity)
     activation_granularity, weight_granularity = granularity
 
-    if not _fp8_mm_compat(weight):
+    if weight.dim() == 5:
+        # weights for conv3d
+        assert isinstance(activation_granularity, PerTensor) and isinstance(
+            weight_granularity, PerTensor
+        ), "5D tensor only supports per tensor activation and weight quantization"
+    elif not _fp8_mm_compat(weight):
         # TODO(future PR): this should really throw an exception instead of silently
         # not doing what the user asked
         return weight
--- a/torchao/quantization/quantize_/workflows/float8/float8_tensor.py
+++ b/torchao/quantization/quantize_/workflows/float8/float8_tensor.py
@@ -39,6 +39,7 @@ from torchao.utils import (
     _is_fbgemm_gpu_genai_available,
     fill_defaults,
     is_sm_at_least_90,
+    is_sm_at_least_100,
 )
 
 __all__ = [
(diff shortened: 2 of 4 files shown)


jerryzh168:
currently waiting for the fbgemm conv op to be available in nightly

`uv pip install --pre fbgemm-gpu-genai --index-url https://download.pytorch.org/whl/nightly/cu128`

torchao/quantization/quant_api.py
```diff
@@ -1813,7 +1813,7 @@ def _float8_dynamic_activation_float8_weight_quantize_tensor(weight, config):
     _check_hardware_support(granularity)
     activation_granularity, weight_granularity = granularity
 
-    if not _fp8_mm_compat(weight):
+    if weight.dim() != 5 and not _fp8_mm_compat(weight):
```
andrewor14: 5 seems a bit arbitrary here without the context, should we add a comment that this is for conv3d?

torchao/utils.py
```diff
@@ -32,6 +32,7 @@
     "is_MI300",
     "is_sm_at_least_89",
     "is_sm_at_least_90",
+    "is_sm_at_least_100",
```
andrewor14, line 35: not super related to this PR but I wonder if we should stop exposing these, we don't expect users to call these themselves

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
```diff
@@ -418,6 +418,108 @@ def _(func, types, args, kwargs):
...
 
 
+def _scaled_conv3d(
+    input_tensor,
+    weight_tensor,
+    bias,
+    stride,
+    padding,
+    dilation,
+):
```
andrewor14, line 429: do we also need to check kernel preference? Like if it's "torch" maybe we should throw an exception since we don't support that yet?

torchao/quantization/quant_api.py
jerryzh168: OK added

torchao/utils.py
jerryzh168: oh, what should user call?

jerryzh168: oh should we use `is_sm_version`? but I think we need `is_sm_at_least`

andrewor14: oh I mean should users even call these helper functions? They know what GPUs they're running on. If they really want to check then maybe they should just check `torch.cuda.get_device_capability() >= (10, 0)` themselves instead of importing our utils. I'd like to keep them private if possible (in a separate PR)

jerryzh168: oh I see, makes sense, yeah this should just be dev only, not user facing

PR #3220, opened 2025-10-21
fix is metadata func

liangel-02:
function should only return false if tensors are not in the allowed subclasses **and** if it is not a regular torch.Tensor

**Testing**
`python test/prototype/safetensors/test_safetensors_utils.py` (fails without this change)


jerryzh168:
any unit test for this change?

PR #3221, opened 2025-10-21
fix mxfp8 matmul benchmark

vkuzo:
Summary:

Adds padding to the scales to properly support shapes where M % 128 != 0

Test Plan:

```
python benchmarks/float8/bench_matmul.py --shape_gen_name custom --recipe mxfp8_cublas --M 17 --K 32 --N 16
```

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo:
> only scale need to be padded?

this fix just aligns the benchmark to use the same padding as is used in the actual code in MXTensor, the benchmark was incorrect if M % 128 != 0

separately, we could also add padding to the underlying data - we can do this in a future PR.

PR #3240, opened 2025-10-24
nvfp4: support inference_mode and rank 3

vkuzo:
Summary:

Adds support for inference mode and rank 3 inputs for nvfp4 inference

Test Plan:

```
pytest test/prototype/mx_formats/test_inference_workflow.py -s -x -k test_inference_workflow_nvfp4
```

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/prototype/test_tensor_conversion.py
+++ b/test/prototype/test_tensor_conversion.py
@@ -34,7 +34,10 @@ from torchao.quantization.quantize_.workflows.intx.intx_opaque_tensor import (
     _is_kernel_library_loaded,
 )
 from torchao.quantization.utils import compute_error
-from torchao.utils import _is_fbgemm_gpu_genai_available
+from torchao.utils import (
+    _is_fbgemm_gpu_genai_available,
+    is_sm_at_least_90,
+)
 
 
 class ToyLinearModelWithTiedEmbedding(torch.nn.Module):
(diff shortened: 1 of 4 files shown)


vkuzo:
* __->__ #3240

test/prototype/mx_formats/test_inference_workflow.py
```diff
@@ -172,6 +172,8 @@ def test_inference_workflow_mx(
     ],
     ids=lambda s: f"{s[0]}x{s[1]}x{s[2]}",
 )
+@pytest.mark.parametrize("use_inference_mode", [False, True])
+@pytest.mark.parametrize("x_rank", [2, 3])
```
jerryzh168, line 176: nit: maybe instead of x_rank, we can follow this: https://github.com/pytorch/ao/blob/e9c7bead90b840b280f97374308255957108ce47/test/quantization/quantize_/workflows/float8/test_float8_tensor.py#L76-L77 ?

torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -576,15 +577,19 @@ def nvfp4_linear(func, types, args, kwargs):
             tensor_amax = torch.max(torch.abs(input_tensor))
             per_tensor_scale = per_tensor_amax_to_scale(tensor_amax)
         else:
-            per_tensor_scale = weight_tensor._act_per_tensor_scale
+            per_tensor_scale = weight_tensor.act_per_tensor_scale
```
jerryzh168, line 580: btw, I saw this function is still using `implements` for both F.linear and aten.linear somehow, it should be changed to https://github.com/pytorch/ao/blob/e9c7bead90b840b280f97374308255957108ce47/torchao/quantization/quantize_/workflows/float8/float8_tensor.py#L251-L252

I guess this should be caught by test as well, maybe we should add a test that can catch this usage

jerryzh168: oh it's because this subclass is not migrated yet, then it's fine

PR #3241, opened 2025-10-24
introduce new int8 quantization API

namgyu-youn:
**Summary**:
Introduce a new tensor subclass API. The main features are
- `Int8Tensor`: Main API, which handles quantization and dequantization operations
- Utility operation functions: Tensor slice, index selection

This api is integrated into global variants (`Int8WeightOnlyConfig`, `Int8DynamicActivationInt8WeightConfig`) using `version`, and not defined as a default.

**Related Issue/PR**: #3038 (reland)

**Test plan**:
test/quantization/quantize_/workflows/int8/test_int8_tensor.py

**Performance**:
The following are the results of https://github.com/pytorch/ao/blob/main/tutorials/quantize_vit/run_vit_b_quant.py with a batch size of 32:

| API | With `torch.compile` | Without `torch.compile` |
|-----|---------------------|------------------------|
| Old | 65.47 ms | 234.39 ms |
| New | 63.30 ms | 239.30 ms |

### DIFF
--- a/torchao/quantization/quantize_/workflows/__init__.py
+++ b/torchao/quantization/quantize_/workflows/__init__.py
@@ -38,7 +38,6 @@ __all__ = [
     "Int4MarlinSparseTensor",
     "Int4PlainInt32Tensor",
     "Int4TilePackedTo4dTensor",
-    "Float8OpaqueTensor",
     "Int8Tensor",
     "QuantizeTensorToInt8Kwargs",
     "Float8Tensor",


test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -0,0 +1,266 @@
...
+        dynamic_tensor = Int8Tensor.from_hp(weight, block_size=[N, K])
+
+        # Static quantization (pre-computed scale)
+        act_scale, _ = choose_qparams_affine(
+            input=input_tensor,
+            mapping_type=MappingType.SYMMETRIC,
+            block_size=(input_tensor.shape[0], K),
+            target_dtype=torch.int8,
+            quant_min=self.int8_min,
+            quant_max=self.int8_max,
```
jerryzh168: nit: we can omit these two args if these are the same as default (-128, 127)

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -0,0 +1,266 @@
...
+        output_original = m(input_tensor)
+        output_quantized = m_q(input_tensor)
+
+        error = compute_error(output_original, output_quantized)
+        assert error > 20, (
+            f"Quantization quality is too low, SQNR: {error}dB (expected > {20}dB)"
+        )
+
+    @common_utils.parametrize("dtype", [torch.bfloat16, torch.float16])
+    def test_quantization_shapes(self, dtype):
```
jerryzh168: this seems to be a combination of two tests, one for dynamic quant one for static quant, can you use something like this: https://github.com/pytorch/ao/blob/e9c7bead90b840b280f97374308255957108ce47/test/quantization/quantize_/workflows/float8/test_float8_tensor.py#L65

also I feel it might be better to not add static quant in this PR, and in a separate PR add both the tensor support and config support for static quant

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,300 @@
...
+        block_size: list[int],
+        act_quant_kwargs: Optional[QuantizeTensorToInt8Kwargs] = None,
+    ):
+        if w.dim() != 2 or len(block_size) != 2:
+            raise ValueError("Expected 2D tensor and block_size length 2")
+
+        if act_quant_kwargs is not None and act_quant_kwargs.static_scale is not None:
+            # INT8 × INT8 (static)
+            scale = act_quant_kwargs.static_scale
+            zero_point = torch.zeros_like(scale, dtype=torch.int8)
```
jerryzh168: I think user should specify static_zero_point as well

but again, it's better to do this in a separate PR, since current state is a half of the static quant feature (no config)

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,300 @@
...
+        if is_per_tensor_weight:
+            result = y_dot.mul_(w_scales.to(intermediate_dtype))
+        else:
+            # Per-row weight scale - transpose and broadcast
+            w_scales_broadcast = w_scales.t().expand_as(y_dot)
+            result = y_dot.mul_(w_scales_broadcast.to(intermediate_dtype))
+
+        # Reshape back to original shape
+        result = result.view(*x_vals.shape[:-1], result.shape[-1])
+        result = result.to(activation_tensor.dtype)
```
jerryzh168: this should follow: https://github.com/pytorch/ao/blob/e9c7bead90b840b280f97374308255957108ce47/torchao/dtypes/uintx/plain_layout.py#L281

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,300 @@
...
+    else:
+        # FP × INT8 (weight-only)
```
jerryzh168, line 278: this should follow https://github.com/pytorch/ao/blob/e9c7bead90b840b280f97374308255957108ce47/torchao/dtypes/uintx/plain_layout.py#L250

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
namgyu-youn: Okay, not sure to remove static flags (although its not fully implemented) before, but small PR should be always better I feel. I will remove `static_scale` and all those supports.

torchao/float8/inference.py
```diff
@@ -140,7 +140,18 @@ def _slice_scale_for_dimension(
     """
     aten = torch.ops.aten
 
-    # Unsupported case for now, this would be 1 scale per data element
+    # Per-tensor quantization (scalar scale)
```
jerryzh168: is this change related?

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -0,0 +1,235 @@
...
+
+    @common_utils.parametrize("dtype", [torch.bfloat16, torch.float16])
+    @common_utils.parametrize(
+        "config",
+        [
+            Int8DynamicActivationInt8WeightConfig(version=2),
+            Int8WeightOnlyConfig(version=2),
+        ],
+    )
+    def test_per_row_scale_shape(self, dtype, config):
```
jerryzh168: can you add a test like this one https://github.com/pytorch/ao/blob/1e473ed94caa060b3fbac96657030c44173759e0/test/quantization/quantize_/workflows/float8/test_float8_tensor.py#L80 to test all the variations and include the shape check there?

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -0,0 +1,235 @@
...
+        # Dynamic: per-row (1D scale [N]), Weight-only: per-tensor (scalar)
+        if isinstance(config, Int8DynamicActivationInt8WeightConfig):
+            self.assertEqual(linear.weight.scale.shape, (N,))
+            self.assertEqual(linear.weight.scale.ndim, 1)
+        else:
+            self.assertEqual(linear.weight.scale.numel(), 1)
+
+    @common_utils.parametrize("dtype", [torch.bfloat16, torch.float16])
+    @common_utils.parametrize("has_bias", [True, False])
+    def test_weight_only_linear_with_bias(self, dtype, has_bias):
```
jerryzh168: this can probably be merged into the linear varaints test as well

jerryzh168 (commented):
thanks, I think the tensor changes looks good, but need to make a linear_variants tests to make sure we cover different aspects of things (e.g. compile), see comments inline


can you also do a e2e perf check with https://github.com/pytorch/ao/blob/main/tutorials/quantize_vit/run_vit_b_quant.py to make sure the performance are the same before and after change for vit model?

also add a kernel check might be useful to make sure we don't regress things: https://github.com/pytorch/ao/blob/1e473ed94caa060b3fbac96657030c44173759e0/test/quantization/quantize_/workflows/float8/test_float8_tensor.py#L422

torchao/float8/inference.py
namgyu-youn: ~It is updated to support more granularity. Without this change, we can't use per-tensor (0D scale) and per-row (1D scale).~

above comment is incorrect and this change is unrelated; https://github.com/pytorch/ao/pull/3241/#discussion_r2515319920

namgyu-youn:
Updated logs:
- Support 3D Tensor input using reshape, similar to `Flot8Tensor`
- Build integrated linear variants test: https://github.com/pytorch/ao/pull/3241#discussion_r2479708489, https://github.com/pytorch/ao/pull/3241#discussion_r2479709364
- Add available kernel test: https://github.com/pytorch/ao/pull/3241#pullrequestreview-3401951826
- Add PERF result in PR contents: https://github.com/pytorch/ao/pull/3241#pullrequestreview-3401951826

Xia-Weiwen:
Hi @namgyu-youn Do you plan to submit another PR for static quantization? We also need static quantization for SmoothQuant. So, we are wondering if you have a plan or we should consider adding it ourselves. Thanks. CC @cyxlily

namgyu-youn:
> Hi @namgyu-youn Do you plan to submit another PR for static quantization? We also need static quantization for SmoothQuant. So, we are wondering if you have a plan or we should consider adding it ourselves. Thanks. CC @cyxlily

Yeah, static quantization support using static/dynamic flags is planned; I hope to show it to your team in the foreseeable future.

Also, in the SmoothQuant case, validating its support for the new quantization APIs (below) has higher priority, I think. Could you look into it?
- W4A16-INT: `Int4WeightOnlyConfig(group_size=32, version=2)`
- W4A16-FP: `Float8WeightOnlyConfig(version=2)`
- W8A8-FP-dynamic: `Float8DynamicActivationFloat8WeightConfig(version=2)`

Xia-Weiwen:
> Yeah, static quantization support using static/dynamic flags is planned; I hope to show it to your team in the foreseeable future.

Thanks. Looking forward to it. If there is anything we can help with, please let us know.
 
> Also, in the SmoothQuant case, validating its support for the new quantization APIs (below) has higher priority, I think. Could you look into it?
> 
> * W4A16-INT: `Int4WeightOnlyConfig(group_size=32, version=2)`
> * W4A16-FP: `Float8WeightOnlyConfig(version=2)`
> * W8A8-FP-dynamic: `Float8DynamicActivationFloat8WeightConfig(version=2)`

By "validating them", do you mean adding test cases? And are W4A16 and W8A16 (I guess there is a typo in your comment) really needed for SmoothQuant? For W4A16 , it would be much the same as AWQ. And for W8A16, I think accuracy is generally good enough without SmoothQuant.

namgyu-youn:
> By "validating them", do you mean adding test cases? And are W4A16 and W8A16 (I guess there is a typo in your comment) really needed for SmoothQuant? For W4A16 , it would be much the same as AWQ. And for W8A16, I think accuracy is generally good enough without SmoothQuant.

Oh yes, it was a typo (W8A16 is right), and W4A16-INT (`Int4WeightOnlyConfig(group_size=32, version=2)`) is of interest. In my last experience and https://arxiv.org/html/2411.02355v3, W4A16-INT is the most efficient choice for synchronous deployments, while W8A8-INT maximize throughput in asynchronous settings.

Because current AWQ/SmoothQuant test is only working with old APIs (version 1), we can replace it with new APIs like `Int4WeightOnlyConfig(group_size=32, version=2)` I guess.

Xia-Weiwen:
> > By "validating them", do you mean adding test cases? And are W4A16 and W8A16 (I guess there is a typo in your comment) really needed for SmoothQuant? For W4A16 , it would be much the same as AWQ. And for W8A16, I think accuracy is generally good enough without SmoothQuant.
> 
> Oh yes, it was a typo (W8A16 is right), and W4A16-INT (`Int4WeightOnlyConfig(group_size=32, version=2)`) is of interest. In my last experience and https://arxiv.org/html/2411.02355v3, W4A16-INT is the most efficient choice for synchronous deployments, while W8A8-INT maximize throughput in asynchronous settings.
> 
> Because current AWQ/SmoothQuant test is only working with old APIs (version 1), we can replace it with new APIs like `Int4WeightOnlyConfig(group_size=32, version=2)` I guess.

I see. Thanks. We will evaluate that.

torchao/float8/inference.py
Xia-Weiwen: So maybe it's better to move this util function to a common place?

Xia-Weiwen:
Hi @namgyu-youn May I know if you have a timeline to land this? Thanks.

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -0,0 +1,291 @@
...
+        ],
+    )
+    @common_utils.parametrize(
+        "sizes",
+        [
+            ((128,), 256, 128),  # 2D
+            ((32, 128), 64, 256),  # 3D
+        ],
+    )
+    def test_int8_linear_quantization_accuracy(
```
jerryzh168: can this one be combined with `test_int8_linear_variants` as well?

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -0,0 +1,291 @@
...
+            )
+            self.assertEqual(weight2.scale, dummy.weight.scale)
+        else:
+            # PerTensor: scale unchanged by slicing
+            self.assertEqual(weight1.scale, dummy.weight.scale)
+            self.assertEqual(weight2.scale, dummy.weight.scale)
+        with self.assertRaises(NotImplementedError):
+            _ = dummy.weight[::2]
+
+    def test_index_select(self):
```
jerryzh168: I think all the tests should be modified to use config as well

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -0,0 +1,291 @@
...
+        config = Int8DynamicActivationInt8WeightConfig(version=2)
+        quantize_(m, config)
+        m = torch.compile(m)
+        x = torch.randn(M, K, device="cuda", dtype=torch.bfloat16)
+
+        out, code = run_and_get_code(m, x)
+        kernels = {}
+
+        # Check for Triton kernels
+        if "torch.ops.triton" in code[0]:
```
jerryzh168: should add some asserts I think

namgyu-youn:
Updated logs:
- Replace `Int8Tensor` to global configs in test: https://github.com/pytorch/ao/pull/3241#discussion_r2488365752
- Consolidate linear variation tests: https://github.com/pytorch/ao/pull/3241#discussion_r2488364893
- Update parser for kernel detection test: https://github.com/pytorch/ao/pull/3241#discussion_r2488370485

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -0,0 +1,262 @@
...
+        quantize_(dummy, config)
+
+        weight1 = dummy.weight.clone().narrow(0, 0, slice_sizes[0])
+        weight2 = dummy.weight.clone().narrow(1, 0, slice_sizes[1])
+
+        self.assertEqual(weight1.qdata, dummy.weight.qdata.narrow(0, 0, slice_sizes[0]))
+        self.assertEqual(weight2.qdata, dummy.weight.qdata.narrow(1, 0, slice_sizes[1]))
+
+        # Int8DynamicActivationInt8WeightConfig uses per-row (PerRow)
+        # Int8WeightOnlyConfig uses per-tensor (PerTensor)
```
jerryzh168: it should be per row I think? https://github.com/pytorch/ao/blob/6815e5713b2021d3e065ec77765e2ab47b1319b6/torchao/quantization/quant_api.py#L1343

torchao/quantization/quant_api.py
```diff
@@ -1363,22 +1365,30 @@ def __post_init__(self):
...
+            weight,
+            mapping_type,
+            block_size,
+            target_dtype,
+            eps=eps,
+            zero_point_dtype=zero_point_dtype,
+        )
+    else:
+        assert config.version == 2, f"Unexpected version: {config.version}"
+        block_size = [weight.shape[0], weight.shape[1]]
```
jerryzh168: why does this default to per tensor? I think it should follow the existing logic from L1376-1378?

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -0,0 +1,262 @@
...
+        quantize_(linear, config)
+
+        tensor = linear.weight
+        dequantized = tensor.dequantize()
+        self.assertEqual(dequantized.shape, test_data.shape)
+        self.assertLess(
+            torch.abs(dequantized - test_data).max().item(),
+            0.1,
+            msg=f"Dequantization error exceeds tolerance of {0.1}",
+        )
```
jerryzh168: maybe check sqnr with `compute_error` instead of hardcoded absolute value?

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -0,0 +1,262 @@
...
+
+        out, code = run_and_get_code(m, x)
+        has_triton = "triton" in code[0].lower()  # Trition
+        has_fbgemm = "fbgemm" in code[0].lower()  # FB-GEMM
+        has_int_mm = "_int_mm" in code[0]  # Int8 MatMul
+
+        self.assertTrue(
+            has_triton or has_fbgemm or has_int_mm,
+            f"No int8 quantization kernels found. has_triton={has_triton}, has_fbgemm={has_fbgemm}, has_int_mm={has_int_mm}",
+        )
```
jerryzh168: is it possible to be more strict to spell out the kernels called and their order? like https://github.com/pytorch/ao/blob/6815e5713b2021d3e065ec77765e2ab47b1319b6/test/quantization/quantize_/workflows/float8/test_float8_tensor.py#L596-L600

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -0,0 +1,262 @@
...
+            torch.nn.Linear(K, N, device="cuda", dtype=torch.bfloat16)
+        )
+        config = Int8DynamicActivationInt8WeightConfig(version=2)
+        quantize_(m, config)
+        m = torch.compile(m)
+        x = torch.randn(M, K, device="cuda", dtype=torch.bfloat16)
+
+        out, code = run_and_get_code(m, x)
+        has_triton = "triton" in code[0].lower()  # Trition
+        has_fbgemm = "fbgemm" in code[0].lower()  # FB-GEMM
```
jerryzh168: we don't expect fbgemm kernel to be used here?

torchao/float8/inference.py
jerryzh168: this can be moved to torchao/quantization/quantize_/common/utils.py I think

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,300 @@
...
+
+from dataclasses import dataclass
+from typing import Optional
+
+import torch
+from torch.utils._python_dispatch import return_and_correct_aliasing
+
+from torchao.float8.inference import (
+    _slice_scale_for_dimension,
+    preprocess_scale,
```
jerryzh168: why are these used?

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,300 @@
...
+                else:
+                    block_size_updated = list(act_kwargs.block_size)[-input_ndim:]
+                act_kwargs = QuantizeTensorToInt8Kwargs(block_size=block_size_updated)
+
+            activation_tensor = _choose_quant_func_and_quantize_tensor(
+                activation_tensor, act_kwargs
+            )
+
+        x_vals = activation_tensor.qdata.reshape(-1, activation_tensor.qdata.shape[-1])
+        x_scales = preprocess_scale(activation_tensor.scale, x_vals.shape)
```
jerryzh168: why is this called? can you match the code with can you match the code with https://github.com/pytorch/ao/blob/6815e5713b2021d3e065ec77765e2ab47b1319b6/torchao/dtypes/uintx/plain_layout.py#L281 exactly?

torchao/float8/inference.py
namgyu-youn: Okay, then I will move this to `torchao/quantization/quantize_/common/utils.py` after this PR.

torchao/quantization/quant_api.py
namgyu-youn: I remember W4A16-INT used per-tensor granularity in test code, but per-row should be right based on version 1; thanks for pointing out.

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
namgyu-youn: I used `preprocess_scale` based on `Float8Tensor`, but `preprocess_scale` is ambiguous and we can follow UINT pattern as you mentioned I feel.

namgyu-youn:
Updated logs:
- Set per-row as a default granularity in both api and test: https://github.com/pytorch/ao/pull/3241#discussion_r2501227851 https://github.com/pytorch/ao/pull/3241#discussion_r2501224824
- Update kernel (trition, int_mm) test using parameterization and `FileCheck`: https://github.com/pytorch/ao/pull/3241#discussion_r2501233098 https://github.com/pytorch/ao/pull/3241#discussion_r2501232521
- Fix quantization ops similar to UINT pattern: https://github.com/pytorch/ao/pull/3241#discussion_r2501237961 https://github.com/pytorch/ao/pull/3241#discussion_r2461423604 https://github.com/pytorch/ao/pull/3241#discussion_r2501238457

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -0,0 +1,243 @@
...
+@common_utils.instantiate_parametrized_tests
+class TestInt8Tensor(TorchAOIntegrationTestCase):
+    def setUp(self):
+        super().setUp()
+
+        self.test_shape = (32, 20)
+        self.dtype = torch.bfloat16
+        self.batch_size = 32
+
+        torch.manual_seed(42)
```
jerryzh168, line 36: nit: this can probably be moved to `TorchAOIntegrationTestCase`
can also add https://github.com/pytorch/ao/commit/8e3b3da5444e69434a5a0997767d6a38a8b40bf8

feel free to do in a separate PR though

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -0,0 +1,243 @@
...
+    def test_creation_and_attributes(self, config):
+        """Test tensor creation, dtypes, and ranges"""
+        linear = torch.nn.Linear(
+            self.test_shape[1],
+            self.test_shape[0],
+            bias=False,
+            dtype=self.dtype,
+            device="cuda",
+        )
+        linear.weight.data = self.weight_fp.cuda()
```
jerryzh168: why do we do this? doesn't sound very necessary?

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -0,0 +1,243 @@
...
+
+    @common_utils.parametrize("dtype", [torch.bfloat16, torch.float16])
+    @common_utils.parametrize(
+        "config",
+        [
+            Int8DynamicActivationInt8WeightConfig(version=2),
+            Int8WeightOnlyConfig(version=2),
+        ],
+    )
+    def test_per_row_scale_shape(self, dtype, config):
```
jerryzh168: nit: can you merge the checks in this test to previous test `test_int8_linear_variants`?

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -0,0 +1,243 @@
...
+        quantize_(linear, config)
+
+        tensor = linear.weight
+        dequantized = tensor.dequantize()
+        self.assertEqual(dequantized.shape, test_data.shape)
+        assert compute_error(dequantized, test_data) > 20, (
+            f"Dequantization error is too high to get a SQNR of {compute_error(dequantized, test_data)}"
+        )
+
+    @common_utils.parametrize(
```
jerryzh168: does this have to be parametrize? I think what we need here is to check the code contains a sequence of ops / kernel calls, like this: https://github.com/pytorch/ao/blob/b4ec4cbab799b5320fc1ca094e0398b3ad967dbf/test/quantization/quantize_/workflows/float8/test_float8_tensor.py#L827-L831

I think we can check 1. the quantize op and then 2. the mm op `extern_kernels._int_mm`, in a single run (see example), that should be enough

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -0,0 +1,243 @@
...
+        "config",
+        [
+            Int8DynamicActivationInt8WeightConfig(version=2),
+            Int8WeightOnlyConfig(version=2),
+        ],
+    )
+    def test_dequantization_accuracy(self, config):
```
jerryzh168, line 178: additional comments for this test, 1. increase the size of linear? 2. I think we don't need to overwrite the weight, we can just save the floating point weight (deepcopy) before quantization and compare the results

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -0,0 +1,243 @@
...
+
+        self.test_shape = (32, 20)
+        self.dtype = torch.bfloat16
+        self.batch_size = 32
+
+        torch.manual_seed(42)
+        self.weight_fp = torch.randn(*self.test_shape, dtype=self.dtype)
+        self.input_fp = torch.randn(*self.test_shape, dtype=self.dtype)
+        self.bias = torch.randn(self.test_shape[0], dtype=self.dtype)
+        self.block_size = list(self.test_shape)
```
jerryzh168: I feel we probably don't need these, it's also easier for people to follow to define everything / most of things in the test itself

torchao/float8/inference.py
```diff
@@ -140,7 +140,18 @@ def _slice_scale_for_dimension(
...
+    if scale.numel() == 1:
```
jerryzh168: note: I think we can just check for ndim consistently everywhere, after https://github.com/pytorch/ao/issues/3324 is fixed

jerryzh168: also isn't handling for per tensor and per row already included in original code? https://github.com/pytorch/ao/blob/b4ec4cbab799b5320fc1ca094e0398b3ad967dbf/torchao/float8/inference.py#L158-L178

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
jerryzh168: yeah we don't need this I think, just need to follow original code exactly

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,301 @@
...
+            # Dynamic activation quantization
+            act_kwargs = weight_tensor.act_quant_kwargs
+            input_ndim = activation_tensor.ndim
+
+            # Ensure block_size matches input tensor dimensions
+            if len(act_kwargs.block_size) != input_ndim:
+                if input_ndim == 3 and len(act_kwargs.block_size) == 2:
+                    block_size_updated = [1] + list(act_kwargs.block_size)
+                else:
+                    block_size_updated = list(act_kwargs.block_size)[-input_ndim:]
```
jerryzh168: we kind of changed the meaning of block_size used in PerBlock quant recently, check https://github.com/pytorch/ao/blob/b4ec4cbab799b5320fc1ca094e0398b3ad967dbf/torchao/quantization/utils.py#L707

when is this code needed?

jerryzh168: in principle we shouldn't update block_size here, but instead, make sure block_size make sense and is consistent throughout the code base

torchao/quantization/quant_api.py
```diff
@@ -1559,19 +1572,43 @@ def get_weight_block_size(x):
...
+    else:
+        from torchao.quantization.quantize_.workflows.int8.int8_tensor import (
+            QuantizeTensorToInt8Kwargs,
+        )
+
+        assert config.version == 2, f"Unexpected version: {config.version}"
+        quantized_weight = Int8Tensor.from_hp(
+            weight,
+            block_size,
+            act_quant_kwargs=QuantizeTensorToInt8Kwargs(block_size=block_size),
```
jerryzh168: why does activation use the same block_size as weight?

jerryzh168: I think we can change block_size argument to `granularity`, if the block_size is unknown for activation since the shape is unknown

jerryzh168 (changes requested):
please rebase, also some comments are not addressed yet I think

namgyu-youn:
> please rebase, also some comments are not addressed yet I think

Thanks for the fast review, but please wait until the review request. I will address all comments this weekend :)

torchao/float8/inference.py
namgyu-youn: Reverted; I miscalculated `block_sizes`, and this change is unrelated. The original code is already handling per-tensor and per-row.

Btw, this function can be improved by separating the granularity I think; please check #3345 for this update.

torchao/quantization/quant_api.py
```diff
@@ -1324,14 +1325,18 @@ class Int8WeightOnlyConfig(AOBaseConfig):
...
+        group_size (version 1) - Controls the granularity of quantization.
+        If None, applies per-channel quantization. Otherwise, applies per-group quantization with the specified group size.
+        granularity (version 2) - Quantization granularity.
+            PerRow() for per-channel quantization, PerTensor() for per-tensor quantization.
         set_inductor_config: bool = True - If True, adjusts `torchinductor` settings to recommended values
             for better performance with this quantization scheme.
     """
 
     group_size: Optional[int] = None
+    granularity: Optional[Union[PerRow, PerTensor]] = PerRow()
```
jerryzh168: nit: the type can just be `Optional[Granularity]` I think, we can do validation on supported granularity later

torchao/quantization/quant_api.py
```diff
@@ -1492,12 +1504,15 @@ class Int8DynamicActivationInt8WeightConfig(AOBaseConfig):
             in original precision during decode operations.
         set_inductor_config: bool = True - If True, adjusts `torchinductor` settings to recommended values
             for better performance with this quantization scheme.
+        version (int): the version of the config, version 1 is using AffineQuantizedTensor that we plan to deprecate/split, version 2 is using Int8Tensor
     """
 
     layout: Optional[Layout] = PlainLayout()
     act_mapping_type: Optional[MappingType] = MappingType.SYMMETRIC
     weight_only_decode: bool = False
+    granularity: Optional[Union[PerRow, PerTensor]] = PerRow()
```
jerryzh168, line 1525: same here

also should this one be a tuple? since there is both activation and weight

or follow this: https://github.com/pytorch/ao/blob/8d546c83f6c8ec38c2c8ce0e41e35bcd12be841d/torchao/quantization/quant_api.py#L1804

torchao/quantization/quant_api.py
```diff
@@ -1545,19 +1557,46 @@ def get_weight_block_size(x):
...
+            quantized_weight, input_quant_func
+        )
+    else:
+        from torchao.quantization.quantize_.workflows.int8.int8_tensor import (
+            QuantizeTensorToInt8Kwargs,
+        )
+
+        assert config.version == 2, f"Unexpected version: {config.version}"
+        # Compute block_size from granularity for activation quantization kwargs
+        block_size = get_block_size(weight.shape, config.granularity)
```
jerryzh168: this is incorrect I think, we don't want to calculate the block_size for activation from weight

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,307 @@
...
+class QuantizeTensorToInt8Kwargs(QuantizeTensorKwargs):
+    """Tensor kwargs for creating int8 tensor (either activation or weight)
+
+    Args:
+        block_size (list[int]): block size for quantization granularity
+        granularity: the granularity for the Tensor, currently either PerRow() or PerTensor()
+        # TODO: Static quantization support using `static_scale`, `static_zero_point`
+    """
+
+    block_size: list[int]
```
jerryzh168: this should probably just take granularity

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,307 @@
...
+
+        if output_dtype is None:
+            output_dtype = self.dtype
+
+        qdata_fp = self.qdata.to(output_dtype)
+        scale = self.scale
+        while scale.ndim < qdata_fp.ndim:
+            scale = scale.unsqueeze(-1)
+
+        scale_expanded = _maybe_expand_scale_to_tensor_shape(scale, qdata_fp.shape)
```
jerryzh168: are these needed?

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,307 @@
...
+        return qdata_fp * scale_expanded.to(output_dtype)
```
jerryzh168: I think we should just call the `dequantize_affine` op directly https://github.com/pytorch/ao/blob/8d546c83f6c8ec38c2c8ce0e41e35bcd12be841d/torchao/quantization/quant_primitives.py#L704

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
jerryzh168: this comment is not addressed

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,307 @@
...
+        )
+        y_dot_scaled = y_dot_scaled.to(x_scales_dtype)
+
+        y = (y_dot_scaled * w_scales).reshape(
+            *x_vals_int8.shape[:-1], y_dot_scaled.shape[-1]
+        )
+
+        # can downcast only at the very end
+        output_dtype = activation_tensor.dtype
+        y = y.to(output_dtype)
```
jerryzh168: nit: can probably do this after bias addition

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,307 @@
...
+        w_vals_int8_t = weight_tensor.qdata.t()
+        m = torch.mm(
+            activation_tensor.reshape(-1, activation_tensor.shape[-1]),
+            w_vals_int8_t.to(activation_tensor.dtype),
+        )
+        y = m * weight_tensor.scale.to(m.dtype)
+        y = y.reshape(*activation_tensor.shape[:-1], weight_tensor.qdata.shape[0])
+        if bias is not None:
+            y += bias
+        return y
```
jerryzh168: can do the cast right before return I think

torchao/quantization/quant_api.py
namgyu-youn: Aren't we already testing `PerRow/PerTensor` in `test_index_select`? `Granularity` for type hint sounds too comprehensive I feel.

torchao/quantization/quant_api.py
namgyu-youn: @jerryzh168 How about dropping `block_size` from kwargs then? We might only go with `granularity`, and get block size by directly using `granularity (PerTensor/PerRow)` I guess.

namgyu-youn: Updated to use `Granularity` instead of `block_size` entirely, let me know if this move is right.

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
namgyu-youn: Yeah `_maybe_expand_scale_to_tensor_shape` is too abstract, and `dequantize_affine` should be better, thanks.

torchao/quantization/quant_api.py
jerryzh168: I think it's better to just do validate on the values instead of both value and type

torchao/quantization/quant_api.py
jerryzh168: here the type should be a Union of single value or a list (for the case when activation and weight do not have the same granularity)

torchao/quantization/quant_api.py
```diff
@@ -1545,19 +1557,44 @@ def get_weight_block_size(x):
...
+            QuantizeTensorToInt8Kwargs,
+        )
+
+        assert config.version == 2, f"Unexpected version: {config.version}"
+
+        quantized_weight = Int8Tensor.from_hp(
+            weight,
+            granularity=config.granularity,
+            act_quant_kwargs=QuantizeTensorToInt8Kwargs(granularity=config.granularity),
+        )
```
jerryzh168, line 1611: can you align the granularity handling with float8? https://github.com/pytorch/ao/blob/2ff1eb2e356275cfbe46832327387d382c72945d/torchao/quantization/quant_api.py#L1825, user should be able to specify different granularity for activation and weight

torchao/quantization/quantize_/common/quantize_tensor_kwargs.py
```diff
@@ -52,5 +54,11 @@ def _choose_quant_func_and_quantize_tensor(
             quant_kwargs.hp_value_ub,
             quant_kwargs.kernel_preference,
         )
+    elif isinstance(quant_kwargs, QuantizeTensorToInt8Kwargs):
+        return Int8Tensor.from_hp(
+            tensor,
+            granularity=quant_kwargs.granularity,
+            act_quant_kwargs=quant_kwargs,
```
jerryzh168: this should not include `act_quant_kwargs`, this function is used to quantize activation only

torchao/quantization/quant_api.py
```diff
@@ -1545,19 +1557,44 @@ def get_weight_block_size(x):
...
+        from torchao.quantization.quantize_.workflows.int8.int8_tensor import (
+            QuantizeTensorToInt8Kwargs,
+        )
+
+        assert config.version == 2, f"Unexpected version: {config.version}"
+
+        quantized_weight = Int8Tensor.from_hp(
+            weight,
+            granularity=config.granularity,
+            act_quant_kwargs=QuantizeTensorToInt8Kwargs(granularity=config.granularity),
```
jerryzh168, line 1610: how are these lines preserved in version 2? https://github.com/pytorch/ao/blob/2ff1eb2e356275cfbe46832327387d382c72945d/torchao/quantization/quant_api.py#L1568-L1571

torchao/quantization/quant_api.py
jerryzh168: yes that's correct, thanks

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,303 @@
...
+        # TODO: Static quantization support using `static_scale`
+
+    Non-Tensor Attributes:
+        granularity: the granularity for quantization (e.g., PerRow(), PerTensor())
+        act_quant_kwargs: flags for dynamic activation quantization
+    """
+
+    # TODO: Static quantization support using `static_scale`
+    tensor_data_names = ["qdata", "scale"]
+    tensor_attribute_names = ["granularity"]
```
jerryzh168, line 57: I think here it should store block_size to align with float8_tensor: https://github.com/pytorch/ao/blob/2ff1eb2e356275cfbe46832327387d382c72945d/torchao/quantization/quantize_/workflows/float8/float8_tensor.py#L246

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
namgyu-youn: Updated by dropping `block_size` from kwargs, thanks for teaching me.

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
namgyu-youn: Removed, this function is too abstract so we don't have to call it, similar to `preprocess_scale`

torchao/quantization/quant_api.py
namgyu-youn: Interesting, but may I split PR for this?

torchao/quantization/quantize_/common/quantize_tensor_kwargs.py
namgyu-youn: Could you elaborate more about it? Is it not the same as
1. W4A16-INT: `weight_tensor.act_quant_kwargs` is None
2. W8A8-INT-dynamic: `weight_tensor.act_quant_kwargs` is not None

? Let me know if I am missing something.

torchao/quantization/quant_api.py
namgyu-youn: We can divide PR for symmetric/asymmetric support if we need to, because reusing `_int8_asymm_per_token_quant` and `_int8_symm_per_token_reduced_range_quant` might make another abstraction. 

Personally, symmetric-only support is still fine because much recent research is choosing it, powered by less overhead (no zero point).

torchao/quantization/quantize_/common/quantize_tensor_kwargs.py
jerryzh168: `_choose_quant_func_and_quantize_tensor` is called in the Int8Tensor linear dispatch to quantize activations, this is trying to quantize activation to Int8, it doesn't make sense to pass around the `quant_kwargs` (which is used to quantize activations) to when you construct the Int8 activation Tensor here.

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,307 @@
...
+    optional_tensor_attribute_names = ["act_quant_kwargs", "block_size", "dtype"]
+
+    def __new__(
+        cls: type,
+        qdata: torch.Tensor,
+        scale: torch.Tensor,
+        granularity: Granularity,
+        block_size: Optional[torch.Size] = None,
```
jerryzh168: why store both granularity and block_size? I think we don't need granularity

jerryzh168: also it's better to match the order of optional args with the order in `optional_tensor_attribute_names` btw

torchao/quantization/quantize_/common/quantize_tensor_kwargs.py
namgyu-youn: Oh I totally missed that... thanks.

torchao/quantization/quant_api.py
namgyu-youn: May I split PR for this and update todo? Want to work with it, but dividing it would be easier for land I feel.

torchao/quantization/quant_api.py
jerryzh168: I think the refactor should preserve the functionality of previous version, this doesn't mean reusing these things, but adding correct args to the constructor so these activation quantization can be expressed in configs

torchao/quantization/quant_api.py
jerryzh168: the current v1 support granularity seems to be activation = PerToken and weight = PerRow, so I think a v2 migration should preserve that

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,303 @@
...
+
+    if not isinstance(weight_tensor, Int8Tensor):
+        raise TypeError(f"Expected weight to be Int8Tensor, got {type(weight_tensor)}")
+
+    output_dtype = activation_tensor.dtype
+
+    if weight_tensor.act_quant_kwargs is not None:
+        activation_tensor = Int8Tensor.from_hp(
+            activation_tensor, weight_tensor.act_quant_kwargs.granularity
+        )
```
jerryzh168, line 248: this is correct, btw I didn't mean to remove these from https://github.com/pytorch/ao/blob/2ff1eb2e356275cfbe46832327387d382c72945d/torchao/quantization/quantize_/common/quantize_tensor_kwargs.py#L33. what I meant was just not pass act_quant_kwargs again there, but anyways we can fix later

torchao/quantization/quant_api.py
jerryzh168: we can also land this and people from our team can follow up to fix the remaining issues, I think that might be faster. This is actually blocking some of our use case now, and we'd like to get int8 working ASAP. please let me know which one would you prefer

torchao/float8/inference.py
```diff
@@ -139,43 +139,50 @@ def _slice_scale_for_dimension(
...
-    # Reconstruct block sizes based on data shape and scale shape
-    block_sizes = tuple(data_shape[i] // scale.shape[i] for i in range(len(data_shape)))
+        If data_shape is [256, 128] and scale shape is [256] (indicating per-row scaling),
+        and we slice data along dim=0 from 64 to 192, the corresponding scale
+    """
+    aten = torch.ops.aten
 
-    if dim >= len(block_sizes):
-        # Slicing beyond the dimensions we care about
+    # Case 1: Per-tensor quantization (scalar scale)
```
jerryzh168: I think it's actually better to write a separate slicing function for int8 (or don't just don't use helper functions, just implement slice directly in int8 tensor), since currently the scale dimension of int8 and float8 are different. we can merge later when these are aligned.

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,303 @@
...
+        raise ValueError(f"Only dim in [0, 1, 2] supported, got dim={dim}")
+
+    if self.qdata.ndim not in [2, 3]:
+        raise ValueError(f"Expected qdata to be 2D or 3D, got {self.qdata.ndim}D")
+
+    if end is None or end > self.shape[dim]:
+        end = self.shape[dim]
+
+    sliced_qdata = aten.slice.Tensor(self.qdata, dim, start, end, step)
+    sliced_scale = _slice_scale_for_dimension(
```
jerryzh168: probably easier to just implement the slice here in the function directly

torchao/quantization/quant_api.py
namgyu-youn: > we can also land this and people from our team can follow up to fix the remaining issues, I think that might be faster. This is actually blocking some of our use case now, and we'd like to get int8 working ASAP. please let me know which one would you prefer

Please feel free to resolve with your teams. I should have divided this PR earlier (W4A16 & W8A8-dynamic) for ASAP update.

torchao/float8/inference.py
namgyu-youn: Is it possible to reuse this update function (no separate implementation) with a dimension check before calling this function?

namgyu-youn:
@jerryzh168 Added a few TODO for not addressed comments, and I will no longer work here as we discussed at https://github.com/pytorch/ao/pull/3241#discussion_r2553354809; let me know if there is anything I can help.

torchao/float8/inference.py
jerryzh168: I feel we can start with separate implementations first since there is no reuse when the dimension does not match

jerryzh168 (commented):
TODO:
1. [before landing the PR] revert changes to `float8/inference.py`, implement the slicing logic in the int8tensor itself

following can be after landing the PR since v2 is not used yet:

2. fix granularity typing to just use Granularity, follow float8 opaque tensor to handle granularity, i.e. have a normliaze_and_validate function: https://github.com/pytorch/ao/blob/3ad4d0acd1cc2dbba74a97a0f181363478655c21/torchao/prototype/float8_opaque_tensor/float8_opaque_tensor.py#L98, we can split out the normalize function in the future
3. allow different input activation quantization settings (symmetric / asymmetric) as shown in https://github.com/pytorch/ao/blob/3ad4d0acd1cc2dbba74a97a0f181363478655c21/torchao/quantization/quant_api.py#L1566-L1569 (can include the weight only decode setting as well, but optional)
4. [optional] add back the https://github.com/pytorch/ao/blob/3ad4d0acd1cc2dbba74a97a0f181363478655c21/torchao/quantization/quantize_/common/quantize_tensor_kwargs.py#L33 changes and use it to quantize activation

namgyu-youn:
> TODO:
> 
> 1. [before landing the PR] revert changes to `float8/inference.py`, implement the slicing logic in the int8tensor itself

Resolved this TODO for early land, please take a look.

jerryzh168 (approved):
looks good I think, thanks @namgyu-youn!

jerryzh168:
@namgyu-youn somehow the CI is not triggered, can you rebase again so we can trigger CI?

namgyu-youn:
> @namgyu-youn somehow the CI is not triggered, can you rebase again so we can trigger CI?

Finished rebase to trigger CI, please take a look.

torchao/quantization/quantize_/workflows/__init__.py
```diff
@@ -34,6 +38,9 @@
     "Int4MarlinSparseTensor",
     "Int4PlainInt32Tensor",
     "Int4TilePackedTo4dTensor",
+    "Float8OpaqueTensor",
```
jerryzh168: this should be removed now I think, it's moved to prototype

torchao/float8/inference.py
jerryzh168, line 1: @namgyu-youn can you revert changes to this file?

jerryzh168:
somehow I still didn't see an option to run all the CI jobs, @namgyu-youn do you mind close this one and reopen another PR to see if it helps?

namgyu-youn:
> somehow I still didn't see an option to run all the CI jobs, @namgyu-youn do you mind close this one and reopen another PR to see if it helps?

Reland at #3391, please take a look. Also, does adding a trigger (`workflow_dispatch` ) to CI make sense? https://github.com/namgyu-youn/ci_test/blob/main/.github/workflows/.test_dispatch.yml showed:

1. With dispatch (trigger button is added) 
```yaml
name: Test Workflow Dispatch

on:
  workflow_dispatch:

jobs:
  test:
    runs-on: ubuntu-latest
    steps:
      - run: echo "Workflow dispatch works!"
```

image

2. Without dispatch (TorchAO; no trigger button):

```yaml
name: Test CI for push

on:
  push:
    branches:
      - main

jobs:
  test:
    runs-on: ubuntu-latest
    steps:
      - run: echo "LGTM"
```

image

jerryzh168:
thanks, that workflow dispatch seems only like a test, doesn't look real

jerryzh168:
OK I think it's some issue with our configs, let me check

PR #3246, opened 2025-10-24
Update config.py to negate the dimension issue for FP8 support for AMD GPUs

kailashg26:
Update the config.py file to fix (“negate”) the dimension mismatch issue that arises when enabling FP8 (8-bit floating point) precision support on AMD GPUs.

**Error:**
image

### DIFF
--- a/torchao/float8/config.py
+++ b/torchao/float8/config.py
@@ -187,7 +187,7 @@ class Float8LinearConfig:
     # inner dimension of a (dim 1) and b (dim 2) with 0s. This is needed for matmuls
     # _scaled_mm since it has the strong constraint that for M,N,K  N, K must be a multiple of 16.
     # This can cause a memory spike however so we keep this off by default.
-    pad_inner_dim: bool = False
+    pad_inner_dim: bool = True
 
     # If True, emulation is used instead of hardware accelerated gemm
     emulate: bool = False


vkuzo:
could you share how this is related to AMD GPUs?  From the screenshot in the PR summary, it looks like the shapes being fed through the network do not adhere to the requirements of `scaled_mm`, which seems like a GPU independent problem.

kailashg26:
@vkuzo I agree, this is GPU independent problem. I just tried on MI300 and MI355 and this problem persists. But when I enable padding it works fine!
The issue related to this PR is https://github.com/meta-pytorch/torchtune/issues/2833#issuecomment-3008413564

vkuzo:
got it, can we just enable padding at the callsite for your use case instead of changing the default?

kailashg26:
@vkuzo not sure how do we do that at the callsite. You mean before I run my script using something like this?

`find /opt/venv/lib/python3.10/site-packages/torchao/float8/config.py -type f -print0 | xargs -0 sed -i 's/pad_inner_dim: bool = False/pad_inner_dim: bool = True/g' \
 `

I was just wondering if this might be very hacky

vkuzo:
usually the user creates a `Float8LinearConfig`, like so

```python
config = Float8LinearConfig(...)
```

in the place where that happens in torchtune, you could set the padding flag to True, or make that user configurable.  Would that work?

kailashg26:
But if we use upstream torchtune we have to submit  PR to upstream one right? Not sure if they are actively accepting PRs

vkuzo:
how about just patching this function (https://github.com/meta-pytorch/torchtune/blob/67ab86b94de9e7ac7dd9850113ebe69e2bbd307c/torchtune/training/quantization.py#L232) in your callsite instead?

```python
def patched_torchtune_convert_to_float8_training(...):
    ... modify config as needed ...

# put this somewhere after torchtune imports but before you run float8 conversion
torchtune.training.quantization.convert_to_float8_training = patched_torchtune_convert_to_float8_training
```

PR #3248, opened 2025-10-27
Support mx_tensor and enable it's test on Intel GPU

xiaowangintel:
The MXFP8/MXFP4/MXFP6 format training and inference is already enabled on CUDA, but not on Intel GPU. This pr use to support mx function of ao on Intel GPU, and enable the corresponding test. Currently, the linear for mx format is **only supported on emulate mode on Intel GPU**. We are working the scaled_mm on Intel and will enable more GEMM choice after the scaled_mm on xpu is ready.

### DIFF
--- a/test/prototype/mx_formats/test_inference_workflow.py
+++ b/test/prototype/mx_formats/test_inference_workflow.py
@@ -36,6 +36,14 @@ if not torch_version_at_least("2.8.0"):
     pytest.skip("Unsupported PyTorch version", allow_module_level=True)
 
 
+devices = []
+if torch.cuda.is_available():
+    devices.append("cuda")
+
+if torch.xpu.is_available():
+    devices.append("xpu")
+
+
 # source: https://stackoverflow.com/a/22638709
 @pytest.fixture(autouse=True)
 def run_around_tests():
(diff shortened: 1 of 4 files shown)


test/prototype/mx_formats/test_inference_workflow.py
```diff
@@ -36,6 +36,14 @@
...
 
 
+devices = []
+if torch.cuda.is_available():
+    devices.append("cuda")
+
+if torch.xpu.is_available():
+    devices.append("xpu")
+
+
```
liangan1, line 46: Suggest to use utils function of torchao to get the available device.

test/prototype/mx_formats/test_mx_tensor.py
```diff
@@ -38,6 +38,14 @@
...
 
 
+devices = []
+if torch.cuda.is_available():
+    devices.append("cuda")
+
+if torch.xpu.is_available():
+    devices.append("xpu")
+
+
```
liangan1, line 48: Suggest to use utils function of torchao to get the available device.

liangan1 (approved):
General LGTM.

test/prototype/mx_formats/test_inference_workflow.py
xiaowangintel: The get_available_devices(https://github.com/pytorch/ao/blob/main/torchao/utils.py#L139) interface can get devices. However, the output includes "cpu" which is not what we need.

test/prototype/mx_formats/test_mx_tensor.py
xiaowangintel: ditto

xiaowangintel:
@jerryzh168 Please help to review, thanks.

xiaowangintel:
This is for the older software stack. Support for the new software stack has already been submitted as a draft PR.

tye1:
> This is for the older software stack. Support for the new software stack has already been submitted as a draft PR.

PR link: https://github.com/pytorch/ao/pull/4468

PR #3249, opened 2025-10-27
[Intel XPU] Enable test/quantization UTs on XPU 

DiweiSun:
This PR is targeted to enable the CI for the test/quantization on XPU device.

### DIFF
--- a/test/quantization/test_quant_api.py
+++ b/test/quantization/test_quant_api.py
@@ -499,6 +499,7 @@ class TestQuantFlow(TestCase):
     @unittest.skipIf(not torch.accelerator.is_available(), "Need GPU available")
     def test_quantized_model_streaming(self):
         device_module = torch.get_device_module(_DEVICE)
+
         def reset_memory():
             gc.collect()
             device_module.empty_cache()
@@ -1110,6 +1111,7 @@ class TestFqnToConfig(TestCase):
     @unittest.skipIf(not torch.accelerator.is_available(), "Need GPU available")
     def test_quantized_model_streaming_fqn_config(self):
         device_module = torch.get_device_module(_DEVICE)
+
         def reset_memory():
             gc.collect()
             device_module.empty_cache()
--- a/test/quantization/test_quant_primitives.py
+++ b/test/quantization/test_quant_primitives.py
@@ -43,6 +43,7 @@ torch.manual_seed(_SEED)
 
 _DEVICE = auto_detect_device()
 
+
 # Helper function to run a function twice
 # and verify that the result is the same.
 # Adds some verification to avoid side effects.


torchao/utils.py
```diff
@@ -148,6 +148,13 @@ def get_available_devices():
     return devices
 
 
+def auto_detect_device():
```
jerryzh168, line 151: does this include cuda?

liangan1: Yes.

test/quantization/test_gptq.py
```diff
@@ -18,13 +18,15 @@
 from torchao._models.llama.tokenizer import get_tokenizer
 from torchao.quantization import Int4WeightOnlyConfig, quantize_
 from torchao.quantization.utils import compute_error
+from torchao.utils import auto_detect_device
 
 torch.manual_seed(0)
 
+_DEVICE = auto_detect_device()
```
jerryzh168, line 25: auto_detect_device seems to be changing what we want to test, I think previous we only want to test on CUDA, can you preserve this?

test/quantization/test_gptq.py
```diff
@@ -18,13 +18,15 @@
...
+
 
 class TestGPTQ(TestCase):
     @unittest.skip("skipping until we get checkpoints for gpt-fast")
-    @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
```
jerryzh168, line 27: just change this to `torch.accelerator.is_available()`?

test/quantization/test_gptq.py
```diff
@@ -104,7 +105,6 @@ def test_gptq_quantizer_int4_weight_only(self):
 
 
 class TestMultiTensorFlow(TestCase):
-    @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
```
jerryzh168, line 107: we don't want to expand test to cpu I think

test/quantization/test_gptq.py
liangan1: We have refine the auto_detect_device functions and cpu will not be included.

test/quantization/test_gptq.py
liangan1: Done

test/quantization/test_gptq.py
liangan1: Done.

test/quantization/test_quant_api.py
```diff
@@ -262,7 +265,7 @@ def api(model):
 
         m2.load_state_dict(state_dict)
         m2 = m2.to(device="cuda")
-        example_inputs = map(lambda x: x.cuda(), example_inputs)
+        example_inputs = map(lambda x: x.to(_DEVICE), example_inputs)
```
jerryzh168, line 268: so when in CPU, _DEVICE will be None now, what happens when we do x.to(None)? I think we don't want auto detect here, since in L267, it is converting model m2 to "cuda"?

test/quantization/test_quant_primitives.py
```diff
@@ -594,7 +597,7 @@ def test_choose_qparams_tensor_asym_eps(self):
     )
     def test_get_group_qparams_symmetric_memory(self):
         """Check the memory usage of the op"""
-        weight = torch.randn(1024, 1024).to(device="cuda")
+        weight = torch.randn(1024, 1024).to(device=_DEVICE)
```
jerryzh168, line 600: this test also has skip if no cuda, so it is still only going to run in cuda right? the change doesn't seem to have any effect right now

test/quantization/test_quant_api.py
jerryzh168, line 1: I think it might make sense if we change all the `skip if no cuda` to `skip if not torch.accelerator.available()` in this file

test/quantization/test_quant_api.py
zxd1997066: changed m2.to(_DEVICE), as it is skip if not torch.accelerator.available(), when in CPU, this case will be skipped

test/quantization/test_quant_api.py
zxd1997066: test_module_fqn_to_config_regex_basic, test_module_fqn_to_config_regex_fullmatch, test_module_fqn_to_config_regex_precedence and test_module_fqn_to_config_regex_precedence2 are not ready for XPU, others are changed to skip if not torch.accelerator.available()

test/quantization/test_quant_primitives.py
zxd1997066: changed skip if no cuda to skip if not torch.accelerator.available().

PR #3251, opened 2025-10-27
create a separate test for mx and nv serialization

vkuzo:
Summary:

Creates a separate test that loads MX and NV torchao checkpoints in a
subprocess, to enforce that `import torchao.prototype.mx_formats` is the
only import needed to load the checkpoint.

When MX|NV is out of prototype, we can change the requirement to `import torchao`.

Also, deletes the old serialization tests, since these are slow and we
only need to run these once per Python-object name to get full coverage.

Test Plan:

```
pytest test/prototype/mx_formats/test_mx_serialization.py -s
```

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/prototype/mx_formats/README.md
+++ b/torchao/prototype/mx_formats/README.md
@@ -73,6 +73,7 @@ Below is a toy training loop. For an example real training loop, see our torchti
 ```python
 import torch
 from torchao.quantization import quantize_
+import torchao.prototype.mx_formats
 from torchao.prototype.mx_formats import MXLinearConfig, MXGemmKernelChoice, ScaleCalculationMode
 
 # on NVIDIA Blackwell GPUs, you can use cuBLAS or CUTLASS mxfp8 kernels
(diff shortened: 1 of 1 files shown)


vkuzo:
* __->__ #3251

test/prototype/mx_formats/test_mx_serialization.py
```diff
@@ -0,0 +1,73 @@
...
+
+    assert fname is not None
+
+    code = f"""
+import torch
+import torchao.prototype.mx_formats
+_ = torch.load('{fname}', weights_only=True)
+    """
+
+    subprocess_out = subprocess.run(["python"], input=code, text=True)
```
andrewor14, line 71: if we're loading in a subprocess seems like we can just run this test in the old file too? Wondering if there's another reason to separate it out?

andrewor14 (approved):

test/prototype/mx_formats/test_mx_serialization.py
vkuzo: it's slow, the new setup runs it 2 times (once for mx, once for nv), the old one ran it once per every tested combination of various things inside the configs, ~60+ times

PR #3257, opened 2025-10-29
add a_1_128_w_128_128 (DeepSeek) float8 scaling for inference

vkuzo:
Summary:

Basic enablement of the a_1_128_w_128_128 float8 scaling recipe in
torchao inference. In detail:
1. updates the triton kernels for this scaling type to (a) be importable in an env without triton (for CI), and (b) adds compile support for the gemm
2. enables the new granularity in various utility functions
3. wires the new granularity through the float8 inference configs
4. adds a test which tests for e2e numerical correctness via SQNR
   comparison vs high precision baseline

For now we only have fallback kernels which requires triton and are numerically
correct but may not reach optimal performance. Performance optimization is
left for future PRs:
1. we should map the gemm to `torch._scaled_mm` for CUDA 12.9+
2. we should enable an fbgemm_gpu_genai path, if available in user env
3. we should map to a triton kernel for quantizing the weights, as
   `torch.compile` is currently known slow for 128x128 block
   quantization

Further accuracy testing and enablement of more features is left for future PRs, to keep PR size small.

Test Plan:

```
pytest test/quantization/quantize_/workflows/float8/test_float8_tensor.py -s -x
pytest test/dtypes/test_affine_quantized_float.py -s -x
```

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/quantization/granularity.py
+++ b/torchao/quantization/granularity.py
@@ -112,8 +112,8 @@ class PerBlock(Granularity):
     * block_size has shape [X, Y]
     * input_tensor shape [A] -> scaling undefined
     * input_tensor shape [A, B] -> scale shape [A // X, B // Y]
-    * input_tensor shape [A, B, C] -> scale shape [1, B // X, C // Y]
-    * input_tensor shape [A, B, C, D] -> scale shape [1, 1, C // X, D // Y], and so on
+    * input_tensor shape [A, B, C] -> scale shape [A, B // X, C // Y]
+    * input_tensor shape [A, B, C, D] -> scale shape [A, B, C // X, D // Y], and so on
 
     Note that `PerBlock((1, Y))` is equivalent to `PerGroup(Y)`
 


vkuzo:
* #3269
* #3265
* #3259
* __->__ #3257

test/quantization/quantize_/workflows/float8/test_float8_tensor.py
```diff
@@ -86,13 +90,21 @@ def test_fp8_linear_variants(
...
-            return unittest.skip(
-                "per tensor with fbgemm kernel preferece does not work yet"
-            )
+        if isinstance(granularity, PerTensor):
+            if kernel_preference is KernelPreference.FBGEMM:
+                return unittest.skip(
+                    "per tensor with fbgemm kernel preference does not work yet"
+                )
+            elif mode == "weight-only":
+                return unittest.skip("unimplemented")
```
vkuzo, line 99: unrelated to goal of this PR, but fixing while I'm here

torchao/kernel/blockwise_quantization.py
```diff
@@ -8,272 +8,301 @@
...
-    assert x.is_contiguous() and s.is_contiguous(), "Input tensors must be contiguous"
-    assert x.dim() == 2 and s.dim() == 2, "Input tensors must have 2 dimensions"
-    M, N = x.size()
-    y = torch.empty_like(x, dtype=torch.get_default_dtype())
-    grid = lambda meta: (
-        triton.cdiv(M, meta["BLOCK_SIZE"]),
-        triton.cdiv(N, meta["BLOCK_SIZE"]),
+from torch.utils._triton import has_triton
+
+if has_triton():
```
vkuzo, line 13: most of the changes in this file is just indent from adding the `if has_triton()` statement

torchao/kernel/blockwise_quantization.py
```diff
@@ -8,272 +8,301 @@
...
+            b_s_ptrs += 1
+
+        c = accumulator.to(c_ptr.dtype.element_ty)
+        offs_m = pid_m * BLOCK_SIZE_M + tl.arange(0, BLOCK_SIZE_M)
+        offs_n = pid_n * BLOCK_SIZE_N + tl.arange(0, BLOCK_SIZE_N)
+        c_ptrs = c_ptr + offs_m[:, None] * N + offs_n[None, :]
+        mask = (offs_m[:, None] < M) & (offs_n[None, :] < N)
+        tl.store(c_ptrs, c, mask=mask)
+
+    @torch.library.custom_op("ao::blockwise_fp8_gemm", mutates_args=())
```
vkuzo, line 79: non-indent change 1

torchao/kernel/blockwise_quantization.py
```diff
@@ -8,272 +8,301 @@
...
+        grid = lambda META: (
+            triton.cdiv(M, META["BLOCK_SIZE_M"]),
+            triton.cdiv(N, META["BLOCK_SIZE_N"]),
+        )
+        blockwise_fp8_gemm_kernel[grid](
+            a, b, c, a_s, b_s, M, N, K, M_BUCKET, BLOCK_SIZE_K=block_size
+        )
+        return c
+
+    @blockwise_fp8_gemm.register_fake
```
vkuzo, line 105: non-indent change 2

torchao/kernel/blockwise_quantization.py
```diff
@@ -8,272 +8,301 @@
...
+        M, N = x.size()
+        y = torch.empty_like(x, dtype=torch.get_default_dtype())
+        grid = lambda meta: (
+            triton.cdiv(M, meta["BLOCK_SIZE"]),
+            triton.cdiv(N, meta["BLOCK_SIZE"]),
+        )
+        fp8_blockwise_weight_dequant_kernel[grid](x, s, y, M, N, BLOCK_SIZE=block_size)
+        return y
+
+else:
```
vkuzo, line 289: non-indent change 3

test/quantization/quantize_/workflows/float8/test_float8_tensor.py
```diff
@@ -86,13 +90,21 @@ def test_fp8_linear_variants(
...
+
+        elif granularity == (PerBlock((1, 128)), PerBlock((128, 128))):
+            if dtype is torch.float32:
```
jerryzh168: nit: it's not very clear what is supported, will it be clearer to reverse this and say:
```
if not SUPPORTED_CONDITION:
   return unittest.skip("unimplemented")
```

torchao/float8/inference.py
```diff
@@ -196,6 +198,36 @@ def _is_tensorwise_scaled(x: torch.Tensor) -> bool:
...
 
 
+def _is_1_128_scaled(x: torch.Tensor) -> bool:
+    """Checks if a quantized tensor is scaled with a block size of 1x128
+    Args:
+        x: quantized tensor (should have `block_size` attribute)
+    """
+    assert hasattr(x, "block_size"), "Expecting input to have `block_size` attribute"
+    b = x.block_size
+    return len(b) >= 2 and math.prod(b[:-1]) == 1 and b[-1] == 128
```
jerryzh168, line 208: is this per block if `len(b) >= 2`? seems like PerGroup? https://github.com/pytorch/ao/blob/1e473ed94caa060b3fbac96657030c44173759e0/torchao/quantization/utils.py#L723

vkuzo: as of today:
1. x.shape can have any rank >= 2
2. block_size is statically assigned during config creation, and is not aware of the rank of x

for example, is x.shape is (1, 4, 8) and block_size is (2, 4), we assume that the leading dim is implied and the scale should be of dim (1, 2, 2)

the code ^ translates between (1) and (2)

jerryzh168: but block_size has to match the dimension of the input tensor so far I think, we can create another type of granularity for the example you described?

but as is, seems like (1, 128) + the semantics above means it's the same as PerGroup(128) is that correct?

vkuzo: I see, I think you are right, this is the same as PerGroup.  Will change it to PerGroup.

vkuzo: we'll have to revisit this for 3d parameters though, because block_size will be 128x128 and it will need to support 2d and 3d weights

torchao/quantization/utils.py
```diff
@@ -706,6 +706,15 @@ def get_block_size(
         return tuple(block_size)
     elif isinstance(granularity, PerBlock):
         block_size = granularity.block_size
+
+        # pad the start of `block_size` with 1s, to make 2d block_size
+        # handle tensors of rank 3+
+        if len(block_size) < len(input_shape):
```
jerryzh168, line 712: should we create a new type, instead of making `block_size` taking more meanings?

vkuzo: hmm, is there any use case for blockwise without allowing leading dims?  I.e. if I say "128x128 blockwise scaling", I'm not sure if there is value in supporting that for only 2d tensors vs supporting that for tensors of any rank>=2.

jerryzh168: I feel the semantics to combining preceding dims and then do a > 1 grouping is a bit weird/complicated for 3D and more actually so was thinking of keeping the PerBlock simple and create a new one if needed

for 128x128, I'm wondering for 3D, is this just (1, 128, 128)? or is there actually need to combine elements from first 2 dims and do a 128 grouping? (e.g. I feel combining first 2 dims of (32, 64, 128) and then do (128, 128) makes it harder to reason about which elements belongs to the same element for dim0)

jerryzh168: discussed offline to repurpose PerBlock to the semantics of adding leading (1,)s like PerGroup, to keep user API simpler

jerryzh168 (approved):
looks good, I think we can also keep the 1x128 block to align with what deepseek is calling it now

vkuzo:
CI failures exist on main branch, landing

PR #3258, opened 2025-10-29
Introduce new W8A8-FP-CSR quantitzation API

namgyu-youn:
**Summary**:
Introduce new W8A8-FP-CSR quantization API, `Float8SemiSparseTensor`, which specializes in semi-sparse pattern using cuSPARSELt accelerations (https://docs.nvidia.com/cuda/cusparselt/)

**Related Issue/PR**: #2752 

**Future Plan**:
This PR only introduces core operations (quantization/dequantization). For better API support, we have to introduce tensor utility operations like indexing and slicing.

**Test Plan**:
test/prototype/quantization/quantize_/float8/test_float8_semisparse_tensor.py

### DIFF
--- a/torchao/prototype/quantization/quantize_/workflows/float8/float8_semisparse_tensor.py
+++ b/torchao/prototype/quantization/quantize_/workflows/float8/float8_semisparse_tensor.py
@@ -142,7 +142,6 @@ class Float8SemiSparseTensor(TorchAOBaseTensor):
 
         # Store fp8 data in both dense and compressed formats
         fp8_data_fp16 = fp8_data.to(torch.float16)
-        from torch.sparse import to_sparse_semi_structured
 
         fp8_compressed = to_sparse_semi_structured(fp8_data_fp16)
 
@@ -180,47 +179,28 @@ def _(func, types, args, kwargs):
     )
 
     assert isinstance(weight_tensor, Float8SemiSparseTensor)
-    assert activation_tensor.shape[-1] == weight_tensor.original_shape[1], (
-        f"Shape mismatch: {activation_tensor.shape} @ {weight_tensor.original_shape}"
-    )
+    assert activation_tensor.dim() == 2, "Only 2D input supported"
+    assert activation_tensor.shape[-1] == weight_tensor.original_shape[1]
 
-    # Flatten batch dimensions for scale computation
-    orig_shape = activation_tensor.shape
-    if activation_tensor.dim() > 2:
-        activation_flat = activation_tensor.view(-1, orig_shape[-1])
-    else:
-        activation_flat = activation_tensor
+    x_scales = _choose_qparams_affine_floatx(activation_tensor, ebits=4, mbits=3)
+    w_scales = weight_tensor.scale
 
-    # Compute dynamic scale for activation quantization
-    x_scales = _choose_qparams_affine_floatx(activation_flat, ebits=4, mbits=3)
-    x_scales = x_scales.unsqueeze(1)  # [batch, 1]
-
-    # Quantize activation
-    scaled_x = activation_flat / x_scales
-    scaled_x = scaled_x.clamp(-448.0, 448.0)
+    # Global normalizer to prevent overflow
+    global_scale = (x_scales.max() * w_scales.max()).sqrt().clamp(min=0.01)
+    x_scales_adj = (x_scales.unsqueeze(1) / global_scale).to(torch.float32)
+    scaled_x = (activation_tensor.to(torch.float32) / x_scales_adj).clamp(-448.0, 448.0)
     x_vals_fp8 = scaled_x.to(torch.float8_e4m3fn)
 
-    # Dequantize both activation and weight before MatMul to avoid FP16 overflow
-    x_dequant = (x_vals_fp8.to(torch.float32) * x_scales.to(torch.float32)).to(
-        torch.float16
+    # MatMul
+    x_padded = SparseSemiStructuredTensorCUSPARSELT._pad_dense_input(
+        x_vals_fp8.to(torch.float16)
     )
-    w_dequant = (
-        weight_tensor.qdata.to(torch.float32)
-        * weight_tensor.scale.unsqueeze(1).to(torch.float32)
-    ).to(torch.float16)
+    y_fp16 = torch.matmul(x_padded, weight_tensor.qdata_compressed.t())
+    y = y_fp16[: activation_tensor.shape[0], :].to(torch.float32)
 
-    # Sparse MatMul with dequntized tensor
-    w_sparse = to_sparse_semi_structured(w_dequant)
-    row = x_dequant.shape[0]
-    x_padded = SparseSemiStructuredTensorCUSPARSELT._pad_dense_input(x_dequant)
-
-    y = torch.matmul(x_padded, w_sparse.t())
-    y = y[:row, :]
-
-    # Reshape to original activation shape
-    if activation_tensor.dim() > 2:
-        y = y.view(*orig_shape[:-1], -1)
-
+    # Restore scale
+    w_scales_fp32 = w_scales.to(torch.float32)
+    y = y * (x_scales_adj * w_scales_fp32.unsqueeze(0) * global_scale)
     y = y.to(activation_tensor.dtype).contiguous()
 
     if bias is not None:


namgyu-youn:
@jcaip could you please check this PR?

jerryzh168:
@namgyu-youn thanks, seems like int8 is WIP in https://github.com/pytorch/ao/pull/2971 although not sure about the status there

jcaip:
cc @namgyu-youn

Can you split this into two PRs? one for int8 and one for float8?

In general I don't think we want to introduce weight-only sparsity configs for int8 and float8 because we don't have mixed-dtype kernel support currently. The only kernels we have are for int8 x int8 2:4 sparse and fp8 x fp8 2:4 sparse.

I would like Int8SemiSparseTensor though, but I think it should live in prototype until we have a user for it.  

Also cc @bbeckca who has been working on fp8xfp8 2:4 sparse tensor subclass migration in #3182.

jerryzh168:
> cc @namgyu-youn
> 
> Can you split this into two PRs? one for int8 and one for float8?
> 
> In general I don't think we want to introduce weight-only sparsity configs for int8 and float8 because we don't have mixed-dtype kernel support currently. The only kernels we have are for int8 x int8 2:4 sparse and fp8 x fp8 2:4 sparse.
> 
> I would like Int8SemiSparseTensor though, but I think it should live in prototype until we have a user for it.
> 
> Also cc @bbeckca who has been working on fp8xfp8 2:4 sparse tensor subclass migration in #3182.

@jcaip if we want to move int8 2:4 sparse to prototype, then we don't need to migrate the tensor I think

namgyu-youn:
Okay, then I'll address only ~W8A8-INT~ W8A8-FP here and keep file structure at the prototype.

jcaip:
cc @namgyu-youn I talked to @bbeckca and I think your PR is closer so lets use it instead. 
Can you remove the int8 changes then and I will give this a review. Thanks for picking this up!

namgyu-youn:
cc @jcaip to request review, thanks.

torchao/prototype/quantization/quantize_/workflows/float8/float8_semisparse_tensor.py
```diff
@@ -0,0 +1,232 @@
...
+from torchao.utils import TorchAOBaseTensor
+
+__all__ = ["Float8SemiSparseTensor"]
+
+aten = torch.ops.aten
+
+
+class Float8SemiSparseTensor(TorchAOBaseTensor):
+    """
+    W8A8-FP-CSR: float8 quantized tensor with 2:4 semi-structured sparsity layout
```
jcaip, line 24: nit: comment looks wrong, CSR is compressed sparse row and it's not the sparse format used here (2:4 sparsity)

torchao/prototype/quantization/quantize_/workflows/float8/float8_semisparse_tensor.py
```diff
@@ -0,0 +1,232 @@
...
+        qdata_compressed: SparseSemiStructuredTensor (compressed format for matmul)
+        scale: scale factors for dequantization
+
+    Non-Tensor Attributes:
+        block_size: block size for quantization granularity
+        original_shape: original uncompressed shape
+        float8_dtype: float8 dtype variant
+    """
+
+    tensor_data_names = ["qdata", "qdata_compressed", "scale"]
```
jcaip, line 37: I think `quantized_sparse_data` and `quantized_sparse_metadata` would be better here for variable names. 

`quantized_sparse_data` holds the specified values and `quantized_sparse_metadata` holds the sparsity metadata.

torchao/prototype/quantization/quantize_/workflows/float8/float8_semisparse_tensor.py
```diff
@@ -0,0 +1,232 @@
...
+
+    def __repr__(self):
+        return (
+            f"{self.__class__.__name__}("
+            f"qdata_shape={self.qdata.shape}, {self.scale=}, "
+            f"{self.block_size=}, {self.shape=}, {self.device=}, {self.dtype=})"
+        )
+
+    @property
+    def qdata_fp8(self):
```
jcaip, line 84: why do we need this?

torchao/prototype/quantization/quantize_/workflows/float8/float8_semisparse_tensor.py
```diff
@@ -0,0 +1,232 @@
...
+
+        # Apply 2:4 sparsity pruning
+        with torch.no_grad():
+            w_sparse = w.clone()
+
+        pruning_inds = w_sparse.abs().view(-1, 4).argsort(dim=1)[:, :2]
+        w_sparse.view(-1, 4).scatter_(1, pruning_inds, value=0)
+
+        # Check for all-zero (sparsity=1) tensor
+        if w_sparse.abs().max() == 0:
```
jcaip, line 125: I think this should be supported actually? I don't see why we should error here.

torchao/prototype/quantization/quantize_/workflows/float8/float8_semisparse_tensor.py
```diff
@@ -0,0 +1,232 @@
...
+        if rows % block_size[0] != 0 or cols % block_size[1] != 0:
+            raise ValueError(
+                f"Dimensions {w.shape} must be divisible by block_size {block_size}"
+            )
+
+        # Apply 2:4 sparsity pruning
+        with torch.no_grad():
+            w_sparse = w.clone()
+
+        pruning_inds = w_sparse.abs().view(-1, 4).argsort(dim=1)[:, :2]
```
jcaip, line 121: you can use this util: https://github.com/pytorch/ao/blob/315e9b4555f2245bb452b9e22f485419188ce8b4/torchao/sparsity/utils.py#L101 here

torchao/prototype/quantization/quantize_/workflows/float8/float8_semisparse_tensor.py
```diff
@@ -0,0 +1,232 @@
...
+        scaled_data = scaled_data.clamp(-448.0, 448.0)
+        fp8_data = scaled_data.to(float8_dtype).contiguous()
+
+        if not torch.isfinite(fp8_data.to(torch.float32)).all():
+            raise ValueError("fp8_data contains NaN/Inf after quantization")
+
+        # Store fp8 data in both dense and compressed formats
+        fp8_data_fp16 = fp8_data.to(torch.float16)
+
+        fp8_compressed = to_sparse_semi_structured(fp8_data_fp16)
```
jcaip, line 146: We should use the torchao cutlass packing kernels here, not the default torch ones: https://github.com/pytorch/ao/blob/315e9b4555f2245bb452b9e22f485419188ce8b4/torchao/dtypes/floatx/cutlass_semi_sparse_layout.py#L171

torchao/prototype/quantization/quantize_/workflows/float8/float8_semisparse_tensor.py
```diff
@@ -0,0 +1,232 @@
...
+            raise ValueError("Input tensor is all zeros after pruning")
+
+        scale = _choose_qparams_affine_floatx(w_sparse, ebits=4, mbits=3)
+
+        if not torch.isfinite(scale).all():
+            raise ValueError("Scale contains NaN or inf values")
+        if not (scale > 0).all():
+            raise ValueError(f"Scale contains non-positive values: min={scale.min()}")
+
+        scale_expanded = scale.unsqueeze(1)
```
jcaip, line 135: Is this different from Float8Tensor, can we use the same scale calculation logic as we use there?

torchao/prototype/quantization/quantize_/workflows/float8/float8_semisparse_tensor.py
```diff
@@ -0,0 +1,232 @@
...
+        if not torch.isfinite(fp8_data.to(torch.float32)).all():
+            raise ValueError("fp8_data contains NaN/Inf after quantization")
+
+        # Store fp8 data in both dense and compressed formats
+        fp8_data_fp16 = fp8_data.to(torch.float16)
+
+        fp8_compressed = to_sparse_semi_structured(fp8_data_fp16)
+
+        return cls(
+            fp8_data,  # dense for dequantization
```
jcaip, line 149: we shouldn't be storing both the dense data and the compressed data, we should be storing the sparse specified values and the sparse metadata.

torchao/prototype/quantization/quantize_/workflows/float8/float8_semisparse_tensor.py
```diff
@@ -0,0 +1,232 @@
...
+            fp8_compressed,  # compressed for matmul
+            scale,
+            block_size,
+            original_shape=w.shape,
+            dtype=w.dtype,
+            float8_dtype=float8_dtype,
+        )
+
+    def dequantize(self, output_dtype: Optional[torch.dtype] = None) -> torch.Tensor:
```
jcaip, line 158: we should multiply by identity matrix to dequantize, like we do here: https://github.com/pytorch/ao/blob/315e9b4555f2245bb452b9e22f485419188ce8b4/torchao/dtypes/floatx/cutlass_semi_sparse_layout.py#L133

torchao/prototype/quantization/quantize_/workflows/float8/float8_semisparse_tensor.py
```diff
@@ -0,0 +1,232 @@
...
+    w_scales = weight_tensor.scale
+
+    # Global normalizer to prevent overflow
+    global_scale = (x_scales.max() * w_scales.max()).sqrt().clamp(min=0.01)
+    x_scales_adj = (x_scales.unsqueeze(1) / global_scale).to(torch.float32)
+    scaled_x = (activation_tensor.to(torch.float32) / x_scales_adj).clamp(-448.0, 448.0)
+    x_vals_fp8 = scaled_x.to(torch.float8_e4m3fn)
+
+    # MatMul
+    x_padded = SparseSemiStructuredTensorCUSPARSELT._pad_dense_input(
```
jcaip, line 195: We should use the torchao cutlass fp8 kernels, which fuse in scale multiplication here.

See https://github.com/pytorch/ao/blob/315e9b4555f2245bb452b9e22f485419188ce8b4/torchao/dtypes/floatx/cutlass_semi_sparse_layout.py#L252

torchao/prototype/quantization/quantize_/workflows/__init__.py
jerryzh168, line 1: delete? we don't want this to be in prototype I think

jerryzh168: this should be add to the init file without the prototype in path

test/prototype/quantization/quantize_/float8/test_float8_semisparse_tensor.py
jerryzh168, line 1: also need to add to `Float8DynamicActivationFloat8WeightConfig`?

namgyu-youn:
> cc @namgyu-youn 
> 
> I think there's a bit of confusion on what the tensor subclass should be storing and how to do the op overload.
> 
> Please take a look at https://github.com/pytorch/ao/pull/3182/files#diff-afc7dd21d2b704181a6fd55be989426c0217a2bbfb694af9eb9746239ec462ed for the appropriate logic / ops to be called. 
> 
> 

@jcaip Thanks a lot for the comprehensive review. I didn't know there was an already opened PR (#3182), and I found my implementation is quite far away (mostly ops, kernel). Therefore, the right move seems to be reopening #3182 and letting me update it after the last review. Is it okay to go with this?
Let me know which move is right for progress. Also, cc @bbeca who already did this work.

jcaip:
@namgyu-youn I think it'll be easier for me to just migrate this over, mind if I take over the PR? #3182 is also quite far from landing.

PR #3261, opened 2025-10-30
[PT2E][X86] Add Inductor fusion passes of float8 qconv for X86Inductor backend

jiayisunx:
Summary:
This PR aims to support QConv weight prepacking, QConv unary fusion, and QConv binary fusion for fp8 data type in Inductor, to lower the ref quant model to X86Inductor backend.

### DIFF
--- a/torchao/quantization/pt2e/inductor_passes/x86.py
+++ b/torchao/quantization/pt2e/inductor_passes/x86.py
@@ -773,7 +773,10 @@ def _register_qconv_weight_prepack_pass(pattern, pass_number, dtype=torch.float3
         if has_free_symbols(x_shape):
             # For dynamic shape case, we can't get activation shape ahead of runtime.
             x_shape = None
-        if is_fp8 and w_scale.target is torch.ops.aten.full.default:
+        if is_fp8:
+            # For float8, we assume the scales are from aten.full.default instead of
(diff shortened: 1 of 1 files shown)


torchao/quantization/pt2e/inductor_passes/x86.py
```diff
@@ -792,14 +773,19 @@ def qconv_weight_prepack(match: Match, *args, **kwargs):
         if has_free_symbols(x_shape):
             # For dynamic shape case, we can't get activation shape ahead of runtime.
             x_shape = None
+        if is_fp8 and w_scale.target is torch.ops.aten.full.default:
```
Xia-Weiwen: Can you explain the special case here? And maybe a comment here will be helpful.

Xia-Weiwen (approved):

torchao/quantization/pt2e/inductor_passes/x86.py
jiayisunx: Added, thanks for your comment!

Xia-Weiwen:
> stamping since this doesn't touch the main torchao API or user API
> 
> can you show any performance metrics before and after the fusion as well

Thanks for reviewing. This fusion passes are actually for lowering, i.e., to fuse the pattern `dq - conv q` to `qconv`. So performance comparisons might not be meaningful.

PR #3265, opened 2025-10-30
Makes fallback float8 1x128 by 128x128 gemm output bfloat16

vkuzo:
Summary:

For now, we just care about bf16 output.  We can add fp32 and a flag to
control it later, if needed.

Test Plan:

```
pytest test/quantization/quantize_/workflows/float8/test_float8_tensor.py -s -k fp8_linear_variants -x
```

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo:
* #3279
* #3269
* __->__ #3265

vkuzo:
> LGTM. Btw in my benchmarks I found the torch._scaled_mm cutlass kernel for blockwise gemms to be much faster than the triton kernels. This was a few months ago, you can run the benchmarks scripts in this dir if you want: https://github.com/pytorch/ao/tree/main/benchmarks/prototype/blockwise_fp8_training

Yes, two things for ^:
1. we need https://github.com/pytorch/pytorch/pull/166752 to land in core for the scaling to be numerically correct before we enable it in torchao
2. we need a fallback path that works on various GPUs, torch.scaled_mm blockwise scaling only works on Hopper

PR #3266, opened 2025-10-30
DTensor support for bfloat16 stochastic rounding

EduardDurech:
Resolves https://github.com/pytorch/ao/issues/2296

Upstream torch is not taking this seriously https://github.com/pytorch/pytorch/issues/156649

### DIFF
--- a/torchao/optim/quant_utils.py
+++ b/torchao/optim/quant_utils.py
@@ -148,7 +148,15 @@ def _fp32_to_bf16_sr(_x_f32: Tensor) -> Tensor:
     # x_f32_bits = (x_f32_bits + rand_16bit) & 0xFFFF0000
     x_bf16_trunc = x_f32_bits.view(torch.float32).bfloat16()
 
-    return DTensor.from_local(
-        x_bf16_trunc, _x_f32.device_mesh, _x_f32.placements,
-        run_check=False, shape=tuple(_x_f32.shape), stride=tuple(_x_f32.stride()),
-    ) if is_dt else x_bf16_trunc
+    return (
+        DTensor.from_local(
+            x_bf16_trunc,
+            _x_f32.device_mesh,
+            _x_f32.placements,
+            run_check=False,
+            shape=tuple(_x_f32.shape),
+            stride=tuple(_x_f32.stride()),
+        )
+        if is_dt
+        else x_bf16_trunc
+    )
--- a/test/test_low_bit_optim.py
+++ b/test/test_low_bit_optim.py
@@ -124,8 +124,8 @@ class TestQuantize(TestCase):
     def test_bf16_stochastic_round_dtensor(self, device, compile):
         pytest.importorskip("torch.distributed")
         import torch.distributed as dist
-        from torch.distributed.tensor import DTensor, Replicate
         from torch.distributed.device_mesh import init_device_mesh
+        from torch.distributed.tensor import DTensor, Replicate
 
         created_pg = False
         if dist.is_available() and not dist.is_initialized():


torchao/optim/quant_utils.py
```diff
@@ -5,6 +5,13 @@
 # LICENSE file in the root directory of this source tree.
 import torch
 from torch import Tensor
+try:
```
vkuzo: is this for different PyTorch versions? If yes, could you clarify which specific versions?

in general we support 3 most recent stable PyTorch releases max, so if it's older than that I'd just leave it out

torchao/optim/quant_utils.py
```diff
@@ -5,6 +5,13 @@
...
+    from torch.distributed.tensor import DTensor
+except Exception:
+    try:
+        from torch.distributed._tensor import DTensor
+    except Exception:
+        DTensor = tuple()
```
vkuzo: rethrow this instead returning tuple

torchao/optim/quant_utils.py
```diff
@@ -142,4 +152,9 @@ def _fp32_to_bf16_sr(x_f32: Tensor) -> Tensor:
     )
     # alternative, slightly faster
     # x_f32_bits = (x_f32_bits + rand_16bit) & 0xFFFF0000
-    return x_f32_bits.view(torch.float32).bfloat16()
+    x_bf16_trunc = x_f32_bits.view(torch.float32).bfloat16()
+
+    return DTensor.from_local(
```
vkuzo: looks reasonable, can we add a test to cover?

torchao/optim/quant_utils.py
EduardDurech: @vkuzo considering the last commit https://github.com/pytorch/pytorch/commit/995df34b19387f672fc49ba4197e56c11ad52f7f seems it was 2.7.0 so I'll remove the backwards compatibility

EduardDurech:
@vkuzo done

```shell
test/test_low_bit_optim.py::TestQuantize::test_bf16_stochastic_round_dtensor_device_cpu_compile_False PASSED
test/test_low_bit_optim.py::TestQuantize::test_bf16_stochastic_round_dtensor_device_cpu_compile_True PASSED
test/test_low_bit_optim.py::TestQuantize::test_bf16_stochastic_round_dtensor_device_cuda_compile_False PASSED
test/test_low_bit_optim.py::TestQuantize::test_bf16_stochastic_round_dtensor_device_cuda_compile_True PASSED
```

EduardDurech:
@vkuzo could you check the updates, thanks

gau-nernst:
Are there any performance hit from this?

EduardDurech:
> Are there any performance hit from this?

Compared to what? It doesn't work otherwise, you either do FSDP or no DTensors or this with FSDP2 or DTensors, else it doesn't work, it falls back to old logic if not using DTensors so it doesn't affect the old path other than the overhead of checking if it's a DTensor; the only baseline is to compare no DTensor with it in a compiled graph and DTensor with it in a compiled graph, and it's the expected minimal speedup of FSDP2 to FSDP, the other thing would be to fix this in Torch but they aren't taking it seriously https://github.com/pytorch/pytorch/issues/156649

gau-nernst:
@EduardDurech im asking if there is a perf hit for the single gpu case. But considering that the optimizer is already using explicit fullgraph=True, maybe it's fine. Though I dont know if the extra logic changes how torch.compile partition the graph into individual kernels, if any. Anyway, it would be a good sanity check to ensure there is no speed regression...

EduardDurech:
@gau-nernst I get what you mean, I can't test myself but the cases
1. Single GPU, no DTensor, no compile graph - one extra isinstance and assignment
2. Single GPU, no DTensor, compile graph - isinstance guard, so should just be the same as before
3. Single GPU, DTensor - to_local() and from_local(), but no other option, if it makes a noticeable difference though don't have much time to test

EduardDurech:
P.S. PyTorch has an open PR from a week ago, when this is merged the logic can change https://github.com/pytorch/pytorch/pull/169340

PR #3279, opened 2025-11-03
make float8 a1x128_w128x128 granularity serializeable

vkuzo:
Summary:

Change the granularity for this scaling type from
`PerBlock((1, 128)), PerBlock((128, 128))` to `PerBlock([1, 128]), PerBlock([128, 128])`,
to get around the current limitation of config serialization not
supporting tuples.

Error when serializing tuples:
https://gist.github.com/vkuzo/ab4d6aec83cb98ad9417898d2c024a2c

Supporting tuples and lists would be a better long term fix, but the
workaround will unblock further work in the short term.

Test Plan:

```
pytest test/quantization/quantize_/workflows/float8/test_float8_tensor.py -s -x
```

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo:
* __->__ #3279

torchao/quantization/quant_api.py
```diff
@@ -1782,7 +1782,6 @@ def __post_init__(self):
                 KernelPreference.AUTO,
                 KernelPreference.TORCH,
             ), "unimplemented"
-            assert self.mm_config is None, "unimplemented"
```
jerryzh168, line 1785: is this supported now?

vkuzo: yes, this is tested with the serialization flow.  When the object is created a default value of `mm_config` is set, and then when object is saved and reloaded, that value then makes it to the constructor.

PR #3283, opened 2025-11-03
fix unit test to use no grad

liangel-02:
**Summary** 
Updating tests to use no grad due to [166367](https://github.com/pytorch/pytorch/pull/166367) breaking CI. PR was since reverted but updating in anticipation of it relanding. 

**Test**
`python test/quantization/quantize_/workflows/float8/test_float8_tensor.py`

### DIFF
--- a/test/quantization/quantize_/workflows/float8/test_float8_tensor.py
+++ b/test/quantization/quantize_/workflows/float8/test_float8_tensor.py
@@ -80,6 +80,7 @@ class ToyConvModel(torch.nn.Module):
 class TestFloat8Tensor(TorchAOIntegrationTestCase):
     def setUp(self):
         self.GPU_DEVICES = ["cuda"] if torch.cuda.is_available() else []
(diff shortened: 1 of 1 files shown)


test/quantization/quantize_/workflows/float8/test_float8_tensor.py
```diff
@@ -37,6 +37,7 @@
 
 # Needed since changing args to function causes recompiles
 torch._dynamo.config.cache_size_limit = 128
+torch.set_grad_enabled(False)
```
andrewor14: I think this disables it for all the tests, that's why some unrelated optim tests are failing? Maybe we should scope it to within the specific float8 test that's failing? @liangel-02 @jerryzh168 what do you think

jerryzh168: yeah I think we can put this in setUp of the test case

PR #3294, opened 2025-11-05
Add deprecation warnings for various inference configs

vkuzo:
Summary:

Adds warnings that the following configs will be moved to prototype in a
future release:
* `Int8DynamicActivationInt4WeightConfig`
* `Int4DynamicActivationInt4WeightConfig`
* `GemliteUIntXWeightOnlyConfig`
* `Float8StaticActivationFloat8WeightConfig`
* `UIntXWeightOnlyConfig`
* `FPXWeightOnlyConfig`

See https://github.com/pytorch/ao/issues/2752 for more context

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/quantization/test_quant_api.py
+++ b/test/quantization/test_quant_api.py
@@ -835,7 +835,9 @@ class TestQuantFlow(TestCase):
                 # Each call should have at least one warning.
                 # Some of them can have two warnings - one for deprecation,
                 # one for moving to prototype
-                self.assertTrue(len(_warnings) > 0)
+                # 1 warning - just deprecation
+                # 2 warnings - deprecation and prototype warnings
+                self.assertTrue(len(_warnings) in (1, 2))
                 found_deprecated = False
                 for w in _warnings:
                     if "is deprecated and will be removed in a future release" in str(


vkuzo:
* __->__ #3294

torchao/quantization/quant_api.py
```diff
@@ -592,6 +592,9 @@ def __post_init__(self):
         torch._C._log_api_usage_once(
             "torchao.quantization.Int8DynamicActivationInt4WeightConfig"
         )
+        warnings.warn(
+            "`Int8DynamicActivationInt4WeightConfig` will be moving to prototype in a future release of torchao. Please see https://github.com/pytorch/ao/issues/2752 for more details."
```
andrewor14, line 596: I feel this config can just be removed in the future in favor of `Int8DynamicActivationIntXWeightConfig`. Also maybe add the replacement in the warning?

test/quantization/test_quant_api.py
```diff
@@ -826,19 +826,23 @@ def test_config_deprecation(self):
...
-            self.assertEqual(len(_warnings), len(deprecated_apis_to_args))
-            for w in _warnings:
-                self.assertIn(
-                    "is deprecated and will be removed in a future release",
-                    str(w.message),
-                )
+                # Each call should have at least one warning.
+                # Some of them can have two warnings - one for deprecation,
+                # one for moving to prototype
+                self.assertTrue(len(_warnings) > 0)
```
andrewor14: I think we want to assert `len(_warnings) == 1` here, since we moved the warnings context manager outside the loop? E.g. if it's 2 then that means we've logged a warning each time we call the API, which could be very noisy

torchao/quantization/quant_api.py
```diff
@@ -963,6 +966,9 @@ def __post_init__(self):
         torch._C._log_api_usage_once(
             "torchao.quantization.Int4DynamicActivationInt4WeightConfig"
         )
+        warnings.warn(
+            "`Int4DynamicActivationInt4WeightConfig` will be moving to prototype in a future release of torchao. Please see https://github.com/pytorch/ao/issues/2752 for more details."
```
andrewor14, line 970: Do we know if anyone's actually using these configs? Is it moving to prototype just in case someone is still using them?

torchao/quantization/quant_api.py
vkuzo: I agree long term, I want to prioritize moving these out of main folder more urgently to clean up the main folder, and deleting outright is less urgent.  If someone wants to own deleting outright on a tight timeline, sounds good to me!

torchao/quantization/quant_api.py
vkuzo: Moving to prototype is to clean up main folder faster, and we can delete from prototype at a later time.  Note that there are some internal use cases using some of these configs.

test/quantization/test_quant_api.py
vkuzo: I can fix this, we can assert for length 1 to 2.  The 2nd warning is the one being added in this PR.

andrewor14 (approved):

PR #3296, opened 2025-11-05
float8 inference: fix bmm semantics

vkuzo:
Summary:

Fixes the `Float8Tensor` `torch.bmm` override to match the semantics of the
high precision op. Specifically, input 1 is of shape (B, M, K) and input
2 is of shape (B, K, N).

Previously, the shape expectation from `torch.bmm` was not consistent between high precision and quantized versions, which is confusing.

This is important for quantizing LLaMa 4 MoE variants, which use
`torch.bmm` in the HF implementation.

Test Plan:

```
pytest test/quantization/quantize_/workflows/float8/test_float8_tensor.py -s -x -k bmm
```

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo:
* #3303
* __->__ #3296

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
```diff
@@ -422,24 +422,25 @@ def _(func, types, args, kwargs):
...
             and input_tensor.block_size[-1] == input_tensor.shape[-1]
         ), "bmm only works for per row activation quantization"
 
-        orig_out_features = b_data.shape[-2]
+        orig_out_features = b_data.shape[-1]
 
         res = torch.ops.fbgemm.f8f8bf16_rowwise_batched(
             a_data,
-            b_data,
+            b_data.transpose(-2, -1),
```
jerryzh168, line 441: will performance be a concern?

test/quantization/quantize_/workflows/float8/test_float8_tensor.py
```diff
@@ -444,25 +444,27 @@ def test_bmm(self):
...
-        m = M(weight).eval()
+
+        B, M, K, N = 10, 32, 128, 256
+
+        input = torch.randn(B, M, K, dtype=dtype, device=device)
+        weight = torch.randn(B, N, K, dtype=dtype, device=device)
+        m = Model(weight).eval()
         original = m(input)
-        # we need to transpose the weight first for bmm
-        m.weight = torch.nn.Parameter(m.weight.transpose(1, 2).contiguous())
```
jerryzh168, line 462: this was from llama-models I think: https://github.com/meta-llama/llama-models/blob/0e0b8c519242d5833d8c11bffc1232b77ad7f301/models/llama4/quantization/loader.py#L142, although not as important now

but I guess the important thing is how do we implement it in a way that it can be used by different implementations, would current fp8 bmm implementation work for different ways people use bmm?

vkuzo: this is overriding `torch.bmm`, so we definitely should match the semantics of `torch.bmm` in terms of input shapes.  It doesn't make sense to do a bmm with shapes that aren't (B, M, K) and (B, K, N).  If that breaks llama-models, then they should fix it to match bmm semantics.

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
vkuzo: transpose is just metadata, no impact on performance

PR #3298, opened 2025-11-06
Fix module name extraction logic in quant_api.py

jcaip:
att

### DIFF
--- a/test/quantization/test_quant_api.py
+++ b/test/quantization/test_quant_api.py
@@ -1123,7 +1123,7 @@ class TestFqnToConfig(TestCase):
         self.assertLess(memory_streaming, memory_baseline)
 
     @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
-    def test_quantized_nested_module(self):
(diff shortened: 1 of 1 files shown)


test/quantization/test_quant_api.py
```diff
@@ -1122,6 +1122,31 @@ def reset_memory():
...
+            def __init__(self):
+                super().__init__()
+                self.nested = NestedModule()
+                self.linear1 = torch.nn.Linear(16, 16)
+
+        m = TopLevelModule()
+        quant_config = FqnToConfig(
+            {
+                "nested.linear": Int8WeightOnlyConfig(),
+                "linear1": Int8WeightOnlyConfig(),
```
jerryzh168, line 1142: nit: use new tensors, e.g. Int4WeightOnlyConfig (with Int4Tensor), since we are moving away from AQT

jerryzh168: this comment is missed I think @jcaip

jcaip: oh sorry didn't see this - let me update all of the configs in the tests

PR #3303, opened 2025-11-06
Enable `PerRow(axis)` to support axes other than `-1`

vkuzo:
Summary:

Enables per-row float8 quantization of tensors with shape (B, K, N) across the K dimension.

TODO before land:
1. [done] audit `PerRow` callsites across the codebase and see if anything else needs changing
2. [done] verify serialization will not break

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/quantization/quantize_/workflows/float8/float8_tensor.py
+++ b/torchao/quantization/quantize_/workflows/float8/float8_tensor.py
@@ -11,7 +11,6 @@ from typing import List, Optional
 import torch
 from torch.utils._python_dispatch import return_and_correct_aliasing
 
-from torchao.dtypes.utils import get_out_shape
 from torchao.float8.inference import (
     Float8MMConfig,
     FP8Granularity,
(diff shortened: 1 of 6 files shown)


vkuzo:
* __->__ #3303

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
```diff
@@ -180,6 +180,7 @@ def from_hp(
             and _is_fbgemm_gpu_genai_available()
             and is_sm_at_least_90()
             and isinstance(granularity, PerRow)
+            and granularity.axis in (-1, len(hp_tensor.shape))
```
jcaip: Should this be `range(len(hp_tensor.shape))`? I think
```
input_tensor shape [A, B, C], axis 1 -> scale_shape [1, B, 1]
```

would fail this check

vkuzo: I think it should actually be `(-1, len(hp_tensor.shape) - 1)`, will fix before landing

the goal is to only take this path for "rowwise across the last dimension", which is the default behavior.  The fbgemm kernel does not support quantizing across anything other than the last dimension.

torchao/quantization/utils.py
```diff
@@ -723,8 +723,12 @@ def get_block_size(
                 f"Not all shapes in input shape {input_shape} are divisible by block size {block_size}"
             )
         return block_size
-    elif isinstance(granularity, (PerRow, PerToken)):
+    elif isinstance(granularity, PerToken):
         return (1,) * (len(input_shape) - 1) + (input_shape[-1],)
+    elif isinstance(granularity, PerRow):
+        block_size = [1] * len(input_shape)
+        block_size[granularity.axis] = input_shape[granularity.axis]
```
jerryzh168: nit: row_dim to reduce confusion with PerAxis

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
```diff
@@ -438,7 +440,7 @@ def _(func, types, args, kwargs):
 
         res = torch.ops.fbgemm.f8f8bf16_rowwise_batched(
             a_data,
-            b_data.transpose(-2, -1),
+            b_data.transpose(-2, -1).contiguous(),
```
jerryzh168, line 480: should we add a model level test for this change as well

jerryzh168: oh sorry just saw this

jerryzh168 (approved):
looks good

torchao/testing/utils.py
```diff
@@ -456,15 +458,15 @@ def _test_slice_and_copy_similar_to_vllm(self, config: AOBaseConfig):
             param = l.weight
             param_data = param.data
             param_data = param_data.narrow(output_dim, start_idx, shard_size)
-            orig_value = param_data.qdata[0][0]
+            orig_values = param_data.qdata[0]
```
vkuzo, line 461: the original check failed by chance, the two tensors had the same value at [0][0].  Checking the first row is more resistant to chance.

PR #3304, opened 2025-11-06
Implement aten.add for IntxUnpackedToInt8Tensor

jackzhxng:


metascroy:
I think we should not add an implementation for add unless there is a usecase.  It is not required for quantizing embedding/linear, the main uses for quantize_.

PR #3309, opened 2025-11-07
Skip quantization when channels_out / channels_in are not multiple of 16

jerryzh168:
Summary:
The underlying fbgemm conv3d kernel for float8 only supports channels_out/channels_in are both multiples of 16 so we skip the shapes that doesn't satisfy the requirements for now, we can expand the support to do padding if needed in the future

Test Plan:
(in B200 machine) python test/quantization/quantize_/workflows/float8/test_float8_tensor.py -k test_fp8_conv_skip_quant

### DIFF
--- a/test/quantization/quantize_/workflows/float8/test_float8_tensor.py
+++ b/test/quantization/quantize_/workflows/float8/test_float8_tensor.py
@@ -17,6 +17,7 @@ from torch.testing._internal.common_utils import run_tests
 
 from torchao.quantization import (
     Float8DynamicActivationFloat8WeightConfig,
+    Float8Tensor,
     Float8WeightOnlyConfig,
     Granularity,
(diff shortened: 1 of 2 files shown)


jerryzh168:
CI error unrelated

test/quantization/quantize_/workflows/float8/test_float8_tensor.py
```diff
@@ -349,19 +348,14 @@ def test_fp8_conv_variants(
         self,
         dtype: torch.dtype,
         compile: bool,
-        granularity,
         inference_mode: bool,
         kernel_preference: KernelPreference,
```
jerryzh168, line 352: sorry just noticed that this is not removed, the test is skipped in CI so should not be ran for now, will remove in next PR

PR #3311, opened 2025-11-07
Add torch 2.9 in regression tests

jainapurva:
Add regression test to test torchao against the latest torch 2.9.1

Minor update: Change torch from 2.7.0 -> 2.7.1

### DIFF
--- a/torchao/quantization/pt2e/utils.py
+++ b/torchao/quantization/pt2e/utils.py
@@ -861,7 +861,7 @@ def _get_aten_graph_module_for_pattern(
 
     if torch.__version__.startswith("2.9"):
         # PyTorch 2.9 adds _guards_fn nodes to exported graphs.
-        # These have errors only on torch 2.9 and 2.9.0
+        # These have errors only on torch 2.9.0 and 2.9.1
         for node in list(aten_pattern.graph.nodes):  # type: ignore[union-attr]
             if node.op == "call_module" and node.name == "_guards_fn":
                 aten_pattern.graph.erase_node(node)  # type: ignore[operator, union-attr]


torchao/quantization/pt2e/utils.py
```diff
@@ -859,6 +859,15 @@ def _get_aten_graph_module_for_pattern(
...
             aten_pattern.graph.erase_node(node)  # type: ignore[operator, union-attr]
 
+    # PyTorch 2.9+ adds _guards_fn nodes to exported graphs.
+    # These should not be part of pattern matching, so remove them.
+    for node in list(aten_pattern.graph.nodes):  # type: ignore[union-attr]
+        if node.op == "call_module" and node.name == "_guards_fn":
+            aten_pattern.graph.erase_node(node)  # type: ignore[operator, union-attr]
+            # Also remove the _guards_fn module from the graph module if it exists
+            if hasattr(aten_pattern, "_guards_fn"):
+                delattr(aten_pattern, "_guards_fn")
```
andrewor14: I ran into this myself yesterday. This was resolved if I upgrade to the latest main (torch-2.10.0dev), do we still need to do this? Will CI fail without this?

jainapurva: It doesn't happen on main, but I'm adding tests for torch2.9, hence will need to do this. Probably one work around could be to put this block of code in a condition to run only in case of torch 2.9 or 2.9.1

andrewor14: Does it also fail with 2.9.1? I only tested with 2.9.0. If so that sounds good to me

jainapurva: Yes, it fails for both 2.9.0 and 2.9.1. I added a condition to run it only if it's torch 2.9

PR #3315, opened 2025-11-08
Add per tensor fp8 conv2d support

jerryzh168:
Summary:
Add fp8 conv2d support, using the same conv3d kernels, by setting the D dimension to 1.

1. unsqueeze both input and weight in dim 2 ( the D dimension)
2. call fp8 conv3d op from fbgemm `torch.ops.fbgemm.f8f8bf16_conv`
3. assert D dimension shape to be 1 and call sequeeze at dim 2: res.squeeze(2) to remove the D dimension

Test Plan:
```
python test/quantization/quantize_/workflows/float8/test_float8_tensor.py -k test_unsqueeze_conv2d_weight 
```

### DIFF
--- a/torchao/quantization/quant_api.py
+++ b/torchao/quantization/quant_api.py
@@ -1816,13 +1816,19 @@ def _float8_dynamic_activation_float8_weight_quantize_tensor(weight, config):
     _check_hardware_support(granularity)
     activation_granularity, weight_granularity = granularity
 
-    if weight.dim() == 5:
-        # weights for conv3d
+    # Note: right now we assume it's weights of conv2d and conv3d purely based
+    # on the dimension of weight, currently there is no conflict with linear 2d
+    # and moe weights 3d
+    # if we need to support conv1d, which also has 3d weight, we may have to
+    # pass around the module as well to distinguish between conv1d and 3d moe weight
+    if weight.dim() in [4, 5]:
+        # weights for conv2d or 3d
         assert isinstance(activation_granularity, PerTensor) and isinstance(
             weight_granularity, PerTensor
-        ), "5D tensor only supports per tensor activation and weight quantization"
+        ), "4D/5D tensor only supports per tensor activation and weight quantization"
 
-        # weight dim: (C_out, C_in, K1, K2, K3)
+        # conv3d weight dim: (C_out, C_in, K1, K2, K3)
+        # conv2d weight dim: (C_out, C_in, K1, K2)
         # skip quantization when either C_out or C_in
         # is not a multiple of 16
         if weight.shape[0] % 16 != 0 or weight.shape[1] % 16 != 0:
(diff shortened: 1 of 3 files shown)


torchao/quantization/quantize_/workflows/float8/float8_tensor.py
```diff
@@ -572,10 +573,71 @@ def _(func, types, args, kwargs):
...
+        # (N, C, H, W) --> (N, C, 1, H, W)
+        input_tensor = input_tensor.unsqueeze(2)
+        weight_tensor = weight_tensor.unsqueeze(2)
+        assert tuple(output_padding) == (0, 0), (
+            f"Only (0, 0) is supported for `output_padding`, got: f{output_padding}"
+        )
+        padding = [0, *padding]
+        stride = [1, *stride]
+        dilation = [1, *dilation]
+        res = _quantize_and_scaled_conv3d(
```
andrewor14, line 597: is this unsqueezing to turn 2d into 3d? Does fbgemm only have 3d conv kernels?

torchao/quantization/quant_api.py
```diff
@@ -1816,13 +1816,14 @@ def _float8_dynamic_activation_float8_weight_quantize_tensor(weight, config):
     _check_hardware_support(granularity)
     activation_granularity, weight_granularity = granularity
 
-    if weight.dim() == 5:
-        # weights for conv3d
+    if weight.dim() in [4, 5]:
```
andrewor14, line 1824: is there anything more robust we can check here?

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
```diff
@@ -596,16 +658,36 @@ def _(func, types, args, kwargs):
...
+    # (N, C, H, W) --> (N, C, 1, H, W)
+    # memory_format of both tensors should be torch.channels_last
+    # and it should be preserved with unsqueeze(2) (becoming torch.channels_last_3d)
+    assert input_tensor.is_contiguous(
+        memory_format=torch.channels_last
+    ) and weight_tensor.qdata.is_contiguous(memory_format=torch.channels_last), (
+        "Please make sure both activation and weights are in the `channels_last` memory_format"
+    )
+    input_tensor = input_tensor.unsqueeze(2)
+    weight_tensor = weight_tensor.unsqueeze(2)
```
andrewor14, line 673: Can we reuse this code, e.g. have aten.conv2d call into aten.convolution?

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
jerryzh168: yeah this turns 2d conv to 3d and it's because fbgemm only supports 3d conv right now

torchao/quantization/quant_api.py
jerryzh168: we can't really distinguish from here whether it's linear and conv weight I think (although right now seems linear is 2d/3d and conv is 4d/5d, maybe conv1d could have 3d weight, which is an overlap with linear)

but we could potentially separate the handling of conv and linear by passing around the module as well, if this is needed in the future

I can add a comment for now

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
jerryzh168: yeah I think that should be possible for both aten.conv2d and aten.conv3d, I can refactor that in next PR?

PR #3321, opened 2025-11-10
Pin pytest==8.4.2

huydhn:
Testing to see if this fixes https://github.com/pytorch/ao/actions/runs/19239143137/job/54997062535 and https://github.com/pytorch/ao/actions/runs/19239143151

cc @jerryzh168


atalman (approved):

jerryzh168 (approved):
I think we can land this, seems L4 test error is gone now

PR #3327, opened 2025-11-11
Use conda libgcc-ng 11.2

atalman:


jerryzh168:
thanks, looks like all other jobs are fixed except for the nightly, what is the issue there?

huydhn:
I guess if we land this, then nightly will be fixed by tmrw?

jerryzh168:
> I guess if we land this, then nightly will be fixed by tmrw?

I see, why is that? this is talking about torch cpu nightly I think

but I can merge now since it's better than current state

PR #3332, opened 2025-11-12
[xpu][feature] Enable fp8 on XPU

Stonepia:
# Motivation 
We have enabled the [scaled_mm ](https://github.com/pytorch/pytorch/pull/166056/files)in the PyTorch for XPU, so the fp8 support has the initial stage for tensorwise/rowwise scaling. This PR is used to help the user to run the fp8 models/tests with torchAO. 

User can use  same quantization config Float8DynamicActivationFloat8WeightConfig as CUDA to run the fp8 models on XPU, the device dispatch is finished by the scaled_mm op.

This PR has no API related change. It does the following:
1. Fix the CUDA related API assertions.
2. Set `use_fast_accum=False` for XPU, because XPU does not have the fast_accum.
3. Enable fp8 related tests.

### DIFF
--- a/torchao/float8/inference.py
+++ b/torchao/float8/inference.py
@@ -288,7 +288,7 @@ def _check_hardware_support(
         assert torch.xpu.is_available() or (
             torch.cuda.is_available() and is_sm_at_least_89() or is_MI300()
         ), (
-            "Float8 dynamic quantization requires CUDA compute capability ≥8.9 or MI300+."
+            "Float8 dynamic quantization requires CUDA compute capability ≥8.9 or MI300+ or XPU."
         )
     elif is_a_1_128_w_128_128:
         # TODO(future PR): look into AMD support


liangan1:
@Stonepia  since the scaled_mm has been merged. pls rebase this pr again.

liangan1 (approved):

torchao/testing/utils.py
```diff
@@ -496,15 +497,15 @@ def _test_slice_and_copy_similar_to_vllm(self, config: AOBaseConfig):
         # and does not use tensor parallelism
 
         dtype = torch.bfloat16
-        device = "cuda"
-        l = torch.nn.Linear(1024, 1024, device="cuda", dtype=dtype)
+        device = get_current_accelerator_device()
```
jerryzh168, line 501: what happens when this is None?

Stonepia: Thanks for the remainder! I added an assertion in https://github.com/pytorch/ao/pull/3332/commits/a1a15bcfea847be8b636345e2adf3660e6d61678

BTW this is the definition of accelerator. 

```C++
# torch/..../DeviceAccelerator.cpp
bool isAccelerator(c10::DeviceType device_type) {
  switch (device_type) {
    case at::kCUDA:
    case at::kMTIA:
    case at::kXPU:
    case at::kHIP:
    case at::kMPS:
    case at::kHPU:
    case at::kPrivateUse1:
      return true;
    default:
      return false;
  }
}

```

jerryzh168: oh, I mean we use this everywhere currently, if it's None, then `torch.nn.Linear(1024, 1024, device=device, dtype=dtype)` will be using CPU?

you can fix this callsite with assert, but will need to do an audit for all the callsites in the codebase to make sure the None case is also properly handled I think

test/quantization/quantize_/workflows/float8/test_float8_tensor.py
```diff
@@ -40,6 +41,9 @@
 # Needed since changing args to function causes recompiles
 torch._dynamo.config.cache_size_limit = 128
 
+_DEVICE = get_current_accelerator_device()
+assert _DEVICE is not None, "Accelerator device is required for float8 tests"
```
jerryzh168: this will make the CI fail I think, we could be running torchao in CPU only hardwares

Stonepia: Thanks for the quick reply! Yes this should not be added. I deleted it in https://github.com/pytorch/ao/pull/3332/commits/440de1d2f5ee0af9ef47f1e7a938d6040b0cd3f9

Meanwhile, since the whole fp8 tests is wrapped like below, so I didn't do any other specific check.

```Python
@unittest.skipIf(
    not torch.accelerator.is_available(), "skipping when gpu is not available"
)
class TestFloat8Tensor(TorchAOIntegrationTestCase):
```

torchao/testing/utils.py
Stonepia: Thanks for the pointing out. 

Yes there should be another PR to make the following code change:

```diff
- @unittest.skipIf(_DEVICE is None, "skipping when GPU is not available")
+ @unittest.skipIf(torch.accelerator.is_available() "skipping when GPU is not available")
...

- def get_current_accelerator_device():
+ def get_current_device():
    if torch.accelerator.is_available():
        return torch.accelerator.current_accelerator()
    else:
-        return None
+       return "cpu"
```

For the callsite in `_test_slice_and_copy_similar_to_vllm()` specifically, I think using an assert is ok? Because this test formerlly in design should only on cuda, so we need an assertion to make sure there is a GPU.

test/quantization/quantize_/workflows/float8/test_float8_tensor.py
Stonepia: This change is in https://github.com/pytorch/ao/pull/3332/commits/440de1d2f5ee0af9ef47f1e7a938d6040b0cd3f9

torchao/testing/utils.py
Stonepia: Changed in https://github.com/pytorch/ao/pull/3332/commits/c84a5abc94cb354d7adaad76679cd221bc4ad13c

liangan1:
@Stonepia  pls rebase this PR.

torchao/testing/utils.py
Stonepia: Emmm, I rethinked about it. This test specifically only test on CUDA, so I still prefer to add an assertion here, it should not fallback to CPU from my perspective.

But I do think the `get_current_accelerator_device()` should not silently return None, and it at least should have a error handling. I will submit another pr for that.

liangan1: @jerryzh168 this is a utils function which is used by other UTs, and these UTs are triggered only when the torch.accelerator.is_available(), So this assert will be always true.

jerryzh168: OK, I think it's probably better to make this more "self contained" (meaning without relying on it's been called only after `torch.accelerator.is_available()`) but can be done separately

liangan1: Thanks.  @Stonepia  pls take this AR.

PR #3335, opened 2025-11-13
[Intel XPU] Add common code for UTs on XPU

zxd1997066:
This PR is targeted to add common code for UTs on XPU

### DIFF
--- a/torchao/utils.py
+++ b/torchao/utils.py
@@ -137,7 +137,7 @@ def get_available_devices():
     return devices
 
 
-def auto_detect_device():
+def get_current_accelerator_device():
     if torch.accelerator.is_available():
         return torch.accelerator.current_accelerator()
     else:


.github/scripts/ci_test_xpu.sh
```diff
@@ -14,4 +14,16 @@ python3 -c "import torch; import torchao; print(f'Torch version: {torch.__versio
 
 pip install pytest expecttest parameterized accelerate hf_transfer 'modelscope!=1.15.0'
 
-pytest -v -s torchao/test/quantization/quantize_/workflows/int4/test_int4_plain_int32_tensor.py 
+pytest -v -s torchao/test/quantization/quantize_/workflows/int4/test_int4_plain_int32_tensor.py
+
+pytest -v -s torchao/test/quantization/
```
jerryzh168, line 17: does these tests run on xpu?

.github/scripts/ci_test_xpu.sh
```diff
@@ -14,4 +14,16 @@ python3 -c "import torch; import torchao; print(f'Torch version: {torch.__versio
 
 pip install pytest expecttest parameterized accelerate hf_transfer 'modelscope!=1.15.0'
 
-pytest -v -s torchao/test/quantization/quantize_/workflows/int4/test_int4_plain_int32_tensor.py 
+pytest -v -s torchao/test/quantization/quantize_/workflows/int4/test_int4_plain_int32_tensor.py
```
liangan1:
```suggestion

```

.github/scripts/ci_test_xpu.sh
liangan1: These UTs will be run on XPU finally.  The current TorchAO CI should only cover the CUDA device and skip on XPU. We plan to firstly add these test folders into the XPU ci which is triggered by ciflow/xpu and then enabled the XPU UTs file by file to reduce the review effort.

torchao/testing/utils.py
```diff
@@ -98,6 +98,47 @@ def wrapper(*args, **kwargs):
     return decorator
 
 
+def skip_if_no_xpu(message=None):
```
jerryzh168: currently these things are a bit confusing I think, since I remember some of these work with unittests and others work with pytest. can you confirm what these are working with? I think ideally we should create two versions, one for unittests one for pytest

zxd1997066: Created two versions for both unittest and pytest

torchao/utils.py
```diff
@@ -137,6 +137,13 @@ def get_available_devices():
     return devices
 
 
+def auto_detect_device():
```
jerryzh168: nit: it would be better to change this to something like `get_current_accelerator_device` to be clearer I think

jerryzh168 (approved):
lg, see one comment inline

torchao/utils.py
liangan1: Make sense. This name is more reasonable. Refined.

PR #3337, opened 2025-11-13
[pt2e] Fix QAT annotations for special qspecs

andrewor14:
**Summary:** In pt2e QAT, we first annotate the nodes to be quantized and then perform the pattern replacement (i.e. QAT fusion) in the prepare step. After this pattern replacement, old nodes are replaced with new nodes, and any references to the old nodes must also be updated to refer to the new nodes.

This commit fixes a bug where, for special qspecs like the `SharedQuantizationSpec` and `DerivedQuantizationSpec`, we only update the values of a node's `input_qspec_map`, not the keys. As a result, the keys still refer to the old nodes that do not exist anymore after the QAT fusion.

**Test Plan:**
```
python test/quantization/pt2e/test_quantize_pt2e_qat.py -k test_qat_shared_qspec
```

### DIFF
--- a/test/quantization/pt2e/test_quantize_pt2e_qat.py
+++ b/test/quantization/pt2e/test_quantize_pt2e_qat.py
@@ -8,7 +8,7 @@
 import copy
 import operator
 import unittest
-from typing import Any, Optional
+from typing import Any, Optional, Tuple
 
 import torch
 from torch.ao.quantization import QConfigMapping
@@ -46,6 +46,7 @@ from torchao.quantization.pt2e.quantizer import (
     QuantizationAnnotation,
     QuantizationSpec,
     Quantizer,
+    SharedQuantizationSpec,
 )
 from torchao.testing.pt2e._xnnpack_quantizer import (
     XNNPACKQuantizer,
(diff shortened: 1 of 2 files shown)


andrewor14:
> ah makes sense, can you add a test

Yep, this is in progress, just wanted to push first for Xing to try

test/quantization/pt2e/test_quantize_pt2e_qat.py
```diff
@@ -878,6 +879,44 @@ class TestQuantizePT2EQAT_ConvBn2d(TestQuantizePT2EQAT_ConvBn_Base):
...
+                        (src, dest) = qspec.edge_or_node
+                        nodes_to_check.update([src, dest])
+
+        # assert that none of the nodes refer to old nodes
+        self.assertEqual(len(nodes_to_check), 5)
+        num_batch_norm_nodes_checked = 0
+        for n in nodes_to_check:
+            if n.target == torch.ops.aten.batch_norm.default:
+                num_batch_norm_nodes_checked += 1
+            self.assertTrue(n not in old_nodes, "found old node in qspec")
```
jerryzh168: is it possible to also check they are all in new_nodes?

PR #3343, opened 2025-11-15
Add `enable_fusion_modeling` for conv2d and conv3d

jerryzh168:
Summary:
This is to have more accurate benchmarking for fusion

Test Plan:
```
RECIPE_NAME="tensorwise"
SHAPE_GEN_NAME="custom"
SCRIPT_PATH="benchmarks/float8/float8_inference_roofline.py"
M=1
K=160
N=320
D=3
H=194
W=130
kernel_size=3

OUTPUT_FILE="~/local/tmp/test_${M}_${K}_${N}_${D}_${H}_${W}_${kernel_size}.csv"
python $SCRIPT_PATH $OUTPUT_FILE \
    --recipe_name $RECIPE_NAME \
    --shape_gen_name $SHAPE_GEN_NAME \
    --M $M --K $K --N $N  \
    --D $D --H $H --W $W  \
    --kernel_size $kernel_size \
    --enable_fusion_modeling \
    --op_name conv3d
```
Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/benchmarks/float8/float8_inference_roofline.py
+++ b/benchmarks/float8/float8_inference_roofline.py
@@ -370,13 +370,27 @@ def run(
         if do_benchmarks:
             # create the model
             if op_name == "conv2d":
-                m_orig = nn.Sequential(
-                    nn.Conv2d(K_val, N_val, kernel_size, bias=False)
-                ).to(memory_format=torch.channels_last)
+                if not enable_fusion_modeling:
+                    m_orig = nn.Sequential(
+                        nn.Conv2d(K_val, N_val, kernel_size, bias=False)
+                    ).to(memory_format=torch.channels_last)
+                else:
+                    m_orig = nn.Sequential(
+                        nn.ReLU(),
+                        nn.Conv2d(K_val, N_val, kernel_size, bias=False),
+                        nn.ReLU(),
+                    ).to(memory_format=torch.channels_last)
             elif op_name == "conv3d":
-                m_orig = nn.Sequential(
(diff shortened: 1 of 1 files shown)


benchmarks/float8/float8_inference_roofline.py
```diff
@@ -370,13 +370,27 @@ def run(
...
-                ).to(memory_format=torch.channels_last)
+                if not enable_fusion_modeling:
+                    m_orig = nn.Sequential(
+                        nn.Conv2d(K_val, N_val, kernel_size, bias=False)
+                    ).to(memory_format=torch.channels_last)
+                else:
+                    m_orig = nn.Sequential(
+                        nn.ReLU(),
+                        nn.Conv2d(K_val, N_val, kernel_size, bias=False),
+                        nn.ReLU(),
```
vkuzo, line 381: is there a reason to have an epilogue here?  It's nice to have a prologue to measure the fusion of preceding activation, but unless the conv kernel does epilogue fusion I'm not sure we need an epilogue here, it would just measure the same thing for both bf16 and lowp.

jerryzh168: oh I just saw conv was typically surrounded by activations/norms, do we only support prologue fusions typically?

jerryzh168: just checked, we'll have epilogue fusion for conv op. and can explore prologue fusion for the quant kernels later. so the current test is OK I think

PR #3345, opened 2025-11-16
Separate granularity check in scale slicing for less abstraction

namgyu-youn:
**Summary**:
For better readability and robust handling, explicitly separate quantization granularity cases (per-tensor, per-row, per-block) with early returns for simpler cases.

**Test plan**:
test/quantization/quantize_/workflows/float8/test_float8_tensor.py

### DIFF
--- a/torchao/float8/inference.py
+++ b/torchao/float8/inference.py
@@ -142,40 +142,40 @@ def _slice_scale_for_dimension(
     """
     aten = torch.ops.aten
 
-    # Unsupported case for now, this would be 1 scale per data element
-    if scale.shape == data_shape:
-        return aten.slice.Tensor(scale, dim, start, end, step)
-
-    # Reconstruct block sizes based on data shape and scale shape
-    block_sizes = tuple(data_shape[i] // scale.shape[i] for i in range(len(data_shape)))
-
-    if dim >= len(block_sizes):
-        # Slicing beyond the dimensions we care about
+    # Case 1: Per-tensor quantization (scalar scale)
+    if scale.numel() <= 1:
         return scale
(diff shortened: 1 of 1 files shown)


torchao/float8/inference.py
```diff
@@ -142,40 +142,40 @@ def _slice_scale_for_dimension(
...
-    if scale.shape == data_shape:
-        return aten.slice.Tensor(scale, dim, start, end, step)
-
-    # Reconstruct block sizes based on data shape and scale shape
-    block_sizes = tuple(data_shape[i] // scale.shape[i] for i in range(len(data_shape)))
-
-    if dim >= len(block_sizes):
-        # Slicing beyond the dimensions we care about
+    # Case 1: Per-tensor quantization (scalar scale)
+    if scale.numel() <= 1:
```
jerryzh168, line 146: does it work with special casing per tensor and per row?

torchao/float8/inference.py
```diff
@@ -142,40 +142,40 @@ def _slice_scale_for_dimension(
...
-                "Slicing with step > 1 is not implemented for scale tensors."
-            )
+    # There is blocking in this dimension
+    # Calculate which scale elements correspond to the sliced data
+    scale_start = start // block_size_for_dim if start is not None else None
+    scale_end = (
+        (end + block_size_for_dim - 1) // block_size_for_dim
+        if end is not None
+        else None
+    )
```
jerryzh168, line 176: I feel ideally this code should be general enough to handle all cases

namgyu-youn:
Close because current implementation is still enough to handle

PR #3352, opened 2025-11-18
Align memory_format for conv2d/3d in Float8Tensor with hp Tensor

jerryzh168:
Summary:
att, we want to make sure the output of

`F.conv3d(input, weight, ...)` and `F.conv3d(input, fp8_weight, ...)` have the same memory_format

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/quantization/quantize_/workflows/float8/float8_tensor.py
+++ b/torchao/quantization/quantize_/workflows/float8/float8_tensor.py
@@ -533,36 +533,60 @@ def _quantize_and_scaled_conv3d(
             )
 
     assert kernel_choice == "fbgemm", "Only fbgemm kernel choice is supported currently"
+    input_qdata = input_tensor.qdata
+    weight_qdata = weight_tensor.qdata
+
+    is_input_channels_last = input_qdata.is_contiguous(
+        memory_format=torch.channels_last_3d
+    )
+    is_weight_channels_last = weight_qdata.is_contiguous(
+        memory_format=torch.channels_last_3d
+    )
+
+    # convert the input/weight to channels_last_3d memory_format here
+    # to make sure we can call the fbgemm conv
+    # kernel, it should be a no-op if both activation and weight are in
+    # channels_last_3d memory_format
+    input_qdata = input_qdata.contiguous(memory_format=torch.channels_last_3d)
+    weight_qdata = weight_qdata.contiguous(memory_format=torch.channels_last_3d)
+
     # move C_in to last dim
     # after permute: (N, D, H, W, C_in)
-    act_qdata = input_tensor.qdata.permute([0, 2, 3, 4, 1])
+    input_qdata = input_qdata.permute([0, 2, 3, 4, 1])
 
     # move C_in to last dim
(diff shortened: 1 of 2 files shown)


torchao/quantization/quantize_/workflows/float8/float8_tensor.py
```diff
@@ -558,11 +567,22 @@ def _quantize_and_scaled_conv3d(
     )
     # output shape after permute: N, C_out, D_out, H_out, W_out
     output = output.permute([0, 4, 1, 2, 3])
+
+    # aligning the semantics with bfloat16 conv ops, the
+    # output should use channels_last format as long as any of the
+    # input or weight is channels_last
+    if is_input_channels_last or is_weight_channels_last:
+        output = output.to(memory_format=torch.channels_last_3d)
```
jbschlosser: I think this is the right thing to semantics-wise, but note that this will incur a copy if the output isn't already in channels_last. Ideally, the kernel itself would output into channels_last directly to avoid the copy.

Edit: oh I think you're already aware of this :)

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
```diff
@@ -542,9 +549,11 @@ def _quantize_and_scaled_conv3d(
...
     weight_qdata = weight_tensor.qdata.permute([0, 2, 3, 4, 1])
 
-    assert act_qdata.is_contiguous() and weight_qdata.is_contiguous(), (
-        "Please make sure both activation and weights are in the `channels_last_3d` memory_format"
-    )
+    # calling contiguous() here to make sure we can call the fbgemm conv
+    # kernel, it should be a no-op if both activation and weight are in
+    # channels_last_3d memory_format
+    act_qdata = act_qdata.contiguous()
+    weight_qdata = weight_qdata.contiguous()
```
jbschlosser: ~~I don't think these `contiguous()` calls are the right thing to do here - note that this will clobber `channels_last` for the activation and weight. Calling `contiguous(memory_format=torch.channels_last)` would be more correct~~

Edit: from offline discussion, can't forget the `permute()`! we want a `contiguous()` `(N, D, H, W, C_in)` tensor, which is equivalent to a properly-permuted, `contiguous(channels_last)` `(N, C_in, D, H, W)` tensor

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
jerryzh168: yes, the fbgemm kernel should output a tensor that's already in this format, so this becomes a no-op when either input or weight is in channels_last format

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
jerryzh168: refactored this to first to contiguous(memory_format=torch.channels_last_3d) and then do permute to make it easier to follow

vkuzo (approved):

PR #3370, opened 2025-11-21
Add PerBlock to safe globals

andrewor14:
**Summary:** Add PerBlock to safe globals so users don't have to do this themselves when they load config.json with PerBlock.

```
WeightsUnpickler error: Unsupported global: GLOBAL torchao.quantization.granularity.PerBlock was not an allowed global by default. Please use `torch.serialization.add_safe_globals([torchao.quantization.granularity.PerBlock])` or the `torch.serialization.safe_globals([torchao.quantization.granularity.PerBlock])` context manager to allowlist this global if you trust this class/function.
```

**Test Plan:**
```
python test/core/test_config.py -k test_granularity_serialization
```

### DIFF
--- a/test/core/test_config.py
+++ b/test/core/test_config.py
@@ -6,6 +6,7 @@
 
 import json
 import os
+import subprocess
 import tempfile
 import warnings
 from dataclasses import dataclass
@@ -23,7 +24,11 @@ from torchao.prototype.awq import (
     AWQConfig,
     AWQStep,
 )
-from torchao.quantization import PerBlock
+from torchao.quantization import (
+    PerBlock,
+    PerRow,
+    PerTensor,
+)
 from torchao.quantization.quant_api import (
     Float8DynamicActivationFloat8WeightConfig,
     Float8DynamicActivationInt4WeightConfig,
@@ -36,10 +41,11 @@ from torchao.quantization.quant_api import (
     Int8DynamicActivationInt8WeightConfig,
     Int8WeightOnlyConfig,
     ModuleFqnToConfig,
-    PerRow,
     UIntXWeightOnlyConfig,
+    quantize_,
 )
 from torchao.sparsity.sparse_api import BlockSparseWeightConfig, SemiSparseWeightConfig
+from torchao.utils import is_sm_at_least_89
 
 # Define test configurations as fixtures
 configs = [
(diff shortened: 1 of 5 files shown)


vkuzo:
I think we should
1. put this in the same place where the other granularities are exposed for serialization
2. add a test (same as we should have for other granularities)

andrewor14:
> put this in the same place where the other granularities are exposed for serialization

Today these are added in observer.py, I think it just so happens that they're called when loading that file indirectly: https://github.com/pytorch/ao/blob/main/torchao/quantization/observer.py#L356

Should I move all of these to granularity.py to be more explicit?

jerryzh168:
> > put this in the same place where the other granularities are exposed for serialization
> 
> Today these are added in observer.py, I think it just so happens that they're called when loading that file indirectly: [`main`/torchao/quantization/observer.py#L356](https://github.com/pytorch/ao/blob/main/torchao/quantization/observer.py#L356)
> 
> Should I move all of these to granularity.py to be more explicit?

yeah I think just adding all used granularity would be good, and remove these from observer.py

torchao/quantization/quant_api.py
```diff
@@ -332,10 +332,10 @@ def insert_observers_(
     ```
         import torch
         import torch.nn as nn
+        from torchao.quantization.granularity import PerTensor
```
jerryzh168: nit: just `import torchao.quantization import PerTensor` is better I think

test/dtypes/test_affine_quantized_tensor_parallel.py
```diff
@@ -22,7 +22,7 @@
     Int8DynamicActivationInt8WeightConfig,
     Int8WeightOnlyConfig,
 )
-from torchao.quantization.observer import PerRow, PerTensor
+from torchao.quantization.granularity import PerRow, PerTensor
```
jerryzh168: nit: remove granularity

test/core/test_config.py
```diff
@@ -155,6 +160,41 @@ def test_reconstructable_dict_file_round_trip(config):
...
+        torch.save(m.state_dict(), f.name)
+        fname = f.name
+
+    assert fname is not None
+
+    code = f"""
+import torch
+import torchao
+_ = torch.load('{fname}', weights_only=True)
+    """
```
jerryzh168, line 194: why do we need these instead of just code?

andrewor14: It's because by the time we run the test we have already imported everything. This starts a fresh environment and shows you only need to import torchao for loading to work. Copied from: https://github.com/pytorch/ao/blob/main/test/prototype/mx_formats/test_mx_serialization.py#L36

andrewor14:
Test failures don't seem related. Thanks, merging this

PR #3371, opened 2025-11-21
Add conv2d support for IntxUnpackedToInt8Tensor

andrewor14:
**Summary:** This enables `Int8DynamicActivationIntxWeightConfig` and `IntxWeightOnlyConfig` for conv2d.

**Test Plan:**
```
python test/quantization/quantize_/workflows/intx/test_intx_unpacked_to_int8_tensor.py -k test_conv2d
```

### DIFF
--- a/torchao/quantization/quant_api.py
+++ b/torchao/quantization/quant_api.py
@@ -2337,20 +2337,32 @@ def _intx_weight_only_quantize_tensor(
     intx_packing_format = config.intx_packing_format
     intx_choose_qparams_algorithm = config.intx_choose_qparams_algorithm
 
-    assert weight.dim() == 2, (
-        f"IntxWeightOnlyConfig only works for 2-d Tensor, got: {weight.dim()}"
-    )
+    if weight.dim() == 2:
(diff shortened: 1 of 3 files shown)


torchao/quantization/quant_api.py
```diff
@@ -2350,7 +2350,11 @@ def _intx_weight_only_quantize_tensor(
     else:
         raise ValueError(f"granularity must be PerGroup or PerAxis, got {granularity}")
 
-    block_size = (1, group_size)
+    if weight.dim() == 2:
+        block_size = (1, group_size)
+    else:
+        assert weight.dim() == 4
+        block_size = (1, group_size, 1, 1)
```
metascroy, line 2365: I think this doesn't work with the logic to select groupsize as group_size=weight.shape[-1] with peraxis granularity on line 2349?

Maybe introduce a new "input_dim" variable that is -1 (or 1) for linear, and it is 1 for conv2d?

andrewor14: ah you're right, will fix

Issue #3324, opened 2025-11-11
Align `_choose_scale_float8`'s use of `block_size` with other code


jerryzh168:
Recently I found that for `_choose_scale_float8`:

`len(block_size) == 0` for per tensor: https://github.com/pytorch/ao/blob/b4ec4cbab799b5320fc1ca094e0398b3ad967dbf/torchao/quantization/quant_primitives.py#L2303

`keepdim=True` for others: https://github.com/pytorch/ao/blob/b4ec4cbab799b5320fc1ca094e0398b3ad967dbf/torchao/quantization/quant_primitives.py#L2314

While for `_choose_qparams_affine` (for int):

we use `keepdim=False`: https://github.com/pytorch/ao/blob/b4ec4cbab799b5320fc1ca094e0398b3ad967dbf/torchao/quantization/quant_primitives.py#L1552-L1553 that will make per tensor quantization to have a scalar scale automatically.

I think we should fix this discrepancy to reduce confusion, by aligning `_choose_scale_float8` with the int one, i.e. change `keepdim=True` to `keepdim=False` and make sure the tests work.

we can also add some docs afterwards to both ops to clarify.

jerryzh168:
@namgyu-youn I updated this issue a bit, I think we can actually align the integer `choose_qparams_affine` with the float8 one, and keep the original rank instead of shrink it, this would make the scale/zero_point easier to handle overall

created a separate issue for float8: https://github.com/pytorch/ao/issues/3414

AryanBagade:
@jerryzh168 if this issue is open for external contributors, shall i work on it?

jerryzh168:
yes, please go ahead, thanks for helping! we also have internal tests that maybe affected so I can help with testing that when you put up the PR

AryanBagade:
@jerryzh168 created a PR, thanks for letting me contribute, once this is merged will work on its twin issue !

namgyu-youn:
> [@namgyu-youn](https://github.com/namgyu-youn) I updated this issue a bit, I think we can actually align the integer `choose_qparams_affine` with the float8 one, and keep the original rank instead of shrink it, this would make the scale/zero_point easier to handle overall
> 
> created a separate issue for float8: [#3414](https://github.com/pytorch/ao/issues/3414)

Thanks so much for the info Jerry. I was too busy to work on this, and hope it will be resolved by @AryanBagade .

[...]

PR #3374, opened 2025-11-22
[WIP] update keepdim params for 1D scale factor

namgyu-youn:
**Overview**:
In `_choose_scale_float8`, the per-tensor quantization case (`len(block_size) == 0`) uses `tensor.amax(keepdim=True)` while `_choose_qparams_affine` uses `torch.amax(..., keepdim=False)` for the same purpose.

This PR aligns `_choose_scale_float8` with `_choose_qparams_affine` by using `tensor.amax(keepdim=False)` for 1D scale factor.

**Related Issue/PR**: #3324

**Test Plan**: test/quantization/test_quant_primitives.py

### DIFF
--- a/torchao/quantization/quant_primitives.py
+++ b/torchao/quantization/quant_primitives.py
@@ -2311,7 +2311,7 @@ def _choose_scale_float8(
             block_size, tensor.shape
         )
         tensor_reshaped = tensor.view(shape_for_reduction)
-        max_abs = tensor_reshaped.abs().amax(dim=reduction_dims, keepdim=True)
+        max_abs = tensor_reshaped.abs().amax(dim=reduction_dims, keepdim=False)
         if hp_value_lb is not None or hp_value_ub is not None:
             max_abs = torch.clamp(max_abs, min=hp_value_lb, max=hp_value_ub)
         scale = max_abs / quant_max


namgyu-youn:
Local test code:
```python
import torch
from torchao.quantization.quant_primitives import _choose_scale_float8, _choose_scale_float8_old

a = torch.randn(4, 4)
scale = _choose_scale_float8(a, block_size=(4, 1))  # keepdim = False
print(scale)

scale_old = _choose_scale_float8_old(a, block_size=(4, 1))  # keepdim = True
print(scale_old)
```

And the result is (same):
```
tensor([[0.0023, 0.0044, 0.0040, 0.0027]])
tensor([[0.0023, 0.0044, 0.0040, 0.0027]])
```

jerryzh168:
this is not enough I think there are https://github.com/pytorch/ao/blob/2ff1eb2e356275cfbe46832327387d382c72945d/torchao/quantization/quant_primitives.py#L2320-L2323 that will expand the dimension of scale

I actually tried this locally, and doesn't seem to work very well. I think we can leave this for now, but one thing that is useful is to remove the need to do https://github.com/pytorch/ao/blob/2ff1eb2e356275cfbe46832327387d382c72945d/torchao/quantization/quant_primitives.py#L2332, I don't know why we need this if float8 is already using a scale that's matching the dimension of input: https://github.com/pytorch/ao/blob/2ff1eb2e356275cfbe46832327387d382c72945d/torchao/quantization/quant_primitives.py#L2318-L2323

maybe try removing calls to `_maybe_expand_scale_to_tensor_shape` in code and make sure all tests still pass is a good task to work on 


also another thing we can do is to simplify the implementation of https://github.com/pytorch/ao/blob/2ff1eb2e356275cfbe46832327387d382c72945d/torchao/float8/inference.py#L130 since for float8 scale is always matching the dim of input

PR #3383, opened 2025-11-24
3/x mx cleanup: speed up local test suite

vkuzo:
Summary:

* Before, worst case: ~4 minutes
* Before, best case: ~2 minutes
* After: ~82 sec

Changes:
1. test smaller shapes in triton kernel tests, sometimes the large
   shapes are expensive in local runs
2. remove testing of mxfp4 + compile, it's really slow and not useful
   yet

Test Plan:

```
```

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo:
* #3398
* #3387
* #3386
* #3385
* __->__ #3383

test/prototype/mx_formats/test_mx_linear.py
```diff
@@ -258,7 +260,6 @@ def test_activation_checkpointing():
     "scale_calculation_mode",
     [
         ScaleCalculationMode.FLOOR,
-        ScaleCalculationMode.CEIL,
```
danielvegamyhre, line 261: is this removed because in practice we're only using floor and rceil?

vkuzo: yep!

CEIL is still tested for eager, this removes it from compile

jerryzh168 (approved):

PR #3384, opened 2025-11-24
Implements rounding mode for NVFP4 tensor

### LINKED ISSUE #3264: [RFC] NVFP4 Rounding Modes
The current rounding mode for NVFP4 tensors in TorchAO is round-to-nearest. The purpose of this issue is to discuss support for other rounding modes.

### What rounding modes are available?

- Stochastic Rounding (RS)
- Round Nearest (RN)
- Round-zero (RZ)

### Where do we need different rounding modes?

- NVFP4 Training Recipe (https://github.com/pytorch/torchtitan/issues/1962)
  - RS for gradients
  - RN for weights and activation
- _AdamW in torchao.optim supports BF16 stochastic rounding: https://github.com/pytorch/ao/blob/1e473ed94caa060b3fbac96657030c44173759e0/torchao/optim/README.md?plain=1#L71-L77
- INT8 Quantization has stochastic rounding mode in TorchAO: https://github.com/pytorch/ao/blob/1e473ed94caa060b3fbac96657030c44173759e0/torchao/prototype/quantized_training/int8.py#L24-L26

### Existing RN Kernels

- Eager path: https://github.com/pytorch/ao/blob/1e473ed94caa060b3fbac96657030c44173759e0/torchao/prototype/custom_fp_utils.py#L27-L30
- torch.compile: https://github.com/pytorch/ao/blob/1e473ed94caa060b3fbac96657030c44173759e0/torchao/prototype/mx_formats/kernels.py#L1492-L1493
  - Uses `cvt.rn.satfinite.e2m1x2.f32 inline asm`

### Possible RS Kernels implementation
- Emulated implementation from @slayton58. I've quickly written this in triton syntax but it probably makes most sense to write in pytorch eager similar to RN (`_f32_to_floatx_unpacked`).
  ```python
  @triton.jit
  def float_rs(x, seed, offset):
      """
      Apply stochastic rounding when casting from float32 to NVFP4.
      
      Args:
          x: Input tensor (float32)
          seed: Random seed for the random number generator
          offset: Offset for random number generation (should be unique per element)
      
      Returns:
          Stochastically rounded tensor
      """
      
      # Scale down by 2^(-125) to normalize range
      downscale_factor = tl.math.exp2(-125.0)
      x = x * downscale_factor
      
      # Create 32-bit pseudorandom value
      rnd = tl.randint(seed, offset)
      
      # Isolate lower 22 bits for randomness injection
      # Process: left-shift by 10, then right-shift by 10
      rnd_shifted = (rnd << 10) >> 10
      
      # Reinterpret float bits as unsigned integer
      xb = x.to(tl.uint32, bitcast=True)
      
      # Inject randomness into the discarded precision bits
      yb = xb + rnd_shifted
      
      # Clear the lower 22 bits to perform rounding
      yb = (yb >> 22) << 22
      
      # Reinterpret integer bits back as floating point
      y = yb.to(tl.float32, bitcast=True)
      
      # Restore original magnitude by scaling up
      upscale_factor = tl.math.exp2(125.0)
      y = y * upscale_factor
      
      return y
  ```
- Use an inline asm triton kernel using `cvt.rs.satfinite.e2m1x4.f32` for stochastic rounding similar to RN.

### Integration
- A possible integration point for NVFP4 Training Recipe use case is to specify the rounding mode in `to_nvfp4` calls.
  ```python
  class RoundingMode(Enum):
      RN = "round_nearest"
      RS = "round_stochastic"
      RZ = "round_zero"
  
  def to_nvfp4(
          data_hp: torch.Tensor,
          block_size: int = 16,
          per_tensor_scale: Optional[torch.Tensor] = None,
          act_per_tensor_scale: Optional[torch.Tensor] = None,
          is_swizzled_scales: bool = False,
          use_triton_kernel: bool = False,
          act_quant_kwargs: Optional[QuantizeTensorToNVFP4Kwargs] = None,
          rounding_mode: RoundingMode = RoundingMode.RN,
      ):
      ...
      if use_triton_kernel:
          blockwise_scales, data_lp = triton_quantize_nvfp4(data_hp, per_tensor_scale, rounding_mode)
      else:
          blockwise_scales, data_lp = nvfp4_quantize(
              data_hp, block_size, per_tensor_scale, rounding_mode
          )
  ```
- We should discuss if we need to support rounding mode more generically to support other use cases like _AdamW, and int8 training.

### Test Plan
- TODO

CC: @slayton58, @ngimel, @ptrblck, @eqy

syed-ahmed:
Closes https://github.com/pytorch/ao/issues/3264. 
Code Assistant Used: Claude Code Opus 4.6
## Summary

Implements RS (stochastic rounding) support for NVFP4 as described in the RFC. RN remains the default.

## Key implementation details

**Triton kernel path**: Uses hardware `cvt.rs.satfinite.e2m1x4.f32` PTX inline asm. The RS instruction converts 4 floats and 1 random uint32 into 2 packed FP4 bytes, while the RN instruction converts 2 floats into 1 byte. Since the RN path uses `pack=4` (8 floats per invocation), the RS path issues two `cvt.rs` calls to match. With `pack=4`, 4 `rbits` values are loaded but only 2 are consumed per invocation — the other 2 are wasted to keep the output layout identical to RN.

**Seed determinism**: Triton path takes an explicit `seed` parameter for `tl.randint`; PyTorch path uses `torch.manual_seed`. Same seed produces bitwise-identical results within each path. The two paths use different RNGs so they diverge for RS even with the same seed.

**Validation**: Both paths validate `rounding_mode` using `not in RoundingMode`, raising `ValueError` for invalid values.

**`RoundingMode` enum uses `int` values** (`RN=0`, `RS=1`) rather than string values from the RFC, since the Triton kernel needs an integer `tl.constexpr`.

## Tests

Single parametrized `test_f4_rounding` in `test_kernels.py` covers rounding mode (RN/RS/invalid), kernel (PyTorch/Triton), seed determinism (same/different seeds), and value axes. Verifies RN is biased to nearest, RS is unbiased in expectation, same seed is deterministic, different seeds diverge, and invalid mode raises.

Existing `test_nvfp4_tensor.py` tests are parametrized over `rounding_mode` and `_triton_kernel_params`. Triton-vs-PyTorch equivalence uses SQNR threshold of 40 for RN and 8 for RS (different RNGs). End-to-end matmul uses SQNR threshold of 16 for RN and 6 for RS.

## Test Environment
```
Collecting environment information...                                                                                     
PyTorch version: 2.12.0a0+gitbabda95                                                                                      
Is debug build: False                                                                                                     
CUDA used to build PyTorch: 13.2                                                                                          
ROCM used to build PyTorch: N/A                                                                                           
                                                                                                                          
OS: Ubuntu 24.04.4 LTS (aarch64)                                                                                          
GCC version: (Ubuntu 13.3.0-6ubuntu2~24.04.1) 13.3.0                                                                      
Clang version: Could not collect                                                                                          
CMake version: version 3.31.6                                                                                             
Libc version: glibc-2.39                                                                                                  
                                                                                                                          
Python version: 3.12.3 (main, Jan 22 2026, 20:57:42) [GCC 13.3.0] (64-bit runtime)                                        
Python platform: Linux-6.14.0-1008-nvidia-64k-aarch64-with-glibc2.39                                                      
Is CUDA available: True                                                                                                   
CUDA runtime version: 13.2.50                                                                                             
CUDA_MODULE_LOADING set to: LAZY                                                                                          
GPU models and configuration:                                                                                             
GPU 0: NVIDIA GB200                                                                                                       
GPU 1: NVIDIA GB200                                                                                                       
GPU 2: NVIDIA GB200                                                                                                       
GPU 3: NVIDIA GB200

Nvidia driver version: 580.105.08                                                                                                                                                                                                                                                          

Versions of relevant libraries:                                                                                     
[pip3] mypy==1.16.0                                       
[pip3] mypy_extensions==1.1.0                             
[pip3] numpy==1.26.4                                      
[pip3] nvidia-cudnn-frontend==1.18.0                                                                                
[pip3] onnx==1.20.0                                       
[pip3] onnx-ir==0.1.16                                    
[pip3] onnxscript==0.6.2                                  
[pip3] optree==0.13.0                                     
[pip3] pytorch-lightning==2.6.1                                                                                     
[pip3] torch==2.12.0a0+gitbabda95                                                                                   
[pip3] torchmetrics==1.8.2                                
[pip3] triton==3.6.0+git9844da95                                                                                    
[conda] Could not collect
```

### DIFF
--- a/torchao/prototype/custom_fp_utils.py
+++ b/torchao/prototype/custom_fp_utils.py
@@ -13,10 +13,10 @@
 #      - Values outside the representable range of Floatx after rounding are clamped to the maximum Floatx
 #      magnitude (sign is preserved).
 
-import torch
 from enum import Enum
-from torch import Tensor
 
+import torch
+from torch import Tensor
 
 
 class RoundingMode(Enum):
@@ -49,6 +49,7 @@ def _f32_to_floatx_unpacked(
     ebits: int,
     mbits: int,
     rounding_mode: RoundingMode = RoundingMode.RN,
+    rand_bits: Tensor | None = None,
 ) -> Tensor:
     """Convert FP32 numbers to sub-byte floating point numbers with the given
     number of exponent and mantissa bits.
@@ -65,8 +66,13 @@ def _f32_to_floatx_unpacked(
         mbits: Number of mantissa bits in the target format (including the
             implicit leading 1).
         rounding_mode: RoundingMode.RN (round to nearest, ties to even)
-            or RoundingMode.RS (stochastic rounding). For deterministic RS,
-            set torch.manual_seed() before calling.
+            or RoundingMode.RS (stochastic rounding).
+        rand_bits: Random int32 tensor for stochastic rounding. Required when
+            rounding_mode is RS. Must have the same shape as the normal-range
+            elements of x. Only the low-order (23 - mbits) bits are used;
+            the caller may pass any int32 random values. The caller is
+            responsible for generating this tensor, which keeps RNG concerns
+            (seeding, CUDA graph safety) outside this function.
 
     Note: there are no special values (NaN, inf) support in this code. Values
     outside the representable range of Floatx after rounding are clamped to the
@@ -160,13 +166,13 @@ def _f32_to_floatx_unpacked(
         # positions. This makes the probability of rounding up proportional
         # to the fractional distance from the lower quantization level,
         # providing an unbiased estimator.
-        rand_bits = torch.randint(
-            0,
-            (1 << (MBITS_F32 - mbits)),
-            normal_x.shape,
-            dtype=torch.int32,
-            device=normal_x.device,
-        )
+        if rand_bits is None:
+            raise ValueError(
+                "rand_bits is required for stochastic rounding (RoundingMode.RS). "
+                "Caller must generate rand_bits externally."
+            )
+        # Mask to only the bits that fit in the truncated mantissa positions.
+        rand_bits = rand_bits & ((1 << (MBITS_F32 - mbits)) - 1)
         val_to_add = (exp_bias - F32_EXP_BIAS) << MBITS_F32
         normal_x += val_to_add
         normal_x += rand_bits
--- a/torchao/prototype/custom_fp_utils.py
+++ b/torchao/prototype/custom_fp_utils.py
@@ -13,7 +13,7 @@
 #      - Values outside the representable range of Floatx after rounding are clamped to the maximum Floatx
 #      magnitude (sign is preserved).
 
-from enum import Enum
+from enum import Enum, auto
 
 import torch
 from torch import Tensor
--- a/torchao/prototype/custom_fp_utils.py
+++ b/torchao/prototype/custom_fp_utils.py
@@ -8,12 +8,32 @@
 # It has been refactored to support any sub-byte FP dtypes. However, some behaviors of MX dtypes remain:
 #   1. No encodings are reserved for special values (+/-inf, NaN).
 #   2. When downcasting from FP32 to Floatx,
-#      - Rounding mode is round to nearest, ties to even.
+#      - Rounding mode defaults to round to nearest, ties to even.
+#      - Stochastic rounding is also supported for training use cases.
 #      - Values outside the representable range of Floatx after rounding are clamped to the maximum Floatx
 #      magnitude (sign is preserved).
 
 import torch
+from enum import Enum
 from torch import Tensor
+
+
+
+class RoundingMode(Enum):
+    """
+    Enum representing rounding modes for sub-byte floating point quantization.
+
+    RN: Round to nearest, ties to even. This is the standard IEEE 754 default
+        rounding mode. Used for weights and activations.
+
+    RS: Stochastic rounding. Adds uniformly distributed random noise to the
+        mantissa bits below the target precision before truncating. Provides
+        an unbiased estimator of the original value in expectation. Needed
+        for gradient quantization in NVFP4 training.
+    """
+
+    RN = 0
+    RS = 1
 
 
 def _n_ones(n: int) -> int:
@@ -24,7 +44,12 @@ EBITS_F32, MBITS_F32 = 8, 23
 F32_EXP_BIAS = _n_ones(EBITS_F32 - 1)
 
 
-def _f32_to_floatx_unpacked(x: Tensor, ebits: int, mbits: int) -> Tensor:
+def _f32_to_floatx_unpacked(
+    x: Tensor,
+    ebits: int,
+    mbits: int,
+    rounding_mode: RoundingMode = RoundingMode.RN,
+) -> Tensor:
     """Convert FP32 numbers to sub-byte floating point numbers with the given
     number of exponent and mantissa bits.
 
@@ -33,6 +58,15 @@ def _f32_to_floatx_unpacked(x: Tensor, ebits: int, mbits: int) -> Tensor:
     in the least significant bits. e.g.
       fp4: bits 0-3 empty and bits 4-7 in fp4_e2m1 encoding
       fp6: bits 0-1 empty and bits 2-7 in fp6_e2m3 or fp6_e3m2 encoding
+
+    Args:
+        x: Input tensor of dtype torch.float.
+        ebits: Number of exponent bits in the target format.
+        mbits: Number of mantissa bits in the target format (including the
+            implicit leading 1).
+        rounding_mode: RoundingMode.RN (round to nearest, ties to even)
+            or RoundingMode.RS (stochastic rounding). For deterministic RS,
+            set torch.manual_seed() before calling.
 
     Note: there are no special values (NaN, inf) support in this code. Values
     outside the representable range of Floatx after rounding are clamped to the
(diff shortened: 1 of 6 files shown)


drisspg (commented):
We should have the lower level ops all be functional, e.g. take in a seed anlgous to what is done for the tritonkernel, but same for the _f32_to_floatx_unpacked


And then for how users use this e.g.:             _seed = torch.randint(2**31, (1,)).item()

in the nvfp4 tensors I am not entirely sure I like this..

syed-ahmed:
I think that should be fine, and I can make the seed handling more functional. As a side note, may be need to review TE's rng_state mechanism a little bit and make sure we have full control over the determinism knobs.

drisspg:
TBH this PR has sparked some convo on how we expect users to handle RNG for RS mode.

I am not sure I have perosnally have an infromed enough opinion on the difficulties and nuance here. 

maybe @vkuzo has some better ideas here

My intial thought is lower level is functional. And then user sets a seed (philox state),  ensure that we properly increment philox offset before every invocation of RS which will pull from there to generate the source of randomness. Also ensure that cudagraphs work here. I bet inductor will not behave well here though.

@jbschlosser  just to throw some more ideas into the ring

jbschlosser:
I'm currently looking into the API design for a new set of "stateless RNG" APIs (i.e. pseudorandom number generation without pytorch maintaining global PRNG state). My concern is that this composes well with RS mode as implemented in this PR, and, more generally, with any implementations of stochastic rounding throughout torch / torchao.

> My intial thought is lower level is functional. And then user sets a seed (philox state), ensure that we properly increment philox offset before every invocation of RS which will pull from there to generate the source of randomness.

so I generally agree with Driss's statement here. I think I'd want to see the lower level ops accept a philox seed / offset and that should satisfy composability with a new stateless RNG API.

syed-ahmed:
Thanks @drisspg and @jbschlosser for the input! I'll refactor the ops to be functional and only take seed, and remove the .item(). For the eager path, should `_f32_to_floatx_unpacked` take a seed (and so consume the seed in a torch.Generator, and that Generator gets passed to rand_ints) or should we just have a rand_int tensor as input?

syed-ahmed:
Ok, addressed the comments. 
- .item() call is removed. Triton kernel now loads a seed.
- eager path expects a rand_bits tensor now keeping generator state outside.
- to_nvfp4 expects caller to pass rand_bits.
- added a CUDA graph composabilty test for both eager and torch.compile.

torchao/prototype/custom_fp_utils.py
```diff
@@ -8,14 +8,34 @@
 # It has been refactored to support any sub-byte FP dtypes. However, some behaviors of MX dtypes remain:
 #   1. No encodings are reserved for special values (+/-inf, NaN).
 #   2. When downcasting from FP32 to Floatx,
-#      - Rounding mode is round to nearest, ties to even.
+#      - Rounding mode defaults to round to nearest, ties to even.
+#      - Stochastic rounding is also supported for training use cases.
 #      - Values outside the representable range of Floatx after rounding are clamped to the maximum Floatx
 #      magnitude (sign is preserved).
 
+from enum import Enum
```
crcrpar:
```suggestion
from enum import Enum, auto
```

torchao/prototype/custom_fp_utils.py
```diff
@@ -8,14 +8,34 @@
...
+        rounding mode. Used for weights and activations.
+
+    RS: Stochastic rounding. Adds uniformly distributed random noise to the
+        mantissa bits below the target precision before truncating. Provides
+        an unbiased estimator of the original value in expectation. Needed
+        for gradient quantization in NVFP4 training.
+    """
+
+    RN = 0
+    RS = 1
```
crcrpar:
```suggestion
    RN = auto()
    RS = auto()
```

torchao/prototype/custom_fp_utils.py
crcrpar, line 1: purely cosmetic but `auto` can be used while it'd start with 1, not 0 -- https://docs.python.org/3/library/enum.html#enum.auto

torchao/prototype/mx_formats/kernels.py
```diff
@@ -53,13 +54,23 @@ def get_bits(x: torch.Tensor) -> str:
 ZERO_POINT_FIVE_BITS_F32 = 0x3F000000
 
 
-def f32_to_f4_unpacked(x):
+def f32_to_f4_unpacked(x, rounding_mode=RoundingMode.RN, rand_bits=None):
```
crcrpar, line 61: i feel a type annotation is good to have for this method

```python
def f32_to_f4_unpacked(x: torch.Tensor, rounding_mode: RoundingMode = RoundingMode.RN, rand_bits: torch.Tensor | None = None):
```

torchao/prototype/mx_formats/kernels.py
```diff
@@ -628,7 +708,7 @@ def triton_quantize_nvfp4(
         return scales, xq.view(torch.uint8)
 
     @triton_quantize_nvfp4.register_fake
-    def _(x, per_tensor_scale=None):
+    def _(x, per_tensor_scale=None, rounding_mode=0, seed=None):
```
crcrpar: I'm not quite confident but shouldn't this be?
```suggestion
    def _(x, per_tensor_scale=None, rounding_mode=RoundingMode.RN, seed=None):
```

torchao/prototype/mx_formats/kernels.py
```diff
@@ -664,7 +743,9 @@ def triton_mx_block_rearrange(scale_tensor: torch.Tensor) -> torch.Tensor:
         raise AssertionError("needs torch version 2.8+ and triton")
 
     def triton_quantize_nvfp4(
-        x: torch.Tensor, tensor_scale: Optional[torch.Tensor] = None
+        x: torch.Tensor,
+        tensor_scale: Optional[torch.Tensor] = None,
+        rounding_mode: int = 0,
```
crcrpar: ditto
```suggestion
        rounding_mode: RoundingMode = RoundingMode.RN,
```

torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -670,6 +702,10 @@ def nvfp4_quantize(
     assert data_hp.size(-1) % block_size == 0, "K dim must be divisible by block_size"
     assert data_hp.is_contiguous(), "Only support contiguous data for now"
     assert block_size == 16, "NVFP4 requires block_size=16"
+    if rounding_mode == RoundingMode.RS:
+        assert rand_bits is not None and rand_bits.numel() > 1, (
```
drisspg, line 839: this numel check si werid it should just do a direct shape compare right?

drisspg: hmm I guess it needs to have two differtn forms 1 for triton and one for fallback?

syed-ahmed: Yes. For triton, I only need  a fresh seed which informs the philox in triton where to start from and the offsets ensure getting the correct random number. For the fallback, we are generating all the randoms outside the quantize.

torchao/prototype/mx_formats/kernels.py
```diff
@@ -425,6 +436,295 @@ def triton_mx_block_rearrange(scale_tensor: torch.Tensor) -> torch.Tensor:
...
+            + (pid_m * tl.num_programs(0) + pid_n) * (32 * 16)
+            + offs_m * 16
+            + offs_n,
+            packed_scales,
+        )
+
+        # Output offsets for packed FP4 storage [128, 32]
+        offs_m = pid_m * 128 + tl.arange(0, 128)[:, None]
+        offs_n = pid_n * 32 + tl.arange(0, 32)[None, :]
+        out_offs = offs_m * (N // 2) + offs_n
```
jbschlosser, line 610: I'm worried about how these offsets are calculated wrt randomness below during stochastic rounding application. Passing a seed only to this kernel does not seem like enough to avoid undesirable RNG reuse. I think at the very least, we'd want to accept a base offset to start from. And even more ideally from a control perspective, we'd allow for a full set of (seed, offset) pairs controlling every single application of rounding (but I realize there will likely be perf issues with this).

syed-ahmed: Thanks @jbschlosser for the comment. 

Would it be better to write a fused kernel for now, similar to MSLK: https://github.com/meta-pytorch/MSLK/blob/main/mslk/quantize/triton/fp4_quantize.py#L723 when composing with random hadamard transform (https://github.com/pytorch/ao/issues/4040)?

It seems like there is some discussion on how to support stochastic rounding better here: https://github.com/pytorch/pytorch/issues/175409, so not sure anymore if the approach in this PR would make sense for UX and performance.

syed-ahmed: @rdspring1 added an offset base now:
```
offset_base = tl.load(offset_ptr)
# Place local index in upper 32 bits and offset_base in lower 32 bits
offset = (tl.cast(out_offs, tl.int64) << 32) | offset_base
```

torchao/prototype/mx_formats/__init__.py
```diff
@@ -16,4 +17,5 @@
     "NVFP4DynamicActivationNVFP4WeightConfig",
     "NVFP4ObservedLinear",
     "NVFP4WeightOnlyConfig",
+    "RoundingMode",
```
jerryzh168, line 20: nit: probably make sense to qualify this a bit more, e.g. `NVFP4RoudningMode`?

PR #3388, opened 2025-11-25
[mxfp8 moe training] add roofline script

danielvegamyhre:
## Summary
This gives us complete observability into the `_to_mxfp8_then_scaled_grouped_mm` autograd func, generating top level roofline against bf16 as well for each individual kernel roofline vs actual perf.

includes:
- fwd and bwd
- configurable power limit, N, K, G dims
- plots for roofline vs actual for e2e op fwd + bwd and individual kernels

Results indicate the perf optimization opportunities, in order of impact, are:
1. 2d-2d mxfp8 grouped gemm underperforming (by far the biggest impact)
2. blocked format scale conversion kernels running at low mem bw utilization

Note the scale conversion to blocked format kernels are terribly inefficient, but the actual execution time is so small that it does not move the needle for net speedup. We should improve these, but they are a lower priority than improving the GEMMs. 

Llama4:

llama_rooflines

DSV3 671b:
dsv3_rooflines

### DIFF
--- a/benchmarks/prototype/moe_training/mxfp8/.gitignore
+++ b/benchmarks/prototype/moe_training/mxfp8/.gitignore
@@ -1,0 +1,5 @@
+# Ignore CSV output files
+*.csv
+
+# Ignore PNG plot files
+*.png
--- a/benchmarks/prototype/moe_training/mxfp8/roofline_unified.py
+++ b/benchmarks/prototype/moe_training/mxfp8/roofline_unified.py
@@ -1,0 +1,1737 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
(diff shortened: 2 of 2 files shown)


vkuzo:
how about making https://github.com/pytorch/ao/blob/main/benchmarks/float8/float8_roofline.py include MoE instead?

danielvegamyhre:
> how about making https://github.com/pytorch/ao/blob/main/benchmarks/float8/float8_roofline.py include MoE instead?

I'm trying to do something a bit different actually. Basically complete observability for the entire _to_mxfp8_then_scaled_grouped_mm op, with rooflines vs actual for both the end-to-end up as well as the individual kernels, all in one centralized view. To help us quickly diagnose performance issues. Check out the updated PR description and plots I've added.

danielvegamyhre:
gpu_name_to_specs is great, thanks! updated accordingly.

will discuss 2nd item with you today

PR #3401, opened 2025-11-27
Fixing subclassing

Krishn1412:
- Fix subclass initialization by ensuring each subclass gets its own per-class dispatch tables and correctly inherits parent op mappings without recursion.
- Look at [PR](https://github.com/pytorch/ao/pull/3153).

### DIFF
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -413,6 +413,50 @@ class TestTorchAOBaseTensor(unittest.TestCase):
 
         self.assertEqual(counter["calls"], 1)
 
+    def test_op_overwrite(self):
+        counters = {"parent": 0, "child": 0}
+
+        class Parent(TorchAOBaseTensor):
+            tensor_data_names = ["qdata"]
+            tensor_attribute_names = ["attr"]
+
+            def __new__(cls, qdata, attr):
+                r = torch.Tensor._make_wrapper_subclass(cls, qdata.shape)
+                r.qdata = qdata
+                r.attr = attr
+                return r
+
+            def __init__(self, qdata, attr):
+                pass
+
+        @Parent.implements([torch.ops.aten.cat.default])
+        def parent_cat(func, types, args, kwargs):
+            counters["parent"] += 1
+
+        class Child(Parent):
+            tensor_data_names = ["qdata"]
+            tensor_attribute_names = ["attr"]
+
+        @Child.implements([torch.ops.aten.cat.default])
+        def child_cat(func, types, args, kwargs):
+            counters["child"] += 1
+
+        # Parent instance should call parent implementation
+        p1 = Parent(torch.randn(2, 3), "a")
+        p2 = Parent(torch.randn(2, 3), "b")
+        torch.ops.aten.cat.default([p1, p2], 0)
+
+        self.assertEqual(counters["parent"], 1)
+        self.assertEqual(counters["child"], 0)
+
+        # Child instance should call child implementation
+        c1 = Child(torch.randn(2, 3), "a")
+        c2 = Child(torch.randn(2, 3), "b")
+        torch.ops.aten.cat.default([c1, c2], 0)
+
+        self.assertEqual(counters["parent"], 1)
+        self.assertEqual(counters["child"], 1)
+
     def test_multiple_inheritance(self):
         counters = {"A": 0, "B": 0}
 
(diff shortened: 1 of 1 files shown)


Krishn1412:
Can you take a look @jerryzh168?

Krishn1412:
Hello @jerryzh168, happy new year! Can you please take a look at this

test/test_utils.py
```diff
@@ -342,6 +342,91 @@ def fake_linear(func, types, args, kwargs):
...
+        class Child(Parent):
+            tensor_data_names = ["qdata"]
+            tensor_attribute_names = ["attr"]
+
+        # Table checks
+        self.assertIn(torch.ops.aten.cat.default, Parent._ATEN_OP_TABLE[Parent])
+        self.assertIn(torch.ops.aten.cat.default, Child._ATEN_OP_TABLE[Child])
+
+        # Ensure child table is distinct
+        self.assertIsNot(Parent._ATEN_OP_TABLE, Child._ATEN_OP_TABLE)
```
jerryzh168: nit: we can have Child reimplement `torch.ops.aten.cat.default` and make sure the correct op is invoked

test/test_utils.py
```diff
@@ -342,6 +342,91 @@ def fake_linear(func, types, args, kwargs):
             counter["calls"], 2, "Expected fake_linear to be called via aten.t.default"
         )
 
+    def test_subclassing(self):
+        class Parent(TorchAOBaseTensor):
+            tensor_data_names = ["qdata"]
+            tensor_attribute_names = ["attr"]
+
+        Parent._ATEN_OP_TABLE[Parent]["op_parent"] = "parent_impl"
+        Parent._TORCH_FN_TABLE[Parent]["fn_parent"] = "parent_fn_impl"
```
jerryzh168: `_ATEN_OP_TABLE` and `_TORCH_FN_TABLE` are implementation details, I feel we should not rely on these for testing, I think it's better to just use parent / child counters to make sure the correct ops are called and overwriting happens

same for all the other tests, I think we can rely on counter in op implementations (user code) instead of the implementation details

jerryzh168:
sorry for the delay and thanks for working on this, I just left some comments about testing

jerryzh168:
@Krishn1412 do you plan to update the PR? I think this is an important item to fix

Krishn1412:
Hello @jerryzh168, I will update it over the weekend. Sorry about the delay.

test/test_utils.py
```diff
@@ -342,6 +342,124 @@ def fake_linear(func, types, args, kwargs):
...
+            tensor_attribute_names = ["attr"]
+
+        t1 = Child(torch.randn(2, 3), "a")
+        t2 = Child(torch.randn(2, 3), "b")
+
+        torch.ops.aten.cat.default([t1, t2], 0)
+
+        self.assertEqual(counter["calls"], 1)
+
+    def test_multiple_inheritance(self):
```
jerryzh168, line 460: I think before the more complicated case, we can add one more test here, for single inheritance, and Child overwrite the op that's already defined by Parent (e.g. cat), and when we call into the implementation from Child class instead of Parent class, (name can be: test_op_overwrite), also verify that parent subclass instance is still calling the parent implementation instead of the child implementation as well

test/test_utils.py
```diff
@@ -342,6 +342,124 @@ def fake_linear(func, types, args, kwargs):
...
+
+        class C(A, B):
+            tensor_data_names = ["a"]
+            tensor_attribute_names = ["b"]
+
+        t = C(torch.randn(3), "x")
+        torch.ops.aten.neg.default(t)
+
+        self.assertEqual(counters["A"], 0)
+        self.assertEqual(counters["B"], 1)
```
jerryzh168, line 505: can add additional test here for C to overwrite the neg function, and then only counters["C"] is updated as well

jerryzh168: also check the behavior of instance of A and B when we call the neg op when a new neg op is implemented in C (similar to the single inheritance example)

Krishn1412:
hey @jerryzh168, is it the last test what you want in you second comment?

jerryzh168:
> hey @jerryzh168, is it the last test what you want in you second comment?

yeah that looks good, I was thinking of just adding the test to the multiple inheritance test, but I think it's better to keep this in a separate test

torchao/utils.py
```diff
@@ -880,10 +880,14 @@ def __init_subclass__(cls, **kwargs):
         # inherit the torch function and dispatch implementations from direct parent classes
         # e.g. for `class C(B, A)`, C.__bases__ == (B, A)
```
jerryzh168, line 881: maybe before landing, probably better to add an example of what `_ATEN_OP_TABLE` and `_TORCH_FN_TABLE` looks like for parent and current `cls` (child) so it's easier to understand for future devs

jerryzh168: since CI is passing already, I can follow up with the docs, thanks for the contribution!

Issue #3290, opened 2025-11-04
FP8 Blockwise Training Tracker


danielvegamyhre:
We want to support DeepSeekV3-style FP8 blockwise training in torchao for both dense and MoE models. 

# Support for dense models (linears)
- [x] Functionality
    - [x] 1x128 quantization for LHS activations, write to row major layout
    - [x] 128x1 quantization for RHS activations, write to col major layout
    - [x] 128x128  quantization for weights, write to col major layout
    - [x] 1x128 @ 128x128 gemm, use for:
         - [x] `output = input @ weight.t()`
         - [x] `dgrad = grad_output @ weight`
    - [x] 1x128 @ 128x1 gemm, use for:
         - [x]  `wgrad = grad_output.t() @ input`
    - [x] Autograd function implementing forward and backward
    - [ ] DTensor handling for TP support
    - [ ] Custom ops around all custom kernels for `torch.compile` composability
    - [ ] Tests for FSDP, TP
    - [x] `quantize_` model conversion api peforming module swap of nn.Linear to FP8BlockwiseLinear (wraps autograd func)
- [ ] Performance
    - [ ] all quantization kernels run at 80%+ of peak achievable memory bandwidth on Hopper
         - [ ] benchmark scripts for each quantization kernel 
    - [ ] all gemm kernels run at 80%+ of peak achievable TFLOPs/sec on Hopper 
         - [ ] benchmark scripts for each gemm
- [ ] Integration into torchtitan, validate with Llama3
 
# Support for MoE layers (grouped GEMMs)
- [ ] Functionality
    - [ ] 128x128 quantization compatible with 3d expert weights, write to per-expert col major layout (e.g. shape (E,N,K) with strides (N*K,1,N)) 
    - [ ] 1x128 @ 128x128 scaled grouped gemm
         - [ ] `output = input @ weight.transpose(-2,-1)`
         - [ ] `dgrad = grad_output @ weight`
    - [ ] 1x128 @ 128x1 scaled grouped gemm
         - [ ] `wgrad = grad_output.transpose(-2,-1) @ input`
- [ ] Performance
    - [ ] all quantization kernels run at 80%+ of peak achievable memory bandwidth on Hopper
         - [ ] benchmark scripts for each quantization kernel 
    - [ ] all gemm kernels run at 80%+ of peak achievable TFLOPs/sec on Hopper 
         - [ ] benchmark scripts for each gemm
[...]

PR #3402, opened 2025-11-28
FP8 Blockwise Training: `triton_op` for dense model

agolajko:
Added two `triton_op`s to the GEMM kernels

re: #3290
cc @danielvegamyhre

### DIFF
--- a/torchao/prototype/blockwise_fp8_training/kernels.py
+++ b/torchao/prototype/blockwise_fp8_training/kernels.py
@@ -96,6 +96,7 @@ def triton_fp8_gemm_1x128_128x128_kernel(
     tl.store(c_ptrs, c, mask=c_mask)
 
 
+@triton_op("torchao::triton_fp8_gemm_1x128_128x128", mutates_args={})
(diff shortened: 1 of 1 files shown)


danielvegamyhre:
this is fine to do, however, it's primarily the quantization kernels we need custom ops for, since for the GEMMs we're going to use torch._scaled_mm which dispatches to a cutlass kernel that is more performant than these triton ones

danielvegamyhre (approved):

agolajko:
By quantization kernels you mean the eg [this](https://github.com/pytorch/ao/blob/main/torchao/prototype/blockwise_fp8_training/kernels.py#L337)?

All of these are decorated with @triton_op already

What am I missing?

danielvegamyhre:
> By quantization kernels you mean the eg [this](https://github.com/pytorch/ao/blob/main/torchao/prototype/blockwise_fp8_training/kernels.py#L337)?
> 
> All of these are decorated with @triton_op already
> 
> What am I missing?

ah i forgot we already did that, been a while since i looked at this code. thanks!

PR #3156, opened 2025-10-11
Introduce SINQ calibration-free quantization algorithm

namgyu-youn:
**Summary**:
Introduce SINQ: Sinkhorn-Normalized Quantization for calibration-free weight quantization (https://arxiv.org/abs/2509.22944).

SINQ uses **dual-axis scaling** (row + column) vs. HQQ's single-axis approach, achieving 1) 2-3x faster quantization time and 2) better imbalance handling.

<details>
<summary>(TL;DR) What is SINQ?</summary>

### Single-scale (Scales + Shifts)
In normally, weight-only quantization algorithms defined as

$\hat{W} = \vec{s} \odot(Q +\vec{z})$

where `w_hat` is N × M matrix, `\vec{s}` is a N × 1 scale factor, `Q` is a quantized N × M matrix, and `\vec{z}` is a shift.

### Dual-Scacles (SINQ)
Unlike above scales+shift approach, SINQ supply two vectors,

$\hat{W} = \vec{s} \odot Q \odot \vec{t}$

where `\vec{s}` is a N × 1 vector, `\vec{t}` is a 1 × M vector and the rest is as above.

2-axis scale factor efficiently collects spatial outlier distribution.

image

> If we don't mind the potential additonal overhead, `Q` can be updated to `Q+\vec{z}` with shifting

## Representation Space (Matrix Imbalance)
Matrix imbalance (i.e., outlier) is inconvenient to optimize with gradient descent, because of sparse gradients interrupt maximum and minimum operations. HIGGS used rotations (hadamard transform to normalize weight distribution), and AWQ/SmoothQuant used channel-wise scaling to minimizing errors by outliers.

SINQ uses sinkhorn iteration to **normalize both row/column std** (Algorithm 1).

## SINQ: Algorithm 1
Iteratively normalize the standard deviation of the rows and columns of the matrix (weight) to be quantized. Then apply a standard quantization method (e.g., RTN)

image

</details>

**Test/Future Plan:**
The commented test shows quantization quality and speed comparison with HQQ. Full TorchAO integration with `IntxWeightOnlyConfig` and `Int8DynamicActivationIntxWeightConfig` is planned.

**Related Issue/PR**: [#3106](https://github.com/pytorch/ao/pull/3106#issuecomment-3382559912)

PR #3106, opened 2025-10-01
Add outlier in AWQ test cases

namgyu-youn:
**Summary**:
Similar to #3101 , this test inject outliers to reflect input distribution. Toy model is updated to use direct dtype, device callsite

**Test plan**:
pytest -sv test/prototype/test_awq.py

PR #3403, opened 2025-11-30
Add unit test for HQQ

namgyu-youn:
**Summary**:
Add HQQ unit test because there was only an e2e test.

**Test Plan**:
```
pytest -sv test/quantization/test_quant_primitives.py
```

**Related Issue/PR**: https://github.com/pytorch/ao/pull/3156#issuecomment-3554170604

### DIFF
--- a/test/quantization/test_quant_primitives.py
+++ b/test/quantization/test_quant_primitives.py
@@ -30,6 +30,7 @@ from torchao.quantization.quant_primitives import (
 # TODO: remove test for utils?
 from torchao.quantization.utils import (
     _quantize_activation_per_token_absmax,
+    compute_error,
     get_block_size,
     get_group_qparams_symmetric,
     groupwise_affine_dequantize_tensor_from_qparams,
@@ -896,6 +897,9 @@ class TestQuantPrimitives(unittest.TestCase):
         reconstructed = qdata.to(input.dtype) * scale_expanded
         self.assertFalse(torch.isnan(reconstructed).any())
         self.assertEqual(reconstructed.shape, input.shape)
+
+        error = compute_error(input, reconstructed)
+        self.assertLess(error, 25)
 
     def test_float8_blockwise_scaling(self):
         M, K = 512, 1024


torchao/_models/_eval.py
```diff
@@ -106,7 +106,7 @@ def max_length(self):
 
     @property
     def max_gen_toks(self):
-        return 50
+        return 512
```
namgyu-youn: This is updated because the GSM8K task is CoT reasoning (long-context)

namgyu-youn:
cc @metascroy can you take a look into this? I will move to E2E support for SINQ after this PR as we talked with https://github.com/pytorch/ao/pull/3156#issuecomment-3576954522; we can add SINQ to this e2e benchmark I feel.

metascroy:
Thanks!  The unit test looks good to me.  I'll let @jainapurva review the benchmark script, as I'm less familiar with that part of the codebase and if there are coding standards that are usually followed.

jerryzh168:
intx is targeting mobile right? so testing performance on server doesn't seem to be helpful?

namgyu-youn:
> intx is targeting mobile right? so testing performance on server doesn't seem to be helpful?

@jerryzh168 Sorry I misunderstood the difference between intx (ExecuTorch/mobile) and int4 (CUDA), and actually my use case is CUDA (not mobile) deployment. Working more for int4 should be better to me, so I will move to SINQ support for int4 (as talked with https://github.com/pytorch/ao/pull/3106#issuecomment-3382559912).

namgyu-youn:
> Thanks! The unit test looks good to me. I'll let @jainapurva review the benchmark script, as I'm less familiar with that part of the codebase and if there are coding standards that are usually followed.

@metascroy Updated PR for only adding the HQQ unit test (drop intx benchmark). Please check the above comment.

Let me know if a dropped benchmark would be needed for ExecuTorch; if needed, I will submit a PR for ExecuTorch.

metascroy:
> > Thanks! The unit test looks good to me. I'll let @jainapurva review the benchmark script, as I'm less familiar with that part of the codebase and if there are coding standards that are usually followed.
> 
> @metascroy Updated PR for only adding the HQQ unit test (drop intx benchmark). Please check the above comment.
> 
> Let me know if a dropped benchmark would be needed for ExecuTorch; if needed, I will submit a PR for ExecuTorch.

The benchmark isn't necessary for ExecuTorch.

metascroy:
PR looks good!  If unit tests pass, we can merge.

test/quantization/test_quant_primitives.py
```diff
@@ -863,6 +864,39 @@ def test_choose_qparams_and_quantize_scale_only_sinq(self):
...
+
+        qdata, scale = _choose_qparams_and_quantize_scale_only_hqq(
+            input,
+            block_size=block_size,
+            qmin=qmin,
+            qmax=qmax,
+            iters=20,
+        )
+
+        # Check quantized data shape and dtype
```
jerryzh168, line 884: test doesn't look very hard to pass? would it be better to make it stronger by comparing with the results of https://github.com/pytorch/ao/blob/8b2367bda34af74d739bc7ffda06942a7c069763/torchao/quantization/quant_primitives.py#L34?

test/quantization/test_quant_primitives.py
```diff
@@ -863,6 +864,39 @@ def test_choose_qparams_and_quantize_scale_only_sinq(self):
...
+        num_groups = input.shape[1] // block_size[1]
+        self.assertEqual(scale.shape, (input.shape[0], num_groups))
+        self.assertEqual(scale.dtype, input.dtype)
+        self.assertTrue((scale > 0).all())
+
+        # Test reconstruction is possible
+        scale_expanded = scale.repeat_interleave(block_size[1], dim=1)
+        reconstructed = qdata.to(input.dtype) * scale_expanded
+        self.assertFalse(torch.isnan(reconstructed).any())
+        self.assertEqual(reconstructed.shape, input.shape)
```
jerryzh168, line 899: also here can we compare `reconstructed` with original `input` (with sqnr) to make sure they are close?

test/quantization/test_quant_primitives.py
namgyu-youn: How about adding intx+hqq to
https://github.com/pytorch/ao/blob/8b2367bda34af74d739bc7ffda06942a7c069763/benchmarks/quantization/create_quantized_model.py#L30-L35

instead of comparing them here? Both are different HQQ implementation (optmized for NVIDIA vs. no HW optimization), so a sanity check might not be fair here. If we need to compare them, we can do some experiments with the new eval script I feel.

test/quantization/test_quant_primitives.py
namgyu-youn: Updated to compare reconstruction error, thanks.

namgyu-youn:
@metascroy @jerryzh168 could you help this to merge?

PR #3404, opened 2025-11-30
Introduce Stochastic Rounding

namgyu-youn:
**Summary**:
Stochastic rounding is widely used primitive for low-precision training task.

Unlike deterministic rounding, stochastic rounding is unbiased (`E[SR(x)] = x`), which helps prevent error accumulation in gradient updates and low-bit training.

**Test Plan**:
```
pytest -sv test/quantization/test_quant_primitives.py
pytest -sv test/prototype/test_quantized_training.py
```

### DIFF
--- a/test/quantization/test_quant_primitives.py
+++ b/test/quantization/test_quant_primitives.py
@@ -9,12 +9,6 @@
 import unittest
 
 import torch
-from torch.testing._internal.common_utils import (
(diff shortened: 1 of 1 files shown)


namgyu-youn:
Will reland this after better understanding. This PR approach is really thin and not appliable to quantized training

PR #3391, opened 2025-11-26
Introduce int8 quantization api (version 2)

namgyu-youn:
**Summary**:
Introduce a new tensor subclass API. The main features are
- `Int8Tensor`: Main API, which handles quantization and dequantization operations
- Utility operation functions: Tensor slice, index selection

This API is integrated into global variants (`Int8WeightOnlyConfig`, `Int8DynamicActivationInt8WeightConfig`) using `version`, and not defined as a default.

**Related Issue/PR**: #3241 (reland)

**Test plan**: pytest -sv test/quantization/quantize_/workflows/int8/test_int8_tensor.py

**PERF Test**:
https://github.com/pytorch/ao/blob/main/tutorials/quantize_vit/run_vit_b_quant.py with a batch size of 32:

| API | With `torch.compile` | Without `torch.compile` |
|-----|---------------------|------------------------|
| Old | 65.47 ms | 234.39 ms |
| New | 63.30 ms | 239.30 ms |

**Future Plan**: https://github.com/pytorch/ao/pull/3241#pullrequestreview-3502633317

PR #3407, opened 2025-12-01
Int8Tensor migration cleanup

jcaip:
Summary:

This PR cleans up several things in `Int8Tensor`:
- adds support for act_mapping_type (SYMMETRIC vs ASYMMETRIC) for parity with v1 config
- changes scales logic to be `keepdim`, previous scales for a 2d weight tensor was a 1d vector, now for a M x N weight is it is a M x 1 vector. This is the same as float8tensor, so we can reuse the same utility functions now. 
- adds test for different combinations of `granularity` and `act_mapping_type` currently only symmetric activation quantization is supported. 
- updated `_choose_quant_func_and_quantize_tensor` with int8 quantize kwargs. 

Test Plan:

To ensure BC:
```
pytest test/quantization/test_quant_api.py
```

To test Int8Tensor:
```
pytest test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/quantization/quantize_/workflows/int8/test_int8_tensor.py
+++ b/test/quantization/quantize_/workflows/int8/test_int8_tensor.py
@@ -109,6 +109,11 @@ class TestInt8Tensor(TorchAOIntegrationTestCase):
         self.assertEqual(model_q.linear2.weight.scale.ndim, 2)
 
         if compile:
+            if isinstance(config, Int8WeightOnlyConfig) and isinstance(
+                config.granularity, PerTensor
+            ):
+                # currently the inductor lowering for weight only quant in core does not support per-tensor gpu, so this errors. Skipping for now, but will address this in core
+                return
             model_q = torch.compile(model_q, fullgraph=True)
 
         output_fp = model(input_tensor)


torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,352 @@
...
+        block_size = get_block_size(hp_tensor.shape, granularity)
+        block_size = list(block_size)
+
+        scale, zero_point = choose_qparams_affine(
+            input=hp_tensor,
+            mapping_type=MappingType.SYMMETRIC,
+            block_size=block_size,
+            target_dtype=torch.int8,
+            quant_min=hp_value_lb,
+            quant_max=hp_value_ub,
```
jerryzh168: `hp_value_lb` and `hp_value_ub` are actually for the (unquantized) floating point input tensor, but quant_min/quant_max are for the (quantized) int8 tensor, since here it's int8 which means (-128, 127), I feel we could just not pass around these args (these will be derived from `torch.int8` target dtype directly)

jerryzh168:
also since we have https://github.com/pytorch/ao/pull/3391 but there are some infra issues so some Cis are not running. I'd suggest to recreate this PR exact and land that first, and then we can do additional fixes on top of that to recognize contributions from OSS

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,302 @@
...
+        act_quant_kwargs: flags for dynamic activation quantization
+    """
+
+    # TODO: Static quantization support using `static_scale`
+    tensor_data_names = ["qdata", "scale"]
+    tensor_attribute_names = []
+    optional_tensor_attribute_names = [
+        "block_size",
+        "act_quant_kwargs",
+        "dtype",
```
jerryzh168: nit: should dtype be optional? or just required?

Oh I guess this is following float8 tensor, and only used for dequantize for now. I think one thing to make sure if just it still works when `dtype` == None

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -0,0 +1,302 @@
...
+            scale=scale,
+            zero_point=zero_point,
+            output_dtype=torch.int8,
+        )
+
+        # make scale the correct dim
+        if isinstance(granularity, PerRow):
+            scale = scale.unsqueeze(1)
+        elif isinstance(granularity, PerTensor):
+            scale = scale.unsqueeze(0).unsqueeze(1)
```
jerryzh168: these might be confusing, we could align with float8 path and have a `keepdim=True` variant for `choose_qparams_affine` I guess: https://github.com/pytorch/ao/blob/73730e8730a5bac22985a53c2cc4d9a3c17b362a/torchao/quantization/quant_primitives.py#L1553-L1554

like `choose_qparams_affine_keepdim`

torchao/quantization/quant_api.py
```diff
@@ -1509,7 +1520,9 @@ class Int8DynamicActivationInt8WeightConfig(AOBaseConfig):
     layout: Optional[Layout] = PlainLayout()
     act_mapping_type: Optional[MappingType] = MappingType.SYMMETRIC
     weight_only_decode: bool = False
+    granularity: Optional[Union[PerRow, PerTensor]] = PerRow()
```
jerryzh168: nit: typing needs to be updated to `Optional[Union[Granularity, List[Granularity, Granularity]]]` I think

torchao/quantization/quant_api.py
```diff
@@ -1524,52 +1537,66 @@ def __post_init__(self):
...
+            _layout=layout,
+            zero_point_domain=weight_zero_point_domain,
+        )
+        new_weight = to_linear_activation_quantized(new_weight, input_quant_func)
+        return new_weight
+    else:
+        activation_granularity, weight_granularity = _normalize_granularity(
+            config.granularity
+        )
+        act_quant_kwargs = QuantizeTensorToInt8Kwargs(
```
jerryzh168: should we also do some validations here on what are the combinations that's supported?

torchao/quantization/quant_api.py
```diff
@@ -1524,52 +1537,66 @@ def __post_init__(self):
...
+            activation_granularity,
+            # hp_value_lb=activation_value_lb,
+            # hp_value_ub=activation_value_ub,
```
jerryzh168: nit: remove

torchao/quantization/quant_api.py
```diff
@@ -1332,7 +1336,9 @@ class Int8WeightOnlyConfig(AOBaseConfig):
     """
 
     group_size: Optional[int] = None
+    granularity: Granularity = PerRow()
     set_inductor_config: bool = True
+    version: int = 2
```
jerryzh168: btw, I think we should break BC in a separate PR

torchao/quantization/quant_api.py
```diff
@@ -1509,7 +1520,9 @@ class Int8DynamicActivationInt8WeightConfig(AOBaseConfig):
     layout: Optional[Layout] = PlainLayout()
     act_mapping_type: Optional[MappingType] = MappingType.SYMMETRIC
     weight_only_decode: bool = False
+    granularity: Optional[Union[PerRow, PerTensor]] = PerRow()
     set_inductor_config: bool = True
+    version: int = 2
```
jerryzh168: same here, I think we should set this to 1 first and break BC separately to reduce the scope of changes

jerryzh168:
also for all the failed CIs, I think it's because you bumped versions, I think we can keep BC for all these older tests by setting version to 1 explicitly: https://github.com/pytorch/ao/blob/73730e8730a5bac22985a53c2cc4d9a3c17b362a/test/dtypes/test_affine_quantized_float.py#L114

while migrating the ones that we think that's going to be useful to test_int8_tensor.py

torchao/quantization/quant_api.py
jcaip: This should just be, granularity and activation_mapping_type (symmetric vs asymmetric), which is what the v1 config supports. Will add a test for these combos

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -140,93 +138,29 @@ def from_hp(
...
-            output_dtype = self.dtype
-
-        block_size = get_block_size(self.qdata.shape, self.granularity)
-
         return dequantize_affine(
             input=self.qdata,
-            block_size=block_size,
-            scale=self.scale,
+            block_size=self.block_size,
+            scale=self.scale.squeeze(),
```
jerryzh168: this is probably not very robust, maybe we can add a keepdim for both quantize_affine and dequantize_affine as well for now?

in the future we can just refactor these ops to use expanded scale/zero_point: https://github.com/pytorch/ao/issues/3324

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -54,21 +56,24 @@ class Int8Tensor(TorchAOBaseTensor):
...
-        granularity: Optional[Granularity] = None,
-        block_size: Optional[torch.Size] = None,
+        block_size: Optional[List[int]] = None,
         act_quant_kwargs: Optional[QuantizeTensorToInt8Kwargs] = None,
         dtype: Optional[torch.dtype] = None,
     ):
         kwargs = {
             "device": qdata.device,
-            "dtype": dtype or scale.dtype,
+            "dtype": dtype,
```
jerryzh168, line 80: is dtype becoming required? if so, we can move dtype from `optional_tensor_attribute_names` to `tensor_attribute_names` and change its order in the arglist

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -319,39 +266,42 @@ def _(func, types, args, kwargs):
...
+    """Select operation for Int8Tensor"""
+    old_int8_tensor, dim, index = args
+    assert dim == 0, f"Int8Tensor aten.select.int with {dim=} is not yet supported"
+    assert len(old_int8_tensor.qdata.shape) == len(old_int8_tensor.scale.shape), (
+        "unsupported"
+    )
+    assert len(old_int8_tensor.qdata.shape) == len(old_int8_tensor.block_size), (
+        "unsupported"
+    )
+    new_int8_tensor = old_int8_tensor.__class__(
```
jerryzh168: nit: we can just use `Int8Tensor` here to be more explicit I think

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -29,13 +30,14 @@
 
 @dataclass
 class QuantizeTensorToInt8Kwargs(QuantizeTensorKwargs):
-    """Tensor kwargs for creating int8 tensor (either activation or weight)
+    """Tensor kwargs for creating int8 tensor from high precision
```
jerryzh168: this is also just for activation I think, we can include that in the comment as well

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -54,21 +56,24 @@ class Int8Tensor(TorchAOBaseTensor):
 
     # TODO: Static quantization support using `static_scale`
     tensor_data_names = ["qdata", "scale"]
-    tensor_attribute_names = ["granularity"]
-    optional_tensor_attribute_names = ["act_quant_kwargs", "block_size", "dtype"]
+    tensor_attribute_names = []
+    optional_tensor_attribute_names = [
+        "block_size",
```
jerryzh168: also, should block_size be optional?

torchao/quantization/quant_api.py
```diff
@@ -1592,24 +1588,28 @@ def _int8_dynamic_activation_int8_weight_quantize_tensor(weight, config):
...
+        quantized_weight = to_linear_activation_quantized(new_weight, input_quant_func)
     else:
         from torchao.quantization.quantize_.workflows.int8.int8_tensor import (
             QuantizeTensorToInt8Kwargs,
         )
 
+        assert config.granularity in {PerRow(), PerTensor()}, (
+            "Only PerRow and PerTensor are supported"
+        )
+        weight_granularity, act_granularity = _normalize_granularity(config.granularity)
```
jerryzh168: `_normalize_granularity` includes both normalization and validation right now I think and it's float8 specific, I feel we should split normalize and validation and validation should be Int8 specific, see: https://github.com/pytorch/ao/blob/16aad7c2f4073d415a541cbfd00aa40fe4ef7162/torchao/prototype/float8_opaque_tensor/inference_workflow.py#L60-L62 for example

jerryzh168: (`_normalize_and_validate_granularity` is probably a better name)

torchao/quantization/quant_api.py
```diff
@@ -1592,24 +1588,28 @@ def _int8_dynamic_activation_int8_weight_quantize_tensor(weight, config):
...
-        quantized_weight = to_linear_activation_quantized(
-            quantized_weight, input_quant_func
-        )
+        quantized_weight = to_linear_activation_quantized(new_weight, input_quant_func)
     else:
         from torchao.quantization.quantize_.workflows.int8.int8_tensor import (
             QuantizeTensorToInt8Kwargs,
         )
 
+        assert config.granularity in {PerRow(), PerTensor()}, (
```
jerryzh168, line 1597: nit: if only these two values are allowed, maybe we can change the type of `config.granularity` to `Granularity` (also dropping Optional since it's not Optional)

torchao/quantization/quant_api.py
jcaip: Yeah I think we should just restrict this to PerRow and PerTensor for now. I think probably we don't need a function for this

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
jcaip: Took a look and the code as currently written will be fine for both flattend and non-flattened case, we just need keepdims for quantize_affine

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
jcaip: I think its optional here because its kwarg that defaults to None, I'm a little unclear on what the default block_size should be though. should we default to PerTensor?

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -237,14 +175,16 @@ def _(func, types, args, kwargs):
...
-    if not isinstance(weight_tensor, Int8Tensor):
-        raise TypeError(f"Expected weight to be Int8Tensor, got {type(weight_tensor)}")
+    assert isinstance(weight_tensor, Int8Tensor), (
+        f"Expected weight to be Int8Tensor, got {type(weight_tensor)}"
+    )
 
     output_dtype = activation_tensor.dtype
 
     if weight_tensor.act_quant_kwargs is not None:
         activation_tensor = Int8Tensor.from_hp(
```
jerryzh168: mapping_type not passed?

jerryzh168: also should this use the `_choose...` API?

torchao/quantization/quant_api.py
```diff
@@ -1592,24 +1588,32 @@ def _int8_dynamic_activation_int8_weight_quantize_tensor(weight, config):
...
 
         # TODO: Symmentric/Asymmetric choice for weight quantization
         # https://github.com/pytorch/ao/pull/3241#discussion_r2551515539
-        # TODO: Add block_size args to return in from_hp
-        # https://github.com/pytorch/ao/pull/3241#discussion_r2552016429
         quantized_weight = Int8Tensor.from_hp(
             weight,
-            granularity=config.granularity,
-            act_quant_kwargs=QuantizeTensorToInt8Kwargs(granularity=config.granularity),
+            granularity=weight_granularity,
```
jerryzh168, line 1612: mapping_type for weight is not passed?

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
jcaip: I was just following how Float8Tensor did it: https://github.com/pytorch/ao/blob/2872c9b8255ffa9e7597a7a825b7066e0652d236/torchao/quantization/quantize_/workflows/float8/float8_tensor.py#L109

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
jcaip: `from_hp` calls `_choose_qparams_affine` under the hood.

torchao/quantization/quant_api.py
jcaip: the config doesn't have an option for the weight_mapping_type, so we just use the default (symmetric)

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
jerryzh168: oh sorry I meant this: https://github.com/pytorch/ao/blob/0975a4027df983c45a0799c59cbf18786b6fae1a/torchao/quantization/quantize_/common/quantize_tensor_kwargs.py#L33

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
jerryzh168: probably just make it required, it will be simpler I think, value of block_size is dependent on input so we can't really specify a per tensor block_size without knowing the shape of input

torchao/quantization/quant_api.py
```diff
@@ -1522,9 +1526,7 @@ class Int8DynamicActivationInt8WeightConfig(AOBaseConfig):
     layout: Optional[Layout] = PlainLayout()
     act_mapping_type: Optional[MappingType] = MappingType.SYMMETRIC
     weight_only_decode: bool = False
-    # TODO: Revisit for supported granularitys
-    # https://github.com/pytorch/ao/pull/3241#discussion_r2551497849
-    granularity: Optional[Granularity] = PerRow()
+    granularity: Union[PerRow, PerTensor] = PerRow()
```
jerryzh168: nit: just `granularity: Granularity`?

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -105,32 +112,29 @@ def __repr__(self):
     @classmethod
     def from_hp(
         cls,
-        w_hp: torch.Tensor,
+        hp_tensor: torch.Tensor,
         granularity: Granularity = PerRow(),
```
jerryzh168: nit: we can make granularity required I think

torchao/quantization/quantize_/common/quantize_tensor_kwargs.py
```diff
@@ -53,4 +55,11 @@ def _choose_quant_func_and_quantize_tensor(
             quant_kwargs.kernel_preference,
         )
 
+    if isinstance(quant_kwargs, QuantizeTensorToInt8Kwargs):
+        return Int8Tensor.from_hp(
+            tensor,
+            quant_kwargs.granularity,
+            quant_kwargs.mapping_type,
```
jerryzh168: nit: pass the optional arg mapping_type by name is probably better

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -29,18 +33,22 @@
...
 class QuantizeTensorToInt8Kwargs(QuantizeTensorKwargs):
-    """Tensor kwargs for creating int8 tensor (either activation or weight)
+    """Tensor kwargs for creating int8 tensor for activation.
 
     Args:
         granularity: the granularity for the Tensor, currently either PerRow() or PerTensor()
+        mapping_type: whether to use symmetric or asymmetric quant, only symmetric is supported currently
     """
 
     granularity: Granularity = PerRow()
```
jerryzh168: nit: same here, we can just make granularity required I think

jerryzh168:
forgot to mention, will this flatten stuff affect performance?

jcaip:
> forgot to mention, will this flatten stuff affect performance?

Didn't run benchmarks, but flatten should just be a view for 2d -> 1d so there shouldn't be a perf hit.

PR #3408, opened 2025-12-01
add an example for quantizing LLaMa 4 Scout

vkuzo:
Summary:

Adds an e2e example of how to use torchao to quantize LLaMa 4 Scout.

Note that this needs:
* a recent `transformers` version (higher than 4.57, not officially
  released yet so user needs to build from source)
* a recent `fbgemm_gpu` version nightly from `2025.11.22` or after
* to run this in vLLM, https://github.com/vllm-project/vllm/pull/28421
  is needed (not yet landed).

Test Plan:

```bash
with-proxy time python examples/quantize_llama_4.py ~/local/tmp/20251201_test/
```

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/examples/quantize_llama_4.py
+++ b/examples/quantize_llama_4.py
@@ -22,7 +22,7 @@ from transformers import AutoModelForCausalLM, AutoTokenizer, TorchAoConfig
 
 from torchao.quantization import (
     Float8DynamicActivationFloat8WeightConfig,
-    ModuleFqnToConfig,
+    FqnToConfig,
     PerRow,
 )
 from torchao.quantization.quantize_.workflows.float8.float8_tensor import (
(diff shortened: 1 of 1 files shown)


vkuzo:
* __->__ #3408

examples/quantize_llama_4.py
```diff
@@ -0,0 +1,223 @@
...
+
+def get_quantization_config():
+    expert_3d_weight_single_config = Float8DynamicActivationFloat8WeightConfig(
+        # the weights of this model are stored in (B, K, N) layout, and we
+        # need to quantize rowwise across the K axis, which is `PerRow(1)`.
+        granularity=[PerRow(), PerRow(1)],
+        # guard against activations with groups of all-zeroes
+        activation_value_lb=1.0e-12,
+    )
+    module_fqn_to_config = ModuleFqnToConfig(
```
jcaip: we should use `FqnToConfig` instead of `ModuleFqnToConfig`

PR #3418, opened 2025-12-03
[Reland][PT2E][X86] Add Inductor fusion passes of float8 qconv for X8…

jiayisunx:
…6Inductor backend

Reland https://github.com/pytorch/ao/pull/3261

### DIFF
--- a/test/quantization/pt2e/test_x86inductor_fusion.py
+++ b/test/quantization/pt2e/test_x86inductor_fusion.py
@@ -142,10 +142,22 @@ class FP8QDQLinear(torch.nn.Module):
 
 
 class FP8QDQConv2d(torch.nn.Module):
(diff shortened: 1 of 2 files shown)


Xia-Weiwen (approved):

Xia-Weiwen:
The CI failures are unrelated.

jerryzh168:
looks like still fails? https://github.com/pytorch/ao/actions/runs/19906303811/job/57063261477

Xia-Weiwen:
> looks like still fails? https://github.com/pytorch/ao/actions/runs/19906303811/job/57063261477

Oh... Wondering why CI didn't catch that. Is the case run in CI?

jerryzh168:
yeah not exactly sure why..

Xia-Weiwen:
> yeah not exactly sure why..

Then I feel we are not sure how to merge this properly. Shall I resubmit so that you can import it to run internal tests? Is that enough to avoid such failures?

PR #3419, opened 2025-12-03
Roofline quantized conv3d/2d layer 

jainapurva:
This pull request extends the float8 inference roofline benchmarking code to support convolution operations (conv2d and conv3d) in addition to linear layers. It introduces new utilities and refactors the workflow to enable roofline modeling and kernel benchmarking for convolutional operations, including calculation of equivalent GEMM dimensions and measurement of kernel times. A conv kernel (FBGEMM) is a combination of im2col + implicit GEMM

### DIFF
--- a/benchmarks/float8/float8_inference_roofline.py
+++ b/benchmarks/float8/float8_inference_roofline.py
@@ -492,8 +492,6 @@ def run(
         assert recipe_name in _SUPPORTED_CONV_RECIPES, (
             f"Recipe '{recipe_name}' is not supported for {op_name}. "
             f"Supported recipes for conv operations: {_SUPPORTED_CONV_RECIPES}. "
-            f"Note: rowwise requires per-tensor for 4D/5D tensors, "
-            f"mxfp8/mxfp4/nvfp4 have block size or implementation constraints."
         )
 
     if op_name == "conv2d":


benchmarks/float8/float8_inference_roofline.py
```diff
@@ -347,44 +413,105 @@ def run(
...
         b_bf16_e2e_time_s, b_fp8_e2e_time_s = 0, 0
-        if do_benchmarks:
+        # Check hardware requirements for conv operations
+        skip_conv_benchmarks = (
+            do_benchmarks
+            and op_name in ("conv2d", "conv3d")
+            and not is_sm_at_least_100()
+        )
+
+        if skip_conv_benchmarks:
```
jerryzh168: I feel conditions seems a bit convoluted here

maybe:
```
if do_benchmarks:
    if op_name in ("conv2d", "conv3d") and not is_sm_at_least_100():
        print warning
   else:
      # can also move this part to a function to make it clearer
       ....
```

benchmarks/float8/float8_inference_roofline.py
```diff
@@ -347,44 +413,105 @@ def run(
...
+                stride=stride,
+                padding=padding,
+            )
 
-            r_fp8_ovhd_time_s = None
-            r_fp8_gemm_and_ovhd_s = None
-            r_speedup = None
+            # use roofline model to estimate gemm time using equivalent GEMM dims
+            r_bf16_gemm_time_s = float(
+                bf16_gemm_time_sympy.subs(M, gemm_M).subs(K, gemm_K).subs(N, gemm_N)
```
jerryzh168, line 657: is the memory operations of conv the same as linear well? https://github.com/pytorch/ao/blob/0975a4027df983c45a0799c59cbf18786b6fae1a/torchao/testing/training/roofline_utils.py#L332

benchmarks/float8/float8_inference_roofline.py
```diff
@@ -260,20 +330,17 @@ def run(
         print("fp8_ovhd_time_sympy", fp8_ovhd_time_sympy)
         print()
     else:
-        # TODO: enable roofline analysis for conv
         pass
```
jerryzh168: if we share the same thing, we should remove this if/else branch and inline the code in if here I think

benchmarks/float8/float8_inference_roofline.py
```diff
@@ -347,44 +413,105 @@ def run(
...
+            )
+            r_fp8_ovhd_time_s = float(
+                fp8_ovhd_time_sympy.subs(M, gemm_M).subs(K, gemm_K).subs(N, gemm_N)
+            )
+            r_fp8_gemm_and_ovhd_s = r_fp8_gemm_time_s + r_fp8_ovhd_time_s
+            r_speedup = r_bf16_gemm_time_s / (r_fp8_gemm_time_s + r_fp8_ovhd_time_s)
 
-            # real gemm benchmark time, also not added yet
-            # if enabled, also measured observed gemm time
+            # gemm benchmarks for conv not implemented, as conv uses implicit GEMM
```
jerryzh168: we should run the conv ops I think?

benchmarks/float8/float8_inference_roofline.py
jainapurva: As conv is an implicit gemm, I'm assuming the memory operations for gemm and conv should be same.

benchmarks/float8/float8_inference_roofline.py
jainapurva: I kept if/else just incase of an unexpected op_name. We can either make the code verify op_name in beginning to avoid any errors, or assume the input for op_name will always be correct

jerryzh168: we already verify that in L264 I think

benchmarks/float8/float8_inference_roofline.py
jainapurva: We're running conv op in benchmarks

jerryzh168: that's a bit different I think, that one is doing e2e speedup. 

we can do the same that linear is doing: https://github.com/pytorch/ao/blob/2ae2994b5021c3e636490c8ab7e771e72a0349cd/benchmarks/float8/float8_inference_roofline.py#L88 that only run the fp8 conv op itself (without all the quant ops for act and weight)

benchmarks/float8/float8_inference_roofline.py
jbschlosser: implicit gemm means that one of the operands of the matmul is not fully materialized (specifically, the one corresponding with the convolutional filters), so I'd expect somewhat different memory access behavior for implicit gemms: i.e. accessing the same memory locations multiple times to effectively "unfold" the implicitly-materialized operand. should this be accounted for here?

benchmarks/float8/utils.py
```diff
@@ -400,6 +400,41 @@ def get_gpu_kernel_gemm_time_s(f, *args, **kwargs):
...
+        "aten::slow_conv_dilated3d",
+        "torchao::_conv2d_fp8_inner",
+        "torchao::_conv3d_fp8_inner",
+        "fbgemm::f8f8bf16_conv",
+    }
+
+    # Filter out aten::fill_ and other non-conv operations
+    filtered_data = {k: v for k, v in data.items() if k in expected_conv_kernels}
+
+    assert len(filtered_data) >= 1, f"unexpected data: {data}"
```
jbschlosser: maybe the error message should indicate something about potential incompleteness of the above expected conv kernel list?

benchmarks/float8/utils.py
```diff
@@ -400,6 +400,41 @@ def get_gpu_kernel_gemm_time_s(f, *args, **kwargs):
...
+    # Filter to only conv-related kernels and remove aten::fill_
+    expected_conv_kernels = {
+        "aten::conv2d",
+        "aten::conv3d",
+        "aten::convolution",
+        "aten::cudnn_convolution",
+        "aten::slow_conv_dilated2d",
+        "aten::slow_conv_dilated3d",
+        "torchao::_conv2d_fp8_inner",
+        "torchao::_conv3d_fp8_inner",
```
jerryzh168: where are these from?

benchmarks/float8/utils.py
```diff
@@ -400,6 +400,41 @@ def get_gpu_kernel_gemm_time_s(f, *args, **kwargs):
...
+        "fbgemm::f8f8bf16_conv",
```
jerryzh168: we just updated this to mslk btw

benchmarks/float8/float8_inference_roofline.py
```diff
@@ -165,6 +166,278 @@ def do_matmul(A, B):
...
+    else:
+        raise ValueError(f"Unsupported op_name: {op_name}")
+
+    # Measure bf16 conv time
+    bf16_time_s = get_gpu_kernel_conv_time_s(
+        lambda x, w: conv_fn(x, w, **conv_kwargs), x_bf16, w_bf16
+    )
+
+    # For fp8 conv timing, we need to use fbgemm operator
+    if recipe_name in ("mxfp4_cutlass", "nvfp4"):
```
jerryzh168: these are not supported, so we don't need to check for these I think, we can add a check to say what are the supported recipe_names

benchmarks/float8/float8_inference_roofline.py
```diff
@@ -165,6 +166,278 @@ def do_matmul(A, B):
...
+        # For now, skip fp4 conv operations as they're more complex to set up
+        f8_time_s = 0.0
+    else:
+        # Try to use fbgemm fp8 conv operator
+        try:
+            # Check if fbgemm fp8 conv is available
+            if not hasattr(torch.ops.fbgemm, "f8f8bf16_conv"):
```
jerryzh168: should use https://github.com/pytorch/ao/blob/e4bfca1fb2bf02800eae66579d39777564601b6c/torchao/quantization/quantize_/workflows/float8/float8_tensor.py#L44 to check I think

benchmarks/float8/float8_inference_roofline.py
```diff
@@ -347,125 +620,149 @@ def run(
...
 
-            # real gemm benchmark time, also not added yet
-            # if enabled, also measured observed gemm time
+            # measure actual conv kernel times (without quant overhead)
             b_bf16_gemm_time_s, b_fp8_gemm_time_s = 0, 0
-            # gemm roofline ratio achieved in real benchmark
             rb_bf16_gemm_ratio = -1
             rb_fp8_gemm_ratio = -1
+            if do_benchmarks:
+                try:
```
jerryzh168: I think we should remove try / except to make it simpler

benchmarks/float8/float8_inference_roofline.py
```diff
@@ -165,6 +166,278 @@ def do_matmul(A, B):
...
+        f8_time_s = 0.0
+    else:
+        # Try to use fbgemm fp8 conv operator
+        try:
+            # Check if fbgemm fp8 conv is available
+            if not hasattr(torch.ops.fbgemm, "f8f8bf16_conv"):
+                print(
+                    "Warning: fbgemm.f8f8bf16_conv not available, skipping fp8 conv timing"
+                )
+                f8_time_s = 0.0
```
jerryzh168: nit: I think we can just error out, and talk about how user can turn off the conv benchmarking by setting `do_benchmarks` to False?

this will also reduce the indentation here

benchmarks/float8/float8_inference_roofline.py
```diff
@@ -165,6 +166,278 @@ def do_matmul(A, B):
...
+                        w_fp8,
+                    )
+                else:
+                    # conv2d not yet implemented for fbgemm
+                    print(
+                        "Warning: fbgemm fp8 conv2d not implemented, skipping fp8 conv timing"
+                    )
+                    f8_time_s = 0.0
+
+        except Exception as e:
```
vkuzo: just raise the exception

benchmarks/float8/float8_inference_roofline.py
jainapurva: As discussed offline, we are already overestimating the calculation for memory gem time, hence we can keep it the same.

benchmarks/float8/utils.py
jainapurva: These we placeholders. Cleaned it up

Issue #2917, opened 2025-09-02
Intel XPU CI Enabling


liangan1:
# Motivation 

To improve quality on Intel XPU device, we plan to enable the CI/CD process on Intel XPUs. The CI is based on the docker env and we will firstly enable the CI infra and then enable the UTs feature by feature. 

**Attention**: XPU CI is only triggered by "ciflow/xpu" label and the test folders or test_xxx file should be included by the [.github/scripts/ci_test_xpu.sh](https://github.com/pytorch/ao/pull/3335/files#diff-b202517cefd1a330b08ef698a15f2292ba85091bbe9332109e4b7e774a23adba)

# Status
   The common dependency, e.g.,  CI infra(#3027) and test folders(#3335) has been landed.  The feature will be enabled one by one. Community owner suggests us to make the PR as small as possible(Not change more than 2 files per PR) to reduce the review effort.  

# Task List

@zxd1997066 is the owner of this task list. 

- [x]  #3027 
- [x] #3335 
- [x] #3350 
- [x] #3351 
- [x] #3355 
- [x] #3366 
- [ ] #3332 
- [x] #3177   
- [ ] #3248 
- [ ] #2802 
- [ ] #2803

zxd1997066:
- [ ] https://github.com/pytorch/ao/pull/3377
- [ ] https://github.com/pytorch/ao/pull/3409
- [ ] https://github.com/pytorch/ao/pull/3423
- [ ] https://github.com/pytorch/ao/pull/3424
- [ ] https://github.com/pytorch/ao/pull/3434
- [ ] https://github.com/pytorch/ao/pull/3475
- [ ] https://github.com/pytorch/ao/pull/3500

jerryzh168:
let's restrict to stable features in https://github.com/pytorch/ao/tree/main/torchao/quantization/quantize_ for now

PR #3424, opened 2025-12-03
[xpu][test] Port test_float8_utils.py and test_hqq_affine.py UT files to intel XPU

zxd1997066:
For https://github.com/pytorch/ao/issues/2917, This PR is targeted to port test/hqq/test_hqq_affine.py and test/float8/test_float8_utils.py to intel XPU.


zxd1997066:
cc @liangan1

PR #3429, opened 2025-12-03
Add pruning-aware training in torchao.prototype.pat

lisjin:
Adding our pruning-aware training (PAT) library as a prototype. The original library is under fairinternal/qpat but we would like to surface it in torchao for broader adoption.

The interface is almost identical to torchao.prototype.parq, but we use (group) Lasso instead of piecewise-affine regularization. More details on code organization and usage can be found in the README.


lisjin:
@andrewor14 Let me know if anything needs to be cleared up in this diff. I'm hoping to update [D88501706](https://www.internalfb.com/diff/D88501706) so that it imports from torchao.prototype.pat instead of copying code.

torchao/prototype/pat/optim/pruneopt.py
```diff
@@ -0,0 +1,402 @@
...
+    from torch.distributed.tensor.experimental import local_map
+    from torch.distributed.tensor.placement_types import Partial, Replicate, Shard
+
+
+class PruneOptimizer(Optimizer):
+    """PruneOptimizer assembles functionalities of the following objects:
+    a base optimizer (e.g., SGD or AdamW)
+        - update the latent variables for QAT
+    Other parameters:
+        warmup_steps: int >= 0
```
andrewor14, line 34: This is the central API right, can we add an example usage in this docstring?

torchao/prototype/pat/layers/masked_layernorm.py
```diff
@@ -0,0 +1,47 @@
...
+    count = torch.sum(nz_mask, dim=norm_dims, keepdim=True).clamp_(min=1)
+    mean = torch.sum(a, dim=norm_dims, keepdim=True) / count
+
+    a_center = torch.where(nz_mask, a - mean, torch.zeros_like(a))
+    biased_var = torch.sum(a_center.pow(2), dim=norm_dims, keepdim=True) / count
+    out = a_center * torch.rsqrt(biased_var + eps)
+    return out
+
+
+class MaskedLayerNorm(nn.LayerNorm):
```
andrewor14, line 33: Seems like this is not used anywhere other than in tests. Can we delete this? Am I missing something?

torchao/prototype/pat/utils/__init__.py
```diff
@@ -0,0 +1,14 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
andrewor14: nit: can we just flatten this and make this `utils/distributed.py -> distributed_utils.py` and `utils/torch.py -> utils.py`? Seems like we don't need an extra module?

torchao/prototype/pat/utils/distributed.py
```diff
@@ -0,0 +1,56 @@
...
+
+from torch import Tensor
+from torch import distributed as dist
+
+try:
+    from torch.distributed.tensor import DTensor
+
+    HAS_DTENSOR = True
+except ImportError:
+    HAS_DTENSOR = False
```
andrewor14: torchao only supports the last 3 stable versions of pytorch (in this case 2.9, 2.8, and 2.7). Seems like we don't need this try except since DTensor has been in this location since 2.4+ already

torchao/prototype/pat/optim/pruneopt.py
```diff
@@ -0,0 +1,402 @@
...
+from torch.optim import Optimizer
+
+from ..utils import HAS_DTENSOR, instantiate_module, is_dtensor, is_main_process
+from ..utils.distributed import _maybe_async_aggregate, _sum_async_streams
+from ..utils.torch import get_index_linspace
+
+if HAS_DTENSOR:
+    from torch.distributed.tensor import distribute_tensor
+    from torch.distributed.tensor.experimental import local_map
+    from torch.distributed.tensor.placement_types import Partial, Replicate, Shard
```
andrewor14: These 3 imports should be in 2.7.0+, probably don't need to guard these imports

torchao/prototype/pat/optim/nm_sgd.py
```diff
@@ -0,0 +1,48 @@
...
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+
+import torch
+from torch.optim import Optimizer
+
+from .pruneopt import PruneOptimizer
+
+
+class NMSGDOptimizer(PruneOptimizer):
```
andrewor14: General question: I notice a lot of APIs in this PR that are not used or referenced anywhere. Are these all user-facing APIs? If so can we document them somewhere (e.g. main README) and explain how they're related to the main `PruneOptimizer` API? If they're not user-facing APIs and they're not used, do we still need them?

Some examples:
- NMSGDOptimizer
- ProxNuclearNorm
- all the groupers like QKSVDGrouper

andrewor14:
Hi @lisjin looks good overall. My main comment is just my confusion about how the APIs are used, seems like the code snippet in the main README only references 1 or 2 of these, so it's unclear to me how the rest are related. Would be great if you can clarify this in documentation.

Separately do you have any initial results? If so, would be great to include these in the README too.

torchao/prototype/pat/optim/pruneopt.py
lisjin: Good call—I updated the README example to include keyword args like warmup_steps and reg_lambda

torchao/prototype/pat/layers/masked_layernorm.py
lisjin: I'm hoping to keep this class since it's important for converting pruned models to their compressed inference-ready forms. This functionality can be added to PAT in the future.

torchao/prototype/pat/optim/nm_sgd.py
lisjin: * The NMSGDOptimizer was written by a summer intern last year and has shown promising results. Since it's an experimental feature, we don't have unit tests for it yet.
* ProxNuclearNorm is important for applying low-rank pruning to embeddings. [Here](https://github.com/fairinternal/qpat/blob/main/configs/prune_config_svd.yaml)'s an example config.
* The other groupers are more experimental. It would be great to keep them around so that we can stay in sync with  the [original repo](https://github.com/fairinternal/qpat), but I can also remove them if you'd like.

lisjin:
@andrewor14 Thanks for taking the time to review this back in Dec! I found out in January that the team I was collaborating with no longer needed to use PAT in torchao. However, now @Ninja91 and his team are planning to experiment with PAT. Could you please check that my fixes addressed all your comments? I've also added some initial results on unstructured pruning to the README.

torchao/prototype/pat/distributed_utils.py
```diff
@@ -0,0 +1,50 @@
...
+# LICENSE file in the root directory of this source tree.
+
+from typing import List
+
+from torch import Tensor
+from torch import distributed as dist
+from torch.distributed.tensor import DTensor
+
+
+def is_main_process():
```
andrewor14: are these functions meant to be called by the user? If not we should put an _ before the name to mark it as such

torchao/prototype/pat/distributed_utils.py
```diff
@@ -0,0 +1,50 @@
...
+    rank = dist.get_rank() if dist.is_available() and dist.is_initialized() else 0
+    return rank == 0
+
+
+def is_dtensor(x):
+    return isinstance(x, DTensor)
+
+
+class NoopHandle:
```
andrewor14: same here, these seem like util classes/functions

torchao/prototype/pat/README.md
```diff
@@ -0,0 +1,61 @@
...
+# split params into prunable and non-prunable groups
+weights = [p for name, p in model.named_parameters() if name.endswith("weight")]
+others = [p for name, p in model.named_parameters() if not name.endswith("weight")]
+
+# apply row-wise group Lasso regularization to the weights
+param_groups = [
+    {
+        "params": weights",
+        "group_type": "pat.group.Dim0Grouper",
+        "prox_type": "pat.prox.ProxGroupLasso",
```
andrewor14: Should these take in actual classes instead of strings of classes? Seems like it'll be more robust

torchao/prototype/pat/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+    AttentionHeadGrouperDim1,
+    ConvFilterGrouper,
+    Dim0Grouper,
+    Dim1Grouper,
+    ElemGrouper,
+    LayerGrouper,
+    PackedSVDGrouper,
+    QKGrouper,
+    QKSVDGrouper,
+    SVDGrouper,
```
andrewor14, line 16: I guess these classes are intended for users to import to use with `PruneOptimizer`? If so they should be documented somewhere (README.md is fine). Otherwise we shouldn't expose them as public APIs, and shouldn't keep them around if they're not used

torchao/prototype/pat/optim/nm_sgd.py
andrewor14: I see, we can keep them if we document them somewhere. If they're experimental we can mark them as such in the README. In general public APIs should have associated documentation somewhere, otherwise users won't be able to find them

torchao/prototype/pat/README.md
lisjin: Ah this usage is actually outdated. I updated it a while back to accept strings like `"Dim0Grouper"` and `"ProxGroupLasso"` so that there's no dependency on import structure. The README is fixed to reflect this.

lisjin:
@andrewor14 Thanks for the suggestions again. Here's what I've updated in the latest commit:
- Removed experimental classes like `NMSGDOptimizer`, `QKGrouper`, `QKSVDGrouper`
- Documented all remaining grouper and proximal mapping classes in a new table of the README
- Added underscores to non user-facing methods in distributed_utils.py

Let me know if anything's missing—this is very much a research prototype :)

PR #3411, opened 2025-12-02
[xpu][test] Port 2 test/prototype/test_{parq, quantized_training} UT files to intel XPU

zxd1997066:
For https://github.com/pytorch/ao/issues/2917, This PR is targeted to port test/prototype/test_parq.py and test/prototype/test_quantized_training.py to intel XPU.

PR #3432, opened 2025-12-04
Revert "[xpu][test] Port 2 test/prototype/test_{parq, quantized_training} UT files to intel XPU"

metascroy:
Reverts pytorch/ao#3411

Looks like this breaks CI:

image


jerryzh168:
cc @zxd1997066 @liangan1

PR #3434, opened 2025-12-04
[reland][xpu]Quantized training on intel XPU

zxd1997066:
## Introduction 
quantized training is one of features listed in the #3576, Since the design can be reused by XPU, we just need to enable the UTs to ensure the feature works on the Intel XPU.


zxd1997066:
cc @liangan1

liangan1 (approved):

vkuzo:
what do performance / accuracy results look like on your hardware

PR #3435, opened 2025-12-04
Add support for MXFP8 All gather 

avizon-aws:
Currently when trying to use TP All gather with MXFP8, it gives an unimplemented error for the all gather operation. This PR adds support for MXFP8 all gather which will enable better performance as it will reduce the bandwidth required for the gather compute.

### DIFF
--- a/.github/workflows/4xH100_tests.yml
+++ b/.github/workflows/4xH100_tests.yml
@@ -47,3 +47,4 @@ jobs:
         pip install . --no-build-isolation
         ./test/float8/test_everything_multi_gpu.sh
         ./test/prototype/mx_formats/test_mx_dtensor.sh
+        ./test/prototype/mx_formats/test_mxfp8_allgather.sh


torchao/prototype/tests/test_mxfp8_allgather.py
```diff
@@ -0,0 +1,108 @@
...
+            
+            # Perform all_gather
+            gathered_mx = torch.ops._c10d_functional.all_gather_into_tensor.default(
+                local_mx,
+                world_size,
+                "0",
+            )
+            gathered_mx = torch.ops._c10d_functional.wait_tensor.default(gathered_mx)
+            
+            # ✅ Verify type
```
danielvegamyhre: please remove emojis

danielvegamyhre:
cc @vkuzo @drisspg

torchao/prototype/tests/test_mxfp8_allgather.py
avizon-aws: removed.

vkuzo:
looks reasonable, would you mind rebasing as we've landed some refactors to MXTensor recently?

just curious, which framework are you using TP + mxfp8 from?

avizon-aws:
I think I already rebased just before creating the PR (yesterday night), so it should be up to date with mainline. I fixed the ruff formating to pass the workflows, I think one of the workflow was failing due to a missing tag on the PR, i think the right tag should be "add new feature", i wasnt able to add it, I think i dont have the permission for it.

I tried the TP AG using native pytorch, i created a test script for it, didnt use any other framework currently. I updated the sharding strategy for to_mx due to which it required the TP AG.

danielvegamyhre:
lgtm! will leave approval for @vkuzo who i think is also taking a look at this.

@avizon-aws to fix the code analysis ci job, please run the linter with `ruff check --fix torchao/ test/` and `ruff format torchao/ test/`

avizon-aws:
Thanks @danielvegamyhre , i have used the commands shared above to fix the formatting.

torchao/prototype/mx_formats/mx_tensor.py
```diff
@@ -840,3 +840,82 @@ def mx_select(func, types, args, kwargs):
...
+        ),  # The scale factors, Need to cast to uint8 as float8_e8m0fnu is not support for all gather.
+        group_tag,
+        *args[2:],
+        **kwargs,
+    )
+
+    gathered_scale = gathered_scale.view(torch.float8_e8m0fnu)
+
+    # Return new MXTensor with gathered data
+    return MXTensor(
```
vkuzo, line 883: this definitely looks like it needs a rebase, we have cleaned up some of the fields in MXTensor.  `_pack_fp6` is no longer there at all, and `_gemm_kernel_choice` has been renamed to `kernel_preference`.

torchao/prototype/tests/test_mxfp8_allgather.py
```diff
@@ -0,0 +1,122 @@
...
+
+@instantiate_parametrized_tests
+class MXFP8OnDeviceAllGatherTest(MultiProcessTestCase):
+    def setUp(self) -> None:
+        super().setUp()
+        self._spawn_processes()
+
+    @property
+    def world_size(self) -> int:
+        return 4
```
vkuzo: maybe make this 2 to mininize the requirements to run?

torchao/prototype/mx_formats/mx_tensor.py
avizon-aws: I see, I will check it again and fix it, thanks for the review.

torchao/prototype/tests/test_mxfp8_allgather.py
avizon-aws: Will change it to 2.

torchao/prototype/mx_formats/mx_tensor.py
vkuzo: also, can we add the new test runs to CI? CI is currently green but my guess would be that the new test isn't running there.

avizon-aws: yeah, that sounds good, could you advise how to do that, I am unfamiliar with that part.

vkuzo: this should be the right line: https://github.com/pytorch/ao/blob/69ce0fdf7125ee54558257e8214216efe157d170/.github/workflows/1xH100_tests.yml#L54, either this isn't capturing your test or your test is skipped in CI

vkuzo: you should make sure your test is in `test/prototype/mx_formats` to follow repo conventions

avizon-aws: I see, i think the test might be getting skipped due to:

```
if not torch.cuda.is_available() or torch.cuda.get_device_capability() != (10, 0):
    pytest.skip("Test requires CUDA build on SM100", allow_module_level=True)
```

avizon-aws: I will move it to `test/prototype/mx_formats`, currently it is in `torchao/prototype/tests/test_mxfp8_allgather.py`

avizon-aws:
I see some issues during rebasing, will fix them and then ping for review.

avizon-aws:
@vkuzo , i have fixed the rebase

avizon-aws:
I think the test is failing in the CI because of version and compatibility issues (I have run the test locally and it passes on compatibility >=9)), I will have to add some conditions due to which the test might need to be skipped for certain CUDA versions.

avizon-aws:
I added checks to ensure the test runs only on CUDA and also for compatibility >=9  (H100 and above as FP8 is supported well)

avizon-aws:
It seems like there is only 1 H100 GPU and hence the test is failing, because the world size required for my test is 2. What are the recommendations to proceed here? There are 2 options:

1. Perform the test on CPU
2. Skip the test

I think skipping the test is risky, I would prefer option 1

vkuzo:
https://github.com/pytorch/ao/blob/51fd90e50dd42deb9556b3519c41106e76f689ab/.github/workflows/4xH100_tests.yml#L49 is a job that runs with 4 H100 GPUs. I apologize for not linking this file originally!

vkuzo:
thank you! this should be good to land once we get CI to pass

avizon-aws:
Thanks for sharing the above file, I will have to update the test to not use pytest and then explicitly create a new .sh script that calls my test file using torchrun instead because currently I was using the distributedtest utility from pytorch, will update and share.

avizon-aws:
@vkuzo , i have updated the CR, i think it should pass the CI this time.

vkuzo:
CI looks good, thank you!

danielvegamyhre:
@avizon-aws @vkuzo hmm the test is failing in main branch CI with permission denied error trying to run the script. @avizon-aws if you can help take a look that would be great, i'll investigate on our side as well: https://github.com/pytorch/ao/actions/runs/20110277303/job/57705428270

avizon-aws:
@danielvegamyhre , based on the error, it seems like the command to execute the bash script i.e. `./test_mxfp8_allgather.sh` is failing, I think it could be because we need to give executable permissions to this file and hence the error. I checked the other .sh scripts in the 4xH100 tests.yml, but i couldnt see any explicit code for adding the permission. Kindly advise.

danielvegamyhre:
> @danielvegamyhre , based on the error, it seems like the command to execute the bash script i.e. `./test_mxfp8_allgather.sh` is failing, I think it could be because we need to give executable permissions to this file and hence the error. I checked the other .sh scripts in the 4xH100 tests.yml, but i couldnt see any explicit code for adding the permission. Kindly advise.

I'm pretty sure we set the executable bit before committing the script to the repo. e.g., 

- chmod +x ./my_script
- git add, git commit etc

avizon-aws:
Yeah, so I meant that its possible that my script (test_mxfp8_all_gather.sh) is missing the permissions. I can create another PR where i add the permissions and then add the file again, I think it should fix the issue. But I am wondering why the test did not fail in the CI of the PR? Because this test was not run in PR CI?

Edit: I checked the permission:

```
git ls-files -s /home/avizon/torchao_PR/ao/test/prototype/mx_formats/test_mxfp8_allgather.sh
100644 180375af405d346e260ce5317c695c5a10da48dc 0       test/prototype/mx_formats/test_mxfp8_allgather.sh
```
So that is indeed the case.

avizon-aws:
I created a PR for this: https://github.com/pytorch/ao/pull/3479, this should fix the issue.

PR #3441, opened 2025-12-05
[bc-breaking] update default int8 tensor version to v2

jcaip:


jerryzh168:
will also need to change all these old tests to use v1, e.g. in https://github.com/pytorch/ao/blob/main/test/dtypes/test_affine_quantized.py

PR #3442, opened 2025-12-05
Add int8 static quantization workflow

jcaip:
This PR adds in a new static quant workflow based off of Int8Tensor (#3407). 

It introduces a new config, `Int8StaticActivationInt8WeightConfig` which requires a scale tensor and granularity

```python
    scale=int8_input.scale.detach(), granularity=PerRow
)
quantize_(model_static_quant, static_config)
```

Currently PerRow and PerTensor symmetric quant is support only. 

This scale tensor is stored on the weight Int8Tensor under `activation_scale`, and is used to create a new activation Int8Tensor for static quantization. 

It would be nice to store this scale tensor in `QuantizeTensorToInt8Kwargs` but unfortunately this breaks dynamo tracing, as we store the quant kwargs as an object for the weight tensor and we are unable to fakeify them properly. 

As a result, we need to keep track and pass scale outside of this `Kwargs` object.

### DIFF
--- a/test/quantization/quantize_/workflows/int8/test_int8_tensor.py
+++ b/test/quantization/quantize_/workflows/int8/test_int8_tensor.py
@@ -230,38 +230,7 @@ class TestInt8Tensor(TorchAOIntegrationTestCase):
 class TestInt8StaticQuant(TorchAOIntegrationTestCase):
     @common_utils.parametrize("granularity", [PerRow(), PerTensor()])
     @common_utils.parametrize("dtype", [torch.bfloat16])
-    def test_static_activation_per_row_int8_weight_earger(self, granularity, dtype):
-        M, N, K = 32, 32, 32
-        input_tensor = torch.randn(M, K, dtype=dtype, device="cuda")
-
-        model_static_quant = (
-            torch.nn.Linear(K, N, bias=False).eval().to(device="cuda", dtype=dtype)
-        )
-        model_dynamic_quant = copy.deepcopy(model_static_quant)
-
-        dynamic_config = Int8DynamicActivationInt8WeightConfig(
-            version=2, granularity=granularity
-        )
-        quantize_(model_dynamic_quant, dynamic_config)
-
-        dynamic_quantize_out = model_dynamic_quant(input_tensor)
-
-        int8_input = _choose_quant_func_and_quantize_tensor(
-            input_tensor, model_dynamic_quant.weight.act_quant_kwargs
-        )
-
-        static_config = Int8StaticActivationInt8WeightConfig(
-            scale=int8_input.scale.detach().clone(), granularity=granularity
-        )
-        quantize_(model_static_quant, static_config)
-
-        static_quantize_out = model_static_quant(input_tensor)
-        torch.testing.assert_close(dynamic_quantize_out, static_quantize_out)
-
-    @common_utils.parametrize("granularity", [PerRow(), PerTensor()])
-    @common_utils.parametrize("dtype", [torch.bfloat16])
-    def test_static_activation_per_row_int8_weight_compile(self, granularity, dtype):
-        # for compile, we can't compare dynamic vs static because we may get slightly different qparams
+    def test_static_activation_per_row_int8_weight(self, granularity, dtype):
         torch.compiler.reset()
 
         M, N, K = 32, 32, 32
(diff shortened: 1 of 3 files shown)


torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -116,22 +118,34 @@ def from_hp(
...
+                quant_min=-128,
+                quant_max=127,
+                scale_dtype=hp_tensor.dtype,
+                zero_point_dtype=torch.int8,
+                keepdim=True,
+            )
+        else:
+            # Scale can be provided in the case of static quant
+            assert scale.ndim == hp_tensor.ndim
+            if isinstance(granularity, PerTensor):
```
jcaip: note: I changed these checks in #3468

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -89,13 +89,15 @@ def __init__(
         block_size: List[int],
         dtype: torch.dtype,
         act_quant_kwargs: Optional[QuantizeTensorToInt8Kwargs] = None,
+        activation_scale=None,
```
jerryzh168: nit: act_scale to align with `act_quant_kwargs`, also optional tensor should come before optional tensor attributes I think (example: https://github.com/pytorch/ao/blob/08e5e203f1074a1663fe8293e184b252ea359f99/test/test_utils.py#L257)

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
jerryzh168: I think we typically also check the shape of `scale` tensor as well, like these https://github.com/pytorch/ao/blob/08e5e203f1074a1663fe8293e184b252ea359f99/torchao/float8/inference.py#L181-L198

torchao/quantization/quant_api.py
```diff
@@ -1634,7 +1634,90 @@ def _int8_dynamic_activation_int8_weight_transform(
...
+    act_mapping_type: Optional[MappingType] = MappingType.SYMMETRIC
+    set_inductor_config: bool = True
+    version: int = 1
+
+    def __post_init__(self):
+        torch._C._log_api_usage_once(
+            "torchao.quantization.Int8StaticActivationInt8WeightConfig"
+        )
+        if isinstance(self.granularity, PerTensor):
+            assert self.scale.numel() == 1
```
jerryzh168: nit: also check the shapes, and check PerRow as well

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -221,5 +225,93 @@ def test_available_gpu_kernels(self):
...
+            scale=int8_input.scale.detach().clone(), granularity=granularity
+        )
+        quantize_(model_static_quant, static_config)
+
+        static_quantize_out = model_static_quant(input_tensor)
+        torch.testing.assert_close(dynamic_quantize_out, static_quantize_out)
+
+    @common_utils.parametrize("granularity", [PerRow(), PerTensor()])
+    @common_utils.parametrize("dtype", [torch.bfloat16])
+    def test_static_activation_per_row_int8_weight_compile(self, granularity, dtype):
```
jerryzh168: nit: just merge with the previous test with `compile` arg? can use a branch for different testing as well, or just always use sqnr

torchao/quantization/quant_api.py
```diff
@@ -1634,7 +1634,90 @@ def _int8_dynamic_activation_int8_weight_transform(
...
+
+
+@dataclass
+class Int8StaticActivationInt8WeightConfig(AOBaseConfig):
+    """
+    Configuration for applying float8 static symmetric quantization to
+
+    Args:
+        scale (torch.Tensor): The scale tensor for activation quantization.
+        activation_dtype (torch.dtype): The target data type for activation quantization. Default is torch.float8_e4m
```
jerryzh168: e4m3?

torchao/quantization/quant_api.py
```diff
@@ -1634,7 +1634,90 @@ def _int8_dynamic_activation_int8_weight_transform(
...
+        weight_dtype (torch.dtype): The target data type for weight quantization. Default is torch.float8_e4m
```
jerryzh168: e4m3?

torchao/quantization/quant_api.py
jcaip: I think we might want to enable scales to be None, for passing `Int8StaticActivationInt8WeightConfig()` as a base config, we can discuss on #3468

jerryzh168: OK makes sense, sounds good, this is easier for user I think, otherwise they have to do a separate flow to get the scale here

PR #3443, opened 2025-12-05
skip certain mxfp8 tests for cuda < 12.8

danielvegamyhre:
When preparing the release, I found certain tests depending on the mxfp8_cuda extension would run when the extension was not built (for example, with cuda 12.6). 

We only build the mxfp8_cuda kernels for CUDA 12.8 and above, so we shouldn't run the test for older versions of CUDA.

However, we should still run the test without explicitly checking if the extension is available because this will help alert us to any build issues, such as the C++ extension not being built in 12.8 nightlies

### DIFF
--- a/test/prototype/mx_formats/test_mx_linear.py
+++ b/test/prototype/mx_formats/test_mx_linear.py
@@ -26,6 +26,7 @@ from torchao.prototype.mx_formats.mx_linear import (
 from torchao.quantization import quantize_
 from torchao.quantization.utils import compute_error
 from torchao.utils import (
+    is_cuda_version_at_least,
     is_sm_at_least_89,
     is_sm_at_least_100,
     torch_version_at_least,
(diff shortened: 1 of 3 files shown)


test/prototype/mx_formats/test_mx_linear.py
```diff
@@ -50,12 +51,25 @@ def run_around_tests():
 elem_dtypes = (
     [
         # test each dtype
-        (torch.float8_e4m3fn, torch.float8_e4m3fn, torch.float8_e4m3fn),
+        (
```
vkuzo, line 54: nit: unrelated, split to separate PR?

torchao/utils.py
```diff
@@ -512,9 +513,11 @@ def _same_metadata(self: TorchAOBaseTensor, src: TorchAOBaseTensor) -> bool:
         if hasattr(self, "optional_tensor_data_names"):
             # either both are None or both are not Tensors and the shape match
             _optional_tensor_shape_match = all(
-                getattr(self, t_name).shape == getattr(src, t_name).shape
-                if getattr(self, t_name) is not None
-                else getattr(src, t_name) is None
+                (
```
vkuzo, line 516: nit: unrelated, split to separate PR?

test/prototype/mx_formats/test_mx_linear.py
danielvegamyhre: it's autoformatting, not an actual code change. I think maybe ruff got a version update or something, not sure why it's suddenly changing the format, but code analysis passes

(same with other comment below)

vkuzo: you can make sure your ruff version matches https://github.com/pytorch/ao/blob/69ce0fdf7125ee54558257e8214216efe157d170/dev-requirements.txt#L33 to avoid this

danielvegamyhre: hmm it does match this, maybe this code hadn't been formatted in a while, not sure.

vkuzo: it could be that both syntaxes are correct, and if you ran it on your original code it wouldn't do these changes. I've seen this behavior before.

vkuzo: not worth revert, but IMO worth not doing this in general and making sure formatting changes are separate from logic changes

PR #3447, opened 2025-12-05
Align _choose_qparams_affine with _choose_scale_float8 behavior

AryanBagade:
Changes keepdim default from False to True in _choose_qparams_affine to match _choose_scale_float8 behavior. This ensures scale/zero_point maintain the same rank as input tensor, making downstream handling more consistent.

Part 1 of fixing #3324

### Core Changes (`torchao/quantization/quant_primitives.py`)
- Changed `keepdim: bool = False` → `keepdim: bool = True` in both `choose_qparams_affine` (line 1220) and `_choose_qparams_affine` (line 1526)
- Added reshape logic (lines 1600-1608) to match `_choose_scale_float8` behavior
- Saved `original_input_size` before reshaping to compute correct output shape
- Added documentation explaining the alignment with `_choose_scale_float8`

### Workflow Simplification (`torchao/quantization/quantize_/workflows/intx/intx_unpacked_to_int8_tensor.py`)
- Removed manual reshape logic (lines 247-254) that is no longer needed

### Test Updates(`test/quantization/test_quant_primitives.py`)
- Updated 3 test cases to squeeze scale/zero_point before comparison with reference values
- All 7 `test_choose_qparams` tests now pass

### DIFF
--- a/torchao/quantization/quant_primitives.py
+++ b/torchao/quantization/quant_primitives.py
@@ -1535,7 +1535,7 @@ def _choose_qparams_affine(
        and `zero_point_domain`
 
     Note:
-        keepdim defaults to True to align with _choose_scale_float8 behavior. This ensures
+        Set keepdim=True to align with _choose_scale_float8 behavior. This ensures
         scale/zero_point maintain the same rank as input, making it easier to handle downstream.
     """
     quant_min, quant_max = _get_and_check_qmin_qmax(target_dtype, quant_min, quant_max)


torchao/quantization/quant_primitives.py
```diff
@@ -1591,6 +1598,15 @@ def _choose_qparams_affine(
...
             zero_point_dtype = torch.int32
 
+    # Reshape scale and zero_point to match expected output shape
+    # This aligns with _choose_scale_float8 behavior
+    if keepdim:
+        output_shape = [
+            original_input_size[i] // block_size[i] for i in range(len(block_size))
+        ]
+        scale = scale.reshape(output_shape)
+        zero_point = zero_point.reshape(output_shape)
```
jerryzh168, line 1608: is this needed? `keepdim=True` is already used in int8_tensor.py

Edit: oh OK I think it is needed because the reshapes we are doing before https://github.com/pytorch/ao/blob/aa21b808ad625f89a761ef31e1c81d3903306e7e/torchao/quantization/quant_primitives.py#L1554

AryanBagade: Yes, this reshape is needed! **int8_tensor.py** passes scale/zero_point directly to `quantize_affine`, which internally reshapes the scale at line 461. So it doesn't need the output to be pre-reshaped. but the thing is, **IntxUnpackedToInt8Tensor.__init__** (lines131-136) asserts that `scale.shape` must **exactly** match `tuple(n_blocks)` before passing to `quantize_affine`:
```python
  assert scale.shape == tuple(n_blocks), ( 
  f"Expected scale to have shape {n_blocks} (inferred from
  block_size={block_size}), but got {scale.shape}"
  )

Comment (AryanBagade) [reply torchao/quantization/quant_primitives.py:1608]: basically without this reshape:
  - keepdim=True gives scale shape like (1, 5, 1) for block_size (10, 4) on input (10, 20)
  - But IntxUnpackedToInt8Tensor expects (1, 5)
  - The assertion would fail!

**Comment by jerryzh168** [comment]
Thanks, I think it's a good start, we can remove `keepdim` arg in next PR after this PR is merged

File: torchao/quantization/quant_primitives.py (hunk 1 above)
Comment (jerryzh168) [reply torchao/quantization/quant_primitives.py:1608]: yeah that's correct, I'll approve the CI to run all the tests to see, especially this one: https://github.com/pytorch/ao/blob/aa21b808ad625f89a761ef31e1c81d3903306e7e/test/quantization/test_quant_primitives.py#L569

Comment (jerryzh168) [reply torchao/quantization/quant_primitives.py:1608]: also in a future PR we could remove some of the reshaping logic in quantize_affine/dequantize _affine as well: https://github.com/pytorch/ao/blob/aa21b808ad625f89a761ef31e1c81d3903306e7e/torchao/quantization/quant_primitives.py#L453-L461

also eventually remove the block_size arg from these ops (bc-breaking)

Comment (AryanBagade) [reply torchao/quantization/quant_primitives.py:1608]: sure, sounds like a plan, happy to contribute

**Comment by AryanBagade** [comment]
I see 25 integration tests failed due to backward compatibility issues with the `keepdim=True` default change

**Comment by jerryzh168** [comment]
it's expected, I think maybe just don't change the default for now, but turn the keepdim to True in these tests one by one to make sure these tests are fixed, and alls the callsites are fixed before making the switch would be better

**Comment by AryanBagade** [comment]
Really sorry, had a super busy schedule

**Comment by AryanBagade** [comment]
@jerryzh168 could you please run the CI, made the required changes!

**Comment by AryanBagade** [comment]
@jerryzh168 could you please let me know what further steps to take!

**Comment by jerryzh168** [comment]
@AryanBagade https://github.com/pytorch/ao/actions/runs/20121850704/job/57745005031?pr=3447 is failing but likely is fixed in main already, can you rebase on main branch? I have imported and confirmed that this change doesn't break any of the internal test, so we should be able to merge after you rebase

File: torchao/quantization/quant_primitives.py [hunk 2]
```diff
@@ -1532,6 +1533,10 @@ def _choose_qparams_affine(
     2. find min_val/max_val based on the dimension for reduction
     3. calculate quantization parameters based on min_val/max_val based on args like `preserve_zero`
        and `zero_point_domain`
+
+    Note:
+        keepdim defaults to True to align with _choose_scale_float8 behavior. This ensures
```
jerryzh168: right now it's defaults to `False`, so should change the comment here

AryanBagade:
@jerryzh168  Rebased on main and fixed the docstring comment as requested. Ready for re-import and merge!

torchao/quantization/quant_primitives.py
```diff
@@ -1532,6 +1533,10 @@ def _choose_qparams_affine(
     2. find min_val/max_val based on the dimension for reduction
     3. calculate quantization parameters based on min_val/max_val based on args like `preserve_zero`
        and `zero_point_domain`
+
+    Note:
+        Set keepdim=True to align with _choose_scale_float8 behavior. This ensures
```
jerryzh168, line 1538: please change this to False, thanks

jerryzh168: oh sorry, seems OK, it's not talking about default

PR #3448, opened 2025-12-05
Add nightly regression tests for CUDA 12.8 and 12.9 on H100/A100

jainapurva:
We need to test TorchAO more extensively with the various cuda versions (12.8, 12.9). A nightly test run for those will help us catch regressions faster. This is a test PR to see what failures occur on these versions


jerryzh168:
is this too many tests to run?

cc @huydhn @atalman

jainapurva:
> is this too many tests to run?
> 
> cc @huydhn @atalman

If these are too many tests, we can make this a weekly test, or as a test run for release process, rather than running it nightly

PR #3449, opened 2025-12-05
create a new accuracy eval script for official README.md eval accuracy

vkuzo:
Summary:

Creates a standalone eval script for generating accuracy metrics for
quantization README.md, based on the HuggingFace model definition of
LLaMa 3.1 8B

Why new script?
1. the current `prod` script in
   https://github.com/pytorch/ao/blob/main/torchao/_models/llama/eval.py
   uses a custom model definition, this was pre-HF integration, it's better to use HF's model definition now
2. we have HummingBird scripts in
   https://github.com/pytorch/ao/tree/40c4f44677ae11166c3dcfbb9189cfa78789390c/.github/scripts/torchao_model_releases,
   but they seem pretty verbose and hard to use/modify
3. we have
   https://github.com/pytorch/ao/blob/main/benchmarks/_models/eval_hf_models.py,
   I copy-pasted and modified this for the current PR. The script above
   didn't work as is for various reasons, and also seemed to be hard to
   use/modify, for main README.md it's important to have a very simple
   standalone script.

We should probably do a pass on the naming before landing.

Future work:
1. add metrics for `int4_weight_only_hqq` (need to run on A100)
2. add metrics for 'int4 weight float8 activation' (currently doesn't work with HF accelerate)
3. add metrics for `mxfp8` and `nvfp4` (need to run on B200)
4. make the parsing of logs automated
5. also add a similar script for performance benchmarks, using vllm
6. delete https://github.com/pytorch/ao/blob/main/torchao/_models/llama/

Test Plan:

```
// debug run on small model
with-proxy time ./benchmarks/quantization/eval_accuracy_for_readme.sh facebook/opt-125m

// real run
with-proxy time ./benchmarks/quantization/eval_accuracy_for_readme.sh
```

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/benchmarks/quantization/eval_accuracy_for_readme.sh
+++ b/benchmarks/quantization/eval_accuracy_for_readme.sh
@@ -21,9 +21,9 @@ time python -u benchmarks/quantization/eval_accuracy_for_readme.py $BASE_ARGS 2>
 
 # quantized recipes
 # note:
-# * `int4_weight_float8_rowwise_activation` doesn't work with dtype_map auto: https://gist.github.com/vkuzo/6b128681b628744d445c553cdeac8a85
-# * `int4_weight_only_hqq` only works on A100
-for quant_recipe in float8_rowwise int4_weight_float8_rowwise_activation int4_weight_only_hqq int8_weight_only int8; do
+# * `int4_groupwise_hqq_weight_float8_rowwise_activation` doesn't work with dtype_map auto: https://gist.github.com/vkuzo/6b128681b628744d445c553cdeac8a85
+# * `int4_groupwise_hqq_weight_only` only works on A100
+for quant_recipe in float8_rowwise int4_groupwise_weight_float8_rowwise_activation int4_groupwise_hqq_weight_only int8_rowwise_weight_only int8_rowwise; do
   time python -u benchmarks/quantization/eval_accuracy_for_readme.py $BASE_ARGS --quant_recipe_name $quant_recipe 2>&1 | tee -a "$LOG_FILE"
 done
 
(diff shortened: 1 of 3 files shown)


vkuzo:
* #3466
* __->__ #3449

benchmarks/quantization/eval_accuracy_for_readme.sh
```diff
@@ -0,0 +1,30 @@
...
+# baseline
+# note: the -u flag is to prevent python from buffering stdout and stderr
+# and make the output log file be in chronological order
+time python -u benchmarks/quantization/eval_accuracy_for_readme.py $BASE_ARGS 2>&1 | tee "$LOG_FILE"
+
+# quantized recipes
+# note:
+# * `int4_groupwise_hqq_weight_float8_rowwise_activation` doesn't work with dtype_map auto: https://gist.github.com/vkuzo/6b128681b628744d445c553cdeac8a85
+# * `int4_groupwise_hqq_weight_only` only works on A100
+for quant_recipe in float8_rowwise int4_groupwise_weight_float8_rowwise_activation int4_groupwise_hqq_weight_only int8_rowwise_weight_only int8_rowwise; do
```
jerryzh168, line 26: nit: `int4_groupwise_weight_float8_rowwise_activation` --> `float8_rowwise_activation_int4_groupwise_weight` to match the config name order?

vkuzo: I'm matching the order in https://github.com/pytorch/ao/tree/main?tab=readme-ov-file#stable-workflows

overall if we want to standardize this everywhere sounds reasonable, IMO let's do that in a separate "rename-only" PR?

PR #3450, opened 2025-12-05
Fix NVFP4 QAT convert path

andrewor14:
**Summary:** The previous convert path through `QATConfig` did not swap `NVFP4FakeQuantizedLinear` back to `torch.nn.Linear`. The numerics tests still passed because this fake quantized linear happen to match the PTQ numerics exactly.

**Test Plan:**
```
```

### DIFF
--- a/test/quantization/test_qat.py
+++ b/test/quantization/test_qat.py
@@ -2106,29 +2106,46 @@ class TestQAT(TestCase):
         Test QAT with `NVFP4FakeQuantizeConfig`.
         """
         from torchao.prototype.mx_formats import NVFP4DynamicActivationNVFP4WeightConfig
-        from torchao.prototype.qat import NVFP4FakeQuantizeConfig
+        from torchao.prototype.mx_formats.nvfp4_tensor import NVFP4Tensor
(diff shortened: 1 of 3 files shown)


test/quantization/test_qat.py
```diff
@@ -2132,6 +2137,18 @@ def test_qat_nvfp4(self, use_per_tensor_scale: bool):
         sqnr = compute_error(out, baseline_out).item()
         self.assertGreaterEqual(sqnr, float("inf"))
 
+        # Compare converted values
```
jerryzh168, line 2140: oh we didn't compare convert results before?

jerryzh168 (approved):

test/quantization/test_qat.py
andrewor14: It's tested here: https://github.com/pytorch/ao/blob/048568f4fb48a7ebdcff2c616fc396b21a9c00c5/test/quantization/test_qat.py#L2085

We just never explicitly checked it's using tensor subclasses after convert (tests still passed because QAT prepare mimics PTQ exactly)

PR #3455, opened 2025-12-05
update version compatibility table

danielvegamyhre:
for 0.15.0 release. will need to cherry pick this into the release branch i think, the release runbook had the steps out of order, i updated it accordingly.

### DIFF
--- a/torchao/__init__.py
+++ b/torchao/__init__.py
@@ -76,10 +76,12 @@ elif not ("+git" in __version__) and not ("unknown" in __version__):
         # Built against torch 2.9.0
         (_parse_version("0.14.1"), _parse_version("2.9.0")),
         (_parse_version("0.14.1"), _parse_version("2.10.0.dev")),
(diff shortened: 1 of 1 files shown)


torchao/__init__.py
```diff
@@ -76,6 +76,9 @@ def _parse_version(version_string):
         # Built against torch 2.9.0
         (_parse_version("0.14.1"), _parse_version("2.9.0")),
         (_parse_version("0.14.1"), _parse_version("2.10.0.dev")),
+        # Built against torch 2.9.1
+        (_parse_version("0.15.0"), _parse_version("2.9.1")),
+        (_parse_version("0.15.0"), _parse_version("2.10.0.dev")),
         # Current torchao version
```
jerryzh168, line 82: should the current torchao version also be updated?

danielvegamyhre: Thanks, updated

PR #3461, opened 2025-12-06
Fix type annotation of _choose_qparams_affine/gguf(): input = None

Freed-Wu:


jerryzh168 (approved):
seems reasonable, not sure why it's Optional before

Freed-Wu:
> seems reasonable,

@jerryzh168 Can we merge this PR?

PR #3468, opened 2025-12-08
enable smoothquant for int8 static tensor

jcaip:
This PR hooks up the static quant workflow added in #3442 to the prototype smoothquant API. 

You can use the new flow like follows:
```python
    Int8StaticActivationInt8WeightConfig,
)
from torchao.prototype.smoothquant import (
    SmoothQuantConfig
)

config = SmoothQuantConfig(
            base_config=Int8StaticActivationInt8Weight(granularity=PerRow()),
            step=SmoothQuantStep.PREPARE,
            alpha=0.5,
        )

quantize_(model, config)

# Perform calibration with test data
model(*x)

config.step = SmoothQuantStep.CONVERT
quantize_(model, config)

# model will now be statically quantized with the inputs used in smoothquant observer. 
model(*x)
```

Test Plan:
```
pytest test/quantization/quantize_/workflows/int8/test_int8_tensor.py
pytest test/prototype/test_smoothquant.py 
```

### DIFF
--- a/test/quantization/quantize_/workflows/int8/test_int8_tensor.py
+++ b/test/quantization/quantize_/workflows/int8/test_int8_tensor.py
@@ -300,11 +300,15 @@ class TestInt8StaticQuant(TorchAOIntegrationTestCase):
         static_out_compile = model_static_quant(input_tensor)
         sqnr_static_compile = compute_error(model_out_baseline, static_out_compile)
 
-        assert sqnr_static_compile == sqnr_dynamic_compile, (
-            f"Static SQNR mismatch: compile={sqnr_static_compile} vs eager={sqnr_static_eager}"
+        assert abs(sqnr_static_compile - sqnr_dynamic_compile) < 1, (
+            f"Static compile vs dynamic compile SQNR difference too large: "
+            f"static_compile={sqnr_static_compile} vs dynamic_compile={sqnr_dynamic_compile}, "
+            f"diff={abs(sqnr_static_compile - sqnr_dynamic_compile)}"
         )
-        assert sqnr_static_eager == sqnr_dynamic_eager, (
-            f"Static eager vs dynamic eager SQNR mismatch: {sqnr_static_eager} vs {sqnr_dynamic_eager}"
+        assert abs(sqnr_static_eager - sqnr_dynamic_eager) < 1, (
+            f"Static eager vs dynamic eager SQNR difference too large: "
+            f"static_eager={sqnr_static_eager} vs dynamic_eager={sqnr_dynamic_eager}, "
+            f"diff={abs(sqnr_static_eager - sqnr_dynamic_eager)}"
         )
         # eager numerics should match exactly
         # for compile, we can't compare dynamic vs static because we may get slightly different qparams when fused


jcaip:
cc @Xia-Weiwen and @cyxlily fyi

torchao/prototype/smoothquant/api.py
```diff
@@ -120,6 +137,7 @@ def _smooth_quant_transform(
     qw = to_weight_tensor_with_linear_activation_scale_metadata(
```
jerryzh168: we should not be using this, please check awq on how this should be implemented in the new stack: https://github.com/pytorch/ao/blob/08e5e203f1074a1663fe8293e184b252ea359f99/torchao/prototype/awq/api.py#L108-L113

torchao/quantization/quant_api.py
```diff
@@ -1658,7 +1658,7 @@ class Int8StaticActivationInt8WeightConfig(AOBaseConfig):
         version (int): the version of the config
     """
 
-    scale: torch.Tensor
+    scale: torch.Tensor = None
```
jerryzh168: nit: Optional[torch.Tensor]

test/prototype/test_smoothquant.py
```diff
@@ -83,7 +85,10 @@ def setUpClass(cls):
     @common_utils.parametrize(
         "base_config",
         [
-            Int8DynamicActivationInt8WeightConfig(),
+            Int8DynamicActivationInt8WeightConfig(version=2),
+            # TODO: not sure if we should allow not passing scales as part of static config?
```
jerryzh168: yeah I think it's fine

side note: we may need a separate API/flow for plain static quant without Smoothquant if needed.

cyxlily:
@jcaip Our customer needs activation quantization PerTensor and weight quantization PerRow. Will you implement it, or may I create a new PR to do it?

jcaip:
@cyxlily feel free to open a new PR for activation per tensor x weight per row, it's not something im planning to do currently. 

Thank you for your smoothquant pr btw, I used it to implement this.

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -221,5 +225,93 @@ def test_available_gpu_kernels(self):
...
+        model_static_quant = torch.compile(model_static_quant, fullgraph=True)
+
+        static_out_compile = model_dynamic_quant(input_tensor)
+        sqnr_static_compile = compute_error(model_out_baseline, static_out_compile)
+
+        assert (
+            sqnr_static_compile
+            == sqnr_static_eager
+            == sqnr_dynamic_compile
+            == sqnr_dynamic_eager
```
jerryzh168: are we trying to say `static_out_compile` == `static_out_eager` == `dynamic_out_compile` == `dynamic_out_eager` here? if so I think it might be clearer just to assert all these are equal to each other

torchao/prototype/smoothquant/api.py
```diff
@@ -95,8 +97,18 @@ def _smooth_quant_transform(
     else:
         raise ValueError(f"Unexpected step: {step}")
 
+    if isinstance(base_config, Int8StaticActivationInt8WeightConfig):
```
jerryzh168: I think we shouldn't have specific config here, maybe change this to a similar protocol like `SupportsActivationPreScaling` for config?

torchao/quantization/quant_api.py
jerryzh168: also maybe `static_scale` might be more descriptive I feel

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -292,6 +302,8 @@ def _(func, types, args, kwargs):
             block_size,
             self.dtype,
             act_quant_kwargs=self.act_quant_kwargs,
+            act_scale=self.act_scale,
```
jerryzh168: I guess slice doesn't work for static quant int8 before, can you add a test for that?

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -322,6 +334,8 @@ def _(func, types, args, kwargs):
         old_int8_tensor.scale[index],
         old_int8_tensor.block_size[1:],
         old_int8_tensor.dtype,
+        old_int8_tensor.act_scale,
```
jerryzh168: same for this one, seems like select op breaks before with static quant

torchao/prototype/smoothquant/api.py
jcaip: I think figuring out how to do this generally will need a bit more design, we'd need to figure out how to map to the appropriate `QuantizeTensorToInt/FloatXKwargs` object. Agree we should be able to do this though, but can I address in a later PR?

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -106,6 +108,7 @@ def __repr__(self):
             f"qdata={self.qdata}, "
             f"scale={self.scale}, "
             f"act_scale={self.act_scale}, "
+            f"act_pre_scale={self.act_scale}, "
```
cyxlily: act_pre_scale

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
jcaip: For slice and select, they work when granularity is PerDim(=-1), but not otherwise. we now throw an exception when passing PerDim != -1 and I added a test for both.

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -297,17 +300,96 @@ def test_static_activation_per_row_int8_weight(self, granularity, dtype):
...
+        1. PerRow(dim != -1) activation quantization raises an error (per-feature not supported)
+        2. PerRow(dim=-1) and PerTensor() work correctly with weight slicing
+
+        Per-feature activation quantization (PerRow(dim=0)) would require slicing
+        act_quant_scale when weight is sliced, which is not currently supported.
+        Per-token activation quantization (PerRow(dim=-1)) should work correctly
+        with weight slicing since act_quant_scale doesn't need to be sliced.
+        """
+
+        with self.assertRaises(ValueError) as cm:
```
jerryzh168, line 323: nit: probably better to have a separate test, seems weird to test this here

torchao/prototype/smoothquant/api.py
jerryzh168: we can get this info from a callback I think, like base_config.get_activation_quant_kwargs(), I feel you can iterating from that, but not from the current state

torchao/prototype/smoothquant/api.py
```diff
@@ -111,15 +123,21 @@ def _smooth_quant_transform(
     linear.bias = observed_linear.bias
 
     # Quantize weights
+    if isinstance(base_config, Int8StaticActivationInt8WeightConfig):
+        base_config.static_scale = activation_scale
```
jerryzh168: can do the same as `SupportsActivationPreScaling`, like `IsStaticQuantizationConfig`

torchao/quantization/quant_api.py
```diff
@@ -1182,14 +1182,14 @@ class Int8StaticActivationInt8WeightConfig(AOBaseConfig):
...
-        scale (torch.Tensor): The scale tensor for activation quantization.
+        static_scale (torch.Tensor): The scale tensor for activation quantization.
         granularity (Granularity): The granularity of quantization. PerRow() and PerTensor() are supported currently
         act_mapping_type (MappingType): The mapping type for activation quantization. only SYMMETRIC is supported currently
         set_inductor_config (bool): if True, adjusts `torchinductor` settings to recommended values.
         version (int): the version of the config
     """
 
-    scale: torch.Tensor
+    static_scale: Optional[torch.Tensor] = None
```
jerryzh168: nit: should we say `act_static_scale` to be more concrete?

torchao/quantization/quant_api.py
```diff
@@ -1231,7 +1239,7 @@ def _int8_static_activation_int8_weight_transform(
             granularity=activation_granularity,
             mapping_type=config.act_mapping_type,
         ),
-        act_scale=config.scale.detach(),
+        act_quant_scale=config.static_scale.detach(),
```
jerryzh168: nit: should align the name, probably also `act_static_scale`

or `act_quant_scale` is fine as well since it corresponds better with `act_pre_scale`

jcaip: let's use `act_quant_scale`, i think that's clearer

PR #3470, opened 2025-12-09
simplify accuracy eval

vkuzo:
Summary:

1. move model creation to a separate script, so it can be reused by future perf benchmarks
2. move lm_eval call to shell script, which is simpler than calling it from a python subprocess
3. simplify the for loop by moving the `None` baseline case inside the loop

Test Plan:

```bash
benchmarks/quantization/eval_accuracy_for_readme.sh
```

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo:
* #3473
* #3472
* __->__ #3470

PR #3471, opened 2025-12-09
use python version agnostic binding for mxfp8 cuda kernels

danielvegamyhre:
## Summary 
- Migrate from PyBind to TORCH_LIBRARY API, which is python version agnostic
- Update setup.py to use python limited (version agnostic) tag
- Use pattern seen in `torchao/ops.py` for other CUDA C++ extensions
  - build `torchao._C_mxfp8` so file (lands in build/ dir) instead of separate torchao.prototype.mxpf8_cuda`extension (landed in torchao/prototype)
  - Define new op schema for 2d quantization and 3d quantization kernels
  - Update custom ops, meta functions, and custom sharding registrations to wrap new custom op

### Context

While doing the 0.15.0 torchao release and testing the test build for cuda 12.8, and i found the "torchao.prototype.mxfp8_cuda" c++ extension cannot be found (import error, module not found). we only build the extension for cuda 12.8+, so i checked the logs and i see logs indicating it was built: https://github.com/pytorch/ao/actions/runs/20046209265/job/57498462190

so then i checked the local installation itself, and i do see a .so file for the extension in the torchao/prototype dir, so it is definitely being built.

i tried asking claude about this and it says the build for python3.10 must match the python version in the conda env due to ABI incompatibility (i'm using python 3.12). as a test, i tried a fresh conda env with python 3.10, and instead of module not found, i get an undefined symbol error, so that does seem to indicate some python ABI issue.

asking @drisspg he said we should be building with a py agnostic flag, so i looked into this and we are doing this for other c++ extensions but not mxfp8_cuda, so I am fairly certain this is the root cause and this PR will fix the issue.

### DIFF
--- a/benchmarks/mx_formats/cast_bench.py
+++ b/benchmarks/mx_formats/cast_bench.py
@@ -366,19 +366,19 @@ def run(
         bps = (bytes_r + bytes_w) / (time_us / 1e6)
 
     elif mode == "dim1_mxfp8_cuda_rceil":
-        from torchao.prototype import mxfp8_cuda
+        from torchao.prototype.mx_formats.kernels import mxfp8_quantize_cuda
 
-        _, y_d1, _, s_d1 = mxfp8_cuda.quantize(
+        _, y_d1, _, s_d1 = mxfp8_quantize_cuda(
             x, rowwise=False, colwise=True, scaling_mode="rceil"
         )
 
         for _ in range(2):
-            __ = mxfp8_cuda.quantize(
+            __ = mxfp8_quantize_cuda(
                 x, rowwise=False, colwise=True, scaling_mode="rceil"
             )
 
         time_us = benchmark_cuda_function_in_microseconds(
-            lambda x: mxfp8_cuda.quantize(
+            lambda x: mxfp8_quantize_cuda(
                 x, rowwise=False, colwise=True, scaling_mode="rceil"
             ),
             x,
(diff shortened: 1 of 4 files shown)


atalman (approved):
lgtm

drisspg:
spoke offline, we need to be using torchlib and not use pybind

test/prototype/mx_formats/test_kernels.py
```diff
@@ -587,48 +585,16 @@ def test_cuda_mx_dim1_numerics(M, K, input_dtype, scaling_mode):
...
-
-@pytest.mark.skipif(
-    not is_sm_at_least_100(),
-    reason="MXFP8 requires CUDA capability 10.0 or greater",
-)
-@pytest.mark.skipif(
-    not is_cuda_version_at_least(12, 8),
-    reason="CUDA version >= 12.8 required for MXFP8 CUDA kernels",
-)
-def test_cuda_mx_dim1_invalid_block_size():
```
danielvegamyhre, line 617: deleting this test since block_size of 32 is hard coded in the python wrapper for the kernel now, since we always use this for mxfp8

drisspg:
you should be able to use nm -D to also investigate the symbols and ensure there are none from python

danielvegamyhre:
current CI failures will be resolved once this rollback in upstream pytorch is included in the next torch nightly: https://github.com/pytorch/pytorch/pull/169985

danielvegamyhre:
other upstream pytorch issue causing CI issues has now been resolved: https://github.com/pytorch/pytorch/issues/170184

drisspg (approved):

danielvegamyhre:
mac-os test failure is unrelated

PR #3472, opened 2025-12-09
refactor accuracy eval script to be organized by hardware

vkuzo:
Summary:

We have a many-to-many mapping of recipe to device it can run on.
Reorganizing the accuracy script to reflect this, since we don't have a
machine that can run all the workflows.

Test Plan:

```
// test on h100
benchmarks/quantization/eval_accuracy_for_readme.sh h100
```

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/.github/scripts/torchao_model_releases/quantize_and_upload.py
+++ b/.github/scripts/torchao_model_releases/quantize_and_upload.py
@@ -8,8 +8,13 @@ import argparse
 from typing import List
 
 import torch
+import transformers
 from huggingface_hub import ModelCard, get_token, whoami
 from transformers import AutoModelForCausalLM, AutoTokenizer, TorchAoConfig
+
+_transformers_version = str(transformers.__version__)
+if _transformers_version >= "5":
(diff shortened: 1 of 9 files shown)


vkuzo:
* #3473
* __->__ #3472

benchmarks/quantization/eval_accuracy_for_readme.sh
```diff
@@ -1,27 +1,95 @@
...
-rm $LOG_FILE
-touch $LOG_FILE
+set -e
 
-QUANT_RECIPES=(
+# Define all available quantization recipes
+QUANT_RECIPES_ALL=(
   # no quantization (baseline)
   "None"
   "float8_rowwise"
```
jainapurva, line 30: Just checking, is this the only float8 recipe we want to showcase?

vkuzo: this is covering all of the server GPU recipes from https://github.com/pytorch/ao/tree/main?tab=readme-ov-file#stable-workflows, and leaving out all other recipes

PR #3473, opened 2025-12-09
make eval script also handle performance measurement

vkuzo:
Summary:

1. refactors the eval script to also handle performance measurement in
   vllm
2. adds a simple `vllm bench latency` script to bench in vllm for prefill and decode

Also, add convenience flags to skip model creation, lm_eval, vllm as
needed to enable running just a single model + single step.

Test Plan:

```
with-proxy ./benchmarks/quantization/measure_accuracy_and_performance.sh h100
// full output: https://www.internalfb.com/phabricator/paste/view/P2094641791
```

Results (on H100):

```
Library Versions:
================================================================================
torch.__version__: 2.9.0+cu128
torch.cuda.get_device_name(): NVIDIA H100
torchao.__version__: 0.14.0+git5c8a14207
vllm.__version__: 0.13.0

Quantization Recipe Results:
================================================================================
+--------------------------+--------------+--------------+--------------+--------------+-----------+----------+-----------+-----------+
| Recipe                   |   Checkpoint |     Wikitext |   Winogrande |   Winogrande |   Prefill |   Decode |   Speedup |   Speedup |
|                          |         (GB) |   Perplexity |          Acc |       Stderr |    toks/s |   toks/s |   Prefill |    Decode |
+==========================+==============+==============+==============+==============+===========+==========+===========+===========+
| None                     |        16.08 |       7.5435 |       0.7419 |       0.0123 |   30946.5 |  6612    |     1     |     1     |
+--------------------------+--------------+--------------+--------------+--------------+-----------+----------+-----------+-----------+
| float8_rowwise           |         9.1  |       7.5919 |       0.7348 |       0.0124 |   45312.5 |  8025.95 |     1.464 |     1.214 |
+--------------------------+--------------+--------------+--------------+--------------+-----------+----------+-----------+-----------+
| int8_rowwise_weight_only |         9.11 |       7.5561 |       0.7427 |       0.0123 |   28231.9 |  4309.8  |     0.912 |     0.652 |
+--------------------------+--------------+--------------+--------------+--------------+-----------+----------+-----------+-----------+
| int8_rowwise             |         9.1  |       7.6567 |       0.738  |       0.0124 |           |          |           |           |
+--------------------------+--------------+--------------+--------------+--------------+-----------+----------+-----------+-----------+

```

Results (on B200):

```
torch.__version__: 2.9.0+cu128
torch.cuda.get_device_name(): NVIDIA B200
torchao.__version__: 0.16.0+git5ad41c539
vllm.__version__: 0.13.0

Quantization Recipe Results:
================================================================================
+----------------+--------------+--------------+--------------+--------------+-----------+----------+-----------+-----------+
| Recipe         |   Checkpoint |     Wikitext |   Winogrande |   Winogrande |   Prefill |   Decode |   Speedup |   Speedup |
|                |         (GB) |   Perplexity |          Acc |       Stderr |    toks/s |   toks/s |   Prefill |    Decode |
+================+==============+==============+==============+==============+===========+==========+===========+===========+
| None           |        16.08 |       7.5435 |       0.7427 |       0.0123 |   59099.9 |  14380   |     1     |     1     |
+----------------+--------------+--------------+--------------+--------------+-----------+----------+-----------+-----------+
| mxfp8          |         9.32 |       7.6034 |       0.7316 |       0.0125 |           |          |           |           |
+----------------+--------------+--------------+--------------+--------------+-----------+----------+-----------+-----------+
| nvfp4          |         6.05 |       8.4459 |       0.7135 |       0.0127 |  102786   |  15218.9 |     1.739 |     1.058 |
+----------------+--------------+--------------+--------------+--------------+-----------+----------+-----------+-----------+
| float8_rowwise |         9.1  |       7.6263 |       0.7348 |       0.0124 |   69313.7 |  15984   |     1.173 |     1.112 |
+----------------+--------------+--------------+--------------+--------------+-----------+----------+-----------+-----------+

```

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo:
* __->__ #3473

namgyu-youn:
@vkuzo Thanks for working on this new benchmark module; this looks much better than the old versions. I have some questions about this work:

1. How about using an academic notation like `W8A16-INT` or `W8A8-INT`, instead of `int8_rowwise_weight_only` or `int8_rowwise`? Abstractions like granularity or HQQ can be added at the end of them I feel.

2. Is there a plan supporting calibration-based apis like AWQ, SmoothQuant, and GPTQ? Those configs might be:

https://github.com/pytorch/ao/blob/415e0e8ad01011367ca6e45b14016250c997352d/.github/scripts/torchao_model_releases/quantize_and_upload.py#L739-L808

vkuzo:
> How about using an academic notation like W8A16-INT or W8A8-INT, instead of int8_rowwise_weight_only or int8_rowwise? Abstractions like granularity or HQQ can be added at the end of them I feel.

I want the solution to the "string to config" problem to be fully decoupled from having good benchmarks, to not block progress on either.  I went with something simple for this PR, if torchao converges to a "string to config" util that everyone is good with we can switch these benchmarks to use it.

I would note that `W8A16-INT` is ambiguous as it does not specify how exactly things are quantized, so we'll need something more specific.

vkuzo:
> Is there a plan supporting calibration-based apis like AWQ, SmoothQuant, and GPTQ?

We should support them eventually!

namgyu-youn:
Thanks for clarify.

btw, I have been engaged implementing Triton kernel for Integer MatMul, which will replace `torch._int_mm()`, because that kernel  failed to generate CUDA graph with lower dimension (similar to: https://gist.github.com/vkuzo/5bf389079442bb9851ef315cdcb797b4). It only happened with `torch.compile()`:

https://github.com/pytorch/ao/blob/27c5eb969f323f08b2870f98045a326771c4b515/torchao/kernel/intmm.py#L46-L52

Just want to make it clear, does Triton kernel for MatMul make sense? I saw this on docs (maybe contribution guide), but not sure yet because: 1)  only a few kernels, 2) even though trition kernel for MatMul exist (https://github.com/pytorch/ao/blob/main/torchao/kernel/intmm_triton.py; pressuming this kernel is outdated), it doesn't used accros ao.

Stonepia:
Hi @vkuzo , May I ask what's the plan of the future running scripts?

For example, for LLama model, I can see the plan is: 
1. benchmarks/_models/eval_hf_models.py  <-- deleted 
2. benchmarks/quantization/measure_accuracy_and_performance.sh <-- Prefer this
3. torchao/_models/llama/benchmarks.sh  <--- Going to be deleted

From the `measure_accuracy_and_performance.sh`, we could only benchmark through `vllm bench`, and the accuracy check goes to `lm_eval`. Seems that the plan is, torchao not responsible for holding any running scripts, just a wrapper, torchao prefer to call tools to reduce efforts.

May I ask the following questions?

1. What is the recommended one for doing the simple performance collection task? 

My concern is that we could reach peak performance by doing some tricks like modifying how the model runs. Like `torch.compile` for example. But if we totally rely on the benchmarking with vllm bench, do we need to move all of these tricks inside vLLM?

2. Current scripts do not take care of the memory constraint GPUs. It gets OOM easily on my local XPU machines on int4-wo models. This may be because a) lm_eval have a copy b) some potential bugs missing. In any way, if we don't keep one script on torchao side, we need to do debugging on lm_eval, is this correct?

Thanks!

vkuzo:
> Seems that the plan is, torchao not responsible for holding any running scripts, just a wrapper, torchao prefer to call tools to reduce efforts.

The motivation for the recent changes is (a) modernize benchmarking for recent changes in the field (such as popularity of vllm) and (b) get away from custom model definitions in torchao (since they are not widely used).  For LLMs, we want to benchmark in vLLM to ensure we use the latest and greatest in the LLM OSS community.   We are also adding a diffusers benchmark in https://github.com/pytorch/ao/pull/3502.

> What is the recommended one for doing the simple performance collection task?

I'd actually recommend the diffusers benchmark from https://github.com/pytorch/ao/pull/3502 once it lands, it's simpler than vllm because this is not an autoregressive model - no prefill vs decode, no k-v cache, etc.

> My concern is that we could reach peak performance by doing some tricks like modifying how the model runs.

Yes, and vLLM has all of those tricks, so if you run the vLLM benchmarks you will capture the tricks :)

> Current scripts do not take care of the memory constraint GPUs. It gets OOM easily on my local XPU machines on int4-wo models.

Improvements to the scripts are welcome! Please feel free to submit a PR, or file an issue if you'd like help from torchao team.

vkuzo:
> It gets OOM easily on my local XPU machines on int4-wo models.

Is this something that worked in the old scripts but does not in the new ones? If so, let's just fix it in the new scripts - let me know more context.

Stonepia:
> Is this something that worked in the old scripts but does not in the new ones? If so, let's just fix it in the new scripts - let me know more context.

Hi @vkuzo , thanks for the reply!
For the OOM, this is not the regression. Should be because of two reasons:
1. The llama/generate.py first load the model, then `to(GPU)`, then quant. This `to(GPU)` will cause OOM.
2. The lm_eval seems maintain the duplicate copy of model (or maybe not delete the model on time), so on the small memory machine, it would cause OOM.

I will also do some more detailed analysis.

Thanks for the help and happy new year~!

xiaowangintel:
Hi, @vkuzo, I have some questions.

I would like to clarify whether measure_accuracy_and_performance.sh is considered a stable and recommended benchmarking method for TorchAO.

Specifically, is this script intended to serve as a long-term reference for measuring both accuracy and performance across different TorchAO features and backends? 

In addition, we would like to understand whether other low-precision data types and quantization methods (e.g., AWQ, GPTQ) are evaluated using the same benchmarking methodology. If not, what are the key differences in evaluation setup, and how should results across different quantization approaches be compared in a consistent and fair manner?

vkuzo:
> I would like to clarify whether measure_accuracy_and_performance.sh is considered a stable and recommended benchmarking method for TorchAO. Specifically, is this script intended to serve as a long-term reference for measuring both accuracy and performance across different TorchAO features and backends?

yes

> In addition, we would like to understand whether other low-precision data types and quantization methods (e.g., AWQ, GPTQ) are evaluated using the same benchmarking methodology. If not, what are the key differences in evaluation setup, and how should results across different quantization approaches be compared in a consistent and fair manner?

https://github.com/pytorch/ao/pull/3602 adds a similar path for calibration based approaches

xiaowangintel:
@vkuzo, thanks for your reply!

PR #3479, opened 2025-12-10
Added permission to test script

avizon-aws:
This PR adds the permission to execute the shell script.

```
git ls-files -s test_mxfp8_allgather.sh        
100755 180375af405d346e260ce5317c695c5a10da48dc 0       test_mxfp8_allgather.sh
```


jerryzh168:
@danielvegamyhre this should be not user facing btw (since tests are not user facing public APIs), all the tags (improvement, bug fix, new feature, docs) by default refers to stable user features like `quantize_`, configs etc. just FYI for future reference.

PR #3486, opened 2025-12-14
Parse device at model release script

namgyu-youn:
Summary:
Add `device` as a parser for custom GPU usage like `cuda:1`. `model_id & quant` parsers are updated to require input.

Test Plan:
python .github/scripts/torchao_model_releases/quantize_and_upload.py  --model_id Qwen/Qwen3-8B  --quant AWQ-INT4

### DIFF
--- a/.github/scripts/torchao_model_releases/quantize_and_upload.py
+++ b/.github/scripts/torchao_model_releases/quantize_and_upload.py
@@ -1003,7 +1003,7 @@ if __name__ == "__main__":
         "--device",
         type=str,
         default="cuda:0",
-        help="Device to run the model on (e.g., 'cuda', 'cuda:0', 'cpu'). Default is 'cuda:0'",
+        help="Device for model loading and quantization (e.g., 'cuda', 'cuda:0', 'cpu'). Default is 'cuda:0'",
     )
     args = parser.parse_args()
     quantize_and_upload(


.github/scripts/torchao_model_releases/quantize_and_upload.py
```diff
@@ -964,7 +975,25 @@ def filter_fn_skip_lmhead(module, fqn):
...
+        type=str,
+        default="cuda:0",
+        help="Device to run the model on (e.g., 'cuda', 'cuda:0', 'cpu'). Default is 'cuda:0'",
+    )
     args = parser.parse_args()
+
+    # Validate CUDA devicedevice
+    try:
+        device = torch.device(args.device)
+        if not torch.cuda.is_available():
```
jerryzh168: why checking cuda here if have a device arg?

namgyu-youn: Sorry it is the wrong check because user might fail to pass with `--device cpu`, this should be removed

jerryzh168:
seems OK, probably we should make "auto" work in the end

.github/scripts/torchao_model_releases/quantize_and_upload.py
```diff
@@ -989,6 +999,12 @@ def filter_fn_skip_lmhead(module, fqn):
         default=False,
         help="Flag to indicate whether push model card to huggingface hub or not",
     )
+    parser.add_argument(
+        "--device",
+        type=str,
+        default="cuda:0",
+        help="Device to run the model on (e.g., 'cuda', 'cuda:0', 'cpu'). Default is 'cuda:0'",
```
jerryzh168: nit: clarify this is device for loading and quantizing the model (running is more for verfication/sanity check)

PR #3488, opened 2025-12-15
[float8] fix the "Cannot set version_counter for inference tensor" for transpose

### LINKED ISSUE #3487: Cannot set version_counter for inference tensor
# Reproduer

When I tried to run the following:

```Python
import torch
import torch.nn as nn
from torchao.quantization import quantize_, Float8DynamicActivationFloat8WeightConfig, PerTensor

device = torch.device("cuda")
#device = torch.device("xpu")

# Create a simple linear layer
linear = nn.Linear(32, 48, bias=True).to(device).to(torch.bfloat16).eval()
linear.requires_grad_(False)

print("Quantizing with version=2...")
quantize_(
    linear,
    Float8DynamicActivationFloat8WeightConfig(granularity=PerTensor(), version=2),
    device=device,
)

# Create input
input_tensor = torch.randn(16, 32, dtype=torch.bfloat16, device=device)

# Test forward pass inside inference_mode
print("\nTesting forward pass inside inference_mode...")
with torch.inference_mode():
    output = linear(input_tensor)

```

I got the following error on both CUDA/XPU:

```
Testing forward pass inside inference_mode...
  File "/root/tongsu/test_full_flow.py", line 26, in <module>
    output = linear(input_tensor)
  File "/root/miniforge3/envs/tongsu_1212/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1776, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/root/miniforge3/envs/tongsu_1212/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1787, in _call_impl
    return forward_call(*args, **kwargs)
  File "/root/miniforge3/envs/tongsu_1212/lib/python3.10/site-packages/torch/nn/modules/linear.py", line 134, in forward
    return F.linear(input, self.weight, self.bias)
  File "/root/tongsu/ao/torchao/utils.py", line 632, in _dispatch__torch_function__
    return cls._TORCH_FN_TABLE[cls][func](func, types, args, kwargs)
  File "/root/tongsu/ao/torchao/utils.py", line 435, in wrapper
    return _func(f, types, args, kwargs)
  File "/root/tongsu/ao/torchao/quantization/quantize_/workflows/float8/float8_tensor.py", line 266, in _
    return _float8_addmm_impl(input_tensor, weight_tensor.t(), bias)
  File "/root/tongsu/ao/torchao/utils.py", line 634, in _dispatch__torch_function__
    return func(*args, **kwargs)
RuntimeError: Cannot set version_counter for inference tensor
```

This supposed to be because I should put the tensor creation and usage all inside the `with torch.inference_mode():`

```Python
# This works correctly, which put everything inside the with context

with torch.inference_mode():
   linear = nn.Linear(32, 48, bias=True).to(device).to(torch.bfloat16).eval()
   ...
```

However, maybe we could do something further to make the initialization more user friendly?

# Env

```
torch: Version: 2.10.0.dev20251210+cu128
torchao: Version: 0.16.0+gitff6d9e244
```

# Problem Analysis

As the error message suggests, the failure happened on this:
```python
#float8_tensor.py
@implements(aten.linear.default)
@implements_torch_function(torch.nn.functional.linear)
...

return _float8_addmm_impl(input_tensor, weight_tensor.t(), bias) # <---- This weight_tensor.t()
```
It actually happened on the `weight_tensor.t()`. 

As the error message suggests, should be in the dispatching process in:

```
  File "/root/.../ao/torchao/utils.py", line 634, in _dispatch__torch_function__
    return func(*args, **kwargs)
```

So the quick solution is to avoid this dispatching by register the `implements_torch_function(torch.Tensor.t)` directly:

# Possible Fix
In torchao, there is a fix by registering like below. I created a PR at #3488 , but I don't know if this is a proper fix

```diff
@implements(aten.t.default)
def _(func, types, args, kwargs):
...

+@implements_torch_function(torch.Tensor.t)
+def _(func, types, args, kwargs):
+    print("DEBUG: implements_torch_function(torch.Tensor.t)")
+    assert len(args) == 1
+    self = args[0]
+    assert len(self.block_size) == 2
+    new_tensor = self.__class__(
+        self.qdata.t(),
+        self.scale.t(),
+        (self.block_size[1], self.block_size[0]),
+        self.mm_config,
+        self.act_quant_kwargs,
+        self.kernel_preference,
+        self.dtype,
+    )
+    return new_tensor
```

This output is like below:

```
DEBUG: implements_torch_function(torch.Tensor.t)
```

So the difference is (by asking GPT):

This is the success path:
```
Python: weight_tensor.t()
    ↓
__torch_function__ intercepts
    ↓
Your handler runs, returns new_tensor (is_inference=True)
    ↓
Returns DIRECTLY to Python caller ← No C++ ADInplaceOrView involved!
    ↓
SUCCESS
```

This is the failure path:

```
Python: weight_tensor.t()
    ↓
__torch_function__ - NOT registered, falls through
    ↓
func(*args, **kwargs) with DisableTorchFunctionSubclass
    ↓
C++ dispatch starts
    ↓
ADInplaceOrView::t() runs ← This is the problem!
    ↓
Calls at::_ops::t::redispatch() → eventually hits __torch_dispatch__
    ↓
Your @implements(aten.t.default) handler runs, returns new_tensor (is_inference=True)
    ↓
Returns to ADInplaceOrView::t()
    ↓
ADInplaceOrView::t() calls as_view(base, output) on YOUR returned tensor
    ↓
as_view() calls set_version_counter() on output (the inference tensor)
    ↓
ERROR: Cannot set version_counter for inference tensor
```

Stonepia:
Fix #3487

We got `Cannot set version_counter for inference tensor` when handling on float8 tensors. The root cause is that the torch `return_and_correct_aliasing` sets version counters, so this PR adds the direct dispatch path, to avoid going to the tensor_dispatch logic.

### DIFF
--- a/torchao/quantization/quantize_/workflows/float8/float8_tensor.py
+++ b/torchao/quantization/quantize_/workflows/float8/float8_tensor.py
@@ -763,14 +763,19 @@ def _(func, types, args, kwargs):
     for i in range(len(self.block_size)):
         block_size[i] = min(block_size[i], sliced_data.shape[i])
 
-    return Float8Tensor(
-        sliced_data,
-        sliced_scale,
-        block_size,
-        self.mm_config,
-        self.act_quant_kwargs,
-        self.kernel_preference,
-        dtype=self.dtype,
+    return return_and_correct_aliasing(
+        func,
+        args,
+        kwargs,
+        Float8Tensor(
+            sliced_data,
+            sliced_scale,
+            block_size,
+            self.mm_config,
+            self.act_quant_kwargs,
+            self.kernel_preference,
+            dtype=self.dtype,
+        ),
     )
 
 


jerryzh168:
@Stonepia can you add a test, I confirmed this fixes the issue you posted: https://github.com/pytorch/pytorch/issues/170419

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
```diff
@@ -965,6 +965,23 @@ def _(func, types, args, kwargs):
...
+    new_tensor = self.__class__(
+        self.qdata.t(),
+        self.scale.t(),
+        (self.block_size[1], self.block_size[0]),
+        self.mm_config,
+        self.act_quant_kwargs,
+        self.kernel_preference,
+        self.dtype,
+    )
+    return new_tensor
```
bdhirsh, line 1002: This will sever the autograd graph if you call `y = float8_tensor.t()`

jerryzh168: is this true for all the other ops implemented for Float8Tensor in this file?

bdhirsh: from a quick audit of the file - I think the answer is no, because this is the only op that gets a torch_function handler, *and* returns the output through a call to `new_tensor = self.__class__(...)`. (for __torch_dispatch__ handlers this is fine because __torch_dispatch__ runs below autograd, but not torch_function

Stonepia: May I ask, is using the `implements_torch_function` the best way?

Because a few lines above, it has the code that implements `aten.t`. I think this might be the default behavior by design (although it gets wrong):

```python
@implements(aten.t.default)
...
```
The `@implements_torch_function(torch.Tensor.t)` will override the above behavior.

Stonepia:
> @Stonepia can you add a test, I confirmed this fixes the issue you posted: [pytorch/pytorch#170419](https://github.com/pytorch/pytorch/issues/170419)

Hi @jerryzh168 , thanks for the help on this issue! I put the two reported tests in https://github.com/pytorch/ao/pull/3488/commits/d1fcd08918bc586c8a9bb58c196305795c104301

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
jerryzh168: OK, since this is only for inference, autograd is not a concern for now

@Stonepia we need to use `implements_torch_function` for torch.Tensor.t since it's a torch function, not aten function, `implements` is reserved for aten functions only

jerryzh168: @Stonepia one thing I'm wondering is whether we need `return_and_correct_aliasing` here as well, like other functions in the file

Edit: nvm, I just tested, this is not needed for `implements_torch_function`

jerryzh168:
marking this as ready for review, we would like to merge this ASAP to unblock pytorch release

Stonepia:
Hi @jerryzh168 , the `test_tensor_slice_in_inference_mode()` seems can't be passed. So I added another PR in https://github.com/pytorch/ao/pull/3590 to make this pr to have less code. May I ask a favor for reviewing?

Stonepia:
The CI failure seems not related to this PR. Thanks for the help!

[Run Regression Tests / test-nightly (CPU Nightly, linux.4xlarge, --pre torch --index-url https://download.pytorch.org/wh... / linux-job](https://hud.pytorch.org/pr/pytorch/ao/3488#59835380361) ([gh](https://github.com/pytorch/ao/actions/runs/20776016902/job/59835380361))
test/quantization/pt2e/test_x86inductor_fusion.py::TestPatternMatcher::test_qconv2d_silu_int8_mixed_bf16_cpu

jerryzh168:
@Stonepia please wait a bit, I'm also exploring just fixing this in vllm directly (adding inference_mode() during quantization)

jerryzh168:
I tried with adding inference_mode in vllm for weight loading, but seems that this is still needed, I think we can merge this first

Issue #3490, opened 2025-12-15
FqnToConfig doesn't handle module swaps


andrewor14:
A user tried to use `FqnToConfig` with `QATConfig`, which uses module swap. However, `FqnToConfig` seems to only work with tensor subclasses, not module swaps:
https://github.com/pytorch/ao/blob/ff6d9e24421cf4b1e98a2ae139144adbc038e6f8/torchao/quantization/quant_api.py#L485-L491

Looks like we do need to reassign for module swaps.

**Minimal repro:**

```
import torch
from torchao.quantization import quantize_, FqnToConfig, Int4WeightOnlyConfig
from torchao.quantization.qat import QATConfig
from torchao.prototype.mx_formats import NVFP4DynamicActivationNVFP4WeightConfig

class M(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.l = torch.nn.Linear(4096, 4096)
    def forward(self, x):
        return self.l(x)

m = M()
print("BEFORE:", type(m.l), m.l)
base_config = NVFP4DynamicActivationNVFP4WeightConfig()
name_to_config = {"l": QATConfig(base_config, step="prepare")}
quantize_(m, FqnToConfig(name_to_config), filter_fn=None) # doesn't work
print("AFTER:", type(m.l), m.l)
```
Output: (module was not swapped)
```
BEFORE: <class 'torch.nn.modules.linear.Linear'> Linear(in_features=4096, out_features=4096, bias=True)
AFTER: <class 'torch.nn.modules.linear.Linear'> Linear(in_features=4096, out_features=4096, bias=True)
```

andrewor14:
cc @jcaip

PR #3492, opened 2025-12-15
Make FqnToConfig handle module swap configs

andrewor14:
**Summary:** `FqnToConfig` does not handle module swap configs like `QATConfig` today because it does not reassign the replaced modules, but instead assumes there is no need to do so because the underlying tensors are swapped. This breaks when the user tries to use `FqnToConfig` with `QATConfig`, which does not rely on tensor subclasses. This commit fixes this by making the replacement logic in `FqnToConfig` more general.

Fixes https://github.com/pytorch/ao/issues/3490.

**Test Plan:**
```
python test/quantization/test_quant_api.py -k test_fqn_config_quantized_nested_module_module_swap
```

### DIFF
--- a/torchao/quantization/quant_api.py
+++ b/torchao/quantization/quant_api.py
@@ -480,19 +480,23 @@ def quantize_(
             raise ValueError(
                 "Custom filter_fn and FqnToConfig were both specified. Only filter_fn=None is supported when FqnToConfig is specified."
             )
-
-        for module_fqn, module in model.named_modules():
+        named_modules = dict(model.named_modules())
+        for module_fqn, module in named_modules.items():
             if (
                 fqn_matches_fqn_config(module_fqn, config)
                 or _module_param_matches_fqn_config(module, module_fqn, config)
                 or ("_default" in config.fqn_to_config and _is_linear(module))
             ):
-                # this replaces inplace, so no need to reassign
-                _fqn_to_config_handler(module, module_fqn, config)
+                replacement = _fqn_to_config_handler(module, module_fqn, config)
                 if device is not None:
-                    module.to(device=device)
-        return
-    if isinstance(config, AOBaseConfig):
+                    replacement = replacement.to(device=device)
+                # handle module swap
+                if replacement is not module and module_fqn != "":
+                    child_name = module_fqn.split(".")[-1]
+                    parent_fqn = module_fqn.removesuffix(child_name).removesuffix(".")
+                    parent_module = named_modules[parent_fqn]
+                    setattr(parent_module, child_name, replacement)
+    elif isinstance(config, AOBaseConfig):
         filter_fn = _is_linear if filter_fn is None else filter_fn
         handler = _QUANTIZE_CONFIG_HANDLER[type(config)]
         # for each linear in the model, apply the transform if filtering passes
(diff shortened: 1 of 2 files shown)


torchao/quantization/quant_api.py
```diff
@@ -480,19 +517,8 @@ def quantize_(
...
-                or _module_param_matches_fqn_config(module, module_fqn, config)
-                or ("_default" in config.fqn_to_config and _is_linear(module))
-            ):
-                # this replaces inplace, so no need to reassign
-                _fqn_to_config_handler(module, module_fqn, config)
-                if device is not None:
-                    module.to(device=device)
-        return
-    if isinstance(config, AOBaseConfig):
+        _fqn_to_config_replace(config, model, device=device)
```
vkuzo: this is hard to follow, would it be cleaner to use iteration instead of recursion here

torchao/quantization/quant_api.py
```diff
@@ -480,19 +517,8 @@ def quantize_(
...
             )
-
-        for module_fqn, module in model.named_modules():
-            if (
-                fqn_matches_fqn_config(module_fqn, config)
-                or _module_param_matches_fqn_config(module, module_fqn, config)
-                or ("_default" in config.fqn_to_config and _is_linear(module))
-            ):
-                # this replaces inplace, so no need to reassign
-                _fqn_to_config_handler(module, module_fqn, config)
```
jcaip, line 491: Maybe we can implement something like
```python
def set_module_by_fqn(model, fqn, new_module):
    # set model[fqn] to new_module
```

so then we can keep it iterative? Only thing is that we have to get the modules in top-down order (this does [appear](https://github.com/pytorch/pytorch/blob/d38164a545b4a4e4e0cf73ce67173f70574890b6/torch/nn/modules/module.py#L2823) to be how named_modules is implemented) 
```python
    for module_fqn, module in model.named_modules():
            if (
                fqn_matches_fqn_config(module_fqn, config)
                or _module_param_matches_fqn_config(module, module_fqn, config)
                or ("_default" in config.fqn_to_config and _is_linear(module))
            ):
                new_module = _fqn_to_config_handler(module, module_fqn, config)
                # reassign module
                set_module_by_fqn(model, fqn, new_module) 
```

andrewor14:
Rewrote using iteration, check again? @vkuzo @jcaip

vkuzo (approved):

torchao/quantization/quant_api.py
```diff
@@ -480,19 +480,23 @@ def quantize_(
...
-                    module.to(device=device)
-        return
-    if isinstance(config, AOBaseConfig):
+                    replacement = replacement.to(device=device)
+                # handle module swap
+                if replacement is not module and module_fqn != "":
+                    child_name = module_fqn.split(".")[-1]
+                    parent_fqn = module_fqn.removesuffix(child_name).removesuffix(".")
+                    parent_module = named_modules[parent_fqn]
+                    setattr(parent_module, child_name, replacement)
```
jcaip, line 498: I think we're missing a return here at the end

andrewor14: I changed the next case to elif, is that OK?

jcaip: oh missed that, yeah lgtm

andrewor14:
Test failure doesn't seem related, merging this. Thanks

PR #3494, opened 2025-12-16
[ZENDNN] Adding support for slice in opaque

Priyjain-amd:
Add support `Int4OpaqueTensor` from torchao in vLLM's LLaMA model implementation.
These changes enable vLLM to load models quantized with `torchao` using `int4_packing_format="opaque"` without errors during weight loading.

Encountered error: NotImplementedError: Int4OpaqueTensor dispatch: attempting to run unimplemented operator/function: 
func=<OpOverload(op='aten.slice', overload='Tensor')>

Issue #3499

### DIFF
--- a/torchao/core/config.py
+++ b/torchao/core/config.py
@@ -200,6 +200,7 @@ ALLOWED_AO_MODULES = {
     "torchao.prototype.parq.quant",
     "torchao.quantization.quantize_.common",
     "torchao.quantization.quantize_.workflows",
+    "torchao.prototype.quantization.int4",
 }
 
 
--- a/torchao/core/config.py
+++ b/torchao/core/config.py
@@ -203,6 +203,35 @@ ALLOWED_AO_MODULES = {
 }
 
 
+def _register_int4_module() -> None:
+    module_spec = importlib.util.find_spec("torchao.prototype.quantization.int4")
+    if module_spec is None:
+        module_spec = importlib.util.find_spec("torchao.prototype.int4_opaque_tensor")
+    if module_spec is None:
+        return
+
+    int4_module = importlib.import_module(module_spec.name)
+    if not hasattr(int4_module, "Int4WeightOnlyOpaqueTensorConfig") and hasattr(
+        int4_module, "PrototypeInt4WeightOnlyConfig"
+    ):
+        int4_module.Int4WeightOnlyOpaqueTensorConfig = (
+            int4_module.PrototypeInt4WeightOnlyConfig
+        )
+    ALLOWED_AO_MODULES.add(module_spec.name)
+
+    try:
+        from torchao.prototype.safetensors.safetensors_utils import (
+            ALLOWED_CLASSES,
+            ALLOWED_TENSORS_SUBCLASSES,
+        )
+
+        if "Int4OpaqueTensor" not in ALLOWED_TENSORS_SUBCLASSES:
+            ALLOWED_TENSORS_SUBCLASSES.append("Int4OpaqueTensor")
+        ALLOWED_CLASSES["Int4OpaqueTensor"] = int4_module.Int4OpaqueTensor
+    except (ImportError, AttributeError):
+        pass
+
+
 def config_from_dict(data: Dict[str, Any]) -> AOBaseConfig:
     """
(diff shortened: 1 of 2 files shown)


torchao/prototype/int4_opaque_tensor/int4_opaque_tensor.py
```diff
@@ -242,6 +242,69 @@ def _(func, types, args, kwargs):
...
+    scale_and_zero = aten.slice(self.scale_and_zero, 1 - dim, start_sz, end_sz, step)
+
+    # Calculate new shape after slicing
+    new_shape = list(self.shape)
+    new_shape[dim] = end - start
+
+    block_size = list(self.block_size)
+    block_size[dim] = min(block_size[dim], new_shape[dim])
+
+    return Int4TilePackedTo4dTensor(
```
jerryzh168: Int4OpqueTensor?

Priyjain-amd: I accidentally committed the incorrect changes; they’re resolved now.

Xia-Weiwen:
The `Int4OpqueTensor` does not support slice I think. The tensor data are no longer arranged as [N, K/2] in memory after packing. And if you unpack the tensor before slice, you won't be able to pack it again since the slice you get is not contiguous in memory. So, basically if you want to do a slice operation, you will probably need to use a plain tensor rather than `Int4OpqueTensor`.
If you need slice for tensor parallel (TP), I guess you need to split tensors to each rank before packing and pack each slice separately on each rank.

aadwived:
Hello @Xia-Weiwen ,

**Requirement of the Slice op**
We need the slice op support for `Int4OpaqueTensor` because we want to run quantized TorchAO models with vLLM.
During vLLM weight loading, the loader calls `param_data.narrow(...)` to split fused weights (e.g., QKV). `narrow `lowers to `aten.slice.Tensor`, and without a slice implementation on `Int4OpaqueTensor` we hit:

```
... linear.py: weight_loader -> param_data = param_data.narrow(...)
    ... torchao/utils.py dispatch -> NotImplementedError: Int4OpaqueTensor dispatch ... aten.slice.Tensor 
```
This prevents the model from loading/executing.

**Contiguity and Repack-ability of Tensors**
The slicing needs are limited, and we can design the op with safeguards rather than making the opaque format generally slice-able:
  

-   `step == 1` only (matches `narrow `usage).
-   `dim == 0` (N / out_features) slicing is supported: it corresponds to taking contiguous row ranges, and we slice `qdata `(dim 0) + `scale_and_zero `(N dimension) consistently.
-   `dim == 1` (K / in_features) slicing is not supported, except for identity mapping (`start=0, end=K`), since arbitrary K slicing would not preserve the packed layout/contiguity assumptions.

If you are satisfied with the response, we can re-configure the op implementation accordingly. Please let me know if you have any more concerns.

**Question**
Does TorchAO have plans (or a preferred approach) to ensure tensor subclasses / quantization configs are compatible with vLLM’s loader expectations (which currently relies on narrow/slice during load), or should this compatibility be handled within TorchAO tensor subclasses as we’re proposing here?

Thanks,
Aakar Dwivedi

CC : @Priyjain-amd and @jerryzh168

aadwived:
Hello @Xia-Weiwen ,

Can you please go through the comment above and let me know your opinion?

Xia-Weiwen:
Hi @aadwived The concern is whether the implementation is correct or not. The int4 opaque tensor is in a packed layout in memory. It's data is rearranged to 3D (N is tiled), rather than 2D. You may check out `_convert_weight_to_int4pack_for_cpu` in PyTorch source for more details: https://github.com/pytorch/pytorch/blob/c926d4534966f289400a44932e3a2c1e610a2832/aten/src/ATen/native/cpu/int4mm_kernel.cpp#L579
So, calling `aten.slice` directly on the tensor data along N can be incorrect. Two methods I came up with are:
1. Check N dimensions before slicing. Slicing a N tile is not allowed.
2. Unpack the opaque tensor, slice it, then pack the result again.

Given that most models have an N that is a multiple of 64, method 1 looks simpler. Does that meet your needs?

PR #3498, opened 2025-12-17
[mxfp8] support RCEIL in triton_to_mxfp8_dim0 kernel with inline PTX

danielvegamyhre:
## Summary
- Support RCEIL in triton_to_mxfp8_dim0 kernel with inline PTX, for consistency with other kernels. We are moving toward RCEIL as the sensible default for accuracy + perf. 
- After this change, for MoE training code, all quantization ops have an option that supports RCEIL.

Reference: https://github.com/pytorch/ao/blob/85557135c93d3429320a4a360c0ee9cb49f84a00/torchao/csrc/cuda/mx_kernels/mxfp8_quantize.cuh#L211

## Tests
- `pytest test/prototype/mx_formats/test_kernels.py -k dim0`
- `pytest pytest test/prototype/mx_formats/test_mx_linear.py`

## Benchmarks
```
[danvm@devgpu031.atn1 ~/ao (rceil2)]$ cd /home/danvm/ao && PYTHONPATH=/home/danvm/ao:$PYTHONPATH conda run -n torch2 CUDA_VISIBLE_DEVICES=7 python benchmarks/mx_formats/cast_bench.py --mode dim0_mxfp8_triton_rceil
M 16384 K 16384 BLOCK_SIZE 32
GPU: NVIDIA B200
torch version: 2.11.0.dev20251216+cu128
triton version: 3.6.0
mode: dim0_mxfp8_triton_rceil
time_us 139.45600390434265
mem_bw_gbps 5834.779093183679
```

### DIFF
--- a/test/prototype/mx_formats/test_mx_linear.py
+++ b/test/prototype/mx_formats/test_mx_linear.py
@@ -125,14 +125,9 @@ def test_linear_eager_vs_hp(
             ScaleCalculationMode.FLOOR,
             ScaleCalculationMode.RCEIL,
         ):
-            pytest.skip("unsupported configuration")
-        elif (
-            scale_calculation_mode == ScaleCalculationMode.RCEIL
-            and not is_sm_at_least_100()
-        ):
-            pytest.skip(
-                "triton dim1 mxfp8 quantization kernel requires sm100 for RCEIL scaling mode"
-            )
+            pytest.skip("triton mxfp8 quantization kernels only require sm100")
+        if not is_sm_at_least_100():
+            pytest.skip("triton mxfp8 quantization kernels require sm100")
     elif mxfp8_cast_kernel_choice == MXFP8Dim1CastKernelChoice.CUDA:
         if scale_calculation_mode not in (
             ScaleCalculationMode.FLOOR,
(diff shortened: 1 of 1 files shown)


torchao/prototype/mx_formats/kernels.py
```diff
@@ -179,23 +179,46 @@ def _triton_calculate_scale(x, axis):
...
+            # Casting uint16 to uint8 naturally truncates to the low byte.
+            scale_e8m0_biased = tl.inline_asm_elementwise(
+                asm="cvt.rp.satfinite.ue8m0x2.f32 $0, 0.0, $1;",
+                constraints="=h,r",
+                args=[scale_input.to(tl.float32, bitcast=False)],
+                dtype=tl.uint16,
+                is_pure=True,
+                pack=1,
+            ).to(tl.uint8)
+        else:
```
drisspg, line 202: static assert the scaling mode is whats expected

danielvegamyhre: mode is a tl.constexpr so it should be known at compile time, what am i missing?

drisspg: Static assert in the else path that the string is what you expect

danielvegamyhre: ah I see what you mean - done

danielvegamyhre:
@drisspg CI is green now if you want to take another look (only failing test is unrelated).

I had to require sm100 for some tests, namely triton dim0 and dim1 kernels, for both rceil and floor, since they share the "calculate scale" triton jit func and just compiling it at all with the inline PTX requires sm100 now

PR #3500, opened 2025-12-17
[xpu] INT8 quantization on Intel XPU

zxd1997066:
# Motivation
This feature is one of features listed in the #3576. There is the XMX intrinsic on the Intel GPUs, e.g., Intel Arc B580, to speedup the int8 computation. This PR is target to ensure the int8 quantization works on the XPU and user can use same quantization config with CUDA. 
# Implementation

# Usage
Users can use same quantization config, such as Int8StaticActivationInt8WeightConfig or Int8DynamicActivationInt8WeightConfig, as CUDA to run the fp8 models on XPU. This device dispatched by the PyTorch op.
```

#For static int8 quantization
            granularity=granularity,
        )
```

### DIFF
--- a/test/quantization/quantize_/workflows/int8/test_int8_tensor.py
+++ b/test/quantization/quantize_/workflows/int8/test_int8_tensor.py
@@ -28,7 +28,7 @@ from torchao.quantization.quantize_.workflows.int8.int8_tensor import (
 )
 from torchao.quantization.utils import compute_error, get_block_size
 from torchao.testing.model_architectures import ToyTwoLinearModel
-from torchao.testing.utils import TorchAOIntegrationTestCase
+from torchao.testing.utils import TorchAOIntegrationTestCase, skip_if_xpu
 from torchao.utils import (
     get_available_devices,
     get_current_accelerator_device,
@@ -234,7 +234,7 @@ class TestInt8Tensor(TorchAOIntegrationTestCase):
             "extern_kernels._int_mm", 1
         ).check_count("triton_poi_fused", 1).run(code[0])
 
-    @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
+    @skip_if_xpu("XPU enablement in progress")
     @common_utils.parametrize("config", INT8_TEST_CONFIGS)
     def test_pin_memory(self, config):
         linear = torch.nn.Linear(


zxd1997066:
cc @liangan1

liangan1 (approved):

liangan1:
@xiaowangintel  pls refine the pr desc. You should highlight:

1. The related op has been enabled in the pytorch core, pls add the related pr link. 
2.  For the end user, what is the E2E quantization config on XPU? 

pls refer to https://github.com/pytorch/ao/pull/3332

liangan1:
@zxd1997066  pined memory is N/A to XPU. pls skip these failed cases on XPU related to pined memory.

test/dtypes/test_nf4.py
```diff
@@ -757,7 +757,7 @@ def world_size(self) -> int:
         return 2
 
     @skip_if_lt_x_gpu(2)
-    @unittest.skipIf(not torch.accelerator.is_available(), "Need GPU available")
+    @unittest.skipIf(not torch.cuda.is_available(), "Need GPU available")
```
liangan1: nf4 is not related to this PR. pls remove it and submit it in another pr.

zxd1997066: done

vkuzo:
what does performance/accuracy look like

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -60,7 +61,7 @@ def test_creation_and_attributes(self, config):
             self.test_shape[0],
             bias=False,
             dtype=self.dtype,
-            device="cuda",
+            device=_DEVICE,
```
jerryzh168: what happens when _DEVICE is None?

jerryzh168: I feel a safer way might be to do `_DEVICE = get_current_accelerator_device()` inside the test function since it's guarded by `@unittest.skipIf(not torch.accelerator.is_available(), "Need GPU available")`

also we could think of just removing the None branch for `get_current_accelerator_device` for simplicity and not having to deal with the None case (error out for None)

xiaowangintel:
> what does performance/accuracy look like

@vkuzo  Thanks for your comments. 
For accuracy details as follows:

meta-llama/Llama-3.1-8B-Instruct on Intel GPU and A100 using torchao int8-dynamic-quantization. The model configs as below:
Precision : torch.bfloat16
quantization configuration : Int8DynamicActivationInt8WeightConfig
dataset : wikitext
Result:
The perplexity res:

> Intel GPU: 9.583
> A100: 9.578

We can not provide the performance data directly here due to the internal legal process. But we will provide it once the benchmark tests are ready on XPU and will update the readme in another PR.

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
zxd1997066: Thanks, Removed None branch for ```get_current_accelerator_device``` https://github.com/pytorch/ao/pull/3500/files#diff-d2a11602a79e83305208472f1abe6a4106f02ce62a7f9524007181813863fcf6L158

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -38,9 +38,10 @@
         version=2, granularity=PerRow(), act_mapping_type=MappingType.SYMMETRIC
     ),
 ]
+_DEVICE = get_current_accelerator_device()
```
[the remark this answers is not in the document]
vkuzo: it's confusing to have this as a global as cpu is always available and it's no longer clear whether this is cpu or cuda/xpu/etc

vkuzo (changes requested):

torchao/utils.py
```diff
@@ -155,8 +155,6 @@ def get_available_devices():
 def get_current_accelerator_device():
     if torch.accelerator.is_available():
         return torch.accelerator.current_accelerator()
-    else:
```
jerryzh168: it will still return None by default, I think there should be an assert here, it might break some tests so probably better to do in a separate PR

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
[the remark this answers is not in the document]
zxd1997066: Thanks for your review, moved ```_DEVICE = get_current_accelerator_device()``` inside the test function, because of  ```@unittest.skipIf(not torch.accelerator.is_available(), "Need GPU available")```, it will not be cpu

torchao/utils.py
zxd1997066: Thanks, will do that in https://github.com/pytorch/ao/pull/3634

liangan1:
@vkuzo according to the semantic of pytorch, cpu is not the accelerator device.

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
liangan1: @zxd1997066  I suppose we should use _DEVICE = get_current_accelerator_device() inside the test function since it's guarded by @unittest.skipIf(not torch.accelerator.is_available() and remove the NONE of get_current_accelerator_device  in another PR.

zxd1997066: Yes, I already use _DEVICE = get_current_accelerator_device() inside the test function in this PR and submit another PR https://github.com/pytorch/ao/pull/3634 to remove the NONE of get_current_accelerator_device

liangan1:
@jerryzh168  @vkuzo  We have refined it according to comments. pls help to review again. Thanks.

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -128,7 +129,7 @@ def test_int8_linear_variants(
         )
 
     @common_utils.parametrize("config", INT8_TEST_CONFIGS)
-    @common_utils.parametrize("device", ["cpu", "cuda"])
+    @common_utils.parametrize("device", ["cpu", _DEVICE])
```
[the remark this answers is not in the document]
vkuzo: it's confusing to use persistent state for available devices.  How about something like this?

```python
def get_available_devices() -> List[torch.device]:
    ...

@common_utils.parametrize("device", get_available_devices())
    ...
    # no persistent state to track, device is passed in to the test args, logic is simple
    x = torch.randn(..., device=device)
```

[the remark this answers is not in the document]
zxd1997066: Thanks for your suggestion, updated.

torchao/utils.py
```diff
@@ -153,8 +153,15 @@ def get_available_devices():
 
 
 def get_current_accelerator_device():
+<<<<<<< HEAD
```
jerryzh168: this is still here

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -224,6 +225,7 @@ def test_available_gpu_kernels(self):
             "extern_kernels._int_mm", 1
         ).check_count("triton_poi_fused", 1).run(code[0])
 
+    @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
```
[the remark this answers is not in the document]
jerryzh168: is this because only cuda is supported for the pin_memory test?

torchao/utils.py
zxd1997066: missed during rebase, just removed it.

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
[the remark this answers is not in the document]
zxd1997066: refined the skip for XPU only.

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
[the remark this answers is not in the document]
zxd1997066: Hi @vkuzo , I already updated to @common_utils.parametrize("device", get_available_devices()) https://github.com/pytorch/ao/pull/3500/changes#diff-76d4f5ccf412b0c545cc447dced3b9235139c14230c9fc9c7efa5dfb3fd9c298R139, anything else also need changes?

cc @jerryzh168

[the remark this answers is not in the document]
jerryzh168: please remove `self._DEVICE ` and `_DEVICE` in the file and replace with the same pattern everywhere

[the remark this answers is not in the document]
zxd1997066: Thanks, will update for this.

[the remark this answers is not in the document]
zxd1997066: seems get_current_accelerator_device() can not be used as ```@common_utils.parametrize("device", [get_current_accelerator_device()])```, as it will raise AssertionError on cpu. Removed ```self._DEVICE```  and ```_DEVICE```, and defined ```device``` inside the test function https://github.com/pytorch/ao/pull/3782

PR #3501, opened 2025-12-17
Fix NVFP4 QAT mixed precision

andrewor14:
**Summary:** This commit adds support for bf16 activations + fp32 weights mixed precision for NVFP4 QAT, which previously threw a dtype assertion error:
```
File "ao/torchao/prototype/qat/nvfp4.py", line 159, in forward
  assert fq.dtype == x.dtype
```

**Test Plan:**
```
python test/quantization/test_qat.py -k test_nvfp4_fake_quanitzed_linear_mixed_precision
```

### DIFF
--- a/torchao/prototype/qat/nvfp4.py
+++ b/torchao/prototype/qat/nvfp4.py
@@ -39,6 +39,7 @@ class _NVFP4QuantizedForwardFakeQuantizedBackward(torch.autograd.Function):
     """
 
     @staticmethod
+    @torch.amp.custom_fwd(device_type="cuda")
     def forward(
         ctx,
         _input: torch.Tensor,
@@ -87,6 +88,7 @@ class _NVFP4QuantizedForwardFakeQuantizedBackward(torch.autograd.Function):
         )
 
     @staticmethod
+    @torch.amp.custom_bwd(device_type="cuda")
     def backward(ctx, grad_output: torch.Tensor) -> torch.Tensor:
         _input, weight = ctx.saved_tensors
         assert isinstance(_input, NVFP4Tensor)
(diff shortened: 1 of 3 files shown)


torchao/prototype/qat/nvfp4.py
```diff
@@ -156,7 +156,7 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
         fq = _NVFP4QuantizedForwardFakeQuantizedBackward.apply(
```
vkuzo, line 158: should the cast be inside `_NVFP4QuantizedForwardFakeQuantizedBackward`?

andrewor14: moved it in

torchao/prototype/qat/nvfp4.py
vkuzo, line 83: why is this returning the wrong dtype?

andrewor14: Seems to be caused by adding the bias:

https://github.com/pytorch/ao/blob/1f9bfd7ebf054c616543052c5767a8e6972cbdf8/torchao/prototype/mx_formats/nvfp4_tensor.py#L495

Before this line result was bf16, after this line it's fp32. Do you think we should cast the bias here instead?

torchao/prototype/qat/nvfp4.py
```diff
@@ -39,6 +39,7 @@ class _NVFP4QuantizedForwardFakeQuantizedBackward(torch.autograd.Function):
     """
 
     @staticmethod
+    @torch.amp.custom_fwd(device_type="cuda")
```
andrewor14, line 42: https://docs.pytorch.org/docs/stable/notes/amp_examples.html#functions-with-multiple-inputs-or-autocastable-ops

vkuzo: is this still needed after the change in nvfp4 tensor?

andrewor14: Yep, that one is just for fp32 inputs + fp32 weights during inference, this is for mixed precision during training

PR #3514, opened 2025-12-18
Remove dType Field

syedshazli:
In this file, there are multiple comments in the code that say 'remove the dType field as it is not used.' However, the dType field is still instantiated as None in the class.

This PR aims to remove the dType field during initialization of any modules that may have previously referenced it.

### DIFF
--- a/torchao/quantization/linear_quant_modules.py
+++ b/torchao/quantization/linear_quant_modules.py
@@ -143,7 +143,7 @@ class WeightOnlyInt4Linear(torch.nn.Module):
             "scales_and_zeros",
             torch.zeros(
                 (in_features // groupsize, out_features, 2),
-                dtype=self.scales_precision,
+                precision=self.scales_precision,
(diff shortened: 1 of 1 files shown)


syedshazli:
Signed the form.

syedshazli:
Adding commits, I need to replace mentions of dtype with precision only in modules where dtype is to be deprecated

syedshazli:
Tests are failing as they still reference dtype. I'm wondering if it's okay for me to edit the tests, or if this deprecation should still expect all tests to pass as is

jerryzh168:
sorry just saw this, this is the older part of the code base that we'd like to just remove actually, so I think we'd accept a PR that removes these

syedshazli:
Oh okay I can remove it from the tests then.

I'd also be happy to deprecate or refactor other aspects of the "old" part of the codebase after I finish up some of these other PRs

PR #3515, opened 2025-12-18
Remove ::kernels for namespacing

syedshazli:
This PR addresses a TODO in torchao\csrc\cpu\torch_free_kernels\fallback\matmul\fp32_a_channelwise_8bit_b_fp32_c.h which says "TODO: Remove all ::kernels. No need for extra namespace."

I presume this meant remove all namespaces within the codebase that reference ::kernels, which is why this commit is so large. If this is not the case and was only meant for the .h file, I can change the commit to reflect this.

### DIFF
--- a/torchao/csrc/cpu/shared_kernels/benchmarks/benchmark_linear_8bit_act_xbit_weight.cpp
+++ b/torchao/csrc/cpu/shared_kernels/benchmarks/benchmark_linear_8bit_act_xbit_weight.cpp
@@ -18,7 +18,7 @@ template <int weight_nbit, bool has_weight_zeros, bool has_bias, bool has_clamp>
 UKernelConfig get_ukernel_config() {
   UKernelConfig config;
 
-  namespace ukernel = torchao::kernels::cpu::aarch64::linear::
+  namespace ukernel = torchao::cpu::aarch64::linear::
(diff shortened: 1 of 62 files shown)


syedshazli:
gentle ping @metascroy

metascroy:
In theory, I don't have an issue with the PR, but it does seem unnecessary.

Did you run all the tests and confirm everything still works and builds?

metascroy:
@jerryzh168 why are CI tests not being run on torchao APIs

jerryzh168:
there was some issue a while ago in the test infra I think, I'd recommend rebase on main and try again

metascroy:
@syedshazli can you rebase.

Did you try running the tests?

syedshazli:
@metascroy I remember running the tests a while ago and they seemed fine. Not sure why they failed here, I'll rebase and see what happened

syedshazli:
In terms of the necessity of the deprecation, it likely is quite unnecessary haha. I was just trying to get my feet wet with the torch/ao repo finding any # TODO in the codebase.

metascroy:
On second thought, I'm quite hesitant to make this change without more testing across torchao -> pytorch / torchao -> executorch boundaries.

Given the change seems unnecessary as well, I'm going to close the PR.

PR #3517, opened 2025-12-19
Add GPTQ to prototype

jcaip:
Summary:

This PR adds GPTQ support to `torchao.prototype.gptq`.

It is exposed via a new config, GPTQConfig, which can have two steps, "observe" and "convert". 

When `quantize_(model, GPTQConfig(step="observe"))` is run, observer tensors are attached to the weight tensors, which keep track of linear / torch.bmm ops and updates the hessian matrix based on the observed inputs. 

When `quantize_(model, GPTQConfig(step="convert"))` is run, we will find any observer tensors, take the Hessian and do int4 GPTQ quantization to find the weights. The core of this function is in `gptq_quantize`. 

Currently Int4WeightOnlyConfig and Int8WeightOnlyConfig are supported. 

Also included is an example script, `gptq_example.py` that does sequential / nonsequential quantization on helllaswag for a simple example. 

I am comparing against this survey paper: https://arxiv.org/pdf/2409.11055v1 We look to be on par accuraccy wise for int8 / int4 GPTQ.  (GPTQ** is llm-compressor, GPTQ* is autogptq)

To reproduces these results please run:
```
python torchao/prototype/gptq/gptq_example.py --quantization baseline
python torchao/prototype/gptq/gptq_example.py --quantization int4-rtn
python torchao/prototype/gptq/gptq_example.py --quantization int4-gptq-sequential
python torchao/prototype/gptq/gptq_example.py --quantization int4-gptq-nonsequential
python torchao/prototype/gptq/gptq_example.py --quantization int8-rtn
python torchao/prototype/gptq/gptq_example.py --quantization int8-gptq-sequential
python torchao/prototype/gptq/gptq_example.py --quantization int8-gptq-nonsequential

```
<html><head></head><body><p>Benchmark Results:</p>

Model | Method | W/A | Storage (GB) | BBH (3-shot) | TorchAO Results
-- | -- | -- | -- | -- | --
Llama-3.1-8B-it | FP16 | 16 / 16 | 16 | 51.35 | 51.08
  | FP8 | 8 / 8 | 8 | 50.78 (↓0.57) |  
  | INT8 | 8 / 16 | 8 | N/A | 41.82
  | INT4 | 4 / 16 | 4 | N/A | 50.91
  | GPTQ* | 4 / 16 | 4 | 48.27 (↓3.08) |  
  | GPTQ** | 4 / 16 | 4 | 48.48 (↓2.87) | 48.32
  | GPTQ** | 8 / 16 | 8 | 51.29 (↑0.06) | 51.10
  | SmoothQuant | 8 / 8 | 8 | 51.32 (↓0.03) |  
  | AWQ | 4 / 16 | 4 | 48.25 (↓3.10) |  

</body></html>

Timings:

running `python torchao/prototype/gptq/gptq_example.py --quantization int4-gptq-sequential` yields a quantized model in 6.6 minutes, while running the llmcompressor example [here](https://github.com/vllm-project/llm-compressor/blob/main/examples/quantization_w4a16/llama3_example.py) takes 8 minutes, so we're approximately ~13% faster. 

Need to profile the runs to get a sense of where the speedup is happening

Test Plan:

```
pytest test/prototype/gptq/test_gptq.py 
```

### DIFF
--- a/test/prototype/gptq/test_gptqv2.py
+++ b/test/prototype/gptq/test_gptqv2.py
@@ -17,7 +17,7 @@ from torchao.prototype.gptq import (
 from torchao.prototype.gptq.observer import GPTQObserverTensor
 from torchao.quantization import Int4WeightOnlyConfig, Int8WeightOnlyConfig, quantize_
 from torchao.quantization.granularity import PerRow
-from torchao.utils import _is_fbgemm_gpu_genai_available
+from torchao.utils import _is_mslk_available
 
 
 def _calculate_hessian(inputs, device=None):
@@ -216,7 +216,7 @@ class TestGPTQObserverTensor:
                 Int4WeightOnlyConfig(group_size=128),
                 id="int4",
                 marks=pytest.mark.skipif(
-                    not _is_fbgemm_gpu_genai_available(),
+                    not _is_mslk_available(),
                     reason="fbgemm_gpu not available",
                 ),
             ),
@@ -297,7 +297,7 @@ class TestGPTQFlow:
                 Int4WeightOnlyConfig(group_size=128),
                 id="int4",
                 marks=pytest.mark.skipif(
-                    not _is_fbgemm_gpu_genai_available(),
+                    not _is_mslk_available(),
                     reason="fbgemm_gpu not available",
                 ),
             ),
@@ -360,7 +360,7 @@ class TestGPTQFlow:
                 Int4WeightOnlyConfig(group_size=128),
                 id="int4",
                 marks=pytest.mark.skipif(
-                    not _is_fbgemm_gpu_genai_available(),
+                    not _is_mslk_available(),
                     reason="fbgemm_gpu not available",
                 ),
             ),
@@ -437,7 +437,7 @@ class TestGPTQFlow:
                 Int4WeightOnlyConfig(group_size=128),
                 id="int4",
                 marks=pytest.mark.skipif(
-                    not _is_fbgemm_gpu_genai_available(),
+                    not _is_mslk_available(),
                     reason="fbgemm_gpu not available",
                 ),
             ),
@@ -508,7 +508,7 @@ class TestGPTQFlow:
                 Int4WeightOnlyConfig(version=2),
                 id="int4",
                 marks=pytest.mark.skipif(
-                    not _is_fbgemm_gpu_genai_available(),
+                    not _is_mslk_available(),
                     reason="fbgemm_gpu not available",
                 ),
             ),
--- a/torchao/prototype/gptq/api.py
+++ b/torchao/prototype/gptq/api.py
@@ -13,7 +13,7 @@ import torch
 import torch.nn as nn
 
 try:
-    from fbgemm_gpu.experimental.gen_ai.quantize import int4_row_quantize_zp, pack_int4
+    from mslk.quantize import int4_row_quantize_zp, pack_int4
 except:
     int4_row_quantize_zp = None
     pack_int4 = None
--- a/.github/workflows/1xH100_tests.yml
+++ b/.github/workflows/1xH100_tests.yml
@@ -25,7 +25,7 @@ jobs:
         include:
           - name: H100
             runs-on: linux.aws.h100
-            torch-spec: '--pre torch torchvision torchaudio fbgemm-gpu-genai --index-url https://download.pytorch.org/whl/nightly/cu126'
+            torch-spec: '--pre torch torchvision torchaudio mslk --index-url https://download.pytorch.org/whl/nightly/cu126'
             gpu-arch-type: "cuda"
             gpu-arch-version: "12.4"
     permissions:
@@ -50,5 +50,7 @@ jobs:
         pytest test/integration --verbose -s
         pytest test/dtypes/test_affine_quantized_float.py --verbose -s
         python test/quantization/quantize_/workflows/float8/test_float8_tensor.py
+        python test/quantization/quantize_/workflows/int4/test_int4_tensor.py
+        python test/quantization/quantize_/workflows/int4/test_int4_preshuffled_tensor.py
         ./test/float8/test_everything_single_gpu.sh
         pytest test/prototype/mx_formats/ -s
--- a/.github/workflows/doc_build.yml
+++ b/.github/workflows/doc_build.yml
@@ -61,6 +61,8 @@ jobs:
     runs-on: [self-hosted, linux.2xlarge]
     needs: build_docs
     if: ${{ github.event_name == 'pull_request' || github.event_name == 'workflow_dispatch' }}
+    permissions:
+      pull-requests: write
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -77,7 +79,37 @@ jobs:
           s3-bucket: doc-previews
           if-no-files-found: error
           path: docs
-          s3-prefix: pytorch/ao/${{ github.event.pull_request.number }}
+          s3-prefix: pytorch/ao/${{ github.event_name == 'pull_request' && github.event.pull_request.number || github.run_id }}
+      - name: Get preview URL
+        if: ${{ github.event_name == 'workflow_dispatch' }}
+        id: preview-url
+        run: |
+          echo "url=https://docs-preview.pytorch.org/pytorch/ao/${{ github.run_id }}/index.html" >> $GITHUB_OUTPUT
+      - name: Find associated PR and post comment
+        if: ${{ github.event_name == 'workflow_dispatch' }}
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const prs = await github.rest.pulls.list({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              head: `${context.repo.owner}:${context.ref.replace('refs/heads/', '')}`,
+              state: 'open'
+            });
+            console.log(`Found ${prs.data.length} open PRs for branch ${context.ref}`);
+            if (prs.data.length > 0) {
+              const prNumber = prs.data[0].number;
+              console.log(`Posting comment to PR #${prNumber}`);
+              const previewUrl = '${{ steps.preview-url.outputs.url }}';
+              await github.rest.issues.createComment({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                issue_number: prNumber,
+                body: `## 📄 Doc Preview (workflow_dispatch)\n\nYou triggered a doc build manually via workflow_dispatch.\nThe pytorchbot doc preview link will not work for this build.\n\n**Use this link instead:** ${previewUrl}`
+              });
+            } else {
+              console.log('No open PR found for this branch');
+            }
 
   upload:
     runs-on: ubuntu-latest
--- a/.github/workflows/release_model.yml
+++ b/.github/workflows/release_model.yml
@@ -19,7 +19,7 @@ jobs:
         include:
           - name: H100
             runs-on: linux.aws.h100
-            torch-spec: '--pre torch torchvision torchaudio fbgemm-gpu-genai --index-url https://download.pytorch.org/whl/nightly/cu126'
+            torch-spec: '--pre torch torchvision torchaudio mslk --index-url https://download.pytorch.org/whl/nightly/cu126'
             gpu-arch-type: "cuda"
             gpu-arch-version: "12.4"
     permissions:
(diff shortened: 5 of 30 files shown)


test/prototype/gptq/test_gptqv2.py
```diff
@@ -0,0 +1,453 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
vkuzo, line 1: nit: put in `test/prototype/gptq` folder to match codebase convention

torchao/prototype/gptq/__init__.py
```diff
@@ -0,0 +1,269 @@
...
+
+from torchao.core.config import AOBaseConfig
+from torchao.quantization import Int4Tensor
+from torchao.quantization.quant_api import _module_extra_repr
+from torchao.quantization.transform_module import register_quantize_module_handler
+
+from .observer import ObserverTensor
+
+
+@dataclass
```
vkuzo: nit: put logic in `some_file.py` instead of `__init__.py`

torchao/prototype/gptq/__init__.py
```diff
@@ -0,0 +1,269 @@
...
+    H[dead, dead] = 1
+    W[:, dead] = 0
+
+    damp = percdamp * torch.mean(torch.diag(H))
+    diag = torch.arange(columns, device=device)
+    H[diag, diag] += damp
+    H = torch.linalg.cholesky(H)
+    H = torch.cholesky_inverse(H)
+    H = torch.linalg.cholesky(H, upper=True)
+    Hinv = H
```
vkuzo: nit: add some comments linking to the right page of the paper

torchao/prototype/gptq/__init__.py
```diff
@@ -0,0 +1,269 @@
...
+
+    all_qparams = []
+
+    for W_quantize_block, block_start in zip(
```
vkuzo: nit: add comments as needed to explain how this code is following the algorithm in the paper

torchao/prototype/gptq/__init__.py
```diff
@@ -0,0 +1,269 @@
...
+        shape=W.shape,
+        act_pre_scale=None,
+    )
+    return res
+
+
+def _calculate_hessian(inputs, device=None):
+    """Calculate Hessian matrix from input activations for GPTQ.
+
+    DEPRECATED: This function is kept for backward compatibility in tests only.
```
vkuzo: backward compatibility with what?

torchao/prototype/gptq/observer.py
```diff
@@ -0,0 +1,96 @@
...
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+
+import torch
+import torch.nn.functional as F
+
+from torchao.utils import TorchAOBaseTensor
+
+
+class ObserverTensor(TorchAOBaseTensor):
```
vkuzo: the logic here is specific to GPTQ, can we ensure the name specifies that

jerryzh168:
can you show e2e accuracy results as well

jcaip:
> I guess with this it's good that we don't need MultiTensor stuff anymore. I do remember there were some inplace op considerations before:
> 
> https://github.com/pytorch/ao/blob/f59a19bf0d97fd3672bf33497fd0b3508ec13254/torchao/quantization/GPTQ/GPTQ.py#L349
> 
> how is it addressed here?

The in-place op considerations are specific to MultiTensor, so if we avoid multi tensor we don't have to worry about them. 

We needed Multi-tensor before to support propagating quantized outputs to the subsequent observer modules. Right now we instead quantize one transformer layer at a time via some custom code in gptq_example.py. This mirrors the approach that https://github.com/IST-DASLab/MoE-Quant takes. 

llm comperssor does this by tracing the fx graph, it is possible to do something similar if we want.

torchao/prototype/gptq/api.py
```diff
@@ -0,0 +1,330 @@
...
+
+CONFIG_TO_TORCHAO_BASE_TENSOR = {
+    Int4WeightOnlyConfig: Int4Tensor,
+    Int8WeightOnlyConfig: Int8Tensor,
+}
+
+
+@dataclass
+class GPTQConfig(AOBaseConfig):
+    """Config for GPTQ quantization
```
jcaip, line 42: add explanation that this is with unquantized activations out of the box

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -308,6 +308,58 @@ def test_static_activation_per_row_int8_weight(self, granularity, dtype):
         # for compile, we can't compare dynamic vs static because we may get slightly different qparams when fused
         torch.testing.assert_close(dynamic_out_eager, static_out_eager)
 
+    @common_utils.parametrize("dtype", [torch.bfloat16])
+    def test_int8_weight_only_v1_vs_v2_comparison(self, dtype):
```
vkuzo: just test v2 and remove v1?

torchao/prototype/gptq/gptq_example.py
```diff
@@ -0,0 +1,368 @@
...
+from torchao.prototype.gptq import GPTQConfig
+from torchao.quantization import Int4WeightOnlyConfig, Int8WeightOnlyConfig, quantize_
+from torchao.quantization.granularity import PerRow
+
+
+def sequential_quantize(
+    model,
+    calibration_data: List[torch.Tensor],
+    config: Any,
+) -> None:
```
vkuzo, line 40: add docblock explaining that user is expected to implement this for their model

torchao/prototype/gptq/gptq_example.py
```diff
@@ -0,0 +1,368 @@
...
+    # run with no grad otherwise keeping all the tensors around for the backwards will cause oom
+    with torch.no_grad():
```
vkuzo: can be moved to a decorator on `sequential_quantize` to save a level of indent

torchao/prototype/gptq/api.py
```diff
@@ -0,0 +1,330 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
namgyu-youn, line 1: nit: rename to `api.py`, instead of `gptq.py` to following AWQ/SmoothQuant patterns?

torchao/prototype/gptq/gptq.py
```diff
@@ -0,0 +1,330 @@
...
+# LICENSE file in the root directory of this source tree.
+
+import types
+from dataclasses import dataclass
+from functools import partial
+from typing import Union
+
+import torch
+import torch.nn as nn
+from fbgemm_gpu.experimental.gen_ai.quantize import int4_row_quantize_zp, pack_int4
```
namgyu-youn: Can we focus on W8A8-INT only, instead of W4A16-INT for early land? fbgemm dependency issue has been reported many times I remember.

torchao/prototype/gptq/api.py
```diff
@@ -0,0 +1,330 @@
...
+    Int4WeightOnlyConfig,
+    Int8WeightOnlyConfig,
+    _module_extra_repr,
+)
+from torchao.quantization.transform_module import register_quantize_module_handler
+from torchao.quantization.utils import get_block_size
+
+from .observer import GPTQObserverTensor
+
+CONFIG_TO_TORCHAO_BASE_TENSOR = {
```
namgyu-youn, line 34: Is this needed? Can we drop this?

torchao/prototype/gptq/api.py
```diff
@@ -0,0 +1,330 @@
...
+    for chunk, scale_chunk, zero_chunk in zip(x_chunks, scales_list, zeros_list):
+        chunk_float = chunk.to(torch.float32) + 2 ** (n_bit - 1)
+        min_val = zero_chunk - scale_chunk * (2 ** (n_bit - 1))
+        dequant = chunk_float * scale_chunk + min_val
+        dequant_chunks.append(dequant)
+
+    return torch.cat(dequant_chunks, dim=-1)
+
+
+def gptq_quantize(H: torch.Tensor, W: torch.Tensor, config: GPTQConfig):
```
namgyu-youn, line 183: I think we should not implement quantization module inside GPTQ algorithm. To me, the main difference between AWQ/SmoothQuant/GPTQ vs. RTN is **preprocessing something** before applying RTN. 

It can be (1) activation outlier in SmoothQuant, (2) hessian inverse in GPTQ, and (3) salient weight in AWQ. To reproduce this, most of the research and we have been used calibration-flow (`lm-eval`).

Therefore, we should follow decomposed workflow:
1. Prepare: Initialization before running the algorithm; we might need linear layer for this
2. Observe: Compute "something" based on algorithm. Note that this step should based on calibration-flow
3. Store computed value inside chunk or something
4. Convert: Apply (3) to high-precision model, not quantized model
5. quantize: RTN using `quantize_(model, base_config)`

test/prototype/gptq/test_gptq.py
```diff
@@ -0,0 +1,534 @@
...
+)
+from torchao.prototype.gptq.observer import GPTQObserverTensor
+from torchao.quantization import Int4WeightOnlyConfig, Int8WeightOnlyConfig, quantize_
+from torchao.quantization.granularity import PerRow
+
+
+def _calculate_hessian(inputs, device=None):
+    """Calculate Hessian matrix from input activations for GPTQ.
+
+    DEPRECATED: This function is kept for backward compatibility in tests only.
```
vkuzo: nit: update docblock

torchao/prototype/gptq/gptq.py
```diff
@@ -0,0 +1,330 @@
...
+    """
+    This function implements the GPTQ algorithm described in this paper: https://arxiv.org/abs/2210.17323 (Algorithm 1)
+
+    GPTQ quantizes weights column-by-column while propagating quantization errors to subsequent columns.
+    This minimizes the overall error: argmin W' ||WX - W'X||_2^2
+
+    For example we can seee how GPTQ improves accuracy on a toy example:(2x2 matrix, 2-bit symmetric quantization):
+
+    Example:
+        Given: W = [[1.2, 0.8],    X = [[1.0],      Original output: W @ X = [[2.0],
```
vkuzo: nit: can we switch this to do `X @ W.T` to match the common case? `X @ W` is also fine, it's confusing to have `W` first here.

torchao/prototype/gptq/api.py
```diff
@@ -0,0 +1,330 @@
...
+                            :, group_start - block_start : group_end - block_start
+                        ],
+                        base_config.granularity,
+                    )
+
+            # Quantize each column and propagate errors to subsequent columns
+            for i in range(group_start - block_start, group_end - block_start):
+                w = W_quantize_block[:, i].unsqueeze(1)
+                if isinstance(base_config, Int4WeightOnlyConfig):
+                    q = _int4_row_quantize_zp_precomputed_qparams(
```
vkuzo, line 303: can probably improve how int4/int8/etc logic works in these two for loops, but better in a future PR

torchao/prototype/gptq/gptq.py
```diff
@@ -0,0 +1,330 @@
...
+                W_quantize_block[:, i:] -= err1.matmul(
+                    Hinv_quantize_block[i, i:].unsqueeze(0)
+                )
+                Err1[:, i] = err1.flatten()
+
+        # Lazy Batch-Updates: We process B columns at a time with local updates above.
+        # Once a block is fully processed, perform global updates to H^-1 and W using batched versions of the error propagation equations.
+        W[:, block_end:] -= Err1.matmul(Hinv[block_start:block_end, block_end:])
+
+    if "cuda" in device.type:
```
vkuzo: does this support cpu? maybe just drop cpu for now? my guess is things will be too slow to be useful on cpu

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -137,6 +137,8 @@ def from_hp(
                 scale_dtype=hp_tensor.dtype,
                 zero_point_dtype=torch.int8,
                 keepdim=True,
+                # Use float32 eps for all dtypes to avoid accuracy loss with bfloat16 (bfloat16 eps is ~65x larger).
```
vkuzo, line 144: separate PR

torchao/prototype/gptq/api.py
vkuzo: @namgyu-youn , the GPTQ algorithm doesn't cleanly translate to your steps 1-6 above without either program capture, custom model forward or other tricks.  The current PR takes a "custom model forward" approach.  I would recommend improving the usability in future PRs.

namgyu-youn: > @namgyu-youn , the GPTQ algorithm doesn't cleanly translate to your steps 1-6 above without either program capture, custom model forward or other tricks. The current PR takes a "custom model forward" approach. I would recommend improving the usability in future PRs.

Oh sorry I had to review OBQ again. GPTQ can't be done with 1-6 steps like AWQ/SmoothQuant. Mark as out of topic

torchao/prototype/gptq/gptq.py
jcaip: I think this just needs to be gated properly like the other fbgemm imports

torchao/prototype/gptq/api.py
jcaip: We need something to map from the configs to the tensor, which we don't have currently afaik.

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
jcaip: landed in #3664

torchao/prototype/gptq/gptq.py
jcaip: It should work but i agree this will be quite slow on CPU.

PR #3522, opened 2025-12-21
[mxfp8] bug fix and test updates related to new triton_calculate_scale param

danielvegamyhre:
- Second call site with new `triton_calculate_scale` param needs to be updated, this slipped through in #3498, motivation for getting blackwells in CI soon!
- Update triton mx dim1 tests to test both floor and rceil. Even though we aren't using it right now, just for completeness

## Tests
- `pytest test/prototype/mx_formats/test_kernels.py ` verified whole test suite is passing locally this time

### DIFF
--- a/torchao/prototype/mx_formats/kernels.py
+++ b/torchao/prototype/mx_formats/kernels.py
@@ -17,6 +17,7 @@ from torchao.prototype.custom_fp_utils import (
     _f32_to_floatx_unpacked,
     _floatx_unpacked_to_f32,
 )
+from torchao.prototype.mx_formats.config import ScaleCalculationMode
 from torchao.utils import (
     is_cuda_version_at_least,
     is_sm_at_least_100,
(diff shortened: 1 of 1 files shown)


danielvegamyhre:
cc @drisspg for review

torchao/prototype/mx_formats/kernels.py
```diff
@@ -639,7 +639,9 @@ def custom_triton_to_mxfp8_dim1_sharding(x, inner_block_size=32):
         return acceptable_shardings
 
     def triton_to_mxfp8_dim1_reference(
-        x_hp: torch.Tensor, block_size
+        x_hp: torch.Tensor,
```
drisspg, line 643: if we can we should try to have 1 default arg at the top user facing and then not populate for inner funcs should help with these types of bug

danielvegamyhre: the launcher functions do have defaults for scaling_mode (`rceil`) that are passed in to the actual kernels. (let me know if i misunderstood your comment)

danielvegamyhre: oh you mean in this torch reference impl? i will add a default.

PR #3523, opened 2025-12-22
[xpu][feature] Add xpu to fp8 quant API

Stonepia:
We have added the XPU support for `torch.scaled_mm`, so the fp8 related op in torchao could also be enabled. This pr adds the xpu to the torchao'a quant API.
To reduce review overhead, this PR does not have the tests included. We will enable tests in separate PR. Thanks!


liangan1 (approved):

Stonepia:
We decided to close this pr and put everything in https://github.com/pytorch/ao/pull/3332. Thus close it. Thanks for the help!

PR #3524, opened 2025-12-22
Separately control the activation quantization granularity

cyxlily:
Based on #3468
Separately control the activation quantization granularity and weight quantization granularity


torchao/prototype/smoothquant/api.py
```diff
@@ -95,6 +95,14 @@ def _smooth_quant_transform(
     else:
         raise ValueError(f"Unexpected step: {step}")
 
+    if isinstance(base_config, Int8StaticActivationInt8WeightConfig):
```
jerryzh168: I think smoothquant should not mention specific configs

torchao/quantization/quant_api.py
```diff
@@ -1381,6 +1381,7 @@ class Int8StaticActivationInt8WeightConfig(AOBaseConfig):
 
     scale: torch.Tensor
     granularity: Granularity = PerRow()
+    act_granularity: Granularity = PerTensor()
```
jerryzh168: also we just use tuple, like this: https://github.com/pytorch/ao/blob/428bbcf8e8ec3bdfb0c3ae38e787827d1b540e29/torchao/quantization/quant_api.py#L1592

torchao/quantization/quant_api.py
```diff
@@ -1398,8 +1423,18 @@ def _int8_static_activation_int8_weight_transform(
     *,
     parameter_name="weight",
 ):
-    assert config.granularity in {PerRow(), PerTensor()}, (
-        "Only PerRow and PerTensor is supported currently"
+    assert isinstance(config.granularity, (tuple, list)) and len(config.granularity) == 2, (
```
Xia-Weiwen: According to the code above, it can only be a list, right?

test/prototype/test_smoothquant.py
```diff
@@ -101,7 +105,15 @@ def test_smoothquant_accuracy(self, alpha, base_config, device, input_dtype):
 
         # Step 1. Basic quantization
         basic_model = deepcopy(m)
-        quantize_(basic_model, base_config)
+        if isinstance(base_config, Int8StaticActivationInt8WeightConfig):
+            quantize_(
+                basic_model,
+                Int8DynamicActivationInt8WeightConfig(
```
Xia-Weiwen, line 109: Dynamic or static?

cyxlily: When testing static, It uses dynamic as the basic model. Otherwise, directly quantizing the static config as the basic config will result in errors due to missing act_scale computed by static_scale.

torchao/prototype/smoothquant/api.py
cyxlily: Thanks for the comment, updated.

torchao/quantization/quant_api.py
cyxlily: Thanks for the comment, updated.

torchao/quantization/quant_api.py
cyxlily: Thanks for the comment, updated.

test/prototype/test_smoothquant.py
```diff
@@ -83,9 +83,14 @@ def setUpClass(cls):
     @common_utils.parametrize(
         "base_config",
         [
-            Int8DynamicActivationInt8WeightConfig(version=2),
```
Xia-Weiwen, line 86: Granularity is an optional argument. We should also test the default value. The default value should not be None.

torchao/prototype/smoothquant/api.py
```diff
@@ -96,11 +96,14 @@ def _smooth_quant_transform(
...
-        base_config.get_act_quant_kwargs()
-        if isinstance(base_config, IsStaticQuantizationConfig)
-        else None
-    )
+    if isinstance(base_config, IsStaticQuantizationConfig):
+        # Static quantization
+         quant_kwargs = QuantizeTensorToInt8Kwargs(
+             granularity=base_config.granularity[0],
+             mapping_type=base_config.act_mapping_type,
+         )
```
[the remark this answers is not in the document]
Xia-Weiwen: Try pre-commit to fix format issues locally

torchao/quantization/quant_api.py
```diff
@@ -1064,7 +1089,7 @@ class Int8DynamicActivationInt8WeightConfig(AOBaseConfig):
     layout: Optional[Layout] = PlainLayout()
     act_mapping_type: Optional[MappingType] = MappingType.SYMMETRIC
     weight_only_decode: bool = False
-    granularity: Granularity = PerRow()
+    granularity: Optional[Union[Granularity, List[Granularity]]] = None
```
Xia-Weiwen: Better not to change the default value. It should work without setting granularity explicitly.

torchao/quantization/quant_api.py
```diff
@@ -1183,14 +1219,17 @@ class Int8StaticActivationInt8WeightConfig(AOBaseConfig):
...
+            activations and weights) or a tuple of two granularities (first for activations, second for weights).
+            If None, defaults to PerRow for both. Only PerTensor and PerRow are supported.
         act_mapping_type (MappingType): The mapping type for activation quantization. only SYMMETRIC is supported currently
         set_inductor_config (bool): if True, adjusts `torchinductor` settings to recommended values.
         version (int): the version of the config
     """
 
     act_quant_scale: Optional[torch.Tensor] = None
-    granularity: Granularity = PerRow()
+    granularity: Optional[Union[Granularity, List[Granularity]]] = None
```
Xia-Weiwen: Same here. Should keep the default value

torchao/quantization/quant_api.py
```diff
@@ -1227,8 +1270,16 @@ def _int8_static_activation_int8_weight_transform(
     *,
     parameter_name="weight",
 ):
-    assert config.granularity in {PerRow(), PerTensor()}, (
-        "Only PerRow and PerTensor is supported currently"
+    assert isinstance(config.granularity, list) and len(config.granularity) == 2, (
+        "granularity should be a tuple or list of length 2"
+    )
+    granularity = config.granularity
+    act_granularity, weight_granularity = granularity
```
Xia-Weiwen:
```suggestion
    activation_granularity, weight_granularity = config.granularity
```

torchao/quantization/quant_api.py
```diff
@@ -1240,14 +1291,11 @@ def _int8_static_activation_int8_weight_transform(
...
 
-    activation_granularity = config.granularity
-    weight_granularity = config.granularity
-
     quantized_tensor = Int8Tensor.from_hp(
         getattr(module, parameter_name),
         granularity=weight_granularity,
         act_quant_kwargs=QuantizeTensorToInt8Kwargs(
-            granularity=activation_granularity,
+            granularity=act_granularity,
```
Xia-Weiwen: We can avoid this change by using the same name

Xia-Weiwen (changes requested):
The comments by copilot make sense to me. Please also take a look.

Xia-Weiwen (approved):

namgyu-youn:
@cyxlily hi, may I ask if you are planning mixed-granularity techniques like DeepGEMM?

Xia-Weiwen:
> @cyxlily hi, may I ask if you are planning mixed-granularity techniques like DeepGEMM?

Per-tensor quant for activation and per-channel (per-row) quant for weight is a common configuration for int8. Unrelated to DeepGEMM I think.

namgyu-youn:
> Per-tensor quant for activation and per-channel (per-row) quant for weight is a common configuration for int8. Unrelated to DeepGEMM I think.

Make sense, thanks.

Xia-Weiwen:
Hi @jerryzh168 Can you please review this PR again? Thanks.

torchao/prototype/smoothquant/api.py
```diff
@@ -96,11 +96,12 @@ def _smooth_quant_transform(
...
 
-    quant_kwargs = (
-        base_config.get_act_quant_kwargs()
-        if isinstance(base_config, IsStaticQuantizationConfig)
-        else None
-    )
+    if isinstance(base_config, IsStaticQuantizationConfig):
+        # Static quantization
+        quant_kwargs = base_config.get_act_quant_kwargs()
+        quant_kwargs.granularity = base_config.granularity[0]
```
jerryzh168: why do we need this? why doesn't the `base_config.get_act_quant_kwargs()` set the granularity correctly?

torchao/prototype/smoothquant/api.py
```diff
@@ -96,11 +96,11 @@ def _smooth_quant_transform(
     else:
         raise ValueError(f"Unexpected step: {step}")
 
-    quant_kwargs = (
-        base_config.get_act_quant_kwargs()
-        if isinstance(base_config, IsStaticQuantizationConfig)
-        else None
-    )
+    if isinstance(base_config, IsStaticQuantizationConfig):
```
jerryzh168: maybe just revert this change?

torchao/quantization/quant_api.py
```diff
@@ -1047,6 +1047,27 @@ def _float8_cutlass_quant_sparse(
     )
 
 
+def _process_granularity(
```
jerryzh168: if this Int8 only, can you move this to Int8Tensor?

torchao/quantization/quant_api.py
```diff
@@ -1082,6 +1107,9 @@ def __post_init__(self):
             "torchao.quantization.Int8DynamicActivationInt8WeightConfig"
         )
 
+        act_granularity, weight_granularity = _process_granularity(self.granularity)
+        self.granularity = [act_granularity, weight_granularity]
```
jerryzh168: it's not a good idea to change the config I think, it's better to do the processing in `_int8_dynamic_activation_int8_weight_quantize_tensor` and have a local variable for this

torchao/quantization/quant_api.py
```diff
@@ -1047,6 +1047,27 @@ def _float8_cutlass_quant_sparse(
...
+    processed_granularity = None
+    if granularity is None:
+        processed_granularity = (PerRow(), PerRow())
+    elif isinstance(granularity, (PerTensor, PerRow)):
+        processed_granularity = (granularity, granularity)
+    elif isinstance(granularity, (tuple, list)) and len(granularity) == 2:
+        processed_granularity = tuple(granularity)
+    else:
+        raise ValueError(f"Invalid granularity specification: {granularity}.")
+    return processed_granularity
```
jerryzh168: this is a mix of normalization and validation, I feel it might be better to split these 2 steps clearly

normalize: turn None to default 2 item tuple, turn 1 item to a tuple,  return 2 items directly, reject 3 items and above
--> makes sure the output is a 2 item tuple


validation: make sure the returned tuple is supported

torchao/quantization/quant_api.py
```diff
@@ -1131,11 +1159,16 @@ def get_weight_block_size(x):
         )
         quantized_weight = to_linear_activation_quantized(new_weight, input_quant_func)
     else:
-        assert config.granularity in {PerRow(), PerTensor()}, (
-            "Only PerRow and PerTensor are supported"
+        assert isinstance(config.granularity, list) and len(config.granularity) == 2, (
+            "granularity should be a tuple or list of length 2"
+        )
+        act_granularity, weight_granularity = config.granularity
+        assert act_granularity in {PerRow(), PerTensor()}, (
```
jerryzh168: nit: it's better to check with `isinstance(act_granularity, (PerRow, PerTensor))` I feel, similar for `weight_granularity`

also this can be done in a separate validation function

torchao/quantization/quant_api.py
```diff
@@ -1208,12 +1244,14 @@ def __post_init__(self):
         torch._C._log_api_usage_once(
             "torchao.quantization.Int8StaticActivationInt8WeightConfig"
         )
+        act_granularity, weight_granularity = _process_granularity(self.granularity)
+        self.granularity = [act_granularity, weight_granularity]
 
         # Validate activation granularity for static quantization
-        if isinstance(self.granularity, PerRow) and self.granularity.dim != -1:
+        if isinstance(self.granularity[0], PerRow) and self.granularity[0].dim != -1:
```
jerryzh168: nit: probably better to name this, e.g. `act_granularity = granularity[0]`

torchao/quantization/quant_api.py
```diff
@@ -1223,8 +1261,15 @@ def get_act_quant_kwargs(self) -> QuantizeTensorToInt8Kwargs:
         Returns:
             QuantizeTensorToInt8Kwargs with the configured granularity and mapping type.
         """
+        # If granularity is a list/tuple with 2 elements, use the first one for activation quantization
+        # Otherwise, use the granularity as is
+        if isinstance(self.granularity, (list, tuple)) and len(self.granularity) == 2:
```
jerryzh168: we can call normalize helper function to normalize self.granularity first

torchao/quantization/quant_api.py
```diff
@@ -1236,8 +1281,15 @@ def _int8_static_activation_int8_weight_transform(
     *,
     parameter_name="weight",
 ):
-    assert config.granularity in {PerRow(), PerTensor()}, (
-        "Only PerRow and PerTensor is supported currently"
+    assert isinstance(config.granularity, list) and len(config.granularity) == 2, (
```
jerryzh168: calls to _normalize and _validate functions can be done here before use

jerryzh168:
also is this only supporting CPU or both CUDA and CPU? (e.g. `act_granularity, weight_granularity = (PerTensor(), PerRow())`

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -25,11 +29,65 @@
...
+    "QuantizeTensorToInt8Kwargs",
+    "_normalize_granularity_int8",
+    "_validate_granularity_int8",
+]
 
 aten = torch.ops.aten
 
 
+def _normalize_granularity_int8(
+    granularity: Optional[
```
Xia-Weiwen: Let's define a type for this so that it can be reused.
```python
Int8Granularity = Optional[
    Union[
        Granularity,
        Tuple[Granularity, Granularity],
        list[Granularity],
    ]
]

def _normalize_granularity_int8(
    granularity: Int8Granularity,
) -> Tuple[Granularity, Granularity]:

class Int8DynamicActivationInt8WeightConfig(AOBaseConfig):
    granularity: Int8Granularity = PerRow()
```

torchao/quantization/quant_api.py
```diff
@@ -1192,14 +1196,17 @@ class Int8StaticActivationInt8WeightConfig(AOBaseConfig):
 
     Args:
         act_quant_scale (torch.Tensor): The scale tensor for activation quantization.
-        granularity (Granularity): The granularity of quantization. PerRow() and PerTensor() are supported currently
+        granularity (Optional[Union[Granularity, List[Granularity]]]):
```
Xia-Weiwen: Please update this comment after refining the typing

torchao/quantization/quant_api.py
```diff
@@ -1192,14 +1196,17 @@ class Int8StaticActivationInt8WeightConfig(AOBaseConfig):
...
+            activations and weights) or a tuple of two granularities (first for activations, second for weights).
+            If None, defaults to PerRow for both. Only PerTensor and PerRow are supported.
         act_mapping_type (MappingType): The mapping type for activation quantization. only SYMMETRIC is supported currently
         set_inductor_config (bool): if True, adjusts `torchinductor` settings to recommended values.
         version (int): the version of the config
     """
 
     act_quant_scale: Optional[torch.Tensor] = None
-    granularity: Granularity = PerRow()
+    granularity: Optional[Union[Granularity, List[Granularity]]] = PerRow()
```
Xia-Weiwen: Please update here after refining the typing

torchao/quantization/quant_api.py
```diff
@@ -1236,9 +1239,17 @@ def _int8_static_activation_int8_weight_transform(
...
     parameter_name="weight",
 ):
-    assert config.granularity in {PerRow(), PerTensor()}, (
-        "Only PerRow and PerTensor is supported currently"
+    activation_granularity, weight_granularity = _normalize_granularity_int8(
+        config.granularity
     )
+    _validate_granularity_int8((activation_granularity, weight_granularity))
+    # Validate activation granularity for static quantization
+    if isinstance(activation_granularity, PerRow) and activation_granularity.dim != -1:
```
Xia-Weiwen: If we require activation tensor to be 2d, `dim=1` should be valid, too.

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -25,11 +29,65 @@
...
+        raise ValueError(
+            f"Unsupported activation granularity type: {type(act_granularity)}. "
+            f"Only PerTensor and PerRow are supported."
+        )
+
+    if not isinstance(weight_granularity, (PerTensor, PerRow)):
+        raise ValueError(
+            f"Unsupported weight granularity type: {type(weight_granularity)}. "
+            f"Only PerTensor and PerRow are supported."
+        )
```
Xia-Weiwen: Do we support per-row quantization of activation and per-tensor quantization of weight?
And CUDA does not support mixed granularities, we need to ensure it's guarded elsewhere.

torchao/quantization/quant_api.py
```diff
@@ -1056,6 +1058,10 @@ class Int8DynamicActivationInt8WeightConfig(AOBaseConfig):
     Args:
         layout: Optional[Layout] = PlainLayout() - Tensor layout for the quantized weights. Controls how the
             quantized data is stored and accessed.
+        granularity (Optional[Union[Granularity, List[Granularity]]]):
```
Xia-Weiwen: Please update this comment after refining the typing

Xia-Weiwen (changes requested):

torchao/quantization/quantize_/workflows/__init__.py
```diff
@@ -54,4 +57,7 @@
     "IntxPackingFormat",
     "IntxUnpackedToInt8Tensor",
     "IntxOpaqueTensor",
+    "Int8Granularity",
+    "_normalize_granularity_int8",
+    "_validate_granularity_int8",
```
jerryzh168: these should not be top level API I think, please follow @Xia-Weiwen's Float8OpaqueTensor code: https://github.com/pytorch/ao/blob/79c8d01f0adc539eb376ddcfe60215441ad9017e/torchao/prototype/float8_opaque_tensor/float8_opaque_tensor.py#L98 (defining as a class util method for Int8Tensor, but still kept as 2 methods, I think that's clearer)

torchao/quantization/quant_api.py
```diff
@@ -1192,14 +1197,17 @@ class Int8StaticActivationInt8WeightConfig(AOBaseConfig):
...
+            activations and weights) or a tuple of two granularities (first for activations, second for weights).
+            If None, defaults to PerRow for both. Only PerTensor and PerRow are supported.
         act_mapping_type (MappingType): The mapping type for activation quantization. only SYMMETRIC is supported currently
         set_inductor_config (bool): if True, adjusts `torchinductor` settings to recommended values.
         version (int): the version of the config
     """
 
     act_quant_scale: Optional[torch.Tensor] = None
-    granularity: Granularity = PerRow()
+    granularity: Int8Granularity = PerRow()
```
jerryzh168: nit: we don't need `Int8Granularity`, it's fine to just use `Granularity` and do validation later

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -115,6 +122,55 @@ def __repr__(self):
...
+        granularity: Optional[
+            Union[
+                Granularity,
+                Tuple[Granularity, Granularity],
+                list[Granularity],
+            ]
+        ],
+    ) -> Tuple[Granularity, Granularity]:
+        normalized_granularity = None
+        supported = (PerTensor, PerRow)
```
jerryzh168: this should not appear in normalize btw, normalize should be very simple, just making sure the output is (act_granularity, weight_granularity) tuple, nothing else

jerryzh168: and we should be able to move normalize to a common util in the future, should be nothing specific to int8

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -115,6 +122,55 @@ def __repr__(self):
...
+                )
+        else:
+            raise ValueError(
+                f"Invalid granularity type: {granularity}. "
+                f"Expected None, Granularity, or tuple/list of 2 Granularities."
+            )
+        return normalized_granularity
+
+    @classmethod
+    def _validate_granularity_int8(
```
jerryzh168: nit: _validate_granularity would be enough since this is defined in Int8Tensor, same for normalize

torchao/quantization/quant_api.py
```diff
@@ -1236,9 +1241,19 @@ def _int8_static_activation_int8_weight_transform(
     *,
     parameter_name="weight",
 ):
-    assert config.granularity in {PerRow(), PerTensor()}, (
-        "Only PerRow and PerTensor is supported currently"
+    activation_granularity, weight_granularity = Int8Tensor._normalize_granularity_int8(
+        config.granularity
     )
+    Int8Tensor._validate_granularity_int8((activation_granularity, weight_granularity))
```
jerryzh168: maybe validation can happen once in `__post_init__`?

torchao/quantization/quant_api.py
```diff
@@ -1236,9 +1241,19 @@ def _int8_static_activation_int8_weight_transform(
...
+    # Validate activation granularity for static quantization
+    if isinstance(
+        activation_granularity, PerRow
+    ) and activation_granularity.dim not in [-1, 1]:
```
jerryzh168: why does it become [-1, 1] now? previous is just -1

jerryzh168: if you want to add this, it should be a separate PR, and also need a test for the addition

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
jerryzh168: yeah if something is only supported in CPU, please make sure it's guarded, and also make sure it's tested

seems we are extending the support in the PR, and I didn't see a test for the main config with things like (PerTensor(), PerRow()), in https://github.com/pytorch/ao/blob/main/test/quantization/quantize_/workflows/int8/test_int8_tensor.py

test/prototype/test_smoothquant.py
```diff
@@ -84,8 +84,19 @@ def setUpClass(cls):
         "base_config",
         [
             Int8DynamicActivationInt8WeightConfig(version=2),
+            Int8DynamicActivationInt8WeightConfig(version=2, granularity=PerRow()),
```
jerryzh168: also I think we should first add test in https://github.com/pytorch/ao/blob/main/test/quantization/quantize_/workflows/int8/test_int8_tensor.py, smoothquant support can be a separate PR after the basic things are tested first

torchao/quantization/quant_api.py
Xia-Weiwen: Hi @jerryzh168  I think the case is that activation is a 2d tensor and the users write `PerRow(dim=1)`, which is also valid. What do you think? Thanks.

cyxlily:
> also is this only supporting CPU or both CUDA and CPU? (e.g. `act_granularity, weight_granularity = (PerTensor(), PerRow())`

Yes, this supports both CPU and CUDA.

torchao/quantization/quant_api.py
jerryzh168: I think it's better to put these into a separate PR, it's confusing to change this at the same time of doing refactor

torchao/quantization/quant_api.py
```diff
@@ -1208,12 +1221,16 @@ def __post_init__(self):
...
             "torchao.quantization.Int8StaticActivationInt8WeightConfig"
         )
+        act_granularity, weight_granularity = Int8Tensor._normalize_granularity(
+            self.granularity
+        )
+        Int8Tensor._validate_granularity((act_granularity, weight_granularity))
 
         # Validate activation granularity for static quantization
-        if isinstance(self.granularity, PerRow) and self.granularity.dim != -1:
+        if isinstance(act_granularity, PerRow) and act_granularity.dim != -1:
```
jerryzh168: this is also doing validating granularity? it should be merged `_validate_granularity` I think

test/prototype/test_smoothquant.py
jerryzh168: please split the changes in this file to a separate PR to keep the current PR clear

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
cyxlily: Yes, we support per-row quantization of activation and per-tensor quantization of weight, and support mixed granularities on CUDA too.

torchao/quantization/quant_api.py
cyxlily: It's only validating static granularity. _validate_granularity is for both dynamic and static.

test/prototype/test_smoothquant.py
cyxlily: Thanks for comment, create a separate PR: https://github.com/pytorch/ao/pull/3847

torchao/quantization/quant_api.py
jerryzh168: I see, OK. seems to me then it might be better to move _validate_granularity to this file and define 

`_validate_granularity_int8_dynamic` and `_validate_granularity_int8_static` util functions here to be more consistent

cyxlily: Thanks for comment, updated.

torchao/quantization/quant_api.py
```diff
@@ -1047,6 +1047,45 @@ def _float8_cutlass_quant_sparse(
     )
 
 
+def _validate_granularity(
```
jerryzh168: I feel it's fine to repeat this code for now

jerryzh168 (approved):
looks good, thanks! left one comment inline

torchao/quantization/quant_api.py
cyxlily: Thanks for the comment. I just found that both dynamic and static need to check PerPow(dim=-1), because they both don't support per-feature activation quantization due to slicing limitations. I unified them into _validate_granularity_int8.

jerryzh168:
@Xia-Weiwen please wait for us to merge for PRs that's not CPU only

Xia-Weiwen:
> @Xia-Weiwen please wait for us to merge for PRs that's not CPU only

Oh, sorry about that. I merged it because I saw you approved it.

PR #3525, opened 2025-12-22
Support NVFP4 fp32 bias

andrewor14:
**Summary:** Today we hit this error with fp32 inputs + bias:

```
RuntimeError: Bias is not supported when module weight is in fp32
(out_dtype=Float32). Please use bfloat16 or float16 weights,
or remove the bias from the linear layer.
```

This is thrown by `NVFP4DynamicActivationNVFP4WeightConfig` but it's trying to guard against this underlying `_scaled_mm` error:

```
RuntimeError: Bias is not supported when out_dtype is set to Float32
```

This commit works around these errors by adding the bias separately in this case, similar to what float8 does.

**Test Plan:**

```
```

### DIFF
--- a/torchao/prototype/mx_formats/nvfp4_tensor.py
+++ b/torchao/prototype/mx_formats/nvfp4_tensor.py
@@ -470,13 +470,36 @@ def _addmm_nvfp4_dispatch(
         assert b.per_tensor_scale is None and a.per_tensor_scale is None
         scale_result = None
 
-    # THIS IS A WORKAROUND:
-    # RuntimeError: CUDA error: CUBLAS_STATUS_INVALID_VALUE when calling
+    # THIS IS A WORKAROUND FOR TWO ERRORS:
+    #
+    # (1) RuntimeError: CUDA error: CUBLAS_STATUS_INVALID_VALUE when calling
     # When we have per-tensor scaling, we need to apply it before bias
     # since bias is not quantized
-    should_add_bias_separately = (scale_result is not None) and (bias is not None)
+    #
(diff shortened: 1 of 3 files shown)


torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -470,11 +470,17 @@ def _addmm_nvfp4_dispatch(
...
-    # THIS IS A WORKAROUND:
-    # RuntimeError: CUDA error: CUBLAS_STATUS_INVALID_VALUE when calling
+    # THIS IS A WORKAROUND FOR TWO ERRORS:
+    #
+    # (1) RuntimeError: CUDA error: CUBLAS_STATUS_INVALID_VALUE when calling
     # When we have per-tensor scaling, we need to apply it before bias
     # since bias is not quantized
-    should_add_bias_separately = (scale_result is not None) and (bias is not None)
+    #
+    # (2) RuntimeError: Bias is not supported when out_dtype is set to Float32
```
drisspg, line 479: Okay this is what I thought would happen

andrewor14: Yeah but this only happens if `per_tensor_scale=None` (by default it is not), so users generally won't run into the `_scaled_mm` error. Either way this PR fixes that case

drisspg: can you just create a note somewhere on the conversions / casting path for the paths?

andrewor14: added a comment

Issue #3373, opened 2025-11-22
Support offloading feature in diffuser


jerryzh168:
1. diffuser currently is using older torchao versions, we should update it to stable 0.14.1 (and use nightly for testing)

2. Repro error and add missing support torchao to unskip the following tests:

Note: right now they are using int8 weight only: https://github.com/huggingface/diffusers/blob/d5da453de56fe73e0cfd26204ccca441af568ca1/tests/quantization/torchao/test_torchao.py#L660, but we haven't fully [migrated](https://github.com/pytorch/ao/issues/2752#issue-3316116488) this config yet, maybe we can start with [float8 weight only config](https://github.com/pytorch/ao/blob/2ff1eb2e356275cfbe46832327387d382c72945d/torchao/quantization/quant_api.py#L1623) which is using the new Float8Tensor design and expand to other quant types.

https://github.com/huggingface/diffusers/blob/d5da453de56fe73e0cfd26204ccca441af568ca1/tests/quantization/torchao/test_torchao.py#L664

https://github.com/huggingface/diffusers/blob/d5da453de56fe73e0cfd26204ccca441af568ca1/tests/quantization/torchao/test_torchao.py#L673

int8 weight only is still more important, but migration is in progress: https://github.com/pytorch/ao/pull/3241/files, we can use float8 to figure out what fix is needed first and then do the same fix on the migrated int8 tensor.

cc @sayakpaul

jerryzh168:
repro 1 (pin memory):


```
import torch 
from torchao.quantization import quantize_, Int8WeightOnlyConfig 


module = torch.nn.Linear(3072, 3072)
quantize_(module, Int8WeightOnlyConfig())

for p in module.parameters():
    p.data.cpu().pin_memory()
```

```
Traceback (most recent call last):
  File "/fsx/sayak/diffusers/check_pin_torchao.py", line 9, in <module>
    p.data.cpu().pin_memory()
  File "/fsx/sayak/miniconda3/envs/diffusers/lib/python3.10/site-packages/torchao/utils.py", line 658, in _dispatch__torch_function__
    return func(*args, **kwargs)
  File "/fsx/sayak/miniconda3/envs/diffusers/lib/python3.10/site-packages/torchao/utils.py", line 678, in _dispatch__torch_dispatch__
    raise NotImplementedError(
[...]

PR #3526, opened 2025-12-22
fp8 pinned mem

liangel-02:
Enabling pinned memory for Float8Tensor per [#3373](https://github.com/pytorch/ao/issues/3373)

Test: `python test/quantization/quantize_/workflows/float8/test_float8_tensor.py -k test_pin_memory`


liangel-02:
cc @sayakpaul

sayakpaul (approved):
Thanks! This should be enough. But I have asked some questions about this in our internal channel @jerryzh168.

jerryzh168:
> Thanks! This should be enough. But I have asked some questions about this in our internal channel @jerryzh168.

I have replied, I think you just need to switch version to 2

PR #3533, opened 2025-12-23
Remove permute op for fp8 conv

jerryzh168:
Summary:
Previously we need to permute input, weight and output because of the shape expectations of `torch.ops.mslk.f8f8bf16_conv` (working with channels_last shape + contigous memory_format) now that mslk has supported `channels_first shape (original shape) + channels_last memory_format`, we don't need to do permute before and after the fp8 conv. This PR removes these permute ops.

We hope this can speedup the model using fp8 conv, but will need to test this in downstream workloads

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/quantization/quantize_/workflows/float8/float8_tensor.py
+++ b/torchao/quantization/quantize_/workflows/float8/float8_tensor.py
@@ -47,6 +47,9 @@ from torchao.utils import (
     is_sm_at_least_100,
 )
 
+if _is_mslk_available():
+    import mslk.conv  # noqa: F401
+
 __all__ = [
     "Float8Tensor",
     "QuantizeTensorToFloat8Kwargs",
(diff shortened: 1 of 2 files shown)


torchao/quantization/quantize_/workflows/float8/float8_tensor.py
```diff
@@ -549,19 +549,16 @@ def _quantize_and_scaled_conv3d(
...
-    input_qdata = input_qdata.permute([0, 2, 3, 4, 1])
-
-    # move C_in to last dim
-    # after permute: (C_out, K1, K2, K3, C_in)
-    weight_qdata = weight_qdata.permute([0, 2, 3, 4, 1])
-
     input_scale = input_tensor.scale
     weight_scale = weight_tensor.scale
 
     import mslk.conv  # noqa: F401
```
vkuzo: move to top of file?

vkuzo (approved):

torchao/quantization/quantize_/workflows/float8/float8_tensor.py
jerryzh168: OK, sg. this is not always available, so will need to guard it with `_is_mslk_available`

also currently fbgemm is imported in the is available function: https://github.com/pytorch/ao/blob/428bbcf8e8ec3bdfb0c3ae38e787827d1b540e29/torchao/utils.py#L1141 we'll have to refactor that as well I think

PR #3538, opened 2025-12-24
[CPU][PT2E] Add inductor lowering path for quantize_affine_float8_non_decomposed

Xia-Weiwen:
As the title.


Xia-Weiwen:
UT failures are unrelated and they will be fixed via https://github.com/pytorch/ao/pull/3542

PR #3544, opened 2025-12-25
[ROCM] Add MI350 support for MXFP8 colwise quantization.

xiaobochen-amd:
## Summary
Add ROCm MI350 (gfx950) support for MXFP8 quantization kernel.

## Changes
- Implement `mxfp8_quantize` for ROCm in `mxfp8_extension.cpp` and `mxfp8_rocm.hip`
- Support colwise quantization with column-major output layout (matching CUDA API)
- Support both FLOOR and RCEIL scaling modes
- Add MI350 to test conditions in `test_kernels.py`

## Testing
- Validated against CUDA reference implementation on MI350
- All `test_cuda_mx_dim1_numerics` tests pass for FLOOR and RCEIL modes

```
docker:  rocm/primus:v25.10

torch==2.11.0.dev20251221+rocm7.1
```

## Bench
```
python3 ./benchmarks/mx_formats/cast_bench.py --mode=dim1_mxfp8_cuda_rceil

M 16384 K 16384 BLOCK_SIZE 32
GPU: AMD Instinct MI355X
torch version: 2.10.0.dev20251112+rocm7.1
triton version: 3.4.0
mode: dim1_mxfp8_cuda_rceil
time_us 3423.0724573135376
mem_bw_gbps 237.70895479045635
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -581,6 +581,10 @@ def get_extensions():
     # Add CUDA source files if needed
     if use_cuda:
         sources += cuda_sources
+    if use_rocm:
+        extra_compile_args["nvcc"].append("--offload-arch=gfx942")
+        extra_compile_args["nvcc"].append("--offload-arch=gfx950")
+        sources += rocm_sources
 
     # Add MXFP8 cuda extension dir
     mxfp8_extension_dir = os.path.join(extensions_dir, "cuda", "mx_kernels")
@@ -589,12 +593,15 @@ def get_extensions():
     )
     sources = [s for s in sources if s not in mxfp8_sources_to_exclude]
 
+    # Add MXFP8 rocm extension dir
+    mxfp8_rocm_extension_dir = os.path.join(extensions_dir, "rocm", "mx_kernels")
+    mxfp8_rocm_sources_to_exclude = list(
+        glob.glob(os.path.join(mxfp8_rocm_extension_dir, "**/*"), recursive=True)
+    )
+    sources = [s for s in sources if s not in mxfp8_rocm_sources_to_exclude]
+
     # TOOD: Remove this and use what CUDA has once we fix all the builds.
-    # TODO: Add support for other ROCm GPUs
-    if use_rocm:
-        extra_compile_args["nvcc"].append("--offload-arch=gfx942")
-        sources += rocm_sources
-    else:
+    if not use_rocm:
         # Remove ROCm-based sources from the sources list.
         extensions_rocm_dir = os.path.join(extensions_dir, "rocm")
         rocm_sources = list(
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -736,6 +743,37 @@ def get_extensions():
...
+            os.path.join(mxfp8_rocm_extension_dir, "mxfp8_rocm.hip"),
+        ]
+
+        # Only add the extension if the source files exist
+        mxfp8_rocm_src_files_exist = all(os.path.exists(f) for f in mxfp8_rocm_sources)
+        if mxfp8_rocm_src_files_exist:
+            print("Building mxfp8_rocm extension")
+            ext_modules.append(
+                CUDAExtension(
+                    name="torchao._C_mxfp8",
```
danielvegamyhre, line 759: I think this can't have the same extension name as the CUDA C++ mxfp8 extension or it will be overwritten/conflict. 

We should have a separate extension for ROCM

torchao/csrc/rocm/mx_kernels/mxfp8_rocm.hip
```diff
@@ -0,0 +1,151 @@
...
+    if (scale == 0) return 1.0f;
+    return exp2f(FP32_EXPONENT_BIAS - static_cast<float>(scale));
+}
+
+// MXFP8 colwise quantization kernel
+// Input: row-major [rows, cols]
+// Output: column-major [rows, cols] with stride {1, rows}
+// Scales: column-major [cols, num_row_blocks] with stride {1, cols}
+template<typename IType, typename OType, int BLOCK_SIZE, ScalingMode MODE>
+__global__ void mxfp8_quantize_colwise_kernel(
```
danielvegamyhre, line 58: Can you please update [this benchmark script](https://github.com/pytorch/ao/blob/3955b6ccbd8c084c8c543a466d2b9a136a0c3531/benchmarks/mx_formats/cast_bench.py#L96) with this kernel, and include the results in the PR description?

torchao/prototype/mx_formats/kernels.py
```diff
@@ -1138,9 +1139,9 @@ def triton_mxfp8_dequant_dim0(
         raise AssertionError("needs torch version 2.8+ and triton")
 
 
-mxfp8_cuda_extension_available = is_sm_at_least_100() and is_cuda_version_at_least(
-    12, 8
-)
+mxfp8_cuda_extension_available = (
+    is_sm_at_least_100() and is_cuda_version_at_least(12, 8)
```
danielvegamyhre, line 1143: The mxfp8 c++ extension includes various kernels we have no ROCM equivalent for, I don't think we should be re-using the same extension for CUDA and AMD, would be better to keep them separate

setup.py
xiaobochen-amd: I was looking at torchao/csrc/rocm/swizzle. It seems that it can share torchao._C with CUDA. Could you help me understand why MX needs a separate one instead?

torchao/prototype/mx_formats/kernels.py
xiaobochen-amd: I’d like to confirm whether, for this specific case, there are any clearly unsupported functionalities on the ROCm side. While the underlying C++ implementations can be different, the python-side interface and call path are identical. In that situation, would maintaining two separate python code introduce unnecessary redundancy?

vkuzo:
> mem_bw_gbps 237.70895479045635

what % of peak memory bandwidth is this kernel achieving?

xiaobochen-amd:
I tested this on MI355; the theoretical peak bandwidth is 8 TB/s.

vkuzo:
> I tested this on MI355; the theoretical peak bandwidth is 8 TB/s.

makes sense.  What performance do you see with just using torch.compile + triton?  We only added the CUDA custom kernel because it considerably beat what we could get for free from the compiler.  I think it only makes sense to add a AMD custom kernel here if (a) performance is better than what we can get without a custom kernel with compile and (b) performance is "good" (maybe >40% peak mem bw).  Without those requirements, we are adding software complexity without performance benefits.

xiaobochen-amd:
> > I tested this on MI355; the theoretical peak bandwidth is 8 TB/s.
> 
> makes sense. What performance do you see with just using torch.compile + triton? We only added the CUDA custom kernel because it considerably beat what we could get for free from the compiler. I think it only makes sense to add a AMD custom kernel here if (a) performance is better than what we can get without a custom kernel with compile and (b) performance is "good" (maybe >40% peak mem bw). Without those requirements, we are adding software complexity without performance benefits.

Good suggestion. It might make sense to try a triton version and see how the performance looks.

vkuzo:
> Good suggestion. It might make sense to try a triton version and see how the performance looks.

I'd recommend:
1. just use torch.compile
2. if (1) is not good enough, try a handwritten triton kernel, only check in if better than 1
3. if (2) is not good enough, try a lower level language, only check in if better than 2

xiaobochen-amd:
https://github.com/pytorch/ao/pull/3620

PR #3545, opened 2025-12-25
update version compatibility to allow torch 2.11.0.dev

danielvegamyhre:
Stacked PRs:
 * #3546
 * __->__#3545

--- --- ---

### update version compatibility to allow torch 2.11.0.dev
- update version compatibility to allow torch 2.11.0.dev
- Remove torch 2.9.0.dev nightlies, which are no longer active/indexed
- Also add logic so that if one .so fails to load, we still try to load the remaining ones, rather than skipping them. I added this because I found this bug where `_C_cutlass_100a.abi3.so` was failing to load and preventing mxfp8 .so from being loaded:
- Link `cuda` lib in mxfp8_cuda c++ extension for mxfp8 blocked layout kernel in next PR in stack

`Loading /home/danvm/ao/torchao/_C_cutlass_100a.abi3.so
Failed to import cpp extensions: operator torchao::_linear_8bit_act_1bit_weight does not exist`

### DIFF
--- a/torchao/__init__.py
+++ b/torchao/__init__.py
@@ -70,9 +70,7 @@ elif not ("+git" in __version__) and not ("unknown" in __version__):
     torchao_pytorch_compatible_versions = [
         # Built against torch 2.8.0
         (_parse_version("0.13.0"), _parse_version("2.8.0")),
-        (_parse_version("0.13.0"), _parse_version("2.9.0.dev")),
         (_parse_version("0.14.0"), _parse_version("2.8.0")),
-        (_parse_version("0.14.0"), _parse_version("2.9.0.dev")),
         # Built against torch 2.9.0
         (_parse_version("0.14.1"), _parse_version("2.9.0")),
         (_parse_version("0.14.1"), _parse_version("2.10.0.dev")),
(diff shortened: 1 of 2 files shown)


torchao/__init__.py
```diff
@@ -108,10 +109,15 @@ def _parse_version(version_string):
     try:
         from pathlib import Path
 
+        # Load abi3 .so files and cpython .so files
         so_files = list(Path(__file__).parent.glob("_C*.so"))
         if len(so_files) > 0:
             for file in so_files:
-                torch.ops.load_library(str(file))
+                print(f"Loading {file}")
```
danielvegamyhre: note: will change these to logger.debug() before landing

danielvegamyhre:
cc @vkuzo @andrewor14 for review

torchao/__init__.py
```diff
@@ -82,6 +82,7 @@ def _parse_version(version_string):
         # Current torchao version
         (_parse_version("0.16.0.dev"), _parse_version("2.9.1")),
         (_parse_version("0.16.0.dev"), _parse_version("2.10.0.dev")),
```
liangel-02, line 82: should we remove this line if the torch version has been bumped?

liangel-02 (approved):

torchao/__init__.py
danielvegamyhre: hmm i'm not sure, there are still plenty of 2.10 nightlies active in the index people could be using: https://download.pytorch.org/whl/nightly/torch/

we also still have 2.9 nightlies listed in the compatibility table, which don't exist in the index anymore, so can probably be removed: https://github.com/pytorch/ao/blob/27c5eb969f323f08b2870f98045a326771c4b515/torchao/__init__.py#L73

IMO we should:
- keep 2.10 nightlies while they are still active/indexed
- remove 2.9 nightlies which are no longer active/indexed

going to discuss in the team chat

PR #3546, opened 2025-12-25
[mxfp8 moe training] add CUDA kernel for per group blocked layout with groups along M

danielvegamyhre:
Stacked PRs:
 * __->__#3546
 * #3545

--- --- ---

### [mxfp8 moe training] add CUDA kernel for per group blocked layout with groups along M

- ~3.5x to ~12x faster than Triton

### Summary of kernel design

Terminology:
- "sf_tile" = a 128x4 scale factor tile
- "chunk" = a 128xCHUNK_WIDTH group of scale factor tiles, where CHUNK_WIDTH can be either 64 or 128
- "super block" = a group of chunks processed by one thread block. Stacked vertically along the row dimension into shape (e.g., 4 chunks per threadblock / `CHUNKS_PER_TB` would have shape (512, CHUNK_WIDTH)

Design:
- Pipelined design with double-buffering where we overlap "async TMA load of chunk N+1" with the "process and async TMA store of chunk N"
  - Motivation: in the original non-pipelined kernel, I found this kernel to be extremely ALU / integer arithmetic heavy in (1) the "group lookup" logic (described below) and (2) "per-group, per scale factor tile transformation to blocked layout" logic. Each warp was occupying SM resources and blocking other warps from being scheduled to due useful work until the heavy pointer arithmetic, SMEM layout change, and the store to GMEM was complete. Ideally, I would like to prefetch and have that next set of data in flight while I do all this heavy pointer arithmetic. Thus, I arrived at this pipelining approach. 
- We divide the work up into super blocks, each handled by one thread block, which iterates through the chunks in the superblock, pipelining as described above.
- Grid calculation:
    - Input group end offsets are dynamic and may result in partial tiles, requiring padding in our output tensor to avoid writing partial/invalid scale factor tiles. They are also only available on the device, thus we cannot directly use them to calculate our grid on the host for the kernel launch without doing a device-to-host sync, which will kill performance. Therefore, we allocate an extra `num_groups` thread blocks along the row dimension, to handle the worst case where every group boundary splits a scale factor tile and requires padding.
 - Inside the kernel, we do a prefix sum over the superblocks required for each group, and then use that to determine (1) which group this thread block should operate and (2) which local superblock we are within that group. 
 - The threadblock then uses that information to iterate through the chunks in that superblock, overlapping "async TMA load of chunk N+1" with "process in SMEM and async TMA store of chunk N+1" 

### Note on smem bank conflicts

We have four-way bank conflicts on writes for the blocked layout transform in SMEM. However, I think addressing them might be a net performance loss because then each scale factor tile would no longer be in contiguous layout in SMEM, thus preventing us from directly using a 1D TMA async store to GMEM. We would either have to use (1) blocking global stores or (2) cp.async, which uses threads and registers unlike TMA and complicates our pipeline stage coordination logic using 2 different forms of async data movement. It is unclear / difficult to reason about if this would be a net benefit, and I am happy with the big perf gains in the current implementation, and need to prioritize other things now.
  
### Benchmarks 
Looking at the best performing CUDA kernel configs (chunk width, chunks per threadblock) for each tested shape: 

- ~7x to 165x faster than torch impl using `torch.compile` (still requires d2h sync)
- ~3.5x to ~12x faster than Triton

Note: i didn't measure memory bandwidth utilization since this is an unusual kernel that is not cleanly compute bound (gemm) or memory bandwidth bound (quantization). it looks to be "uniform datapath" (identical instructions in each warp?) bound from what I see in NCU. 

```
input_shape      chunk_width    chunks_per_tb    torch_time_us    triton_time_us    cuda_time_us  triton_vs_torch    cuda_vs_torch      cuda_vs_triton
-------------  -------------  ---------------  ---------------  ----------------  --------------  -----------------  ---------------  ----------------
(32768, 160)              64                1           122.91             60.45           17.41  2.03x              7.06x                        3.47x
(32768, 160)              64                4           560.88             52.19           15.39  10.75x             36.44x                       3.39x
(32768, 160)              64                8           557.07             37.89           17.44  14.70x             31.94x                       2.17x
(131072, 160)             64                1          1867.25            128              31.78  14.59x             58.76x                       4.03x
(131072, 160)             64                4          1856.22            136.22           25.63  13.63x             72.42x                       5.31x
(131072, 160)             64                8          1853.86            279.55           23.55  6.63x              78.71x                      11.87x
(131072, 64)              64                1          1595.49            148.48           15.49  10.75x             103.01x                      9.59x
(131072, 64)              64                4          1528.14            103.42           15.39  14.78x             99.28x                       6.72x
(131072, 64)              64                8          1542.43            109.57           17.44  14.08x             88.44x                       6.28x
(131072, 224)             64                1          4884.77            234.4            39.97  20.84x             122.22x                      5.86x
(131072, 224)             64                4          4943.9             189.44           31.78  26.10x             155.59x                      5.96x
(131072, 224)             64                8          4923.94            211.97           29.73  23.23x             165.63x                      7.13x```

### DIFF
--- a/torchao/csrc/cuda/mx_kernels/mx_block_rearrange_2d_M_groups.cu
+++ b/torchao/csrc/cuda/mx_kernels/mx_block_rearrange_2d_M_groups.cu
@@ -1,0 +1,553 @@
+/**
+ * MX Block Rearrange 2D Kernel for M-Groups (Groups along row dimension)
+ *
+ * TERMINOLOGY:
+ * - SF Tile: A 128x4 scale factor tile (512 bytes in blocked layout)
+ *   - SF_ROWS = 128 rows
+ *   - SF_COLS = 4 columns
+ *   - Each SF tile is stored contiguously in the output blocked layout
+ *
+ * - Chunk: A 128xCHUNK_WIDTH region (128 rows × 64 or 128 columns)
+ *   - One TMA load brings one chunk from GMEM to SMEM
+ *   - Contains multiple SF tiles horizontally (CHUNK_WIDTH/SF_COLS = 16 or 32)
+ *   - Each chunk is processed independently in the pipeline
+ *
+ * - Superblock: CHUNKS_PER_TB chunks stacked vertically
+ *   - Processed by one thread block
+ *   - Height = SF_ROWS * CHUNKS_PER_TB (512, 1024, or 2048 rows)
+ *   - Width = CHUNK_WIDTH (64 or 128 columns)
+ *   - Grid is organized as (num_col_chunks, num_row_superblocks)
+ */
+#include <cstdint>
+#include <cstdio>
+#include <cstdlib>
+
+#include <cuda.h>
+#include <cuda/barrier>
+#include <cuda/ptx>
+#include <cuda_fp8.h>
+#include <cuda_runtime.h>
+#include <cudaTypedefs.h>
+#include "ptx.cuh"
+
+// Overloaded error checking for both CUDA driver API (CUresult) and runtime API (cudaError_t)
+inline void cuda_check_impl(CUresult result, const char* file, int line) {
+    if (result != CUDA_SUCCESS) {
+        const char* error_str;
+        cuGetErrorString(result, &error_str);
+        fprintf(stderr, "CUDA Driver Error at %s:%d - %s\n", file, line, error_str);
+        exit(EXIT_FAILURE);
+    }
+}
+
+inline void cuda_check_impl(cudaError_t result, const char* file, int line) {
+    if (result != cudaSuccess) {
+        fprintf(stderr, "CUDA Runtime Error at %s:%d - %s\n", file, line,
+                cudaGetErrorString(result));
+        exit(EXIT_FAILURE);
+    }
+}
+
+#define CUDA_CHECK(result) cuda_check_impl((result), __FILE__, __LINE__)
+
+// Get the driver entry point for cuTensorMapEncodeTiled (used for TMA tensor maps)
+inline void* get_driver_ptr() {
+    static void *driver_ptr = nullptr;
+    if (!driver_ptr) {
+        cudaDriverEntryPointQueryResult result;
+        CUDA_CHECK(cudaGetDriverEntryPoint("cuTensorMapEncodeTiled", &driver_ptr,
+                                cudaEnableDefault, &result));
+    }
+    return driver_ptr;
+}
+
+// Device helper: ceiling division
+__device__ __forceinline__ int ceil_div(int a, int b) {
+    return (a + b - 1) / b;
+}
+
+#define SF_ROWS 128
+#define SF_COLS 4
+
+// Each thread reads 16 bytes (vectorized uint4 load)
+#define BYTES_PER_THREAD 16
+
+// Double buffering for pipelined TMA loads/stores
+#define NUM_BUFFERS 2
+
+
+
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by danielvegamyhre** [comment]
cc @drisspg for review. let me know if there's anything i can do to make this easier to review. i included a summary of the kernel design in the PR description, and tried to ensure sufficient but not excessive comments in the code

**Comment by danielvegamyhre** [comment]
also @slayton58 if you have time to review this CUDA kernel that would be much appreciated! i am pretty happy with the current results but am curious to get feedback on the overall design and/or any thoughts on how it can be improved.

File: torchao/csrc/cuda/mx_kernels/mx_block_rearrange_2d_M_groups.cu [hunk 1]
```diff
@@ -0,0 +1,552 @@
...
+        return min(CHUNK_WIDTH, remaining_cols);
+    };
+    const int cols_in_chunk = compute_cols_in_chunk();
+
+    // The tensor map box width is min(CHUNK_WIDTH, scale_cols)
+    // TMA always loads this full box size (with OOB elements zero-filled)
+    const int tma_box_width = min(CHUNK_WIDTH, scale_cols);
+
+    // Func for async TMA load of a chunk of data to the given SMEM buffer
+    auto load_chunk = [&](int chunk_idx, int buf_idx) {
```
slayton58: Ditto here

slayton58:
> also @slayton58 if you have time to review this CUDA kernel that would be much appreciated! i am pretty happy with the current results but am curious to get feedback on the overall design and/or any thoughts on how it can be improved.

I've taken a quick look through - if it'll wait to merge until I'm back (01/05) then I can do a proper one.

My main comment is around using lambdas in the the kernel instead of splitting out into "proper" `__device__` functions - I think this makes the code harder to grok by adding stuff inline to the main `__global__` method rather than keeping it separate. Aside from that quibble, I think the code is nicely laid out, the pipelined nature of it is actually, imo, a really nice simple example of that concept already, and if the lambdas were to be split out, it would be even better. 

On a perf note, did you look at using `STG` for smem->gmem stores instead of TMA?  `STG` is fire-and-forget, so you won't block threads, and can potentially have lower latencies.

danielvegamyhre:
> I've taken a quick look through - if it'll wait to merge until I'm back (01/05) then I can do a proper one.
sounds good, I need to land this asap for scale testing deepseekv3.

> My main comment is around using lambdas in the the kernel instead of splitting out into "proper" `__device__` functions - I think this makes the code harder to grok by adding stuff inline to the main `__global__` method rather than keeping it separate. Aside from that quibble, I think the code is nicely laid out, the pipelined nature of it is actually, imo, a really nice simple example of that concept already, and if the lambdas were to be split out, it would be even better.

good point! will refactor to proper __device__ functions for readability.

> On a perf note, did you look at using `STG` for smem->gmem stores instead of TMA? `STG` is fire-and-forget, so you won't block threads, and can potentially have lower latencies.

interesting, i had mistakenly thought `st.global`/STG would be blocking, and only `cp.async` or TMA/`cp.async.bulk` would be non-blocking. will try it out. i might need to land this first though to do some dsv3 experiments i need to do, and then try it out later to see if we can improve further.

danielvegamyhre:
>My main comment is around using lambdas in the the kernel instead of splitting out into "proper" __device__ functions - I think this makes the code harder to grok by adding stuff inline to the main __global__ method rather than keeping it separate.

I agree with this in principle, so I tried refactoring but quickly realized `load_chunk` will require 9 args and `process_chunk` will require 20 (!) params, which makes it more messy/confusing imo..  

anyway, need to land now in order to keep moving, but am curious to get more feedback when you're back if you have time

PR #3552, opened 2025-12-26
delete outdated llama benchmarks

vkuzo:
Summary:

Since https://github.com/pytorch/ao/pull/3473 landed which benchmarks
LLaMa models with HF model definition and vllm, we no longer need the
previous benchmarks we had which used a custom model definition.
Deleting all of them.

Note that `benchmarks/_models/` is using HF model definition, but
deleting them too as https://github.com/pytorch/ao/pull/3473 is a more
modern version of these which includes performance.

Test Plan: CI

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo:
* #3554
* __->__ #3552

PR #3554, opened 2025-12-26
remove `torchao/prototype/moe_quant`

vkuzo:
Summary:

This is not used, removing.

Test Plan: CI

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/dtypes/test_affine_quantized_float.py
+++ b/test/dtypes/test_affine_quantized_float.py
@@ -434,9 +434,9 @@ class TestAffineQuantizedFloat8Compile(InductorTestCase):
                 expected_scale,
                 float8_dtype=float8_dtype,
             )
-            torch.testing.FileCheck().check(f"{quantize_affine_float8}.default").run(
-                code_q
-            )
+            # After lowering the op is not in the output code but the base name is
+            quant_op_base_name = f"{quantize_affine_float8}".split(".")[-1]
+            torch.testing.FileCheck().check(quant_op_base_name).run(code_q)
             test_dq, (code_dq,) = torch._inductor.utils.run_and_get_code(
                 torch.compile(dequantize_affine_float8),
                 test_q,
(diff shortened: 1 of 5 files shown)


vkuzo:
* __->__ #3554

liangan1:
@vkuzo  is there any plan to add the new design for moe quant?

liangan1:
cc @xiaowangintel

vkuzo:
> @vkuzo is there any plan to add the new design for moe quant?

Yes
1. we don't currently plan to provide any "MoE modules" (hence this PR which deletes "MoE modules")
2. we do plan to improve PyTorch Core MoE support (such as `torch._grouped_mm`) and support quantization as needed.  So, if user writes their MoE with `torch._grouped_mm`, torchao plans to support quantizing that to `torch._scaled_grouped_mm`.
3. we have tentative plans to improve the integration of torchao checkpoints into vLLM for MoE models, using vLLM's MoE kernels

liangan1:
> > @vkuzo is there any plan to add the new design for moe quant?
> 
> Yes
> 
> 1. we don't currently plan to provide any "MoE modules" (hence this PR which deletes "MoE modules")
> 2. we do plan to improve PyTorch Core MoE support (such as `torch._grouped_mm`) and support quantization as needed.  So, if user writes their MoE with `torch._grouped_mm`, torchao plans to support quantizing that to `torch._scaled_grouped_mm`.
> 3. we have tentative plans to improve the integration of torchao checkpoints into vLLM for MoE models, using vLLM's MoE kernels

Thanks for your info. But I suppose the item 2 should only works for the fp8/fp4, how about int4 and int8? besides, most of the MoE model definition in the HF/transformers is not based on the grouped_mm, is there any plan to extend the torch._grouped_mm adoption scope in the  HF/transformers?

vkuzo:
> But I suppose the item 2 should only works for the fp8/fp4, how about int4 and int8?

int4 and int8 could be supported as well, whether the kernel lives in Core, torchao or somewhere else can be case by case in the short term

> most of the MoE model definition in the HF/transformers is not based on the grouped_mm

yes, the MoE authoring story is very fragmented.  Long term, we want PyTorch core to have the right primitives to make MoE authoring easy, and in torchao have a story to easily quantize them.  Short term, we may have to have case-by-case workarounds.  We do plan to work on adoption of grouped_mm.

PR #3558, opened 2025-12-31
Matmul kernel preference support for `Int8Tensor`

namgyu-youn:
**Summary**:
Add kernel routing support (`kernel_preference`) for `Int8Tensor — "auto", "pytorch", and "triton"

**Motivation**:
`torch._int_mm` (INT8 MatMul kernel in pytorch internal) requires M (batch size) > 16, which failed CUDA graph capture like vLLM — https://gist.github.com/vkuzo/5bf389079442bb9851ef315cdcb797b4.

For better vLLM integration and performance, I would like to support alternative INT8 MatMul kernel support like the Triton-based `scaled_int8_mm`. This kernel was implemented by @gau-nernst 

**Example**:
```python
# Auto routing (dedault)
config = Int8DynamicActivationInt8WeightConfig(
    kernel_preference="auto"  # default, select based on setup

# Use ao kernel (inside pytorch)
config = Int8DynamicActivationInt8WeightConfig(
    kernel_preference="pytorch"  # Use ao kernel (inside pytorch)
)

# Use custom triton kernel
config = Int8DynamicActivationInt8WeightConfig(
    granularity=PerRow(),  # Note: support PerRow only
    kernel_preference="triton"  # Use Triton kernel
)

# Common quantization flow
quantize_(model, config)
```

Test plan:
```
pytest -sv test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```


vkuzo:
check out https://github.com/pytorch/ao/blob/main/torchao/prototype/quantized_training/int8_mm.py, maybe we can improve the existing one instead?

namgyu-youn:
> check out https://github.com/pytorch/ao/blob/main/torchao/prototype/quantized_training/int8_mm.py, maybe we can improve the existing one instead?

Thanks I didn't know it before. Updated PR entirely ~because improvement after promotion looks better.~ to use that triton kernel, could you look into it?

namgyu-youn:
@vkuzo could you please look again this updated PR?

namgyu-youn:
@vkuzo could you please look again this updated PR? also cc @jerryzh168

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -57,13 +58,15 @@ class Int8Tensor(TorchAOBaseTensor):
     Non-Tensor Attributes:
         granularity: the granularity for quantization (e.g., PerRow(), PerTensor())
         act_quant_kwargs: flags for dynamic activation quantization
+        mm_config: Matmul kernel to use - "pytorch" (default) or "triton"
```
vkuzo: it would be better to follow this design: https://github.com/pytorch/ao/blob/a5f2693089b4c6528f019a0fb17235c9f22180a9/torchao/quantization/quantize_/workflows/float8/float8_tensor.py#L98

vkuzo (changes requested):
let's be consistent with `Float8Tensor` for this logic

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
namgyu-youn: Actually, I didn't follow the `Float8Tensor` pattern. The differences can be summarized as:
- `Float8Tensor`: (1) default is None, (2) uses `mm_config` and `kernel_preference` for kernel selection
- `Int8Tensor`: (1) default is 'pytorch' (preserves existing behavior), (2) uses `mm_config` only

In my opinion, (1) having a default helps with simpler logic, and (2) only one config is needed for kernel selection. Could you please check again?

namgyu-youn: Personally, I don't understand why `Float8Tensor` needs 3 configs to route the kernel. 

It seems to work like: 1) `kernel_preference` is defined by the user, 2) `kernel_choice` is defined at runtime, 3) `mm_config` is defined after passing runtime? Is it possible to make it simpler?

vkuzo: for what you are adding here, `kernel_preference` is the existing abstraction we have, so I'm recommending to use that to stay consistent across the codebase.  `mm_config` shouldn't be related to this, and `kernel_choice` should not matter because it's not in the BC surface.

namgyu-youn: Understood, I will update to use https://github.com/pytorch/ao/blob/main/torchao/quantization/quantize_/common/kernel_preference.py then. Thanks for pointing it out.

namgyu-youn:
@vkuzo updated based on your suggestion and rebase, could you look into it?

torchao/quantization/quantize_/common/kernel_preference.py
```diff
@@ -30,9 +30,13 @@ class KernelPreference(str, Enum):
     """
     FBGEMM = "fbgemm"
 
+    """Use Triton kernels for quantization and matrix multiplication
+    """
+    TRITON = "triton"
```
vkuzo: looks good

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -233,16 +243,30 @@ def _(func, types, args, kwargs):
...
+        elif weight_tensor.kernel_preference == KernelPreference.TORCH:
+            # PyTorch kernel only applies row scaling, apply col scaling separately
+            intermediate_dtype = (
+                torch.float if x_scales.dtype == torch.half else x_scales.dtype
+            )
+            y_dot_scaled = int_scaled_matmul(
+                tmp, w_vals_int8_t, x_scales.reshape(-1, 1).to(intermediate_dtype)
+            ).to(output_dtype)
+            y = y_dot_scaled * w_scales.flatten()
+        else:
```
vkuzo: `AUTO` should always be supported

torchao/quantization/quant_api.py
```diff
@@ -1059,6 +1059,8 @@ class Int8DynamicActivationInt8WeightConfig(AOBaseConfig):
         set_inductor_config: bool = True - If True, adjusts `torchinductor` settings to recommended values
             for better performance with this quantization scheme.
         version (int): the version of the config, version 1 is using AffineQuantizedTensor that we plan to deprecate/split, version 2 is using Int8Tensor
+        kernel_preference (KernelPreference): Kernel preference for matmul operations. TORCH uses int_scaled_matmul,
```
vkuzo: put AUTO first

torchao/quantization/quant_api.py
```diff
@@ -1067,6 +1069,7 @@ class Int8DynamicActivationInt8WeightConfig(AOBaseConfig):
     granularity: Granularity = PerRow()
     set_inductor_config: bool = True
     version: int = 1
+    kernel_preference: KernelPreference = KernelPreference.TORCH
```
vkuzo: default to AUTO

namgyu-youn:
It seems that CI failure was caused by uninstalled Triton. Should I (1) add Triton to a dependency or (2) make unit test to skip if Triton is uninstalled?

namgyu-youn:
Just updated unit test to skip if Triton is not installed. This would be safer for CI.

namgyu-youn:
@vkuzo could you please look again this?

namgyu-youn:
Updated to conditional import Triton `torchao/kernel/__init__.py`, CI should be green now I believe.

namgyu-youn:
@vkuzo could you please look again this?

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -289,31 +291,34 @@ def test_pin_memory(self, config):
...
+        # Verify output shape and dtype
+        self.assertEqual(output.shape, (M, N))
+        self.assertEqual(output.dtype, torch.bfloat16)
 
-        sqnr = compute_error(output_v1, output_v2)
-        self.assertGreater(sqnr, 30, f"v1 vs v2 SQNR too low: {sqnr}")
+        # Verify correctness by comparing with reference
+        output_ref = torch.nn.functional.linear(input_tensor, weight_ref)
+        sqnr = compute_error(output_ref, output)
+        self.assertGreater(sqnr, 20, f"SQNR is too low: {sqnr} dB (expected > 20 dB)")
```
jerryzh168, line 337: we should also check numerical consistency between kernel preferences I think, can you add a new test similar to https://github.com/pytorch/ao/blob/436b2e9aeb01843e0241c38c4f9a69182b7693d9/test/quantization/quantize_/workflows/float8/test_float8_tensor.py#L763

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -290,27 +300,58 @@ def _(func, types, args, kwargs):
...
+        kernel_choice = None
+        if weight_tensor.kernel_preference == KernelPreference.AUTO:
+            # AUTO: prefer Triton on CUDA for rowwise scaled tensors
+            is_rowwise = tuple(weight_tensor.block_size) == (1,) * (
+                weight_tensor.dim() - 1
+            ) + (weight_tensor.shape[-1],)
+            kernel_choice = (
+                "triton" if tmp.device.type == "cuda" and is_rowwise else "torch"
+            )
+        elif weight_tensor.kernel_preference == KernelPreference.TRITON:
```
jerryzh168: if triton only supports rowwise we should do a check here and error out when user is not using rowwise here I think

torchao/kernel/int8_mm.py
jerryzh168, line 1: I'm not how this op has the same numerics as the other int8 mm ops actually, we should check

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
namgyu-youn: sure, added `test_kernel_preference_numerical_equivalence` to check numerical consistency.

torchao/kernel/int8_mm.py
namgyu-youn: Do you mean "numerics", not hardware behavior? Can we check under `test_kernel_preference_numerical_equivalence`?

jerryzh168: yes, we can check numerics with `test_kernel_preference_numerical_equivalence`

namgyu-youn:
@jerryzh168 could you please review this PR again?

namgyu-youn:
@jerryzh168 Could you please look into this PR again? I would like to improve `Int8Tensor` pipeline (benchmark: #4244 ), and this kernel is a core feature in that plan.

namgyu-youn:
After careful consideration, I am wondering if kernel preference is truly needed. For easy maintenance, it's not worth adding this router. Maybe we can just change kernels after gpu perf tests like #4244?

PR #3580, opened 2026-01-06
[mxfp8 moe training] mxfp8 token permute autograd func + triton kernels

danielvegamyhre:
Stacked PRs:
 * #3606
 * #3585
 * #3584
 * #3583
 * #3582
 * #3581
 * __->__#3580
 * #3579

--- --- ---

### Tests 
- `PYTHONPATH=/home/$USER/ao:$PYTHONPATH pytest test/prototype/moe_training/ep/test_permute.py`
- Tests for backward require integration with other EP pipeline components, so this is tested in the integration test added in #3584

PR #3579, opened 2026-01-06
[mxfp8 moe training] mxfp8 a2a_dispatch autograd function

danielvegamyhre:
Stacked PRs:
 * #3606
 * #3585
 * #3584
 * #3583
 * #3582
 * #3581
 * #3580
 * __->__#3579

--- --- ---

## Tests
- `PYTHONPATH=/home/$USER/ao:$PYTHONPATH pytest test/prototype/moe_training/ep/test_a2a_dispatch.py`
- Tests for backward require integration with other EP pipeline components, so this is tested in the integration test added in #3584

PR #3581, opened 2026-01-06
[mxfp8 moe training] mxfp8 unpermute autograd function

danielvegamyhre:
Stacked PRs:
 * #3606
 * #3585
 * #3584
 * #3583
 * #3582
 * __->__#3581
 * #3580
 * #3579

--- --- ---

### Tests
- There are no useful tests to add for forward() here, because the autograd function forward() uses the same code as the bf16 impl for forward. The difference where MXFP8 is used is in **backward**.
- Tests for backward require integration with other EP pipeline components, so this is tested in the integration test added in https://github.com/pytorch/ao/pull/3584

### DIFF
--- a/torchao/prototype/moe_training/ep/unpermute.py
+++ b/torchao/prototype/moe_training/ep/unpermute.py
@@ -1,0 +1,156 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import torch
+
+from torchao.prototype.mx_formats.mx_tensor import MXTensor
+
+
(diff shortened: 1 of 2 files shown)


torchao/prototype/moe_training/ep/unpermute.py
```diff
@@ -0,0 +1,180 @@
...
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import torch
+
+from torchao.prototype.mx_formats.mx_tensor import MXTensor
+
+
+class _Unpermute(torch.autograd.Function):
```
vkuzo: name seems ambiguous

danielvegamyhre: Updated with more descriptive names

vkuzo (approved):
didn't read closely, stamp for prototype

PR #3582, opened 2026-01-06
[mxfp8 moe training] mxfp8 a2a_combine autograd function

danielvegamyhre:
Stacked PRs:
 * #3606
 * #3585
 * #3584
 * #3583
 * __->__#3582
 * #3581
 * #3580
 * #3579

--- --- ---

### Tests
- There are no useful tests to add for forward() here, because the autograd function forward() uses the same code as the bf16 impl for forward. The difference where MXFP8 is used is in backward.
- Tests for backward require integration with other EP pipeline components, so this is tested in the integration test added in https://github.com/pytorch/ao/pull/3584

### DIFF
--- a/torchao/prototype/moe_training/ep/a2a_combine.py
+++ b/torchao/prototype/moe_training/ep/a2a_combine.py
@@ -1,0 +1,190 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+import torch
+import torch.distributed as dist
+from torch.distributed._functional_collectives import all_to_all_single
+
+from torchao.prototype.mx_formats.config import ScaleCalculationMode
+from torchao.prototype.mx_formats.kernels import triton_to_mxfp8_dim0
+from torchao.prototype.mx_formats.mx_tensor import MXTensor
+
+
+class _A2ACombineHPFwdMXFP8Bwd(torch.autograd.Function):
+    """
+    All-to-all combine with MXFP8 quantization in backward.
+
+    Forward:
+        - Takes bf16 input
+        - Performs all-to-all in bf16 (no quantization)
+        - Returns bf16 output
+
+    Backward:
+        - Takes bf16 gradient input
+        - Dynamically quantizes to mxfp8
+        - Performs inverse all-to-all on qdata and scales
+        - Returns MXTensor wrapping the gradient qdata and scales
+    """
+
+    @staticmethod
+    def forward(
+        ctx,
+        input: torch.Tensor,
+        output_splits: list[int],
+        input_splits: list[int],
+        group: dist.ProcessGroup,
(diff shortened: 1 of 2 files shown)


torchao/prototype/moe_training/ep/a2a_combine.py
```diff
@@ -0,0 +1,190 @@
...
+import torch
+import torch.distributed as dist
+from torch.distributed._functional_collectives import all_to_all_single
+
+from torchao.prototype.mx_formats.config import ScaleCalculationMode
+from torchao.prototype.mx_formats.kernels import triton_to_mxfp8_dim0
+from torchao.prototype.mx_formats.mx_tensor import MXTensor
+
+
+class _A2ACombine(torch.autograd.Function):
```
vkuzo: name should explain quantization

torchao/prototype/moe_training/ep/a2a_combine.py
```diff
@@ -0,0 +1,190 @@
...
+                output_split_sizes=ctx.input_splits,
+                input_split_sizes=ctx.output_splits,
+                group=ctx.group,
+            )
+            grad_input = torch.ops._c10d_functional.wait_tensor(grad_input)
+
+        return grad_input, None, None, None, None, None, None
+
+
+def a2a_combine(
```
vkuzo: name more descriptive

torchao/prototype/moe_training/ep/a2a_combine.py
```diff
@@ -0,0 +1,190 @@
...
+    input: torch.Tensor,
+    output_splits: list[int],
+    input_splits: list[int],
+    group: dist.ProcessGroup,
+    scaling_mode: ScaleCalculationMode = ScaleCalculationMode.RCEIL,
+    block_size: int = 32,
+    mxfp8_bwd: bool = True,
```
vkuzo, line 165: why do we need both

torchao/prototype/moe_training/ep/a2a_combine.py
danielvegamyhre: updated

torchao/prototype/moe_training/ep/a2a_combine.py
danielvegamyhre: updated

torchao/prototype/moe_training/ep/a2a_combine.py
danielvegamyhre: to allow for easier accuracy experiments, since when using low precision people often do ablations studying accuracy and perf for quantizing different individual parts of the model, different combinations, etc. To find some optimal trade-off they're happy with and de-risk the large pretraining run.

to use the torch default high precision impl the user can just flip the flag, instead of importing different stuff, adding branching logic, using `wait_tensor` (can be confusing at first) etc - all of that is done for you already.

this idea didn't occur to me until this point in the dev process, so the other primitives don't have similar flags yet, but i am keen on adding them /making them consistent.

what do you think?

PR #3583, opened 2026-01-06
[mxfp8 moe training] handle pre-quantized inputs/grads in forward/backward of _MXFP8GroupedMM autograd func

danielvegamyhre:
Stacked PRs:
 * #3606
 * #3585
 * #3584
 * __->__#3583
 * #3582
 * #3581
 * #3580
 * #3579

--- --- ---

### [mxfp8 moe training] handle pre-quantized inputs/grads in forward/backward of _MXFP8GroupedMM autograd func

- In the mxfp8 ep pipeline, the forward inputs may be prequantized, and backward inputs (upstream grads) may be quantized. So we need to add handling for this in both places.
- Assert `wgrad_with_hp` is true if input is pre-quantized (MXTensor). We can revisit this if/when we have a fast fused kerel for "dequant dim0 row major -> requant dim1 col major"
- This PR also refactors to simplify the code  since it was getting a bit complicated / branchy.

## Tests
- `pytest test/prototype/moe_training/test_scaled_grouped_mm.py`

### DIFF
--- a/torchao/prototype/moe_training/scaled_grouped_mm.py
+++ b/torchao/prototype/moe_training/scaled_grouped_mm.py
@@ -20,6 +20,7 @@ from torchao.prototype.moe_training.kernels.mxfp8 import (
     mx_block_rearrange_2d_M_groups_cuda,
     mxfp8_quantize_cuda_3d,
     triton_mx_block_rearrange_2d_K_groups,
+    triton_mx_block_rearrange_2d_M_groups,
     triton_mx_block_rearrange_per_group_3d,
 )
 from torchao.prototype.moe_training.utils import (
@@ -29,8 +30,11 @@ from torchao.prototype.mx_formats.config import (
     MXFP8Dim1CastKernelChoice,
     ScaleCalculationMode,
 )
-from torchao.prototype.mx_formats.kernels import triton_to_mxfp8_dim0
-from torchao.prototype.mx_formats.mx_tensor import to_mx
+from torchao.prototype.mx_formats.kernels import (
+    triton_mxfp8_dequant_dim0,
+    triton_to_mxfp8_dim0,
+)
+from torchao.prototype.mx_formats.mx_tensor import MXTensor, to_mx
 from torchao.prototype.mx_formats.utils import _to_mxfp8_dim1_kernel_wrapper
 from torchao.quantization.quantize_.common import KernelPreference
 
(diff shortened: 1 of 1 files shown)


torchao/prototype/moe_training/scaled_grouped_mm.py
```diff
@@ -282,196 +286,379 @@ def backward(ctx, grad_output: torch.Tensor):
...
-                scaling_mode=scale_calculation_mode,
-            )
-            B_scales, B_data = to_mx(
-                B_t.transpose(-2, -1),
-                elem_dtype=torch.float8_e4m3fn,
-                block_size=block_size,
-                scaling_mode=scale_calculation_mode,
+        if isinstance(input_act, MXTensor):
+            assert wgrad_with_hp, (
+                "wgrad_with_hp must be True when input_act is MXTensor (pre-quantized)"
```
vkuzo: this should say something like "only wgrad_with_hp is supported in this path, TODO implement the others"

torchao/prototype/moe_training/scaled_grouped_mm.py
```diff
@@ -708,10 +896,11 @@ def _to_mxfp8_then_scaled_grouped_mm(
         - offs (int32 torch.Tensor): The offsets to use to mark the end index of each group along the dim0 of the A tensor.
         - block_size (int): The block size to use for mxpf8 quantization. Currently only 32 is supported.
         - out_dtype (Optional[torch.dtype]): The dtype of the output tensor. Default is torch.bfloat16.
-        - emulated (bool): Whether to use the emulated mxpf8 scaled grouped mm kernel (for testing).
+        - emulated (bool): Deprecated parameter, kept for backwards compatibility. Ignored.
```
vkuzo: why is this deprecated? this is important for CI

torchao/prototype/moe_training/scaled_grouped_mm.py
```diff
@@ -708,10 +896,11 @@ def _to_mxfp8_then_scaled_grouped_mm(
...
         - use_triton_for_dim0_cast (bool): Whether to use Triton for the dim0 cast. Default true. If false, use torch native implementation.
         - wgrad_with_hp (bool): Whether to compute weight gradients in high precision.
         - scale_calculation_mode (ScaleCalculationMode): The mode to use for scale calculation.
+        - use_cuda_for_blocked_layout (bool): Whether to use CUDA kernel for per-group scale factor blocked layout conversion.
```
vkuzo: `cuda_kernel`

torchao/prototype/moe_training/scaled_grouped_mm.py
danielvegamyhre: we now have access to `linux.dgx.b200` now for CI, so don't need the emulation for it. my current plan is to:

1. add b200 ci worfklow yaml with some tests we want to run
2. then fully remove these emulated funcs in follow up PRs after this

vkuzo: I think it's better to support emulation everywhere for the convenience of being able to run most of the API (except real kernels) on most CUDA machines.  This is useful pretty often, and will be useful until it will be easy to get a b200 for any developer at any time.

danielvegamyhre: i totally agree in principle, and if you feel strongly, i can create an issue w/ specific items needed to support emulated mode, could be a good BE project. 

however, at this stage of the project, any meaningful work that can be done will require sm100, all tests/benchmarks are gated on sm100, and anyone on the team can easily get access to a b200 relatively instantly now, which wasn't the case ~6 months ago.

torchao/prototype/moe_training/scaled_grouped_mm.py
danielvegamyhre: Updated to message similar to this

torchao/prototype/moe_training/scaled_grouped_mm.py
danielvegamyhre: discussed offline, going to support emulated mode for non-sm100 CI and dev envs to prevent incidental breakage https://github.com/pytorch/ao/issues/3598

PR #3586, opened 2026-01-07
[CPU] improve quantization for int8 sdpa

Valentine233:
Description:

1. Avoid duplicated `min`/`max` operations inside and outside the `vec::conversion`.
2. Further improvement the conversion from fp32 to uint8 by adopting `SaturateU8` intrinsic.

Validation on ViT:

1. Accuracy is the same as before.
2. E2E performance improves 1%.

### DIFF
--- a/torchao/prototype/inductor/codegen/cpp_int8_sdpa_template.py
+++ b/torchao/prototype/inductor/codegen/cpp_int8_sdpa_template.py
@@ -45,6 +45,7 @@ inline void store(scalar_t* dst, at::vec::Vectorized<scalar_t> src, int size=at:
   src.store(dst, size);
 }
 
+#ifdef CPU_CAPABILITY_AVX512
 // A fast version for uint8_t conversion with SaturateU8 intrinsic
 at::vec::Vectorized<uint8_t> inline convert_float_to_uint8(
     at::vec::Vectorized<float> src) {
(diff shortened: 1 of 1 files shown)


Valentine233:
@mingfeima, please help review the PR.

torchao/prototype/inductor/codegen/cpp_int8_sdpa_template.py
```diff
@@ -45,8 +45,26 @@
...
+  // uint32 to uint8
+  auto v3 = _mm512_cvtusepi32_epi8(v2);
+  return _mm512_castsi128_si512(v3);
+}
+
+template <typename scalar_t>
+inline typename std::enable_if_t<std::is_same_v<scalar_t, unsigned char>, void>
+store(scalar_t* dst, at::vec::Vectorized<float> src, int size=at::vec::Vectorized<float>::size()) {
+  auto res = convert_float_to_uint8(src);
+  res.store(dst, size);
```
mingfeima, line 71: in case `size = at::vec::Vectorized<float>::size()` which is 16

`res` should be 128bit, `__m128i`. we should be using store of 128bit in `convert_float_to_uint8`?

torchao/prototype/inductor/codegen/cpp_int8_sdpa_template.py
```diff
@@ -45,8 +45,26 @@
   src.store(dst, size);
 }
 
+// A fast version for uint8_t conversion with SaturateU8 intrinsic
+at::vec::Vectorized<uint8_t> inline convert_float_to_uint8(
```
howardzhang-cv, line 50: still new to this code so not sure of this, but should there be some guard here to ensure AVX512 capability before using these operations?

Valentine233: Thanks, you are right! Although we have the AVX512 guard outside of int8 sdpa template, we can still add a guard here to ensure the safety. The guard is added now.

howardzhang-cv:
Still a bit new so ideally we can get some other people to take a look before merging, but it looks good to me!

Valentine233:
Thanks for your review! @howardzhang-cv 
Do you have any more comment? @drisspg @jerryzh168

PR #3590, opened 2026-01-07
Fix the slicing error

Stonepia:
Fixes https://github.com/pytorch/pytorch/issues/164872
This PR adds the `implement_torch_function` for `__getitem__`, so when slicing, it won't trigger the "Cannot set version_counter for inference tensor"

cc: @jerryzh168 @bdhirsh


Stonepia:
Honestly, I am not very sure this change is ok. Seems that I am pushing too many code into the fp8 support, but these code are not elegant at all and seems more like a workaround...

PR #3591, opened 2026-01-07
Remove external wrapper in SmoothQuant

namgyu-youn:
**Summary**:
SmoothQuant used an external wrapper (`WeightTensorWithLinearActivationScaleMetadata`, `to_weight_tensor_with_linear_activation_scale_metadata`), which handles metadata for the pre-scaling factor. This approach makes SmoothQuant implementation ambiguous

Instead of using this wrapper, we want to store pre-scaling factor (`act_pre_scale`) inside the quantization subclass like `Int4Tensor` and `Int8Tensor`. This opens orthogonal approaches like AWQ+HQQ or SmoothQuant+HQQ, which can be improved compared to AWQ/SmoothQuant.

**Test plan**:
```
pytest -sv test/quantization/quantize_/workflows/int8/test_int8tensor.py
```

### DIFF
--- a/torchao/quantization/quantize_/workflows/int8/int8_tensor.py
+++ b/torchao/quantization/quantize_/workflows/int8/int8_tensor.py
@@ -55,8 +55,7 @@ class Int8Tensor(TorchAOBaseTensor):
         scale: scale factors for dequantization
 
     Non-Tensor Attributes:
-        granularity: the granularity for quantization
-            (e.g., PerRow(), PerTensor())
+        granularity: the granularity for quantization (e.g., PerRow(), PerTensor())
         act_quant_kwargs: flags for dynamic activation quantization
 
     Optional Tensor Data Attributes:


vkuzo:
we should store the final scaling factor in `act_pre_scale` and not have any wrappers

also, +1 to removing `WeightTensorWithLinearActivationScaleMetadata`, we should just use the tensors directly and set `act_pre_scale` as needed

namgyu-youn:
> we should store the final scaling factor in `act_pre_scale` and not have any wrappers
> 
> also, +1 to removing `WeightTensorWithLinearActivationScaleMetadata`, we should just use the tensors directly and set `act_pre_scale` as needed

Oh I missed mixed algorithm case. We can use orthogonal approaches like AWQ+HQQ, by storing `act_pre_scale` in a tensor directly, right?

https://github.com/pytorch/ao/blob/aabe7dd9cf650f300f120f2b09f3868fdf3efbe6/.github/scripts/torchao_model_releases/quantize_and_upload.py#L728-L741

Storing `act_pre_scale` inside the tensor subclass should be better, I will move into it.

namgyu-youn:
@vkuzo Updated based on your suggestion. Could you please take a look into it?

vkuzo:
looks reasonable to me, thanks for working on this! @jerryzh168  should also take a look

namgyu-youn:
@jerryzh168 gentle ping to request a review :)

namgyu-youn:
Hmm... unfortunately I just found this PR is duplicate (#3468). Not sure why there is too much change in #3468, but close this one.

PR #3595, opened 2026-01-07
Add deprecation warning for `Float8DynamicActivationFloat8SemiSparseWeightConfig`

jcaip:
att


torchao/quantization/quant_api.py
```diff
@@ -1758,6 +1758,11 @@ def __post_init__(self):
 def _float8_dynamic_activation_float8_semi_sparse_weight_transform(
     module: torch.nn.Module, config: Float8DynamicActivationFloat8SemiSparseWeightConfig
 ):
+    warnings.warn(
```
andrewor14, line 1761: should this warning be in `__post_init__` instead?

andrewor14 (approved):

torchao/quantization/quant_api.py
jcaip: All the other warnings are in the transform function, so I put this there to be consistent.

Issue #3596, opened 2026-01-07
Deprecation for version=1 Int8WeightOnlyConfig and Int8DynamicActivationInt8WeightConfig


jcaip:
As part of #2752 we added a new Int8Tensor that does not use AQT in #3407. 

We plan to deprecate version=1 support of `Int8WeightOnlyConfig` and `Int8DynamicActivationInt8WeightConfig` and migrate users to the new workflow. 

version=1 support will be removed in a future torchao release

namgyu-youn:
@jcaip hey, sorry for bothering this, but I just opened PR for deprecation (same as this issue). Could you look into 
https://github.com/pytorch/ao/pull/3592 ?

PR #3597, opened 2026-01-07
Update Int8WeightOnlyConfig and Int8DynamicActivationInt8WeightConfig to use version=2 by default

jcaip:
Summary:

att, I also adding a deprecation warning to both configs pointing to #3596 

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:


torchao/quantization/quant_api.py
```diff
@@ -1265,6 +1265,9 @@ def __post_init__(self):
 
 def _int8_dynamic_activation_int8_weight_quantize_tensor(weight, config):
     if config.version == 1:
+        warnings.warn(
```
andrewor14: same here, probably better if this is in `__post_init__`

andrewor14:
Seems like this is actually BC-breaking. Did you get the chance to check all internal callsites? There may be users using this code path today

Issue #2815, opened 2025-08-20
Benchmark AWQ and SmoothQuant within vLLM ecosystem


namgyu-youn:
SmoothQuant and AWQ operate using a similar intuition, scaling up weights and scaling down activations. For deeper understanding, we want to benchmark them within the vLLM ecosystem for model quality evaluation, using `vllm/lm-eval` and `vllm/benchmarks/benchmark_latency`: https://vllm-ascend.readthedocs.io/en/latest/developer_guide/evaluation/using_lm_eval.html, https://github.com/vllm-project/vllm/blob/main/benchmarks/benchmark_latency.py.

Ideally, we want to update the user guide and docs and rely on vLLM ecosystem for the following: https://github.com/pytorch/ao/blob/main/torchao/prototype/awq/example.py, https://github.com/pytorch/ao/blob/main/torchao/prototype/awq/README.md, https://github.com/pytorch/ao/blob/main/torchao/prototype/smoothquant/example.py, https://github.com/pytorch/ao/blob/main/torchao/prototype/smoothquant/README.md

- Related Issue/PR: https://github.com/pytorch/ao/pull/2728#discussion_r2285972094 https://github.com/mit-han-lab/llm-awq/issues/130

namgyu-youn:
@jerryzh168 Hi, I have a question while addressing this issue.

For the benchmark, we need three metrics: 1) latency/throughput (using TTFT and TPOT), 2) memory usage, 3) accuracy (perplexity; ppl). We previously agreed on fully-customized datasets (tasks) for calibration flow in https://github.com/pytorch/ao/pull/2728#discussion_r2317503241, but wondering if `get_calib_dataset()` can substitute the calibration flow in `TransformerEvalWrapper()` although there is accuracy loss.

My reasoning is mostly about OSS (HF/vLLM) support:
1. `lm_eval` is only designed to evaluate accuracy (ppl), not evaluating 1) and 2), and is not compatible with vLLM/HF.
2. Based on the above, we should produce our own benchmark for latency and memory usage, making too complex.
3. Unlike `lm_eval`, the Hugging Face (HF) format is a widely used case, opening tickets to [HF - Benchmarks](https://huggingface.co/docs/transformers/main/en/benchmarks) and [vLLM - Benchmark](https://vast.ai/article/how-to-benchmark-an-LLM-with-vLLM-in-10-minutes).

[...]

PR #3602, opened 2026-01-08
Add AWQ/SmoothQuant to benchmark module

namgyu-youn:
**Summary**:
Add AWQ/SmoothQuant to the benchmark module. `vllm bench` didn't work and will be fixed in a future PR.
- fix: #2815 

**Test plan**:
```
./benchmarks/quantization/calibration_based/measure_accuracy_and_performance.sh awq_int4_weight_only meta-llama/Llama-3.2-1B
./benchmarks/quantization/calibration_based/measure_accuracy_and_performance.sh smoothquant_int8 meta-llama/Llama-3.2-1B
```

**Results** on A100:
```
+----------------------+--------------+--------------+--------------+--------------+
| Recipe               |   Checkpoint |     Wikitext |   Winogrande |   Winogrande |
|                      |         (GB) |   Perplexity |          Acc |       Stderr |
+======================+==============+==============+==============+==============+
| awq_int4_weight_only |         1.06 |      13.5956 |       0.6022 |       0.0138 |
+----------------------+--------------+--------------+--------------+--------------+
| smoothquant_int8     |         1.52 |      15.6757 |       0.5762 |       0.0139 |
+----------------------+--------------+--------------+--------------+--------------+
```


namgyu-youn:
@pytorchbot label "topic: for developers"

also cc @vkuzo working on this benchmark :)

benchmarks/quantization/create_quantized_model.py
```diff
@@ -9,25 +9,131 @@
 
 import torch
 from transformers import AutoModelForCausalLM, AutoTokenizer, TorchAoConfig
+from transformers.quantizers.auto import get_hf_quantizer
 from utils import string_to_config
 
+from torchao._models._eval import TransformerEvalWrapper
```
vkuzo: we were trying to get away from needing wrappers :)

is there a way to calibrate without a wrapper like this one?

benchmarks/quantization/utils.py
```diff
@@ -47,5 +47,16 @@ def string_to_config(s):
             use_dynamic_per_tensor_scale=True,
             use_triton_kernel=True,
         )
+    elif s == "awq_int4_weight_only":
```
vkuzo: either have a new function for these more complicated objects, or ensure that existing callsites are updated.  `string_to_config` should have a simple return type that doesn't vary based on the arguments.

vkuzo:
i think it may be simpler to have a separate script for calibration based workflows, that way the no-calibration script can stay simple

benchmarks/quantization/create_quantized_model.py
namgyu-youn: Yes actually we are already using `lm-eval` directly in AWQ example

https://github.com/pytorch/ao/blob/b78bd40e020436d135acd06e20012be87075fe37/torchao/prototype/awq/example.py#L98-L121

. Maybe we can also update other workflows not to use wrapper in a future PR?

namgyu-youn: Deprecation for wrapper (`TransformerEvalWrapper`): #3617

namgyu-youn:
Updated to separate AWQ/SmoothQuant flows entirely. It can be summarized as:
- Workflow: `quantize_and_upload.py` vs. `calibration.py`
- Configs: `string_to_config()` vs. `string_to_calibration_config`

vkuzo:
> i think it may be simpler to have a separate script for calibration based workflows, that way the no-calibration script can stay simple

to clarify, I meant a separate bash script, to keep everything simple in the main production benchmark scripts

the calibration flows are all prototype, IMO it's cleaner to keep it separate and consider merging to main flow as they mature

namgyu-youn:
> > i think it may be simpler to have a separate script for calibration based workflows, that way the no-calibration script can stay simple
> 
> to clarify, I meant a separate bash script, to keep everything simple in the main production benchmark scripts
> 
> the calibration flows are all prototype, IMO it's cleaner to keep it separate and consider merging to main flow as they mature

Understood, separating them entirely is much better for maintenance. Could you check the updated (separated) workflows?

benchmarks/quantization/calibration_based/measure_accuracy_and_performance.sh
```diff
@@ -0,0 +1,164 @@
...
+
+set -e
+
+# Define all available calibration-based quantization recipes
+QUANT_RECIPES_ALL=(
+  "awq_int4_weight_only"
+  "smoothquant_int8"
+)
+
+# Define recipes that are known to be broken in vllm
```
vkuzo: let's just add vllm later? not useful to add it now since it does not work yet

benchmarks/quantization/calibration_based/create_quantized_model.py
```diff
@@ -0,0 +1,89 @@
...
+        tasks=tasks,
+        limit=limit,
+        batch_size=1,
+    )
+    quantize_(model, config_class(base_config, step="convert"), filter_fn=filter_fn)
+    load_config = config_class(base_config, step="prepare_for_loading")
+    model.config.quantization_config = TorchAoConfig(load_config)
+
+
+def quantize_model_and_save(model, recipe, base_config, output_dir, tasks, limit):
```
vkuzo: nit: add types to vars, and also the name for `base_config` should be something like `base_config_cls`

benchmarks/quantization/calibration_based/measure_accuracy_and_performance.sh
```diff
@@ -0,0 +1,164 @@
...
+MODEL_ID="${2:-meta-llama/Llama-3.1-8B}"
+
+# Get log file as third positional argument (optional)
+LOG_FILE="${3:-benchmarks/data/calibration_based/accuracy_and_performance_log.txt}"
+
+# Get calibration tasks as fourth positional argument (optional)
+CALIB_TASKS="${4:-wikitext}"
+
+# Get calibration limit as fifth positional argument (optional)
+CALIB_LIMIT="${5:-10}"
```
vkuzo: is 10 a good default? Usually people use 100-500 samples.

namgyu-youn: Low iteration is for fast experiment, even though we lose stability. I remember 128/256/512 are used in many papers, and we can lead external users by filing docs.

benchmarks/quantization/calibration_based/measure_accuracy_and_performance.sh
namgyu-youn: Actually I am planning to debug into it after this PR. How about keeping this block for faster debugging?

benchmarks/quantization/calibration_based/create_quantized_model.py
namgyu-youn: Done

benchmarks/quantization/calibration_based/measure_accuracy_and_performance.sh
vkuzo: i'd say adding commented out / broken code with the intention to fix it later is a bit of an anti-pattern, it's usually cleaner to just add it in the PR where it works.  If you add it here there is no way to test it and it's a bit confusing to people who read your code.

benchmarks/quantization/calibration_based/measure_accuracy_and_performance.sh
vkuzo: right, the default should be targeted towards useful accuracy numbers, not debugging

benchmarks/quantization/calibration_based/measure_accuracy_and_performance.sh
namgyu-youn: Understood, thanks for pointing it out.

benchmarks/quantization/calibration_based/measure_accuracy_and_performance.sh
namgyu-youn: I see, updated to 256. Let me know if a higher one (maybe 512) is needed.

namgyu-youn:
vLLM failure log:
- AWQ: https://gist.github.com/namgyu-youn/dff3e22320b028b28f7d533727a88bb1
- SmoothQuant: https://gist.github.com/namgyu-youn/0dca97ff669cfebfcb3af522ae10ea83

namgyu-youn:
@vkuzo could you please look again this?

benchmarks/quantization/calibration_based/measure_accuracy_and_performance.sh
```diff
@@ -0,0 +1,121 @@
...
+#   ./measure_accuracy_and_performance.sh awq_int4_weight_only           # Run single recipe
+#   ./measure_accuracy_and_performance.sh all meta-llama/Llama-3.2-1B    # Custom model with all recipes
+#   ./measure_accuracy_and_performance.sh awq_int4_weight_only meta-llama/Llama-3.2-8B my_log.txt  # All custom args
+#   SKIP_MODEL_CREATE=1 ./measure_accuracy_and_performance.sh all        # Skip model creation, only run eval
+#   SKIP_LM_EVAL=1 ./measure_accuracy_and_performance.sh all             # Skip lm_eval, only create models
+
+set -e
+
+# Define all available calibration-based quantization recipes
+QUANT_RECIPES_ALL=(
```
vkuzo, line 36: nit: this seems to be missing a baseline (bfloat16 without quantization), without a baseline it's hard to interpret whether the results of quantization are good or not

benchmarks/quantization/calibration_based/utils.py
```diff
@@ -0,0 +1,24 @@
...
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+from torchao.quantization import (
+    Int4WeightOnlyConfig,
+    Int8DynamicActivationInt8WeightConfig,
+)
+
+
+def string_to_calibration_config(s):
```
vkuzo, line 12: nit: just inline this to your main function, since you don't have any other users of this

benchmarks/quantization/calibration_based/measure_accuracy_and_performance.sh
namgyu-youn: How about running "None" from upstream then? Or we can run `lm-eval` and `vllm serve` directly for original model.

benchmarks/quantization/calibration_based/utils.py
namgyu-youn: Isn't separating base configs better for (1) safe logic (not all subclass supports AWQ/SmoothQuant yet), (2) future extensibility (like GPTQ)? 

Although mapping pattern is quiet complex, there might be only a few code reductions with inline.

benchmarks/quantization/calibration_based/measure_accuracy_and_performance.sh
vkuzo: right, just make it easy for the user by including that logic here? 10x easier to just do it by default vs requiring the user to jump to a different script

namgyu-youn: I think we can lead users to upstream, even though their might be confusion. 

It's mostly because this module should run calibration flow, instead of direct `lm-eval & vllm` run like upstream. We can build skip logic for None case, but it would make it hard to maintain.

namgyu-youn: @vkuzo oh sorry, after careful look, I found we can just `pass` calibration-flow. Added "None" option, sorry for my bad again.

vkuzo (approved):

vkuzo:
thanks for working on this! would you mind rebasing so we can run a fresh CI

namgyu-youn:
> thanks for working on this! would you mind rebasing so we can run a fresh CI

Sure, finished rebase; could you trigger ci?

namgyu-youn:
CI failure seems related to network issue (timeout), maybe rerun to fix?

Issue #3599, opened 2026-01-08
MXFP8 MoE training prototype graduation tracker


danielvegamyhre:
- [ ] Support emulated mode #3598 
- [ ] Fix torch reference for per group blocked layout for groups along K
- [ ] Add B200 to CI #2964 and add tests to workflow triggered by label

PR #3603, opened 2026-01-08
[mxfp8 moe training] fix torch ref impl of SF blocked layout per group along K

danielvegamyhre:
Stacked PRs:
 * #3604
 * __->__#3603

--- --- ---

### [mxfp8 moe training] fix torch ref impl of SF blocked layout per group along K

Part of #3599

### DIFF
--- a/torchao/prototype/moe_training/kernels/mxfp8/quant.py
+++ b/torchao/prototype/moe_training/kernels/mxfp8/quant.py
@@ -93,6 +93,7 @@ def torch_to_blocked_2d_K_groups(
         blocked_scales: Tensor
         start_row_after_padding: Tensor of shape (num_groups,) which contains the start row after padding for each group.
     """
+    # TODO: add diagram of this transformation to the docs, and link in the docstring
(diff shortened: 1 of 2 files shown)


torchao/prototype/moe_training/kernels/mxfp8/quant.py
vkuzo, line 94: would be nice to include a simple example of toy input values and what this function calculates for output values

vkuzo (approved):
stamp

torchao/prototype/moe_training/kernels/mxfp8/quant.py
danielvegamyhre: i could link to a diagram, the visualization makes it much easier to understand. i added a TODO to add this to the docs, with a link here. 

a proper diagram is needed to show (1) the layout transformation within the tile,  (2) how tiles within a group are laid out in memory together, and (3) how different groups of scale factor tiles are laid out in memory with respect to each other.

PR #3610, opened 2026-01-09
[claude] Make MXFP8 cuda kernels ABI stable

andrewor14:
Test Plan:
```
pytest test/prototype/mx_formats/test_kernels.py -v -s -k cuda
pytest test/prototype/moe_training/test_kernels.py -v -s -k cuda_mx_block
pytest test/prototype/moe_training/test_training.py -v -s
```

Prompt:
```
Make these two files ABI stable:
torchao/csrc/cuda/mx_kernels/mxfp8_cuda.cu
torchao/csrc/cuda/mx_kernels/mxfp8_extension.cpp

Read these for instructions:
pytorch/docs/source/notes/libtorch_stable_abi.md
cppdocs/_sources/stable.rst.txt

Use these files for an example:
Before: flash-attention/hopper/flash_api.cpp
After: flash-attention/hopper/flash_api_stable.cpp

Additional instructions:
Replace at::zeros with torch::stable::new_zeros
Replace at::empty with torch::stable::new_empty
Replace TORCH_CHECK with STD_TORCH_CHECK without changing the content of the check, just replace the function name
Replace c10::cuda::CUDAGuard with DeviceGuard
When calling tensor.stride(), no need to check for tensor.numel() > 0 beforehand
When using cudaStream_t, remember to include cuda_runtime.h
When calling aoti_torch_get_current_cuda_stream, get the device index from a tensor, not from the general torch::stable::accelerator::getCurrentDeviceIndex();

Don’t do these things:
Don’t use torch::stable::new_empty_strided since this API does not exist
Don’t define a dummy _C module that can be accessed from python
Don’t declare aoti_torch_get_current_cuda_stream, just include it from torch/csrc/inductor/aoti_torch/c/shim.h and add -DUSE_CUDA to both cxx and nvcc in setup.py
Don’t box kernels manually, just use TORCH_BOX
Don’t change switch statements into if else blocks
```

Follow-up prompts:
```
Revert the content changes of STD_TORCH_CHECK, just replace the function name

For the changes that involved tensor.data_ptr<some_dtype>() in the old code, revert the changes and replace it with tensor.const_data_ptr<some_dtype>()
```

Output:
```
● I've successfully made both files ABI stable. Here's a summary of the changes:

  Changes to mxfp8_cuda.cu:

  1. Updated includes: Replaced ATen and torch/extension.h with stable ABI headers (torch/csrc/stable/tensor.h, torch/csrc/inductor/aoti_torch/c/shim.h, torch/headeronly/core/ScalarType.h, torch/headeronly/util/Exception.h, torch/headeronly/util/shim_utils.h)
  2. Added cuda_runtime.h for cudaStream_t
  3. Changed tensor type: torch::Tensor → torch::stable::Tensor (with using declaration)
  4. Changed scalar type enums: torch::kFloat32 → torch::headeronly::ScalarType::Float, etc.
  5. Changed error checking: TORCH_CHECK → STD_TORCH_CHECK
  6. Changed CUDA stream access: at::cuda::getCurrentCUDAStream() → aoti_torch_get_current_cuda_stream(input.get_device_index(), &stream_ptr) with TORCH_ERROR_CODE_CHECK
  7. Fixed stride access: .strides()[i] → .stride(i) and .sizes()[i] → .size(i)

  Changes to mxfp8_extension.cpp:

  1. Updated includes: Replaced torch/library.h, ATen/ATen.h, ATen/cuda/CUDAContext.h, c10/cuda/CUDAGuard.h with stable headers
  2. Changed tensor type: at::Tensor → torch::stable::Tensor
  3. Changed DeviceGuard: c10::cuda::CUDAGuard → torch::stable::accelerator::DeviceGuard (with input.get_device_index() instead of input.device())
  4. Changed error checking: TORCH_CHECK → STD_TORCH_CHECK
  5. Changed tensor creation:
    - at::empty → torch::stable::new_empty
    - at::zeros → torch::stable::new_zeros
    - For strided tensors: Created with transposed dimensions then used torch::stable::transpose to get column-major layout
  6. Changed CUDA stream access: Used aoti_torch_get_current_cuda_stream with device index from tensor
  7. Changed library registration: TORCH_LIBRARY_IMPL → STABLE_TORCH_LIBRARY_IMPL with TORCH_BOX wrapper

  Changes to setup.py:

  1. Added -DUSE_CUDA to both cxx and nvcc compiler arguments for the mxfp8 extension
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -761,7 +761,11 @@ def get_extensions():
 
         # Only add the extension if the source files exist AND we are building for sm100
         mxfp8_src_files_exist = all(os.path.exists(f) for f in mxfp8_sources)
-        if mxfp8_src_files_exist and build_for_sm100a:
+        if (
+            mxfp8_src_files_exist
+            and build_for_sm100a
+            and _torch_version_at_least("2.11.0")
+        ):
             print("Building mxfp8_cuda extension")
             ext_modules.append(
                 CUDAExtension(
@@ -775,11 +779,16 @@ def get_extensions():
                             f"-DPy_LIMITED_API={min_supported_cpython_hexcode}",
                             "-std=c++20",
                             "-O3",
+                            "-DUSE_CUDA",
+                            # define TORCH_TARGET_VERSION with min version 2.11 for ABI stable Float8_e8m0fnu
+                            "-DTORCH_TARGET_VERSION=0x020b000000000000",
                         ],
                         "nvcc": nvcc_args
                         + [
                             "-gencode=arch=compute_100,code=sm_100",
                             "-gencode=arch=compute_120,code=compute_120",
+                            "-DUSE_CUDA",
+                            "-DTORCH_TARGET_VERSION=0x020b000000000000",
                         ],
                     },
                 ),
--- a/torchao/csrc/cuda/mx_kernels/mxfp8_cuda.cu
+++ b/torchao/csrc/cuda/mx_kernels/mxfp8_cuda.cu
@@ -7,26 +7,33 @@
 // CUDA bridge for MXFP8 quantization
 
 #include "mxfp8_quantize.cuh"
-#include <ATen/cuda/CUDAContext.h>
-#include <string>
-#include <torch/extension.h>
 
+#include <torch/csrc/stable/tensor.h>
+#include <torch/csrc/inductor/aoti_torch/c/shim.h>
+#include <torch/headeronly/core/ScalarType.h>
+#include <torch/headeronly/util/Exception.h>
+#include <torch/headeronly/util/shim_utils.h>
+
+#include <cuda_runtime.h>
+#include <string>
+
+using torch::stable::Tensor;
 
 namespace mxfp8 {
 
 // Convert PyTorch scalar type to our DType enum
-DType get_input_dtype(const torch::Tensor &t) {
+DType get_input_dtype(const Tensor &t) {
   switch (t.scalar_type()) {
-  case torch::kFloat32:
+  case torch::headeronly::ScalarType::Float:
     return DType::kFloat32;
-  case torch::kFloat16:
+  case torch::headeronly::ScalarType::Half:
     return DType::kFloat16;
-  case torch::kBFloat16:
+  case torch::headeronly::ScalarType::BFloat16:
     return DType::kBFloat16;
-  case torch::kUInt8:
+  case torch::headeronly::ScalarType::Byte:
     return DType::kByte;
   default:
-    TORCH_CHECK(false, "Unsupported input tensor dtype: ", t.scalar_type());
+    STD_TORCH_CHECK(false, "Unsupported input tensor dtype: ", t.scalar_type());
   }
 }
 
@@ -36,7 +43,7 @@ ScaleCalculationMode get_scaling_mode(const std::string &scaling_mode) {
   } else if (scaling_mode.compare("rceil") == 0) {
       return ScaleCalculationMode::RCEIL;
   } else {
-      TORCH_CHECK(false, "Unsupported scaling mode: ", scaling_mode, ". Only ['floor', 'rceil'] are supported.");
+      STD_TORCH_CHECK(false, "Unsupported scaling mode: ", scaling_mode, ". Only ['floor', 'rceil'] are supported.");
   }
 }
 
@@ -45,16 +52,16 @@ DType get_output_dtype(const std::string &fp8_format) {
   if (fp8_format.compare("e4m3") == 0) {
     return DType::kFloat8E4M3;
   } else {
-    TORCH_CHECK(false, "Unsupported FP8 format: ", fp8_format,
+    STD_TORCH_CHECK(false, "Unsupported FP8 format: ", fp8_format,
                 ". Only 'e4m3' is supported.");
   }
 }
 
-void mxfp8_quantize_cuda(const torch::Tensor &input,
-                         torch::Tensor &output_rowwise,
-                         torch::Tensor &output_colwise,
-                         torch::Tensor &scales_rowwise,
-                         torch::Tensor &scales_colwise,
+void mxfp8_quantize_cuda(const Tensor &input,
+                         Tensor &output_rowwise,
+                         Tensor &output_colwise,
+                         Tensor &scales_rowwise,
+                         Tensor &scales_colwise,
                          int64_t scale_dim_x,
                          int64_t scale_dim_y,
                          const std::string &fp8_format,
@@ -79,14 +86,16 @@ void mxfp8_quantize_cuda(const torch::Tensor &input,
           ? reinterpret_cast<e8m0_t *>(scales_colwise.data_ptr())
           : nullptr;
 
-  // Get CUDA stream
-  cudaStream_t stream = at::cuda::getCurrentCUDAStream();
+  // Get CUDA stream using stable ABI
+  void* stream_ptr = nullptr;
+  TORCH_ERROR_CODE_CHECK(aoti_torch_get_current_cuda_stream(input.get_device_index(), &stream_ptr));
+  cudaStream_t stream = static_cast<cudaStream_t>(stream_ptr);
 
-  // Get strides of scale ptrs
-  int64_t scale_rowwise_stride_dim0 = scales_rowwise.strides()[0];
-  int64_t scale_rowwise_stride_dim1 = scales_rowwise.strides()[1];
-  int64_t scale_colwise_stride_dim0 = scales_colwise.strides()[0];
-  int64_t scale_colwise_stride_dim1 = scales_colwise.strides()[1];
+  // Get strides of scale ptrs (guard against 1D empty tensors when rowwise/colwise is false)
+  int64_t scale_rowwise_stride_dim0 = scales_rowwise.dim() >= 2 ? scales_rowwise.stride(0) : 0;
+  int64_t scale_rowwise_stride_dim1 = scales_rowwise.dim() >= 2 ? scales_rowwise.stride(1) : 0;
+  int64_t scale_colwise_stride_dim0 = scales_colwise.dim() >= 2 ? scales_colwise.stride(0) : 0;
+  int64_t scale_colwise_stride_dim1 = scales_colwise.dim() >= 2 ? scales_colwise.stride(1) : 0;
 
 #if defined(DEBUG)
   printf("mxfp8_quantize_cuda:\n");
@@ -94,8 +103,12 @@ void mxfp8_quantize_cuda(const torch::Tensor &input,
   printf("scaling_mode: %s\n", scaling_mode.c_str());
   printf("Scale dim x: %ld\n", scale_dim_x);
   printf("Scale dim y: %ld\n", scale_dim_y);
-  printf("Rowwise scale shape: %ld x %ld\n", scales_rowwise.sizes()[0], scales_rowwise.sizes()[1]);
-  printf("Colwise scale shape: %ld x %ld\n", scales_colwise.sizes()[0], scales_colwise.sizes()[1]);
+  printf("Rowwise scale shape: %ld x %ld\n",
+         scales_rowwise.dim() >= 1 ? scales_rowwise.size(0) : 0,
+         scales_rowwise.dim() >= 2 ? scales_rowwise.size(1) : 0);
+  printf("Colwise scale shape: %ld x %ld\n",
+         scales_colwise.dim() >= 1 ? scales_colwise.size(0) : 0,
+         scales_colwise.dim() >= 2 ? scales_colwise.size(1) : 0);
   printf("scale_rowwise_stride_dim0 = %ld\n", scale_rowwise_stride_dim0);
   printf("scale_rowwise_stride_dim1 = %ld\n", scale_rowwise_stride_dim1);
   printf("scale_colwise_stride_dim0 = %ld\n", scale_colwise_stride_dim0);
--- a/torchao/csrc/cuda/mx_kernels/mxfp8_extension.cpp
+++ b/torchao/csrc/cuda/mx_kernels/mxfp8_extension.cpp
@@ -1,19 +1,29 @@
-// MXFP8 extension using TORCH_LIBRARY (CPython ABI agnostic)
-#include <torch/library.h>
-#include <ATen/ATen.h>
-#include <ATen/cuda/CUDAContext.h>
-#include <c10/cuda/CUDAGuard.h>
+// MXFP8 extension using STABLE_TORCH_LIBRARY (ABI stable)
+#include <torch/csrc/stable/tensor.h>
+#include <torch/csrc/stable/library.h>
+#include <torch/csrc/stable/ops.h>
+#include <torch/csrc/stable/accelerator.h>
+#include <torch/csrc/inductor/aoti_torch/c/shim.h>
+
+#include <torch/headeronly/core/ScalarType.h>
+#include <torch/headeronly/util/Exception.h>
+#include <torch/headeronly/util/shim_utils.h>
+
+#include <cuda_runtime.h>
 #include <cstdint>
 #include <string>
+
+using torch::stable::Tensor;
+namespace tsa = torch::stable::accelerator;
 
 namespace mxfp8 {
 
 // Forward declarations
-void mxfp8_quantize_cuda(const at::Tensor &input,
-                         at::Tensor &output_rowwise,
-                         at::Tensor &output_columnwise,
-                         at::Tensor &scales_rowwise,
-                         at::Tensor &scales_colwise,
+void mxfp8_quantize_cuda(const Tensor &input,
+                         Tensor &output_rowwise,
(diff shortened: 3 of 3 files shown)


torchao/csrc/cuda/mx_kernels/mxfp8_cuda.cu
```diff
@@ -1,26 +1,33 @@
 // CUDA bridge for MXFP8 quantization
 
 #include "mxfp8_quantize.cuh"
+#include <ATen/Tensor.h>
```
janeyx99: You should not include any headers that aren't stable!

torchao/csrc/cuda/mx_kernels/mxfp8_cuda.cu
```diff
@@ -74,7 +81,7 @@ void mxfp8_quantize_cuda(const torch::Tensor &input,
           : nullptr;
 
   // Get CUDA stream
-  cudaStream_t stream = at::cuda::getCurrentCUDAStream();
+  cudaStream_t stream = xf_getCurrentCUDAStream();
```
janeyx99: you probably don't wanna use the xf utils, there should be a stable api for this already

janeyx99 (commented):
The best way to test this is by setting up a build where you can pass in -DTORCH_TARGET_VERSION and build the kernel + ensure it builds and runs. I'd recommend setting up a separate stable wheel for the migration to make testing easier as you slowly migrate kernels to the stable wheel

Edit: I just learned that you only have 3 files to migrate. In that case you do have the option of just biting the bullet and changing the setup.py in torchao to add this flag: https://github.com/Dao-AILab/flash-attention/pull/2155/files.

Note you want 20a0 and not 2090 though, for 2.10.

torchao/csrc/cuda/mx_kernels/mxfp8_cuda.cu
```diff
@@ -1,26 +1,33 @@
...
 #include <string>
-#include <torch/extension.h>
 
+#include <torch/csrc/stable/tensor.h>
+#include <torch/csrc/inductor/aoti_torch/c/shim.h>
+#include <torch/headeronly/core/ScalarType.h>
+#include <torch/headeronly/util/Exception.h>
+
+// Declare the CUDA stream function that's behind #ifdef USE_CUDA in shim.h
+extern "C" AOTITorchError aoti_torch_get_current_cuda_stream(int32_t device_index, void** ret_stream);
```
janeyx99: I don't think you need this, you should just pass -DUSE_CUDA as a flag in setup.py.

Though hmm @mikaylagawarecki should we automatically pass this in through CUDAExtension for better UX?

torchao/csrc/cuda/mx_kernels/mxfp8_cuda.cu
```diff
@@ -1,26 +1,33 @@
...
 
 namespace mxfp8 {
 
 // Convert PyTorch scalar type to our DType enum
-DType get_input_dtype(const torch::Tensor &t) {
-  switch (t.scalar_type()) {
-  case torch::kFloat32:
+DType get_input_dtype(const Tensor &t) {
+  auto scalar_type = t.scalar_type();
+  if (scalar_type == torch::headeronly::ScalarType::Float) {
```
janeyx99: you should still be able to use a switch statement i believe, but no biggie

torchao/csrc/cuda/mx_kernels/mxfp8_extension.cpp
```diff
@@ -40,94 +50,129 @@ void launch_mx_block_rearrange_2d_M_groups_cuda(
...
-  at::Tensor output_rowwise, output_colwise;
-  at::Tensor scales_rowwise, scales_colwise;
+  Tensor output_rowwise, output_colwise;
+  Tensor scales_rowwise, scales_colwise;
 
   if (rowwise) {
     const int64_t num_col_blocks = (cols + scale_dim_x - 1) / scale_dim_x;
-    output_rowwise = at::empty({rows, cols}, options_fp8);
-    scales_rowwise = at::empty({rows, num_col_blocks}, options_scale);
+    output_rowwise = torch::stable::empty(
```
janeyx99: new_empty?

here and below

janeyx99 (commented):
Yea, this looks close. Here's when I'd recommend just building the wheel and letting the error messages and test cases get you 100% there.

torchao/csrc/cuda/mx_kernels/mxfp8_cuda.cu
mikaylagawarecki: This or -DUSE_CUDA would both work

tbh I'm not sure that silently defining USE_CUDA in CUDAExtension is good UX 🤔  (Where did USE_CUDA originate from? Is it a pytorch thing or a cuda thing? I'm wondering whether users might also define USE_CUDA themselves for separate purposes hmm, wdyt?)

torchao/csrc/cuda/mx_kernels/mxfp8_cuda.cu
```diff
@@ -128,7 +137,10 @@ void mxfp8_quantize_3d_cuda(const torch::Tensor &input,
       reinterpret_cast<e8m0_t *>(scales_colwise.data_ptr());
 
   // Get CUDA stream
-  cudaStream_t stream = at::cuda::getCurrentCUDAStream();
+  void* stream_ptr = nullptr;
+  TORCH_ERROR_CODE_CHECK(
+      aoti_torch_get_current_cuda_stream(input.get_device_index(), &stream_ptr));
```
andrewor14: @mikaylagawarecki @janeyx99 this is failing compilation:
``` 
      /home/andrewor/local/ao/torchao/csrc/cuda/mx_kernels/mxfp8_cuda.cu(141): error: identifier "aoti_torch_get_current_cuda_stream" is undefined
          if ((aoti_torch_get_current_cuda_stream(input.get_device_index(), &stream_ptr)) != 0) { torch::headeronly::detail::throw_exception("aoti_torch_get_current_cuda_stream(input.get_device_index(), &stream_ptr)", "/home/andrewor/local/ao/torchao/csrc/cuda/mx_kernels/mxfp8_cuda.cu", 141); }
               ^
      
      2 errors detected in the compilation of "/home/andrewor/local/ao/torchao/csrc/cuda/mx_kernels/mxfp8_cuda.cu".
```

What's the right way to make this compile? I'm already importing `torch/csrc/inductor/aoti_torch/c/shim.h` according to https://docs.pytorch.org/cppdocs/stable.html#getting-the-current-cuda-stream

torchao/csrc/cuda/mx_kernels/mxfp8_extension.cpp
```diff
@@ -40,89 +50,80 @@ void launch_mx_block_rearrange_2d_M_groups_cuda(
...
-    // for colwise scaling, we can avoid uncoalesced writes to global memory.
-    // This is because each of the 32 threads in a warp will be computing
-    // a scale for a different column of 32 input data values, then each writing
-    // that scale to global memory - so the stride along this `col` dim should be 1
-    // so writes can be coalesced into a single transaction.
-    scales_colwise = at::empty_strided({cols, num_row_blocks}, {1, cols}, options_scale);
+    // Create column-major tensor by creating transposed shape and transposing
+    // We need shape {rows, cols} with strides {1, rows}, so create {cols, rows} and transpose
+    Tensor output_colwise_tmp = torch::stable::new_empty(input, {cols, rows}, torch::headeronly::ScalarType::Float8_e4m3fn);
+    output_colwise = torch::stable::transpose(output_colwise_tmp, 0, 1);
```
andrewor14, line 154: @janeyx99 @danielvegamyhre is this right? Seems like claude is doing this for all column major tensors

torchao/csrc/cuda/mx_kernels/mxfp8_extension.cpp
```diff
@@ -234,16 +228,18 @@ at::Tensor mx_block_rearrange_2d_M_groups(
...
-  auto output = at::zeros({padded_rows, padded_cols},
-                            at::TensorOptions()
-                                .dtype(scales_tensor.scalar_type())
-                                .device(scales_tensor.device()));
+  auto output = torch::stable::new_zeros(scales_tensor, {padded_rows, padded_cols}, scales_tensor.scalar_type());
 
   // Get raw pointers
   const uint8_t* scales_ptr = reinterpret_cast<const uint8_t*>(scales_tensor.data_ptr());
-  const int32_t* offsets_ptr = input_group_end_offsets.data_ptr<int32_t>();
+  const int32_t* offsets_ptr = static_cast<const int32_t*>(input_group_end_offsets.data_ptr());
```
andrewor14: do we need this change?

setup.py
```diff
@@ -745,11 +748,13 @@ def get_extensions():
                             f"-DPy_LIMITED_API={min_supported_cpython_hexcode}",
                             "-std=c++17",
                             "-O3",
+                            "-DUSE_CUDA",
                         ],
                         "nvcc": nvcc_args
                         + [
                             "-gencode=arch=compute_100,code=sm_100",
                             "-gencode=arch=compute_120,code=compute_120",
+                            "-DUSE_CUDA",
```
andrewor14, line 790: @janeyx99 @mikaylagawarecki I needed to add this for `aoti_torch_get_current_cuda_stream`. Is this what you would recommend?

torchao/csrc/cuda/mx_kernels/mxfp8_extension.cpp
danielvegamyhre: looks right to me, it allocates empty shape `(cols, rows)` with strides `(rows, 1)` (row major) then transpose so we have `(rows, cols)` with strides `(1, rows)` (col major). since transpose is just a metadata change this shouldn't impact perf

danielvegamyhre: also as an aside (and a note to myself) i think the original code for the scales can be simplified, but will leave it for a future PR to avoid merge conflicts with this

torchao/csrc/cuda/mx_kernels/mxfp8_extension.cpp
mikaylagawarecki: You can use `input_group_end_offsets.const_data_ptr<int32_t>()` the stable `Tensor.data_ptr()` doesn't accept templates atm hence the issue :(

for future reference, there's also .mutable_data_ptr<T>() in stable

setup.py
```diff
@@ -745,11 +748,13 @@ def get_extensions():
                             f"-DPy_LIMITED_API={min_supported_cpython_hexcode}",
                             "-std=c++17",
                             "-O3",
+                            "-DUSE_CUDA",
```
janeyx99, line 782: also pass in a -DTORCH_TARGET_VERSION=0x020a000000000000 for 2.10

danielvegamyhre:
@andrewor14 i'll build and do some tests with this PR locally and follow up with the results

danielvegamyhre:
@andrewor14 when I run unit tests I see runtime errors like this:

```
RuntimeError: call, /home/dev/.conda/envs/torch/lib/python3.12/site-packages/torch/include/torch/csrc/stable/stableivalue_conversions.h:530, Not yet supported ScalarType 44, please file an issue describing your use case.
```

Repro command:
- `pytest test/prototype/moe_training/test_training.py -v -s     `

@janeyx99 any thoughts on this?

danielvegamyhre:
also @andrewor14 i just landed a test fix for an unrelated issue #3728, i would suggest rebasing on main now to get cleaner test signal

andrewor14:
Hmm I'm hitting a different error after rebasing and running the test:

```
pytest test/prototype/moe_training/test_training.py -v -s 
```

Error:
```
FAILED test/prototype/moe_training/test_training.py::test_moe_training[recipe_config1-False-target_fqns0] -

RuntimeError: CUDA error: CUBLAS_STATUS_INVALID_VALUE when calling
`cublasGemmEx( handle, opa, opb, m, n, k, &falpha, a, CUDA_R_16BF, lda,
b, CUDA_R_16BF, ldb, &fbeta, c, std::is_same_v<C_Dtype, float...
```
Any ideas? Wonder if it's a shape error? We did add some transposes in this PR. @danielvegamyhre

janeyx99:
> ```
> RuntimeError: call, /home/dev/.conda/envs/torch/lib/python3.12/site-packages/torch/include/torch/csrc/stable/stableivalue_conversions.h:530, Not yet supported ScalarType 44, please file an issue describing your use case.
> ```
> 
> Repro command:
> 
> * `pytest test/prototype/moe_training/test_training.py -v -s     `
> 
> @janeyx99 any thoughts on this?

Oh boy, this is cuz our shim doesn't currently support the 44th ScalarType that is Float8_e8m0fnu --> is that needed? I'll patch something for 2.11 if so, but it would mean we can't migrate this til then :/

andrewor14:
Marking this PR as draft so we don't accidentally merge it. Seems like this is blocked on Float8_e8m0fnu ABI stable support for now, probably until torch 2.11

torchao/csrc/cuda/mx_kernels/mxfp8_cuda.cu
```diff
@@ -73,23 +80,29 @@ void mxfp8_quantize_cuda(const torch::Tensor &input,
...
-  // Get strides of scale ptrs
-  int64_t scale_rowwise_stride_dim0 = scales_rowwise.strides()[0];
-  int64_t scale_rowwise_stride_dim1 = scales_rowwise.strides()[1];
-  int64_t scale_colwise_stride_dim0 = scales_colwise.strides()[0];
-  int64_t scale_colwise_stride_dim1 = scales_colwise.strides()[1];
+  // Get strides of scale ptrs (guard against 1D empty tensors when rowwise/colwise is false)
+  int64_t scale_rowwise_stride_dim0 = scales_rowwise.dim() >= 2 ? scales_rowwise.stride(0) : 0;
+  int64_t scale_rowwise_stride_dim1 = scales_rowwise.dim() >= 2 ? scales_rowwise.stride(1) : 0;
+  int64_t scale_colwise_stride_dim0 = scales_colwise.dim() >= 2 ? scales_colwise.stride(0) : 0;
+  int64_t scale_colwise_stride_dim1 = scales_colwise.dim() >= 2 ? scales_colwise.stride(1) : 0;
```
andrewor14, line 98: Hey @danielvegamyhre I had to add the dim check to pass tests for some reason. Does this look right? Here's what claude said:

```
  The Problem:                                                                                                                                                                                                                
                                                                                                                                                                                                                              
  In your commit, you changed from .strides()[i] to .stride(i) to use the stable ABI. However, the stable ABI's .stride(i) method validates the dimension index, while .strides()[i] accessed the array directly without      
  bounds checking.                                                                                                                                                                                                            
                                                                                                                                                                                                                              
  When rowwise=False, scales_rowwise is created as a 1D tensor with shape {0}. When the code calls scales_rowwise.stride(1), it fails because dimension index 1 is out of range for a 1D tensor (valid indices are only -1 and
   0).                                                                                                                                                                                                                        
                                                                                                                                                                                                                              
  The Fix (in torchao/csrc/cuda/mx_kernels/mxfp8_cuda.cu):                                                                                                                                                                    
                                                                                                                                                                                                                              
  Guard the stride accesses with dimension checks:                                                                                                                                                                            
                                                                                                                                                                                                                              
  // Get strides of scale ptrs (guard against 1D empty tensors when rowwise/colwise is false)                                                                                                                                 
  int64_t scale_rowwise_stride_dim0 = scales_rowwise.dim() >= 2 ? scales_rowwise.stride(0) : 0;                                                                                                                               
  int64_t scale_rowwise_stride_dim1 = scales_rowwise.dim() >= 2 ? scales_rowwise.stride(1) : 0;                                                                                                                               
  int64_t scale_colwise_stride_dim0 = scales_colwise.dim() >= 2 ? scales_colwise.stride(0) : 0;                                                                                                                               
  int64_t scale_colwise_stride_dim1 = scales_colwise.dim() >= 2 ? scales_colwise.stride(1) : 0;                                                                                                                               
                                                                                                                                                                                                                              
  This was actually a latent bug in the original code - the unstable ABI's .strides()[i] silently returned garbage/undefined values for out-of-bounds access. The stable ABI's stricter bounds checking exposed it. 
```

setup.py
```diff
@@ -755,11 +755,16 @@ def get_extensions():
...
+                            # define TORCH_TARGET_VERSION with min version 2.11 for ABI stable Float8_e8m0fnu
+                            "-DTORCH_TARGET_VERSION=0x020b000000000000",
```
andrewor14, line 784: Note: updated the target version to 2.11 cause we need https://github.com/pytorch/pytorch/pull/173669

jerryzh168: should we add a guard for torch_version_at_least("2.11.0") as well?

torchao/csrc/cuda/mx_kernels/mxfp8_cuda.cu
danielvegamyhre: Looks good to me

PR #3611, opened 2026-01-09
[mxfp8 moe training] export wgrad_with_hp recipe to quantize_ api

danielvegamyhre:
Stacked PRs:
 * #3651
 * __->__#3611

--- --- ---

[mxfp8 moe training] expose wgrad_with_hp recipe to quantize_ api

### DIFF
--- a/torchao/prototype/moe_training/conversion_utils.py
+++ b/torchao/prototype/moe_training/conversion_utils.py
@@ -20,6 +20,7 @@ logger: logging.Logger = logging.getLogger(__name__)
 class MoEScalingType(Enum):
     FP8_ROWWISE = "fp8_rowwise"
     MXFP8 = "mxfp8"
+    MXFP8_WGRAD_WITH_HP = "mxfp8_wgrad_with_hp"
 
 
 class MoETrainingConfig(AOBaseConfig):
--- a/benchmarks/prototype/moe_training/benchmark_scaled_grouped_mm_dq.py
+++ b/benchmarks/prototype/moe_training/benchmark_scaled_grouped_mm_dq.py
@@ -85,7 +85,11 @@ def get_configs() -> List[ExperimentConfig]:
         (128000, 2048, 7168, 4),
         (128000, 2048, 7168, 8),
     ]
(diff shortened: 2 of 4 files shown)


torchao/prototype/moe_training/conversion_utils.py
```diff
@@ -20,6 +20,7 @@
 class MoEScalingType(Enum):
     FP8_ROWWISE = "fp8_rowwise"
     MXFP8 = "mxfp8"
+    MXFP8_WGRAD_WITH_HP = "mxfp8_wgrad_with_hp"
```
vkuzo, line 23: i'd recommend having a separate thing for "grad_weight in hp" instead of coupling it to scaling type, for separation of concerns

danielvegamyhre: can you clarify what you mean? i am following the pattern from float8 training for consistency (`rowwise` vs `rowwise with gw hp`): https://github.com/pytorch/ao/blob/40cd03f4fe1992f07cdcfbd980bdc91d06d7673a/torchao/float8/config.py#L135

vkuzo: derived dtype (fp8 rowwise, mxfp8, etc) should be in a separate configuration variable vs how the 3 gemms are scaled (all quantized, grad_weight in hp, etc).  The code you linked in `float8/config.py` separates them.

danielvegamyhre:
linking to #3599

torchao/prototype/moe_training/conversion_utils.py
danielvegamyhre: i think the simplest solution is just to rename `MoEScalingType` -> `GroupedMMRecipeName` or something with "recipe name" in the name, which is what this config really is storing.  fp8 rowwise is a recipe, mxfp8 wgrad with hp is recipe, etc. 

i like the idea of knobs to control every cast, every gemm but until we support that i really want to avoid premature sprawl of additional configs when we only really support 2 recipes (fp8 rowwise to be deprecated), it makes the code very tedious to work with

danielvegamyhre: syncing offline and working on a follow up

PR #3614, opened 2026-01-10
Your diff, D90273266, broke some tests

jerryzh168:
Differential Revision: D90338397


jerryzh168:
fixed in https://github.com/pytorch/ao/pull/3621

PR #3615, opened 2026-01-10
Add aten.narrow support for inference_mode in Float8Tensor

jerryzh168:
## Summary
  - Extract common slicing logic into `_slice_impl` helper function
  - Add `aten.narrow.default` implementation for Float8Tensor
 

## Test plan
  - [x] `test_narrow_inference_mode` - verifies narrow works correctly with and without `torch.inference_mode()`
  - [x] All existing slice tests pass (18 tests)
  - [x] Full test suite passes (434 passed, 76 skipped)
  
  
  Created with claude


jerryzh168:
cc @manuelcandales seems mps tests are failing again

PR #3620, opened 2026-01-12
[ROCm] Refine from_recipe_name to support mxfp8 on rocm.

RuibinCheung:
## Summary
Support mxfp8 on gfx950 by refine `MXLinearConfig.from_recipe_name `

It will encounter error when called triton's quantize kernel with rceil on rocm because PTX instructions was called in kernel. So we chose implementation of `torch.compile` to workaround.

## TODO
* Implement quantize kernel with rceil mode in triton.

### DIFF
--- a/torchao/prototype/mx_formats/config.py
+++ b/torchao/prototype/mx_formats/config.py
@@ -166,6 +166,7 @@ class MXLinearConfig(AOBaseConfig):
         elif recipe_name is MXLinearRecipeName.MXFP8_CUBLAS:
             return MXLinearConfig(
                 kernel_preference=KernelPreference.AUTO,
+                # CUDA quantization kernels are not supported on ROCm. Fallback to triton.
                 mxfp8_cast_kernel_choice=MXFP8Dim1CastKernelChoice.TRITON
                 if is_ROCM()
                 else MXFP8Dim1CastKernelChoice.CUDA,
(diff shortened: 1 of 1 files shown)


torchao/prototype/mx_formats/config.py
```diff
@@ -165,12 +166,16 @@ def from_recipe_name(
         elif recipe_name is MXLinearRecipeName.MXFP8_CUBLAS:
             return MXLinearConfig(
                 kernel_preference=KernelPreference.AUTO,
-                mxfp8_cast_kernel_choice=MXFP8Dim1CastKernelChoice.CUDA,
+                mxfp8_cast_kernel_choice=MXFP8Dim1CastKernelChoice.TRITON
```
vkuzo, line 170: style nit: can we rewrite as below to improve readability

```python
# add a descriptive comment here
mxfp8_cast_kernel_choice = MXFP8Dim1CastKernelChoice.TRITON is if_ROCM() else MXFP8Dim1CastKernelChoice.CUDA

return MXLinearConfig(..., mxfp8_cast_kernel_choice)
```

vkuzo:
lgtm, could we update the style per my comment

torchao/prototype/mx_formats/config.py
RuibinCheung: ACK. Thanks.

RuibinCheung:
> lgtm, could we update the style per my comment

Thanks for your suggestion. But the ruff-format will modify the code if I follow your style. I can't find a method to disable ruff-format in a line.

Do you have any further suggestions?

RuibinCheung:
Hi @vkuzo, I saw my PR was blocked by Dr.CI. Could you help me trigger it ?

PR #3624, opened 2026-01-12
Add compatibility table entry for torchao 0.15.1

jerryzh168:
Summary:
att

Test Plan:
going to test later after branch cut for 0.15.1

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/__init__.py
+++ b/torchao/__init__.py
@@ -77,6 +77,9 @@ elif not ("+git" in __version__) and not ("unknown" in __version__):
         # Built against torch 2.9.1
         (_parse_version("0.15.0"), _parse_version("2.9.1")),
         (_parse_version("0.15.0"), _parse_version("2.10.0.dev")),
+        # Built against torch 2.9.1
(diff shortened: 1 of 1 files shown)


torchao/__init__.py
```diff
@@ -77,6 +77,9 @@ def _parse_version(version_string):
         # Built against torch 2.9.1
         (_parse_version("0.15.0"), _parse_version("2.9.1")),
         (_parse_version("0.15.0"), _parse_version("2.10.0.dev")),
+        # Built against 2.10
+        (_parse_version("0.15.1"), _parse_version("2.10.0")),
```
atalman: @jerryzh168 I believe we can however we would need to release it after torch release. Scheduled for 1.21.2026 . As per: https://dev-discuss.pytorch.org/t/pytorch-2-10-rc1-produced-for-pytorch-torchvision-torchaudio/3281

jerryzh168: OK updated to 2.9.1 so we can release earlier

PR #3633, opened 2026-01-14
Fix fake fusion for convolutions without bias (#3353)

jerryzh168:
Summary:

Fix `AttributeError` when performing fake fusion on convolution layers without bias by creating a zero-filled bias parameter instead of attempting to access requires_grad on None.

Reviewed By: jerryzh168

Differential Revision: D87356763


jerryzh168:
fixing ruff errors from https://github.com/pytorch/ao/pull/3353

cc @balakrishnanv

andrewor14 (approved):

PR #3634, opened 2026-01-14
[xpu][test] Remove None in get_current_accelerator_device

zxd1997066:
Per discussion in https://github.com/pytorch/ao/pull/3500#discussion_r2688437729, we try to remove the None branch and add an assert in ```get_current_accelerator_device```


liangan1:
@zxd1997066  pls rebase the code and fix the ci fails.

test/quantization/test_quant_primitives.py
```diff
@@ -600,6 +598,7 @@ def test_choose_qparams_tensor_asym_eps(self):
     )
     def test_get_group_qparams_symmetric_memory(self):
         """Check the memory usage of the op"""
+        _DEVICE = get_current_accelerator_device()
         weight = torch.randn(1024, 1024).to(device=_DEVICE)
         device_module = torch.get_device_module(_DEVICE)
```
liangan1:
```suggestion
        device = get_current_accelerator_device()
        weight = torch.randn(1024, 1024).to(device=device)
        device_module = torch.get_device_module(device)
```

liangan1: Since _DEVICE is not a global variable, I think it should follow the same name style as others. dit in other changed files of this PR.

test/integration/test_integration.py
```diff
@@ -730,6 +736,7 @@ def test_weight_only_groupwise_embedding_quant(self):
     @unittest.skipIf(not torch.accelerator.is_available(), "Need GPU available")
     def test_weight_only_quant_force_mixed_mm(self, device, dtype):
```
liangan1, line 738: I suppose we should use  @unittest.skipIf(device != get_current_accelerator_device(), "xxxx") instead of skipping it in the function.

test/integration/test_integration.py
```diff
@@ -86,9 +87,8 @@
 
 torch.manual_seed(0)
 config.cache_size_limit = 100
-_DEVICE = get_current_accelerator_device()
 
-COMMON_DEVICES = ["cpu", _DEVICE]
+COMMON_DEVICES = get_available_devices()
```
liangan1: This change will remove the cpu from the list?

test/dtypes/test_uintx.py
```diff
@@ -30,8 +30,7 @@
 )
 
 group_sizes = [32, 64, 128]
-_DEVICE = get_current_accelerator_device()
-devices = ["cpu", _DEVICE]
+devices = get_available_devices()
```
liangan1: This change will remove the cpu from the list? do it makes sense?

liangan1:
@zxd1997066  The common comments for all files in this PR:
1. The _DEVICE is a local variable with your change, I think you should follow the same name style as other variable.
2. cpu is removed in some cases, this may change the test scope of the original UTs. I suggest to keep the original semantics.

test/integration/test_integration.py
zxd1997066: no, see https://github.com/pytorch/ao/blob/main/torchao/utils.py#L144
```
def get_available_devices():
    devices = ["cpu"]
    if torch.cuda.is_available():
        devices.append("cuda")
    elif torch.xpu.is_available():
        devices.append("xpu")
    if torch.mps.is_available():
        devices.append("mps")
    return devices
```
it contains cpu

test/dtypes/test_uintx.py
zxd1997066: it contains cpu, see https://github.com/pytorch/ao/blob/main/torchao/utils.py#L144

liangan1: You are right. I mix this with get_current_accelerator_device

test/integration/test_integration.py
```diff
@@ -730,10 +736,6 @@ def test_weight_only_groupwise_embedding_quant(self):
     @unittest.skipIf(not torch.accelerator.is_available(), "Need GPU available")
     def test_weight_only_quant_force_mixed_mm(self, device, dtype):
         undo_recommended_configs()
-        if device != _DEVICE:
```
liangan1, line 733: We can not remove these device != _DEVICE: directly, how do ensure the device!=get_current_accelerator_device()? you may need to add @unittest context for this function.

test/integration/test_integration.py
```diff
@@ -924,11 +922,10 @@ class TestAutoQuant(unittest.TestCase):
             ],
         )
     )
+    @unittest.skipIf(not torch.accelerator.is_available(), "Need GPU available")
     def test_autoquant_one_input(self, device, dtype, m, k, n):
         undo_recommended_configs()
         print("(m, k, n): ", (m, k, n))
-        if device != _DEVICE or not torch.accelerator.is_available():
-            self.skipTest(f"autoquant currently does not support {device}")
```
liangan1, line 931: tido

test/integration/test_integration.py
jerryzh168: I'm also unclear on what is this doing and why is this removed, please add a comment

test/dtypes/test_affine_quantized.py
```diff
@@ -226,8 +230,9 @@ def test_test_copy__apply(self, apply_quant):
         "apply_quant", get_quantization_functions(False, True, _DEVICE, False)
```
jerryzh168, line 216: nit: _DEVICE is removed already?

test/integration/test_integration.py
zxd1997066: ```def test_weight_only_quant_force_mixed_mm(self, device, dtype):``` directly get the parameters from ```@parameterized.expand(COMMON_DEVICE_DTYPE)```, ```@unittest.skipIf``` can not get the variable ```device```,  will update to keep the skip in function.

test/integration/test_integration.py
zxd1997066: Yes, per CI results, it can not removed, will update and keep the skip.

test/dtypes/test_affine_quantized.py
zxd1997066: global variable ```_DEVICE``` is removed https://github.com/pytorch/ao/pull/3634/files#diff-31b1ffcd78674b79cc65749176354ea4743683070120034709c1da7a3eac31f6L48, local ```_DEVICE``` is defined in https://github.com/pytorch/ao/pull/3634/files#diff-31b1ffcd78674b79cc65749176354ea4743683070120034709c1da7a3eac31f6R91, as this function is guarded by ```@unittest.skipIf(not torch.accelerator.is_available(), "Need GPU available") ```, it will not be cpu

jerryzh168: OK, sg

test/integration/test_integration.py
zxd1997066: ```get_available_devices()``` can not be used here, as it will return ```['cpu', 'cuda']```, but currently ```COMMON_DEVICES``` is ```['cpu', device(type='cuda')]```, which makes difference in test (e.g. test_autoquant_one_input), will update.

Issue #3414, opened 2025-12-02
Delete special handling of per tensor quant in `_choose_scale_float8`


jerryzh168:
Recently I found that for `_choose_scale_float8`:

`len(block_size) == 0` for per tensor: https://github.com/pytorch/ao/blob/b4ec4cbab799b5320fc1ca094e0398b3ad967dbf/torchao/quantization/quant_primitives.py#L2303

This seems unnecessary, and I feel we should merge this branch with the others

KaranPradhan266:
@jerryzh168 I’d love to take this on, may I work on this issue?

jerryzh168:
@KaranPradhan266 yeah sure, please go ahead

PR #3635, opened 2026-01-14
Unify float8 scale selection for per-tensor quantization

MagellaX:
## Summary
- Remove the per-tensor special case in _choose_scale_float8 and use the unified reduction path
- Update float8 tests to use per-tensor block_size == input_tensor.shape

#3414

### DIFF
--- a/torchao/quantization/quant_primitives.py
+++ b/torchao/quantization/quant_primitives.py
@@ -148,34 +148,49 @@ _SUB_BYTE_INT_BOUNDS: Dict[Union[torch.dtype, TorchAODType], Tuple[int, int]] = 
     TorchAODType.INT7: (-(2**6), 2**6 - 1),
 }
 
-_SUB_BYTE_UINT_BOUNDS = {}
-
-
-def _maybe_register_dtype(mapping, dtypename: str, bounds: Tuple[int, int], bitwidth: int):
-    if hasattr(torch, dtypename):
-        dtype = getattr(torch, dtypename)
-        mapping[dtype] = bounds
-        _DTYPE_TO_BIT_WIDTH[dtype] = bitwidth
-
-
-for bits in range(1, 8):
-    _maybe_register_dtype(_SUB_BYTE_UINT_BOUNDS, f"uint{bits}", (0, 2**bits - 1), bits)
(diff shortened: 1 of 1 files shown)


torchao/quantization/quant_primitives.py
```diff
@@ -148,49 +148,34 @@ class TorchAODType(Enum):
...
-        torch.uint1: 1,
-        torch.uint2: 2,
-        torch.uint3: 3,
-        torch.uint4: 4,
-        torch.uint5: 5,
-        torch.uint6: 6,
-        torch.uint7: 7,
-    }
-)
+_SUB_BYTE_UINT_BOUNDS = {}
```
jerryzh168: could you split this change to a separate PR? seems it's not related to the fp8 choose_scale op change right?

MagellaX: Thanks! You’re right, that change is unrelated. I’m cleaning the branch now and will update the PR with only the float8 scale change.

jerryzh168: @MagellaX seems still not updated yet

torchao/quantization/quant_primitives.py
```diff
@@ -2275,32 +2260,25 @@ def _choose_scale_float8(
...
-        if hp_value_lb is not None or hp_value_ub is not None:
-            max_abs = torch.clamp(max_abs, min=hp_value_lb, max=hp_value_ub)
-        scale = max_abs / quant_max
-        # Reshape scale back to match the expected output shape
-        # The scale tensor should have the same shape as the input divided by block_size
-        output_shape = [
-            input_size // block_size[i] for i, input_size in enumerate(tensor.shape)
-        ]
-        scale = scale.reshape(output_shape)
+    shape_for_reduction, reduction_dims = _get_reduction_params(
```
jerryzh168, line 2283: these looks good to me I think, as long as all the tests are passing

torchao/quantization/quant_primitives.py
MagellaX: i think it is now!!!

jerryzh168:
this breaks BC for `_choose_scale_float8` for per tensor block_size but since `_choose_scale_float8` is a private op, we don't need to mark it as BC breaking

Issue #2146, opened 2025-04-28
AssertionError During Quantization of `torch.empty_like()`, `torch.ones_like`, and `torch.randn_like`


defaultd661:
### 🐛 Describe the bug

Similar to [#146621](https://github.com/pytorch/pytorch/issues/146621), when quantizing a model containing a `torch.empty_like()`, `torch.ones_like`, and `torch.randn_like` operations using `PT2E (prepare_pt2e)`, the process fails with an assertion error inside `_maybe_insert_input_observers_for_node`. The root cause is that `torch.empty_like()`, `torch.ones_like`, and `torch.randn_like` have `kwargs`, but currently the code assumes that most `aten` ops (except a few listed ones) should not have `kwargs`. 

### torch.empty_like
```
import torch
from torch.ao.quantization.quantize_pt2e import prepare_pt2e
from torch.ao.quantization.quantizer import Quantizer, QuantizationSpec
from torch.ao.quantization.observer import HistogramObserver
from torch.ao.quantization.quantizer.utils import _annotate_input_qspec_map, _annotate_output_qspec

class EmptyLikeModule(torch.nn.Module):

    def forward(self, t: torch.Tensor):
        return torch.empty_like(t)

class TestQuantizer(Quantizer):

    def annotate(self, model: torch.fx.GraphModule) ->torch.fx.GraphModule:
        qspec = QuantizationSpec(torch.int8, HistogramObserver, qscheme=
            torch.per_tensor_symmetric)
        for node in model.graph.nodes:
            for input_node in node.all_input_nodes:
                _annotate_input_qspec_map(node, input_node, qspec)
            _annotate_output_qspec(node, qspec)
        return model

    def validate(self, model: torch.fx.GraphModule) ->None:
        pass
    
def test_bug():
    model = EmptyLikeModule()
    exported_model = torch.export.export(model, (torch.randn(10),))
    prepared_model = prepare_pt2e(exported_model.graph_module,
        TestQuantizer())

if __name__ == '__main__':
    test_bug()
```

### torch.ones_like
```
import torch
from torch.ao.quantization.quantize_pt2e import prepare_pt2e
from torch.ao.quantization.quantizer import Quantizer, QuantizationSpec
from torch.ao.quantization.observer import HistogramObserver
from torch.ao.quantization.quantizer.utils import _annotate_input_qspec_map, _annotate_output_qspec

class OnesLikeModule(torch.nn.Module):

    def forward(self, t: torch.Tensor):
[...]

PR #3642, opened 2026-01-15
Allow non-tensor kwargs in prepare_pt2e

Erik-Lundell:
Add a list of permitted kwargs in
_maybe_insert_input_observers_for_node:
device, pin_memory, and memory_format.

Some ops are already permitted, showing that
the assert is not a hard limit. The kwargs
in question should not affect or be affected
by quantization, and there is no clear reason
for why all kwargs should be disallowed.

There has been at least two discussions on this issue https://github.com/pytorch/ao/issues/2146
https://github.com/pytorch/pytorch/issues/146621,

and it has caused a follow-up bug in Executorch:
https://github.com/pytorch/executorch/issues/16541

### DIFF
--- a/test/quantization/pt2e/test_quantize_pt2e.py
+++ b/test/quantization/pt2e/test_quantize_pt2e.py
@@ -1463,9 +1463,9 @@ class TestQuantizePT2E(PT2EQuantizationTestCase):
     def test_quantize_kwargs(self):
         """Ensure non-tensor kwargs pass quantization, tensor kwargs don't"""
 
-        class EmptyLikeModule(torch.nn.Module):
+        class OnesLikeModule(torch.nn.Module):
             def forward(self, t: torch.Tensor):
-                return torch.empty_like(t, device="cpu", pin_memory=False)
+                return torch.ones_like(t, device="cpu", pin_memory=False)
 
         class ClampModule(torch.nn.Module):
             def __init__(self, max: float | torch.Tensor, use_out: bool):
@@ -1521,8 +1521,8 @@ class TestQuantizePT2E(PT2EQuantizationTestCase):
         example_inputs = torch.randn(1, 2, 3, 3)
         quantizer = BackendAQuantizer()
 
-        # Example 1: empty_like with device and pin_memory passes.
-        m = self._quantize(EmptyLikeModule(), quantizer, (example_inputs,))
+        # Example 1: ones_like with device and pin_memory passes.
+        m = self._quantize(OnesLikeModule(), quantizer, (example_inputs,))
         validate_quantization(m)
         m(example_inputs)
 


test/quantization/pt2e/test_quantize_pt2e.py
```diff
@@ -1460,6 +1460,40 @@ def validate(self, model: torch.fx.GraphModule) -> None:
...
+
+        class BackendAQuantizer(Quantizer):
+            def annotate(self, model: torch.fx.GraphModule) -> torch.fx.GraphModule:
+                qspec = QuantizationSpec(
+                    torch.int8,
+                    observer.default_observer,
+                    qscheme=torch.per_tensor_symmetric,
+                )
+                for node in model.graph.nodes:
+                    input_qspec_map = {
```
jerryzh168, line 1492: this is trying to quantize all inputs for all nodes? maybe add a check to make sure the observers are inserted as well?

test/quantization/pt2e/test_quantize_pt2e.py
```diff
@@ -1460,6 +1460,40 @@ def validate(self, model: torch.fx.GraphModule) -> None:
...
+
+            def validate(self, model: torch.fx.GraphModule) -> None:
+                pass
+
+        m = EmptyLikeModule()
+        example_inputs = torch.randn(1, 2, 3, 3)
+        m = torch.export.export(m, (example_inputs,), strict=True).module()
+        m = prepare_pt2e(m, BackendAQuantizer())
+        m(example_inputs)
+        convert_pt2e(m)
```
jerryzh168, line 1546: nit: also run the model after convert to make sure

torchao/quantization/pt2e/prepare.py
```diff
@@ -544,12 +544,19 @@ def _maybe_insert_input_observers_for_node(
     # Clone has a memory_format kwarg, zeros_like has a pin_memory kwarg, and
     # gelu has a has an approximate kwarg that persist in exported graph.
     # This is just a work around for these.
-    assert (
-        node.target == torch.ops.aten.clone.default
-        or node.target == torch.ops.aten.zeros_like.default
-        or node.target == torch.ops.aten.gelu.default
-        or len(node.kwargs) == 0
-    ), " expecting kwargs for aten op IR to be empty"
+    # We also permit some general kwargs that shouldn't be affected by quantization.
```
jerryzh168: also it seems to me this is just an assert for aten IR, seems unnecessary to put here in the pt2e flow, would it work if just remove this?

also you mentioned `torch.add(x,y, alpha=0.8).` in https://github.com/pytorch/ao/issues/2146#issuecomment-3376286612, does that work with pt2e quant right now?

jerryzh168: oh I guess there is still some value here since if we remove these, kwargs will just be ignored, it will be silent failure if there are kwargs that are tensor. I feel in the end, we want to detect if there is tensor args in kwargs dict and error out based on that

JacobSzwejbka: @Erik-Lundell Can you just loop over the leaves of kwargs and assert that none are tensors. We are separately seeing this same problem for the dtype arg

Erik-Lundell: Sure, I'd prefer that. I went with a defensive approach since I didn't know the reason for the assert. If you both believe that the guard is only necessary for tensors, let's go for that.

Erik-Lundell: Not sure about the torch.add(x,y, alpha=...) situation though. I guess it is up to the backend to handle it. In the Arm backend we solve it by decomposing in transform_for_annotation for example.

test/quantization/pt2e/test_quantize_pt2e.py
Erik-Lundell: Added a check that we have q/dq visitors afterwards.

torchao/quantization/pt2e/prepare.py
Erik-Lundell: It was actually very hard to find a case with tensor kwargs, let me know what you think.

jerryzh168: yeah I think if you just change the assert for tensor kwargs, the add use case should be supported as well.

also for tensor kwargs, we could support it as well if needed, but seems no need to do that now

torchao/quantization/pt2e/prepare.py
```diff
@@ -541,15 +541,13 @@ def _maybe_insert_input_observers_for_node(
...
-    # gelu has a has an approximate kwarg that persist in exported graph.
-    # This is just a work around for these.
-    assert (
-        node.target == torch.ops.aten.clone.default
-        or node.target == torch.ops.aten.zeros_like.default
-        or node.target == torch.ops.aten.gelu.default
-        or len(node.kwargs) == 0
-    ), " expecting kwargs for aten op IR to be empty"
+    # Check that no kwargs are Tensors. Otherwise they silently go unquantized.
+    disallowed_kwarg_types = (torch.fx.Node,)
```
jerryzh168: probably should look at all args recursively, e.g. dict of list of dict, that contains a torch.fx.Node anywhere

Erik-Lundell: Done. Can you approve workflow to see that tests pass?

torchao/quantization/pt2e/prepare.py
```diff
@@ -541,15 +542,19 @@ def _maybe_insert_input_observers_for_node(
...
-        or len(node.kwargs) == 0
-    ), " expecting kwargs for aten op IR to be empty"
+    # Check that no kwargs are Tensors. Otherwise they silently go unquantized.
+    def _validate_kwarg(name: str, val):
+        disallowed_kwarg_types = (torch.fx.Node,)
+        if isinstance(val, disallowed_kwarg_types):
+            raise AssertionError(
+                f"Kwarg '{name}' contained type {type(val)} which is not supported for annotation."
+            )
+        if isinstance(val, Sequence):
```
jerryzh168: seems this does not include the case when `val` is a dict?

Erik-Lundell: No, added it and made the check for Sequence more precise, it caused an error with infinite recursion

jerryzh168:
please fix ruff: https://github.com/pytorch/ao/actions/runs/21207774639/job/61049126964?pr=3642

Erik-Lundell:
> please fix ruff: https://github.com/pytorch/ao/actions/runs/21207774639/job/61049126964?pr=3642

That's an annoying thing to miss, fixed it. Also addressed an issue causing flakiness. Slowly getting there :)

PR #3643, opened 2026-01-15
[ROCM] Fix float8 rounding test case to enable ROCM

alex-minooka:
### Summary

Modified test case to accept ROCM specific value

ROCm's torch.log2 implementation has higher numerical precision than NVIDIA's, causing the largest normal fp32 number test case to floor to 2^127 instead of overflowing to inf. Updated test to accept platform-specific expected values and removed skip_if_rocm decorator to enable the test on ROCm.

### DIFF
--- a/test/float8/test_float8_utils.py
+++ b/test/float8/test_float8_utils.py
@@ -9,7 +9,7 @@ import pytest
 import torch
 
 from torchao.float8.float8_utils import _round_scale_down_to_power_of_2
-from torchao.utils import is_ROCM
+from torchao.testing.utils import skip_if_rocm
 
 
 # source for notable single-precision cases:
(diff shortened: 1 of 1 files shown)


test/float8/test_float8_utils.py
```diff
@@ -23,7 +24,7 @@
         ("inf", float("inf"), float("inf")),
         ("nan", float("nan"), float("nan")),
         ("smallest positive subnormal number", 2**-126 * 2**-23, 2**-126 * 2**-23),
-        ("largest normal number", 2**127 * (2 - 2**-23), float("inf")),
+        ("largest normal number", 2**127 * (2 - 2**-23), 2**127 if is_ROCM() else float("inf")),
```
vkuzo: add a comment with context from the PR summary? great to have this kind of info inline in the code

alex-minooka: Added inline comment

vkuzo: nit: move the comment above the code instead of below

alex-minooka: Done. Moved it above.

test/float8/test_float8_utils.py
```diff
@@ -32,7 +35,7 @@
         # ("largest subnormal number", [2**-126 * (1 - 2**-23), 1.1754943508222875e-38]),
     ],
 )
-@skip_if_rocm("ROCm enablement in progress")
+# @skip_if_rocm("ROCm enablement in progress")
```
vkuzo: nit: either revert this change or remove the comment?

alex-minooka: removed comment

alex-minooka:
Going to reopen pull request, this was on main branch of fork and was interfering with other prs.

PR #3651, opened 2026-01-16
[mxfp8 moe training] rename MoEScalingType -> ScaledGroupedMMRecipe

danielvegamyhre:
[mxfp8 moe training] rename MoEScalingType -> ScaledGroupedMMRecipe

## Summary
- `MoEScalingType` has ended up being not a scaling type, but rather a recipe name. The name should reflect that.
- Also, `ScaledGroupedMMRecipe` aligns better with the `ScaledGroupedMMTensor` which dispatches with this recipe as an arg.

Note: I think we should also probably rename `MoETrainingConfig` to align with this naming pattern. 

## Tests
- `pytest test/prototype/moe_training/test_training.py`
- `pytest test/prototype/moe_training/test_scaled_grouped_mm.py`


danielvegamyhre:
@vkuzo let me know what you think of this as an intermediate solution. i will rename the enum in torchtitan as well if we proceed.

PR #3652, opened 2026-01-17
Update TorchAOBaseTensor docs

jerryzh168:
Summary:
formatting + removing some deprecated helper functions in doc

Test Plan:
check docs page for the PR

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/utils.py
+++ b/torchao/utils.py
@@ -397,7 +397,19 @@ and dispatching to these implementations.
 
 
 def _implements(cls, aten_ops):
-    """Decorator to implement aten ops for __torch_dispatch__."""
+    """Decorator for implementing aten ops like `torch.ops.aten.linear.default` for
+    tensor subclass, the implemented functions are called in ``__torch_dispatch__`` callback
+    for ``torch.Tensor`` subclasses
+
+    Examples::
+
+        implements = MyTensor.implements
+
+        @implements(torch.ops.aten.linear.default):
+        def _(func, types, args, kwargs):
+            ...
+
+    """
     if not hasattr(cls, "_ATEN_OP_TABLE"):
         cls._ATEN_OP_TABLE = {}
     if cls not in cls._ATEN_OP_TABLE:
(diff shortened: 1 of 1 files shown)


torchao/utils.py
```diff
@@ -730,7 +730,7 @@ def _get_to_kwargs(self, *args, **kwargs):
 
 
 class TorchAOBaseTensor(torch.Tensor):
-    """A util tensor subclass that provides commonly used functions
+    r"""A util tensor subclass that provides commonly used functions
```
danielvegamyhre, line 779: why add this r"" prefix here?

jerryzh168: seems current format is messed up: https://docs.pytorch.org/ao/main/generated/torchao.utils.TorchAOBaseTensor.html#torchao.utils.TorchAOBaseTensor, e.g. `abc` is treated as highlight, so was just trying to see if r""" would help

jerryzh168: actually I just checked https://github.com/pytorch/pytorch/blob/v2.9.1/torch/nn/modules/conv.py#L374, r""" probably won't help, I'll follow that instead

jerryzh168:
doc build is only enabled when docs/ files changed: https://github.com/pytorch/ao/commit/52280bbb69e29ccde28b529157e313f849bd9ff0
manually triggering doc build in https://github.com/pytorch/ao/actions/workflows/doc_build.yml

torchao/utils.py
```diff
@@ -397,7 +397,18 @@ def torch_version_at_least(min_version):
...
+
+    Examples:
+
+        implements = MyTensor.implements
+
+        @implements(torch.ops.aten.linear.default):
+        def _(func, types, args, kwargs):
+            ...
+
+    """
```
svekars: I think for this example code block to render correctly (right now, it renders like [this](https://docs-preview.pytorch.org/pytorch/ao//generated/torchao.utils.TorchAOBaseTensor.html#torchao.utils.TorchAOBaseTensor.implements)), you need to either add double colon `::` after Examples, or use Google-style/doctest docstrings like:

```
 """Decorator for implementing aten ops like ``torch.ops.aten.linear.default`` for
    tensor subclass.

    Examples:
        >>> implements = MyTensor.implements
        >>> @implements(torch.ops.aten.linear.default)
        ... def _(func, types, args, kwargs):
        ...     pass
    """
```

torchao/utils.py
```diff
@@ -995,6 +1019,7 @@ def __repr__(self):
         )
 
     def get_layout(self):
+        """DEPRECATED"""
```
svekars: Use a Sphinx directive `.. deprecated::`, like:
```
.. deprecated:: <version>
        Use :meth:`new_method_name` instead, or provide explanation why it's deprecated.
```

jerryzh168:
docs: https://docs-preview.pytorch.org/pytorch/ao//generated/torchao.utils.TorchAOBaseTensor.html

PR #3617, opened 2026-01-10
Deprecate `TransformerEvalWrapper`, `LMEvalInputRecorder`

namgyu-youn:
**Summary**:
Follow-up #3720, AWQ have been used `TransformerEvalWrapper`, which wrap an ao model to run `lm-eval`. Instead, update to run `lm-eval` directly by using `HFLM`. 

**Change log**:
```py
# Old
from torchao._models._eval import TransformerEvalWrapper

TransformerEvalWrapper(
    model=model,
    tokenizer=tokenizer,
    max_seq_length=max_seq_length,
).run_eval(
    tasks=tasks,
    limit=calibration_limit,
)
```

```py
# New (direct lm-eval)
from lm_eval import evaluator
from lm_eval.models.huggingface import HFLM

evaluator.simple_evaluate(
    HFLM(pretrained=model, tokenizer=tokenizer),
    tasks=tasks,
    limit=calibration_limit,
    batch_size=1,
)
```

**Test plan**:
```
python quantize_and_upload.py --model_id meta-llama/Llama-3.2-1B --quant AWQ-INT4
```

PR #3655, opened 2026-01-17
Add static quant flow support for `Float8StaticActivationFloat8WeightConfig`

jerryzh168:
* __->__ #3655

Summary:

Add AWQ-style observer flow support to Float8StaticActivationFloat8WeightConfig for static float8 quantization.

- Added Float8StaticStep enum with PREPARE and CONVERT values following the AWQ API pattern
- Added Float8ObservedLinear class that wraps a linear layer with an AffineQuantizedMinMaxObserver to collect activation statistics during calibration
- Updated Float8StaticActivationFloat8WeightConfig with a new step parameter supporting two workflows:
  a. AWQ-style observer flow: Use step="prepare" to insert observers, calibrate with representative data, then step="convert" to convert to quantized model
  b. Direct quantization: Provide act_quant_scale directly (existing behavior preserved)
- Handler transforms:
  - step="prepare": Replaces nn.Linear with Float8ObservedLinear containing an observer
  - step="convert": Extracts scale from observer and creates quantized weights with PrototypeFloat8Tensor

with claude-code

Test Plan:

Run the new observer flow test:
python -m pytest test/prototype/test_prototype_float8_tensor.py::TestFloat8StaticActivation::test_static_quant_flow_with_observers -v

Run all static quantization tests:
python -m pytest test/prototype/test_prototype_float8_tensor.py -v

Reviewers:

Subscribers:

Tasks:

Tags:


test/prototype/test_prototype_float8_tensor.py
```diff
@@ -52,6 +53,27 @@ def forward(self, x):
         return self.conv(x)
 
 
+class ToyLinearModel(torch.nn.Module):
```
namgyu-youn:
```suggestion
from torchao.testing.model_architectures import ToyTwoLinearModel
```

torchao/prototype/quantization/quant_api.py
```diff
@@ -323,66 +360,189 @@ def __post_init__(self):
...
-    activation_dtype = config.activation_dtype
-    weight_dtype = config.activation_dtype
-
-    granularity = config.granularity
-    if granularity is None:
-        granularity = PerRow()
-    elif isinstance(granularity, tuple):
-        raise ValueError(
-            "Tuple granularity is not supported. Please specify a single granularity value."
+    from torchao.quantization.observer import AffineQuantizedMinMaxObserver
```
namgyu-youn: So in the future, we can use custom observers like AWQ/SmoothQuant maybe?

torchao/prototype/quantization/quant_api.py
```diff
@@ -323,66 +360,189 @@ def __post_init__(self):
...
+
+    step = config.step
+    granularity = config.granularity if config.granularity is not None else PerTensor()
+
+    if step == Float8StaticStep.PREPARE or step == "prepare":
+        # Create observer and wrap linear
+        observer = AffineQuantizedMinMaxObserver(
```
namgyu-youn, line 460: So in the future, we can expand to more observers like `AWQObserver/SmoothQuantObserver`? If so, we might need something like
```python
class StaticObserver(Enum):
    MINMAX = "AffineQuantizedMinMaxObserver"
    AWQ = "AWQObserver"
    SMOOTHQUANT = "SmoothQuantObserver"
```

test/prototype/test_prototype_float8_tensor.py
```diff
@@ -244,6 +266,70 @@ def test_fp8_conv_variants(
...
+
+        This follows the AWQ pattern:
+        1. Prepare model by inserting observers (step="prepare")
+        2. Calibrate with representative data
+        3. Convert observed model to quantized model (step="convert")
+        """
+        torch.compiler.reset()
+        torch.manual_seed(42)
+
+        dtype = torch.bfloat16
```
namgyu-youn:
```suggestion
        dtype = self.dtype
```

namgyu-youn (commented):
Left a few comment, thanks for this reference code!

torchao/prototype/quantization/quant_api.py
namgyu-youn: Maybe we can define `ObservedLinear` as a base config, and build enum backend inside https://github.com/pytorch/ao/blob/main/torchao/quantization/observer.py

torchao/prototype/quantization/quant_api.py
```diff
@@ -323,66 +360,189 @@ def __post_init__(self):
...
+                raise ValueError(f"{self.step} is not one of {all_step_values}")
+
+
+# Float8 static quantization step enum (similar to AWQStep)
+class Float8StaticStep(str, Enum):
+    PREPARE = "prepare"
+    CONVERT = "convert"
+
+
+class Float8ObservedLinear(torch.nn.Linear):
```
jcaip, line 391: Do you have thoughts on creating an observer tensor subclass so we're not limited to `nn.Linear` modules? Thinking about how to support F.linear / torch.bmm. Would the original flow still work?

torchao/prototype/quantization/quant_api.py
```diff
@@ -323,66 +360,189 @@ def __post_init__(self):
...
+            act_quant_scale = act_quant_scale.view(1, 1)
+
+        if config.set_inductor_config:
+            torchao.quantization.utils.recommended_inductor_config_setter()
+
+        activation_dtype = config.activation_dtype
+        weight_dtype = config.weight_dtype
+
+        # Create quantized weight tensor
+        quantized_tensor = PrototypeFloat8Tensor.from_hp(
```
jcaip, line 492: Do we still want to use the prototype tensor for this?

torchao/prototype/quantization/quant_api.py
```diff
@@ -323,66 +360,189 @@ def __post_init__(self):
...
+            bias=module.bias is not None,
+            device=module.weight.device,
+            dtype=module.weight.dtype,
+        )
+        linear.weight = torch.nn.Parameter(quantized_tensor, requires_grad=False)
+        linear.bias = module.bias
+        return linear
+
+    elif step is None:
+        # Direct quantization path - use provided act_quant_scale
```
jcaip, line 520: Thoughts on if it makes sense to separate this into two configs `AWQConfig`, and `Float8StaticActivationFloat8WeightConfig`, like we do in #3517?

Idea being you would call `AWQConfig(step="prepare"), AWQConfig(step="convert")` and `Float8StaticActivationFloat8WeightConfig` must take in a scale.

torchao/prototype/quantization/quant_api.py
jerryzh168: for different observers, I think we can worry about that later, I'd imagine it would be more of moving average, histogram etc.

the API design is following AWQ/smoothquant, but we don't need to combine these together I think

torchao/prototype/quantization/quant_api.py
jerryzh168: we can use that when it's needed I feel, I feel the module swap might be simpler to start with, and we may not necessarily have use cases of F.linear/torch.bmm for float8 static quant, so probably better to start with this for now

torchao/prototype/quantization/quant_api.py
jerryzh168: oh, will merge after this

torchao/prototype/quantization/quant_api.py
jerryzh168: you mean 
```
StaticQuantConfig(Float8StaticActivationFloat8WeightConfig(), step=“prepare”)
StaticQuantConfig(Float8StaticActivationFloat8WeightConfig(), step=“convert”)
```

?

yeah I thought about this as well, seems there are some duplicate here talking about static quant twice, also we'd need an API / callback in the `Float8StaticActivationFloat8WeightConfig` to make things configurable, like `Float8ObservedLinear`, `AffineQuantizedMinMaxObserver`

something like:

```
class Float8StaticActivationFloat8WeightConfig:
   ...
   static_quant_observed_linear = Float8ObservedLinear
   static_quant_observer = AffineQuantizedMinMaxObserver
```

just adding new abstraction, we can check if it is worth unifying when we have more static quant flows I think

torchao/prototype/quantization/quant_api.py
```diff
@@ -323,66 +360,189 @@ def __post_init__(self):
...
+                self.step = self.step.lower()
+            all_step_values = [s.value for s in Float8StaticStep]
+            if self.step not in all_step_values and self.step not in list(
+                Float8StaticStep
+            ):
+                raise ValueError(f"{self.step} is not one of {all_step_values}")
+
+
+# Float8 static quantization step enum (similar to AWQStep)
+class Float8StaticStep(str, Enum):
```
namgyu-youn, line 386: btw, I think we can use `ObserverStep` (common enum for observer steps) for future extensibility. We can update AWQ/SmoothQuant to reuse common flow.

https://github.com/pytorch/ao/blob/3350b2fd9ccae1e3868045ca90514f5d1ae95a84/torchao/prototype/awq/core.py#L21-L24

and

https://github.com/pytorch/ao/blob/3350b2fd9ccae1e3868045ca90514f5d1ae95a84/torchao/prototype/smoothquant/core.py#L13-L16

But not in this PR. Let me know if this makes sense.

torchao/prototype/quantization/quant_api.py
```diff
@@ -298,16 +300,51 @@ class Float8StaticActivationFloat8WeightConfig(AOBaseConfig):
...
+            PREPARE: insert observers to linear
+            CONVERT: convert the observed linear modules to linear modules with quantized weights
+            Can use the corresponding string "prepare", "convert" for simplicity
+            If not provided, act_quant_scale must be provided for direct quantization.
+        act_quant_scale (Optional[torch.Tensor]): The scale tensor for activation quantization.
+            Required when step is not provided.
         activation_dtype (torch.dtype): The target data type for activation quantization. Default is torch.float8_e4m3fn
         weight_dtype (torch.dtype): The target data type for weight quantization. Default is torch.float8_e4m3fn
-        granularity (FP8Granularity): The granularity of quantization. PerRow() and PerTensor() are supported currently.
+        granularity (FP8Granularity): The granularity of quantization. Only PerTensor() is supported for static activation quantization because the scale must be fixed at calibration time and work for any batch size at inference.
```
namgyu-youn: nit: `PerTensor` directly, because it is only supported granularity at now?

torchao/prototype/quantization/quant_api.py
jerryzh168: sounds OK, let's see if there are more similar cases coming up before we unify

torchao/prototype/quantization/quant_api.py
jerryzh168: I think probably still better to have this even that's the only option for now, since we don't talk about granularity in the name

jerryzh168: oh or you mean the comment? I have updated it to PerTensor() only

namgyu-youn: Oh actually I was wondering if we can drop `FP8Granularity` (just use union or something), because didn't understand why dtype-specific granularity is needed.

namgyu-youn: But out of this topic I feel, we can just use `FP8Granularity` preserving existing behavior.

torchao/prototype/quantization/quant_api.py
jcaip: I would be interested in using this for GPTQ as well

torchao/prototype/quantization/quant_api.py
jcaip: yeah agree this is good for now, was just curious.

torchao/prototype/quantization/quant_api.py
jcaip: sorry think I was a bit confused myself here actually, I thought that 
```
    Example (Observer flow):
        # Step 1: Prepare model by inserting observers
        quantize_(model, Float8StaticActivationFloat8WeightConfig(step="prepare"))

        # Step 2: Calibrate with representative data
        for batch in calibration_data:
            model(batch)

        # Step 3: Convert observed model to quantized model
        quantize_(model, Float8StaticActivationFloat8WeightConfig(step="convert"))
```
would be how you do AWQ quantization, I realize now this is just for generic rtn quant.

I'm less worried about the static quant API unification, more so it seems like there's some repeated observe - prepare - convert logic thats shared across smoothquant, this, gptq and awq, that might be worth consolidating cc @namgyu-youn.

torchao/prototype/quantization/quant_api.py
```diff
@@ -323,66 +360,189 @@ def __post_init__(self):
...
 
+        # Allow string step values for convenience
+        if self.step is not None:
+            if isinstance(self.step, str):
+                self.step = self.step.lower()
+            all_step_values = [s.value for s in Float8StaticStep]
+            if self.step not in all_step_values and self.step not in list(
+                Float8StaticStep
+            ):
+                raise ValueError(f"{self.step} is not one of {all_step_values}")
```
jcaip, line 369: fine to add in later PR, but we should add an implementation of `https://github.com/pytorch/ao/blob/c6bc74c5689328882128fb8645a98073e52604c0/torchao/quantization/quantize_/common/protocol.py#L39` here so it implements the protocol.

torchao/prototype/quantization/quant_api.py
```diff
@@ -298,16 +300,51 @@ class Float8StaticActivationFloat8WeightConfig(AOBaseConfig):
     """
     Configuration for applying float8 static symmetric quantization to both activation and weight.
 
+    This supports two workflows:
+
+    1. AWQ-style observer flow (recommended):
```
jcaip: nit: maybe just Observer based flow here?

torchao/prototype/quantization/quant_api.py
jerryzh168: that's right, yeah we could unify separately

torchao/prototype/quantization/quant_api.py
jerryzh168: oh, I think we should just use Granularity if we want to improve this, I can change to that

PerTensor is more intended to be used as a value, so probably should not use it as a type to reduce confusions

torchao/prototype/quantization/quant_api.py
jerryzh168: OK, can do now since your PR is landed

torchao/prototype/quantization/quant_api.py
jerryzh168: OK, seems like we have one for attention, will probably need to explore this a bit

torchao/prototype/quantization/quant_api.py
namgyu-youn: @jcaip opened https://github.com/pytorch/ao/pull/3678 for common workflow, could you look into it?

torchao/prototype/quantization/quant_api.py
namgyu-youn: @jcaip to me, AWQ/SmoothQuant/GPTQ works like "apply something before RTN". For example, in SmoothQuant, "something" here can be activation pre-scaling. After applying it, it works like RTN, as you mentioned.

We can decompose SmoothQuant into
(1) prepare: Insert observer to linear layers (before applying algorithm)
(2) observe: Run algorithm (i.e., compute activation scaling factor)
(3) convert: Apply "something" to linear module before quantization (i.e., apply activation pre-scale)
(4) quantize: RTN

Also I think we don't have to consolidate (1)-(3), because we might use `lm-eval` (calibration-flow) between them. We can merge to follow other subclass design, but it might be too heavy.

namgyu-youn (approved):
Looks good to me, will move into integer flow after this land.

torchao/prototype/quantization/quant_api.py
namgyu-youn: @jcaip to me, AWQ/SmoothQuant works like **apply something before RTN**. For example, in SmoothQuant, "something" here can be "activation pre-scaling". After applying it, it works like RTN, as you mentioned.

We can decompose SmoothQuant into
(1) prepare (Step 1): Insert observer to linear layers (before applying algorithm)
(2) observe (Step 2): Run algorithm based on calibration flow (i.e., compute activation scaling factor with `lm-eval`)
(3) convert (Step 3.1): Apply (2) value to high-precision model (i.e., apply activation pre-scale)
(4) quantize (Step 3.2): RTN

Let me know if my explanation is incorrect. Also, I think we don't have to consolidate (1)-(3), because we might use `lm-eval` (calibration-flow) between them. We can merge to follow other subclass patterns, but it might be too heavy.

jcaip: For GPTQ, step 3 and 4 and kind of interleaved, but overall agree with your breakdown. I think I actually agree with your point on not consoldiating the code logic here, but mostly I want to make sure user APIs / experience for AWQ/SmoothQuant/GPTQ/observer-rtn are generally consistent. 

For example it seems like we have duplicated example scripts for AWQ/GPTQ/Smoothquant, it might be a good idea to consolidate this so we can report benchmarks across the three techniques vs just solo. I think there's a quite a bit of low hanging fruit here.

vkuzo:
> it seems like there's some repeated observe - prepare - convert logic thats shared across smoothquant, this, gptq and awq, that might be worth consolidating cc @namgyu-youn.

I wouldn't overindex on this. IMO it's better to ensure the workflows have good accuracy and performance individually on a good set of models + hardware types first, and worry about this later.  If logic is duplicated for this but that leads to simpler overall code, that is okay.

torchao/prototype/quantization/quant_api.py
namgyu-youn: @jcaip yeah I confused GPTQ; it's not the same as the 1-4 breakdown. 😅

For AWQ/SmoothQuant/GPTQ, I think integration/consolidation can be done later due to a few issues:
(1) Only a few subclasses have experimental support (I remember only W4A16-INT (tensor core layout) supports AWQ)
(2) vLLM serve currently fails: https://gist.github.com/namgyu-youn/dff3e22320b028b28f7d533727a88bb1

Also I have a few related PRs that would be great if you could review:
(3) Minor cleanup like #3617 is in progress
(4) Initial benchmark is in progress at https://github.com/pytorch/ao/pull/3602 (mostly there I believe)

Wondering if you'll continue working on AWQ/SmoothQuant/GPTQ. I saw you're planning comprehensive experiments based on https://arxiv.org/pdf/2409.11055v1, so I'd like to hear your future plans before consolidation. Personally planning doc update (https://github.com/pytorch/ao/blob/main/docs/source/static_quantization.rst) after #3687 land. What about your thoughts?

Issue #3636, opened 2026-01-14
mx_block_rearrange_2d_M_groups_cuda does not support scales_tensor with M-group size = 44


eee4017:
### Description

We are hitting a runtime failure when using the newly added CUDA kernel `mx_block_rearrange_2d_M_groups_cuda` introduced in PR #3546.

Our workload requires a specific per-group size (44), which appears to be unsupported by the kernel. The failure manifests as a CUDA driver error at runtime.

**Input configuration:**

```python
scales_tensor: torch.Size([98048, 44])
dtype = torch.float8_e8m0fnu

input_group_end_offsets: torch.Size([8])
dtype = torch.int32

chunk_width = 64
chunks_per_tb = 8
```

**Error:**

```text
CUDA workload error: CUDA Driver Error at /workspace/ao/torchao/csrc/cuda/mx_kernels/mx_block_rearrange_2d_M_groups.cu:431
invalid argument
```

cc @danielvegamyhre @drisspg

danielvegamyhre:
at first glance this looks like a validation bug but will investigate. 

in practice you'll need to pad group sizes to be multiple of 32, the scaling block size for mxfp8. 
this is because in the backward pass, for `wgrad = grad_out^T @ input_act` the groups are along the contracting dim, where the 1x32 (LHS) and 32x1 (RHS) scaling blocks are applied, and we can't have a scaling block cross group boundaries.

for example, in TorchTitan, we do this [here](https://github.com/pytorch/torchtitan/blob/9f211ec199bc887901b874edd6af5a20527a4175/torchtitan/models/moe/utils.py#L52), we fuse this padding to the kernel which does "permutation."

A better name would be `gather_and_fill` or similar, really. what it does is generate indexes to rearrange/gather tokens from layout:

`[rank0 tokens for expert 0, rank0 tokens for expert 1, rank1 tokens for expert0, rank1 tokens for expert1, ...]` 

to layout:

`[tokens for expert 0 from all ranks, tokens for expert 1 from all ranks, ...]` - INCLUDING adding padding tokens as necessary to ensure group sizes are divisible by 32.

If you want, we have our own version of this operation in torchao you can use: https://github.com/pytorch/ao/blob/c6d21f9b82819a7e3546a267eafb15a5b8cf3951/torchao/prototype/moe_training/ep/permute.py#L167


i will figure out why this kernel is not handling this properly, though.

danielvegamyhre:
[...]

PR #3656, opened 2026-01-17
[mxfp8 moe training] cuda blocked layout kernel handling for skinnier scale tensors

danielvegamyhre:
Stacked PRs:
 * #3658
 * __->__#3656

--- --- ---

## Summary
Fixes https://github.com/pytorch/ao/issues/3636

### Context
Bug report shows the user has a scales tensor with 44 columns, which this kernel does not handle well for 2 reasons:
1. TMA 2d transfers require stride be divisible by 16, and for this row major scale tensor, stride would be 44, resulting in a non-informative CUDA error when creating the Tensormap.
    - `"globalStrides array, which specifies tensor stride of each of the lower tensorRank - 1 dimensions in bytes, must be a multiple of 16 and less than 2^40.'"` - [source](https://docs.nvidia.com/cuda/archive/12.6.3/cuda-driver-api/group__CUDA__TENSOR__MEMORY.html) 
3. Even for tensors with stride % 16 == 0, the minimum CHUNK_WIDTH the kernel handles is 64 (this corresponds to a model dim of 2048, as 2048/32==64). For smaller models like DSV3 16b with intermediate dim = 1408 in the MoE layer, this kernel will fail. 

### Fix
- Add validation with clear error messages enforcing scale_cols % 16 == 0, which will direct the user to use the more flexible, but slower, Triton kernel if they have some odd model dim that results in scale_cols not divisible by 16.
- Support chunk_size 16, 32 to handle smaller model dims.

### Tests
- `pytest test/prototype/moe_training/test_kernels.py -v -s -k cuda_mx_block  `


drisspg (approved):
image

PR #3664, opened 2026-01-20
Fix Int8Tensor v2 to use float32 eps for better bfloat16 precision

jcaip:
- Add explicit eps=torch.finfo(torch.float32).eps to Int8Tensor.from_hp
- Ensures consistent behavior with v1 AffineQuantizedTensor
- Fixes SQNR regression with bfloat16 inputs (bfloat16 eps=0.0078 vs float32 eps=1.19e-07)


vkuzo (approved):

amdfaa:
This is failing on both rocm and cuda.
https://hud.pytorch.org/hud/pytorch/ao/3350b2fd9ccae1e3868045ca90514f5d1ae95a84/1?per_page=50&name_filter=&mergeEphemeralLF=true

jcaip:
@amdfaa thanks for flagging, should be able to fix this in a bit.

PR #3606, opened 2026-01-09
[mxfp8 moe training] add torch.compile support

danielvegamyhre:
Stacked PRs:
 * #3611
 * __->__#3606

--- --- ---

[mxfp8 moe training] add test for torch.compile

## Summary
- Add `@torch._dynamo.nonstrict_trace` to autograd functions to relax tensor metadata matching restrictions between forward output vs backward input
- Wrap `fill_indices_kernel` in custom op
- Pass process group name (instead of process group itself), then resolve/lookup the group by name inside the function, as a workaround until https://github.com/pytorch/pytorch/pull/172566 lands
- Add e2e test for forward and backward pass comparing outputs and grads for regular bf16 e2e EP pipline vs mxfp8 EP pipeline

PR #3667, opened 2026-01-21
[mxfp8 moe training] register constant with pytree

danielvegamyhre:
After #3606 `@torch._dynamo.nonstrict_trace` was added to support autograd functions having different tensor subclasses for forward output and backward input.

The e2e integration tests for compile and eager passed and it was landed, but it turns out if you explicitly specify a scaling recipe instead of defaulting, it breaks with this error:

```
torch._dynamo.exc.Unsupported: Invalid input type for nonstrict_trace-ed function Explanation: For nonstrict_trace-ed functions, only basic types (e.g., torch.Tensor, int, float) or pytree containers of those are allowed as inputs. The provided argument contains an unsupported type. Hint: Use one of the following to register the type with pytree: * torch.utils._pytree.register_constant * torch.utils._pytree.register_dataclass * torch.utils._pytree.register_pytree_node
```

Basically we need to register the enum with pytree, which is done in this PR.

### Tests
Added unit test cases which fail without this change.
- `pytest test/prototype/moe_training/test_scaled_grouped_mm.py -k dq`

### DIFF
--- a/test/prototype/moe_training/test_scaled_grouped_mm.py
+++ b/test/prototype/moe_training/test_scaled_grouped_mm.py
@@ -8,6 +8,7 @@ import pytest
 import torch
 from torch.nn import functional as F
 
+from torchao.prototype.mx_formats.config import ScaleCalculationMode
 from torchao.utils import is_sm_version, torch_version_at_least
 
 # We need to skip before doing any imports which would use triton, since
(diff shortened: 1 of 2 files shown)


liangel-02 (approved):

danielvegamyhre:
test failures unrelated, landing

PR #3672, opened 2026-01-21
[ROCM] Bringing Float8 Util Unit Tests for ROCM

alex-minooka:
Summary
Modified test case to accept ROCM specific value

ROCm's torch.log2 implementation has higher numerical precision than NVIDIA's, causing the largest normal fp32 number test case to floor to 2^127 instead of overflowing to inf. Updated test to accept platform-specific expected values and removed skip_if_rocm decorator to enable the test on ROCm.


alex-minooka:
@vkuzo reopened this PR https://github.com/pytorch/ao/pull/3643 here.

PR #3673, opened 2026-01-21
[ROCM] Float8 Test Base Unit Tests for ROCM

alex-minooka:
### Summary

ROCM supports float8 data type on mi300 and mi350, tests are now able to run.

### DIFF
--- a/test/float8/test_base.py
+++ b/test/float8/test_base.py
@@ -231,7 +231,7 @@ class TestFloat8TrainingTensor:
         ],
     )
     @unittest.skipIf(not torch.cuda.is_available(), "CUDA not available")
-    @unittest.skipIf(not is_ROCM and not is_sm_at_least_90(), "Requires CUDA capability >= 9.0")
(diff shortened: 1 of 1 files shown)


test/float8/test_base.py
```diff
@@ -652,7 +656,7 @@ class TestNumerics:
             torch.float8_e5m2fnuz,
         ],
     )
-    @unittest.skipIf(not torch.cuda.is_available(), "CUDA not available")
+    @unittest.skipIf(not is_ROCM() and not torch.cuda.is_available(), "CUDA not available")
```
vkuzo, line 659: is `is_ROCM()` different from `torch.cuda.is_available()`? I thought ROCM was a subset of cuda in terms of how it's integrated with PyTorch?

alex-minooka: is_ROCM() additionally checks for hip availability. 

On another note, I think this PR is doing the same thing as https://github.com/pytorch/ao/pull/3682, but the 3682 PR adds additional functionality that I think would be better for bringing back more ROCm tests moving forward. If 3682 could get looked at first I think it would make the process smoother.

PR #3674, opened 2026-01-21
[ROCM] Bring back Float8 test integrations unit tests

alex-minooka:
### Summary

Bringing back unit tests that were previously skipped by currently pass on ROCm

### DIFF
--- a/test/float8/test_numerics_integration.py
+++ b/test/float8/test_numerics_integration.py
@@ -25,7 +25,8 @@ from torchao.float8.float8_linear_utils import (
 from torchao.float8.float8_utils import compute_error
 from torchao.testing.training.test_utils import get_test_float8_linear_config
 from torchao.utils import (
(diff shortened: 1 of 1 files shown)


test/float8/test_numerics_integration.py
```diff
@@ -154,9 +155,8 @@ def _test_impl(self, config: Float8LinearConfig) -> None:
         [ScalingType.DYNAMIC],
     )
     @pytest.mark.skipif(
-        not is_sm_at_least_89(), reason="requires SM89 compatible machine"
+        not is_ROCM() and not is_sm_at_least_89(), reason="requires SM89 compatible machine"
```
wenchenvincent: We need to also differentiate the gpu targets on rocm

alex-minooka: added

Issue #3669, opened 2026-01-21
Torchao import caches CUDA state which can mess up device placement with Ray


patrickvonplaten:
As described in this issue: https://github.com/vllm-project/vllm/issues/25113#issuecomment-3397351690 importing torchao can have some nasty side effects such as messing with the CUDA state. If this happens before Ray is started in multi-node environments, Ray won't be able to correctly detect CUDA Device IDs causing vLLM multi-node servings to fail.

Can we make the torchao init more light weight to not influence the CUDA state? torchao is integrated quite deeply into libraries such as transformers which are imported everywhere.

vkuzo:
yes, we should do that, PR is up

PR #3676, opened 2026-01-21
make torchao import not initialize cuda

vkuzo:
Summary:

Should fix https://github.com/pytorch/ao/issues/3669. The approach is to make everything that initializes cuda state lazy.

Test Plan:

```
pytest -s -x test/test_lean_import.py
```

let's see how CI looks


jerryzh168:
looks like the CI errors are relevant

PR #3682, opened 2026-01-22
[ROCm] Enable Float8 base test suite

ZhiweiYan-96:
# Description

Enable float8 base tests in `test/float8/test_base.py` on ROCm backend.  Add utils for judging the Float8 support status on different platforms.

# Test

```bash
  pytest -sv test/float8/test_base.py
```

### DIFF
--- a/torchao/utils.py
+++ b/torchao/utils.py
@@ -1095,7 +1095,7 @@ def is_sm_at_least_100():
     )
 
 
-def platform_supports_float8():
+def platform_supports_float8_tensorwise():
     if torch.version.hip:
         return is_MI300() or is_MI350()
     else:
@@ -1111,7 +1111,7 @@ def platform_supports_float8_blockwise():
     if torch.version.hip:
         return is_MI300() or is_MI350()
     else:
-        return is_sm_at_least_100()
+        return is_sm_at_least_90()
 
 def is_cuda_version_at_least(major: int, minor: int) -> bool:
     if not torch.cuda.is_available():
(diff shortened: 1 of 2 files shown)


ZhiweiYan-96:
hi, @alex-minooka Mind have a look for this PR? I just see your PR https://github.com/pytorch/ao/pull/3673 here also unskip the case in `test_base.py`.  We may consider merging the change in your/my pr.

Any suggestion on my change in the `torchao/utils.py`? The utils might be more straightforward for choosing skip condition.

torchao/utils.py
```diff
@@ -1095,6 +1095,24 @@ def is_sm_at_least_100():
     )
 
 
+def platform_supports_float8():
```
vkuzo: name it `float8_tensorwise`?

torchao/utils.py
```diff
@@ -1095,6 +1095,24 @@ def is_sm_at_least_100():
...
+    if torch.version.hip:
+        return is_MI300() or is_MI350()
+    else:
+        return is_sm_at_least_90()
+
+def platform_supports_float8_blockwise():
+    if torch.version.hip:
+        return is_MI300() or is_MI350()
+    else:
+        return is_sm_at_least_100()
```
vkuzo: I don't think this is correct, blockwise should work on CUDA capability 9.0+

vkuzo (commented):
lg overall, left some comments inline

torchao/utils.py
ZhiweiYan-96: Thanks, that's more reasonable. Modified

torchao/utils.py
ZhiweiYan-96: great, change the version here.

ZhiweiYan-96:
Could you please help add ciflow/rocm and have a review for this PR @jithunnair-amd @vkuzo @alex-minooka？I don't have the access to add label. Thank you!

PR #3687, opened 2026-01-22
Add Static Activation Quantization Wrapper

namgyu-youn:
**Summary**:

- related to: #3655

Quantization tensor subclass (e.g., `Int8Tensor`, `Float8Tensor`) for static activation workflow requires a pre-computed activation scale (`act_quant_scale`), which means the activation should be updated before quantization. `act_quant_scale` can be computed by algorithms like AWQ/SmoothQuant. 

For example, AWQ workflow can be decomposed into:
```
(1) Prepare: Observer initialization
(2) Observe: Compute `act_quant_scale` based on algorithm and calibration flow (via `lm-eval`)
(3) Convert: Apply the `act_quant_scale` computed in (2)
(4) Quantize (RTN): Quantize using `quantize_(model, config)`
```

However, on the subclass (e.g., `Float8Tensor`, `Int8Tensor`) side, above workflow breaks subclass pattern, which should only address storage+inference ops. To handle this issue, this PR adds external wrapper to handle static activation workflow.

**Future Plan**:
- [ ] Update `Int8Tensor` to use this observer. (ref: https://github.com/pytorch/ao/pull/3687/changes/b373e101dafa670aa8d29841d0deb7d92f00d62e)
- [ ] Enum for custom use like AWQ
```py
class StaticObserver(Enum):
    """Observers supported to static quantization flow."""

    MINMAX = "affine"  # AffineQuantizedMinMaxObserver
    AWQ = "awq"  # AWQObserver
    SMOOTHQUANT = "smoothquant"  # SmoothQuantObserver
```
- [ ] Deprecate `Float8ObservedLinear` and update all callsite to use `ObserverLinear`
- [ ] Add NXFP format observer (see also: https://github.com/pytorch/ao/pull/3866#discussion_r2800398463)

**Test Plan**: CI

### DIFF
--- a/torchao/quantization/quantize_/common/observer_module.py
+++ b/torchao/quantization/quantize_/common/observer_module.py
@@ -11,7 +11,7 @@ from torchao.quantization.observer import AffineQuantizedMinMaxObserver
 
 # TODO: Build enum backend for more observers like AWQObserver, PerChannelHistogramObserver
 # User should be able to handle `act_obs` directly for easy customization
-class ObservedLinear(torch.nn.Linear):
+class ObservedTensor(torch.nn.Linear):
     """
     A linear module with an observer for static quantization.
 
@@ -59,7 +59,7 @@ class ObservedLinear(torch.nn.Linear):
         cls,
         float_linear: torch.nn.Linear,
         act_obs: "AffineQuantizedMinMaxObserver",
-    ) -> "ObservedLinear":
+    ) -> "ObservedTensor":
         """Create an observed linear from a float linear module."""
         observed_linear = cls(
             float_linear.in_features,


torchao/quantization/observer.py
```diff
@@ -346,3 +346,68 @@ def calculate_qparams(self):
             self.preserve_zero,
             self.zero_point_domain,
         )
+
+
+# TODO: Build enum backend for more observers like AWQObserver, PerChannelHistogramObserver
+# User should be able to handle `act_obs` directly for easy customization
+class ObservedLinear(torch.nn.Linear):
```
jerryzh168: is it possible to move this to https://github.com/pytorch/ao/tree/main/torchao/quantization/quantize_/common (create a new file like `observed_module.py`)

namgyu-youn: Sure, moved to `torchao/quantization/quantize_/common/observer_module.py`

jerryzh168:
@namgyu-youn we are actually thinking of move the ObservedLinear approach to use tensor subclass for observation: https://github.com/pytorch/ao/pull/3866#discussion_r2798336560

can you do that refactor first since you are working on this?

jerryzh168:
current PR is a mix of refactor + int8 changes, so would be good to split anyways. I'd recommend to:
PR 1: do observed linear --> observed tensor refactor
PR 2: int8 changes

namgyu-youn:
> @namgyu-youn we are actually thinking of move the ObservedLinear approach to use tensor subclass for observation: [#3866 (comment)](https://github.com/pytorch/ao/pull/3866#discussion_r2798336560)
> 
> can you do that refactor first since you are working on this?

Sure, small PR should always be better. Could you look into updated version?

jerryzh168:
@namgyu-youn I think it would be good to just refactor this to use `ObservedTensor` instead of `ObservedLinear` instead as I mentioned before.

Also in terms of labels, we are moving to per module release: https://github.com/pytorch/ao/issues/3889, and we'll be using labels from https://github.com/pytorch/ao/labels?q=module going forward, this PR belongs to `module: inference` I think

namgyu-youn:
@jerryzh168 renamed based on your suggestion, could you look into it?

jerryzh168:
@namgyu-youn the issue is not about renaming, it's about the flow itself, what we want is:

prepared model:
  linear.weight = ObservedTensor(...)

converted model:
  linear.weight = QuantizedTensor(linear.weight, ...)

namgyu-youn:
> @namgyu-youn the issue is not about renaming, it's about the flow itself, what we want is:
> 
> prepared model: linear.weight = ObservedTensor(...)
> converted model: linear.weight = QuantizedTensor(linear.weight, ...)

If I understood correctly, do you mean inherit subclass (like `Int8Tensor`, `Float8Tensor`) directly, not by linear module? So we can replace `ObservedTensor` with subclass at convert stage?

jerryzh168:
> If I understood correctly, do you mean inherit subclass (li

yes that's correct

namgyu-youn:
@jerryzh168 I think reland is better for fresh CI and updated context. Could you look into #3925 ?

PR #3688, opened 2026-01-22
Fix doc-preview S3 path for workflow_dispatch runs

svekars:
When triggered via `workflow_dispatch`, the `s3-prefix` was resolving to `pytorch/ao/{empty string}`,
causing files to upload to `pytorch/ao//index.html` (double slash). This made the preview inaccessible since the URL path didn't match.

Fix: Use `github.run_id` as fallback when PR number is unavailable, ensuring a valid path like
`pytorch/ao/{run_id}/index.html`.


jerryzh168:
let's test with manual run, seems the doc is not generated here by default (which is expected I guess)

PR #3708, opened 2026-01-23
[mxfp8 moe training] fallback cuda kernel for when input doesn't meet 2d TMA constraints

danielvegamyhre:
[mxfp8 moe training] fallback cuda kernel for when input doesn't meet 2d TMA constraints

## Context
- In https://github.com/pytorch/ao/issues/3636 the user reported a bug training DeepSeekV3 16b. TL;DR is the CUDA blocked layout kernel for groups along M doesn't support input tensors which don't meet 2d TMA constraint of stride being a multiple of 16 bytes.
- Short term mitigation (https://github.com/pytorch/ao/pull/3656) was to assert this constraint is met, and throw an informative error prompting the user to use the slower but more flexible Triton kernel for this instead.
- This PR lands a proper fix, which is a simple non-pipelined CUDA kernel that is:
   - ~5.5x faster than Triton for DSV3 16b shapes, but handles arbitrary column width
   - Dispatching to use faster pipelined kernel if we can, otherwise fallback to this simpler kernel.

## Tests
- `pytest test/prototype/moe_training/test_kernels.py -v -s -k cuda_mx_block  `

## Benchmarks
```
input_shape      chunks_per_tb    torch_time_us    triton_time_us    cuda_time_us  triton_speedup    cuda_speedup
-------------  ---------------  ---------------  ----------------  --------------  ----------------  --------------
(131072, 44)                 1          1105.65            123.94           19.46  8.92x             56.83x
(131072, 44)                 4          1101.41            109.57           19.46  10.05x            56.61x
(131072, 44)                 8          1042.5             207.97           19.46  5.01x             53.58x
```

E2E in Torchtitan on DSV3 16b I see an extra ~3% TPS speedup with single node dp2ep and the mxfp8_wgrad_with_hp + mxfp8 all2all/expert parallel enabled.

### DIFF
--- a/torchao/csrc/cuda/mx_kernels/mxfp8_extension.cpp
+++ b/torchao/csrc/cuda/mx_kernels/mxfp8_extension.cpp
@@ -39,6 +39,16 @@ void launch_mx_block_rearrange_2d_M_groups_cuda(
     int chunks_per_tb,
     cudaStream_t stream);
 
+void launch_mx_block_rearrange_2d_simple_cuda(
+    const uint8_t* scales_ptr,
+    int scale_rows,
+    int scale_cols,
+    const int32_t* input_group_end_offsets,
+    uint8_t* output_scales_ptr,
+    int num_groups,
+    int chunk_width,
+    cudaStream_t stream);
+
 // Helper for tensor validation
 void check_cuda_tensor(const at::Tensor &t, const char *name) {
   TORCH_CHECK(t.is_cuda(), name, " must be a CUDA tensor");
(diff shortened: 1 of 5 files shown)


torchao/csrc/cuda/mx_kernels/mxfp8_extension.cpp
```diff
@@ -39,6 +39,19 @@ void launch_mx_block_rearrange_2d_M_groups_cuda(
     int chunks_per_tb,
     cudaStream_t stream);
 
+<<<<<<< Updated upstream
```
slayton58: Remove merge detritus :)

PR #3718, opened 2026-01-26
[X86] Re-enable some inductor-related test cases

Xia-Weiwen:
These test cases were removed from PyTorch during the migration of PT2E. They test the X86Inductor backend.

### DIFF
--- a/torchao/testing/pt2e/utils.py
+++ b/torchao/testing/pt2e/utils.py
@@ -222,7 +222,7 @@ class PT2ENumericDebuggerTestCase(TestCase):
         )
 
 
-def get_default_quantizer(is_qat, is_dynamic):
+def get_default_x86_quantizer(is_qat, is_dynamic):
     """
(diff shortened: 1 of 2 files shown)


torchao/testing/pt2e/utils.py
```diff
@@ -213,3 +220,276 @@ def assertNodeSourcesEqual(self, node_source_1, node_source_2):
             node_source_1.name == node_source_2.name
             and node_source_1.graph_id == node_source_2.graph_id
         )
+
+
+def get_default_quantizer(is_qat, is_dynamic):
```
jerryzh168: nit: `get_default_cpu_quantizer`?

jerryzh168 (approved):

torchao/testing/pt2e/utils.py
Xia-Weiwen: Thanks. I have renamed it `get_default_x86_quantizer` which may be more accurate.

PR #3721, opened 2026-01-26
fix `test_parq` for transformers=5.0.0 release

vkuzo:
Summary:

The recent `transformers` v5 release broke the following test:

```bash
pytest test/prototype/test_parq.py -s -x -k test_tied_weights_quantization
```

You can see the failure in recent CI for PRs such as https://github.com/pytorch/ao/pull/3676

Fixing it here by moving the test case to the syntax expected by transformers v5

Test Plan:


andrewor14:
@lisjin

lisjin:
Thanks for the fix @vkuzo

PR #3724, opened 2026-01-26
[mxfp8 moe training] add emulated mode

danielvegamyhre:
Stacked PRs:
 * #3735
 * #3734
 * __->__#3724

--- --- ---

## Summary
Support emulated mode in `_to_mxfp8_then_scaled_grouped_mm` autograd function, which can be used for non-sm100 CI or developers without sm100 devices.

Emulated mode does the following:
   - Uses torch native dim0 quant, asserts `use_triton_for_dim0_cast` is False
   - Uses torch native 3d quantization along dim1 in backward, instead of CUDA kernel for 3d dim1 quant
   - Use torch native dim1 quant instead of CUDA kernel for 2d dim1 quant  
   - Uses hardware agnostic Triton kernel for scale factor blocked layout per group along M. Asserts `use_cuda_kernel_for_blocked_layout` is false
   - Dequantize and do bf16 `torch._grouped_mm` for all grouped gemms

## Tests
Added test cases for emulated mode. 
- `pytest test/prototype/moe_training/test_scaled_grouped_mm.py -v -s `
- `pytest test/prototype/mx_formats/test_mx_tensor.py -v -s`
- `pytest test/prototype/mx_formats/test_kernels.py -v -s`

Tested on H100 as well and the MXFP8 tests pass:
- `pytest test/prototype/moe_training/test_scaled_grouped_mm.py -v -s -k mx`

As an aside, some fp8 rowwise tests are failing now on h100 with odd errors like:
- `CUDA error: CUBLAS_STATUS_INVALID_VALUE when calling `::cublasLtMatmulDescSetAttribute(descriptor(), attr, &value, sizeof(value))``
- OOM (??) 

so we should deprecate this soon.

### DIFF
--- a/test/prototype/moe_training/test_scaled_grouped_mm.py
+++ b/test/prototype/moe_training/test_scaled_grouped_mm.py
@@ -42,6 +42,7 @@ from torchao.prototype.moe_training.utils import (
     generate_jagged_offs,
 )
 from torchao.prototype.mx_formats.mx_tensor import to_mx
+from torchao.quantization.quantize_.common import KernelPreference
 from torchao.testing.utils import skip_if_rocm
 
 # Needed since changing args to function causes recompiles
@@ -318,13 +319,15 @@ def test_emulate_mxfp8_grouped_gemm_2d_2d(M, N, num_experts):
 
 
 @skip_if_rocm("ROCm not supported")
-@pytest.mark.parametrize(
-    "M,K,N", [(16640, 5120, 8192), (131072, 5120, 8192), (131072, 8192, 5120)]
-)
+@pytest.mark.parametrize("M,K,N", [(32768, 5120, 8192), (16640, 7168, 2048)])
 @pytest.mark.parametrize("num_experts", (2, 4, 8, 16))
 @pytest.mark.parametrize("use_triton_for_dim0_cast", (True, False))
 @pytest.mark.parametrize("wgrad_with_hp", (True, False))
 @pytest.mark.parametrize("use_compile", (True, False))
+@pytest.mark.parametrize(
+    "kernel_preference", (KernelPreference.AUTO, KernelPreference.EMULATED)
+)
+@pytest.mark.parametrize("use_cuda_kernel_for_blocked_layout", (True, False))
 @pytest.mark.parametrize(
     "scale_mode", (ScaleCalculationMode.FLOOR, ScaleCalculationMode.RCEIL)
 )
@@ -336,8 +339,33 @@ def test_mxfp8_grouped_gemm_with_dq_fwd_bwd(
     use_triton_for_dim0_cast,
     wgrad_with_hp,
     use_compile,
+    kernel_preference,
+    use_cuda_kernel_for_blocked_layout,
     scale_mode,
 ):
+    # Emulated mode with compile is not supported
+    if kernel_preference == KernelPreference.EMULATED and use_compile:
+        pytest.skip(
+            "Skipping use_compile=True with kernel_preference=EMULATED, not currently supported"
+        )
+
+    if kernel_preference == KernelPreference.EMULATED and use_triton_for_dim0_cast:
+        pytest.skip("Triton kernel not supported in emulated mode.")
+
+    if (
+        kernel_preference == KernelPreference.EMULATED
+        and use_cuda_kernel_for_blocked_layout
+    ):
+        pytest.skip(
+            "CUDA blocked layout per group along M kernel not supported in emulated mode."
+        )
+
+    # MXFP8 hardware path requires SM100
+    if kernel_preference != KernelPreference.EMULATED and not is_sm_version(10, 0):
+        pytest.skip(
+            f"Skipping MXFP8 hardware mode tests, only supported on compute capability 10.0 and found {torch.cuda.get_device_capability()}"
+        )
+
     block_size = 32
     x = torch.randn(M, K, dtype=torch.bfloat16, device="cuda", requires_grad=True)
     w = torch.randn(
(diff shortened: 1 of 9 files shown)


torchao/prototype/moe_training/kernels/mxfp8/quant.py
```diff
@@ -701,11 +701,13 @@ def _blocked_group_start_idx(
     return group_start_idx
 
 
-mxfp8_cuda_extension_available = is_sm_at_least_100() and is_cuda_version_at_least(
-    12, 8
+_mxfp8_cuda_kernels_available = (
```
vkuzo, line 704: this should be done lazily, see https://github.com/pytorch/ao/pull/3676 for context

torchao/prototype/moe_training/scaled_grouped_mm.py
```diff
@@ -60,6 +76,8 @@ def _quantize_then_scaled_grouped_mm(
             and in column-major memory layout.
         offs (int32 torch.Tensor): The offsets to use to mark the starting index of each group along dim0 of the A tensor.
         out_dtype (Optional[torch.dtype]): The dtype of the output tensor. Currently only torch.bfloat16 is supported.
+        scaling_type (MoEScalingType): The scaling type to use for quantization.
+        emulated (bool): Whether to use emulated mode (dequantize and use BF16 grouped mm). Only applies to MXFP8 scaling types.
```
vkuzo: how about using `KernelPreference` (example: https://github.com/pytorch/ao/blob/70c31df4c5e78d4975a385ad6a561925568aab31/torchao/prototype/mx_formats/mx_tensor.py#L550), to align with future unification between MoE and dense code

test/prototype/moe_training/test_scaled_grouped_mm.py
```diff
@@ -336,8 +339,33 @@ def test_mxfp8_grouped_gemm_with_dq_fwd_bwd(
     use_triton_for_dim0_cast,
     wgrad_with_hp,
     use_compile,
+    kernel_preference,
+    use_cuda_kernel_for_blocked_layout,
     scale_mode,
 ):
+    # Emulated mode with compile is not supported
+    if kernel_preference == KernelPreference.EMULATED and use_compile:
+        pytest.skip(
```
vkuzo, line 348: just curious why not

test/prototype/moe_training/test_training.py
```diff
@@ -81,6 +90,21 @@ def test_moe_training(target_fqns: list[str], compile: bool, recipe_config: dict
         recipe_config["min_param_grad_sqnr"],
     )
     assert torch.cuda.is_available()
+
+    if kernel_preference == KernelPreference.EMULATED:
+        # FP8_ROWWISE doesn't support emulated mode
+        if recipe == MoEScalingType.FP8_ROWWISE:
+            pytest.skip(
+                "Skipping FP8 rowwise tests with kernel_preference=EMULATED, emulated mode only applies to MXFP8"
```
vkuzo, line 98: why not have this

torchao/prototype/moe_training/scaled_grouped_mm.py
```diff
@@ -334,6 +354,27 @@ def forward(
...
             Output tensor, shape (M, N)
         """
+        assert kernel_preference in (
+            KernelPreference.AUTO,
+            KernelPreference.EMULATED,
+        ), "kernel_preference must be AUTO or EMULATED"
+
+        # emulated mode validation
+        emulated = kernel_preference == KernelPreference.EMULATED
+        assert emulated or _SM100_KERNELS_AVAILABLE, (
```
vkuzo, line 364: how about something like this instead

```
if kernel_preference == auto:
    if is_sm100(...):
      ... call sm100 kernels...
    else:
      raise AssertionError('unsupported hardware')
eilf kernel_preference == emulated:
   ... emulation
else:
    raise AssertionError('unsupported kernel_preference')
```

torchao/prototype/mx_formats/kernels.py
```diff
@@ -166,76 +166,65 @@ def pack_uint4(uint8_data: torch.Tensor) -> torch.Tensor:
...
-                asm="cvt.rp.satfinite.ue8m0x2.f32 $0, 0.0, $1;",
-                constraints="=h,r",
-                args=[scale_input.to(tl.float32, bitcast=False)],
-                dtype=tl.uint16,
-                is_pure=True,
-                pack=1,
-            ).to(tl.uint8)
-        else:
-            tl.static_assert(SCALING_MODE == "floor")
+    def triton_to_mxfp8_dim1_reference(
```
vkuzo, line 169: why so many changes to this file?

if this is because we are making them lazy, then IMO this should be a separate PR

test/prototype/moe_training/test_training.py
danielvegamyhre: we could, we just don't, it would be a separate project, and i'd prefer to just deprecate/remove fp8 rowwise MoE

torchao/prototype/mx_formats/kernels.py
danielvegamyhre: it's because for emulated mode to work on non-sm100 machines rather than crash at compile time, I had to separate out the kernels that can compile on non-sm100 versus those that require sm100 to compile. 

The diff view is hard to read but if you open up the file itself, the kernels are grouped in 2 sections now. 

1. Hardware agnostic kernels that just require torch 2.7 and triton available: https://[github.com/pytorch/ao/blob/ec683d488a3d98bfacb1761256a267e173b4c116/torchao/prototype/mx_formats/kernels.py#L164](http://github.com/pytorch/ao/blob/ec683d488a3d98bfacb1761256a267e173b4c116/torchao/prototype/mx_formats/kernels.py#L164)

2. Kernels that require Sm100: https://github.com/pytorch/ao/blob/ec683d488a3d98bfacb1761256a267e173b4c116/torchao/prototype/mx_formats/kernels.py#L683

test/prototype/moe_training/test_scaled_grouped_mm.py
danielvegamyhre: i was spending a lot of time on cryptic compile errors trying to get it working until i decided it isn't really necessary for emulated mode. the non-emulated path compiles successfully.

PR #3725, opened 2026-01-26
[claude] Make rowwise_scaled_linear_sparse_cutlass ABI stable

andrewor14:
**Test Plan:**
```
pytest test/test_ops_rowwise_scaled_linear_sparse_cutlass.py
```
(needs #3768)

**Prompt:**

```
Make everything under this directory ABI stable:
ao/torchao/csrc/cuda/rowwise_scaled_linear_sparse_cutlass

Read these for instructions:
pytorch/docs/source/notes/libtorch_stable_abi.md
cppdocs/_sources/stable.rst.txt

Use these files for an example:
Before: flash-attention/hopper/flash_api.cpp
After: flash-attention/hopper/flash_api_stable.cpp

Additional instructions:
Replace TORCH_CHECK with STD_TORCH_CHECK without changing the error message
Replace c10::cuda::CUDAGuard with DeviceGuard
When calling aoti_torch_get_current_cuda_stream, get the device index from a tensor, not from the general torch::stable::accelerator::getCurrentDeviceIndex()
Where possible, use get_device_index instead of get_device
Before checking if CUDA_VERSION is defined, remember to include cuda.h

Don’t do these things:
Don’t define a dummy _C module that can be accessed from python
Don’t declare aoti_torch_get_current_cuda_stream, just include it from torch/csrc/inductor/aoti_torch/c/shim.h and add -DUSE_CUDA to both cxx and nvcc in setup.py if needed
Also add -DTORCH_TARGET_VERSION=0x020a000000000000 to both cxx and nvcc in setup.py
Don’t box kernels manually, just use TORCH_BOX
```

**Follow-up prompts:**

```
Seems like STD_CUDA_KERNEL_LAUNCH_CHECK is undefined, is it missing an import?
```

```
Revert all checks that check if the layout is strided, where the original code looks like this:

TORCH_CHECK(Wq.layout() == at::Layout::Strided, …)

Then, rewrite these checks using the following instead to make it ABI stable:

using torch::headeronly::Layout
int32_t wlayout;
TORCH_ERROR_CODE_CHECK(aoti_torch_get_layout(Wq.get(), wlayout));
STD_TORCH_CHECK(torch::stable::detail::to<Layout>(
    torch::stable::detail::from(wlayout)) == Layout::Strided, …
)

Do not change the error message.
```

```
Can you save "torch::stable::detail::to<Layout>(torch::stable::detail::from(xq_layout))" to a variable to reduce duplicate code? Same for the other checks
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,6 +8,7 @@ import glob
 import json
 import os
 import pickle
+import re
 import subprocess
 import sys
 import time
@@ -42,6 +43,38 @@ def read_requirements(file_path):
 def read_version(file_path="version.txt"):
     with open(file_path, "r") as file:
         return file.readline().strip()
+
+
+# Check if torch version is at least 2.10.0 (for stable ABI support)
+# util copied from torchao/utils.py 1
+def _parse_version(version_string):
+    """
+    Parse version string representing pre-release with -1
+
+    Examples: "2.5.0.dev20240708+cu121" -> [2, 5, -1], "2.5.0" -> [2, 5, 0]
+    """
+    # Check for pre-release indicators
+    is_prerelease = bool(re.search(r"(git|dev)", version_string))
+    match = re.match(r"(\d+)\.(\d+)\.(\d+)", version_string)
+    if match:
+        major, minor, patch = map(int, match.groups())
+        if is_prerelease:
+            patch = -1
+        return [major, minor, patch]
+    else:
+        raise ValueError(f"Invalid version string format: {version_string}")
+
+
+def _is_fbcode():
+    return not hasattr(torch.version, "git_version")
+
+
+def _torch_version_at_least(min_version):
+    if _is_fbcode():
+        return True
+
+    # Parser for local identifiers
+    return _parse_version(torch.__version__) >= _parse_version(min_version)
 
 
 SPINQUANT_REL_PATH = Path("torchao") / "prototype" / "spinquant"
@@ -616,6 +649,7 @@ def get_extensions():
 
     use_cutlass = False
     cutlass_90a_sources = None
+    stable_cutlass_90a_sources = None
     build_for_sm90a = False
     build_for_sm100a = False
     if use_cuda and not IS_WINDOWS:
@@ -649,19 +683,23 @@ def get_extensions():
         build_for_sm90a, build_for_sm100a = get_cutlass_build_flags()
         # Define sm90a sources
         cutlass_90a_sources = [
-            os.path.join(
-                extensions_cuda_dir,
-                "rowwise_scaled_linear_sparse_cutlass",
-                "rowwise_scaled_linear_sparse_cutlass_f8f8.cu",
-            ),
+            # TODO: move this to stable_cutlass_90a_sources in #3727
             os.path.join(
                 extensions_cuda_dir,
                 "to_sparse_semi_structured_cutlass_sm9x",
                 "to_sparse_semi_structured_cutlass_sm9x_f8.cu",
             ),
         ]
+
+        stable_cutlass_90a_sources = [
+            os.path.join(
+                extensions_cuda_dir,
+                "rowwise_scaled_linear_sparse_cutlass",
+                "rowwise_scaled_linear_sparse_cutlass_f8f8.cu",
+            ),
+        ]
         for dtypes in ["e4m3e4m3", "e4m3e5m2", "e5m2e4m3", "e5m2e5m2"]:
-            cutlass_90a_sources.append(
+            stable_cutlass_90a_sources.append(
                 os.path.join(
                     extensions_cuda_dir,
                     "rowwise_scaled_linear_sparse_cutlass",
@@ -670,6 +708,7 @@ def get_extensions():
             )
         # Always remove sm90a sources from main sources
         sources = [s for s in sources if s not in cutlass_90a_sources]
+        sources = [s for s in sources if s not in stable_cutlass_90a_sources]
 
     else:
         # Remove CUTLASS-based kernels from the sources list.  An
@@ -731,6 +770,7 @@ def get_extensions():
             )
 
     # Only build the cutlass_90a extension if sm90a is in the architecture flags
+    # TODO: delete this after #3726 and #3727
     if (
         cutlass_90a_sources is not None
         and len(cutlass_90a_sources) > 0
@@ -747,6 +787,38 @@ def get_extensions():
                 cutlass_90a_sources,
                 py_limited_api=True,
                 extra_compile_args=cutlass_90a_extra_compile_args,
+                extra_link_args=extra_link_args,
+            )
+        )
+
+    # Only build the cutlass_90a extension if sm90a is in the architecture flags
+    # and if torch version >= 2.10
+    if (
+        stable_cutlass_90a_sources is not None
+        and len(stable_cutlass_90a_sources) > 0
+        and build_for_sm90a
+        and _torch_version_at_least("2.10.0")
+    ):
+        stable_cutlass_90a_extra_compile_args = copy.deepcopy(extra_compile_args)
+        # Only use sm90a architecture for these sources, ignoring other flags
+        stable_cutlass_90a_extra_compile_args["nvcc"].append(
+            "-gencode=arch=compute_90a,code=sm_90a"
+        )
+        # Add -DUSE_CUDA for stable ABI support (using features in torch 2.10)
+        stable_cutlass_90a_extra_compile_args["cxx"].append("-DUSE_CUDA")
+        stable_cutlass_90a_extra_compile_args["cxx"].append(
+            "-DTORCH_TARGET_VERSION=0x020a000000000000"
+        )
+        stable_cutlass_90a_extra_compile_args["nvcc"].append("-DUSE_CUDA")
+        stable_cutlass_90a_extra_compile_args["nvcc"].append(
+            "-DTORCH_TARGET_VERSION=0x020a000000000000"
+        )
+        ext_modules.append(
+            extension(
+                "torchao._C_cutlass_90a_stable",
+                stable_cutlass_90a_sources,
+                py_limited_api=True,
+                extra_compile_args=stable_cutlass_90a_extra_compile_args,
                 extra_link_args=extra_link_args,
             )
         )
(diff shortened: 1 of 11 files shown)


torchao/csrc/cuda/rowwise_scaled_linear_sparse_cutlass/rowwise_scaled_linear_sparse_cutlass.cuh
```diff
@@ -223,27 +272,33 @@ void rowwise_scaled_linear_sparse_kernel_cutlass_sm9x(
...
 template<typename DtypeXq, typename DtypeWq, typename... Types>
 static void select_config(
-    const at::Tensor& Xq, const at::Tensor& X_scale, const at::Tensor& Wq,
-    const at::Tensor& W_meta, const at::Tensor& W_scale, const at::Tensor& bias,
-    at::Tensor& Y) {
-  const auto dprops = at::cuda::getCurrentDeviceProperties();
+    const Tensor& Xq, const Tensor& X_scale, const Tensor& Wq,
+    const Tensor& W_meta, const Tensor& W_scale, const Tensor& bias,
+    Tensor& Y) {
+  const auto dprops = get_device_prop(Xq);
```
andrewor14: @mikaylagawarecki @janeyx99 claude is adding a lot of extra code just to replace `at::cuda::getCurrentDeviceProperties`. What is the right way to migrate this?

torchao/csrc/cuda/rowwise_scaled_linear_sparse_cutlass/rowwise_scaled_linear_sparse_cutlass.cuh
```diff
@@ -37,6 +43,49 @@
...
+  device_properties[device_index] = device_prop;
+}
+
+cudaDeviceProp* get_device_prop(const Tensor& t) {
+  initVectors();
+  int device_index = t.get_device();
+  std::call_once(device_flags[device_index], initDeviceProperty, device_index);
+  return &device_properties[device_index];
+}
+} // anonymous namespace
```
andrewor14, line 116: @mikaylagawarecki here

mikaylagawarecki: Yes this looks right, iiuc it is almost an exact copy of what's in flash_ap_stable.cpp, and only change is how the device_index is acquired on line 83 and I think that is an improvement :)

a small nit is maybe use `t.get_device_index()` instead of `t.get_device()`

janeyx99:
LGTM, my main qs for these sort of PRs are "what's the test plan?". Does it pass existing tests?

jerryzh168:
> LGTM, my main qs for these sort of PRs are "what's the test plan?". Does it pass existing tests?

we added 2.10 in CI recently, probably can run the related test with that

setup.py
```diff
@@ -649,21 +669,26 @@ def get_extensions():
...
-                extensions_cuda_dir,
-                "rowwise_scaled_linear_sparse_cutlass",
-                "rowwise_scaled_linear_sparse_cutlass_f8f8.cu",
-            ),
+            # TODO: move this to stable_cutlass_90a_sources in #3727
             os.path.join(
                 extensions_cuda_dir,
                 "to_sparse_semi_structured_cutlass_sm9x",
                 "to_sparse_semi_structured_cutlass_sm9x_f8.cu",
             ),
```
andrewor14, line 691: @jerryzh168 I also separated `rowwise_scaled_linear_sparse_cutlass` and `to_sparse_semi_structured_cutlass_sm9x` here to make my PR pass tests.

I'm thinking once I land this you can move these few lines to `stable_cutlass_90a_sources`? Then eventually we can delete `cutlass_90a_sources`. What do you think

jerryzh168: yeah I can rebase if you can get the CI to pass

setup.py
```diff
@@ -753,6 +780,38 @@ def get_extensions():
             )
         )
 
+    # Only build the cutlass_90a extension if sm90a is in the architecture flags
+    # and if torch version >= 2.10
+    if (
+        stable_cutlass_90a_sources is not None
+        and len(stable_cutlass_90a_sources) > 0
+        and build_for_sm90a
+        and _parse_version(torch.__version__) >= [2, 10, 0]
```
jerryzh168: do we need to count for fbcode as well

jerryzh168: I was copy pasting all these functions https://github.com/pytorch/ao/pull/3727/changes#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R155-R182 seems much more complicated than before

torchao/csrc/cuda/rowwise_scaled_linear_sparse_cutlass/rowwise_scaled_linear_sparse_cutlass.cuh
```diff
@@ -223,27 +287,30 @@ void rowwise_scaled_linear_sparse_kernel_cutlass_sm9x(
...
+  status = gemm_op.initialize(arguments, workspace.data_ptr(), stream);
   CUTLASS_STATUS_CHECK(status, OPERATOR_NAME);
 
   // Perform mixed datatypes GEMM operation.
-  status = gemm_op.run(at::cuda::getCurrentCUDAStream());
+  status = gemm_op.run(stream);
   CUTLASS_STATUS_CHECK(status, OPERATOR_NAME);
 
-  C10_CUDA_KERNEL_LAUNCH_CHECK();
+  STD_CUDA_KERNEL_LAUNCH_CHECK();
```
andrewor14, line 318: @janeyx99 this doesn't exist today, what is a good replacement?

torchao/csrc/cuda/rowwise_scaled_linear_sparse_cutlass/rowwise_scaled_linear_sparse_cutlass.cuh
```diff
@@ -353,144 +420,135 @@ dispatch_on_bias(
...
   // Validate layouts of arguments.
-  TORCH_CHECK(Xq.dim() >= 2, OPERATOR_NAME,
+  STD_TORCH_CHECK(Xq.dim() >= 2, OPERATOR_NAME,
               " : Expected Xq argument to be 2D or higher-dimensional tensor, "
               " got ", Xq.dim(), " dims");
-  TORCH_CHECK(Xq.layout() == at::Layout::Strided, OPERATOR_NAME,
-              " : Expected Xq argument to be strided, got layout ",
-              Xq.layout());
-  TORCH_CHECK(X_scale.dim() == Xq.dim() - 1, OPERATOR_NAME,
+  STD_TORCH_CHECK(Xq.is_contiguous() || Xq.stride(-1) == 1, OPERATOR_NAME,
```
andrewor14: @janeyx99 this doesn't seem right, is there a good replacement for tensor.layout()?

torchao/csrc/cuda/rowwise_scaled_linear_sparse_cutlass/rowwise_scaled_linear_sparse_cutlass.cuh
janeyx99: huh...this should exist?

torchao/csrc/cuda/cutlass_extensions/common.h
```diff
@@ -10,7 +10,7 @@
 
 #define CUTLASS_STATUS_CHECK(status, message_prefix)                           \
   {                                                                            \
-    TORCH_CHECK(status == cutlass::Status::kSuccess, message_prefix,           \
+    STD_TORCH_CHECK(status == cutlass::Status::kSuccess, message_prefix,           \
```
jerryzh168: you'll need to change this locally in rowwise_scaled_linear_sparse_cutlass.cuh file I think

torchao/csrc/cuda/rowwise_scaled_linear_sparse_cutlass/rowwise_scaled_linear_sparse_cutlass.cuh
janeyx99: Would is_contiguous be too strict for the kernel? If not, is_contiguous is even more strict than layout is strided.

2.10 doesn't have Tensor.layout yet but we can add for 2.11. In the meantime, a not so nice workaround would be to move this check to python :/

torchao/csrc/cuda/cutlass_extensions/common.h
andrewor14: Hmm what do you mean? I already replaced all TORCH_CHECK in the .cuh file, but this one was failing compilation because it was inlined (and TORCH_CHECK was not replaced)

jerryzh168: oh, I followed the same fix in my PR first but found that `[torchao/csrc/cuda/cutlass_extensions/common.h](https://github.com/pytorch/ao/pull/3725/files/6589ec0d78011214d24c850b5579ad5c16d5c35f#diff-d7b8367f1b64ac9e770e0f286c78b8bd83a515e113c89496fe62ef023c1e7057)` is used by cu files both in your PR and my PR, and stable API include files (imports) are needed to use `STD_TORCH_CHECK` I think, so we can't modify this in the common.h (since the other half of the files are not using stable ABI in both our PRs), so we just have to do this locally https://github.com/pytorch/ao/pull/3727/changes#diff-818fa0804aed61661b3208bb032fdc1692086e8ce1e9ec0d191261e0fb8b93dbR37-R43 instead of in common.h

this is something that can be simplified if we do the stable ABI change in the same PR, but it's not too complicated to workaround

torchao/csrc/cuda/rowwise_scaled_linear_sparse_cutlass/rowwise_scaled_linear_sparse_cutlass.cuh
janeyx99: from offline discussion: this does exist!

jerryzh168 (approved):
please add fbcode check, before merging

torchao/csrc/cuda/rowwise_scaled_linear_sparse_cutlass/rowwise_scaled_linear_sparse_cutlass.cuh
```diff
@@ -30,14 +40,81 @@
...
+  cudaDeviceProp device_prop{};
+  cudaError_t err = cudaGetDeviceProperties(&device_prop, device_index);
+  if (err != cudaSuccess) {
+    STD_TORCH_CHECK(false, "cudaGetDeviceProperties failed: " +
+                               std::string(cudaGetErrorString(err)));
+  }
+  device_properties[device_index] = device_prop;
+}
+
+cudaDeviceProp* get_device_prop() {
```
andrewor14, line 97: here

Issue #3516, opened 2025-12-18
Making torchao ABI stable and moving closer to python only


jerryzh168:
Follow up of https://github.com/pytorch/ao/issues/1747, here is what we plan to do to make torchao ABI compatible and closer to python only, after this is done, torchao will be compatible with all pytorch versions and we don't need to worry about https://github.com/pytorch/ao/issues/2919

please feel free to pick up the tasks by adding your name to Status column

| Status | Assignee | File | Description | Plan |
| :---- | :---- | :---- | :---- | :---- |
| Deleted #3520 | @howardzhang-cv  | torchao/csrc/cuda/fp6\_llm/fp6\_linear.cu | FP6 linear layer | delete |
| Deleted #3612 | @howardzhang-cv  | torchao/csrc/cuda/marlin\_qqq/marlin\_qqq\_kernel.cu | Marlin QQQ quantization | delete |
| Deleted #3744 | @howardzhang-cv  | torchao/csrc/cuda/activation24/sparse\_gemm.cu | 2:4 sparse GEMM | delete |
| Deleted #3744 | @howardzhang-cv | torchao/csrc/cuda/activation24/sparsify24.cu | 2:4 sparsification | delete |
| Deleted #3613 | @howardzhang-cv | torchao/csrc/cuda/sparse\_marlin/marlin\_kernel\_nm.cu | N:M sparse Marlin | delete |
| Deleted #3722 | @howardzhang-cv | torchao/csrc/cuda/tensor\_core\_tiled\_layout/tensor\_core\_tiled\_layout.cu | Tensor core tiled layout | delete |
| ABI stable #3610 | @andrewor14 @danielvegamyhre  | torchao/csrc/cuda/mx\_kernels/mxfp8\_cuda.cu | MXFP8 CUDA kernels | Make ABI compatible |
| ABI stable #3610 | @andrewor14 @danielvegamyhre  | torchao/csrc/cuda/mx\_kernels/mxfp8\_extension.cpp | MXFP8 CUDA kernels | Make ABI compatible |
| No need to change | @andrewor14 @danielvegamyhre | torchao/csrc/cuda/mx_kernels/mx_block_rearrange_2d_M_groups.cu | MXFP8 CUDA kernels | Make ABI compatible |
[...]

PR #3727, opened 2026-01-26
Make to_sparse_semi_structured_cutlass_sm9x ABI stable

jerryzh168:
* __->__ #3727

Summary:
part of https://github.com/pytorch/ao/issues/3516

This PR also changed 

```
Float8DynamicActivationFloat8WeightConfig(
  version=2,
  packing_format=Float8PackingFormat.SPARSE_CUTLASS,
  granularity=PerRow(),
),
```

to only be available after torch 2.10.0, because we made the ops used ABI stable, and stable ABI APIs are only available after 2.10.0

Test Plan:
pip install -e . --no-build-isolation

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -789,9 +789,10 @@ def get_extensions():
                 "-DTORCH_TARGET_VERSION=0x020a000000000000",
             ]
         )
+        # stable ABI cutlass_90a module
         ext_modules.append(
             extension(
-                "torchao._C_cutlass_90a_stable",
+                "torchao._C_cutlass_90a",
                 cutlass_90a_sources,
                 py_limited_api=True,
                 extra_compile_args=cutlass_90a_extra_compile_args,


jerryzh168:
confirmed with Randy that it's fine that we change the availability of 

```
Float8DynamicActivationFloat8WeightConfig(
                    version=2,
                    packing_format=Float8PackingFormat.SPARSE_CUTLASS,
                    granularity=PerRow(),
                ),
```

to be only available after 2.10.0 since they are using nightly / pytorch main

setup.py
```diff
@@ -443,6 +451,7 @@ def get_extensions():
         print("If you'd like to compile ROCm extensions locally please install ROCm")
 
     use_cuda = torch.version.cuda and CUDA_HOME is not None
+    print("use cuda:", use_cuda)
```
andrewor14: remove?

setup.py
```diff
@@ -659,11 +669,6 @@ def get_extensions():
                 "rowwise_scaled_linear_sparse_cutlass",
                 "rowwise_scaled_linear_sparse_cutlass_f8f8.cu",
             ),
-            os.path.join(
-                extensions_cuda_dir,
-                "to_sparse_semi_structured_cutlass_sm9x",
-                "to_sparse_semi_structured_cutlass_sm9x_f8.cu",
-            ),
```
andrewor14, line 699: Do we need to separate these? Everything in the original list will either be made ABI stable or deleted. Maybe we can just keep this block but only build if torch >= 2.10?

setup.py
```diff
@@ -148,6 +148,12 @@ def use_debug_mode():
     _get_cuda_arch_flags,
 )
 
+from packaging.version import Version
+
+# Check if torch version is at least 2.10.0 (for stable ABI support)
+_torch_version_str = torch.__version__.split("+")[0].split("a")[0].split("b")[0]
```
andrewor14: a bit hard to read, should we just copy this one? https://github.com/pytorch/ao/blob/8940de9ce01345e44bdc063ebf7a0b2c5405a70d/torchao/__init__.py#L33

setup.py
andrewor14: oh I see, it's just a temporary thing since rowwise_scaled_linear_sparse_cutlass is not ABI stable yet in this PR?

setup.py
jerryzh168: yeah I'm still debugging the CI error right now..

setup.py
jerryzh168: OK sure

jerryzh168:
@janeyx99 @mikaylagawarecki @andrewor14 please take a look, I will remove the debug print before landing

torchao/csrc/cuda/to_sparse_semi_structured_cutlass_sm9x/to_sparse_semi_structured_cutlass_sm9x.cuh
```diff
@@ -27,16 +33,35 @@
...
+#define CUTLASS_STATUS_CHECK(status, message_prefix)                           \
+  {                                                                            \
+    STD_TORCH_CHECK(status == cutlass::Status::kSuccess, message_prefix,       \
+                    " : Got CUTLASS error: ", cutlassGetStatusString(status)); \
+  }
 #endif
 
 #define OPERATOR_NAME "to_sparse_semi_structured_cutlass_sm9x"
 
+// Macro for checking CUDA kernel launch errors (replacement for C10_CUDA_KERNEL_LAUNCH_CHECK)
```
janeyx99: does STD_CUDA_KERNEL_LAUNCH_CHECK not work here?

setup.py
```diff
@@ -149,6 +150,38 @@ def use_debug_mode():
 )
 
 
+# Check if torch version is at least 2.10.0 (for stable ABI support)
+# util copied from torchao/utils.py
```
janeyx99, line 154: why not import the util

setup.py
```diff
@@ -758,6 +802,60 @@ def get_extensions():
...
+            ]
+        )
+
+        print(
+            "cutlass_90a_stable_extra_compile_args cxx:",
+            cutlass_90a_stable_extra_compile_args["cxx"],
+        )
+        ext_modules.append(
+            extension(
+                "torchao._C_cutlass_90a_stable",
```
janeyx99: Would you maintain an unstable version for torch 2.10 minus? It might be wise to enable users to build these kernels on an older torch still?

torchao/csrc/cuda/to_sparse_semi_structured_cutlass_sm9x/to_sparse_semi_structured_cutlass_sm9x.cuh
```diff
@@ -113,56 +138,66 @@ to_sparse_semi_structured_kernel_cutlass_sm9x(const at::Tensor& W) {
...
   // Validate the input tensor layout.
-  TORCH_CHECK(W.dim() == 2, OPERATOR_NAME,
+  STD_TORCH_CHECK(W.dim() == 2, OPERATOR_NAME,
               " : Expected W argument to be 2D tensor,  got ", W.dim(),
               " dims");
-  TORCH_CHECK(W.layout() == at::Layout::Strided, OPERATOR_NAME,
-              " : Expected W argument to be strided, got layout ",W.layout());
+
+  // Note: layout() check removed as torch::stable::Tensor doesn't support it.
+  // The stride check below implicitly validates the tensor is strided.
```
janeyx99: Talked offline, but you could use the raw shim like so:

```
using torch::headeronly::Layout
int32_t wlayout;
AOTI_TORCH_ERROR_CODE_CHECK(aoti_torch_get_layout(W.get(), wlayout));
STD_TORCH_CHECK(torch::stable::detail::to<Layout>(
      torch::stable::detail::from(wlayout)) == Layout::Strided, ...)
```

setup.py
jerryzh168: this is the code that is installing torchao, so it is assuming torchao is not installed yet, you mean just import the file? is that too confusing?

setup.py
andrewor14: I think this is just a temporary thing to keep this PR and https://github.com/pytorch/ao/pull/3725 separate. After both PRs are landed there will be a single `torchao._C_cutlass_90a` and it will be the stable ABI version

torchao/csrc/cuda/to_sparse_semi_structured_cutlass_sm9x/to_sparse_semi_structured_cutlass_sm9x.cuh
andrewor14: TORCH_ERROR_CODE_CHECK

torchao/csrc/cuda/to_sparse_semi_structured_cutlass_sm9x/to_sparse_semi_structured_cutlass_sm9x.cuh
jerryzh168: let me try

setup.py
andrewor14: I think we want to keep setup.py as simple as possible. Importing torchao/utils.py might have side effects outside the scope of this file

setup.py
jerryzh168: yeah and we'll drop support for 2.10 minus afterwards

torchao/csrc/cuda/to_sparse_semi_structured_cutlass_sm9x/to_sparse_semi_structured_cutlass_sm9x.cuh
```diff
@@ -113,56 +125,71 @@ to_sparse_semi_structured_kernel_cutlass_sm9x(const at::Tensor& W) {
...
-to_sparse_semi_structured_cutlass_sm9x(const at::Tensor& W) {
+std::tuple<Tensor, Tensor>
+to_sparse_semi_structured_cutlass_sm9x(const Tensor& W) {
 #if defined(BUILD_TO_SPARSE_SEMI_STRUCTURED_CUTLASS_SM9X)
-  const auto dprops = at::cuda::getCurrentDeviceProperties();
-  const auto is_sm9x = dprops->major == 9;
+  // Get device properties using raw CUDA API.
+  int device_id = W.get_device();
+  cudaDeviceProp device_prop;
+  cudaError_t err = cudaGetDeviceProperties(&device_prop, device_id);
```
andrewor14: @janeyx99 @mikaylagawarecki is this right? I had to do something more complicated in my PR: https://github.com/pytorch/ao/pull/3725/changes#r2746685516

setup.py
andrewor14: gonna remove these before land?

torchao/csrc/cuda/to_sparse_semi_structured_cutlass_sm9x/to_sparse_semi_structured_cutlass_sm9x.cuh
andrewor14: (copied from the FA3 stable ABI PR: https://github.com/Dao-AILab/flash-attention/pull/1791)

janeyx99: I believe the more complicated code in FA3 is needed for perf, so you cache the device properties and only call it once.

jerryzh168: oh I see, I guess I can follow the same pattern, maybe just move that function to common.h?

PR #3706, opened 2026-01-23
Added benchmarking on Flux.1-schnell for new fp8 sdpa API

howardzhang-cv:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.13.0) (oldest at bottom):
* #3830
* #3781
* #3780
* #3779
* #3732
* #3731
* #3730
* __->__ #3706
* #3729

Summary: Added new benchmark folder in benchmarks/prototype/fp8_sdpa_inference
Benchmark uses Flux.1-schnell and tests LPIPs and runtime for fp8 SDPA versus regular SDPA versus fp8 linear

Test Plan: python benchmarks/fp8_sdpa_inference/eval_flux_fp8_sdpa.py --mode fp8_sdpa
python benchmarks/fp8_sdpa_inference/eval_flux_fp8_sdpa.py --mode fp8_linear

PR #3729, opened 2026-01-27
fp8_sdpa torch.compile functionality

howardzhang-cv:
* #3830
* #3781
* #3780
* #3779
* #3732
* #3731
* #3730
* #3706
* __->__ #3729

Summary:
Add in explicit dropout, attn_mask, enable_gqa check to fp8_sdpa
More closely mirrors F.scaled_dot_product_attention


torchao/prototype/fp8_sdpa_inference/fp8_sdpa_attention.py
```diff
@@ -44,13 +45,16 @@ def fp8_sdpa_parallel(
...
         scale: Optional scale factor for attention
+        enable_gqa: Whether to enable Grouped Query Attention (GQA)
         num_chunks: Number of chunks to split the S dimension into.
                     If None, automatically selects based on GPU SM count.
     """
     if attn_mask is not None:
         raise ValueError("attn_mask is not supported for FP8 SDPA")
     if dropout_p != 0.0:
         raise ValueError("dropout_p must be 0.0 for FP8 SDPA")
+    if enable_gqa:
```
drisspg: hmm I think it should be, do you know why we need to error for this

howardzhang-cv: Actually didn't realize that FA3 natively supported already. Was going to add in explicit support for enable_gqa in the pytorch side first then come back and remove this, but I realized it just auto works out of the box so I'll go ahead and remove this

drisspg: Yeah I think as long as you dont error in the _FA3 impl for this setting it shoudl be fine since the C++ filter checks this

PR #3730, opened 2026-01-27
Added new fp8_rope_sdpa_inference API

howardzhang-cv:
* #3830
* #3781
* #3780
* #3779
* #3732
* #3731
* __->__ #3730
* #3706
* #3729

## Summary
This PR introduces a new method that does RoPE + fp8 quantization in one operation, to be fed into the fp8 SDPA with FA3 backend.
- In order to replace multiple modules at once, it creates a custom attention module that includes the projection layers, norms, RoPE, and SDPA. In the new version, it replaces the RoPE + SDPA with the RoPE/fp8 quantization operation + fp8 SDPA.
- The RoPE/fp8 quantization operation is done in Triton, and uses a 3-kernel setup. The first kernel performs RoPE and calculates maxes on chunks of the sequence. The second kernel calculates the max of the entire head. The third kernel uses the scales to quantize to fp8, again done on chunks of the sequence. The chunking setup has much better SM utilization, so it's quite a bit faster, but since FA3 uses per-head quantization, it needs to be split up into 3 kernels separately.

## Results
Results: Here are the results of using this method. I'm also including the results of the other methods further up the stack here as well, and keeping those as draft PRs, just so the discussion is in one place.
#### Performance Table
| | fp8 | fp8 | fp8_rope | fp8_rope | fp8_ropeh | fp8_ropeh | fp8_rope_vh | fp8_rope_vh |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| Compile | NO | YES | NO | YES | NO | YES | NO | YES |
| Runtime | 1.04s | 0.78s | 0.84s | 0.76s | 0.93s | 0.87s | 0.89s | 0.82s |
| Speedup | 1.05x | 1.12x | 1.32x | 1.15x | 1.18x | 1.01x | 1.24x | 1.06x |
| LPIPS | .224 | .224 | .226 | .225 | .137 | .136 | .148 | .147 |

Results calculated on flux.1-schnell model using 4 inference steps on the DrawBench benchmark (as in PR #3502). All speedups and LPIPS image quality calculations are done against the standard model in bf16. Lower is better for LPIPS.
**fp8**: fp8 quantization Triton op into fp8 SDPA (as in PR #3641)
**fp8_rope**: this PR, using RoPE + fp8 quantization Triton op into fp8 SDPA
**fp8_ropeh**: RoPE + Hadamard transformation on QKV + fp8 quantization Triton op into fp8 SDPA (as in PR #3732)
**fp8_rope_vh**: RoPE + Hadamard transformation on V only + fp8 quantization Triton op into fp8 SDPA (as in PR #3780)

#### Results Summary
It seems like without torch.compile, the RoPE + fp8 quantization in one op does actually help runtime by quite a bit, but using torch.compile gets rid of most of that benefit. I logged the torch.compile and saw that it does not fuse the RoPE and fp8 quantization ops together, but it seems like that fusion only makes it slightly faster.
The Hadamard transformations do substantially help lower the LPIPS, but at the cost of runtime. In the torch.compiled version, including Hadamard on QKV means it is basically the same runtime. I looked into the literature and people typically get most of the benefit of Hadamard from just transforming V, so I tested that variant as well, and found that to be true. The LPIPS is only slightly higher (but still way better than without Hadamard), and we get back some runtime.

## Concerns
1. Using this method always requires knowing the model's attention module and making a copy of it, since it replaces multiple ops in a row, unlike a linear layer, or just SDPA by itself, where you can just overwrite that one op. This means the Python part needs to be specifically designed for each model, though the Triton part works for most models.
2. The RoPE + fp8 quantization does not actually lead to that much speedup when using torch.compiled. The speedup only goes from 1.12x to 1.15x. I imagine that most people who would use this would be using torch.compiled?

Because of the above, not sure if this method is actually useful, fp8_sdpa (from PR #3641) already provides a majority of the benefit of the FA3 fp8 SDPA and it's model agnostic. Thoughts?


torchao/prototype/fp8_rope_sdpa_inference/fp8_rope_sdpa_attention.py
```diff
@@ -0,0 +1,212 @@
...
+    B, S, H, D = query.shape
+    cos, sin = freqs_cis
+
+    # Ensure cos/sin are on the correct device and contiguous
+    cos = cos.to(query.device).contiguous()
+    sin = sin.to(query.device).contiguous()
+
+    # Fused RoPE + quantization using Triton kernel
+    # Input: [B, S, H, D], Output: [B, H, S, D] (ready for SDPA)
+    q_fp8, k_fp8, v_fp8, q_descale, k_descale, v_descale = fp8_rope_quantize_func(
```
drisspg, line 74: noob q?  did we measure against what torch.compile does for this pattern?

torchao/prototype/fp8_rope_sdpa_inference/fp8_rope_sdpa_attention.py
```diff
@@ -0,0 +1,212 @@
...
+    k_scale = torch.where(k_max > eps, FP8_MAX / k_max, torch.ones_like(k_max))
+    v_scale = torch.where(v_max > eps, FP8_MAX / v_max, torch.ones_like(v_max))
+
+    # Quantize in fp32 before casting to fp8
+    q_fp8 = (query_rope.float() * q_scale).to(torch.float8_e4m3fn)
+    k_fp8 = (key_rope.float() * k_scale).to(torch.float8_e4m3fn)
+    v_fp8 = (value_t.float() * v_scale).to(torch.float8_e4m3fn)
+
+    # Compute descales
+    q_descale = (
```
drisspg: nit; I think there are some common functions that can be be turned into utilities

torchao/prototype/fp8_rope_sdpa_inference/fp8_rope_sdpa_quantization.py
```diff
@@ -0,0 +1,523 @@
...
+import torch
+import triton
+import triton.language as tl
+
+# =============================================================================
+# Helper functions
+# =============================================================================
+
+
+def _compute_num_chunks(tensor: torch.Tensor, S: int) -> int:
```
drisspg, line 32: we should try implementing this in helion

torchao/prototype/fp8_rope_sdpa_inference/fp8_rope_sdpa_quantization.py
```diff
@@ -0,0 +1,523 @@
...
+    # Make tensors contiguous if needed
+    if not q.is_contiguous():
+        q = q.contiguous()
+    if not k.is_contiguous():
+        k = k.contiguous()
+    if not v.is_contiguous():
+        v = v.contiguous()
+    if not cos.is_contiguous():
+        cos = cos.contiguous()
+    if not sin.is_contiguous():
```
drisspg: I thikn you can just call .contiguous it does the `if not check`

torchao/prototype/fp8_rope_sdpa_inference/fp8_rope_sdpa_quantization.py
```diff
@@ -0,0 +1,523 @@
...
+        q_descale,
+        k_descale,
+        v_descale,
+        H,
+        num_chunks,
+    )
+
+    # Phase 2: Apply RoPE and quantize
+    grid_phase2 = (B, H, num_chunks)
+    rope_qkv_phase2_kernel[grid_phase2](
```
drisspg, line 534: Nit you call this 1 kernel but this is  3 kernels

torchao/prototype/fp8_rope_sdpa_inference/fp8_rope_sdpa_attention.py
howardzhang-cv: Yes, torch.compile does not fuse RoPE together with the quantize kernel, so this is technically faster, but only by a really thin margin (see table in the PR summary).

torchao/prototype/fp8_rope_sdpa_inference/fp8_rope_sdpa_attention.py
howardzhang-cv: Removed the reference implementation

torchao/prototype/fp8_rope_sdpa_inference/fp8_rope_sdpa_quantization.py
howardzhang-cv: Will work on this now and compare the Helion implementation to the Triton one

torchao/prototype/fp8_rope_sdpa_inference/fp8_rope_sdpa_attention.py
```diff
@@ -0,0 +1,94 @@
...
+
+    # Fused RoPE + quantization using Triton kernel
+    # Input: [B, S, H, D], Output: [B, H, S, D] (ready for SDPA)
+    q_fp8, k_fp8, v_fp8, q_descale, k_descale, v_descale = fp8_rope_quantize_func(
+        query, key, value, cos, sin, num_chunks
+    )
+
+    # Call PyTorch's FP8 SDPA
+    with sdpa_kernel(SDPBackend.FLASH_ATTENTION):
+        out = _scaled_dot_product_attention_quantized(
```
jerryzh168, line 80: q: is softmax quantized? is there anyway to control the accuracy of this op?

torchao/prototype/fp8_rope_sdpa_inference/fp8_rope_sdpa_quantization.py
howardzhang-cv: Fixed

torchao/prototype/fp8_rope_sdpa_inference/fp8_rope_sdpa_quantization.py
howardzhang-cv: Yeah I've changed the comments to reflect that. I originally had it as one kernel, but found that the chunking pattern and splitting it up into 2 phases with a reduction kernel in between was much faster.

howardzhang-cv:
> I think overall this PR needs a good /deslop pass tbh

Yeah sorry, that's my bad. Probably should have kept it as a draft for now. I did a pass and removed a lot of the unnecessary slop and cleaned up some of the code. Overall, still not too sure about this method though. Might just turn it back into a draft while we discuss as per the concerns in the summary above?

torchao/prototype/fp8_rope_sdpa_inference/fp8_rope_sdpa_attention.py
howardzhang-cv: Softmax is done in fp32. Unless we modify the flash attention 3 code, there's no way to control the accuracy of the softmax.

namgyu-youn:
> 2. The RoPE + fp8 quantization does not actually lead to that much speedup when using torch.compiled. The speedup only goes from 1.12x to 1.15x. I imagine that most people who would use this would be using torch.compiled?

If there is no meaningful speedup with torch.compile, can we reuse `_choose_scale_float8() + _quantize_affine_float8()`, like other FP8 subclasses pattern?

https://github.com/pytorch/ao/blob/6ad7c40461a5c8e79e442f17d8c25b47f0ae652f/torchao/quantization/quantize_/workflows/float8/float8_tensor.py#L30-L34

PR #3742, opened 2026-01-28
[DRAFT] Move all torch._scaled_mm calls to F.scaled_mm

slayton58:
* __->__ #3742

All `torch._scaled_mm` calls & references should now be replaced with
`F.scaled_mm`.

Changed some logic w.r.t. nvfp4 gemm + per-tensor-scales and bias to
reflect improved functionality.

Test Plan:

Relevant tests I've found pass:
```
pytest test/float8
pytest test/prototype/mx_formats
pytest test/prototype/blockwise_fp8_training
```
Note that there are plenty of failing tests, but some, like fp8 FSDP fail on my test machine without this patch, others, like DeepSeek-style gemms have to be running through triton on this (B200) node, as functionality through `F.scaled_mm` only exists for sm90 (Hopper). 

NOTE: This requires https://github.com/pytorch/pytorch/pull/173646 to be merged, otherwise compiled tests will fail.

Signed-off-by: Simon Layton <simonlayton@meta.com>

### DIFF
--- a/test/prototype/blockwise_fp8_training/test_blockwise_kernels.py
+++ b/test/prototype/blockwise_fp8_training/test_blockwise_kernels.py
@@ -24,7 +24,7 @@ from torchao.prototype.blockwise_fp8_training.kernels import (
     triton_fp8_gemm_1x128_128x128,
 )
 from torchao.testing.utils import skip_if_rocm
-from torchao.utils import is_sm_at_least_90
+from torchao.utils import is_sm_exactly_90
 
 BLOCKWISE_SIZE_MNK = [
     # (128, 128, 128),
@@ -39,7 +39,7 @@ BLOCKWISE_SIZE_MNK = [
 
 
 @pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA not available")
-@pytest.mark.skipif(not is_sm_at_least_90(), reason="Requires CUDA capability >= 9.0")
+@pytest.mark.skipif(not is_sm_exactly_90(), reason="Requires CUDA capability == 9.0")
 @pytest.mark.skipif(
     version.parse(triton.__version__) < version.parse("3.3.0"),
     reason="Triton version < 3.3.0, test skipped",
@@ -64,7 +64,7 @@ def test_triton_fp8_gemm_1x128_128x128(M, N, K, dtype):
 
 
 @pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA not available")
-@pytest.mark.skipif(not is_sm_at_least_90(), reason="Requires CUDA capability >= 9.0")
+@pytest.mark.skipif(not is_sm_exactly_90(), reason="Requires CUDA capability == 9.0")
 @pytest.mark.skipif(
     version.parse(triton.__version__) < version.parse("3.3.0"),
     reason="Triton version < 3.3.0, test skipped",
@@ -90,7 +90,7 @@ def test_triton_fp8_gemm_1x128_128x1(M, N, K, dtype):
 
 
 @skip_if_rocm("ROCm not supported")
-@pytest.mark.skipif(not is_sm_at_least_90(), reason="Requires CUDA capability >= 9.0")
+@pytest.mark.skipif(not is_sm_exactly_90(), reason="Requires CUDA capability == 9.0")
 @pytest.mark.parametrize("block_size", [128, 256])
 def test_triton_quantize_fp8_act_quant_lhs(block_size):
     device = "cuda"
@@ -137,7 +137,7 @@ def test_triton_quantize_fp8_act_quant_lhs(block_size):
 
 
 @skip_if_rocm("ROCm not supported")
-@pytest.mark.skipif(not is_sm_at_least_90(), reason="Requires CUDA capability >= 9.0")
+@pytest.mark.skipif(not is_sm_exactly_90(), reason="Requires CUDA capability == 9.0")
 @pytest.mark.parametrize("block_size", [128, 256])
 def test_triton_quantize_fp8_act_quant_rhs(block_size: int):
     device = "cuda"
@@ -184,7 +184,7 @@ def test_triton_quantize_fp8_act_quant_rhs(block_size: int):
 
 
 @skip_if_rocm("ROCm not supported")
-@pytest.mark.skipif(not is_sm_at_least_90(), reason="Requires CUDA capability >= 9.0")
+@pytest.mark.skipif(not is_sm_exactly_90(), reason="Requires CUDA capability == 9.0")
 @pytest.mark.parametrize("block_size", [128, 256])
 @pytest.mark.parametrize("M,K", [(4096, 1024), (4096, 4 * 4096)])
 def test_triton_quantize_fp8_act_quant_transposed_lhs(M, K, block_size: int):
@@ -233,7 +233,7 @@ def test_triton_quantize_fp8_act_quant_transposed_lhs(M, K, block_size: int):
 
 
 @skip_if_rocm("ROCm not supported")
-@pytest.mark.skipif(not is_sm_at_least_90(), reason="Requires CUDA capability >= 9.0")
+@pytest.mark.skipif(not is_sm_exactly_90(), reason="Requires CUDA capability == 9.0")
 @pytest.mark.parametrize("block_size", [128, 256])
 @pytest.mark.parametrize("M,K", [(4096, 1024), (4096, 4 * 4096)])
 def test_triton_quantize_fp8_weight_quant_rhs(M, K, block_size: int):
@@ -279,7 +279,7 @@ def test_triton_quantize_fp8_weight_quant_rhs(M, K, block_size: int):
 
 
 @skip_if_rocm("ROCm not supported")
-@pytest.mark.skipif(not is_sm_at_least_90(), reason="Requires CUDA capability >= 9.0")
+@pytest.mark.skipif(not is_sm_exactly_90(), reason="Requires CUDA capability == 9.0")
 @pytest.mark.parametrize("block_size", [128, 256])
 def test_triton_quantize_fp8_weight_quant_transposed_rhs(block_size: int):
     device = "cuda"
--- a/test/prototype/mx_formats/test_inference_workflow.py
+++ b/test/prototype/mx_formats/test_inference_workflow.py
@@ -32,6 +32,9 @@ torch.manual_seed(2)
 if not torch_version_at_least("2.8.0"):
     pytest.skip("Unsupported PyTorch version", allow_module_level=True)
 
+# Check if MSLK is available for MXFP4
+mslk_available = "USE_MSLK" in torch.__config__.show()
+
 
 # source: https://stackoverflow.com/a/22638709
 @pytest.fixture(autouse=True)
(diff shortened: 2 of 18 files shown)


vkuzo:
include test plan in PR summary? all MX functionality and all float8 (training, qat, inference) functionality should likely be tested

slayton58:
> include test plan in PR summary? all MX functionality and all float8 (training, qat, inference) functionality should likely be tested

I see a lot of tests that fail that either aren't due to these changes (fp8 FSDP for instance fails with & without changes), or due to not-available-functionality (no MSLK/FBGEMM_GEN_AI fails rather than skips) and so on.

I'll add the tests I know have to be run, happy to take advice on others.

vkuzo:
> pytest test/float8

for float8 training local testing:

```bash
./test/float8/test_everything_multi_gpu.sh
./test/float8/test_everything_single_gpu.sh
```

for float8 inference:

```bash
pytest test/quantization/quantize_/workflows/float8 -s
```

torchao/float8/float8_utils.py
```diff
@@ -27,6 +31,132 @@
...
+try:
+    from torch._inductor.utils import infer_scale_swizzle as _infer_scale_swizzle
+
+    infer_scale_swizzle = _infer_scale_swizzle
+except ImportError:
+    from collections.abc import (
+        Callable,
+    )
+
+    def _infer_scale_swizzle_impl(
```
vkuzo: why not just pass the swizzle enum directly? seems simpler and clearer to just choose the enum from the underlying recipe, and the recipe is always known in all of this product code

torchao/float8/inference.py
```diff
@@ -103,26 +108,35 @@ def addmm_float8_unwrapped_inference(
     versions of the linear module.
     """
 
+    scale_recipe_a, swizzle_a = infer_scale_swizzle(a_data, a_scale)
```
vkuzo: just choose swizzle here? its swizzled for nvfp4 and mx, and not swizzled otherwise?

vkuzo: there is also the https://github.com/pytorch/ao/blob/2cd02cfed317c08fed009b0f9e1d65a184d420af/torchao/prototype/mx_formats/mx_tensor.py#L455 and https://github.com/pytorch/ao/blob/2cd02cfed317c08fed009b0f9e1d65a184d420af/torchao/prototype/mx_formats/nvfp4_tensor.py#L74 `is_swizzled_scales` fields on the relevant tensors which could be used, but once the logic gets here the scale is already swizzled anyways

torchao/prototype/blockwise_fp8_training/linear.py
```diff
@@ -23,6 +25,29 @@
...
+    a: torch.Tensor,
+    b: torch.Tensor,
+    scale_a: torch.Tensor,
+    scale_b: torch.Tensor,
+    out_dtype: torch.dtype,
+) -> torch.Tensor:
+    scale_recipe_a, swizzle_a = infer_scale_swizzle(a, scale_a)
+    scale_recipe_b, swizzle_b = infer_scale_swizzle(a, scale_a)
+
+    return F.scaled_mm(
```
vkuzo: why do we need to wrap this, just call directly? the API is good, i feel like wrapping obfuscates things

torchao/float8/inference.py
vkuzo: since we are in `float8/inference.py`, should be no swizzling?

slayton58: This is effectively "just choosing swizzle" -- yes, it's set for mxfp{4,8} and nvfp4, `None` otherwise.

In format-specific files I've just filled in what `swizzle_{a,b}` have to be, but this seems to be a more general interface.

torchao/prototype/blockwise_fp8_training/linear.py
slayton58: The triton & `F.scaled_mm` calls have different signatures - either I need to setup partial functions for each of the unique gemm calls (1x128 x 128x128, 1x128 x 1x128 and so on) with scale_type etc. filled in, or a wrapper. I felt one wrapper was cleaner.

torchao/float8/float8_utils.py
slayton58: There seem to be two kinds of `scaled_mm` calls in the codebase:

1. General APIs that take `(a, b, scale_a, scale_b)` (like `float8_addmm_unwrapped` as an example) that then call down to `F.scaled_mm` -- these could (at least theoretically) pass anything.
2. Calls from type-specific bindings (nvfp4 and deepseek are the ones that immediately come to mind), which directly call `F.scaled_mm` and we know everything.

In the case of (1) I infer what the scale/swizzle types should be because a) I don't know much about what's going to be passed, and b) I haven't wanted to change the API to these general calls, especially when it's not strictly necessary.

vkuzo: > Calls from type-specific bindings

the intent for torchao code this PR is touching is for this style of code, there aren't general recipe-agnostic APIs in here

torchao/float8/inference.py
vkuzo: IMO we should inline how the swizzle is chosen instead of obfuscating, I want to see the real F.scaled_mm call without parsing through various helpers

in this callsite, it should be "no swizzle" as it's for float8

torchao/prototype/blockwise_fp8_training/linear.py
```diff
@@ -42,7 +67,7 @@ def forward(ctx, x, weight, block_size, out_dtype=torch.bfloat16, use_triton=Fal
         )
 
         # out = input @ weight.T
-        fp8_gemm = triton_fp8_gemm_1x128_128x128 if use_triton else torch._scaled_mm
+        fp8_gemm = triton_fp8_gemm_1x128_128x128 if use_triton else torch_scaled_mm_wrap
```
vkuzo: i kinda get why this is wrapping, but IMO the torch function should be the reference, and any custom functions that don't match torch should wrap

torchao/float8/inference.py
vkuzo: reading some of the other code in this PR, I think i feel like this:
* don't wrap F.scaled_mm (this is the real PyTorch API, just call it and make easy to follow)
* can wrap other gemms where the interface doesn't match (and ideally delete this if the PyTorch API improves)

wdyt

slayton58: Are you suggesting to remove the abstractions like `float8_addmm_unwrapped` and move `F.scaled_mm` calls directly into the recipes?

test/prototype/blockwise_fp8_training/test_blockwise_kernels.py
```diff
@@ -24,7 +24,7 @@
     triton_fp8_gemm_1x128_128x128,
 )
 from torchao.testing.utils import skip_if_rocm
-from torchao.utils import is_sm_at_least_90
+from torchao.utils import is_sm_exactly_90
```
vkuzo, line 27: nit: add a comment why only sm90 (pasted from our group chat)?

test/prototype/mx_formats/test_inference_workflow.py
```diff
@@ -95,6 +98,8 @@ def test_inference_workflow_mx(
     elif elem_dtype == torch.float4_e2m1fn_x2:
         if not is_sm_at_least_100() and not emulate:
             pytest.skip("CUDA capability >= 10.0 required for mxfp4 gemm")
+        elif not mslk_available:
```
vkuzo, line 101: hmm, I kind of prefer to see an error message:
1. in CI, we always install MSLK
2. locally, I'd rather know that I forgot to install MSLK rather than see a skip

torchao/float8/float8_utils.py
```diff
@@ -28,6 +29,32 @@
 }
 
 
+def infer_float8_scaling(
```
vkuzo, line 32: maybe `get_float8_scaling` or something similar? `infer` sounds like making an educated guess, but this function is exact

torchao/prototype/blockwise_fp8_training/linear.py
```diff
@@ -158,10 +190,13 @@ def __init__(
...
         )
-        assert is_sm_at_least_90(), "Only support SM90"
+        # Note: Triton kernels are broken (as-of 01/29/26), so must only run on
+        #       H100 w/cublasLt
+        assert is_sm_exactly_90(), "Only supported on SM90 + cublasLt"
         self.block_size = block_size
         self.dtype = dtype
-        self.use_triton = use_triton
+        # Must use triton on non-Hopper.
+        self.use_triton = use_triton or not is_sm_exactly_90()
```
vkuzo, line 199: throw an exception instead of silently changing the user preference?

agreed that its better to auto-decide this, but current API assumes user chooses it.

torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -462,57 +467,50 @@ def _addmm_nvfp4_dispatch(
...
-    # -----------------------------------------------------------------------------
-    result = torch._scaled_mm(
+
+    # Bias is not supported when output dtype == Float32, so note that, and add separately,
+    # post-scaled_mm call.
+    # Note: scaled_mm _does_ support bias w/per-tensor scales, which was previously
+    #       not supported.
+    should_add_bias_separately = (a._orig_dtype == torch.float32) and (bias is not None)
+
+    result = scaled_mm(
```
vkuzo, line 500: nit: how about just writing out the two different calls?  It's hard to follow the current logic where scales and swizzles are constructed based on vals of other args.  There are only 2 cases here, and being able to just stare at the code and instantly see the `F.scaled_mm` call with the right args is nice.

vkuzo:
looks great, thank you so much!

torchao/prototype/blockwise_fp8_training/linear.py
danielvegamyhre: I have no strong opinion here, just wanted to mention that if it makes things easier for you, feel free to remove the Triton fp8 blockwise gemms. 

I wrote the "1x128_128x1" one (used for wgrad in backward) and copied the "1x128_128x128" one from DSV3 inference code (used for forward and dgrad), but IIRC they are both much slower than the cuBLAS kernel torch scaled_mm dispatches to. 

If you want to remove and just always dispatch to scaled_mm, that's fine.

nikhil-arm:
Hello, May I Please know the plans to merge this PR?

The existing MXFP8 path in torchao uses aten::scaled_mm instead of F.scaled_mm ( scaled_mm_v2). 
Weirdly swizzle is hardcoded (assuming cuda always?) to be always true in the standard user facing MXDynamicActivationMXWeightConfig used for MXFP8.  https://github.com/pytorch/ao/pull/3200

There is also this PR: https://github.com/pytorch/ao/pull/4290, which appears to introduce further backend-specific branching rather than converging on a common API across the MXFP paths in TorchAO and cleaning up the kernel mapping in PyTorch. It would be useful to understand whether this is now the intended direction and how out of tree backends ( non CUDA/ROCM) are supposed to handle this.

cc: @vkuzo

vkuzo:
Hi @nikhil-arm , IMO https://github.com/pytorch/ao/pull/4290 is the way to go here.  

> appears to introduce further backend-specific branching rather than converging on a common API across the MXFP paths in TorchAO

Unfortunately, the backends are actually different.  NVIDIA hardware requires swizzled scales, and ROCm hardware requires unswizzled scales. So, there are two choices:
1. keep API the same, decide swizzling under the hood
2. expose swizzling with a config, make the user decide which config to use based on their target hardware

We are explicitly going for option 2 for the sake of simplicity.

PR #3678, opened 2026-01-21
Consolidate enum for observer step

namgyu-youn:
**Summary**:
Calibration-based algorithms (AWQ, SmoothQuant) operate independently of quantization granularity. To improve extensibility and consistency, add common enum instead of algorithm-specific definitions.

**Test plan**:
```bash
pytest -sv test/core/test_config.py
pytest -sv test/prototype/test_awq.py
pytest -sv test/prototype/test_smoothquant.py
```

**Example**: SmoothQuant ([arxiv:2211.10438](https://arxiv.org/abs/2211.10438)) decomposes into:

Quantization Flow
1. **Prepare**: Insert observers into linear layers
2. **Calibrate**: Compute scaling factors via calibration algorithm
3. **Convert**: Apply calibration results and quantize weights

Inference Flow: Load quantized model directly

PR #3757, opened 2026-01-29
Update GPTQ to use common enum step

namgyu-youn:
Summary:
Follow-up to #3517 and #3678. For future extensibility, update GPTQ to use a common enum for the quantization observer step.

Test plan:
```
pytest -sv test/prototype/gptq/test_gptqv2.py
```


torchao/prototype/gptq/api.py
```diff
@@ -82,8 +85,9 @@ def _gptq_config_transform(
 ) -> torch.nn.Module:
     """Unified transform handler that uses explicit step control."""
     tensor = getattr(module, parameter_name)
+    step = config.step
 
-    if config.step == "observe":
+    if step == QuantizationStep.PREPARE:
```
jerryzh168: does it work if user passes "prepare"?

namgyu-youn: Do you mean "convert" directly without "prepare"? That might not work because the "convert" stage requires GPTQ subclass (`GPTQObserverTensor`) input.

jerryzh168: I mean the string, I think we should be using strings instead of enum by default since it doesn't require people to import the QuantizationStep object

if you have time please submit a PR to change all QuantizationStep.PREPARE etc. to strings like "prepare", that would be easier to use I think

jerryzh168 (changes requested):
please use strings so users don't need to import the enum

torchao/prototype/gptq/api.py
namgyu-youn: OK, so in the future, we can (1) drop `QuantizationStep` entirely (update to string-based) and (2) file up https://github.com/pytorch/ao/blob/main/docs/source/eager_tutorials/static_quantization.rst how each stage works?

PR #3763, opened 2026-01-29
[5/x] docs update: initial clean up of quantization API reference section

vkuzo:
Summary:

1. delete "dtypes" as its outdated
2. organize by weight dtype
3. make sure that every config from
   https://github.com/pytorch/ao/issues/2752 is in the docs
4. move the examples for how to use the configs from quantization
   readme.md to the docs

This is definitely still not 100% clean, but way better than before and
we can improve further in future PRs.

Test Plan: look at new docs in CI


vkuzo:
* #3767
* #3766
* #3765
* #3764
* __->__ #3763

torchao/quantization/quant_api.py
```diff
@@ -1280,6 +1305,12 @@ class Float8WeightOnlyConfig(AOBaseConfig):
 
     Note:
         The actual matmul will be computed in original precision of the weight tensor.
+
+    Example::
+
+        # for torch 2.5+
```
jerryzh168, line 1311: nit: can probably remove since we don't support lower than 2.5 torch versions

torchao/quantization/quant_api.py
```diff
@@ -1398,6 +1429,15 @@ class Float8DynamicActivationFloat8WeightConfig(AOBaseConfig):
         set_inductor_config (bool): if True, adjusts `torchinductor` settings to recommended values.
         version (int): the version of the config, version 1 is deprecated, version 2 is using Float8Tensor (default)
 
+    Example::
+
+        # Tensorwise scaling
+        from torchao.quantization import quantize_, Float8DynamicActivationFloat8WeightConfig, PerTensor
+        quantize_(model, Float8DynamicActivationFloat8WeightConfig(granularity=PerTensor()))
+
+        # Rowwise scaling (recommended for torch 2.5+)
```
jerryzh168, line 1438: same here (removing torch 2.5+)

torchao/quantization/README.md
```diff
@@ -63,91 +63,19 @@ vllm bench throughput --num_prompts 128 --input_len 32 --output_len 2048 --max_m
...
-
-```python
-# for torch 2.5+
-from torchao.quantization import quantize_, PerRow, Float8DynamicActivationFloat8WeightConfig
-quantize_(model, Float8DynamicActivationFloat8WeightConfig(granularity=PerRow()))
-```
-
-Per-row scaling is only supported for bfloat16 weight and activation. This API is only tested on H100. Hardware with CUDA compute capability 8.9 or greater is required.
-
-#### Workaround with `unwrap_tensor_subclass` for `export`, `AOTI` and `torch.compile`
```
jerryzh168, line 133: maybe verify what is the current state of these things before removing? or add the current state as well in the current PR?

jerryzh168 (approved):
stamping, can address these later, please see comments inline

PR #3768, opened 2026-01-29
Fix rowwise_scaled_linear_sparse_cutlass test

andrewor14:
**Summary:**

```
pytest test/test_ops_rowwise_scaled_linear_sparse_cutlass.py
```

This gives

```
E       RuntimeError: The size of tensor a (4) must match the size of tensor b (65) at non-singleton dimension 1
```

Seems like it's caused by https://github.com/pytorch/ao/pull/3759. We just need to update the test.

**Test Plan:**

```
pytest test/test_ops_rowwise_scaled_linear_sparse_cutlass.py
```

### DIFF
--- a/test/test_ops_rowwise_scaled_linear_sparse_cutlass.py
+++ b/test/test_ops_rowwise_scaled_linear_sparse_cutlass.py
@@ -71,10 +71,10 @@ def run_test_for_op(
     W_aqt = W_quant_func(W, Wq_dtype)
 
     Xq = X_aqt.tensor_impl.float8_data
-    X_scale = X_aqt.tensor_impl.scale
+    X_scale = X_aqt.tensor_impl.scale.squeeze(-1)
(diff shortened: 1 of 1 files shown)


jerryzh168 (approved):

jerryzh168:
should we add the test run to CI?

andrewor14:
I think there are reasons why it wasn't added to CI, maybe because it needs special hardware or something? I'll let @jcaip do that when he gets back

jerryzh168:
> I think there are reasons why it wasn't added to CI, maybe because it needs special hardware or something? I'll let @jcaip do that when he gets back

I think you can probably add to H100 test, but I'm also enabling the v2 version using this op in my PR any ways: https://github.com/pytorch/ao/pull/3727/changes#diff-bd400f440bccb67a4d23b5cc5c3ce843a1a6219354e25d69cd167827f62e415b

PR #3774, opened 2026-01-30
[mxpf8 moe training] add unit tests to CI

danielvegamyhre:
Stacked PRs:
 * #3789
 * #3775
 * __->__#3774

--- --- ---

### [mxpf8 moe training] add unit tests to CI

### DIFF
--- a/.github/workflows/1xH100_tests.yml
+++ b/.github/workflows/1xH100_tests.yml
@@ -49,8 +49,9 @@ jobs:
         pip install . --no-build-isolation -vv
         pytest test/integration --verbose -s
         pytest test/dtypes/test_affine_quantized_float.py --verbose -s
-        pytest test/quantization/quantize_/workflows/float8 -v
+        pytest test/quantization/quantize_/workflows/float8 --verbose -s
         python test/quantization/quantize_/workflows/int4/test_int4_tensor.py
         python test/quantization/quantize_/workflows/int4/test_int4_preshuffled_tensor.py
         ./test/float8/test_everything_single_gpu.sh
-        pytest test/prototype/mx_formats/ -s
+        pytest test/prototype/mx_formats/ --verbose -s
+        pytest test/prototype/moe_training/test_scaled_grouped_mm.py --verbose -s
--- a/test/prototype/moe_training/test_scaled_grouped_mm.py
+++ b/test/prototype/moe_training/test_scaled_grouped_mm.py
@@ -49,8 +49,12 @@ from torchao.testing.utils import skip_if_rocm
 torch._dynamo.config.cache_size_limit = 1000
 
 
+@pytest.mark.skipif(
+    True,
+    reason="Skipping FP8 rowwise test pending fix for https://github.com/pytorch/ao/issues/3788",
+)
 @skip_if_rocm("ROCm not supported")
-@pytest.mark.parametrize("m", [131072])
+@pytest.mark.parametrize("m", [4096])
 @pytest.mark.parametrize("n", [8192])
 @pytest.mark.parametrize("k", [5120])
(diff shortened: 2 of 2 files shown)


.github/workflows/1xH100_tests.yml
```diff
@@ -54,3 +54,4 @@ jobs:
         python test/quantization/quantize_/workflows/int4/test_int4_preshuffled_tensor.py
         ./test/float8/test_everything_single_gpu.sh
         pytest test/prototype/mx_formats/ -s
+        pytest test/prototype/moe_training/test_scaled_grouped_mm.py -v -s -k mx
```
vkuzo: this works but is confusing, I don't really expect a name filter to apply in CI.  Can we just skip properly in the test file and remove this?

vkuzo: specifically, the `-k mx` part

danielvegamyhre: is there a way to skip the test in CI but keep it runnable locally? i just wanted certain prototype tests to not be blocking in CI but still be usable for development

is there an env var I can check or something?

danielvegamyhre: i think in the workflow yaml i can just `export IS_CI=1` then pytest skipIf that is set

test/prototype/moe_training/test_scaled_grouped_mm.py
```diff
@@ -49,8 +50,13 @@
...
 
 
+# context: https://github.com/pytorch/ao/issues/3788
+@pytest.mark.skipif(
+    os.environ.get("IS_CI"), 
+    "Skipping FP8 rowwise test in CI"
+)
 @skip_if_rocm("ROCm not supported")
-@pytest.mark.parametrize("m", [131072])
+@pytest.mark.parametrize("m", [4096])
```
danielvegamyhre, line 57: unrelated but fixes OOM on a fp8 test case

.github/workflows/1xH100_tests.yml
vkuzo: just gate it on cuda capability 10.0?

it's confusing to run something locally but skip in CI even it can pass in CI

danielvegamyhre: The MX tests are gated on sm100 and all run in CI. 

It is the fp8 rowwise tests that I want to skip in CI so they are non-blocking but keep runnable locally. If I just gate on sm 90 they will still run in CI.

a different solution would be to move fp8 rowwise tests to a different file, and don't run that that one in CI.

i think being able to be more specific about which tests are blocking in CI is a good thing / useful though?

vkuzo: > non-blocking but keep runnable locally

to keep it simple, if the test works, run it in CI? If it fails, skip it everywhere?  "run this locally but not CI" is a footgun.

danielvegamyhre: i made the change, skip with linked github issue to fix.

edit: need to update yaml now, will after meeting

danielvegamyhre: @vkuzo ok this is ready now

.github/workflows/1xH100_tests.yml
```diff
@@ -54,3 +54,4 @@ jobs:
         python test/quantization/quantize_/workflows/int4/test_int4_preshuffled_tensor.py
         ./test/float8/test_everything_single_gpu.sh
         pytest test/prototype/mx_formats/ -s
+        pytest test/prototype/moe_training/test_scaled_grouped_mm.py -v -rs
```
vkuzo: nit: it's weird for the pytest flags to be inconsistent between invocations, I'd advocate for the same flags everywhere

PR #3777, opened 2026-01-30
[Reland][X86] Re-enable some inductor-related test cases

Xia-Weiwen:
Reland https://github.com/pytorch/ao/pull/3718 and skip rocm

### DIFF
--- a/test/quantization/pt2e/test_x86inductor_cpp_wrapper.py
+++ b/test/quantization/pt2e/test_x86inductor_cpp_wrapper.py
@@ -24,7 +24,6 @@ from torch.testing._internal.common_device_type import (
 from torch.testing._internal.common_utils import (
     IS_MACOS,
     IS_WINDOWS,
-    TEST_WITH_ROCM,
     slowTest,
 )
 from torch.testing._internal.inductor_utils import HAS_CPU
@@ -175,7 +174,7 @@ def make_dynamic_cls(cls, xfail_prop="_expected_failure_dynamic"):
     )
 
 
-if RUN_CPU and not TEST_WITH_ROCM and torch_version_at_least("2.8.0"):
+if RUN_CPU and (torch.version.hip is None) and torch_version_at_least("2.8.0"):
 
     class BaseTest(NamedTuple):
         name: str


andrewor14:
Hi @Xia-Weiwen sorry but we need to revert this again since it broke some internal builds. Seems like importing a test within a test is a problem.

test/quantization/pt2e/test_x86inductor_cpp_wrapper.py
```diff
@@ -0,0 +1,295 @@
...
+
+from torchao.utils import torch_version_at_least
+
+try:
+    try:
+        from . import (
+            test_x86inductor_fusion,
+        )
+    except ImportError:
+        import test_x86inductor_fusion
```
andrewor14, line 39: Hi @Xia-Weiwen can you clean up this import? It's a bit messy. Usually when we want to import a test utils we put it under torchao/testing instead. Do you mind fixing this in the reland PR?

Xia-Weiwen: Hi @andrewor14  We import another test file in this test file because it want to reuse the test cases in that test file. This test file tests Inductor's CPP wrapper and we create test cases from the cases in the other file which tests the python wrapper. I am doing this because this is how torch handles the case. These test cases were in PyTorch previously and I am just moving them here.
https://github.com/pytorch/pytorch/blob/5ad9b89c54e1eba6fcd2ea2d0b44dad12c66c255/test/inductor/test_cpu_cpp_wrapper.py#L22C1-L41C1
Perhaps I should keep everything the same as in PyTorch. Do you have any suggestions?

CC @jerryzh168

Xia-Weiwen: @andrewor14  @jerryzh168 I have updated this part but not sure if it would work:
```python
try:
    try:
        from . import (
            test_x86inductor_fusion,
        )
    except ImportError:
        import test_x86inductor_fusion  # @manual
except unittest.SkipTest:
    if __name__ == "__main__":
        sys.exit(0)
    raise
```
Note: the change is `# @manual`.
The new PR is here: https://github.com/pytorch/ao/pull/3826

Can you please import it to fbcode to verify? Thanks.

Issue #3576, opened 2026-01-06
[RFC] XPU Enabling Plan for TorchAO


liangan1:
## Motivation
 
Since the software stack is almost 1x1 mapping between the CUDA and XPU, we plan to align the torchAO feature with CUDA on the XPU(except for the CUDA specific features) to help user seamlessly transfer from CUDA device to XPU device.
In general, we will first focus on features used by the internal tasks. In the past few months, we have upstream the int4, int8 and fp8 with tensor- and channel-wise scaling. Since the mxfp8 and mxfp4 will be natively supported by the Intel Crescent Island(CRI) data center GPU, these two data type will be the highest priority for XPU. As for other features in the torchAO, we will gradually enable them by priority.
## Goal

- Similar user experience as CUDA
- Almost zero migration effort from CUDA to XPU
- On-par feature scope with CUDA and same code quality which is ensured by the CI.

## Methodology

- Reuse TorchAO code as much as possible
E.g., Same quantization config for int8/fp8/mxfp8 with CUDA
- Device dispatch is processed by PyTorch and Keep the kernel in PyTorch core
E.g., [scaled_mm](https://github.com/pytorch/pytorch/pull/166056/files) for fp8, [int_mm ](https://github.com/pytorch/pytorch/pull/166056/files)for int8
- Only defined XPU-specific config when the CUDA version is not applicable for XPU
 E.g., Int4PlainInt32Tensor for XPU used by woq-int4

## Features Plan

liangan1:
cc @jerryzh168  @Stonepia  @xiaowangintel

vkuzo:
thank you, @liangan1 ! I have a couple of questions:
1. do we have these devices in CI? 
2. what is the best place to understand the performance characteristics of each of these GPUs for various low precision dtypes?  Are there technical specification sheets available? If yes, maybe we can link them to this issue?  Specifically, it would be great to know, for each GPU, (a) which dtypes are supported, (b) peak OPS for dot product by dtype, (c) peak memory bandwidth.
3. do we have a process already if there are XPU-specific failures in the future - would someone from Intel be available to help support?

liangan1:
> thank you, @liangan1 ! I have a couple of questions:
> 1. do we have these devices in CI? 
[...]

PR #3782, opened 2026-01-30
[reland][xpu] INT8 quantization on Intel XPU

zxd1997066:
# Motivation
This feature is one of features listed in the #3576. There is the XMX intrinsic on the Intel GPUs, e.g., Intel Arc B580, to speedup the int8 computation. This PR is target to ensure the int8 quantization works on the XPU and user can use same quantization config with CUDA. 
# Implementation

# Usage
Users can use same quantization config, such as Int8StaticActivationInt8WeightConfig or Int8DynamicActivationInt8WeightConfig, as CUDA to run the fp8 models on XPU. This device dispatched by the PyTorch op.
```

#For static int8 quantization
            granularity=granularity,
        )
```

### DIFF
--- a/test/quantization/quantize_/workflows/int8/test_int8_tensor.py
+++ b/test/quantization/quantize_/workflows/int8/test_int8_tensor.py
@@ -301,14 +301,15 @@ class TestInt8Tensor(TorchAOIntegrationTestCase):
             weight_cpu.dequantize(), weight_pinned.dequantize(), atol=0, rtol=0
         )
 
-    def test_int8_weight_only_v1_v2_per_group_equivalence(self):
+    @common_utils.parametrize("device", get_available_devices())
+    def test_int8_weight_only_v1_v2_per_group_equivalence(self, device):
         """Test that v1 per-group and v2 PerGroup produce equivalent outputs."""
         torch.manual_seed(42)
         group_size = 64
         K, N = 256, 128
 
         model_v1 = ToyTwoLinearModel(
-            K, N, K, dtype=torch.bfloat16, device="cuda"
+            K, N, K, dtype=torch.bfloat16, device=device
         ).eval()
         model_v2 = copy.deepcopy(model_v1)
 
@@ -318,7 +319,7 @@ class TestInt8Tensor(TorchAOIntegrationTestCase):
         quantize_(model_v1, config_v1)
         quantize_(model_v2, config_v2)
 
-        input_tensor = torch.randn(32, K, dtype=torch.bfloat16, device="cuda")
+        input_tensor = torch.randn(32, K, dtype=torch.bfloat16, device=device)
 
         with torch.no_grad():
             output_v1 = model_v1(input_tensor)


test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -227,6 +236,7 @@ def test_available_gpu_kernels(self):
             "extern_kernels._int_mm", 1
         ).check_count("triton_poi_fused", 1).run(code[0])
 
+    @skip_if_xpu("XPU enablement in progress")
```
liangan1:
```suggestion
    @skip_if_xpu("pin memory is not supported on XPU")
```

liangan1 (approved):

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -162,8 +168,9 @@ def test_slice(self, config, device, dtype):
     def test_index_select(self, config):
         """test that `x_0 = x[0]` works when `x` is a 2D quantized tensor."""
         N, K = 256, 512
-        x = torch.randn(N, K, device="cuda", dtype=torch.bfloat16)
-        linear = torch.nn.Linear(K, N, bias=False, dtype=torch.bfloat16, device="cuda")
+        device = get_current_accelerator_device()
```
jerryzh168: this is not that different, can you use `@common_utils.parametrize("device", get_available_devices())` (used in L140) everywhere?

jerryzh168 (changes requested):
please see inline comments

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
zxd1997066: Sure, let me have a try.

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -205,20 +213,21 @@ def test_dequantization_accuracy(self, config):
     @unittest.skipIf(
         not torch_version_at_least("2.7.0"), "torch 2.6.0 and below has custom fx pass"
     )
-    def test_available_gpu_kernels(self):
+    @common_utils.parametrize("device", get_available_devices())
+    def test_available_gpu_kernels(self, device):
```
zxd1997066, line 256: Hi @jerryzh168 , seems ```@common_utils.parametrize("device", get_available_devices())``` is not suitable for each case, for this one, ```get_available_devices()``` will return ```["cpu", "cuda"]```, and ```test_available_gpu_kernels_device_cpu``` encounter the runtime error ```RuntimeError: Expected to find "triton_per_fused" but did not find it```

jerryzh168: is this resolved now?

jerryzh168: this seems reasonable, since kernels are device specific, I think you show restrict the devices to cuda or expand the test to include expected kernels for other devices (cpu/xpu) as well

zxd1997066: Thanks, updated to use ```get_current_accelerator_device()``` for this case.

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
liangan1: @zxd1997066 updated?

zxd1997066: yes, only ```test_available_gpu_kernels``` uses ```get_current_accelerator_device()``` now. others are using ```@common_utils.parametrize("device", get_available_devices())``` https://github.com/pytorch/ao/pull/3782#discussion_r2752279853

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -226,15 +235,16 @@ def test_available_gpu_kernels(self):
         torch.compiler.reset()
 
         M, K, N = 128, 256, 512
+        device = get_current_accelerator_device()
```
jerryzh168: this should be removed I think

jerryzh168: we should probably try to remove `get_current_accelerator_device` in the end and replace everything with `get_available_devices`

zxd1997066: Hi @jerryzh168 , ```get_available_devices``` seems can not be used here, as it will return ["cpu", "cuda"], and ```test_available_gpu_kernels_device_cpu``` will encounter the runtime error ```RuntimeError: Expected to find "triton_per_fused" but did not find it```

So I changed to ```device = "xpu" if torch.xpu.is_available() else "cuda"```

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -226,15 +234,16 @@ def test_available_gpu_kernels(self):
         torch.compiler.reset()
 
         M, K, N = 128, 256, 512
+        device = "xpu" if torch.xpu.is_available() else "cuda"
```
jerryzh168: just device from arg?

zxd1997066: Thanks, updated

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
jerryzh168: @zxd1997066 if you have time probably makes sense to put up a PR to remove `get_current_accelerator_device` and refactor all the callsites I think

jerryzh168:
seems xpu tests are failing

liangan1:
> * [xpu-test / test](https://hud.pytorch.org/pr/pytorch/ao/3782#68311744102) ([gh](https://github.com/pytorch/ao/actions/runs/23476976654/job/68311744102))
>   `torchao/test/prototype/safetensors/test_safetensors_support.py::TestSafeTensors::test_safetensors_sharded_config7_act_pre_scale_False`

Yes. you are right. The UT fails is not related to this PR changes and will be fixed by  https://github.com/pytorch/ao/pull/4049

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
zxd1997066: Sure, I will take a look.

zxd1997066:
> > * [xpu-test / test](https://hud.pytorch.org/pr/pytorch/ao/3782#68311744102) ([gh](https://github.com/pytorch/ao/actions/runs/23476976654/job/68311744102))
> >   `torchao/test/prototype/safetensors/test_safetensors_support.py::TestSafeTensors::test_safetensors_sharded_config7_act_pre_scale_False`
> 
> Yes. you are right. The UT fails is not related to this PR changes and will be fixed by #4049

Yes, it will be fixed by https://github.com/pytorch/ao/pull/4049

PR #3784, opened 2026-01-30
[X86] Update smooth quant fusion pattern

cyxlily:
The SmoothQuant implementation has changed in Torchao recently, so we need to update the fusion patterns in FX graph passes of Inductor accordingly.
Besides, we added a `reduce_range` argument in `Int8Tensor` to accommodate CPUs without AVX512_VNNI to avoid overflow.
Unit tests are updated too.


torchao/quantization/pt2e/inductor_passes/x86.py
```diff
@@ -1772,65 +1778,102 @@ def _register_smooth_quant_int_mm_pattern():
...
     # When torch.compile'ing with dynamic=True, the expand node and the two tailing reshape nodes exist
     # When torch.compile'ing with dynamic=False, they don't exist
-    def get_pattern_no_bias(expand_a_scale: bool, reshape_a: bool = True):
-        return CallFunction(
-            aten.mul.Tensor,
-            CallFunction(
+    def get_pattern_no_bias(
+        expand_a_scale: bool, reshape_a: bool = True, convert_a: bool = False
+    ):
+        if convert_a:
```
Xia-Weiwen: the patterns in `if` and `else` are much the same. Please extract the common part for reuse.

torchao/quantization/pt2e/inductor_passes/x86.py
```diff
@@ -1772,65 +1778,102 @@ def _register_smooth_quant_int_mm_pattern():
 
     # When torch.compile'ing with dynamic=True, the expand node and the two tailing reshape nodes exist
     # When torch.compile'ing with dynamic=False, they don't exist
-    def get_pattern_no_bias(expand_a_scale: bool, reshape_a: bool = True):
-        return CallFunction(
-            aten.mul.Tensor,
-            CallFunction(
+    def get_pattern_no_bias(
+        expand_a_scale: bool, reshape_a: bool = True, convert_a: bool = False
```
Xia-Weiwen: I remember you removed the expand op. Is expand_a_scale always false? If so, why not removing it?

torchao/quantization/pt2e/inductor_passes/x86.py
```diff
@@ -1772,65 +1778,102 @@ def _register_smooth_quant_int_mm_pattern():
...
-            ),
-            Arg(),
-        ),
-        KeywordArg("out_shape_with_bias"),
+    pattern_with_bias_1_c1 = CallFunction(
+        aten.add.Tensor,
+        pattern_no_bias_1_c1,
+        KeywordArg("bias"),
+    )
+    pattern_with_bias_1_c2 = CallFunction(
```
Xia-Weiwen: Please use meaningful names rather than `c1` and `c2`.

torchao/quantization/pt2e/inductor_passes/x86.py
```diff
@@ -1846,7 +1889,7 @@ def _with_outer_reshape(pattern):
     )
 
     def _validate_pattern(match: Match):
-        if len(match.nodes) not in [4, 5, 6, 7, 10]:
+        if len(match.nodes) not in [4, 5, 7, 9]:
```
Xia-Weiwen: Please add a comment here as copilot commented for better readablity.

torchao/quantization/pt2e/inductor_passes/x86.py
```diff
@@ -1890,10 +1931,13 @@ def _int_mm_weight_prepack(match: Match, *args, **kwargs):
...
             x = kwargs["a"]
             weight = kwargs["b"]
-            dtype = kwargs["dtype"]
+            x_scale_dtype = kwargs["x_scale_dtype"]
             x_scale = kwargs["x_scale"]
             w_scale = kwargs["w_scale"]
             x_shape = x.meta.get("tensor_meta").shape
+            dtype = kwargs.get("dtype")
+            if dtype is None:
+                dtype = x_scale_dtype
```
```suggestion
            dtype = kwargs.get("dtype", x_scale_dtype)
```

Xia-Weiwen (commented):
Thanks for the PR. Please also add a UT.

torchao/kernel/intmm.py
```diff
@@ -127,7 +127,8 @@ def int_scaled_matmul(
     assert M == scales1.size(0) or scales1.numel() == 1
     assert 1 == scales1.size(1)
     assert scales1.is_contiguous()
-    scales1 = scales1.expand((M, N))
+    if scales1.device.type != "cpu":
```
jerryzh168: why is this change needed?

cyxlily: Broadcasting is automatic during scale multiplication, so the expand operation is unnecessary and adds overhead.

jerryzh168: I see, is this true for CUDA as well?

also is this optimized away by torch.compile?

Xia-Weiwen: This change is removed (moved to another PR https://github.com/pytorch/ao/pull/4103)

torchao/quantization/pt2e/inductor_passes/x86.py
```diff
@@ -1825,43 +1828,46 @@ def _with_outer_reshape(pattern):
...
     def _validate_pattern(match: Match):
-        if len(match.nodes) not in [4, 5, 6, 7, 10]:
+        # Valid node counts correspond to different pattern variations:
+        # 4: pattern1_with_no_outer_or_act_reshape (int_mm + convert + mul + mul)
+        # 6: pattern_no_bias_1 (reshape + int_mm + convert + mul + mul + reshape)
+        # 7: pattern_with_bias_1 (pattern_no_bias_1 + add)
+        # 8: pattern_no_bias_1_with_output_convert (pattern_no_bias_1 with dot scaled + output convert)
+        # 9: pattern_with_bias_1_with_output_convert (pattern_with_bias_1 with dot scaled + output convert)
+        if len(match.nodes) not in [4, 6, 7, 8, 9]:
             return False
```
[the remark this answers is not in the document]
Xia-Weiwen: @cyxlily Please take a look.

torchao/quantization/pt2e/inductor_passes/x86.py
```diff
@@ -1787,36 +1787,39 @@ def _register_smooth_quant_int_mm_pattern():
 
     # When torch.compile'ing with dynamic=True, the expand node and the two tailing reshape nodes exist
     # When torch.compile'ing with dynamic=False, they don't exist
```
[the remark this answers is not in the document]
Xia-Weiwen: @cyxlily Please take a look.

torchao/quantization/pt2e/inductor_passes/x86.py
```diff
@@ -1787,36 +1787,39 @@ def _register_smooth_quant_int_mm_pattern():
...
-    def get_pattern_no_bias(expand_a_scale: bool, reshape_a: bool = True):
-        return CallFunction(
+    def get_pattern_no_bias(reshape_a: bool = True, convert_a: bool = False):
+        int_mm_pattern = CallFunction(
```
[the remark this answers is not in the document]
Xia-Weiwen: @cyxlily Please consider renaming here.

test/quantization/pt2e/test_x86inductor_fusion.py
```diff
@@ -2907,6 +2889,260 @@ def matcher_check_fn():
         if test_for_pointwise_binary:
             self.assertEqual(counters["inductor"]["qlinear_binary_matcher_count"], 1)
 
+    @skipIfNoDynamoSupport
+    @skipIfNoONEDNN
+    @parametrize("dtype", [torch.float32, torch.bfloat16])
+    def test_da8w8_2d_hits_no_outer_or_act_reshape_pattern(self, dtype):
```
Xia-Weiwen: Can we merge these test cases to reduce the number of them? For example, we can use arguments to control bias, dynamic, etc.

torchao/quantization/pt2e/inductor_passes/x86.py
```diff
@@ -1780,43 +1780,52 @@ def _register_linear_dynamic_fp16_weight_prepack():
 def _register_smooth_quant_int_mm_pattern():
     """
     The pattern is:
-      (no bias) reshape -> _int_mm -> convert_element_type -> (expand ->) mul -> mul -> reshape
-    or
-      (with bias) pattern_no_bias -> add (-> reshape -> reshape)
+       (no bias)
+         (optional) reshape(a) -> _int_mm -> convert_element_type -> mul(x_scale)
+         (optional) convert_element_type (output-convert variant) -> mul(w_scale)
```
Xia-Weiwen: It looks a bit confusing. Can we use `[]` to mark an optional node instead of writing it explicitly? For example, `[reshape(a)] -> _int_mm`. And the `(output-convert variant)` comment is unnecessary.

torchao/quantization/pt2e/inductor_passes/x86.py
```diff
@@ -1825,43 +1834,51 @@ def _with_outer_reshape(pattern):
...
+    # The following pattern covers both torchao
+    # Int8DynamicActivationInt8WeightConfig linear and smooth-quant.
+    # It matches the Int8DynamicActivationInt8WeightConfig case where
+    # both activation and weights are symmetrically quantized.
+    # It also matches smooth-quant when the graph is lowered to the
+    # same int_mm + scaling form, especially for 2D input shapes.
+    # Since add is not currently being used as a oneDNN post-op, but
+    # is unfused, we don't need this pattern with bias.
+    # Ideally, we should add mul + add post-op support in ATen int8
+    # oneDNN linear op.
```
Xia-Weiwen: The comment is too long and is hard to read after formatting. Please simplify it and format it with bullets.

Xia-Weiwen: Just one sentence is enough: The following pattern covers both torchao DA8W8 and SmoothQuant.

cyxlily: Thanks for the comments, updated.

torchao/quantization/pt2e/inductor_passes/x86.py
cyxlily: Thanks for the comments, updated.

test/quantization/pt2e/test_x86inductor_fusion.py
cyxlily: Thanks for the comments, updated.

Xia-Weiwen:
@claude  review

Xia-Weiwen:
@claude review again

Xia-Weiwen:
@claude review again.

Xia-Weiwen:
@claude review again

Xia-Weiwen:
@claude review again

Xia-Weiwen:
@claude review again

Xia-Weiwen:
@claude review again

Xia-Weiwen:
@claude review again

Xia-Weiwen (approved):

torchao/quantization/quant_api.py
```diff
@@ -1053,6 +1055,7 @@ class Int8DynamicActivationInt8WeightConfig(AOBaseConfig):
         Union[Granularity, Tuple[Granularity, Granularity], list[Granularity]]
     ] = PerRow()
     set_inductor_config: bool = True
+    reduce_range: Optional[bool] = False
```
jerryzh168: should just be bool

torchao/quantization/quant_api.py
```diff
@@ -1078,6 +1081,9 @@ def __post_init__(self):
             "Please set it to MappingType.SYMMETRIC or "
             "MappingType.ASYMMETRIC."
         )
+        assert self.reduce_range in (True, False), (
```
jerryzh168: if this is only for x86, maybe do a check to make sure user is using cpu when the flag is set to True?

torchao/quantization/quant_api.py
```diff
@@ -1175,6 +1186,9 @@ def __post_init__(self):
...
```

jerryzh168:
seems reduce_range is not tested, I'd recommend split the reduce_range change and the fusion change in two separate PRs to make it clearer

torchao/quantization/quant_api.py
Xia-Weiwen: Thanks for reviewing. This is for compatibility purpose (per review comments from claude). If it is not optional, it may break backward compatibility if users saved quantized model with old torchao and now want to load it with the new torchao.

torchao/quantization/quant_api.py
Xia-Weiwen: If it's not on CPU, `reduce_range` still works so not necessary to check I think.

Xia-Weiwen:
> seems reduce_range is not tested, I'd recommend split the reduce_range change and the fusion change in two separate PRs to make it clearer

Hi @jerryzh168 Thanks for reviewing. We add this flag because some or all of the CI machines do not have AVX512_VNNI and the test cases would fail. So, we only test `reduce_range=True` here. This is also the case in pytorch, where we simply skip correctness test of int8 matmul or use small int8 values intentionally to avoid overflow.
A better way to handle this may be that we check AVX512_VNNI availability and set `reduce_range` accordingly. Though `reduce_range=True` will still not be tested in CI but code looks better and it can be tested in our own environments. How does that sound to you?

cyxlily:
> seems reduce_range is not tested, I'd recommend split the reduce_range change and the fusion change in two separate PRs to make it clearer

Thanks for comments, split the reduce_range to the PR https://github.com/pytorch/ao/pull/4266

jerryzh168 (approved):
@Xia-Weiwen please feel free to be the codeowner for cpu only changes

you can add yourself to https://github.com/pytorch/ao/blob/main/CODEOWNERS as well for relevant files

torchao/quantization/quant_api.py
jerryzh168: probably better to add the check if we don't expect other backend to use it, this will help us decrease the user surface and make it less error prone for users I think

jerryzh168:
> Hi @jerryzh168 Thanks for reviewing. We add this flag because some or all of the CI machines do not have AVX512_VNNI and the test cases would fail. So, we only test `reduce_range=True` here. This is also the case in pytorch, where we simply skip correctness test of int8 matmul or use small int8 values intentionally to avoid overflow.

can this be set automatically when we are in CPU + not in `AVX512_VNNI` machine? ideally we make the user's life easier

maybe one question here is do we always prepare and serve the quantized model in the same type of machine or not

cyxlily:
> > Hi @jerryzh168 Thanks for reviewing. We add this flag because some or all of the CI machines do not have AVX512_VNNI and the test cases would fail. So, we only test `reduce_range=True` here. This is also the case in pytorch, where we simply skip correctness test of int8 matmul or use small int8 values intentionally to avoid overflow.
> 
> can this be set automatically when we are in CPU + not in `AVX512_VNNI` machine? ideally we make the user's life easier
> 
> maybe one question here is do we always prepare and serve the quantized model in the same type of machine or not

Thanks for comments, I removed the manual reduce_range setting and set it automatically in int8 tensor.

Xia-Weiwen:
@claude Review again

PR #3795, opened 2026-02-02
mxfp4 and nvfp4: add test for packing shape conversions and fix bug

vkuzo:
Summary:

1. add test coverage for utility functions that convert between packed
   and unpacked shapes
2. fix bug uncovered by (1)

This will be useful as we add support for `aten.expand` further up the
stack.

Test Plan:

```
pytest test/prototype/mx_formats -s -x
```

PR #3797, opened 2026-02-02
make 4 bit shape calculation functions more generic

vkuzo:
Summary:

1. make the util function that gets the 4-bit packed shape from original shape
   more generic by using strides
2. delete the unused `view` op handling and the now-unused function to
   go from 4-bit packed shape to high precision shape

Test Plan:

```bash
time pytest test/prototype/mx_formats/ -s -x
```


vkuzo:
* #3798
* __->__ #3797
* #3795

PR #3801, opened 2026-02-03
[CPU] support scaled embedding bag pattern match for FP8 output

yanbing-j:
This PR is to add pattern match support for scaled embedding bag with FP8 output.

### DIFF
--- a/torchao/quantization/pt2e/inductor_passes/x86.py
+++ b/torchao/quantization/pt2e/inductor_passes/x86.py
@@ -53,20 +53,11 @@ _VIEW_METHOD_OPS = [
 ]
 
 # Dynamo/FX sometimes carries ScalarType as an int enum (e.g. 24 for float8_e4m3fn).
-# For readability and comparisons, normalize back to torch.dtype when possible using _fallback_enum_to_dtype table.
+# For readability and comparisons, normalize back to torch.dtype when possible using _fallback_enum_to_dtype table,
+# which is from https://github.com/pytorch/pytorch/blob/main/torch/headeronly/core/ScalarType.h#L103-L149.
 _fallback_enum_to_dtype: dict[int, torch.dtype] = {
-    0: torch.uint8,
-    1: torch.int8,
-    2: torch.int16,
-    3: torch.int32,
-    4: torch.int64,
-    5: torch.float16,
     6: torch.float,
-    7: torch.double,
-    11: torch.bool,
-    15: torch.bfloat16,
     24: torch.float8_e4m3fn,
-    25: torch.float8_e5m2,
 }
 
 """


torchao/quantization/pt2e/inductor_passes/x86.py
```diff
@@ -2946,10 +2952,66 @@ def scaled_embedding_bag(match: Match, *args, **kwargs):
...
+            normalized_o_dtype = _normalize_dtype(kwargs["o_dtype"])
+            output_type = normalized_o_dtype
+
+            def _extract_const_float(val) -> float | None:
+                # Prefer extracting from python scalars and FX node structure
+                if isinstance(val, (int, float, bool)):
+                    return float(val)
+                if isinstance(val, torch.fx.Node):
+                    meta_val = val.meta.get("val", None)
+                    if isinstance(meta_val, (int, float, bool)):
```
LevelDownRefine:
```suggestion
                    if isinstance(meta_val, (int, float)):
```

torchao/quantization/pt2e/inductor_passes/x86.py
```diff
@@ -2946,10 +2952,66 @@ def scaled_embedding_bag(match: Match, *args, **kwargs):
...
+                        return float(meta_val)
+                    # Common pattern: aten.full([1], fill_value, dtype=float)
+                    if val.target is torch.ops.aten.full.default and len(val.args) >= 2:
+                        fill_value = val.args[1]
+                        if isinstance(fill_value, (int, float, bool)):
```
LevelDownRefine:
```suggestion
                        if isinstance(fill_value, (int, float)):
```

test/quantization/pt2e/test_x86inductor_fusion.py
```diff
@@ -199,7 +199,15 @@ def _test_common(
                 clone_inputs = self._clone_inputs(inputs)
                 expected = mod(*inputs)
                 actual = torch.compile(mod, **compile_options)(*clone_inputs)
-                torch.testing.assert_close(actual, expected, atol=atol, rtol=rtol)
+                if inputs[0].dtype == torch.float8_e4m3fn:
```
Xia-Weiwen: Looks like we can always call `.float()` to avoid `if ... else`

torchao/quantization/pt2e/inductor_passes/x86.py
```diff
@@ -2946,10 +2952,66 @@ def scaled_embedding_bag(match: Match, *args, **kwargs):
...
-            o_scale = kwargs["o_inv_scale"]
+
+            def _normalize_dtype(dtype_or_enum: Any) -> torch.dtype | Any:
+                # Dynamo/FX sometimes carries ScalarType as an int enum (e.g. 24 for float8_e4m3fn).
+                # For readability and comparisons, normalize back to torch.dtype when possible.
+                if isinstance(dtype_or_enum, torch.dtype):
+                    return dtype_or_enum
+                if isinstance(dtype_or_enum, int):
+                    # Fallback mapping for common ScalarType enums.
+                    _fallback_enum_to_dtype: dict[int, torch.dtype] = {
```
Xia-Weiwen: Where can we find this mapping? Can we import something from pytorch to avoid defining this table here and lower the risk of potential change in the future?
If this table must be explicit defined in torchao, can we put it somewhere more common?

torchao/quantization/pt2e/inductor_passes/x86.py
```diff
@@ -2946,10 +2952,66 @@ def scaled_embedding_bag(match: Match, *args, **kwargs):
...
+                            and len(data) == 1
+                            and isinstance(data[0], (int, float))
+                        ):
+                            return float(data[0])
+                return None
+
+            o_scale_maybe = _extract_const_float(kwargs["o_inv_scale"])
+            if o_scale_maybe is None:
+                assert False, "Output scale is not a constant float."
+            o_scale = o_scale_maybe
```
```suggestion
            o_scale = _extract_const_float(kwargs["o_inv_scale"])
            assert o_scale is not None, "Output scale is not a constant float."
```

test/quantization/pt2e/test_x86inductor_fusion.py
yanbing-j: This is in `_test_common`, and may include other dtypes, like int... Can we always call `.float()` here?

torchao/quantization/pt2e/inductor_passes/x86.py
yanbing-j: The mapping can be found in https://github.com/pytorch/pytorch/blob/main/torch/headeronly/core/ScalarType.h#L103-L149. Need more investigation to import directly from pytorch. I can add TODO here.
Do you have any suggestions to put this table? I suppose it is a general requirement.

test/quantization/pt2e/test_x86inductor_fusion.py
Xia-Weiwen: I think so. It looks of low risk.

torchao/quantization/pt2e/inductor_passes/x86.py
Xia-Weiwen: I see. Since it's common, maybe it's better to put it outside the registration and have a comment explain this.

torchao/quantization/pt2e/inductor_passes/x86.py
```diff
@@ -2946,10 +2960,50 @@ def scaled_embedding_bag(match: Match, *args, **kwargs):
...
+
+            def _normalize_dtype(dtype_or_enum: Any) -> torch.dtype | Any:
+                # Dynamo/FX sometimes carries ScalarType as an int enum (e.g. 24 for float8_e4m3fn).
+                # For readability and comparisons, normalize back to torch.dtype when possible.
+                if isinstance(dtype_or_enum, torch.dtype):
+                    return dtype_or_enum
+                if isinstance(dtype_or_enum, int):
+                    # Fallback mapping for common ScalarType enums.
+                    if dtype_or_enum in _fallback_enum_to_dtype:
+                        return _fallback_enum_to_dtype[dtype_or_enum]
```
yanbing-j, line 2976: Need add assert here to filter dtypes except for float and e4m3. When encounter other dtytpes in the future, I will update accordingly.

yanbing-j:
Hi @jerryzh168 , could you please help review this PR? The CI failure is not related to PR code change.

PR #3807, opened 2026-02-03
LowBitPacking kernels on x86

JacobSzwejbka:
Summary: Buckify these AO packing kernels, add fixes to make them feature complete, and modify masking behavior to make them match arm kernel behavior. The masking is sort of weird what happens iiuc is that Arm kernels assume all values in range so it doesnt bother masking say the high bits of an input value of 4 bit quant. The cpp kernels previously masked the bits which is also sort of incorrect since it would be changing the input value (which was an invalid input anyway since it was out of range). I went ahead and removed the masking since it makes it easier to verify the kernels are correct.

Differential Revision: D91728275


metascroy:
Looks good!

I think you need to run the ao linter (ruff).

metascroy (approved):

PR #3816, opened 2026-02-04
fix torchao circular imports for BUCK

vkuzo:
Summary:

TL;DR; is:
1. the torchao import chain has gotten messy
2. https://github.com/pytorch/ao/pull/3676 broke some internal builds introducing circular import errors. I don't have a clean repro, but from the traces the errors all step from the fact that `torchao.dtype` and `torchao.quantization` import from each other in various places.

This PR attempts to eliminate (2) by ensuring that `torchao.dtypes` is always initialized first, and any imports from `torchao.dtypes` into `torchao.quantization` are done lazily. I did it this way because `torchao.dtypes` will be removed from the codebase soon.

WIP

TODO fill me out

Test Plan: CI, needs validation internally


torchao/dtypes/uintx/int4_xpu_layout.py
```diff
@@ -39,6 +38,8 @@ def _aqt_is_xpu_layout_uint4(aqt):
 
 
 def _linear_bf16_act_uint4_weight_float_zero_check(input_tensor, weight_tensor, bias):
+    from torchao.quantization.quant_primitives import ZeroPointDomain
```
andrewor14, line 41: add some comments about why we're importing it here? Otherwise I can imagine the next person coming in trying to "clean up" the code and introducing this problem again

vkuzo: this is enforced by the new `torchao/__init__.py` structure, where `dtypes` is imported before `quantization`.  Now if someone imports quantization from dtypes directly, `import torchao` will fail with a circular import error.

PR #3826, opened 2026-02-05
[Reland][X86] Re-enable some inductor-related test cases (2)

Xia-Weiwen:
Reland again since https://github.com/pytorch/ao/pull/3777 is reverted.


test/quantization/pt2e/test_x86inductor_cpp_wrapper.py
```diff
@@ -0,0 +1,295 @@
...
+        slow: bool = False
+        func_inputs: list = None
+        code_string_count: dict = {}
+        test_build_separate: bool = False
+
+    for item in [
+        BaseTest(
+            "test_qconv2d",
+            "cpu",
+            test_x86inductor_fusion.TestPatternMatcher(),
```
jerryzh168: it's probably better to put the common things in torchao.testing.pt2e instead

Xia-Weiwen: Sorry, I don't think it's common. Can you please instruct what to do? Thanks.

jerryzh168: seems these are the tests for x86_inductor, wondering why do you need 2 files? can you just merge test/quantization/pt2e/test_x86inductor_cpp_wrapper.py into test_x86_inductor_fusion.py?

Xia-Weiwen: Ok. That should work. I will do that. Thanks.

Xia-Weiwen: Hi @jerryzh168 I have updated this part. Would you mind importing it to your inner workflow to avoid any issues? Thanks.

andrewor14:
Test changes look good to me, will let @jerryzh168 do the final stamp.

I did try to import this into our internal workflows, but it's complaining that the PR was not created using ghstack so I could not import it in the end, so not sure if there's a way to test it other than just merging

test/quantization/pt2e/test_x86inductor_fusion.py
```diff
@@ -3472,6 +3277,247 @@ def matcher_check_fn():
...
+
+
+RUN_CPU = HAS_CPU and not IS_MACOS and torch.version.hip is None
+
+if RUN_CPU and torch_version_at_least("2.8.0"):
+
+    class BaseTest(NamedTuple):
+        name: str
+        device: str = "cpu"
+        tests: TestCase = TestCase()
```
jerryzh168, line 3412: nit: naming seems confusing? should this just be `test_case`?

test/quantization/pt2e/test_x86inductor_fusion.py
```diff
@@ -3472,6 +3277,247 @@ def matcher_check_fn():
...
+            item.device,
+            item.tests,
+            item.condition,
+            item.slow,
+            item.func_inputs,
+            item.code_string_count,
+            item.test_build_separate,
+        )
+
+    copy_tests(
```
jerryzh168, line 3503: seems a bit complicated, I think ideally these can be simplified, not sure if it's possible.

jerryzh168 (approved):
I feel ideally the tests can be written in a more explicitly way (i.e. no util functions to generate test case etc.)

but also this part is relatively independent and I assume maintained by intel, I think it's fine to land as well.

Xia-Weiwen:
Hi @jerryzh168 Thanks for reviewing. Since we are just moving these tests from torch to torchao, I didn't think carefully about how to simplify them and I am not sure if it's worth the effort. Intuitively, the tests for CPP wrapper are based on those for Python wrapper, so it makes sense that we create CPP wrapper cases based on Python wrapper cases. Please let me know if there is a straightforward way to simplify it that comes to you. Thanks.

Xia-Weiwen:
> Test changes look good to me, will let @jerryzh168 do the final stamp.
> 
> I did try to import this into our internal workflows, but it's complaining that the PR was not created using ghstack so I could not import it in the end, so not sure if there's a way to test it other than just merging

@andrewor14 Thanks. I will merge it later. If it breaks something internal again, please let me know.

Xia-Weiwen:
Hi @jerryzh168 I am going to merge this if you don't mind. Thanks.

jerryzh168:
@Xia-Weiwen I can try to import this to fbcode first if you are worried about this breaking fbcode CI

Xia-Weiwen:
> @Xia-Weiwen I can try to import this to fbcode first if you are worried about this breaking fbcode CI

Thanks a lot

Xia-Weiwen:
Hi @jerryzh168 How is the import going? Thanks.

jerryzh168:
@Xia-Weiwen seems OK, I think we can merge

Xia-Weiwen:
> @Xia-Weiwen seems OK, I think we can merge

Great. Thanks.

PR #3827, opened 2026-02-05
Add configurable scale dtype for `Int8Tensor`

### LINKED ISSUE #3819: How to set Int8Tensor scale dtype?
Hi, 

I'm trying to quantize a model with compute type bfloat16 and storage type Int8Tensor but precision is suboptimal as Int8Tensor.scale is in hp tensor's dtype. bfloat16 has a very small mantissa so eps is 0.0078125 which is too large to give any happy precision - lots of good bits left unused but using up space anyway. Is there a way to specify the scale dtype as float16 or even float32 without destroying performance? Performance is real good with compilation!

Any thoughts?

namgyu-youn:
**Summary**:
`Int8Tensor` used BF16 as a hard-coded scale dtype. BF16 has better range than FP16, but poor precision with only 7 mantissa bits, causing quantization accuracy loss. This PR adds configurable scale dtype to address this issue.
- Fixes #3819

**Test plan**:
pytest -sv test/quantization/quantize_/workflows/int8/test_int8_tensor.py

### DIFF
--- a/test/quantization/quantize_/workflows/int8/test_int8_tensor.py
+++ b/test/quantization/quantize_/workflows/int8/test_int8_tensor.py
@@ -418,11 +418,12 @@ class TestInt8StaticQuant(TorchAOIntegrationTestCase):
         self.assertGreater(sqnr, 40, f"SQNR too low: {sqnr}")
 
     @common_utils.parametrize("scale_dtype", [torch.float16, torch.bfloat16])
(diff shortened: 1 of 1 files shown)


jerryzh168:
@namgyu-youn since the issue is resolved now: https://github.com/pytorch/ao/issues/3819#issuecomment-3860235172 I think we don't want to land this for now to keep the feature simpler

PR #3829, opened 2026-02-05
Add HQQ option in UIntxWeightOnlyConfig

mergennachin:
We'll be leaving accuracy on the top. Seems like it's an easy thing to do in general to run HQQ.

```
USE_CPP=1 TORCHAO_BUILD_EXPERIMENTAL_MPS=1 pip install . --no-build-isolation
python -m pytest torchao/experimental/ops/mps/test/test_quantizer.py
```

### DIFF
--- a/torchao/experimental/quant_api.py
+++ b/torchao/experimental/quant_api.py
@@ -7,7 +7,8 @@
 import logging
 import sys
 from dataclasses import dataclass
-from typing import Optional
+from enum import Enum
+from typing import Optional, Union
 
 import torch
 import torch.nn as nn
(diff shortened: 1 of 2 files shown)


torchao/experimental/quant_api.py
```diff
@@ -87,14 +111,43 @@ def __init__(
         else:
             self.register_parameter("bias", None)
 
-    def quantize_and_pack_weights(self, weights, nbit, group_size):
+    def quantize_and_pack_weights(
```
metascroy, line 114: This should probably be migrated to quantize_, but that's a separate issue. cc @manuelcandales

metascroy (approved):

PR #3832, opened 2026-02-05
[bug][pt2e] Fix Python 3.14 compatibility for EdgeOrNode type alias

oelachqar:
## Summary

Python 3.14 made `typing.Union` attributes read-only . This causes an `AttributeError` when importing the quantizer module:

```
AttributeError: 'typing.Union' object has no attribute '__module__'
```

This PR guards the `EdgeOrNode.__module__` assignment with a version check to maintain compatibility across Python versions.

## Reproduction

```bash
# Create a fresh Python 3.14 environment
uv venv --python 3.14 /tmp/test-py314
source /tmp/test-py314/bin/activate

# Install torchao (or PyTorch with torch.ao)
uv pip install torch torchao

# This will fail on Python 3.14 without the fix:
python -c "from torchao.quantization.pt2e.quantizer.quantizer import EdgeOrNode"
```

**Expected error (before fix):**
```
  File "<string>", line 1, in <module>
  File ".../torchao/quantization/pt2e/quantizer/quantizer.py", line 96, in <module>
    EdgeOrNode.__module__ = "torchao.quantization.pt2e.quantizer.quantizer"
AttributeError: 'typing.Union' object has no attribute '__module__'
```

## Details

The `__module__` assignment is cosmetic - it improves how the type alias appears in error messages and documentation. It is not required for functionality, so skipping it on Python 3.14+ seems safe.

## Test Plan

- Verified import works on Python 3.14

### DIFF
--- a/torchao/quantization/pt2e/quantizer/quantizer.py
+++ b/torchao/quantization/pt2e/quantizer/quantizer.py
@@ -5,6 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 # mypy: allow-untyped-defs
+import sys
 from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
 from typing import Callable, Optional, Union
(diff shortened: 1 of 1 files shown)


jerryzh168 (approved):
seems OK, thanks

PR #3834, opened 2026-02-06
[ROCm][INT4] Configurable ntile size for TilePacked format

ZhiweiYan-96:
# Motivation 
Fix a UT failure 
```
pytest -sv test/integration/test_integration.py -k test_int4_weight_only_quant_subclass_api_grouped_5
``` 
The failed case is with shape (m, k, n)=(256, 256,8). The n dimension is smaller than the Matrix Core nTileSize=16 on AMD, while reasonable for Nv TensorCore with nTileSize=8

According to the code at https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/cuda/int4mm.cu#L1116
```
auto nTiles = (B.size(0) / nTileScaleFactor);
```
We can infer that 
1. n/8 > 16, where 16=nTileScaleFactor*nTileSizeTensor otherwise, nTiles would be 0.(This is bug!)
2. n/8 must be a mulitple of 8, otherwise, there would be fractional number of tiles.

This PR fix it by using a proper padding size when calling `find_mmultiple` utils.

# Testing

```
pytest -sv test/integration/test_integration.py -k test_int4_weight_only_quant_subclass_api_grouped_5
```

### DIFF
--- a/torchao/quantization/quant_api.py
+++ b/torchao/quantization/quant_api.py
@@ -768,8 +768,8 @@ class Int4WeightOnlyConfig(AOBaseConfig):
     int4_choose_qparams_algorithm: Int4ChooseQParamsAlgorithm = (
         Int4ChooseQParamsAlgorithm.TINYGEMM
     )
-    version: int = 2
     int4_tile_packed_ntile: int = 8
+    version: int = 2
 
     def __post_init__(self):
         assert self.int4_tile_packed_ntile in [8, 16], (
--- a/torchao/quantization/quantize_/workflows/int4/int4_tile_packed_to_4d_tensor.py
+++ b/torchao/quantization/quantize_/workflows/int4/int4_tile_packed_to_4d_tensor.py
@@ -98,8 +98,8 @@ class Int4TilePackedTo4dTensor(TorchAOBaseTensor):
         cls,
         hp_tensor: torch.Tensor,
         block_size: List[int],
-        ntile_size: Optional[int] = 8,
         int4_choose_qparams_algorithm: Int4ChooseQParamsAlgorithm = Int4ChooseQParamsAlgorithm.TINYGEMM,
+        ntile_size: Optional[int] = 8,
     ):
         assert len(block_size) == hp_tensor.ndim, (
             f"Expecting the length of block_size to be equal to the dimension of the weight, got {block_size=} and {hp_tensor.ndim=}"


ZhiweiYan-96:
@XiaobingSuper Mind take a look？

torchao/quantization/quantize_/workflows/int4/int4_tile_packed_to_4d_tensor.py
```diff
@@ -127,7 +127,8 @@ def from_hp(
 
         # Pre-process: pad to required dimensions
         in_features = find_multiple(orig_in_features, 1024)
-        out_features = find_multiple(orig_out_features, 8)
+        n_tile = 16 if orig_out_features < 16 and torch.version.hip else 8
```
XiaobingSuper: need to` n_tile = 16 if torch.version.hip else 8`? find_multiple will do a padding according to the given tile size.

ZhiweiYan-96: That's right.

ZhiweiYan-96:
hi, @petrex @jithunnair-amd Could you please take a look for reviewing this PR and add `ciflow/rocm` label for testing? Thanks.

XiaobingSuper:
@jerryzh168 could you help review it? Thanks!

ZhiweiYan-96:
The two failures are irrelevant to this PR.

```
test/prototype/test_parq.py::TestTorchAoConfigIntegration::test_tied_weights_quantization - AttributeError: 'list' object has no attribute 'keys'

 torch.distributed.DistNetworkError: The server socket has failed to listen on any local network address. port: 29500, useIpv6: false, code: -98, name: EADDRINUSE, message: address already in use
```

vkuzo:
hi @ZhiweiYan-96 , looks reasonable overall.  IMO we should make this user configurable instead of automatically selected, it's confusing when a packing format behaves differently based on the environment.  Can we add this to the config instead of selecting it automatically?  You can make the config clearly state which value the user needs to select on ROCm.

ZhiweiYan-96:
Thanks, @vkuzo ,  we can make n_tile_size as a configurable attributes in `Int4WeightOnlyConfig`, which defaults to 8 (nv) and    16 for the ROCm users.

torchao/quantization/quant_api.py
XiaobingSuper, line 755: add int4_tile_packed_ntile description here? also note that int4_tile_packed_ntile only works Int4PackingFormat.TILE_PACKED_TO_4D case.

torchao/quantization/quant_api.py
```diff
@@ -767,6 +767,8 @@ class Int4WeightOnlyConfig(AOBaseConfig):
     int4_choose_qparams_algorithm: Int4ChooseQParamsAlgorithm = (
         Int4ChooseQParamsAlgorithm.TINYGEMM
     )
+    # ntile size for TILE_PACKED_TO_4D format, 8 for CUDA platform, 16 for ROCm platform
+    int4_tile_packed_ntile: int = 8
```
XiaobingSuper: Do we need to add a check to ensure it only supports a limited set of values?

torchao/quantization/quant_api.py
ZhiweiYan-96: nice catch, moved

torchao/quantization/quant_api.py
ZhiweiYan-96: yes, added. https://github.com/pytorch/ao/pull/3834/changes/b390fec93994d91276a5d03b48d78c3bf6c71a5a..470b7ccca937f0b7beca642836666b7f5db3adde#diff-bf4d50867e3d649de2d89146592bf47d2f258c4c19126c8acf0e120ee904b726R775

torchao/quantization/quantize_/workflows/int4/int4_tile_packed_to_4d_tensor.py
```diff
@@ -98,6 +98,7 @@ def from_hp(
         cls,
         hp_tensor: torch.Tensor,
         block_size: List[int],
+        ntile_size: Optional[int] = 8,
```
vkuzo: nit: maybe make this argument last, to avoid breaking any existing callsites that specify arguments positionally

vkuzo:
looks good! can we just make the new argument last, and after that if CI is green lgtm!

torchao/quantization/quantize_/workflows/int4/int4_tile_packed_to_4d_tensor.py
ZhiweiYan-96: thanks @vkuzo ! Move the argument to the last one https://github.com/pytorch/ao/pull/3834/changes#diff-632459893f0de165d48a87a900c6535f5b392520b47fffeddaa2175f5904b6baR102. The CI is green now

XiaobingSuper (approved):

PR #3835, opened 2026-02-06
[X86] Fix incorrect weight scale in fp8 qconv

Yuxingwang-intel:
Because the same buffer name was reused, the later-registered convolution overwrote the previous one, causing multiple convolutions to incorrectly share the same weight scale, which led to accuracy issues in FP8 qconv.

### DIFF
--- a/torchao/quantization/pt2e/inductor_passes/x86.py
+++ b/torchao/quantization/pt2e/inductor_passes/x86.py
@@ -795,7 +795,7 @@ def _register_qconv_weight_prepack_pass(
             )
             with torch.utils._python_dispatch._disable_current_modes():
                 w_scale_tensor = torch.tensor([w_scale.args[1]])
-            scale_name = f"{conv_node}"
+            scale_name = f"{conv_node}_scale"
             match.graph.owning_module.register_buffer(scale_name, w_scale_tensor)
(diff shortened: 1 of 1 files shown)


Yuxingwang-intel:
@Xia-Weiwen Could you please help me review this PR. Thanks!

torchao/quantization/pt2e/inductor_passes/x86.py
```diff
@@ -795,8 +795,9 @@ def qconv_weight_prepack(match: Match, *args, **kwargs):
             )
             with torch.utils._python_dispatch._disable_current_modes():
                 w_scale_tensor = torch.tensor([w_scale.args[1]])
-            match.graph.owning_module.register_buffer("w_scale", w_scale_tensor)
-            w_scale = match.graph.create_node("get_attr", "w_scale")
+            scale_name = f"{conv_node}"
```
```suggestion
            scale_name = f"{conv_node}_scale"
```
seems better?

Xia-Weiwen (approved):

Xia-Weiwen:
Can you add a test case for it?

Xia-Weiwen:
The CI failure is irrelevant to this PR. And we plan to enable correctness test for quantization path in another PR since it is clearer.

PR #3839, opened 2026-02-06
Rename is_bias to has_bias in SmoothQuantObservedLinear module

jerryzh168:
att, to make it clearer

### DIFF
--- a/torchao/prototype/smoothquant/core.py
+++ b/torchao/prototype/smoothquant/core.py
@@ -77,12 +77,12 @@ class SmoothQuantObservedLinear(torch.nn.Linear):
         in_features: int,
         out_features: int,
         obs: SmoothQuantObserver,
-        is_bias: bool = False,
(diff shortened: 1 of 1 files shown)


torchao/prototype/smoothquant/core.py
```diff
@@ -77,12 +77,12 @@ def __init__(
...
         out_features: int,
         obs: SmoothQuantObserver,
-        is_bias: bool = False,
+        has_bias: bool = False,
         device=None,
         dtype=None,
     ):
         super().__init__(
-            in_features, out_features, bias=is_bias, device=device, dtype=dtype
+            in_features, out_features, bias=has_bias, device=device, dtype=dtype
```
jerryzh168, line 85: initially I thought maybe we can remove this since we set bias tensor later anyways, but then I think this is still useful for us to get the default bias registration code from parent

PR #3845, opened 2026-02-09
[X86] Update the call path to access CPUInfo properties

Yuxingwang-intel:
Update the call path to access CPUInfo properties following https://github.com/pytorch/pytorch/pull/173433

### DIFF
--- a/torchao/csrc/cpu/aten_kernels/quantized_sdpa.cpp
+++ b/torchao/csrc/cpu/aten_kernels/quantized_sdpa.cpp
@@ -22,6 +22,9 @@
 #include <omp.h>
 #include <torch/all.h>
 #include <torch/csrc/autograd/function.h>
+#include <iostream>
+#include <fstream>
+#include <string>
 
 namespace torchao {
 
@@ -878,7 +881,7 @@ inline void _fp8_dequant_quant_fusion_kernel(
   }
 }
 
-// UINT8 - one parallel loop with u8u8s32 GEMM 
+// UINT8 - one parallel loop with u8u8s32 GEMM
 template <typename scalar_t, typename mask_t,
           int64_t q_split_size, int64_t kv_split_size,
           bool use_one_parallel_loop,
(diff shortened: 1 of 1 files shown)


Xia-Weiwen:
CC @mingfeima

Xia-Weiwen (approved):

mingfeima (approved):

Yuxingwang-intel:
> The `l2_cache_size` here also seems unavailable with new API: https://github.com/pytorch/ao/blob/main/torchao/csrc/cpu/aten_kernels/quantized_sdpa.cpp#L2221. As lacking of the CI for CPP kernels, could you please locally build the repo with `USE_CPU_KERNELS=1` and ensure `test/test_ops.py` and `test/prototype/inductor/test_qsdpa_fusion.py` to pass?

The update has been applied. The test/test_ops.py and test/prototype/inductor/test_qsdpa_fusion.py have been verified to pass locally.

Xia-Weiwen:
Hi @mingfeima There are new changes in `torchao/csrc/cpu/aten_kernels/quantized_sdpa.cpp`. Can you please review the workaround to get L2 cache size? Thanks.

mingfeima (approved):

Xia-Weiwen:
Hi @jerryzh168 @janeyx99 @metascroy @drisspg Can you please review this, especially the changes to `setup.py`? Thanks.

jerryzh168:
@Xia-Weiwen btw we are updating release notes to per module highlights: https://github.com/pytorch/ao/issues/3889

wondering which flow this PR belongs to? please add correct labels

Xia-Weiwen:
> setup.py changes looks OK, although it's getting more complicated and we should probably think about how to make it clearer in the future

Thanks for reviewing. Perhaps we can move x86-related content to a separate file. Not sure if that would be a good idea. If so, we can work on that later.

Xia-Weiwen:
> @Xia-Weiwen btw we are updating release notes to per module highlights: #3889
> 
> wondering which flow this PR belongs to? please add correct labels

Thanks for the reminder. The `not user facing` label should be good enough.

jerryzh168:
> > setup.py changes looks OK, although it's getting more complicated and we should probably think about how to make it clearer in the future
> 
> Thanks for reviewing. Perhaps we can move x86-related content to a separate file. Not sure if that would be a good idea. If so, we can work on that later.

maybe structuring them into a single function in the same file might be good enough, if that's possible. we also have to cleanup for other backends as well, right now seems quite complicated

Xia-Weiwen:
> > > setup.py changes looks OK, although it's getting more complicated and we should probably think about how to make it clearer in the future
> > 
> > 
> > Thanks for reviewing. Perhaps we can move x86-related content to a separate file. Not sure if that would be a good idea. If so, we can work on that later.
> 
> maybe structuring them into a single function in the same file might be good enough, if that's possible. we also have to cleanup for other backends as well, right now seems quite complicated

Good idea. Thanks.

PR #3847, opened 2026-02-10
Add mixed granularity tests for smooth quant

cyxlily:
Split from https://github.com/pytorch/ao/pull/3524
Test different activation and weight granularities for smooth quantization.


Xia-Weiwen (approved):

test/prototype/test_smoothquant.py
```diff
@@ -84,8 +84,19 @@ def setUpClass(cls):
         "base_config",
         [
             Int8DynamicActivationInt8WeightConfig(version=2),
+            Int8DynamicActivationInt8WeightConfig(version=2, granularity=PerRow()),
```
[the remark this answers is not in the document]
Xia-Weiwen: Makes sense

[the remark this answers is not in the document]
Xia-Weiwen: @cyxlily Please take a look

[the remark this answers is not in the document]
cyxlily: Thanks for the comment, updated.

jerryzh168 (approved):

test/prototype/test_smoothquant.py
```diff
@@ -88,8 +88,17 @@ def setUpClass(cls):
         "base_config",
         [
             Int8DynamicActivationInt8WeightConfig(version=2),
-            Int8StaticActivationInt8WeightConfig(granularity=PerRow()),
+            Int8DynamicActivationInt8WeightConfig(version=2, granularity=PerTensor()),
```
jerryzh168: nit: version=2 is the default now, feel free to remove

test/prototype/test_smoothquant.py
```diff
@@ -88,8 +88,17 @@ def setUpClass(cls):
...
+            Int8DynamicActivationInt8WeightConfig(
+                version=2, granularity=[PerRow(), PerTensor()]
```
jerryzh168: actually why is this a list? 

I saw type is changed to https://github.com/pytorch/ao/blob/9058b58a4f0ac5ea3c0fca4cc8f4f225d79f7137/torchao/quantization/quant_api.py#L856

I feel we should not have both tuple and list, can you have a followup PR to change this to `Union[Granularity, Tuple[Granularity, Granularity]]` to reduce the complexity here?

cyxlily: Thanks for comments, updated the UT.
For the followup PR, the JSON encoder explicitly raises NotImplementedError for tuples: https://github.com/pytorch/ao/blob/main/torchao/core/config.py#L146
So even though users can pass a Tuple, internally it still needs to be stored as a list for serialization.
Does it still make sense to change the API from list to tuple?

test/prototype/test_smoothquant.py
cyxlily: Thanks for comments, updated.

test/prototype/test_smoothquant.py
jerryzh168: oh I see, in that case we should remove tuple and just use List, I've seen the json error before as well

test/prototype/test_smoothquant.py
```diff
@@ -87,9 +87,14 @@ def setUpClass(cls):
     @common_utils.parametrize(
         "base_config",
         [
-            Int8DynamicActivationInt8WeightConfig(version=2),
-            Int8StaticActivationInt8WeightConfig(granularity=PerRow()),
+            Int8DynamicActivationInt8WeightConfig(),
+            Int8DynamicActivationInt8WeightConfig(granularity=PerTensor()),
+            Int8DynamicActivationInt8WeightConfig(granularity=(PerRow(), PerTensor())),
```
jerryzh168: let's just use list and drop tuple in the next PR

cyxlily: OK, created the PR to remove tuple: https://github.com/pytorch/ao/pull/4347

PR #3850, opened 2026-02-10
fix test_bf16_stochastic_round_dtensor

vkuzo:
Summary:

This test started failing on recent versions of PyTorch nightly, the
test cases passed individually but failed when run in sequence.
What I think was broken is that the ports were not released by the OS
fast enough between the tests. I did not debug what specifically
changed in PyTorch, fixing defensively by using a FileStore instead of
kTCPStore, which does not use any ports.

Test Plan:

```
pytest test/test_low_bit_optim.py -k test_bf16_stochastic_round_dtensor
```


vkuzo:
* __->__ #3850

PR #3857, opened 2026-02-11
Added new API for low precision fp8 attention using FA3

howardzhang-cv:
* #3947
* #3960
* #3930
* #3929
* #3865
* __->__ #3857
* #3959

 ---                                                                                                                                                                                                                                               
## Summary                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          
  - Added RoPE fusion compile path for FA3 FP8 low-precision attention (fuse_rope=True)                                                                                                                                                             
  - New elementary block: fp8_fa3_rope_sdpa — fused RoPE + FP8 quantization + low-precision SDPA
  - New Triton kernel for fused RoPE + QKV quantization with layout transpose ([B,S,H,D] → [B,H,S,D])
  - RoPE fusion method: Custom Inductor backend that traces the FX graph, detects RoPE + SDPA patterns (NeoX half-split and FLUX interleaved formats), and replaces them with fp8_fa3_rope_sdpa custom ops. Falls back to fp8_fa3_sdpa for SDPA
  nodes without RoPE.
  - Causal mask detection: Pre-flight forward pass identifies HuggingFace-style materialized causal masks so the fusion pass can strip them and use is_causal=True instead.
  - Added compiled model wrapper (_FP8FlashAttentionCompiledWrapper) with @torch._dynamo.disable to prevent re-tracing.
  - Added RoPE SDPA numerical accuracy tests and fuse_rope parametrization on model-level tests.

### New Files
  - shared_utils/fusion_utils.py: Shared FX graph fusion pass — RoPE pattern detection, SDPA detection, transpose unwrapping, parameterized graph surgery
  - shared_utils/custom_ops.py: Factory functions to register backend-specific custom ops with register_fake, and helpers to build fusion passes and compile functions
  - fp8_fa3/fusion_pass.py: FA3-specific custom op registration, rope_sdpa_fusion_pass, and compile_with_fp8_fusion entry point
  - quantization/triton_rope_qkv_quantization.py: Fused RoPE + QKV FP8 quantization Triton kernel

### Modified Files
  - shared_utils/attention.py: Added _fp8_rope_sdpa shared implementation
  - shared_utils/wrapper.py: Added _FP8FlashAttentionCompiledWrapper
  - shared_utils/setup.py: Added compile path routing via compile_fn parameter, moved detect_causal_mask to fusion_utils.py
  - quantization/quantization.py: Added _fp8_rope_sdpa_quantize
  - fp8_fa3/attention.py: Added fp8_fa3_rope_sdpa elementary block
  - fp8_fa3/setup.py: Passes compile_with_fp8_fusion as compile_fn
  - test_fp8_attention.py: Added TestFP8RopeSDPANumericalAccuracy, fuse_rope parametrization on model test

## Test Plan
`python -m pytest test/prototype/attention/test_fp8_attention.py -v`

## Example Usage
```python
  from torchao.prototype.attention import (
      AttentionBackend,
      LowPrecisionAttentionConfig,
      apply_low_precision_attention,
  )

  model = MyModel()

  # Enable low precision attention with config
  config = LowPrecisionAttentionConfig(
      backend=AttentionBackend.FP8_FA3,
  )
  model = apply_low_precision_attention(model, config)
  # Run torch.compile automatically applies rope fusion
  model = torch.compile(model)

  # Flash activation is handled internally by the wrapper
  output = model(inputs)

```
  ---
## Results
#### Single-Layer Results
Results directly comparing FA3 SDPA versus FA3 fp8 SDPA (including quantization time):
<img width="645" height="373" alt="image" src="https://github.com/user-attachments/assets/f64cba6d-4ac9-41b7-b0f7-bf93c67f6c13" />
#### Llama3 Model Results
Results comparing Llama3 model with FA3 SDPA versus Llama3 using the FA3 fp8 wrapper. Uses RoPE fusion.
Perplexity: 6.19 -> 6.24
<img width="634" height="285" alt="image" src="https://github.com/user-attachments/assets/7e454bdc-b0d2-43e6-8293-32012abde5f8" />

### DIFF
--- a/.github/workflows/claude-code.yml
+++ b/.github/workflows/claude-code.yml
@@ -1,0 +1,17 @@
+name: Claude Code
+
+on:
+  issue_comment:
+    types: [created]
+  issues:
+    types: [opened]
+
+jobs:
+  claude-code:
+    uses: pytorch/test-infra/.github/workflows/_claude-code.yml@main
+    permissions:
+      contents: read
+      pull-requests: write
+      issues: write
+      id-token: write
+    secrets: inherit
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,3 @@
+# TorchAO Claude Instructions
+
+Fill me in
--- a/test/integration/test_integration.py
+++ b/test/integration/test_integration.py
@@ -643,13 +643,13 @@ class TestSubclass(unittest.TestCase):
         )
 
     @parameterized.expand(COMMON_DEVICE_DTYPE)
-    @skip_if_rocm("ROCm enablement in progress")
     @skip_if_xpu("XPU enablement in progress")
     def test_int4_weight_only_quant_subclass_api_grouped(self, device, dtype):
         if dtype != torch.bfloat16:
             self.skipTest(f"Fails for {dtype}")
         if device == "cpu":
             self.skipTest("Only CUDA is supported for int4 weight only quantization v2")
+        ntile_size = 16 if torch.version.hip else 8
         for test_shape in [(256, 256, 16), (256, 256, 8)]:
             for groupsize in [64, 32]:
 
@@ -659,6 +659,7 @@ class TestSubclass(unittest.TestCase):
                         Int4WeightOnlyConfig(
                             group_size=groupsize,
                             int4_packing_format=Int4PackingFormat.TILE_PACKED_TO_4D,
+                            int4_tile_packed_ntile=ntile_size,
                         ),
                     )
 
--- a/torchao/prototype/attention/fp8_fa3/setup.py
+++ b/torchao/prototype/attention/fp8_fa3/setup.py
@@ -1,0 +1,27 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""FP8 FA3 backend setup."""
+
+import torch.nn as nn
+
+from torchao.prototype.attention.config import LowPrecisionAttentionConfig
+from torchao.prototype.attention.shared_utils.setup import setup_fp8_backend
+
+
+def setup_fp8_fa3(
+    model: nn.Module,
+    config: LowPrecisionAttentionConfig,
+) -> nn.Module:
+    """Set up FP8 FA3 attention on *model* and wrap it."""
+    from torchao.prototype.attention.fp8_fa3.attention import fp8_fa3_sdpa
+
+    return setup_fp8_backend(
+        model,
+        config,
+        flash_impl_name="FA3",
+        sdpa_fn=fp8_fa3_sdpa,
+    )
--- a/torchao/prototype/attention/quantization/quantization.py
+++ b/torchao/prototype/attention/quantization/quantization.py
@@ -1,0 +1,50 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+FP8 quantization for attention inputs.
+"""
+
+from typing import Tuple
+
+import torch
+
+
+def _fp8_sdpa_quantize(
+    q: torch.Tensor,
+    k: torch.Tensor,
+    v: torch.Tensor,
+) -> Tuple[
+    torch.Tensor,
+    torch.Tensor,
+    torch.Tensor,
+    torch.Tensor,
+    torch.Tensor,
+    torch.Tensor,
+]:
+    """Quantize Q, K, V to FP8 with per-head scaling."""
+    if q.dim() != 4:
+        raise ValueError(f"Expected 4D tensor for q, got {q.dim()}D")
+    if k.dim() != 4:
+        raise ValueError(f"Expected 4D tensor for k, got {k.dim()}D")
+    if v.dim() != 4:
+        raise ValueError(f"Expected 4D tensor for v, got {v.dim()}D")
+    if k.shape != v.shape:
+        raise ValueError(f"K and V shape mismatch: {k.shape} vs {v.shape}")
+    if q.shape[0] != k.shape[0]:
+        raise ValueError(f"Batch size mismatch: {q.shape[0]} vs {k.shape[0]}")
+    if q.shape[1] % k.shape[1] != 0:
+        raise ValueError(
+            f"Q head count ({q.shape[1]}) must be a multiple of K head count ({k.shape[1]})"
+        )
+    if q.shape[3] != k.shape[3]:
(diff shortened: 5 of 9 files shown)


howardzhang-cv:
Assuming this is gonna take a bit of back and forth to land to get it just right, since this is a user-facing change. @drisspg please take a look and let me know if the API looks okay

torchao/prototype/attention/fp8_fa3/wrappers.py
```diff
@@ -0,0 +1,99 @@
...
+    config: Optional[LowPrecisionAttentionConfig] = None,
+) -> nn.Module:
+    """Wrap model so its forward pass uses FP8 FA3 attention."""
+    if config is None:
+        config = LowPrecisionAttentionConfig()
+
+    original_forward = model.forward
+
+    def wrapped_forward(*args, **kwargs):
+        with _fp8_fa3_attention_context(config):
```
jerryzh168: not sure if we talked about this before, what's the thoughts on this v.s. following other torchao APIs, e.g.
```
q = Float8Tensor.from_hp(q)
k = Float8Tensor.from_hp(k)
v = Float8Tensor.from_hp(v)
# dispatch to `_fp8_fa3_sdpa` in Float8Tensor implementation
F.scaled_dot_product_attention(q, k, v, is_causal=True)
```
?

howardzhang-cv: Yeah, ideally we follow other torchao APIs but it's a bit difficult for attention. This first backend/recipe works well because it is a simple replacement of F.scaled_dot_product_attention, but as we add other features such as RoPE fusion or RoPE + Hadamard it becomes much more difficult, since it becomes model specific (i.e. if there is RoPE followed by SDPA, replace with RoPE + quantization fused kernel into fp8 SDPA). To future-proof other attention backends and recipes, I think it's better to have it as its own separate APIs. What are your thoughts?

jerryzh168: I thought the fused kernel replacement should happen in inductor? or is there a reason why we have to hand replace these in eager mode?

howardzhang-cv: Inductor unfortunately does not fuse RoPE with the quantization kernel

jerryzh168: oh I mean is this something that's not possible to do in inductor or just something that does not exist right now

howardzhang-cv: Not too sure, I feel like it should be possible in inductor? But it definitely is something that does not exist right now. Could look into it for the future, but for now, I think it would be simpler just to move this into prototype and have it be available through that. Will work on moving this over.

torchao/prototype/attention/fp8_fa3/quantization.py
```diff
@@ -0,0 +1,149 @@
...
+    if k.shape != v.shape:
+        raise ValueError(f"K and V shape mismatch: {k.shape} vs {v.shape}")
+    if q.shape[0] != k.shape[0]:
+        raise ValueError(f"Batch size mismatch: {q.shape[0]} vs {k.shape[0]}")
+    if q.shape[1] != k.shape[1]:
+        raise ValueError(f"Head count mismatch: {q.shape[1]} vs {k.shape[1]}")
+    if q.shape[3] != k.shape[3]:
+        raise ValueError(f"Head dim mismatch: {q.shape[3]} vs {k.shape[3]}")
+
+    if torch.compiler.is_compiling():
```
vkuzo: is there a test comparing numerics between these two paths

torchao/prototype/attention/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+quantization strategies and different PyTorch core attention backends.
+
+Note: Low-precision attention only supports inference (forward pass).
+Backward pass is not supported by the underlying backends.
+
+Example::
+
+    from torchao.prototype.attention import apply_low_precision_attention
+
+    model = MyTransformer()
```
vkuzo: we should specify what kind of syntax is automatically converted to low precision here.  F.SDPA? something else?

torchao/prototype/attention/api.py
```diff
@@ -0,0 +1,193 @@
...
+        )
+
+    from torch._inductor.compile_fx import compile_fx
+
+    from torchao.prototype.attention.fp8_fa3.fusion_pass import (
+        rope_sdpa_fusion_pass,
+    )
+
+    def fp8_attention_backend(gm, example_inputs):
+        """Custom Inductor backend that applies the RoPE + FP8 fusion pass."""
```
vkuzo: this should specify what kind of user syntax counts as "attention"

torchao/prototype/attention/config.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class AttentionBackend(str, Enum):
+    """
+    Backend kernel for computing attention.
+
+    Different backends have different hardware requirements and capabilities.
+    """
+
+    FP8_FA3 = "fa3"
```
vkuzo: make the string match the enum value name

torchao/prototype/attention/api.py
```diff
@@ -0,0 +1,193 @@
...
+        # cached and FA3 activation is a harmless no-op for the fused
+        # custom ops (they select the backend internally).
+        activate_flash_attention_impl("FA3")
+        try:
+            return compiled_mod(*args, **kwargs)
+        finally:
+            restore_flash_attention_impl()
+
+
+def apply_low_precision_attention(
```
vkuzo, line 60: I think this should be explicit that parts of torch.compile is used to do the logic swap

torchao/prototype/attention/fp8_fa3/quantization.py
howardzhang-cv: The numerics are exactly the same between the two paths. I tested the runtime and it seems like the triton implementation was much faster (~100 ms difference on llama3 model with 124k sequence length)

torchao/prototype/attention/fp8_fa3/wrappers.py
howardzhang-cv: @jerryzh168 I've gotten the rope fusion to work by using the inductor path. I've moved everything to a prototype folder. Due to the way attention works, I think it's better if it's own API, as it doesn't fit cleanly with others for now (especially with rope fusion and hadamard and stuff requiring inductor path). What do you think?

torchao/prototype/attention/__init__.py
howardzhang-cv: Added

torchao/prototype/attention/api.py
howardzhang-cv: Added

torchao/prototype/attention/config.py
howardzhang-cv: Fixed

torchao/prototype/attention/api.py
howardzhang-cv: Added, additionally added a warning as well to be even more explicit

PR #3865, opened 2026-02-12
Added benchmarking for new torchao low precision attention api

howardzhang-cv:
* #3947
* #3960
* #3930
* #3929
* __->__ #3865
* #3857
* #3959

## Summary
- Added new benchmark for new low precision attention API
- Can set baseline and test models between different backends: (fa2, fa3, fa3_fp8, fa4, fa4_fp8)
- uses flux.1-schnell model, 4 inference steps, DrawBench prompts
- has options to control number of prompts, torch.compile usage, warmup_iters, using debug prompts, number of inference steps, rope fusion
- Following the guidelines of https://github.com/pytorch/ao/pull/3502

## Example Run
python benchmarks/prototype/attention/eval_flux_model.py --baseline fa3 --test fa3_fp8 --compile

### DIFF
--- a/test/prototype/attention/test_fp8_attention.py
+++ b/test/prototype/attention/test_fp8_attention.py
@@ -20,20 +20,20 @@ from torchao.utils import torch_version_at_least
 if torch_version_at_least("2.11.0"):
     from torchao.prototype.attention.utils import _is_fa3_available, _is_hopper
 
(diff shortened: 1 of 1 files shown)


vkuzo:
do you want to eventually compare this with results from quantizing linear, moe, etc? if yes, could just build into existing scripts from the start, if no then separate script seems fine

PR #3866, opened 2026-02-12
Add per_tensor_scale static quantization flow support for NVFP4DynamicActivationNVFP4WeightConfig

jerryzh168:
* __->__ #3866

Summary:

 - Add static quantization support to NVFP4DynamicActivationNVFP4WeightConfig using a prepare/calibrate/convert observer flow
- Only the activation per_tensor_scale is determined statically; block-level dynamic quantization is unchanged
- Add NVFP4ObservedLinear to track activation amax during calibration
- Add step field ("prepare"/"convert") to config; auto-sets use_dynamic_per_tensor_scale=False

Test Plan:
- pytest test/prototype/mx_formats/test_inference_workflow.py::test_nvfp4_static_quantization_flow -xvs — prepare → calibrate → convert e2e flow, verifies observer insertion, quantized weight properties, SQNR
- pytest test/prototype/mx_formats/test_inference_workflow.py::test_nvfp4_static_vs_dynamic_quantization -xvs — static matches dynamic when calibrated on same input (torch.equal)
- pytest test/prototype/mx_formats/test_inference_workflow.py -x — no regressions on existing tests

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/prototype/mx_formats/inference_workflow.py
+++ b/torchao/prototype/mx_formats/inference_workflow.py
@@ -178,9 +178,8 @@ class NVFP4DynamicActivationNVFP4WeightConfig(AOBaseConfig):
           per_tensor_scale into the quantized weight tensor
         - At inference, the static activation per_tensor_scale is read from the
           weight tensor instead of being computed dynamically
-        - Note: activations are still dynamically quantized to FP4 at runtime
-          using the static per_tensor_scale; only the per_tensor_scale computation
-          is static, not the quantization itself
+        - Note: activation per-block scales are still computed dynamically at
+          inference time
 
     Note: When step is specified, use_dynamic_per_tensor_scale is automatically
     set to False.


torchao/prototype/mx_formats/inference_workflow.py
```diff
@@ -35,6 +39,49 @@
 )
 
 
+class NVFP4ObservedLinear(torch.nn.Linear):
```
vkuzo, line 42: it's a bit confusing to do observation using a linear override and then do inference with weight tensor subclass override.  Can we use the same method (weight tensor subclass override) in both places?

torchao/prototype/mx_formats/inference_workflow.py
```diff
@@ -35,6 +39,49 @@
...
+    def __init__(
+        self,
+        in_features: int,
+        out_features: int,
+        bias: bool = True,
+        device=None,
+        dtype=None,
+    ):
+        super().__init__(in_features, out_features, bias, device, dtype)
+        self.amax = torch.tensor(0.0, device=device)
```
vkuzo, line 62: init with `-inf`

torchao/prototype/mx_formats/inference_workflow.py
```diff
@@ -113,9 +160,32 @@ class NVFP4DynamicActivationNVFP4WeightConfig(AOBaseConfig):
...
+        - At inference, the static activation per_tensor_scale is read from the
+          weight tensor instead of being computed dynamically
+
+    Note: When step is specified, use_dynamic_per_tensor_scale is automatically
+    set to False.
+
     Configuration parameters:
     - use_triton_kernel: bool, whether to use fused triton kernel for activation scaling (default: True)
     - use_dynamic_per_tensor_scale: bool, whether to dynamically compute per tensor scale (default: True)
+    - step: Optional[QuantizationStep], the quantization step for observer-based flow
```
vkuzo, line 190: `step` is ambiguous, can this be named more clearly? `workflow_step`?

same for `QuantizationStep`.  Should that be something like `QuantizationWorkflowStep`?

torchao/prototype/mx_formats/inference_workflow.py
```diff
@@ -134,46 +205,115 @@ def __post_init__(self):
                 "NVFP4DynamicActivationNVFP4WeightConfig requires PyTorch 2.8 or later"
             )
 
+        # Allow string step values for convenience
```
vkuzo: just use `QuantizationStep[self.step]` to cast from string to enum?

test/prototype/mx_formats/test_inference_workflow.py
```diff
@@ -277,3 +277,150 @@ def test_nvfp4_quantize_3d_param_similar_to_vllm(self):
...
+            use_triton_kernel=False,
+        ),
+    )
+
+    # Test inference - when calibrating on the same input, static and dynamic
+    # should produce the same result since the per_tensor_scale will match
+    y_dynamic = m_dynamic(x)
+    y_static = m_static(x)
+
+    torch.testing.assert_close(y_dynamic, y_static)
```
vkuzo: test bitwise equivalence?

andrewor14:
Related issue: https://github.com/pytorch/ao/issues/2572

torchao/prototype/mx_formats/inference_workflow.py
jerryzh168: OK, it's using module right now just because it's simpler. Jesse also mentioned this as well in terms of this won't work for group_mm etc. I can refactor this to use tensor subclass as well.

can I do in a separate PR? there are a few places doing the same right now

torchao/prototype/mx_formats/inference_workflow.py
jerryzh168: this is max of absolute values actually so the min should be 0?

torchao/prototype/mx_formats/inference_workflow.py
jerryzh168: sounds good, let me do in a separate PR to refactor all occurrences?

torchao/prototype/mx_formats/inference_workflow.py
```diff
@@ -113,9 +160,32 @@ class NVFP4DynamicActivationNVFP4WeightConfig(AOBaseConfig):
...
+    - A global per_tensor_scale (float32)
+    - Per-block scales (float8_e4m3fn, block_size=16)
+
+    The activation per_tensor_scale can be determined in two ways:
+
+    1. Dynamic (default, step=None, use_dynamic_per_tensor_scale=True):
+        - Both weight and activation per_tensor_scale are computed at runtime
+          from the tensor amax
+
+    2. Static via observer flow (step="prepare"/"convert"):
```
andrewor14: Should this be under a separate config? I feel it's a bit weird to have a static flow option in a "dynamic activation" config

torchao/prototype/mx_formats/inference_workflow.py
```diff
@@ -126,54 +196,117 @@ class NVFP4DynamicActivationNVFP4WeightConfig(AOBaseConfig):
 
     use_triton_kernel: bool = True
     use_dynamic_per_tensor_scale: bool = True
+    step: Optional["QuantizationStep"] = None
```
andrewor14, line 201: If you separate it into a new config then step can be not optional, which will simplify the branches below

torchao/prototype/mx_formats/inference_workflow.py
jerryzh168: hmmm, there is still a "dynamic" part in this config, the static part is only for the global tensorwise scale that is applied before the dynamic quantization actually, it is a bit confusing use case by itself..

jerryzh168: I can clarify in the docstring

torchao/prototype/mx_formats/inference_workflow.py
```diff
@@ -113,9 +160,35 @@ class NVFP4DynamicActivationNVFP4WeightConfig(AOBaseConfig):
...
+
+    2. Static per_tensor_scale via observer flow (step="prepare"/"convert"):
+        - Weight per_tensor_scale is computed from weight amax at convert time
+        - Activation per_tensor_scale is determined statically during calibration:
+          step="prepare" inserts observers, then after running calibration data,
+          step="convert" extracts the observed amax and bakes the activation
+          per_tensor_scale into the quantized weight tensor
+        - At inference, the static activation per_tensor_scale is read from the
+          weight tensor instead of being computed dynamically
+        - Note: activations are still dynamically quantized to FP4 at runtime
```
jcaip: nit: this note is kind of confusing, maybe something more like: 
```
        - Note: When using (step="prepare"/"convert") activation per-block scales are still dynamically quantized. 

```

torchao/prototype/mx_formats/inference_workflow.py
andrewor14: ah ok makes sense

PR #3871, opened 2026-02-12
[QAT] Add low-bit FSDP all-gather path for affine fake quant tensor

### LINKED ISSUE #1224: [QAT] Low-bit FSDP all-gather for QAT
Had this idea and discussed briefly with @andrewor14.

Conceptually the current QAT + FSDP looks like this

 - sharded FP32 weight -> all-gather in BF16 -> fake quantize

However, we can do low-bit all-gather, since the weight can be quantized before all-gather

- sharded FP32 weight -> (real) quantize -> all-gather in low-bit -> dequantize

In terms of perf, basically we are comparing between (ignoring potential fusion surrounding this)
1. BF16 all-gather + fake quantize
2. (Real) quantize (1/NGPU) + Low-bit all-gather + Dequant

This might be a small perf win, especially when distributed comm is bottleneck. Might be useful for QAT recipes in torchtune.

This is probably a low priority. Need to quantify the speedup, if any.

In terms of implementation, we can follow float8 design (https://github.com/pytorch/ao/blob/000a49026459dd1dadf5ca34322d98e7b1680250/torchao/float8/fsdp_utils.py)
- A tensor subclass to hold original weight + use FSDP2 all-gather extension: possibly extend this https://github.com/pytorch/ao/blob/000a49026459dd1dadf5ca34322d98e7b1680250/torchao/quantization/qat/affine_fake_quantized_tensor.py
- Another tensor subclass to hold quantized weight. If AQT has basic support for backward, maybe we can use AQT directly. Otherwise, need to have another subclass.

MagellaX:
## What
- add FSDP pre/post all-gather hooks to `_AffineFakeQuantizedTensor`
- quantize local shards before all-gather and dequantize gathered weights after all-gather
- preserve quantization metadata on the tensor subclass for deterministic behavior

## Tests
- add two QAT tests covering low-bit all-gather roundtrip and per-tensor block size behavior
- `python -m pytest test/quantization/test_qat.py -k "affine_fake_quantized_tensor_fsdp_all_gather" -q`

closes #1224

### DIFF
--- a/test/quantization/test_qat.py
+++ b/test/quantization/test_qat.py
@@ -309,6 +309,127 @@ class TestQAT(TestCase):
         )
         torch.testing.assert_close(out, out_ptq, atol=0, rtol=0)
 
+    def test_affine_fake_quantized_tensor_fsdp_all_gather_low_bit_roundtrip(self):
+        from torchao.quantization.qat.affine_fake_quantized_tensor import (
+            _to_affine_fake_quantized,
+        )
+
+        class _MPPolicy:
+            def __init__(self, param_dtype: torch.dtype):
+                self.param_dtype = param_dtype
+
+        torch.manual_seed(self.SEED)
+        x = torch.randn(8, 16, dtype=torch.float32)
+        block_size = (1, 4)
+        target_dtype = torch.int8
+        mp_policy = _MPPolicy(torch.bfloat16)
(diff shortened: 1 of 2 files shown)


MagellaX:
Would anybody be able to review this??

PR #3872, opened 2026-02-12
WIP [mx][test] Refactor skip logic in test_mx_linear

ugolowic:
I plan to enable MX format tests on XPU by parameterizing the existing CUDA tests with a device argument (cuda/xpu). The current skip logic is spread across complex if/else blocks, which makes it hard to add new parameters safely. This PR refactors those conditions into kernel/recipe compatibility matrices with helper functions that perform the skips based on compatibility.

@vkuzo please take a look.

### DIFF
--- a/test/prototype/mx_formats/test_mx_linear.py
+++ b/test/prototype/mx_formats/test_mx_linear.py
@@ -5,6 +5,8 @@
 # LICENSE file in the root directory of this source tree.
 
 import copy
+from dataclasses import dataclass
+from typing import Callable, Optional, Set
 
 import pytest
 import torch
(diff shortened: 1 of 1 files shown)


vkuzo:
context on which XPU hardware will support MX formats would be useful here to motivate the need for this

ugolowic:
@vkuzo this change is not XPU-related; it only cleans up the existing CUDA tests. I’ll include the details you requested in follow‑up XPU changes.

vkuzo:
I think to motivate the need for the current PR, we should clarify what exactly "MX on XPU" means - what hardware supports it, how many users are there, etc.  Without this data, it's hard to justify increase in complexity for adding XPU support.

ugolowic:
Got it—I’ll follow up on the hardware support.
In the meantime, since this change clarifies existing code without adding test complexity, would you be willing to share your feedback?

vkuzo:
yeah, this PR seems to make things more complicated for the only callsite currently using the abstractions, if statements are simpler to understand for a single callsite.  I think the value from something like this would come with more callsites, and it would probably need some repo-wide design on the new abstractions.

PR #3877, opened 2026-02-12
Fix FqnToConfig module skipping for _default and regex modules

### LINKED ISSUE #3867: FqnToConfig doesn't skip some modules when set _default.
### Description

When using `FqnToConfig` (or `ModuleFqnToConfig`) with a configuration that includes both:
- A regex pattern matching to `None` (intended to skip quantization for certain modules)
- A `_default` configuration (intended to quantize other modules)

Modules that match the `None` pattern are incorrectly still quantized by `_default` instead of being skipped.

### Steps to Reproduce

```
from collections import OrderedDict
import torch
import torch.nn as nn
    quantize_,
    ModuleFqnToConfig,
)

class TestModel(nn.Module):
    def __init__(self):
        super().__init__()
        self.time_embed = nn.Sequential(
            nn.Linear(1024, 1024),
            nn.ReLU(),
            nn.Linear(1024, 1024),
        )
        self.linear1 = nn.Linear(1024, 1024)

    def forward(self, x):
        x = self.time_embed(x)
        x = self.linear1(x)
        return x

model = TestModel()
model.eval()

# Expected: time_embed should NOT be quantized, linear1 should be quantized
cfg = ModuleFqnToConfig(
    module_fqn_to_config=OrderedDict([
        ("re:.*time_embed.*", None),  # Skip time_embed
        ("_default", Int8DynamicActivationInt8WeightConfig()),  # Quantize other modules
    ])
)

quantize_(model, cfg, filter_fn=None)
```

error logs: 
  File "/root/workspace/optkit/examples/ttt.py", line 43, in <module>
    quantize_(model, cfg, filter_fn=None)
  File "/usr/local/lib/python3.10/dist-packages/torchao/quantization/quant_api.py", line 502, in quantize_
    _fqn_to_config_handler(module, module_fqn, config)
  File "/usr/local/lib/python3.10/dist-packages/torchao/quantization/quant_api.py", line 2690, in _fqn_to_config_handler
    return handler(module, c)
  File "/usr/local/lib/python3.10/dist-packages/torchao/quantization/quant_api.py", line 1629, in _int8_dynamic_activation_int8_weight_transform
    assert hasattr(module, "weight"), (
AssertionError: applying int8 dynamic activation int8 weight quant requires module to have weight attributebut {module} does not have one

jcaip:
Summary:

Fixes #3867. We were missing handling None appropriately like we do for exact fqn match earlier in the function.

Test Plan:

```
pytest test/quantization/test_quant_api.py -k
test_fqn_to_config_regex_skip
```

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/quantization/quant_api.py
+++ b/torchao/quantization/quant_api.py
@@ -1980,6 +1980,8 @@ def _fqn_to_config_handler(
                 if c is not None:
                     handler = _QUANTIZE_CONFIG_HANDLER[type(c)]
                     return handler(module, c)
+                else:
+                    return module
 
     # If no module_fqn or parameter_fqn matches, then we apply _default
     if not parameter_config_found:
--- a/test/quantization/test_quant_api.py
+++ b/test/quantization/test_quant_api.py
@@ -692,6 +692,43 @@ class TestFqnToConfig(TestCase):
         assert str(linear_model).startswith(expected_starting_str)
         assert str(linear_model.linear1.weight) in str(linear_model)
 
+    def test_fqn_to_config_regex_skip(self):
+        """Test that regex pattern with None config skips matching modules."""
+
+        class TestModel(torch.nn.Module):
+            def __init__(self):
+                super().__init__()
+                self.time_embed = torch.nn.Sequential(
+                    torch.nn.Linear(128, 128), torch.nn.Linear(128, 128)
+                )
+                self.linear1 = torch.nn.Linear(128, 128)
+
+            def forward(self, x):
+                x = self.time_embed(x)
+                x = self.linear1(x)
+                return x
+
+        model = TestModel().eval()
+
+        cfg = FqnToConfig(
+            {
+                "re:.*time_embed.*": None,
+                "_default": Float8WeightOnlyConfig(),
+            }
+        )
+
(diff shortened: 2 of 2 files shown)


torchao/quantization/quant_api.py
jerryzh168, line 1830: we should update docs to talk about this as well I think?

also any plans to deprecate _default?

jcaip: This was a bug with the actual behavior not matching the documented behavior, now we match the docs, so they shouldn't need to be updated. 

>also any plans to deprecate _default?

Yeah, planning to do this for 0.17 release. 

Also should I mark this bc-breaking? top level-api behavior changed but it was the wrong behavior in the first place.

jerryzh168: oh I see. I think it's fine to say bug fix then. We are also planning to deprecate these flags and just use `module: training` / `module: inference` / `module: qat` flags that only capture important commits

jcaip: I think we still want bc-breaking as well though, right? maybe not as top-leve module but would be good to be able to tag bc-breaking prs.

PR #3878, opened 2026-02-12
Update pytorch_sphinx_theme2 version to 0.4.3

svekars:
- Update pytorch_sphinx_theme2 version to 0.4.3
- Add logo


jerryzh168:
@svekars looks like build docs failed right now, do you know how can we fix? I was actually thinking of reverting currently https://github.com/pytorch/ao/pull/3879

PR #3880, opened 2026-02-13
use helion instead of triton for low precision attention quantization kernels

howardzhang-cv:
* __->__ #3880
* #3865
* #3857

Summary:

Test Plan: python benchmarks/prototype/attention/eval_flux_model.py --compile --num_prompts 1

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchao/prototype/attention/__init__.py
+++ b/torchao/prototype/attention/__init__.py
@@ -35,7 +35,10 @@ Example::
 """
 
 from torchao.prototype.attention.api import apply_low_precision_attention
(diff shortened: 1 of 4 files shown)


howardzhang-cv:
@yf225 Here is the compiler error. I didn't do the MRE since I was afraid some context would be lost since I'm not sure exactly what part leads to the error. [helion_compiler_error.txt](https://github.com/user-attachments/files/25277320/helion_compiler_error.txt)

howardzhang-cv:
Merged this with https://github.com/pytorch/ao/pull/3857

PR #3882, opened 2026-02-13
Add support for `torch._higher_order_ops.scan` in pt2e quantization

jerryzh168:
* __->__ #3882

Summary:

- Add support for torch._higher_order_ops.scan in pt2e quantization by recursively applying prepare_pt2e and convert_pt2e to the scan op's `combine_fn` subgraph using the same quantizer from the top-level module
- Register scan's combine_fn subgraph in _get_control_flow_submodules alongside existing cond and map_impl handlers
- In convert_pt2e, process scan subgraphs before the main conversion passes to prevent DuplicateDQPass from encountering stale observer references during recursive graph linting

Test Plan:
- Added test_scan_op_quantization which exports a model using scan with a linear layer in the combine_fn, runs prepare_pt2e + calibration + convert_pt2e, and verifies observers are inserted after prepare, q/dq
ops are present after convert, and the quantized model produces correct output
- Ran the full test_quantize_pt2e.py test suite (59 tests) — all pass

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/quantization/pt2e/test_quantize_pt2e.py
+++ b/test/quantization/pt2e/test_quantize_pt2e.py
@@ -14,7 +14,6 @@ import unittest
 
 import torch
 from torch import Tensor
-from torch.ao.ns.fx.utils import compute_sqnr
 from torch.ao.quantization import QConfigMapping
 from torch.ao.quantization.qconfig import (
     QConfig,
@@ -63,6 +62,7 @@ from torchao.quantization.pt2e.quantizer.composable_quantizer import (  # noqa: 
 from torchao.quantization.pt2e.quantizer.embedding_quantizer import (  # noqa: F811
     EmbeddingQuantizer,
 )
+from torchao.quantization.utils import compute_error
 from torchao.testing.model_architectures import ConvWithSharedWeightInExportedModel
 from torchao.testing.pt2e._xnnpack_quantizer import (
     XNNPACKQuantizer,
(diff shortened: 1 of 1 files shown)


test/quantization/pt2e/test_quantize_pt2e.py
```diff
@@ -14,6 +14,7 @@
 
 import torch
 from torch import Tensor
+from torch.ao.ns.fx.utils import compute_sqnr
```
andrewor14: should we just copy this now to make it easier to delete torch.ao in the future?

torchao/quantization/pt2e/quantize_pt2e.py
```diff
@@ -113,6 +113,19 @@ def calibrate(model, data_loader):
...
+    # This is done after the top-level prepare since prepare() does not
+    # modify scan subgraphs (they are separate GraphModules), so ordering
+    # does not matter here.
+    for node in model.graph.nodes:
+        if node.op == "call_function" and node.target is torch.ops.higher_order.scan:
+            scan_combine_fn_node = node.args[0]
+            assert isinstance(scan_combine_fn_node, Node)
+            assert scan_combine_fn_node.op == "get_attr"
+            assert isinstance(scan_combine_fn_node.target, str)
+            scan_combine_fn = model.get_submodule(scan_combine_fn_node.target)
```
andrewor14, line 126: does torch.export convert `scan` into a sub graphmodule?

andrewor14 (approved):

test/quantization/pt2e/test_quantize_pt2e.py
jerryzh168: oh right, we shouldn't import from there, I'll change this to import from torch.quantization.utils, thanks for catching this

torchao/quantization/pt2e/quantize_pt2e.py
jerryzh168: scan is an call_function op that has a submodule arg I think, something like:

```
submodule_arg = getattr(...)
scan_out = scan(..., submodule_arg)
```

PR #3883, opened 2026-02-13
Remove Float8DynamicActivationFloat8SemiSparseWeightConfig

jcaip:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #3887
* #3886
* #3885
* #3884
* __->__ #3883

This config was deprecated in favor of Float8DynamicActivationFloat8WeightConfig
with packing_format=Float8PackingFormat.SPARSE_CUTLASS and granularity=PerRow().
Remove the class definition, handler, and all references from imports, tests, and
benchmarks.

Co-authored-by: Cursor <cursoragent@cursor.com>

Differential Revision: [D93251982](https://our.internmc.facebook.com/intern/diff/D93251982)

PR #3884, opened 2026-02-13
Remove Int8DynamicActivationInt4WeightConfig

jcaip:
* #3905
* #3904
* #3903
* #3902
* #3901
* #3894
* #3887
* __->__ #3884
* #3883

This config was deprecated in favor of Int8DynamicActivationIntxWeightConfig.
Remove the class definition, handler, and all references from imports, tests,
QAT code, benchmarks, and documentation. Update QAT docs to reference
Int4WeightOnlyConfig as the example base config.

### DIFF
--- a/docs/source/eager_tutorials/finetuning.rst
+++ b/docs/source/eager_tutorials/finetuning.rst
@@ -205,13 +205,14 @@ because we are not actually casting the fake quantized values.
 
 .. code:: py
 
-  from torchao.quantization import quantize_, Int4WeightOnlyConfig
+  from torchao.quantization import quantize_, Int8DynamicActivationIntxWeightConfig
   from torchao.quantization.qat import QATConfig
+  from torchao.quantization.granularity import PerGroup
 
   model = get_model()
 
   # prepare: swap `torch.nn.Linear` -> `FakeQuantizedLinear`
-  base_config = Int4WeightOnlyConfig(group_size=32)
+  base_config = Int8DynamicActivationIntxWeightConfig(weight_dtype=torch.int4, weight_granularity=PerGroup(32))
   quantize_(model, QATConfig(base_config, step="prepare"))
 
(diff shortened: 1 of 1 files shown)


docs/source/eager_tutorials/finetuning.rst
```diff
@@ -205,13 +205,13 @@ because we are not actually casting the fake quantized values.
 
 .. code:: py
 
-  from torchao.quantization import quantize_, Int8DynamicActivationInt4WeightConfig
+  from torchao.quantization import quantize_, Int4WeightOnlyConfig
```
jerryzh168: replacement for this config is `Int8DynamicActivationIntxWeightConfig` I think, maybe change to that?

test/quantization/test_da8w4_cpu.py
```diff
@@ -4,177 +4,4 @@
...
-                )
-            assert torch.allclose(y, y_ref)
-
-
-common_utils.instantiate_parametrized_tests(TestDa8w4Cpu)
-
-
-if __name__ == "__main__":
-    run_tests()
+# Tests removed: Int8DynamicActivationInt4WeightConfig has been removed.
```
jerryzh168, line 7: just remove the file?

also cc @Xia-Weiwen we are using `Int8DynamicActivationIntxWeightConfig` to replace `Int8DynamicActivationInt4WeightConfig` now, in case you want to keep the CPU version of this config

andrewor14:
Remember to import this. There were a lot of internal use cases!

test/quantization/test_da8w4_cpu.py
Xia-Weiwen: Oh, Thanks @jerryzh168 I missed this message. And the DA8W4 workflow is broken on CPU now. I will work on a PR to fix the workflow and add these test cases back.

PR #3887, opened 2026-02-13
Remove UIntXWeightOnlyConfig

jcaip:
* #3905
* #3904
* #3903
* #3902
* #3901
* #3894
* __->__ #3887
* #3884
* #3883

This config was deprecated and scheduled for deletion. Remove the class
definition, handler, and all references from imports, tests, benchmarks,
and the autoround eval script. This also removes the entire BC import
block from quant_api.py since all prototype configs have been removed.


jerryzh168:
btw we should also open PRs in diffusers to remove some of these options

PR #3892, opened 2026-02-16
Build: put x86 options in a dedicated function

Xia-Weiwen:
As the title. To make `setup.py` more organized.


Xia-Weiwen:

PR #3895, opened 2026-02-17
Add precomputed scale support to MXTensor and fix swizzled scales

jcaip:
* #3935
* #3897
* #3899
* __->__ #3895

This PR adds in support for passing a precomputed scale to MXTensor, via a `scale` kwarg in `to_mx`. 
When specified, we will use the precomputed scale to do MX quantization instead of calculating a scale dynamically.

This is needed for GPTQ. 

Test:
```python
pytest test/prototype/mx_formats_test_mx_tensor.py -k test_to_mx_precomputed_scale
```

### DIFF
--- a/torchao/prototype/mx_formats/mx_tensor.py
+++ b/torchao/prototype/mx_formats/mx_tensor.py
@@ -524,9 +524,9 @@ def tensor_size_fp4x2_to_hp(orig_size, is_contiguous):
 class MXTensor(TorchAOBaseTensor):
     tensor_data_names = ["qdata", "scale"]
     tensor_attribute_names = [
-        "elem_dtype",
+        "_elem_dtype",
         "block_size",
-        "orig_dtype",
+        "_orig_dtype",
         "kernel_preference",
         "act_quant_kwargs",
         "is_swizzled_scales",
@@ -573,9 +573,9 @@ class MXTensor(TorchAOBaseTensor):
         ), "unsupported"
         self.qdata = qdata
         self.scale = scale
-        self.elem_dtype = elem_dtype
+        self._elem_dtype = elem_dtype
         self.block_size = block_size
-        self.orig_dtype = orig_dtype
+        self._orig_dtype = orig_dtype
         self.kernel_preference = kernel_preference
         self.act_quant_kwargs = act_quant_kwargs
         self.is_swizzled_scales = is_swizzled_scales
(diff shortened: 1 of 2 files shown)


torchao/prototype/mx_formats/mx_tensor.py
```diff
@@ -348,6 +348,101 @@ def to_mx(
...
+):
+    """
+    Like to_mx, but uses a precomputed e8m0 scale instead of computing one
+    from the data.  This is useful for GPTQ which computes the scale once
+    per group and then quantizes each column individually with that same
+    scale.
+
+    Returns (scale_e8m0_biased, data_lp) -- the same signature as to_mx.
+    """
+    assert data_hp.dtype in (
```
vkuzo: can we reuse all of this logic instead of copy-paste

jcaip: Sure, I can add a scale kwarg to `to_mx` and reuse the logic, does that sound good?

vkuzo:
I think it's cleaner to structure the code reuse this way (names not real, just demo)

```python
def to_mx(x_hp, ...):
    scale = calculate_scale(x_hp, ...)
    qdata = to_mx_with_precomputed_scale(x_hp, scale, ...)

def to_mx_with_precomputed_scale(x_hp, scale, ...):
    ...
```

jcaip:
@jcaip has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D95073527).

torchao/prototype/mx_formats/mx_tensor.py
```diff
@@ -103,64 +103,18 @@ class QuantizeTensorToMXKwargs(QuantizeTensorKwargs):
...
     elem_dtype: Union[torch.dtype, str],
     block_size: int,
     scaling_mode: ScaleCalculationMode = ScaleCalculationMode.FLOOR,
     is_swizzled_scales: bool = False,
-):
-    """
-    Takes a high precision tensor and converts to MX scale and raw data, in
-    naive layout (scale and raw data are separate tensors).
+    scale: Optional[torch.Tensor] = None,
+) -> tuple[torch.Tensor, torch.Tensor]:
```
vkuzo, line 198: it's weird to pass a scale here and then return it, feels like it would be cleaner for this to be a separate function

```
def to_mx_with_precomputed_scale(..., scale, ...) -> torch.Tensor
```

vkuzo: not sure why you need to change the current function if you already have `scale_and_cast_hp_data`

torchao/prototype/mx_formats/mx_tensor.py
```diff
@@ -103,52 +103,90 @@ class QuantizeTensorToMXKwargs(QuantizeTensorKwargs):
...
-def _to_mx_rceil(
+def to_mx_with_precomputed_scale(
     data_hp: torch.Tensor,
-    max_abs: torch.Tensor,
-    max_pos: float,
-) -> tuple[torch.Tensor, torch.Tensor]:
+    scale: torch.Tensor,
+    elem_dtype: Union[torch.dtype, str],
+    block_size: int,
+    scaling_mode: ScaleCalculationMode = ScaleCalculationMode.FLOOR,
```
vkuzo, line 111: if scale is already computed, can we delete this arg?

jcaip: We still need this for `ScaleCalculationMode.RCEIL`, which has a different codepath from the other scaling modes in `scale_and_cast_hp_data`

vkuzo: I see, looks like a lot of technical debt has accumulated in the code which does the casting to mx.  How about:
* option 1: we clean up the casting code, unifying the two paths that cast the high precision data, then continue with this PR.  This should be tested carefully for numerics and performance
* option 2: you write your own helper for the RCEIL path to mx (copy-paste as needed), file an issue to do the cleanup in option 1, and we can unify later after someone (probably me) does option 1

let me know which one looks better?

PR #3862, opened 2026-02-11
[moe training] refactor configs, recipes; support converting linears + grouped gemms in a single quantize_() call

danielvegamyhre:
# Summary
To prepare the MoE training prototype to stable we need the dev experience to be consistent with the rest of torchao, namely the fp8/mxfp8 linear training.

This PR:
- Splits FP8 grouped mm and MXFP8 grouped mm related code into separate files to organize by derived dtype
- Update APIs and configs to align with the patterns of MXLinearConfig, MXLinear.
- Add support for converting both linears and grouped gemms in a single `quantize_()` call

These files contain the important changes, everything else is peripheral:
- `torchao/prototype/moe_training/config.py`
- `torchao/prototype/moe_training/conversion_utils.py`
- `torchao/prototype/moe_training/mxfp8_grouped_mm.py`
- `torchao/prototype/moe_training/fp8_grouped_mm.py`
-  `torchao/prototype/moe_training/tensor.py`
- `torchao/quantization/quant_api.py`

#### New model conversion API

```python
    model = MoEModel()

    config = FqnToConfig(
        fqn_to_config=OrderedDict(
            [
                (
                    "re:.*experts.*",
                    MXFP8GroupedMMConfig.from_recipe(MXFP8GroupedMMRecipe.MXFP8_RCEIL),
                ),
                (
                    "re:^(pre_moe|post_moe)$",
                    MXLinearConfig.from_recipe_name(MXLinearRecipeName.MXFP8_CUBLAS_RCEIL),
                ),
            ]
        )
    )

    quantize_(model, config, filter_fn=None)
```

#### New recipes

```python

class FP8GroupedMMRecipe(Enum):
    """FP8 recipes for grouped matrix multiplication."""
    FP8_ROWWISE = "fp8_rowwise"


class MXFP8GroupedMMRecipe(Enum):
    """MXFP8 recipes for grouped matrix multiplication."""
    # TODO: add floor variants
    MXFP8_RCEIL = "mxfp8_rceil"
    MXFP8_RCEIL_WGRAD_WITH_HP = "mxfp8_rceil_wgrad_with_hp"
    MXFP8_EMULATED_RCEIL = "mxfp8_emulated_rceil"
```

#### New configs

```python
class GroupedMMConfig(AOBaseConfig):
    """Base configuration for grouped matrix multiplication. Not intended to be used directly."""
    pass


@dataclass
class FP8GroupedMMConfig(GroupedMMConfig):
[...]

PR #3898, opened 2026-02-17
[mxfp8 moe training] move mxfp8 grouped mm code into mx_formats

danielvegamyhre:
Next PR in stack:
- #3922 

Prior PR in stack:  
- #3862 

## Summary
- Move mxfp8 moe training code:
    - from: `torchao/prototype/moe_training`  
    - to: `torchao/prototype/mx_formats/grouped_mm/`
- Make imports lazy to avoid circular dependencies

 Note: This PR doesn't move the fp8 moe training code, which for now still lives in in `torchao/prototype/moe_training`. Subsequent PRs will handle this.

## Tests
- `./test/prototype/moe_training/test_everything.sh`


danielvegamyhre:
another incremental refactor toward end state, cc @vkuzo for review

vkuzo:
I don't think just moving the files to a different folder is worth the effort (arguably worse for mx_formats since it becomes more confusing), IMO better to only move things to `mx_formats` if we are also making the abstractions consistent at the same time.

danielvegamyhre:
> IMO better to only move things to `mx_formats` if we are also making the abstractions consistent at the same time.

The mxfp8 grouped mm configs and apis were made more consistent with the MX linear configs in the prior PR. 


> I don't think just moving the files to a different folder is worth the effort (arguably worse for mx_formats since it becomes more confusing)

i am surprised by this comment, last time we talked you said: `"no separate moe_training and mx_formats folders (combined one needs a name, and inference stuff might be moved out)"`  which i agree with and this PR is part of the work on that.

i think the end state of this PR and subsequent ones will be much better organized than the current state - let's clarify the current state and desired end state here to make sure we are aligned:

Current:
- `torchao.prototype.mx_formats` - has training for mxfp8 and inference for mxfp8. has inference for nvfp4. also supports misc other mx formats that are not really used iirc
- `torchao.prototype.moe_training.mxfp8_grouped_mm`: training for mxfp8 gmms
- `torchao.prototype.moe_training.fp8_grouped_mm`: training for fp8 gmms
- `torchao.float8` - training for fp8 linears

intended end state:
- `torchao.training.fp8` - further organized into linears, grouped mms
- `torchao.training.mxfp8` - further organized into linears, grouped mms
- `torchao.training.nvfp4` - (when it exists - same^)
- (inference related code in `torchao.prototype.mx_formats` moved to different location(s), TBD)

PR #3906, opened 2026-02-17
Make `_emulated_mxfp8_scaled_grouped_mm_2d_2d` torch.compile compatible

lyprince:
* Replace loop-based dequantization in `_emulated_mxfp8_scaled_grouped_mm_2d_2d` with vectorized ops, removing the `@torch.compiler.disable` decorator
* The previous implementation used `offs.tolist()` and a Python loop with data-dependent control flow, which prevented `torch.compile` from tracing through the emulated MXFP8 wgrad path
* Remove the `pytest.skip` for `EMULATED` + `use_compile=True` in `test_mxfp8_grouped_gemm_with_dq_fwd_bwd`


danielvegamyhre:
this is much better, thanks!

lyprince:
Looks like CI is blocked by PR label check?

PR #3910, opened 2026-02-17
Add flashinfer-python to CI

jerryzh168:
* #3912
* #3911
* __->__ #3910

Summary:
Adding flashinfer as an optional depenency so we can use their nvfp4_quantize op
(it's faster for some workloads that we are interested in)

Test Plan:
tested `_is_flashinfer_available` locally

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/.github/scripts/torchao_model_releases/quantize_and_upload.py
+++ b/.github/scripts/torchao_model_releases/quantize_and_upload.py
@@ -7,6 +7,7 @@
 import argparse
 from typing import List
 
+import huggingface_hub
 import torch
 import transformers
 from huggingface_hub import ModelCard, get_token, whoami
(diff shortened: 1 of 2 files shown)


torchao/utils.py
```diff
@@ -1270,6 +1270,13 @@ def _is_mslk_available():
     return True
 
 
+def _is_flashinfer_available():
+    return (
+        importlib.util.find_spec("flashinfer") is not None
+        and importlib.util.find_spec("tvm_ffi") is not None
```
jerryzh168, line 1278: there are more like these that I need to add, but I'll do it in https://github.com/pytorch/ao/pull/3912

PR #3911, opened 2026-02-17
Refactor `use_triton_kernel` to use `nvfp4_quantize_kernel_choice`

jerryzh168:
* #3912
* __->__ #3911

Summary:
This is to prefer the addition of flashinfer quantize kernel path in next PR

use_triton_kernel==True --> `QuantizeToNVFP4KernelChoice.MSLK`
use_triton_kernel==False --> `QuantizeToNVFP4KernelChoice.TRITON`

Note: this breaks BC for the users of the prototype API

for configs whose default is `use_triton_kernel = True` (e.g. `NVFP4DynamicActivationNVFP4WeightConfig`), an error will be thrown when the flag is set to False,
for configs whose default is `use_triton_kernel = False` (e.g. `NVFP4FakeQuantizeConfig`), an error will be thrown when the flag is set to True

we'll make these changes internally later

Test Plan:
python test/prototype/mx_formats/test_inference_workflow.py

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/prototype/mx_formats/test_inference_workflow.py
+++ b/test/prototype/mx_formats/test_inference_workflow.py
@@ -188,7 +188,10 @@ def test_inference_workflow_nvfp4(
         and quantize_to_nvfp4_kernel_choice == QuantizeToNVFP4KernelChoice.MSLK
     ):
         pytest.skip("unsupported configuration")
-    if quantize_to_nvfp4_kernel_choice == QuantizeToNVFP4KernelChoice.MSLK and not use_dynamic_per_tensor_scale:
+    if (
+        quantize_to_nvfp4_kernel_choice == QuantizeToNVFP4KernelChoice.MSLK
+        and not use_dynamic_per_tensor_scale
+    ):
         pytest.skip("unsupported configuration")
 
     if use_inference_mode and (
--- a/test/quantization/pt2e/test_x86inductor_fusion.py
+++ b/test/quantization/pt2e/test_x86inductor_fusion.py
@@ -3093,6 +3093,9 @@ class TestDynamicPatternMatcher(TestPatternMatcherBase):
 
     @skipIfNoDynamoSupport
     @skipIfNoONEDNN
+    @unittest.skipIf(
+        torch_version_at_least("2.11.0.dev"), "Doesn't work with torch 2.11.0.dev+"
+    )
     def test_q_attention_block(self):
         for annotate_matmul in [True, False]:
             self._test_q_attention_block_helper(annotate_matmul=annotate_matmul)
@@ -3100,6 +3103,9 @@ class TestDynamicPatternMatcher(TestPatternMatcherBase):
     @skipIfNoDynamoSupport
     @skipIfNoONEDNN
     @skipIfNoFloat8Support
+    @unittest.skipIf(
+        torch_version_at_least("2.11.0.dev"), "Doesn't work with torch 2.11.0.dev+"
+    )
     def test_fp8_q_attention_block(self):
         for annotate_matmul in [True, False]:
             self._test_q_attention_block_helper(


torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -150,13 +169,46 @@ def to_nvfp4(
         assert len(data_hp.shape) in (2, 3), "unsupported"
         leading_dims, M, K = data_hp.shape[:-2], data_hp.shape[-2], data_hp.shape[-1]
 
-        if use_triton_kernel:
+        if nvfp4_quantize_kernel_choice == NVFP4QuantizeKernelChoice.TRITON:
```
vkuzo: this logic is choosing a single kernel, can we just have one if statement instead of adding an intermediate variable `kernel_choice`?

torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -150,13 +169,46 @@ def to_nvfp4(
...
+            kernel_choice = "triton"
+        elif nvfp4_quantize_kernel_choice == NVFP4QuantizeKernelChoice.FLASHINFER:
```
vkuzo: is this PR adding flashinfer or is that the next PR?

torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -150,13 +169,46 @@ def to_nvfp4(
...
+            assert per_tensor_scale is not None, (
+                "flashinfer nvfp4_quantize requires per_tensor_scale"
+            )
+            assert is_swizzled_scales, (
+                "flashinfer nvfp4_quantize only supports swizzled scales"
+            )
+            # flashinfer uses global_sf = (F8E4M3_MAX * F4_E2M1_MAX) / amax
+            # which is 1 / per_tensor_scale
+            global_sf = 1.0 / per_tensor_scale
+            data_lp, blockwise_scales = flashinfer_nvfp4_quantize(
```
vkuzo: are `data_lp` and `blockwise_scales` bitwise equivalent to the torch and triton paths?

torchao/prototype/mx_formats/inference_workflow.py
```diff
@@ -185,7 +186,7 @@ class NVFP4DynamicActivationNVFP4WeightConfig(AOBaseConfig):
     set to False.
 
     Configuration parameters:
-    - use_triton_kernel: bool, whether to use fused triton kernel for activation scaling (default: True)
+    - nvfp4_quantize_kernel_choice: NVFP4QuantizeKernelChoice, kernel preference for quantization (default: NVFP4QuantizeKernelChoice.TRITON)
```
vkuzo: I think we should keep the boolean for BC for one more torchao release, all of our code examples use it and there is already non-trivial OSS usage.  How about we keep it for now, throw a warning if the user sets it, and delete it in the next release?

torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -32,18 +33,36 @@
...
-from torchao.utils import TorchAOBaseTensor, fill_defaults
+from torchao.utils import TorchAOBaseTensor, _is_flashinfer_available, fill_defaults
 
 E4M3_EPS = torch.finfo(torch.float8_e4m3fn).tiny
 
 aten = torch.ops.aten
 
 
+class NVFP4QuantizeKernelChoice(str, Enum):
+    """Enum for specifying the kernel used for NVFP4 quantization."""
```
vkuzo: nit: make this more specific to explain what exactly this kernel is doing, "nvfp4 quantization" is correct but ambiguous

torchao/prototype/mx_formats/inference_workflow.py
jerryzh168: makes sense, will add this back

torchao/prototype/mx_formats/nvfp4_tensor.py
jerryzh168: sg. btw, I saw `block_size: Block size for quantization (must be 16)`, is this true? why do we make this an argument if it has to be fixed?

torchao/prototype/mx_formats/nvfp4_tensor.py
jerryzh168: ah sorry, this should be in the next PR

torchao/prototype/mx_formats/nvfp4_tensor.py
jerryzh168: these are not bitwise equivalent I think, tested in next PR

torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -39,12 +41,63 @@
 aten = torch.ops.aten
 
 
+class NVFP4QuantizeKernelChoice(str, Enum):
```
vkuzo: put in `torchao/prototype/mx_formats/constants.py` (or a similar file if already exists)

torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -61,7 +114,7 @@ class NVFP4Tensor(TorchAOBaseTensor):
         block_size (int): Block size for quantization (fixed at 16)
         orig_dtype (torch.dtype): Original tensor dtype before quantization
         is_swizzled_scales (bool): Whether scales are stored in swizzled (blocked) format
-        use_triton_kernel (bool): Whether to use triton kernels
+        nvfp4_quantize_kernel_choice (NVFP4QuantizeKernelChoice): Kernel preference for quantization
```
vkuzo: `quantize_kernel_choice`? can be consistent for similar functionality in other workflows

or, `quantize_to_nvfp4_kernel_choice`? clearer name

torchao/prototype/qat/nvfp4.py
```diff
@@ -23,12 +31,21 @@ class NVFP4FakeQuantizeConfig(FakeQuantizeConfigBase):
...
     use_per_tensor_scale: bool = True
     use_swizzled_scales: bool = False
-    use_triton_kernel: bool = False
+    nvfp4_quantize_kernel_choice: NVFP4QuantizeKernelChoice = (
+        NVFP4QuantizeKernelChoice.TORCH
+    )
+    use_triton_kernel: Optional[bool] = None
+
+    def __post_init__(self):
+        self.nvfp4_quantize_kernel_choice = _handle_use_triton_kernel(
```
vkuzo: I think this should throw an exception if the user specified `use_triton_kernel=True` and anything other than kernel preference triton, and do nothing else.  Setting `self.use_triton_kernel` to None is confusing here, let's just enforce everything is consistent and keep it.

jerryzh168: this would mean bc-breaking for current callsites, I was planning to not break bc in this PR, and then refactor all OSS and internal callsites, and then break bc, does that sound OK?

torchao/prototype/mx_formats/nvfp4_tensor.py
jerryzh168: OK will change to `quantize_to_nvfp4_kernel_choice` for now

can rename to `quantize_kernel_choice` later when there are similar cases I think

torchao/prototype/qat/nvfp4.py
vkuzo: the BC should not break as long as `use_triton_kernel` and `nvfp4_quantize_kernel_choice` both default to using the triton kernel

jerryzh168: > I think this should throw an exception if the user specified use_triton_kernel=True and anything other than kernel preference triton, and do nothing else.

should this ignore the case of `use_triton_kernel=False` and kernel_choice triton?

> the BC should not break as long as use_triton_kernel and nvfp4_quantize_kernel_choice both default to using the triton kernel

what about user setting `use_triton_kernel=False`?

vkuzo: you can throw an exception if the two vars do not match

torchao/prototype/mx_formats/nvfp4_tensor.py
andrewor14: It's just the specification: https://developer.nvidia.com/blog/introducing-nvfp4-for-efficient-and-accurate-low-precision-inference

jerryzh168: yeah confirmed with Vasiliy that we can remove this arg in the future

jerryzh168:
seems mslk kernels can give us similar performance as flashinfer kernel, this is no longer needed

PR #3912, opened 2026-02-17
Add support for flashinfer quantize kernel option for nvfp4

jerryzh168:
* __->__ #3912
* #3911

Summary:
Added the flashinfer option for better performance on some of the workflow
we are interested in, also added numerical equivalence test between different
nvfp4_quantize_kernel_choice options

Test Plan:

```
pip install flashinfer-python
```
pytest test/prototype/mx_formats/test_nvfp4_tensor.py -k test_kernel_preference_numerical_equivalence

perf test: https://github.com/pytorch/ao/pull/4031

We'll test speedup a bit later

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/benchmarks/float8/float8_inference_roofline.py
+++ b/benchmarks/float8/float8_inference_roofline.py
@@ -45,6 +45,7 @@ from torchao.utils import torch_version_at_least
 # ScalingType and SwizzleType are only available in PyTorch 2.10+
 if torch_version_at_least("2.10.0"):
     from torch.nn.functional import ScalingType, SwizzleType
+from torchao.prototype.mx_formats.config import QuantizeToNVFP4KernelChoice
 from torchao.prototype.mx_formats.inference_workflow import (
     MXDynamicActivationMXWeightConfig,
     NVFP4DynamicActivationNVFP4WeightConfig,
@@ -796,13 +797,16 @@ def run(
                 elif recipe_name == "nvfp4":
                     config = NVFP4DynamicActivationNVFP4WeightConfig(
                         use_dynamic_per_tensor_scale=True,
+                        quantize_to_nvfp4_kernel_choice=QuantizeToNVFP4KernelChoice.FLASHINFER,
                     )
                 elif recipe_name == "nvfp4_static":
                     config_calib = NVFP4DynamicActivationNVFP4WeightConfig(
                         step="prepare",
+                        quantize_to_nvfp4_kernel_choice=QuantizeToNVFP4KernelChoice.FLASHINFER,
                     )
                     config = NVFP4DynamicActivationNVFP4WeightConfig(
                         step="convert",
+                        quantize_to_nvfp4_kernel_choice=QuantizeToNVFP4KernelChoice.FLASHINFER,
                     )
                 else:
                     assert False, "unsupported"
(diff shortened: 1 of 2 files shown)


torchao/quantization/quantize_/common/kernel_preference.py
```diff
@@ -34,6 +34,10 @@ class KernelPreference(str, Enum):
     """Use triton kernels
     """
 
+    FLASHINFER = "flashinfer"
```
vkuzo: I don't think this PR is using `KernelPreference` in the same way as other workflows are.  In this PR we mean "use FlashInfer for the quantize kernel", to follow design of `KernelPreference` the logic in this branch should be "use FlashInfer kernels for quantize and the gemm kernels".

Shorter term I think making the boolean for the quantize kernel into an enum would sound better to me than reusing `KernelPreference` in this way.

Longer term I think we should reconsider `KernelPreference`, there are more differences than similarities between different workflows.

jerryzh168: OK yeah sounds good

yeah it seems kernel preference, packing format are all not really reused

test/prototype/mx_formats/test_nvfp4_tensor.py
```diff
@@ -368,8 +369,8 @@ def test_nvfp4_swizzled_scales_get_scales_method():
     not is_sm_at_least_100(), reason="requires sm100+ for raw intrinsics"
 )
 @torch.no_grad()
-def test_triton_nvfp4_quantize_equivalence(M, N, use_per_tensor_scale, dtype):
-    """Test that Triton and PyTorch NVFP4 quantization produce equivalent results."""
+def test_kernel_choice_numerical_equivalence(M, N, use_per_tensor_scale, dtype):
```
vkuzo: clarify this is for quantization from bf16 to nvfp4

test/prototype/mx_formats/test_nvfp4_tensor.py
```diff
@@ -378,40 +379,52 @@ def test_triton_nvfp4_quantize_equivalence(M, N, use_per_tensor_scale, dtype):
...
+            x.clone(),
+            per_tensor_scale=per_tensor_scale,
+            is_swizzled_scales=True,
+            nvfp4_quantize_kernel_choice=kc,
+        )
+
+        # For kernel choices that use the same quantization algorithm as TORCH
+        # (TRITON should be bitwise identical), verify internal data matches exactly
+        if kc == NVFP4QuantizeKernelChoice.TRITON:
+            torch.testing.assert_close(
```
vkuzo, line 413: add check to ensure bitwise match

vkuzo:
Can we include performance benchmarking in this PR?
1. roofline script with sweep over shapes: https://github.com/pytorch/ao/blob/main/benchmarks/float8/float8_inference_roofline.py
2. nvfp4 individual cast bench: https://github.com/pytorch/ao/blob/f8bbce54eff2993e950ccb724022068e419ce793/benchmarks/mx_formats/cast_bench.py#L96

test/prototype/mx_formats/test_nvfp4_tensor.py
jerryzh168: will rename to `test_quantize_to_nvfp4_kernel_numerical_equivalence`

original test is testing both fp32 and bf16, please let me know if you feel we should remove fp32 test case as well

jerryzh168:
seems mslk kernels can give us similar performance as flashinfer kernel, this is no longer needed

PR #3916, opened 2026-02-19
Add support for `torch._higher_order_ops.while_loop` in pt2e quantization

jerryzh168:
* __->__ #3916

Summary:

- Add support for torch._higher_order_ops.while_loop in pt2e quantization by recursively applying prepare_pt2e and convert_pt2e to the while_loop op's body_fn subgraph (arg index 1) using the same quantizer from
  the top-level module
- Register while_loop's body_fn subgraph in _get_control_flow_submodules alongside existing cond, map_impl, and scan handlers
- In convert_pt2e, process while_loop subgraphs before the main conversion passes to prevent DuplicateDQPass from encountering stale observer references during recursive graph linting
- Only the body_fn subgraph is quantized; the cond_fn subgraph (arg index 0) is not quantized

Test Plan:

- Added test_while_loop_op_quantization which exports a model using while_loop with a linear layer in the body_fn, runs prepare_pt2e + calibration + convert_pt2e, and verifies observers are inserted after
prepare, q/dq ops are present after convert, and the quantized model produces correct output (SQNR 42.8 dB > 35 dB threshold)
- Verified existing test_scan_op_quantization still passes

Summary:

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/quantization/pt2e/test_quantize_pt2e.py
+++ b/test/quantization/pt2e/test_quantize_pt2e.py
@@ -3349,6 +3349,123 @@ class TestQuantizePT2E(PT2EQuantizationTestCase):
         sqnr = compute_error(output_ref, output)
         self.assertGreater(sqnr, 35, f"SQNR too low: {sqnr} dB")
 
+    def test_while_loop_op_quantization(self):
+        """Test that prepare_pt2e and convert_pt2e correctly quantize ops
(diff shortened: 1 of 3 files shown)


Ninja91:
Awesome!

PR #3930, opened 2026-02-21
Added benchmark for LLaMA 3 model for attention tests

howardzhang-cv:
* #3947
* #3960
* __->__ #3930
* #3929
* #3865
* #3857
* #3959

## Summary
- Added new benchmark for new low precision attention API
- uses Llama3-8b model, prefill
- 2-phase test
    - perplexity using the WikiText-2 test set
    - runtime over different sequence lengths (1024 -> 131072)
- Can set baseline and test models between different backends: (fa2, fa3, fa3_fp8, fa4, fa4_fp8)
- has options to control torch.compile usage, warmup_iters, sequence length, number of runtime test iterations, rope fusion

## Example Run
python benchmarks/prototype/attention/eval_llama3_model.py --baseline fa3 --test fa3_fp8

### DIFF
--- a/test/prototype/attention/test_fp8_attention.py
+++ b/test/prototype/attention/test_fp8_attention.py
@@ -20,20 +20,20 @@ from torchao.utils import torch_version_at_least
 if torch_version_at_least("2.11.0"):
     from torchao.prototype.attention.utils import _is_fa3_available, _is_hopper
 
-if _is_hopper() and _is_fa3_available():
-    from torch.nn.attention import (
-        activate_flash_attention_impl,
-        restore_flash_attention_impl,
-    )
+    if _is_hopper() and _is_fa3_available():
+        from torch.nn.attention import (
+            activate_flash_attention_impl,
+            restore_flash_attention_impl,
+        )
 
-    from torchao.prototype.attention import (
-        AttentionBackend,
-        apply_low_precision_attention,
-    )
-    from torchao.prototype.attention.fp8_fa3.attention import (
-        fp8_fa3_rope_sdpa,
-        fp8_fa3_sdpa,
-    )
+        from torchao.prototype.attention import (
+            AttentionBackend,
+            apply_low_precision_attention,
+        )
+        from torchao.prototype.attention.fp8_fa3.attention import (
+            fp8_fa3_rope_sdpa,
+            fp8_fa3_sdpa,
+        )
 
 
 def _rope_cos_sin(S, D, device):


benchmarks/prototype/attention/eval_llama3_model.py
```diff
@@ -0,0 +1,440 @@
...
+            restore_flash_attention_impl()
+        model.config._attn_implementation = old_attn_impl
+
+
+# =============================================================================
+# Helpers
+# =============================================================================
+
+
+def load_wikitext2_tokens(tokenizer, seq_len: int):
```
namgyu-youn: We don't have to define wiki-text tokenizer here. Instead, we can run `lm-eval` directly like

https://github.com/pytorch/ao/blob/450699a26c2b3469a31741b4e70b93a439459b5b/benchmarks/quantization/calibration_based/create_quantized_model.py#L30-L37

benchmarks/prototype/attention/eval_llama3_model.py
```diff
@@ -0,0 +1,440 @@
...
+    chunks = tokens.view(n_chunks, seq_len + 1)
+
+    print(
+        f"  WikiText-2 test: {tokens.numel()} tokens, "
+        f"{n_chunks} chunks of {seq_len} tokens"
+    )
+    return chunks
+
+
+def compute_perplexity(model, chunks, device: str, backend_name: str) -> float:
```
namgyu-youn: This is also not needed. See above comment

benchmarks/prototype/attention/eval_llama3_model.py
```diff
@@ -0,0 +1,440 @@
...
+            loss = F.cross_entropy(
+                logits.view(-1, logits.size(-1)), labels.reshape(-1)
+            )
+            total_loss += loss.item()
+
+    avg_loss = total_loss / n_chunks
+    return math.exp(avg_loss)
+
+
+def benchmark_runtime(
```
namgyu-youn, line 176: Q. Can we compute forward pass latency using vLLM directly, similar to e2e?
https://github.com/pytorch/ao/blob/3d3dd506fe375d18ac79ac1c6a4a00e6ff3f529b/benchmarks/quantization/measure_accuracy_and_performance.sh#L176

benchmarks/prototype/attention/eval_llama3_model.py
```diff
@@ -0,0 +1,440 @@
...
+    "fa4": {
+        "hf_attn_impl": "sdpa",
+        "flash_impl": "FA4",
+        "fp8": False,
+        "label": "FA4 BF16",
+    },
+}
+
+RANDOM_SEED = 42
+DEFAULT_MODEL_ID = "meta-llama/Llama-3.1-8B"
```
namgyu-youn: What does `DEFAULT_MODEL_ID` do? Should it be called by args (`--model_id`) default directly?

benchmarks/prototype/attention/eval_llama3_model.py
howardzhang-cv: Thanks! I changed it to use this instead

benchmarks/prototype/attention/eval_llama3_model.py
howardzhang-cv: Changed

benchmarks/prototype/attention/eval_llama3_model.py
howardzhang-cv: Unfortunately not. Unlike the other quantization APIs in TorchAO, the low precision attention path requires replacing F.scaled_dot_product_attention with a specific attention backend capable of low precision attention (e.g. FA3/4). So we need to ensure that our model calls F.SDPA.

benchmarks/prototype/attention/eval_llama3_model.py
howardzhang-cv: Fixed

benchmarks/prototype/attention/eval_llama3_model.py
```diff
@@ -0,0 +1,534 @@
...
+        "flash_impl": "FA3",
+        "fp8": True,
+        "fp8_backend": AttentionBackend.FP8_FA3,
+        "label": "FA3 FP8",
+    },
+}
+
+RANDOM_SEED = 42
+
+SEQ_LENGTHS = [1024, 2048, 4096, 8192, 16384, 32768]  # , 65536, 131072]
```
namgyu-youn: Q. Are you considering both short and long (> 32k) SeqLen? If so, how about using `longbench_single_e` task  instead of `wiki-text` for long SeqLen? btw, I remember torchao is using wikitext and winogrande only (no reasoning task yet).

Maybe we can separate for small PRs like: (1) short-context only (1024-4096) in this PR, (2) long-context (> 32k) with `longbench_single_e` task support.

benchmarks/prototype/attention/eval_llama3_model.py
```diff
@@ -0,0 +1,534 @@
...
+# Helpers
+# =============================================================================
+
+
+def evaluate_perplexity(model, tokenizer, flash_impl) -> float:
+    """Evaluate perplexity on WikiText-2 using lm_eval.
+
+    For FP8 backends, the model wrapper handles flash activation internally.
+    For non-FP8 backends, flash activation is set for the duration of the
+    evaluation so that lm_eval's internal forward calls use the correct kernel.
```
namgyu-youn: Perplexity isn't related to FP8/FA. We can drop Ln206-Ln208 I feel

benchmarks/prototype/attention/eval_llama3_model.py
```diff
@@ -0,0 +1,534 @@
...
+    # --- Baseline runtime (all sequence lengths) ---
+    print(f"\n  Running baseline ({baseline_label})...")
+    baseline_model, baseline_flash = setup_backend(
+        orig_model,
+        baseline_backend,
+        compile,
+        fuse_rope=fuse_rope,
+    )
+    baseline_runtimes = {}
+    for S in SEQ_LENGTHS:
```
namgyu-youn: I think adding SeqLen to args without this loop would be safer for parallel execution and better oom handling (no try-except).

benchmarks/prototype/attention/eval_llama3_model.py
howardzhang-cv: This sounds like a good plan. For this PR, it's just to focus on validating that the low-precision attention path works. But it would definitely be good to get signal on if the quality worsens on longer sequence lengths as well. I will plan a separate PR to add to this benchmark file.

benchmarks/prototype/attention/eval_llama3_model.py
howardzhang-cv: Dropped

benchmarks/prototype/attention/eval_llama3_model.py
howardzhang-cv: Yes this makes sense, I added it as an argument instead of being a hard-coded list

benchmarks/prototype/attention/eval_llama3_model.py
namgyu-youn: This block still exists :)
```
For FP8 backends, the model wrapper handles flash activation internally.
For non-FP8 backends, flash activation is set for the duration of the
evaluation so that lm_eval's internal forward calls use the correct kernel.
```

PR #3945, opened 2026-02-25
Use relaxed memory ordering for Triton atomics on AMDGPU.

wenchenvincent:
`tl.atomic_add` and the like needs to use relaxed memory ordering on AMDGPU for performance. When the default acquire-release semantic is used, memory fence will be inserted before and after the atomics op and thus hurts performance. Such memory fences are not necessary for the functionality of `atomic_add` and the like (they usually required for `tl.atomic_xchg`).

Here is an example result when running the benchmarking for 
`python benchmarks/prototype/moe_training/fp8_rowwise/bench_triton_fp8_rowwise_3d_transpose_rhs.py`

Before the changes:
image
After the change:
image

### DIFF
--- a/torchao/prototype/hqq/kernels.py
+++ b/torchao/prototype/hqq/kernels.py
@@ -3,6 +3,7 @@
 #
 # This source code is licensed under the BSD 3-Clause license found in the
 # LICENSE file in the root directory of this source tree.
+import torch
 import triton
 import triton.language as tl
 from triton import Config
@@ -389,7 +390,11 @@ def _mixed_mm_kernel(
     if SPLIT_K == 1:
         tl.store(C, acc, mask=mask)
     else:
-        tl.atomic_add(C, acc, mask=mask)
+        # AMD GPUs need relaxed semantics for better performance
+        if tl.constexpr(torch.version.hip is not None):
+            tl.atomic_add(C, acc, mask=mask, sem="relaxed")
+        else:
+            tl.atomic_add(C, acc, mask=mask)
 
 
 _mixed_mm = triton.heuristics(MIXED_MM_HEURISTICS)(_mixed_mm_kernel)
--- a/torchao/prototype/common/triton/matmul.py
+++ b/torchao/prototype/common/triton/matmul.py
@@ -259,7 +259,11 @@ def _kernel(
     if SPLIT_K == 1:
         tl.store(C, acc, mask=mask)
     else:
-        tl.atomic_add(C, acc, mask=mask)
+        # AMD GPUs need relaxed semantics for better performance
+        if tl.constexpr(torch.version.hip is not None):
+            tl.atomic_add(C, acc, mask=mask, sem="relaxed")
+        else:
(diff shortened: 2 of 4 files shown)


danielvegamyhre:
Interesting, I'm surprised the performance impact is so outsized on AMD but not CUDA. I just tested these changes locally on b200 and the results were virtually identical. 

Thanks for improving this! 

Please use `ruff check --fix <dirs>` and `ruff format <dirs>` to fix the linter error

torchao/prototype/hqq/kernels.py
```diff
@@ -152,15 +153,19 @@ def init_to_zero(name):
...
-    "BLOCK_K": lambda args: min(args["BLOCK_K"], args["QGROUP_SIZE"])
-    if not args["TRANSPOSED"]
-    else args["BLOCK_K"],
-    "BLOCK_N": lambda args: min(args["BLOCK_N"], args["QGROUP_SIZE"])
-    if args["TRANSPOSED"]
-    else args["BLOCK_N"],
-    "SPLIT_K": lambda args: 1
-    if args["IS_BFLOAT16"]
-    else args["SPLIT_K"],  # atomic add not supported for bfloat16
+    "BLOCK_K": lambda args: (
```
danielvegamyhre: odd this formatting was changed, is your ruff version freshly pip installed from requirements.txt

wenchenvincent: I directly installed ruff with pip without installing it from dev-requirements.txt. It seems that the ruff version that I used was 0.15.2, which was newer than 0.11.6. Now I reintalled ruff 0.11.6 and rerun formatting on this file but it didn't make it to the original format. Let me know what you would like to do it. Shall I revert to the original format or can we keep this format change as is?

danielvegamyhre: I see yeah just revert the formattting changes to this file please

wenchenvincent: Done.

danielvegamyhre:
@wenchenvincent looks like there is still an issue:

```
torchao/prototype/hqq/kernels.py:393:25: F821 Undefined name `torch`
    |
391 |     else:
392 |         # AMD GPUs need relaxed semantics for better performance
393 |         if tl.constexpr(torch.version.hip is not None):
    |                         ^^^^^ F821
394 |             tl.atomic_add(C, acc, mask=mask, sem="relaxed")
395 |         else:
    |
    ```

wenchenvincent:
> @wenchenvincent looks like there is still an issue:
> 
> ```
> torchao/prototype/hqq/kernels.py:393:25: F821 Undefined name `torch`
>     |
> 391 |     else:
> 392 |         # AMD GPUs need relaxed semantics for better performance
> 393 |         if tl.constexpr(torch.version.hip is not None):
>     |                         ^^^^^ F821
> 394 |             tl.atomic_add(C, acc, mask=mask, sem="relaxed")
> 395 |         else:
>     |
>     ```
> ```

Sorry, my bad! Had an omission when reverting and adding the changes. It should be fixed now.

PR #3946, opened 2026-02-25
Add RunningAbsMaxSmoothQuantObserver for memory-efficient calibration (#3946)

jcaip:
Summary:

Add a memory-efficient SmoothQuant observer that uses running per-channel
absmax instead of storing all calibration inputs. This reduces calibration
memory from O(N x features) to O(features), preventing RAM spikes and OOM
kills when calibrating on large datasets.

- Add RunningAbsMaxSmoothQuantObserver class in core.py
- Add use_running_absmax config option to SmoothQuantConfig
- Export the new observer from the module

Differential Revision: D94260071

### DIFF
--- a/torchao/quantization/quantize_/common/quantization_step.py
+++ b/torchao/quantization/quantize_/common/quantization_step.py
@@ -49,9 +49,13 @@ class QuantizationStep(str, Enum):
     After this step, run calibration data through the model again to collect
     smoothed activation statistics for accurate activation scale computation.
     """
-    PREPARE_FOR_SMOOTHQUANT_SMOOTHING_FACTOR = "prepare_for_smoothquant_smoothing_factor"
+    PREPARE_FOR_SMOOTHQUANT_SMOOTHING_FACTOR = (
+        "prepare_for_smoothquant_smoothing_factor"
+    )
 
     """Compute activation scales from smoothed activation statistics collected
     during the second calibration pass (after PREPARE_FOR_SMOOTHQUANT_SMOOTHING_FACTOR).
     """
-    PREPARE_FOR_SMOOTHQUANT_ACTIVATION_SCALES = "prepare_for_smoothquant_activation_scales"
+    PREPARE_FOR_SMOOTHQUANT_ACTIVATION_SCALES = (
+        "prepare_for_smoothquant_activation_scales"
+    )


torchao/prototype/smoothquant/tests/test_observers.py
```diff
@@ -0,0 +1,249 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
namgyu-youn: Not sure why test script is located inside package. Shouldn't we update  https://github.com/pytorch/ao/blob/main/test/prototype/test_smoothquant.py instead?

torchao/prototype/smoothquant/core.py
```diff
@@ -71,12 +71,86 @@ def calculate_qparams(self, weight_quant_kwargs=None):
             return smoothing_factor, None
 
 
+class RunningAbsMaxSmoothQuantObserver(torch.nn.Module):
```
namgyu-youn, line 78: Like this new logic overally, but do we actually need `RunningAbsMaxSmoothQuantObserver`? To me, minimal `SmoothQuantObserver` update (even though its bc-breaking) looks better.

Also, `RunningAbsMaxSmoothQuantObserver` reminds me `AQT (AffineQuantizedTensor)` nightmare. I am not sure if we really need abstracted notation.

torchao/prototype/smoothquant/tests/test_observers.py
jcaip: Yes, will tell Hossein to do that before merging

torchao/prototype/smoothquant/core.py
jcaip: Yeah agree SmoothQuantObserver abstraction is kind of funky. 

I think the best way this mess is to define a ObserverTensor class that we can then modify appropriately. That way we don't need to have SmoothQuantObserver, SmoothQuantObservedLinear and can just use a single SmoothQuantObserverTensor. 

Do you know if smoothquant API is used much anywhere? The only thing that concerns me is that by default replacing SmoothQuantObserver with RunningAbsMaxSmoothQuantObserver is that we need two passes through the data for RunningAbsMaxSmoothQuantObserver, so existing workflows might fail. 

We do need a way to test numerics though between SmoothQuantObserver and RunningAbsMaxSmoothQuantObserver, so even if we make RunningAbsMaxSmoothQuantObserver we'll still need to store this code somewhere. I think it's better to leave the cleanup as a follow up.

namgyu-youn: > I think the best way this mess is to define a ObserverTensor class that we can then modify appropriately. That way we don't need to have SmoothQuantObserver, SmoothQuantObservedLinear and can just use a single SmoothQuantObserverTensor.

So are you considering common observer class? More generally, quantization subclass (`Float8Tensor/Int8Tensor`) should not contain observer-related ops (like `act_pre_scale`)? Actually I was working for this topic, so it would be great if you can look into it — #3925.

namgyu-youn: > Do you know if smoothquant API is used much anywhere? The only thing that concerns me is that by default replacing SmoothQuantObserver with RunningAbsMaxSmoothQuantObserver is that we need two passes through the data for RunningAbsMaxSmoothQuantObserver, so existing workflows might fail.

Yeah I am also interested in this topic. For example, because AWQ uses greedy search

https://github.com/pytorch/ao/blob/5045d76db86090b8936e1a3630ed52df915ea8bc/torchao/prototype/awq/core.py#L70-L86

, we can try better searching algorithm like binary/tenary. It's GPTQ, but I already observed tenary is 2x faster than greedy: https://github.com/ModelCloud/GPTQModel/pull/2419

Wondering if you have any candidates for these numerical techniques, what about your thought?

namgyu-youn: > We do need a way to test numerics though between SmoothQuantObserver and RunningAbsMaxSmoothQuantObserver, so even if we make RunningAbsMaxSmoothQuantObserver we'll still need to store this code somewhere. I think it's better to leave the cleanup as a follow up.

I think in this early stage, we don't need multiple-configs to store all of them. Also I don't want to see `RunningAbsMaxSmoothQuantObserver` anywhere.

Instead, we can just run perf using https://github.com/pytorch/ao/blob/main/benchmarks/quantization/calibration_based/measure_accuracy_and_performance.sh and update minimally across AWQ/GPTQ/SmoothQuant I feel.

jcaip:
@claude can you fix lint please

namgyu-youn:
@jcaip @jerryzh168 do you have any GPU profiling e2e results for this PR? If I remember correctly, there wasn't a profiling benchmark at the time — just wanted to check if that's been added or if there are any benchmark numbers available I am missing.

PR #3359, opened 2025-11-20
[ci][aarch64] add Linux aarch64 wheels

puneetmatharu:
Adds support for also building Linux AArch64 wheels; currently `ao` only releases Linux x86 wheels.

PR #3954, opened 2026-02-26
Patch nightly Linux AArch64-specific wheel build

puneetmatharu:
## Summary
This PR fixes the nightly Linux AArch64 wheel pipeline so we reliably produce a manylinux wheel with the expected CPU C++ kernels (including the KleidiAI path) and correct runtime linkage for OpenMP. It also fixes runtime linkage issues that caused kernel registration failures at test time.

## Underlying problem
The AArch64 nightly wheel path had four issues:

1. AArch64 CPU kernel build flags were not fully enabled in the wheel build environment.
1. Post-build `auditwheel` logic handled CUDA wheels but not general non-pure AArch64 wheels.
1. OpenMP runtime linking on AArch64 assumed `libomp`, but AArch64 builds of `torch` typically come with `libgomp.so.1`.
1. Runtime symbol resolution for `cpuinfo` could become inconsistent unless the link order was controlled, leading to wrong ISA feature visibility during ukernel selection (resulting in the error `RuntimeError: ukernel_config did not register`).

## What this PR changes

1. `packaging/env_var_script_linux.sh`:

   - Enables AArch64-specific build flags:
     - `USE_CPP=1`
     - `USE_CPU_KERNELS=1`
     - `TORCHAO_BUILD_CPU_AARCH64=1`
     - `TORCHAO_BUILD_KLEIDIAI=1`
     - `TORCHAO_ENABLE_ARM_NEON_DOT=1`
     - `BUILD_TORCHAO_EXPERIMENTAL=1`
   - Adds explicit logging of enabled flags for build transparency.

1. `packaging/post_build_script.sh`:

   - Generalizes auditwheel handling to any non-pure wheel (not CUDA-only).
   - Selects platform dynamically by arch (manylinux_2_28_x86_64 / manylinux_2_28_aarch64).
   - Skips none-any wheels and repair binary wheels consistently.

1. `torchao/csrc/cpu/shared_kernels/Utils.cmake`:

   - Removes duplicate Torch-linking from backend helper to avoid multi-point linking.
   - Improves OpenMP runtime resolution on AArch64 by preferring runtime available in torch lib dir:
     - `gomp/libgomp` first, then `omp/libomp`, with fallback preserved.

1. `torchao/csrc/cpu/CMakeLists.txt`:

   - Strictly speaking, this isn't a change but a (desired) knock-on effect of updating `Utils.cmake` (above):
     - Keep linking to `cpuinfo` before `TORCH_LIBRARIES` for `torchao_ops_aten`; This makes symbol resolution deterministic and avoids CPU feature detection mismatches that can lead to `ukernel_config did not register`.

## Validation

1. Built wheel in Linux AArch64 manylinux environment.
1. Installed built wheel in clean venv with nightly `torch`.
1. Verified binary wheel contents and kernel extension loading.
1. Ran intx opaque tensor accuracy tests against installed wheel:
   ```bash
   python3 -m unittest test_intx_opaque_tensor.py -k test_accuracy
   ```
   **RESULT:** all 8 test_accuracy cases pass for `OPAQUE_TORCHAO_KLEIDIAI`.

### DIFF
--- a/packaging/env_var_script_linux.sh
+++ b/packaging/env_var_script_linux.sh
@@ -20,7 +20,7 @@ if [[ ${CU_VERSION:-} == "cu124" ]]; then
 fi
 
 # Enable C++ kernels + kleidiai in aarch64 build
-if [[ $(uname -m) == "aarch64" ]]; then
+if [[ "$(uname -m)" =~ "^(aarch64|arm64)$" ]]; then
     echo "Enabling aarch64-specific build"
     export USE_CPP=1
     export USE_CPU_KERNELS=1
--- a/torchao/csrc/cpu/CMakeLists.txt
+++ b/torchao/csrc/cpu/CMakeLists.txt
@@ -197,6 +197,7 @@ endif()
 
 # Build ExecuTorch ops
 if(TORCHAO_BUILD_EXECUTORCH_OPS)
+    find_package(Torch REQUIRED)
     # ExecuTorch package is not required, but EXECUTORCH_INCLUDE_DIRS and EXECUTORCH_LIBRARIES must
     # be defined and EXECUTORCH_LIBRARIES must include the following libraries installed by ExecuTorch:
     # libexecutorch.a
@@ -229,4 +230,5 @@ if(TORCHAO_BUILD_EXECUTORCH_OPS)
         endif()
     endif()
     target_link_libraries(torchao_ops_executorch PRIVATE cpuinfo)
+    target_include_directories(torchao_ops_executorch PRIVATE "${TORCH_INCLUDE_DIRS}")
 endif()


puneetmatharu:
@huydhn / @jerryzh168 this PR follows on from the previous one, https://github.com/pytorch/ao/pull/3359. Would you be able to review it?

puneetmatharu:
Bump @huydhn / @jerryzh168.

torchao/csrc/cpu/shared_kernels/Utils.cmake
```diff
@@ -16,12 +17,25 @@ function(target_link_torchao_parallel_backend target_name torchao_parallel_backe
...
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${_OMP_CXX_COMPILE_FLAGS}" PARENT_SCOPE)
 
-        find_package(Torch REQUIRED)
-        include_directories("${TORCH_INCLUDE_DIRS}")
-        target_link_libraries(${target_name} PRIVATE "${TORCH_LIBRARIES}")
-
         target_compile_definitions(${target_name} PRIVATE TORCHAO_PARALLEL_ATEN=1 AT_PARALLEL_OPENMP=1 INTRA_OP_PARALLEL=1)
-        target_link_libraries(${target_name} PRIVATE ${TORCH_INSTALL_PREFIX}/lib/libomp${CMAKE_SHARED_LIBRARY_SUFFIX})
+
+        if(CMAKE_SYSTEM_PROCESSOR MATCHES "^(aarch64|arm64)$")
```
nikhil-arm, line 21:
```suggestion
        if(CMAKE_SYSTEM_NAME STREQUAL "Linux" AND CMAKE_SYSTEM_PROCESSOR MATCHES "^(aarch64|arm64)$")
```

packaging/env_var_script_linux.sh
```diff
@@ -19,8 +19,19 @@ if [[ ${CU_VERSION:-} == "cu124" ]]; then
   TORCH_CUDA_ARCH_LIST="${TORCH_CUDA_ARCH_LIST};9.0"
 fi
 
-# Enable kleidiai for aarch64 build
+# Enable C++ kernels + kleidiai in aarch64 build
 if [[ $(uname -m) == "aarch64" ]]; then
```
nikhil-arm:
```suggestion
if [[ "$(uname -s)" == "Linux" && "$(uname -m)" == "aarch64" ]]; then
```

packaging/env_var_script_linux.sh
```diff
@@ -1,5 +1,5 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates.
-# Copyright (c) 2025 Arm Limited and/or its affiliates.
+# Copyright (c) 2025-2026 Arm Limited and/or its affiliates.
```
nikhil-arm, line 2:
```suggestion

```

packaging/post_build_script.sh
```diff
@@ -1,34 +1,51 @@
 #!/bin/bash
 # Copyright (c) Meta Platforms, Inc. and affiliates.
+# Copyright (c) 2026 Arm Limited and/or its affiliates.
```
nikhil-arm:
```suggestion
```

torchao/csrc/cpu/shared_kernels/Utils.cmake
```diff
@@ -1,4 +1,5 @@
 # Copyright (c) Meta Platforms, Inc. and affiliates.
+# Copyright (c) 2026 Arm Limited and/or its affiliates.
```
nikhil-arm:
```suggestion
```

torchao/csrc/cpu/shared_kernels/Utils.cmake
nikhil-arm: I am suggesting to guard against MACOS CPU

packaging/env_var_script_linux.sh
nikhil-arm: I am suggesting to guard against MACOS CPU

torchao/csrc/cpu/shared_kernels/Utils.cmake
puneetmatharu: ~Excellent point and I agree. I can change this now.~

malfet: Why restrict it to Linux? AO's optimizations should work on MacOS as well (this is what arm64 is for)

packaging/env_var_script_linux.sh
malfet: I don't think this is needed, as aarch64 is Linux only (`uname -m` on Darwin return `arm64`)

torchao/csrc/cpu/shared_kernels/Utils.cmake
nikhil-arm: Good Point. I was trying to restrict the macOS failures in CI.

I think the macOS CI failure is caused by the newly enabled ExecuTorch build path on arm64, which pulls in ExecuTorch headers that depend on PyTorch c10 headers, but the torchao_ops_executorch target on macOS is not getting the required Torch include path, so compilation fails on c10/util/irange.h.

We will investigate this and see if we can correctly enable it for darwin arm64

malfet: But I don't think those are related, ET should have access to c10 headers regadless of the platform, shouldn't it?

[Edit] I'm wrong, it succeeds on restart

malfet (approved):

torchao/csrc/cpu/shared_kernels/Utils.cmake
malfet: Hmm, if this part is for ET integration, then I think it's wrong to make a change, because OpenMP runtime is obviously not available on lot of ET targets

packaging/env_var_script_linux.sh
puneetmatharu: I think this might be _a_ problem (not necessarily _the_ source of the failing tests). The CMake update to `Utils.cmake` catches Linux AND macOS but this only catches Linux. I think these build variables should be used for both...

torchao/csrc/cpu/shared_kernels/Utils.cmake
puneetmatharu: This change is solely meant to fix the `aten_openmp` backend usage on `aarch64`/`arm64`; it should leave the executorch backend route untouched.

puneetmatharu:
The failing `test-nightly` isn't due to us; it's mentioned [here](https://github.com/pytorch/ao/pull/4071).

Also can we add the "not user facing" label to this PR? (Not sure what the syntax is)

torchao/csrc/cpu/shared_kernels/Utils.cmake
```diff
@@ -16,12 +16,25 @@ function(target_link_torchao_parallel_backend target_name torchao_parallel_backe
 
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${_OMP_CXX_COMPILE_FLAGS}" PARENT_SCOPE)
 
-        find_package(Torch REQUIRED)
-        include_directories("${TORCH_INCLUDE_DIRS}")
```
puneetmatharu, line 20: Ah. I think this is the source of the issue. Previously the include dirs were blanket-added to the entire repo (which isn't good CMake anyway; it should be done on a `target_*` basis) when building the Aten ops with the `aten_openmp` parallel backend. When you get to the executorch ops, the torch dirs have already been included. I think the patch will be to find the `Torch` package and target_include the `TORCH_INCLUDE_DIRS`. Will take care of this.

puneetmatharu:
**Update:** the two `test-nightly` regression tests should pass now with https://github.com/pytorch/ao/pull/4071.

torchao/csrc/cpu/CMakeLists.txt
```diff
@@ -229,4 +230,5 @@ if(TORCHAO_BUILD_EXECUTORCH_OPS)
         endif()
     endif()
     target_link_libraries(torchao_ops_executorch PRIVATE cpuinfo)
+    target_include_directories(torchao_ops_executorch PRIVATE "${TORCH_INCLUDE_DIRS}")
```
metascroy, line 233: Why do we need torch include dirs for executorch build?

torchao/csrc/cpu/shared_kernels/Utils.cmake
```diff
@@ -16,12 +16,25 @@ function(target_link_torchao_parallel_backend target_name torchao_parallel_backe
...
-        target_link_libraries(${target_name} PRIVATE "${TORCH_LIBRARIES}")
-
         target_compile_definitions(${target_name} PRIVATE TORCHAO_PARALLEL_ATEN=1 AT_PARALLEL_OPENMP=1 INTRA_OP_PARALLEL=1)
-        target_link_libraries(${target_name} PRIVATE ${TORCH_INSTALL_PREFIX}/lib/libomp${CMAKE_SHARED_LIBRARY_SUFFIX})
+
+        if(CMAKE_SYSTEM_PROCESSOR MATCHES "^(aarch64|arm64)$")
```
metascroy, line 21: Why is this OMP logic specific to aarch64?

torchao/csrc/cpu/CMakeLists.txt
puneetmatharu: It's necessary because previously you had

```cmake
find_package(Torch REQUIRED)
include_directories("${TORCH_INCLUDE_DIRS}")
```

in `Utils.cmake` which blanked added the torch include directories to everything defined in the CMake configuration step (which isn't really how it should be done in modern CMake). Removing those lines identified a dependence of `torchao_ops_executorch` on the torch include directories (specifically for `c10` headers). It's for this reason I added this explicit line here.

torchao/csrc/cpu/shared_kernels/Utils.cmake
puneetmatharu: On AArch64 systems, `torch` wheels ship with `libgomp` not `libomp`. On top of that, there's some slight messiness because the `libgomp.so.1` isn't symlinked to `libgomp.so` so the `find_library` usage is slightly more convoluted than desired to pick it up.

metascroy:
@puneetmatharu @nikhil-arm out of curiosity, what motivated this PR? Are you planning on contributing more kleidiai kernels to torchao?

puneetmatharu:
> @puneetmatharu @nikhil-arm out of curiosity, what motivated this PR? Are you planning on contributing more kleidiai kernels to torchao?

@metascroy The main use case is the Arm [Tool Solutions](https://github.com/ARM-software/Tool-Solutions) library, which we use to build "bleeding edge" AArch64 `torch` wheels (with the latest Arm contributions across the ML stack). We currently have to build `torchao` from source to get native AArch64 wheels, which is not ideal; we'd very much like to remove this need and just `pip install torchao` to get the AArch64-specific wheel. Having this feature would also unblock us on further restructuring of the library.

Issue #3788, opened 2026-01-30
fp8 rowwise scaled_mm error: CUBLAS_STATUS_INVALID_VALUE


danielvegamyhre:
### Repro

On H100 run:
`pytest test/prototype/moe_training/test_scaled_grouped_mm.py -v -rs`

### Error

`RuntimeError: CUDA error: CUBLAS_STATUS_INVALID_VALUE when calling ::cublasLtMatmulDescSetAttribute(descriptor(), attr, &value, sizeof(value))`

jerryzh168:
I've seen these before, for me it seems to be resolved by setting `LD_LIBRARY_PATH` env variable correctly, but not sure if it's the same issue for you

danielvegamyhre:
this self resolved

PR #3956, opened 2026-02-27
Fix backward return count mismatch in _Float8GroupedMM

xiaobochen-amd:


danielvegamyhre:
surprised 1xh100 CI tests didn't catch this

xiaobochen-amd:
Before my previous PR, the counts here were mismatched.

danielvegamyhre:
i think i see the issue, we need to unskip this test: https://github.com/pytorch/ao/blob/4e18d87712a71816fe5754549b4f2bf84082e201/test/prototype/moe_training/test_scaled_grouped_mm.py#L58

xiaobochen-amd:
> i think i see the issue, we need to unskip this test:
> 
> https://github.com/pytorch/ao/blob/4e18d87712a71816fe5754549b4f2bf84082e201/test/prototype/moe_training/test_scaled_grouped_mm.py#L58

When I tested locally, it was set to unskip and the test passed. I just noticed that https://github.com/pytorch/ao/issues/3788
 hasn’t been fixed, so I left it as is.

danielvegamyhre:
> > i think i see the issue, we need to unskip this test:
> > https://github.com/pytorch/ao/blob/4e18d87712a71816fe5754549b4f2bf84082e201/test/prototype/moe_training/test_scaled_grouped_mm.py#L58
> 
> When I tested locally, it was set to unskip and the test passed. I just noticed that #3788 hasn’t been fixed, so I left it as is.

I see, let me test on h100 if the issue still exists, i think it may have been a transient env/build issue we never root caused

danielvegamyhre:
@xiaobochen-amd i tested and and the original cublas error is resolved, but there is a new error (numerics mismatch over the threshold). forward pass outputs are identical with torch.equal, however, gradients are slightly different, with most columns identical but some columns requiring atol/rtol=1 to pass.

I'll create an issue for this on the CUDA side. It doesn't block this, we can leave the test as skipped.

PR #3959, opened 2026-02-27
Add FP8 FA3 low-precision attention with monkey-patch SDPA path

howardzhang-cv:
* #3947
* #3960
* #3930
* #3929
* #3865
* #3857
* __->__ #3959

---

## Summary                                                                                                                                                                                                                                                                                                  
  - Added new folder for low-precision attention APIs in torchao/prototype/attention                                                                                                                                                                
  - New API for FP8 FA3 low-precision attention with two components:
    - Elementary block: fp8_fa3_sdpa — a direct drop-in replacement for F.scaled_dot_product_attention that users can integrate into their model manually. Performs per-head FP8 quantization of Q, K, V followed by low-precision SDPA.            
    - Simple wrapper: apply_low_precision_attention — wraps any model to automatically replace all SDPA calls with the FP8 variant. No torch.compile required.                                                                                      
  - New Triton kernel for fused QKV FP8 quantization (3-phase: absmax reduction, scale computation, quantize)
  - Causal mask detection: Pre-flight forward pass identifies HuggingFace-style materialized causal masks so the wrapper can strip them and use is_causal=True instead.
  - Flash attention activation is handled internally by the wrapper — no manual activate_flash_attention_impl / restore_flash_attention_impl calls needed.
  - Added new test folder for low-precision attention APIs in test/prototype/attention

### Folder Breakdown
  - torchao/prototype/attention: new folder for low-precision attention APIs
    - __init__.py: public exports (apply_low_precision_attention, AttentionBackend, LowPrecisionAttentionConfig)
    - api.py: user-facing entry point that validates config and dispatches to the correct backend
    - config.py: AttentionBackend enum and LowPrecisionAttentionConfig dataclass
    - utils.py: hardware capability checks, backend availability detection
    - shared_utils/: shared infrastructure used by backend implementations
        - attention.py: shared _fp8_sdpa implementation (quantize + SDPA)
      - wrapper.py: _FP8FlashAttentionMonkeyPatchWrapper — replaces F.scaled_dot_product_attention during forward, manages flash activation internally
      - setup.py: setup_fp8_backend — builds the wrapper with causal mask detection
    - fp8_fa3/: FA3-specific backend
        - attention.py: fp8_fa3_sdpa elementary block
      - setup.py: thin wrapper calling setup_fp8_backend with FA3 parameters
    - quantization/: shared FP8 quantization kernels
        - quantization.py: _fp8_sdpa_quantize — calls fused Triton kernels for per-head Q, K, V quantization
      - triton_qkv_quantization.py: fused QKV FP8 quantization Triton kernel
  - test/prototype/attention: new folder for low-precision attention API tests
    - test_fp8_attention.py: numerical accuracy tests (eager SDPA) and model-level API tests (simple wrapper)

## Test Plan
`python -m pytest test/prototype/attention/test_fp8_attention.py -v`

## Example Usage
```python
  from torchao.prototype.attention import (
      AttentionBackend,
      LowPrecisionAttentionConfig,
      apply_low_precision_attention,
  )

  model = MyModel()

  # Simple SDPA replacement — no torch.compile needed
  config = LowPrecisionAttentionConfig(backend=AttentionBackend.FP8_FA3)
  model = apply_low_precision_attention(model, config)

  # Flash activation is handled internally by the wrapper
  output = model(inputs)
```
  ---
## Results
#### Single-Layer Results
Results directly comparing FA3 SDPA versus FA3 fp8 SDPA (including quantization time):
<img width="645" height="373" alt="image" src="https://github.com/user-attachments/assets/f64cba6d-4ac9-41b7-b0f7-bf93c67f6c13" />
#### Llama3 Model Results
Results comparing Llama3 model with FA3 SDPA versus Llama3 using the FA3 fp8 wrapper. Does not use RoPE fusion.
Perplexity: 6.19 -> 6.25
<img width="368" height="171" alt="image" src="https://github.com/user-attachments/assets/f7534b8b-6914-4e6b-b568-55e38bdfcb2b" />

### DIFF
--- a/test/prototype/attention/test_fp8_attention.py
+++ b/test/prototype/attention/test_fp8_attention.py
@@ -20,17 +20,17 @@ from torchao.utils import torch_version_at_least
 if torch_version_at_least("2.11.0"):
     from torchao.prototype.attention.utils import _is_fa3_available, _is_hopper
 
-if _is_hopper() and _is_fa3_available():
-    from torch.nn.attention import (
-        activate_flash_attention_impl,
-        restore_flash_attention_impl,
-    )
+    if _is_hopper() and _is_fa3_available():
+        from torch.nn.attention import (
+            activate_flash_attention_impl,
+            restore_flash_attention_impl,
+        )
 
-    from torchao.prototype.attention import (
-        AttentionBackend,
-        apply_low_precision_attention,
-    )
-    from torchao.prototype.attention.fp8_fa3.attention import fp8_fa3_sdpa
+        from torchao.prototype.attention import (
+            AttentionBackend,
+            apply_low_precision_attention,
+        )
+        from torchao.prototype.attention.fp8_fa3.attention import fp8_fa3_sdpa
 
 
 class SimpleAttentionModel(nn.Module):


torchao/prototype/attention/api.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+def apply_low_precision_attention(
+    model: nn.Module,
+    config: Optional[LowPrecisionAttentionConfig] = None,
+) -> nn.Module:
+    """
+    Apply low-precision attention to a model.
+
+    Depending on the configuration, the model is either:
```
vkuzo: this is a bit confusing, can we make the distinction more clear? either with a boolean in this API, or just have two APIs, etc

torchao/prototype/attention/api.py
```diff
@@ -0,0 +1,127 @@
...
+    - **Monkey-patch path** (``fuse_rope=False``, default): wraps the model
+      so that ``F.scaled_dot_product_attention`` is replaced with the FP8
+      backend at call time.  No ``torch.compile`` is needed.
+    - **Compile path** (``fuse_rope=True``): internally calls
```
vkuzo: if `fuse_rope` also changes compile settings, let's call it something like `fuse_rope_using_torch_compile`?

torchao/prototype/attention/config.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+    Configuration for low-precision attention inference.
+
+    Note: Some low-precision attention only supports inference (forward pass).
+    Backward pass may not be supported by the underlying backends.
+
+    Args:
+        backend: Attention backend to use. If None (default), automatically
+            selected based on hardware capabilities.
+        use_hadamard: Apply Hadamard transform. Options:
```
vkuzo: this looks like a boolean but is actually an enum, maybe the name can be clearer?

torchao/prototype/attention/config.py
```diff
@@ -0,0 +1,58 @@
...
+            - None: No Hadamard transform (default)
+            - "v": Apply Hadamard to V only
+            - "qkv": Apply Hadamard to Q, K, and V
+        fuse_rope: If True, the model is compiled with a custom Inductor
```
vkuzo: this seems to be about fusing rope as well as implementation details of compile usage, can the name reflect this?

torchao/prototype/attention/api.py
```diff
@@ -0,0 +1,127 @@
...
+
+def apply_low_precision_attention(
+    model: nn.Module,
+    config: Optional[LowPrecisionAttentionConfig] = None,
+) -> nn.Module:
+    """
+    Apply low-precision attention to a model.
+
+    Depending on the configuration, the model is either:
+    - **Monkey-patch path** (``fuse_rope=False``, default): wraps the model
```
vkuzo: if user starts with mokey patch patch and calls torch.compile on it, do they get the compile path? if yes, why not just delete the setting for `compile path` from this API and have it automatically work if compile is on?

torchao/prototype/attention/api.py
howardzhang-cv: Yeah that makes sense, I changed the name

torchao/prototype/attention/config.py
howardzhang-cv: Changed to hadamard_mode

torchao/prototype/attention/config.py
howardzhang-cv: Changed to fuse_rope_using_torch_compile

torchao/prototype/attention/api.py
howardzhang-cv: No, if the user starts with monkey patch path and calls torch.compile, it will still just be a torch.compiled model but the F.SDPA is replaced with the triton quant kernel + low precision SDPA. There will be no rope fusion like the compile path. I originally had it set so that the compile path is always on, and it would simply fallback to the simpler path if no RoPE is detected, but I figured that wasn't good since the compile path uses the pre-grad IR path, and the default should be a simpler path that doesn't use that.

test/prototype/attention/test_fp8_attention.py
```diff
@@ -0,0 +1,262 @@
...
+    skip_msg: str
+
+
+def _probe_eager_quantized_sdpa(sdpa_fn, flash_impl: str) -> bool:
+    """Try a tiny quantized SDPA call to verify the backend works in eager mode.
+
+    FA3 uses _scaled_dot_product_attention_quantized internally,
+    which requires FA3 activation. This probe catches mismatches.
+    """
+    try:
```
drisspg: can you run a deslop pass on some of this, I kind of doubt you need code like this

test/prototype/attention/test_fp8_attention.py
```diff
@@ -0,0 +1,262 @@
...
+        activate_flash_attention_impl(flash_impl)
+        try:
+            q = torch.randn(1, 1, 4, 64, device="cuda", dtype=torch.bfloat16)
+            with torch.no_grad():
+                sdpa_fn(q, q, q, is_causal=False)
```
drisspg: also we should just know right?

torchao/prototype/attention/shared_utils/attention.py
```diff
@@ -0,0 +1,135 @@
...
+
+_TORCH_VERSION_AT_LEAST_2_11 = torch_version_at_least("2.11.0")
+
+if _TORCH_VERSION_AT_LEAST_2_11:
+    from torch.nn.attention import SDPBackend, sdpa_kernel
+    from torch.nn.attention.experimental._scaled_dot_product_attention_quantized import (
+        _scaled_dot_product_attention_quantized,
+    )
+
+_MIN_VERSION_ERROR = (
```
drisspg: I feel like these global error strings is weird Claude coding we could just inlin e

torchao/prototype/attention/shared_utils/attention.py
```diff
@@ -0,0 +1,135 @@
...
+            f"dropout_p must be 0.0 for FP8 {backend_name}, got {dropout_p}"
+        )
+
+    input_dtype = query.dtype
+
+    # Ensure Triton kernels launch on the correct device.
+    # In the monkey-patch path (no torch.compile), accelerate's hooks move
+    # tensors to the correct device but don't call torch.cuda.set_device().
+    # Triton dispatches based on current_device(), not tensor device, so
+    # without this guard the kernel launches on the wrong GPU's stream.
```
drisspg, line 60: this is kinda weird to me .. hmmm

torchao/prototype/attention/shared_utils/setup.py
```diff
@@ -0,0 +1,197 @@
...
+    if q_len != kv_len:
+        return False
+
+    # Build reference causal mask on the same device / shape and compare.
+    ref = torch.tril(torch.ones(q_len, kv_len, dtype=torch.bool, device=mask.device))
+    # Broadcast: mask may be [B, 1, Q, KV] or [B, H, Q, KV]; ref is [Q, KV].
+    return torch.equal(mask.broadcast_to(mask.shape), ref.expand_as(mask))
+
+
+def detect_causal_mask(
```
drisspg: is this really the best way?

torchao/prototype/attention/shared_utils/wrapper.py
```diff
@@ -0,0 +1,171 @@
...
+
+# ============================================================================
+# Causal mask stripping helper
+# ============================================================================
+
+
+def _make_causal_aware_sdpa(fp8_sdpa_fn: Callable, strip_causal_mask: bool) -> Callable:
+    """Wrap an FP8 SDPA function to strip materialized causal masks.
+
+    HuggingFace models (e.g. LLaMA) pass a materialized lower-triangular
```
drisspg: I'm surprised this is true.. are you sure ?

torchao/prototype/attention/shared_utils/attention.py
howardzhang-cv: Oh yeah, I was actually planning on asking you this later in the 1/1. This was weird to me too, but for some reason without this guard, triton does actually dispatch to the wrong device leading to NaNs in multi-gpu setups.

torchao/prototype/attention/shared_utils/setup.py
howardzhang-cv: Yeah, it's surprisingly difficult to have a general way to see if the attention mask is just a causal mask. Claude suggested a bunch of ideas that imo either leaned too heavily on user interaction or was not robust enough. We can't do it in the pre-grad IR pass because we don't have access to the attention mask values to see if it's causal, so I came up with this method where we do a fake forward pass to see if the masks are causal to get rid of them. I could definitely just be doing something dumb though, there likely is a better way.

torchao/prototype/attention/shared_utils/wrapper.py
howardzhang-cv: Yeah I had to verify to see if it was true as well. In "SDPA" attention mode, huggingface materializes an attention mask instead of just using causal. Since FA3/FA4 aren't compatible with attention masks and only work with causal, I had to do the whole sketchy "detect mask" thing. On Blackwell though, it doesn't do this, only on Hopper.

torchao/prototype/attention/api.py
drisspg: where does this stand post convo?

howardzhang-cv: We'll keep the higher-level API with the --fuse_rope option. If the user enables fuse_rope, we send a warning stating that they must explicitly run torch.compile to enable the rope fusion and low precision attention pathway. No more magic behind-the-scenes compile

torchao/prototype/attention/config.py
```diff
@@ -0,0 +1,58 @@
...
+    Backward pass may not be supported by the underlying backends.
+
+    Args:
+        backend: Attention backend to use. If None (default), automatically
+            selected based on hardware capabilities.
+        hadamard_mode: Apply Hadamard transform. Options:
+            - None: No Hadamard transform (default)
+            - "v": Apply Hadamard to V only
+            - "qkv": Apply Hadamard to Q, K, and V
+        fuse_rope_using_torch_compile: If True, fuse RoPE + quantization + SDPA into optimized
```
vkuzo: if user is using torch.compile, do we still need this? shouldn't we just pick the best available fusion in torch.compile land?

torchao/prototype/attention/api.py
```diff
@@ -0,0 +1,100 @@
...
+)
+from torchao.prototype.attention.utils import (
+    _check_backend_available,
+    _get_available_backend,
+)
+
+
+def apply_low_precision_attention(
+    model: nn.Module,
+    config: Optional[LowPrecisionAttentionConfig] = None,
```
vkuzo: thoughts about removing the config and just adding arguments here? will be more discoverable

howardzhang-cv: Yup, that makes sense to me. I originally had it as a config because I want it to be usable as we start adding more and more parameters. But, I think for what it is now, it makes more sense to have it as arguments. I have made the change.

torchao/prototype/attention/config.py
howardzhang-cv: Unfortunately we'll still be needing this flag, since we need to use the pre-grad IR pass, which is a global setting, and there is no way to selectively use the pre-grad IR pass for just the selected module (as far as I know but I could totally be wrong). It's one of the reasons why I first chose to have the behind the scenes compile, so we can set the global flag, compile that one module, and reset it to default. 

What I have now is we return a compile backend, and in the user warning, we tell them they must run torch.compile(module, backend=returned_backend) to compile that module with the specific backend that they need (which is different depending on FA3/FA4). We still need the flag, and the user cannot just call torch.compile on the whole model with us automatically picking the best available fusion for that specific module, but I think it's probably still better than having a magic compile the user doesn't know about?

PR #3968, opened 2026-02-28
TrainingWeightWrapperTensor base class; subclasses for FP8/MXFP8 with grouped_mm and linear overrides

danielvegamyhre:
## Tensor subclass changes
- `TrainingWeightWrapperBaseTensor`: base Common logic for FSDP, __torch_dispatch__, subclass initialization etc is in this base class (not to be used directly) 
    - Common base class also enables common model conversion / param wrapping code 
- MXFP8 and FP8 tensor subclasses inherit from this and implement the override __torch_function__ with the specific grouped_mm and linear overrides, dispatching to the appropriate autograd functions wrapping our kernels

## Autograd function changes
- Add new `_to_mxfp8_then_scaled_mm` autograd func to support linear op overrides. Supports `wgrad_with_hp` as well.

## Other
- Delete `MXLinear` and `MXLinearConfig` so we don't have two diverging ways of doing mxfp8 dense training. This also removes MXFP4 training support but nobody is using this as far as we know so not creating tech debt is preferable.

## Tests
- `./test/prototype/moe_training/test_everything.sh`
- `pytest test/prototype/mx_formats/test_mx_linear.py`
- `pytest test/prototype/mx_formats/test_mx_tensor.py`


torchao/prototype/moe_training/config.py
```diff
@@ -33,14 +33,19 @@ class MXFP8GroupedMMRecipe(Enum):
     MXFP8_EMULATED_RCEIL = "mxfp8_emulated_rceil"
 
 
-class GroupedMMConfig(AOBaseConfig):
-    """Base configuration for grouped matrix multiplication. Not intended to be used directly."""
+class TrainingBaseConfig(AOBaseConfig):
```
vkuzo: the name is very generic, how about `TrainingOpBaseConfig` to clarify this is for a single op

torchao/prototype/moe_training/config.py
```diff
@@ -33,14 +33,19 @@ class MXFP8GroupedMMRecipe(Enum):
...
+    Purpose is to support generic model conversion function for linear and grouped gemm
+    low precision training.
+    """
 
     pass
 
 
 @dataclass
-class FP8GroupedMMConfig(GroupedMMConfig):
+class FP8GroupedMMConfig(TrainingBaseConfig):
```
vkuzo: `Float8` instead of `Fp8`, to match PyTorch naming for float8?

torchao/prototype/moe_training/config.py
```diff
@@ -67,23 +72,19 @@ def from_recipe(
 # register as pytree constant so we can use dynamo nonstrict trace in torchao.prototype.moe_training.ep
 @register_as_pytree_constant
 @dataclass
-class MXFP8GroupedMMConfig(GroupedMMConfig):
+class MXFP8TrainingConfig(TrainingBaseConfig):
```
vkuzo: `MXFP8OpTrainingConfig`?

torchao/prototype/moe_training/tensor.py
```diff
@@ -232,49 +205,128 @@ def fsdp_post_all_gather(
...
-            wgrad_with_hp=config.wgrad_with_hp,
-            scale_calculation_mode=config.scale_calculation_mode,
-        )
-    else:
-        raise ValueError(f"Unsupported config type: {type(config)}")
+
+    @classmethod
+    def __torch_function__(cls, func, types, args, kwargs={}):
+        # grouped_mm op override
+        if func.__name__ == cls.grouped_mm_func_name:
```
vkuzo: this is confusing, can this just state the op directly since we are already inside the float8 wrapper?

torchao/prototype/moe_training/tensor.py
```diff
@@ -232,49 +205,128 @@ def fsdp_post_all_gather(
...
+    """
+    A subclass of torch.Tensor that overrides the grouped_mm and linear ops
+    to use dynamic quantization of inputs to MXFP8, then runs the MXFP8 grouped_mm/linear,
+    based on the training config.
+    """
+
+    @classmethod
+    def __torch_function__(cls, func, types, args, kwargs={}):
+        # grouped_mm op override
+        if func.__name__ == cls.grouped_mm_func_name:
```
vkuzo: just say the op directly?

torchao/prototype/moe_training/tensor.py
```diff
@@ -232,49 +205,128 @@ def fsdp_post_all_gather(
...
+            if A_is_2d and B_is_2d_or_3d and offs is not None:
+                return _quantize_then_scaled_grouped_mm(
+                    A,
+                    B,
+                    offs=offs,
+                    config=config,
+                )
+
+        # linear op override
+        elif func.__name__ in cls.mm_func_names:
```
vkuzo: just put the ops here? making the code reader jump around to know which ops go here is confusing

vkuzo:
looks good, I care about cleaning up the `func.__name__ == cls.grouped_mm_func_name` and `elif func.__name__ in cls.mm_func_names` the most from my nit comments, thank you!

vkuzo (approved):
lg if CI passes and you are sure this does not regress anything

danielvegamyhre:
addressed comments, will land once CI green

test/prototype/mx_formats/test_mx_linear.py
```diff
@@ -1,480 +0,0 @@
...
-    "scale_calculation_mode",
-    [
-        ScaleCalculationMode.FLOOR,
-        # even + compile does not work yet:
-        # https://gist.github.com/vkuzo/1a04845cd503b1c75291aa1ea3bf79c4
-        # ScaleCalculationMode.EVEN,
-        ScaleCalculationMode.RCEIL,
-    ],
-)
-def test_linear_compile(
```
vkuzo, line 332: @danielvegamyhre was this moved over? i can't find it

danielvegamyhre: Sort of - linear tests were replaced with training [test cases here](https://github.com/pytorch/ao/blob/b3e0db2fae37427b867f8d2b43a0d94d1a474249/test/prototype/moe_training/test_training.py#L44) - `"shared_experts"` FQN is a linear layer, and there is a `compile` boolean parameterization as well, so it tests mxfp8 linear in eager and with compile.

PR #3952, opened 2026-02-25
Expand Triton autotune configs for MoE FP8 kernels to improve AMD GPU performance

brucechanglongxu:
The MoE FP8 Triton kernels in `float8_rowwise.py` and `jagged_float8_scales.py` each have a single hardcoded autotune config. This means the autotuner never actually tunes anything — it just uses the one config regardless of hardware or problem size.

This PR gates an expanded autotune search space behind `torch.version.hip`, so AMD gets 8-16 candidate configs per kernel while NVIDIA keeps the original single config unchanged. The original values are always included in the AMD search space, so the autotuner can only do better (or equal).

This matters for AMD GPUs in particular — AMD wavefronts are 64 threads (vs 32 on NVIDIA), so the best `num_warps` and pipelining depth tend to differ. Benchmarks on MI250X show 4-15% improvement on Llama4 shapes (see comments). Initial H100 testing with an ungated expanded search space showed ~18% regression on the atomic kernel for larger shapes, which motivated the gating approach.

Complements #3945 (relaxed atomics on AMDGPU), same kernel files.

## Test plan
- [x] `ruff check` / `ruff format` clean
- [x] Benchmarked on AMD MI250X (4-15% faster, results in comments)
- [x] Benchmarked on NVIDIA H100 (no regression with gated configs, results in comments)

cc: @BowenBao

PR #3972, opened 2026-03-02
Optimize FP8 colwise scales kernel for AMD GPUs in MoE backward pass

lizamd:
## Summary

Builds on #3952 which expanded Triton autotune configs for AMD GPUs. This PR further optimizes the `_triton_fp8_per_group_colwise_scales_kernel` (used in the backward pass of `_Float8GroupedMM`) with two interdependent changes:

1. **Remove redundant `.t().contiguous().t()` memory copies** before calling `triton_fp8_per_group_colwise_scales` in the backward pass. The kernel already handles arbitrary strides via its stride parameters, so these full-tensor copies (which convert row-major to column-major) are unnecessary.

2. **Use larger Triton autotune configs** (`BLOCK_SIZE=128/256`, `BLOCK_SIZE_ITER=128/256`, `num_warps=8`) for the colwise scales kernel on AMD GPUs, replacing the 16 smaller configs from #3952 (`BLOCK_SIZE=32/64`). With row-major input (from change 1), larger block sizes enable contiguous column access patterns, reducing grid block count by 4-8x.

These two changes are interdependent: removing the memory copy passes row-major tensors to the kernel, and the larger block sizes are optimal for row-major input. With column-major input (the old behavior), the larger block sizes actually perform worse due to strided access patterns.

Note: #3952's expanded configs for `float8_rowwise.py` (the 3D transpose kernel) are unchanged — this PR only modifies the `jagged_float8_scales.py` configs.

## Ablation Study

The `.t().contiguous().t()` was originally added because with small block sizes, column-major input was faster. Our ablation shows neither change works alone — they must be applied together:

End-to-end training with [torchtitan](https://github.com/pytorch/torchtitan) on 8x MI300X with DeepSeek-MoE-16B (EP=8, batch=1, seq_len=4096, torch.compile enabled):

| `.t().contiguous().t()` | BLOCK_SIZE | Input Layout | TPS | Notes |
|---|---|---|---|---|
| Kept | 32/64 (#3952) | col-major | 136 | Baseline |
| **Removed** | 32/64 (#3952) | row-major | 137 | No improvement — small blocks still do strided column access |
| Kept | 128/256 (this PR) | col-major | 127 | **Worse** — large blocks + col-major = bad stride pattern |
| **Removed** | **128/256 (this PR)** | **row-major** | **642** | **4.7x — both changes together** |

**Why the interdependency?** The colwise scales kernel iterates over rows in blocks of `BLOCK_SIZE_ITER` and over columns in blocks of `BLOCK_SIZE`. With small `BLOCK_SIZE` (32/64), each block covers few columns so the strided access from column-major layout is tolerable, and the `.t().contiguous().t()` copy makes those accesses contiguous. With large `BLOCK_SIZE` (128/256), each block covers many more columns — row-major input means these columns are contiguous in memory, enabling efficient wide loads. But column-major input with large blocks creates large strides that kill performance.

## Benchmarks

End-to-end training with [torchtitan](https://github.com/pytorch/torchtitan) on 8x MI300X with DeepSeek-MoE-16B (EP=8, seq_len=4096, torch.compile enabled):

| Batch Size | Baseline TPS (#3952) | Optimized TPS | Speedup |
|-----------|---------------------|--------------|---------|
| 1 | 136 | 642 | 4.7x |
| 4 | 500 | 1,865 | 3.7x |

Profiler analysis shows `_triton_fp8_per_group_colwise_scales_kernel` was the dominant bottleneck (83% of GPU time at baseline). After optimization, kernel time dropped from 11.8s to 4.1s with 10x fewer kernel launches (86,265 → 8,671).

## Test plan
- [x] Verified training convergence (loss decreasing normally) with torchtitan DeepSeek-MoE-16B
- [x] Profiled with PyTorch profiler to confirm kernel speedup
- [x] Ablation study confirming both changes are needed together
- [ ] Run existing unit tests: `pytest test/prototype/moe_training/ -v`

### DIFF
--- a/torchao/prototype/moe_training/fp8_grouped_mm.py
+++ b/torchao/prototype/moe_training/fp8_grouped_mm.py
@@ -214,9 +214,7 @@ class _Float8GroupedMM(torch.autograd.Function):
         # needed for grad_B: grad_output_t @ A
         # Use transpose method to avoid uncoalesced memory accesses.
         grad_out_data_colwise, grad_out_scales = triton_fp8_per_group_colwise_scales(
-            grad_output.t()
-            .contiguous()
-            .t(),  # Quantization is over 2x faster when input is col major, even with this transformation
+            grad_output,
             offs,
             float8_dtype,
             round_scales_to_power_of_2=True,
@@ -225,9 +223,7 @@ class _Float8GroupedMM(torch.autograd.Function):
         grad_output_t_scales = grad_out_scales.t()
 
         A_data_col_major, A_scales = triton_fp8_per_group_colwise_scales(
-            A.t()
-            .contiguous()
-            .t(),  # Quantization is over 2x faster when input is col major, even with this transformation
+            A,
             offs,
             float8_dtype,
             round_scales_to_power_of_2=True,
--- a/torchao/prototype/moe_training/kernels/jagged_float8_scales.py
+++ b/torchao/prototype/moe_training/kernels/jagged_float8_scales.py
@@ -41,14 +41,20 @@ if torch_version_at_least("2.7.0") and has_triton():
     if torch.version.hip is not None:
         kernel_configs_2D = [
             triton.Config(
-                {"BLOCK_SIZE": block_size, "BLOCK_SIZE_ITER": block_size_iter},
-                num_warps=warps,
-                num_stages=stages,
-            )
-            for block_size in [32, 64]
-            for block_size_iter in [64, 128]
-            for warps in [4, 8]
-            for stages in [2, 3]
+                {"BLOCK_SIZE": 128, "BLOCK_SIZE_ITER": 128},
+                num_warps=8,
+                num_stages=2,
+            ),
+            triton.Config(
+                {"BLOCK_SIZE": 128, "BLOCK_SIZE_ITER": 256},
+                num_warps=8,
+                num_stages=2,
+            ),
+            triton.Config(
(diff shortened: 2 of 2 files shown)


danielvegamyhre:
>Remove redundant .t().contiguous().t()

This is interesting. I originally added this as a temporary measure because I found that the quantization kernels were two times as fast with the tensor in col major, even with this memory layout transform. We pivoted to mxfp8 and i never got a chance to investigate further.

Are you able to benchmark this change on a H100 as well?

lizamd:
we can test on MI350 also. H100 is not tested yet, we can add it

danielvegamyhre:
> we can test on MI350 also. H100 is not tested yet, we can add it

ok please let me know

alex-minooka:
> > we can test on MI350 also. H100 is not tested yet, we can add it
> 
> ok please let me know

image

Here is some data on h100, with the configurations listed in the image. PR gives slight performance improvement on 1 node.

danielvegamyhre:
@alex-minooka @lizamd looks great! will land once CI is green. for future reference, please include microbenchmarks in the PR description (e.g., in this could you would run `~/ao/benchmarks/prototype/moe_training/fp8_rowwise/bench_triton_fp8_per_group_colwise_scales.py` - you can find other relevant kernel microbenchmarking scripts for fp8 moe training in there as well)

PR #3973, opened 2026-03-02
Add fused FP8 rowwise scale+cast kernel for MoE forward pass

lizamd:
## Summary

Follow-up to #3972 (which builds on #3952). Adds a fused Triton kernel for the forward pass of `_Float8GroupedMM`.

The fused kernel (`triton_fp8_rowwise_2d_scale_and_cast`) replaces the 3-kernel sequence in the forward pass:
1. `tensor_to_scale(A, axiswise_dim=-1)` — computes per-row absmax and scale
2. `A_scaled = A.to(float32) * A_scales` — applies scale
3. `A_fp8 = to_fp8_saturated(A_scaled, fp8_dtype)` — clamps and casts to FP8

The fused kernel performs per-row absmax and FP8 cast in a single kernel launch with two passes, benefiting from L2 cache reuse on the second pass.

Depends on #3972.

## Benchmarks

End-to-end training with [torchtitan](https://github.com/pytorch/torchtitan) on 8x MI300X with DeepSeek-MoE-16B (EP=8, batch=4, seq_len=4096, torch.compile enabled):

| Config | TPS | Improvement over baseline (#3952) |
|--------|-----|----------------------------------|
| Baseline (#3952 only) | 500 | — |
| With #3972 | 1,865 | 3.7x |
| **With #3972 + this PR** | **2,153** | **4.3x** |

This PR provides an additional ~15% throughput improvement on top of #3972.

## Test plan
- [x] Verified training convergence (loss decreasing normally) with torchtitan DeepSeek-MoE-16B
- [x] Profiled with PyTorch profiler to confirm kernel is used
- [ ] Run existing unit tests: `pytest test/prototype/moe_training/ -v`

### DIFF
--- a/benchmarks/prototype/moe_training/fp8_rowwise/bench_colwise_block_configs.py
+++ b/benchmarks/prototype/moe_training/fp8_rowwise/bench_colwise_block_configs.py
@@ -7,198 +7,198 @@
 # Sweeps (BLOCK_SIZE, BLOCK_SIZE_ITER, num_warps) for the colwise FP8 kernel
 # on representative DeepSeek-MoE-16B training shapes (MI300X).
 #
-# Run via subprocess-per-config to isolate GPU context failures.
 # Usage:
 #   cd ~/ao
 #   python benchmarks/prototype/moe_training/fp8_rowwise/bench_colwise_block_configs.py
 
 import itertools
-import json
-import os
-import subprocess
-import sys
+from dataclasses import dataclass
+from typing import List
 
 import torch
+import triton
+from tabulate import tabulate
+from triton.testing import do_bench
 
-# ---------------------------------------------------------------------------
-# When called as a worker subprocess (BENCH_CFG env var set), run one config.
-# ---------------------------------------------------------------------------
-if "BENCH_CFG" in os.environ:
-    import triton
-    import triton.language as tl
-    from triton.testing import do_bench
+from torchao.prototype.moe_training.kernels.jagged_float8_scales import (
+    _triton_fp8_per_group_colwise_scales_kernel,
+)
+from torchao.prototype.moe_training.utils import generate_jagged_offs
 
-    from torchao.prototype.moe_training.utils import generate_jagged_offs
+device = torch.device("cuda")
 
-    EPS = 1e-12
-    FP8_DTYPE_MAP = {
-        "float8_e4m3fn": (torch.float8_e4m3fn, tl.float8e4nv),
-        "float8_e4m3fnuz": (torch.float8_e4m3fnuz, tl.float8e4b8),
-    }
-
-    # Kernel uses same dimension convention as production:
-    #   input shape = (K, N) where K=token rows (jagged), N=hidden cols
-    #   offsets mark group boundaries along K
-    #   scales buffer = N * N_GROUPS
-    #   BLOCK_SIZE tiles N (columns), BLOCK_SIZE_ITER tiles K (rows, inner loop)
-    @triton.jit
-    def _colwise_kernel(
-        input_ptr, offsets_ptr, out_ptr, scales_ptr,
-        K: tl.int64, N: tl.int64, N_GROUPS: tl.int64,
-        str_ir: tl.int64, str_ic: tl.int64,
-        str_or: tl.int64, str_oc: tl.int64,
-        fp8_min: tl.constexpr, fp8_max: tl.constexpr,
-        in_dtype: tl.constexpr, out_dtype: tl.constexpr,
-        BLOCK_SIZE: tl.constexpr, BLOCK_SIZE_ITER: tl.constexpr, EPS: tl.constexpr,
-    ):
-        # block_col_id tiles the N (hidden/col) dimension
-        bcol = tl.program_id(0)
-        gidx = tl.program_id(1)
-        # group boundaries along K (token) dimension
-        rs = tl.load(offsets_ptr + gidx - 1, mask=gidx > 0, other=0)
-        re = tl.load(offsets_ptr + gidx)
-        col_offs = (bcol * BLOCK_SIZE + tl.arange(0, BLOCK_SIZE)).to(tl.int64)
-        amax = tl.zeros((BLOCK_SIZE,), dtype=in_dtype)
-        for r0 in range(rs, re, BLOCK_SIZE_ITER):
-            row_offs = (r0 + tl.arange(0, BLOCK_SIZE_ITER)).to(tl.int64)
-            offs = row_offs[:, None] * str_ir + col_offs[None, :] * str_ic
-            mask = (row_offs[:, None] < re) & (col_offs[None, :] < N)
-            d = tl.load(input_ptr + offs, mask=mask, other=0.0).to(in_dtype)
-            amax = tl.maximum(amax, tl.max(tl.abs(d), axis=0)).to(in_dtype)
-        amax = amax.to(tl.float64)
-        s = (fp8_max / tl.clamp(amax, min=EPS, max=float("inf"))).to(tl.float32)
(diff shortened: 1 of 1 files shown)


danielvegamyhre:
Looks great @lizamd thanks for adding this - I'll take a closer look tomorrow, btw there's a linter issue you'll need to fix, try making sure `ruff` version is same on as in dev requirements.txt and run:

- `ruff check --fix <dirs>`
- `ruff format <dirs>`

torchao/prototype/moe_training/kernels/float8_rowwise.py
```diff
@@ -494,6 +494,224 @@ def _fake_triton_fp8_rowwise_3d_transpose_rhs_fused_reduction(
...
+            triton.Config(
+                {"BLOCK_SIZE_K": 128},
+                num_warps=4,
+                num_stages=4,
+            )
+        ]
+
+    @triton.autotune(configs=fused_2d_kernel_configs, key=["K"])
+    @triton.jit
+    def _triton_fp8_rowwise_2d_fused_scale_and_cast_kernel(
```
danielvegamyhre, line 511: Looks reasonable. Please add (1) a unit test comparing to pytorch reference impl, in ~/ao/test/prototype/moe_training somewhere, and (2) a microbenchmark comparing to torch.compile performance in ~/ao/benchmarks/prototype/moe_training somewhere. for the benchmark, please copy/paste one of the existing kernel microbenchmark scripts in that dir to get started, we use the same consistent structure / benchmarking scaffold code

we should always add unit tests and microbenchmarks for every kernel we add.

danielvegamyhre (approved):

benchmarks/prototype/moe_training/fp8_rowwise/bench_triton_fp8_rowwise_2d_fused_scale_and_cast.py
```diff
@@ -0,0 +1,268 @@
...
+        A_data = to_fp8_saturated(A_scaled, float8_dtype)
+        return A_data, A_scales
+
+    run_original_compiled = torch.compile(run_original)
+
+    # --- Column 2: 3-kernel sequence inside a compiled graph with an opaque boundary.
+    # Simulates the actual MoE forward pass: torch.compile cannot fuse across the
+    # tensor_to_scale call because it's treated as an opaque custom op, leaving
+    # 3 separate kernel launches — exactly what the fused kernel replaces in practice.
+    def run_compiled_graph_unfused(A: torch.Tensor):
```
danielvegamyhre: why do you want to measure this? seems like "torch.compile of torch native impl with fullgraph=True" versus "triton kernel" is the only comparison we need

lizamd: `torch.compile` of the isolated 3-kernel sequence is the best-case scenario for the unfused path — the compiler sees only those 3 ops and can fuse them freely. But in actual MoE training, `triton_fp8_rowwise_2d_scale_and_cast` is called as a `custom_op` inside a much larger compiled graph (see `fp8_grouped_mm.py`). In that context, `torch.compile` also sees `tensor_to_scale` as opaque (it goes through its own dispatch path), so the 3 kernels remain separate launches. The `compile+opaque` column simulates this by wrapping `tensor_to_scale` with `torch.compiler.disable`, which is closer to what the compiler actually sees during training. The `compile-isolated` column is still included for reference, but the speedup that matters for this PR is `triton vs compile+opaque`.

danielvegamyhre: > But in actual MoE training, triton_fp8_rowwise_2d_scale_and_cast is called as a custom_op inside a much larger compiled graph (see fp8_grouped_mm.py). In that context, torch.compile also sees tensor_to_scale as opaque (it goes through its own dispatch path)

If we are using the triton kernel wrapped in a custom op, `tensor_to_scale` won't be opaque/untraceable, it will just not be used at all. 

And in the torch native implementation when we _do_ use `tensor_to_scale`, it will not be opaque, it will be traceable, and torch.compile definitely can do operator fusion across this boundary.

lizamd:
Thanks for the review — this is a good point and worth clarifying.

You're correct that in the **forward pass**, `tensor_to_scale` is not called at all for `A` — `triton_fp8_rowwise_2d_scale_and_cast` already replaced it (line 107). So the "compile+opaque" column doesn't represent the forward path.

However, it accurately represents the **backward pass**. In `_Float8GroupedMM.backward` ([fp8_grouped_mm.py:167-175](torchao/prototype/moe_training/fp8_grouped_mm.py#L167-L175)), `grad_output` is still quantized using the unfused 3-op pattern:
```python
grad_output_scales = tensor_to_scale(grad_output, ...)                     # kernel 1
grad_output_scaled = grad_output.to(torch.float32) * grad_output_scales   # kernel 2
grad_output_data_row_major = to_fp8_saturated(grad_output_scaled, ...)    # kernel 3
```
Because this runs inside `torch.autograd.Function.backward`, torch.compile cannot trace into it — the entire backward is an opaque boundary by definition, regardless of whether `tensor_to_scale` itself is traceable. These 3 launches cannot be fused by the compiler.

The `torch.compiler.disable(tensor_to_scale)` in the benchmark simulates exactly this: a compiled graph where the compiler cannot see across the `tensor_to_scale` call, leaving 3 separate launches — which is what `triton_fp8_rowwise_2d_scale_and_cast` would replace in the backward.

Applying the fused kernel to `grad_output` in backward is a valid follow-up, though the primary bottleneck there is `triton_fp8_per_group_colwise_scales` (lines 216/225) rather than the rowwise quantization.

danielvegamyhre:
> Because this runs inside `torch.autograd.Function.backward`, torch.compile cannot trace into it — the entire backward is an opaque boundary by definition, regardless of whether `tensor_to_scale` itself is traceable. These 3 launches cannot be fused by the compiler.

Thanks @lizamd, this is actually not true though - torch.compile does trace through the backward pass via AOTAutograd, produces graphs, which inductor codegens into Triton kernels. 

Please let me know if I have misunderstood your comment.

lizamd:
You're right, thanks for the correction. I've simplified the benchmark to just torch.compile vs triton as you suggested — removed the compile+opaque column entirely. Updated in 2d97f7e.

danielvegamyhre:
@lizamd ok great! please rebase on main to resolve merge conflicts, then we can land

lizamd:
Rebased on main, conflicts resolved. Ready for landing.

danielvegamyhre:
@lizamd ruff lint issue to fix before landing

benchmarks/prototype/moe_training/fp8_rowwise/bench_colwise_block_configs.py
```diff
@@ -0,0 +1,205 @@
...
+        run()
+    torch.cuda.synchronize()
+
+    t_us = do_bench(run, return_mode="median") * 1e3
+    print(json.dumps({"t_us": t_us}))
+    sys.exit(0)
+
+
+# ---------------------------------------------------------------------------
+# Main driver: spawn one subprocess per config.
```
danielvegamyhre: can you clarify we are doing this, instead of just running the autotuner and letting it cache the best config?

danielvegamyhre: @lizamd please fix ruff/lint error and remove this multi-process benchmarking approach, stick with the standard benchmarking method done in other files

danielvegamyhre:
@lizamd benchmark script has a merge conflict to resolve

danielvegamyhre:
@lizamd looks like ci is failing on this pr

PR #3985, opened 2026-03-04
[not for land] [mxfp8 training] fix TP bug

danielvegamyhre:


pianpwk:
On my side the [LINEAR rank=*] logs are only trigged from the non-parallelized model; this patch gets SQNR from 23 -> 50:
```diff
+++ b/torchao/prototype/moe_training/tensor.py
@@ -21,6 +21,7 @@ from torchao.prototype.moe_training.config import (
 )
 from torchao.prototype.moe_training.utils import _quantize_then_scaled_grouped_mm
 from torchao.prototype.mx_formats.mx_linear import _to_mxfp8_then_scaled_mm
+from torchao.prototype.mx_formats.mx_tensor import MXTensor
 from torchao.utils import TorchAOBaseTensor
 
 aten = torch.ops.aten
             
             return _to_mxfp8_then_scaled_mm(
@@ -357,6 +359,37 @@ class MXFP8TrainingWeightWrapperTensor(TrainingWeightWrapperBaseTensor):
             with torch._C.DisableTorchFunctionSubclass():
                 return func(*args, **kwargs)
 
+    @classmethod
+    def __torch_dispatch__(cls, func, types, args, kwargs={}):
+        # Intercept aten.mm.default to apply MXFP8 quantization.
+        # This is needed because DTensor decomposes F.linear (a CompositeImplicitAutograd op)
+        # into aten.t + aten.mm, bypassing our __torch_function__ override for "linear".
+        # Without this, the TP/SP model would compute with regular BF16 mm.
+        if func == torch.ops.aten.mm.default:
+            A, B = args[0], args[1]
+
+            if not isinstance(A, cls) and isinstance(B, cls):
+                config = B.config
+
+                if isinstance(config, MXFP8TrainingOpConfig):
+                    input_hp = A
+                    # B._data is the transposed weight (from aten.t in linear decomposition).
+                    # _to_mxfp8_then_scaled_mm expects weight in [out, in] layout and
+                    # internally computes input @ weight.t(), so we pass B._data.t()
+                    # to recover the original weight layout:
+                    #   input @ (B._data.t()).t() = input @ B._data
+                    # which matches the aten.mm(input, weight_t) semantics.
+                    weight_hp = B._data.t().contiguous()
+                    return _to_mxfp8_then_scaled_mm(
+                        input_hp,
+                        weight_hp,
+                        kernel_preference=config.kernel_preference,
+                        scale_calculation_mode=config.scale_calculation_mode,
+                        wgrad_with_hp=config.wgrad_with_hp,
+                    )
+
+        return super().__torch_dispatch__(func, types, args, kwargs)
+
```

danielvegamyhre:
> On my side the [LINEAR rank=*] logs are only trigged from the non-parallelized model; this patch gets SQNR from 23 -> 50:
> 
> ```diff
> +++ b/torchao/prototype/moe_training/tensor.py
> @@ -21,6 +21,7 @@ from torchao.prototype.moe_training.config import (
>  )
>  from torchao.prototype.moe_training.utils import _quantize_then_scaled_grouped_mm
>  from torchao.prototype.mx_formats.mx_linear import _to_mxfp8_then_scaled_mm
> +from torchao.prototype.mx_formats.mx_tensor import MXTensor
>  from torchao.utils import TorchAOBaseTensor
>  
>  aten = torch.ops.aten
>              
>              return _to_mxfp8_then_scaled_mm(
> @@ -357,6 +359,37 @@ class MXFP8TrainingWeightWrapperTensor(TrainingWeightWrapperBaseTensor):
>              with torch._C.DisableTorchFunctionSubclass():
>                  return func(*args, **kwargs)
>  
> +    @classmethod
> +    def __torch_dispatch__(cls, func, types, args, kwargs={}):
> +        # Intercept aten.mm.default to apply MXFP8 quantization.
> +        # This is needed because DTensor decomposes F.linear (a CompositeImplicitAutograd op)
> +        # into aten.t + aten.mm, bypassing our __torch_function__ override for "linear".
> +        # Without this, the TP/SP model would compute with regular BF16 mm.
> +        if func == torch.ops.aten.mm.default:
> +            A, B = args[0], args[1]
> +
> +            if not isinstance(A, cls) and isinstance(B, cls):
> +                config = B.config
> +
> +                if isinstance(config, MXFP8TrainingOpConfig):
> +                    input_hp = A
> +                    # B._data is the transposed weight (from aten.t in linear decomposition).
> +                    # _to_mxfp8_then_scaled_mm expects weight in [out, in] layout and
> +                    # internally computes input @ weight.t(), so we pass B._data.t()
> +                    # to recover the original weight layout:
> +                    #   input @ (B._data.t()).t() = input @ B._data
> +                    # which matches the aten.mm(input, weight_t) semantics.
> +                    weight_hp = B._data.t().contiguous()
> +                    return _to_mxfp8_then_scaled_mm(
> +                        input_hp,
> +                        weight_hp,
> +                        kernel_preference=config.kernel_preference,
> +                        scale_calculation_mode=config.scale_calculation_mode,
> +                        wgrad_with_hp=config.wgrad_with_hp,
> +                    )
> +
> +        return super().__torch_dispatch__(func, types, args, kwargs)
> +
> ```

thanks for looking at this @pianpwk, knowing that linear is decomposed into aten.t + aten.mm is super useful. quick question, i thought __torch_dispatch__ runs below autograd so we can't run autograd functions there otherwise it won't run backwards properly? is that not the case? this is why i have the autograd functions in __torch_function__ instead

danielvegamyhre:
@pianpwk 

>it needs to intercept aten::mm at the torch_dispatch level instead.

that makes sense, but in the past when i tried to use autograd functions in __torch_dispatch__, i found only the forward pass ran, never the backward pass, which is problematic for this use case since we need to control the backward pass as well. my understand was that autograd is not captured at the __torch_dispatch__level, am i mistaken on that?

when Dtensor decomposes linear into aten.t + aten.mm, does it skip __torch_function__ entirely and go straight to dispatch? there is no way to intercept in __torch_function__?

pianpwk:
sorry, I think I was also partly mistaken. I'm not 100% sure what causes the (linear -> t + view + mm) decomposition for the DTensor + MXFP8 composition case, but you should also be able to intercept mm for it at the __torch_function__ level as well.

Let me dig a bit more...

danielvegamyhre:
> sorry, I think I was also partly mistaken. I don't think DTensor CIA handling is why the (linear -> t + view + mm) decomposition happens anymore, and you should also be able to intercept mm at the **torch_function** level as well.
> 
> Let me dig a bit more...

ok thanks again for your help - for what it's worth i've been trying to figure out how to intercept mm at torch dispatch level as well for TP case but i don't see any meaningful func names, just `__get__` at function level followed by `mm.default` at dispatch level:

```
TP case
[DISPATCH rank=0] aten.t.default args=['Wrapper(torch.Size([64, 64]), mean=-0.001793)']
func __get__
[DISPATCH rank=0] aten.mm.default args=['Tensor(torch.Size([256, 64]), mean=0.496094)', 'Wrapper(torch.Size([64, 64]), mean=-0.001793)']
not preserving subclass aten.mm.default
[DISPATCH rank=0] aten.t.default args=['Wrapper(torch.Size([64, 64]), mean=-0.000147)']
func __get__
[DISPATCH rank=0] aten.mm.default args=['Tensor(torch.Size([256, 64]), mean=0.496094)', 'Wrapper(torch.Size([64, 64]), mean=-0.000147)']
not preserving subclass aten.mm.default
[DISPATCH rank=0] aten.t.default args=['Wrapper(torch.Size([64, 64]), mean=0.000112)']
func __get__
[DISPATCH rank=0] aten.mm.default args=['Tensor(torch.Size([256, 64]), mean=-0.000790)', 'Wrapper(torch.Size([64, 64]), mean=0.000112)']
not preserving subclass aten.mm.default
```

pianpwk:
> but i don't see any meaningful func names, just __get__ at function level followed by mm.default at dispatch level:

my bad, this only triggered because I had a custom dispatch mode on (DebugMode), otherwise it's at dispatch level. I get your point about autograd though, let me ask around. I tried reversing (parallelize before quantize) but that hit other composability issues.

danielvegamyhre:
> > but i don't see any meaningful func names, just **get** at function level followed by mm.default at dispatch level:
> 
> my bad, this only triggered because I had a custom dispatch mode on (DebugMode), otherwise it's at dispatch level. I get your point about autograd though, let me ask around. I tried reversing (parallelize before quantize) but that hit other composability issues.

sounds good, thanks again - i will keep debugging on my side as well

pianpwk:
@danielvegamyhre I vibecoded this reordering fix which seems to work? https://github.com/pytorch/ao/pull/4010

danielvegamyhre:
> @danielvegamyhre I vibecoded this reordering fix which seems to work? #4010

@pianpwk per our discussion on that PR and elsewhere, do you know of a way to intercept the linear op when the wrapping is DTensor(MXFP8TrainingTensor(..))?  that is the fundamental issue i am trying to solve, from logging it seems to be decomposed into

- \_\_get\_\_ (not sure what this is)
- t
- mm

torchao/prototype/moe_training/tensor.py
```diff
@@ -263,14 +278,11 @@ class MXFP8TrainingWeightWrapperTensor(TrainingWeightWrapperBaseTensor):
     @classmethod
     def __torch_function__(cls, func, types, args, kwargs={}):
         # grouped_mm op override
+        print("[TORCH_FUNCTION]", func.__name__)
```
andrewor14, line 281: remove these before landing?

torchao/prototype/mx_formats/mx_tensor.py
```diff
@@ -170,7 +170,8 @@ def to_mx(
     assert data_hp.shape[-1] % block_size == 0, (
         f"the last dimension of shape {data_hp.shape} must be divisible by block_size {block_size}"
     )
-    assert data_hp.is_contiguous(), "unsupported"
+    if not data_hp.is_contiguous():
+        assert data_hp.is_contiguous(), "unsupported"
```
andrewor14, line 174: a bit confused by this, if it's not contiguous it would fail like before, so is there a reason behind this change?

torchao/prototype/moe_training/tensor.py
```diff
@@ -40,6 +40,24 @@
     torch.ops.aten.clone.default,
     torch.ops.aten.transpose.int,
     torch.ops.aten.t.default,
+    # required for TP - scatter_ is used to distribute weights
+    torch.ops.c10d.scatter_.default,
```
andrewor14, line 44: so this is the real fix right? Do we need the other ops you commented out?

torchao/prototype/mx_formats/mx_tensor.py
danielvegamyhre: @andrewor14 sorry i linked the wrong PR, this is not the one that will address test failures, this is a WIP draft for an issue we are still trying to find a proper solution for - please disregard this PR.

PR #3992, opened 2026-03-04
[ROCm] Enable FSDP2 Float8 and affine quantized tensor parallel tests on ROCm

brucechanglongxu:
Currently these tests are blanket-skipped on ROCm even though most of them have no FP8 dependency. This patch removes the unconditional ROCm skips and replaces the hardware capability gates with proper checks that cover both CUDA (SM89+/SM90+) and ROCm (MI300/MI350).

Changes in `test/float8/test_fsdp2/test_fsdp2.py`:
- Collapse the two separate module-level skips (one for `is_sm_at_least_89`, one for ROCm) into a single gate that also accepts MI300/MI350.
- Use `e4m3_dtype` from config instead of hardcoded `torch.float8_e4m3fn` in `test_amax_allreduce_device_mesh`, since MI300 uses `float8_e4m3fnuz`.

Changes in `test/dtypes/test_affine_quantized_tensor_parallel.py`:
- Remove the module-level `pytest.skip` on ROCm that was blocking all TP tests including Int8wo, Int4wo, and Int8dq which don't require FP8.
- Fix the Float8 TP class gate to use `is_sm_at_least_90()` instead of raw `get_device_capability() >= (9, 0)`. The old check incorrectly passes on MI250X (gfx90a reports capability 9.0 but lacks FP8).

Tested on MI250X (gfx90a, 8 GPUs): FSDP2 Float8 tests correctly skip, affine quantized TP passes 4/6 (Int8wo 3/3, Int8dq 1/1, Float8 classes correctly not defined).

cc: @danielvegamyhre

### DIFF
--- a/test/float8/test_fsdp2/test_fsdp2.py
+++ b/test/float8/test_fsdp2/test_fsdp2.py
@@ -48,7 +48,10 @@ from torchao.testing.training.fsdp2_utils import (
 from torchao.utils import is_MI300, is_MI350, is_sm_at_least_89
 
 if not (is_sm_at_least_89() or is_MI300() or is_MI350()):
-    pytest.skip("Requires FP8-capable GPU (CUDA SM89+, MI300, or MI350)", allow_module_level=True)
+    pytest.skip(
+        "Requires FP8-capable GPU (CUDA SM89+, MI300, or MI350)",
+        allow_module_level=True,
+    )
 
 
 class TestFloat8Common:


danielvegamyhre:
@brucechanglongxu please fix ruff linter error

fyi @jerryzh168 as this touches AQT test as well

danielvegamyhre:
please fix ruff linter @brucechanglongxu

PR #4010, opened 2026-03-05
[TP] reorder MXFP8 wrapper over DTensor

pianpwk:
For https://github.com/pytorch/pytorch/issues/177059, https://github.com/pytorch/ao/pull/3985

The original attempt to handle TP + MXFP8 wrapped DTensor over the MXFP8 subclass. The MXFP8 subclass intends to capture at the torch_function level, and use custom autograd functions to control fwd/bwd behavior. Because DTensor has no fwd/bwd coupling, and because it CIA-decomposes aten::linear, this ordering does not work; MXFP8 tensor does not see aten::linear at __torch_function__

This PR reverses the order to MXFP8(DTensor), allowing aten::linear interception and fwd/bwd control. Relies on https://github.com/pytorch/pytorch/pull/177234 landing in pytorch.

I understand other dtypes still need reordering?

More details in discussion of https://github.com/pytorch/pytorch/issues/177059

### DIFF
--- a/torchao/prototype/mx_formats/mx_tensor.py
+++ b/torchao/prototype/mx_formats/mx_tensor.py
@@ -354,7 +354,7 @@ def get_fp_scale(scale_e8m0):
     s_offset = scale_e8m0.to(torch.int16) - E8M0_EXPONENT_BIAS
     # TODO(later): it would be nice if there was a way to do the 2^x operation
     # in PyTorch without creating a tensor of twos
-    two = torch.full_like(s_offset, 2.0)
+    two = torch.full_like(s_offset, 2.0, dtype=torch.float32)
     # pow(two, s_offset) can be out of range of floating point formats.
     # TODO(later): handle this for float16 if we decide to support float16
     # scales.


test/prototype/mx_formats/test_mx_dtensor.py
```diff
@@ -43,52 +46,69 @@ def setup_distributed():
...
 
 if __name__ == "__main__":
     device_mesh = setup_distributed()
     tests = [
-        _test_dtensor_cast_to_mxfp4,
-        _test_mxfp8_mlp_tensor_parallelism,
+        _test_dtensor_cast_to_mxfp8,
+        _test_mxfp8_mlp_tensor_parallelism_emulated,
     ]
+    # The auto test requires the mxfp8_quantize CUDA kernel to be built
```
danielvegamyhre: we have a flag for this you can check instead of this try/except test: https://github.com/pytorch/ao/blob/5a5029dd2ec0c3f046a41b4472d7c1c943d7e3f2/torchao/prototype/moe_training/kernels/mxfp8/quant.py#L736

torchao/testing/training/dtensor_utils.py
```diff
@@ -145,38 +164,98 @@ def _test_lowp_mlp_tensor_parallelism_base(
...
+
+    sp_out = sp_model(x_sp_input)
+    sp_out.backward(go_sp)
+
+    global_out = toy_model_fp8(x)
+    global_out.backward(go)
+
+    if is_mxfp8:
+        # MXFP8 block quantization may produce small numerical differences
+        # between TP and non-TP due to block boundary alignment, so use
```
danielvegamyhre: this isn't quite accurate, the mxfp8 quantization will fail if the dim we're scaling along isn't evenly divisible by 32 (1x32 scaling factor size). this test passes previously with `torch.testing.assert_close` when we had a "MXLinear" module swap method (swaps nn.Linear for the mxfp8 linear layer). Using this tensor subclass op override method the numerics should still be equivalent.

danielvegamyhre:
thanks for investigating this further @pianpwk ! i added a couple notes to assist

torchao/prototype/moe_training/tensor.py
```diff
@@ -67,7 +69,7 @@ def __new__(
             dtype=tensor.dtype,
             layout=tensor.layout,
             device=tensor.device,
-            pin_memory=tensor.is_pinned(),
+            pin_memory=tensor.is_pinned() if not isinstance(tensor, DTensor) else False,
```
danielvegamyhre: why is this necessary?

torchao/prototype/mx_formats/utils.py
```diff
@@ -206,10 +206,19 @@ def _to_mxfp8_dim1_kernel_wrapper(
...
         )
+        # Transpose placements to match the transposed local data.
+        # a_data_local.t() flips the sharded dimension for 2D tensors.
+        transposed_placements = []
+        for p in a_data.placements:
+            if isinstance(p, Shard):
+                assert a_data.ndim == 2
+                transposed_placements.append(Shard(1 - p.dim))
+            else:
+                transposed_placements.append(p)
```
danielvegamyhre: hmm, can you help me understand why this is necessary? it looks like this code unconditionally does this placement transpose if the input is a 2d dtensor - shouldn't this only happen if we are doing a transpose op? and shouldn't dtensor update the sharding dims/placements automatically when a transpose happens?

pianpwk: you know what, in this the rewrapping seems to keep the old order `DTensor(MXFP8(...))`, I imagine we'd need to reverse this too (the transpose is happening on the "local" MXFP8 tensor, so when we rewrap with DTensor specs we adjust the sharded dim accordingyl)

pianpwk:
@danielvegamyhre updated the PR to cover more spots where we should reorder. I didn't review carefully but at a high-level it should make more sense?

Also had to override a DTensor sharding rule for scaled_mm - I'll upstream that to pytorch instead, if it turns out to be valid.

danielvegamyhre:
nice @pianpwk, qq, why did you need to register a custom sharding strategy for scaled_mm, does the one in core have a bug?

torchao/prototype/moe_training/tensor.py
pianpwk: I guess it's a workaround, we'll see `*** NotImplementedError: Operator aten.is_pinned.default does not have a sharding strategy registered.` without guarding this. probably better to add a strategy

pianpwk: linking https://github.com/pytorch/pytorch/pull/177235

torchao/prototype/mx_formats/mx_tensor.py
```diff
@@ -578,28 +598,6 @@ def to_mx(
...
-                is_swizzled_scales,
-            )
-            return DTensor.from_local(
-                inner_mx_tensor,
-                data_lp.device_mesh,
-                data_lp.placements,
-                run_check=False,
-                shape=data_lp.size(),
-                stride=data_lp.stride(),
-            )
```
pianpwk, line 664: consequence of reversing order?

torchao/testing/training/dtensor_utils.py
```diff
@@ -145,38 +164,97 @@ def _test_lowp_mlp_tensor_parallelism_base(
...
+        )
+
+        out_proj_grad2_sqnr = compute_error(
+            tp_model.ffn.out_proj.weight.grad, sp_model2.ffn.out_proj.weight.grad
+        )
+        assert out_proj_grad2_sqnr >= MIN_SQNR, (
+            f"out_proj.weight.grad (sp_model2) SQNR {out_proj_grad2_sqnr} < {MIN_SQNR}"
+        )
+    else:
+        torch.testing.assert_close(tp_out, global_out)
```
danielvegamyhre: we need to always use assert_close instead of sqnr for this test

pianpwk: seems good now?

torchao/prototype/moe_training/tensor.py
danielvegamyhre: i see, can we land that PR first then remove this change?

test/prototype/mx_formats/test_mx_dtensor.py
```diff
@@ -43,52 +46,92 @@ def setup_distributed():
...
+    if _mxfp8_cuda_kernels_available:
+        try:
+            from torchao.prototype.mx_formats.kernels import mxfp8_quantize_cuda
+
+            t = torch.randn(32, 32, device="cuda", dtype=torch.bfloat16)
+            mxfp8_quantize_cuda(t, rowwise=False, colwise=True)
+            del t
+            tests.append(_test_mxfp8_mlp_tensor_parallelism_auto)
+        except Exception:
+            print("Skipping auto test: mxfp8_quantize CUDA kernel not available")
```
danielvegamyhre: if `_mxfp8_cuda_kernels_available` let's just append the test case without doing the kernel dispatch test

torchao/testing/training/dtensor_utils.py
```diff
@@ -90,7 +90,17 @@ def _test_lowp_mlp_tensor_parallelism_base(
         rowwise_parallel_cls = Float8RowwiseParallel
         prepare_input_cls = PrepareFloat8ModuleInput
 
+    # For MXFP8: parallelize first, then quantize.
+    # This puts MXFP8 wrapper on top of DTensor so __torch_function__
+    # intercepts F.linear before DTensor can trigger premature all-gathers.
```
danielvegamyhre, line 95: @vkuzo this wrapping order change to `MXFP8WeightWrapperTensor(Dtensor(...))` is pretty fundamental one, if you'd like to review as well. see the comment above, it is working cleanly now after @pianpwk landed a fix to dtensor sharding rules for 1d/flattened scale factors for `torch._scaled_mm`: https://github.com/pytorch/pytorch/pull/177234

torchao/prototype/moe_training/tensor.py
danielvegamyhre: nevermind i see it is landed now, let's remove this if/else here then

torchao/prototype/mx_formats/mx_tensor.py
danielvegamyhre: makes sense, this would rewrap as "Dtensor(MXTensor(...))" which the opposite order of what we are doing now. nice that all this can be removed now, cleaner

torchao/prototype/mx_formats/mx_tensor.py
```diff
@@ -703,6 +653,29 @@ def _get_gemm_choice(
...
+            t = (
+                t.view(torch.uint8)
+                .redistribute(placements=t_placements)
+                .view(torch.float8_e8m0fnu)
+            )
+        out = local_map(
+            to_blocked,
+            in_placements=(t_placements,),
+            out_placements=t_placements,
+        )(t)
```
danielvegamyhre, line 673: to confirm my understanding, local_map just runs the function (to_blocked) on each local shard as if it were a plain tensor, and then rewraps the output in a dtensor according to out_placements right

torchao/prototype/mx_formats/kernels.py
```diff
@@ -1122,35 +1122,41 @@ def custom_mxfp8_quantize_cuda_dim1_sharding(
...
+            [Shard(0)] + non_tensor_args,
+        )
+
+        # When input is sharded along dim 1:
+        # output_colwise: cols are sharded → Shard(1)
+        # scales_colwise: col dim is sharded → Shard(0)
+        rule_shard_dim1 = (
+            [Replicate(), Shard(1), Replicate(), Shard(0)],
+            [Shard(1)] + non_tensor_args,
+        )
```
danielvegamyhre, line 1157: why did these dim1 quantization kernel sharding rules need to be updated?

also, i thought the rule tuple order was (inputs, outputs) but it seems like this is the opposite, do i have it backwards?

torchao/prototype/mx_formats/mx_tensor.py
pianpwk: Yep!

torchao/prototype/mx_formats/kernels.py
pianpwk: Ahh ordering is [*outputs, *inputs], I think that's the issue: https://github.com/pytorch/pytorch/blob/6511db6ea2ae6133a5260076862f63540564897a/torch/distributed/tensor/experimental/_register_sharding.py#L22

danielvegamyhre:
@pianpwk when i use latest torch nightly, checkout this PR and run `./test/prototype/mx_formats/test_mx_dtensor.sh` the test fails with: 

```
[rank0]:   File "/home/dev/ao/torchao/prototype/mx_formats/mx_tensor.py", line 409, in to_dtype
[rank0]:     data_hp = data_hp * s_fp
[rank0]:               ~~~~~~~~^~~~~~
[rank0]: RuntimeError: The size of tensor a (8) must match the size of tensor b (4) at non-singleton dimension 0
```

pianpwk:
> ./test/prototype/mx_formats/test_mx_dtensor.sh



> @pianpwk when i use latest torch nightly, checkout this PR and run `./test/prototype/mx_formats/test_mx_dtensor.sh` the test fails with:
> 
> ```
> [rank0]:   File "/home/dev/ao/torchao/prototype/mx_formats/mx_tensor.py", line 409, in to_dtype
> [rank0]:     data_hp = data_hp * s_fp
> [rank0]:               ~~~~~~~~^~~~~~
> [rank0]: RuntimeError: The size of tensor a (8) must match the size of tensor b (4) at non-singleton dimension 0
> ```

hmm not able to repro this

danielvegamyhre:
> > ./test/prototype/mx_formats/test_mx_dtensor.sh
> 
> > @pianpwk when i use latest torch nightly, checkout this PR and run `./test/prototype/mx_formats/test_mx_dtensor.sh` the test fails with:
> > ```
> > [rank0]:   File "/home/dev/ao/torchao/prototype/mx_formats/mx_tensor.py", line 409, in to_dtype
> > [rank0]:     data_hp = data_hp * s_fp
> > [rank0]:               ~~~~~~~~^~~~~~
> > [rank0]: RuntimeError: The size of tensor a (8) must match the size of tensor b (4) at non-singleton dimension 0
> > ```
> 
> hmm not able to repro this

did fresh uninstall/install and it isn't happening anymore, idk what happened. this lgtm - thanks!

Issue #3928, opened 2026-02-20
[Float8DynamicActivationFloat8WeightConfig] Enable Dynamic Quantized models to be exportable using torch.export.export


asfiyab-nvidia:
I'm using the script below to quantize the activation and weights of a toy linear model and want to export it using `torch.export.export`. but I run into `NotImplementedError: UserDefinedObjectVariable(QuantizeTensorToFloat8Kwargs)`. Based on my investigation, QuantizeTensorToFloat8Kwargs is a dataclass inheriting from QuantizeTensorKwargs(abc.ABC). Dynamo has no special handling for dataclass and it falls through to the generic UserDefinedObjectVariable, whose as_proxy() hits the base `NotImplementedError`. Is there a recommendation on how I can work around this issue?

```
# Based on https://github.com/pytorch/ao/blob/main/tutorials/developer_api_guide/my_dtype_tensor_subclass.py
import torch
import torchao
from torchao.quantization import quantize_, Float8DynamicActivationFloat8WeightConfig, PerRow
from torchao.quantization.quantize_.workflows import Float8Tensor

class Float8TensorNonDecomposed(Float8Tensor):
    """Float8Tensor with non-decomposed dequantize"""
    @classmethod
    def from_float8_tensor(cls, tensor: Float8Tensor) -> "Float8TensorNonDecomposed":
        """Convert a Float8Tensor to Float8TensorNonDecomposed by copying its internals."""
        tensor_data, tensor_attrs = tensor.__tensor_flatten__()
        tensor_data_dict = {name: getattr(tensor, name) for name in tensor_data}
        print(tensor_attrs)
        return cls.__tensor_unflatten__(
            tensor_data_dict, tensor_attrs, tensor.shape, tensor.stride()
        )

    def dequantize(self, output_dtype=None):
        if output_dtype is None:
            output_dtype = torch.float16
        return torch.ops.torchao.dequantize_affine_float8_non_decomposed.default(
            self.qdata, self.scale, output_dtype
        )

class linear_model(torch.nn.Module):
    def __init__(self) -> None:
        super().__init__()
        self.linear = torch.nn.Linear(16, 32)

    def forward(self, x: torch.Tensor) -> torch.Tensor:
        return self.linear(x)

def convert_float8_to_float8_non_decomposed(model: torch.nn.Module) -> torch.nn.Module:
    """Convert all Float8Tensor parameters in the model to Float8TensorNonDecomposed."""
    for name, module in model.named_modules():
[...]

PR #4011, opened 2026-03-05
Make QuantizeTensorToFloat8Kwargs a frozen Dataclass to enable torch export

asfiyab-nvidia:
Regular Dataclasses are not supported by the Dynamo Tracer as they're dispatched to `UserDefinedObjectVariable` which doesn't have an implementation for `as_proxy`. However, Frozen Dataclasses are supported by the Dynamo Tracer as the `as_proxy` method is defined for them ([source](https://github.com/pytorch/pytorch/blob/main/torch/_dynamo/variables/user_defined.py#L2253)).

Marking `QuantizeTensorToFloat8Kwargs` as a frozen dataclass enables torch.export on the `Float8Tensor` subclass.

Fixes https://github.com/pytorch/ao/issues/3928

### DIFF
--- a/torchao/quantization/quantize_/workflows/float8/float8_tensor.py
+++ b/torchao/quantization/quantize_/workflows/float8/float8_tensor.py
@@ -58,7 +58,7 @@ __all__ = [
 aten = torch.ops.aten
 
 
-@dataclass
+@dataclass(frozen=True)
 class QuantizeTensorToFloat8Kwargs(QuantizeTensorKwargs):
     """Tensor kwargs for creating float8 tensor (either activation or weight)
 


asfiyab-nvidia:
@jcaip can you please help with the review? Thanks

asfiyab-nvidia:
@jcaip I added another [comment ](https://github.com/pytorch/ao/issues/3928#issuecomment-4014627925) on the [issue](https://github.com/pytorch/ao/issues/3928) that motivates this change. Based on the comment, would you still recommend the change the PR proposes to maintain compatibility with PyTorch < 2.7 where  `unwrap_tensor_subclass` is neeeded?

PR #4016, opened 2026-03-06
[ROCM] Float8 deepseekv3_671b IntOverflow in triton kernels during training 

alex-minooka:
Issue
When running training on deepseekv3_671b on 1 node reduced layers, triton kernels were producing IntOverflow exception after a few iterations.

` 
    File "/opt/conda/lib/python3.11/site-packages/triton/runtime/jit.py", line 744, in run
      kernel.run(grid_0, grid_1, grid_2, stream, kernel.function, kernel.packed_metadata, launch_metadata,
    File "/opt/conda/lib/python3.11/site-packages/triton/backends/amd/driver.py", line 828, in __call__
      self.launch(self.launch_cooperative_grid, gridX, gridY, gridZ, stream, function, profile_scratch, *args)
  OverflowError: signed integer is greater than maximum
`
---

Changes:
Two fixes applied to both _triton_fp8_per_group_rowwise_scales_kernel and _triton_fp8_per_group_colwise_scales_kernel:

Removed unused num_elements parameter — This was computed as hp_tensor.numel() and passed to the Triton kernels but never referenced in the kernel bodies. For DeepSeek V3's large MoE tensors (256 experts, dim=7168), numel() can exceed 2^31 - 1, and the AMD Triton driver packs int kernel args as signed 32-bit integers, directly causing the OverflowError: signed integer is greater than maximum at kernel launch.

Cast pointer arithmetic to tl.int64 — All stride multiplications and offset computations inside both kernels now use 64-bit integers (block_row_offs.to(tl.int64), stride_input_row.to(tl.int64), etc.). This prevents potential int32 overflow in pointer calculations like block_row_offs * stride_input_row for large activation tensors (e.g., ~1M routed tokens × stride 7168 ≈ 7B, which overflows int32).

### DIFF
--- a/torchao/prototype/moe_training/kernels/jagged_float8_scales.py
+++ b/torchao/prototype/moe_training/kernels/jagged_float8_scales.py
@@ -164,13 +164,13 @@ if torch_version_at_least("2.7.0") and has_triton():
         offsets_ptr,
         out_ptr,
         scales_ptr,
-        M: int,
-        K: int,
-        N_GROUPS: int,
-        stride_input_row: int,
-        stride_input_col: int,
-        stride_output_row: int,
-        stride_output_col: int,
+        M: tl.int64,
+        K: tl.int64,
+        N_GROUPS: tl.int64,
+        stride_input_row: tl.int64,
+        stride_input_col: tl.int64,
+        stride_output_row: tl.int64,
+        stride_output_col: tl.int64,
         fp8_dtype_min: tl.constexpr,
         fp8_dtype_max: tl.constexpr,
         input_dtype: tl.constexpr,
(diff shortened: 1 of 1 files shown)


torchao/prototype/moe_training/kernels/jagged_float8_scales.py
```diff
@@ -194,15 +191,24 @@ def _triton_fp8_per_group_rowwise_scales_kernel(
         group_col_end_idx = tl.load(offsets_ptr + offset_idx)
         block_row_offs = block_row_id * BLOCK_SIZE + tl.arange(0, BLOCK_SIZE)
 
+        # cast to int64 to avoid overflow in pointer arithmetic for large tensors
+        block_row_offs_i64 = block_row_offs.to(tl.int64)
```
danielvegamyhre: You can just update the type annotation in the kernel signature to int64 i think, can you try

alex-minooka: ack

PR #4025, opened 2026-03-08
Upgrade torchao build configs to C++20 (#4025)

r-barnes:
Summary:

Update CMAKE_CXX_STANDARD and compiler flags from c++17 to c++20
in torchao's setup.py, CMakeLists.txt, and Makefile files.

Reviewed By: andrewor14

Differential Revision: D95103084


r-barnes:
If regression tests don't fail after rebase, we should be good.

PR #4030, opened 2026-03-09
fix nvfp4 static roofline and add nvfp4 dynamic roofline

vkuzo:
Summary:

the nvfp4 inference roofline we had mistakenly used static global scaling, even
though it was communicated as dynamic global scaling

1. add a proper static roofline
2. fix the docs to properly call out the static numbers as static
3. add the dynamic nvfp4 roofline recipe and add the numbers to the docs

this is in preparation for a future PR which will hook up MSLK's
to_nvfp4 kernel

Test Plan:

https://gist.github.com/vkuzo/4dcfba91f56cc9663b7197b94e2b4791


vkuzo:
* #4031
* __->__ #4030

PR #4038, opened 2026-03-10
[float8] fix `==` typo in `float8_transpose` axiswise dim remapping

### LINKED ISSUE #4444: [float8] `float8_transpose` uses `==` instead of `=` when remapping `_axiswise_dim`
## Bug

[`torchao/float8/float8_ops.py:174-179`](https://github.com/pytorch/ao/blob/main/torchao/float8/float8_ops.py#L174-L179) uses `==` (comparison, result discarded) where `=` (assignment) was clearly intended:

```python
new_axiswise_dim = old_axiswise_dim
if old_axiswise_dim is not None:
    if old_axiswise_dim == 0:
        new_axiswise_dim == -1   # ← comparison, discarded
    else:
        new_axiswise_dim == 0    # ← comparison, discarded
```

`new_axiswise_dim` is never reassigned, so after `Float8TrainingTensor.transpose(0, 1)` or `.t()` the `_axiswise_dim` attribute keeps the original value (e.g. stays `0` after a transpose that should flip it to `-1`). This silently produces wrong axis metadata that downstream matmul / scale-flattening code relies on. The companion `float8_view` (`float8_ops.py:213-239`) already does the correct remap, so this one function is just the outlier.

## Repro

```python
import torch
from torchao.float8 import e4m3_dtype
from torchao.float8.float8_scaling_utils import hp_tensor_and_scale_to_float8
from torchao.float8.float8_utils import tensor_to_scale

a = torch.rand((16, 16), dtype=torch.bfloat16)
scale = tensor_to_scale(a, e4m3_dtype)
fp8 = hp_tensor_and_scale_to_float8(a, scale, e4m3_dtype, axiswise_dim=0)
assert fp8.t()._axiswise_dim == -1, (
    f"expected -1 after transpose, got {fp8.t()._axiswise_dim}"
)
# fails on current main with: expected -1 after transpose, got 0
```

Verified the bug still exists on `main` at `b4805e2e8`.

## Fix

PR #4038 — replace `==` with `=` (2 chars), plus extends `test_transpose` to assert `_axiswise_dim` on the output (the existing test only checked `_data` / `_scale`).

## Related

Sibling typo bug in the same module: #4445 (`cc1` vs `cc2` in `Float8LinearConfig.__post_init__`). Originally bundled with this one in #4443, then split for one-to-one PR auto-close.

GotFusion:
Fixes #4444.

## Summary

In `torchao/float8/float8_ops.py::float8_transpose`, the lines that remap `_axiswise_dim` after transpose use `==` (comparison) instead of `=` (assignment), so the comparison result is discarded and `new_axiswise_dim` is never updated:

```python
if old_axiswise_dim is not None:
    if old_axiswise_dim == 0:
        new_axiswise_dim == -1   # no-op comparison
    else:
        new_axiswise_dim == 0    # no-op comparison
```

After this patch `Float8TrainingTensor.transpose(0, 1)` and `.t()` correctly flip the axiswise dim (`0` ↔ `-1`), matching what `float8_view` (lines 213-239 in the same file) already assumes for reshape.

Verified the bug still exists on `main` at `b4805e2e8` (HEAD as of this rebase).

## Test Plan

Extends the existing `TestFloat8TrainingTensor.test_transpose` (which previously only asserted `_data` / `_scale` on the transposed tensor) to also assert `_axiswise_dim` on both `transpose(0, 1)` and `.t()` outputs, for all parametrized `axiswise_dim` values (`None`, `0`, `-1`).

```bash
pytest test/float8/test_base.py -k 'TestFloat8TrainingTensor and transpose' -v
```

The new assertions fail on `main` (the bug) and pass after the fix.

## Changes since the previous version of this PR

The previous version of this PR also touched two other files. Both have been **split out**:

1. **`torchao/quantization/linear_activation_quantized_tensor.py`** (`tensor_impl` property passthrough)
   — the file was deleted from upstream `main` in the interim (class `LinearActivationQuantizedTensor` no longer exists anywhere in the repo per code search), so this hunk is obsolete.

2. **`torchao/optim/adam.py`** (DTensor branch around `torch.compile(single_param_adam)`)
   — this is a separate functional concern that deserves its own PR with a clearer FSDP2 reproducer. Will file separately if still needed after verification on current `main`.

This PR is now **only** the one-line typo fix plus the regression test that locks it in (+7/-2 lines).

Addresses @vkuzo's earlier ask for a test plan.

## Related

- Sibling typo bug + PR: #4445 / #4115

### DIFF
--- a/test/float8/test_base.py
+++ b/test/float8/test_base.py
@@ -144,11 +144,16 @@ class TestFloat8TrainingTensor:
 
             fp8_a_transposed = fp8_a.transpose(0, 1)
             fp8_b_t = fp8_b.t()
+            expected_axiswise_dim = (
+                None if axiswise_dim is None else (-1 if axiswise_dim == 0 else 0)
+            )
 
             torch.testing.assert_close(
                 (fp8_a_transposed._data, fp8_a_transposed._scale),
                 (fp8_b_t._data, fp8_b_t._scale),
             )
+            assert fp8_a_transposed._axiswise_dim == expected_axiswise_dim
+            assert fp8_b_t._axiswise_dim == expected_axiswise_dim
 
     @pytest.mark.parametrize("shape", [(8, 16), (4, 8, 16), (2, 4, 8, 16)])
     @pytest.mark.parametrize("axiswise_dim", [0, -1])
--- a/torchao/float8/float8_ops.py
+++ b/torchao/float8/float8_ops.py
@@ -174,9 +174,9 @@ def float8_transpose(aten_op, args, kwargs=None):
     new_axiswise_dim = old_axiswise_dim
     if old_axiswise_dim is not None:
         if old_axiswise_dim == 0:
-            new_axiswise_dim == -1
+            new_axiswise_dim = -1
         else:
-            new_axiswise_dim == 0
+            new_axiswise_dim = 0
 
     return Float8TrainingTensor(
         new_data,


vkuzo:
@GotFusion , thank you. Could you include a test plan in the PR summary? Our CI is green, your fix looks logically correct, I just want to know if there is a gap in our test coverage.

vkuzo:
@claude review this PR

GotFusion:
Added regression coverage for the `_axiswise_dim` remap in `test_transpose` and updated the PR summary with the test plan. The test now explicitly checks the expected output dim for `None`, `0`, and `-1`, which was the previous coverage gap.

GotFusion:
@claude review this PR

GotFusion:
@vkuzo Rebased on top of current `main` (`b4805e2e8`) and addressed your earlier ask:

1. **Test plan added to the PR summary** (also reproduced here):
   ```bash
   pytest test/float8/test_base.py -k 'TestFloat8TrainingTensor and transpose' -v
   ```
   The new assertions on `_axiswise_dim` (added by the bundled regression-test commit) fail on `main` and pass after the fix — this addresses the coverage gap that the bot review flagged.

2. **Scope reduced** to just the `==` → `=` typo fix in `float8_ops.py` + its regression test (+7/-2 lines total). The previous version of the PR included two unrelated hunks that have been split off:
   - `torchao/quantization/linear_activation_quantized_tensor.py` — file deleted from `main` in the interim, so the hunk is obsolete.
   - `torchao/optim/adam.py` DTensor branch — separate functional concern, will file as its own PR if still needed after I reverify the FSDP2 case on current `main`.

PR is now `MERGEABLE` again and the single commit has a self-contained title / test plan. PTAL when you have a moment 🙏

vkuzo:
@claude review this PR

GotFusion:
Friendly ping @vkuzo 🙏

Quick recap of current state:

- ✅ Sibling PR #4115 was merged by you on 5/28
- ✅ @claude reviewed this one and concluded **"LGTM. Ship it."** ([5/27 review](https://github.com/pytorch/ao/pull/4038#issuecomment-3551175728))
- ✅ All 15+ CI checks green (1×H100, 1×L4, CUDA, CPU, MPS, aarch64, Ruff, CLA, Meta Import…)
- ✅ Rebased to current main, `MERGEABLE`
- ✅ Test plan in the PR summary addresses your earlier ask

Heads up: a duplicate one-liner PR #4453 was filed for the same issue #4444 on 5/29 — same `==` → `=` fix, but **no regression test**. Since the test in this PR was added per your earlier request and locks in the corrected behavior, this PR is the more complete one to land.

Would appreciate a final 🟢 when you have a moment. Thanks!

vkuzo:
thanks for the fix and sorry for the delayed response!

PR #4042, opened 2026-03-11
[ROCm] Add NVFP4 emulation support for ROCm (MI300/MI350)

brucechanglongxu:
Adds `KernelPreference.EMULATED` support to NVFP4, mirroring what MXTensor already has. On ROCm (MI300X specifically, which lacks native FP4 GEMM), this lets NVFP4 quantization run by dequantizing both operands back to bf16/fp32 and doing a regular `torch.mm`/`torch.addmm` instead of going through `_scaled_mm`.

The main change is in `_addmm_nvfp4_dispatch` in `nvfp4_tensor.py` -- when the resolved kernel preference is EMULATED, we short-circuit before the `_scaled_mm` path and dequantize + matmul instead. `kernel_preference` gets threaded through `NVFP4Tensor.__new__`, `to_nvfp4`, and all the aten op implementations (transpose, slice, view, select, copy) so it survives tensor manipulation. There's a `_resolve_nvfp4_kernel_preference` helper that handles the case where two NVFP4Tensor inputs might have different preferences (same idea as `_get_gemm_choice` for MX).

On the config side, `NVFP4DynamicActivationNVFP4WeightConfig` gets a `kernel_preference` field. The hard `assert is_sm_at_least_100()` in the dynamic quantization path is replaced with a check that actually distinguishes between platforms -- AUTO still requires SM100+ on CUDA or gfx950 on ROCm, but EMULATED runs anywhere. The preference propagates through `QuantizeTensorToNVFP4Kwargs` into the activation quantization at runtime.

Tests: removed `@skip_if_rocm` and the SM100+ skipif decorators from `test_inference_workflow_nvfp4`, `test_nvfp4_matmul_with_amax`, `test_nvfp4_static_quantization_flow`, `test_nvfp4_static_vs_dynamic_quantization`, and `test_serialization`. Each test now checks `is_ROCM()`/`is_MI350()` at the top and picks EMULATED vs AUTO accordingly. On ROCm we also force `use_triton_kernel=False` since the Triton quantization kernel uses CUDA PTX inline asm.

Tested on MI300X (gfx942) in a ROCm Docker container -- quantize/dequantize round-trips, emulated matmul SQNR, dynamic and static quantization flows via `quantize_()`, static-vs-dynamic equivalence, and serialization all pass.

### DIFF
--- a/torchao/prototype/mx_formats/inference_workflow.py
+++ b/torchao/prototype/mx_formats/inference_workflow.py
@@ -298,6 +298,10 @@ def _nvfp4_inference_linear_transform(
     elif step is None:
         # Dynamic quantization
         kernel_pref = config.kernel_preference
+        assert kernel_pref in (KernelPreference.AUTO, KernelPreference.EMULATED), (
+            f"Unsupported kernel_preference={kernel_pref}, "
+            f"only AUTO and EMULATED are supported"
+        )
         if kernel_pref == KernelPreference.AUTO:
             if is_ROCM():
                 from torchao.utils import is_MI350
--- a/torchao/prototype/mx_formats/nvfp4_tensor.py
+++ b/torchao/prototype/mx_formats/nvfp4_tensor.py
@@ -447,12 +447,9 @@ def nvfp4_select(func, types, args, kwargs):
 
 def _resolve_nvfp4_kernel_preference(a_pref, b_pref):
     """Resolve kernel preference between two NVFP4Tensors."""
-    if a_pref == KernelPreference.AUTO:
-        return b_pref
-    if b_pref == KernelPreference.AUTO:
-        return a_pref
     assert a_pref == b_pref, (
-        "Both NVFP4Tensor inputs must have the same kernel preference"
+        f"Both NVFP4Tensor inputs must have the same kernel preference, "
+        f"got {a_pref} and {b_pref}"
     )
     return a_pref
 


torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -427,17 +440,42 @@ def nvfp4_select(func, types, args, kwargs):
         old.is_swizzled_scales,
         old.use_triton_kernel,
         old.act_quant_kwargs,
+        old.kernel_preference,
     )
     return return_and_correct_aliasing(func, args, kwargs, new)
 
 
+def _resolve_nvfp4_kernel_preference(a_pref, b_pref):
```
vkuzo, line 448: I think this should assert that `apref == b_pref` in all cases, to keep it simple

torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -427,17 +440,42 @@ def nvfp4_select(func, types, args, kwargs):
...
 ) -> torch.Tensor:
     """
     Core implementation shared between nvfp4_mm, nvfp4_addmm, and nvfp4_linear.
     The only difference is whether bias is None or not.
     """
+    gemm_choice = _resolve_nvfp4_kernel_preference(
+        a.kernel_preference, b.kernel_preference
+    )
+
+    if gemm_choice == KernelPreference.EMULATED:
```
vkuzo, line 468: lg!

torchao/prototype/mx_formats/inference_workflow.py
```diff
@@ -305,11 +297,21 @@ def _nvfp4_inference_linear_transform(
 
     elif step is None:
         # Dynamic quantization
-        assert is_sm_at_least_100(), (
-            "NVFP4 DYNAMIC mode is only supported on sm100+ machines"
-        )
-
-        weight = getattr(module, parameter_name)
+        kernel_pref = config.kernel_preference
+        if kernel_pref == KernelPreference.AUTO:
```
vkuzo, line 305: we should also assert that `kernel_pref` is either `AUTO` or `EMULATED`, since the other cases are not supported

torchao/prototype/mx_formats/nvfp4_tensor.py
brucechanglongxu: Done — simplified to just assert a_pref == b_pref unconditionally.

torchao/prototype/mx_formats/inference_workflow.py
brucechanglongxu: Added — asserts kernel_pref is AUTO or EMULATED before proceeding.

brucechanglongxu:
@vkuzo Addressed both review comments:
- Simplified `_resolve_nvfp4_kernel_preference` to assert `a_pref == b_pref` unconditionally
- Added assertion that `kernel_pref` is either AUTO or EMULATED

Ready for re-review when you have a chance.

brucechanglongxu:
@vkuzo Gentle ping for re-review — I addressed both of your review comments (simplified `_resolve_nvfp4_kernel_preference` to assert unconditionally, added assertion that `kernel_pref` is either AUTO or EMULATED). This is mergeable and ready for another look.

brucechanglongxu:
@vkuzo NVFP4 emulation on ROCm — hoping to land by end of week; addressed earlier comments, re-review appreciated.

PR #4045, opened 2026-03-11
[moe training] apply per group padding to fp8 grouped mm

danielvegamyhre:
## Context
A test for fp8 MoE training is failing:
```
FAILED test/prototype/moe_training/test_training.py::test_moe_training[recipe_config0-auto-True-target_fqns0] - torch._dynamo.exc.BackendCompilerFailed: backend='inductor' raised:
RuntimeError: Expected mat_a stride along 0 dim to be multiple of 16 bytes, got 16392.
```
This is coming from the fp8 rowwise scaled grouped mm in core. This because 2d TMA needs to have global stride of the target tensor be divisible by 16 bytes, and in the backward pass for `dW = dO^t @ X` the groups are along the contracting dim, and the stride is determined by the group size.

## Summary
- This PR applies the padding/unpadding kernels we wrote for mxfp8 grouped mm, and uses them for fp8 as well to align groups to 16 bytes.
- Move padding/unpadding helpers to `utils.py` to import and use in both fp8 and mxfp8
- Temporarily skip compile test for fp8 due to https://github.com/pytorch/ao/issues/4048 fyi @brucechanglongxu
- Add tensor subclass unwrapping autograd func so we always pass plain tensors into autograd funcs, to simplify space where weird errors can occur.

## Tests
- `pytest test/prototype/moe_training/test_training.py -v -s `
- `pytest test/prototype/moe_training/test_fp8_grouped_mm.py `

### DIFF
--- a/torchao/prototype/moe_training/utils.py
+++ b/torchao/prototype/moe_training/utils.py
@@ -1,3 +1,5 @@
+"""Shared utilities for MoE training: FP8/MXFP8 quantization, group padding/unpadding, and scaled grouped matmul."""
+
 import random
 import warnings
 from typing import Optional, Tuple


drisspg:
@claude please review this PR

drisspg (approved):

drisspg:
@claude add a very short file descirption doc block  to the top of torchao/prototype/moe_training/utils.py

danielvegamyhre:
fyi @brucechanglongxu @lizamd this will reduce the net speedup from FP8 row-wise grouped MM; however, it is necessary because torchtitan is removing the token group padding (context: https://github.com/pytorch/torchtitan/pull/2255)

these padding/unpadding CUDA kernels offer an immediate fix, but is non-optimal since it requires an expensive extra copy of the (usually large) input activations.

i am working on a triton + symmetric memory implementation of a fused all2all + padding operation that eliminates this extra copy.

PR #4069, opened 2026-03-13
[moe training] Optimize FP8 MoE backward pass: fused colwise kernel + AMD tuning

lizamd:
## Summary

Optimizes the FP8 MoE backward pass for DeepSeek-MoE-16B training on MI300X.

### Changes

**1. Fused FP8 rowwise kernel for forward pass** (`kernels/float8_rowwise.py`)
- Adds `triton_fp8_rowwise_2d_scale_and_cast`: a fused Triton kernel that computes scales and casts to FP8 in a single pass, avoiding a second read of the input tensor.
- Used in the forward and backward paths of `_Float8GroupedMM`.

**2. Dual colwise kernel in backward pass** (`kernels/jagged_float8_scales.py`, `fp8_grouped_mm.py`)
- Adds `triton_fp8_per_group_colwise_scales_dual`: a single kernel that quantizes two tensors (grad_output and A) colwise in one launch, replacing two sequential calls.

**3. AMD runtime fixes and tuning** (`config.py`, `utils.py`, `kernels/jagged_float8_scales.py`)
- Add `pad_token_groups_for_grouped_mm: bool = False` field to `Float8TrainingOpConfig`. Default False avoids a D2H sync on AMD where the CUDA padding kernel is unavailable and the torch fallback does `group_sizes.tolist()` which breaks `torch.compile`.
- **Single fixed Triton config on AMD**: The default config (`BLOCK_SIZE=128, BLOCK_SIZE_ITER=128, num_warps=8`) was poorly tuned for MI300X. A sweep over `(BLOCK_SIZE, BLOCK_SIZE_ITER, num_warps)` on representative DeepSeek-MoE-16B shapes found `BLOCK_SIZE=32, BLOCK_SIZE_ITER=128, num_warps=4` runs 2-4x faster — smaller block size reduces register pressure and improves occupancy on MI300X.

| Shape | Old (BS=128, BSI=128, w=8) | Tuned (BS=32, BSI=128, w=4) |
|-------|------|------|
| K=2048, E=64 | 573 us | 243 us (2.4x) |
| K=5120, E=64 | 1339 us | 492 us (2.7x) |
| K=2048, E=128 | 417 us | 218 us (1.9x) |
| K=5120, E=128 | 1005 us | 515 us (2.0x) |

**4. Benchmarks**
- `bench_triton_fp8_rowwise_2d_fused_scale_and_cast.py`
- `bench_triton_fp8_per_group_colwise_scales_dual.py`
- `bench_colwise_block_configs.py` (block size sweep tool)

## End-to-end results (DeepSeek-MoE-16B, 8xMI300X, batch=4, torch.compile=True)

| | TPS |
|--|--|
| Baseline (origin/main) | ~2,480 |
| This PR | ~10,500 |
| **Speedup** | **4.2x** |

## Key finding: Block config matters on MI300X

Profiling showed `_triton_fp8_per_group_colwise_scales_kernel` dominated backward pass time (83% of GPU time, 86k launches/step). The default block config was not tuned for MI300X — a sweep over `(BLOCK_SIZE, BLOCK_SIZE_ITER, num_warps)` on representative shapes found a 2-4x faster config (see table above). The dual colwise kernel and fused rowwise kernel further reduce memory bandwidth overhead.

## Test plan
- [ ] `python test/prototype/moe_training/test_kernels.py`
- [ ] `python test/prototype/moe_training/test_fp8_grouped_mm.py`
- [ ] End-to-end: DeepSeek-MoE-16B training on 8xMI300X with torchtitan

### DIFF
--- a/test/prototype/moe_training/test_distributed.py
+++ b/test/prototype/moe_training/test_distributed.py
@@ -17,13 +17,19 @@ import os
 import pytest
 import torch
 
-from torchao.utils import is_MI300, is_MI350, is_sm_at_least_89
+from torchao.utils import is_MI300, is_MI350, is_ROCM, is_sm_at_least_89
 
 if not torch.cuda.is_available() or not (
     is_sm_at_least_89() or is_MI300() or is_MI350()
 ):
     pytest.skip(
         "Requires FP8-capable GPU (CUDA SM89+, MI300, or MI350)",
+        allow_module_level=True,
+    )
+
+if is_ROCM():
+    pytest.skip(
+        "Distributed MoE tests require world_size=4; ROCm CI has 1 device",
         allow_module_level=True,
     )
 


danielvegamyhre:
@lizamd wow, e2e training speedup looks very compelling - is the baseline using torch.compile?

torchao/prototype/moe_training/config.py
```diff
@@ -71,6 +75,10 @@ def from_recipe(
             raise ValueError(f"Unsupported FP8 recipe: {recipe}")
 
 
+# Backward-compatibility alias: torchtitan uses the Feb 2026 name FP8GroupedMMConfig.
+FP8GroupedMMConfig = Float8TrainingOpConfig
```
danielvegamyhre: Can we just update torchtitan instead of maintaining BC? both torchao and torchtitan have not done a major release yet, still iterating quickly on minor releases

lizamd:
Thanks @danielvegamyhre! Yes, both baseline and new numbers use `torch.compile` — the `deepseek_v3_16b` config has `CompileConfig(enable=True, components=["model", "loss"])` by default. So the comparison is apples-to-apples.

The 4.2x speedup comes entirely from the kernel-level optimizations: eliminating Triton autotune D2H sync overhead (the dominant factor), a better block config for the colwise kernel (2-3x faster kernel execution), and the fused dual colwise kernel launch.

test/quantization/test_da8w4_cpu.py
```diff
@@ -1,7 +0,0 @@
-# Copyright (c) Meta Platforms, Inc. and affiliates.
-# All rights reserved.
```
danielvegamyhre: unintentional change?

danielvegamyhre:
> Thanks @danielvegamyhre! Yes, both baseline and new numbers use `torch.compile` — the `deepseek_v3_16b` config has `CompileConfig(enable=True, components=["model", "loss"])` by default. So the comparison is apples-to-apples.
> 
> The 4.2x speedup comes entirely from the kernel-level optimizations: eliminating Triton autotune D2H sync overhead (the dominant factor), a better block config for the colwise kernel (2-3x faster kernel execution), and the fused dual colwise kernel launch.

nice, i'm really excited to see this! Quick question:

>Triton autotune D2H sync overhead 

what do you mean by this, can you elaborate? the triton autotuning will be done the first time the kernel is executed, then the best configs are cached and re-used. there should be no overhead from this, and no d2h sync?

the only way re-compilation would become an issue is the autotuner config `key` included something that changes (like the number of rows/tokens, for example, which changes based on routing)

benchmarks/quantization/eval_accuracy_and_perf_of_flux.py
```diff
@@ -71,12 +76,18 @@ def print_pipeline_architecture(pipe):
 
 
 def generate_image(
-    pipe, prompt: str, seed: int, device: str, num_inference_steps: int
+    pipe,
+    prompt: str,
+    seed: int,
+    device: str,
```
danielvegamyhre, line 82: are these changes included unintentionally? maybe you need to rebase?

lizamd:
You are completely right, thank you for the correction. The autotune key `["K", "N_GROUPS"]` uses the hidden dim (`K` = 2048 or 5120) and number of experts (`N_GROUPS` = 64 or 128), both of which are fixed across training steps. So the Triton cache works correctly after warmup — no repeated D2H syncs.

The "~33 unique shapes / ~100 syncs per step" claim in the PR description was wrong. I will correct it.

The actual source of the 4.2x speedup is:
1. **Better block config**: The default `BLOCK_SIZE=128, BLOCK_SIZE_ITER=128, num_warps=8` was a poor fit for MI300X. Benchmarking found `BLOCK_SIZE=32, BLOCK_SIZE_ITER=128, num_warps=4` runs 2-4x faster on representative DeepSeek-MoE-16B shapes — smaller block size reduces register pressure and improves occupancy.
2. **Dual colwise kernel**: Halves kernel launches for backward quantization.
3. **Fused rowwise kernel**: Single-pass scale+cast avoids re-reading the input tensor.

torchao/prototype/moe_training/config.py
lizamd: Agreed — removing the alias. Will send a follow-up PR to torchtitan to use `Float8TrainingOpConfig` directly.

test/quantization/test_da8w4_cpu.py
lizamd: Yes, unintentional — already removed in the latest push.

benchmarks/quantization/eval_accuracy_and_perf_of_flux.py
lizamd: Yes, unintentional — already removed in the latest push.

danielvegamyhre:
@claude please review this PR, in particular the triton kernels themselves and the multi-process logic in the benchmarking which seems unexpected/unnecessary to me

danielvegamyhre:
@lizamd please remove the multi-process benchmarking approach and stick with standard benchmarking method used in other files - thanks!

lizamd:
hii @danielvegamyhre I updated based on your comments, seems all checks passed, ready to land? thank you!

torchao/prototype/moe_training/config.py
```diff
@@ -57,7 +57,11 @@ class Float8TrainingOpConfig(TrainingOpBaseConfig):
     # Output dtype for the FP8 grouped GEMMs.
     out_dtype: Optional[torch.dtype] = torch.bfloat16
 
-    # TODO: support pad_token_groups_for_grouped_mm field like MXFP8TrainingOpConfig
+    # Whether to pad the token group sizes to multiples of 16.
+    # On AMD/ROCm this must be False because the CUDA padding kernel is unavailable
+    # and the torch fallback (torch_pad_token_groups) does group_sizes.tolist() which
+    # causes a D2H sync that breaks torch.compile.
+    pad_token_groups_for_grouped_mm: bool = False
```
danielvegamyhre, line 64: in fp8 grouped mm autograd func can you assert this is false since it isn't supported yet?

danielvegamyhre:
@lizamd can you check the ROCM test failures and skip / fix as needed?

danielvegamyhre:
> @lizamd can you check the ROCM test failures and skip / fix as needed?

@brucechanglongxu added skips for blockwise tests on rocm so maybe you need to rebase. also need to disable the distributed tests on rocm, they need 4 devices

danielvegamyhre:
test failure unrelated, landing

PR #4072, opened 2026-03-13
clean up flux-1.schnell benchmark and add to docs

vkuzo:
Summary:

Combining various improvements to the flux-1.schnell benchmark in a
single PR:

* set `num_inference_steps` to 4 to match the default for this model
* turn on cuda graphs via `reduce-overhead` to improve nvfp4 performance for batch size 1,
  this required patching the transformer blocks forward using code I got
  from @jbschlosser
* add a larger batch size for additional performance metrics at larger
  shapes
* remove torch.compile from vae, as compile times are too long and the
  goal of this benchmark is to compare recipes and track performance
  improvements, not achieve best possible latency. Now perf results for
  a single configuration take ~30s - ~1 min, and overall run on one B200
  is ~20 mins.
* fix quant recipe to exclude one more layer with small shapes
* add the results to documentation

I want to land this so we can see the e2e metrics lift from eventually
landing https://github.com/pytorch/ao/pull/4031

Test Plan:

```
./benchmarks/quantization/eval_accuracy_and_perf_of_flux.sh
...
| experiment | lpips_avg | time_s_bsz_1 | speedup_bsz_1 | time_s_bsz_4 | speedup_bsz_4 |
| ---------- | --------- | ------------- | -------------- | ------------- | -------------- |
| bfloat16 | 0 | 0.4178 | 1.00 | 1.4914 | 1.00 |
| float8_rowwise | 0.1236| 0.3455 | 1.21 | 1.1986 | 1.24 |
| mxfp8 | 0.1260 | 0.3673 | 1.14 | 1.2820 | 1.16 |
| nvfp4 | 0.2694 | 0.3308 | 1.26 | 1.1334 | 1.32 |
```

PR #4073, opened 2026-03-13
speed up mxfp8 rceil quantization kernel

vkuzo:
Summary:

Ensure we don't clamp in inductor, as triton does it for us.
This optimization is already there for FLOOR, also adding it
to the RCEIL path.

copy-pasting as we need to unify the scaling anyways so it will be
faster do that as a separate PR.

For large shapes, this brings us from 3.7 TB/s to 6.1 TB/s on B200.

```bash
> python benchmarks/mx_formats/cast_bench.py --mode dim0_mxfp8_rceil
...
// before
time_us 220.12799978256226                                                                                                                                                                                                       
mem_bw_gbps 3696.4628616248297
// after
time_us 134.17600095272064                                                                                                                                                                                                       
mem_bw_gbps 6064.385361184824 
```

Test Plan:

### DIFF
--- a/torchao/prototype/mx_formats/mx_tensor.py
+++ b/torchao/prototype/mx_formats/mx_tensor.py
@@ -147,7 +147,17 @@ def _to_mx_rceil(
     )
 
     # scale and saturated cast the data elements to max of target dtype
-    data_lp = torch.clamp(data_hp * descale_fp, min=-1 * max_pos, max=max_pos)
+    data_lp = data_hp * descale_fp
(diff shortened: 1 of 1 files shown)


vkuzo:
* __->__ #4073
* #4072

vkuzo:
need to debug why the improvement is measurable in the cast benchmark but not in relu-linear microbenchmarks or the e2e flux-1.schnell benchmarks, likely an additional fix is needed

Issue #4050, opened 2026-03-11
Triton + Symmetric Memory fused all2all dispatch + token group padding kernel


danielvegamyhre:
We landed some decent CUDA Kernels to do per group padding the "naive" way (extra kernel that incurs an extra copy of the large input activations).

Fusing the padding into the all2all needed to avoid the extra copy incurred by the standalone padding kernel described above, which hurt our speedup. The benefit of this approach is that, with the all-to-all dispatch, the receiver ranks are already going to be allocating a buffer for the incoming tokens. If we write those tokens to locations aligned with multiples of 32, we avoid the need for this expensive extra copy.
While we're doing this, we can also write incoming tokens grouped by local expert, instead of grouped by remote/source rank, in order to avoid the token shuffle kernel step (i.e., 2d on-device all to all with no d2h sync)

PR #4086, opened 2026-03-14
Fuse expert grouping and padding into the MXFP8 EP dispatch path

MagellaX:
## Summary
This PR wires the fused MXFP8 dispatch-and-group path into the expert-parallel flow.

Instead of doing:
- MXFP8 all-to-all dispatch
- separate expert regroup / permutation
- separate padding for grouped GEMM

the EP path can now dispatch directly into grouped, padded MXFP8 layout that is ready for grouped GEMM.

## What changed
- Added `a2a_dispatch_and_group_mxfp8_fwd_hp_bwd(...)` in `torchao/prototype/moe_training/ep/a2a_dispatch.py`
- Integrated the fused path into the MXFP8 expert-parallel example
- Finished the grouped MXFP8 all-to-all kernel plumbing in `torchao/prototype/moe_training/kernels/mxfp8/comms.py`
- Added EP-level coverage for:
  - fused dispatch/group metadata
  - forward integration through grouped GEMM / unpermute
  - compiled EP pipeline

## Why
`#4050` is about removing the extra post-dispatch regroup/padding step from the MXFP8 EP flow.

This change moves that layout work into the dispatch path itself, so the output of dispatch is already:
- grouped by local expert
- padded to grouped-GEMM alignment
- still in MXFP8

## Validation
Local:
- `ruff check --fix`
- `ruff format`
- `python -m compileall ...`

Hardware:
- Modal `4x B200`

## Notes
- This PR focuses on correctness and EP-path integration.
- It does not include a separate end-to-end performance benchmark in this diff.

addresses #4050


danielvegamyhre:
Hi @MagellaX looks like this is using my WIP PR #4066 is that right?

torchao/prototype/moe_training/ep/a2a_dispatch.py
```diff
@@ -175,3 +183,99 @@ def a2a_dispatch_mxfp8_fwd_hp_bwd(
...
+    )
+    assert torch.equal(input_splits_tensor, expected_input_splits), (
+        f"input_splits {input_splits} do not match num_tokens_per_expert-derived splits {input_splits_tensor.tolist()}"
+    )
+
+    max_input_rows_per_rank = torch.tensor(
+        [input.shape[0]], device=input.device, dtype=torch.int64
+    )
+    dist.all_reduce(max_input_rows_per_rank, op=dist.ReduceOp.MAX, group=group)
+    max_output_rows_per_rank = int(max_input_rows_per_rank.item()) * ep_degree + (
```
danielvegamyhre: d2h sync here (can't do this, kills perf)

torchao/prototype/moe_training/ep/a2a_dispatch.py
```diff
@@ -175,3 +183,99 @@ def a2a_dispatch_mxfp8_fwd_hp_bwd(
...
+    )
+
+    expected_output_splits = torch.tensor(
+        output_splits, dtype=torch.int64, device=input.device
+    )
+    assert torch.equal(output_splits_tensor, expected_output_splits), (
+        f"kernel output_splits {output_splits_tensor.tolist()} do not match expected {output_splits}"
+    )
+
+    padded_rows = int(padded_group_end_offsets[-1].item())
```
danielvegamyhre: d2h sync here from .item()

torchao/prototype/moe_training/ep/a2a_dispatch.py
```diff
@@ -175,3 +183,99 @@ def a2a_dispatch_mxfp8_fwd_hp_bwd(
...
+            num_local_experts,
+            ep_degree,
+            padded_rows,
+            block_size,
+            use_cpu=True,
+        )
+    )
+    permuted_indices = permuted_indices.to(device=input.device, dtype=torch.int32)
+    assert torch.equal(group_offsets.to(torch.int64), padded_group_end_offsets), (
+        f"group_offsets {group_offsets.tolist()} do not match padded_group_end_offsets {padded_group_end_offsets.tolist()}"
```
danielvegamyhre: f-strings are not evaluated lazily, the .tolist() will cause a d2h sync

torchao/prototype/moe_training/ep/a2a_dispatch.py
```diff
@@ -175,3 +183,99 @@ def a2a_dispatch_mxfp8_fwd_hp_bwd(
...
+    expected_output_splits = torch.tensor(
+        output_splits, dtype=torch.int64, device=input.device
+    )
+    assert torch.equal(output_splits_tensor, expected_output_splits), (
+        f"kernel output_splits {output_splits_tensor.tolist()} do not match expected {output_splits}"
+    )
+
+    padded_rows = int(padded_group_end_offsets[-1].item())
+    permuted_indices, num_tokens_per_expert_padded, group_offsets = (
+        generate_permute_indices(
```
danielvegamyhre: this function is used for permuting the tokens from rank-major to expert-major and adding per group padding, it should not be needed in this impl since the a2a should write directly to expert-major

torchao/prototype/moe_training/ep/a2a_dispatch.py
```diff
@@ -175,3 +183,99 @@ def a2a_dispatch_mxfp8_fwd_hp_bwd(
...
+        input_splits, dtype=torch.int64, device=input.device
+    )
+    assert torch.equal(input_splits_tensor, expected_input_splits), (
+        f"input_splits {input_splits} do not match num_tokens_per_expert-derived splits {input_splits_tensor.tolist()}"
+    )
+
+    max_input_rows_per_rank = torch.tensor(
+        [input.shape[0]], device=input.device, dtype=torch.int64
+    )
+    dist.all_reduce(max_input_rows_per_rank, op=dist.ReduceOp.MAX, group=group)
```
danielvegamyhre: i'm confused why there is a new all-reduce being added here, we shouldn't need to do this, we just overallocate the sym mem buffer for the worst case of all tokens being routed to the same rank

MagellaX:
> Hi @MagellaX looks like this is using my WIP PR #4066 is that right?


ahh i didn't even knew that this PR was there :) at the first place, but now looking at the code, i can see the overlap more clearly.

i mean it's definitely in the low-level fused expert-major MXFP8 a2a work in your `#4066`. that part is very much the same direction.

what I added on top here is mostly the EP-path wiring around it
so yeah, the kernel-level idea overlaps with `#4066`, while this PR is more about carrying that path through the actual EP API + tests.

it wasn’t intentionally based on your PR, but after comparing the code, I agree they’re close enough... :)

torchao/prototype/moe_training/ep/a2a_dispatch.py
MagellaX: yeah, makes sense, this should not be doing a host sync here. I was using it just to size the buffer conservatively in the EP wrapper, but that’s not the right shape for the final path.

torchao/prototype/moe_training/ep/a2a_dispatch.py
MagellaX: yep nice catch....

torchao/prototype/moe_training/ep/a2a_dispatch.py
MagellaX: agreed, after looking more closely at `#4066`, this is the clearest signal that I was still partially thinking in terms of the old rank-major -> expert-major regroup path. if the fused a2a writes directly to expert-major padded layout, this extra step should not exist.

torchao/prototype/moe_training/ep/a2a_dispatch.py
MagellaX: yeah, fair point. I added that to derive a max row count for the wrapper allocation, but that’s not the right approach if we’re just overallocating symm-mem for the true worst case....so yeah, this should not be added.

MagellaX:
hey yo!! @danielvegamyhre let me know

MagellaX:
> > hey yo!! @danielvegamyhre let me know
> 
> lmk any updates!! @danielvegamyhre, btw, loved your blog on mxfp8 gemm!

PR #4090, opened 2026-03-16
Add CuTeDSL kernel for 3D tensor quantization to MXFP8

alexsamardzic:


benchmarks/mx_formats/mxfp8_backend_bench.py
```diff
@@ -0,0 +1,133 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
```
danielvegamyhre: nice @alexsamardzic! please let me know when this is ready for review, quick early comment though: if you want to keep this new bench script for dev purposes that is fine, would you mind also updating the existing bench_quantize_3d.py script to benchmark this kernel as well?

alexsamardzic:
The kernel works on everything Hopper and above.  It supports both floor and rceil rounding.

To test:
```sh
pytest test/prototype/moe_training/test_kernels.py -k cuda_mx_dim1_3d_numerics
```

To benchmark:
```sh
for shape in \
  "8 5120 5120" \
  "8 7168 5120" \
  "8 7168 7168" \
  "8 8192 8192"
do
  read -r E N K <<< "$shape"
  python benchmarks/mx_formats/mxfp8_backend_bench.py \
    --dtype bf16 \
    --scaling-mode floor \
    --backends cuda,cutedsl \
    --check-results \
    --stage-count 2 \
    --E "$E" --N "$N" --K "$K"
done
```

Benchmarking results:
```GPU: NVIDIA B200
CC: (10, 0)
shape=(E,N,K)=(8,5120,5120) dtype=torch.bfloat16 scaling_mode=floor stage_count=2
[cuda      ]    0.131 ms    4.847 TB/s   y_stride=(26214400, 1, 5120) s_shape=(8, 160, 5120)
[cutedsl   ]    0.123 ms    5.165 TB/s   y_stride=(26214400, 1, 5120) s_shape=(8, 160, 5120)
diff(cuda vs cutedsl): y_max_abs=0.0 s_max_abs=0.0
check(cuda vs cutedsl): PASS (atol=0.0)
GPU: NVIDIA B200
CC: (10, 0)
shape=(E,N,K)=(8,7168,5120) dtype=torch.bfloat16 scaling_mode=floor stage_count=2
[cuda      ]    0.180 ms    4.934 TB/s   y_stride=(36700160, 1, 7168) s_shape=(8, 224, 5120)
[cutedsl   ]    0.170 ms    5.232 TB/s   y_stride=(36700160, 1, 7168) s_shape=(8, 224, 5120)
diff(cuda vs cutedsl): y_max_abs=0.0 s_max_abs=0.0
check(cuda vs cutedsl): PASS (atol=0.0)
GPU: NVIDIA B200
CC: (10, 0)
shape=(E,N,K)=(8,7168,7168) dtype=torch.bfloat16 scaling_mode=floor stage_count=2
[cuda      ]    0.250 ms    4.983 TB/s   y_stride=(51380224, 1, 7168) s_shape=(8, 224, 7168)
[cutedsl   ]    0.236 ms    5.283 TB/s   y_stride=(51380224, 1, 7168) s_shape=(8, 224, 7168)
diff(cuda vs cutedsl): y_max_abs=0.0 s_max_abs=0.0
check(cuda vs cutedsl): PASS (atol=0.0)
GPU: NVIDIA B200
CC: (10, 0)
shape=(E,N,K)=(8,8192,8192) dtype=torch.bfloat16 scaling_mode=floor stage_count=2
[cuda      ]    0.326 ms    4.998 TB/s   y_stride=(67108864, 1, 8192) s_shape=(8, 256, 8192)
[cutedsl   ]    0.307 ms    5.304 TB/s   y_stride=(67108864, 1, 8192) s_shape=(8, 256, 8192)
diff(cuda vs cutedsl): y_max_abs=0.0 s_max_abs=0.0
check(cuda vs cutedsl): PASS (atol=0.0)
```

Note: nvidia-cutlass-dsl and apache-tvm-ffi packages are now dependencies.

benchmarks/mx_formats/mxfp8_backend_bench.py
alexsamardzic: It is ready for review. Please note that PR keeps the CUDA kernel for now too, will be removed when we're eventually happy with this one.

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
```diff
@@ -0,0 +1,857 @@
...
+        import cutlass  # noqa: F401
+        import cutlass.cute as cute  # noqa: F401
+        import tvm_ffi  # noqa: F401
+    except Exception:
+        return False
+    return True
+
+
+def _cutedsl_compile_options() -> str:
+    return "--enable-tvm-ffi"
```
danielvegamyhre: this enables us to use torch tensors directly instead of converting to cute tensor via `from_dlpack`, is that right?

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
```diff
@@ -0,0 +1,857 @@
...
+    import cutlass
+    import cutlass.cute as cute
+    import cutlass.utils as utils
+    from cutlass.base_dsl._mlir_helpers import arith as _dsl_arith
+    from cutlass.cute.nvgpu import cpasync, tcgen05
+    from cutlass.cute.runtime import make_fake_stream, make_fake_tensor
+
+    # PTX lowering note (checked by dumping PTX):
+    # - RCEIL path currently does not lower to C++'s inline-PTX
+    #   `cvt.rp.satfinite.ue8m0x2.f32` scale conversion.
```
danielvegamyhre: could we just use inline ptx for this like we do in this triton kernel for 2d inputs: https://github.com/pytorch/ao/blob/960f30729038eb58818a468d38e05aa031cfaca3/torchao/prototype/mx_formats/kernels.py#L497

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
```diff
@@ -0,0 +1,857 @@
...
+    else:
+        raise ValueError(
+            f"Unsupported input dtype for CuTeDSL quantize_3d: {input_dtype_name}"
+        )
+
+    # Warp-specialized TMA kernel:
+    # - warp 0: producer (issues TMA G2S and S2G)
+    # - warps [1..compute_warps]: consumers (quantize)
+    # Note: we intentionally keep store on warp 0 (no dedicated store
+    # warp).  A split load-warp/store-warp design was tested and
```
danielvegamyhre, line 115: >A split load-warp/store-warp design was tested and mostly regressed throughput
 
interesting! thanks for noting this, that would have been one of my first questions..

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
```diff
@@ -0,0 +1,857 @@
...
+                if n < N:
+                    vals_block[i] = cutlass.Float32(sIN_tile[0, n_base + i, k_rel])
+                else:
+                    vals_block[i] = cutlass.Float32(0.0)
+            return vals_block
+
+        @cute.jit
+        def _compute_amax(self, vals_block: cute.Tensor):
+            amax = cutlass.Float32(0.0)
+            for i in range(SCALE_DIM_N_VALUE):
```
danielvegamyhre: interesting, there is no builtin amax helper for this?

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
```diff
@@ -0,0 +1,857 @@
...
+                av = vals_block[i]
+                if av < 0:
+                    av = -av
+                if av > amax:
+                    amax = av
+            return amax
+
+        @cute.jit
+        def _compute_scale_rceil(
```
danielvegamyhre, line 225: is this based on the torch impl here? https://github.com/pytorch/ao/blob/960f30729038eb58818a468d38e05aa031cfaca3/torchao/prototype/mx_formats/mx_tensor.py#L154

we should try using this method with inline PTX, i think it would be more instruction efficient: https://github.com/pytorch/ao/blob/960f30729038eb58818a468d38e05aa031cfaca3/torchao/prototype/mx_formats/kernels.py#L474

benchmarks/mx_formats/mxfp8_backend_bench.py
alexsamardzic: Updated `bench_quantize_3d.py` to run CuTeDSL kernel too.

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
```diff
@@ -0,0 +1,857 @@
...
+                        if q < -F8_MAX:
+                            q = -F8_MAX
+                    q_vals4[j] = q
+
+                q_fp8_vec4 = q_vals4.load().to(cutlass.Float8E4M3FN)
+                q_fp8_vals4 = cute.make_rmem_tensor((chunk_vec,), cutlass.Float8E4M3FN)
+                q_fp8_vals4.store(q_fp8_vec4)
+
+                for j in range(chunk_vec):
+                    sOUT_tile[0, sout_base + j, k_rel] = q_fp8_vals4[j]
```
danielvegamyhre: (new to cutedsl, apologies if this comment is off base, let me know): 

this will do 4 vectorized 4 byte rmem->smem writes, right? from what i can see this would appear to be simple row major layout, is that right?

we need the smem layout to be in ((32,4),4) blocked layout for tcgen05.mma usage, and write directly to that layout in smem before we do the TMA copy smem->gmem. if we store in simple row major format we pay the tax of an extra kernel dispatch doing this layout transformation on the scales, which is what we are currently doing and would like to avoid.

probably need to do scattered / uncoalesced 4 byte vectorized stores to smem for this, just make sure to avoid single byte shared stores / `STS.U8` 

if you aren't familiar with this layout i added some diagrams to our recent MXFP8 training blog that help visualize it: https://pytorch.org/blog/mxfp8-training-for-moes-1-3x-training-speedup-vs-bf16-for-llama4-scout-on-gb200-cluster-using-torchao-and-torchtitan/

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
```diff
@@ -0,0 +1,857 @@
...
+            tINs, tINg = cpasync.tma_partition(
+                tma_atom_in,
+                0,
+                cta_layout,
+                sIN_for_tma_partition,
+                gIN_for_tma_partition,
+            )
+            tINg_stage0 = tINg[(None, 0)]
+            tINs_stage0 = tINs[(None, 0)]
+            if warp_idx == 0:
```
danielvegamyhre, line 432: only warp 0 does anything here, why not have this guard at the top of the jit function? does it not affect how the cutedsl is lowered to ptx then sass in this case?

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
```diff
@@ -0,0 +1,857 @@
...
+                                        n_base,
+                                        k_rel,
+                                    )
+
+                                    amax = self._compute_amax(vals_block)
+
+                                    scale_biased, inv_scale = (
+                                        self._compute_scale_from_amax(amax, USE_RCEIL)
+                                    )
+                                    scales_colwise_u8[e, n_block, k] = cutlass.Uint8(
```
danielvegamyhre: correct me if i'm wrong (new to cutedsl) but this seems like this is storing scales in simple row major format, can we store directly in ((32,4),4) layout for tcgen05.mma usage, so we can avoid dispatching the separate Triton kernel to do this blocked layout transformation to prepare for the mxfp8 grouped mm: https://github.com/pytorch/ao/blob/960f30729038eb58818a468d38e05aa031cfaca3/torchao/prototype/moe_training/mxfp8_grouped_mm.py#L258

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
```diff
@@ -0,0 +1,857 @@
...
+    import cuda.bindings.driver as cuda
+
+    stream = cuda.CUstream(int(torch.cuda.current_stream().cuda_stream))
+    n_blocks = N // block_size
+    k_cta_tiles = ceil_div(K, tile_k * k_tiles_per_cta)
+    n_cta_tiles = ceil_div(N, tile_n)
+
+    # FIXME: remove chunking once CuTeDSL/IR indexing is proven safe
+    # above 2^31.
+    # Work around internal index overflow on very large tensors by
```
danielvegamyhre: is it possible to just use int64 instead of int32 to avoid this overflow? or is this literally internal cutedsl core  code that we cannot modify? it would surprise me if it cannot handle large tensors where stride is greater than int32 max value

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
danielvegamyhre: quack code pointer that may be useful here: 
- [tile_atom_to_shape_SF](https://github.com/Dao-AILab/quack/blob/542cf51952cc023f3543fd37c6537b9bde7f517a/quack/gemm_sm100.py#L507)

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
alexsamardzic: Yes, that's correct.

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
alexsamardzic: Indeed there is.  Updated, 5% performance improvement for rceil case.

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
alexsamardzic: Updated.  No performance gains here, but indeed it is a proper change.

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
alexsamardzic: I based floor/rceil implementation on C++ code.  Anyway, I've updated it to follow Triton implementation, and use inline PTX for the conversion.  About 1% performance improvement here.

EDIT: Realized that the performance improvement is only because of inline PTX, so reverted to implementation based on C++ code, with inline PTX.

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
alexsamardzic: Updated, see [below](https://github.com/pytorch/ao/pull/4090#discussion_r2948440317).

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
alexsamardzic: Hmm, initially it was not working and my conclusion was that CuTeDSL compiler forces uses of int32 internally.  In the meantime, I've upgrade CuTeDSL package, and now it seems to work.  So I removed the chunking loop, only one kernel call there now.

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
alexsamardzic: Added an optional output mode to produce blocked layout, while keeping the existing layout as the default so current callers and tests remain unchanged. The new mode writes scales in the same layout currently produced by `triton_mx_block_rearrange_per_group_3d`, so grouped MM can switch to it later and skip the extra rearrange kernel; at that point, the current default will be switched, and writing into the plain layout removed.

<details>
<summary>Script to validate the new output layout</summary>

```Python
import argparse

import torch

from torchao.prototype.moe_training.kernels.mxfp8.quant import (
    mxfp8_quantize_cuda_3d,
    torch_to_blocked_per_group_3d,
)


def main():
    parser = argparse.ArgumentParser()
    parser.add_argument("--E", type=int, default=8)
    parser.add_argument("--N", type=int, default=7168)
    parser.add_argument("--K", type=int, default=7168)
    parser.add_argument("--dtype", choices=["bf16", "fp32"], default="bf16")
    parser.add_argument("--scaling-mode", choices=["floor", "rceil"], default="floor")
    parser.add_argument("--stage-count", type=int, default=2)
    parser.add_argument("--seed", type=int, default=0)
    args = parser.parse_args()

    dtype = {
        "bf16": torch.bfloat16,
        "fp32": torch.float32,
    }[args.dtype]

    torch.manual_seed(args.seed)
    x = torch.randn((args.E, args.N, args.K), device="cuda", dtype=dtype)

    y_logical, s_logical = mxfp8_quantize_cuda_3d(
        x,
        scaling_mode=args.scaling_mode,
        backend="cutedsl",
        stage_count=args.stage_count,
        blocked_scale_output=False,
    )
    y_blocked, s_blocked = mxfp8_quantize_cuda_3d(
        x,
        scaling_mode=args.scaling_mode,
        backend="cutedsl",
        stage_count=args.stage_count,
        blocked_scale_output=True,
    )

    s_blocked_ref = torch_to_blocked_per_group_3d(s_logical.transpose(-2, -1))

    print(f"GPU: {torch.cuda.get_device_name()}")
    print(f"CC: {torch.cuda.get_device_capability()}")
    print(
        f"shape=(E,N,K)=({args.E},{args.N},{args.K}) "
        f"dtype={dtype} scaling_mode={args.scaling_mode} stage_count={args.stage_count}"
    )
    print("y equal:", torch.equal(y_logical, y_blocked))
    print("logical scales shape:", tuple(s_logical.shape))
    print("blocked scales shape:", tuple(s_blocked.shape))
    print("blocked ref shape:", tuple(s_blocked_ref.shape))
    print("blocked equal:", torch.equal(s_blocked, s_blocked_ref))

    y_diff = (y_logical != y_blocked).sum().item()
    s_diff = (s_blocked != s_blocked_ref).sum().item()
    print("y mismatches:", y_diff)
    print("blocked scale mismatches:", s_diff)

    if s_diff:
        idx = (s_blocked != s_blocked_ref).nonzero()[0].tolist()
        print("first blocked mismatch idx:", idx)
        print("blocked kernel:", s_blocked[tuple(idx)].item())
        print("blocked ref   :", s_blocked_ref[tuple(idx)].item())


if __name__ == "__main__":
    main()
```
</details>

<details>
<summary>Command to run the script above</summary>

```sh
python /tmp/compare_cutedsl_blocked_scales.py --E 8 --N 7168 --K 7168 --dtype bf16 --scaling-mode rceil --stage-count 2
```
</details>

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
alexsamardzic: Updated to write one uint32 instead of four fp8, another 4% performance improvement here.

For the output layout update, see [below](https://github.com/pytorch/ao/pull/4090#discussion_r2948649044).

alexsamardzic:
@danielvegamyhre: Thanks for all the comments, in particular for pointing to possible performance improvements.  I've made updates according to all of your suggestions, and we're now getting pretty close to the 6.4 TB/s - here are latest rceil benchmarking results on B200 (note that these are for the default output layout, new variant with blocked output layout is significantly slower):

```
shape=(E,N,K)=(8,5120,5120) dtype=torch.bfloat16 scaling_mode=rceil stage_count=2
[cuda      ]    0.117 ms    5.431 TB/s   y_stride=(26214400, 1, 5120) s_shape=(8, 160, 5120)
[cutedsl   ]    0.105 ms    6.066 TB/s   y_stride=(26214400, 1, 5120) s_shape=(8, 160, 5120)
diff(cuda vs cutedsl): y_max_abs=0.0 s_max_abs=0.0
check(cuda vs cutedsl): PASS (atol=0.0)

shape=(E,N,K)=(8,7168,5120) dtype=torch.bfloat16 scaling_mode=rceil stage_count=2
[cuda      ]    0.162 ms    5.494 TB/s   y_stride=(36700160, 1, 7168) s_shape=(8, 224, 5120)
[cutedsl   ]    0.144 ms    6.183 TB/s   y_stride=(36700160, 1, 7168) s_shape=(8, 224, 5120)
diff(cuda vs cutedsl): y_max_abs=0.0 s_max_abs=0.0
check(cuda vs cutedsl): PASS (atol=0.0)

shape=(E,N,K)=(8,7168,7168) dtype=torch.bfloat16 scaling_mode=rceil stage_count=2
[cuda      ]    0.224 ms    5.565 TB/s   y_stride=(51380224, 1, 7168) s_shape=(8, 224, 7168)
[cutedsl   ]    0.201 ms    6.197 TB/s   y_stride=(51380224, 1, 7168) s_shape=(8, 224, 7168)
diff(cuda vs cutedsl): y_max_abs=0.0 s_max_abs=0.0
check(cuda vs cutedsl): PASS (atol=0.0)

shape=(E,N,K)=(8,8192,8192) dtype=torch.bfloat16 scaling_mode=rceil stage_count=2
[cuda      ]    0.291 ms    5.592 TB/s   y_stride=(67108864, 1, 8192) s_shape=(8, 256, 8192)
[cutedsl   ]    0.262 ms    6.202 TB/s   y_stride=(67108864, 1, 8192) s_shape=(8, 256, 8192)
diff(cuda vs cutedsl): y_max_abs=0.0 s_max_abs=0.0
check(cuda vs cutedsl): PASS (atol=0.0)
```

I'll check the blog you pointed to for further improvement ideas.

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
```diff
@@ -0,0 +1,940 @@
...
+                row_mod_32 = row_in_block % cutlass.Int64(32)
+                dest_in_block = (
+                    row_mod_32 * cutlass.Int64(16)
+                    + row_div_32 * cutlass.Int64(4)
+                    + col_in_block
+                )
+                block_offset = block_row * (
+                    cutlass.Int64(128) * padded_scale_cols
+                ) + block_col * cutlass.Int64(512)
+                scales_out_u8[e, block_offset + dest_in_block] = scale_u8
```
danielvegamyhre: you mentioned perf is much worse if we write scales to blocked layout - i wonder if doing the manual calculations above, as we did in the triton kernel, if we can use cute layouts to achieve this more efficiently in smem first, before writing to gmem? this example from @drisspg may be relevant here: https://github.com/drisspg/transformer_nuggets/blob/0cf46e64d7d160ae72f1724a48ad4a38c6c3e526/transformer_nuggets/cute/to_blocked.py#L11

danielvegamyhre: or maybe we can use this helper? https://github.com/NVIDIA/cutlass/blob/1b741cabaab20892a1e3c6e4fe6cda3ff9f8e568/python/CuTeDSL/cutlass/utils/blockscaled_layout.py#L64

danielvegamyhre: or this one for doing layout transformation in smem before writing to gmem? perf is probably worse for blocked layout output write now because it is doing uncoalesced scattered writes to gmem to achieve this blocked layout.

https://github.com/NVIDIA/cutlass/blob/1b741cabaab20892a1e3c6e4fe6cda3ff9f8e568/python/CuTeDSL/cutlass/utils/blockscaled_layout.py#L88

if we do the layout transform in smem, there will be bank conflicts but we can then do coalesced vectorized writes to gmem (or TMA stores) which should be a net benefit. 

let me think if there is a swizzle that will avoid/minimize bank conflicts will still enabling efficient gmem store patterns..

benchmarks/prototype/moe_training/mxfp8/bench_quantize_3d.py
```diff
@@ -117,12 +119,33 @@ def using_to_mx(x: torch.Tensor) -> torch.Tensor:
...
     time_cuda_3d_us = benchmark_cuda_function_in_microseconds(
         mxfp8_quantize_cuda_3d,
         input_tensor,
         block_size=block_size,
         scaling_mode=str(config.scaling_mode.value),
+        backend="cuda",
+    )
+
+    # bench 3d CuTeDSL kernel
+    _, _ = mxfp8_quantize_cuda_3d(
```
danielvegamyhre: looks like this is accidentally running cuda kernel for cutedsl benchmark

drisspg:
@claude also do a PR review pass here

benchmarks/prototype/moe_training/mxfp8/bench_quantize_3d.py
danielvegamyhre: nevermind sorry didn't see that you added a 'backend' flag to the existing kernel wrapper...

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
drisspg: in that linked code I assume we have the scale data and then do the layout transformation in smem and store to gmem.  I searched over alot of swizzles and that was working best. And indead some  smem bank conflicts experimental were worth the coalesed writes

drisspg:
Can you add the benchmark numbers to the summary

torchao/prototype/moe_training/kernels/mxfp8/quant.py
```diff
@@ -846,20 +857,46 @@ def triton_mx_block_rearrange_2d_K_groups(
         )
 
 
+_lib_mxfp8 = torch.library.Library("torchao", "FRAGMENT")
+try:
```
drisspg: why this code? does not have cutedsl throw a runtime error?

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
```diff
@@ -0,0 +1,940 @@
...
+import os
+from typing import Tuple
+
+import torch
+
+from torchao.utils import ceil_div
+
+
+def _cutedsl_runtime_available() -> bool:
+    try:
```
drisspg: lets use importlib here I think we have soemthign in inductor cutedsl that I have found robust for determining if libs are installed

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
drisspg: nit: just inline

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
```diff
@@ -0,0 +1,940 @@
...
+    return True
+
+
+def _cutedsl_compile_options() -> str:
+    return "--enable-tvm-ffi"
+
+
+def _make_tile_smem_layouts(cute, tile_n: int, tile_k: int):
+    smem_layout_in = cute.make_layout(
+        (1, tile_n, tile_k),
```
drisspg, line 40: Do you need the (tile_N*_tile_K)? for 0 dim and that is why `make_ordered_layout` wouldnt work?

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
```diff
@@ -0,0 +1,940 @@
...
+    if variant_name:
+        if variant_name not in _CUTEDSL_VARIANTS:
+            valid = ", ".join(sorted(_CUTEDSL_VARIANTS.keys()))
+            raise ValueError(
+                f"Unknown TORCHAO_MXFP8_CUTEDSL_VARIANT={variant_name!r}. "
+                f"Valid values: {valid}"
+            )
+        return variant_name, _CUTEDSL_VARIANTS[variant_name]
+
+    # Default table tuned for current measured best behavior.
```
drisspg: hmm this feels like overkill? is there noticiebal wins to autotuning for certain shapes?

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
```diff
@@ -0,0 +1,940 @@
...
+    COMPUTE_WARPS = compute_warps
+    TILE_N = tile_n
+    TILE_K = tile_k
+    K_TILES_PER_CTA = k_tiles_per_cta
+    IS_FULL_K_TILES_VALUE = is_full_k_tiles
+    IS_BLACKWELL_VALUE = is_blackwell
+    BLOCKED_SCALE_OUTPUT_VALUE = blocked_scale_output
+
+    THREADS_PER_BLOCK = (1 + COMPUTE_WARPS) * 32
+    if COMPUTE_WARPS < 1:
```
drisspg: nit, seeing as users are not setting these we shoould jsut assert all these

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
```diff
@@ -0,0 +1,940 @@
...
+    INV_F8_MAX = cutlass.Float32(1.0 / 448.0)
+
+    @dsl_user_op
+    def _cvt_rp_satfinite_ue8m0x2_f32(
+        a: cutlass.Float32,
+        *,
+        loc=None,
+        ip=None,
+    ) -> cutlass.Uint16:
+        return cutlass.Uint16(
```
drisspg, line 157: im surprised new veresions of dsl dont support this. If so I think we should see if there is any SSA overrride that has this

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
alexsamardzic: Done.

torchao/prototype/moe_training/kernels/mxfp8/quant.py
alexsamardzic: Yes, removed the try block.

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
alexsamardzic: Updated.

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
alexsamardzic: Yes, it's needed, as `make_ordered_layout` puts 1 as the first stride.

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
alexsamardzic: This is for the development, the env. variable and unused configs will be removed.  Added a FIXME to clarify it.

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
alexsamardzic: Updated.

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
alexsamardzic: Indeed, but haven't found anything in the CUTLASS source tree or CuTeDSL package.

alexsamardzic:
Current performance results:

| shape | cuda us | plain us | blocked us | cuda TB/s | plain TB/s | blocked TB/s |
|---|---:|---:|---:|---:|---:|---:|
| (8, 5120, 5120) | 117.0 | 105.0 | 107.0 | 5.436 | 6.068 | 5.951 |
| (8, 7168, 5120) | 162.0 | 144.0 | 147.0 | 5.496 | 6.187 | 6.038 |
| (8, 7168, 7168) | 224.0 | 201.0 | 205.0 | 5.573 | 6.197 | 6.077 |
| (8, 8192, 8192) | 291.0 | 262.0 | 267.0 | 5.594 | 6.212 | 6.102 |

<details>
<summary>Script to generate the results above</summary>

```Python
import re
import subprocess

SHAPES = [
    (8, 5120, 5120),
    (8, 7168, 5120),
    (8, 7168, 7168),
    (8, 8192, 8192),
]

COMMON_ARGS = [
    "python",
    "benchmarks/mx_formats/mxfp8_backend_bench.py",
    "--dtype",
    "bf16",
    "--scaling-mode",
    "rceil",
    "--backends",
    "cuda,cutedsl",
    "--check-results",
    "--stage-count",
    "2",
]

BLOCKED_ARGS = [
    "python",
    "benchmarks/mx_formats/mxfp8_backend_bench.py",
    "--dtype",
    "bf16",
    "--scaling-mode",
    "rceil",
    "--backends",
    "cuda,cutedsl",
    "--check-results",
    "--stage-count",
    "2",
    "--blocked-scale-output",
    "--blocked-k-blocks-per-tb",
    "1",
    "--blocked-swizzle",
    "5,2,5",
]

LINE_RE = re.compile(r"^\[(cuda|cutedsl)\s+\]\s+([0-9.]+) ms\s+([0-9.]+) TB/s")


def run_bench(args, e, n, k):
    cmd = args + ["--E", str(e), "--N", str(n), "--K", str(k)]
    out = subprocess.check_output(cmd, text=True, cwd="/data/openteams/scratch/ao")
    rows = {}
    for line in out.splitlines():
        m = LINE_RE.match(line.strip())
        if m:
            backend, ms, tbps = m.groups()
            rows[backend] = (float(ms), float(tbps))
    return rows


print("| shape | cuda us | plain us | blocked us | cuda TB/s | plain TB/s | blocked TB/s |")
print("|---|---:|---:|---:|---:|---:|---:|")

for e, n, k in SHAPES:
    plain = run_bench(COMMON_ARGS, e, n, k)
    blocked = run_bench(BLOCKED_ARGS, e, n, k)

    cuda_ms, cuda_tbps = plain["cuda"]
    plain_ms, plain_tbps = plain["cutedsl"]
    blocked_ms, blocked_tbps = blocked["cutedsl"]

    print(
        f"| ({e}, {n}, {k}) | "
        f"{cuda_ms * 1000:.1f} | "
        f"{plain_ms * 1000:.1f} | "
        f"{blocked_ms * 1000:.1f} | "
        f"{cuda_tbps:.3f} | "
        f"{plain_tbps:.3f} | "
        f"{blocked_tbps:.3f} |"
    )
```
</details>

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
alexsamardzic: The perf for generating blocked layout was not really "much" worse initially, and with the changes in the meantime, writing into blocked layout is just up to 2% slower than writing to "plain" layout, see current performance results [below](https://github.com/pytorch/ao/pull/4090#issuecomment-4091570159).  Tried several variants, the "layout conversion" kernel is really not fast, and the best variant in the context of this quantization kernel turned out to be to do the shuffle when writing to GMEM; other variants are not practical because of block shapes that work best for the quantization part.

danielvegamyhre: >other variants are not practical because of block shapes that work best for the quantization part.

Hmm, i notice you're limited to single byte stores right now due to 128x32 or 256x32 input block reducing to 128x1 or 256x1 scale factors. 

Have you tried expanding threadblock size to 128x128, so the resulting scale factor tile is 128x4 and you can write it directly as a 512 byte contiguous chunk of global memory in ((32,4),4) layout?

alexsamardzic: Yes, tried that - the perfomance is ~20% slower.

danielvegamyhre: interesting! hmm i am curious why that is, wonder what NCU shows? it's fine though, i am happy with the current performance for writing to blocked layout, would rather we spend the time working on other things now. @drisspg i am happy with landing this, i will wait for your comments though if you have any feedback on cute tips / tricks/ best practices etc.

alexsamardzic: > hmm i am curious why that is, wonder what NCU shows?

If I remember it correctly, as several variations got tried, it was just that larger tiles make quantization part less efficient (lower occupancy, more register pressure, etc.).

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
```diff
@@ -0,0 +1,937 @@
...
+                    sIN_staged.iterator + 1 * stage_elems, smem_layout_in
+                )
+                sOUT_tile1 = cute.make_tensor(
+                    sOUT_staged.iterator + 1 * stage_elems, smem_layout_out
+                )
+
+            if tidx == 0:
+                cpasync.prefetch_descriptor(tma_atom_in)
+                cpasync.prefetch_descriptor(tma_atom_out)
+                cute.arch.mbarrier_init(tma_mbar_ptr0, 1)
```
drisspg, line 561: could we use a pipeline object instead of raw mbar?

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
```diff
@@ -0,0 +1,937 @@
...
+            sOUT_tile: cute.Tensor,
+            n0: cutlass.Int64,
+            n_base: cutlass.Int32,
+            k_rel: cutlass.Int32,
+            N: cutlass.Int64,
+            USE_RCEIL: cutlass.Constexpr[bool],
+        ):
+            chunk_vec = 4
+            num_chunks = SCALE_DIM_N_VALUE // chunk_vec
+            for c in range(num_chunks):
```
drisspg, line 413: I feel like tensor SSA could clean this up a little

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
```diff
@@ -0,0 +1,937 @@
...
+            k: cutlass.Int64,
+            n_blocks: cutlass.Int64,
+            scale_biased: cutlass.Int32,
+            BLOCKED_SCALE_OUTPUT: cutlass.Constexpr[bool],
+        ):
+            scale_u8 = cutlass.Uint8(scale_biased)
+            if cutlass.const_expr(BLOCKED_SCALE_OUTPUT):
+                padded_scale_cols = (
+                    (n_blocks + cutlass.Int64(3)) // cutlass.Int64(4)
+                ) * cutlass.Int64(4)
```
drisspg: I saw your comment. I'm not sure if that I totally understand what it means, but basically this is this modulus floor diff math. This is the exact type of thing that I would expect you to be able to construct a layout, maybe even on CPU, and have it handle the indexing. I think Daniel linked the one kernel where I set up the output scale layouts, assuming row major construction.

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
```diff
@@ -0,0 +1,937 @@
...
+            bits = _dsl_arith.bitcast(descale.ir_value(), _dsl_arith.T.i32())
+            exponent = (bits >> cutlass.Int32(23)) & cutlass.Int32(0xFF)
+            mantissa = bits & cutlass.Int32(0x7FFFFF)
+            if exponent == cutlass.Int32(0xFF):
+                if mantissa != cutlass.Int32(0):
+                    scale_biased = cutlass.Int32(0xFF)
+                else:
+                    scale_biased = cutlass.Int32(0xFE)
+            else:
+                if mantissa > cutlass.Int32(0):
```
drisspg: do you think some Where expressions could clean this up? This is some crazy code, lol.

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
alexsamardzic: Updated.  It is tiny bit faster this way too.

My comment was about the fact that your kernel does the "shuffle" in SMEM.  I tried to do that here, but the performance was better with doing it on store.  TBH, I'm not yet fully confident with CuTeDSL layouts and it took some experimentation to get this right, but indeed it should be more readable this way too.

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
alexsamardzic: Updated.

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
alexsamardzic: It reports a compile error, as both operands in comparisons are scalar.  I've updated some other segments of code to use where expressions.

(The floor/rceil code is pretty much literal translation of my understanding of C++ kernel.)

alexsamardzic:
Here are some questions and comments regarding the remaining changes before the PR eventually merged:

1. ~Do we make `nvidia-cutlass-dsl` and `apache-tvm-ffi` dependencies (i.e. do we add them to `dev-requirements.txt`) or it's user responsibility to install these if CuTeDSL functionality wanted?~
2. ~Do we want to remove C++ kernel altogether, or we want to keep an ability to switch between "cuda"/"cutedsl" backends for now?~
3. ~Do we want to keep the ability to save to "plain" layout, or we want to keep only the code saving to "blocked" layout?~
4. ~The test need updates.  Testing "blocked" layout is needed.  Also, rceil should be added into the test, as at the moment both "cuda" and "cutedsl" kernels fail for some tests if rceil added.~
5. ~Shall I remove my `benchmarks/prototype/moe_training/mxfp8/bench_quantize_3d.py` benchmark from the PR?  It was useful for both profiling and validation throughout the development, but `benchmarks/mx_formats/mxfp8_backend_bench.py` has CuTeDSL benchmarking added too now.~

Edit: crossed out completed changes.

danielvegamyhre:
> Here are some questions and comments regarding the remaining changes before the PR eventually merged:
> 
> 1. Do we make `nvidia-cutlass-dsl` and `apache-tvm-ffi` dependencies (i.e. do we add them to `dev-requirements.txt`) or it's user responsibility to install these if CuTeDSL functionality wanted?

i think runtime dependency is fine but it would be best to include as a dependency we ship. I will check if anyone on the team has objections to including it in our builds. (@drisspg any thoughts?)

> 2. Do we want to remove C++ kernel altogether, or we want to keep an ability to switch between "cuda"/"cutedsl" backends for now?

We should delete the cuda c++ kernel once the cutedsl is one is done

> 3. Do we want to keep the ability to save to "plain" layout, or we want to keep only the code saving to "blocked" layout?

We only need the blocked layout, we can delete the support for writing "plain" scales (row major)

> 4. The test need updates.  Testing "blocked" layout is needed.  Also, rceil should be added into the test, as at the moment both "cuda" and "cutedsl" kernels fail for some tests if rceil added.

This is surprising, can you share a specific test case that fails so i can also take a look? Did you make sure to update the reference `to_mx()` call to use RCEIL as well?

> 5. Shall I remove my `benchmarks/prototype/moe_training/mxfp8/bench_quantize_3d.py` benchmark from the PR?  It was useful for both profiling and validation throughout the development, but `benchmarks/mx_formats/mxfp8_backend_bench.py` has CuTeDSL benchmarking added too now.

Yes let's just keep one benchmark script and delete the other. I prefer keeping "benchmarks/prototype/moe_training/mxfp8/bench_quantize_3d.py" since the structure is consistent with the rest of kernel benchmark scripts.

alexsamardzic:
> > 4. The test need updates.  Testing "blocked" layout is needed.  Also, rceil should be added into the test, as at the moment both "cuda" and "cutedsl" kernels fail for some tests if rceil added.
> 
> This is surprising, can you share a specific test case that fails so i can also take a look? Did you make sure to update the reference `to_mx()` call to use RCEIL as well?
> 

Indeed, `scaling_mode=scaling_mode` should be added to `to_mx` call.  I'll do that, and add `ScaleCalculationMode.RCEIL` to the parametrization.

Edit: Pushed.

alexsamardzic:
~An additional question:  Is it important to keep support for anything except CC 10.x?  At some point, I had kernel working on anything CC 9.0 and above, but would need to retest.~

Changed to check for 10.x only.

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
alexsamardzic: Tried that initially, and it was slower.  I reconstructed the [patch](https://github.com/user-attachments/files/26145894/diff.txt), still about 2% slower.

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
```diff
@@ -0,0 +1,947 @@
...
+            )
+
+            blocked_scale_layout = cute.make_layout((1,))
+            e_scale_stride = cutlass.Int64(0)
+            if cutlass.const_expr(BLOCKED_SCALE_OUTPUT_VALUE):
+                padded_scale_cols = cute.round_up(n_blocks, 4)
+                k_block_tiles = cute.ceil_div(K, 128)
+                n_block_tiles = padded_scale_cols // cutlass.Int64(4)
+                blocked_scale_layout = cute.make_layout(
+                    ((32, 4, k_block_tiles), (4, n_block_tiles)),
```
danielvegamyhre, line 775: curious, why isn't this something like `(rest_n, rest_k, 32, 4, 4)`? (just for my own understanding, i am newer to cute layouts)

alexsamardzic:
For the reference, here is the latest version of benchmarking script used throughout the development (script removed from PR):

<details>
<summary>Benchmarking script</summary>

```Python
import argparse
from typing import Dict, Tuple

import torch

from torchao.prototype.moe_training.kernels.mxfp8.quant import mxfp8_quantize_cuda_3d
from torchao.prototype.mx_formats.utils import from_blocked


def _parse_backends(s: str) -> list[str]:
    out = [x.strip() for x in s.split(",") if x.strip()]
    for b in out:
        if b not in {"cuda", "cutedsl"}:
            raise ValueError(f"Unsupported backend={b!r}, expected cuda/cutedsl")
    return out


def _dtype_from_str(s: str) -> torch.dtype:
    if s == "bf16":
        return torch.bfloat16
    if s == "fp32":
        return torch.float32
    raise ValueError(f"Unsupported dtype={s}")


def _tbps(num_bytes: int, ms: float) -> float:
    return num_bytes / (ms / 1e3) / 1e12


def _benchmark(fn, warmup: int, iters: int) -> float:
    for _ in range(warmup):
        fn()
    torch.cuda.synchronize()
    start = torch.cuda.Event(enable_timing=True)
    end = torch.cuda.Event(enable_timing=True)
    start.record()
    for _ in range(iters):
        fn()
    end.record()
    torch.cuda.synchronize()
    return start.elapsed_time(end) / iters


def _max_abs_diff(a: torch.Tensor, b: torch.Tensor) -> float:
    return (a.float() - b.float()).abs().max().item()


def _run_3d(args) -> None:
    dtype = _dtype_from_str(args.dtype)
    backends = _parse_backends(args.backends)
    E, N, K = args.E, args.N, args.K

    props = torch.cuda.get_device_properties(torch.cuda.current_device())
    cc = torch.cuda.get_device_capability()
    print(f"GPU: {props.name}")
    print(f"CC: {cc}")
    print(
        f"shape=(E,N,K)=({E},{N},{K}) dtype={dtype} scaling_mode={args.scaling_mode} "
        f"stage_count={args.stage_count}"
    )

    x = torch.randn((E, N, K), device="cuda", dtype=dtype) * 1000
    n_blocks = N // 32
    bytes_moved = (
        x.numel() * x.element_size()  # input
        + x.numel()
        * torch.tensor([], dtype=torch.float8_e4m3fn).element_size()  # output
        + (E * n_blocks * K)
        * torch.tensor([], dtype=torch.float8_e8m0fnu).element_size()  # scale
    )

    outs: Dict[str, Tuple[torch.Tensor, torch.Tensor]] = {}
    for b in backends:
        try:
            fn = lambda b=b: mxfp8_quantize_cuda_3d(  # noqa: E731
                x,
                block_size=32,
                scaling_mode=args.scaling_mode,
                backend=b,
                stage_count=args.stage_count,
            )
            ms = _benchmark(fn, args.warmup, args.iters)
            y, s = fn()
            outs[b] = (y, s)
            print(
                f"[{b:<10}]    {ms:.3f} ms    {_tbps(bytes_moved, ms):.3f} TB/s   "
                f"y_stride={tuple(y.stride())} s_shape={tuple(s.shape)}"
            )
        except Exception as e:
            print(f"[{b:<10}] FAILED after 0.00s: {type(e).__name__}: {e}")

    if args.check_results and "cuda" in outs:
        y_ref, s_ref = outs["cuda"]
        for b in backends:
            if b == "cuda" or b not in outs:
                continue
            y, s = outs[b]
            dy = _max_abs_diff(y_ref, y)
            if b == "cutedsl":
                s = torch.stack(
                    [
                        from_blocked(s[e], K, n_blocks).transpose(-2, -1).contiguous()
                        for e in range(E)
                    ],
                    dim=0,
                ).to(s_ref.dtype)
            ds = _max_abs_diff(s_ref, s)
            print(f"diff(cuda vs {b}): y_max_abs={dy} s_max_abs={ds}")
            ok = dy <= args.atol and ds <= args.atol
            print(f"check(cuda vs {b}): {'PASS' if ok else 'FAIL'} (atol={args.atol})")
            if not ok:
                raise RuntimeError(
                    f"Result mismatch for backend={b}: y_diff={dy}, s_diff={ds}, atol={args.atol}"
                )


def main() -> None:
    parser = argparse.ArgumentParser()
    parser.add_argument("--dtype", choices=("bf16", "fp32"), default="bf16")
    parser.add_argument("--scaling-mode", choices=("floor", "rceil"), default="floor")
    parser.add_argument("--backends", default="cuda,cutedsl")
    parser.add_argument("--warmup", type=int, default=20)
    parser.add_argument("--iters", type=int, default=100)
    parser.add_argument("--check-results", action="store_true")
    parser.add_argument("--atol", type=float, default=0.0)
    parser.add_argument("--stage-count", type=int, default=2)

    parser.add_argument("--E", type=int, default=8)
    parser.add_argument("--N", type=int, default=7168)
    parser.add_argument("--K", type=int, default=2048)

    args = parser.parse_args()
    _run_3d(args)


if __name__ == "__main__":
    main()
```
</details>

<details>
<summary>Command to run the script (if saved as <tt>bench.py</tt>)</summary>

```sh
for shape in \
"8 5120 5120" \
"8 7168 5120" \
"8 7168 7168" \
"8 8192 8192"
do
  read -r E N K <<< "$shape"
  python bench.py \
    --dtype bf16 \
    --scaling-mode rceil \
    --backends cuda,cutedsl \
    --check-results \
    --stage-count 2 \
    --E "$E" --N "$N" --K "$K"
done
```
</details>

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
danielvegamyhre: is it because we still need two "modes" for row and col dimensions to easily map the 512 byte scale factor tiles to a 2d coordinate space? so we have the `(32,4, k_block_tiles)` mode to select "which tile along the row dim", and `(4, n_block_tiles)` as "which tile along column dim" ?

alexsamardzic:
> > 2. Do we want to remove C++ kernel altogether, or we want to keep an ability to switch between "cuda"/"cutedsl" backends for now?
> 
> We should delete the cuda c++ kernel once the cutedsl is one is done
> 
> > 3. Do we want to keep the ability to save to "plain" layout, or we want to keep only the code saving to "blocked" layout?
> 
> We only need the blocked layout, we can delete the support for writing "plain" scales (row major)

I made the change for 3., not yet for 2., and CUDA C++ kernel is still the default.  The thing is, the contract for `mxfp8_quantize_cuda_3d` will change when I do 2.  Do we treat this one as for internal torchao use only?

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
alexsamardzic: > curious, why isn't this something like (rest_n, rest_k, 32, 4, 4)? (just for my own understanding, i am newer to cute layouts)

These two should be equivalent.  I guess it boils down to which one seems easier to understand.

alexsamardzic: > is it because we still need two "modes" for row and col dimensions to easily map the 512 byte scale factor tiles to a 2d coordinate space? so we have the `(32,4, k_block_tiles)` mode to select "which tile along the row dim", and `(4, n_block_tiles)` as "which tile along column dim" ?

Yes, exactly.

drisspg:
1. yeah lets make optional for now

delete c++ and only right swizzled for testing we can just `deswizzle` the swizzled output to plain if need be. And then https://github.com/pytorch/ao/blob/cd0e858c9ce51e96b57194b5c78a1c60c7705628/torchao/prototype/moe_training/mxfp8_grouped_mm.py#L447-L460 we just merge into 1 call

alexsamardzic:
> 1. yeah lets make optional for now

Updated.  The operator will report missing packages in case of failure.
 

> delete c++ and only right swizzled for testing we can just `deswizzle` the swizzled output to plain if need be. And then
>   [ ... ]
> we just merge into 1 call

Done.

alexsamardzic:
All updates made, I think this is ready for merge now.

To test/benchmark:

```
pytest test/prototype/moe_training/test_kernels.py -k cuda_mx_dim1_3d_numerics

pytest test/prototype/moe_training/test_mxfp8_grouped_mm.py -k dq_fwd_bwd

python -m benchmarks.prototype.moe_training.mxfp8.bench_quantize_3d

python -m benchmarks.prototype.moe_training.bench_2d_3d_grouped_gemm
```

danielvegamyhre:
> All updates made, I think this is ready for merge now.
> 
> To test/benchmark:
> 
> ```
> pytest test/prototype/moe_training/test_kernels.py -k cuda_mx_dim1_3d_numerics
> 
> pytest test/prototype/moe_training/test_mxfp8_grouped_mm.py -k dq_fwd_bwd
> 
> python -m benchmarks.prototype.moe_training.mxfp8.bench_quantize_3d
> 
> python -m benchmarks.prototype.moe_training.bench_2d_3d_grouped_gemm
> ```

I did some final manual tests and everything looks good, planning to land once CI is done!

danielvegamyhre:
landing, test error unrelated:

```
AttributeError: 'LinearActivationQuantizedTensor' object has no attribute 'tensor_impl'

To execute this test, run the following from the base repo dir:
    python test/dtypes/test_affine_quantized_tensor_parallel.py TestInt8dqAffineQuantizedTensorParallel.test_tp_bfloat16
    ```

PR #4091, opened 2026-03-16
Fix _is_128_128_scaled false positive for PerTensor on 128x128 linears

### LINKED ISSUE #4089: Float8DynamicActivationFloat8WeightConfig with PerTensor granularity false-positives _is_128_128_scaled on 128x128 linear
`Float8DynamicActivationFloat8WeightConfig()` with default `PerTensor` granularity on a 128x128 linear layer causes `_is_128_128_scaled` to false-positive, since the block_size ends up matching the tensor shape. The code path then incorrectly assumes 128x128 blockwise scaling is in use, which breaks `test_workflow_e2e_numerics` on `ToyLinearModel`.

`PerRow` granularity is unaffected. Not platform-specific — ran into this while enabling ROCm inference quantization tests in #4044.

```python
from torchao.quantization.quant_api import Float8DynamicActivationFloat8WeightConfig, quantize_

model = ToyLinearModel()  # contains 128x128 linear
config = Float8DynamicActivationFloat8WeightConfig()  # default PerTensor granularity
quantize_(model, config)
```

brucechanglongxu:
`_is_128_128_scaled` checks whether `block_size == (128, 128)`, but for PerTensor granularity, `block_size` is set to the full tensor shape. On a 128x128 weight this means `block_size = (128, 128)`, so the function returns True and routes to `blockwise_fp8_gemm` instead of the standard path.

The fix excludes the tensorwise case (`block_size == shape`) so only actual blockwise 128x128 scaling matches. PerRow is unaffected since its block_size is `(1, K)`.

Found while enabling Float8 inference tests on ROCm (#4044) — the test model (`ToyLinearModel`) has a 128x128 linear, and `Float8DynamicActivationFloat8WeightConfig()` with default PerTensor granularity was hitting this.

Fixes #4089

### DIFF
--- a/torchao/float8/inference.py
+++ b/torchao/float8/inference.py
@@ -210,7 +210,7 @@ def _is_128_128_scaled(x: torch.Tensor) -> bool:
     """
     assert hasattr(x, "block_size"), "Expecting input to have `block_size` attribute"
     b = x.block_size
-    return len(b) == 2 and b[0] == 128 and b[1] == 128
+    return len(b) == 2 and b[0] == 128 and b[1] == 128 and not _is_tensorwise_scaled(x)
 
 
 def _granularity_is_a_1_128_w_128_128(


brucechanglongxu:
@vkuzo This fixes the `_is_128_128_scaled` false positive from #4089 — when a weight is exactly 128×128, PerTensor granularity sets `block_size = (128, 128)` which matches the blockwise check and routes through `blockwise_fp8_gemm` instead of the standard path. The fix excludes the tensorwise case so only actual blockwise 128×128 scaling matches. Found while enabling Float8 inference tests on ROCm (#4044).

brucechanglongxu:
@vkuzo Gentle ping — this fixes the `_is_128_128_scaled` false positive you identified (PerTensor on exactly 128×128 linears matching the blockwise check). Small, targeted fix, mergeable and ready for review.

brucechanglongxu:
@vkuzo Small fix for _is_128_128_scaled false positive you called out — hoping to land by end of week.

PR #4092, opened 2026-03-16
[ROCm] Enable Int4 weight-only quantization (TILE_PACKED_TO_4D) on ROCm

brucechanglongxu:
Int4 weight-only quantization with TILE_PACKED_TO_4D hits a CUDA compute capability check that rejects ROCm GPUs, even though `_weight_int4pack_mm` is supported on MI300/MI350 via HIP (pytorch/pytorch#129710).

Three changes:

1. Gate the capability check on `torch.version.cuda` so ROCm isn't rejected. The existing TODO on that line already says "relax checks to enable quantizing in other platforms."

2. `Int4WeightOnlyConfig.int4_tile_packed_ntile` defaults to 8, but ROCm needs 16. The docstring already documents this but the code doesn't act on it. Changed the default to `None` with auto-detection in `__post_init__` (16 on ROCm, 8 on CUDA), so ROCm users don't need to pass it manually.

3. Updated `test_int4_weight_only_quant_subclass_api_grouped` to use the auto-detected ntile and skip N<16 shapes on ROCm (upstream `_weight_int4pack_mm` has a `qScaleAndZeros` shape mismatch at small N with the 16x16 MFMA tiles).

### DIFF
--- a/torchao/quantization/quant_api.py
+++ b/torchao/quantization/quant_api.py
@@ -538,7 +538,6 @@ class Int4WeightOnlyConfig(AOBaseConfig):
          currently support TINYGEMM ("tinygemm") and HQQ ("hqq"), used in version 2 only
         `set_inductor_config`: if True, adjusts `torchinductor` settings to recommended values. used in both version 1 and 2
         `version`: version of the config to use, default is 2
-        `int4_tile_packed_ntile`: ntile size for TILED_PACKED_TO_4D format, auto-detected as 8 for CUDA and 16 for ROCm if not specified
 
     Example:
 
@@ -552,15 +551,9 @@ class Int4WeightOnlyConfig(AOBaseConfig):
     int4_choose_qparams_algorithm: Int4ChooseQParamsAlgorithm = (
         Int4ChooseQParamsAlgorithm.TINYGEMM
     )
-    int4_tile_packed_ntile: Optional[int] = None
     version: int = 2
 
     def __post_init__(self):
-        if self.int4_tile_packed_ntile is None:
-            self.int4_tile_packed_ntile = 16 if torch.version.hip else 8
-        assert self.int4_tile_packed_ntile in [8, 16], (
-            "int4_tile_packed_ntile must be either 8 or 16"
-        )
         torch._C._log_api_usage_once("torchao.quantization.Int4WeightOnlyConfig")
 
 
@@ -573,7 +566,6 @@ def _int4_weight_only_quantize_tensor(weight, config):
     group_size = config.group_size
     int4_choose_qparams_algorithm = config.int4_choose_qparams_algorithm
     int4_packing_format = config.int4_packing_format
-    int4_tile_packed_ntile = config.int4_tile_packed_ntile
 
     if weight.shape[-1] % group_size != 0:
         logger.info(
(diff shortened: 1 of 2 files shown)


brucechanglongxu:
@vkuzo @danielvegamyhre This enables Int4 weight-only quantization (TILE_PACKED_TO_4D) on ROCm. Three changes: (1) gate the CUDA capability check on `torch.version.cuda` so ROCm isn't rejected, (2) auto-detect `ntile` (16 on ROCm, 8 on CUDA) in `__post_init__` so users don't need to pass it manually, (3) skip N<16 shapes on ROCm in the test due to upstream `_weight_int4pack_mm` shape mismatch with 16×16 MFMA tiles. Would appreciate a review when you have a chance.

brucechanglongxu:
@vkuzo @danielvegamyhre Gentle ping — enables Int4 weight-only quantization (TILE_PACKED_TO_4D) on ROCm. Three changes: gates the CUDA capability check, auto-selects FNUZ dtypes on AMD, and removes ROCm test skips. Mergeable and ready for review.

torchao/quantization/quant_api.py
```diff
@@ -768,10 +768,12 @@ class Int4WeightOnlyConfig(AOBaseConfig):
     int4_choose_qparams_algorithm: Int4ChooseQParamsAlgorithm = (
         Int4ChooseQParamsAlgorithm.TINYGEMM
     )
-    int4_tile_packed_ntile: int = 8
+    int4_tile_packed_ntile: Optional[int] = None
```
jerryzh168: I actually didn't see people really change this argument for CUDA, does rocm set this to different values? if not I think we could just hardcode this and remove the arg

brucechanglongxu: Yes, ROCm requires a different value — `_weight_int4pack_mm` on HIP uses ntile=16 while CUDA uses ntile=8. The auto-detection in `__post_init__` avoids users needing to know this platform detail; without it, ROCm users would need to manually pass `int4_tile_packed_ntile=16` every time.

That said, if you'd prefer to keep the API simpler, I could hardcode it as `int = 8` and override to 16 internally when `torch.version.hip` is set (moving the platform logic into `from_hp()` or the quantization path rather than the config). Let me know which approach you prefer.

brucechanglongxu:
@jerryzh168 Rebased after your approval; replied on the ntile thread. OK to merge when CI looks good on your side.

torchao/quantization/quant_api.py
jerryzh168: yeah I think we can just hardcode this at the time when we need to use this value, no need to expose to config or store in the tensor if neither cuda nor hip are going to need multiple values

brucechanglongxu:
@jerryzh168 Addressed the ntile feedback: dropped `int4_tile_packed_ntile` from `Int4WeightOnlyConfig` and resolve 8 vs 16 inside `Int4TilePackedTo4dTensor.from_hp` when `ntile_size` is unset (CUDA 8, HIP 16). No config field for this anymore. Branch rebased and pushed.

brucechanglongxu:
@jerryzh168 Could you re-review when you have time? `int4_tile_packed_ntile` is removed from the config; ntile defaults are applied inside `Int4TilePackedTo4dTensor.from_hp` only (8 CUDA / 16 HIP).

brucechanglongxu:
@jerryzh168 You approved earlier; latest ntile changes are on the branch. Trying to land by end of week — OK to merge if CI is green on your side?

Issue #4097, opened 2026-03-17
aten._has_compatible_shallow_copy_type', overload='default' dispatcher failed for LinearActivationQuantizedTensor in inference mode


xiaowangintel:
Describe the bug:
Reproducer:
```
import copy

import torch
import torch.nn as nn
from torchao.quantization import quantize_

from transformers import (
    AutoConfig,
    AutoModelForCausalLM,
    AutoTokenizer,
    GenerationConfig,
)
from torchao.quantization import quantize_, Int8DynamicActivationInt8WeightConfig
from transformers import TorchAoConfig

qconfig = TorchAoConfig(Int8DynamicActivationInt8WeightConfig())
model_id = "meta-llama/Llama-3.1-8B-Instruct"
load_dtype = torch.bfloat16
device = "cuda"
config = AutoConfig.from_pretrained(model_id, trust_remote_code=False)

with torch.inference_mode():
    model = AutoModelForCausalLM.from_pretrained(
        model_id,
        torch_dtype=load_dtype,
        config=config,
        low_cpu_mem_usage=True,
        trust_remote_code=False,
        device_map=device,
        quantization_config=qconfig,
        attn_implementation="sdpa",
    )

```

Error Information:
```
  File "/ao/torchao/utils.py", line 684, in _dispatch__torch_dispatch__
    raise NotImplementedError(
NotImplementedError: LinearActivationQuantizedTensor dispatch: attempting to run unimplemented operator/function: func=<OpOverload(op='aten._has_compatible_shallow_copy_type', overload='default')>, types=(<class 'torchao.quantization.linear_activation_quantized_tensor.LinearActivationQuantizedTensor'>,), arg_types=(<class 'torchao.quantization.linear_activation_quantized_tensor.LinearActivationQuantizedTensor'>, <class 'torchao.quantization.linear_activation_quantized_tensor.LinearActivationQuantizedTensor'>), kwarg_types={}
```

Versions:
```
Collecting environment information...
PyTorch version: 2.10.0+cu130
Is debug build: False
CUDA used to build PyTorch: 13.0
ROCM used to build PyTorch: N/A

OS: Ubuntu 22.04.3 LTS (x86_64)
GCC version: (Ubuntu 11.4.0-1ubuntu1~22.04.3) 11.4.0
Clang version: Could not collect
CMake version: Could not collect
Libc version: glibc-2.35

[...]

PR #4098, opened 2026-03-17
Fix aten._has_compatible_shallow_copy_type', overload='default' dispatcher failed for LinearActivationQuantizedTensor in inference mode

xiaowangintel:
Fixes https://github.com/pytorch/ao/issues/4097
### Summary
This PR fixes a bug in torchao/quantization/linear_activation_quantized_tensor.py where LinearActivationQuantizedTensor without dispatch of aten._has_compatible_shallow_copy_type.

### DIFF
--- a/torchao/quantization/linear_activation_quantized_tensor.py
+++ b/torchao/quantization/linear_activation_quantized_tensor.py
@@ -301,6 +301,12 @@ def _(func, types, args, kwargs):
     )
 
 
+@implements_torch_function(torch._has_compatible_shallow_copy_type)
+def _(func, types, args, kwargs):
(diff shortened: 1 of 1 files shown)


jerryzh168:
@xiaowangintel this is part of our deprecated stack, wondering when do you see this error? which config are you using

xiaowangintel:
@jerryzh168 Thank you for your reply. The default version is version 1 in TorchAO, so we both support the two version. It can execute normally when provided with Int8DynamicActivationInt8WeightConfig(version=2). We will focus on version 2 later.

xiaowangintel:
@liangan1 FYI

PR #4100, opened 2026-03-17
Deprecate TorchAODType

andrewor14:
**Summary:** This was added for torch 2.5 as a placeholder for torch.int4. Now all torch versions we support have these dtypes, so this class is no longer needed and should not be used by anyone. We will remove it next release (0.18.0).

**Test Plan:** Manual testing

**Deprecation notes:**

Before:
```
TorchAODType.INT1
TorchAODType.INT2
TorchAODType.INT3
TorchAODType.INT4
TorchAODType.INT5
TorchAODType.INT6
TorchAODType.INT7
```

After:
```
torch.int1
torch.int2
torch.int3
torch.int4
torch.int5
torch.int6
torch.int7
```

### DIFF
--- a/torchao/quantization/quant_primitives.py
+++ b/torchao/quantization/quant_primitives.py
@@ -5,7 +5,8 @@
 # LICENSE file in the root directory of this source tree.
 
 import math
-from enum import Enum, auto
+import warnings
+from enum import Enum, EnumMeta, auto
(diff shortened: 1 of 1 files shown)


jerryzh168:
@andrewor14 might be good to add a deprecate note in the summary as well so you can copy paste when we do the release notes

andrewor14:
Ok added

PR #4103, opened 2026-03-18
[X86] intmm: Use u8s8 when only support avx512-vnni

cyxlily:
Use u8s8 matmul for intmm on X86 CPU when only support avx512-vnni for better performance.


torchao/kernel/intmm.py
```diff
@@ -133,8 +133,16 @@ def int_scaled_matmul(
     if check_cpu_version(scales1.device):
         # CPU prefers decomposed version of int_scaled_matmul
         # to leverage the fusion capability of Inductor
-        c = torch._int_mm(a, b)
-        return c.to(scales1.dtype) * scales1
+
+        if not torch.cpu._is_amx_tile_supported() and torch.cpu._is_vnni_supported():# uint8 path
```
Xia-Weiwen: A potential lint issue here: no spaces before comment. Did you run pre-commit?

Xia-Weiwen: Please
1. extract the CPU part as a function and call it here.
2. Explain more details in comments:
    1. For platforms with AVX512_VNNI support but without AMX, we convert to u8s8 to use AVX512_VNNI instructions for better performance.
    2. For other platforms, s8s8 computation is done with AMX or reference implementation.

Xia-Weiwen:
@claude review

torchao/kernel/intmm.py
```diff
@@ -105,6 +105,35 @@ def int_matmul(a: torch.Tensor, b: torch.Tensor) -> torch.Tensor:
...
+        a (torch.Tensor): The first matrix to multiply (int8).
+        b (torch.Tensor): The second matrix to multiply (int8).
+        scales1 (torch.Tensor): The scaling factors, shape (M, 1).
+
+    Returns:
+        torch.Tensor: The result of the scaled matrix multiplication.
+    """
+    if (
+        not torch.cpu._is_amx_tile_supported() and torch.cpu._is_vnni_supported()
+    ):  # u8s8: convert to uint8 to use AVX512_VNNI instructions for better performance
```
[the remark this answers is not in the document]
Xia-Weiwen: Yeah. @cyxlily please check attributes like `hasattr(torch.cpu, "_is_vnni_supported")`

torchao/kernel/intmm.py
```diff
@@ -105,6 +105,35 @@ def int_matmul(a: torch.Tensor, b: torch.Tensor) -> torch.Tensor:
...
+        # on platforms with AVX512_VNNI support but without AMX
+        a = (a.to(torch.int32) + 128).to(torch.uint8)
+        c = torch._int_mm(a, b)
+        comp = b.sum(dim=0, keepdim=True, dtype=torch.int32) * 128
+        c.sub_(comp)
+        return c.to(scales1.dtype) * scales1
```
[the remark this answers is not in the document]
Xia-Weiwen: @cyxlily Please consider adding a test case by monkeypatching

cyxlily:
@claude I have updated the codes. Review again.

Xia-Weiwen:
@claude codes are updated per your (and copilot's) comments. Review again.

torchao/kernel/intmm.py
```diff
@@ -127,14 +186,12 @@ def int_scaled_matmul(
     assert M == scales1.size(0) or scales1.numel() == 1
     assert 1 == scales1.size(1)
     assert scales1.is_contiguous()
-    scales1 = scales1.expand((M, N))
     assert scales1.dim() == 2
```
jerryzh168: should the assert be moved as well, previusly this is after expand

test/kernel/test_autotuner.py
```diff
@@ -96,5 +96,88 @@ def test_int_scaled_mm(self, device, dtype):
         torch.testing.assert_allclose(out32_1, out32_2)
 
 
+class TestIntScaledMatmulCPUPaths(unittest.TestCase):
+    """
+    Tests for the CPU-specific paths inside _int_scaled_matmul_cpu.
+    Because the u8s8 VNNI branch is gated on runtime CPU feature detection,
+    CI machines are unlikely to exercise it naturally.  We monkeypatch the
+    two helper functions so each branch can be tested on any machine.
```
jerryzh168, line 104: this seems a bit confusing, you mean you have to monkeypatch the `_cpu_is_amx_tile_supported` and `_cpu_is_vnni_supported` to run the test, I thought these have to reflect what the hardware is doing? what are the flags of the CI machines before monkeypatch? and did you only change the flag from True to False to test reference?

Xia-Weiwen: Thanks for reviewing. This is to ensure both paths are tested since the hardware for CI most likely does not have these ISA support. These paths work on all platforms while performance is the best with certain ISA support.

torchao/kernel/intmm.py
cyxlily: The CPU also needs to check that scales1.dim() == 2, otherwise calculation with 3D scales1 will also be incorrect. And if scales1 is 1D, 'assert 1 == scales1.size(1)' will cause an error, so I think it's better to move the check for scales1.dim() == 2 at the beginning.

cyxlily:
@jerryzh168 Could you review again?

test/kernel/test_autotuner.py
jerryzh168: so if hardware doesn't have instruction support:
_cpu_is_amx_tile_supported = False and _cpu_is_vnni_supported = False

and you set one of these to True, e.g. _cpu_is_amx_tile_supported = True, what happens?

can you provide more details on these in each of the test?

Xia-Weiwen: Setting these flags in different ways will lead to s8s8 or u8s8 paths (according to the rules we defined in torchao/kernel/intmm.py). Both paths work on all platforms because `torch._int_mm` for CPU calls oneDNN under the hood and oneDNN takes care of everything. This is for test only to ensure both paths work as expected in terms of functionality.

Xia-Weiwen: Otherwise only one path will be tested in CI. The purpose is to test both paths in CI to ensure they work.

torchao/kernel/intmm.py
```diff
@@ -124,17 +183,15 @@ def int_scaled_matmul(
...
     M, K = a.shape
     K, N = b.shape
+    assert scales1.dim() == 2
     assert M == scales1.size(0) or scales1.numel() == 1
     assert 1 == scales1.size(1)
     assert scales1.is_contiguous()
-    scales1 = scales1.expand((M, N))
-    assert scales1.dim() == 2
 
     if check_cpu_version(scales1.device):
```
jerryzh168, line 193: `check_cpu_version` seems too vague I feel, maybe just something like `is_device_type(scales1.device, "cpu")`

torchao/kernel/intmm.py
```diff
@@ -124,17 +183,15 @@ def int_scaled_matmul(
     """
     M, K = a.shape
     K, N = b.shape
+    assert scales1.dim() == 2
```
jerryzh168, line 188: are you sure to check this before expand? does orignal op work with 1d scale

Xia-Weiwen: > are you sure to check this before expand? does orignal op work with 1d scale

It does not work with 1d scale. the `assert 1 == scales1.size(1)` below is not added by us and it assumes  scales1 is 2d but previously it was only checked after that. So, we think it is better to check that at the beginning.

torchao/kernel/intmm.py
Xia-Weiwen: > `check_cpu_version` seems too vague I feel, maybe just something like `is_device_type(scales1.device, "cpu")`

Thanks for the suggestion. However, this utility is not added by this PR. It is defined here: https://github.com/pytorch/ao/blob/9051a2f0a16cd6f75da48c62fb713b3c61be38fa/torchao/utils.py#L1242 How about fixing it in another PR?

torchao/kernel/intmm.py
```diff
@@ -105,6 +105,63 @@ def int_matmul(a: torch.Tensor, b: torch.Tensor) -> torch.Tensor:
...
+        return torch.cpu._is_vnni_supported()
+    return False
+
+
+def _int_scaled_matmul_cpu(
+    a: torch.Tensor, b: torch.Tensor, scales1: torch.Tensor
+) -> torch.Tensor:
+    """
+    CPU-optimized path for scaled integer matrix multiplication.
+    CPU prefers decomposed version to leverage the fusion capability of Inductor.
```
Xia-Weiwen, line 139: Please clarify here: `It goes to s8s8 or u8s8 path according to ISA support of hardware. The selection is for performance only and both paths should work regardless of ISA support.`

jerryzh168 (approved):

torchao/kernel/intmm.py
jerryzh168: yeah it should be fixed in a separate PR

Xia-Weiwen:
CI failures are unrelated.

PR #4109, opened 2026-03-18
Fix deprecated TreeSpec CTOR

JacobSzwejbka:
Getting absolutely spammed with "/Users/jakeszwe/anaconda3/envs/opt/lib/python3.12/copyreg.py:99: FutureWarning: `isinstance(treespec, LeafSpec)` is deprecated, use `isinstance(treespec, TreeSpec) and treespec.is_leaf()` instead." 

when lowering models in ET. Seeing if this fixes.


JacobSzwejbka:
Ah regression tests caught that ao runs on older torch versions so added the dumb try except for now.

Issue #4054, opened 2026-03-11
Triton -> Helion for blockwise quantization kernel improvements


danielvegamyhre:
1. Run benchmark scripts for quantization kernels (not the gemm kernels): https://github.com/pytorch/ao/tree/main/benchmarks/prototype/blockwise_fp8_training
2. Document memory bandwidth utilization achieved on H100 for each with [DSV3 671b shapes](https://github.com/pytorch/torchtitan/blob/ea4989e0dfc15d214e69c9569b358da322087f46/torchtitan/models/deepseek_v3/__init__.py#L135-L136)
    - Use 32768 and 131072 as the M dimension (local batch size * sequence length) to simulate medium vs large local batch sizes
3. Identify weakest quantization kernels, prototype Helion alternatives with unit tests and benchmarks.

iamzainhuda:
with bandwidth benchmarking script, on H100 SXM5 variant, we achieve following mem bandwidth utiil%. we apply 92% as achievable bandwidth percent (as observed in meta h100 variant under roofline_utils). we see util above 80% for these kernels across M=32768 and M=131072.

<img width="871" height="336" alt="Image" src="https://github.com/user-attachments/assets/68b61ac6-901b-45da-bc9a-6e2f567551df" />

danielvegamyhre:
Great, I think we have good coverage for v0 of this feature then, we can potentially optimize further later on if needed. Thanks for investigating and validating this. Let's close this for now and build out the rest of it!

PR #4112, opened 2026-03-19
add bandwidth benchmarking for fp8 quant blockwise

iamzainhuda:
## Summary
As part of #4054, this adds a kernel-only bandwidth benchmark for the blockwise FP8 quantization kernels. The benchmark preallocates outputs, validates the low-level launch path against the public wrapper outputs, times the Triton kernel launches with L2 flushing between iterations, and reports effective logical IO bandwidth against peak and achievable device-bandwidth references.

It also adds benchmarks/prototype/blockwise_fp8_training/README.md documenting the benchmark methodology and current H100 results.

To run
```
python -m benchmarks.prototype.blockwise_fp8_training.benchmark_quant_kernel_bandwidth
```

CLI usage:                                                                                                                                          
  - ```--m-values 32768 131072``` sets the M shapes to benchmark                                                                                            
  - ```--k 4096``` sets the shared feature dimension                                                                                                           
  - ```--block-size 128``` sets the quantization block size                                                                                                 
  - ```--csv /path/to/results.csv``` writes the results table to CSV                                                                                       
  - ```--use-roofline-utils``` uses the static roofline_utils bandwidth reference instead of CUDA device properties

### Per-shape Results
Tested with shapes 32768 and 131072 to reflect real world training:

| kernel | shape | kernel_us | effective_logical_io_gbps | logical_io_vs_peak_% | logical_io_vs_achievable_% |
|---|---|---:|---:|---:|---:|
| act_quant_transposed_lhs | 32768x4096 | 154.46 | 2633.9 | 78.6 | 85.4 |
| weight_quant_transposed_rhs | 32768x4096 | 150.53 | 2675.2 | 79.8 | 86.7 |
| act_quant_lhs | 32768x4096 | 150.86 | 2696.8 | 80.4 | 87.4 |
| act_quant_rhs | 32768x4096 | 148.70 | 2736.0 | 81.6 | 88.7 |
| weight_quant_rhs | 32768x4096 | 144.99 | 2777.3 | 82.8 | 90.1 |
| weight_quant_transposed_rhs | 131072x4096 | 581.89 | 2768.1 | 82.6 | 89.8 |
| act_quant_lhs | 131072x4096 | 586.98 | 2772.5 | 82.7 | 89.9 |
| act_quant_transposed_lhs | 131072x4096 | 581.47 | 2798.7 | 83.5 | 90.7 |
| act_quant_rhs | 131072x4096 | 562.56 | 2892.8 | 86.3 | 93.8 |
| weight_quant_rhs | 131072x4096 | 555.30 | 2900.7 | 86.5 | 94.1 |

### DIFF
--- a/benchmarks/prototype/blockwise_fp8_training/benchmark_quant_kernel_bandwidth.py
+++ b/benchmarks/prototype/blockwise_fp8_training/benchmark_quant_kernel_bandwidth.py
@@ -11,24 +11,15 @@ from pathlib import Path
 from typing import Callable, Iterable, List, Optional, Tuple
 
 import torch
-import triton
 from tabulate import tabulate
-from torch.library import wrap_triton
 
-from torchao.float8.config import e4m3_dtype
-from torchao.kernel.autotuner import do_bench_triton
+from benchmarks.utils import benchmark_cuda_function_in_microseconds
 from torchao.prototype.blockwise_fp8_training.kernels import (
-    EPS,
     triton_fp8_blockwise_act_quant_lhs,
-    triton_fp8_blockwise_act_quant_lhs_kernel,
     triton_fp8_blockwise_act_quant_rhs,
-    triton_fp8_blockwise_act_quant_rhs_kernel,
     triton_fp8_blockwise_act_quant_transposed_lhs,
-    triton_fp8_blockwise_act_quant_transposed_lhs_kernel,
     triton_fp8_blockwise_weight_quant_rhs,
-    triton_fp8_blockwise_weight_quant_rhs_kernel,
     triton_fp8_blockwise_weight_quant_transposed_rhs,
-    triton_fp8_blockwise_weight_quant_transposed_rhs_kernel,
 )
 from torchao.testing.training.roofline_utils import gpu_name_to_specs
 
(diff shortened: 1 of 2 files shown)


iamzainhuda:
changing benchmarking to time kernel execution only, current times include wrapper overhead

benchmarks/prototype/blockwise_fp8_training/benchmark_quant_kernel_bandwidth.py
```diff
@@ -0,0 +1,745 @@
...
+    m, k = x.size()
+    fp8_max = torch.finfo(buffers.y.dtype).max
+
+    def grid(meta):
+        return (
+            triton.cdiv(m, meta["NUM_GROUPS"]),
+            triton.cdiv(k, meta["BLOCK_SIZE"]),
+        )
+
+    wrap_triton(triton_fp8_blockwise_act_quant_lhs_kernel)[grid](
```
danielvegamyhre: why are we defining launchcers and wrapping the kernels directly, we shouldn't need to do that, all the kernels already have launcherse i believe https://github.com/pytorch/ao/blob/5ee9094f6715ebb509c4306ac432a50cba4b5a0e/torchao/prototype/blockwise_fp8_training/kernels.py#L459

iamzainhuda: i had initially done it like this to measure strictly the triton launch into the preallocated output since i considered the additional work under the launcher as "overhead" (input checks, output alloc, etc) that shouldn't be counted. 

i've changed it to time the kernel through launcher already defined like the other benchmarking scripts

danielvegamyhre:
Thanks for adding this @iamzainhuda !

PR #4117, opened 2026-03-19
[mxfp8] fix exp2f_rcp to handle special case of unbiased exponent of 127

danielvegamyhre:
Copying updated `exp2f_rcp` from [here](https://github.com/NVIDIA/TransformerEngine/blob/b7598aa887eb7d619d64c90692980009669379bf/transformer_engine/common/util/ptx.cuh#L332-L341) to properly handle case of unbiased exponent  of 127.

## Build test
- `USE_CPP=1 pip install --no-build-isolation -e ~/ao -v`

### DIFF
--- a/test/prototype/mx_formats/test_kernels.py
+++ b/test/prototype/mx_formats/test_kernels.py
@@ -612,42 +612,6 @@ def test_cuda_mx_dim1_numerics(M, K, input_dtype, scaling_mode):
     not is_cuda_version_at_least(12, 8),
     reason="CUDA version >= 12.8 required for MXFP8 CUDA kernels",
 )
-def test_cuda_mxfp8_rceil_exp2f_rcp_unbiased_exp_127():
-    """
-    Test edge case that occurs when the unbiased exponent is 127, requiring
-    the reciprocal to be 2^-127.
-    """
-    device = "cuda"
-    scaling_mode_str = "rceil"  # exp2f_rcp only is called for rceil scale rounding mode
-
-    value = float("inf")
-
-    M, K = 64, 64
-    input_tensor = torch.full((M, K), value, dtype=torch.float32, device=device)
-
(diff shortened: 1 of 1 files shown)


vkuzo:
can we have a test?

danielvegamyhre:
> can we have a test?

@vkuzo @slayton58 correct me if i'm wrong but i don't think this bug is reachable from our pytorch code? to reach a biased scale of 254 we need a very large value with exponent bits of 254, and we only can pass bf16 or fp32 (same dynamic range, so let's use fp32). 

the max fp32 value is [(2 - (2**-23)) * (2**127)](https://en.wikipedia.org/wiki/Single-precision_floating-point_format) which has exponent 254. However, when using RCEIL scaling (only path that uses `exp2f_rcp`) this is [divided by 448 and converted with the cvt.sat.finite inline ptx](https://github.com/pytorch/ao/blob/5ee9094f6715ebb509c4306ac432a50cba4b5a0e/torchao/csrc/cuda/mx_kernels/mxfp8_quantize.cuh#L534) and this results in `scale.view(uint8)` of 247, which is below the target value of 254 to hit this code path. 

tl;dr We can't create large enough test inputs for the kernel to hit this path since they aren't representable in fp32 or bf16.

danielvegamyhre:
@vkuzo @slayton58 ok i determined we can pass `"inf"` which will result in biased exponent of 254 and returns `scales.view(uint8) == 254`. So i guess this case isn't reachable via finite values but is special case for infinity?

slayton58:
> @vkuzo @slayton58 ok i determined we can pass `"inf"` which will result in biased exponent of 254 and returns `scales.view(uint8) == 254`. So i guess this case isn't reachable via finite values but is special case for infinity?

Sounds about right - even looking at MXFP4 we'd still only hit 252, which is still < 254. I think we'd need a format where `max(fmt) == 1` to hit this, which we don't have.

PR #4125, opened 2026-03-20
Fix test_sparse_api failures for builds without hipSPARSELt (#4125)

bbeckca:
Summary:

What: Add a guard in test/sparsity/test_sparse_api.py so tests that depend on `hipSPARSELt` only run when the backend is available.

Why: These tests currently fail on environments without hipSPARSELt, for example OSS AO nightly. Since the functionality is environment dependent, these test should be conditionally skipped. https://fburl.com/9fpszdqg

```
FAILED test/sparsity/test_sparse_api.py::TestSemiStructuredSparse::test_sparse - RuntimeError: hipSPARSELt not supported on your machine.
```

Differential Revision: D97540060

### DIFF
--- a/test/sparsity/test_sparse_api.py
+++ b/test/sparsity/test_sparse_api.py
@@ -29,6 +29,8 @@ class TestSemiStructuredSparse(common_utils.TestCase):
     @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
     @skipIfRocmVersionLessThan((7, 0))
     def test_sparse(self):
+        if not torch.backends.cusparselt.is_available():
+            self.skipTest("Need cuSPARSELt or hipSPARSELt")
         input = torch.rand((128, 128)).half().cuda()
         model = (
             nn.Sequential(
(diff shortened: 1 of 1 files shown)


test/sparsity/test_sparse_api.py
```diff
@@ -46,9 +48,6 @@ def test_sparse(self):
         sparsify_(model, semi_sparse_weight())
         sparse_result = model(input)
 
-        if compile:
```
jcaip, line 49: why are we deleting the compile test?

bbeckca: We're not setting any `@common_utils.parametrize("compile", [True, False])` so I don't think this path was being used. Would you prefer we enable it as opposed to removing?

jcaip: Oh I see, hmm can we add @common_utils.parametrize("compile", [True, False]) to the test as well then?

bbeckca: Update: Aligned offline. @jcaip will update compile test in follow up diff.

PR #4132, opened 2026-03-22
[ROCm] Enable FP8 QAT test on MI300/MI350

brucechanglongxu:
The FP8 QAT test (`test_quantize_api_fp8_fp8`) was gated on `is_sm_at_least_89()`, which excluded AMD GPUs. MI300 and MI350 both support FP8 dynamic activation + FP8 weight quantization with PerTensor and PerRow granularity, so this updates the skip condition to also allow MI300/MI350.

Tested locally on MI300X -- both PerTensor and PerRow variants pass.

### DIFF
--- a/test/quantization/test_qat.py
+++ b/test/quantization/test_qat.py
@@ -100,6 +100,8 @@ from torchao.utils import (
     _is_mslk_available,
     get_current_accelerator_device,
     is_fbcode,
+    is_MI300,
+    is_MI350,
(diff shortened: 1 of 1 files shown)


brucechanglongxu:
@andrewor14 Thanks for the approval! This is mergeable and CI is green (the CUDA nightly failure looks like a pre-existing infra issue). Could you merge when you get a chance?

PR #4136, opened 2026-03-22
[ROCm] Fix BlockwiseQuantLinear FP8 dtype handling for AMD GPUs

brucechanglongxu:
`BlockwiseQuantLinear` doesn't work on MI300 — the dtype whitelist rejects fnuz variants, the default is hardcoded to `e4m3fn`, and there's a bug where `self.dtype` reads the class attribute (`bfloat16`) instead of storing the constructor arg, so the forward pass would always blow up in `fp8_blockwise_act_quant`.

Fixed by adding fnuz dtypes to the accepted set, defaulting to `e4m3_dtype` from config (picks the right variant per hardware), and storing the dtype properly as `self.fp8_dtype`.

Tested on MI300X — all four FP8 dtype variants construct fine, invalid dtypes get rejected, forward pass works.

### DIFF
--- a/torchao/prototype/blockwise_fp8_inference/blockwise_linear.py
+++ b/torchao/prototype/blockwise_fp8_inference/blockwise_linear.py
@@ -7,25 +7,33 @@
 import torch
 from torch import nn
 
+from torchao.float8.config import e4m3_dtype
 from torchao.kernel.blockwise_quantization import (
     blockwise_fp8_gemm,
     fp8_blockwise_act_quant,
 )
 
+_SUPPORTED_FP8_DTYPES = [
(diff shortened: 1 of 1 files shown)


brucechanglongxu:
@vkuzo @danielvegamyhre This fixes FP8 dtype handling in `BlockwiseQuantLinear` for AMD GPUs — auto-selects FNUZ variants on ROCm so blockwise inference doesn't silently use unsupported IEEE FP8 dtypes. Mergeable, would appreciate a review.

brucechanglongxu:
@vkuzo @danielvegamyhre Hoping to land the ROCm stack PRs by end of week — could one of you review when you have bandwidth?

PR #4141, opened 2026-03-23
[mxfp8 training] require torch nightly and cuda 12.8+

danielvegamyhre:
## Summary
- #4010 depends on changes in pytorch core that recently landed, so we need to assert the user has torch nightly installed for MXFP8 training

### DIFF
--- a/torchao/prototype/moe_training/conversion_utils.py
+++ b/torchao/prototype/moe_training/conversion_utils.py
@@ -30,6 +30,16 @@ def _get_tensor_cls_for_config(
     )
 
     if isinstance(config, MXFP8TrainingOpConfig):
+        from torch.distributed.tensor import _dispatch, _ops
+
+        pytorch_version_supported = hasattr(
+            _ops, "scaled_mm_single_dim_strategy"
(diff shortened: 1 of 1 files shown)


torchao/prototype/moe_training/conversion_utils.py
```diff
@@ -30,6 +30,11 @@ def _get_tensor_cls_for_config(
     )
 
     if isinstance(config, MXFP8TrainingOpConfig):
+        # MXFP8 training requires torch nightly build for cuda 12.8+
+        assert "dev" in torch.__version__ and "cu12" in torch.version.cuda, (
```
vkuzo: would this work on any nightly (even before the PT core fix)? Also, would this break on stable when the corresponding PyTorch stable version is released?  Ideally this should:
1. fail when on a version of PyTorch without the core fix
2. pass on nightly, stable or built-from-source versions of PyTorch with the core fix

danielvegamyhre: i updated it to check if pytorch installation has certain attributes that indicate the necessary PRs were included

PR #4074, opened 2026-03-13
Deprecate AQT and related classes

andrewor14:
**Summary:** Deprecate AffineQuantizedTensor, AQTTensorImpl, Layout, and all subclasses of the above in torchao.dtypes. We are planning to remove these classes in the future, so we deprecate them here in advance.

```
● Here are all the classes that now have new deprecation warnings:

  Base classes (torchao/dtypes/utils.py):
  1. Layout
  2. PlainLayout
  3. AQTTensorImpl

  torchao/dtypes/affine_quantized_tensor.py:
  4. AffineQuantizedTensor

  torchao/dtypes/floatx/:
  5. Float8Layout
  6. Float8AQTTensorImpl
  7. CutlassSemiSparseLayout
  8. CutlassSemiSparseTensorImpl

  torchao/dtypes/uintx/:
  9. TensorCoreTiledLayout
  10. TensorCoreTiledAQTTensorImpl
  11. SemiSparseLayout
  12. SemiSparseAQTTensorImpl
  13. Int4CPULayout
  14. Int4CPUAQTTensorImpl
  15. Int4XPULayout
  16. Int4XPUAQTTensorImpl
  17. QDQLayout
  18. QDQTensorImpl
  19. PlainAQTTensorImpl
  20. PackedLinearInt8DynamicActivationIntxWeightLayout
  21. PackedLinearInt8DynamicActivationIntxWeightAQTTensorImpl
```

More context here: https://github.com/pytorch/ao/issues/2752

**Test Plan:**
Manual testing

PR #4151, opened 2026-03-23
Delete deprecated PlainLayout, PlainAQTTensorImpl and related v1 code paths

jerryzh168:
* __->__ #4151

- Remove PlainLayout class from dtypes/utils.py
- Delete torchao/dtypes/uintx/plain_layout.py
- Remove int8 weight and int8 dynamic activation dispatch from AQT dispatch table
- Remove AQT embedding dispatch (used PlainAQTTensorImpl)
- Change Int8WeightOnlyConfig default to version=2 (removes v1 AQT path)
- Change Int8DynamicActivationInt8WeightConfig default to version=2 (removes v1 AQT path)
- Remove PlainLayout from public exports
- Update tests to use v2 tensor types


andrewor14:
Technically these were only deprecated recently (as of #4074). We can still delete them now, just want to point out it hasn't been 1 release yet

jerryzh168:
> Technically these were only deprecated recently (as of #4074). We can still delete them now, just want to point out it hasn't been 1 release yet

I think it's fine since the v1 config has been deprecated for a long time

PR #4154, opened 2026-03-23
Remove version compatibility table

andrewor14:
**Summary:** As of https://github.com/pytorch/ao/issues/3516, torchao is now ABI stable, so we no longer need this version compatibility table. We do require torch 2.11+ because the mxfp8 kernels rely on stable ABI features added in torch 2.11.

Note: there are some leftover kernels under torchao/csrc/cpu and rocm that are not ABI stable yet. These would require additional ABI stable support from PyTorch and so are left for the future.

**Test Plan:** Manual testing

### DIFF
--- a/torchao/__init__.py
+++ b/torchao/__init__.py
@@ -49,6 +49,7 @@ def _parse_version(version_string):
 
 
 skip_loading_so_files = False
+_skip_reason = None
 force_skip_loading_so_files = (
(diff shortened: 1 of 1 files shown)


jerryzh168 (approved):
LGTM, for people who want to use the few kernels in torchao like mxfp8, they have to upgrade torch to >= 2.11.0 now. I assume this is OK since most of torchao features don't require these kernels.

PR #4156, opened 2026-03-24
[mxfp8 training] add cutedsl kernel for mxfp8 quantation along dim0

danielvegamyhre:
## Summary
- Created 2d version of the 3d weight quantization kernel from @alexsamardzic. It accepts 2d tensor in row major layout of shape (M,K) and each CTA iterates along K 
- Added tests and benchmark scripts comparing it to the current 2 stage approach (quant kernel -> blocked layout kernel)

## Tests
- `pytest test/prototype/moe_training/test_kernels.py -s -k 2d_numerics`
- `pytest test/prototype/moe_training/test_mxfp8_grouped_mm.py -k dq_fwd_bwd`

## Benchmarks (writing scales directly to blocked layout for tcgen05.mma)
- 1.29x to 2.88x faster than existing for local batch size = 1 (2.8 to 4.2 tb/s)
- 1.08x to 1.18x faster than existing for local batch size = 4 (4.4 to 5.8 tb/s)
- 1.10x to 1.24x faster than exisitng for local batch size = 16 (5.8 to 6.4 tb/s)

```
input_shape     scaling_mode      num_groups    cutedsl_blocked_us    triton+rearrange_us  speedup      cutedsl_gbps    triton+rearrange_gbps
--------------  --------------  ------------  --------------------  ---------------------  ---------  --------------  -----------------------
(8192, 2048)    floor                      8                 18.03                  51.14  2.84x              2820.3                    994.5
(8192, 2048)    rceil                      8                 17.47                  50.37  2.88x              2910.7                   1009.7
(8192, 7168)    floor                      8                 42.02                  55.3   1.32x              4236.4                   3219
(8192, 7168)    rceil                      8                 42.02                  54.3   1.29x              4236.4                   3277.8
(32768, 2048)   floor                      8                 46.14                  54.27  1.18x              4408.5                   3748.2
(32768, 2048)   rceil                      8                 46.08                  54.27  1.18x              4414.6                   3748.2
(32768, 7168)   floor                      8                125.95                 136.38  1.08x              5652.8                   5220.4
(32768, 7168)   rceil                      8                121.89                 136.19  1.12x              5841.3                   5227.8
(131072, 2048)  floor                      8                140.32                 154.69  1.10x              5798.9                   5260.2
(131072, 2048)  rceil                      8                136.22                 154.69  1.14x              5973.2                   5260.2
(131072, 7168)  floor                      8                457.76                 556.1   1.21x              6221.5                   5121.3
(131072, 7168)  rceil                      8                443.46                 547.84  1.24x              6422.1                   5198.5
```

## Next steps
- Fused per-group padding into quantization kernel for the "non-HybridEP" and "non-EP" cases. I have been working on this but it is difficult to get good performance so far.

### DIFF
--- a/torchao/prototype/moe_training/kernels/mxfp8/cute_utils.py
+++ b/torchao/prototype/moe_training/kernels/mxfp8/cute_utils.py
@@ -170,7 +170,7 @@ if _cutedsl_runtime_available():
         return scale_biased, inv_scale
 
     @cute.jit
-    def load_vals_4B_chunk_full(
+    def load_vals_chunk_full(
         vals_block: cute.Tensor,
         local_base: cutlass.Int32,
     ):
@@ -187,13 +187,13 @@ if _cutedsl_runtime_available():
             Register tensor of shape (4,) containing the loaded float32 values
         """
         chunk_vec = 4
-        vals_4B_chunk = cute.make_rmem_tensor((chunk_vec,), cutlass.Float32)
+        vals_chunk = cute.make_rmem_tensor((chunk_vec,), cutlass.Float32)
         for j in range(chunk_vec):
-            vals_4B_chunk[j] = vals_block[local_base + j]
-        return vals_4B_chunk
+            vals_chunk[j] = vals_block[local_base + j]
+        return vals_chunk
 
     @cute.jit
-    def load_vals_4B_chunk_tail(
+    def load_vals_chunk_tail(
         vals_block: cute.Tensor,
         dim0: cutlass.Int64,
         sout_base: cutlass.Int32,
@@ -217,11 +217,11 @@ if _cutedsl_runtime_available():
             with out-of-bounds positions set to 0.0
         """
         chunk_vec = 4
-        vals_4B_chunk = cute.make_rmem_tensor((chunk_vec,), cutlass.Float32)
+        vals_chunk = cute.make_rmem_tensor((chunk_vec,), cutlass.Float32)
         for j in range(chunk_vec):
             idx = dim0 + sout_base + j
             if idx < dim_size:
-                vals_4B_chunk[j] = vals_block[local_base + j]
+                vals_chunk[j] = vals_block[local_base + j]
             else:
-                vals_4B_chunk[j] = cutlass.Float32(0.0)
-        return vals_4B_chunk
+                vals_chunk[j] = cutlass.Float32(0.0)
+        return vals_chunk
--- a/torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
+++ b/torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
@@ -15,8 +15,8 @@ from .cute_utils import (
     F8_MAX,
     compute_amax,
     compute_scale_from_amax,
-    load_vals_4B_chunk_full,
-    load_vals_4B_chunk_tail,
+    load_vals_chunk_full,
+    load_vals_chunk_tail,
 )
 
 
@@ -188,7 +188,7 @@ def _compile_mxfp8_quantize_3d_cutedsl(
                 scales_expert[e, n_block, k] = scale_u8
 
         @cute.jit
-        def _store_q_fp8_4B_chunk(
+        def _store_q_fp8_chunk(
             self,
             q_fp8_vals4: cute.Tensor,
             sOUT_tile: cute.Tensor,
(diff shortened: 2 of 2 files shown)


benchmarks/mx_formats/cast_bench.py
```diff
@@ -125,11 +126,31 @@ def run(
...
         "dim1_mxfp8_cuda_floor",
         "dim1_mxfp8_cuda_rceil",
+        "dim0_mxfp8_cutedsl_2d_floor",
+        "dim0_mxfp8_cutedsl_2d_rceil",
     )
 
     x = torch.randn(M, K, dtype=torch.bfloat16, device="cuda") * 1000
 
-    if mode == "dim0":
+    if mode == "memcpy":
```
danielvegamyhre, line 135: note: added this so we can get a sense of best achievable bandwidth utilization for small shapes like (4096, 7168) which memcpy gets 4.5 tb/s. this is useful for cases like seq_len=4096, microbatch_size=1

drisspg:
Can we reuse some of the code between the two?

danielvegamyhre:
> Can we reuse some of the code between the two?

Yeah, will do. Will ping when it's done

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_2d.py
```diff
@@ -0,0 +1,866 @@
...
+        (tile_m, tile_k),
+        stride=(tile_k, 1),
+    )
+    return smem_layout_in, smem_layout_out
+
+
+# Config format:
+# (compute_warps, tile_m, tile_k, k_tiles_per_cta)
+_CUTEDSL_CONFIGS = {
+    "bf16_default": (4, 128, 32, 4),
```
danielvegamyhre, line 50: note: got better perf with slightly different compute_warps and tiles_per_cta for 2d vs 3d

danielvegamyhre:
@drisspg ok i extracted out various shared helpers into cute_utils.py and import them in both 2d/3d quant kernels

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_2d.py
```diff
@@ -0,0 +1,866 @@
...
+            smem_layout_in, smem_layout_out = _make_tile_smem_layouts(
+                cute, TILE_M, TILE_K
+            )
+            # SM >= 100 (Blackwell and beyond, including consumer SM12x and
+            # SM13x): use tcgen05.CtaGroup.ONE for the optimised single-CTA
+            # Blackwell TMA load path.
+            if cutlass.const_expr(IS_BLACKWELL_VALUE):
+                g2s_op = cpasync.CopyBulkTensorTileG2SOp(tcgen05.CtaGroup.ONE)
+            else:
+                g2s_op = cpasync.CopyBulkTensorTileG2SOp()
```
alexsamardzic: At the end, I enabled the 3D kernel for Blackwell only (for easier maintenance, and as GEMM kernels, that these kernels are preparing data for, are Blackwell only), so we can remove `IS_BLACKWELL_VALUE` and keep only branches where this value is true, for both kernels.

alexsamardzic:
> i extracted out various shared helpers into cute_utils.py and import them in both 2d/3d quant kernels

I'm wondering, is it possible to extract more, for example `_issue_tma_store` maybe could have a constexpr arg `IS_INPUT_2D` and then the part of the body that differs between the two coded like:

```Python
if cutlass.const_expr(IS_INPUT_2D):
    sOUT_for_tma_partition = cute.group_modes(sOUT_tile, 0, 1)
    gOUT_for_tma_partition = cute.group_modes(gOUT_tile, 0, 1)
else:
    sOUT_for_tma_partition = cute.group_modes(sOUT_tile, 0, 2)
    gOUT_for_tma_partition = cute.group_modes(gOUT_tile, 0, 2)
```

and alike.  Maybe the kernel itself should be split into more `@cute.jit` functions, so that there is less duplication between the two.  (I know it's boring to do this, but it may make implementing FP4 kernels simpler too.)

danielvegamyhre:
> > i extracted out various shared helpers into cute_utils.py and import them in both 2d/3d quant kernels
> 
> I'm wondering, is it possible to extract more, for example `_issue_tma_store` maybe could have a constexpr arg `IS_INPUT_2D` and then the part of the body that differs between the two coded like:
> 
> ```python
> if cutlass.const_expr(IS_INPUT_2D):
>     sOUT_for_tma_partition = cute.group_modes(sOUT_tile, 0, 1)
>     gOUT_for_tma_partition = cute.group_modes(gOUT_tile, 0, 1)
> else:
>     sOUT_for_tma_partition = cute.group_modes(sOUT_tile, 0, 2)
>     gOUT_for_tma_partition = cute.group_modes(gOUT_tile, 0, 2)
> ```
> 
> and alike. Maybe the kernel itself should be split into more `@cute.jit` functions, so that there is less duplication between the two. (I know it's boring to do this, but it may make implementing FP4 kernels simpler too.)

Yeah I considered extracting the tma load/store but as you noted there are some 2d/3d specific differences and I didn't want to "over-generify / over-abstract" the code... Either way could be reasonable I guess

drisspg:
can you explain the differences between 2d and 3d impls such that we can't have these be the same kernel but with different layouts? 

genuine Q; 

like is one for the packaged inputs the other for the 3d weights, does outer batch dim + different layout not handle this

danielvegamyhre:
> can you explain the differences between 2d and 3d impls such that we can't have these be the same kernel but with different layouts?
> 
> genuine Q;
> 
> like is one for the packaged inputs the other for the 3d weights, does outer batch dim + different layout not handle this

the main reason is that the 2d case has to handle (1) dynamic group sizes and (2) perform per group padding. This isn't done in the kernel yet, I am breaking it into logical steps of "get correct numerics and good perf for simple 2d case" and then once we have that solid starting point, work on adding this, rather than try to tackle it all at once.

drisspg:
@claude review this pr please

danielvegamyhre:
ok @drisspg @alexsamardzic latest changes:
- claude's first comment/"bug" is wrong in this case, ignored (explicit fence isn't necessary after mbarrier completes, it is implicit). i think it was looking at an old version of 3d kernel that had that pattern and got confused, not sure.
- removed IS_BLACKWELL branching and only kept Blackwell path, we don't need to support other architectures, we have emulated mode for that
- unified more helper functions in cute_utils.py and import to use in 2d/3d kernels
- removed unused K param from _select_cutedsl_config. also updated to use actual `torch.dtype` instead of a string  `str(torch.dtype)` - was a bit odd
- renamed functions to make data movement more obvious/clear (register -> smem, reg -> gmem etc)
- added docstrings to functions to clarify inputs/outputs and functionality
- fixed var names in cast_bench.py to clarify we quant along dim0

danielvegamyhre:
@drisspg @alexsamardzic i also confirmed the auto-vectorization is working, since the code as written has loops of 4 byte stores etc but confirmed the SASS shows 16 byte shared load/stores (`LDS.128`/`STS.128`). 

we do have bank conflicts on both smem read/writes that could potentially be addressed in a follow up (e.g. for stores, write to smem with manual swizzle and do tma store with this swizzle set). not sure if worth the lift though, sometimes it ends up being as fast or faster to just let the serialization into sequential access happen...

PR #4157, opened 2026-03-24
[mxfp8 training] refactor mxfp8 grouped mm autograd func

danielvegamyhre:
## Summary
While attempting to integrate the cutedsl 2d mxfp8 quantization kernel from https://github.com/pytorch/ao/pull/4156 into the autograd func, it became apparent the "emulated vs non-emulated" branches everywhere, every helper, etc. were becoming messy/confusing.

This PR refactors the autograd func to have 2 clean code paths: auto vs emulated. The bifurcation happens early, and all code after that point just has one distinct path of kernels/functions, making it easier to reason about and modify.

## Tests
- `pytest test/prototype/moe_training/test_kernels.py -s -k 2d_numerics`
- `pytest test/prototype/moe_training/test_mxfp8_grouped_mm.py -k dq_fwd_bwd`

### DIFF
--- a/torchao/prototype/moe_training/mxfp8_grouped_mm.py
+++ b/torchao/prototype/moe_training/mxfp8_grouped_mm.py
@@ -172,9 +172,12 @@ class _MXFP8GroupedMM(torch.autograd.Function):
             KernelPreference.EMULATED,
         ), "kernel_preference must be AUTO or EMULATED"
 
-        # emulated mode validation
-        emulated = kernel_preference == KernelPreference.EMULATED
-        assert emulated or _SM100_KERNELS_AVAILABLE, (
+        # Determine which path to use: AUTO or EMULATED
+        use_emulated = (
+            kernel_preference == KernelPreference.EMULATED
+            or not _SM100_KERNELS_AVAILABLE
+        )
+        assert not use_emulated or kernel_preference == KernelPreference.EMULATED, (
             "SM100 kernels not available. Please use use torchao CUDA 12.8+ build on SM100/100a device(s). "
             "Otherwise, set kernel_preference=KernelPreference.EMULATED (emulated mode implements basic functionality without efficient kernels)."
         )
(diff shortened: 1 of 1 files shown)


torchao/prototype/moe_training/mxfp8_grouped_mm.py
```diff
@@ -172,9 +172,12 @@ def forward(
             KernelPreference.EMULATED,
         ), "kernel_preference must be AUTO or EMULATED"
 
-        # emulated mode validation
-        emulated = kernel_preference == KernelPreference.EMULATED
-        assert emulated or _SM100_KERNELS_AVAILABLE, (
+        # Determine which path to use: AUTO or EMULATED
+        use_emulated = (
+            kernel_preference == KernelPreference.EMULATED
+            or not _SM100_KERNELS_AVAILABLE
```
vkuzo, line 178: I would prefer deleting this, the code should honor the configuration specified by the user, emulation should be explicitly opt-in.  Turning on emulation automatically is a gotcha IMO.

torchao/prototype/moe_training/mxfp8_grouped_mm.py
```diff
@@ -320,23 +284,19 @@ def backward(ctx, grad_output: torch.Tensor):
...
-        assert emulated or _SM100_KERNELS_AVAILABLE, (
-            "SM100 kernels not available. Please use use torchao CUDA 12.8+ build on SM100/100a device(s)."
-            "Otherwise, set kernel_preference=KernelPreference.EMULATED (emulated mode implements basic functionality without efficient kernels)."
-        )
-
         # Pad grad_output if padding was used in forward (needed for both dgrad and wgrad)
         if pad_token_groups_for_grouped_mm:
             # padded start/end offsets same as what we saved from forward.
             # will be original offsets if we aren't using pad/unpad path
+            kernel_preference = (
```
vkuzo, line 297: this is unexpected, `KernelPreference` is a top level user facing object, now we are here and we have a boolean, and we are converting it back to `KernelPreference`.  Why not just keep the boolean?

danielvegamyhre: yeah this line is weird, the boolean was to avoid repeatedly recomputing the same "if kernel_pref==emulated or not _S100_kernels available" everywhere, and pad_token function was originally designed to be called at the top level where the kernel_pref was available and it has it's own internal logic for handling emualted vs kernel path. if we kept this, i would refactor pad_tokens to just accept the boolean and align with the autograd func refactor in that way.

However, i saw your other comment about not turning on emulated unless explicitly requested, which i agreee makes more sense, so this no longer applies. We don't need the boolean anymore, can just use kernel pref everyhwere directly, and assert at the beginning of forward that if `kernel_pref != emulated` that `_SM100_kernels_available == true`. Will refactor accordingly

PR #4159, opened 2026-03-24
[X86] multi-ISA portable kernel compilation and runtime dispatch

Xia-Weiwen:
## Summary
There is a problem that the X86 kernels in torchao can only run on the machine where they are built. It is because there is no runtime dispatch mechanism for different ISA. That means, if one builds a wheel of Torchao (or a docker image, etc.) with X86 kernels inside, the wheel (docker image) can only be used on machines with the same or higher capability:
Suppose machine A has AVX512 support and machine B only has AVX2 support:
- Build on A + run on A: Yes
- Build on B + run on B: Yes
- Build on A + run on B: No (runtime error: illegal instruction)
- Build on B + run on A: Yes but limited features and low performance

The solution is to add a dynamic dispatch mechanism so that the kernels work on different platforms with a single build.
This PR introduces a light-weight dynamic dispatch mechanism for X86 C++ kernels so that we can build one wheel (or docker image) for all X86 platforms.

## Changes
**Kernel file changes** (kernel files under torchao/csrc/cpu/aten_kernels/)
- Rename kernel files so that their names end with `_krnl.cpp`.
- Replace the anonymous namespace with named namespaces `CPU_CAPABILITY` and it is replaced by ISA names during compilation so that kernels for different ISA live in different namespaces.
- Move op registration to dispatch.cpp (see below).

**Dispatch** (dispatch.cpp)
- Add a new dispatch.cpp file.
- It defines a simple dispatcher to select kernel of the proper ISA at first run.
- It defines wrapper functions and registers them as ops.
- It accepts the ISA level set via a environment variable, which is useful for testing and validation. For example, we can test all ISA paths on high-capability platforms rather than testing each path on different machines. This resembles the `ATEN_CPU_CAPABILITY=avx512` for PyTorch.

**Design Doc** (dynamic_dispatch.md)
- Add a new file to explain the design and how to add a new kernel.

**Build system changes** (setup.py)
 - Add a new `X86KernelBuild` class, which encapsulates all x86-CPU-specific build logic: compiler discovery, compile flags, source filtering, and pre-compilation for different ISA. This keeps CPU-specific code isolated from the rest of the build and from CUDA/other-platform builds.
 - GCC 15 is needed to build all features (especially AVX10.2). GCC 11 is the minimum supported version to build AVX512 code.
 - Pre-compilation of ISA-specific code: at build time, `_krnl.cpp` files are copied to a temp directory and compiled with specific flags for each ISA. The resulting `.o` is linked as an extra_object alongside the main build.
 - No impact on non-CPU builds: all X86-specific flags and sources are gated by `USE_CPU_KERNELS=1`, `is_x86_64` and `is_linux`. The `is_x86_64` is to ensure it's not ARM.

## Test plan
The existing test cases should cover all the changes, especially:
```
pytest -sv test/test_ops.py
pytest -sv test/prototype/test_da8w4_cpu.py
pytest -sv test/prototype/test_float8_opaque_tensor.py
```
One can set the `TORCHAO_CPU_DISPATCH` flag to control which ISA kernel to use. It is useful for testing all paths on a single machine. The `TORCHAO_CPU_DISPATCH` can be set to `DEFAULT`(scalar), `AVX512`, `AMX` and `AVX10_2`.
One can set `TORCHAO_CPU_DISPATCH_DEBUG=1` to see the kernel dispatch info. The info is printed only once.

**Note**: these test cases are not run in CI because x86 kernels are not built in CI. They are tested locally.

Assisted by copilot.


Xia-Weiwen:
@claude review

Xia-Weiwen:
@claude review again.

Xia-Weiwen:
Hi @mingfeima @xuhancn Could you please review this PR? Thanks.

torchao/csrc/cpu/aten_kernels/da8w4_linear_krnl.cpp
```diff
@@ -2,24 +2,32 @@
 #include <ATen/cpu/vec/vec.h>
 #include <ATen/native/CPUBlas.h>
 #include <c10/util/Unroll.h>
+#include <mutex>
+#include "utils.h"
```
mingfeima, line 6:
```suggestion
#include <mutex> // global flag cpublas_once init
#include "utils.h"
```

torchao/csrc/cpu/aten_kernels/float8_linear_krnl.cpp
```diff
@@ -11,28 +11,33 @@
...
-namespace {
+#if defined(CPU_CAPABILITY_AVX10_2)
+  #define ISA_NAMESPACE avx10_2
+#elif defined(CPU_CAPABILITY_AVX512)
+  #define ISA_NAMESPACE avx512
+#else
+  #define ISA_NAMESPACE default_scalar
+#endif
+
+namespace ISA_NAMESPACE {
```
mingfeima, line 22: what's the matter with anonymous namespace, this style looks very strange.

torchao/csrc/cpu/aten_kernels/dispatch.cpp
```diff
@@ -0,0 +1,258 @@
...
+declare_all_kernels(avx10_2)
+declare_all_kernels(avx512)
+declare_all_kernels(default_scalar)
+
+/********** DA8W4 Linear Kernel Dispatch **********/
+declare_da8w4_linear_prepack_impl {
+#if defined(BUILD_AVX10_2) && __GNUC__ >= 15
+  if (kHasAVX10_2) {
+    return avx10_2::call_da8w4_linear_prepack_impl();
+  }
```
mingfeima, line 161: does AO follows the ISA dispatch logic from Aten, why we are using ISA as namespace here?

torchao/csrc/cpu/aten_kernels/float8_linear_krnl.cpp
```diff
@@ -163,11 +160,11 @@ static void cvt_f8e4m3_to_bf16(
...
-      size_t vec_len = cols / 32 * 32;
-      for (; i < vec_len; i += 32) {
-        __m256i fp8_vec = _mm256_loadu_si256((__m256i*)&in[r * stride + i]);
-        __m512bh bf16_vec = cvt_e4m3_bf16_intrinsic_no_nan(fp8_vec);
-        _mm512_storeu_si512((__m512i*)(out + r * cols + i), (__m512i)bf16_vec);
+      size_t vec_len = cols / vec_len * vec_len;
+      for (; i < vec_len; i += vec_len) {
+        __m128i fp8_vec = _mm_loadu_si128((__m128i*)&in[r * stride + i]);
+        __m256bh bf16_vec = cvt_fp8e4m3_to_bf16(fp8_vec);
+        _mm256_storeu_si256((__m256i*)(out + r * cols + i), (__m256i)bf16_vec);
```
mingfeima, line 167: L163 is wrong, will result into illegal memory access.

do you have test cases to cover this branch.

torchao/csrc/cpu/aten_kernels/float8_linear_krnl.cpp
mingfeima, line 157: this branch is not needed. just handle [rows, cols] with stride [stride, 1] will be enough. And check the input tensor to be lasdim contiguous.

mingfeima (changes requested):
test cases need to cover both contiguous and non-contigous inputs.

torchao/csrc/cpu/aten_kernels/float8_linear_krnl.cpp
xuhancn: It is a part of dynamic dispatch: https://github.com/intel/intel-extension-for-pytorch/blob/main/docs/tutorials/features/isa_dynamic_dispatch.md#dynamic-dispatch-design

setup.py
```diff
@@ -382,7 +384,205 @@ def bool_to_on_off(value):
...
+            ["-O3", "-std=c++20", "-fPIC", "-fopenmp"]
+            + include_flags
+            + ["-I", aten_kernels_dir]
+        )
+        avx512_defines = ["-DCPU_CAPABILITY_AVX512", "-DCPU_CAPABILITY_AVX512_VNNI"]
+        avx10_2_defines = avx512_defines + ["-DCPU_CAPABILITY_AVX10_2"]
+        build_configs = [
+            {
+                "isa": "avx512",
+                "gcc_min_ver": X86KernelBuild._MINIMUM_GCC_MAJOR,
```
xuhancn, line 535: Check gcc version is not robust method, please check compiler capability dry run build, please reference: https://github.com/pytorch/pytorch/blob/cc3b1bccdd8c8bfd80e0f2de94f1299824d7ab8f/cmake/Modules/FindAVX.cmake#L48-L75
A python implement is better for AO.

torchao/csrc/cpu/aten_kernels/scaled_embedding_bag_krnl.cpp
```diff
@@ -45,7 +45,15 @@
 
 namespace torchao {
 
-namespace {
+#if defined(CPU_CAPABILITY_AVX10_2)
+  #define ISA_NAMESPACE avx10_2
```
xuhancn, line 49: we can define ISA_NAMESPACE in compiler ISA check stage, please reference: https://github.com/pytorch/pytorch/blob/be5f4f9611e30fc4111508d994974f28772f5d50/cmake/Codegen.cmake#L443-L447

Xia-Weiwen:
@claude review

torchao/csrc/cpu/aten_kernels/da8w4_linear_krnl.cpp
Xia-Weiwen: This is not needed. I have removed it. Thanks.

torchao/csrc/cpu/aten_kernels/float8_linear_krnl.cpp
Xia-Weiwen: I have updated this part to be aligned the torch/IPEX. The namespace is now `CPU_CAPABILITY`, which will be replaced by ISA name during compilation, e.g., AVX512, etc. Thanks.

torchao/csrc/cpu/aten_kernels/dispatch.cpp
Xia-Weiwen: The AO does not have a dynamic dispatch mechanism. This PR is adding one, which is similar to those in Torch and IPEX but more light weight. Thanks.

torchao/csrc/cpu/aten_kernels/float8_linear_krnl.cpp
Xia-Weiwen: This part is updated in another PR. Thanks.

torchao/csrc/cpu/aten_kernels/float8_linear_krnl.cpp
Xia-Weiwen: This part is updated in another PR. Thanks.

setup.py
Xia-Weiwen: Now it uses C++ code snippets to test the capability of the compiler. Thanks.

torchao/csrc/cpu/aten_kernels/scaled_embedding_bag_krnl.cpp
Xia-Weiwen: I have updated this part to be aligned with Torch/IPEX (`namespace CPU_CAPABILITY`). Thanks.

Xia-Weiwen:
Hi @mingfeima @xuhancn Can you please review again? Thanks.

torchao/csrc/cpu/aten_kernels/utils.h
```diff
@@ -98,3 +102,24 @@ template <typename T> constexpr int get_vnni_size() { return vnni_traits<T>::siz
             )                                                                           \
         )                                                                               \
     )
+
+enum DispatchMode {
+  MODE_DEFAULT = 0,
+  MODE_AVX512 = 1, // Build with AVX512. Brgemm disabled manually.
+  MODE_AMX = 2, // Build with AVX512. Brgemm enabled manually.
```
xuhancn, line 109: Why here define `MODE_AMX` but not dispatch this ISA?

setup.py
```diff
@@ -382,7 +384,289 @@ def bool_to_on_off(value):
...
+            + include_flags
+            + ["-I", aten_kernels_dir]
+        )
+        avx512_defines = ["-DCPU_CAPABILITY_AVX512", "-DCPU_CAPABILITY_AVX512_VNNI"]
+        avx512_isa_flags = [
+            "-mavx512f",
+            "-mavx512bw",
+            "-mavx512vl",
+            "-mavx512dq",
+            "-mavx512vnni",
```
xuhancn, line 611: As PyTorch defination, avx512vnni is not belong to avx512core, why we add it to this scope?

torchao/csrc/cpu/aten_kernels/utils.h
Xia-Weiwen: Thanks for the comment. This is for manual dispatch, which is used for testing and validation. When users set the dispatch level to AMX, it means AMX kernel is allowed but it is subject to hardware availability.
We don't have an explicit ISA level called AMX because we don't write AMX code explicitly. We use oneDNN instead.

setup.py
Xia-Weiwen: Thanks for the comment. It's because we don't have implementations with pure AVX512-VNNI and we don't write AMX code explicitly. So, both AVX512-VNNI and AMX are put in the AVX512 ISA level and it's determined by oneDNN runtime check to use AMX/VNNI or not.

mingfeima (approved):

Xia-Weiwen:
Hi @jerryzh168 @vkuzo We would like to hear your comments on this PR. With this patch, we can build one wheel for all x86 CPU platforms (Linux only). This potentially enables us to release wheels for x86 in the future. Of course, that is not included in this PR and more discussions are needed. Thanks.

torchao/csrc/cpu/aten_kernels/dispatch.cpp
```diff
@@ -0,0 +1,237 @@
...
+    }
+  });
+
+  static const char* env_dispatch_debug = std::getenv(TORCHAO_CPU_DISPATCH_DEBUG_ENV);
+  static bool dispatch_debug = env_dispatch_debug ? std::string(env_dispatch_debug) == "1" : false;
+
+  static KernelDispatcher dispatcher = []() {
+    KernelDispatcher d;
+    // Select ISA level based on runtime detection (kHas*) and compile-time checks (BUILD_*)
+#if defined(BUILD_AVX10_2)
```
jerryzh168, line 173: do you really need to implement a custom dispatcher? or just adding macros for the implementation is enough:

```
... op(...) {
  #if defined(BUILD_AVX10_2)
   if (...) {
     impl 1
     return ...;
  }
  #endif

  #if defined(BUILD_AVX512) 
   if (...) {
     impl 2
     return ...;
  }
  #endif

  #if ..
   if (...) {
     impl 3
     return ...;
  }
   #endif
}
```

Xia-Weiwen: Thanks for the comment. The dispatcher provides a unified management of the dispatching logic. Without the dispatcher, we need similar code in the kernel wrapper functions below. Besides, the dispatcher's implementation is simple and straightforward. So, the dispatcher would make it easier for maintenance.

jerryzh168: cc @ezyang in case there are macro/hardware based dispatcher utilities that we can reuse in pytorch, but I think in the end it's user's (maintainers') choice

Xia-Weiwen: Thanks. If there is existing utilities that can be reused here, we can switch to that.

ezyang: Uhhhhhh. Well, PyTorch *core* has a dynamic dispatch mechanism for ISA, but I don't know how well it will work out of tree, and certainly not with stable ABI. @janeyx99 do we have anything like that? I think it would probably be valuable for stable ABI.

Arranging to make sure the recompilations of the source file with different ISA don't clobber each other with shared symbols is surprisingly complicated; if you were using LLMs to implement this I'd definitely have them look at the PyTorch ref implementation.

Xia-Weiwen: > Uhhhhhh. Well, PyTorch _core_ has a dynamic dispatch mechanism for ISA, but I don't know how well it will work out of tree, and certainly not with stable ABI. @janeyx99 do we have anything like that? I think it would probably be valuable for stable ABI.
> 
> Arranging to make sure the recompilations of the source file with different ISA don't clobber each other with shared symbols is surprisingly complicated; if you were using LLMs to implement this I'd definitely have them look at the PyTorch ref implementation.

@ezyang Thanks for your comment. We are indeed taking the PyTorch impl as the ref and recompilations of source files are exactly what the PyTorch's dynamic dispatch does. The implementation in this PR is just a simplified version of that in PyTorch core. And just my two cents: recompilations reduce duplicate code in source files and makes it easier to maintain. Otherwise, we need to write a kernel and register an op for each ISA, which leads to duplicate code. Not sure if there is an even simpler way.

Xia-Weiwen:
Hi @vkuzo @janeyx99 Could you please review this PR? Thanks.

torchao/csrc/cpu/aten_kernels/dispatch.cpp
janeyx99: Ah the APIs in PyTorch are indeed not ABI stable today, and we probably will do some level of support for them if we need to focus more on CPU perf. If the code in this PR makes torchao not ABI stable anymore then that would be a blocker, though I haven't checked.

Xia-Weiwen: @janeyx99 Thanks for your comment. I am still a little confused. What should I do? Do you have any suggestions? Thanks.

Xia-Weiwen:
Hi @vkuzo @janeyx99 Could you please review this PR? Thanks.

Xia-Weiwen:
Hi @vkuzo @janeyx99 @ezyang Do you have any concern about this PR? Thanks.

janeyx99:
@Xia-Weiwen I'm not the best person to review this code, I believe @vkuzo would have more context

Xia-Weiwen:
> @Xia-Weiwen I'm not the best person to review this code, I believe @vkuzo would have more context

Got it. Thanks.

vkuzo:
@claude does this PR regress ABI compatibility/stability in any way? any other concerns?

Xia-Weiwen:
Land as CI shows green.

PR #4160, opened 2026-03-24
Add zero_point and act_mapping_type checks in int8 quantization

cyxlily:
For static SmoothQuant with mapping_type == MappingType.SYMMETRIC, activation_tensor.zero_point is generated by observed_linear.obs.calculate_qparams(weight_quant_kwargs=quant_kwargs). It is not None, but a tensor of zeros. We only need to apply the compensation when act_quant_kwargs.mapping_type == MappingType.ASYMMETRIC.

### DIFF
--- a/torchao/quantization/quant_api.py
+++ b/torchao/quantization/quant_api.py
@@ -1105,9 +1105,13 @@ class Int8DynamicActivationInt8WeightConfig(AOBaseConfig):
                 self.granularity
             )
             _validate_granularity_int8(act_granularity, weight_granularity)
-            assert self.act_mapping_type is not None, (
+            assert self.act_mapping_type in (
+                MappingType.SYMMETRIC,
+                MappingType.ASYMMETRIC,
+            ), (
                 "Int8DynamicActivationInt8WeightConfig requires "
-                "`act_mapping_type` to be set (not be None). "
+                "`act_mapping_type` in (MappingType.SYMMETRIC, "
+                "MappingType.ASYMMETRIC). "
                 "Please set it to MappingType.SYMMETRIC or "
                 "MappingType.ASYMMETRIC."
             )
@@ -1245,9 +1249,13 @@ class Int8StaticActivationInt8WeightConfig(AOBaseConfig):
             self.granularity
         )
         _validate_granularity_int8(act_granularity, weight_granularity)
-        assert self.act_mapping_type is not None, (
+        assert self.act_mapping_type in (
+            MappingType.SYMMETRIC,
+            MappingType.ASYMMETRIC,
+        ), (
             "Int8StaticActivationInt8WeightConfig requires "
-            "`act_mapping_type` to be set (not be None). "
+            "`act_mapping_type` in (MappingType.SYMMETRIC, "
+            "MappingType.ASYMMETRIC). "
             "Please set it to MappingType.SYMMETRIC or "
             "MappingType.ASYMMETRIC."
         )


torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -301,7 +301,10 @@ def _(func, types, args, kwargs):
         # Asymmetric activation zero_point correction:
         # Y = (X_int @ W_int^T) * s_x * s_w - zp_x * s_x * row_sum(W_int)^T * s_w
         # The first term is y_dot_scaled * w_scales. The second is the correction.
-        if activation_tensor.zero_point is not None:
+        if (
+            weight_tensor.act_quant_kwargs.mapping_type == MappingType.ASYMMETRIC
```
Xia-Weiwen, line 305: check `weight_tensor.act_quant_kwargs is not none`?

cyxlily: Already check earlier:  https://github.com/pytorch/ao/blob/main/torchao/quantization/quantize_/workflows/int8/int8_tensor.py#L268

torchao/quantization/quant_api.py
```diff
@@ -1226,6 +1244,7 @@ class Int8StaticActivationInt8WeightConfig(AOBaseConfig):
             activations and weights) or a tuple / list of two granularities (first for activations, second for weights).
             If None, defaults to PerRow for both. Only PerTensor and PerRow are supported.
         act_mapping_type (MappingType): The mapping type for activation quantization. SYMMETRIC and ASYMMETRIC are supported.
```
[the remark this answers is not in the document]
Xia-Weiwen: This patch makes sense to me. @cyxlily Please take a look.

torchao/quantization/quant_api.py
```diff
@@ -1068,6 +1068,21 @@ def _validate_granularity_int8(
...
+    if act_mapping_type is None:
+        return MappingType.SYMMETRIC
+
+    supported = (MappingType.SYMMETRIC, MappingType.ASYMMETRIC)
+    if act_mapping_type not in supported:
+        raise ValueError(
+            f"Unsupported act_mapping_type: {act_mapping_type}. "
+            f"Only {supported} are supported for int8 activation quantization."
+        )
+    return act_mapping_type
```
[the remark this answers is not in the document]
Xia-Weiwen: @cyxlily Can we ensure `act_mapping_type=None` is tested in UT?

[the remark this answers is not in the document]
cyxlily: Thanks for the comment, add unit test test_act_mapping_type_none_defaults_to_symmetric.

torchao/quantization/quant_api.py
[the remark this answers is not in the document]
cyxlily: Thanks for the comment, updated.

Xia-Weiwen:
Hi @jerryzh168 Could you please take a look. The CI failures are unrelated

torchao/quantization/quant_api.py
```diff
@@ -1058,6 +1058,21 @@ def _validate_granularity_int8(
         )
 
 
+def _validate_act_mapping_type_int8(
```
jerryzh168: this is not validate? validate should just assert when something fails, not return something

jerryzh168: this is mixing up normalize and validate into the same function I believe

torchao/quantization/quant_api.py
```diff
@@ -1217,7 +1235,7 @@ class Int8StaticActivationInt8WeightConfig(AOBaseConfig):
             The granularity for quantization. Can be either a single granularity (applied to both
             activations and weights) or a tuple / list of two granularities (first for activations, second for weights).
             If None, defaults to PerRow for both. Only PerTensor and PerRow are supported.
-        act_mapping_type (MappingType): The mapping type for activation quantization. SYMMETRIC and ASYMMETRIC are supported.
+        act_mapping_type (Optional[MappingType]): The mapping type for activation quantization. SYMMETRIC and ASYMMETRIC are supported.
```
jerryzh168: why do we need the Optional? why not just default to SYMMETRIC

torchao/quantization/quant_api.py
cyxlily: Thanks for the comment, remove this validate function and add assert.

torchao/quantization/quant_api.py
cyxlily: Thanks for the comment, remove the Optional.

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -301,7 +301,10 @@ def _(func, types, args, kwargs):
         # Asymmetric activation zero_point correction:
         # Y = (X_int @ W_int^T) * s_x * s_w - zp_x * s_x * row_sum(W_int)^T * s_w
         # The first term is y_dot_scaled * w_scales. The second is the correction.
-        if activation_tensor.zero_point is not None:
+        if (
```
jerryzh168, line 304: this change of `if` also changed what goes to else, is that intended?

should we just add an assert of `weight_tensor.act_quant_kwargs.mapping_type == MappingType.ASYMMETRIC` in the original if branch?

jerryzh168: ideally we should avoid these multiple condition statements, it makes the code hard to understand and more error prone

jerryzh168: please add the motivation to PR summary, I don't quite understand why this change

jerryzh168:
took another look, `weight_tensor.act_quant_kwargs` is probably the arg to check to see this is weight only quant or not

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
cyxlily: Thanks, for the comment, I updated the PR summary. For static SmoothQuant with SYMMETRIC mapping, the asymmetric-compensation branch is expected to be skipped. Execution then proceeds with the regular scaled matmul output path, not the weight-only path. The weight-only path is only reached via the outer else of if weight_tensor.act_quant_kwargs is not None:.

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -301,7 +301,10 @@ def _(func, types, args, kwargs):
...
+            weight_tensor.act_quant_kwargs.mapping_type == MappingType.ASYMMETRIC
+            and activation_tensor.zero_point is not None
+        ):
             w_row_sums = weight_tensor.qdata.sum(dim=-1)  # (N,)
```
jerryzh168, line 308: I see, can you add a comment for the symmetric case

PR #4165, opened 2026-03-24
Add flag for users to silence version compatibility warning

andrewor14:
**Summary:** Many users wish to just use the python only APIs but still see this warning when they import torchao, e.g.

https://github.com/pytorch/ao/issues/2919#issuecomment-3821689925 https://github.com/pytorch/ao/issues/2919#issuecomment-3849920557

For example, this warning still triggers when they use torchao 0.17.0 with torch 2.10, which is within the 3 most recent torch versions that we support for python only APIs. This commit adds an optional env var users can explicitly set to disable this warning.

**Test Plan:** Manual testing.

### DIFF
--- a/torchao/__init__.py
+++ b/torchao/__init__.py
@@ -76,11 +76,13 @@ elif not ("+git" in __version__) and not ("unknown" in __version__):
         skip_loading_so_files = True
         _skip_reason = (
             f"Skipping import of cpp extensions due to incompatible torch version. "
-            f"Please upgrade to torch >= 2.11.0 (found {torch.__version__})."
+            f"Please upgrade to torch >= 2.11.0 (found {torch.__version__}). "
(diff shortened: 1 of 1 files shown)


torchao/__init__.py
```diff
@@ -76,11 +76,13 @@ def _parse_version(version_string):
         skip_loading_so_files = True
         _skip_reason = (
             f"Skipping import of cpp extensions due to incompatible torch version. "
-            f"Please upgrade to torch >= 2.11.0 (found {torch.__version__})."
+            f"Please upgrade to torch >= 2.11.0 (found {torch.__version__}). "
+            f"Set TORCHAO_SILENCE_VERSION_COMPATIBILITY_WARNING=1 to silence this warning."
```
jerryzh168, line 80: can this be set by flags/ways related to the `logger.warning` call itself (e.g. setting logger level?) instead we add a new torchao-only flag?

jerryzh168:
nit: for label, maybe we can use `module: core` for now?

PR #4166, opened 2026-03-24
remove per kernel fp8 triton quantized benchmarks

iamzainhuda:
### Summary 
This removes the redundant bench_triton_fp8_* quant benchmark scripts under benchmarks/prototype/blockwise_fp8_training and consolidates their correctness coverage into benchmark_quant_kernel_bandwidth.py.

The consolidated benchmark now has an optional --check-correctness mode that compares each Triton quant kernel against the corresponding Torch reference implementation, including layout/stride checks where applicable. The weight-kernel references were adjusted to match Triton’s column-major outputs and scales, which was required for weight_quant_rhs and weight_quant_transposed_rhs.
  
### Testing
  - python -m py_compile benchmarks/prototype/blockwise_fp8_training/benchmark_quant_kernel_bandwidth.py
  - python -m benchmarks.prototype.blockwise_fp8_training.benchmark_quant_kernel_bandwidth --help
  - python -m benchmarks.prototype.blockwise_fp8_training.benchmark_quant_kernel_bandwidth --check-correctness on H100, completed successfully with no correctness failures.

### DIFF
--- a/benchmarks/prototype/blockwise_fp8_training/benchmark_quant_kernel_bandwidth.py
+++ b/benchmarks/prototype/blockwise_fp8_training/benchmark_quant_kernel_bandwidth.py
@@ -420,7 +420,9 @@ def _print_results(
                     if achievable_utils
                     else None
                 ),
-                _format_optional_float(min(achievable_utils) if achievable_utils else None),
+                _format_optional_float(
+                    min(achievable_utils) if achievable_utils else None
+                ),
             ]
         )
     overall_rows.sort(


benchmarks/prototype/blockwise_fp8_training/benchmark_quant_kernel_bandwidth.py
```diff
@@ -439,22 +526,45 @@ def parse_args():
...
+        action="store_true",
+        help=(
+            "Run a Torch reference implementation once per valid shape/kernel "
+            "pair and assert the Triton output matches."
+        ),
+    )
+    parser.add_argument(
+        "--correctness-rtol",
+        type=float,
+        default=1e-2,
```
danielvegamyhre: what atol/rtol are needed for this to pass? did you reference the unit tests?

benchmarks/prototype/blockwise_fp8_training/benchmark_quant_kernel_bandwidth.py
```diff
@@ -439,22 +526,45 @@ def parse_args():
...
 
     torch.random.manual_seed(67)
-    args = parse_args()
     bandwidth_spec = _resolve_gpu_specs(use_roofline_utils=args.use_roofline_utils)
 
     measurements, skipped = _run_suite(
         m_values=args.m_values,
         k=args.k,
         block_size=args.block_size,
         bandwidth_spec=bandwidth_spec,
```
danielvegamyhre, line 576: can we also update this script to only measure against peak achievable bandwidth as defined in the gpu_spec in roofline_utils.py we import (not measure against the "advertised max theoretical bandwidth" number)

benchmarks/prototype/blockwise_fp8_training/benchmark_quant_kernel_bandwidth.py
iamzainhuda: the per kernel scripts used 1e-2 so i kept that behavior. correctness passes with tolerances set to 0. for the same kernels in our testing we have tolerances set to 0, so i can set that as default here too: https://github.com/pytorch/ao/blob/main/test/prototype/blockwise_fp8_training/test_blockwise_kernels.py#L127

danielvegamyhre: sounds good, yeah let's use 0 if we can

benchmarks/prototype/blockwise_fp8_training/benchmark_quant_kernel_bandwidth.py
iamzainhuda: right now script reports both, do you want me to remove the max theoretical bandwidth numbers altogether? 
image

danielvegamyhre: let's just keep vs_achievable

iamzainhuda: done, added print to make it clear to user as well what achievable % is used

PR #4167, opened 2026-03-25
Fix rocm CI

jerryzh168:
* __->__ #4167

Summary:
The fix skips the pytorch version check when kernel_preference is EMULATED, since the emulated mode doesn't need scaled_mm_single_dim_strategy or is_pinned_handler from torch.distributed.tensor. The hardware
MXFP8 paths (AUTO kernel preference) still get the version check.

Test Plan:
CI

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/prototype/moe_training/test_training.py
+++ b/test/prototype/moe_training/test_training.py
@@ -1,3 +1,9 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+
 import copy
 
 import pytest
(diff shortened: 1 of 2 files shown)


danielvegamyhre:
@jerryzh168 instead of carving out this case for emulated mode can we just skip rocm for the whole test_training.py file for ROCM, and keep this version check as is? nobody is using mxfp8 + rocm (not even possible right now, we have several cuda-only kernels with no rocm alternative, just emulated mode.

i think allowing ROCM+MXFP8 tests in CI was a mistake, this was my bad, i shouldn't have approved the PR.

jerryzh168:
@danielvegamyhre sounds good, I'll change the PR to skip the training test for rocm

PR #4173, opened 2026-03-25
Add WAN style rope detection pattern for low precision attention API

howardzhang-cv:
* __->__ #4173

## Quick Summary
- Added rope fusion detection for WAN style models that use RoPE. This is for the low precision attention API, to enable rope fusion.
- Previously, the fallback path would work, allowing for torch.compile to work with the monkey patched F.SDPA. However, the RoPE operation was not fused into a single kernel with the quantization to fp8.

## Results
Previous results on WAN model:
| Config | Median Time (s) | Speedup |
| --- | --- | --- |
| bf16 baseline | 100.85 | 1.00x |
| fp8_attn | 67.15 | **1.50x** |
| bf16 baseline + torch.compile | 81.32 | 1.00x |
| fp8_attn + compile | 46.88 | **1.73x** |

New results with rope fusion:
| Config | Median Time (s) | Speedup |
| --- | --- | --- |
| fp8_attn + rope fusion + compile | 44.24 | **1.84x** |

### DIFF
--- a/torchao/prototype/attention/shared_utils/fusion_utils.py
+++ b/torchao/prototype/attention/shared_utils/fusion_utils.py
@@ -692,6 +692,23 @@ def _detect_neox_rope(node: Node) -> Optional[RoPEMatch]:
     return _try_match(right, left)
 
 
+def _parse_stride2_index(idx) -> Optional[int]:
+    """Parse ``(..., slice(start, None, 2))`` index tuples.
+
+    Returns the slice ``start`` value (defaulting to 0), or None if the
+    index does not match the expected stride-2 pattern.
+    """
+    if not isinstance(idx, tuple) or len(idx) < 1:
+        return None
+    last = idx[-1]
+    if not isinstance(last, slice) or last.step != 2:
+        return None
+    for entry in idx[:-1]:
+        if entry is not Ellipsis and entry != slice(None):
+            return None
+    return last.start if last.start is not None else 0
+
+
 def _get_setitem_stride2_slice(node: Node) -> Optional[Tuple[int, Node]]:
     """Check if node is ``operator.setitem(tensor, (..., slice(start, None, 2)), value)``.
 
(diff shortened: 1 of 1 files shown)


torchao/prototype/attention/shared_utils/fusion_utils.py
```diff
@@ -690,9 +692,204 @@ def _try_match(x_mul: Node, rot_mul: Node) -> Optional[RoPEMatch]:
...
+    if not isinstance(value, Node):
+        return None
+    # idx should be (Ellipsis, slice(start, None, 2))
+    if not isinstance(idx, tuple) or len(idx) < 1:
+        return None
+    last = idx[-1]
+    if not isinstance(last, slice) or last.step != 2:
+        return None
+    # Check preceding entries are Ellipsis or slice(None)
+    for entry in idx[:-1]:
```
drisspg: these should  share ode right?

howardzhang-cv: I debated that but kept it separate just for cleaner separation of operations when we're trying to pattern match rope. Like for example, when we pattern match, it'll say ("find setitem operations with stride 2 slices, do stuff etc., find getitem operations with stride 2 slices"). I thought it was slightly cleaner to keep them separate.

howardzhang-cv: Actually, I just thought about it a bit more and I think it's better to combine them. Probably better to avoid the case where someone changes one but forgets to change the other.

drisspg (approved):
I think that if we are goin to add more patterns we should have some tests for the patterns we want to match against besides just integration tests

howardzhang-cv:
> I think that if we are goin to add more patterns we should have some tests for the patterns we want to match against besides just integration tests

Good point. I'll work on setting those up. I'll do it in a separate PR to keep it clean.

PR #4178, opened 2026-03-25
[mxfp8 moe training] register sharding rules for cutedsl 2d quant kernel

danielvegamyhre:
[mxfp8 moe training] register sharding rules for cutedsl 2d quant kernel

# Summary
- Register sharding rules for new cutedsl 2d tensor quantization kernel. 

## Tests
- Need kernel to support fused per group padding then integrate it into the autograd func, then we can add a test. This is pretty straightforward though, and follows the same pattern as our other kernels.

### DIFF
--- a/torchao/prototype/moe_training/kernels/mxfp8/quant.py
+++ b/torchao/prototype/moe_training/kernels/mxfp8/quant.py
@@ -8,6 +8,8 @@ from typing import Tuple
 
 import torch
 from torch import Tensor
+from torch.distributed.tensor import Replicate, Shard
+from torch.distributed.tensor.experimental import register_sharding
 from torch.utils._triton import has_triton
 
 from torchao.prototype.moe_training.kernels.mxfp8.cute_utils import (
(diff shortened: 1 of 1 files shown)


torchao/prototype/moe_training/kernels/mxfp8/quant.py
```diff
@@ -975,6 +977,18 @@ def _fake_mxfp8_quantize_2d_cutedsl_custom_op(
     return q_data, scales
 
 
+@register_sharding(torch.ops.torchao._mxfp8_quantize_2d_cutedsl_custom_op.default)
+def custom_triton_to_mxfp8_dim0_sharding(
```
vkuzo: `custom_triton` is unexpected here, from the PR title this is for a cutedsl kernel?

accepting to unblock

danielvegamyhre: woops, thank you, renamed accordingly

PR #4180, opened 2026-03-26
Passing alg_id by semi sparse config (#4180)

RandySheriff:
Summary:
X-link: https://github.com/pytorch/pytorch/pull/178458

As titled ~

Reviewed By: ZihaoLiu0927

Differential Revision: D98240336

### DIFF
--- a/torchao/sparsity/sparse_api.py
+++ b/torchao/sparsity/sparse_api.py
@@ -8,7 +8,7 @@ from dataclasses import dataclass
 from typing import Callable, Optional
 
 import torch
-from torch.sparse import to_sparse_semi_structured
+from torch.sparse import SparseSemiStructuredTensor, to_sparse_semi_structured
 
 from torchao.core.config import AOBaseConfig
 from torchao.prototype.sparsity.sparsifier.weight_norm_sparsifier import (
@@ -75,6 +75,10 @@ class SemiSparseWeightConfig(AOBaseConfig):
     Configuration for converting the weight of linear modules to semi-structured (2:4) sparsity
     """
 
+    def __init__(self, alg_id: int = SparseSemiStructuredTensor._DEFAULT_ALG_ID):
+        super().__init__()
+        self.alg_id = alg_id
+
     def __post_init__(self):
         torch._C._log_api_usage_once("torchao.sparsity.SemiSparseWeightConfig")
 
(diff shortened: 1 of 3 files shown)


torchao/sparsity/sparse_api.py
```diff
@@ -75,6 +73,9 @@ class SemiSparseWeightConfig(AOBaseConfig):
     Configuration for converting the weight of linear modules to semi-structured (2:4) sparsity
     """
 
+    def __init__(self, alg_id: int = SparseSemiStructuredTensor._DEFAULT_ALG_ID):
+        self.alg_id = alg_id
```
jerryzh168, line 80: isn't this supposed to be figured out by calling some search function?

RandySheriff: exactly ~ we are going to invoke mm_search in production logic and pass it in here

jerryzh168:
for my understanding, is the flow something like the following:

```
1. initialization:
linear.weight = to_sparse_weight(linear.weight, alg_id=default)

2. runtime:
def linear_dispatch(args, kwargs, ...):
    activation = args[0]
    weight = args[1]
    optimal_alg_id = _some_function(activation, weight)
    weight.alg_id = optional_alg_id
    return sparse_linear(activation, weight.data, weight.alg_id)
```

RandySheriff:
> for my understanding, is the flow something like the following:
> 
> ```
> 1. initialization:
> linear.weight = to_sparse_weight(linear.weight, alg_id=default)
> 
> 2. runtime:
> def linear_dispatch(args, kwargs, ...):
>     activation = args[0]
>     weight = args[1]
>     optimal_alg_id = _some_function(activation, weight)
>     weight.alg_id = optional_alg_id
>     return sparse_linear(activation, weight.data, weight.alg_id)
> ```

yes, and it will be like:

during model lowering:

1.  mock_tensor = ...
2. alg_id = mm_search(mock_tensor)
3. linear.weight = to_sparse_weight(linear.weight, alg_id=alg_id)

linear_dispatch will occur during serving
The purpose of the diff is to let #2 to get alg_id and pass it to #3.
Without the diff, there is no way to assign alg_id to the semi tensor.

jerryzh168:
@RandySheriff thanks for the explanations, wondering if we want to add some test to use `alg_id`? looks like right now it is not used

RandySheriff:
> @RandySheriff thanks for the explanations, wondering if we want to add some test to use `alg_id`? looks like right now it is not used

sure ~ suggest a file where some unit test can be add?

jerryzh168:
please fix ruff

test/sparsity/test_sparse_api.py
```diff
@@ -39,7 +39,7 @@ def test_sparse(self):
         apply_fake_sparsity(model)
         dense_result = model(input)
 
-        sparsify_(model, semi_sparse_weight())
+        sparsify_(model, semi_sparse_weight(alg_id=0))
```
jerryzh168: looks good

Issue #4138, opened 2026-03-23
Buffer mutation tracking lost after convert_pt2e for index_put_


tom-arm:
## Summary

When using `prepare_pt2e` / `convert_pt2e` to quantize a module that mutates a registered buffer via `index_put_`, the resulting exported program loses its `buffers_to_mutate` mapping. The FP32 exported program correctly records the buffer mutation, but after quantization the buffer is renamed to a frozen parameter (`_frozen_param0`) and the mutation relationship is silently dropped.

## Reproduction

Below is a reproduction script with a simple quantizer for `index_put_`. Note this annotation behviour is similar to the `TOSAQuantizer` in ExecuTorch where this issue was found. 

```python
import torch
from torch.export import export
from torchao.quantization.pt2e.observer import MinMaxObserver
from torchao.quantization.pt2e.quantize_pt2e import convert_pt2e, prepare_pt2e
from torchao.quantization.pt2e.quantizer import (
    QuantizationAnnotation,
    QuantizationSpec,
    Quantizer,
    SharedQuantizationSpec,
)
from torchao.quantization.pt2e.quantizer.quantizer import Q_ANNOTATION_KEY


class MutatingBufferModule(torch.nn.Module):
    def __init__(self) -> None:
        super().__init__()
        self.register_buffer("buf", torch.zeros(4, dtype=torch.float32))

    def forward(self, x: torch.Tensor, idx: torch.Tensor) -> torch.Tensor:
        updated = self.buf.index_put_((idx,), x)
        return updated.clone()


class IndexPutQuantizer(Quantizer):
    def __init__(self) -> None:
        super().__init__()
        self.qspec = QuantizationSpec(
            dtype=torch.int8,
            observer_or_fake_quant_ctr=MinMaxObserver.with_args(eps=2**-12),
            quant_min=-128,
            quant_max=127,
            qscheme=torch.per_tensor_affine,
        )

    def annotate(self, model: torch.fx.GraphModule) -> torch.fx.GraphModule:
        for node in model.graph.nodes:
            if node.op != "call_function":
                continue
            if node.target != torch.ops.aten.index_put_.default:
                continue

            dst = node.args[0]
            value = node.args[2]
            node.meta[Q_ANNOTATION_KEY] = QuantizationAnnotation(
                input_qspec_map={
                    dst: self.qspec,
                    value: SharedQuantizationSpec((dst, node)),
                },
                output_qspec=SharedQuantizationSpec((dst, node)),
                _annotated=True,
[...]

PR #4181, opened 2026-03-26
Prevent folding of mutable input to copy_ and put_ operators

tom-arm:
* Find mutable buffers for put_ operations as well as copy_.
* Trace the graph to the buffer and mark nodes found as mutable to prevent folding.

### DIFF
--- a/torchao/quantization/pt2e/constant_fold.py
+++ b/torchao/quantization/pt2e/constant_fold.py
@@ -94,19 +94,66 @@ class ConstantFolder(torch.fx.Interpreter):
         # Identify mutable buffers by finding copy_ operations
         self.mutable_buffers = self._find_mutable_buffers()
 
+    def _is_mutable_buffer(self, node: torch.fx.Node) -> bool:
+        """Check if a node is a mutable buffer."""
+        named_buffers = dict(self.module.named_buffers())
+        if node.op == "placeholder":
+            return True
+
+        if node.op == "get_attr" and str(node.target) in named_buffers:
+            return True
+
+        return False
+
+    def _collect_mutable_chain(self, start: torch.fx.Node) -> set[torch.fx.Node]:
+        """Collects all nodes that lead to a mutable buffer."""
+        stack = [start]
+        node_parents: dict[torch.fx.Node, list[torch.fx.Node]] = {}
+
+        # Generate a map from a node to its parents
+        while stack:
+            node = stack.pop()
+            if node in node_parents:
+                continue
+
+            parents = [
+                arg
+                for arg in pytree.arg_tree_leaves(*node.args, **node.kwargs)
+                if isinstance(arg, torch.fx.Node)
+            ]
+            node_parents[node] = parents
+            stack.extend(parents)
+
+        # Visit the graph in topological order and trace from the buffer to the start node
+        connected = set()
+        for node in self.graph.nodes:
+            if node not in node_parents:
+                continue
+            if self._is_mutable_buffer(node) or any(
+                parent in connected for parent in node_parents[node]
+            ):
+                connected.add(node)
+
+        return connected if start in connected else set()
+
     def _find_mutable_buffers(self) -> set[torch.fx.Node]:
-        """Find mutable buffers by identifying copy_ operations.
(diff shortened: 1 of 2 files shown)


tom-arm:
Partly solves the issue raised here: https://github.com/pytorch/ao/issues/4138. It should stop the mutable buffer being folded into a constant

torchao/quantization/pt2e/constant_fold.py
```diff
@@ -94,19 +94,66 @@ def __init__(
...
         for node in self.module.graph.nodes:
             if (
                 node.op == "call_function"
                 and hasattr(node.target, "_schema")
-                and "copy_" in str(node.target)
+                and ("copy_" in str(node.target) or "put_" in str(node.target))
             ):
-                # The first argument of copy_ is the mutable buffer
+                # The first argument of copy_ or put_ is the mutable input
+                # Work through the graph to find all nodes that lead to a mutable buffer
```
jerryzh168: nit: Work --> Walk

torchao/quantization/pt2e/constant_fold.py
```diff
@@ -94,19 +94,66 @@ def __init__(
...
+        named_buffers = dict(self.module.named_buffers())
+        if node.op == "placeholder":
+            return True
+
+        if node.op == "get_attr" and str(node.target) in named_buffers:
+            return True
+
+        return False
+
+    def _collect_mutable_chain(self, start: torch.fx.Node) -> set[torch.fx.Node]:
```
jerryzh168: can this implementation be simplified a bit based on https://github.com/pytorch/pytorch/blob/56e17506de10782f7dbd5eef32b17b6f2250b1b9/torch/ao/quantization/fx/utils.py#L195

torchao/quantization/pt2e/constant_fold.py
tom-arm: Done

torchao/quantization/pt2e/constant_fold.py
tom-arm: Thanks for the suggestion! This does simplify things

torchao/quantization/pt2e/constant_fold.py
```diff
@@ -94,19 +95,37 @@ def __init__(
...
-                and "copy_" in str(node.target)
+                and ("copy_" in str(node.target) or "put_" in str(node.target))
             ):
-                # The first argument of copy_ is the mutable buffer
+                # The first argument of copy_ or put_ is the mutable input.
+                # If any producer in the chain is a mutable buffer, mark
+                # all producers as mutable to prevent constant folding.
                 if len(node.args) > 0 and isinstance(node.args[0], torch.fx.Node):
-                    mutable_buffers.add(node.args[0])
+                    producer_nodes = collect_producer_nodes(node.args[0])
```
jerryzh168, line 123: oh didn't know this works as is, but great it works

jerryzh168 (approved):
Thanks for the fix!

torchao/quantization/pt2e/constant_fold.py
```diff
@@ -10,6 +10,7 @@
 import torch
 import torch.utils._pytree as pytree
 from torch._inductor.freezing_utils import maybe_set_is_frozen_param
+from torch.ao.quantization.fx.utils import collect_producer_nodes
```
jerryzh168, line 13: btw, we can also copy this util to torchao, since we want to avoid dep on torch.ao.quantization, but better to do this in a separate PR, please feel free to put up PR for this one as well, otherwise I can work on this a bit later

tom-arm: thanks for letting me know, I don't mind doing that for you.

so just copy the `collect_producer_nodes` function to `torchao/quantization/utils.py`?

jerryzh168: thanks, that would be great, you can copy that here: https://github.com/pytorch/ao/blob/main/torchao/quantization/pt2e/graph_utils.py

tom-arm: I have raised a PR for this here: https://github.com/pytorch/ao/pull/4294

PR #4182, opened 2026-03-26
support Per Group in Float8DynamicActivationFloat8WeightConfig (#4182)

zlin888:
Summary:

this is supported in mtia. added hw check the guard this support. No kernel is added because it is replying on the plain pytorch implementaion

Differential Revision: D97987011

### DIFF
--- a/torchao/float8/inference.py
+++ b/torchao/float8/inference.py
@@ -17,6 +17,7 @@ from torchao.float8.types import FP8Granularity
 from torchao.utils import (
     is_MI300,
     is_MI350,
+    is_mtia,
     is_sm_at_least_89,
 )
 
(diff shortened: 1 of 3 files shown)


zlin888:
@jerryzh168 @jcaip pls take a look. thanks!

torchao/float8/inference.py
```diff
@@ -304,5 +315,10 @@ def _check_hardware_support(
         assert is_sm_at_least_89(), (
             "Float8 1x128 activation and 128x128 weight scaling requires CUDA compute capability ≥8.9."
         )
+    elif is_per_group:
+        logger.warning(
```
jerryzh168: we probably want to add at least `is_sm_at_least_89`? since IIRC that's the first hardware that introduced fp8 tensor core

PR #4186, opened 2026-03-26
fix undefined values for tail elements in act quant kernels

iamzainhuda:
### Summary
Fixed a tail-handling bug in the blockwise FP8 activation quantization kernels that was corrupting reciprocal scale tensors and causing ```test_triton_fp8_gemm_1x128_128x128``` to fail on ragged shapes. 

The GEMM consumes a_s / b_s scale tensors produced by the quantization kernels, and those scale tensors could be corrupted at the tensor edges. In the LHS activation quant kernel, masked tail lanes were still storing scales into a compact column-major as_strided buffer. Because that buffer has no padding, logically invalid row writes could alias valid scale entries from the next column. The GEMM then used those corrupted scales and SQNR collapsed on small or ragged M.

Updated RHS and LHS blockwise quant kernels with ```tl.load(..., other=0.0)``` for masked tail loads. And masked reciprocral scale ```tl.store``` calls so invalid lanes don't write to scale buffers.

Failures before fix:
image

### Testing
  - ```pytest -q test/prototype/blockwise_fp8_training/test_blockwise_kernels.py -k test_triton_fp8_gemm_1x128_128x128```
  - ```pytest -q test/prototype/blockwise_fp8_training/test_blockwise_kernels.py``` 
     - now passes entirely (42 passed)

### DIFF
--- a/torchao/prototype/blockwise_fp8_training/kernels.py
+++ b/torchao/prototype/blockwise_fp8_training/kernels.py
@@ -318,7 +318,7 @@ def triton_fp8_blockwise_act_quant_lhs_kernel(
     k_offs = pid_k * BLOCK_SIZE + tl.arange(0, BLOCK_SIZE)
     x_offs = m_offs[:, None] * x_stride_dim_0 + k_offs[None, :] * x_stride_dim_1
     x_mask = (m_offs[:, None] < M) & (k_offs[None, :] < K)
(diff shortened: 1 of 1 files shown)


danielvegamyhre:
(this fix is indeed needed, just note that the triton fp8 blockwise gemms were experimental and not performant, not actually used. we use torch._scaled_mm which dispatches to cublas)

PR #4187, opened 2026-03-26
add torch.compile test for Float8BlockwiseLinear

iamzainhuda:
## Summary
- Add `torch.compile` testing to `Float8BlockwiseLinear` forward and backward passes, verifying fullgraph compilation, numerical correctness, and no recompilation across repeated calls with the same shapes
- Refactored existing `test_blockwise_quant_linear_fwd_bwd` into a shared `_run_blockwise_quant_linear_fwd_bwd` helper that supports both eager and compiled execution paths
- Fixes backward pass assertions to correctly compare gradients (input grad vs input grad, weight grad vs weight grad)

## Testing

```
pytest test/prototype/blockwise_fp8_training/test_blockwise_linear.py
```

- **Eager mode (`test_blockwise_quant_linear_fwd_bwd`)**: Parametrized across `in_features=[4096]`, `out_features=[128256]`, `batch_size=[1, 8]`, `block_size=[128]` — validates forward SQNR >= 25.0 and backward grad SQNR >= 30.0 against a reference `nn.Linear`
- **Compile mode (`test_blockwise_quant_linear_compile_fullgraph_fwd_bwd`)**: Runs with `torch.compile(fullgraph=True)` using `CompileCounterWithBackend("inductor")` to assert:
  - The model traces into a **single compiled frame** (no graph breaks)
  - **No recompilation** on a second forward/backward call with the same input shapes
  - Numerical correctness matches the eager reference (same SQNR thresholds)
- Both paths check for NaN-free outputs and gradients

### DIFF
--- a/test/prototype/blockwise_fp8_training/test_blockwise_kernels.py
+++ b/test/prototype/blockwise_fp8_training/test_blockwise_kernels.py
@@ -109,7 +109,8 @@ def test_triton_quantize_fp8_act_quant_lhs(block_size):
     x[0, :block_size] = 0.0
 
     triton_fp8, triton_scale = triton_fp8_blockwise_act_quant_lhs(
-        x, block_size=block_size
+        x,
+        block_size=block_size,
     )
 
     ref_fp8, ref_scale = torch_blockwise_scale_act_quant_lhs(x, tile_size=block_size)
@@ -151,7 +152,8 @@ def test_triton_quantize_fp8_act_quant_rhs(block_size: int):
     x[:block_size, :block_size] = 0.0
 
     triton_fp8, triton_scale = triton_fp8_blockwise_act_quant_rhs(
-        x, block_size=block_size
+        x,
+        block_size=block_size,
     )
 
     ref_fp8, ref_scale = torch_blockwise_scale_act_quant_rhs(x, block_size=block_size)
@@ -186,16 +188,15 @@ def test_triton_quantize_fp8_act_quant_rhs(block_size: int):
 )
 @pytest.mark.parametrize("block_size", [128, 256])
 @pytest.mark.parametrize("M,K", [(4096, 1024), (4096, 4 * 4096)])
-def test_triton_quantize_fp8_act_quant_transposed_lhs(
-    M, K, block_size: int
-):
+def test_triton_quantize_fp8_act_quant_transposed_lhs(M, K, block_size: int):
     device = "cuda"
     x = torch.randn(M, K, device=device)
 
     x[0, :block_size] = 0.0
 
     triton_fp8, triton_scale = triton_fp8_blockwise_act_quant_transposed_lhs(
-        x, block_size=block_size
+        x,
+        block_size=block_size,
     )
 
     ref_fp8, ref_scale = torch_blockwise_scale_act_quant_lhs(
@@ -239,7 +240,8 @@ def test_triton_quantize_fp8_weight_quant_rhs(M, K, block_size: int):
     x[:block_size, :block_size] = 0.0
 
     triton_fp8, triton_scale = triton_fp8_blockwise_weight_quant_rhs(
-        x, block_size=block_size
+        x,
+        block_size=block_size,
     )
     ref_fp8, ref_scale = torch_blockwise_scale_weight_quant(x, tile_size=block_size)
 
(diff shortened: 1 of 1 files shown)


danielvegamyhre:
@iamzainhuda i think there was a miscommunication, we don't want to directly wrap an individual triton custom op in torch.compile and test - we want to compile a full blockwise linear layer (this [test](https://github.com/pytorch/ao/blob/02105d46c61dc80a8c9d39d5836e827ba3af8439/test/prototype/blockwise_fp8_training/test_blockwise_linear.py#L30)) and ensure there are no graph breaks (fullgraph=True) and that numerics of outputs/grads undergo the pass with the same threshold testing as the eager mode tests.

JiwaniZakir (commented):
The `_maybe_compile` helper in `test_blockwise_kernels.py` is clean and avoids repetition, but wrapping a freshly-created lambda each call means `torch.compile` receives a new function object on every test invocation with no opportunity for cache reuse across parametrized runs — this is fine for correctness but adds compile overhead. More importantly, there are no `torch._dynamo.reset()` calls between tests: compiled artifacts from one parametrized combination (e.g., `block_size=128`) can leak into the next (`block_size=256`), which can mask failures or produce misleading error messages in CI. It would be worth adding a `torch._dynamo.reset()` at the start of each test body when `use_compile=True`, or using a pytest fixture/autouse teardown for that. Also, the `use_compile` parametrize decorator is placed *closer* to the function than `block_size` in all cases, which means `use_compile` varies as the inner loop in the test matrix — this is a minor point but being consistent across all five tests about decorator ordering is good hygiene and it is consistent here.

iamzainhuda:
> @iamzainhuda i think there was a miscommunication, we don't want to directly wrap an individual triton custom op in torch.compile and test - we want to compile a full blockwise linear layer (this [test](https://github.com/pytorch/ao/blob/02105d46c61dc80a8c9d39d5836e827ba3af8439/test/prototype/blockwise_fp8_training/test_blockwise_linear.py#L30)) and ensure there are no graph breaks (fullgraph=True) and that numerics of outputs/grads undergo the pass with the same threshold testing as the eager mode tests.

that makes sense (esp w.r.t understanding graph breaks), updated the PR. added a compile test in blockwise_linear with a full layer, used a smaller shape for compile time sake. verified full graph compilation, numerical correctness and no erroneous recompilation across multiple calls.

test/prototype/blockwise_fp8_training/test_blockwise_linear.py
```diff
@@ -41,33 +39,119 @@ def test_blockwise_quant_linear_fwd_bwd(
...
+    out_features,
+    batch_size,
+    block_size,
+):
+    _run_blockwise_quant_linear_fwd_bwd(
+        in_features,
+        out_features,
+        batch_size,
+        block_size,
+        compile_mode=True,
```
danielvegamyhre, line 156: can we just set this "use compile" bool as a pytest.mark.parametrize parameter, and combine the eager and compile unit tests into one test (rather than duplicating just to change this bool?). to keep things simple and be consistent with the usual parttern in other torchao tests.

or is there any reason these are separated now?

test/prototype/blockwise_fp8_training/test_blockwise_linear.py
```diff
@@ -41,33 +39,119 @@ def test_blockwise_quant_linear_fwd_bwd(
...
     ).cuda()
-
     layer_test = Float8BlockwiseLinear.from_float(copy.deepcopy(layer_ref))
+    compiled_frame_counter = None
+    compiled_step = None
+
+    if compile_mode:
+        with torch._dynamo.config.patch(trace_autograd_ops=True):
+            torch._dynamo.reset()
+            compiled_frame_counter = CompileCounterWithBackend("inductor")
```
danielvegamyhre, line 49: this is cool, hadn't used this before

test/prototype/blockwise_fp8_training/test_blockwise_linear.py
iamzainhuda: i had it separated for 1) compile test uses a smaller shape for compile time/runtime cost and 2) easier to reason about failures when a "test_linear_compile" is failing vs "test_linear_fwd_bwd" as i felt they were testing reasonably different things. so it'd make less sense to have them both as "test_linear_fwd_bwd". happy to move them together if you'd like!

danielvegamyhre: ok sure, sgtm. i think the compile time thing is a bit overkill for a unit test unless it's really long, but i agree the test failures will be more readable at a glance, since with parameterized `use_compile` bool the failed config just shows up as like `test_this_thing[4096-4096-1-128-true]` which gets confusing especially if we add more bool test params later

PR #4194, opened 2026-03-27
add hadamard option to low precision attention api

howardzhang-cv:
* __->__ #4194

## Summary
- Added Hadamard on QKV tensor support for the low precision attention API, passed through from apply_low_precision_attention to select the hadamard fused kernel
- Added new kernels (triton_hadamard_qkv_quantization.py and triton_hadamard_rope_qkv_quantization.py) in triton for fused hadamard and QKV quantization (with rope fusion as well)
  - Because of the way per-head quantization works with the sequence chunks, we need to store the hadamard outputs in a temp buffer, which eliminates some of the benefit of the fusion. However, it still saves one global read of the QKV tensors, which experiments show still benefits runtime quite a bit, so the fusion is still much better than just running hadamard separately.
- Added hadamard option to the benchmarks
- Replaced some duplicate compute_num_chunks code

#### Single Attention Layer
image
For a single attention layer, it got slower (from 1.36x speedup to 1.29x speedup on the highest sequence length). The SQNR does not really improve because we're testing with random tensors, which are pretty uniform already (Hadamard is intended to spread out intensity for better quantization accuracy)

#### LLaMA3 Prefill
image
Perplexity used to go from 7.54 -> 7.62. Now it is noticeably better, going from 7.54 -> 7.57. The speedup dropped from 1.23x to 1.15x at the highest sequence length.

### DIFF
--- a/test/prototype/attention/test_fp8_attention.py
+++ b/test/prototype/attention/test_fp8_attention.py
@@ -107,8 +107,7 @@ class TestFP8FA3Attention(TestCase):
     )
     @common_utils.parametrize("shape", [(2, 8, 1024, 64), (1, 16, 1024, 128)])
     @common_utils.parametrize("dtype", [torch.bfloat16, torch.float16])
-    @common_utils.parametrize("hadamard", ["NONE", "QKV"])
-    def test_sdpa_accuracy(self, shape, dtype, hadamard):
+    def test_sdpa_accuracy(self, shape, dtype):
         B, H, S, D = shape
         q = torch.randn(B, H, S, D, device="cuda", dtype=dtype)
         k = torch.randn(B, H, S, D, device="cuda", dtype=dtype)
@@ -120,7 +119,7 @@ class TestFP8FA3Attention(TestCase):
         activate_flash_attention_impl("FA3")
         try:
             with torch.no_grad():
-                out_fp8 = fp8_fa3_sdpa(q, k, v, is_causal=False, hadamard=hadamard)
+                out_fp8 = fp8_fa3_sdpa(q, k, v, is_causal=False)
         finally:
             restore_flash_attention_impl()
 
@@ -128,7 +127,7 @@ class TestFP8FA3Attention(TestCase):
         self.assertGreater(
             sqnr.item(),
             25.0,
-            f"SQNR {sqnr.item():.2f} dB below 25 dB for shape={shape}, dtype={dtype}, hadamard={hadamard}",
+            f"SQNR {sqnr.item():.2f} dB below 25 dB for shape={shape}, dtype={dtype}",
         )
 
     @unittest.skipUnless(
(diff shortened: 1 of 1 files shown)


drisspg:
@claude /pr-review detailed

Please do a deep review focused on subtle correctness bugs and regressions that are easy for human reviewers to miss. Prioritize findings over summary.

Specifically, look for:

incorrect edge-case behavior
missing or weak test coverage
backward-compatibility risks
silent behavioral changes
dtype/device/shape/stride/autograd/dispatch issues
aliasing, mutation, lifetime, and ownership bugs
concurrency or thread-safety problems
error-path bugs and partial-state failures
performance traps that come from correctness fixes or hidden extra work
Ignore style, formatting, and nits unless they hide a real bug. Be skeptical of code that “looks fine” and focus on failure modes, invariants, and assumptions not enforced by
tests.

drisspg:
@claude 
Please do a deep review focused on subtle correctness bugs and regressions that are easy for human reviewers to miss. Prioritize findings over summary.

Specifically, look for:

incorrect edge-case behavior
missing or weak test coverage
backward-compatibility risks
silent behavioral changes
dtype/device/shape/stride/autograd/dispatch issues
aliasing, mutation, lifetime, and ownership bugs
concurrency or thread-safety problems
error-path bugs and partial-state failures
performance traps that come from correctness fixes or hidden extra work
Ignore style, formatting, and nits unless they hide a real bug. Be skeptical of code that “looks fine” and focus on failure modes, invariants, and assumptions not enforced by
tests.

howardzhang-cv:
@drisspg Can you take another look? Just addressed some of the claude review comments:
1. I fixed the benchmark_sdpa bug.
2. Added a comment for the kv_temp. I think it's better to keep it as the same for now, considering this kernel is only called internally and will always be used in a stream and called in a way that there won't be conflict, but I did add a comment acknowledging potential issues.
3. Added some tests for hadamard, was going to do this in a separate PR after implementing Hadamard mode for V-only, but doing it here makes sense too.
4. Changed from str(hadamard) to hadamard.value instead to avoid the pytorch 2.11 issue.

A lot of the other comments I think aren't really applicable in my opinion or are wrong, but happy to look into it more if you think I should.

test/prototype/attention/test_fp8_attention.py
```diff
@@ -118,15 +120,15 @@ def test_sdpa_accuracy(self, shape, dtype):
...
             with torch.no_grad():
-                out_fp8 = fp8_fa3_sdpa(q, k, v, is_causal=False)
+                out_fp8 = fp8_fa3_sdpa(q, k, v, is_causal=False, hadamard=hadamard)
         finally:
             restore_flash_attention_impl()
 
         sqnr = compute_error(out_ref, out_fp8)
         self.assertGreater(
             sqnr.item(),
             25.0,
```
drisspg, line 129: should we expect/check sqnr to be higher w/ hadamard?

test/prototype/attention/test_fp8_attention.py
```diff
@@ -153,23 +156,26 @@ def test_rope_sdpa_accuracy(self, shape, dtype):
         activate_flash_attention_impl("FA3")
         try:
             with torch.no_grad():
-                out_fp8 = fp8_fa3_rope_sdpa(q, k, v, cos, sin, is_causal=False)
+                out_fp8 = fp8_fa3_rope_sdpa(
```
drisspg: ditto

test/prototype/attention/test_fp8_attention.py
howardzhang-cv: The tests here were more just making sure Hadamard was calculated correctly (with inverse and stuff). Because it's uniformly sampled, Hadamard doesn't actually improve SQNR. It only improves when there's outliers. I just added a separate set of tests to check that Hadamard improves accuracy when the inputs have outliers.

test/prototype/attention/test_fp8_attention.py
howardzhang-cv: see above

PR #4198, opened 2026-03-29
Refactor SM requirement device check

namgyu-youn:
`_is_hopper()` was a device-specific check targeting only Hopper GPUs, while `is_sm_at_least_90()` correctly expresses a minimum SM requirement, making the code forward-compatible with future SM90+ architectures like Blackwell.


namgyu-youn:
@pytorchbot label "module: not user facing"

cc @howardzhang-cv

PR #4200, opened 2026-03-30
Create recipe for flux2pro running on AMD (#4200)

hualazimi0425:
Summary:

Tested based on D98755945

Reviewed By: colinchan15

Differential Revision: D98537991

### DIFF
--- a/test/quantization/quantize_/workflows/float8/test_float8_tensor.py
+++ b/test/quantization/quantize_/workflows/float8/test_float8_tensor.py
@@ -38,6 +38,7 @@ from torchao.utils import (
     is_sm_at_least_100,
     torch_version_at_least,
 )
+from unittest.mock import patch
 
 # Needed since changing args to function causes recompiles
 torch._dynamo.config.cache_size_limit = 128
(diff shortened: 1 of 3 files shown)


hualazimi0425:
@vkuzo @jerryzh168 Could you please add the `module:float8` label to this PR? CI requires it and I don't have permission as an external contributor. Would also appreciate a review when you get a chance. Thanks\!

test/quantization/quantize_/workflows/float8/test_float8_tensor.py
```diff
@@ -1489,5 +1490,118 @@ def test_create_tensor_out_of_inference_mode(self):
 
 common_utils.instantiate_parametrized_tests(TestFloat8Tensor)
 
+
+class TestMI350HardwareSupport(common_utils.TestCase):
```
jerryzh168, line 1494: btw I don't think we need to test these flags, instead a simplified version of this: https://github.com/pytorch/ao/blob/7dd8be5dd4acc6e81783aa0c41bd88fcfc27a213/test/quantization/quantize_/workflows/float8/test_float8_tensor.py#L365-L388 would be more helpful

PR #4201, opened 2026-03-30
[mxfp8 training] update triton_to_mxfp8_dim0 nan handling; fix offset int32 overflow issue

danielvegamyhre:
## Summary
There are inconsistencies in how torch vs triton handle special values like NaN in the mxfp8 quantization. This PR:
- Adds new unit tests for special values and subnormals, ensuring triton matches torch behavior
- Updates triton to match torch behavior
- Updates both triton and torch to have explicit handling for inf and Nan when computing the scale reciprocal 

## Behavior
- If any input element in 1x32 block is NaN, the scale will be NaN, and thus the output 1x32 block will be NaN
- Reciprocal handling of special values (nan, inf) is now consistent for torch and triton 

## Benchmarks
```
(torch) dev@gpu-dev-e6e9fb8d:~/ao$ CUDA_VISIBLE_DEVICES=1 PYTHONPATH=/home/$USER/ao:$PYTHONPATH python benchmarks/mx_formats/cast_bench.py --mode dim0_mxfp8_triton_rceil --M 32768 --K 7168
M 32768 K 7168 BLOCK_SIZE 32
GPU: NVIDIA B200
torch version: 2.11.0+cu130
triton version: 3.6.0
mode: dim0_mxfp8_triton_rceil
time_us 121.88799679279327
mem_bw_gbps 5841.289731017193
```

## Additional bug fix
- I found for very large tensors, the offset calculations would exceed int32 maximum value and overflow, resulting in illegal memory access. This was discovered using a "repro tensor" given to me by someone who reported the kernel producing a NaN unexpectedly, and I think this overflow is responsible.

### DIFF
--- a/test/prototype/mx_formats/test_kernels.py
+++ b/test/prototype/mx_formats/test_kernels.py
@@ -819,6 +819,9 @@ def test_triton_mxfp8_dim0_overflow_underflow(scaling_mode):
 @pytest.mark.skipif(
     not is_cuda_version_at_least(12, 8), reason="CUDA version >= 12.8 required"
 )
+@pytest.mark.skipif(
+    not is_sm_at_least_100(), reason="CUDA capability 10.0 or greater required"
+)
 @pytest.mark.parametrize("scaling_mode", (ScaleCalculationMode.RCEIL,))
 def test_all_nan_block_scale_behavior(scaling_mode):
     """


danielvegamyhre:
edit: nevermind, managed to recover performance

vkuzo:
@claude review this PR

danielvegamyhre:
Responding to claude comments:

1. >dim1 kernel inconsistency

Updated dim1 kernel to be consistent. It isn't actually used in practice anymore and we'll probably delete soon but for consistency's sake I updated.

2. >But the Triton kernel at line 483 uses bare tl.max(x, axis=axis) where x is tl.abs(x_block_r). Since abs(NaN) = NaN and tl.max will propagate NaN, the Triton kernel will produce a NaN max_abs when any element in the block is NaN. 

This is not true, I confirmed via manual testing that `tl.max` ignores NaNs, so the scale will not be NaN if any values in the 1x32 block are non-NaN. This matches the torch ref, which explicitly ignores nans using `torch.where`.

3. > In PyTorch, torch.clamp propagates NaN (NaN stays NaN), so this is fine. However, this is a subtle invariant worth a brief comment

Added comment noting this

4. > typo in comment

Fixed typo in comment.

5. >5. All-NaN block edge case (mx_tensor.py:217-219)

After some testing, my current understanding is:
- CUDA/TE with `__hmax` :
      - block of mixed real numbers and NaNs = real number
      - block of all nans = nan
-  Triton with `tl.max` amax calculation:
     - block of mixed real numbers and NaNs = real number
     - block of all nans = nan
- Torch (current impl) amax calculation:
    - block of mixed real numbers and NaNs = real number
    - block of all nans = **-inf** (discrepancy!)
    
I have updated the torch impl to match the CUDA/TE and Triton behavior of "block of all nans => nan amax"

6. >dim1 special value test coverage

Not adding since i intend to delete dim1 triton kernel in the next PR in the stack.


7. >variable naming consistency in dim1 kernel

Updated

test/prototype/mx_formats/test_kernels.py
```diff
@@ -625,3 +625,266 @@ def test_cuda_mx_dim0_not_supported():
...
+    )
+
+
+@pytest.mark.skipif(not has_triton(), reason="unsupported without triton")
+@pytest.mark.skipif(
+    not is_sm_at_least_100() and not is_MI350(),
+    reason="mxfp8 requires CUDA capability 10.0 or greater or ROCm gfx950 or greater.",
+)
+@pytest.mark.parametrize("scaling_mode", (ScaleCalculationMode.RCEIL,))
+def test_triton_mxfp8_dim0_overflow_underflow(scaling_mode):
```
slayton58, line 733: I guess I'm not sure exactly what this is testing..

`inputs > 448.` means we'll have a `scaling_factor < 1`, and so the underflow vals should be 0 after conversion, but this isn't actually tested

test/prototype/mx_formats/test_kernels.py
```diff
@@ -625,3 +625,266 @@ def test_cuda_mx_dim0_not_supported():
...
+    """Test with values near overflow and underflow thresholds."""
+    # Values near float8_e4m3fn limits
+    f8_max = torch.finfo(torch.float8_e4m3fn).max  # ~448
```
slayton58: Tiny nit, we have exact values for these, the approx `~` shouldn't be there. max == 448, min subnormal == 2^-9.

test/prototype/mx_formats/test_kernels.py
```diff
@@ -625,3 +625,266 @@ def test_cuda_mx_dim0_not_supported():
...
+    torch.testing.assert_close(x_s_t, x_s_ref, rtol=0, atol=0)
+
+
+@pytest.mark.skipif(not has_triton(), reason="unsupported without triton")
+@pytest.mark.skipif(
+    not is_sm_at_least_100() and not is_MI350(),
+    reason="mxfp8 requires CUDA capability 10.0 or greater or ROCm gfx950 or greater.",
+)
+@pytest.mark.parametrize("scaling_mode", (ScaleCalculationMode.RCEIL,))
+def test_triton_mxfp8_dim0_extreme_range(scaling_mode):
```
slayton58: Same comment on what's being tested - we're testing that things match a reference, but not necessarily that the observed behaviour is actually correct -- both imo should be tested. This is also a good candidate to merge with the overflow/underflow test above, as I think they only differ in terms of what the extreme values are (which is parametrize-able)

test/prototype/mx_formats/test_kernels.py
```diff
@@ -625,3 +625,266 @@ def test_cuda_mx_dim0_not_supported():
...
+    torch.testing.assert_close(x_s_t, x_s_ref, rtol=0, atol=0)
+
+
+@pytest.mark.skipif(not has_triton(), reason="unsupported without triton")
+@pytest.mark.skipif(
+    not is_sm_at_least_100() and not is_MI350(),
+    reason="mxfp8 requires CUDA capability 10.0 or greater or ROCm gfx950 or greater.",
+)
+@pytest.mark.parametrize("scaling_mode", (ScaleCalculationMode.RCEIL,))
+def test_triton_mxfp8_dim0_denormals_subnormals(scaling_mode):
```
slayton58: Same comment on what's being tested.

test/prototype/mx_formats/test_kernels.py
```diff
@@ -625,3 +625,266 @@ def test_cuda_mx_dim0_not_supported():
...
+    assert not x_mx_t.isnan().any(), "quantized tensor should not contain NaNs"
+    assert not x_s_t.isnan().any(), "scales should not contain NaNs"
+    torch.testing.assert_close(x_mx_t, x_mx_ref, rtol=0, atol=0)
+    torch.testing.assert_close(x_s_t, x_s_ref, rtol=0, atol=0)
+
+
+@pytest.mark.skipif(not has_triton(), reason="unsupported without triton")
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA not available")
+@pytest.mark.parametrize("scaling_mode", (ScaleCalculationMode.RCEIL,))
+def test_all_nan_block_scale_behavior(scaling_mode):
```
slayton58, line 826: This is the opposite of what we discussed, no? We discussed mixed real + NaN -> scale = NaN

test/prototype/mx_formats/test_mx_tensor.py
```diff
@@ -341,6 +341,74 @@ def test_exponent_nan_in(elem_dtype):
     assert not torch.any(torch.isnan(tensor_mx.scale[1:]))
 
 
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA not available")
+@pytest.mark.parametrize("elem_dtype", SUPPORTED_ELEM_DTYPES)
+def test_all_nan_blocks(elem_dtype):
```
slayton58, line 337: Same again w.r.t. mixed behaviour

test/prototype/mx_formats/test_kernels.py
danielvegamyhre: Oh, in this test the comments are outdated but it does indeed validate mixed real + nan = scale nan. Updated the comment.

test/prototype/mx_formats/test_mx_tensor.py
danielvegamyhre: Ah this test was missed when i updated the behavior.. thanks. Will rerun all tests to double check

test/prototype/mx_formats/test_kernels.py
danielvegamyhre: I was just testing the torch vs triton handling of special cases is consistent, since it previously wasn't. I was treating the torch reference the source of truth for correctness, but since we are also changing aspects of that here, you make a good point we should also validate the result here as well. 

Updated to verify both consistency and expected results (i.e. validate which vals underflow and which don't)

test/prototype/mx_formats/test_kernels.py
danielvegamyhre: Makes sense, I deleted some tests and focused on having fewer, higher value tests.

PR #4206, opened 2026-03-31
[mx] Triton RCEIL rounding for non-CUDA backends

ugolowic:
In the context of https://github.com/pytorch/ao/issues/3576:
The `USE_PTX` flag in MXFP8 triton kernels was gated by `not IS_ROCM`, which assumed only CUDA and ROCm backends exist. On any other non-CUDA device (e.g. XPU), this incorrectly enabled CUDA PTX inline assembly (`cvt.rp.satfinite.ue8m0x2.f32`), causing errors.

### DIFF
--- a/torchao/prototype/mx_formats/kernels.py
+++ b/torchao/prototype/mx_formats/kernels.py
@@ -25,6 +25,7 @@ from torchao.utils import (
     is_mslk_version_at_least,
     is_ROCM,
     is_sm_at_least_100,
+    is_XPU,
 )
 
 logger = logging.getLogger(__name__)
@@ -455,11 +456,13 @@ else:
         raise AssertionError("needs triton")
 
 
-_triton_kernels_available = (
-    has_triton()
-    and torch.cuda.is_available()
-    and (is_sm_at_least_100() and is_cuda_version_at_least(12, 8))
+_triton_kernels_available = has_triton() and (
+    (
+        torch.cuda.is_available()
+        and (is_sm_at_least_100() and is_cuda_version_at_least(12, 8))
+    )
     or (is_ROCM() and is_MI350())
+    or is_XPU()
 )
 
 if _triton_kernels_available:
@@ -468,6 +471,7 @@ if _triton_kernels_available:
     from torch.library import triton_op, wrap_triton
 
     IS_ROCM = tl.constexpr(is_ROCM())
+    IS_XPU = tl.constexpr(is_XPU())
 
     @triton.jit
     def _calculate_reciprocal_scale(scale_e8m0_biased):
@@ -689,7 +693,7 @@ if _triton_kernels_available:
             col_rcp_scale_fp32, col_scale_e8m0_r = _triton_calculate_scale_rceil(
                 x_block_abs_t_r,
                 axis=1,
-                USE_PTX=not IS_ROCM,
+                USE_PTX=(not IS_ROCM and not IS_XPU),
             )
         else:
             tl.static_assert(SCALING_MODE == "floor")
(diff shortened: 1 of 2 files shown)


vkuzo:
@claude review this PR

JiwaniZakir (commented):
The substitution of `is_ROCM()` with `torch.cuda.is_available()` in `triton_mxfp8_dequant_dim0` has a subtle correctness risk: on ROCm, `torch.cuda.is_available()` can return `True` because ROCm provides a HIP-based CUDA compatibility layer. If that's the case on a ROCm machine, `IS_CUDA` would be `True` and the PTX path would be taken—which is precisely what the original `not IS_ROCM` guard was meant to prevent, since PTX is CUDA-specific and will fail on ROCm. A more robust predicate would be something like `torch.version.hip is None` (or re-using `is_ROCM()` to keep the logic symmetrical), which distinguishes a true CUDA build from a ROCm/HIP build regardless of what `torch.cuda.is_available()` reports. It's also worth confirming whether there's a test that exercises the `SCALING_MODE == "nearest_even"` branch (lines ~686 and ~796) on a ROCm device so this regression would be caught automatically.

ugolowic:
I've updated the change fixing the logic errors.

@JiwaniZakir , thanks for your comments. Re: "nearest_even" - did you mean "floor"? The else branch is covered by UTs.

JiwaniZakir:
The fix correctly narrows `USE_PTX` to CUDA-only rather than `not IS_ROCM` — that assumption was always fragile and would break on any new backend. One thing worth verifying: the Triton RCEIL (`tl.constexpr` rounding) path produces numerically equivalent results to `cvt.rp.satfinite.ue8m0x2.f32` for the UE8M0 format, since the PTX instruction handles saturation. If there are edge cases (NaN, inf, subnormals), the fallback path should be tested to confirm matching behavior across backends. Also check whether `IS_CUDA` is being determined at kernel compile time via `tl.constexpr` — if it's evaluated at Python level before JIT compilation, that's fine, but if it's inside a triton kernel it needs to be a constexpr to avoid issues.

ugolowic:
> The fix correctly narrows `USE_PTX` to CUDA-only rather than `not IS_ROCM` — that assumption was always fragile and would break on any new backend. One thing worth verifying: the Triton RCEIL (`tl.constexpr` rounding) path produces numerically equivalent results to `cvt.rp.satfinite.ue8m0x2.f32` for the UE8M0 format, since the PTX instruction handles saturation. If there are edge cases (NaN, inf, subnormals), the fallback path should be tested to confirm matching behavior across backends. Also check whether `IS_CUDA` is being determined at kernel compile time via `tl.constexpr` — if it's evaluated at Python level before JIT compilation, that's fine, but if it's inside a triton kernel it needs to be a constexpr to avoid issues.

@JiwaniZakir
- there's no `IS_CUDA` - what do you mean?
- I believe the cases you mention are implemented in test_kernels.py - see https://github.com/pytorch/ao/pull/4213

kgajdamo (approved):
Looks good to me

xiaowangintel:
@ugolowic Please merge https://github.com/pytorch/ao/pull/4206 and https://github.com/pytorch/ao/pull/4213 first. And, could you please rebase to get onto a revision with green CI.

ugolowic:
> @ugolowic Please merge #4206 and #4213 first. And, could you please rebase to get onto a revision with green CI.

@xiaowangintel, see comment: https://github.com/pytorch/ao/pull/4271#discussion_r3257743637 - do you want me to merge all three commits in one PR?

ugolowic:
Moved the commit to https://github.com/pytorch/ao/pull/4213

PR #4208, opened 2026-03-31
Fix NF4Tensor narrow() crash with recent PyTorch nightly

vkuzo:
(95% claude)

Recent PyTorch nightly added a stricter check that prevents creating differentiable views on tensors that already have autograd metadata. When DTensor's _maybe_unpad_tensor calls narrow() on an NF4Tensor (after detach()), it fell through to the default __torch_function__ fallback path, which triggered a RuntimeError:

  "Attempted to make a tensor into a differentiable view, but the
   tensor already had autograd metadata associated with it."

This adds a torch.Tensor.narrow handler in __torch_function__ for NF4Tensor (similar to the existing view_as handler) that returns a fresh NF4Tensor with detached inner tensors, avoiding the autograd metadata conflict. It currently only supports identity narrows (start=0, length=size[dim]), which covers DTensor's redistribute unpadding path when tensor dimensions evenly divide across the mesh.

### DIFF
--- a/test/dtypes/test_nf4.py
+++ b/test/dtypes/test_nf4.py
@@ -771,9 +771,6 @@ class TestComm(FSDPTest):
 
     @skip_if_lt_x_gpu(2)
     @unittest.skipIf(not torch.accelerator.is_available(), "Need GPU available")
-    @unittest.skip(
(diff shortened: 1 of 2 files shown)


jerryzh168:
looks like there are some new test failures

andrewor14 (approved):

PR #4211, opened 2026-04-01
Remove `check_cpu_version` and `check_xpu_version` helpers

Xia-Weiwen:
This PR removes `check_cpu/xpu_version` since they are ambiguous and changes all uses of these helpers to `_is_device("cpu/xpu", tensor.device)`.


torchao/utils.py
```diff
@@ -1239,16 +1241,18 @@ def is_cuda_version_at_least(major: int, minor: int) -> bool:
     return (cuda_major, cuda_minor) >= (major, minor)
 
 
-def check_cpu_version(device, version="2.6.0"):
+def is_cpu_and_torch_version_ok(device, torch_version="2.6.0"):
```
jerryzh168: do we need this?

I'm wondering if we should just have a general device type check and remove these two helper functions

torchao/utils.py
```diff
@@ -1239,16 +1241,18 @@ def is_cuda_version_at_least(major: int, minor: int) -> bool:
...
     if isinstance(device, torch.device):
         device = device.type
-    return device == "cpu" and torch_version_at_least(version)
+    return device == "cpu" and torch_version_at_least(torch_version)
 
 
-def check_xpu_version(device, version="2.8.0"):
+def is_xpu_and_torch_version_ok(device, torch_version="2.8.0"):
```
jerryzh168: lowest torch support is 2.8.0 now: https://github.com/pytorch/ao/blob/main/.github/workflows/regression_test.yml

I think you don't need to check torch version anymore

jerryzh168: also we might be upgrading this to 2.9.1 soon as well, since 2.11.0 is out

Xia-Weiwen: Oh, thanks. I will remove these two helpers

torchao/utils.py
Xia-Weiwen: They are removed. Thanks.

torchao/utils.py
Xia-Weiwen: They are removed. Thanks.

jerryzh168:
@claude review

torchao/quantization/utils.py
```diff
@@ -485,9 +479,9 @@ def groupwise_affine_dequantize_tensor_from_qparams(
     assert groupsize > 1
     assert w_int4x8.dim() == 2
     # need to handle single column case so check for dtype/size from groupwise_affine_quantize_tensor_from_qparams path
-    if (w_int4x8.dtype == torch.uint8 or w_int4x8.shape[-1] > 1) and not (
-        check_cpu_version(w_int4x8.device)
-    ):
+    if (
```
jerryzh168: this condition seems complicated, might be good to simplify if possible

jerryzh168:
> There's also an existing is_device() helper in torchao/dtypes/utils.py:94 specifically for this purpose. 

I feel we should move this helper to `torchao/utils.py` and use this helper everywhere

torchao/quantization/utils.py
Xia-Weiwen: Oh, I am not familiar with the code here. And I feel the condition cannot be simplified. However, we can probably add a comment here to clarify the condition.
CC the last author @yanbing-j

Xia-Weiwen:
> > There's also an existing is_device() helper in torchao/dtypes/utils.py:94 specifically for this purpose.
> 
> I feel we should move this helper to `torchao/utils.py` and use this helper everywhere

Hi @jerryzh168 I feel that util function does not really simplify code.
```
is_device("cpu", tensor.device)
tensor.device.type == "cpu"

not is_device("cpu", tensor.device)
tensor.device.type != "cpu"
```
However, if you think the util function is better, I will use that.

Xia-Weiwen:
> > > There's also an existing is_device() helper in torchao/dtypes/utils.py:94 specifically for this purpose.
> > 
> > 
> > I feel we should move this helper to `torchao/utils.py` and use this helper everywhere
> 
> Hi @jerryzh168 I feel that util function does not really simplify code.
> 
> ```
> is_device("cpu", tensor.device)
> tensor.device.type == "cpu"
> 
> not is_device("cpu", tensor.device)
> tensor.device.type != "cpu"
> ```
> 
> However, if you think the util function is better, I will use that.

@jerryzh168 I have removed the `is_device` utils and added `is_on_device(tensor, device)` and `not_on_device(tensor, devices)` in torchao.utils. These two utils might looks less confusing and more convenient. How does that sound to you? Thanks.

Xia-Weiwen:
@claude Please review again

jerryzh168:
> > > There's also an existing is_device() helper in torchao/dtypes/utils.py:94 specifically for this purpose.
> > 
> > 
> > I feel we should move this helper to `torchao/utils.py` and use this helper everywhere
> 
> Hi @jerryzh168 I feel that util function does not really simplify code.
> 
> ```
> is_device("cpu", tensor.device)
> tensor.device.type == "cpu"
> 
> not is_device("cpu", tensor.device)
> tensor.device.type != "cpu"
> ```
> 
> However, if you think the util function is better, I will use that.

the util is to make sure it works for multiple situations, like device being string, being torch.device type etc., it's more for robustness of the check

I don't understand the motivation of `is_on_device` and `is_not_on_device`, why `is_device` is not enough here?

torchao/quantization/utils.py
```diff
@@ -465,11 +462,9 @@ def groupwise_affine_quantize_tensor_from_qparams(
         quant_max,
     )
     if w.shape[-1] > 1:
-        if (not (check_cpu_version(int_data.device))) and (
-            not (check_xpu_version(int_data.device))
-        ):
+        if not_on_device(int_data, ["cpu", "xpu"]):
```
jerryzh168: I see, I don't think we need special util just for having a list of args, I think it's better to just use `is_device`

Xia-Weiwen:
> > > > There's also an existing is_device() helper in torchao/dtypes/utils.py:94 specifically for this purpose.
> > > 
> > > 
> > > I feel we should move this helper to `torchao/utils.py` and use this helper everywhere
> > 
> > 
> > Hi @jerryzh168 I feel that util function does not really simplify code.
> > ```
> > is_device("cpu", tensor.device)
> > tensor.device.type == "cpu"
> > 
> > not is_device("cpu", tensor.device)
> > tensor.device.type != "cpu"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > However, if you think the util function is better, I will use that.
> 
> the util is to make sure it works for multiple situations, like device being string, being torch.device type etc., it's more for robustness of the check
> 
> I don't understand the motivation of `is_on_device` and `is_not_on_device`, why `is_device` is not enough here?

The motivation is to simplify code and make it more straightforward: the new utils work on tensors so that callers don't have to write `tensor.device` or `tensor.deice.type`, just `tensor`.
 ```
is_device("cpu", tensor.device)
tensor.device.type == "cpu"
is_on_device(tensor, "cpu")

not is_device("cpu", tensor.device) and not is_device("xpu", tensor.device)
tensor.device.type != "cpu" and tensor.device.type != "xpu"
not_on_device(tensor, ["cpu", "xpu"])
```
Anyway, I will rollback to `is_device` if you don't like them :)

Xia-Weiwen:
@claude review

jerryzh168 (approved):
yeah I think this is cleaner, please update the summary as well

PR #4213, opened 2026-04-01
[xpu][mx][test] Enable test_kernels on xpu

ugolowic:

This PR contains 3 commits necessary for enabling mx kernels tests on xpu:

### 1. Triton RCEIL rounding for non-CUDA backends
The `USE_PTX` flag in MXFP8 triton kernels was gated by `not IS_ROCM`, which assumed only CUDA and ROCm backends exist. On any other non-CUDA device (e.g. XPU), this incorrectly enabled CUDA PTX inline assembly (`cvt.rp.satfinite.ue8m0x2.f32`), causing errors.

### 2. Fix NaN scale propagation in RCEIL triton kernel for SPIR-V backends
On XPU's SPIR-V backend, `.to(uint8)` silently destroys NaN values: 

`fptoui(NaN)`, used by `.to(uint8)` is undefined behavior per the LLVM spec (returning 0 instead of 255) - see [The LLVM LangRef `fptoui` semantics section](https://llvm.org/docs/LangRef.html#fptoui-to-instruction)
The existing NaN workaround in `_triton_calculate_scale_rceil` correctly detects NaN blocks via `x != x + tl.max(nan_mask)`, but then the NaN injected into `max_abs` is subsequently killed by `.to(uint8)` further down the chain.

The proposed solution is to move the NaN correction after the `log2 -> ceil -> clamp -> +127 -> to(uint8)` chain.

### 3. Enable tests to run on xpu
Remove hardcoded `device="cuda"` and use `current_accelerator`. 
Add ids to existing test parameters for clearer test names.

### DIFF
--- a/test/prototype/mx_formats/test_kernels.py
+++ b/test/prototype/mx_formats/test_kernels.py
@@ -44,6 +44,7 @@ from torchao.prototype.mx_formats.kernels import (
 from torchao.prototype.mx_formats.mx_tensor import ScaleCalculationMode, to_dtype, to_mx
 from torchao.prototype.mx_formats.utils import to_blocked
 from torchao.utils import (
+    get_available_devices,
     is_cuda_version_at_least,
     is_MI350,
     is_sm_at_least_100,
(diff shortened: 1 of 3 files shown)


liangan1:
@claude  review this PR.

test/prototype/mx_formats/test_kernels.py
```diff
@@ -916,19 +949,23 @@ def test_all_nan_block_scale_behavior(scaling_mode):
 
 @pytest.mark.skipif(not has_triton(), reason="unsupported without triton")
 @pytest.mark.skipif(
-    not is_sm_at_least_100() and not is_MI350(),
+    torch.cuda.is_available() and not is_sm_at_least_100() and not is_MI350(),
```
ugolowic, line 948: @jerryzh168, I've used your suggestion from the other PR. On a side note, I think there's a logic error in this skipif. Shouldn't it be `is_sm_at_least_100() or not is_MI350()`? The two conditions are mutually exclusive.

test/prototype/mx_formats/test_kernels.py
```diff
@@ -445,16 +435,18 @@ def triton_to_mxfp8_dim0_reference(
...
-@pytest.mark.parametrize("K", (128, 256))
+@pytest.mark.parametrize("M", (128, 256), ids=lambda x: f"M={x}")
+@pytest.mark.parametrize("K", (128, 256), ids=lambda x: f"K={x}")
 @pytest.mark.parametrize(
     "scaling_mode", (ScaleCalculationMode.FLOOR, ScaleCalculationMode.RCEIL)
 )
 def test_triton_mxfp8_dim1_randn(M, K, scaling_mode):
-    x = torch.randn(M, K, dtype=torch.bfloat16, device="cuda")
+    x = torch.randn(
+        M, K, dtype=torch.bfloat16, device=torch.accelerator.current_accelerator()
```
xiaowangintel, line 444: 1. No Hardcoded Device Strings
Use @parametrize for Device Parameterization
2. All device-dependent tests should be decorated with @parametrize("device", get_available_devices()) and accept device as a function argument.

PR #3934, opened 2026-02-23
[optim] Add appearance dtype for optim subclasses to make things work better with DTensor

gau-nernst:
Fixes #3314

Context: https://github.com/pytorch/ao/issues/3314#issuecomment-3942475942

PR #4216, opened 2026-04-01
Add aten.view.dtype dispatch to optim state subclasses

winglian:
PR #3934 added an "appearance dtype" to OptimState8bit (e.g. bf16 wrapper over uint8 codes), but didn't implement aten.view.dtype. 

torch.compile'd 8bit Adam optimizer (e.g. adamw_torch_8bit in HuggingFace transformers) crashes on PyTorch 2.9.1+

fix: register aten.view.dtype for all three optim subclasses, updating only the appearance dtype (consistent with existing aten._to_copy.default behavior).

### DIFF
--- a/torchao/optim/subclass_4bit.py
+++ b/torchao/optim/subclass_4bit.py
@@ -195,10 +195,10 @@ def _(func, types, args, kwargs):
     x, shape = args
 
     if tuple(x.shape) == tuple(shape):
-        return OptimState4bit(x.codes, x.scale, x.qmap, x.signed, x._shape)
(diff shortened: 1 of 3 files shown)


gau-nernst:
@winglian Do you have a minimal reproducible example? I'm not sure why this is needed.

Also, `aten.view.dtype` doesn't make sense to me for low-bit tensor subclass...

andrewor14:
@claude review please

andrewor14:
@claude can you suggest a unit test

gau-nernst:
I don't think implementing aten.view.dtype is right. It means "reinterpreting bytes", which doesnt make sense for scaled low-bit. It would be great if there is a reproducible example to see if implementing this op is actually necessary. Then we can add it to the FSDP test as well.

PR #4229, opened 2026-04-02
fix blockwise FP8 scaled_mm scale layout in Float8BlockwiseLinear

iamzainhuda:
### Summary

- replace the non-Triton blockwise FP8 matmul path with functional scaled_mm semantics for forward, grad_x, and grad_weight
- fix the BlockWise128x128 RHS scale layout by padding the K-block dimension to a multiple of 4, as required by cuBLASLt
- fix the BlockWise1x128 RHS scale orientation in grad_weight by transposing the activation scales before the matmul
- use aten._scaled_mm_v2 only under torch.compile(fullgraph=True) so the compiler can trace the op without graph-breaking on the Python F.scaled_mm wrapper

#### What was going wrong?

The issue was not just that we were calling torch._scaled_mm, but that this path in our blockwise FP8 linear code was not matching the cuBLASLt scale-layout contract for blockwise scaling.
In particular:
- grad_x = grad_output @ weight uses RHS BlockWise128x128 scales
- cuBLASLt requires those scales to be K-major, with the K-block dimension padded to a multiple of 4
- our code was passing the unpadded layout, which caused incorrect behavior in the scaled-mm backend

There was a second layout bug in grad_weight = grad_output^T @ x:
- the RHS uses BlockWise1x128 scaling
- the quantized activation scales had the right values but the wrong orientation for the RHS scaled-mm call
- transposing those scales fixes the contract mismatch

Also to note:
- eager mode uses torch.nn.functional.scaled_mm
- compile mode calls aten._scaled_mm_v2 directly because, in this torch build, Dynamo fullgraph cannot trace through the Python F.scaled_mm wrapper even though it lowers to the same underlying op

### Testing

```
 pytest -q test/prototype/blockwise_fp8_training/test_blockwise_linear.py
```

cc @slayton58 @drisspg as part of #4209

### DIFF
--- a/torchao/prototype/blockwise_fp8_training/linear.py
+++ b/torchao/prototype/blockwise_fp8_training/linear.py
@@ -5,6 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 import torch
+import torch.nn.functional as F
 from torch import nn
 
 from torchao.core.config import AOBaseConfig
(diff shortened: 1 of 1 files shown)


drisspg:
@claude Please do a deep review focused on subtle correctness bugs and regressions that are easy for human reviewers to miss. Prioritize findings over summary. I dont care about anything that isnt a real bug drop all the noise but that, if you there isnt anything to report just say so with a small comment

drisspg:
> lso to note:
eager mode uses torch.nn.functional.scaled_mm
compile mode calls aten._scaled_mm_v2 directly because, in this torch build, Dynamo fullgraph cannot trace through the Python F.scaled_mm wrapper even though it lowers to the same underlying op

can you open a small reproducer on pytorch for this, this seems high priority to fix

torchao/prototype/blockwise_fp8_training/linear.py
```diff
@@ -23,6 +24,55 @@
 from torchao.utils import is_sm_at_least_90
 
 
+def _pad_blockwise_128x128_scale_k_major(scale: torch.Tensor) -> torch.Tensor:
+    # cuBLASLt requires BLK128x128 scales to be K-major with the K stride padded to a multiple of 4.
+    padded_k_blocks = (scale.shape[0] + 3) // 4 * 4
+    if padded_k_blocks == scale.shape[0]:
+        return scale
+
+    padded_scale = scale.new_full((scale.shape[1], padded_k_blocks), 1.0).transpose(
```
danielvegamyhre, line 33: note for next steps it would be useful to do this padding as part of the quantization kernel, rather than a separate padding step afterward, in order avoid an extra copy here.

danielvegamyhre (approved):
LGTM, please note for next steps we should do scale factor padding as part of the quantization kernel, rather than a separate kernel, to avoid the extra copy. Please also check on red CI jobs before landing

torchao/prototype/blockwise_fp8_training/linear.py
iamzainhuda: got it, i'll add that as a todo

iamzainhuda:
> > lso to note:
> > eager mode uses torch.nn.functional.scaled_mm
> > compile mode calls aten._scaled_mm_v2 directly because, in this torch build, Dynamo fullgraph cannot trace through the Python F.scaled_mm wrapper even though it lowers to the same underlying op
> 
> can you open a small reproducer on pytorch for this, this seems high priority to fix

created issue in pytorch/pytorch with small reproduce, filed as bug: https://github.com/pytorch/pytorch/issues/180604

PR #4237, opened 2026-04-03
add gptq algorithm diagram

vkuzo:
Summary:

Adding a diagram of the GPTQ update loop to simplify future work on this
code.

Also, var renames to simplify:
* W_t_quantize_block -> B_cur
* k_block_start -> B_cur_k_start
* k_block_end -> B_cur_k_end
* Err1 -> B_cur_Err1
* Hinv_quantize_block -> Hinv_cur
* k_group_start -> G_k_start
* k_group_end -> G_k_end

Test Plan: CI, GPTQ output did not change locally before+after


vkuzo:
* __->__ #4237

vkuzo:
@claude verify that all var renames are correct, that the diagram matches the code, and that this is a pure refactor (no logic change)

torchao/prototype/gptq/api.py
```diff
@@ -263,43 +263,85 @@ def gptq_quantize(H: torch.Tensor, W_t: torch.Tensor, config: GPTQConfig):
...
+    #    |-----------------------------------------|
+    #  N | 0 1 2 3 | 4 5 6 7 | ...
+    #    | ...
+    #    |-----------------------------------------|
+    #
+    # 1. start with W_t, with shape [N, K]
+    #    * B1, ..., BN are chunks of size [N, B], where B is a hyperparameter of GPTQ
+    #    * G1, ..., GN are chunks of size [N, G], where G is group_size of the quantization recipe
+    #
+    # 2. triple for loop, with every loop chunking along the K dimension:
```
jerryzh168, line 282: is this the same thing the code is doing? if so, would it better to just do inline comment for code (including refactor code itself to be easier to read by adding helper functions etc.)?

vkuzo:
I think the diagram is really useful. What you suggested is good as well, I
think this PR should land and we could implement your suggestion in a
future PR.

jerryzh168 (approved):

PR #4247, opened 2026-04-07
[nvfp4_training] Add Triton kernel for global amax of columnwise RHT (SM90+)

rdspring1:
## Summary                                                                                                                                                              
                                                                                                                                                                          
  - Adds `triton_rht_amax` (`hadamard_amax_triton.py`): a persistent, warp-specialized Triton kernel that applies the Randomized Hadamard Transform (RHT) to the input and reduces to a scalar global absolute maximum, without materializing the full post-RHT tensor                                                                            
  - This is a prerequisite building block for `triton_rht_quantize_row_col`: the global amax determines the per-tensor decode scale (`global_amax / (FP8_E4M3_MAX × FP4_E2M1_MAX)`) used in two-level NVFP4 quantization                                                                                                                    
  - Adds `_compute_pid` and related RHT matrix helpers to `hadamard_utils.py`                                                                             
                                                                                                                                                                          
## Key design choices                                                                                                                                                  
  - **Persistent grid**: kernel launches with `NUM_SMS` CTAs and each CTA iterates over all tiles, amortizing launch overhead
  - **Warp specialization**: producer warps issue TMA loads; consumer warps run `wgmma` for the RHT matrix multiply
  - **No output buffer**: per-CTA cumulative max is reduced with one `atomic_max` per CTA, avoiding a full `(N, M)` intermediate allocation

## Performance
| Model | Shape | M | N | time_us | gbps |
|---|---|---:|---:|---:|---:|
| Llama 3 8B | hidden-state input | 2048 | 4096 | 19.488 | 860.900 |
| Llama 3 8B | mlp.down input | 2048 | 14336 | 31.744 | 1849.810 |
| Llama 3 70B | hidden-state input | 2048 | 8192 | 25.600 | 1310.720 |
| Llama 3 70B | mlp.down input | 2048 | 28672 | 46.048 | 2550.390 |
                                                                                                                                                                          
## Test plan                                                                                                                                                            
  - [ ] `pytest test/prototype/mx_formats/test_hadamard_amax_triton.py -v`                                                                                                
  - [ ] Covered transitively by `test_hadamard_quantize_row_col_triton.py` — incorrect amax would break scale computation and fail SQNR/bitwise tests

### DIFF
--- a/benchmarks/prototype/nvfp4_training/README.md
+++ b/benchmarks/prototype/nvfp4_training/README.md
@@ -1,0 +1,55 @@
+# NVFP4 Training Benchmarks
+
+This directory contains benchmarking scripts for the NVFP4 training kernels
+under `torchao.prototype.mx_formats`.
+
+## Hadamard Amax Benchmark
+
+Benchmarks `triton_rht_amax` — the fused Randomized Hadamard Transform + amax
+reduction kernel used in NVFP4 training.
+
+```bash
+python -m benchmarks.prototype.nvfp4_training.bench_hadamard_amax
+```
+
+To run model-derived representative shapes:
+
+```bash
+python -m benchmarks.prototype.nvfp4_training.bench_hadamard_amax --shape-set representative-models
+```
+
+What it reports:
+
+- `time_us`: median kernel runtime in microseconds
+- `gbps`: effective memory bandwidth (input read bytes / time)
+
+### Methodology
+
+- Sweeps M ∈ {128, 256, 1024, 8192} × N ∈ {128, 256, 512, 1024, 2048, 4096, 8192, 16384, 32768}
+- Uses `benchmark_cuda_function_in_microseconds` from `benchmarks/utils.py`,
+  which wraps `triton.testing.do_bench` and returns the median.
+- Bandwidth is computed from input read bytes only (bfloat16 input, scalar output).
+
+### Representative Model Results
+
+The following shapes are activation-input matrices `(M, N)` for representative
+linear layers. `M` is `batch_size * sequence_length` except for the DeepSeek-V3
+routed expert rows, where `M` is the average per-expert token count:
+`4096 tokens * 8 experts per token / 256 routed experts = 128`.
+
+Run environment: NVIDIA GB200, PyTorch 2.12.0a0, Triton 3.7.0.
+
+| Model | Shape | M | N | time_us | gbps |
+|---|---|---:|---:|---:|---:|
+| DeepSeek-V3 671B | attn.wkv_b input | 4096 | 512 | 43.328 | 96.804 |
+| DeepSeek-V3 671B | attn.wq_b input | 4096 | 1536 | 46.048 | 273.256 |
+| DeepSeek-V3 671B | shared expert w2 input | 4096 | 2048 | 48.128 | 348.596 |
+| DeepSeek-V3 671B | hidden-state input | 4096 | 7168 | 45.088 | 1302.350 |
+| DeepSeek-V3 671B | attn.wo input | 4096 | 16384 | 64.512 | 2080.510 |
+| DeepSeek-V3 671B | dense ffn.w2 input | 4096 | 18432 | 68.672 | 2198.780 |
+| DeepSeek-V3 671B | avg routed expert w2 input | 128 | 2048 | 42.720 | 12.273 |
+| DeepSeek-V3 671B | avg routed expert w1/w3 input | 128 | 7168 | 42.240 | 43.442 |
+| Llama 3 8B | hidden-state input | 2048 | 4096 | 46.496 | 360.831 |
+| Llama 3 8B | mlp.down input | 2048 | 14336 | 46.080 | 1274.310 |
+| Llama 3 70B | hidden-state input | 2048 | 8192 | 44.032 | 762.047 |
+| Llama 3 70B | mlp.down input | 2048 | 28672 | 58.368 | 2012.070 |
--- a/test/prototype/mx_formats/test_hadamard_amax_triton.py
+++ b/test/prototype/mx_formats/test_hadamard_amax_triton.py
@@ -1,0 +1,97 @@
+"""Tests for triton_rht_amax (SM100+ kernel)."""
+
+import pytest
+import torch
+from torch.utils._triton import has_triton
+
+from torchao.prototype.mx_formats.hadamard_utils import (
+    get_hadamard_matrix,
+    get_rht_matrix,
+    get_wgrad_sign_vector,
+)
+from torchao.utils import is_sm_at_least_100
+
+_HARDCODED_SIGN_VECTOR = (
+    1,
+    1,
+    1,
+    -1,
+    1,
+    -1,
+    -1,
+    -1,
+    -1,
+    -1,
+    -1,
+    1,
+    -1,
+    1,
+    -1,
+    -1,
+)
+
+
+# M=32 excluded: all BLOCK_M configs (64, 128) exceed M=32 → all autotune configs fail.
+_M_VALUES = [64, 96, 128, 160, 256, 512]
+# N=100 excluded: TMA TensorDescriptor requires stride % 16 bytes == 0;
+# for bf16 this means N % 8 == 0. N=100 (100*2=200 bytes, 200%16=8) fails.
+_N_VALUES = [128, 200, 256, 384, 512, 1024]
+
+
+@torch.no_grad()
+def test_get_rht_matrix_with_hardcoded_sign_vector():
+    rht_matrix = get_rht_matrix(sign_vector=_HARDCODED_SIGN_VECTOR, device="cpu")
+
+    expected_signs = torch.tensor(_HARDCODED_SIGN_VECTOR, dtype=torch.bfloat16)
+    expected = torch.diag(expected_signs) @ get_hadamard_matrix(16, device="cpu")
+    torch.testing.assert_close(rht_matrix, expected, atol=0, rtol=0)
+
+
+@torch.no_grad()
+def test_get_rht_matrix_with_generated_sign_matches_sampled_signs():
+    get_rht_matrix.cache_clear()
+    torch.manual_seed(42)
+    expected_signs = get_wgrad_sign_vector(16, device="cpu")
+
+    get_rht_matrix.cache_clear()
+    torch.manual_seed(42)
+    rht_matrix = get_rht_matrix(sign_vector=None, device="cpu")
+
+    expected = torch.diag(expected_signs) @ get_hadamard_matrix(16, device="cpu")
+    torch.testing.assert_close(rht_matrix, expected, atol=0, rtol=0)
+
+
+@pytest.mark.parametrize(
+    "sign_vector",
+    [None, _HARDCODED_SIGN_VECTOR],
+    ids=["generated_sign_vector", "hardcoded_sign_vector"],
+)
+@pytest.mark.skipif(not has_triton(), reason="unsupported without triton")
+@pytest.mark.skipif(not is_sm_at_least_100(), reason="Requires SM100+")
+@pytest.mark.parametrize("N", _N_VALUES, ids=lambda n: f"N{n}")
+@pytest.mark.parametrize("M", _M_VALUES, ids=lambda m: f"M{m}")
+@torch.no_grad()
+def test_triton_rht_amax_vs_reference(M, N, sign_vector):
+    """triton_rht_amax must match the reference RHT matmul amax exactly (bitwise)."""
+    from torchao.prototype.mx_formats.hadamard_amax_triton import triton_rht_amax
+
+    torch.manual_seed(42)
+    A = torch.randn(M, N, dtype=torch.bfloat16, device="cuda")
+
+    get_rht_matrix.cache_clear()
+    if sign_vector is None:
+        torch.manual_seed(42)
+    B = get_rht_matrix(sign_vector=sign_vector, device="cuda")
+    ref_rht_amax = (
+        (A.t().reshape(N * M // 16, 16) @ B).to(torch.bfloat16).abs().max().float()
+    )
+    ref_amax = A.abs().max().float()
+
+    get_rht_matrix.cache_clear()
+    if sign_vector is None:
+        torch.manual_seed(42)
+
+    # Check RHT amax and regular amax are bitwise identical to reference.
+    triton_rht_amax_val, triton_amax_val = triton_rht_amax(A, sign_vector=sign_vector)
+    torch.testing.assert_close(triton_rht_amax_val, ref_rht_amax, atol=0, rtol=0)
+    torch.testing.assert_close(triton_amax_val, ref_amax, atol=0, rtol=0)
--- a/torchao/prototype/mx_formats/hadamard_amax_triton.py
+++ b/torchao/prototype/mx_formats/hadamard_amax_triton.py
@@ -1,0 +1,199 @@
+"""
+Triton kernel for Randomized Hadamard Transform (RHT) with fused global amax reduction.
+
+Entry point: triton_rht_amax(A) returns a scalar float32 global absolute maximum of
+the post-RHT output without materializing the full (N, M) output tensor. Uses a
+persistent warp-specialized TMA kernel with per-CTA cumulative max and one atomic_max
+per CTA into a caller-provided scalar buffer.
+"""
+
+import itertools
+
+import torch
+import torch.nn.functional as F
+import triton
+import triton.language as tl
+
+from torchao.prototype.mx_formats.hadamard_utils import _compute_pid, get_rht_matrix
+from torchao.utils import is_sm_at_least_100
+
+# SM100+ autotune configs. BLOCK_M must be divisible by 16 (RHT reshape constraint).
+HADAMARD_TILE_SHAPES: list[tuple[int, int]] = [
+    (64, 32),
+    (64, 64),
+    (64, 128),
+    (128, 32),
+    (128, 64),
+]
+
+HADAMARD_CONFIGS: list[triton.Config] = [
+    triton.Config(
+        {"BLOCK_M": bm, "BLOCK_N": bn, "NUM_STAGES": ns},
+        num_warps=nw,
+        num_stages=ns,
+    )
+    for (bm, bn), ns, nw in itertools.product(
+        HADAMARD_TILE_SHAPES,
+        [2, 3, 4],  # NUM_STAGES
+        [4, 8],  # NUM_WARPS
+    )
+]
+
+
+@triton.autotune(configs=HADAMARD_CONFIGS, key=["M", "N"])
+@triton.jit
+def _hadamard_amax_kernel(
+    a_ptr,
+    b_ptr,
+    global_rht_amax_ptr,
+    global_a_amax_ptr,
+    M,
+    N,
+    BLOCK_M: tl.constexpr,
+    BLOCK_N: tl.constexpr,
+    GROUP_SIZE_N: tl.constexpr,
+    NUM_SMS: tl.constexpr,
+    NUM_STAGES: tl.constexpr,
+):
+    """Persistent RHT kernel with fused amax reduction; no output tensor written."""
+    # Create TMA descriptors in-kernel from raw pointers and shape/stride
+    a_desc = tl.make_tensor_descriptor(
+        a_ptr,
+        shape=[M, N],
+        strides=[N, 1],
+        block_shape=[BLOCK_M, BLOCK_N],
+    )
+    b_desc = tl.make_tensor_descriptor(
+        b_ptr,
+        shape=[16, 16],
+        strides=[16, 1],
+        block_shape=[16, 16],
+    )
+
+    # Persistent grid-stride loop
+    start_pid = tl.program_id(0)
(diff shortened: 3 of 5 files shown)


torchao/prototype/mx_formats/hadamard_utils.py
```diff
@@ -0,0 +1,77 @@
...
+
+import math
+import functools
+import torch
+import triton
+import triton.language as tl
+
+
+def get_wgrad_sign_vector(device) -> torch.Tensor:
+    """Hard-coded random signs for Hadamard transform."""
```
slayton58: Wait, why is this hard-coded vs. generated?

torchao/prototype/mx_formats/hadamard_utils.py
```diff
@@ -0,0 +1,77 @@
...
+                [1, 1, 1, 1, 1, 1, 1, 1, -1, -1, -1, -1, -1, -1, -1, -1],
+                [1, -1, 1, -1, 1, -1, 1, -1, -1, 1, -1, 1, -1, 1, -1, 1],
+                [1, 1, -1, -1, 1, 1, -1, -1, -1, -1, 1, 1, -1, -1, 1, 1],
+                [1, -1, -1, 1, 1, -1, -1, 1, -1, 1, 1, -1, -1, 1, 1, -1],
+                [1, 1, 1, 1, -1, -1, -1, -1, -1, -1, -1, -1, 1, 1, 1, 1],
+                [1, -1, 1, -1, -1, 1, -1, 1, -1, 1, -1, 1, 1, -1, 1, -1],
+                [1, 1, -1, -1, -1, -1, 1, 1, -1, -1, 1, 1, 1, 1, -1, -1],
+                [1, -1, -1, 1, -1, 1, 1, -1, -1, 1, 1, -1, 1, -1, -1, 1],
+            ],
+            dtype=torch.float32,
```
slayton58: Is there a reason this needs to be 32b? Won't we save BW later if we store this natively in something lower (4b for instance), along with avoiding casts (See [here](https://github.com/pytorch/ao/pull/4247/changes#diff-eff9c1a9cce14be6ba193e088702759209bccddcc20b763d012288a5cd8833cfR157))

torchao/prototype/mx_formats/hadamard_utils.py
```diff
@@ -0,0 +1,77 @@
...
+            device=device,
+        )
+        * hadamard_scale
+    )
+
+
+@functools.lru_cache(maxsize=None)
+def get_rht_matrix(with_random_sign_mask: bool, device) -> torch.Tensor:
```
slayton58: Should have argument for hadamard dimension (kwarg-only perhaps, default=16?). Should probably also have an option dtype argument too (seeing notes above)

torchao/prototype/mx_formats/hadamard_amax_triton.py
```diff
@@ -0,0 +1,176 @@
...
+        # TritonGPUAutomaticWarpSpecialization MLIR pass
+        abs_a_t_rht = tl.abs(a_t_rht)
+        cumulative_max = tl.maximum(cumulative_max, abs_a_t_rht)
+
+    # Get scalar max for this block and update global max with atomic max operation
+    tile_max = tl.max(tl.max(cumulative_max, axis=1), axis=0)
+    tl.atomic_max(global_max_ptr, tile_max.to(tl.float32))
+
+
+def triton_rht_amax(
```
slayton58, line 127: No option for non-global amax domain

torchao/prototype/mx_formats/hadamard_amax_triton.py
```diff
@@ -0,0 +1,176 @@
...
+            N,
+            BLOCK_M=cfg.BLOCK_M,
+            BLOCK_N=cfg.BLOCK_N,
+            GROUP_SIZE_N=GROUP_SIZE_N,
+            NUM_SMS=NUM_SMS,
+            NUM_STAGES=cfg.NUM_STAGES,
+            num_warps=cfg.NUM_WARPS,
+        )
+
+    best = get_best_config(cache_key, HADAMARD_CONFIGS, benchmark_fn)
```
slayton58: Is there a reason you're re-implementing autotune?

test/prototype/mx_formats/test_hadamard_amax_triton.py
```diff
@@ -0,0 +1,34 @@
...
+@pytest.mark.parametrize("M", _M_VALUES, ids=lambda m: f"M{m}")
+@torch.no_grad()
+def test_triton_rht_amax_vs_reference(M, N):
+    """triton_rht_amax must match the reference RHT matmul amax exactly (bitwise)."""
+    torch.manual_seed(42)
+    A = torch.randn(M, N, dtype=torch.bfloat16, device="cuda")
+
+    # Reference: same deterministic matrix (lru_cached, hard-coded sign vector)
+    B = get_rht_matrix(with_random_sign_mask=True, device="cuda")
+    ref_amax = (A.t().reshape(N * M // 16, 16) @ B).to(torch.bfloat16).abs().max().float()
```
slayton58: This is (as written) doing `A (bf16) @ B (fp32)` - intended?

torchao/prototype/mx_formats/hadamard_amax_triton.py
```diff
@@ -0,0 +1,176 @@
...
+
+    Args:
+        A: (M, N) bfloat16 tensor, row-major. M must be divisible by 16.
+
+    Returns:
+        Scalar float32 tensor containing max(abs(RHT(A))).
+
+    Raises:
+        NotImplementedError: If hardware is pre-SM90.
+        AssertionError: If A is not bfloat16, not 2-D, not contiguous, or M % 16 != 0.
```
slayton58: Nit: `ValueError`

torchao/prototype/mx_formats/hadamard_utils.py
```diff
@@ -0,0 +1,77 @@
...
+    """Construct matrix used in random Hadamard transform."""
+    hadamard_dimension = 16
+    if with_random_sign_mask:
+        signs = get_wgrad_sign_vector(device=device)
+    else:
+        signs = torch.ones(1, dtype=torch.float32, device=device)
+    sign_matrix = signs * torch.eye(
```
slayton58: We should also be able to specify different sign vectors if desired.

torchao/prototype/mx_formats/autotune_configs.py
```diff
@@ -0,0 +1,101 @@
...
+        fn()
+        end.record()
+        torch.cuda.synchronize()
+        times.append(start.elapsed_time(end))
+
+    times.sort()
+    return times[len(times) // 2]
+
+
+def get_best_config(
```
danielvegamyhre: why is this necessary? triton autotuner automatically caches the best config for the given keys.

torchao/prototype/mx_formats/autotune_configs.py
```diff
@@ -0,0 +1,101 @@
...
+    BLOCK_N: int
+    NUM_STAGES: int
+    NUM_WARPS: int
+
+
+# Process-lifetime cache: maps (kernel_name, *shape_dims, device_index) -> KernelConfig
+_autotune_cache: dict[tuple, KernelConfig] = {}
+
+
+def do_bench(fn: Callable, warmup_iters: int = 3, bench_iters: int = 10) -> float:
```
danielvegamyhre: can you use our existing benchmark util for this instead of defining a new one: https://github.com/pytorch/ao/blob/707bee86677ce4ab7b699693e37635db025f1f49/benchmarks/utils.py#L107

danielvegamyhre (commented):
When adding new kernels could you add microbenchmarks in `ao/benchmarks/prototype`? Feel free to start a `nvfp4_training` directory in there

torchao/prototype/mx_formats/hadamard_utils.py
danielvegamyhre: we could add the kwarg with default 16 and for now just assert == 16, wdyt? only support 16 for now, build out prototype quickly and support generating other RHT matrix sizes later?

slayton58: Yep, that's perfectly fine - I'm concerned with getting the API as correct as possible here

test/prototype/mx_formats/test_hadamard_amax_triton.py
rdspring1: B from `get_rht_matrix` is `return rht_matrix.to(dtype=torch.bfloat16)`, so it is always `bf16`

torchao/prototype/mx_formats/hadamard_amax_triton.py
rdspring1: Added `scaling_type: F.ScalingType = F.ScalingType.TensorWise`. The function throw `ValueError` if it is anything except `F.ScalingType.TensorWise`

torchao/prototype/mx_formats/hadamard_utils.py
rdspring1: Converted `get_wgrad_sign_vector` to generate random sign vector.

torchao/prototype/mx_formats/hadamard_utils.py
rdspring1: Added `dtype` argument with `torch.bfloat16` default

torchao/prototype/mx_formats/hadamard_utils.py
rdspring1: Changes `get_rht_matrix` to
```python
def get_rht_matrix(
    sign_vector: tuple[int, ...] | None,
    device,
    dtype: torch.dtype = torch.bfloat16,
    hadamard_dimension: int = 16,
) -> torch.Tensor:
```
* If `sign_vector` is `None`, it calls `get_wgrad_sign_vector`. Otherwise the `sign_vector` tuple is converted to torch.tensor. It is a tuple so it is hashable for `lru_cache`.

torchao/prototype/mx_formats/hadamard_utils.py
rdspring1: Added `sign_vector` argument.

torchao/prototype/mx_formats/hadamard_amax_triton.py
rdspring1: Removed.

torchao/prototype/mx_formats/autotune_configs.py
rdspring1: Added micro benchmark for the triton kernel.

torchao/prototype/mx_formats/autotune_configs.py
rdspring1: Removed. I replace host-side TMA descriptor with in-kernel ones so it is compatible with `@triton.autotune` decorator.

benchmarks/prototype/nvfp4_training/README.md
```diff
@@ -0,0 +1,25 @@
+# NVFP4 Training Benchmarks
```
danielvegamyhre, line 1: Can you add the benchmark results on some representative shapes (i.e., DSV3 671b with seq_len=4096, microbatch_size of 1 sequence)? In the PR description and/or in this MD file

torchao/prototype/mx_formats/hadamard_amax_triton.py
```diff
@@ -0,0 +1,189 @@
...
+import itertools
+import triton
+import triton.language as tl
+import torch
+import torch.nn.functional as F
+from torchao.prototype.mx_formats.hadamard_utils import get_rht_matrix, _compute_pid
+from torchao.utils import is_sm_at_least_90
+
+# SM90+ autotune configs. BLOCK_M must be divisible by 16 (RHT reshape constraint).
+HADAMARD_CONFIGS: list[triton.Config] = [
```
slayton58, line 29: How long does autotune take to run for each config? I don't think I'm too concerned, but it's worth knowing and keeping an eye on

torchao/prototype/mx_formats/hadamard_utils.py
```diff
@@ -0,0 +1,88 @@
...
+        torch.ones(shape, dtype=dtype, device=device),
+        -torch.ones(shape, dtype=dtype, device=device),
+    )
+
+
+def get_hadamard_matrix(
+    hadamard_dimension: int, device, dtype: torch.dtype = torch.bfloat16
+) -> torch.Tensor:
+    """Construct a 16x16 Hadamard matrix (scaled by 1/sqrt(16))."""
+    assert hadamard_dimension == 16, "Only hadamard dimension 16 is supported."
```
slayton58: Nit: `assert` to:

```
if hadamard_dimension != 16:
   raise ValueError(...)
```
? 

Core has moved more to `raise` vs. `assert` - I'll let @danielvegamyhre decide if we care enough to want to fix.

torchao/prototype/mx_formats/hadamard_utils.py
```diff
@@ -0,0 +1,88 @@
...
+    sign_vector: tuple[int, ...] | None,
+    device,
+    dtype: torch.dtype = torch.bfloat16,
+    hadamard_dimension: int = 16,
+) -> torch.Tensor:
+    """Construct an RHT matrix from an explicit sign vector or a generated sign vector."""
+    if sign_vector is None:
+        signs = get_wgrad_sign_vector(hadamard_dimension, device=device, dtype=dtype)
+    else:
+        assert len(sign_vector) == hadamard_dimension, (
```
slayton58: Nit: `assert` vs. `raise` again.

torchao/prototype/mx_formats/hadamard_utils.py
```diff
@@ -0,0 +1,88 @@
...
+        )
+        signs = torch.tensor(sign_vector, dtype=dtype, device=device)
+    sign_matrix = signs * torch.eye(hadamard_dimension, dtype=dtype, device=device)
+    return sign_matrix @ get_hadamard_matrix(
+        hadamard_dimension, device=device, dtype=dtype
+    )
+
+
+@triton.jit
+def _compute_pid(tile_id, num_pid_in_group, num_pid_n, GROUP_SIZE_N: tl.constexpr):
```
slayton58: Is this stable? In that does triton always layout threads this way?

danielvegamyhre:
@rdspring1 please address final comments and make sure CI is green, thanks!

benchmarks/prototype/nvfp4_training/README.md
rdspring1: I added results for DSV3 and Llama3 8B/70B.

| Model | Shape | M | N | time_us | gbps |
|---|---|---:|---:|---:|---:|
| DeepSeek-V3 671B | attn.wkv_b input | 4096 | 512 | 43.328 | 96.804 |
| DeepSeek-V3 671B | attn.wq_b input | 4096 | 1536 | 46.048 | 273.256 |
| DeepSeek-V3 671B | shared expert w2 input | 4096 | 2048 | 48.128 | 348.596 |
| DeepSeek-V3 671B | hidden-state input | 4096 | 7168 | 45.088 | 1302.350 |
| DeepSeek-V3 671B | attn.wo input | 4096 | 16384 | 64.512 | 2080.510 |
| DeepSeek-V3 671B | dense ffn.w2 input | 4096 | 18432 | 68.672 | 2198.780 |
| DeepSeek-V3 671B | avg routed expert w2 input | 128 | 2048 | 42.720 | 12.273 |
| DeepSeek-V3 671B | avg routed expert w1/w3 input | 128 | 7168 | 42.240 | 43.442 |
| Llama 3 8B | hidden-state input | 2048 | 4096 | 46.496 | 360.831 |
| Llama 3 8B | mlp.down input | 2048 | 14336 | 46.080 | 1274.310 |
| Llama 3 70B | hidden-state input | 2048 | 8192 | 44.032 | 762.047 |
| Llama 3 70B | mlp.down input | 2048 | 28672 | 58.368 | 2012.070 |

torchao/prototype/mx_formats/hadamard_utils.py
rdspring1: changed to `raise ValueError`

torchao/prototype/mx_formats/hadamard_amax_triton.py
rdspring1: I estimated autotune cost using `TRITON_PRINT_AUTOTUNING=1` on `triton_rht_amax`, with an isolated `TRITON_CACHE_DIR` to control cache effects. I got the total autotuning time for one shape (4096, 7168) and then divided by number of configs (24) to estimate mean time per config.

Results on GB200:
  - Cold Triton cache: `11.53-11.60s` total autotune time, about `0.48s/config`.
  - Warm Triton compile cache, fresh process: `3.54s` total autotune time, about `0.15s/config`.
  - Steady-state benchmark runtime for the selected config was around `39-44 us`.

torchao/prototype/mx_formats/hadamard_utils.py
rdspring1: This helper does not assume any Triton-specific thread layout guarantee. It only converts a linear logical index into a grouped 2D traversal order. As with CUDA thread blocks, actual execution order is not something we should rely on. At best, this ordering is a hint for better locality.

danielvegamyhre: my understanding is it is basically a simpler version of a Hilbert Curve (i.e., a cache aware tile processing order) - is that right?

rdspring1:
@danielvegamyhre I addressed the last round of reviewer comments and the CI looks green. The
`test/prototype/mx_formats/test_mx_tensor.py::test_cast_to_float8_e4m3fn_saturation_behavior` failure isn't related to this PR.

danielvegamyhre (approved):

rdspring1:
@danielvegamyhre Can you merge this PR? I added performance numbers to the summary. Thanks!

danielvegamyhre:
@rdspring1 i am confused by the benchmarks, what hardware is this on, the numbers look extremely low for b200/gb200?

also, for these DSV3 shapes:

- `avg routed expert w2 input`
- `avg routed expert w1/w3 input`

why are you dividing the M dimension up into only a 128 row chunk? wouldn't the kernel operate on the full tensor containing input activations for all local experts?

rdspring1:
@danielvegamyhre 
### Changes to benchmark:
The benchmark was setup to measure `triton_rht_amax` from cold-start. I changed the script to measure only the triton kernel itself.
  - RHT matrix is precomputed before do_bench
  - scalar output buffers are preallocated before do_bench
  - Triton allocator setup is outside the timed closure
  - the timed closure only launches _hadamard_amax_kernel
  - Mostly changed results for smaller kernels.

### Updated benchmark results:
NVIDIA GB200 with ~8 TB/s memory bandwidth.

**Copy kernel baseline:**
| Shape | M | N | copy_us | copy read-equivalent GB/s |
|---|---:|---:|---:|---:|
| Llama 3 8B hidden-state input | 2048 | 4096 | 12.128 | 1383.3 |
| Llama 3 8B mlp.down input | 2048 | 14336 | 24.576 | 2389.3 |
| Llama 3 70B hidden-state input | 2048 | 8192 | 16.384 | 2048.0 |
| Llama 3 70B mlp.down input | 2048 | 28672 | 42.848 | 2740.9 |

**`triton_rht_amax` results:**
| Model | Shape | M | N | time_us | gbps |
|---|---|---:|---:|---:|---:|
| Llama 3 8B | hidden-state input | 2048 | 4096 | 19.488 | 860.900 |
| Llama 3 8B | mlp.down input | 2048 | 14336 | 31.744 | 1849.810 |
| Llama 3 70B | hidden-state input | 2048 | 8192 | 25.600 | 1310.720 |
| Llama 3 70B | mlp.down input | 2048 | 28672 | 46.048 | 2550.390 |

* `triton_rht_amax` reads input tensor and writes two scalars, so I compared read-equivalent GB/s. 

| Shape | Current GB/s | Vs local copy read-equivalent | Vs ~8 TB/s single-GPU peak |
|---|---:|---:|---:|
| Llama 3 8B hidden | 860.9 | ~62% | ~11% |
| Llama 3 8B mlp.down | 1849.8 | ~77% | ~23% |
| Llama 3 70B hidden | 1310.7 | ~64% | ~16% |
| Llama 3 70B mlp.down | 2550.4 | ~93% | ~32% |

### DSV3:
This triton_rht_amax kernel doesn't handle MOE layers natively. **I went ahead and removed those shapes.**
For a fallback, you'd padded to nearest 128 and quantize each tensor separately. This is what earlier benchmarks were measuring.
```
For each expert:
  row_global_amax, col_global_amax = triton_rht_amax(expert_tokens_padded_to_128)
```

danielvegamyhre:
>This triton_rht_amax kernel doesn't handle MOE layers natively. I went ahead and removed those shapes.
For a fallback, you'd padded to nearest 128 and quantize each tensor separately. This is what earlier benchmarks were measuring.

Ok makes sense, yeah let's just focus on linear layers for now. MoEs are a separate problem which I am prototyping some things internally for atm.

Thanks for benchmarking against memcpy for the given shapes, provides better context than hardware theoretical max bw in these cases.

danielvegamyhre (approved):

rdspring1:
@danielvegamyhre can you merge this PR? The H100 failure looks unrelated; something about `mslk.so: undefined symbol`.

PR #4251, opened 2026-04-08
[xpu][mx] Enable mx matmul tests on xpu

ugolowic:

To enable matmul tests on xpu, the "new" PyTorch matmul API needs to be used with torch 2.10+.

The older `torch._scaled_mm` infers scaling type and memory layout from tensor shapes alone. This only works on CUDA with the specific swizzled scale layout the hardware expects. 

PyTorch 2.10 introduced `torch.functional.scaled_mm` as the public API. It takes explicit `ScalingType` and `SwizzleType` arguments, making it possible to specify non-swizzled layouts needed by backends like XPU.

### DIFF
--- a/test/prototype/mx_formats/test_mx_mm.py
+++ b/test/prototype/mx_formats/test_mx_mm.py
@@ -13,37 +13,20 @@ from torch.nn.functional import ScalingType, SwizzleType
 from torchao.float8.float8_utils import compute_error
 from torchao.prototype.mx_formats.mx_tensor import MXTensor
 from torchao.prototype.mx_formats.utils import to_blocked
-from torchao.utils import is_sm_at_least_100
-
-
-def _mxfp4_scaled_mm(a_data, b_data, a_scale_block, b_scale_block):
-    """Wrapper for F.scaled_mm with MXFP4 configuration."""
-    return F.scaled_mm(
-        a_data.view(torch.float4_e2m1fn_x2),
-        b_data.view(torch.float4_e2m1fn_x2),
-        scale_a=a_scale_block,
-        scale_recipe_a=ScalingType.BlockWise1x32,
-        scale_b=b_scale_block,
-        scale_recipe_b=ScalingType.BlockWise1x32,
-        swizzle_a=SwizzleType.SWIZZLE_32_4_4,
-        swizzle_b=SwizzleType.SWIZZLE_32_4_4,
-        output_dtype=torch.bfloat16,
-    )
+from torchao.utils import (
+    is_sm_at_least_100,
+    torch_version_at_least,
+)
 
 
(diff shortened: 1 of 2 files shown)


test/prototype/mx_formats/test_mx_mm.py
```diff
@@ -69,20 +45,55 @@ def run_matrix_test(M: int, K: int, N: int, format) -> float:
...
-    not is_sm_at_least_100(), reason="CUDA capability >= 10.0 required for mxfloat8"
+    not torch.accelerator.is_available(), reason="No accelerator available"
+)
+@pytest.mark.skipif(
+    torch.cuda.is_available() and not is_sm_at_least_100(),
+    reason="CUDA capability >= 10.0 required for mxfloat8",
+)
+@pytest.mark.skipif(
+    torch.xpu.is_available() and not torch_version_at_least("2.13.0"),
+    reason="XPU blockwise scaled_mm requires oneDNN 3.11+ (PyTorch 2.13+) (pytorch/pytorch#173630, #176043, #176064)",
```
xiaowangintel: The reason does not match above check. There is no need to specifically require oneDNN 3.11+ here.

torchao/prototype/mx_formats/mx_tensor.py
```diff
@@ -822,8 +842,12 @@ def _addmm_mx_dispatch(
                 scale_recipe_a=ScalingType.BlockWise1x32,
                 scale_b=b_scale_block,
                 scale_recipe_b=ScalingType.BlockWise1x32,
-                swizzle_a=SwizzleType.SWIZZLE_32_4_4,
-                swizzle_b=SwizzleType.SWIZZLE_32_4_4,
+                swizzle_a=SwizzleType.SWIZZLE_32_4_4
```
xiaowangintel: swizzle_a and swizzle_b are identical, so this can be simplified here.

xiaowangintel:
@claude review this PR.

test/prototype/mx_formats/test_mx_mm.py
```diff
@@ -22,38 +22,14 @@
...
-        swizzle_a=SwizzleType.SWIZZLE_32_4_4,
-        swizzle_b=SwizzleType.SWIZZLE_32_4_4,
-        output_dtype=torch.bfloat16,
-    )
-
-
 def run_matrix_test(M: int, K: int, N: int, format) -> float:
     dtype = torch.bfloat16
-    device = torch.device("cuda")
+    device = torch.accelerator.current_accelerator()
```
xiaowangintel, line 24: 1. No Hardcoded Device Strings
2. Use @parametrize for Device Parameterization
All device-dependent tests should be decorated with @parametrize("device", get_available_devices()) and accept device as a function argument.

PR #4254, opened 2026-04-09
Support 32x32 scaling for weights in MXFP8 weight quantization kernel

### LINKED ISSUE #4185: Support 32x32 scaling for weights in MXFP8 weight quantization kernel
32x32 scaling is more performant by avoiding an extra scale calculation in the backward pass, and it's also better for accuracy by ensuring that there's no params getting underflowed in forward (and not contributing to output) but not getting underflowed in backward (and thus getting a gradient).

alexsamardzic:
Fixes #4185

### DIFF
--- a/test/prototype/moe_training/test_mxfp8_grouped_mm.py
+++ b/test/prototype/moe_training/test_mxfp8_grouped_mm.py
@@ -30,6 +30,10 @@ if not (
 pytest.importorskip("triton", reason="Triton required to run this test")
 
 from torchao.float8.float8_utils import compute_error
+from torchao.prototype.moe_training.kernels.mxfp8 import (
+    mx_block_rearrange_2d_M_groups_cuda,
+    mxfp8_quantize_cuda_3d,
+)
 from torchao.prototype.moe_training.mxfp8_grouped_mm import (
     _emulated_mxfp8_scaled_grouped_mm_2d_2d,
     _emulated_mxfp8_scaled_grouped_mm_2d_3d,
@@ -40,6 +44,7 @@ from torchao.prototype.moe_training.utils import (
     _to_mxfp8_per_group_rowwise,
     generate_jagged_offs,
 )
+from torchao.prototype.mx_formats.kernels import triton_to_mxfp8_dim0
 from torchao.prototype.mx_formats.mx_tensor import MXTensor, to_mx
 from torchao.quantization.quantize_.common import KernelPreference
 from torchao.testing.utils import skip_if_rocm
@@ -49,30 +54,134 @@ torch._dynamo.config.cache_size_limit = 1000
 
 
 @skip_if_rocm("ROCm not supported")
+@pytest.mark.skipif(
+    not is_sm_version(10, 0),
+    reason="3D MXFP8 quantization requires SM100",
+)
 @pytest.mark.parametrize("M,K,N", [(1024, 1024, 1024), (1024, 2048, 4096)])
 @pytest.mark.parametrize("num_experts", (1, 8, 16))
-def test_emulate_mxfp8_grouped_gemm_2d_3d(M, K, N, num_experts):
+@pytest.mark.parametrize("scale_block_k", (1, 32), ids=("32x1_n", "32x32_n"))
+@pytest.mark.parametrize(
+    "scale_mode", (ScaleCalculationMode.FLOOR, ScaleCalculationMode.RCEIL)
+)
+def test_emulate_mxfp8_grouped_gemm_2d_3d(
+    M, K, N, num_experts, scale_block_k, scale_mode
+):
     x = torch.randn(M, K, dtype=torch.bfloat16, device="cuda")
     w = torch.randn(num_experts, N, K, dtype=torch.bfloat16, device="cuda")
     offs = generate_jagged_offs(num_experts, M)
-    x_ref, w_ref, offs_ref = x.clone(), w.clone(), offs.clone()
+    offs_ref = offs.clone()
 
     # Quantize inputs to mxpf8 for emulated mxfp8 scaled grouped mm
     block_size = 32
-    x_scale, x_fp8 = to_mx(x, elem_dtype=torch.float8_e4m3fn, block_size=block_size)
-
-    # To cast B_t per-expert to mxfp8 across dim1, we transpose the experts, cast along dim -1, then untranspose.
-    w_scale, w_fp8 = to_mx(
-        w,
+    x_scale, x_fp8 = to_mx(
+        x,
         elem_dtype=torch.float8_e4m3fn,
         block_size=block_size,
+        scaling_mode=scale_mode,
     )
+    x_ref = (
+        x_fp8.reshape(M, K // block_size, block_size).to(torch.bfloat16)
+        * x_scale.unsqueeze(-1).to(torch.bfloat16)
+    ).reshape(M, K)
+
+    if scale_block_k == 1:
+        w_scale, w_fp8 = to_mx(
+            w,
+            elem_dtype=torch.float8_e4m3fn,
+            block_size=block_size,
+            scaling_mode=scale_mode,
+        )
+    else:
+        w_fp8, w_scale = mxfp8_quantize_cuda_3d(
+            w,
+            block_size=block_size,
+            scale_block_n=block_size,
+            scale_block_k=scale_block_k,
+            scaling_mode=scale_mode.value.lower(),
+            blocked_scale_output=False,
+        )
+        w_scale = w_scale.repeat_interleave(block_size, dim=1)
+    w_ref = (
+        w_fp8.reshape(num_experts, N, K // block_size, block_size).to(torch.bfloat16)
+        * w_scale.unsqueeze(-1).to(torch.bfloat16)
+    ).reshape(num_experts, N, K)
 
     ref_out = torch._grouped_mm(
         x_ref, w_ref.transpose(-2, -1), offs=offs_ref, out_dtype=torch.bfloat16
     )
     out = _emulated_mxfp8_scaled_grouped_mm_2d_3d(
         x_fp8, x_scale, w_fp8, w_scale, offs=offs, out_dtype=torch.bfloat16
+    )
+
+    sqnr = compute_error(ref_out, out)
+    min_sqnr = 27.0
+    assert sqnr >= min_sqnr, f"sqnr {sqnr} is too low, must be >= {min_sqnr}"
+
+
+@skip_if_rocm("ROCm not supported")
+@pytest.mark.skipif(
+    not is_sm_version(10, 0),
+    reason="3D MXFP8 quantization and MXFP8 grouped GEMM require SM100",
+)
+@pytest.mark.parametrize("M,K,N", [(1024, 1024, 1024), (1024, 2048, 4096)])
+@pytest.mark.parametrize("num_experts", (1, 8, 16))
+@pytest.mark.parametrize("scale_block_k", (1, 32), ids=("32x1_n", "32x32_n"))
+@pytest.mark.parametrize(
+    "scale_mode", (ScaleCalculationMode.FLOOR, ScaleCalculationMode.RCEIL)
+)
+def test_mxfp8_grouped_gemm_2d_3d(M, K, N, num_experts, scale_block_k, scale_mode):
+    grad_out = torch.randn(M, N, dtype=torch.bfloat16, device="cuda")
+    w = torch.randn(num_experts, N, K, dtype=torch.bfloat16, device="cuda")
+    offs = generate_jagged_offs(num_experts, M)
+    offs_ref = offs.clone()
+
+    # Real SM100 grouped MM: 1x32-scaled A @ 3D-quantized B -> grad_input.
+    block_size = 32
+    grad_out_fp8, grad_out_scale = triton_to_mxfp8_dim0(
+        grad_out,
+        inner_block_size=block_size,
+        scaling_mode=scale_mode.value.lower(),
+    )
+    w_fp8, w_scale = mxfp8_quantize_cuda_3d(
+        w,
+        block_size=block_size,
+        scale_block_n=block_size,
+        scale_block_k=scale_block_k,
+        scaling_mode=scale_mode.value.lower(),
+        blocked_scale_output=True,
+    )
+    w_fp8_ref, w_scale_ref = mxfp8_quantize_cuda_3d(
+        w,
+        block_size=block_size,
+        scale_block_n=block_size,
+        scale_block_k=scale_block_k,
+        scaling_mode=scale_mode.value.lower(),
+        blocked_scale_output=False,
+    )
+
+    grad_out_scale_ref = grad_out_scale.repeat_interleave(block_size, dim=1)
+    grad_out_ref = grad_out_fp8.to(torch.bfloat16) * grad_out_scale_ref.to(
+        torch.bfloat16
+    )
+    if scale_block_k == 1:
+        w_scale_ref = w_scale_ref.repeat_interleave(block_size, dim=1)
+    else:
+        w_scale_ref = w_scale_ref.repeat_interleave(
+            block_size, dim=1
+        ).repeat_interleave(block_size, dim=2)
+    w_ref = w_fp8_ref.to(torch.bfloat16) * w_scale_ref.to(torch.bfloat16)
+    ref_out = torch._grouped_mm(
+        grad_out_ref, w_ref, offs=offs_ref, out_dtype=torch.bfloat16
+    )
+    A_scales_blocked = mx_block_rearrange_2d_M_groups_cuda(grad_out_scale, offs)
+    out = torch._scaled_grouped_mm(
+        grad_out_fp8,
+        w_fp8,
+        A_scales_blocked,
+        w_scale,
+        offs=offs,
+        out_dtype=torch.bfloat16,
     )
 
     sqnr = compute_error(ref_out, out)
--- a/torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
+++ b/torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
@@ -5,6 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 import functools
+import math
 from typing import Tuple
 
 import torch
@@ -35,20 +36,18 @@ def _make_tile_smem_layouts(cute, tile_n: int, tile_k: int):
 # Config format:
 # (compute_warps, tile_n, tile_k, k_tiles_per_cta)
 _CUTEDSL_CONFIGS = {
-    "bf16_default": (6, 32, 128, 4),
+    "bf16_32x1": (4, 32, 128, 4),
+    "bf16_32x32": (4, 32, 128, 4),
     "fallback": (6, 32, 128, 2),
 }
 
 
 def _select_cutedsl_config(
     input_dtype_name: str,
-    scaling_mode: str,
-    K: int,
+    scale_block_k: int,
 ) -> Tuple[str, Tuple[int, int, int, int]]:
-    del K
-
     if input_dtype_name == "torch.bfloat16":
-        config_name = "bf16_default"
+        config_name = "bf16_32x32" if scale_block_k == 32 else "bf16_32x1"
     else:
         config_name = "fallback"
     return config_name, _CUTEDSL_CONFIGS[config_name]
@@ -64,6 +63,8 @@ def _compile_mxfp8_quantize_3d_cutedsl(
     requested_stage_count: int,
     k_tiles_per_cta: int,
     is_full_k_tiles: bool,
+    scale_block_n: int,
+    scale_block_k: int,
     blocked_scale_output: bool,
 ):
     import cuda.bindings.driver as cuda
@@ -101,6 +102,8 @@ def _compile_mxfp8_quantize_3d_cutedsl(
     TILE_K = tile_k
     K_TILES_PER_CTA = k_tiles_per_cta
     IS_FULL_K_TILES_VALUE = is_full_k_tiles
+    SCALE_DIM_N_VALUE = scale_block_n
+    SCALE_DIM_K_VALUE = scale_block_k
     BLOCKED_SCALE_OUTPUT_VALUE = blocked_scale_output
 
     THREADS_PER_BLOCK = (1 + COMPUTE_WARPS) * 32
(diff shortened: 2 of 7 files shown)


alexsamardzic:
Fixes #4185.

To test:

```
pytest -q test/prototype/moe_training/test_kernels.py -k "test_cuda_mx_3d_cutedsl_numerics and 32x32"
```

To benchmark:

```
python benchmarks/prototype/moe_training/mxfp8/bench_quantize_3d.py
```

Benchmarking results:

```
input_shape       scaling_mode                  scale_block_k    cuda_2d_us    cutedsl_3d_us    to_mx_us    cuda_2d_gbps    cutedsl_3d_gbps    to_mx_gbps
----------------  --------------------------  ---------------  ------------  ---------------  ----------  --------------  -----------------  ------------
(1, 8192, 5120)   ScaleCalculationMode.FLOOR                1        36.864           29.728     119.728        3448.89             4276.77      1061.91
(1, 8192, 5120)   ScaleCalculationMode.FLOOR               32       nan               30.752      73.664         nan                4134.36      1725.94
(1, 8192, 5120)   ScaleCalculationMode.RCEIL                1        35.872           28.704     118.752        3544.26             4429.34      1070.63
(1, 8192, 5120)   ScaleCalculationMode.RCEIL               32       nan               28.704      72.704         nan                4429.34      1748.73
(1, 7168, 2048)   ScaleCalculationMode.FLOOR                1        38.304           24.64       83.872        1161.73             1805.96       530.558
(1, 7168, 2048)   ScaleCalculationMode.FLOOR               32       nan               20.096      70.72          nan                2214.32       629.227
(1, 7168, 2048)   ScaleCalculationMode.RCEIL                1        33.984           18.624      83.968        1309.41             2389.33       529.951
(1, 7168, 2048)   ScaleCalculationMode.RCEIL               32       nan               19.808      69.824         nan                2246.51       637.302
(8, 8192, 5120)   ScaleCalculationMode.FLOOR                1      1524.64           168.864    1689.5           667.121            6023.3        602.022
(8, 8192, 5120)   ScaleCalculationMode.FLOOR               32       nan              168        1816.64          nan                6054.28       559.89
(8, 8192, 5120)   ScaleCalculationMode.RCEIL                1      1505.44           163.68     1688.42          675.629            6214.07       602.41
(8, 8192, 5120)   ScaleCalculationMode.RCEIL               32       nan              162.976    1817.6           nan                6240.91       559.594
(8, 7168, 2048)   ScaleCalculationMode.FLOOR                1       541.536           66.56      602.08          657.374            5348.43       591.27
(8, 7168, 2048)   ScaleCalculationMode.FLOOR               32       nan               65.568     644             nan                5429.35       552.782
(8, 7168, 2048)   ScaleCalculationMode.RCEIL                1       535.488           65.376     602.112         664.798            5445.29       591.238
(8, 7168, 2048)   ScaleCalculationMode.RCEIL               32       nan               64.544     644.032         nan                5515.49       552.754
(32, 7168, 2048)  ScaleCalculationMode.FLOOR                1      2128.03           228.512    2358.21          669.147            6231.47       603.834
(32, 7168, 2048)  ScaleCalculationMode.FLOOR               32       nan              229.312    2544.69          nan                6209.73       559.584
(32, 7168, 2048)  ScaleCalculationMode.RCEIL                1      2101.38           220.144    2358.29          677.635            6468.34       603.814
(32, 7168, 2048)  ScaleCalculationMode.RCEIL               32       nan              220.192    2535.33          nan                6466.93       561.65
(32, 8192, 5120)  ScaleCalculationMode.FLOOR                1      6057.66           651.264    6706.26          671.624            6247.04       606.669
(32, 8192, 5120)  ScaleCalculationMode.FLOOR               32       nan              651.536    7228.51          nan                6244.44       562.837
(32, 8192, 5120)  ScaleCalculationMode.RCEIL                1      5978.61           623.824    6707.81          680.505            6521.83       606.528
(32, 8192, 5120)  ScaleCalculationMode.RCEIL               32       nan              624.304    7229.57          nan                6516.82       562.755
```

To test with grouped MM, apply this [patch](https://github.com/user-attachments/files/26612478/diff.txt), and run (tolerances updated as 32x32 gives "worse" quantization quality than 32x1):

```
pytest -q test/prototype/moe_training/test_mxfp8_grouped_mm.py -k test_mxfp8_grouped_gemm_with_dq_fwd_bwd
```

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
```diff
@@ -187,6 +192,92 @@ def _store_scale(
...
+                scales_expert[n_row, k_block] = scale_u8
+            else:
+                scales_expert[e, n_block, k_block] = scale_u8
+
+        @cute.jit
+        def _warp_reduce_max(
+            self,
+            amax: cutlass.Float32,
+        ):
+            amax = cute.arch.fmax(
```
danielvegamyhre: using warp reduction for this sense - one nit but i think this is better done via a for loop, see example from quack:

```python
@cute.jit
def warp_reduce(val: cute.Numeric,
                op: Callable,
                width: cutlass.Constexpr = 32) -> cute.Numeric:
    for i in range(int(math.log2(width))):
        # cute.arch.shuffle_sync_bfly will read from another thread's registers
        val = op(val, cute.arch.shuffle_sync_bfly(val, offset=1 << i))
    return val
```

torchao/prototype/moe_training/kernels/mxfp8/quant.py
```diff
@@ -1282,16 +1379,50 @@ def fused_unpad_token_groups_cuda(
 def mxfp8_quantize_cuda_3d(
     x: torch.Tensor,
     block_size: int = 32,
+    scale_block_k: int = 1,
```
danielvegamyhre: i think this API is a bit confusing - the function is just called "mxfp8_quantize_cuda_3d" and combined with the "scale_block_k" param (without corresponding scale_block_n param) seems to communicate that it is a kernel for scaling across K.

I think we should update this to be:

```python
def mxfp8_quantize_3d_cutedsl(
   x: torch.Tensor,
   scale_block_n: Literal[1, 32],
   scale_block_k: Literal[1, 32],
) ...
   # .. (validate only 32x1 and 32x32 are currently supported)
   # .. (in the future would be nice to support 1x32 as well)
```

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
```diff
@@ -598,16 +693,31 @@ def __call__(
...
+                    blocked_scale_layout = cute.make_layout(
+                        ((32, 4, scale_row_tiles), (4, scale_col_tiles)),
+                        stride=(
+                            (16, 4, cutlass.Int64(128) * padded_scale_cols),
+                            (1, cutlass.Int64(512)),
+                        ),
+                    )
+                else:
+                    padded_scale_cols = cute.round_up(n_blocks, 4)
+                    k_block_tiles = cute.ceil_div(K, 128)
```
danielvegamyhre: i just noticed, shouldn't we always be padding the scale tensor rows/cols such that is evenly divisible into 128x4 cols (or 4x128, depending on scaling dim)? for both 32x1 scaling and 32x32 scaling? 

this seems to assume that in the case of 32x1 scaling, we assume K is divisible by 128, but do we actually validate that, or will it just silently fail/crash?

i would prefer we don't do if/else here, and just unconditionally use padded rows/cols.

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
alexsamardzic: Updated.

torchao/prototype/moe_training/kernels/mxfp8/quant.py
alexsamardzic: Changed.  (Probably 1x32 will not be a straightforward extension like 32x32.)

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
alexsamardzic: Fixed.

torchao/prototype/moe_training/kernels/mxfp8/quant.py
danielvegamyhre: Yeah I agree, based on my experience with the kernels for 1x32 vs 32x1 scaling on 2D tensors.

This can just be the user-facing API, that dispatches to the appropriate kernel based on the scale_k/scale_n values

torchao/prototype/moe_training/mxfp8_grouped_mm.py
```diff
@@ -663,6 +663,7 @@ def __compute_dgrad_sm100(
     weight_e4m3, weight_scales_blocked = mxfp8_quantize_cuda_3d(
         weight._data if hasattr(weight, "_data") else weight,
         block_size,
+        scale_block_n=block_size,
```
danielvegamyhre, line 666: can you make this more explicit by also specifying `scale_block_k=1`? so it is clear at a glance we are still using 32x1 scaling here, not 32x32 (yet).

alexsamardzic: Fixed.

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
```diff
@@ -698,6 +791,7 @@ def __call__(
 def mxfp8_quantize_cutedsl_3d(
     x: torch.Tensor,
     block_size: int = 32,
+    scale_block_k: int = 1,
```
danielvegamyhre, line 824: can we make everything consistent with the user-facing API where `scale_block_n` and `scale_block_k` are both explicit, rather than N being implicit and K being explicit? i.e. in the kernel as well

alexsamardzic: Sure, fixed - sorry about that.  For the kernel, as discussed above, considerable changes would be probably needed for `scale_block_n` different than 32, but it still make sense to have it explicit.

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
```diff
@@ -35,20 +36,24 @@ def _make_tile_smem_layouts(cute, tile_n: int, tile_k: int):
...
 
 
 def _select_cutedsl_config(
     input_dtype_name: str,
     scaling_mode: str,
+    scale_block_n: int,
     K: int,
+    scale_block_k: int,
 ) -> Tuple[str, Tuple[int, int, int, int]]:
+    del scale_block_n
```
danielvegamyhre: one last nit: in this particular helper function, i think it is confusing to pass `K` and `scale_block_n` args then immediately delete them via `del K` and `del scale_block_n`. let's just only pass the args we need here.

the other places you added explicit scale_block_n/scale_block_k look good (main kernel entry point, user facing api/wrapper, etc).

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
```diff
@@ -35,20 +36,24 @@ def _make_tile_smem_layouts(cute, tile_n: int, tile_k: int):
 # Config format:
 # (compute_warps, tile_n, tile_k, k_tiles_per_cta)
 _CUTEDSL_CONFIGS = {
-    "bf16_default": (6, 32, 128, 4),
+    "bf16_1x32": (4, 32, 128, 4),
```
danielvegamyhre: should this one be 32x1 (not 1x32) since it is quantizing shape (E,N,K) with 32x1 scaling along N

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
```diff
@@ -187,6 +197,80 @@ def _store_scale(
...
+            lane: cutlass.Int32,
+            USE_RCEIL: cutlass.Constexpr[bool],
+            BLOCKED_SCALE_OUTPUT: cutlass.Constexpr[bool],
+        ):
+            amax = compute_amax(vals_block)
+            if cutlass.const_expr(SCALE_DIM_K_VALUE == 32):
+                amax = self._warp_reduce_max(amax)
+            scale_biased, inv_scale = compute_scale_from_amax(amax, USE_RCEIL)
+            if cutlass.const_expr(SCALE_DIM_K_VALUE == 32):
+                if lane == cutlass.Int32(0):
```
danielvegamyhre: based on your `store_scale_32x32` implementation here:

```python
            if cutlass.const_expr(BLOCKED_SCALE_OUTPUT):
                n_row = n_block * cutlass.Int64(32) + cutlass.Int64(lane)
                scales_expert[n_row, k_block] = scale_u8
```

don't we need all 32 lanes to store the warp reduced scale factor, not just thread 0, in order to replicate it to all 32 rows of the SF tile?

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
```diff
@@ -598,11 +686,18 @@ def __call__(
             blocked_scale_layout = cute.make_layout((1,))
             e_scale_stride = cutlass.Int64(0)
             if cutlass.const_expr(BLOCKED_SCALE_OUTPUT_VALUE):
-                padded_scale_cols = cute.round_up(n_blocks, 4)
-                k_block_tiles = cute.ceil_div(K, 128)
-                n_block_tiles = padded_scale_cols // cutlass.Int64(4)
+                scale_rows = K if cutlass.const_expr(SCALE_DIM_K_VALUE == 1) else N
```
danielvegamyhre, line 714: i'm a bit confused, why does the number of rows in the scale factor tensor depend on the granularity of our scale factor size (32x32 vs 32x1)?  same question for `scale_cols` below. is this right? if so, could you please add a brief comment explaining

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
alexsamardzic: Changed.

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
alexsamardzic: That was a typo, fixed.

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
alexsamardzic: Indeed that was a bug.  I actually encountered it when I tried end-to-end test, and fixed it, but instead of adding the fix to commit, it stayed in the [patch](https://github.com/user-attachments/files/26612478/diff.txt) attached to my first commit.  I've pushed this change now, and here is the proper patch for end-to-end testing: [diff-end-to-end-test.txt](https://github.com/user-attachments/files/26710402/diff-end-to-end-test.txt).  Shall we add this patch, to be an option for given grouped MM test?

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
alexsamardzic: I think the confusion comes from the fact that I tried to keep single scale factor for 32x32 block as an option for the quantization side of the things, to be ready in case we decide to try with grouped MM kernel that would handle this situation.  This variant should be under `blocked_scale_output = False` condition.  I've added some comments to clarify it, and I've also added an explicit test variant for this case.  Is this explaining it, or you had something else on mind here?

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
```diff
@@ -528,18 +629,19 @@ def kernel(
...
+                                            k_block
+                                            if cutlass.const_expr(
+                                                SCALE_DIM_K_VALUE == 32
+                                            )
+                                            else cutlass.Int64(0),
+                                            lane,
+                                            USE_RCEIL,
                                             BLOCKED_SCALE_OUTPUT_VALUE,
                                         )
                                         self._quantize_store_tail(
```
danielvegamyhre, line 631: Comparing the `if cutlass.const_expr(IS_FULL_K_TILES)` "true" path vs "false" path, the code looks identical except:
- in the 'false' path, it has an extra condition `k_in_bounds` (e.g. `if k_rel < TILE_K and k_in_bounds: ...`) 
- at the end we call`self._quantize_store_full` vs `self._quantize_store_tail` 

Could we simplify this by:
1. always just doing the k_in_bounds check (unnecessary for full blocks but i doubt that 1 boolean check will hurt perf?) 
2. executing the shared code (which would now be one unified path)
3. doing the if/else at the end for full vs tail store?

if i missed something, please let me know

alexsamardzic: Updated.

danielvegamyhre: Benchmarks still look the same after the change?

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
```diff
@@ -187,6 +191,99 @@ def _store_scale(
...
+            USE_RCEIL: cutlass.Constexpr[bool],
+            BLOCKED_SCALE_OUTPUT: cutlass.Constexpr[bool],
+        ):
+            amax = compute_amax(vals_block)
+            if cutlass.const_expr(SCALE_DIM_K_VALUE == 32):
+                amax = self._warp_reduce_max(amax)
+            scale_biased, inv_scale = compute_scale_from_amax(amax, USE_RCEIL)
+            if cutlass.const_expr(SCALE_DIM_K_VALUE == 32):
+                if cutlass.const_expr(BLOCKED_SCALE_OUTPUT):
+                    self._store_scale_32x32(
```
danielvegamyhre, line 257: Can you please add a comment to clarify the blocked path uses all 32 lanes to replicate the warp-reduced scale, versus the non-blocked (row major) path uses a single lane to write a single non replicated scale factor for the 32x32 block?

Just to make this calling logic more immediately understandable, so the reader doesn't have to dive deeper into the code to understand the higher level calling logic.

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
```diff
@@ -187,6 +191,99 @@ def _store_scale(
...
+                        scales_expert,
+                        e,
+                        n_block,
+                        k_block,
+                        lane,
+                        scale_biased,
+                        BLOCKED_SCALE_OUTPUT,
+                    )
+            else:
+                self._store_scale(
```
danielvegamyhre, line 277: nit but can we call this `_store_scale_32x1` now to be explicit, since this kernel now supports different scale granularities. it would be preferable for both 32x32 and 32x1 be explicit, rather than one be implicit and one be explicit.

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
danielvegamyhre: Would this complexity of swapping K/N for rows/cols be eliminated if we only supported blocked layout for 32x32?

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
alexsamardzic: Yes, benchmarked with and without this change, no noticeable difference.

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
alexsamardzic: Done.

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
alexsamardzic: Updated.

torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_3d.py
alexsamardzic: Sorry, I think I answered a different question above. You were asking about the `blocked_scale_output = True` case, while I replied about the compact `blocked_scale_output = False` 32x32 representation. In the blocked path this is intentional: 32x1 and 32x32 map to different logical scale layouts before blocking, so scale_rows / scale_cols do depend on the granularity there.

danielvegamyhre: i see, so to summarize my understanding:

- for 32x1 blocked layout the scale shape is (K, N//32) since the grouped gemm expects this shape
- for 32x32 blocked layout, we conceptually have one scale per (N//32, K//32) block, then replicate it by having every lane write it to its associated row, giving us (N, K//32). 

is that correct? also, have you tested the 32x32 blocked outputs with the torch._scaled_grouped_mm mxfp8 grouped gemm?

alexsamardzic: Yes, this is correct.

As for the testing:  What is tested for 32x32 at the moment is the quantization itself in test_cuda_mx_3d_cutedsl_numerics.  If you want, I can add a grouped MM test that consumes the blocked 32x32 scales.

danielvegamyhre: Yes please add a test that confirms A @ B with 1x32 for A and 32x32 for B works, that would be great

alexsamardzic: Done.

test/prototype/moe_training/test_mxfp8_grouped_mm.py
```diff
@@ -51,28 +55,66 @@
 @skip_if_rocm("ROCm not supported")
 @pytest.mark.parametrize("M,K,N", [(1024, 1024, 1024), (1024, 2048, 4096)])
 @pytest.mark.parametrize("num_experts", (1, 8, 16))
-def test_emulate_mxfp8_grouped_gemm_2d_3d(M, K, N, num_experts):
-    x = torch.randn(M, K, dtype=torch.bfloat16, device="cuda")
+@pytest.mark.parametrize("scale_block_k", (1, 32), ids=("32x1", "32x32"))
+@pytest.mark.parametrize(
+    "scale_mode", (ScaleCalculationMode.FLOOR, ScaleCalculationMode.RCEIL)
+)
+def test_emulate_mxfp8_grouped_gemm_2d_3d(
```
danielvegamyhre, line 67: @alexsamardzic it seems like this test was changed from testing the emulated path, to testing the non-emulated/real-sm100 path? instead we should keep the test for the emulated path, and make a new test for this real 1x32 @ 32x32 case (make sure to add skip if not on sm100)

alexsamardzic: Done.  We have now `test_emulate_mxfp8_grouped_gemm_2d_3d` and `test_mxfp8_grouped_gemm_2d_3d`, that both test both 32x1 and 32x32 variants that we have at the moment.

danielvegamyhre (approved):

PR #4259, opened 2026-04-09
Improve HQQ scale-only

metascroy:
Improves HQQ scale-only by falling back to affine for groups where HQQ hurts max reconstruction error at the expense of MSE:

1. Initialize scale using `choose_qparams_affine` with `SYMMETRIC_NO_CLIPPING_ERR`, which correctly handles asymmetric ranges.
2. Run the standard HQQ least-squares iteration (unchanged).
3. Per-group fallback: if HQQ's max reconstruction error exceeds affine's for any group, fall back to affine's scale for that group.

## Results

Tested across 4 models: **Phi-4-mini**, **Llama-3.2-1B**, **Gemma-3-4B**, **Qwen3-4B**.

| Metric | HQQ (hybrid) vs Affine | HQQ (hybrid) vs Original HQQ |
|---|---|---|
| **MSE (avg reconstruction error)** | **12–90% lower** ✅ | ~6% higher (small tradeoff) |
| **Max reconstruction error** | **Always ≤ affine** (guaranteed by per-group fallback) ✅ | **0–51% lower** ✅ |
| **Clipping rate** | 4.2–4.4% (vs affine's 0–3.8%) | Down from 5.1–5.8% ✅ |

The hybrid approach retains HQQ's MSE optimization advantage while guaranteeing max reconstruction error never exceeds affine's, eliminating the pathological outlier errors that degraded generation quality.

**Running lm-eval benchmarks now**

### DIFF
--- a/torchao/quantization/quant_primitives.py
+++ b/torchao/quantization/quant_primitives.py
@@ -2061,11 +2061,28 @@ def _choose_qparams_and_quantize_scale_only_hqq(
     n_groups = k // group_size
     Wg = W.view(n, n_groups, group_size)
 
-    # Initialize per-block scales as max-abs / qabs
-    # scale.shape = [n, n_groups]
-    qabs = max(abs(qmin), abs(qmax)) or 1
-    scale = (Wg.abs().amax(dim=2) / qabs).clamp_min(compute_eps)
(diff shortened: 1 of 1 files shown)


metascroy:
Model evaluations are not showing clear benefit in general

PR #4261, opened 2026-04-10
Remove LinearActivationWeightObservedTensor and insert_observers_

jerryzh168:
* #4264
* #4263
* #4262
* __->__ #4261

Summary:
Delete `linear_activation_weight_observed_tensor.py` and the
`insert_observers_` function that depended on it. Also remove the
`TestLinearObserver` test class and unused imports.

BC-breaking notes:
```
# 0.17
from torchao.quantization.quant_api import insert_observers_
insert_observers_(linear_module, input_act, output_act)
# 0.18
`insert_observers_` function is removed since it's using our older stack
```

Test Plan:
pytest test/quantization/test_observer.py -x


torchao/quantization/quant_api.py
```diff
@@ -254,87 +250,6 @@ def replace_conv2d_1x1(conv):
     )
 
 
-def insert_observers_(
```
andrewor14, line 257: This was not deprecated previously. Is anyone using this?

andrewor14:
@claude review

andrewor14 (approved):
Stamping to unblock. We should make sure there are no internal and external users

torchao/quantization/quant_api.py
jerryzh168: it's used in tutorials I think: https://github.com/pytorch/ao/blob/5750fa8a1226504f1309af81693bab35e5a5dd30/docs/source/eager_tutorials/static_quantization.rst#L118, but we can probably remove that tutorial for now as well

jerryzh168: oh actually that is a separate definition, not related to this function

jerryzh168: I can add a bc breaking note, seems low risk to break this

PR #4266, opened 2026-04-10
Add reduce_range to avoid overflow in int8 tensor

cyxlily:
Add reduce_range optional flag. Default False. If True, use reduced int8 range [-64, 63] instead of full range [-128, 127] to reduce overflow risk on platforms without VNNI instructions. Kept optional for backward compatibility with older call sites and serialized configs.


test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -515,5 +516,128 @@ def test_int8_weight_only_v2_correct_eps(self, dtype, device):
         self.assertGreater(sqnr, 40, f"SQNR too low: {sqnr}")
 
 
+@common_utils.instantiate_parametrized_tests
+class TestInt8CPU(TorchAOIntegrationTestCase):
```
Xia-Weiwen:
```suggestion
class TestInt8TensorCPU(TorchAOIntegrationTestCase):
```

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -515,5 +516,128 @@ def test_int8_weight_only_v2_correct_eps(self, dtype, device):
...
+    @common_utils.parametrize("reduce_range", [False, True])
+    @common_utils.parametrize(
+        "granularity",
+        [PerRow(), PerTensor(), (PerRow(), PerTensor()), (PerTensor(), PerRow())],
+    )
+    @common_utils.parametrize(
+        "act_mapping_type", [MappingType.SYMMETRIC, MappingType.ASYMMETRIC]
+    )
+    def test_dynamic_config_reduce_range(
```
Xia-Weiwen:
```suggestion
    def test_int8tensor_cpu(
```

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -515,5 +516,128 @@ def test_int8_weight_only_v2_correct_eps(self, dtype, device):
         self.assertGreater(sqnr, 40, f"SQNR too low: {sqnr}")
 
 
+@common_utils.instantiate_parametrized_tests
+class TestInt8CPU(TorchAOIntegrationTestCase):
+    @common_utils.parametrize("reduce_range", [False, True])
```
Xia-Weiwen: Please put this in the end.

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -515,5 +516,128 @@ def test_int8_weight_only_v2_correct_eps(self, dtype, device):
...
+        x = torch.randn(32, 256, dtype=torch.bfloat16, device=device)
+        x_int8 = _choose_quant_func_and_quantize_tensor(x, w.act_quant_kwargs)
+        self.assertTrue(
+            torch.all(x_int8.qdata >= expected_min)
+            and torch.all(x_int8.qdata <= expected_max)
+        )
+
+        y_fp = linear_fp(x)
+        y = linear(x)
+        self.assertGreater(
```
Xia-Weiwen: Since the reference output is computed by fake quantization, it's better to assert results are all close rather than to use SQNR.

Xia-Weiwen (approved):

Xia-Weiwen:
@claude review

cyxlily:
Thanks for the suggestion.
1. We need to keep reduce_range as [Optional[bool]] for backward compatibility with old quantized models. Otherwise we will got error:
[FAIL] 1 Int8Tensor(s) failed dequantize(): ['fc.weight']
2. I updated the CPU test so we still run the setup/quantization/forward path on non-VNNI CPUs, but we skip the SQNR (compute_error) assertion specifically when reduce_range=False and VNNI is unavailable. TestInt8DynamicQuant and TestInt8StaticQuant only test GPU.
3. In this test, I intentionally did not use a fake-quantized reference with assertAllClose.
The goal here is to measure end-to-end degradation from quantization by comparing FP output and quantized output, rather than checking closeness to a specific reference path.

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -12,18 +12,24 @@
 from torch.testing import FileCheck
 from torch.testing._internal import common_utils
 
+from torchao.kernel.intmm import _cpu_is_vnni_supported
```
Xia-Weiwen: Put the util func in torchao/utils.py?

Xia-Weiwen:
@claude review again

Xia-Weiwen (approved):

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -43,10 +52,15 @@ class QuantizeTensorToInt8Kwargs(QuantizeTensorKwargs):
     Args:
         granularity: the granularity for the Tensor, currently either PerRow() or PerTensor()
         mapping_type: whether to use symmetric or asymmetric quant
+        reduce_range: optional flag. If True, use reduced int8 range [-64, 63]
```
jerryzh168: nit: we should be consistent on using numbers v.s. `_REDUCED_QUANT_MIN` etc. I think, right now the quant_api is using the variable but here it's using variables.

jerryzh168: OK to use variables but might also be good to mention what these values are in the docstring as well

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
```diff
@@ -515,5 +521,93 @@ def test_int8_weight_only_v2_correct_eps(self, dtype, device):
         self.assertGreater(sqnr, 40, f"SQNR too low: {sqnr}")
 
 
+@common_utils.instantiate_parametrized_tests
+class TestInt8TensorCPU(TorchAOIntegrationTestCase):
```
jerryzh168: nit: maybe create a new `test_int8_tensor_cpu.py` it will be easier to separate tests by device for CI in the future if we want.

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
jerryzh168: as commented in https://github.com/pytorch/ao/pull/3784#issuecomment-4240324904 I'm wondering if we can just set it automatically and print out a warning, instead of leave this to user?

cyxlily: Thanks for comments, updated.

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
cyxlily: Thanks for comments, updated.

test/quantization/quantize_/workflows/int8/test_int8_tensor.py
cyxlily: Thanks for comments, updated.

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -26,15 +26,28 @@
...
 
 aten = torch.ops.aten
 
+_REDUCED_QUANT_MIN = -64
+_REDUCED_QUANT_MAX = 63
+_FULL_QUANT_MIN = -128
+_FULL_QUANT_MAX = 127
+
+
+def _should_use_reduced_range(tensor: torch.Tensor) -> bool:
```
vkuzo: I think the user should set this and torchao can make it easy to pick the right setting, automagically setting this based on hardware is a footgun

vkuzo (changes requested):

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
jerryzh168: OK, we can expose this as an arg, and by default it will be autopicked based on device, and don't require people to set it? 

```
# reduce_range auto-picked, but can be overwritten
# by default, auto picked based on device
config = Int8DynamicActivationInt8WeightConfig()

# explicit overwrite
config = Int8DynamicActivationInt8WeightConfig(reduce_range=True/False)
```


or are you talking about user should always explicitly set this arg in the config, like changing all callsites to:
```
config = Int8DynamicActivationInt8WeightConfig(reduce_range=get_recommended_reduce_range())
```

Xia-Weiwen: @vkuzo @jerryzh168 Thanks for the comments. To make UX better, we can probably do as @jerryzh168 suggested:
- if users don't set it explicitly, it is decided automatically (reduce_range=true if on CPU and no VNNI, false otherwise)
- if users set it explicitly, we just use whichever the users set and the users are responsible for the results.

How does that sound to you? Thanks.

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -36,17 +37,25 @@
...
+
+
 @dataclass
 class QuantizeTensorToInt8Kwargs(QuantizeTensorKwargs):
     """Tensor kwargs for creating int8 tensor for activation.
 
     Args:
         granularity: the granularity for the Tensor, currently either PerRow() or PerTensor()
         mapping_type: whether to use symmetric or asymmetric quant
+        reduce_range: If None, choose the int8 quantization range automatically.
```
vkuzo: I'm going to quote "[simple over easy](https://docs.pytorch.org/docs/stable/community/design.html#principle-2-simple-over-easy)" here: let's make it simple by requiring the user to specify numerics upfront.  It's ok to ask users on hardware that needs `reduce_range` to specify the required flag manually.

vkuzo (changes requested):

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
Xia-Weiwen: @vkuzo Thanks for the comments. Just to confirm, you don't have concern if `reduce_range` is optional and if it's None we use `reduce_range=False`, do you? This is for backward compatibility in case where people saved their quantized model without this flag and want to load it with a new version of torchao.

vkuzo: yes, that sounds good.  It's also find to provide a helper function which returns the "correct" value based on hardware, but the default value should be simple and deterministic (reduce_range False).

Xia-Weiwen (approved):

Xia-Weiwen:
@claude review

Xia-Weiwen:
@claude review

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
cyxlily: Thanks for the comments. I've set reduce_range default to False. @vkuzo @jerryzh168 please review.

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -43,10 +47,13 @@ class QuantizeTensorToInt8Kwargs(QuantizeTensorKwargs):
     Args:
         granularity: the granularity for the Tensor, currently either PerRow() or PerTensor()
         mapping_type: whether to use symmetric or asymmetric quant
+        reduce_range: optional flag for reduced int8 quantization range to avoid overflow
+            on CPU without VNNI support. Users can call should_reduce_range() to help determine.
     """
 
     granularity: Granularity
     mapping_type: MappingType = MappingType.SYMMETRIC
+    reduce_range: Optional[bool] = False
```
jerryzh168: can you add a comment on why Optional, and whether we can remove it in the future?

I don't think we need to guarantee BC for old checkpoints btw

cyxlily: Thanks for the comments, removed `Optional` as it's not needed to guarantee BC for old checkpoints.

Xia-Weiwen:
Hi @jerryzh168 Could you please review this PR again?

torchao/quantization/quant_api.py
```diff
@@ -881,6 +885,11 @@ def __post_init__(self):
             "Please set it to MappingType.SYMMETRIC or "
             "MappingType.ASYMMETRIC."
         )
+        if self.reduce_range is None:
+            self.reduce_range = False
+        assert self.reduce_range in (True, False), (
+            "`reduce_range` must be True or False. None is defaulted to False."
+        )
```
jerryzh168: probably not needed?

torchao/quantization/quant_api.py
```diff
@@ -978,6 +992,11 @@ def __post_init__(self):
             "Please set it to MappingType.SYMMETRIC or "
             "MappingType.ASYMMETRIC."
         )
+        if self.reduce_range is None:
+            self.reduce_range = False
+        assert self.reduce_range in (True, False), (
```

test/quantization/quantize_/workflows/int8/test_int8_tensor_cpu.py
```diff
@@ -0,0 +1,117 @@
...
+
+        if config_mode == "dynamic":
+            config = Int8DynamicActivationInt8WeightConfig(
+                version=2,
+                granularity=granularity,
+                act_mapping_type=act_mapping_type,
+                reduce_range=reduce_range,
+            )
+        else:
+            act_granularity, _ = Int8Tensor._normalize_granularity(granularity)
```
jerryzh168, line 73: nit: add assert for config_mode == "static"

test/quantization/quantize_/workflows/int8/test_int8_tensor_cpu.py
```diff
@@ -0,0 +1,117 @@
...
+        if is_ROCM():
+            self.skipTest("Don't test CPU for ROCM version of torch")
+
+        torch.compiler.reset()
+
+        M, N, K = 64, 256, 256
+        input_tensor = torch.randn(M, K, dtype=dtype, device=device)
+        model = ToyTwoLinearModel(K, N, K, dtype=dtype, device=device).eval()
+        model_q = copy.deepcopy(model)
+        reduce_range = should_reduce_range(input_tensor.device)
```
jerryzh168, line 60: what's the effect when reduce_range is not set correctly for cpu? should we test that as well?

jerryzh168 (approved):
looks good overall

test/quantization/quantize_/workflows/int8/test_int8_tensor_cpu.py
cyxlily: Thanks for the comments. On CPUs without VNNI, when using torch.compile with int8 matmul, there may be accuracy degradation compared to eager mode due to overflow in oneDNN qlinear. The `reduce_range` setting is designed to prevent this.
We don't need to test incorrect settings because this UT focuses on the recommended path, and I've already provided the helper function `should_reduce_range()` and clear UT Notes.

Xia-Weiwen:
Merge as CI failures are unrelated.

PR #4267, opened 2026-04-10
Fix DTensor errors in PruneOptimizer

lisjin:
Quick caveat: the majority of this PR is new test cases for PruneOptimizer. Those alone account for ~1K lines—let me know if that's too many.

The main change is to update PruneOptimizer to support group-level regularization terms (gamma_index_slope) when using FSDP (i.e., DTensor params). In the past, the the sharding logic for DTensor caused issues when applying PAT's proximal mapping in parallel on the tensor groups (e.g., rows, columns, attention heads).

There is also support for an iterative reweighting on the tensor groups, which Lin developed, based on [this paper](http://arxiv.org/abs/0711.1612). It is still in experimental stages but may prove useful in our collaborations with other teams.

*Edit (4/15)*: See #4283 for the separate branch that iterative reweighting has been moved to.


lisjin:
@claude Please do the following:
1. Summarize the changes in PruneOptimizer (torchao/prototype/pat/optim/pruneopt.py) and which new test cases validate the changes.
2. Give a quick summary of the iterative reweighting method (torchao/prototype/pat/optim/iterative_reweight.py) as it's used in PruneOptimizer. Use the test case test/prototype/pat/test_pat_pruneopt.py for guidance.
3. Review the PR as a whole.

lisjin:
@claude From your previous task/comment, please repeat (1) and (2) using layman's terms only and use at most 300 words for each. Instead of giving pointers to exact line numbers and files, just summarize the practical insights. Then, verify that the test cases mentioned in the last comment were deduplicated.

jerryzh168:
are PruneOptimizer and iterative reweighting two separate features? does it make sense to split them into multiple PRs?

lisjin:
> are PruneOptimizer and iterative reweighting two separate features? does it make sense to split them into multiple PRs?

Hi @jerryzh168, you can think of iterative reweighting as a new feature for PruneOptimizer (which was added in #3429). It would be great to keep them together since they're integrated tightly, but I can also split them if it's too confusing.

jerryzh168:
> > are PruneOptimizer and iterative reweighting two separate features? does it make sense to split them into multiple PRs?
> 
> Hi @jerryzh168, you can think of iterative reweighting as a new feature for PruneOptimizer (which was added in #3429). It would be great to keep them together since they're integrated tightly, but I can also split them if it's too confusing.

OK then it's fine, I think @andrewor14 or @danielvegamyhre should be able to help review this

andrewor14:
@claude @lisjin can the DTensor fixes in PruneOptimizer be moved to a separate PR? Seems like iterative reweighting is an additional feature on top of the fixes, please let me know if I understood this correctly. It would make things easier to review that way. Also, are any of the test changes independent of both of the above features?

lisjin:
@andrewor14 I just removed the iterative reweighting changes and put them in #4283 (stacked on top of this branch). Let me know if anything's unclear

andrewor14:
Great thanks @lisjin!

andrewor14:
@claude review now? Lisa just separated out the iterative reweighting changes

andrewor14 (approved):

PR #4270, opened 2026-04-13
[optim] Add GrokAdamW optimizer with low-bit quantization support

### LINKED ISSUE #1032: Support for GrokAdamW
Hi there,

It would be great to have support for **GrokAdamW** optimizer but with low bit quantization.

You can check reference implementation: https://github.com/cognitivecomputations/grokadamw

It has shown promising results already.

vaibhavhariram:
Fixes #1032

Implements GrokAdamW optimizer — a combination of AdamW with Grokfast's adaptive frequency amplification for accelerating grokking (delayed generalization).

**Key features:**
- EMA filter per parameter to track and amplify slow-varying gradients
- Layer-wise beta1 decay for differentiated learning rates across layers
- Adaptive EMA momentum based on configurable grokking signal functions
- Optional gradient clipping
- Low-bit quantization support: GrokAdamW8bit, GrokAdamW4bit, GrokAdamWFp8

**Implementation:** Follows the existing `_AdamBase` / `single_param_adam` pattern. Adds a `single_param_grokadamw` function that applies the Grokfast EMA filter before the core AdamW update, maintaining compatibility with `torch.compile` and FSDP2.

**References:**
- [GrokAdamW reference implementation](https://github.com/cognitivecomputations/grokadamw)
- [Grokfast: Accelerated Grokking by Amplifying Slow Gradients](https://arxiv.org/abs/2405.20233)
- [Grokking: Generalization Beyond Overfitting](https://arxiv.org/abs/2201.02177)

**Test plan:** Added unit tests following the existing optimizer test pattern.


vaibhavhariram:
@gau-nernst @jerryzh168 @vkuzo 

CLA signed, ready for review. Any blockers?

vkuzo:
Thanks for the PR! Do we have e2e results on a representative model, with perf/accuracy metrics showing that the implementation in this PR is working as expected and provides a tangible benefit over other optimizers?

vaibhavhariram:
@vkuzo apologies for the long silence on this, that's on me.

On the e2e question: I'd like to run the same setup torchao already uses for low-bit optimizer benchmarks. ViT-H (630M) on resisc45 via `benchmarks/benchmark_low_bit_adam.py`, BF16 AMP, compiled, 1 epoch, fixed seed and report peak memory, wall time, and accuracy across a six-run grid: AdamW fp32, AdamW8bit, GrokAdamW fp32, and the three quantized variants this PR adds.

That isolates two separate claims. The quantization claim (GrokAdamW8bit/4bit/Fp8 vs. GrokAdamW fp32) I expect to hold, same memory profile as the existing AdamW8bit/4bit/Fp8. The optimizer claim (GrokAdamW vs. AdamW) I'm less sure about: grokking is a delayed-generalization effect usually demonstrated on small algorithmic tasks, and I doubt it shows up in a 1-epoch ViT-H finetune. If it doesn't, I'd rather report that honestly than go looking for a task where it wins.

Which raises a scoping question before I spend the GPU time. #1032 is closed as not planned, and msaroufim's original note there suggested new optimizers are generally fine to implement downstream. If torchao's position is that GrokAdamW doesn't belong upstream regardless of the numbers, I'd rather hear that now than send you a benchmark table you didn't want to review. If the memory-parity result would be sufficient, I'll run it and post the full grid.

Either answer is useful to me, just let me know which.

vaibhavhariram:
Thanks for the review, and apologies for the long silence.

Honest answer: no, I don't have e2e results on a representative model yet. The current benchmark is a 1M-param toy transformer on CPU, which only validates that the optimizer runs and converges — it doesn't demonstrate a tangible benefit.

To properly answer your question, I'd need to show convergence curves + memory savings on a real model (e.g., ViT or GPT-2 fine-tune on GPU) and compare against both `torch.optim.AdamW` and the existing `AdamW8bit` in torchao. I'm happy to run that if you think the direction is still worth pursuing.

That said, I want to be upfront: I saw that #1032 was closed as not planned and that the general position is new optimizers are better implemented downstream. If the bar for merging a new optimizer into `torchao/optim/` is higher than "it works correctly with low-bit states" — i.e., it needs to demonstrably beat existing options on standard benchmarks — I'd understand if this doesn't clear it. GrokAdamW's value proposition is around grokking dynamics (delayed generalization), which is hard to show on a typical fine-tuning benchmark.

Let me know if it's worth running the numbers, or if this is better kept as a downstream package that builds on torchao's quantized optimizer infrastructure.

vkuzo:
> If the bar for merging a new optimizer into torchao/optim/ is higher than "it works correctly with low-bit states" — i.e., it needs to demonstrably beat existing options on standard benchmarks — I'd understand if this doesn't clear it.

I'd say that's a minimum, it would be better to also provide evidence of product-market fit, i.e. that people will actually use it.

vaibhavhariram:
Thanks — that's a clear answer and a reasonable bar.

I don't have evidence of demand for GrokAdamW specifically, and I don't think
I can manufacture it. Closing this rather than leaving it open.

One thing that might be worth separating from this PR: the reason implementing
this took a fork rather than composition is that _AdamBase doesn't expose its
quantized-state machinery to downstream optimizers. If torchao's position is
that new optimizers belong downstream, making that infrastructure reusable
would serve that directly. Happy to open an issue if it's of interest — no
expectation either way.

PR #4272, opened 2026-04-13
[xpu][mx][test] Enable mx serialization tests on xpu

ugolowic:

Parametrize test_serialization over available devices (CUDA, XPU) instead of hardcoding CUDA-only.

Replace hardcoded `python` with `sys.executable` in subprocess calls to ensure the test uses the same interpreter (important when running inside a virtualenv).

### DIFF
--- a/test/prototype/mx_formats/test_mx_serialization.py
+++ b/test/prototype/mx_formats/test_mx_serialization.py
@@ -6,6 +6,7 @@
 
 import os
 import subprocess
+import sys
 import tempfile
 
 import pytest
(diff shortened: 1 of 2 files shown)


test/prototype/mx_formats/test_mx_serialization.py
```diff
@@ -27,16 +28,22 @@
...
-@pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA not available")
-@pytest.mark.skipif(not is_sm_at_least_100(), reason="needs CUDA capability 10.0+")
+@pytest.mark.skipif(
+    not torch.accelerator.is_available(), reason="No accelerator available"
+)
+@pytest.mark.skipif(
+    torch.cuda.is_available() and not is_sm_at_least_100(),
+    reason="needs CUDA capability 10.0+",
+)
 @pytest.mark.parametrize("recipe_name", ["mxfp8", "nvfp4"])
```
kgajdamo, line 32: Will the test fail if nvfp4 is run on xpu?

kgajdamo (approved):
Looks good to me

xiaowangintel:
@claude review this PR.

test/prototype/mx_formats/test_mx_serialization.py
ugolowic: It doesn't fail, serialization isn't affected by data format here.

test/prototype/mx_formats/test_mx_serialization.py
```diff
@@ -27,16 +28,22 @@
...
 def test_serialization(recipe_name):
     """
     Ensure that only `import torchao.prototype.mx_formats` is needed to load MX
     and NV checkpoints.
     """
+    device = torch.accelerator.current_accelerator()
```
xiaowangintel, line 38: 1. No Hardcoded Device Strings
2. Use @parametrize for Device Parameterization
All device-dependent tests should be decorated with @parametrize("device", get_available_devices()) and accept device as a function argument.

PR #4273, opened 2026-04-13
update profiling util to dump trace per rank

danielvegamyhre:
## Summary
I want to be able to compare traces from different ranks, i.e. if there is a straggler. This change will support that.


jerryzh168 (approved):

jerryzh168:
can probably mark `modue: not user facing` for utils used by developers

danielvegamyhre:
> can probably mark `modue: not user facing` for utils used by developers

Makes sense, updated

PR #4278, opened 2026-04-14
Fix race condition in mxfp8 CUDA kernels

lw:
A race condition was present in two kernels due to a bad ordering between a syncthreads and an async-proxy fence. The fence is needed because it makes sure that the calling thread's writes to shmem are visible in the async proxy. However, the operation we're synchronizing with is the TMA write issues by thread 0, hence we need to establish a causality link between _all_ the fences performed by _all_ threads, and the issuing of the TMA load by thread 0. Thus the syncthreads must be inserted in between these two operations.

The CUDA programming guide is very explicit about this, in [section 10.29.1. Using TMA to transfer one-dimensional arrays](https://docs.nvidia.com/cuda/cuda-c-programming-guide/#using-tma-to-transfer-one-dimensional-arrays):

Screenshot 2026-04-14 at 11 48 31
Screenshot 2026-04-14 at 11 48 52

### DIFF
--- a/torchao/csrc/cuda/mx_kernels/mx_block_rearrange_2d_M_groups.cu
+++ b/torchao/csrc/cuda/mx_kernels/mx_block_rearrange_2d_M_groups.cu
@@ -316,10 +316,9 @@ __global__ void mx_blocked_layout_2d_M_groups_kernel(
                 ) = data;
             }
         }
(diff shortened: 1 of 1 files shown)


danielvegamyhre (approved):
Good catch @lw, thanks for the fix!

PR #4283, opened 2026-04-15
Add iterative reweighting to PruneOptimizer

lisjin:
See #4267 for the original context on iterative reweighting. This diff is stacked on top of it for better readability.


andrewor14:
@claude review

lisjin:
@andrewor14 would this one be fine to merge too?

andrewor14:
Looks good, I think we should address some of claude's feedback, seems like it's mostly doc suggestions

andrewor14 (approved):
Stamping to unblock, but please add some docs/comments accordingly

PR #4293, opened 2026-04-17
[mxfp8 moe training] fuse dynamic per-group padding into cutedsl 2d mxfp8 quant kernel

### LINKED ISSUE #4184: fuse dynamic per group padding into cutedsl 2d MXFP8 quantization kernel
For the non-EP case, we need to fuse per group padding to nearest multiple of 32/128 into the MXFP8 quantization kernel added in #4156 , to avoid the expensive extra copy in the standalone padding kernel.

MagellaX:
Closes #4184.

## Summary

Today, the non-EP MoE forward path in `_MXFP8GroupedMM.forward` pads the
activation along M (`pad_token_groups`) and then calls the CuTeDSL 2D MXFP8
quantize kernel from #4156 on the fully padded tensor. That pad is a full
input-sized memory copy whose only purpose is to give the quantize kernel
aligned TMA tiles.

This PR adds a CuTeDSL kernel that takes the unpadded input directly and
emits an upper-bound-sized padded output in a single launch, avoiding the
intermediate copy entirely. It uses `cute.domain_offset` for unaligned TMA
loads at each group's real row base, and zeroes padding rows in shared
memory before the TMA store.

## What's new

- `torchao/prototype/moe_training/kernels/mxfp8/cutedsl_quantize_2d_1x32_padded.py`
  with `Mxfp8Quantize2dPaddedKernel` and public wrapper
  `mxfp8_quantize_cutedsl_2d_1x32_padded` (registered as
  `torchao::mxfp8_quantize_2d_1x32_padded_cutedsl`).
- `compute_fused_padding_metadata` host helper in `quant.py`: on-device pass
  that computes padded offsets and per-M-tile origins with no D2H sync,
  matching the upper-bound allocation semantics of `pad_token_groups`.
- Tests and a bench script (see below).

### Drive-by fix (first commit)

`_compile_mxfp8_quantize_2d_cutedsl` in the base kernel (#4156) was
`@functools.cache`d on an `offs: Tensor` argument whose only compile-time use
is an `offs is not None` branch. Because `hash(Tensor)` is identity-based,
every call with a freshly-allocated `padded_offs` missed the cache and
triggered a full CuTeDSL re-JIT (~1.4s of overhead per call on B200). Changed
the cache key to `has_offs: bool`; no functional change. Happy to split this
into its own PR if you'd prefer.... :) lmk

## Scope (v1)

- `alignment_size=128` only. The issue asks for 32/128; 32 is deferred to a
  follow-up since it needs a different TMA tile / scale-row layout. A clean
  error is raised for 32.
- Non-EP only, as the issue requests.
- Not yet wired into `_MXFP8GroupedMM.forward`; this PR adds the kernel and
  leaves integration to a small follow-up so the diff stays focused.

## Test plan

New test `test_cuda_mx_dim0_2d_padded_fused_numerics` in
`test/prototype/moe_training/test_kernels.py`: **72 cases** covering
{bf16, fp32} × {FLOOR, RCEIL} × varied M/K shapes including non-aligned K
(157, 300, 1025, 7177, 16392) × pad-needed and pad-is-noop. Each case asserts
bit-exact equality (`torch.equal`) against the two-kernel reference for both
the quantized data tensor and the blocked scales (via `from_blocked` to
ignore uninitialized gap bytes in the tcgen05 blocked layout). **All 72 pass
on B200 (SM 10.0).**

## Benchmark

`benchmarks/prototype/moe_training/mxfp8/bench_cutedsl_quantize_2d_1x32_padded.py`
compares three paths at DeepSeek-V3-shaped activations, 8 groups, unaligned
per-token offsets, on B200 with `torch` nightly + cu128,
`do_bench(return_mode="median")`:

| input_shape M×K | triton µs | cutedsl_base µs | **fused µs** | **vs base** | **vs triton** | fused GB/s |
|-----------------|-----------|-----------------|--------------|-------------|---------------|------------|
| 8192×2048       | 317       | 214             | 229          | 0.94×       | 1.38×         | 232        |
| 8192×7168       | 354       | 261             | 232          | 1.13×       | 1.53×         | 801        |
| 32768×2048      | 343       | 252             | 225          | 1.12×       | 1.53×         | 915        |
| 32768×7168      | 538       | 534             | 228          | 2.34×       | 2.36×         | 3150       |
| 131072×2048     | 578       | 570             | 231          | 2.47×       | 2.50×         | 3534       |
| 131072×7168     | 1749      | 1686            | 549          | **3.07×**   | **3.18×**     | **5200**   |

At realistic DeepSeek-V3 activations the fused kernel hits ~5.1 TB/s
effective HBM bandwidth (~64% of B200 HBM3e peak), a 3× speedup over the
production pad + CuTeDSL pipeline. The small-shape regression (0.94× at
8192×2048) is the SMEM zero-fill overhead for padding rows dominating when
the pad copy it replaces is already tiny; at any realistic MoE activation
size the fused path is a clear win.

cc @danielvegamyhre

### DIFF
--- a/torchao/prototype/moe_training/kernels/mxfp8/triton_dispatch_quantize.py
+++ b/torchao/prototype/moe_training/kernels/mxfp8/triton_dispatch_quantize.py
@@ -1,0 +1,533 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Fused MoE dispatch + MXFP8 quantization + blocked-scale rearrangement kernel.
+
+This module implements Triton kernels that fuse, in one GPU pass:
+
+  * per-row gather from a source tensor via a permutation index vector
+    (MoE token dispatch / per-group padding),
+  * per-row MXFP8 quantization (1x32 scaling granularity), and
+  * direct write of scales in the tcgen05 blocked layout consumed by
+    ``torch._scaled_grouped_mm``.
+
+Motivation (ao#4184). The existing MoE forward pipeline on SM 10.x does three
+sequential memory passes: pad token groups -> MXFP8 quantize -> rearrange
+scales into the blocked layout. For realistic batch sizes (total_M in the
+4k-32k range on DeepSeek-V3-like shapes) this is memory-bandwidth bound, and
+the padding copy alone burns ~1/3 of the budget. Fusing these three steps into
+one kernel lets us touch global memory exactly once per token.
+
+The same core kernel handles both the non-EP path (source rows are in
+expert-major order, we only pad) and the EP path (``src_indices`` is an
+arbitrary permutation produced by ``generate_permute_indices``): both reduce
+to "for each output row r, gather x[src_indices[r], :] or emit zeros if
+src_indices[r] == -1".
+
+Layout contract (v1, alignment = 128):
+
+  * Per-group sizes are multiples of 128 so that groups align with the
+    128-row blocks of the blocked scale layout. This lets each CTA write
+    exactly one (128, 4) blocked tile with no cross-group bookkeeping.
+  * The blocked scale output matches the layout produced by
+    ``triton_mx_block_rearrange_2d_M_groups`` for alignment=128 (identical
+    bytewise on non-gap cells).
+"""
+
+from typing import Tuple
+
+import torch
+from torch import Tensor
+from torch.utils._triton import has_triton
+
+from torchao.utils import ceil_div, torch_version_at_least
+
+# v1 tuning parameters (see module docstring).
+_ALIGNMENT = 128
+_BLOCK_ROWS = 128
+_SCALE_BLOCK_COLS = 4
+_SCALE_BLOCK_SIZE = 32
+_COL_TILE_SIZE = 128  # == _SCALE_BLOCK_COLS * _SCALE_BLOCK_SIZE
+
+
+def _validate_src_indices(
+    x: Tensor,
+    src_indices: Tensor,
+) -> Tuple[int, int, int, int]:
+    assert x.dtype == torch.bfloat16, f"x must be bfloat16, got {x.dtype}"
+    assert x.is_contiguous(), "x must be contiguous"
+    assert x.ndim == 2, f"x must be 2D, got {x.ndim}D"
+    assert src_indices.dtype == torch.int32, (
+        f"src_indices must be int32, got {src_indices.dtype}"
+    )
+    assert src_indices.ndim == 1, "src_indices must be 1D"
+    unpermuted_M, K = x.shape
+    padded_M = src_indices.shape[0]
+    assert K % _SCALE_BLOCK_SIZE == 0, (
+        f"K must be divisible by {_SCALE_BLOCK_SIZE}, got K={K}"
+    )
+    assert padded_M % _ALIGNMENT == 0, (
+        f"src_indices length must be a multiple of {_ALIGNMENT}, got {padded_M}"
+    )
+    return unpermuted_M, K, padded_M, K // _SCALE_BLOCK_SIZE
+
+
+if torch_version_at_least("2.7.0") and has_triton():
+    import triton
+    import triton.language as tl
+
+    from torchao.prototype.mx_formats.kernels import (
+        _triton_calculate_scale_floor,
+        _triton_calculate_scale_rceil,
+    )
+
+    def _dispatch_quantize_autotune_configs():
+        # BLOCK_ROWS and COL_TILE_SIZE are locked by the blocked scale layout
+        # in v1; we only sweep over warp/stage counts. A broader tuning pass
+        # can come once we have more shapes to profile against.
+        configs = []
+        for num_warps in (4, 8):
+            for num_stages in (2, 3, 4):
+                configs.append(
+                    triton.Config({}, num_warps=num_warps, num_stages=num_stages)
+                )
+        return configs
+
+    @triton.jit
+    def _quantize_row_tile_and_store(
+        x_block,
+        m_offs,
+        col_offs,
+        row_mask,
+        col_mask,
+        qdata_ptr,
+        scales_ptr,
+        K,
+        padded_scale_cols,
+        pid_m,
+        pid_k,
+        BLOCK_ROWS: tl.constexpr,
+        COL_TILE_SIZE: tl.constexpr,
+        SCALE_BLOCK_SIZE: tl.constexpr,
+        SCALE_BLOCK_COLS: tl.constexpr,
+        SCALING_MODE: tl.constexpr,
+        USE_PTX: tl.constexpr,
+    ):
+        """Given a (BLOCK_ROWS x COL_TILE_SIZE) bf16 tile, MXFP8-quantize it
+        rowwise and write fp8 row-major + one (128,4) scale tile in the
+        tcgen05 blocked layout."""
+        x_block_r = x_block.reshape(
+            BLOCK_ROWS * SCALE_BLOCK_COLS, SCALE_BLOCK_SIZE
+        )
+        x_block_abs_r = tl.abs(x_block_r)
+
+        if SCALING_MODE == "rceil":
+            descale_fp32_r, scale_e8m0_r = _triton_calculate_scale_rceil(
+                x_block_abs_r, axis=1, USE_PTX=USE_PTX
+            )
+        else:
+            tl.static_assert(SCALING_MODE == "floor")
+            descale_fp32_r, scale_e8m0_r = _triton_calculate_scale_floor(
+                x_block_abs_r, axis=1
+            )
+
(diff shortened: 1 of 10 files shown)


danielvegamyhre:
I tried this as well, but unforutnately the performance is just not good enough for realistic batch sizes (total_M 4096 to 32768). if you can get it above 5k gbps or so for those shapes maybe we can land, but i think the appraoch is fundamentally flawed, the padding needs to be fused into the dispatch when were shuffling the data already, rather than do it separately later.

MagellaX:
> I tried this as well, but unforutnately the performance is just not good enough for realistic batch sizes (total_M 4096 to 32768). if you can get it above 5k gbps or so for those shapes maybe we can land, but i think the appraoch is fundamentally flawed, the padding needs to be fused into the dispatch when were shuffling the data already, rather than do it separately later.


yeah so i pushed the changes right there (commit `77da665`) — full rewrite per your feedback, padding is fused into the dispatch now instead of being a separate pass after.

dropped CuTeDSL entirely (its python API doesn't expose arbitrary per-row gather for the TMA path) and went single-pass Triton. two entry points: `triton_mxfp8_dispatch_and_quantize(x, src_indices)` for the EP path (takes the `-1`-sentinel permutation `generate_permute_indices` already produces) and `triton_mxfp8_pad_and_quantize(x, group_offsets)` as a non-EP drop-in for `pad_token_groups` → `triton_to_mxfp8_dim0` → `triton_mx_block_rearrange_2d_M_groups`.

B200, alignment=128, 8 groups, rceil:

| M | K | fused µs | fused GB/s | vs today's a128 3-stage |
|--:|--:|---------:|-----------:|------------------------:|
| 16384 | 7168 | 70.8 | **5140** | 7.02x |
| 32768 | 5120 | 93.2 | **5515** | 5.60x |
| 32768 | 7168 | 126 | **5713** | 4.47x |

peak 5.71 TB/s HBM (~72% of B200 HBM3e), so the >5k bar is cleared on the realistic shapes (M ≥ 16k, K ≥ 5k). full sweep across M ∈ {4k..32k} × K ∈ {2k, 5k, 7k} × groups ∈ {4, 8, 16} is **4.1x–8.1x vs. a128 3-stage** and **2.0x–3.3x vs. today's a32 CUDA fused-pad path**. 68 parametrize cases pass bit-exact vs. the 3-stage reference.

still scoped to alignment=128 / no autograd wiring per your earlier "keep v1 tight" call --alignment=32 and the forward integration are follow-ups. lmk what you want to see next... happy to implement :)

torchao/prototype/moe_training/kernels/mxfp8/triton_dispatch_quantize.py
```diff
@@ -0,0 +1,533 @@
...
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Fused MoE dispatch + MXFP8 quantization + blocked-scale rearrangement kernel.
+
+This module implements Triton kernels that fuse, in one GPU pass:
+
+  * per-row gather from a source tensor via a permutation index vector
+    (MoE token dispatch / per-group padding),
```
danielvegamyhre, line 12: it seems like the dispatch kernel doesn't actually do any dispatching, i don't see any comms. 

don't worry about this, i already have this working in internal code, and it is heavily co-designed with other surrounding parts that aren't open source yet.

one thing you could add if you want, is a kernel that quantizes "grad_out" to both non-transpose and transpose variants, for use in the backward pass:

dgrad = grad_out @ weight
wgrad = grad_out_t @ x

so we need to read the bf16 grad_out once, and write 2 tensors:
1. grad_out_e4m3, grad_out_e8m0_scales
2. grad_out_t_e4m3, grad_out_t_e8m0_scales

e4m3 data should be row major. 
e8m0 scales should be in blocked layout (see [this 'row major to blocked layout' transform kernel, this new kernel should write directly to the blocked format).

we should aim for 90%+ of memcpy bandwidth utilization on B200 (1000W power draw) for shape (4, 4096, 2048). if you get most of the way there, feel free to put up a PR and we can iterate/optimize on it together. 

this would be super useful!

MagellaX:
Across the whole sweep: 1.33x–6.96x over the 4-kernel baseline, consistent ~1.1–1.2 TB/s effective BW (~20–32% of measured B200 bf16 memcpy, which is ~5 TB/s on this rig). That's a real drop-in win but still short of the "90% of memcpy" bar - the remaining headroom is in Blackwell TMA / warp-specialized stores..

MagellaX:
## What changed

Added a **size-based dispatch** inside `triton_mxfp8_quantize_dim0_dim1`:

- **Small shapes** (`M * N < 32M` elements) → single-CTA fused kernel (reads bf16 once, emits all 4 outputs)
- **Large shapes** → split into two standalone kernels (`_dim0_blocked_kernel` + `_dim1_blocked_kernel`) launched on **separate CUDA streams** so they overlap on the HBM controller

Both paths write scales **directly in the tcgen05 `(128, 4)` blocked layout** (no rearrange kernel). Autotune sweep now mirrors production: `(BM, BN) ∈ {(128,128), (128,256), (256,128), (256,256), (512,128), (512,256)}` with `num_warps ∈ (4,8)`, `num_stages ∈ (2,3,4)`, plus config pruning so autotuner never picks a tile larger than the problem.

## B200 numbers

| M | N | 4k_us | fused_us | speedup | % memcpy |
|---|---|---|---|---|---|
| 4096 | 2048 | 278.7 | **38.9** | **7.16x** | **32.4%** |
| 8192 | 2048 | 295.6 | 64.5 | 4.58x | 27.4% |
| 16384 | 2048 | 342.0 | 113.0 | 3.03x | 24.9% |
| 32768 | 2048 | 449.5 | 196.6 | 2.29x | 25.5% |
| 8192 | 5120 | 360.1 | 126.0 | 2.86x | 27.2% |
| 8192 | 7168 | 398.6 | 180.2 | 2.21x | 24.8% |
| 16384 | 5120 | 455.4 | 234.4 | 1.94x | 25.8% |
| 16384 | 7168 | 548.9 | 325.0 | 1.69x | 25.0% |
| 32768 | 5120 | 762.9 | 470.2 | 1.62x | 23.9% |

**Peak 32.4% of memcpy BW** (up from 24.9%), **1.62x–7.16x** speedup over the 4-kernel production path across the full DeepSeek-V3-like backward sweep. All 14 `test_triton_mxfp8_quantize_dim0_dim1_numerics` tests still pass on B200.

## Honest read/opinion on the 90% target

The dim1 output is fundamentally a transposed store (`(N, M)` row-major from an `(M, N)` input), so every row of that output lives on a different HBM row. Production's standalone `to_mxfp8_dim1_kernel` alone caps at ~44% memcpy on B200, and here we're doing that work **plus** the dim0 write and all the tcgen05 swizzle writes. The split-stream trick got dim1 out from under dim0's register pressure and bought overlap on the HBM controller, which is the cleanest thing Triton gives us here without dropping to TMA/CUTLASS or warp specialization. 90% memcpy for a dual-output backward quant with a transposed store doesn't seem realistic in Triton on B200 today; getting into the 25-35% band while maintaining a 1.6-7x win over the production path is the practical ceiling I could find, and I think that's a strong PR on its own...lmk your thoughts

MagellaX:
hey @danielvegamyhre any thoughts????

MagellaX:
hey man @danielvegamyhre can you hit me up with an update??

MagellaX:
hey @danielvegamyhre, pushed another update.

I kept this one focused on the reviewable pieces from the 90%-memcpy exploration:

- added a CuTeDSL grad_out path that emits both dim0 and dim1 MXFP8 forms:
  - a two-stream wrapper over the existing CuTeDSL dim0/dim1 kernels
  - a single-read CuTeDSL prototype that reads bf16 `grad_out` once and writes row-major dim0 qdata, transposed dim1 qdata, and both blocked e8m0 scale tensors
- extended the CUDA MXFP8 TMA quant kernel so scales can be written directly in blocked layout
- starts rowwise output TMA earlier when both rowwise/colwise outputs are active, so the two output paths overlap a bit better
- added SM10-gated numerics tests for both new CuTeDSL grad paths against the existing CuTeDSL references
- added a small benchmark script for this grad_out quantization path

I intentionally left the raw CUDA/NVRTC scratch probes out of the PR to keep the diff sane. Those were useful for mapping the ceiling, but they are not clean enough to review as part of this PR.

Validation:
- `ruff check` passed on the touched Python files
- `py_compile` passed
- Modal B200 targeted run: `90 passed, 1 warning` for the MXFP8/MoE kernel numerics selectors, including the new CuTeDSL stream + single-read tests

On the 90% memcpy target: the best raw CUDA/TMA prototype I found was correct for the target shape `(4, 4096, 2048)` / flattened `16384 x 2048`, and sits around ~88-89% of the fast repeated memcpy baseline. Some individual runs crossed 90%, but not stably enough that I’d want to claim it as solved. This update keeps the useful production-facing parts in the PR, and we can keep iterating on the lower-level kernel separately if you think that direction is still worth pushing...lmk

MagellaX:
Any thoughts guys?? @danielvegamyhre @jerryzh168 @vkuzo

PR #4295, opened 2026-04-17
add FSDP and TP tests for Float8BlockwiseLinear

iamzainhuda:
### Summary                                                                                                                                                                                                                                                    

This adds distributed coverage for Float8BlockwiseLinear on top of the DTensor sharding rules we already added for the blockwise FP8 kernels. The goal is to validate the full module-level training path before wiring the kernels into a real torchtitan run.               
            
Added a `torchao::blockwise_scaled_mm` wrapper around the fallback `_scaled_mm` path and registered DTensor sharding for it. This was necessary because TP on `Float8BlockwiseLinear(use_triton=False)` routes through `aten._scaled_mm_v2`, and that op did not have the DTensor sharding propagation we needed. Without this, the Triton path worked under TP, but the `_scaled_mm` fallback path failed.         
                    
Added the following tests to the following files:
- `test/prototype/blockwise_fp8_training/test_dtensor.py`
  - `test_linear_tp_parity`
    - Adds a 2-GPU tensor-parallel parity test for a quantized `ToyModel`.
    - Uses the standard FFN TP plan:
      - `w1`: colwise
      - `w2`: colwise
      - `out_proj`: rowwise
    - Runs forward, backward, and optimizer step parity against an unsharded reference for both:
      - `_scaled_mm`
      - Triton GEMM

- `test/prototype/blockwise_fp8_training/test_fsdp2.py`
  - `test_fsdp2_parity`
    - Add a 2-GPU FSDP2 parity test for a quantized `ToyModel`.
    - `fully_shard()` is applied on the FFN submodules plus the parent module.

- `test/prototype/blockwise_fp8_training/test_fsdp2_tp.py`
  - Add a manual 4-GPU `FSDP2 + TP` parity test on a 2D mesh with `(dp, tp) = (2, 2)`.
  - Apply TP on the FFN first, then FSDP2 on the DP mesh.
  - This is the closest test to how it will be integrated into torchtitan.             

All of the three tests use the same testing metholdogy:

- build identical reference and distributed `ToyModel` instances
- quantize both with `Float8BlockwiseLinearConfig()`
- run both execution paths by toggling `use_triton`
- feed the same per-replica inputs to reference and distributed models
- compare:
  - forward outputs
  - losses
  - gradients after the expected distributed reduction
  - parameter values after optimizer step         

### Testing

2-GPU DTensor + TP coverage:

```bash
NCCL_SOCKET_IFNAME=lo pytest test/prototype/blockwise_fp8_training/test_dtensor.py -v
```

2-GPU FSDP2 coverage:

```bash
NCCL_SOCKET_IFNAME=lo pytest test/prototype/blockwise_fp8_training/test_fsdp2.py -v
```

4-GPU FSDP2 + TP coverage:

```bash
NCCL_SOCKET_IFNAME=lo NCCL_DEBUG=WARN torchrun --standalone --nproc_per_node=4 test/prototype/blockwise_fp8_training/test_fsdp2_tp.py
```

### DIFF
--- a/torchao/prototype/blockwise_fp8_training/kernels.py
+++ b/torchao/prototype/blockwise_fp8_training/kernels.py
@@ -45,6 +45,18 @@ def _pad_blockwise_128x128_scale_k_major(scale: torch.Tensor) -> torch.Tensor:
     return padded_scale
 
 
+def _prepare_blockwise_scaled_mm_rhs_scale(
+    scale: torch.Tensor,
+    scale_recipe: F.ScalingType | int,
+) -> torch.Tensor:
+    scale_recipe_value = (
+        scale_recipe.value if isinstance(scale_recipe, F.ScalingType) else scale_recipe
+    )
+    if scale_recipe_value != F.ScalingType.BlockWise128x128.value:
+        return scale
+    return _pad_blockwise_128x128_scale_k_major(scale)
+
+
 def _is_column_major(x: torch.Tensor) -> bool:
     assert x.ndim == 2 or x.ndim == 3, "input tensor must be 2D or 3D"
     return x.stride(-2) == 1
@@ -110,8 +122,7 @@ def blockwise_scaled_mm(
         a = a.contiguous()
     if not _is_column_major(b):
         b = b.t().contiguous().t()
-    if scale_recipe_b == F.ScalingType.BlockWise128x128.value:
-        b_s = _pad_blockwise_128x128_scale_k_major(b_s)
+    b_s = _prepare_blockwise_scaled_mm_rhs_scale(b_s, scale_recipe_b)
 
     return torch.ops.aten._scaled_mm_v2.default(
         a,
--- a/torchao/prototype/blockwise_fp8_training/kernels.py
+++ b/torchao/prototype/blockwise_fp8_training/kernels.py
@@ -31,12 +31,21 @@ fp8_gemm_configs_max_autotune = [
 
 EPS = 1e-12
 
+# PyTorch exposes ScalingType in different places across versions. When the
+# enum is unavailable, fall back to the integer recipe IDs expected by
+# aten._scaled_mm_v2.
+_BLOCKWISE_1X128_SCALING_TYPE_ID = 4
+_BLOCKWISE_128X128_SCALING_TYPE_ID = 5
 _SCALING_TYPE = getattr(F, "ScalingType", getattr(torch._C, "_ScalingType", None))
 BLOCKWISE_1X128_SCALING_TYPE = (
-    _SCALING_TYPE.BlockWise1x128 if _SCALING_TYPE is not None else 4
+    _SCALING_TYPE.BlockWise1x128
+    if _SCALING_TYPE is not None
+    else _BLOCKWISE_1X128_SCALING_TYPE_ID
 )
 BLOCKWISE_128X128_SCALING_TYPE = (
-    _SCALING_TYPE.BlockWise128x128 if _SCALING_TYPE is not None else 5
+    _SCALING_TYPE.BlockWise128x128
+    if _SCALING_TYPE is not None
+    else _BLOCKWISE_128X128_SCALING_TYPE_ID
 )
 
 
@@ -45,7 +54,11 @@ def _scaling_type_value(scale_recipe) -> int:
 
 
 def _pad_blockwise_128x128_scale_k_major(scale: torch.Tensor) -> torch.Tensor:
-    # cuBLASLt requires BLK128x128 scales to be K-major with the K stride padded to a multiple of 4.
+    # cuBLASLt's BLK128x128 FP8 scale layout is K-major with
+    # L = ceil(K / 128) rounded up to a multiple of 4. For fp32 scales, this
+    # makes the stride between scale columns a 16-byte multiple, satisfying the
+    # backend layout/alignment requirement.
+    # https://docs.nvidia.com/cuda/cublas/index.html#scaling-factors-layouts
     padded_k_blocks = (scale.shape[0] + 3) // 4 * 4
     if padded_k_blocks == scale.shape[0]:
         return scale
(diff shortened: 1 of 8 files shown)


torchao/prototype/blockwise_fp8_training/kernels.py
```diff
@@ -62,6 +105,78 @@ def _blockwise_gemm_shardings():
...
+            [Replicate(), Replicate(), Replicate(), None, Replicate(), None, None],
+        ),
+        ([Shard(0)], [Shard(0), Replicate(), Shard(0), None, Replicate(), None, None]),
+        ([Shard(1)], [Replicate(), Shard(1), Replicate(), None, Shard(1), None, None]),
+        ([Partial()], [Shard(1), Shard(0), Shard(1), None, Shard(0), None, None]),
+    ]
+
+
+@torch.library.custom_op("torchao::blockwise_scaled_mm", mutates_args=())
+def blockwise_scaled_mm(
```
iamzainhuda, line 123: moved here because i wanted to register sharding rules for the scaled_mm. felt it made sense to move these definitions in kernel.py

torchao/prototype/blockwise_fp8_training/kernels.py
```diff
@@ -34,6 +31,52 @@
 
 EPS = 1e-12
 
+_SCALING_TYPE = getattr(F, "ScalingType", getattr(torch._C, "_ScalingType", None))
```
danielvegamyhre, line 39: why is this necessary? torch changes between versions?

torchao/prototype/blockwise_fp8_training/kernels.py
```diff
@@ -34,6 +31,52 @@
...
+BLOCKWISE_1X128_SCALING_TYPE = (
+    _SCALING_TYPE.BlockWise1x128 if _SCALING_TYPE is not None else 4
```
danielvegamyhre: what are the magic variables 4/5? can we define them as a constant or at least add a comment to make it more easily interpretable

torchao/prototype/blockwise_fp8_training/kernels.py
```diff
@@ -34,6 +31,52 @@
...
+    _SCALING_TYPE.BlockWise128x128 if _SCALING_TYPE is not None else 5
+)
+
+
+def _scaling_type_value(scale_recipe) -> int:
+    return scale_recipe.value if hasattr(scale_recipe, "value") else scale_recipe
+
+
+def _pad_blockwise_128x128_scale_k_major(scale: torch.Tensor) -> torch.Tensor:
+    # cuBLASLt requires BLK128x128 scales to be K-major with the K stride padded to a multiple of 4.
```
danielvegamyhre: Does the multiple of 4 requirement stem from TMA where the tensor's global stride must be a multiple of 16 bytes? i.e., 4 fp32 scales = 16 bytes

If possible, we should explain why this requirement exists in the comment so we have a full understanding

torchao/prototype/blockwise_fp8_training/kernels.py
```diff
@@ -62,6 +105,78 @@ def _blockwise_gemm_shardings():
...
+    a_s: torch.Tensor,
+    scale_recipe_a: int,
+    b_s: torch.Tensor,
+    scale_recipe_b: int,
+    out_dtype: torch.dtype,
+) -> torch.Tensor:
+    if not _is_row_major(a):
+        a = a.contiguous()
+    if not _is_column_major(b):
+        b = b.t().contiguous().t()
```
danielvegamyhre: can we just assert the data is in the expected layout, instead of silently doing these expensive copies to transform it (which hurts perf)? 

i.e., if someone imports and uses blockwise_scaled_mm and calls it with inputs in the wrong layouts, it should fail loudly with clear error messages describing the necessary layouts, rather than "succeed" with bad perf, leaving the user potentially confused why fp8 is not producing speedup vs bf16 when rooflines for the shapes suggest it should.

torchao/prototype/blockwise_fp8_training/linear.py
```diff
@@ -24,52 +30,68 @@
...
+    *,
+    use_triton: bool,
+    triton_kernel,
     mat_a: torch.Tensor,
     mat_b: torch.Tensor,
     scale_a: torch.Tensor,
     scale_recipe_a,
     scale_b: torch.Tensor,
     scale_recipe_b,
+    triton_scale_b: torch.Tensor | None = None,
```
danielvegamyhre, line 76: is the new `triton_scale_b` arg necessary because the Triton gem requires that the B tensor scale be in a different layout than what torch _scaled_mm requires?

if so, in a follow up, we should try to address that - having 2 separate scale_b params for different kernels is pretty gnarly 

alternatively, if the fixes in core _scaled_mm are going to land soon (that will enable to stop using the triton kernel at all), then we can leave this `triton_scale_b` arg for now and just delete all the triton gemm stuff once the fix in core lands.

danielvegamyhre (approved):
LGTM with some minor comments

torchao/prototype/blockwise_fp8_training/kernels.py
iamzainhuda: yeah, this is to guard against different torch versions. older pytorch has torch._C._ScalingType and newer has torch.nn.functional.ScalingType

torchao/prototype/blockwise_fp8_training/kernels.py
iamzainhuda: i'm unsure if this is because of TMA but i've updated it to add more detail directly from NVIDIA's cuBLAS docs:

```
CUBLASLT_MATMUL_MATRIX_SCALE_BLK128x128_32F scaling mode, the scaling factors are 
-major and the stride between the consecutive columns must be a multiple of 4. Let 
, where the denotes rounding up to the nearest multiple of 4. 
 ```

reference: 
https://docs.nvidia.com/cuda/cublas/index.html#scaling-factors-layouts

torchao/prototype/blockwise_fp8_training/kernels.py
iamzainhuda: yup agreed and changed to assertions. oversight on my part where it's unclear to user/hurts perf

torchao/prototype/blockwise_fp8_training/linear.py
iamzainhuda: yeah, agreed its not great. the triton and scaled_mm paths expect the RHS scale tensor in different physical layouts for the grad_weight case. the triton 1x128_128x1 kernel consumes b_s in row-major layout whereas scaled_mm expects the transposed scale layout, so with the arg we let each path receive the layout it expects without adding a copy/transpose on the hot path

i believe the changes in core will be landed soon so we can get rid of the triton gemm work when it does. https://github.com/pytorch/pytorch/pull/180668

danielvegamyhre: ok sounds good. please make sure this is explained in the docstring somewhere.

iamzainhuda: yup added a comment explaining it

Issue #1912, opened 2025-03-17
placeholder for tracking saturated cast logic to float8_e4m3fn


vkuzo:
Today the cast to `torch.float8_e4m3fn` is unsaturated in PyTorch eager mode, and satured in triton.  For now creating a placeholder to improve this, and we might move this to PyTorch core after some discussions.

PR #4297, opened 2026-04-19
fix(utils): propagate non_blocking in TorchAOBaseTensor._to_copy and _get_to_kwargs

Dev-next-gen:
## Problem

`_get_to_kwargs` explicitly discarded the `non_blocking` argument parsed from `torch._C._nn._parse_to`, with a comment saying it is "not very useful for most tensor subclasses". As a result, any call to `tensor.to(device, non_blocking=True)` on a `TorchAOBaseTensor` subclass silently became a blocking transfer at the inner-tensor level.

This matters in practice for async CPU→GPU offloading workflows such as `diffusers` `enable_group_offload(use_stream=True)`: the diffusers hook schedules copies with `non_blocking=True` so that the transfer stream and the compute stream can overlap. Because the flag was dropped, all copies became blocking, negating the overlap benefit.

On AMD ROCm (gfx1xxx) the missing `non_blocking` also interacts with a separate stream-ordering race (fixed in huggingface/diffusers#13502): the default stream can race ahead of "blocking" copies that the OS scheduler hasn't committed yet, producing device-mismatch errors in the first matmul.

## Fix

1. `_get_to_kwargs`: include `non_blocking` in the returned kwargs dict.
2. `TorchAOBaseTensor._to_copy.default`: pop `non_blocking` from kwargs and forward it to every inner `.to()` call for both `tensor_data_names` and `optional_tensor_data_names`.

The change is backward-compatible: when `non_blocking=False` (the default), behaviour is identical to before.

## Tested on

| GPU | 5× AMD RX 7800 XT (gfx1101) |
| ROCm | 7.1 |
| PyTorch | 2.7 |
| Model | FLUX.1-dev, `Int8WeightOnlyConfig` via torchao |
| Config | block-level group offload + `use_stream=True` (diffusers) |

## Related

- huggingface/diffusers#13502 — companion stream sync fix on the diffusers side
- Reproducer and full benchmark: https://github.com/Dev-next-gen/flux-amd-rocm

### DIFF
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -56,6 +56,39 @@ class TestTorchAOBaseTensor(unittest.TestCase):
         with self.assertRaisesRegex(NotImplementedError, "arg_types"):
             l.weight = torch.nn.Parameter(MyTensor(l.weight))
 
+    def test_get_to_kwargs_non_blocking(self):
+        """Verify _get_to_kwargs parses and returns the non_blocking flag."""
+
+        class MyTensor(TorchAOBaseTensor):
+            tensor_data_names = ["qdata"]
+            tensor_attribute_names = ["attr", "device"]
+
+            def __new__(cls, qdata, attr="attr", device=None):
+                if device is None:
+                    device = qdata.device
+                kwargs = {"device": device, "dtype": qdata.dtype}
+                r = torch.Tensor._make_wrapper_subclass(cls, qdata.shape, **kwargs)
+                r.qdata = qdata
+                r.attr = attr
+                return r
+
+            def __init__(self, qdata, attr="attr", device=None):
+                pass
+
+        t = MyTensor(torch.randn(4, 4))
+
+        # non_blocking=True is preserved
+        kwargs = t._get_to_kwargs(device="cpu", non_blocking=True)
+        self.assertTrue(kwargs["non_blocking"])
+
+        # non_blocking=False (explicit) is preserved
+        kwargs = t._get_to_kwargs(device="cpu", non_blocking=False)
+        self.assertFalse(kwargs["non_blocking"])
+
+        # default (not specified) → False
+        kwargs = t._get_to_kwargs(device="cpu")
+        self.assertFalse(kwargs["non_blocking"])
+
     def _test_default_impls_helper(self, lp_tensor, lp_tensor_for_copy):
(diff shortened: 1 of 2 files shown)


Dev-next-gen:
Hi @jerryzh168 @andrewor14 — gentle ping 🙏

PR is ~9+ days old, CLA signed (confirmed by meta-cla bot), but the CI workflows haven't run
yet — they're stuck in "awaiting approval" state for first-time contributors (visible on
https://hud.pytorch.org/pr/pytorch/ao/4297 — all jobs in U/Workflow Startup Failure since open).
  
Could a maintainer kick off the workflow approval whenever convenient? It's a small fix
(+10/-10 LOC, 1 file) that propagates `non_blocking` through `TorchAOBaseTensor._to_copy` /
`_get_to_kwargs`. Backward-compatible (no-op when `non_blocking=False`).
  
Companion PR on diffusers side: huggingface/diffusers#13502 — together they unblock async
CPU↔GPU offloading for FLUX.1-dev int8 on AMD ROCm.
  
Happy to iterate on review feedback once CI is green. Thanks!

Dev-next-gen:
@jerryzh168 @vkuzo quick note for whoever reviews: the H100 failure (`test_cast_to_float8_e4m3fn_saturation_behavior` in `test/prototype/mx_formats/test_mx_tensor.py:701`) appears unrelated to this PR. 
The test has an inline `# TODO(#1912): make the saturated cast work in eager mode and remove this test` — a recent torch nightly likely resolved #1912, breaking this now-obsolete guard. This PR doesn't touch `mx_formats`. 
Happy to file a separate issue if useful.

jerryzh168 (commented):
looks good, can you add a test for this?

Dev-next-gen:
@jerryzh168 added the requested test in `test/test_utils.py::TestTorchAOBaseTensor::test_get_to_kwargs_non_blocking` — covers `non_blocking=True`, `non_blocking=False`, and default (unspecified) cases. CPU-only, runs in every matrix slot. CI is back in `awaiting workflow approval` after the new push — could you re-trigger when you review?

jerryzh168 (approved):

Dev-next-gen:
@jerryzh168 thanks for the quick turnaround on the review! Let me know if anything else is needed from my side to land this.

PR #4302, opened 2026-04-20
hook up gptq prototype to nvfp4

vkuzo:
Summary:

For now, this is a numerical reference which hooks up nvfp4 and verifies
that a minimal unit test (random data + toy model) works as expected,
NVFP4 + GPTQ loss is significanly lower than baseline loss.

Future TODOs:
1. validate on e2e model
2. optimize dense performance
3. add moe support (will require custom fwd to ensure every expert sees
   calibration data)

Test Plan:

```
> pytest test/prototype/gptq/ -s -k nvfp4
...
test/prototype/gptq/test_gptqv2.py GPTQ loss: 0.1582, Naive loss: 0.9259
```

### DIFF
--- a/torchao/prototype/gptq/api.py
+++ b/torchao/prototype/gptq/api.py
@@ -319,6 +319,7 @@ def gptq_quantize(H: torch.Tensor, W_t: torch.Tensor, config: GPTQConfig):
         block_size = list(block_size)
         group_size = block_size[-1]
     elif isinstance(base_config, NVFP4DynamicActivationNVFP4WeightConfig):
+        assert base_config.use_dynamic_per_tensor_scale, "unsupported"
         group_size = 16
         block_size = [1, group_size]
         # for per-tensor nvfp4, we need to calculate the global scale over the
--- a/torchao/prototype/gptq/api.py
+++ b/torchao/prototype/gptq/api.py
@@ -19,6 +19,25 @@ except:
     pack_int4 = None
 
 from torchao.core.config import AOBaseConfig
+from torchao.prototype.mx_formats.constants import F4_E2M1_MAX
+from torchao.prototype.mx_formats.inference_workflow import (
+    NVFP4DynamicActivationNVFP4WeightConfig,
+)
+from torchao.prototype.mx_formats.kernels import (
+    f4_unpacked_to_f32,
+    f32_to_f4_unpacked,
+    pack_uint4,
+)
+from torchao.prototype.mx_formats.nvfp4_tensor import (
+    NVFP4Tensor,
+    QuantizeTensorToNVFP4Kwargs,
+    nvfp4_quantize,
+    per_tensor_amax_to_scale,
+)
+from torchao.prototype.mx_formats.utils import (
+    hp_data_dims_to_swizzled_scale_dims_nvfp4,
+    to_blocked,
+)
 from torchao.quantization import Int4Tensor, Int8Tensor
 from torchao.quantization.granularity import PerRow
 from torchao.quantization.quant_api import (
@@ -33,6 +52,7 @@ from .observer import GPTQObserverTensor
 CONFIG_TO_TORCHAO_BASE_TENSOR = {
     Int4WeightOnlyConfig: Int4Tensor,
     Int8WeightOnlyConfig: Int8Tensor,
+    NVFP4DynamicActivationNVFP4WeightConfig: NVFP4Tensor,
 }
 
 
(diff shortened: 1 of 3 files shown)


vkuzo:
* #4313
* #4310
* #4303
* __->__ #4302

torchao/prototype/gptq/gptq_example.py
```diff
@@ -230,7 +230,8 @@ def main():
 
     # Generate output directory name from args
     model_name = args.model_id.split("/")[-1]  # Get last part of model ID
-    output_dir = f"{model_name}_{args.quantization}"
+    # TODO(before land): make output dir configurable
```
vkuzo: all these changes should be in the next PR, will fix before review/land

torchao/prototype/gptq/api.py
```diff
@@ -385,6 +498,48 @@ def gptq_quantize(H: torch.Tensor, W_t: torch.Tensor, config: GPTQConfig):
...
+        scale_N, scale_K = hp_data_dims_to_swizzled_scale_dims_nvfp4(N, K)
+        combined_scale_swizzled = combined_scale_swizzled.view(scale_N, scale_K)
+
+        result = NVFP4Tensor(
+            qdata,
+            combined_scale_swizzled,
+            block_size=group_size,
+            orig_dtype=W_t.dtype,
+            per_tensor_scale=nvfp4_global_scale,
+            # TODO(future): get act_per_tensor_scale from calibration data?
```
jerryzh168, line 536: it's fine we don't have this use case yet I think

torchao/prototype/gptq/api.py
```diff
@@ -385,6 +498,48 @@ def gptq_quantize(H: torch.Tensor, W_t: torch.Tensor, config: GPTQConfig):
...
+        combined_scale_swizzled = to_blocked(combined_scale).flatten()
+        scale_N, scale_K = hp_data_dims_to_swizzled_scale_dims_nvfp4(N, K)
+        combined_scale_swizzled = combined_scale_swizzled.view(scale_N, scale_K)
+
+        result = NVFP4Tensor(
+            qdata,
+            combined_scale_swizzled,
+            block_size=group_size,
+            orig_dtype=W_t.dtype,
+            per_tensor_scale=nvfp4_global_scale,
```
jerryzh168, line 535: although we do have use case of nvfp4_global_scale being generated from static quantization (can be added later)

torchao/prototype/gptq/api.py
```diff
@@ -241,6 +318,15 @@ def gptq_quantize(H: torch.Tensor, W_t: torch.Tensor, config: GPTQConfig):
         block_size = get_block_size(W_t.shape, base_config.granularity)
         block_size = list(block_size)
         group_size = block_size[-1]
+    elif isinstance(base_config, NVFP4DynamicActivationNVFP4WeightConfig):
+        group_size = 16
+        block_size = [1, group_size]
+        # for per-tensor nvfp4, we need to calculate the global scale over the
+        # entire tensor before we enter the GPTQ loop
+        tensor_amax = torch.max(torch.abs(W_t))
+        nvfp4_global_scale = per_tensor_amax_to_scale(tensor_amax)
```
jerryzh168, line 328: do we need to assert that `use_dynamic_per_tensor_scale` is True here?

jerryzh168:
@claude review

torchao/prototype/gptq/api.py
```diff
@@ -179,6 +203,59 @@ def _int4_row_dequantize_zp(
     return torch.cat(dequant_chunks, dim=-1)
 
 
+def _nvfp4_with_precalculated_scales_qdq(
```
jerryzh168, line 206: should this share code with `_nvfp4_with_precalculated_scales_q` so it's less error prone? they should use the same quantization code right

jerryzh168 (approved):
LGTM, please see comments inline

torchao/prototype/gptq/api.py
vkuzo: the mathematical code is pretty simple and the shape operations are different (qdq is input always shaped [N, 1] and the q input is always shaped [N, k_slice], which requires different broadcasting behavior), I think it's ok to keep them separate for simplicity

jerryzh168: sounds OK to keep separate for readability, my main worry was about the consistency of the two implementations, is there a test to make sure these two code paths have the same quantization code?

vkuzo: I agree ^ is useful, I'm just punting it until later.  There is a TODO in the code to track.  The numerical tests we already have would also capture any divergence indirectly.

PR #4306, opened 2026-04-21
[mxfp8 moe training] integrate 1x32 cutedsl quantization kernel

danielvegamyhre:
## Summary
- Use cutedsl mxfp8 quantization kernel for 1x32 scaling of 2d tensors in `_compute_fwd` and `_compute_dgrad`

## Other changes
- Remove pad_token_groups_for_grouped_mm test case - to be deprecated since we are now validating token group sizes are a multiple of 128 (no padding needed)
- Update tests to use `multiple_of=128`, per above (quantization kernels validate group sizes)
- Remove extraneous extra "__" that should be single "_" in some method names
- Remove `_extract_or_quantize_dim0_auto` and just inline the logic, it was obfuscating things:

## Tests
- `pytest test/prototype/moe_training/ -v -s`

## Benchmarks
Planning to do after integrating the 32x1 kernel next

### DIFF
--- a/torchao/prototype/moe_training/mxfp8_grouped_mm.py
+++ b/torchao/prototype/moe_training/mxfp8_grouped_mm.py
@@ -550,10 +550,6 @@ def _compute_fwd_sm100(
             scaling_mode=scale_calculation_mode.value.lower(),
             offs=padded_group_end_offsets,
         )
-        # TODO: cutedsl shouldn't flatten scales
-        input_act_scales_blocked = input_act_scales_blocked.view(
-            input_act_e4m3.shape[0], -1
-        )
 
     # Quantize weights along dim0 (after transposing from (E, K, N) to (E, N, K))
     weight_e4m3, weight_scales = triton_to_mxfp8_dim0(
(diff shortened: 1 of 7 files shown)


torchao/prototype/moe_training/mxfp8_grouped_mm.py
```diff
@@ -558,19 +532,32 @@ def _compute_fwd_sm100(
...
+                padded_input_act.scales, padded_group_end_offsets
+            )
+    else:
+        input_act_e4m3, input_act_scales_blocked = mxfp8_quantize_2d_1x32_cutedsl(
+            padded_input_act,
+            scaling_mode=scale_calculation_mode.value.lower(),
+            offs=padded_group_end_offsets,
+        )
+        # TODO: cutedsl shouldn't flatten scales
+        input_act_scales_blocked = input_act_scales_blocked.view(
```
drisspg, line 554: shoudl we just fix here?

drisspg (approved):

torchao/prototype/moe_training/mxfp8_grouped_mm.py
danielvegamyhre: Ok, fixed both 1x32 and 32x1 variants to return 2D scales aligned with the shape torch._scaled_grouped_mm expects for consumption. Updated tests accordingly.

PR #4307, opened 2026-04-21
[PT2E][Docs] Calibration is a must for both static and dynamic quantization

Xia-Weiwen:
This pull request adds a note in the PT2E docs stating that the calibration is mandatory.

### DIFF
--- a/docs/source/pt2e_quantization/index.rst
+++ b/docs/source/pt2e_quantization/index.rst
@@ -52,7 +52,7 @@ API Example::
   m = prepare_pt2e(m, quantizer)
 
   # run calibration
-  # calibration is a must for both static and dynamic quantization
+  # must run calibration for both static and dynamic quantization
   # calibrate(m, sample_inference_data)
 
   m = convert_pt2e(m)
--- a/docs/source/pt2e_quantization/index.rst
+++ b/docs/source/pt2e_quantization/index.rst
@@ -52,7 +52,9 @@ API Example::
   m = prepare_pt2e(m, quantizer)
 
   # run calibration
+  # calibration is a must for both static and dynamic quantization
   # calibrate(m, sample_inference_data)
+
   m = convert_pt2e(m)
 
   # Step 3. lowering
--- a/torchao/quantization/pt2e/convert.py
+++ b/torchao/quantization/pt2e/convert.py
@@ -203,32 +203,6 @@ def _replace_observer_with_quantize_dequantize_node_decomposed(
         # 1. extract information for inserting q/dq node from activation_post_process
         node_type = "call_function"
         quantize_op: Optional[Callable] = None
-
-        # In the dynamic quantization case, weight's is_dynamic flag is false, but calibration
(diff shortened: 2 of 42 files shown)


torchao/quantization/pt2e/convert.py
```diff
@@ -203,6 +203,32 @@ def add_quantize_dequantize_node_info(qdq_node, original_node):
         # 1. extract information for inserting q/dq node from activation_post_process
         node_type = "call_function"
         quantize_op: Optional[Callable] = None
+
+        # In the dynamic quantization case, weight's is_dynamic flag is false, but calibration
+        # is skipped so weight observers may never have been run. Detect this by checking
+        # if the observer is uninitialized (empty min_val) and its input is a constant
+        # weight tensor, and run the observer eagerly in that case.
+        if (
+            hasattr(activation_post_process, "min_val")
```
jerryzh168, line 212: this condition is too ad hoc I think, won't work if `activation_post_process` is more general

Xia-Weiwen: Oh. It checks if the observer has not been run and its argument is a constant (from getattr). It looks complicated but I did not figure out a better way. Do you have any suggestions 😂 Thanks.

Xia-Weiwen: Hi @jerryzh168 Do you have any good ideas to simplify this or to fix the issue in another way? Thanks.

jerryzh168: what is the use case for dynamic quantization? why is just going through an example input for dynamic quant flow not good? I think we need example input for torch.export as well anyways

jerryzh168: I feel we could also just add some asserts saying observer should always run, or just make sure we have proper docs saying calibration step can't be skipped

Xia-Weiwen: Hi @jerryzh168 Thanks for the suggestions. We already have a warning here: https://github.com/pytorch/ao/blob/67a78e5814ceac00fa2a82514511b2f728fca3bd/torchao/quantization/pt2e/utils.py#L148
So, I just add a comment in the docs to say calibration is a must for dynamic quantization. Please take a look.

Xia-Weiwen:
Hi @jerryzh168 Could you please review again? Thanks.

docs/source/pt2e_quantization/index.rst
```diff
@@ -52,7 +52,9 @@ API Example::
   m = prepare_pt2e(m, quantizer)
 
   # run calibration
+  # calibration is a must for both static and dynamic quantization
```
jerryzh168: nit: `must run calibration for both static and dynamic quantization`

jerryzh168 (approved):
thanks, can change the comment to be a bit more natural

Xia-Weiwen:
> thanks, can change the comment to be a bit more natural

Updated. Thanks.

PR #4315, opened 2026-04-22
make NVFP4Tensor handle per-expert outer scale

vkuzo:
Summary:

NVFP4 MoE kernels usually do per-expert outer scale.
This PR extends `NVFP4Tensor` to support this.

Test Plan:

```
pytest test/prototype/mx_formats/test_nvfp4_tensor.py -s
```

### DIFF
--- a/torchao/prototype/mx_formats/nvfp4_tensor.py
+++ b/torchao/prototype/mx_formats/nvfp4_tensor.py
@@ -104,6 +104,11 @@ class NVFP4Tensor(TorchAOBaseTensor):
             requires_grad=False,
         )
 
+        if per_tensor_scale is not None:
+            # 0: per-tensor scale, with shape (,)
+            # 3: per-expert scale, with shape (E, 1, 1)
+            assert len(per_tensor_scale.shape) in (0, 3), "unsupported"
+
         self.qdata = qdata
         self.scale = scale
         self.block_size = block_size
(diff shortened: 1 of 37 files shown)


vkuzo:
* #4321
* #4320
* #4319
* #4316
* __->__ #4315

torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -224,7 +229,7 @@ def dequantize(self, output_dtype: Optional[torch.dtype] = None) -> torch.Tensor
         return result
 
     def get_hp_scales(self) -> torch.Tensor:
-        """Get the scales of the NVFP4Tensor in original dtype.
+        """Get the scales of the NVFP4Tensor in float32.
```
jerryzh168, line 232: not for this PR: I just checked the usage of `self._orig_dtype` seems it's duplicated with `self.dtype` now

jerryzh168 (approved):
would be good to add a comment for attribute `per_tensor_scale` for NVFP4Tensor I think, also probably try to restrict this to weight only since it doesn't apply to activations?

vkuzo:
> would be good to add a comment for attribute per_tensor_scale for NVFP4Tensor I think

sure, fixed!

> also probably try to restrict this to weight only since it doesn't apply to activations?

it could apply to activations (per-token-group)

PR #4324, opened 2026-04-23
fix(fp8_fa3): remove broken import and fix setup_fp8_fa3 signature

### LINKED ISSUE #4318: Import of nonexistent config module crashes fp8_fa3/setup.py
### Description

# Missing config module breaks fp8_fa3/setup.py

## What happened

I was trying to import `setup_fp8_fa3` directly and hit this:

```
ModuleNotFoundError: No module named 'torchao.prototype.attention.config'
```

Turns out `torchao/prototype/attention/fp8_fa3/setup.py` is importing from a module that doesn't exist:

```python
from torchao.prototype.attention.config import LowPrecisionAttentionConfig
```

There is no `config.py` (or `config/` package) anywhere under `torchao/prototype/attention/`. The class `LowPrecisionAttentionConfig` was never defined.

## Impact

`setup_fp8_fa3` is completely broken -- you can't even import it. The function signature references a type that doesn't exist:

```python
def setup_fp8_fa3(
    model: nn.Module,
    config: LowPrecisionAttentionConfig,  # <-- undefined
) -> nn.Module:
```

However, the public API (`apply_low_precision_attention`) is fine. It calls `shared_utils/setup.py:setup_fp8_backend()` directly, passing a plain string `"FA3"` for the backend name, so the broken function is never reached through normal usage.

## Affected code

| File | Line | Problem |
|---|---|---|
| `fp8_fa3/setup.py` | 11 | `from torchao.prototype.attention.config import LowPrecisionAttentionConfig` -- module doesn't exist |
| `fp8_fa3/setup.py` | 17 | `config: LowPrecisionAttentionConfig` parameter type is undefined |

## How to fix

Two reasonable options:

**1. Create the missing config module**

Add `torchao/prototype/attention/config.py` defining `LowPrecisionAttentionConfig` and have `setup_fp8_fa3` translate the config into `setup_fp8_backend` arguments. This was probably the original intent -- the config was planned but never implemented.

or 

**2. Remove setup_fp8_fa3 entirely**

Since `apply_low_precision_attention` already handles the FA3 path through `setup_fp8_backend`, and `setup_fp8_fa3` is the only consumer of this nonexistent config, we can just remove `setup_fp8_fa3` and stop exporting it from `fp8_fa3/__init__.py`.

1 makes sense if there's a plan to add more backend-specific configuration later.

### Steps to reproduce

```python
## Steps to reproduce


from torchao.prototype.attention.fp8_fa3.setup import setup_fp8_fa3
# ModuleNotFoundError: No module named 'torchao.prototype.attention.config'
```

### Environment

```text
- torch: 2.11.0+cpu
- torchao: 0.18.0+gitde873d15
- Hardware: RTX 4060 
- OS: CachyOS Linux 7.0.0-1 (x86_64)
```

Anai-Guo:
## Problem

Closes #4318.

`torchao/prototype/attention/fp8_fa3/setup.py` contains two bugs that make `setup_fp8_fa3` completely unusable:

**1. Import from a non-existent module:**
```python
from torchao.prototype.attention.config import LowPrecisionAttentionConfig
```
There is no `config.py` or `config/` package under `torchao/prototype/attention/`. Importing `setup_fp8_fa3` raises `ModuleNotFoundError` immediately.

**2. Incorrect call to `setup_fp8_backend`:**
```python
# OLD (broken)
return setup_fp8_backend(
    model,
    config,               # should be flash_impl_name: str
    flash_impl_name="FA3",
    sdpa_fn=fp8_fa3_sdpa, # no such parameter
)
```
The actual signature is `setup_fp8_backend(model, flash_impl_name, hadamard="NONE")`.

Note: the public `apply_low_precision_attention` in `api.py` is unaffected — it calls `setup_fp8_backend` correctly with `"FA3"` directly.

## Fix

- Remove the broken `LowPrecisionAttentionConfig` import
- Update `setup_fp8_fa3` to accept `hadamard: str = "NONE"` (matching `setup_fp8_backend`)
- Fix the call to `setup_fp8_backend` to use the correct signature

## Before / After

```python
# BEFORE (broken)
from torchao.prototype.attention.config import LowPrecisionAttentionConfig

def setup_fp8_fa3(model: nn.Module, config: LowPrecisionAttentionConfig) -> nn.Module:
    return setup_fp8_backend(model, config, flash_impl_name="FA3", sdpa_fn=fp8_fa3_sdpa)

# AFTER (fixed)
def setup_fp8_fa3(model: nn.Module, hadamard: str = "NONE") -> nn.Module:
    return setup_fp8_backend(model, flash_impl_name="FA3", hadamard=hadamard)
```

### DIFF
--- a/test/prototype/attention/test_fp8_attention.py
+++ b/test/prototype/attention/test_fp8_attention.py
@@ -6,6 +6,7 @@
 
 """Tests for FP8 low-precision attention (FA3 backend on Hopper)."""
 
+import inspect
 import unittest
 
 import torch
@@ -443,6 +444,38 @@ class TestHadamardAccuracy(TestCase):
                 f"had={err_had.item():.6f} vs none={err_none.item():.6f} "
                 f"for shape={shape}, dtype={dtype}",
             )
+
+
+class TestFP8FA3SetupRegression(TestCase):
+    """Regression coverage for the broken import + signature in
+    ``torchao/prototype/attention/fp8_fa3/setup.py`` (issue #4318).
+
+    ``setup_fp8_fa3`` previously imported a non-existent
+    ``torchao.prototype.attention.config`` module, so importing it raised
+    ``ModuleNotFoundError`` before any FA3 hardware path was reached, and its
+    signature did not match ``setup_fp8_backend``. Both checks are
+    hardware-independent and run on CPU.
+    """
+
+    @unittest.skipUnless(
+        torch_version_at_least("2.11.0"),
+        "fp8_fa3 setup imports torch.nn.attention helpers added in torch 2.11",
+    )
+    def test_setup_fp8_fa3_imports_without_error(self):
+        from torchao.prototype.attention.fp8_fa3.setup import setup_fp8_fa3
+
+        self.assertTrue(callable(setup_fp8_fa3))
+
+    @unittest.skipUnless(
+        torch_version_at_least("2.11.0"),
+        "fp8_fa3 setup imports torch.nn.attention helpers added in torch 2.11",
+    )
+    def test_setup_fp8_fa3_signature_matches_backend(self):
+        from torchao.prototype.attention.fp8_fa3.setup import setup_fp8_fa3
+
+        params = inspect.signature(setup_fp8_fa3).parameters
+        self.assertEqual(list(params), ["model", "hadamard"])
+        self.assertEqual(params["hadamard"].default, "NONE")
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


Anai-Guo:
Added a small CPU regression test (`TestFP8FA3SetupRegression` in `test/prototype/attention/test_fp8_attention.py`) that would have caught this: it imports `setup_fp8_fa3` (previously `ModuleNotFoundError` from the non-existent `...attention.config` module) and asserts the signature matches `setup_fp8_backend` — `(model, hadamard="NONE")`. Both checks are hardware-independent (no Hopper/FA3 required), guarded only by `torch_version_at_least("2.11.0")`. `ruff check` + `ruff format --check` pass.

PR #4325, opened 2026-04-23
[fp8 training] support linear op overrides in Float8TrainingWeightWrapperTensor

danielvegamyhre:
Stacked PRs:
 * __->__#4325

--- --- ---

## Summary
- `torchao.float8` module swap design is interfering with some torchtitan refactor/design goals. We would like to migrate to the "weight wrapper tensor subclass" approach asap. MXFP8 training already works this way for both linear and grouped_mm ops, but for float8 training, it only is supported for grouped_mm. This PR adds support for linears.

## Tests
- Add test cases in `test_tensor.py`

### DIFF
--- a/torchao/prototype/moe_training/config.py
+++ b/torchao/prototype/moe_training/config.py
@@ -6,12 +6,14 @@
 
 from dataclasses import dataclass
 from enum import Enum
-from typing import Optional
+from typing import Literal, Optional
 
 import torch
 from torch import nn
 
 from torchao.core.config import AOBaseConfig
+from torchao.float8.config import Float8LinearConfig
+from torchao.float8.float8_training_tensor import LinearMMConfig, ScaledMMConfig
 from torchao.prototype.mx_formats.config import ScaleCalculationMode
 from torchao.quantization.quantize_.common import KernelPreference
 from torchao.quantization.transform_module import register_quantize_module_handler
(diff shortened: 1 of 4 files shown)


torchao/prototype/moe_training/config.py
```diff
@@ -64,6 +66,39 @@ class Float8TrainingOpConfig(TrainingOpBaseConfig):
     # causes a D2H sync that breaks torch.compile.
     pad_token_groups_for_grouped_mm: bool = False
 
+    # Recipe for the float8 linear op override ("tensorwise" or "rowwise").
+    float8_linear_recipe: Literal["tensorwise", "rowwise"] = "rowwise"
```
danielvegamyhre: note for reviewer: to simplify the config, i chose to just have a simple recipe name, and the `__post_init__` does the plumbing to assemble the necessary configs for the autograd function. 

for more control, the user/caller can still modify the `config._float8_linear_config` object as they wish.

vkuzo: support `rowwise_with_grad_weight_in_hp` too

vkuzo (approved):

torchao/prototype/moe_training/config.py
danielvegamyhre: Updated

PR #4326, opened 2026-04-24
Fix int8 dynamic activation quantization accuracy regression from v2 tensor migration

htzo:
The v2 int8 dynamic activation quantization path introduced in #4151 uses `hp_tensor.dtype` as `scale_dtype` in `choose_qparams_affine`. When the input tensor is bfloat16, this computes quantization scales in bfloat16 precision instead of float32, causing ~0.6 dB SQNR degradation and measurable accuracy loss on downstream tasks (e.g. MMLU on Llama-3.1-8B-Instruct).

This PR replaces `hp_tensor.dtype` with configurable `scale_dtype` so it is configurable per-platform.

## Regression Cause

The original code path hardcoded `scale_dtype=torch.float32`. The refactor in #4151 changed this to `hp_tensor.dtype`, which for bfloat16 models means scale computation happens at reduced precision. Since bfloat16 has only 8 bits of mantissa (vs float32's 24), the quantization scales lose precision, degrading output quality.

## Validation

`Llama-3.1-8B-Instruct` restores top-1 accuracy in tests to v1.

No built-in tests should be affected.

### DIFF
--- a/torchao/quantization/quantize_/workflows/int8/int8_tensor.py
+++ b/torchao/quantization/quantize_/workflows/int8/int8_tensor.py
@@ -43,10 +43,13 @@ class QuantizeTensorToInt8Kwargs(QuantizeTensorKwargs):
     Args:
         granularity: the granularity for the Tensor, currently either PerRow() or PerTensor()
         mapping_type: whether to use symmetric or asymmetric quant
+        scale_dtype: dtype for quantization scale computation. Defaults to torch.float32.
+            Using a lower-precision dtype (e.g. bfloat16) may reduce accuracy.
     """
 
     granularity: Granularity
     mapping_type: MappingType = MappingType.SYMMETRIC
+    scale_dtype: Optional[torch.dtype] = torch.float32
 
 
 class Int8Tensor(TorchAOBaseTensor):
@@ -170,6 +173,7 @@ class Int8Tensor(TorchAOBaseTensor):
         scale: Optional[torch.Tensor] = None,
         zero_point: Optional[torch.Tensor] = None,
         act_quant_kwargs: Optional[QuantizeTensorToInt8Kwargs] = None,
+        scale_dtype: Optional[torch.dtype] = torch.float32,
         act_quant_scale: Optional[torch.Tensor] = None,
         act_quant_zero_point: Optional[torch.Tensor] = None,
         act_pre_scale: Optional[torch.Tensor] = None,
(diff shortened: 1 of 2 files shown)


Xia-Weiwen (approved):

Xia-Weiwen:
@claude review

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -43,10 +43,13 @@ class QuantizeTensorToInt8Kwargs(QuantizeTensorKwargs):
     Args:
         granularity: the granularity for the Tensor, currently either PerRow() or PerTensor()
         mapping_type: whether to use symmetric or asymmetric quant
+        scale_dtype: dtype for quantization scale computation. Defaults to torch.float32.
+            Using a lower-precision dtype (e.g. bfloat16) may reduce accuracy.
     """
 
     granularity: Granularity
     mapping_type: MappingType = MappingType.SYMMETRIC
+    scale_dtype: Optional[torch.dtype] = torch.float32
```
Xia-Weiwen: To keep the original bahavior default, the default value here is better to be None and when it's None, scale_dtype = hp_tensor.dtype.

Xia-Weiwen (commented):
Users don't call `from_hp` directly. How do users set this flag in their script?

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
jerryzh168: I don't think we want to add this to API, it can just be hidden the same way as before:

https://raw.githubusercontent.com/pytorch/ao/662c670db64da7c3c0a66d626a40311125838316/torchao/quantization/quant_api.py

we can also put up a separate PR to hardcode this to float32 and check if it's fine as well

htzo: Simplified the patch to a one-line hardcode of `scale_dtype=torch.float32`, matching the old behavior without a new API. It seems to be working - accuracy is fine, and tests are passing.

andrewor14:
@claude review

andrewor14:
Hi @haotongzou or @claude, can you point me to the original code before #4151 that used `scale_dtype` as torch.float32?

jerryzh168 (approved):
please rebase

Xia-Weiwen (approved):

PR #4327, opened 2026-04-24
nvfp4 gptq for bmm

vkuzo:
Summary:

Extend GPTQ coverage for bmm, formulating the bmm as a 3d case of mm.
This involves:
1. refactoring the 2d code to make it easily extendable to 3d
2. the existing bmm logic was numerically incorrect (used a single
   hessian), modify it to instead use E K by K Hessians for an `E, N, K`
   input shape, route to the 2D hessian logic E times. This is slow
   but we can optimize later.

We test numerical correctness by bitwise matching E 2d hessian
calculations to the 3D one.

Test Plan:

```
torchao/prototype/gptq/gptq_nvfp4_llama3_2_1b_nonsequential_wikitext.sh
// gptq accuracy unchanged
```

### DIFF
--- a/test/prototype/gptq/test_gptqv2.py
+++ b/test/prototype/gptq/test_gptqv2.py
@@ -102,8 +102,10 @@ class TestGPTQObserverTensor:
         # Check hp_data is stored correctly
         torch.testing.assert_close(observer.hp_data, weight)
 
-        # Check hessian is initialized as None
-        assert observer.hessian is None
+        # Check hessian is initialized as zeros
+        assert torch.equal(
+            observer.hessian, torch.zeros(64, 64, dtype=torch.float32, device="cuda")
+        )
 
         # Check total_batches is initialized as 0
         assert observer.total_batches == 0
@@ -120,7 +122,9 @@ class TestGPTQObserverTensor:
 
         # Test hessian attribute
         assert hasattr(observer, "hessian")
-        assert observer.hessian is None
+        assert torch.equal(
+            observer.hessian, torch.zeros(32, 32, dtype=torch.float32, device="cuda")
+        )
 
         # Test total_batches attribute
         assert hasattr(observer, "total_batches")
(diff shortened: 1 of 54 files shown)


vkuzo:
* #4331
* #4330
* #4329
* #4328
* __->__ #4327

test/prototype/gptq/test_gptqv2.py
```diff
@@ -108,31 +108,7 @@ def test_observer_tensor_creation(self):
         )
 
         # Check total_batches is initialized as 0
-        assert observer.total_batches == 0
-
-    @pytest.mark.skipif(not torch.cuda.is_available(), reason="Need CUDA available")
-    def test_observer_tensor_attributes(self):
```
vkuzo, line 114: this test is not that useful, deleting instead of updating with new contents of the observer tensor

torchao/prototype/gptq/observer.py
```diff
@@ -11,54 +11,92 @@
...
+    def __new__(cls, hp_data: torch.Tensor, total_batches, hessian=None):
         shape = hp_data.shape
         kwargs = {}
         kwargs["device"] = hp_data.device
         kwargs["dtype"] = hp_data.dtype
         kwargs["requires_grad"] = False
         return torch.Tensor._make_wrapper_subclass(cls, shape, **kwargs)  # type: ignore[attr-defined]
 
-    def __init__(self, hp_data: torch.Tensor, total_batches: int, hessian=None):
+    def __init__(self, hp_data: torch.Tensor, total_batches, hessian=None):
```
jerryzh168, line 26: can you add some docs for `total_batches`?

jerryzh168 (approved):

torchao/prototype/gptq/observer.py
vkuzo: sure, let me do that in a future PR (planning today), as we also need to change the definition of this for grouped_mm to sample each token equally instead of sampling each batch equally

PR #4345, opened 2026-04-28
Fix symmetric quantization scale to match C++ ChooseQuantizationParams (#180990)

fantaosha:
Summary:
Fix inconsistencies between Python `_calculate_qparams` in `observer.py` and C++ `ChooseQuantizationParams` in `caffe2/aten/src/ATen/native/quantized/cpu/QuantUtils.h` for symmetric quantization.

Two changes:

1. **Scale formula:** Replace `scale = max(|min|, |max|) / 127.5` (which treats the quantization range as perfectly symmetric) with `scale = max(|min|/128, |max|/127)`, matching the C++ logic that correctly accounts for two's complement asymmetry (`symmetric_qmin=-128`, `symmetric_qmax=127` for qint8).

2. **Precision:** Promote `min_val`/`max_val` to `float64` for intermediate computation, matching the C++ code which uses `double` for scale and zero_point arithmetic, then cast `scale` back to `float32` at the end.

Together, these caused QAT training (which uses the C++ kernel via `FusedMovingAvgObsFakeQuantize`) and model conversion (which calls Python `_calculate_qparams`) to produce different quantization parameters. The affine/asymmetric path is unaffected.

X-link: https://github.com/pytorch/pytorch/pull/180990
Approved by: https://github.com/jerryzh168

Differential Revision: D101413736

### DIFF
--- a/torchao/quantization/pt2e/observer.py
+++ b/torchao/quantization/pt2e/observer.py
@@ -383,10 +383,19 @@ class UniformQuantizationObserverBase(ObserverBase):
             )
 
         quant_min, quant_max = self.quant_min, self.quant_max
+
+        # Use double precision for intermediate computation to match C++
+        # ChooseQuantizationParams in
+        # caffe2/aten/src/ATen/native/quantized/cpu/QuantUtils.h,
+        # which uses double for scale and zero_point arithmetic then
+        # casts to float at the end.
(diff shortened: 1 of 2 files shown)


jerryzh168 (commented):
could you add a test for the flag as well

PR #4347, opened 2026-04-28
Simplify granularity: accept list not tuple

cyxlily:
Simplify int8 tensor granularity by removing tuple support and accepting only list


andrewor14:
I remember we need tuples not lists for serializability, maybe @vkuzo can confirm?

jerryzh168 (approved):

PR #4362, opened 2026-05-01
Fix INT8 quantization precision loss for bf16 inputs in IntxUnpackedToInt8Tensor

mergennachin:
Cast to float32 before choose_qparams_affine and quantize_affine in
from_hp. Without this, bf16 inputs with small magnitudes (abs_mean ~0.01)
collapse all per-group scales to a single value, using only ~3% of the
int8 dynamic range.

Example: a (128, 256) bf16 weight scaled by 0.01 with group_size=32:
  Before: all scales identical, qdata in [-4, 4], SQNR ~13 dB
  After:  per-group scales preserved, qdata in [-128, 127], SQNR ~43 dB

This affects models with normalized attention projections (e.g., weights
after QK-norm) where weight magnitudes are significantly smaller than 1.0.

Affects both AFFINE and HQQ_SCALE_ONLY paths.

### DIFF
--- a/torchao/quantization/quantize_/workflows/intx/intx_unpacked_to_int8_tensor.py
+++ b/torchao/quantization/quantize_/workflows/intx/intx_unpacked_to_int8_tensor.py
@@ -216,13 +216,18 @@ class IntxUnpackedToInt8Tensor(TorchAOBaseTensor):
             )
         elif intx_choose_qparams_algorithm == IntxChooseQParamsAlgorithm.HQQ_SCALE_ONLY:
             qdata, scale = _choose_qparams_and_quantize_scale_only_hqq(
-                hp_tensor, block_size, qmin, qmax
+                hp_tensor.float(), block_size, qmin, qmax
             )
             qdata = qdata.to(torch.int8)
             zero_point = torch.zeros_like(scale, dtype=torch.int8)
         elif intx_choose_qparams_algorithm == IntxChooseQParamsAlgorithm.AFFINE:
+            # Cast to float32 for numerical precision: bf16 inputs with
+            # small magnitudes (e.g., abs_mean ~0.01) lose per-group scale
+            # granularity in bf16, collapsing all scales to a single value
+            # and wasting most of the int8 range.
+            hp_float = hp_tensor.float()
             scale, zero_point = choose_qparams_affine(
-                hp_tensor,
+                hp_float,
                 mapping_type,
                 block_size,
                 target_dtype=torch.int8,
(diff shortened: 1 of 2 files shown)


torchao/quantization/quantize_/workflows/intx/intx_unpacked_to_int8_tensor.py
```diff
@@ -216,13 +216,18 @@ def from_hp(
...
             zero_point = torch.zeros_like(scale, dtype=torch.int8)
         elif intx_choose_qparams_algorithm == IntxChooseQParamsAlgorithm.AFFINE:
+            # Cast to float32 for numerical precision: bf16 inputs with
+            # small magnitudes (e.g., abs_mean ~0.01) lose per-group scale
+            # granularity in bf16, collapsing all scales to a single value
+            # and wasting most of the int8 range.
+            hp_float = hp_tensor.float()
             scale, zero_point = choose_qparams_affine(
-                hp_tensor,
+                hp_float,
```
metascroy, line 230: If this is an underlying issue with choose_qparams_affine, it seems like it should be fixed there?

Internally this function does do float casting, so the issue you're seeing is it likely has a buggy default for eps-clamping:

https://github.com/pytorch/ao/blob/104f01a85252ac82fcac7ad689c4926bce66d61b/torchao/quantization/quant_primitives.py#L1527

The default value of:

eps = torch.finfo(scale_dtype).eps

might not be good. Claude suggests using one of these instead:

* eps = torch.finfo(scale_dtype).tiny * torch.finfo(scale_dtype).eps

* eps = torch.finfo(scale_dtype).tiny

cc @jerryzh168 for thoughts

torchao/quantization/quantize_/workflows/intx/intx_unpacked_to_int8_tensor.py
```diff
@@ -216,13 +216,18 @@ def from_hp(
             )
         elif intx_choose_qparams_algorithm == IntxChooseQParamsAlgorithm.HQQ_SCALE_ONLY:
             qdata, scale = _choose_qparams_and_quantize_scale_only_hqq(
-                hp_tensor, block_size, qmin, qmax
+                hp_tensor.float(), block_size, qmin, qmax
```
metascroy, line 219: This function already promotes to fp32 internally: https://github.com/pytorch/ao/blob/104f01a85252ac82fcac7ad689c4926bce66d61b/torchao/quantization/quant_primitives.py#L2040

Did you notice this cast make a difference for HQQ? Or was it primarily for affine?

PR #4364, opened 2026-05-01
Handle saturated eager float8 casts

iamzainhuda:
## Summary

Update the MX tensor float8 saturation test to run only on PyTorch versions where core eager casts for finite-overflow `float8_e4m3fn` values saturate instead of returning `NaN`.

- gate the test to PyTorch `2.12.0.dev0+`
- link the upstream core fix: https://github.com/pytorch/pytorch/pull/178817
- require eager and compiled/Triton casts to saturate to `448` / `-448`
- document the existing torchao saturation guards in shared float8, affine float8 quantization, and MX conversion helpers

This keeps the coverage for saturated float8 casts while avoiding compatibility code for the older broken eager behavior.

## Test Plan

- `ruff check test/prototype/mx_formats/test_mx_tensor.py`
- `ruff format --check test/prototype/mx_formats/test_mx_tensor.py`
- `pytest -q test/prototype/mx_formats/test_mx_tensor.py::test_cast_to_float8_e4m3fn_saturation_behavior`

### DIFF
--- a/test/prototype/mx_formats/test_mx_tensor.py
+++ b/test/prototype/mx_formats/test_mx_tensor.py
@@ -694,11 +694,18 @@ def test_cast_to_float8_e4m3fn_saturation_behavior():
         device="cuda",
     )
 
-    # verify that in eager mode PyTorch casting to float8 is unsaturated
+    # Eager mode used to produce NaNs for out-of-range e4m3fn casts, but newer
+    # PyTorch versions saturate to the finite max like the compiled path.
     data_in_range_f8 = data_in_range_bf16.to(torch.float8_e4m3fn)
     data_out_of_range_f8 = data_out_of_range_bf16.to(torch.float8_e4m3fn)
     assert not torch.any(torch.isnan(data_in_range_f8))
-    assert torch.all(torch.isnan(data_out_of_range_f8))
+    eager_cast_saturates = not torch.any(torch.isnan(data_out_of_range_f8))
+    if eager_cast_saturates:
+        torch.testing.assert_close(
+            data_in_range_f8, data_out_of_range_f8, atol=0, rtol=0
+        )
+    else:
+        assert torch.all(torch.isnan(data_out_of_range_f8))
 
     # verify that in triton, casting to float8 is saturated
     # for simplicity, use torch.compile to generate triton code
--- a/test/prototype/mx_formats/test_mx_tensor.py
+++ b/test/prototype/mx_formats/test_mx_tensor.py
@@ -669,9 +669,11 @@ def test_index_select():
     not is_sm_at_least_89(),
     reason="float8 in triton requires CUDA capability 8.9 or greater",
 )
+@pytest.mark.skipif(
+    not torch_version_at_least("2.12.0.dev0"),
+    reason="eager float8_e4m3fn casts saturate in PyTorch 2.12+",
+)
 def test_cast_to_float8_e4m3fn_saturation_behavior():
-    # TODO(#1912): make the saturated cast work in eager mode and remove this
-    # test
     max_val = torch.finfo(torch.float8_e4m3fn).max
 
     # create example data inside the representable range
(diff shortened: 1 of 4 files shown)


test/prototype/mx_formats/test_mx_tensor.py
```diff
@@ -694,11 +694,18 @@ def test_cast_to_float8_e4m3fn_saturation_behavior():
         device="cuda",
     )
 
-    # verify that in eager mode PyTorch casting to float8 is unsaturated
+    # Eager mode used to produce NaNs for out-of-range e4m3fn casts, but newer
```
vkuzo: do we have a link to which PR fixed this in pytorch/pytorch?

vkuzo: if this works in core now, IMO we should just gate this test to only run in the necessary PyTorch version and delete the old code, it's not worth the complexity to support the old broken behavior

iamzainhuda: fixed behaviour in https://github.com/pytorch/pytorch/pull/178817, gated test to only run on 2.12+

iamzainhuda: had codex look through for any code relying on the old finite overflow to NaN behaviour there wasn't any. it did update different code sites to document clamping behaviours and what users should expect.

danielvegamyhre:
LGTM, will let @vkuzo stamp so he can confirm his feedback was addressed

torchao/prototype/mx_formats/mx_tensor.py
```diff
@@ -363,12 +363,9 @@ def to_mx(
...
         ):
-            # As of 20250317, the Pytorch eager mode cast to `torch.float8_e4m3fn`
-            # is unsaturated. This cast is saturated in triton. If we are compute bound,
-            # we see a speedup if we remove this redundant clamp if we are compiling
-            # to triton.
-            # TODO(#1912): make the saturated cast work in eager mode and remove this
-            # workaround.
+            # Keep eager conversion saturated across PyTorch versions and FP8
+            # dtypes. Inductor/Triton casts already saturate, so compiled mode
+            # skips this redundant clamp for better fusion/codegen.
```
vkuzo: hmm, I think the previous comment was more accurate.  Maybe just modify the TODO linking to the core fix PR, and saying we can remove the workaround after we stop supporting PyTorch 2.11?

torchao/float8/float8_utils.py
```diff
@@ -119,12 +119,11 @@ def to_fp8_saturated(x: torch.Tensor, float8_dtype: torch.dtype):
     """Converts a tensor to a saturated fp8 tensor.
 
     Note:
-        The default behavior in PyTorch for casting to `float8_e4m3fn`
-        and `e5m2` is to not saturate. In this context, we should saturate.
-        A common case where we want to saturate is when the history of a
-        tensor has a maximum value of `amax1`, and the current amax value
-        is `amax2`, where `amax1 < amax2`. This is common when using delayed
-        scaling.
+        This helper clamps explicitly before casting so torchao has stable
```
vkuzo: I think the old comment was outdated and the new one doesn't seem correct. My understanding:
* in PyTorch 2.11 and older eager mode did not saturate, so we saturate here.  Compile saturated, and this extra clamp was a no-op in accuracy and regressed performance
* in PyTorch 2.12 and newer, eager mode now saturates and this clamp is no longer needed

maybe just update the comment to say that?

torchao/quantization/quant_primitives.py
```diff
@@ -239,6 +239,9 @@ def backward(ctx, gy: torch.Tensor) -> torch.Tensor:
 class _RoundToFloat8(torch.autograd.Function):
     """
     Implementation of `tensor.to(float8_dtype)` with backward STE.
+
+    Callers that require saturated FP8 semantics should clamp before invoking
```
vkuzo: seems like this is only true for PyTorch 2.11 and older? if we are calling this out we should make sure it's correct across PT versions

vkuzo:
code lg, can we make sure if we are changing the comments the new ones reflect the state of the world?

torchao/float8/float8_utils.py
iamzainhuda: updated and added a TODO once we're no longer supporting PT 2.11

torchao/prototype/mx_formats/mx_tensor.py
iamzainhuda: yeah that makes sense, i left the key triton details in. i reworked the comment

torchao/quantization/quant_primitives.py
iamzainhuda: updated to specify pytorch version

iamzainhuda:
@vkuzo let me know if this is good to land!

vkuzo (approved):
lg!

PR #4374, opened 2026-05-09
[mxfp8 training] add MXFP8Linear

danielvegamyhre:
## Summary
- In torchtitan we are refactoring how quantized training is implemented:
    - Previous: initialize model definition with standard nn.Linear / torch._grouped_mm ops. Use torchao model converter to do module swaps or wrap nn.Parameters in tensor subclass with op overrides
    - New: apply quantization to model **config** first, then construct model based on the config, directly initializing MXFP8Linear / MXFP8Experts rather than doing a model conversion.
- This requires we implement the building blocks wrapping our "dynamic quant + compute" autograd funcs, to be directly initialized.

## Tests
- Added `test_mxfp8_linear.py`

## Misc changes
- Migrated `mx_formats/mx_linear.py` -> `moe_training/mxfp8_linear` since A) it was only used in moe_training dir, and B) `moe_training` will soon become just `training` and be the v2 for the torchao training design - so both linears and grouped gemms will live here. More refactors will be done to materialize this.

Authored using Claude

### DIFF
--- a/test/prototype/moe_training/test_mxfp8_linear.py
+++ b/test/prototype/moe_training/test_mxfp8_linear.py
@@ -1,0 +1,115 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+
+import pytest
+import torch
+import torch.nn.functional as F
+
+from torchao.quantization.utils import compute_error
+from torchao.utils import is_sm_at_least_100, torch_version_at_least
+
+if not (torch_version_at_least("2.7.0") and torch.cuda.is_available()):
+    pytest.skip("CUDA and PyTorch 2.7.0+ required", allow_module_level=True)
+
+from torchao.prototype.moe_training.mxfp8_linear import MXFP8Linear
+from torchao.quantization.quantize_.common.kernel_preference import KernelPreference
+
+
+def _build_linear_pair(
+    in_features: int,
+    out_features: int,
+    bias: bool,
+    kernel_preference: KernelPreference,
+    wgrad_with_hp: bool,
+):
+    ref = torch.nn.Linear(
+        in_features, out_features, bias=bias, device="cuda", dtype=torch.bfloat16
+    )
+    mxfp8 = MXFP8Linear(
+        in_features,
+        out_features,
+        bias=bias,
+        device="cuda",
+        dtype=torch.bfloat16,
+        kernel_preference=kernel_preference,
+        wgrad_with_hp=wgrad_with_hp,
+    )
+    # Share weights/bias so the only difference is the quantized matmul path.
+    with torch.no_grad():
+        mxfp8.weight.copy_(ref.weight)
+        if bias:
+            mxfp8.bias.copy_(ref.bias)
+    return ref, mxfp8
+
+
+@pytest.mark.skipif(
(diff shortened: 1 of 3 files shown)


test/prototype/moe_training/test_mxfp8_linear.py
```diff
@@ -0,0 +1,118 @@
...
+        if bias:
+            mxfp8.bias.copy_(ref.bias)
+    return ref, mxfp8
+
+
+def _requires_real_mxfp8(kernel_preference: KernelPreference):
+    if kernel_preference == KernelPreference.EMULATED:
+        return
+    if not is_sm_at_least_100():
+        pytest.skip("Real MXFP8 kernels require SM100+")
```
drisspg: NIT: I think if you do this as a decorator, you get better PyTests, like skip execution, non blocking though

test/prototype/moe_training/test_mxfp8_linear.py
```diff
@@ -0,0 +1,118 @@
...
+    "kernel_preference", [KernelPreference.EMULATED, KernelPreference.AUTO]
+)
+def test_mxfp8_linear_fwd_bwd_sqnr(bias, wgrad_with_hp, kernel_preference):
+    _requires_real_mxfp8(kernel_preference)
+
+    M, K, N = 1024, 1024, 2048
+    ref, mxfp8 = _build_linear_pair(K, N, bias, kernel_preference, wgrad_with_hp)
+
+    x_ref = torch.randn(M, K, dtype=torch.bfloat16, device="cuda", requires_grad=True)
+    x_mxfp8 = x_ref.clone().detach().requires_grad_(True)
```
drisspg, line 64: Nit; If you wanted to save one of these activations, you could just use torch.autograd.grad.

torchao/prototype/moe_training/mxfp8_linear.py
```diff
@@ -260,3 +261,32 @@ def backward(ctx, grad_output_hp: torch.Tensor):
...
+        output = _to_mxfp8_then_scaled_mm(
+            input,
+            self.weight,
+            kernel_preference=self.kernel_preference,
+            scale_calculation_mode=ScaleCalculationMode.RCEIL,
+            wgrad_with_hp=self.wgrad_with_hp,
+        )
+        if self.bias is not None:
+            output = output + self.bias.to(output.dtype)
+        return output
```
drisspg, line 300: Should MXFP8Linear accept an MXFP8TrainingOpConfig?

drisspg (approved):

torchao/prototype/moe_training/mxfp8_linear.py
danielvegamyhre: I am conflicted on this cause MXFP8TrainingOpConfig contains `pad_token_grouped_for_grouped_mm`, which makes sense for the generic model conversion approach that wraps params in a tensor subclass that overrides both `linear` and `grouped_mm` - however, if we are just directly initializing a MXFP8Linear it is weird/confusing to have this unused field. 

I would rather just have the few keyword args that the MXFP8Linear needs directly in the constructor, to keep it simple.

drisspg: are we plannig to support more / different backward options?

danielvegamyhre: None currently on my radar (besides wgrad_with_hp, which is already supported for both mxfp8 linear and grouped_mm)

PR #4377, opened 2026-05-11
[pat] (1/5) DTensor proxmap fix + low-rank speedups

lisjin:
### Summary
- `local_map` was discarding the in-place mutation written by the prox step because its implicit redistribute returned a fresh tensor.
- Wrap each prox application in an explicit `redistribute(...)` → in-place prox → `copy_(...)` round-trip, and pass `redistribute_inputs=False` so `local_map` stops re-shuffling the input under us.
- Modernize typing across pat/optim and pat/distributed_utils (typing.{List,Tuple,Dict,Set} -> builtins) and rewrite test_pat_proxmap.py to be DTensor-aware and exercise the redistribute path.
- Speed up the low-rank prox + reconstruction along the same DTensor / `vmap` path.

### Changes
- `torchao/prototype/pat/optim/pruneopt.py`: redistribute/copy-back wrapper around the prox call; flip `redistribute_inputs=False`.
- `torchao/prototype/pat/optim/{proxmap,group_lasso,lasso,nuclear_norm}.py`, `torchao/prototype/pat/{utils,distributed_utils}.py`: modernize typing
  (`typing.{List,Tuple,Dict,Set}` → builtins) — no behavior change.
- `group/low_rank.py`: reject DTensor inputs to `SVDGrouper`; add `deterministic` class flag; reconstruct as `(U * S) @ Vh` to skip `diag` / `diag_embed` materialization.
- `utils.py`: parameterize `use_deterministic_algorithms(mode=True)` and short-circuit when the requested mode already matches.
- `test/prototype/pat/test_pat_proxmap.py`: rewritten to be DTensor-aware and exercise the redistribute path.

### Test plan
- `python -m unittest discover -s test/prototype/pat -p 'test_pat_*.py'`

PR #4378, opened 2026-05-11
[pat] (2/5) Add QKGrouper and refactor AttentionHeadGrouper for vmap

lisjin:
### Summary
- `QKGrouper`: new grouper that prunes only the Q and K slabs of a packed QKV tensor, leaving V untouched.
- `AttentionHeadGrouperDim0`/`Dim1`: stop in-place-mutating `self.p.data` and reassign `self.p` instead, so `torch.vmap` (used by `PruneOptimizer`'s `local_map` path from PR1) can iterate over the grouped view without tripping the in-place check.

### Changes
- `torchao/prototype/pat/group/attention.py`
- `torchao/prototype/pat/group/__init__.py`: re-export `QKGrouper`.

Depends on #4377

### DIFF
--- a/torchao/prototype/pat/group/__init__.py
+++ b/torchao/prototype/pat/group/__init__.py
@@ -7,6 +7,7 @@
 from .attention import (  # noqa: F401
     AttentionHeadGrouperDim0,
     AttentionHeadGrouperDim1,
+    QKGrouper,
 )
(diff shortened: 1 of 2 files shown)


lisjin:
@claude Please review changes

andrewor14 (approved):

PR #4379, opened 2026-05-11
[pat] (3/5) Add KElementGrouper

lisjin:
### Summary
- New `KElementGrouper` that splits each row of a parameter tensor into fixed-size chunks of `k` elements.
- Non-divisible row sizes are zero-padded inside the context and the padding is stripped on exit, so the original tensor shape is preserved for downstream consumers.

### Changes
- `torchao/prototype/pat/group/k_element.py`
- `torchao/prototype/pat/group/__init__.py`: re-export `KElementGrouper`.
- `test/prototype/pat/test_pat_k_element_grouper.py`

Depends on #4378

### DIFF
--- a/test/prototype/pat/test_pat_k_element_grouper.py
+++ b/test/prototype/pat/test_pat_k_element_grouper.py
@@ -1,0 +1,137 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+
+import math
+import unittest
+
+import torch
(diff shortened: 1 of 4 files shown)


lisjin:
@claude Please review changes

lisjin:
@claude For issue 1, we don't need to check >2D tensors as we only use KElementGrouper on linear layers. For issue 2, PAT will never regularize bias tensors (1D). Finally, the check in issue 3 seems like overkill since we only use `KElementGrouper(k=1)` for element-wise pruning, or `KElementGrouper(k=4)` for 2:4 sparse pruning. In other words, `k` will generally be very small. Please re-evaluate your review.

andrewor14 (approved):

PR #4424, opened 2026-05-18
Convert `quant_lift_up` to `CustomGraphPass` with `uuid()`

frgossen:
Convert the `quant_lift_up` pre-grad custom pass in the x86 inductor passes from a plain function to a `QuantLiftUp(CustomGraphPass)` subclass that implements `uuid()`. This enables the AOTAutograd cache when `pre_grad_pass_timing` is "late".

A module-level `quant_lift_up` instance preserves backward compatibility with `x86_inductor_quantizer` registration.

### DIFF
--- a/torchao/quantization/pt2e/inductor_passes/x86.py
+++ b/torchao/quantization/pt2e/inductor_passes/x86.py
@@ -15,6 +15,7 @@ from typing import Any
 
 import torch
 from torch._dynamo.utils import counters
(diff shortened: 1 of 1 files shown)


vkuzo:
@claude review this PR

frgossen:
Addressed review feedback: narrowed `uuid()` return type from `Any | None` to `bytes` for consistency with `_CustomPass` in `qsdpa_fusion.py`.

PR #4425, opened 2026-05-19
[CPU] disable a few test cases due to an issue in torch nightly

Xia-Weiwen:
As the title
Guilty commit in torch: https://github.com/pytorch/pytorch/commit/868be1cb1f59e9aa4e52bced64f10940fad6828b


Valentine233 (approved):
The issue would break TorchAO CIs, so we skip the UTs now.
We should continue to figure out why the guilty commit affects the UTs.

PR #4427, opened 2026-05-19
Fix Python 3.14 compatibility of quantization/pt2e

rgommers:
Avoids the error:
```
AttributeError: 'typing.Union' object has no attribute '__module__'
```

This is a known issue, in Python 3.14 it's no longer supported to do this.

Closes gh-3619
Supersedes and closes gh-3657

The same kind of fix was already merged in gh-3832, this is just more of the same.

Issue #3243, opened 2025-10-24
TorchAO Missing 3.13T (free-threading) Wheels


Qubitium:
Latest `0.14.1` cuda builds does produce wheels for `3.13t` which is the `nogil` build of Python. 

On Ubuntu 24.04 x86_64

```py
# pip install torchao==0.14.1 --index-url https://download.pytorch.org/whl/cu130 -U
Looking in indexes: https://download.pytorch.org/whl/cu130
ERROR: Could not find a version that satisfies the requirement torchao==0.14.1 (from versions: none)
ERROR: No matching distribution found for torchao==0.14.1

# python --version
Python 3.13.8

# python --version
Python 3.13.8

# pip show torch
Name: torch
Version: 2.9.0+cu130
Summary: Tensors and Dynamic neural networks in Python with strong GPU acceleration
Home-page: https://pytorch.org
Author: 
Author-email: PyTorch Team <packages@pytorch.org>
License: BSD-3-Clause
Location: /root/vm313t/lib/python3.13t/site-packages
Requires: filelock, fsspec, jinja2, networkx, nvidia-cublas, nvidia-cuda-cupti, nvidia-cuda-nvrtc, nvidia-cuda-runtime, nvidia-cudnn-cu13, nvidia-cufft, nvidia-cufile, nvidia-curand, nvidia-cusolver, nvidia-cusparse, nvidia-cusparselt-cu13, nvidia-nccl-cu13, nvidia-nvjitlink, nvidia-nvshmem-cu13, nvidia-nvtx, setuptools, sympy, triton, typing-extensions
Required-by: accelerate, bitblas, causal_conv1d, flash_attn, GPTQModel, lm_eval, MemLord, peft, torchvision
```

Reported here
https://github.com/pytorch/ao/issues/2919#issuecomment-3443814140

And reproduced by another user here:
https://github.com/pytorch/ao/issues/2919#issuecomment-3444060877

janeyx99:
Ah, 3.13t and 3.14t both do not support CPython stable ABI, so there will have to be different wheels for those until maybe 3.15t if this PEP is approved/implemented https://peps.python.org/pep-0803/

Qubitium:
@janeyx99 It's been 6 months. Pretty please?

And it's now PEP approved as of 18 days ago.
<img width="1568" height="1106" alt="Image" src="https://github.com/user-attachments/assets/7d9e481e-c104-47c4-a2a7-0314defe193c" />

[...]

PR #4428, opened 2026-05-19
Enable building under a free-threaded Python interpreter

rgommers:
Free-threaded interpreters do not support the CPython Limited C API, and hence use of the limited API has to be disabled. With this one change, `torchao` builds fine.

The `is_freethreaded` function is in widespread use, and documented as the canonical way of doing this check here:
https://py-free-threading.github.io/running-gil-disabled/#check-if-using-a-free-threaded-build

The changes to `setup.py` mirror how this was done in other projects, e.g., [vllm#29241](https://github.com/vllm-project/vllm/pull/29241).

This implements the first step of https://github.com/pytorch/ao/issues/3243#issuecomment-4428954678. I have tested this locally, and (after fixing a couple of unrelated issues, namely gh-4426, gh-4427) with this change `torchao` can be installed under a free-threading Python 3.14t interpreter, and almost all tests pass (the ones that don't look 3.14 rather than free-threading related). 

***

The second commit also fixes a problem under a default interpreter: the `_C_mxfp8` extension module was not being compiled against the CPython Limited API, and hence wheels were not `abi3`-compatible.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -793,7 +793,9 @@ def get_extensions():
 
     maybe_cpython_limited_api = []
     if not is_freethreaded():
-        maybe_cpython_limited_api += [f"-DPy_LIMITED_API={min_supported_cpython_hexcode}"]
+        maybe_cpython_limited_api += [
+            f"-DPy_LIMITED_API={min_supported_cpython_hexcode}"
+        ]
 
     extra_link_args = []
     extra_compile_args = {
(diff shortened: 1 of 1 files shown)


vkuzo:
@claude review this PR, make sure to consider all the different hardwares that torchao targets

rgommers:
> 1. The `_C_mxfp8` extension defines `Py_LIMITED_API` via compile flags but never sets `py_limited_api=True` as a kwarg to `CUDAExtension()`. This is a pre-existing inconsistency (not introduced by this PR) and doesn't affect correctness for the free-threading change.

This finding looks correct and important to fix. @vkuzo I'm happy to do that in either this PR or a follow-up PR, whatever seems best to you. It looks like a one-liner fix, inserting `py_limited_api=not is_freethreaded()` for that extension.

vkuzo:
> This finding looks correct and important to fix. 

thank you for helping! want to amend this PR so it's all in one go?

Also, CI uncovered a ruff failure, you can run `ruff format && ruff check . --fix` from root to resolve

rgommers:
> thank you for helping! want to amend this PR so it's all in one go?

Thanks, done! Also edited the PR description to point out the bug fix.

PR #4431, opened 2026-05-20
fix(pt2e/constant_fold): preserve shared get_attr target when only some uses are dead (#4420)

### LINKED ISSUE #4420: constant_fold deletes shared get_attr target while another get_attr for the same target is still live
### Description

### Steps to reproduce

```python
import torch


class Root(torch.nn.Module):
   def __init__(self):
      super().__init__()

root = Root().eval()

g = Graph()
x = g.placeholder("x")

g.output((folded, live))
gm = GraphModule(root, g)

print("Before:")
print(gm.graph)

constant_fold(gm)

>>

      self.lint()
      raise RuntimeError(
```

### Environment

```text
torch: 2.11.0+cpu
torchao: 0.17.0+git02105d46c
```

Anai-Guo:
## Summary

Fixes #4420.

`torchao.quantization.pt2e.constant_fold.constant_fold` previously called `delattr(gm, node.target)` for every dead `get_attr` node without checking whether **other live** `get_attr` nodes still pointed at the same target. When an FX graph contains multiple `get_attr` nodes referencing the same buffer/parameter (e.g. ExecuTorch graphs where a shared buffer is read through several `get_attr` nodes), the attribute would be deleted as soon as any one of them became dead, leaving the remaining live `get_attr` nodes dangling and causing the immediately-following `graph.lint()` to raise:

```
RuntimeError: Node shared_1 target shared references nonexistent attribute shared of <graph module>
```

This PR splits the loop in two: first collect the dead `get_attr` nodes, then build the set of targets that still have at least one live `get_attr`, and only call `delattr` for targets that fall entirely out of that set.

## Reproducer (from #4420)

```python
import torch

class Root(torch.nn.Module):
    def __init__(self):
        super().__init__()

root = Root().eval()
g = Graph()
x = g.placeholder("x")
g.output((folded, live))
gm = GraphModule(root, g)

constant_fold(gm)  # before: RuntimeError on graph.lint(); after: ok
```

## Test plan

- [x] Reproducer from #4420 no longer raises.
- [ ] Existing `test/quantization/pt2e/` tests should continue to pass (no behavioural change when at most one `get_attr` references a given target).

---

AI-assisted, human reviewed.

### DIFF
--- a/torchao/quantization/pt2e/constant_fold.py
+++ b/torchao/quantization/pt2e/constant_fold.py
@@ -353,11 +353,24 @@ def constant_fold(
         erased_params = []
         for node in gm.graph.find_nodes(op="get_attr"):
             if len(node.users) == 0:
-                if hasattr(gm, node.target):
-                    delattr(gm, node.target)
                 erased_params.append(node)
 
+        # Only delete the underlying module attribute when no other live
+        # get_attr node still references the same target. Without this guard,
+        # graphs that contain multiple get_attr nodes pointing at the same
+        # buffer/parameter would lose the attribute as soon as any one of
+        # them became dead, leaving the remaining live get_attr dangling and
+        # making the subsequent graph.lint() fail.
+        live_get_attr_targets = {
+            n.target
+            for n in gm.graph.find_nodes(op="get_attr")
+            if len(n.users) > 0
+        }
         for node in erased_params:
+            if node.target not in live_get_attr_targets and hasattr(
+                gm, node.target
+            ):
+                delattr(gm, node.target)
             gm.graph.erase_node(node)
 
         gm.graph.eliminate_dead_code()
--- a/torchao/quantization/pt2e/constant_fold.py
+++ b/torchao/quantization/pt2e/constant_fold.py
@@ -362,14 +362,10 @@ def constant_fold(
         # them became dead, leaving the remaining live get_attr dangling and
         # making the subsequent graph.lint() fail.
         live_get_attr_targets = {
-            n.target
-            for n in gm.graph.find_nodes(op="get_attr")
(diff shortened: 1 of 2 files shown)


jerryzh168:
makes sense, can you add a test?

Anai-Guo:
Added a regression test in `test/quantization/pt2e/test_quantize_pt2e.py` (`test_constant_fold_preserves_shared_get_attr_target`) — it builds a tiny FX graph with two `get_attr` nodes pointing at the same buffer (one live, one dead), runs `constant_fold`, and verifies the buffer survives and the recompiled module still executes. On the pre-fix code this fails at `graph.lint()` after the shared attribute is deleted.

Let me know if you'd prefer it co-located somewhere else or matched against an existing constant-fold test class.

torchao/quantization/pt2e/constant_fold.py
```diff
@@ -353,11 +353,24 @@ def constant_fold(
         erased_params = []
         for node in gm.graph.find_nodes(op="get_attr"):
             if len(node.users) == 0:
```
jerryzh168, line 355: wait, didn't we already check for number of users here?

andrewor14:
@claude review

Anai-Guo:
Thanks for the thorough review! I've implemented the optional test-strengthening note in c63c6dd: the regression test now also asserts that exactly one `get_attr` node referencing `"shared"` remains after folding, so it pins down both halves of the behavior — the buffer attribute is preserved **and** the dead `get_attr` node is actually erased.

I left the `live_get_attr_targets` recompute as-is (the second `find_nodes` pass) since, as you noted, the current two-loop form reads more clearly and the extra scan is negligible here.

PR #4432, opened 2026-05-21
Add algorithm search to FP8 sparse linear benchmark (#4432)

gyllstromk:
Summary:
What: Adds algorithm search support (--search-alg) to the FP8 sparse linear benchmark. Threads a new alg_id parameter through the quantization config (Float8DynamicActivationFloat8WeightConfig), the sparse tensor class, and down into the _cslt_sparse_mm kernel call. When --search-alg is passed, the benchmark calls _cslt_sparse_mm_search to find the best algorithm for each shape and benchmarks with that algorithm to report the speedup.

Why: hipSPARSELt supports multiple algorithms for sparse matmul, and the default (alg_id=0) isn't always the fastest. This lets us find the best algorithm for a given shape and measure the performance benefit.

Differential Revision: D102683062

Pulled By: gyllstromk

### DIFF
--- a/torchao/quantization/quant_api.py
+++ b/torchao/quantization/quant_api.py
@@ -1169,6 +1169,7 @@ class Float8DynamicActivationFloat8WeightConfig(AOBaseConfig):
     kernel_preference: KernelPreference = KernelPreference.AUTO
     set_inductor_config: bool = True
     version: int = 2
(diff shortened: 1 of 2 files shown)


vkuzo:
@claude review this PR

vkuzo (approved):

PR #4402, opened 2026-05-14
switch CI to PyTorch 2.12

vkuzo:
Summary:

bumping stable CI jobs from PyTorch 2.11 to 2.12

Test Plan: CI

PR #4437, opened 2026-05-22
fix rocm CI

vkuzo:
Summary:

In https://github.com/pytorch/ao/pull/4402 we switched CI to
PyTorch 2.12.

This broke ROCM because the ROCM jobs pinned rocm7.0, and the PyTorch
nightly for that is not available for v2.12+

fix is to pin a more recent rocm version

Test Plan: CI


vkuzo:
* __->__ #4437

drisspg (approved):

PR #4456, opened 2026-05-29
[CI] Switch H100 jobs to linux_job_v3 (OSDC/ARC)

huydhn:
## What

Switches torchao's H100 CI jobs from the EC2-based `linux_job_v2` to the OSDC (ARC) `linux_job_v3` reusable workflow, to evaluate running H100 tests on on-site data-center (EKS/ARC) runners.

- **`1xH100_tests.yml`**: `linux.aws.h100` → `mt-l-x86iamx-22-225-h100`
- **`4xH100_tests.yml`**: `linux.aws.h100.4` → `mt-l-x86iamx-88-900-h100-4`
- Both now call `pytorch/test-infra/.github/workflows/linux_job_v3.yml@linux-job-v3-osdc`.

## How v3 differs from v2

`linux_job_v3` runs the job **inside a job-level `container:`** on an ARC runner (instead of running on an EC2 host and launching a nested `docker run`). The script runs directly in the container; GPUs are exposed via `--gpus all` + NVIDIA driver capabilities. The almalinux-builder CUDA images (resolved from `gpu-arch-version`) are pre-built and pulled as the job container.

## Note on the pip guard

The ARC image injects a CPU-only pip mirror (`PIP_EXTRA_INDEX_URL=.../whl/cpu`), which otherwise resolves `torch` to a `+cpu` wheel even with `--index-url .../cuXXX`. The torch install line is prefixed with `PIP_CONFIG_FILE=/dev/null PIP_EXTRA_INDEX_URL=` so the CUDA index is the only source.

## Status

**Draft** — depends on `linux_job_v3.yml` landing in `pytorch/test-infra` (currently on branch `linux-job-v3-osdc`, validated by `test_linux_job_v3.yml`). The `@linux-job-v3-osdc` ref should be bumped to `@main` once that PR merges.

PR #4457, opened 2026-05-30
[CI] Switch release_model H100 job to linux_job_v3 (OSDC/ARC)

huydhn:
## What

Migrates the `release_model.yml` H100 job from EC2 `linux_job_v2` to the OSDC (ARC) `linux_job_v3` reusable workflow.

- Runner: `linux.aws.h100` → `mt-l-x86iamx-22-225-h100` (from `.github/arc.yaml`).
- `uses: …/linux_job_v3.yml@main`.
- Pre-install torch's pure-python deps from the in-cluster pypi-cache, then install torch from the literal nightly cu126 index. On ARC, `cache-enforcer` iptables-blocks `files.pythonhosted.org`, so the deps (which the nightly index references there) must come from the reachable cache; `mslk` is a `+cuXXX` wheel on the pytorch index and stays in `torch-spec`.

Part of the torchao H100/A100 → OSDC migration. Companion PRs: #4456 (1x/4xH100). Depends on `linux_job_v3` (now on `pytorch/test-infra@main`).

### DIFF
--- a/.github/workflows/release_model.yml
+++ b/.github/workflows/release_model.yml
@@ -18,14 +18,14 @@ jobs:
       matrix:
         include:
           - name: H100
-            runs-on: linux.aws.h100
+            runs-on: mt-l-x86iamx-22-225-h100
             torch-spec: '--pre torch torchvision torchaudio mslk --index-url https://download.pytorch.org/whl/nightly/cu126'
             gpu-arch-type: "cuda"
             gpu-arch-version: "12.4"
     permissions:
       id-token: write
       contents: read
-    uses: pytorch/test-infra/.github/workflows/linux_job_v2.yml@main
+    uses: pytorch/test-infra/.github/workflows/linux_job_v3.yml@main
     with:
       timeout: 90
       runner: ${{ matrix.runs-on }}
@@ -38,7 +38,11 @@ jobs:
         export PATH=/opt/rh/devtoolset-10/root/usr/bin/:$PATH
         python -m pip install --upgrade pip
         pip install uv
-        pip install ${{ matrix.torch-spec }}
+        # Pre-install torch/vision's pure-python deps from the in-cluster pypi-cache for speed.
+        pip install filelock typing-extensions "setuptools<82" sympy networkx jinja2 fsspec numpy pillow
+        # Clear PIP_EXTRA_INDEX_URL (the runner's default cpu /whl/cpu/) so it can't supply a
+        # +cpu torch; the torch-spec's --index-url makes the literal nightly cu126 index the only source.
+        PIP_EXTRA_INDEX_URL= pip install ${{ matrix.torch-spec }}
         uv pip install -r dev-requirements.txt
         pip install . --no-build-isolation
         HF_MODEL_ID=${{ github.event.inputs.hf_model_id }}


.github/workflows/release_model.yml
```diff
@@ -38,7 +38,11 @@ jobs:
         export PATH=/opt/rh/devtoolset-10/root/usr/bin/:$PATH
         python -m pip install --upgrade pip
         pip install uv
-        pip install ${{ matrix.torch-spec }}
+        # Pre-install torch/vision's pure-python deps from the in-cluster pypi-cache for speed.
+        pip install filelock typing-extensions "setuptools<82" sympy networkx jinja2 fsspec numpy pillow
+        # Clear PIP_EXTRA_INDEX_URL (the runner's default cpu /whl/cpu/) so it can't supply a
+        # +cpu torch; the torch-spec's --index-url makes the literal nightly cu126 index the only source.
+        PIP_EXTRA_INDEX_URL= pip install ${{ matrix.torch-spec }}
```
malfet, line 45: Err, is it documented anywhere?

malfet, line 45: Why one can not leave this section as is?

huydhn: First I want to set some context here:

* OSDC runners are not allowed to directly access pypi or other registries until we can increase the number of public facing IP address. So, access to files.pythonhosted.org is blocked by firewall.  Only access to download.pytorch.org is allowed after https://github.com/pytorch/ci-infra/pull/545 as LF is the owner there.
* The runner automatically set some local env variable like `PIP_EXTRA_INDEX_URL` to point to the local cache at http://pypi-cache-cpu.pypi-cache.svc.cluster.local:8080.

These are 2 new pypi-cache compatibility issues that I found today with the setup, one issue per line.  I have created 2 tracking issue for them here to follow on this next week:

1. The first issue is that we point to files.pythonhosted.org in download.pytorch.org for some common python packages. For example, https://download.pytorch.org/whl/nightly/filelock.  I observed that a command like `pip install torch --index-url https://download.pytorch.org/whl/nightly/cu130`, by fixing the index to download.pytorch.org, could get all packages from download.pytorch.org but not those from files.pythonhosted.org. Running `pip install filelock typing-extensions "setuptools<82" sympy networkx jinja2 fsspec numpy pillow` is a quick way to fetch these packages separately without setting the download.pytorch.org index.  It works because it correctly use the local cache without referring to download.pytorch.org. Tracking issue https://github.com/pytorch/ci-infra/issues/660

3. The second issue is that `PIP_EXTRA_INDEX_URL` is set wrongly to `PIP_EXTRA_INDEX_URL=http://pypi-cache-cpu.pypi-cache.svc.cluster.local:8080/whl/cpu/`.  With this variable set, torch CPU is somehow preferred over the correct CUDA version. Unset it is a quick way to tell pip to look into https://download.pytorch.org/whl/nightly/cu130 instead. Tracking issue https://github.com/pytorch/ci-infra/issues/661

I don't have the full context on these behavior yet and need more time to look into them. (1) is gotcha while (2) is a bug.  So, these changes are here to unblock torchao CI if needed.  I agree that we shouldn't need them once (1) and (2) are fixed.

vkuzo:
@claude review

vkuzo (approved):

vkuzo:
@huydhn should this job have a succeeding 1xh100 run? looks like the one in the current PR was cancelled

huydhn:
> @huydhn should this job have a succeeding 1xh100 run? looks like the one in the current PR was cancelled

I need to rebase to get the changes from the other 2 PRs, it looks like jobs are running now, so let's merge once they pass

PR #4458, opened 2026-05-30
[CI] Move H100/A100 benchmark & tutorial jobs to OSDC (ARC)

huydhn:
## What

Moves the direct `runs-on:` H100/A100 jobs onto OSDC (ARC). These don't use `linux_job`, so per the agreed approach they get a job-level `container:` added in place rather than being converted to the reusable workflow.

| Workflow | Runner: EC2 → ARC | Upload |
|---|---|---|
| `run_microbenchmarks.yml` | `linux.aws.h100` → `mt-l-x86iamx-22-225-h100` | yes |
| `dashboard_perf_test.yml` | `linux.aws.a100` → `mt-l-x86iavx512-11-125-a100` | yes |
| `run_tutorials.yml` | `linux.aws.a100` → `mt-l-x86iavx512-11-125-a100` | no |

Each job now runs inside `pytorch/almalinux-builder:cuda13.0` with `options: --gpus all` (the container CUDA version is independent of the pip-installed cu126 torch; cuda13.0 matches the ARC nodes' driver). torch's pure-python deps are pre-installed from the in-cluster pypi-cache, and `PIP_EXTRA_INDEX_URL` is cleared on the torch install so the default cpu index can't supply a `+cpu` torch (same pattern as the H100 PRs).

The two upload jobs get `id-token: write` + a `configure-aws-credentials` step using `role/gha_workflow_upload-benchmark-results` (ARC pods have no EC2 instance profile).

Scope: **H100/A100 only** — regular GPU (g5/g6/L4) and CPU jobs are intentionally left on EC2. Companion PRs: #4456 (1x/4xH100), #4457 (release_model). Depends on `linux_job_v3`-era infra (now on `pytorch/test-infra@main`).

## Needs validation
- `setup-miniconda` + benchmark uploads inside the ARC container (esp. that `role/gha_workflow_upload-benchmark-results` is assumable from these pods).
- `actions/checkout`/JS actions inside `almalinux-builder` (git present; node provided by the ARC container hook).

### DIFF
--- a/.github/workflows/dashboard_perf_test.yml
+++ b/.github/workflows/dashboard_perf_test.yml
@@ -11,7 +11,14 @@ on:
 jobs:
   benchmark:
     timeout-minutes: 500
-    runs-on: linux.aws.a100
+    runs-on: mt-l-x86iavx512-11-125-a100
+    permissions:
+      id-token: write
+      contents: read
+    # OSDC (ARC): run inside a CUDA container; --gpus all exposes the A100.
+    container:
+      image: pytorch/almalinux-builder:cuda13.0
+      options: --gpus all
     strategy:
       matrix:
         torch-spec:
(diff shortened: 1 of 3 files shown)


vkuzo:
@claude review

vkuzo (approved):

PR #4312, opened 2026-04-21
docs(finetuning): replace TorchTune QAT section with Unsloth

Anai-Guo:
## Summary

Closes #4287

Replaces the "Option 1: TorchTune QAT Integration" section in the fine-tuning tutorial with an Unsloth integration section, as requested in the linked issue.

Changes:
- Update the option list to reference Unsloth instead of TorchTune
- Replace the entire TorchTune QAT section (install commands, tune CLI invocations, eval tables, TorchTune-specific image) with Unsloth content including:
  - Installation via `pip install --upgrade unsloth unsloth_zoo`
  - Python code for `FastLanguageModel.from_pretrained` + `get_peft_model` with `qat_scheme="int4"`
  - Supported QAT schemes: `int4`, `int8-int4`, `fp8-int4`, `fp8-fp8`
  - Experiment results table (Wikitext PPL) for Llama-3.2-3B, Gemma3-4B-it, Gemma3-12B-it from the [QAT experiments page](https://docs.pytorch.org/ao/main/workflows/qat.html)
  - Links to the Unsloth QAT tutorial and the ao QAT experiments page
- Update Option 3 description to reference Unsloth instead of TorchTune

## Test plan

- [ ] Verify RST renders correctly in Sphinx (`docs/source/eager_tutorials/finetuning.rst`)
- [ ] Confirm code snippets match the current Unsloth API
- [ ] Confirm experiment numbers match https://docs.pytorch.org/ao/main/workflows/qat.html

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4459, opened 2026-06-02
fix(Int8DynActInt4WeightQuantizer): use scales_precision for runtime modules

### LINKED ISSUE #2571: Int8DynActInt4WeightQuantizer quantize() not propagating the scales_precision argument
Hello,

In  [`Int8DynActInt4WeightQuantizer`](https://github.com/pytorch/ao/blob/main/torchao/quantization/linear_quant_modules.py#L617), the `scales_precision` argument of `replace_linear_8da4w` is set to `self.precision` instead of `self.scales_precision`.

This introduces an inconsistency between calling `_create_quantized_state_dict(model)` and calling `quantize(model).state_dict()`.

Anai-Guo:
## Summary

`Int8DynActInt4WeightQuantizer._convert_for_runtime` passed `self.precision` as the `scales_precision` argument of `replace_linear_8da4w`, so the runtime `Int8DynActInt4WeightLinear` modules were built with the wrong scales dtype. The code even carried a `# TODO: this should be self.scales_precision?` next to the line.

This diverges from `_create_quantized_state_dict`, which uses `self.scales_precision` for the scales. As reported in #2571, that mismatch makes `quantize(model).state_dict()` inconsistent with `_create_quantized_state_dict(model)` whenever `precision != scales_precision`.

## Fix

Pass `self.scales_precision` (instead of `self.precision`) as the `scales_precision` argument, matching `_create_quantized_state_dict`.

```python
 replace_linear_8da4w(
     model,
     self.groupsize,
     self.padding_allowed,
     self.precision,
-    # TODO: this should be self.scales_precision?
-    self.precision,
+    self.scales_precision,
 )
```

Fixes #2571

## Test plan
- [ ] Instantiate `Int8DynActInt4WeightQuantizer(precision=torch.float32, scales_precision=torch.bfloat16)`, quantize a small linear model, and confirm the runtime module scales dtype now matches `_create_quantized_state_dict` output.

### DIFF
--- a/test/quantization/test_quant_api.py
+++ b/test/quantization/test_quant_api.py
@@ -269,14 +269,20 @@ class TestQuantFlow(TestCase):
             precision=torch.float32,
             scales_precision=torch.bfloat16,
         )
-        m = ToyLinearModel().eval()
+        m = ToyLinearModel(bias=True).eval()
         example_inputs = m.example_inputs()
         m = quantizer.quantize(m)
         for name in ("linear1", "linear2"):
             mod = getattr(m, name)
             assert isinstance(mod, Int8DynActInt4WeightLinear)
+            # scales/zeros buffers follow scales_precision...
             self.assertEqual(mod.scales.dtype, torch.bfloat16)
             self.assertEqual(mod.zeros.dtype, torch.bfloat16)
+            # ...while the activation precision and bias buffer still follow
+            # precision. Asserting these guards against a future change that
+            # over-corrects by routing everything through scales_precision.
+            self.assertEqual(mod.precision, torch.float32)
+            self.assertEqual(mod.bias.dtype, torch.float32)
         m(*example_inputs)
 
     @unittest.skipIf(not torch.accelerator.is_available(), "Need GPU available")
--- a/test/quantization/test_quant_api.py
+++ b/test/quantization/test_quant_api.py
@@ -255,6 +255,30 @@ class TestQuantFlow(TestCase):
         assert isinstance(m.linear2, Int8DynActInt4WeightLinear)
         m(*example_inputs)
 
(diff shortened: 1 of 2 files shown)


andrewor14:
@claude review

Anai-Guo:
Thanks @claude for the review. I took the non-blocking suggestion and tightened the regression test to document the full dtype contract: the test now uses `ToyLinearModel(bias=True)` and additionally asserts `mod.precision == torch.float32` and `mod.bias.dtype == torch.float32` alongside the existing bfloat16 scales/zeros checks. This makes it fail not only on the original bug but also on a future change that over-corrects by routing `precision` through `scales_precision`.

andrewor14:
Thanks, will merge this for now but I'm actually surprised this code still exists. We meant to deprecate/delete it, will do that separately.

andrewor14:
Will wait for Ci to pass first

Anai-Guo:
@andrewor14 The remaining red checks here look like pre-existing nightly/infra breakage rather than anything from this change:

- All the **release** test jobs pass: `test (CPU 2.12)`, `test (CUDA 2.12)`, `test (SM-89)`, plus the CPU/MPS ops jobs.
- Only the `test-nightly (CPU Nightly / CPU Nightly (Clang) / CUDA Nightly)` jobs and the ~2-min `test (H100, ...)` setup fail.
- The exact same `test-nightly` jobs are failing identically on unrelated **docs-only** PRs (e.g. #4312, #4475), which can't possibly touch CPU/CUDA test behavior — so the failures are environmental (nightly `--pre torch` / docker-hash), not introduced here.

So the dtype regression test added in this PR is green on the release runners. Happy to rebase if a fresh nightly helps, but it should be safe to merge once you're comfortable.

andrewor14 (approved):

PR #4460, opened 2026-06-02
Add blockwise FP8 MoE functionality with emulated GEMM

iamzainhuda:
## Summary
- add blockwise FP8 MoE training config and grouped-MM autograd path
- move grouped blockwise custom ops and scale-layout prep into torchao.prototype.blockwise_fp8_training.grouped_kernels
- add tests for grouped weight quantization, fwd/bwd, compile, and quantize dispatch

this uses an emulated gemm path as pytorch core today does not have a fp8 blockwise scaled grouped gemm. the emulated path serves as a correctness bridge so we can verify numerics.

**Follow up PRs will make this performant, this initial PR is to setup the wiring and get the quantization numerics in place before moving further**

Part of #3290

## Tests
```
  pytest -q test/prototype/moe_training/test_fp8_blockwise_grouped_mm.py
```
  For the broader related coverage:
```
  pytest -q test/prototype/moe_training/test_tensor.py
  pytest -q test/prototype/moe_training/test_fp8_grouped_mm.py
  pytest -q test/prototype/blockwise_fp8_training/test_blockwise_kernels.py
```

### DIFF
--- a/test/prototype/moe_training/test_fp8_blockwise_grouped_mm.py
+++ b/test/prototype/moe_training/test_fp8_blockwise_grouped_mm.py
@@ -20,7 +20,7 @@ if not (
 pytest.importorskip("triton", reason="Triton required to run this test")
 
 from torchao.prototype.moe_training.blockwise_fp8.grouped_mm import (
-    _to_fp8_blockwise_then_scaled_grouped_mm,
+    _to_fp8_blockwise_then_emulated_scaled_grouped_mm,
 )
 from torchao.quantization.utils import compute_error
 from torchao.testing.utils import skip_if_rocm
@@ -41,7 +41,9 @@ def _make_column_major_weight_t(E: int, N: int, K: int) -> torch.Tensor:
         (torch.tensor([129, 384, 500], dtype=torch.int32), True),
     ],
 )
-def test_fp8_blockwise_grouped_mm_fwd_bwd(offs, pad_token_groups_for_grouped_mm):
+def test_fp8_blockwise_emulated_grouped_mm_fwd_bwd(
+    offs, pad_token_groups_for_grouped_mm
+):
     torch.manual_seed(0)
     offs = offs.cuda()
     E = offs.numel()
@@ -53,7 +55,7 @@ def test_fp8_blockwise_grouped_mm_fwd_bwd(offs, pad_token_groups_for_grouped_mm)
     A_ref = A.detach().clone().requires_grad_(True)
     B_t_ref = B_t.detach().clone().requires_grad_(True)
 
-    out = _to_fp8_blockwise_then_scaled_grouped_mm(
+    out = _to_fp8_blockwise_then_emulated_scaled_grouped_mm(
         A,
         B_t,
         offs,
@@ -73,13 +75,15 @@ def test_fp8_blockwise_grouped_mm_fwd_bwd(offs, pad_token_groups_for_grouped_mm)
 
 
 @skip_if_rocm("ROCm not supported")
-def test_fp8_blockwise_grouped_mm_compile_aligned_groups():
+def test_fp8_blockwise_emulated_grouped_mm_compile_aligned_groups():
     E, M, K, N = 2, 256, 128, 128
     A = torch.randn(E * M, K, dtype=torch.bfloat16, device="cuda")
     B_t = _make_column_major_weight_t(E, N, K)
     offs = torch.arange(M, (E + 1) * M, M, device="cuda", dtype=torch.int32)
 
-    compiled = torch.compile(_to_fp8_blockwise_then_scaled_grouped_mm, fullgraph=True)
+    compiled = torch.compile(
+        _to_fp8_blockwise_then_emulated_scaled_grouped_mm, fullgraph=True
+    )
     out = compiled(A, B_t, offs, pad_token_groups_for_grouped_mm=False)
 
     assert out.shape == (E * M, N)
--- a/test/prototype/moe_training/test_fp8_blockwise_grouped_mm.py
+++ b/test/prototype/moe_training/test_fp8_blockwise_grouped_mm.py
@@ -6,7 +6,6 @@
 
 import pytest
 import torch
-from torch import nn
 
 from torchao.utils import is_MI300, is_MI350, is_sm_at_least_90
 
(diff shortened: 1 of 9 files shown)


vkuzo:
@claude review

torchao/prototype/moe_training/utils.py
```diff
@@ -375,12 +375,24 @@ def _quantize_then_scaled_grouped_mm(
         config (TrainingOpBaseConfig): Configuration for quantization recipe, etc
     """
     from torchao.prototype.moe_training import (
+        _to_fp8_blockwise_then_scaled_grouped_mm,
```
vkuzo: I think the right time to land blockwise change to this function would be after there is a fast grouped gemm + cast kernels, it seems too early now

torchao/prototype/moe_training/config.py
```diff
@@ -24,6 +24,7 @@ class Float8TrainingRecipe(Enum):
     """FP8 recipes for grouped matrix multiplication."""
 
     FP8_ROWWISE = "fp8_rowwise"
+    FP8_BLOCKWISE = "fp8_blockwise"
```
vkuzo: seems too early for this, we should have the kernels first

torchao/prototype/moe_training/blockwise_fp8/grouped_mm.py
```diff
@@ -0,0 +1,236 @@
...
+)
+from torchao.prototype.moe_training.utils import (
+    conditional_nostrict_trace,
+    pad_token_groups,
+    unpad_token_groups,
+)
+
+
+@conditional_nostrict_trace
+def _to_fp8_blockwise_then_scaled_grouped_mm(
```
vkuzo: this is reasonable, I'd recommend giving it a subdirectory instead of putting directly into `moe_training` until there are kernels

torchao/prototype/blockwise_fp8_training/grouped_kernels.py
```diff
@@ -0,0 +1,321 @@
...
+    offs: torch.Tensor,
+    out_dtype: torch.dtype,
+    block_size: int = 128,
+    use_native_grouped_gemm: bool = False,
+) -> torch.Tensor:
+    assert _is_row_major(a), "blockwise_scaled_grouped_mm expected row-major A"
+    assert _is_column_major(b), "blockwise_scaled_grouped_mm expected column-major B"
+    assert offs is not None and offs.dtype == torch.int32, "offs must be int32"
+    b_s = _prepare_grouped_rhs_scale(b_s, scale_recipe_b)
+    # this flag is here for when we can use pytorch scaled grouped mm with blockwise
```
vkuzo: nit: don't add it if it doesn't work, add it later when that path actually exists.  You can do

```
# TODO(future): hook up F.scaled_grouped_mm once it supports float blockwise
```

test/prototype/moe_training/test_fp8_blockwise_grouped_mm.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+def _make_column_major_weight_t(E: int, N: int, K: int) -> torch.Tensor:
+    weight = torch.randn(E, N, K, dtype=torch.bfloat16, device="cuda")
+    return weight.contiguous().transpose(-2, -1)
+
+
+@skip_if_rocm("ROCm not supported")
+@pytest.mark.parametrize("E,K,N", [(1, 128, 128), (3, 256, 384)])
+def test_grouped_weight_quant_transposed_rhs_matches_dense(E, K, N):
```
vkuzo: isn't this always true now because of what the function is currently doing? if yes, then i'd delete this test and only add it when the kernel is implementing a grouped gemm without a cpu for loop

torchao/prototype/moe_training/utils.py
iamzainhuda: yup makes sense. users can be confused otherwise that this functionality is "ready"

test/prototype/moe_training/test_fp8_blockwise_grouped_mm.py
iamzainhuda: yeah that's true, this is just validating the current per expert loop behaviour. i've removed it. i also removed some other tests (quantize dispatch test) since we removed the blockwise from the configs earlier as well

torchao/prototype/moe_training/blockwise_fp8/grouped_mm.py
iamzainhuda: kk, i put these under blockwise_fp8 subdirectory under moe_training

torchao/prototype/blockwise_fp8_training/grouped_kernels.py
```diff
@@ -0,0 +1,296 @@
...
+    scale_out = torch.empty_strided(
+        (E, N // block_size, K // block_size),
+        ((N // block_size) * (K // block_size), 1, N // block_size),
+        dtype=torch.float32,
+        device=weight_t.device,
+    )
+    return q_out, scale_out
+
+
+# NOTE: only for emulated backend, we can remove once we have native scaled grouped gemm
```
vkuzo: nit: I don't think this will be removed, people usually expand the scale for the real gemm kernel as well

torchao/prototype/blockwise_fp8_training/grouped_kernels.py
```diff
@@ -0,0 +1,296 @@
...
+) -> torch.Tensor:
+    a_scale = _expand_blockwise_scale(a, a_s, scale_recipe_a, block_size)
+    b_scale = _expand_blockwise_scale(b, b_s, scale_recipe_b, block_size)
+    a_hp = a.to(torch.bfloat16) * a_scale.to(torch.bfloat16)
+    b_hp = b.to(torch.bfloat16) * b_scale.to(torch.bfloat16)
+    return torch._grouped_mm(a_hp, b_hp, offs=offs, out_dtype=out_dtype)
+
+
+@torch.library.custom_op("torchao::blockwise_scaled_grouped_mm", mutates_args=())
+def blockwise_scaled_grouped_mm(
```
vkuzo: nit: call it emulated since the only logic today is emulated

torchao/prototype/moe_training/blockwise_fp8/grouped_mm.py
```diff
@@ -0,0 +1,227 @@
...
```
vkuzo: call it emulated until there is a real path, this is early prototype so you can change names at will

torchao/prototype/moe_training/config.py
```diff
@@ -227,8 +227,8 @@ def __hash__(self):
         )
 
 
-@register_quantize_module_handler(Float8TrainingOpConfig)
 @register_quantize_module_handler(MXFP8TrainingOpConfig)
+@register_quantize_module_handler(Float8TrainingOpConfig)
```
vkuzo, line 230: revert?

vkuzo (approved):
lg! I'd just make sure the names in the code as well as the PR title make it clear the gemm is emulated for now.

PR #4461, opened 2026-06-03
Remove MSLK FP8 tensorwise support

cthi:
As titled, we remove this support from AO, as we plan to deprecate the FP8 tensorwise support in MSLK. Using `KernelPreference.TORCH` is better as it would suit `_scaled_mm` which has better support/performance through cublas and inductor integration.


vkuzo:
@claude review

vkuzo (approved):

Issue #4408, opened 2026-05-15
[RFC] MoE Quantization-Aware Training (QAT) in TorchAO


CaoSY:
## Motivation

Quantization-aware training for MoE models is becoming a production standard. **DeepSeek-V4** applies MXFP4 QAT to its MoE expert weights during post-training. **Kimi K2 Thinking** employs INT4 QAT on its MoE components during post-training, achieving lossless 2× speedup (as reported in the model card). There is also an open request for MXFP4 QAT support for **GPT-OSS** (a 120B MoE model) in torchao (issue [#3547](https://github.com/pytorch/ao/issues/3547)).

TorchAO's current QAT support partially addresses this need: it handles `nn.Linear` layers (Q, K, V, O projections in attention) via module swap (`FakeQuantizedLinear`). However, the MoE expert weights — where the majority of parameters reside — are 3D `nn.Parameter` tensors routed through ops such as `torch._grouped_mm`, `torch.bmm`, and `F.linear`, which fall entirely outside the existing QAT path.

`torchao.prototype.moe_training` has already established a parameter-swap pattern (tensor subclass on `param.data`) for low-precision MoE training that composes with FSDP2, Expert Parallel, and `torch.compile`. This RFC proposes extending QAT to MoE models using the same parameter-swap approach.

## Goal

Deliver MoE QAT support using the parameter-swap pattern from `moe_training`, starting with FP8 row-wise weight-only fake quantization and designed for future extension to other precisions and activation QAT.

### Features Plan

| Precision | Weight-only QAT | Activation+Weight QAT |
|-----------|----------------|----------------------|
| FP8 rowwise | Planned | Planned |
| Int4 weight-only | Planned | — |
| Int8 dynamic activation + IntX weight | Planned | Planned |
| MXFP8 / MXFP4 | Planned | Planned |
| NVFP4 | Planned | Planned |

Additionally, we plan to support a dual-precision QAT path: fake-quantize from the original precision to an ultra-low precision (e.g., MXFP4) for storage, then dequantize to a low precision (e.g., MXFP8) for computation. This decouples storage precision from compute precision and enables reusing the existing low-precision training pipeline which is usually customized by users. This is inspired by **DeepSeek-V4**, where the path is FP32 → MXFP4 (1×32) → FP8 (128×128).

### Compatibility

- **FSDP2**, **Expert Parallel**, **torch.compile**: planned
[...]

PR #4464, opened 2026-06-04
Add FP8 Row-wise QAT for MoE

CaoSY:
Implements the FP8 row-wise path of https://github.com/pytorch/ao/issues/4408. See #4408 for full RFC scope.

## Implemented

- `MoEQATConfig` with `params_filter_fn` for per-parameter filtering
- `FakeQuantizedWeightWrapperBaseTensor` — base tensor subclass with `__torch_dispatch__`, FSDP2 hooks, flatten/unflatten
- `Float8FakeQuantizedWeightWrapperTensor` — subclass with `__torch_function__` that intercepts `mm`, `bmm`, `linear`, `matmul`, `addmm`, `_grouped_mm` and applies FP8 row-wise fake quantization
- Deferred fake quantization: slicing/indexing/reshaping preserves the wrapper; quantization only at computation ops
- Activation fake quantization (opt-in via `activation_config`)
- `_is_expert` example filter function for targeting MoE submodules
- Prepare → convert lifecycle with `quantize_` integration
- FSDP2 hooks (`fsdp_pre_all_gather` / `fsdp_post_all_gather`) with mixed-precision support

## Test Suite

Unit tests under `test/prototype/moe_qat/`:

| File | Scope |
|---|---|
| `test_wrapper_tensor.py` | `__init__`, `__deepcopy__`, flatten/unflatten, meta, FSDP hooks |
| `test_wrapper_ops.py` | `__torch_dispatch__`, `__torch_function__`, op tests, `torch.compile` |
| `test_moe_qat_config.py` | Config validation |
| `test_moe_qat_transform.py` | Prepare/convert lifecycle, filter functions |
| `test_training.py` | Full-model QAT training SQNR; QAT + `torch.compile` on MoE |
| `test_distributed.py` | FSDP, EP, TP, EP+TP, FSDP+TP; `c10d.scatter_` unit test |

Run all unit tests with `./test/prototype/moe_qat/test_everything.sh`. Distributed tests are included in that script via `./test/prototype/moe_qat/test_distributed.sh` (requires `torchrun --nproc_per_node=4`). All tests run on both CPU and GPU via `target_devices` parametrization.

## Known Limitations (not passing, work in progress)

- **`torch.compile` + activation QAT with DTensor**: Activation fake-quant under Dynamo tracing fails with `RuntimeError: backward() called inside a functorch transform`
- **`torch.compile` + EP+TP**: `device_mesh._get_submesh` is not a recognized inductor op
- **`torch.compile` + `fullgraph=True` + activation QAT + `use_grouped_mm=False`**: Per-expert for-loop has data-dependent token counts that cannot be guarded on
- **convert after FSDP forward+backward**: Not supported — FSDP2 manages `param.data` internally during the all-gather cycle
- **activation QAT + `_grouped_mm`**: Forward SQNR becomes NaN randomly across runs when activation fake quantization is combined with grouped_mm (root cause not yet isolated)
- `_stable_hash_for_caching` not implemented — produces a warning under the nightly PyTorch build. Correctness is unaffected; the default fallback uses pickle-based hashing
- Code is not formatted via ruff yet. Will do after bugs are cleared.


andrewor14:
Hi @CaoSY, I left a high level comment here: https://github.com/pytorch/ao/issues/4408#issuecomment-4674456719. Please take a look, thanks.

PR #4465, opened 2026-06-04
Implement aten.cat for Int8QuantizedTrainingLinearWeight (#2619)

### LINKED ISSUE #2619: Int8QuantizedTrainingLinearWeight missing aten.cat support for distributed training
When running `accelerator.prepare(model)` on a model with int8 quantisation applied, an error in pytorch emits that we're missing `aten.cat` in the implementation.

I've taken a wack at hacking it together, but I'm not certain it's correct - it does "work" but i had to dequantize where maybe I should not have to, but just didn't know a better way.

my current workaround (sorry, it's horrible):

```py
try:
    from torchao.prototype.quantized_training.int8 import Int8QuantizedTrainingLinearWeight, implements
    import torch
    
    test_tensor = Int8QuantizedTrainingLinearWeight.from_float(torch.randn(2, 2))
    try:
        torch.cat([test_tensor, test_tensor], dim=0)
        print("aten.cat already implemented for Int8QuantizedTrainingLinearWeight")
    except NotImplementedError:
        @implements(torch.ops.aten.cat.default)
            """Implement concatenation for Int8QuantizedTrainingLinearWeight (needed for DDP)."""
            tensors = args[0]
            dim = args[1] if len(args) > 1 else kwargs.get('dim', 0)
            
            has_int8 = any(isinstance(t, Int8QuantizedTrainingLinearWeight) for t in tensors)
            if not has_int8:
                return func(tensors, dim, **kwargs)
            
            ndims = set()
            for t in tensors:
                if isinstance(t, Int8QuantizedTrainingLinearWeight):
                    ndims.add(2)
                else:
                    ndims.add(t.ndim)
            
            if len(ndims) > 1:
                flattened = []
                for t in tensors:
                    if isinstance(t, Int8QuantizedTrainingLinearWeight):
                        flattened.append(t.dequantize().flatten())
                    else:
                        flattened.append(t.flatten())
                
                return func(flattened, 0, **kwargs)
            
            all_int8 = all(isinstance(t, Int8QuantizedTrainingLinearWeight) for t in tensors)
            
            if not all_int8:
                dequantized = []
                for t in tensors:
                    if isinstance(t, Int8QuantizedTrainingLinearWeight):
                        dequantized.append(t.dequantize())
                    else:
                        dequantized.append(t)
                return func(dequantized, dim, **kwargs)
            
            if dim == 0:
                int_data_list = [t.int_data for t in tensors]
                scale_list = [t.scale for t in tensors]
                
                cat_int_data = torch.cat(int_data_list, dim=0)
                cat_scale = torch.cat(scale_list, dim=0)
                
                return Int8QuantizedTrainingLinearWeight(cat_int_data, cat_scale)
            
            elif dim == 1:
                int_data_list = [t.int_data for t in tensors]
                cat_int_data = torch.cat(int_data_list, dim=1)
                
                return Int8QuantizedTrainingLinearWeight(cat_int_data, tensors[0].scale)
            
            else:
                dequantized = [t.dequantize() for t in tensors]
                return func(dequantized, dim, **kwargs)
        
except ImportError:
    # torchao int8 not being used
    pass
except Exception as e:
    print(f"Warning: Failed to monkeypatch int8 cat operation: {e}")
    print("DDP may not work with int8 quantization")
```

lollinng:
### Problem (#2619)

Wrapping an int8-quantized-training model for distributed training — e.g. `accelerator.prepare(model)` or DDP — calls `torch.cat` on the parameters, but `Int8QuantizedTrainingLinearWeight` had no `aten.cat.default` handler, so it raised:

```
NotImplementedError: ... attempting to run unimplemented operator/function: aten.cat.default ...
```

### Fix

Add an `aten.cat` handler as the inverse of the existing `aten.split.Tensor` handler. The subclass uses **row-wise** quantization (`int_data` is `[rows, cols]` int8, `scale` is `[rows]`), so concatenating along **dim 0** (the row dim) just concatenates `int_data` and `scale` — each row keeps its own scale, so the result is **exact with no dequantization**.

`dim != 0` (and mixed input types) raise `NotImplementedError`, mirroring `split`'s dim-0-only contract: a column-wise cat would mix columns governed by different per-row scales and can't be represented as a single `Int8QTLinearWeight`. (This matches the concern raised in the issue thread.)

### Testing done

Added `test_int8_weight_only_cat`, parametrized over `_DEVICES` (runs on CPU here and CUDA in CI). Verified locally on CPU with the repo source:

- **Before** (no handler): `torch.cat([w1, w2], dim=0)` → `NotImplementedError`.
- **After**: returns `Int8QuantizedTrainingLinearWeight(shape=(10, 8))`, and `out.dequantize()` matches `torch.cat([w1.dequantize(), w2.dequantize()], dim=0)` exactly; `dim=-2` resolves to the row dim; `dim=1` raises.

```
test_int8_weight_only_cat_device_cpu PASSED
```

`ruff check` and `ruff format --check` (0.11.6) clean on both files.

The INT8 quantized-training path has no compute-capability gate, so this runs identically on CPU and on older GPUs. AI assistance (Claude Code) was used; I reviewed the change and ran the test.

Fixes #2619

### DIFF
--- a/test/prototype/test_quantized_training.py
+++ b/test/prototype/test_quantized_training.py
@@ -98,6 +98,37 @@ class TestQuantizedTraining(TestCase):
         if bias:
             torch.testing.assert_close(linear_fp32.bias.grad, linear_int8.bias.grad)
 
+    @parametrize("device", _DEVICES)
+    def test_int8_weight_only_cat(self, device):
+        from torchao.prototype.quantized_training.int8 import (
+            Int8QuantizedTrainingLinearWeight,
+        )
+
+        w1 = Int8QuantizedTrainingLinearWeight.from_float(
+            torch.randn(4, 8, device=device)
+        )
+        w2 = Int8QuantizedTrainingLinearWeight.from_float(
+            torch.randn(6, 8, device=device)
+        )
+
+        out = torch.cat([w1, w2], dim=0)
+        self.assertIsInstance(out, Int8QuantizedTrainingLinearWeight)
+        self.assertEqual(out.shape, (10, 8))
+        # row-wise scales are preserved, so the cat is exact (no requantization)
+        torch.testing.assert_close(
+            out.dequantize(),
+            torch.cat([w1.dequantize(), w2.dequantize()], dim=0),
+        )
+
+        # negative dim resolves to the row dim
+        torch.testing.assert_close(
+            torch.cat([w1, w2], dim=-2).dequantize(), out.dequantize()
+        )
+
+        # cat along the feature dim is not representable with per-row scales
+        with self.assertRaises(NotImplementedError):
+            torch.cat([w1, w1], dim=1)
+
     @parametrize("leading_dims", [(), (2,), (2, 4)])
     @parametrize("bias", [False, True])
     @parametrize("device", _DEVICES)
--- a/torchao/prototype/quantized_training/int8.py
+++ b/torchao/prototype/quantized_training/int8.py
@@ -278,6 +278,31 @@ def _(func, types, args, kwargs):
     return out
 
 
+# Inverse of aten.split.Tensor above. Needed for distributed training, where
+# wrapping a model with DDP / `accelerator.prepare()` concatenates parameters.
+# With row-wise quantization each row carries its own scale, so concatenating
+# along dim 0 (the row dim) just concatenates int_data and scale and needs no
+# dequantization. Concatenating along dim 1 would mix columns governed by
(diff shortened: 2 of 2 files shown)


vkuzo:
@claude review

vkuzo:
lgtm, can you sign the CLA agreement?

lollinng:
Hey I have signed

vkuzo (approved):

PR #4466, opened 2026-06-05
pt2e: preserve mutable buffer inputs during prepare

mansnils:
Skip inserting observers on arg0 of mutating ops when that argument traces back to a named buffer. Arg0 is the tensor being mutated, so it must remain the original module buffer.

Without this, prepare+convert rewrites:

res = index_put_(dequantize(quantize(buf)), idx, dequantized_x)

instead of:

res = index_put_(buf, idx, dequantized_x)

The first form mutates a dequantized temporary rather than the buffer, so re-export no longer reports buffers_to_mutate. Add coverage for index_put_ to verify buffers_to_mutate is preserved after convert_pt2e.

Co-Authored-By: Tom Allsop <tom.allsop@arm.com>

Fix for: https://github.com/pytorch/ao/issues/4138

### DIFF
--- a/torchao/quantization/pt2e/prepare.py
+++ b/torchao/quantization/pt2e/prepare.py
@@ -432,6 +432,7 @@ def _is_mutable_buffer_arg(
         return False
 
     schema = node.target._schema
+    # This covers in-place ops such as copy_, put_, scatter_, fill_, add_
     if (
         not schema.arguments
         or not schema.arguments[0].alias_info
--- a/torchao/quantization/pt2e/prepare.py
+++ b/torchao/quantization/pt2e/prepare.py
@@ -418,9 +418,9 @@ def _get_obs_or_fq_map(
 
 
 def _is_mutable_buffer_arg(
-    model: torch.nn.Module,
     node: Node,
     arg: Node,
+    buffer_names: set[str],
 ) -> bool:
     """Return whether arg is a mutating op destination backed by a module buffer."""
     if (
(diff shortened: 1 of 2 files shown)


andrewor14:
@claude review

mansnils:
I am not able to add labels. Should we implement Claude's suggestions?

vkuzo:
yeah, (1) and (2) from Claude's suggestions would be good to implement!

torchao/quantization/pt2e/prepare.py
```diff
@@ -416,6 +417,34 @@ def _get_obs_or_fq_map(
...
+        return False
+
+    if "copy_" not in str(node.target) and "put_" not in str(node.target):
+        return False
+
+    producer_nodes = collect_producer_nodes(arg)
+    if producer_nodes is None:
+        return False
+
+    named_buffers = dict(model.named_buffers(remove_duplicate=False))
```
vkuzo: your call on whether to implement claude's suggestion here, I think the runtime of the pass before/after is important.  If runtime doesn't meaningfully change, fine to leave IMO.

andrewor14:
@claude can you check if this PR fixes https://github.com/pytorch/ao/issues/4138?

andrewor14 (approved):
Stamping to unblock. @mansnils please feel free to implement (1) and (2), I think they're good suggestions. For (1) please add a comment for example ops this covers, e.g. put_, copy_

mansnils:
> yeah, (1) and (2) from Claude's suggestions would be good to implement!

Done

mansnils:
> Stamping to unblock. @mansnils please feel free to implement (1) and (2), I think they're good suggestions. For (1) please add a comment for example ops this covers, e.g. put_, copy_

Done

andrewor14:
@claude latest changes good to go?

andrewor14:
Thanks!

liangel-02:
@mansnils sorry reverting this because some internal tests are failing, see https://github.com/pytorch/ao/pull/4531

xingguo01:
@liangel-02 highly appreciated if you can share what regressions this change caused in your internal test so we can address it. To add a bit of context, this is the current bottleneck when we develop kv cache of SLM for TOSA/VGF INT PATH in Executorch.

PR #4467, opened 2026-06-05
Add workflow to triage main CI failures with Claude

andrewor14:
**Summary:** Adds .github/workflows/ci-failure-issue.yml. When a CI workflow fails
on a push to `main`, it asks Claude to triage the failure and file (or update) a
`ci-failure` GitHub issue with a root-cause analysis and suggested fix.

**Workflow:**
1. Triggers on `workflow_run: completed` for the watched workflows, gated to real
   failures on main pushes (conclusion == failure, head_branch == main,
   event == push).
2. Pre-check: if an open `ci-failure` issue already references the failing commit,
   add a one-line "also failed" note and stop -- so one bad commit that breaks
   many workflows yields a single issue, not one per workflow.
3. Otherwise, check out the failing commit and assume the Bedrock OIDC role
   (environment: bedrock), similar to _claude-code.yml in test-infra.
4. Claude reads the failing logs (treated as untrusted input), dedups against open
   issues, and emits a structured decision that a separate job acts on.

**Security precautions:**
- Split into two jobs so untrusted CI logs and the issue-write token never share a
  context: read-only `triage` reads logs, write-only `act` posts issues.
- A machine-readable dedup marker (`<!-- ci-failure-sha:<sha> -->`) is matched
  exactly instead of a bare SHA substring, preventing dedup spoofing/false matches.
- All posted text is run through a best-effort secret scrub that redacts common
  token/key shapes before it reaches a public issue.
- A rolling-24h burst cap limits new issue creation so a runaway or log-injection
  can't spam the tracker.
- The prompt treats CI logs as untrusted, ignores embedded instructions, and
  forbids quoting credentials.
- Context and inputs are passed via env vars (never inlined) to avoid shell
  injection, with the SHA, run id, and issue number validated.

A `workflow_dispatch` trigger is included so the workflow can be triggered
manually against a specific failing run for testing/debugging.

### DIFF
--- a/.github/workflows/ci-failure-issue.yml
+++ b/.github/workflows/ci-failure-issue.yml
@@ -1,0 +1,467 @@
+name: Triage CI failure on main
+
+# When a watched CI workflow fails on a push to `main`, triage it into a single
+# `ci-failure` issue (root cause + suggested fix), or comment on the existing one.
+#
+# Split into two jobs so the untrusted CI logs and the issue-write token never
+# meet (see PR #4467 review):
+#   1. triage -- read-only. A cheap pre-check skips the rest when an open issue
+#      already tracks this commit or the daily issue cap is hit; otherwise Claude
+#      reads the logs and emits a structured decision (--json-schema) as an
+#      artifact. Holds NO `issues: write`, so an injected log can't write anything.
+#   2. act -- write. Claude-free code that validates that decision (or the
+#      pre-check result) and performs the single issue create/comment.
+#
+# So one bad commit breaking many workflows yields one issue, not one per workflow.
+# Auth reuses test-infra's Bedrock OIDC role (environment: bedrock); `act` writes
+# via its GITHUB_TOKEN. Patterned on pytorch/pytorch's claude-autorevert-advisor.yml.
+
+on:
+  workflow_run:
+    # Must match each workflow's `name:`. These run on pushes to `main`
+    # (`xpu-test` excluded -- tags/cron only). Keep it an explicit list, no
+    # wildcards: the name is interpolated into the prompt.
+    workflows:
+      - "Run Regression Tests"
+      - "Run Regression Tests on ROCm"
+      - "Run Regression Tests (aarch64)"
+      - "Run 1xH100 Tests"
+      - "Run 4xH100 tests"
+      - "Run 1xL4 Tests"
+      - "Run TorchAO Experimental MPS Tests"
+      - "Code Analysis with Ruff"
+      - "Build Docs"
+      - "Build Linux Wheels (x86)"
+      - "Build Linux Wheels (AArch64)"
+      # CodeQL is GitHub default-setup scanning (no workflow file), so
+      # `workflow_run` may not fire for it. Listed optimistically -- verify.
+      - "CodeQL"
+    types: [completed]
+  # Manual trigger for debugging/backfill of a specific failing run. There is no
+  # `workflow_run` context here, so the same fields are taken from these inputs
+  # (resolved in the "Resolve failing-run context" step below).
+  workflow_dispatch:
+    inputs:
+      run_id:
+        description: "Failing workflow run id to triage"
+        required: true
+        type: string
+      head_sha:
+        description: "Commit SHA the run failed on"
+        required: true
+        type: string
+      workflow_name:
+        description: "Name of the failing workflow"
+        required: true
+        type: string
+
+# One triage per commit at a time (spanning both jobs), so siblings don't each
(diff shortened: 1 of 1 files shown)


vkuzo:
@claude review, focus on any security attack vectors that could be present with this approach

andrewor14:
@claude review again?

andrewor14:
@claude `ci-failure` exists, I added this manually. Also added a comment to explain "breadcrumb" comment is best effort. Any other thoughts?

andrewor14:
@drisspg can you help take a look?

.github/workflows/ci-failure-issue.yml
```diff
@@ -0,0 +1,177 @@
...
+#
+# A cheap pre-check skips the (expensive) Claude step when an open issue already
+# references the failing commit, so one bad commit that breaks many workflows
+# yields a single issue, not one per workflow.
+#
+# Auth reuses test-infra's Bedrock OIDC role (environment: bedrock); Claude
+# creates/comments via the job's GITHUB_TOKEN.
+
+on:
+  workflow_run:
```
izaitsevfb, line 20: you probably want to add workflow_dispatch here to be able to debug this workflow

.github/workflows/ci-failure-issue.yml
```diff
@@ -0,0 +1,177 @@
...
+                 - Representative error: ...
+
+                 ## Root cause
+                 ...
+
+                 ## Classification
+                 real regression / flaky test / infra failure -- with justification
+
+                 ## Suggested fix
+                 <description; diff if you can>
```
izaitsevfb: please add usage reporting, [see example here](https://github.com/pytorch/pytorch/blob/8bc0aa73bdf79a93193f5307d6658ac94ff0547b/.github/workflows/claude-autorevert-advisor.yml#L167)

```
      - name: Upload usage metrics
        if: always()
        uses: pytorch/test-infra/.github/actions/upload-claude-usage@main
```

izaitsevfb (commented):
mandatory: please add usage tracking.

I'd also highly recommend considering splitting into two dependent jobs (first one has read-only permissions, produces structured artifact, that the second job verifies and acts upon). that is relatively cheap and tightens the security.


Also consider the following list (produced by my ai-agents):

1. Burst/rate cap in the pre-check (24h creation ceiling)
2. Secret-scrub before posting log excerpts to a public issue
3. Machine-readable dedup marker — write `<!-- ci-failure-sha:<sha> -->` and match on that exact marker instead of body-substring contains(SHA); kills the spoof/alert-DoS where a pre-opened issue containing a target SHA silences triage (codex+fable LOW).
4. Bump model claude-opus-4-6 → 4-8 (already flagged by the claude bot; triage quality over noisy logs benefits directly).

andrewor14:
@claude Please review the latest changes again, can you double check if I addressed all outstanding comments?

andrewor14:
@izaitsevfb can you take another look? Thanks.

izaitsevfb (approved):
stamping to unblock, but I'd strongly recommend to address the following:


1. The dedup marker is never enforced mechanically. 

2. `act` never verifies/injects the marker. If Claude omits/typos it (or is injected into dropping it), the issue is invisible to the next pre-check 


currently the prompt-injected claude could update / replace any issue. the better design would be to include the marker **programmatically** into the issue / comment (during `act`),  and when updating the issues (during `act`) verify that the marker is present

.github/workflows/ci-failure-issue.yml
```diff
@@ -0,0 +1,425 @@
...
+          fi
+
+          # Not tracked yet -> we would file a NEW issue. Enforce the rolling-24h
+          # ceiling first so a burst can't spam the tracker. `--state all` counts
+          # created issues regardless of whether they were since closed.
+          since=$(date -u -d '24 hours ago' +%Y-%m-%dT%H:%M:%SZ)
+          recent=$(gh issue list --repo "$REPO" --state all --label ci-failure \
+            --limit 200 --json createdAt \
+            --jq "[.[] | select(.createdAt >= \"${since}\")] | length")
+          if [ "$recent" -ge "$MAX_ISSUES_PER_DAY" ]; then
```
izaitsevfb, line 184: burst cap is TOCTOU-racy across SHAs — different commits run in different concurrency groups; each sees 9 recent, all create → cap of 10 overshoots. ideally re-check in `act` before create

andrewor14:
@claude review latest changes, are all comments addressed?

vkuzo:
I took a look manually, excited about the automation. My main concern is the security aspect but that seems ok due to precedent in core + discussion / iteration on this PR.  Let's try it!

vkuzo (approved):

PR #4389, opened 2026-05-13
[moe training] Optimize tensorwise FP8 grouped GEMM

rishisinhanj:
## Summary
This adds a tensorwise FP8 grouped GEMM path for MoE training and optimizes its quantization overhead enough to make the expert backward path materially faster on DeepSeek-V3-sized shapes.

The implementation uses single-scalar tensorwise scaling, caches/materializes the layouts needed by `torch._scaled_grouped_mm`, and replaces the slow layout materialization plus contention-heavy tensorwise amax path with tuned Triton kernels. The dual-layout 2D quantizer now writes row-major and column-major FP8 outputs from tiled blocks, while the amax path uses a staged two-kernel reduction for DeepSeek-sized tensors with an atomic fallback for tensors with too many partial blocks.

## What changed
- Add TensorWise FP8 grouped GEMM plumbing and recipe selection for MoE training.
- Add fused Triton 2D and 3D tensorwise quantization kernels, including row-major and column-major layout materialization for grouped GEMM requirements.
- Add a tiled 2D dual-layout quantization kernel so one pass materializes both grouped-GEMM layouts without a separate copy/transpose path.
- Add a staged tensorwise amax reduction and route both 2D activation/grad quantization and 3D expert-weight quantization through it.
- Add focused correctness tests for staged amax vs the prior atomic path, 2D and dual-layout quantization parity, and grouped-GEMM forward/backward with nonuniform offsets.
- Add a DeepSeek-V3 MoE grouped-GEMM benchmark covering BF16, rowwise FP8, tensorwise FP8, fwd/bwd/fwd+bwd, JSON output, and optional trace capture.

## Performance notes
Tiled dual-layout quantization removes the previous explicit layout materialization overhead. On the original TensorWise quantization benchmark, the tiled kernel reduced the dual-layout quantization path from roughly millisecond-scale PyTorch/layout work to a single fused tiled Triton path for the FP8 row-major and column-major outputs.

Focused staged-amax benchmark on MI355X, DeepSeek bs=2 local shapes:

| Target | Shape | Amax Speedup | 2D Quant Speedup | Dual-Layout Quant Speedup |
|---|---:|---:|---:|---:|
| gate/up input | `65536x7168` | `1.86x` | `1.17x` | `1.19x` |
| gate/up grad out | `65536x2048` | `1.36x` | `1.09x` | `1.06x` |
| down input | `65536x2048` | `1.36x` | `1.09x` | `1.06x` |
[...]

PR #4469, opened 2026-06-08
[moe training] Use non-atomic Triton scatter for permute_and_pad backward

rishisinhanj:
## Summary

This is a **general, recipe-independent** optimization of the MoE token-permute
backward. `permute_and_pad` gathers `x_padded[permuted_indices, :]`; its autograd
backward is PyTorch's default `indexing_backward_kernel`, an **atomic** scatter.
Since `permuted_indices` is a bijection over valid (non-padding) positions, the
atomics are unnecessary. We replace the backward with `_PermuteBF16FwdBF16Bwd`,
whose backward is a **non-atomic Triton scatter** (`_triton_permute_bwd`). The
forward is unchanged.

Because the permute runs on **BF16 tokens upstream of quantization**, the fix is
independent of the scaling recipe — it benefits the **rowwise**, **tensorwise**,
and **MXFP8** paths alike (the MXFP8 permute path `_PermuteMXFP8FwdHPBwd` already
uses `_triton_permute_bwd`). No torchtitan changes are needed;
`TorchAOTokenDispatcher` already calls `permute_and_pad`.

This builds on the tensorwise grouped-GEMM path from #4389. Per the discussion
there, the goal of landing the tensorwise path is to establish a **working
tensorwise baseline on ROCm** to compare rowwise and the in-progress MXFP8
implementation against — not to beat BF16 with tensorwise. This permute fix is
shared infrastructure underneath all three recipes.

## What this fixes

In FP8 MoE training the routing tokens are padded to an alignment multiple for
the grouped GEMM, so the permute backward runs an atomic `indexing_backward`
over the **padded** tensor (grid `[8320,28,1]`). That makes FP8's routing
backward roughly **2x BF16's** — purely an artifact of padding + atomics, not
real work. This PR removes that FP8-specific penalty and restores **parity with
BF16's routing backward**.

## What changed

- Add `_PermuteBF16FwdBF16Bwd` + `_triton_permute_bwd` op/kernel (skips `-1`
  padding rows); `permute_and_pad` uses it.
- Fixes: load grad with `other=0.0` (not the `-1` sentinel); `.contiguous()`
  guards for non-contiguous autograd grads.
- **New test** `test/prototype/moe_training/ep/test_permute_backward.py`:
  backward parity vs the PyTorch reference (bit-exact gradients, incl.
  zero-token experts).
- **New benchmark** `benchmarks/prototype/moe_training/ep/bench_permute_backward.py`:
  reproduces the default-`indexing_backward`-vs-Triton speedup table below at
  DeepSeek-V3 shapes.

## Why scoped to the permute backward

We prototyped Triton backwards for the other two routing ops and dropped both —
PyTorch's defaults already handle them well:

- **Unpermute** (`out[idx] = routed; out[:-1]`): default backward is a plain
  **gather**, already efficient. The Triton version was ~break-even and required
  a torchtitan change. Redundant.
- **Dispatch gather** (`x[token_indices]` with top_k duplicates): a true
  scatter-add; PyTorch sorts the indices and does a non-atomic segmented
  reduction, which **beat** our atomic Triton kernel (~2x slower). Regressive.

Only the permute is a genuine win: an atomic scatter over a **bijection**, where
a non-atomic Triton scatter is strictly better.

## Performance (kernel-level)

MI355X (gfx950), DeepSeek-V3 671B 4-layer, D=7168, 256 experts, EP=8.

Microbenchmark — `permute_and_pad` backward, default `indexing_backward` vs Triton:

| M (tokens) | default (us) | Triton (us) | speedup |
|---:|---:|---:|---:|
| 4096  | 1137 | 149 | 7.7x |
| 8192  | 1321 | 274 | 4.8x |
| 16384 | 2121 | 547 | 3.9x |

E2E trace — routing-backward `indexing_backward`, restored to BF16 parity:

| config | `indexing_backward` in backward |
|---|---|
| BF16 (reference) | 9.5 ms |
| FP8 rowwise — before | 17.7 ms |
| FP8 rowwise — this PR | 9.4 ms + 0.9 ms Triton |
| FP8 tensorwise — before | 17.3 ms |
| FP8 tensorwise — this PR | 9.5 ms + 0.9 ms Triton |

Per step the permute backward drops from ~8.5 ms (4x atomic `indexing_backward`,
~2.1 ms/call) to ~1.1 ms (`_triton_permute_bwd`, ~9x per call).

**Scope note:** this removes the FP8-specific permute penalty (~7-9 ms/step
backward) and is a parity fix; it does not by itself close the FP8 vs BF16 E2E
gap. That remaining gap stems largely from the **layout requirements of
`torch._scaled_grouped_mm`** — column-major operands, 16-element alignment
padding, and materializing both row- and column-major FP8 copies — which drive
the extra `chunk_cat` and device-to-device copies in the FP8 path.

## Test plan

```
pytest -q test/prototype/moe_training/ep/test_permute_backward.py   # 5 passed (MI355X)
python benchmarks/prototype/moe_training/ep/bench_permute_backward.py
```

### DIFF
--- a/benchmarks/prototype/moe_training/ep/bench_permute_backward.py
+++ b/benchmarks/prototype/moe_training/ep/bench_permute_backward.py
@@ -69,10 +69,18 @@ def bench(M, D, num_experts, ep_degree, warmup=20, iters=100):
 
 def main():
     print(f"Device: {torch.cuda.get_device_name(0)} | torch {torch.__version__}\n")
-    print("permute_and_pad backward: default indexing_backward vs Triton non-atomic scatter")
+    print(
+        "permute_and_pad backward: default indexing_backward vs Triton non-atomic scatter"
+    )
     hdr = ("M", "D", "experts", "ep", "default(ms)", "triton(ms)", "speedup")
-    print(f"{hdr[0]:>7}{hdr[1]:>7}{hdr[2]:>9}{hdr[3]:>4}{hdr[4]:>13}{hdr[5]:>12}{hdr[6]:>9}")
-    for M, D, ne, ep in [(4096, 7168, 256, 8), (8192, 7168, 256, 8), (16384, 7168, 256, 8)]:
+    print(
+        f"{hdr[0]:>7}{hdr[1]:>7}{hdr[2]:>9}{hdr[3]:>4}{hdr[4]:>13}{hdr[5]:>12}{hdr[6]:>9}"
+    )
+    for M, D, ne, ep in [
+        (4096, 7168, 256, 8),
+        (8192, 7168, 256, 8),
+        (16384, 7168, 256, 8),
+    ]:
         d, t = bench(M, D, ne, ep)
         print(f"{M:>7}{D:>7}{ne:>9}{ep:>4}{d:>13.4f}{t:>12.4f}{d / t:>8.2f}x")
 
--- a/benchmarks/prototype/moe_training/ep/bench_permute_backward.py
+++ b/benchmarks/prototype/moe_training/ep/bench_permute_backward.py
@@ -1,0 +1,81 @@
+"""Microbenchmark: permute_and_pad backward — Triton non-atomic scatter vs the
+default PyTorch indexing_backward_kernel (atomic scatter).
+
+The permute is applied to BF16 tokens upstream of FP8 quantization, so this
+kernel is identical for both fp8_rowwise and fp8_tensorwise recipes.
+
+Run (inside the training container, fresh torchao installed):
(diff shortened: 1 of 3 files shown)


rishisinhanj:
@danielvegamyhre Could we get a comment or changes on this pr to be successfully merged? Also the previous #4389 Thanks!

rishisinhanj:
> @danielvegamyhre Could we get a comment or changes on this pr to be successfully merged? Also the previous #4389 Thanks!

gentle ping @danielvegamyhre

Issue #4376, opened 2026-05-11
Float8DynamicActivationFloat8WeightConfig is inference-only — backward fails on torch._scaled_mm; document or add training-aware variant


genno-whittlery:
## TL;DR

`Float8DynamicActivationFloat8WeightConfig` (fp8 W8A8 inference config) **forward** works fine in a LoRA-training setup with the `filter_fn=skip_lora_modules` pattern. But `loss.backward()` crashes immediately with:

```
RuntimeError: derivative for aten::_scaled_mm is not implemented
```

Either:
1. Documentation could be clearer that this config is inference-only (the README and docstring don't make that obvious; people doing LoRA-on-quantized-base hit it).
2. A training-aware variant of `Float8DynamicActivationFloat8WeightConfig` (custom forward+backward `torch.autograd.Function` around `torch._scaled_mm`) would unlock a meaningful speedup for LoRA training on Blackwell — currently the only working *training* fp8 path through torchao is `torchao.float8.convert_to_float8_training`, which keeps weights in bf16 (no memory savings — non-starter for dual-32 GB consumer setups).

OneTrainer ships a working reference implementation at https://github.com/Nerogar/OneTrainer/blob/master/modules/module/quantized/LinearW8A8.py — `LinearW8A8` with hand-rolled forward/backward `torch.autograd.Function` around `torch._scaled_mm` (fp8) and `torch._int_mm` (int8), sustained 1.75 / 0.95 s/it on 2× RTX 5090 for FLUX.2 LoRA at rank 32.

## Context

Surfaced while validating a dual-GPU model-parallel FLUX.2 LoRA training path across four trainers (https://github.com/genno-whittlery/flux2-dual-gpu-lora — ai-toolkit, musubi-tuner, OneTrainer, DiffSynth-Studio).

Three of the four use torchao `Float8WeightOnlyConfig` (memory savings, bf16 matmul) — that path works cleanly with the LoRA-skip `filter_fn` from https://github.com/huggingface/peft/issues/3223. Training sustains 2.22–2.85 s/it depending on trainer.

The fourth (OneTrainer) uses its own `LinearW8A8` for proper W8A8 compute on Blackwell tensor cores, hitting 0.95 s/it (int8) or 1.75 s/it (fp8). That's a real 2-3× compute speedup over the bf16-matmul band.

[...]

Issue #4022, opened 2026-03-06
Next steps for MXFP8 training


danielvegamyhre:
Functionality improvements
- [ ] [P0] Dynamic per group padding/unpadding kernels
    - Needed as short/medium term solution so users have performant-ish way to pad token groups to multiples of mxfp8 scaling block size (32), which is required to use the differentiable mxfp8 grouped mm. Using naive torch padding requires d2h sync and kills perf, so custom kernels are needed, which causes usage friction.
    - Status: WIP https://github.com/pytorch/ao/pull/3998

Performance improvements
- [ ] [P0] Fused all2all dispatch + padding kernel
    - Needed to avoid the extra copy incurred by the standalone padding kernel described above, which hurt our speedup. The benefit of this approach is that, with the all-to-all dispatch, the receiver ranks are already going to be allocating a buffer for the incoming tokens. If we write those tokens to locations aligned with multiples of 32, we avoid the need for this expensive extra copy.    
    - While we're doing this, we can also write incoming tokens grouped by local expert, instead of grouped by remote/source rank, in order to avoid the token shuffle kernel step
    - Status: WIP
- [ ] [P1] Faster 3d weight quantization kernel for backward pass dgrad computation with RCEIL scaling, that writes scales directly to ((32,4),4) layout for tcgen05 mma 
    - Current: ~5 tb/s, goal: 6.4 tbs
    - Currently writes scales in row major, requires additional lightweight kernel for per group blocked layout 
- [ ] [P1] Faster dim0 quantization kernel with RCEIL scaling that writes scales directly to ((32,4),4) layout for tcgen05 mma 
    - Current: ~5.5 tb/s, Goal: ~6.4 tb/s.  
    - Currently writes scales in row major, requires additional lightweight kernel for per group blocked layout 
- [ ] [P1] dim1 quantization kernel with RCEIL scaling that writes scales directly to ((32,4),4) layout for tcgen05 mma (current: 5.9 tb/s but writes scales in row major, requires extra kernel for per group blocked layout)
    - Currently writes scales in row major, requires additional lightweight kernel for per group blocked layout

PR #4471, opened 2026-06-09
[mxfp8] Skip weight gradient for frozen weights in mx_mm backward

ultism:
## Summary

When the weight does not require grad — a frozen base in LoRA or frozen-layer finetuning — autograd discards `grad_weight`, so computing it in `mx_mm.backward` (the wgrad GEMM plus its two `dim1` MXFP8 casts) is pure overhead.

This guards the weight-gradient computation on `ctx.needs_input_grad[1]` and returns `grad_weight=None` for a frozen weight. It's the same idiom already used by the `int8_mixed_precision` and `bitnet` training linears.

```python
# input_t @ grad_output = grad_weight
if not ctx.needs_input_grad[1]:
    grad_weight = None
elif wgrad_with_hp:
    ...
```

## Eager vs compile

This is an **eager-mode** win. Under `torch.compile`, AOTAutograd already specializes the backward on `requires_grad` and prunes the unused `grad_weight` subgraph via DCE, so the compiled frozen path is already optimal; this change just brings eager to parity. It does **not** change compiled behavior, and `ctx.needs_input_grad` is a Python bool (a trace-time constant), so the branch does not introduce a graph break.

Measured on an RTX 5060 Ti (sm120), frozen vs trainable `fwd+bwd` for a frozen linear, ratio = trainable / frozen (>1 means the wgrad work is skipped). Absolute numbers vary run-to-run with GPU clocks; the within-run ratio is the signal:

| path | shape (M, K, N) | eager (this PR) | eager (main) | compile (main) |
|---|---|---|---|---|
| LLaMA MLP up | 8192, 4096, 14336 | **1.54x** | 1.01x | 1.52x |
| LLaMA MLP dn | 8192, 14336, 4096 | **1.57x** | 1.00x | 1.24x |

So on `main` eager does not skip the wgrad (≈1.0x) while compile already does (≈1.2–1.5x); this PR closes the eager gap.

## Test

`test_mxfp8_linear_frozen_weight_skips_wgrad` (EMULATED + AUTO): a frozen weight gets `grad_weight=None`, a trainable one does not, `grad_input` is bit-identical between the two, and the frozen path performs strictly fewer `dim1` casts (verified by counting `_to_mxfp8_dim1_kernel_wrapper` calls). This fails on `main` (the frozen path still runs the wgrad casts) and passes here.

`test_mxfp8_linear_frozen_weight_compile` (EMULATED + AUTO): compiles the frozen module with `fullgraph=True` and checks the result matches eager and `weight.grad is None`, locking in compile-safety.

Verified locally on sm120: the full `test_mxfp8_linear.py` suite passes; `ruff check` and `ruff format --check` clean.

## Context

Frozen-base / parallel-adapter (LoRA) training on a quantized base is the motivating use case; see #4376 for the analogous float8 discussion and #4022 for the MXFP8 training roadmap. Touches the same `mx_mm.backward` as #4470 (a non-contiguous `grad_output` fix) but is independent of it.

### DIFF
--- a/test/prototype/moe_training/test_mxfp8_linear.py
+++ b/test/prototype/moe_training/test_mxfp8_linear.py
@@ -4,6 +4,8 @@
 # This source code is licensed under the BSD 3-Clause license found in the
 # LICENSE file in the root directory of this source tree.
 
+from unittest.mock import patch
+
 import pytest
 import torch
 import torch.nn.functional as F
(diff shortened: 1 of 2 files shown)


vkuzo:
@claude review

vkuzo:
lgtm and CI failures look unrelated. @ultism can you rebase, your other PR which we just merged has a conflict with this one. Thank you!

@andrewor14 can you help land this after I go OOO please? Thank you!

vkuzo (approved):

ultism:
Rebased onto current main (df97218e8), resolving the conflict with #4470 as requested — both test additions are kept (`test_mxfp8_linear_noncontiguous_grad_output` from #4470 and the two frozen-weight tests here). The `ctx.needs_input_grad[1]` guard now sits on the updated backward (with `wgrad_with_hp`). @vkuzo ready for another look, thanks!

PR #4473, opened 2026-06-09
Integrate DeepGEMM FP8 blockwise MoE backend

iamzainhuda:
## Summary

Adds an optional **DeepGEMM backend** for the FP8 blockwise MoE grouped GEMM, alongside the existing emulated path. When DeepGEMM is installed, `KernelPreference.AUTO` dispatches the forward, dgrad, and wgrad grouped GEMMs to DeepGEMM; otherwise it falls back to the emulated kernels. 

Two main parts are added here: backend-selection layer and DeepGEMM-layout FP8 quantization kernels that write DeepGEMM's exact data/scale contracts directly (no dispatch-time transpose/copy).

## DeepGEMM layout/scale contract vs TorchAO

DeepGEMM's grouped FP8 kernels are expect a different operand memory layout and scale orientation and differ from the layouts of TorchAO's existing blockwise quantizers (and `torch._grouped_mm`) produce. 
- TorchAO's public expert weight is `(E, K, N)` transposed in per-expert column-major layout, its 1x128 activation scales are stored as `(M_blocks, K)` (column-major-friendly for `torch._scaled_mm`), and its grouped RHS scales follow the `torch._grouped_mm` convention. 
- DeepGEMM instead wants: the **forward RHS** as `(E, N, K)` row-major with contiguous `K` and `(E, N//128, K//128)` scales; the **dgrad RHS** as `(E, K, N)` row-major with contiguous `N` and `(E, K//128, N//128)` scales; and for the **K-grouped wgrad**, each operand as a flat per-expert `(D, expert_tokens)` 
- K-major buffer (experts concatenated) with scales as `(D, total_valid_token_blocks)` — i.e. the token dimension becomes the GEMM's contraction axis and the scale axes are transposed relative to TorchAO's.

Rather than quantize in the TorchAO layout and then transpose/copy at dispatch time, we add kernels that write DeepGEMM's exact contract directly. 
- The weight quantizers read the contiguous `(E*N, K)` view of the column-major weight and emit DeepGEMM's `(E, N, K)`/`(E, K, N)` data and block-transposed scales straight from the cast — no separate layout-copy kernel. 
- For the wgrad operands there are two routes selected per operand by `_DEEPGEMM_DIRECT_K_GROUPED_QUANT_MIN_DIM`: wide operands take the **direct K-grouped quant** that writes the flat `(D, expert_tokens)` K-major buffer and `(D, blocks)` scales in one pass; narrower operands reuse TorchAO's existing transposed-LHS/RHS quantizers and are then flattened per-expert and have their scales transposed (`(M_blocks, K) → (K, M_blocks)`) at launch time to match DeepGEMM's K-major contract. The on-device `grouped_layout` (one int32 expert id per row, `-1` for padding) is what tells DeepGEMM's contiguous M-grouped kernel which expert each row belongs to.

### H100 per-kernel roofline (balanced tokens, M=32768, E=8, N=2048, K=7168)

Roofline refs: mem BW peak 3352 GB/s / achievable 3084 GB/s (92%); FP8 compute peak 1979 TFLOP/s / achievable 1544 TFLOP/s (78%).

Quant/cast kernels are reported as % of achievable memory bandwidth; DeepGEMM grouped GEMMs as % of achievable FP8 compute.

| kernel | µs | % of achievable roofline |
|---|--:|--:|
| fwd: act_quant_lhs | 253.5 | 91.1 (BW) |
| fwd: weight_quant_forward_rhs | 128.1 | 89.2 (BW) |
| **fwd: deepgemm_grouped_mm** | 779.3 | **80.0 (compute)** |
| bwd: act_quant_lhs(grad_out) | 77.6 | 85.0 (BW) |
| bwd: weight_quant_dgrad_rhs | 126.3 | 90.5 (BW) |
| **bwd: deepgemm_grouped_mm_dgrad** | 778.6 | **80.1 (compute)** |
| bwd: wgrad_quant_lhs(grad_out) [transposed] | 85.1 | 77.5 (BW) |
| bwd: wgrad_quant_rhs(A) [direct] | 275.3 | 83.8 (BW) |
| **bwd: deepgemm_grouped_mm_wgrad** | 2105.4 | **29.6 (compute)** |

### DIFF
--- a/torchao/prototype/blockwise_fp8_training/deepgemm_metadata.py
+++ b/torchao/prototype/blockwise_fp8_training/deepgemm_metadata.py
@@ -25,6 +25,47 @@ class DeepGemmKGroupedQuantMetadata:
         return self.q_offset_by_block.numel()
 
 
+@dataclass(frozen=True)
+class DeepGemmGroupedOffsetPlan:
+    group_end_offsets: torch.Tensor
+    group_sizes: list[int]
+    ks_tensor: torch.Tensor
+    grouped_layout: torch.Tensor
+    original_group_sizes: list[int] | None = None
+    padded_group_start_offsets: torch.Tensor | None = None
+
+    @property
+    def total_tokens(self) -> int:
+        return sum(self.group_sizes)
+
+    def groups_are_block_aligned(self, block_size: int) -> bool:
+        return all(group_size % block_size == 0 for group_size in self.group_sizes)
+
+    def valid_scale_blocks(self, block_size: int) -> int:
+        assert self.groups_are_block_aligned(block_size), (
+            "valid scale blocks require block-aligned group sizes"
+        )
+        return sum(group_size // block_size for group_size in self.group_sizes)
+
+    def k_quant_metadata(
+        self,
+        block_size: int,
+        dim: int,
+    ) -> DeepGemmKGroupedQuantMetadata:
+        return build_deepgemm_k_grouped_quant_metadata(
+            self.group_end_offsets,
+            self.group_sizes,
+            block_size,
+            dim,
+        )
+
+    def m_grouped_layout(self, num_rows: int) -> torch.Tensor:
+        assert self.grouped_layout.numel() == num_rows, (
+            "cached DeepGEMM grouped layout does not match operand rows"
+        )
+        return self.grouped_layout
+
+
 def group_sizes_from_offsets(group_end_offsets: torch.Tensor) -> list[int]:
     assert group_end_offsets is not None and group_end_offsets.dtype == torch.int32, (
         "group_end_offsets must be int32"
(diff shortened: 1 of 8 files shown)


iamzainhuda:
@claude review

andrewor14:
@claude review

torchao/prototype/blockwise_fp8_training/deepgemm_metadata.py
```diff
@@ -0,0 +1,203 @@
...
+    def valid_tokens(self) -> int:
+        return sum(self.group_sizes)
+
+    @property
+    def valid_blocks(self) -> int:
+        return self.q_offset_by_block.numel()
+
+
+@dataclass(frozen=True)
+class DeepGemmGroupedOffsetPlan:
```
andrewor14, line 49: add some docstrings here (and all other new classes)? Explain what each of the fields refer to?

torchao/prototype/blockwise_fp8_training/deepgemm_metadata.py
```diff
@@ -0,0 +1,203 @@
...
+    layout[~valid] = -1
+    return layout.contiguous()
+
+
+def build_deepgemm_k_grouped_quant_metadata(
+    group_end_offsets: torch.Tensor,
+    group_sizes: list[int],
+    block_size: int,
+    dim: int,
+) -> DeepGemmKGroupedQuantMetadata:
```
andrewor14, line 209: need some docstrings on these helper functions too. What are they doing on a high level, what the args are etc

torchao/prototype/moe_training/blockwise_fp8/grouped_mm.py
```diff
@@ -115,6 +159,7 @@ def forward(
             padded_group_end_offsets,
             out_dtype,
             block_size,
+            deepgemm_offset_plan=deepgemm_offset_plan,
```
andrewor14: seems a bit weird to pass this deepgemm specific plan to a general backend plan. Given a specific backend the plan is the same for all grouped_mm calls right? Can we just embed this offset plan in the backend_plan itself so you don't need to pass it to every grouped_mm call?

torchao/prototype/moe_training/blockwise_fp8/grouped_mm_backend.py
```diff
@@ -0,0 +1,338 @@
...
+        block_size: int,
+        dtype: torch.dtype,
+        *,
+        deepgemm_offset_plan: DeepGemmGroupedOffsetPlan | None,
+    ) -> torch.Tensor:
+        raise NotImplementedError
+
+
+class _EmulatedGroupedMMBackendPlan(_GroupedMMBackendPlan):
+    kind = _GroupedMMBackend.EMULATED
```
andrewor14: nit: maybe just `backend` to be more specific?

torchao/prototype/moe_training/blockwise_fp8/grouped_mm_backend.py
```diff
@@ -0,0 +1,338 @@
...
+    kernel_preference: KernelPreference,
+    A: torch.Tensor,
+    out_dtype: torch.dtype,
+    block_size: int,
+    group_end_offsets: torch.Tensor,
+    *,
+    original_group_end_offsets: Optional[torch.Tensor] = None,
+    padded_group_start_offsets: Optional[torch.Tensor] = None,
+    num_rows: Optional[int] = None,
+) -> _GroupedMMBackendSelection:
```
andrewor14: the docstring here should list all the scenarios in which emulated is selected vs deepgemm

andrewor14:
@claude can you verify if the refactor is done correctly such that the original emulated path is unchanged? Trace through the code paths and print them here. Then do the same for the new deepgemm code path

andrewor14 (commented):
Looks great, just mostly comments about docs and arg signatures. Looks like claude pointed out there's potentially a wrong shape. Can you look into that?

Also just wondering if you've tried this on a real model yet to see if we get some reasonable loss, not blocking for this PR but would be great to see some real examples

torchao/prototype/moe_training/blockwise_fp8/grouped_mm.py
iamzainhuda: yeah this makes sense, changed it so its part of that backends plan now

torchao/prototype/moe_training/blockwise_fp8/grouped_mm_backend.py
iamzainhuda: added

torchao/prototype/moe_training/blockwise_fp8/grouped_mm_backend.py
iamzainhuda: changed to make it clearer:
  - _GroupedMMBackend enum -> _GroupedMMBackendKind
  - _GroupedMMBackendPlan -> _GroupedMMBackend
  - _EmulatedGroupedMMBackendPlan -> _EmulatedGroupedMMBackend
  - _DeepGemmGroupedMMBackendPlan -> _DeepGemmGroupedMMBackend

iamzainhuda:
@andrewor14 addressed the comments

andrewor14:
@claude review again?

torchao/prototype/moe_training/blockwise_fp8/grouped_mm_backend.py
```diff
@@ -0,0 +1,322 @@
...
+from torchao.prototype.blockwise_fp8_training.kernels import (
+    BLOCKWISE_1X128_SCALING_TYPE,
+    _scaling_type_value,
+    triton_fp8_blockwise_act_quant_rhs,
+    triton_fp8_blockwise_act_quant_transposed_lhs,
+)
+from torchao.quantization.quantize_.common import KernelPreference
+
+
+class _GroupedMMBackendKind(str, Enum):
```
andrewor14, line 41: nit: "kind" still sounds a bit vague. Backend "name"? or just "backend" and name the other one something else?

andrewor14 (approved):
LGTM, thanks!

PR #4137, opened 2026-03-23
[xpu][test] Enable MoE tests for Intel XPU

kgajdamo:
The PR enables a subset of MoE tests for Intel XPU and is part of the [XPU Enabling Plan for TorchAO](https://github.com/pytorch/ao/issues/3576).

Details:
- The PR doesn’t introduce any new logic to the tests.
- Adds a device parameter to be able to run the tests on the XPU.
- Skips not yet supported scenarios.

PR #3823, opened 2026-02-04
[xpu][float8] device agnostic test base,compile, numerics_integration

arlesniak:
This PR activates a subset of float8 tests for Intel XPU as part of the [XPU Enabling Plan for TorchAO](https://github.com/pytorch/ao/issues/3576).

Details:
- The PR does not add any new test logic.
- It introduces a device parameter to allow running the tests on XPU.
- Unsupported scenarios are skipped.

PR #4474, opened 2026-06-11
Enable pt2e XPU tests in test_x86inductor_fusion

chunhuanMeng:
- Add 35 XPU test cases migrated from pytorch release/2.9
- Add get_default_xpu_quantizer() in torchao/testing/pt2e/utils.py
- Auto-detect XPU device in _generate_ref_quantized_model
- Add @config.patch({"fx_graph_cache": False}) to 4 flaky XPU tests
- Remove --ignore for test_x86inductor_fusion.py in CI script

### DIFF
--- a/.github/scripts/ci_test_xpu.sh
+++ b/.github/scripts/ci_test_xpu.sh
@@ -16,8 +16,7 @@ python -c "import torch; import torchao; print(f'Torch version: {torch.__version
 
 python -m pip install pytest expecttest parameterized accelerate hf_transfer 'modelscope!=1.15.0' transformers tabulate fire
 
-pytest -v -s --ignore=torchao/test/quantization/pt2e/test_x86inductor_fusion.py \
-        torchao/test/quantization/pt2e/ \
+pytest -v -s torchao/test/quantization/pt2e/ \
         torchao/test/quantization/*.py \
         torchao/test/dtypes/ \
         torchao/test/float8/ \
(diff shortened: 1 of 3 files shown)


zxd1997066:
The test is previously ignored because the the cpu part in this file cost too much time and it is not related to xpu

message from Karol:

I've noticed that we started hitting the timeout for `xpu-test` workflow.
The timeout is set to 2 hours (https://github.com/pytorch/ao/blob/main/.github/workflows/xpu_test.yml#L28).
 
On Artur's PR [#3823](https://github.com/pytorch/ao/pull/3823) we had two failed runs due to this problem.
On the other hand, Kinga's PR succeeded ([#4137](https://github.com/pytorch/ao/pull/4137s))
 
I did a small research.
Both runs used the identical Docker image and the identical test files (within what was reached).
There are some problematic (time consuming) parts of the test.
Here is timing comparison of the bottlenecks:
 
 


| Phase                                         | CI machine 1 (bad)        | CI machine 2 (good)     |
| --------------------------------------------- | ---------------------- | -------------------- |
| `test_x86inductor_fusion.py` total            | 56 min                 | 46 min               |
| — of which: `DynamicShapesCppWrapperCpuTests` | 21 min                 | 15 min               |
| `test_int8_tensor.py` compile_True_xpu        | incomplete (canceled)  | 9 min (all 36 tests) |
| Total test time                               | >107 min (interrupted) | 91:17

xiaowangintel:
@chunhuanMeng test_x86inductor_fusion.py is an x86-specific test file. Adding these tests there will introduce a large number of tests that are unrelated to XPU. Please move these tests to a more appropriate test file.

PR #4477, opened 2026-06-12
 Refactor NVFP4Tensor: align the swizzle execution with the configured swizzle setting

xiaowangintel:
## Summary

This applies the same principle introduced in [pytorch/ao#4844](https://github.com/pytorch/ao/pull/4844) ("Refactor MXTensor: align the swizzle execution with the configured swizzle setting", now merged into `main`) to `NVFP4Tensor`.

Previous designs always produced swizzled scales regardless of the config. This silently ignored the configured swizzle setting on non-CUDA backends and re-encoded backend-specific behavior instead of relying on PyTorch's own contract: `F.scaled_mm` already takes the scale layout as explicit `swizzle_a` / `swizzle_b` arguments.

## Config changes

Per review feedback on #4844 (swizzle knobs on user-facing configs should be a `SwizzleType` enum, not a bare `bool`; tensor-internal attributes stay `bool`), renamed:

- `NVFP4DynamicActivationNVFP4WeightConfig.is_swizzled: bool = True` → `swizzled_type: SwizzleType = SwizzleType.SWIZZLE_32_4_4`
- `NVFP4WeightOnlyConfig.is_swizzled: bool = True` → `swizzled_type: SwizzleType = SwizzleType.SWIZZLE_32_4_4`

**What each backend accepts (matches PyTorch's own support matrix):**

| | CUDA (SM100+) | ROCm | XPU |
|---|---|---|---|
| `swizzle_a` / `swizzle_b` | `SWIZZLE_32_4_4` | `NO_SWIZZLE` | `NO_SWIZZLE` |

### Usage

```python
# default — blocked/swizzled scales (CUDA)
config = NVFP4DynamicActivationNVFP4WeightConfig(
    swizzled_type=SwizzleType.SWIZZLE_32_4_4,
)

# no swizzle (ROCm / XPU)
config = NVFP4DynamicActivationNVFP4WeightConfig(
    swizzled_type=SwizzleType.NO_SWIZZLE,
)
```

### DIFF
--- a/torchao/prototype/mx_formats/inference_workflow.py
+++ b/torchao/prototype/mx_formats/inference_workflow.py
@@ -221,6 +221,7 @@ class NVFP4DynamicActivationNVFP4WeightConfig(AOBaseConfig):
     use_triton_kernel: bool = True
     use_dynamic_per_tensor_scale: bool = True
     step: Optional["QuantizationStep"] = None
+    is_swizzled: bool = True
 
     def __post_init__(self):
         if isinstance(self.step, str):
@@ -245,6 +246,7 @@ def _nvfp4_inference_linear_transform(
     - None (default): Original dynamic quantization behavior
     """
     weight = getattr(module, parameter_name)
+    is_swizzled_scales = config.is_swizzled
     if weight.shape[-2] % 16 != 0 or weight.shape[-1] % 16 != 0:
         raise RuntimeError(
             f"NVFP4 only supports weight shape with last 2 dims divisible by 16, got {weight.shape}"
@@ -270,14 +272,14 @@ def _nvfp4_inference_linear_transform(
         act_quant_kwargs = QuantizeTensorToNVFP4Kwargs(
             use_dynamic_per_tensor_scale=False,
             use_triton_kernel=config.use_triton_kernel,
-            is_swizzled_scales=True,
+            is_swizzled_scales=is_swizzled_scales,
         )
 
         quantized_weight = NVFP4Tensor.to_nvfp4(
             weight,
             per_tensor_scale=weight_per_tensor_scale,
             act_per_tensor_scale=act_per_tensor_scale.detach(),
-            is_swizzled_scales=True,
+            is_swizzled_scales=is_swizzled_scales,
             use_triton_kernel=False,  # Always use traditional construction for weights
             act_quant_kwargs=act_quant_kwargs,
         )
@@ -298,8 +300,8 @@ def _nvfp4_inference_linear_transform(
 
     elif step is None:
         # Dynamic quantization
-        assert is_sm_at_least_100(), (
-            "NVFP4 DYNAMIC mode is only supported on sm100+ machines"
+        assert is_sm_at_least_100() or torch.xpu.is_available(), (
+            "NVFP4 DYNAMIC mode is only supported on sm100+ machines or XPU"
         )
 
         weight = getattr(module, parameter_name)
@@ -321,13 +323,13 @@ def _nvfp4_inference_linear_transform(
         act_quant_kwargs = QuantizeTensorToNVFP4Kwargs(
             use_dynamic_per_tensor_scale=config.use_dynamic_per_tensor_scale,
             use_triton_kernel=config.use_triton_kernel,
-            is_swizzled_scales=True,
+            is_swizzled_scales=is_swizzled_scales,
         )
 
         quantized_weight = NVFP4Tensor.to_nvfp4(
             weight,
             per_tensor_scale=per_tensor_scale,
-            is_swizzled_scales=True,
+            is_swizzled_scales=is_swizzled_scales,
             use_triton_kernel=False,  # Always use traditional construction for weights
             act_quant_kwargs=act_quant_kwargs,
         )
@@ -368,6 +370,7 @@ class NVFP4WeightOnlyConfig(AOBaseConfig):
     """
 
     use_dynamic_per_tensor_scale: bool = True
+    is_swizzled: bool = True
 
 
 @register_quantize_module_handler(NVFP4WeightOnlyConfig)
@@ -383,6 +386,8 @@ def _nvfp4_weight_only_linear_transform(
             f"NVFP4 only supports weight shape with last 2 dims divisible by 16, got {weight.shape}"
         )
 
+    is_swizzled_scales = config.is_swizzled
+
     per_tensor_scale = None
     if config.use_dynamic_per_tensor_scale:
         tensor_amax = torch.max(torch.abs(weight))
@@ -391,7 +396,7 @@ def _nvfp4_weight_only_linear_transform(
     quantized_weight = NVFP4Tensor.to_nvfp4(
         weight,
         per_tensor_scale=per_tensor_scale,
-        is_swizzled_scales=True,
+        is_swizzled_scales=is_swizzled_scales,
         act_quant_kwargs=None,
     )
     # Set triton preference after construction
--- a/torchao/prototype/mx_formats/inference_workflow.py
+++ b/torchao/prototype/mx_formats/inference_workflow.py
@@ -250,13 +250,13 @@ def _nvfp4_inference_linear_transform(
     - None (default): Original dynamic quantization behavior
     """
     weight = getattr(module, parameter_name)
-    is_swizzled = config.swizzled_type == SwizzleType.SWIZZLE_32_4_4
     if weight.shape[-2] % 16 != 0 or weight.shape[-1] % 16 != 0:
         raise RuntimeError(
             f"NVFP4 only supports weight shape with last 2 dims divisible by 16, got {weight.shape}"
         )
 
     step = config.step
+    is_swizzled = config.swizzled_type == SwizzleType.SWIZZLE_32_4_4
     if step == QuantizationStep.PREPARE or step == "prepare":
         return NVFP4ObservedLinear.from_float(module)
 
--- a/torchao/prototype/mx_formats/inference_workflow.py
+++ b/torchao/prototype/mx_formats/inference_workflow.py
@@ -225,7 +225,7 @@ class NVFP4DynamicActivationNVFP4WeightConfig(AOBaseConfig):
     use_triton_kernel: bool = True
     use_dynamic_per_tensor_scale: bool = True
     step: Optional["QuantizationStep"] = None
-    is_swizzled: bool = True
+    swizzled_type: SwizzleType = SwizzleType.SWIZZLE_32_4_4
 
     def __post_init__(self):
         if isinstance(self.step, str):
@@ -250,7 +250,7 @@ def _nvfp4_inference_linear_transform(
     - None (default): Original dynamic quantization behavior
     """
     weight = getattr(module, parameter_name)
-    is_swizzled_scales = config.is_swizzled
+    is_swizzled_scales = config.swizzled_type == SwizzleType.SWIZZLE_32_4_4
     if weight.shape[-2] % 16 != 0 or weight.shape[-1] % 16 != 0:
         raise RuntimeError(
             f"NVFP4 only supports weight shape with last 2 dims divisible by 16, got {weight.shape}"
@@ -374,7 +374,8 @@ class NVFP4WeightOnlyConfig(AOBaseConfig):
     """
 
     use_dynamic_per_tensor_scale: bool = True
-    is_swizzled: bool = True
+    # How to store block scales.
+    swizzled_type: SwizzleType = SwizzleType.SWIZZLE_32_4_4
 
 
 @register_quantize_module_handler(NVFP4WeightOnlyConfig)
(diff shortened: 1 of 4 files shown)


torchao/prototype/mx_formats/inference_workflow.py
```diff
@@ -249,6 +250,7 @@ def _nvfp4_inference_linear_transform(
     - None (default): Original dynamic quantization behavior
     """
     weight = getattr(module, parameter_name)
+    is_swizzled_scales = config.swizzled_type == SwizzleType.SWIZZLE_32_4_4
```
liangan1:
```suggestion
    is_swizzled = config.swizzled_type == SwizzleType.SWIZZLE_32_4_4
```

torchao/prototype/mx_formats/inference_workflow.py
```diff
@@ -274,14 +276,14 @@ def _nvfp4_inference_linear_transform(
         act_quant_kwargs = QuantizeTensorToNVFP4Kwargs(
             use_dynamic_per_tensor_scale=False,
             use_triton_kernel=config.use_triton_kernel,
-            is_swizzled_scales=True,
+            is_swizzled_scales=is_swizzled_scales,
```
liangan1:
```suggestion
            is_swizzled_scales=is_swizzled,
```

torchao/prototype/mx_formats/inference_workflow.py
```diff
@@ -274,14 +276,14 @@ def _nvfp4_inference_linear_transform(
...
         )
 
         quantized_weight = NVFP4Tensor.to_nvfp4(
             weight,
             per_tensor_scale=weight_per_tensor_scale,
             act_per_tensor_scale=act_per_tensor_scale.detach(),
-            is_swizzled_scales=True,
+            is_swizzled_scales=is_swizzled_scales,
```
liangan1:
```suggestion
            is_swizzled_scales=is_swizzled,
```

torchao/prototype/mx_formats/inference_workflow.py
```diff
@@ -325,13 +327,13 @@ def _nvfp4_inference_linear_transform(
         act_quant_kwargs = QuantizeTensorToNVFP4Kwargs(
             use_dynamic_per_tensor_scale=config.use_dynamic_per_tensor_scale,
             use_triton_kernel=config.use_triton_kernel,
-            is_swizzled_scales=True,
+            is_swizzled_scales=is_swizzled_scales,
```
liangan1:
```suggestion
            is_swizzled_scales=is_swizzled,
```

torchao/prototype/mx_formats/inference_workflow.py
```diff
@@ -325,13 +327,13 @@ def _nvfp4_inference_linear_transform(
...
         )
 
         quantized_weight = NVFP4Tensor.to_nvfp4(
             weight,
             per_tensor_scale=per_tensor_scale,
-            is_swizzled_scales=True,
+            is_swizzled_scales=is_swizzled_scales,
```
liangan1:
```suggestion
            is_swizzled_scales=is_swizzled,
```

torchao/prototype/mx_formats/inference_workflow.py
```diff
@@ -387,6 +391,8 @@ def _nvfp4_weight_only_linear_transform(
             f"NVFP4 only supports weight shape with last 2 dims divisible by 16, got {weight.shape}"
         )
 
+    is_swizzled_scales = config.swizzled_type == SwizzleType.SWIZZLE_32_4_4
```
liangan1:
```suggestion
    is_swizzled = config.swizzled_type == SwizzleType.SWIZZLE_32_4_4
```

torchao/prototype/mx_formats/inference_workflow.py
```diff
@@ -395,7 +401,7 @@ def _nvfp4_weight_only_linear_transform(
     quantized_weight = NVFP4Tensor.to_nvfp4(
         weight,
         per_tensor_scale=per_tensor_scale,
-        is_swizzled_scales=True,
+        is_swizzled_scales=is_swizzled_scales,
```
liangan1:
```suggestion
        is_swizzled_scales=is_swizzled,
```

torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -539,7 +551,6 @@ def _addmm_nvfp4_dispatch(
     should_add_bias_separately = (
         scale_result is not None or a.orig_dtype == torch.float32
     ) and (bias is not None)
-    # should_add_bias_separately = bias is not None
```
liangan1, line 542: Restore this change.

torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -498,24 +500,34 @@ def _addmm_nvfp4_dispatch(
...
     if b.is_swizzled_scales:
         b_scale_blocked = b.scale.t()  # Already swizzled
     else:
-        b_scale = b.scale.t().view(N, K // b.block_size)
-        b_scale_blocked = to_blocked(b_scale)
+        b_scale_blocked = b.scale.t().view(N, K // b.block_size)
+
+    swizzle = (
+        SwizzleType.SWIZZLE_32_4_4 if a.is_swizzled_scales else SwizzleType.NO_SWIZZLE
+    )
```
liangan1:
```suggestion
    swizzle_type = (
        SwizzleType.SWIZZLE_32_4_4 if a.is_swizzled_scales else SwizzleType.NO_SWIZZLE
    )
```

torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -558,13 +569,17 @@ def _addmm_nvfp4_dispatch(
...
         a.qdata.view(torch.float4_e2m1fn_x2),
         b.qdata.view(torch.float4_e2m1fn_x2),
-        a_scale_blocked.view(torch.float8_e4m3fn),
-        b_scale_blocked.view(torch.float8_e4m3fn),
+        scale_a=a_scale_blocked.view(torch.float8_e4m3fn),
+        scale_recipe_a=ScalingType.BlockWise1x16,
+        scale_b=b_scale_blocked.view(torch.float8_e4m3fn),
+        scale_recipe_b=ScalingType.BlockWise1x16,
+        swizzle_a=swizzle,
+        swizzle_b=swizzle,
```
liangan1:
```suggestion
        swizzle_a=swizzle_type,
        swizzle_b=swizzle_type,
```

xiaowangintel: Done

torchao/prototype/mx_formats/nvfp4_tensor.py
xiaowangintel: Done

torchao/prototype/mx_formats/inference_workflow.py
xiaowangintel: Done

torchao/prototype/mx_formats/inference_workflow.py
xiaowangintel: Done

torchao/prototype/mx_formats/inference_workflow.py
xiaowangintel: Done

torchao/prototype/mx_formats/inference_workflow.py
xiaowangintel: Done

torchao/prototype/mx_formats/inference_workflow.py
xiaowangintel: Done

torchao/prototype/mx_formats/inference_workflow.py
xiaowangintel: Done

torchao/prototype/mx_formats/nvfp4_tensor.py
xiaowangintel: Done

xiaowangintel:
@claude review this pr.

xiaowangintel:
@claude review this pr.

torchao/prototype/mx_formats/nvfp4_tensor.py
```diff
@@ -535,7 +539,7 @@ def _addmm_nvfp4_dispatch(
     # since bias is not quantized
     #
     # (2) RuntimeError: Bias is not supported when out_dtype is set to Float32
-    # This is not supported by _scaled_mm
+    # This is not supported by F.scaled_mm
```
liangan1, line 542: Have you double check this?

xiaowangintel: Yes, the details refer to the corresponding implementation in https://github.com/pytorch/pytorch/blob/300f0c8298dc4610f1e13f6b7f8e0366fdd04204/aten/src/ATen/native/cuda/ScaledBlas.cpp#L1556-L1558

liangan1:
@claude review again.

liangan1 (approved):

torchao/prototype/mx_formats/inference_workflow.py
```diff
@@ -302,8 +304,8 @@ def _nvfp4_inference_linear_transform(
 
     elif step is None:
         # Dynamic quantization
-        assert is_sm_at_least_100(), (
-            "NVFP4 DYNAMIC mode is only supported on sm100+ machines"
+        assert is_sm_at_least_100() or torch.xpu.is_available(), (
```
vkuzo, line 307: is nvfp4 supported on every single XPU chip? if not, this should be gated to versions that do support it

vkuzo:
@claude review

torchao/prototype/mx_formats/inference_workflow.py
liangan1: Thanks for your comments. @vkuzo. XPU use different philosophy to CUDA. oneDNN is the compute backend of XPU and provides broad platform support for a specific data type, even if this data type is not natively supported by the HW. 

Let us take the nvfp4 as example, it can be used by any Intel XPU platform. When it is not natively supported by a specific HW, the oneDNN kernel will do the  C= dequantize(A) * dequantize(B). This fallback implementation allows users to run an nvfp4 model on any XPU platform. For the compute bound case(e.g., prefill in LLM inference), this fallback may not be as good as the natively supported HW, but for the memory bound case(e.g., decoding in LLM inference), this fallback implementation of nvfp4 can still get similar perf with natively support HW.

liangan1:
@claude  review.

torchao/prototype/mx_formats/inference_workflow.py
vkuzo: @liangan1 , that makes sense, unfortunately this is in conflict with the design philosophy of torchao on the same problem, which is:
1. user specifies a quant config
2. the software either executes it or throws an error.  There is never any silent change from what user requested to what hardware executes.

We believe that this is the simplest way to go, and it prevents silent gotchas when user expects recipe X but the hardware silently runs recipe Y.

Can we follow ^ in torchao for XPU? I'd actually recommend changing the pytorch core kernels for scaled_mm on XPU to do the same.  

Note that we can still express "dequantize weight and run bf16 gemm", that is usually called weight-only quant and we have support for this in other torchao workflows.  I'd recommend exposing this functionality, if you want it on XPU, through the weight-only path.

We also have emulation mode for debugging, which is equivalent to what you said.  It's just not exposed as a production path.

vkuzo: I'd also recommend ^ for all other quant recipes that XPU supports

liangan1: Thanks for you comments. @vkuzo.  Given a quantization workflow, let us break it into the recipe from AO, the operator from PyTorch and the kernel from the Library. 

1. The recipe level in torchAO is decided by the quantization config. e.g.,
```python
 # no swizzle (ROCm / XPU)
config = NVFP4DynamicActivationNVFP4WeightConfig(
    swizzled_type=SwizzleType.NO_SWIZZLE,
)
```
This config means that the Linear module will use nvfp4 data type and no swizzle scales layout for both activation and weights(A4W4). 

2. The operator level in PyTorch decides the semantics of the compute, dtype and layout information of arguments. e.g., the F.scaled_mm will do the NO_SWILLZLE A4W4 computation from the quantization config.
 
3. The kernel is the implementation of the operator on the given hardware. The data type is natively supported by the HW just means that a specific intrinsic is provided by this platform to speed up the compute of this data type, but the semantics of operator are not impacted. For a NO_SWIZZLE A4W4 kernel, the differences are mainly in the intrinsic level: when nvfp4 natively supported, we only use one intrinsic to do this computation, but multiple intrinsics are needed when nvfp4 is not natively supported by HW platform.

So, we don't silently change the user-expected behavior with quantization config in AO. The kernel provide the mathematically equivalent implementation whatever this data type is natively supported by this HW platform.

PR #4480, opened 2026-06-12
[PT2E][X86] Fall back to plain linear annotation for reused nn.Linear with fusable post op

### LINKED ISSUE #4478: X86InductorQuantizer fusion annotation crashes when one nn.Linear has multiple call sites (unrolled loops, reused modules)
## Bug description

`X86InductorQuantizer`'s fused-pattern annotation passes raise a `ValueError`
whenever a single `nn.Linear` module is invoked more than once in the exported
graph and any call site is followed by a fusable op (`add` or `relu`).

`get_source_partitions` groups all call sites of one module into a single
`SourcePartition`, so the partition has one output node per call site. The
fusion passes then reject it:

- linear + add (`_annotate_linear_binary_unary` → `_get_output_nodes_of_partitions`):
  `ValueError: Input partition has more than one output node`
- linear + relu (`_annotate_linear_unary`):
  `ValueError: Linear partition cannot have more than one output node`

Plain (non-fused) linear annotation handles multiple call sites correctly, i.e.
the same model quantizes fine if the add/relu is removed, so the
single-call-site assumption is specific to the fusion matching.

Module reuse like this is common: unrolled sampling loops of diffusion /
flow-matching models (our case: `torch.export` unrolls a 10-step denoising
loop, so every linear in the denoiser has 10 call sites). Related but distinct from
pytorch/pytorch#142035 (tied weights between two different modules); this
reproduces with a single module and no weight tying on current versions.

## Minimal reproduction

```python
import torch
from torchao.quantization.pt2e.quantize_pt2e import prepare_pt2e
from torchao.quantization.pt2e.quantizer.x86_inductor_quantizer import (
    X86InductorQuantizer,
    get_default_x86_inductor_quantization_config,
)

class SharedLinearLoop(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.projection = torch.nn.Linear(8, 8)

    def forward(self, x):
        for _ in range(2):
            x = x + self.projection(x)
        return x

example = (torch.randn(2, 8),)
exported = torch.export.export(SharedLinearLoop().eval(), example).module()
quantizer = X86InductorQuantizer()
quantizer.set_global(get_default_x86_inductor_quantization_config())
prepare_pt2e(exported, quantizer)
# ValueError: Input partition has more than one output node
```

Replacing `x = x + self.projection(x)` with `x = torch.relu(self.projection(x))`
fails in `_annotate_linear_unary` instead; `x = self.projection(x)` alone
(no fusable neighbor) quantizes successfully.

## Expected behavior

Either fuse per call site (each call site of the shared linear forms its own
linear+add instance), or skip fusion annotation for multi-call-site partitions
and fall back to plain linear annotation — but do not raise and abort
`prepare_pt2e` for the whole model.

## Versions

- torch 2.12.0+cu130
- torchao 0.17.0+cu130
- Python 3.14.4, Linux

waynehacking8:
Fixes #4478

## Problem

`prepare_pt2e` with `X86InductorQuantizer` raises a `ValueError` whenever a single `nn.Linear` module has multiple call sites in the exported graph (e.g. it is called inside an unrolled loop — common for unrolled sampling loops of diffusion / flow-matching models) and any call site is followed by a fusable `add`/`relu`:

```
ValueError: Input partition has more than one output node          # linear + add
ValueError: Linear partition cannot have more than one output node # linear + relu
```

Root cause: under non-strict export (the current `torch.export.export` default), `get_source_partitions` groups all call sites of one module instance into a single `SourcePartition`, so the partition has one output node per call site. The linear fusion annotation passes (`_annotate_linear_unary`, `_annotate_linear_binary_unary` via `_get_output_nodes_of_partitions`) assume exactly one output node and raise, aborting quantization of the whole model. (Under `strict=True` each call site gets its own partition, which is why existing tests never hit this.)

## Fix

- `_annotate_linear_unary` / `_annotate_linear_binary_unary`: skip fused partition candidates where any partition has more than one output node (same `continue` style as the other pattern checks in this file) instead of raising.
- `_annotate_linear`: annotate each linear call site (output node) of a partition instead of raising on multi-output partitions, so the skipped call sites still get plain (non-fused) linear annotation.

Behavior change: a reused linear with a fusable neighbor is now quantized without fusion (input q/dq per call site, per-channel weight dq per call site) instead of crashing `prepare_pt2e`. Fusion behavior for single-call-site modules is unchanged.

Note: the conv fusion paths share the same single-call-site assumption and would crash the same way for a reused conv; this PR is scoped to the linear paths reported in the issue, and I'm happy to extend to conv in a follow-up if desired.

## Before / after

Minimal repro from the issue (`x = x + self.projection(x)` twice with one shared `nn.Linear`):

- Before: `ValueError: Input partition has more than one output node` (and the relu variant fails with `Linear partition cannot have more than one output node`)
- After: `prepare_pt2e` / `convert_pt2e` succeed; converted graph contains 2 quantized linear call sites (2 input q/dq pairs, 2 per-channel weight dq)

## Test

- Added `TestHelperModules.SharedLinearModule` and `test_linear_reused_with_fusable_post_op` in `test/quantization/pt2e/test_x86inductor_quantizer.py`, covering both `add` and `relu` post ops. The test checks node occurrence and annotation statistics (both call sites annotated as plain quantized linears, post ops unannotated).
- `_test_quantizer` gained a `strict` parameter (default `True`, all existing callers unchanged); the new test uses `strict=False` since only non-strict export produces multi-output partitions.
- The new test fails with the exact `ValueError`s from the issue without the source change and passes with it.
- Full `test/quantization/pt2e/test_x86inductor_quantizer.py`: 51 passed (CPU, torch 2.12).
- `ruff check` / `ruff format` (0.11.6, per pre-commit) clean.

> **Note on export mode:** this targets the **default (non-strict)** `torch.export.export` path. PT2E docs never pass `strict=` (all use `torch.export.export(m, example_inputs).module()`), and recent PyTorch defaults export to non-strict — so the documented happy path for any model that reuses an `nn.Linear` in a loop hits this crash. The prior tests only avoided it because `_test_quantizer` hard-codes `strict=True`.

---

This PR was written with the assistance of Claude (AI); I have reviewed and validated all changes.

 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/test/quantization/pt2e/test_x86inductor_quantizer.py
+++ b/test/quantization/pt2e/test_x86inductor_quantizer.py
@@ -470,6 +470,24 @@ class TestHelperModules:
                 tmp = self.linear(x)
                 return tmp + self.linear2(tmp)
 
+    class SharedLinearModule(torch.nn.Module):
+        """One Linear module with multiple call sites (e.g. an unrolled loop),
+        each followed by a fusable post op (add or relu).
+        """
+
+        def __init__(self, post_op: str = "add") -> None:
+            super().__init__()
+            self.linear = torch.nn.Linear(in_features=16, out_features=16)
+            self.post_op = post_op
+
+        def forward(self, x):
+            for _ in range(2):
+                if self.post_op == "add":
+                    x = x + self.linear(x)
+                else:
+                    x = torch.relu(self.linear(x))
+            return x
+
     class Conv2dAddModule2(torch.nn.Module):
         def __init__(
             self,
@@ -668,12 +686,13 @@ class X86InductorQuantTestCase(QuantizationTestCase):
         is_qat=False,
         debug=False,
         lower=False,
+        strict=True,
     ):
         m_eager = model.train() if is_qat else model.eval()
 
         # program capture
         m = copy.deepcopy(m_eager)
-        m = torch.export.export(m, example_inputs, strict=True).module()
+        m = torch.export.export(m, example_inputs, strict=strict).module()
 
         # QAT Model failed to deepcopy
         export_model = m if is_qat else copy.deepcopy(m)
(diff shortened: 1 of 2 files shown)


Xia-Weiwen:
@claude review this. And also check: is `strict=True` of `torch.export.export` required for PT2E quantization in the documents?

Xia-Weiwen:
@waynehacking8 Thanks for the fix. Can you please take a look at the comments from claude? Seems the test case needs improvement.

waynehacking8:
Thanks @Xia-Weiwen, and thanks @claude for the careful review — both points are addressed in 0b35d4d.

**Test coverage fix (claude's point #1).** You're right that the `relu` variant used functional `torch.relu`, which never forms a `[Linear, nn.ReLU]` partition, so it only hit the plain `_annotate_linear` fallback and left the new `_annotate_linear_unary` guard untested. `SharedLinearModule` now uses a reused `nn.ReLU` **module**, so a `[Linear, ReLU]` sequential partition forms and the guard at `_annotate_linear_unary` is exercised.

I verified the coverage is real, not just that the test still passes: with the `_annotate_linear_unary` multi-output guard temporarily removed, the `relu` case fails with `ValueError: Input partition has more than one output node`; with the guard in place it passes. The `node_occurrence` / annotation-stat expectations are unchanged (the reused linear still falls back to plain annotation, relu stays unannotated). I also tightened the `SharedLinearModule` docstring to make the module-vs-functional distinction explicit (claude's point #3).

**`strict=True` for PT2E (your question + claude's answer).** Confirmed: `strict=True` is **not** required and isn't used anywhere in the PT2E docs — every example (`pt2e_quant_x86_inductor.rst:266`, `pt2e_quant_ptq.rst:67`, `pt2e_quant_qat.rst:38`, `pt2e_quant_openvino_inductor.rst:111`) calls `torch.export.export(m, example_inputs).module()` with the default. Since `torch.export.export` now defaults to **non-strict**, the documented happy path is exactly the code path this bug hits — the existing tests only avoided it because `_test_quantizer` hard-codes `strict=True`. I'll note in the PR description that the fix targets the default (non-strict) export behavior, not an opt-in edge case.

**Conv paths (claude's point #2).** Agreed the conv annotation paths share the same single-output assumption (`_annotate_conv2d` raises at the same class of error for a reused conv). I scoped this PR to linear deliberately; happy to open a tracking issue and follow up with the conv fix in a separate PR.

waynehacking8:
Opened the conv follow-up as #4482 (stacked on this branch, since its test needs the `strict=False` helper param here). It mirrors this fix across `_annotate_conv2d` + the three conv fusion paths, with the same coverage-proof approach. Happy to fold it into this PR instead if you'd prefer a single "reused nn.Linear/nn.Conv2d" change — just let me know.

Xia-Weiwen:
@claude review again

Xia-Weiwen (approved):
Thanks for the fix

waynehacking8:
CI note: the failing `test_qconv2d_maxpool2d_linear_dynamic_cpu_cpp_wrapper` (`'quantized::max_pool2d' not found`) **also fails on clean `main` (c42673d)** with no changes from this PR — i.e. it's pre-existing and unrelated. This PR only touches `_annotate_linear` / the linear fusion guards; the non-cpp-wrapper variant of that same test passes locally. Verified by running it on a clean main worktree.

Issue #4116, opened 2026-03-19
Linear+BatchNorm fusion produces incorrect model in most cases


StrycekSimon:
In `prepare_pt2e`, there is a call to `_fuse_linear_bn_` (see [quantize_pt2e.py](https://github.com/pytorch/ao/blob/1d75a0779f2e0095a08b3753f4a4f1b9666ccfe0/torchao/quantization/pt2e/quantize_pt2e.py#L106)) that unconditionally fuses BatchNorm stats to Linear weights tensor. The problem is that this fusion silently produces invalid model in some cases.

**More context**
The fusion is correct only when both layers work with the same dimension. Linear layer works strictly with only the last dimension of it's input. BatchNorm on the other hand works in channels dimension (the second one in most cases) except for one case when using BatchNorm1d and input shape `(N, C)` (see [official PyTorch documentation](https://docs.pytorch.org/docs/stable/generated/torch.nn.BatchNorm1d.html) for more info). Only in this one case both BatchNorm and Linear works on the same dim and are correctly fusable. When we use incompatible Linear+BatchNorm that accidentally has matching last dim size with number of channels, the fusion silently produces a model with incorrect Linear weights tensor.

**Example**
In typical scenario, the last dim won't match the number of channels and will fail:
```python
inp = torch.randn((2, 2, 3)) # (N, C, L)
lin = nn.Linear(in_features=3, out_features=5) # Weight of shape (5, 3)
bn = nn.BatchNorm2d(num_features=2) # Running var shape (2, )

# With the following fusing equation:
# linear_w_fused = linear_w * (gamma / sqrt(var + eps))
# And the following shapes:
# linear_w.shape: (5, 3); var.shape: (2, )
# We cannot perform the multiplication.
```
but when `inp = torch.randn((2, 5, 3))`, the quantization passes and produces incorrect Linear weights.

jcaip:
cc @SS-JIA @andrewor14 @jerryzh168

PR #4484, opened 2026-06-13
[pt2e] Fuse and fold linear + bn in QAT like conv + bn

### LINKED ISSUE #4463: Linear+BatchNorm1d not folded during QAT, unlike Conv+BN
### Description

I'm using the PT2E QAT flow (`prepare_qat_pt2e` + `convert_pt2e`), training from scratch and then lowering the converted graph to a custom backend. Conv+BN gets folded into the conv at convert time, which works great. Linear+BatchNorm1d doesn't get folded at all though, so a `batch_norm` node survives `convert_pt2e` and my backend has to special-case it.

As far as I can tell the QAT fuse/fold code in `qat_utils.py` is conv-only: `_fuse_conv_bn_qat` runs during prepare and `_fold_conv_bn_qat` runs during convert, but there's no linear equivalent in either step. The PTQ path did get linear+BN folding recently (#3917, plus the 2D correctness fix in #4242), so it looks like the QAT side just hasn't been done yet rather than being intentionally skipped. Wanted to check before putting up a PR.

Small repro, conv folds but linear doesn't:

```python
import torch
from torch.export import export
from torchao.quantization.pt2e.quantize_pt2e import prepare_qat_pt2e, convert_pt2e
from torchao.testing.pt2e._xnnpack_quantizer import (
    XNNPACKQuantizer, get_symmetric_quantization_config,
)

def count_bn(gm):
    return sum(n.op == "call_function" and "batch_norm" in str(n.target)
               for n in gm.graph.nodes)

class ConvBN(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.c = torch.nn.Conv2d(3, 4, 3, padding=1, bias=False)
        self.bn = torch.nn.BatchNorm2d(4)
    def forward(self, x): return torch.relu(self.bn(self.c(x)))

class LinearBN(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.fc = torch.nn.Linear(8, 16, bias=False)
        self.bn = torch.nn.BatchNorm1d(16)
    def forward(self, x): return torch.relu(self.bn(self.fc(x)))

q = XNNPACKQuantizer().set_global(
    get_symmetric_quantization_config(is_qat=True, is_per_channel=True)
)
for name, m, ex in [("Conv2d+BN2d", ConvBN(), torch.randn(4, 3, 8, 8)),
                    ("Linear+BN1d", LinearBN(), torch.randn(4, 8))]:
    gm = export(m, (ex,)).module()
    prep = prepare_qat_pt2e(gm, q)
    for _ in range(4):
        prep(torch.randn_like(ex))
    conv = convert_pt2e(prep)
    print(f"{name}: surviving batch_norm nodes = {count_bn(conv)}")
```

`Conv2d+BN2d` prints 0, `Linear+BN1d` prints 1.

Couple of things I want to flag so we're on the same page:

- I need the QAT version specifically, not the PTQ `_fuse_linear_bn_`. That one collapses BN into the linear weights up front, before training. Since I'm training from scratch, BN has to stay live and keep adapting during training and only fold at convert, which is what the conv QAT path already does.
- I know from #4116 / #4242 that linear+BN is only safe to fold for 2D `(N, C)` activations, since Linear operates on the last dim while BatchNorm1d normalizes dim 1, and those only line up at 2D. I'd carry the same guard over and skip 3D+ inputs with a warning instead of folding them.

If this seems reasonable I'm happy to send a PR that mirrors the conv path for linear and reuses the existing `fold_bn_weights_into_linear_node` for the convert-time fold. And if there's a reason it was left out that I'm not seeing, let me know.

cc @jerryzh168 @andrewor14

discobot:
Fixes #4463.

`prepare_qat_pt2e`/`convert_pt2e` only register conv variants in the QAT
fuse/fold passes, so a `batch_norm` node survives convert for
Linear+BatchNorm1d. This mirrors the conv path for `F.linear` in
`qat_utils.py`, as discussed in the issue: the pattern helpers are already
parameterized by `conv_fn`, the approximate-fusion math yields the right
broadcast shapes for a 2-D weight (matching `nniqat.LinearBn1d`), and the
convert step reuses the existing `fold_bn_weights_into_linear_node` from the
PTQ work. The 2-D-input-only guard from #4116/#4242 is enforced as a match
filter at prepare time (skip + warn), which is sufficient — an unfused linear
never produces the quantized QAT pattern at convert.

One thing the issue didn't surface: the `qat_utils.py` changes alone aren't
enough for the repro to fold. Since QAT annotates **before** fusing (PTQ fuses
before annotating), the quantizer must place the output qspec at the BN/relu
like `conv_bn`/`conv_bn_relu` do; the plain `linear` annotator instead inserts
an output fake-quantize between the linear and the rescaling `div` inside the
fused subgraph, and the convert-time pattern never matches. So this also adds
`linear_bn`/`linear_bn_relu` QAT-only annotators to the testing
XNNPACKQuantizer, mirroring `_do_annotate_conv_bn`. Backend quantizers that
want the fold need the same annotation shape.

Also extended the convert-time `num_batches_tracked` `add_` cleanup to match
BatchNorm1d in addition to BatchNorm2d — this is needed for linear+bn1d and
also removes the dangling `add_` previously left behind for conv1d+bn1d folds.

Added `TestQuantizePT2EQAT_LinearBn` covering: prepared/converted graph
structure and exact numerics parity with FX QAT (`nniqat.LinearBn1d`) for
bias/no-bias and per-channel/per-tensor, relu variant, BN erasure at convert,
and the 3-D input skip. Full `test_quantize_pt2e_qat.py` passes on CPU.

### DIFF
--- a/test/quantization/pt2e/test_quantize_pt2e_qat.py
+++ b/test/quantization/pt2e/test_quantize_pt2e_qat.py
@@ -204,6 +204,7 @@ class PT2EQATTestCase(QuantizationTestCase):
         has_bias: bool = True,
         is_cuda: bool = False,
         expected_conv_literal_args: Optional[tuple[Any, ...]] = None,
+        is_linear: bool = False,
         # TODO: set this to true by default
         verify_convert: bool = False,
     ):
@@ -215,6 +216,7 @@ class PT2EQATTestCase(QuantizationTestCase):
             has_bias=has_bias,
             is_cuda=is_cuda,
             expected_conv_literal_args=expected_conv_literal_args,
+            is_linear=is_linear,
             verify_convert=verify_convert,
         )
         self._verify_symmetric_xnnpack_qat_graph_helper(
@@ -225,6 +227,7 @@ class PT2EQATTestCase(QuantizationTestCase):
             has_bias=has_bias,
             is_cuda=is_cuda,
             expected_conv_literal_args=expected_conv_literal_args,
+            is_linear=is_linear,
             verify_convert=verify_convert,
         )
 
@@ -237,6 +240,7 @@ class PT2EQATTestCase(QuantizationTestCase):
         has_bias: bool = True,
         is_cuda: bool = False,
         expected_conv_literal_args: Optional[tuple[Any, ...]] = None,
+        is_linear: bool = False,
         verify_convert: bool = False,
     ):
         """
@@ -290,7 +294,10 @@ class PT2EQATTestCase(QuantizationTestCase):
         (conv_node, scale_factor_reshape_node) = div_scale_factor_node.args
         conv_op = conv_node.target
         self.assertEqual(div_scale_factor_node.target, torch.ops.aten.div.Tensor)
-        self.assertTrue(_is_conv_node(conv_node))
+        if is_linear:
+            self.assertEqual(conv_node.target, torch.ops.aten.linear.default)
+        else:
+            self.assertTrue(_is_conv_node(conv_node))
         self.assertEqual(
             scale_factor_reshape_node.target, torch.ops.aten.reshape.default
         )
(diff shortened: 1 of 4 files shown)


andrewor14:
@claude review

PR #4507, opened 2026-06-18
Add weight-only quantization MoE example

### LINKED ISSUE #729: MoE example
@felipemello1 shared this PR with me  https://github.com/vllm-project/vllm/pull/7415

My sense is we should already be able to support this with 

```python
from torchao.quantization.quant_api import quantize_, int4_weight_only
quantize_(m, int8_weight_only())
```

Would just need a good example to showcase this

Also cc @jcaip and @cpuhrsch who've thought a lot more about MoE than me

mykolas-perevicius:
### Why this PR

Issue #729 notes that `quantize_()` should already support Mixture-of-Experts (MoE) models, but there is no example or tutorial demonstrating weight-only quantization on an MoE — users have no reference for the workflow even though the API reportedly covers it. The existing `examples/quantize_llama_4.py` quantizes routed experts to **float8 w8a8 dynamic** (activation + weight), not weight-only, and `torchao/prototype/moe_training/` is MoE *training*, a different feature. So the weight-only-MoE showcase gap is real.

I verified the maintainer's claim before writing anything: applying `Int8WeightOnlyConfig` to an MoE block's expert linears works unmodified — expert weights become `Int8Tensor`, the model shrinks ~3.9x, and outputs stay within ~45 dB SQNR of fp32. The contribution is therefore purely additive: a runnable example plus a unit test, with no core changes.

### What this PR does

Adds `examples/quantize_moe.py`: it builds a small token-choice top-2 MoE block (a softmax router plus `nn.Linear` experts) and quantizes **only the expert weights** via `quantize_(model, Int8WeightOnlyConfig(), filter_fn=is_expert_linear)`. The router is deliberately left in high precision — quantizing it would change token-to-expert routing decisions, not just numerics.

The script is self-verifying: it prints before/after weight types and serialized sizes, runs a forward pass, reports SQNR vs the fp32 baseline, and asserts (experts quantized, router not, ≥1.5x smaller, SQNR > 25 dB), exiting non-zero on any failure.

- `--dtype int8|int4` and `--device` flags; int4 is gated behind a hardware/dependency warning.
- The docstring points users with real fused-3D-expert checkpoints (e.g. `meta-llama/Llama-4-Scout-17B-16E-Instruct`) at the `FqnToConfig` + `PerRow(1)` pattern, mirroring `examples/quantize_llama_4.py`.

It also adds a unit test — `TestQuantFlow.test_int8_weight_only_moe_experts_only` in `test/quantization/test_quant_api.py` (with a module-level `ToyMoEModel` modeled on the file's `ToyLinearModel`) — since CI lints examples but does not execute them, so without a test MoE support stays demonstrated-but-unguarded.

### Relevant issues

Closes #729

### Test plan / evidence

CI lints `examples/` but does not execute them (only `tutorials/` run via `run_tutorials.yml`), so the example is self-asserting and the unit test provides the regression guard.

**CPU (macOS, torch 2.12):** example runs twice identically — experts → `Int8Tensor`, router `float32`, serialized 8.40 → 2.15 MB (3.90x), SQNR 45.1 dB, all asserts pass. `pytest test/quantization/test_quant_api.py -k moe` → 1 passed.

**CUDA (RTX 3090, SM 8.6, torch 2.12.1+cu130):**

```
$ python examples/quantize_moe.py --device cuda
baseline | expert weight: Parameter (torch.float32) | serialized: 8.40 MB
int8 wo  | expert weight: Int8Tensor | router: Parameter (torch.float32) | serialized: 2.15 MB (3.90x smaller) | SQNR vs float32: 45.1 dB
quantization of the MoE model succeeded

$ pytest test/quantization/test_quant_api.py -k moe -q
1 passed

$ pytest test/quantization/test_quant_api.py
18 passed, 28 skipped, 0 failed
```

The 28 skips are all hardware-gated (`Need SM 8.9+` / `Checkpoints are produced in SM90+` — float8 paths on this SM 8.6 card) or pre-existing unconditional skips; none are introduced by this change.

### Acceptance criteria

- [x] Tests added for new behavior (`test_int8_weight_only_moe_experts_only` + `ToyMoEModel`)
- [x] All tests passing (CPU and CUDA; pre-existing skips/failures unrelated, verified via `git stash`)
- [x] Follows project style guide (`ruff check` + `ruff format` clean, ruff 0.11.6; BSD-3 header via `scripts/check_copyright_header.py`)
- [x] No breaking changes (purely additive — new example + new test, no core changes)
- [x] Documentation updated where applicable (example docstring with run instructions, hardware note, and fused-3D-expert pointer)

### Open questions for maintainers

1. **Example location** — top-level `examples/quantize_moe.py` (matching `examples/quantize_llama_4.py`, added in #3408) vs `examples/inference/` (hinted by the `literalinclude` in `quant_api.py`)?
2. **Keep or drop the unit test?** #3408 was single-file. Since CI never runs examples, I added the test so MoE support is regression-checked — happy to drop it if you prefer the single-file shape.
3. **`examples/README.md` entry** — should I add one? #3408 didn't.
4. **Supported public path for int4?** The int4 default path requires `mslk >= 1.0.0`, but the `mslk` package on public PyPI is a `0.0.0` placeholder and `torchao/utils.py:1226` gates real availability on `is_fbcode()`. On a fresh RTX 3090 with the public wheel, `--dtype int4` raises `ImportError: Requires mslk >= 1.0.0` (same on CPU and CUDA). What's the supported public way to exercise int4 weight-only? (The example gates int4 behind a flag + warning, so the int8 showcase is unaffected.)

### Note

Minor and separate from this PR: `torchao/quantization/quant_api.py:1499` has an invalid escape `'\.'` in a docstring that triggers `SyntaxWarning: invalid escape sequence` on import under Python ≥ 3.12. Happy to send a one-line `r"""` micro-fix as its own PR.

### DIFF
--- a/examples/quantize_moe.py
+++ b/examples/quantize_moe.py
@@ -1,0 +1,166 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+A script demonstrating weight-only quantization of a Mixture-of-Experts (MoE)
+model with `quantize_()`.
+
+It builds a small token-choice top-2 MoE block (a router plus ``nn.Linear``
+experts), quantizes only the expert weights — the router is left in high
+precision so routing decisions are unchanged — and verifies the result:
+expert weights become quantized tensor subclasses, the serialized model
+shrinks roughly 4x (int8), and a forward pass stays numerically close to the
+float32 baseline (SQNR).
+
+Run on CPU (default, int8 weight-only):
+
+    python examples/quantize_moe.py
+
+Run int4 weight-only (requires CUDA and the `mslk` package):
+
+    python examples/quantize_moe.py --dtype int4 --device cuda
+
+Note: real MoE checkpoints (e.g. `meta-llama/Llama-4-Scout-17B-16E-Instruct`)
+often store experts as fused 3D ``(num_experts, K, N)`` parameters instead of
+``nn.Linear`` modules. For those, configure quantization per-parameter with
+``FqnToConfig`` and ``PerRow(1)`` granularity — see
+``examples/quantize_llama_4.py`` for a complete float8 version of that flow.
+"""
+
+import argparse
+import copy
+import io
+
+import torch
+import torch.nn.functional as F
+from torch import nn
+
+from torchao.quantization import Int4WeightOnlyConfig, Int8WeightOnlyConfig, quantize_
+from torchao.quantization.utils import compute_error
+
+
+class Expert(nn.Module):
+    """A standard FFN expert: up-projection, SiLU, down-projection."""
+
+    def __init__(self, dim, hidden_dim):
+        super().__init__()
+        self.up = nn.Linear(dim, hidden_dim, bias=False)
+        self.down = nn.Linear(hidden_dim, dim, bias=False)
+
+    def forward(self, x):
+        return self.down(F.silu(self.up(x)))
+
+
+class MoEBlock(nn.Module):
+    """A token-choice top-k MoE block: softmax router + ``nn.Linear`` experts."""
+
+    def __init__(self, dim=256, hidden_dim=512, num_experts=8, top_k=2):
+        super().__init__()
+        self.router = nn.Linear(dim, num_experts, bias=False)
+        self.experts = nn.ModuleList(
+            Expert(dim, hidden_dim) for _ in range(num_experts)
+        )
+        self.top_k = top_k
+
+    def forward(self, x):
+        batch, seq, dim = x.shape
+        flat = x.reshape(-1, dim)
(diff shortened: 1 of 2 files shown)


mykolas-perevicius:
Hi @jcaip @msaroufim — this is my first contribution to torchao, addressing #729.

It adds a runnable weight-only quantization example for MoE models (`examples/quantize_moe.py`) plus a unit test. It quantizes only the expert linears via `quantize_(..., filter_fn=...)` — the router is kept in high precision so routing decisions are unchanged — and the script self-verifies the size reduction (~3.9x int8) and SQNR. Validated end-to-end on CPU (macOS) and CUDA (RTX 3090).

I left a few open questions in the description that I'd value your steer on — especially the preferred example location (`examples/` vs `examples/inference/`) and whether you'd like the unit test kept (happy to drop it for a single-file PR). Would appreciate a review when you have time. Thank you!

PR #4508, opened 2026-06-18
[x86] Fix missing fake metadata in quantized concat linear CPU fusion

Yuxingwang-intel:
This PR adds fake tensor metadata for the fused qlinear node created by the quantized concat linear CPU fusion pass.

The pass replaces multiple qlinear nodes with one fused qlinear node followed by split_with_sizes. Since the new qlinear node is manually created, it does not automatically carry meta["val"]. Without this metadata, FakeTensorUpdater may treat the fused qlinear output as None when updating the following split_with_sizes node.

The fused output shape is inferred from the original qlinear users.


Yuxingwang-intel:
Hi @Valentine233, could you please help review this PR. Thanks!

Xia-Weiwen (approved):

PR #4513, opened 2026-06-19
Add fused MXFP4 (E2M1) + Random Hadamard Transform CuTe DSL quantize cast (SM100)

santoshmo:
**What:** A single-pass, max-bandwidth CuTe DSL kernel that applies a per-32-block Random Hadamard Transform
  (register FWHT(32) + sign vector) and quantizes to MXFP4 — E2M1 data (2 codes/byte) + E8M0 block scales — writing
  packed `[M, K//2]` qdata and cuBLAS-blocked (or plain) scales in one streaming pass. Each thread owns one full
  32-element block: it reads the 32 inputs with forced 128-bit loads, runs FWHT+sign in registers, computes the
  E8M0 scale, and writes the 16 packed output bytes with a single 128-bit store plus one scale byte. No shared
  memory, no TMA, no warp specialization; heavy per-thread ILP hides the load latency and the E2M1/E8M0
  conversions; the saturating `cvt.rn.satfinite.e2m1x2` removes the need for explicit ±6 clamps. Lives in
  `torchao/prototype/mx_formats/cutedsl/`; opt-in via `MXTensor.to_mx(..., mxfp4_cast_kernel_choice=CUTEDSL,
  rht_sign_vector=...)`. Self-contained (own E2M1 PTX pack + register FWHT); reuses only existing torchao infra
  (`to_blocked`, `ScaleCalculationMode`, the CuTe DSL runtime gate, shared E8M0/cvt helpers). RHT is fused into the
  cast to suppress outliers before 4-bit quant without a separate transform pass.

  ### Accuracy
  - **Bit-exact** (`rtol=0, atol=0`) packed qdata **and** swizzled scales — validated byte-for-byte across
  scaling_mode {FLOOR, RCEIL} × swizzled {on, off} × dtype {bf16, fp32} × shapes {128×128, 128×384, 256×512, 
  1024×2048} (incl. single-tile / OOB). `test_mxfp4_rht_cutedsl.py` (16 tests) passes; the op is bit-exact vs
  torchao's emulated `to_mx(torch.float4_e2m1fn_x2)` on identical RHT input.
  - **SQNR** vs the high-precision dense `(x @ H)·sign` reference: ~18.8 dB (FLOOR / RCEIL).
  - **Component-level:** register FWHT matches the dense Hadamard to ~7e-7; the FP4 E8M0 scale helpers are 
  bit-exact vs eager `to_mx`.

  ### Performance (B200, CUPTI device time, bf16 in; RHT path)
  | Shape (M×K)  | Latency | Effective BW |
  |--------------|---------|--------------|
  | 8192 × 8192  | ~53 µs  | ~3.2 TB/s    |
  | 2048 × 8192  | ~15 µs  | ~2.9 TB/s    |
  | 4096 × 4096  | ~18 µs  | ~2.4 TB/s    |

  Reported as CUPTI kernel self-time (true kernel throughput), excluding the per-call host dispatch/alloc overhead
  a wall-clock timer would fold in — `benchmarks/mx_formats/cast_bench.py` was switched to device time for the
  cutedsl MXFP4 modes. The RHT path is FWHT(32)-compute-bound (~2.4–3.2 TB/s, ~30–40% of HBM peak); the plain
  no-RHT cast (not exposed by this RHT-only op) reaches ~5 TB/s (≈ HBM roofline).

  **Note:** Ships FLOOR + RCEIL; EVEN/CEIL are not included.

### DIFF
--- a/benchmarks/mx_formats/cast_bench.py
+++ b/benchmarks/mx_formats/cast_bench.py
@@ -114,6 +114,27 @@ def benchmark_cuda_function_in_microseconds(f, *args):
         return ((end - start) / 100) * 1e6  # Convert to microseconds
 
 
+def benchmark_cuda_function_device_time_us(f, *args, iters=100, warmup=20):
+    """Mean true on-GPU kernel time in microseconds (CUPTI device time).
+
+    Unlike ``benchmark_cuda_function_in_microseconds`` (a GPU-timeline span that
+    still captures per-call host launch/alloc overhead), this sums the CUDA
+    kernel self-times so the result reflects kernel throughput. Use for kernels
+    whose Python wrapper has non-trivial host cost (e.g. the CuTeDSL casts). The
+    warmup also covers the one-time CuTeDSL JIT compile.
+    """
+    from torch.profiler import ProfilerActivity, profile
+
+    for _ in range(warmup):
+        f(*args)
+    torch.cuda.synchronize()
+    with profile(activities=[ProfilerActivity.CUDA]) as prof:
+        for _ in range(iters):
+            f(*args)
+        torch.cuda.synchronize()
+    return sum(k.self_device_time_total for k in prof.key_averages()) / iters
+
+
 def run(
     M: int = 16384,
     K: int = 16384,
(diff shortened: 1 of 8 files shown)


santoshmo:
Consolidated with NVFP4 in #4517

PR #4516, opened 2026-06-19
Near-roofline NVFP4 quantization kernel 

santoshmo:
**What:** A single-pass, max-bandwidth CuTe DSL kernel that quantizes to NVFP4 — E2M1 data (2 codes/byte) +
  two-level E4M3 block scales (per-16-block E4M3 × a per-tensor global scale) — writing packed `[M, K//2]` qdata
  and cuBLAS-blocked (or plain) E4M3 scales, with an **optional** fused per-16-block Random Hadamard Transform
  (register FWHT(16) + sign vector). Each thread owns two consecutive 16-blocks (= 32 input elements = 16 output
  bytes): it reads them with forced 128-bit loads, computes the two block scales, and writes the 16 packed bytes
  with a single 128-bit store. No shared memory, no TMA, no warp specialization; heavy per-thread ILP hides the
  load latency and the E2M1/E4M3 conversions; the saturating `cvt.rn.satfinite.e2m1x2` removes the need for
  explicit ±6 clamps. Lives in `torchao/prototype/mx_formats/cutedsl/`; exposed as a selectable backend via
  `NVFP4Tensor.to_nvfp4(..., use_cutedsl_kernel=True)` (alongside the eager and Triton backends), and as the
  default for the gated cutedsl op `nvfp4_rht_quantize_cutedsl_2d` (the prior warp-specialized TMA kernel is kept
  as a `use_maxbw=False` fallback). Self-contained (own E2M1 PTX pack + register FWHT); reuses only existing
  torchao infra (`to_blocked` / `from_blocked`, `per_tensor_amax_to_scale`, the CuTe DSL runtime gate, shared
  E4M3/cvt helpers).

  ### Accuracy
  - **Byte-for-byte identical** packed qdata **and** scales vs the prior warp-specialized TMA kernel — validated
  across RHT {on, off} × swizzled {on, off} × dtype {bf16, fp32} × shapes incl. single-tile / OOB (24/24 op-level
  combos; 60/60 vs a kernel-faithful replica).
  - **vs eager `nvfp4_quantize`:** fp32 fully bit-exact; bf16 scale bytes 100% bit-exact, qdata within ≤1 FP4 code 
  step on ≤0.4% of elements (benign multiply-vs-divide reassociation; identical SQNR).
  - **vs the real Triton NVFP4 kernel** (rowwise path): scale bytes 100% bit-exact, qdata within ≤1 FP4 step 
  (several shapes fully bit-exact).
  - **SQNR** vs input ~20.4 dB — identical to the eager and Triton paths. `test_nvfp4_rht_cutedsl.py` passes; 
  `NVFP4Tensor` non-Triton tests pass.

  ### Performance (B200, CUPTI device time, @8192×8192, bf16 in)
  | Mode                 | Latency | Effective BW |
  |----------------------|---------|--------------|
  | plain, unswizzled    | ~34 µs  | ~5.1 TB/s    |
  | plain, swizzled      | ~37 µs  | ~4.6 TB/s (**1.66×** the legacy TMA path) |
  | RHT, swizzled        | ~52 µs  | ~3.3 TB/s    |

  Reported as CUPTI kernel self-time (true kernel throughput). The plain cast is bandwidth-bound and approaches the
  HBM roofline at scale (~6.2 TB/s @ 16384×16384); the swizzled scatter store costs ~10%, and the RHT path is 
  FWHT(16)-compute-bound. `NVFP4Tensor`'s backend uses the plain path (no RHT).

  **Note:** `use_cutedsl_kernel` is 2D + scalar/`None` `per_tensor_scale` only (asserts otherwise); 3D per-expert 
  cases use the existing eager/Triton paths.

### DIFF
--- a/benchmarks/mx_formats/cast_bench.py
+++ b/benchmarks/mx_formats/cast_bench.py
@@ -8,8 +8,13 @@ from typing import Tuple
 
 import fire
 import torch
-import triton
-from triton.testing import do_bench
+
+try:
+    import triton
+    from triton.testing import do_bench
+except ImportError:
+    triton = None
+    do_bench = None
 
 from torchao.prototype.mx_formats.config import ScaleCalculationMode
 from torchao.prototype.mx_formats.kernels import (
@@ -94,7 +99,19 @@ def to_nvfp4_reference_triton_swizzle(x_hp):
 
 
 def benchmark_cuda_function_in_microseconds(f, *args):
-    return do_bench(lambda: f(*args), return_mode="median") * 1e3
+    if do_bench is not None:
+        return do_bench(lambda: f(*args), return_mode="median") * 1e3
+    else:
+        # Fallback timing when triton is not available
+        import time
+
+        torch.cuda.synchronize()
+        start = time.perf_counter()
+        for _ in range(100):
+            f(*args)
+        torch.cuda.synchronize()
+        end = time.perf_counter()
+        return ((end - start) / 100) * 1e6  # Convert to microseconds
 
 
 def run(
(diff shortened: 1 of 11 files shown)


santoshmo:
Consolidated with MXFP4 in #4517

PR #4518, opened 2026-06-20
fix: use `Int4WeightOnlyConfig.group_size` in fake quant configs

### LINKED ISSUE #3572: `group_size` hardcoded in `_infer_fake_quantize_configs`.
Hi - I noticed today that `group_size` has been hardcoded to 128 [here](https://github.com/pytorch/ao/blob/main/torchao/quantization/qat/fake_quantize_config.py#L397) for `Int4WeightOnlyConfig`. Is this intentional?

oriollinan:
Following the discussion with @andrewor14 in #3572, this PR updates the `Int4WeightOnlyConfig` QAT path to use the configured `group_size` instead of the hardcoded value `128`.

Previously, `_infer_fake_quantize_configs` always generated fake quantization configs with `group_size=128`, regardless of the value specified in `Int4WeightOnlyConfig`. As a result, QAT did not honor the configured group size, unlike the corresponding PTQ path.

Fixes #3572

### DIFF
--- a/test/quantization/test_qat.py
+++ b/test/quantization/test_qat.py
@@ -1951,7 +1951,8 @@ class TestQAT(TestCase):
         self.assertEqual(weight_config.group_size, 128)
         self.assertEqual(weight_config.activation_dtype, e4m3_dtype)
 
-    def test_infer_int4_weight_only_config(self):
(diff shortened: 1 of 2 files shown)


vkuzo:
@claude review

PR #4531, opened 2026-06-23
Revert "pt2e: preserve mutable buffer inputs during prepare"

liangel-02:
Reverts pytorch/ao#4466

failing internal test


andrewor14 (approved):

tom-arm:
Is it possible to share what regressions this change caused?

PR #4534, opened 2026-06-25
Use smallest_normal instead of eps to clamp quant scales

GregoryComer:
In _choose_qparams_affine, set the default eps (used to clamp scales to avoid zero) to finfo smallest_normal instead of eps. This significantly improves accuracy on some models when using bf16 scales.

This is because eps is the delta from 1 to the next representable value. This can be significantly larger than the actual smallest non-zero value. For bf16, it ends up clamping at 2^-7, which very course. Bf16 can represent much smaller values. In practice, this means small scales in bf16 get clamped to a much larger value, which degrades accuracy.

The issue doesn't really arise with f16/f32 because the eps is pretty small, even if it is many orders of magnitude larger than the smallest non-zero value that can be represented. Bf16 is particularly vulnerable due to the tiny fraction. 

**Question for reviewers: Is there any other reason to clamp to eps other than to avoid zero scales?**

End-to-end perplexity numbers:
Gemma3-1b, quantized to int4, GS=32/256, wikitext, 40k tokens, 1024-token windows:
| Activations | Group size | Before | After | Delta |
|---|---|---|---|---|
| f32  | 32  | 29.9809 | 29.9809 | 0.000 |
| f16  | 32  | 29.4767 | 29.4741 | −0.003 |
| bf16 | 32  | 40.5104 | 29.9401 | −10.57 |
| f32  | 256 | 66.6410 | 66.6410 | 0.000 |
| f16  | 256 | 64.8682 | 64.8673 | −0.001 |
| bf16 | 256 | 90.9888 | 63.5593 | −27.43 |

On Gemma3-1b, about 52% of scales get clamped up currently in bf16, often by orders of magnitude. Post-fix, no scales are clamped as there are legitimate bf16 representations that are greater than zero.

### DIFF
--- a/test/quantization/test_quant_primitives.py
+++ b/test/quantization/test_quant_primitives.py
@@ -589,7 +589,7 @@ class TestQuantPrimitives(unittest.TestCase):
         scale, zero_point = choose_qparams_affine(
             input, mapping_type, block_size, dtype
         )
-        eps = torch.finfo(torch.float32).eps
+        eps = torch.finfo(torch.float32).smallest_normal
         self.assertEqual(scale, eps)
 
     @unittest.skipIf(
--- a/torchao/prototype/parq/quant/uniform_torchao.py
+++ b/torchao/prototype/parq/quant/uniform_torchao.py
@@ -86,7 +86,7 @@ class UnifTorchaoQuantizer(Quantizer):
     ) -> tuple[Tensor, Tensor]:
         self._init_quant_min_max(b)
         if self.eps is None:
-            self.eps = torch.finfo(p.dtype).eps
+            self.eps = torch.finfo(p.dtype).smallest_normal
 
         # assume that p has already been grouped in QuantOptimizer.step
         block_size = (1, p.size(-1)) if dim is not None else p.size()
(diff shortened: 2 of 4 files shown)


GregoryComer:
Note: I intentionally didn't touch the other variants of `_choose_qparams_affine in the file` to minimize blast radius, but I expect they would benefit from the same tweak. Feel free to let me know if we want to update those, too.

Edit: Tests enforce parity, so I've updated the other variants.

GregoryComer:
Converting back to draft to fix the test failure.

Edit: Need to update parq logic to match as tests expect exact equality.

metascroy:
LGTM!

This is the same fix I suggested on this PR from Mergen as well: https://github.com/pytorch/ao/pull/4362

I don't see an issue, but maybe someone from AO could approve to make sure there is no blast radius on server side (cc @andrewor14).

(Note, I don't think this bug affects the default export paths we have on ET, which use HQQ).

metascroy (approved):

GregoryComer:
The remaining test failures (test_fsdp2_correctness and test_uneven_shard) fail on trunk with the same errors, so they appear unrelated. See https://github.com/pytorch/ao/actions/runs/28133078180/job/83313695594.

andrewor14:
@claude review, analyze potential blast radius?

andrewor14:
Looks good. I agree this is the more correct fix. The blast radius seems to only affect the older int configs, seems fine. But yeah there wasn't really any particular reason to use eps instead. We just treated eps as the smallest possible value before.

andrewor14 (approved):

PR #4535, opened 2026-06-25
[XPU][CI] Add test/low_bit_optim.py to XPU CI

karol-brejna-i:
Adds test/test_low_bit_optim.py test as part of XPUs CI.
The test takes about 150 seconds on BMG B570.

### DIFF
--- a/.github/scripts/ci_test_xpu.sh
+++ b/.github/scripts/ci_test_xpu.sh
@@ -23,4 +23,5 @@ pytest -v -s --ignore=torchao/test/quantization/pt2e/test_x86inductor_fusion.py 
         torchao/test/float8/ \
         torchao/test/integration/test_integration.py \
         torchao/test/prototype/ \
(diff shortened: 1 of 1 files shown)


karol-brejna-i:
The test passes on BMG, but seems to fail on PVC.
The runner is a PVC node (dut7361_pvc_card_5).

`torchao/test/test_low_bit_optim.py::TestQuantize::test_bf16_stochastic_round_device_xpu_compile_False` uses randomness for rounding the numbers. PVC implementation is probably different then BMG. 
On the other hand, it seems that the tolerance margin in the test is too high (1e-5).  
(See: https://github.com/pytorch/ao/blob/260ac851fdca24130b921e663e50b710205d1287/test/test_low_bit_optim.py#L109-L120)

Let's think what should be our next steps here.

karol-brejna-i:
I start investigating the problem on PVC.

karol-brejna-i:
Replaced by #4576

PR #4537, opened 2026-06-26
Re-enable building under a free-threaded Python interpreter

nascheme:
This PR re-applies the changes from #4428.

Free-threaded interpreters do not support the CPython Limited C API, and hence use of the limited API has to be disabled. With this one change, `torchao` builds fine.

The `is_freethreaded` function is in widespread use, and documented as the canonical way of doing this check here:
https://py-free-threading.github.io/running-gil-disabled/#check-if-using-a-free-threaded-build

PR #4428 (commit f43befe6b) was inadvertently reverted by #4446 ("Back out D105996100 — breaks `fbpkg build light_cli`"). That back-out was meant to revert only the multi-ISA work in #4159 (D105996100), but instead of computing a reverse diff it restored a whole-file snapshot of setup.py that predated both #4159 and #4428, silently clobbering the free-threading support.

Re-applied on top of the current `add_options_for_x86` base; the only adaptation from the original is the placement of `is_freethreaded()` (both functions are kept).

(cherry picked from commit f43befe6b7b8a75776bfe905dfa8ddbc287fa9fd)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -541,7 +541,9 @@ def get_extensions():
 
     maybe_cpython_limited_api = []
     if not is_freethreaded():
-        maybe_cpython_limited_api += [f"-DPy_LIMITED_API={min_supported_cpython_hexcode}"]
+        maybe_cpython_limited_api += [
+            f"-DPy_LIMITED_API={min_supported_cpython_hexcode}"
+        ]
 
     extra_link_args = []
     extra_compile_args = {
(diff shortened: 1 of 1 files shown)


vkuzo:
@claude review, why was the original PR reverted?

vkuzo:
lgtm! and sorry for long delay, I was OOO and now back.

Could you please fix `ruff`, by running `ruff format && ruff check . --fix` from root? then should be good to go

nascheme:
Reformatted with ruff, check passes without errors.

PR #4540, opened 2026-06-27
Preserve nested control-flow subgraphs during pt2e convert

### LINKED ISSUE #4455: convert_pt2e crashes in DuplicateDQPass — GraphModule rebuild drops submodules only referenced by other submodules' internal graphs
### Description

When calling convert_pt2e(prepared_model) on a model where one submodule's internal graph references another submodule (common in complex architectures like Qwen2.5), the conversion crashes inside DuplicateDQPass with:
```
RuntimeError: Node submod_3 target submod_1 references nonexistent attribute submod_1 of ...
```

### Steps to reproduce

```python
class Inner(torch.nn.Module):
    """Leaf module with a parameter — will become 'submod_1'."""
    def __init__(self):
        super().__init__()
        self.weight = torch.nn.Parameter(torch.randn(4, 4))

    def forward(self, x):
        return torch.nn.functional.linear(x, self.weight)


class Block(torch.nn.Module):
    """Composite module that calls Inner — will become 'submod_3'."""
    def __init__(self, inner: Inner):
        super().__init__()
        self.inner = inner  # registered as 'submod_1'

    def forward(self, x):
        return self.inner(x)

def build_buggy_model():
    """
    Build a model whose structure mimics the failure case:

        main graph  ──calls──▶  submod_3 (Block)
        submod_3.graph  ──calls──▶  submod_1 (Inner)

    submod_1 is NOT directly called by the main graph, only by submod_3's
    internal graph.
    """
    inner = Inner()
    block = Block(inner)

    model = torch.nn.Module()
    model.add_module("submod_1", inner)   # registered on top level
    model.add_module("submod_3", block)   # registered on top level

    # Build a main graph that ONLY calls submod_3
    graph = Graph()
    x = graph.placeholder("x")
    out = graph.call_module("submod_3", args=(x,))
    graph.output(out)
    model.graph = graph

    return model
```

### Environment

```text
torch version:2.9.0+cu128
torchao:0.17.0
Hardware:cpu
OS:ubuntu
```

atharvjairath:
Fixes #4455.

`GraphModule.delete_all_unused_submodules()` only inspects the immediate graph, so `convert()` could delete submodules that are referenced only from nested control-flow subgraphs. In nested `while_loop` cases, this left dangling references and caused `DuplicateDQPass` graph linting to fail.

This adds a recursion-aware cleanup wrapper that preserves submodules referenced from nested `GraphModule` children, and adds a regression test covering nested `while_loop` conversion.

Test plan:
- `python -m ruff check torchao/quantization/pt2e/convert.py test/quantization/pt2e/test_quantize_pt2e.py`
- `python -m pytest test/quantization/pt2e/test_quantize_pt2e.py::TestQuantizePT2E::test_nested_while_loop_convert_preserves_subgraphs -q`
- `python -m pytest test/quantization/pt2e/test_duplicate_dq.py -q`
- `git diff --check`

### DIFF
--- a/test/quantization/pt2e/test_quantize_pt2e.py
+++ b/test/quantization/pt2e/test_quantize_pt2e.py
@@ -3628,6 +3628,54 @@ class TestQuantizePT2E(PT2EQuantizationTestCase):
         sqnr = compute_error(output_ref, output)
         self.assertGreater(sqnr, 35, f"SQNR too low: {sqnr} dB")
 
+    def test_nested_while_loop_convert_preserves_subgraphs(self):
+        """Regression test for https://github.com/pytorch/ao/issues/4455.
+
+        When a while_loop body contains a nested while_loop, the inner cond/body
+        subgraphs are referenced only from inside the outer body subgraph. The
+        cleanup in convert() (``delete_all_unused_submodules``) only inspects the
+        top-level graph, so it used to delete those inner subgraphs, leaving
+        dangling references that crashed graph linting in DuplicateDQPass.
+        """
+        from torch._higher_order_ops.while_loop import while_loop
+
+        class NestedWhileLoopModel(torch.nn.Module):
+            def __init__(self):
+                super().__init__()
+                self.linear = torch.nn.Linear(5, 5)
+
+            def forward(self, x):
+                def inner_cond(j, y):
+                    return j < 2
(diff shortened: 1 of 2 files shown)


vkuzo:
@claude review

PR #4542, opened 2026-06-28
Fix constant_fold cleanup for shared get_attr targets

### LINKED ISSUE #4420: constant_fold deletes shared get_attr target while another get_attr for the same target is still live

yinli-systems:
## Summary

Fixes #4420.

`constant_fold` deleted a GraphModule attribute as soon as it found one dead `get_attr` node. When another live `get_attr` node referenced the same target, the graph became invalid and failed linting with:

```text
Node shared_1 target shared references nonexistent attribute shared
```

This change snapshots the targets still used by live `get_attr` nodes, deletes an underlying attribute only when no live node references it, and avoids attempting to delete the same dead target more than once. Dead graph nodes are still removed as before.

The regression constructs two `get_attr` nodes for one registered buffer: one becomes dead after constant folding while the other remains in a live addition. It verifies graph linting, attribute retention, and both folded and live outputs.

## Validation

- `PYTHONDONTWRITEBYTECODE=1 PYTHONPATH=. .venv/bin/python -m unittest test.quantization.pt2e.test_quantize_pt2e.TestConstantFold.test_shared_get_attr_target_stays_live` (1 passed)
- Fault injection with the old cleanup logic: the same test fails with the expected nonexistent-attribute graph error.
- `ruff==0.11.6 check` on both changed files.
- `ruff==0.11.6 format --check` on both changed files.
- `git diff upstream/main --check`.

## AI assistance

AI tooling assisted with implementation review and validation. I reviewed the final diff and reproduced both the old failure and the fixed behavior locally.

### DIFF
--- a/test/quantization/pt2e/test_quantize_pt2e.py
+++ b/test/quantization/pt2e/test_quantize_pt2e.py
@@ -22,7 +22,7 @@ from torch.ao.quantization.qconfig import (
     per_channel_weight_observer_range_neg_127_to_127,
     weight_observer_range_neg_127_to_127,
 )
-from torch.fx import Node, symbolic_trace
+from torch.fx import Graph, GraphModule, Node, symbolic_trace
 from torch.testing import FileCheck
 from torch.testing._internal.common_quantization import (
     NodeSpec as ns,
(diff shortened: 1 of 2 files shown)


yinli-systems:
AI-assisted duplicate audit (reviewed for factual accuracy):

> #4431 predates this PR, implements the same live-target guard, already received maintainer feedback, and its regression additionally asserts that the dead duplicate `get_attr` node is erased. I revalidated that the underlying bug still exists on current `main` and that the equivalent fix passes on PyTorch 2.13, but this PR has no unique fix left to contribute.

Closing this duplicate so review can consolidate on #4431.

PR #4544, opened 2026-06-29
fix(_is_mslk_available): guard against missing CUDA/ROCm runtime

space192:
## Summary

`_is_mslk_available()` crashes with `OSError` when mslk is installed but no CUDA/ROCm runtime is present (e.g. CPU-only submission nodes).

The function is called at module level in `torchao/quantization/quantize_/workflows/float8/float8_tensor.py`, so the crash propagates through any transitive import chain that touches torchao — including `transformers`, `vllm`, etc. — even when the user never explicitly imports mslk or uses FP8.

**Repro:**
```bash
# On a CPU-only machine with mslk installed from the CUDA PyTorch index
python -c "import torchao"
# OSError: Could not load this library: .../mslk/mslk.so
```

## Fix

Check `torch.cuda.is_available()` before attempting `import mslk`. This covers both CUDA and ROCm (HIP maps to the CUDA API surface). A warning is emitted so users know mslk was skipped and why.

## Test plan

- Verified on a CPU-only node: `import torchao` succeeds with a warning instead of crashing.
- On GPU nodes: behavior is unchanged, mslk loads normally.


vkuzo:
I think it would be even better to fix this in MSLK, but a workaround sounds reasonable

@claude review

space192:
Thanks @vkuzo for the review ! Applied the changes claude suggested :)

Yeah this is more of a workaround than a fix, but felt like the change is small enough to be part of ao :)

space192:
not sure the CI failing is related to my changes 🤔

vkuzo:
@claude review

vkuzo:
@space192 thank you! any chance you could help rebase to kick off CI? For some reason it did not trigger on your latest changes.

space192:
@vkuzo just rebased on main ! we should be good now :) Thanks for your review btw !

vkuzo (approved):

PR #4545, opened 2026-06-29
Add MoE blockwise FP8 DTensor and compile coverage

iamzainhuda:
## Summary
- add DTensor TP support for blockwise FP8 MoE grouped GEMM, covering fwd, dgrad, and wgrad placements
- add custom-op wrappers and DTensor sharding rules for the emulated and DeepGEMM grouped GEMM paths so `torch.compile` can hit the production kernels
- add end-to-end FSDP2/TP coverage matching the blockwise FP8 linear tests, plus shared DTensor test utilities reused by both paths

this keeps the MoE path aligned with the blockwise FP8 linear DTensor design: shared DTensor helpers handle local tensor unwrap/rewrap, replicated metadata, placement checks, and common test parity utilities. The grouped MoE code still exercises the quantized training path rather than a BF16-only fallback.

Part of #3290

## Tests
```
pytest -q test/prototype/moe_training/test_config.py
NCCL_SOCKET_IFNAME=lo pytest -q test/prototype/moe_training/test_fp8_blockwise_grouped_mm.py
NCCL_SOCKET_IFNAME=lo pytest -q test/prototype/moe_training/test_fp8_blockwise_fsdp2.py
NCCL_SOCKET_IFNAME=lo pytest -q test/prototype/blockwise_fp8_training/test_dtensor.py
NCCL_SOCKET_IFNAME=lo pytest -q test/prototype/blockwise_fp8_training/test_fsdp2.py
```

For the broader related coverage:
```
NCCL_SOCKET_IFNAME=lo pytest -q test/prototype/moe_training/test_fp8_blockwise_deepgemm_backend.py
NCCL_SOCKET_IFNAME=lo pytest -q test/prototype/moe_training/test_fp8_blockwise_fsdp2_tp.py
```


iamzainhuda:
@claude review

iamzainhuda:
@claude review this PR again

test/prototype/moe_training/test_config.py
```diff
@@ -0,0 +1,28 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
vkuzo, line 1: https://github.com/pytorch/ao/tree/main/torchao/testing is the right top level dir for this, test utilities should not be mixed with product code. Same for other files in this PR

torchao/prototype/blockwise_fp8_training/kernels.py
```diff
@@ -91,17 +94,13 @@ def _is_row_major(x: torch.Tensor) -> bool:
...
-        (list(shard_dim0_outputs), [Shard(0), None, None]),
-        (list(shard_dim1_outputs), [Shard(1), None, None]),
-    ]
+def _ensure_column_major(x: torch.Tensor) -> torch.Tensor:
+    if _is_column_major(x):
+        return x
+    # DTensor sharding can materialize local shards of a logically column-major
+    # operand with regular row-major strides. Preserve the logical values while
+    # restoring the physical RHS layout required by _scaled_mm_v2.
+    return x.transpose(-2, -1).contiguous().transpose(-2, -1)
```
vkuzo, line 105: hmm, this is unexpected. Do we know why this is needed?

vkuzo:
do we need TP for experts? what's the motivation for that?

test/prototype/moe_training/test_config.py
iamzainhuda: makes sense, moved. kept the suite-specific builders local under `test/prototype`

torchao/prototype/blockwise_fp8_training/kernels.py
iamzainhuda: at the `blockwise_scaled_mm` custom-op boundary: DTensor local dispatch / redistribution can materialize the local RHS shard in a regular contiguous layout even when the logical operand is the transposed FP8 RHS produced by the quantizer, and `_scaled_mm_v2` requires a physically column-major RHS

iamzainhuda: i actually saw this issue in a job i ran internally, i can make this fix in a seperate PR if thats cleaner

PR #4551, opened 2026-07-02
Reuse direct grouped FP8 quantizers for blockwise MoE and simplify DeepGEMM path

iamzainhuda:
## Summary
- Move the performant direct grouped weight quantizers into neutral `grouped_weight_quant.py`, shared by the DeepGEMM and emulated blockwise MoE backends.
- Use one DeepGEMM-specific direct K-grouped wgrad activation quantizer for both operands at every supported dimension, removing dimension-based kernel routing and the generic-layout repack fallback (previously was based on 4096 dim routing based off flawed benchmarking).
- Delete the obsolete grouped weight quantizer wrapper custom ops from `grouped_kernels.py` now that both backends reuse the direct grouped weight quantizers.
- Updated tests and benchmarks

## Why the DeepGEMM-specific activation quantizer remains

For wgrad, each expert computes `(D_out, expert_tokens) @ (D_in, expert_tokens).T`. The token count is the ragged K dimension. DeepGEMM `k_grouped_fp8_gemm_nt_contiguous` therefore expects flat expert-major buffers: each expert is stored as `(D, expert_tokens)`, and those expert buffers are concatenated.

The generic TorchAO quantizers read the same row-major `(M_tokens, D)` activations but write either `(D, M_tokens)` or logical `(M_tokens, D)` with column-major storage. A plain flatten preserves dimension-major order across all experts, not the expert-major order required by DeepGEMM. Using these outputs requires slicing by expert and repacking into a new flat buffer.

The DeepGEMM-specific quantizer writes the final expert-major layout directly, with scales shaped `(D, total_valid_token_blocks)`. This avoids the intermediate TorchAO layout and repack.

### H100 measurements

Same-shape measurements on H100, balanced `E=8`, block size `128`, median of five benchmark runs. Modeled bandwidth counts the bf16 input read, FP8 output write, and FP32 scale write. The H100 reference is `3352.3 GB/s` peak and `3084.1 GB/s` achievable; `% BW` is relative to modeled achievable bandwidth, not a hardware-counter measurement.

| `(M_tokens, D)` | Direct quant | Direct GB/s | Direct % BW | TorchAO LHS quant + repack | TorchAO RHS quant + repack |
|---|---:|---:|---:|---:|---:|
| `(16384, 2048)` | `46.8 us` | `2171` | `70.4%` | `264.1 us` | `273.5 us` |
| `(16384, 7168)` | `140.2 us` | `2539` | `82.3%` | `495.2 us` | `492.1 us` |

Across all measured shapes, the direct quantizer achieved `2165-2592 GB/s`, or `70-84%` of modeled achievable H100 bandwidth, and was faster end-to-end. In eager execution, direct quantization was `3.4x-5.5x` faster than generic quantization plus repack; with `torch.compile(fullgraph=True)`, it remained `1.9x-5.9x` faster.

The DeepGEMM wgrad path therefore uses the direct quantizer unconditionally for both operands. The emulated backend continues to use the generic TorchAO quantizers because it consumes their native layouts.

## Test Plan
- `pytest -q test/prototype/moe_training/test_fp8_blockwise_deepgemm_backend.py`
- `pytest -q test/prototype/moe_training/test_fp8_blockwise_grouped_mm.py` 
- `python -m benchmarks.prototype.blockwise_fp8_training.bench_moe_grouped_kernels --shapes 16384,2048,7168,8`


torchao/prototype/blockwise_fp8_training/grouped_kernels.py
```diff
@@ -17,8 +23,6 @@
     _is_row_major,
     _prepare_blockwise_scaled_mm_rhs_scale,
     _scaling_type_value,
-    triton_fp8_blockwise_weight_quant_rhs,
```
vkuzo, line 20: we should also delete these functions if they are unused after this PR

torchao/prototype/blockwise_fp8_training/grouped_kernels.py
```diff
@@ -78,32 +82,14 @@ def triton_fp8_blockwise_weight_quant_grouped_transposed_rhs(
...
-        q, scale = triton_fp8_blockwise_weight_quant_transposed_rhs(
-            expert_weight,
-            block_size=block_size,
-            dtype=dtype,
-        )
-        q_out[expert_idx].copy_(q)
-        scale_out[expert_idx].copy_(scale)
-    return q_out, scale_out
+    # The direct kernel writes the transpose-compatible row-major layout;
+    # transpose views preserve TorchAO's public column-major grouped RHS contract.
```
vkuzo: I think this comment should explain what is happening in terms of `E, N, K`

vkuzo:
directionally looks good, can you add a PR description, test plan and address the comments?

torchao/prototype/blockwise_fp8_training/grouped_kernels.py
iamzainhuda: agreed! coming in next version - didn't get a chance to push it yet

iamzainhuda:
> directionally looks good, can you add a PR description, test plan and address the comments?

added - good to review now

vkuzo:
> Move direct grouped weight quantization into a neutral grouped_weight_quant module shared by the DeepGEMM and emulated blockwise MoE backends.

can we just leave them in the old location to preserve git history?

iamzainhuda:
> > Move direct grouped weight quantization into a neutral grouped_weight_quant module shared by the DeepGEMM and emulated blockwise MoE backends.
> 
> can we just leave them in the old location to preserve git history?

yeah sure, i moved it because it's shared in both backends. but is the thinking here since this is temporary (until we have kernel in core) to keep it separate?

vkuzo:
> Keep direct grouped weight quantization in the original deepgemm_quant.py location while sharing it between the DeepGEMM and emulated blockwise MoE backends.

Not sure I agree with this, the kernel that prepares the activations and weights for a float8 blockwise grouped gemm should not be related to "deepgemm", as the layout is just row-major.  I think in the current case, the right thing to do is to either revert the original PR that created the triton kernels that we don't actually need, or just delete them and switch the references to the existing kernels.  It's a bit confusing to couple deletion of kernels which were incorrectly added with code movement of old kernels into the same place.

iamzainhuda:
> > Keep direct grouped weight quantization in the original deepgemm_quant.py location while sharing it between the DeepGEMM and emulated blockwise MoE backends.
> 
> Not sure I agree with this, the kernel that prepares the activations and weights for a float8 blockwise grouped gemm should not be related to "deepgemm", as the layout is just row-major. I think in the current case, the right thing to do is to either revert the original PR that created the triton kernels that we don't actually need, or just delete them and switch the references to the existing kernels. It's a bit confusing to couple deletion of kernels which were incorrectly added with code movement of old kernels into the same place.

i agree that float8 blockwise grouped quantization should not be DeepGEMM-specific. That was the motivation for the earlier `grouped_weight_quant.py` iteration: keep the performant direct grouped quant kernels and use them for both DeepGEMM and emulated backends, and delete the older grouped wrapper kernels that are no longer needed. we established that the backend difference here is layout, so the emulated path can use transpose views and the DeepGEMM path can consume the direct layout

i combined deletion and move into one PR since they are linked. once both backends are sharing the same performant kernels, old grouped wrappers are dead code, and the shared kernels should live somewhere where it is clear they are shared among the two backends.

i can split the cleanup in a separate PR if you'd prefer. and for this PR i can delete the obsolete grouped wrappers and point both backends to the quant kernels in deepgemm_quant.py. and follow up PR with a rename/move to a generic location. i still prefer to do it in one PR, since end state of this one leaves the kernels shared in the two backends under deepgemm_quant name

for now the PR is following the "split PR" path i outlined earlier. to make sure we're on same page, in this PR:
- the performant quant kernels we will use in moe blockwise fp8 training are in deepgemm_quant.py
- the naming of the said kernels are now generic --> removed "deepgemm" from the naming
- deletes the obsolete grouped wrapper kernels in grouped_kernels.py
- points both deepgemm and emulated backend paths to the kernels in deepgemm_quant.py
- no grouped_weight_quant.py remains

vkuzo:
> points both deepgemm and emulated backend paths to the kernels in deepgemm_quant.py

i think the file name and dir names should also not contain `deepgemm` for kernels which are not specific to deepgemm.

iamzainhuda:
> > points both deepgemm and emulated backend paths to the kernels in deepgemm_quant.py
> 
> i think the file name and dir names should also not contain `deepgemm` for kernels which are not specific to deepgemm.

agreed - the shared kernels are now in ```grouped_weight_quant.py```

iamzainhuda:
State of the world after this PR is:

We have shared (for emulated and DeepGEMM backend) grouped weight quant kernels. 

The remaining DeepGEMM specific quant kernel is for DeepGEMM wgrad API. DeepGEMM K-grouped wgrad takes flat 1D buffers where each expert is concat as K-major slice on logical shape (dim, expert tokens), the scales and group_sizes are passed as Python list. Direct quant kernel writes the flat grouped 1d layout directly. 

this is needed because at dims 4096+ the direct kernel is better, because we can avoid the extra read/write that comes from flattening/repacking.  So, for DeepGEMM, the routing is dim >= 4096 uses the direct path and dim <4096 we use the torchAO quantizer + flatten path

benchmarks:
image

vkuzo:
here is my summary of the layout differences:

```
# layout requirements for wgrad 2dx2d
# PyTorch: grad_out_t [N, M] row-major, A [M, K] col-major, offs
# DeepGemm:
#  * grad_out_t is [N * M], laid out as [N * M_0, ..., N * M_e-1], each N * M_n block row-major
#  * A is [M * K], laid out as [K * M_0, ..., K * M_e-1], each K * M_0 block row-major
```

lmk if that matches your understanding.  If it does, I'd recommend:
1. keep two kernels for the 2dx2d case (as you already have)
2. clearly explain the layout differences between them (I think this part needs some work)
3. gate which kernel is used by a top level config (backend emulated vs backend deepgemm)

iamzainhuda:
> here is my summary of the layout differences:
> 
> ```
> # layout requirements for wgrad 2dx2d
> # PyTorch: grad_out_t [N, M] row-major, A [M, K] col-major, offs
> # DeepGemm:
> #  * grad_out_t is [N * M], laid out as [N * M_0, ..., N * M_e-1], each N * M_n block row-major
> #  * A is [M * K], laid out as [K * M_0, ..., K * M_e-1], each K * M_0 block row-major
> ```
> 
> lmk if that matches your understanding. If it does, I'd recommend:
> 
> 1. keep two kernels for the 2dx2d case (as you already have)
> 2. clearly explain the layout differences between them (I think this part needs some work)
> 3. gate which kernel is used by a top level config (backend emulated vs backend deepgemm)

Yup,

For 2) I added more comments to be clear in the code where things are happening and why.

For 3) It's already gated through ``` Float8TrainingOpConfig.kernel_preference``` , `EMULATED` forces the emulated backend and ```AUTO``` selects DeepGEMM when installed otherwise it falls back to emulated. There is no specifics "DeepGEMM" selection, if you want one I can add it. But AUTO and EMULATED felt sufficient here.

torchao/prototype/blockwise_fp8_training/deepgemm_quant.py
```diff
@@ -344,10 +95,11 @@ def triton_fp8_blockwise_act_quant_k_grouped_deepgemm(
     """
     Quantize activations directly for DeepGEMM K-grouped wgrad.
 
-    Input is logical (M_tokens, D), already padded/grouped by expert. Output is
-    a flat K-major DeepGEMM buffer: for each expert, data is stored as
-    (D, expert_tokens), then concatenated with the next expert. Scales are
-    stored as (D, total_valid_token_blocks).
+    Input is logical ``(M, D)``, already padded and concatenated by expert.
```
vkuzo: what is `D`? How about rewriting the docs in `M, K, N, E` nomenclature to be consistent with other casts related to scaled variants of `grouped_mm`?

torchao/prototype/blockwise_fp8_training/deepgemm_quant.py
```diff
@@ -390,54 +142,33 @@ def _triton_fp8_blockwise_act_quant_k_grouped_deepgemm_with_group_sizes(
     )
 
     D = x.shape[1]
+    assert block_size == 128, "DeepGEMM K-grouped quantization requires block_size=128"
```
vkuzo, line 154: is `k-grouped` from the literature? I understand what your code is doing but not clear why it's described as `k-grouped`.

torchao/prototype/moe_training/blockwise_fp8/grouped_mm_backend.py
```diff
@@ -103,28 +101,34 @@ def quantize_forward_rhs(
         block_size: int,
         dtype: torch.dtype,
     ) -> tuple[torch.Tensor, torch.Tensor]:
-        # The emulated backend consumes TorchAO's grouped RHS layout:
+        # The shared direct quantizer writes forward RHS as row-major
+        # (E, N, K) data with (E, N_blocks, K_blocks) scales. Transpose views
+        # convert that to TorchAO's grouped RHS contract for output = A @ B_t:
```
vkuzo: this contract is not from torchao, it's from the requirements of `torch._grouped_mm` and `torch._scaled_grouped_mm`

torchao/prototype/blockwise_fp8_training/deepgemm_quant.py
iamzainhuda: Updated the docstring to use E, M, K, and N throughout

torchao/prototype/moe_training/blockwise_fp8/grouped_mm_backend.py
iamzainhuda: Fixed. Both forward and dgrad comments now attribute the RHS layout to the contract required by torch._grouped_mm and torch._scaled_grouped_mm

torchao/prototype/blockwise_fp8_training/deepgemm_quant.py
iamzainhuda: This is DeepGEMM API terminology from k_grouped_fp8_gemm_nt_contiguous. DeepGEMM calls it K-grouped because the per-expert token count M_e is the varying reduction/K extent of (N, M_e) @ (M_e, K). I added the explanation to the comments.

torchao/prototype/moe_training/blockwise_fp8/grouped_mm.py
```diff
@@ -44,6 +44,12 @@ def _to_fp8_blockwise_then_scaled_grouped_mm(
 
     A has shape (M, K). B_t has shape (E, K, N), transposed and in
     per-expert column-major layout.
+
+    ``kernel_preference`` selects one backend for the complete autograd
+    operation. ``EMULATED`` uses PyTorch grouped-mm layouts and kernels;
+    ``AUTO`` uses DeepGEMM-specific layouts and kernels when supported and
```
vkuzo: repo's convention is for `AUTO` to only choose between real non-emulated backends.  This should be something like

DEEPGEMM for deep_gemm grouped_mm
AUTO, if you want, but it can only call DEEPGEMM or fail
EMULATED for emulation, AUTO cannot reach here

torchao/prototype/moe_training/blockwise_fp8/grouped_mm_backend.py
```diff
@@ -20,14 +20,12 @@
...
-from torchao.prototype.blockwise_fp8_training.deepgemm_quant import (
-    triton_fp8_blockwise_weight_quant_grouped_rhs_deepgemm,
-    triton_fp8_blockwise_weight_quant_grouped_transposed_rhs_deepgemm,
-)
 from torchao.prototype.blockwise_fp8_training.grouped_kernels import (
     emulated_blockwise_scaled_grouped_mm,
-    triton_fp8_blockwise_weight_quant_grouped_rhs,
-    triton_fp8_blockwise_weight_quant_grouped_transposed_rhs,
+)
+from torchao.prototype.blockwise_fp8_training.grouped_weight_quant import (
```
vkuzo, line 26: for this PR importing from `blockwise_fp8_training` is fine, IMO we should consolidate in `moe_training` before finalizing how torchtitan imports this logic

vkuzo (approved):

torchao/prototype/moe_training/blockwise_fp8/grouped_mm.py
iamzainhuda: Addressed in e22497b62. Added an explicit `KernelPreference.DEEPGEMM`; `AUTO` and `DEEPGEMM` now select DeepGEMM or fail when it is unsupported, while `EMULATED` is the only path to emulation

PR #4552, opened 2026-07-03
optim: cache qmap values as tuples

nascheme:
Free-threading audit finding: `get_qmap_signed`/`get_qmap_unsigned` in `subclass_4bit.py` and `subclass_8bit.py` cache a `list` via `lru_cache(maxsize=1)`. The list is returned by reference to every caller, so it's shared mutable state under free-threaded Python. Wrap the cached value in `tuple(...)` -- `torch.tensor(...)` accepts any iterable, so no call site changes are needed.


gau-nernst (approved):

vkuzo:
@claude review

PR #4553, opened 2026-07-03
kernel: avoid partial Triton lazy initialization

nascheme:
`blockwise_quantization.py` and `bsr_triton_ops.py` both used a module-level `_triton_initialized = False` flag set to `True` before the kernel handles were assigned. Under free-threaded Python, one thread can set the flag and yield during the slow `import triton`, while another thread observes the flag set but the kernel globals still `None`.

Replace the bool-flag pattern with a lock-guarded builder that publishes an immutable dataclass of the built handles only once construction completes, so callers either see nothing or the fully-built set.

### DIFF
--- a/torchao/kernel/blockwise_quantization.py
+++ b/torchao/kernel/blockwise_quantization.py
@@ -5,38 +5,43 @@
 # LICENSE file in the root directory of this source tree.
 
 import math
-from typing import Tuple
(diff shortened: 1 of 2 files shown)


vkuzo:
@claude review

vkuzo (approved):

vkuzo:
thank you!

PR #4554, opened 2026-07-03
kernel: synchronize autotuner cache updates

nascheme:
`get_best_config_fn` in `torchao/kernel/autotuner.py` had a lazy-init-then-read-modify-write on the module-level `BEST_CONFIGS` dict, followed by `_save_best_configs(BEST_CONFIGS)` pickling it directly. Under free-threaded Python this is a check-then-act race on init, plus `pickle.dump` can iterate the dict while another thread is still inserting into it.

Guard lazy init and the miss-path write with a lock, and pickle a snapshot taken under that same lock rather than the live dict.

### DIFF
--- a/torchao/kernel/autotuner.py
+++ b/torchao/kernel/autotuner.py
@@ -6,6 +6,7 @@
 import logging
 import os
 import pathlib
+import threading
 
 import torch
 import triton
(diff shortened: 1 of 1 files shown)


vkuzo:
@claude review

vkuzo (approved):

PR #4555, opened 2026-07-03
float8: snapshot FSDP precomputed scale

nascheme:
`fsdp_pre_all_gather` on `WeightWithDynamicFloat8CastTensor` checks `self._precomputed_scale is not None` and then reads `self._precomputed_scale` again to use it. `precompute_float8_dynamic_scale_for_fsdp` can reassign that attribute on the same shared instance between the check and the use, so under free-threaded Python the value used may differ from the one checked.

Read `self._precomputed_scale` into a local once and use that for both the check and the call.


vkuzo:
@claude review, and explain why this is needed for freethreaded Python

PR #4558, opened 2026-07-06
Expose FP8 blockwise grouped MM function

iamzainhuda:
## Summary
- Expose `_to_fp8_blockwise_then_scaled_grouped_mm` from the prototype MoE namespaces for direct TorchTitan integration without a tensor subclass or `quantize_`.
- Keep the established `_to_*` naming used by the existing grouped-MM entry points.

## Test Plan
`pytest -q test/prototype/moe_training/test_fp8_blockwise_grouped_mm.py::test_fp8_blockwise_grouped_mm_emulated_backend`

### DIFF
--- a/torchao/prototype/moe_training/__init__.py
+++ b/torchao/prototype/moe_training/__init__.py
@@ -6,6 +6,17 @@ from torchao.prototype.moe_training.mxfp8_grouped_mm import (
 )
 
 __all__ = [
+    "fp8_blockwise_grouped_mm",
     "_to_mxfp8_then_scaled_grouped_mm",
     "_to_fp8_rowwise_then_scaled_grouped_mm",
 ]
+
+
+def __getattr__(name: str):
+    if name == "fp8_blockwise_grouped_mm":
+        from torchao.prototype.moe_training.blockwise_fp8 import (
+            fp8_blockwise_grouped_mm,
+        )
+
+        return fp8_blockwise_grouped_mm
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
--- a/torchao/prototype/moe_training/__init__.py
+++ b/torchao/prototype/moe_training/__init__.py
@@ -6,17 +6,17 @@ from torchao.prototype.moe_training.mxfp8_grouped_mm import (
 )
 
 __all__ = [
-    "fp8_blockwise_grouped_mm",
+    "_to_fp8_blockwise_then_scaled_grouped_mm",
     "_to_mxfp8_then_scaled_grouped_mm",
     "_to_fp8_rowwise_then_scaled_grouped_mm",
 ]
 
 
 def __getattr__(name: str):
-    if name == "fp8_blockwise_grouped_mm":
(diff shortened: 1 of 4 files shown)


torchao/prototype/moe_training/blockwise_fp8/grouped_mm.py
```diff
@@ -29,7 +29,7 @@
 
 
 @conditional_nostrict_trace
-def _to_fp8_blockwise_then_scaled_grouped_mm(
+def fp8_blockwise_grouped_mm(
```
vkuzo: the old name sounds more correct than the new, why change?

torchao/prototype/moe_training/__init__.py
```diff
@@ -6,6 +9,7 @@
 )
 
 __all__ = [
+    "fp8_blockwise_grouped_mm",
```
vkuzo: I think all of these variables should be named consistently with each other, so something like `_to_fp8_blockwise_1x128_128x128_then_scaled_grouped_mm`

torchao/prototype/moe_training/blockwise_fp8/grouped_mm.py
iamzainhuda: i was still workshopping this change - i don't like how the main API export is a private helper name but i know in other places we have it like `_to_mxfp8_then_scaled_grouped_mm`

vkuzo: `to_fp8_blockwise_then_scaled_grouped_mm` is consistent with the others

if you want to remove the leading underscore, sounds reasonable to me.  For a prototype, I think leading underscore is fine.

iamzainhuda: got it - i left the naming unchanged. just adding the test now in this PR

torchao/prototype/moe_training/__init__.py
```diff
@@ -6,6 +6,17 @@
 )
 
 __all__ = [
+    "_to_fp8_blockwise_then_scaled_grouped_mm",
```
vkuzo, line 9: I think if we are exposing this as a public API, then it's time to refactor `torchao/prototype/moe_training/blockwise_fp8` and fold it into `torchao/prototype/moe_training`, to match existing code organization

iamzainhuda:
Closing as superseded. The underlying  already exists in #4551, so the package re-export and duplicate emulated-backend test are not needed. #4559 now imports the function from its existing module path and is based directly on #4551.

iamzainhuda:
Clarification: the underlying `_to_fp8_blockwise_then_scaled_grouped_mm` already exists in #4551.

torchao/prototype/moe_training/__init__.py
vkuzo: name is better, I'm ok to land this, but we should really remove the `blockwise_fp8` folder before we start marketing / using this too much from other repos

PR #4559, opened 2026-07-06
Make FP8 blockwise MoE DeepGEMM compile clean

iamzainhuda:
## Summary
- Add a precomputed FP8 blockwise grouped-MM plan path so DeepGEMM host metadata is prepared before fullgraph compile.
- Wrap DeepGEMM grouped forward and wgrad launches in custom ops and add fullgraph/fake-contract coverage.
- Bind a precomputed plan to the exact `offs` tensor used at runtime. Supplying a plan is an explicit DeepGEMM selection; normal `AUTO` selection still falls back to the emulated backend when DeepGEMM is unsupported.

## Compile issue
The DeepGEMM K-grouped wgrad launch needs per-expert group sizes as host-side Python data (`list[int]`) and `ks` metadata derived from `offs`. Building that metadata while tracing can require tensor-to-host extraction and break `torch.compile(..., fullgraph=True)`. Raw Python DeepGEMM extension calls also do not provide Dynamo/AOTAutograd with fake tensor shape behavior.

`prepare_fp8_blockwise_grouped_mm_plan(offs)` materializes static host metadata before compilation. The grouped forward and flat-operand wgrad launches are represented by `torch.library.custom_op` operators with fake implementations.

## Test Plan
- `pytest -q test/prototype/moe_training/test_fp8_blockwise_grouped_mm.py`
- `pytest -q test/prototype/moe_training/test_fp8_blockwise_deepgemm_backend.py`

### DIFF
--- a/torchao/prototype/blockwise_fp8_training/deepgemm_grouped_kernels.py
+++ b/torchao/prototype/blockwise_fp8_training/deepgemm_grouped_kernels.py
@@ -7,7 +7,6 @@ from __future__ import annotations
 
 import importlib
 from dataclasses import dataclass
-from enum import Enum
 from functools import lru_cache
 from typing import Optional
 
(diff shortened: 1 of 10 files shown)


torchao/prototype/blockwise_fp8_training/deepgemm_grouped_kernels.py
```diff
@@ -369,6 +369,60 @@ def prepare_deepgemm_wgrad_plan(
...
+    b_s: torch.Tensor,
+    grouped_layout: torch.Tensor,
+    out_dtype: torch.dtype,
+    block_size: int = 128,
+) -> torch.Tensor:
+    _require_deep_gemm()
+    capabilities = _get_deepgemm_capabilities()
+    grouped_gemm = capabilities.m_grouped_gemm
+    if grouped_gemm is None:
+        raise ImportError(
```
vkuzo, line 239: seems more like a runtime error

vkuzo:
> Add a precomputed FP8 blockwise grouped-MM plan path so DeepGEMM host metadata is prepared before fullgraph compile.

can you explain what is going on here in more detail?

torchao/prototype/blockwise_fp8_training/deepgemm_grouped_kernels.py
iamzainhuda: i was thinking here`ImportError` made more sense since this checks if the required symbol exists.

PR #4560, opened 2026-07-07
Implement `cat`, `transpose`, `mm` and `addmm` ops in Int8Tensor

draghan:
In commit https://github.com/pytorch/pytorch/commit/1b6442e463d53f7c7d5f714d00e1aacc7caf8531 PyTorch introduced optimization for XPU, which merges several `F.linear` calls that share the same input into one wide matmul, then slices the result back apart. 

This is spawning instances of `aten.cat.default`, `aten.addmm.default`, `aten.transpose.int` and `aten.mm.default` ops when TorchInductor is compiling certain models in `Int8DynamicActivationInt8WeightConfig(version=2)` quantization.

It was encountered during compilation of quantized Deepseek-OCR-2 model with the nightly torchAO and PyTorch+XPU:
`torch==2.14.0.dev20260707+xpu`
`torchao==0.18.0.dev20260707+xpu`

A standalone self-contained reproduction script:
```python
#!/usr/bin/env python
"""
Minimal reproducer for inductor `batch_linear_lhs` pre-grad fusion + torchao W8A8.
`BatchLinearLHSFusion` fuses >=5 `F.linear(x, w_i)` calls that share the same
2D input `x` into cat/transpose/mm/addmm on the concatenated weight. With W8A8
the weights are torchao `Int8Tensor`, so this drives those ops onto the subclass:
    torch.cat([w1..wn])  -> aten.cat.default
    torch.transpose      -> aten.transpose.int
    torch.mm             -> aten.mm.default      (no-bias group)
    torch.addmm          -> aten.addmm.default   (with-bias group)

The same quantized model + input is compiled and run on XPU twice: once with the
batch_linear_lhs optimization enabled and once with it disabled. The two outputs
must match, proving the fused int8 ops produce the same result as the unfused path.
"""

import torch
import importlib
import torch.nn as nn
from torchao.quantization import quantize_, Int8DynamicActivationInt8WeightConfig
import torch._inductor.config as inductor_config
from torch._dynamo.utils import counters

# disable caching so the fusion pass actually runs every time
inductor_config.force_disable_caches = True

# assert your torch version is indeed supporting this optimization
def _has_batch_linear_lhs_device_gating() -> bool:
    """True if the installed torch has the batch_linear_lhs device-gating
    introduced by pytorch commit 1b6442e."""

    try:
        pre_grad = importlib.import_module("torch._inductor.fx_passes.pre_grad")
    except ImportError:
        return False
    # `_resolve_pre_grad_fusion_options` is what makes {"devices": ["xpu"]} work.
    return hasattr(pre_grad, "_resolve_pre_grad_fusion_options")


assert _has_batch_linear_lhs_device_gating(), (
    "Installed torch lacks batch_linear_lhs device gating "
    "(pytorch commit 1b6442e) - this repro script has no use here"
)


# >= MIN_FUSE_SET_SIZE (5); linears must share the same 2D input to be fused.
NUM_LAYERS = 8
IN_FEATURES = 512
OUT_FEATURES = 512
BATCH = 32


class SharedInputLinears(nn.Module):
    """A with-bias group (-> addmm) and a bias-free group (-> mm), same input."""
    def __init__(self):
        super().__init__()
        self.with_bias = nn.ModuleList(
            nn.Linear(IN_FEATURES, OUT_FEATURES, bias=True) for _ in range(NUM_LAYERS)
        )
        self.no_bias = nn.ModuleList(
            nn.Linear(IN_FEATURES, OUT_FEATURES, bias=False) for _ in range(NUM_LAYERS)
        )
    def forward(self, x):
        acc = None
        for layer in self.with_bias:
            acc = layer(x) if acc is None else acc + layer(x)
        for layer in self.no_bias:
            acc = acc + layer(x)
        return acc


# Build the reference fp model and input once, then quantize on XPU. Both runs
# use the exact same quantized weights and input.
torch.manual_seed(0)
model = SharedInputLinears().to(device="xpu", dtype=torch.bfloat16).eval()
quantize_(model, Int8DynamicActivationInt8WeightConfig(version=2))
x = torch.randn(BATCH, IN_FEATURES, device="xpu", dtype=torch.bfloat16)


def run(enable_fusion):
    print(f"Running with batched fusion {'on' if enable_fusion else 'off'}")

    # batch_linear_lhs is gated on the input device type appearing in "devices".
    inductor_config.pre_grad_fusion_options = (
        {"batch_linear_lhs": {"devices": ["xpu"]}} if enable_fusion else {}
    )
    torch._dynamo.reset()
    counters.clear()
    compiled = torch.compile(model, backend="inductor")
    with torch.inference_mode():
        out = compiled(x)
    torch.xpu.synchronize()
    fired = counters["inductor"].get("batch_linear_lhs", 0)
    if enable_fusion:
        assert fired > 0, "batch_linear_lhs did not fire when enabled"
    else:
        assert fired == 0, "batch_linear_lhs fired when disabled"
    return out.float().cpu()


result_batched_fusion_off = run(enable_fusion=False)
result_batched_fusion_on = run(enable_fusion=True)

torch.testing.assert_close(result_batched_fusion_on, result_batched_fusion_off, rtol=2e-2, atol=2e-2)
print("[OK] batch_linear_lhs on XPU matches the unfused path.")

```


torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -519,5 +519,107 @@ def _(func, types, args, kwargs):
...
+    if first.zero_point is not None:
+        zero_point = func([t.zero_point for t in tensors], dim)
+
+    return Int8Tensor(
+        qdata,
+        scale,
+        list(first.block_size),
+        first.dtype,
+        zero_point=zero_point,
+        act_quant_scale=first.act_quant_scale,
```
vkuzo, line 639: we should also assert that the activation metadata matches, and throw an exception if it does not

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -519,5 +519,107 @@ def _(func, types, args, kwargs):
...
+    )
+    return return_and_correct_aliasing(func, args, kwargs, new_int8_tensor)
+
+
+@implements(aten.mm.default)
+def _(func, types, args, kwargs):
+    input_tensor, mat2 = args
+    if isinstance(mat2, Int8Tensor):
+        if isinstance(input_tensor, Int8Tensor):
+            input_tensor = input_tensor.dequantize()
```
vkuzo: this does not match the assumptions in torchao inference, we should not silently dequantize tensors.  Instead, this code should honor the configuration in `input_tensor` and `mat2`'s `act_quant_kwargs` metadata fields, and throw an exception if it cannot be honored.  Same for other callsites in this file.

liangan1: @draghan I have same comment with @vkuzo. The AO should use int8 gemm to do inference instead of use dequantize path. If the error is caused by https://github.com/pytorch/pytorch/commit/1b6442e463d53f7c7d5f714d00e1aacc7caf8531. The fix should be in the pytorch side and should not change the behavior of  AO.

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
```diff
@@ -519,5 +519,107 @@ def _(func, types, args, kwargs):
...
+    return func(input_tensor.dequantize(), mat2)
+
+
+@implements(aten.addmm.default)
+def _(func, types, args, kwargs):
+    bias, input_tensor, mat2 = args
+    beta = kwargs.get("beta", 1)
+    alpha = kwargs.get("alpha", 1)
+    if isinstance(mat2, Int8Tensor) and beta == 1 and alpha == 1:
+        if isinstance(input_tensor, Int8Tensor):
```
liangan1: Ditto

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
draghan: Thanks, done.

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
draghan: Done, I've routed the gemm through the `aten.linear.default` with assertions on quantizations of activation and weight matrices.

torchao/quantization/quantize_/workflows/int8/int8_tensor.py
draghan: Done, no silent dequantizations.

liangan1 (approved):

PR #4561, opened 2026-07-07
cpu: lock ukernel registration tables

nascheme:
Both `linear_8bit_act_xbit_weight/kernel_selector.h` and `groupwise_lowbit_weight_lut/kernel_selector.h` define a `UKernelConfigRegistrationTable` whose backing `std::unordered_map` is mutated on cache miss (`register_ukernel_config`) and read (`get_ukernel_config`) with no synchronization; the structs are used as function-local statics and the code was annotated "Not thread safe". Under free-threaded CPython the GIL no longer serializes callers, so two threads hitting a cold cache concurrently can race the map insertion (rehash during lookup, duplicate-key throw).

Add a `mutable std::mutex` to each struct and take a `std::lock_guard` in `register_ukernel_config` and `get_ukernel_config`. Contention is low since cache fills are rare, so a plain `std::mutex` is used rather than `std::shared_mutex`. The stale "Not thread safe" annotations are updated.

With the lock in place, two threads that both miss the cache can still hit the "already registered" throw in `register_ukernel_config`; a follow-up PR makes registration idempotent.

Assisted-by: Claude


vkuzo:
@claude review

PR #4562, opened 2026-07-07
cpu: make ukernel registration idempotent

nascheme:
Stacked on #4561 (the first commit here); only the second commit is new.

With the table mutex from #4561 in place, the call site in `select_ukernel_config` is still racy:

```cpp
auto ukernel = table.get_ukernel_config(header, uarch);  // lock A
if (ukernel.has_value()) return ukernel.value();
register_ukernel_config<weight_nbit>(table, format, uarch);  // lock B
```

Lock A and lock B are independent critical sections. Two threads can both observe the missing entry, both call `register_ukernel_config`, and the second thread's "already registered" throw fires as a TOCTOU false positive under free-threaded Python.

Make registration idempotent: first registration wins, duplicates are silent no-ops (`try_emplace` semantics). The registered config is deterministic for a given (header, uarch) key, so dropping the duplicate loses nothing. Single-threaded behavior is unchanged. Applies to both `linear_8bit_act_xbit_weight` and `groupwise_lowbit_weight_lut`, which have structurally identical tables.

Reproduced on x86 with a free-threaded interpreter: concurrent cold-cache calls to the 8-bit-act linear op raised "UKernelConfig is already registered for this format" before this change, pass after.

Assisted-by: Claude

### DIFF
--- a/torchao/csrc/cpu/shared_kernels/groupwise_lowbit_weight_lut/kernel_selector.h
+++ b/torchao/csrc/cpu/shared_kernels/groupwise_lowbit_weight_lut/kernel_selector.h
@@ -53,11 +53,13 @@ struct UKernelConfigRegistrationTable {
     auto header = format.to_packed_weights_header();
     auto key = make_key(header, uarch);
     std::lock_guard<std::mutex> guard(mu_);
-    if (registration_table_.find(key) != registration_table_.end()) {
-      throw std::runtime_error(
-          "UKernelConfig is already registered for this format");
-    }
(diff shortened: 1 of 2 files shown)


vkuzo:
@claude review

vkuzo:
lg, can we rebase to get around the merge conflicts?

nascheme:
I did a merge and resolved conflicts.  If you prefer a rebase then let me know, that's easy to do.  I think both ways result in the same single commit if you do "squash merge" in Github.

vkuzo:
@claude review

PR #4564, opened 2026-07-10
Revert "Revert "pt2e: preserve mutable buffer inputs during prepare""

tom-arm:
Reverts pytorch/ao#4531


andrewor14:
Made the fix to the internal code, going to try to land this from there, which should auto merge this PR

Issue #4541, opened 2026-06-28
Module-wise type hinting and docstring coverage audit for PT2E


GiGiKoneti:
### Description

This is an RFC proposing a systematic, module-wise type hinting and docstring coverage audit for the PyTorch 2 Export (PT2E) sub-modules. Currently, several core files bypass static type checking via `# mypy: allow-untyped-defs` and lack comprehensive docstrings. 

To keep code review manageable and avoid large PRs, we will break this down into directory-based, focused pull requests targeting the following 18 eligible files:

#### 1. Core PT2E Utility & Graph Processing (`torchao/quantization/pt2e/`)
* `torchao/quantization/pt2e/__init__.py`
* `torchao/quantization/pt2e/graph_utils.py`
* `torchao/quantization/pt2e/export_utils.py`
* `torchao/quantization/pt2e/utils.py`
* `torchao/quantization/pt2e/prepare.py`
* `torchao/quantization/pt2e/qat_utils.py`
* `torchao/quantization/pt2e/fake_quantize.py`
* `torchao/quantization/pt2e/observer.py`
* `torchao/quantization/pt2e/reference_representation_rewrite.py`

#### 2. Backend Quantizers & Passes (`torchao/quantization/pt2e/quantizer/`)
* `torchao/quantization/pt2e/quantizer/quantizer.py`
* `torchao/quantization/pt2e/quantizer/utils.py`
* `torchao/quantization/pt2e/quantizer/arm_inductor_quantizer.py`
* `torchao/quantization/pt2e/quantizer/x86_inductor_quantizer.py`
* `torchao/quantization/pt2e/quantizer/xpu_inductor_quantizer.py`
* `torchao/quantization/pt2e/quantizer/duplicate_dq_pass.py`
* `torchao/quantization/pt2e/quantizer/embedding_quantizer.py`
* `torchao/quantization/pt2e/quantizer/port_metadata_pass.py`

#### 3. Compiler Passes (`torchao/quantization/pt2e/inductor_passes/`)
* `torchao/quantization/pt2e/inductor_passes/x86.py`

---

#### Non-Eligible / Excluded Scope
* `torchao/quantization/pt2e/convert.py` (Too high complexity/dynamic FX transformations).
* `torchao/kernel/bsr_triton_ops.py` (Triton JIT compilation does not cleanly support static type checking).
[...]

PR #4565, opened 2026-07-11
docs/types(pt2e): annotate and document DuplicateDQPass (#4541)

Anai-Guo:
## Summary

Part of the PT2E type-hinting & docstring audit proposed in #4541. This PR
covers a single small, self-contained file:
`torchao/quantization/pt2e/quantizer/duplicate_dq_pass.py`.

- Remove the file-level `# mypy: allow-untyped-defs`.
- Add the one missing annotation: `_maybe_duplicate_dq(...) -> None` (it was the
  only untyped def; `DuplicateDQPass.call` and the nested `maybe_replace_node`
  were already annotated).
- Add module-, function-, class-, and method-level docstrings.

No behavior change — the executable lines are byte-for-byte identical; only the
mypy pragma, the one return annotation, and docstrings differ.

## Why this is safe

`allow-untyped-defs` only suppresses checking of *untyped* function bodies, so
`call` was already type-checked and passing. The only newly-checked body is
`_maybe_duplicate_dq`, whose types are straightforward (`GraphModule`/`Node`
in, `None` out); the existing `# type: ignore[assignment]` on the
`user.args`/`user.kwargs` rewrites is preserved.

`ruff check` passes and no line exceeds 88 columns.

### DIFF
--- a/torchao/quantization/pt2e/quantizer/duplicate_dq_pass.py
+++ b/torchao/quantization/pt2e/quantizer/duplicate_dq_pass.py
@@ -4,7 +4,15 @@
 # This source code is licensed under the BSD 3-Clause license found in the
 # LICENSE file in the root directory of this source tree.
 
-# mypy: allow-untyped-defs
+"""FX pass that duplicates shared dequantize nodes so each consumer owns one.
+
+When a single ``dequantize`` node feeds several downstream users, those users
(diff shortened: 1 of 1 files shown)


andrewor14:
@claude review please

andrewor14 (approved):

PR #4573, opened 2026-07-14
fix(int4/tile_packed): validate block_size against scale_and_zero to prevent OOB read

### LINKED ISSUE #4572: Bounds-check `block_size` against `scale_and_zero` in the int4 tile-packed load path (avoids an OOB read in `_weight_int4pack_mm`)
**What.** When an `Int4TilePackedTo4dTensor` is reconstructed from a checkpoint (via `__tensor_unflatten__`), the
stored `block_size` is pinned as a field with no validation against `scale_and_zero`. At inference,
`groupsize = block_size[-1]` is passed to `torch.ops.aten._weight_int4pack_mm(act, qdata, groupsize, scale_and_zero)`.
The aten op validates that `groupsize ∈ {32,64,128,256}` but not that it is *consistent* with the `scale_and_zero`
group extent, so a checkpoint declaring a valid-but-inconsistent smaller group size makes the tinygemm kernel
iterate more groups than the scale buffer holds and read out of bounds.

**Why it matters (robustness).** A malformed/adversarial checkpoint loaded through the safetensors path (which
torchao supports as a non-pickle format — `TorchAoHfQuantizer.is_serializable == True`, subclass flattened to
safetensors with `block_size` in metadata) can crash the process. The int8/intx sibling already re-derives
`n_blocks = qdata.shape // block_size` and asserts `scale.shape == n_blocks`
(`intx_unpacked_to_int8_tensor.py`); the int4 tile-packed path skips that check.

**Reproduction (torchao 0.17.0, torch 2.13.0+cu130, A100, compute-sanitizer memcheck).** Quantize a Linear with
`Int4WeightOnlyConfig(group_size=128, int4_packing_format=TILE_PACKED_TO_4D)`, serialize via
`flatten_tensor_state_dict`, edit the `block_size` in the metadata JSON from `[1,128]` to `[1,32]`,
`unflatten_tensor_state_dict` (restores it unchecked), and run forward:
```
Invalid __global__ read of size 4 bytes
    at ... tinygemm_m16n8k16_chunk_kernel<...BLayout_TC_int4<8,32>...>
    Access ... 12289 bytes after the nearest allocation of size 8192 bytes  (= scale_and_zero)
    Host Frame: at::native::_weight_int4pack_mm_cuda(...) in libtorch_cuda.so
```
Control (unedited `block_size=[1,128]`) runs clean, 0 errors.

**Proposed fix.** Validate `block_size` against `scale_and_zero` in the int4 tile-packed constructor, mirroring the
int8/intx sibling (`intx_unpacked_to_int8_tensor.py`, which already does `n_blocks = qdata.shape // block_size;
assert scale.shape == n_blocks`). Concretely, in `Int4TilePackedTo4dTensor.__init__`
(`torchao/quantization/quantize_/workflows/int4/int4_tile_packed_to_4d_tensor.py`):

```python
def __init__(self, qdata, scale_and_zero, block_size, shape, act_pre_scale=None):
    groupsize = block_size[-1]
    # scale_and_zero is laid out with one (scale, zero) pair per group along K.
    k = shape[-1]
    torch._check(
        groupsize in (32, 64, 128, 256),
        lambda: f"int4 tile-packed groupsize must be in (32,64,128,256), got {groupsize}",
    )
    torch._check(
        k % groupsize == 0 and scale_and_zero.shape[0] == k // groupsize,
        lambda: (
            f"block_size {block_size} is inconsistent with scale_and_zero "
            f"(expected {k // groupsize} groups along K, got {scale_and_zero.shape[0]})"
        ),
    )
    self.qdata = qdata
    self.scale_and_zero = scale_and_zero
    self.block_size = block_size
    self.act_pre_scale = act_pre_scale
```

(Adjust the exact `scale_and_zero` group dimension to the current layout.) Optionally also validate `groupsize`
against the `qScaleAndZeros` extent inside `_weight_int4pack_mm` so every caller of the core op is hardened, not
just torchao — happy to send that as a separate PyTorch-core PR.

Anai-Guo:
## Summary

Fixes #4572.

`Int4TilePackedTo4dTensor.__init__` stores `block_size` as a field with no validation against `scale_and_zero`. At inference, `groupsize = block_size[-1]` is passed to `torch.ops.aten._weight_int4pack_mm(...)`. The aten op only checks that `groupsize ∈ {32,64,128,256}`, not that it is *consistent* with the `scale_and_zero` group extent. So a checkpoint that declares a valid-but-inconsistent smaller group size — restored unchecked via `__tensor_unflatten__` from the non-pickle safetensors path — makes the tinygemm kernel iterate more groups than the scale buffer holds and read out of bounds (`Invalid __global__ read ... after the nearest allocation of size ... (= scale_and_zero)`).

The int8/intx sibling already guards this: `intx_unpacked_to_int8_tensor.py` re-derives `n_blocks = qdata.shape // block_size` and asserts `scale.shape == n_blocks`. This PR adds the equivalent check to the int4 tile-packed path.

## Detail

The check is added in `__init__` so it covers both fresh quantization and checkpoint reconstruction. Two notes on correctness:

- **Group-count dimension.** `pack_tinygemm_scales_and_zeros` lays `scale_and_zero` out as `(..., k // group_size, n, 2)`, transposing the group axis to `-3`. This holds for both the 2D layout (`[k//g, n, 2]`) and the MoE layout with a leading experts dim (`[experts, k//g, n, 2]`). The check therefore uses `scale_and_zero.shape[-3]`, **not** `shape[0]` — the latter would be the experts count on the MoE path and would false-trip.
- **Convention.** Uses `assert` to match the surrounding file and the int8 sibling.

The normal quantize path satisfies the invariant by construction, so no valid checkpoint is affected; only inconsistent `(block_size, scale_and_zero)` pairs are rejected with a clear message before they can reach the kernel.

### DIFF
--- a/test/quantization/quantize_/workflows/int4/test_int4_tile_packed_to_4d_tensor.py
+++ b/test/quantization/quantize_/workflows/int4/test_int4_tile_packed_to_4d_tensor.py
@@ -191,6 +191,41 @@ class TestInt4TilePackedTo4dTensor(TorchAOIntegrationTestCase):
             param.data.scale_and_zero.data_ptr() == param_data.scale_and_zero.data_ptr()
         )
 
+    def test_block_size_validated_against_scale_and_zero(self):
+        # Regression test for #4572: a block_size whose group size is smaller than
+        # the one scale_and_zero was packed with claims more groups along K than the
+        # tensor holds, which makes the tinygemm kernel read it out of bounds.
+        device = "cuda"
+        linear = torch.nn.Linear(
+            1024, 256, bias=False, dtype=torch.bfloat16, device=device
+        )
+        quantize_(linear, INT4_CONFIG)
+        qw = linear.weight
+        # scale_and_zero holds 1024 // 128 == 8 groups along K; a pinned group size
+        # of 32 would have the kernel walk 32 groups.
+        with self.assertRaisesRegex(AssertionError, "inconsistent with scale_and_zero"):
+            Int4TilePackedTo4dTensor(
+                qw.qdata,
+                qw.scale_and_zero,
+                [1, 32],
+                qw.shape,
+                act_pre_scale=qw.act_pre_scale,
+            )
+
+    def test_padded_in_features_accepted(self):
+        # from_hp pads K up to a multiple of 1024 before computing qparams but stores
+        # the unpadded shape, so scale_and_zero legitimately holds more groups than
+        # in_features // group_size. That must not trip the validation above.
+        device = "cuda"
+        linear = torch.nn.Linear(
+            1536, 256, bias=False, dtype=torch.bfloat16, device=device
+        )
+        quantize_(linear, INT4_CONFIG)
+        qw = linear.weight
+        # K padded 1536 -> 2048, so 2048 // 128 == 16 groups for in_features 1536.
+        self.assertEqual(qw.scale_and_zero.shape[-3], 16)
+        self.assertEqual(qw.shape[-1], 1536)
+
     def test_cant_initialize_in_cpu(self):
         config = INT4_CONFIG
         linear = torch.nn.Linear(128, 256, dtype=torch.bfloat16)
--- a/torchao/quantization/quantize_/workflows/int4/int4_tile_packed_to_4d_tensor.py
+++ b/torchao/quantization/quantize_/workflows/int4/int4_tile_packed_to_4d_tensor.py
@@ -87,6 +87,27 @@ class Int4TilePackedTo4dTensor(TorchAOBaseTensor):
         self.block_size = block_size
         self.act_pre_scale = act_pre_scale
 
+        # Validate block_size against scale_and_zero so a checkpoint that pins an
+        # inconsistent (but individually valid) block_size cannot make the tinygemm
+        # kernel iterate more groups than scale_and_zero holds and read out of bounds.
+        # This mirrors the int8/intx sibling (intx_unpacked_to_int8_tensor.py), which
+        # re-derives n_blocks from block_size and asserts the scale shape matches.
+        # scale_and_zero is packed by pack_tinygemm_scales_and_zeros to
+        # (..., k // group_size, n, 2), so the group count along K lives at dim -3
+        # for both the 2D and MoE (leading experts dim) layouts.
+        group_size = block_size[-1]
+        k = shape[-1]
+        assert k % group_size == 0, (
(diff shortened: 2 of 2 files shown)


vkuzo:
@claude review

Anai-Guo:
Thanks @vkuzo — the review caught a real bug and the CI red was exactly it. Pushed a fix, though I landed on a different invariant than the one suggested, because deriving the count from the padded K is still wrong on two paths.

**The blocker is confirmed.** `from_hp` pads K to `find_multiple(in_features, 1024)` before computing qparams but stores the unpadded `original_shape`, so `scale_and_zero` holds `padded_k // group_size` groups. My equality false-tripped on any `in_features` that isn't a multiple of 1024 — including `Linear(128, 256)` in this very test file, which is what `test_int4_weight_only_quant_subclass_api_grouped_5` was hitting.

**Why not `n_groups = find_multiple(shape[-1], 1024) // group_size`:**

- **K-slices to non-1024 boundaries.** The `aten.slice` impl sets `new_shape[1] = end - start` and slices `scale_and_zero` by group ratio, so the result carries exactly `(end - start) // group_size` groups. Rounding the stored K back up to 1024 expects more groups than are there — e.g. a 2048→768 shard has 6 groups but the check would demand 8. The existing `test_slice` only passes because it deliberately uses multiples of 1024 ("Note: we use multiples of 1024 for both in_features and out_features").
- **Empty K shards.** `block_size[dim] = min(block_size[dim], new_shape[dim])` makes `group_size` 0 when a rank gets an empty shard, so `k_padded // group_size` raises `ZeroDivisionError`.

**What I pushed instead** — assert the lower bound that actually rules out the over-read, i.e. the groups present must cover K:

```python
group_size = block_size[-1]
k = shape[-1]
n_groups = scale_and_zero.shape[-3]
assert n_groups * group_size >= k, ...
```

`n_groups * group_size >= k` is equivalent to `n_groups >= k / group_size`, which is precisely the condition that the kernel's group walk stays in bounds — so it still catches #4572 — while padded weights (more groups than K implies), exact K-slices, and empty shards (`0 * 0 >= 0`) all satisfy it. It's also division-free, so the empty-shard case can't raise.

Also added two tests per your suggestion: `test_padded_in_features_accepted` pins the `in_features=1536` → 16-group padding case that broke this, and `test_block_size_validated_against_scale_and_zero` pins the inconsistent `block_size` the guard exists for.

PR #4576, opened 2026-07-16
[xpu][test] Relax test_bf16_stochastic_round tolerance

karol-brejna-i:
# Goal of the PR

This PR:
- enables `test/test_low_bit_optim.py` in XPU CI (`.github/scripts/ci_test_xpu.sh` modification)
- fixes `test_bf16_stochastic_round`, whose tolerance was tighter than the test's own
sampling noise (causing CI to fail).

The tolerance is changed from `atol=3e-5, rtol=3e-5` to `atol=1e-5, rtol=2e-4`.
Although the problem was detected on PVC 1100 card, **it is not really device specific** - it is reproducible on CPU/CUDA (different seed or sample size). 

----------
### This is not a hardware issue

- **Reproducible on CPU.** With the original tolerance, CPU fails on ~2% of random draws (7 out of 300 seeds). The same is true on CUDA and on every XPU device tested. 
- **The kernel is correct.** Measured bias is consistent with zero on CPU, BMG and PVC alike. There is no numerical defect to fix.
- It **looks** device-specific** because `TestCase.setUp` calls `torch.manual_seed(1234)`. Each (device, torch build) combination therefore repeats one fixed draw. A device whose draw happens to fall outside the tolerance fails on every run; a device whose draw falls inside passes on every run.

### Why the change is needed

The test averages `N = 100_000` stochastic roundings and compares the mean back to the original fp32 value:

```python
x = torch.rand(32, device=device) * 100
x_rep = x.view(-1, 1).repeat(1, 100_000)
x_rep_bf16 = func(x_rep)
torch.testing.assert_close(x_rep_bf16.float().mean(1), x, atol=3e-5, rtol=3e-5)
```

Stochastic rounding is unbiased, so this mean converges to `x` — but only up to the **standard error of the mean**. What the assertion sees is sampling noise, not a rounding error, and the old tolerance sat just ~2.5 sigma above that noise. A few percent of random draws land outside it, on any backend.

### How the new value was chosen

For `x` in `[2^e, 2^(e+1))` the bf16 spacing is `ulp = 2^(e-7)`. One stochastic rounding has variance `ulp^2 * p(1-p) <= ulp^2/4`, so the mean of `N` of them has

```
sigma_rel = (ulp/x) / (2*sqrt(N)) <= 2^-7 / (2*sqrt(100_000)) = 1.24e-5
```

This is **scale-invariant**, so `rtol` is what matters here and `atol` only guards values near zero. The bound is worst when `x` sits just above a power of two, where `ulp` doubles but `x` does not — and indeed the observed worst element is `x = 66.761`, just above 64.

`rtol = 2e-4` is **~16 sigma**, which is far enough out that a false failure is not a practical concern, and is still ~20-40x tighter than the error any non-stochastic implementation would produce (`2^-8 = 3.9e-3` for round-to-nearest, `2^-7 = 7.8e-3` for truncation). Verified that the test still fails if the rounding is replaced by truncation, by round-to-nearest-even, or by stochastic rounding with the probability skewed by 10%.


xiaowangintel:
## Summary
This failure is not device-specific. After changing the test seed configuration, the same failure was reproduced on other hardware platforms as well：

| Scenario | Seed | Result |
|---|---|---|
| XPU (original CI failure) | 1234 (default) | ❌ | 
| CPU, `manual_seed(42)` | 42 | ❌ | 
| CUDA, `manual_seed(16)` | 16 | ❌ |


The PyTorch tolerance used for FP32 → BF16 conversion is:
rtol = 0.016 (1.6e-2)
atol = 1e-5
And can be used to justify adopting the same tolerance in TorchAO.

karol-brejna-i:
> ## Summary
> This failure is not device-specific. After changing the test seed configuration, the same failure was reproduced on other hardware platforms as well：
> 
> Scenario	Seed	Result
> XPU (original CI failure)	1234 (default)	❌
> CPU, `manual_seed(42)`	42	❌
> CUDA, `manual_seed(16)`	16	❌
> The PyTorch tolerance used for FP32 → BF16 conversion is: rtol = 0.016 (1.6e-2) atol = 1e-5 And can be used to justify adopting the same tolerance in TorchAO. I verified this on XPU, CPU, and CUDA. This test passed.

Thanks — your CPU (seed 42) and CUDA (seed 16) reproductions confirm this isn't device-specific. That matches what I measured, and I've reworked the PR description around it.



On `rtol=1.6e-2` though, I'd push back: **it makes the test ineffective.**

That value is PyTorch's tolerance for a *single* fp32 → bf16 rounding. But this test averages 100,000 stochastic roundings, and the mean converges far tighter than one rounding does — the sampling noise here is only ~1.2e-5 relative. So `1.6e-2` is ~1300x larger than anything the test can actually produce.

The practical consequence: `1.6e-2` is above the error of *every* bf16 rounding mode (truncation is at most 7.8e-3, round-to-nearest 3.9e-3). So the test passes even if stochastic rounding is removed entirely. I checked — replacing the kernel with plain truncation:

```python
def broken_truncate(x):
    return (x.view(torch.int32) & 0xFFFF0000).view(torch.float32).bfloat16()
```

still passes at `1.6e-2` on 200/200 seeds. That's why it passed on XPU/CPU/CUDA for you — but a kernel with SR deleted would pass there too, so it doesn't tell us the kernel is correct.

I'd suggest **`rtol=2e-4`** instead (keeping your `atol=1e-5`). It's ~16 sigma above the noise, so no flakiness — 0/100 seeds fail on PVC 1100 — but still tight enough to catch truncation, round-to-nearest, or a 10%-skewed SR probability.

PR #4583, opened 2026-07-20
Add QAT flow for CoreML codebook quantization

andrewor14:
**Summary:** Adds a QAT flow for CoreML-style codebook (lookup-table) quantization that mirrors the numerics of the PTQ config CodebookWeightOnlyConfig and plugs into the standard QATConfig prepare/convert mechanism. On a refresh step the fake quantization matches PTQ.

- CodebookFakeQuantizeConfig / CodebookFakeQuantizer, wired into the QAT config-inference and from_config paths; a straight-through estimator lets gradients flow to the underlying weight.
- The codebook (k-means centroids) is expensive and non-differentiable, so it is refreshed every `refresh_interval` steps (default 100); between refreshes only the cheap nearest-centroid assignment is recomputed. `refresh_interval=1` gives exact PTQ numerics every step.
- The cached codebook is a persistent buffer, so it is saved in the prepared checkpoint. Convert reuses the persisted codebook via a new `custom_codebook` path and runs no new k-means, so a model can be trained, checkpointed, then loaded and converted later.

Example usage:

```python
import torch
from torchao.quantization.qat import QATConfig
from torchao.prototype.quantization.codebook_coreml import CodebookWeightOnlyConfig

base_config = CodebookWeightOnlyConfig(dtype=torch.uint4, block_size=[-1, 16])

# training session (needs coremltools)
quantize_(model, QATConfig(base_config, step="prepare"))
train_loop(model)                                  # populates the codebook
torch.save(model.state_dict(), "prepared.pt")      # codebook saved in state_dict

# later session
model = build_model()
quantize_(model, QATConfig(base_config, step="prepare"))    # codebook starts empty
model.load_state_dict(torch.load("prepared.pt"))            # restores the codebook
quantize_(model, QATConfig(base_config, step="convert"))    # reuses it, no k-means
output = model(example_input)
```

**Test Plan:**
```
pytest test/prototype/test_codebook_coreml_qat.py
```

### DIFF
--- a/test/prototype/test_codebook_coreml_qat.py
+++ b/test/prototype/test_codebook_coreml_qat.py
@@ -1,0 +1,313 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+import copy
+import unittest
+
+import torch
+
+from torchao.prototype.qat import (
+    CodebookFakeQuantizeConfig,
+    CodebookFakeQuantizer,
+)
+from torchao.prototype.quantization.codebook_coreml import (
+    CodebookQuantizedTensor,
+    CodebookWeightOnlyConfig,
+)
+from torchao.quantization import quantize_
+from torchao.quantization.qat import QATConfig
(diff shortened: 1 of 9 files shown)


andrewor14:
@claude review

Issue #4582, opened 2026-07-20
Rowwise fp8 (Float8DynamicActivationFloat8Weight, PerRow) fails with CUBLAS_STATUS_NOT_SUPPORTED for large-K + small-M GEMMs on H100


sayakpaul:
## Summary

`Float8DynamicActivationFloat8WeightConfig(granularity=PerRow())` produces a `Linear` that raises `CUDA error: CUBLAS_STATUS_NOT_SUPPORTED` at inference for **large-K + small-M** GEMMs on H100. Concretely, a `Linear(9216 -> 3072)` fails when the activation's leading dim `M <= 256` (e.g. `M=127`), while:

- the **same** rowwise Linear works for `M >= 257`,
- a smaller-K Linear (`3072 -> 3072`) works even at `M=127`,
- **`PerTensor` granularity works** at the same failing shape (workaround).

So the failure is specific to the **rowwise (axiswise) fp8 `_scaled_mm` path** for large-K + small-M problems. `quantize_` happily quantizes the layer, but it then crashes on perfectly valid inputs.

This showed up in a real model: the audio feed-forward `linear_out` is `Linear(9216 -> 3072)` and the audio stream has 127 tokens, so `M=127` — every audio/video inference crashes, while the image path (large `M`) is fine.

## Repro (torch + torchao only)

```python
import torch
from torchao.quantization import (
    Float8DynamicActivationFloat8WeightConfig, PerRow, PerTensor, quantize_,
)

def try_linear(K, N, M, granularity, tag):
    lin = torch.nn.Linear(K, N, bias=False).to("cuda", torch.bfloat16)
    quantize_(lin, Float8DynamicActivationFloat8WeightConfig(granularity=granularity))
    x = torch.randn(M, K, device="cuda", dtype=torch.bfloat16)
    try:
        y = lin(x); torch.cuda.synchronize()
        print(f"[{tag}] Linear({K}->{N}) M={M}: OK -> {tuple(y.shape)}")
    except Exception as e:
        print(f"[{tag}] Linear({K}->{N}) M={M}: FAIL -> {str(e).splitlines()[0]}")

try_linear(9216, 3072, 127, PerRow(),   "repro    ")  # FAIL
try_linear(9216, 3072, 256, PerRow(),   "M<=256   ")  # FAIL
try_linear(9216, 3072, 257, PerRow(),   "M>=257   ")  # OK
try_linear(9216, 3072, 512, PerRow(),   "large-M  ")  # OK
[...]

PR #4588, opened 2026-07-21
add warnings for float8 rowwise inference on CUDA 12.9

vkuzo:
Summary:

Adds warning for the case hit in https://github.com/pytorch/ao/issues/4582

Note that:
* things are fixed with a CUDA 13.0 upgrade
* the shapes on which this is broken in CUDA 12.9 are not known (we can probe, but we can't know for sure without source code access)

So, we just warn and nudge the user to upgrade their CUDA

Test Plan: tested with Claude - warns on CUDA 12.9, does not warn on CUDA 12.6 or 13.0

### DIFF
--- a/torchao/quantization/quant_api.py
+++ b/torchao/quantization/quant_api.py
@@ -1282,6 +1282,44 @@ def _float8_dynamic_activation_float8_weight_quantize_tensor(weight, config):
         return quantized_weight
 
 
+def _maybe_warn_rowwise_fp8_cuda_12_9(
+    config: Float8DynamicActivationFloat8WeightConfig,
+) -> None:
+    if not torch.cuda.is_available():
+        return
+    if not torch.version.cuda.startswith("12.9"):
+        return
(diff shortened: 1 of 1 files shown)


drisspg (approved):

torchao/quantization/quant_api.py
```diff
@@ -1282,13 +1282,51 @@ def _float8_dynamic_activation_float8_weight_quantize_tensor(weight, config):
         return quantized_weight
 
 
+def _maybe_warn_rowwise_fp8_cuda_12_9(
+    config: Float8DynamicActivationFloat8WeightConfig,
+) -> None:
+    cuda_version = torch.version.cuda
+    if cuda_version is None or not cuda_version.startswith("12.9"):
```
sayakpaul: Should there be a validation around `if not torch.cuda.is_available()`?

sayakpaul (approved):

torchao/quantization/quant_api.py
vkuzo: fixed

PR #4591, opened 2026-07-22
fix(float8nocompile): correct swapped bounds-mask axes in _to_fp8_row_major_t (#4584)

### LINKED ISSUE #4584: float8nocompile `_to_fp8_row_major_t`: transposed-store bounds mask has its axes swapped
### Description

The transposed store address is
https://github.com/pytorch/ao/blob/5f17fde9fc575435a4211fbe9c8ea4c0fba012d2/torchao/prototype/float8nocompile/kernels/fp8_dynamic_tensorwise.py#L147-L150
so `block_col_offs` indexes the output **row**. 

But the bounds mask is 
https://github.com/pytorch/ao/blob/5f17fde9fc575435a4211fbe9c8ea4c0fba012d2/torchao/prototype/float8nocompile/kernels/fp8_dynamic_tensorwise.py#L151-L153
`block_row_offs` checked against `output_num_rows`, i.e. axes swapped relative to the address. 
For rectangular / non-tile-aligned inputs whole tiles are masked off (leaving torch.empty garbage) or written out of bounds

The test below fails (`torchao/prototype/float8nocompile/kernels/fp8_dynamic_tensorwise_test.py`)

### Steps to reproduce

```python
@pytest.mark.parametrize(
    "algo",
    [KernelAlgorithm.REDUCTION, KernelAlgorithm.ATOMIC_MAX],
)
@pytest.mark.parametrize(
    "input_shape",
    [(2, 4), (32, 16), (512, 512)],
)
def test_fp8_hp_to_fp8_row_major_t(input_shape: tuple[int, int], algo: KernelAlgorithm):
    assert torch.cuda.is_available()
    device = "cuda"
    # input_bf16 = torch.tensor(
    #     [[1, 2, 3], [4, 5, 6]], dtype=torch.bfloat16, device=device
    # )
    input_bf16 = torch.randn((32, 256), dtype=torch.bfloat16, device=device)
    x_bf16 = input_bf16.clone().detach().to(device)
    y_bf16 = input_bf16.clone().detach().to(device)

    # production implementation
    x_fp8_row_major = hp_tensor_to_float8_dynamic(
        x_bf16,
        torch.float8_e4m3fn,
        LinearMMConfig(),
    )
    x_fp8_row_major_t = x_fp8_row_major.t().contiguous()

    # float8nocompile triton implementation
    y_fp8_row_major_t = hp_to_fp8_row_major_t(
        y_bf16,
        torch.float8_e4m3fn,
        LinearMMConfig(),
        algo=algo,
    )

    # check scales
    assert torch.eq(x_fp8_row_major_t._scale, y_fp8_row_major_t._scale)

    # check data
    assert torch.all(torch.eq(x_fp8_row_major_t._data, y_fp8_row_major_t._data))

    # check shapes
    assert x_fp8_row_major_t.shape == y_fp8_row_major_t.shape

    # check strides
    assert x_fp8_row_major_t.stride() == y_fp8_row_major_t.stride()

    # check memory layout
    assert is_row_major(x_fp8_row_major_t.stride())
    assert is_row_major(y_fp8_row_major_t.stride())

    # check underlying memory layout
    assert (
        x_fp8_row_major_t._data.storage().tolist()
        == y_fp8_row_major_t._data.storage().tolist()
    )

    # assert that error is raised when input tensor is not contiguous
    with pytest.raises(AssertionError, match="tensor must be contiguous"):
        hp_to_fp8_row_major(
            y_bf16.t(),  # transpose so tensor memory layout is no longer contiguous
            torch.float8_e4m3fn,
            LinearMMConfig(),
        )
```

### Environment
- torch 2.13.0+cu126
- torchao 0.18.0+gitb7f01f18
- H100
- Ubuntu 22.04.4 LTS

Anai-Guo:
## Summary

`_to_fp8_row_major_t` in `torchao/prototype/float8nocompile/kernels/fp8_dynamic_tensorwise.py` builds the transposed store address so that the output **row** is indexed by `block_col_offs` and the output **column** by `block_row_offs`:

```python
out_offs = (
    block_col_offs[:, None] * output_stride_row
    + block_row_offs[None, :] * output_stride_col
)
```

but the bounds mask checks the opposite axes:

```python
out_mask = (block_row_offs[:, None] < output_num_rows) & (
    block_col_offs[None, :] < output_num_cols
)
```

so the mask axes are swapped relative to `out_offs`. For square, tile-aligned inputs the two coincide, but for **rectangular / non-tile-aligned** shapes whole valid output tiles get masked off (leaving the `torch.empty` garbage in the result) or are written out of bounds.

## Fix

- Swap the mask axes so `block_col_offs` is bounded by `output_num_rows` and `block_row_offs` by `output_num_cols`, matching `out_offs`.
- Make `test_fp8_hp_to_fp8_row_major_t` actually use its parametrized `input_shape` (as the sibling `test_fp8_hp_to_fp8_row_major` already does) instead of a fixed `2x3` tensor. The fixed tensor fit inside a single tile and never hit the boundary, which is why the parametrized `(32, 16)` / `(512, 512)` cases were not exercising the transposed store and the bug slipped through.

Fixes #4584.

### DIFF
--- a/torchao/prototype/float8nocompile/kernels/fp8_dynamic_tensorwise_test.py
+++ b/torchao/prototype/float8nocompile/kernels/fp8_dynamic_tensorwise_test.py
@@ -93,9 +93,11 @@ def test_fp8_hp_to_fp8_row_major(input_shape: tuple[int, int], algo: KernelAlgor
 def test_fp8_hp_to_fp8_row_major_t(input_shape: tuple[int, int], algo: KernelAlgorithm):
     assert torch.cuda.is_available()
     device = "cuda"
-    input_bf16 = torch.tensor(
-        [[1, 2, 3], [4, 5, 6]], dtype=torch.bfloat16, device=device
-    )
+    # Use the parametrized shape (matching test_fp8_hp_to_fp8_row_major) so the
+    # rectangular / non-tile-aligned cases actually exercise the transposed
+    # store's bounds mask. A fixed tiny tensor fit in a single tile and never
+    # hit the boundary, hiding the swapped-axis mask bug.
+    input_bf16 = torch.randn(input_shape, dtype=torch.bfloat16, device=device)
     x_bf16 = input_bf16.clone().detach().to(device)
     y_bf16 = input_bf16.clone().detach().to(device)
 
--- a/torchao/prototype/float8nocompile/kernels/fp8_dynamic_tensorwise_test.py
+++ b/torchao/prototype/float8nocompile/kernels/fp8_dynamic_tensorwise_test.py
@@ -88,7 +88,11 @@ def test_fp8_hp_to_fp8_row_major(input_shape: tuple[int, int], algo: KernelAlgor
 )
 @pytest.mark.parametrize(
     "input_shape",
-    [(2, 4), (32, 16), (512, 512)],
+    # (129, 16) is deliberately not a multiple of the 128-wide block: it makes the
+    # row and column block starts differ, which is exactly what the swapped-axis
+    # bounds mask got wrong. The other shapes are single-block or square and pass
+    # under both the old and the new mask.
+    [(2, 4), (32, 16), (129, 16), (512, 512)],
 )
 def test_fp8_hp_to_fp8_row_major_t(input_shape: tuple[int, int], algo: KernelAlgorithm):
     assert torch.cuda.is_available()
--- a/torchao/prototype/float8nocompile/kernels/fp8_dynamic_tensorwise.py
+++ b/torchao/prototype/float8nocompile/kernels/fp8_dynamic_tensorwise.py
@@ -148,8 +148,14 @@ def _to_fp8_row_major_t(
         block_col_offs[:, None] * output_stride_row
         + block_row_offs[None, :] * output_stride_col
     )
-    out_mask = (block_row_offs[:, None] < output_num_rows) & (
-        block_col_offs[None, :] < output_num_cols
+    # The transposed store indexes the output row by block_col_offs and the
+    # output column by block_row_offs, so the bounds mask must be checked
+    # against the same axes. Using block_row_offs against output_num_rows here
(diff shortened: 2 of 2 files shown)


torchao/prototype/float8nocompile/kernels/fp8_dynamic_tensorwise_test.py
```diff
@@ -93,9 +93,11 @@ def test_fp8_hp_to_fp8_row_major(input_shape: tuple[int, int], algo: KernelAlgor
...
     assert torch.cuda.is_available()
     device = "cuda"
-    input_bf16 = torch.tensor(
-        [[1, 2, 3], [4, 5, 6]], dtype=torch.bfloat16, device=device
-    )
+    # Use the parametrized shape (matching test_fp8_hp_to_fp8_row_major) so the
+    # rectangular / non-tile-aligned cases actually exercise the transposed
+    # store's bounds mask. A fixed tiny tensor fit in a single tile and never
+    # hit the boundary, hiding the swapped-axis mask bug.
+    input_bf16 = torch.randn(input_shape, dtype=torch.bfloat16, device=device)
```
Sanjays2402, line 104: None of these shapes actually regresses the mask fix: `(2, 4)` and `(32, 16)` launch one program on each axis for every candidate config, while `(512, 512)` is square and tile-aligned, so the old and new masks are numerically identical. Could we include `(129, 16)`? That forces unequal block starts even with the 128x128 config and fails against the old mask.

Anai-Guo:
Thanks — good catch, you're right that none of the original shapes actually regress under the old mask.

Added `(129, 16)` in 32598e8. With `BLOCK_SIZE = 128` that shape gives `num_pid_m = 2`, `num_pid_n = 1`, so the second row-block starts at 128 while the column-block start stays at 0 — the two axes are no longer interchangeable and the swapped mask writes out of the tensor's valid region. The other three shapes are either single-block on both axes or square and tile-aligned, so old and new masks coincide there.

PR #4593, opened 2026-07-23
Restore fp32 matmul precision in TestQuantFlow (#4593)

georgehong:
Summary:

pytorch/pytorch #185236 breaks some tests due to detecting fp32 precision flag leaks, so some forward fixing is required for tests.

Differential Revision: D113332658

### DIFF
--- a/test/quantization/test_quant_api.py
+++ b/test/quantization/test_quant_api.py
@@ -158,6 +158,17 @@ class TestQuantFlow(TestCase):
         ["xpu"] if torch.xpu.is_available() else []
     )
 
+    def setUp(self):
+        super().setUp()
+        # quantize_() sets a global float32 matmul precision; snapshot fp32_precision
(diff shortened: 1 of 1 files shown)


test/quantization/test_quant_api.py
```diff
@@ -158,6 +158,17 @@ class TestQuantFlow(TestCase):
         ["xpu"] if torch.xpu.is_available() else []
     )
 
+    def setUp(self):
+        super().setUp()
+        # quantize_() sets a global float32 matmul precision; snapshot fp32_precision
```
vkuzo, line 163: wait wut

vkuzo:
@claude "# quantize_() sets a global float32 matmul precision; snapshot fp32_precision" - where specifically does this happen

vkuzo (approved):

vkuzo:
@georgehong thank you! I'll let you land to not mess up your diff train, lmk if anything else is needed!

PR #4487, opened 2026-06-15
[nvfp4_training] Add CuteDSL NVFP4 RHT kernels

tbqh:
Added a CuteDSL (`nvidia-cutlass-dsl`) backend for the NVFP4 training RHT amax and quantization kernels, as a faster alternative to the Triton kernels from #4369.

### Benchmarks

Run environment: NVIDIA GB200, PyTorch 2.12.0a0, Triton 3.7.0, nvidia-cutlass-dsl.

**Hadamard amax** (`cutedsl_rht_amax` vs `triton_rht_amax`):

| Model | Shape | M | N | cutedsl_us | triton_us | speedup |
|---|---|---:|---:|---:|---:|---:|
| Llama 3 8B  | hidden-state input | 2048 | 4096  | 11.36 | 17.11 | 1.51x |
| Llama 3 8B  | mlp.down input     | 2048 | 14336 | 20.96 | 27.55 | 1.31x |
| Llama 3 70B | hidden-state input | 2048 | 8192  | 16.17 | 21.55 | 1.33x |
| Llama 3 70B | mlp.down input     | 2048 | 28672 | 39.58 | 43.20 | 1.09x |

**Hadamard quantize row+col** (`cutedsl_rht_quantize_row_col` vs `triton_rht_quantize_row_col`, RTNE):

| Model | Shape | M | N | cutedsl_us | triton_us | speedup |
|---|---|---:|---:|---:|---:|---:|
| Llama 3 8B  | hidden-state input | 2048 | 4096  | 13.89 | 24.36 | 1.75x |
| Llama 3 8B  | mlp.down input     | 2048 | 14336 | 34.91 | 60.68 | 1.74x |
| Llama 3 70B | hidden-state input | 2048 | 8192  | 23.73 | 37.21 | 1.57x |
| Llama 3 70B | mlp.down input     | 2048 | 28672 | 67.94 | 113.74| 1.67x |

**2D weight quantize** (`cutedsl_weight_quantize_2d` vs `triton_weight_quantize_2d`, no RHT):

| Model | Weight | M (out) | N (in) | cutedsl_us | triton_us | speedup |
|---|---|---:|---:|---:|---:|---:|
| Llama 3 8B  | mlp.gate/up | 14336 | 4096  | 65.98  | 248.33 | 3.76x |
| Llama 3 8B  | mlp.down    | 4096  | 14336 | 66.29  | 248.53 | 3.75x |
[...]

PR #4598, opened 2026-07-24
[nvfp4_training] Support NVFP4 GroupedExperts on Blackwell

rdspring1:
# Summary

This PR adds a **differentiable NVFP4 grouped GEMM** stack for Mixture-of-Experts training on Blackwell (SM100+). It provides a single autograd entry point, `_to_nvfp4_then_scaled_grouped_mm`, whose forward, dgrad, and wgrad each fuse a per-group **randomized Hadamard transform (RHT)** with **NVFP4** quantization (E2M1 4-bit codes + per-16 E4M3 block scales + a per-tensor scale) and feed `torch.nn.functional.scaled_grouped_mm`.

Three grouped Triton kernels do the quantization work. All group-aware via cumulative token-end offsets (jagged expert groups, 128-row aligned). `_NVFP4GroupedMM` (`nvfp4_training/nvfp4_grouped_mm.py`) wires the kernels into an `autograd.Function` with same training recipe as Linear kernels.

- `triton_group_rht_amax` — per-group row/col amax after RHT.
- `triton_group_rht_quantize_row_col` — fused rowwise + columnwise NVFP4 activation quantization.
- `triton_group_weight_quantize_2d` — per-expert 2D NVFP4 weight quantization (W and Wᵀ).
- `triton_group_weight_amax` — per-group amax

# Kernels

## 1. `triton_group_rht_amax` (`group_hadamard_amax_triton.py`)

Computes, per expert group `g`, `col_amax[g] = max|RHT(Aᵀ)|` and `row_amax[g] = max|A|` without materializing the transformed output. These become the per-tensor NVFP4 scales for the row and col quantizer.

- **Two Triton kernels.** 
- A **persistent per-group-CTA** kernel (warp-specialized, TMA descriptor for the 16×16 Hadamard matrix, elementwise cumulative max with a single `atomic_max` per CTA) recovers single-tensor bandwidth for large groups.
- A **tiled** kernel handles small groups. The wrapper picks between them by average group size and SM occupancy.
- Group membership via binary search over cumulative token-end offsets
- 128-row-aligned boundaries so a tile never straddles two experts.
- Autotune key `["N"]` (per-tile occupancy tracks `N`, not total token count — keeps CUDA-graph capture warm under variable token counts).

<details>
<summary>triton_group_rht_amax — DSV3 671B (E=128), effective memory bandwidth</summary>

| Projection | M | N | time (µs) | GB/s | % peak |
|-----------|---|---|-----------|------|--------|
| gate/up (w1/w3) | 2048 | 7168 | 2068.51 | 1816.81 | 22.9% |
| down (w2) | 7168 | 2048 | 2064.38 | 1820.44 | 23.0% |
</details>

## 2. `triton_group_rht_quantize_row_col` (`group_rht_quantize_row_col_triton.py`)

Fuses two NVFP4 quantizations in one 128×128 tile per CTA:
- **Rowwise** — raw `A`, scaled by `row_amax[g]` → row FP4 codes + swizzled FP8 scales (forward-GEMM LHS).
- **Columnwise** — RHT of `Aᵀ`, scaled by `col_amax[g]` → col FP4 codes + swizzled FP8 scales (saved for wgrad).
- Optional **Philox stochastic rounding** driven by a caller-owned 4-element `rng_state` (col/row seed+offset), advanced by the caller so it stays CUDA-graph-safe across replays.
- `num_warps` pinned at 4 (register-heavy body over-subscribes at 8); autotune key `["N", "STOCHASTIC_ROUNDING"]`.

<details>
<summary>triton_group_rht_quantize_row_col — DSV3 671B (E=128), effective memory bandwidth</summary>

Device peak memory bandwidth: 7928.1 GB/s (GB200)

| Projection | M | N | Rounding | time (µs) | GB/s | % peak |
|-----------|---|---|----------|-----------|------|--------|
| gate/up (w1/w3) | 2048 | 7168 | rtne | 2635.78 | 2227.82 | 28.1% |
| gate/up (w1/w3) | 2048 | 7168 | rs   | 5411.82 | 1085.04 | 13.7% |
| down (w2) | 7168 | 2048 | rtne | 2614.05 | 2246.33 | 28.3% |
| down (w2) | 7168 | 2048 | rs   | 5400.75 | 1087.26 | 13.7% |

`rtne` = round-to-nearest-even (forward path); `rs` = stochastic rounding (backward path). RS roughly halves bandwidth due to the per-element Philox draws.
</details>

## 3. `triton_group_weight_quantize_2d` (`group_quantize_2d_triton.py`)

- Per-expert 16×16-block NVFP4 quantization of both `W` and `Wᵀ` (no RHT), scaled by a per-expert amax. 
- Grid `(M//128, N//128, E)`, one tile per CTA, with **int64** per-expert base pointers so large `E·M·N` weights address correctly. 
- Autotune key `["M", "N"]` (weight dims are fixed at model-definition time, so keying on them is safe and re-tunes nothing per step).

<details>
<summary>triton_group_weight_quantize_2d — DSV3 671B (E=128), effective memory bandwidth</summary>

| Projection | M | N | time (µs) | GB/s | % peak |
|-----------|---|---|-----------|------|--------|
| gate/up (w1/w3) | 2048 | 7168 | 2628.61 | 2233.89 | 28.2% |
| down (w2) | 7168 | 2048 | 2606.08 | 2253.20 | 28.4% |
</details>

## 4. `triton_group_weight_amax` (group_weight_amax_triton.py)

- Computes, per expert group `g`, `amax[g] = max|A|`. It is equivalent to `x.float().abs().max()` and `torch.linalg.vector_norm(dtype=torch.float)`.

<details>
<summary>triton_group_weight_amax performance</summary>

| shape | MB | eager `.float().abs().max()` | vector_norm | triton E=1 | main/triton |
| --- | ---: | ---: | ---: | ---: | ---: |
| 671B expert 2048×7168 | 29 | 67.2 | 15.2 | 10.5 | 6.39x |
| Llama 3 8B gate/up 14336×4096 | 117 | 274.5 | 42.0 | 26.8 | 10.26x |
| Llama 3 70B gate/up 28672×8192 | 470 | 1038.1 | 132.3 | 81.8 | 12.69x |

</details>

### Benchmark Details
- Shapes from `deepseek_v3_shapes.py`. 671B uses 256 experts at EP=2 → 128 local experts, `dim=7168`, `moe_hidden=2048`. 
- GB/s is measured byte traffic (bf16 read + FP4 code writes + FP8 scale writes) over median kernel time.

# `GroupedExperts` — the TorchTitan integration vehicle

The kernels are consumed by a `GroupedExperts` nn.Module in `nvfp4_training/nvfp4_single_gpu_example.py`). It holds expert weights `w1/w2/w3` and calls `_to_nvfp4_then_scaled_grouped_mm` for the gate, up, and down projections with `SiLU(gate) * up`.

```python
gate = _to_nvfp4_then_scaled_grouped_mm(x, self.w1, sign_vector, sr_seed, offs=offsets)
up   = _to_nvfp4_then_scaled_grouped_mm(x, self.w3, sign_vector, sr_seed, offs=offsets)
hidden = F.silu(gate) * up
out  = _to_nvfp4_then_scaled_grouped_mm(hidden, self.w2, sign_vector, sr_seed, offs=offsets)
```

Input is pre-routed and packed by expert and `offsets = cumsum(num_tokens_per_expert)` gives the jagged group boundaries.

— NVFP4 currently integrates at the **functional-op level** inside `GroupedExperts.forward`, because it needs an explicit RHT `sign_vector` and stochastic-rounding `sr_seed` that don't fit the auto-quantize op-swap. A future `NVFP4…OpConfig` wrapper could follow the MXFP8 subclass pattern if transparent `quantize_(model, ...)` integration is desired.

# Training Loss Convergence

Eager Trainer 200M-token run for NVFP4 and BF16

<details>
<summary>Training Run Setup</summary>

- Model: DeepSeek V3 16B
- Trainer: eager
- Precision path: BF16
- Compile: TorchTitan eager compile enabled
- Parallelism: 4x FSDP, TP 1, EP 4
- Local batch size: 16
- Global batch size: 128
- Gradient accumulation steps: 2
- Sequence length: 4096
- Tokens per optimizer step: 524,288
- Steps requested/completed: 382
- Tokens requested/processed: 200,278,016
- Dataset: c4
</details>

image


rdspring1:
@vkuzo The plan is to use OAI Triton kernels as a jumping off point and the add optimized CuTeDSL kernels like https://github.com/pytorch/ao/pull/4487.

vkuzo:
@claude review.  Is there a plain pytorch (no triton) reference for these new kernels?

rdspring1:
table4_training_fwd_bwd_sweep

From Table 4 of my [kernel analysis](https://github.com/rdspring1/nvfp4-training-harness/blob/deepseek-debugmodel/deepseek_v3/kernel_analysis/nvfp4_grouped_gemm_crossover.md) of `GroupedExperts`.

The chart compares fwd+bwd time (ms) of `GroupedExperts` of bf16, mxfp8, TE nvfp4, and AO Triton nvfp4 against number of tokens per expert. 8 experts for grouped GEMM with even balance. Looking at DSV4 and GLM 5.2, the number of experts per GPU is [6,14] with [16K, 37K] tokens per expert. At 32K tokens per expert, the quantization kernels are the gap between mxfp8 and TE nvfp4 performance.

@vkuzo I added pytorch oracles for `triton_group_rht_amax` and `triton_group_weight_quantize_2d`.

torchao/prototype/moe_training/nvfp4_training/nvfp4_single_gpu_example.py
```diff
@@ -0,0 +1,168 @@
...
+        self.register_buffer(
+            "sr_seed", torch.tensor([1234], dtype=torch.int64, device=device)
+        )
+
+        nn.init.normal_(self.w1, mean=0.0, std=0.02)
+        nn.init.normal_(self.w2, mean=0.0, std=0.02)
+        nn.init.normal_(self.w3, mean=0.0, std=0.02)
+
+    def forward(
+        self, x: torch.Tensor, num_tokens_per_expert: torch.Tensor
```
vkuzo, line 85: does torchtitan's MoE logic give us `num_tokens_per_expert`? just curious why the API is set up this way vs just taking `offsets` directly

torchao/prototype/moe_training/nvfp4_training/nvfp4_single_gpu_example.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+def main() -> None:
+    if not torch.cuda.is_available():
+        print("Skipping NVFP4 example: CUDA is not available.")
+        return
+    if not has_triton():
+        print("Skipping NVFP4 example: Triton is not available.")
+        return
+    if not torch_version_at_least("2.10.0"):
```
vkuzo: optional: you can gate to PyTorch 2.13+ if you want - we don't officially support older PyTorch versions anymore in torchao, for simplicity.

torchao/prototype/moe_training/nvfp4_training/nvfp4_grouped_mm.py
```diff
@@ -0,0 +1,346 @@
...
+from torchao.prototype.mx_formats.nvfp4_tensor import per_tensor_amax_to_scale
+from torchao.quantization.quantize_.common import KernelPreference
+from torchao.utils import is_sm_at_least_100
+
+_ALIGNMENT = 128
+_SCALE_RECIPE = [F.ScalingType.BlockWise1x16, F.ScalingType.TensorWise]
+_SWIZZLE = [F.SwizzleType.SWIZZLE_32_4_4, F.SwizzleType.NO_SWIZZLE]
+
+
+def _scaled_grouped_mm(
```
vkuzo: just inline this?

torchao/prototype/moe_training/nvfp4_training/nvfp4_grouped_mm.py
```diff
@@ -0,0 +1,346 @@
...
+        scale_recipe_b=_SCALE_RECIPE,
+        swizzle_a=_SWIZZLE,
+        swizzle_b=_SWIZZLE,
+        offs=group_end_offsets,
+        output_dtype=torch.bfloat16,
+    )
+
+
+@conditional_nostrict_trace
+def _to_nvfp4_then_scaled_grouped_mm(
```
vkuzo: I think the function name should include RHT and RS

test/prototype/moe_training/nvfp4_training/test_group_rht_quantize_row_col_triton.py
```diff
@@ -0,0 +1,507 @@
...
+        hs,
+        spec.shape_rep,
+        amax_row,
+        amax_col,
+        None,
+        False,
+    )
+    _check_output_shapes(spec, qa, sfa, qd, sfd)
+
+    expected_col_sf = torch.empty(
```
vkuzo, line 244: can we factor L236:L272 into a function `triton_group_rht_quantize_row_col_ref`, to make it clear that we have a reference, a triton implementation of it, and they match up to a tolerance

test/prototype/moe_training/nvfp4_training/test_group_quantize_2d_triton.py
```diff
@@ -0,0 +1,153 @@
...
+    actual_codes, actual_scales, _, _ = triton_group_weight_quantize_2d(
+        weights, global_amax, num_tensors=E
+    )
+
+    for expert in range(E):
+        expected_scales, expected_codes = nvfp4_quantize(
+            weights[expert],
+            per_tensor_scale=per_tensor_amax_to_scale(global_amax[expert]),
+        )
+        expected_scales = to_blocked(expected_scales).view_as(actual_scales[expert])
```
vkuzo: can we wrap this into a helper `group_weight_quantize_2d_ref`

torchao/prototype/moe_training/nvfp4_training/nvfp4_single_gpu_example.py
rdspring1: Yes, `GroupedExperts` from TorchTitan provides `num_tokens_per_expert`. This "nvfp4" `GroupedExperts` is intended to be used by the converter. Taking offsets directly would break that replacement interface, so the current API is intentional.

torchao/prototype/moe_training/nvfp4_training/nvfp4_grouped_mm.py
rdspring1: Renamed to `_to_nvfp4_rht_rs_then_scaled_grouped_mm`

rdspring1:
@vkuzo @andrewor14 I have early training runs for full DSV3 671B with 64 GB300s.

## DSV3 671B Run:
* MXFP8 quantized all 61 layers including attention linears + FFN + shared experts + routed experts.
* NVFP4 quantized the first 51 FFN layers covering dense FFN + shared experts + routed experts. Attention and last 10 layers remained in BF16.
* OAI Triton NVFP4 is 5.5% slower than MXFP8 but 15.4% faster than BF16.

### 16 nodes / 64 GPUs, ep-32, bs-8, seq 4096, gb300, 1h TimeLimit

| cell | TFLOP/s/GPU | tps | vs bf16 | peak memory | max step |
| --- | ---: | ---: | ---: | ---: | ---: |
| bf16 | 731.6 | 2602 | — | 225.46 GiB (81.50%) | 270 |
| nvfp4-bf16-tail-attn | 844.4 | 3004 | +15.4% | 227.03 GiB (82.07%) | 280 |
| mxfp8 | 890.8 | 3169 | +21.8% | 222.49 GiB (80.43%) | 320 |

curves-16n-1h-no-cluster


## Updates: 
- [Correctness] Fixed a bug in colwise scale factors in `group_rht_quantize_row_col_triton.py` where I failed to account for the groups when writing to gmem.
- [Optimization] Replaced global amax calculation for 2d weight quantization with triton kernel. Linear is a subcase of grouped version, so reused the kernel for the linear case.

benchmarks/prototype/nvfp4_training/deepseek_v3_shapes.py
```diff
@@ -0,0 +1,60 @@
+"""TorchTitan DeepSeek-V3 expert-weight shapes shared by grouped benchmarks."""
+
+from dataclasses import dataclass
+
+
+@dataclass(frozen=True)
+class DeepSeekV3ModelShape:
```
vkuzo: maybe `Config` instead of `Shape`, to better match what is in this struct?

test/prototype/moe_training/nvfp4_training/test_hadamard_quantize_row_col.py
```diff
@@ -304,24 +309,53 @@ def _unpack_fp4_magnitudes(codes: torch.Tensor) -> torch.Tensor:
...
 def _assert_zero_quantized(
     codes: torch.Tensor,
     scales: torch.Tensor,
     dequantized: torch.Tensor,
 ) -> None:
+    """All-zero input packs to zero codes, stores a zero block scale, and
+    dequantizes to exactly zero.
+
+    Both backends emit 0 (not E4M3 eps) for a zero block, matching
+    TransformerEngine, which applies no lower clamp to the per-vector scale.
```
vkuzo, line 406: can you explain the motivation behind this behavior change

test/prototype/moe_training/nvfp4_training/test_hadamard_quantize_row_col.py
```diff
@@ -351,14 +375,22 @@ def _assert_near_zero_values_do_not_saturate(
...
-    div.rn.f32, causing ~0.2% nibble differences at FP4 midpoints. Use the SQNR
-    test for quantization quality validation.
+    Not bitwise: the kernels are matched against TransformerEngine, which uses
+    correctly-rounded div_rn and no lower clamp on the per-vector scale, whereas
+    the mx_formats reference multiplies by a reciprocal and applies an E4M3_EPS
+    floor. The scales are mathematically identical and differ only in fp32
+    rounding order, so they land on the same or an adjacent representable e4m3
+    value. Being bitwise-exact with TE and bitwise-exact with the torch
+    reference are mutually exclusive; TE is the ground truth these kernels
+    interoperate with.
```
vkuzo: I get what this is saying, but it's a bit confusing.  How about:
1. check in a plain PyTorch reference of the behavior you want (if I understand this PR correctly, you are matching TE bitwise)
2. ensure that your fused kernels match the reference in (1)

Then you can document where are the places that (1) currently diverges from other potential references, such as the `mx_formats` prototype package.  Long term we should align numerics, it doesn't need to block this PR.

torchao/prototype/moe_training/nvfp4_training/_cutedsl_kernels_impl.py
```diff
@@ -383,10 +382,12 @@ def _quant16_from_amax(
 ):
     """Quantize 16 f32 values to NVFP4 (w0,w1 packed u32, pvscale_fp8) using a given block amax
     (1x16 or a shared 16x16 amax). sr selects stochastic rounding over RTNE."""
-    # Clamp to [eps, max]: a zero/near-zero block stores eps, not 0 (matches triton, and keeps
-    # the decode reciprocal finite).
+    # Cap at FP8_E4M3_MAX only, no lower clamp: pvscale is non-negative and TE emits a zero
```
vkuzo: can you explain what TE is doing for an all-zero block and why?  Emitting a zero scale for an all-zero block sounds problematic to me, because then there is no way to recover the original high precision data from (qdata, scale).

vkuzo:
> MXFP8 quantized all 61 layers including attention linears + FFN + shared experts + routed experts.
> NVFP4 quantized the first 51 FFN layers covering dense FFN + shared experts + routed experts. Attention and last 10 layers remained in BF16.

This makes sense based on our earlier explanations of the nvfp4 pretraining best practices.  Optional nit: would be clearer if the graphs had "mxfp8" and "nvfp4_except_attn_and_last_10_layers", to make it clear that this is not an apples to apples comparison of applying mxfp8/nvfp4 to the same set of layers.

vkuzo:
@rdspring1, thanks for adding this! This is a great update and high level numerics data you shared looks good.  Can we just expand more on the reasoning behind the numerics change (I commented inline), and then this should be good to go

torchao/prototype/moe_training/nvfp4_training/_cutedsl_kernels_impl.py
rdspring1: TE computes the stored per-block decode scale as `E4M3(block_amax * global_encode_scale / 6)`. For an all-zero block, `block_amax == 0`, so the decode scale is zero. TE caps the reciprocal encoding multiplier at `FP32_MAX`. Multiplying the block’s zero inputs by that finite multiplier still produces zero FP4 data. Dequantization therefore reconstructs the original all-zero block exactly. The actual lossy edge case here is a small nonzero block whose E4M3 scale underflows to zero. TE flattens that block on dequantization, and this kernel preserves that behavior for compatibility.

test/prototype/moe_training/nvfp4_training/test_hadamard_quantize_row_col.py
rdspring1: I see. The current test conflates the intended TE semantics with comparison against the existing `mx_formats` implementation, which has different operation ordering and an E4M3 epsilon floor. I’ve added a local plain-PyTorch reference that reproduces TE’s global-scale and per-block-scale arithmetic, including the absence of a lower clamp, and require the fused kernels’ scale outputs to match it bitwise. 

This compatibility test measures the difference from `mx_formats` kernels. Packed FP4 codes will remain quality-tested rather than bitwise-tested because the fused kernels use an approximate reciprocal.

test/prototype/moe_training/nvfp4_training/test_hadamard_quantize_row_col.py
rdspring1: This test change accompanies removal of the `E4M3_EPS` lower clamp from the quantization kernels. TransformerEngine computes the per-block scale and casts it directly to E4M3, so an all-zero block stores scale zero rather than epsilon. For an exactly zero block, this does not change the quantized or dequantized values. It does matter for sufficiently small nonzero blocks whose scale underflows to zero. Those values are lost on dequantization. We accept that lossy edge case to match the TransformerEngine representation.

benchmarks/prototype/nvfp4_training/deepseek_v3_shapes.py
rdspring1: I renamed it to `DeepSeekV3ModelConfig`.

rdspring1:
@vkuzo I updated the comments for the numerics changes. Since I was using TE as training baseline and for debugging, I switched the quantization kernels to be bitwise with their numerics. I forgot that the `mxformat` kernels made a different choice.

> NVFP4 quantized the first 51 FFN layers covering dense FFN + shared experts + routed experts. Attention and last 10 layers remained in BF16.

This ^^^ is one of the conservative NVFP4 recipes for DSV3. It's useful for checking numerics of the quantization kernels. On the TorchTitan side, I'll get a more performant recipe.

rdspring1:
@vkuzo I replaced `/` with `tl.div_rn` in remaining triton kernels and consolidated the TE pytorch references in test/prototype/moe_training/nvfp4_training/nvfp4_reference.py. I redid the full DSV3 671B with 64 GB300s from earlier. It doesn't show significant divergence from earlier run.

curves-16n-1h-triton-eq-mxfp8-bf16

vkuzo:
@claude review

vkuzo:
@rdspring1 , thank you!

Claude's "Backward still mints device RNG" comment seems relevant, thoughts?

rdspring1:
@vkuzo "Backward still mints device RNG" by claude is wrong. Grouped kernels are following the same pattern as the linear kernels. `torch.randint` creates new random numbers under cuda graphs. The forward pass doesn't use stochastic rounding, so doesn't use RNG.

More lengthy response from Claude:
SR offsets come from the default CUDA generator (no generator= arg), as in nvfp4_linear.nvfp4_mm_triton. This is the graph-safe construction, not an exception to it: the kernel tl.loads seed and offset from device pointers instead of baking them into launch args, and the framework advances the default generator between replays, so each replay gets fresh noise. A host-side or ctx-threaded offset would freeze at capture time or force a cudagraph skip. sr_seed fixes the Philox key only; there is no bitwise SR control today.

vkuzo:
> @vkuzo "Backward still mints device RNG" by claude is wrong. Grouped kernels are following the same pattern as the linear kernels. torch.randint creates new random numbers under cuda graphs.

you are right, i created a quick local repro to convince myself: https://gist.github.com/vkuzo/738684d4462bee60d221753ea49e2246

vkuzo (approved):
great work! We can iterate in future PRs, this is in a good place to land. Thank you!

PR #4603, opened 2026-07-24
fix(autocast): resolve autocast state from the input's device type (#4601)

### LINKED ISSUE #4601: [Bug] Hardcoded CUDA autocast queries cause silent fp32 fallback on non-CUDA backends
### Description

Three locations hardcode CUDA autocast queries, causing mixed-precision to silently skip dtype casting on non-CUDA backends:

### 1. `torchao/float8/float8_linear.py:258`

```python
def forward(self, input: torch.Tensor) -> torch.Tensor:
    if torch.is_autocast_enabled():          # ← defaults to "cuda"
        autocast_dtype = torch.get_autocast_gpu_dtype()  # ← returns CUDA's dtype
        input = input.to(autocast_dtype)
```

`torch.is_autocast_enabled()` with no argument defaults to `"cuda"` (see `VariableTypeManual.cpp`: `device_type=\"cuda\"`), so it always checks CUDA autocast regardless of the actual device. On non-CUDA devices with device autocast active, this returns `False` → input stays in fp32.

### 2. `torchao/prototype/quantized_training/int8_mixed_precision.py:170`

```python
@Int8MixedPrecisionTrainingLinearWeight.implements(torch.nn.functional.linear)
    if torch.is_autocast_enabled("cuda"):    # ← hardcoded "cuda"
        dtype = torch.get_autocast_gpu_dtype()
```

### 3. `torchao/prototype/quantized_training/bitnet.py:145`

```python
@BitNetTrainingLinearWeight.implements_torch_function(F.linear)
    if torch.is_autocast_enabled("cuda"):    # ← hardcoded "cuda"
        dtype = torch.get_autocast_gpu_dtype()
```

### Impact

On non-CUDA backends with autocast enabled:
- Inputs are **not** cast to the autocast dtype (e.g. bf16) before quantization
- Quantization runs in fp32 — numerically correct but slower and diverges from CUDA behavior
- Silent performance regression with no warning or error

### Proposed fix

Use the input tensor's device type instead of hardcoding `"cuda"`:

```python
# float8_linear.py:258
device_type = input.device.type
if torch.is_autocast_enabled(device_type):
    autocast_dtype = torch.get_autocast_dtype(device_type)
    input = input.to(autocast_dtype)

# int8_mixed_precision.py:170 / bitnet.py:145
device_type = args[0].device.type
if torch.is_autocast_enabled(device_type):
    dtype = torch.get_autocast_dtype(device_type)
```

On CUDA, behavior is identical (`input.device.type == "cuda"`). On non-CUDA devices, autocast is now correctly detected.

### Steps to reproduce

```python
import torch
from torchao.quantization import quantize_, Float8DynamicActivationFloat8WeightConfig
from torchao.quantization.granularity import PerRow

# Any non-CUDA accelerator with autocast support
device = "xpu"  # or "mlu", "npu", etc.

# Enable autocast on the non-CUDA device
with torch.autocast(device_type=device, dtype=torch.bfloat16):
    x = torch.randn(4, 256, device=device)
    # Float8Linear.forward checks torch.is_autocast_enabled()
    # → returns False (defaults to "cuda", not the device's autocast)
    # → input stays fp32, quantization runs in fp32 instead of bf16
    y = model(x)

# Expected: input cast to bf16 before quantization (same as CUDA behavior)
# Actual: input stays fp32, numerically correct but ~2× slower


For the training paths (`int8_mixed_precision.py`, `bitnet.py`):


with torch.autocast(device_type=device, dtype=torch.bfloat16):
    # torch.is_autocast_enabled("cuda") → False on non-CUDA
    # → dtype cast skipped entirely
    y = model(x)
```

### Environment

```text
- torchao version: 0.17.0
- Relevant: the device must support `torch.autocast` (most PrivateUse1 backends do)
```

Anai-Guo:
Fixes #4601.

## Problem

Three call sites query CUDA's autocast state regardless of where the tensor actually lives:

| file | line | code |
|---|---|---|
| `torchao/float8/float8_linear.py` | 258 | `torch.is_autocast_enabled()` — the no-arg form defaults to `"cuda"` |
| `torchao/prototype/quantized_training/int8_mixed_precision.py` | 170 | `torch.is_autocast_enabled("cuda")` |
| `torchao/prototype/quantized_training/bitnet.py` | 145 | `torch.is_autocast_enabled("cuda")` |

On a non-CUDA accelerator with autocast active this returns `False`, so the input is never cast to the autocast dtype and quantization silently runs in fp32 — numerically fine, but slower and divergent from the CUDA path, with no warning. All three also call the deprecated `torch.get_autocast_gpu_dtype()`.

The `float8_linear.py` site already carried a `# For now, hardcode to GPU's autocast dtype / if we need CPU support in the future, we can add it` note, so this is the follow-up that comment anticipated.

## Fix

Resolve the device type from the input tensor:

```python
device_type = input.device.type
if torch.amp.is_autocast_available(device_type) and torch.is_autocast_enabled(device_type):
    autocast_dtype = torch.get_autocast_dtype(device_type)
    input = input.to(autocast_dtype)
```

## Why the `is_autocast_available` guard

The issue proposes calling `torch.is_autocast_enabled(device_type)` directly. That is not safe on its own — it raises for device types that have no autocast dispatch key:

```python
>>> torch.is_autocast_enabled("meta")
RuntimeError: unknown device type for autocast in get_autocast_dispatch_key_from_device_type
>>> torch.amp.is_autocast_available("meta")
False
```

Meta tensors show up in ordinary flows (`with torch.device("meta")` model init, FakeTensor tracing), so an unguarded lookup would turn a working path into a hard error. The `is_autocast_available()` short-circuit keeps those paths on the existing "autocast off" branch.

## Behavior

- **CUDA**: unchanged — `input.device.type == "cuda"`, so the same autocast state and dtype are read.
- **CPU / XPU / PrivateUse1 accelerators**: autocast is now detected on the device that is actually in use.
- **meta and other non-autocast device types**: treated as autocast-disabled, as today.

Verified against the pinned `ruff` 0.11.6 from `.pre-commit-config.yaml` (`check` and `format --check` both clean).

I did not add a regression test: reproducing the original miss needs a backend with autocast support other than CUDA, and I don't have one in CI reach to validate the test against. Happy to add one if you can point me at a suitable target (CPU autocast plus `Float8LinearConfig(emulate=True)` looks like it could work, but I couldn't verify it here).

### DIFF
--- a/torchao/float8/float8_linear.py
+++ b/torchao/float8/float8_linear.py
@@ -255,10 +255,11 @@ class Float8Linear(torch.nn.Linear):
     def forward(self, input: torch.Tensor) -> torch.Tensor:
         # Duplicate the autocast logic for F.linear, so that the output
         # of our module has the right original precision
-        if torch.is_autocast_enabled():
-            # For now, hardcode to GPU's autocast dtype
-            # if we need CPU support in the future, we can add it
-            autocast_dtype = torch.get_autocast_gpu_dtype()
+        device_type = input.device.type
+        if torch.amp.is_autocast_available(device_type) and torch.is_autocast_enabled(
+            device_type
+        ):
+            autocast_dtype = torch.get_autocast_dtype(device_type)
             input = input.to(autocast_dtype)
 
         output = matmul_with_hp_or_float8_args.apply(
--- a/torchao/prototype/quantized_training/bitnet.py
+++ b/torchao/prototype/quantized_training/bitnet.py
@@ -142,8 +142,11 @@ class BitNetTrainingLinearWeight(TorchAOBaseTensor):
 
 @BitNetTrainingLinearWeight.implements_torch_function(F.linear)
 def _(func, types, args, kwargs):
-    if torch.is_autocast_enabled("cuda"):
-        dtype = torch.get_autocast_gpu_dtype()
+    device_type = args[0].device.type
+    if torch.amp.is_autocast_available(device_type) and torch.is_autocast_enabled(
+        device_type
+    ):
+        dtype = torch.get_autocast_dtype(device_type)
         args = tuple(x.to(dtype) if x is not None else x for x in args)
     return _BitNetTrainingLinear.apply(*args, **kwargs)
 
--- a/torchao/prototype/quantized_training/int8_mixed_precision.py
+++ b/torchao/prototype/quantized_training/int8_mixed_precision.py
@@ -167,8 +167,11 @@ class Int8MixedPrecisionTrainingLinearWeight(TorchAOBaseTensor):
 
 @Int8MixedPrecisionTrainingLinearWeight.implements(torch.nn.functional.linear)
 def _(func, types, args, kwargs):
-    if torch.is_autocast_enabled("cuda"):
-        dtype = torch.get_autocast_gpu_dtype()
+    device_type = args[0].device.type
+    if torch.amp.is_autocast_available(device_type) and torch.is_autocast_enabled(
+        device_type
+    ):
+        dtype = torch.get_autocast_dtype(device_type)
         args = tuple(x.to(dtype) if x is not None else x for x in args)
     return _Int8MixedPrecisionTrainingLinearFunction.apply(*args, **kwargs)
 


ATMmonitor667:
Hi @Anai-Guo,

I saw your note about regression coverage. I independently prepared hardware-independent tests in draft PR #4741:

- CPU autocast coverage for the Int8 mixed-precision and BitNet handlers
- `meta` device coverage for the autocast-availability guard
- A device-generic expected autocast dtype in the existing Float8Linear test

The quantized-training tests mock the underlying autograd function, allowing them to verify that tensors are forwarded as bf16 without requiring a non-CUDA accelerator kernel.

I do not want to duplicate your work. Would you like me to extract these tests into a test-only commit or patch that you can incorporate into #4603? If so, I can do that and close #4741.

Anai-Guo:
Thanks @ATMmonitor667, that would be very welcome — please go ahead and extract them into a test-only commit/patch I can fold into #4603, and close #4741. Hardware-independent coverage (CPU autocast + `meta` device guard) is exactly what this was missing. Appreciate the coordination to avoid duplication.

PR #4581, opened 2026-07-20
[test][xpu]Skip test_comm as XPU CI does not support distributed test currently

zxd1997066:
Fix https://github.com/pytorch/ao/actions/runs/29694233643/job/88212101392#step:18:46555 oneCCL could not be initialized! Could not load any plugin.

Skip it as XPU CI does not support distributed test currently.

PR #4628, opened 2026-07-28
[xpu][test] Use device-agnostic device list in test_quantized_training

karol-brejna-i:
### What

Replaces the hardcoded CUDA device list in `test/prototype/test_quantized_training.py` with the
existing helper:

```diff
-_DEVICES = ["cpu"] + (["cuda"] if torch.cuda.is_available() else [])
+_DEVICES = get_available_devices()
```

### Why

The five int8 weight-only tests are device-parametrized and use only plain torch ops, but the hardcoded list means they never run on the XPU CI lane — they silently collect as CPU-only. `get_available_devices()` (`torchao/utils.py`) already returns `["cpu"]` + `"cuda"` or `"xpu"` (+ optional `"mps"`), so no new helper is needed.

No library code changes. No behavioural change on CUDA: on a CUDA machine the helper returns `['cpu', 'cuda']`, exactly the previous list.

### Validation

Run on an Intel XPU node (Python 3.12, `torch 2.14.0.dev20260714+xpu`, `pytorch-triton-xpu` present, single card via `ZE_AFFINITY_MASK=0`):

```bash
pytest -rs -v test/prototype/test_quantized_training.py
```

|              | before | after |
|--------------|-------:|------:|
| passed       |     16 |    32 |
| skipped      |     20 |     4 |
| failed/error |      0 |     0 |

16 new test IDs appear and pass — the `..._device_xpu` variants of `test_int8_stochastic_rounding`, `test_int8_weight_only_correctness`, `test_int8_weight_only_cat`, `test_int8_weight_only_compile` (including the `compile=True` parametrizations, i.e. inductor-xpu) and `test_int8_weight_only_training`.

CPU-only collection is unaffected, which was was checked explicitly:

```bash
# NB: ZE_AFFINITY_MASK="" does *not* hide the XPU; an out-of-range mask does.
CUDA_VISIBLE_DEVICES="" ZE_AFFINITY_MASK=99 pytest -q --collect-only \
  test/prototype/test_quantized_training.py
```

Collection succeeds, with the same test count as `main`.

### Notes

- `TestFSDP2` is intentionally left CUDA-hardcoded: the XPU CI lane is single-card and has no distributed support today (see #4581). Both of its tests already carry `skipif(not torch.cuda.is_available())` and skip cleanly on XPU with a visible reason.

### DIFF
--- a/test/prototype/test_quantized_training.py
+++ b/test/prototype/test_quantized_training.py
@@ -34,7 +34,10 @@ from torchao.prototype.quantized_training import (
     quantize_int8_rowwise,
 )
 from torchao.quantization.quant_api import quantize_
-from torchao.utils import get_available_devices, torch_version_at_least
+from torchao.utils import (
+    get_available_devices,
+    torch_version_at_least,
+)
 
 if common_utils.SEED is None:
     common_utils.SEED = 1234
--- a/test/prototype/test_quantized_training.py
+++ b/test/prototype/test_quantized_training.py
@@ -34,12 +34,12 @@ from torchao.prototype.quantized_training import (
     quantize_int8_rowwise,
 )
 from torchao.quantization.quant_api import quantize_
-from torchao.utils import torch_version_at_least
+from torchao.utils import get_available_devices, torch_version_at_least
 
 if common_utils.SEED is None:
     common_utils.SEED = 1234
 
-_DEVICES = ["cpu"] + (["cuda"] if torch.cuda.is_available() else [])
+_DEVICES = get_available_devices()
 
 
 def _reset():


test/prototype/test_quantized_training.py
```diff
@@ -34,12 +34,12 @@
     quantize_int8_rowwise,
 )
 from torchao.quantization.quant_api import quantize_
-from torchao.utils import torch_version_at_least
+from torchao.utils import get_available_devices, torch_version_at_least
```
xiaowangintel: (
get_available_devices,
torch_version_at_least,
)
Please follow the import format used above.

karol-brejna-i: Updated.

liangan1:
@claude review

PR #4629, opened 2026-07-28
Make the QAT tests to be device agnostic

karol-brejna-i:
### Summary

Makes QAT (fp8, MX) and integration export tests device-agnostic so they run on non-CUDA accelerators (like Intel XPU) instead of being skipped by hardcoded CUDA checks:

1. **`test/quantization/test_qat.py`**:
   - Replaced CUDA-specific checks with accelerator-agnostic equivalents (`torch.accelerator.is_available()`, `get_current_accelerator_device()`).
   - Scoped SM89 / compute-capability checks to CUDA only, preventing spurious skips on non-NVIDIA accelerators with native fp8 support.
   - Removed unnecessary GPU gates from `test_mx_fake_quantize_config` and `test_mx_config_error_handling` (pure Python/dataclass checks that can safely run on CPU).

2. **`test/integration/test_export.py`**:
   - Made `test_export_float8` device-agnostic using the current accelerator device instead of hardcoded `"cuda"`.
   - Scoped the sm89 check to CUDA only.

CUDA-only tests (MSLK, NVFP4, SM100-specific kernels) remain untouched.

### Test Plan

- **XPU**: Verified on Intel GPU (PVC / BMG).
  - `test/quantization/test_qat.py`: 29 previously skipped tests now pass.
  - `test/integration/test_export.py`: `test_export_float8` passes.
  - No new failures or regressions.
- **CPU**: Verified tests collect and run cleanly when no accelerator is present.
- **CUDA**: Changes are non-breaking; CUDA paths retain existing capability checks.

### DIFF
--- a/test/quantization/test_qat.py
+++ b/test/quantization/test_qat.py
@@ -1871,7 +1871,7 @@ class TestQAT(TestCase):
     @unittest.skipIf(
         torch.cuda.is_available()
         and not (is_sm_at_least_89() or is_MI300() or is_MI350()),
-        "Need sm89+ or MI300/MI350",
+        "Need CUDA sm89+ or ROCm MI300/MI350",
     )
     def test_quantize_api_fp8_fp8(self, granularity: Granularity):
         """
--- a/test/quantization/test_qat.py
+++ b/test/quantization/test_qat.py
@@ -1867,9 +1867,10 @@ class TestQAT(TestCase):
         self.assertGreaterEqual(convert_sqnr, target_convert_sqnr)
 
     @parametrize("granularity", [PerTensor(), PerRow()])
-    @unittest.skipIf(not _CUDA_IS_AVAILABLE, "skipping when cuda is not available")
+    @unittest.skipIf(not torch.accelerator.is_available(), "Need GPU available")
     @unittest.skipIf(
-        not (is_sm_at_least_89() or is_MI300() or is_MI350()),
+        torch.cuda.is_available()
+        and not (is_sm_at_least_89() or is_MI300() or is_MI350()),
         "Need sm89+ or MI300/MI350",
     )
     def test_quantize_api_fp8_fp8(self, granularity: Granularity):
@@ -2360,7 +2361,6 @@ class TestQAT(TestCase):
         torch.testing.assert_close(m.linear2.weight.scale, scale2)
         torch.testing.assert_close(m.sub.linear.weight.scale, sub_scale)
 
-    @unittest.skipIf(not _CUDA_IS_AVAILABLE, "skipping when cuda is not available")
     def test_mx_fake_quantize_config(self):
         """Test MXFakeQuantizeConfig dataclass with various element dtypes."""
         from torchao.prototype.mx_formats.config import ScaleCalculationMode
@@ -2394,7 +2394,7 @@ class TestQAT(TestCase):
         )
         self.assertEqual(config_fp8_e5m2.dtype, torch.float8_e5m2)
 
-    @unittest.skipIf(not _CUDA_IS_AVAILABLE, "skipping when cuda is not available")
+    @unittest.skipIf(not torch.accelerator.is_available(), "Need GPU available")
     @parametrize("bias", [True, False])
     @parametrize("input_shape", [(128, 256), (1, 128, 256), (2, 4, 128, 256)])
     @parametrize(
@@ -2406,6 +2406,7 @@ class TestQAT(TestCase):
         from torchao.prototype.qat import MXFakeQuantizeConfig, MXFakeQuantizedLinear
 
         K, N = 256, 128
+        device = get_current_accelerator_device()
 
         activation_config = MXFakeQuantizeConfig(
             dtype=dtype,
@@ -2416,14 +2417,14 @@ class TestQAT(TestCase):
             block_size=32,
         )
 
-        linear = torch.nn.Linear(K, N, bias=bias, device="cuda", dtype=torch.bfloat16)
+        linear = torch.nn.Linear(K, N, bias=bias, device=device, dtype=torch.bfloat16)
         mx_linear = MXFakeQuantizedLinear.from_linear(
             linear,
             activation_config=activation_config,
             weight_config=weight_config,
         )
 
-        x = torch.randn(*input_shape, device="cuda", dtype=torch.bfloat16)
+        x = torch.randn(*input_shape, device=device, dtype=torch.bfloat16)
         y_ref = linear(x)
         y_mx = mx_linear(x)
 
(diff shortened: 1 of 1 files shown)


test/quantization/test_qat.py
```diff
@@ -2406,6 +2406,7 @@ def test_mx_fake_quantized_linear_forward(self, bias, input_shape, dtype):
         from torchao.prototype.qat import MXFakeQuantizeConfig, MXFakeQuantizedLinear
 
         K, N = 256, 128
+        device = get_current_accelerator_device()
```
xiaowangintel: get_current_accelerator_device may be removed. Please use torch.accelerator.get_current_accelerator() instead.

test/quantization/test_qat.py
```diff
@@ -2438,18 +2439,19 @@ def test_mx_fake_quantized_linear_forward(self, bias, input_shape, dtype):
...
 
-    @unittest.skipIf(not _CUDA_IS_AVAILABLE, "skipping when cuda is not available")
+    @unittest.skipIf(not torch.accelerator.is_available(), "Need GPU available")
     @parametrize("bias", [True, False])
     def test_mx_fake_quantized_linear_backward(self, bias):
         """Test MXFakeQuantizedLinear backward pass."""
         from torchao.prototype.qat import MXFakeQuantizeConfig, MXFakeQuantizedLinear
 
         M, K, N = 128, 256, 128
+        device = get_current_accelerator_device()
```
xiaowangintel: Ditto

test/quantization/test_qat.py
```diff
@@ -2487,18 +2489,19 @@ def test_mx_fake_quantized_linear_backward(self, bias):
...
         self.assertGreaterEqual(w_grad_sqnr, 3.0)
 
-    @unittest.skipIf(not _CUDA_IS_AVAILABLE, "skipping when cuda is not available")
+    @unittest.skipIf(not torch.accelerator.is_available(), "Need GPU available")
     def test_mx_fake_quantized_linear_to_linear(self):
         """Test converting MXFakeQuantizedLinear back to nn.Linear."""
         from torchao.prototype.qat import MXFakeQuantizeConfig, MXFakeQuantizedLinear
 
         K, N = 256, 128
+        device = get_current_accelerator_device()
```
xiaowangintel: Ditto

test/quantization/test_qat.py
```diff
@@ -2555,18 +2557,19 @@ def test_mx_matmul_sqnr(self, shapes):
         from torchao.prototype.qat import MXFakeQuantizeConfig, MXFakeQuantizedLinear
 
         M, K, N = shapes
+        device = get_current_accelerator_device()
```
xiaowangintel: Ditto

test/quantization/test_qat.py
```diff
@@ -2575,19 +2578,20 @@ def test_mx_matmul_sqnr(self, shapes):
...
 
-    @unittest.skipIf(not _CUDA_IS_AVAILABLE, "skipping when cuda is not available")
+    @unittest.skipIf(not torch.accelerator.is_available(), "Need GPU available")
     def test_mx_training_simulation(self):
         """Simulate a simple training loop with MX QAT."""
         from torchao.prototype.qat import MXFakeQuantizeConfig, MXFakeQuantizedLinear
 
         M, K, N = 128, 256, 128
         num_steps = 5
+        device = get_current_accelerator_device()
```
xiaowangintel: Ditto

test/quantization/test_qat.py
```diff
@@ -1867,9 +1867,10 @@ def _test_quantize_api_against_ptq(
         self.assertGreaterEqual(convert_sqnr, target_convert_sqnr)
 
     @parametrize("granularity", [PerTensor(), PerRow()])
-    @unittest.skipIf(not _CUDA_IS_AVAILABLE, "skipping when cuda is not available")
+    @unittest.skipIf(not torch.accelerator.is_available(), "Need GPU available")
     @unittest.skipIf(
-        not (is_sm_at_least_89() or is_MI300() or is_MI350()),
+        torch.cuda.is_available()
+        and not (is_sm_at_least_89() or is_MI300() or is_MI350()),
         "Need sm89+ or MI300/MI350",
```
xiaowangintel: For CUDA, SM89+ is required. For ROCm, MI300/MI350 is required.

xiaowangintel: For CUDA, SM89+ is required. For ROCm, MI300/MI350 is required.

xiaowangintel: For CUDA, SM89+ is required. For ROCm, MI300/MI350 is required.

karol-brejna-i:
@xiaowangintel @liangan1 All the required changes were applied. Apart fro unrelated ROCM, the CI is green. 
Please, confirm the review and approve.

test/quantization/test_qat.py
```diff
@@ -2365,7 +2366,6 @@ def test_range_learning_convert_pass_qparams(
         torch.testing.assert_close(m.linear2.weight.scale, scale2)
         torch.testing.assert_close(m.sub.linear.weight.scale, sub_scale)
 
-    @unittest.skipIf(not _CUDA_IS_AVAILABLE, "skipping when cuda is not available")
```
liangan1, line 2368: @unittest.skipIf(not torch.accelerator.is_available(), "Need GPU available")

liangan1:
@claude  review.

test/quantization/test_qat.py
karol-brejna-i: Please, note that the method only builds `MXFakeQuantizeConfig` and checks its dataclass fields — it never allocates a tensor or touches a device (and there is pur Python only validation in `_post_init()`).
That is not device specific and I removed the CUDA gate.

liangan1 (approved):

vkuzo:
@claude review

vkuzo:
could you look into the two yellow suggestions from claude?

liangan1:
> could you look into the two yellow suggestions from claude?

Thanks @vkuzo. We will do it

karol-brejna-i:
Thanks for the review.

On the `test_export_float8` point: I believe this one is a false positive — the accelerator gate wasn't lost, it's on the enclosing class rather than the method:
```python
# test/integration/test_integration.py:711
@unittest.skipIf(not torch.accelerator.is_available(), "requires gpu")
class TestExport(unittest.TestCase):
```

It skips all the tests inside the class before it ever looks at the method decorator, so on a CPU-only runner the method not executed.

I remember checking this on an XPU box with the GPU hidden via `ZE_AFFINITY_MASK=99`. CPU-only runs before/after this change are byte-identical (`8 passed, 52 skipped`). So the change doesn't affect CPU-only behavior.

It is confirmed by this PR's own CI: on the CPU-only job (`test (CPU 2.13, linux.4xlarge, ...)`, run [34864858253](https://github.com/pytorch/ao/actions/runs/34864858253/job/104046053037?pr=4629#step:17:2184)):
```
test/integration/test_integration.py::TestExport::test_export_float8 SKIPPED
```

Let me think a little bit about the other remark.

karol-brejna-i:
@vkuzo  As for the second Claude's note, I would opt for keeping my changes in place. 
Here is the explanation.  

### Regarding the suggestion to use  `get_current_accelerator_device()`:

My understanding is that transitioning from `get_current_accelerator_device()` to `current_accelerator()` was an established direction. As noted in this PR (and a few others, I believe), `get_current_accelerator_device()` is deprecated (or there is deprecation plan for it) and expected to be removed soon (it simply wraps return torch.accelerator.current_accelerator() internally).


### Regarding the suggestion to reuse the module-level `_DEVICE` constant:

This constant does not seem to represent the target implementation and is likely to change. 
This is highlighted by the existing comment:
```python
# TODO: put this in a common test utils file
```

Additionally, `_DEVICE` is currently unused. While it is declared at `test_qat.py:109`, the references to `_DEVICE` at lines 617, 669, and 713 are local variables that shadow the module-level constant. Reusing it here would be its first actual use, rather than following an established pattern in the codebase.

liangan1:
@claude review this pr

liangan1 (approved):
LGTM @vkuzo  Pls help to review it again. Thanks.

test/integration/test_integration.py
```diff
@@ -769,7 +769,8 @@ def forward(self, x):
         self.assertTrue(torch.equal(after_export, ref))
 
     @unittest.skipIf(
-        not is_sm_at_least_89(), "Requires GPU with compute capability >= 8.9"
+        torch.cuda.is_available() and not is_sm_at_least_89(),
```
vkuzo, line 772: I don't understand this, so the test will run if cuda is not available? that is not expected.  Can we gate it with an allowlist for hardware that supports the capability

liangan1: Thanks for your comments. We will work on the capability check.

karol-brejna-i: > Thanks for your comments. We will work on the capability check.

@liangan1 Should we revert this change for now and reintroduce it once we've finalized the design for the XPU capability check?"

PR #4632, opened 2026-07-28
Fix speedup column name for FP4 recipes in bench_matmul

jagadish-amd:
Use fp4_speedup for mxfp4_cutlass and nvfp4; keep fp8_speedup for FP8 recipes.


jagadish-amd:
@vkuzo can u plz review and approve/merge, thanks.

vkuzo (approved):

PR #4635, opened 2026-07-29
Fix partial K tile masking in int_scaled_matmul

### LINKED ISSUE #4633: `int_scaled_matmul` reads out of bounds and returns wrong results when K is even but not a multiple of `BLOCK_K
### Description

`scaled_matmul_kernel_with_block_pointers` turns off the K-axis bounds mask when `EVEN_K` is true. The host sets `EVEN_K = (K % 2 == 0)`, but what the kernel actually needs is `K % BLOCK_K == 0`. When K is even and is *not* a whole number of `BLOCK_K` tiles, the last K tile is loaded with no mask: it reads past the end of A and B and adds whatever is there into the result.

Host predicate — `torchao/kernel/intmm_triton.py:311`:

```python
EVEN_K=(K % 2 == 0),
```

Kernel — `torchao/kernel/intmm_triton.py:227-233`:

```python
for k in range(K, 0, -BLOCK_K):
    if EVEN_K:
        a = tl.load(A)                                  # <-- no mask
        b = tl.load(B)                                  # <-- no mask
    else:
        a = tl.load(A, mask=rk[None, :] < k, other=0.0)
        b = tl.load(B, mask=rk[:, None] < k, other=0.0)
```

`K % 2 == 0` does not imply `K % BLOCK_K == 0` — for example K=34 with BLOCK_K=32.

Worth noting why this bites immediately: the M and N axes *are* protected (`rm % M` / `rn % N` wraparound plus a masked store), so an over-read there folds back into the tensor. The K axis has no protection at all, so a wrong K predicate is a straight out-of-bounds read.

### Steps to reproduce

```python
import torch
import triton

from torchao.kernel.intmm_triton import int_scaled_matmul_kernel

M = N = 64
K, BLOCK_K = 34, 32  # K is even, but 34 % 32 == 2

# Pad A and B with a sentinel so the out-of-bounds read is deterministic.
a_pad = torch.full((M, K + BLOCK_K), 127, dtype=torch.int8, device="cuda")
a_pad[:, :K] = torch.randint(-128, 128, (M, K), dtype=torch.int8, device="cuda")
b_pad = torch.full((K + BLOCK_K, N), 127, dtype=torch.int8, device="cuda")
b_pad[:K, :] = torch.randint(-128, 128, (K, N), dtype=torch.int8, device="cuda")
a, b = a_pad[:, :K], b_pad[:K, :]

scales = torch.ones((M, 1), dtype=torch.float32, device="cuda")
c = torch.zeros((M, N), dtype=torch.float32, device="cuda")

# Pin the config: BLOCK_K is normally picked by timing, so an autotuned run
# could pass by luck on one machine and fail on another.
config = triton.Config(
    {"BLOCK_M": 32, "BLOCK_N": 32, "BLOCK_K": BLOCK_K, "GROUP_M": 8},
    num_stages=1,
    num_warps=2,
)

out = int_scaled_matmul_kernel(a, b, scales.expand(M, N), c, config)
ref = (a.float() @ b.float()) * scales
print("max abs error:", (out - ref).abs().max().item())

## Output:
##
## max abs error: 483870.0
##
```

### Environment

```text
torchao `main` @ `b7f01f18` (`intmm_triton.py` unchanged since `3bad6a2`)
torch 2.13.0+cu126, triton 3.7.1
NVIDIA H100 NVL (sm90)
```

bm1016bm-svg:
## Summary
- mask the final K tile unless `K` is divisible by the selected `BLOCK_K`
- add a sentinel-padded CUDA regression for `K=34`, `BLOCK_K=32`
- pin the Triton config in the test so autotuning cannot hide the bug

Fixes #4633

## Test plan
- `uvx --from ruff==0.11.6 ruff check test/kernel/test_autotuner.py`
- `uvx --from ruff==0.11.6 ruff format test/kernel/test_autotuner.py`
- `git diff --check`
- CUDA regression not run locally because Torch/Triton are not installed in this Windows environment; CI provides the required runtime

### DIFF
--- a/test/kernel/test_autotuner.py
+++ b/test/kernel/test_autotuner.py
@@ -98,6 +98,37 @@ class TestQuantFlow(unittest.TestCase):
         assert out32_2.dtype == out32_1.dtype
         torch.testing.assert_allclose(out32_1, out32_2)
 
+    @unittest.skipIf(not torch.cuda.is_available(), "Need CUDA available")
+    def test_int_scaled_mm_partial_k_tile(self):
+        import triton
+
+        from torchao.kernel.intmm_triton import int_scaled_matmul_kernel
+
+        m = n = 64
+        k, block_k = 34, 32
+        a_pad = torch.full((m, k + block_k), 127, dtype=torch.int8, device="cuda")
+        b_pad = torch.full((k + block_k, n), 127, dtype=torch.int8, device="cuda")
+        a_pad[:, :k] = torch.randint(-128, 128, (m, k), dtype=torch.int8, device="cuda")
+        b_pad[:k] = torch.randint(-128, 128, (k, n), dtype=torch.int8, device="cuda")
+        a, b = a_pad[:, :k], b_pad[:k]
+        scales = torch.ones((m, 1), dtype=torch.float32, device="cuda")
+        output = torch.zeros((m, n), dtype=torch.float32, device="cuda")
+        config = triton.Config(
+            {
+                "BLOCK_M": 32,
+                "BLOCK_N": 32,
+                "BLOCK_K": block_k,
+                "GROUP_M": 8,
+            },
+            num_stages=1,
+            num_warps=2,
+        )
+
+        result = int_scaled_matmul_kernel(a, b, scales, output, config)
+        expected = (a.float() @ b.float()) * scales
+
+        torch.testing.assert_close(result, expected)
+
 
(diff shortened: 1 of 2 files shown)


bm1016bm-svg:
Closing as superseded by #4717. That merged change removed the experimental torchao/kernel/intmm_triton.py path containing this bug; current int_scaled_matmul uses safe_int_mm/torch._int_mm instead. Rebasing this PR would only restore deleted code, so there is no valid patch left to merge.

PR #4636, opened 2026-07-29
[xpu][test][float8] Enable test_float8_utils on XPU

karol-brejna-i:
## Summary

Enables `test_round_scale_down_to_power_of_2_valid_inputs` to run on XPU (in addition to CUDA).

### Changes

- Replace `@unittest.skipIf(not torch.cuda.is_available(), ...)` with `torch.accelerator.is_available()`
- Resolve device inside the test body via `get_current_accelerator_device()` (never at module scope)
- Replace `.cuda()` calls with `.to(device)`

`test_non_float32_input` is already device-independent.

### Validation

All checks performed on an XPU machine (torch 2.14.0.dev20260714+xpu):

- Numerics probe — `_round_scale_down_to_power_of_2` tested on CPU and XPU for all 8 parametrized cases (one, inf, nan, smallest subnormal, largest normal, smallest normal, largest-less-than-one, smallest-greater-than-one). All cases: OK on both devices.
- Before/after run (`pytest -q -rs test/float8/test_float8_utils.py`, XPU box):
  - Before (main): 8 passed, 8 skipped (`CUDA not available`)
  - After (this PR): 16 passed, 0 skipped
- Static checks — `ruff check` and `ruff format --check` both clean.

### DIFF
--- a/test/float8/test_float8_utils.py
+++ b/test/float8/test_float8_utils.py
@@ -10,11 +10,12 @@ import torch
 
 from torchao.float8.float8_utils import _round_scale_down_to_power_of_2
 from torchao.testing.utils import skip_if_rocm
+from torchao.utils import get_current_accelerator_device
 
 
 # source for notable single-precision cases:
 # https://en.wikipedia.org/wiki/Single-precision_floating-point_format
-@unittest.skipIf(not torch.cuda.is_available(), "CUDA not available")
+@unittest.skipIf(not torch.accelerator.is_available(), "GPU not available")
 @pytest.mark.parametrize(
     "test_case",
     [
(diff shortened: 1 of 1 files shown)


karol-brejna-i:
Note: please add the `ciflow/xpu` label to trigger the XPU CI lane (per `.github/scripts/ci_test_xpu.sh`). The label may need to be created first if it doesn't exist in this repo yet.

karol-brejna-i:
Need to wait until CI "global" fail will be over (see: https://github.com/pytorch/ao/issues/4625).

test/float8/test_float8_utils.py
```diff
@@ -37,9 +38,10 @@ def test_round_scale_down_to_power_of_2_valid_inputs(
     test_case: dict,
 ):
     test_case_name, input, expected_result = test_case
+    device = get_current_accelerator_device()
```
xiaowangintel: get_current_accelerator_device may be removed. Please use torch.accelerator.get_current_accelerator() instead.

karol-brejna-i:
CI (except for ROCM job which is broken for some time)  is green.

PR #4639, opened 2026-07-29
Fix CI test failures: fp32 precision leak and ROCm rowwise-fp8 warning

vkuzo:
Fixes two independent failures surfaced by recent CI runs.

1. fp32 precision flag leak (1xL4, 1xH100, CPU/CUDA nightly)

   PyTorch's new test harness fails any test that mutates a global fp32
   matmul precision flag without restoring it. quantize_() sets tf32 on
   torch.backends.{cuda,mkldnn}.matmul.fp32_precision, so tests that call
   it leak that state and trip the checker. Following the same fix as
   #4593 (which patched TestQuantFlow), add setUp/tearDown that snapshot
   and restore the two flags to the remaining affected classes:
     - TestAffineQuantizedFloat8Compile (test/dtypes/test_affine_quantized_float.py)
     - TestInt4OpaqueTensor (test/prototype/test_int4_opaque_tensor.py)
     - TestAWQ (test/prototype/test_awq.py)

2. ROCm rowwise-fp8 warning crash

   test_quantize_api_fp8_fp8_granularity1 (PerRow) failed on ROCm with
   "AttributeError: 'NoneType' object has no attribute 'startswith'" in
   _maybe_warn_rowwise_fp8_cuda_12_9 (added in #4588). On ROCm/HIP builds
   torch.cuda.is_available() is True but torch.version.cuda is None, so
   torch.version.cuda.startswith("12.9") raised. Guard for None so ROCm
   short-circuits out of the CUDA-12.9-specific warning.

### DIFF
--- a/test/dtypes/test_affine_quantized_float.py
+++ b/test/dtypes/test_affine_quantized_float.py
@@ -46,6 +46,17 @@ class ToyLinearModel(torch.nn.Module):
 
 
 class TestAffineQuantizedFloat8Compile(InductorTestCase):
+    def setUp(self):
+        super().setUp()
+        # quantize_() sets a global float32 matmul precision; snapshot fp32_precision
+        self._prev_cuda_matmul_fp32 = torch.backends.cuda.matmul.fp32_precision
+        self._prev_mkldnn_matmul_fp32 = torch.backends.mkldnn.matmul.fp32_precision
+
+    def tearDown(self):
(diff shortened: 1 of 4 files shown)


vkuzo:
* __->__ #4639

andrewor14 (approved):

vkuzo:
actually, i like https://github.com/pytorch/ao/pull/4631 better

PR #4643, opened 2026-07-29
Stop quantize_ from leaking global fp32 matmul precision (complete fix)

vkuzo:
This is a more complete version of https://github.com/pytorch/ao/pull/4631.
That PR fixed the root cause but its CI still failed with 14 "fp32 precision
flag leak detected" errors in test/prototype/test_quantized_training.py,
because removing the leak unmasked a second, pre-existing leak in that test.

Root cause (from #4631):

recommended_inductor_config_setter() called
torch.set_float32_matmul_precision("high"). That is not an inductor config,
it is process-global state: it moves torch.backends.{cuda,mkldnn}.matmul.
fp32_precision from 'none' to 'tf32' and never restores them. quantize_()
runs that setter by default, so any test that quantizes a model leaks the
flags and trips PyTorch's new TestCase.tearDown invariant check.

Changes carried over from #4631:

- Drop the set_float32_matmul_precision call from
  recommended_inductor_config_setter(); the five inductor knobs are
  unchanged. Callers that want tf32 set it themselves.
- Drop the mirror line from undo_recommended_configs() in
  test/integration/test_integration.py, which only existed to undo the
  setter and leaked 'none' -> 'ieee' itself.
- Add TestQuantFlow::test_quantize_does_not_leak_fp32_precision.

Additional fix (what #4631 missed / left as follow-up):

test/prototype/test_quantized_training.py has its own leak: _reset() calls
torch.set_float32_matmul_precision("highest") (== 'ieee') and never restores
it. Before #4631 this was masked -- each test called _reset() and then
quantize_(), and quantize_()'s setter reset the flags back to 'tf32', which
happened to match the (also leaked) 'tf32' baseline, so tearDown saw no
change. Once quantize_() stops touching the flags, _reset()'s 'ieee' survives
to tearDown and the check fires ('tf32' -> 'ieee'), failing 14 tests.

Fix: add setUp/tearDown to TestQuantizedTraining that snapshot and restore
the two fp32 matmul precision flags, matching the pattern already used by
TestQuantFlow. _reset() still sets "highest" during the test body (needed for
correctness), but the flags are restored before torch's leak check runs.

Verified: the new leak test and a previously-failing training test pass, and
a direct simulation of the tearDown leak-check semantics from a 'tf32'
baseline shows the flags restored after tearDown (no leak).

### DIFF
--- a/test/integration/test_integration.py
+++ b/test/integration/test_integration.py
@@ -116,7 +116,6 @@ def undo_recommended_configs():
     torch._inductor.config.force_fuse_int_mm_with_mul = False
     torch._inductor.config.fx_graph_cache = False
     torch._inductor.config.triton.unique_kernel_names = False
-    torch.set_float32_matmul_precision("highest")
 
 
 def combine_parameters(a, b):
--- a/test/prototype/test_quantized_training.py
+++ b/test/prototype/test_quantized_training.py
@@ -51,6 +51,19 @@ def _reset():
 
 
 class TestQuantizedTraining(TestCase):
+    def setUp(self):
+        super().setUp()
+        # _reset() calls torch.set_float32_matmul_precision("highest"), which mutates
+        # the process wide fp32 precision flags; snapshot them so tearDown can restore
+        # and torch's TestCase.tearDown does not flag a "fp32 precision flag leak".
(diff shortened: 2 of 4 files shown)


vkuzo:
* __->__ #4643

drisspg (approved):

PR #4644, opened 2026-07-29
Split PT2E tests into dedicated CPU and GPU CI workflows

vkuzo:
Summary:

analysis: https://gist.github.com/vkuzo/1267ce1872eca8fbd583bdc48779a220 

The PT2E test suite (test/quantization/pt2e) is slow -- ~40+ min, almost
entirely torch.export / torch.compile / Inductor codegen -- and CPU/x86-centric.
Running it on every regression leg, and especially on the expensive GPU runner,
is wasteful and largely redundant.

This change:
- Excludes PT2E from the existing workflows that run the full suite
  (regression_test.yml, regression_test_rocm.yml, nightly_smoke_test.yml) via
  --ignore=test/quantization/pt2e.
- Adds regression_test_pt2e_cpu.yml: runs the FULL PT2E suite on an Intel CPU
  runner (nightly + torch 2.13.0 pinned). The x86 Inductor tests gate on CPU ISA
  features (AMX tile, AVX512 VNNI), so an Intel CPU runner is where they are
  actually exercised.
- Adds regression_test_pt2e_gpu.yml: runs the accelerator-relevant PT2E tests on
  CUDA (nightly + 2.13.0 pinned) and ROCm (nightly). The slow x86 Inductor tests
  (test_x86inductor_fusion.py, test_x86inductor_quantizer.py) are excluded here
  since they are CPU-only and add ~45 min with zero GPU coverage.

Each workflow has comments explaining what runs where and why.

Test Plan:
- Validated YAML for all changed/new workflows.
- CI will exercise the new PT2E CPU/GPU workflows.

### DIFF
--- a/.github/workflows/nightly_smoke_test.yml
+++ b/.github/workflows/nightly_smoke_test.yml
@@ -39,4 +39,7 @@ jobs:
         pip install ${{ matrix.torch-spec }}
         pip install -r dev-requirements.txt
         pip install . --no-build-isolation
-        pytest test --verbose -s
+        # PT2E tests (test/quantization/pt2e) are excluded here and run in the
+        # dedicated regression_test_pt2e_cpu.yml / regression_test_pt2e_gpu.yml
+        # workflows instead.
+        pytest test --verbose -s --ignore=test/quantization/pt2e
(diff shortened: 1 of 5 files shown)


vkuzo:
* __->__ #4644

drisspg (approved):

jerryzh168 (approved):

Issue #4647, opened 2026-07-29
MXFP8 cutedsl kernels not compatible with nvidia-cutlass-dsl 4.6.0


andrewor14:
**Summary:** MXFP8 cutedsl kernels fail to compile under nvidia-cutlass-dsl 4.6.0 (required by vllm nightly `1.0.0.dev20260723`). The TMA verifier became stricter in 4.6.0 and now rejects the shared-memory layout that used in these kernels, failing at compile time:
- `mxfp8_quantize_cutedsl_3d`
-  `mxfp8_quantize_2d_1x32_cutedsl`
- `mxfp8_quantize_2d_32x1_cutedsl`

**Error:**
```
Verification failed: error: "tINs, tINg = cpasync.tma_partition(...):
'cute_nvgpu.atom.tma_partition' op tma_partition: shared-memory operand stride at mode 1 is
16 bits, not a multiple of the required 128-bit TMA smem alignment (element=16b, stride=1 elements).
```

**Minimal repro:** works with 4.5.0 but errors with 4.6.0
```
import torch
from torchao.prototype.moe_training.kernels.mxfp8 import mxfp8_quantize_cuda_3d
x = torch.randn(1, 32, 32, device="cuda", dtype=torch.bfloat16)
mxfp8_quantize_cuda_3d(
    x, 32, scale_block_dim1=32, scale_block_dim2=1, scaling_mode="rceil"
)
```

<details>
<summary>Claude analysis:</summary>

```
  Decoding the smem layout ((1,32),128):((4096,128),1):
  - mode 0 = (1,32) with strides (4096,128)
  - mode 1 = 128 with stride 1 ← the flagged one

  So the verifier is looking at mode 1 — a dimension of size 128 with element stride = 1 = 16 bits (one bf16) — and demanding that stride be a multiple of 128 
  bits (16 bytes). 16 is not a multiple of 128, so it errors.

  What it's actually rejecting

  It's rejecting a bare linear stride-1 (contiguous) dimension as the TMA smem operand's innermost mode. TMA addresses shared memory in 16-byte (128-bit) 
  granular units — the hardware's swizzle/box addressing steps in whole 16-byte chunks, so any stride the TMA descriptor uses must be ≥16 bytes and
  16-byte-aligned. A stride of a single 2-byte element is sub-granular — the TMA descriptor literally can't encode it.

[...]

PR #4649, opened 2026-07-29
Add _MXFP8GroupedMM variant with bf16 backward

andrewor14:
**Summary:**

Add `_MXFP8GroupedMMFwdBF16Bwd`, which does real MXFP8 forward but BF16 backward backward (straight-through), for QAT or RL trainer where a low-precision backward isn't needed. Selectable via a new `bf16_bwd` flag on `_to_mxfp8_then_scaled_grouped_mm` and `MXFP8TrainingOpConfig`. Also add KernelPreference.TRITON, which runs pure triton kernels for the forward quantization to work around an mxfp8 cutedsl issue (see pytorch/ao#4647).

Example usage:

```
from torchao.prototype.moe_training.config import MXFP8TrainingOpConfig
from torchao.prototype.mx_formats.config import ScaleCalculationMode
from torchao.quantization.quantize_.common import KernelPreference
from torchtitan.models.common.moe import GroupedExperts

config = MXFP8TrainingOpConfig(
    kernel_preference=KernelPreference.TRITON,
    scale_calculation_mode=ScaleCalculationMode.RCEIL,
    bf16_bwd=True,
)
quantize_(model, config=config, filter_fn=lambda m, fqn: isinstance(m, GroupedExperts))
```

**Test Plan:**

```
pytest test/prototype/moe_training/test_mxfp8_grouped_mm.py -k bf16_bwd_forward_matches_default
pytest test/prototype/moe_training/test_mxfp8_grouped_mm.py -k bf16_bwd_triton_matches_emulated
```

Real e2e RL tests TBD.

### DIFF
--- a/torchao/prototype/moe_training/config.py
+++ b/torchao/prototype/moe_training/config.py
@@ -170,6 +170,9 @@ class MXFP8TrainingOpConfig(TrainingOpBaseConfig):
     # Whether to pad the token group sizes to multiples of 32 (MXFP8 scaling block size).
     pad_token_groups_for_grouped_mm: bool = False
 
+    # QAT variant: quantize the forward to MXFP8 but compute the backward in BF16
+    bf16_bwd: bool = False
+
     @classmethod
     def from_recipe(
         cls,
@@ -212,6 +215,7 @@ class MXFP8TrainingOpConfig(TrainingOpBaseConfig):
                 and self.scale_calculation_mode == other.scale_calculation_mode
                 and self.pad_token_groups_for_grouped_mm
                 == other.pad_token_groups_for_grouped_mm
+                and self.bf16_bwd == other.bf16_bwd
             )
         return NotImplemented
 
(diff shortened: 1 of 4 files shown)


vkuzo:
why do we need cutedsl 4.6.0? can we use 4.5.2?

andrewor14:
> why do we need cutedsl 4.6.0? can we use 4.5.2?

It's because vllm nightly (since 7/16/26) has 4.6.0 has a hard dependency, and I feel it's [unlikely they'll downgrade](https://github.com/vllm-project/vllm/pull/47442). torchtitan RL is expected to work with vllm nightly

torchao/prototype/moe_training/config.py
vkuzo, line 165: @andrewor14 , i think instead of `bf16_bwd` you can set this flag to true and also add `dgrad_with_hp` to get the same effect but matching the existing code better.

andrewor14:
Going to close this to reopen with ghstack

PR #4652, opened 2026-07-30
float8: work around Triton fp8 store miscompile in compiled training

vkuzo:
Summary:
Compiled float8 training (e.g. FSDP2 + torch.compile) could intermittently
produce NaN/Inf losses. Root cause is a Triton miscompile: when a kernel
contains `min.NaN`/`max.NaN` PTX instructions, Triton mis-lowers a neighboring
transposed, vectorized 1-byte (fp8) store, writing garbage bytes that decode as
NaN/Inf. The computed value is correct (the contiguous store of the same value
is clean); only the transposed store is corrupted. Filed upstream as
https://github.com/triton-lang/triton/issues/11111.

The bug became visible after PyTorch inductor PR
https://github.com/pytorch/pytorch/pull/186933, which changed the Triton codegen
for min/max from the `tl.where`-based `triton_helpers.{minimum,maximum}` to
`tl.{minimum,maximum}(a, b, tl.PropagateNan.ALL)`. The two forms are numerically
identical, but only the `PropagateNan.ALL` form lowers to the `min.NaN`/`max.NaN`
instructions that trip the store bug.

Timeline: nightlies were GOOD through 2026-06-13 and BAD starting 2026-06-14,
bracketing #186933. The failure is flaky (~25%/run) because inductor autotunes
the fused kernel's launch config and only a subset of configs trip the store
miscompile; the miscompile itself is deterministic per config.

This change reverts the inductor min/max codegen back to the numerically-identical
`triton_helpers` (`tl.where`) form, which avoids emitting `min.NaN`/`max.NaN`.
It is applied from `convert_to_float8_training` so it only affects users of the
float8 training product, not every `import torchao`. The reverted form is always
correct, so this is a numeric no-op even after Triton fixes the underlying bug.

Test Plan:
Repro'd the FSDP2 float8 + torch.compile NaN in a standalone script; verified the
config that deterministically NaN'd now produces finite losses with the workaround
applied, and confirmed `convert_to_float8_training` (but not bare `import torchao`)
flips the codegen.

### DIFF
--- a/torchao/float8/_inductor_patch.py
+++ b/torchao/float8/_inductor_patch.py
@@ -1,0 +1,68 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+Workaround for a Triton miscompile that produces spurious NaN/Inf in compiled
+float8 backward kernels (e.g. FSDP2 float8 training with torch.compile).
+
+Root cause: PyTorch inductor PR pytorch/pytorch#186933 changed the Triton codegen
+for ``minimum``/``maximum`` from the ``tl.where``-based ``triton_helpers.{minimum,
+maximum}`` to ``tl.{minimum,maximum}(a, b, tl.PropagateNan.ALL)``. The two forms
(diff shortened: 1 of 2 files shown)


vkuzo:
* __->__ #4652

andrewor14:
@claude review

drisspg (approved):

PR #4653, opened 2026-07-30
mxfp8: use versioned TMA lookup with CUDA 13

lw:
The MXFP8 CUDA kernels cast cuTensorMapEncodeTiled to the v12.0 driver ABI, but cudaGetDriverEntryPoint chooses an ABI from the build-time runtime. On CUDA 13.2 that lookup returned cudaErrorInvalidValue before either quantization kernel could launch.

For CUDA 13 and newer, use cudaGetDriverEntryPointByVersion(..., 12000, ...) at both TMA lookup sites. This matches PFN_cuTensorMapEncodeTiled_v12000 and keeps the extension on the same stable ABI with newer toolkits. Retain cudaGetDriverEntryPoint for older toolkits, preserving the previous path where it already selects the expected ABI and avoiding a dependency on the newer versioned runtime API.


lw:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.16.0) (oldest at bottom):
* __->__ #4653

vkuzo:
@claude review

PR #4654, opened 2026-07-30
mxfp8: make RCEIL NaN handling CSE-friendly

lw:
RCEIL conversion currently creates a fresh float('nan') for each quantizer. NaN is non-reflexive, so AOTAutograd's FX CSE cannot treat otherwise identical projection casts as the same expression.

When the encoded E8M0 exponent is 255, descale is already NaN. Reuse that computed value for the reciprocal branch instead of creating a new scalar constant. This preserves the existing NaN propagation while making repeated quantizers structurally commonable.

Add an FX regression that runs the same CSE pass used by AOTAutograd and verifies repeated RCEIL branches share the computed NaN operand. The existing RCEIL numerical test continues to cover NaN and special-value output bytes.

### DIFF
--- a/torchao/prototype/mx_formats/mx_tensor.py
+++ b/torchao/prototype/mx_formats/mx_tensor.py
@@ -188,6 +188,7 @@ def _to_mx_rceil(
         # exponent is 255 exactly when descale is NaN. Reusing descale avoids
         # materializing distinct NaN constants that prevent graph CSE when the
         # same activation is quantized for multiple linear projections.
+        # See https://github.com/pytorch/pytorch/issues/191013.
         descale,
         torch.where(
             exponent == 254,  # Inf case -> return 2^-127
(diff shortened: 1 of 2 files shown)


lw:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.16.0) (oldest at bottom):
* __->__ #4654

vkuzo:
@claude review

torchao/prototype/mx_formats/mx_tensor.py
```diff
@@ -185,7 +185,11 @@ def _to_mx_rceil(
     # Ref: https://github.com/NVIDIA/TransformerEngine/blob/b7598aa887eb7d619d64c90692980009669379bf/transformer_engine/common/util/ptx.cuh#L332-L341
     rcp_fp32 = torch.where(
         exponent == 255,  # NaN case -> stays NaN
-        float("nan"),
+        # exponent is 255 exactly when descale is NaN. Reusing descale avoids
+        # materializing distinct NaN constants that prevent graph CSE when the
+        # same activation is quantized for multiple linear projections.
+        # See https://github.com/pytorch/pytorch/issues/191013.
+        descale,
```
vkuzo, line 192: nice

vkuzo (approved):

PR #4659, opened 2026-07-30
Add core support for static Range Quantization (SRQ) and sub-byte embedding optimizations

pculliton:
## Summary
This PR introduces generic support for static range activation scales (SRQ) and output scaling, along with row-level embedding lookups and selection dispatchers inside the TorchAO quantization workflows. These changes provide the necessary low-level primitives to load and execute pre-calibrated low-bit quantized weight subclasses (such as `int2` and `int4`) efficiently on CUDA.

---

## Motivation
Our primary motivation is to add int2 and SRQ support to TorchAO.

To that end, this PR adds generic primitive support inside `IntxUnpackedToInt8Tensor` and `Int8Tensor` for:
1.  **Static activation and output range scaling** in linear layers.
2.  **Row-level embedding dequantization slicing** to keep memory footprints low.
3.  **Basic row-indexing (`aten.select.int`)** support on subclasses.

---

### 1. Core Subclass & Primitive Updates (`torchao/quantization/quantize_/workflows`)
*   **`intx/intx_unpacked_to_int8_tensor.py`**:
    *   Added `INT8_SYM_STATIC` to the `IntxUnpackedToInt8TensorActivationQuantization` configuration enum.
    *   Added `act_quant_scale` and `output_quant_scale` parameters to the `IntxUnpackedToInt8Tensor` constructor to hold static calibration scales.
    *   Implemented `aten.embedding.default` and `aten.select.int` dispatcher overrides for subclasses with row-wise granularity (`block_size[0] == 1`). Instead of dequantizing the entire weight matrix, it slices the quantized data first and dequantizes only the accessed rows, preventing OOM.
    *   Dispatched compile runs of linear layers with static activation scales to the optimized `int_scaled_matmul` CUDA kernel.
*   **`int8/int8_tensor.py`**:
    *   Added `output_quant_scale` support to the constructor and static linear dispatcher (applying quantize-dequantize scaling to linear output).
    *   Merged output scale handling with the recently introduced upstream `reduce_range` logic.

### 2. Configuration & Serialization Core
*   **`quant_api.py`**:
    *   Exposed and registered `Int8StaticActivationIntxWeightConfig` in the public config API.
    *   Updated tensor config constructors to automatically cast serialized list formats back to `torch.Tensor` during deserialization loading.
*   **`core/config.py`**:
    *   Updated `ConfigJSONEncoder` to serialize `torch.Tensor` parameters (such as activation scales) into JSON-compatible lists.
*   **`prototype/safetensors/safetensors_utils.py`**:
    *   Registered `IntxUnpackedToInt8TensorActivationQuantization` in `ALLOWED_CLASSES` to enable safe model serialization.

### 3. Unit Tests (`test/quantization/quantize_/workflows/intx`)
*   **`test_intx_unpacked_to_int8_tensor.py`**:
    *   Added `test_int8_static_activation_intx_weight_config` to verify compilation correctness of static range activation configurations.
    *   Added `test_intx_unpacked_embedding_sliced` to verify row-level embedding lookups.
    *   Added `test_select` to verify indexing slice dispatch.

---

## Verification
All changes were verified locally using the standard TorchAO unit testing pipeline, as well as our added tests. All 232 test cases passed successfully:
```bash
python test/quantization/quantize_/workflows/intx/test_intx_unpacked_to_int8_tensor.py
```


torchao/quantization/quantize_/workflows/intx/intx_unpacked_to_int8_tensor.py
```diff
@@ -322,20 +379,99 @@ def _(func, types, args, kwargs):
...
+                x_scales_expanded = (
+                    act_quant_scale.view(-1, 1).expand(tmp.shape[0], 1).contiguous()
+                )
+                intermediate_dtype = (
+                    torch.float
+                    if act_quant_scale.dtype == torch.half
+                    else act_quant_scale.dtype
+                )
+
+                y_dot_scaled = int_scaled_matmul(
```
metascroy: Why not make this cuda usecase its own tensor subclass? It seems different enough from the rest of the intx tensor?

metascroy (changes requested):

torchao/quantization/quantize_/workflows/intx/intx_unpacked_to_int8_tensor.py
pculliton: Hmm, I think that would make sense, thanks! I'll split it up.

PR #4661, opened 2026-07-30
fix h100 sparse CI

vkuzo:
The `1xH100` CI leg was failing the float8 sparse cutlass tests with
  `Could not run 'torchao::to_sparse_semi_structured_cutlass_sm9x_f8' with
  arguments from the 'CUDA' backend` — the compiled CUDA op was never registered.

  Root cause is an import-path/packaging interaction, not an ABI or code bug:

  - CI installs torchao with a **non-editable** `pip install .`, which puts the
    compiled extensions (`_C_cutlass_90a.abi3.so`, etc.) only in the site-packages
    wheel, leaving the source checkout's `torchao/` dir without any `.so`.
  - pytest is run from the repo root. Because `test/` is a package (`test/__init__.py`)
    and #4651 added `test/conftest.py`, pytest must import `test.conftest`, which
    prepends the **repo root** to `sys.path`. That makes `import torchao` resolve to
    the source tree, shadowing the installed wheel.
  - The source `torchao/` has no `.so`, so the loader's `_C*.so` glob is empty, the
    extension is never loaded, and the CUDA op is never registered.

  ## Fix
  
  Switch the H100 workflow to an editable install (`pip install -e .`). Editable runs
  `build_ext --inplace`, dropping the compiled extensions into the source tree that
  pytest actually imports — so the shadowing becomes harmless and the CUDA ops load.

  ## Test plan
  
  CI: the float8 sparse cutlass tests now run (instead of erroring/skipping) on the
  `1xH100` leg.

### DIFF
--- a/.github/workflows/1xH100_tests.yml
+++ b/.github/workflows/1xH100_tests.yml
@@ -69,7 +69,15 @@ jobs:
         pip install apache-tvm-ffi
         pip install nvidia-ml-py
         uv pip install -r dev-requirements.txt
-        pip install . --no-build-isolation -vv
+        # Editable install so build_ext --inplace drops the compiled extensions
+        # (e.g. _C_cutlass_90a.abi3.so) into the in-tree torchao/ dir. pytest is run
+        # from the repo root and, because test/ is a package with a conftest.py,
+        # prepends the repo root to sys.path -- so `import torchao` resolves to the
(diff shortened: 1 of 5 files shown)


vkuzo:
* __->__ #4661

andrewor14 (approved):

PR #4526, opened 2026-06-23
[test][xpu] Device agnostic test_config.py, test_model_architecture.py and test_utils.py

zxd1997066:
This PR is target to device agnostic test_utils.py and add test_config.py, test_model_architecture.py and test_utils.py in intel GPU CI

PR #4664, opened 2026-07-30
make tensor subclass copy_ error report exactly which metadata mismatched

### LINKED ISSUE #3046: make it easier to debug tensor subclass metadata mismatch
I'm hacking on a vLLM PoC, and I saw a stack trace like below:

```python
[rank0]: ValueError: Not supported args for copy_ due to metadata mismatch: (Float8Tensor(self.act_quant_kwargs=None, self.qdata=tensor([[nan, nan, nan,  ..., nan, nan, nan],
[rank0]:         [nan, nan, nan,  ..., nan, nan, nan],
[rank0]:         [nan, nan, nan,  ..., nan, nan, nan],
[rank0]:         ...,
[rank0]:         [nan, nan, nan,  ..., nan, nan, nan],
[rank0]:         [nan, nan, nan,  ..., nan, nan, nan],
[rank0]:         [nan, nan, nan,  ..., nan, nan, nan]], device='cuda:0',
[rank0]:        dtype=torch.float8_e4m3fn), self.scale=tensor([[0.],
[rank0]:         [0.],
[rank0]:         [0.],
[rank0]:         ...,
[rank0]:         [0.],
[rank0]:         [0.],
[rank0]:         [0.]], device='cuda:0', dtype=torch.float32), self.block_size=[1, 2048], self.mm_config=Float8MMConfig(emulate=False, use_fast_accum=True, pad_inner_dim=False), self.kernel_preference=<KernelPreference.AUTO: 'auto'> self.shape=torch.Size([1408, 2048]), self.device=device(type='cuda', index=0), self.dtype=torch.bfloat16), Float8Tensor(self.act_quant_kwargs=None, self.qdata=tensor([[ -24.,  -18.,   44.,  ...,  104., -320., -176.],
[rank0]:         [ 160.,  128.,  128.,  ...,  112.,  -14.,   44.],
[rank0]:         [ -22.,   13.,   22.,  ...,   48.,  -13.,   72.],
[rank0]:         ...,
[rank0]:         [ -56., -112., -352.,  ..., -256.,   20.,   96.],
[rank0]:         [ 144.,  224.,  -52.,  ...,   40., -208.,  -15.],
[rank0]:         [-144.,  240.,  224.,  ...,  224.,   26.,  -80.]],
[rank0]:        dtype=torch.float8_e4m3fn), self.scale=tensor([[0.0001],
[rank0]:         [0.0001],
[rank0]:         [0.0001],
[rank0]:         ...,
[rank0]:         [0.0001],
[rank0]:         [0.0001],
[rank0]:         [0.0002]], dtype=torch.float32), self.block_size=[1, 2048], self.mm_config=Float8MMConfig(emulate=False, use_fast_accum=True, pad_inner_dim=False), self.kernel_preference=<KernelPreference.AUTO: 'auto'> self.shape=torch.Size([1408, 2048]), self.device=device(type='cpu'), self.dtype=torch.bfloat16))

```

It would be great if torchao told me exactly which attribute is not matching instead of me having to parse the wall of text.

sainikhiljuluri:
## Summary

`TorchAOBaseTensor`'s generic `aten.copy_.default` handler raised a `ValueError` that interpolated both tensors' full `repr`s — multi-KB walls of tensor data — without saying which metadata field actually mismatched. This PR makes the error name exactly the mismatched fields with concise values, e.g.:

```
ValueError: Not supported args for copy_ due to metadata mismatch: shape: torch.Size([256, 128]) vs torch.Size([256, 64]); qdata.shape: torch.Size([256, 128]) vs torch.Size([256, 64]); block_size: [1, 128] vs [1, 64]
```

Fixes #3046

## What changed

- `torchao/utils.py`: factored the comparisons of `_same_metadata` into `_metadata_mismatch_details(self, src) -> List[str]`, which returns a human-readable entry per mismatched field (type, outer shape, `tensor_data_names` shapes, `optional_tensor_data_names` presence/shapes, `tensor_attribute_names` / `optional_tensor_attribute_names` values). `_same_metadata` is now `not _metadata_mismatch_details(...)` (same name, same boolean semantics), and the `copy_` handler reports the joined details instead of dumping both tensors.
- The comparison semantics are unchanged, except two cases that previously crashed with `AttributeError` before reaching the `ValueError` at all:
  - `self`'s optional tensor set while `src`'s is `None` (`None.shape`) — now reported as `zero_point: Tensor vs None`;
  - type mismatch between structurally different subclasses (`getattr(src, ...)` on a missing field) — now reported as `type: A vs B` via an early return.
- `test/test_utils.py`: 8 new CPU tests (no CUDA gate): attribute mismatch names the field and both values (and the message contains no tensor dumps), type mismatch, inner-tensor shape mismatch with equal outer shapes (packed-format case), outer-shape mismatch with equal inner shapes (reported on its own, not conflated with `qdata.shape`), optional-tensor `None` vs set in both directions, multiple mismatches all reported, matching-metadata `copy_` still succeeds, and an end-to-end case on a shipped subclass rather than a test-local one.

## Test plan

```
pytest test/test_utils.py -q
# before: 9 passed, 3 skipped  |  after: 17 passed, 3 skipped
# TDD: the 7 new error-message tests were written first and verified failing on main
# (5 on the old wall-of-text message, 2 on the pre-existing AttributeError crashes)

ruff check --isolated --select F821,F823,W191 && ruff check && ruff format --check
# all pass
```

The end-to-end check on a real subclass is now a test rather than a manual step:
`test_copy__real_tensor_subclass_names_the_single_field` builds two `Int8Tensor`s on
CPU differing only in `block_size` and asserts the three properties #3046 asks for —
the differing field is named, the message contains no tensor data, and it stays short
(83 chars, against the multi-KB dump in the issue).

## Risk / BC notes

- The error message prefix `Not supported args for copy_ due to metadata mismatch:` is preserved; only the payload after it changed. This exact string appears nowhere else in the repo, so no other tests depend on it.
- `NVFP4Tensor` has its own separate `_same_metadata`/`copy_` implementation (`torchao/prototype/mx_formats/nvfp4_tensor.py`) and is intentionally not touched here; can follow up if wanted.

### A note on `device`

The trace in #3046 ends with `self.device=device(type='cuda', index=0)` against
`device(type='cpu')`, so it seems worth being explicit: **`device` is not part of the
metadata `_same_metadata` compares — neither before nor after this PR.** The comparison
is driven by `tensor_data_names` / `tensor_attribute_names`, and e.g.
`Float8Tensor.tensor_attribute_names` is `[]`. With only `.device` differing, the check
finds no mismatch and the `copy_` proceeds.

Making a device difference a reported mismatch would be a behaviour change, not a
message change — cross-device `copy_` is legal today and would start raising — so this
PR deliberately leaves the comparison set alone and only changes how an already-detected
mismatch is *reported*. Happy to add device as a separate PR if you'd prefer those
semantics.

---

*This PR was developed with AI assistance.*

### DIFF
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -9,6 +9,7 @@ from unittest.mock import patch
 import torch
 import torch.nn.functional as F
 
+from torchao.quantization import Int8Tensor
 from torchao.testing.utils import skip_if_no_cuda
 from torchao.utils import TorchAOBaseTensor, torch_version_at_least
 
@@ -399,9 +400,8 @@ class TestTorchAOBaseTensor(unittest.TestCase):
         with self.assertRaisesRegex(ValueError, r"type: MyTensor vs OtherTensor"):
             lp_tensor.copy_(other_tensor)
 
-    def test_copy__tensor_shape_mismatch_error_names_tensor(self):
-        """Inner tensor shape mismatch should be reported by name even when
-        the outer shapes match (e.g. packed formats)"""
+    def _get_packed_copy_test_tensor_class(self):
+        """Outer shape is decoupled from the inner tensor shape, as in packed formats"""
 
         class MyPackedTensor(TorchAOBaseTensor):
             tensor_data_names = ["qdata"]
@@ -418,10 +418,31 @@ class TestTorchAOBaseTensor(unittest.TestCase):
             def __init__(self, qdata, shape, device=None):
                 pass
 
+        return MyPackedTensor
+
+    def test_copy__tensor_shape_mismatch_error_names_tensor(self):
+        """Inner tensor shape mismatch should be reported by name even when
+        the outer shapes match (e.g. packed formats)"""
+        MyPackedTensor = self._get_packed_copy_test_tensor_class()
         lp_tensor = MyPackedTensor(torch.randn(2), torch.Size([4]))
         lp_tensor_for_copy = MyPackedTensor(torch.randn(4), torch.Size([4]))
         with self.assertRaisesRegex(
             ValueError, r"qdata\.shape: torch\.Size\(\[2\]\) vs torch\.Size\(\[4\]\)"
+        ):
+            lp_tensor.copy_(lp_tensor_for_copy)
+
+    def test_copy__outer_shape_mismatch_reported_on_its_own(self):
+        """An outer shape mismatch is reported by itself when the inner tensors agree.
+
+        Packed formats decouple the two, so `shape` and `qdata.shape` are independent
+        signals and the message must not conflate them.
+        """
+        MyPackedTensor = self._get_packed_copy_test_tensor_class()
+        lp_tensor = MyPackedTensor(torch.randn(4), torch.Size([4]))
+        lp_tensor_for_copy = MyPackedTensor(torch.randn(4), torch.Size([2, 2]))
+        with self.assertRaisesRegex(
+            ValueError,
+            r"metadata mismatch: shape: torch\.Size\(\[4\]\) vs torch\.Size\(\[2, 2\]\)$",
         ):
             lp_tensor.copy_(lp_tensor_for_copy)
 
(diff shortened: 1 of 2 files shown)


sainikhiljuluri:
Rebased onto `main` to clear a conflict introduced by #4526, which made `test_utils.py` device-agnostic and dropped the `skip_if_no_cuda` import. Resolved by keeping that removal — the tests added here don't need a device — so the only import this adds is `Int8Tensor`. `test/test_utils.py` passes 20/20 locally.

This is waiting on a maintainer releasing the `action_required` workflows so CI can run. Happy to adjust anything once it's green.

PR #4665, opened 2026-07-31
Fold quantization into mutable buffer storage (#4665)

ethansfng:
Summary:

Add an opt-in `fold_quantize_into_mutable_buffers=False` argument to `convert_pt2e`. When enabled with `fold_quantize`, it recognizes registered floating-point state with an explicit buffer Q/DQ read boundary and exactly one terminal `copy_` sourced from a matching static Q/DQ write boundary. The write no longer has to be reached through a single-user read-to-write path, so the dequantized buffer read may fan out through arbitrary floating-point computation. The matcher does not infer mutation from operator schemas or synthesize a writeback.

Use the terminal write qparams as the canonical storage domain. Quantize the backing buffer in that domain, retarget the input DQ to the integer buffer, and preserve the output Q as the value written by `copy_`. Reject aliased buffers and incompatible tensor metadata, update FX metadata, and propagate the option into supported control-flow subgraphs. The option remains disabled by default and requires `fold_quantize=True`.

Reviewed By: DrJessop

Differential Revision: D113826710

### DIFF
--- a/test/quantization/pt2e/test_quantize_pt2e.py
+++ b/test/quantization/pt2e/test_quantize_pt2e.py
@@ -3435,6 +3435,372 @@ class TestQuantizePT2E(PT2EQuantizationTestCase):
             ViewM(), example_inputs, {"select_scatter": "buf"}
         )
 
+    def _prepare_mutable_buffer_quantization(
+        self,
+        module: torch.nn.Module,
+        example_inputs: tuple[torch.Tensor, torch.Tensor],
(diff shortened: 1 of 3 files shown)


vkuzo:
@claude review

vkuzo:
@claude review

vkuzo:
@claude explain what the PR is doing, include an ascii diagram of before vs after

vkuzo (approved):

PR #4686, opened 2026-08-04
Expose compile_backend for low-bit Adam optimizers

### LINKED ISSUE #955: [MPS] torchao low-bit-precision optim does not expose 'backend' argument to torch.compile
on apple mps platforms, torchao training works great until we involve the AdamW8bit optimiser:

```
    assert wrapper_code_gen_cls is not None, f"Device {device_type} not supported"
torch._dynamo.exc.BackendCompilerFailed: backend='inductor' raised:
AssertionError: Device mps not supported

Set TORCH_LOGS="+dynamo" and TORCHDYNAMO_VERBOSE=1 for more information
```

and conditionally setting:

```py
    if torch.backends.mps.is_available():
        import torch._dynamo
        torch._dynamo.config.suppress_errors = True
```

to suppress this error and 'fallback to eager mode' does not work in this situation, it merely hides the notice that this parameter could be set.

using `aot_eager` is required instead of `inductor` for MPS, but additionally it kinda limits third-party backends.

Can we expose the `backend` parameter, perhaps as `compile_backend` ?

Abhisri436:
## What does this PR do?

This PR adds an optional `compile_backend` argument to torchao’s low-bit Adam optimizers. When provided, the value is forwarded to the internal `torch.compile` call as the `backend` argument. When it is not provided, the existing default behavior is preserved and no `backend` argument is passed.

## Why was this PR needed?

Issue #955 requests a way for torchao low-bit optimizers to expose the `backend` argument used by `torch.compile`. During investigation, I found that the shared `_AdamBase.step()` path calls `torch.compile(single_param_adam, fullgraph=True, dynamic=False)` directly, while the public optimizer constructors do not accept a backend option. This means users are locked into the default compile backend, which can be a problem on platforms such as MPS where an alternative backend like `aot_eager` may be needed.

## What changed?

- Added `compile_backend=None` to the low-bit Adam optimizer path in `torchao/optim/adam.py`
- Forwarded `compile_backend` through the public optimizer constructors:
  - `Adam8bit`
  - `AdamW8bit`
  - `Adam4bit`
  - `AdamW4bit`
  - `AdamFp8`
  - `AdamWFp8`
- Passed `backend=self.compile_backend` to `torch.compile` only when `compile_backend` is provided
- Added tests for constructor support, backend forwarding, and preserved default behavior
- Updated `torchao/optim/README.md` with a short usage note

## What are the relevant issue numbers?

Closes #955

## Testing

I ran the following checks locally:

```bash
ruff check --isolated --select F821,F823,W191
ruff check torchao/optim/adam.py test/test_low_bit_optim.py
ruff format --check torchao/optim/adam.py test/test_low_bit_optim.py
python -m pytest test/test_low_bit_optim.py -k "compile_backend or smoke" -q
python -m pytest test/test_low_bit_optim.py -q

### DIFF
--- a/test/test_low_bit_optim.py
+++ b/test/test_low_bit_optim.py
@@ -7,6 +7,7 @@ import copy
 import shutil
 import tempfile
 from pathlib import Path
+from unittest.mock import patch
 
 import pytest
 import torch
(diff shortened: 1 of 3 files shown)


gau-nernst:
This does not resolve #955. It merely adds cosmetics. aot_eager is only for debugging, it does not codegen fused kernels

Abhisri436:
Hi, thank you for the feedback and clarification. I had interpreted #955 as an API-forwarding issue for exposing the `torch.compile` backend argument, but I understand now that this does not address the underlying fused-kernel/codegen problem.

PR #4709, opened 2026-08-06
Enable rowwise MXFP8 CUDA quantization

lw:


lw:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.17.0) (oldest at bottom):
* #4725
* #4733
* #4732
* __->__ #4709

torchao/csrc/cuda/mx_kernels/mxfp8_extension.cpp
```diff
@@ -112,7 +112,6 @@ mxfp8_quantize(const Tensor& input, bool rowwise, bool colwise,
                const std::string &scaling_mode) {
 
   // Validate inputs
-  STD_TORCH_CHECK(!rowwise, "rowwise scaling is not supported yet");
```
lw, line 115: This is the core reason behind this change. I found that rowwise was effectively implemented by the kernel, but disabled in the entry point. I found that odd, but please let me know if there's a reason.

vkuzo:
@claude review

test/prototype/mx_formats/test_kernels.py
```diff
@@ -566,38 +566,62 @@ def test_rearrange(shape):
...
-    torch.testing.assert_close(s_d1, s_d1_ref, rtol=0, atol=0)
-
-    # check quantized values
-    torch.testing.assert_close(y_d1, y_d1_ref, rtol=0, atol=0)
-    assert y_d1.stride() == y_d1_ref.stride(), "quantized tensor strides do not match"
+    if rowwise:
+        # Rowwise uses independent 1x32 groups, including when colwise is also
+        # requested from the fused kernel.
+        torch.testing.assert_close(s_d0, s_d0_ref, rtol=0, atol=0)
+        torch.testing.assert_close(y_d0, y_d0_ref, rtol=0, atol=0)
```
vkuzo, line 610: lg

vkuzo (approved):
lg if claude + local testing + perf all pass!

vkuzo:
@claude review

lw:
Did some local perf benchmarks:
- the old colwise-only case remains stable (within noise: measured a +0.03% latency)
- the rowwise and colwise+rowwise cases (which were disabled before) have been slowed down by the extra bounds check I added (between 1% and 5% slower), but that seems unavoidable for now, hence no way around that cost

PR #4710, opened 2026-08-06
Fix flaky test_fp32_fp8_multi_module_parity

### LINKED ISSUE #4708: test_fp32_fp8_multi_module_parity (FSDP2 float8): NaN/0.0 loss parity mismatch on 4xH100 (torch 2.13.0)
Commits affected: `ff9db6ed5a4805465237cf8e81721ee2ea2d557a`
Run: https://github.com/pytorch/ao/actions/runs/31037168764

## Summary
`test_fp32_fp8_multi_module_parity` in `test/float8/test_fsdp2/test_fsdp2.py` fails a float8-vs-fp32 loss parity assertion at training iteration 4 with degenerate, rank-inconsistent loss values (NaN / 0.0) — a numerical-instability signature unrelated to the triggering commit.

## Failure details
- Failed job(s): `Run 4xH100 tests` → `test (H100 2.13, mt-l-x86iamx-88-900-h100-4, torch==2.13.0 torchvision==0.28.0 --index-url .../cu126)`. The sibling nightly-torch job (`--pre torch`) in the same run **passed**.
- Failing test: `TestFloat8MultiThread.test_fp32_fp8_multi_module_parity` (1 error out of 8 tests in `test/float8/test_everything_multi_gpu.sh`).
- Representative error (`torchao/testing/training/fsdp2_utils.py:59`, `check_parity_no_mp`):
  ```
  AssertionError: Scalars are not close!
  Thread 0: Expected nan but got 0.0    (iter: 4, loss-ref: 0.0,       loss-fp8: nan)
  Thread 1: Expected 0.0 but got 0.0005306489765644073
                                        (iter: 4, loss-ref: 0.0005306, loss-fp8: 0.0)
  ```
  Job then exits 1: `RuntimeError: Command bash /__w/_temp/exec_script failed with exit code 1`.

## Root cause
The reference (fp32) and fp8 losses collapse to degenerate values (0.0 and NaN) at iter 4, and the two threads report mutually inconsistent reference losses. Losses that are simultaneously zero/NaN and disagree across ranks indicate numerical instability in the fp8 parity check rather than a deterministic computational error.

The triggering commit `ff9db6e` ("Remove deprecated torchao/dtypes package") only deletes the deprecated `torchao/dtypes` package (edits limited to `torchao/__init__.py`, `torchao/core/config.py`, `torchao/dtypes/__init__.py`, `torchao/prototype/mx_formats/kernels.py`, and one prototype script). It does not touch any float8 or FSDP2 training code, and nothing under `torchao/float8` or `torchao/testing/training` imports `torchao.dtypes`. The immediate parent commit (`ac1a803c60`) and the preceding ~14 runs of this workflow all passed, so the failure does not track the code change.

## Classification
flaky test — justification: (1) the failing code path is untouched by the commit; (2) parent and prior runs are green; (3) the assertion values are degenerate and rank-inconsistent (NaN vs 0.0), the classic signature of fp8 numerical noise rather than a reproducible regression; (4) the parallel nightly-torch job passed on the identical commit.

## Suggested fix
- Rerun the job to confirm non-determinism (expected to pass on retry).
- If it recurs, harden `test_fp32_fp8_multi_module_parity` against fp8 numerical instability: seed the RNG deterministically before model/data init, and/or guard the parity `assertEqual` in `torchao/testing/training/fsdp2_utils.py` against NaN/0.0 degenerate losses (skip or fail with a clear message when reference loss is 0.0/NaN, since relative-difference is `inf` there). Consider slightly relaxed tolerances or more iterations before comparison to avoid the early-iteration degenerate-loss window.

andrewor14:
**Summary:** Fixes #4708, where the fp8 loss intermittently goes NaN due to a test-only race condition in the multi-threaded process group:

1. FSDPTestMultiThread runs each rank as a thread on one device with its own FSDP stream
2. ProcessLocalGroup reduces on rank 0's stream without waiting on the others
3. fp8 amax all-reduce in fsdp_pre_all_gather reads a value whose kernel hasn't run, yielding NaN losses

This is a test-only failure. With a real NCCL process group, collectives are stream-ordered, so the reduce can't run ahead of the amax that feeds it. The multi-process tests covering the same fp8 all-gather path are unaffected.

Fix: serialize FSDP2's communication streams onto the current stream for TestFloat8MultiThread.

**Test Plan:**

```
pytest test/float8/test_fsdp2/test_fsdp2.py
```

### DIFF
--- a/test/float8/test_fsdp2/test_fsdp2.py
+++ b/test/float8/test_fsdp2/test_fsdp2.py
@@ -54,6 +54,62 @@ if not (is_sm_at_least_89() or is_MI300() or is_MI350()):
     )
 
 
+def _serialize_fsdp_comm_streams(test_cls) -> None:
+    """
+    Serialize all FSDP2's communication streams for this test to bypass the
+    following race condition:
+
+      1. FSDPTestMultiThread runs each rank as a thread on one device, and each
+         thread gets its own FSDP all-gather copy-in stream.
+      2. With fp8 all-gather, torchao computes the weight amax in
+         fsdp_pre_all_gather, so it lands on that per-thread stream.
+      3. ProcessLocalGroup reduces on rank 0's thread and stream, ordered only
+         by CPU condition variables; it never waits on the other thread's
+         stream.
+      4. Rank 0 can therefore read an amax whose kernel has not run yet, giving
+         a garbage scale and a NaN loss.
+
+    We achieve this by pointing all streams to the current stream, which
+    restores the ordering.
+
+    Note: This uses a private torch.distributed.fsdp API. If upstream changes
+    we will fail loudly here instead of silently allowing the race condition to
+    resurface.
+    """
+    try:
+        from torch.distributed.fsdp._fully_shard._fsdp_param_group import (
+            FSDPCommContext,
+        )
+    except ImportError as e:
+        raise ImportError(
+            "Cannot import FSDPCommContext to serialize FSDP2's communication "
+            "streams. This private API moved; update _serialize_fsdp_comm_streams "
+            "or these multi-threaded tests will race on the fp8 amax all-reduce."
+        ) from e
+
+    orig_lazy_init = FSDPCommContext.lazy_init
+
+    def serialized_lazy_init(self, device: torch.device) -> None:
+        orig_lazy_init(self, device)
+        current_stream = self.device_handle.current_stream()
+        required_streams = (
+            "all_gather_copy_in_stream",
+            "all_gather_stream",
+            "reduce_scatter_stream",
+        )
+        missing = [attr for attr in required_streams if not hasattr(self, attr)]
+        if missing:
+            raise AttributeError(
+                f"FSDPCommContext is missing {missing}; update "
+                "_serialize_fsdp_comm_streams for this version of torch."
+            )
+        for attr in required_streams:
+            setattr(self, attr, current_stream)
+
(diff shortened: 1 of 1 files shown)


andrewor14:
@claude review, look at https://github.com/pytorch/ao/issues/4708 for additional context

test/float8/test_fsdp2/test_fsdp2.py
```diff
@@ -54,6 +54,62 @@
...
+
+    We achieve this by pointing all streams to the current stream, which
+    restores the ordering.
+
+    Note: This uses a private torch.distributed.fsdp API. If upstream changes
+    we will fail loudly here instead of silently allowing the race condition to
+    resurface.
+    """
+    try:
+        from torch.distributed.fsdp._fully_shard._fsdp_param_group import (
```
vkuzo, line 80: is this a new api in nightly? also, is the regression on stable, nightly or both?

vkuzo (approved):

test/float8/test_fsdp2/test_fsdp2.py
andrewor14: seems like this API existed since 2.6 and has been unchanged since then. I think the regression is on both stable and nightly

vkuzo: sg, in that case you could just import it directly without the try/except

PR #4714, opened 2026-08-06
Fix float8 zero scale quantization

### LINKED ISSUE #4713: Float8Tensor zero inputs quantize to NaN
### Description

Quantizing an all-zero tensor through the public Float8Tensor API currently produces NaN float8 data and NaN dequantized output.

This happens because  computes . For an all-zero block,  is 0, so  is 0, and the following affine float8 quantization path divides 0 by 0.

### Actual behavior

Both  and dequantized output contain NaN values.

### Expected behavior

Zero input should quantize to finite zero float8 data, and dequantize back to all zeros.

### Notes

The nearby floatx scale path already clamps amax to a small positive floor before dividing by the float max. Applying the same guard to  fixes this without changing non-zero normal inputs.

Rajioba1:
## Summary

Fixes #4713.

Float8Tensor.from_hp currently produces NaN qdata and NaN dequantized output for all-zero tensors because _choose_scale_float8 returns a zero scale for zero blocks. The later affine float8 quantization path divides 0 by 0.

This clamps the amax value to a small positive floor before dividing by the float8 max, matching the nearby floatx scale helper pattern.

## Test Plan

- CUDA_VISIBLE_DEVICES="" python -m pytest -q test/quantization/quantize_/workflows/float8/test_float8_tensor.py::TestFloat8TensorCPU -q
- CUDA_VISIBLE_DEVICES="" python -m pytest -q test/quantization/test_quant_primitives.py -q
- ruff check torchao/quantization/quant_primitives.py test/quantization/quantize_/workflows/float8/test_float8_tensor.py
- ruff format --check torchao/quantization/quant_primitives.py test/quantization/quantize_/workflows/float8/test_float8_tensor.py

### DIFF
--- a/test/quantization/quantize_/workflows/float8/test_float8_tensor.py
+++ b/test/quantization/quantize_/workflows/float8/test_float8_tensor.py
@@ -193,6 +193,19 @@ class ToyLoRAModel(torch.nn.Module):
 
 
 # TODO: move tests in test_affine_quantized_float.py here after we migrated all implementations
+class TestFloat8TensorCPU(common_utils.TestCase):
+    @common_utils.parametrize("granularity", [PerTensor(), PerRow()])
+    def test_zero_tensor_quantizes_to_finite_zero(self, granularity):
+        x = torch.zeros(2, 4, dtype=torch.float32)
+
+        x_fp8 = Float8Tensor.from_hp(x, torch.float8_e4m3fn, granularity)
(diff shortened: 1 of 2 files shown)


Rajioba1:
I do not have permission to apply upstream labels from the fork. Could a maintainer please add the appropriate module label? I believe module: core fits this quant primitive / Float8Tensor change.

Rajioba1:
I have signed the CLA

PR #4725, opened 2026-08-07
Refine MXFP8 CUDA quantization semantics

lw:
Here's the steps of quantization, with the respective options and pitfalls.

1. **Compute the abs-max of the quantization tile**
   The main decision is what to do with NaNs: propagate them (one NaN "pollutes" the whole tile) or ignore them (amax is NaN only if the whole tile is NaN). CUDA and Triton default to ignore, PyTorch and CuTeDSL default to propagate. Here I opted to go for propagate, for a bunch of reasons:
   - It allows for a consistent behavior with fp4, where a NaN in the high-precision input must cause a NaN to appear in the scale, as the individual fp4 (or even fp6) output element cannot encode a NaN
   - It means the behavior is consistent and easily-reproducible with the standard PyTorch operators
   
   (Note, an earlier version of this PR made the opposite choice and went for "ignore", but I changed my mind)

2. **Convert the amax / fp8_max to a ue8m0 scale**
   - We need to decide what to do about Inf. Here I decided to make it produce a scale of NaN, as saturating it to the largest scale might end up hiding a numerical bug, and would cause an infinitely large quantization error.
   - Another catch is to make sure to handle subnormals correctly, since the lowest ue8m0 value is just outside the range of normals.
   - Also note that doing this via log2+ceil (or floor) is mathematically correct but numerically incorrect because of some inaccuracies of these functions (which are also quite expensive on GPU!).

   The best approach seems to be bit-manipulation (or, on Blackwell GPUs, just relying on the built-in PTX, albeit with the correct modifiers).

3. **Taking the reciprocal of the scale**
   NaNs must pass-through unchanged. And again there's the subnormal value (corresponding to 2^-127) which might need special care. A safe approach is to reinterpret this as 254 - x in uint8.

4. **Converting the reciprocal back to a fp32**
   Pretty much the symmetrical problems: handling NaN, mapping 2^-127 to a subnormal, avoiding exp2 due to numerical issues (and slowness). And again the solution is bit-manipulation or PTX builtins.

5. **Converting data to e4m3 or e5m2**
   This is pretty straightforward if we use the right builtins. One catch is that the floor mode needs to do clamping. It might be done automatically by the conversion if we use saturation, but if this isn't the case we need to do it manually.

I'd appreciate some discussion of the above decisions, so that we get to a large consensus. The one I'm most unsure about is the ignore-vs-propagate for NaNs.

I encoded this semantic on all the backends: torch (eager + compiled), CUDA C++, Triton, CuTeDSL. The FlyDSL changes are agent-authored and unreviewed!

I also added extensive testing for all the delicate edge cases that I found. In principle every change I did would now cause a test to fail if it were reverted.

I added some inline comments in this PR to highlight some of the above issues, but didn't do it with perfect coverage.


lw:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.17.0) (oldest at bottom):
* __->__ #4725

torchao/csrc/cuda/mx_kernels/mxfp8_cuda.cu
```diff
@@ -26,8 +26,6 @@ DType get_input_dtype(const Tensor &t) {
   switch (t.scalar_type()) {
   case torch::headeronly::ScalarType::Float:
     return DType::kFloat32;
-  case torch::headeronly::ScalarType::Half:
-    return DType::kFloat16;
```
lw: Oops, spurious changes, will move them to another PR. The goal here was to remove fp16 support, which is advertised at a high-level but unsupported in the kernel.

torchao/csrc/cuda/mx_kernels/mxfp8_quantize.cuh
```diff
@@ -123,29 +124,36 @@ template <> struct DataTypeTraits<nv_bfloat16> {
   }
 };
 
-__device__ static __forceinline__ e8m0_t
-calculate_e8m0_biased_scale(const float amax) {
-  // torchao ref:
-  // https://github.com/pytorch/ao/blob/00417b8b33abb75c54cdb347bd320fb6ac0a4d94/torchao/prototype/mx_formats/mx_tensor.py#L239
-  const int32_t int_amax = *reinterpret_cast<const int32_t *>(&amax);
-  const int32_t extracted_pow2 =
-      ((int_amax >> FP32_MANTISSA_BITS) & 0b11111111) - FP32_EXPONENT_BIAS;
```
lw, line 131: This wasn't handling subnormals correctly.

torchao/csrc/cuda/mx_kernels/mxfp8_quantize.cuh
```diff
@@ -209,63 +217,6 @@ template <typename T, int N> struct Vec {
...
-// https://github.com/NVIDIA/TransformerEngine/blob/1ae1d228d725a488621deba685bd26d6ee1cdb21/transformer_engine/common/utils.cuh#L937
-__device__ __forceinline__ e8m0_t float_to_e8m0(float val) {
-  // TODO: nan/inf needs to be set for any value
-  // of nan/inf in input not just amax.
-  if (isnan(val)) {
-    return 0xFF;
-  }
-  if (isinf(val)) {
-    return 0xFE;
-  }
```
lw, line 233: This was causing Inf to become 2^127. It was clearly intentional, but I'd argue against it.

torchao/csrc/cuda/mx_kernels/mxfp8_quantize.cuh
```diff
@@ -209,63 +217,6 @@ template <typename T, int N> struct Vec {
...
-#if ((__CUDA_ARCH_HAS_FEATURE__(SM100_ALL)) ||                                 \
-     (__CUDA_ARCH_HAS_FEATURE__(SM101_ALL)) ||                                 \
-     (__CUDA_ARCH_HAS_FEATURE__(SM120_ALL)))
-  uint16_t out;
-  asm volatile("{\n"
-               "cvt.rp.satfinite.ue8m0x2.f32  %0, 0.0, %1;\n"
```
lw, line 239: Similrly, `.satfinite` is also causing Inf to become 2^127. But we can just remove that modifier.

torchao/csrc/cuda/mx_kernels/mxfp8_quantize.cuh
```diff
@@ -472,42 +419,26 @@ __device__ inline void copy_2d_to_shared(void *smem,
...
 ////////////////////////////////////////////////////////////////////////////////
 
-/**
- * Convert e8m0 biased scale to float32 scale following torchao implementation
- * torchao ref:
- * https://github.com/pytorch/ao/blob/00417b8b33abb75c54cdb347bd320fb6ac0a4d94/torchao/prototype/mx_formats/mx_tensor.py#L275C1-L277C30
- */
-__device__ __forceinline__ float e8m0_to_scale_fp32(e8m0_t e8m0_biased_scale) {
-  int32_t exponent_as_int32 = static_cast<int32_t>(e8m0_biased_scale);
-  int32_t float_bits = exponent_as_int32 << FP32_MANTISSA_BITS;
```
lw, line 479: Isn't handling 2^-127 correctly.

torchao/csrc/cuda/mx_kernels/mxfp8_quantize.cuh
```diff
@@ -472,42 +419,26 @@ __device__ inline void copy_2d_to_shared(void *smem,
...
-  // Scale the input value
-  float data_lp = input_value * inv_scale_fp32;
-
-  // Apply torchao-style clamping
-  // torchao ref:
-  // https://github.com/pytorch/ao/blob/00417b8b33abb75c54cdb347bd320fb6ac0a4d94/torchao/prototype/mx_formats/mx_tensor.py#L301C23-L301C74
-  data_lp = min(data_lp, F8E4M3_MAX);
-  data_lp = max(data_lp, -F8E4M3_MAX);
-
-  return static_cast<OType>(data_lp);
```
lw, line 507: Must use explicit clamping, because the conversion function being used doesn't make it clear

torchao/csrc/cuda/mx_kernels/mxfp8_quantize.cuh
```diff
@@ -790,27 +698,23 @@ __global__ void __launch_bounds__(MXFP8_THREADS_PER_CHUNK)
           // Load from shared memory into thread local registers
           in.load_from(&in_sh[buff][shmem_offset_y][shmem_offset_x]);
 
-          float thread_amax = 0;
+          float thread_amax = CUDART_NAN_F;
```
lw: The "null" value must become NaNs, so that a tile full of NaNs becomes a scale of NaN. Otherwise it would become 0.

torchao/csrc/cuda/mx_kernels/mxfp8_quantize.cuh
```diff
@@ -790,27 +698,23 @@ __global__ void __launch_bounds__(MXFP8_THREADS_PER_CHUNK)
...
-            // Load and convert to float
-            float elt = DataTypeTraits<IType>::to_float(in.data.elt[j]);
-            in_compute[j] = elt;
-
-            // Update thread local amax
-            if (!out_of_bounds) {
-              thread_amax = fmaxf(thread_amax, fabsf(elt));
-            }
+            const float elt = DataTypeTraits<IType>::to_float(in.data.elt[j]);
+            in_compute[j] = out_of_bounds ? 0.0f : elt;
```
lw: I wonder if the out-of-bounds data should be NaN too?

torchao/prototype/moe_training/kernels/mxfp8/cute_utils.py
```diff
@@ -49,39 +49,170 @@ def _cutedsl_runtime_available() -> bool:
...
+        return unpack(packed, cutlass.Float8E8M0FNU, loc=loc, ip=ip)[0]
 
-        Uses inline PTX on Blackwell-family targets because CuTeDSL does not
-        currently lower this conversion to `cvt.rp.satfinite.ue8m0x2.f32` on its own.
-        """
-        return cutlass.Uint16(
-            llvm.inline_asm(
-                T.i16(),
-                [cutlass.Float32(a).ir_value(loc=loc, ip=ip)],
-                "cvt.rp.satfinite.ue8m0x2.f32 $0, 0.0, $1;",
```
lw, line 76: Again the `.satfinite` issue (we want this when doing fp32->e4m3, but not when doing fp32->e8m0)

torchao/prototype/moe_training/kernels/mxfp8/cute_utils.py
```diff
@@ -92,36 +223,21 @@ def compute_amax(vals_block: cute.Tensor):
         Returns:
             The absolute maximum value as Float32
         """
-        vals_vec = vals_block.load()
-        abs_vec = cute.where(vals_vec < 0, -vals_vec, vals_vec)
-        return cutlass.Float32(
-            abs_vec.reduce(cute.ReductionOp.MAX, cutlass.Float32(0.0), 0)
```
lw, line 98: This reduction propagates NaNs

torchao/prototype/moe_training/kernels/mxfp8/cute_utils.py
```diff
@@ -92,36 +223,21 @@ def compute_amax(vals_block: cute.Tensor):
...
-        # referene: https://github.com/pytorch/ao/blob/ac0b820899b0a5d415310f798c9c96b5a5973f53/torchao/csrc/cuda/mx_kernels/mxfp8_quantize.cuh#L538
         descale = amax * INV_F8_MAX
-        scale_biased = cutlass.Int32(_cvt_rp_satfinite_ue8m0x2_f32(descale))
-        inv_scale = cutlass.Float32(1.0)
-        if scale_biased == 0xFF:
-            inv_scale = cutlass.Float32(0.0)
-        elif scale_biased == 0:
-            inv_scale = cute.exp2(cutlass.Float32(126.0))
-        else:
-            inv_scale = cute.exp2(cutlass.Float32(127 - scale_biased))
```
lw, line 123: I'm puzzled by what's going on here. NaN becomes 0, the inverse of 2^-127 is 2^126, ...

torchao/prototype/mx_formats/kernels.py
```diff
@@ -480,8 +480,8 @@ def _calculate_reciprocal_scale(scale_e8m0_biased):
                 scale_e8m0_biased == 254,  # Inf case -> return 2^-127
                 2**-127,
                 tl.where(
-                    scale_e8m0_biased == 0,  # Zero case -> return 1.0 (no scaling)
-                    1.0,
```
lw, line 484: I think this is a bug carried over from TransformerEngine, and since fixed there: https://github.com/NVIDIA/TransformerEngine/pull/2647 and https://github.com/NVIDIA/TransformerEngine/pull/3262

torchao/prototype/mx_formats/mx_tensor.py
```diff
@@ -150,64 +202,26 @@ def _to_mx_rceil(
...
-        # See https://github.com/pytorch/pytorch/issues/191013.
-        descale,
-        torch.where(
-            exponent == 254,  # Inf case -> return 2^-127
-            2**-127,
-            # Normal case
-            torch.where(
-                exponent == 0,
-                1.0,
-                torch.exp2(E8M0_EXPONENT_BIAS - exponent.to(torch.float32)),
```
lw, line 200: Again the TE issue, plus https://github.com/NVIDIA/TransformerEngine/issues/2408

torchao/prototype/mx_formats/kernels.py
```diff
@@ -551,17 +548,17 @@ def _triton_calculate_scale_floor(
...
-        bf16_exp_bias = 127
+        fp32_mbits = 23
+        fp32_exp_bias = 127
 
         # Find the maximum absolute value for each row
         max_abs = tl.max(x, axis=axis)
 
         # Original floor implementation
         # Calculate the e8m0 scale by extracting the exponent (floor)
-        max_abs = max_abs.to(tl.bfloat16)
```
lw, line 562: So, this one is subtle: going fp32->bf16->e8m0 (instead of directly fp32->e8m0) can round some fp32 values across the threshold that makes them round up instead of down!

torchao/prototype/mx_formats/kernels.py
```diff
@@ -521,21 +524,15 @@ def _triton_calculate_scale_rceil(x, axis, USE_PTX: tl.constexpr):
...
                 constraints="=h,r",
                 args=[scale_input.to(tl.float32, bitcast=False)],
                 dtype=tl.uint16,
                 is_pure=True,
                 pack=1,
             ).to(tl.uint8)
         else:
-            # Fallback implementation
-            scale_e8m0_unbiased = tl.clamp(
-                tl.ceil(tl.log2(scale_input)),
```
lw, line 534: Log2 is too inaccurate for such an operation

torchao/csrc/cuda/mx_kernels/mxfp8_cuda.cu
lw: Moved to https://github.com/pytorch/ao/pull/4732

torchao/csrc/cuda/mx_kernels/mxfp8_quantize.cuh
```diff
@@ -282,9 +219,7 @@ __forceinline__ __device__ float subwarp_reduce_max_broadcast(const float val) {
   for (int offset = subwarp_width / 2; offset > 0; offset /= 2) {
     const float val_other =
         __shfl_down_sync(0xFFFFFFFF, val_tmp, offset, subwarp_width);
-    __builtin_assume(val_tmp >= 0);
-    __builtin_assume(val_other >= 0);
```
lw, line 286: These assumption inequalities do not hold if the values are NaN.

vkuzo:
> 1. The main decision is what to do with NaNs: propagate them (one NaN "pollutes" the whole tile) or ignore them (amax is NaN only if the whole tile is NaN).

I think code which contains snippets equivalent to `torch.max(torch.abs(tensor))` should propagate NaNs.

First, core precedent:

```python
>>> import torch
>>> some_nans = torch.tensor([float('nan'), 1.0, -1.0])
>>> all_nans = torch.tensor([float('nan'), float('nan')])
>>> some_nans
tensor([nan,  1., -1.])
>>> all_nans
tensor([nan, nan])
>>> some_nans.abs().max()
tensor(nan)
>>> all_nans.abs().max()
tensor(nan)
```

Second, torchao precedent: https://gist.github.com/vkuzo/498a4e65ac6012bf4394d8461b002a9d. Clauded and I did not check every line, but the intent has always been to propagate.

Third, @lw's example of fp4, which I have not yet thought about for NaN semantics but what @lw  is saying makes perfect sense.  The fp4 encoding does not have a NaN, so if we want sane NaN handling it must be in the scale.

test/prototype/moe_training/test_kernels.py
```diff
@@ -21,6 +21,10 @@ def _is_sm_10x() -> bool:
         allow_module_level=True,
     )
 
+from test.prototype._mxfp8_test_utils import (
```
vkuzo: I think this import will not work properly in fbcode, can we move this to `torchao.testing` which does work consistently across OSS and fbcode?

torchao/testing/_mxfp8_test_utils.py
```diff
@@ -0,0 +1,283 @@
...
+
+@dataclass(frozen=True)
+class MXFP8SemanticCases:
+    names: tuple[str, ...]
+    inputs: torch.Tensor
+    expected_data: torch.Tensor
+    expected_scales: torch.Tensor
+
+
+def make_mxfp8_semantic_cases(
```
vkuzo, line 22: love it

test/prototype/_mxfp8_test_utils.py
```diff
@@ -0,0 +1,283 @@
...
+    # Rceil rounds 480/448 up to a scale of 2, whereas floor just looks at its
+    # previous power of 2 and treats it the same as 448, yielding a scale of 1.
+    # Floor might therefore have a tendency to overflow when computing the data.
+    # We use this to check for proper clamping (implicit or explicit).
+    satfinite_scale = 127 if is_floor else 128  # 1.0 or 2.0
+    satfinite_data = 0x7E if is_floor else 0x77  # 448.0 or 240.0
+
+    for idx, name in enumerate(names):
+        if name == "zero":
+            # Leave eveything at zero. Note this means the scale is 2^-127.
```
vkuzo: nitty, but would be cleaner to initialize the zero case here

vkuzo:
> 2. Convert the amax / fp8_max to a ue8m0 scale
> We need to decide what to do about Inf. Here I decided to make it produce a scale of NaN, as saturating it to the largest scale might end up hiding a numerical bug, and would cause an infinitely large quantization error.

inf amax -> nan scale makes sense to me practically! 

> Another catch is to make sure to handle subnormals correctly, since the lowest ue8m0 value is just outside the range of normals.

sg, I did not read the code line by line but this wasn't really thought through before and I trust the test case

> Also note that doing this via log2+ceil (or floor) is mathematically correct but numerically incorrect because of some inaccuracies of these functions (which are also quite expensive on GPU!).

+1

vkuzo:
> 3. Taking the reciprocal of the scale
> NaNs must pass-through unchanged. And again there's the subnormal value (corresponding to 2^-127) which might need special care. A safe approach is to reinterpret this as 254 - x in uint8.
> 4. Converting the reciprocal back to a fp32
> Pretty much the symmetrical problems: handling NaN, mapping 2^-127 to a subnormal, avoiding exp2 due to numerical issues (and slowness). And again the solution is bit-manipulation or PTX builtins.

lg and the after is strictly better than before!

vkuzo:
> Converting data to e4m3 or e5m2
This is pretty straightforward if we use the right builtins. One catch is that the floor mode needs to do clamping. It might be done automatically by the conversion if we use saturation, but if this isn't the case we need to do it manually.

lg and looks strictly better than before, as we did not previously cover e5m2 here

vkuzo (approved):
I agree with all points in 1-5.  I didn't read every line of the implementation - lmk if I should

vkuzo:
we should just change

```
from test.prototype._mxfp8_test_utils import ...
```

to 

```
from torchao.testing.{foobar} import ...
```

to make fbcode happy

slayton58:
> I'd appreciate some discussion of the above decisions, so that we get to a large consensus. The one I'm most unsure about is the ignore-vs-propagate for NaNs.

I am _very_ firmly in favour of propagate.
* Consistent behaviour across fp{4,6,8} or any other potential output type that doesn't encode NaN as you mention
* Faster checks for "is this scaled tensor valid" -- scan over the scale values (so 32x fewer) rather than output values thems. You can still then identify individual bad input elements by looking over the values that contributed to a bad scale.

Inf amax -> NaN scale makes sense to me as well, I consider inf/nan values in an input to be effectively broken, so I'd want them flagged.

PR #4742, opened 2026-08-11
Remove unused quantization code

iamzainhuda:
## Summary

This removes 185 lines of unreachable, shadowed, or already-broken quantization code. Repository-wide exact-symbol searches found no live callers for the removed helpers. No supported top-level APIs are changed.

### Why each removal is dead

- **`torchao/prototype/mx_formats/mx_funcs.py`**: this module has no inbound imports and cannot be imported because it references the deleted `torchao.prototype.mx_formats.mx_ops` module. The active MX linear dispatch is implemented in `mx_tensor.py`.
- **Stale codebook `__all__` entries**: `QuantizedLutEmbedding` and `EmbeddingLutQuantizer` are not defined or imported anywhere in the package. Their only repository occurrences were the two strings in `__all__`, which made wildcard imports request nonexistent attributes.
- **First `GGUFQuantizedTensor._apply_fn_to_data` definition**: the class defines this method a second time later in the same class body, so Python always overwrites the first definition. The unreachable version also contains the typo `super_block_min_sclae`; the live second implementation is unchanged.
- **`_embedding_extra_repr` and `_get_linear_subclass_inserter`**: exact repository searches find only their definitions; neither is called, imported, registered, tested, documented, or included in `quant_api.__all__`. `_get_linear_subclass_inserter` is distinct from the live `_get_subclass_inserter`, which remains exported and used by `benchmark_aq.py` and `test_quant_api.py`. The `_linear_extra_repr` and `_quantization_type` definitions remain; only their now-dead imports into `quant_api.py` are removed.
- **Private FloatX primitive chain**: `_choose_qparams_affine_floatx`, `_quantize_affine_floatx`, and `_dequantize_affine_floatx` have no callers in source, tests, docs, examples, or benchmarks. Their previous consumer disappeared with the legacy `quant_llm_linear` path. `_ONES_TABLE` and the three removed imports existed only to support this chain. The underlying custom-FP helpers remain in `prototype/custom_fp_utils.py`, where active MX kernels still use them.
- **`call_from_inner_tensors` in NF4**: this helper has no callers and is not registered in the NF4 `__torch_dispatch__` or `__torch_function__` tables. The active `apply_to_inner_tensors` helper and all NF4 dispatch behavior remain unchanged.
- **`_MultiInput`**: this private GPTQ/model-evaluation remnant is never imported or constructed. Its only other occurrence was its self-reference when slicing.

Public GitHub code search found no source consumers of `_get_linear_subclass_inserter`; matches were generated package metadata rather than imports or calls. Internal fbcode usage cannot be checked from this environment and should be confirmed through the normal internal review process.

## Impact

- 185 lines removed across 7 files
- 3 unused private entries removed from `quant_primitives.__all__`
- 2 invalid entries removed from `codebook_groupwise.__all__`
- no stable top-level torchao API changes

## Test plan

- `ruff check torchao/quantization/quant_primitives.py torchao/quantization/quant_api.py torchao/quantization/utils.py torchao/quantization/quantize_/workflows/nf4/nf4_tensor.py torchao/prototype/quantization/codebook_groupwise/__init__.py torchao/prototype/quantization/gguf/gguf_quantized_tensor.py`
- `pytest -q test/quantization/test_quant_primitives.py`
- `pytest -q test/quantization/quantize_/workflows/nf4/test_nf4_tensor.py`
- `pytest -q test/prototype/test_gguf_quant.py test/prototype/test_codebook_quant.py`
- `pytest -q test/prototype/mx_formats/test_mx_tensor.py`
- `CUDA_VISIBLE_DEVICES='' pytest -q test/quantization/test_quant_api.py`

Local result: 255 passed, 54 skipped.


torchao/prototype/quantization/codebook_groupwise/__init__.py
```diff
@@ -4,6 +4,4 @@
 __all__ = [
     "CodebookQuantizedPackedTensor",
     "GroupwiseLutWeightConfig",
-    "QuantizedLutEmbedding",
```
iamzainhuda, line 7: these do not exist anymore

torchao/prototype/quantization/gguf/gguf_quantized_tensor.py
```diff
@@ -66,18 +66,6 @@ def __init__(
         self.quantized_block_min = quantized_block_min
         self.int_data = int_data
 
-    def _apply_fn_to_data(self, fn):
```
iamzainhuda, line 69: class defines this function twice. the first implementation never executes because python overrides with the second one. so so we delete it

torchao/prototype/quantization/gguf/gguf_quantized_tensor.py
```diff
@@ -66,18 +66,6 @@ def __init__(
...
-        return self.__class__(
-            self.n_blocks_per_superblock,
-            fn(self.super_block_scale_scale),
```
iamzainhuda, line 72: there is also a typo in the overridden fn

vkuzo:
thank you! I'd recommend:
1. these changes seem unrelated to each other, would be good to split to one PR per change instead of all together
2. make sure CI is green

iamzainhuda:
Superseded by focused, independently reviewable PRs:

- #4748: remove orphaned MX dispatch module
- #4752: remove stale codebook exports
- #4750: remove shadowed GGUF tensor method
- #4751: remove unused quantization API helpers
- #4749: remove unused FloatX quantization primitives
- #4747: remove unused NF4 helper
- #4753: remove unused multi-input helper

Each replacement is based independently on the latest `main` and changes one file.

PR #4749, opened 2026-08-13
Remove unused FloatX quantization primitives

iamzainhuda:
## Summary

Remove `_choose_qparams_affine_floatx`, `_quantize_affine_floatx`, and `_dequantize_affine_floatx`, plus their private `__all__` entries and now-unused support imports/table.

Repository-wide searches find no callers in source, tests, docs, examples, or benchmarks. Their previous consumer disappeared with the legacy `quant_llm_linear` path. `_ONES_TABLE` and the removed imports existed only for this chain. The underlying custom-FP helpers remain in `prototype/custom_fp_utils.py`, where active MX kernels still use them.

This removes 70 lines and three unused private module exports; no top-level API changes.

Split from #4742.

## Test plan

- `pytest -q test/quantization/test_quant_primitives.py`

Result: 26 passed.

### DIFF
--- a/torchao/quantization/quant_primitives.py
+++ b/torchao/quantization/quant_primitives.py
@@ -11,11 +11,6 @@ from typing import Callable, Dict, List, Optional, Tuple, Union
 
 import torch
 
-from torchao.prototype.custom_fp_utils import (
-    _f32_to_floatx_unpacked,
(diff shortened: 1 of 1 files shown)


vkuzo (approved):
nice find!

vkuzo:
@claude review

PR #4752, opened 2026-08-13
Remove stale codebook exports

iamzainhuda:
## Summary

Remove `QuantizedLutEmbedding` and `EmbeddingLutQuantizer` from `codebook_groupwise.__all__`.

Neither name is defined or imported by the package. Their only repository occurrences are these two strings, so wildcard imports currently request nonexistent attributes. The two real codebook exports remain unchanged.

This removes two invalid API entries.

Split from #4742.

## Test plan

- `python -c "from torchao.prototype.quantization.codebook_groupwise import *; assert CodebookQuantizedPackedTensor and GroupwiseLutWeightConfig"`
- `pytest -q test/prototype/test_codebook_quant.py`

Result: 4 passed.


vkuzo:
@claude review

vkuzo (approved):

PR #4764, opened 2026-08-16
Fix: use 'is None' instead of 'not' for tensor checks in FixedQParamO…

gaurav-redhat:
## Summary

`AffineQuantizedFixedQParamObserver` in `torchao/quantization/observer.py` uses `if not scale` and `if not zero_point` to check for missing arguments. This is incorrect for tensors:

1. **Crash** — Multi-element tensors (e.g. per-channel scales) raise `RuntimeError: Boolean value of Tensor with more than one value is ambiguous`
2. **Silent data corruption** — A valid single-element `scale=0.0` is treated as falsy and overwritten with `1.0`

### Buggy code (3 locations)

```python
# __init__ (lines 227-230)
if not scale:              # BUG: should be 'is None'
    scale = torch.Tensor([1])
if not zero_point:         # BUG: should be 'is None'
    zero_point = torch.zeros_like(scale)

# set_qparams (line 235)
if not zero_point:         # BUG: should be 'is None'
    zero_point = torch.zeros_like(scale)

### DIFF
--- a/test/quantization/test_observer.py
+++ b/test/quantization/test_observer.py
@@ -286,6 +286,51 @@ class TestQuantFlow(TestCase):
         )
 
 
+    def test_fixed_qparam_observer_multi_element_scale(self):
+        """Regression test: 'if not scale' crashes on multi-element tensors."""
+        scale = torch.tensor([1.0, 2.0, 3.0])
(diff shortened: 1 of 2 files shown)


andrewor14:
@claude review

PR #4766, opened 2026-08-17
Fix cvt_packfloat compatibility with nvidia-cutlass-dsl 4.6.0+

andrewor14:
## Summary
`nvidia-cutlass-dsl` 4.6.0 removed a leading argument in a few nvvm ops. The current callsite in torchao (introduced in #4725) still includes this leading argument, resulting in the following error:

```
TypeError: cvt_packfloat_f32() takes 4 positional arguments but 5 were given
```

Correct signature:

```
\# 4.5.x
nvvm.cvt_packfloat_f32(T.i32(), src_a, src_b, src_c, to, rnd=..., sat=...)
\# 4.5.x[cu13], 4.6.0+
nvvm.cvt_packfloat_f32(src_a, src_b, src_c, to, rnd=..., sat=...)
```

Note that we cannot simply gate this on the verison because different variants within the same version can have different signatures. Here is the complete table:

```
nvidia-cutlass-dsl          4.4.x, 4.5.x    accepts `res` == T.i32()
nvidia-cutlass-dsl[cu13]    4.4.x, 4.5.x    no `res`
nvidia-cutlass-dsl          4.6.0+          no `res`
nvidia-cutlass-dsl[cu13]    4.6.0+          no `res`
```

Full list of affected ops used by `cute_utils.py`:

```
fmax
cvt_packfloat
cvt_packfloat32
```

## Test Plan

Full version sweep across the following versions. vllm nightly currently requires 4.6.2.

```
nvidia-cutlass-dsl          4.5.3, 4.6.0, 4.6.2, 4.7.1
nvidia-cutlass-dsl[cu13]    4.5.3, 4.6.0, 4.6.2, 4.7.1
```

Repro 1:
```
import torch
from torchao.prototype.moe_training.kernels.mxfp8 import (
    mxfp8_quantize_2d_1x32_cutedsl,
)

x = torch.randn(128, 2048, device="cuda", dtype=torch.bfloat16)
offs = torch.tensor([128], device="cuda", dtype=torch.int32)
mxfp8_quantize_2d_1x32_cutedsl(x, scaling_mode="rceil", offs=offs)
```

Repro 2:
```
import torch
from torchao.prototype.moe_training.kernels.mxfp8 import (
    mxfp8_quantize_cuda_3d,
)

x = torch.randn(2, 256, 256, device="cuda", dtype=torch.bfloat16)
mxfp8_quantize_cuda_3d(
    x, scale_block_dim1=32, scale_block_dim2=32, scaling_mode="rceil",
)
```

### DIFF
--- a/torchao/prototype/moe_training/kernels/mxfp8/cute_utils.py
+++ b/torchao/prototype/moe_training/kernels/mxfp8/cute_utils.py
@@ -7,6 +7,7 @@
 """Shared utilities for CuTeDSL quantization kernels."""
 
 import importlib.util
+import inspect
 
 # Runtime package detection
 _CUTEDSL_RUNTIME_PACKAGES = {
@@ -52,6 +53,25 @@ if _cutedsl_runtime_available():
     from cutlass._mlir import ir
     from cutlass._mlir.dialects import arith, llvm, nvvm, vector
     from cutlass.cutlass_dsl import T, dsl_user_op
+
+    # For some NVVM ops, older cutlass-dsl builds take the op's result type as
+    # a leading `res` argument, while newer ones infer it from the operands.
+    # However, we cannot simply gate this on the version, since the same
+    # version can ship either signature depending on the CUDA extra it is
+    # installed with:
+    #
+    #   nvidia-cutlass-dsl          4.4.x, 4.5.x    accepts `res`
+    #   nvidia-cutlass-dsl[cu13]    4.4.x, 4.5.x    no `res`
+    #   nvidia-cutlass-dsl          4.6.0+          no `res`
+    #   nvidia-cutlass-dsl[cu13]    4.6.0+          no `res`
+    #
+    # TODO: delete this once we drop 4.5.x support.
+    def _accepts_res_arg(op) -> bool:
+        return "res" in inspect.signature(op).parameters
+
+    _FMAX_ACCEPTS_RES_ARG = _accepts_res_arg(nvvm.fmax)
+    _CVT_PACKFLOAT_F32_ACCEPTS_RES_ARG = _accepts_res_arg(nvvm.cvt_packfloat_f32)
+    _CVT_PACKFLOAT_ACCEPTS_RES_ARG = _accepts_res_arg(nvvm.cvt_packfloat)
 
     # FP8 constants
     INV_F8_MAX = cutlass.Float32(1.0 / 448.0)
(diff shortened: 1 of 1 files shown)


andrewor14:
@claude review

janbernloehr:
Thanks for addressing the `cvt_packfloat_f32` / `cvt_packfloat` signature change.

One additional compatibility point may be worth including before merge: `cute_utils.py` also calls `nvvm.fmax` with a leading result type (`T.f32()`), and the 4.6.x API changed that result-type convention as well. The same compatibility handling should cover the `fmax_nan_f32` call.

Also, since `nvvm.FPRoundingMode` is now used on the pre-4.6 branch, could we confirm that symbol exists across all supported 4.5.x versions? If not, enum selection should be version- or feature-gated separately from the positional-argument handling.

It would be useful to test both a 4.5.x and 4.6.x CuTeDSL environment before merging, since the goal is to preserve older-version support while adding 4.6+ compatibility.

andrewor14:
@claude review again? I just rebased. The naming change was actually made before this PR

andrewor14:
Hi @janbernloehr, thanks for the suggestion. I fixed it for fmax too. Will test on 4.5.0 now. I did actually confirm that `FPRoundingMode` was in 4.5.x as well as all future versions including 4.6.x and 4.7.x here: https://github.com/pytorch/ao/pull/4765.

andrewor14:
@claude review latest changes again

torchao/prototype/moe_training/kernels/mxfp8/cute_utils.py
```diff
@@ -53,6 +54,23 @@ def _cutedsl_runtime_available() -> bool:
...
+    # For some NVVM ops, older cutlass-dsl builds take the op's result type as
+    # a leading `res` argument, while newer ones infer it from the operands.
+    # However, we cannot simply gate this on the version, since the same
+    # version can ship either signature depending on the CUDA extra it is
+    # installed with:
+    #
+    #   nvidia-cutlass-dsl          4.4.x, 4.5.x    accepts `res`
+    #   nvidia-cutlass-dsl[cu13]    4.4.x, 4.5.x    no `res`
+    #   nvidia-cutlass-dsl          4.6.0+          no `res`
+    #   nvidia-cutlass-dsl[cu13]    4.6.0+          no `res`
```
vkuzo, line 66: can we add a TODO here to go back to a version gate once 4.5.x ages out of pytorch support

vkuzo (approved):

PR #4775, opened 2026-08-17
Add opt-in KleidiAI support to BUCK (#4775)

JakeStevens:
Summary:

Enabling `TORCHAO_ENABLE_KLEIDI` exposed prevoiusly uncompiled function definitions inside the headers, which failed with missing-prototypes

Moving them to inline to match other usages like linear/ for these types of header-provided implementations.

Reviewed By: digantdesai

Differential Revision: D116355119


digantdesai (approved):
Review automatically exported from Phabricator review in Meta.

PR #4784, opened 2026-08-19
Fix inverted bias flag in swap_conv2d_1x1_to_linear

### LINKED ISSUE #4615: Inverted bias flag in replace_conv2d_1x1 inside swap_conv2d_1x1_to_linear
## Description

In `swap_conv2d_1x1_to_linear`, the `bias` argument passed to `nn.Linear` is logically inverted — it creates a bias when the original conv has none, and skips it when the conv has one.

**File:** `torchao/quantization/quant_api.py`  
**Line:** 234–238

## Current code

```python
def replace_conv2d_1x1(conv):
    assert conv.kernel_size == (1, 1)
    lin = torch.nn.Linear(
        conv.in_channels, conv.out_channels, bias=(conv.bias is None)  # inverted
    )
    lin.weight = torch.nn.Parameter(conv.weight.squeeze(-1, -2))
    lin.bias = conv.bias
    return PermuteSandwich(lin)
```

## Steps to reproduce
```python
import torch
import torch.nn as nn

# Step 1: Create a Conv2d WITHOUT bias
conv = nn.Conv2d(3, 16, 1, bias=False)

# Step 2: See what the buggy code passes to nn.Linear
print(f"conv.bias is None     = {conv.bias is None}")      # True
print(f"conv.bias is not None = {conv.bias is not None}")   # False

# Step 3: Buggy path — nn.Linear allocates an unnecessary bias
lin_buggy = nn.Linear(3, 16, bias=(conv.bias is None))   # bias=True (WRONG)
print(f"lin_buggy.bias = {lin_buggy.bias is not None}")    # True — wasted allocation

# Step 4: Fixed path — nn.Linear correctly skips bias
lin_fixed = nn.Linear(3, 16, bias=(conv.bias is not None))  # bias=False (CORRECT)
print(f"lin_fixed.bias = {lin_fixed.bias is not None}")     # False — clean

# Step 5: After lin.bias = conv.bias, both end up the same
lin_buggy.bias = conv.bias   # overwrites with None
lin_fixed.bias = conv.bias   # already None

print(f"lin_buggy.bias is None = {lin_buggy.bias is None}")  # True
print(f"lin_fixed.bias is None = {lin_fixed.bias is None}")  # True
```

amanyagami:
## Summary
In `replace_conv2d_1x1` (used by `swap_conv2d_1x1_to_linear`), the `bias=` argument passed to the `nn.Linear` constructor is inverted relative to intent:

```python
lin = torch.nn.Linear(
    conv.in_channels, conv.out_channels, bias=(conv.bias is None)  # should be 'is not None'
)
lin.weight = torch.nn.Parameter(conv.weight.squeeze(-1, -2))
lin.bias = conv.bias
```

**Important note found during verification:** this does **not** currently cause a user-observable bug. The very next line, `lin.bias = conv.bias`, unconditionally overwrites whatever the constructor allocated — `nn.Module.__setattr__` re-registers the `bias` parameter regardless of the module's initial `bias=` flag, as long as the target attribute name already exists in `_parameters` (which it always does after `nn.Linear.__init__`). So the final module ends up with the correct bias either way today.

It's still worth fixing because:
1. The flag is backwards relative to its obvious intent and is misleading to future readers.
2. It's technically wasteful in the `bias=True`-when-unnecessary case (allocates and discards a placeholder bias Parameter).
3. It's a latent foot-gun — if the explicit `lin.bias = conv.bias` override is ever refactored away, the inverted flag would become a real, silent bug.

## Fix
`bias=(conv.bias is not None)`, matching intent.

## Test
Added `test_swap_conv2d_1x1_to_linear_bias` in `test/quantization/test_quant_api.py`, covering both the bias and no-bias case, checking both the resulting module's `bias` attribute and full numerical equivalence to the original conv's output.

Fixes #4615

### DIFF
--- a/test/quantization/test_quant_api.py
+++ b/test/quantization/test_quant_api.py
@@ -48,6 +48,7 @@ from torchao.quantization.quant_api import (
     PerRow,
     PerTensor,
     _replace_with_custom_fn_if_matches_filter,
+    swap_conv2d_1x1_to_linear,
 )
 from torchao.quantization.quant_primitives import MappingType
 from torchao.quantization.utils import compute_error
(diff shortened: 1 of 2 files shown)


amanyagami:
recheck

amanyagami:

PR #4785, opened 2026-08-19
Remove dead, broken torchao/prototype/attention/fp8_fa3/setup.py

### LINKED ISSUE #4318: Import of nonexistent config module crashes fp8_fa3/setup.py

amanyagami:
## Summary
`torchao/prototype/attention/fp8_fa3/setup.py` is orphaned dead code:

1. It imports `torchao.prototype.attention.config.LowPrecisionAttentionConfig`. That module/class has never existed anywhere in this repo, so simply `import`ing this file raises `ModuleNotFoundError` (the crash reported in #4318).
2. Even setting that aside, `setup_fp8_fa3()` calls `setup_fp8_backend(model, config, flash_impl_name="FA3", sdpa_fn=fp8_fa3_sdpa)`, which does not match the real signature of `setup_fp8_backend` in `shared_utils/setup.py` — `def setup_fp8_backend(model, flash_impl_name, hadamard="NONE")`. There's no `config` or `sdpa_fn` parameter at all.
3. Nothing in the codebase imports or calls `setup_fp8_fa3` (`grep -rn "setup_fp8_fa3"` only matches its own definition).

The real, current, tested entry point for this feature is `torchao.prototype.attention.api.apply_low_precision_attention()`, which already correctly calls `shared_utils.setup.setup_fp8_backend()`.

## Fix
Delete the file rather than patch it to match a signature that already has a working call site elsewhere — keeping it around as a second, subtly-different, untested/unreachable path would just be a bug waiting to happen if something ever does start importing it.

## Test
Added `test/prototype/attention/test_fp8_fa3_package_import.py`, which:
- imports `torchao.prototype.attention.fp8_fa3` and checks its `__all__` members are present
- imports the public `apply_low_precision_attention` API
- asserts `torchao.prototype.attention.fp8_fa3.setup` no longer resolves

None of this requires CUDA/Hopper hardware — it's purely import-time correctness (verified with a CPU-only torch + triton install).

Fixes #4318


amanyagami:
Cleanup — `torchao/prototype/attention/fp8_fa3/setup.py` is dead code that can't even be imported (references a `LowPrecisionAttentionConfig` class that doesn't exist anywhere in the repo, per #4318). Removing it rather than trying to fix code nothing calls. CI green.

amanyagami:

PR #4787, opened 2026-08-19
Make GPTQ device-agnostic (remove CUDA-only assert and torch.cuda.synchronize)

### LINKED ISSUE #4600: GPTQ hardcodes CUDA device — blocks all non-CUDA backends
### Description

`torchao/prototype/gptq/api.py` has two CUDA-specific lines:

**Line 390** — device type assertion:
```python
assert device.type == "cuda", "GPTQ only supports CUDA currently"
```

**Line 534** — CUDA synchronize:
```python
torch.cuda.synchronize()
```

The GPTQ algorithm (Hessian-based column-wise error propagation) is implemented in pure PyTorch (`torch.diag`, `torch.linalg`, `matmul`, indexing) and works on any accelerator device. These two lines are the only blockers.

### Steps to reproduce

```python
import torch
from torchao.prototype.gptq import gptq_quantize

# Any non-CUDA accelerator (via PrivateUse1)
device = "mlu"
calibration_data = torch.randn(8, 256, device=device)

gptq_quantize(model, calibration_data, Int4WeightOnlyConfig())
# → AssertionError: GPTQ only supports CUDA currently


Even after removing the assert manually, line 534 crashes:


torch.cuda.synchronize()
# → RuntimeError: No CUDA GPUs are available
```

### Environment

```text
- torchao version: 0.17.0 
- The GPTQ algorithm itself is pure PyTorch and device-agnostic; only these two lines block non-CUDA devices
```

amanyagami:
## Summary
`gptq_quantize()` in `torchao/prototype/gptq/api.py` had two CUDA-specific lines:

```python
assert device.type == "cuda", "GPTQ only supports CUDA currently"
...
torch.cuda.synchronize()
```

The GPTQ algorithm itself (Hessian-based column-wise error propagation via `torch.diag`, `torch.linalg.cholesky`, `matmul`, indexing) is pure PyTorch and device-agnostic — these two lines are the only blockers, as also noted in the issue.

## Fix
- Drop the CUDA-only assert.
- Replace the unconditional `torch.cuda.synchronize()` with `torch.accelerator.synchronize(device)`, guarded to skip when `device.type == "cpu"` (CPU needs no stream sync and isn't a `torch.accelerator` device).

## Test
Added `test/prototype/gptq/test_gptq_device_agnostic.py`, which runs `gptq_quantize` end-to-end **on CPU** with `Int8WeightOnlyConfig` (int4 needs `fbgemm_gpu`/`mslk`, not available here) and checks the result is a valid `Int8Tensor` of the correct shape, dequantizing back to a sane result.

- Verified this test fails with the exact `AssertionError: GPTQ only supports CUDA currently` from the issue on the unpatched code.
- Passes with the fix.
- The existing CUDA-gated suite (`test/prototype/gptq/test_gptqv2.py`) still collects and skips cleanly without CUDA (1 passed, 16 skipped).

Fixes #4600

### DIFF
--- a/test/prototype/gptq/test_gptq_device_agnostic.py
+++ b/test/prototype/gptq/test_gptq_device_agnostic.py
@@ -1,0 +1,63 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Regression test for https://github.com/pytorch/ao/issues/4600.
+
+gptq_quantize() previously had two CUDA-only lines:
+  - `assert device.type == "cuda", "GPTQ only supports CUDA currently"`
+  - `torch.cuda.synchronize()`
+even though the GPTQ algorithm itself (torch.diag, torch.linalg, matmul,
+indexing) is pure PyTorch and device-agnostic. This test runs it on CPU,
+which used to fail with an AssertionError (or, if that assert were removed
+by hand, a `RuntimeError: No CUDA GPUs are available` from the
+unconditional torch.cuda.synchronize()).
+"""
+
+import pytest
+import torch
+import torch.nn.functional as F
+
+from torchao.utils import torch_version_at_least
+
+pytestmark = pytest.mark.skipif(
+    not torch_version_at_least("2.11.0"),
+    reason="GPTQ prototype requires PyTorch 2.11+",
+)
+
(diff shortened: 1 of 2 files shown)


amanyagami:
`gptq_quantize()` has a CUDA-only device assert and an unconditional `torch.cuda.synchronize()`, so it can't run on other accelerator backends even though nothing else in the function is CUDA-specific. Small fix, CI green.

amanyagami:

PR #4788, opened 2026-08-19
Make autocast dtype detection device-agnostic (not CUDA-only)

### LINKED ISSUE #4601: [Bug] Hardcoded CUDA autocast queries cause silent fp32 fallback on non-CUDA backends

amanyagami:
## Summary
Three call sites hardcoded CUDA autocast queries, so on non-CUDA devices with device autocast active they always saw autocast as disabled and silently left the input in its original (usually fp32) dtype instead of casting it to the active autocast dtype:

1. `torchao/float8/float8_linear.py:258` — `Float8Linear.forward` used bare `torch.is_autocast_enabled()` (implicitly `device_type="cuda"`) and `torch.get_autocast_gpu_dtype()`.
2. `torchao/prototype/quantized_training/int8_mixed_precision.py:170` and `torchao/prototype/quantized_training/bitnet.py:145` — both explicitly hardcoded `torch.is_autocast_enabled("cuda")` and `torch.get_autocast_gpu_dtype()`.

## Fix
As suggested in the issue: use the input tensor's own device type instead of hardcoding `"cuda"`:
```python
device_type = input.device.type
if torch.is_autocast_enabled(device_type):
    dtype = torch.get_autocast_dtype(device_type)
```
On CUDA this is behaviorally identical (`input.device.type == "cuda"`); on any other autocast-supporting device (XPU, MLU, NPU, ...) it now works correctly.

## Test
Added `test/quantization/test_device_agnostic_autocast.py`, covering all three call sites entirely **on CPU** under `torch.autocast(device_type="cpu")`. Since the underlying fp8/int8 matmul kernels are GPU-only, each test monkeypatches the wrapped compute op (`matmul_with_hp_or_float8_args`, `_Int8MixedPrecisionTrainingLinearFunction`, `_BitNetTrainingLinear`) to capture the dtype it actually receives, so only the autocast-detection logic under test runs — no GPU kernel is exercised.

- Verified all three tests fail with `dtype == torch.float32` (autocast silently skipped) on the unpatched code — reproducing the exact bug described in the issue.
- Pass with `dtype == torch.bfloat16` after the fix.
- Existing `test/float8/test_base.py` (125 tests) and `test/prototype/test_quantized_training.py` (36 tests) still collect cleanly.

Fixes #4601

### DIFF
--- a/test/quantization/test_device_agnostic_autocast.py
+++ b/test/quantization/test_device_agnostic_autocast.py
@@ -1,0 +1,114 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Regression test for https://github.com/pytorch/ao/issues/4601.
+
+Three call sites hardcoded CUDA autocast queries
+(`torch.is_autocast_enabled()` with no argument, or explicitly
+`torch.is_autocast_enabled("cuda")`, plus `torch.get_autocast_gpu_dtype()`),
+so on a non-CUDA device with device autocast active they always saw
+autocast as disabled and left the input in its original (usually fp32)
+dtype instead of casting it to the active autocast dtype:
+
+  - torchao/float8/float8_linear.py: Float8Linear.forward
+  - torchao/prototype/quantized_training/int8_mixed_precision.py:
+    the F.linear override for Int8MixedPrecisionTrainingLinearWeight
+  - torchao/prototype/quantized_training/bitnet.py: the F.linear override
+    for BitNetTrainingLinearWeight
+
+Each test below runs entirely on CPU under `torch.autocast(device_type="cpu")`
+and checks the dtype actually seen by the wrapped compute op, bypassing the
+op itself (fp8/int8 matmul kernels are GPU-only) via a lightweight monkeypatch
+so only the autocast-detection logic under test executes.
+"""
+
+import torch
+import torch.nn.functional as F
+from torch.testing._internal.common_utils import TestCase, run_tests
+
+
+class TestDeviceAgnosticAutocast(TestCase):
+    def test_float8_linear_respects_cpu_autocast(self):
+        from torchao.float8 import float8_linear as m
+        from torchao.float8.config import Float8LinearConfig
+
+        captured = {}
+
+        class FakeMatmul:
+            @staticmethod
+            def apply(input, weight_t, linear_mm_config, config):
+                captured["dtype"] = input.dtype
+                return input @ weight_t
+
+        orig = m.matmul_with_hp_or_float8_args
+        m.matmul_with_hp_or_float8_args = FakeMatmul
+        try:
+            lin = m.Float8Linear(
+                8, 8, bias=False, config=Float8LinearConfig(emulate=True)
+            )
+            x = torch.randn(4, 8)
+            with torch.autocast(device_type="cpu", dtype=torch.bfloat16):
+                lin(x)
+        finally:
+            m.matmul_with_hp_or_float8_args = orig
+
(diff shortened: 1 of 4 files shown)


amanyagami:
Three call sites hardcode CUDA when checking whether autocast is active, so on a non-CUDA device with device autocast enabled they silently skip the cast and leave the input in its original dtype. Fix makes the checks device-agnostic, CI green, still reproducing on `main`.

amanyagami:

PR #4789, opened 2026-08-19
Fix constant_fold deleting attrs still referenced by live get_attr nodes

### LINKED ISSUE #4420: constant_fold deletes shared get_attr target while another get_attr for the same target is still live

amanyagami:
## Summary
`torchao.quantization.pt2e.constant_fold.constant_fold`'s cleanup pass deleted the underlying module attribute for a `get_attr` node as soon as it found one dead `get_attr` node for that target — even if another `get_attr` node with the same target was still live elsewhere in the graph. That leaves the graph with a live `get_attr` node pointing at an attribute that no longer exists, which `graph.lint()` (called from `eliminate_dead_code`) rejects:

```
```

This was hit in ExecuTorch on a model where the same buffer/parameter is referenced through multiple `get_attr` nodes.

## Fix
Compute the set of targets that still have at least one live `get_attr` node *before* deleting anything, and only `delattr` a target when it isn't in that set — i.e. only once every `get_attr` node for that target is dead.

## Test
Added `test/quantization/pt2e/test_constant_fold.py` using the exact repro from the issue: two `get_attr` nodes sharing one buffer target, one folded away (dead), one live.

- Verified it fails with the issue's exact `RuntimeError` on the unpatched code.
- Passes with the fix, and the resulting graph module still runs and produces correct values for both outputs.
- The pre-existing `test/quantization/pt2e/test_quantize_pt2e.py` suite (70 tests, which exercises `constant_fold` indirectly via lowering/quantize_pt2e) still passes.

Fixes #4420

### DIFF
--- a/test/quantization/pt2e/test_constant_fold.py
+++ b/test/quantization/pt2e/test_constant_fold.py
@@ -1,0 +1,66 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Regression test for https://github.com/pytorch/ao/issues/4420.
+
+constant_fold's cleanup pass deleted the underlying module attribute for a
+get_attr node as soon as it found one dead get_attr node for that target -
+even if another get_attr node with the same target was still live. That
+left the graph with a live get_attr pointing at an attribute that no
+longer existed, which graph.lint() rejects.
+"""
+
+import torch
+from torch.fx import Graph, GraphModule
+from torch.testing._internal.common_utils import TestCase, run_tests
+
+from torchao.quantization.pt2e.constant_fold import constant_fold
+
+
+class Root(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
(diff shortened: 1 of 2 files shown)


amanyagami:
@jerryzh168 @andrewor14 this one is 14 lines in `pt2e/constant_fold.py` plus a test, still present on `main`.

Several `get_attr` nodes can share one target when a buffer or parameter is referenced more than once. The cleanup loop deletes the module attribute as soon as it sees *any* `get_attr` node for that target with no users, without checking whether the other nodes pointing at the same target are still live. The graph is then left with a `get_attr` referencing an attribute that no longer exists, which fails later at runtime rather than here.

The fix collects the set of targets that still have live `get_attr` users and only deletes attributes outside it.

amanyagami:

PR #4790, opened 2026-08-19
Preserve dropout p / batch_norm eps,momentum across QAT train<->eval switches

### LINKED ISSUE #2980: PT2 Qat eval
### 🐛 Describe the bug

In QAT (Quantization-Aware Training), it is necessary to call torch.ao.quantization.allow_exported_model_train_eval(model) along with model.eval() and model.train().

This actually triggers patterns in torch/ao/quantization/pt2e/export_utils.py to replace dropout and batchnorm:
```
    def dropout_train(x):
            return F.dropout(x, p=0.5, training=True, inplace=inplace)

    def dropout_eval(x):
            return F.dropout(x, p=0.5, training=False, inplace=inplace)
    def bn_train(
        x: torch.Tensor,
        bn_weight: torch.Tensor,
        bn_bias: torch.Tensor,
        bn_running_mean: torch.Tensor,
        bn_running_var: torch.Tensor,
    ):
        return F.batch_norm(
            x, bn_running_mean, bn_running_var, bn_weight, bn_bias, training=True
        )

    def bn_eval(
        x: torch.Tensor,
        bn_weight: torch.Tensor,
        bn_bias: torch.Tensor,
        bn_running_mean: torch.Tensor,
        bn_running_var: torch.Tensor,
    ):
        return F.batch_norm(
            x, bn_running_mean, bn_running_var, bn_weight, bn_bias, training=False
        )
```
The above patterns will forcibly modify the p parameter of dropout and the momentum and eps(defaults 1e-5) parameters of batch_norm.

These parameter modifications directly affect the original model's training hyperparameters. I tried do PT2 QAT with YOLOv5, it was found that changes to eps (yolov5 configs 1e-3)  can cause a drop of one point in floating-point precision accuracy.

it confused me by several weeks. 

### Versions

torch2.6

cc @jerryzh168 @jianyuh @raghuramank100 @jamesr66a @vkuzo @jgong5 @Xia-Weiwen @leslie-fang-intel

amanyagami:
## Summary
`_replace_dropout`/`_replace_batchnorm` in `torchao/quantization/pt2e/export_utils.py` (used by `move_exported_model_to_eval`/`move_exported_model_to_train`, which QAT users call via `allow_exported_model_train_eval` + `model.train()`/`model.eval()`) rewrote the model's dropout/batch_norm aten nodes on every train↔eval switch.

The match patterns use `ignore_literals=True` so they match dropout/batch_norm calls with **any** `p`/`momentum`/`eps` — but the *replacement* patterns baked in hardcoded literals (dropout `p=0.5`, batch_norm's default `momentum=0.1`/`eps=1e-5`). So every train↔eval switch silently overwrote the model's actual configured values with those defaults. There was even a maintainer TODO acknowledging the batch_norm half of this:
```python
# TODO(Leslie): This function still fails to support custom momentum and eps value.
```
Reported to cause a real accuracy drop on a YOLOv5 model configured with `eps=1e-3`.

## Fix
Use `torch.fx.subgraph_rewriter`'s `replacement_callback` hook instead of a static replacement pattern. For each match, look up the actual matched node via `match.nodes_map`, read its real `p` (dropout) or `momentum`/`eps`/`cudnn_enabled` (batch_norm), and build the replacement graph with those values instead of the hardcoded ones — only the `training` flag is toggled.

## Test
Added two regression tests in `test/quantization/pt2e/test_quantize_pt2e.py`:
- `test_move_exported_model_dropout_preserves_custom_p` — `nn.Dropout(0.7)`, checked through export → eval → train.
- `test_move_exported_model_bn_preserves_custom_momentum_and_eps` — `nn.BatchNorm2d(eps=1e-3, momentum=0.3)`, same cycle.

- Verified both fail on the unpatched code (values silently reset to `0.5`/`0.1`/`1e-5`).
- Pass with the fix, across train→eval→train.
- The full `test/quantization/pt2e/test_quantize_pt2e.py` suite (72 tests, including the pre-existing `test_allow_exported_model_train_eval`/`test_move_exported_model_bn`/`test_move_exported_model_dropout*`) still passes.

Fixes #2980

### DIFF
--- a/test/quantization/pt2e/test_quantize_pt2e.py
+++ b/test/quantization/pt2e/test_quantize_pt2e.py
@@ -2317,6 +2317,41 @@ class TestQuantizePT2E(PT2EQuantizationTestCase):
     def test_move_exported_model_dropout_inplace(self):
         self._test_move_exported_model_dropout(inplace=True)
 
+    def test_move_exported_model_dropout_preserves_custom_p(self):
+        """
+        Regression test for https://github.com/pytorch/ao/issues/2980.
+
+        move_exported_model_to_eval/train used to always rewrite dropout's
+        `p` to a hardcoded 0.5, silently discarding any custom dropout rate.
+        """
+
+        class M(torch.nn.Module):
+            def __init__(self) -> None:
+                super().__init__()
+                self.dropout = torch.nn.Dropout(0.7)
+
+            def forward(self, x):
+                return self.dropout(x)
+
+        example_inputs = (torch.randn(1),)
+        m = M().train()
+        m = torch.export.export(m, example_inputs, strict=True).module()
+        target = torch.ops.aten.dropout.default
+
+        dropout_node = self._get_node(m, target)
+        self.assertEqual(dropout_node.args[1], 0.7)
+        self.assertTrue(dropout_node.args[2])
+
+        torchao.quantization.pt2e.move_exported_model_to_eval(m)
+        dropout_node = self._get_node(m, target)
+        self.assertEqual(dropout_node.args[1], 0.7)
+        self.assertFalse(dropout_node.args[2])
+
+        torchao.quantization.pt2e.move_exported_model_to_train(m)
+        dropout_node = self._get_node(m, target)
+        self.assertEqual(dropout_node.args[1], 0.7)
+        self.assertTrue(dropout_node.args[2])
+
     def _get_bn_train_eval_ops(self):
         return (
             torch.ops.aten.batch_norm.default,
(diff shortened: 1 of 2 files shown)


amanyagami:
@andrewor14 flagging this QAT one — `_replace_dropout`/`_replace_batchnorm` bake hardcoded literals (`p=0.5`, `momentum=0.1`, `eps=1e-5`) into the replacement pattern regardless of what the model actually uses, so a non-default dropout `p` or batch_norm `eps`/`momentum` gets silently reset on every train↔eval switch under `allow_exported_model_train_eval`. Fix plus a test, CI green.

amanyagami:

PR #4791, opened 2026-08-19
Fix dtype mismatch in Float8Linear backward when weight scaling is DISABLED

### LINKED ISSUE #4616: [Bug] DISABLED backward path crashes with dtype mismatch under autocast
### Description

`torchao/float8/float8_linear.py` backward crashes with `RuntimeError: expected scalar type BFloat16 but found Float` when backward scaling is `ScalingType.DISABLED` and the model runs under `torch.autocast(dtype=bfloat16)`.

**Line 128-129** — DISABLED branch uses saved `weight_hp_t` directly:
```python
elif c.cast_config_weight_for_grad_input.scaling_type is ScalingType.DISABLED:
    weight_t_maybe_fp8_dim0 = weight_hp_t  # float32 (master weight)
```

**Line 143-145** — `torch.mm` with mismatched dtypes:
```python
grad_input = torch.mm(
    grad_output_reshaped_maybe_fp8_dim0,  # bf16 (from autograd under autocast)
    weight_t_maybe_fp8_dim0.t(),          # float32 ← mismatch
)
```

Under autocast, `Float8Linear.forward` receives `input` as bf16 and `self.weight.t()` as float32 (master weight). Both are saved via `ctx.save_for_backward(input_hp, weight_hp_t)` — a (bf16, float32) pair. When backward scaling is active, `hp_tensor_to_float8_dynamic` casts both to float8, so the mismatch is masked. When DISABLED, the raw saved tensors go directly to `torch.mm`, which does not perform automatic dtype promotion.

### Proposed fix

```python
# Line 128-129: align dtype with grad_output
elif c.cast_config_weight_for_grad_input.scaling_type is ScalingType.DISABLED:
    weight_t_maybe_fp8_dim0 = weight_hp_t.to(
        grad_output_reshaped_maybe_fp8_dim0.dtype
    )
```

`.to(dtype)` is a no-op when dtypes already match (no autocast → both float32), so this has zero cost outside the autocast + DISABLED scenario. The `grad_weight` path is unaffected — both its operands (`grad_output` and `input_hp`) are bf16 under autocast.

### Steps to reproduce

```python
import torch
from torchao.float8 import Float8LinearConfig, CastConfig, ScalingType
from torchao.float8.float8_linear import Float8Linear

# Explicitly DISABLE all backward scaling
disabled = CastConfig(scaling_type=ScalingType.DISABLED)
config = Float8LinearConfig(
    cast_config_grad_output=disabled,
    cast_config_weight_for_grad_input=disabled,
    cast_config_input_for_grad_weight=disabled,
    cast_config_grad_output_for_grad_weight=disabled,
)

layer = Float8Linear(64, 64, config=config).cuda()
x = torch.randn(8, 64, device="cuda", requires_grad=True)

with torch.autocast("cuda", dtype=torch.bfloat16):
    y = layer(x)

y.sum().backward()
# → RuntimeError: expected scalar type BFloat16 but found Float


Without autocast (both operands float32), the same config works fine — confirming this is specifically a mixed-precision dtype alignment issue.
```

### Environment

```text
- torchao version: 0.18.0 (any version with `torchao.float8.float8_linear`)
- Hardware: reproduces on CUDA (no special hardware needed); affects any backend using `ScalingType.DISABLED` backward with autocast
- Files: `torchao/float8/float8_linear.py` line 128-129 (1 location, ~3 lines fix)
```

amanyagami:
## Summary
`matmul_with_hp_or_float8_args.backward`'s `grad_input` calculation, when `cast_config_weight_for_grad_input.scaling_type` is `ScalingType.DISABLED`, used the saved high precision weight (`weight_hp_t`) as-is in `torch.mm` against `grad_output`. Under autocast, `Float8Linear.forward` casts its input (and therefore the output, and therefore `grad_output` during backward) to the autocast dtype (e.g. bf16), while `weight_hp_t` — the fp32 master weight — is saved untouched. `torch.mm` does not promote dtypes, so this crashes with:
```
RuntimeError: expected m1 and m2 to have the same dtype, but got: c10::BFloat16 != float
```

## Fix
As proposed in the issue: cast `weight_hp_t` to `grad_output`'s dtype in the DISABLED branch. `.to(dtype)` is a no-op outside autocast (both already the same dtype), so this has zero cost outside the autocast + DISABLED scenario. The `grad_weight` path needs no change — both its operands are already the autocast dtype in that scenario.

## Test
Added `test_backward_disabled_grad_input_dtype_mismatch` in `test/float8/test_base.py`. It reproduces the exact dtype mismatch directly via `matmul_with_hp_or_float8_args` — constructing `input_hp` as bf16 (mirroring what an autocast region produces) and `weight_hp_t` as fp32 (the master weight) is sufficient to trigger the same mismatch on **any device**, no autocast context or CUDA required.

- Verified it fails with the issue's exact `RuntimeError` on the unpatched code.
- Passes with the fix, and both `input_hp.grad`/`weight_hp_t.grad` come back with the correct dtypes.
- The full `test/float8/test_base.py` suite still passes (28 tests runnable without CUDA, 98 CUDA-gated and cleanly skipped).

Fixes #4616

### DIFF
--- a/test/float8/test_base.py
+++ b/test/float8/test_base.py
@@ -320,6 +320,53 @@ class TestFloat8Linear:
         if m_ref.bias is not None:
             torch.testing.assert_close(m_ref.bias.grad, m_fp8.bias.grad)
 
+    def test_backward_disabled_grad_input_dtype_mismatch(self):
+        """
+        Regression test for https://github.com/pytorch/ao/issues/4616.
+
+        matmul_with_hp_or_float8_args.backward's grad_input calculation, when
+        cast_config_weight_for_grad_input.scaling_type is DISABLED, used the
+        saved high precision weight (weight_hp_t) as-is in torch.mm against
+        grad_output. Under autocast, Float8Linear.forward casts its input
+        (and therefore the output, and therefore grad_output during
+        backward) to the autocast dtype (e.g. bf16), while weight_hp_t - the
+        fp32 master weight - is saved untouched. torch.mm does not promote
+        dtypes, so this crashed with:
+            RuntimeError: expected m1 and m2 to have the same dtype, ...
+
+        This reproduces the dtype mismatch directly via
+        matmul_with_hp_or_float8_args, without needing autocast or CUDA:
+        constructing input_hp as bf16 (mirroring what an autocast region
+        would produce) and weight_hp_t as fp32 (the master weight) is
+        sufficient to trigger the same mismatch on any device.
+        """
+        from torchao.float8.float8_linear import matmul_with_hp_or_float8_args
+
+        disabled = CastConfig(scaling_type=ScalingType.DISABLED)
+        config = Float8LinearConfig(
+            emulate=True,
+            cast_config_grad_output=disabled,
+            cast_config_weight_for_grad_input=disabled,
+            cast_config_input_for_grad_weight=disabled,
+            cast_config_grad_output_for_grad_weight=disabled,
+        )
+
+        input_hp = torch.randn(8, 64, dtype=torch.bfloat16, requires_grad=True)
+        weight_hp_t = torch.randn(64, 64, dtype=torch.float32, requires_grad=True)
+
+        y = matmul_with_hp_or_float8_args.apply(
+            input_hp, weight_hp_t, LinearMMConfig(), config
+        )
+        assert y.dtype == torch.bfloat16
+        # Used to raise:
+        #   RuntimeError: expected m1 and m2 to have the same dtype, but
+        #   got: c10::BFloat16 != float
+        y.sum().backward()
+        assert input_hp.grad is not None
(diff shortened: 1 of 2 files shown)


amanyagami:
@vkuzo this one is 8 lines in `float8_linear.py` plus a test, and still reproduces on `main`.

In the `ScalingType.DISABLED` branch of the backward, `weight_t_maybe_fp8_dim0` is set to the saved high-precision weight as-is. Under autocast that can be fp32 while `grad_output_reshaped_maybe_fp8_dim0` is the autocast dtype (bf16), and the `torch.mm` below does not promote — so it raises on dtype mismatch. Outside autocast both are already the same dtype, so the `.to(...)` is a no-op there.

Only the DISABLED path is affected, which is presumably why it has survived.

amanyagami:

PR #4793, opened 2026-08-19
Mask a_s/b_s scale loads in blockwise FP8 training GEMM kernels

### LINKED ISSUE #4634: Blockwise FP8 training GEMMs read past the end of both scale tensors
### Description

Both Triton GEMMs in `torchao/prototype/blockwise_fp8_training/kernels.py` load their scale tensors without a bounds mask, while every other load and the final store in those kernels is masked. When the autotuner's tile does not divide `M` or `N`, the scale loads dereference memory past the end of `a_s` / `b_s`.

| kernel | unmasked loads |
| --- | --- |
| `triton_fp8_gemm_1x128_128x128_kernel` | `kernels.py:250-251` |
| `triton_fp8_gemm_1x128_128x1_kernel` | `kernels.py:368-369` |

The GEMM output is correct — the out-of-range lanes are dropped by `c_mask` at the store — so this is a memory-safety bug, not a numerical one. It is usually hidden by slack in the caching allocator's slabs, and surfaces as an intermittent `CUDA error: unspecified launch failure` when a scale tensor sits at the tail of a mapped region.

Present on `main` at `03ca489d`.

In the code below, it passes when run normally — which is the problem:

```
$ python repro.py
ok, no NaNs: True
```

Under the sanitizer it faults. `PYTORCH_NO_CUDA_MEMORY_CACHING=1` gives each scale tensor its own `cudaMalloc`, so the stray read is not absorbed by slack after the tensor:

```
$ PYTORCH_NO_CUDA_MEMORY_CACHING=1 compute-sanitizer --tool memcheck python repro.py
========= Invalid __global__ read of size 4 bytes
=========     at triton_fp8_gemm_1x128_128x128_kernel+0xfa0 in kernels.py:251
=========     by thread (64,0,0) in block (0,0,0)
=========     Access to 0x7fa99a418204 is out of bounds
=========     and is 1 bytes after the nearest allocation at 0x7fa99a418200 of size 4 bytes
=========         Host Frame: triton_fp8_gemm_1x128_128x128 in kernels.py:291
========= ERROR SUMMARY: 68 errors
```

### Steps to reproduce

```python
# repro.py
import torch

from torchao.prototype.blockwise_fp8_training.kernels import (
    triton_fp8_blockwise_act_quant_lhs,
    triton_fp8_blockwise_weight_quant_transposed_rhs,
    triton_fp8_gemm_1x128_128x128,
)

M = N = K = 128

A = torch.randn(M, K, dtype=torch.bfloat16, device="cuda")
B = torch.randn(N, K, dtype=torch.bfloat16, device="cuda")

A_q, A_s = triton_fp8_blockwise_act_quant_lhs(A)
B_t_q, B_t_s = triton_fp8_blockwise_weight_quant_transposed_rhs(B)

C = triton_fp8_gemm_1x128_128x128(A_q, B_t_q, A_s, B_t_s, out_dtype=torch.bfloat16)
torch.cuda.synchronize()
print("ok, no NaNs:", not C.isnan().any().item())
```

### Environment

```text
- NVIDIA B200 (SM100)
- torch 2.13.0+cu130, triton 3.7.1
- `compute-sanitizer` from CUDA 13.1.1
- torchao `main` at `03ca489d`
```

amanyagami:
## Summary
Both `triton_fp8_gemm_1x128_128x128_kernel` and `triton_fp8_gemm_1x128_128x1_kernel` in `torchao/prototype/blockwise_fp8_training/kernels.py` load their per-tile `a_s`/`b_s` scale values without a bounds mask, while every other load and the final store in these kernels is masked. When the autotuner's tile does not evenly divide `M` or `N`, the scale loads for the tail tile dereference memory past the end of `a_s`/`b_s`.

The GEMM's numeric output is unaffected — the out-of-range lanes are dropped by `c_mask` at the store — so this is a memory-safety bug, not a numerical one. The issue reports it via compute-sanitizer as an `Invalid __global__ read`, which intermittently surfaces as `CUDA error: unspecified launch failure` depending on allocator slack.

## Fix
`mask=offs_m < M` for the `a_s` load, `mask=offs_n < N` for the `b_s` load — matching the bounds every sibling load (`a_mask`, `b_mask`) and the final store (`c_mask`) already use in the same kernels.

## Test
These kernels require CUDA to run (real `@triton.jit` GEMMs targeting H100/B200), so I can't exercise them in this environment. Instead, added `test/prototype/blockwise_fp8_training/test_gemm_scale_bounds.py`, which isolates the exact `offs_m`/`offs_n` → scale-index arithmetic in plain Python for tile geometries where `M`/`N` are **not** a multiple of the block size (the exact case the issue reports), and checks:
- unmasked indices do go out of `[0, M)`/`[0, N)` for these shapes (reproducing the bug's precondition)
- masked indices (`mask=offs_m < M` / `mask=offs_n < N`) never do, and still cover every valid row/col at least once

This proves the fix's indices are correct independent of actually running the Triton kernel. End-to-end verification (e.g. under `compute-sanitizer`, per the issue's own repro) still needs GPU CI — flagging this explicitly.

Fixes #4634

### DIFF
--- a/test/prototype/blockwise_fp8_training/test_gemm_scale_bounds.py
+++ b/test/prototype/blockwise_fp8_training/test_gemm_scale_bounds.py
@@ -1,0 +1,92 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Regression test for https://github.com/pytorch/ao/issues/4634.
+
+Both Triton GEMMs in torchao/prototype/blockwise_fp8_training/kernels.py
+(`triton_fp8_gemm_1x128_128x128_kernel` and
+`triton_fp8_gemm_1x128_128x1_kernel`) load their per-tile `a_s`/`b_s` scale
+values without a bounds mask, while every other load and the final store in
+those kernels is masked. When the autotuner's tile does not evenly divide
+`M` or `N`, the scale loads dereference memory past the end of `a_s`/`b_s`
+for the tail tile - a real out-of-bounds read reported via
+compute-sanitizer in the issue (the GEMM's numeric output happens to still
+be correct, since `c_mask` drops those lanes at the store; this is a
+memory-safety bug, not a numerical one).
+
+These kernels require CUDA to run (they're `@triton.jit`-compiled GEMMs
+targeting an H100/B200), so they can't be exercised in this environment.
+Instead, this test isolates the exact `offs_m`/`offs_n` -> scale-index
+arithmetic in plain Python for tile geometries where `M`/`N` are *not* a
+multiple of the block size, and checks whether any program instance's scale
+load index falls outside the valid `[0, M)`/`[0, N)` range - i.e. whether
+the load the kernel issues is actually in-bounds, mirroring
+`mask=offs_m < M` / `mask=offs_n < N`.
+"""
+
+import itertools
+
+from torch.testing._internal.common_utils import TestCase, run_tests
+
+
+def _scale_load_indices(M, N, block_size_m, block_size_n):
+    """Return the set of row indices used for the `a_s` load (indexed by
+    `offs_m`) and column indices used for the `b_s` load (indexed by
+    `offs_n`), across every program instance the kernel launches - exactly
+    the indices `tl.load(..., mask=offs_m < M)` / `mask=offs_n < N` would
+    guard.
+    """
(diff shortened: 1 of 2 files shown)


amanyagami:
@vkuzo similar area to #4791/#4792 — masks the `a_s`/`b_s` scale loads in the blockwise FP8 training GEMM kernels so an out-of-bounds tile reads garbage scale instead of being masked off. Small diff plus a test, CI green, still reproducing on `main`.

amanyagami:

Issue #3637, opened 2026-01-14
Update documents addressing AQT workflow


namgyu-youn:
AQT (`AffineQuantizedTensor`) is a deprecated workflow, but several documents still reference it. We should either update these documents to reflect current best practices or deprecate them.

- [x] API Reference Dtype: Removed since https://docs.pytorch.org/ao/main/api_ref_quantization.html supersedes it.
	- Source: https://github.com/pytorch/ao/blob/main/docs/source/api_ref_dtypes.rst 
	- Rendering: https://docs.pytorch.org/ao/main/api_ref_dtypes.html
	- fixed at: #3639 

- [ ] Static Quantization: Update to use `Int8StaticActivationInt8WeightConfig` after #3687 land.
	- Source: https://github.com/pytorch/ao/blob/main/docs/source/static_quantization.rst
	- Rendering: https://docs.pytorch.org/ao/main/static_quantization.html

jerryzh168:
(1) sounds good to drop dtypes page
(2) I think for static quant, we could add an official API similar to AWQ/SmoothQuant actually (observer should be configurable for each tensor subclass instead of hardcoded like AWQ/SmoothQuant, can use protocol for that as well I think), that should be able to compose with any tensor subclass that supports this, it's probably better to work on that first, it could also simplify the tutorial

namgyu-youn:
> (1) sounds good to drop dtypes page

https://github.com/pytorch/ao/pull/3639 for the first one, could you look into it?

namgyu-youn:
> (2) I think for static quant, we could add an official API similar to AWQ/SmoothQuant actually (observer should be configurable for each tensor subclass instead of hardcoded like AWQ/SmoothQuant, can use protocol for that as well I think), that should be able to compose with any tensor subclass that supports this, it's probably better to work on that first, it could also simplify the tutorial

Do you mean calibration logic (observers, scale computation, `act_pre_scale`) is handled by the API layer via protocol, and subclass should not handle them (we are storing `act_pre_scale` inside subclass)? If so, would it be decomposed to: 1) static quant protocol, 2) subclass (`Int4Tensor/Int8Tensor`), 3) observer (AWQ/SmoothQuant), 4) API layer? Let me know if I understood right

jerryzh168:
[...]

PR #4794, opened 2026-08-19
Fix broken example code in the static quantization tutorial

amanyagami:
## Summary
`docs/source/eager_tutorials/static_quantization.rst`'s `QuantizedLinear` example had three compounding bugs that made it non-functional for any reader following it verbatim (tracked under #3637, updating docs that reference outdated workflows):

1. `QuantizedLinear.from_observed()` passed an extra `target_dtype` argument into `cls(...)` that `QuantizedLinear.__init__` never accepted (it's unused) — raised:
   ```
   TypeError: QuantizedLinear.__init__() takes 7 positional arguments but 8 were given
   ```

2. `Int8Tensor.from_hp`'s static-scale path asserts `scale.ndim == hp_tensor.ndim`, but `AffineQuantizedMinMaxObserver.calculate_qparams()` returns scale/zero_point with the block dims squeezed out (e.g. shape `(64,)` for a per-row scale on a `(64, 64)` weight, not `(64, 1)`) — raised:
   ```
   AssertionError (in Int8Tensor.from_hp)
   ```

3. Once 1–2 are fixed, `forward()` manually built a *second* `Int8Tensor` for the activation and passed both quantized tensors into `F.linear`. `Int8Tensor`'s `F.linear` dispatch has no support for a pre-quantized activation tensor — only a plain high precision one, optionally paired with the *weight* tensor's `act_quant_kwargs`/`act_quant_scale`/`act_quant_zero_point` fields for static (fixed-scale) activation quantization. This raised:
   ```
   NotImplementedError: Int8Tensor dispatch: attempting to run unimplemented operator/function: func=aten.view.default, ...
   ```

## Fix
Fold the calibrated activation scale/zero_point into the weight tensor via `act_quant_kwargs`/`act_quant_scale`/`act_quant_zero_point` on `Int8Tensor.from_hp` — the currently supported way to do static activation quantization — and have `forward()` pass the raw activation straight into `F.linear`; `Int8Tensor`'s dispatch quantizes it internally using the fixed statistics. Also fixed the reshape needed for bug 2, dropped the dead `target_dtype` argument for bug 1, and updated the tutorial's illustrative REPL output (repr type name, tensor shape, and where the activation scale now lives) to match reality.

## Test
Added `test/quantization/test_static_quantization_doc_example.py`, which mirrors this example (minus `torch.compile`/CUDA, so it runs on CPU) end to end.
- Verified it reproduces the original `TypeError` on the unpatched example code (and would hit the other two errors in sequence if that one were papered over — traced each independently while debugging).
- The full corrected pipeline runs cleanly and produces a NaN-free, correctly-shaped output with the fix.
- Verified the edited `.rst` still parses cleanly with `docutils`.

This leaves one remaining checklist item open in #3637 (updating the linked #3687, which was closed without merging) — that's a separate, larger scope change (a new `Int8StaticActivationInt8WeightConfig`) rather than a doc bug, so I've left it for a maintainer to scope.

### DIFF
--- a/docs/source/eager_tutorials/static_quantization.rst
+++ b/docs/source/eager_tutorials/static_quantization.rst
@@ -145,6 +145,9 @@ There are multiple ways to actually quantize the model. Here we walk through the
 
    from torchao.quantization import Int8Tensor
    from torchao.quantization import PerRow, PerTensor
+   from torchao.quantization.quantize_.workflows.int8.int8_tensor import (
+       QuantizeTensorToInt8Kwargs,
+   )
 
    class QuantizedLinear(torch.nn.Module):
        def __init__(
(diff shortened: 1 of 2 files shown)


amanyagami:
Docs-only fix — the `QuantizedLinear` example in the static quantization tutorial has three compounding bugs that make it fail for anyone following it verbatim (tracked under #3637). CI green, would appreciate a look whenever convenient.

amanyagami:

PR #4795, opened 2026-08-19
Delete unused ParetoQ prototype

iamzainhuda:
## Summary

- delete the self-contained `torchao.prototype.paretoq` research snapshot
- delete its dedicated test, which is the package's only external consumer
- remove the package-specific Ruff exclusion
- retain the unrelated ParetoQ paper citation in CUTLASS publications

ParetoQ has no package exports and no consumers in TorchAO source, docs, examples, benchmarks, or CI beyond its own test. The implementation is a copied Transformers training stack with cwd-sensitive imports and a pinned dependency file, and it has not received feature maintenance since its initial import.

This removes 2,244 package text LOC, 52 test LOC, and approximately 3.65 MB of image assets. The already-merged broken eval launcher cleanup in #4774 is part of the base.

## Test plan

- `ruff check .`
- `pytest --collect-only -q test/prototype` (5,460 tests collected)
- package import smoke test
- verify no tracked TorchAO references to the ParetoQ package or its deep symbols remain
- `git diff --check`


iamzainhuda:
@claude review

andrewor14 (approved):

andrewor14:
@liuzechun Just a heads up, we're cleaning up torchao to delete unused / inactive features. Please let us know if you have any concerns about deleting ParetoQ from torchao. I think we prefer to redirect users to https://github.com/facebookresearch/ParetoQ since that's more actively maintained.

PR #4046, opened 2026-03-11
[xpu][test] Port test/prototype/quant_logger/test_quant_logger.py to Intel GPU

zxd1997066:
For https://github.com/pytorch/ao/issues/2917, This PR is targeted to port test/prototype/quant_logger/test_quant_logger.py to Intel GPU.

PR #4796, opened 2026-08-19
Delete unused quantization logger prototype

iamzainhuda:
## Summary

- delete `torchao.prototype.quant_logger` and its seven exported APIs
- delete the package's dedicated test and four documentation examples
- remove its API reference, tutorial, and both toctree entries

The package has no production consumers outside its own test/documentation island. This removes 310 package LOC and 787 lines repository-wide. The old open XPU test-port PR #4046 only changes the now-deleted dedicated test and does not represent a downstream consumer.

## Test plan

- `ruff check .`
- `pytest --collect-only -q test/prototype` (5,455 tests collected; only the five deleted quant-logger tests are absent)
- package import smoke test
- verify no tracked references to the package, API page, or tutorial remain
- `git diff --check`

A local Sphinx build was not available because Sphinx is not installed in this environment; the removed pages were also removed from both toctrees.


iamzainhuda:
@claude review

andrewor14 (approved):

PR #4798, opened 2026-08-20
[nvfp4_training] Optimize CuteDSL kernels for NVFP4 training

rdspring1:
# Summary

Brings the CuteDSL NVFP4 quantization kernels to parity-or-better with the Triton backend across the DeepSeek-V3 shapes, adds a fast-math path to **both** backends, and promotes CuteDSL + fast math to the default via `KernelPreference.AUTO`.

CuteDSL now leads Triton by **1.6x on grouped RTNE, 2.1x on grouped SR, 1.7-1.8x on the grouped amax, and 1.8x on grouped 2D weights** at 671B, and fast math is worth a further 1.3-1.4x on top. Round-to-nearest-even output stays bitwise identical to both the Triton backend and the TransformerEngine-derived reference on every path.

The latest configuration is NVFP4 for all dense linears, shared and routed experts and MXFP8 attention with fast math enabled. In E2E run with 64 GB300 GPUs for 1 hour, NVFP4 is **37.3% faster than BF16 and +12.8% faster than MXFP8** (1004.6 vs 731.6 and 890.8 TFLOP/s) while using less memory than either (217.08 GiB, −8.38 GiB vs bf16 and −5.41 GiB vs mxfp8). The training loss curve is inline with BF16 and MXFP8.

<details>
<summary><strong>DSV3 671B E2E Run, GBS 512, 64 GB300 GPUs, Training Loss, Throughput, and Memory</strong></summary>

image

**Throughput** (steps >= 50):

| cell | TFLOP/s/GPU | tok/s/GPU | vs bf16 | vs mxfp8 |
|---|---:|---:|---:|---:|
| BF16 | 731.6 | 2602 | — | −17.9% |
| NVFP4 F0L10, BF16 ATTN, Triton, PR 4598 | 847.4 | 3014 | +15.8% | −4.9% |
| MXFP8 | 890.8 | 3168 | +21.8% | — |
| NVFP4 F0L0 + MXFP8 ATTN, CuteDSL exact | 989.0 | 3518 | +35.2% | +11.0% |
| **NVFP4 F0L0 + MXFP8 ATTN, CuteDSL fast** | **1004.6** | **3573** | **+37.3%** | **+12.8%** |

**Memory** (peak reserved per GPU, 276 GiB device):

| cell | peak | of 276 GiB | vs bf16 | headroom |
|---|---:|---:|---:|---:|
| nvfp4 F0L10, bf16 attn, Triton, PR 4598| 227.03 GiB | 82.07% | +1.57 GiB | 48.97 GiB |
| bf16 | 225.46 GiB | 81.50% | — | 50.54 GiB |
| mxfp8 | 222.49 GiB | 80.43% | −2.97 GiB | 53.51 GiB |
| nvfp4 F0L0 + mxfp8 attn, CuteDSL exact | 217.08 GiB | 78.48% | −8.38 GiB | 58.92 GiB |
| **nvfp4 F0L0 + mxfp8 attn, CuteDSL fast** | **217.08 GiB** | **78.48%** | **−8.38 GiB** | **58.92 GiB** |

</details>

### Key Changes

1. Replace random bit generation in Linear CuteDSL kernels from MurmurHash3 to Philox
2. Add fast-math path to Triton and CuteDSL
3. Create CuteDSL versions of the grouped RHT quantize, grouped amax, and grouped 2D weight kernels
4. Enable CuteDSL and fast math as the default

# Details

## Fast Math Path

Fast math matches TransformerEngine under `NVTE_USE_FAST_MATH=1` and does two things:

- **Skips the FP32 -> BF16 -> FP32 round-through of the tcgen05 RHT accumulator.** Gated per call site: it applies to the columnwise path only, since the rowwise path has no accumulator.
- **Replaces the correctly rounded `div.rn` encode reciprocal with `rcp.approx.ftz.f32`** (one `MUFU.RCP`).

Triton previously had no fast variant at all. Adding it keeps the two backends bitwise identical to each other **and** to TE in either mode, so `AUTO` and `TRITON` agree.

Weight quantization is deliberately untouched: without an RHT there is no accumulator to skip, and TE has no 2D fast path either (`quantize_transpose_nvfp4.cuh` has zero `use_fast_math` references). The amax kernels are always exact, in TE and CuteDSL both.

## KernelPreference.CuteDSL and Fast Math as the default

`NVFP4TrainingConfig.kernel_preference` moves `TRITON` -> `AUTO` and `use_fast_math` defaults `True`. `AUTO` resolves to CuteDSL on SM100 with the CuteDSL runtime present and falls back to Triton otherwise.

Both defaults are documented in the config with the exact settings that recover prior numerics (`kernel_preference=KernelPreference.TRITON, use_fast_math=False`), so a loss-curve diff across this branch does not silently carry two moved variables.

Fast-vs-exact measures 30.2-32.6 dB SQNR on the linear path and 30.4-31.9 grouped — roughly 10 dB quieter than NVFP4's own ~20 dB quantization noise. The error is large and rare rather than small and uniform (~1.2% of code bytes differ): skipping the bf16 round-through moves a value by ~2**-9, but an element near an E2M1 midpoint then flips a whole FP4 step.

## Philox divergence between CuteDSL and Triton

**Triton and CuteDSL are bitwise identical for RTNE but not for SR**

Triton's SR path calls `tl.randint(seed, offset)` once per packed byte. Two things waste work there:

- `tl.randint` computes a full Philox-4x32-10 draw (4 words) and returns **one** of them.
- The `cvt.rs` conversion goes through `tl.inline_asm_elementwise` with `pack=4`, and `pack` applies uniformly to every argument. So the asm block receives **4** random words while its two `cvt.rs.satfinite.e2m1x4.f32` instructions consume only **2** (`$9`, `$10`) — the other two are computed and discarded. There is no way to request a different pack factor for the random-bit argument than for the data arguments.

CuteDSL is not bound by that shape. `philox4_all` draws **one counter per 16-element block and consumes all four output words**, computing all four in the last round: **34 multiplies per block instead of 124**. The counter is derived from tile coordinates rather than a running per-thread value, because the persistent CLC scheduler's visit order is not fixed — so results stay reproducible for a given `rng_state`.

The consequence: CuteDSL's SR stream is a *different, equally valid* stream, not a matching one. It is validated on properties rather than bit-equality — every SR code within one FP4 magnitude step of the bitwise-checked RTNE code, reconstruction SQNR, unbiasedness at exact-halfway values, and `rng_state` reproducibility.

This is the one behavioural asymmetry in `AUTO`: on a node without the CuteDSL runtime, the silent Triton fallback changes backward-pass numerics under SR, not just which kernel runs. Pin `kernel_preference` explicitly for runs that must reproduce bitwise across machines.

# Performance: CuteDSL vs Triton

GB200, CUDA 13.4, PyTorch 2.15.0a0+git0f3e7e2, Triton 3.8.0, nvidia-cutlass-dsl 4.5.2.
Device kernel self-time in microseconds, median of three full script passes at 15 warmups and 50 timed profiler iterations. 
`E = 4` local experts, the EP=64 training layout.

Reproducible with:

```bash
python -m benchmarks.prototype.nvfp4_training.bench_group_rht_quantize_row_col --experts 4 --math all
python -m benchmarks.prototype.nvfp4_training.bench_group_hadamard_amax --experts 4
python -m benchmarks.prototype.nvfp4_training.bench_group_quantize_2d
```

<details>
<summary><strong>DeepSeek-V3 671B with 4 local experts</strong></summary>

### Grouped RHT quantize — round-to-nearest-even

| projection | math | CuteDSL | Triton | speedup | CuteDSL fast/exact |
|---|---|---:|---:|---:|---:|
| gate/up (w1/w3) | standard | 53.95 | 87.49 | **1.62x** | — |
| gate/up (w1/w3) | fast | 39.73 | 64.16 | **1.62x** | 1.36x |
| down (w2) | standard | 55.23 | 86.91 | **1.57x** | — |
| down (w2) | fast | 41.14 | 63.71 | **1.55x** | 1.34x |

### Grouped RHT quantize — stochastic rounding

| projection | math | CuteDSL | Triton | speedup | CuteDSL fast/exact |
|---|---|---:|---:|---:|---:|
| gate/up (w1/w3) | standard | 74.75 | 159.96 | **2.14x** | — |
| gate/up (w1/w3) | fast | 54.93 | 141.28 | **2.57x** | 1.36x |
| down (w2) | standard | 76.51 | 159.30 | **2.08x** | — |
| down (w2) | fast | 56.51 | 140.86 | **2.49x** | 1.35x |

The SR lead is larger than the RTNE lead precisely because of the Philox change: SR costs
CuteDSL ~1.4x its own RTNE time against Triton's ~1.8x.

### Grouped amax and 2D weights (no fast-math variant — see above)

| kernel | projection | CuteDSL | Triton | speedup |
|---|---|---:|---:|---:|
| grouped RHT amax | gate/up | 22.95 | 39.91 | **1.74x** |
| grouped RHT amax | down | 22.42 | 40.54 | **1.81x** |
| grouped 2D weight | gate/up | 59.76 | 109.31 | **1.83x** |
| grouped 2D weight | down | 60.02 | 107.99 | **1.80x** |

### Fast-math value by backend

| path | CuteDSL fast/exact | Triton fast/exact |
|---|---|---|
| grouped quantize, RTNE | 1.32-1.38x | 1.28-1.36x |
| grouped quantize, SR | 1.27-1.36x | 1.13-1.14x |

Under RTNE both backends gain about the same. The split appears under SR, where Triton
gains only 1.13-1.14x: the Philox work it still carries dominates what fast math removes.

</details>

## AI tools used
- Claude Opus 5 (1M context) and OAI GPT 5.6 via Codex


rdspring1:
@vkuzo After talking with MLPerf folks, I've changed the recipe to quantize all the FFN layers with NVFP4 while the Attention layers follow the MXFP8 recipe in TorchTitan. I'm currently running DSV3 16B 1500 step run from [Experiment 2: DeepSeek-V3 16B MoE Loss Convergence Validation](https://pytorch.org/blog/enabling-up-to-41-faster-pre-training-mxfp8-and-deepep-for-deepseek-v3-on-b200-with-torchtitan/).

Can you advise on how you want users to pick between (CuteDSL, Triton) and (Standard, Fast-Math)? At the moment, this PR has `KernelPreference.Auto` and fast-math set as defaults to make the convergence runs easier.

vkuzo:
> Triton's SR path calls tl.randint(seed, offset)

How about using `tl.randint4x`?

rdspring1:
`tl.randint` already is `tl.randint4x` under the hood. It computes 4 random numbers via philox but only returns the first one. Swapping the call changes nothing by itself.

The ideal cost for quantizing 16 fp32 to fp4 is 1 Philox-4x32-10 invocation for 4 `cvt.rs`.
- `cvt.rs.satfinite.e2m1x4.f32 half, {$a,$b,$c,$d}, $r` consumes 4 fp32 and one 32-bit register of random bits. It writes one `.b16 = 4 fp4 (e2m1)`.
- One Philox invocation yields four 32-bit registers, so 4 `cvt.rs` requires 1 Philox invocation.

Quantizing 16 fp32 to fp4 yields 8 packed uint8. A `tl.inline_asm_elementwise` with `pack=4` covers 4 packed uint8 = 2 `cvt.rs`, so the 8 take 2 asm blocks = 4 `cvt.rs`. Triton runs 8 philox invocations because `offset` is per packed uint8, so 32 random numbers are produced of which 4 are read.

There are two separate areas that waste random numbers. `tl.randint` discards 3 out of 4, leaving 8 random numbers. Then `pack=4` gives each asm block 4 output elements, so rbits contributes operands `$9..$12`, but the two `cvt.rs` reference only `$9` and `$10` while `$11` and `$12` sit in the constraint string. They stay live so their Philox invocations still execute.

`tl.randint4x` can't close this because `tl.inline_asm_elementwise` is a map where each output element computes its own value independently. CuteDSL has no such problem because a thread just owns 16 values, calls Philox once, and uses all four random numbers as plain registers. The limitation isn't in the RNG generation. It's that Triton's tensor model can't express counter sharing across lanes.

rdspring1:
@vkuzo I completed the following experiments using the `KernelPreference.CuteDSL` and with `fast_math` enabled.

# DeepSeek-V3 with NVFP4 on GB300 — Convergence, Throughput, Power

Three experiments against a BF16 and MXFP8 baseline. All runs use torchtitan and torchao with pinned revisions. For each experiment every arm shares one container, so the precisions are arm-locked. The only intended difference is the quantization converters. The NVFP4 quantization scheme is NVFP4 F0L0 FFN + MXFP8 ATTN.

- **Experiment 1 — 16B, 1500 steps, convergence.** NVFP4 trains **15.4% faster** and holds **12.6% less memory** than BF16 but finishes **+0.050 nats** higher in loss, while MXFP8 gives up **7.8% throughput** for only **+0.010 nats**.
- **Experiment 2 — 671B, 256 GPUs, throughput.** At M = 131,072 NVFP4 delivers **+13.8% throughput** over MXFP8 and **10.79 GiB less** memory per GPU, with no arm showing instability.
- **Experiment 3 — 671B, 64 GPUs, power.** NVFP4 reaches **1.65× BF16 perf/watt** — 37.6% more throughput while drawing 16.6% *less* power, i.e. **39.4% less energy per token** — with MXFP8 at 1.31×.

---

## Experiment 1 — DSV3 16B model -- Training Convergence for 1500 steps

Does NVFP4 converge like MXFP8 and BF16 when routing is held fixed?

loss-fixed-routing

| arm | final loss |
| --- | ---: |
| BF16 | **3.10751** | — |
| MXFP8 | 3.11726 | +0.00975 |
| NVFP4 | 3.15782 | +0.05031 |

All three reached step 1500 with **zero NaN**. NVFP4 is makes a real trade: +0.050 nats for +15.4% throughput and 17.9 GiB memory savings. Switching some layers to higher precision could reduce the training loss gap.

<details>
<summary>Experiment 1 configuration</summary>

| | |
| --- | --- |
| model | DeepSeek-V3 16B MoE, 15,706,484,224 params (dense 858M / sparse 14.85B / active 2.66B) |
| hardware | GB300, 16 nodes × 4 = 64 GPUs, 276.62 GiB each |
| mesh | pp 1, dp_replicate 1, dp_shard 64, cp 1, tp 1, **ep 32** |
| batch | local 32, **global 4096 sequences**, grad accum 2, seq_len 8192 |
| schedule | 1500 steps, warmup 200, lr 2.2e-4, AdamW fused, betas (0.9, 0.95), wd 0.1 |
| routing | **forced round-robin** (every expert gets an identical token count) |
| AC / compile | FullAC, `torch.compile` per TransformerBlock |
| dataset | c4_test, `c4/en_train_3gb` |
| HybridEP | hidden 2048, max_tokens_per_rank 262144, 2 local experts, ep_size 32 |
| MXFP8 arm | GroupedExperts → dynamic `mxfp8_rceil` for `grouped_mm`; Linear → MXFP8Linear |
| NVFP4 arm | GroupedExperts → dynamic NVFP4 for `grouped_mm`; Linear → MXFP8Linear **and** NVFP4Linear |

Routing is **forced round-robin**, which removes expert specialisation. It makes grouped-GEMM shapes deterministic and is what the throughput comparison wants, but the router never learns to route — so these losses are not a statement about end-task quality.

</details>

---

## Experiment 2 — DSV3 671B model -- Throughput at 256 GPUs

| precision | TFLOP/s/GPU | tokens/s/GPU | peak memory |
| --- | ---: | ---: | ---: |
| MXFP8 | 889.47 | 3,164 | 164.54 GiB |
| NVFP4 | **1,012.34** | 3,601 | **153.75 GiB** |
| | **+13.8%** | | **−10.79 GiB** |

NVFP4 wins on both axes. At this batch size the 10.79 GiB is the difference between fitting a larger batch and not.

The margin is **strongly dependent on M**, the tokens entering one expert invocation. A separate sweep across M = 8,192 → 131,072 found NVFP4 moving from **−23.3%** (slower than MXFP8) to **+14.1%**, crossing over between M = 8,192 and 16,384. The row above is the top of that curve; it should not be quoted without its batch size. This parallelization scheme was selected because it gets the best performance with TorchTitan on GB300 measured July 2026.

<details>
<summary>Experiment 2 configuration</summary>

| | |
| --- | --- |
| model | DeepSeek-V3 671B |
| hardware | GB300, 64 nodes × 4 = **256 GPUs** |
| parallelism | EP 64, PP 1, TP 1, CP 1 |
| batch | `local_batch_size` 16, GBS 4096 sequences, seq_len 4096 |
| M | 131,072 tokens per expert invocation |
| steps | 50 |
| AC / compile | full activation checkpointing, `torch.compile` on model and loss |
| routing | forced load balance |

</details>

---

## Experiment 3 — DSV3 671B model -- Power at 64 GPUs

1 Hz per-GPU `nvidia-smi` sampling on every node, ~60 min per arm.

| precision | TFLOP/s/GPU | draw/GPU | 64-GPU total | % of 1400 W cap | SM clock | **TFLOP/s per W** | **J/token** |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| BF16 | 749.16 | 1,070.5 W | 68.51 kW | 76.5% | 1,946 MHz | 0.700 | 0.4017 |
| MXFP8 | 918.81 | 1,004.0 W | 64.26 kW | 71.7% | 2,057 MHz | 0.915 | 0.3072 |
| NVFP4 | **1,031.09** | **892.4 W** | **57.12 kW** | 63.7% | 2,067 MHz | **1.155** | **0.2434** |

**NVFP4 is 1.65× BF16 on perf/watt and it is not a trade** — 37.6% more throughput while drawing 16.6% *less* power. MXFP8 sits at 1.31×.

- **Nothing is cap-limited.** All three run well under 1400 W, NVFP4 at 63.7%. The differences are arithmetic efficiency, not throttling.
- **BF16 draws the most power at the lowest clock**, with GPU utilisation pinned at 99% on all three arms — more tensor-core work per unit time, not a stall.
- **Cross-GPU spread is ~1.1% per arm**, so no straggler skews a mean.

**These are GPU package draw only.** Switches, NVLink spine, CPUs, fans and PSU loss are invisible to `nvidia-smi`, and that is the share a rack-level comparison turns on. Those components are reported to sit near-constant at
15–25 kW regardless of workload, so they dilute the advantage: adding a flat 20 kW puts NVFP4 at **+58.0%** perf/watt over BF16 rather than +65%.

<details>
<summary>Experiment 3 configuration</summary>

| | |
| --- | --- |
| model | DeepSeek-V3 671B |
| hardware | GB300, 16 nodes × 4 = **64 GPUs** |
| parallelism | EP 32, PP 1, TP 1, CP 1 |
| batch | `local_batch_size` 8, GBS 512 sequences, seq_len 4096 |
| M | 32,768 tokens per expert invocation |
| steps | 1200 requested in a 3600 s limit — **every arm ends on TIMEOUT by design** |
| sampling | `nvidia-smi` at 1 Hz: `power.draw`, `power.limit`, `clocks.gr`, `pstate`, utilisation, temperature |
| analysis | leading 40% of samples dropped as compile/warm-up ramp |

**The sampler is free.** The same geometry had already been measured without any probe: BF16 2,602, MXFP8 3,153, NVFP4 3,608 tok/s/GPU. All three arms here come out **1.6–3.6% faster**, so 1 Hz sampling across 64 GPUs costs no measurable throughput and the power figures belong to an undisturbed workload.

</details>

---

## Conclusion

Experiments 2 and 3 measure **cost**; Experiment 1 measures **accuracy**. They point in opposite directions and should be quoted together:

| arm | energy per token | 16B loss vs BF16 |
| --- | ---: | ---: |
| MXFP8 | −23.5% | +0.0098 nats |
| NVFP4 | **−39.4%** | **+0.0503 nats** |

NVFP4 buys roughly 1.7× the energy saving of MXFP8 for roughly 5× the loss penalty. Two further limits worth stating: every margin here is **batch-size dependent** (NVFP4 is *slower* than MXFP8 at small M), and the 671B numbers are throughput and power only.

vkuzo:
> tl.randint already is tl.randint4x under the hood. It computes 4 random numbers via philox but only returns the first one. Swapping the call changes nothing by itself.

right, the code could use all 4 values instead of ignoring 3 of them

In any case, this is a moot point if cuteDSL is faster than triton for these, which I believe.  Let's just delete the triton kernels in a follow-up?

vkuzo:
> Three experiments against a BF16 and MXFP8 baseline. All runs use torchtitan and torchao with pinned revisions. For each experiment every arm shares one container, so the precisions are arm-locked. The only intended difference is the quantization converters. The NVFP4 quantization scheme is NVFP4 F0L0 FFN + MXFP8 ATTN.

Could you specify what "F0L0 FFN" is? Also, curious why experiment 1 used mxfp8 attn at all.  Was mxfp8 attn used only in the "nvfp4" branch, or also in the "mxfp8" branch?  That seems unrelated to nvfp4.

Any thoughts on why the loss curve for the nvfp4 branch seems to be trending in the wrong direction?

torchao/prototype/moe_training/nvfp4_training/nvfp4_linear.py
```diff
@@ -137,13 +128,17 @@ def _weight_quantize_2d(x: torch.Tensor, use_cutedsl: bool):
...
+
+    ``use_cutedsl`` selects the quantize backend; it reaches the RHT amax and
+    the 2D weight quantize. The name carried ``_triton`` while this was the
+    only path -- it dispatches both now, so do not read it as triton-only.
 
     3 GEMMs:
-      forward:   x_row @ W.T  = output         (triton RHT rowwise + 2D weight)
-      backward:  dy_sr @ W.T  = grad_input      (triton SR rowwise + 2D weight)
-      backward:  dy_col.T @ x_col = grad_weight (triton col RHT + SR for dy; saved col for x)
+      forward:   x_row @ W.T  = output          (RHT rowwise + 2D weight)
```
vkuzo: the RHT should be on both operands of `grad_weight`, see https://arxiv.org/pdf/2509.25149 figure 5.  I think the code should be already right, can we fix the comment since we already touching this code?

vkuzo:
@claude review, focus on whether the cutedsl path matches the triton path, on whether the new logic is well covered with tests, and any important edge cases you can think of

vkuzo:
thanks @rdspring1 ! Directionally this looks good, I had a few questions/comments inline and kicked off agent review

torchao/prototype/moe_training/nvfp4_training/nvfp4_linear.py
rdspring1: Updated to:
```
     3 GEMMs -- each operand as (orientation, transform, rounding, name):
       forward:   x_row @ W.T      = output      (rowwise x; rowwise 2D-quantized W) -- no RHT or SR
       saved:     x_col, Wt                      (colwise RHT x; colwise 2D-quantized W, no RHT)
       backward:  dy_row @ Wt.T    = grad_input  (rowwise SR dy; saved colwise W, no SR)
       backward:  dy_col.T @ x_col = grad_weight (colwise RHT SR dy; saved colwise RHT x, no SR)
```

rdspring1:
@vkuzo 

> Was mxfp8 attn used only in the "nvfp4" branch, or also in the "mxfp8" branch?

The MXFP8 arm is the TorchTitan MXFP8 config for DSV3. All FFN layers and Attention linears - wq_a, wq_b, wkv_a, wkv_b, wo. NVFP4 F0L0 + MXFP8 attention means NVFP4 recipe is applied to the FFN layers - dense, shared and routed experts. Everything else is the same as MXFP8 arm.

> Also, curious why experiment 1 used mxfp8 attn at all.

The attention linears are roughly 1/3 FLOPS of DSV3 with MoE FFN layers containing the remaining 2/3. If you only swap the FFN layers but keep attention BF16, you don't see any speed up over MXFP8 recipe. Experiment 1-3 use the same recipe to show benefit of NVFP4 over MXFP8. 

> Any thoughts on why the loss curve for the nvfp4 branch seems to be trending in the wrong direction?

- Loss gaps can appear in training loss. For real training, you usually measure downstream accuracy or steps to target validation loss.

**Experiment 4: DeepSeek-V3 671B on GB300 — MLPerf DSV3 Training task**

<details>

<summary> Task Description </summary>
The MLPerf DeepSeek-V3 training task is a large-scale MoE pretraining benchmark that measures how quickly a system can train the 671B-parameter DeepSeek-V3 model from a provided warm-start checkpoint to a target validation loss, rather than training the model from scratch. The workload uses C4 data with 4096-token sequences and is designed to represent steady-state training, stressing the full distributed training stack including MLA, expert parallelism, all-to-all communication, grouped expert GEMMs, optimizer execution, and overlap across TP/PP/CP/DP dimensions. Because the benchmark starts from a stabilized checkpoint with balanced expert routing, its primary goal is to compare end-to-end system performance and time-to-quality for production-scale DeepSeek-V3 training.

</details>

validation_curves_three_arms

* In this task, the models are trained until they cross held-out validation threshold of 3.6. NVFP4 crosses at step 45, MXFP8 at 43, BF16 at 44–45.

vkuzo:
> The MXFP8 arm is the TorchTitan MXFP8 config for DSV3

ah, I thought you were talking about mxfp8 SDPA, but you were talking about the attention linears, sgtm

vkuzo (approved):

rdspring1:
Thanks @vkuzo!

PR #4800, opened 2026-08-20
Add device-specific tensor registration pattern and XPU MX example

xiaowangintel:
### Motivation

torchao's tensor formats currently only support CUDA. Adding support for new devices (e.g., XPU) requires modifying core files like `mx_tensor.py` and `inference_workflow.py` with device-specific branches, which doesn't scale.

This PR introduces a generic device-specific tensor registration mechanism and uses it to add Intel XPU support. This change does not affect any of the existing functionality.

### Design
**Generic registration in `torchao/utils.py`** 

```python
# Register a device-specific tensor
register_ao_tensor(MXTensor, "xpu", MXTensorXPU)

# Look up at runtime (falls back to generic_type if no device-specific tensor registered)
mx_cls = get_ao_tensor(MXTensor, device_type)
```

Any tensor type inheriting from `TorchAOBaseTensor` can use this pattern to register device-specific variants without modifying core files.

**XPU implementation in `torchao/prototype/mx_formats/xpu/`**:

### DIFF
--- a/torchao/prototype/mx_formats/inference_workflow.py
+++ b/torchao/prototype/mx_formats/inference_workflow.py
@@ -31,7 +31,7 @@ from torchao.quantization.transform_module import (
     register_quantize_module_handler,
 )
 from torchao.quantization.utils import _module_extra_repr, _quantization_type
-from torchao.utils import is_sm_at_least_100
+from torchao.utils import get_ao_tensor, is_sm_at_least_100
 
 
 class NVFP4ObservedLinear(torch.nn.Linear):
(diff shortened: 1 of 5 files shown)


torchao/prototype/mx_formats/inference_workflow.py
vkuzo, line 129: IMO the config should encode 100% of the information needed to understand the serialized checkpoint.  Therefore, if different devices require different layouts, that information should be encoded in the workflow config and passed to tensor configs, not automatically inferred from the tensor devices.

PR #4808, opened 2026-08-20
Fix _is_128_128_scaled false-positive on PerTensor-scaled 128x128 tensors

### LINKED ISSUE #4089: Float8DynamicActivationFloat8WeightConfig with PerTensor granularity false-positives _is_128_128_scaled on 128x128 linear

amanyagami:
## Summary
`_is_128_128_scaled` (torchao/float8/inference.py) inferred "128x128 blockwise scaling" purely from `block_size == (128, 128)`. That's ambiguous: `get_block_size()` gives a `PerTensor`-granularity tensor a `block_size` equal to its full shape, so a weight whose shape happens to be exactly 128x128 also gets `block_size == (128, 128)`, even though it was never blockwise-scaled.

`Float8DynamicActivationFloat8WeightConfig()`'s dispatch (`quantize_/workflows/float8/float8_tensor.py`) uses `_is_128_128_scaled` to decide whether the activation tensor must be 1x128 scaled. For a 128x128 `PerTensor`-quantized linear this misfires and requires a plain per-tensor-quantized activation to be 1x128 scaled instead, breaking inference:

```python
from torchao.quantization.quant_api import Float8DynamicActivationFloat8WeightConfig, quantize_

model = ToyLinearModel()  # contains 128x128 linear
config = Float8DynamicActivationFloat8WeightConfig()  # default PerTensor granularity
quantize_(model, config)  # -> breaks test_workflow_e2e_numerics
```

One of the four call sites (the `KernelPreference.AUTO` mslk-selection check) already worked around this ambiguity with an extra `and not _is_tensorwise_scaled(weight_tensor)` check; the other three (including the one that actually causes the reported failure, at the `_is_128_128_scaled(weight_tensor)` branch that decides between the blockwise GEMM path and the normal `addmm_float8_unwrapped_inference` path) did not.

## Fix
Move that exclusion into `_is_128_128_scaled` itself, so every call site gets it consistently (the now-redundant explicit check at the one site that already had it was left as-is, to keep this diff minimal).

## Test
Added `test/float8/test_inference_granularity_helpers.py`, which exercises `_is_128_128_scaled`/`_is_tensorwise_scaled` directly against plain CPU tensors carrying a `block_size` attribute (the only thing these helpers actually touch — no float8 dtype or GPU compute is involved, so this runs on CPU-only CI):
- A 128x128 tensor with `block_size == (128, 128)` (matching its full shape, i.e. `PerTensor`) is no longer misidentified as 128x128-blockwise.
- A 256x256 tensor tiled into 128x128 blocks (`block_size == (128, 128)` but the tensor's full shape is `(256, 256)`) is still correctly identified as 128x128-blockwise.
- Verified the first assertion fails against the unpatched code (reproduces the reported bug) and passes with the fix.

Fixes #4089

### DIFF
--- a/test/float8/test_inference_granularity_helpers.py
+++ b/test/float8/test_inference_granularity_helpers.py
@@ -1,0 +1,67 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Regression test for https://github.com/pytorch/ao/issues/4089.
+
+`_is_128_128_scaled` used to infer "128x128 blockwise scaling" purely from
+`block_size == (128, 128)`. That's ambiguous: a `PerTensor`-granularity
+weight whose shape happens to be exactly 128x128 also gets `block_size ==
+(128, 128)` (one block spanning the whole tensor, from `get_block_size`),
+even though it was never blockwise-scaled. Downstream dispatch code
+(`torchao/quantization/quantize_/workflows/float8/float8_tensor.py`) uses
+`_is_128_128_scaled` to decide whether the *activation* tensor must be 1x128
+scaled, which then fails for a plain PerTensor-quantized activation - this is
+exactly the crash the issue reports for `ToyLinearModel`'s 128x128 linear.
+
+These tests exercise `_is_128_128_scaled`/`_is_tensorwise_scaled` directly
+with plain CPU tensors carrying a `block_size` attribute (the only thing
+these helpers actually touch - no float8 dtype or GPU compute is involved),
+so they run happily in CPU-only CI.
+"""
+
+import torch
+from torch.testing._internal.common_utils import TestCase, run_tests
+
+from torchao.float8.inference import _is_128_128_scaled, _is_tensorwise_scaled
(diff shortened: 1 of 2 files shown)


amanyagami:
@vkuzo similar area to #4791/#4792/#4793 — `_is_128_128_scaled` in `torchao/float8/inference.py` false-positives on a `PerTensor`-scaled 128x128 tensor, misclassifying it as blockwise-scaled. Small fix plus a test, CI green, still reproducing on `main`.

amanyagami:

PR #4809, opened 2026-08-20
Bounds-check block_size against packed tensors in Int4TilePackedTo4dTensor

### LINKED ISSUE #4572: Bounds-check `block_size` against `scale_and_zero` in the int4 tile-packed load path (avoids an OOB read in `_weight_int4pack_mm`)

amanyagami:
## Summary
`Int4TilePackedTo4dTensor` stored `block_size` with no validation against the actual packed `qdata`/`scale_and_zero` tensors. At inference, `groupsize = block_size[-1]` is handed straight to `torch.ops.aten._weight_int4pack_mm(act, qdata, groupsize, scale_and_zero)`, which trusts `groupsize` to be consistent with how many `(scale, zero)` groups `scale_and_zero` actually holds. The aten op validates that `groupsize ∈ {32,64,128,256}` but not that it's *consistent* with `scale_and_zero`'s actual group extent, so a checkpoint whose `block_size` metadata was edited to declare a smaller (but still individually valid) groupsize than the one `qdata`/`scale_and_zero` were actually packed with makes the tinygemm kernel iterate more groups than `scale_and_zero` holds — reading out of bounds.

This is unreachable via `from_hp` (which always derives `block_size` from the tensor it packs); the vulnerable path is deserializing a tampered/inconsistent checkpoint that skips `from_hp` — e.g. via `__tensor_unflatten__` or direct construction, as flagged in the issue for the safetensors path (`TorchAoHfQuantizer.is_serializable == True`, subclass flattened to safetensors with `block_size` in metadata).

## Fix
`qdata` is packed as `[..., n/8, k_padded/(inner_k_tiles*16), 32, inner_k_tiles/2]` (`inner_k_tiles` fixed to 8, see the class docstring), so the actual packed K extent is always recoverable from `qdata`'s shape regardless of what `block_size` claims. `scale_and_zero` holds one `(scale, zero)` pair per group along that same K extent. `__init__` now cross-checks `block_size`'s groupsize against both, rejecting an inconsistent checkpoint there instead of letting it reach the tinygemm kernel.

I verified this holds for every existing non-adversarial construction path (`from_hp`, `slice`, `select`, MoE 3D weights) using `device="meta"` tensors — all group-aligned slicing (the kind used throughout the existing test suite; `test_slice` explicitly notes it uses multiples of 1024 so padding doesn't affect the result) satisfies the new check. Only a misaligned slice boundary (not exercised anywhere in the repo) or an actually-tampered checkpoint trips it.

## Test
Added `test/quantization/quantize_/workflows/int4/test_int4_tile_packed_to_4d_tensor_bounds_check.py`. The real `_convert_weight_to_int4pack`/`_weight_int4pack_mm` ops only have CUDA kernels registered (confirmed — no CPU kernel exists, matching the existing `test_cant_initialize_in_cpu` test in the sibling test file), but the packed *shapes* they produce are fully determined by simple arithmetic and reproducible with `device="meta"` tensors. I verified the shape formula directly against the real op's `Meta` dispatch key (`test_packed_qdata_shape_matches_real_op`), then used it to construct `Int4TilePackedTo4dTensor` instances — covering 2D and MoE 3D weights, and the issue's exact tampered-checkpoint scenario (packed at `group_size=128`, `block_size` edited to claim `32`) — with no GPU involved.

Fixes #4572

### DIFF
--- a/test/quantization/quantize_/workflows/int4/test_int4_tile_packed_to_4d_tensor_bounds_check.py
+++ b/test/quantization/quantize_/workflows/int4/test_int4_tile_packed_to_4d_tensor_bounds_check.py
@@ -1,0 +1,134 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Regression test for https://github.com/pytorch/ao/issues/4572.
+
+`Int4TilePackedTo4dTensor` stored `block_size` with no validation against the
+actual packed `qdata`/`scale_and_zero` tensors. At inference,
+`groupsize = block_size[-1]` is handed straight to
+`torch.ops.aten._weight_int4pack_mm(act, qdata, groupsize, scale_and_zero)`,
+which trusts `groupsize` to be consistent with how many (scale, zero) groups
+`scale_and_zero` actually holds. A checkpoint whose `block_size` metadata was
+edited to declare a smaller (but still individually valid) groupsize than the
+one `qdata`/`scale_and_zero` were actually packed with makes the tinygemm
+kernel iterate more groups than `scale_and_zero` holds, reading out of bounds
+(this is unreachable via `from_hp`, which always derives `block_size` from
+the tensor it packs -- the vulnerable path is deserializing a tampered/
+inconsistent checkpoint that skips `from_hp`, e.g. via `__tensor_unflatten__`
+or direct construction).
+
+The real tinygemm ops (`_convert_weight_to_int4pack` / `_weight_int4pack_mm`)
+only have CUDA kernels registered, but the packed *shapes* they produce are
+fully determined by simple arithmetic and are reproducible with `device="meta"`
+tensors (verified below against the real op via the `Meta` dispatch key) --
+so this test constructs `Int4TilePackedTo4dTensor` instances directly with
+correctly- and incorrectly-shaped meta tensors, with no GPU involved.
+"""
+
+import torch
+from torch.testing._internal.common_utils import TestCase, run_tests
+
+from torchao.quantization.quantize_.workflows.int4.int4_tile_packed_to_4d_tensor import (
+    Int4TilePackedTo4dTensor,
+)
+
+# Fixed for this packing format, see the class docstring.
+_INNER_K_TILES = 8
+
+
+def _make_qdata_and_scale_and_zero(n, k, group_size, num_experts=None):
+    """Build correctly-shaped, empty meta qdata/scale_and_zero tensors for an
+    (n, k) weight (or a batch of `num_experts` of them), matching the shapes
+    `Int4TilePackedTo4dTensor.from_hp` actually produces (confirmed against
+    the real `_convert_weight_to_int4pack` op run on `device="meta"`, since
+    it only has CUDA kernels registered otherwise).
+    """
+    packed_n, packed_k_groups = n // 8, k // (_INNER_K_TILES * 16)
+    num_groups = k // group_size
+    if num_experts is None:
+        qdata_shape = (packed_n, packed_k_groups, 32, _INNER_K_TILES // 2)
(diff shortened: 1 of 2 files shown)


amanyagami:
@jerryzh168 @andrewor14 flagging this one as well — adds a bounds-check so an out-of-range `block_size` against a packed `Int4TilePackedTo4dTensor` raises instead of reading/writing out of bounds. 8 lines plus a test, CI green, still reproducing on `main`.

amanyagami:

PR #4717, opened 2026-08-07
Delete experimental custom Triton autotuner from torchao/kernel

vkuzo:
Summary:

The custom autotuner in torchao/kernel was an experimental, off-by-default
feature (gated behind TORCHAO_AUTOTUNER_ENABLE, which defaults to 0). It was
never enabled in any normal code path -- the only place that set the env var
was its own test. This deletes the autotuner and all code that becomes
orphaned once it's gone.

Deleted:
- torchao/kernel/autotuner.py (autotuning harness; only consumer was intmm_triton)
- torchao/kernel/intmm_triton.py (Triton int8 kernels + torchao::int_matmul /
  int_scaled_matmul custom ops, only reached when the autotuner was enabled)
- torchao/kernel/configs/data_a100.pkl and the now-empty configs/ dir
- torchao/kernel/README.md (documented only the autotuner env vars)
- benchmarks/intmm.py, benchmarks/print_config_shapes.py (imported the removed
  int_matmul / autotuner)

Edited:
- torchao/kernel/intmm.py: removed the intmm_triton import, AUTOTUNER_ENABLE,
  int_matmul(), both AUTOTUNER_ENABLE branches, and the now-unused os import
  and scales1.expand(). safe_int_mm, int_scaled_matmul, and the CPU AMX/VNNI
  path are unchanged.
- setup.py: dropped the "torchao.kernel.configs" package-data entry.
- test/kernel/test_autotuner.py -> test/kernel/test_intmm.py: removed the
  env-var setup and the int_matmul test; kept the safe_int_mm float8,
  int_scaled_matmul, and CPU-path tests.
- docs: removed the int_matmul bullet in quantization_overview.rst and
  repointed the custom-triton-kernel example in contributor_guide.rst to the
  surviving bsr_triton_ops.py (the old autotuner link pointed at /main/ and
  would 404).

No functional change for any non-autotuner path.

Test Plan:

pytest test/kernel/test_intmm.py
-> 9 passed

ruff check torchao/kernel/intmm.py test/kernel/test_intmm.py \
  torchao/prototype/quantized_training/int8_mm.py setup.py
-> All checks passed!

Verified imports still resolve:
  python -c "import torchao.quantization; import torchao.sparsity.blocksparse; \
    from torchao.kernel import intmm; assert not hasattr(intmm, 'int_matmul')"

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

PR #4810, opened 2026-08-20
Fix wrong variable in test_handle_save_load_meta_impl error message

### LINKED ISSUE #4339: Parametrized dtype overridden in test_autotuner; wrong variable in test_integration error message
## Bug 1: Parametrized `dtype` silently overridden in test_autotuner.py

All three test methods in `TestQuantFlow` are parametrized over `dtype` (both `torch.bfloat16` and `torch.float16`), but each one immediately overwrites it with a hardcoded `dtype = torch.bfloat16`:

| Method | Line | Override |
|---|---|---|
| `test_int_mm` | 40 | `dtype = torch.bfloat16` |
| `test_int_mm_float8` | 62 | `dtype = torch.bfloat16` |
| `test_int_scaled_mm` | 88 | `dtype = torch.bfloat16` |

This means the `float16` test cases never actually test float16. They silently run with bfloat16 instead, giving false test coverage.

Additionally, `test_int_scaled_mm` asserts `out32_1.dtype == torch.bfloat16` on line 96, which is hardcoded instead of using the parametrized `dtype`. If the override were removed without fixing this assertion, float16 tests would fail.

## Bug 2: Wrong variable in error message in test_integration.py

Line 641-642:

```python
assert SQNR(ref_f, test) > min_sqnr, (
    f"got sqnr: {SQNR(ref_f, ref_q)}, expected: {min_sqnr}"
)
```

The assertion checks `SQNR(ref_f, test)`, but the error message prints `SQNR(ref_f, ref_q)`. `ref_q` is the compiled quantized reference computed earlier, not the value being tested. When this assertion fails, the error message shows the wrong SQNR value — the one between the float reference and the compiled quantized reference, not between the float reference and the actual test output.

Should be:
```python
f"got sqnr: {SQNR(ref_f, test)}, expected: {min_sqnr}"
```

## Reproduction

Bug 1: Run `pytest test/kernel/test_autotuner.py -v` and observe that `test_int_mm_cuda_float16` and similar cases produce identical results to the bfloat16 variants.

Bug 2: No runtime failure — the bug only manifests in error messages when the assertion fails.

amanyagami:
## Summary
`TestSaveLoadMeta._test_handle_save_load_meta_impl` in `test/integration/test_integration.py` has two near-identical assertions comparing a float reference (`ref_f`) against a compiled-quantized output. The second one checks `SQNR(ref_f, test) > min_sqnr` (correct), but its failure message printed `SQNR(ref_f, ref_q)` instead — `ref_q` is the *other* assertion's compiled reference, computed earlier in the method, not the value the second assertion actually checked. On failure this would report the wrong number, from the wrong tensor pair, making the failure harder to debug.

## Fix
```diff
         assert SQNR(ref_f, test) > min_sqnr, (
-            f"got sqnr: {SQNR(ref_f, ref_q)}, expected: {min_sqnr}"
+            f"got sqnr: {SQNR(ref_f, test)}, expected: {min_sqnr}"
         )
```

## Test
The method itself needs `torch.compile(mode="max-autotune")` plus a real quantization API and isn't meaningfully runnable/verifiable without a GPU. Added `test/integration/test_save_load_meta_error_message.py`, which instead reads the method's source directly (it's wrapped by `@torch.no_grad()`/`@run_supported_device_dtype`, which defeats `inspect.getsource` on the function object) and checks that every `SQNR` assertion's failure message references the same variable its condition actually checked — i.e. that this exact copy-paste mistake can't silently reappear. Verified it fails against the unpatched source and passes with the fix.

This is bug 2 of #4339. Bug 1 (a parametrized `dtype` silently overridden in `test/kernel/test_autotuner.py`) no longer applies — that file was deleted along with `torchao/kernel/intmm_triton.py` in #4717.

Fixes #4339

### DIFF
--- a/test/integration/test_integration.py
+++ b/test/integration/test_integration.py
@@ -609,7 +609,7 @@ class TestSaveLoadMeta(unittest.TestCase):
         test = model_qc(x).detach()
 
         assert SQNR(ref_f, test) > min_sqnr, (
-            f"got sqnr: {SQNR(ref_f, ref_q)}, expected: {min_sqnr}"
+            f"got sqnr: {SQNR(ref_f, test)}, expected: {min_sqnr}"
         )
         self.assertTrue(torch.equal(ref_q, test))
 
--- a/test/integration/test_save_load_meta_error_message.py
+++ b/test/integration/test_save_load_meta_error_message.py
@@ -1,0 +1,63 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Regression test for https://github.com/pytorch/ao/issues/4339 (bug 2).
+
+`TestSaveLoadMeta._test_handle_save_load_meta_impl` in test_integration.py
+has two near-identical assertions comparing a float reference (`ref_f`)
+against a compiled-quantized output. The second one checked
+`SQNR(ref_f, test) > min_sqnr` (correct) but its error message printed
+`SQNR(ref_f, ref_q)` (`ref_q` is the *other* assertion's compiled reference,
+computed earlier in the method) instead of `SQNR(ref_f, test)` -- so on
+failure it reported the wrong number, from the wrong tensor pair.
+
+The method itself needs `torch.compile(mode="max-autotune")` plus a real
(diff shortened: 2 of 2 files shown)


amanyagami:
Small one — the second SQNR assertion in `test_handle_save_load_meta_impl` prints the wrong reference/output pair in its failure message, which makes a real failure harder to debug. One-line fix, CI green.

amanyagami:

PR #4817, opened 2026-08-24
Fix NVFP4/MX QAT backward dropping the bias gradient

shoemoney:
The custom autograd functions in `torchao/prototype/qat/nvfp4.py` and `torchao/prototype/qat/mx.py` accept `bias` in forward and pass it to the addmm dispatch, but their backward returns `None` in the bias gradient slot. After `loss.backward()`, `linear.bias.grad` is `None`, so the optimizer never updates any bias for the entire QAT run. There is no error; the model just trains with frozen biases and silently degraded accuracy.

The mainline sibling `FakeQuantizedLinear` (`torchao/quantization/qat/linear.py`, `F.linear(x, w, self.bias)`) trains bias normally through autograd, so only the NVFP4 and MX prototype paths are affected. The existing regression tests for these paths use bias=False models (`M`/`Sub` in `test_qat.py`), which is why they cannot catch this.

Fix: record bias presence on `ctx` in forward, and return the bias gradient in backward: `grad_output.sum(0)` for NVFP4 (grad_output is always 2D there, the module flattens 3D inputs before `apply`) and `grad_output_2d.sum(0)` for MX. Return `None` when bias is absent.

Tests: `M`/`Sub` gain an optional `bias` arg (default False, existing callers unchanged). `test_qat_nvfp4_training` is parametrized over `use_bias` and asserts bias gradients flow and are nonzero. `test_mx_fake_quantized_linear_backward` (already parametrized over bias) now asserts `bias.grad` exists and matches the reference `nn.Linear` bias gradient. Full `test_qat.py`: 141 passed, 48 skipped (CUDA-gated), no failures. Verified the gradient mechanism on CPU by calling the autograd functions directly: bias grad was `None` before and matches `grad_output.sum(0)` after; the new test parametrizations themselves are CUDA-gated like their neighbors.

### DIFF
--- a/torchao/prototype/qat/mx.py
+++ b/torchao/prototype/qat/mx.py
@@ -117,7 +117,6 @@ class _MXQuantizedForwardFakeQuantizedBackward(torch.autograd.Function):
 
         ctx.save_for_backward(_input_2d, weight)
         ctx.orig_shape = orig_shape
-        ctx.bias_present = bias is not None
 
         # Use addmm when bias is present, mm otherwise
         if bias is not None:
@@ -152,7 +151,7 @@ class _MXQuantizedForwardFakeQuantizedBackward(torch.autograd.Function):
         grad_weight = torch.mm(grad_output_2d.t(), _input_2d)
 
         grad_input = grad_input_2d.view(*orig_shape)
-        grad_bias = grad_output_2d.sum(0) if ctx.bias_present else None
+        grad_bias = grad_output_2d.sum(0) if ctx.needs_input_grad[2] else None
         return grad_input, grad_weight, grad_bias, None, None
 
 
--- a/torchao/prototype/qat/mx.py
+++ b/torchao/prototype/qat/mx.py
@@ -117,6 +117,7 @@ class _MXQuantizedForwardFakeQuantizedBackward(torch.autograd.Function):
 
         ctx.save_for_backward(_input_2d, weight)
         ctx.orig_shape = orig_shape
+        ctx.bias_present = bias is not None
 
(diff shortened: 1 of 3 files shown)


vkuzo:
@claude review

shoemoney:
The Check PR Labels failure wants a module: label, which I cannot add as an outside contributor. `module: qat` fits this change if a maintainer can apply it. The test-rocm nightly failure is unrelated to this diff (CPU-only gradient path change).

torchao/prototype/qat/mx.py
```diff
@@ -117,6 +117,7 @@ def forward(
 
         ctx.save_for_backward(_input_2d, weight)
         ctx.orig_shape = orig_shape
+        ctx.bias_present = bias is not None
```
vkuzo: instead of creating a new flag we should be able to use `ctx.needs_input_grad[2]`

vkuzo:
thanks for the fix! I commented with a simplification, after that is addressed this should be good to go

torchao/prototype/qat/mx.py
shoemoney: Done, both files now use ctx.needs_input_grad[2] and the flag is gone. Behavior verified unchanged on CPU (bias grad matches grad_output.sum(0) through real autograd for MX end to end, and via direct backward for NVFP4); test_qat.py still 141 passed. It also handles the requires_grad=False case the flag missed. Thanks for the label and the review.

vkuzo:
@claude review

andrewor14 (approved):

andrewor14:
Looks great, thanks for the fix!

PR #4820, opened 2026-08-24
[mxfp8 training] Add cudnn wrapper for fused grouped GEMM + SwiGLU + MXFP8 quantization ops for MoE routed experts

wolfcomos:
## Summary

This PR adds four torch custom ops that run a routed-experts (grouped) MLP entirely in MXFP8 with the GEMM↔activation boundaries fused, so the intermediate activations are quantized in-kernel and the BF16 hidden tensor is never materialized:

- `mxfp8_grouped_gemm_swiglu_fwd_cudnn` — FC1 grouped GEMM on MXFP8 operands + SwiGLU + dual MXFP8 RCEIL requantization (rowwise 1x32 and columnwise 32x1) in one launch; also emits the BF16 pre-activation `z` consumed unchanged by the backward op.
- `mxfp8_grouped_gemm_cudnn` — grouped GEMM on prequantized MXFP8 operands → BF16 (serves FC2 forward and FC1 dgrad).
- `mxfp8_grouped_gemm_dswiglu_bwd_cudnn` — FC2-dgrad grouped GEMM + dSwiGLU + dual MXFP8 requantization.
- `mxfp8_grouped_gemm_wgrad_cudnn` — ragged-K grouped weight-gradient GEMM (serves both FC1 and FC2 wgrads).

The device kernels are the grouped-GEMM wrapper family shipped by the `cudnn` python package (>= 1.27, `cudnn-frontend`); this module contributes the op registration, fakes, and layout plumbing. The package is imported lazily inside the op bodies — torchao gains **no hard import dependency**; on machines without it, the first op launch fails with the actionable `import cudnn` error. The consuming composite lives in torchtitan (pytorch/torchtitan#4257's override module); a follow-up torchtitan PR wires these ops in as a selectable fusion plan. This torchao PR is ops-only.

## Contracts

The kernels have hard alignment contracts, documented at the op boundary and guaranteed by the caller:

- **Per-expert row counts must be multiples of 256** (`ROW_GROUP_ALIGNMENT`, the kernels' fixed pad size). This is the one contract that fails *silently* if violated — sub-256 groups produce nondeterministic corruption, not an error — so the callers' padded token dispatcher is the load-bearing guarantee.
- Feature dims (`D`, `F`) multiples of 128 (`DIM_ALIGNMENT`); `R * max(dim) < 2**31` (int32 element indexing); 16-byte pointer alignment.
- Layouts: activations carry rowwise 1x32 blocked scales; columnwise operands carry **per-group** blocked scales (logical `[cols, rows_g/32]` per expert), sized by the ALLOCATED rows (`round_up(cols,128) * R/32` bytes — padding past `offsets[-1]` is dead bytes the kernels never read); columnwise qdata is accepted in both majors (no transpose copies on the hot path); `offsets` is an int32 `[G]` exclusive-end cumsum. Rows past `offsets[-1]` of kernel-allocated outputs are garbage and read-forbidden.

The ops perform no input validation: the composite caller constructs every operand from its own quantizers and dispatcher, and out-of-contract metadata fails loudly at the cudnn plan-building layer.

All four ops are **bitwise deterministic** across runs, including with the kernels' dynamic scheduling enabled (byte-verified on GB200).

## Performance

End-to-end in torchtitan (DeepSeek-V3 16B, 4x GB200 at 2062 MHz application clocks, EP=4, TP=1, seq 4096, MXFP8 MoE recipe, 50 steps, interleaved A/B rounds, pooled medians of steps 11-50). Arms: the stock MXFP8 path (pad 128), the same path pad-matched to these kernels' 256-row groups, and the fused composite backed by this PR's ops:

```
arm                        bs4 tokens/s   bs8 tokens/s
-------------------------  -------------  -------------
stock MXFP8 (pad 128)      14,289         18,706
stock MXFP8 (pad 256)      13,258         17,882
fused (this PR's ops)      13,563 (+2.3%) 18,783 (+5.0%)
```

The fused gain vs the pad-matched baseline grows with batch (fixed pad tax dilutes; fusion win scales with expert work); at bs8 the fused arm is already net-positive against the pad-128 stock path. Loss trajectories passed paired-seed parity gates (26-step and 50-step budgets with sign-mixing criteria) and a 3-seed x 500-step convergence protocol.

## API

```python
from torchao.prototype.moe_training.kernels.mxfp8.cudnn_grouped_mlp import (
    mxfp8_grouped_gemm_swiglu_fwd_cudnn,   # (x_q, x_sf, w13_q, w13_sf, offsets)
    mxfp8_grouped_gemm_cudnn,              # (a_q, a_sf, b_q, b_sf, offsets)
    mxfp8_grouped_gemm_dswiglu_bwd_cudnn,  # (dy_q, dy_sf, w2_col_q, w2_col_sf, z_bf16, offsets)
    mxfp8_grouped_gemm_wgrad_cudnn,        # (dy_col_q, dy_col_sf, x_col_q, x_col_sf, offsets)
    is_supported,                    # (model_dim, hidden_dim) -> bool
    DIM_ALIGNMENT, ROW_GROUP_ALIGNMENT, SCALE_BLOCK_SIZE,
)
```

`w13_q` is `[G, 2F, D]` in the kernel's 32-block interleaved gate/up order (the layout a fused w13 projection pack produces). Registered as torch custom ops with fake impls; zero-row inputs (`R == 0`) return empty outputs without touching the backend.

## Numerics and tests

16 tests, ~23 s on GB200 (skipped cleanly, no collection errors, without SM100 or the `cudnn` package — CI-safe):

- **Per-op PyTorch references**: each of the four ops has a standalone reference built from nothing but dequantization, per-expert FP32 `torch` matmuls, and `F.silu`/`torch.sigmoid`, and a per-op test validating the op output against it at every tested shape (GEMM-exact outputs measure 105–inf dB vs a 50 dB gate; requantized outputs ~31.5 dB vs a 27 dB gate). Shape cases cover a zero-token expert group, D≠F, G=1, and the 16B production dims.

## Reproducing

```bash
pytest test/prototype/moe_training/test_mxfp8_grouped_mlp.py -q
```

## Follow-ups

1. The torchtitan fusion-plan wiring PR (stacked on pytorch/torchtitan#4257) that consumes these ops end-to-end.
2. A fixed-capacity (static-shape) dispatcher mode would make the %256 contract satisfiable without per-step host-synced offsets, enabling CUDA-graph capture; today the padded dynamic dispatcher is the validated seam.


vkuzo:
this is interesting!  Will take a further look, but I have two initial comments:

1. Can we change torchao/prototype/moe_training/kernels/mxfp8/cutedsl_grouped_mlp.py to something like cudnn_grouped_mlp.py, and make the op names state clearly that this is calling cudnn?

2. Also, for the tests, can we make sure that each op has a plain pytorch reference, and we are validating that the numerics match the reference?

wolfcomos:
> this is interesting! Will take a further look, but I have two initial comments:
> 
> 1. Can we change torchao/prototype/moe_training/kernels/mxfp8/cutedsl_grouped_mlp.py to something like cudnn_grouped_mlp.py, and make the op names state clearly that this is calling cudnn?
> 2. Also, for the tests, can we make sure that each op has a plain pytorch reference, and we are validating that the numerics match the reference?

Thanks for the reviews! Just got those done: renamed the file and refactored the test for just pytorch reference and dequantize to compare SQNR for numerics.

vkuzo:
@claude review

vkuzo:
@wolfcomos thank you! directionally this looks good.  Can we wait until there is a clear path on https://github.com/pytorch/torchtitan/pull/4257 and land them close together, just to make it easier to do any relevant changes that are uncovered whenever we figure out the final torchtitan integration path?

wolfcomos:
> @wolfcomos thank you! directionally this looks good. Can we wait until there is a clear path on [pytorch/torchtitan#4257](https://github.com/pytorch/torchtitan/pull/4257) and land them close together, just to make it easier to do any relevant changes that are uncovered whenever we figure out the final torchtitan integration path?

Yes, thanks a lot for reviews! Sounds totally good to me, will wait until mxfp8 torchtitan interface cleared. Thanks!

PR #4824, opened 2026-08-24
[pt2e][qat] Fix qparams mismatch in FusedMovingAvgObsFakeQuantize

andrewor14:
**Summary:** In PT2E QAT, `FusedMovingAvgObsFakeQuantize` offers
two different qparams calculations with slightly differing numerics:

(1) `torch.fused_moving_avg_obs_fake_quant`

```
# aten/src/ATen/native/quantized/cuda/FusedObsFakeQuant.cu
int symmetric_qmin = -((qmax - qmin) / 2 + 1);
int symmetric_qmax = (qmax - qmin) / 2;
double max_scale = std::max(
    fabs(min_val / symmetric_qmin), fabs(max_val / symmetric_qmax));
min_val = max_scale * symmetric_qmin;
max_val = max_scale * symmetric_qmax;
min_val = std::min(min_val, 0.f);
max_val = std::max(max_val, 0.f);
scale[i] = (static_cast<double>(max_val) - min_val) / (qmax - qmin);
```

(2) `MovingAverageMinMaxObserver.calculate_qparams`

```
max_val_pos = torch.max(-min_val_neg, max_val_pos)
scale = max_val_pos / (float(quant_max - quant_min) / 2)
scale = torch.max(scale, self.eps)
```

For [-8, 7], (1) divides by 8 or 7 while (2) divides by 7.5,
so e.g. min=-1.149422, max=2.533506 gives 0.361929 vs 0.337801.

Although only (1) is ever called during training, (2) is
actually called during `convert_pt2e` after training when
swapping the FakeQuantize classes to qdq ops, resulting in
slight numerical divergences.

This commit fixes this by adding a flag `use_kernel_params`
that allows convert to read the qparams from (1) the qparam
buffers set by the fused kernel. This resolves the duplicate
observer semantics in `FusedMovingAvgObsFakeQuantize`, which
was the root cause of the numerical divergence.

Note: this flag defaults to False and is opt-in only when
instantiating your own quantizer. Thus, existing behavior
is unchanged. An example of how to set this flag through
your quantizer can be found in
`test_qparams_use_kernel_qparams_through_quantizer`.

**Test Plan:**

```
pytest test/quantization/pt2e/test_quantize_pt2e_qat.py -k TestFusedMovingAvgObsFakeQuantizeQParams
```

### DIFF
--- a/test/quantization/pt2e/test_quantize_pt2e_qat.py
+++ b/test/quantization/pt2e/test_quantize_pt2e_qat.py
@@ -6,6 +6,7 @@
 
 # Owner(s): ["oncall: quantization"]
 import copy
+import dataclasses
 import operator
 import unittest
 from typing import Any, Optional, Tuple
@@ -60,6 +61,15 @@ class PT2EQATTestCase(QuantizationTestCase):
     """
     Base QuantizationTestCase for PT2E QAT with some helper methods.
     """
+
+    # Model weights are drawn from the ambient RNG when the test builds its
+    # model, so without seeding here the weights depend on which tests ran
+    # before, and numerics comparisons pass or fail depending on test order.
+    SEED = 0
+
+    def setUp(self):
+        super().setUp()
+        torch.manual_seed(self.SEED)
 
     class _BaseConvBnModel(torch.nn.Module):
         def __init__(
@@ -436,6 +446,7 @@ class TestQuantizePT2EQAT_ConvBn_Base(PT2EQATTestCase):
     #   'QuantizationConfig' object has no attribute '__bool__'
 
     def setUp(self):
+        super().setUp()
         # NB: Skip the test if this is a base class, this is to handle the test
         # discovery logic in buck which finds and runs all tests here including
         # the base class which we don't want to run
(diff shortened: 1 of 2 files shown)


andrewor14:
@claude review

andrewor14:
@claude Regarding (1) and (2), I don't think your claim is correct. We don't actually update `self.scale` and `self.zero_point` if fake quant is disabled. This is where we set it, and it's guarded by `fake_quant_on`: https://github.com/pytorch/pytorch/blob/a44339dbbb3c4b1e298961e53b31870d03769dd6/aten/src/ATen/native/quantized/cuda/FusedObsFakeQuant.cu#L56

Here's a quick repro to prove that:

```
import torch
from torchao.quantization.pt2e import FusedMovingAvgObsFakeQuantize, MovingAverageMinMaxObserver

# observer = on, fake_quant = off -> self.scale is not updated
fq = FusedMovingAvgObsFakeQuantize(
    observer=MovingAverageMinMaxObserver, quant_min=-8, quant_max=7,
    dtype=torch.qint8, qscheme=torch.per_tensor_symmetric)
fq.enable_observer()
fq.disable_fake_quant()
fq(torch.tensor([-1.149422, 2.533506]))
print(fq.scale)  # tensor([1.])  <- scale still at its init value
```

So I think the docstring is correct as written. Please take another look?

andrewor14:
@claude please review again, I did not gate on `fake_quant_enabled` on purpose since the user may turn off fake quant before converting. In general we shouldn't give different numerics depending on whether the user has fake quant enabled or not. The downside is potentially stale qparams if the user observes without fake quantizing for a while, but I think this is a less common use case.

vkuzo (approved):

andrewor14:
@claude review again after the latest changes. I intentionally avoided gating on `fake_quant_enabled` during `calculate_qparams`, since users may deliberately disable fake quant after training (and there are real use cases that do this). I made the new behavior opt-in instead to minimize impact on existing users.

PR #4839, opened 2026-08-27
Make MXTensor subclass-friendly

ahmadsarvmeily:
**Summary**
This PR makes `MXTensor` subclass-friendly and modifies hard-coded internal construction paths that would otherwise force the base `MXTensor` type.

It does this in two layers:

1. Make `MXTensor` itself override-friendly.
2. Add explicit MX factory functions and migrate internal TorchAO call sites to use them.

The idea is, by calling `set_default_mx_tensor_class(MyMXTensor)`, the internals that call the new factory functions `to_mx_tensor(...)` and `make_mx_tensor(...)` instead of the explicit `MXTensor.to_mx` and `MXTensor(...)` will automatically use `MyMXTensor`.

It is insufficient to simply rely on changing the Config. For example, if we take `MXDynamicActivationMXWeightConfig`, we can specify the following info: `block_size, activation_dtype, weight_dtype, KernelPreference, ScaleCalculationMode`. However, the quantization algorithm itself in `MXTensor.to_mx` cannot be modified if needed. In this case, more flexibility is required than what can be specified in any Config.

Note that the behaviour of `MXTensor.to_mx` and `MXTensor(...)` remain untouched: If used explicitly, these will _always_ produce `MXTensor` with no implicit behaviour. Only the new factory functions will try to resolve which default `MXTensor` class to use.

### DIFF
--- a/test/prototype/mx_formats/test_mx_tensor.py
+++ b/test/prototype/mx_formats/test_mx_tensor.py
@@ -29,6 +29,10 @@ from torchao.prototype.mx_formats.mx_tensor import (
     _to_mx_rceil,
     to_dtype,
     to_mx,
+    get_default_mx_tensor_class,
+    make_mx_tensor,
+    set_default_mx_tensor_class,
+    to_mx_tensor,
 )
 from torchao.prototype.mx_formats.utils import from_blocked, to_blocked
 from torchao.quantization.quantize_.common import KernelPreference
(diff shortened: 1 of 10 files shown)


vkuzo:
@ahmadsarvmeily , curious on what you are trying to achieve?

ahmadsarvmeily:
> @ahmadsarvmeily , curious on what you are trying to achieve?

@vkuzo Hi, I want to be able to use a subclass of `MXTensor` (e.g. `MyMXTensor`) automatically in paths that currently hardcode `MXTensor` specifically. For example, the main `quantize_` API with MX configs. This includes being able to use other libraries that use TorchAO, such as TorchTitan, without modifying that code.

I want to be able to customise the behaviour of `to_mx` and other MX functions e.g. `mx_mm`. The latter is already overridable via `@MyMXTensor.implements` so that works. But `to_mx` cannot be easily overridden without monkey-patching the function, which is unstable and undesirable.

vkuzo:
@ahmadsarvmeily , thanks for explaining.  I'd actually recommend monkey patching, or figuring out how to make the config more flexible.  I think adding a factory here is a significant jump in system complexity, which is not desireable.

ahmadsarvmeily:
@vkuzo To give a bit more context, this is to enable support for our own device, and we would need to provide our own `to_mx` kernel. It would not be possible simply by making the config more flexible. I appreciate that the proposed solution may not be ideal, but monkey-patching would not be a long-term solution for us, if it's at all avoidable.

Another potential solution could involve stripping the implementation detail from `MXTensor.to_mx` and similar functions so that each device/backend can register its own implementations, including the default GPU implementation. This would bring MXTensor closer to being a public interface. Would this be better?

vkuzo:
> To give a bit more context, this is to enable support for our own device, and we would need to provide our own to_mx kernel. It would not be possible simply by making the config more flexible. I appreciate that the proposed solution may not be ideal, but monkey-patching would not be a long-term solution for us, if it's at all avoidable.

Are you trying to do this in-tree or out-of-tree?  Is this for a backend with publically known requirements?  We definitely want to be friendly to multiple backends (and we support NVIDIA, rocm, xpu today), and the design should enable per-backend logic.  I do want to keep it simple if possible.

Note: we are working on adding a `quantize_tensor` function to PyTorch core, so `to_mx` will cease to exist in the future.

ahmadsarvmeily:
@vkuzo Out-of-tree. and the backend requirements are not publicly known, no. Is there somewhere I can read more about the new `quantize_tensor` and how it differs, and do you know when it can be expected? Will it be easier to override with custom behaviour?

vkuzo:
it's here: https://github.com/pytorch/pytorch/issues/195481, and https://docs.pytorch.org/docs/2.13/notes/extending.html#extending-torch-python-api has ideas on how to extend it.  For example, you could write a `__torch_function__` or a `__torch_dispatch__` override for `quantize_tensor` that has custom behavior for your device type.

ahmadsarvmeily:
@vkuzo Thank you for sharing this. Just to test my understanding, does this mean that the `MXTensor` class will disappear as well as `to_mx`? I noticed it doesn't appear in the RFC you mentioned. What would happen to the existing MX operation implementations that are decorated with `@MXTensor.implements` such as `mx_mm` in `mx_tensor.py`? Because currently those also need to be overridden

vkuzo:
`MXTensor` is a wrapper around various ops.  The ops are going into PyTorch core.  `MXTensor` the glue code will stay in torchao.

ahmadsarvmeily:
@vkuzo So in theory, after the implementation of `quantize_tensor`, I simply need to:
1. override this function using the "extending torch python API" link you sent
2. make this function return my out-of-tree `MXTensor` subclass (e.g.  `MyMXTensor`)
3. provide my implementations (out of tree) using `@MyMXTensor.implements` as before

and then it should automatically use it in the main API paths such as `_quantize`?

For example, in the `_quantize` path and other key paths, there are still usages of the constructor `MXTensor(...)` (these are the ones I replaced with a `make_mx_tensor` factory function in this PR) and these wouldn't automatically use a custom class even with `quantize_tensor` replacing `to_mx`, unless the explicit MXTensor constructors disappear as well. Will this be the case?

vkuzo:
my comment was just about overriding `F.quantize_tensor`, when it lands.  It was not related to `MXTensor`.  PyTorch core ops have a very fleshed out extensibility story, you'll be able to use it for free once the quantization ops are in core.

`MXTensor` glue code aims to stay as simple as possible, there is no plan to support out of tree non-public backends for it at the moment.

ahmadsarvmeily:
@vkuzo Ok, understood, thanks for clarifying.

If I reduce the scope of these PR changes to only include point 1:

> Make MXTensor itself override-friendly.

and remove this part:

> Add explicit MX factory functions and migrate internal TorchAO call sites to use them.

Would that be acceptable? So basically the only thing that changes is that certain `@implements` decorated functions such as `mx_linear` use `type(tensor)` instead of `MXTensor`. That way, our out-of-tree subclass can cut down on code duplication, and functionally things will be identical

vkuzo:
I think that could potentially make sense if there was an OSS project with meaningful usage that would benefit from it, or a significant number of private projects with meaningful usage. I am hesitant to do this for a single private project, as it would introduce complexity for the entire system.  I'd encourage to just duplicate what you need instead.

ahmadsarvmeily:
Ok, no worries

PR #4844, opened 2026-08-28
Refactor MXTensor: align the swizzle execution with the configured swizzle setting

xiaowangintel:
## Summary

This PR refactors `MXTensor`'s scaled-mm dispatch so it actually follows the configured `swizzled_type`. Previous designs always produced swizzled scales regardless of the config. This made `swizzled_type=SwizzleType.NO_SWIZZLE` silently ignored. But, the config itself is meant to be device-agnostic. PyTorch already models this correctly: F.scaled_mm takes the scale layout as explicit swizzle_a / swizzle_b arguments. This PR aligns ao with that contract by ensuring the configured layouts are honored consistently; in particular, swizzle_a and swizzle_b remain consistent, and NO_SWIZZLE skips both to_block and maybe_dtensor_to_blocked, allowing ao to inherit PyTorch’s backend support matrix instead of re-encoding backend-specific behavior.

## Align with PyTorch's API
**What each backend accepts of PyTorch**

| | CUDA | ROCm | XPU |
|---|---|---|---|
| `A` (activation) | `[M, K]` | `[M, K]` | `[M, K]` |
| `B` (weight) | `[K, N]`  | same | same |
| `scale` dtype | `float8_e8m0fnu` | same | same |
| `scale` logical shape | to_blocked | `scale_a: [M, K/32]`, `scale_b: [N, K/32]`  | `scale_a: [M, K/32]`, `scale_b: [N, K/32]`  |
| **`swizzle_a` / `swizzle_b`** | **`SWIZZLE_32_4_4`** | **`NO_SWIZZLE`** | **`NO_SWIZZLE`** |

## Usage

```python
# default — blocked/swizzled scales
config = MXDynamicActivationMXWeightConfig(
    swizzled_type = SwizzleType.SWIZZLE_32_4_4,
)

# no swizzle
config = MXDynamicActivationMXWeightConfig(
    swizzled_type = SwizzleType.NO_SWIZZLE,
)
```

### DIFF
--- a/test/prototype/mx_formats/test_mx_mm.py
+++ b/test/prototype/mx_formats/test_mx_mm.py
@@ -13,7 +13,7 @@ from torch.nn.functional import ScalingType, SwizzleType
 from torchao.float8.float8_utils import compute_error
 from torchao.prototype.mx_formats.mx_tensor import MXTensor
 from torchao.prototype.mx_formats.utils import to_blocked
-from torchao.utils import is_sm_at_least_100
+from torchao.utils import is_ROCM, is_sm_at_least_100
 
 
 def _mxfp4_scaled_mm(
@@ -68,7 +68,11 @@ def _mxfp8_scaled_mm(
 
 def run_matrix_test(M: int, K: int, N: int, format, device="cuda") -> float:
     dtype = torch.bfloat16
-    swizzle = SwizzleType.NO_SWIZZLE if device == "xpu" else SwizzleType.SWIZZLE_32_4_4
+    swizzle = (
+        SwizzleType.SWIZZLE_32_4_4
+        if device == "cuda" and not is_ROCM()
+        else SwizzleType.NO_SWIZZLE
+    )
 
     a = torch.rand((M, K), dtype=dtype, device=device)
     b = torch.rand((N, K), dtype=dtype, device=device)
--- a/test/prototype/mx_formats/test_mx_mm.py
+++ b/test/prototype/mx_formats/test_mx_mm.py
@@ -22,6 +22,7 @@ def _mxfp4_scaled_mm(
     """Wrapper for F.scaled_mm with MXFP4 configuration."""
     match swizzle:
         case SwizzleType.NO_SWIZZLE:
+            assert a_scale.is_contiguous()
             a_scale_block = a_scale
             b_scale_block = b_scale
         case SwizzleType.SWIZZLE_32_4_4:
@@ -48,6 +49,7 @@ def _mxfp8_scaled_mm(
     """Wrapper for torch._scaled_mm with MXFP8 configuration."""
     match swizzle:
         case SwizzleType.NO_SWIZZLE:
+            assert a_scale.is_contiguous()
             a_scale_block = a_scale
             b_scale_block = b_scale.t()
         case SwizzleType.SWIZZLE_32_4_4:
--- a/test/prototype/mx_formats/test_mx_mm.py
+++ b/test/prototype/mx_formats/test_mx_mm.py
@@ -16,8 +16,19 @@ from torchao.prototype.mx_formats.utils import to_blocked
 from torchao.utils import is_sm_at_least_100
 
 
-def _mxfp4_scaled_mm(a_data, b_data, a_scale_block, b_scale_block):
+def _mxfp4_scaled_mm(
+    a_data, b_data, a_scale, b_scale, swizzle=SwizzleType.SWIZZLE_32_4_4
+):
     """Wrapper for F.scaled_mm with MXFP4 configuration."""
+    match swizzle:
+        case SwizzleType.NO_SWIZZLE:
+            a_scale_block = a_scale.contiguous()
+            b_scale_block = b_scale.contiguous()
+        case SwizzleType.SWIZZLE_32_4_4:
+            a_scale_block = to_blocked(a_scale)
+            b_scale_block = to_blocked(b_scale)
+        case _:
+            raise ValueError(f"Unsupported swizzle type: {swizzle}")
     return F.scaled_mm(
         a_data.view(torch.float4_e2m1fn_x2),
         b_data.view(torch.float4_e2m1fn_x2),
@@ -25,24 +36,47 @@ def _mxfp4_scaled_mm(a_data, b_data, a_scale_block, b_scale_block):
         scale_recipe_a=ScalingType.BlockWise1x32,
         scale_b=b_scale_block,
         scale_recipe_b=ScalingType.BlockWise1x32,
-        swizzle_a=SwizzleType.SWIZZLE_32_4_4,
-        swizzle_b=SwizzleType.SWIZZLE_32_4_4,
+        swizzle_a=swizzle,
+        swizzle_b=swizzle,
         output_dtype=torch.bfloat16,
     )
 
 
-def run_matrix_test(M: int, K: int, N: int, format) -> float:
+def _mxfp8_scaled_mm(
+    a_data, b_data, a_scale, b_scale, swizzle=SwizzleType.SWIZZLE_32_4_4
+):
+    """Wrapper for torch._scaled_mm with MXFP8 configuration."""
+    match swizzle:
+        case SwizzleType.NO_SWIZZLE:
+            a_scale_block = a_scale.contiguous()
+            b_scale_block = b_scale.t().contiguous()
+        case SwizzleType.SWIZZLE_32_4_4:
+            a_scale_block = to_blocked(a_scale)
+            b_scale_block = to_blocked(b_scale)
+        case _:
+            raise ValueError(f"Unsupported swizzle type: {swizzle}")
+    return torch._scaled_mm(
+        a_data,
+        b_data,
+        a_scale_block,
+        b_scale_block,
+        out_dtype=torch.bfloat16,
+    )
+
+
+def run_matrix_test(M: int, K: int, N: int, format, device="cuda") -> float:
     dtype = torch.bfloat16
-    device = torch.device("cuda")
+    swizzle = SwizzleType.NO_SWIZZLE if device == "xpu" else SwizzleType.SWIZZLE_32_4_4
 
     a = torch.rand((M, K), dtype=dtype, device=device)
     b = torch.rand((N, K), dtype=dtype, device=device)
 
     fmt = torch.float8_e4m3fn if format == "fp8" else torch.float4_e2m1fn_x2
+
     mx_func = (
-        partial(torch._scaled_mm, out_dtype=torch.bfloat16)
+        partial(_mxfp8_scaled_mm, swizzle=swizzle)
         if format == "fp8"
-        else _mxfp4_scaled_mm
+        else partial(_mxfp4_scaled_mm, swizzle=swizzle)
     )
 
     a_mx = MXTensor.to_mx(a, fmt, 32)
@@ -56,20 +90,22 @@ def run_matrix_test(M: int, K: int, N: int, format) -> float:
     a_scale = a_mx.scale.view(M, K // 32)
     b_scale = b_mx.scale.view(N, K // 32)
 
-    a_scale_block = to_blocked(a_scale)
-    b_scale_block = to_blocked(b_scale)
-
     out_hp = a_mx.dequantize(torch.bfloat16) @ b_mx.dequantize(
         torch.bfloat16
     ).transpose(-1, -2)
-    out = mx_func(a_data, b_data, a_scale_block, b_scale_block)
+
+    out = mx_func(a_data, b_data, a_scale, b_scale)
 
     return compute_error(out_hp, out).item()
 
 
-@pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA not available")
 @pytest.mark.skipif(
-    not is_sm_at_least_100(), reason="CUDA capability >= 10.0 required for mxfloat8"
+    not torch.accelerator.is_available(),
+    reason="CUDA or XPU not available",
+)
+@pytest.mark.skipif(
+    torch.cuda.is_available() and not is_sm_at_least_100(),
+    reason="CUDA capability >= 10.0 required for mxfloat8",
 )
 @pytest.mark.parametrize(
     "size",
@@ -90,8 +126,9 @@ def run_matrix_test(M: int, K: int, N: int, format) -> float:
 )
 @pytest.mark.parametrize("format", ["fp8", "fp4"])
 def test_matrix_multiplication(size, format):
+    device = torch.accelerator.current_accelerator().type
     M, K, N = size
-    sqnr = run_matrix_test(M, K, N, format)
+    sqnr = run_matrix_test(M, K, N, format, device=device)
     threshold = 80.0
     assert sqnr >= threshold, (
         f"{format} SQNR {sqnr} below threshold for dims {M}x{K}x{N}"
--- a/test/prototype/mx_formats/test_mx_mm.py
+++ b/test/prototype/mx_formats/test_mx_mm.py
@@ -22,8 +22,8 @@ def _mxfp4_scaled_mm(
     """Wrapper for F.scaled_mm with MXFP4 configuration."""
     match swizzle:
         case SwizzleType.NO_SWIZZLE:
-            a_scale_block = a_scale.contiguous()
-            b_scale_block = b_scale.contiguous()
+            a_scale_block = a_scale
+            b_scale_block = b_scale
         case SwizzleType.SWIZZLE_32_4_4:
             a_scale_block = to_blocked(a_scale)
             b_scale_block = to_blocked(b_scale)
@@ -48,8 +48,8 @@ def _mxfp8_scaled_mm(
     """Wrapper for torch._scaled_mm with MXFP8 configuration."""
     match swizzle:
         case SwizzleType.NO_SWIZZLE:
-            a_scale_block = a_scale.contiguous()
-            b_scale_block = b_scale.t().contiguous()
+            a_scale_block = a_scale
+            b_scale_block = b_scale.t()
         case SwizzleType.SWIZZLE_32_4_4:
             a_scale_block = to_blocked(a_scale)
             b_scale_block = to_blocked(b_scale)
@@ -84,6 +84,7 @@ def run_matrix_test(M: int, K: int, N: int, format, device="cuda") -> float:
 
     a_data = a_mx.qdata
     b_data = b_mx.qdata
+    assert a_data.is_contiguous()
     assert b_data.is_contiguous()
     b_data = b_data.transpose(-1, -2)
 
--- a/torchao/prototype/mx_formats/inference_workflow.py
+++ b/torchao/prototype/mx_formats/inference_workflow.py
@@ -111,8 +111,8 @@ class MXDynamicActivationMXWeightConfig(AOBaseConfig):
     # How to calculate the block scales
     scaling_mode: ScaleCalculationMode = ScaleCalculationMode.RCEIL
 
-    # Whether to store block scales in swizzled (blocked) layout.
-    # CUDA uses True (blocked layout), XPU uses False (row-major).
+    # How to store block scales.
+    # CUDA uses swizzle layout, XPU uses not swizzle.
     is_swizzled_scales: bool = True
 
     def __post_init__(self):
--- a/torchao/prototype/mx_formats/inference_workflow.py
+++ b/torchao/prototype/mx_formats/inference_workflow.py
@@ -114,7 +114,7 @@ class MXDynamicActivationMXWeightConfig(AOBaseConfig):
 
     # How to store block scales.
     # CUDA uses swizzle layout, XPU uses not swizzle.
-    swizzled_type: bool = SwizzleType.SWIZZLE_32_4_4
+    swizzled_type: SwizzleType = SwizzleType.SWIZZLE_32_4_4
 
     def __post_init__(self):
         assert self.activation_dtype == self.weight_dtype, (
@@ -144,9 +144,7 @@ def _mx_inference_linear_transform(
         elem_dtype=config.activation_dtype,
         block_size=config.block_size,
         kernel_preference=config.kernel_preference,
-        is_swizzled_scales=True
-        if config.swizzled_type == SwizzleType.SWIZZLE_32_4_4
-        else False,
+        is_swizzled_scales=config.swizzled_type == SwizzleType.SWIZZLE_32_4_4,
         scaling_mode=config.scaling_mode,
     )
 
@@ -157,9 +155,7 @@ def _mx_inference_linear_transform(
         block_size=config.block_size,
         kernel_preference=config.kernel_preference,
         act_quant_kwargs=act_quant_kwargs,
-        is_swizzled_scales=True
-        if config.swizzled_type == SwizzleType.SWIZZLE_32_4_4
-        else False,
+        is_swizzled_scales=config.swizzled_type == SwizzleType.SWIZZLE_32_4_4,
         scaling_mode=config.scaling_mode,
     )
 
--- a/torchao/prototype/mx_formats/inference_workflow.py
+++ b/torchao/prototype/mx_formats/inference_workflow.py
@@ -12,6 +12,7 @@ from typing import Optional
 import torch
 import torch.nn.functional as F
 from torch import Tensor
+from torch.nn.functional import SwizzleType
 
 from torchao.core.config import AOBaseConfig
 from torchao.prototype.mx_formats.config import _validate_elem_dtype
@@ -113,7 +114,7 @@ class MXDynamicActivationMXWeightConfig(AOBaseConfig):
 
     # How to store block scales.
     # CUDA uses swizzle layout, XPU uses not swizzle.
-    is_swizzled_scales: bool = True
+    swizzled_type: bool = SwizzleType.SWIZZLE_32_4_4
 
     def __post_init__(self):
         assert self.activation_dtype == self.weight_dtype, (
@@ -143,7 +144,9 @@ def _mx_inference_linear_transform(
         elem_dtype=config.activation_dtype,
         block_size=config.block_size,
         kernel_preference=config.kernel_preference,
-        is_swizzled_scales=config.is_swizzled_scales,
+        is_swizzled_scales=True
+        if config.swizzled_type == SwizzleType.SWIZZLE_32_4_4
+        else False,
         scaling_mode=config.scaling_mode,
     )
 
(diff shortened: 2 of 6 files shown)


liangan1 (approved):

test/prototype/mx_formats/test_mx_mm.py
```diff
@@ -16,33 +16,67 @@
...
 
 
-def _mxfp4_scaled_mm(a_data, b_data, a_scale_block, b_scale_block):
+def _mxfp4_scaled_mm(
+    a_data, b_data, a_scale, b_scale, swizzle=SwizzleType.SWIZZLE_32_4_4
+):
     """Wrapper for F.scaled_mm with MXFP4 configuration."""
+    match swizzle:
+        case SwizzleType.NO_SWIZZLE:
+            a_scale_block = a_scale.contiguous()
```
vkuzo: can this be replaced by 'assert a_scale.contiguous()`? ideally the caller prepares the scale in the right format

torchao/prototype/mx_formats/inference_workflow.py
```diff
@@ -111,6 +111,10 @@ class MXDynamicActivationMXWeightConfig(AOBaseConfig):
     # How to calculate the block scales
     scaling_mode: ScaleCalculationMode = ScaleCalculationMode.RCEIL
 
+    # How to store block scales.
+    # CUDA uses swizzle layout, XPU uses not swizzle.
+    is_swizzled_scales: bool = True
```
vkuzo: can we use the PyTorch core enum here? it can still be a bool in the tensor, to minimize risk.

torchao/prototype/mx_formats/mx_tensor.py
```diff
@@ -789,38 +789,46 @@ def _addmm_mx_dispatch(
...
-            b_scale_block = maybe_dtensor_to_blocked(b_scale)
+            b_scale_block = b.scale.view(N, K // b.block_size)
 
         if a.elem_dtype == torch.float8_e4m3fn:
             assert b.elem_dtype == torch.float8_e4m3fn
+            a_scale_v1 = a_scale_block.view(torch.float8_e8m0fnu)
+            b_scale_v1 = b_scale_block.view(torch.float8_e8m0fnu)
+            if not b.is_swizzled_scales:
+                # v1 API expects scale_b as (K//32, N)
+                b_scale_v1 = b_scale_v1.t().contiguous()
```
vkuzo: can you explain why this is needed

vkuzo: in the current design, the data should already be in the right format when it arrives to `_addmm_mx_dispatch`.  It's not expected to have to transpose or modify the layout in memory here.

vkuzo:
@claude review, and also try to answer questions from my comments

torchao/prototype/mx_formats/mx_tensor.py
```diff
@@ -789,38 +789,46 @@ def _addmm_mx_dispatch(
...
-            b_scale_block = maybe_dtensor_to_blocked(b_scale)
+            b_scale_block = b.scale.view(N, K // b.block_size)
 
         if a.elem_dtype == torch.float8_e4m3fn:
             assert b.elem_dtype == torch.float8_e4m3fn
+            a_scale_v1 = a_scale_block.view(torch.float8_e8m0fnu)
+            b_scale_v1 = b_scale_block.view(torch.float8_e8m0fnu)
+            if not b.is_swizzled_scales:
+                # v1 API expects scale_b as (K//32, N)
+                b_scale_v1 = b_scale_v1.t()
```
vkuzo: why is this needed? the caller should already have transposed the scale

test/prototype/mx_formats/test_mx_mm.py
xiaowangintel: Done

torchao/prototype/mx_formats/inference_workflow.py
xiaowangintel: Done

torchao/prototype/mx_formats/mx_tensor.py
xiaowangintel: torch._scaled_mm and F.scaled_mm expect b_scale in different layouts: (K // block_size, N) for torch._scaled_mm and (N, K // block_size) for F.scaled_mm. As a result, we currently need to handle b_scale differently depending on which API is being called.

torchao/prototype/mx_formats/mx_tensor.py
```diff
@@ -789,38 +789,46 @@ def _addmm_mx_dispatch(
...
-            b_scale = b.scale.t().view(N, K // b.block_size)
-            b_scale_block = maybe_dtensor_to_blocked(b_scale)
+            b_scale_block = b.scale.view(N, K // b.block_size)
 
         if a.elem_dtype == torch.float8_e4m3fn:
             assert b.elem_dtype == torch.float8_e4m3fn
+            a_scale_v1 = a_scale_block.view(torch.float8_e8m0fnu)
+            b_scale_v1 = b_scale_block.view(torch.float8_e8m0fnu)
+            if not b.is_swizzled_scales:
+                # v1 API expects scale_b as (K//32, N)
```
vkuzo: is this accurate?

xiaowangintel: Thanks for your comments. I’ve updated the code and expanded the MX inference tests([Refactor MXTensor: align the swizzle execution with the configured swizzle setting by xiaowangintel …](https://github.com/pytorch/ao/pull/4844/changes#diff-778d06f2a0a401ab4ca8464fda9fb5460da972acd7e20372e5f50c5dd0f94527R71)) from K=32 to K=[64, 128, 256] to ensure the correctness of the results. Since K=block_size=32 is too restrictive, it cannot effectively catch potential logic errors in the view operation. The accuracy is correct now.

xiaowangintel:
@claude review this pr.

xiaowangintel:
@claude review this pr.

xiaowangintel:
> **Claude finished @xiaowangintel's task in 4m 48s** —— [View job](https://github.com/pytorch/ao/actions/runs/33188804030)
> 
> ### Code Review
> * [x]  Read changed files & gather context
> * [x]  Review core logic (`mx_tensor.py`, `inference_workflow.py`)
> * [x]  Review tests
> 
> This iteration addresses most of the earlier thread. Confirming what's now fixed, then a few remaining items — one of which is a potential correctness concern worth verifying on XPU.
> 
> #### ✅ Resolved since the last round
> * **Type annotation** — `swizzled_type: SwizzleType = SwizzleType.SWIZZLE_32_4_4` (`inference_workflow.py:117`) is now correctly typed (was `bool`).
> * **Ternary simplification** — `is_swizzled_scales=config.swizzled_type == SwizzleType.SWIZZLE_32_4_4` (`:147`, `:158`) is clean.
> * **The core bug the PR targets** — the old non-swizzled branches in `_addmm_mx_dispatch` still called `maybe_dtensor_to_blocked` (i.e. _swizzled anyway_). The new code just views, so `NO_SWIZZLE` finally means no swizzle. This is the right fix and matches the PR motivation. ✅
> * **Latent `a`/`b` coupling** — the new `assert a.is_swizzled_scales == b.is_swizzled_scales` (`mx_tensor.py:812`) guards the case I flagged last time where `swizzle` was derived from `a` while the transpose keyed off `b`. Good defensive assert.
> 
> #### 1. Possible non-contiguous `scale_b` on the fp8 `NO_SWIZZLE` path — `mx_tensor.py:797`
> ```python
> b_scale_block = b.scale.view(K // b.block_size, N)
> ```
> 
> In `mx_linear` the weight always arrives as `args[1].t()`, and `mx_t` sets `scale = old.scale.t()` — so `b.scale` is a **transposed, non-contiguous** view already shaped `(K//bs, N)`. `.view(K//bs, N)` is therefore an identity that preserves the non-contiguous strides, and `b_scale_block` is passed straight into `torch._scaled_mm` (`:805`). If the fp8 backend requires a contiguous `scale_b` (the fp8 branch previously used `.t().contiguous()`), this could error or silently misread on the non-swizzled path.
> 
> Please confirm the mxfp8 + `NO_SWIZZLE` case is actually exercised on XPU hardware (the CI here is CUDA-only, and on CUDA this branch isn't taken). If `torch._scaled_mm` needs contiguity, add `.contiguous()`; if not, worth a one-line comment saying the non-contiguous transposed layout is intentional. [Investigate →](https://claude.ai/code?q=In%20torchao%2Fprototype%2Fmx_formats%2Fmx_tensor.py%20around%20line%20797%20on%20branch%20xw%2Fmx_logic_refactor%2C%20verify%20whether%20torch._scaled_mm%20requires%20a%20contiguous%20scale_b%20on%20the%20fp8%20NO_SWIZZLE%20path%20where%20b.scale%20is%20a%20transposed%20non-contiguous%20view&repo=pytorch/ao)
> 
> Also note the two branches now expect _different_ `b_scale` layouts — fp8 wants `(K//bs, N)`, fp4 wants `(N, K//bs)` (hence the extra `.t()` at `:820`). A one-line comment on each stating the expected shape would save the next reader the trace-through.
> 
> #### 2. Layout still modified in `_addmm_mx_dispatch` — `mx_tensor.py:819-820` (@vkuzo's point 3/4)
> ```python
> if not b.is_swizzled_scales:
>     b_scale_block = b_scale_block.t()
> ```
> 
> This is the residual of @vkuzo's "data should already be in the right format when it arrives" comment. It's now a `.t()` **view** (no `.contiguous()` copy), so the per-forward hot-path copy is gone — perf concern resolved. The remaining item is design-purity: ideally the caller / `to_mx` stores the weight scale in the API-expected layout so the dispatch only _views_. Lower priority now that it's copy-free, but still the thing the thread was pointing at.
> 
> #### 3. Comment still embeds device assumptions — `inference_workflow.py:116`
> ```python
> # How to store block scales.
> # CUDA uses swizzle layout, XPU uses not swizzle.
> ```
> 
> This is exactly what @vkuzo flagged: the config is device-agnostic, so the comment shouldn't re-encode the CUDA/XPU coupling the PR removes. Describe the _layout_ instead, e.g. "swizzled/blocked scale layout vs. plain layout; caller selects based on backend."
> 
> #### 4. Minor
> * `a_scale_block = a.scale.view(M, K // a.block_size).contiguous()` (`:792`) — for a freshly-quantized activation the scale should already be contiguous, so `.contiguous()` is a no-op guard. Fine to keep; just confirming it isn't masking an unexpected copy each forward.
> * `.view(...)` calls silently assume the underlying layout is view-compatible. With the expanded `K=[64,128,256]` test coverage (good change — `K=block_size=32` hid these bugs), that's now exercised at least on the swizzled path.
> 
> #### Positives
> * `to_mx`: `to_blocked(...)` → `maybe_dtensor_to_blocked(...)` (`:405`) is a good consolidation; the non-DTensor path falls through to `to_blocked` (`:755`), so behavior is preserved.
> * Test refactor to `torch.accelerator` + `_mxfp8_scaled_mm`/`_mxfp4_scaled_mm` helpers with an explicit `assert a_scale.is_contiguous()` in the `NO_SWIZZLE` branch is a solid way to pin the contract.
> 
> Nothing here blocks correctness for the supported CUDA path; **#1 is the one to verify before merge** (XPU fp8 non-swizzled contiguity), and #2/#3 are the remaining follow-ups from @vkuzo's thread. · branch `xw/mx_logic_refactor`

Both MXFP8 and MXFP4 are covered by test_inference_workflow_mx.

xiaowangintel:
@claude review this pr.

torchao/prototype/mx_formats/mx_tensor.py
liangan1: @xiaowangintel @vkuzo  This complex semantics are from the different API calls for mxfp8(torch._scaled_mm) and mxfp4(F.scaled_mm). I suppose this gap is due to the API data support matrix gap between the F.scaled_mm and torch._scaled_mm in the old pytorch version which is used by the PR to enabled mxfp8 in AO. F.scaled_mm has provided full data type support. 
https://github.com/pytorch/pytorch/blob/v2.13.0/aten/src/ATen/native/cuda/ScaledBlas.cpp#L1486  
```C++
// dispatch to appropriate lower-level calls for error checking & execution
  if (gemm_impl == ScaledGemmImplementation::TENSORWISE_TENSORWISE) {
    return _scaled_tensorwise_tensorwise(mat_a, mat_b, scale_a[0], scale_b[0], bias, out_dtype_, use_fast_accum, out);
  } else if (gemm_impl == ScaledGemmImplementation::ROWWISE_ROWWISE) {
    return _scaled_rowwise_rowwise(mat_a, mat_b, scale_a[0], scale_b[0], bias, out_dtype_, use_fast_accum, out);
  } else if (gemm_impl == ScaledGemmImplementation::BLOCK_128x128_1x128) {
    return _scaled_block128x128_block1x128(mat_a, mat_b, scale_a[0], scale_b[0], bias, out_dtype_, use_fast_accum, out);
  } else if (gemm_impl == ScaledGemmImplementation::BLOCK_1x128_128x128) {
    return _scaled_block1x128_block128x128(mat_a, mat_b, scale_a[0], scale_b[0], bias, out_dtype_, use_fast_accum, out);
  } else if (gemm_impl == ScaledGemmImplementation::BLOCK_1x128_1x128) {
    return _scaled_block1x128_block1x128(mat_a, mat_b, scale_a[0], scale_b[0], bias, out_dtype_, use_fast_accum, out);
  } else if (gemm_impl == ScaledGemmImplementation::MXFP8_MXFP8) {
    return _scaled_mxfp8_mxfp8(mat_a, mat_b, scale_a[0], swizzle_a_enum[0], scale_b[0], swizzle_b_enum[0], bias, out_dtype_, out);
  } else if (gemm_impl == ScaledGemmImplementation::NVFP4_NVFP4) {
    return _scaled_nvfp4_nvfp4(mat_a, mat_b, scale_a[0], swizzle_a_enum[0], scale_b[0], swizzle_b_enum[0], bias, out_dtype_, out,
                               scale_a[1], scale_b[1]);
  } else if (gemm_impl == ScaledGemmImplementation::NVFP4_NVFP4_SINGLE_SCALE) {
    return _scaled_nvfp4_nvfp4(mat_a, mat_b, scale_a[0], swizzle_a_enum[0], scale_b[0], swizzle_b_enum[0], bias, out_dtype_, out);
  } else if (gemm_impl == ScaledGemmImplementation::MXFP4_MXFP4) {
    return _scaled_mxfp4_mxfp4(mat_a, mat_b, scale_a[0], swizzle_a_enum[0], scale_b[0], swizzle_b_enum[0], bias, out_dtype_, out);
  } else {
    TORCH_CHECK_VALUE(false, "Invalid state - found an implementation, but not really");
  }
```

So, I suggest to use the unify API call like PyTorch does for all data types. F.scaled_mm is the recommended API. And then the semantic should be clean enough. The perf risk for this API change is low. \
- For CUDA, switching the MXFP8 path from `torch._scaled_mm` to `F.scaled_mm` should have low kernel-level performance risk. Both paths eventually invoke the same `_scaled_gemm` / cuBLASLt implementation when
using BlockWise1x32 with the same SWIZZLE_32_4_4 scale buffers.
- For XPU, both path all use oneDNN backend.

xiaowangintel: @vkuzo  @liangan1 The refactoring is complete, and only F.scaled_mm is retained.

xiaowangintel:
@claude review this pr.

torchao/prototype/mx_formats/mx_tensor.py
```diff
@@ -785,45 +785,35 @@ def _addmm_mx_dispatch(
...
-                scale_a=a_scale_block,
-                scale_recipe_a=ScalingType.BlockWise1x32,
-                scale_b=b_scale_block,
-                scale_recipe_b=ScalingType.BlockWise1x32,
-                swizzle_a=SwizzleType.SWIZZLE_32_4_4,
-                swizzle_b=SwizzleType.SWIZZLE_32_4_4,
-                bias=bias,
-                output_dtype=torch.bfloat16,
-            )
+            b_scale_block = b.scale.view(K // b.block_size, N).t()
```
liangan1:
```suggestion
            b_scale_block = b.scale.t().view(N, K // b.block_size)
```

torchao/prototype/mx_formats/mx_tensor.py
```diff
@@ -785,45 +785,35 @@ def _addmm_mx_dispatch(
...
         assert a.block_size == 32, f"Invalid block size {a.block_size}"
         assert b.block_size == 32, f"Invalid block size {b.block_size}"
+        assert a.is_swizzled_scales == b.is_swizzled_scales
 
         if a.is_swizzled_scales:
             a_scale_block = a.scale
         else:
-            a_scale = a.scale.view(M, K // a.block_size)
-            a_scale_block = maybe_dtensor_to_blocked(a_scale)
+            a_scale_block = a.scale.view(M, K // a.block_size).contiguous()
```
liangan1:
```suggestion
            a_scale_block = a.scale.view(M, K // a.block_size)
```

liangan1:
@claude  per your latest review comments:
1 & 2 has been resolved. 
For the t.() in the _addmm_mx_dispatch ,  I want to give more explanation:

The b arg of _addmm_mx_dispatch is b(K, N) and it from the follow 3 ops:
 https://github.com/xiaowangintel/ao/blob/52b541d1c1f35be8dfba265ba61d767f7d48b02c/torchao/prototype/mx_formats/mx_tensor.py#L836

def mx_mm(func, types, args, kwargs):  **b(K, N)** #
def mx_addmm(func, types, args, kwargs): **b(K, N)**
def mx_linear(func, types, args, kwargs): **b(N,K )**

While the F.scaled_mm need the (N,K) layout, So we need the t() in the _addmm_mx_dispatch to call F.scaled_mm. 

As for the **3: fp8 production path moved from torch._scaled_mm → F.scaled_mm — version floor?**

The current torchao baseline already requires PyTorch 2.10 or newer (8e40aef4, which dropped support for PyTorch < 2.10). PyTorch 2.10's F.scaled_mm already dispatches the BlockWise1x32 MXFP8 recipe through _scaled_mm_v2 to the MXFP8_MXFP8 CUDA implementation. Therefore, there should be no supported PyTorch version where the legacy SM100 MXFP8 path works but the public F.scaled_mm path lacks that recipe.

As for the **4:  Minor: dropped the a.elem_dtype == b.elem_dtype assertions**

F.scaled_mm rejects mixed MXFP8/MXFP4 recipes in the ATen recipe checker, so this should not cause silent incorrect results.

liangan1:
> @claude per your latest review comments: 1 & 2 has been resolved. For the t.() in the _addmm_mx_dispatch , I want to give more explanation:
> 
> The b arg of _addmm_mx_dispatch is b(K, N) and it from the follow 3 ops: https://github.com/xiaowangintel/ao/blob/52b541d1c1f35be8dfba265ba61d767f7d48b02c/torchao/prototype/mx_formats/mx_tensor.py#L836
> 
> def mx_mm(func, types, args, kwargs): **b(K, N)** # def mx_addmm(func, types, args, kwargs): **b(K, N)** def mx_linear(func, types, args, kwargs): **b(N,K )**
> 
> While the F.scaled_mm need the (N,K) layout, So we need the t() in the _addmm_mx_dispatch to call F.scaled_mm.
> 
> As for the **3: fp8 production path moved from torch._scaled_mm → F.scaled_mm — version floor?**
> 
> The current torchao baseline already requires PyTorch 2.10 or newer ([8e40aef](https://github.com/pytorch/ao/commit/8e40aef406926f0d6acd0a1ddaa8c89c909ba64d), which dropped support for PyTorch < 2.10). PyTorch 2.10's F.scaled_mm already dispatches the BlockWise1x32 MXFP8 recipe through _scaled_mm_v2 to the MXFP8_MXFP8 CUDA implementation. Therefore, there should be no supported PyTorch version where the legacy SM100 MXFP8 path works but the public F.scaled_mm path lacks that recipe.
> 
> As for the **4: Minor: dropped the a.elem_dtype == b.elem_dtype assertions**
> 
> F.scaled_mm rejects mixed MXFP8/MXFP4 recipes in the ATen recipe checker, so this should not cause silent incorrect results.

@claude  go

liangan1:
@vkuzo  can you help to review this pr again. thanks.

liangan1 (approved):

torchao/prototype/mx_formats/mx_tensor.py
```diff
@@ -402,7 +402,7 @@ def to_mx(
     if is_swizzled_scales:
         leading_dims, M, K = orig_shape[:-2], orig_shape[-2], orig_shape[-1]
         scale_shape = (math.prod(leading_dims) * M, K // block_size)
-        scale = to_blocked(scale_e8m0_biased.view(scale_shape)).flatten()
+        scale = maybe_dtensor_to_blocked(scale_e8m0_biased.view(scale_shape)).flatten()
```
vkuzo: please revert this back to the original location, dtensor code should be kept as high level as possible, it's not ideal to couple it to the function which casts to mx

liangan1:
```suggestion
        scale = to_blocked(scale_e8m0_biased.view(scale_shape)).flatten()
```

liangan1: Thanks for your guidance. Done.

torchao/prototype/mx_formats/mx_tensor.py
```diff
@@ -785,45 +785,35 @@ def _addmm_mx_dispatch(
         assert b.qdata.t().is_contiguous()
         assert a.block_size == 32, f"Invalid block size {a.block_size}"
         assert b.block_size == 32, f"Invalid block size {b.block_size}"
+        assert a.is_swizzled_scales == b.is_swizzled_scales
 
         if a.is_swizzled_scales:
             a_scale_block = a.scale
         else:
-            a_scale = a.scale.view(M, K // a.block_size)
-            a_scale_block = maybe_dtensor_to_blocked(a_scale)
```
vkuzo, line 793: this code lost `maybe_dtensor_to_blocked`, why?

liangan1: maybe_dtensor_to_blocked is to get the swizzle layout while this is no need for NO_SWIZLE path, So we should remove it.

vkuzo: you are removing it everywhere, that seems risky. Ideally this PR should not change anything about what `maybe_dtensor_to_blocked` is doing. If a change is actually needed, please explain the change thoroughly and verify (by running the relevant tests) that it is safe.

liangan1: Thanks for comments @vkuzo. In the original implementation, the scales of both activation and weight were converted to the block swizzle layout  through _**maybe_dtensor_to_blocked**_  for the NO_SWIZZLE path. Which means that, the scaled_mm always uses the swizzled layout for the NO_SWIZZLE path, and this behavior is not aligned with the semantics of the NO_SWIZZLE of F.scaled_mm in PyTorch, So we remove these 2 **_maybe_dtensor_to_blocked_** in the NO_SWIZZLE branch. 

As for the risk, we have add the NO_SWIZLLE test for MXTensor and the related CI is green. 
https://github.com/pytorch/ao/pull/4844/changes#diff-778d06f2a0a401ab4ca8464fda9fb5460da972acd7e20372e5f50c5dd0f94527R110-R116

liangan1: Let use look into this code snapshot, the F.scaled_mm always use the swizzled layout even when it is NO_SWILLZE. This should not be expected by the MXTensor API design. The implementation and API of semantics are not aligned. 

The API design:

```python 

class MXTensor(TorchAOBaseTensor):
    tensor_data_names = ["qdata", "scale"]
    tensor_attribute_names = [
        "elem_dtype",
        "block_size",
        "orig_dtype",
        "kernel_preference",
        "act_quant_kwargs",
        "is_swizzled_scales",
    ]

    def __new__(
        cls,
        qdata,
        scale,
        elem_dtype,
        block_size,
        orig_dtype,
        kernel_preference,
        act_quant_kwargs,
        is_swizzled_scales, # swizzle info is configurable 
    ):
```
SWIZZLE_32_4_4 is always used even for NO_SWILLE. 

```python 
 if b.is_swizzled_scales:
            b_scale_block = b.scale.t()
        else:
            b_scale = b.scale.t().view(N, K // b.block_size)
            b_scale_block = maybe_dtensor_to_blocked(b_scale)

        if a.elem_dtype == torch.float8_e4m3fn:
            assert b.elem_dtype == torch.float8_e4m3fn
            res = torch._scaled_mm(
                a.qdata,
                b.qdata,
                a_scale_block.view(torch.float8_e8m0fnu),
                b_scale_block.view(torch.float8_e8m0fnu),
                bias=bias,
                out_dtype=torch.bfloat16,
            )
        else:
            assert a.elem_dtype == torch.float4_e2m1fn_x2
            assert b.elem_dtype == torch.float4_e2m1fn_x2
            # FP4 operations using F.scaled_mm
            res = F.scaled_mm(
                a.qdata.view(torch.float4_e2m1fn_x2),
                b.qdata.view(torch.float4_e2m1fn_x2),
                scale_a=a_scale_block,
                scale_recipe_a=ScalingType.BlockWise1x32,
                scale_b=b_scale_block,
                scale_recipe_b=ScalingType.BlockWise1x32,
                swizzle_a=SwizzleType.SWIZZLE_32_4_4,
                swizzle_b=SwizzleType.SWIZZLE_32_4_4,
                bias=bias,
                output_dtype=torch.bfloat16,
            )

```

vkuzo: I see, I was misreading the code, I'm sorry

vkuzo:
@claude review

vkuzo:
> maybe_dtensor_to_blocked is now dead code

could you remove this function, since it's dead code now? Otherwise lgtm!

liangan1:
> > maybe_dtensor_to_blocked is now dead code
> 
> could you remove this function, since it's dead code now? Otherwise lgtm!

@xiaowangintel  let us fix it ASAP.

torchao/prototype/mx_formats/mx_tensor.py
liangan1: NP. May be we need to make it more clean in the PR desc.

xiaowangintel:
> inference_workflow.py



> > > maybe_dtensor_to_blocked is now dead code
> > 
> > 
> > could you remove this function, since it's dead code now? Otherwise lgtm!
> 
> @xiaowangintel let us fix it ASAP.

@vkuzo @liangan1 Done

vkuzo:
@claude review

vkuzo (approved):

PR #4856, opened 2026-08-31
Interleave W3A8 activations for Arm NEON prefill

JacobSzwejbka:
This interleaves packed int8 activations so NEON dot-product lanes map directly to M rows. It optimizes the 4-row and 8-row W3A8 kernels around that layout, adds dedicated handling for 3-row and other partial-M tails, and extends correctness coverage for awkward prefill shapes.\n\nThis is pull request 4 of 5 and depends on pull request 4858. Pull request 5 is 4854.


JacobSzwejbka:
The interleaved activation layout and padded M-tail handling are now included directly in pull request 4855.

PR #4857, opened 2026-08-31
Add Arm NEON W3A8 4x8 prefill kernel

JacobSzwejbka:
This adds the native non-LUT Arm NEON W3A8 4x8 dot-product kernel. It extends activation packing to four-row tiles, wires the kernel into dispatch, reuses the 2x8 path for row tails, and adds correctness coverage for weight zero points, bias, clamp, and partial rows.\n\nThis is pull request 2 of 5 and depends on pull request 4855. Pull request 3 is 4858, pull request 4 is 4856, and pull request 5 is 4854.


JacobSzwejbka:
Superseded by the simplified single 8x8 prefill kernel in pull request 4855.

PR #4859, opened 2026-08-31
Optimize Arm NEON W3A8 decode

JacobSzwejbka:
This optimizes the existing Arm NEON 1x8x16 W3A8 decode path for both asymmetric weights with groupwise zero points and symmetric weights without weight zero points.

A signed W3 weight wᵢ is stored as an unsigned value uᵢ from 0 through 7:

wᵢ = uᵢ − 4

Previously, the kernel converted every unpacked weight vector from uᵢ to wᵢ inside the inner K loop.

The original symmetric-weight integer computation is:

Y = Σᵢ(aᵢ − zₐ)wᵢ

Here, aᵢ is an activation qvalue and zₐ is the activation zero point. Substituting wᵢ = uᵢ − 4 gives:

Y = Σᵢ(aᵢ − zₐ)(uᵢ − 4)

Expanding gives:

Y = Σᵢaᵢuᵢ − zₐΣᵢuᵢ − 4Σᵢaᵢ + 4Kzₐ

The signed weight sum is:

Σᵢwᵢ = Σᵢuᵢ − 4K

Therefore:

−zₐΣᵢwᵢ = −zₐΣᵢuᵢ + 4Kzₐ

Substituting this into the expanded expression gives:

Y = Σᵢaᵢuᵢ − zₐΣᵢwᵢ − 4Σᵢaᵢ

This lets the kernel use the unpacked unsigned weights directly. For asymmetric weights, the offset is folded into the existing weight-zero correction. For symmetric weights, activation packing stores one int32 activation sum per quantization group, and the kernel applies the final correction by subtracting 4Σᵢaᵢ. The activation zero correction remains −zₐΣᵢwᵢ because activation dequantization uses aᵢ − zₐ.

The packed-weight format and AOT weight packing remain unchanged. The implementation continues to use Arm NEON dot-product instructions and does not require i8mm.

On an M5 Max at M=1 and K=4096, the asymmetric path improved compute-only latency by approximately 1.09 to 1.14 times at group size 32, 1.22 to 1.24 times at group size 128, and 1.25 to 1.29 times at group size 256 across N values of 1024, 4096, and 11008.

For symmetric weights at N=K=4096 with per-channel quantization, the complete single-panel operator, including activation quantization and packing but excluding weight packing, improved from approximately 257 microseconds to 197 microseconds, or approximately 1.30 times. Output checksums matched exactly.

Correctness coverage includes symmetric and asymmetric W3 weights, multiple output-channel tiles, multiple quantization groups, output-channel tails, bias, and clamp.

This is pull request 1 of 3.

### DIFF
--- a/torchao/csrc/cpu/torch_free_kernels/aarch64/linear/channelwise_8bit_activation_groupwise_lowbit_weight/channelwise_8bit_activation_groupwise_lowbit_weight.h
+++ b/torchao/csrc/cpu/torch_free_kernels/aarch64/linear/channelwise_8bit_activation_groupwise_lowbit_weight/channelwise_8bit_activation_groupwise_lowbit_weight.h
@@ -50,6 +50,35 @@ inline size_t packed_activations_offset(
   return (m_idx / mr) * packed_activations_size_mr_rows;
 }
 
+// The symmetric W3 decode kernel keeps packed weights in their unsigned 0..7
+// representation. Store activation sums so it can remove the resulting +4
+// offset after the dot product without changing the weight format.
+inline size_t packed_activations_with_qvals_sum_size(
+    int m,
+    int k,
+    int group_size,
+    bool has_weight_zeros,
+    int mr,
+    int kr,
+    int sr) {
+  (void)has_weight_zeros;
+  return packed_activations_size(
+      m, k, group_size, true, mr, kr, sr);
+}
+
+inline size_t packed_activations_with_qvals_sum_offset(
+    int m_idx,
+    int k,
+    int group_size,
+    bool has_weight_zeros,
+    int mr,
+    int kr,
+    int sr) {
+  (void)has_weight_zeros;
+  return packed_activations_offset(
+      m_idx, k, group_size, true, mr, kr, sr);
+}
+
 template <int mr_, int kr_, int sr_>
 void pack_activations(
     void* packed_activations,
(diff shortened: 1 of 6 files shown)


torchao/csrc/cpu/torch_free_kernels/aarch64/linear/channelwise_8bit_activation_groupwise_lowbit_weight/channelwise_8bit_activation_groupwise_lowbit_weight.h
```diff
@@ -50,6 +50,35 @@ inline size_t packed_activations_offset(
   return (m_idx / mr) * packed_activations_size_mr_rows;
 }
 
+// The symmetric W3 decode kernel keeps packed weights in their unsigned 0..7
+// representation. Store activation sums so it can remove the resulting +4
+// offset after the dot product without changing the weight format.
```
digantdesai, line 55: FYI, packing is a non-public, contract between your AOT packer and the runtime kernel. So if we want to change this or add more 2/4b to do this trick we have to worry about BC.

JacobSzwejbka: Im changing the activation packing here not weight so runtime only

torchao/csrc/cpu/torch_free_kernels/aarch64/linear/channelwise_8bit_activation_groupwise_lowbit_weight/kernel_1x8x16_f32_neondot-impl.h
```diff
@@ -190,6 +198,26 @@ void kernel_1x8x16_f32_neondot(
                 weight_q_cols67_1,
                 (uint8_t*)weight_data_byte_ptr,
                 lut);
+          } else if constexpr (use_unsigned_w3) {
+            uint8x16_t unsigned_weights[8];
+            torchao::bitpacking::vec_unpack_128_uintx_values<3>(
```
digantdesai, line 203: did you see if there is any perf optimization opportunity with unpack here (or by packing it differently)? We can revisit this iff prefill is too slow compared to 4b *or* decode is not faster than 4b.

digantdesai (approved):

torchao/csrc/cpu/torch_free_kernels/aarch64/linear/channelwise_8bit_activation_groupwise_lowbit_weight/kernel_1x8x16_f32_neondot-impl.h
JacobSzwejbka: I didnt want to mess with the weight packing

PR #4862, opened 2026-09-01
Fix MXFP8 quantize failing when it is the first CUDA call on a thread

anijain2305:
Written by Agent

## Summary

`mxfp8_quantize_cuda` fails with `CUDA error: an illegal instruction was encountered`
whenever it is the **first CUDA kernel launched on a given thread**. Any prior CUDA op
on that thread avoids it. Shape, dtype, alignment and scaling orientation are all
irrelevant.

This is easy to hit in practice because **autograd runs backward on worker threads**.

## Repro

Fails on `main`, passes with this PR. Use a fresh process: the fault leaves the CUDA
context in a sticky error state.

```python
import threading
import torch
from torchao.prototype.mx_formats.kernels import mxfp8_quantize_cuda

t = torch.randn(64, 128, device="cuda", dtype=torch.bfloat16)
err = []

def target():
    try:
        mxfp8_quantize_cuda(t, rowwise=False, colwise=True, scaling_mode="rceil")
        torch.cuda.synchronize()
    except Exception as e:
        err.append(e)

th = threading.Thread(target=target)
th.start()
th.join()
print("FAULT" if err else "ok")
```

`compute-sanitizer` attributes it to the TMA kernel:

```
========= Illegal instruction
=========     at void mxfp8_quantize_kernel<__nv_bfloat16, __nv_fp8_e4m3, 32ul, 1ul,
              (ScaleCalculationMode)1>(CUtensorMap_st, CUtensorMap_st, CUtensorMap_st, ...)+0x270
=========     Host Frame: MXFP8Quantizer::quantize(...) in _C_mxfp8...so
```

## Root cause

The op encodes TMA descriptors with `cuTensorMapEncodeTiled`, a **driver** API that
requires a current CUDA context. The runtime binds the primary context to a thread
lazily, so a thread that has not used CUDA has none:

```
main thread          : cuCtxGetCurrent -> 0xaaab0d10e1b0
fresh thread, before : cuCtxGetCurrent -> NULL      <- descriptors are encoded here
fresh thread, after  : cuCtxGetCurrent -> 0xaaab0d10e1b0
```

`mxfp8_quantize` does already construct a `tsa::DeviceGuard` before encoding, but that
does not force the binding: `c10::cuda::ExchangeDevice` returns early without calling
`cudaSetDevice` when the requested device already equals the current one, which is
exactly the single-device case.

```cpp
if (cur_device < 0) {
    C10_CUDA_CHECK(cudaGetDevice(&tmp_device));
    cur_device = static_cast<DeviceIndex>(tmp_device);
    if (to_device == cur_device) {
      return cur_device;      // no cudaSetDevice, so no context is bound
    }
}
```

Encoding then returns `CUDA_ERROR_INVALID_CONTEXT`, and because the result was
discarded the kernel launched with an unpopulated `CUtensorMap` and executed TMA
instructions against a garbage descriptor.

## Changes

1. **Bind the context explicitly** before encoding, in `mxfp8_quantize_cuda`.
2. **Check the `cuTensorMapEncodeTiled` result** in `create_2D_tensor_map` and
   `create_3D_tensor_map_output`. Both discarded it, which is why a failed encode
   surfaced as an illegal instruction inside the kernel -- or, worse, as a sticky error
   on an unrelated later launch, which makes this very hard to diagnose.
   `mx_block_rearrange_2d_M_groups.cu` already checks its result; this makes the two
   paths consistent.

## Test plan

New `test_cuda_mx_quantize_on_a_fresh_thread` runs the quantize from a fresh
`threading.Thread`; it fails on `main` and passes here.

| check | before | after |
| --- | --- | --- |
| cold-thread repro | FAULT | ok |
| autograd-backward repro | FAULT | ok |
| warm-with matrix (none / current_device / kernel) | `none` FAULT | all ok |
| `test/prototype/mx_formats/test_kernels.py` | -- | 150 passed, 2 skipped |

Verified on NVIDIA GB200 (sm_100), CUDA 13.0, driver 580.126.20. The same code path
also serves sm_90a, which I was not able to test.

### DIFF
--- a/test/prototype/mx_formats/test_kernels.py
+++ b/test/prototype/mx_formats/test_kernels.py
@@ -601,6 +601,41 @@ def test_rearrange(shape):
     not is_cuda_version_at_least(12, 8),
     reason="CUDA version >= 12.8 required for MXFP8 CUDA kernels",
 )
+def test_cuda_mx_quantize_on_a_fresh_thread():
+    """The kernel must work when it is the first CUDA call on its thread.
+
+    It encodes TMA descriptors through the driver API, which needs a current
+    context. A thread that has not used CUDA yet does not have one, and a
+    DeviceGuard does not bind it when the device already matches. Autograd runs
+    backward on worker threads, so this is the common case for an MXFP8
+    backward. Without the fix the kernel raises a cudaErrorIllegalInstruction.
+    """
+    import threading
+
+    x = torch.randn(64, 128, device="cuda", dtype=torch.bfloat16)
+    failure = []
+
+    def run():
+        try:
+            mxfp8_quantize_cuda(x, rowwise=False, colwise=True, scaling_mode="rceil")
+            torch.cuda.synchronize()
+        except Exception as e:  # noqa: BLE001
+            failure.append(e)
+
+    thread = threading.Thread(target=run)
+    thread.start()
+    thread.join()
+    assert not failure, f"quantize failed on a fresh thread: {failure[0]}"
+
+
+@pytest.mark.skipif(
+    not is_sm_at_least_100(),
+    reason="MXFP8 requires CUDA capability 10.0 or greater",
+)
+@pytest.mark.skipif(
+    not is_cuda_version_at_least(12, 8),
+    reason="CUDA version >= 12.8 required for MXFP8 CUDA kernels",
+)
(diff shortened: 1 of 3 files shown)


vkuzo:
@claude review

vkuzo (approved):

PR #4863, opened 2026-09-02
Honor range-learned scales in the tied embedding quantization path (#4863)

telgamal-1:
Summary:
X-link: https://github.com/pytorch/executorch/pull/22476

For a `share_emb: true` checkpoint trained with
`weight_quantization_min_max_learning: true`, the tied embedding path silently
discarded every range-learned weight scale and re-derived qparams from the
dequantized weights. The untied path did not, so the same checkpoint exported
two ways differed by ~6.5 dB SNR against the rlformers reference.

Learned ranges survive export only via `QATConfig(step="convert")`, whose
transform pulls `custom_scale`/`custom_zero_point` off the module's
`weight_fake_quantizer` and disables qparam re-derivation. The untied path wraps
both its embedding and linear configs in it. `TiedEmbeddingQuantizer` used a bare
`Int8DynamicActivationIntxWeightConfig`, so the convert step never ran.

`_apply_embedding_quantization` already receives `range_learning`, but returned
early into `apply_torchao_embedding_quantization` without forwarding it. Thread
it through to `TiedEmbeddingQuantizer` and wrap its config the same way the
non-shared path does.

The `QATConfig` import in `TiedEmbeddingQuantizer` is function-local:
`torchao.quantization.qat` pulls in `torchao.prototype.qat`, which cycles back
through this module at import time.

Only the tied path changes. `EmbeddingQuantizer` (torchao embedding quantization
without shared embeddings) has the same gap and is left alone here.

Reviewed By: YIWENX14

Differential Revision: D118365125

### DIFF
--- a/torchao/prototype/quantization/embedding/api.py
+++ b/torchao/prototype/quantization/embedding/api.py
@@ -294,10 +294,12 @@ class TiedEmbeddingQuantizer:
         weight_dtype: torch.dtype = torch.int4,
         granularity: Granularity = PerAxis(0),
         mapping_type: MappingType = MappingType.ASYMMETRIC,
+        range_learning: bool = False,
     ):
         self.weight_dtype = weight_dtype
         self.granularity = granularity
         self.mapping_type = mapping_type
+        self.range_learning = range_learning
 
     def quantize(
         self,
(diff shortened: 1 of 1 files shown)


andrewor14:
@claude review

torchao/prototype/quantization/embedding/api.py
```diff
@@ -294,10 +294,12 @@ def __init__(
         weight_dtype: torch.dtype = torch.int4,
         granularity: Granularity = PerAxis(0),
         mapping_type: MappingType = MappingType.ASYMMETRIC,
+        range_learning: bool = False,
```
andrewor14, line 297: I guess this Quantizer itself isn't doing range_learning, but only loading in ranged learned qparams? @telgamal-1 @claude can you suggest a better name for this?

andrewor14 (approved):
Stamping to unblock, but please add a test.

@claude can you add a short test for this path? Don't push, just print it out here and tell us where to put it (make sure it passes)

Issue #4845, opened 2026-08-28
Int8Tensor/Float8Tensor missing aten.view/abs/arithmetic op registrations in 0.18 (quantize fails silently, unquantized)


JWLHS:
## Summary

torchao `0.18.0+xpu` rewrote the quantized tensor subclasses (`Int8Tensor` / `Float8Tensor`) to use `__torch_dispatch__` tables, but several operators are **missing from the dispatch registrations** of exactly these two classes. This breaks `quantize_()` whenever the flow calls `aten.view` on an `Int8Tensor` or `aten.abs`/arithmetic ops on a `Float8Tensor` — e.g. re-quantizing an already-quantized module goes through `_choose_qparams_affine` which calls `input.view(...)`.

Worse, the failure is **swallowed by per-module try/except in callers**: the module is left **unquantized** and only a `Failed to quantize ...` message is printed, so users believe quantization is on when it isn't.

## Environment

- OS: Windows 11
- Python: 3.12.9
- torch: `2.13.0+xpu` (Intel Arc A770)
- torchao: `0.18.0+xpu` (from `https://download.pytorch.org/whl/xpu`)

## Reproduction

```python
import torch
from torchao.quantization.quant_api import quantize_ as torchao_quantize_
from torchao.quantization.quant_api import Int8WeightOnlyConfig, Float8WeightOnlyConfig

model = torch.nn.Sequential(torch.nn.Linear(64, 128), torch.nn.Linear(128, 16)).to("xpu")

# First pass: OK (weights become Int8Tensor)
torchao_quantize_(model, Int8WeightOnlyConfig())

# Second pass on an already-quantized module (what ai-toolkit's quantize() does
# when it quantizes the container, then re-visits each child Linear):
torchao_quantize_(model[0], Int8WeightOnlyConfig())
# NotImplementedError: Int8Tensor dispatch: attempting to run unimplemented
# operator/function: func=<OpOverload(op='aten.view', overload='default')>, ...

# Direct repro:
model[0].weight.view(64, 128)  # same NotImplementedError
```

Float8Tensor equivalent:

```python
model = torch.nn.Sequential(torch.nn.Linear(64, 128), torch.nn.Linear(128, 16)).to("xpu")
torchao_quantize_(model, Float8WeightOnlyConfig())
model[0].weight.abs()  # NotImplementedError: Float8Tensor dispatch: aten.abs unimplemented
torchao_quantize_(model[0], Float8WeightOnlyConfig())
[...]

PR #4873, opened 2026-09-08
Make quantize_ skip already-quantized weights

vkuzo:
Summary:
A second quantize_() over a module whose weight is already a TorchAO
quantized tensor re-ran from_hp() on it, hitting unimplemented ops
(aten.view for Int8Tensor, aten.abs for Float8Tensor) and raising
NotImplementedError (#4845). This extends the default _is_linear filter
to skip any weight that is already a TorchAOBaseTensor (with a warning),
making re-application a no-op for all configs. The existing QAT
_AffineFakeQuantizedTensor skip is kept first and silent.

Test Plan:
pytest test/quantization/test_quant_api.py -k reapply
pytest test/quantization/test_qat.py -k "convert or prepare or ptq or api"

### DIFF
--- a/torchao/quantization/quant_api.py
+++ b/torchao/quantization/quant_api.py
@@ -167,6 +167,7 @@ def _replace_with_custom_fn_if_matches_filter(
 
 def _is_linear(mod, *args):
     # avoid circular dependencies
+    from torchao.prototype.gptq.observer import GPTQObserverTensor
     from torchao.quantization.qat.affine_fake_quantized_tensor import (
         _AffineFakeQuantizedTensor,
     )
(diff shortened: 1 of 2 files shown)


vkuzo:
* __->__ #4873

andrewor14:
@claude review

torchao/quantization/quant_api.py
```diff
@@ -170,16 +170,33 @@ def _is_linear(mod, *args):
...
-        and not isinstance(mod.weight, _AffineFakeQuantizedTensor)
         and not isinstance(mod, nn.modules.linear.NonDynamicallyQuantizableLinear)
-    )
+    ):
+        return False
+
+    # Skip fake-quantized (QAT) weights. This is intended when a weight is shared
+    # by multiple linear modules, and is not an error, so skip silently.
+    if isinstance(mod.weight, _AffineFakeQuantizedTensor):
+        return False
```
andrewor14, line 186: fine to keep this case, but no one is actually using this path anymore. I should just delete it

andrewor14 (approved):
Stamping to unblock, but looks like the CI failures are real. For now maybe we can just skip the GPTQ tests and delete it for real later? This PR does break GPTQ I think.

andrewor14:
@claude survey all other `TorchAOBaseTensor`s in the repo, is `GPTQObserverTensor` the only one where we would call `quantize_` on?

PR #4874, opened 2026-09-08
Add code owners for low-bit CPU kernels

JacobSzwejbka:
Adds @digantdesai and @metascroy as the code owners for the low-bit CPU kernel implementation under:

- torchao/csrc/cpu/shared_kernels/
- torchao/csrc/cpu/torch_free_kernels/

These specific rules override the broader torchao/csrc ownership rule for those trees.

Related W3 activation-packing parallelization: #4854

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -25,8 +25,8 @@
 /torchao/csrc/ @jerryzh168 @vkuzo @andrewor14
 
 # Low-bit CPU kernels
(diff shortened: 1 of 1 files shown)


metascroy (approved):

vkuzo (commented):
can we add vkuzo and andrewor14 to these dirs as well, to ensure oncall can always fix

vkuzo (approved):

PR #4875, opened 2026-09-08
Build CUDA wheels for ARM64

andrewor14:
Build CUDA wheels in addition to CPU-only wheels for aarch64, so users on machines with GB200 and GB300 can access kernels like `mxfp8_quantize_cuda`. These cuda kernels are only accessible on x86_64 machines today (B200 and B300).


andrewor14:
@claude review

vkuzo (approved):

PR #4363, opened 2026-05-01
Fix regex escape characters in documentation

DrStone1971:
Fix for vllm

error:

vllm
/home/tore/.venvs/comfy/lib/python3.12/site-packages/torchao/quantization/quant_api.py:1501: SyntaxWarning: invalid escape sequence '\.'
  """Configuration class for applying different quantization configs to modules or parameters based on their fully qualified names (FQNs).

PR #4876, opened 2026-09-09
Fix invalid escape sequences and select W605 so they stay fixed

Dev-next-gen:
Importing `torchao.quantization.quant_api` on Python 3.12+ prints:

```
torchao/quantization/quant_api.py:1542: SyntaxWarning: invalid escape sequence '\.'
```

There are six such literals in the repo, twelve escapes in total:

| file | what it is |
|---|---|
| `torchao/quantization/quant_api.py:1542` | `FqnToConfig` docstring, in the `re:` examples |
| `benchmarks/float8/utils.py` ×5 | regexes written as plain (non-raw) string literals |

The benchmark ones are the more interesting half. They are regexes — `re.match("\# kernel path: .*", line)`, `re.match("([\w_]+) = async_compile.*", line)` and so on — that happen to work only because Python passes an unrecognised escape through unchanged. Python has warned about this since 3.12 and [makes it a `SyntaxError` in 3.15](https://docs.python.org/3/whatsnew/3.12.html#deprecated), at which point `quant_api` stops importing at all.

## This changes nothing today

Making the literals raw is a no-op for current behaviour, because Python already kept the backslashes:

```
identical=True  compiled_identical=True   r'.* ([0-9\.]+)GB/s.*(triton_[a-z_0-9]+)'
identical=True  compiled_identical=True   r'.* \[__output_code\] (.*)'
identical=True  compiled_identical=True   r'\# kernel path: .*'
identical=True  compiled_identical=True   r'([\w_]+) = async_compile.*'
identical=True  compiled_identical=True   r"''', device_str='cuda'\)"
```

Same strings, same compiled patterns. The docstring becomes `r"""` so the documented regexes keep rendering as written rather than needing doubled backslashes.

## Why this also touches ruff.toml

The one-line `quant_api.py` fix has been submitted three times already — #3825 (closed), #4035 and #4363 (both still open) — and has not stuck. All three touch only that docstring; none touch the five benchmark regexes, and none touch the lint config. Nothing in CI can see this class of problem: `ruff.toml` selects `["F", "I"]`, and the rule that catches invalid escape sequences is `W605`, which lives in `W`.

So the file's own header, `# NOTE: The AO repo is completely linted.`, is not quite true for this class. Selecting `W605` makes it true and stops the next one from landing.

If you would rather take one of the existing PRs for the docstring, I am happy to rebase this down to the benchmark regexes plus the lint rule — the point of it is the rule.

## Verification

`ruff check` over the repo (venv excluded), pinned rule sets:

| | `["F", "I"]` | `["F", "I", "W605"]` |
|---|---|---|
| this branch | All checks passed | All checks passed |
| `main` | All checks passed | **Found 12 errors** |

So the added rule flags exactly what this PR fixes, and the fix introduces no findings under the existing selection. A `compile()` sweep over every `.py` in the repo reports zero `SyntaxWarning`s on this branch, against six on `main`.

### DIFF
--- a/benchmarks/float8/utils.py
+++ b/benchmarks/float8/utils.py
@@ -131,7 +131,7 @@ def parse_bw_and_kernel_name(line):
         0.257ms         0.537 GB         2092.43GB/s     triton_red_fused_native_layer_norm_0
     Output: the bandwidth value and the kernel name, or None and None
     """
-    result = re.search(".* ([0-9\.]+)GB/s.*(triton_[a-z_0-9]+)", line)
+    result = re.search(r".* ([0-9\.]+)GB/s.*(triton_[a-z_0-9]+)", line)
     if result:
         return result.group(1), result.group(2)
     else:
@@ -321,7 +321,7 @@ def update_triton_kernels_in_prof_chome_trace_with_torch_logs(
     torch_logs_only = []
     for line in torch_logs_str:
         line = line.replace("\n", "")
-        match = re.match(".* \[__output_code\] (.*)", line)
+        match = re.match(r".* \[__output_code\] (.*)", line)
         if match:
             torch_logs_only.append(match.group(1))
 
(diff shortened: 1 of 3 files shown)


vkuzo:
@claude review

vkuzo (approved):

PR #4840, opened 2026-08-27
Remove Float8 allow-in-graph custom autograd wrappers (#4840)

Jiapli930:
Summary:

Route four TorchAO Float8 autograd.Function classes through the standard Dynamo AutogradFunctionApply HOP so AOTAutograd cache can key and serialize them without local trampoline closures.

Post: https://fb.workplace.com/groups/1075192433118967/permalink/2032824680689066/

 {F1995106320}

Differential Revision: D117758450

PR #4879, opened 2026-09-10
Make float8 autograd Function allow_in_graph opt-in via config

vkuzo:
Summary:

PR #4840 removed the `@torch._dynamo.allow_in_graph` decorators from the
float8 training-path autograd Functions so that AOTAutograd could cache
the compiled graphs. Without the decorator, Dynamo routes these Functions
through the `AutogradFunctionApply` HOP and traces their forward/backward
into static fx subgraphs. This freezes data-dependent Python branches
(e.g. `tensor_already_casted_to_fp8`) at trace time: the backward is
traced in isolation with a synthesized high-precision cotangent, so the
`grad_output` fp8 guard is constant-folded into an unconditional
high-precision cast. At runtime (e.g. tensor parallel, where
`NoopFwToFloat8BwDynamic.backward` casts the grad to fp8 first) a
`Float8TrainingTensor` flows into that frozen `torch.abs` node and fails
with `NotImplementedError: aten.abs.default`, which broke the distributed
CI job.

This change makes the behavior opt-in instead of a hard switch:

- Add `Float8LinearConfig._autograd_fn_allow_in_graph`, defaulting to
  `True` (the pre-#4840 behavior).
- Restore the `@torch._dynamo.allow_in_graph` decorators, but split each
  Function into an undecorated `_*_impl` holding the logic plus a thin
  decorated subclass. This covers `matmul_with_hp_or_float8_args`,
  `NoopFwToFloat8BwDynamic`, `_ToFloat8ConstrFunc`, and
  `_FromFloat8ConstrFunc`.
- Call sites select the decorated class vs. the `_impl` based on
  `config._autograd_fn_allow_in_graph`, threading the flag through
  `hp_tensor_to_float8_dynamic` / `hp_tensor_and_scale_to_float8` and the
  tensor-parallel input/output prepare hooks.

With the default (`True`), fusion and the data-dependent branches behave
correctly (fixing the CI regression). Users who want AOTAutograd caching
can set the flag to `False` and accept the static-graph limitations. The
`test_inductor_aot_autograd_cache_rowwise_with_gw_hp` cache test is
updated to opt into `False`, since caching is the path it exercises.

Test Plan:

pytest test/float8/test_compile.py::test_inductor_aot_autograd_cache_rowwise_with_gw_hp
NCCL_DEBUG=WARN torchrun --nproc_per_node 2 test/float8/test_dtensor.py

Both pass. Also verified eager forward/backward for both flag values.

### DIFF
--- a/test/float8/test_compile.py
+++ b/test/float8/test_compile.py
@@ -252,6 +252,11 @@ def test_inductor_aot_autograd_cache_rowwise_with_gw_hp():
     config = Float8LinearConfig.from_recipe_name(
         Float8LinearRecipeName.ROWWISE_WITH_GW_HP
     )
+    # AOTAutograd caching requires the plain (undecorated) autograd Functions;
+    # the `allow_in_graph` trampolines are not cacheable. Opt out of
+    # `allow_in_graph` so the cache path under test is exercised.
+    # hack around config being frozen
+    object.__setattr__(config, "_autograd_fn_allow_in_graph", False)
     model = Float8Linear.from_float(
         nn.Linear(16, 32, bias=True, device=device, dtype=torch.bfloat16),
         config,
--- a/torchao/float8/config.py
+++ b/torchao/float8/config.py
@@ -245,6 +245,22 @@ class Float8LinearConfig:
     # same value in the forward pass as the backward passes.
     round_scales_to_power_of_2: bool = False
 
+    # If True (the default), the float8 `torch.autograd.Function`s used in the
+    # training path (matmul, casts, and the grad_output cast in tensor parallel)
+    # are decorated with `@torch._dynamo.allow_in_graph`. This keeps them opaque
+    # to Dynamo (Dynamo does not trace their Python bodies) while still letting
+    # AOTAutograd trace through them and Inductor fuse across the boundary.
+    #
+    # If False, the plain (undecorated) autograd Functions are used. Dynamo then
+    # routes them through the `AutogradFunctionApply` HOP, tracing forward and
+    # backward into static fx subgraphs. This is compatible with AOTAutograd
+    # caching, but the static subgraphs freeze any data-dependent Python branches
+    # (e.g. `tensor_already_casted_to_fp8`) at trace time, which can be incorrect
+    # when the runtime dtype differs from what was traced (see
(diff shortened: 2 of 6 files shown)


vkuzo:
* __->__ #4879

andrewor14:
@claude review

torchao/float8/config.py
```diff
@@ -245,6 +245,22 @@ class Float8LinearConfig:
...
+    #
+    # If False, the plain (undecorated) autograd Functions are used. Dynamo then
+    # routes them through the `AutogradFunctionApply` HOP, tracing forward and
+    # backward into static fx subgraphs. This is compatible with AOTAutograd
+    # caching, but the static subgraphs freeze any data-dependent Python branches
+    # (e.g. `tensor_already_casted_to_fp8`) at trace time, which can be incorrect
+    # when the runtime dtype differs from what was traced (see
+    # https://github.com/pytorch/ao/pull/4840). Set to False only if you
+    # understand this tradeoff.
+    _autograd_fn_allow_in_graph: bool = True
```
andrewor14, line 262: Is this bug triggered only with TP, since that path casts the grad output to fp8 (and `tensor_already_casted_to_fp8` takes the wrong path and tries to cast it twice)? So the user would only want to enable this flag if they're not using TP?

test/float8/test_compile.py
```diff
@@ -252,6 +252,11 @@ def test_inductor_aot_autograd_cache_rowwise_with_gw_hp():
     config = Float8LinearConfig.from_recipe_name(
         Float8LinearRecipeName.ROWWISE_WITH_GW_HP
     )
+    # AOTAutograd caching requires the plain (undecorated) autograd Functions;
+    # the `allow_in_graph` trampolines are not cacheable. Opt out of
+    # `allow_in_graph` so the cache path under test is exercised.
+    # hack around config being frozen
+    object.__setattr__(config, "_autograd_fn_allow_in_graph", False)
```
andrewor14, line 259: why not just set `config._autograd_fn_allow_in_graph` directly?

andrewor14 (approved):
Looks good. I think claude's comment (1) is worth fixing.

@claude can you also confirm that this PR adds back all the decorators removed in https://github.com/pytorch/ao/pull/4840 in the default case?

PR #4882, opened 2026-09-11
Support MXTensor cat for shared-input linear fusion

lchen2331:
## Summary

Add `aten.cat` support for `MXTensor` so `batch_linear_lhs` can concatenate weights for linear layers sharing the same input, such as Q/K/V and gate/up projections, to reduce redundant activation quantization and GEMM launches and improve model performance.

## Implementation

- Support concatenation along the output-feature dimension (`dim=0`) for 2D weights with contiguous qdata and matching quantization metadata.
- Concatenate qdata directly, without dequantization or requantization.
- Concatenate plain scales directly. For swizzled scales, restore the logical layout, concatenate, then repack to handle per-input padding correctly.

## Tests

Cover aligned and unaligned shapes, plain and swizzled scales, byte-level equality, and rejection of incompatible weights.
### Model validation
Tested on the Llama 3.1 8B model.

Before fusion, the Q, K, and V projections are executed as three separate linear operations. Although they share the same input activation, its quantized result is redundantly written to three separate output buffers:
```
        # File: transformers/src/transformers/models/llama/modeling_llama.py:262 in forward, code: query_states = self.q_proj(hidden_states).view(hidden_shape).transpose(1, 2)
        l__self___model_layers_0_self_attn_q_proj_weight: "MXTensor(bf16[4096, 4096])" = self.L__self___model_layers_0_self_attn_q_proj_weight
        linear: "bf16[1, 938, 4096]" = torch._C._nn.linear(hidden_states_2, l__self___model_layers_0_self_attn_q_proj_weight, None);  l__self___model_layers_0_self_attn_q_proj_weight = None
        view: "bf16[1, 938, 32, 128]" = linear.view((1, 938, -1, 128));  linear = None
        query_states: "bf16[1, 32, 938, 128]" = view.transpose(1, 2);  view = None

        # File: transformers/src/transformers/models/llama/modeling_llama.py:263 in forward, code: key_states = self.k_proj(hidden_states).view(hidden_shape).transpose(1, 2)
        l__self___model_layers_0_self_attn_k_proj_weight: "MXTensor(bf16[1024, 4096])" = self.L__self___model_layers_0_self_attn_k_proj_weight
        linear_1: "bf16[1, 938, 1024]" = torch._C._nn.linear(hidden_states_2, l__self___model_layers_0_self_attn_k_proj_weight, None);  l__self___model_layers_0_self_attn_k_proj_weight = None
        view_1: "bf16[1, 938, 8, 128]" = linear_1.view((1, 938, -1, 128));  linear_1 = None
        key_states: "bf16[1, 8, 938, 128]" = view_1.transpose(1, 2);  view_1 = None

        # File: transformers/src/transformers/models/llama/modeling_llama.py:264 in forward, code: value_states = self.v_proj(hidden_states).view(hidden_shape).transpose(1, 2)
        l__self___model_layers_0_self_attn_v_proj_weight: "MXTensor(bf16[1024, 4096])" = self.L__self___model_layers_0_self_attn_v_proj_weight
        linear_2: "bf16[1, 938, 1024]" = torch._C._nn.linear(hidden_states_2, l__self___model_layers_0_self_attn_v_proj_weight, None);  hidden_states_2 = l__self___model_layers_0_self_attn_v_proj_weight = None
        view_2: "bf16[1, 938, 8, 128]" = linear_2.view((1, 938, -1, 128));  linear_2 = None
        value_states: "bf16[1, 8, 938, 128]" = view_2.transpose(1, 2);  view_2 = None
```
The corresponding quantization kernel contains three identical stores:

```
tl.store(out_ptr0 + (x0), tmp4, None)
tl.store(out_ptr1 + (x0), tmp4, None)
tl.store(out_ptr2 + (x0), tmp4, None)
```

After fusion, the three projection weights are concatenated and the Q/K/V projections are computed with a single matrix multiplication, followed by a split of the output:
```
        cat_65: "MXTensor(bf16[6144, 4096])" = torch.cat([l__self___model_layers_0_self_attn_q_proj_weight, l__self___model_layers_0_self_attn_k_proj_weight, l__self___model_layers_0_self_attn_v_proj_weight], dim = 0);  l__self___model_layers_0_self_attn_q_proj_weight = l__self___model_layers_0_self_attn_k_proj_weight = l__self___model_layers_0_self_attn_v_proj_weight = None
        t: "MXTensor(bf16[4096, 6144])" = torch.t(cat_65);  cat_65 = None
        mm: "bf16[938, 6144]" = torch.mm(flatten, t);  flatten = t = None
        reshape_32: "bf16[1, 938, 6144]" = torch.reshape(mm, (sym_size_int, sym_size_int_1, 6144));  mm = sym_size_int = sym_size_int_1 = None
        split = torch.functional.split(reshape_32, [4096, 1024, 1024], dim = -1);  reshape_32 = None
```
The shared quantized activation is now written only once:
```
tl.store(out_ptr0 + (x0), tmp4, None)
```
This reduces the Q/K/V path from three linear operations to one matrix multiplication and removes two redundant stores of the shared quantized activation.


test/prototype/mx_formats/test_mx_tensor.py
```diff
@@ -47,6 +48,61 @@
...
+@pytest.mark.parametrize(
+    "rows,width,dim", [((128, 256), 128, 0), ((127, 2, 129), 160, -2)]
+)
+def test_cat(rows, width, dim):
+    device = torch.accelerator.current_accelerator()
+    is_swizzled_scales = device.type == "cuda" and not is_ROCM()
+    inputs = [torch.randn(n, width, device=device, dtype=torch.bfloat16) for n in rows]
+    mx_inputs = [
+        MXTensor.to_mx(
+            x, torch.float8_e4m3fn, 32, is_swizzled_scales=is_swizzled_scales
```
xiaowangintel: 1. Using torch.float8_e4m3fn only covers MXFP8, so the test coverage is too narrow. Please use elem_dtype to cover MXFP8, MXFP4, etc., and pass it as a function parameter.
2. According to the MX recipe, the default BLOCK_SIZE should be sufficient; no additional configuration is needed.

test/prototype/mx_formats/test_mx_tensor.py
```diff
@@ -47,6 +48,61 @@
...
+        )
+        for x in inputs
+    ]
+    result = torch.cat(mx_inputs, dim=dim)
+    expected = MXTensor.to_mx(
+        torch.cat(inputs),
+        torch.float8_e4m3fn,
+        32,
```
xiaowangintel: ditto

test/prototype/mx_formats/test_mx_tensor.py
```diff
@@ -47,6 +48,61 @@
...
+        is_swizzled_scales=is_swizzled_scales,
+    )
+    assert isinstance(result, MXTensor)
+    assert result.shape == expected.shape
+    assert result.is_swizzled_scales == is_swizzled_scales
+    for field in ("qdata", "scale"):
```
xiaowangintel, line 79: This check can be changed to a numerical comparison between result and expected.

test/prototype/mx_formats/test_mx_tensor.py
```diff
@@ -47,6 +48,61 @@
...
+    not torch.accelerator.is_available(), reason="Accelerator not available"
+)
+def test_cat_rejects_incompatible_weights():
+    device = torch.accelerator.current_accelerator()
+    is_swizzled_scales = device.type == "cuda" and not is_ROCM()
+    first, second = [
+        MXTensor.to_mx(
+            torch.randn(rows, 32, device=device, dtype=torch.bfloat16),
+            torch.float8_e4m3fn,
+            32,
```
xiaowangintel: Using torch.float8_e4m3fn only covers MXFP8, so the test coverage is too narrow. Please use elem_dtype to cover MXFP8, MXFP4, etc., and pass it as a function parameter.
According to the MX recipe, the default BLOCK_SIZE should be sufficient; no additional configuration is needed.

xiaowangintel:
@claude review this pr.

test/prototype/mx_formats/test_mx_tensor.py
lchen2331: Thanks for pointing this out. I've updated it to use `elem_dtype` as a function parameter, with the default `BLOCK_SIZE` from the MX recipe.

test/prototype/mx_formats/test_mx_tensor.py
lchen2331: Thanks. I've added the numerical comparison.

lchen2331:
> **Claude finished @xiaowangintel's task in 2m 32s** —— [View job](https://github.com/pytorch/ao/actions/runs/34918140098)
> 
> ### PR Review: Support MXTensor cat for shared-input linear fusion
> * [x]  Read `torchao/prototype/mx_formats/mx_tensor.py` (`mx_cat`)
> * [x]  Read test changes in `test/prototype/mx_formats/test_mx_tensor.py`
> * [x]  Verify helper symbols (`from_blocked`, `to_blocked`, `hp_data_dims_to_swizzled_scale_dims_mx`, `is_ROCM`)
> * [x]  Check whether prior review comments were addressed
> * [x]  Assess correctness across formats (MXFP8/FP6/FP4) and swizzled scales
> 
> Nice, focused change. The `mx_cat` implementation is clean and the byte-level equality test is the right way to validate that cat preserves the quantized representation. Below is my feedback.
> 
> ### Prior review comments — look addressed ✅
> The earlier comments from @xiaowangintel appear resolved by later commits:
> 
> * **Use `elem_dtype` to cover MXFP8/MXFP4/etc.** → `test_cat` now parametrizes over `SUPPORTED_ELEM_DTYPES` (commit `659ccd5`).
> * **Default `BLOCK_SIZE` is sufficient, no extra config** → tests now call `MXTensor.to_mx(x, elem_dtype, is_swizzled_scales=...)` with the default block size.
> * **Change the check to a numerical comparison** → test now does `torch.testing.assert_close(result.dequantize(), expected.dequantize(), rtol=0, atol=0)` plus byte-level `torch.equal` on `qdata`/`scale`.
> 
> ### Correctness — looks sound
> * **Non-swizzled path** (`mx_tensor.py:1013-1014`): scales viewed as `uint8` then `cat` along dim 0 matches `to_mx(cat(inputs))` layout. Viewing as `uint8` to dodge unsupported CUDA E8M0 `cat` layouts is a reasonable workaround, consistent with the existing all-gather code.
> * **Swizzled path** (`mx_tensor.py:1015-1024`): restoring logical rows via `from_blocked` per input, concatenating, then `to_blocked` once for the combined shape correctly handles per-input padding. This mirrors how `to_mx` produces swizzled scales, so the byte-equality assertion should hold.
> * **FP4 packing**: `cat` along dim 0 is safe since packing is along the last dim; `MXTensor.__new__` unpacks the size correctly. Good.
> * **Validation loop**: mismatched dim-1 sizes are caught by `aten.cat` on `qdata`; transposed (non-contiguous) inputs and mismatched metadata are rejected explicitly. Covered by `test_cat_rejects_incompatible_weights`.
> 
> ### Suggestions / questions
> 1. **`test_cat` runs `is_swizzled_scales=True` on every accelerator, but the rejection test gates swizzled to CUDA-non-ROCm** (`is_swizzled_scales = device.type == "cuda" and not is_ROCM()`). Is the swizzled cat path expected to pass on ROCm/XPU? The cat logic itself is pure-torch and device-agnostic, but if swizzled scales aren't a supported layout on those backends the parametrization may produce failures (or validate a config that never occurs in practice). Consider gating `is_swizzled_scales` in `test_cat` the same way, or confirm it's intentional.
> 2. **Minor — redundant per-tensor `ndim` check** (`mx_tensor.py:1005-1006`): `first.ndim != 2` is already asserted at `:984`. The per-tensor loop re-checks `tensor.ndim != 2`, which is useful for catching a later 3D tensor, but the metadata tuple would already differ in most such cases. Not a bug — just noting it's slightly belt-and-suspenders.
> 3. **Consider documenting the dim-1 (contracting dim) match requirement.** It's enforced transitively by `aten.cat` on `qdata`, but an explicit check would give a clearer error message than the raw aten `RuntimeError` for the "same input, different K" mistake.
> 4. **`aten.cat.default([...], 0)` vs. the normalized `dim`**: since only 2D/dim=0 is supported and `dim` is normalized to 0 before this point, hardcoding `0` is fine — but a brief comment noting "only dim=0 reaches here" would help future readers who see the normalization logic above and the literal `0` below.
> 
> Overall this looks correct and well-tested for the intended Q/K/V and gate/up fusion use case. The main thing I'd want confirmed before merge is item **#1** (swizzled coverage on non-CUDA accelerators).
> 
> Note: the CI failure reported by Dr. CI (`test-rocm ... Could not load credentials from any providers`) is an infra/credentials issue, unrelated to this change. • Branch: `mxfp8-cat-only`

1. **Swizzled test coverage**
   Good catch. I updated `test_cat` to gate `is_swizzled_scales` consistently with the rejection test, so swizzled coverage only runs on CUDA non-ROCm.

2. **Per-tensor `ndim` check**
   I’d prefer to keep this check. The MX metadata does not include shape or `ndim`, so later inputs can still have matching metadata while being non-2D.

3. **Explicit K-dimension check**
   I’d prefer to rely on `aten.cat` here, since it already validates that all non-concatenated dimensions match. Adding a separate K check would duplicate that validation.

4. **Comment for hardcoded `dim=0`**
   I kept this as-is. Unsupported dimensions are already rejected above, so only `dim=0` reaches this code path, and an additional comment would be redundant.

xiaowangintel:
@claude review this pr.

liangan1 (approved):

xiaowangintel (approved):

vkuzo:
> Add aten.cat support for MXTensor so batch_linear_lhs can concatenate weights for linear layers sharing the same input, such as Q/K/V and gate/up projections, to reduce redundant activation quantization and GEMM launches and improve model performance.

could you link to the actual use case 

> For swizzled scales, restore the logical layout, concatenate, then repack to handle per-input padding correctly.

It would be better to only allow `cat` along swizzle atom boundaries, and throw an exception otherwise

lchen2331:
> > Add aten.cat support for MXTensor so batch_linear_lhs can concatenate weights for linear layers sharing the same input, such as Q/K/V and gate/up projections, to reduce redundant activation quantization and GEMM launches and improve model performance.
> 
> could you link to the actual use case
> 
> > For swizzled scales, restore the logical layout, concatenate, then repack to handle per-input padding correctly.
> 
> It would be better to only allow `cat` along swizzle atom boundaries, and throw an exception otherwise

Thanks for the suggestion. The direct use case is Inductor’s [`BatchLinearLHSFusion`](https://github.com/pytorch/pytorch/blob/main/torch/_inductor/fx_passes/group_batch_fusion.py#L579-L586), which concatenates weights with `torch.cat` when fusing linear layers that share the same input. This requires `MXTensor` to support `aten.cat`. Examples include Llama’s Q/K/V and gate/up projections.

I’ve also updated the implementation as suggested to only allow concatenation along swizzle atom boundaries and raise an exception for unsupported boundaries.

test/prototype/mx_formats/test_mx_tensor.py
```diff
@@ -47,6 +48,68 @@
...
+    "rows,width,dim",
+    [((128, 256), 128, 0), ((128, 256), 160, -2), ((127, 2, 129), 160, -2)],
+)
+@pytest.mark.parametrize("elem_dtype", SUPPORTED_ELEM_DTYPES)
+def test_cat(rows, width, dim, elem_dtype):
+    device = torch.accelerator.current_accelerator()
+    is_swizzled_scales = device.type == "cuda" and not is_ROCM()
+    inputs = [torch.randn(n, width, device=device, dtype=torch.bfloat16) for n in rows]
+    mx_inputs = [
+        MXTensor.to_mx(x, elem_dtype, is_swizzled_scales=is_swizzled_scales)
```
vkuzo, line 64: we should cover transposed tensors, either test that it works or throw an exception if not contiguous

lchen2331: Thanks for pointing this out. This is already covered by `test_cat_rejects_incompatible_weights`. Transposed `MXTensor`s have non-contiguous `qdata`, so `mx_cat` explicitly rejects them with `NotImplementedError`. The test verifies this with `torch.cat([first.t(), first.t()])`.

vkuzo:
@claude review

vkuzo (approved):

PR #4884, opened 2026-09-11
[a16-support] Add int16 fake quantization

insoochung:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.15.0) (oldest at bottom):
* #4891
* #4890
* #4889
* #4888
* #4887
* #4886
* #4885
* __->__ #4884
* #4883

IntxFakeQuantizeConfig rejects torch.int16 even though the affine primitives define its quantization bounds and bit width.

Allow torch.int16 and extend the per-tensor numerical test to cover dynamic and first-use static fake quantization.

Calibration and conversion behavior remain out of scope.
@exported-using-ghexport

Differential Revision: [D117600354](https://our.internmc.facebook.com/intern/diff/D117600354/)

**NOTE FOR REVIEWERS**: This PR has internal Meta-specific changes or comments, please review them on [Phabricator](https://our.internmc.facebook.com/intern/diff/D117600354/)!

Differential Revision: [D117600354](https://our.internmc.facebook.com/intern/diff/D117600354)

PR #4883, opened 2026-09-11
[a16-support] Add per-tensor fake quantization

insoochung:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.15.0) (oldest at bottom):
* #4891
* #4890
* #4889
* #4888
* #4887
* #4886
* #4885
* #4884
* __->__ #4883

IntxFakeQuantizeConfig rejects PerTensor even though the shared block-size and affine primitives support it.

Accept PerTensor objects and strings, then route IntxFakeQuantizer through the shared affine fake-quantization path. Preserve the existing dynamic and first-use static behavior.

Calibration and conversion behavior remain out of scope.
@exported-using-ghexport

Differential Revision: [D117600352](https://our.internmc.facebook.com/intern/diff/D117600352/)

**NOTE FOR REVIEWERS**: This PR has internal Meta-specific changes or comments, please review them on [Phabricator](https://our.internmc.facebook.com/intern/diff/D117600352/)!

Differential Revision: [D117600352](https://our.internmc.facebook.com/intern/diff/D117600352)

PR #4885, opened 2026-09-11
[a16-support] Add no-clipping symmetric mapping

insoochung:
* #4891
* #4890
* #4889
* #4888
* #4887
* #4886
* __->__ #4885
* #4884
* #4883

IntxFakeQuantizeConfig rejects MappingType.SYMMETRIC_NO_CLIPPING_ERR even though the affine primitives already implement it.

Allow the mapping, preserve symmetric classification, and propagate it through grouped parameter selection. Add A16 per-tensor coverage for the Quanty scale rule and grouped propagation coverage.

Calibration and conversion behavior remain out of scope.
@exported-using-ghexport

Differential Revision: [D117600355](https://our.internmc.facebook.com/intern/diff/D117600355/)

**NOTE FOR REVIEWERS**: This PR has internal Meta-specific changes or comments, please review them on [Phabricator](https://our.internmc.facebook.com/intern/diff/D117600355/)!

Differential Revision: [D117600355](https://our.internmc.facebook.com/intern/diff/D117600355)

### DIFF
--- a/torchao/quantization/qat/fake_quantize_config.py
+++ b/torchao/quantization/qat/fake_quantize_config.py
@@ -117,7 +117,10 @@ class IntxFakeQuantizeConfig(FakeQuantizeConfigBase):
                    with separate `group_size` kwarg, Alternatively, just set the
                    `group_size` kwarg and leave this field empty.
                4) 'per_tensor': equivalent to PerTensor()
-        mapping_type: whether to use symmetric (default) or asymmetric quantization
+        mapping_type: mapping from floating-point values to integers. Supported
+            values are `MappingType.SYMMETRIC`,
+            `MappingType.SYMMETRIC_NO_CLIPPING_ERR`, and
+            `MappingType.ASYMMETRIC`.
             Alternatively, set `is_symmetric` (bool) and leave this field empty.
         scale_precision: scale dtype (default torch.fp32)
         zero_point_precision: zero point dtype (default torch.int32)
@@ -129,8 +132,8 @@ class IntxFakeQuantizeConfig(FakeQuantizeConfigBase):
     Keyword args:
         group_size: size of each group in per group fake quantization,
             can be set instead of `granularity`
-        is_symmetric: whether to use symmetric or asymmetric quantization,
-            can be set instead of `mapping_type`
+        is_symmetric: whether to use symmetric or asymmetric quantization.
+            Use `mapping_type` to select `MappingType.SYMMETRIC_NO_CLIPPING_ERR`.
 
     Example usage::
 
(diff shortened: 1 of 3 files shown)


andrewor14:
@claude review

torchao/quantization/qat/fake_quantize_config.py
```diff
@@ -336,6 +358,11 @@ def __setattr__(self, name: str, value: Any):
         if name == "group_size":
             super().__setattr__("granularity", PerGroup(value))
         elif name == "is_symmetric":
+            if value and getattr(self, "mapping_type", None) in (
+                MappingType.SYMMETRIC,
+                MappingType.SYMMETRIC_NO_CLIPPING_ERR,
+            ):
+                return
```
andrewor14: Hmm I don't think we should special case this behavior. If users want `SYMMETRIC_NO_CLIPPING_ERR` then they should explicitly set `mapping_type` in the constructor instead of relying on this shorthand. This is really a convenience method

torchao/quantization/qat/fake_quantize_config.py
```diff
@@ -325,9 +340,16 @@ def group_size(self) -> int:
     @property
     def is_symmetric(self) -> bool:
         """
-        Return True if mapping type is symmetric, else False (asymmetric).
+        Return True for either symmetric mapping type.
+
+        Setting this property to True preserves an existing symmetric mapping,
+        including SYMMETRIC_NO_CLIPPING_ERR. Setting it to False selects
+        ASYMMETRIC. Set mapping_type directly to choose a specific symmetric mode.
```
andrewor14: I would prefer if we revert this behavior. We can still return True here if it's `MappingType.SYMMETRIC_NO_CLIPPING_ERR` but I don't think we should change the setter

torchao/quantization/qat/fake_quantize_config.py
```diff
@@ -117,7 +117,10 @@ class IntxFakeQuantizeConfig(FakeQuantizeConfigBase):
                    with separate `group_size` kwarg, Alternatively, just set the
                    `group_size` kwarg and leave this field empty.
                4) 'per_tensor': equivalent to PerTensor()
-        mapping_type: whether to use symmetric (default) or asymmetric quantization
+        mapping_type: mapping from floating-point values to integers. Supported
+            values are `MappingType.SYMMETRIC`,
+            `MappingType.SYMMETRIC_NO_CLIPPING_ERR`, and
+            `MappingType.ASYMMETRIC`.
             Alternatively, set `is_symmetric` (bool) and leave this field empty.
```
andrewor14, line 124: then we should update this to say: Setting `is_symmetric=True` is a shorthand for selecting `MappingType.SYMMETRIC`.

andrewor14 (changes requested):
Agreed with claude, we should fix the is_symmetric setter semantics

insoochung:
**Regarding the three comments (by @andrewor14):** Fixed default values &  docstrings to be consistent with assumption `is_symmetric=True is MappingType.SYMMETRIC`
**Regarding the claude comment:** Both `SYMMETRIC` and `SYMMETRIC_NO_CLIPPING_ERR` map to `is_symmetric=True`, so a round trip through the Boolean cannot preserve the exact mapping. Assigning `True` canonicalizes the mapping to `SYMMETRIC`. AFAIK from code search is_symmetric seem to be legacy flag - lossy roundtrip is not an issue if we stick to mapping_type enums.

andrewor14:
@claude review again?

andrewor14 (approved):
Thanks, looks great!

PR #4886, opened 2026-09-11
[a16-support] Persist static fake-quantizer state

insoochung:
* #4891
* #4890
* #4889
* #4888
* #4887
* __->__ #4886
* #4885
* #4884
* #4883

Static IntxFakeQuantizer scale and zero point values are plain attributes, so state_dict omits them and module moves leave them behind.

Register them as buffers for static non-range-learning configurations. Materialize the buffer slots when initialized state is loaded. Dynamic and range-learning behavior remains unchanged.

Older checkpoints without these qparams load with empty buffers and retain first-use initialization.
@exported-using-ghexport

Differential Revision: [D117772566](https://our.internmc.facebook.com/intern/diff/D117772566/)

**NOTE FOR REVIEWERS**: This PR has internal Meta-specific changes or comments, please review them on [Phabricator](https://our.internmc.facebook.com/intern/diff/D117772566/)!

Differential Revision: [D117772566](https://our.internmc.facebook.com/intern/diff/D117772566)

### DIFF
--- a/torchao/quantization/qat/fake_quantizer.py
+++ b/torchao/quantization/qat/fake_quantizer.py
@@ -198,8 +198,18 @@ class IntxFakeQuantizer(FakeQuantizerBase):
         torch._C._log_api_usage_once("torchao.quantization.qat.IntxFakeQuantizer")
         self.config = config
         self.enabled = True
-        self.scale: Optional[torch.Tensor] = None
-        self.zero_point: Optional[torch.Tensor] = None
+        self.scale: Optional[torch.Tensor]
+        self.zero_point: Optional[torch.Tensor]
+        if config.is_dynamic or config.range_learning:
+            self.scale = None
+            self.zero_point = None
+        else:
+            self.register_buffer(
+                "scale", torch.empty(0, dtype=config.scale_precision)
+            )
+            self.register_buffer(
+                "zero_point", torch.empty(0, dtype=config.zero_point_precision)
+            )
 
         # For range learning only
         # TODO: make this configurable?
@@ -344,7 +354,13 @@ class IntxFakeQuantizer(FakeQuantizerBase):
         """
         Return whether we need to compute new scales and zero points.
         """
-        return self.config.is_dynamic or self.scale is None or self.zero_point is None
+        return (
+            self.config.is_dynamic
+            or self.scale is None
+            or self.zero_point is None
+            or self.scale.numel() == 0
+            or self.zero_point.numel() == 0
+        )
 
     def _maybe_update_qparams_for_range_learning(self) -> None:
         """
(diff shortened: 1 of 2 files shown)


andrewor14:
@claude review

torchao/quantization/qat/fake_quantizer.py
```diff
@@ -369,6 +383,36 @@ def _maybe_update_qparams_for_range_learning(self) -> None:
...
+        local_metadata,
+        strict,
+        missing_keys,
+        unexpected_keys,
+        error_msgs,
+    ):
+        for name in ("scale", "zero_point"):
+            key = prefix + name
+            if name in self._buffers and key not in state_dict:
+                state_dict[key] = self._buffers[name]
```
andrewor14, line 404: I think this needs a comment / docstring saying that these are optional buffers. Right now we're mutating the incoming state dict, which is safe since it's a copy, but it's still a bit unexpected especially if the user specifies `strict=True`

torchao/quantization/qat/fake_quantizer.py
```diff
@@ -198,8 +198,16 @@ def __init__(self, config: IntxFakeQuantizeConfig):
...
+        self.scale: Optional[torch.Tensor]
+        self.zero_point: Optional[torch.Tensor]
+        if config.is_dynamic or config.range_learning:
+            self.scale = None
+            self.zero_point = None
+        else:
+            self.register_buffer("scale", torch.empty(0, dtype=config.scale_precision))
+            self.register_buffer(
+                "zero_point", torch.empty(0, dtype=config.zero_point_precision)
+            )
```
andrewor14: Note that this is not `torch.compile` friendly since we change buffer shapes during training. I would add a TODO comment here saying allow users to pass in expected scale and zero point shapes instead, so the buffers will remain the same shape throughout training, to make this more compatible with `torch.compile`

andrewor14, line 213: Can you add a comment here explaining how this works? Something like "Initialize empty buffers here since we do not know the qparam shapes yet, will be populated either from the state dict or from the first forward"

andrewor14: I would also add a comment explaining how this would compose with `torch.compile`. It won't work out of the box today because we resize the buffers during the first forward. The user would need to do something like:

```
# Insert fake quantizers
quantize_(model, qat_config)

# Initialize all buffers to the right shapes
with torch.no_grad():
  model(*example_inputs)

# The rest of training happens after compile
model = torch.compile(model)
```

andrewor14 (approved):
Thanks, code looks good, please add some comments to explain the subtle behavior.

torchao/quantization/qat/fake_quantizer.py
insoochung: Documented that static qparam buffers are optional for compatibility with older checkpoints.

torchao/quantization/qat/fake_quantizer.py
insoochung: 1. Documented that an eager forward must initialize static buffer shapes before torch.compile.
2. Added a comment explaining that checkpoint loading or the first eager forward initializes the empty buffer shapes.

PR #4887, opened 2026-09-11
[a16-support] Add static activation calibration

insoochung:
* #4891
* #4890
* #4889
* #4888
* __->__ #4887
* #4886
* #4885
* #4884
* #4883

Static IntxFakeQuantizer currently derives quantization parameters from its first input and has no multi-batch calibration lifecycle.

Add a static calibration lifecycle that:
- collects temporary ranges across multiple batches while fake quantization is disabled;
- supports per-tensor, per-axis, and per-group configurations with stable qparam shapes;
- finalizes persistent qparams using the configured dtype bounds; and
- restores the fake-quantization state that was active before calibration.

Dynamic, per-token, and range-learning behavior remains unchanged. Calibration ranges are temporary and do not enter the state dictionary.
@exported-using-ghexport

Differential Revision: [D117798007](https://our.internmc.facebook.com/intern/diff/D117798007/)

**NOTE FOR REVIEWERS**: This PR has internal Meta-specific changes or comments, please review them on [Phabricator](https://our.internmc.facebook.com/intern/diff/D117798007/)!

Differential Revision: [D117798007](https://our.internmc.facebook.com/intern/diff/D117798007)

### DIFF
--- a/test/quantization/test_qat.py
+++ b/test/quantization/test_qat.py
@@ -84,6 +84,7 @@ from torchao.quantization.quant_primitives import (
     ZeroPointDomain,
     _fake_quantize_affine,
     choose_qparams_affine,
+    choose_qparams_affine_with_min_max,
     dequantize_affine,
     quantize_affine,
 )
@@ -478,6 +479,128 @@ class TestQAT(TestCase):
         torch.testing.assert_close(restored(x), expected, atol=0, rtol=0)
         torch.testing.assert_close(restored.scale, fake_quantizer.scale)
         torch.testing.assert_close(restored.zero_point, fake_quantizer.zero_point)
+
+    def test_fake_quantizer_static_calibration(self):
+        activation_config = IntxFakeQuantizeConfig(
+            torch.int16,
+            PerTensor(),
+            MappingType.SYMMETRIC_NO_CLIPPING_ERR,
+            is_dynamic=False,
+        )
+        weight_config = IntxFakeQuantizeConfig(
+            torch.int4,
+            PerGroup(4),
+            MappingType.SYMMETRIC_NO_CLIPPING_ERR,
+        )
+        linear = FakeQuantizedLinear(
+            4,
+            3,
+            activation_config=activation_config,
+            weight_config=weight_config,
+        )
+        activation_fake_quantizer = linear.activation_fake_quantizer
+        weight_fake_quantizer = linear.weight_fake_quantizer
+        activation_fake_quantizer.enable_calibration()
+
+        calibration_inputs = [
+            torch.tensor([[-4.0, -1.0, 0.0, 2.0]]),
+            torch.tensor([[-2.0, 1.0, 3.0, 6.0]]),
+        ]
+        # Calibration bypasses activation fake quantization but keeps weight fake quantization enabled.
+        for x in calibration_inputs:
+            expected = F.linear(x, weight_fake_quantizer(linear.weight))
+            torch.testing.assert_close(linear(x), expected, atol=0, rtol=0)
+
+        expected_min = torch.tensor(-4.0)
+        expected_max = torch.tensor(6.0)
+        torch.testing.assert_close(activation_fake_quantizer.min_val, expected_min)
+        torch.testing.assert_close(activation_fake_quantizer.max_val, expected_max)
+        self.assertFalse(activation_fake_quantizer.enabled)
+        self.assertTrue(weight_fake_quantizer.enabled)
+
+        qmin, qmax = _DTYPE_TO_QVALUE_BOUNDS[torch.int16]
+        expected_scale, expected_zero_point = choose_qparams_affine_with_min_max(
+            expected_min,
+            expected_max,
+            MappingType.SYMMETRIC_NO_CLIPPING_ERR,
+            (),
+            torch.int16,
+            qmin,
+            qmax,
+            scale_dtype=activation_config.scale_precision,
+            zero_point_dtype=activation_config.zero_point_precision,
+        )
+        # Finalization derives fixed qparams and restores activation fake quantization.
+        activation_fake_quantizer.finalize_calibration()
+        torch.testing.assert_close(activation_fake_quantizer.scale, expected_scale)
+        torch.testing.assert_close(
+            activation_fake_quantizer.zero_point, expected_zero_point
+        )
+        self.assertFalse(activation_fake_quantizer.observer_enabled)
+        self.assertTrue(activation_fake_quantizer.enabled)
+
+        state_dict = copy.deepcopy(activation_fake_quantizer.state_dict())
+        restored = IntxFakeQuantizer(activation_config)
+        restored.load_state_dict(state_dict)
+        torch.testing.assert_close(restored.min_val, expected_min)
+        torch.testing.assert_close(restored.max_val, expected_max)
+        x = torch.tensor([[-3.0, 0.0, 1.0, 5.0]])
+        torch.testing.assert_close(
+            restored(x), activation_fake_quantizer(x), atol=0, rtol=0
+        )
+
+        # Older checkpoints load with empty ranges and can be recalibrated.
+        legacy_state_dict = copy.deepcopy(state_dict)
+        legacy_state_dict.pop("min_val")
+        legacy_state_dict.pop("max_val")
+        legacy_restored = IntxFakeQuantizer(activation_config)
+        legacy_restored.load_state_dict(legacy_state_dict)
+        self.assertEqual(legacy_restored.min_val.numel(), 0)
+        self.assertEqual(legacy_restored.max_val.numel(), 0)
+
+    def test_fake_quantizer_static_calibration_preserves_range_dtype(self):
+        config = IntxFakeQuantizeConfig(
+            torch.int16,
+            PerTensor(),
+            MappingType.SYMMETRIC_NO_CLIPPING_ERR,
+            is_dynamic=False,
+            scale_precision=torch.float32,
+        )
+        fake_quantizer = IntxFakeQuantizer(config)
+
+        for input_dtype in (torch.float16, torch.bfloat16):
+            fake_quantizer.enable_calibration()
+            fake_quantizer(torch.tensor([[-4.0, 6.0]], dtype=input_dtype))
+            self.assertEqual(fake_quantizer.min_val.dtype, torch.float32)
+            self.assertEqual(fake_quantizer.max_val.dtype, torch.float32)
+
+    @parametrize(
+        "granularity,is_dynamic,range_learning",
+        [
+            (PerTensor(), True, False),
+            (PerTensor(), False, True),
+            (PerGroup(4), False, False),
+        ],
+    )
+    def test_fake_quantizer_static_calibration_rejects_unsupported_config(
+        self,
+        granularity: Granularity,
+        is_dynamic: bool,
+        range_learning: bool,
+    ):
+        config = IntxFakeQuantizeConfig(
+            torch.int16,
+            granularity,
+            MappingType.SYMMETRIC_NO_CLIPPING_ERR,
+            is_dynamic=is_dynamic,
+            range_learning=range_learning,
+        )
+        fake_quantizer = IntxFakeQuantizer(config)
+        error = "Calibration is only supported for static per-tensor quantization"
+        with self.assertRaisesRegex(ValueError, error):
+            fake_quantizer.enable_calibration()
+        with self.assertRaisesRegex(ValueError, error):
+            fake_quantizer.finalize_calibration()
 
     def _set_ptq_weight(
         self,
--- a/torchao/quantization/qat/fake_quantizer.py
+++ b/torchao/quantization/qat/fake_quantizer.py
@@ -25,6 +25,7 @@ from torchao.quantization.quant_primitives import (
     _quantize_affine_float8,
     _Round,
     choose_qparams_affine,
+    choose_qparams_affine_with_min_max,
 )
 from torchao.quantization.utils import (
     _get_per_token_block_size,
@@ -198,6 +199,7 @@ class IntxFakeQuantizer(FakeQuantizerBase):
         torch._C._log_api_usage_once("torchao.quantization.qat.IntxFakeQuantizer")
         self.config = config
         self.enabled = True
+        self.observer_enabled = False
         self.scale: Optional[torch.Tensor]
         self.zero_point: Optional[torch.Tensor]
         if config.is_dynamic or config.range_learning:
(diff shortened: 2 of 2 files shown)


andrewor14:
@claude review

torchao/quantization/qat/fake_quantizer.py
```diff
@@ -198,6 +199,7 @@ def __init__(self, config: IntxFakeQuantizeConfig):
         torch._C._log_api_usage_once("torchao.quantization.qat.IntxFakeQuantizer")
         self.config = config
         self.enabled = True
+        self.observer_enabled = False
```
andrewor14: I think we need to document how this interacts with "self.enabled". Right now it's a bit confusing, seems like `self.enabled` is really `self.fake_quant_enabled`, but due to BC we can't just rename the flag. So the behavior as written is:

```
self.observer_enabled / self.enabled:
- True / True = observe first then return fake quantized value
- True / False = observe only, return original value
- False / True = don't observe, just fake quantize
- False / False = don't observe, return original value
```

torchao/quantization/qat/fake_quantizer.py
```diff
@@ -242,6 +263,66 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        if (
+            self.config.is_dynamic
+            or self.config.range_learning
+            or not isinstance(self.config.granularity, PerTensor)
+        ):
+            raise ValueError(
+                "Calibration is only supported for static per-tensor quantization"
+            )
+
+    def _update_calibration_ranges(self, x: torch.Tensor) -> None:
```
andrewor14: can you add some docstrings here and other places, just a quick 1 line to explain what these methods are doing

torchao/quantization/qat/fake_quantizer.py
```diff
@@ -242,6 +263,66 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+            self.config.zero_point_precision,
+        )
+        self.observer_enabled = False
+        self.enabled = True
+
+    def _validate_calibration_config(self) -> None:
+        if (
+            self.config.is_dynamic
+            or self.config.range_learning
+            or not isinstance(self.config.granularity, PerTensor)
```
andrewor14: do we want to gate on per tensor? Technically calibration can be done for other modes too right?

torchao/quantization/qat/fake_quantizer.py
```diff
@@ -242,6 +263,66 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+    def enable_calibration(self) -> None:
+        """Enable static per-tensor range collection and disable fake quantization."""
+        self._validate_calibration_config()
+        self.min_val.resize_(0)
+        self.max_val.resize_(0)
+        self.observer_enabled = True
+        self.enabled = False
+
+    def finalize_calibration(self) -> None:
+        """Finalize static per-tensor quantization parameters."""
```
andrewor14: can this docstring be a bit more informative? Something like compute qparams based on min and max vals observed during calibration. It should also explicitly mention that it disables observation and re-enables fake quant.

torchao/quantization/qat/fake_quantizer.py
```diff
@@ -242,6 +263,66 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
...
+        self._validate_calibration_config()
```
andrewor14: I think we should also assert that observer_enabled=True and enabled=False here right? I mean outside `_validate_calibration_config` but in `finalize_calibration`

torchao/quantization/qat/fake_quantizer.py
```diff
@@ -242,6 +263,66 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
         else:
             raise ValueError("Unknown granularity '%s'" % self.config.granularity)
 
+    def enable_calibration(self) -> None:
+        """Enable static per-tensor range collection and disable fake quantization."""
```
andrewor14: Nothing about this method is specific to "per-tensor", I think we can remove that from this comment

test/quantization/test_qat.py
```diff
@@ -473,6 +474,128 @@ def test_fake_quantizer_static_state_dict_per_group(self):
...
+        )
+        linear = FakeQuantizedLinear(
+            4,
+            3,
+            activation_config=activation_config,
+            weight_config=weight_config,
+        )
+        activation_fake_quantizer = linear.activation_fake_quantizer
+        weight_fake_quantizer = linear.weight_fake_quantizer
+        activation_fake_quantizer.enable_calibration()
```
andrewor14: In this test we only calibrate activations. Is the reason because your use case requires only static quant for activations? Technically users can calibrate both weight and activations at the same time right? I imagine they would do something like:

```
for module in model.modules():
    if isinstance(module, IntxFakeQuantizer):
        module.enable_calibration()
```

torchao/quantization/qat/fake_quantizer.py
```diff
@@ -208,6 +210,22 @@ def __init__(self, config: IntxFakeQuantizeConfig):
             self.register_buffer(
                 "zero_point", torch.empty(0, dtype=config.zero_point_precision)
             )
+        self.min_val: Optional[torch.Tensor]
+        self.max_val: Optional[torch.Tensor]
+        if (
+            not config.is_dynamic
+            and not config.range_learning
+            and isinstance(config.granularity, PerTensor)
```
andrewor14: same here, no need to gate on per tensor here

torchao/quantization/qat/fake_quantizer.py
```diff
@@ -208,6 +210,22 @@ def __init__(self, config: IntxFakeQuantizeConfig):
...
+        ):
+            self.register_buffer(
+                "min_val", torch.empty(0, dtype=config.scale_precision)
+            )
+            self.register_buffer(
+                "max_val", torch.empty(0, dtype=config.scale_precision)
+            )
```
andrewor14: Do these need to be buffers? Is the use case to do calibration -> save to checkpoint -> load from checkpoint -> QAT? I'm not sure how valuable this use case would be. I would prefer to keep these as simple tensors if possible to simplify the code

andrewor14 (changes requested):
Hi @insoochung, I think we should consider a different approach here. This PR in its current state couples QAT with calibration, but these should be separate processes. This complicates the QAT code significantly.

Instead, we should use the following flow:

```
nn.Linear -> ObservedLinear -> FakeQuantizedLinear
```

There is already an existing example of an observed linear called: `Float8ObservedLinear`. It achieves something very similar to what you want I think. I also think we should put this new feature under `torchao/prototype/quantization` since it would require adding a few new configs/modules.

torchao/quantization/qat/fake_quantizer.py
insoochung: `min_val` and `max_val` have been removed from buffer

torchao/quantization/qat/fake_quantizer.py
insoochung: Added a table describing the 4 possible cases involving the two flags.

torchao/quantization/qat/fake_quantizer.py
insoochung: Good point
- Now allowed: Static calibration now allows per-tensor, per-axis, and per-group configurations
- Still not allowed: Static per-token, dynamic, and range-learning configurations are unsupported - they do not provide the same fixed-shape calibration

torchao/quantization/qat/fake_quantizer.py
insoochung: Expanded the docstring to cover qparam calculation, observation shutdown, and fake-quantization state restoration.

insoochung:
Hi @andrewor14, thanks for the review. We wanted to support a specific QAT flow with the set of PRs. I'll review if moving the feature set to be based on ObservedLinear can support it and if not, will review some other options and sync with you.

insoochung:
Per @andrewor14  suggestion calibration now uses `IntxObservedLinear`... but there is a specific flow that we'd like to support with these set of PRs, I'll need some time to review if the design change still supports it. Please hold off on the review for now, thanks!

torchao/prototype/quantization/observed_qat.py
```diff
@@ -0,0 +1,213 @@
...
+
+class IntxMinMaxObserver(AffineQuantizedMinMaxObserver):
+    """Min-max observer with a reset operation for repeated QAT calibration."""
+
+    def forward(self, input: torch.Tensor) -> torch.Tensor:
+        if input.numel() == 0:
+            return input
+        observed_input = input.detach()
+        if self.scale_dtype is not None:
+            observed_input = observed_input.to(dtype=self.scale_dtype)
```
andrewor14, line 52: What's the reasoning behind this cast? The fake quantizer doesn't do this so this cast actually diverges from the numerics there, so the observation may not be as accurate

torchao/prototype/quantization/observed_qat.py
```diff
@@ -0,0 +1,213 @@
...
+        if isinstance(self.granularity, PerTensor):
```
andrewor14, line 53: Do we need this special case? I think `AffineQuantizedMinMaxObserver` already handles it

torchao/prototype/quantization/observed_qat.py
```diff
@@ -0,0 +1,213 @@
...
+        or config.range_learning
+        or not isinstance(config.granularity, (PerTensor, PerAxis, PerGroup))
+    ):
+        raise ValueError(
+            "Calibration is only supported for static per-tensor, per-axis, "
+            "or per-group activation quantization"
+        )
+
+
+class IntxMinMaxObserver(AffineQuantizedMinMaxObserver):
```
andrewor14, line 44: More generally, this class seems to mirror `AffineQuantizedMinMaxObserver` with slight differences, that makes it difficult to maintain. I'm wondering if we can just `AffineQuantizedMinMaxObserver` directly? If the "reset" functionality is the only substantial functionality gap then we can add it there in the parent class

torchao/prototype/quantization/observed_qat.py
```diff
@@ -0,0 +1,213 @@
...
+from torchao.quantization.utils import get_block_size
+
+
+def _validate_activation_config(config: IntxFakeQuantizeConfig) -> None:
+    if config.zero_point_domain != ZeroPointDomain.INT:
+        raise ValueError("Calibration requires an integer zero-point domain")
+    if (
+        config.is_dynamic
+        or config.range_learning
+        or not isinstance(config.granularity, (PerTensor, PerAxis, PerGroup))
```
andrewor14, line 36: I think if we only test `PerTensor` case then we should just drop support for `PerAxis` and `PerGroup` for now, since those are not guaranteed to work / match numerically between observer and fake quantizer

torchao/prototype/quantization/observed_qat.py
```diff
@@ -0,0 +1,213 @@
...
+        return input
+
+    def reset_min_max(self) -> None:
+        """Remove temporary ranges before a new calibration cycle."""
+        for name in ("min_val", "max_val"):
+            if hasattr(self, name):
+                delattr(self, name)
+
+
+class IntxObservedLinear(FakeQuantizedLinear):
```
andrewor14, line 91: I think this name is not representative of what the class is doing. It's really doing observer + fake quant. More specifically, it's doing observation on activations only, not weights. Maybe call it `IntxFakeQuantizedLinearWithActivationCalibration` or something?

torchao/prototype/quantization/observed_qat.py
```diff
@@ -0,0 +1,213 @@
...
+            dtype=linear.weight.dtype,
+        )
+        observed_linear.weight = linear.weight
+        observed_linear.bias = linear.bias
+        observed_linear.train(linear.training)
+        return observed_linear
+
+
+@dataclass
+class IntxObservedLinearConfig(AOBaseConfig):
```
andrewor14, line 190: Same here, maybe call it `QATWithActivationCalibrationConfig` or something to be more descriptive? Just make sure the name is consistent with what you call the linear class

torchao/prototype/quantization/observed_qat.py
```diff
@@ -0,0 +1,213 @@
...
+
+class IntxObservedLinear(FakeQuantizedLinear):
+    """Persistent QAT linear with repeatable static activation calibration."""
+
+    def __init__(
+        self,
+        in_features: int,
+        out_features: int,
+        bias: bool = False,
+        activation_config: Optional[IntxFakeQuantizeConfig] = None,
```
andrewor14, line 99: this doesn't seem optional since we're doing static fake quant there

torchao/prototype/quantization/observed_qat.py
```diff
@@ -0,0 +1,213 @@
...
+        )
+        observed_linear.weight = linear.weight
+        observed_linear.bias = linear.bias
+        observed_linear.train(linear.training)
+        return observed_linear
+
+
+@dataclass
+class IntxObservedLinearConfig(AOBaseConfig):
+    """Configuration that prepares linear modules for observed integer QAT."""
```
andrewor14, line 191: Please update these docstrings (here and in the linear class) to explicitly state that this is for activation calibration only, with the goal of doing static activation fake quant during QAT

torchao/prototype/quantization/observed_qat.py
```diff
@@ -0,0 +1,213 @@
...
+
+    def reset_min_max(self) -> None:
+        """Remove temporary ranges before a new calibration cycle."""
+        for name in ("min_val", "max_val"):
+            if hasattr(self, name):
+                delattr(self, name)
+
+
+class IntxObservedLinear(FakeQuantizedLinear):
+    """Persistent QAT linear with repeatable static activation calibration."""
```
andrewor14, line 92: Can we add an example usage showing how to use this class properly? With the different `enable_calibrations` and `finalize_calibrations` APIs, along with what state this class is in after we call each

andrewor14 (commented):
Hi @insoochung, thanks for addressing the comments. I realize that the simple transformation of `ObservedLinear -> FakeQuantizedLinear` does not address your use case, which calibrates activations while fake quantizing the weights, so these cannot be separated easily. This is a pretty specific use case and I think `torchao/prototype` is definitely a better place for it.

The latest changes look great. They are definitely in the right direction. I left some further suggestions on the observer abstractions, naming, and docs. Once you address them I can go ahead and merge this. Thanks for your work!

insoochung:
@andrewor14 
Updated the internal diffs with changes that:

- replace the feature-specific observer with the shared `AffineQuantizedMinMaxObserver`;
- rename the module to `IntxStaticActQATLinear` to match TorchAO’s existing ...QATLinear naming and identify its static-activation behavior;
- rename its config to `IntxStaticActQATConfig`;
- rename the implementation file to `static_activation_qat.py`; and
- preserve multi-batch calibration, repeatable recalibration, and the original A16W4 QAT path.

The GitHub export is currently blocked because D1–D4 are on GitHub `main`, but their DiffTrain import has not reached `fbsource`. I am waiting for that synchronization before retrying the D5 export.

PR #4894, opened 2026-09-14
Make quantize_ MoE-aware with staged Int4 expert auto-targeting

eryk-roch:
Add phase-1 MoE-aware targeting to quantize_ and expose helper APIs for loader integrations. 

- extend default matching to linear-or-MoE-expert modules
- add shared parameter target resolution (explicit name, weight, auto-discovery)
- stage auto-targeting without .weight to Int4 plain_int32 only
- add config_targets_parameter and config_prefers_cpu_checkpoint_staging
- add tests for MoE targeting heuristics and staged plain_int32 guard


xiaowangintel:
This PR is not in the right direction.

PR #4909, opened 2026-09-17
Use the ExecuTorch public threading interface

JacobSzwejbka:
## Summary

ExecuTorch 1.5 intentionally stopped publishing `extension/threadpool/threadpool.h`, but TorchAO’s ExecuTorch parallel adapter still included it. Move the adapter to the shipped `runtime/kernel/thread_parallel_interface.h` API and link the appropriate ExecuTorch runtime/threadpool target.

This preserves ExecuTorch’s configured thread count for TorchAO tiling instead of approximating it from hardware concurrency.

Depends on pytorch/executorch#22908, which adds the public thread-count accessor.

## Test plan

- Built the ExecuTorch threadpool tests and TorchAO ExecuTorch operator targets together in fbsource: 25 passed, 0 failed
- `git diff --check`

AI-assisted: Authored with Codex.


JacobSzwejbka:
oh scott got it with https://github.com/pytorch/ao/pull/4930

